PR #9875, opened 2018-11-02
don't serve lwt migrated to static themes to android

### LINKED ISSUE #9762: Find a way to stop Android being passed updates to static themes from LWTS
When the migration from LWT to static themes happens if Android Firefox requests an update it could be update from a working LWT to a non-working static theme. (hattip to @krupa) for pointing out this case)

I've confirmed with Andrew Swan that https://bugzilla.mozilla.org/show_bug.cgi?id=1430276 is platform independent so if we serve updates to android this is likely to happen.

Since support for static themes in Android is low priority we need to investigate the easiest ways to prevent the updates.

eviljeff:
fixes #9762

### DIFF
--- a/services/theme_update.py
+++ b/services/theme_update.py
@@ -4,6 +4,8 @@ import re
 from time import time
 from wsgiref.handlers import format_date_time
 
+from six import text_type
+
 from olympia.constants import base
 
 from services.utils import (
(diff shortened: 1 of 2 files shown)


services/theme_update.py
```diff
@@ -36,8 +38,8 @@ def get_headers(self, length):
 class MigratedUpdate(ThemeUpdate):
 
     def get_data(self):
-        if hasattr(self, 'row'):
-            return self.row
+        if hasattr(self, 'data'):
+            return self.data
```
diox: ! So before, we were re-executing that query ...

eviljeff: yep...

PR #9880, opened 2018-11-04
Fixes #9831, Removing Compatibility-reports and Compat_totals table

### LINKED ISSUE #9831: Remove compatibility-related database tables
Following #9515, these database tables are no longer used and should be removed:
- `compatibility_reports`
- ~~`compat`~~
- `compat_totals`

cc @wagnerand to confirm that we can do that.

miphilomath:
@diox Please review. Fixes Issue, mozilla/addons#6134.

Also, I have a query on, "Where can I see the names of all the tables in the database present if they are public?"

### DIFF
--- a/src/olympia/migrations/1053-remove-compatibility-related-tables.sql
+++ b/src/olympia/migrations/1053-remove-compatibility-related-tables.sql
@@ -1,2 +1,3 @@
-DROP TABLE 'compatibility_reports';
-DROP TABLE 'compat_totals';
+DROP TABLE IF EXISTS 'compatibility_reports';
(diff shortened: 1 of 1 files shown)


src/olympia/migrations/1053-remove-compatibility-related-tables.sql
```diff
@@ -0,0 +1,2 @@
+DROP TABLE 'compatibility_reports';
+DROP TABLE 'compat_totals';
```
championshuttler: New line at the end of the file please :)

miphilomath: Done!

PR #9887, opened 2018-11-05
Remove 'boost-webextensions-in-search' waffle switch

### LINKED ISSUE #9886: Remove `boost-webextensions-in-search` waffle
I noticed that the `boost-webextensions-in-search` waffle switch is still used - it's been `on` for ages, never flipped back, and now that we are about to remove legacy add-ons and only show compatible add-ons to users anyway, it's kinda useless. What's worse, it's checking for the waffle at search time (my fault), affecting performance of the search API a little.

Let's remove it and the associated boost.

diox:
Fix #9886

### DIFF
--- a/src/olympia/migrations/1054-remove-boost-webextensions-in-search-waffle.sql
+++ b/src/olympia/migrations/1054-remove-boost-webextensions-in-search-waffle.sql
@@ -1,0 +1,1 @@
+DELETE FROM waffle_switch_mkt WHERE name = 'boost-webextensions-in-search';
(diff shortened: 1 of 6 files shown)


src/olympia/migrations/1054-remove-boost-webextensions-in-search-waffle.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM waffle_switch_mkt WHERE name = 'boost-webextensions-in-search';
```
eviljeff: `waffle_switch_mkt`?
whatyearisit

diox: Lol messed up my copypasting :)

PR #9905, opened 2018-11-05
Make exact matches against add-on names in search language-aware

### LINKED ISSUE #6898: Search: Support translations for exact name matches
Follow-up issue to #6837 which implemented basic exact matching support.

We need to extend that support for translations. We already have a `name_translations` field so let's see if we can re-use that in a good way, otherwise let's think about restructuring things.

diox:
This makes us try to do an exact match against both the raw name in the current language, and the raw name in the default locale for each add-on (only boosting once, thanks to a `dis_max` query).

The same kind of logic is also added to description and summary matches.

Extra debugging features (search query naming, API returning matched_queries information, commented out script to update expected scores in test_search_ranking.py) are included but shouldn't affect production (or dev/stage) environments.

Fix #6898

### DIFF
--- a/src/olympia/search/tests/test_search_ranking.py
+++ b/src/olympia/search/tests/test_search_ranking.py
@@ -1,23 +1,42 @@
 # -*- coding: utf-8 -*-
 import json
 
+from django.conf import settings
+
 from olympia import amo
 from olympia.amo.tests import (
     APITestClient, ESTestCase, reverse_ns, create_switch)
+from olympia.constants.search import SEARCH_LANGUAGE_TO_ANALYZER
 
 
 class TestRankingScenarios(ESTestCase):
     client_class = APITestClient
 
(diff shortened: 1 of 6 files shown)


src/olympia/search/tests/test_search_ranking.py
```diff
@@ -1,23 +1,42 @@
...
+        def get_name_from_result(item, expected_lang):
+            name = item['name'].get(expected_lang)
+            if name is None:
+                name = item['name'].get(item['default_locale'], '??????')
+            return name
+
+        # Use v4dev version to ensure we get objects for translations all the
+        # time. We don't necessarily specify the language in all tests, but we
+        # want objects all the time for simplicity.
+        url = reverse_ns('addon-search', api_version='v4dev')
```
eviljeff: Heh, _something_ uses the v4dev api finally :tada:

Issue #9857, opened 2018-10-31
Use "Edit Product Page" instead of "Edit Listing" or "Edit Information"


jvillalobos:
What we used to call "listing page" is now being called "product page". To keep the terminology consistent, we need to update some references to it in the devhub.

- [ ] In the logged in view in `/developers/`, rename the `Edit Listing` link as `Edit Product Page`.
- [ ] In the logged in view in `/developers/`, next to the `Welcome back` title, replace `creating an appealing listing` for `creating an appealing product page`.
- [ ] In `/developers/addons/` and `/developers/themes/`, rename the `Edit Information` link as `Edit Product Page`.
- [ ] In the Edit view for an add-on (e.g. `/developers/addon/amo-admin-assistant/edit`), change the page title to `Edit Product Page`.
- [ ] In the Edit view for an add-on (e.g. `/developers/addon/amo-admin-assistant/edit`), rename the `Edit Information` link in the sidebar as `Edit Product Page`. This applies to all edit pages that use this sidebar.

AlexandraMoga:
Verified fixed on AMO -dev with FF63, Win10x64
> * In the logged in view in `/developers/`, rename the `Edit Listing` link as `Edit Product Page`
> * In the logged in view in `/developers/`, next to the `Welcome back` title, replace `creating an appealing listing` for `creating an appealing product page`.
![image](https://user-images.githubusercontent.com/31961530/48474752-e3407b80-e803-11e8-8a15-354d0de00465.png)

> * In `/developers/addons/` and `/developers/themes/`, rename the `Edit Information` link as `Edit Product Page`.
![image](https://user-images.githubusercontent.com/31961530/48474839-18e56480-e804-11e8-9639-fcae025bfd51.png)


> * In the Edit view for an add-on (e.g. `/developers/addon/amo-admin-assistant/edit`), change the page title to `Edit Product Page`.
> * In the Edit view for an add-on (e.g. `/developers/addon/amo-admin-assistant/edit`), rename the `Edit Information` link in the sidebar as `Edit Product Page`. This applies to all edit pages that use this sidebar.
![image](https://user-images.githubusercontent.com/31961530/48474906-3adee700-e804-11e8-801e-6320ea7d7202.png)

I think there are more sections that require this change, so I filed a follow up to address those as well.

PR #9924, opened 2018-11-07
"Edit Listing|Information" -> "Edit Product Page"

eviljeff:
quick fix for #9857

### DIFF
--- a/src/olympia/devhub/templates/devhub/addons/listing/item_actions_theme.html
+++ b/src/olympia/devhub/templates/devhub/addons/listing/item_actions_theme.html
@@ -4,7 +4,7 @@
     <li>
       <a href="{{ addon.get_dev_url() }}" class="tooltip"
          title="{{ _("Edit details about this theme's listing.") }}">
-        {{ _('Edit Listing') }}</a>
+        {{ _('Edit Product Page') }}</a>
     </li>
   {% endif %}
   <li>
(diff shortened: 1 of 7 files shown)


diox:
Nota: don't forget to run the l10n extraction script after making that change (we haven't really ran it recently cause it wasn't needed)

eviljeff:
> Nota: don't forget to run the l10n extraction script after making that change (we haven't really ran it recently cause it wasn't needed)

I don't follow.  How is this different commit than any other change to the templates, so requires the l10n extraction script to be specifically run?

diox:
It's not any different, I haven't committed any changes to templates recently so I haven't run it, but realized with this change that we should :)

src/olympia/devhub/templates/devhub/addons/listing/item_actions_theme.html
```diff
@@ -4,7 +4,7 @@ <h5>{{ _('Actions') }}</h5>
     <li>
       <a href="{{ addon.get_dev_url() }}" class="tooltip"
          title="{{ _("Edit details about this theme's listing.") }}">
```
wagnerand: Can you update the tooltip here too, please?

eviljeff: https://github.com/mozilla/addons-server/pull/9924/commits/dc25ec385d52cde50abd032d068270f3e63196b7

PR #9929, opened 2018-11-07
Add option for fetch_prod_addons command to fetch specific guid(s)

### LINKED ISSUE #9928: Add option allowing fetch_prod_addons to fetch specific guid(s)
`fetch_prod_addons` only fetches the ~ 25k most popular add-ons by default, because of our own limitations, so it can be missing a few key add-ons. It would be useful if it could fetch specific guid(s) that the user would be passing on the command line to test some known scenarios.

diox:
Fix #9928

### DIFF
--- a/src/olympia/landfill/management/commands/fetch_prod_addons.py
+++ b/src/olympia/landfill/management/commands/fetch_prod_addons.py
@@ -22,8 +22,12 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
-            '--max', metavar='MAX', type=int,
+            '--max', metavar='max', type=int,
             help='max amount of pages to fetch.'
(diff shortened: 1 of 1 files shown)


src/olympia/landfill/management/commands/fetch_prod_addons.py
```diff
@@ -25,6 +25,10 @@ def add_arguments(self, parser):
             '--max', metavar='MAX', type=int,
```
eviljeff: why is this `MAX` and later it's `options.get('max')`?

diox: Good question :) I don't know, it's copied from EnTeQuAk's script, but I'll fix it and see if it still works :)

PR #9935, opened 2018-11-08
Make fetch_prod_addons import strict compatibility info and exclude legacy add-ons

### LINKED ISSUE #9934: Avoid importing legacy add-ons in fetch_prod_addons
`fetch_prod_addons` command doesn't filter out legacy add-ons. It allowed me to debug issues reported a while ago with legacy add-ons, but now the legacy add-ons are getting in the way. This is especially important as it does not import `strict_compatibility` info, so I'm seeing legacy add-ons in my local searches even with specifying `app` and `appversion` correctly...

We could just wait until we got rid of legacy add-ons entirely, which should happen any day now, but might as well fix it to make things more reliable.

diox:
Fix #9934

### DIFF
--- a/src/olympia/landfill/management/commands/fetch_prod_addons.py
+++ b/src/olympia/landfill/management/commands/fetch_prod_addons.py
@@ -29,6 +29,10 @@ class Command(BaseCommand):
             '--guid', metavar='guid', type=str,
             help='specific guid(s) to fetch.'
         )
+        parser.add_argument(
+            '--type', metavar='type', type=str,
+            help='only consider this specific add-on type'
+        )
 
     def handle(self, *args, **options):
         if not settings.DEBUG:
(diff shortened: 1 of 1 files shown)


src/olympia/landfill/management/commands/fetch_prod_addons.py
```diff
@@ -70,6 +74,10 @@ def _handle_addon(self, addon_data):
                 'platform': amo.PLATFORM_DICT[files[0]['platform']].id,
                 'size': files[0]['size'],
                 'is_webextension': files[0]['is_webextension'],
+                'is_mozilla_signed_extension': (
+                    files[0]['is_mozilla_signed_extension']),
+                'strict_compatibility': (
+                    version['is_strict_compatibility_enabled'])
             }
```
eviljeff: Is there anything else we're not importing?  Maybe there could be a test that would break if we added something to the serializer and didn't added it to here/an exclusion list?

diox: A bit actually yeah. There is a TODO in the file to do license, ratings, previews, android compat data, category info. And well, since the stuff I added in that PR was not part of that TODO, I guess more is still missing ;) Ultimately though, everything we need to test search relevancy is already there now I think, at least for Firefox add-ons.

Issue #9912, opened 2018-11-06
Add timing metrics to LWT migration


eviljeff:
We know roughly how long 500/1000 LWT take* to migrate, from watching kibana logs, but we don't know how long exactly each step in the migration takes.   Once we know how long things take we can look to optimize the slow parts.

*1000 LWT took 65 minutes on -dev last time it was run.  Though, a) there have been a number of changes since then and b) prod could|should be quicker because more resources.

AlexandraMoga:
@eviljeff should this be a qa:not needed?

diox:
@eviljeff can you add your findings in this issue ?

eviljeff:
datadog screenshot:
![screenshot 2018-11-11 14 04 36](https://user-images.githubusercontent.com/768592/48314210-2c06f300-e5be-11e8-9a50-12d780f36128.png)

diox:
It's 03:38 am and I can't sleep... can't really code either, but `utc_millesecs_from_epoch` *is* suspicious:
```python
def utc_millesecs_from_epoch(for_datetime=None):
    """
    Returns millesconds from the Unix epoch in UTC.

    If `for_datetime` is None, the current datetime will be used.
    """
    if not for_datetime:
        for_datetime = datetime.datetime.now()
    return calendar.timegm(for_datetime.utctimetuple()) * 1000
```

That's not good... 

```
>>> from olympia.amo.utils import utc_millesecs_from_epoch
>>> from datetime import datetime, timedelta
>>> now = datetime.now()
>>> later = now + timedelta(milliseconds=42)
>>> now == later
False
>>> utc_millesecs_from_epoch(now) == utc_millesecs_from_epoch(later)
True
>>> utc_millesecs_from_epoch(later)
1541989568000
>>> utc_millesecs_from_epoch(now)
1541989568000
```

Something like the following doesn't lose the precision (I don't know if there is a better way to do this - our dates are currently always UTC, so this should be fine?):
```python
def utc_millesecs_from_epoch(for_datetime=None):
    if not for_datetime:
        for_datetime = datetime.datetime.now()
return int(float(for_datetime.strftime('%s.%f')) * 1000)
```

eviljeff:
Mmm, parsing a string to get back the seconds as a float... that seems... indirect.

[...]

PR #9937, opened 2018-11-08
Add more detailed timing in add_static_theme_from_lwt 

eviljeff:
follow on from #9912 because all the time was spent in the `add_static_theme_from_lwt` method and need to get breakdown


eviljeff:
#yolo'ing this because it's only timing for a command that's only currently being run on -dev.  I can fix any problems in a follow-up :p

Issue #9867, opened 2018-11-01
Initial setup: Indexation never finishes


wagnerand:
refs #9856

For me, indexation never finishes. I set up a vanilla macOS 10.14 machine and followed the instructions above. The last (descriptive) output I get is `Running all indexation tasks`. After that, it's just dots. I had it running for about an hour, then canceled it. Even after restarting the containers, I get `CommandError: Indexation already occurring - use --force to bypass` when trying to run `./manage.py reindex`. Adding the `--force` option will start the process, but again it doesn't seem to finish.

[docker-celery.log](https://github.com/mozilla/addons-server/files/2538821/docker-celery.log)

[container_web.log](https://github.com/mozilla/addons-server/files/2538835/container_web.log)

[container_es.log](https://github.com/mozilla/addons-server/files/2538836/container_es.log)

OS: macOS 10.14.1
Docker: 18.06.1-ce-mac73 (26764)

diox:
This should now be fixed in master @willdurand @wagnerand . Let me know if it works for you.

diox:
Also, just in case, I'd still advise doing a `manage.py reindex --wipe` after you update to master, if you're not completely starting from scratch, just in case your index is in a weird shape because of that bug.

PR #9939, opened 2018-11-08
Make reindex create the index+alias properly even where there is nothing to index

diox:
Fix https://github.com/mozilla/addons-server/issues/9867

### DIFF
--- a/src/olympia/lib/es/tests/test_commands.py
+++ b/src/olympia/lib/es/tests/test_commands.py
@@ -24,6 +24,9 @@ class TestIndexCommand(ESTestCase):
         # created during this test run.
         self.indices = self.es.indices.stats()['indices'].keys()
 
+        self.addons = []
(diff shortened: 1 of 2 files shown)


src/olympia/lib/es/tests/test_commands.py
```diff
@@ -24,6 +24,9 @@ def setUp(self):
         # created during this test run.
         self.indices = self.es.indices.stats()['indices'].keys()
 
+        self.addons = []
+        self.expected = self.addons[:]
```
eviljeff: I like the consistency, but this is just ` self.expected = []` 😄

diox: Yeah, I went back and forth changing `self.addons` in `setUp` instead of the test method and in the end left `self.expected` as is, eh, it's working :)

PR #9942, opened 2018-11-08
Add 'is_custom' flag to License object in version detail

### LINKED ISSUE #9914: Please return a flag for a custom license in the add-on detail API response
Following up on https://github.com/mozilla/addons-frontend/issues/3697#issuecomment-436254019, we currently get the following for the license in `current_version` from the [add-on detail API](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#detail):
- id
- name
- url

If the license is a custom license (i.e., the text of the license is stored with the version information), then the only indication we have of that is that the `url` points back to the app (e.g., `https://addons-dev.allizom.org/en-US/firefox/addon/adblock-edge/license/1.6b`). 

We would like to also get an explicit indication that the version contains a custom license in this case. This could mean adding a property to `current_version` for `has_custom_license` or something along those lines.

eviljeff:
fixes #9914


eviljeff:
@bobsilverberg the addon detail endpoint includes the version detail for `current_version`, which would include the license.

bobsilverberg:
> @bobsilverberg the addon detail endpoint includes the version detail for current_version, which would include the license.

Ok, I misinterpreted. I thought because I saw `license.text` right next to `license.is_custom` that we'd only get the latter when we used the version detail endpoint. I see now that there's a comment that explains that `license.text` is omitted from add-on detail endpoint.

Thanks!

PR #9949, opened 2018-11-09
Validate that source files are actual working archives

### LINKED ISSUE #9572: Validate submitted source code
https://github.com/mozilla/addons-server/blob/d71d63cc1e350ddebe495cda6ea4247edcb35414/src/olympia/versions/models.py#L42-L44 allows different file formats for uploading source code.

We rely on them being valid for doing code-search (and review of course), but we don't do any kind of validation (other than checking the file extension). The result is, that there are some files either being corrupt/incomplete or have the wrong file extension (presumably to make the upload pass), but in fact use an unsupported compression algorithm.

Therefore, we need to add some validation to uploaded source code files:
* Check the compression algorithm, for example using the file signature (magic number), instead of just the file extension and
* Verify the archive is valid.

Looking into [pyunpack](https://github.com/ponty/pyunpack) might be worthwhile.

Justification for P2: The daily code-extraction script that provides the data for code-search relies on valid files and usage of one of the supported compression algorithms above. We recently mitigated some issues, but we cannot anticipate or support every potential compression algorithm.

diox:
Fix #9572

### DIFF
--- a/src/olympia/devhub/forms.py
+++ b/src/olympia/devhub/forms.py
@@ -37,7 +37,8 @@ from olympia.translations.forms import TranslationFormMixin
 from olympia.translations.models import Translation, delete_translation
 from olympia.translations.widgets import (
     TranslationTextarea, TranslationTextInput)
-from olympia.versions.models import ApplicationsVersions, License, Version
+from olympia.versions.models import (
+    VALID_SOURCE_EXTENSIONS, ApplicationsVersions, License, Version)
 
 from . import tasks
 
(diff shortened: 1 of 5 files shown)


src/olympia/devhub/forms.py
```diff
@@ -288,12 +291,32 @@ def save(self, commit=True):
...
-                    'file {extensions}.'.format(
-                        extensions=VALID_SOURCE_EXTENSIONS)))
+        if source:
+            try:
+                if source.name.endswith('.zip'):
+                    zip_file = SafeZip(source)
+                    # testzip() returns None if there are no broken CRCs.
+                    if zip_file.zip_file.testzip() is not None:
+                        raise zipfile.BadZipfile()
+                elif source.name.endswith(('.tar.gz', '.tar.bz2')):
```
eviljeff: Could we do something that `VALID_SOURCE_EXTENSIONS` here instead?  Seems weird we have that constant but don't actually use to determine what is a valid extension.

eviljeff: I thought something like `elif os.path.splitext(source.name)[1] in VALID_SOURCE_EXTENSIONS:` but yeah could as easily be removed entirely.

diox: `elif os.path.splitext(source.name)[1] in VALID_SOURCE_EXTENSIONS` doesn't separate `.tar.foo` from `.zip` though, and it'd be kinda wonky to do an explicit `.zip` check above and not below, so... yeah, best avoided I think. I'll add a commit to try to get rid of that constant entirely.

diox: Nope, I have to keep it, or something like that. We need the extension and `splitext` doesn't handle `.tar.gz` (how could it...) so we have to hardcode a check against .tar.gz somewhere. Might as well keep this then. It does have a comment next to the constant so hopefully that will be enough.

PR #9971, opened 2018-11-13
Postpone statistics processing by a couple hours

bqbn:
Data pipeline may fail and retry, when that happens the cron may exit early
due to non-existing S3 files.

* [X] Add a description of the the changes introduced in this PR.

### DIFF
--- a/scripts/crontab/crontab.tpl
+++ b/scripts/crontab/crontab.tpl
@@ -34,17 +34,17 @@ HOME=/tmp
 
 # Update ADI metrics from S3.
 # Once per day after 0800 UTC
-30 9 * * * %(django)s update_counts_from_file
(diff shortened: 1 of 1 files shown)


eviljeff:
Only a few weeks ago these crons started at 08:00 and now they'll be starting at 11:30 - did something change that means things are often 3 1/2 hours late?

bqbn:
What I've noticed is that the data pipeline has to re-try processing the logs more often recently. When they retry, data for our cron jobs are delayed.

The purpose of this PR is to reduce the chance that we have to manually re-run the cron jobs on our end.

EnTeQuAk:
Can or should we land this anytime soon?

bqbn:
This can land upon approval.

PR #9982, opened 2018-11-13
Treat an empty locale string in FeaturedCollection as None

### LINKED ISSUE #9980: Featured themes & extensions are empty on dev
Featured themes & extensions are empty on dev. At first I blamed my changes from https://github.com/mozilla/addons-server/issues/9770 and reverted them, but it didn't help. I think it's related to https://github.com/mozilla/addons-server/issues/9964 instead - not the changes themselves, but something else in the admin that was triggered when QA verified that issue.

diox:
This ensures that whatever happens, a `FeaturedCollection` with an empty locale is considered as such.

Fix #9980


diox:
Yeah. But I wasn't entirely sure of the impact outside of ES, so I focused on that for now.

PR #9988, opened 2018-11-14
expose rating flags in detail response when`?include_flags_for=<user>`

### LINKED ISSUE #7037: Show that a review has already been flagged (Review API)

Here are STRs on `addons-frontend`

* Sign in as a developer
* Go to one of your add-ons
* Read all reviews
* Flag some reviews as bug reports, etc
* Get a snack, switch computers (or just reload the page)
* Sign in again to finish flagging reviews

### What happened?

As soon as you reload the page, there is no indication that you have flagged any reviews.

the UI should show that you already flagged a review so you don't have to flag it again.

We need additional info from the API for this.

Originally reported by a developer here: https://github.com/mozilla/addons-frontend/issues/3952

Maybe the [review object](https://addons-server.readthedocs.io/en/latest/topics/api/ratings.html#detail) can include a flagged reason for the authenticated user but I know things like that introduce caching challenges. We could query a separate endpoint but this would add an additional API request of course.

eviljeff:
fixes #7037

### DIFF
--- a/docs/topics/api/ratings.rst
+++ b/docs/topics/api/ratings.rst
@@ -27,7 +27,7 @@ user has already posted a rating for the current version of an add-on.
 .. http:get:: /api/v4/ratings/rating/
 
     :query string addon: The :ref:`add-on <addon-detail>` id, slug, or guid to fetch ratings from. When passed, the ratings shown will always be the latest posted by each user on this particular add-on (which means there should only be one rating per user in the results), unless the ``version`` parameter is also passed.
-    :query string include_flags_for: The user id to show flags for. If the request is made with an authenticated user matching this parameter value, a ``flags`` property will be added to the response as described below in :ref:`ratings <rating-detail-object>`.
+    :query string show_flags_for: The user id to show flags for. If the request is made with an authenticated user matching this parameter value, a ``flags`` property will be added to the response as described below in :ref:`ratings <rating-detail-object>`.
     :query string show_permissions_for: The user id to show permissions for. If the request is made with an authenticated user matching this parameter value, and the ``addon`` parameter is also present, a ``can_reply`` property will be added to the response as described below.
     :query string filter: The :ref:`filter(s) <rating-filtering-param>` to apply.
     :query int user: The user id to fetch ratings from.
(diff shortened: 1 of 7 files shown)


docs/topics/api/ratings.rst
```diff
@@ -78,12 +79,16 @@ This endpoint allows you to fetch a rating by its id.
...
     .. _rating-detail-object:
 
+    :query string include_flags_for: The user id to show flags for. If the request is made with an authenticated user matching this parameter value, a ``flags`` property will be added to the response as described below.
     :>json int id: The rating id.
     :>json object addon: A simplified :ref:`add-on <addon-detail-object>` object that contains only a few properties: ``id``, ``name``, ``icon_url`` and ``slug``.
     :>json string|null body: The text of the rating.
     :>json boolean is_deleted: Boolean indicating whether the rating has been deleted or not.
     :>json boolean is_latest: Boolean indicating whether the rating is the latest posted by the user on the same add-on.
     :>json int previous_count: The number of ratings posted by the user on the same add-on before this one.
+    :>json object flags[]: A list of flags the user requesting has previously applied to this add-on (that haven't been processed by moderators already).  In the current implementation the list will only ever contain one item, or be an empty list if there are no flags from the user.  Only present if ``include_flags_for`` parameter sent.
```
kumar303: typo: ...has previously applied to this rating (not add-on)

kumar303: Saying that the list will only ever contain one item seems unimportant to me.

docs/topics/api/ratings.rst
```diff
@@ -27,6 +27,7 @@ user has already posted a rating for the current version of an add-on.
 .. http:get:: /api/v4/ratings/rating/
 
     :query string addon: The :ref:`add-on <addon-detail>` id, slug, or guid to fetch ratings from. When passed, the ratings shown will always be the latest posted by each user on this particular add-on (which means there should only be one rating per user in the results), unless the ``version`` parameter is also passed.
+    :query string include_flags_for: The user id to show flags for. If the request is made with an authenticated user matching this parameter value, a ``flags`` property will be added to the response as described below in :ref:`ratings <rating-detail-object>`.
```
diox: Can't we combine it with `show_permissions_for` (possibly renaming it in the process) ? Or at least, use the same nomenclature ? I don't care that much about which verb to use, but we should make up our mind between `show` and `include` if we are to have 2 different parameters.

eviljeff: Could change it to `show_flags_for`.  I considered combining it but it doesn't require the `addon` parameter to be present - it could be used for a list of all ratings for a user too.  If we _do_ want to replace `show_permissions_for` with a combined `show_flags_and_permissions_for`* then we'd either have to drop the requiring `addon` validation check or (unnecessarily) require it when you just wanted flag info. 

*(Worst. Name. Ever)

diox: That's a good point. Let's keep the 2 separate, `show_permissions_for` really needs an `addon`, it'd be awkward to let it work without. Since we've started with the word `show` I'd be in favor of following that, so `show_flags_for` gets my vote (I'm biased because I'm the one who introduced `show_permissions_for`, though :)

eviljeff: https://github.com/mozilla/addons-server/pull/9988/commits/f4bf3d514e65fb61a6dafafc552539de0bcfadc3

docs/topics/api/ratings.rst
eviljeff: https://github.com/mozilla/addons-server/pull/9988/commits/f4bf3d514e65fb61a6dafafc552539de0bcfadc3

src/olympia/ratings/views.py
```diff
@@ -412,6 +433,7 @@ def get_paginated_response(self, data):
                     'the user id of the authenticated user')
             extra_data['can_reply'] = (
                 self.check_can_reply_permission_for_ratings_list())
+        self.should_include_flags()
```
diox: Why do we need that here ?

eviljeff: `should_include_flags` has the validation in it so if we don't call it here (or somewhere else in the view explicitly) then when there aren't any results to serialize an invalid request (e.g. they aren't authenticated, the `show_flags_for` parameter doesn't match the user) isn't raised in the API response.

This test https://github.com/mozilla/addons-server/pull/9988/files/f4bf3d514e65fb61a6dafafc552539de0bcfadc3#diff-dbe5b121611cfb846d3adfc78246f12aR1309 and the other two tests on invalid list requests fail.

diox: Ok, that'll need a comment :)

eviljeff: https://github.com/mozilla/addons-server/pull/9988/commits/a844a565ea9d2169cffae6d6a5404e2e9e3717ce

PR #9999, opened 2018-11-15
allow ratings api to be filtered by rating score

### LINKED ISSUE #9173: Get add-on reviews / ratings by star number
We would like the [rating listing API](https://addons-server.readthedocs.io/en/latest/topics/api/ratings.html#get--api-v4-ratings-rating-) to return only reviews for a given star (e.g. one star reviews, five star reviews, etc). 

This will support https://github.com/mozilla/addons-frontend/issues/5923 which is a UI for clicking stars to filter reviews.

The add-on review listing now has a graph that shows a breakdown of reviews by star number. We would like to allow users to click on the star graph and see only reviews by that star number. For example, they could click on the bottom bar to see only one star reviews. This will help users make better decisions about whether to install the add-on or not.

eviljeff:
fixes #9173

### DIFF
--- a/docs/topics/api/ratings.rst
+++ b/docs/topics/api/ratings.rst
@@ -27,13 +27,14 @@ user has already posted a rating for the current version of an add-on.
 .. http:get:: /api/v4/ratings/rating/
 
     :query string addon: The :ref:`add-on <addon-detail>` id, slug, or guid to fetch ratings from. When passed, the ratings shown will always be the latest posted by each user on this particular add-on (which means there should only be one rating per user in the results), unless the ``version`` parameter is also passed.
-    :query string show_flags_for: The user id to show flags for. If the request is made with an authenticated user matching this parameter value, a ``flags`` property will be added to the response as described below in :ref:`ratings <rating-detail-object>`.
-    :query string show_permissions_for: The user id to show permissions for. If the request is made with an authenticated user matching this parameter value, and the ``addon`` parameter is also present, a ``can_reply`` property will be added to the response as described below.
-    :query string filter: The :ref:`filter(s) <rating-filtering-param>` to apply.
-    :query int user: The user id to fetch ratings from.
-    :query boolean show_grouped_ratings: Whether or not to show ratings aggregates for this add-on in the response (Use "true"/"1" as truthy values, "0"/"false" as falsy ones).
-    :query int version: The version id to fetch ratings from.
     :query string exclude_ratings: Exclude ratings by their ``id``. Multiple ratings can be specified, separated by comma(s).
+    :query string filter: The :ref:`filter(s) <rating-filtering-param>` to apply.
+    :query string score: Only include ratings that have been given a specific ``score``. Multiple scores can be specified, separated by comma(s).
+    :query string show_flags_for: The user id to show flags for. If the request is made with an authenticated user matching this parameter value, a ``flags`` property will be added to the response as described below in :ref:`ratings <rating-detail-object>`.
+    :query boolean show_grouped_ratings: Whether or not to show ratings aggregates for this add-on in the response (Use "true"/"1" as truthy values, "0"/"false" as falsy ones).
+    :query string show_permissions_for: The user id to show permissions for. If the request is made with an authenticated user matching this parameter value, and the ``addon`` parameter is also present, a ``can_reply`` property will be added to the response as described below.
+    :query int user: The user id to fetch ratings from.
+    :query int version: The version id to fetch ratings from.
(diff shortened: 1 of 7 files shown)


src/olympia/ratings/views.py
```diff
@@ -389,6 +393,14 @@ def filter_queryset(self, qs):
                 # because the frontend wants to call this before and after
                 # having posted a new rating, and needs accurate results.
                 self.pagination_class = OneOrZeroPageNumberPagination
+            if score_filter:
+                try:
+                    scores = [int(score) for score in score_filter.split(',')]
+                except ValueError:
```
kumar303: You may need to handle non-ascii input, too

```
Python 2.7.10 (default, Feb  7 2017, 00:08:15)
>>> int(u'Krstić')
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'decimal' codec can't encode character u'\u0107' in position 5: invalid decimal Unicode string
```

docs/topics/api/ratings.rst
```diff
@@ -27,13 +27,14 @@ user has already posted a rating for the current version of an add-on.
 .. http:get:: /api/v4/ratings/rating/
 
     :query string addon: The :ref:`add-on <addon-detail>` id, slug, or guid to fetch ratings from. When passed, the ratings shown will always be the latest posted by each user on this particular add-on (which means there should only be one rating per user in the results), unless the ``version`` parameter is also passed.
+    :query string exclude_ratings: Exclude ratings by their ``id``. Multiple ratings can be specified, separated by comma(s).
+    :query string filter: The :ref:`filter(s) <rating-filtering-param>` to apply.
+    :query string score: Only include ratings that have been given a specific ``score``. Multiple scores can be specified, separated by comma(s).
```
kumar303: 👍 

I like the comma separated scores

src/olympia/ratings/views.py
eviljeff: They're from `request.GET` so the input should already be encoded as a byte string.  It's easily testable - it's the same method that `exclude_ratings` uses parses the input so if you can't 500 on `exclude_ratings=Krstić` then it's fine.

diox: Actually it does - It's just that `UnicodeEncodeError` - which is raised here - is inheriting from `ValueError`.

docs/topics/api/ratings.rst
eviljeff: Considered it, but how?  `filter` looks like `filter=with_deleted,without_empty_body` but score needs values for which star rating(s) you want to include.  Double `=` like `filter=score=2` (if that's even supported)?  Some other syntax like `:` (that we'd only use here)?

diox: Mmm true, that's annoying. I wish I had thought of that when adding `filter`... I would have maybe used a more clever format... Ignore this then.

src/olympia/migrations/1055-review-indexes-on-deleted-and-rating.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX `reviews_deleted_894a7213` ON `reviews` (`deleted`);
+CREATE INDEX `reviews_rating_0f3b945f ` ON `reviews` (`rating`);
```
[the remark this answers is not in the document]
eviljeff: https://github.com/mozilla/addons-server/pull/9999/commits/651ae3cbeb858d7bf0a84799e91f087372ed0448

src/olympia/migrations/1055-review-indexes-on-deleted-and-rating.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX `reviews_deleted_addon_id_rating_3d18fa24_idx` ON `reviews` (`deleted`, `addon_id`, `rating`);
+CREATE INDEX `reviews_deleted_user_id_rating_b865f159_idx ` ON `reviews` (`deleted`, `user_id`, `rating`);
```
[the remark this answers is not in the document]
diox: Actually... testing locally with different scenarios and lots of ratings to mimic prod, I can't figure out a good index for it. We'd need `reply_to` and `created` in there as well as they are part of the most frequent queries... Locally and even on stage, no matter what I do, with or without the filtering on `rating`, MySQL seems to insist on using the index on `addon` alone instead. We could force it, but I'm hesitant to do it, I'm not sure we'd do a better job than MySQL's query planner.

Let's drop these indexes entirely for now - don't add *any* index in that PR, and we'll monitor the API & database performance and see if something needs to be done later.

Issue #9804, opened 2018-10-25
Search: searching for "Website vertrauen" doesn't find WOT - Sicher surfen highest in search results


AlexandraMoga:
Follow up for #6838 

1. Load AMO stage
2. Do the following search: https://addons.allizom.org/de/firefox/search/?q=Website%20vertrauen

Expected result:
[WOT - Sicher surfen](https://addons.allizom.org/de/firefox/addon/wot-safe-browsing-tool/?src=search) is found highest in search results list, because it includes the "Website vertrauen" string in its description

Actual result:
[WOT - Sicher surfen](https://addons.allizom.org/de/firefox/addon/wot-safe-browsing-tool/?src=search) can be found only on [page 3](https://addons.allizom.org/de/firefox/search/?page=3&q=Website%20vertrauen)

![image](https://user-images.githubusercontent.com/31961530/47484019-50837100-d843-11e8-905b-b78787d28b15.png)

diox:
The results that appear first have "Website" in their name, but not "vertrauen". Need to investigate this more, I wonder if this is because of the fuzzy match on the name (which does *not* specify `"operator": "and"`). Will know more by reproducing locally and looking at the explanation for the score.

AlexandraMoga:
Verified fixed on AMO stage with FF63, Win10x64

WOT - Sicher surfen is now found when searching for "Website vertrauen"

![image](https://user-images.githubusercontent.com/31961530/48825634-a3801380-ed70-11e8-9098-c5dd75a44e24.png)

PR #10004, opened 2018-11-16
Decrease prefix for fuzzy name matching in search, but force all terms to be found

### LINKED ISSUE #8207: Searching for an Image toggle addon and the correct results are buried.
I came across this this post
https://www.reddit.com/r/firefox/comments/8huosl/help_addin_to_hide_images_in_firefox_59_and_higher/
saying, in part, "no matter how much I search for I can not find a working plugin that switches off/on showing/loading of images in web pages."

I did a search on AMO: https://addons.mozilla.org/en-US/firefox/search/?q=toggle+images 
using "toggle" instead of hide and only one hit was near the bottom of the page that was relevant.  ( I believe the original legacy addon (or one of them at least) used the word toggle.)
I sorted through the first 7 of 19 pages and found a total of 8 relevant hits (that all do about the same thing).  Here's the addons if it's helpful.

https://addons.mozilla.org/en-US/firefox/addon/image-block-x/
https://addons.mozilla.org/en-US/firefox/addon/image-video-block/
https://addons.mozilla.org/en-US/firefox/addon/easy-image-blocker/
https://addons.mozilla.org/en-US/firefox/addon/image-blocker-plus/
https://addons.mozilla.org/en-US/firefox/addon/block-image-video/
https://addons.mozilla.org/en-US/firefox/addon/hide-images-per-tab/
https://addons.mozilla.org/en-US/firefox/addon/image-hide/
https://addons.mozilla.org/en-US/firefox/addon/block-images-webextensions/ 

This seems a bit rubbish notably because of how modern users currently work. They're very used to Google quality search results, seldom make it to the bottom of the page and almost never go to page 2.

I've noticed several users on reddit complaining in recent history of poor search results on AMO.

diox:
Going to help (but not completely fix) #8207
Fix https://github.com/mozilla/addons/issues/567
Fix https://github.com/mozilla/addons-server/issues/9804

### DIFF
--- a/src/olympia/search/tests/test_search_ranking.py
+++ b/src/olympia/search/tests/test_search_ranking.py
@@ -473,8 +473,8 @@ class TestRankingScenarios(ESTestCase):
             average_daily_users=566337, weekly_downloads=150000,
             summary=None,
             description=(
-                'button, click that button, 1-Click Youtube Video '
-                'Downloader is a click click great tool')),
+                'This addon contains Amazon 1-Click Lock in its description '
+                ' but not in its name.')),
         amo.tests.addon_factory(
             name='Amazon 1-Click Lock', type=amo.ADDON_EXTENSION,
             average_daily_users=50, weekly_downloads=1, summary=None)
@@ -484,8 +484,11 @@ class TestRankingScenarios(ESTestCase):
     def test_scenario_tab_center_redux(self):
         self._check_scenario('tab center redux', (
             ['Tab Center Redux', 69.5371],
-            ['Tab Mix Plus', 0.06526235],
-            ['Redux DevTools', 0.044507127],
+            # Those used to be found but we now require all terms to be present
+            # through operator: and on the fuzzy name query (and they have
+            # nothing else to match).
+            # ['Tab Mix Plus', 0.06526235],
+            # ['Redux DevTools', 0.044507127],
         ))
 
     def test_scenario_open_image_new_tab(self):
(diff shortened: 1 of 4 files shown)


src/olympia/search/tests/test_search_ranking.py
```diff
@@ -484,8 +484,11 @@ def setUpTestData(cls):
     def test_scenario_tab_center_redux(self):
         self._check_scenario('tab center redux', (
             ['Tab Center Redux', 69.5371],
-            ['Tab Mix Plus', 0.06526235],
-            ['Redux DevTools', 0.044507127],
+            # Those used to be found but we now require all terms to be present
+            # through operator: and on the fuzzy name query (and they have
+            # nothing else to match).
+            # ['Tab Mix Plus', 0.06526235],
+            # ['Redux DevTools', 0.044507127],
```
eviljeff: What would now happen when searching for `tab center`?  Would it match `Tab Center Redux` still? What about searching for `tab center pro` (or similar?)

diox: `tab center` does match `tab center redux` easily - both terms match, don't even need the fuzzy filter, it also matches the prefix rule and the match phrase rule, so tab center redux would be fairly high in the results. Locally it's the first result for me.

`tab center pro` wouldn't find it however, and that's the biggest change. Locally it doesn't find anything anymore, because that last bit, "pro", doesn't match anything that's also matching the first two. On prod, at the moment, it does find a bunch of add-ons, some relevant, some not. Those 3 issues I linked all complain about this behavior, so that's why I'm changing it here.

Do you think it's sensible ? You can test locally too, if you import prod add-ons using `manage.py fetch_prod_addons` (takes a while)

diox: As discussed on IRC, I've made the fuzzy query less rigid in https://github.com/mozilla/addons-server/pull/10004/commits/3a51d5cc99e15f399c6d2727d4e2989e4e7d4ec7 with a `minimum_should_match` instead of the `operator: and`.

PR #10017, opened 2018-11-19
Added footer link tests.

jrbenny35:
Adds footer link tests that run only on the desktop resolution.


tests/ui/pages/desktop/home.py
```diff
@@ -10,7 +10,7 @@ class Home(Base):
     _extensions_category_locator = (By.CLASS_NAME, 'Home-CuratedCollections')
     _featured_extensions_locator = (By.CLASS_NAME, 'Home-FeaturedExtensions')
     _featured_themes_locator = (By.CLASS_NAME, 'Home-FeaturedThemes')
-    _hero_locator = (By.CLASS_NAME, 'HomeHeroBanner')
+    _hero_locator = (By.CLASS_NAME, 'Hero-name-HomeHeroGuides')
```
willdurand: I am sorry but this fixes the symptoms and not the root cause. Addons-frontend should NOT be run in development mode for the uitests.

PR #10027, opened 2018-11-20
Fix utc_millesecs_from_epoch to not discard milliseconds

### LINKED ISSUE #9961: fix `utc_millesecs_from_epoch` method
It loses millisecond accuracy.  To copy from #9912:
```
>>> from olympia.amo.utils import utc_millesecs_from_epoch
>>> from datetime import datetime, timedelta
>>> now = datetime.now()
>>> later = now + timedelta(milliseconds=42)
>>> now == later
False
>>> utc_millesecs_from_epoch(now) == utc_millesecs_from_epoch(later)
True
>>> utc_millesecs_from_epoch(later)
1541989568000
>>> utc_millesecs_from_epoch(now)
1541989568000
```

diox:
Fix #9961

### DIFF
--- a/src/olympia/amo/utils.py
+++ b/src/olympia/amo/utils.py
@@ -989,7 +989,7 @@ def utc_millesecs_from_epoch(for_datetime=None):
     seconds = time.mktime(for_datetime.utctimetuple())
     # timetuple() doesn't care about more precision than seconds, but we do.
     # Add microseconds as a fraction of a second to keep the precision.
-    seconds += for_datetime.microsecond / 1000000.
+    seconds += for_datetime.microsecond / 1000000.0
     # Now convert to milliseconds.
     return int(seconds * 1000)
 
--- a/src/olympia/amo/utils.py
+++ b/src/olympia/amo/utils.py
@@ -1,4 +1,3 @@
-import calendar
 import collections
 import contextlib
 import datetime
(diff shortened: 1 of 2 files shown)


src/olympia/amo/utils.py
```diff
@@ -986,7 +985,13 @@ def utc_millesecs_from_epoch(for_datetime=None):
     """
     if not for_datetime:
         for_datetime = datetime.datetime.now()
-    return calendar.timegm(for_datetime.utctimetuple()) * 1000
+    # Number of seconds.
+    seconds = time.mktime(for_datetime.timetuple())
```
eviljeff: should this be `utctimetuple` again?

diox: Yeah. Ultimately it doesn't matter as we only deal with UTC dates in addons-server, but we never know...

src/olympia/amo/utils.py
```diff
@@ -986,7 +985,13 @@ def utc_millesecs_from_epoch(for_datetime=None):
     """
     if not for_datetime:
         for_datetime = datetime.datetime.now()
-    return calendar.timegm(for_datetime.utctimetuple()) * 1000
+    # Number of seconds.
+    seconds = time.mktime(for_datetime.utctimetuple())
+    # timetuple() doesn't care about more precision than seconds, but we do.
+    # Add microseconds as a fraction of a second to keep the precision.
+    seconds += for_datetime.microsecond / 1000000.
```
eviljeff: Are you _sure_ this works as expected?  I checked in a shell and this expression always resolves to an integer value, meaning we instantly lose the fraction of a second you just added.  You'd need `float(...)` around the microsecond value, afaik.  Or have `seconds` be `milliseconds` on the line above and just divide `.microsecond` by `1000` instead

diox: `microsecond` is converted into a float by dividing with a float `1000000` has the dot at the end.

diox:
```python
>>> seconds = time.mktime(for_datetime.utctimetuple())
>>> seconds
1542711112.0
>>> seconds += for_datetime.microsecond / 1000000.
>>> seconds
1542711112.344284
```

Everything looks ok ?

eviljeff: doh, I missed the `.`

eviljeff: `1000000.0` for skim readers like myself?

PR #10042, opened 2018-11-21
Enable memcached by default for tests. Upgrade python-memcached

### LINKED ISSUE #8602: Use memcached in our unittests by default
We are using LocMemCache in our unittests, potentially for performance reasons. The problem is that we won't run into any of memcached problems, e.g a max cache-key length of 250 characters and others. (e.g https://github.com/mozilla/addons-server/issues/8598)

Let's use memcached by default.

EnTeQuAk:
* Adds a test to test unicode support for latest python-memcached
  versions
* Configures caches key prefix so that tests don't conflict with each other
* Remove a few cache.clear calls with `cache.delete` calls that avoids
deleting all memcached entries while other tests are running
* Remove zadmin.memcached as it's crazy dangerous, such power should
only be wielded by ops-folks who wake up at night when pressed
* Fixes waffle helpers to call .flush() to flush appropriate caches

This also fixes how parameters are forwarded to memcached and fix KEY_PREFIX initialization

We are now making sure parameters from `settings.CACHES` are properly forwarded
to the actual cache backend (now memcached). These errors were not seen
since LocMemCache didn't require any parameters.

The `KEY_PREFIX` setting is set to a uuid instead of the xdist-slaveid to
ensure test functions are properly isolated from each other.

`cache.clear` is not called anywhere unless absolutely necessary (which is nowhere right now) to reduce
the side-effects of our test-setup.

Fixes #8602


src/olympia/amo/tests/__init__.py
```diff
@@ -228,6 +228,9 @@ def create_sample(name=None, **kw):
     if not created:
         sample.__dict__.update(kw)
         sample.save()
+
+    sample.flush()
```
diox: Shouldn't the cache be flushed when saving ? Same for the other 2 ?

src/olympia/bandwagon/models.py
```diff
@@ -30,7 +30,7 @@ class TopTags(object):
     """Descriptor to manage a collection's top tags in cache."""
 
     def key(self, obj):
-        return '%s:top-tags:%s' % (settings.CACHE_PREFIX, obj.id)
+        return 'top-tags:{id}'.format(id=obj.id)
```
diox: We should create an issue to remove this, I don't think the frontend cares about top tags at all...

src/olympia/amo/tests/__init__.py
EnTeQuAk: I think this may even be old code, I went back and forth on this quite a lot so I may have missed it. I'll double-check and will probably remove it.

django-waffle does indeed call `.flush()` on save and it shouldn't be necessary for the create case.

PR #10044, opened 2018-11-21
restrict screenshot upload dimensions and ratio

### LINKED ISSUE #9487: Restrict screenshot upload dimensions and ratio
For add-on submission and editing, the screenshots need to meet these restrictions:
- 4x3 width to height ratio.
- PNG or JPEG, PNG recommended
- 1000x750 px minimum
- 2400x1800 px recommended _and_ maximum

eviljeff:
fixes #9487


src/olympia/devhub/views.py
```diff
@@ -953,6 +953,22 @@ def ajax_upload_image(request, upload_type, addon_id=None):
                 errors.append(ugettext('Image must be exactly {0} pixels '
                                        'wide and {1} pixels tall.')
                               .format(expected_size[0], expected_size[1]))
+        content_waffle = waffle.switch_is_active('content-optimization')
+        if check.is_image() and upload_type == 'preview' and content_waffle:
+            min_size = amo.ADDON_PREVIEW_SIZES.get('min')
+            # * 100 to get a nice integer to compare against rather than 1.3333
+            required_ratio = min_size[0] * 100 / min_size[1]
+            with storage.open(loc, 'rb') as fp:
+                actual_size = Image.open(fp).size
```
EnTeQuAk: Hmm, this view is opening the image twice now in case of a persona, this should be unified and things like fetching the actual size should be integrated into the `ImageCheck` class or restructure that in another way.

Also, calling `check.is_image` twice opens the actual image again and calls `.verify()` which we're doing twice now, all that can also be a DOS vector in the end.

I get that this isn't part of the PR directly but at the very least we should file a higher priority issue to clean this up.

eviljeff: Btw, there aren't any persona uploads any more, so what happens with personas doesn't matter. (Code will be removed once the migration is done and we definitely don't need to upload any to test something)

EnTeQuAk: True, that code is still executed though, even though `.is_animated()` call and others, this view is opening and processing that one image multiple times, which is really bad.

eviljeff: https://github.com/mozilla/addons-server/pull/10044/commits/be4c761d94f93cc127e669081c063c614407da16

PR #10059, opened 2018-11-23
have combined name and summary length restriction

### LINKED ISSUE #9243: Restrict length of add-on name and summary
We need to limit this during the submission flow and in the developer hub management pages.

eviljeff:
fixes #9243


eviljeff:
> It's kinda sad that django validators don't properly work with multiple fields so this mixin is actually needed.

yeah, django form/model validation is lacking compared to DRF

PR #10062, opened 2018-11-23
Drop obsolete addon properties

### LINKED ISSUE #9785: Dev Hub Edit Information page contains some sections that are no longer used
STR:

1. Log in to Developer Hub
2. Go to Manage My submissions and select an add-on to open its Edit information page
3. Scroll down to the Technical Details section and notice the available edit options

Actual result:
Developer Comments, Required Add-ons and Add-on Flags don't seem to be used anymore in the new listing pages. The Required Add-ons search field is also returning suggestions of disabled legacy add-ons which should o longer exist on the site.

Expected result:
If there is no future plan to use these fields, they should be removed from the Edit Information page.

Cc @jvillalobos for his input

image

eviljeff:
fixes #9785 - there's a migration to nullify `auto_repackage` and `externalsoftware` columns but there will need to be a follow-up that actually drops the columns and deletes the `addons_dependencies` table.


EnTeQuAk:
:fire:

PR #10084, opened 2018-11-27
removes 2 columns and table

### LINKED ISSUE #10080: drop auto_repackage, externalsoftware columns from addons; drop addons_dependencies table
follow-up from #9785 - needs to be merged _after_ 2018-11-29

championshuttler:
Fixes #10080


eviljeff:
just bumped the migration filename because 1060 was taken in the meantime.

PR #10118, opened 2018-11-29
autocrop name/summary in non-default locales on submit

### LINKED ISSUE #9784: remove old submission flow feature that prepopulates description field with cropped summary text in en-US
Follow up for #9489 

STR:
1. Log in to AMO dev
2. Create a new extension package and make sure that the `"description"` in your manifest has more than 250 characters
3. Submit your add-on
4. Observe the Summary and Description fields content in the details submission step

Actual result:
A portion of the `"description"`provided is displayed in the Summary field, while the Description has the entire text, including the portion displayed as Summary. The character count in the Description field is 0.

Expected result:
Since dev hub is limiting the summary to 250 characters, maybe a check should be included in the linter (?) to make sure that the manifest `"description"` has the required length. 

Notes:
- For non-extension types, where the Description field is not present during submission, Summary will include the entire manifest `"description"` , but the field will specify by how many characters the maximum length is exceeded.
- reproduced on AMO -dev with FF62, Win10x64

description_submission

### LINKED ISSUE #10079: add-ons with long descriptions in the manifest cause problems on upload
followup for #9243 - only a bug with `content-optimization` waffle enabled.

On upload we parse the add-on for any names and descriptions in any locales, and then use them to populate the name and description of the add-on on AMO.  We then show a single locale to the developer in the upload flow (the default locale).  If the name + summary are >70 characters then the developer is forced to change them.  But if the name + summary > 70 in _other_ locales then a validation error is raised on the form but the developer _can't_ change them.

Some options:
a) auto-crop the summary (and/or name) in other locales automatically.  We could show a warning to the developer on the form at least.
b) don't validate the lengths of the other locales.  Could show a warning on the page.  Easiest but probably causes problems down the line.
c) change devhub submit flow to support editing multiple locales (an unknown amount of engineering work + needs UX so would have to be a Q1 task.  Also would need a) or b) to happen if the waffle was to be enabled before then)

@jvillalobos

eviljeff:
fixes #10079 (+ fixed #9784 while I was at it)

### DIFF
--- a/src/olympia/devhub/forms.py
+++ b/src/olympia/devhub/forms.py
@@ -5,7 +5,7 @@ import zipfile
 
 from django import forms
 from django.conf import settings
-from django.core.validators import MinLengthValidator
+from django.core.validators import MaxLengthValidator, MinLengthValidator
 from django.db.models import Q
 from django.forms.models import BaseModelFormSet, modelformset_factory
 from django.forms.widgets import RadioSelect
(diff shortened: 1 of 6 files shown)


EnTeQuAk:
I restarted travis, let's see, r+ once it's green (we may have to restart the py3 tasks for that)

PR #10128, opened 2018-11-30
Remove super common 2 letter words from search dictionary_decompounder

### LINKED ISSUE #10127: Unexpected search results for "websocket" searchterm

I searched for the term _websocket_: https://addons.mozilla.org/en-US/firefox/search/?q=websocket&appver=64.0&platform=mac

### What happened?

The few top results look ok, but then I see lots of unrelated results:
* Language packs
* Personas Plus
* a theme about yahoo
* a spell checker
* ...

screenshot_2018-11-30 search results for websocket add-ons for firefox en-us

Seeing only _websocket_ related add-ons.

diox:
'so' and 'up' are far too common and makes us return bad matches too
easily.

Fix #10127


diox:
Yeah I don't love having to adjust the expected scores every time (although I have it automated) - In the future I might just make it mandatory in a couple tests and instead of the score expose which queries match in each test and assert on that.

PR #10139, opened 2018-12-03
Increase weight by 2000 if add-on uses coinminer flag.

### LINKED ISSUE #7052: Increase weight of add-ons flagged for using coin miners
This is a followup for https://github.com/mozilla/addons-linter/issues/1643. This flag should give increased weight for all add-ons that have it.

It should have a weight of 20, and it shouldn't add up (multiple instances is the same as one instance).

EnTeQuAk:
Fixes #7052


wagnerand:
Since the issue was filed initially, we changed a lot of the weights. The maximum weight is now much higher which means adding 200 does not guarantee the add-on showiing up on page 1 (or even close to the top of the queue).

Therefore, I suggest we increase the weight be 2000.

EnTeQuAk:
Thanks @wagnerand, I changed the weight. re-r? @eviljeff / @wagnerand?

PR #10142, opened 2018-12-03
Proposal: Reverse 'Continue' and 'Cancel and Disable...' buttons.

### LINKED ISSUE #7556: Pressing Enter will disable the submission process instead of finish it
Steps to reproduce:
1.Go to dev hub and start a submission for a new webextension
2.After a successful validation, press "Submit version" button without completing all mandatory fields(ie. categories or change the slug if it was used before)
3.Complete the mandatory fields and press Enter keyboard button

Expected results:
The webextension is successfully submitted

Actual results:
Submission is cancelled and webextension is in an incomplete state

NOTES:
Tested using Windows 10x64, FF Release, AMO-dev
Reproducible on all servers

Please see video for this bug(Enter was pressed at the end of video) : 
enter

EnTeQuAk:
Personally, the current order adds a lot of stress to the user to click
the correct button.

When you upload your file (very first screen) the Continue button is on
the very left side, all continuous screens switch it to be on the right
hand site. Even though the 'Cancel' button is red, it takes a bit of
time to parse which button you're going to press and ending up in a
disabled/incomplete state isn't really nice.

This changes that behaviour and fixes #7556 along the way.

screenshot from 2018-12-03 13-07-10

cc @jvillalobos or @wagnerand for opinions on that.


wagnerand:
I had the exact same experience a couple of times, so I am in favor of changing it.

EnTeQuAk:
Switched the buttons for theme submission too.

screenshot from 2018-12-04 07-48-19

EnTeQuAk:
Seems this PR is actually breaking the CircleCI tests, I'll test this a bit more

src/olympia/devhub/templates/devhub/addons/submit/describe.html
```diff
@@ -235,13 +235,13 @@ <h3>{{ _('Describe Add-on') }}</h3>
       </div>
     {% endif %}
     <div class="submission-buttons addon-submission-field">
-      <button class="delete-button" type="submit" formnovalidate
-              formaction="{{ url('devhub.addons.cancel', addon.slug) }}">
-          {{ _('Cancel and Disable Version') }}
+      <button type="submit">
+        {{ _('Continue') }}
```
jrbenny35: does this do the same as the old code? ```{{ _('Submit Version') }}```?

EnTeQuAk: Good idea, I copied the wrong copy. Thanks!

EnTeQuAk:
> Hmm, the test fallure doesn't seem to be related to the changes, unless something else has changed as a result. Looking at the html of the error page, the id id_name doesn't exist though, so something is fishy.

That's indeed weird, the `id_name` is definitely there in the templates though and nothing I touched here.

EnTeQuAk:
Interesting, so according to the screenshot the devhub upload test ends up with an incomplete submission…

EnTeQuAk:
ah ha! Found it, I was only looking at the upload but `DevhubSubmission` uses `.submission-buttons > button:nth-child(2)` which is now wrong, let's see…

PR #10145, opened 2018-12-03
Remove raven.js from legacy + devhub pages.

### LINKED ISSUE #4845: Consider not reporting frontend JS errors to Sentry
Currently, any Javascript error gets reported to Sentry here: https://sentry.prod.mozaws.net/operations/frontend-prod/

@jasonthomas says that this generates a lot of traffic to Sentry which is one reason that Sentry is slow to render new events (for any other project).

Perhaps we could turn off frontend JS errors until we need them? In other words, maybe we could introduce a waffle switch that can be toggled only when someone knows about an error they would like to capture in prod. There are downsides to this of course but maybe they are worth it.

EnTeQuAk:
Unless I'm mistaken there isn't anyone to really look at this and most
errors and problems should be related to legacy pages anyway.

Let's remove things we're not using or looking at. I'm happy to discuss this further though.

Fixes #4845

Once this lands, ops can remove the projects from sentry.


kumar303:
This stops reporting all Devhub JS errors FYI.

EnTeQuAk:
Certainly true, we never looked at them though and sentry overreported quite a lot of things. Given that, it's best to rethink the current situation and maybe end up with a different system / different reporting strategy. That's why I went ahead and removed it for now, we can rethink how we do reporting on JS errors for devhub once legacy pages are removed and we have a clear picture on what pages we'll keep in the future.

kumar303:
Yep, seems like a good plan (I certainly wasn't looking at the errors 😁 ). I just wanted to clarify that this removes error reporting from more than just legacy pages.

Issue #10150, opened 2018-12-03
Fix recent circleci test failures


EnTeQuAk:
It seems something rating related changed on addons-frontend.

```

base_url = 'http://olympia-frontend.test:3000'
selenium = <selenium.webdriver.firefox.webdriver.WebDriver (session="4403ff9b-8557-481d-9739-9417b18cff64")>
category = 'Top Rated', sort_attr = 'rating'

    @pytest.mark.parametrize('category, sort_attr', [
        ['Most Users', 'users'],
        ['Top Rated', 'rating']])
    def test_sorting_by(base_url, selenium, category, sort_attr):
        """Test searching for an addon and sorting."""
        Home(selenium, base_url).open()
        addon_name = 'Ui-addon'
        selenium.get('{}/search/?&q={}&sort={}'.format(
            base_url, addon_name, sort_attr)
        )
        search_page = Search(selenium, base_url)
        results = [getattr(i, sort_attr)
>                  for i in search_page.result_list.extensions]

addon_name = 'Ui-addon'
base_url   = 'http://olympia-frontend.test:3000'
category   = 'Top Rated'
i          = <pages.desktop.search.ResultListItems object at 0x7f08eb0aa290>
search_page = <pages.desktop.search.Search object at 0x7f08eb0fead0>
selenium   = <selenium.webdriver.firefox.webdriver.WebDriver (session="4403ff9b-8557-481d-9739-9417b18cff64")>
sort_attr  = 'rating'

tests/ui/test_search.py:47: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pages.desktop.search.ResultListItems object at 0x7f08eb0aa290>

    @property
    def rating(self):
        """Returns the rating"""
        rating = self.find_element(
            *self._rating_locator).get_property('title')
>       return int(rating.split()[1])
E       ValueError: invalid literal for int() with base 10: 'are'

rating     = 'There are no ratings yet'
self       = <pages.desktop.search.ResultListItems object at 0x7f08eb0aa290>

tests/ui/pages/desktop/search.py:81: ValueError
------------------------------- pytest-selenium --------------------------------
Driver log: /tmp/pytest-of-seluser/pytest-0/test_sorting_by_Resolution__101/driver.log
[...]

PR #10152, opened 2018-12-03
Xfail filter search tests.

jrbenny35:
Search tests are failing due to a change with filters. I want to xfail these and talk to @diox about a good fix.


EnTeQuAk:
Fixes https://github.com/mozilla/addons-server/issues/10150 - should be added to the commit message

Issue #8302, opened 2018-05-18
Stop using username


jvillalobos:
AMO has a username field and a display name field. The distinction exists because usernames used to be for authentication, and they had stricter formatting requirements (like no spaces). The display name was meant as a more user-friendly name that would fall back to the username if not set.

Since we're using FxA for auth on AMO, this distinction doesn't make sense anymore. To address issues like https://github.com/mozilla/addons-server/issues/8296, I think we should merge both fields and have a single display name field with the same rules as the current display name, with the addition of uniqueness as a restriction.

eviljeff:
the username is used in urls though

eviljeff:
What should happen for currently non-unique display names?

Tbh, it would be a lot more straightforward to drop `display_name` instead

jvillalobos:
> Tbh, it would be a lot more straightforward to drop display_name instead

How about this:

1) Drop `display_name`.
2) Relax the rules for `username` so they are "unique `display_name`".
3) Notify developers about the changes.

This way we don't break existing URLs and developers get the choice to switch back to their nicer names if they don't mind the change in their user URLs. Display names can be encoded so they work as slugs, right?

eviljeff:
I did some local testing.  Our current username validation rejects any spaces and most special characters (anything other than letters+numbers+`-`+`~`).  However, when I edited via django-admin, which bypassed any validation, I was able to set a username such as `This is a special username!` which worked in the url fine(*).  

So we can pretty much allow anything, but encoding "abnormal" characters results in crappy looking urls when encoded, so we usually don't.  It can also complicate things like passing the username as a parameter (we do it in some of the APIs) as a comma, ampersand, etc would need to be encoded to stop them from breaking something.  The biggest risk is existing code in addons-server or addons-frontend that does an append or string substitution without encoding the parts first (worst case I can think of is it might allow masquerading as another username).  So if we relax from letters+numbers+`-`+`~` I would advise the least relaxation possible to have "nice" usernames)

(*) Firefox displayed it in the url bar as `/en-GB/firefox/user/This is a special username!/` and it's encoded as `/en-GB/firefox/user/This%20is%20a%20special%20username!/` when shown as text.

eviljeff:
@jvillalobos what would be the minimum relaxing that would be needed to have "nice" usernames?  Is allowing spaces sufficient?  We already support mixed case.

jvillalobos:
[...]

PR #10173, opened 2018-12-05
stop using usernames in collections too

### LINKED ISSUE #10172: support ids / drop usernames from collections
we currently only support usernames in all collection related (legacy) views, and check username, instead of id, in the collection APIs.  This needs to be changed to support either just support id, or support id and username equally so #8302 can happen

eviljeff:
fixes #10172 and supports #8302


src/olympia/bandwagon/templates/bandwagon/user_listing.html
```diff
@@ -1,6 +1,6 @@
 {% extends "base.html" %}
 
-{# L10n: {0} is a username. #}
+{# L10n: {0} is a the user's name. #}
```
diox: Extra `a`

PR #10206, opened 2018-12-12
Add additional search tests

jrbenny35:
*Check that search suggestion list exact matches higher than other suggestions
*Check that long terms searching don't break results page
*Verify blank search functionality
*Check that selecting an item from search suggestions can be made by clicking on it
*Check that ESC dismisses the suggestion search results
*Check that suggestions change if chars are added or deleted
*Check that special chars don't break search
*Check that caps and lowercases have same search suggestions

These test cases can be found [here](https://testrail.stage.mozaws.net/index.php?/suites/view/2816&group_by=cases:section_id&group_id=24581&group_order=asc) if you have access.

### DIFF
--- a/tests/ui/test_search.py
+++ b/tests/ui/test_search.py
@@ -1,5 +1,10 @@
 # -*- coding: utf-8 -*-
+import time
+
 import pytest
+from selenium.common.exceptions import NoSuchElementException
+from selenium.webdriver.common.action_chains import ActionChains
+from selenium.webdriver.common.keys import Keys
 
 from pages.desktop.home import Home
 from pages.desktop.search import Search
@@ -9,7 +14,7 @@ from pages.desktop.search import Search
 def test_search_loads_and_navigates_to_correct_page(base_url, selenium):
     page = Home(selenium, base_url).open()
     addon_name = page.featured_extensions.list[0].name
-    search = page.header.search_for(addon_name)
+    search = page.search.search_for(addon_name)
     search_name = search.result_list.extensions[0].name
     assert addon_name in search_name
     assert search_name in search.result_list.extensions[0].name
@@ -19,7 +24,7 @@ def test_search_loads_and_navigates_to_correct_page(base_url, selenium):
 def test_search_loads_correct_results(base_url, selenium):
     page = Home(selenium, base_url).open()
     addon_name = page.featured_extensions.list[0].name
-    items = page.search_for(addon_name)
+    items = page.search.search_for(addon_name)
     assert addon_name in items.result_list.extensions[0].name
 
 
@@ -27,7 +32,7 @@ def test_search_loads_correct_results(base_url, selenium):
 def test_legacy_extensions_do_not_load(base_url, selenium):
     page = Home(selenium, base_url).open()
     term = 'Video Download Manager'
-    items = page.search_for(term)
+    items = page.search.search_for(term)
     for item in items.result_list.extensions:
         assert term not in item.name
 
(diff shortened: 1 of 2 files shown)


EnTeQuAk:
There are a few flake8 issues that should be resolved, looking at https://travis-ci.org/mozilla/addons-server/jobs/467120724#L1086-L1095

tests/ui/test_search.py
```diff
@@ -53,8 +58,87 @@ def test_sorting_by(base_url, selenium, category, sort_attr):
...
+
+@pytest.mark.nondestructive
+def test_search_suggestion_term_is_higher(base_url, selenium):
+    page = Home(selenium, base_url).open()
+    term = 'Ui-Addon'
+    items = page.search.search_for(term, execute=False)
+    for count, item in enumerate(items):
+        if item.name == term:
+            assert count < 1
+
```
EnTeQuAk: Don't we have a finite number of items we expect here? It'd be better to explicitly test them instead of a `for` loop that wouldn't fail in case we only see one item (for whatever reason).

Also, I don't quite get that loop, isn't that almost equivalent to just writing ``assert items[0].name == term``? `count < 1` will only be true for the very first entry.

tests/ui/test_search.py
```diff
@@ -53,8 +58,87 @@ def test_sorting_by(base_url, selenium, category, sort_attr):
...
+    # Sleep to let autocomplete update.
+    time.sleep(2)
+    assert term in items[0].name
+
+
+@pytest.mark.nondestructive
+def test_special_chars_dont_break_suggestions(base_url, selenium):
+    page = Home(selenium, base_url).open()
+    term = 'Ui-Addon-Install'
+    special_chars = '$%^&*'
```
EnTeQuAk: Let's add some fancy unicode characters to the mix?

tests/ui/test_search.py
```diff
@@ -53,8 +58,87 @@ def test_sorting_by(base_url, selenium, category, sort_attr):
...
+@pytest.mark.nondestructive
+def test_capitalization_has_same_suggestions(base_url, selenium):
+    page = Home(selenium, base_url).open()
+    term = 'Ui-Addon-Install'
+    items = page.search.search_for(term, execute=False)
+    assert term == items[0].name
+    items = page.search.search_for(term.capitalize(), execute=False)
+    # Sleep to let autocomplete update.
+    time.sleep(2)
+    assert term == items[0].name
```
EnTeQuAk: Maybe add a test that `term.lower()` leads to the same results?

Generally, I think we should even be able to say `items == items` to some degree (not sure what kind of object `items[0]` is, maybe convert them to dictionaries to simplify that conversion, `==` will only work for objects that implement `__eq__`)

tests/ui/test_search.py
```diff
@@ -53,8 +58,87 @@ def test_sorting_by(base_url, selenium, category, sort_attr):
...
+    time.sleep(2)
+    assert term in items[0].name
+
+
+@pytest.mark.nondestructive
+def test_blank_search_loads_results_page(base_url, selenium):
+    page = Home(selenium, base_url).open() 
+    results = page.search.search_for('', execute=True)
+    assert results.result_list.extensions[0].name == 'Ui-Addon'
+    assert results.result_list.themes[0].name == 'Ui-Test Theme'
```
EnTeQuAk: Given that a blank search should issue all add-ons (more or less, depending on platform filters etc) we should be able to expect a specific count of `results` here.

Also, in all other tests the result of `search_for` is called `items`, here it's called `results` (which, I personally like better).

tests/ui/test_search.py
```diff
@@ -53,8 +58,87 @@ def test_sorting_by(base_url, selenium, category, sort_attr):
...
+@pytest.mark.nondestructive
+def test_search_suggestion_term_loads_correct_suggestions(base_url, selenium):
+    page = Home(selenium, base_url).open()
+    term = 'Ui-Addon'
+    items = page.search.search_for(term, execute=False)
+    assert term == items[0].name
+    term = '-Install'
+    items = page.search.search_for(term, execute=False)
+    # Sleep to let autocomplete update.
+    time.sleep(2)
```
EnTeQuAk: I wonder if `0.8` or so would suffice? Autocomplete should be fairly fast.

tests/ui/test_search.py
```diff
@@ -53,8 +58,87 @@ def test_sorting_by(base_url, selenium, category, sort_attr):
...
+
+@pytest.mark.nondestructive
+def test_capitalization_has_same_suggestions(base_url, selenium):
+    page = Home(selenium, base_url).open()
+    term = 'Ui-Addon-Install'
+    items = page.search.search_for(term, execute=False)
+    assert term == items[0].name
+    items = page.search.search_for(term.capitalize(), execute=False)
+    # Sleep to let autocomplete update.
+    time.sleep(2)
```
EnTeQuAk: I wonder if `0.8` or so would suffice? Autocomplete should be fairly fast.

tests/ui/test_search.py
jrbenny35: ```items``` is a list of results from the autocomplete box.

tests/ui/test_search.py
jrbenny35: Right, so when ```search_for``` has ```execute``` as True, it will load the search results page. I don't quite know of a way to query for how many add-ons there are at one time, so I just figured I'd test to make sure themes and add-ons loaded.

tests/ui/test_search.py
jrbenny35: Worth a shot!

tests/ui/test_search.py
```diff
@@ -79,43 +79,49 @@ def test_search_suggestion_term_is_higher(base_url, selenium):
...
-    special_chars = '$%^&*'
-    items = page.search.search_for(term, execute=False)
-    assert term == items[0].name
-    items = page.search.search_for(special_chars, execute=False)
-    assert term in items[0].name
+    special_chars = u'%ç√®å'
+    suggestions = page.search.search_for(term, execute=False)
+    assert term == suggestions[0].name
+    suggestions = page.search.search_for(
+        special_chars.encode('utf-8').decode('utf-8'), execute=False)
```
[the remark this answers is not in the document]
jrbenny35: Hmm, I am not good with unicode and this seemed to work.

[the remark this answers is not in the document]
EnTeQuAk: Do you get any errors with `.search_for(special_chars, execute=False)` without the de/encoding?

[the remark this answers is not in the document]
jrbenny35: Yeah because we have ```# -*- coding: utf-8 -*-``` set and I get UnicodeError

EnTeQuAk:
Can you paste that traceback maybe?


On Fri, Dec 14, 2018, at 5:57 PM, Benjamin Forehand Jr wrote:
> *@jrbenny35* commented on this pull request.


> 
> In tests/ui/test_search.py[1]:


> >  @pytest.mark.nondestructive def
> >  test_special_chars_dont_break_suggestions(base_url, selenium): page
> >  = Home(selenium, base_url).open() term = 'Ui-Addon-Install' -
> >  special_chars = '$%^&*' -    items = page.search.search_for(term,
> >  execute=False) -    assert term == items[0].name -    items =
> >  page.search.search_for(special_chars, execute=False) -    assert
> >  term in items[0].name +    special_chars = u'%ç√®å' +
> >  suggestions = page.search.search_for(term, execute=False) +
> >  assert term == suggestions[0].name +    suggestions =
> >  page.search.search_for( +        special_chars.encode('utf-8').decode('utf-
> >  8'), execute=False)
>> Yeah because we have # -*- coding: utf-8 -*- set and I get
> UnicodeError> — You are receiving this because your review was requested. Reply to
> this email directly, view it on GitHub[2], or mute the thread[3].
Links:

  1. https://github.com/mozilla/addons-server/pull/10206#discussion_r241822128
  2. https://github.com/mozilla/addons-server/pull/10206#discussion_r241822128

jrbenny35:
I think I got it. Idk what I was doing wrong.

tests/ui/test_search.py
EnTeQuAk: Wanna try this out still?

jrbenny35:
I tried the ```0.8``` sleeps but it failed. Plus I think for CI, 0.8 would be to short as well.

PR #10217, opened 2018-12-14
remove collection.top_tags

### LINKED ISSUE #10048: Remove collection.top_tags
Top tags aren't used by the new frontend so we should remove all that functionality.

I'm specifically talking about `bandwagon:tasks::collection_meta` and `bandwagon:tags::collection_detail` which is only used for the legacy site which isn't used anymore and `top_tags` being empty is being handled properly so we can simply remove it.

dbxnr:
Fixes #10048 

Removes references to collection.top_tags

### DIFF
--- a/src/olympia/bandwagon/models.py
+++ b/src/olympia/bandwagon/models.py
@@ -117,8 +117,6 @@ class Collection(ModelBase):
 
     objects = CollectionManager()
 
-    top_tags = TopTags()
-
     class Meta(ModelBase.Meta):
         db_table = 'collections'
         unique_together = (('author', 'slug'),)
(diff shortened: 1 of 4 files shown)


EnTeQuAk:
There's a few linting issues you'll need to a ddress: https://travis-ci.org/mozilla/addons-server/jobs/467808003#L1081-L1083

src/olympia/bandwagon/models.py
```diff
@@ -117,8 +117,6 @@ class Collection(ModelBase):
 
     objects = CollectionManager()
 
-    top_tags = TopTags()
```
EnTeQuAk: The `TopTags` at the beginning of the file is now unused and can be removed as well.

PR #10230, opened 2018-12-17
Decode guid using base64 if prefixed by rta: in search API

diox:
This supports the "Return to AMO" feature from [bug 1468680](https://bugzilla.mozilla.org/show_bug.cgi?id=1468680).

Fix mozilla/addons#6258

### DIFF
--- a/src/olympia/search/filters.py
+++ b/src/olympia/search/filters.py
@@ -1,4 +1,6 @@
+from django.conf import settings
 from django.utils import translation
+from django.utils.http import urlsafe_base64_decode
 from django.utils.translation import ugettext
 
 import colorgram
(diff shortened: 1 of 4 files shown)


src/olympia/search/filters.py
```diff
@@ -135,7 +137,23 @@ class AddonGuidQueryParam(AddonQueryParam):
...
+        # only in the search API): if the guid param matches this format, and
+        # the feature is enabled through a setting, then we decode it and
+        # proceed. A 400 is returned if the format is recognized but the
+        # feature is disabled through the setting, acting as a kill-switch.
+        if value.startswith('rta:') and '@' not in value:
+            if settings.RETURN_TO_AMO is not True:
+                raise ValueError('RTA is currently disabled')
+            # Any ValueError will trigger a 400.
+            try:
+                value = urlsafe_base64_decode(value[4:])
```
eviljeff: You could technically encoded a list of guids, comma-separated, because we're not just returning the value at this point.  Maybe we want this (though I'm not sure why) and maybe it's a bug report waiting to happen when a guid containing a comma that isn't appropriately encoded is sent by Firefox.  
In either case, the docs above say " a single guid can be passed in base64url format," which isn't entirely correct.

PR #10255, opened 2018-12-20
Allow .tgz again for source uploads

### LINKED ISSUE #10254: Allow tgz files for source code bundles
- Try to upload a tar that is gz compressed with the file extension "tgz" as source code bundle for an extension version

### What happened?
Only "tar.gz" (among other formats) is accepted as file extension for gz compressed tars.

Since "tgz" is a shorter form of "tar.gz" it would also be accepted. This is especially useful since `npm pack` produces a "tgz" archive of extension source code without any node_modules or build artifacts if ignores are configured correctly.

See https://docs.npmjs.com/cli/pack for the documentation of `npm pack`.

diox:
Fix #10254

### DIFF
--- a/src/olympia/devhub/tests/test_views_submit.py
+++ b/src/olympia/devhub/tests/test_views_submit.py
@@ -675,6 +675,18 @@ class TestAddonSubmitSource(TestSubmitBase):
             oct(os.stat(self.get_version().source.path)[stat.ST_MODE]))
         assert mode == '0100644'
 
+    def test_submit_source_tgz(self):
+        response = self.post(
+            has_source=True, source=self.generate_source_compressed_tar(
+                suffix='.tgz'))
+        self.assert3xx(response, self.next_url)
+        self.addon = self.addon.reload()
+        assert self.get_version().source
+        assert self.addon.needs_admin_code_review
+        mode = (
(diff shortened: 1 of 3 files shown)


src/olympia/devhub/tests/test_views_submit.py
```diff
@@ -675,6 +675,18 @@ def test_submit_source_targz(self):
             oct(os.stat(self.get_version().source.path)[stat.ST_MODE]))
         assert mode == '0100644'
 
+    def test_submit_source_tgz(self):
+        response = self.post(
+            has_source=True, source=self.generate_source_compressed_tar(
```
EnTeQuAk: I think `generate_source_compressed_tar` only passes in `w:bz2` or `w:gz` and this should require a regular, uncompressed `w` mode.

diox: No ? `.tgz` *should* be compressed with `w:gz`, it's a `.tar.gz` with a shorter suffix, that's all.

EnTeQuAk: eh, apologies, I accidentally read `.tar` and not `.tgz` for some reason, all good!

PR #10260, opened 2018-12-21
stop exceptions in the except losing all previous ES updates and prev…

### LINKED ISSUE #10030: Some themes appear to be "duplicated" in search results after the migration to static themes
STR:

1. Open AMO -dev
2. Search for "custom"
3. Observe the first 2 search results

Actual result:
By comparing the two results, it seams that they refer to the same theme. One has a preview while the other doesn't. The details page for either of the themes shows no preview.  On the user profile page, there are also two entries for this theme. [1]

Expected result:
There is only one entry for this theme returned in search results

Notes:
- there is another strange behavior with this example: if you try to search for something else after having searched for "custom", the search results will still show one theme - the one without a preview - at the top of the list [2]
- another example that shows the same behavior is https://addons-dev.allizom.org/en-US/firefox/addon/turntablism/

@eviljeff this could actually be an anomaly created during the migration which also caused https://github.com/mozilla/addons-server/issues/9724

[1]

[2]
image

eviljeff:
should fix #10030 ...?

### DIFF
--- a/src/olympia/addons/tasks.py
+++ b/src/olympia/addons/tasks.py
@@ -710,9 +710,9 @@ def migrate_lwts_to_static_themes(ids, **kw):
 
     # Incoming ids should already by type=persona only
     lwts = Addon.objects.filter(id__in=ids)
-    pause_all_tasks()
     for lwt in lwts:
         static = None
+        pause_all_tasks()
         try:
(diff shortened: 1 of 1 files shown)


eviljeff:
@muffinresearch I'd like to include this in the 2019-01-07 OOB push so it's there for the LWT migration.

PR #10262, opened 2018-12-21
stop using username as a fallback when display name isn't set

### LINKED ISSUE #10261: stop using username as a fallback when display_name isn't set
part of #8302

eviljeff:
fixes #10261 - part of #8302


EnTeQuAk:
Restarted circleci tests for now, let's see

PR #10270, opened 2018-12-22
display_name limited to between 2 and 50 characters

### LINKED ISSUE #10265: restrict display_name to between 2 and 50 charactors
part of #8302

eviljeff:
fixes #10265

### DIFF
--- a/src/olympia/accounts/tests/test_views.py
+++ b/src/olympia/accounts/tests/test_views.py
@@ -1120,6 +1120,21 @@ class TestAccountViewSetUpdate(TestCase):
             'username': [u'Enter a valid username consisting of letters, '
                          u'numbers, underscores or hyphens.']}
 
+    def test_display_name_valid(self):
+        self.client.login_api(self.user)
+        response = self.patch(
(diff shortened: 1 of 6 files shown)


src/olympia/accounts/tests/test_views.py
```diff
@@ -1120,6 +1120,21 @@ def test_username_valid(self):
             'username': [u'Enter a valid username consisting of letters, '
                          u'numbers, underscores or hyphens.']}
 
+    def test_display_name_valid(self):
```
EnTeQuAk: `_valid` is probably a misleading postfix since it actually tests the validation (I'd call it `_validation` or so). Maybe add a valid-case for sake of completeness, I know there is one in `test_a_new_user_is_created_and_logged_in` but testing it explicitly to show that "things are still working" while being in context might be a good idea.

eviljeff: https://github.com/mozilla/addons-server/pull/10270/commits/8713834749f38ef0685abf6d92abdfd11a85bb91

PR #10321, opened 2019-01-07
Force pygit2 global search path to be set to HOME environment variable.

### LINKED ISSUE #10320: GitError: failed to stat '/root/.gitconfig'
https://sentry.prod.mozaws.net/operations/olympia-dev/issues/5099287/

```
GitError: failed to stat '/root/.gitconfig'
  File "django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "newrelic/hooks/framework_django.py", line 544, in wrapper
    return wrapped(*args, **kwargs)
  File "olympia/amo/decorators.py", line 31, in wrapper
    return func(request, *args, **kw)
  File "olympia/devhub/views.py", line 1380, in submit_addon_upload
    request, None, channel_id, 'devhub.submit.source')
  File "django/utils/decorators.py", line 185, in inner
    return func(*args, **kwargs)
  File "olympia/devhub/views.py", line 1327, in _submit_upload
    user=request.user)
  File "olympia/addons/models.py", line 626, in from_upload
    channel=channel, parsed_data=parsed_data)
  File "olympia/versions/models.py", line 250, in from_upload
    version=version, author=upload.user)
  File "olympia/lib/git.py", line 165, in extract_and_commit_from_version
    branch = repo.find_or_create_branch(BRANCHES[version.channel])
  File "olympia/lib/git.py", line 227, in find_or_create_branch
    branch = self.git_repository.branches.get(name)
  File "django/utils/functional.py", line 35, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "olympia/lib/git.py", line 80, in git_repository
    bare=False)
  File "pygit2/__init__.py", line 149, in init_repository
    check_error(err)
  File "pygit2/errors.py", line 64, in check_error
    raise GitError(message)
```

I'm a bit stumped that this happens but it seems for some reason pygit/libgit checks this file and can't, although I think it should be able to inside the docker container…

This looks like some kind of user mismatch

EnTeQuAk:
It's a bit problematic since there isn't a real good default value since
it depends on how the app is hosted, if `HOME` is set correctly and many
other circumstances. Let's enforce it to be the users home to make sure
we are accessing a directory that we have the permissions to.

Fixes #10320

### DIFF
--- a/src/olympia/lib/tests/test_git.py
+++ b/src/olympia/lib/tests/test_git.py
@@ -3,6 +3,7 @@ import subprocess
 import zipfile
 
 import pytest
+import pygit2
 import mock
 
 from django.conf import settings
@@ -36,6 +37,22 @@ def test_temporary_worktree():
     assert not os.path.exists(worktree.temp_directory)
     output = subprocess.check_output('git worktree list', shell=True, env=env)
     assert worktree.name not in output
+
+
+def test_enforce_pygit_global_search_path_to_home():
+    # https://github.com/mozilla/addons-server/issues/10320
+    pygit2.settings.search_path[pygit2.GIT_CONFIG_LEVEL_GLOBAL] = '/root'
+
+    assert (
+        pygit2.settings.search_path[pygit2.GIT_CONFIG_LEVEL_GLOBAL] ==
+        '/root')
+
+    os.environ['HOME'] = settings.GIT_FILE_STORAGE_PATH
+    repo = AddonGitRepository(1)
+
+    assert (
+        pygit2.settings.search_path[pygit2.GIT_CONFIG_LEVEL_GLOBAL] ==
+        settings.GIT_FILE_STORAGE_PATH)
 
 
 def test_git_repo_init():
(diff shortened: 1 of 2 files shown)


src/olympia/lib/tests/test_git.py
```diff
@@ -38,6 +39,24 @@ def test_temporary_worktree():
     assert worktree.name not in output
 
 
+def test_enforce_pygit_global_search_path_to_home():
+    # https://github.com/mozilla/addons-server/issues/10320
+    pygit2.settings.search_path[pygit2.GIT_CONFIG_LEVEL_GLOBAL] = '/root'
```
diox: Shouldn't you revert this to the original value ? And also make sure that it is indeed reverted, regardless of whether the test succeeds or not ?

EnTeQuAk: indeed, I'll fix that.

EnTeQuAk: Fixed in 9b2d5a5

PR #10324, opened 2019-01-07
Force display_name to have at least one printable charactor

### LINKED ISSUE #10266: don't allow a display_name purely of non-printable|displayable charactors
> Other than not allowing display names that are just spaces, current character restrictions to the display name should remain the same.

part of #8302 - bonus marks if it can fix #3388 too

We might want to do something like @diox said in #8302:
> I think for this, if we want to go further, my suggestion in #10097 (comment) should work - Using unicodedata.category + the Unicode Character Database properties values we can figure out whether each character is meant to be a letter in any alphabet, or punctation, space character of any kind, symbol, etc

eviljeff:
fixes #10266

### DIFF
--- a/src/olympia/accounts/tests/test_views.py
+++ b/src/olympia/accounts/tests/test_views.py
@@ -1136,6 +1136,13 @@ class TestAccountViewSetUpdate(TestCase):
                 'Ensure this field has no more than 50 characters.']}
 
         response = self.patch(
+            data={'display_name': u'\x7F\u20DF'})
+        assert response.status_code == 400
+        assert json.loads(response.content) == {
+            'display_name': [
+                'Must contain at least one printable character.']}
+
+        response = self.patch(
             data={'display_name': 'a' * 50})
         assert response.status_code == 200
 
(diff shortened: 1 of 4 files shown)


src/olympia/api/validators.py
```diff
@@ -0,0 +1,17 @@
+import unicodedata
+
+from django.utils.translation import ugettext
+
+from rest_framework import serializers
+
+
+class OneOrMorePrintableCharacterValidator:
+    def __init__(self, unicode_categories=None):
+        self.unicode_categories = unicode_categories or ('L', 'N', 'P', 'S')
```
diox: Can you add a comment linking to http://www.unicode.org/reports/tr44/tr44-6.html#Property_Values and explaining which categories those are, for future reference ?

eviljeff: https://github.com/mozilla/addons-server/pull/10324/commits/fbb73165a83cf05a76479bd5373bed318d79caf7

src/olympia/accounts/tests/test_views.py
```diff
@@ -1135,6 +1135,13 @@ def test_display_name_validation(self):
             'display_name': [
                 'Ensure this field has no more than 50 characters.']}
 
+        response = self.patch(
+            data={'display_name': u'\x7F\u20DF'})
+        assert response.status_code == 400
+        assert json.loads(response.content) == {
+            'display_name': [
+                'Must contain at least one printable character.']}
```
EnTeQuAk: Let's add just one more test that simply adds one printable character and then succeeds to show that it's indeed working as expected with mixed strings.

PR #10337, opened 2019-01-08
Changes needed for tests to run under Python 3 (staying compatible with Python 2)

diox:
(Hopefully) All the changes we need to get tests to actually run (not pass...) under Python 3.

WIP.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -7,8 +7,8 @@ on module-level, they should instead be added to hooks or fixtures directly.
 import os
 import uuid
 
-import responses
 import pytest
+import responses
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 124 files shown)


diox:
We have our first tests passing under Python 3 with that PR \o/ And a bunch of failures as well, but there should be plenty of easy pickings.

EnTeQuAk:
Execnet seems to fail hilariously in https://travis-ci.org/mozilla/addons-server/jobs/476928430 - but that's worth an entirely different issue, might even be fun to debug.

src/olympia/addons/forms.py
```diff
@@ -608,7 +608,7 @@ def save(self):
             'display_username': self.request.user.name
         }
         changed = False
-        for k, v in persona_data.iteritems():
+        for k, v in six.iteritems(persona_data):
```
EnTeQuAk: Why not simply do `persona_data.items()` if we don't necessarily care about performance implications here?

PR #10340, opened 2019-01-08
Use supervisor from master rather than the release on pypi to get py3 suport

### LINKED ISSUE #10339: use the python3 compatible version of supervisord
> supervisor is only required by dev.txt, maybe we could get away with installing it from a known commit directly, since there are no prerelease packages yet?

> Do you mean a commit to a supervisor source-code that is python3 compatible?

> That's what I meant yes - their master has Python 3 support - it's still experimental and that's why there hasn't been a release yet, but maybe it'd be enough for our needs.

eviljeff:
fixes #10339


requirements/dev_without_hash.txt
```diff
@@ -0,0 +1,3 @@
+
```
EnTeQuAk: That empty line probably isn't needed

PR #10347, opened 2019-01-09
add six.moves shim for urllib2 and HTMLParser

### LINKED ISSUE #10346: urllib2 doesn't exist in python3
https://travis-ci.org/mozilla/addons-server/jobs/477279508#L2393 et al.

eviljeff:
fixes #10346

### DIFF
--- a/src/olympia/devhub/tasks.py
+++ b/src/olympia/devhub/tasks.py
@@ -6,12 +6,12 @@ import os
 import socket
 import subprocess
 import tempfile
-import urllib2
 import shutil
 
 from copy import deepcopy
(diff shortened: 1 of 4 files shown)


src/olympia/devhub/tasks.py
```diff
@@ -949,11 +949,11 @@ def failed_validation(*messages):
 
 def _fetch_content(url):
```
diox: Is this used anywhere ?

eviljeff: Doesn't appear so.  I'll remove it.

PR #10352, opened 2019-01-09
don't add re.L (re.LOCALE) flag for regexs in python3

### LINKED ISSUE #10349: don't add re.L flag to regex in python3 - it's invalid for text_type strings
see https://docs.python.org/3/library/re.html#re.LOCALE - using it on a standard unicode string causes an error in 3.6+

eviljeff:
fixes #10349 
but it's kinda difficult to tell as it just exposes other test fails.
https://travis-ci.org/eviljeff/olympia/jobs/477334678 had the fails mentioning re.LOCALE and now https://travis-ci.org/eviljeff/olympia/jobs/477365985 doesn't.

### DIFF
--- a/src/olympia/ratings/forms.py
+++ b/src/olympia/ratings/forms.py
@@ -1,4 +1,5 @@
 import re
+import six
 
 from six.moves.urllib.parse import unquote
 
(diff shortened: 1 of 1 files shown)


src/olympia/ratings/forms.py
```diff
@@ -42,7 +43,8 @@ class RatingForm(RatingReplyForm):
     rating = forms.ChoiceField(
         zip(range(1, 6), range(1, 6)), label=_(u'Rating')
     )
-    flags = re.I | re.L | re.U | re.M
+    # re.L flag has been removed in py3.6 for text_type strings.
```
diox: Maybe make the comment even more explicit ? The docs say `Unicode matching is already enabled by default in Python 3 for Unicode (str) patterns, and it is able to handle different locales/languages` ; so it's not so much that it's been removed for text type strings, it's that it's enabled by default anyway. I was concerned before reading this that we would be removing something without providing a solution.

PR #10362, opened 2019-01-10
Handle empty recommendations result in get_disco_recommendations()

### LINKED ISSUE #10174: TAARDIS: Add curation override mechanism to discovery endpoint
[PRD: with details of the override feature is here](https://docs.google.com/document/d/1m_zmiVh1Oq8sz4NPjEQJz5k3MJZSVdpfr6Ar2IvyQpI/edit?ts=5bd1d4bd#heading=h.jvjx2h18ebr7
).  

See [the documentation for what TAAR is expecting here](https://github.com/mozilla/taar/blob/e4efb742f2117a73cddcf9c3ad70eaa27c6a9a30/API.md#get-addon-recommendations)

It's likely this can be an addition to the existing Disco pane curation tool to provide the hint as to what data needs to be sent to TAAR to configure overrides.

The rationale behind sending data to TAAR when we want to request recommendations is that this is easier than requiring TAAR to reach into AMO's data to find this out.

*Note: since this will receive a fair amount of traffic we'll need to optimize this so that the data lookup doesn't weigh down the `Disco -> AMO -> TAAR -> AMO -> Disco` round-trip response time.*

diox:
Fixes #10174

### DIFF
--- a/src/olympia/discovery/tests/test_utils.py
+++ b/src/olympia/discovery/tests/test_utils.py
@@ -128,6 +128,16 @@ def test_get_disco_recommendations(call_recommendation_server):
 
 
 @mock.patch('olympia.discovery.utils.call_recommendation_server')
+def test_get_disco_recommendations_empty(call_recommendation_server):
+    call_recommendation_server.return_value = None
+    recommendations = get_disco_recommendations('0', [])
+    assert recommendations == []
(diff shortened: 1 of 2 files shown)


diox:
I thought about doing that, but then you'd get an useless database query doing `.filter(guid__in=guids)` so I figured I'd go all the way...

PR #10370, opened 2019-01-11
Don't allow name or subtitle to be comprised of only symbols

### LINKED ISSUE #10097: Don't allow name or subtitle to be comprised of only symbols
Followup for #10093 

These fields shouldn't be exclusively comprised of symbols, so we need to detect there is at least one alphanumeric character on each. I'm using "alphanumeric" loosely here, since Chinese characters should be accepted, for example.

diox:
Fix #10097

### DIFF
--- a/src/olympia/amo/validators.py
+++ b/src/olympia/amo/validators.py
@@ -1,0 +1,35 @@
+import unicodedata
+
+from django.core.exceptions import ValidationError
+from django.utils.deconstruct import deconstructible
+from django.utils.translation import ugettext_lazy as _
+
+
+@deconstructible
+class OneOrMorePrintableCharacterValidator(object):
+    """Validate that the value contains at least one printable character."""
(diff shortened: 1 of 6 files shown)


src/olympia/amo/validators.py
```diff
@@ -0,0 +1,35 @@
...
+    def __call__(self, value):
+        for character in value:
+            if unicodedata.category(character)[0] in self.unicode_categories:
+                return
+        raise ValidationError(self.message)
+
+
+@deconstructible
+class OneOrMoreLetterOrNumberCharacterValidator(
+        OneOrMorePrintableCharacterValidator):
```
EnTeQuAk: Say that 3 times in a row, fast :rofl:

diox:
Requesting a second look from @eviljeff since he just implemented something similar for user names, want to make sure we're on the same page.

Issue #8779, opened 2018-07-06
Automate product-details sync as part of l10n job


EnTeQuAk:
In https://github.com/mozilla/addons-server/issues/8597 we are going to commit our product-details json files into the repository.

Let's automate their sync as part of our l10n job.

EnTeQuAk:
Not needed anymore with #10371 being merged.

Issue #2739, opened 2016-05-25
Add ka-GE, Georgian (Georgia) to AMO_LANGUAGES


EnTeQuAk:
refs https://github.com/mozilla/addons-server/issues/2479 but unfortunately I noticed that `ka-GE` isn't something we have in https://svn.mozilla.org/libs/product-details/json/languages.json which is used for the actual translation of the language.

Need to figure out how to get it there...

EnTeQuAk:
I unassigned myself, I don't have time looking into that anytime soon.

PR #10371, opened 2019-01-11
Remove django-product-details dependency.

### LINKED ISSUE #8850: Do not call update_product_details on deployment
Related to https://sentry.prod.mozaws.net/operations/olympia-prod/issues/367996/

```
IOError: [Errno 13] Permission denied: u'/data/olympia/src/olympia/../../src/olympia/lib/product_json/thunderbird.json'
(7 additional frame(s) were not displayed)
...
  File "product_details/management/commands/update_product_details.py", line 171, in download_json_file
    self._storage.update(json_file, resp.text, resp.headers.get('Last-Modified'))
  File "product_details/storage.py", line 233, in update
    shutil.move(tf.name, filename)
  File "python2.7/shutil.py", line 316, in move
    copy2(src, real_dst)
  File "python2.7/shutil.py", line 144, in copy2
    copyfile(src, dst)
  File "python2.7/shutil.py", line 97, in copyfile
    with open(dst, 'wb') as fdst:

IOError: [Errno 13] Permission denied: u'/data/olympia/src/olympia/../../src/olympia/lib/product_json/thunderbird.json'
```

Besides that this permission error should not happen we also should simply call `update_product_details` while building the docker image instead of during the deployment.

Advantage of this would be that docker-image building is part of our regular circleci-tests and something we get notified about in case of errors

### LINKED ISSUE #9516: Remove django-product-details dependency
Once https://github.com/mozilla/addons-server/issues/9515 lands there is almost no use of data from product-details.

Current occurrences are:

* Fetching language-related data, mostly their translations (in featured collections admin, stats
* Access to current and historic Firefox releases (in stats, via `product_details.firefox_history_major_releases`)
* Compatibility related tools, that will be removed with #9515 

We also get all language translations from product-details.

addons-frontend specifies all languages statically in https://github.com/mozilla/addons-frontend/blob/master/src/core/languages.js and we should be able to do the same in addons-server. We can have a management command to generate that file maybe, depends on the mood of the implementor.

EnTeQuAk:
* Add new olympia.core.languages module as a replacement for supported
  languages
* Implement new ApplicationsVersions.get_latest_application_version
helper

Fixes #9516 

Fixes #8850 along the way, removes the need for #8779 and unblocks #2739


src/olympia/versions/models.py
```diff
@@ -825,6 +826,20 @@ class Meta:
...
+                AppVersion.objects
+                .filter(
+                    ~models.Q(version__contains='*'),
+                    application=self.application)
+                .order_by('-version_int'))
+            return qs.first()
+
+        cache_key = 'appversion:{}:latest'.format(self.application)
+        return cache_get_or_set(
+            cache_key, _fetch_version, timeout=60 * 60 * 24)
```
EnTeQuAk: Actually… thinking about this after a weekend… I don't think it has to be cached since it's used only very rarely. I'll keep it for the review though for now, removing it is easy. Opinions about that?

diox: I think you can remove the caching right now. The query should be pretty fast already, there are not a lot of entries, there are indexes helping it already, and as you pointed out, it's going to be used so rarely it probably doesn't matter.

EnTeQuAk: Done in 6db8a494cd

PR #10372, opened 2019-01-11
Generate and use /static/js/i18n/{locale}.js instead of using a dynamic view

### LINKED ISSUE #9861: Explore the possibility of returning the file path of jsi18n.js as "/static/js/jsi18n-%locale%.js", or generating it as a physical file at deployment time
This relates to https://bugzilla.mozilla.org/show_bug.cgi?id=1501687.

We'd like to request for the -dev team to explore the possibility of either returning the file path for jsi18.js as `/static/js/jsi18n-%locale%.js` (as requested in that ticket), or, generating physical versions of this file and put it in `static/js/` directory at deployment time.

This helps avoid having to configure Nginx for special cases, as well as making accessing this file consistent as other script files (from Nginx point of view).

### What happened?
n/a

Just to see if you can do this or not. If it's not feasible to do it in the app, we can fall back to do it at the Nginx level.

diox:
Fix #9861


.gitignore
```diff
@@ -44,3 +44,4 @@ docker*.yml
 .cache
 .vscode
 src/olympia/discovery/strings.jinja2
+static/js/i18n/*.js
```
eviljeff: nit, but move it next to the two other `static/` gitignore lines above? (not that we have any order or logic to the file, I know)

src/olympia/amo/management/commands/generate_jsi18n_files.py
```diff
@@ -0,0 +1,22 @@
...
+from django.views.i18n import javascript_catalog
+
+
+class Command(BaseCommand):
+    help = 'Generate static jsi18n files for each locale we support'
+
+    def handle(self, *args, **options):
+        fake_request = HttpRequest()
+        for lang in settings.AMO_LANGUAGES:
+            filename = os.path.realpath(os.path.join(
```
eviljeff: you need `.realpath` here?

src/olympia/urls.py
```diff
@@ -75,6 +75,8 @@
 
     # Javascript translations.
     # Should always be called with a cache-busting querystring.
+    # Still served for the moment, but superseded by a fully static version we
+    # generate at deploy-time through our generate_jsi18n_files command.
```
eviljeff: mmm, what's the logic behind not just deleting it in one go?  The templates have been changed (above) to point to the new location so what would be using it the old dynamic url?

diox: I was worried about old pages cached in proxies/browser caches mostly. Not a huge deal admittedly, but I'm not sure what's out there...

src/olympia/amo/management/commands/generate_jsi18n_files.py
diox: I'll remove it. My original intention was to add a check to see where the path would end up, in case a rogue language string was added to `AMO_LANGUAGES`, but I didn't follow up with that.

.gitignore
diox: edb7d63e7b0679c5128f516a04c5ed8824ac7d95

src/olympia/amo/management/commands/generate_jsi18n_files.py
diox: f59a2a1015cca3bae6f601b4775ba5c486ddaade

src/olympia/templates/impala/base.html
```diff
@@ -205,7 +205,7 @@ <h3>{{ _('On the go?') }}</h3>
     {% endblock %}
     {# js #}
     {% block site_js %}
-      <script src="{{ CDN_HOST }}{{ cache_buster(url('jsi18n')) }}"></script>
+      <script src="{{ static('js/i18n/%s.js'|format(LANG)) }}"></script>
```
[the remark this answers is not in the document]
diox: Mmm didn't think of that. Yes, we should. Then the developer can update the files if need be.

PR #10400, opened 2019-01-15
have command to only regenerate missing theme previews

### LINKED ISSUE #10391: enhance recreate_theme_previews to only backfill missing previews
`process_addons --task=recreate_theme_previews` either requires specific ids (`--ids=...`) or us to want all the previews to be regenerated.  In the interest of efficiency it would be good to have a similar command that only generates previews for the themes that don't have any currently (or don't have all 3? not sure)

eviljeff:
fixes #10391 (and a bug with #10383)

### DIFF
--- a/src/olympia/addons/management/commands/process_addons.py
+++ b/src/olympia/addons/management/commands/process_addons.py
@@ -7,7 +7,9 @@ from celery import chord, group
 from olympia import amo
 from olympia.addons.models import Addon
 from olympia.addons.tasks import (
-    add_dynamic_theme_tag, add_firefox57_tag, bump_appver_for_legacy_addons,
+    add_dynamic_theme_tag,
+    add_firefox57_tag,
+    bump_appver_for_legacy_addons,
     disable_legacy_files,
     extract_colors_from_static_themes,
     find_inconsistencies_between_es_and_db,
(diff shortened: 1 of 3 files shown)


src/olympia/addons/tasks.py
```diff
@@ -937,19 +938,23 @@ def extract_colors_from_static_themes(ids, **kw):
...
     for addon in addons:
-        log.info('Recreating previews for theme: %s' % addon.id)
         version = addon.current_version
         if not version:
             continue
         try:
+            if only_missing:
+                with_size = (VersionPreview.objects.filter(version=version)
+                             .exclude(sizes={}).count())
+                if with_size == 3:
```
diox:
```suggestion
                if with_size == `len(amo.THEME_PREVIEW_SIZES)`
```

eviljeff: yeah... I was concerned about `len(amo.THEME_PREVIEW_SIZES)` reducing from 3 (the original plan last year was to standardize on 1 size and drop the others).  But I guess it's better the task code changes and so tests will fail and hopefully trigger us to look into it more.

src/olympia/addons/management/commands/process_addons.py
```diff
@@ -75,9 +77,18 @@
     'recreate_theme_previews': {
         'method': recreate_theme_previews,
         'qs': [
-            Q(type=amo.ADDON_STATICTHEME, status_in=[
```
diox: :scream:

src/olympia/addons/tasks.py
```diff
@@ -937,19 +938,23 @@ def extract_colors_from_static_themes(ids, **kw):
...
     for addon in addons:
-        log.info('Recreating previews for theme: %s' % addon.id)
         version = addon.current_version
         if not version:
             continue
         try:
+            if only_missing:
+                with_size = (VersionPreview.objects.filter(version=version)
+                             .exclude(sizes={}).count())
+                if with_size == len(amo.THEME_PREVIEW_SIZES):
```
diox: :+1:

src/olympia/addons/management/commands/process_addons.py
eviljeff: I know 🙄

PR #10410, opened 2019-01-16
Replace pytest-responses with our own custom implementation.

### LINKED ISSUE #10399: Replace pytest-responses with custom code
Reading upstream comments about upgrading to pytest 3.6+ in https://github.com/getsentry/pytest-responses/issues/8 makes me think if we should just pull that code into addons-server, it's not that much and should be fairly easy to maintain.

EnTeQuAk:
Fixes #10399

Closes #10378
Closes #10317

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -39,6 +39,27 @@ def mock_elasticsearch():
 
 
 @pytest.fixture(autouse=True)
+def start_responses_mocking(request):
+    """Enable ``responses`` this enforcing us to explicitly mark tests
+    that require internet usage.
+    """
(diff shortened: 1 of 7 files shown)


EnTeQuAk:
cc @diox and @eviljeff - as a matter of fact, we **should not be using @responses.activate** since it messes with what urls are activated and our pytest responses reset fixtures fall over because of that. Removing it actually fixes some random failure occurrences that have been happening in the past afair

EnTeQuAk:
It looks like the circleci failures are already on master for quite some time

Issue #10413, opened 2019-01-16
Make python3 fixes to "users" tests


eviljeff:
https://travis-ci.org/mozilla/addons-server/jobs/480257962

AlexandraMoga:
Please add STR or the label "qa:not needed".
Thanks

PR #10435, opened 2019-01-17
various python3 related fixes

eviljeff:
part of #10413 - more needed.

### DIFF
--- a/src/olympia/amo/views.py
+++ b/src/olympia/amo/views.py
@@ -93,7 +93,7 @@ def handler404(request):
 
 @non_atomic_requests
 def handler500(request):
-    if request.is_legacy_api:
+    if getattr(request, 'is_legacy_api', False):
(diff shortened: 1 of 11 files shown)


src/olympia/amo/views.py
```diff
@@ -93,7 +93,7 @@ def handler404(request):
 
 @non_atomic_requests
 def handler500(request):
-    if request.is_legacy_api:
+    if getattr(request, 'is_legacy_api', False):
```
diox: Why is this necessary?

eviljeff: It probably isn't in the end - there's an exception thrown, seemingly before the `is_legacy_api` property is added to `request`, and without this the real exception is replaced with one about `is_legacy_api` being undefined..

diox: ok, it kinda makes sense anyway...

PR #10446, opened 2019-01-18
Remove explicit rss link to blog.mozilla.org feed.

### LINKED ISSUE #10439: The 'Latest News' link from DevHub homepage doesn't lead to the required page
STR:

1. Open the Developer Hub homepage while being logged out - https://addons.mozilla.org/en-US/developers/
2. Scroll down and click on the 'Latest News' link

Actual result:
You are prompted to Open or Save an `.xml` file

Expected result:
The link should redirect to the correct page, or if the page is no longer available the link  should be removed

Notes:
reproduced on all AMO servers with FF64, Win10x64

EnTeQuAk:
There are other places on AMO where we explicitly use a separate link to
make the user subscribe to the RSS feed so those should be fine.

Fixes #10439

We could do a separate link like in other places like

but for that we'll need a new icon :see_no_evil: 

Opinions? cc @jvillalobos

### DIFF
--- a/src/olympia/devhub/templates/devhub/includes/blog_posts.html
+++ b/src/olympia/devhub/templates/devhub/includes/blog_posts.html
@@ -1,9 +1,6 @@
 <div class="blog-posts item">
   <h3>
     {{ _('Developer News') }}
-    <a title="{{ _('Subscribe to this feed') }}" class="subscribe-feed"
-       href="http://blog.mozilla.com/addons/feed/">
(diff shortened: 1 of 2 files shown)


EnTeQuAk:
QA: This pull request also removes the rss feed subscription link from the "All Add-ons" page in devhub.

screenshot_2019-01-18 manage my submissions developer hub add-ons for firefox

Issue #4503, opened 2017-01-27
Make sure travis tests run on our production MySQL version


EnTeQuAk:
Currently our Travis tests run on MySQL 5.5, we are using 5.6 and aim to use 5.7 in the near future.

We should run these on Travis too.

muffinresearch:
See https://github.com/travis-ci/travis-ci/issues/5122#issuecomment-355403249

AlexandraMoga:
@EnTeQuAk is QA needed?

PR #10455, opened 2019-01-18
Use utf8mb4 for connections to MySQL in dev

autrilla:
The database itself has been migrated, now we just need olympia to use utf8mb4 to connect to it!


EnTeQuAk:
Still trying to reproduce this locally, running the failed tests locally doesn't seem to make them fail. Newly created tables are `CHARSET=utf8mb4` so I think this should be identical. :thinking:

EnTeQuAk:
With 27d66ee we're going to fix https://github.com/mozilla/addons-server/issues/4503 as well.

Issue #10438, opened 2019-01-17
Local devhub not accepting upload file


jrbenny35:
### Describe the problem and steps to reproduce it:

- Pull latest docker images
- Start and configure server, creating dummy data
- Navigate to http://olympia.test/developers and sign in
- Go through addon upload process

### What happened?

- When you get to the page to actually upload the addon, you can select a file but the ```Continue``` button stays greyed out.

### What did you expect to happen?

- ```Continue``` button to become enabled and clickable.

EnTeQuAk:
I seem to be able to reproduce this, I wonder if 3b2a5855e7cc0bf9f84c7786bcb1f1e71789e9f1 broke this somehow.

diox:
Hey sorry about that. Any chance that ui tests could spit out the JavaScript errors that come up ? This would have make it more obvious.

EnTeQuAk:
That'd be really ideal, @jrbenny35 any idea about that? I think we'd generally failures on JavaScript failures since they shouldn't occur. Especially with our well defined tests.

jrbenny35:
Where did this error exist? We use a good log reporter that is easy to add things too so as long as it can access them, it's possible.

EnTeQuAk:
The error was a JavaScript error raised by our scripts in the devhub. It couldn't access a function that was meant to be defined (but wasn't because another script 404'ed) and threw an exception.

AlexandraMoga:
Could you please mark this as qa:not needed?
Thanks

Issue #10423, opened 2019-01-16
Make python3 fixes for "amo-locales-and-signing" job


diox:
Python 3 fixes for https://travis-ci.org/mozilla/addons-server/jobs/480441084

diox:
In #10525 I've added this FIXME to `src/olympia/lib/crypto/tests/test_signing.py` that it'd be nice to implement as part of this issue:
```
# FIXME: need to add a test with a unicode filename inside the package
# itself, and then check the signature.
```

diox:
Note that the job is now called `amo-lib-locales-and-signing`.

eviljeff:
The only remaining fails in this job are in test_middleware.py - `test_source_with_wrong_unicode_get` and `test_trailing_slash_middleware`.  They're both caused by `safe_query_string` (`iri_to_uri` really) encoding special characters in the query string as two escaped characters rather than one in python2.  

I haven't yet conclusively determine which is "correct" but receiving a single character in the query string and redirecting it with two seems wrong.

If this isn't an easy fix then we could mark skipif and have all the jobs be passing under python3 (then look at the skips)

AlexandraMoga:
Please add a qa:not needed label if there is no QA required.
Thanks

PR #10457, opened 2019-01-18
Use new /sign/file autograph endpoint

### LINKED ISSUE #9308: sign addons with autograph /sign/file
Per https://github.com/mozilla/addons-server/pull/9306#issuecomment-417300984, we want to switch addons-server to the autograph [`/sign/file` endpoint](https://github.com/mozilla-services/autograph/tree/6bc03e3ce0ce8aae6e31b8240159ec2c7bceef86/signer/xpi#signature-request).

This will let us eventually migrate to SHA256 digests for the PKCS7 signature and add COSE Signatures for https://bugzilla.mozilla.org/show_bug.cgi?id=1471186

We'll probably want to:

- [ ] QA the autograph-signed addons (this probably entails fixing ~~https://github.com/mozilla-services/autograph/issues/130~~ (edit: not necessary; closed) and https://github.com/mozilla-services/autograph/issues/122 and making sure autograph XPIs pass other signing-clients and addons-server unit tests)  
- [ ] determine peak load and network capacity and load test autograph

EnTeQuAk:
Fixes #9308

Various cleanups along the way, this PR is porting some code over from `signing-clients` so that we'll be able to remove that dependency once we don't have to do the actual signing on our end anymore.

I'm also renaming `crypto/packaged.py` into `crypto/signing.py` since it feels more fitting.


src/olympia/lib/crypto/signing.py
```diff
@@ -0,0 +1,278 @@
...
+                'pkcs7_digest': 'SHA1',
+                'cose_algorithms': ['ES256']
+            },
+        }]
+
+        with statsd.timer('services.sign.addon.autograph'):
+            response = requests.post(
+                '{server}/sign/file'.format(server=conf['server_url']),
+                json=signing_request,
+                auth=HawkAuth(id=conf['user_id'], key=conf['key']))
```
EnTeQuAk: cc @g-k to just double-check the code above and that the docs are right.

g-k: :+1: lgtm

Also, not sure how much detail you want in include about signature validation in the docs, but:

* the relevant pref is "security.signed_app_signatures.policy"
* in Fx nightly (and stable I believe) it's currently set to 2 for COSEAndPKCS7WithSHA1OrSHA256 in https://searchfox.org/mozilla-central/rev/dac799c9f4e9f5f05c1071cba94f2522aa31f7eb/security/manager/ssl/tests/unit/test_signed_apps.js#190-198

EnTeQuAk:
This PR is ready for a first look but realizing that CircleCI tests are failing I think I'd like to fix the CircleCI tests via https://github.com/mozilla/addons-server/issues/10438 - or a separate PR if I realize that this is turning into too much work.

EnTeQuAk:
https://github.com/mozilla/addons-server/issues/10438 will be fixed via #10467 so this isn't part of this PR

src/olympia/lib/crypto/signing.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+def call_signing(file_obj):
+    """Get the jar signature and send it to the signing server to be signed."""
+    # Extract jar signature.
+    conf = settings.AUTOGRAPH_CONFIG
+    log.debug('Calling autograph service: {0}'.format(
+        conf['server_url']))
+
+    if waffle.sample_is_active('activate-autograph-file-signing'):
```
diox: Because of the waffle this `call_signing()` function is now super long and hard to follow. Can you break it down?

diox: Also, add a migration to add the default values for the waffle in the database ?

EnTeQuAk: Fixed in 4590507769, I hope that's a bit easier to read?

EnTeQuAk:
Actually, some documentation about this: https://github.com/mozilla-services/autograph/blob/master/signer/xpi/README.rst#signature-request

src/olympia/lib/crypto/signing.py
```diff
@@ -53,19 +55,79 @@ def get_id(addon):
...
-    """Get the jar signature and send it to the signing server to be signed."""
-    # Extract jar signature.
+def autograph_sign_file(file_obj):
+    """Sign `file_obj` via autographs /sign/file endpoint.
+
+    :returns: The certificates serial number.
+    """
+    conf = settings.AUTOGRAPH_CONFIG
+
+    with storage.open(file_obj.current_file_path) as fobj:
```
diox: You should open in `rb` mode I think since you want to encode in base64, which takes bytes.

diox: Then again this probably doesn't matter in Py2 and you've got more potential for breakage through other calls below that deal with bytes in various ways... So maybe don't bother for now, it will have to be fixed in https://github.com/mozilla/addons-server/issues/10423

EnTeQuAk: Perfect, yeah, happy to tackle that as part of #10423 then

PR #10461, opened 2019-01-18
Python 3 fixes for the 'addons' job

### LINKED ISSUE #10419: Make python3 fixes to "addons" job
Python 3 fixes for https://travis-ci.org/mozilla/addons-server/jobs/480441081

diox:
Fix #10419


src/olympia/addons/models.py
```diff
@@ -187,7 +186,8 @@ def featured(self, app, lang=None, type=None):
         Filter for all featured add-ons for an application in all locales.
         """
         ids = get_featured_ids(app, lang, type)
-        return manual_order(self.listed(app), ids, 'addons.id')
+        qs = manual_order(self.listed(app), ids, 'addons.id')
+        return qs
```
eviljeff: not that it matters, but this isn't a necessary change

src/olympia/addons/tests/test_admin.py
```diff
@@ -226,7 +226,7 @@ def test_can_list_with_addons_edit_permission(self):
         self.client.login(email=user.email)
         response = self.client.get(self.list_url, follow=True)
         assert response.status_code == 200
-        assert '/addon/bar-replacement/' in response.content
+        assert '/addon/bar-replacement/' in response.content.decode('utf-8')
```
eviljeff: I've been wondering if we should adopt a best practise?  
`assert '/addon/bar-replacement/' in response.content` to
`assert b'/addon/bar-replacement/' in response.content` or
`assert '/addon/bar-replacement/' in response.content.decode('utf-8')` or
`assert '/addon/bar-replacement/' in force_text(response.content)`

src/olympia/addons/tests/test_views.py
```diff
@@ -62,8 +63,9 @@ def check_cat_sidebar(url, addon):
     """Ensures that the sidebar shows the categories for the correct type."""
     for type_ in [amo.ADDON_EXTENSION, amo.ADDON_THEME, amo.ADDON_SEARCH]:
         addon.update(type=type_)
-        r = Client().get(url)
-        assert pq(r.content)('#side-nav').attr('data-addontype') == str(type_)
+        response = Client().get(url)
+        assert pq(response.content)('#side-nav').attr('data-addontype') == str(
```
eviljeff: change to `text_type`

src/olympia/addons/tests/test_admin.py
diox: Been wondering about that too. I think any of the last 2 are best, but I sometimes just used the second one out of lazyness.

eviljeff: yeah I've been inconsistently making different changes in different places too.  I guess it's fine.  (until it's not)

src/olympia/addons/models.py
diox: Yep, that was a leftover from some debugging I did, removed in 23dcd85a29a21c098ebdb6029fc9b20f86aeb2fb

src/olympia/addons/tests/test_views.py
diox: Fixed in 71d47d41e01d219aedf5bd6e2bf74c21ec370019 with a few more str/bytes fixes that will matter latter for the `es` job.

Issue #10428, opened 2019-01-16
Switch nginx config so that the new front-end is the default


bobsilverberg:
The current configuration is that all traffic is routed to the legacy Django site and only routes that are specifically configured in nginx are routed to the new front end.

With the user legacy pages no longer used, we would like to reverse this, so all traffic would default to the new front end, and only specifically overridden routes would go to the legacy site.

Some routes that would need to go to the legacy site are:
- /developers/
- /admin/
- /reviewers/

Are there any others that I have missed?

diox:
Plenty more :)

https://github.com/mozilla/addons-server/blob/6b06f38de6faf50e73857d70843afd8eb52c2f84/src/olympia/urls.py#L23 has a list of includes, everything included at the root needs to checked as well.

At the very least:
- /api/
- /files/
- /downloads/
- /robots.txt
- /contribute.json
- /services/
- /__version__
- /opensearch.xml

Also need to be careful about URLs that use the locale and/or app name in it.

eviljeff:
we could/should move robots.txt, contribute.json and opensearch.xml to addons-frontend. (?)

/developers, /admin and /reviewers could have locale at the front but, afaik, app name is only for frontend pages.

diox:
The legacy API does have both locale and app name. /opensearch.xml has it too, optionally (it's not in the list of urls exempted from it in addons-server). /files/ and /downloads/ are not exempted either (though /downloads/ doesn't need the locale)

willdurand:
> we could/should move robots.txt, contribute.json and opensearch.xml to addons-frontend. (?)

Quick notes:

- `robots.txt` is generated by addons-server–I don't know if we have enough knowledge to do that on the frontend
- `opensearch.xml` has been added to addons-frontend this week (part of next week's tag) and we use the URL with both the app and locale

muffinresearch:
> opensearch.xml has been added to addons-frontend this week (part of next week's tag) and we use the URL with both the app and locale

@willdurand Is the frontend serving opensearch.xml or just pointing to it?

muffinresearch:
> we could/should move robots.txt, contribute.json and opensearch.xml to addons-frontend. (?)
> 
> /developers, /admin and /reviewers could have locale at the front but, afaik, app name is only for frontend pages.

Moving anything should be out of scope for this issue, but we can definitely file issues to start thinking about it. I still hope we can continue to move in the direction that at some point addons-server with be just API.

willdurand:
[...]

PR #10468, opened 2019-01-21
Remove shield studies 1-16

### LINKED ISSUE #10448: Remove old shield studies
We have ~16 shield studies as static pages in our repository, are they still needed and used? I see in datadog ~1-5 requests to all our static pages and most certainly go to the about page.

cc @Osmose (feel free to redirect me to someone else)

EnTeQuAk:
Fixes #10448

This keeps the pioneer pages intact because I think they're still useful
since pioneer is the mechanism to remove the need to add separate pages
for every shield study.

This PR doesn't 404 on those pages but redirects to the relevant
support pages.


src/olympia/lib/settings_base.py
```diff
@@ -1050,6 +1050,8 @@ def get_db_config(environ_var, atomic_requests=True):
 VALIDATION_FAQ_URL = ('https://wiki.mozilla.org/Add-ons/Reviewers/Guide/'
                       'AddonReviews#Step_2:_Automatic_validation')
 
+SHIELD_STUDIES_SUPPORT_URL = 'https://support.mozilla.org/kb/shield'
```
EnTeQuAk: cc @mythmon - that's the page I'll redirect users to when hitting shield-study 2-16, the pioneer page is being kept intact.

EnTeQuAk:
We could certainly do that, cc @autrilla. They're hard 301s for now, we can remove them later once uplifted into nginx. On the other hand, the traffic on this is **very** low so I don't think we should care. Every new rule on nginx affects **every** request that's incoming so we shouldn't add rules like that for pages that most certainly won't be requested much more than a few times a month from old links on the internet.

EnTeQuAk:
Actually, let me revert my previous comment… I think it makes sense to uplift these changes into nginx with #10428 but this isn't necessary to do immediately so I'm going to merge this anyway as cleanup and we'll be able to move this later if we want to.

Issue #10472, opened 2019-01-21
response mocking in python3 isn't working locally (just basket)


eviljeff:
Requests that hit the basket api are supposed to be automatically caught and a mocked response returned.  The basket API is hit each time a user's (own) profile is accessed (for notifications) so a number of accounts and users tests rely on the responses containing `token` in the JSON block response.
In python2 everything works fine; in Travis on python3 everything works fine; locally with python3 the tests fail.

EnTeQuAk:
This sounds a lot like a bytes vs unicode error on some internal "check if this url matches ..." or something like that. Just a wild guess.

diox:
It's because our docker is running Python 3.5, which hasn't got this bit: https://bugs.python.org/issue10976

basket-client is passing the result of `response.content` from a `requests` call directly to `json.loads()`.

PR #10492, opened 2019-01-23
Use latest-py3 tag in docker-compose instead of a separate image name

diox:
Follow-up for #10472 - it turns out it's easier to just have docker hub build a separate tag of the same repo.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,7 +36,7 @@ services:
   web3:
     extends:
       service: web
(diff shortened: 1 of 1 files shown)


diox:
Note: the first build is still in progress as I'm submitting this PR. That's https://cloud.docker.com/u/addons/repository/registry-1.docker.io/addons/addons-server/hub-builddetail/bzwaku8skbgtupvq8lz7lom

PR #10494, opened 2019-01-23
Remove idna dependency, it's not used anymore.

EnTeQuAk:
With https://github.com/mozilla/addons-server/pull/10488 we don't have
to depend on idna being installed anymore.


EnTeQuAk:
Actually, silly me, this is still required by python-requests.

PR #10505, opened 2019-01-25
Update 1070-remove-name-column to drop fk first

eviljeff:


eviljeff:
Although... this succeeded on addons-dev already?

diox:
mmmm I'm not sure the constraint existed on dev/stage/prod... I don't see it on stage...

eviljeff:
it's there locally but it's not on -dev either.
🤷‍♂️

diox:
IMHO, we should add your change... commented out. If someones needs to run the migration locally they can uncomment it, and since dev/stage/prod don't have the FK it should just work out of the box in those envs. (Need to double-check if prod is the same as stage in that regard)

PR #10508, opened 2019-01-25
expose python version in __version__

### LINKED ISSUE #9396: Add a file with the current deployed python version
From: https://github.com/mozilla/addons-server/pull/9390#issuecomment-419951143

> I thought it redundant and could be confused with prod_common. But I could see an argument for having a deploy.txt, or something similarly named, if we are going to have to specify the filename somewhere (though ideally it'd all be done via the make recipes which has python version detection).

eviljeff:
fixes #9396

### DIFF
--- a/src/olympia/amo/tests/test_views.py
+++ b/src/olympia/amo/tests/test_views.py
@@ -488,6 +488,5 @@ class TestVersion(TestCase):
         assert res._headers['content-type'] == (
             'Content-Type', 'application/json')
         content = json.loads(force_text(res.content))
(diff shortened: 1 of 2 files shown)


diox:
Do we really want this, up to the micro version ? Through this is a bit like security through obscurity, generally servers tend to hide which version of various third-party software they are on. We should probably ask ops about it first.

eviljeff:
I can remove the micro version so it'd be just "2.7" and "3.6" or whatever.  For the purpose of the issue I don't think it matters.

PR #10510, opened 2019-01-25
Check add-on guids for return to AMO against disco content

### LINKED ISSUE #10504: Check safe-list for RTA (Return to AMO) guids
We'd like to check `rta` prefixed links for RTA (Return to AMO) against the list of add-ons that is used for the pool for TAARDIS (recommendations in the disco pane).

If the add-on is in the list then the response should be handed back, if it isn't the result should be a 400 error as it is for a bad guid.

diox:
Disco editorial content acts as a safe-list of curated content we know RTA is safe to use on. It's disabled for the rest, returning a 400 as with other failure cases in that feature.

Fix #10504


src/olympia/addons/tests/test_views.py
```diff
@@ -2255,8 +2256,10 @@ def test_get_queryset_excludes(self):
             key.endswith('.raw') for key in source_keys
         )
 
-    def perform_search(self, url, data=None, expected_status=200, **headers):
-        with self.assertNumQueries(0):
+    def perform_search(
+            self, url, data=None, expected_status=200, expected_queries=0,
+            **headers):
```
eviljeff: nit: this isn't the style we usually write overflowing function definitions in.  Elsewhere we'd write it as:
```python
def perform_search(self, url, data=None, expected_status=200,
                   expected_queries=0, **headers):
    with self.assertNumQueries(expected_queries):
````

File: src/olympia/addons/tests/test_views.py [hunk 2]
```diff
@@ -3052,8 +3070,10 @@ def tearDown(self):
         self.empty_index('default')
         self.refresh()
 
-    def perform_search(self, url, data=None, expected_status=200, **headers):
-        with self.assertNumQueries(0):
+    def perform_search(
```
eviljeff: same here I guess (and so on below)

src/olympia/search/filters.py
```diff
@@ -144,19 +145,29 @@ def get_value(self):
         # a specific API but rather encodes the guid and adds a prefix to it,
         # only in the search API): if the guid param matches this format, and
         # the feature is enabled through a setting, then we decode it and
-        # proceed. A 400 is returned if the format is recognized but the
-        # feature is disabled through the setting, acting as a kill-switch.
+        # check it against DiscoItems, which contains the list of add-ons
+        # susceptible to appear in disco pane, acting as a list of "safe"
```
eviljeff: > susceptible to appear in disco pane

"selected to appear in disco pane"?

diox: Technically no. It's the list of items that we have an editorial translation for, which is used by TAAR as the list of items that *can* appear, but it doesn't mean they necessarily will. I'm open to better wording suggestions though.

PR #10517, opened 2019-01-25
Use utf8mb4 for stage MySQL connections

autrilla:
The -stage database was just migrated to utf8mb4, so this can be enabled.


diox:
need to fix `src/olympia/conf/stage/settings.py:48:80: E501 line too long (92 > 79 characters)` though

autrilla:
Looks like a test is failing due to implicit ordering changing?

```
/home/travis/build/mozilla/addons-server/src/olympia/amo/utils.py:150: UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'olympia.addons.models.Addon'> AddonQuerySet.
```

PR #10525, opened 2019-01-28
Python 3 fixes for the 'main' job.

### LINKED ISSUE #10424: Make python3 fixes for "main" job
Python 3 fixes for https://travis-ci.org/mozilla/addons-server/jobs/480441086

diox:
Includes a change in the way jobs are split because there are still some broken things in
files/ and lib/

Fix #10424

### DIFF
--- a/src/olympia/bandwagon/views.py
+++ b/src/olympia/bandwagon/views.py
@@ -1,9 +1,7 @@
 import functools
-import hashlib
 
 from django import http
 from django.conf import settings
-from django.core.cache import cache
 from django.core.exceptions import PermissionDenied
 from django.db.models import Q
 from django.db.transaction import non_atomic_requests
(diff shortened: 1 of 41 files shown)


src/olympia/bandwagon/views.py
```diff
@@ -121,14 +119,7 @@ def collection_listing(request, base=None):
                   .filter(type=amo.COLLECTION_FEATURED)
                   .exclude(addon_count=0)
     )
-    # Counts are hard to cache automatically, and accuracy for this
```
diox: Removed for simplicity, it's only in a legacy view that we should no longer hit anyway.

src/olympia/files/tests/test_file_viewer.py
```diff
@@ -480,8 +484,11 @@ def test_unzip_fatal(self):
     def test_read(self):
         zip_file = SafeZip(self.xpi_path('langpack-localepicker'))
         assert zip_file.is_valid
-        assert 'locale browser de' in zip_file.read('chrome.manifest')
+        assert b'locale browser de' in zip_file.read('chrome.manifest')
 
+    @pytest.mark.skipif(
```
diox: Like the other `skipIf` I added this should be double-checked later. https://github.com/mozilla/addons-server/issues/10527

src/olympia/lib/crypto/tests/test_signing.py
```diff
@@ -174,6 +175,9 @@ def test_sign_file_non_ascii_filename(self):
         signing.sign_file(self.file_)
         self.assert_signed()
 
+    # FIXME: need to add a test with a unicode filename inside the package
+    # itself, and then check the signature.
+
```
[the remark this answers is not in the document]
diox: Yeah, IMHO it should be part of https://github.com/mozilla/addons-server/issues/10423 - I think we need this test for peace of mind when doing the Python 3 port.

diox:
Being able to build on it is why I left some fixes for files and crypto even though I ended up moving the tests to a separate job. It was 90% done so I figured might as well commit that early.

PR #10526, opened 2019-01-28
drop webext_permission_descriptions table

### LINKED ISSUE #10523: drop webext_permission_descriptions table from the database
follow-on from #10389 - #10389 dropped the `WebextPermissionDescription` model so we need to drop the database table too in an sql migration.

**Must land after the 2019.01.31 tag**

championshuttler:
Fixes #10523 

Remove webext_permission_descriptions table from DB.


EnTeQuAk:
This also needs a (tested!) comment similar to 3114c2f23c88500540f3645c89368175e3d601bd because we have to drop the appropriate translation foreign keys.

We're also going to have to drop the foreign key on our -dev, -stage and -prod environments before dropping the table.

According to 5c3cfd8385e78b64a65d119b4cb0af742c6c5b4e (and I checked on -dev at least) the constraint name is `webext_permission_descriptions_translation_id`

PR #10553, opened 2019-01-31
Remove amo-validator, remove support for uploading legacy-addons.

### LINKED ISSUE #10502: drop amo-validator
amo-validator is only used for legacy addons, which have long since been disallowed for general submission on AMO.  #10445 moved search plugin validator into addons-server directly which was the last remaining type of supported addon handled by amo-validator.  As python3 compatability work is highlighting problems with attempting to use a python2-only package under python3 we may as well just drop it.  The only use case for the amo-validator today is admin-submitted updates to old add-ons, and some system add-ons, both of which should be thoroughly checked for compatablity before upload anyway.

EnTeQuAk:
This is dropping most of our compatibility checks for legacy add-ons and
completely blocks them from being uploaded.

We workaround this block for mozilla-signed add-ons by issuing the same
error, just as a warning instead of an actual error.

This affected *loads* of our tests so the PR contains more test-changes
then I'd like.

In addition to that this PR adds file extension validation for uploads,
we only did that in our JavaScript and never in our code. So from now
on, we have a hard limit to only allow whatever is in
`VALID_EXTENSIONS`.

Further...

* drop amo-validator and related dependencies
* remove `dump_apps` management command
* removed unused fixture files from the repo

Fixes #10502

Future improvement idea for our test-suite: Unify add-on file fixtures
under one single folder to avoid any duplication.

There is most certainly some more improvements that could be made to
code and test structure but I went for the least changes, it's way too
huge already :-/


EnTeQuAk:
I still need to click through all uploads etc to test this via the UI, I'll post screenshots but please feel free to give this a first look through.

src/olympia/devhub/file_validation_annotations.py
```diff
@@ -30,135 +27,29 @@ def insert_validation_message(results, type_='error', message='', msg_id='',
...
-        msg = ugettext(
-            u'Legacy add-ons are not compatible with Firefox 57 or higher. '
-            u'Use a maxVersion of 56.* or lower.')
-
-        insert_validation_message(
-            results, message=msg, msg_id='legacy_addons_max_version')
+    description = ugettext(
+        u'Add-ons for Thunderbird and SeaMonkey are now listed and '
+        u'maintained on addons.thunderbird.net. You can use the same '
+        u'account to update your add-ons on the new site.')
```
eviljeff: Not sure I understand here - the message is that legacy add-ons aren't supported but the description is about Thunderbird and Seamonkey add-ons.  But I don't see any code (remaining) that checks if the add-on is acutally targeting Thunderbird/Seamonkey.

src/olympia/devhub/file_validation_annotations.py
```diff
@@ -238,7 +101,7 @@ def annotate_akismet_spam_check(results, akismet_results):
                 msg_id='akismet_is_spam_%s' % field)
 
 
-def annotate_search_plugin_validation(results, file_path, channel):
+def annotate_search_plugin_validation(results, file_path, is_listed):
```
eviljeff: Can we leave it as `channel` please?  It doesn't appear to be a necessary change.

Really, I was supposed to change all the functions to use `channel` rather than `is_listed` when I implemented version channels but when faced with the huge number of validation tests that specified `is_listed` I gave up.  So if anything, code that still uses `is_listed` should be changed to use `channel` rather than the other way around.

src/olympia/devhub/tests/test_views_submit.py
```diff
@@ -2107,29 +2107,6 @@ def test_success(self):
         log_items = ActivityLog.objects.for_addons(self.addon)
         assert log_items.filter(action=amo.LOG.ADD_VERSION.id)
 
-    @mock.patch('olympia.devhub.views.sign_file')
-    def test_experiments_are_auto_signed(self, mock_sign_file):
```
eviljeff: So the conclusion was that webextension experiments are completely gone, not just replaced with something in the manifest?

src/olympia/files/utils.py
```diff
@@ -81,6 +81,12 @@ class ParseError(forms.ValidationError):
 _lock_count = {}
 
 
+# Acceptable extensions.
+# This is being used by `parse_addon` so please make sure we don't have
+# to touch add-ons before removing anything from this list.
+VALID_EXTENSIONS = ('.crx', '.xpi', '.jar', '.xml', '.json', '.zip')
```
eviljeff: Should this be in base/constants.py?

src/olympia/devhub/file_validation_annotations.py
```diff
@@ -30,135 +27,29 @@ def insert_validation_message(results, type_='error', message='', msg_id='',
...
-            u'Add-ons for Thunderbird and SeaMonkey are now listed and '
-            u'maintained on addons.thunderbird.net. You can use the same '
-            u'account to update your add-ons on the new site.')
-
-        insert_validation_message(
-            results, message=msg, msg_id='thunderbird_and_seamonkey_migration')
-
-    # Legacy add-ons are no longer supported on AMO (if waffle is enabled).
-    elif (is_extension_or_complete_theme and
-            waffle.switch_is_active('disallow-legacy-submissions')):
```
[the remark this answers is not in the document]
EnTeQuAk: Not fully, I'll file follow-up issues.

src/olympia/devhub/file_validation_annotations.py
EnTeQuAk: Given that Thunderbird/Seamonkey don't support WebExtensions at all I thought we don't necessarily have to double-check them all.

Maybe it's worth having a test just for sake of completeness that tests thunderbird/seamonkey?

src/olympia/devhub/file_validation_annotations.py
EnTeQuAk: Gotcha, I'll do that!

src/olympia/devhub/file_validation_annotations.py
eviljeff: Probably.  It's just the way I read it was a developer uploads a (Firefox compatible) legacy add-on and we tell them about Thunderbird and Seamonkey, which would be confusing

src/olympia/devhub/tests/test_views_submit.py
EnTeQuAk: There is another test, `test_experiments_inside_webext_are_auto_signed` that already tests WebExtension style experiments (manifest.json based). So this test here got removed.

src/olympia/files/utils.py
EnTeQuAk: Now that you say that, I think it kinda makes sense. I'll move it there.

src/olympia/devhub/file_validation_annotations.py
EnTeQuAk: Indeed, yeah, I merged the messages because in the end it's all legacy add-ons and we don't support them. I'd even guess that most add-ons that run into that barrier may even be Thunderbird add-ons given that Firefox doesn't have support for legacy add-ons for quite some time.

src/olympia/devhub/file_validation_annotations.py
EnTeQuAk: Fixed in d86e43092c (hoping that this doesn't break too much)

src/olympia/files/utils.py
EnTeQuAk: Fied in 14a14bc500

src/olympia/devhub/tests/test_tasks.py
```diff
@@ -753,60 +673,44 @@ def test_no_upgrade_annotation_no_version(self):
...
         assert not upload.valid
 
-    def test_submit_legacy_extension_1st_version_in_that_channel_reverse(self):
-        file_ = get_addon_file('valid_firefox_addon.xpi')
-        addon = addon_factory(
-            version_kw={'version': '0.1',
-                        'channel': amo.RELEASE_CHANNEL_LISTED})
+    def test_submit_legacy_dictionary_disabled(self):
+        # Should not error: non-extensions types are not affected by the
+        # restriction, even if they target 57.
```
diox: mmm? but they should, and they do - the validation results do contain an error with the `legacy_addons_unsupported` message.

src/olympia/devhub/tests/test_views_validation.py
```diff
@@ -243,38 +239,52 @@ def test_context_and_content(self):
...
-        self.url = reverse('devhub.upload')
-        data = open(get_image_path('animated.png'), 'rb')
-        self.client.post(self.url, {'upload': data})
-        # Make sure it was called with listed=True.
-        assert validate_mock.call_args[1]['listed']
+        url = reverse('devhub.upload')
+
+        fpath = 'src/olympia/files/fixtures/files/webextension_no_id.xpi'
+
+        with open(fpath) as file_:
```
diox: I know there are lots of tests broken in the same way already, and I'm not suggesting to fix those in this PR, but here and below it should be:
```suggestion
        with open(fpath, 'rb') as file_:
```

The person who will end up porting these tests to Python 3 will thank you :)

eviljeff: I already have the devhub py3 branch in progress :)
Although I didn't touch `test_views_validation.py` yet because validator.

src/olympia/devhub/file_validation_annotations.py
eviljeff: I forgot to come back to this before I r+'d - I don't think we should be returning an error message mentioning Thunderbird or Seamonkey unless the developer has specified Thunderbird or Seamonkey in the install.rdf.  Can we leave that check in?  (iirc the app compatibility is in addons-server rather than the validator)

src/olympia/devhub/tests/test_tasks.py
EnTeQuAk: I think I simply failed to remove that comment. Fixed in 0c2bdf06d3, nice find!

src/olympia/devhub/file_validation_annotations.py
```diff
@@ -47,6 +48,21 @@ def annotate_legacy_addon_restrictions(results, parsed_data, error=True):
...
+
+    targets_thunderbird_or_seamonkey = False
+    thunderbird_or_seamonkey = {amo.THUNDERBIRD.guid, amo.SEAMONKEY.guid}
+
+    for ctx in extractor.rdf.objects(None, extractor.uri('targetApplication')):
+        if extractor.find('id', ctx) in thunderbird_or_seamonkey:
+            targets_thunderbird_or_seamonkey = True
+
+    description = description if targets_thunderbird_or_seamonkey else []
+
```
eviljeff: wfm 👍

Issue #10564, opened 2019-02-01
Loads of deadlocks happening durign submission


EnTeQuAk:
We are seeing lots of deadlocks during submission, `devhub.tasks.submit_file` to be precise.

EnTeQuAk:
Sentry issue: [OLYMPIA-PROD-2DK](https://sentry.prod.mozaws.net/operations/olympia-prod/issues/4833802/)

EnTeQuAk:
Sentry issue: [OLYMPIA-PROD-2DN](https://sentry.prod.mozaws.net/operations/olympia-prod/issues/4848460/)

EnTeQuAk:
Sentry issue: [OLYMPIA-PROD-27D](https://sentry.prod.mozaws.net/operations/olympia-prod/issues/4449820/)

EnTeQuAk:
Sentry issue: [OLYMPIA-PROD-2AB](https://sentry.prod.mozaws.net/operations/olympia-prod/issues/4586297/)

EnTeQuAk:
I'm actually going to close this and re-open if needed.

PR #10573, opened 2019-02-01
Remove our own atomic decorator, run create_version_for_upload in 'read committed'

EnTeQuAk:
This is an attempt to fix our enormous amounts of deadlocks. Now I
finally found a reasonable candidate.

`devhub.tasks:create_version_for_upload` is being wrapped in a custom
atomic decorator that sets the transaction level to 'SERIALIZABLE'. To
be fair, I don't think it needs that.

We switched to 'read committed' some time ago and I think it makes sense
to unify this here too.

Refs #10564

### DIFF
--- a/src/olympia/amo/decorators.py
+++ b/src/olympia/amo/decorators.py
@@ -5,7 +5,6 @@ import json
 from django import http
 from django.conf import settings
 from django.core.exceptions import PermissionDenied
-from django.db import connection, transaction
 
 import olympia.core.logger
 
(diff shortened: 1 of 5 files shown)


EnTeQuAk:
Reading what strong guarantees [LOCK IN SHARE MODE](https://mariadb.com/kb/en/library/lock-in-share-mode/) gives it feels like that's a perfect candidate for dead-locking. Especially since that method does quite a lot of selects on fairly often used tables.

EnTeQuAk:
QA: This affects primarily the signing API and add-on submissions (new uploads and updates). No visible changes, except that things should continue to work as expected.

PR #10579, opened 2019-02-01
Fix devhub tests for python3 

### LINKED ISSUE #10421: Make python3 fixes to "devhub" job
Python 3 fixes for https://travis-ci.org/mozilla/addons-server/jobs/480441082

eviljeff:
fixes #10421


eviljeff:
ugh, I got the dreaded 120 too

diox:
Shower thought: keep moving the fixed jobs to python3working, but allow it to fail for now. Merge a bunch of things, then circle back to find a fix for the error 120 (I already spent hours on it with no success..)

eviljeff:
If we move them to the working section then all the prs and commits look like they're failing, which makes it harder to see when they're **actually** failing. I think devhub is just going to have to remain in the non-working section along with main for now.

PR #10581, opened 2019-02-03
Fixes #10511: Added ugettext() around exceptions raised by API search filters

### LINKED ISSUE #10511: Translate search API error messages
The search API can raise some errors when the parameters being passed are incorrect. For instance, if the client provides an invalid combination of app/type/category parameters, it will return `Invalid combination of "app", "type" and "category" parameters.`

Although these error messages are developer oriented, they could be surfaced by the frontend to users if they manually change query parameters in the URLs. It's fine to keep them relatively technical, but we should translate them just in case.

It's just a matter of finding the `ValueError` being raised in https://github.com/mozilla/addons-server/blob/master/src/olympia/search/filters.py and use `ugettext()` on them.

suhailsinghbains:
Added ugettext() where ValueError was being raised

### DIFF
--- a/src/olympia/search/filters.py
+++ b/src/olympia/search/filters.py
@@ -42,7 +42,9 @@ class AddonQueryParam(object):
             value = self.get_value_from_reverse_dict()
         if self.is_valid(value):
             return value
-        raise ValueError('Invalid "%s" parameter.' % self.query_param)
+        raise ValueError(
+            ugettext('Invalid "%s" parameter.' % self.query_param)
+        )
 
     def is_valid(self, value):
         return value in self.valid_values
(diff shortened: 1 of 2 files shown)


src/olympia/search/filters.py
```diff
@@ -240,10 +240,10 @@ def __init__(self, request):
             self.reverse_dict = [CATEGORIES[app][type_] for type_ in types]
         except KeyError:
             raise ValueError(
-                'Invalid combination of "%s", "%s" and "%s" parameters.' % (
+                ugettext('Invalid combination of "%s", "%s" and "%s" parameters.' % (
```
diox: This makes the line too long, so you'll have to wrap it differently.

PR #10589, opened 2019-02-04
Python 3 fixes for 'files'

diox:
https://github.com/mozilla/addons-server/issues/10424

I'll add a commit that moves them back into `main` (or maybe `addons` ? Seems more logical) but first want to see travis pass, confirming what I'm seeing locally.


src/olympia/files/tests/test_views.py
```diff
@@ -457,13 +458,9 @@ def test_unicode(self):
         res = self.client.get(self.file_url(u'\u1109\u1161\u11a9'))
```
diox: This test was fixed by changing the zip file to set the relevant encoding flag in it, because Python 3 behavior for zipfiles without that flag is different now.

Before, it used to return the filenames as bytes, and we'd decode that using utf-8 ourselves. Now, *if the flag is absent* it uses `cp437`, which the the "legacy" encoding for zipfiles, and otherwise falls back to utf-8. The end result for zip files without the flag set is the filenames may look weird in some circumstances, but ultimately should still be coherent.

src/olympia/files/forms.py
```diff
@@ -38,8 +38,7 @@ def render_options(self, context):
         def option(files, label=None, deleted=False, channel=None):
             # Make sure that if there's a non-disabled version,
             # that's the one we use for the ID.
-            files.sort(lambda a, b: ((a.status == amo.STATUS_DISABLED) -
-                                     (b.status == amo.STATUS_DISABLED)))
+            sorted(files, key=lambda a: a.status == amo.STATUS_DISABLED)
```
EnTeQuAk: Wait, that won't work… `sorted` doesn't do the sorting inline but returns the sorted result.

src/olympia/files/tests/test_views.py
EnTeQuAk: Especially since we don't really touch or modify or extract them anymore. I think that should work, might be worth getting QA trying to break it though.

src/olympia/files/forms.py
diox: Ahah I'm dumb :) I wanted to return it but forgot. Funny enough the test wasn't right and I changed it... I guess it's still not right since it's passing.

diox: This is fixed in https://github.com/mozilla/addons-server/pull/10589/commits/1b8f1cd72ee2f1ff0188df07b21ab0944baeb0d5

PR #10595, opened 2019-02-05
Fix missing languages in stats from product_details.

EnTeQuAk:
Some parts actually... need all the languages.

Fixes mozilla/addons#891

### DIFF
--- a/src/olympia/core/languages.py
+++ b/src/olympia/core/languages.py
@@ -1,3 +1,7 @@
+# Languages that we support on AMO, e.g we have translations for them
+# and they're tested.
+# They're automatically added to `settings.AMO_LANGUAGES` in
+# `lib.settings_base`
 LANGUAGE_MAPPING = {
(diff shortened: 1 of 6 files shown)


src/olympia/core/languages.py
```diff
@@ -232,3 +236,399 @@
...
+    },
+    u'csb': {
+        'english': u'Kashubian',
+        'native': u'Kasz\xebbsczi'
+    },
+    u'cy': {
+        'english': u'Welsh',
+        'native': u'Cymraeg'
+    },
+    u'dbg': {
```
diox: Don't need that one

EnTeQuAk: Indeed, removed in 175d9cd4a7

EnTeQuAk:
175d9cd4a7 should fix the py3 tests, I probably should start running all (py3 and py2) tests locally first at some point.

PR #10598, opened 2019-02-05
Remaining Python 3 fixes for lib/amo (except lib/git.py)

diox:
Progress towards fixing https://github.com/mozilla/addons-server/issues/10423

### DIFF
--- a/src/olympia/amo/tests/test_middleware.py
+++ b/src/olympia/amo/tests/test_middleware.py
@@ -116,8 +116,8 @@ class TestNoDjangoDebugToolbar(TestCase):
     def test_no_django_debug_toolbar(self):
         with self.settings(DEBUG=False):
             res = self.client.get(reverse('home'), follow=True)
-            assert 'djDebug' not in res.content
-            assert 'debug_toolbar' not in res.content
+            assert b'djDebug' not in res.content
+            assert b'debug_toolbar' not in res.content
 
 
 def test_hide_password_middleware():
--- a/src/olympia/amo/tests/test_views.py
+++ b/src/olympia/amo/tests/test_views.py
@@ -332,10 +332,10 @@ class TestOtherStuff(TestCase):
         self.assertCloseToNow(response['Expires'],
                               now=datetime.now() + timedelta(days=365))
 
-        en = self.client.get(reverse('jsi18n')).content
+        en = self.client.get(reverse('jsi18n')).content.decode('utf-8')
 
         with self.activate('fr'):
-            fr = self.client.get(reverse('jsi18n')).content
+            fr = self.client.get(reverse('jsi18n')).content.decode('utf-8')
 
         assert en != fr
 
@@ -424,7 +424,7 @@ class TestRobots(TestCase):
         url = reverse('collections.list')
         response = self.client.get('/robots.txt')
         assert response.status_code == 200
-        assert b'Disallow: %s' % url in response.content
+        assert 'Disallow: %s' % url in response.content.decode('utf-8')
 
     @override_settings(ENGAGE_ROBOTS=True)
     def test_allow_mozilla_collections(self):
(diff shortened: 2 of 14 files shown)


diox:
Travis didn't agree (well, it was green, but that's because those tests haven't been added to the "working" section yet). I've made more fixes, another round will be necessary but they'll probably be more involved so that'll be in a different PR.

src/olympia/amo/tests/test_views.py
```diff
@@ -433,7 +433,7 @@ def test_allow_mozilla_collections(self):
                              settings.TASK_USER_ID)
         response = self.client.get('/robots.txt')
         assert response.status_code == 200
-        assert 'Allow: {}'.format(url) in response.content
+        assert b'Allow: {}'.format(url) in response.content
```
EnTeQuAk: I'm sure bytes don't have a `.format()` method in python 3. I think you should decode `response.content` instead

src/olympia/lib/crypto/signing.py
```diff
@@ -130,7 +130,7 @@ def autograph_sign_data(file_obj):
     signed_manifest = six.text_type(jar.signatures)
 
     signing_request = [{
-        'input': b64encode(force_bytes(signed_manifest)),
+        'input': force_text(b64encode(force_bytes(signed_manifest))),
```
EnTeQuAk: Just double checked, this is indeed what we want/need here.

diox: Yeah. It was broken because the contents weren't `str`, so that was failing to be encoded in json.

src/olympia/amo/tests/test_views.py
diox: Yeah, that's right. I really should ban the habit of using `b'...' in response.content`.

diox: https://github.com/mozilla/addons-server/pull/10598/commits/5d120d5615b5e994375eb638789fe082975484ac

src/olympia/lib/crypto/signing.py
diox: Also had to do this in the other function in https://github.com/mozilla/addons-server/pull/10598/commits/1bee2d718984369645ba1868add692413fbea051 (locally this was working, which makes it difficult to track)

src/olympia/amo/tests/test_middleware.py
```diff
@@ -116,8 +116,8 @@ class TestNoDjangoDebugToolbar(TestCase):
     def test_no_django_debug_toolbar(self):
         with self.settings(DEBUG=False):
             res = self.client.get(reverse('home'), follow=True)
-            assert 'djDebug' not in res.content
-            assert 'debug_toolbar' not in res.content
+            assert b'djDebug' not in res.content
+            assert b'debug_toolbar' not in res.content
```
EnTeQuAk: Let's use `force_text(res.content)` everywhere maybe?

EnTeQuAk: Or `u'djDebug' not in res.content.decode('utf-8')` but we should try to find one style to do this in all our tests primarily

src/olympia/signing/tests/test_views.py
```diff
@@ -523,8 +523,8 @@ def test_akismet_reports_created_spam_outcome_action_taken(self):
 
         validation_response = self.get(self.url(self.guid, '3.0'))
         assert validation_response.status_code == 200
-        assert 'spam' in validation_response.content
-        data = json.loads(validation_response.content)
+        assert b'spam' in validation_response.content
```
EnTeQuAk: should probably use `.decode(...)` or `force_text` too here.´ imho

src/olympia/amo/tests/test_middleware.py
eviljeff: fwiw, we've used the `b'foo'` fix _a lot_ in our other test changes (i.e. I'm not saying we shouldn't do this - and I made the same suggestion in another pr - but we're already far down the road of testing for bytestrings in raw `.content`)

EnTeQuAk: Yeah, thinking about it, I don't think it matters much anyway. Any way works in regards to testing so :man_shrugging:

eviljeff:
merging this to make identifying the remaining broken tests easier

PR #10610, opened 2019-02-06
Fix lib/git related tests for python 3 compatibility.

EnTeQuAk:
refs #10423


eviljeff:
you need to fix the test fails, obviously 😛

Issue #10471, opened 2019-01-21
Add error::BytesWarning to pytest filterwarnings


EnTeQuAk:
To issue warnings about `str(bytes_instance)` and `str(bytearray_instance)` as well as bytes/str comparisons we need to add `error:BytesWarning` to our `filterwarnings` settings in `setup.cfg`

That'll help make sure to catch those nifty little edge-cases.

eviljeff:
#10622 is passing, apart from the ES based tests. But I'm not sure the ES fails are fixable.  This is a sample error:
```
____________ ERROR at setup of TestAddonIndexerWithES.test_mapping _____________
src/olympia/amo/tests/__init__.py:878: in setUpClass
    super(ESTestCase, cls).setUpClass()
.tox/es/lib/python3.6/site-packages/django/test/testcases.py:1042: in setUpClass
    cls.setUpTestData()
src/olympia/amo/tests/__init__.py:883: in setUpTestData
    setup_es_test_data(cls.es)
src/olympia/amo/tests/__init__.py:97: in setup_es_test_data
    es.cluster.health()
.tox/es/lib/python3.6/site-packages/elasticsearch/client/utils.py:73: in _wrapped
    return func(*args, params=params, **kwargs)
.tox/es/lib/python3.6/site-packages/elasticsearch/client/cluster.py:33: in health
    'health', index), params=params)
.tox/es/lib/python3.6/site-packages/elasticsearch/client/utils.py:47: in _make_path
    quote_plus(_escape(p), b',*') for p in parts if p not in SKIP_IN_PATH)
.tox/es/lib/python3.6/site-packages/elasticsearch/client/utils.py:47: in <genexpr>
    quote_plus(_escape(p), b',*') for p in parts if p not in SKIP_IN_PATH)
E   BytesWarning: Comparison between bytes and string (it looks like ES is not running, try starting it or don't run ES tests: make test_no_es)
```
Which is from https://github.com/elastic/elasticsearch-py/blob/5.5.x/elasticsearch/client/utils.py#L47 - `SKIP_IN_PATH` is [defined](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L9) as `(None, '', b'', [], ())` - its a tuple of mixed binary and unicode strings so the generator will _always_ fail, regardless of whether `parts` contains binary or unicode strings.

eviljeff:
We could just not run the `es` job with `-bb`, I suppose.

[...]

PR #10622, opened 2019-02-07
try running pytest via python -m to force py3 bytes errors

eviljeff:
for #10471


pytest
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+exec python -bb -m pytest "$@"
```
eviljeff: Do we need this?  It's to make local testing easier but it's only and extra few chars.  Anyway, if we do then someone not on Windows will need to change the file executable flag 😉

diox: I vote we skip it, because it's confusing to have it in our repos like this.

eviljeff:
So, this isn't working right now - https://travis-ci.org/mozilla/addons-server/jobs/490603932#L2660 is the test that specifically verifies the BytesWarning error is raised and it's not being raised.
Locally I get _a lot_ of errors (from the base templates, at least)

pytest
eviljeff: dropped in https://github.com/mozilla/addons-server/pull/10622/commits/5030b022401034a7110a2d41737ab90b52980b69

diox:
So, there are 3 issues here:
1. For some reason the `BytesWarning` error is not raised on travis, preventing us from properly seeing the errors.
2. `elasticsearch-py` has a [`SKIP_IN_PATH = (None, '', b'', [], ())`](https://github.com/elastic/elasticsearch-py/blob/3e015cf5e51d7c515803d799c55e8e8eaf481ba3/elasticsearch/client/utils.py#L9) that it uses on different kinf of data with a `in`, so sometimes it ends up comparing `bytes` and `str`, raising the exception.
3. `waffle` converts all its cache keys to `bytes`, but django cache backend deal with `str`. That lead to a concatenation between `bytes` and `str`, raising the exception.

Once these issues are worked around locally, I only get 2 failures that need our attention. Here they are with their fix:
```diff
diff --git a/src/olympia/addons/tests/test_cron.py b/src/olympia/addons/tests/test_cron.py
index a8aa66f763..6ef836e3b5 100644
--- a/src/olympia/addons/tests/test_cron.py
+++ b/src/olympia/addons/tests/test_cron.py
@@ -160,14 +160,12 @@ class TestHideDisabledFiles(TestCase):
 
         # Check that we called `move_stored_file` for f2 properly
         f2 = self.f2
-        mv_mock.assert_called_with(
-            force_bytes(f2.file_path), force_bytes(f2.guarded_file_path))
+        mv_mock.assert_called_with(f2.file_path, f2.guarded_file_path)
 
         # Check that we called `move_stored_file` for f1 properly
         f1 = self.f1
         mv_mock.call_args = mv_mock.call_args_list[0]
-        mv_mock.assert_called_with(
-            force_bytes(f1.file_path), force_bytes(f1.guarded_file_path))
+        mv_mock.assert_called_with(f1.file_path, f1.guarded_file_path)
 
         # Make sure we called `mv` twice despite an `IOError` for the first
         # file
--- a/src/olympia/files/models.py
+++ b/src/olympia/files/models.py
@@ -154,7 +154,7 @@ class File(OnChangeMixin, ModelBase):
         assert parsed_data is not None
 
         file_ = cls(version=version, platform=platform)
-        upload.path = force_bytes(nfd_str(upload.path))
+        upload.path = force_text(nfd_str(upload.path))
         ext = force_text(os.path.splitext(upload.path)[1])
         if ext == '.jar':
             ext = '.xpi'
@@ -290,17 +290,13 @@ class File(OnChangeMixin, ModelBase):
 
     def move_file(self, source, destination, log_message):
         """Move a file from `source` to `destination`."""
-        # Make sure we are passing bytes to Python's io system.
-        source, destination = force_bytes(source), force_bytes(destination)
-
         try:
             if storage.exists(source):
                 log.info(log_message.format(
                     source=source, destination=destination))
                 move_stored_file(source, destination)
         except (UnicodeEncodeError, IOError):
-            msg = 'Move Failure: {} {}'.format(
-                force_bytes(source), force_bytes(destination))
+            msg = 'Move Failure: {} {}'.format(source, destination)
             log.exception(msg)
 
     def hide_disabled_file(self):
```

We should test that fix to see if it makes sense, particularly with python 2, but I think it does :) The problem is though, do we want to continue investing into enabling `-bb` given the 3 issues above.

eviljeff:
I think we should fix those two files, but beyond that I don't think it's practical to do much else.

The elasticsearch-py comparision seems harmless (they're testing for empty string); and the waffle bug is weird but I can't think of how it would actually affect us (unless we're choosing very unwise waffle names or deciding to specify the waffle names in bytestring).

If `-bb` becomes the default in a future python we could re-open this, but unless we discover some way to have `-bb` run properly under Travis there's nothing else we can do here.

PR #10640, opened 2019-02-08
switch local dev and addons-dev to python3

### LINKED ISSUE #10639: run addons-dev and local dev on python3
Stage one of the python3 upgrade is complete - tests pass.
Initial testing locally doesn't throw up any errors so it's time to have addons-dev use python3.  Stage and prod will only get python3 after a sufficiently long period of time.

eviljeff:
fixes #10639


circle.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           name: Build docker image and push to repo
           command: |
             docker version
-            docker build -t app:build -f Dockerfile.deploy .
+            docker build -t app:build -f Dockerfile.py3.deploy .
```
EnTeQuAk: These are used for addons-dev.allizom.org

eviljeff: yep.  The commit is misnamed - the issue was about having addons-dev run python3 also.

diox: So, the main problem with that approach (with the other change below in the same file) is that we'd build `latest` (for dev) *exclusively* for python 3, and tags (for stage/prod) *exclusively* for python 2.

Also note that currently, for local dev, we use a different dockerhub repo (https://hub.docker.com/r/addons/addons-server instead of https://hub.docker.com/r/mozilla/addons-server) that is configured to build `latest` and `latest-py3` (it's not currently configured to build a Python 3 image for tags/branches, but it could be trivially by just adding it to the docker hub settings). It'd be nice to be consistent (it'd be even nicer to use the same repo for everything, but eh, baby steps).

So, I think we need to do the following:
- Build `latest` (still py2) and `latest-py3` (obviously py3) with circle ci. Same for tags.
- Deploy `latest-py3` to dev (`latest` would be unused except for if/when we need to switch back) and don't change deploy for the rest for now.

Dockerfile
```diff
@@ -1,7 +1,9 @@
-FROM python:2.7.15-slim-stretch
+FROM python:3.6-slim-stretch
```
diox: If you do that, it breaks the current setup. You're referencing `latest-py3` in the `docker-compose.yml`, so I don't think it's necessary to change the main `Dockerfile` just yet? `latest-py3` is setup to be built automatically by docker hub from the repo using `Dockerfile.python3`.

diox: The alternative would be to switch local env completely to python 3 without possibility of going back, in which case you wouldn't want to reference `latest-py3` anymore in `docker-compose.yml`, but I think it's a little early for that. Switching to python 3 by default makes sense, but we need to be able to quickly switch back, even locally.

eviljeff: Okay, I wasn't sure if having an inconsistent Dockerfile to what we use in docker-compose.yml was going to cause problems.

circle.yml
eviljeff: Sounds good.  What lines are you suggesting be changed exactly?

diox: I'm not sure :) @autrilla do you know how we can do what I propose above efficiently? Can this be done in parallel ? Wouldn't want to make the deploys slower.

Dockerfile
eviljeff: https://github.com/mozilla/addons-server/pull/10640/commits/00681c1aecdee381a03c1669122a9fbf1c67bb7e reverts the DockerFile changes

circle.yml
eviljeff: I reversed the Dockerfile changes but I'm not clear how it affects the way forward here.  We _do_ want the docker image deployed on dev to be the py3 one.

autrilla: > I'm not sure :) @autrilla do you know how we can do what I propose above efficiently? Can this be done in parallel ? Wouldn't want to make the deploys slower.

This can certainly be done in parallel with the (relatively) new CircleCI workflows. E.g. autograph runs several things in parallel:

- [CircleCI config](https://github.com/mozilla-services/autograph/blob/master/.circleci/config.yml)
- [Example workflow](https://circleci.com/workflow-run/3f354e0b-275c-433b-9731-382cd2e95a13), shows a bunch of things running in parallel

diox: We're already using workflows so it doesn't sound like we'd need to add much. We only need to care about the master branch deploys right now, so I think we need to:
- Split `build-deploy` into 2 parts: a build part and a deploy part
- Clone the build part, do a py2 and a py3 version
- Make the deploy part require both builds in the workflow, and make it push both `${DOCKERHUB_REPO}:latest` and `${DOCKERHUB_REPO}:latest-py3`.

Bonus point for folding the build-release into the same jobs (just handling the fact that the docker tag is different) since it's essentially copy/pasted right now.

Once this is done, have ops make our pipeline on their end use the `latest-py3` image - but crucially, we'll have both py2 and py3 versions available, letting us switch to either if needed.

.circleci/config.yml
```diff
@@ -1,11 +1,16 @@
...
-  build-deploy: # build for the master branch
-    machine: true
-    working_directory: ~/addons-server
+commands:
+  build_and_push_container_image:
+    description: "Builds and pushes a Docker image"
+    parameters:
+      dockerfile:
+        type: string
+        default: "Dockerfile"
```
diox: It doesn't matter much because you always pass it explicitly, but default to `Dockerfile.py2.deploy` for consistency ?

Dockerfile.py3.deploy
```diff
@@ -0,0 +1,104 @@
+FROM python:3.6-slim-stretch
```
diox: Can you make the naming consistent with the one we use for local dev, which is `Dockerfile.python3` ? Obviously the CircleCI config would then need to be updated.


 (FWIW the reason it wasn't called `Dockerfile.py3` was because some editors consider `.py3` files as python files by default)

PR #10656, opened 2019-02-11
Start mocking elasticsearch earlier, before loading django fixtures

### LINKED ISSUE #10655: Start mocking elasticsearch earlier, before loading django fixtures
We noticed while playing with https://github.com/mozilla/addons-server/issues/10471 that we aren't mocking Elasticsearch as early as possible in tests. Specifically, the mocking starts after django fixtures has been loaded, slowing down tests as we have to index addons from those fixtures. Let's fix that.

diox:
We were mocking es using a pytest autouse fixture, which are apparently called too late, after the loading of django fixtures. This was causing useless index calls, possibly slowing down the tests
slightly.

Fix #10655

### DIFF
--- a/src/olympia/amo/tests/__init__.py
+++ b/src/olympia/amo/tests/__init__.py
@@ -356,6 +356,10 @@ ES_patchers = [
 
 
 def start_es_mocks():
+    # Before starting to mock, stop them first. That way we ensure we're not
+    # trying to mock over an existing mock, which would be problematic since
+    # we use spec=True.
+    stop_es_mocks()
     for patch in ES_patchers:
         patch.start()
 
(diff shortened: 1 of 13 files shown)


src/olympia/amo/tests/__init__.py
```diff
@@ -424,10 +411,33 @@ def grant_permission(user_obj, rules, name):
...
 
 
-class TestCase(PatchMixin, InitializeSessionMixin, BaseTestCase):
+class TestCase(PatchMixin, InitializeSessionMixin, test.TestCase):
     """Base class for all amo tests."""
     client_class = TestClient
 
+    def _pre_setup(self):
+        super(TestCase, self)._pre_setup()
+        self.client = self.client_class()
```
diox: Actually... we probably don't need that, that's what django does already...

src/olympia/amo/tests/__init__.py
```diff
@@ -356,6 +356,10 @@ def days_ago(days):
 
 
 def start_es_mocks():
+    # Before starting to mock, stop them first. That way we ensure we're not
+    # trying to mock over an existing mock, which would be problematic since
+    # we use spec=True.
+    stop_es_mocks()
```
eviljeff: Nit but... this seems a bit wasteful when we know the es mocks won't have started when it's called from `TestCase`, and wasteful a second time when we're stopping the mocks and starting them again in the autouse fixture.  Is there a way of checking that we've initialized the mocks once already?  If we can't think of anything I guess it'd be a future enhancement kind of thing.

diox: I tried... there is no public API to check that a patcher has been started :(

PR #10673, opened 2019-02-12
install uwsgi with pip to get python2/3 compatablity

### LINKED ISSUE #10639: run addons-dev and local dev on python3

eviljeff:
fixes #10639 again (hopefully)

### DIFF
--- a/Dockerfile.python2.deploy
+++ b/Dockerfile.python2.deploy
@@ -37,8 +37,6 @@ RUN apt-get update && apt-get install -y \
         python-pip                       \
         nodejs                           \
         uuid-dev                         \
-        uwsgi                            \
-        uwsgi-plugin-python              \
         # Git, because we're using git-checkout dependencies
         git                              \
         # Dependencies for mysql-python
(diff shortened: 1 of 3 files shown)


Dockerfile.python2.deploy
```diff
@@ -79,6 +72,11 @@ RUN pip install --no-cache-dir --exists-action=w --no-deps -r requirements/syste
     && pip install --no-cache-dir --exists-action=w --no-deps -r requirements/prod_py2.txt         \
     && pip install --no-cache-dir --exists-action=w --no-deps -e .
 
+# Install uwsgi statsd exporter to collect metrics from uwsgi when deployed
+WORKDIR /usr/lib/uwsgi/plugins
+RUN uwsgi --build-plugin https://github.com/Datadog/uwsgi-dogstatsd && \
+    rm -rf uwsgi-dogstatsd
```
diox: I think we need to `WORKDIR /data/olympia` after that (maybe we can even move the `WORKDIR` statement above, I'm not sure)

Dockerfile.python3.deploy
```diff
@@ -79,6 +72,11 @@ RUN pip3 install --no-cache-dir --exists-action=w --no-deps -r requirements/syst
     && pip3 install --no-cache-dir --exists-action=w --no-deps -r requirements/prod_py3.txt         \
     && pip3 install --no-cache-dir --exists-action=w --no-deps -e .
 
+# Install uwsgi statsd exporter to collect metrics from uwsgi when deployed
+WORKDIR /usr/lib/uwsgi/plugins
+RUN uwsgi --build-plugin https://github.com/Datadog/uwsgi-dogstatsd && \
+    rm -rf uwsgi-dogstatsd
+
```
diox: Same comment about `WORKDIR` from the other `Dockerfile`

PR #10683, opened 2019-02-13
Fix codestyle and docs jobs

### LINKED ISSUE #10676: Fix codestyle and docs jobs in travis
After #10592 the codestyle job isn't actually doing a flake8 check because the `make` invocation is missing the `-f` to specify the correct Makefile.  (It didn't need one before because the recipe was in the primary Makefile before, and now it's in Makefile-docker).
Also the docs job is broken (for an unknown period of time) for the opposite reason - the recipe is in Makefile, not Makefile-docker.

We should fix the things.  Bonus for simplifying and making it so we don't have to do `make -f ...` everywhere in tox.ini

diox:
Fix #10676


docs/_build/.gitignore
```diff
@@ -1 +0,0 @@
-*
```
eviljeff: was this deliberate, btw?

diox: Oops, absolutely not. I played with force-rebuilding the docs a bunch of times and at some point the result of a `rm *` must have slipped through. I'll revert that.

PR #10684, opened 2019-02-13
Expose version list for diffing, rework reviewers api to be more concise

### LINKED ISSUE #10432: Reviewers/Browse API: Expose all related versions
https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#browse Exposes details for a specific version but for the addons-code-manager to properly implement the version-selector we'll have to expose a list of versions that the user can diff against.

Refs https://github.com/mozilla/addons-code-manager/issues/39

EnTeQuAk:
We're now mounting any listing/browsing/future diffing APIs under the
already existing addon/<addon_id>/ endpoint which makes the whole
reviewer APIs behave more consistently.

Also implements the listing of versions that can be used for browsing
and diffing.

Fixes #10432

cc @willdurand @bobsilverberg and @kumar303 that the URLs will unfortunately change and now require an addon-id, that will be given through the link to the code-manager. That addon-id will then be used to retrieve the list of versions that can be used for diffing or other actions later.


kumar303:
Thanks, noted.

EnTeQuAk:
Meh, travis didn't communicate the progress of that test-suite, both are actually green :)

src/olympia/reviewers/views.py
```diff
@@ -1240,43 +1241,85 @@ def flags(self, request, **kwargs):
...
+
+    def check_object_permissions(self, request, obj):
+        """Check permissions against the parent add-on object."""
+        return super(ReviewAddonVersionViewSet, self).check_object_permissions(
+            request, obj.addon)
+
+    def list(self, request, *args, **kwargs):
+        """Return all (re)viewable versions for this add-on.
+
+        Full list, no pagination."""
```
diox: Why no pagination? This seems like a bad idea, some add-ons have a lot of versions...

EnTeQuAk: Because that list is being shown in a drop-down of some kind and I personally don't think fetching the versions through pagination is a worthwhile undertaking.

I don't know how the current file-viewer handles huge amounts of versions but I'd guess, it just looks very stupid.

EnTeQuAk:
@diox or @eviljeff does one of you have some time for this today? I have other PRs in the queue that built on top of this one, as it refactors how the API works.

docs/topics/api/reviewers.rst
```diff
@@ -84,6 +84,29 @@ add-on.
...
+        permission for listed add-ons as well as ``Addons:ReviewUnlisted`` for
+        unlisted add-ons. Additionally the current user can also be the owner
+        of the add-on.
+
+.. http:get:: /api/v4/reviewers/addon/(int:addon_id)/versions/
+
+    :>json int id: The version id.
+    :>json string channel: The version channel, which determines its visibility on the site. Can be either ``unlisted`` or ``listed``.
+    :>json string version: The version number string for the version.
+    :>json string url: The (absolute) version detail URL.
```
eviljeff: reminder that this will be dropped in #10698 so you should remove it from here too.

src/olympia/reviewers/views.py
eviljeff: yeah... (see https://reviewers.addons.mozilla.org/en-US/firefox/files/compare/1683453...1679753/ for 83 versions - it's slow)

But if it _only_ every returns minimal information (like the serializer does now) then it's probably okay.  It'll just get terrible quickly if we start returning nested serializers and other objects (e.g. translations).  Maybe add a comment to the serializer for future us to not add anything crazy to it.

src/olympia/reviewers/serializers.py
```diff
@@ -203,3 +203,14 @@ def get_validation_url(self, obj):
...
+
+class DiffableVersionSerializer(VersionSerializer):
+    should_show_channel = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Version
+        fields = ('id', 'channel', 'url', 'version', 'should_show_channel')
+
+    def get_should_show_channel(self, obj):
+        return self.context.get('should_show_channel', False)
```
eviljeff: So we return the same value for every version in the list response?

EnTeQuAk: Correct, I didn't want to nest the list of versions tbh so I decided to list that for every version.

On the other hand, we're returning the `channel` for every version so the client might be able to figure this out themselves? cc @willdurand (some context: If we have mixed listed/unlisted versions that can be compared we want to show the channel, otherwise we don't need to)

src/olympia/reviewers/views.py
EnTeQuAk: I could even make the query a bit quicker by reducing the number of fields we are selecting to the ones we actually need which should be fine then.

eviljeff: `id`, `channel` `version` shouldn't be terrible.

src/olympia/reviewers/serializers.py
willdurand: Thanks. Could you maybe add this note about channels in the API doc?

src/olympia/reviewers/views.py
diox: Best way to make it super fast (besides caching and accepting that the first request is going to be slow) would be to only select columns that are part of the same composite index ; that way MySQL wouldn't need to fetch the data from disk. Unfortunately we don't have an index that match those fields...

EnTeQuAk: I personally don't think we have to make it super-fast, fast enough should be fine as long as the loading doesn't time-out and stays responsive for the user.

src/olympia/reviewers/serializers.py
EnTeQuAk: Added in 3801b2d

docs/topics/api/reviewers.rst
EnTeQuAk: Fixed in 3119c5f

Issue #9349, opened 2018-09-04
File watching uses lots of CPU running the docker dev-env under OSX


muffinresearch:
Running the docker development environment is currently spinning up a development version of addons-frontend for ui-testing here: https://github.com/mozilla/addons-frontend/blob/3beddb39fee76d6a2891d5b91ee9d2464504bf01/package.json#L69

This is going to use webpack dev server and additional file watching to know when to restart the server. When I run the docker env on OSX I see 400% CPU usage and the whole thing is unusable. To use addons-server under docker I have to comment out the frontend part of the `docker-compose.yml`.

Unless there's a good fix to make file-watching performant (unlikely since these problems have been a problem in docker for OSX for a long time) we should consider switching the frontend for ui-testing to use a production build without any file watching. I think we're unlikely to start developing the frontend under docker since it's much easier to point the frontend at -dev or stage using the development server directly from addons-frontend.

/cc @jrbenny35

EnTeQuAk:
I agree, the docker setup of addons-frontend isn't really usable to start developing on addons-frontend and it's not really meant to be.

So we should really be using the setup without any file-watching but still make it listen directly to the local addons-server api.

PR #10692, opened 2019-02-14
Fix addons-frontend running as default frontend.

EnTeQuAk:
The frontend runs quite nicely with the changes from addons-nginx container but the initial page load runs into some wacky connection errors.

https://www.irccloud.com/pastebin/gTMteTUf/

This change fixes this.

Fixes https://github.com/mozilla/addons-server/issues/10855


muffinresearch:
See also https://github.com/mozilla/addons-server/issues/9349

PR #10695, opened 2019-02-14
Raise an exception when we can't sign a file instead of returning None

### LINKED ISSUE #10693: An add-on can end up being approved but not signed
https://github.com/mozilla/addons/issues/909 is actually several bugs linked together, this issue is about the last part.

If an add-on somehow manages to get into awaiting review state with no apps (or no Firefox apps) attached to its version, here is what will happen when approving it (via reviewer tools or `auto_approve`):
- It goes through `process_public()`, which [calls `sign_file(file_)`](https://github.com/mozilla/addons-server/blob/39a6b46d894b987d402744b59c30dd4d42de837d/src/olympia/reviewers/utils.py#L641-L643).
- `sign_file(file_obj)` calls `supports_firefox(file_obj)`, finds that the add-on does not support Firefox, [returns without signing](https://github.com/mozilla/addons-server/blob/39a6b46d894b987d402744b59c30dd4d42de837d/src/olympia/lib/crypto/signing.py#L214-L218)
- `process_public()` happily ignores that and set the addon public

The moral of the story is that `process_public()` and other code calling signing should check the return value of `sign_file()`, and do /something/ if it returns `None`, keeping in mind that it can happen inside `auto_approve`, so just blindly raising an exception is wrong. In addition `sign_file()` needs to be changed to not just return `None` when a file is already signed with Mozilla cert (`is_mozilla_signed_extension`). It could return the `File` object in that case.

diox:
This will prevent us from accidentally making public an add-on we can't sign for some reason. That will make such issues more explicit by either returning a 500 to the user (unlisted signing API), raising an exception that will show up in sentry (tasks) or leave the version awaiting review (every case above + auto_approve). The affected add-ons will still be somewhat broken but they'll be in a more consistent state.

Fixes #10693

### DIFF
--- a/src/olympia/addons/tests/test_commands.py
+++ b/src/olympia/addons/tests/test_commands.py
@@ -165,6 +165,8 @@ def test_approve_addons_approve_files_no_review_type():
     assert file_.reload().status == amo.STATUS_PUBLIC
 
 
+@pytest.mark.django_db
+@mock.patch('olympia.reviewers.utils.sign_file', lambda f: None)
 def test_approve_addons_approve_files(use_case, mozilla_user):
     """Files are approved using the correct review type.
 
--- a/src/olympia/reviewers/management/commands/auto_approve.py
+++ b/src/olympia/reviewers/management/commands/auto_approve.py
@@ -13,6 +13,7 @@ import olympia.core.logger
 
 from olympia import amo
 from olympia.files.utils import atomic_lock
+from olympia.lib.crypto.signing import SigningError
 from olympia.reviewers.models import (
     AutoApprovalNotEnoughFilesError, AutoApprovalNoValidationResultError,
     AutoApprovalSummary, clear_reviewing_cache, set_reviewing_cache)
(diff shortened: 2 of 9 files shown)


src/olympia/addons/tests/test_commands.py
```diff
@@ -165,6 +165,8 @@ def test_approve_addons_approve_files_no_review_type():
     assert file_.reload().status == amo.STATUS_PUBLIC
 
 
+@pytest.mark.django_db
+@mock.patch('olympia.reviewers.utils.sign_file', lambda f: None)
```
eviljeff: curious, does `lambda f: None` work better than setting `return_value`?

src/olympia/reviewers/management/commands/auto_approve.py
```diff
@@ -99,15 +100,17 @@ def process(self, version):
...
 
         except (AutoApprovalNotEnoughFilesError,
-                AutoApprovalNoValidationResultError):
+                AutoApprovalNoValidationResultError,
+                SigningError):
             log.info(
                 'Version %s was skipped either because it had no '
-                'file or because it had no validation attached.', version)
+                'files or had no validation attached to its files, or signing '
+                'failed', version)
```
eviljeff: Should we break out `SigningError` from `AutoApprovalNoValidationResultError` for logging and stats counting?

src/olympia/addons/tests/test_commands.py
diox: The nice thing about the `lambda f: None` as second argument is that I don't have to pass the mock as argument. But I could use `autospec=True` instead, that would be more reliable.

src/olympia/reviewers/management/commands/auto_approve.py
diox: My thought process was, if it's a signing error, there will be something else to tell us, but you're right, maybe that's not enough. I'll see what I can do about separating it from the rest.

diox: I did in 17699e15672de53f5ad0fad2ed0221bede58e3b9 and I ended up also reworking the way the transaction wraps everything, so that we can roll back the `AutoApprovalSummary` on `SigningError` to have better consistency. Please re-review when tests pass.

PR #10698, opened 2019-02-14
drop version detail link

### LINKED ISSUE #10645: URL `/addon/:slug/versions/:versionId/` is not redirected anymore
I found a new issue with this URL (found in the Dev Hub): https://addons-dev.allizom.org/en-US/firefox/addon/guarani-paraguay-lang-pack/versions/66.0buildid20190207161357/

The new frontend does not support URLs like this, we only support `/addon/:slug/versions/`.

@bobsilverberg @muffinresearch do you remember any decision about that? (i.e. about not supporting a URL for a specific version of an add-on).

_Originally posted by @willdurand in https://github.com/mozilla/addons-server/issues/10428#issuecomment-461726490_

eviljeff:
fixes #10645

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -162,13 +162,12 @@ class SimpleVersionSerializer(MinimalVersionSerializer):
     is_strict_compatibility_enabled = serializers.SerializerMethodField()
     license = CompactLicenseSerializer()
     release_notes = TranslationSerializerField()
-    url = serializers.SerializerMethodField()
 
     class Meta:
         model = Version
         fields = ('id', 'compatibility', 'edit_url', 'files',
                   'is_strict_compatibility_enabled', 'license',
-                  'release_notes', 'reviewed', 'url', 'version')
+                  'release_notes', 'reviewed', 'version')
 
     def to_representation(self, instance):
         # Help the LicenseSerializer find the version we're currently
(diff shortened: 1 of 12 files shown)


src/olympia/addons/serializers.py
```diff
@@ -162,13 +162,12 @@ class SimpleVersionSerializer(MinimalVersionSerializer):
     is_strict_compatibility_enabled = serializers.SerializerMethodField()
     license = CompactLicenseSerializer()
     release_notes = TranslationSerializerField()
-    url = serializers.SerializerMethodField()
```
eviljeff: I was worried about breaking the api but it seems we never documented url in response anyway!

EnTeQuAk: Actually, I make use of it in https://github.com/mozilla/addons-server/pull/10684/files#diff-461691f0e19e8fe3e4d86b4b69858cb5R208

EnTeQuAk: So, maybe, if we can keep it? Otherwise I'll re-implement it for my own serializer there.

EnTeQuAk: On the other hand… I'll just drop it from my PR too, I don't think we'll need it anyway.

eviljeff: yeah I dropped it because there isn't a (publicly accessible) unique url for a version any more.  Just returning the /versions page for each version seemed redundant.

PR #10700, opened 2019-02-15
django2.2 changes and deprecation removals

### LINKED ISSUE #10674: Fix known backwards incompatible changes in django2.0,1,2
essentially we need to read through:
https://docs.djangoproject.com/en/2.2/internals/deprecation/#deprecation-removed-in-2-0
https://docs.djangoproject.com/en/2.2/internals/deprecation/#deprecation-removed-in-2-1
https://docs.djangoproject.com/en/2.2/releases/2.0/#backwards-incompatible-changes-in-2-0
https://docs.djangoproject.com/en/2.2/releases/2.1/#backwards-incompatible-changes-in-2-1
https://docs.djangoproject.com/en/2.2/releases/2.2/#backwards-incompatible-changes-in-2-2
and identify things that may break.   Even where we appear to use the feature now changed it's possible we won't actually see any difference so probably easier to err on the side of trivial changes only unless/until we get test failures and errors.

eviljeff:
I think this covers the remaining things I found, so fixes #10674 (the proof of the eating will be when we actually start testing with django2.2)

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -2858,8 +2858,8 @@ class TestAddonSearchView(ESTestCase):
         addon_factory()
         self.reindex(Addon)
 
-        # need to keep this force_text because pre django2.2
-        # urlsafe_base64_encode returns a bytestring and a string after
+        # We need to keep force_text because urlsafe_base64_encode only starts
+        # returning a string from Django 2.2 onwards, before that a bytestring.
         param = 'rta:%s' % force_text(
             urlsafe_base64_encode(force_bytes(addon.guid)))
 
@@ -2878,8 +2878,8 @@ class TestAddonSearchView(ESTestCase):
         addon_factory()
         self.reindex(Addon)
 
-        # need to keep this force_text because pre django2.2
-        # urlsafe_base64_encode returns a bytestring and a string after
+        # We need to keep force_text because urlsafe_base64_encode only starts
+        # returning a string from Django 2.2 onwards, before that a bytestring.
         param = 'rta:%s' % force_text(
             urlsafe_base64_encode(force_bytes(addon.guid)))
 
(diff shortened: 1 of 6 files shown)


src/olympia/addons/tests/test_views.py
```diff
@@ -2858,6 +2858,8 @@ def test_filter_by_guid_return_to_amo(self):
         addon_factory()
         self.reindex(Addon)
 
+        # need to keep this force_text because pre django2.2
+        # urlsafe_base64_encode returns a bytestring and a string after
```
EnTeQuAk: The comment is confusing (to me at least), maybe `need to keep force_text because urlsafe_base64_encode only starts returning a string from Django 2.2 onwards, before that a bytestring.` or so?

src/olympia/search/filters.py
```diff
@@ -163,7 +163,8 @@ def get_value(self):
                     ugettext('Return To AMO is currently disabled')
                 )
             try:
-                value = force_text(urlsafe_base64_decode(value[4:]))
+                value = force_text(
+                    urlsafe_base64_decode(force_text(value[4:])))
```
EnTeQuAk: Does the `_decode` version return a string too after 2.2? If so, let's mark it here too.

eviljeff: No, it's always returned bytes, like the underlying https://docs.python.org/3/library/base64.html#base64.urlsafe_b64decode method, but django changed to require a string as input.  

I didn't bother adding a comment as it will actually fail with an exception in django2.2 if we remove the `force_text` wrapper from the input, unlike like `param = 'rta:%s' % force_text(urlsafe_base64_encode(force_bytes(addon.guid)))` lines where it wouldn't and we'd silently end up with `param = "rta:b'foo''`.  I can add a comment if you think it's clearer though?

eviljeff: https://github.com/mozilla/addons-server/pull/10700/commits/8abb7668faa2b989928a4821c662a16be507169d

src/olympia/addons/tests/test_views.py
eviljeff: https://github.com/mozilla/addons-server/pull/10700/commits/8abb7668faa2b989928a4821c662a16be507169d

PR #10719, opened 2019-02-18
Don't expose 'image' as binary boolean value.

### LINKED ISSUE #10685: Don't expose image binary type
In the reviewers browse api we're exposing the field `binary` which can be a boolean or string but really, should only be a boolean.

We are already exposing the mimetype so the client should be able to handle that properly.

EnTeQuAk:
Fixes #10685

### DIFF
--- a/docs/topics/api/reviewers.rst
+++ b/docs/topics/api/reviewers.rst
@@ -131,7 +131,7 @@ This endpoint allows you to browse through the contents of an Add-on version.
     :>json string file.content: Raw content of the requested file.
     :>json string file.selected_file: The selected file, either from the ``file`` parameter or the default (manifest.json, install.rdf or package.json for Add-ons as well as the XML file for search engines).
     :>json array file.entries[]: The complete file-tree of the extracted XPI.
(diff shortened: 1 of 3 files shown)


eviljeff:
Mmm, so I get
> In the reviewers browse api we're exposing the field `binary` which can be a boolean or string but really, should only be a boolean.

but I don't follow the logic where an image isn't `binary=True` - an image literally is a binary file, rather than a text file.

eviljeff:
Does `binary` need to be replaced with `should_be_displayed_inline` (or something less wordy)?  Wouldn't that be a decision for the client anyway?  (so is this property necessary at all?)

EnTeQuAk:
It would, you're right. I think we may want to rename this to `is_blocklisted_binary` or so, as this is there to avoid the client to do anything at all with .exe files and friends.

Does that sound more reasonable?

EnTeQuAk:
> what is the request/need from the frontend exactly?

We don't want the frontend to touch / open executable files ideally. That's why that's in place but tbh, I don't know how useful it is given that add-ons have changed dramatically in the past.

cc @wagnerand maybe for some ideas

EnTeQuAk:
This will be reworked

Issue #10720, opened 2019-02-18
Stop storing ip_address in abuse reports


diox:
Split from https://github.com/mozilla/addons-server/issues/10643 - in [the PRD for Abuse Reporting in Firefox](https://docs.google.com/document/d/1K5Th9UmpayfVMXadkxU6cDO3sFKOwYozQgyoN0WEGwY/edit?ts=5c4b6e8d#heading=h.af3x6jga7lsy), @jvillalobos said we should stop storing IP Addresses of abuse reports reporters. 

We should determine the region from GeoIP (assuming we have access to a reliable database for this) and store that instead.

EnTeQuAk:
Why do we even have to store the region of the reporter?

Afaik we were storing the ip-addresses to block abuse-reports abuses.

diox:
Not my decision but I assume it can be useful for context, when reporting add-ons that are specific to a country for instance.

jvillalobos:
The region can be useful to identify trends in reporting, and reporting abuses, to an extent. We haven't been using the IP address at all, and it's a privacy issue to keep those stored indefinitely if we're not making use of them.

diox:
We should have access to the GeoIP City/Country database. I'm assuming Country is enough ? Do we need to convert the ip addresses to countries in existing reports ?

jvillalobos:
I believe country should be enough. As for existing reports, either converting or clearing the IP addresses would be good (converting is better).

diox:
QA:
- This should be visible in the admin tool/reviewer tools, instead of the IP the country code is now shown for abuse reports. I'll fire a script to backfill existing data (and then a new issue will get rid of the `ip_address` field entirely, but at the moment it should no longer be used and be completely hidden)

AlexandraMoga:
@diox I've checked both admin and rev tools to see if the IP address is stored anywhere:

[admin] - the Ip address field is empty
![image](https://user-images.githubusercontent.com/31961530/53251519-820f2a80-36c5-11e9-8c54-cab57e75d6a6.png)

[rev tools] - I can't see a user Ip stored anywhere in https://reviewers.addons-dev.allizom.org/en-US/reviewers/abuse-reports/search_by_image
![image](https://user-images.githubusercontent.com/31961530/53251591-b256c900-36c5-11e9-9250-01d1539be4b3.png)

However, the new Country Code field doesn't seem to be populated for any of the abuse reports that I've submitted:

[...]

PR #10733, opened 2019-02-19
Augment the Abuse Report API with new fields

### LINKED ISSUE #10643: Augment abuse API with new fields for add-on reporting within Firefox
PRD: https://docs.google.com/document/d/1K5Th9UmpayfVMXadkxU6cDO3sFKOwYozQgyoN0WEGwY/edit?ts=5c4b6e8d#heading=h.af3x6jga7lsy

The Firefox Add-ons Manager is going to implement an abuse report feature similar to the existing one on AMO. To that end we need to add some fields to the [current abuse API](https://addons-server.readthedocs.io/en/latest/topics/api/abuse.html) to store additional data that can be sent by Firefox. The PRD specifies we want the following fields (need to sort out how to reconcile this with the current API):

- Client ID hash
- Add-on GUID
- Add-on name
- Add-on summary (description field in manifest)
- Add-on version
- Signature state (curated, non-curated, partner, unsigned, temporary installation)
- Application (Firefox, Firefox for Android)
- Application version
- Application locale
- Operating system
- Operating system version
- Install date
- Abuse report flags (malware, spam, broken search, broken new tab, broken websites, etc.)
- Abuse report text
- Region (tbd - fwiw the current model stores the ip address directly)
- Install origin
- Install method
- Entry point (post-uninstall prompt, options menu, context menu, etc.)

diox:
Those new fields are meant to be used for add-on reporting from within Firefox itself.
Includes a refactor of the abuse API internals to make adding those fields a lot easier.

Fixes #10643

### DIFF
--- a/docs/topics/api/abuse.rst
+++ b/docs/topics/api/abuse.rst
@@ -40,7 +40,7 @@ to if necessary.
     :<json string|null install_date: The add-on install date.
     :<json string|null operating_system: The client's operating system.
     :<json string|null operating_system_version: The client's operating system version.
-    :<json string|null reason: The reason for the report. Can be ``malware``, ``spam_or_advertising``, ``search_takeover``, ``new_tab_takeover``, `breaks_website``, `offensive``, `does_not_match_description`` or ``does_not_work``.
+    :<json string|null reason: The reason for the report. Can be ``malware``, ``spam_or_advertising``, ``search_takeover``, ``new_tab_takeover``, ``breaks_website``, ``offensive``, ``does_not_match_description`` or ``does_not_work``.
     :>json object|null reporter: The user who submitted the report, if authenticated.
     :>json int reporter.id: The id of the user who submitted the report.
     :>json string reporter.name: The name of the user who submitted the report.
--- a/docs/topics/api/abuse.rst
+++ b/docs/topics/api/abuse.rst
@@ -19,7 +19,7 @@ Submitting an add-on abuse report
 The following API endpoint allows an abuse report to be submitted for an Add-on,
 either listed on https://addons.mozilla.org or not.
 Authentication is not required, but is recommended so reports can be responded
-to if nessecary.
+to if necessary.
 
 .. http:post:: /api/v4/abuse/report/addon/
 
@@ -27,6 +27,20 @@ to if nessecary.
 
     :<json string addon: The id, slug, or guid of the add-on to report for abuse (required).
     :<json string message: The body/content of the abuse report (required).
+    :<json string|null addon_install_entry_point: The add-on install entry point. Accepted values: TBD.
+    :<json string|null addon_install_method: The add-on install method. Accepted values: TBD.
+    :<json string|null addon_install_origin: The add-on install origin.
+    :<json string|null addon_name: The add-on name in the locale used by the client.
+    :<json string|null addon_signature: The add-on signature state. Accepted values: TBD.
+    :<json string|null addon_summary: The add-on summary in the locale used by the client.
+    :<json string|null addon_version: The add-on version string.
+    :<json string|null application: The application used by the client. Can be either ``firefox`` or ``android``.
+    :<json string|null application_locale: The locale used by the client for the application.
+    :<json string|null client_id: The client's hashed telemetry ID.
+    :<json string|null install_date: The add-on install date.
+    :<json string|null operating_system: The client's operating system.
+    :<json string|null operating_system_version: The client's operating system version.
+    :<json string|null reason: The reason for the report. Can be ``malware``, ``spam_or_advertising``, ``search_takeover``, ``new_tab_takeover``, `breaks_website``, `offensive``, `does_not_match_description`` or ``does_not_work``.
     :>json object|null reporter: The user who submitted the report, if authenticated.
     :>json int reporter.id: The id of the user who submitted the report.
     :>json string reporter.name: The name of the user who submitted the report.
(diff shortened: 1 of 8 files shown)


docs/topics/api/abuse.rst
```diff
@@ -19,14 +19,28 @@ Submitting an add-on abuse report
...
+    :<json string|null addon_signature: The add-on signature state. Accepted values: TBD.
+    :<json string|null addon_summary: The add-on summary in the locale used by the client.
+    :<json string|null addon_version: The add-on version string.
+    :<json string|null application: The application used by the client. Can be either ``firefox`` or ``android``.
+    :<json string|null application_locale: The locale used by the client for the application.
+    :<json string|null client_id: The client's hashed telemetry ID.
+    :<json string|null install_date: The add-on install date.
+    :<json string|null operating_system: The client's operating system.
+    :<json string|null operating_system_version: The client's operating system version.
+    :<json string|null reason: The reason for the report. Can be ``malware``, ``spam_or_advertising``, ``search_takeover``, ``new_tab_takeover``, `breaks_website``, `offensive``, `does_not_match_description`` or ``does_not_work``.
```
EnTeQuAk: at ``breaks_website`` you missed a "`", same for ``offensive`` and `does_not_match_description``

docs/topics/api/abuse.rst
```diff
@@ -37,6 +51,20 @@ to if nessecary.
     :>json int|null addon.id: The add-on id on AMO. If the guid submitted didn't match a known add-on on AMO, then null.
     :>json string|null addon.slug: The add-on slug. If the guid submitted didn't match a known add-on on AMO, then null.
     :>json string message: The body/content of the abuse report.
+    :>json string|null addon_install_entry_point: The add-on install entry point.
```
EnTeQuAk: These (and below) look duplicated?

src/olympia/abuse/models.py
```diff
@@ -14,6 +14,41 @@
...
+        None: None,
+        1: 'malware',
+        2: 'spam_or_advertising',
+        3: 'search_takeover',
+        4: 'new_tab_takeover',
+        5: 'breaks_websites',
+        6: 'offensive',
+        7: 'does_not_match_description',
+        8: 'does_not_work',
+    }
```
EnTeQuAk: I wonder if we should be using constants for values `None` to `8` but as this may change back and forth with the PRD I think it's fine for now. Maybe keep that in mind?

src/olympia/abuse/models.py
```diff
@@ -33,14 +68,51 @@ class AbuseReport(ModelBase):
         on_delete=models.SET_NULL)
     message = models.TextField()
 
+    # Extra optional fields for more information, giving some context that is
+    # meant to be extracted automatically by the client (i.e. Firefox) and
+    # submitted via the API.
+    client_id = models.CharField(
+        default=None, max_length=64, blank=True, null=True)
+    addon_name = models.CharField(
+        default=None, max_length=255, blank=True, null=True)
```
EnTeQuAk: Are these so long max lengths because we usually don't care? Addon name is currently limited to 50 characters in the devhub form for example.

src/olympia/abuse/models.py
```diff
@@ -33,14 +68,51 @@ class AbuseReport(ModelBase):
...
+        null=True)
+
     class Meta:
         db_table = 'abuse_reports'
 
     def send(self):
         if self.reporter:
             user_name = '%s (%s)' % (self.reporter.name, self.reporter.email)
         else:
-            user_name = 'An anonymous coward'
```
EnTeQuAk: wtf :fearful: (don't look at git blame, it hurts)

src/olympia/abuse/models.py
```diff
@@ -50,6 +122,12 @@ def send(self):
         send_mail(
             six.text_type(self), msg, recipient_list=(settings.ABUSE_EMAIL,))
 
+    def save(self, *args, **kwargs):
+        creation = not self.pk
+        super(AbuseReport, self).save(*args, **kwargs)
+        if creation:
+            self.send()
```
EnTeQuAk: I always wondered if a `post_save` might be cleaner as we already get the information if we are creating or updating and Django can determine that on it's own :thinking:

src/olympia/abuse/serializers.py
```diff
@@ -1,16 +1,79 @@
...
+        if target_name not in data:
+            msg = serializers.Field.default_error_messages['required']
+            raise serializers.ValidationError({
+                target_name: [msg]
+            })
+
+    def to_internal_value(self, data):
+        output = super(BaseAbuseReportSerializer, self).to_internal_value(data)
+        request = self.context['request']
+        output['ip_address'] = request.META.get('REMOTE_ADDR')
```
EnTeQuAk: We should have `ip_address` defined in the serializer so that it gets validated and we don't store unvalidated input.

src/olympia/abuse/tests/test_views.py
```diff
@@ -138,6 +140,102 @@ def test_throttle(self):
...
+        for field in ('message', 'client_id', 'addon_name', 'addon_summary',
+                      'addon_version', 'application_version',
+                      'application_locale', 'operating_system',
+                      'addon_install_origin'):
+            assert getattr(report, field) == data[field], field
+        # More complex comparisons:
+        assert report.addon_signature is None
+        assert report.application == amo.FIREFOX.id
+        assert report.install_date == datetime(2004, 8, 15, 16, 23, 42)
+        assert report.reason == 2  # Spam / Advertising
```
EnTeQuAk: Really should be constants :laughing:

docs/topics/api/abuse.rst
diox: Those are for the output, above is for the input

src/olympia/abuse/models.py
diox: Yeah... none of the choices are final at this point so I didn't want to go too far with the constants yet... Also, we really should be using something like https://github.com/twidi/django-extended-choices

src/olympia/abuse/models.py
diox: The add-on doesn't have to be listed, and it could have been installed from before the limits were put in place in devhub anyway. I didn't have any direction on this so I just picked the standard largish value.

src/olympia/abuse/models.py
diox: Yeah, I always struggle with this choice. Signals can be hard to follow, so I tend to try to override `save` these days...

src/olympia/abuse/serializers.py
diox: The problem is we don't want it to be submitted directly. And we don't want to error on it either. So I think it's the best compromise. It will go away in https://github.com/mozilla/addons-server/issues/10720 anyway.

docs/topics/api/abuse.rst
EnTeQuAk: Eh, indeed… `<` vs `>` - sorry, I didn't see that

src/olympia/abuse/models.py
EnTeQuAk: > django-extended-choices

This sounds great, let's do this at some point. Might cleanup our code a lot

src/olympia/abuse/models.py
EnTeQuAk: Sounds fair :+1:

src/olympia/abuse/serializers.py
EnTeQuAk: Ah, if it goes away (I forgot) that's fine then.

PR #10734, opened 2019-02-19
Drop LWT editing

### LINKED ISSUE #10728: remove edit LWT support from AMO
Now all LWT have been converted to Static Themes we can drop support in the code base for them.

1st step - stop any remaining rejected LWT from being edited so they can't be resubmitted.

eviljeff:
This just disables the editing of LWT.  We can't remove the view (easily) because `Addon.get_dev_url()` is called for both LWT add-ons and non-LWT add-ons so we'll need to link to _something_ while we have LWT showing up in users's devhub pages.  Fixes #10728


src/olympia/devhub/templates/devhub/personas/edit.html
```diff
@@ -70,230 +58,9 @@ <h3 class="author">
...
-            </div>
-          </div>
-        </div>
-      </fieldset>
-      <div class="listing-footer">
-        <button class="button prominent" disabled type="submit">{{ _('Save Changes') }}</button>
-      </div>
-    </form>
+		<p>
+			  <h1 class="error">Lightweight Theme editing has been disabled.</h1>
```
diox: Untranslated string

eviljeff: It's not going to get translated before the tag anyway, but sure.

diox: If it somehow stays in for longer than anticipated it's best to be prepared :)

PR #10745, opened 2019-02-21
Stop storing ip_address in abuse reports - lookup country via geoip instead

### LINKED ISSUE #10720: Stop storing ip_address in abuse reports

diox:
Fixes #10720

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,12 +46,21 @@ RUN apt-get update && apt-get install -y \
         pngcrush \
         # our makefile and ui-tests require uuid to be installed
         uuid \
+        # Use libmaxmind for speedy geoip lookups
+        libmaxminddb0                    \
+        libmaxminddb-dev                 \Z
     && rm -rf /var/lib/apt/lists/*
 
 RUN apt-get update && apt-get -t stretch-backports install -y \
         # For git-based files storage backend
         libgit2-dev \
     && rm -rf /var/lib/apt/lists/*
+
+ADD http://geolite.maxmind.com/download/geoip/database/GeoLite2-Country.mmdb.gz /tmp
+
+RUN mkdir -p /usr/local/share/GeoIP \
+ && gunzip -c /tmp/GeoLite2-Country.mmdb.gz > /usr/local/share/GeoIP/GeoLite2-Country.mmdb \
+ && rm -f /tmp/GeoLite2-Country.mmdb.gz
 
 # Compile required locale
 RUN localedef -i en_US -f UTF-8 en_US.UTF-8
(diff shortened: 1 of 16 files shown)


Dockerfile
```diff
@@ -46,13 +46,22 @@ RUN apt-get update && apt-get install -y \
         pngcrush \
         # our makefile and ui-tests require uuid to be installed
         uuid \
+        # Use libmaxmind for speedy geoip lookups
+        libmaxminddb0                    \
+        libmaxminddb-dev                 \Z
```
EnTeQuAk: Sure that `\Z` is what you wanted?

src/olympia/abuse/management/commands/migrate_abuse_reports_to_country_code.py
```diff
@@ -0,0 +1,19 @@
...
+        if not value:
+            raise CommandError('GeoIP lookups does not appear to be working.')
+
+    def handle(self, *args, **kwargs):
+        self.setup_check()
+
+        qs = AbuseReport.objects.filter(ip_address__isnull=False)
+        for report in qs:
+            value = AbuseReport.lookup_country_code_from_ip(report.ip_address)
+            report.update(country_code=value)
```
EnTeQuAk: This needs logging. I don't know how many abuse reports there are, but if there are many this may take some time?

Dockerfile
diox: No that's a - weird - typo...

src/olympia/abuse/management/commands/migrate_abuse_reports_to_country_code.py
diox: The number if known, just don't want to publicly share it - it might take some time but I think it'll be low enough that it can be done in one go. In any case, it's really just a bonus, there is no real need to do the migration, it's just nice to have.

diox: Added logging in e7b2b9e82ad1c744da7df7933a8bd5131db3b78a, and also stop loading useless fields and made the queryset avoid instances that already had a country code set

PR #10750, opened 2019-02-22
Implement mimetype categorization for easier file-type / preview detection 

### LINKED ISSUE #10685: Don't expose image binary type

EnTeQuAk:
This implements more advanced file-type detection based on `python-magic` thus we don't rely on the files extensions anymore but actually verify their content (header) before serving it to the client and reviewer.

Fixes #10685  (and actually rewrites it's meaning)


docs/topics/api/reviewers.rst
```diff
@@ -131,7 +131,7 @@ This endpoint allows you to browse through the contents of an Add-on version.
     :>json string file.content: Raw content of the requested file.
     :>json string file.selected_file: The selected file, either from the ``file`` parameter or the default (manifest.json, install.rdf or package.json for Add-ons as well as the XML file for search engines).
     :>json array file.entries[]: The complete file-tree of the extracted XPI.
-    :>json boolean|string file.entries[].binary: ``True`` if the file is a binary file (e.g an .exe, dll, java, swf file), ``'image'`` if the file is an image or ``False`` otherwise. If ``False`` or ``'image'`` the file should be presentable to the user.
+    :>json boolean file.entries[].binary: ``True`` if the file is a binary file (e.g an .exe, dll, java, swf file) or ``False`` otherwise. If ``False`` the file should be presentable to the user.
```
diox: Didn't notice that before: should we say `class` and not `java` ? `java` files are text, technically :)

src/olympia/reviewers/serializers.py
```diff
@@ -123,27 +134,28 @@ def _fetch_entries():
...
-        if ext in denied_extensions:
-            return True
+        if entry.type == 'tree':
+            return 'application/octet-stream', 'directory'
 
-        bytes_ = tuple(bytearray(memoryview(blob)[:4]))
+        # Hardcoding the maximum amount of bytes to read here
+        # until https://github.com/ahupp/python-magic/commit/50e8c856
+        # lands in a release and we can read that value from libmagic
+        bytes_ = io.BytesIO(memoryview(blob)).read(1048576)
```
diox: I don't get this. Isn't that value the default ? Is something else altering the max value if we don't set it ?

EnTeQuAk: Updated the comment, the value is the default, correct, but I have to wrap this to not exhaust/read the whole blob into memory again (e.g for large files).

docs/topics/api/reviewers.rst
EnTeQuAk: eh, indeed

EnTeQuAk: This needs updating anyway

EnTeQuAk: Updated in 2c0d38b690

Dockerfile
```diff
@@ -28,6 +28,7 @@ RUN apt-get update && apt-get install -y \
         zlib1g-dev \
         libffi-dev \
         libssl-dev \
+        libmagic-dev \
```
diox: You don't need this in `.travis.yml` as well ?

EnTeQuAk: It's already installed in travis.

PR #10751, opened 2019-02-22
Ignore reviewer lock when auto-approving langpacks

### LINKED ISSUE #10747: Open review page can block auto-approval indefinitely
The Firefox Release Team ran into [this problem recently](https://bugzilla.mozilla.org/show_bug.cgi?id=1528161), where a language pack wasn't being signed due to its review page being opened by a user with review permissions. This blocked release of a beta version of Firefox.

The intermediate state between submission and auto-approval is just an artifact of the way auto-approval was implemented, and we have ways to turn off auto-approval for individual add-ons. So, I see no reason for the review page lock to block auto-approval, other than prevent unexpected behavior if an action is taken on that review page based on its previous state. My suggestion in that case would be to show an error, refresh the page state, and ask the reviewer to try again. Not great, but it's not a common problem, AFAIK.

/cc @diox @wagnerand @kewisch

diox:
Fix #10747

### DIFF
--- a/src/olympia/reviewers/models.py
+++ b/src/olympia/reviewers/models.py
@@ -1062,8 +1062,15 @@ class AutoApprovalSummary(ModelBase):
 
     @classmethod
     def check_is_locked(cls, version):
-        locked = get_reviewing_cache(version.addon.pk)
-        return bool(locked) and locked != settings.TASK_USER_ID
+        # Langpacks are submitted as part of Firefox release process and should
+        # never be blocked by a reviewer having opened the review page - they
+        # should always be auto-approved anyway.
+        is_langpack = version.addon.type == amo.ADDON_LPAPP
+        locked_by = get_reviewing_cache(version.addon.pk)
+        return (
+            not is_langpack and
+            bool(locked_by) and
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/models.py
```diff
@@ -1062,8 +1062,15 @@ def check_uses_native_messaging(cls, version):
...
-        locked = get_reviewing_cache(version.addon.pk)
-        return bool(locked) and locked != settings.TASK_USER_ID
+        # Langpacks are submitted as part of Firefox release process and should
+        # never be blocked by a reviewer having opened the review page - they
+        # should always be auto-approved anyway.
+        is_langpack = version.addon.type == amo.ADDON_LPAPP
+        locked_by = get_reviewing_cache(version.addon.pk)
+        return (
+            not is_langpack and
+            bool(locked_by) and
```
eviljeff: nit: isn't `locked_by` always an int (or null)?  Why do we need to force it to `bool`?

diox: It's to force the method to always return a boolean. Not stricly necessary but makes it more predictable / easier to use & test.

PR #10773, opened 2019-02-24
Update django-multidb-router terminology

### LINKED ISSUE #10763: Update django-multidb-router terminology
django-multidb-router updated it's terminology in 0.9, changing from the terms "master" and "slave" to "primary" and "replica".  The existing terms [largely](https://github.com/mozilla/addons-server/pull/10762) still work in 0.9 but are deprecated and will be removed in a future version.  We should change the deprecated terms in advance.

https://github.com/jbalogh/django-multidb-router/blob/master/docs/index.rst#upgrading

championshuttler:
Fixes #10763


src/olympia/conf/prod/settings.py
```diff
@@ -40,7 +40,7 @@
 
 SERVICES_DATABASE = get_db_config('SERVICES_DATABASE_URL')
 
-SLAVE_DATABASES = ['slave']
+REPLICA_DATABASES = ['slave']
```
championshuttler: @eviljeff do we also need to change `slave` to replica?

src/olympia/conf/stage/settings.py
```diff
@@ -52,7 +52,7 @@
 
 SERVICES_DATABASE = get_db_config('SERVICES_DATABASE_URL')
 
-SLAVE_DATABASES = ['slave']
+REPLICA_DATABASES = ['slave']
```
championshuttler: Same here.

src/olympia/conf/dev/settings.py
```diff
@@ -53,7 +53,7 @@
 
 SERVICES_DATABASE = get_db_config('SERVICES_DATABASE_URL')
 
-SLAVE_DATABASES = ['slave']
+REPLICA_DATABASES = ['slave']
```
championshuttler: @eviljeff do we also need to change `slave` to replica?

championshuttler:
Few instances `get_slave` of are already changed here https://github.com/mozilla/addons-server/pull/10762/files#diff-75582f759d2e917a1f6cae5848afd139R507

eviljeff:
You started this too early 😄  - the patch to upgrade to 0.9 wasn't merged so the tests are being run on 0.8 which still uses the old terms.  Please either `git fetch` `git rebase`, or `git pull`.

src/olympia/conf/dev/settings.py
eviljeff: it's not needed to support newer versions of django-multi-db, (`slave` refers to the key in a setting we define) but it would be more consistent.  If you do you need to change the dict key a few lines above (line 50 in this file) also to `'replica'`

src/olympia/conf/prod/settings.py
eviljeff: same as in /dev/settings.py

src/olympia/conf/stage/settings.py
eviljeff: same as in /dev/settings.py

src/olympia/lib/settings_base.py
```diff
@@ -1419,9 +1419,9 @@ def read_only_mode(env):
     env['READ_ONLY'] = True
 
     # Replace the default (master) db with a slave connection.
-    if not env.get('SLAVE_DATABASES'):
+    if not env.get('REPLICA_DATABASES'):
         raise Exception('We need at least one slave database.')
-    slave = env['SLAVE_DATABASES'][0]
+    slave = env['REPLICA_DATABASES'][0]
```
eviljeff: note for later: we will have to co-ordinate with ops to update the env variables too.

src/olympia/conf/dev/settings.py
```diff
@@ -47,13 +47,13 @@
 
 DATABASES = {
     'default': get_db_config('DATABASES_DEFAULT_URL', charset='utf8mb4'),
-    'slave': get_db_config(
+    'replica': get_db_config(
         'DATABASES_SLAVE_URL', atomic_requests=False, charset='utf8mb4'),
```
championshuttler: I guess we should change the DATABASES_SLAVE_URL` to `DATABASES_REPLICA_URL` to keep consistency @eviljeff Let me know if we need to it as well. Thanks.

eviljeff: Sure.  It's another external env variable so we'll need to ops to update those on their end too.

src/olympia/conf/prod/settings.py
```diff
@@ -35,14 +35,14 @@
 
 DATABASES = {
     'default': get_db_config('DATABASES_DEFAULT_URL', charset='utf8mb4'),
-    'slave': get_db_config(
-        'DATABASES_SLAVE_URL', atomic_requests=False, charset='utf8mb4',
+    'replica': get_db_config(
+        'DATABASES_REPLICA_URL', atomic_requests=False, charset='utf8mb4',
```
championshuttler: note for later: we will have to co-ordinate with ops to update the env variables too. (Copying eviljeff comment from below.)

eviljeff:
Needs https://bugzilla.mozilla.org/show_bug.cgi?id=1530760 to be completed first before merging

PR #10786, opened 2019-02-25
Tweak abuse reporting API fields to match existing API and PRD updates

### LINKED ISSUE #10753: Update Abuse Report API with choices from the PRD
The Abuse Report API from #10643 now has documented choices for the fields we care about in [the PRD](https://docs.google.com/document/d/1K5Th9UmpayfVMXadkxU6cDO3sFKOwYozQgyoN0WEGwY/edit?ts=5c4b6e8d#), we should implement that.

Use the occasion to use `extended-choices` to make the constants a little more easy to deal with, and don't forget documentation, including documenting `application_version` parameter that has been forgotten in #10643.

diox:
We now have the expected values for some of the fields. In addition, some of the fields that had equivalents in our other APIs are renamed to match the rest of the API better.

Fixes #10753


diox:
Depends on https://github.com/twidi/django-extended-choices/pull/48 + a new release of django-extended-choices integrating this change, and removal of `future` from our dependencies.

docs/topics/api/abuse.rst
```diff
@@ -58,8 +62,9 @@ to if necessary.
     :>json string|null addon_signature: The add-on signature state.
     :>json string|null addon_summary: The add-on summary in the locale used by the client.
     :>json string|null addon_version: The add-on version string.
-    :>json string|null application: The application used by the client.
-    :>json string|null application_locale: The locale used by the client for the application.
+    :>json string|null app: The application used by the client.
+    :>json string|null appversion: The locale used by the client for the application.
+    :<json string|null lang: The language code of the locale used by the client for the application.
     :>json string|null client_id: The client's hashed telemetry ID.
```
eviljeff: I know this isn't a change in this PR, but it's pretty weird to specify all the properties twice because you get back what you submitted (and why do we do that anyway?)

src/olympia/abuse/tests/test_views.py
```diff
@@ -147,9 +147,6 @@ def test_throttle(self):
         assert response.status_code == 429
 
     def test_optional_fields(self):
-        # FIXME: when choices for addon_signature, addon_install_method and
-        # addon_install_entry_point are finalized, update this test. In the
-        # meantime we pass None.
```
eviljeff: it didn't need to be fixed?

docs/topics/api/abuse.rst
diox: Technically the values could differ if we do some post-save treatment on them, or there could be extra stuff returned (which is the case here: `addon` and `reporter` keys are added). It could be more elegantly described in the docs though.

src/olympia/abuse/tests/test_views.py
diox: that's what this PR is about but I've added tests for the choices elsewhere (in `test_models` and `test_serializers`)

PR #10797, opened 2019-02-26
Expose more specific mimetype.

### LINKED ISSUE #10796: Correct mimetype for text/plain files
The browse api returns currently `text/plain` for all text files but we should make the mimetype more specific for readable text files and allow them to be classified as `text/css`, `text/javascript` etc

EnTeQuAk:
The browse api returns currently `text/plain` for all text files but we should make the mimetype more specific for readable text files and allow them to be classified as `text/css`, `text/javascript` etc

Fixes #10796

### DIFF
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -99,7 +99,6 @@ class FileEntriesSerializer(FileSerializer):
                 path = force_text(entry_wrapper.path)
                 blob = entry_wrapper.blob
 
-                mime, encoding = mimetypes.guess_type(entry.name)
                 sha_hash = (
                     get_sha256(io.BytesIO(memoryview(blob)))
                     if not entry.type == 'tree' else '')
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/serializers.py
```diff
@@ -157,6 +156,13 @@ def get_entry_mime_type(self, entry, blob):
...
         known_types = ('image', 'text')
 
+        if mime_type == 'text':
+            # Allow text mimetypes to be more specific for readable
+            # files. `python-magic`/`libmagic` usually just returns
+            # plain/text but we should use actual types like text/css or
+            # text/javascript.
+            mime, _ = mimetypes.guess_type(entry.name)
+
         return mime, 'binary' if mime_type not in known_types else mime_type
```
diox: FWIW I found this one-liner confusing to read - the inline `if` applies to the second part but that wasn't immediately obvious :)

EnTeQuAk: I'll keep that in mind when touching it the next time.

PR #10806, opened 2019-02-27
Add django admin for versions

### LINKED ISSUE #9304: Link to the Version file from /admin/models/files/ triggers a 504
Follow up for #8985 

STR:
1. Log in to AMO with an Admin account
2. Go to the Admin pages and open the File management tool - `/admin/models/files/`
3. Open any File details  - [see this example](https://addons-internal.stage.mozaws.net/en-US/admin/models/files/file/325883/change/?_changelist_filters=q%3Ddevhubrelease2308)
4. Click on the link next to the Version field

Actual result:
The link is triggering a 504 when opened

Expected result:
The link should open `https://addons-internal.stage.mozaws.net/en-US/admin/models/versions/version/<number>/change/`

Notes:
- reproduced on AMO stage with FF61, Win10x64

image

### LINKED ISSUE #10775: Django model admin for versions returns 504

Django model admin pages for versions seems to return 504 always. Example. https://addons-internal.prod.mozaws.net/en-US/admin/models/versions/version/1988144/change/

### What happened?

HTTP 504: Gateway Timeout

Display the versions admin page.

wagnerand:
Fix #10775.
Fix #9304 (probably).

### DIFF
--- a/src/olympia/versions/admin.py
+++ b/src/olympia/versions/admin.py
@@ -14,7 +14,7 @@ class VersionAdmin(admin.ModelAdmin):
     view_on_site = False
     readonly_fields = ('version', 'channel')
 
-    raw_id_fields = ('addon', 'license', 'release_notes')
+    raw_id_fields = ('addon', 'license')
 
     fieldsets = (
         (None, {
--- a/src/olympia/versions/admin.py
+++ b/src/olympia/versions/admin.py
@@ -1,3 +1,4 @@
+from django import forms
 from django.contrib import admin
 
 from .models import License, Version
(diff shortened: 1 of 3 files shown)


src/olympia/versions/admin.py
```diff
@@ -9,5 +10,30 @@ class LicenseAdmin(admin.ModelAdmin):
     ordering = ('builtin',)
 
 
+class VersionAdmin(admin.ModelAdmin):
+    view_on_site = False
+    readonly_fields = ('version', 'channel')
+
+    raw_id_fields = ('addon', 'license', 'release_notes')
```
diox: You shouldn't need to add `release_notes` to the raw id fields ? It's a translated field, it should be handled already.

wagnerand: Done in 2b74fb06c3.

wagnerand:
@EnTeQuAk since I am changing permissions, would you mind giving this another quick look? Thanks!

src/olympia/versions/tests/test_admin.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def test_detail_view_has_download_link(self):
+        addon = addon_factory()
+        version = addon.current_version
+        detail_url = reverse(
+            'admin:versions_version_change', args=(version.pk,)
+        )
+        user = user_factory()
+        self.grant_permission(user, 'Admin:Tools')
+        self.grant_permission(user, 'Admin:Advanced')
```
EnTeQuAk: Can we have a separate call before the actual "200 response" to double-check that just doing `self.client.get(detail_url...)` without `Admin:Advanced` leads to an error response? (Should be 403)

wagnerand: Done in f6b3e29559.

PR #10829, opened 2019-03-01
frontend has the user admin edit url in a slightly different location

### LINKED ISSUE #10828: 'users.admin_edit' url is inconsistent with addons-frontend
`'users.admin_edit'` should be `/user/<user-id>/edit` - it was `/users/edit/<user-id>` on legacy frontend.  It's only referenced in reviewer tools and admin so it seems safe to just update it to the current correct url

eviljeff:
fixes #10828

### DIFF
--- a/src/olympia/users/urls.py
+++ b/src/olympia/users/urls.py
@@ -12,6 +12,7 @@ detail_patterns = [
     url(r'^themes(?:/(?P<category>[^ /]+))?$', views.themes,
         name='users.themes'),
     url(r'^abuse', views.report_abuse, name='users.abuse'),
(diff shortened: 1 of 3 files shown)


eviljeff:
Tests what exactly?  There's no great reason why whoever wrong the django code decided to have the url one way and the whoever wrote the react code decided to have the url the other way.

EnTeQuAk:
> Tests  what exactly?  There's no great reason why whoever wrong the django  code decided to have the url one way and the whoever wrote the react  code decided to have the url the other way.

Indeed, lgtm anyway.

Issue #10849, opened 2019-03-05
Compare-API: Add support for renames and copies


EnTeQuAk:
Git properly supports finding copies and renames.

Let's implement finding them and expose them properly.

```python
        opts = pygit2.GIT_DIFF_FIND_RENAMES | pygit2.GIT_DIFF_FIND_COPIES
        diff.find_similar(opts, SIMILARITY_THRESHOLD, SIMILARITY_THRESHOLD)
```

PR #10850, opened 2019-03-05
Implement low-level git diff functionality.

### LINKED ISSUE #10848: Implement low-level diff functionality
Part of #10323 

* Implement low-level git diff functionality
* Make sure initial commits are covered
* Add ability to find renames or copies

EnTeQuAk:
This implements all the necessary changes to diff two versions against
each other and also handles initial commits.

Fixes #10848

### DIFF
--- a/src/olympia/lib/git.py
+++ b/src/olympia/lib/git.py
@@ -366,6 +366,8 @@ class AddonGitRepository(object):
 
         This doesn't contain the ``EXTRACTED_PREFIX`` prefix folder.
         """
+        # When `commit` is a commit hash, e.g passed to us through the API
+        # serializers we have to fetch the actual commit object to proceed.
         if isinstance(commit, six.string_types):
             commit = self.git_repository[commit]
 
--- a/src/olympia/lib/git.py
+++ b/src/olympia/lib/git.py
@@ -395,10 +395,13 @@ class AddonGitRepository(object):
                     path=tree_entry.name)
 
     def get_diff(self, commit, parent=None):
-        """Get a diff from `commit` to `parent`.
+        """Get a diff from `parent` to `commit`.
 
         If `parent` is not given we assume it's the first commit and handle
         it accordingly.
+
+        We are resolving renames and copies according to a 50% similarity
+        threshold.
         """
         if parent is None:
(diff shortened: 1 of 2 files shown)


src/olympia/lib/git.py
```diff
@@ -370,3 +393,91 @@ def iter_tree(self, tree):
...
+            context_lines=sys.maxsize,
+            interhunk_lines=0)
+
+        opts = pygit2.GIT_DIFF_FIND_RENAMES | pygit2.GIT_DIFF_FIND_COPIES
+        diff.find_similar(opts, SIMILARITY_THRESHOLD, SIMILARITY_THRESHOLD)
+
+        # TODO (cgrebs): double-check if we are indeed processing files
+        # multiple times (e.g on renames, copies etc)
+        # I think this happens when renames are < similarity threshold or so
+        # needs some more tests though...
```
eviljeff: Issue for follow-up?

src/olympia/lib/git.py
```diff
@@ -346,6 +366,9 @@ def get_root_tree(self, commit):
 
         This doesn't contain the ``EXTRACTED_PREFIX`` prefix folder.
         """
+        if isinstance(commit, six.string_types):
+            commit = self.git_repository[commit]
```
eviljeff: When does this happen?

EnTeQuAk:
> Is there a test that checks SIMILARITY_THRESHOLD and file moves? (Might have missed it - some of the tests are pretty long)

Sorry, I forgot to mention this. I'll add this in a separate PR via https://github.com/mozilla/addons-server/issues/10849 since renames and copies will involve even more test-helper code and I wanted this to be easier to review here.

src/olympia/lib/git.py
EnTeQuAk: Added a comment in 58e3ce9ba2, it basically happens in the drf api serializers which simply grab `version.git_hash` and pass that on as `commit`

When that function is used internally (to hide the `extracted/` folder implementation detail) we are passing along actual commit objects.

src/olympia/lib/git.py
EnTeQuAk: I'll do that as part of #10849

PR #10874, opened 2019-03-06
Implement compare API.

### LINKED ISSUE #10323: Implement files/compare API
Follow-up from #10322 which should implement the API related to comparing two different versions of the same add-on.

EnTeQuAk:
This implements fetching a diff by pathspec, we need to implement this
upstream so it's implemented as a workaround for now.

This removes support for copy and rename finding for now. I filed #10849
to fix this properly in a separate step.

The actual diff output format is hard-coded and not yet implemented as
serializers for now which makes experimenting much easier in case we
need to adapt the format depending on how the client implements it.

Fixes #10323

### DIFF
--- a/docs/topics/api/reviewers.rst
+++ b/docs/topics/api/reviewers.rst
@@ -122,6 +122,8 @@ This endpoint allows you to browse through the contents of an Add-on version.
 
     Inherits most properties from :ref:`version detail <version-detail-object>` except ``files``.
 
+    .. _reviewers-versions-browse-detail:
+
     :param file: The specific file in the XPI to retrieve. Defaults to manifest.json, install.rdf or package.json for Add-ons as well as the XML file for search engines.
     :>json string validation_url_json: The url to the addons-linter validation report, rendered as JSON.
     :>json string validation_url: The url to the addons-linter validation report, rendered as HTML.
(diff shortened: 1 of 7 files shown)


EnTeQuAk:
I'm trying to figure out that travis failure - https://travis-ci.org/mozilla/addons-server/jobs/502683722#L2981-L2986 - which works perfectly fine for me locally.

Should be ready for a first look though.

src/olympia/reviewers/tests/test_views.py
```diff
@@ -5289,6 +5233,70 @@ def test_unlisted_version_user_but_not_author(self):
...
+    client_class = APITestClient
+    __test__ = True
+
+    def setUp(self):
+        super(TestReviewAddonVersionViewSetDetail, self).setUp()
+
+        # TODO: Most of the initial setup could be moved to
+        # setUpTestData but unfortunately paths are setup in pytest via a
+        # regular autouse fixture that has function-scope so functions in
+        # setUpTestData don't use proper paths (cgrebs)
```
diox:
```suggestion
        # setUpTestData doesn't use proper paths (cgrebs)
```

docs/topics/api/reviewers.rst
```diff
@@ -139,3 +141,116 @@ This endpoint allows you to browse through the contents of an Add-on version.
...
+
+This endpoint allows you to compare two Add-on versions with each other.
+
+    .. note::
+        Requires authentication and the current user to have ``ReviewerTools:View``
+        permission for listed add-ons as well as ``Addons:ReviewUnlisted`` for
+        unlisted add-ons. Additionally the current user can also be the owner
+        of the add-on.
+
+.. http:get:: /api/v4/reviewers/addon/(int:addon_id)/versions/(int:version_id)/compare/(int:version_id)/
```
willdurand: This URL is not exposed actually. It should be `compare_to` and  not `compare`.

EnTeQuAk: Indeed, fixed in 7a84c2597f

PR #10888, opened 2019-03-08
Fix ui tests for new nginx config

jrbenny35:
I don't think we have an issue but the UI tests have been failing for a while now due to the nginx config and some other problems. This should fix them. 

Everything routes through nginx and ```http://olympia.test```. These can also been run locally outside of the docker container. I will add a PR with instructions later.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,6 +145,7 @@ jobs:
             # Make sure everything is started
             docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml up -d
             sleep 10
+            docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml sudo chown -R seluser:seluser .
             # Start Test in Firefox docker container
             docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml exec selenium-firefox tox -e ui-tests
       - store_artifacts:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,22 +115,27 @@ jobs:
           command: |
             set -x
             sudo sysctl -w vm.max_map_count=262144
+            # Install UUID
+            sudo apt-get update -qqy && sudo apt-get -qqy install uuid
+            # Pull images
             docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml pull --quiet
+            #Set Fxa Username
+            export UITEST_FXA_EMAIL=uitest-$(uuid)@restmail.net
+            # Start Containers
             docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml up -d
             sleep 20
-            docker-compose ps
-            scripts/ui-test-hostname-setup.sh
+            docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml ps
             # Make sure dependencies get updated in worker and web container
-            docker-compose exec worker make -f Makefile-docker update_deps update_assets
-            docker-compose restart worker
-            docker-compose exec web make -f Makefile-docker update_deps update_assets
-            docker-compose restart web
-            # Delete any existing uitest users within restmail. TEMP FIX: https://github.com/mozilla/addons-server/issues/8980
-            wget --method=DELETE http://restmail.net/mail/uitest
+            docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml exec worker make -f Makefile-docker update_deps update_assets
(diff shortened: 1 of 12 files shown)


tests/ui/test_devhub.py
```diff
@@ -52,7 +52,7 @@ def test_devhub_logout(base_url, selenium, devhub_login):
 @pytest.mark.nondestructive
 def test_devhub_register(base_url, selenium):
     """Test register link loads register page."""
-    selenium.get('{}/en-US/developers/'.format(base_url))
+    selenium.get('{}/developers'.format(base_url))
```
eviljeff: /developers should still work with and without the locale prefix (without the prefix will 302 to with the prefix)

.circleci/config.yml
```diff
@@ -112,25 +112,32 @@ jobs:
...
+            # Install UUID
+            sudo apt-get update -qqy && sudo apt-get -qqy install uuid
+            export UITEST_FXA_EMAIL=uitest-$(uuid)@restmail.net
+            # Pull images
             docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml pull --quiet
+            # Start Containers
             docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml up -d
             sleep 20
-            docker-compose ps
-            scripts/ui-test-hostname-setup.sh
```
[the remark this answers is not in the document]
jrbenny35: Yup!

Makefile-docker
```diff
@@ -7,8 +7,8 @@ export PIP_COMMAND=pip${PYTHON_VERSION_MAJOR}
 export PYTHON_COMMAND=python${PYTHON_VERSION_MAJOR}
 
 # FXA username and password for ui-tests
-export FXA_EMAIL=uitest-$(shell uuid)@restmail.net
-export FXA_PASSWORD=uitester
+export FXA_EMAIL=${UITEST_FXA_EMAIL}
+export FXA_PASSWORD=${UITEST_FXA_PASSWORD}
```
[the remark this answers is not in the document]
jrbenny35: I could remove this as well.

jrbenny35:
Getting random ```Running all indexation tasks . . .```. Using ssh there doesn't seem to be a problem or there aren't any logs that point to what is actually going on.

Makefile-docker
```diff
@@ -172,24 +172,31 @@ setup-ui-tests:
...
-.PHONY: run-ui-tests
-run-ui-tests: setup-ui-tests
-	# Generate test add-ons and force a reindex to make sure things are updated
-	$(PIP_COMMAND) install --progress-bar=off --no-deps -r requirements/uitests.txt
+	$(PYTHON_COMMAND) manage.py generate_addons --app firefox $(NUM_ADDONS)
+	$(PYTHON_COMMAND) manage.py generate_addons --app android $(NUM_ADDONS)
+	$(PYTHON_COMMAND) manage.py generate_themes $(NUM_THEMES)
 	$(PYTHON_COMMAND) manage.py generate_default_addons_for_frontend
+
+	# Now that addons have been generated, reindex.
```
diox: You shouldn't need a `--wipe` the second time around.

docker-compose.yml
```diff
@@ -39,6 +41,8 @@ services:
       - ./site-static:/srv/site-static
       - ./storage/shared_storage/uploads:/srv/user-media
       - ./storage/files:/srv/user-media/addons
+    ports:
+      - "80:80"
```
diox: why do you need to expose ports on the worker?

.circleci/config.yml
```diff
@@ -110,11 +110,16 @@ jobs:
...
+          name: Set hosts
+          command: |
+            echo 127.0.0.1 olympia.test | sudo tee -a /etc/hosts
+            cat /etc/hosts
       - run:
           name: Start container, verify it's running and start tests
           environment:
-            UITEST_FXA_EMAIL: uitest-$(uuid)@restmail.net
             MOZ_HEADLESS: 1
+            PYTEST_ADDOPTS: "--reruns 2"
```
eviljeff: Is this because there are intermittents, or do some tests fail the first time?

jrbenny35: 2 main reasons. 
1. Sometimes there is just a bit of slowness and it effects some tests more than other, i.e.: the devhub upload test. It sometimes times out because the addon doesn't get uploaded.

2. Geckodriver sometimes cannot start marionette on a port and we get an error that marionette couldn't find an open port. 

Note these are ONLY for CI. A local test run does not have reruns.

PR #10905, opened 2019-03-11
set MIGRATED_LWT_UPDATES_ENABLED=True

### LINKED ISSUE #9941: Re-enable LWT to static themes updates for Android
Once [bug 1429488](https://bugzilla.mozilla.org/show_bug.cgi?id=1429488) lands, the current level of theme support for LWTs in Android (header image and accent colour) will work with static themes, too.

This means that it should then be safe to serve updates to migrated LWTs to Android users as well, and also allow normal access to static themes (although ideally we'd suppress themes without a header image for Android users, because those will remain somewhat pointless for now - no idea whether that's easily possible, though).

eviljeff:
fixes #9941

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1822,7 +1822,7 @@ AWS_STATS_S3_PREFIX = env('AWS_STATS_S3_PREFIX', default='amo_stats')
 
 MIGRATED_LWT_DEFAULT_OWNER_EMAIL = 'addons-team+landfill-account@mozilla.com'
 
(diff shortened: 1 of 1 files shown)


eviljeff:
We have a test with the setting set to `True`: https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/tests/test_theme_update.py#L59 and a test with the setting set to `False`: https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/tests/test_theme_update.py#L102

I'm not clear on what additional tests we could have.

PR #10926, opened 2019-03-13
rm legacy search views

### LINKED ISSUE #10918: remove legacy search views
We want to start removing more parts of the legacy frontend so search can go next.  The search templates (the search box) is only still exposed on the statistics pages (so it's being removed)

eviljeff:
fixes #10918


src/olympia/search/tests/test_views.py
```diff
@@ -1,1048 +0,0 @@
...
-        response = self.client.get(self.url, {'q': 'Addon'})
-        assert addon.pk in self.get_results(response)
-
-        self.make_addon_unlisted(addon)
-        addon.reload()
-        self.refresh()
-        response = self.client.get(self.url, {'q': 'Addon'})
-        assert addon.pk not in self.get_results(response)
-
-    def test_find_addon_default_non_en_us(self):
```
diox: This test might be useful to keep and move elsewhere. I don't think we have an equivalent in the API tests

eviljeff: I think this is covered in https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/tests/test_views.py#L2937:L2973 (with less comments)

PR #10970, opened 2019-03-20
Remove all traces of jetpack-specific code

### LINKED ISSUE #7189: Remove "jetpack" tags from existing add-ons and associated code

upload a sdk addon and amo always adds the keyword "jetpack" automatically, which already sucks ...

now i ported my first extension to webext , since it still has the @jetpack in its id, amo will still add jetpack to the keywords .

i suggest removing that feature completely also for older addons

thanx for listening :)

diox:
Fixes #7189


src/olympia/files/admin.py
```diff
@@ -35,14 +35,14 @@ class FileAdmin(admin.ModelAdmin):
...
         ('Details', {
             'fields': (
-                'jetpack_version', 'cert_serial_num', 'original_status'),
+                'cert_serial_num', 'original_status'),
         }),
         ('Flags', {
             'fields': (
                 'is_restart_required', 'strict_compatibility',
-                'requires_chrome', 'binary', 'binary_components',
-                'is_signed', 'is_multi_package', 'is_experiment',
```
eviljeff: A follow-up issue at least, but is `is_multi_package` still used?  multiple packaged addons were a legacy only thing, iirc.

diox: You're right it's not, I'll file an issue and take care of it.

diox: https://github.com/mozilla/addons-server/issues/10978

PR #10990, opened 2019-03-22
Stop building Python 2 images

### LINKED ISSUE #10986: Stop building python 2 images
The opposite of https://github.com/mozilla/addons-server/issues/10385 : now that we're running Python 3 in all environements are are not likely to switch back, let's stop building Python 3 images.

diox:
Fixes #10986


.circleci/config.yml
```diff
@@ -28,41 +28,23 @@ commands:
...
-  build-py2:
+  build:
     machine: true
     working_directory: ~/addons-server
     steps:
       - checkout
       - build_and_push_container_image:
           image_tag: latest
-          dockerfile: "Dockerfile.python2.deploy"
+          dockerfile: "Dockerfile.deploy"
```
diox: This reverts us back to having a single `Dockerfile.deploy` that builds an image called `latest` (and below `"${CIRCLE_TAG}"` for tags). So this will need ops to change the jenkins pipeline to match.

docker-compose.yml
```diff
@@ -22,7 +22,7 @@ x-env-mapping: &env
 services:
   worker: &worker
     <<: *env
-    image: addons/addons-server:latest-py3
+    image: addons/addons-server:latest
```
diox: This requires us to push the python3 image to `addons/addons-server:latest`.

diox: @EnTeQuAk once I merge, can you take care of that bit please ? I'm still operating with limited quota.

EnTeQuAk: absolutely! Just ping me in IRC and I'll rebuild and upload most recent images.

PR #10991, opened 2019-03-22
Truncate data before saving akismet records

### LINKED ISSUE #10212: Add-on submission fails with a server error if the name is too long when urlencoded
Also reported some way back in https://bugzilla.mozilla.org/show_bug.cgi?id=1193265

**STR:**

1. Upload a new add-on with non-ascii characters in name and summary - see attached extension
2. Try to complete the submission

**Actual result:**
A server error is received when the add-on is submitted

**Expected result:**
Add-on is submitted with no errors

**Notes:**
- in my case, I was using [Amharic](https://en.wikipedia.org/wiki/Amharic)
- reproduced on all AMO server with FF63, WIn10x64
- could probably have the same root cause as  #8923  

[Unicode add-on.zip](https://github.com/mozilla/addons-server/files/2675489/Unicode.add-on.zip)

diox:
This should fix issues at submission time with add-ons that have a particularly complex slug that causes the submission URL, which we record as the referrer, to be > 255 chars.

@eviljeff do you think it's the right approach ? I figured we didn't care *that* much if those fields were truncated, and setting a limit to avoid recording super long date makes sense to me.

Fixes #10212


eviljeff:
> @eviljeff do you think it's the right approach ? I figured we didn't care that much if those fields were truncated, and setting a limit to avoid recording super long date makes sense to me.

I suppose it's okay.  We're potentially reducing the accuracy of Akismet if the long data is associated with other spammy submissions around the web (for `user_agent` anyway) but it's an edge case.

PR #11016, opened 2019-03-26
Add new code-manager FxA configs

### LINKED ISSUE #11015: Add a new "code-manager" FxA configuration in settings.py
In order to allow authentication in https://github.com/mozilla/addons-code-manager, we need to add a new FxA configuration (`FXA_CONFIG`) named `code-manager` in the django settings.

willdurand:
Fixes #11015

---

I adapted the `amo` configs in each env. I don't really understand why we specify `/fxa-authenticate` in `redirect_url` though. I believe this is never used (FxA config has its own "redirect URI" that points to the correct API endpoint).

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -83,7 +83,7 @@ FXA_CONFIG = {
         'content_host': 'https://stable.dev.lcip.org',
         'oauth_host': 'https://oauth-stable.dev.lcip.org/v1',
         'profile_host': 'https://stable.dev.lcip.org/profile/v1',
-        'redirect_url': 'http://localhost:3000/fxa-authenticate',
+        'redirect_url': 'http://olympia.test/api/v3/accounts/authenticate/',
         'scope': 'profile',
     },
     'local': {
@@ -94,16 +94,18 @@ FXA_CONFIG = {
         'content_host': 'https://stable.dev.lcip.org',
         'oauth_host': 'https://oauth-stable.dev.lcip.org/v1',
         'profile_host': 'https://stable.dev.lcip.org/profile/v1',
-        'redirect_url': 'http://localhost:3000/api/v3/accounts/authenticate/',
+        'redirect_url': 'http://olympia.test/api/v3/accounts/authenticate/?config=local', # noqa
         'scope': 'profile',
     },
     'code-manager': {
-        'client_id': env('CODE_MANAGER_FXA_CLIENT_ID', default='CHANGE_ME'),
-        'client_secret': env('CODE_MANAGER_FXA_CLIENT_SECRET', default='CHANGE_ME'), # noqa
+        'client_id': env('CODE_MANAGER_FXA_CLIENT_ID', default='a98b671fdd3dfcea'), # noqa
+        'client_secret': env(
+            'CODE_MANAGER_FXA_CLIENT_SECRET',
+            default='d9934865e34bed4739a2dc60046a90d09a5d8336cf92809992dec74a4cff4665'),  # noqa
         'content_host': 'https://stable.dev.lcip.org',
         'oauth_host': 'https://oauth-stable.dev.lcip.org/v1',
         'profile_host': 'https://stable.dev.lcip.org/profile/v1',
-        'redirect_url': 'http://localhost:3000/fxa-authenticate',
+        'redirect_url': 'http://olympia.test/api/v4/accounts/authenticate/?config=code-manager', # noqa
         'scope': 'profile',
     },
 }
--- a/settings.py
+++ b/settings.py
@@ -83,7 +83,7 @@ FXA_CONFIG = {
         'content_host': 'https://stable.dev.lcip.org',
         'oauth_host': 'https://oauth-stable.dev.lcip.org/v1',
         'profile_host': 'https://stable.dev.lcip.org/profile/v1',
-        'redirect_url': 'http://olympia.test/api/v3/accounts/authenticate/',
+        'redirect_url': 'http://localhost:3000/fxa-authenticate',
         'scope': 'profile',
     },
     'local': {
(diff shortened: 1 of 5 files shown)


settings.py
```diff
@@ -97,8 +97,17 @@
         'redirect_url': 'http://localhost:3000/api/v3/accounts/authenticate/',
         'scope': 'profile',
     },
+    'code-manager': {
+        'client_id': env('CODE_MANAGER_FXA_CLIENT_ID', default='CHANGE_ME'),
+        'client_secret': env('CODE_MANAGER_FXA_CLIENT_SECRET', default='CHANGE_ME'), # noqa
```
eviljeff: can you use the dashboard to create some credentials (like the others do)

src/olympia/conf/prod/settings.py
```diff
@@ -101,9 +101,20 @@
...
+        'client_id': env('CODE_MANAGER_FXA_CLIENT_ID'),
+        'client_secret': env('CODE_MANAGER_FXA_CLIENT_SECRET'),
+        'content_host': 'https://accounts.firefox.com',
+        'oauth_host': 'https://oauth.accounts.firefox.com/v1',
+        'profile_host': 'https://profile.accounts.firefox.com/v1',
+        'redirect_url':
+            'https://addons.mozilla.org/api/v4/accounts/authenticate/',
+        'scope': 'profile',
+        'skip_register_redirect': True,
+    },
```
eviljeff: Are prod and stage ready to be enabled yet?

willdurand: no, not yet but the config can still be added, no?

settings.py
willdurand: yep

willdurand:
I also updated the `redirect_url` params with better values. For local dev and -dev settings, I read the values using the FxA console.

settings.py
```diff
@@ -83,7 +83,7 @@
         'content_host': 'https://stable.dev.lcip.org',
         'oauth_host': 'https://oauth-stable.dev.lcip.org/v1',
         'profile_host': 'https://stable.dev.lcip.org/profile/v1',
-        'redirect_url': 'http://localhost:3000/fxa-authenticate',
+        'redirect_url': 'http://olympia.test/api/v3/accounts/authenticate/',
```
kumar303: As I recall, `redirect_url` has no effect whatsoever. The value configured at https://oauth-stable.dev.lcip.org/console/ will override it.

eviljeff: We like to keep them in sync for reference for reference though

willdurand: yep, but I prefer to use the values that should be there. That will avoid confusion in the future I believe. On a related note, FxA should really take `redirect_url` into account (and check its value against the redirect URI it knows).

eviljeff: You shouldn't change this without changing https://oauth-stable.dev.lcip.org/console/client/0f95f6474c24c1dc or replacing it with a new oauth client id/secret pair.

settings.py
```diff
@@ -94,11 +94,22 @@
         'content_host': 'https://stable.dev.lcip.org',
         'oauth_host': 'https://oauth-stable.dev.lcip.org/v1',
         'profile_host': 'https://stable.dev.lcip.org/profile/v1',
-        'redirect_url': 'http://localhost:3000/api/v3/accounts/authenticate/',
+        'redirect_url': 'http://olympia.test/api/v3/accounts/authenticate/?config=local', # noqa
```
eviljeff: similarly here - you can't update it.  Also, localhost is correct - it's for when addons-frontend is run separately from addons-server and running on localhost.

settings.py
kumar303: > That will avoid confusion in the future I believe.

It's still pretty confusing 🤣 but, yeah, whatever works. I think we should delete them but maybe something will break, who knows.

src/olympia/conf/dev/settings.py
```diff
@@ -118,7 +118,7 @@
         'content_host': 'https://stable.dev.lcip.org',
         'oauth_host': 'https://oauth-stable.dev.lcip.org/v1',
         'profile_host': 'https://stable.dev.lcip.org/profile/v1',
-        'redirect_url': 'https://amo.addons-dev.allizom.org/fxa-authenticate',
+        'redirect_url': 'https://addons-dev.allizom.org/api/v3/accounts/authenticate/?config=amo', # noqa
```
eviljeff: if you want to change this check with ops the value of `AMO_FXA_CLIENT_ID` and verify the `redirect_url` specified on FxA.

src/olympia/conf/dev/settings.py
```diff
@@ -127,12 +127,21 @@
         'content_host': 'https://stable.dev.lcip.org',
         'oauth_host': 'https://oauth-stable.dev.lcip.org/v1',
         'profile_host': 'https://stable.dev.lcip.org/profile/v1',
-        'redirect_url': 'http://localhost:3000/fxa-authenticate',
+        'redirect_url': 'https://addons-dev.allizom.org/api/v3/accounts/authenticate/?config=local', # noqa
```
eviljeff: same here, though I suspect, like with `settings.py`, localhost is correct.

settings.py
willdurand: ooops, I think I messed up something because I did not intend to change this value that way :/

src/olympia/conf/dev/settings.py
willdurand: This is this one: https://oauth-stable.dev.lcip.org/console/client/cb6a5ba26feea430

src/olympia/conf/dev/settings.py
willdurand: Fixed this one too after having fetched the client ID on the server.

willdurand:
> the fxa oauth dashboard before changing the others.

Which dashboard? I believe prod/stage are ok now. I put the same things as in -dev, and values should be the same between these envs.

willdurand:
@eviljeff:

- stage config: https://oauth.accounts.firefox.com/v1/client/9dbca21ec249771d
- prod config: https://oauth.accounts.firefox.com/v1/client/3a1f53aabe17ba32

PR #11021, opened 2019-03-26
Add devhub upload, approve, install

jrbenny35:
This adds one of the more requested tests to CI. The full upload, submit, approve and install of an add-on. 

I had to create a little script to run the ```auto-approve``` command in the background because there were some issues with contacting the database from the tests with pytest-django. This can always be changed in the future though.


tests/ui/conftest.py
```diff
@@ -43,6 +44,16 @@ def firefox_notifications(notifications):
...
 
 
+@pytest.fixture(scope='session')
+def django_db_setup():
+    # print(settings.DATABASES)
+    settings.DATABASES['default'] = {
+        'ENGINE': 'olympia.core.db.mysql',
+        'HOST': 'mysqld',
+        'NAME': 'olympia',
+    }
```
EnTeQuAk: Is this actually still needed?

jrbenny35: No its not

settings_test.py
```diff
@@ -28,6 +28,37 @@
 STATIC_URL = '%s/static/' % CDN_HOST
 MEDIA_URL = '%s/user-media/' % CDN_HOST
 
+# We are setting memcached here to make sure our test setup is as close
```
diox: What's this copy/paste ?

jrbenny35: A bad one. Should be fixed

tests/ui/docker-compose.selenium.yml
```diff
@@ -15,7 +15,7 @@ x-env-mapping: &env
     - TERM=xterm-256color
     - UITEST_FXA_EMAIL
     - UITEST_FXA_PASSWORD=uitester
-    - MOZ_HEADLESS=1
+    - MOZ_HEADLESS
```
EnTeQuAk: Is this unsetting this or what is this doing? This isn't clear to be just from looking at it.

tests/ui/run_auto_approve.py
```diff
@@ -0,0 +1,15 @@
+import subprocess
+import time
+
+
+# Script for running the auto approve command every 10 seconds for 5 minutes.
+def approve_addons():
+    start_time = time.time()
+    # run for 5 minutes max
+    while time.time() != start_time + 300:
+        subprocess.call(["python", "manage.py", "auto_approve"])
```
EnTeQuAk: Let's adapt our single-quote style here too

tests/ui/test_devhub.py
```diff
@@ -57,3 +61,28 @@ def test_devhub_register(base_url, selenium):
...
+@pytest.mark.desktop_only
+@pytest.mark.nondestructive
+@pytest.mark.withoutresponses
+def test_devhub_addon_upload_approve_install(
+        base_url, selenium, devhub_upload, firefox, firefox_notifications):
+    """Test uploading an addon via devhub."""
+    'ui-test-addon-2' in devhub_upload.addons[-1].name
+    # We have to wait for approval
+    time.sleep(15)
+    selenium.get('{}/addon/{}'.format(base_url, 'ui-test_devhub_ext/'))
```
EnTeQuAk: Unsure if maybe https://docs.python.org/2/library/urlparse.html#urlparse.urljoin leads to a bit nicer code here… but given that we're joining two paths onto the base url it might not…

tests/ui/docker-compose.selenium.yml
jrbenny35: Just getting it from the ENV and passing it in

jrbenny35:
Looks like I am getting that missing checkbox error which is caused by a failed upload or something

tests/ui/test_devhub.py
jrbenny35: I could just add the second path to the string so its just one set of ```{}``` instead of 2

PR #11027, opened 2019-03-26
Move download url view from api to regular view.

### LINKED ISSUE #10994: Links to download files in the Reviewers API responses don't work because of an authentication problem
We cannot use the `download_url` in code-manager because those links require a token-based authentication (like API calls). Yet, this URL will be given to the users in order to download files.

EnTeQuAk:
This is needed to allow us to authenticate a user based on already set
cookies instead of unnecessarily complicated through an API.

Fixes #10994


eviljeff:
Lots of test fails though

EnTeQuAk:
> Lots of test fails though

I hate last minute changes, I'll fix them.

Issue #11014, opened 2019-03-26
'View on site' option from admin collection pages triggers a 500 server error


AlexandraMoga:
Regression from https://github.com/mozilla/addons-server/pull/10983
Somewhat similar to https://github.com/mozilla/addons-server/issues/10862

**STR:**

1. Log in to AMO -dev with an admin account
2. Open the Collections management tool
3. Select a collection from the list - i.e. https://addons-internal.dev.mozaws.net/en-US/admin/models/bandwagon/collection/5497518/change/
4. Click on the 'View on site' button in the top right corner

**Actual result:**
A 500 error is triggered for the following page: https://addons-internal.dev.mozaws.net/en-US/firefox/collections/4757633/webdeveloper/

**Expected result:**
The public collection page should be displayed

**Notes:**
- reproduced on AMO -dev with FF66, Win10x64

![2019-03-26_11-52-23](https://user-images.githubusercontent.com/31961530/54988312-d6f0da00-4fbe-11e9-8226-3431d597aa66.gif)

eviljeff:
it's the same as #10862 - the admin instance of AMO doesn't have frontend setup so any links to frontend pages return errors.

AlexandraMoga:
Verified fixed on -dev with FF66, Win10x64

![2019-04-10_11-48-20](https://user-images.githubusercontent.com/31961530/55865331-cd529f00-5b86-11e9-8fd1-da1faa5ab28c.gif)

PR #11032, opened 2019-03-27
Allow urls to new frontend views to link to the public facing site on internal instances

### LINKED ISSUE #10862: Trying to access a public profile from django admin Users will result in a 500 error
This looks like a regression from https://github.com/mozilla/addons-server/pull/10833

STR:
1. Log in to AMO -dev with an admin account
2. Access the User profiles management tool
3. Open the details of a user from the list - i.e. https://addons-internal.dev.mozaws.net/en-US/admin/models/users/userprofile/10641829/change/
4. Click on `View on site` button on the top right corner

Actual result:
A server error (500) is received

Expected result:
The public profile of the user is opened

Notes:
- the link should be pointing to the new frontend profile page
- reproduced on AMO -dev with FF65, Win10x64

eviljeff:
fixes #10862 and #11014 and any other place /admin links to a frontend view.

### DIFF
--- a/src/olympia/amo/models.py
+++ b/src/olympia/amo/models.py
@@ -2,10 +2,13 @@ import contextlib
 import os
 import time
 
+from urllib.parse import urljoin
+
 from django.conf import settings
 from django.core.files.storage import default_storage as storage
 from django.db import models
 from django.db.models.query import ModelIterable
+from django.urls.exceptions import Resolver404
 from django.utils import timezone, translation
 
 import elasticsearch
@@ -13,6 +16,7 @@ import multidb.pinning
 
 import olympia.core.logger
 
+from olympia.amo.urlresolvers import resolve
 from olympia.translations.hold import save_translations
 
 from . import search
@@ -369,7 +373,15 @@ class ModelBase(SearchMixin, SaveUpdateMixin, models.Model):
         base_manager_name = 'objects'
 
     def get_absolute_url(self, *args, **kwargs):
-        return self.get_url_path(*args, **kwargs)
+        from olympia.amo.views import frontend_view  # circular import
+
+        relative_url = self.get_url_path(*args, **kwargs)
+        try:
+            is_frontend = resolve(relative_url).func == frontend_view
+        except Resolver404:
+            is_frontend = False
+        site = settings.EXTERNAL_SITE_URL if is_frontend else settings.SITE_URL
+        return urljoin(site, relative_url)
 
     def serializable_reference(self):
         """Return a tuple with app label, model name and pk to be used when we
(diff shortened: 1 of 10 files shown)


eviljeff:
Actually, still not sure about my approach here.   Maybe I should be just adding handling for an external url directly in each model's `get_url_path`

EnTeQuAk:
Yes, I personally think that a models `get_url_path` is the better approach.

This gives us much more freedom I think.

eviljeff:
Although, the negative of that approach is duplicating the same logic in ~8 different model's `get_url_path`

diox:
Maybe we could have a generic implementation in `ModelBase` ? I really don't like touching `frontend_view` either.

diox:
Note: the API already has to deal with something similar. Serializers all wrap `get_url_path()` in a `absolutify()` call. Maybe there is something to refactor there so that the admin etc also do this.

eviljeff:
Re-implemented how this works to replace `get_absolute_url` instead.  A follow-up would be to replace the `absolutify(instance.get_url_path()` in the API serailizers.

src/olympia/reviewers/views.py
```diff
@@ -1201,8 +1202,9 @@ def download_git_stored_file(request, version_id, filename):
         raise http.Http404()
 
     actual_blob = serializer.git_repo[blob_or_tree.oid]
+
     response = http.FileResponse(
-        memoryview(actual_blob),
+        io.BytesIO(memoryview(actual_blob)),
```
EnTeQuAk: That shouldn't be in here.

src/olympia/reviewers/tests/test_views.py
```diff
@@ -5682,6 +5686,10 @@ def _test_url_success(self):
         response = self.client.get(url)
         assert response.status_code == 200
 
+        content = b''.join(response.streaming_content).decode('utf-8')
+        assert content.startswith('{')
+        assert '"manifest_version": 2' in content
```
EnTeQuAk: That shouldn't be in here.

src/olympia/reviewers/tests/test_views.py
```diff
@@ -5637,6 +5637,10 @@ def test_download_basic(self):
             response['Content-Disposition'] ==
             'attachment; filename="manifest.json"')
 
+        content = b''.join(response.streaming_content).decode('utf-8')
+        assert content.startswith('{')
+        assert '"manifest_version": 2' in content
```
EnTeQuAk: That shouldn't be in here.

src/olympia/reviewers/views.py
eviljeff: rebased and force pushed to get rid of the extra commit

Issue #10887, opened 2019-03-07
Link to code-manager file viewer from reviewer tools


kumar303:
Once we have https://github.com/mozilla/addons-code-manager deployed to [dev](https://github.com/mozilla/addons-code-manager/issues/327), let's add a link to the file viewer from reviewer tools. 

Let's add a link on this panel below, **in addition** to the existing *Contents* link. This will help the reviewers start testing out the new site. They will still be able to access the old file viewer if they need to.

<img width="261" alt="screenshot 2019-03-07 16 30 07" src="https://user-images.githubusercontent.com/55398/53994002-93f5c200-40f6-11e9-99bb-5f4fb8b8a23e.png">

This screenshot is from https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/couponshelper12#history

The link will obviously need a [Geocities icon](https://gifcities.org/?q=new) like:

![new3](https://user-images.githubusercontent.com/55398/53994118-ff3f9400-40f6-11e9-940e-ab9b7f546b1a.gif)

kumar303:
When discussing this with security, we want to only show this link to one or two reviewers while testing it. Maybe we can use a waffle flag that lets us turn it on for select users.

kumar303:
I turned on this flag to make it accessible to all users on -dev

AlexandraMoga:
Verified fixed on AMO -dev

A link to the new code manager is present on an addon review page  - except for static themes
![image](https://user-images.githubusercontent.com/31961530/55312768-f2505f00-546e-11e9-8b1f-25eb30405742.png)


I found an issue with search tools for which I filed #11085 


@kumar303 @jvillalobos regarding static themes, should we consider adding a link to code-manager on the review page? I can see that we could easily support this, so it is just a matter of deciding if we want to expose this link or not. 
https://code.addons-dev.allizom.org/en-US/browse/616605/versions/1655484/

jvillalobos:
Current theme review pages don't have source links, so let's keep it that way for now

PR #11038, opened 2019-03-27
Link to Code Manager from reviewer tools

kumar303:
Fixes https://github.com/mozilla/addons-server/issues/10887

I used a waffle flag because we want to run this side by side in production (to test it) and only give one or two reviewers access -- probably just Andreas. We need to make sure security is tight before reviewers can rely on the tool to make decisions.

When the flag is on:

Screenshot 2019-03-27 15 37 29


settings.py
```diff
@@ -124,6 +124,8 @@
 # Validation key we need to send in POST response.
 INBOUND_EMAIL_VALIDATION_KEY = 'totally-unsecure-validation-string'
 
+CODE_MANAGER_URL = 'http://localhost:3000'
```
eviljeff: Isn't this the same port that addons-frontend would run on?  You can't run both at once?

src/olympia/conf/stage/settings.py
```diff
@@ -149,3 +149,5 @@
     'amo-account-change-stage')
 
 VAMO_URL = 'https://versioncheck.allizom.org'
+
+CODE_MANAGER_URL = 'https://code.addons.allizom.org'
```
eviljeff: I wish we didn't need to define this in each conf/.../settings.py file but that's some refactoring away 😞

src/olympia/reviewers/templates/reviewers/includes/files_view.html
```diff
@@ -21,5 +21,28 @@
...
   </span>
 </li>
+{% if waffle.flag('code-manager') %}
+  <li class="code-manager-links">
+    <div>
+      <strong>Code Manager</strong>
+      <img width="25" height="9" src="{{ static('img/reviewers/new.gif') }}">
+    </div>
+    <a
+      href="{{ code_manager_url('/browse/{}/versions/{}/'.format(addon.pk, version.pk)) }}"
```
eviljeff: I wonder if this `.format` substitution should be moved down to `code_manager_url` also - i.e. add `addon`, `version` and (optional) `diff` args like `def code_manager_url(addon, version, diff=None)`  (and alter the url from browse to compare based on if diff is `None`)

settings.py
kumar303: That's true, it is the same port and you can't run both. Honestly, I don't think it would be uncommon to hook up a local `addons-server` to a local `addons-code-manager` so this is probably fine just as a reference. I could change it to the dev URL but I'm not sure if that would be useful since the data won't match (URLs will be 404s).

src/olympia/reviewers/templates/reviewers/includes/files_view.html
kumar303: The point of `code_manager_url()` is simply to share the common host prefix logic. I don't see any other shared functionality besides that. 

The two URLs are distinct; they both use add-on IDs and version IDs but in different ways. If I were to move that logic into `code_manager_url()` then it would be implemented with an if statement switching between the two URLs and I don't think that would be helpful. Also, it would have a negative side affect of forcing the reader to inspect `code_manager_url()` to understand what the final URL will look like.

src/olympia/conf/stage/settings.py
kumar303: Why not? Each URL is unique for each environment.

eviljeff: Ideally we'd just do `CODE_MANAGER_URL = 'https://code.' + DOMAIN` in settings_base.py but as we re-define `DOMAIN` in each `settings.py` we have to then (re-)define `CODE_MANAGER_URL` anyway.  It much harder to distinguish genuine differences between prod, stage & dev in the settings but it needs some serious and careful refactoring of all our settings files to not redefine `DOMAIN` and others settings each time.

settings.py
eviljeff: I thought you'd be running it on a different port by default anyway for that reason.  But I suppose it's an edge case you'd need that link to work reliably.

src/olympia/conf/stage/settings.py
kumar303: oh, I see what you mean. That makes sense. I updated the patch to use env vars, too.

PR #11047, opened 2019-03-28
drop global stats and dashboard

### LINKED ISSUE #11037: drop site-wide /statistics dashboard
From https://github.com/mozilla/addons-server/issues/10920#issuecomment-472155187

> /statistics/ should be dropped

(per addon statistics dashboards should stay as they are)

eviljeff:
fixes #11037


src/olympia/stats/models.py
```diff
@@ -105,18 +105,6 @@ class Meta:
...
 
 
-class GlobalStat(models.Model):
-    id = PositiveAutoField(primary_key=True)
-    name = models.CharField(max_length=255)
-    count = models.IntegerField()
-    date = models.DateField()
-
-    class Meta:
-        db_table = 'global_stats'
```
EnTeQuAk: Added https://github.com/mozilla/addons-server/issues/11107 to drop that table.

eviljeff: 🙊 I keep forgetting to file those follow up issues

PR #11051, opened 2019-03-28
Add CORS to validation results endpoint

### LINKED ISSUE #11048: Add CORS headers to `validation.json` URLs
The code-manager client fetches the linter data using the `validation.json` endpoints:

```
https://addons-dev.allizom.org/en-US/developers/addon/androidadbplus2/file/252861/validation.json
```

Because we are not on the same domain (`code.a.m.o` vs` a.m.o`) and CORS are not configured, we cannot retrieve the data in -dev (and therefore stage and prod). We don't have this problem in local dev because we're using an insecure proxy.

willdurand:
Fixes #11048

---

This patch adds:

1. a `CODE_MANAGER_URL` variable in all envs (similar to `SERVICES_URL`)
2. CORS headers to the `validation.json` endpoint

It should be enough for code-manager although I did not try because it is complicated to make addons-server work with code-manager locally. I'll be able to test this patch in -dev though (using Chrome "Local Overrides" and code-manager in -dev).

### DIFF
--- a/src/olympia/devhub/tests/test_views_validation.py
+++ b/src/olympia/devhub/tests/test_views_validation.py
@@ -3,6 +3,7 @@ import json
 import shutil
 
 from django.core.files.storage import default_storage as storage
+from django.test.utils import override_settings
 
 import mock
 import waffle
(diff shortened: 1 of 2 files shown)


src/olympia/devhub/tests/test_views_validation.py
```diff
@@ -212,6 +213,15 @@ def test_json_results_get_not_cached(self):
 
         assert self.client.get(self.json_url).status_code == 405
 
+    def test_cors_headers_are_sent(self):
+        code_manager_url = 'https://my-code-manager-url.example.org'
+        with override_settings(CODE_MANAGER_URL=code_manager_url):
+            r = self.client.get(self.json_url)
```
eviljeff: avoid single char variables as they can conflict during `ipdb` debugging.

PR #11055, opened 2019-03-28
drop frontend specific links/popups from base template

### LINKED ISSUE #10952: tidy up statistics dashboard base templates
#10920 has been logged to reimplement the statistics dashboards on addons-frontend but in the meantime we should remove as much of the legacy-frontend UI from the base templates used for statistics as possible (e.g. drop the "Extensions Themes Collections More…" links and popups).  The search box was removed in #10918

eviljeff:
fixes #10952 and from wherever else we still use the base templates.


eviljeff:
Looks like:
Screenshot 2019-03-28 18 15 21

PR #11056, opened 2019-03-28
Safe redirect

### LINKED ISSUE #11018: Does Django's `is_safe_url` accept subdomains?
Looking at the code of the `accounts/authenticate` endpoint, it looks like it is using `is_safe_url` to determine whether to send a redirect to the path specified by the client (`to` query parameter sent when starting the login dance by calling `login/start`) or redirect to the homepage.

https://github.com/mozilla/addons-server/blob/077126cd57508e6a534b5e779db436a3414d3bac/src/olympia/accounts/views.py#L355-L377

https://github.com/mozilla/addons-server/blob/077126cd57508e6a534b5e779db436a3414d3bac/src/olympia/accounts/views.py#L89-L93

In addons-frontend, we pass a relative path, e.g., `/en-US/firefox/`, to `login/start` but for code-manager, the API and the frontend don't have the same domain. I believe this can be a problem because it looks like passing a relative path in `?to=` means the API uses its own domain.

Yet, if we pass an absolute URL, e.g., `https://code.addons-dev.allizom.org/`, I believe `is_safe_url` will reject this URL because it does not have the same domain as `settings.DOMAIN`, hence the question in the title.

willdurand:
Fixes #11018 

---

That was a bit more work that I would have expected... Even though the test suite says so, I am not sure it works super well for local dev. It redirects to absolute URLs with `olympia.test` but I had troubles with `localhost`.

### DIFF
--- a/src/olympia/accounts/views.py
+++ b/src/olympia/accounts/views.py
@@ -31,6 +31,7 @@ from rest_framework.response import Response
 from rest_framework.status import HTTP_204_NO_CONTENT
 from rest_framework.views import APIView
 from rest_framework.viewsets import GenericViewSet
+from six.moves.urllib_parse import urlparse
 from waffle.decorators import waffle_switch
 
 import olympia.core.logger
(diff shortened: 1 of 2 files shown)


src/olympia/accounts/views.py
```diff
@@ -86,8 +86,17 @@
 API_TOKEN_COOKIE = 'frontend_auth_token'
 
 
-def safe_redirect(url, action):
-    if not is_safe_url(url, allowed_hosts=(settings.DOMAIN,)):
+def _is_safe_url(url, request):
+    """Override the Django `is_safe_url()` to pass a configured list of allowed
+    hosts and enforce HTTPS."""
+    allowed_hosts = (settings.DOMAIN, settings.CODE_MANAGER_DOMAIN, )
```
eviljeff: nit: you don't need the trailing comma in a tuple with >1 items.

settings.py
```diff
@@ -44,6 +44,7 @@
 SERVICES_URL = SITE_URL
 
 CODE_MANAGER_URL = os.environ.get('CODE_MANAGER_URL') or 'http://localhost:3000'
+CODE_MANAGER_DOMAIN = urlparse(CODE_MANAGER_URL).netloc
```
eviljeff: I'm not sure about having both `CODE_MANAGER_URL` and `CODE_MANAGER_URL` settings - can't you just do `urlparse(settings.CODE_MANAGER_URL).netloc` when you need the domain in `_is_safe_url`?

diox:
Works for me. I don't particularly care about trying to support anything other than `olympia.test`.

settings.py
willdurand: Yep, I guess I could do that.

PR #11076, opened 2019-03-29
Update docs for new procedures regarding integration testing.

jrbenny35:


docs/topics/development/testing.rst
```diff
@@ -80,9 +80,31 @@ the respective docker image, use the following command::
...
+    docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml exec web make -f Makefile-docker update_deps update_assets
+
+Followed by::
+
+    docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml restart web
+
+To setup the data within the database that the tests will look for run this command::
+
+    make -f Makefile-docker setup-ui-tests
+
```
eviljeff: Why do you have to do `make -f Makefile-docker`?  make is written to use `Makefile-docker` within the docker containers.

tests/ui/README.md
```diff
@@ -17,19 +17,15 @@ Follow the instructions found [here][addons-server-docs].
...
 
 To run the tests, execute the commands below:
 ```sh
 # Make sure all required containers are running.
 docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml up -d
-docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml exec --user root selenium-firefox tox -e ui-tests
+docker-compose -f docker-compose.yml -f tests/ui/docker-compose.selenium.yml exec selenium-firefox tox -e ui-tests
 ```
-WARNING: This will WIPE the database as the test will create specific data for itself to look for.
-If you have anything you don't want to be deleted, please do not run these tests.
```
eviljeff: it doesn't wipe the database any more?  (last time I tried it, it did!)

jrbenny35: It does but not that command. I'll move that under the ```make initialize_docker``` command.

docs/topics/development/testing.rst
jrbenny35: Hmm okay Ill confirm before changing.

jrbenny35: So when I run ```make setup-ui-tests``` I get an error that says ```No rule to make target 'setup-ui-tests'.  Stop.``` I know ```make initialize_docker``` exists within ```Makefile-os``` maybe that's why?

eviljeff: I'm not sure where this command is expected to be run.  If you see `Makefile` - the code to identify which makefile to use expects a folder to be present to identify it's running inside the web or worker container.

Commands which are expected to run outside of docker should be in Makefile-os; commands that run inside the container, in Makefile-docker.  (We have to override this specifically when running tests on Travis because the tests aren't run inside docker containers)

jrbenny35: Its meant to run inside the docker container as it is executing a bunch of database commands.

PR #11080, opened 2019-03-29
Make sure FxA state is correct when initiating login

willdurand:
This patch is a follow-up of #11018. The code dealing with the return
from FxA is correct, however FxA currently gets a state with an
`next_path` that is not set to the correct URL, mainly because the URL
given _via_ the `to` parameter was rejected.

This patch unifies the `is_safe_url` behavior for both the "login" and
the "authenticate" views.

I did not know why I had troubles with `localhost` in the previous patch,
but now I know: this was the issue. Everything works just fine locally!

### DIFF
--- a/src/olympia/accounts/utils.py
+++ b/src/olympia/accounts/utils.py
@@ -11,10 +11,22 @@ from django.utils.http import is_safe_url
 
 import boto3
 import six
-from six.moves.urllib_parse import urlencode
+from six.moves.urllib_parse import urlencode, urlparse
 
 from olympia.accounts.tasks import primary_email_change_event
 from olympia.core.logger import getLogger
+
+
+def _is_safe_url(url, request):
+    """Override the Django `is_safe_url()` to pass a configured list of allowed
+    hosts and enforce HTTPS."""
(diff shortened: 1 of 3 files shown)


src/olympia/accounts/utils.py
```diff
@@ -11,12 +11,24 @@
...
 from olympia.core.logger import getLogger
 
 
+def _is_safe_url(url, request):
+    """Override the Django `is_safe_url()` to pass a configured list of allowed
+    hosts and enforce HTTPS."""
+    allowed_hosts = (
+        settings.DOMAIN,
+        urlparse(settings.CODE_MANAGER_URL).netloc,
+    )
```
EnTeQuAk: I wonder if this should be a setting.

There is already an `ALLOWED_HOSTS` setting but that's being used for something else entirely (e.g csrf protection).

So maybe `ALLOWED_LOGIN_HOSTS` which is being set to these values? But then again we'd have to overwrite it explicitly in every hosting environments setting then which is kinda nasty too.

Maybe it's fine like it is here, just thinking loud :thinking:

PR #11088, opened 2019-04-01
Add reviewer name alias for reviewers to hide their actual name

### LINKED ISSUE #11002: [Reviewer Tools] Anonymising Reviewer Emails
It is a fact the developers are often not happy if their extensions are rejected during a review. While it should not cause any further complications, in reality that is not always the case. 
There are instance where the dissatisfied developers resort to targeting the reviewer via other AMO systems as a consequence. 
There are even situations where developer blames the addon-reviewer for issues not even related to the review or reviewer.

AMO systems such as reviewer's addon user-reveiw, addon support & addon Abuse Reports are then targeted for harassment and retaliation.

For example:
```
Harassment of reviewer via user-review system:
https://github.com/mozilla/addons/issues/962
https://github.com/mozilla/addons-server/issues/9686

Harassment of reviewer via add-on support pages:
https://github.com/erosman/support/issues/47
https://github.com/erosman/support/issues/51

Harassment of reviewer via add-on Abuse Reports (since 2015)
https://reviewers.addons.mozilla.org/en-US/reviewers/abuse-reports/firemonkey

Harassment of reviewer via discourse.mozilla.org
(there have been instances in the past)
```

IMHO, review emails could be generalised and anonymised (similar to auto-approval) in order to avoid such occurrences and to enable reviewers to review add-ons without the fear of becoming a victim of harassment or reprisal from unhappy developers.

@caitmuenster  @wagnerand  @kewisch

diox:
It's used when exposing reviewer name to developers for reviewer actions - regular actions still use the normal name, even when made by a reviewer.

Fixes #11002

### DIFF
--- a/docs/topics/api/accounts.rst
+++ b/docs/topics/api/accounts.rst
@@ -59,6 +59,7 @@ If you have `Users:Edit` permission you will see these extra fields for all user
 
     :>json boolean deleted: Is the account deleted.
     :>json string|null display_name: The name chosen by the user.
+    :>json string|null reviewer_name: The reviewer name chosen by the user, if set Only available for users with any kind of reviewer permission.
     :>json string email: Email address used by the user to login and create this account.
     :>json string fxa_edit_email_url: The configured URL for editing the user's email on FxA.
     :>json string last_login: The date of the last successful log in to the website.
--- a/docs/topics/api/accounts.rst
+++ b/docs/topics/api/accounts.rst
@@ -59,7 +59,7 @@ If you have `Users:Edit` permission you will see these extra fields for all user
 
     :>json boolean deleted: Is the account deleted.
     :>json string|null display_name: The name chosen by the user.
-    :>json string|null reviewer_name: The reviewer name chosen by the user, if set Only available for users with any kind of reviewer permission.
+    :>json string|null reviewer_name: The reviewer name chosen by the user, if set. Only available for users with any kind of reviewer permission.
     :>json string email: Email address used by the user to login and create this account.
     :>json string fxa_edit_email_url: The configured URL for editing the user's email on FxA.
     :>json string last_login: The date of the last successful log in to the website.
--- a/docs/topics/api/activity.rst
+++ b/docs/topics/api/activity.rst
@@ -39,6 +39,11 @@ Review Notes Detail
 
 This endpoint allows you to fetch a single review note for a specific version of an add-on.
 
+    .. note::
+        To allow reviewers to stay anonymous if they wish, the ``user`` object ``name`` can point to
+        their "reviewer" name depending on the action. In addition all other fields in that object,
+        despite being present for backwards-compatibility, are set to ``null``.
+
 .. http:get:: /api/v4/addons/addon/(int:addon_id|string:addon_slug|string:addon_guid)/versions/(int:id)/reviewnotes/(int:id)/
 
     .. _review-notes-version-detail-object:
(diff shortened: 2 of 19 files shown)


src/olympia/accounts/serializers.py
```diff
@@ -78,19 +78,28 @@ class UserProfileSerializer(PublicUserProfileSerializer):
...
-            'last_login_ip', 'read_dev_agreement', 'permissions',
-            'fxa_edit_email_url', 'username',
+            'deleted', 'display_name', 'email', 'fxa_edit_email_url',
+            'last_login', 'last_login_ip', 'permissions', 'picture_upload',
+            'read_dev_agreement', 'reviewer_name', 'username'
         )
         writeable_fields = (
             'biography', 'display_name', 'homepage', 'location', 'occupation',
-            'picture_upload',
+            'picture_upload', 'reviewer_name',
```
eviljeff: If we're exposing `reviewer_name` in the serializer response it should be documented in the api docs.  

How is displaying/editing `reviewer_name` going to work in the frontend?  Is there an issue logged to expect it and show it to the user?  There won't be any problem with having a writable field on the form that is only sometimes available?

src/olympia/activity/serializers.py
```diff
@@ -37,3 +36,16 @@ def get_action(self, obj):
...
+        avoid revealing actual name of reviewers for their review actions if
+        they have set an alias.
+
+        id, username and url are present for backwards-compatibility only."""
+        return {
+            'id': None,
+            'username': None,
+            'url': None,
+            'name': obj.author_name,
+        }
```
eviljeff: This will remove id, username and url for all activity log responses, not just reviewer related ones (`self.action in constants.activity.LOG_REVIEW_QUEUE_DEVELOPER`)

docs/topics/api/activity.rst
```diff
@@ -39,17 +39,22 @@ Review Notes Detail
 
 This endpoint allows you to fetch a single review note for a specific version of an add-on.
 
+    .. note::
+        To allow reviewers to stay anonymous if they wish, the ``user`` object ``name`` can point to
+        their "reviewer" name depending on the action. In addition all other fields in that object,
+        despite being present for backwards-compatibility, are set to ``null``.
```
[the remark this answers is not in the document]
diox: To keep things simple I've done that for all actions, since [the request in the issue](https://github.com/mozilla/addons-server/issues/11002#issuecomment-476845441) was to remove the links to user profiles entirely anyway.

I should remove all that in v4 yeah - I did the simplest thing that was still compatible with v3.

src/olympia/accounts/serializers.py
diox: I'll add docs and file an issue for that yeah - it was mentioned in the tasks to do in the addons-server issue.

If the user is a reviewer, they are supposed to see an extra field. I don't think it should be too much of a problem on the frontend.

src/olympia/activity/serializers.py
diox: In theory this serializer is only used by `VersionReviewNotesViewSet` which does that filtering already. But you're right, I should also do the check here just in case, relying on the view to filter is a bad idea (it's a vestige from an older version of this patch that always returned the reviewer alias for all actions)

src/olympia/accounts/serializers.py
diox: https://github.com/mozilla/addons-frontend/issues/7806

docs/topics/api/activity.rst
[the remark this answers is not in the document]
diox: 10bc305de5fe220fa4cf7ed75aabeb5070049026

src/olympia/activity/serializers.py
diox: 10bc305de5fe220fa4cf7ed75aabeb5070049026. Note that we don't need the extra check for the `name` since `author_name` does it for us.

docs/topics/api/accounts.rst
```diff
@@ -59,6 +59,7 @@ If you have `Users:Edit` permission you will see these extra fields for all user
 
     :>json boolean deleted: Is the account deleted.
     :>json string|null display_name: The name chosen by the user.
+    :>json string|null reviewer_name: The reviewer name chosen by the user, if set Only available for users with any kind of reviewer permission.
```
eviljeff: nit: " if set Only available" is missing some punctuation.

docs/topics/api/activity.rst
```diff
@@ -39,17 +39,19 @@ Review Notes Detail
...
 
     .. _review-notes-version-detail-object:
 
     :>json int id: The id for a review note.
     :>json string action: The :ref:`type of review note<review-note-action>`.
     :>json string action_label: The text label of the action.
-    :>json int user.id: The id of the reviewer or author who left the review note.
     :>json string user.name: The name of the reviewer or author.
-    :>json string user.url: The link to the profile page for of the reviewer or author.
-    :>json string user.username: The username of the reviewer or author.
```
eviljeff: Add some details to the API changelog about this.

PR #11092, opened 2019-04-01
remove legacy_discovery

### LINKED ISSUE #11089: remove legacy discovery modules
they're only used by legacy frontend now so we can delete them

eviljeff:
fixes #11089


eviljeff:
logged #11110

PR #11104, opened 2019-04-02
Expose risk weights breakdown in reviewer tools

### LINKED ISSUE #11096: Expose risk weights breakdown in reviewer tools
We calculate the risk of each add-on file using a number of weights specified in [this spreadsheet](https://docs.google.com/spreadsheets/d/1CZEECHHqEmK87fNiWNslAsFQBL7zxYF-c63-zZVZuqw/edit?ts=59c9d01f#gid=491359892), and currently show the final number in review pages and the review queue. This is the first step to make this breakdown more visible, per discussions with @atsay and @muffinresearch, among others.

For this issue, we need to:
- [ ] Store, for each add-on file, the value of each of the risk factors.
- [ ] Show the breakdown in the reviewer tools. I suggest a tooltip in the Weight number that appears in each item in the review history ([for example](https://reviewers.addons.mozilla.org/reviewers/review/adblock-plus#history)). The tooltip could say something like:
> Admin review flag: 0
> Size of code changes: 70
> etc.

And something like "Risk breakdown not available." for older versions.

Followup step is to have a dedicated page for it.

diox:
Fixes #11096

### DIFF
--- a/src/olympia/reviewers/models.py
+++ b/src/olympia/reviewers/models.py
@@ -13,6 +13,8 @@ from django.utils.translation import ugettext, ugettext_lazy as _
 
 import six
 
+from django_extensions.db.fields.json import JSONField
+
 import olympia.core.logger
 
 from olympia import amo
@@ -792,6 +794,7 @@ class AutoApprovalSummary(ModelBase):
         choices=amo.AUTO_APPROVAL_VERDICT_CHOICES,
         default=amo.NOT_AUTO_APPROVED)
     weight = models.IntegerField(default=0)
+    weight_info = JSONField(default={}, null=True)
     confirmed = models.NullBooleanField(default=None)
 
     class Meta:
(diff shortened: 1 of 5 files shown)


src/olympia/reviewers/models.py
```diff
@@ -13,6 +13,8 @@
 
 import six
 
+from django_extensions.db.fields.json import JSONField
```
eviljeff: I keep thinking there's a native `JSONField` now in django but it seems not 🐼

src/olympia/reviewers/templates/reviewers/review.html
```diff
@@ -101,7 +101,7 @@ <h3 id="history">
                   </div>
               {% endif %}
               {% if version.autoapprovalsummary %}
-                <div class="file-weight"><strong>{{ _('Weight:') }}</strong> {{ version.autoapprovalsummary.weight }}</div>
+                <div class="file-weight" title="{{ version.autoapprovalsummary.get_pretty_weight_info()|join('\n') }}"><strong>{{ _('Weight:') }}</strong> {{ version.autoapprovalsummary.weight }}</div>
```
eviljeff: Not essential, but I'd be interested to see a screenshot of what this looks like.  (If it's too much hassle doesn't matter - it can wait until it's live on addons-dev)

src/olympia/reviewers/models.py
```diff
@@ -850,9 +850,10 @@ def calculate_weight(self):
                         files__status=amo.STATUS_DISABLED)
                 .distinct().count() * 10, 100),
         }
-        factors.update(self.calculate_static_analysis_weight_factors())
-        self.weight = sum(factors.values())
-        return factors
+        self.weight_info.update(
+            self.calculate_static_analysis_weight_factors())
```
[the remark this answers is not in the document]
diox: It's done when calling `create_summary_for_version()` (modified below)

src/olympia/reviewers/models.py
diox: There is one... for postgres :disappointed:

src/olympia/reviewers/templates/reviewers/review.html
diox: Ask and you shall receive! 
risk_exposition

Note that yes, the keys of the dict are directly exposed without a translation, but I thought it was enough for now. We might want prettier strings later.

PR #11105, opened 2019-04-02
Allow credentials (CORS) for accounts/session API endpoint

### LINKED ISSUE #11100: Allow credentials to be included in cross-site HTTP requests (accounts/session endpoint)
The current CORS configuration returns `Access-Control-Allow-Origin: *` in the API responses, which makes it impossible to send credentials (cookies) along with an API call from a client hosted on a different domain.

This is problematic because code-manager cannot send credentials when it needs to. The error in Chrome or FF is:

```
Response to preflight request doesn't pass access control check: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.
```

Code-manager needs to send credentials when logging out a user: https://github.com/mozilla/addons-code-manager/issues/515.

The CORS configuration should allow credentials for this endpoint. ~~According to the CORS middleware source code, setting `CORS_ALLOW_CREDENTIALS` to `True` will add an extra header (`Access-Control-Allow-Credentials`) and change the `Access-Control-Allow-Origin` value to the `origin` of the request.~~ Read below, we decided to only change the CORS config for the `accounts/session` endpoint.

willdurand:
Fixes #11100

---

API call (with the authenticated by-passed):

```
$ http OPTIONS http://olympia.test/api/v4/accounts/session/ Origin:http://foobar.exampl.com
HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: accept, accept-encoding, authorization, content-type, dnt, origin, user-agent, x-csrftoken, x-requested-with
Access-Control-Allow-Methods: DELETE, GET, OPTIONS, PATCH, POST, PUT
Access-Control-Allow-Origin: http://foobar.exampl.com
Access-Control-Max-Age: 86400
Allow: DELETE, OPTIONS
Connection: keep-alive
Content-Length: 0
Content-Security-Policy: default-src 'self'; object-src 'none'; script-src https://ssl.google-analytics.com/ga.js https://www.google.com/recaptcha/ https://www.gstatic.com/recaptcha/ https://addons.cdn.mozilla.net http://www.google-analytics.com 'self'; child-src 'self' https://www.google.com/recaptcha/; form-action 'self' https://developer.mozilla.org; base-uri 'self' https://addons.mozilla.org; frame-src 'self' https://www.google.com/recaptcha/; media-src https://videos.cdn.mozilla.net; connect-src 'self' https://sentry.prod.mozaws.net https://addons.cdn.mozilla.net; style-src 'self' 'unsafe-inline' https://addons.cdn.mozilla.net; font-src 'self' https://addons.cdn.mozilla.net; img-src 'self' data: blob: https://ssl.google-analytics.com https://addons.cdn.mozilla.net https://static.addons.mozilla.net https://sentry.prod.mozaws.net http://www.google-analytics.com; report-uri /csp-report
Date: Tue, 02 Apr 2019 12:46:59 GMT
Server: nginx/1.15.8
Vary: User-Agent
X-AMO-Request-ID: 9ca36fa5c009493cbc23422ee80f9170
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
```

```
$ http DELETE http://olympia.test/api/v4/accounts/session/ Origin:http://foobar.exampl.com
HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: http://foobar.exampl.com
Allow: DELETE, OPTIONS
Connection: keep-alive
Content-Length: 11
Content-Security-Policy: default-src 'self'; object-src 'none'; script-src https://ssl.google-analytics.com/ga.js https://www.google.com/recaptcha/ https://www.gstatic.com/recaptcha/ https://addons.cdn.mozilla.net http://www.google-analytics.com 'self'; child-src 'self' https://www.google.com/recaptcha/; form-action 'self' https://developer.mozilla.org; base-uri 'self' https://addons.mozilla.org; frame-src 'self' https://www.google.com/recaptcha/; media-src https://videos.cdn.mozilla.net; connect-src 'self' https://sentry.prod.mozaws.net https://addons.cdn.mozilla.net; style-src 'self' 'unsafe-inline' https://addons.cdn.mozilla.net; font-src 'self' https://addons.cdn.mozilla.net; img-src 'self' data: blob: https://ssl.google-analytics.com https://addons.cdn.mozilla.net https://static.addons.mozilla.net https://sentry.prod.mozaws.net http://www.google-analytics.com; report-uri /csp-report
Content-Type: application/json
Date: Tue, 02 Apr 2019 12:47:33 GMT
Server: nginx/1.15.8
Set-Cookie: frontend_auth_token=""; expires=Thu, 01 Jan 1970 00:00:00 GMT; Max-Age=0; Path=/
Set-Cookie: multidb_pin_writes=y; expires=Tue, 02 Apr 2019 12:47:48 GMT; Max-Age=15; Path=/
Vary: User-Agent
X-AMO-Request-ID: 1c5bd4e152904ca3b5b8c71bb9b2ed94
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
X-XSS-Protection: 1; mode=block

{
    "ok": true
}
```

### DIFF
--- a/src/olympia/accounts/views.py
+++ b/src/olympia/accounts/views.py
@@ -395,15 +395,14 @@ def logout_user(request, response):
 class SessionView(APIView):
     permission_classes = [IsAuthenticated]
 
-    def add_common_cors_headers(self, request, response):
-        origin = request.META.get('HTTP_ORIGIN')
-        response[ACCESS_CONTROL_ALLOW_ORIGIN] = origin
-        response[ACCESS_CONTROL_ALLOW_CREDENTIALS] = 'true'
-
     def options(self, request, *args, **kwargs):
         response = Response()
         response['Content-Length'] = '0'
-        self.add_common_cors_headers(request, response)
+        origin = request.META.get('HTTP_ORIGIN')
+        if not origin:
+            return response
+        response[ACCESS_CONTROL_ALLOW_ORIGIN] = origin
+        response[ACCESS_CONTROL_ALLOW_CREDENTIALS] = 'true'
         # Mimics the django-cors-headers middleware.
         response[ACCESS_CONTROL_ALLOW_HEADERS] = ', '.join(
             corsheaders_conf.CORS_ALLOW_HEADERS
@@ -419,7 +418,11 @@ class SessionView(APIView):
     def delete(self, request, *args, **kwargs):
         response = Response({'ok': True})
         logout_user(request, response)
-        self.add_common_cors_headers(request, response)
+        origin = request.META.get('HTTP_ORIGIN')
+        if not origin:
+            return response
+        response[ACCESS_CONTROL_ALLOW_ORIGIN] = origin
+        response[ACCESS_CONTROL_ALLOW_CREDENTIALS] = 'true'
         return response
 
 
--- a/src/olympia/accounts/views.py
+++ b/src/olympia/accounts/views.py
@@ -17,6 +17,11 @@ from django.utils.translation import ugettext, ugettext_lazy as _
 import six
 import waffle
 
+from corsheaders.conf import conf as corsheaders_conf
+from corsheaders.middleware import (
+    ACCESS_CONTROL_ALLOW_ORIGIN, ACCESS_CONTROL_ALLOW_CREDENTIALS,
+    ACCESS_CONTROL_ALLOW_HEADERS, ACCESS_CONTROL_ALLOW_METHODS,
+    ACCESS_CONTROL_MAX_AGE)
 from rest_framework import serializers
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.decorators import action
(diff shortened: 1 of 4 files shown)


src/olympia/accounts/views.py
```diff
@@ -385,12 +390,36 @@ def logout_user(request, response):
         API_TOKEN_COOKIE, domain=settings.SESSION_COOKIE_DOMAIN)
 
 
+# This view is not covered by the CORS middleware, see:
+# https://github.com/mozilla/addons-server/issues/11100
 class SessionView(APIView):
     permission_classes = [IsAuthenticated]
 
+    def add_common_cors_headers(self, request, response):
+        origin = request.META.get('HTTP_ORIGIN')
```
diox: if `origin` is not set, we probably don't want to add the headers, and return straight away?

willdurand: ah yes, that would make sense.

diox: Don't forget to do something sensible for `OPTIONS` too (not sure what)

willdurand: Looks like the middleware does not add anything when `Origin` is missing.

src/olympia/accounts/views.py
```diff
@@ -385,12 +390,39 @@ def logout_user(request, response):
...
 
+    def options(self, request, *args, **kwargs):
+        response = Response()
+        response['Content-Length'] = '0'
+        origin = request.META.get('HTTP_ORIGIN')
+        if not origin:
+            return response
+        response[ACCESS_CONTROL_ALLOW_ORIGIN] = origin
+        response[ACCESS_CONTROL_ALLOW_CREDENTIALS] = 'true'
+        # Mimics the django-cors-headers middleware.
```
kumar303: Re-implementing parts of `django-cors-headers` makes me wonder if we should consider an alternate approach. What about doing the logout as a redirect, as mentioned in https://github.com/mozilla/addons-server/issues/11100#issuecomment-478919885 ? That seems like a reasonable solution.

EnTeQuAk: I don't think that won't be a big problem given it's only for this view and only for logout and as such should only be happening once in the whole code.

src/olympia/amo/tests/test_views.py
```diff
@@ -377,6 +378,12 @@ def test_cors_api_v4(self):
         assert not response.has_header('Access-Control-Allow-Credentials')
         assert response['Access-Control-Allow-Origin'] == '*'
 
+    def test_cors_excludes_accounts_session_endpoint(self):
+        assert re.match(
+            settings.CORS_URLS_REGEX,
+            reverse('v4:accounts.session'),
```
EnTeQuAk: This should probably use `reverse_ns` to make it work for all api versions maybe. Nitpick though-

willdurand: I tried but `reverse_ns` returns an absolute URL and the regex is for a (relative) path.

willdurand:
Looks like Circle CI is failing for other reasons (`fxa.errors.ClientError: Account already exists` in the logs).

diox:
I'm having doubts over something: shouldn't we validate the origin ? Technically internal authentication should be required, so CSRF shouldn't be an issue, but maybe we want to be extra-safe ?

src/olympia/amo/tests/test_views.py
eviljeff: yeah... we're quite careful to avoid using django's reverse and `v4:xxx` syntax for API urls.  
`from rest_framework.reverse import reverse` should have been used instead though - it doesn't return absolute urls by default.

willdurand: Let me fix that in a follow-up patch, thanks!

PR #11119, opened 2019-04-03
Correctly stream the raw file content.

### LINKED ISSUE #1118: add-on valdation failed with 1 error, but in the validation report is no error or warning
i have tried to upload a little addon to the amo, but the validation failed with 1 error. so looked into the full validation report, but there was no error or warning...
submit
validation_report

EnTeQuAk:
If passed a bytestring as data the streamed content will be bonkers, if
passed a BytesIO file-like object the actual data will be correctly
streamed.

In theory, with a BytesIO wrapping the memoryview we should be keeping
the advantages of using `memoryview` for the blob, which is reduced
memory usage. In practice this probably won't matter much.

I have no idea how to actually test the memory-exhausting to be honest,
so this is primarily guesswork and docs-reading.

Fixes #1118

### DIFF
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -1,5 +1,6 @@
 import json
 import time
+import io
 
 from collections import OrderedDict, defaultdict
 from datetime import date, datetime, timedelta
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/views.py
```diff
@@ -1201,8 +1202,9 @@ def download_git_stored_file(request, version_id, filename):
         raise http.Http404()
 
     actual_blob = serializer.git_repo[blob_or_tree.oid]
+
     response = http.FileResponse(
-        memoryview(actual_blob),
+        io.BytesIO(memoryview(actual_blob)),
```
eviljeff: yeah it seems like there should be a more direct way without having to wrap `serializer.git_repo[blob_or_tree.oid]` in both a `memoryview` and `io.BytesIO` but finding out would just be trial and error and it seems you've done that.

PR #11124, opened 2019-04-03
Indicate file status in file.entries.

### LINKED ISSUE #11123: Indicate which git file status in compare response
Blocks https://github.com/mozilla/addons-code-manager/issues/413

We should indicate in the compare API the added/changed/modified status of files.

EnTeQuAk:
We're now adding a new `status` property to `file.entries` which
indicates the git status of the file.

This will make it easier to show added/deleted/removed files.

This does not yet cover renamed/copied status as this will happen in #10849

Fixes #11123


src/olympia/reviewers/tests/test_serializers.py
```diff
@@ -152,6 +155,94 @@ def test_get_entries_cached(self):
...
+        readme_data = data['entries']['test.txt']
+        assert readme_data['depth'] == 0
+        assert readme_data['filename'] == u'test.txt'
+        assert readme_data['sha256'] == (
+            'f8b40fc302692ea4f552cb3d60bc89dd8b4616e398de5585e471cee73e2c0618')
+        assert readme_data['mimetype'] == 'text/plain'
+        assert readme_data['mime_category'] == 'text'
+        assert readme_data['path'] == u'test.txt'
+        assert readme_data['size'] == 18
+        assert readme_data['status'] == 'A'
```
kumar303: I don't see a test for showing deleted files. Is there one? I'm a bit confused about how we will access deleted files because it doesn't really make sense to include deleted files in `data['entries']`. Or does it?

Some details: https://github.com/mozilla/addons-code-manager/issues/413#issuecomment-479659974

EnTeQuAk: You're right, I added proper support for file deletion in 2367135ea9.

So, deleted files are in `data['entries']`, only they have their status `'D'` for deleted and most of their properties set to `null` given that we don't care anymore (and would have to fetch the previous commit to fetch the data which is unnecessary)

This is the first try at this, very happy to hear feedback regarding structure and data. For me, it'd be easier to just have the status in the end-structure but I figured it might be easier for you to reuse the same data-structure.

Does that work for you like that?

kumar303: Yep, this looks good: https://github.com/mozilla/addons-server/commit/2367135ea975e0f0aec0ff4c8f4f8927de0d63ae#diff-b02a60627dc42d205bfd52770cafe54cR250

Thanks!

docs/topics/api/reviewers.rst
```diff
@@ -158,12 +158,24 @@ This endpoint allows you to compare two Add-on versions with each other.
...
+    Properties specific to this endpoint:
 
-    The following output represents this git diff.
+    :>json array file.entries[]: The complete file-tree of the extracted XPI.
+    :>json string files.entries[].status: Status of this file, see https://git-scm.com/docs/git-status#_short_format
+    :>json int|null file.entries[].depth: Level of folder-tree depth, starting with 0.
+    :>json string file.entries[].filename: The filename of the file.
+    :>json string file.entries[].path: The absolute path (from the root of the XPI) of the file.
+    :>json string|null file.entries[].sha256: SHA256 hash.
+    :>json string|null file.entries[].mimetype: The determined mimetype of the file or ``application/octet-stream`` if none could be determined.
```
kumar303: Maybe this should be more clear that deletions are `null`? Otherwise it sounds like the mimetype would be `application/octet-stream`

EnTeQuAk: Fixed in 61da03f197

src/olympia/lib/git.py
```diff
@@ -402,46 +403,41 @@ def iter_tree(self, tree):
...
 
-    def get_diff(self, commit, parent=None, pathspec=None):
-        """Get a diff from `parent` to `commit`.
-
-        If `parent` is not given we assume it's the first commit and handle
-        it accordingly.
-
-        We are resolving renames and copies according to a 50% similarity
-        threshold.
+    @lru_cache(maxsize=None)
```
diox: `maxsize=None` ? This seems quite dangerous, the cache will grow forever in the python process, which can itself live for quite a while...

diox: I think what you need here is something that caches on the object instance - `self` - so that it goes away after you're done with that instance. Just doing `lru_cache(maxsize=1)` might work but it's per-process so it may be it's a little risky. A quick search suggest decorating with `lru_cache` in the object's `__init__`, since the cache is attached to the function being decorated, doing it at `__init__` time might be safe (still need `maxsize=1` though).

Will need some tests in any case.

EnTeQuAk: Solved this a bit differently in fb60d6d, I hope that works too? Added a test as well.

EnTeQuAk: I figured that would be easier to implement / test compared to re-using lru_cache here too. Especially since lru_cache(maxsize=1) wouldn't give us any benefits of the actual data structure and with the destruction of `self` happening right after serialization we should be safe.

kumar303:
@diox @eviljeff could either of you help us land this next week while Chris is out on PTO? It's a high priority issue for the code manager.

EnTeQuAk:
This should be all ready, just needs another look regarding the latest cache related changes.

Issue #11034, opened 2019-03-27
Return absolute URLs in `validationURL` fields


willdurand:
The `validationURL` field returned by the browse and compare APIs have relative URLs as values. In code-manager, we would need absolute URLs instead, mainly to fix this issue: https://github.com/mozilla/addons-code-manager/issues/480.

willdurand:
Actually, let's not do this. This will break our nice local dev env in code-manager, forcing developers to log in locally (localhost) and on -dev.

EnTeQuAk:
re-opening according to the discussion in #11128

EnTeQuAk:
Sorry I stole this issue, I've been touching that part so I thought I'd just do it along the way too.

PR #11128, opened 2019-04-03
Return relative download URLs

### LINKED ISSUE #11035: Return relative URLs in `downloadURL` fields
We added `downloadURL` fields to the browse and compare APIs recently. The values returned by the API are absolute URLs so far, but it'd be easier for code-manager (especially for the local dev setup) to use relative URLs.

Can we have that?

willdurand:
Fixes #11035

### DIFF
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -15,7 +15,6 @@ from django.utils.encoding import force_text
 from django.utils.timezone import FixedOffset
 
 from olympia.amo.urlresolvers import reverse
-from olympia.amo.templatetags.jinja_helpers import absolutify
 from olympia.addons.serializers import (
     VersionSerializer, FileSerializer, SimpleAddonSerializer)
 from olympia.addons.models import AddonReviewerFlags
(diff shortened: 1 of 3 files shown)


diox:
mmm @eviljeff and I had the opposite idea in mind to fix https://github.com/mozilla/addons-server/issues/10862 and https://github.com/mozilla/addons-server/issues/11014

eviljeff:
yeah, I'm not sure what #11035 is fixing exactly.

willdurand:
> mmm @eviljeff and I [had the opposite idea in mind](https://github.com/mozilla/addons-server/pull/11032#issuecomment-478918719) to fix #10862 and #11014

I am not sure how these issues are related to #11035: the problem with the reviewers API is that it is not on the same domain as the client. Having relative URLs will allow code-manager to build the final URLs, ensuring it works for any env, including local dev.

The URL to the validation results is a relative URL and thanks to that, we were able to fetch the results and then develop the UI for them.

Currently, downloading a file in local dev is broken because it calls the -dev server and not the local proxy (like for API calls, validation results, etc.)

eviljeff:
Thinking just about the API, imo, all urls to something external to the API, including django views that are currently served under the same domain as the API, should be returned as absolute urls.  

That said, my involvement with the APIs relating to the code manager has been limited to looking at pull requests so 🤷

willdurand:
If that makes everything consistent, I can still re-open https://github.com/mozilla/addons-server/issues/11034 and make these absolute urls work again in code-manager (local dev). That will probably be a matter of removing the base URL.

diox:
I think that might be a lot better in the long run. As @eviljeff pointed out we're kinda moving towards every non-API URL returned by the API be absolute - and that's why this PR conflicts with that work.

willdurand:
That makes sense, thanks!

PR #11156, opened 2019-04-05
Allow using guids in add-on urls

### LINKED ISSUE #11154: Allow accessing reviewer and django admin add-ons page using add-on id
Admins constantly have to convert between the add-on ID (guid) and the AMO ID.
Could we add url handlers for the following urls?

```
/reviewers/review/<guid>
/reviewers/review-listed/<guid>
/reviewers/review-unlisted/<guid>
/reviewers/review-content/<guid>
/admin/models/addons/addon/<guid>
/admin/addon/manage/<guid>
```

Should we also add a handler for the public pages `/firefox/addon/<guid>` ?

wagnerand:
Fix #11154.

### DIFF
--- a/src/olympia/addons/admin.py
+++ b/src/olympia/addons/admin.py
@@ -1,15 +1,17 @@
 from six.moves.urllib_parse import urlencode
 
-from django import forms
+from django import http, forms
 from django.conf import settings
 from django.contrib import admin
 from django.core import validators
+from django.shortcuts import get_object_or_404
 from django.urls import resolve
 from django.utils.html import format_html
 from django.utils.translation import ugettext, ugettext_lazy as _
 
 from olympia import amo
 from olympia.access import acl
+from olympia.addons.models import Addon
 from olympia.amo.urlresolvers import reverse
 from olympia.ratings.models import Rating
 from olympia.zadmin.admin import related_content_link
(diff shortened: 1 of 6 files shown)


src/olympia/addons/admin.py
```diff
@@ -83,6 +85,29 @@ def status_with_admin_manage_link(self, obj):
                            link, obj.get_status_display())
     status_with_admin_manage_link.short_description = _(u'Status')
 
+    def change_view(self, request, object_id, form_url='', extra_context=None):
+        lookup_field = Addon.get_lookup_field(object_id)
```
diox: We need something more DRY here... Can you extract the common code with `addon_view` decorator and re-use that ?

wagnerand: I decided to duplicate this code because here we are redicrecting to the AMO id (vs the slug). I think if we want to keep the redirects in place, having those distinct implementations make sense because we probably don't want this one re-used on a public page. Let me know what you think.

wagnerand:
Could you more specific what tests are missing? I added tests for accessing the urls by add-on id, tests for the other lookup keys were already in place, together with their implementation.

diox:
Looks like I was looking at an earlier diff, didn't see those tests :see_no_evil:

src/olympia/addons/tests/test_admin.py
```diff
@@ -177,6 +177,32 @@ def test_can_not_edit_without_addons_edit_permission(self):
...
+    def test_access_using_slug(self):
+        addon = addon_factory(guid='@foo')
+        self.detail_url = reverse(
+            'admin:addons_addon_change', args=(addon.slug,)
+        )
+        user = user_factory()
+        self.grant_permission(user, 'Admin:Tools')
+        self.grant_permission(user, 'Addons:Edit')
+        self.client.login(email=user.email)
+        response = self.client.get(self.detail_url, follow=True)
```
diox: I think I'd rather see a test proving we redirect to the right URL (so `folllow=False`, you can use our `assert3xx` method with the expected URL). Same thing below.

PR #11167, opened 2019-04-08
remove legacy rating views; add RatingFlagSerializer

### LINKED ISSUE #11155: remove legacy views from /ratings
they're not referenced or accessible any more

eviljeff:
fixes #11155 

Most of the change is deleting the legacy rating view but I also had to add a serializer (I guess I could have done it directly in the view, but its done now) to handle flagging ratings because we reused the legacy ui view&form for the API too.


src/olympia/ratings/serializers.py
```diff
@@ -7,19 +7,29 @@
...
 
 
+# This matches the following three types of patterns:
+# http://... or https://..., generic domain names, and IPv4 octets. It does not
+# match IPv6 addresses or long strings such as "example dot com".
+link_pattern = re.compile(
+    r'((://)|'  # Protocols (e.g.: http://)
+    r'((\d{1,3}\.){3}(\d{1,3}))|'
+    r'([0-9a-z\-%%]+\.(%s)))' % '|'.join(TLDS),
+    (re.I | re.U | re.M))
```
EnTeQuAk: I get the feeling that you may be able to re-use `django.core.validators:URLValidator.regex` instead which allows much more stuff and is also much better tested (e.g, way more thorough tld restrictions etc).

https://github.com/django/django/blob/master/django/core/validators.py#L95

EnTeQuAk: I'm perfectly aware that this only got copied so treat this as a nitpick and nothing blocking merge.

eviljeff: I just tried it but `URLValidator`'s regex has `^` so it doesn't match.

PR #11227, opened 2019-04-19
Store add-on guid when saving an Abuse Report against an add-on in database

### LINKED ISSUE #11219: Store current add-on guid when saving an Abuse Report against an add-on in database
In the Abuse Reporting API, when we save an abuse report against an add-on in the database, we don't save the `guid` - we only do for add-ons we don't know about.

In the admin, we want to be able to find all reports for a given guid, so we need to also store the guid that the add-on had when the report was created.

diox:
Fixes #11219

### DIFF
--- a/src/olympia/migrations/1087-abuse-add-guid-to-existing-addon-reports.sql
+++ b/src/olympia/migrations/1087-abuse-add-guid-to-existing-addon-reports.sql
@@ -1,0 +1,5 @@
+UPDATE `abuse_reports`
+    INNER JOIN `addons` ON (`abuse_reports`.`addon_id` = `addons`.`id`)
+    SET `abuse_reports`.`guid` = `addons`.`guid`
+    WHERE `abuse_reports`.`addon_id` IS NOT NULL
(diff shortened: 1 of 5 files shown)


src/olympia/migrations/1087-abuse-add-guid-to-existing-addon-reports.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE `abuse_reports`
+    INNER JOIN `addons` ON (`abuse_reports`.`addon_id` = `addons`.`id`)
```
eviljeff: for add-ons that have been deleted, but then replaced with a new add-on listing, `guid` will be a string like `guid-reused-by-pk-xxx` where x is the new add-on id.  Maybe that's fine just for backfilling, but it's not really the guid so just pointing it out.

diox: Yeah, I had that in mind - I think that's fine since it's just for backfilling.

PR #11238, opened 2019-04-23
Add list and search views for abuse reports admin

### LINKED ISSUE #11133: List and search view for abuse reports
Part of the the [abuse reporting project](https://docs.google.com/document/d/1K5Th9UmpayfVMXadkxU6cDO3sFKOwYozQgyoN0WEGwY/edit#). 

The abuse reporting admin needs a list / search view. It should list all reports by default, with the most recent ones at the top. It should be accessible to users with the  (new) `AbuseReports:Edit` permission.

The following fields should describe report entries:
* Add-on name and version
* Signing state of the add-on version (signed, partner, curated. If not yet available: self-hosted, listed, featured)
* Category (spam, malicious, etc.)
* State (untriaged/valid/suspicious/deleted)
* Excerpt from the abuse report text

When searching, the following fields should be available for filtering:

* Reports for a specific add-on (by id, name, or slug)
* Search for specific text keywords in the report text
* Category
* State (untriaged/valid/suspicious/deleted)
* For reports marked valid, a time range in which the report was marked
* Minimum amount of abuse reports in a (user-specified) timeframe

The following bulk actions should be available:
* Delete all checked reports
* Mark all checked reports as suspicious
* Mark all checked reports as valid

Here's a sketch of this view (slightly outdated, so it's just a rough guide):

Screen Shot 2019-04-03 at 3 08 06 PM

diox:
Includes new 'state' property that allows AbuseReports to be soft-deleted, marked as valid or suspicious. The default state is untriaged.

Fixes #11133


src/olympia/abuse/models.py
```diff
@@ -160,10 +193,7 @@ def target(self):
 
     @property
     def type(self):
-        with translation.override(settings.LANGUAGE_CODE):
-            type_ = (translation.ugettext(amo.ADDON_TYPE[self.addon.type])
-                     if self.addon else 'User' if self.user else 'Addon')
-        return type_
+        return 'User' if self.user else 'Addon'
```
eviljeff: So we're dropping the ability to differentiate between different types of Add-on in the emails, admin, etc?  Themes abuse reports would be indistinguishable from Extension reports.

diox: Ah, I had not considered the impact to themes. It got in my way because the main thing I needed to differentiate between was users and addons. @wagnerand do you know if we care ? If we do, I can add it back, I'll need to figure out how to implement type filtering in the admin differently...

wagnerand: I think it would be good to have since theme admins are not necessarily the same group of people as extension admins. If it's major effort though, let's talk about it.

diox: 27215b8d382d88841b0bbffdf91cd31be4dc4424 restores it. The drawback is that it can be a little confusing because the admin doesn't let you filter that precisely (it's either addons or users when filtering by type) even though it displays the precise type in the corresponding column.

PR #11275, opened 2019-04-29
Add model support for recommendations

### LINKED ISSUE #11064: Add a recommended add-on to the list
Currently we have the recommended add-on list used by TAAR and return to AMO this list is a good place to start.

We can continue to use this list, but we will need to be able to do the following:

* A newly recommended add-on will need to be placed into a "pending recommendation" state.
* The next version uploaded will be pushed into a different review queue (the review flow is a independent project), the next version will be signed as recommended only once the review has passed.
* Unlisted add-ons can be recommended

For more details see the PRD: https://docs.google.com/document/d/1fRxLy7h02tn1HBIlsTU_6h9DHmN8c3N6bg4JoxHF7J8/edit#heading=h.pbx0vxiwi6qq

During the changes we must ensure that the TAAR and RTAMO lists continue to function.

### LINKED ISSUE #11278: expose recommended status in the API
The API needs to expose recommended status so Frontend can badge it appropriately.  Addon will expose `is_recommended`.  Version will expose something - probably a similar `is_recommended` field - tbc.

supports mozilla/addons-frontend#7793

eviljeff:
fixes #11064 ~?  Probably needs some reviewer tools changes to complete the issue.~ and fixes #11278 and fixes #11275


src/olympia/constants/base.py
```diff
@@ -75,6 +75,16 @@
 VALID_ADDON_STATUSES = (STATUS_NOMINATED, STATUS_APPROVED)
 VALID_FILE_STATUSES = (STATUS_AWAITING_REVIEW, STATUS_APPROVED)
 
+RECOMMENDATION_STATUS_CHOICES = {
```
diox: Re-using statuses mmm... Cute, and I appreciate that it helps us avoid death by a thousand constants, but I feel like this will be more confusing than helping. Not strongly opposed to it though

EnTeQuAk: I agree, it'll mix- and match different meanings of constants which will be confusing when grepping for a specific constant, see it's usage etc.

src/olympia/versions/models.py
```diff
@@ -131,6 +131,10 @@ class Version(OnChangeMixin, ModelBase):
     git_hash = models.CharField(max_length=40, blank=True)
     source_git_hash = models.CharField(max_length=40, blank=True)
 
+    recommendation_status = models.PositiveSmallIntegerField(
+        choices=amo.RECOMMENDATION_VERSION_STATUS_CHOICES.items(),
+        default=amo.STATUS_NULL)
+
```
diox: Do we really need that extra field ? We're going to have to update `status` on the `DiscoveryItem` anyway when a version is approved, right ?

eviljeff: We need to track which specific version (or file) has been approved - otherwise the developer could delete the approved versions and we'd be recommending a version that was unreviewed.  I'm 99% certain this was something we agreed was necessary in the previous meeting :)

I'm not sure if we need 3 states on `DiscoveryItem` though - we have NULL, AWAITING_REVIEW and APPROVED but NULL and some other non-NULL state are probably enough because if the `Addon.current_version` isn't `STATUS_APPROVED` then it's not offered as recommended.

diox: You're totally right, we did agree on that, I even pointed out in the PRD that it'd be stored per version/file. Carry on :) 

Maybe a simple boolean would work for both fields ?

eviljeff: Replaced both `DiscoveryItem` and `Version` field with `BooleanField`s https://github.com/mozilla/addons-server/pull/11275/commits/9e44673d0f3352b206ac6e188f2ea2c1d1a10dc3

src/olympia/constants/base.py
eviljeff: https://github.com/mozilla/addons-server/pull/11275/commits/9e44673d0f3352b206ac6e188f2ea2c1d1a10dc3 makes the constants redundant now.

diox:
Not sure if we need to add reviewer tools support in this PR, but I think this would be the ideal moment to add the recommended state in the admin.

eviljeff:
I'm thinking now that https://github.com/orgs/mozilla/projects/70 covers anything reviewer related.  Adding the code which would set `recommendation_approved` during a review isn't something we want to do until the queues are split out or a standard manual review would inadvertently approve the addon for recommended too.

I'll add the fields to admin.  I have a commit that adds API support too which I may as well include.

src/olympia/addons/indexers.py
```diff
@@ -317,7 +318,8 @@ def extract_document(cls, obj):
         attrs = ('id', 'average_daily_users', 'bayesian_rating',
                  'contributions', 'created',
                  'default_locale', 'guid', 'hotness', 'icon_hash', 'icon_type',
-                 'is_disabled', 'is_experimental', 'last_updated',
+                 'is_disabled', 'is_experimental', 'is_recommended',
```
diox: If you add it there, it needs to always exist - see the `getattr` below, it doesn't have a default value, because it doesn't know what makes sense for each individual field. Since it's a new property, you should explicitly do a separate `getattr(obj, `is_recommended`, False)` or something like that instead, then later, when we've reindexed, we'll be free to change this back to your original version

src/olympia/addons/models.py
```diff
@@ -1268,6 +1268,18 @@ def get_featured_by_app(self):
             out[app].add(locale or None)
         return out
 
+    @cached_property
```
diox: Do we need it to be a `@cached_property`? Do we end up requesting that twice in a row somewhere for the same instance ? This can cause some headaches if it's cached early...

eviljeff: it has to be a `@cached_property` so the ES serializer can set it's value based on what's in the index.

src/olympia/addons/indexers.py
eviljeff: Mmm, I thought that would be a problem in the _serializer_ not the indexer?  `_attach_fields` would try to do `value = data.get(field_name, None)` and then `setattr(obj, field_name, value)` (i.e. `setattr(addon_instance, 'is_recommended', data.get('is_recommended', None))`) and end up setting it to `None`.  `None` isn't strictly `false` I suppose... and we are returning it...

eviljeff: Maybe always cast it to `bool` in `to_representation`?

diox: Sorry, totally misread, it is indeed only a problem in the serializer, not the indexer.

src/olympia/addons/models.py
diox: Ah, I see.

src/olympia/addons/indexers.py
diox: I think it's fine as is. It can only be `None` initially... it will be `False` after we reindex...

docs/topics/api/addons.rst
```diff
@@ -175,6 +175,7 @@ This endpoint allows you to fetch a specific add-on by id, slug or guid.
     :>json boolean is_disabled: Whether the add-on is disabled or not.
     :>json boolean is_experimental: Whether the add-on has been marked by the developer as experimental or not.
     :>json boolean is_featured: The add-on appears in a featured collection.
+    :>json boolean is_recommended: The add-on is recommended by Mozilla.
```
diox: Add it to changelog

eviljeff: https://github.com/mozilla/addons-server/pull/11275/commits/0d23c4147daf414a2057bb574d3a487413dae91b

PR #11310, opened 2019-05-02
Add rate-limiting to add-on signing API

### LINKED ISSUE #11287: Add rate-limiting to the submission process
Add-on submission process doesn't have rate-limiting inside the app, it's done by ops at the nginx level instead. We should move it to inside addons-server to have finer grained control over it like we do for other APIs.

We should rely on DRF's throttling classes for that, with the catch that it needs to work with devhub submissions as well, which are not API based. We probably want to modify:
- `olympia.signing.views.VersionView` (`put` and `post`)
- `olympia.devhub.views.upload_for_version`
- `olympia.devhub.views.submit_version_upload`

diox:
Fixes #11287

### DIFF
--- a/src/olympia/api/throttling.py
+++ b/src/olympia/api/throttling.py
@@ -1,7 +1,9 @@
 import re
+import time
 
 from django.conf import settings
 
+from rest_framework.permissions import SAFE_METHODS
 from rest_framework.throttling import UserRateThrottle
 
 
@@ -11,7 +13,23 @@ from rest_framework.throttling import UserRateThrottle
 
 
 class GranularUserRateThrottle(UserRateThrottle):
+    """
+    Throttling class that works like DRF's UserRateThrottle but supports
+    granular rates like 1/5second, and can be deactivated through a
+    API_THROTTLING django setting.
+
+    Its scope defaults to `user` but in most cases we'll want the child class
+    to override that and add a custom rate.
+    """
+
     RATE_REGEX = r'(?P<num>\d+)\/(?P<period_num>\d{0,2})(?P<period>\w)'
+    timer = time.time
+
+    def __init__(self):
+        super().__init__()
+        # Re-initialize timer at __init__() to allow freeze_gun.freeze_time()
+        # to work properly in tests.
+        self.timer = time.time
 
     def allow_request(self, request, view):
         if settings.API_THROTTLING:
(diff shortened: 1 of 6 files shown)


src/olympia/api/throttling.py
```diff
@@ -11,7 +13,23 @@
...
+    """
+
     RATE_REGEX = r'(?P<num>\d+)\/(?P<period_num>\d{0,2})(?P<period>\w)'
+    timer = time.time
+
+    def __init__(self):
+        super().__init__()
+        # Re-initialize timer at __init__() to allow freeze_gun.freeze_time()
+        # to work properly in tests.
+        self.timer = time.time
```
EnTeQuAk: But then you don't really need the class-property `timer` as subclasses will have to overwrite `__init__` anyway.

diox: I don't understand your point. Subclasses don't overwrite `__init__` - without that patch, they only have the class property, with it they have an instance one. That makes all the difference with `freeze_gun`.

src/olympia/api/throttling.py
```diff
@@ -28,3 +46,32 @@ def parse_rate(self, rate):
...
+    scope = 'anon'
+
+    def get_cache_key(self, request, view):
+        return self.cache_format % {
+            'scope': self.scope,
+            'ident': self.get_ident(request)  # This will get the IP.
+        }
+
+
+class ThrottleOnlyUnsafeMethodsMixin():
```
eviljeff: I can't help thinking this should be a wrapper class, or a class variable (e.g. `unsafe_methods_only = False` by default, that can be overridden by subclasses) in `GranularUserRateThrottle`.  But personal opinion and not a blocker given the issue priority.

src/olympia/lib/settings_base.py
```diff
@@ -1816,6 +1816,8 @@ def get_raven_release():
 
 MIGRATED_LWT_UPDATES_ENABLED = True
 
+ADDON_UPLOAD_RATE_LIMITS_BYPASS_EMAILS = ('addons+release@mozilla.org',)
```
eviljeff: Just confirming this email is correct because I've not seen it before.  (Maybe it should be a permission group we add users to, rather than a hardcoded email address?)

diox: Was confirmed with @Callek on slack. I thought about doing a group+permission but I kinda wanted to restrict it more, I'd like to make sure the list is always under control. I don't see any valid use cases other than langpacks submission for new Firefox releases atm.

src/olympia/api/throttling.py
diox: Yeah I can do that in a follow-up. Was trying not to impact other existing views for now.

PR #11325, opened 2019-05-03
Fix abuse report type for unknown add-on types

### LINKED ISSUE #11305: 500 Internal Server Error when a 3 digit number is typed in the search field
Followup for https://github.com/mozilla/addons-server/issues/11133

STR:
1. Go to  /models/abuse/abusereport/ on AMO dev 
2. Type in the search field "123" or "12a" or "559" then click the Search button 
3. Observe the page

Expected result:
Eventually no results are returned.

Actual result:
500 Internal Server Error. 

Notes:
Reproducible on AMO dev - FF66(Win10). 

In the Search field

500 server error for 3 digit numbers

I reproduced this issue every time with "By minimum reports count (grouped by guid)" filter when using the following scenario: I tried to search using a value (0 for the first attempt) and then I changed the value to "69" and clicked "Submit".

500 server error by minimum reports count

diox:
Fixes #11305

### DIFF
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -210,8 +210,12 @@ class AbuseReport(ModelBase):
     @property
     def type(self):
         with translation.override(settings.LANGUAGE_CODE):
-            type_ = (translation.ugettext(amo.ADDON_TYPE[self.addon.type])
-                     if self.addon else 'User' if self.user else 'Addon')
+            if self.addon and self.addon.type in amo.ADDON_TYPE:
+                type_ = (translation.ugettext(amo.ADDON_TYPE[self.addon.type]))
+            elif self.user:
+                type_ = 'User'
+            else:
+                type_ = 'Addon'
         return type_
 
     def __str__(self):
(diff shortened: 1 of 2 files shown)


src/olympia/abuse/models.py
```diff
@@ -210,8 +210,12 @@ def target(self):
     @property
     def type(self):
         with translation.override(settings.LANGUAGE_CODE):
-            type_ = (translation.ugettext(amo.ADDON_TYPE[self.addon.type])
-                     if self.addon else 'User' if self.user else 'Addon')
+            if self.addon and self.addon.type in amo.ADDON_TYPE:
+                type_ = (translation.ugettext(amo.ADDON_TYPE[self.addon.type]))
```
eviljeff: 🤔 why _do_ we have this literal `translation.ugettext(amo.ADDON_TYPE[self.addon.type])` rather than `amo.ADDON_TYPE[self.addon.type]`?

diox: To ensure it's... *not* translated in whatever the current language is in the current thread but instead rendered in english (it's inside a `translation.activate()`, the tests cover that)

eviljeff: ah, missed that.

PR #11332, opened 2019-05-03
Rename .git* files on extraction

### LINKED ISSUE #10611: Add test that makes gitignore files are ignored on our side

willdurand:
Fixes #10611

---

This patch adds a list of files to rename during extraction.

### DIFF
--- a/src/olympia/lib/tests/test_git.py
+++ b/src/olympia/lib/tests/test_git.py
@@ -377,6 +377,40 @@ def test_extract_and_commit_source_from_version_no_dotgit_clash(settings):
 
 
 @pytest.mark.django_db
+def test_extract_and_commit_source_from_version_rename_dotgit_files(settings):
+    addon = addon_factory(
+        file_kw={'filename': 'webextension_no_id.xpi'},
+        version_kw={'version': '0.1'})
+
+    # Generate source file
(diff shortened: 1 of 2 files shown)


src/olympia/lib/tests/test_git.py
```diff
@@ -376,6 +376,40 @@ def test_extract_and_commit_source_from_version_no_dotgit_clash(settings):
...
+    addon = addon_factory(
+        file_kw={'filename': 'webextension_no_id.xpi'},
+        version_kw={'version': '0.1'})
+
+    # Generate source file
+    source = temp.NamedTemporaryFile(suffix='.zip', dir=settings.TMP_PATH)
+    with zipfile.ZipFile(source, 'w') as zip_file:
+        zip_file.writestr('manifest.json', '{}')
+        zip_file.writestr('.gitattributes', '')
+        zip_file.writestr('.gitignore', '')
```
kumar303: Is there a test to make sure nested `.gitignore` files are also renamed? I can see that they are but this is something we should have a regression test for since it's super important.

willdurand: I don't think we test that indeed, all files to rename could potentially be nested. Could you file an issue please?

kumar303: Sure. Filed https://github.com/mozilla/addons-server/issues/11370

PR #11353, opened 2019-05-07
Implement a new draft-comment model.

### LINKED ISSUE #11319: Add new draft-comment model
Part of #11277

A new `DraftCommentLog` model should be added to allow us to store comments that are added through an API that should be visible on the reviewer-page.

Screenshot from 2019-05-03 11-22-52

Ideally, to simplify implementation and usage, we won't need it to be linked to an ActivityLog entry. If a review action is submitted the draft comment log will be cleared.

EnTeQuAk:
This new model will be used to present draft-comments written in the
code-manager and displayed during review.

The related rest-api comes in a second step.

Fixes #11319

### DIFF
--- a/src/olympia/activity/models.py
+++ b/src/olympia/activity/models.py
@@ -162,6 +162,20 @@ class GroupLog(ModelBase):
         ordering = ('-created',)
 
 
+class DraftComment(ModelBase):
+    """A model that allows us to draft comments for reviews before we have
+    an ActivityLog instance ready.
+
+    This is being used by the commenting API by the code-manager.
+    """
+    id = PositiveAutoField(primary_key=True)
+    comments = models.TextField()
+    version = models.OneToOneField(Version, on_delete=models.CASCADE)
+
+    class Meta:
+        db_table = 'log_activity_comment_draft'
+
+
 class ActivityLogManager(ManagerBase):
     def for_addons(self, addons):
         if isinstance(addons, Addon):
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -790,8 +790,15 @@ def review(request, addon, channel=None):
             request, ugettext('Self-reviews are not allowed.'))
         return redirect(reverse('reviewers.dashboard'))
 
-    # Get the current info request state to set as the default.
-    form_initial = {'info_request': addon.pending_info_request}
+    form_initial = {
+        # Get the current info request state to set as the default.
+        'info_request': addon.pending_info_request,
+    }
+
+    try:
+        form_initial['comments_draft'] = version.draftcomment.comments
+    except Version.draftcomment.RelatedObjectDoesNotExist:
+        pass
 
     form_helper = ReviewHelper(
         request=request, addon=addon, version=version,
@@ -831,6 +838,11 @@ def review(request, addon, channel=None):
         redirect_url = reverse('reviewers.unlisted_queue_all')
 
     if request.method == 'POST' and form.is_valid():
+        try:
+            version.draftcomment.delete()
+        except Version.draftcomment.RelatedObjectDoesNotExist:
+            pass
+
         form.helper.process()
         amo.messages.success(
             request, ugettext('Review successfully processed.'))
--- a/src/olympia/migrations/1090-create-log-activity-comment-draft-table.sql
+++ b/src/olympia/migrations/1090-create-log-activity-comment-draft-table.sql
@@ -1,0 +1,12 @@
+CREATE TABLE `log_activity_comment_draft` (
+  `id` integer UNSIGNED AUTO_INCREMENT NOT NULL PRIMARY KEY,
+  `created` datetime(6) NOT NULL,
+  `modified` datetime(6) NOT NULL,
+  `comments` longtext NOT NULL,
+  `version_id` integer UNSIGNED NOT NULL
+);
+
+ALTER TABLE `log_activity_comment_draft`
+  ADD CONSTRAINT `log_activity_comment_draft_version_id_b9633528_fk_versions_id`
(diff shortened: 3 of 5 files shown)


src/olympia/reviewers/views.py
```diff
@@ -1386,6 +1400,27 @@ def retrieve(self, request, *args, **kwargs):
...
+    def draft_comment(self, request, **kwargs):
+        version = self.get_object()
+
+        if request.method == 'GET':
+            instance = get_object_or_404(DraftComment, version=version)
+            serializer = DraftCommentSerializer(instance, data=request.data)
+            return Response(serializer.data)
+        elif request.method == 'DELETE':
+            DraftComment.objects.filter(version=version).delete()
+            return Response(status=status.HTTP_202_ACCEPTED)
```
willdurand: Isn't `delete()` on a model synchronous? If so, shouldn't we return a 204?

src/olympia/activity/models.py
```diff
@@ -162,6 +162,20 @@ class Meta:
...
 
 
+class DraftComment(ModelBase):
+    """A model that allows us to draft comments for reviews before we have
+    an ActivityLog instance ready.
+
+    This is being used by the commenting API by the code-manager.
+    """
+    id = PositiveAutoField(primary_key=True)
+    comments = models.TextField()
```
willdurand: Sorry for the naive question: is bleach used somewhere so that we do not store XSS in the db?

src/olympia/reviewers/views.py
```diff
@@ -831,6 +842,9 @@ def review(request, addon, channel=None):
         redirect_url = reverse('reviewers.unlisted_queue_all')
 
     if request.method == 'POST' and form.is_valid():
+        if comments_draft:
+            comments_draft.delete()
```
eviljeff: I'd move this below `form.helper.process()` in case there's an exception thrown in there.

src/olympia/reviewers/views.py
eviljeff: you're correct, but there may not be any `DraftComment` instance for that version to delete.  (should we throw different codes depending on if you try to `DELETE` the comment for a version when there aren't any?)

src/olympia/reviewers/views.py
```diff
@@ -1386,6 +1400,27 @@ def retrieve(self, request, *args, **kwargs):
...
+
+        if request.method == 'GET':
+            instance = get_object_or_404(DraftComment, version=version)
+            serializer = DraftCommentSerializer(instance, data=request.data)
+            return Response(serializer.data)
+        elif request.method == 'DELETE':
+            DraftComment.objects.filter(version=version).delete()
+            return Response(status=status.HTTP_202_ACCEPTED)
+
+        instance, _ = DraftComment.objects.get_or_create(version=version)
```
eviljeff: I know the method is limited to `GET`, `DELETE` and `PUT` but I'd be happier if this was also in a `elif request.method == 'PUT`:` block.

src/olympia/reviewers/views.py
```diff
@@ -1386,6 +1400,27 @@ def retrieve(self, request, *args, **kwargs):
         )
         return Response(serializer.data)
 
+    @action(
+        detail=True,
+        methods=['get', 'put', 'delete'],
+        permission_classes=ReviewAddonVersionMixin.permission_classes)
```
eviljeff: This would let add-on authors delete the `DraftComment` of a reviewer?  That seems... not good opsec.

EnTeQuAk:
> What's the expected workflow here btw - the reviewer loads the code manager from the review page and writes the draft comment and then returns to the review page - would they have to reload the page to see it? (Thinking back to my reviewer days I would generally open the file viewer in a new tab)

Correct, that's the workflow for this. There may be some javascript magic on the reviewer page at some point to prefill this maybe but we should keep things simple for now. Eventually, more and more functionality will move to the code-manager and will replace functionality on the reviewer tools to avoid the two-tools-switching.

src/olympia/reviewers/views.py
EnTeQuAk: I kinda think 204 is what I originally wanted. I don't think we need a different status code for different states, e.g an instance existing vs one doesn't. `204` mostly is just a success response.

src/olympia/activity/models.py
EnTeQuAk: I checked most of the templates that render the comments and they're using `|safe`, except for `reviewers/includes/history.html` which is rendering activity log entries, indeed. Let me play around a bit with XSS and check if we should sanitize this, we probably should but given that the developer may quote code-parts and we're rendering it to HTML and email this may be tricky.

src/olympia/reviewers/views.py
EnTeQuAk: uhm, fair point, I'll limit this. Actually, fair point, should this actually be limited to the actual author of the draft comment? e.g, if two reviewers are reviewing the same version, should they produce their own draft comment instances? 

cc @wagnerand for some ideas about this too

wagnerand: > if two reviewers are reviewing the same version, should they produce their own draft comment instances?

I'd say yes.

src/olympia/reviewers/views.py
EnTeQuAk: Fixed in 2047d45

src/olympia/reviewers/views.py
EnTeQuAk: Fixed in 2047d45

src/olympia/reviewers/views.py
EnTeQuAk: Alrighty, implemented in fc8d0f1 and 80159dc

Issue #10071, opened 2018-11-26
Make add-on popularity boost in search easier to tweak


diox:
As explained in #9741, when we index add-ons, we store a special `boost` property that depends on the add-on popularity (number of users). #9741 is about investigating that boost, potentially lowering it, but in the meantime, we should try to move that boost to a query-time boost, because it would make things easier to debug - having to re-index every time you change the value is annoying.

AlexandraMoga:
As stated in #10171, validating search on -dev has become extremely difficult, so, until a better solution is found, we're going to have to wait for the changes to land in prod in order to properly verify search issues.

AlexandraMoga:
I've tried several search scenarios on prod and overall I received relevant results, unlike -dev and stage.

I'm marking this issue as verified fixed.

PR #11373, opened 2019-05-08
boost a search result based on Addon.is_recommended

### LINKED ISSUE #11066: Update search to prioritize recommended add-ons
Search should boost recommended add-ons above non-recommended content.

See the PRD for further details: https://docs.google.com/document/d/1fRxLy7h02tn1HBIlsTU_6h9DHmN8c3N6bg4JoxHF7J8/edit#heading=h.5kvm14zbadi2

### LINKED ISSUE #11410: Don't raise an exception if there isn't a `current_version` in `is_recommended()`
An add-on with no `current_version` (either purely unlisted or listed with all its versions deleted) that we're trying to determine the recommendation status of can cause us to raise an exception.

### Steps to reproduce
- Create a purely unlisted add-on, make it recommendable by adding a discovery item for it through the admin w/ the property checked.
- Also try with a purely listed add-on for which all versions are deleted after it has been approved & set as recommendable in discovery admin
- Try to index it, or try to call the detail API for that add-on w/ the right permissions 

- It should work. Notably, once indexed the add-on should no longer show up in searches if it was listed before.

eviljeff:
fixes #11066 by adding a x20 (actually x5, compared to non-recommended, non-experimental add-ons) boost to recommended add-ons.  Not sure if that's enough, but we can increase it.
+fixes #11410

### DIFF
--- a/src/olympia/search/tests/test_filters.py
+++ b/src/olympia/search/tests/test_filters.py
@@ -102,7 +102,7 @@ class TestQueryFilter(FilterTestsBase):
         assert expected in should
 
         functions = qs['query']['function_score']['functions']
-        assert len(functions) == 2
+        assert len(functions) == 3
         assert functions[0] == {
             'field_value_factor': {
                 'field': 'average_daily_users', 'modifier': 'log2p'
@@ -121,6 +121,10 @@ class TestQueryFilter(FilterTestsBase):
             },
             'weight': 4.0
         }
+        assert functions[2] ==  {
+            'filter': {
+                'term': {'is_recommended': True}},
+            'weight': 5.0}
         return qs
 
     def test_no_rescore_if_not_sorting_by_relevance(self):
--- a/src/olympia/search/tests/test_filters.py
+++ b/src/olympia/search/tests/test_filters.py
@@ -121,7 +121,7 @@ class TestQueryFilter(FilterTestsBase):
             },
             'weight': 4.0
         }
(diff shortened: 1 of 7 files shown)


diox:
Looks like I kept the field for backwards-compatibility but forgot to file an issue to then remove it later. My bad... Gotta check that we did a full reindex since that date before removing it. In any case the recommended boost should be done at query time by tweaking the scoring functions.

EnTeQuAk:
> Wait, I actually did change that to a query time boost a while ago as part of #10089 (fixing #10071). Did I forget to remove the indexer part ?

I knew I remembered something but couldn't find it so I thought I rememberd wrong. Thanks for clarifying!

src/olympia/search/tests/test_filters.py
```diff
@@ -121,6 +121,10 @@ def _test_q(self, qs):
             },
             'weight': 4.0
         }
+        assert functions[2] == {
+            'filter': {
+                'term': {'is_recommended': True}},
```
diox: Kind of an edge case but I think we need to repeat the other boosts in `functions[1]` as well to ensure we don't boost an add-on that is somehow `is_recommended` but not public ? Not sure if that's possible.

eviljeff: It's possible for a non-public add-on to be `is_recommended` but a non-public add-on wouldn't be in the search results so doesn't need bumping (that's my thinking).

PR #11384, opened 2019-05-09
Normalize SVG and BMP mimetypes

### LINKED ISSUE #11382: Normalize SVG mimetypes to image/svg+xml
Server-side issue of https://github.com/mozilla/addons-code-manager/issues/719

### LINKED ISSUE #11383: Alias `image/x-ms-bmp` mimetype to `image/bmp`
See also: https://github.com/mozilla/addons-code-manager/issues/718

---

We want to alias the `image/x-ms-bmp` mimetype to `image/bmp` so that code-manager does not maintain a huge list of supported mimetypes with all the weird existing mimetypes.

willdurand:
Fixes #11382, which should also fix https://github.com/mozilla/addons-code-manager/issues/719
Fixes #11383, which should also fix https://github.com/mozilla/addons-code-manager/issues/718

---

I found the BMP examples in http://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html (found this link in http://fileformats.archiveteam.org/wiki/BMP).

The SVG content has been extracted from the original SVG that is causing issues. This is how libmagic detects it: https://github.com/file/file/blob/31a82018c2ed153b84ea5e115fa855de24eb46d1/magic/Magdir/sgml#L12-L13.


src/olympia/reviewers/tests/test_serializers.py
```diff
@@ -324,6 +324,12 @@ def test_serialize_deleted_file(self):
         (MagicMock(type='blob'), 'foo.json', 'text', 'application/json'),
         (MagicMock(type='tree'), 'foo', 'directory',
                                  'application/octet-stream'),
+        (MagicMock(type='blob'), 'image-svg-without-xml.svg', 'image',
+                                 'image/svg+xml'),
+        (MagicMock(type='blob'), 'bmp-v3.bmp', 'image', 'image/bmp'),
+        (MagicMock(type='blob'), 'bmp-v4.bmp', 'image', 'image/bmp'),
+        (MagicMock(type='blob'), 'bmp-v5.bmp', 'image', 'image/bmp'),
```
EnTeQuAk: There's different versions? :rofl:

willdurand:
@EnTeQuAk I have one test case that passes locally but it fails on Travis CI: https://travis-ci.org/mozilla/addons-server/jobs/530193027#L2688-L2692.

It's weird because, according to the test case, the mimetype is asserted first and we expect `image/svg+xml`, so why would the category be `text` and not `image`? 🤔

EnTeQuAk:
> It's weird because, according to the test case, the mimetype is asserted first and we expect image/svg+xml, so why would the category be text and not image? thinking

That's hard to say, the whole compat mapping really is there so that tests on travis work properly - at least that was it's initial implementation idea.

Travis uses an older version of `file` and as such a different mime-mapping.

It could be, that the file is detected as XML and as such declared as `text/xml` maybe?

willdurand:
> It could be, that the file is detected as XML and as such declared as `text/xml` maybe?

That's what I think too, but why would the assertion on mimetype pass?

willdurand:
> That's what I was thinking too, but why would the assertion on mimetype pass?

So we concluded that the mimetype was maybe detected as `text/*` and then a completely new mimetype was returned by `mimetypes.guess_type()`. That would explain the correct mimetype value with the wrong category. Last commit changes the logic to only second guess `text/plain`.

PR #11402, opened 2019-05-13
Return diff as an object or null, not an array

### LINKED ISSUE #11392: Don't return an array for `diff` in Compare API response
The [response to the Compare endpoint](https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#compare) returns a single diff, in a property called `diff`, but that actually contains an array, which only ever contains a single diff. So let's change it to just return the diff object in the `diff` property, rather than an array with a single diff.

willdurand:
Fixes #11392

---

I am not sure what the best Python idiom is for returning the first item of a list or `None`.

### DIFF
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -307,6 +307,10 @@ class FileEntriesDiffSerializer(FileEntriesSerializer):
             parent=parent,
             pathspec=[self.get_selected_file(obj)])
 
+        # Because we're always specifying `pathspec` with the currently
+        # selected file we can inline the diff because there will always be
+        # one.
+        # See: https://github.com/mozilla/addons-server/issues/11392
         return next(iter(diff), None)
 
     def get_entries(self, obj):
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -307,7 +307,7 @@ class FileEntriesDiffSerializer(FileEntriesSerializer):
             parent=parent,
             pathspec=[self.get_selected_file(obj)])
 
-        return diff
+        return next(iter(diff), None)
 
     def get_entries(self, obj):
         """Overwrite `FileEntriesSerializer.get_entries to inject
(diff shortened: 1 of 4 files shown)


src/olympia/reviewers/serializers.py
```diff
@@ -307,7 +307,7 @@ def get_diff(self, obj):
             parent=parent,
             pathspec=[self.get_selected_file(obj)])
 
-        return diff
+        return next(iter(diff), None)
```
EnTeQuAk: I think we should be implementing this in `lib/git::get_diff` instead to make sure we implement this as low-level as possible. Make sure `changes` is a dictionary and render the patch datastructure in there.

Additionally, raise an error in case we hit an edge-case somewhere where it's not the case anymore.

I looked at various diffs and couldn't find a scenario tbh, technically I'm still not 100% sure if there aren't any edgecases I'm not thinking about. But neither for renames, copies or others the diffs aren't issued more than once per file.

willdurand: @EnTeQuAk the `lib/git::get_diff` returns a diff between two commits, which means diffs for different files. Specifying `pathspec` helps focus on one specific file but I believe `pathspec` could also be a partial path like a directory name.

In the serializer, we pass the file name to `lib/git::get_diff` to make sure we retrieve only one diff for this specific file. This is why we can't have more than one diff in the list returned by `lib/git::get_diff`.

I think `lib/git::get_diff` does the right thing and has the correct API/return value. I don't think we should change the return type to be an array, `None` or a single item depending on what's supplied to this function because that'd be consistent.

If we want to implement this in `lib/git`, then I'd rather suggest a `get_diff_for_file` function that calls `get_diff` and returns `next(iter(diff), None)`.

EnTeQuAk: You're right, I forgot that this is only because we are specifying `pathspec`.

Maybe add a comment before the `return` mentioning that? Something like, "Because we're always specifying pathspec with the currently selected file we can inline the diff because there will always be one." or so.

With that, I think we should be okay

willdurand:
@bobsilverberg this is ready, feel free to land it when you need it.

eviljeff:
Needs an entry on the API changelog list (in overview.rst)

PR #11435, opened 2019-05-15
themes in the featured collection should be seen as recommended too

### LINKED ISSUE #11429: Have feature themes collection membership be treated as recommended
(At least at first) themes aren't going to go through the same process as extensions to get recommended - they don't have code so don't need a code review so we'll reuse the existing way featured themes are selected (by membership of special collection)

eviljeff:
fixes #11429

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -1270,15 +1270,22 @@ class Addon(OnChangeMixin, ModelBase):
 
     @cached_property
     def is_recommended(self):
+        from olympia.bandwagon.models import CollectionAddon
         from olympia.discovery.models import DiscoveryItem
 
         try:
             item = self.discoveryitem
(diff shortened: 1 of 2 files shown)


diox:
Why can't they be added to the disco items like normal ? If we do this I'm worried there will be a disconnect between AMO and Firefox: they are not going to get the signature bits, so they aren't going to be shown as recommended in Firefox itself.

eviljeff:
@diox a question to bring up with Jorge (in the issue or PRD) - he wanted recommended themes to still be a collection.

jvillalobos:
The need to have the collection is that the featured/recommended themes list is maintained by relatively low-tech community members. Also, since that list will be managed differently and will be cycled regularly, it doesn't make sense to add any of the review / signing bits to it. It will be inconsistent either way. Not showing the badge for installed themes in about:addons is okay.

PR #11436, opened 2019-05-15
waffle the search boost for recommended behind 'api-recommendations-priority'

### LINKED ISSUE #11430: Put the backend recommendation API changes behind a waffle
Some of the things we'll be doing, like changing ranking and only exposing recommended add-ons in certain places, are only going to work when we have all the work done and have all the add-ons we want to recommended already reviewed, so we should put the changes behind a waffle.

eviljeff:
fixes #11430

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -1816,6 +1816,7 @@ class TestAddonSearchView(ESTestCase):
         self.refresh()
         query = (u'남포역립카페추천 ˇjjtat닷컴ˇ ≡제이제이♠♣ 남포역스파 '
                  u'남포역op남포역유흥≡남포역안마남포역오피 ♠♣')
+        assert switch_is_active('api-recommendations-priority')
         data = self.perform_search(self.url, {'q': query})
         # No results, but no 500 either.
         assert data['count'] == 0
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -963,6 +963,8 @@ class TestAddonSearchView(ESTestCase):
         self.url = reverse_ns('addon-search')
         self.create_switch('return-to-amo', active=True)
         switch_is_active('return-to-amo')
+        self.create_switch('api-recommendations-priority', active=True)
+        switch_is_active('api-recommendations-priority')
 
     def tearDown(self):
         super(TestAddonSearchView, self).tearDown()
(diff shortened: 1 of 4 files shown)


eviljeff:
No idea about the test fail - it doesn't fail locally 😿

src/olympia/addons/tests/test_views.py
```diff
@@ -963,6 +963,8 @@ def setUp(self):
         self.url = reverse_ns('addon-search')
         self.create_switch('return-to-amo', active=True)
         switch_is_active('return-to-amo')
+        self.create_switch('api-recommendations-priority', active=True)
+        switch_is_active('api-recommendations-priority')
```
EnTeQuAk: Did you forget an `assert`?

src/olympia/addons/tests/test_views.py
```diff
@@ -1826,6 +1829,8 @@ class TestAddonAutoCompleteSearchView(ESTestCase):
 
     def setUp(self):
         super(TestAddonAutoCompleteSearchView, self).setUp()
+        self.create_switch('api-recommendations-priority', active=True)
+        switch_is_active('api-recommendations-priority')
```
EnTeQuAk: Missing the `assert` too I think

src/olympia/search/filters.py
```diff
@@ -736,19 +736,24 @@ def apply_search_query(self, search_query, qs, sort=None):
...
+            ]
+        else:
+            recommendation_functions = []
 
         # Assemble everything together
         qs = qs.query(
             'function_score',
             query=query.Bool(should=primary_should + secondary_should),
-            functions=functions)
+            functions=functions + recommendation_functions)
```
EnTeQuAk: Instead of this, you could just do `functions.append(...)` after the `if switch_is_active` maybe?

src/olympia/addons/tests/test_views.py
eviljeff: no, it's just to check it so it's cached and the query doesn't show in when we're counting the number of db queries during ES related API calls.

eviljeff: see the lines above

src/olympia/addons/tests/test_views.py
eviljeff: same

src/olympia/search/filters.py
eviljeff: I thought it was clearer to use `+` but can do

eviljeff: https://github.com/mozilla/addons-server/pull/11436/commits/53e9f1a467932369904c5256cac664d79490558b

Issue #11329, opened 2019-05-03
0 and negative values are allowed when using "By minimum reports count" filter


ioanarusiczki:
Followup for https://github.com/mozilla/addons-server/issues/11133

STR;
1. Go to admin abuse reports /en-US/admin/models/abuse/abusereport/
2. Scroll down to "By minimum reports count" filter and type in 0 then a negative value

Expected result:
0 or negative values are not allowed. 

Actual result:
Negative values can be submitted and return results. 

Notes:
Reproducible in admin tool dev with FF66 (Win10). 

![0 and negative](https://user-images.githubusercontent.com/33448286/57137972-14186b00-6dba-11e9-88d9-d554816285a6.gif)

ioanarusiczki:
@diox I verified it on admin tools dev  - it won't allow negative values. 
![negative is not allowed](https://user-images.githubusercontent.com/33448286/58154775-19f7c280-7c7b-11e9-808d-5797f35d5cc6.png)


0 can be submitted but won't have an effect on the number of results displayed. It has the same effect as if I'd use 1.  I don't see it as a problem. Let me know if it's not ok from your point of view.

Issue #11308, opened 2019-05-02
User abuse reports - search results should be returned when searching by the user name or user id 


ioanarusiczki:
Followup for https://github.com/mozilla/addons-server/issues/11133

STR:
1. Go on dev to /en-US/admin/models/abuse/abusereport/
2. Filter the reports by type: users
3. Try to see the reports made for one user only by typing in the user's name (or user's id)

Expected:
Reports for the specified user are listed.

Actual result:
No results found. 

Notes:
For an addon the abuse reports are displayed when searching by the addon's name, slug or guid.
Reproducible on admin tool dev - FF66(Win10). 

![search using a users name](https://user-images.githubusercontent.com/33448286/57083373-1ec2f980-6d01-11e9-8175-c56ec3b77b8f.gif)

diox:
Yeah, I should be smarter about this. It should totally depend on the type selected, there is no need to search through add-on fields if users type is selected, and vice-versa.

ioanarusiczki:
@diox Here are my test results on django admin dev - FF67(Win10). 
- searching by id or GUID works for both users and addons, I did not find problems.
- searching by Name for Users doesn't really work, at least not with most of the Names I tested with. In many cases, if I type in the complete names I get 0 search results.
I'll give examples for this situation:
BobsDisplayName 
UserWithNoPerms 
LexaDev
IoanaTestingStageandDevlongstringlong
Itsy Bitsy Spider
Amoga-dev 
Jorge Villalobos 

These are some examples of Names that returned the expected result:
CosminSV
kumar_dev2 
testemaildev 

- searching by Name addons works well in most of the cases. Still, if a name is longer it will return less results than it should.
Examples 
"X-notifier (for Gmail,Hotmail,Yahoo,AOL ...)" ->130 search results using the id or GUID then 13 using it's complete Name.
"Awesome Screenshot Plus - Capture, Annotate & More"-> 122 search results using the id down to 61 when I try with the complete Name.
"Ecosia - The search engine that plants trees!" -> I get 5 results if I type in just "Ecosia" (or search it by GUID or ID), and 1 result if I use it's complete Name.

Issue #11307, opened 2019-05-02
Search by add-on id in abuse reports admin is not returning results


ioanarusiczki:
Followup for https://github.com/mozilla/addons-server/issues/11133

> When searching, the following fields should be available for filtering:
> - Reports for a specific add-on (by id, name, or slug)

STR:
1. Load AMO dev admin tools via /en-US/admin/models/abuse/abusereport/
2. Try to make a search using an add-ons id (I tried with 6 digit ids such as 502857 or 616681)
3. Observe the results

Expected:
The addon abuse report is available.

Actual result:
404 not found

Notes:
Search can be made using the name, guid or slug of an add-on. 

![search using an id](https://user-images.githubusercontent.com/33448286/57082262-110c7480-6cff-11e9-85b9-851233bd7635.gif)

diox:
It's trivial but I forgot to implement this.

diox:
>  Is this alright or the Search should appear for "All" too (perhaps to search by Name if not ID).

It's by design yes, to limit performance issues with search.

PR #11438, opened 2019-05-15
Make Abuse Report Admin search depend on report type being searched

diox:
Fixes https://github.com/mozilla/addons-server/issues/11329
Fixes https://github.com/mozilla/addons-server/issues/11308
Fixes https://github.com/mozilla/addons-server/issues/11307

### DIFF
--- a/src/olympia/abuse/admin.py
+++ b/src/olympia/abuse/admin.py
@@ -191,7 +191,7 @@ class AbuseReportAdmin(CommaSearchInAdminMixin, admin.ModelAdmin):
         if type_ == 'addon':
             search_fields = (
                 'addon__name__localized_string', 'addon__slug', 'addon_name',
-                'guid', 'message', 'addon__id',
+                '=guid', 'message', 'addon__id',
             )
         elif type_ == 'user':
             search_fields = (
--- a/src/olympia/abuse/admin.py
+++ b/src/olympia/abuse/admin.py
@@ -212,7 +212,7 @@ class AbuseReportAdmin(CommaSearchInAdminMixin, admin.ModelAdmin):
         elif type_ == 'user':
             search_field = 'user_id'
         else:
-            search_field = 'pk'
+            search_field = super().get_search_id_field(request)
         return search_field
 
     def get_search_results(self, request, qs, search_term):
--- a/src/olympia/abuse/admin.py
+++ b/src/olympia/abuse/admin.py
@@ -12,6 +12,7 @@ from rangefilter.filter import (
 from olympia import amo
 from olympia.access import acl
 from olympia.addons.models import Addon
+from olympia.amo.admin import CommaSearchInAdminMixin
 from olympia.translations.utils import truncate_text
 
 from .models import AbuseReport
@@ -146,7 +147,7 @@ class DateRangeFilter(FakeChoicesMixin, DateRangeFilterBase):
         yield all_choice
 
 
-class AbuseReportAdmin(admin.ModelAdmin):
+class AbuseReportAdmin(CommaSearchInAdminMixin, admin.ModelAdmin):
     actions = ('delete_selected', 'mark_as_valid', 'mark_as_suspicious')
     date_hierarchy = 'modified'
     list_display = ('target_name', 'guid', 'type', 'state', 'distribution',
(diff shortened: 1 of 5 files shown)


src/olympia/abuse/admin.py
```diff
@@ -185,6 +183,38 @@ def get_actions(self, request):
...
 
+    def get_search_fields(self, request):
+        """
+        Return search fields according to the type filter.
+        """
+        type_ = request.GET.get('type')
+        if type_ == 'addon':
+            search_fields = (
+                'addon__name__localized_string', 'addon__slug', 'addon_name',
+                '=guid', 'message', 'addon__id',
```
eviljeff: is `addon_id` more efficient than `addon_id` (and should it be `'=addon_id'` or `'addon_id__exact'`?)

src/olympia/abuse/admin.py
```diff
@@ -185,6 +183,38 @@ def get_actions(self, request):
...
+            )
+        elif type_ == 'user':
+            search_fields = (
+                'message', 'user__id', '^user__username', '^user__email',
```
eviljeff: similar question with `user__id`

src/olympia/abuse/admin.py
```diff
@@ -185,6 +183,38 @@ def get_actions(self, request):
...
+        Return the field to use when all search terms are numeric, according to
+        the type filter.
+        """
+        type_ = request.GET.get('type')
+        if type_ == 'addon':
+            search_field = 'addon_id'
+        elif type_ == 'user':
+            search_field = 'user_id'
+        else:
+            search_field = 'pk'
```
eviljeff: I don't suppose it matters, but `super().get_search_id_field(request)` instead when type isn't `addon` or `user`?

src/olympia/abuse/admin.py
diox: Yes, but that doesn't work :( Django doesn't let me, and I didn't want to add even more custom code to `get_search_fields()` (I originally had, but then with the other recent changes that was a little too much in this method).

I haven't looked at the consequences of using `addon__id` besides the fact it's working but hopefully the resulting query is not that bad.

src/olympia/abuse/admin.py
diox: same

src/olympia/abuse/admin.py
diox: it doesn't exist, it's a custom method

src/olympia/abuse/admin.py
eviljeff: what doesn't work?  `=` is old syntax, supported here: https://github.com/mozilla/addons-server/pull/11438/files#diff-673ddf9dd55a8432ead7781c781b0d6fR39 and `__exact` is the new syntax, supported in the same method.

src/olympia/abuse/admin.py
eviljeff: You're overriding the mixin `get_search_id_field` method - why wouldn't it exist? https://github.com/mozilla/addons-server/pull/11438/files#diff-673ddf9dd55a8432ead7781c781b0d6fR10

src/olympia/abuse/admin.py
diox: I'm talking about using `addon_id`. It raises an exception, because it's trying to use the field as a string instead of an id. Weirdly, it works with `addon__id` and `id` - django docs warn that this is meant for text search originally... 

I don't think I need `=`/`exact` with it but I'll check. I did try and it didn't solve the original issue (brokeness with `addon_id`)

src/olympia/abuse/admin.py
diox: Ah, sorry, yeah, I can call `super()` and get the mixin implementation yes. Misread where the comment was.

src/olympia/abuse/admin.py
diox: - With `=addon__id`, it seems to be working
- With `addon__id__exact`, `TestAbuse.test_search_addon` fails with `ValueError: invalid literal for int() with base 10: 'sterious'`
- With `=addon_id`, that test fails with `django.core.exceptions.FieldError: Related Field got invalid lookup: iexact`
- With `addon_id__exact`, that test fails with `ValueError: invalid literal for int() with base 10: 'sterious'`

Similar results with `user_id` (using `TestAbuse.test_search_user`).

Conclusion, I'm going to keep traversing the relation, but use the `=` prefix.

diox: Note that as soon as we have a list of search terms (more than 1 term), all numeric, we switch to the `get_search_id_field()` where we don't have that problem.

Issue #11442, opened 2019-05-15
Admin tool for blocked email addresses


jvillalobos:
From the [Account Submission Restrictions PRD](https://docs.google.com/document/d/1FfFcydrcm_TC4sQHXWQfJg2_11RvvYUANDy0hqMlH7U/edit#).

We need an admin tool that lets AMO admins add email addresses to be blocked. It should be possible to add individual email addresses or entire email domains (`@somethingbad.com`, for instance).

If it's easy, it should also be possible to uses masks (`*@*.somethingbad.com`, for example). Otherwise it can be done in a followup.

AlexandraMoga:
@EnTeQuAk @diox which admin permissions are required to access the Email/IP blocking models?
So far I have tested with a full admin account, but maybe I should also check with more specialized permissions?

diox:
@AlexandraMoga once my fix for https://github.com/mozilla/addons-server/issues/11497 goes through (it's making its way to stage as I write this) it'll require 'Admin:Advanced'.

AlexandraMoga:
@diox users with `Admin:Advanced` can edit an entry from the blocked emails/IPs list, but they cannot deleted or add items. Was this intended?

diox:
Oversight on my part :(

diox:
I'll fix and cherry-pick that as well.

PR #11465, opened 2019-05-17
First part of adding user restriction model and basic admin.

### LINKED ISSUE #11440: Admin tool for blocked IP addresses

We need an admin tool that lets AMO admins add IP addresses to be blocked. It should be possible to add individual IP addresses. 

If it's easy, it should also be possible to add masks to the addresses (`64.10.*.*`, for example). Otherwise it can be done in a followup.

EnTeQuAk:
This adds a `UserRestriction` model that allows admins to add
restrictions based on ip-address, cidr network or email.

The admin is currently very basic and will most certainly be improved
over time.

Fixes #11440 (already contains CIDR support)
Refs #11442 (missing email domain support, going to push that in a
follow-up)

I was originally thinking about adding advanced database lookups for CIDRs but that turned out to make the patch explode so that's going to land in a separate PR, same as email domain support.


diox:
For lookups, I wonder how many rules there will be in the database. Maybe fetching them all and then comparing one by one isn't so bad ?

diox:
One thing I was expecting in terms of models was one model per type of restriction (one for ip, one for emails), allowing different permissions for each in the admin, and then only one field in them (you don't need both `ip_address` and `network`).

Then, each model would provide a `allow_request(request)` method as well as a `get_error_message()` method, and the submission code would just iterate on all known models (so 2 atm), ask them if the request is allowed, and return the appropriate error message if it isn't.

EnTeQuAk:
> For lookups, I wonder how many rules there will be in the database. Maybe fetching them all and then comparing one by one isn't so bad ?

Which is what we do with `DeniedName` already - which is actually cached. Maybe I agree, it won't be so bad but an actual database lookup won't be either and isn't too much to bother. It's a nice to have and optimization later.

To keep in mind, we're doing that check over and over again, for quite a few operations so it sums up.

> One thing I was expecting in terms of models was one model per type of restriction (one for ip, one for emails), allowing different permissions for each in the admin, and then only one field in them (you don't need both ip_address and network).

Kinda like form validators are being iterated over, yeah, I like it. I didn't think about the permissions but it makes sense, especially when considering that blocking whole networks shouldn't just be done by anyone.

I'll change the PR and throw it your way for review in a bit.

diox:
> To keep in mind, we're doing that check over and over again, for quite a few operations so it sums up.

tbh I don't care that much if it's a bit slow. It's only in the uploads, and there is rate-limiting.

diox:
I refactored my branch following that idea, completely untested right now but it will give you an idea of what I'm heading towards: https://github.com/diox/olympia/commit/79efcfeafc93ae8de09a744947ab9b26ad8af248

EnTeQuAk:
@diox refactored my work in 117fedf as well, I haven't played around much with the admin yet and will probably add a bit more functionality once I do but it should be working for now.

src/olympia/users/models.py
```diff
@@ -637,23 +638,79 @@ def fetch_names():
...
+            remote_addr = ipaddress.ip_address(request.META.get('REMOTE_ADDR'))
+        except ValueError:
+            # If we don't have a valid ip address, let's deny
+            # TODO: Verify this is what we want…
+            return False
+
+        restrictions = IPNetworkUserRestriction.objects.all()
+
+        for restriction in restrictions:
+            if remote_addr in list(restriction.network.hosts()):
```
diox: You shouldn't need to list the hosts. `remote_addr in restriction.network` should work

src/olympia/users/models.py
```diff
@@ -637,23 +638,79 @@ def fetch_names():
         return any(n in name for n in blocked_list)
 
 
-@python_2_unicode_compatible
-class UserRestriction(ModelBase):
+class IPAddressUserRestriction(ModelBase):
```
diox: We don't need both `IPAddressUserRestriction` and `IPNetworkUserRestriction`, the former is just a special case of the latter. If an IP is entered in the admin we can convert it to a network automatically by just making it a `/32`

src/olympia/users/models.py
diox: (Bonus point if using `in` is that it works with `/32` networks, where as `hosts()` doesn't seem to)

src/olympia/users/models.py
```diff
@@ -637,23 +638,79 @@ def fetch_names():
...
+            if remote_addr in list(restriction.network.hosts()):
+                return False
+        return True
+
+    @classmethod
+    def get_error_message(self, request):
+        return IPAddressUserRestriction.get_error_message()
+
+
+class EmailUserRestriction(ModelBase):
```
diox: Similarly to what I'm suggesting above eventually this should hold both domain (w/ wildcard) and email restrictions.

src/olympia/users/models.py
EnTeQuAk: Interesting, I'll try that out and add a test.

src/olympia/users/models.py
EnTeQuAk: I kinda thought we wanted to split that up to simplify permission management? We could split it in the admin anyway I think…

diox: We do want different permissions for email and ip bans, but there is no need to differentiate between ip and network, or email and domain, respectively.

src/olympia/users/models.py
diox: Add a couple tests for /24 and /32 networks just in case. When we have time to make the admin a little more fancy we can play with showing how many hosts would be affected, that kind of thing (maybe display like `ipcalc` does, with `HostMin` and `HostMax`, etc)

PR #11475, opened 2019-05-20
Add stat for account creation

### LINKED ISSUE #11474: Add a metric for account creation
So that we have a better understanding of account creations which is wrapped as part of the login view, let's add an explicit metric for that.

muffinresearch:
Fixes #11474


src/olympia/accounts/views.py
```diff
@@ -135,6 +136,7 @@ def register_user(sender, request, identity):
     user = UserProfile.objects.create_user(
         email=identity['email'], username=None, fxa_id=identity['uid'])
     log.info('Created user {} from FxA'.format(user))
+    statsd.incr('accounts.account_created_from_fxa');
```
diox: Extra `;` at the end, you've done too much JavaScript :)

muffinresearch: Hah good catch! 😅

Issue #11464, opened 2019-05-16
unhide_disabled_files command stuck and won't finish itself


bqbn:
### Describe the problem and steps to reproduce it:

Addons-server has an hourly command called "unhide_disabled_files" as per https://github.com/mozilla/addons-server/blob/master/scripts/crontab/crontab.tpl#L17

But this command seems to stuck and won't finish. For example,

```
$ sudo docker ps
CONTAINER ID        IMAGE                                                    COMMAND                  CREATED             STATUS              PORTS               NAMES
ce82f609045f        hub.prod.mozaws.net/pipelines/addons-server:2019.05.16   "python -W ignore::D…"   About an hour ago   Up About an hour                        mystifying_neumann
bb633f3e70f6        hub.prod.mozaws.net/pipelines/addons-server:2019.05.16   "python -W ignore::D…"   2 hours ago         Up 2 hours                              silly_engelbart
52657d081064        hub.prod.mozaws.net/pipelines/addons-server:2019.05.16   "python -W ignore::D…"   3 hours ago         Up 3 hours                              quirky_curran
f07194159c61        hub.prod.mozaws.net/pipelines/addons-server:2019.05.16   "python -W ignore::D…"   4 hours ago         Up 4 hours                              adoring_wilson
92522dc693a7        hub.prod.mozaws.net/pipelines/addons-server:2019.05.16   "python -W ignore::D…"   5 hours ago         Up 5 hours                              agitated_brown
1a8bc3708e8c        hub.prod.mozaws.net/pipelines/addons-server:2019.05.16   "python -W ignore::D…"   6 hours ago         Up 6 hours                              cranky_vaughan
817d5ecf9bda        hub.prod.mozaws.net/pipelines/addons-server:2019.05.16   "python -W ignore::D…"   7 hours ago         Up 7 hours                              trusting_hugle
```

The oldnest one has been running for 7 hours as of this time.

### What happened?

Not sure.

We tried to run it manually, and it stuck as well, and there was no output so it's hard to tell what's happening.

### What did you expect to happen?

The command should finish, and ideally within an hour, because it is an hourly cron job.

### Anything else we should know?
(Please include a link to the page, screenshots and any relevant files.)

bqbn:
I think it has to do with the fact that the `guarded-addons` directory contains too many sub-directories,

```
drwxrwxr-x  327212 olympia olympia  8656896 May 17 00:55 guarded-addons
```

[...]

PR #11476, opened 2019-05-20
Try to simplify hide_disabled_files/unhide_disabled_files crons

diox:
Naive attempt at fixing #11464 

We don't care about what files are currently in storage in guarded-addons, so we shouldn't need to walk through the directory.


diox:
I wouldn't rush on making them more DRY. Note that the queryset also has a subtle difference, so we'd need 2 params anyway... But more importantly we may end up having to add more complexity for edge cases...

PR #11525, opened 2019-05-28
Add command to delete ratings caused by armagadd-on

### LINKED ISSUE #11348: Delete negative ratings caused by Armagadd-on 2.0
Let's delete all reviews and ratings under 4 stars, starting with ID: 1327807. [We're still getting lots of them](https://sql.telemetry.mozilla.org/queries/62630/source), so we still have to wait abit to figure out where to finish.

/cc @atsay

diox:
Fixes #11348

### DIFF
--- a/src/olympia/addons/management/commands/process_addons.py
+++ b/src/olympia/addons/management/commands/process_addons.py
@@ -18,6 +18,9 @@ from olympia.addons.tasks import (
 from olympia.amo.utils import chunked
 from olympia.devhub.tasks import get_preview_sizes, recreate_previews
 from olympia.lib.crypto.tasks import sign_addons
+from olympia.ratings.tasks import (
+    delete_armagaddon_ratings_for_addons, get_armagaddon_ratings_filters
+)
 from olympia.reviewers.tasks import recalculate_post_review_weight
 from olympia.versions.compare import version_int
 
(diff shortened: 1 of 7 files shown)


diox:
It will delete all ratings between those dates *below* 4 stars. The issue is not about deleting specified ids but deleting all ratings below 4 stars created before and after the 2 IDs given, I merely translated the IDS into datetimes to allow to easier testing by QA on dev/stage.

EnTeQuAk:
> It will delete all ratings between those dates below 4 stars. The issue is not about deleting specified ids but deleting all ratings below 4 stars created before and after the 2 IDs given, I merely translated the IDs into datetimes to allow to easier testing by QA on dev/stage.

Shame on me, I misread the issue comments about ratings from specific user-ids. Nevermind then, now that works perfectly for me I think.

src/olympia/addons/management/commands/process_addons.py
```diff
@@ -147,6 +154,12 @@ def add_arguments(self, parser):
             type=int,
             help='Only apply task to the first X addon ids.')
 
+        parser.add_argument(
+            '--distinct',
+            action='store_true',
+            dest='distinct',
+            help='Apply DISTINCT() to the query.')
```
EnTeQuAk: Instead of requiring a `--distinct` on the command line I personally would prefer forwarding this as configuration via `'kwargs': {'distinct': True},` or so so that this stays an implementation detail rather than being an argument.

diox: https://github.com/mozilla/addons-server/pull/11525/commits/66c04f1d5981bdd62b031fda23d70ae2b168e5c2

PR #11527, opened 2019-05-28
search engine plugins to auto approval

### LINKED ISSUE #11073: Switch Search Engine add-ons to Auto-Approval
Part of [Review Flow for Recommended Extensions](https://docs.google.com/document/d/1kqDR2KHfbgLmaVZ-b-HgbSgpwU6eKA_D9voQkJ9M3io/edit#).

I believe these two add-on types currently require pre-review. Let's switch them to auto-approval, since they don't pose any significant risk and aren't getting much community activity.

eviljeff:
fixes #11073

### DIFF
--- a/src/olympia/versions/models.py
+++ b/src/olympia/versions/models.py
@@ -9,6 +9,7 @@ import django.dispatch
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.files.storage import default_storage as storage
 from django.db import models, transaction
+from django.db.models import Q
 from django.utils.encoding import force_text, python_2_unicode_compatible
 from django.utils.functional import cached_property
 from django.utils.translation import ugettext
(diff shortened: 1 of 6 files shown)


diox:
Also need to modify `Version.is_ready_for_auto_approval()` - which it used in reviewer tools to decide whether or not to fetch & display auto approval info in the review page.

eviljeff:
Ugh, apparently the logic _wasn't_ already spread around enough places 😒

diox:
Refactors welcome :) For this particular case I think we could try to remove that method entirely and always try to fetch the autoapprovalsummary - it's from a time where we still had to deal with legacy extensions.

eviljeff:
ended up making the the autoapproval query a manager method so we can query if a version would be in the queryset and determine with certainty if the version will be attempted for autoapproval.

src/olympia/versions/models.py
```diff
@@ -81,6 +82,19 @@ def latest_public_compatible_with(self, application, appversions):
             files__status=amo.STATUS_APPROVED,
         ).order_by('-created')
 
+    def auto_approve(self):
```
diox: nit: doesn't block this, but I'm not a fan of the name here, it implies that calling this will auto approve versions in the queryset. And it should have a docstring.

eviljeff: name suggestions?  `get_auto_approvable`?

diox: even `auto_approvable` would work for me I think. `Version.objects.auto_approvable()` seems good.

eviljeff: https://github.com/mozilla/addons-server/pull/11527/commits/c4f09470196bdb3002b3181213a1f625710ad228

PR #11589, opened 2019-06-05
Resign all add-ons with new COSE signature.

### LINKED ISSUE #11323: Re-sign listed add-ons with COSE signature
This bug is to cover re-signing all *listed* add-ons for FF 57+ with the updated signature that has been available for new versions uploaded since the April 4th push.

From @diox on irc:

> the command used to resign add-ons still exists, it currently it resigns *all* versions *including unlisted* of *all* add-ons that are not disabled/incomplete it sends an email to developers as well see https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/crypto/tasks.py

## Open questions:

* Does we need to re-sign all versions or just latest?
* What does the email need to say?

EnTeQuAk:
This is re-using the old sign_addons task but renames the argument
needed for `process_addons --task=` command so that we make sure the new
code won't be used by accident by some cron or so.

* Cleans up some old jetpack/legacy add-on related code along the way

Fixes #11323


src/olympia/lib/crypto/tasks.py
```diff
@@ -8,68 +8,37 @@
...
+    u'Your Firefox extension has been re-signed with a stronger signature')
 
-We recommend that you give them a try to make sure they don't have any
-unexpected problems: {addon_url}
+MAIL_COSE_MESSAGE = u'''
+Hello,
 
-If you are unfamiliar with the extension signing requirement, please read the
-following documents:
+Mozilla has recently upgraded the signing for Firefox extensions and themes to
```
diox: `extensions` is probably wrong here, that will affect themes and langpacks as well, right ? This is why the original message said `add-on`.

EnTeQuAk: It's saying "extensions and themes" but you're right, language packs are affected as well. cc @caitmuenster on this

src/olympia/lib/crypto/tasks.py
```diff
@@ -91,49 +60,58 @@ def get_new_version_number(version):
...
-    }
-    mail_subject, mail_message = reasons[kw.get('reason', 'default')]
+    mail_subject, mail_message = MAIL_COSE_SUBJECT, MAIL_COSE_MESSAGE
+
+    # query everything except for search-plugins as they're generally
+    # not signed
+    current_versions = (
+        Addon.objects
+        .filter(id__in=addon_ids)
+        .exclude(type=amo.ADDON_SEARCH)
```
diox: Shouldn't this be done earlier, in `process_addons` task declaration ?

src/olympia/lib/crypto/tasks.py
```diff
@@ -91,49 +60,58 @@ def get_new_version_number(version):
...
+        .values_list('_current_version', flat=True))
+    qset = (
+        Version.objects.filter(id__in=current_versions)
+        .exclude(addon__type=amo.ADDON_SEARCH))
```
diox: Shouldn't be needed.

src/olympia/lib/crypto/tasks.py
EnTeQuAk: Actually, I just re-read the `process-addon` command and I think you're right, all that should be moving there. I'll fix that.

src/olympia/lib/crypto/tasks.py
diox: Wow I'm blind. The problem is present on the line below though, it says `all extension versions`.

EnTeQuAk: dictionaries are affected as well imho

EnTeQuAk: Question is, does "Extension" in a wider sense include themes and the others too? Not sure about our general wording in this regard.

EnTeQuAk:
except for the wording question this should be ready for another review.

src/olympia/lib/crypto/tasks.py
diox: Historically it didn't, at least...

src/olympia/addons/management/commands/process_addons.py
```diff
@@ -41,7 +41,10 @@
         ]},
     'resign_addons_for_cose': {
         'method': sign_addons,
-        'qs': []},
+        'qs': [
+            ~Q(type=amo.ADDON_SEARCH)
```
diox: Now that I think about it we probably want to restrict to public add-ons only. No point in re-signing others. The task will filter some of the non-public out because it looks at `current_version`, but that's not enough.

src/olympia/lib/crypto/tasks.py
caitmuenster: Good points & thanks for the catch! Made a few tweaks here: https://docs.google.com/document/d/167MXie1DEXDVEM0j2V8ALC8MBSjjKqNp6Ebd9TV9gMg/edit#

EnTeQuAk:
* Updated messaging and added test
* We're now only re-signing add-ons created before April 5th since we introduced the new signing on April 4th
* Only public add-ons are being considered

re-r? @diox

src/olympia/lib/crypto/tests/test_signing.py
```diff
@@ -451,213 +451,188 @@ def test_sign_full(self, mock_sign_file):
...
+            assert new_current_version.version == '0.0.2.1-signed'
+            assert new_current_version.version_int == version_int(
+                '0.0.2.1-signed')
+            assert file_hash != new_file.generate_hash()
+
+            # Verify that the old version hasn't been resigned
+            self.version.reload()
+            assert self.version.version == '0.0.1'
+            assert self.version.version_int == version_int('0.0.1')
+            assert file_hash != new_file.generate_hash()
```
diox: This doesn't make sense, it has nothing to do with `self.version`. Wrong copy/paste ?

EnTeQuAk: Fixed in 63f046794a, indeed, that didn't make any sense :roll_eyes: yeah, some copy&paste error :-/

Issue #11586, opened 2019-06-05
Accessing the taar service may log sensitive information


bqbn:
### Describe the problem and steps to reproduce it:
https://bugzilla.mozilla.org/show_bug.cgi?id=1546132 requested that we set up HTTP basic auth when talking to taar service.

If that's enabled, the application may log sensitive information such as

`"Calling recommendation server: https://<user>:<pass>@taar.dev.mozaws.net/v1/api/recommendations/12345678-90ab-cdef-1234-567890abcdef/"` 

We should make sure no secret is logged before enabling HTTP basic auth when talking to taar service.

### What happened?
Application logs sensitive information if HTTP basic auth is used for taar service.

### What did you expect to happen?
Do not log sensitive info when talking to taar service.

### Anything else we should know?
n/a

diox:
I think we can just remove that log statement entirely.

PR #11598, opened 2019-06-06
Stop logging recommendations server endpoint

diox:
Fixes https://github.com/mozilla/addons-server/issues/11586


muffinresearch:
Is there a chance this could be logged in future? I wonder if the logging calls should be setup so anything that looks like basic auth is automatically stripped?

diox:
Does that log provides any useful information though ? Logging the full URL of the taar endpoint being called doesn't seem particularly useful, and it's a bit spammy since it gets calls so many times.

PR #11627, opened 2019-06-12
Sign add-ons with recommendation signer based on DiscoveryItem.

### LINKED ISSUE #11062: Update autograph calls to send the right parameters for recommended extensions
Autograph is going to be updated to provide support for adding to the signature of an extension so that Firefox will know that a curated extension is going to be used.

Once that support is ready and we have already done the backend work to support the identification of recommended add-ons we can then conditionally send different params to autograph to request the correct signature attribution.

https://bugzilla.mozilla.org/show_bug.cgi?id=1548647

EnTeQuAk:
This also prepares for the possibility that we may have to use different
credentials for the recommendations signer. In case we don't, ops only
has to set the creds twice which seems fair.

Fixes #11062

### DIFF
--- a/scripts/autograph_travis_test_config.yaml
+++ b/scripts/autograph_travis_test_config.yaml
@@ -245,4 +245,7 @@ authorizations:
       key: fs5wgcer9qj819kfptdlp8gm227ewxnzvsuj9ztycsx08hfhzu
       signers:
           - webextensions-rsa
+    - id: bob
+      key: 9vh6bhlc10y63ow2k4zke7k0c3l9hpr8mo96p92jmbfqngs9e7d
+      signers:
           - webextensions-rsa-with-recommendation
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,17 +69,17 @@ before_script:
 
 script:
   - |
-    if [[ $TRAVIS_EVENT_TYPE != "cron" ]]; then
-      if [[ $TOXENV == "amo-lib-locales-and-signing" ]]; then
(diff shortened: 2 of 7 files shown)


scripts/autograph_travis_test_config.yaml
```diff
@@ -245,4 +245,7 @@ authorizations:
       key: fs5wgcer9qj819kfptdlp8gm227ewxnzvsuj9ztycsx08hfhzu
       signers:
           - webextensions-rsa
+    - id: bob
+      key: 9vh6bhlc10y63ow2k4zke7k0c3l9hpr8mo96p92jmbfqngs9e7d
+      signers:
```
EnTeQuAk: These credentials have to match https://github.com/mozilla-services/autograph/blob/master/autograph.yaml#L1053 (and respectively for `alice`)

EnTeQuAk:
> This makes sense. Do we have any end-to-end test in reviewer tools that include signing ? Maybe it's worth adding one w/ a recommendable add-on...

We don't unfortunately, https://github.com/mozilla/addons-server/blob/0a1419cd012668d747ff5441374481018c079b26/src/olympia/reviewers/tests/test_utils.py#L1289-L1315 tests this mostly but doesn't test the whole flow. Guess it's worth adding a test for that that doesn't rely on mocking.

EnTeQuAk:
> Maybe it's worth adding one w/ a recommendable add-on...

Added d99ab37 regarding end-to-end tests.

PR #11645, opened 2019-06-14
rm all the persona stuff from devhub

### LINKED ISSUE #10785: remove LWT support from devhub
might have to split up into stages/separate issues.  reviewer tools was already removed in #10755

eviljeff:
fixes #10785


eviljeff:
note: really, we can't merge this before all the LWT are removed, or they will break in devhub

PR #11682, opened 2019-06-21
Compare API: Render unmodified files.

### LINKED ISSUE #11639: Please return a full file from the Compare API when there are no changes
Currently the Compare API returns an empty file for the file contents when there are no changes. Instead of that could it please return the full contents of the file, with all lines showing as unchanged?

This will fix https://github.com/mozilla/addons-code-manager/issues/831 and should also help with the work on https://github.com/mozilla/addons-code-manager/issues/672.

EnTeQuAk:
Currently we followed how ``git diff`` works and didn't render any
changes for unmodified files.

To help with some frontend issues (e.g showing linter messages in compare view for unmodified files, and others) we now need to include unmodified files.
That alone would not render their contents, so we're including a faked
hunk that shows all lines as unmodified.

Fixes #11639

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,7 +20,6 @@
 Addons-Server
 =============
 
-foo
 Welcome to the Addons Server repository!  Please feel free to visit the web page of the current project hosted on `addons.mozilla.org`_. If you want to install it follow our guide located in `install docs`_.  We'd love your help!  You can come talk to us on `irc.mozilla.org #amo`_ if you have any questions.
 
 Please report bugs here: https://github.com/mozilla/addons/issues or https://github.com/mozilla/addons-server/issues
--- a/README.rst
+++ b/README.rst
@@ -20,6 +20,7 @@
 Addons-Server
 =============
 
+foo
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -20,6 +20,7 @@
 Addons-Server
 =============
 
+foo
```
kumar303: Like Kung Fu?

EnTeQuAk: Doh, I was test-generating git diffs and failed to not commit that change :rofl:

src/olympia/lib/git.py
```diff
@@ -525,6 +532,46 @@ def _render_patch(self, patch, commit, parent):
...
+                    'content': line,
+                    'type': GIT_DIFF_LINE_CONTEXT,
+                    'old_line_number': lineno,
+                    'new_line_number': lineno,
+                }
+                for lineno, line in enumerate(actual_blob.data.split(b'\n'))
+            ]
+
+            hunks.append({
+                'header': '@@ -0 +0 @@',
```
EnTeQuAk: @bobsilverberg 	I'm going to merge this PR to test it myself on -dev even maybe but we may need to tweak the faked hunk a bit, depending on the parsing of the used library. I looked into the source-code, and the fake should work relatively fine - just so that you're aware in case you run into any kinds of issues while testing this.

PR #11690, opened 2019-06-21
Update 'file' version to fix json<->html detection for reviewe…

### LINKED ISSUE #11534: Browse API returns wrong content type for youtube-download-video

* Sign into prod with a reviewer account
* Load https://code.addons.mozilla.org/en-US/browse/1482277/versions/3189741/?path=data%2Fconfig.json

It makes this API call
```
https://addons.mozilla.org/api/v5/reviewers/addon/1482277/versions/3189741/?file=data%2Fconfig.json
```

### What happened?

The API response specifies the mime type of the file `data/config.json` incorrectly as `text/html`.

It should be `application/json`

This renders the file without syntax highlighting. See https://github.com/mozilla/addons-code-manager/issues/790

Other json files do not typically have this problem.

EnTeQuAk:
Fixes #11534

~~Given the updates, and the fact that the Ubuntu Bionic update seems to be on beta on travis (not yet announced) we might need to double-check if we want all that.~~

This is upgrading libmagic-dev on our docker image and on travis to make use of at least of version 5.34 which seems to work perfectly fine. Some more analysis can be found in #11534 

I don't see many other way around the issue of upgrading the `file` tool though.


EnTeQuAk:
Okay, this was a false attempt, Ubuntu Bionic uses `file` 5.32 but we want 5.35 at least :-/

I'll continue to investigate.

.travis.yml
```diff
@@ -52,6 +52,12 @@ before_install:
   - mysql -e 'create database olympia;'
   - export GOPATH=$HOME/go
   - export PATH=$HOME/usr/local/go/bin:$GOPATH/bin:$PATH
+  # Unfortunately we require the most recent libmagic-dev version to make use of more recent
+  # JSON detection features which we require for the code-manager related APIs
+  - wget http://security.ubuntu.com/ubuntu/pool/main/f/file/libmagic-dev_5.35-4_amd64.deb
+  - wget http://security.ubuntu.com/ubuntu/pool/main/f/file/libmagic1_5.35-4_amd64.deb
+  - wget http://security.ubuntu.com/ubuntu/pool/main/f/file/libmagic-mgc_5.35-4_amd64.deb
+  - sudo dpkg -i libmagic-dev_5.35-4_amd64.deb libmagic1_5.35-4_amd64.deb libmagic-mgc_5.35-4_amd64.deb
```
diox: Can you figure out a way to add a repo instead like you do in the `Dockerfile` ? Really not a fan of all those hardcoded versions here.

EnTeQuAk: Probably, I'll try to make use of https://packages.ubuntu.com/disco/libmagic-dev which is at least a released (non testing) version.

PR #11706, opened 2019-06-25
Always translate custom_heading in DiscoveryItem._build_heading()

### LINKED ISSUE #11704: heading_text is not localized in discovery API response
The `heading_text` property of the discovery API response includes the English text.

### What happened?
Recommended extensions in about:addons are shown with an English heading and a localized description.

Both the heading and description are localized.

https://services.addons.mozilla.org/api/v4/discovery/?lang=hsb
https://bugzilla.mozilla.org/show_bug.cgi?id=1559160

diox:
Fixes #11704


diox:
e02c91838ed3688961704d6aeb505c8f7851aee5 adds a test (impossible to do without mocking since those translations can potentially change often)

PR #11713, opened 2019-06-26
Don't render fake hunk files for binary files such as images.

### LINKED ISSUE #11712: 500 errors received in Compare file for images
Followup for https://github.com/mozilla/addons-server/issues/11639
STR:
1. Load a compare view such as 
https://code.addons-dev.allizom.org/en-US/compare/390151/versions/1688198...1688302/
https://code.addons.allizom.org/en-US/compare/1002795/versions/2474830...2474841
2. Open a folder that contains images and click a file such as https://code.addons.allizom.org/en-US/compare/1002795/versions/2474830...2474841?path=icon%2F176.png

Expected result:
No differences displayed 

Actual result:
Internal server error received for https://addons.allizom.org/api/v5/reviewers/addon/1002795/versions/2474830/compare_to/2474841/?file=icon/176.png&lang=en-US

Notes: 
this issue is reproducible on -dev and -stage - FF67 -Win10
I could not reproduce it with text files or in the versions page (browse view). 

500 error compare view

EnTeQuAk:
* Moves mimetype detection to olympia.lib.git
* Only generates fake hunks for files that are 'text' files (not binary
or images)

Fixes #11712

This might need a cherry-pick as it fixes a regresson in a new feature for the code-manager.


src/olympia/lib/git.py
```diff
@@ -59,6 +62,74 @@
...
+    # See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#textjavascript  # noqa
+    'application/javascript': 'text/javascript',
+}
+
+
+# Some official mimetypes belong to the `text` category, even though their
+# names don't include `text/`.
+MIMETYPE_CATEGORY_MAPPING = {
+    'application/json': 'text',
+    'application/xml': 'text',
```
diox: I know you only moved that but I suspect you'll need `application/xhtml+xml`, `application/rss+xml`, `application/atom+xml` etc too

EnTeQuAk: I added https://github.com/mozilla/addons-server/issues/11731 to not blow up the patch even more.

EnTeQuAk: Actually… after investigating this in #11731 I figured that `libmagic` doesn't know about these specific types.

XHTML is simply returned as `text/html` and RSS/Atom is returned as `text/xml` which is perfectly fine too.

PR #11717, opened 2019-06-26
Don't return subject line in bounce email

### LINKED ISSUE #11707: Do not bounce emails for spammers
Our email provider has reported that spammers are able to spoof sender addresses and have AMO bounce emails for them.

Please see details in https://bugzilla.mozilla.org/show_bug.cgi?id=1561413

This hurts our email reputation and can get us banned by ISPs.

### What happened?

See https://bugzilla.mozilla.org/show_bug.cgi?id=1561413 for details.

We should not bounce emails from spammers. On the other hand, we should probably still allow valid email bounces to happen.

n/a

eviljeff:
fixes #11707


eviljeff:
Changed mind about the usefulness of this change.

PR #11727, opened 2019-06-27
Remove all_personas column from collections table

### LINKED ISSUE #11718: drop 'all_personas' column from 'collections' table
#11705 removes the column from the model; we should drop the column from the database with a migration in _the following milestone/deploy_

Abhi-khandelwal:
Fixes #11718

### DIFF
--- a/src/olympia/migrations/1099-remove-all-personas-column.sql
+++ b/src/olympia/migrations/1099-remove-all-personas-column.sql
@@ -1,0 +1,2 @@
+ALTER TABLE `collections`
+ DROP COLUMN `all_personas`;


eviljeff:
Note: do _not_ merge this until after the 2017.07.04 milestone is tagged.

willdurand:
For some reasons, I have troubles running `make update_docker`:

```
[...]

Running SQL migration 1103:
 ALTER TABLE `collections`
 DROP COLUMN IF EXISTS `all_personas`;

Error: Had trouble running this: BEGIN;
ALTER TABLE `collections`
 DROP COLUMN IF EXISTS `all_personas`;

UPDATE schema_version SET version = 1103;
COMMIT;
stdout: b''
stderr: b"ERROR 1064 (42000) at line 2: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'IF EXISTS `all_personas`' at line 2\n"

[...]
```

I'd suggest we use `IF EXISTS` clauses, but on the other hand, I am not sure to understand why this migration is failing.

diox:
`DROP COLUMN IF EXISTS` isn't supported by MySQL AFAIK (sadly). My guess is you don't need that migration because your database was created from a addons-server clone that didn't have that field in the model already. Easiest workaround is to run `./schematic src/olympia/migrations -u 1103` from the `web` container to have it ignore that migration

willdurand:
> `DROP COLUMN IF EXISTS` isn't supported by MySQL AFAIK (sadly).

Correct.

> My guess is you don't need that migration because your database was created from a addons-server clone that didn't have that field in the model already. Easiest workaround is to run `./schematic src/olympia/migrations -u 1103` from the `web` container to have it ignore that migration

Yeah, I ended up doing this. I am wondering if this migration will be a problem with a fresh install, though.

diox:
On completely fresh installs, we ignore the migrations up to the current number. If the migrations aren't not completely in sync (because we remove the model field one week before removing the sql column) you can get issues like this one if you're unlucky. https://github.com/mozilla/addons/issues/941 also throws a wrench into this...

PR #11730, opened 2019-06-27
Remove all usage of 'six' and as much compat code as I could find.

### LINKED ISSUE #11728: Remove six py2/py3 compat code
Python 3 support is now live for quite some time.

Let's remove our compatibility code.

EnTeQuAk:
Cleans up some imports along the way.

Fixes #11728


conftest.py
```diff
@@ -138,10 +137,8 @@ def test_pre_setup(request, tmpdir, settings):
     from waffle.utils import get_cache as waffle_get_cache
     from waffle import models as waffle_models
 
-    # Ignore ResourceWarning for now. It's a Python 3 thing so it's done
-    # dynamically here.
-    if six.PY3:
-        warnings.filterwarnings('ignore', category=ResourceWarning)  # noqa
+    # Ignore ResourceWarning for now. It's a Python 3 thing :-/
+    warnings.filterwarnings('ignore', category=ResourceWarning)  # noqa
```
eviljeff: we can have this in `setup.cfg` now - it was only in here because we needed to be able to only enable it for python 3.

EnTeQuAk: Indeed! Fixed in bd499fec29

PR #11738, opened 2019-06-28
Optimize number of queries for content review & auto approved reviewer queues

### LINKED ISSUE #11733: Reduce number of queries made by reviewer tools queues
Some reviewer tools queues, at least auto-approved and content-review, make an extra query per item in the queue. That makes the number of sql queries blow up and in fact, locally, I can't load those queues when they are full (we have pagination at 200 items, it blows up before that).

The reason is quite subtle: 
We correctly do `select_related('_current_version__autoapprovalsummary')`, so a `JOIN` is performed for both `_current_version` and `autoapprovalsummary`, but then the default `transformer()` method on `Addon` overrides the `_current_version` on each add-on with ones from a (single) query it makes. That, in turn, erases all `autoapprovalsummary` fetched by the initial query.

Fixing this is not trivial. In reviewer tools we don't need the transformer most of the time... except for files, which are preloaded from the `Version` default `transformer()` method. So adding a `only_translations()` to the queryset, preventing the `Addon` default `transformer()` to run, still causes an extra query to be made per result, this time to fetch the `files`... And `files` is currently an invalid target for `select_related()` on a `Version`, because it's a many-to-one relation. Need to investigate if using `prefetch_related()` instead would work.

diox:
Fixes #11733


src/olympia/reviewers/tests/test_views.py
```diff
@@ -1981,7 +1981,17 @@ def test_only_viewable_with_specific_permission(self):
     def test_results(self):
         self.login_with_permission()
         self.generate_files()
-        self._test_results()
+        with self.assertNumQueries(0):
```
EnTeQuAk: `0`?

diox: That was for testing, it should be 23 as well...

PR #11740, opened 2019-06-28
Only bounce invalid emails when they aren't spammy

### LINKED ISSUE #11707: Do not bounce emails for spammers

eviljeff:
fixes #11707 hopefully

### DIFF
--- a/src/olympia/activity/utils.py
+++ b/src/olympia/activity/utils.py
@@ -34,6 +34,7 @@ REPLY_TO_PREFIX = 'reviewreply+'
 # Group for users that want a copy of all Activity Emails.
 ACTIVITY_MAIL_GROUP = 'Activity Mail CC'
 NOTIFICATIONS_FROM_EMAIL = 'notifications@%s' % settings.INBOUND_EMAIL_DOMAIN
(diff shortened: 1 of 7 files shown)


src/olympia/activity/utils.py
```diff
@@ -128,9 +129,17 @@ def add_email_to_activity_log_wrapper(message):
...
-        except Exception:
-            log.error('Bouncing invalid email failed.')
+            spam_rating = float(spam_rating)
+        except ValueError:
+            spam_rating = 0.0
+        if spam_rating < SOCKETLABS_SPAM_THRESHOLD:
+            try:
+                bounce_mail(message, reason)
+            except Exception:
+                log.error('Bouncing invalid email failed.')
```
diox: `log.exception` would give us the stack trace when needed

PR #11756, opened 2019-07-02
Normalize emails when checking restrictions

### LINKED ISSUE #11755: Normalize emails when checking restrictions
When we check an email in `EmailReputationRestriction` and `EmailUserRestriction`, we should normalize them by removing any `.` and everything after `+`. That will make restrictions more effective against variations of the same email address, since some email providers like gmail let you have multiple aliases that way.

For the reputation part, this will require the fraud pipeline to also apply the normalization before we can deploy this.

diox:
Fixes #11755

### DIFF
--- a/src/olympia/signing/tests/test_views.py
+++ b/src/olympia/signing/tests/test_views.py
@@ -141,7 +141,7 @@ class TestUploadVersion(BaseUploadVersionTestMixin, TestCase):
 
     def test_user_does_not_own_addon(self):
         self.user = UserProfile.objects.create(
-            read_dev_agreement=datetime.now())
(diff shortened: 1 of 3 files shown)


src/olympia/signing/tests/test_views.py
```diff
@@ -141,15 +141,15 @@ def test_new_addon_random_slug_unlisted_channel(self):
 
     def test_user_does_not_own_addon(self):
         self.user = UserProfile.objects.create(
```
eviljeff: Certainly not a requirement for this patch but we should replace these `UserProfile.objects.create()` with `user_factory`

diox: I almost did but didn't want to dig into any potential repercussions just now

PR #11757, opened 2019-07-02
Apply upload restrictions to known disposable email providers

### LINKED ISSUE #11633: Port pre-fxa email domain blocks to new admin

We had a list of blocked email domains, which were all well known throwaway email providers. This was removed on the move to FxA (see https://github.com/mozilla/addons-server/pull/3589), and we should add them back to the new list.

diox:
Use a new model to make it easier on the db (don't have to load everything since there are no wildcards to support, we can just do a single efficient db query)

Fixes #11633


diox:
Depends on fixes in https://github.com/mozilla/addons-server/pull/11756 for the failing tests

src/olympia/devhub/tests/test_permissions.py
```diff
@@ -29,6 +31,13 @@ def test_has_permission_user_has_not_read_agreement(self):
             'The Firefox Add-on Distribution Agreement also links to our '
             'Privacy Notice which explains how we handle your information.')
 
+    def test_has_permission_disposable_email(self):
+        DisposableEmailDomainRestriction.objects.create(domain='example.com')
+        assert not self.permission.has_permission(self.request, self.view)
+        assert self.permission.message == (
+            'The email address you used for your developer account is not '
+            'allowed for add-on submission.')
```
eviljeff: nit: `'The email address used for your account is not allowed for add-on submission.'`?

diox: Since we're already using that string and I think it was given by product I'm not changing it here. I do like your version better though, I'll file an issue about getting it changed (p5 :p)

src/olympia/migrations/1100-add-disposable-emails-domain-restriction.sql
```diff
@@ -0,0 +1,4544 @@
+CREATE TABLE `users_disposable_email_domain_restriction` (
+    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `created` datetime(6) NOT NULL,
+    `modified` datetime(6) NOT NULL,
+    `domain` varchar(255) NOT NULL UNIQUE
+);
+
+INSERT INTO `users_disposable_email_domain_restriction` (`domain`, `created`, `modified`) VALUES
```
bqbn: Would it be a good idea to just create the table during the migration, and load the data via an after-push script? From a push perspective, the less to do during the migration the better.

diox: Locally inserting those rows was almost instantaneous so I figured it wasn't much of a problem. Was it long on stage?

bqbn: Nah, it's not a big deal and it's below a second in this case. I was more wondering from a best/standard practice perspective.

diox: It's quite a rare occurrence so we don't have established best practices for it. I'd have probably done this via a script if it was addons/versions/files/users table, like we did when we mass-disabled legacy add-ons for instance.

PR #11774, opened 2019-07-03
Pass the count when paginating reviewer tools queues

### LINKED ISSUE #11769: Reviewer tools queues pagination evaluates the full queries instead of `LIMIT`ing
Reviewer tools queues don't paginate properly. For the reviewer everything looks ok, they get the right page, but underneath we run the full, unlimited query. For queues with thousands of rows this is incredibly inefficient. It's causing https://github.com/mozilla/addons-server/issues/11767 at the moment because on dev/stage there are 300k items in the content review queue.

Reviewer tools queues do this:
```python
    table = TableObj(data=qs, order_by=order_by)
    per_page = request.GET.get('per_page', REVIEWS_PER_PAGE)
    try:
        per_page = int(per_page)
    except ValueError:
        per_page = REVIEWS_PER_PAGE
    if per_page <= 0 or per_page > REVIEWS_PER_PAGE_MAX:
        per_page = REVIEWS_PER_PAGE
    page = paginate(request, table.rows, per_page=per_page, count=qs.count())
    table.set_page(page)
```

When `paginate()` is called, that instantiates a `Paginator`, which does the page calculations. When doing this, it calls `self.count`, which is implemented like this:
```python
    @cached_property
    def count(self):
        """Return the total number of objects, across all pages."""
        c = getattr(self.object_list, 'count', None)
        if callable(c) and not inspect.isbuiltin(c) and method_has_no_args(c):
            return c()
        return len(self.object_list)
```

So, if we're paginating a queryset, we call `count()`, and then we paginate as normal. However, we're calling `paginate()` with `table.rows`, which does *not* have a `count` callable attribute, so we end up falling back to `len(self.object_list`), which evaluates the full query and then counts the number of items in it in python, which is incredibly bad for performance.

diox:
If we don't do that, because we pass table.rows to the paginator and it doesn't have a count callable attribute, django's `Paginator` falls back to doing `len()`, which evaluates the full query instead!

Also add django-debug-toolbar on our local envs when `DEBUG` is `True` to make this sort of thing easier to spot.

Fixes #11769


EnTeQuAk:
On second thought… before merging this, can we make sure that the debug-toolbar commit doesn't get cherry-picked?

diox:
It should automatically make it work as long as `DEBUG` is `True`, on local envs. On dev/stage/prod, or if `DEBUG` is `False`, it should not nothing (the app isn't even in `INSTALLED_APPS` or `MIDDLEWARE` on dev/stage/prod)

PR #11776, opened 2019-07-03
Add command to mark migrated public themes has having passed content review

### LINKED ISSUE #11767: The Content review queue is inaccessible on -dev and stage due to an increased number of pending items
**STR**

1. Log in to AMO dev or stage with a reviewer account
2. Try to access the Content review queue - https://reviewers.addons.allizom.org/en-US/reviewers/queue/content_review

**Actual result:**
The content review queue has too many entries, exploring its contents is very difficult and most of the time I'm receiving a 504 when I'm trying to open it.

**Expected result:**
The content review queue is functional

**Notes:**
- it was pretty hard to check where all these new pending items where coming from, due to the 504s, but at a first glance it looks like all the migrated themes are now part of this queue - on stage, for example, there are 386052 pending items
- the regression could come from either https://github.com/mozilla/addons-server/pull/11750 or https://github.com/mozilla/addons-server/pull/11738

diox:
Fixes #11767

### DIFF
--- a/src/olympia/addons/management/commands/process_addons.py
+++ b/src/olympia/addons/management/commands/process_addons.py
@@ -9,6 +9,7 @@ from olympia import amo
 from olympia.addons.models import Addon
 from olympia.addons.tasks import (
     add_dynamic_theme_tag,
+    content_approve_migrated_themes,
     delete_addons,
     extract_colors_from_static_themes,
     find_inconsistencies_between_es_and_db,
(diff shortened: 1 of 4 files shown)


diox:
This is a one-off, but we'll want that in production as well.

EnTeQuAk:
> This is a one-off, but we'll want that in production as well.

Alrighty then

Issue #11721, opened 2019-06-26
Ui-test failure: Unable to locate element: .Home-FeaturedExtensions


muffinresearch:
Unable to locate element: .Home-FeaturedExtensions

From: https://circleci.com/gh/mozilla/addons-server/34155

jrbenny35:
This should be fixed with https://github.com/mozilla/addons-server/issues/11565

Issue #11720, opened 2019-06-26
Ui-test failure: Account already exists


muffinresearch:
https://circleci.com/gh/mozilla/addons-server/34162

```sh
    output = self.handle(*args, **options)
  File "/code/src/olympia/landfill/management/commands/generate_default_addons_for_frontend.py", line 77, in handle
    serializer = GenerateAddonsSerializer()
  File "/code/src/olympia/landfill/serializers.py", line 50, in __init__
    self.fxa_id = self._create_fxa_user()
  File "/code/src/olympia/landfill/serializers.py", line 61, in _create_fxa_user
    fxa_account.password)
  File "/usr/local/lib/python3.6/site-packages/fxa/core.py", line 64, in create_account
    resp = self.apiclient.post(url, body)
  File "/usr/local/lib/python3.6/site-packages/fxa/_utils.py", line 335, in post
    return self.request("POST", url, json, **kwds)
  File "/usr/local/lib/python3.6/site-packages/fxa/_utils.py", line 321, in request
    raise fxa.errors.ClientError(body)
fxa.errors.ClientError: Account already exists
Makefile-docker:161: recipe for target 'setup-ui-tests' failed
make: *** [setup-ui-tests] Error 1
Exited with code 2
```

Issue #11691, opened 2019-06-23
Move frontend-only related ui-tests to addons-frontend test suite


EnTeQuAk:
We discussed this in Whistler. Now that addons-frontend only serves the API, DevHub, reviewer tools and smaller other pages we should move all pure frontend related ui-tests to the addons-frontend test suite.

This should relate to `tests/ui/{test_home,test_search}` tests.

While we're at it, the `test_login` tests have been skipped for a very long time. Either revisit them and see if they can be moved and fixed or remove them altogether and create an issue to work on them separately.

`test_install` and `test_devhub` should stay in addons-server for now.

It can be done in a separate step but see if any of the test-setup data isn't needed for the remaining tests, if so, remove it.

When touching this, make sure to keep #11565 in mind and remove that test along the way (ideally this should be all separate steps / prs).

File separate issues as needed.

suhailsinghbains:
Hi, Can I please work on this issue?
Thanks

EnTeQuAk:
This is an assigned issue and not necessarily that is open for contributions. I'm sorry about that, but there's a whole list of other issues where contributions are very welcome! https://github.com/mozilla/addons-server/labels/contrib%3A%20welcome

suhailsinghbains:
Okay, thanks :)

jrbenny35:
This depends on: https://github.com/mozilla/addons-server/issues/11565

eviljeff:
Are there blockers on removing the ui-tests from addons-server now _before_ some subset of them have been re-added to addons-frontend? We'd lose test coverage on those tests in the mean time but it'd only be for a few weeks, right?

The code itself will remain in the git[hub] history for reference.

jrbenny35:
@eviljeff the job on circleci can be disabled and I plan to move the tests soon. I don't think there is a problem leaving the code here for now until its moved as its pretty out of the way. If you want to disable the current job, I am okay with that

jrbenny35:
I'll have to re-update ```pytest-xdist``` when the tests are moved.

jrbenny35:
Done here https://github.com/mozilla/addons-frontend/pull/9018

PR #11811, opened 2019-07-09
Initial attempt at fixing integration-tests.

### LINKED ISSUE #11565: CircleCI builds are failing
Moving this to addons-frontend, now that legacy frontend is gone is another thing but we should put circleci tests to green in the meantime.

```
tests/ui/test_home.py::test_category_section_loads_correct_category[Resolution: 414x738] FAILED
```

And it seems to fail because `.CategoryHeader-name` isn't present on this page: 

index

Which links to this commit on addons-frontend: https://github.com/mozilla/addons-frontend/commit/f7d85d5d88233d60cb19726c57463f13da14f0c0

jrbenny35:
This hopefully fixes the integration tests before the BIG SPLIT. 
Fixes #11565 #11720 #11721
Works on #11691

### DIFF
--- a/src/olympia/landfill/serializers.py
+++ b/src/olympia/landfill/serializers.py
@@ -17,10 +17,11 @@ from rest_framework import serializers
 
 import olympia.core.logger
 
-from olympia.amo.tests import user_factory, addon_factory, copy_file_to_temp
+from olympia.amo.tests import user_factory, addon_factory, copy_file_to_temp, version_factory
 from olympia import amo
-from olympia.addons.models import AddonUser, Preview, Addon
+from olympia.addons.models import AddonUser, Preview, Addon, Persona
 from olympia.addons.utils import generate_addon_guid
+from olympia.amo.utils import days_ago
 from olympia.constants.applications import APPS, FIREFOX
 from olympia.constants.base import (
     ADDON_EXTENSION,
@@ -29,12 +30,16 @@ from olympia.constants.base import (
 )
 from olympia.devhub.forms import icons
 from olympia.landfill.collection import generate_collection
-from olympia.landfill.generators import generate_themes
+from olympia.landfill.generators import generate_themes, create_theme
 from olympia.landfill.user import generate_user
 from olympia.files.tests.test_file_viewer import get_file
 from olympia.ratings.models import Rating
 from olympia.users.models import UserProfile
 from olympia.devhub.tasks import create_version_for_upload
+from olympia.discovery.models import DiscoveryItem
+
+from .version import generate_version
+
 
 log = olympia.core.logger.getLogger('z.users')
 
@@ -90,8 +95,30 @@ class GenerateAddonsSerializer(serializers.Serializer):
 
         """
         for _ in range(10):
+            addon = addon_factory(
+                status=amo.STATUS_APPROVED,
+                version_kw={
+                    'recommendation_approved': True,
+                    'nomination': days_ago(6)
+                })
             AddonUser.objects.create(
-                user=user_factory(), addon=addon_factory())
+                user=user_factory(), addon=addon)
+            DiscoveryItem.objects.create(
+                recommendable=True, addon=addon)
+
+    def create_generic_featured_themes(self):
+        for _ in range(10):
+            addon = addon_factory(
+                status=amo.STATUS_APPROVED,
+                type=ADDON_PERSONA,
+                version_kw={
+                    'recommendation_approved': True,
+                    'nomination': days_ago(6)
+                })
+            generate_version(addon=addon)
+            addon.update_version()
+            DiscoveryItem.objects.create(
+                recommendable=True, addon=addon)
 
     def create_named_addon_with_author(self, name, author=None):
         """Create a generic addon and a user.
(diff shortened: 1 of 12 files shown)


EnTeQuAk:
You may have to take a look at the code-style related failures in travis: https://travis-ci.org/mozilla/addons-server/jobs/556537194#L515-L536

The circleci issue doesn't look permanent but you may have a much better idea about what's going on there.

src/olympia/landfill/serializers.py
```diff
@@ -103,25 +130,36 @@ def create_named_addon_with_author(self, name, author=None):
         if author is None:
             author = user_factory()
         try:
-            generate_user(author)
+            user = UserProfile.objects.create(username=author, email=f'{author}@email.com')
+            user.id = 4757633
```
EnTeQuAk: You will have to use `user.update(id=settings.4757633)` ideally to not hardcode the id.

src/olympia/landfill/serializers.py
```diff
@@ -103,25 +130,36 @@ def create_named_addon_with_author(self, name, author=None):
...
                 name=u'{}'.format(name),
                 slug=u'{}'.format(name),
+                version_kw={
+                'recommendation_approved': True,
+                'nomination': days_ago(6)
+                }
             )
             addon.save()
         else:
+            author.id = 4757633
```
EnTeQuAk: Same here, you will have to use author.update(id=settings.TASK_USER_ID)` ideally to not hardcode the id.

src/olympia/landfill/serializers.py
```diff
@@ -103,25 +130,36 @@ def create_named_addon_with_author(self, name, author=None):
...
             # If the user is already made, use that same user,
             # if not use created user
             addon = addon_factory(
-                status=STATUS_APPROVED,
+                status=amo.STATUS_APPROVED,
                 users=[UserProfile.objects.get(username=author)],
                 name=u'{}'.format(name),
                 slug=u'{}'.format(name),
+                version_kw={
+                'recommendation_approved': True,
```
EnTeQuAk: Look at the codestyle issues in travis, this needs indentation.

tests/ui/pages/desktop/base.py
```diff
@@ -11,7 +11,7 @@ class Base(Page):
 
     def __init__(self, selenium, base_url, locale='en-US', **kwargs):
         super(Base, self).__init__(
-            selenium, base_url, locale=locale, timeout=30, **kwargs)
+            selenium, base_url, locale=locale, timeout=10, **kwargs)
```
EnTeQuAk: Is that a performance optimization or why did we change that?

tests/ui/pages/desktop/devhub.py
```diff
@@ -142,12 +142,14 @@ class Header(Region):
         _sign_out_locator = (By.CSS_SELECTOR, '.DevHub-Navigation-SignOut > a')
 
         def click_login(self):
+            self.selenium.find_element_by_id("djHideToolBarButton").click()
```
EnTeQuAk: Is that actually needed, does it avoid the other buttons to be clicked? That'd be funny

tests/ui/pages/desktop/devhub.py
```diff
@@ -133,7 +133,7 @@ class Header(Region):
 
         _register_link_locator = (
             By.CSS_SELECTOR,
-            '.DevHub-Navigation-Register > a:nth-child(1)'
+            'li.show-on-desktop:nth-child(7) > a:nth-child(1)'
```
EnTeQuAk: Why is that necessary? The class `DevHub-Navigation-Register` still exists, these pages haven't changed since forever.

tests/ui/test_search.py
```diff
@@ -127,6 +120,7 @@ def test_long_terms_dont_break_suggestions(base_url, selenium):
 
 
 @pytest.mark.nondestructive
+@pytest.mark.skip(reason='Search Behavior has changed')
```
EnTeQuAk: Let's create an issue if that leads to skipping the test.

tests/ui/pages/desktop/base.py
jrbenny35: This can be removed

tests/ui/pages/desktop/devhub.py
jrbenny35: Yeah it is somehow blocking the other buttons.

tests/ui/pages/desktop/devhub.py
jrbenny35: I think this was a change due to the django toolbar. It should be able to be reverted.

tests/ui/test_search.py
jrbenny35: Trying this again with proper recommendations seems to work fine, so I'll "unskip" it.

jrbenny35:
The circleci failure is an error from what seems to be the devhub "Submit a new add-on" page. Doesn't seem like the checkboxes are clickable. I cant reproduce it locally though..

tests/ui/pages/desktop/devhub.py
EnTeQuAk: Probably worth adding a comment. Also, in case we disable debug toolbar by default locally in the future, does that work then too or does the "click" fail loudly?

jrbenny35: It will fail to find the element, and that would be loud.

EnTeQuAk: Should be fine I guess :man_shrugging:

EnTeQuAk:
> This hopefully fixes the integration tests before the BIG SPLIT.
Fixes #11565 #11691 #11720 #11721

It prepares for  #11691, it doesn't really fix it though

PR #11814, opened 2019-07-11
 Reviewers Browse-API: render empty string for content for binary files. 

### LINKED ISSUE #11782: Reviewers Browse API: Stop rendering content for binary files
As discussed in #11759, we are currently sending content for binary files for the browse views but aren't generating fake-hunks for binary files in the compare view.

We should simply stop returning content for binary files completely and offer the user a download-link instead.

That'll also prevent us from unforeseeable security issues.

EnTeQuAk:
Fixes #11782

### DIFF
--- a/src/olympia/lib/git.py
+++ b/src/olympia/lib/git.py
@@ -114,7 +114,14 @@ def get_mime_type_for_blob(tree_or_blob, name, blob):
         # Allow text mimetypes to be more specific for readable files.
         # `python-magic`/`libmagic` usually just returns plain/text but we
         # should use actual types like text/css or text/javascript.
-        mimetype, _ = mimetypes.guess_type(name)
+        guessed_mimetype, _ = mimetypes.guess_type(name)
+
+        # If the file for some reason doesn't have a known file extension
+        # (could happen for text files like `README`, `LICENSE` etc)
+        # don't null the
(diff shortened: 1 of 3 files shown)


src/olympia/lib/git.py
```diff
@@ -114,7 +114,14 @@ def get_mime_type_for_blob(tree_or_blob, name, blob):
...
+
+        # If the file for some reason doesn't have a known file extension
+        # (could happen for text files like `README`, `LICENSE` etc)
+        # don't null the
+        if guessed_mimetype is not None:
+            mimetype = guessed_mimetype
+
         # Re-apply compatibility mappings since `guess_type()` might return
         # a completely different mimetype.
         mimetype = MIMETYPE_COMPAT_MAPPING.get(mimetype, mimetype)
```
eviljeff: nit: you only need to re-apply the mapping if `guessed_mimetype` wasn't `None`.
i.e. line 123 would be `mimetype = MIMETYPE_COMPAT_MAPPING.get(guessed_mimetype, guessed_mimetype)` instead.

EnTeQuAk: Fixed in e93094e437 if I understood correctly what you meant. That makes it cleaner indeed!

PR #11823, opened 2019-07-12
Match last login ip against ip filter

### LINKED ISSUE #11822: Check last login ip against IP restrictions
We should also match the last login ip against the IP restrictions in case the IP address changed between login and submission.

wagnerand:
Fix #11822.

### DIFF
--- a/src/olympia/users/models.py
+++ b/src/olympia/users/models.py
@@ -669,7 +669,7 @@ class IPNetworkUserRestriction(ModelBase):
         """
         try:
             remote_addr = ipaddress.ip_address(request.META.get('REMOTE_ADDR'))
-            if (request.user):
+            if request.user:
                 user_last_login_ip = ipaddress.ip_address(
                     request.user.last_login_ip)
         except ValueError:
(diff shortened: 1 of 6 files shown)


src/olympia/users/models.py
```diff
@@ -669,16 +669,21 @@ def allow_request(cls, request):
         """
         try:
             remote_addr = ipaddress.ip_address(request.META.get('REMOTE_ADDR'))
+            if (request.user):
```
EnTeQuAk: No need for parenthesis here, just do `if request.user`

src/olympia/users/models.py
```diff
@@ -669,16 +669,21 @@ def allow_request(cls, request):
...
         restrictions = IPNetworkUserRestriction.objects.all()
 
         for restriction in restrictions:
-            if remote_addr in restriction.network:
-                log.info('Restricting request from %s %s (%s)',
+            if (remote_addr in restriction.network or
+               user_last_login_ip in restriction.network):
+                log.info('Restricting request from %s %s, %s %s (%s)',
                          'ip', remote_addr,
+                         'user last login ip', user_last_login_ip,
```
EnTeQuAk: Maybe naming this `last_login_ip` instead of the spaces makes it easier to match later on?

PR #11831, opened 2019-07-12
Disable frontend related tests.

jrbenny35:
Should work on #11691 

This disables the frontend related tests and only runs devhub and install tests.


jrbenny35:
> Can we open an issue to roll back this change once #11691 has been implemented? So that we not accidentally forget to run newly added tests?

I'm not sure I follow. Devhub tests will always start with ```test_devhub```

PR #11844, opened 2019-07-16
Drop monthly_pick table

### LINKED ISSUE #11843: drop `monthly_pick` table
After https://github.com/mozilla/addons-server/pull/11840/ we should be able to drop the actual database table for the `monthly_pick` table.

Let's do this in the next tag or so.

Abhi-khandelwal:
Fixes #11843

### DIFF
--- a/src/olympia/migrations/1105-drop-monthly-pick-table.sql
+++ b/src/olympia/migrations/1105-drop-monthly-pick-table.sql
@@ -1,0 +1,1 @@
+DROP TABLE monthly_pick;


src/olympia/migrations/1105-drop-monthly-pick-table.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE monthly_pick;
```
championshuttler: Add a new line after this.

PR #11847, opened 2019-07-17
Normalise xpi_info['guid'] before equality check

### LINKED ISSUE #11174: Add-on ID comparison is case-sensitive when new version, case-insensitive when new add-on
Critical error when uploading new version: "The add-on ID in your manifest.json or install.rdf (NoticeCase@xxx) does not match the ID of your add-on on AMO (noticecase@xxx)"

Critical error when uploading new add-on: "Duplicate add-on ID found."

### What happened?
Signing of the add-on was prevented in both cases.

Either allow a new version of existing add-on or allow a new add-on.

dbxnr:
Fixes #11174 - normalises xpi_info['guid'] before equality check, performs iexact match on database query to ignore case.

### DIFF
--- a/src/olympia/devhub/tests/test_views_validation.py
+++ b/src/olympia/devhub/tests/test_views_validation.py
@@ -117,14 +117,14 @@ class TestUploadErrors(BaseUploadTest):
         upper_guid = addon.guid.upper()
 
         with self.assertRaisesRegex(forms.ValidationError,
-                                    "does not match the ID of your add-on"
-                                    ) as e:
+                                    "does not match the ID of your add-on"):
             xpi_info = check_xpi_info({'guid': upper_guid,
                                        'version': '1.0'},
                                       addon=addon)
         xpi_info = check_xpi_info({'guid': upper_guid,
                                    'version': '1.0'})
         assert xpi_info['guid'] == upper_guid
+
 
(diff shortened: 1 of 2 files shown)


wagnerand:
According to @rpl, Firefox treats add-on ids as case-sensitive. Could you update your PR to make all checks case-sensitive?

EnTeQuAk:
> According to @rpl, Firefox treats add-on ids as case-sensitive. Could you update your PR to make all checks case-sensitive?

The checks are currently case-sensitive, so this seems to be needless to be changed.

I'm going to close the PR for now, and the related issue as the current behavior seems to represent what Firefox does and expects.

Correct me if I'm mistaken though

EnTeQuAk:
re-opening, see reasoning in my comment in the linked issue.

If you have the time, we'd appreciate you update your PR to make all the checks indeed case sensitive and add a few tests to test that new behavior.

dbxnr:
I've updated the PR to perform case-sensitive SQL queries ([__regex](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#regex)).  The CI failures unfortunately seem unrelated to my changes, unless I'm missing something

EnTeQuAk:
I'm closing the PR, last time we discussed this we ended up deciding that the current behavior is perfectly fine for now.

Issue #11830, opened 2019-07-12
Implement admin for Primary Hero Shelf


jvillalobos:
Part of [this PRD](https://docs.google.com/document/d/12aD_v5AOj1Y6VmT4xOuyNcrKfZxqhP8R_AOHroQVLV8/edit#).

The Primary Hero Shelf in the homepage will feature one extension, chosen from a subset of Recommended extensions:

<img width="623" alt="Screen Shot 2019-07-03 at 4 02 16 PM" src="https://user-images.githubusercontent.com/62568/60627685-16904480-9dac-11e9-8d5b-438676aeeb44.png">

Since this will only work with Recommended extensions, it seems sensible to extend the admin we already have for it.

For the admin of a Recommended extension, add:

* Featured Image, chosen from a pool of predetermined images.
* Background color, used for the gradient.
* Featured on Primary Hero Shelf, checkbox that adds or removes the extension from the set. The checkbox shouldn't be toggled on if the other two properties haven't been set, or if the extension is only in the Recommendable state. There should always be at least one extension with the checkbox turned on.

jvillalobos:
The background color will be chosen from a fixed palette, to be provided by @brassy- here.

jvillalobos:
It should also be possible to feature _unlisted_ Recommended extensions in the homepage (and we most likely will at launch). The Call to Action will point to the homepage URL, so having one set should be another requirement to enable the checkbox in the admin, for unlisted ones.

eviljeff:
> It should also be possible to feature unlisted Recommended extensions in the homepage (and we most likely will at launch).

Okay, we could have an additional setting which would mark the item as referring to an unlisted-only add-on and then frontend would know to display it differently.

Unlisted versions can't be approved for recommended though, currently, so it's not possible to have an add-on without a listed version be recommended  - that's #11284.   (And we'd be referring the user to an install that wouldn't be marked in Firefox as recommended)

eviljeff:
> The background color will be chosen from a fixed palette

would this be defined by two colours if it's for a gradient (a start and an end colour)?

bobsilverberg:
> > The background color will be chosen from a fixed palette
> 
> would this be defined by two colours if it's for a gradient (a start and an end colour)?

What I recall from the last meeting we had is that there will always be a single color that represents the start of the gradient, and then each gradient will be defined as having a different end color. If that is the case then the user would only need to choose one color for the gradient (the end color). 

[...]

PR #11858, opened 2019-07-18
create primary hero shelve model, serializer, view

### LINKED ISSUE #11842: create API to expose primary shelf config to frontend
all the admin model data needs to be exposed in a single API endpoint for frontend to consume

eviljeff:
fixes #11842 - adds the base and required code to implement the API response for primary hero shelf.
The `/hero/` endpoint will include the secondary shelf too when implemented.


src/olympia/hero/models.py
```diff
@@ -0,0 +1,16 @@
+from django.conf import settings
+from django.db import models
+
+from olympia.amo.models import ModelBase
+from olympia.discovery.models import DiscoveryItem
+
+
+class PrimaryHero(ModelBase):
+    image = models.CharField(max_length=255)
+    background_color = models.CharField(max_length=7)
```
bobsilverberg: There was some discussion today about how we should record this color choice. It seemed like using css color variables, which will be defined in an .scss file, would be a good idea, but no decision was made. I take it that this format assumes a hex color? I see no reason why that wouldn't work for the front end to consume but before we commit to a restrictive format we should perhaps decide what's going to work best.

src/olympia/hero/models.py
```diff
@@ -0,0 +1,16 @@
...
+    enabled = models.BooleanField(db_index=True, null=False, default=False,)
+    disco_addon = models.OneToOneField(DiscoveryItem, on_delete=models.CASCADE)
+
```
bobsilverberg: I don't see it listed in the issue, nor in the requirements doc, but I recall a discussion about the ability to pin a single add-on to always use in the Primary Hero area, rather than allowing random images to cycle. If we are still planning on implementing that, this seems like the place we'd store that piece of data.

@jvillalobos Do we still plan to implement the ability to pin an individual add-on in the Primary Hero area?

src/olympia/hero/models.py
eviljeff: It could change depending on the answer to the question in the doc, but I was assuming it would be a hex color value, yeah.  Return color names/id (which would have to be kept in sync) seems more complicated than showing swatches in the admin and returning the hex color.

src/olympia/hero/models.py
eviljeff: It got dropped as a requirement.  You could still do that by setting the other PrimaryHero objects to `enabled=False` though.

jvillalobos: Pinning was dropped. Like Andrew said, it's enough to have a single extension in the homepage featured group.

eviljeff:
After an (overly long) discussion on irc we're sticking with `hero` for now but until the feature launches it could be changed.

src/olympia/hero/models.py
eviljeff: I don't think we can merge this until https://github.com/mozilla/addons-server/issues/11830#issuecomment-513230982 is answered because it'll be a pain to change how colours are represented.

eviljeff: added https://github.com/mozilla/addons-server/pull/11858/commits/acaeb79aa6532653fdefda2c308bec7062c07268 for the gradient pair
(+ https://github.com/mozilla/addons-server/pull/11858/commits/53f56d7813cb0b4868c1ef43be7184c4b5584028 for the docs)

src/olympia/hero/models.py
```diff
@@ -0,0 +1,23 @@
...
+from olympia.amo.models import ModelBase
+from olympia.discovery.models import DiscoveryItem
+
+
+GRADIENT_START_COLOR = '#20123A'
+
+
+class PrimaryHero(ModelBase):
+    image = models.CharField(max_length=255)
+    background_color = models.CharField(max_length=7)
```
bobsilverberg: Do you think it makes sense to keep this as `background_color` when it's being used for the gradient end colour? Perhaps a more descriptive name like `gradient_end_color` would be better?

eviljeff: https://github.com/mozilla/addons-server/pull/11858/commits/73904864a04a87e4a736750c627cea266dafa2c4

PR #11859, opened 2019-07-18
add discovery=true filter to discovery editoral endpoint

### LINKED ISSUE #11851: Add param to filter the /api/v4/discovery/editorial/ to only show add-ons with a recommended status.
See also https://github.com/mozilla/taar/issues/157

To support moving TAAR to be base of only add-ons in the Recommended Extensions (RecEx) programme, we need to add a param to the /api/v4/discovery/editorial/ so that it only shows add-ons with a recommended status. That is not just "recommendable" but add-ons that are approved and have latest recommended signed version.

For example:

 `/api/v4/discovery/editorial/?recex=true`

Would show only RecEx add-ons.

Without this param would show everything. The latter will be useful for getting a head start on translations ahead of the recommendations process being completed.

Once TAAR is using the filtered list, non-recommendable content could be cleaned up by editorial.

At the same time as fixing this bug, it would also be good to fix the documentation to reflect that this end point is used for both translations and a source for TAAR.

eviljeff:
fixes #11851

### DIFF
--- a/docs/topics/api/discovery.rst
+++ b/docs/topics/api/discovery.rst
@@ -54,10 +54,12 @@ Editorial Content
 
 This endpoint allows you to fetch all editorial content for Discovery Pane
 Recommendations. This is used internally to generate .po files containing the
-strings the content team came up with.
+strings defined by the content team.  It is also used by TAAR service to obtain a list
+of appropriate add-ons to recommended.
 
  .. http:get:: /api/v4/discovery/editorial/
 
+    :query boolean recommended: Filter to only add-ons recommended by Mozilla.  Only ``recommended=true`` is supported.
     :>json array results: The array containing the results for this query. There is no pagination, all results are returned.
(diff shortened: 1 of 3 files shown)


docs/topics/api/discovery.rst
```diff
@@ -54,10 +54,12 @@ Editorial Content
 
 This endpoint allows you to fetch all editorial content for Discovery Pane
 Recommendations. This is used internally to generate .po files containing the
-strings the content team came up with.
+strings defined by the content team.  It is also used by TAAR service to obtain a list
+of appropriate add-ons to recommended.
 
  .. http:get:: /api/v4/discovery/editorial/
 
+    :query boolean recommended: Filter to only add-ons recommended by Mozilla.  Only ``recommended=true`` is supported.
```
EnTeQuAk: "Filter to only **show** add-ons recommended..." I think :thinking:

Issue #11810, opened 2019-07-09
Bulk-banning users times out


wagnerand:
### Describe the problem and steps to reproduce it:

Bulk-banning a bunch of users makes the admin instance respond with a timeout.

### What happened?

Timeout :(

### What did you expect to happen?

All bans go through without a timeout.

### Anything else we should know?
(Please include a link to the page, screenshots and any relevant files.)

muffinresearch:
@eviljeff if you have a chance to look at this when you're back that would be appreciated.

eviljeff:
need some more detail here - is it any users or just specific ones?  And how many is "a bunch"?

wagnerand:
Any users. Yesterday I tried with 80 and 120 (and for a test run even just with a dozen or so) and every time it timed out.

muffinresearch:
@wagnerand when it times out what are you seeing? Nginx timeout or Django? Just wondering if any external config changes could have affected this?

wagnerand:
I think it's an nginx timeout.

muffinresearch:
I'm not seeing anything obvious in the nginx config repo, but maybe @bqbn could confirm?

bqbn:
> Just wondering if any external config changes could have affected this?

There has no Nginx changes recently. Also we haven't touched Nginx configurations that may affect timing for a long time. Most changes we do for Nginx are route changes these days.

> I think it's an nginx timeout.

Could you provide any logs (from the browser's console for example) to help us investigate where the issue may happen? And which endpoint(s) were you hitting when you saw the timeout? Your source IP and approximate access time would be helpful too. Basically we'll need something that we can use to grep in Nginx logs to see what happens.

If as you said, even banning a dozen or so users leads to timeout, that doesn't sound to be quite efficient. Maybe we should consider doing the bulk banning in an asynchronous job (e.g. done by celery).

eviljeff:
> If as you said, even banning a dozen or so users leads to timeout, that doesn't sound to be quite efficient. Maybe we should consider doing the bulk banning in an asynchronous job (e.g. done by celery).

I want to understand why it went from quick and efficient when it landed a few weeks to inefficient now.

muffinresearch:
Before we try and spend time on fixing perf I think it would be good to make sure we're sure of where the problem lies.

- Is there a good way to see all sentry errors for just the admin instance?
- Is there a clear way to see metrics for just the admin instance we could also try and start adding some additional metrics to see what parts specifically are slow.
[...]

PR #11895, opened 2019-07-23
don't pass a set to `index_addons.delay`

eviljeff:
might fix the errors in the logs listed in #11810


eviljeff:
what would the test be testing?

EnTeQuAk:
> what would the test be testing?

Ideally we'd be able to reproduce the exception we see? At the very least test that a list is supplied to that task, and document it properly.

PR #11928, opened 2019-07-25
Limit abuse report message to 10k characters.

### LINKED ISSUE #11921: Limit body of abuse reports to 10,000 characters
We need to limit the body of abuse reports (both on AMO and coming from Firefox) to avoid issues and prevent misuse. The limit should be 10,000 characters.

EnTeQuAk:
Fixes #11921

### DIFF
--- a/src/olympia/abuse/tests/test_views.py
+++ b/src/olympia/abuse/tests/test_views.py
@@ -166,6 +166,24 @@ class AddonAbuseViewSetTestBase(object):
                           u'[Extension] Abuse Report for %s' % addon.name)
         assert report.message == ''
 
+    def test_message_length_limited(self):
+        addon = addon_factory()
+
+        response = self.client.post(
+            self.url,
+            data={'addon': str(addon.id),
+                  'message': 'a' * 10000})
+        assert response.status_code == 201
+
+        response = self.client.post(
+            self.url,
+            data={'addon': str(addon.id),
+                  'message': 'a' * 10001})
+        assert response.status_code == 400
+        assert json.loads(response.content) == {
+            'message': ['Ensure this field has no more than 10000 characters.']
+        }
+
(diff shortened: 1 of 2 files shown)


EnTeQuAk:
> 10,000 astonished

I think, it's just to limit it to something :see_no_evil:

src/olympia/abuse/tests/test_views.py
```diff
@@ -166,6 +166,24 @@ def test_message_can_be_blank_if_reason_is_provided(self):
...
+                  'message': 'a' * 10000})
+        assert response.status_code == 201
+
+        response = self.client.post(
+            self.url,
+            data={'addon': str(addon.id),
+                  'message': 'a' * 10001})
+        assert response.status_code == 400
+        assert json.loads(response.content) == {
+            'message': ['Ensure this field has no more than 10000 characters.']
```
willdurand: Could we configure a slightly better message? see: https://github.com/mozilla/addons-frontend/issues/8385#issuecomment-514990492

EnTeQuAk: "Please ensure this field has no more than 10000 characters"? I couldn't think of anything nicer.

eviljeff:
`Ensure this field has fewer than 10000 characters.` is less correct (because fewer than is `<10000` rather than `<=10000`) but is a little shorter?

src/olympia/abuse/tests/test_views.py
eviljeff: `Ensure this field has fewer than 10000 characters.` is less correct (because fewer than is `<10000` rather than `<=10000`) but is a little shorter?

willdurand: I think prepending the message with "Please" would make it more user friendly.

PR #11949, opened 2019-07-29
Add admin support for adding/editing primary hero shelf items

### LINKED ISSUE #11830: Implement admin for Primary Hero Shelf

eviljeff:
fixes #11830


eviljeff:
list
Screenshot 2019-07-29 14 03 10
detail
Screenshot 2019-07-29 14 03 42

PR #11954, opened 2019-07-29
Add python script for fxa email gen.

### LINKED ISSUE #11952: Write a python script to generate a UUID for the fxa email within the integration-tests.
It seems that the UUID package will randomly not resolve. This add extra flakiness to the integration tests on circleci that really shouldn't exist. We should build a simple python script to generate an fxa email the tests can use.

jrbenny35:
Fixes #11952

### DIFF
--- a/tests/ui/scripts/generate_fxa_email.py
+++ b/tests/ui/scripts/generate_fxa_email.py
@@ -2,9 +2,7 @@ import uuid
 
 
 def create_fxa_email():
-    id = uuid.uuid1()
-    email = f"uitest-{id}@restmail.net"
-    print(f"{email}")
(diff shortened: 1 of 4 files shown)


tests/ui/scripts/generate_fxa_email.py
```diff
@@ -0,0 +1,11 @@
+import uuid
+
+
+def create_fxa_email():
+    id = uuid.uuid1()
+    email = f"uitest-{id}@restmail.net"
+    print(f"{email}")
```
eviljeff: Heh, you could do this in less lines, but meh.

jrbenny35: Challenge excepted

PR #11973, opened 2019-07-30
implement jinja bytecode cache

### LINKED ISSUE #11969: Implement jinja bytecode cache
We dropped template caching during our upgrade to django-jinja and never ended up re-implementing it.

Let's do that.

Theoretically it's just a matter of adding `django_jinja.builtins.extensions.CacheExtension` to the extensions list, and adding

```python
            "bytecode_cache": {
                "name": "default",
                "backend": "django_jinja.cache.BytecodeCache",
                "enabled": True,
            },
```

to the `OPTIONS` configuration (and `enabled: False` to the local config)

Abhi-khandelwal:
Fixes #11969


EnTeQuAk:
I'm really sorry but #11969 isn't a good issue to be tackled by a contributor. We're going to have to write some tests about that and it needs thorough testing. With that I'm going to close this PR but please feel encouraged to look into other issues with the label `contrib: welcome`

PR #11985, opened 2019-07-31
Add YaraResults model

### LINKED ISSUE #11981: Create database table and model for `YaraResults`
In order to store the results of the `yara` execution, we need a table and a Django model, named `YaraResults`. This model will have 1-1 relations with `FileUpload` (not null) and `versions.Version` (nullable).

willdurand:
Fixes #11981 

---

This patch adds a new table: `yara_results`, used to store the results of `yara`.

### DIFF
--- a/src/olympia/migrations/1107-yara.sql
+++ b/src/olympia/migrations/1107-yara.sql
@@ -1,0 +1,11 @@
+CREATE TABLE `yara_results` (
+    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `created` datetime(6) NOT NULL,
+    `modified` datetime(6) NOT NULL,
+    `upload_id` int(11) NOT NULL,
+    `version_id` integer unsigned,
+    `matches` longtext,
+
+    FOREIGN KEY (`upload_id`) REFERENCES `file_uploads` (`id`),
+    FOREIGN KEY (`version_id`) REFERENCES `versions` (`id`)
+);
--- a/src/olympia/migrations/1107-yara.sql
+++ b/src/olympia/migrations/1107-yara.sql
@@ -1,11 +1,14 @@
 CREATE TABLE `yara_results` (
-    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `id` int(11) AUTO_INCREMENT NOT NULL PRIMARY KEY,
     `created` datetime(6) NOT NULL,
     `modified` datetime(6) NOT NULL,
-    `upload_id` int(11),
-    `version_id` integer unsigned,
-    `matches` longtext,
+    `upload_id` int(11) DEFAULT NULL,
+    `version_id` int(10) unsigned DEFAULT NULL,
+    `matches` longtext NOT NULL,
 
-    FOREIGN KEY (`upload_id`) REFERENCES `file_uploads` (`id`) ON DELETE SET NULL,
-    FOREIGN KEY (`version_id`) REFERENCES `versions` (`id`)
+    UNIQUE KEY `upload_id` (`upload_id`),
+    UNIQUE KEY `version_id` (`version_id`),
+
+    CONSTRAINT `yara_results_upload_id_5cf355f9_fk_file_uploads_id` FOREIGN KEY (`upload_id`) REFERENCES `file_uploads` (`id`) ON DELETE SET NULL,
+    CONSTRAINT `yara_results_version_id_b32a0f70_fk_versions_id` FOREIGN KEY (`version_id`) REFERENCES `versions` (`id`) ON DELETE CASCADE
 );
--- a/src/olympia/migrations/1107-yara.sql
+++ b/src/olympia/migrations/1107-yara.sql
@@ -2,10 +2,10 @@ CREATE TABLE `yara_results` (
     `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
     `created` datetime(6) NOT NULL,
     `modified` datetime(6) NOT NULL,
-    `upload_id` int(11) NOT NULL,
+    `upload_id` int(11),
     `version_id` integer unsigned,
     `matches` longtext,
 
-    FOREIGN KEY (`upload_id`) REFERENCES `file_uploads` (`id`),
+    FOREIGN KEY (`upload_id`) REFERENCES `file_uploads` (`id`) ON DELETE SET NULL,
     FOREIGN KEY (`version_id`) REFERENCES `versions` (`id`)
 );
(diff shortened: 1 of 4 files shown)


eviljeff:
I don't know if you strictly _need_ a `__init__.py` file in every app's folder these days, but it's best practise too.

willdurand:
> test for models.py please 😉

Thanks. I wonder what needs to be tested here? Would you have any pointers? I looked at several `test_models.py` and most tests are for computed props and other methods on the model.

willdurand:
> I don't know if you strictly _need_ a `__init__.py` file in every app's folder these days, but it's best practise to.

ah, good call!

src/olympia/yara/models.py
```diff
@@ -0,0 +1,20 @@
+from django.db import models
+from django_extensions.db.fields.json import JSONField
+
+from olympia.amo.models import ModelBase
+from olympia.files.models import FileUpload
+
+
+class YaraResults(ModelBase):
```
eviljeff: maybe you want this to be called just `YaraResult` to avoid having specify the plural

willdurand: I did that initially but it sounds weird.. Maybe I should pick a different name than `results` (although I like it).

willdurand:
> Thanks. I wonder what needs to be tested here? Would you have any pointers? I looked at several `test_models.py` and most tests are for computed props and other methods on the model.

I ended up adding a smoke test for the creation, and I added a new method. I wanted to add it later, but that makes sense here too.

src/olympia/yara/models.py
willdurand: I switched to `YaraResult` 🤷‍♀

willdurand:
I just noticed that I forgot to add a `__init__.py` file in the `tests/` folder. This is done.

src/olympia/yara/tests/test_models.py
```diff
@@ -0,0 +1,33 @@
...
+from olympia.files.models import FileUpload
+from olympia.yara.models import YaraResult
+
+
+class TestYaraResult(TestCase):
+    def test_create(self):
+        addon = addon_factory()
+        upload = FileUpload.objects.create(addon=addon)
+
+        r = YaraResult.objects.create(upload=upload)
```
EnTeQuAk: nitpick, let's use names > 3 characters, `result` should be working fine here.

src/olympia/yara/tests/test_models.py
```diff
@@ -0,0 +1,33 @@
...
+
+        assert r.id is not None
+        assert r.upload == upload
+        assert r.matches == []
+        assert r.version is None
+
+    def test_add_match(self):
+        addon = addon_factory()
+        upload = FileUpload.objects.create(addon=addon)
+        r = YaraResult.objects.create(upload=upload)
```
EnTeQuAk: Same here

src/olympia/yara/models.py
```diff
@@ -0,0 +1,26 @@
...
+    matches = JSONField(default=[])
+    version = models.OneToOneField('versions.Version',
+                                   related_name='yara_results',
+                                   on_delete=models.CASCADE,
+                                   null=True)
+
+    class Meta:
+        db_table = 'yara_results'
+
+    def add_match(self, rule, tags=[], meta={}):
```
EnTeQuAk: Don't use `{}` or `[]` (mutable objects) as defaults in Python arguments.

```python
In [2]: def append(num, numlist=[]):
   ...:     numlist.append(num)
   ...:     print(numlist)
   ...:     return numlist
   ...:

In [3]: append(5)
[5]
Out[3]: [5]

In [4]: append(7)
[5, 7]
Out[4]: [5, 7]

In [5]: append(2)
[5, 7, 2]
Out[5]: [5, 7, 2]
```

Use `None` as default values and do something like `[] if tags is not None else tags` or `tags or []` which should both work fine.

willdurand: ah, right. I keep forgetting this..

src/olympia/yara/models.py
```diff
@@ -0,0 +1,26 @@
...
+from django_extensions.db.fields.json import JSONField
+
+from olympia.amo.models import ModelBase
+from olympia.files.models import FileUpload
+
+
+class YaraResult(ModelBase):
+    upload = models.OneToOneField(FileUpload,
+                                  related_name='yara_results',
+                                  on_delete=models.CASCADE)
```
eviljeff: Note: `FileUpload` instances are _supposed_ to be a short term thing before the addon/version has been created, and ignored afterwards.  There's a cron job that deletes old `FileUpload`s, which would (because `on_delete=models.CASCADE`) end up deleting the `YaraResult` too (though I think the cron might be broken).  So you possibly don't want `CASCADE` for this fk.

eviljeff: It depends on how long you intend for `YaraResult`s to be kept around, of course.

willdurand: Ah, correct. I believe removing a `FileUpload` should not remove the `YaraResult` because this relation is only useful during the submission process, until we can attach a `Version` to the `YaraResult`. So I changed (1) the column to be nullable and (2) the FK constraint to `SET NULL` on delete.

willdurand: I tried to delete a `FileUpload` and [everything](https://github.com/mozilla/addons-server/compare/master...willdurand:yara) continues to work as expected.

src/olympia/migrations/1107-yara.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE `yara_results` (
+    `id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY,
+    `created` datetime(6) NOT NULL,
+    `modified` datetime(6) NOT NULL,
+    `upload_id` int(11),
+    `version_id` integer unsigned,
+    `matches` longtext,
+
+    FOREIGN KEY (`upload_id`) REFERENCES `file_uploads` (`id`) ON DELETE SET NULL,
+    FOREIGN KEY (`version_id`) REFERENCES `versions` (`id`)
```
eviljeff: This sql is missing the named `CONSTRAINT` we usually have.  Fwiw, I find the easiest way to get the exact same sql as django would generate is to run the tests with `--create-db` then grab the sql from the `olympia_test` database with `SHOW CREATE TABLE xxx`.

willdurand: Thanks!

src/olympia/yara/tests/test_models.py
```diff
@@ -0,0 +1,33 @@
...
+        rule = 'some-yara-rule'
+        tags = ['foo']
+        meta = {'description': 'some description for some-yara-rule'}
+        result.add_match(rule=rule, tags=tags, meta=meta)
+
+        assert result.matches == [{
+            'rule': rule,
+            'tags': tags,
+            'meta': meta,
+        }]
```
diox: You could add a simple test creating a yararesult with a fileupload attached and then delete the fileupload and make sure the yararesult is still here, proving you've set `on_delete` correctly.

willdurand: just added that

Issue #11780, opened 2019-07-03
Implement admin for Secondary Hero Shelf Promo modules


jvillalobos:
Part of [this PRD](https://docs.google.com/document/d/12aD_v5AOj1Y6VmT4xOuyNcrKfZxqhP8R_AOHroQVLV8/edit#).

The Secondary Hero Shelf in the homepage will have three promotional modules with an icon, some text and a Call to Action:

<img width="623" alt="Screen Shot 2019-07-03 at 4 22 47 PM" src="https://user-images.githubusercontent.com/62568/60628507-ee561500-9dae-11e9-840b-d64f60697694.png">

We need an admin tool for editing the contents of these modules:

* Icon, chosen from an existing pool
* Description, up to 50 characters, localized
* Call to Action, up to 20 characters, localized
* URL

The localization should work in the same way as the copy for Recommended extensions. The pool of icons will probably be [these](https://design.firefox.com/icons/viewer/#), but please confirm with Product when you start working on this :). There will always be exactly 3 modules.

The first version of this module will have the same three "guide page" modules that we currently have on the homepage (Call to Action is "Learn More", icons are TBD).

muffinresearch:
See attachment for SVG icons from Scott: [icons.zip](https://github.com/mozilla/addons-server/files/3411347/icons.zip)

AlexandraMoga:
Verified on -dev with Fx68

Adding to the results mentioned in https://github.com/mozilla/addons-server/issues/11779#issuecomment-519093310:
- each secondary hero needs to have three modules attached to it, otherwise the form can't be saved
- each module needs to have a corresponding icon and description, otherwise the form can't be saved 
- all the enabled entries are returned by the `/v4/api/hero/secondary/`

There is an issue with the icons and their radio buttons, for which I filed #12097

PR #11988, opened 2019-07-31
add secondaryhero shelf - shelf&blurb

### LINKED ISSUE #11779: Implement admin for Secondary Hero Shelf Blurb module
Part of [this PRD](https://docs.google.com/document/d/12aD_v5AOj1Y6VmT4xOuyNcrKfZxqhP8R_AOHroQVLV8/edit#).

The Secondary Hero Shelf in the homepage will have a module with some text and a Call to Action on the left-hand side:

Screen Shot 2019-07-03 at 4 22 47 PM

We need an admin tool for editing the contents of this module:

* Headline, up to 50 characters, localized
* Description, up to 100 characters, localized
* Call to Action, up to 20 characters, localized, optional
* URL, optional

The localization should work in the same way as the copy for Recommended extensions. The Call to Action and URL go together, so it shouldn't be possible to have one without the other.

The first version of this module will have similar content as currently on the homepage:
* Headline: "Extensions are like apps for your browser."
* Description: "They add features to Firefox to make browsing faster, smarter, or just plain fun."
* Call to Action: "See all extensions"
* URL: https://addons.mozilla.org/extensions/

eviljeff:
fixes #11779 - not particularly useful until #11780 is done too, admittedly.


eviljeff:
rebase to fix conflicts

docs/topics/api/hero.rst
```diff
@@ -15,13 +15,15 @@ Combined Hero Shelves
 
 .. _hero-shelves:
 
-This convienence endpoint serves a single, randomly selected, primary hero shelf.
+This convienence endpoint serves a single, randomly selected, primary hero shelf,
```
EnTeQuAk: the comma at the end shouldn't be needed?

docs/topics/api/hero.rst
```diff
@@ -15,13 +15,15 @@ Combined Hero Shelves
...
+and a single, randomly selected secondary hero shelf.
 
 
 .. http:get:: /api/v4/hero/
 
     :query string lang: Activate translations in the specific language for that query. (See :ref:`translated fields <api-overview-translations>`)
     :>json object primary: A :ref:`primary hero shelf <primary-hero-shelf>`.
+    :>json object secondary: A :ref:`primary hero shelf <primary-hero-shelf>`.
```
EnTeQuAk: I think you wanted to link to the **secondary** hero shelf item here

docs/topics/api/hero.rst
```diff
@@ -46,3 +48,23 @@ small number of shelves this endpoint is not paginated.
...
+
+This endpoint returns all enabled secondary hero shelves.  As there will only ever be a
+small number of shelves - and likely only one - this endpoint is not paginated.
+
+
+.. http:get:: /api/v4/hero/secondary/
+
+    :>json array results: The array containing the results for this query.
+    :>json string results[].headline: The headline for this item. May contain some HTML tags.
+    :>json string results[].description: The description for this item. May contain some HTML tags.
```
EnTeQuAk: I wonder if we should mention if they're escaped or not here?

src/olympia/hero/models.py
```diff
@@ -101,3 +101,21 @@ def clean(self):
...
+    description = models.CharField(max_length=100, blank=False)
+    cta_url = models.CharField(max_length=255, blank=True)
+    cta_text = models.CharField(max_length=20, blank=True)
+    enabled = models.BooleanField(db_index=True, null=False, default=False)
+
+    def __str__(self):
+        return str(self.headline)
+
+    def clean(self):
+        both_or_neither = not (bool(self.cta_text) ^ bool(self.cta_url))
```
EnTeQuAk: 🤓

src/olympia/hero/tests/test_views.py
```diff
@@ -67,14 +76,53 @@ def test_basic(self):
...
+
+        hero_a.update(enabled=True)
+        hero_b.update(enabled=True)
+        # don't enable the 3rd PrimaryHero object
+        response = self.client.get(self.url)
+        assert response.status_code == 200
+        assert response.json() == {
+            'results': [
+                SecondaryHeroShelfSerializer(instance=hero_a).data,
+                SecondaryHeroShelfSerializer(instance=hero_b).data]}
```
EnTeQuAk: Either here or in the serializer; we should test that html-tags are exported the way we expect them to.

docs/topics/api/hero.rst
eviljeff: Depends if you think there should be a pause in the sentence.

docs/topics/api/hero.rst
eviljeff: I copied this line from the discovery api documentation (and the primary shelf documentation above).  Currently it's just a char field that could contain literally anything entered in the admin, and when #11989 is implemented it'll be appropriately escaped as per the discovery api/model.  I don't know what is appropriate documentation for this at this point.  (Or even if frontend would expect html or text only for these fields until Bob is back)

src/olympia/hero/tests/test_views.py
eviljeff: Perhaps I'll just leave out all mention of escaping until the html (escaping) has been implemented (or it's been confirmed frontend doesn't want html anyway)

docs/topics/api/hero.rst
eviljeff: https://github.com/mozilla/addons-server/pull/11988/commits/7ea155b54d52a51a8e9f29235d787574c59cac66

docs/topics/api/hero.rst
eviljeff: I decided to leave out any mention of html and escaping because it's not clear what the implementation will be yet.  If there ends up being html in #11989 I'll add docs then (and tests)

PR #11999, opened 2019-08-01
add logging statement for guid reuse

### LINKED ISSUE #11835: log add-on guid reuse to our logging pipeline
from https://github.com/mozilla/addons-server/pull/11824/ @EnTeQuAk said:
> Just an improvement idea: log add-on guid reuse to our logging pipeline, so that maybe in the future it can be part of malicious detection etc.

eviljeff:
fixes #11835 - is this sufficient to end up in the pipeline for anti-spam purposes?


wagnerand:
Could we log both the old and new PK? Looks great otherwise.

eviljeff:
Just the most recent previous pk?  We could get them all now (with `ReusedGUID.objects.filter(...)`)

wagnerand:
Hm, good question. Let's start with the most recent, and expand later if necessary.

Issue #11984, opened 2019-07-31
Create a Celery chords for tasks between `repack` and `handle_upload_validation_result`


willdurand:
After #9068, we should create a _chord_ to run the Celery tasks of the submission process in parallel. The chord should be executed after _repack_ and should end with `handle_upload_validation_result`.

Doing this will allow us to add more tasks in parallel, such as the yara task described in #11983.

willdurand:
@diox I suppose we could do that as soon as #11983 lands, no? It seems only the linter task would be part of the set of parallel tasks, right? Once we have that, it should be relatively easy to add the yara task (#11983).

diox:
Yes, and in fact that'd probably be a great thing to try without adding extra features like yara, just making a (dumb) chord and see if it still works.

tapaswenipathak:
Hello folks: Can I PR for the ticket?

EnTeQuAk:
> Hello folks: Can I PR for the ticket?

This issue isn't really suitable for contributors as.

Please take a look at other issues that have the [`contrib: welcome`](https://github.com/mozilla/addons-server/labels/contrib%3A%20welcome) label.

willdurand:
We decided that running the `chord` after the linter would make sense to only run the scanners and other tools on "valid" add-ons.

AlexandraMoga:
I've submitted samples for all the supported add-on types, I've triggered validation warnings and errors, I've submitted add-ons with many files. I haven't noticed any issues during upload and validation. 

@willdurand I've also used the upload script that uses the API signing endpoint and for a batch of 100 add-ons, one of them was created without a version. 
This is the example (you need to be an admin to see it):
https://addons-dev.allizom.org/en-US/developers/addon/e884c0b902ec4c4e8ac6/versions
I understand you observed this issue in prod as well?

willdurand:
>  I understand you observed this issue in prod as well?

I filed https://github.com/mozilla/addons-server/issues/12572, which should be related.

AlexandraMoga:
If the issue does exist in prod, it should not be a regression from this patch. 
I'll cover a few more submissions on stage and if everything goes well, I'm going to mark the issue as verified fixed.

PR #12002, opened 2019-08-01
Add a Django admin for `YaraResult`

### LINKED ISSUE #11982: Create a Django admin for `YaraResults`
In order to view the `yara` results, we need a simple read-only Django admin. The list view should display:

- upload id (link to the detail view)
- add-on name with a link to the reviewer page
- channel
- matched rules

Screen Shot 2019-07-31 at 11 09 24

The default detail view is probably fine, although we can probably pretty print the JSON column:

Screen Shot 2019-07-31 at 11 09 58

willdurand:
Fixes #11982 

---

I suppose this should be done at the end but I am waiting for #11984, so there we are. A few things:

- the add/edit/delete abilities have been removed
- not sure what else to test on this admin, I did not want to write assertions for HTML given that the admin is read-only for now
- I don't know how to restrict this admin to AMO Admins only (unless it is done already). I know about `has_module_permission` but I don't know what we should be doing here.
- I think we'd need a new permission, eventually. This dashboard should be visible to AMO admins for now but maybe it'd make sense that *I* (and other people in our team) have access to it

You can add some data into the `yara_results` database to try this new admin:

```sql
-- without any version attached and a single match
insert into yara_results (created, modified, upload_id, matches) values (now(), now(), 1, '[{"rule":"dummy-rule","tags":[],"meta":{"description": "desc for dummy-rule"}}]');
-- without any version attached and two matches
insert into yara_results (created, modified, upload_id, matches) values (now(), now(), 2, '[{"rule":"dummy-rule","tags":[],"meta":{"description": "desc for dummy-rule"}}, {"rule":"dummy-rule-2","tags":[],"meta":{"description": "desc for dummy-rule-2"}}]');
-- without any version attached and no matches
insert into yara_results (created, modified, upload_id, matches) values (now(), now(), 3, '[]');
-- with a version attached, no upload and a single match
insert into yara_results (created, modified, matches, version_id) values (now(), now(), '[{"rule":"dummy-rule","tags":[],"meta":{"description": "desc for dummy-rule"}}]', 30);
```

## Screenshots

List view:

Screen Shot 2019-08-01 at 19 48 45

Detail view:

Screen Shot 2019-08-01 at 19 25 48

Screen Shot 2019-08-01 at 19 25 51

Screen Shot 2019-08-01 at 19 25 54

### DIFF
--- a/src/olympia/yara/admin.py
+++ b/src/olympia/yara/admin.py
@@ -1,0 +1,54 @@
+import json
+
+from django.contrib import admin
+from django.utils.html import format_html
+
+from olympia import amo
+from olympia.amo.urlresolvers import reverse
+
+from .models import YaraResult
+
+
+@admin.register(YaraResult)
+class YaraResultAdmin(admin.ModelAdmin):
+    actions = None
+
+    list_display = ('id', 'formatted_addon', 'channel', 'matched_rules')
+    list_select_related = ('version',)
+
+    fields = ('id', 'upload', 'formatted_addon', 'channel',
+              'formatted_matches')
+
+    # Remove the "add" button
+    def has_add_permission(self, request):
+        return False
+
+    # Remove the "delete" button
+    def has_delete_permission(self, request, obj=None):
+        return False
+
+    # Read-only mode
+    def has_change_permission(self, request, obj=None):
+        return False
+
+    def formatted_addon(self, obj):
+        if obj.version:
+            return format_html(
+                '<a href="{}">{} (version: {})</a>',
+                reverse('reviewers.review', args=[obj.version.addon.slug]),
+                obj.version.addon.name,
+                obj.version.id
+            )
+        return "-"
+    formatted_addon.short_description = 'Add-on'
+
+    def channel(self, obj):
+        if obj.version:
+            return ('listed' if obj.version.channel ==
+                    amo.RELEASE_CHANNEL_LISTED else 'unlisted')
+        return "-"
+    channel.short_description = 'Channel'
+
+    def formatted_matches(self, obj):
+        return format_html('<pre>{}</pre>', json.dumps(obj.matches, indent=4))
+    formatted_matches.short_description = 'Matches'
--- a/src/olympia/yara/admin.py
+++ b/src/olympia/yara/admin.py
@@ -12,6 +12,7 @@ from .models import YaraResult
 @admin.register(YaraResult)
 class YaraResultAdmin(admin.ModelAdmin):
     actions = None
+    view_on_site = False
 
     list_display = ('id', 'formatted_addon', 'channel', 'matched_rules')
     list_select_related = ('version',)
--- a/src/olympia/yara/tests/test_admin.py
+++ b/src/olympia/yara/tests/test_admin.py
@@ -16,7 +16,7 @@ class TestYaraResultAdmin(TestCase):
         super().setUp()
 
         self.user = user_factory()
-        self.grant_permission(self.user, '*:*')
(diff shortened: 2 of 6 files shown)


src/olympia/yara/admin.py
```diff
@@ -0,0 +1,54 @@
...
+from olympia import amo
+from olympia.amo.urlresolvers import reverse
+
+from .models import YaraResult
+
+
+@admin.register(YaraResult)
+class YaraResultAdmin(admin.ModelAdmin):
+    actions = None
+
```
diox: Add `view_on_site = False` as well (I guarantee you that's the first thing QA will try :stuck_out_tongue: )

src/olympia/yara/tests/test_admin.py
```diff
@@ -0,0 +1,94 @@
...
+    def test_has_change_permission(self):
+        assert self.admin.has_change_permission(request=None) is False
+
+    def test_formatted_addon(self):
+        addon = addon_factory()
+        version = version_factory(
+            addon=addon,
+            channel=amo.RELEASE_CHANNEL_LISTED
+        )
+        r = YaraResult(version=version)
```
diox: We typically avoid one-letter variables in favor of something more descriptive - in that case, I'd go for `result` (same for other instances of `r` below)

src/olympia/yara/tests/test_admin.py
```diff
@@ -0,0 +1,94 @@
...
+from olympia.yara.admin import YaraResultAdmin
+from olympia.yara.models import YaraResult
+
+
+class TestYaraResultAdmin(TestCase):
+    def setUp(self):
+        super().setUp()
+
+        self.user = user_factory()
+        self.grant_permission(self.user, '*:*')
```
diox: Modify `src/olympia/constants/permissions.py` to give the `yara.view_yararesult` permission to admins with `ADMIN_ADVANCED`, and use that permission instead of `*:*` (which very few people have, we try not to depend on that)

diox: Later we can decide on a more specific permission and add a new group with that, but I think for now advanced admins should be enough. Maybe @wagnerand has other ideas though

src/olympia/yara/tests/test_admin.py
willdurand: argh, right. Sorry for that.

src/olympia/yara/tests/test_admin.py
wagnerand: Wfm.

willdurand:
This is ready for another look I think, the CI build fails but it is not related to this patch. It tries to fetch a different repo/branch for some reason...

src/olympia/yara/tests/test_admin.py
```diff
@@ -0,0 +1,94 @@
...
+        self.grant_permission(self.user, 'Admin:Advanced')
+        self.client.login(email=self.user.email)
+        self.list_url = reverse('admin:yara_yararesult_changelist')
+
+        self.admin = YaraResultAdmin(model=YaraResult, admin_site=AdminSite())
+
+    def test_list_view(self):
+        response = self.client.get(self.list_url)
+        assert response.status_code == 200
+
```
diox: I like to add a simple test proving that without the specific permission (but an `Admin:<something>` permission left so that the user can still access django admin itself), `self.client.get(self.list_url)` returns a 403.

willdurand: Good call, thanks!

willdurand:
Merging as the circle-ci failure is not related to this patch (it's related to FxA and it's a known issue at the time of writing this comment).

PR #12004, opened 2019-08-02
optimize queries made per hero api request

### LINKED ISSUE #11971: optimize queries for hero views
We should check if the hero views aren't doing a stupid number of queries on each API request and optimize.

eviljeff:
fixes #11971

### DIFF
--- a/src/olympia/hero/views.py
+++ b/src/olympia/hero/views.py
@@ -1,3 +1,6 @@
+from django.db.transaction import non_atomic_requests
+from django.utils.decorators import classonlymethod
+
 from rest_framework.mixins import ListModelMixin
 from rest_framework.response import Response
 from rest_framework.views import APIView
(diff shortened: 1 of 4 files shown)


src/olympia/hero/views.py
```diff
@@ -25,6 +32,11 @@ def list(self, request, *args, **kwargs):
         # Simulate pagination-like results, without actual pagination.
         return Response({'results': serializer.data})
 
+    @classonlymethod
+    def as_view(cls, actions=None, **initkwargs):
+        view = super().as_view(actions=actions, **initkwargs)
+        return non_atomic_requests(view)
```
diox: With our `NonAtomicRequestsForSafeHttpMethodsMiddleware` that shouldn't be necessary for `GET` operations anymore. Did you notice extra the transaction statements when calling the API ?

eviljeff: I did, yeah.  There was a savepoint at the start and end.

diox: I think now that we have the middleware those only happens in tests, because of the way tests are wrapped in a transaction. But I guess that's fine to keep.

PR #12005, opened 2019-08-02
Create a run_yara task

### LINKED ISSUE #11983: Create a task to run `yara`
We need a new Celery task to run yara. The task will:

- only target web extensions
- ~~unzip file uploads~~ edit: we can directly read the content of (each file of) the xpi
- compile the yara rules
- run yara on all the files ~~of the unzipped upload~~
- create a new `YaraResult` (cf. #11981) object
- ~~delete the unzipped file upload~~

If an exception occurs, we should log it but we should not raise, otherwise it will break the submission flow (and that's not what we want).

In addition to that, we also need a waffle ~~flag~~ switch to enable/disable this task. **– Edit:** this will be implemented in https://github.com/mozilla/addons-server/issues/12006

Important: This task should be part of a Celery chord once we upgrade to Celery 4.2.

willdurand:
Fixes #11983 

---

This patch adds a new `run_yara` task that is NOT run yet (because we need #11984).

In order to try this task, you need to:

1. clone our private yara repository in the `addons-server` folder
2. apply the patch below (copy this patch then `cat` it to `patch -p1` and voilà):

```diff
﻿﻿﻿﻿diff --git a/src/olympia/devhub/tasks.py b/src/olympia/devhub/tasks.py
index 5f551e3a90..213111239a 100644
--- a/src/olympia/devhub/tasks.py
+++ b/src/olympia/devhub/tasks.py
@@ -5,6 +5,7 @@ import json
 import os
 import subprocess
 import tempfile
+import waffle
 
 from copy import deepcopy
 from decimal import Decimal
@@ -42,6 +43,7 @@ from olympia.files.utils import (
 from olympia.files.tasks import repack_fileupload
 from olympia.versions.models import Version
 from olympia.devhub import file_validation_annotations as annotations
+from olympia.yara.tasks import run_yara
 
 
 log = olympia.core.logger.getLogger('z.devhub.task')
@@ -266,6 +268,13 @@ def handle_upload_validation_result(
     FileUpload instance."""
     upload = FileUpload.objects.get(pk=upload_pk)
 
+    # Run Yara. This cannot be asynchronous because we have no way to know
+    # whether the task will complete before we attach a `Version` to it
+    # later in the submission process...
+    # TODO: move this task in the submission chord once it is possible.
+    # TODO: use a waffle switch to enable/disable this task.
+    run_yara(upload.pk)
+
     # Check for API keys in submissions.
     # Make sure it is extension-like, e.g. no search plugin
     try:
diff --git a/src/olympia/versions/models.py b/src/olympia/versions/models.py
index c76fd08945..5f351122e7 100644
--- a/src/olympia/versions/models.py
+++ b/src/olympia/versions/models.py
@@ -35,6 +35,7 @@ from olympia.files import utils
 from olympia.files.models import File, cleanup_file
 from olympia.translations.fields import (
     LinkifiedField, PurifiedField, TranslatedField, save_signal)
+from olympia.yara.models import YaraResult
 
 from .compare import version_dict, version_int
 
@@ -262,6 +263,15 @@ class Version(OnChangeMixin, ModelBase):
         storage.delete(upload.path)
         version_uploaded.send(sender=version)
 
+        try:
+            yara_result = YaraResult.objects.get(upload_id=upload.id)
+            if yara_result:
+                yara_result.version = version
+                yara_result.save()
+        except:
+            log.exception('Error trying to assign a Version from upload %s '
+                          'to a YaraResult', upload.id)
+
         # Extract this version into git repository
         transaction.on_commit(
             lambda: extract_version_to_git_repository(version, upload))
```

3. upload a new add-on (see our slack channel for an example)

A few other things to consider:

- We might find another way to consume the private yara rules in the near future, but this works well: it's straightforward and it works with Docker. We'll need a solution for when we'll deploy and run this task though, but that's still under discussion.
- The task will be behind a waffle switch. I'd like to only add the task to the chord when the switch is `on` (see: https://github.com/mozilla/addons-server/compare/master...willdurand:yara)


willdurand:
I don't think the circle-ci failure is caused by this patch, so it should be ready for a review.

src/olympia/yara/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+from olympia.files.models import FileUpload
+from olympia.files.utils import SafeZip
+
+from .models import YaraResult
+
+log = olympia.core.logger.getLogger('z.yara.task')
+
+
+@task
+def run_yara(upload_pk):
```
diox: I know the whole repo is lacking in that department, but a docstring would be nice.

And in fact it may be a good place to document that this task can *only* work as part of the submission process - because otherwise it's too late, we keep old `FileUpload`s for a little while but we move the xpi as soon as we create the version. It tripped me up for a second when I tried to re-run the task on an add-on I had created earlier.

Issue #12007, opened 2019-08-02
`YaraResult.matched_rules` should return a unique list


willdurand:
It is possible to have a Yara rule that matches more than once for a given add-on, most likely because the rules are applied to all the files and it is possible that a rule gets triggered by more than one of these files.

The `matched_rules` property on the `YaraResult` model should return a unique list of rule names, instead of returning all the rule names and potentially duplicate names.

AlexandraMoga:
Please add STR or the qa:not needed label
Thanks

PR #12008, opened 2019-08-02
YaraResult.matched_rules should return unique rule names

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12007

### DIFF
--- a/src/olympia/yara/models.py
+++ b/src/olympia/yara/models.py
@@ -28,4 +28,4 @@ class YaraResult(ModelBase):
 
     @property
     def matched_rules(self):
-        return [match['rule'] for match in self.matches]
+        return sorted({match['rule'] for match in self.matches})
--- a/src/olympia/yara/tests/test_models.py
+++ b/src/olympia/yara/tests/test_models.py
@@ -76,3 +76,14 @@ class TestYaraResult(TestCase):
             result.add_match(rule=match.rule, tags=match.tags, meta=match.meta)
 
         assert result.matched_rules == [rule1, rule2]
+
+    def test_matched_rules_returns_unique_list(self):
+        result = self.create_yara_result()
+        rule1 = 'rule-1'
+        rule2 = 'rule-2'
+
+        for rule in [rule1, rule2, rule1, rule2]:
+            match = self.create_fake_yara_match(rule=rule)
+            result.add_match(rule=match.rule, tags=match.tags, meta=match.meta)
+
+        assert result.matched_rules == [rule1, rule2]


src/olympia/yara/models.py
```diff
@@ -28,4 +28,9 @@ def add_match(self, rule, tags=None, meta=None):
 
     @property
     def matched_rules(self):
-        return [match['rule'] for match in self.matches]
+        matched_rules = []
+        for match in self.matches:
+            rule = match['rule']
+            if rule not in matched_rules:
+                matched_rules.append(rule)
+        return matched_rules
```
eviljeff: return `list({match['rule'] for match in self.matches})`

willdurand: Will this keep the order of the items in the list? I am not sure I need this but when I used `list(set())`, another test case failed because `rule1` came after `rule2`.

eviljeff: It won't.  Though, thinking about it, if having duplicates is a problem why not check uniqueness in `add_match` instead?

willdurand: We should store all the occurrences in database because it will give hints about which rule matched on which file. This computed property is used in the list view of the Django admin.

Maybe it does not matter if the order is not preserved?

eviljeff: I wouldn't have thought it would matter, except for readability in django admin, though I don't have full sight of the whole project so 🤷‍♂

willdurand: This property is used in the right column here:



I suppose if we don't preserve the order and a user refreshes the list many times, then the values in this column will likely change, but that does not look like a huge problem... So maybe we should keep things simple?

diox: You could return fully `sorted` rules every time to make the order stable ?

willdurand: Yep, that would make sense, I guess. So `list(sorted(set()))`?

eviljeff: `sorted({match['rule'] for match in self.matches})` should be enough - `sorted` returns a list from an iterable.

src/olympia/yara/models.py
```diff
@@ -28,4 +28,4 @@ def add_match(self, rule, tags=None, meta=None):
 
     @property
     def matched_rules(self):
-        return [match['rule'] for match in self.matches]
+        return sorted({match['rule'] for match in self.matches})
```
EnTeQuAk: It doesn't really make sense to use a `set` here instead of a list comprehension. Actually, you can just use a generator instead too `sorted((match['rule'] for match in self.matches))` - although I think using a list comprehension here might even be the "fastest" (not that we care…) since it doesn't need to be computed on every iteration compared to a generator or a set (where we have the unique checks every time).

PR #12009, opened 2019-08-02
drop Personas model

### LINKED ISSUE #11801: theme_update_counts_from_file management command got OOM killed
There is a cron job that runs `./manage.py theme_update_counts_from_file` daily (i.e. https://github.com/mozilla/addons-server/blob/master/scripts/crontab/crontab.tpl#L32). This job wasn't able to finish for 2019-07-04 date and it got OOM killed in the middle of running.

The kernel message showed the process used more than 6 GB memory before it's killed.

```
[Jul 6 04:48] gollum invoked oom-killer: gfp_mask=0x201da, order=0, oom_score_adj=0
[  +0.001888] gollum cpuset=/ mems_allowed=0
[  +0.001039] CPU: 1 PID: 2132 Comm: gollum Not tainted 3.10.0-862.14.4.el7.x86_64 #1

... ... 

[  +0.001939] Out of memory: Kill process 625 (python) score 779 or sacrifice child
[  +0.001822] Killed process 625 (python) total-vm:6469572kB, anon-rss:6276464kB, file-rss:0kB, shmem-rss:0kB
```

### What happened?

We were alerted for missing stats for 2019-07-04, so we tried to manually re-run all the stats related cron jobs for the date, and during the process, we found that the `theme_update_counts_from_file` job wasn't able to finish and got OOM killed.

The job to finish without being OOM killed.

N/A

### LINKED ISSUE #12010: drop the Persona model
Now we don't have any LWT we can drop the `Persona` model and all references to it.

eviljeff:
fixes #12010 + resolves #11801 too


src/olympia/addons/models.py
```diff
@@ -1603,195 +1590,8 @@ class AddonReviewerFlags(ModelBase):
...
-    @cached_property
-    def listed_authors(self):
-        return self.addon.listed_authors
-
-
 class MigratedLWT(OnChangeMixin, ModelBase):
-    lightweight_theme = models.ForeignKey(
-        Addon, unique=True, related_name='migrated_to_static_theme',
-        on_delete=models.CASCADE)
+    lightweight_theme_id = models.PositiveIntegerField()
```
eviljeff: This is so we can keep the MigratedLWT model around and serve updates after the LWT have been hard-deleted.  I'm assuming changing this and dropping the constraint (see migration) is enough.

src/olympia/stats/models.py
```diff
@@ -88,23 +88,6 @@ class Meta:
...
-    per Persona).
-
-    """
-    id = PositiveAutoField(primary_key=True)
-    persona_id = models.PositiveIntegerField()
-    popularity = models.PositiveIntegerField()
-    movers = models.FloatField()
-
-    class Meta:
-        db_table = 'theme_update_counts_bulk'
```
diox: Do we have an issue to drop tables for models like this one that you're removing?

eviljeff: #12011

PR #12021, opened 2019-08-05
Fix wrong comment in description of queries

diox:


diox:
Ah yeah, I grepped for that earlier but somehow missed a couple in hero test views as well. I'll look into those.

PR #12023, opened 2019-08-05
Clear last_login_ip on users deleted for more than a year

### LINKED ISSUE #11560: Preserve IP addresses for banned users for a limited time

In order to fight bad actors, we need to preserve the `last_login_ip` of a banned user for some time.

### What happened?

`last_login_ip` gets cleared on account ban.

Keep `last_login_ip` for some time.

We should add a daily cronjob that clears IP addresses from banned users if their IP address hasn't been used by another user in 12 months (measured from the last login of the banned user. Ban date would be preferred, but that's a little harder to fetch, until we do #11458).

Legal and Trust/Security have signed-off on this. @muffinresearch can confirm if needed.

diox:
Fixes #11560

### DIFF
--- a/src/olympia/users/management/commands/clear_old_last_login_ip.py
+++ b/src/olympia/users/management/commands/clear_old_last_login_ip.py
@@ -1,0 +1,21 @@
+from datetime import datetime, timedelta
+
+from django.core.management.base import BaseCommand
+
+import olympia.core.logger
+from olympia.users.models import UserProfile
+
+
+log = olympia.core.logger.getLogger('z.users')
+
+
+class Command(BaseCommand):
+    help = 'Clear last_login_ip on users deleted more than a year ago'
+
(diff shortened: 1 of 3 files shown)


src/olympia/users/management/commands/clear_old_last_login_ip.py
```diff
@@ -0,0 +1,21 @@
...
+log = olympia.core.logger.getLogger('z.users')
+
+
+class Command(BaseCommand):
+    help = 'Clear last_login_ip on users deleted more than a year ago'
+
+    def handle(self, *args, **options):
+        a_year_ago = datetime.now() - timedelta(days=365)
+        qs = UserProfile.objects.filter(
+            deleted=True, modified__lt=a_year_ago).exclude(last_login_ip='')
```
wagnerand: Don't you want `modified__gt` here?

Issue #11771, opened 2019-07-03
Add metrics for account restrictions


muffinresearch:
The goal here is to add metrics for when a restriction prevents an action. This way we can graph the number of restrictions over time.

diox:
It can be useful, so I'll add it, but I think it'll be difficult to interpret. Failing a restriction check can happen at submission time, but also just when trying to get an API key. In addition, we'd log a failure every time the affected user reloads the page, re-submit their add-on, etc.

PR #12024, opened 2019-08-05
Increment a statsd counter whenever we allow/prevent a restricted action

diox:
Fixes https://github.com/mozilla/addons-server/issues/11771


diox:
I dunno. The problem is, since we're checking all restrictions, multiple classes might fail for a single request. So logging each class independently might make things even harder to track. But I can add a *second* statsd counter for that, I suppose.

PR #12025, opened 2019-08-05
 Do not expose heading_text in the discovery pane serializer. 

### LINKED ISSUE #11817: Don't show "Custom heading" in the discovery card
On the discovery pane and on the recommeded section in extensions manager, we're not going to display the heading. 
recommeded

cc @devaneymoz if more details are needed.

EnTeQuAk:
Fixes #11817

### DIFF
--- a/docs/topics/api/overview.rst
+++ b/docs/topics/api/overview.rst
@@ -332,6 +332,7 @@ v4 API changelog
 * 2019-07-25: added /hero/ endpoint to expose recommended addons and other content to frontend to allow customizable promos https://github.com/mozilla/addons-server/issues/11842.
 * 2019-08-01: added alias ``edition=MozillaOnline`` for ``edition=china`` in /discovery/ endpoint.
 * 2019-08-08: add support for externally hosted addons to /hero/ endpoints.  https://github.com/mozilla/addons-server/issues/11882
+* 2019-08-08: removed ``heading_text`` property from discovery api. https://github.com/mozilla/addons-server/issues/11817
 
 ----------------
 v5 API changelog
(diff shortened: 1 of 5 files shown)


diox:
Docs need to be updated. Also, we'll need to provide clear instructions for how to test this in-client to QA.

EnTeQuAk:
QA: We're going to have to make sure that…

* All Firefox releases with the recommended section work with this change (without ``heading_text`` in the API)
* The HTML based discovery pane used by older clients is also unaffected

Once this lands on -dev and on -dev later on it should be fairly straight forward to test / see any issues.

docs/topics/api/overview.rst
```diff
@@ -332,6 +332,7 @@ v4 API changelog
 * 2019-07-25: added /hero/ endpoint to expose recommended addons and other content to frontend to allow customizable promos https://github.com/mozilla/addons-server/issues/11842.
 * 2019-08-01: added alias ``edition=MozillaOnline`` for ``edition=china`` in /discovery/ endpoint.
 * 2019-08-08: add support for externally hosted addons to /hero/ endpoints.  https://github.com/mozilla/addons-server/issues/11882
+* 2019-08-08: removed ``heading_text`` property from discovery api. https://github.com/mozilla/addons-server/issues/11817
```
diox: In `discovery.rst` there is also a mention of `heading_text` that needs to be removed.

EnTeQuAk: :see_no_evil: apparently I never committed that… fixed in 1155b51ffa

eviljeff:
Is there a follow-up to this that removes the property off the model?  And presumably drops `heading` (the html variant) too because it'll be unused everywhere after the older discovery goes away?

EnTeQuAk:
> Is there a follow-up to this that removes the property off the model? And presumably drops heading (the html variant) too because it'll be unused everywhere after the older discovery goes away?

Not sure, I'll open an issue to ask that question at least. I think for now we just wanted to drop it from the api.

PR #12032, opened 2019-08-05
Create `ScannersResult` model and database table

### LINKED ISSUE #12027: Create database table and model for `ScannersResult`
In order to store the results of the different scanner executions, we need a table and a Django model, named `ScannersResult`. This model will have 1-1 relations with `FileUpload` (not null) and `versions.Version` (nullable). The model will store the JSON `results` and a reference to the scanner that has produced these results.

willdurand:
Fixes #12027 

---

- `SCANNERS` is a ~~list of constants~~ a list containing all the scanners with their human-friendly/code names
- ~~each scanner is declared with a constant and a strictly positive integer. Maybe `int(10)` is a bit too much here... WDYT?~~ switched to small integer
- Adding a new scanner would be a matter of defining a new constant and adding a new tuple to `SCANNERS`

This is part of [a bigger patch](https://github.com/mozilla/addons-server/compare/master...willdurand:scanners).

### DIFF
--- a/src/olympia/constants/scanners.py
+++ b/src/olympia/constants/scanners.py
@@ -1,0 +1,5 @@
+CUSTOMS = 1
+
+SCANNERS = [
+    (CUSTOMS, 'customs'),
+]
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -477,6 +477,7 @@ INSTALLED_APPS = (
     'olympia.pages',
     'olympia.ratings',
     'olympia.reviewers',
+    'olympia.scanners',
     'olympia.search',
     'olympia.stats',
     'olympia.tags',
--- a/src/olympia/migrations/1117-scanners.sql
+++ b/src/olympia/migrations/1117-scanners.sql
@@ -1,0 +1,17 @@
+CREATE TABLE `scanners_results` (
+    `id` int(11) NOT NULL AUTO_INCREMENT,
+    `created` datetime(6) NOT NULL,
+    `modified` datetime(6) NOT NULL,
+    `upload_id` int(11) DEFAULT NULL,
+    `results` longtext NOT NULL,
+    `scanner` smallint(5) unsigned NOT NULL,
+    `version_id` int(10) unsigned DEFAULT NULL,
+
+    PRIMARY KEY (`id`),
+
+    UNIQUE KEY `upload_id` (`upload_id`),
+    UNIQUE KEY `version_id` (`version_id`),
+
+    CONSTRAINT `scanners_results_upload_id_9259a7bf_fk_file_uploads_id` FOREIGN KEY (`upload_id`) REFERENCES `file_uploads` (`id`),
+    CONSTRAINT `scanners_results_version_id_dd07be31_fk_versions_id` FOREIGN KEY (`version_id`) REFERENCES `versions` (`id`)
+);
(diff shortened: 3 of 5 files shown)


eviljeff:
Isn't this part of the `yara` app you added?  Or `yara` would be part of this app?

willdurand:
> Isn't this part of the `yara` app you added? Or `yara` would be part of this app?

Yara and the scanners are quite different so there are 2 apps: `yara` and `scanners`.

eviljeff:
> Yara and the scanners are quite different so there are 2 apps: yara and scanners.

How are they different, conceptually?   Even the model looks the same (though they wouldn't necessarily have to share the same model)

willdurand:
> How are they different, conceptually? Even the model looks the same (though they wouldn't necessarily have to share the same model)

The way I see the different tools is:

- Yara is more a gate-keeper, standalone (in part because this is an existing tool/lib)
- The scanners (that we built) collect data that will be either shown to the reviewer and/or sent to another machine learning tool (api)

I found it cleaner to not merge both 🤷‍♂

EnTeQuAk:
> > 
> 
> The way I see the different tools is:
> 
>     * Yara is more a gate-keeper, standalone (in part because this is an existing tool/lib)
> 
>     * The scanners (that we built) collect data that will be either shown to the reviewer and/or sent to another machine learning tool (api)
> 
> 
> I found it cleaner to not merge both man_shrugging

Hmm, I think I have to agree with @eviljeff here. I didn't think about that earlier but if we are creating a `scanners` app with all the necessary code to run the various scanners it'd probably make sense to eventually move `run_addons_linter` from `devhub.tasks` there too to have one single point of entry to see how all our submission related (whether they're blocking or just collecting data is imho irrelevant) tasks.

Maybe @diox has another opinion about that.

willdurand:
> Hmm, I think I have to agree with @eviljeff here. I didn't think about that earlier but if we are creating a `scanners` app with all the necessary code to run the various scanners it'd probably make sense to eventually move `run_addons_linter` from `devhub.tasks` there too to have one single point of entry to see how all our submission related (whether they're blocking or just collecting data is imho irrelevant) tasks.

I agree.

I feel like `yara` should not belong to it, though. Yara won't share anything with the scanners, while the linter and the other scanners could be based on the same task code, similar api, no django admin, etc.

diox:
> each scanner is declared with a constant and a strictly positive integer. Maybe int(10) is a bit too much here... WDYT?

Doesn't really matter but you can use `PositiveSmallIntegerField` which gives you a `smallint(5)`

willdurand:
> Doesn't really matter but you can use `PositiveSmallIntegerField` which gives you a `smallint(5)`

ah yes, thanks! I missed this field :/

willdurand:
@diox any thoughts on the `app` debate?

diox:
I think it's too early to tell for sure, so I don't mind your layout.

willdurand:
Ok, then it should be ready for a review. The circle-ci build is failing for another reason.. and it's obscure.

EnTeQuAk:
> Ok, then it should be ready for a review. The circle-ci build is failing for another reason.. and it's obscure.

This is related to https://github.com/mozilla/addons-server/pull/12058 so yeah, don't worry about it.

willdurand:
I just updated this patch to avoid a migration ID conflict. R? ❤️

eviljeff:
Reminder to bump the migration # before merging.

PR #12036, opened 2019-08-06
Preload 'addon' and it's _current_version to avoid unneeded sql queries.

### LINKED ISSUE #12035: Optimize SQL for discovery items
Screenshot from 2019-08-06 08-49-32

EnTeQuAk:
This brings the count down from ~500 queries to <10 queries.

Fixes #12035

Screenshot from 2019-08-06 08-49-32

Screenshot from 2019-08-06 08-53-58

### DIFF
--- a/src/olympia/discovery/admin.py
+++ b/src/olympia/discovery/admin.py
@@ -4,8 +4,10 @@ from django.contrib.admin.widgets import ForeignKeyRawIdWidget
 from django.utils import translation
 from django.utils.html import format_html
 from django.utils.safestring import mark_safe
+from django.db.models import Prefetch
 
 from olympia.addons.models import Addon
+from olympia.versions.models import Version
 from olympia.discovery.models import DiscoveryItem
 from olympia.hero.admin import PrimaryHeroInline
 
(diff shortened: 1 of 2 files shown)


src/olympia/discovery/admin.py
```diff
@@ -81,7 +83,19 @@ class Media:
...
+        qset = (
+            DiscoveryItem.objects.all()
+            .select_related('primaryhero')
+            .prefetch_related(
+                Prefetch(
+                    'addon',
+                    queryset=Addon.unfiltered.all().only_translations()),
+                Prefetch(
+                    'addon___current_version',
+                    queryset=Version.unfiltered.all().only_translations())))
```
diox: Actually... if we want to push this further... I think you can remove that second prefetch, and add `select_related('_current_version')` on the `Addon` queryset that is being prefetched above. It will prevent versions transformers from running, including the translations one, but we don't need translations on `Version` here. Worth a try locally.

diox: Also in any case, we should add a comment for that method explaining that we're forced to use `prefetch_related()` because of translations, and that we use `only_translations()` because we don't care about the other transforms.

EnTeQuAk: Fixed in 3a791c7 and 6dfdbe3

Issue #11797, opened 2019-07-05
Review page is making way too many (duplicated) sql queries


diox:
Loading a simple content review page locally, I was greeted with this in the Django Debug Toolbar SQL panel:
> 413.36 ms (122 queries including 85 similar and 81 duplicates )

We're duplicating the query to fetch the add-on - and the ones made by transforms ! - multiple times, and also loading versions etc several times. Some might be unavoidable, but we should at least work to reduce that number and make it more bearable.

Note that this was with an add-on that had *one* version. It might be worse for add-ons that have more.

muffinresearch:
@EnTeQuAk any ideas on what a fix for this would take in terms of time?

EnTeQuAk:
Ideally not too much, I'll be looking at this today and tomorrow and see where we'll end up. Just noticed the higher priority issue #11902 and maybe investigation shows other slow downs too.

muffinresearch:
Marking this as p2 too since it's probably related to #11902.

EnTeQuAk:
Unassigning the milestone but keeping the issue open for now.

@diox and I looked into it and landed #12039 which should improve the speed already. We will file separate issues for all the other performance optimisations we may find.

lavish205:
> 413.36 ms (122 queries including 85 similar and 81 duplicates )

Could you please tell in which method/file these queries are happening?

diox:
@lavish205 this is in the `olympia.reviewers.views.review` page, but we've identified where most of the queries come from and filed new issues referencing this one - this is more a tracking issue now.

The extra queries are caused by deeply rooted issues that are hard to fix without being extremely familiar with our internals, though. Not really good bugs for contributors.

diox:
We've fixed the biggest problems in linked issues. The plan is to re-assess in production once today's push goes through, using reviewer feedback and the [dumb dashboard in Grafana](https://earthangel-b40313e5.influxcloud.net/d/sGEtfYcZk/amo-reviewer-tools-performance?orgId=1) I made - that dashboard is not super useful because of the low traffic reviewer tools get, but hopefully should be enough to see the results. If we notice enough of an improvement we'll close this issue.

diox:
> If we notice enough of an improvement we'll close this issue.

We did.

PR #12038, opened 2019-08-06
Reduce amount of queries on reviewe page

EnTeQuAk:
Refs #11797 but not necessarily fixes it.

On my tests we're down from ~130 queries to ~89 queries while reducing a few more in some edge-cases.

Unfortunately, to be honest, my debugging skills end here mostly and I can't figure out why version transforms still aren't properly run :disappointed:

Maybe someone else has an idea and is able to improve this number further.

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/review.html
+++ b/src/olympia/reviewers/templates/reviewers/review.html
@@ -75,7 +75,7 @@
   <div class="results">
     <div class="results-inner">
       <table id="review-files" class="item-history">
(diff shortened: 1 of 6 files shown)


EnTeQuAk:
> E   AssertionError: 122 != 89 : 122 queries executed, 89 expected

Thanks travis… apparently my local tests are relying on something travis doesn't do :-/

EnTeQuAk:
Closing this for now.

PR #12044, opened 2019-08-06
Dropping Personas and related tables

### LINKED ISSUE #12011: drop personas and related tables from DB
follow on from #12010 - can't be merged until after the 2019-08-08 tag, and we might want to keep the data around longer?

Tables:
`personas`
`theme_update_counts_bulk`
`theme_user_counts`

Abhi-khandelwal:
Fixes #12011


src/olympia/migrations/1113-drop-personas-and-related-tables.sql
```diff
@@ -0,0 +1,3 @@
+DROP TABLE IF EXISTS ` personas`;
```
eviljeff: you've got a space in the table name

PR #12058, opened 2019-08-07
Add additional docker compose 'ups'

jrbenny35:
Trying to fix the current error on CircleCI about a container not being found.

### DIFF
--- a/docker/uwsgi.ini
+++ b/docker/uwsgi.ini
@@ -17,10 +17,13 @@ enable-threads  = true
 
 # Run watchmedo (via watchdog) to implement unrestricted
 # autoreload capability.
-# fs-reload = %(base)/docker/artifacts/
-# safe-pidfile = %(base)/docker/artifacts/addons-server-uwsgi-master.pid
-# attach-daemon = setsid watchmedo shell-command --patterns="*.py" --recursive --command='/usr/bin/touch %(safe-pidfile)' %(base)
-# max-requests = 5000
+fs-reload = %(base)/docker/artifacts/
+# CI is only being set on circleci. See addons-server issue #12064
+if-not-env = CI
+safe-pidfile = %(base)/docker/artifacts/addons-server-uwsgi-master.pid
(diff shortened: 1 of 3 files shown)


docker/uwsgi.ini
```diff
@@ -17,10 +17,10 @@ enable-threads  = true
...
 # Run watchmedo (via watchdog) to implement unrestricted
 # autoreload capability.
-fs-reload = %(base)/docker/artifacts/
-safe-pidfile = %(base)/docker/artifacts/addons-server-uwsgi-master.pid
-attach-daemon = setsid watchmedo shell-command --patterns="*.py" --recursive --command='/usr/bin/touch %(safe-pidfile)' %(base)
-max-requests = 5000
+# fs-reload = %(base)/docker/artifacts/
+# safe-pidfile = %(base)/docker/artifacts/addons-server-uwsgi-master.pid
+# attach-daemon = setsid watchmedo shell-command --patterns="*.py" --recursive --command='/usr/bin/touch %(safe-pidfile)' %(base)
+# max-requests = 5000
```
eviljeff: I think we want this for local env, so you can't disable it generally

EnTeQuAk: I'd probably try to use https://uwsgi-docs.readthedocs.io/en/latest/ConfigLogic.html?highlight=if%20env#if-env and only set `safe-pidfile` and `attach-daemon` in case of that environment variable exists. Or, do the reverse, do something like

```ini
# MOZ_HEADLESS is only being set in docker-compose.selenium.yml which is only used by circleci
if-env = MOZ_HEADLESS
attach-daemon =
endif =
```

or so… this is untested, I don't know if unsetting variables like that works.

jrbenny35: Since this only effects CI why don't we use a circleci env variable. https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables, so ```if-not-env = CI```

jrbenny35:
Seems to work on CI

Issue #12006, opened 2019-08-02
Execute the `run_yara` task during submission (behind a waffle switch)


willdurand:
The `run_yara` task (cf. #11983) should run during submission. It should be part of the "submission" chord added in #11984. In order to enable/disable this task, we should use a waffle switch.

willdurand:
It would probably be a good idea to time this task (with statsd).

willdurand:
Marking this issue as blocked by https://github.com/mozilla/addons-server/issues/11984 (which is blocked by https://github.com/mozilla/addons-server/issues/9068), unless we are OK with running this task synchronously.

AlexandraMoga:
Same as #12208, should we be able to test this in -dev as well?

willdurand:
Added "state: blocked" since we cannot test in in -dev yet.

vcarciu:
I verified the issue using a code sample from @willdurand (eval(function(p,a,c,k,e,d) {});) and the "dean_edwards_packer" rule was correctly identified.

Marking issue as verified

PR #12059, opened 2019-08-07
Run yara task on submission (with waffle switch)

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12006

---

This patch allows to run `yara` on submission, but only if the linter results don't contain errors.

⚠️ It calls `run_yara` directly and does not fire a new task.

You can test it by:

1. cloning the yara repo in `/path/to/addons-server/private/`
2. turn the waffle switch named `enable-yara` on:

        ./manage.py waffle_switch enable-yara on


src/olympia/versions/models.py
```diff
@@ -262,6 +263,15 @@ def from_upload(cls, upload, addon, selected_apps, channel,
         storage.delete(upload.path)
         version_uploaded.send(sender=version)
 
+        try:
+            yara_result = YaraResult.objects.get(upload_id=upload.id)
+            if yara_result:
+                yara_result.version = version
+                yara_result.save()
+        except:
```
diox: Shouldn't need to catch anything other that `YaraResult.DoesNotExist`

willdurand:
I was thinking that, _maybe_, we could get this enabled in -dev if we manage to find a way to deploy the Yara rules.. but since we don't know how to that yet, I don't know..

willdurand:
@diox I know we mentioned lambda functions for the scanners but yara is a bit different, and if we are going to store the rules in the db in the future (likely), then we should keep the task in addons-server.

That would mean two things:

1. building private Docker images for deployment
2. adding the yara rules to these private images

It would require:

1. a deploy key added to the yara repo
2. the build step should clone the yara repo in `private/` after having cloned the addons-server repo

WDYT? If that could work, then we should ping ops to start addressing the two (1) and then (2).

diox:
I agree we should keep this one in addons-server. That deploy strategy makes sense to me, maybe ops have other suggestions though. In any case the details are mostly up to them.

willdurand:
@oremj @bqbn would it be possible to start working on [this deploy strategy](https://github.com/mozilla/addons-server/pull/12059#issuecomment-525269369) for -dev only?

willdurand:
@bqbn lets move the discussion to an issue, I filed https://github.com/mozilla/addons-server/issues/12211.

src/olympia/migrations/1124-run-yara-waffle-switch.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO waffle_switch (name, active, note, created, modified) VALUES ('enable-yara', 0, 'Enable the yara task during submission.', NOW(), NOW());
```
diox: Migration needs to be moved to a new number now

willdurand:
> Doesn't have to be in this PR, but can you add a couple of dumb `Makefile` actions to:
> 
>     * Initialize `yara` & `customs` clones in `private/`
> 
>     * Update `yara` & `customs`
> 
> 
> Maybe even a 3rd one to launch customs locally.

I was thinking about doing this but then we would disclose the git repo urls. I don't know if this is risky or not.

PR #12066, opened 2019-08-08
Correctly set domain in author confirmation emails

diox:
Follow-up https://bugzilla.mozilla.org/show_bug.cgi?id=1563534


AlexandraMoga:
This is verified fixed on -dev.

PR #12073, opened 2019-08-08
Upgrade Celery to 4.3

EnTeQuAk:
This PR upgrades to Celery 4.3, introduces a small hack because of a regression.

This PR does not introduce any helpers to make use of celery workers inside the tests, we delayed that to a later point to unblock the upgrade.

This PR does fix some inconsistencies with elasticsearch mocks along the way as they were making some tests fail (see comment logs), the mocks are now applied and rolled back much less than before which makes them easier to reason about and to debug.

As a follow-up I'm already working on a PR to move the repack task together with the validation tasks into a chain as mentioned in the comments from @diox

Any thoughts so far?

Notes:

* acks_late now doesn't acknowledge in case of failures or timeouts (which is the case for `index_addons`). We could use `task_acks_on_failure_or_timeout` but I figured the new behavior should be fine.
* any `CELERY_ALWAYS_EAGER` setting in local_settings.py must be replaced by `CELERY_TASK_ALWAYS_EAGER`


src/olympia/amo/tests/__init__.py
```diff
@@ -408,6 +411,32 @@ def activate_locale(locale=None, app=None):
...
+    for task_id in task_ids:
+        while task_id not in _celery_task_results:
+            if time_slept > max_wait:
+                if throw:
+                    raise AssertionError(
+                        f'Task didn\'t return in {max_wait} seconds')
+                else:
+                    break
+            time.sleep(0.05)
+            time_slept += 0.05
```
willdurand: It might be worth extracting `0.05` into a new variable.

src/olympia/amo/tests/test_celery.py
```diff
@@ -1,67 +1,129 @@
...
-        patch = mock.patch('olympia.amo.celery.cache')
-        self.cache = patch.start()
-        self.addCleanup(patch.stop)
 
-        patch = mock.patch('olympia.amo.celery.statsd')
-        self.statsd = patch.start()
-        self.addCleanup(patch.stop)
+@pytest.mark.celery_worker_test
+def test_celery_worker_test_runs_through_worker():
+    result = sleeping_task.delay()
```
willdurand: I think this test case would be easier to understand if `sleeping_task` was receiving a `sleep_time` argument set to `0.3` here.

src/olympia/amo/tests/test_celery.py
```diff
@@ -1,67 +1,129 @@
...
+    # then remove it as it won't be in CELERY_ROUTES
+    queues_in_queues.remove(settings.CELERY_TASK_DEFAULT_QUEUE)
+
+    queues_in_routes = set(
+        [c['queue'] for c in settings.CELERY_TASK_ROUTES.values()])
+    assert queues_in_queues == queues_in_routes
+
+
+@task(ignore_result=False)
+def fake_task():
```
willdurand: Most tasks here are "fake" but this one is probably an `ignore_result_task` (given that the other one is the default task, because it has no task options).

src/olympia/amo/tests/test_celery.py
```diff
@@ -1,67 +1,129 @@
...
+
+@pytest.mark.celery_worker_test
+def test_celery_default_ignore_result():
+    result = default_task.delay().get()
+    assert result is None
+
+
+@pytest.mark.celery_worker_test
+def test_celery_explicit_dont_ignore_result():
+    result = fake_task.delay().get()
```
willdurand: if you rename the task as mentioned above, then the test case would read better IMO

EnTeQuAk:
> I was expecting the diff to be much bigger, thanks +1 I left some basic comments only so someone else should take a look.

Heh, the bigger refactoring comes afterwards :see_no_evil: but isn't too much code, as most tests already exist. Thanks for the comments!

EnTeQuAk:
It seems, the hanging in the test is primarily being caused by the celery session worker not properly working under multi-processing tests being run (via pytest -n). I'm not sure yet why and what a proper fix would be…

EnTeQuAk:
Alrighty, tests are passing. I'll continue testing it locally and set it as ready to review asap.

src/olympia/amo/tests/test_send_mail.py
```diff
@@ -269,6 +269,11 @@ def attach_alternative(*args, **kwargs):
 
     @mock.patch('olympia.amo.tasks.EmailMessage')
     def test_async_will_retry_default(self, backend):
+        # Monkeypatch Celerys ".get()" inside async task error
+        # until https://github.com/celery/celery/issues/4661 (which isn't just
+        # about retries but a general regression) fixed.
```
diox: This comment and the clones below should mention that this celery issue manifests itself in eager mode so that we don't have a heart attack when quickly scanning comments in that file :)

src/olympia/lib/es/tests/test_commands.py
```diff
@@ -26,6 +26,10 @@ def setUp(self):
 
         self.addons = []
         self.expected = self.addons[:]
+        # Monkeypatch Celerys ".get()" inside async task error
+        # until https://github.com/celery/celery/issues/4661 (which isn't just
+        # about retries but a general regression) fixed.
```
diox: As above... but this test is doing threads to mimic a reindexation taking place so maybe it should use "real" celery and not eager mode ?

requirements/system.txt
```diff
@@ -120,6 +120,6 @@ six==1.12.0 \
     --hash=sha256:d16a0141ec1a18405cd4ce8b4613101da75da0e9a7aec5bdd4fa804d0e0eba73
 urllib3==1.25.3 \
     --hash=sha256:b246607a25ac80bedac05c6f282e3cdaf3afb65420fd024ac94435cabe6e18d1 \
-    --hash=sha256:dbe59173209418ae49d485b87d1681aefa36252ee85884c31346debd19463232 # pyup: <1.25
```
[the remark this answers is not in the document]
EnTeQuAk: No, not really. But I saw that and thought, making a separate PR just for that feels wrong. I can exclude it though.

src/olympia/lib/es/tests/test_commands.py
EnTeQuAk: Indeed, but tbh, I'd like to tackle that in a separate step. I spend quite some time to move this to a version where it uses a separate worker but continuously ran into transaction barriers. Also keeps the patch small (hoping to use that as an excuse…)

src/olympia/amo/tests/test_send_mail.py
EnTeQuAk: Fixed in 55fc2e2852

EnTeQuAk:
Interestingly it got fixed early this year apparently - https://github.com/tox-dev/tox/commit/88d28c136c788c701bcddf49f1f613498a6fd9c1 - maybe it's broken again?

EnTeQuAk:
Fixed in 0c81bfaf57 - let's see

EnTeQuAk:
The current hang in travis seems to be reproducible locally and related to a specific task not being registered when celery worker related tasks are being run together with other tests.

willdurand:
@EnTeQuAk do you know why this is still failing?

EnTeQuAk:
I re-implemented the state processing by using celery's own testing manager in eaf3047131 - let's see if this fixes the issues I had. This now waits much more efficiently for a certain state and should adapt to different environments (local vs travis) much easier.

EnTeQuAk:
Okay, this is still weird. It seems that the worker stops working at some point. I wonder why, I'll see if I can attach myself to all the worker events and get some more insight.

I wonder if one worker for the whole test session seems like a good idea but eventually trips up somehow…

EnTeQuAk:
Alrighty, it seems I am finally able to reproduce this issue by forcibly clearing any dangling caches

```diff
diff --git a/src/olympia/amo/tests/__init__.py b/src/olympia/amo/tests/__init__.py
index 3386152d76..21692b5aa3 100644
--- a/src/olympia/amo/tests/__init__.py
+++ b/src/olympia/amo/tests/__init__.py
@@ -1024,6 +1024,9 @@ class CeleryWorkerTestCase(TestCase):
         desc = f'waiting for tasks to be {states.join(",")}'

         def _is_result_task_started(results, **kwargs):
+            # Clear out any result caches we may have dangling around
+            [result.forget() for result in results]
+
```

Why this doesn't happen locally for me, I don't quite know yet. But it seems to be a start.

EnTeQuAk:
@diox so, for some context on this.

With the Celery 4.3 upgrade I implemented the ability to have tests use an actual celery worker instead of relying on always-eager mode. Unfortunately that specifically doesn't quite seem to work.

Specifically I'm talking about the failing tests in https://travis-ci.org/mozilla/addons-server/jobs/584588722#L1131-L1151 - they're only testing the in-testing workers - switching back to using always-eager mode fixes them immediately.

That'd be my scenario to unblock this PR, switch back to relying on always-eager for now and split the work to use actual worker threads into a separate issue/pr later on.

Mabye you have an idea why they're failing - or generally I'd be interested in your opinion about how much time we really want to spend to make use of actual worker-threads in our tests compared to using always-eager-mode.

I am able to reproduce the failing test locally now, using https://github.com/mozilla/addons-server/pull/12073#issuecomment-531233691 but that didn't quite help me discovering what is going on - for now I do have no idea.

setup.cfg
```diff
@@ -5,6 +5,8 @@ markers =
     es_tests: mark a test as an elasticsearch test.
     needs_locales_compilation: mark a test as needing compiled locales to work.
     allow_external_http_requests: mark a test to allow external http requests and disable responses.
+    celery_worker_tests: mark a test to require a running celery worker
+    static_assets: Tests that are depending on `compress_assets` and require to be separated with other tests.
```
EnTeQuAk: This is only here because of a pytest warning I fixed, happy to back it out.

diox:
> That'd be my scenario to unblock this PR, switch back to relying on always-eager for now and split the work to use actual worker threads into a separate issue/pr later on.
> 
> Maybe you have an idea why they're failing - or generally I'd be interested in your opinion about how much time we really want to spend to make use of actual worker-threads in our tests compared to using always-eager-mode.

No idea yet, but we can solve that later. So go ahead and switch back to always eager for now - We want to unblock the upgrade asap to start implementing and testing https://github.com/mozilla/addons-server/issues/11984

On avoiding always eager in general: I *hate* always eager, but as long as it's kinda supported, I'm not sure it's worth the effort to get rid of it in 99% of our tests. In most of our tests, the task isn't what's being tested, the fact that it's ran in eager mode is not great, it's kind of a hack, but it's fine for now.

EnTeQuAk:
> On avoiding always eager in general: I hate always eager, but as long as it's kinda supported, I'm not sure it's worth the effort to get rid of it in 99% of our tests. In most of our tests, the task isn't what's being tested, the fact that it's ran in eager mode is not great, it's kind of a hack, but it's fine for now.

Absolutely, that's also not what I wanted to do - that's why I created a `CeleryWorkerTestCase` class that works similar to the elasticsearch test case, only for tests that explicitly test the tasks.

Alrighty, I'll roll back the worker thread tests to use always-eager again and make the PR ready for review tomorrow morning. Thanks!

EnTeQuAk:
@diox I cleaned up the PR and updated the description. This should be ready for a review now.

willdurand:
> Absolutely, that's also not what I wanted to do - that's why I created a `CeleryWorkerTestCase` class that works similar to the elasticsearch test case, only for tests that explicitly test the tasks.
> 
> Alrighty, I'll roll back the worker thread tests to use always-eager again and make the PR ready for review tomorrow morning. Thanks!

@EnTeQuAk is there an existing issue to keep track of that?

requirements/prod.txt
```diff
@@ -126,12 +126,12 @@ bleach==3.1.0 \
...
     --hash=sha256:c96be2c71ed4893df29f0ce11cc3562109d9c14de6d7a1bd9f82c3356303fdf4
-celery==4.1.1 \
-    --hash=sha256:6fc4678d1692af97e137b2a9f1c04efd8e7e2fb7134c5c5ad60738cdd927762f \
-    --hash=sha256:d1f2a3359bdbdfb344edce98b8e891f5fe64f8a11c5a45538ec20ac237c971f5 # pyup: <4.2
 botocore==1.12.230 \
     --hash=sha256:18b7064b4bc6f7cb295aa48917eb91efe817cc0111ffbcb6559f50da0cee599d \
     --hash=sha256:30ff956bfef7dd91f35235ac705f9457f575071c2a577a01f8a2cf19eea95d02
+celery==4.3.0 \
+    --hash=sha256:4c4532aa683f170f40bd76f928b70bc06ff171a959e06e71bf35f2f9d6031ef9 \
+    --hash=sha256:528e56767ae7e43a16cfef24ee1062491f5754368d38fcfffa861cdb9ef219be
```
eviljeff: gah! the celery dependency was in the middle of the boto updates for a reason - otherwise each daily boto3 update causes a merge conflict with the accompanying botocore dep.

PR #12083, opened 2019-08-09
add a banned property for the date when a user is banned

### LINKED ISSUE #11876: Allow identification of banned accounts on database level
Currently, banned accounts are identified by the following conditions:

`users.deleted = 1 AND users.email IS NOT NULL AND users.fxa_id IS NOT NULL`.

This makes it problematic to to identify banned accounts on redash (which is needed for reasons), since we cannot expose the email to redash.

Let's add some way to identify banned users and distinguish them from users who deleted their accounts themselves.

One idea would be to add new column to the `users` table but I was told that migration would be quite expensive since the table is quite large.

eviljeff:
fixes #11876 - I guess?  I thought I'd have to change some check for banned-ness, but we only do on logon and the existing check is probably still appropriate?

### DIFF
--- a/src/olympia/users/tests/test_models.py
+++ b/src/olympia/users/tests/test_models.py
@@ -119,7 +119,7 @@ class TestUserProfile(TestCase):
         user.ban_and_disable_related_content()
         user.reload()
         assert user.deleted
-        assert user.banned
+        self.assertCloseToNow(user.banned)
         assert user.email == 'jbalogh@mozilla.com'
         assert user.auth_id
         assert user.fxa_id == '0824087ad88043e2a52bd41f51bbbe79'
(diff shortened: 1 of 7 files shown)


src/olympia/users/tests/test_models.py
```diff
@@ -119,6 +119,7 @@ def test_ban_and_disable_related_content(
         user.ban_and_disable_related_content()
         user.reload()
         assert user.deleted
+        assert user.banned
```
diox: Use `self.assertCloseToNow(user.banned)` (and below too)

eviljeff: https://github.com/mozilla/addons-server/pull/12083/commits/992099fd18f0dab09512ce5faf7c5585ce50de40

eviljeff:
https://github.com/mozilla/addons-server/pull/12083/commits/3638199b9a70bf1157d7e6604ca2be5996c82bc4 adds the backfilling.  Doesn't fail with a syntax error locally, at least.

PR #12085, opened 2019-08-09
Block adding author to an add-on if they don't meet the display name requirement

### LINKED ISSUE #11634: Block adding new author to add-on based on display name

Similar to https://github.com/mozilla/addons-server/issues/11448, we should block adding authors to an add-on if they don't meet the display name requirement.

The error message (in the tooltip) should be "The account needs a display name before it can be added as an author."

diox:
Fixes #11634

### DIFF
--- a/src/olympia/devhub/forms.py
+++ b/src/olympia/devhub/forms.py
@@ -471,12 +471,22 @@ class AuthorWaitingConfirmationForm(AuthorForm):
 
     def clean_user(self):
         user = self.cleaned_data.get('user')
-        if user and not EmailUserRestriction.allow_email(user.email):
-            raise forms.ValidationError(EmailUserRestriction.error_message)
+        if user:
+            if not EmailUserRestriction.allow_email(user.email):
+                raise forms.ValidationError(EmailUserRestriction.error_message)
 
-        if user and self.addon.authors.filter(pk=user.pk).exists():
-            raise forms.ValidationError(
-                ugettext('An author can only be present once.'))
+            if self.addon.authors.filter(pk=user.pk).exists():
+                raise forms.ValidationError(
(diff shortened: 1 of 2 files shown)


src/olympia/devhub/forms.py
```diff
@@ -471,12 +471,22 @@ class Meta(AuthorForm.Meta):
...
+                    ugettext('An author can only be present once.'))
+
+            name_validators = user._meta.get_field('display_name').validators
+            try:
+                for validator in name_validators:
+                    validator(user.display_name)
+            except forms.ValidationError:
+                raise forms.ValidationError(ugettext(
+                    'The account needs a display name before it can be added '
+                    'as an author.'))
```
eviljeff: Mmm, what are those validators?  If they're the ones defined on the model then neither of them are _strictly_ checking if the display name has been set - the first is checking for one or more printable characters, the second that the length is >=2.

diox: Well it works for empty string display names :) Interestingly it doesn't for `None`, the validator raises `TypeError`. I'll add something for that.

diox: (The message is relatively generic but the intent is definitely to block adding people that don't have a /valid/ display name, and that includes people who don't have any)

diox: eb839f999e162d4bb3af66b7b0749f4714e88d68

eviljeff: Should the error be something more along the lines of the account needing a _valid_ display name before it can be added then?

diox: Copy was from product :) Old accounts excepted, it shouldn't be possible for users to set invalid display names, so I guess that message still works.

PR #12113, opened 2019-08-14
Re-introduce fake persona categories, addons-frontend is not ready yet

diox:
Fixes https://github.com/mozilla/addons-frontend/issues/8480

### DIFF
--- a/src/olympia/constants/categories.py
+++ b/src/olympia/constants/categories.py
@@ -408,6 +408,7 @@ CATEGORIES_NO_APP = {
         'general': StaticCategory(name=_(u'General'))
     },
 }
+
 CATEGORIES_NO_APP[_ADDON_PERSONA] = {
     slug: copy.copy(cat)
     for slug, cat in CATEGORIES_NO_APP[ADDON_STATICTHEME].items()}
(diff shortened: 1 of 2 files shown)


src/olympia/constants/categories.py
```diff
@@ -524,6 +525,8 @@ def _immutable(self, *args):
         },
     },
 }
+CATEGORIES[FIREFOX.id][_ADDON_PERSONA] = copy.copy(CATEGORIES[FIREFOX.id][ADDON_STATICTHEME])
+CATEGORIES[ANDROID.id][_ADDON_PERSONA] = copy.copy(CATEGORIES[ANDROID.id][ADDON_STATICTHEME])
```
eviljeff: You shouldn't need the `ANDROID` copy - it wasn't there before - frontend just assumed the same for android as desktop.

src/olympia/constants/categories.py
```diff
@@ -408,6 +408,7 @@ def _immutable(self, *args):
         'general': StaticCategory(name=_(u'General'))
     },
 }
+CATEGORIES_NO_APP[_ADDON_PERSONA] = copy.copy(CATEGORIES_NO_APP[ADDON_STATICTHEME])
```
eviljeff: iirc it needs to be:
```
CATEGORIES_NO_APP[_ADDON_PERSONA] = {
    slug: copy.copy(cat)
    for slug, cat in CATEGORIES_NO_APP[ADDON_STATICTHEME].items()}
``` 
as before because the `StaticCategory` needs to be copied too (`copy.copy` will shallow copy) - it gets modified in the next few lines.

PR #12117, opened 2019-08-14
combine accounts-users-and-ratings job into other jobs

### LINKED ISSUE #12116: drop the seperate accounts-users-and-ratings travis job
the actual tests are quick, it's the setup that takes the time so we can get an overall timing saving by moving the tests into other jobs.

eviljeff:
closes #12116 
I tried combing a few other jobs but there wasn't any particular benefit.  For the mozilla travis jobs they all run at once so the overall time is limited by the slowest job; for our own forks there appears to be a 5 job simultaneous limit, so it's about the combinations of jobs.

base line reference: https://travis-ci.org/eviljeff/olympia/builds/571801374
accounts-users-and-ratings into other jobs: https://travis-ci.org/eviljeff/olympia/builds/571811384

other combinations I tried without success:
https://travis-ci.org/eviljeff/olympia/builds/571914106 (roughly the same time or longer)
https://travis-ci.org/eviljeff/olympia/builds/571926016 (roughly the same time or longer)
https://travis-ci.org/eviljeff/olympia/builds/571820223 (test fails - probably because es_tests don't play well with pytest-xdist)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,9 @@ jobs:
     - { env: TOXENV=docs }
     - { env: TOXENV=assets }
     - { env: TOXENV=es }
-    - { env: TOXENV=addons-versions-and-files }
+    - { env: TOXENV=addons-versions-files-ratings }
     - { env: TOXENV=devhub }
     - { env: TOXENV=reviewers-and-zadmin }
-    - { env: TOXENV=accounts-users-and-ratings }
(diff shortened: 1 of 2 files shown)


diox:
Wait which job has the users and accounts tests now with those changes?

eviljeff:
"main"

PR #12121, opened 2019-08-15
move all the make commands that operate on a single container to Makefile-docker

eviljeff:
I did some tinkering with the `Makefile`s because I was bored with not having the uwgi restart command inside docker  Now with magic to redirect any unknown make recipe to the make inside the web container.


EnTeQuAk:
This seems to break circleci:


```bash
npm install --prefix /deps/
npm WARN deprecated circular-json@0.3.3: CircularJSON is in maintenance only, flatted is its successor.
npm WARN deprecated nomnom@1.8.1: Package no longer supported. Contact support@npmjs.com for more info.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@2.0.7 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@2.0.7: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

added 10 packages from 7 contributors, removed 11 packages, updated 71 packages, moved 1 package and audited 1115 packages in 18.206s
found 0 vulnerabilities

for dest in @claviska/jquery-minicolors/jquery.minicolors.css @claviska/jquery-minicolors/jquery.minicolors.png  ; do cp /deps/node_modules/$dest static/css/node_lib/ ; done
for dest in less/dist/less.js jqmodal/jqModal.js jquery/dist/jquery.js jquery.browser/dist/jquery.browser.js jquery.cookie/jquery.cookie.js @claviska/jquery-minicolors/jquery.minicolors.js jquery-pjax/jquery.pjax.js jszip/dist/jszip.js timeago/jquery.timeago.js underscore/underscore.js  ; do cp /deps/node_modules/$dest static/js/node_lib/ ; done
for dest in jquery-ui/ui/version.js jquery-ui/ui/data.js jquery-ui/ui/disable-selection.js jquery-ui/ui/ie.js jquery-ui/ui/keycode.js jquery-ui/ui/escape-selector.js jquery-ui/ui/labels.js jquery-ui/ui/jquery-1-7.js jquery-ui/ui/plugin.js jquery-ui/ui/safe-active-element.js jquery-ui/ui/safe-blur.js jquery-ui/ui/scroll-parent.js jquery-ui/ui/focusable.js jquery-ui/ui/tabbable.js jquery-ui/ui/unique-id.js jquery-ui/ui/position.js jquery-ui/ui/widget.js jquery-ui/ui/widgets/menu.js jquery-ui/ui/widgets/mouse.js jquery-ui/ui/widgets/autocomplete.js jquery-ui/ui/widgets/datepicker.js jquery-ui/ui/widgets/sortable.js ; do cp /deps/node_modules/$dest static/js/node_lib/ui/ ; done
# If changing this here, make sure to adapt tests in amo/test_commands.py
manage.py compress_assets
make: manage.py: Command not found
Makefile-docker:157: recipe for target 'update_assets' failed
make: *** [update_assets] Error 127
Exited with code 2
```

Makefile-docker
```diff
@@ -1,5 +1,7 @@
 export PIP_COMMAND=pip3
-export PYTHON_COMMAND=python3
+export PYTHON_COMMAND=
```
eviljeff: eek, I didn't mean to change this line.  It's probably what caused the circleci fail

Makefile-os
```diff
@@ -52,35 +17,13 @@ update_docker:
...
-.PHONY: dbshell
-dbshell:
-	docker-compose exec web ./manage.py dbshell
-
-.PHONY: reload-uwsgi
-reload-uwsgi:
-	docker-compose exec web uwsgi --reload /code/docker/artifacts/addons-server-uwsgi-master.pid
+%:
+	# This directs any other recipe (command) to the web container's make.
+	docker-compose exec web make $(MAKECMDGOALS) ARGS=$(ARGS)
```
muffinresearch: For permissions, I wonder if this should be specified as the `olympia` user see also https://github.com/mozilla/addons-server/pull/12153

eviljeff: We could.  I've never found it necessary locally when running the commands but I don't suppose it hurts.

eviljeff: Are there any downsides to this?  I.e. sometimes we might not _want_ commands executed as `olympia`?

According to the docker docs we can specify a default user with `USER olympia` in the Dockerfile so perhaps that's an alternative so we don't need to specify on each docker-compose command.

muffinresearch: I think the main thing to think about is that if it's not `olympia` it'll be `root` for the case where the make file is running on the container, which seems to have been the source of perms issues. That might not make sense for every command but then lower privileges probably won't hurt, but we'll need to test it out.

eviljeff: I added `USER olympia` to the Dockerfile instead.  https://github.com/mozilla/addons-server/pull/12121/commits/208fcb712ab1d45d06bb8f66256ce8b7c927e22b

PR #12135, opened 2019-08-16
change primary hero gradients to be photon color names instead of hex colors

### LINKED ISSUE #12132: Send the photon color names for gradient for the Primary Hero shelf
In order to avoid having to use inline styles in the front-end for `linear-gradient`, and also to support the use of the `:hover` pseudo-selector for the cta button in the Primary Hero shelf, it would be better if the hero API could provide information that can be used to determine a `className` defined in the front-end.

We discussed that if the hero API returns the names of photon colors as the values for `start` and `end` for `gradient`, then the front-end can concatenate those into a `className` which would include those colors in its definition. This would allow for some synchronization of the colors between  server and front-end without having to manually keep them in sync.

The bottom line of the change is this:

Instead of sending hex colors for  `start` and `end` for `gradient`, send the corresponding photon color names.

eviljeff:
fixes #12132

### DIFF
--- a/docs/topics/api/hero.rst
+++ b/docs/topics/api/hero.rst
@@ -41,8 +41,8 @@ small number of shelves this endpoint is not paginated.
     :query string lang: Activate translations in the specific language for that query. (See :ref:`translated fields <api-overview-translations>`)
     :>json array results: The array containing the results for this query.
     :>json object results[].gradient: The background colors used for the gradient.
-    :>json string results[].gradient.start: The starting color name for gradient - typically top or left. The names are the photon color variables from https://github.com/FirefoxUX/photon-colors/blob/master/photon-colors.scss
-    :>json string results[].gradient.end: The ending color name for gradient - typically bottom or right. The names are the photon color variables from https://github.com/FirefoxUX/photon-colors/blob/master/photon-colors.scss
+    :>json string results[].gradient.start: The starting color name for gradient - typically top or left. The name is from the `photon color variables <https://github.com/FirefoxUX/photon-colors/blob/master/photon-colors.scss>`_.
+    :>json string results[].gradient.end: The ending color name for gradient - typically bottom or right. The name is from the `photon color variables <https://github.com/FirefoxUX/photon-colors/blob/master/photon-colors.scss>`_.
     :>json string results[].featured_image: The image used to illustrate the item.
     :>json string|null results[].description: The description for this item, if any. May contain some HTML tags.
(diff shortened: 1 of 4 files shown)


docs/topics/api/hero.rst
```diff
@@ -41,8 +41,8 @@ small number of shelves this endpoint is not paginated.
     :query string lang: Activate translations in the specific language for that query. (See :ref:`translated fields <api-overview-translations>`)
     :>json array results: The array containing the results for this query.
     :>json object results[].gradient: The background colors used for the gradient.
-    :>json string results[].gradient.start: The starting color for gradient - typically top or left.
-    :>json string results[].gradient.end: The ending color for gradient - typically bottom or right.
+    :>json string results[].gradient.start: The starting color name for gradient - typically top or left. The names are the photon color variables from https://github.com/FirefoxUX/photon-colors/blob/master/photon-colors.scss
```
bobsilverberg: Nit: Should be "the name is from", not "the names are", as this is listed once for each name.

docs/topics/api/hero.rst
```diff
@@ -41,8 +41,8 @@ small number of shelves this endpoint is not paginated.
...
+    :>json string results[].gradient.end: The ending color name for gradient - typically bottom or right. The names are the photon color variables from https://github.com/FirefoxUX/photon-colors/blob/master/photon-colors.scss
```
EnTeQuAk: Given its restructuredtext you can use a link and a description instead too

docs/topics/api/hero.rst
```diff
@@ -41,8 +41,8 @@ small number of shelves this endpoint is not paginated.
     :query string lang: Activate translations in the specific language for that query. (See :ref:`translated fields <api-overview-translations>`)
     :>json array results: The array containing the results for this query.
     :>json object results[].gradient: The background colors used for the gradient.
-    :>json string results[].gradient.start: The starting color for gradient - typically top or left.
-    :>json string results[].gradient.end: The ending color for gradient - typically bottom or right.
+    :>json string results[].gradient.start: The starting color name for gradient - typically top or left. The name is from the `photon color variables <https://github.com/FirefoxUX/photon-colors/blob/master/photon-colors.scss>`_.
+    :>json string results[].gradient.end: The ending color name for gradient - typically bottom or right. The name is from the `photon color variables <https://github.com/FirefoxUX/photon-colors/blob/master/photon-colors.scss>`_.
```
eviljeff: There's something weird happening with the RST -> html rendering meaning there's an extra `<p>` inserted in the middle of the gradient.end line after the `:`.  If either one of the links is broken/removed then there isn't an extra `<p>` - it only occurs when both of them are in the same list of parameters/response.

Screenshot 2019-08-19 18 00 17

eviljeff:
I'm going to merge this despite the docs formatting weirdness but it'd be good to work out how to fix it anyway.

PR #1052, opened 2015-12-04
Give a 503 when in read-only for signing (see #1050)

mstriemer:
This just gives a 503 on the upload version API since I think that's the only one we care about now and I wanted this to be small.

This is related to #1050.

PR #12141, opened 2019-08-19
Revert "Update kombu to 4.6.4"

willdurand:
Reverts mozilla/addons-server#12124

This should fix the `build` build and re-enable deployments. There are new issues with this latest release in https://github.com/celery/kombu/issues, and the stack trace in Circle-CI mentions `kombu`: https://circleci.com/gh/mozilla/addons-server/35915.

From one of the GitHub issues:

> It had been mentioned that kombu 4.6 was breaking celery workers pings: #1052
The guilty code was reverted, then, in 4.6.4, the revert has been reverted. 052f760

In addition, all `build` builds fail after this commit (there are green statuses in the list of commits but they don't run the `build` build).

### DIFF
--- a/requirements/prod.txt
+++ b/requirements/prod.txt
@@ -246,9 +246,9 @@ jmespath==0.9.4 \
     --hash=sha256:3720a4b1bd659dd2eecad0666459b9788813e032b83e7ba58578e48254e0a0e6 \
     --hash=sha256:bde2aef6f44302dfb30320115b17d030798de8c4110e28d5cf6cf91a7a31074c
 # kombu is required by celery
-kombu==4.6.4 \
-    --hash=sha256:55274dc75eb3c3994538b0973a0fadddb236b698a4bc135b8aa4981e0a710b8f \
-    --hash=sha256:e5f0312dfb9011bebbf528ccaf118a6c2b5c3b8244451f08381fb23e7715809b
+kombu==4.6.3 \
+    --hash=sha256:55b71d3785def3470a16217fe0780f9e6f95e61bf9ad39ef8dce0177224eab77 \
+    --hash=sha256:eb365ea795cd7e629ba2f1f398e0c3ba354b91ef4de225ffdf6ab45fdfc7d581
 # lxml is required by pyquery
 lxml==4.4.1 \
     --hash=sha256:02ca7bf899da57084041bb0f6095333e4d239948ad3169443f454add9f4e9cb4 \


EnTeQuAk:
The issue in circleci though is related to a dependency missing. `importlib-metadata` is present in `tests.txt` and `travis_base.txt` though. I wonder if it should actually be in `prod.txt` though if kombu is using it in regular code :thinking:

willdurand:
I don't know.. I think we should revert the patch that introduced the issue and then work on updating `kombu` correctly in a follow-up patch.

diox:
It should definitely be in `prod.txt` now with the latest version of kombu: https://github.com/celery/kombu/commit/193054d24331a10d76a50a0501d396a9dbc1a0a6

I submitted https://github.com/mozilla/addons-server/pull/12142

diox:
I merged https://github.com/mozilla/addons-server/pull/12142 (tests passed, including circleci), if this fixes dev this PR should no longer be necessary.

willdurand:
> I merged #12142 (tests passed, including circleci), if that fixes dev this PR should no longer be necessary.

Thanks. I hope this was only that (the missing dep) and not a bigger change. The recent issues in the kombu repo are a bit worried though... (not to mention that 4.6 is not the stable version).

diox:
Interesting, I didn't know 4.6 wasn't stable... Maybe we should pin it down. Thoughts, @EnTeQuAk ?

EnTeQuAk:
> Interesting, I didn't know 4.6 wasn't stable... Maybe we should pin it down. Thoughts, @EnTeQuAk ?

Where is it stated that kombu 4.6 isn't a stable version? Celery 4.1.x defines `kombu>=4.2.0,<5.0` as the supported kombu range while celery 4.2.x pins this down to 
`kombu>=4.2.0,<4.4` and celery 4.3.x goes back to allowing everything `<5.0` so upstream is a bit confusing here :thinking:

willdurand:
> Where is it stated that kombu 4.6 isn't a stable version?

Kombu documentation: http://docs.celeryproject.org/projects/kombu/

> This document is for Kombu's development version, which can be significantly different from previous releases. Get the stable docs here: 4.5.

willdurand:
Could we agree on merging this revert patch to fix the -dev deployments and file an issue to update Kombu correctly in a follow-up? The `build` build is still failing but for something else that time: https://circleci.com/gh/mozilla/addons-server/35923

diox:
Yeah. The weird thing is circle ci build passed for my PR..

Issue #11342, opened 2019-05-06
Apply rate-limiting from the signing API to devhub


diox:
Rate-limiting added in #11287 should also affect devhub.

AlexandraMoga:
@diox I've tried again to trigger this rate-limiting error in dev hub, but I'm still not running into it.

Here is what I did:
1. from the same account (ID 10641840) and the same IP
2. within the same hour (start time May 7, 2019, 7:35 a.m - end time May 7, 2019, 8:33 a.m)
3. submitted 53 unlisted add-ons in devhub and haven't run into an error

Here is the activity log for these actions - https://addons-internal.dev.mozaws.net/en-US/admin/models/activity/activitylog/?created__day=7&created__month=5&created__year=2019&p=0&user=10641840

Maybe I'm misinterpreting the formula -  _3/minute and 20/hour per user, 6/minutes and 50/hour per ip_ - 
Considering that the submissions were made by the same user from the same IP I was expecting to be blocked at some point.

AlexandraMoga:
I've given it another try on stage, after seeing that the API part - #11287 - was giving promising results.

Attempting unlisted submissions in devhub:

Rule/minute
- same user, same IP  - 3 successful submissions, 4th submission blocked
- Same user, different IPs -  3 successful submissions, 4th submission blocked
- Different users (I've used 4), same IP - 6 successful submissions, 7th submission blocked

Rule/hour
 - same user, same IP - 20 successful submissions, 21st submission blocked

@diox the rate-limiting rules seem to be working in devhub.  There could be, of course, more scenarios run to cover all the variables, but I believe that doing this manually from devhub is too costly.

PR #12144, opened 2019-08-19
process_addons task to hard (or soft) delete obsolete addon types

eviljeff:
needed for #11342 - the issue will be closed when the commands have been run and the obsolete addons are gone.

### DIFF
--- a/src/olympia/addons/management/commands/process_addons.py
+++ b/src/olympia/addons/management/commands/process_addons.py
@@ -18,6 +18,7 @@ from olympia.addons.tasks import (
     repack_themes_for_69,
 )
 from olympia.abuse.models import AbuseReport
+from olympia.constants.base import _ADDON_PERSONA, _ADDON_WEBAPP
 from olympia.amo.utils import chunked
 from olympia.devhub.tasks import get_preview_sizes, recreate_previews
 from olympia.lib.crypto.tasks import sign_addons
@@ -153,8 +154,8 @@ tasks = {
             Q(type__in=(amo.ADDON_THEME,
                         amo.ADDON_LPADDON,
                         amo.ADDON_PLUGIN,
-                        9,  # ADDON_PERSONA,
-                        11,  # ADDON_WEBAPP,
+                        _ADDON_PERSONA,
+                        _ADDON_WEBAPP,
                         ))
         ],
         'allowed_kwargs': ('with_deleted',),
--- a/src/olympia/addons/management/commands/process_addons.py
+++ b/src/olympia/addons/management/commands/process_addons.py
@@ -10,6 +10,7 @@ from olympia.addons.models import Addon
 from olympia.addons.tasks import (
     add_dynamic_theme_tag,
     content_approve_migrated_themes,
+    delete_addons,
     extract_colors_from_static_themes,
     find_inconsistencies_between_es_and_db,
     migrate_webextensions_to_git_storage,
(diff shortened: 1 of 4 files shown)


src/olympia/addons/management/commands/process_addons.py
```diff
@@ -145,7 +146,19 @@ def get_recalc_needed_filters():
...
                  addonapprovalscounter__last_content_review__isnull=True)]
-    }
+    },
+    'delete_obsolete_addons': {
+        'method': delete_addons,
+        'qs': [
+            Q(type__in=(amo.ADDON_THEME,
+                        amo.ADDON_LPADDON,
+                        amo.ADDON_PLUGIN,
+                        9,  # ADDON_PERSONA,
```
diox: nit: Personally I'd use `_ADDON_PERSONA` here and even add `_ADDON_WEBAPP` to our base constants and use that as well.

src/olympia/addons/tasks.py
```diff
@@ -330,12 +330,16 @@ def recreate_theme_previews(addon_ids, **kw):
...
-    addons = Addon.objects.filter(pk__in=addon_ids).no_transforms()
-    for addon in addons:
-        addon.delete(send_delete_email=False)
+def delete_addons(addon_ids, with_deleted=False, **kw):
+    log.info('[%s@%s] %sDeleting addons starting at id: %s...'
+             % ('Hard ' if with_deleted else '', len(addon_ids),
+                delete_addons.rate_limit, addon_ids[0]))
+    addons = Addon.unfiltered.filter(pk__in=addon_ids).no_transforms()
+    if with_deleted:
+        addons.delete()
```
diox: Add a comment making what's going on a little clearer (explain it's a hard-delete because we're using the queryset method)

src/olympia/addons/management/commands/process_addons.py
```diff
@@ -145,7 +146,19 @@ def get_recalc_needed_filters():
...
+                        11,  # ADDON_WEBAPP,
+                        ))
+        ],
+        'allowed_kwargs': ('with_deleted',),
```
diox: IMHO the behavior of the task would be clearer if the argument was called `hard_delete` and not `with_deleted`. `with_deleted` makes sense in `process_addons` because it's used to fetch stuff, but it makes less sense in the task where we're using `unfiltered` anyway. So instead of passing it with `allowed_kwargs`, rename it and pass it through `kwargs` ?

src/olympia/addons/management/commands/process_addons.py
eviljeff: yeah I was just being lazy because `_`'d properties aren't exposed in `amo.`

src/olympia/addons/management/commands/process_addons.py
eviljeff: I wrote it that way originally.  But I couldn't see a use-case where you wouldn't also need `with_deleted` when using `hard_delete`, because you otherwise couldn't hard-delete anything previously soft-deleted otherwise, and for anything not soft-deleted you'd want to soft-delete first to attempt file system clean-up, etc.  
So I removed `hard_delete` and just relied on `with_deleted` - I can add it back though.

src/olympia/addons/management/commands/process_addons.py
eviljeff: yeah I was just being lazy because `_`'d properties aren't exposed in `amo.`

src/olympia/addons/management/commands/process_addons.py
diox: I see. It's fine, it's just a little confusing because the name doesn't really suit the task. I guess just adding a comment in the task will be enough.

Issue #12146, opened 2019-08-19
Synchronize add-on name changes with Basket


diox:
Follow-up for https://github.com/mozilla/addons-server/issues/12104 : also synchronize Add-on name changes.

This requires some refactoring to be able to detect those changes properly (because they are not a simple database field - the value in the `addons` table actually never changes)

diox:
I think to do this we'll have to refactor the way translations being held in `olympia.translations.hold._to_save.translations` are processed, adding the name of the field & parent model&pk. The alternative is messing with devhub forms but it's not trivial to detect a change there because `changed_data` won't work (like with using `on_change`, there is no actual field value change here, since we store a fk to the `translations` table and that stays the same)

PR #12159, opened 2019-08-20
Synchronize specific UserProfile, Addon and AddonUser changes with Basket

### LINKED ISSUE #12104: Synchronize some user/add-on data with Salesforce through Basket API
We want to automatically synchronize designated user/add-on fields between Salesforce and AMO. 

[PRD](https://docs.google.com/document/d/1PPy8H_foBrgY-Ls3rQANham65N54-SkzDfPrXmvNq1s/edit)
[Data to sync](https://docs.google.com/spreadsheets/d/1WMsA9ukyLVlyiNtveVijdsbfdcDMkkhsGZ56NrGjMQQ/edit)
[Basket API](https://github.com/mozmeao/basket/issues/223)

diox:
Fixes #12104

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -1509,6 +1509,34 @@ def watch_disabled(old_attr=None, new_attr=None, instance=None, sender=None,
             file_.hide_disabled_file()
 
 
+@Addon.on_change
+def watch_changes(old_attr=None, new_attr=None, instance=None, sender=None,
+                  **kwargs):
+    if old_attr is None:
+        old_attr = {}
+    if new_attr is None:
+        new_attr = {}
+
+    changes = {
+        x for x in new_attr
+        if not x.startswith('_') and new_attr[x] != old_attr.get(x)
+    }
+    basket_relevant_changes = (
+        # Some changes are not tracked here:
+        # - Any authors changes (separate model)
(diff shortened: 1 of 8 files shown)


src/olympia/addons/models.py
```diff
@@ -1509,6 +1509,34 @@ def watch_disabled(old_attr=None, new_attr=None, instance=None, sender=None,
...
+
+    changes = {
+        x for x in new_attr
+        if not x.startswith('_') and new_attr[x] != old_attr.get(x)
+    }
+    basket_relevant_changes = (
+        # Some changes are not tracked here:
+        # - Any authors changes (separate model)
+        # - Creation/Deletion of unlisted version (separate model)
+        # - Name change (separate model, not implemented yet)
```
eviljeff: Separate model for name change?  Doesn't changing `name` update `last_updated` on `Addon`?

diox: My idea was to listen to translation changes (hence the remark about it being a separate model). Relying on `last_updated` is tricky, because it's modified by a cron that acts on a queryset, making it harder to catch and sync individual changes without completely changing the way it's set.

diox: (In any case I made a separate issue for that problem: https://github.com/mozilla/addons-server/issues/12146 )

PR #12168, opened 2019-08-21
Fix broken hu locale

### LINKED ISSUE #12167: Broken locale
From: https://travis-ci.org/mozilla/addons-server/jobs/574765362

```
>>> Working on: /home/travis/build/mozilla/addons-server/locale/hu/LC_MESSAGES/django.po

E201: invalid variables: {addon}

1687:msgid "Author invitation for {addon_name}"

1688:msgstr "Szerzői meghívás a(z) {addon} kiegészítőhöz."
```

willdurand:
fixes #12167, hopefully

### DIFF
--- a/locale/hu/LC_MESSAGES/django.po
+++ b/locale/hu/LC_MESSAGES/django.po
@@ -1685,7 +1685,7 @@ msgid "An author has been added to your add-on"
 msgstr "A kiegészítőhöz egy szerző hozzárendelésre került"
 
 msgid "Author invitation for {addon_name}"
-msgstr ""
+msgstr "Szerzői meghívás a(z) {addon_name} kiegészítőhöz."
 
 msgid "A confirmation email has been sent to {email}"
 msgstr "Megerősítő e-mail elküldve ide: {email}"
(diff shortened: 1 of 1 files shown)


locale/hu/LC_MESSAGES/django.po
```diff
@@ -1685,7 +1685,7 @@ msgid "An author has been added to your add-on"
 msgstr "A kiegészítőhöz egy szerző hozzárendelésre került"
 
 msgid "Author invitation for {addon_name}"
-msgstr "Szerzői meghívás a(z) {addon} kiegészítőhöz."
+msgstr ""
```
eviljeff: you can change it to `Szerzői meghívás a(z) {addon_name} kiegészítőhöz.` so it's not broken.

willdurand: Are you sure this is the right translation? I don't speak Hungarian... I feel like letting localizers update this string would be safer but 🤷‍♂

eviljeff: I don't, but I wasn't updating the translation from what was written, it was just correcting the broken `{addon_name}` token.

willdurand: I see, but why is that better than letting localizers update the entire localized string? In the frontend, I think we've seen such issues where the localized string was completely different (not only a problem with the placeholder name but with the whole string)

diox: When there is only a single placeholder, and it looks like an obvious typo in the placeholder name, we tend to fix it ourselves, it's just faster.

PR #12181, opened 2019-08-22
Trigger an add-on sync with basket when an add-on name changes

### LINKED ISSUE #12146: Synchronize add-on name changes with Basket

diox:
Fixes #12146

### DIFF
--- a/src/olympia/translations/hold.py
+++ b/src/olympia/translations/hold.py
@@ -8,9 +8,10 @@ from django.core.signals import request_finished
 _to_save = local()
 
 
-def add_translation(instance, translation, field=None):
+def add_translation(*, instance, translation, field):
     """
-    Queue a translation that needs to be saved for a particular instance.
+    Queue a `translation` that needs to be saved for a particular `field` on
+    `instance`.
     """
     if not hasattr(_to_save, 'translations'):
         _to_save.translations = {}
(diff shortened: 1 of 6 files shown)


src/olympia/translations/hold.py
```diff
@@ -1,21 +1,22 @@
...
-    generate the key, call make_key.
+    Queue a translation that needs to be saved for a particular instance.
     """
     if not hasattr(_to_save, 'translations'):
         _to_save.translations = {}
-
+    key = make_key(instance)
     _to_save.translations.setdefault(key, [])
-    _to_save.translations[key].append(translation)
+    _to_save.translations[key].append((field.name, translation))
```
eviljeff: If `field` is `None` (the kw default) this will fail.

diox: Yeah... I thought about checking if `field is None` and throw an exception, but that doesn't buy us much compared to the `AttributeError` that would be raised if someone does call that function without a `field` set correctly... And silently ignoring it doesn't seem right either.

diox: https://github.com/mozilla/addons-server/pull/12181/commits/5d24fcfd4889869cd7db05d63a4cc1a0ee03ac53 Removes the default while keeping it a keyword argument.

PR #12190, opened 2019-08-23
Add AbuseReport.addon_install_source ; relax addon_install_method

### LINKED ISSUE #12060: Allow arbitrary values for `addon_install_method` in the Abuse Reports API
The Abuse Reports API currently allows a limited set of values for `addon_install_method`, and doesn't accept reports using other values. On the Firefox side this was solved by defaulting reports to the "Other" method if the value didn't fit the predefined set. Some changes happened in those values since first established that have caused lots of reports to be classified under "Other".

To address this problem, we should do the following:
* Remove the value restriction on the API and allow arbitrary values.
* On the AMO side, map unknown values to "Other", but log those values so they can be looked at and new install methods are added when necessary.
* It might also be a good idea to do the `-` -> `_` mapping that [Firefox currently does](https://searchfox.org/mozilla-central/source/toolkit/mozapps/extensions/AbuseReporter.jsm#152-185).
* Add the new constants that are missing, in preparation for Firefox sending them. @rpl should know what those are.

diox:
`AbuseReport.addon_install_source` is added to match what Firefox has internally, and to help support future versions that may introduce new possible values, the `AbuseReport` serializer now
accepts unknown values for `addon_install_method` and `addon_install_source` and turns them into "other" instead of returning a 400.

Fixes #12060

### DIFF
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -64,6 +64,9 @@ class AbuseReport(ModelBase):
         ('OTHER', 127, 'Other'),
     )
 
+    # https://searchfox.org
+    # /mozilla-central/source/toolkit/components/telemetry/Events.yaml#122-131
+    # Firefox submits values in lowercase, with '-' and ':' changed to '_'.
     ADDON_INSTALL_METHODS = APIChoicesWithNone(
         ('AMWEBAPI', 1, 'Add-on Manager Web API'),
         ('LINK', 2, 'Direct link'),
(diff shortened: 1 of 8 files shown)


src/olympia/abuse/models.py
```diff
@@ -72,12 +75,42 @@ class AbuseReport(ModelBase):
...
+        # install source and method into addon_install_method before deciding
+        # to split the two like Firefox does, so these 6 values are only kept
+        # for backwards-compatibility with older reports and older versions of
+        # Firefox that still only submit that.
         ('FILE_URL', 8, 'File URL'),
         ('ENTERPRISE_POLICY', 9, 'Enterprise Policy'),
         ('DISTRIBUTION', 10, 'Included in build'),
         ('SYSTEM_ADDON', 11, 'System Add-on'),
         ('TEMPORARY_ADDON', 12, 'Temporary Add-on'),
         ('SYNC', 13, 'Sync'),
```
eviljeff: Should these description strings be labelled as obsolete too? I.e. in the string itself.

diox: Yeah that's a good point, I'll do that.

diox: Actually, I think it's better to do that once Firefox 70 is released (with the patch from https://bugzilla.mozilla.org/show_bug.cgi?id=1572173) so that it's less confusing. I'll file an issue for us to update the strings later.

PR #12166, opened 2019-08-21
Add task to run customs

willdurand:
TODO:

- [ ] write tests
- [ ] add waffle switch

PR #12191, opened 2019-08-23
Add task to run `customs` (over HTTP)

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12208

---

~~This is basically #12166 but using HTTP/RPC to run the `customs` scanner.~~

This patch adds a task to run scanners _via_ RPC. It adds the actual code to run `customs` behind a waffle switch. It is synchronous for the moment, because we need to update Celery first. I believe we can safely try this patch in -dev for a while though.

## Try it!

1. clone `customs` locally in `addons-server/private/addons-customs-scanner` and checkout the `functions` branch
2. checkout this patch locally
3. build new docker services:

    ```
    docker-compose -f docker-compose.yml -f docker-compose.private.yml build
    docker-compose -f docker-compose.yml -f docker-compose.private.yml  up -d
    ```

4. run `make update_docker`
5. enable the waffle switch named `enable-customs`:

    ```
    ./manage.py waffle_switch enable-customs on
    ```

6. upload a new xpi and you should have a new entry in the `scanners_results` table shortly after

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,7 @@ tmp/*
 # private projects
 addons-yara
 private/
+
+# do not ignore the following files
+!docker-compose.private.yml
+!private/README.md
--- a/.gitignore
+++ b/.gitignore
@@ -48,4 +48,4 @@ supervisord.pid
 tmp/*
 
 # private projects
-addons-yara
+private/
--- a/.gitignore
+++ b/.gitignore
@@ -48,4 +48,5 @@ supervisord.pid
 tmp/*
 
 # private projects
+addons-yara
 private/
--- a/settings.py
+++ b/settings.py
@@ -137,7 +137,7 @@ INBOUND_EMAIL_SECRET_KEY = 'totally-unsecure-secret-string'
 INBOUND_EMAIL_VALIDATION_KEY = 'totally-unsecure-validation-string'
 
 # Sectools
-CUSTOMS_API_URL = 'http://host.docker.internal:10101/'
+CUSTOMS_API_URL = 'http://customs:10101/'
 CUSTOMS_API_KEY = 'customssecret'
 
 # If you have settings you want to overload, put them in a local_settings.py.
--- a/settings.py
+++ b/settings.py
@@ -136,6 +136,10 @@ INBOUND_EMAIL_SECRET_KEY = 'totally-unsecure-secret-string'
 # Validation key we need to send in POST response.
 INBOUND_EMAIL_VALIDATION_KEY = 'totally-unsecure-validation-string'
 
+# Sectools
+CUSTOMS_API_URL = 'http://host.docker.internal:10101/'
+CUSTOMS_API_KEY = 'customssecret'
+
 # If you have settings you want to overload, put them in a local_settings.py.
 try:
     from local_settings import *  # noqa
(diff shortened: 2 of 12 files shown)


willdurand:
@diox external contributors will never have access to these scanners, so how should this be handled? (@eviljeff mentioned this problem in IRC the other day). I think adding a waffle switch would work because we do want to make sure those tools are running in the hosted envs (so adding some sort of checks would be tricky) and it'll be easy for our team to enable them locally too. WDYT?

willdurand:
Other question: if I add more tests and the waffle switch to this patch, would it be OK to land it?

I deployed the lambda functions on GCP (using my dev account), so we could probably configure the env vars, turn the switch ON and see how it behaves in -dev. WDYT?

diox:
A waffle switch works for me. Alternatively, we could make the whole thing depend on the presence of the env variables even locally, but I guess it would be more annoying for us to work with...

Haven't looked at the patch in detail yet but the idea seems sound.

settings.py
```diff
@@ -136,6 +136,10 @@
 # Validation key we need to send in POST response.
 INBOUND_EMAIL_VALIDATION_KEY = 'totally-unsecure-validation-string'
 
+# Sectools
+CUSTOMS_API_URL = 'http://host.docker.internal:10101/'
```
diox: I assume this would be https for dev/stage/prod, right ?

src/olympia/scanners/tasks.py
```diff
@@ -0,0 +1,80 @@
...
+                             'not exist.' .format(upload.path))
+
+        result = ScannersResult()
+        result.upload = upload
+        result.scanner = scanner
+
+        with statsd.timer('devhub.{}'.format(scanner_name)):
+            headers = {'Authorization': 'Bearer {}'.format(api_key)}
+            with open(upload.path, 'rb') as xpi:
+                response = requests.post(url=api_url,
```
diox: I think we'll want to pass a `timeout` here. Technically it will matter less once this is run as a proper task, because we can just rely on the task execution time limits, but still a good thing to have, with a setting to control the value.

settings.py
diox: As mentioned on IRC this doesn't work for me with Linux as the host. We need a more robust solution.

willdurand: > I assume this would be https for dev/stage/prod, right ?

Yes

> As mentioned on IRC this doesn't work for me with Linux as the host. We need a more robust solution.

😢 working on it

src/olympia/scanners/tasks.py
willdurand: Good idea, thanks

.gitignore
```diff
@@ -48,4 +48,4 @@ supervisord.pid
 tmp/*
 
 # private projects
-addons-yara
+private/
```
diox: It's fine, but it will break yara locally while https://github.com/mozilla/addons-server/pull/12059 is not merged, so maybe only add `private/` in this PR, without removing `addons-yara` ? No big deal though, we can handle that in #12059 shortly.

willdurand: ah yes, good catch, thanks

willdurand:
I fixed the `.gitignore` and added a `README` in the `private/` folder to add the instructions from this PR description.

Issue #10917, opened 2019-03-12
django database definition should match our prod db schema


eviljeff:
part of #3705 

(from https://github.com/mozilla/addons-server/issues/3705#issuecomment-471957170):
> make sure we are able to generate a schema locally that looks like -dev, or go the other way around and implement a migration that migrates -dev, -stage and -prod to something like -local.

eviljeff:
A took a look at this but it's not a quick task.  The differences fall largely into these categories that we'll need to handle:
* different datatypes (e.g. django uses `double`; our db specifies `float`)
* indexes with different names
* db defaults (django doesn't auto generate them; we have them for most columns)

And some irrelevant differences like `int(10)` vs `int(11)` which only affects formatting.

We'll have to change our db schema for datatype differences - there's no other way (short of subclassing the Field classes and adding to our custom mysql db layer).  Index differences could be changed in either but each name change involves dropping the existing constraint and index and adding a new constraint and key (costly and fragile).  We can maybe use https://github.com/3YOURMIND/django-add-default-value to add database defaults (or drop them from the schema for now)

AlexandraMoga:
Is any QA required for this?

PR #12192, opened 2019-08-23
change -dev to match what django would create.  abuse_reports and addons

eviljeff:
part of #10917

this just touches `addons` and `abuse_reports` tables

It doesn't handle the mismatched fk constraint names yet because there's no way to do it in sql, and it's not clear how it can be done in django either - [`Meta.indexes`](https://docs.djangoproject.com/en/2.2/ref/models/indexes/) and [`Meta.constraints`](https://docs.djangoproject.com/en/2.2/ref/models/constraints/) are for additional table indexes and general constraints respectively, whereas fk constraints are a little different.  Replacing https://github.com/django/django/blob/2.2.4/django/db/backends/base/schema.py#L1001 may be the only way... 😨


eviljeff:
Also, this is a result of a verrryyyy slow manual, line by line, method of looking at the diffs for the end goal of having the databases match when there's +600 line differences in a diff between a local test DB and addons-dev and this only modifies 33 columns.  I may have to reconsider the strategy.

diox:
> It doesn't handle the mismatched fk constraint names yet because there's no way to do it in sql

Can't we drop them and add them back ?

> Also, this is a result of a verrryyyy slow manual, line by line, method of looking at the diffs for the end goal of having the databases match when there's +600 line differences in a diff between a local test DB and addons-dev and this only modifies 33 columns. I may have to reconsider the strategy.

What kind of other differences are there ? Can you share the part of the diff that's relevant for this PR ?

eviljeff:
>> It doesn't handle the mismatched fk constraint names yet because there's no way to do it in sql

> Can't we drop them and add them back ?

Of course, but adding constraints is generally slow on large tables - there are 10 fks in the addons table and addons is quite large.  (Plus the other tables not considered for this patch)

eviljeff:
> What kind of other differences are there ? 

It's the scale of the inconsistencies that's the problem rather than different kinds of difference.  I mentioned in the main kinds in https://github.com/mozilla/addons-server/issues/10917#issuecomment-475235011 - though I've now concluded styling differences (`int(10)`) need to be addressed too because their noise is getting in the way of noticing other things we need to fix.

> Can you share the part of the diff that's relevant for this PR ?

I don't have a unified diff.  These are the sql outputs I was comparing (constraints listed separately at the bottom):
[test_olympia-db_schema.txt](https://github.com/mozilla/addons-server/files/3538382/test_olympia-db_schema.txt)
[addons_dev-db_schema.txt](https://github.com/mozilla/addons-server/files/3538383/addons_dev-db_schema.txt)

src/olympia/migrations/1121-update-db-schema-to-match-django.sql
```diff
@@ -0,0 +1,36 @@
...
+    MODIFY `eula` int(10) unsigned,
+    MODIFY `homepage` int(10) unsigned,
+    MODIFY `icon_hash` varchar(8),
+    MODIFY `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
+    MODIFY `inactive` tinyint(1) NOT NULL,
+    MODIFY `modified` datetime(6) NOT NULL,
+    MODIFY `name` int(10) unsigned,
+    MODIFY `last_updated` datetime(6) NOT NULL,
+    MODIFY `privacypolicy` int(10) unsigned,
+    MODIFY `publicstate` tinyint(1) NOT NULL,
```
diox: Where does that come from ?

diox: Ah, typo, `publicstats` !

eviljeff: d'oh

src/olympia/migrations/1121-update-db-schema-to-match-django.sql
```diff
@@ -0,0 +1,36 @@
...
+    MODIFY `description` int(10) unsigned,
+    MODIFY `developercomments` int(10) unsigned,
+    MODIFY `eula` int(10) unsigned,
+    MODIFY `homepage` int(10) unsigned,
+    MODIFY `icon_hash` varchar(8),
+    MODIFY `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
+    MODIFY `inactive` tinyint(1) NOT NULL,
+    MODIFY `modified` datetime(6) NOT NULL,
+    MODIFY `name` int(10) unsigned,
+    MODIFY `last_updated` datetime(6) NOT NULL,
```
diox: This is the one that fails locally for me, and it would also fail on stage (probably dev & prod as well):
```sql
> SELECT * FROM addons WHERE last_updated IS NULL;
(...)
12250 rows in set (0.07 sec)
```

eviljeff: oh, that's a mistake.  The django field definition is `last_updated = models.DateTimeField(       db_index=True, null=True,...)` so I shouldn't have added `NOT NULL` to it.

src/olympia/migrations/1121-update-db-schema-to-match-django.sql
```diff
@@ -0,0 +1,36 @@
...
+    MODIFY `status` int(10) unsigned NOT NULL,
+    MODIFY `summary` int(10) unsigned,
+    MODIFY `supportemail` int(10) unsigned,
+    MODIFY `supporturl` int(10) unsigned,
+    MODIFY `textreviewscount` int(10) unsigned NOT NULL,
+    MODIFY `totaldownloads` int(10) unsigned NOT NULL,
+    MODIFY `totalreviews` int(10) unsigned NOT NULL,
+    MODIFY `viewsource` tinyint(1) NOT NULL,
+    MODIFY `weeklydownloads` int(10) unsigned,
+    DROP `whiteboard`;
```
diox: Can you add a comment explaining that it should be commented out if it fails locally ?

src/olympia/migrations/1121-update-db-schema-to-match-django.sql
```diff
@@ -0,0 +1,36 @@
+ALTER TABLE `abuse_reports`
+    MODIFY `created` datetime(6) NOT NULL,
+    MODIFY `modified` datetime(6) NOT NULL,
+    MODIFY `guid` varchar(255),
+    MODIFY `state` NOT NULL;
```
diox: Need the column definition for this line to work

junngo:
When I run command(`make update_docker`) locally after this PR, I happened error in my case.
So I run command that is `make initialize_docker` and I solved :)
If you happen error, You try to run initialize command.

Issue #12147, opened 2019-08-19
Add task to synchronize user/add-on data with Basket


diox:
Prerequisite for https://github.com/mozilla/addons-server/issues/12104

This cover just the task, #12104 will take care of calling it in various places. It needs to be behind a waffle switch, and documentation for what we send to what endpoints will need to be included.

PR #12203, opened 2019-08-27
Add log statement when synchronizing an object with basket

diox:
Extra logging for https://github.com/mozilla/addons-server/issues/12147

### DIFF
--- a/src/olympia/users/models.py
+++ b/src/olympia/users/models.py
@@ -962,6 +962,9 @@ def watch_changes(old_attr=None, new_attr=None, instance=None,
     )
     if any(field in changes for field in basket_relevant_changes):
         from olympia.amo.tasks import sync_object_to_basket
+        log.info(
+            'Triggering a sync of %s %s with basket because of %s change',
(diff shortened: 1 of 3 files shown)


src/olympia/users/models.py
```diff
@@ -962,6 +962,9 @@ def watch_changes(old_attr=None, new_attr=None, instance=None,
     )
     if any(field in changes for field in basket_relevant_changes):
         from olympia.amo.tasks import sync_object_to_basket
+        log.info(
+            'Triggering a sync of %s %s with basket because of %s change',
+            'userprofile', instance.pk, 'attribute')
```
eviljeff: Why the unnecessary string substitution in each `log.info`?

diox: Just to make sure the string was consistent without having to write a wrapper for it

eviljeff: you could have specified the string as a constant.

eviljeff:
I didn't r+ before you merged 😛

diox:
woops sorry about that :)

PR #12205, opened 2019-08-27
Don't synchronize deleted add-ons with basket when their authors change

diox:
#12147

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -1634,7 +1634,7 @@ def addon_user_sync(sender=None, instance=None, **kwargs):
     # Basket doesn't care what role authors have or whether they are listed
     # or not, it just needs to be updated whenever an author is added/removed.
     created_or_deleted = 'created' not in kwargs or kwargs.get('created')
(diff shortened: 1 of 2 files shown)


src/olympia/addons/models.py
```diff
@@ -1634,7 +1634,7 @@ def addon_user_sync(sender=None, instance=None, **kwargs):
     # Basket doesn't care what role authors have or whether they are listed
     # or not, it just needs to be updated whenever an author is added/removed.
     created_or_deleted = 'created' not in kwargs or kwargs.get('created')
-    if created_or_deleted:
+    if created_or_deleted and instance.addon.status != amo.STATUS_DELETED:
```
eviljeff: I assume the case when an add-on is deleted (and so the user isn't seen as an author any longer) is captured elsewhere?

diox: Add-on deletion is implemented in the `@Addon.on_change` `watch_changes` function since it's actually a `status` change on `Addon`. It's covered by tests.

PR #12220, opened 2019-08-29
Autogenerate makefile help

willdurand:
This is how it looks:

Screen Shot 2019-08-29 at 14 38 32

### DIFF
--- a/Makefile-docker
+++ b/Makefile-docker
@@ -213,7 +213,7 @@ dbshell: ## connect to a database shell
 initialize: update_deps initialize_db update_assets populate_data ## init the dependencies, the database, and assets
 
 .PHONY: reload-uwsgi
-reload-uwsgi:
+reload-uwsgi: ## force django code reload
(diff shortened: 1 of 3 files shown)


eviljeff:
The "Host only commands" section is broken in the screenshot.

Makefile-docker
```diff
@@ -211,52 +190,57 @@ setup-codestyle:
...
 .PHONY: dbshell
-dbshell:
+dbshell: ## connect to a database shell
 	$(PYTHON_COMMAND) ./manage.py dbshell
 
 .PHONY: initialize
-initialize: update_deps initialize_db update_assets populate_data
+initialize: update_deps initialize_db update_assets populate_data ## init the dependencies, the database, and assets
 
 .PHONY: reload-uwsgi
```
eviljeff: It didn't have one before, but it'd be nice if this had a comment too so it shows up

willdurand:
> The "Host only commands" section is broken in the screenshot.

oops, I did not notice that, sorry.

PR #12222, opened 2019-08-29
Fix abuse report API docs & app_profile value for install source

diox:
Fixes https://github.com/mozilla/addons-server/issues/12060


src/olympia/abuse/models.py
```diff
@@ -96,7 +96,7 @@ class AbuseReport(ModelBase):
         ('ABOUT_DEBUGGING', 2, 'Add-ons Debugging'),
         ('ABOUT_PREFERENCES', 3, 'Preferences'),
         ('AMO', 4, 'AMO'),
-        ('APP:PROFILE', 5, 'app:profile'),
+        ('APP_PROFILE', 5, 'App Profile'),
```
eviljeff: checking this doesn't need a DB migration for any existing records with the value `APP:PROFILE`

diox: We just store the integer value (`5`) so no (and it's a recent introduction anyway, no clients except maybe Firefox Nightly would have sent this parameter, and if they did they would have sent `app_profile` which would not have worked before this patch)

PR #12230, opened 2019-08-30
Require email confirmation before creating the first developer API key

diox:
Fixes #11635

PR #12240, opened 2019-08-30
Add view to download FileUpload

### LINKED ISSUE #12232: Allow FileUpload to be downloaded (from the scanners)
While testing the customs integration in -dev with some large add-ons, I discovered that add-ons larger than 10MB are rejected by the customs cloud function. This is because GCP cloud functions have very strict and fixed limits: https://cloud.google.com/functions/quotas#resource_limits. AWS has a limit of 6MB.

Currently, it does not seem possible to upload add-ons larger than 40-50MB because of the soft time limit (and the fact that repack and linting run within the same task).

That being said, we should be able to upload add-ons with a size of 10 to 30MB without any issue, but then the customs cloud function will not be able to scan these add-ons.

Therefore, we need a different solution to pass the xpi to these scanners. The idea would be to send a download URL instead of the xpi itself and let the cloud function download the xpi.

The download URL should be authenticated. We can generate a token in the `FileUpload` (on creation) and use it, so that a download URL is bound to only one `FileUpload` and we can't retrieve other files with the same token (if the download URL leaks for example).

The cloud function will validate the download URL sent (making sure it is a valid URL with a whitelisted host). An issue will be filed in the customs repo for that.

This issue is about:

1. Generating a token when a `FileUpload` is created (this needs a new column in the database table)
2. Creating a method on `FileUpload` to return its very own authenticated download URL
3. Adding a new view/endpoint to be able to download `FileUpload`

Did I miss anything? Any suggestions?

willdurand:
Fixes #12232 

---

Do we want to document this endpoint?


src/olympia/files/views.py
```diff
@@ -204,3 +206,20 @@ def serve(request, viewer, key):
...
+
+def serve_file_upload(request, uuid):
+    upload = shortcuts.get_object_or_404(FileUpload, uuid=uuid)
+    access_token = request.GET.get('access_token')
+
+    if not access_token:
+        log.error('Denying access to %s, no token.' % upload.id)
+        raise PermissionDenied
+
+    if access_token != upload.access_token:
```
diox: If we want to be serious about it... Use `constant_time_compare` from `django.utils.crypto`.

willdurand: oh, yeah. That's a good call, I forgot about timing attack on string comparison..

src/olympia/files/tests/test_views.py
```diff
@@ -728,3 +731,31 @@ def test_files_list_uses_correct_links(self):
...
+        assert resp.status_code == 403
+
+    def test_get(self):
+        resp = self.client.get(self.upload.get_authenticated_download_url())
+
+        assert resp.status_code == 200
+        assert resp['content-type'] == 'application/octet-stream'
+        # This view returns an `HttpResponseSendFile`, which defines this
+        # attribute.
+        assert resp.path == self.upload.path
```
willdurand: Initially, I wanted to assert on the content but I can't get it to work...

src/olympia/files/views.py
diox: To be fair... to exploit this, you'd need to guess the uuid of your target (good luck!), find the token using some kind of timing attack, and do all that for the very limited amount of time the FileUpload still have an xpi attached... all this just to download an xpi :)

src/olympia/files/tests/test_views.py
diox: `HttpResponseSendFile` shouldn't return content, it relegates the job to nginx. What you need though is `@override_settings(XSENDFILE=True)`, and then assert on `response[settings.XSENDFILE_HEADER]` being equals to the path you're looking for.

willdurand: woah, ok. TIL

src/olympia/migrations/1126-add-access_token-to-file_upload.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `file_uploads` ADD COLUMN `access_token` varchar(40) DEFAULT NULL;
```
diox: Migration number is obsolete (sorry)

diox:
> Do we want to document this endpoint?

Not as part of the API docs at least. You can add a docstring to your function explaining why it's there though.

src/olympia/files/models.py
```diff
@@ -566,6 +573,16 @@ def add_file(self, chunks, filename, size):
         self.hash = 'sha256:%s' % hash_func.hexdigest()
         self.save()
 
+    def generate_access_token(self):
+        return binascii.hexlify(os.urandom(20)).decode()
```
EnTeQuAk: We now have pretty similar code like that in api.models for ApiToken, accounts.utils regarding fxa and others, I wonder if we could move that out. On the other hand, we won't gain much so it's rather a nit rather than a requirement.

diox: This would be a good follow-up after https://github.com/mozilla/addons-server/pull/12230 is also merged. I thought about it and then discarded the idea because the use cases were a little different, and ultimately it's just a one-liner, but maybe it's worth consolidating.

EnTeQuAk: Yeah, in particular that (not complex) one-linter part is what makes it only a nit. I wrote the comment down, wanted to deleted at first but then thought… meh… could be done, not a requirement though, so :man_shrugging:

Issue #5712, opened 2017-06-21
Refactor indices for our statistics models


EnTeQuAk:
This issue is a meta-issue to document what needs to be done to refactor our database indices on some of our stats models.

Current issue: a full reindex on -dev doesn't work because theme_user_counts is queried by date which is not indexed.

Our solution: We'll need a (reversed) composite key covering `(date, addon_id)` which should match perfectly how we query that model most of the time and allows mysql to query that exact index for `date` only searches.

Now, we do that for a few other models. While looking through our indices on our stats models I saw some quite gruelsome stuff, so let's start cleaning that up a bit.

`stats_collections_counts`: 

* [ ] has an index on `collection_id` (`collection_id`), let's drop that. It's a relict from MySQL < 5.x, these days creates that index automatically
* [ ] has an index on `date`, let's drop that
* [ ] has an index on `count`, let's drop that
* [ ] add index on (date, addon_id)

`download_counts`:
* [ ] drop index on `addon_id`, not needed anymore, mysql creates these automatically
* [ ] drop index on `count`
* [ ] rename `date_2` unique composite index on (date, addon_id) to something django would generate. MySQL supports proper index renames these days (`ALTER TABLE RENAME INDEX`, new in 5.7)
* [ ] Remoe usage of `USE KEY (`addon_and_count`)` in addons/cron.py

`update_counts` (https://github.com/mozilla/addons-server/issues/8475):

* [ ] Drop `addon_id` index, not needed anymore
* [ ] Drop `count` index, should not be used (and is relatively useless imho)
* [ ] Drop `addon_and_count` (`addon_id`,`count`)
* [ ] Drop `addon_date_idx` (`addon_id`,`date`)
* [ ] Add (date, addon_id) index

`theme_user_counts` (https://github.com/mozilla/addons-server/pull/5716):

* [x] Drop `addon_date_idx` (`addon_id`,`date`)
* [x] Add new (date, addon_id) index

NOTE: This only covers models related to our reindex for now, the others don't look any better but I wanted to sandbox my problem a bit, the rest will need a follow up later.

Refs https://github.com/mozilla/addons-server/issues/3705

EnTeQuAk:
Unassigning myself for the time being, there is a few split-off issues and steps are documented.

PR #12255, opened 2019-09-02
Declare some indexes that already exist on dev/stage/prod in our models

diox:
Part of #10917

This updates ~ half (*) of our models (to match existing schema in dev/stage/prod. The idea is to leverage Django index naming capabilities to avoid migrations as much as possible.

(*) Went through all the tables from `m` to `z`.

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -32,8 +32,8 @@ from olympia.addons.utils import (
 from olympia.amo.decorators import use_primary_db
 from olympia.amo.fields import PositiveAutoField
 from olympia.amo.models import (
-    BasePreview, BaseQuerySet, ManagerBase, ModelBase, OnChangeMixin,
-    SaveUpdateMixin, SlugField, manual_order)
+    BasePreview, BaseQuerySet, LongNameIndex, ManagerBase, ModelBase,
+    OnChangeMixin, SaveUpdateMixin, SlugField, manual_order)
 from olympia.amo.templatetags import jinja_helpers
 from olympia.amo.urlresolvers import reverse
 from olympia.amo.utils import (
@@ -1579,13 +1579,24 @@ class AddonReviewerFlags(ModelBase):
 
 class MigratedLWT(OnChangeMixin, ModelBase):
     lightweight_theme_id = models.PositiveIntegerField()
-    getpersonas_id = models.PositiveIntegerField(db_index=True)
+    getpersonas_id = models.PositiveIntegerField()
     static_theme = models.ForeignKey(
         Addon, unique=True, related_name='migrated_from_lwt',
         on_delete=models.CASCADE)
 
     class Meta:
         db_table = 'migrated_personas'
+        indexes = [
+            LongNameIndex(
+                fields=('lightweight_theme_id',),
+                name='migrated_personas_lightweight_theme_id_fk_addons_id'),
+            LongNameIndex(
+                fields=('static_theme',),
+                name='migrated_personas_static_theme_id_fk_addons_id'),
+            LongNameIndex(
+                fields=('getpersonas_id',),
+                name='migrated_personas_getpersonas_id'),
+        ]
 
 
 class AddonCategory(models.Model):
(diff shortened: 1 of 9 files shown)


src/olympia/stats/models.py
```diff
@@ -19,34 +19,31 @@ class StatsSearchMixin(SearchMixin):
...
-
-        # additional indices on this table (in dev, stage and prod):
-        # * KEY `addon_and_count` (`addon_id`,`count`)
-        # * KEY `addon_date_idx` (`addon_id`,`date`)
-
-        # in our (dev, stage and prod) database:
-        # UNIQUE KEY `date_2` (`date`,`addon_id`)
-        unique_together = ('date', 'addon')
+        indexes = [
+            # FIXME: some of these might redundant.
```
EnTeQuAk: Given that it's been an ongoing investigation, there's https://github.com/mozilla/addons-server/issues/5712 that covers that part. Maybe worth linking to it?

src/olympia/stats/models.py
```diff
@@ -55,7 +52,11 @@ class UpdateCount(StatsSearchMixin, models.Model):
 
     class Meta:
         db_table = 'update_counts'
-
-        # Additional indices on this table (on dev, stage and prod):
-        # * KEY `addon_and_count` (`addon_id`,`count`)
-        # * KEY `addon_date_idx` (`addon_id`,`date`)
+        indexes = [
+            # FIXME: some of these might redundant.
```
EnTeQuAk: Same here, maybe link to https://github.com/mozilla/addons-server/issues/5712?

src/olympia/stats/models.py
diox: 1163b56517f2961ae5f136dc6d937d276fcfc897

src/olympia/stats/models.py
diox: 1163b56517f2961ae5f136dc6d937d276fcfc897

src/olympia/addons/models.py
```diff
@@ -1579,13 +1579,24 @@ class AddonReviewerFlags(ModelBase):
...
     static_theme = models.ForeignKey(
         Addon, unique=True, related_name='migrated_from_lwt',
         on_delete=models.CASCADE)
 
     class Meta:
         db_table = 'migrated_personas'
+        indexes = [
+            LongNameIndex(
+                fields=('lightweight_theme_id',),
+                name='migrated_personas_lightweight_theme_id_fk_addons_id'),
```
eviljeff: I wonder if we should actually drop this as it's not a foreign key any longer.

src/olympia/tags/models.py
```diff
@@ -28,6 +28,13 @@ class Tag(ModelBase):
     class Meta:
         db_table = 'tags'
         ordering = ('tag_text',)
+        indexes = [
+            models.Index(fields=('denied', 'num_addons'),
+                         name='tag_blacklisted_num_addons_idx')
+        ]
+        constraints = [
+            models.UniqueConstraint(fields=('tag_text',), name='tag_text')
```
eviljeff: should we be declaring `unique=True` in the column definition too for single column constraints?

eviljeff: same question about the (non-fk) columns that aren't declared with `db_index=True` I suppose.

diox: It's not necessary if it's in `constraints`/`indexes`. In fact, at some point I ran into some issues when I had both and a custom name, I think django was confused and tried to add the index twice (I think it was with the `getpersonas_id` field, which is why I removed its `db_index` property)

src/olympia/addons/models.py
diox: We can. I suspect it's no longer necessary anyway...

diox:
> We should probably decide on best practises for db_index=True and unique=True too - should we consistently not define them in the field kw if defined in the classes Meta instead? Not specific to this issue/pr and nothing to do here, just saying.

We should definitely not declare them in the field if they are declared in the meta properties.

What I'm unsure about is whether we should completely replace `db_index=True` and `unique=True` with the meta properties going forward. On the one hand, declaring them in the field is straightforward, hard to forget when you're adding a new field. On the other, it's not flexible, and we have to use `indexes` and `constraints` for our indexes with custom names anyways...

PR #12257, opened 2019-09-02
add --batch-size param to process_addons

eviljeff:


src/olympia/addons/tests/test_commands.py
```diff
@@ -122,6 +122,30 @@ def test_process_addons_limit_addons():
         assert calls[0]['kwargs']['args'] == [addon_ids[:2]]
 
 
+@freeze_time('2019-04-01')
+@pytest.mark.django_db
+def test_process_addons_batch_size():
+    addon_ids = [
+        addon_factory(status=amo.STATUS_APPROVED).id for _ in range(101)
```
diox: This is going to be terribly slow... I would skip this and mock `chunked` instead.

eviljeff: https://github.com/mozilla/addons-server/pull/12257/commits/3e67803523aa009737c0f85be789955c54d43290

PR #12281, opened 2019-09-05
db schema update for collections onwards

eviljeff:
part of #10917 - should be the remaining indexes/uniques (that aren't foreign keys)

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -1727,11 +1727,15 @@ class AddonUserPendingConfirmation(SaveUpdateMixin, models.Model):
 
     class Meta:
         db_table = 'addons_users_pending_confirmation'
-        unique_together = (('addon', 'user'),)
         indexes = [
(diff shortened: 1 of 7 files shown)


src/olympia/scanners/models.py
```diff
@@ -20,4 +20,8 @@ class ScannersResult(ModelBase):
 
     class Meta:
         db_table = 'scanners_results'
-        unique_together = ('upload', 'scanner', 'version',)
```
eviljeff: drive-by fix so we have no more `unique_together`

src/olympia/addons/models.py
```diff
@@ -1727,12 +1727,16 @@ def __init__(self, *args, **kwargs):
 
     class Meta:
         db_table = 'addons_users_pending_confirmation'
-        unique_together = (('addon', 'user'),)
```
eviljeff: drive-by fix so we have no more `unique_together`

PR #12289, opened 2019-09-06
Create stale.yml

### LINKED ISSUE #12288: Configure Probot Stale
Enabled in https://bugzilla.mozilla.org/show_bug.cgi?id=1579128

We just need to configure it now.

muffinresearch:
Fixes  #12288

This patch adds a stale config. Once this is good, this process can be run for all other add-ons repos since probot stale has been enabled on all of them.

Should we add a `state: never-stale` label to the label exemptions list for an explicit opt-out?

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -16,7 +16,7 @@ exemptLabels:
   - "contrib: good first bug"
   - "contrib: welcome"
   - "component: security"
-  - "type: papercut"
+  - "neverstale"
 
 # Set to true to ignore issues in a project (defaults to false)
 exemptProjects: true
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -22,10 +22,10 @@ exemptLabels:
 exemptProjects: true
 
 # Set to true to ignore issues in a milestone (defaults to false)
-exemptMilestones: false
+exemptMilestones: true
 
 # Set to true to ignore issues with an assignee (defaults to false)
-exemptAssignees: false
+exemptAssignees: true
 
 # Label to use when marking as stale
 staleLabel: "state: stale"
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,11 +1,11 @@
 # Configuration for probot-stale - https://github.com/probot/stale
 
 # Number of days of inactivity before an Issue or Pull Request becomes stale
-daysUntilStale: 60
+daysUntilStale: 180
 
 # Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
 # Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
-daysUntilClose: 7
+daysUntilClose: 14
 
 # Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
 onlyLabels: []
@@ -33,8 +33,8 @@ staleLabel: "state: stale"
 # Comment to post when marking as stale. Set to `false` to disable
 markComment: >
   This issue has been automatically marked as stale because it has not had
-  recent activity. It will be closed if no further activity occurs. Thank you
-  for your contributions.
+  recent activity. If you think this bug should stay open, please comment on 
+  the issue with further details. Thank you for your contributions.
 
 # Comment to post when removing the stale label.
 # unmarkComment: >
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -16,12 +16,13 @@ exemptLabels:
   - "contrib: good first bug"
   - "contrib: welcome"
   - "component: security"
+  - "type: papercut"
 
 # Set to true to ignore issues in a project (defaults to false)
 exemptProjects: true
 
 # Set to true to ignore issues in a milestone (defaults to false)
-exemptMilestones: true
+exemptMilestones: false
 
 # Set to true to ignore issues with an assignee (defaults to false)
 exemptAssignees: false
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,62 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - "contrib: maybe good first bug"
+  - "contrib: good first bug"
+  - "contrib: welcome"
+  - "component: security"
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: true
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: true
+
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: false
+
+# Label to use when marking as stale
+staleLabel: "state: stale"
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+
+# Comment to post when removing the stale label.
+# unmarkComment: >
+#   Your comment here.
+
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,62 @@
...
+  - "contrib: maybe good first bug"
+  - "contrib: good first bug"
+  - "contrib: welcome"
+  - "component: security"
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: true
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: true
```
muffinresearch: On reflection, I'm thinking that this one is ok to be false.

.github/stale.yml
```diff
@@ -0,0 +1,62 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
```
EnTeQuAk: I find 60 days a relatively short amount of time. I could imagine that issues won't get tackled for a quarter and maybe only get more activity / re-triaged after a quarter of work has been done and we go on to the next things to do? So maybe 90-100 days may be more safe to really make sure the issue hasn't been considered for more than a quarter?

.github/stale.yml
```diff
@@ -0,0 +1,62 @@
...
+exemptAssignees: false
+
+# Label to use when marking as stale
+staleLabel: "state: stale"
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
```
EnTeQuAk: I wonder if some kind of call to action might be worth adding, something like "If you think this issue should not be closed, please do let us know and we can re-triage."

:thinking:

.github/stale.yml
eviljeff: idea: could leave the stale at 60 and change `daysUntilClose` to 30?

.github/stale.yml
```diff
@@ -0,0 +1,62 @@
...
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - "contrib: maybe good first bug"
+  - "contrib: good first bug"
+  - "contrib: welcome"
+  - "component: security"
```
eviljeff: my preference would be against a straight "neverstale" label because it doesn't explain _why_ we don't want it never to be stale, but perhaps there are other labels we could add to the list that would be used for issues we really want to fix... eventually.  I thought about 'priority: p2` but those should never be left alone for 60 days.  

Finally a reason for the `priority: p2.5` label 🤣

eviljeff: Or, we just have some ongoing projects papercuts and other things we're intending to fix that we can add the issues to to keep them non-stale.  Mmm, yes, I think I'd prefer that.

.github/stale.yml
muffinresearch: Then again its marked as stale and doesn't get a response within a certain window it probably never will. Maybe 14 days to allow a bit more time would be the middle ground?

.github/stale.yml
muffinresearch: > Or, we just have some ongoing projects papercuts and other things we're intending to fix that we can add the issues to to keep them non-stale. Mmm, yes, I think I'd prefer that.

So you want "type: papercut" added here, any others?

.github/stale.yml
EnTeQuAk: Given that PTO can (and should… to get a reasonable amount of breathe time) be > 14 days, I'd also vote for 30 days as someone (the author) being disconnected for more than 30 days is probably rare.

Obviously, the author can reply to an closed issue and we can handle that too… not sure what feels best, I think I'd go for @eviljeff's idea

.github/stale.yml
muffinresearch: > my preference would be against a straight "neverstale" label because it doesn't explain _why_ we don't want it never to be stale, but perhaps there are other labels we could add to the list that would be used for issues we really want to fix... eventually. I thought about 'priority: p2` but those should never be left alone for 60 days.
> 
> Finally a reason for the `priority: p2.5` label 🤣

The use-case for an explicit `neverstale` label of some kind is having a direct way to tell stalebot to stay out of the way. It wouldn't always be the case that a label in `exemptLabels` would be a good fit. If that label is used then I'd expect the comments to provide a rationale.

.github/stale.yml
muffinresearch: I think that's already implied no?

EnTeQuAk: Being explicit is always better than implicitly implying something. Especially for people not too familiar with the project or open source contributions often need a bit more hand-holding. I also think it's a good habit to encourage people to help us prioritize their issues.

.github/stale.yml
eviljeff: > So you want "type: papercut" added here, any others?

no, I mean we'd have actual project to collect issues we specifically care about and are intending to fix.  Issues in projects are already excluded according the config.

.github/stale.yml
muffinresearch: On the basis that an issue can still be re-opened I don't see too much of an issue with shorter times overall. This might be a case of trying it and adapting the rules if they feel like they're not working. Also could be interesting to see what bugzilla's settings are for things like this if we can find them.

.github/stale.yml
muffinresearch: > > So you want "type: papercut" added here, any others?
> 
> no, I mean we'd have actual project to collect issues we specifically care about and are intending to fix. Issues in projects are already excluded according the config.

Ah ok, I'll pull that.

.github/stale.yml
diox: Looks like I'm in the minority here... I was thinking of having a much higher delay like 180 days! While issues can still be re-opened, it's slightly harder to find issues that have been deleted (more noise to go through).

.github/stale.yml
```diff
@@ -0,0 +1,63 @@
...
+  - "neverstale"
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: true
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: false
```
diox: I would ignore those. I know it's unlikely to happen though :)

.github/stale.yml
```diff
@@ -0,0 +1,63 @@
...
+
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: false
```
diox: I would ignore those. I try to keep my backlog of things I want to work on to less than a page of results (25), so I stay assigned on bugs I really want to fix someday. Doesn't mean I will of course, but I'd prefer to keep them opened, and unassign myself the day I get tired of seeing them and admit I'll never fix them.

.github/stale.yml
diox: :+1: on @EnTeQuAk 's suggestion, I think it's not explicit enough. Maybe word that in such a way that encourages people to help in some way? A less aggressive version of "send us your patch if you think it should stay opened" :-)

muffinresearch: Fair enough, I think it would be better to re-word the comment, rather than try to add something to reinforce what's already there. To me  "It will be closed if no further activity occurs" is quite clear and unambiguous about what will happen. 

If we include something with a more concrete action re: having users ask for issues to be re-triaged, then we'll need to be sure that this is a process that will happen and doesn't get missed. Any thoughts on how to do that?

.github/stale.yml
muffinresearch: Does everyone do that though? 😀

diox: Maybe not, but I think it should be the assignee's responsibility to close or unassign issues already assigned to them, not a bot.

.github/stale.yml
muffinresearch: > Looks like I'm in the minority here... I was thinking of having a much higher delay like 180 days! While issues can still be re-opened, it's slightly harder to find issues that have been deleted (more noise to go through).

I've no problem with the initial delay being longer. 6 months seems ok for what we want this for.

muffinresearch: I've updated  `daysUntilStale: 180` and `daysUntilClose: 14`

Does that seem ok to start?

.github/stale.yml
muffinresearch: Reworded to:

```yaml
markComment: >
  This issue has been automatically marked as stale because it has not had
  recent activity. If you think this bug should stay open, please comment on 
  the issue with further details. Thank you for your contributions.
```

PR #12293, opened 2019-09-06
Migrations to bring dev/stage/prod columns and constraints closer to django

diox:
Tables m to z, except for the following which are not included on purpose:
- `translations`, `update_counts`, `users`, `versions`: waiting for next push window so that the migration step of the push doesn't take too much time
- `waffle_flag`, `waffle_flag_groups`, `waffle_flag_users`, `waffle_sample`, `waffle_switch`: since those are coming from the external `waffle` app, need additional investigation into whether they should be done via schematic or via django-migrations

Part of #10917


diox:
Depends on https://github.com/mozilla/addons-server/pull/12283 being merged first

PR #12302, opened 2019-09-09
Add UPDATE statement to fix broken data in preview table

diox:


diox:
Yeah tested on dev earlier

PR #12305, opened 2019-09-09
Convert recently created tables default charset to utf8mb4

### LINKED ISSUE #12304: Convert recently created tables and switch database default charset to utf8mb4
Part of #3705 and #10917

Tables created recently with `schematic` are not in `utf8mb4`, so we need to fix that and make `utf8mb4` the default at the database level.

```sql
> SHOW VARIABLES LIKE  'char%';
+--------------------------+-------------------------------------------+
| Variable_name            | Value                                     |
+--------------------------+-------------------------------------------+
| character_set_client     | utf8mb4                                   |
| character_set_connection | utf8mb4                                   |
| character_set_database   | utf8                                      |
| character_set_filesystem | binary                                    |
| character_set_results    | utf8mb4                                   |
| character_set_server     | utf8                                      |
| character_set_system     | utf8                                      |
| character_sets_dir       | /rdsdbbin/mysql-5.7.16.R1/share/charsets/ |
+--------------------------+-------------------------------------------+
```

`character_set_database` is the one that needs to be fixed.

diox:
Fixes #12304

### DIFF
--- a/src/olympia/migrations/1181-utf8mb4.sql
+++ b/src/olympia/migrations/1181-utf8mb4.sql
@@ -1,0 +1,15 @@
+ALTER TABLE `editor_subscriptions` COMMENT '';
+
+ALTER TABLE `addons_reusedguid` CONVERT TO CHARACTER SET utf8mb4;
+ALTER TABLE `addons_users_pending_confirmation` CONVERT TO CHARACTER SET utf8mb4;
+ALTER TABLE `api_apikeyconfirmation` CONVERT TO CHARACTER SET utf8mb4;
+ALTER TABLE `hero_primaryhero` CONVERT TO CHARACTER SET utf8mb4;
+ALTER TABLE `hero_secondaryhero` CONVERT TO CHARACTER SET utf8mb4;
+ALTER TABLE `hero_secondaryheromodule` CONVERT TO CHARACTER SET utf8mb4;
+ALTER TABLE `log_activity_comment_draft` CONVERT TO CHARACTER SET utf8mb4;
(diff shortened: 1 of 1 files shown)


bqbn:
Question: Why this issue even exists? The reason I'm asking is that we converted all tables to utf8mb4 in Q1, since then, shouldn't we always create tables with that charset? 


Unless the so called "recently created tables" haven't been created in -stage or -prod yet, and only exist in -dev environment, in which case, maybe it is better to fix them manually in -dev environment. That way we don't have to run `alter database` in -prod.

diox:
@bqbn this issue exist precisely because the `ALTER DATABASE` had not been ran, at least not on dev & stage. Migration that created tables since we did the conversion in Q1 were created using the implicit database charset, which was wrong in those cases, resulting in us needing to fix the tables.

I can't check prod, but I suspect the problem is also present there. In any case, the `ALTER DATABASE` statement is trivial, it doesn't make any modifications to the tables themselves (and the tables that do need fixing, which we're taking care of in a migration, are quite small)

src/olympia/migrations/1179-utf8mb4.sql
```diff
@@ -0,0 +1,15 @@
+ALTER TABLE `editor_subscriptions` COMMENT '';
```
eviljeff: ugh, is this the syntax to clear a comment?

diox: Yeah, I couldn't find another way to do it...

bqbn:
I had thought `ALTER DATABASE` had been run when we did the migration. Sorry about that. 😞

PR #12316, opened 2019-09-10
Move CI envvar definition to selenium override config.

EnTeQuAk:
This should fix autoreload.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,6 @@ x-env-mapping: &env
     - TERM=xterm-256color
     - UITEST_FXA_EMAIL
     - UITEST_FXA_PASSWORD=uitester
-    - CI=${CI}
 
 services:
   worker: &worker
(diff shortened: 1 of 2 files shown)


EnTeQuAk:
```
Hint: Exit code 137 typically means the process is killed because it was running out of memory
Hint: Check if you can optimize the memory usage in your app
Hint: Max memory usage of this container is 7429345280
 according to /sys/fs/cgroup/memory/memory.max_usage_in_bytes
```

This is the issue we originally ran ran into, let me see how I broke CI :man_shrugging:

jrbenny35:
How was autoreload broken?

EnTeQuAk:
The problem is that if-not-env only checks for the environment variable existence, not it's value. I hoped that moving the forward definition to a different config file would fix that he apparently it breaks ci again by not correctly avoiding autoreload to start there...

EnTeQuAk:
Trying out a different approach to how we're handling the environment variable in uwsgi, let's see

PR #12317, opened 2019-09-10
Send download_url to the scanners

### LINKED ISSUE #12236: Update `run_scanner` to send a `FileUpload` download URL instead of the content
Instead of sending the xpi content, we should send the secured download URL implemented in #12232 to the cloud functions. That will fix the GCP limitation on the request body size.

willdurand:
Fixes #12236

---

You need to checkout the `aws` branch in the customs repo and the rebuild the `customs` Docker image.

You need to checkout the `functions` branch in the wat repo and the rebuild the `wat` Docker image.

~~⚠️ the `wat` repo has not been updated yet, so it does not work with wat.~~

### DIFF
--- a/docker-compose.private.yml
+++ b/docker-compose.private.yml
@@ -9,7 +9,17 @@ services:
   customs:
     build:
       context: ./private/addons-customs-scanner
+    environment:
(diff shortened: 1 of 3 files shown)


willdurand:
rebased and ready for r?

PR #12319, opened 2019-09-10
Allow admins to reset API keys for a developer or group of developers

### LINKED ISSUE #12280: Allow admins to reset API Credentials confirmation status for a developer or group of developers
Followup for https://github.com/mozilla/addons-server/issues/11635

> After confirmation, re-generating keys shouldn't require confirmation again. However, it should be possible to reset the confirmation status of any or all developers in order to require confirmation again. By "possible" I just mean it should be visible in the user admin and it should be easy to ask ops to change this flag for a group of users if needed.

STR:
The following actions should be available in admin tools:
1. Reset the API Credentials confirmation status for developers.
2.  Reset the API Credentials confirmation status for developers.

diox:
This includes the API Credentials Confirmation state, so any affected developers will have to go through the email confirmation again.

Fixes #12280

### DIFF
--- a/src/olympia/users/templates/admin/users/userprofile/change_form.html
+++ b/src/olympia/users/templates/admin/users/userprofile/change_form.html
@@ -10,10 +10,6 @@
               They have to be *after* the other submit buttons, otherwise they
               could become the default submit button of the page
              {% endcomment %}
-             {% if can_ban %}
-                 <button formaction="{% url 'admin:users_userprofile_ban' original.pk|admin_urlquote %}"
-                         type="submit">{% trans "Ban" %}</button>
-             {% endif %}
              {% if has_users_edit_permission %}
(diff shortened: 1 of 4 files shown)


src/olympia/users/templates/admin/users/userprofile/change_form.html
```diff
@@ -14,6 +14,12 @@
                  <button formaction="{% url 'admin:users_userprofile_ban' original.pk|admin_urlquote %}"
```
eviljeff: you forgot to remove this block still referencing `can_ban`

diox: woops good eye, fixed in ab63a486f8627b7308d5cb3c27bafdbb65099372

Issue #12314, opened 2019-09-10
Remove the fallback code in `HttpResponseSendFile`


willdurand:
In https://github.com/mozilla/addons-server/issues/12313, we decided to leverage the XSENDFILE feature locally. That way, we're using the same feature pretty much everywhere.

We could now remove the fallback code in `HttpResponseSendFile` and update the test suite accordingly. I believe we could also remove the `XSENDFILE` config parameter.

PR #12328, opened 2019-09-11
Cleanup HttpResponseSendFile and remove XSENDFILE config param

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12314

---

This patch:

- removes the `XSENDFILE` config parameter
- removes the logic to stream the content of an `HttpResponseSendFile`
when `XSENDFILE = False`
- adds a new test case for `HttpResponseSendFile`
- removes some more code in `HttpResponseSendFile`, which I think is not
needed. According to this code:
https://github.com/django/django/blob/b616908ce1ad46457c4b49fe098320bb27e09581/django/http/response.py#L107-L137,
Django automatically does the conversion and it uses
[`email.header.Header`](https://docs.python.org/3/library/email.header.html#email.header.Header)
to mime encode the value of the header

### DIFF
--- a/settings_test.py
+++ b/settings_test.py
@@ -56,9 +56,6 @@ TASK_USER_ID = 1337
 ES_DEFAULT_NUM_REPLICAS = 0
 ES_DEFAULT_NUM_SHARDS = 1
 
-# Set to True if we're allowed to use X-SENDFILE.
-XSENDFILE = True
-
 # Don't enable the signing by default in tests, many would fail trying to sign
 # empty or bad zip files, or try posting to the endpoints. We don't want that.
 SIGNING_SERVER = ''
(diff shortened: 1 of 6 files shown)


EnTeQuAk:
> Questions: why is the test file suffixed with _ (test_utils_.py)? Can we rename it?

I don't think there is one, let's rename it. It's been named like that since 2014 7e47bea2b07ea2e97d2338c8e0cc23fe9cd6e622

diox:
>> Questions: why is the test file suffixed with _ (test_utils_.py)? Can we rename it?
> I don't think there is one, let's rename it. It's been named like that since 2014 7e47bea

I also wondered the same thing when I joined the project... IIRC a while ago, addons-server, when it was called zamboni, had a dependency on the [`test_utils` package](https://pypi.org/project/test-utils/). To prevent potential import conflicts the convention to add an underscore to any file called `test_utils` was adopted. The dependency is long gone but we never renamed the files...

PR #12329, opened 2019-09-11
Update remaining table schemas (m-z)

diox:
This includes dropping some indexes that were added by mistake duplicating already existing ones, and taking care of the remaining changes on "big" tables (`translations`, `versions`, `update_counts`, `users`)

part of #10917


diox:
Ah forgot `version_previews`, I'll add it, it's part of the same group of tables for which I added an extra duplicate index by mistake. The waffle ones we need to determine if the included migrations would be enough in `django-waffle` for them.

Issue #12321, opened 2019-09-10
Make reviewer tools abuse reports page faster


diox:
The reviewer tools abuse reports page is pretty slow when there are a full page of abuse reports shown, because it duplicates the add-on query at well as all its transforms. Let's fix that.

Issue #11775, opened 2019-07-03
Double entry for the Abuse Reports section on the /reviewers/abuse-reports/ page


AlexandraMoga:
**STR**

1. Log in to AMO stage with a reviewer account
2. Open to the Auto-approval queue and select an add-on to review
3. Click on Abuse reports to view all the reports for the selected add-on - see https://reviewers.addons.allizom.org/en-US/reviewers/abuse-reports/botblockerlocale

**Actual result:**
There are two entries for the Abuse reports section

**Expected result:**
Abuse reports section appears only once

**Notes:**
- reproduced on -dev and stage with Fx67, WIn10x64

![double abuse](https://user-images.githubusercontent.com/31961530/60598919-3b89b500-9db6-11e9-9dd7-980e8803ea3e.png)

Patil2099:
Can I work on this @eviljeff ?

eviljeff:
I think you just need to drop the abuse report inclusion from the specific abuse-reports template, but I've not looked through the code.

Patil2099:
I will look through that template and inform you about it. @eviljeff

Patil2099:
@eviljeff  I saw that reviewers/addon_details_box.html had a abuse reports table in it so I removed the table from abuse_reports.html because addon_details_box.html was included at various places but the abuse_reports.html was not so I changed abuse_reports.html. The expected result was achieved by it.

Vishal19111999:
Hey! is this repo good for beginners? I have experience in python and C++.

diox:
Sure. This is probably just a template inclusion problem, @Patil2099 already worked on fixing this in https://github.com/mozilla/addons-server/pull/11961 but we closed the PR because they couldn't get addons-server to work locally. If someone could test that fix locally or even write a unit test in reviewers/tests/test_views.py (maybe just modifying TestReview.test_abuse_reports() is enough) that'd be great.

diox:
Taking this myself, because the right fix is actually slightly more involved than that.

AlexandraMoga:
Verified fixed on -dev

There's only one entry for abuse reports now:
![2019-09-16_14-15-17](https://user-images.githubusercontent.com/31961530/64953794-75823780-d88c-11e9-8920-b62dc7b9de8c.gif)

Issue #11365, opened 2019-05-08
Theme images section is empty in the abuse reports page


ioanarusiczki:
Followup for https://github.com/mozilla/addons-server/issues/11137

STR:
1. User needs an admin account - load an abuse report page for a static theme
2. Observe the Theme images section

Expected:
A preview of the theme is available.

Actual result:
No image.

Notes:
This issue is reproducible on AMO dev - FF66(Win10). 

![empty](https://user-images.githubusercontent.com/33448286/57358695-a768ec00-717e-11e9-82ef-fef6f61155ba.png)

eviljeff:
the template context is missing `version`

diox:
Coming back to this. I think we want to display the `current_version` - the most likely scenario for abuse reports against themes is not going to be themes installed in Firefox anyway, but rather themes listed on AMO and someone noticing them while browsing AMO.

ioanarusiczki:
@diox I checked on some pages on dev such as 
https://reviewers.addons-dev.allizom.org/en-US/reviewers/abuse-reports/space-stars
https://reviewers.addons-dev.allizom.org/en-US/reviewers/abuse-reports/underwater-search?page=1
or
https://reviewers.addons-dev.allizom.org/en-US/reviewers/abuse-reports/sunset-over-water
I think the content has doubled and I still don't see the theme preview displayed. 
I see Abuse Reports(7) then Theme Images which is empty then again an Abuse Report section (looks to me like it's twice the same content). 

![layout problems](https://user-images.githubusercontent.com/33448286/64540624-2aff3900-d329-11e9-8c3a-2515f5d27fc4.gif)

I'm going to reopen.

diox:
Ah, I fixed the images in the abuse report *admin* tool, not the reviewers one. Still, I'll do it there as well. The doubling is tracked in https://github.com/mozilla/addons-server/issues/11775 but I might tackle that as the same time.

ioanarusiczki:
@diox Looks good to me - verified on AMO dev - FF69(Win10). 
Also, the doubling is fixed too. 

![images for themes](https://user-images.githubusercontent.com/33448286/64857609-afa1ce00-d62d-11e9-8e75-002e9e5bd411.gif)

Issue #12043, opened 2019-08-06
Extraneous queries in review page caused by ReviewForm


diox:
Part of #11797

`ReviewForm` loads all a bunch of `Version` from the addon to allow the reviewer to reject multiple versions at once. It's bound to create duplicate queries with the loading of the latest and current version that the view already does, not to mention the display of the 10 versions shown in the history, but we can live with that. The main problem is that we only use the version number and pk for each version, but currently we go through the transformer and load translations, files & applications versions for these versions. It only happens once (thanks to being done by a transformer at the queryset level), but still, that's data we don't need to fetch.

It doesn't seem trivial to fix however. I've tried adding a `.no_transforms()` to the queryset being set in the `__init__` of the form, and that made things worse for some reason.

diox:
Downgrading priority because it actually has a low impact. It would be nice to fix but that's just 2 extra queries and some columns.

diox:
Not worth the effort.

PR #12332, opened 2019-09-11
Make fewer SQL queries in reviewer pages displaying abuse reports

diox:
In addition, in the abuse reports view itself, prevent showing the abuse reports twice and display theme images from current version.

Fixes https://github.com/mozilla/addons-server/issues/12321
Fixes https://github.com/mozilla/addons-server/issues/11775
Fixes https://github.com/mozilla/addons-server/issues/11365
Helps https://github.com/mozilla/addons-server/issues/12043

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -1218,7 +1218,7 @@ class Addon(OnChangeMixin, ModelBase):
         return not self.is_deleted
 
     def has_listed_versions(self):
-        return self.versions.filter(
+        return self.current_version or self.versions.filter(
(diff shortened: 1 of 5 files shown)


src/olympia/addons/models.py
```diff
@@ -1218,7 +1218,7 @@ def can_be_deleted(self):
         return not self.is_deleted
 
     def has_listed_versions(self):
-        return self.versions.filter(
+        return self.current_version or self.versions.filter(
```
diox: Doesn't look like much, but it's one (or more) extra query avoided on every page in devhub, in some cases the detail API, and many reviewer pages.

PR #12356, opened 2019-09-13
expose site_status in UserProfileSerializer and /site/

### LINKED ISSUE #11493: Add banner to user and developer facing AMO pages
Until we have a fix for the certificate issue, let's put up a banner on all AMO pages that face users and developers that says: 

.You may be experiencing issues with extensions and themes for Firefox. Please see https://discourse.mozilla.org/t/certificate-issue-causing-add-ons-to-be-disabled-or-fail-to-install/39047 for more information."

eviljeff:
fixes #11493

### DIFF
--- a/docs/topics/api/overview.rst
+++ b/docs/topics/api/overview.rst
@@ -257,7 +257,7 @@ Site Status
 .. _`api-site-status`:
 
 This special endpoint just returns if the site is in read only mode, and if there is a site notice currently in effect.
-See :ref:`maintainance mode <api-overview-maintainance>` for mode details of when the site is read only and how requests are affected.
+See :ref:`maintainance mode <api-overview-maintainance>` for more details of when the site is read only and how requests are affected.
 
 
 .. http:get:: /api/v4/site/
--- a/docs/topics/api/overview.rst
+++ b/docs/topics/api/overview.rst
@@ -264,7 +264,7 @@ See :ref:`maintainance mode <api-overview-maintainance>` for more details of whe
 
     .. _site-status-object:
 
-    :>json boolean read_only: Is the site in read-only mode.
+    :>json boolean read_only: Whether the site in read-only mode.
     :>json string notice: A site-wide notice about any current known difficulties or restrictions.  If this API is being consumed by a tool/frontend it should be displayed to the user.
 
 
--- a/docs/topics/api/overview.rst
+++ b/docs/topics/api/overview.rst
@@ -256,7 +256,7 @@ Site Status
 
 .. _`api-site-status`:
 
-This special endpoint just returns if the site is in read only mode, and if there is a site notice currently in effect.
+This special endpoint returns if the site is in read only mode, and if there is a site notice currently in effect.
 See :ref:`maintainance mode <api-overview-maintainance>` for more details of when the site is read only and how requests are affected.
 
 
@@ -265,7 +265,7 @@ See :ref:`maintainance mode <api-overview-maintainance>` for more details of whe
     .. _site-status-object:
 
     :>json boolean read_only: Whether the site in read-only mode.
-    :>json string notice: A site-wide notice about any current known difficulties or restrictions.  If this API is being consumed by a tool/frontend it should be displayed to the user.
+    :>json string|null notice: A site-wide notice about any current known difficulties or restrictions.  If this API is being consumed by a tool/frontend it should be displayed to the user.
 
 
 ------------
--- a/docs/topics/api/overview.rst
+++ b/docs/topics/api/overview.rst
@@ -89,6 +89,8 @@ constants are as follows:
     ========================  =========================================================
 
 
+.. _api-overview-maintainance:
+
 ~~~~~~~~~~~~~~~~~
 Maintainance Mode
 ~~~~~~~~~~~~~~~~~
@@ -111,6 +113,7 @@ The response when returning ``HTTP 503 Service Unavailable`` in case of read-onl
     }
 
 In case we are not in read-only mode everything should be back working as normal.
+To check if the site is in read-only mode before submitting a response the :ref:`site status api<api-site-status>` can be checked.
 
 .. _api-overview-pagination:
 
@@ -246,6 +249,25 @@ specified.
 
 .. _api-versions-list:
 
+
+-----------
+Site Status
+-----------
+
+.. _`api-site-status`:
+
+This special endpoint just returns if the site is in read only mode, and if there is a site notice currently in effect.
+See :ref:`maintainance mode <api-overview-maintainance>` for mode details of when the site is read only and how requests are affected.
+
+
+.. http:get:: /api/v4/site/
+
+    .. _site-status-object:
+
+    :>json boolean read_only: Is the site in read-only mode.
+    :>json string notice: A site-wide notice about any current known difficulties or restrictions.  If this API is being consumed by a tool/frontend it should be displayed to the user.
+
+
 ------------
 API Versions
 ------------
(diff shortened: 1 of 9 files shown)


docs/topics/api/overview.rst
```diff
@@ -246,6 +249,25 @@ specified.
...
 .. _api-versions-list:
 
+
+-----------
+Site Status
+-----------
+
+.. _`api-site-status`:
+
+This special endpoint just returns if the site is in read only mode, and if there is a site notice currently in effect.
```
willdurand: Not sure I fully understand this sentence.

docs/topics/api/overview.rst
```diff
@@ -246,6 +249,25 @@ specified.
...
+See :ref:`maintainance mode <api-overview-maintainance>` for mode details of when the site is read only and how requests are affected.
```
willdurand:
```suggestion
See :ref:`maintainance mode <api-overview-maintainance>` for more details of when the site is read only and how requests are affected.
```

docs/topics/api/overview.rst
```diff
@@ -111,6 +113,7 @@ The response when returning ``HTTP 503 Service Unavailable`` in case of read-onl
     }
 
 In case we are not in read-only mode everything should be back working as normal.
+To check if the site is in read-only mode before submitting a response the :ref:`site status api<api-site-status>` can be checked.
```
willdurand:
```suggestion
To check if the site is in read-only mode before submitting a response, the :ref:`site status api<api-site-status>` can be called.
```

docs/topics/api/overview.rst
```diff
@@ -246,6 +249,25 @@ specified.
...
+
+This special endpoint just returns if the site is in read only mode, and if there is a site notice currently in effect.
+See :ref:`maintainance mode <api-overview-maintainance>` for mode details of when the site is read only and how requests are affected.
+
+
+.. http:get:: /api/v4/site/
+
+    .. _site-status-object:
+
+    :>json boolean read_only: Is the site in read-only mode.
```
willdurand:
```suggestion
    :>json boolean read_only: Whether the site in read-only mode.
```

docs/topics/api/overview.rst
```diff
@@ -246,6 +249,25 @@ specified.
...
+    :>json string notice: A site-wide notice about any current known difficulties or restrictions.  If this API is being consumed by a tool/frontend it should be displayed to the user.
```
willdurand: Will the message be localized?

src/olympia/accounts/tests/test_serializers.py
```diff
@@ -250,6 +251,21 @@ def test_validate_homepage(self):
...
             assert serializer.validate_homepage(allowed_url) == allowed_url
 
+    def test_site_status(self):
+        data = super(TestUserProfileSerializer, self).test_basic()
+        assert data['site_status'] == {
+            'read_only': False,
+            'notice': None,
+        }
+
+        set_config('site_notice', 'THIS is NOT Á TEST!')
```
willdurand: We could maybe have one `assert` here to cover the case where the site is not in read only mode but there is a message to display to the users.

src/olympia/amo/views.py
```diff
@@ -131,3 +134,15 @@ def _frontend_view(*args, **kwargs):
...
+class SiteStatusView(APIView):
+    authentication_classes = []
+    permission_classes = []
+
+    @classmethod
+    def as_view(cls, **initkwargs):
+        return non_atomic_requests(super().as_view(**initkwargs))
+
+    def get(self, request, format=None):
+        return Response(SiteStatusSerializer(object()).data)
```
willdurand: Why do we need `object()` here?

src/olympia/accounts/serializers.py
```diff
@@ -79,12 +80,13 @@ class UserProfileSerializer(PublicUserProfileSerializer):
     reviewer_name = serializers.CharField(
         min_length=2, max_length=50, allow_blank=True,
         validators=[OneOrMorePrintableCharacterAPIValidator()])
+    site_status = SiteStatusSerializer(source='*')
```
willdurand: Why do we need a `source='*'` here?

docs/topics/api/overview.rst
eviljeff: No.

docs/topics/api/overview.rst
```diff
@@ -246,6 +249,25 @@ specified.
...
+This special endpoint just returns if the site is in read only mode, and if there is a site notice currently in effect.
+See :ref:`maintainance mode <api-overview-maintainance>` for more details of when the site is read only and how requests are affected.
+
+
+.. http:get:: /api/v4/site/
+
+    .. _site-status-object:
+
+    :>json boolean read_only: Is the site in read-only mode.
+    :>json string notice: A site-wide notice about any current known difficulties or restrictions.  If this API is being consumed by a tool/frontend it should be displayed to the user.
```
willdurand: This should probably return `null` when it is not set, no?

src/olympia/amo/views.py
eviljeff: It's just how serializers are written - SiteStatusSerializer isn't really serializing input or retrieved data in the strictest sense, and serializers are supposed to be serializing _something_. You either have to pass an instance of an object (an empty object is the most trivial example), or pass data, but then you have to call validate too (and extra few lines of code)

src/olympia/accounts/serializers.py
eviljeff: similar to the `object()` above - when a serializer is used as a field inside a serializer it's passed some data from the parent object/data, and by default it's the name of the field.  As `UserProfile` doesn't contain a `site_status` property it fails so you have to give it a source of another field.  `source='*'` is just passing the entire UserProfile.  Could be any property.

src/olympia/amo/views.py
willdurand: ah, ok I see now, thanks

src/olympia/accounts/serializers.py
willdurand: I see, thanks. Could you maybe add a comment above this line to explain that?

docs/topics/api/overview.rst
EnTeQuAk: This kinda sounds like it'll be blocking if the site is not in read-only mode which isn't quite what you wanted to say I suppose. I wonder if simply removing the "just" makes it much clearer already.

src/olympia/amo/tests/test_views.py
```diff
@@ -443,3 +445,27 @@ def test_version_json(self):
...
+            'notice': None,
+        }
+
+        set_config('site_notice', 'THIS is NOT Á TEST!')
+        with override_settings(READ_ONLY=True):
+            response = self.client.get(self.url)
+        assert response.data == {
+            'read_only': True,
+            'notice': 'THIS is NOT Á TEST!',
+        }
```
EnTeQuAk: Let's add two more test-cases that check for `READ_ONLY=False` but `site_notice` set and `READ_ONLY=True` but `site_notice` unset.

src/olympia/accounts/tests/test_serializers.py
eviljeff: https://github.com/mozilla/addons-server/pull/12356/commits/4acd93a499ade786efc4967b45807f2550d4a248

src/olympia/accounts/serializers.py
eviljeff: https://github.com/mozilla/addons-server/pull/12356/commits/4acd93a499ade786efc4967b45807f2550d4a248

docs/topics/api/overview.rst
eviljeff: https://github.com/mozilla/addons-server/pull/12356/commits/4acd93a499ade786efc4967b45807f2550d4a248

src/olympia/amo/tests/test_views.py
eviljeff: https://github.com/mozilla/addons-server/pull/12356/commits/4acd93a499ade786efc4967b45807f2550d4a248

docs/topics/api/overview.rst
eviljeff: https://github.com/mozilla/addons-server/pull/12356/commits/4acd93a499ade786efc4967b45807f2550d4a248

Issue #12344, opened 2019-09-12
Remove duplicate addon queries in reviewer tools caused by version history


diox:
Part of https://github.com/mozilla/addons-server/issues/11797

On every review page, the add-on and its transformers are run multiple times. One of the culprits is the loading of the activity logs, tracked in https://github.com/mozilla/addons-server/issues/12055, but another is the fact that we end up accessing the `addon` field on every `Version` displayed in the history, but django doesn't know that it's the same `addon` it already has loaded. So every `Version` displayed in the history causes something like 8 extra queries *per version*.

diox:
While working on this I stumbled upon the queries we make to support hard-deleted versions. Before 982e86965d84e0e589268948376d6602a81a451f (deployed in 2016-03-03) versions used to be hard-deleted. But because we wanted to be display reviewer comments left on versions that had since been deleted, a [hack had been put in place years before](https://github.com/mozilla/addons-server/commit/e4e5b977ca3903e868aa9eea80222f7af62fcd7d) to fetch the relevant activity logs and insert fake versions in the review page history.

That hack causes us to load all versions, and is getting in the way of the optimization I'm making here. Since versions have been soft-deleted for more than 3 years now, we are going to remove the hack. I looked at the data with @wagnerand and there are a few more recent logs that show up that need to be investigated first.

diox:
There are 3 kinds of logs that appear because of the hack, that would no longer be displayed in the history once I remove it:
- Thousands of old entries from before we started soft-deleting versions (before 2016-03-03)
- 40 or so logs that appear accidentally because the hack doesn't care about the channel (hard to do with the way the query is set up) and the log actually refers to an unlisted version.
- 3 that are a bit more mysterious because the `ActivityLog` arguments point to a `Version`, but there is no corresponding `VersionLog`. It has to be a bug somewhere but I'm not seeing why. The action for those is either `REQUEST_INFORMATION` or `REVIEWER_REPLY_VERSION`. Maybe the `Version` has somehow been hard-deleted manually by us in the db.

@wagnerand said he wasn't concerned with those versions disappearing from the history, as long as the review page itself for those add-ons would continue working.

diox:
Note: it's not just reviewer tools that would benefit from using `<Addon>.versions. The versions API would as well for instance.

diox:
[...]

PR #12369, opened 2019-09-17
Eliminate a few duplicate queries in reviewer tools

diox:
- Stop fetching all `CommentLog` and `Version`. This was done to display comments from hard-deleted versions, but versions   have been soft-deleted for more than 3 years now.
- Stop re-fetching the add-on and re-calling its default transformer once per version being displayed (7 duplicate queries per version)
- Stop calling the Version transformer twice (2 duplicate queries per version)

Fixes https://github.com/mozilla/addons-server/issues/12344

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/review.html
+++ b/src/olympia/reviewers/templates/reviewers/review.html
@@ -75,7 +75,7 @@
   <div class="results">
     <div class="results-inner">
       <table id="review-files" class="item-history">
-        {% for version in pager.object_list|reverse %}
+        {% for version in pager.object_list %}
         <tr class="listing-header">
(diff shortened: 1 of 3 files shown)


src/olympia/reviewers/templates/reviewers/review.html
```diff
@@ -75,7 +75,7 @@ <h3 id="history">
   <div class="results">
     <div class="results-inner">
       <table id="review-files" class="item-history">
-        {% for version in pager.object_list|reverse %}
+        {% for version in pager.object_list %}
```
diox: `|reverse` is no longer needed now that we directly use the queryset, it's already sorted and paginated in SQL.

Issue #12372, opened 2019-09-18
Fix Userprofile.username charset db inconsistency


eviljeff:
Running migration [1201](https://github.com/mozilla/addons-server/blob/2019.09.19/src/olympia/migrations/1201-users.sql) resulted in the error `stderr: b"ERROR 1062 (23000) at line 2: Duplicate entry 'Austin' for key 'username'\n` seemingly because we are dropping the existing explicit column definition of `CHARACTER SET utf8mb4 COLLATE utf8mb4_bin` and replacing it with the table default.

Attempts to manually fix the data by renaming the "Austin" username to something else, resulted in a different error for the next username, and so on (I abandoned after 3 `UPDATE` to different records).  To unblock the stage deploy I merged https://github.com/mozilla/addons-server/commit/d6cdd59175b649f02e767843f921a882bd9957e8 and cherry-picked to the https://github.com/mozilla/addons-server/tree/2019.09.19-1 tag.

We need to fix the inconsistency and reapply ```ALTER TABLE `users` CHANGE COLUMN `username` `username` VARCHAR (255) NOT NULL;``` to bring the stage&prod database to django norms.

diox:
QA: Let's just test that this hasn't accidentally dropped existing usernames that were not dupes by loading known public collection URLs like https://addons-dev.allizom.org/en-US/firefox/collections/VictorQATester/testcollect/ and making sure they still work.

diox:
Actually, the new migration initially didn't take on stage, it failed with:
`ERROR 1062 (23000) at line 7: Duplicate entry 'anonymous-83f2f3c3aefa9a786cee930e3572eba932' for key 'username'`

The relevant SQL was this:
```sql
CREATE TEMPORARY TABLE `duplicate_usernames`
    SELECT LOWER(`username`) AS `username` FROM `users`
        GROUP BY LOWER(`username`) HAVING COUNT(*) > 1;
UPDATE `users`
    SET `username`=CONCAT('anonymous-', MD5(RAND()), 32)
    WHERE LOWER(`username`) != `username`
    AND LOWER(`username`) IN (SELECT `username` FROM `duplicate_usernames`);
ALTER TABLE `users` CHANGE COLUMN `username` `username` VARCHAR (255) NOT NULL;
```

Before running it, there were ~60k affected users on stage, 3 times that on prod.

[...]

PR #12375, opened 2019-09-18
Anonymize duplicate usernames and fix collation to be case-insensitive

diox:
Fixes https://github.com/mozilla/addons-server/issues/12372

### DIFF
--- a/src/olympia/migrations/1212-unique-lowercase-usernames.sql
+++ b/src/olympia/migrations/1212-unique-lowercase-usernames.sql
@@ -1,5 +1,5 @@
 /* Create a table with all usernames that have duplicates in lowercase. */
-CREATE TABLE `duplicate_usernames`
+CREATE TEMPORARY TABLE `duplicate_usernames`
     SELECT LOWER(`username`) AS `username` FROM `users`
         GROUP BY LOWER(`username`) HAVING COUNT(*) > 1;
 /* Update the username for all users in that table that have a username and lowercase username not being equal. */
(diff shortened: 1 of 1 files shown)


src/olympia/migrations/1212-unique-lowercase-usernames.sql
```diff
@@ -0,0 +1,11 @@
+/* Create a table with all usernames that have duplicates in lowercase. */
+CREATE TABLE `duplicate_usernames`
```
eviljeff: Can we either delete this table at the end of the migration, or file a follow-up to remove it once we're done with whatever we're doing with it?

eviljeff: or is `CREATE TEMPORARY TABLE` a possibility?  (Is the entire migration executed as a single "session")?

diox: I almost did that but I thought we might want to have it around in case we receive some user complaints... though I realize now I'm just storing the username and not the id so I don't know how useful it'd be...

eviljeff: And the ```LOWER(`username`)``` too, so even less useful.  If you want to to keep it around for some reason then issue plz.  Now we've a clean database lets keep it clean 😛

diox: 23584554f80007cae97061c49f26e72dcf7f59c1 makes it a temporary table

PR #12377, opened 2019-09-18
Add statsd.incr() calls in scanners/yara tasks

### LINKED ISSUE #12376: Add statsd counters to track the number of success/failure executions of the security tools
Let's add some more `statsd` calls to track the number of success/failure executions in Grafana and see how the tools perform.

willdurand:
Fixes #12376

### DIFF
--- a/src/olympia/scanners/tests/test_tasks.py
+++ b/src/olympia/scanners/tests/test_tasks.py
@@ -35,8 +35,9 @@ class TestRunScanner(UploadTest, TestCase):
 
     @override_settings(SCANNER_TIMEOUT=123)
     @mock.patch('olympia.scanners.tasks.SCANNERS', MOCK_SCANNERS)
+    @mock.patch('olympia.yara.tasks.statsd.incr')
     @mock.patch('olympia.scanners.tasks.requests.post')
-    def test_run_with_mocks(self, requests_mock):
+    def test_run_with_mocks(self, requests_mock, incr_mock):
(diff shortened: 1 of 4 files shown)


src/olympia/scanners/tests/test_tasks.py
```diff
@@ -62,6 +63,10 @@ def test_run_with_mocks(self, requests_mock):
         assert result.upload == upload
         assert result.scanner == self.FAKE_SCANNER
         assert result.results == scanner_data
+        scanner_name = dict(self.MOCK_SCANNERS).get(self.FAKE_SCANNER)
```
eviljeff: I guess not a question specific to this patch, but if we have to do `dict()` everywhere `SCANNERS` is accessed, why not just define it as a dict in the first place?

willdurand: Can we pass a `dict` to a [Django's `choices` model attribute](https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.Field.choices)?

eviljeff: No, you'd have to do dict.items() to get a list of tuples back.  (Which I realise is just the opposite way around)  You only have to do that in one place vs. converting to a dict in many places though.  A nit for any follow-up patches that touch the scanners/yara code I suppose.

willdurand: yep, I agree it sounds nicer.

Issue #12379, opened 2019-09-13
Validation fails for packages that contains filenames that finish with a "dot".


spoji:
### Describe the problem and steps to reproduce it:

Create a very simple package containing 2 files : 
a manifest.xml
a svg named "www.amazon.com..svg"

Example : [fail.zip](https://github.com/mozilla/addons-linter/files/3610062/fail.zip)
Then upload it to /addon/submit/upload-listed

### What happened?
/addon/submit/upload-listed fails with the following message : 
"our add-on failed validation with 1 error. Sorry, we couldn't load your WebExtension"

If the file "www.amazon.com..svg" is renamed to "www.amazon.com.svg", the upload will not fail.

### Anything else we should know?
This issue has been created after looking at https://github.com/mozilla/addons/issues/1092

willdurand:
It turns out that this is a server issue, in the repack task:

```
[...]

Issue #11861, opened 2019-07-19
Uncaught exception thrown when a zip file member is larger than the limit we set


willdurand:
### Describe the problem and steps to reproduce it:

1. Create a XPI with a large text file (more than 150MB for example)
2. Submit it

### What happened?

So the error is not visible to the end user. There is a validation error page, which is expected but it does not say why the validation has failed.

In the logs, we can see an uncaught exception:

```
TypeError('__str__ returned non-string (type bytes)',)"    
   msg: "XPI parse error"    
   remoteAddressChain: "[REDACTED]"    
   traceback: "Uncaught exception:
  File "/data/olympia/src/olympia/files/utils.py", line 988, in parse_xpi
    xpi_info = Extractor.parse(xpi, minimal=minimal)
  File "/data/olympia/src/olympia/files/utils.py", line 154, in parse
    zip_file = SafeZip(xpi_fobj)
  File "/data/olympia/src/olympia/files/utils.py", line 755, in __init__
    self.is_valid = self.initialize_and_validate()
  File "/data/olympia/src/olympia/files/utils.py", line 775, in initialize_and_validate
    archive_member_validator(self.source, info)
  File "/data/olympia/src/olympia/files/utils.py", line 741, in archive_member_validator
    '%s' % (archive, filename, filesize))
<class 'TypeError'>
TypeError('__str__ returned non-string (type bytes)',)
```

### What did you expect to happen?

I don't know, but the uncaught `TypeError` exception should be handled.

### Anything else we should know?

At first glance, it looks like the log call is the culprit: https://github.com/mozilla/addons-server/blob/17d4886b14c277b238ea7925beb63be52c39e559/src/olympia/files/utils.py#L741-L743

I don't really know what we should be logging here but it has more values than placeholders and the order looks... odd too.

EnTeQuAk:
> I don't really know what we should be logging here but it has more values than placeholders and the order looks... odd too.

There are three placeholders and values that are used to fill them. The problem more likely is that `archive`, `filename` or `filesize` is a bytes type object and the conversion is being wonky…

EnTeQuAk:
Initially I thought that this is a regression of 2de22598bdd14b5bbd7f58d26b9d2fa6448ccdf9 but it doesn't seem likely, at least I don't see changes that may have caused it by looking at `git diff 91c882f549..703cdb9f96`

willdurand:
[...]

PR #12382, opened 2019-09-18
Allow valid dot dot files

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12379
Fixes https://github.com/mozilla/addons-server/issues/11861

---

I created a new function to be able to test it..


willdurand:
mmh wait, the `invalid_webextension.xpi` isn't invalid anymore because it simply contains a file named `a..b.txt`, which would be allowed with this patch.. I need a new invalid webextension.

willdurand:
I updated the `invalid_webextension.xpi` file. In order to do that, I...

1. created a new zip file with a `manifest.json` file and another file named `filee` (with 2 `e`)
2. renamed the zip file from `test.zip` to `test` so that I could edit it in raw mode in vim
3. found the second occurrence of `filee` in the `test` file (binary) and added `/` before this value
4. renamed the `test` file to `test.zip`
5. fixed the zip file with `zip -F test.zip --out test.out`
6. renamed `test.out` to `invalid_webextension.xpi`

These steps allowed me to create an xpi with a file named `/file`.

src/olympia/files/tests/test_utils_.py
```diff
@@ -1189,10 +1189,49 @@ def test_id_to_path(value, expected):
...
+    def test_does_not_raise_when_filename_is_dot_dot_extension(self):
+        utils._validate_archive_member_name_and_size('foo..svg', 123)
+
+    @override_settings(FILE_UNZIP_SIZE_LIMIT=100)
+    def test_raises_when_filesize_is_above_limit(self):
+        with pytest.raises(forms.ValidationError):
+            utils._validate_archive_member_name_and_size(
+                'filename',
+                settings.FILE_UNZIP_SIZE_LIMIT + 100
+            )
```
EnTeQuAk: You could be using `@pytest.mark.parametrize` to simplify this test a bit and make it easier to see the actual scenarios and look through them imo (e.g https://github.com/mozilla/addons-server/blob/af81faedd5aab40c8674c76a7dc8e8670b69e5d5/src/olympia/lib/tests/test_git.py#L1061-L1109)

For using parametrize together with `pytest.raises` you could take a look at http://doc.pytest.org/en/latest/example/parametrize.html#parametrizing-conditional-raising

PR #12404, opened 2019-09-20
Optimize queries made by ActivityLog in reviewer tools & devhub

### LINKED ISSUE #12055: Make fewer queries when displaying activity logs
Part of #11797

A lot of extra queries in reviewer tools (and elsewhere, like devhub) are caused by `ActivityLog`. When an `ActivityLog` is displayed, we load the json containing the representation of the arguments stored when the activity was recorded, and then fetch the corresponding instances from the database:

```python
    @property
    def arguments(self):
        try:
            # d is a structure:
            # ``d = [{'addons.addon':12}, {'addons.addon':1}, ... ]``
            d = json.loads(self._arguments)
        except Exception as e:
            log.debug('unserializing data from addon_log failed: %s' % self.id)
            log.debug(e)
            return None

        objs = []
        for item in d:
            # item has only one element.
            model_name, pk = list(item.items())[0]
            if model_name in ('str', 'int', 'null'):
                objs.append(pk)
            else:
                # Cope with renames of key models:
                if model_name == 'reviews.review':
                    model_name = 'ratings.rating'
                (app_label, model_name) = model_name.split('.')
                model = apps.get_model(app_label, model_name)
                # Cope with soft deleted models and unlisted addons.
                objs.extend(model.get_unfiltered_manager().filter(pk=pk))
        return objs
```

If there was a database relation between the object(s) the activity log is about and the activity log itself, we could try to force the object(s) to be preloaded on each activity instance, something like that. But it's not the case here, so we'll have to be more creative.

diox:
- Cache `ActivityLog.arguments` on the instance to avoid making its queries multiple times when the property is called more than once
- Group all queries made by `ActivityLog.arguments` using a transformer by default
- Avoid loading the version (and therefore its files & applications versions) when figuring out the url for an `Addon`
- Prefetch the user when fetching `ActivityLog`
- Tweak review page and `Version.all_activity`/`activity_transformer` to benefit from those changes

Number of queries made by reviewer tools will differ a lot depending on the situation, but on my local environment, after spinning up lots of various activity:

Queries count in reviewlog page before: > 4500 queries
Queries count in reviewlog page after:    ~ 15 queries
Queries count in busy review page before: > 1000 queries
Queries count in busy review page after:  ~  100 queries

Still a lot to improve on the review page, but definitive improvements across the board in reviewer tools.

Fixes #12055

### DIFF
--- a/src/olympia/amo/models.py
+++ b/src/olympia/amo/models.py
@@ -75,8 +75,10 @@ class BaseQuerySet(models.QuerySet):
         """Remove all transforms except translations."""
         from olympia.translations import transformer
         # Add an extra select so these are cached separately.
-        return (self.no_transforms().extra(select={'_only_trans': 1})
-                .transform(transformer.get_trans))
+        qs = self.no_transforms()
+        if hasattr(self.model._meta, 'translated_fields'):
+            qs = qs.transform(transformer.get_trans)
+        return qs
 
 
 class RawQuerySet(models.query.RawQuerySet):
--- a/src/olympia/activity/models.py
+++ b/src/olympia/activity/models.py
@@ -2,6 +2,7 @@ import json
 import string
 import uuid
 
+from collections import defaultdict
 from copy import copy
 from datetime import datetime
 
@@ -10,6 +11,7 @@ from django.conf import settings
 from django.db import models
 from django.urls import reverse
 from django.utils import timezone
+from django.utils.functional import cached_property
 from django.utils.translation import ugettext
 
 import jinja2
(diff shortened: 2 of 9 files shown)


src/olympia/amo/models.py
```diff
@@ -75,8 +75,10 @@ def only_translations(self):
         """Remove all transforms except translations."""
         from olympia.translations import transformer
         # Add an extra select so these are cached separately.
-        return (self.no_transforms().extra(select={'_only_trans': 1})
```
diox: That `extra()` is a relic from cache-machine days, no longer needed.

PR #12406, opened 2019-09-20
delete any update_counts records that won't be able to satisfy new fk constraint

eviljeff:
pre-empting migration 1200 failing


eviljeff:
Not that this is going to do anything other than note down a fix for reference.  -dev and stage have already gone past this migration and we skipped past it on prod.  Once mozilla-services/cloudops-deployment#3412 is merged schematic won't run on deploy either .

PR #12422, opened 2019-09-24
Split repack and validation tasks

willdurand:


src/olympia/devhub/tests/test_views.py
```diff
@@ -1331,21 +1331,14 @@ def test_no_servererror_on_missing_version(self):
...
-        # message for this kind of situation).
         assert not run_addons_linter_mock.called
         upload = FileUpload.objects.get()
         response = self.client.get(
             reverse('devhub.upload_detail', args=[upload.uuid.hex, 'json']))
         data = json.loads(force_text(response.content))
         message = [(m['message'], m.get('fatal', False))
                    for m in data['validation']['messages']]
-        # We do raise a specific error message explaining that the archive is
-        # not valid instead of a generic exception.
```
willdurand: @diox I reverted changes you made here when you combined the repack and linter tasks

src/olympia/devhub/tests/test_utils.py
```diff
@@ -68,11 +70,12 @@ def check_upload(self, file_upload, listed=True):
...
                    else amo.RELEASE_CHANNEL_UNLISTED)
 
-        # Make sure we run the correct validation task for the upload and we
-        # set up an error handler.
+        # `Validator.validate_upload()` returns a chain with these tasks:
+        self.repack_fileupload.called_once_with(file_upload.pk)
         self.validate_upload.assert_called_once_with(
             file_upload.pk, channel=channel)
-        assert self.validate_upload.return_value.on_error.called
+        # TODO: Make sure we set up an error handler on the validation task:
```
willdurand: @diox I am not sure to know how to re-add this.. Any idea?

src/olympia/devhub/tasks.py
```diff
@@ -180,25 +179,26 @@ def wrapper(task, id_or_path, *args, **kwargs):
 
 
 @validation_task
-@use_primary_db
```
willdurand: Is this still needed?

diox: Yes: it needs to use the primary db when fetching the `FileUpload`, which is created before in the view. Otherwise, if there is a slight delay between primary and replica and the task gets fired too early, it might not be able to fetch the `FileUpload`.

One could argue that it needs to be moved to `@validation_task` instead, though.

willdurand: I am wondering if we need this on the scanner tasks too.

diox: Absolutely. This hasn't been a problem since they weren't called as real tasks, but as soon as they are they'll need it, so might as well include it now.

src/olympia/devhub/tests/test_views.py
diox: mmm you shouldn't need to change this test if everything still works. We *do* want the specific message instead of the generic one. The comment above should also still apply: if the xpi is not a valid zip the chain should abort and the linter should never be called.

willdurand: mmh, interesting. I wrapped the repack task with `validation_task` (otherwise, everything gets broken...) but it is likely why we do not get the specific error. It likely swallows the exception, no?

diox: But `validation_task` *is* the one that returns that `Invalid or corrupt add-on file` message, if `BadZipfile` is raised by the task it's wrapping. And this should be the right exception to catch, since it was working before ? Something is not right.

willdurand: Yeah, I don't get it either to be honest.. I'll investigate

diox: It might be precisely the issue we wanted to avoid by upgrading Celery - it might also be only in tests, since we're running them in eager mode and that messes everything up... do try on your actual env

willdurand: I think I get it:

I added `@validation_task` to the `repack` task, so that we can use its error handling and return the `Invalid or corrupt add-on file` message. The problem is that it means the exception is handled and we do not call `on_error()` because there is not need (the exception has been handled). Instead, `repack` returns the `results` but because `validate_upload` is declared as immutable, the `results` from `repack` are lost.

That is why `validate_upload` is fired.

If we do not use `@validation_task` on the `repack` task, we cannot return a specific message. I forgot about it, but that's what I assumed last night... and that's why I changed this test case.

If we want to return a specific message, I suppose we need to move the logic from `@validation_task` to the `Validator.error_handler` method, assuming raising in a chain calls this handler (which was not the case in Celery < 4.2 if I understood correctly).

diox: Yeah, I think that makes sense. We do want the specific error message, so now that we have a chain the logic for that must be in the error handler.

willdurand:
@diox I updated this patch with one of the approaches we discussed. It seems to work without having too many things to change. I was able to "centralize" the results handling in the `validation_task` decorator itself, so that does look less bad than what I thought.

willdurand:
mmh, I spotted something I did not like. I was breaking the signature of the validation tasks. I don't think we want that.

src/olympia/files/tests/test_tasks.py
```diff
@@ -52,7 +52,7 @@ def test_repacking_xpi_files(self):
         # that structure is restored once the file has been moved.
         upload = self.get_upload('unicode-filenames.xpi')
         original_hash = upload.hash
-        repack_fileupload(upload.pk)
+        repack_fileupload(None, upload.pk)
```
willdurand: We could probably add a test case to make sure the task returns what we give to it.

src/olympia/devhub/tasks.py
```diff
@@ -140,22 +139,41 @@ def create_version_for_upload(addon, upload, channel):
...
+    validation tasks. Should only be called directly by Validator."""
+    results = deepcopy(amo.VALIDATOR_SKELETON_RESULTS)
+    return results
+
+
 def validation_task(fn):
-    """Wrap a validation task so that it runs with the correct flags, then
-    parse and annotate the results before returning."""
+    """Wrap a validation task so that it runs with the correct flags and
+    handles errors.
```
diox: Mention the signature the functions must have, and where `results` is coming from (it's not obvious if you haven't read `create_initial_validation_results()` before)

src/olympia/devhub/tasks.py
```diff
@@ -139,22 +132,56 @@ def create_version_for_upload(addon, upload, channel):
         add_dynamic_theme_tag(version)
 
 
+@task
+def create_initial_validation_results():
```
diox: Need to update `CELERY_TASK_ROUTES` in `settings_base.py` to make sure this task goes to the `devhub` queue, otherwise the chain might not work properly. (Double-check, I think it's also missing `run_wat`)

src/olympia/devhub/tasks.py
```diff
@@ -139,22 +132,56 @@ def create_version_for_upload(addon, upload, channel):
...
+        # return.
         task.ignore_result = True
         try:
-            data = fn(id_or_path, **kwargs)
-            results = json.loads(force_text(data))
-            return results
+            # All validation tasks should receive `results`.
+            if not results:
+                raise Exception('Unexpected call to a validation task without '
+                                '`results`')
```
diox: Maybe `assert` instead, and/or write a test for this.

src/olympia/devhub/tasks.py
```diff
@@ -139,22 +132,56 @@ def create_version_for_upload(addon, upload, channel):
...
+
+            if results['errors'] > 0:
```
diox: Is there a test for this?

src/olympia/devhub/tasks.py
willdurand: ah yes, thanks

src/olympia/devhub/tasks.py
willdurand: added a test case

src/olympia/devhub/tasks.py
willdurand: now there is one

Issue #12417, opened 2019-09-23
drop 'skip_register_redirect' in fxa configs


eviljeff:
to quote myself from #10487 
> after the authentication has taken place the only difference is whether we redirect users to the edit profile page... something I'm not sure we need anyway because it was originally implemented when addons-frontend was mobile only and we'd not gotten around to implementing the edit page. 🤔 

(splitting this out as an issue so we're clear on the feature removal)

The logic currently is:
- `default` fxa config (used everywhere on django served pages, and possibly frontend*) redirects for a new user everywhere (locally, addons-dev, addons-stage, addons-prod)
- `amo` fxa config (used by frontend*) redirects locally & addons-dev, but doesn't on addons-stage & addons-prod
- `local` fxa config (used for frontend local dev) redirects locally & addons-dev & addons-stage (it isn't available on addons-prod)
- `code-manager` fxa config (used for code manager app) copies the `amo` logic.

IMO, we should either stop redirecting any new user to the profile registration page, or redirect all new users.  The reason it was implemented (not having a frontend edit profile page) is long gone and it adds complexity on the server side (and means as many as 4 times as many fxa client key pairs need to be generated per environment)

@jvillalobos to decide if we want to redirect users who are first time AMO users to the account profile settings page or not.

*frontend uses `amo` on some environments and possibly not others but it's an ops env setting so a little opaque.

eviljeff:
the internal logic is roughly:
- -server gets a `to` url passed along from frontend
- if new_user and not skip_register_redirect then ignore `to` url and edirect to /users/edit
- otherwise, redirect to `to`iss

If we want to force users and/or developers to set their display_name then we probably want to redirect all new users and prevent frontend from opting out some.

jvillalobos:
In which cases would there be a `to` set? Is that initially set by AMO or does it come from somewhere else?

eviljeff:
> In which cases would there be a to set? Is that initially set by AMO or does it come from somewhere else?

It's the page the user was on when they tried to login/register

jvillalobos:
This would make the most sense to me:

- Send all users to the profile edit page.
- Have the button say `Create My Profile` instead of `Update My Profile`.
- If there's a `to` in the URL, send them there after the profile is created.

eviljeff:
> Have the button say Create My Profile instead of Update My Profile.
> If there's a to in the URL, send them there after the profile is created.

[...]

PR #12423, opened 2019-09-24
fxa auth flow changes

### LINKED ISSUE #10487: The authentication dance uses a v3 endpoint (even when v4 is used at the beginning)
When trying to log in a user (for addons-code-manager but also addons-frontend), after the FxA page, a `v3` API endpoint is used instead of `v4`:

screen shot 2019-01-22 at 17 33 55

~~I don't really know why component creates this URL though, FxA maybe?~~

The URL comes from the response to an XHR call to https://oauth-stable.dev.lcip.org/v1/authorization:

```
{"code":"xxxx","state":"yyy8261:Lw","redirect":"http://localhost:3000/api/v3/accounts/authenticate/?config=local&state=yyy%3ALw&code=xxx"}
```

eviljeff:
this would fix #10487 - and #12417 if the outcome of that issue is to always redirect for new users (if it's something else the pr would need changes).

It drops the most of the configs because undefined config names will fall back to `default` - without the `skip_register_redirect` differences they will be the same, apart from on local dev which has some extra functionality, and `config=local` which always redirects to a localhost url.

I generated some new fxa clients (id/secret pairs) for local dev; dev, stage, and prod will need ones generating by ops and adding to the env files.  I left the `api/v3/accounts/authenticate` in place for the meantime so localhost callbacks from fxa on dev & stage should work until there is a new deploy that updates the env.

### DIFF
--- a/docs/topics/api/auth_internal.rst
+++ b/docs/topics/api/auth_internal.rst
@@ -22,7 +22,7 @@ Fetching the token
 A fresh token, valid for 30 days, is automatically generated and added to the
 responses of the following endpoint:
 
-    * ``/api/v4/accounts/authenticate/``
+    * ``/api/auth/authenticate-callback/``
 
 The token is available in two forms:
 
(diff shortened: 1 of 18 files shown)


eviljeff:
I'm going to refactor this to move the code for #12417 and the general changes to a different pr, leaving this one to focus on just the api endpoint changes.

eviljeff:
closing for now, until the pr fixing #12417 is merged.

PR #12438, opened 2019-09-26
Reduce number of queries made by addon versions API and ratings API

### LINKED ISSUE #12396: Remove duplicate addon queries in Versions API
In https://github.com/mozilla/addons-server/issues/12344#issuecomment-531979930 I wrote:
> Note: it's not just reviewer tools that would benefit from using `<Addon>.versions`. The versions API would as well for instance.

So let's do that - currently the versions API repeats the query to load add-on + its transformer for each version it returns.

diox:
Fixes #12396

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -362,8 +362,8 @@ class TestAddonViewSetDetail(AddonAndVersionViewSetDetailMixin, TestCase):
             guid=generate_addon_guid(), name=u'My Addôn', slug='my-addon')
         self._set_tested_url(self.addon.pk)
 
-    def _test_url(self):
-        response = self.client.get(self.url)
+    def _test_url(self, **kwargs):
+        response = self.client.get(self.url, data=kwargs)
         assert response.status_code == 200
         result = json.loads(force_text(response.content))
         assert result['id'] == self.addon.pk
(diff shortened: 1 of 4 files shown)


src/olympia/ratings/tests/test_views.py
```diff
@@ -185,22 +185,18 @@ def test_list_addon_queries(self):
 
         assert Rating.unfiltered.count() == 3
 
-        with self.assertNumQueries(5):
-            # 5 queries:
+        with self.assertNumQueries(7):
+            # 7 queries:
```
EnTeQuAk: **not** approved :hammer: :see_no_evil:

src/olympia/addons/tests/test_views.py
```diff
@@ -377,6 +377,26 @@ def _set_tested_url(self, param):
...
+            # - 1 for applications versions of that version
+            # - 1 for files of that version
+            # - 1 for authors
+            # - 1 for previews
+            # - 1 for license
+            # - 1 for translations of the license
+            # - 1 for discovery item (is_recommended)
+            # - 1 for featured collection presence (is_featured)
+            # - 1 for tags
+            self._test_url(lang='en-US')
```
EnTeQuAk: Is this actually a valid test-case for the actual issue? Unless I'm mistaken we were hitting "duplicated" queries that multiplied by the amount of add-ons shown right? We aren't creating any additional add-ons here to verify that that's no longer the case.

EnTeQuAk: Also, can we easily `select_related` any of the other relations without hurting this or other APIs?

diox: That one specifically is not for the valid issue, it's the add-on detail API. I made it to make sure I didn't regress anything while fixing the versions API.

There are very few single relations in there, so few opportunities for a `select_related()`. I think the only case where that'd work is discovery item - there is also the current version, but we'd skip the Version `transformer` so I don't know what that would do. In any case that'd be a different issue.

src/olympia/ratings/tests/test_views.py
diox: Actual count without the mocks I removed before this patch is like 13 or something, since we were calling all addon transformers for not good reason :scream:

EnTeQuAk: Hehe

PR #12443, opened 2019-09-27
fixed debugtool middleware be after GZip middleware

### LINKED ISSUE #11819: Warnings related to django middleware in local dev
Running `make initialize_docker`, we can see the following warnings in the logs:

```
WARNINGS:
?: (debug_toolbar.W003) debug_toolbar.middleware.DebugToolbarMiddleware occurs before django.middleware.gzip.GZipMiddleware in MIDDLEWARE.
	HINT: Move debug_toolbar.middleware.DebugToolbarMiddleware to after django.middleware.gzip.GZipMiddleware in MIDDLEWARE.
```

junngo:
Fixes #11819 

If the DebugToolbar be before the GZip, appeared the WARNINGS on local dev. I revised that debug tool middleware sequence be after GZip middleware.

**Before**
debug_toolbar_warnings

**After**
Not see the WARNINGS

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -19,13 +19,15 @@ INSTALLED_APPS += (
     'debug_toolbar',
 )
 
-# If the DebugToolbar be before the GZip in MIDDLEWARE, appeared the WARNINGS.
+# django-debug-doolbar middleware needs to be inserted as high as possible
+# but after GZip middleware
 def insert_debug_toolbar_middleware(middlewares):
     ret_middleware = list(middlewares)
 
     for i, middleware in enumerate(ret_middleware):
         if 'GZipMiddleware' in middleware:
-            ret_middleware.insert(i+1, 'debug_toolbar.middleware.DebugToolbarMiddleware')
+            ret_middleware.insert(
+                i + 1, 'debug_toolbar.middleware.DebugToolbarMiddleware')
             break
 
     return tuple(ret_middleware)
--- a/settings.py
+++ b/settings.py
@@ -19,7 +19,18 @@ INSTALLED_APPS += (
     'debug_toolbar',
 )
 
-MIDDLEWARE = ('debug_toolbar.middleware.DebugToolbarMiddleware',) + MIDDLEWARE
+# If the DebugToolbar be before the GZip in MIDDLEWARE, appeared the WARNINGS.
+def insert_debug_toolbar_middleware(middlewares):
+    ret_middleware = list(middlewares)
+
+    for i, middleware in enumerate(ret_middleware):
+        if 'GZipMiddleware' in middleware:
+            ret_middleware.insert(i+1, 'debug_toolbar.middleware.DebugToolbarMiddleware')
+            break
+
+    return tuple(ret_middleware)
+
+MIDDLEWARE = insert_debug_toolbar_middleware(MIDDLEWARE)
 
 DEBUG_TOOLBAR_CONFIG = {
     # Enable django-debug-toolbar locally, if DEBUG is True.


settings.py
```diff
@@ -19,7 +19,17 @@
     'debug_toolbar',
 )
 
-MIDDLEWARE = ('debug_toolbar.middleware.DebugToolbarMiddleware',) + MIDDLEWARE
+# If the DebugToolbar be before the GZip, appeared the WARNINGS.
+def insert_debug_toolbar_middleware(middlewares):
+    ret_middleware = list(middlewares)
+
+    for i, middleware in enumerate(ret_middleware):
+        if 'olympia.api.middleware.GZipMiddlewareForAPIOnly' in middleware:
```
diox: I mentioned in [my comment](https://github.com/mozilla/addons-server/issues/11819#issuecomment-515924406) to re-use the logic Django-Debug-Toolbar already has for this. It's more code, but it matters because otherwise we're hardcoding the name of the middleware, which would break if we ever change it - in other words it lacks flexibility.

junngo:
I re-wrote code, re-using the logic from django-debug-toolbar.
I talk about considerations and my thought.

1. Using django.setup() function
When I re-using the logic, It happens error:
`django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.`
This error happens. When settings files are imported, Some Apps arent' loaded yet. So I had to use this function(`django.setup()`).

2. Reason for making "Insert debug-toolbar function" instead of "Return the position of the GZipMiddleware function"
MIDDLEWARE is tuple. So If I do insert debug middleware on the place outside of function, It looks complicated. Especially When someone add middleware after.

If I wrong or there's a better way, let me know. So, I gonna modify this code again.

junngo:
Test failure happen because I call 'django.setup()' function:
`django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty.`

I could try to add code(e.g)`os.environ.setdefault(...)`) to solve.
But It looks like need to review generally as to whether the direction is right.

diox:
The fact that you have to call `django.setup()` is an issue - it's too soon for this to happen, we're likely to encounter other problems doing that. Have you traced what call cause django to complain that Apps aren't loaded yet without the `django.setup()` bit ?

junngo:
Sure, I traced related middleware.
In the absence of the `django.setup()`, Middleware with issue is as follows.

- olympia.amo.middleware.NonAtomicRequestsForSafeHttpMethodsMiddleware
- olympia.amo.middleware.SetRemoteAddrFromForwardedFor
- olympia.amo.middleware.LocaleAndAppURLMiddleware
- olympia.amo.middleware.RemoveSlashMiddleware
- olympia.amo.middleware.CommonMiddleware
- olympia.amo.middleware.NoVarySessionMiddleware
- olympia.amo.middleware.AuthenticationMiddlewareWithoutAPI
- olympia.amo.middleware.ReadOnlyMiddleware
- olympia.amo.middleware.ScrubRequestOnException
- olympia.amo.middleware.RequestIdMiddleware

diox:
The question is, what causes `django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.` if you don't add `django.setup()`

junngo:
(Sorry, I closed PR by mistake.)
Middlewares with issue aren't loaded yet, When I use the `import_string()` to import this middleware in settings.py.
The reason for using the `import_string()` function is to convert string to class object.

diox:
Stacktrace:
```
  File "/code/src/olympia/amo/middleware.py", line 9, in <module>
    from django.contrib.auth.middleware import AuthenticationMiddleware
  File "/usr/local/lib/python3.6/site-packages/django/contrib/auth/middleware.py", line 4, in <module>
    from django.contrib.auth.backends import RemoteUserBackend
  File "/usr/local/lib/python3.6/site-packages/django/contrib/auth/backends.py", line 5, in <module>
    from django.contrib.auth.models import Permission
  File "/usr/local/lib/python3.6/site-packages/django/contrib/auth/models.py", line 2, in <module>
    from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
  File "/usr/local/lib/python3.6/site-packages/django/contrib/auth/base_user.py", line 47, in <module>
    class AbstractBaseUser(models.Model):
  File "/usr/local/lib/python3.6/site-packages/django/db/models/base.py", line 103, in __new__
    app_config = apps.get_containing_app_config(module)
  File "/usr/local/lib/python3.6/site-packages/django/apps/registry.py", line 252, in get_containing_app_config
    self.check_apps_ready()
  File "/usr/local/lib/python3.6/site-packages/django/apps/registry.py", line 135, in check_apps_ready
    raise AppRegistryNotReady("Apps aren't loaded yet.")
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.

```

Loading modules in settings is always painful since you run the risk to be importing models too early, and that's what's happening here... 

Let's go back to hardcoding the string, then :(

junngo:
I went to back to hardcoding the string.
If you have any changes after review, please tell me again :)

diox:
Thanks. I improved the comment and fixed some formatting, I'll merge as soon as tests pass.

junngo:
Okay. Thank you all the time for your review :)

Issue #12448, opened 2019-09-27
Only set a version to YaraResult/ScannersResult when feature is enabled


willdurand:
We currently always try to set a version to a `YaraResult`/`ScannersResult`, even when the features are not enabled. It's not a problem _per se_ but it is not needed and results in unnecessary error entries in the logs.

**For QA:** please make sure that `YaraResult` or `ScannersResult` can be linked to a version when the version gets created.

willdurand:
I believe we can also restrict to web extensions too.

vcarciu:
Verified as fixed with multiple webextensions. Yara and scanners results are correctly linked to a vers when created.

Issue #12446, opened 2019-09-27
`log.exception()` calls are not sent to Sentry


willdurand:
We use `log.exception()` to log errors in this app and we can see them in Stackdriver. Sentry is supposed to receive these errors too, but that does not seem to be the case.

For example, this part of the code logs exceptions and we can see them in Stackdriver but I was not able to find them in Sentry (I tried to find occurrences of "YaraResult" or "ScannersResult" in any of the environments): https://github.com/mozilla/addons-server/blob/998615a9892831aec3a9b94919fd672a4d5a6417/src/olympia/versions/models.py#L294-L308

wagnerand:
Sounds like something we want to fix.

diox:
Related, we have kind of the opposite problem in tasks now (which wasn't the case before we tweaked the logging config, I believe): https://github.com/mozilla/addons-server/issues/14355

I suspect that, to make it work outside of tasks, we need to do something specific. Maybe we should investigate switching to the sentry sdk python client to replace raven first, to be on a more supported configuration.

eviljeff:
I fixed this in #14640

PR #12449, opened 2019-09-27
Add guards in Version model

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12448

---

If Sentry was receiving the `log.exception()` calls (#12446), we would have noticed this before.

### DIFF
--- a/src/olympia/versions/models.py
+++ b/src/olympia/versions/models.py
@@ -291,26 +291,27 @@ class Version(OnChangeMixin, ModelBase):
 
         version_uploaded.send(instance=version, sender=Version)
 
-        if waffle.switch_is_active('enable-yara'):
-            try:
-                yara_result = YaraResult.objects.get(upload_id=upload.id)
-                yara_result.version = version
-                yara_result.save()
-            except YaraResult.DoesNotExist:
-                log.exception('Could not find a YaraResult for FileUpload %s',
-                            upload.id)
+        if version.is_webextension:
+            if waffle.switch_is_active('enable-yara'):
+                try:
+                    yara_result = YaraResult.objects.get(upload_id=upload.id)
+                    yara_result.version = version
+                    yara_result.save()
+                except YaraResult.DoesNotExist:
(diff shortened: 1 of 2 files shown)


src/olympia/versions/models.py
```diff
@@ -291,21 +291,27 @@ def from_upload(cls, upload, addon, selected_apps, channel,
...
+
+            if (
+                    waffle.switch_is_active('enable-customs') or
+                    waffle.switch_is_active('enable-wat')
+            ):
+                try:
+                    ScannersResult.objects.filter(upload_id=upload.id).update(
+                        version=version
+                    )
+                except ScannersResult.DoesNotExist:
```
diox: Had not noticed before, but this will never be reached. This is only thrown by django when you `.get()` a single instance, otherwise it just happily returns an empty list (or, in the case of your `.update()`, runs an `UPDATE` that doesn't do anything)

willdurand: ha, that's cool. thanks!

willdurand: removed that part

src/olympia/versions/models.py
```diff
@@ -291,21 +291,22 @@ def from_upload(cls, upload, addon, selected_apps, channel,
...
+                    yara_result = YaraResult.objects.get(upload_id=upload.id)
+                    yara_result.version = version
+                    yara_result.save()
+                except YaraResult.DoesNotExist:
+                    log.exception('Could not find a YaraResult for FileUpload '
+                                  '%s', upload.id)
+
+            if (
+                    waffle.switch_is_active('enable-customs') or
+                    waffle.switch_is_active('enable-wat')
```
eviljeff: unnecessary/unusual indentation?

willdurand: How should this be formatted? (it'll be so much better when we'll have Black :p)

eviljeff: meh, I don't care actually.  Nothing to see here.

PR #12475, opened 2019-09-30
Remove .jar support

### LINKED ISSUE #12435: Remove ".jar" support
This came up during an IRC discussion. We probably don't need to support `.jar` files anymore. We should remove this extension from `VALID_ADDON_FILE_EXTENSIONS` and remove potentially related code.

willdurand:
Fixes #12435

---

Not sure whether I should remove more code... There is one part that I left untouched because many test cases failed: https://github.com/mozilla/addons-server/blob/2206b3f27dac4d4d819ca53529731fb7b2e84607/src/olympia/files/utils.py#L797-L803


eviljeff:
That code is supporting the inner zip files of really, really, old xul extensions I was talking about in the issue 😄 
We should drop that too - but split it into a follow-up if it's easier.  It's quite likely we have a bunch of ancient fixtures that use it.

willdurand:
> That code is supporting the inner zip files of really, really, old xul extensions I was talking about in the issue 😄

ah ok, so we're not using this anymore? I was confused because the 6 test cases use a "langpack" file: https://github.com/mozilla/addons-server/blob/d7ae3e57970e551dbcdd643b8037d511823a90fe/src/olympia/files/tests/test_models.py#L1436-L1472

eviljeff:
Oh, I forgot xul langpacks used it until much more recently.  Afaik that's all dead code since webextension based langpacks but more reason to tinker with it in a separate issue/pr.  Code quality clean-up can be a rabbit-hole on addons-server.

willdurand:
ok, I filed https://github.com/mozilla/addons-server/issues/12478

PR #12389, opened 2019-09-19
remove everything akismet

eviljeff:
fixes #12359 
There goes 2100 lines of code I wrote a year ago!

PR #12487, opened 2019-10-01
drop akismet model migration

eviljeff:
follow-up for #12389 that adds the migration to drop the models.

But, we've got a problem - as the app has been removed from `settings.INSTALLED_APPS` then django doesn't see it any more, so doesn't know the model has gone.  Options as a I see it:
- don't remove the app from settings (and the models.py file?); then delete app from settings, and the files/folders themselves in _another_ follow-up.  Cons: 3 step process over 3 weeks and we loose the migration history at the end.
- move the migration files related to the now deleted app to another app (e.g. `amo`) and change the naming as necessary so it's still valid.  Cons: the database table references the original app so django wouldn't know the previous migrations had already been executed (might try to execute them again).
- something else?


diox:
Would [squashing the migrations](https://docs.djangoproject.com/en/2.2/topics/migrations/#migration-squashing) into a different place (the `amo` app as you suggest) work ?

eviljeff:
squashing still seems to require the original migrations be there while some deploys are mid way through (i.e. they've only executed 0001 to add the table, but not 0002 to drop it).  So, if we wanted to keep some sort of record of the migrations existing, we could squash and stick it somewhere else - and it's possible in some circumstances we might need to if other migrations depend on them - but it still ends up being a multi-week process and the app and existing migrations need to stay for the first week.  

Dropping entire apps is kinda rare though, so I don't think we'll come up against it often.

diox:
> So, if we wanted to keep some sort of record of the migrations existing, we could squash and stick it somewhere else - and it's possible in some circumstances we might need to if other migrations depend on them - but it still ends up being a multi-week process and the app and existing migrations need to stay for the first week.

We kinda need to if we want to be nice to local envs - otherwise if you miss the window when the migration is still here you end up with a broken local env...

Issue #12489, opened 2019-10-01
Stop creating a new chain of tasks in devhub.tasks.validate()


diox:
As part of https://github.com/mozilla/addons-server/issues/12415 @willdurand and I investigated issues with Celery error handling and realized that in `devhub.tasks.validate()`, we create a new chain on top of the existing one. This is not great because the new chain doesn't inherit options passed to the initial chain, and make things harder to debug in general (gotta go deeper to inspect the tasks, etc).

Instead, `Validator()` should probably have some method or argument to add something to the task, so that validate()` can just call `apply()` or `delay()` on it.

Note that ideally, we'd remove this wrapping entirely, removing the functionality to trigger the validation tasks synchronously. Might be worth investigating if the bit that does that in `json_file_validation()` is still useful: if it's not, we can greatly simplify the whole thing. **Edit:** covered separately by https://github.com/mozilla/addons-server/issues/12488

PR #12491, opened 2019-10-01
Stop creating a new chain of tasks in devhub.tasks.validate()

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12489
~~⚠️ depends on #12422~~

---

~~Only the last commit is relevant.~~ rebased so there is only one commit now :)


willdurand:
rebased and ready for a review

PR #12493, opened 2019-10-01
Remove ability to force revalidation of a file in devhub

### LINKED ISSUE #12488: Remove the ability to re-validate a file
It seems possible to revalidate a file in DevHub. I am not sure who is able to do that but is it actually needed?

This makes our validation logic quite complex, so if this is not useful/used anymore, we could probably remove it. WDYT?

diox:
We haven't had issues regarding `File` with no `FileValidation` attached in the past 3 years, everything should now be wrapped in transaction(s).

Fixes #12488


src/olympia/devhub/tests/test_views_validation.py
```diff
@@ -404,166 +396,3 @@ def test_upload_addon_version(self):
...
-        })
-        response = self.client.post(
-            reverse('devhub.json_file_validation',
-                    args=[self.addon.slug, self.file.id]), follow=True)
-        assert response.status_code == 200
-        data = json.loads(response.content)
-        doc = pq(data['validation']['messages'][0]['description'][0])
-        assert doc('a').text() == 'https://bugzilla.mozilla.org/'
-
-    def test_opensearch_validation(self):
```
willdurand: Should we keep this test case even if we are going to remove opensearch support soon?

diox: Given that we are removing it soon and it was in the wrong place I wasn't really motivated to keep it...

willdurand:
ugh, sorry for the conflict :(

diox:
It's fine I expected it :)

PR #12496, opened 2019-10-01
Add a chord() to the validation chain()

### LINKED ISSUE #11984: Create a Celery chords for tasks between `repack` and `handle_upload_validation_result`

willdurand:
Fixes #11984  
~~⚠️ depends on #12491~~

---

This patch adds a Celery `chord` and a task that forwards the linter results to the chord's callback.

### DIFF
--- a/src/olympia/devhub/tasks.py
+++ b/src/olympia/devhub/tasks.py
@@ -279,7 +279,7 @@ def forward_linter_results(results, upload_pk):
     linter results to `handle_upload_validation_result()` (the callback of the
     chord).
     """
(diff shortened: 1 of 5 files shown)


willdurand:
This is ready although circle-ci complains about something... I looked at other patches and it complains for the same reasons so 🤷‍♂

Issue #12216, opened 2019-08-29
Add `run_yara` task to the "chord" of tools that run after the linter


willdurand:
Once https://github.com/mozilla/addons-server/issues/11984 has landed, we should move the `run_yara` function call from the `handle_upload_validation_result` task to the Celery chord that should run after the linter and before `handle_upload_validation_result`.

FTR: We added the `run_yara` task in https://github.com/mozilla/addons-server/issues/12006.

PR #12507, opened 2019-10-02
Use `run_yara()` as a Celery task in the validation chain/chord

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12216
⚠️ depends on #12496 

---

This patch moves the `run_yara` task to the validation chain (in the chord for the scanners).


willdurand:
rebased!

src/olympia/devhub/utils.py
```diff
@@ -256,6 +259,9 @@ def __init__(self, file_, addon=None, listed=None, final_task=None):
                 tasks.forward_linter_results.s(file_.pk),
             )
 
+            if waffle.switch_is_active('enable-yara'):
+                tasks_in_parallel = (*tasks_in_parallel, run_yara.s(file_.pk))
```
diox: Nit: any reason for `tasks_in_parallel` to be a `tuple` and not a `list` we'd just `append()` task signatures to ?

src/olympia/yara/tasks.py
```diff
@@ -5,17 +5,18 @@
...
+
 from .models import YaraResult
 
 log = olympia.core.logger.getLogger('z.yara.task')
 
 
-@task
-def run_yara(upload_pk):
+@validation_task
+def run_yara(results, upload_pk):
```
diox: Since the docstring mentions storing the `results` and it now, as a `validation_task` receives a `results` argument (and returns it) it may start to be a little confusing... I don't have good suggestions though, just something to keep in mind maybe.

src/olympia/devhub/tasks.py
```diff
@@ -301,17 +300,6 @@ def handle_upload_validation_result(all_results, upload_pk, channel,
 
     upload = FileUpload.objects.get(pk=upload_pk)
 
-    if waffle.switch_is_active('enable-yara') and results['errors'] == 0:
```
diox: Didn't you lose the `results['errors'] == 0` bit ? The `chord()` will still be run when there are errors, right ? And we don't want that...

src/olympia/devhub/utils.py
willdurand: No, it is mostly because the celery example you started last week used a tuple. I think both are valid.

src/olympia/yara/tasks.py
willdurand: Ah, good catch, thanks. I'll revisit the docstring.

src/olympia/devhub/tasks.py
willdurand: That's a good question. Since `run_yara` is a validation task now, it should not run because `@validation_task` already checks `results['errors']`, no? https://github.com/mozilla/addons-server/blob/4389866635dd83464fd5ad270f080338ea1c645e/src/olympia/devhub/tasks.py#L178-L179

The chord will indeed start but will also early return and the `run_yara` code won't run. Does that sound okay?

If we're not doing this, I don't know how we could do that in the `Validator` since this needs to be decided based on the chain results.

diox: Ah yeah forgot we had this check in `@validation_task` now. That's ok. Maybe worth adding an integration test specifically checking that this particular task is aborting early (I know we have the test for the `validation_task` decorator itself, but this is to be extra paranoid about it)

PR #12508, opened 2019-10-02
Drop auto-approval delay for first version after content approval

### LINKED ISSUE #12429: Drop initial auto-approval delay after content-approval

An initial listed submission is delayed for 24 hours to give content-reviewers a chance to review the listing.

### What happened?

If a content-reviewer approves within that 24 hour window, auto-approval is still paused until 24 hours have passed.

After content-approval, we know the listing is ok to be published, so there is no need to wait until the delay has passed. We can drop the delay at that point and auto-approve in the next cycle.

diox:
Fixes #12429

### DIFF
--- a/src/olympia/reviewers/models.py
+++ b/src/olympia/reviewers/models.py
@@ -17,7 +17,7 @@ import olympia.core.logger
 from olympia import amo
 from olympia.abuse.models import AbuseReport
 from olympia.access import acl
-from olympia.addons.models import Addon
+from olympia.addons.models import Addon, AddonApprovalsCounter
 from olympia.amo.fields import PositiveAutoField
 from olympia.amo.models import ModelBase
 from olympia.amo.templatetags.jinja_helpers import absolutify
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/models.py
```diff
@@ -1178,13 +1178,19 @@ def check_should_be_delayed(cls, version):
...
+        try:
+            content_review = addon.addonapprovalscounter.last_content_review
+        except AddonApprovalsCounter.DoesNotExist:
+            content_review = None
         return (
             not is_langpack and
             version.addon.status == amo.STATUS_NOMINATED and
-            now - nomination < timedelta(hours=24))
+            now - nomination < timedelta(hours=24) and
+            content_review is None)
```
eviljeff: perhaps we should check the content_review was performed after the nomination date.  Otherwise there is an edge-case where an add-on was approved and content reviewed previously; all it's versions were disabled so went back to incomplete; and a new version was uploaded so the add-on becomes nominated again - if that happened we probably would want to treat it like a new add-on... (right?  I'm not certain on this)

diox: mmm I would say no but I'm not certain either. IMHO the main purpose of the delay is to counter bad actors submitting low-effort malware that pretend to be flash player or adblocker or whatever and that would appear in the recently updated add-ons without the delay.

If an add-on has been approved & content reviewed once, that means it passed the delay once already, so I'm not sure it needs to go through it again (if a reviewer rejected it and all its versions the whole add-on would likely be disabled)

diox: cc @wagnerand for input

wagnerand: I think we can ignore that case. Ideally we should reset content-approved whenever name, summary, description, screenshots etc change.

Issue #12510, opened 2019-10-02
Better handle http timeout in `run_scanner`


willdurand:
### Describe the problem and steps to reproduce it:

1. Submit a webextension (`.xpi` file) with a lot of files (we tried with 2142 files). It should force the `customs` lambda to http timeout (after 30s)

### What happened?

We receive the following JSON:

```json
{
  "message": "Endpoint request timed out"
}
```

The `run_customs` task does not fail, it is successful and it writes the JSON content above in the `ScannersResult` entry for the uploaded add-on.

### What did you expect to happen?

The timeout is correctly detected and the `run_customs` task fails.

vcarciu:
Verified as fixex. If customs timeout, scanner results are not created.

PR #12511, opened 2019-10-02
Better handle http timeout in `run_scanner`

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12510

---

API Gateway returns a 504 HTTP error with a JSON payload when a call has reached the Gateway timeout (30s).


willdurand:
I believe the circle-ci failure is unrelated.

PR #12515, opened 2019-10-02
Make repack_fileupload task use local filesystem for extraction

### LINKED ISSUE #12415: Cannot upload big xpis in Stage and -dev
Steps to reproduce:
1.Try to validate a big(around 100MB) xpi to stage or -dev. You can try with https://addons.mozilla.org/en-US/firefox/addon/lolbyte-league-summoner-search/versions/ .

Expected results:
Validation is successful

Actual results:
"Error: Validation was unable to complete successfully due to an unexpected error."

NOTES:
Not reproducible in prod
From @willdurand : "Soft time limit (110s) exceeded for olympia.devhub.tasks.validate_upload" in logs

diox:
Fixes #12415

### DIFF
--- a/src/olympia/files/tasks.py
+++ b/src/olympia/files/tasks.py
@@ -9,7 +9,7 @@ import olympia.core.logger
 from olympia.amo.celery import task
 from olympia.amo.decorators import use_primary_db
 from olympia.amo.storage_utils import move_stored_file
-from olympia.amo.utils import rm_local_tmp_dir
+from olympia.amo.utils import StopWatch
 from olympia.devhub.tasks import validation_task
 from olympia.files.models import File, FileUpload, WebextPermission
 from olympia.files.utils import extract_zip, get_sha256, parse_xpi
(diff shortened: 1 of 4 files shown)


src/olympia/files/tasks.py
```diff
@@ -57,13 +58,18 @@ def repack_fileupload(results, upload_pk):
     # developer, so it could be something else.
     if upload.path.endswith('.xpi'):
         try:
-            tempdir = extract_zip(upload.path)
-        except Exception:
+            tempdir = tempfile.mkdtemp()  # *not* on TMP_PATH, we want local fs
+            zip_file = SafeZip(source=upload.path)
+            zip_file.extract_to_dest(tempdir)
```
eviljeff: nit: this seems like a place where a decorator would be useful so the folder is always tidied up at the end of the block.

willdurand: I know it does not have anything to do with this patch but it'd be nice to track the timing of this task, or even better, to track:

- the extraction time
- the repacking time
- the time to move the file back

diox: We do log start/end timings for all tasks already, but maybe having more granular timings would be nice, yes. I'll look into that.

diox: mmm maybe I should keep using `extract_zip` - which does encapsulate `extract_to_dest`, though without being a decorator, and instead have it accept an argument for the `tempdir`...

diox: Done in 03bc9af8919da3af3b647420c9c54401fbd6fc15

diox: d35c1670e30a8c55b24310b4f4e3bd674fd76c17 adds individual timings

PR #12520, opened 2019-10-03
Removed the settings for "view source"

### LINKED ISSUE #12418: Remove public source code setting from DevHub
We decided to drop the View Source link in add-on product pages some time ago because of security concerns, and because it's easy to have your sources published in external sites like GitHub.

The [setting is still present in the DevHub](https://discourse.mozilla.org/t/this-add-ons-source-code-is-publicly-viewable-is-it/45292/), though, which is confusing developers. It's at the very bottom, under Technical Details. Let's remove it.

ruchikabgosain:
Fixes #12418 

Removed all the settings for View Source. 
cc: @eviljeff @diox

### DIFF
--- a/src/olympia/devhub/templates/devhub/addons/edit/technical.html
+++ b/src/olympia/devhub/templates/devhub/addons/edit/technical.html
@@ -42,34 +42,6 @@
           </tr>
           <tr>
             <th>
-              {{ tip(_("View Source?"),
-                     _("Whether the source of your add-on can be displayed in our "
-                       "online viewer. It is only relevant for listed add-ons.")) }}
-            </th>
-            <td>
-              {{ flags(_("This add-on's source code is publicly viewable."),
-                       addon.view_source if not editable else form.view_source, editable,
-                       _("No, this add-on's source code is not publicly viewable.")) }}
-            </td>
-          </tr>
-          {% endif %}
-          {% if show_listed_fields %}
-          <tr>
-            <th>
-              {{ tip(_("Public Stats?"),
-                     _("Whether the download and usage stats of your add-on can "
-                       "be displayed in our online viewer. "
-                       "It is only relevant for listed add-ons.")) }}
-            </th>
-            <td>
-              {{ flags(_("This add-on's stats are publicly viewable."),
-                       addon.public_stats if not editable else form.public_stats, editable,
-                       _("No, this add-on's stats are not publicly viewable.")) }}
-            </td>
-          </tr>
-          {% endif %}
-          <tr>
-            <th>
               <abbr title="{{ _('Universally unique identifier') }}" class="label">
                 {{ _('UUID') }}
               </abbr>
--- a/src/olympia/devhub/templates/devhub/addons/edit/technical.html
+++ b/src/olympia/devhub/templates/devhub/addons/edit/technical.html
@@ -40,7 +40,6 @@
               {% endif %}
             </td>
           </tr>
-          {% endif %}
           {% if show_listed_fields %}
           <tr>
             <th>
--- a/src/olympia/devhub/templates/devhub/addons/edit/technical.html
+++ b/src/olympia/devhub/templates/devhub/addons/edit/technical.html
@@ -40,6 +40,7 @@
               {% endif %}
             </td>
           </tr>
+          {% endif %}   
           {% if show_listed_fields %}
           <tr>
             <th>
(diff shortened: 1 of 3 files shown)


ruchikabgosain:
ah, I just needed to know how I can find all these files? @diox

diox:
The files would be scattered across the repos, you need to do a search, there are quite a few of them.

diox:
Because the issue has been split in two, you can focus purely on the code in `src/olympia/devhub`. Anything outside that you can ignore, we'll deal with it in a separate issue. Do fix any broken tests but don't bother modifying fixtures.

src/olympia/devhub/templates/devhub/addons/edit/technical.html
```diff
@@ -40,34 +40,6 @@ <h3>
...
-                       "be displayed in our online viewer. "
-                       "It is only relevant for listed add-ons.")) }}
-            </th>
-            <td>
-              {{ flags(_("This add-on's stats are publicly viewable."),
-                       addon.public_stats if not editable else form.public_stats, editable,
-                       _("No, this add-on's stats are not publicly viewable.")) }}
-            </td>
-          </tr>
-          {% endif %}
```
junngo: I think this place that be out of range about "View Source".

src/olympia/devhub/tests/test_views_edit.py
```diff
@@ -1471,7 +1470,6 @@ def test_technical_on(self):
         response = self.client.post(self.technical_edit_url, data)
         addon = self.get_addon()
 
-        assert not addon.view_source
```
junngo: When this part is deleted, the relevant upper line should also be considered.

src/olympia/devhub/templates/devhub/addons/edit/technical.html
ruchikabgosain: I didn't quite get what you are implying? This section does not need to be deleted? 🤔

src/olympia/devhub/tests/test_views_edit.py
ruchikabgosain: I guess you imply `addon = self.get_addon()`?

src/olympia/devhub/templates/devhub/addons/edit/technical.html
eviljeff: You've deleted too much - the stats settings aren't going.

eviljeff:
You should alway look at the travis test fails btw. It shows one of the mistakes

src/olympia/devhub/templates/devhub/addons/edit/technical.html
junngo: I don't know if you meant it. I think that Area "Public Stats?" as well as area "View Source?" is included.
And, as @eviljeff  said that the template grammar is broken.

src/olympia/devhub/tests/test_views_edit.py
junngo: I think that 1469-1471 lines are preceded for work 1474 line.
So, I mentioned my opinion.

ruchikabgosain:
hi @eviljeff I corrected one of the failing tests on Travis; for the other test, I get the following output for multiple test files: 

`E   jinja2.exceptions.TemplateSyntaxError: Unexpected end of template. Jinja was looking for the following tags: 'elif' or 'else' or 'endif'. The innermost block that needs to be closed is 'if'.` 

Can you tell me how to proceed with this one?

src/olympia/devhub/templates/devhub/addons/edit/technical.html
```diff
@@ -40,19 +40,6 @@ <h3>
...
-                     _("Whether the source of your add-on can be displayed in our "
-                       "online viewer. It is only relevant for listed add-ons.")) }}
-            </th>
-            <td>
-              {{ flags(_("This add-on's source code is publicly viewable."),
-                       addon.view_source if not editable else form.view_source, editable,
-                       _("No, this add-on's source code is not publicly viewable.")) }}
-            </td>
-          </tr>
-          {% endif %}
```
eviljeff: The travis fails are because you need this closing `{% endif %}` still (the opening tag is on line 18)

ruchikabgosain:
only one test case fails now: 
`src/olympia/devhub/tests/test_views_edit.py:1481: in test_technical_devcomment_notrequired`
This code fails and gives an assertion error. 

> addon = self.get_addon()
>         for k in data:
>             if k == 'developer_comments':
>                 assert str(getattr(addon, k)) == str(data[k])
>             else:
>                 assert getattr(addon, k) == (data[k] == 'on')
> 

5458E   AssertionError: assert False == True
5459E     -False
5460E     +True

PR #12523, opened 2019-10-03
Preserve IP addresses for deleted users for 6 months

### LINKED ISSUE #12522: Preserve IP addresses for deleted users for 6 months

In order to fight bad actors, we need to preserve the `last_login_ip` of a deleted user for some time (similar to #11560, but for an account deleted by the user themselves.)

### What happened?

`last_login_ip` gets cleared on account deletion.

Keep `last_login_ip` for 6 months.

Legal and Trust/Security have signed-off on this. @muffinresearch can confirm if needed.

diox:
Previously this was done only for banned users and their IP was kept for a year, now this behavior applies to all deleted accounts, but the delay is reduced to 6 months.

Fixes #12522


diox:
Test failures are unrelated

PR #12532, opened 2019-10-04
drop skip_register_redirect + some refactoring

### LINKED ISSUE #12417: drop 'skip_register_redirect' in fxa configs

eviljeff:
fixes #12417

### DIFF
--- a/src/olympia/conf/prod/settings.py
+++ b/src/olympia/conf/prod/settings.py
@@ -85,34 +85,22 @@ FXA_CONFIG = {
     'default': {
         'client_id': env('FXA_CLIENT_ID'),
         'client_secret': env('FXA_CLIENT_SECRET'),
-        'content_host': 'https://accounts.firefox.com',
-        'oauth_host': 'https://oauth.accounts.firefox.com/v1',
-        'profile_host': 'https://profile.accounts.firefox.com/v1',
-        'redirect_url':
-            'https://%s/api/v3/accounts/authenticate/' % DOMAIN,
-        'scope': 'profile',
+        # redirect_url' = 'https://%s/api/v3/accounts/authenticate/' % DOMAIN,
     },
     'amo': {
         'client_id': env('AMO_FXA_CLIENT_ID'),
         'client_secret': env('AMO_FXA_CLIENT_SECRET'),
-        'content_host': 'https://accounts.firefox.com',
-        'oauth_host': 'https://oauth.accounts.firefox.com/v1',
-        'profile_host': 'https://profile.accounts.firefox.com/v1',
-        'redirect_url': 'https://addons.mozilla.org/api/v3/accounts/authenticate/?config=amo', # noqa
-        'scope': 'profile',
-        'skip_register_redirect': True,
+        # redirect_url = 'https://addons.mozilla.org/api/v3/accounts/authenticate/?config=amo', # noqa
     },
     'code-manager': {
         'client_id': env('CODE_MANAGER_FXA_CLIENT_ID'),
         'client_secret': env('CODE_MANAGER_FXA_CLIENT_SECRET'),
-        'content_host': 'https://accounts.firefox.com',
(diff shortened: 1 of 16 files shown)


src/olympia/conf/prod/settings.py
```diff
@@ -85,34 +85,22 @@
...
     'amo': {
         'client_id': env('AMO_FXA_CLIENT_ID'),
         'client_secret': env('AMO_FXA_CLIENT_SECRET'),
-        'content_host': 'https://accounts.firefox.com',
-        'oauth_host': 'https://oauth.accounts.firefox.com/v1',
-        'profile_host': 'https://profile.accounts.firefox.com/v1',
-        'redirect_url': 'https://addons.mozilla.org/api/v3/accounts/authenticate/?config=amo', # noqa
-        'scope': 'profile',
-        'skip_register_redirect': True,
+        # redirect_url = 'https://addons.mozilla.org/api/v3/accounts/authenticate/?config=amo', # noqa
```
EnTeQuAk: Hmm, if you're commenting this one out in all the configs it's probably best to add an additional comment why it's commented out. Are you using this for documentation reasons?

eviljeff: There's no way of knowing what the redirect url is going to if you didn't personally create the fxa client, so I'm documenting it, yes.

EnTeQuAk: Okay, then it'd make sense to add some more context as part of that documentation, e.g the purpose of that line and why it's here etc.

eviljeff: https://github.com/mozilla/addons-server/pull/12532/commits/8a5b67d43e4b41d1ffaf73648c01d0a2ddb58367

src/olympia/lib/settings_base.py
```diff
@@ -1782,6 +1782,9 @@ def get_raven_release():
     ('olympia', 'amo'),
 )
 
+FXA_CONTENT_HOST = 'https://stable.dev.lcip.org'
+FXA_OAUTH_HOST = 'https://oauth-stable.dev.lcip.org/v1'
+FXA_PROFILE_HOST = 'https://stable.dev.lcip.org/profile/v1'
```
diox: Maybe the default should be prod instead, and these reset in `dev/settings.py` and `settings.py` ? We haven't been super consistent with this.

In any case stage should use prod settings for these - that needs to be fixed.

diox: While you're at it, can you get rid of `VALID_LOGIN_REDIRECTS` ? It's used in `devhub/views.py` in `_clean_next_url()` but judging from the contents of that variable it's completely outdated, gotta be obsolete.

Issue #12217, opened 2019-08-29
Add `run_customs` task to the "chord" of tools that run after the linter


willdurand:
Once https://github.com/mozilla/addons-server/issues/11984 has landed, we should move the `run_customs` function call from the `handle_upload_validation_result` task to the Celery chord that should run after the linter and before `handle_upload_validation_result`.

FTR: We added the `run_customs` task in https://github.com/mozilla/addons-server/pull/12191.

vcarciu:
Verified as fixed by submitting various types of addons. Also we run the load script on AMO-dev and all entries were added.

PR #12546, opened 2019-10-07
Run scanners as Celery tasks in the chord of the validation ch…

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12217
Fixes https://github.com/mozilla/addons-server/issues/12224

---

It was easier to move both tasks at the same time.


src/olympia/devhub/utils.py
```diff
@@ -260,6 +261,12 @@ def __init__(self, file_, addon=None, listed=None, final_task=None):
             if waffle.switch_is_active('enable-yara'):
                 tasks_in_parallel.append(run_yara.s(file_.pk))
 
+            if waffle.switch_is_active('enable-customs'):
+                tasks_in_parallel.append(run_customs.s(file_.pk))
+
+            if waffle.switch_is_active('enable-wat'):
+                tasks_in_parallel.append(run_wat.s(file_.pk))
```
willdurand: I ran into some issues locally, could it be because we are modifying the Celery validation task with waffle? (i.e. it is not always exactly the same and maybe Celery and/or Django don't see the same validation task in the end)

diox: I hadn't thought of that. I don't think it should matter ? Ultimately the chain/chord differs yes, but the definition of the tasks included in them doesn't (the tasks are registered even if the waffle is off), so I'd say it should still work...

src/olympia/scanners/tasks.py
```diff
@@ -49,15 +49,15 @@ def run_scanner(upload_pk, scanner, api_url, api_key):
...
             raise ValueError(response.text)
 
-        if response.status_code != 200 or 'error' in results:
-            raise ValueError(results)
+        if response.status_code != 200 or 'error' in scanner_results:
+            raise ValueError(scanner_results)
 
-        result.results = results
+        result.results = scanner_results
         result.save()
```
diox: So we have `result`, `result.results`, `results`, `scanner_results`, and they are all different things... One is a model instance, the others are dicts of data. This is confusing :)

I would do this:
- Rename `scanner_results` to `data`. It's just the response of running the scanner, and we throw it after, so I think we can get away with it having a generic name
- Rename `result` to `scanner_result` (singular), matching the model name (*)
- Keep `results` as it matches the tasks etc.

(*) A bit unfortunate that the model is called `ScannersResult` plural. Maybe we should change that.

What do you think?

willdurand: Yep, that makes sense.

I renamed `result` to `scanners_result` though, to match the model name. I think I used `Scanners` in the model name to match the package name.

diox: Renaming the model is a task for another day, but generally the convention in the Django world is to keep the model name singular (matches with the fact that one models are instances of a single object) and the app name plural (logic makes slightly less sense but I like to see the app as hosting multiple instances).

willdurand: oh ok, I see. I'll file an issue to rename the model + package then.

diox: Only the model needs to be renamed IMHO - the app name, `scanners`, is already good (and we can consider `yara` as invariable :)

willdurand: oh hu, correct. I misread you. I thought "app" was like "olympia" but that's a project, I guess?

diox: Yeah - though addons-server has a custom structure with no real "project" in the Django sense.

willdurand:
Marked this PR as blocked because we're investigating https://github.com/mozilla/addons-server/issues/12572

willdurand:
> Marked this PR as blocked because we're investigating #12572

Looks like it is very _hard_ to reproduce. In -dev and stage today, QA uploaded more than 200 add-ons in each env, and no error has been found...

PR #12550, opened 2019-10-07
Use `is_webextension` from linter results to determine whether to run yara

### LINKED ISSUE #12436: Use `is_webextension` from parsed data to determine whether to run Yara
We currently determine whether a file should be scanned by looking at the `upload.path` value, checking that it ends with `.xpi`. This is technically valid but not very future-proof.

We can use the `is_webextension` prop in the `results` (coming from the linter).

**For QA:** please make sure we can get yara results for web extensions only.

willdurand:
Fixes #12436

### DIFF
--- a/src/olympia/yara/tasks.py
+++ b/src/olympia/yara/tasks.py
@@ -33,9 +33,9 @@ def run_yara(results, upload_pk):
     log.info('Starting yara task for FileUpload %s.', upload_pk)
     upload = FileUpload.objects.get(pk=upload_pk)
 
-    if not upload.path.endswith('.xpi'):
-        log.info('Not running yara for FileUpload %s, it is not a xpi file.',
-                 upload_pk)
+    if not results['metadata']['is_webextension']:
+        log.info('Not running yara for FileUpload %s, it is not a '
+                 'webextension.', upload_pk)
         return results
 
     try:
--- a/src/olympia/yara/tests/test_task.py
+++ b/src/olympia/yara/tests/test_task.py
@@ -4,40 +4,51 @@ from unittest import mock
 
 from django.test.utils import override_settings
 
+from olympia import amo
 from olympia.amo.tests import TestCase
 from olympia.files.tests.test_models import UploadTest
+from olympia.files.utils import parse_addon
 from olympia.yara.models import YaraResult
 from olympia.yara.tasks import run_yara
 
 
 class TestRunYara(UploadTest, TestCase):
-    @mock.patch('yara.compile')
-    def test_skip_non_xpi_files_with_mocks(self, yara_compile_mock):
-        upload = self.get_upload('search.xml')
+    def setUp(self):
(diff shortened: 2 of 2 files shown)


src/olympia/devhub/tasks.py
```diff
@@ -129,10 +129,12 @@ def create_version_for_upload(addon, upload, channel):
 
 
 @task
-def create_initial_validation_results():
+def create_initial_validation_results(minimal_parsed_data):
     """Returns the initial validation results for the next tasks in the
     validation chain. Should only be called directly by Validator."""
     results = deepcopy(amo.VALIDATOR_SKELETON_RESULTS)
+    if minimal_parsed_data:
+        results.update(minimal_parsed_data)
```
willdurand: Maybe this should be added under `results['minimal_parsed_data']` instead of being merged directly in the main dict?

willdurand: I just did that

willdurand: I just did that

src/olympia/devhub/utils.py
```diff
@@ -283,10 +283,11 @@ def __init__(self, file_, addon=None, listed=None, final_task=None):
             self.file = file_
             self.addon = self.file.version.addon
             addon_data = {'guid': self.addon.guid,
-                          'version': self.file.version.version}
+                          'version': self.file.version.version,
+                          'is_webextension': self.file.is_webextension}
```
willdurand: This is not needed but 🤷‍♂

src/olympia/devhub/tasks.py
```diff
@@ -208,7 +211,17 @@ def validate_upload(results, upload_pk, channel):
...
-    return {**results, **json.loads(force_text(data))}
+    updated_results = {**results, **json.loads(force_text(data))}
+
+    # The tasks in the chord of the validation chain need `is_webextension` in
+    # `results['minimal_parsed_data']`. It often exists but it can be missing
+    # when `parse_data()` raised an error in Validator.
+    if not updated_results['minimal_parsed_data']:
+        linter_metadata = updated_results.get('metadata', {})
+        updated_results['minimal_parsed_data'] = {
+            'is_webextension': linter_metadata.get('is_webextension', False),
```
diox: I'm tempted to say we should directly use this all the time instead of updating results with `minimal_parsed_data`. Any reason not to ?

willdurand: Do you mean directly using `results['metadata']['is_webextension']` in the tasks?

diox: Yeah

willdurand: Ok, so I think I am going to update this patch with just that, but I will also revert https://github.com/mozilla/addons-server/issues/12436, which introduced an issue in `master`.

willdurand:
@diox I updated this patch with the older patch + what we discussed here

PR #12573, opened 2019-10-09
Improve yara results admin performance

### LINKED ISSUE #12564: Improve YaraResult admin filtering
https://github.com/mozilla/addons-server/issues/12549 added filtering in `YaraResult` admin. It's binary for now: we allow filtering of results with matched rules and that's it.

The current way of filtering is not great for performance: it's a `WHERE NOT yara_results.matches = '[]'` in SQL. It's a simple condition at least, but the column it's matching against is a `longtext`, and is not indexed, so it's likely to cause issues.

Ideally if we only care about results with matches or not, we should add a simple boolean column, index it and use that for filtering. If we're likely to want more complex filtering however, like what results match which specific set of rules, then we'll need something else...

diox:
Fixes #12564


src/olympia/scanners/admin.py
```diff
@@ -21,6 +22,19 @@ class ScannersResultAdmin(admin.ModelAdmin):
     fields = ('id', 'upload', 'formatted_addon', 'channel', 'scanner',
               'formatted_results')
 
+    def get_queryset(self, request):
+        # We already set list_select_related() so we don't need to repeat that.
+        # We also need to fetch the add-ons though, and because we need their
+        # translations for the name (see formatted_addon() below) we can't use
+        # select_related(). We don't want to run the default transformer though
+        # so we prefetch them with just the translations.
```
willdurand: Thanks for thr explanation, I was wondering how to do that..

src/olympia/scanners/admin.py
```diff
@@ -41,14 +55,13 @@ def formatted_addon(self, obj):
...
-        return "-"
+        return '-'
     formatted_addon.short_description = 'Add-on'
 
     def channel(self, obj):
         if obj.version:
-            return ('listed' if obj.version.channel ==
-                    amo.RELEASE_CHANNEL_LISTED else 'unlisted')
-        return "-"
+            return obj.version.get_channel_display()
```
willdurand: whoops

src/olympia/yara/models.py
```diff
@@ -26,6 +30,10 @@ def add_match(self, rule, tags=None, meta=None):
             'meta': meta or {},
         })
 
+    def save(self, *args, **kwargs):
+        self.has_matches = bool(self.matches)
```
willdurand: Is it ok to only compute this value here? For example, that's something I'd like to reuse in the task to send a statsd value. I guess I can do it after the `save()` but in the future, one might need this value before, maybe?

diox: I considered doing it in `add_match()` too, but I figured it wasn't enough, in case the `matches` were modified manually. I could do both though...

diox: 0a1b4c9834d3d3d8024b742fe47e53aea23e6238

PR #12579, opened 2019-10-10
Run dockerfile commands with correct user.

### LINKED ISSUE #12366: docker builds are broken, unable to find user olympia
```
Building in Docker Cloud's infrastructure...
Cloning into '.'...
Warning: Permanently added the RSA host key for IP address '192.30.253.112' to the list of known hosts.
Reset branch 'master'
Your branch is up-to-date with 'origin/master'.
KernelVersion: 4.4.0-1060-aws
Components: [{u'Version': u'18.03.1-ee-3', u'Name': u'Engine', u'Details': {u'KernelVersion': u'4.4.0-1060-aws', u'Os': u'linux', u'BuildTime': u'2018-08-30T18:42:30.000000000+00:00', u'ApiVersion': u'1.37', u'MinAPIVersion': u'1.12', u'GitCommit': u'b9a5c95', u'Arch': u'amd64', u'Experimental': u'false', u'GoVersion': u'go1.10.2'}}]
Arch: amd64
BuildTime: 2018-08-30T18:42:30.000000000+00:00
ApiVersion: 1.37
Platform: {u'Name': u''}
Version: 18.03.1-ee-3
MinAPIVersion: 1.12
GitCommit: b9a5c95
Os: linux
GoVersion: go1.10.2
Starting build of index.docker.io/addons/addons-server:latest...
Step 1/32 : FROM python:3.6-slim-stretch
---> 56a67187c07a
Step 2/32 : ENV PYTHONDONTWRITEBYTECODE=1
---> Running in f24632e16f72
Removing intermediate container f24632e16f72
---> 59fe80a2b7a5
Step 3/32 : USER olympia
---> Running in 44f911c926d8
Removing intermediate container 44f911c926d8
---> 76b174e18c00
Step 4/32 : RUN touch /addons-server-docker-container
---> Running in 4abf39d7b061
Removing intermediate container 4abf39d7b061
linux spec user: unable to find user olympia: no matching entries in passwd file
```

EnTeQuAk:
This updates various things…

* Remove the default usage of `USER` in the Dockerfile as this may
  be too restrictive for daily usage
* Use `gosu` and update the entrypoint to fix permissions and drop
  properly into the `olympia` user
* Use `docker-compose run` to actually make sure the entrypoint is run
correctly which isn't the case when using `exec`

Fixes #12366 

Please do test these changes thoroughly, I only briefly ran them on macos and windows but I might not know what to look for specifically.

Things worth testing:

* `make initialize` works as expected and doesn't trample directory permissions
* `make shell` works and initializes itself with the `olympia` user
* inside `make shell` a `make update_deps` works too because we're setting correct permissions
* `make update` works as expected

I pushed a separate `fixed-docker-perms` tag to dockerhub and updated the docker-compose file to make testing easier.


Makefile-os
```diff
@@ -11,17 +11,17 @@ help_submake:
 
 .PHONY: update_docker
 update_docker: ## update all the docker images
-	docker-compose exec worker make update_deps
```
diox: Why the change from `exec` to `run` here and below ? The container should already be running, we don't want another one, we want to re-use the one we have.

EnTeQuAk: As written in the description - I noticed that when using `exec` the entrypoint would simply not be executed which lead to many of the permission problems we've seen in folders on the hosts system.

`run` on the other hand does run the entrypoint properly. I haven't found another reasonable solution to this and I'm perfectly aware that it's not ideal but as the original `web` container stays up I don't think it's an issue.

Also, the storage layers (`/deps/`) seem to be shared between those two so that updates do happen on both afaik.

Happy to use `exec` if you know any way to make the entrypoint being executed properly.

docker-compose.yml
```diff
@@ -23,7 +23,7 @@ x-env-mapping: &env
 services:
   worker: &worker
     <<: *env
-    image: addons/addons-server:latest
+    image: addons/addons-server:fixed-docker-perms
```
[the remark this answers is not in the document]
EnTeQuAk: No, as written in the description…

> I pushed a separate fixed-docker-perms tag to dockerhub and updated the docker-compose file to make testing easier.

It's not permanent and will be switched back before I merge this branch.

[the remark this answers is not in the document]
EnTeQuAk: Yes, before I merge that. For testing it's easier to leave it

EnTeQuAk:
Alrighty, I updated the implementation slightly.

* I created a `olympia` user in the `Dockerfile`
* I commented out the entrypoint for now (see below for why)
* I made use of the `user` option in docker-compose.yml

Now, this seems to work nicely on Linux and I'm not able to reproduce any of the filesystem permission issues we had in the past. I'm not quite sure though how other systems do the mapping so I need more testing for whether the uid/guid mapping is still needed or not.

@eviljeff and @muffinresearch are you able to checkout this branch and test if you still see issues with files owned by another account on your systems for files created inside the docker container?

If you don't, then we do have a fairly simple solution at hand.

muffinresearch:
> Alrighty, I updated the implementation slightly.
> 
>     * I created a `olympia` user in the `Dockerfile`
> 
>     * I commented out the entrypoint for now (see below for why)
> 
>     * I made use of the `user` option in docker-compose.yml
> 
> 
> Now, this seems to work nicely on Linux and I'm not able to reproduce any of the filesystem permission issues we had in the past. I'm not quite sure though how other systems do the mapping so I need more testing for whether the uid/guid mapping is still needed or not.
> 
> @eviljeff (for Windows) and @muffinresearch (for OSX) are you able to checkout this branch and test if you still see issues with files owned by another account on your systems for files created inside the docker container?
> 
> If you don't, then we do have a fairly simple solution at hand.

Yep will do.

Dockerfile
```diff
@@ -80,10 +83,15 @@ ENV PIP_SRC=/deps/src/
 ENV NPM_CONFIG_PREFIX=/deps/
 ENV SWIG_FEATURES="-D__x86_64__"
 
+RUN useradd -Md /code/ olympia
```
muffinresearch: Possibly a bit counter-intuitive to set `$HOME` to `/code` even if we did this historically -  does anything actually rely on that behaviour?

EnTeQuAk: I don't think so and I'm not sure we actually need that. I could try without and see where that gets us

scripts/start-docker.sh
```diff
@@ -9,20 +9,34 @@
...
+#     useradd -Md $(pwd) olympia
+# else
+#     useradd -Md $(pwd) -u $uid olympia
+# fi
+
+deps_uid=$(ls -nd . | awk '{ print $3 }')
+
+# Fix /deps/ folder so that we're able to update the image
+# with the `olympia` user
+if [[ $deps_uid -ne $uid ]]; then
```
diox: I guess this doesn't hurt but why would the user be wrong?

EnTeQuAk: With the recent changes, this entrypoint will hopefully be removed - I haven't done that yet while the testing isn't done.

So, don't worry about that just yet.

EnTeQuAk: I put the `chown` in the Dockerfile instead which enforces the correct user assignment.

docker-compose.yml
```diff
@@ -23,9 +23,13 @@ x-env-mapping: &env
 services:
   worker: &worker
     <<: *env
-    image: addons/addons-server:latest
+    build: .
```
diox: As mentioned on IRC, we don't want to keep that, because this forces a local build instead of a pull. If someone wants to locally build an image, they should either use a custom file if they want to use `docker-compose`, or just use regular `docker`.

docker-compose.yml
```diff
@@ -23,9 +23,13 @@ x-env-mapping: &env
...
+    image: addons/addons-server:fixed-docker-perms
     command: supervisord -n -c /code/docker/supervisor-celery.conf
-    entrypoint: ./scripts/start-docker.sh
+    # We are defaulting to the olympia user to avoid any permission related
+    # issues and files owned by root. This can be overwritten through
+    # the --user option when calling docker-compose
+    user: olympia
```
diox: Maybe add a simple `rootshell` target in `Makefile-os` ?

EnTeQuAk:
I updated quite a bunch of things, I'll add more documentation later once I'm in the colo and will see if the circleci builds work for now. I think we got a tiny bit closer to a working docker environment (and dockerhub builds for that matter…)

Dockerfile
```diff
@@ -2,8 +2,14 @@ FROM python:3.6-slim-stretch
 
 ENV PYTHONDONTWRITEBYTECODE=1
 
-# Run everything as olympia user, by default.
-USER olympia
+ARG GROUP_ID=1000
+ARG USER_ID=1000
```
EnTeQuAk: These require `--build-arg GROUP_ID=$(id -g) --build-arg USER_ID=$(id -u)` - I'll add docs about that later - I found that we don't have any docs at all about local building so I'll do that together with fixing circleci automatic builds (for local dev) as this may change requirements.

PR #12580, opened 2019-10-10
Use `is_webextension` from linter results to determine whether…

### LINKED ISSUE #12537: Use `is_webextension` from parsed data to determine whether to run a scanner
We currently determine whether a file should be scanned by looking at the `upload.path` value, checking that it ends with `.xpi`. This is technically valid but not very future-proof.

We can use the `is_webextension` prop in the `results` (coming from the linter).

willdurand:
Fixes #12537

### DIFF
--- a/src/olympia/scanners/tasks.py
+++ b/src/olympia/scanners/tasks.py
@@ -19,17 +19,22 @@ log = olympia.core.logger.getLogger('z.scanners.task')
 def run_scanner(results, upload_pk, scanner, api_url, api_key):
     """
     Run a scanner on a FileUpload via RPC and store the results.
+
+    - `results` are the validation results passed in the validation chain. This
+       task is a validation task, which is why it must receive the validation
+       results as first argument.
+    - `upload_pk` is the FileUpload ID.
     """
     scanner_name = SCANNERS.get(scanner)
     log.info('Starting scanner "%s" task for FileUpload %s.', scanner_name,
              upload_pk)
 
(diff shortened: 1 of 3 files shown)


src/olympia/yara/tasks.py
```diff
@@ -31,13 +31,14 @@ def run_yara(results, upload_pk):
...
     """
     log.info('Starting yara task for FileUpload %s.', upload_pk)
-    upload = FileUpload.objects.get(pk=upload_pk)
 
     if not results['metadata']['is_webextension']:
         log.info('Not running yara for FileUpload %s, it is not a '
                  'webextension.', upload_pk)
         return results
 
+    upload = FileUpload.objects.get(pk=upload_pk)
```
willdurand: I moved this here to avoid a query for nothing.

src/olympia/scanners/tasks.py
```diff
@@ -19,18 +19,23 @@
...
-
-    if not upload.path.endswith('.xpi'):
-        log.info('Not running scanner "%s" for FileUpload %s, it is not a xpi '
-                 'file.', scanner_name, upload_pk)
+    if not results['metadata']['is_webextension']:
+        log.info('Not running scanner "%s" for FileUpload %s, it is not a '
+                 'webextension.', upload_pk)
         return results
 
+    upload = FileUpload.objects.get(pk=upload_pk)
```
willdurand: I moved this here to avoid a query for nothing.

PR #12586, opened 2019-10-10
Add data migration to backfill YaraResult.has_matches

### LINKED ISSUE #12585: Set `has_matches` on existing yara results
In #12564, we improved the performances of the Yara admin. We added a new field but we did not updated the existing results (in the DB). The problem is that the admin does not list all the results with matches because of that. We should set a value to all the existing yara results.

**For QA:** the Yara admin should list the results with matches by default.

diox:
Fixes #12585

### DIFF
--- a/src/olympia/yara/migrations/0003_auto_20191010_1446.py
+++ b/src/olympia/yara/migrations/0003_auto_20191010_1446.py
@@ -1,0 +1,23 @@
+from django.db import migrations
+from django.db.models import Case, Value, When
+
+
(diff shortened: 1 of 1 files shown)


diox:
The file was generated by running `manage.py makemigrations --empty yara` and then writing the migration in the file created by that command.

willdurand:
> The file was generated by running `manage.py makemigrations --empty yara` and then writing the migration in the file created by that command.

ha! I did not know about `--empty`. Very cool, thanks!

Issue #12547, opened 2019-10-07
Review queues table layout is broken


AlexandraMoga:
### Describe the problem and steps to reproduce it:
1. Log in with a reviewer account
2. Go to the Reviewer tools and select a queue to open - i.e. https://reviewers.addons-dev.allizom.org/en-US/reviewers/queue/auto_approved
2. Notice the layout of all the tables present on the page

### What happened?
Table layouts are a bit off compared to stage

### What did you expect to happen?
No layout issues

### Anything else we should know?
This seems to be a regression from https://github.com/mozilla/addons-server/pull/12526 which introduced the `table-layout: fixed` property

[-dev]
![image](https://user-images.githubusercontent.com/31961530/66312922-22932180-e91a-11e9-8cbf-36f36f750cf7.png)

-------------------

[stage]
![image](https://user-images.githubusercontent.com/31961530/66312943-29219900-e91a-11e9-9a2b-2aa4eb1032b2.png)

AlexandraMoga:
Verified fixed on -dev:

![image](https://user-images.githubusercontent.com/31961530/66383377-8b83a380-e9c5-11e9-8d39-dd27c13e4f09.png)

PR #12592, opened 2019-10-10
Fix review logs layout

### LINKED ISSUE #3447: Long review content break the layout in editor tools - add-on review logs
### Steps to reproduce:
1. Open https://addons-dev.allizom.org/en-US/editors/reviewlog
2. Click on the Show Comment link for a review that contains a long comment

Expected results:
There are no layout or display issues.
### Actual results:

The layout breaks.
### Notes/Issues

Verified in FF48(Win7) Issue can be reproduced in addons-dev and addons-stage
Screencast for this issue: 
reviewhistorybug

yatinmaan:
Fixes #3447

`overflow-wrap` doesn't work for tables without `table-layout: fixed` (The `<td>` expands to content size and there's no overflow to trigger the wrapping).

`table-layout: fixed` causes a regression (Ref: #12547).

### DIFF
--- a/static/css/zamboni/reviewers.less
+++ b/static/css/zamboni/reviewers.less
@@ -403,7 +403,7 @@ ul.tabnav a:hover {
 table.data-grid tr.comments td {
     border-top: none;
     background: #def;
-    overflow-wrap: break-word;
+    word-break: break-all;
     hyphens: auto;
 }
 


eviljeff:
The regression before was because the change was too widely applied - the css is shared among all the reviewer tools pages. So what fixed one page broke another. Have you checked this isn't happening again?

yatinmaan:
Yes. This ones much more limited in scope, it only affects rows with the `comments` class.

eviljeff:
🤞

PR #12603, opened 2019-10-14
Don't auto_approve/disable unlisted versions awaiting review accidentally

### LINKED ISSUE #12602: Cope with unlisted add-ons awaiting review better
While unlisted versions are currently automatically being approved, for a limited window of time they are actually in `STATUS_AWAITING_REVIEW`. That window of time is normally extremely short, but if it was an API upload and the celery workers are slow or stuck, it can be longer.

During that window, at least 2 things currently happen that shouldn't:
- `auto_approve` happily processes the version instead of ignoring it
- If a new _listed_ version is uploaded, `disable_old_files()` will disable all versions awaiting review regardless of their channel instead of just disabling those awaiting review in the same channel

diox:
Fixes #12602

### DIFF
--- a/src/olympia/reviewers/tests/test_commands.py
+++ b/src/olympia/reviewers/tests/test_commands.py
@@ -129,6 +129,7 @@ class TestAutoApproveCommand(AutoApproveTestsMixin, TestCase):
         DiscoveryItem.objects.create(
             recommendable=True, addon=recommended_addon)
 
+        # ---------------------------------------------------------------------
         # Add a bunch of add-ons in various states that should not be returned.
         # Public add-on with no updates.
         addon_factory(file_kw={'is_webextension': True})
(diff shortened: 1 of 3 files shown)


src/olympia/reviewers/tests/test_commands.py
```diff
@@ -129,6 +129,7 @@ def test_fetch_candidates(self):
         DiscoveryItem.objects.create(
             recommendable=True, addon=recommended_addon)
 
+        # ---------------------------------------------------------------------
```
eviljeff: 🤔

src/olympia/reviewers/tests/test_commands.py
```diff
@@ -168,6 +169,15 @@ def test_fetch_candidates(self):
...
 
+        # Add-on with an already public version and an unlisted webext
+        # still awaiting review
+        complex_addon_2 = addon_factory(file_kw={'is_webextension': True})
+        version_factory(
+            addon=complex_addon_2, channel=amo.RELEASE_CHANNEL_UNLISTED,
+            file_kw={'is_webextension': True,
+                     'status': amo.STATUS_AWAITING_REVIEW})
+
+        # ---------------------------------------------------------------------
```
eviljeff: 🤔

src/olympia/reviewers/tests/test_commands.py
diox: Yeah... This test method is getting huge and I needed a way to separate the different setup parts... They could be extracted to separate functions, but those would only be useful for this particular test so I dunno...

eviljeff: I wasn't sure if it was something you added during debugging because the line locations seemed a bit arbitrary at first glance.  130 lines of setup is a bit much... but I guess extracting some to separate functions wouldn't really make it more understandable either (perhaps less so).  🤷‍♂

PR #12607, opened 2019-10-14
Add Stackdriver severity when we go through logging formatting

### LINKED ISSUE #11724: Add `severity` and `message` to logging payload to improve stackdriver log viewing
Stackdriver currently displays the whole json payload for each line of log because we don't set a `message` property. Similarly, because we don't set `severity` all messages are considered to be of the same severity, preventing us to filter by log level.

If we were to add those fields it would increase payload size (since they'd be duplicated) but we'd get better logging viewing experience.

diox:
Fixes #11724

### DIFF
--- a/src/olympia/core/logger.py
+++ b/src/olympia/core/logger.py
@@ -31,25 +31,32 @@ class AmoLoggerAdapter(logging.LoggerAdapter):
         return msg, kwargs
 
 
-class Formatter(logging.Formatter):
-    """Formatter that makes sure REMOTE_ADDR and USERNAME are available.
-
-    Relies on AmoLoggerAdapter to make sure those variables will be set."""
-
-    def format(self, record):
-        for name in 'REMOTE_ADDR', 'USERNAME':
-            record.__dict__.setdefault(name, '')
-        return super(Formatter, self).format(record)
-
-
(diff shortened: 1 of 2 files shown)


src/olympia/core/logger.py
```diff
@@ -31,25 +31,32 @@ def process(self, msg, kwargs):
...
+    STACKDRIVER_LEVEL_MAP = {
+        # 800 is EMERGENCY but Python doesn't have that
+        # 700 is ALERT but Python doesn't have that
+        logging.CRITICAL: 600,
+        logging.ERROR: 500,
+        logging.WARNING: 400,
+        # 300 is NOTICE but Python doesn't have that
+        logging.INFO: 200,
+        logging.DEBUG: 100,
+        logging.NOTSET: 0,
```
EnTeQuAk: I wonder if the stackdriver values should be constants as well but I think for the little amount of usage (here + the tests) it should be fine.

diox: They are technically defined in `google-cloud-python` lib but I didn't want to depend on it just for those constants... And yeah, since it's self contained and only used to match `logging` constants I figured redefining them wasn't necessary.

PR #12612, opened 2019-10-15
Devhub landing page footer

### LINKED ISSUE #12555: Implement new footer for devhub landing page
#11651 specified a new footer to match extensionsworkshop but it wasn't implemented with that issue because the mocks are inconsistent: In logged out view it's the new extension-workshop style footer.  But logged in the sketch file shows the *old* footer for desktop size and no footer at all for mobile size.  

It's unclear if those were all deliberate choices or if there was no particular attention to the logged in view so just included a grab of what was there on desktop already.  That the logged in mobile view doesn't include a footer is an existing issue #4693 - we should fix it at the same time.

@jvillalobos to clarify - IMO we should be using the same footer for both logged in and logged out users, and showing it on mobile and desktop.

eviljeff:
fixes #12555 

Screen Shot 2019-10-15 at 10 25 35-fullpage

Screen Shot 2019-10-15 at 10 41 41-fullpage

Screen Shot 2019-10-15 at 10 26 05-fullpage

Screen Shot 2019-10-15 at 10 42 13-fullpage


src/olympia/devhub/templates/devhub/new-landing/components/dev_footer.html
```diff
@@ -0,0 +1,69 @@
+<footer class="DevHub-Footer content">
```
eviljeff: this is just a renamed and then cut down version of /new-landing/components/footer.html - not sure why github's diff viewer hasn't realised that (it did for the `.less` file).

src/olympia/templates/photon-footer.html
```diff
@@ -0,0 +1,73 @@
+<footer class="Footer">
```
eviljeff: This was based on the html from frontend - with a few changes for urls and includes for the template.

src/olympia/devhub/templates/devhub/new-landing/components/dev_footer.html
diox: In git, renames are not actually renames but moving content from the old file to a new one, and the fact that it's a rename is detected through an heuristic... which fails in some cases like this one.

static/css/photon-site/footer.less
```diff
@@ -0,0 +1,194 @@
+.Footer {
```
eviljeff: This was also based on the frontend css.  Stylus is a little different to less but I think the resulting css should be the same if I rewrote it correctly.

src/olympia/templates/photon-footer.html
```diff
@@ -0,0 +1,73 @@
...
+        <div class="Footer-mozilla-link-wrapper">
+            <a class="Footer-mozilla-link" href="https://mozilla.org/" title="Go to Mozilla's homepage"><span class="Icon Icon-mozilla Footer-mozilla-logo"><span class="visually-hidden">Go to Mozilla's homepage</span></span></a>
+        </div>
+
+        <section class="Footer-amo-links">
+            <h4 class="Footer-links-header"><a href="{{ url('home') }}">Add-ons</a></h4>
+
+            <ul class="Footer-links">
+                <li><a href="{{ url('pages.about') }}">About</a></li>
+
```
muffinresearch: Maybe these blanks lines in between list-items could be removed?

eviljeff: The html is directly copied from addons.mozilla.org/ so yes, but there too then?

Issue #12648, opened 2019-10-17
Migrate `YaraResult` entries to new `ScannerResult` entries


willdurand:
This issue depends on https://github.com/mozilla/addons-server/issues/12631.

**For QA:** all the existing Yara results will be available in the scanner django admin.

AlexandraMoga:
The pre-existing yara results can be found in the Scanner module - both with matched and unmatched rules.
I don't know exactly how many entries we had on -dev before the migration, but I can make a note of  all yara entries from stage to double check if the numbers are correctly imported.

AlexandraMoga:
I've verified with stage that all the pre-existing yara entries have been migrated to Scanner results.

PR #12653, opened 2019-10-18
Migrate yara results

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12648

~~:warning: depends on https://github.com/mozilla/addons-server/pull/12647~~

---

I added a default `ordering` in the scanner admin so that we won't have all the newly created yet old scanner results at the top of the page...

### DIFF
--- a/src/olympia/scanners/migrations/0004_auto_20191018_0740.py
+++ b/src/olympia/scanners/migrations/0004_auto_20191018_0740.py
@@ -5,7 +5,10 @@ from olympia.constants.scanners import YARA
 
 class Migration(migrations.Migration):
 
-    dependencies = [('scanners', '0003_auto_20191017_1514')]
+    dependencies = [
+        ('scanners', '0003_auto_20191017_1514'),
+        ('yara', '0003_auto_20191010_1446'),
+    ]
 
     operations = [
         migrations.RunSQL(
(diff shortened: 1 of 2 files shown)


willdurand:
rebased

src/olympia/scanners/migrations/0004_auto_20191018_0740.py
```diff
@@ -0,0 +1,32 @@
...
+from django.db.models import Case, Value, When
+
+from olympia.constants.scanners import YARA
+
+
+def move_yara_results(apps, schema_editor):
+    YaraResult = apps.get_model('yara', 'YaraResult')
+    ScannerResult = apps.get_model('scanners', 'ScannerResult')
+
+    for yara_result in YaraResult.objects.all():
```
diox: I wonder if we could do it in one SQL query, either with raw SQL or with fancy ORM tricks. Otherwise this is going to take a while...

willdurand: Yeah, that's true.

FWIW, there are 10k entries in production.

willdurand: I switched to a SQL query to copy the data.

willdurand:
rebased

Issue #12654, opened 2019-10-18
Remove Yara admin


willdurand:
This issue depends on https://github.com/mozilla/addons-server/issues/12648.

**For QA:** all existing yara results should have been moved to the scanner results in #12648. For this issue, the "yara django admin" should be gone.

AlexandraMoga:
"Yara" admin module is no longer present on -dev:

![image](https://user-images.githubusercontent.com/31961530/67264621-4dea4480-f4b4-11e9-9e14-5f2308fb07c0.png)

PR #12655, opened 2019-10-18
Remove Yara admin and model

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12654

:warning: depends on https://github.com/mozilla/addons-server/pull/12653

---

The `yara` model and other files will be fully deleted later. We likely need them to be able to run the migrations.


willdurand:
ah oops, I always forget about this test case that counts the number of admins. Fixed.

src/olympia/yara/models.py
```diff
@@ -1,41 +0,0 @@
-from django.db import models
```
eviljeff: Does django need a models.py file to "see" the app as possibly having migrations?

src/olympia/yara/migrations/0004_delete_yararesult.py
```diff
@@ -0,0 +1,16 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('yara', '0003_auto_20191010_1446'),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
+            name='YaraResult',
```
diox: We probably don't want to execute this migration in the same week as the one that replaces `YaraResult` by `ScannerResult` to keep backwards-compatibility during deploy... 

... Although, maybe in this case we could cheat and disable the switch before to be sure there are no errors, and re-enable it after the deploy?

willdurand: I'd vote for disabling the switch or put the site in read only mode because otherwise we'll have to copy the data from `yara_results` to `scanners_results` twice and/or force admins to look at 2 django admins to see the results.

willdurand:
I am going to close this branch and reopen a new PR to only remove the django admin.

Issue #12640, opened 2019-10-17
Create a `Rule` model in `scanners` package


willdurand:
We need a `ScannerRule` model to bind a rule to an action.

PR #12656, opened 2019-10-18
Add a Rule model

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12640


diox:
I'm not sold on the prefix thing. Why not have a `scanner` field ? Much easier to enforce/manipulate.

willdurand:
> I'm not sold on the prefix thing. Why not have a `scanner` field ? Much easier to enforce/manipulate.

True. I changed that and it should be ready for a review, thanks!

src/olympia/scanners/models.py
```diff
@@ -1,44 +1,47 @@
...
-        ]
-        indexes = [
-            models.Index(fields=('has_matches',)),
+            models.UniqueConstraint(
+                fields=('upload', 'scanner', 'version'),
+                name='scanners_results_upload_id_scanner_'
+                'version_id_ad9eb8a6_uniq',
+            )
         ]
+        indexes = [models.Index(fields=('has_matches',))]
```
willdurand: I have a `Black` vim plugin to autoformat my code (because I am missing prettier all the time) 😞

src/olympia/constants/scanners.py
```diff
@@ -7,3 +10,12 @@
...
     YARA: 'yara',
 }
+
+# Action IDs are also used for severity (the higher, the more severe).
+NO_ACTION = 1
+FLAG_FOR_HUMAN_REVIEW = 20
+
+ACTIONS = {
+    NO_ACTION: _('no action'),
+    FLAG_FOR_HUMAN_REVIEW: _('flag for human review'),
```
willdurand: I was thinking about adding the rest when we'll implement them, but maybe we can add all options now.

diox: I think that's fine as-is, we can add to the dict as we implement the actions. I started filing issues but we're going to be blocked soon partially (or at least applying actions to *unlisted* submissions will be blocked)

willdurand: Thanks. I see that the unlisted submissions won't be as straightforward as we initially thought indeed :(

src/olympia/scanners/models.py
```diff
@@ -48,4 +51,24 @@ def save(self, *args, **kwargs):
...
+        help_text=_('This is the exact name of the rule used by a scanner.'),
+    )
+    scanner = models.PositiveSmallIntegerField(choices=SCANNERS.items())
+    action = models.PositiveSmallIntegerField(
+        choices=ACTIONS.items(), default=NO_ACTION
+    )
+    is_active = models.BooleanField(default=True)
+
+    class Meta:
+        db_table = 'scanners_rules'
```
diox: Note: it's perfectly fine to do it, but you don't need to bother with setting `db_table` explicitly every time you create a model - the default is fine. We've got a lot of those hanging around because many of our tables are older than the django app :)

PR #12658, opened 2019-10-18
Refactor unlisted submission to go through auto-approval workflow

### LINKED ISSUE #12622: Refactor unlisted submission to go through auto-approval workflow
We want to unify listed & unlisted submission workflow to make it easier to manage, as well as making it possible to apply some actions to unlisted uploads while they are scanned, such as delaying their approval.

For developers, that will mean although unlisted approvals should still be fast, it will no longer be instantaneous.

This means:
- Modifying devhub & API to not auto-sign, instead letting unlisted versions be set to awaiting review and the UI changed to be similar to listed (with a couple extra explanations added to tell developers to wait for signing and where to find the file once it's signed).
- Processing unlisted versions awaiting review, either through `submit_file` task (currently fired for API uploads only), `auto_approve` cron, or a different cron. (Note `auto_approve` was recently fixed to only care about listed versions, that would need to be reverted if going this route). We might want to have them go through `AutoApprovalSummary` too, to be determined.
- Ensuring this whole thing is transparent - and still fast - for langpack submissions that are made as part of Firefox release process - essentially langpacks that are mozilla signed should be exempted from this as much as possible.

This should be done behind a waffle switch and maybe split into several issues if necessary.

diox:
Because unlisted versions can be submitted when the add-on is disabled, this forces us to consider them as candidates for `auto_approve`, but reject them later during processing.

Fixes #12622


diox:
This should work, but marking as draft cause I want to hunt down the places where tests are missing for those changes, as well as manually test it with `web-ext` to make sure it properly waits for the upload to be marked as `processed` even with unlisted.

diox:
> as well as manually test it with `web-ext` to make sure it properly waits for the upload to be marked as processed even with unlisted.

It doesn't: it eventually outputs `Error: Validation took too long to complete` :(

willdurand:
> It doesn't: it eventually outputs `Error: Validation took too long to complete` :(

Yeah, likely due to https://github.com/mozilla/sign-addon/blob/a8877cf22cc2d6cf9234fe06e83088cf29249e53/src/amo-client.js#L43

diox:
Note: I intent to ping releng before this goes live just in case (which is unlikely to happen this week given the issue with `web-ext`) but their release script seem to wait around 10 minutes for each submission (and they do multiple in parallel) so it wouldn't affect them even if they do unlisted submissions.

diox:
Per discussion in https://github.com/mozilla/addons-server/issues/12722 I'll need to make sure unlisted are auto-approved even if recommendable, *without* getting recommended in the process (only listed are recommended)

src/olympia/reviewers/utils.py
```diff
@@ -479,8 +479,6 @@ def set_addon(self, **kw):
```
diox: FIXME: also make sure in `confirm_auto_approved` we set `confirmed` on unlisted as well.

diox: c29507e566113d736ea301cba7efd87f53584840

diox:
`web-ext` 3.2.1 has been released with a fix. We'll be doing some comms about this on Monday, so this is ready for review.

src/olympia/reviewers/models.py
```diff
@@ -1111,6 +1115,16 @@ def check_uses_native_messaging(cls, version):
         return any('nativeMessaging' in file_.webext_permissions_list
                    for file_ in version.all_files)
 
+    @classmethod
+    def check_is_listing_disabled(cls, version):
+        """Check whether the add-on is disabled or incomplete.
+
+        Only applies to listed versions."""
```
eviljeff: should the `addon.is_disabled` check apply to unlisted versions too?

src/olympia/reviewers/models.py
```diff
@@ -1111,6 +1115,16 @@ def check_uses_native_messaging(cls, version):
...
+        return (
+            version.channel == amo.RELEASE_CHANNEL_LISTED and
+            version.addon.is_disabled and
+            version.status != amo.STATUS_NULL)
```
eviljeff: This method would return `False` if the addon was disabled by the user and the status was `STATUS_NOMINATED` or `STATUS_PUBLIC` - is that what we want?  (It would also return false if _we_ disabled the add-on because then the status would be `STATUS_DISABLED_BY_MOZILLA`)

src/olympia/reviewers/models.py
```diff
@@ -1164,6 +1185,7 @@ def check_should_be_delayed(cls, version):
             content_review = None
         return (
             not is_langpack and
+            version.channel == amo.RELEASE_CHANNEL_LISTED and
```
eviljeff: for all of these `check_...`methods, if the check is expensive (db queries) and it always only applies to listed versions we should probably be doing that check first to avoid unnecessary work for unlisted versions.

src/olympia/reviewers/tests/test_commands.py
```diff
@@ -88,15 +95,16 @@ def test_fetch_candidates(self):
 
         # Add a dictionary: it should also be considered.
         dictionary = addon_factory(
+            name='Disctionary',
```
eviljeff: heh, intentional spelling mistake?

src/olympia/reviewers/utils.py
```diff
@@ -924,8 +931,9 @@ def process_public(self):
         assert self.version.channel == amo.RELEASE_CHANNEL_UNLISTED
 
         # Sign addon.
+        self.sign_files()
         for file_ in self.files:
-            sign_file(file_)
+            ActivityLog.create(amo.LOG.UNLISTED_SIGNED, file_, user=self.user)
```
eviljeff: mmm, we didn't log anything before?

src/olympia/signing/tests/test_views.py
```diff
@@ -351,9 +341,8 @@ def test_system_addon_update_allowed(self):
         assert response.status_code == 202
         addon = Addon.unfiltered.filter(guid=guid).get()
         assert addon.versions.count() == 2
-        latest_version = addon.find_latest_version(
+        addon.find_latest_version(
```
eviljeff: obsolete statement

docs/topics/api/signing.rst
```diff
@@ -139,13 +139,9 @@ Once validation is completed (whether it passes or fails) then the
 passed using the ``valid`` property and check the results with
 ``validation_results``.
 
-If validation passed then your add-on will be submitted for review.
-In the case of unlisted add-ons this will happen automatically.  If your add-on
-is listed then it will be reviewed by a human and that will take a bit longer.
-You can check the ``automated_signing`` property to see if signing will happen
```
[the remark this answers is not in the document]
diox: It's still there. I wanted to de-emphasize it as it's misleading and not really relevant anymore since we started to auto-approve listed versions.

src/olympia/devhub/views.py
```diff
@@ -1145,30 +1144,6 @@ def check_validation_override(request, form, addon, version):
         helper.actions['super']['method']()
 
 
-def auto_sign_file(file_):
-    """If the file should be automatically reviewed and signed, do it."""
-    addon = file_.version.addon
-
-    if file_.is_experiment:  # See bug 1220097.
-        ActivityLog.create(amo.LOG.EXPERIMENT_SIGNED, file_)
```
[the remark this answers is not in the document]
diox: Technically they will eventually be auto-signed as normal, it will just take slightly longer. I intended to mention those changes to releng on top of the public blog post we did, but you're right, we need to go wider internally, I'll do that.

src/olympia/reviewers/models.py
diox: See below.

src/olympia/reviewers/models.py
diox: `Version.objects.auto_approvable()` already excludes addons with `STATUS_DISABLED` status. The only thing this check does on top of that is checking if the listing itself is set to "invisible", which behind the scenes uses `disabled_by_user`. Unlisted have always ignored that.

eviljeff: `is_listing_user_disabled`?

diox: I'm not sure we want the last check, `version.status != amo.STATUS_NULL` though...

src/olympia/reviewers/models.py
diox: Yeah. I had that in mind but for the moment there aren't that many extra db queries (a lot of the stuff fetched is useful at some point or another), so I favored readability, even if that meant an extra db query in some cases.

src/olympia/reviewers/tests/test_commands.py
diox: No :)

src/olympia/reviewers/utils.py
diox: We did but not here.

src/olympia/reviewers/models.py
diox: The actual check I wanted to write is:
```python
return (
            version.channel == amo.RELEASE_CHANNEL_LISTED and (
                version.addon.is_disabled or
                version.addon.status == amo.STATUS_NULL)
        )
```

If the version is listed, we care whether the add-on is disabled or incomplete and return `True` (preventing approval) if that's the case.

diox: 88fe4f1685e6d43a39d0e4ccbed8e2d008a11ea5

src/olympia/signing/tests/test_views.py
diox: 93af1645ecebaf7f58ac320f35b2171de7cfe111

src/olympia/reviewers/tests/test_commands.py
diox: d6ed5c4c6159e93286f55f9bc99d9ff0d6c686f0

PR #12667, opened 2019-10-21
Fix id locale

EnTeQuAk:
travis builds are currently broken because of broken locales


EnTeQuAk:
Apparently localizers fixed the issues themselves :man_shrugging:

Issue #12674, opened 2019-10-21
Add a Django admin for the scanner rules


willdurand:
We introduced a new model named `ScannerRule` in #12656. Let's create a django admin for it so that admins can bind rules to actions.

AlexandraMoga:
@willdurand should the entries added in Scanner rules have any effect at this point, or should we only test that the module is working as expected, i.e adding/editing/deleting rules?

willdurand:
> @willdurand should the entries added in Scanner rules have any effect at this point, or should we only test that the module is working as expected, i.e adding/editing/deleting rules?

There is no effect at this point. You can add/edit/delete rules, but it won't be used for anything else.

AlexandraMoga:
Verified fixed on -dev.
For the actions that are currently available, the scanner rules module is working as expected.

PR #12675, opened 2019-10-21
Add Django admin for `ScannerRule`

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12674

---

Add a django admin for the `ScannerRule` model.

Screen Shot 2019-10-21 at 17 55 14

### DIFF
--- a/src/olympia/scanners/tests/test_admin.py
+++ b/src/olympia/scanners/tests/test_admin.py
@@ -14,8 +14,12 @@ from olympia.amo.tests import (
 )
 from olympia.amo.urlresolvers import reverse
 from olympia.constants.scanners import CUSTOMS, WAT, YARA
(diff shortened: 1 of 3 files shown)


src/olympia/scanners/tests/test_admin.py
```diff
@@ -173,3 +177,28 @@ def test_list_can_show_all_entries(self):
...
+        super().setUp()
+
+        self.user = user_factory()
+        self.grant_permission(self.user, 'Admin:Advanced')
+        self.client.login(email=self.user.email)
+        self.list_url = reverse('admin:scanners_scannerrule_changelist')
+
+        self.admin = ScannerRuleAdmin(
+            model=ScannerRule, admin_site=AdminSite()
+        )
```
eviljeff: `self.admin` doesn't seem to be used anywhere.

PR #12685, opened 2019-10-22
Pre-select platforms from addons current version.

### LINKED ISSUE #10398: If "Firefox for Android" was checked for a previous addon version it should remain checked on submitting a new version
From: https://github.com/mozilla/addons-frontend/issues/7394

---

1. Submit a new addon to AMO.
2. Mark ` Firefox for Android` under `Which applications is this version compatible with?`
3. Submit an updated version of the same addon to AMO.
4. Check what the default state is of the `Firefox for Android` checkbox

### What happened?
The checkbox was not checked.

The checkbox should be checked as previous version were also compatible with Firefox for Android. Now it is incredibly easy to miss that checkbox when submitting your addon and accidentally withhold updates from a user group. 

This issue is a result of the discussion in [this bugzilla issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1519725)

EnTeQuAk:
Fixes #10398

This pre-selects platforms from the addons current version if there is
any.

### DIFF
--- a/src/olympia/devhub/tests/test_forms.py
+++ b/src/olympia/devhub/tests/test_forms.py
@@ -25,6 +25,7 @@ from olympia.devhub import forms
 from olympia.files.models import FileUpload
 from olympia.signing.views import VersionView
 from olympia.tags.models import AddonTag, Tag
+from olympia.versions.models import ApplicationsVersions
 
 
 class TestNewUploadForm(TestCase):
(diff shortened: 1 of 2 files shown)


src/olympia/devhub/tests/test_forms.py
```diff
@@ -37,6 +38,23 @@ def test_firefox_default_selected(self):
...
+            version=addon.current_version, application=amo.ANDROID.id,
+            min=appversion, max=appversion)
+
+        upload = FileUpload.objects.create(valid=False)
+        data = {'upload': upload.uuid}
+        request = req_factory_factory('/', post=True, data=data)
+        request.user = user_factory()
+        form = forms.NewUploadForm(data, request=request, addon=addon)
+        assert form.fields['compatible_apps'].initial == [
+            amo.FIREFOX.id, amo.ANDROID.id]
```
eviljeff: can we test first that `form.fields['compatible_apps'].initial` would otherwise not contain `amo.ANDROID.id` if `addon.current_version` hadn't been changed?

Issue #12693, opened 2019-10-22
Add `matched_rules` to `ScannerResult`


willdurand:
In order to work on #12688, we need to associate `ScannerRule` and `ScannerResult` entries when we can. This issue is about doing this.

diox:
That seems reasonable - the `ScannerRule` model is going to be in prod this week so we could prepare everything on all environments before the migration gets deployed. (I like storing the number of matches rather than a boolean, too).

willdurand:
I kept `has_matches` as a boolean field for now and updated the patch to compute its value on `save()`.

willdurand:
See also https://github.com/mozilla/addons-server/pull/12714

AlexandraMoga:
Verified fixed on -dev and stage. 
The matched rule identified for a scanner result is linking to the actual scanner rule details.

PR #12694, opened 2019-10-22
Add `matched_rules` to `ScannerResult`

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12693

---

This patch adds:

- a new M2M relation between a `ScannerResult` and a `ScannerRule` through `ScannerMatch`
- a migration to associate `ScannerResult` with `ScannerRule`


src/olympia/scanners/models.py
```diff
@@ -24,6 +43,7 @@ class ScannerResult(ModelBase):
         null=True,
     )
     has_matches = models.NullBooleanField()
+    matched_rules = models.ManyToManyField(ScannerRule)
```
diox: It might be a good idea to start having an intermediary model right away, even if we don't use it yet. It'll make adding fields/custom logic easier in the future.

willdurand:
@diox Travis-CI is having issues but apart from that, this is ready for a review (it was passing before, I just changed how the model is imported in the migrations).

willdurand:
I created the rules in -dev.

PR #12698, opened 2019-10-23
Added data-no-csrf to newsletter form

psiinon:
This is to address https://github.com/mozilla/addons-frontend/issues/8816
Not adding a `fixes` line as that issue is managed by the baseline scan.


eviljeff:
The issue was moved to mozilla/addons-server#12699 (adding the backlink for reference)

PR #12712, opened 2019-10-24
Add flag for versions needing human reviews, and display them in reviewer tools

### LINKED ISSUE #12670: Implement "Flag for human review" flag
"Flag for human review" is one of the actions we want to implement in the workflow of the scanners. When triggered, this action should not prevent auto-approval but flag the entire add-on for human review in a separate queue.

Reviewers can unflag, but for now that shouldn't prevent scanners from reflagging the add-on for the same reason after a new upload.

In terms of implementation this is probably going to be a new `BooleanField` in `AddonReviewerFlags` called `needs_human_code_review`. Like `needs_admin_code_review` it doesn't block auto-approval (though may bump the weight assigned to the versions approved). The flag should be set before approving the version, allowing us to see all versions posted since the flag was set including the one that triggered the flag.

We'll also need a separate queue in reviewer tools to see those flagged add-ons. We need to clarify how to display unlisted & listed versions in there, since the flag could be triggered by either and the review pages for are separate for each channel.

diox:
This flag is going to be set automatically by scanners. It needs to be set on versions and not add-ons to ensure specific versions that were found to be problematic by scanners are examined by human reviewers.

Fixes #12670


diox:
(Actually a draft, but I messed up when creating the PR)

diox:
Queue looks like this:
Queue
When clicking on the review page links in that queue, the versions that need human review appear in addition to the last 10 versions of the add-on:
Review page when coming from the queue

Approving, confirming approval and rejecting (either the add-on or multiple versions) will all unflag versions that were displayed, as we consider the reviewer saw them during their review.

src/olympia/reviewers/views.py
```diff
@@ -733,14 +752,41 @@ def review(request, addon, channel=None):
...
+        # top of the regular transformers.
+             .transform(Version.transformer_activity)
+    )
+
+    # Specific versions for the reviewer to look at / act upon when POSTing.
+    needs_human_review_versions = []
+    needs_human_review_versions_ids = (
+        request.POST.get(
+            'needs_human_review_versions_ids',
+            request.GET.get('needs_human_review_versions_ids', ''))
```
eviljeff: so the versions table will include the versions that need human review too?  If there are a number of versions it'll make the page pretty long with the duplicate version details.  I'm not sure how likely a UX review is on a reviewer tools patch but have Product at least okay'd this layout?

e.g. Don't display a separate list of versions but highlight the specific versions within the main version list?

src/olympia/reviewers/views.py
```diff
@@ -173,6 +174,12 @@ def dashboard(request):
             'https://wiki.mozilla.org/Add-ons/Reviewers/Guide'
         ),
         ))
+        sections[ugettext('Need Human Review')] = [(
+            ugettext('Need Human Review ({0})').format(
```
eviljeff: I'm still not convinced about "Needing Human Review" as a thing - *all* of the queues contain addons that need some sort of human review.

src/olympia/reviewers/views.py
diox: The problem with highlighting the versions in the page is that you then potentially miss a few, and then you don't know which ones to unset the flag on at submission, since they weren't all shown.

My idea was, when this flag is set by scanners, we should react quickly, and don't let the versions needing review accumulate on any given add-on, so the lack of pagination shouldn't be too much of a problem.

src/olympia/reviewers/views.py
diox: That's fair. I haven't found a good alternative yet though :(

src/olympia/reviewers/views.py
diox: (Note: I do highlight the versions that have the flag set in the "main" history section, regardless of whether or not the query parameter is present)

eviljeff: I don't have an answer either to the problem either - that's why I was pushing chosing the UX solution onto someone in Product 😉

eviljeff:
Can you include a screenshot of an entire review page?

diox:
Here is a full page shot for the add-on I was testing with, which has 6 versions in total, 2 flagged for human review. Note that the extra section would only appear if coming from the needs human review queue.

Screenshot_2019-10-29 WebExt-eval-test Reviewer Tools Add-ons for Firefox

src/olympia/reviewers/views.py
diox: @wagnerand has already seen this in slack, but looping him again and @jvillalobos to see if they have suggestions.

jvillalobos: I agree with @eviljeff that having two lists is a bit much. It's creating duplicate info and making an already very complex page more complex. It's not unlikely that many versions pending human review will accumulate, so that list may become a problem. It's also unlikely that if an add-on accumulates 5+ versions that need human review, that all of them will actually end up being reviewed, instead of maybe the last version or two.

I think that highlighting the versions in the history should be good enough, and it would be the reviewer's responsibility to figure out which versions actually need to be reviewed before submitting their approval.

src/olympia/reviewers/views.py
jvillalobos: "Blocked on Review", or "Signature Pending Review"?

src/olympia/reviewers/views.py
diox: We've been discussing that over IRC with @wagnerand earlier and reached broadly similar conclusions. 

I haven't pushed those changes yet but this is where I landed:
Screenshot_2019-10-29 WebExt-eval-test Reviewer Tools Add-ons for Firefox(2)

Versions needing human review are highlighted in the history, and if there are more needing attention outside of the paginated range there is a mention of that.

src/olympia/reviewers/views.py
diox: The thing is, it doesn't necessarily block approval. Delaying or outright blocking automated approval is yet another step in the actions scanner rules can trigger. This is more of a "the scanner found something, but can't be 100% sure, need a human to double-check". The version can be auto-approved despite this.

jvillalobos: Ah, I see. Other suggestions: "Flagged for Human Review", "Flagged by Scanners", "Auto-flagged". All that said, the wording doesn't matter that much since reviewers will get used to whatever we pick very quickly.

eviljeff: "Scanner Flagged" (similar naming to Admin Flagged)
"Scanner Review" or "Malware Review" (similar naming to Content Review)

diox:
Since the thread with the latest screenshot is going to be outdated, repeating it here:

I haven't pushed those changes yet but this is where I landed:
Screenshot_2019-10-29 WebExt-eval-test Reviewer Tools Add-ons for Firefox(2)

Versions needing human review are highlighted in the history, and if there are more needing attention outside of the paginated range there is a mention of that. 

When a reviewer submits a review:
- If they are approving, clear the flag on the latest version as well as all listed versions. It's assumed that users will automatically update to the latest version, which the reviewer should have looked at. 
- If they are rejecting, clear the flag only on versions being rejected.

Unlisted versions handling will need more work in a follow-up issue. They may require an entirely different workflow.

src/olympia/reviewers/utils.py
```diff
@@ -696,6 +746,11 @@ def process_sandbox(self):
         self.set_files(amo.STATUS_DISABLED, self.files,
                        hide_disabled_file=True)
 
+        # Unset needs_human_review on the latest version - it's the only
+        # version we can be certain that the reviewer looked at.
+        if self.request:
```
eviljeff: mmm, I know we test on `self.request` as a way of determining if it's an auto approval by the cron or a manual one, but it's a bit magic sauce, especially when there no notes in the method about the logic.  I wonder if we should just set a `self.is_not_cron` (with a better name) to be explicit about it.

src/olympia/reviewers/views.py
```diff
@@ -629,6 +639,15 @@ def queue_expired_info_requests(request):
                   qs=qs, SearchForm=None)
 
 
+@permission_or_tools_view_required(amo.permissions.ADDONS_REVIEW)
+def queue_needs_human_review(request):
+    admin_reviewer = is_admin_reviewer(request)
+    qs = Addon.objects.get_needs_human_review_queue(
+        admin_reviewer=admin_reviewer)
+    return _queue(request, NeedsHumanReviewTable, 'needs_human_review',
+                  qs=qs, SearchForm=None)
```
eviljeff: No `SearchForm`?

src/olympia/reviewers/views.py
```diff
@@ -733,6 +752,21 @@ def review(request, addon, channel=None):
...
+        # addon.versions related manager to get `addon` property pre-cached on
+        # each version.
+        addon.versions(manager='unfiltered_for_relations')
+             .filter(channel=channel)
+             .select_related('autoapprovalsummary')
+        # Add activity transformer to prefetch all related activity logs on
+        # top of the regular transformers.
+             .transform(Version.transformer_activity)
+    )
+
```
eviljeff: nit: unnecessary change - you don't need to move this block of code up anymore.

src/olympia/reviewers/views.py
```diff
@@ -872,6 +893,12 @@ def review(request, addon, channel=None):
         verdict_info = summary.calculate_verdict(pretty=True)
         auto_approval_info[a_version.pk] = verdict_info
 
+    # We want to notify the reviewer if there are versions needing extra
+    # attention that are not present in the versions history (which is
+    # paginated).
+    versions_flagged_by_scanners = versions_qs.filter(
```
eviljeff: `additional_versions_flagged_by_scanners` ?  It's a count that doesn't include versions shown on the current page.

src/olympia/reviewers/views.py
diox: I never include it if it's not requested in the issue cause reviewer tools search is a pain to work with. We can add it later if needed.

src/olympia/reviewers/utils.py
diox: 397194b60bd0a50c0d3982a4cd01c8e9717b51b5

src/olympia/reviewers/views.py
diox: I wanted to have that originally but it broke 80 characters limit when included in context below (`additional_versions_flagged_by_scanners=additional_versions_flagged_by_scanners`) so I didn't bother.

Issue #12601, opened 2019-10-14
Delete `scanners_results`/`yara_results` entries not bound to a version after some time


willdurand:
We should delete the scanner and yara results not bound to a version (when no version has been created, likely because the submission process has been aborted) after some time. We can probably do that in the "gc" CRON.

**For QA:** this will likely shorten the list of "all scanner results" but it should only delete entries with no file upload and no version attached to them.

diox:
Maybe just delete all `YaraResult` with `upload` and `version` set to `None` - the `gc` cron already deletes `FileUpload` older than 7 days, so if both fields are set to `None` you're dealing with an aborted submission that has already been considered obsolete since its `FileUpload` has been deleted.

AlexandraMoga:
I've looked inside scanner results and indeed the number of entries with no version attached, which are older than 7 days, have been removed.

@willdurand it looks like the [first scanner object](https://addons-internal.dev.mozaws.net/en-US/admin/models/scanners/scannerresult/1/change/?_changelist_filters=has_matches%3Dall%26p%3D32) hasn't been removed, although there doesn't appear to be any upload or version attached to it.
Also, while verifying this issue, I figured it could be helpful to have a Creation Date column for scanner results. What do you think?

willdurand:
> @willdurand it looks like the [first scanner object](https://addons-internal.dev.mozaws.net/en-US/admin/models/scanners/scannerresult/1/change/?_changelist_filters=has_matches%3Dall%26p%3D32) hasn't been removed, although there doesn't appear to be any upload or version attached to it.

@AlexandraMoga yeah, I've seen this one too. It is because the add-on is marked as deleted. It will be fixed by https://github.com/mozilla/addons-server/issues/12735.

> Also, while verifying this issue, I figured it could be helpful to have a Creation Date column for scanner results. What do you think?

I don't know, maybe?

AlexandraMoga:
> I don't know, maybe?

Well, for me it would have been helpful in this context. There might be other scenarios in which this Creation date could prove useful. 
We could also leave it for the admins to decide if they require this column or not.

willdurand:
> Well, for me it would have been helpful in this context. There might be other scenarios in which this Creation date could prove useful.
> We could also leave it for the admins to decide if they require this column or not.

It is fine to add it I think. Could you please file an issue for it?

PR #12721, opened 2019-10-25
Delete scanner results when both upload and version are None

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12601


src/olympia/amo/cron.py
```diff
@@ -57,6 +58,12 @@ def days_ago(days):
                 pass
         file_upload.delete()
 
+    stale_scanner_results = ScannerResult.objects.filter(
+        upload=None, version=None
+    ).order_by('id')
+    for scanner_result in stale_scanner_results:
+        scanner_result.delete()
+
```
eviljeff: unless there's custom logic in the model's `delete` method, you can just do `ScannerResult.objects.filter(upload=None, version=None).delete()`

willdurand: Oh, I did not know that. Thanks!

Issue #12571, opened 2019-10-09
Add check_for_api_keys_in_file to the validation chain


diox:
https://github.com/mozilla/addons-server/issues/12505 added a celery `chain` to orchestrate the validation tasks. #12217 and #12224 will include the scanners and yara in a `chord` inside that chain to run those in parallel.

There is one last cleanup to do: we should look moving `check_for_api_keys_in_file` task outside of `handle_upload_validation_result()`. It's not an annotation: it should run as part of the validation chain, either in the main chain, or in the `chord` that includes yara and the scanners.

willdurand:
> There is one last cleanup to do: we should look moving `check_for_api_keys_in_file` task outside of `handle_upload_validation_result()`. It's not an annotation: it should run as part of the validation chain, either in the main chain, or in the `chord` that includes yara and the scanners.

If we add this function as a `@task` in the validation `chain` (say, after the linter and before the `chord`), we will not execute the `chord` if an api key is detected (since the function inserts a validation "error" message).

Does that sound okay?

diox:
Yes. The developer will have to resubmit anyway, so I think this is fine.

ioanarusiczki:
@diox Can you give some str for qa. I don't know if I got it right. I should get an error if api keys are detected (should that be inside the manifest.json) ?

diox:
This is a refactor of https://bugzilla.mozilla.org/show_bug.cgi?id=1456797 so the same steps should apply. If a developer puts their api key anywhere in a non-binary file in their add-on, it should fail validation with an appropriate error message and revoke affected credentials automatically shortly (2 minutes is the currently configured time) after.

AlexandraMoga:
Verified on -dev using both devhub and the signing API submission.
If the developers add their API credentials in any of the add-on files, the following validation error will be raised:
![image](https://user-images.githubusercontent.com/31961530/68119498-8dc91700-ff0b-11e9-8b3e-947ae7372a09.png)

Subsequently, the respective API keys will be revoked and the user can generate new ones.

@diox for the API submissions, the validation error is not returned in the response body, so there is no immediate way of telling what caused the failure. There is, however, a link to the validation results which can be opened in devhub, where the reason for the submission failure is clear.  Would it be possible to show this error message directly in the response?

diox:
Changing how validation errors are returned in the API is a big task, out of scope here.

PR #12745, opened 2019-10-28
Use `check_for_api_keys_in_file()` as a validation task

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12571

---

This patch removes the `check_for_api_keys_in_file()` call in `handle_upload_validation_result()` and runs it as a validation task in the validation chain (after the linter).

### DIFF
--- a/src/olympia/devhub/tasks.py
+++ b/src/olympia/devhub/tasks.py
@@ -298,13 +298,6 @@ def handle_upload_validation_result(all_results, upload_pk, channel,
 
     upload = FileUpload.objects.get(pk=upload_pk)
 
-    # Check for API keys in submissions.
-    # Make sure it is extension-like, e.g. no search plugin
-    try:
-        results = check_for_api_keys_in_file(results=results, upload=upload)
-    except (ValidationError, BadZipfile, IOError):
-        pass
-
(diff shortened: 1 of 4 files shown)


willdurand:
Only 1 test case is failing in the ES job but I am not sure why..

diox:
> Only 1 test case is failing in the ES job but I am not sure why.. 

There are a couple intermittent failures in the ES job that have been happening more often recently...

diox:
This doesn't seem to work for me locally - submitting an add-on with `web-ext` that includes my secret key in the content, the `FileUpload` is never marked as `processed` (there is never a `validation` attached), `web-ext` eventually times out after polling for a while.

diox:
mmm I didn't try via devhub...

diox:
mmm that's weird. I need to look in my celery logs, I could see some things happening but I didn't monitor it too closely, I will have a second look.

diox:
All good, sorry, it was an issue on my local env, I had a migration unapplied as the branch I was working on in parallel was a little behind master, that caused an `IntegrityError` when submitting add-ons from your branch even though everything was working from mine.

Tested with devhub and API, it works as expected, the task is fired in the chain as expected, everything is ok.

PR #12773, opened 2019-11-01
support up to Firefox65535! Plus, add a working addon version compare

### LINKED ISSUE #6016: Stop hardcoding '99' as the maximum Firefox version ever possible
In the [version compare code](https://github.com/mozilla/addons-server/blob/ddc439ca0cdf64b6569f36dbdf7c6cd64629e717/src/olympia/versions/compare.py), the [search code](https://github.com/mozilla/addons-server/blob/f74cf659602a1c4f066a537ba52c8d620111cda1/src/olympia/search/views.py#L580-L582) and the [applications definition](https://github.com/mozilla/addons-server/blob/ddc439ca0cdf64b6569f36dbdf7c6cd64629e717/src/olympia/constants/applications.py#L176-L181) the value `99` is hardcoded. It represents the maximum Firefox version ever possible, that value is supposed to be always higher than Firefox's (and other apps).

This results in things like:
```
ipdb> dict_from_int(version_int('99'))
{'pre': None, 'major': 99, 'minor3': 0, 'minor2': 0, 'minor1': 0, 'alpha': None, 'alpha_ver': 0, 'pre_ver': 0}
ipdb> dict_from_int(version_int('99999'))
{'pre': None, 'major': 99, 'minor3': 0, 'minor2': 0, 'minor1': 0, 'alpha': None, 'alpha_ver': 0, 'pre_ver': 0}
```

Breaking some of the compatibility comparison code once Firefox reaches version 99. Firefox not ever reaching version 99 may have been a valid assumption to make [six years ago when it was implemented](https://github.com/mozilla/addons-server/commit/ca6d1a1) but if the naming scheme does not change, that won't be the case in six more years...

eviljeff:
fixes #6016

addon_version_hash isn't used anywhere currently but I thought I'd leave it in because I'm expecting to need it at some point for the blocklist code, and it should hopefully discourage anyone else from using `version_int` for addon versions.


src/olympia/applications/migrations/0002_auto_20191101_1222.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 2.2.6 on 2019-11-01 12:22
+
+from django.db import migrations
+
+from olympia.versions.compare import version_int
+
+
+def increase_version_int_for_asterisk(apps, schema_editor):
+    AppVersion = apps.get_model('applications', 'AppVersion')
+    for app in AppVersion.objects.filter(version='*'):
```
eviljeff: we only need to update "*" because the minor parts are staying at 99.

src/olympia/versions/compare.py
```diff
@@ -39,15 +43,45 @@ def version_dict(version):
...
+    """Suitable for comparing addon version strings that are Chrome compatible
+    plus the limited a, b, pre suffixes we support for app versions.  Returns
+    a very large integer (that's too big to store as a BIGINT in mysql).
+    """
+    vdict = _get_version_dict(
+        version, max_number_minor=MAX_VERSION_PART,
+        max_number_major=MAX_VERSION_PART)
+
+    # use hex numbers to simplify the conversion.
+    # alpha and pre can only be 0,1,2 so will always be a single digit; prevar
```
diox: typo `prevar`

src/olympia/versions/compare.py
```diff
@@ -39,15 +43,45 @@ def version_dict(version):
...
+        max_num = max_number_major if num == 'major' else max_number_minor
+        vdict[num] = min(vdict[num] or 0, max_num)
     vdict['alpha'] = {'a': 0, 'b': 1}.get(vdict['alpha'], 2)
     vdict['pre'] = 0 if vdict['pre'] else 1
+    return vdict
+
+
+def version_int(version):
+    """This is used for converting an app version's version string into a
+    single number for comparision.  To maintain compatibility the minor parts
```
diox: Typo `comparision`

src/olympia/versions/compare.py
```diff
@@ -39,15 +43,45 @@ def version_dict(version):
...
+        version, max_number_minor=99, max_number_major=MAX_VERSION_PART)
 
     vint = '%d%02d%02d%02d%d%02d%d%02d' % (
         vdict['major'], vdict['minor1'], vdict['minor2'], vdict['minor3'],
         vdict['alpha'], vdict['alpha_ver'], vdict['pre'], vdict['pre_ver'])
-    return min(int(vint), MAXVERSION)
+    return min(int(vint), BIGINT_POSITIVE_MAX)
+
+
+def addon_version_hash(version):
```
diox: Not a big fan of calling this an `hash` - it's just a potentially long int... Why not call this `addon_version_int` in fact ?

src/olympia/versions/compare.py
```diff
@@ -39,15 +43,45 @@ def version_dict(version):
...
+    a very large integer (that's too big to store as a BIGINT in mysql).
+    """
+    vdict = _get_version_dict(
+        version, max_number_minor=MAX_VERSION_PART,
+        max_number_major=MAX_VERSION_PART)
+
+    # use hex numbers to simplify the conversion.
+    # alpha and pre can only be 0,1,2 so will always be a single digit; prevar
+    # is parsed as single digit by version_dict.
+    hash_string = ('%x' '%04x' '%04x' '%04x' '%x' '%04x' '%x' '%x') % (
```
diox: As above, not really a hash. `hex_string` would be more correct.

src/olympia/versions/compare.py
eviljeff: yeah, I went back and forward on returning a string instead of a int

eviljeff: https://github.com/mozilla/addons-server/pull/12773/commits/4a7316d59f8020ba0e83a2e1c785c13a36130b9a

PR #12776, opened 2019-11-01
Experimental: Integrate black code formatter

### LINKED ISSUE #8617: Use black for automatically formatting our source code
I've been dreaming to do this on addons-server for a long time and now that black grows to a quite sophisticated tool to do proper source formatting in python projects we may as well try.

Their style guide: http://black.readthedocs.io/en/stable/the_black_code_style.html

Black is very opinionated and obliterates our current imports-style-guide to some extent. It doesn't touch the ordering there at all though, only the wrapping, which is nice, that means we can still combine it with pytest-isort.

Black recently gained the ability to *not* force regular quotes which means we can continue to use single-quotes which would simplify the whole diff and migration a bit.

``black --skip-string-normalization --line-length 79 src/``

The output of that can be seen in [this experimental branch](https://github.com/mozilla/addons-server/compare/black-experiment).

This obviously only works if everyone is doing it and we are enforcing it via CI so we should all be on the same page before we do this.

I do think that this will simplify code reviews and also on-boards contributors much better since no one needs to worry about how the code looks anymore.

It can be integrated into CI using ``black --check`` as part of our ``codestyle`` task. We may want to exclude existing migrations in a final run (I didn't in the experimental branch). There is an active community around it and good [integration into editors and tools ](http://black.readthedocs.io/en/stable/editor_integration.html)

Alrighty, opinions, thoughts, objections? @diox @eviljeff

EnTeQuAk:
Fixes #8617

* Adds flake8 integration (see screenshot below)
* Adds reasonable configuration that is based on our flake8 config so far

Opinions?

Screenshot from 2019-11-01 14-49-30


eviljeff:
Can you point to the changes _you_ made in the commits (as opposed to the codestyle changes that Black carried out)?  Is it just:
https://github.com/mozilla/addons-server/pull/12776/files#diff-522adf759addbd3b193c74ca85243f7d
and
https://github.com/mozilla/addons-server/pull/12776/files#diff-84efda180740ca51fff9fc1935f931e2 ?

How would this work exactly - we'd get a warning if we ran flake8 locally?  (It needs to work via the `make lint-codestyle` command too).  How would commits be automatically formatted and/or validated against Black compliance before merging?

EnTeQuAk:
> Can you point to the changes you made in the commits (as opposed to the codestyle changes that Black carried out)?

Only 4c74335 has been made by black, the rest is my handwriting.

> How would this work exactly - we'd get a warning if we ran flake8 locally? (It needs to work via the make lint-codestyle command too). How would commits be automatically formatted and/or validated against Black compliance before merging?

Correct, it's fully integrated into `flake8` so that all the existing tooling will work just as you're used to.

Running it locally will just be doing `black src/` - or using one of the many (https://black.readthedocs.io/en/stable/editor_integration.html)[Editor integrations]. Maybe I should add something to our local development docs to point contributors and developers into that direction?

EnTeQuAk:
I added some more configuration and ran black again after rebasing on master via 
b79c9e3 

So, to review - my commits are: bd4678e, 4cff3cf, afb2a62, 973f59c

eviljeff:
yeah I think we should mention it somewhere in the docs.  It'd be `src/ services/ tests` like flake8 too, I assume?  

I think I'd prefer a make command for it so we could just do `make black` (can we rename `lint-codestyle` to just `codestyle` or `lint` while we're at it - I can rarely correctly remember hyphenated names)

willdurand:
Very cool. I started to use `Black` locally (using a vim plugin) and shipped a few patches formatted with it. It'd be handy if we could end up with a single commit to apply Black on all the files in this project, so that it is [easy to exclude in `git blame`](https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame).

eviljeff:
maybe we should take the chance to increase max line length (90? 100?) to reduce the number of extra lines?

diox:
Black defaults to 88 characters per line, so maybe we should adopt that.

EnTeQuAk:
> Black defaults to 88 characters per line, so maybe we should adopt that.

I'm running this right now and will push it in a bit (after my 1:1 and after fixing all the conflicts), it does look fairly nice with that.

I just remembered that we kinda talked about not going too wide during one of the all hands and that's why I sticked to the 79 for now.

EnTeQuAk:
Alrighty, I refactored this a bit. It's now looking at line-length of 88 characters, so commits to reivew:

* cd07cec, 4c0e0eb, 37787ed, 7a9eea8, 3257db1, 36e79ae

EnTeQuAk:
Oops, some.of the excludes didn't work, maybe I messed up during rebase. I'll fix that

eviljeff:
flake8 fails in travis.  Does some of the code need manually fixing?

EnTeQuAk:
```
[git:8617-black-formatter (olympia)] ~/Projects/olympia> black src/olympia/reviewers/utils.py
All done! ✨ 🍰 ✨

[git:8617-black-formatter (olympia)] ~/Projects/olympia> flake8 src/olympia/reviewers/utils.py
src/olympia/reviewers/utils.py:714:89: E501 line too long (92 > 88 characters)
```

probably… although I don't really like that this even happens :imp:

EnTeQuAk:
Going to close it for now, going to take another look on a Friday

PR #12788, opened 2019-11-04
allow latest recommended version to be disabled if previous is okay

### LINKED ISSUE #12432: Cannot disable/delete latest version of Recommended Add-on
I released a version of my recommended add-on, FoxyProxy, with a major bug. I was unable to disable or delete it.

Steps:

1. Visit https://addons.mozilla.org/en-US/developers/addon/foxyproxy-standard/versions
2. Under **Listed Versions**, click the Delete/Disable "X" button in the same row as **Currently on AMO** Version 7.1, Status Approved.

3. Experience the message,

> The latest approved version of a Recommended extension cannot be deleted or disabled. Please contact AMO Admins if you need help with this

4. Hope and pray that your can fix the bug, upload new version, and find an admin reviewer to approve the new version really, really fast.

### What happened?

I could not stop hundreds of thousands of FoxyProxy users from automatically installing an update with a major bug.

I expected to be able to disable or delete the buggy version in order to spare my users headaches.

Read more Hemingway, and don't spell his name with two Ms when typing it.

eviljeff:
fixes #12432

### DIFF
--- a/src/olympia/versions/models.py
+++ b/src/olympia/versions/models.py
@@ -635,6 +635,17 @@ class Version(OnChangeMixin, ModelBase):
             for name, background in utils.get_background_images(
                 file_obj, theme_data=None, header_only=header_only).items()}
 
+    def can_be_disabled_and_deleted(self):
+        addon = self.addon
+        if self.recommendation_approved and self == addon.current_version:
+            qs = addon.versions.exclude(id=self.id)
+            previous_version = qs.latest() if qs.exists() else None
+            previous_approved = (
+                previous_version and previous_version.recommendation_approved)
+            if addon.is_recommended and not previous_approved:
+                return False
+        return True
+
 
 def generate_static_theme_preview(theme_data, version_pk):
     """This redirection is so we can mock generate_static_theme_preview, where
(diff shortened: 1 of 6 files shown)


src/olympia/versions/models.py
```diff
@@ -635,6 +635,17 @@ def get_background_images_encoded(self, header_only=False):
             for name, background in utils.get_background_images(
                 file_obj, theme_data=None, header_only=header_only).items()}
 
+    def can_be_disabled_and_deleted(self):
+        addon = self.addon
+        if self.recommendation_approved and self == addon.current_version:
+            qs = addon.versions.exclude(id=self.id)
+            previous_version = qs.latest() if qs.exists() else None
+            previous_approved = (
+                previous_version and previous_version.recommendation_approved)
```
diox: mmm, there are 2 things that strike me as potentially problematic here:
- Not filtering on `channel`. While unlisted versions should never have `recommendation_approved` set currently, it would be good to make sure now that we are future proof.
- More worryingly, this cannot return a version that is already deleted because you're using `addon.versions`, but this could return a version that is already disabled.

eviljeff: > this cannot return a version that is already deleted because you're using addon.versions, but this could return a version that is already disabled.

not returning deleted versions is okay (and what we want); disabled versions should probably be excluded.

diox: yeah my sentence was unclear: we clearly don't want deleted versions, I was highlighting the fact that `addon.versions` buys you this, but not excluding disabled ones.

PR #12793, opened 2019-11-04
Remove dogwrap

bqbn:
This is to get ready to disable the datadog agent in -prod.

* [X] Add a description of the changes introduced in this PR.

### DIFF
--- a/scripts/crontab/gen-cron.py
+++ b/scripts/crontab/gen-cron.py
@@ -26,13 +26,8 @@ def main():
     if not opts.deprecations:
         opts.python += ' -W ignore::DeprecationWarning'
 
-    dogwrap_path = '/usr/local/bin/amo_cron_dogwrap'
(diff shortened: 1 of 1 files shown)


scripts/crontab/gen-cron.py
```diff
@@ -26,13 +26,8 @@ def main():
     if not opts.deprecations:
         opts.python += ' -W ignore::DeprecationWarning'
 
-    dogwrap_path = '/usr/local/bin/amo_cron_dogwrap'
+    ctx = { "django": "cd %s; %s %s manage.py" % (opts.zamboni, opts.python) }
```
eviljeff: This doesn't appear valid python - the string still has 3 substitutions but now only 2 values have been provided

bqbn: Oops, I thought I removed that. 😝 

Thank you for catching this!

PR #12800, opened 2019-11-05
Associate customs lint rules with scanner rules

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12803

---

This patch updates the `extract_rule_names()` method to support `customs`. By doing this, we can associate scanner rules with matched customs rules.

In order to try it, you should either use the latest version of customs locally **or** you update a known scanner result to specify `matchedRules: ["some-name"]`. You will have to register the scanner rules beforehand.

### DIFF
--- a/src/olympia/scanners/models.py
+++ b/src/olympia/scanners/models.py
@@ -54,9 +54,9 @@ class ScannerResult(ModelBase):
     def extract_rule_names(self):
         """This method parses the raw results and returns the (matched) rule
         names. Not all scanners have rules that necessarily match."""
-        if self.scanner is YARA:
+        if self.scanner == YARA:
             return sorted({result['rule'] for result in self.results})
-        if self.scanner is CUSTOMS and 'matchedRules' in self.results:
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/models.py
```diff
@@ -48,10 +54,12 @@ def add_yara_result(self, rule, tags=None, meta=None):
     def extract_rule_names(self):
         """This method parses the raw results and returns the (matched) rule
         names. Not all scanners have rules that necessarily match."""
-        if self.scanner is not YARA:
-            # We do not have support for other scanners yet.
-            return []
-        return sorted({result['rule'] for result in self.results})
+        if self.scanner is YARA:
```
diox: Had not spotted that before, but don't use `is` for integer value comparisons, use `==`. `is` is for identity comparisons, the fact that it works for integers up to 256 is an implementation detail.
```python
>>> a = 1
>>> a is 1
True
>>> a = 257
>>> a is 257
False
```

willdurand: oh I see, thanks

Issue #12767, opened 2019-10-31
Drop Version.version_int


eviljeff:
`Version.version_int` is unused and a complication if we want to change how it's calculated (see #6016).  Lets drop the property from the model.

PR #12805, opened 2019-11-05
migration to drop Version.version_int column

eviljeff:
followup to #12767 pr that drops the column from the db for the removed model field.

Must be merged in a subsequent tag/milestone to 2019.11.07


src/olympia/versions/migrations/0003_drop_version_int_column.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 2.2.6 on 2019-10-30 20:20
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('versions', '0002_version_needs_human_review.py'),
```
EnTeQuAk: You only need the module name, that's why all tests fail. Skip the `.py`

Issue #12749, opened 2019-10-29
Execute the `run_action` task


willdurand:
In #12676, we added a `run_action` celery task that takes a `version_id` as input, retrieves the action to execute (if any) and then execute this action. We should now run this task.

Note: only two actions are available for now:

- `no_action`, which does nothing
- `flag_for_human_review`, which is not implemented yet

AlexandraMoga:
@willdurand are there any specifics to be tested or should I check for regressions only?

willdurand:
> @willdurand are there any specifics to be tested or should I check for regressions only?

You can test the following workflow:

1. configure a rule like the one starting with "D" with action "flag"
2. upload an add-on that is matched by this "D" rule
3. look at the new queue in the reviewer tools

AlexandraMoga:
@willdurand I've verified this on -dev by  enabling the `run-action-in-auto-approve` switch and setting a "flag" action to various scanner rules.

Submitted add-ons that matched any of the flagged rules, will be marked for human review in Rev Tools.
*Note: this applies only to listed versions currently because of #12622 

Example:
[This scanner result](https://addons-internal.dev.mozaws.net/en-US/admin/models/scanners/scannerresult/4981/change/) was automatically flagged for human review - see [add-on review page](https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/ext-with-scanner-rule-5)

willdurand:
> @willdurand I've verified this on -dev by enabling the `run-action-in-auto-approve` switch and setting a "flag" action to various scanner rules.

ah good call, thanks! I forgot to enable the feature in -dev last week..

> *Note: this applies only to listed versions currently because of #12622

Indeed, but we might get this issue fixed this week too!

PR #12806, opened 2019-11-05
Execute run_action() in auto_approve

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12749

---

I added a waffle switch so that we can decide when to turn the feature on and also have a kill switch if needed..


willdurand:
D'oh! The migration fails in most travis jobs because ["waffle is not an app"](https://travis-ci.org/mozilla/addons-server/jobs/607706219). It worked for me locally, though..

PR #12807, opened 2019-11-05
Implement "Delay auto-approval by 24 hours/indefinitely" action

### LINKED ISSUE #12671: Implement "Delay auto-approval by x" action
See also  #12670 which implements another base action that we'll probably want to do first.

"Delay auto-approval by 24 hours" and "Delay auto-approval indefinitely" are two actions we want to implement in the workflow of the scanners. When triggered, each of these actions should:
- Prevent future auto-approval of the add-on (including the version that triggered that action)
- Flag the entire add-on for human review in a separate queue (maybe re-using https://github.com/mozilla/addons-server/issues/12670)

As #12670 reviewers can revert this, but for now that shouldn't prevent scanners from reflagging the add-on for the same reason after a new upload.

Implementation notes:
- Add new `DateTimeField` in `AddonReviewerFlags` called `auto_approval_disabled_until`.
- Scanner action would set a datetime in the future accordingly. "Indefinitely" is implemented using the same field with a date in the far future, separate from the existing `auto_approval_disabled` which is set by reviewers.

If not re-using the flag from #12670 we'll need a separate queue to view flagged add-ons in reviewer tools. If re-using it, we could display the date where auto-approvals would resume for this add-on in that queue.

Everything should apply to both listed & unlisted, but https://github.com/mozilla/addons-server/issues/12622 is a blocker for unlisted so we may start with listed only.

Depends on #12640

diox:
This is separate from `auto_approval_disabled` that can be set by admin reviewers for now, because scanners forcibly set it on the add-on as soon as a version matching a rule with the corresponding action is found.

Also includes a refactor of `AutoApprovalSummary` stolen from #12658 - it didn't turn out to be needed but it's nice to have it to make the code around this easier to understand and modify.

Fixes #12671


src/olympia/addons/models.py
```diff
@@ -1627,6 +1639,8 @@ class AddonReviewerFlags(ModelBase):
     needs_admin_content_review = models.BooleanField(default=False)
     needs_admin_theme_review = models.BooleanField(default=False)
     auto_approval_disabled = models.BooleanField(default=False)
+    auto_approval_disabled_until = models.DateTimeField(
```
eviljeff: `auto_approval_delayed_until`? I think that would be more consistent.

src/olympia/reviewers/models.py
```diff
@@ -225,7 +225,10 @@ def base_query(self):
         query['where'].append(
             f'((addons.addontype_id IN ({types}) '
             'AND files.is_webextension = 0) '
-            'OR addons_addonreviewerflags.auto_approval_disabled = 1)')
+            'OR addons_addonreviewerflags.auto_approval_disabled = 1 '
+            'OR addons_addonreviewerflags.auto_approval_disabled_until > NOW()'
```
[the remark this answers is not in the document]
diox: Yes. The flag isn't necessarily cleared, it could stay at an old date, so if the date it's set to is before `NOW` then we don't need to take it into account.

src/olympia/addons/models.py
diox: I like that. eefefa57c1e3851a0503e3fd3592829d0718335e

PR #12813, opened 2019-11-06
Fix needs_human_review flag clearing when rejecting multiple versions

diox:
Follow-up fix for https://github.com/mozilla/addons-server/issues/12670

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -859,10 +859,11 @@ class ReviewBase(object):
             self.set_files(amo.STATUS_DISABLED, files, hide_disabled_file=True)
             self.log_action(action_id, version=version, files=files,
                             timestamp=timestamp)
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/utils.py
```diff
@@ -859,10 +859,11 @@ def reject_multiple_versions(self):
             self.set_files(amo.STATUS_DISABLED, files, hide_disabled_file=True)
             self.log_action(action_id, version=version, files=files,
                             timestamp=timestamp)
-        # Unset needs_human_review on those versions, we consider that the
-        # reviewer looked at them.
-        if self.human_review:
-            self.data['versions'].update(needs_human_review=False)
```
diox: I moved this and changed the code to update each version separately because the queryset is in a weird state after going through this loop - it's as if it's empty as soon as you try to act on it.

PR #12816, opened 2019-11-06
edit addon block, with ActivityLog logging

### LINKED ISSUE #12464: Edit add-on block

When a block already exists for an add-on, the individual add-on block page should allow admins to edit that block. Note that this restricts add-ons to only having one active block at a time.

The edit page should also display the blocklist history for that add-on, with all of the details (link, reason, time, who made the change, and whether the block applies to old clients).

eviljeff:
fixes #12464


src/olympia/blocklist/admin.py
```diff
@@ -155,31 +158,76 @@ def review_unlisted(self, obj):
...
+
+        logs = ActivityLog.objects.for_addons((obj.addon,)).order_by('created')
+        log_entries_gen = (
+            (log.created.date(),
+             log.user.name,
+             str(log),
+             log.details['min_version'],
+             log.details['max_version'],
+             log.details['reason'])
+            for log in logs)
```
diox: Am I missing something ? You are not filtering on `action`, so you're going to get a bunch of random `ActivityLog` objects that are irrelevant to this, and would not have the keys you're looking for in `log.details`.

eviljeff: ugh, you're right.  It worked locally!  (when I was blocking the addons created during initialization that wouldn't have any other ActivityLog records...)

diox: Does fail locally with `KeyError` because of this.

eviljeff: https://github.com/mozilla/addons-server/pull/12816/commits/be1dec6bfa4cf4cc8f951e3bd7d20c77beff8f78

Issue #12514, opened 2019-10-02
drop is_source_public property from Addon api


eviljeff:
depends on #12418 

quoting @diox:
> The functionality is also exposed in the API under a different name, is_source_public. This needs special care as the API must keep this field for backwards-compatibility purposes. You'll need to modify the declaration of the field in src/olympia/addons/serializers.py to always return False. I suggest making it a serializerMethodField() instead of the BooleanField it is now and implementing a get_is_source_public(self, obj) in the serializer class, simply having it return False. Tests will have to be updated here as well.

ash-13:
Hey @eviljeff I would like to work on it , can i take this up?

eviljeff:
@ash-13 you can work on it.  Though you'll need an understanding of Django rest framework and you will need to look yourself how our `DRF_API_GATES` implementation provides shimming/removing properties from different versions of the API.  It's not a good first bug.

ash-13:
@eviljeff Okay I will start django , then I will start good first but
Then guide me !

junngo:
@eviljeff 
I am going to modify it according to diox guide. And It will be modified so that it can be returned only from v3 api.
If It is right, Can I work on it?

eviljeff:
@junngo go ahead.

eviljeff:
QA: check the API response doesn't contain the `is_source_public` property in /api/v4; but does in api/v3 - where it should always be False.

ioanarusiczki:
@eviljeff I verified on dev for
for https://addons-dev.allizom.org/api/v4/addons/addon/502872/
`is_source_public` is no longer available 

for https://addons-dev.allizom.org/api/v3/addons/addon/502872/
`is_source_public` is false

![source public - false](https://user-images.githubusercontent.com/33448286/68777984-2dd51d80-063b-11ea-9739-dcdcff21303c.png)

PR #12824, opened 2019-11-07
drop is-source-public field except for v3 api

junngo:
#12514 
I modified so that is_source_public field can be returned only from v3 api. I checked to run test case and If it work right on my locally.
Let me know if there's anything wrong with a misprint or something.

### DIFF
--- a/docs/topics/api/overview.rst
+++ b/docs/topics/api/overview.rst
@@ -364,6 +364,7 @@ v4 API changelog
 * 2019-08-22: added ``canned_response`` property to draft comment api. https://github.com/mozilla/addons-server/issues/11807
 * 2019-09-19: added /site/ endpoint to expose read-only mode and any site notice.  Also added the same response to the /accounts/account/ non-public response as a convenience for logged in users. https://github.com/mozilla/addons-server/issues/11493)
 * 2019-10-17: moved /authenticate endpoint from api/v4/accounts/authenticate to version-less api/auth/authenticate-callback https://github.com/mozilla/addons-server/issues/10487
+* 2019-11-14: removed ``is_source_public`` property from addons API https://github.com/mozilla/addons-server/issues/12514
 
 .. _`#11380`: https://github.com/mozilla/addons-server/issues/11380/
 .. _`#11379`: https://github.com/mozilla/addons-server/issues/11379/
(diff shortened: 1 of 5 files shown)


src/olympia/addons/tests/test_serializers.py
```diff
@@ -354,10 +353,17 @@ def test_is_disabled(self):
         assert result['is_disabled'] is True
 
     def test_is_source_public(self):
-        self.addon = addon_factory(view_source=True)
+        self.request.version = 'v3'
+        self.addon = addon_factory()
         result = self.serialize()
```
eviljeff: You shouldn't need to set `self.request.version` to activate the shim - that's what the `override_settings` later on is for.

src/olympia/addons/tests/test_serializers.py
```diff
@@ -354,10 +353,17 @@ def test_is_disabled(self):
...
 
-        assert result['is_source_public'] is True
+        assert result['is_source_public'] is False
```
eviljeff: so, with no shim the `is_source_public` **shouldn't** be present.

src/olympia/addons/tests/test_serializers.py
```diff
@@ -354,10 +353,17 @@ def test_is_disabled(self):
...
+
+        # It's present in v3
+        gates = {None: ('is-source-public-shim',)}
+        with override_settings(DRF_API_GATES=gates):
+            result = self.serialize()
+            assert 'is_source_public' not in result
```
eviljeff: and with `override_settings(...)` the field **should** be present in `result`

junngo:
Thanks for reviewing :) I submitted PR again.
If you have any other ideas, please let me know.

docs/topics/api/overview.rst
```diff
@@ -314,6 +314,7 @@ these docs.
 v4 API changelog
 ----------------
 
+* 2019-11-14: removed ``is_source_public`` property from addons API https://github.com/mozilla/addons-server/issues/12514
```
eviljeff: If you look at dates you can see they're in descending order, so the new entry should be at the end of the list rather than the top

junngo: Oh, Sorry.
I submit patch again.

PR #12825, opened 2019-11-07
Buttons made visible in DevHub mobile view

### LINKED ISSUE #12594: [DevHub] Submit a New Add-on and Theme buttons are not visible on mobile logged in view
1. Log in to the Developer Hub with a user that has some add-on submitted
2. Resize your screen to mobile view
3. Look for the 'Submit a New Add-on' and 'Submit a New Theme' buttons

### What happened?
The buttons are not visible

The buttons should be visible

- the likelihood for users to submit add-ons from a mobile device is small, but we should aim for the same content on desktop and mobile
- furthermore, if a user without submitted add-ons is viewing the page, the buttons are present 

[logged in view - user with addons - buttons not visible]
image

[logged in view - user without addons - buttons are visible]
image

kunalmohan:
Fixes #12594

New Add-on and Theme buttons were not visible in mobile view.
The issue is corrected by removing display none property and
changing flex direction of container to column for device-
width less than 960px.

before-
Screenshot from 2019-11-07 15-10-25

after-
Screenshot from 2019-11-07 15-08-31

### DIFF
--- a/static/css/devhub/new-landing/base.less
+++ b/static/css/devhub/new-landing/base.less
@@ -3,6 +3,7 @@
 
 // Grid
 @grid-max: 960px;
+@small: ~"only screen and (min-width: 550px)";
 @medium: ~"only screen and (min-width: 700px)";
 @large: ~"only screen and (min-width: 960px)";
 @notlarge: ~"only screen and (max-width: 959px)";
--- a/static/css/devhub/new-landing/base.less
+++ b/static/css/devhub/new-landing/base.less
@@ -3,7 +3,6 @@
 
 // Grid
 @grid-max: 960px;
-@small: ~"only screen and (min-width: 550px)";
 @medium: ~"only screen and (min-width: 700px)";
 @large: ~"only screen and (min-width: 960px)";
 @notlarge: ~"only screen and (max-width: 959px)";
--- a/static/css/devhub/new-landing/sections/my-addons.less
+++ b/static/css/devhub/new-landing/sections/my-addons.less
@@ -121,7 +121,7 @@
   align-items: flex-start;
   justify-content: flex-start;
 
-  @media @large {
+  @media only screen and (min-width: 730px) {
     flex-direction: row;
   }
 }
@@ -140,8 +140,12 @@
     margin-right: auto;
     margin-left: 0;
   }
+}
 
-  @media @large {
-    display: inline;
+.DevHub-MyAddons-item-buttons-submit > .Button{
+  width: 100%;
+
+  @media @small{
+    width: auto;
   }
 }
--- a/static/css/devhub/new-landing/sections/my-addons.less
+++ b/static/css/devhub/new-landing/sections/my-addons.less
@@ -116,15 +116,6 @@
 
 .DevHub-MyAddons-item-buttons {
   margin-top: 20px;
-  display: flex;
-  flex-direction: column;
-  align-items: flex-start;
-  justify-content: flex-start;
-
-  @media @medium {
-    flex-direction: row;
-    flex-wrap: wrap;
-  }
 }
 
 .DevHub-MyAddons-item-buttons-all {
@@ -136,11 +127,19 @@
 .DevHub-MyAddons-item-buttons-submit {
 
   margin-left: auto;
-  align-self: flex-end;
-
+  float: right;
 
   html[dir=rtl] & {
     margin-right: auto;
     margin-left: 0;
+    float: left;
+  }
+}
+
+.DevHub-MyAddons-item-buttons-submit > .Button {
+  width: 100%;
+
+  @media @medium {
+    width: auto;
   }
 }
--- a/static/css/devhub/new-landing/sections/my-addons.less
+++ b/static/css/devhub/new-landing/sections/my-addons.less
@@ -121,8 +121,9 @@
   align-items: flex-start;
   justify-content: flex-start;
 
-  @media only screen and (min-width: 730px) {
+  @media @medium {
     flex-direction: row;
+    flex-wrap: wrap;
   }
 }
 
@@ -135,17 +136,11 @@
 .DevHub-MyAddons-item-buttons-submit {
 
   margin-left: auto;
+  align-self: flex-end;
+
 
   html[dir=rtl] & {
     margin-right: auto;
     margin-left: 0;
-  }
-}
-
-.DevHub-MyAddons-item-buttons-submit > .Button{
-  width: 100%;
-
-  @media @small{
-    width: auto;
   }
 }
--- a/static/css/devhub/new-landing/sections/my-addons.less
+++ b/static/css/devhub/new-landing/sections/my-addons.less
@@ -117,9 +117,13 @@
 .DevHub-MyAddons-item-buttons {
   margin-top: 20px;
   display: flex;
-  flex-direction: row;
+  flex-direction: column;
   align-items: flex-start;
   justify-content: flex-start;
+
+  @media @large {
+    flex-direction: row;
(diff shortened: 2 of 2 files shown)


kunalmohan:
I closed the request by mistake.

kunalmohan:
@kumar303 are there any issues with the PR because initially it did not pass some checks? Should I make any more changes?

Sorry for bringing this up so late :sweat_smile:

kumar303:
Hi @kunalmohan. This has been in my review queue but I've had to give my time to other prioritized features. My apologies. I hope to get to it soon. Thanks for taking time to submit a patch.

kunalmohan:
Yeah, I could do that. I'll wait for your confirmation.

eviljeff:
I don't think it matters too much, but what @bobsilverberg said seems fine.  

Another option, if we think keeping them under one another at widths all the way up to 960px is too extreme, is to change the `@media` to `@media @medium` in `.DevHub-MyAddons-item-buttons` and `.DevHub-MyAddons-item-buttons-submit` (I've not experimented to see how this looks)

kunalmohan:
These are ss according to what @eviljeff is suggesting:
800px:
800px


715px:
715px

650px:
650px


For <550px, it is the same as what I have implemented above.

muffinresearch:
> These are ss according to what @eviljeff is suggesting:
> 800px:
> 800px
> 
> 715px:
> 715px
> 
> 650px:
> 650px
> 
> For <550px, it is the same as what I have implemented above.

Hi @kunalmohan This looks pretty good, if it's possible to make the 715px view match the layout shown in the 650px one that would be better and would avoid having the buttons floating in mid air. We'll also want to check that the state represented in the 715px above doesn't appear at some other point when reducing the width once that nit is fixed.

For < 550px the only other change I'd suggest is making the buttons wider to be 100% of the viewport width (less the default gutter either side). 

Let me know if you have any additional questions.

kunalmohan:
@muffinresearch I made the required changes. I had to introduce two new breakpoints, one of which I have introduced in the traditional CSS manner (I didn't want to make a new standard breakpoint that differs from an already defined one by just 30px).

bobsilverberg:
@kunalmohan Thanks for your continued efforts on this, and sorry for not getting back to you sooner. There shouldn't be a need to define a new breakpoint, nor to introduce a custom breakpoint. We should just try to make sure that things look good at the already established breakpoints.

kunalmohan:
Okay, I'll try to get the best results without defining any new breakpoints :sweat_smile:

kunalmohan:
> > These are ss according to what @eviljeff is suggesting:
> > 800px:
> > 800px
> > 715px:
> > 715px
> > 650px:
> > 650px
> > For <550px, it is the same as what I have implemented above.
> 
> Hi @kunalmohan This looks pretty good, if it's possible to make the 715px view match the layout shown in the 650px one that would be better and would avoid having the buttons floating in mid air. We'll also want to check that the state represented in the 715px above doesn't appear at some other point when reducing the width once that nit is fixed.
> 
> For < 550px the only other change I'd suggest is making the buttons wider to be 100% of the viewport width (less the default gutter either side).
> 
> Let me know if you have any additional questions.

I have corrected the display at 715px. And I don't think it would be possible to make the buttons wider for viewport <550px without defining a new breakpoint there.

bobsilverberg:
> > > These are ss according to what @eviljeff is suggesting:
> > > 800px:
> > > 800px
> > > 715px:
> > > 715px
> > > 650px:
> > > 650px
> > > For <550px, it is the same as what I have implemented above.
> > 
> > 
> > Hi @kunalmohan This looks pretty good, if it's possible to make the 715px view match the layout shown in the 650px one that would be better and would avoid having the buttons floating in mid air. We'll also want to check that the state represented in the 715px above doesn't appear at some other point when reducing the width once that nit is fixed.
> > For < 550px the only other change I'd suggest is making the buttons wider to be 100% of the viewport width (less the default gutter either side).
> > Let me know if you have any additional questions.
> 
> I have corrected the display at 715px. And I don't think it would be possible to make the buttons wider for viewport <550px without defining a new breakpoint there.

Thanks @kunalmohan. I see that the smallest breakpoint currently available is 700px. I was thinking you could define the buttons to be 100% width at <550px, as @muffinresearch suggested, but I also see that a breakpoint doesn't exist for that size. @muffinresearch can you suggest how that might be done without introducing a new breakpoint?

eviljeff:
The only sizes currently defined in devhub landing page are:
`<700px` (no `@media`, so "normal" size)
`>=700px` (`@medium`)
`>=960px` (`@large`)

Although it's possible to add extra breakpoints I spent some time tidying up unnecessary breakpoints previously so I'd prefer to stick with those tbh :)  Maybe we can have them max-width up to 700px?  Or just define a max-size directly on the button?

muffinresearch:
Looking at this I think it works better if you just remove the medium break-point change and make the `flex-direction: column;` for all breakpoints in `.DevHub-MyAddons-item-buttons`.

However, after doing this it really starts to remove the need for using flexbox, so instead you could simply ditch all the flex related properties on `.DevHub-MyAddons-item-buttons` and simply float right (left for RTL) `.DevHub-MyAddons-item-buttons-submit`. This then works better across multiple breakpoints.

After that the only change is to fix the button width to ~100%, and override that for larger viewports.

kunalmohan:
> Looking at this I think it works better if you just remove the medium break-point change and make the `flex-direction: column;` for all breakpoints in `.DevHub-MyAddons-item-buttons`.
> 
> However, after doing this it really starts to remove the need for using flexbox, so instead you could simply ditch all the flex related properties on `.DevHub-MyAddons-item-buttons` and simply float right (left for RTL) `.DevHub-MyAddons-item-buttons-submit`. This then works better across multiple breakpoints.
> 
> After that the only change is to fix the button width to ~100%, and override that for larger viewports.

Applying these changes, there is a range of viewport width which has the following layout:
abc
Here the width is 100% (for viewport smaller than `medium`).

For other viewports, layout looks good.

muffinresearch:
This still appears to be using flex-box. Note the part of the comment which refers to dropping that in favour of simple floats.

kunalmohan:
Nope, this does not use flexbox.
abcd

I suppose I could also remove all the margin properties from `.DevHub-MyAddons-item-buttons-submit` after using float.

muffinresearch:
> Nope, this does not use flexbox.
> abcd
> 
> I suppose I could also remove all the margin properties from `.DevHub-MyAddons-item-buttons-submit` after using float.

I still see this for the diff: 

Buttons_made_visible_in_DevHub_mobile_view_by_kunalmohan_·_Pull_Request__12825_·_mozilla_addons-server_-_Firefox_Nightly

Is a commit missing from the PR?

kunalmohan:
> Is a commit missing from the PR?

Yes, I had not pushed the latest changes. I have done so now.

muffinresearch:
Thanks for the patch @kunalmohan

caitmuenster:
Thanks so much for the patch, @kunalmohan! Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

Would you be interested in creating an account on [mozillians.org](https://mozillians.org)? I'd be happy to vouch for you!

kunalmohan:
@caitmuenster Yes, I'd love that :)

kunalmohan:
@caitmuenster could you vouch for me? https://mozillians.org/en-US/u/kunalmohan99/

caitmuenster:
@kunalmohan You're vouched! 🎉 Welcome onboard! I look forward to seeing you around the project.

PR #12832, opened 2019-11-08
replace addon fk on Block model with guid field

### LINKED ISSUE #12815: Store guid in Block model and rely on it as the source of truth
When the Block model was implemented for #12454 it was implemented with an fk to Addon.  Which works great until the developer deletes the Addon instance and we still want the GUID to not change and refer to the new Addon.  So instead the Block needs to store the GUID directly and lookup the Addon instance from the guid.

eviljeff:
fixes #12815

tbh, still not sure if we should be indexing ActivityLog based on GUID or Block record.  (the former is what we normally do; the latter would work even if the Block was deleted)


src/olympia/blocklist/admin.py
```diff
@@ -229,33 +235,36 @@ def delete_model(self, request, obj):
...
     def get_form(self, request, obj=None, **kwargs):
-        if obj:
-            self.addon_instance = obj.addon
-        else:
-            self._guid = request.GET.get('guid')
-            self.addon_instance = Block(guid=self._guid).addon
+        # Grab the obj here so we can reference it in formfield_for_db_field,
+        # or create a new Block instance otherwise.
+        request._obj = (
+            Block(guid=self.get_request_guid(request)) if not obj else obj)
```
diox: I wonder if we could set the choices in this method (below, after `super().get_form(...)` to avoid setting `_obj` on `request` ?

eviljeff: After a few hours of looking through the spaghetti soup of code in django admin model code, django forms, and django fields, I don't have an answer to that.  It certainly _seems_ like that should be possible, if one discovers the right method to hook into.

eviljeff: https://github.com/mozilla/addons-server/pull/12832/commits/7d3f0b97434bc60ca00cfa86962ed39284dbe1b2

diox: My idea was to set the choices without using `formfield_for_dbfield` at all - letting django create the form (calling `super().get_form(...)`) and *then* modify the choices in the returned `form`. Is that not possible ?

eviljeff: like I was saying, it seems like it should be possible, if one finds the right way of doing it.

eviljeff: it's a question of how many (more) hours is to be spent trying to get it to work

eviljeff: We need some way of forcing the field to be a `ChoiceField`, and setting the choices.  Forcing the field to have choices in the model causes django migration problems (and it's difficult to get the correct choices at the correct point when they depend on a query based on the value of another field).  I tried a lot of variations, but at some point you have to stop (and I was already double the time limit for the timebox I set myself)

diox: My naive implementation : https://gist.github.com/diox/5cd0f93cfcc14f9dc4d7e5b20bb7c3c0

It seems to work locally and pass the tests - but maybe more tests are needed.

eviljeff: I had a thought while I was eating just now that something like that might work (if I could remember what the fields property was) - thanks for doing the legwork!

Issue #12851, opened 2019-11-12
Add a state field on the `ScannerResult`


willdurand:
We need a new `state` field to be able to triage scanner results. We'll have 3 choices for now:

- unknown (default)
- true positive (scanner result reported something of interest for us
- false positive

ioanarusiczki:
@willdurand on dev with FF70(Win10):
 - The `state` field is available in the scanner results table and it can be `unknown`, `false positive` or `true positive`. 

I see the number of results at the bottom of the page. The Filter Menu is at the top. When I change the filter By state `All` to `Unknown` or `true positive` I don't see the number of items updates unless I scroll down. It would be better to see it at the top of the page. 
Let me know what you think. 

![number of results and paging](https://user-images.githubusercontent.com/33448286/69226293-931f9600-0b88-11ea-98c9-01696c0fd737.png)

willdurand:
> I see the number of results at the bottom of the page. The Filter Menu is at the top. When I change the filter By state `All` to `Unknown` or `true positive` I don't see the number of items updates unless I scroll down. It would be better to see it at the top of the page.
> Let me know what you think.

This is the default "django admin" layout. Maybe we could change it but it is like that for all other admin pages.

PR #12852, opened 2019-11-12
Add state field to ScannerResult to record feedback

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12851

---

This patch also contains two new methods on the `ScannerResult` class that I will need in an upcoming patch.


src/olympia/scanners/migrations/0015_auto_20191112_1413.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dependencies = [
+        ('scanners', '0014_alter_action_choices_on_scannerrule'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='scannerresult',
+            name='state',
+            field=models.PositiveSmallIntegerField(choices=[(0, 'Unknown'), (1, 'True positive'), (2, 'False positive')], default=0),
```
diox: Django doesn't add database defaults in column definitions (historical design decision, I believe because they want to keep control over the values in python). 

When deploying there will be a small window of time when we are running the old code from previous week tag on the new, updated database schemas from the newer tag. So, if you add a field that doesn't accept null values, the old code will fail when inserting new values as it doesn't know about the new column and it doesn't have a database default.

The moral of the story is: if we want proper backwards-compatibility without having to disable/re-enable scanners are deploy time, you want to make this field accept null values. 

You can either make this value one of the choices (something I've done in new fields for `AbuseReport`, using `null=True, blank=True, default=None`) or make it a separate temporarily accepted value you then migrate away from in a future deploy, but that's more extra work later.

PR #12854, opened 2019-11-12
Add actions to report feedback on ScannerResult

### LINKED ISSUE #12853: Add actions to report feedback on the scanner results
We'd like to report feedback on the scanner results. We have two possibilities:

- false positive: scanner result was incorrect
- true positive: scanner result was correct

When marking a result as true positive, we'll simply update its state (see #12851) and return to the list.

When reporting a result as false positive, we'll create a GitHub issue _draft_ with some relevant information after having updated the state of the result.

willdurand:
Fixes #12853

---

This patch adds the following features:

- mark a scanner result as true positive
- report a scanner result as false positive
- list view hides triaged scanner results by default (only showing results with `state = UNKNOWN`)

I ignored the WAT scanner for now. Its collected data cannot really be true or false positive. We might need a new state value in the future to mark a result as "handled only".

### DIFF
--- a/src/olympia/scanners/models.py
+++ b/src/olympia/scanners/models.py
@@ -1,5 +1,6 @@
 import json
 
+from django.conf import settings
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 from django_extensions.db.fields.json import JSONField
@@ -12,6 +13,7 @@ from olympia.constants.scanners import (
     RESULT_STATES,
     SCANNERS,
     UNKNOWN,
+    WAT,
     YARA,
 )
 from olympia.files.models import FileUpload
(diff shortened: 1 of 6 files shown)


src/olympia/scanners/models.py
```diff
@@ -33,6 +39,10 @@ class ScannerResult(ModelBase):
     matched_rules = models.ManyToManyField(
         'ScannerRule', through='ScannerMatch'
     )
+    state = models.PositiveSmallIntegerField(
```
diox: If it was a NullBooleanField instead, you'd get a fancy green/red tick for free instead of the quite verbose text in the admin (which you can replicate manually by using a method + setting `<methodname>.boolean = True`, but might as well use a true boolean field)

Maybe it's worth exploring, unless you think we're going to get more states than just `true`/`false`/`null` ?

willdurand: I thought about using a nullable boolean field but @wagnerand and I discussed and we thought that we might have more states in the future.

willdurand:
I just rebased this patch.

src/olympia/scanners/models.py
```diff
@@ -88,6 +90,17 @@ def get_scanner_name(self):
...
         return json.dumps(self.results, indent=2)
 
+    def can_report_feedback(self):
+        return (
+            self.has_matches and self.state == UNKNOWN and self.scanner != WAT
+        )
+
+    def get_git_repository(self):
+        return {
+            CUSTOMS: settings.CUSTOMS_GIT_REPOSITORY,
```
diox: I don't think we need it yet, but worth keeping in mind: at some point we may want to start defining a class for each scanner with methods like `run()`, `get_git_repository()`...

PR #12863, opened 2019-11-13
Display auto approval delay as a flag in reviewer queues

### LINKED ISSUE #12849: Reviewer queues should show if the add-on approvals are being delayed
In https://github.com/mozilla/addons-server/issues/12833#issuecomment-552401824 @wagnerand wrote:
> Yes! It would be good if the queue would somehow differentiate between no delay, 24 hours and indefinitely.

Let's add flags for those to all queues as flags, it's the easiest way to show this. We're already fetching the model that has this information anyway.

diox:
Fixes #12849

Add-on with auto-approval temporary delayed:
flags-temporarily

Add-on with auto-approval indefinitely delayed:
flags-indefinitely


diox:
doh, I forgot to handle the special Flagged by scanners queue! Need to add that.

diox:
All this time designing icons in gimp wasted :stuck_out_tongue: 

Given that addons-server is licensed under [BSD-3](https://github.com/mozilla/addons-server/blob/master/LICENSE) I didn't want to re-use 3rd-party icons under a different license, because I didn't want to go through the hassle of specifying a different license for those icons in particular (especially as they are part of a sprite that's not even automatically generated...).

Still, I do see the point about the lack of contrast. I like using an infinity symbol for the "Delayed indefinitely" flag, I'll look into creating a new icon following that idea.

diox:
> Also, I don't think it is related to this patch but I noticed that the reviewer detail page does does not mention that it has not been auto-approved because of the delay:

Yeah, that's because I re-used the same check we have for auto-approval disabled in `AutoApprovalSummary` instead of bothering with a new one. I figured it wasn't that big of a deal, the important point is that the auto-approval was disabled for this version. Affected versions will also get the "Flagged by scanner" highlighting so I think we're fine.

diox:
New icon:

new-icons

PR #12881, opened 2019-11-15
add featured images

### LINKED ISSUE #12818: Add primary hero shelf featured images (again)
Duplicate of #12184 

The primary hero shelf currently only allows a selection of 2 placeholder images. We need a wider selection, as per the PRD, for launch.

Ideally they can be attached to this issue.

eviljeff:
fixes #12818


eviljeff:
It was just because it was only useful for migrations, rather than live code (like how we have all the test util functions in `amo/tests/__init__.py`)

PR #12885, opened 2019-11-15
include ActivityLog by guid too so logs for deleted Blocks persist

### LINKED ISSUE #12840: Block history not populated if the blocked add-on has a previous deleted block
Follow up for #12464 

1. Log in to AMO with an admin account
2. Open the Blocklist module in django admin
3. Add a new add-on to the blocklist
4. Delete the add-on block
5. Add the same add-on to the blocklist again
6. Observe the Block history after you Save your changes

### What happened?
Block history is not populated

Block history is populated

Here is an example: https://addons-internal.dev.mozaws.net/en-US/admin/models/blocklist/block/17/change/
image

eviljeff:
fixes #12840


src/olympia/activity/migrations/0004_auto_20191115_1154.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('activity', '0003_auto_20191115_1151'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='blocklog',
+            name='guid',
+            field=models.CharField(max_length=255),
```
diox: Can't do that yet. Splitting into 2 migrations that run in the same deploy window doesn't solve the problem of having potentially `BlockLog` entries being created by old code after migrations have ran but before the push is done.

eviljeff: It's fine, it's fine 😉 It's not in use yet by anyone so unless QA are unlucky enough to be using django admin during the deploy window it won't affect anyone.

diox: Fair enough. In that case though, you don't need 2 migrations - you can set a `default` in the migration when adding the field.

eviljeff: It was a separate migration so the RunPython can populate the null values before the column is changed to be actually `null=False` (or it fails db integrity checking when the column is added). 
Or are you suggesting setting it to some other non-null value?

diox: Just an empty string ?

eviljeff: https://github.com/mozilla/addons-server/pull/12885/commits/c22bf522fc758f78aa45207927fe472fb1a454eb

Issue #12896, opened 2019-11-18
Unlisted submissions with a delay auto-approval flag are still auto-approved shortly


AlexandraMoga:
Follow up for #12622 and #12671 

### Describe the problem and steps to reproduce it:
1. Log in to AMO -dev
2. Submit a new unlisted extension that contains a [matched rule](https://addons-internal.dev.mozaws.net/en-US/admin/models/scanners/scannerrule/) configured with a Delay auto-approval action (24h or indefinite)
3. Check the add-on status after a few minutes

### What happened?
The add-on is approved

### What did you expect to happen?
The add-on is still in an awaiting review state

### Anything else we should know?
- looking at the [Flagged by scanners](https://reviewers.addons-dev.allizom.org/en-US/reviewers/queue/needs_human_review) queue, the add-ons will appear in the list with the corresponding delay flags:
![image](https://user-images.githubusercontent.com/31961530/69063086-88e48700-0a24-11ea-84be-d7c7a270f7b8.png)
- both add-ons have a a scanner result with a corresponding matched rule

eviljeff:
I don't know if we're relying on these rules working yet. If we are its a P1, otherwise I dunno @diox @willdurand

diox:
We aren't yet, especially for unlisted, but I'll take a look.

diox:
The `auto_approval_delayed_until` field was set properly on the `AddonReviewerFlags` instance, and the `AutoApprovalSummary` instance was created immediately after, but somehow when it ran `check_has_auto_approval_disabled()` it didn't see the delay.

I suspect somehow the `AddonReviewerFlags` instance had already been fetched and cached, as `OneToOneField` are, although I can't find what would have triggered that - it would have needed to be accessed first, but AFAIK that doesn't happen... I'll investigate.

AlexandraMoga:
Verified fixed on -dev. 

Unlisted submissions that matched a rule with a delay auto-approval action will no longer be auto-approved shortly after submission. Those with a 24h delay will be auto-approved after the flag expires.

PR #12904, opened 2019-11-19
Make sure auto-approve runs everything on primary database

diox:
Fixes https://github.com/mozilla/addons-server/issues/12896

Added a couple other commits to help in the future:
- A couple functional tests (were passing already)
- Re-using instance instead of passing the id in `run_action()` since it's not a real task

### DIFF
--- a/src/olympia/scanners/tasks.py
+++ b/src/olympia/scanners/tasks.py
@@ -24,7 +24,6 @@ from olympia.addons.models import AddonReviewerFlags
 from olympia.devhub.tasks import validation_task
 from olympia.files.models import FileUpload
 from olympia.files.utils import SafeZip
-from olympia.versions.models import Version
 
 from .models import ScannerResult, ScannerRule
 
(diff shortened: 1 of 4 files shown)


src/olympia/scanners/tasks.py
```diff
@@ -229,14 +229,14 @@ def _delay_auto_approval_indefinitely(version):
...
+def run_action(version):
     """This function tries to find an action to execute for a given version,
     based on the scanner results and associated rules.
 
     It is not run as a Celery task but as a simple function, in the
     auto_approve CRON."""
-    log.info('Checking rules and actions for version %s.', version_id)
-    version = Version.objects.get(pk=version_id)
+    log.info('Checking rules and actions for version %s.', version.pk)
+    version = Version.objects.get(pk=version.pk)
```
willdurand: Do we really need to re-fetch the object then?

diox: oops, no we don't

diox: f3e63f18c308d6ed22b0cd12bda45d77e68fdebb

PR #12908, opened 2019-11-19
Add definition to scanner rules (Yara)

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12909

---

This patch adds a new `definition` field to the `ScannerRule` model. It is used for Yara only at the moment.


static/js/scanners/scannerrule.js
```diff
@@ -0,0 +1,35 @@
+const selectedScannerIsYara = (select) => {
```
willdurand: @diox do we have unit tests for such js files?

diox: We don't :(

src/olympia/scanners/models.py
```diff
@@ -120,6 +125,32 @@ class Meta:
...
+
+        if len(re.findall(r'rule\s+.+?\s+{', self.definition)) > 1:
+            raise ValidationError(
+                _('Only one Yara rule is allowed in the definition')
+            )
+
+        try:
+            yara.compile(source=self.definition)
+        except Exception:
+            raise ValidationError(_('The definition is not valid'))
```
diox: We should probably try to re-use the exception message here if it's a `yara.SyntaxError`. Showing what the error is is going to be much more user-friendly that just a generic "not valid!" error.

src/olympia/scanners/models.py
```diff
@@ -120,6 +125,32 @@ class Meta:
     def __str__(self):
         return self.name
 
+    def clean(self):
+        if self.scanner == YARA:
+            self.clean_yara()
+
+    def clean_yara(self):
```
diox: For all your definition-related validation errors, you should do `raise ValidationError({'definition': ...})` instead so that the error is assigned to the definition field instead of being put with all the non-field errors at the top.

willdurand: Ha, right. Thanks! I filed https://github.com/mozilla/addons-server/issues/12910

src/olympia/scanners/models.py
willdurand: That's true. I did not find the error messages very user friendly but maybe better than nothing.

diox: Yeah I didn't find them super useful either but it's better than nothing, I guess.

Issue #12910, opened 2019-11-19
Show error messages on the definition field when creating a Yara rule


willdurand:
In #12909, we added a new `definition` field to create Yara rules in the DB. Error messages are rendered globally but we should render them on the `definition` field instead (inline).

AlexandraMoga:
Verified fixed on -dev:

![image](https://user-images.githubusercontent.com/31961530/69435770-dfb1cf80-0d48-11ea-882d-f8fef4747c43.png)

PR #12911, opened 2019-11-19
Better error messages in ScannerRule admin form

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12910

---

After:

Screen Shot 2019-11-19 at 18 44 23

### DIFF
--- a/src/olympia/scanners/models.py
+++ b/src/olympia/scanners/models.py
@@ -131,25 +131,45 @@ class ScannerRule(ModelBase):
 
     def clean_yara(self):
         if not self.definition:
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/models.py
```diff
@@ -131,25 +131,45 @@ def clean(self):
 
     def clean_yara(self):
         if not self.definition:
-            raise ValidationError(_('Yara rules should have a definition'))
+            raise ValidationError(
+                {'definition': _('Yara rules should have a definition')}
```
willdurand: This could be a global error message, maybe.

PR #12925, opened 2019-11-20
Prevent uploading new listed versions while the add-on is "Invisible"

### LINKED ISSUE #4025: Prevent uploading a new version to the listed channel if the add-on is invisible
Invisible add-ons have a special status. Their files are not reviewed, so uploading new ones does not make much sense, so we need to prevent uploading a new version to the listed channel if the add-on is invisible.

diox:
Fixes #4025

### DIFF
--- a/src/olympia/devhub/forms.py
+++ b/src/olympia/devhub/forms.py
@@ -33,7 +33,6 @@ from olympia.addons.utils import verify_mozilla_trademark
 from olympia.amo.fields import HttpHttpsOnlyURLField, ReCaptchaField
 from olympia.amo.forms import AMOModelForm
 from olympia.amo.messages import DoubleSafe
-from olympia.amo.templatetags.jinja_helpers import mark_safe_lazy
 from olympia.amo.urlresolvers import reverse
 from olympia.amo.utils import (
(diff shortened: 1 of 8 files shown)


src/olympia/devhub/forms.py
```diff
@@ -1217,12 +1217,24 @@ class DistributionChoiceForm(forms.Form):
...
-            ('unlisted', mark_safe_lazy(UNLISTED_LABEL))),
+        choices=[],
         initial='listed',
         widget=forms.RadioSelect(attrs={'class': 'channel'}))
 
+    def __init__(self, *args, **kwargs):
+        self.addon = kwargs.pop('addon', None)
+        super().__init__(*args, **kwargs)
+        choices = [
+            ('listed', mark_safe_lazy(self.LISTED_LABEL)),
```
EnTeQuAk: you won't need the `_lazy` call here because you moved this to ``__init__`` and thus translation happens during runtime.

diox: Indeed, thanks. 5beed9ddd20be7badd411385ce052fa8b9c7e197

PR #12928, opened 2019-11-20
Disallow comments submitted for a line number and a null file.

### LINKED ISSUE #12635: Disallow comments submitted for a line number and a null file

Post a comment with `filename: null` and `line: 123`. Example:

```
curl 'http://localhost:3000/api/v5/reviewers/addon/616959/versions/1688394/draft_comments/?lang=en-US'  \
  -H 'Authorization: Bearer eyJh...' \
  -H 'Content-Type: application/json' \
  -H 'Cookie: sessionid=...' \
  --data '{"comment": "A comment", "filename": null, "lineno": 123}'
```

### What happened?

The comment was created.

The request should be rejected with a 400 because commenting on a line number without specifying the file doesn't make sense.

Because the frontend pre-loads all comments for the version, it's causing this error: https://github.com/mozilla/addons-code-manager/issues/1126

EnTeQuAk:
Fixes #12635

### DIFF
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -392,4 +392,9 @@ class DraftCommentSerializer(serializers.ModelSerializer):
                 {'comment': ugettext(
                     'You can\'t submit a comment if `canned_response` is '
                     'defined.')})
+        if data.get('lineno') and not data.get('filename'):
+            raise serializers.ValidationError(
+                {'comment': ugettext(
+                    'You can\'t submit a line number without associating '
+                    'it to a filename.')})
         return data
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/serializers.py
```diff
@@ -386,10 +386,32 @@ def __init__(self, *args, **kwargs):
...
 
+    def get_or_default(self, key, data, default=''):
+        """Return the value of ``key`` in ``data``
+
+        If that key is not present then return the value of ``key`` from
+        ``self.instance`, otherwise return the ``default``.
+
+        This method is a helper to simplify validation for partial updates.
+        """
+        return data.get(key, getattr(self.instance, key, default))
```
eviljeff: Is it common to be requesting a property on an instance that doesn't exist?  (That seems more likely to be due to an error in the code?)
When I read the description I assumed it would be `data.get(key, getattr(self.instance, key) or default)`

EnTeQuAk: Indeed, I agree. I think it'd even be better to raise an error if an attribute doesn't exist, I'll change the code.

I guess it's just been bad habit.

PR #12929, opened 2019-11-20
Improve documentation regarding filename and it's absolute val…

### LINKED ISSUE #12608: Document `filename` limitations in the comment APIs
STR:
1. I used Postman to send a reviewer comment on a specific line for a specific file using the [documentation provided](https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#canned-responses). 
2. The Request JSON Object for a file which is inside a folder or even a sub-folder needs to contain the entire path for the file, else the POST or PATCH operations will be created but not available on the client side (code manager in out case) for that specified file.
POST 
https://addons-dev.allizom.org/api/v4/reviewers/addon/492486/versions/1529211/draft_comments/

Request JSON Object:
`{   "comment":"text now",
	"filename": "_locales/ar/messages.json",
	"lineno":5
}
 
Actual result:
The documentation does not provide this information.

Expected result:
Add for the POST and PATCH operations
filename (string) – The filename  this comment is related to (optional). - (with it's path in case the file is inside a folder)

EnTeQuAk:
Fixes #12608

Does that make more sense now? Looking at it, it would probably make sense to rename `filename` in `filepath` but I guess that's just too much work for that small change.

### DIFF
--- a/docs/topics/api/reviewers.rst
+++ b/docs/topics/api/reviewers.rst
@@ -335,7 +335,7 @@ These endpoints allow you to draft comments that can be submitted through the re
 
     :>json int id: The id of the draft comment object.
     :>json string comment: The comment that is being drafted as part of a review. Specific to a line in a file.
-    :>json string|null filename: The absolute file path a specific comment is related to. Can be ``null`` in case a comment doesn't belong to a specific file but the whole version.
+    :>json string|null filename: The full file path a specific comment is related to. Can be ``null`` in case a comment doesn't belong to a specific file but the whole version.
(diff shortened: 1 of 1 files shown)


docs/topics/api/reviewers.rst
```diff
@@ -349,7 +349,7 @@ These endpoints allow you to draft comments that can be submitted through the re
     Create a draft comment for a specific version.
 
     :<json string comment: The comment that is being drafted as part of a review.
-    :<json string filename: The filename this comment is related to (optional).
+    :<json string filename: The absolute file path this comment is related to (optional). This must represent the absolute path, including sub-folders.
```
kumar303: I think it's misleading to call this an absolute path. It's really a full path, relative to the root of the add-on. For example, `background.js` or `lib/scripts/background.js` -- I wouldn't consider either of those an absolute path.

EnTeQuAk: I tried to clarify this in eeb8317, maybe that's a bit better?

Issue #12874, opened 2019-11-14
AMO doesn't find files for user-deleted add-ons


wagnerand:
### Describe the problem and steps to reproduce it:

AMO cannot find files for user-deleted add-ons. The files linked from the reviewer tools are just 404.

Assumed STR (not tested):
1. Submit an add-on, possibly need to submit a couple of versions.
2. Delete the add-on.
3. Observe the file status, it should be `STATUS_APPROVED`.
4. Try to download the file from the review tools page.

### What happened?

404 - no file.

### What did you expect to happen?

200 - file download.

### Anything else we should know?

It seems files are not having their status updated when a user deletes their add-on.

muffinresearch:
@wagnerand do you have a proposed priority on this. and is this new?

wagnerand:
I am not sure if this is new. File visibility has come up a few times and we have patched some issues, but maybe we missed a case or two.

I am tagging with P2 to have it above the fold. It's not extremely urgent, but when we run into it, it's a rather significant annoyance and could have impact on our ability/speed on reacting to suspicious or malicious add-ons.

diox:
> Observe the file status, it should be `STATUS_APPROVED`.

Should it ? Shouldn't we set the files status to `STATUS_DISABLED` ?

wagnerand:
> > Observe the file status, it should be `STATUS_APPROVED`.
> 
> Should it ? Shouldn't we set the files status to `STATUS_DISABLED` ?

Sorry, poor choice of words. "should" was describing what I experienced, not what the expected correct thing should be.

diox:
QA: with this fixed, the file status will now be `STATUS_DISABLED` when deleting. The version should be deleted as well. As always with disabling files, we will need to wait for up to one hour for the files to be moved to the right storage (`hide_disabled_files` needs to kick in).

Also, this has implications for permissions needed to review deleted add-ons. We should require the normal `Addons:Review` permission to access the listed review page for an add-on that used to have listed versions before being deleted, and `Addons:ReviewUnlisted` otherwise.

It should still be impossible for a developer to access deleted add-ons with the old devhub URL (by id or by slug).

wagnerand:
> It should still be impossible for a developer to access deleted add-ons with the old devhub URL (by id or by slug).

Does that mean slugs are locked forever?

diox:
> Does that mean slugs are locked forever?

No, deleting clears the slug. Note that I said *impossible*.

[...]

PR #12940, opened 2019-11-22
Set Versions as deleted and disable files when deleting an Addon

diox:
Fixes https://github.com/mozilla/addons-server/issues/12874

### DIFF
--- a/src/olympia/addons/decorators.py
+++ b/src/olympia/addons/decorators.py
@@ -43,8 +43,9 @@ def addon_view(
 
         # If the addon has no listed versions it needs either an author
         # (owner/viewer/dev/support) or an unlisted addon reviewer.
-        if not (addon.has_listed_versions(
-            include_deleted=include_deleted_when_checking_versions) or
+        has_listed_versions = addon.has_listed_versions(
+            include_deleted=include_deleted_when_checking_versions)
+        if not (has_listed_versions or
                 owner_or_unlisted_reviewer(request, addon)):
             raise http.Http404
         return f(request, addon, *args, **kw)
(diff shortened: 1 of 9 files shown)


src/olympia/addons/decorators.py
```diff
@@ -40,9 +41,10 @@ def wrapper(request, addon_id=None, *args, **kw):
...
                 return http.HttpResponsePermanentRedirect(url)
 
-        # If the addon is unlisted it needs either an owner/viewer/dev/support,
-        # or an unlisted addon reviewer.
-        if not (addon.has_listed_versions() or
+        # If the addon has no listed versions it needs either an author
+        # (owner/viewer/dev/support) or an unlisted addon reviewer.
+        if not (addon.has_listed_versions(
+            include_deleted=include_deleted_when_checking_versions) or
                 owner_or_unlisted_reviewer(request, addon)):
```
EnTeQuAk: I wonder if we can extract those function calls into separate variables above to make it easier to read:

```python
has_listed_version = addon.has_listed_versions(...)
is_owner_or_unlisted_reviewer = owner_or_unlisted_reviewer(...)

if not has_listed_version or is_owner_or_unlisted_reviewer:
    # ....
```

diox: I've extracted the first one into a separate variable. The second one would make a database query to load the user groups... It's likely the view using that decorator would need to do that at some point, and it's cached into an instance property, but still I'm not sure we want to do that just yet...

EnTeQuAk: That works for me, it's already much more readable. Thanks!

PR #12941, opened 2019-11-22
Add revert action for triaged scanner results

### LINKED ISSUE #12939: Allow reverting human scanner result verdict

When humans happen to click on the wrong button in the scanner admin result list, they need a way to revert that decision.

### What happened?

The actions column shows `-` for triaged results.

We could show a "Retriage" or similar button when a result is triaged.

wagnerand:
Fix #12939.

### DIFF
--- a/src/olympia/scanners/admin.py
+++ b/src/olympia/scanners/admin.py
@@ -216,6 +216,18 @@ class ScannerResultAdmin(admin.ModelAdmin):
             )
         )
 
+    def handle_revert(self, request, pk, *args, **kwargs):
+        result = self.get_object(request, pk)
+        result.update(state=UNKNOWN)
+
+        messages.add_message(
+            request,
+            messages.INFO,
+            'Scanner result {} report has been reverted.'.format(pk),
+        )
+
+        return redirect('admin:scanners_scannerresult_changelist')
+
     def get_urls(self):
         urls = super().get_urls()
         custom_urls = [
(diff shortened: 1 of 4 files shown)


src/olympia/scanners/models.py
```diff
@@ -99,6 +99,11 @@ def can_report_feedback(self):
             self.has_matches and self.state == UNKNOWN and self.scanner != WAT
         )
 
+    def can_revert_feedback(self):
```
willdurand: We should probably have one or two test cases for this method.

src/olympia/scanners/admin.py
```diff
@@ -229,26 +241,38 @@ def get_urls(self):
...
-            reverse(
-                'admin:scanners_scannerresult_handlefalsepositive',
-                args=[obj.pk],
-            ),
-            reverse(
-                'admin:scanners_scannerresult_handletruepositive',
-                args=[obj.pk],
-            ),
-        )
+        if obj.can_report_feedback():
```
willdurand: This makes me think that we should likely use a `<form>` and a `POST` method instead of using `GET` to update things. Not something to change in this patch, though.

src/olympia/scanners/models.py
wagnerand: What exactly should be tested?

src/olympia/scanners/admin.py
wagnerand: Yes, I agree with both. I wanted to have a quick solution to make it into the next tag. Unfortunately, it'll have to wait a little longer now, because we don't push this week.

src/olympia/scanners/models.py
willdurand: - a scanner result that has been previously reported as false or true positive can be reverted
- a default scanner result should not be "revert-able"

wagnerand: 4b1f07febe

src/olympia/scanners/admin.py
willdurand: I'll file an issue :)

PR #12956, opened 2019-11-25
rm backend support for featured collections/addons

### LINKED ISSUE #12907: Remove "Featured" addons support
The Recommend add-ons programme has supplanted Featured add-ons, and has been live for a number of months (i.e. we're not going back).  It's time to remove the supporting code, tools, and API properties (in a v3 compatible way).  This can be done in stages.

eviljeff:
fixes #12907


src/olympia/addons/migrations/0004_auto_20191126_1712.py
```diff
@@ -0,0 +1,28 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name='addon',
+            name='type',
+            field=models.PositiveIntegerField(choices=[(1, 'Extension'), (2, 'Deprecated Complete Theme'), (3, 'Dictionary'), (4, 'Search Engine'), (5, 'Language Pack (Application)'), (6, 'Language Pack (Add-on)'), (7, 'Plugin'), (9, 'Deprecated LWT'), (10, 'Theme (Static)')], db_column='addontype_id', default=1),
+        ),
+        migrations.AlterField(
+            model_name='addoncategory',
+            name='feature',
+            field=models.NullBooleanField(default=None),
```
eviljeff: this is just to give the db column a default value

src/olympia/addons/migrations/0004_auto_20191126_1712.py
```diff
@@ -0,0 +1,28 @@
...
+
+    dependencies = [
+        ('addons', '0003_addonreviewerflags_auto_approval_disabled_until'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='addon',
+            name='type',
+            field=models.PositiveIntegerField(choices=[(1, 'Extension'), (2, 'Deprecated Complete Theme'), (3, 'Dictionary'), (4, 'Search Engine'), (5, 'Language Pack (Application)'), (6, 'Language Pack (Add-on)'), (7, 'Plugin'), (9, 'Deprecated LWT'), (10, 'Theme (Static)')], db_column='addontype_id', default=1),
```
eviljeff: Pretty sure I didn't make this change in the patch, but it seems pretty harmless.

src/olympia/addons/utils.py
```diff
@@ -1,110 +1,26 @@
...
-        qs = qs.filter(is_featured)
-    other_ids = list(qs.distinct().values_list('id', flat=True))
-    random.shuffle(ids)
-    random.shuffle(other_ids)
-    ids += other_ids
-    return list(map(int, ids))
-
-
-@memoize('addons:creatured', timeout=60 * 10)
-def get_creatured_ids(category, lang=None):
```
EnTeQuAk: OMG, `creatured` is gone! :skull: :fire:

src/olympia/addons/migrations/0004_auto_20191126_1712.py
EnTeQuAk: I think I agree, same with `category.type` below which is curious. The values are identical to `ADDON_TYPE` so that should be fine I think. Do you have maybe some local changes that you didn't commit?

eviljeff: I don't.  Don't you get the same thing if you do `./manage.py makemigrations addons` ?

eviljeff: I get the same migrations from a checkout from mozilla/addons-server:master.  Possibly from when I changed the status labels.

PR #12960, opened 2019-11-25
index block by latest version too so shows up on review page

### LINKED ISSUE #12456: Show blocklist actions in review history

When a block is added or deleted, an event should be added to the review history for the affected versions, indicating the date and who made the change. It shouldn't include the bug link or block reason, because it could contain sensitive information and this will be accessible to all reviewers.

eviljeff:
fixes #12456

### DIFF
--- a/src/olympia/blocklist/utils.py
+++ b/src/olympia/blocklist/utils.py
@@ -2,15 +2,14 @@ from olympia import amo
 from olympia.activity import log_create
 
 
-def add_to_latest_version_in_channel(obj, al, channel):
+def add_version_log_for_blocked_versions(obj, al):
     from olympia.activity.models import VersionLog
 
-    version_ids = [
-        id_ for id_, chan in obj.addon_versions.values()
-        if chan == channel]
-    if not version_ids:
-        return
-    VersionLog.objects.create(activity_log=al, version_id=version_ids[-1])
+    VersionLog.objects.bulk_create([
+        VersionLog(activity_log=al, version_id=id_chan[0])
+        for version, id_chan in obj.addon_versions.items()
+        if obj.is_version_blocked(version)
+    ])
 
 
 def block_activity_log_save(obj, change):
@@ -29,8 +28,7 @@ def block_activity_log_save(obj, change):
     al = log_create(
         action, obj.addon, obj.guid, obj, details=details, user=obj.updated_by)
 
-    add_to_latest_version_in_channel(obj, al, amo.RELEASE_CHANNEL_LISTED)
-    add_to_latest_version_in_channel(obj, al, amo.RELEASE_CHANNEL_UNLISTED)
+    add_version_log_for_blocked_versions(obj, al)
 
 
 def block_activity_log_delete(obj, user):
@@ -38,5 +36,4 @@ def block_activity_log_delete(obj, user):
         amo.LOG.BLOCKLIST_BLOCK_DELETED, obj.addon, obj.guid, obj,
         details={'guid': obj.guid}, user=user)
 
-    add_to_latest_version_in_channel(obj, al, amo.RELEASE_CHANNEL_LISTED)
-    add_to_latest_version_in_channel(obj, al, amo.RELEASE_CHANNEL_UNLISTED)
+    add_version_log_for_blocked_versions(obj, al)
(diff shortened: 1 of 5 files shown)


src/olympia/blocklist/utils.py
```diff
@@ -2,6 +2,17 @@
 from olympia.activity import log_create
 
 
+def add_to_latest_version_in_channel(obj, al, channel):
+    from olympia.activity.models import VersionLog
+
+    version_ids = [
+        id_ for id_, chan in obj.addon_versions.values()
```
diox: I don't think it makes sense to use `addon_versions` here: it can be super costly (both in python and db), and unless I'm mistaken it doesn't appear to be used by anything else when saving a block, so you don't gain anything by preloading it. In this case you just need the latest version in each channel, it's probably best to use `find_latest_version()` instead, even if that means making 2 queries.

src/olympia/blocklist/utils.py
```diff
@@ -13,12 +24,19 @@ def block_activity_log_save(obj, change):
...
         'reason': obj.reason,
         'include_in_legacy': obj.include_in_legacy,
+        'comments': f'Versions {obj.min_version} - {obj.max_version} blocked.',
     }
-    log_create(
+    al = log_create(
         action, obj.addon, obj.guid, obj, details=details, user=obj.updated_by)
 
+    add_to_latest_version_in_channel(obj, al, amo.RELEASE_CHANNEL_LISTED)
+    add_to_latest_version_in_channel(obj, al, amo.RELEASE_CHANNEL_UNLISTED)
```
diox: mmm actually this is kind of weird. You're attaching logs to versions that might not have anything to do with the versions you're blocking, they might not even be in the same channel ? That seems wrong.

src/olympia/blocklist/utils.py
eviljeff: For a single block `addon_versions` will have been preloaded for `_get_version_choices`, called by `get_form`.
For multiple blocks, looks like you're right - though it should be more efficient to add in a line to specifically call `Block.preload_addon_versions` and do one (big) query for all Blocks being saved than 2 small queries per Block. (?)

diox: Ah, I hadn't thought about the fact that the choices would have been generated for the single block, makes sense.

For multiple blocks... preloading *all* versions from *all* addons in a multiple block just to access 2 of them per add-on seems a bit much... and likely to cause issues depending on how many add-ons are in the block, since the block is processed by a single task.

src/olympia/blocklist/utils.py
eviljeff: Blocks aren't channel specific so they could affect either channel.  The issue did originally call for a log entry for each version and I asked if that was necessary, given the version is annotated with **Blocked** anyway, and Jorge seemed to think it wasn't.  

I could do it anyway though - it'd be iterating through every version and adding the `VersionLog` link if it matched.  
Some downsides here: 
- if old versions are blocked the message might not be visible in the log on the first page (we probably don't care)
- The logic around logging for editing blocks would be less clear.  E.g. if the range is shortened and a version now _isn't_ blocked what do we do about the existing message?  If the range is expanded do we have to add it to all the new affected versions?  If a block is edited generally do we add the message again to the existing versions?

diox: I think I'd rather have it set to all affected versions and only those, and re-add it when the block is edited - it's a log, let's treat it as append-only (plus it's an edge case anyway).

src/olympia/blocklist/utils.py
eviljeff: Well, we're going to need all the versions to be able to add a `VersionLog` entry for all of them anyway.  

I was thinking the submit task might need changing to batch the guids tbh.  Running `process_input_guids` on a slice of 100 guids each time.  (I'd not planned to add it in this PR)

diox: Yeah. It's fine to do in a follow-up, I guess. I'm slightly worried about the number of versions we can end up loading and then inserting a `VersionLog` for, but I guess we already have similar problems when we display a single block form anyway (albeit without the insertion part)

src/olympia/blocklist/utils.py
eviljeff: https://github.com/mozilla/addons-server/pull/12960/commits/440add512e2381bf9a86430d876ed3eb520dcf95

PR #12983, opened 2019-11-27
change pngs for jpegs + add scott by the window

### LINKED ISSUE #12950: Disco admin needs an additional hero image
We recently uploaded a number of new images to be used in the AMO homepage hero area. However one image we identified to include didn't make it into the admin. Can we please add the attached image? 

hero_worldwideradio

### LINKED ISSUE #12973: Optimize primary Hero images
The images that we were given are PNGs of photographs, so they are much heavier than they should be. Let's convert them to JPEG and optimize their size.

eviljeff:
fixes #12973 + fixes #12950


src/olympia/hero/migrations/0006_auto_20191127_1246.py
```diff
@@ -0,0 +1,30 @@
...
+
+import olympia.hero.models
+
+
+def featured_images_png_to_jpg(apps, schema_editor):
+    PrimaryHero = apps.get_model('hero', 'PrimaryHero')
+    for shelf in PrimaryHero.objects.filter(image__endswith='.png'):
+        new_name = splitext(shelf.image)[0] + '.jpg'
+        shelf.update(image=new_name)
+
```
eviljeff: @bqbn I'm not sure if this is going to hit problems during the prod deployment window.  The migration will be updating the database records to reference the new image files in /static/images/ - will they be there when the migration is run?

bqbn: @eviljeff by "there", do you mean where?

What I can confirm is that before the db migration happens, what's inside the `site-static/` directory in the Django project would have been uploaded to the S3 bucket (e.g. [[1]](https://github.com/mozilla-services/cloudops-deployment/blob/master/projects/addons-server/scripts/push_static_files.sh#L20-L24) runs before the db migration happens).

Note that the website is still running the old app code when the db migration happens.

Let me know if that answers your question or not.

eviljeff: > by "there", do you mean where?

Um, no.  _There_ referring to  /static/images.

> What I can confirm is that before the db migration happens, what's inside the site-static/ directory in the Django project would have been uploaded to the S3 bucket (e.g. [1] runs before the db migration happens).

That answers my question, thanks.  The new images files will exist when the database migration runs to update to point to the new files.  Although... does that mean any images that have been deleted in that tag will be removed too I guess?  We should be remove them in a later tag perhaps.

bqbn: > does that mean any images that have been deleted in that tag will be removed too I guess?

Which files? Can you give some example file names?

Any files that do not exist in the local `site-static/` directory, but exist in the remote `s3://$remoteRepo/static/` directory will be deleted.

eviljeff: > Which files? Can you give some example file names?

All the files shown as removed in this pr

eviljeff: https://github.com/mozilla/addons-server/pull/12983/commits/c3e89d48618b3e6515226f53057b4740acdd1c3f adds the deleted images back so they can still be accessed during the deployment.  We'll have to delete them in the next tag instead.

bqbn: Ah, OK. This makes sense.

Issue #13006, opened 2019-12-02
Fix permissions required to download source code


diox:
I've tweaked the permissions required to download source code in https://github.com/mozilla/addons-server/pull/12987 but I think it's still wrong / not what I intended.

The rules should be:
- If the add-on or version is deleted: admins only
- If the add-on or file is disabled: admins only
- Otherwise: admins and developers

The channel shouldn't matter. "admins" in this case are people with `Reviews:Admin`.

wagnerand:
In any case, only admins should be allowed to access the source code package.

diox:
Wait, not developers ? They have a link in the devhub version edit page.

wagnerand:
Oh, you are right, sorry.

AlexandraMoga:
Verified fixed on -dev:

Reviewers with `Reviews:Admin` permissions can download Source files attached to an add-on regardless of the add-on status: approved, disabled or deleted.
Regular reviewers will not see the Download sources link in the  version history.

Developers can also download the attached source code from the manage versions page if the add-on status is Approved or Disabled.

PR #13007, opened 2019-12-02
Fix permissions required to download source code

diox:
Admins should be allowed by default, not just when the addon/version
is deleted.

Fixes https://github.com/mozilla/addons-server/issues/13006

### DIFF
--- a/src/olympia/versions/tests/test_views.py
+++ b/src/olympia/versions/tests/test_views.py
@@ -522,6 +522,7 @@ class TestDownloadSource(TestCase):
     def test_download_for_admin(self):
         """File downloading is allowed for admins."""
         self.grant_permission(self.user, 'Reviews:Admin')
+        self.addon.authors.clear()
         self.client.login(email=self.user.email)
         assert self.client.get(self.url).status_code == 200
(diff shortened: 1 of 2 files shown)


src/olympia/versions/tests/test_views.py
```diff
@@ -522,6 +522,7 @@ def test_download_for_unlisted_addon_reviewer(self):
     def test_download_for_admin(self):
         """File downloading is allowed for admins."""
         self.grant_permission(self.user, 'Reviews:Admin')
+        self.addon.authors.clear()
```
EnTeQuAk: That doesn't look like it's testing the actual changes. I think there are some tests missing?

diox: It is actually testing the changes : that test covers every case (disabled, deleted, etc) for admins. The problem was, before I added this single line, the user in the test was also an author of the add-on, and thus the test wasn't testing what it should have been testing...

EnTeQuAk: Oh, I missed that. Thanks for the explanation!

Issue #12992, opened 2019-11-29
Default "scanner results" list view should not show the entries without a version


willdurand:
We should not show the scanner results without versions attached by default in the admin list view because there is nothing an admin can do.

We can have a filter to "show everything" if we need.

AlexandraMoga:
Verified fixed on -dev.
The default scanner results list view will not include entries with no version attached.
These type of records can still be viewed if you select "All" in the "By presence of a version" filter.

PR #13009, opened 2019-12-02
Add a new filter to hide scanner results without a version by…

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12992


willdurand:
I just rebased this patch.

Issue #12948, opened 2019-11-22
Open scanner result false positive github issue in new tab.


wagnerand:
### Describe the problem and steps to reproduce it:

The github issue draft for reporting a false positive scanner result should open in a new tab.

### Anything else we should know?
(Please include a link to the page, screenshots and any relevant files.)

willdurand:
Opening the link in a new tab seems doable but the list view will not be refreshed, that's why we do not open the github issue page in a new tab.

Maybe the link to file a github issue should appear in a flash message, WDYT @wagnerand?

wagnerand:
Eventually it would be cool if we did something a little AJAXy. Clicking the button would send a XHR to the server, wait for the result and then update the DOM. At the same time, we could open a new tab to report the issue. I am not sure if that's possible in django admin?

willdurand:
mmm I guess we can do many things but what would be the goal for an ajax admin page? Say we can solve this issue without it? (I *think* we can trigger a page refresh before opening a link to a new tab).

wagnerand:
That works too!

AlexandraMoga:
Verified fixed on -dev.
If you click on the "Report false positive' button for a scanner result, the github repository will open in a new tab. The scanner results list will also be refreshed.
I don't have access to the github repo to see how the  issue created looks like but by looking at what we send in the URL, the necessary information seems to be there.

PR #13012, opened 2019-12-02
Open scanner result false positive github issue in new tab

willdurand:
~~:warning: depends on #13011~~
Fixes https://github.com/mozilla/addons-server/issues/12948

---

It feels a bit like the 2000s but it is simple and it works well.


willdurand:
I just rebased this patch. Let me know what you think :)

Issue #12918, opened 2019-11-20
Add filter for each rule in admin scanner result page


wagnerand:
The admin scanner result page already supports filtering by scanner. In order to focus on specific issues, let's also add a filter for each rule of every scanner (where applicable).

diox:
We should probably also add a link to all results from the rule change/view page (we do something similar in other admin tools, displaying a count of the number of related objects at the same time)

AlexandraMoga:
Scanner results can now be filtered by matched rules. The filter is updated when a a rule is added or deleted.

@diox have you also implemented this:

>We should probably also add a link to all results from the rule change/view page (we do something similar in other admin tools, displaying a count of the number of related objects at the same time)

If I understand correctly, this should have added a link to each scanner rule `/change/` view.

diox:
Ah, it got mixed up in my mind: I wanted to add the link to the results to each rule change page, but I added the link to the rules to each result view page instead :) Both are useful, going to do this.

AlexandraMoga:
Verified fixed on -dev:

- link to matched scanner results in rule change page
![image](https://user-images.githubusercontent.com/31961530/70146455-d579c480-16aa-11ea-8554-7f0d7356bd68.png)

- link to rule change page in scanner result view 
![image](https://user-images.githubusercontent.com/31961530/70146460-d9a5e200-16aa-11ea-9ecf-3ee7e0f33ce0.png)

PR #13020, opened 2019-12-03
Include link to all results for a given rule in admin

diox:
Fixes https://github.com/mozilla/addons-server/issues/12918

### DIFF
--- a/src/olympia/scanners/admin.py
+++ b/src/olympia/scanners/admin.py
@@ -324,11 +324,11 @@ class ScannerRuleAdmin(admin.ModelAdmin):
     list_display = ('name', 'scanner', 'action', 'is_active')
     list_filter = ('scanner', 'action', 'is_active')
     fields = (
-        'created',
-        'modified',
         'name',
         'scanner',
         'action',
+        'created',
+        'modified',
         'definition',
         'matched_results_link',
         'is_active',
--- a/src/olympia/scanners/admin.py
+++ b/src/olympia/scanners/admin.py
@@ -339,7 +339,11 @@ class ScannerRuleAdmin(admin.ModelAdmin):
         count = ScannerResult.objects.filter(matched_rules=obj).count()
         url = reverse('admin:{}_{}_changelist'.format(
             ScannerResult._meta.app_label, ScannerResult._meta.model_name))
-        url = (f'{url}?matched_rules__id__exact={obj.pk}&has_version=all'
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/admin.py
```diff
@@ -323,3 +323,23 @@ class ScannerRuleAdmin(admin.ModelAdmin):
...
+        'name',
+        'scanner',
+        'action',
+        'definition',
+        'matched_results_link',
+        'is_active',
+    )
+    readonly_fields = ('created', 'modified', 'matched_results_link',)
+
+    def matched_results_link(self, obj):
```
diox: Didn't think it was worth a code comment, but FYI I didn't use `zadmin.admin.related_content_link` because I wanted more flexibility around the url and possibly the text/queryset later as well.

src/olympia/scanners/admin.py
```diff
@@ -323,3 +323,23 @@ class ScannerRuleAdmin(admin.ModelAdmin):
...
+    fields = (
+        'created',
+        'modified',
+        'name',
+        'scanner',
+        'action',
+        'definition',
+        'matched_results_link',
+        'is_active',
+    )
```
willdurand: Can we move the date fields down in the list?

src/olympia/scanners/admin.py
```diff
@@ -323,3 +323,23 @@ class ScannerRuleAdmin(admin.ModelAdmin):
...
+    readonly_fields = ('created', 'modified', 'matched_results_link',)
+
+    def matched_results_link(self, obj):
+        count = ScannerResult.objects.filter(matched_rules=obj).count()
+        url = reverse('admin:{}_{}_changelist'.format(
+            ScannerResult._meta.app_label, ScannerResult._meta.model_name))
+        url = (f'{url}?matched_rules__id__exact={obj.pk}&has_version=all'
```
willdurand: Something like this would be more future-proof:

```suggestion
        url = (f'{url}?matched_rules__id__exact={obj.pk}&{WithVersionFilter.parameter_name}=all'
```

src/olympia/scanners/admin.py
```diff
@@ -323,3 +323,23 @@ class ScannerRuleAdmin(admin.ModelAdmin):
...
+               f'&state=all&scanner={obj.scanner}')
```
willdurand: I think we should also rely on the `parameter_name` of the `StateFilter`.

PR #13031, opened 2019-12-04
Don't include pending versions in unlisted review actions affecting multiple versions

### LINKED ISSUE #13008: Unlisted versions can't be manually approved
Prerequisites: Log in with a user with Review:Unlisted permissions

1. Submit a new unlisted add-on that matches a rule with a delay auto-approval action - this will prevent the add-on from being auto-approved 
2. Search for your add-on in the Unlisted Review queue
3. Open the review page of your add-on and approve the version by using the 'Confirm multiple versions' option 
4. Check the status of your add-on after confirming approval

### What happened?
The add-on status is Awaiting Review (Confirmed) - see [example](https://reviewers.addons-dev.allizom.org/en-US/reviewers/review-unlisted/135c15504d6649caaf15)
Reviewers should have an option to manually Approve unlisted add-ons
- by contrast, a listed version that has a delay auto-approval flag will include the manually 'Approve' option in reviewer actions, which will set the add-on status as approved - see [example](https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/human-review-delay3)
- for unlisted versions, reviewers can force auto-approval by clearing the 'Delay auto-approval` flag but don't have the option to manually approve the version
- this scenario can be reproduce both on -dev and stage

diox:
Fixes #13008

### DIFF
--- a/src/olympia/reviewers/forms.py
+++ b/src/olympia/reviewers/forms.py
@@ -346,11 +346,15 @@ class ReviewForm(forms.Form):
                 channel = self.helper.version.channel
             else:
                 channel = amo.RELEASE_CHANNEL_LISTED
+            # For unlisted, we only care about approved versions, reviewers
+            # aren't actively monitoring that queue.
+            if channel == amo.RELEASE_CHANNEL_UNLISTED:
+                statuses = (amo.STATUS_APPROVED, )
+            else:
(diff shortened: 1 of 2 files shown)


diox:
The original issue is only present because QA was trying to use the action on a non-approved version - that shouldn't be allowed, since the action is only meant to confirm approval (and unset the needs human review flag in the process).

Issue #12993, opened 2019-11-29
Improve rule creation page


wagnerand:
For yara rules, the rule name should in the dedicated name field should match the name in the definition, so technically we wouldn't even need to write a name, we could just extract it from the definition. On the other hand, we need it for other scanners.

Hence, let's move the scanner selection to the top of the page, hide the name field by default and only unhide it if the selected scanner is not yara.

ioanarusiczki:
@willdurand 
When a rule for the `Yara` scanner is created it has only the `description` field available where the name of the rule can be added. 
Customs and Wat have the `Name` field.
Works as [described](https://github.com/mozilla/addons-server/pull/13033) by the pull request. 
Verified on dev with FF70(Win10).

PR #13033, opened 2019-12-04
Simplify how yara rules are created

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/12993

---


willdurand:
@diox in the screencast, we can see that the value for "matched results" is not correct when creating a new rule.

diox:
Good catch, https://github.com/mozilla/addons-server/pull/13046 will fix the matched rules being present at creation time.

PR #13042, opened 2019-12-05
add utm params to contribution url in addon api detail response

### LINKED ISSUE #13010: Add UTM params to Contribute URLs
This has been moved from https://github.com/mozilla/addons-frontend/issues/9005, quoting the original issue:

> We should set UTM params in the Contribute button so our referrals can be tracked. In particular, some devs point to the Donate to Mozilla page, and with that data we can figure out how much donations we're driving.

> The params should be `utm_content=product-page-contribute&utm_medium=referral&utm_source=addons.mozilla.org`

Can we change the code the generates the contribute URL to add these UTM params before "outgoingifying" it?

cc @jvillalobos as the OP

eviljeff:
fixes #13010

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -2,6 +2,7 @@ import re
 from urllib.parse import urlsplit, urlunsplit
 
 from django.conf import settings
+from django.http.request import QueryDict
 
 from rest_framework import exceptions, serializers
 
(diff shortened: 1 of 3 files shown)


src/olympia/addons/serializers.py
```diff
@@ -415,6 +416,16 @@ def get_url(self, obj):
         # annoying in subclasses which don't want to load that version.
         return absolutify(obj.get_detail_url())
 
+    def get_contributions_url(self, obj):
+        if not obj.contributions:
+            # don't add anything when it's not set.
+            return obj.contributions
+        parts = urlsplit(obj.contributions)
+        query = '&'.join(
+            s for s in (parts.query, amo.CONTRIBUTE_UTM_PARAMS) if s)
```
willdurand: I would make `CONTRIBUTE_UTM_PARAMS` a `dict` and I would parse `parts.query` with `parse_qs` (probably), then merge both dicts and finally use `urlencode`. That would be arguably more readable that this line and that would allow us to override the `utm_*` query params if the `contributions` URL has query params including one of the UTM params we want to set.

eviljeff: QueryDicts explicitly _do_ allow duplicate keys though https://docs.djangoproject.com/en/2.2/ref/request-response/#querydict-objects.  Do we want to override a user's UTM params with our own anyway? - that seems like a bad thing given that we're adding these to all services not just donate.m.o (and wasn't specified in the issue)

willdurand: ha, right 🤦‍♂ Then, I guess we can keep duplicate keys then and use `QueryDict.update()`, that would still be slightly more readable in my opinion 😛

eviljeff: https://github.com/mozilla/addons-server/pull/13042/commits/6da0d9db2fb30601a126e1188fba8957c85b5aeb
Ironically, I started the original patch using `parse_qs` and then changed to a flat string because I thought it was overwritten and made it less readable.

PR #13055, opened 2019-12-06
Change docker-compose nginx path to serve installable addon.

jrbenny35:
The install-able addon that gets created for the integration tests was not being served correctly from nginx. This should fix that.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -41,7 +41,7 @@ services:
       - ./static:/srv/static
       - ./site-static:/srv/site-static
       - ./storage/shared_storage/uploads:/srv/user-media
-      - ./storage/files:/srv/user-media/addons
+      - ./storage/files:/user-media/addons
     ports:
       - "80:80"
 


docker-compose.yml
```diff
@@ -41,7 +41,7 @@ services:
       - ./static:/srv/static
       - ./site-static:/srv/site-static
       - ./storage/shared_storage/uploads:/srv/user-media
-      - ./storage/files:/srv/user-media/addons
+      - ./storage/files:/user-media/addons
```
EnTeQuAk: I'm a bit confused, because nginx will look at `/srv/user-media/addons` - https://github.com/mozilla/addons-nginx/blob/master/addons.conf#L30-L32

Do you have STR to reproduce the issue you're talking about?

jrbenny35: Sure. If you setup a local instance for development using the ```setup ui-tests``` make command, then navigate to the frontend, try to install the addon named ```Ui-Addon-Install```. It should load an nginx does not exist page.

eviljeff: if this is just for the UI tests maybe it should be in docker-compose.selenium.yml with the other UI test specific settings.

jrbenny35: @eviljeff if a local instance user was to create the site and try an upload, it would be the same problem.

diox: That's weird: I don't see the issue locally, I'm able to upload and download xpis just fine.

jrbenny35: @diox so where does ```srv``` come from? I don't ever see that directory created

jrbenny35:
I found this https://github.com/mozilla/addons-nginx/blob/master/addons.conf#L16

docker-compose.yml
diox: I don't know, I haven't looked, I'm just mentioning that it's already working for me. The directory exists in my `nginx` container and it's full of stuff.

diox: From what I can see `/srv` already exists in the base image, and then those sub-directories exist... precisely because they are mounted through the `docker-compose.yml` directives you wanted to change here.

jrbenny35: Hmm, just tried again and it seems to work. Maybe my docker was having a case of the Fridays. Thanks for all the input.

PR #13058, opened 2019-12-06
handle versions not being 0 - * and existing blocks list needing refresh

### LINKED ISSUE #12991: allow multiple blocks to be "edited"
@wagnerand mentioned on irc that he believes there will be a need in the future to bulk edit blocks.

The current multiple block submission process should work as a basis for this.  It needs:
- #12936 to be done so arbitrary max/min version values can be specified
- changes to overcome (or remove) the filtering out of Blocks that are already 0 - *.  The admin may want to reduce some blocks that are currently 0 - * to cover a smaller range.

One idea I had was that the form wouldn't save but redirect back (similar to how form errors cause the instances not to save) if the min/max versions were changed from 0 - *, with a revised list of guids that would be affected.  There might be better implementations.

eviljeff:
fixes #12991

### DIFF
--- a/src/olympia/blocklist/admin.py
+++ b/src/olympia/blocklist/admin.py
@@ -1,7 +1,7 @@
 from django.contrib import admin
 from django.core.exceptions import PermissionDenied
 from django.forms import modelform_factory
-from django.forms.fields import ChoiceField
+from django.forms.fields import CharField, ChoiceField
 from django.forms.widgets import HiddenInput
 from django.shortcuts import redirect
 from django.template.response import TemplateResponse
@@ -105,6 +105,7 @@ class MultiBlockSubmitAdmin(admin.ModelAdmin):
         'blocks_submitted_count',
         'submission_complete',
     )
+    # Only used for the change view
     fields = (
         'input_guids',
         'blocks_submitted',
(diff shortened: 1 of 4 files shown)


src/olympia/blocklist/admin.py
```diff
@@ -129,21 +138,33 @@ def has_change_permission(self, request, obj=None):
...
             form = MultiBlockForm(request.POST)
-            if form.is_valid():
+            form_valid = form.is_valid()
+            frm_data = form.cleaned_data
+            # Check if the versions specified were the ones we calculated which
+            # Blocks would be updated or skipped on.
+            versions_unchanged = (
+                frm_data['min_version'] == frm_data['existing_min_version'] and
+                frm_data['max_version'] == frm_data['existing_max_version'])
+            if form_valid and versions_unchanged:
```
diox: mmm I think you're better off using raw `form.data` - `cleaned_data` isn't necessarily complete when the form is invalid, so this is kinda dangerous to rely on this. Then you don't need to call `form.is_valid()` early.

eviljeff: I don't think it makes any difference in this case but 🤷‍♂  https://github.com/mozilla/addons-server/pull/13058/commits/3cd2d2de9137d7a3201274fea8c7bd43d0ae2b46

PR #13063, opened 2019-12-06
add a test that fails if any image file is over >200kb + clean up images

### LINKED ISSUE #13035: Add a test for image assets being too large
As a follow-up to https://github.com/mozilla/addons-server/issues/12973

We should consider if we can easily add a basic check to ensure that images that we add to AMO aren't over a certain size. 

We don't do this a lot on the server so if this is a lot of effort we might not do it, but if there's a quick win here we should.

### LINKED ISSUE #13048: Remove the .png files for the primary discovery items
Followup for https://github.com/mozilla/addons-server/issues/12973

STR:
1. Since the .png files are no longer of use these should be removed from the Image section 

Annotation 2019-12-05 141034

Reproducible on dev and stage - FF70(Win10).

eviljeff:
fixes #13035; fixes #13048 and removes a bunch of other image files that were bigger than 200kb (and were unreferenced in the codebase)


src/olympia/amo/tests/test_utils.py
```diff
@@ -254,3 +256,14 @@ def test_adds_etag_header(self):
...
+
+def test_images_are_small():
+    """A test that will fail if we accidentally include a large image."""
+    large_images = []
+    img_path = os.path.join(settings.ROOT, 'static', 'img')
+    for root, dirs, files in os.walk(img_path):
+        large_images += [
+            os.path.join(root, name) for name in files
+            if os.path.getsize(os.path.join(root, name)) > IMAGE_FILESIZE_MAX]
+    assert not large_images
```
EnTeQuAk: Taking "explicit is always better than implicit" into account, I personally would prefer to use `pytest.mark.parametrize` - especially since it'll make finding "where is this image used?"-things easier.

eviljeff: gah! I didn't open the issue so didn't see your edit.  Fwiw, using `parametrize` is 10x slower too (10 seconds vs 1 second)

Issue #12923, opened 2019-11-20
Remove code related to the yara repo


willdurand:
In #12913, we changed how yara rules are fetched (file system -> database). Once this is deployed in production and it works well, we can remove the `YARA_RULES_FILEPATH` constant and update the documentation of the private sectools.

willdurand:
looks like this was fixed already.

PR #13071, opened 2019-12-09
Remove unused settings

willdurand:
refs https://github.com/mozilla/addons-server/issues/12923


willdurand:
> You might want to follow up with ops so the settings are cleared from their env files too (though admittedly we're pretty lazy about doing this normally)

Yep, that's why I did not want to close the issue with this patch. Thanks!

PR #13087, opened 2019-12-11
Add bulk block delete input; refactor bulk guid input

### LINKED ISSUE #12459: Multiple add-on block removal

Enter multiple GUIDs to have their blocks cleared. Details in the mocks.

eviljeff:
fixes #12459

### DIFF
--- a/src/olympia/blocklist/forms.py
+++ b/src/olympia/blocklist/forms.py
@@ -1,0 +1,44 @@
+from django import forms
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext_lazy as _
+
+from olympia.addons.models import Addon
+from olympia.blocklist.models import Block
+
+
+class MultiGUIDInputForm(forms.Form):
+    guids = forms.CharField(
+        widget=forms.Textarea(attrs={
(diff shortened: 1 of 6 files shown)


src/olympia/blocklist/forms.py
```diff
@@ -0,0 +1,44 @@
...
+            'cols': '80', 'rows': '10', 'required wrap': 'off'}))
+
+
+class MultiDeleteForm(MultiGUIDInputForm):
+    existing_block_qs = None
+
+    def clean(self):
+        guids = self.cleaned_data.get('guids', '').splitlines()
+        if len(guids) >= 1:
+            self.existing_block_qs = Block.objects.filter(guid__in=guids)
```
EnTeQuAk: Let's add a comment here (or above where it's initially defined) that the queryset will be forwarded to `admin.actions.delete_selected` and that's why we need the actual queryset and can't do `Block.objects.filter(...).values_list('guid', flat=True)` which would be slightly more efficient.

eviljeff: https://github.com/mozilla/addons-server/pull/13087/commits/2b35a33f085ed5d04ea8e5780b319fee6a099824

Issue #13099, opened 2019-12-13
Collect the number of customs/wat results with matches (statsd)


willdurand:
It'd be useful to know when a scanner (that is not Yara) matched "something". We are already doing this for Yara (#12577) so let's do this for the other scanners too.

ioanarusiczki:
@willdurand Are the instructions for qa the same as those from https://github.com/mozilla/addons-server/issues/12577#issue-504905208 (except that it's customs and wat to be tested instead of yara) ?

willdurand:
> @willdurand Are the instructions for qa the same as those from [#12577 (comment)](https://github.com/mozilla/addons-server/issues/12577#issue-504905208) (except that it's customs and wat to be tested instead of yara) ?

ah sorry, yes!

ioanarusiczki:
@willdurand 
- On the scanners side I did not find any problems for wat and custom scanners on dev. 

- using the [API](https://addons-server.readthedocs.io/en/latest/topics/api/signing.html#put--api-v4-addons-(string-guid)-versions-(string-version)-) I tried to upload a new version using the `channel` listed (the first version is unlisted and uploaded using dev hub). 
PUT https://addons-dev.allizom.org/api/v4/addons/{c67f9ca3-ea61-4dcd-9274-ef9ee32e431b}/versions/5.0/
I receive an error     "error": "You cannot add a listed version to this addon via the API due to missing metadata. Please submit via the website"

 The scenario on stage is not reproducible though (with PUT I could add a new `listed` version) for the add-on below
https://addons.allizom.org/api/v4/addons/{3a80c837-54e9-4cd9-a990-74e9290bb7d1}/versions/4.0/

ioanarusiczki:
@willdurand Nevermind! It's the same on stage meaning submission through API for a listed version is not possible if the previous versions are `unlisted`. The API documentation is a bit unclear here.

PR #13100, opened 2019-12-13
Collect the number of customs/wat results with matches

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13099

---

Not sure why I did not think about adding this sooner.. The test case has been updated so that it is both "successful" and we matched a rule.


willdurand:
Test failures don't seem related.

PR #13105, opened 2019-12-13
each block import to a celery task and don't import existing blocks

### LINKED ISSUE #13104: batch import_blocklist processing and don't try to reimport existing blocks
#12460 was a little naively written - it assumed that regex searches on the database would be relatively speedy, and the command was pretty much guaranteed to finish so would only ever need to be executed once per env.  The experience from addons-dev is some of the regexs take a very long time to run and not all blocks are processed before the command times out - either a database timeout or the dev instance is replaced due to a new deploy.  And then if you start it again the blocks have to be processed again from the beginning.

eviljeff:
fixes #13104 - most of the code change is moving the logic from the management command to a task

### DIFF
--- a/src/olympia/blocklist/tasks.py
+++ b/src/olympia/blocklist/tasks.py
@@ -1,7 +1,22 @@
+import re
+import time
+from datetime import datetime
+
+import olympia.core.logger
+from olympia import amo
+from olympia.addons.models import Addon
 from olympia.amo.celery import task
 from olympia.amo.decorators import use_primary_db
+from olympia.users.utils import get_task_user
 
-from .models import MultiBlockSubmit
+from .models import Block, MultiBlockSubmit
+from .utils import block_activity_log_save
+
+
+log = olympia.core.logger.getLogger('z.amo.blocklist')
+
+bracket_open_regex = re.compile(r'(?<!\\){')
+bracket_close_regex = re.compile(r'(?<!\\)}')
 
 
 @task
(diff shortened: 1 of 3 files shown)


src/olympia/blocklist/tasks.py
```diff
@@ -10,3 +25,60 @@ def create_blocks_from_multi_block(multi_block_submit_id, **kw):
...
     obj.save_to_blocks()
+
+
+@task
+@use_primary_db
+def import_block_from_blocklist(record):
+    log.debug('Processing block id: [%s]', record.get('id'))
+    guid = record.get('guid')
+    if not guid:
+        log.error('GUID is falsey, skipping.')
```
willdurand: Is this supposed to happen? If not, shouldn't we specify the `id` in this log message so that it is easier to debug if we need to.

eviljeff: None of the blocklist records currently on kinto have a empty guid - and new blocks require a guid afaik - so it's it's very unlikely to happen.  I was counting on the previous log entry being `'Processing block id: [%s]', record.get('id')` so it could be traced from there in the worst case.

willdurand: > I was counting on the previous log entry being `'Processing block id: [%s]', record.get('id')` so it could be traced from there in the worst case.

How do you guarantee the exact order? The command fires a new task for each record.

eviljeff: well, it's 3 lines above in the same task so... 😄

PR #13122, opened 2019-12-17
Fix #13088: lower hotness threshold for static themes

### LINKED ISSUE #13088: Lower "hotness" requirement for themes
From @diox: "the 2 conditions to have a hotness score > 0 are: number of users > 1000 and gained at least one user in the past 3 weeks".

This works okay for extensions, but it doesn't work well for themes. I suggest we lower the requirement for themes to 250 users. [66 themes created this year](https://sql.telemetry.mozilla.org/queries/66902/source) meet that bar.

ragnarokatz:
      PRs open for the same issue.
* [x] Add tests to cover the changes added in this PR.

Fixes issue #13088 

- Change function logic suggested by @diox in in lowering hotness requirement for static themes from 1000 down to 250 (the number represent the average daily value for this week)
- Update function docstring to reflect logic change
- Add more test cases to test out different scenarios (not hitting the threshold from either static theme or extension, hitting the threshold for static theme but not extension, hitting the threshold for both)

### DIFF
--- a/src/olympia/addons/cron.py
+++ b/src/olympia/addons/cron.py
@@ -168,7 +168,8 @@ def deliver_hotness():
 
     a = avg(users this week)
     b = avg(users three weeks before this week)
-    hotness = (a-b) / b if a > 1000 and b > 1 else 0
+    threshold = 250 if addon type is theme, else 1000
+    hotness = (a-b) / b if a > threshold and b > 1 else 0
     """
     frozen = set(f.id for f in FrozenAddon.objects.all())
     all_ids = list((Addon.objects.filter(status__in=amo.REVIEWED_STATUSES)
@@ -189,7 +190,8 @@ def deliver_hotness():
 
             # Update the hotness score but only update hotness if necessary.
             # We don't want to cause unnecessary re-indexes
-            if this > 1000 and three > 1:
+            threshold = 250 if addon.type == amo.ADDON_STATICTHEME else 1000
+            if this > threshold and three > 1:
                 hotness = (this - three) / float(three)
                 if addon.hotness != hotness:
                     addon.update(hotness=(this - three) / float(three))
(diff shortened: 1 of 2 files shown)


ragnarokatz:
hello, so the codestyle check is the only thing that has failed, and I can reproduce the errors generated by travisCI by running `flake8` locally as well.

```
flake8 src/ services/ tests/

src/olympia/addons/tests/test_cron.py:385:80: E501 line too long (84 > 79 characters)
src/olympia/addons/tests/test_cron.py:445:5: E303 too many blank lines (2)
src/olympia/addons/tests/test_cron.py:459:80: E501 line too long (89 > 79 characters)
src/olympia/addons/tests/test_cron.py:460:80: E501 line too long (83 > 79 characters)
src/olympia/addons/tests/test_cron.py:466:5: E303 too many blank lines (2)
```

I can fix the ones with too many blank lines. I can fix the first line too long as well. So 385, 445, and 466 will be fixed shortly.

However, I remember someone was telling me that wrapping a long assert statement around with brackets will cause the assert to malfunction, so something like:

from:
```
        assert (self.barely_popular_static_theme.reload().hotness == 0.0058309523809523135)
```

to:
```
        assert (
            self.barely_popular_static_theme.reload().hotness ==
            0.0058309523809523135)
```
will cause the assert to malfunction.

Can someone confirm this? Also, if it is indeed the case, what's the correct way to wrap around a long assert statement that has exceeded 79 characters? Thanks.

@jvillalobos @diox

ragnarokatz:
An additional improvement is to move the magic numbers `250` and `1000` into `constants` module. But in the short amount of time I spent on this project so far, I'm not sure what an appropriate place is to put the variables. I'd probably name them as `HOTNESS_THRESHOLD_ADDON` and `HOTNESS_THRESHOLD_STATICTHEME` or something similar.

ragnarokatz:
It seems that the `TOXENV=es` server is having some trouble connecting to `apt.cache.travis-ci.com`:
https://travis-ci.org/mozilla/addons-server/jobs/626276709?utm_medium=notification&utm_source=github_status

eviljeff:
re: `assert self.barely_popular_static_theme.reload().hotness == 0.0058309523809523135`

Normally, we'd write it as 
```
        assert self.barely_popular_static_theme.reload().hotness == (
            0.0058309523809523135)
```
but it's fine

diox:
Since we don't reference this elsewhere I don't think we need to bother with constants.

PR #13170, opened 2020-01-02
Delete add-ons with no versions as part of garbage collection cron

### LINKED ISSUE #13166: Delete add-ons with no versions as part of garbage collection cronjob
Because of https://github.com/mozilla/addons-server/issues/7135 some `Addon` might be created with no `Version` attached. These incomplete `Addon` instances stay around for no good reason.

#7135 should be fixed but in the meantime, we should delete those extraneous incomplete add-ons from the database, a bit like we delete stale `FileUpload`s. Funny enough, `amo.cron.gc` used to do this 9 years ago, but [it was reverted](https://github.com/jbalogh/zamboni/commit/4e2c688) with this comment added: 
> Incomplete addons cannot be deleted here because when an addon is
> rejected during a review it is marked as incomplete. See [bug 670295](https://bugzilla.mozilla.org/show_bug.cgi?id=670295).

Indeed, we should do more than just checking for `incomplete` status, especially now that pure unlisted add-ons also have this status.

The rules should be something like:
- Add-on with `STATUS_NULL` status
- Created more than 7 days ago
- No versions attached at all, even soft-deleted (this is the important bit)

diox:
Fixes #13166

### DIFF
--- a/src/olympia/amo/cron.py
+++ b/src/olympia/amo/cron.py
@@ -10,6 +10,8 @@ import olympia.core.logger
 
 from olympia import amo
 from olympia.activity.models import ActivityLog
+from olympia.addons.models import Addon
+from olympia.addons.tasks import delete_addons
 from olympia.amo.utils import chunked
 from olympia.bandwagon.models import Collection
 from olympia.constants.base import VALID_ADDON_STATUSES, VALID_FILE_STATUSES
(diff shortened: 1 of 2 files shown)


diox:
Yeah. It's no different that what would happen if the author would delete the incomplete add-on themselves, so I think we're good.

PR #13178, opened 2020-01-05
Add link to developer agreement page in corresponding API error

### LINKED ISSUE #11746: Include link to developer agreement page in corresponding API error
### Is this a feature request or a bug?

Bug.

### What is the current behavior?

I guess the problem might be on the server-side, but I'm not sure where to report that... Anyway, my CI server has been signing and publishing new versions of my extension for quite a while now, successfully. Trying it again today suddenly [turned up](https://gitlab.com/Flockademic/plaudit/-/jobs/242630620) the following error after running `web-ext sign --artifacts-dir web-ext-artifacts --channel=unlisted --id={***} --api-key=*** --api-secret=***`:

> Error: Received bad response from the server while requesting https://addons.mozilla.org/api/v3/addons/%7B82c27c82-29f7-477c-8588-6255c1b59a2a%7D/versions/0.1.27build242630620/
> 
> status: 403
> response: {"detail":"Before starting, please read and accept our Firefox Add-on Distribution Agreement as well as our Review Policies and Rules. The Firefox Add-on Distribution Agreement also links to our Privacy Notice which explains how we handle your information."}

### What is the expected or desired behavior?

At least for a link to https://addons.mozilla.org/en-US/developers/addon/api/key/ to be present, along with a message indicating why it had to be signed again - or not requesting me to sign it again.

It appears this did not affect just me: https://stackoverflow.com/questions/56338055/where-i-should-accept-firefox-add-on-distribution-agreement

### Version information (for bug reports)

* **Firefox version**: 67.0.3
* **Your OS and version**: Ubuntu 18.04
* Paste the output of these commands:

````
node --version && npm --version && web-ext --version
v10.15.3
6.4.1
3.1.0
````

### PR DESCRIPTION
**Pull request by mirefly**
Fixes #11746 
I tried to follow the suggestions in https://github.com/mozilla/addons-server/issues/11746#issuecomment-562966751.

### DIFF
--- a/src/olympia/users/models.py
+++ b/src/olympia/users/models.py
@@ -650,7 +650,14 @@ class DeniedName(ModelBase):
         return any(n in name for n in blocked_list)
 
 
-class IPNetworkUserRestriction(ModelBase):
+class GetErrorMessageMixin():
+
+    @classmethod
+    def get_error_message(cls, is_api):
+        return cls.error_message
+
+
+class IPNetworkUserRestriction(GetErrorMessageMixin, ModelBase):
     id = PositiveAutoField(primary_key=True)
     network = CIDRField(
         blank=True, null=True,
@@ -686,7 +693,7 @@ class IPNetworkUserRestriction(ModelBase):
 
         for restriction in restrictions:
             if (remote_addr in restriction.network or
-               user_last_login_ip in restriction.network):
+                    user_last_login_ip in restriction.network):
                 log.info('Restricting request from %s %s, %s %s (%s)',
                          'ip', remote_addr,
                          'last_login_ip', user_last_login_ip,
@@ -711,7 +718,8 @@ class NormalizeEmailMixin:
         return normalized_email
 
 
-class EmailUserRestriction(NormalizeEmailMixin, ModelBase):
+class EmailUserRestriction(
+        GetErrorMessageMixin, NormalizeEmailMixin, ModelBase):
     id = PositiveAutoField(primary_key=True)
     email_pattern = models.CharField(
         _('Email Pattern'),
@@ -760,7 +768,7 @@ class EmailUserRestriction(NormalizeEmailMixin, ModelBase):
         return True
 
 
-class DisposableEmailDomainRestriction(ModelBase):
+class DisposableEmailDomainRestriction(GetErrorMessageMixin, ModelBase):
     domain = models.CharField(
         unique=True,
         max_length=255,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/olympia/users/models.py [hunk 1]
```diff
@@ -864,13 +873,21 @@ def allow_request(cls, request):
...
                       ' Add-on Distribution Agreement as well as our Review'
                       ' Policies and Rules. The Firefox Add-on Distribution'
                       ' Agreement also links to our Privacy Notice which'
                       ' explains how we handle your information.')
 
+    @classmethod
+    def get_error_message(cls, is_api):
+        if is_api:
+            url = 'https://addons.mozilla.org/en-US/developers/addon/api/key/'
+            return ('{} Please see {}'.format(cls.error_message, url))
```
diox: This should be a translated string as well, so it needs to be wrapped in `_()`. But more importantly, thinking about this more, I don't think we need the full message for the API. Developers should only stumble upon this message if they have already accepted the developer agreement but the document has changed since they agreed to it, and they'll see the full error message when they go to the URL.

So, I propose not re-using the message, and doing this:
```suggestion
            return _('Please read and accept our Firefox on Distribution '
                     'Agreement as well as our Review Policies and Rules '
                     'by visiting {url}'.format(url=url))
```

diox:
`es` travis job failure is unrelated. Thanks for taking care of this, @mirefly !

src/olympia/users/models.py
```diff
@@ -864,13 +873,23 @@ def allow_request(cls, request):
...
     error_message = _('Before starting, please read and accept our Firefox'
                       ' Add-on Distribution Agreement as well as our Review'
                       ' Policies and Rules. The Firefox Add-on Distribution'
                       ' Agreement also links to our Privacy Notice which'
                       ' explains how we handle your information.')
 
+    @classmethod
+    def get_error_message(cls, is_api):
+        if is_api:
+            url = 'https://addons.mozilla.org/en-US/developers/addon/api/key/'
```
diox: Nit: let's not hardcode the URL.

```suggestion
            from olympia.amo.templatetags.jinja_helpers import absolutify
            url = absolutify(reverse('devhub.api_key'))
```
Requires to fix to `test_has_permission_user_has_not_read_agreement_when_using_API` test accordingly (tests run with the `http://testserver` URL prefix, so the URL will be `http://testserver/en-US/developers/addon/api/key/` in tests)

mirefly:
@diox Thanks for reviewing and helpful suggestions. I pushed the updates.

PR #13210, opened 2020-01-09
Add models & tasks to run arbitrary yara rules on existing versions

### LINKED ISSUE #13205: Add models & tasks to run arbitrary yara rules on existing versions
We want to allow searching for code patterns on existing add-ons from the admin. The idea is to use our scanners models & tasks to do this, and yara in particular.

To do this, first we need to refactor the existing models & tasks to be able to run an arbitrary rule on a `Version` (and not just a `FileUpload`) without polluting current results.

diox:
Fixes #13205


src/olympia/scanners/models.py
```diff
@@ -177,6 +231,51 @@ def clean_yara(self):
...
+
+class ScannerQueryResult(AbstractScannerResult):
+    # Has to be overridden, because the parent refers to ScannerMatch.
+    matched_rules = models.ManyToManyField(
+        'ScannerQueryRule', through='ScannerQueryMatch'
+    )
+
+    class Meta(AbstractScannerResult.Meta):
+        db_table = 'scanners_query_results'
+        # FIXME indexes, unique constraints ?
```
willdurand: Is this comment for later?

diox: Yeah. I'm probably going to add more things, and I didn't want to do multiple migrations fiddling with the indexes just now, so I left it for later.

Issue #13181, opened 2020-01-06
Dockerfile build fails to add Geolite2 database dependency


jrbenny35:
This line is failing as it seems the URL is not resolving correctly.

https://github.com/mozilla/addons-server/blob/master/Dockerfile#L59

The integration tests on addons-frontend depend on this Dockerfile being built. https://circleci.com/gh/mozilla/addons-frontend/18914

eviljeff:
I guess it's because of https://blog.maxmind.com/2019/12/18/significant-changes-to-accessing-and-using-geolite2-databases/

eviljeff:
@diox because I believe this was added for #10720

eviljeff:
The same line isn't in `Dockerfile.deploy` so I don't think it will block dev/stage/prod deployments (though I can't see how the geo db is made available in those environments)

diox:
In dev/stage/prod it's being copied from a special ops repo using a private ansible rule.

jrbenny35:
Is there a fix for this planned? Looks like some extra steps need to happen as far as getting a key and whatnot

diox:
We could replace it by a dummy database. We don't really need a real one for our dev envs.

AlexandraMoga:
Please add a qa:not needed label to this issue.
Thanks

PR #13212, opened 2020-01-09
Update Dockerfile to no longer fetch geoip database

diox:
It's not needed for local envs. Geoip queries will fail, but that should be ok.

Fixes https://github.com/mozilla/addons-server/issues/13181


jrbenny35:
Thanks for this @diox

PR #13230, opened 2020-01-11
Fix TestBlockAdminAdd and TestMultiBlockSubmitAdmin

### LINKED ISSUE #13004: fix TestBlockAdminAdd intermittent test fail
```
_____________________ TestBlockAdminAdd.test_review_links ______________________

[gw0] linux -- Python 3.6.7 /home/travis/build/mozilla/addons-server/.tox/main/bin/python3

src/olympia/blocklist/tests/test_admin.py:179: in test_review_links

    assert 'Review Unlisted' in content, content
```

mirefly:
Fixes #13004 
These two tests failed when `version_factory` happened to generate two versions with the same [`version_str`](https://github.com/mirefly/addons-server/blob/fix-TestBlockAdminAdd-test-fail/src/olympia/amo/tests/__init__.py#L856) in one test. Since [`addon_versions`](https://github.com/mirefly/addons-server/blob/fix-TestBlockAdminAdd-test-fail/src/olympia/blocklist/models.py#L78) in `Block` used `version_str` as the key, only one version was actually loaded.
I guess we would not have versions with same `version_str` in one addon in reality (?). So I just pass the `version_str` into `version_factory` and `addon_factory` to fix these two tests.

### DIFF
--- a/src/olympia/blocklist/tests/test_admin.py
+++ b/src/olympia/blocklist/tests/test_admin.py
@@ -175,14 +175,16 @@ class TestBlockAdminAdd(TestCase):
         self.grant_permission(user, 'Reviews:Admin')
         self.client.login(email=user.email)
 
-        addon = addon_factory(guid='guid@', name='Danger Danger')
+        addon = addon_factory(
+            guid='guid@', name='Danger Danger', version_kw={'version': '0.1'})
         response = self.client.get(
             self.single_url + '?guid=guid@', follow=True)
         content = response.content.decode('utf-8')
         assert 'Review Listed' in content
         assert 'Review Unlisted' not in content  # Theres only a listed version
 
-        version_factory(addon=addon, channel=amo.RELEASE_CHANNEL_UNLISTED)
+        version_factory(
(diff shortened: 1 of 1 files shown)


diox:
Impressive. Most impressive.

diox:
> I guess we would not have versions with same `version_str` in one addon in reality

Correct. I forgot why we don't have a database constraint for this (is there a good reason any more? maybe not) but yeah, this shouldn't happen - we have code preventing this.

We've had intermittent failures caused by this exact thing before, but I had completely forgotten about them.

@eviljeff or I will review this on Monday, but I wanted to be the first to congratulate you, I'm impressed!

eviljeff:
> I guess we would not have versions with same version_str in one addon in reality

I also forgot this was a possiblity (or that `addon_versions` needed to rely on the version strings).  Fixing `version_factory` to never generate a version string that's already in use for that add-on is the follow-up.

> I forgot why we don't have a database constraint for this (is there a good reason any more? maybe not)

I can't think of a reason now.  (Though we may be disproved during the database migration to add one).

Issue #13220, opened 2020-01-10
Include list of matched files in customs result


wagnerand:
It would be nice if the customs scan result page would also include the list of matched files, next to the list of matched rules, ideally linking to the file in the file viewer.

Note that for now, we might want to point it to the old file-viewer since most unlisted files have not been extracted yet and thus code-manager doesn't see them.

willdurand:
> It would be nice if the customs scan result page would also include the list of matched files, next to the list of matched rules, ideally linking to the file in the file viewer.

Implementation idea: let's only consider `customs` (so other scanners would not have what is explained thereafter). We have the JSON with all the information we need so we could add a new row in the page (above the raw JSON one). This row should list all the files (with links to the file viewer) for each matched rule.

willdurand:
> Implementation idea: let's only consider `customs` (so other scanners would not have what is explained thereafter). We have the JSON with all the information we need so we could add a new row in the page (above the raw JSON one). This row should list all the files (with links to the file viewer) for each matched rule.

Looks like we can do this for all scanners, actually.

AlexandraMoga:
Verified fixed on -dev and stage. The 'Matched rules' section from a scanner details page is now including a direct link to the file(can be multiple files) that generated it. The link opens in the old file viewer. This will also work for disabled and deleted files, if the user has the correct permissions. 

Note that since `wat` doesn't have any rules yet, this section will be empty, but, assuming that we're going to have rules for `wat` as well, it makes no sense to hide it temporarily. 

![image](https://user-images.githubusercontent.com/31961530/72432994-0f0d1780-37a1-11ea-9868-2cf607e69a78.png)

PR #13238, opened 2020-01-13
Add the list of files related to the matched rules

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13220

---

This patch changes the way we render the matched rules in the django admin. In
the "change view" of a scanner result, we now show both the matched rules and
the associated files (in a table). When possible, we add links to the reviewer
tools.

### DIFF
--- a/src/olympia/scanners/models.py
+++ b/src/olympia/scanners/models.py
@@ -1,6 +1,8 @@
 import json
 import re
 
+from collections import defaultdict
+
 import yara
 
 from django.conf import settings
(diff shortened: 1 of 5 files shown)


src/olympia/scanners/models.py
```diff
@@ -94,6 +94,24 @@ def get_scanner_name(self):
     def get_pretty_results(self):
         return json.dumps(self.results, indent=2)
 
+    def get_files_by_matched_rules(self):
+        res = dict()
```
eviljeff: you can simplify below by using `defaultdict(list)`

src/olympia/scanners/models.py
```diff
@@ -94,6 +94,24 @@ def get_scanner_name(self):
...
+        if self.scanner is YARA:
+            for item in self.results:
+                if item['rule'] not in res:
+                    res[item['rule']] = []
+                res[item['rule']].append(item['meta']['filename'])
+        elif self.scanner is CUSTOMS:
+            scanMap = (self.results['scanMap'] if 'scanMap' in self.results
```
eviljeff: `scanMap = self.results.get('scanMap', {})`

willdurand: ha, I meant to change that, thanks!

src/olympia/scanners/models.py
willdurand: oh, I did not know that, thanks!

Issue #13239, opened 2020-01-13
Add a new permission for the scanners admin


willdurand:
Let's introduce new permissions for accessing the scanners admin so that both admins and privileged users with this new permission can view the scanners rules/results. We'll have permissions to view the scanner results, view the scanner rules and edit the scanner rules.

Admins should still be able to access the scanners admin and triage the scanner results.

AlexandraMoga:
I've verified the new scanner admin permissions on -dev and I have the following results:

Prior condition: these permissions need to be combined with the `Admin:Tools` permission in order to be able to access the `/admin/models/`

1. User with `Admin:ScannersResultsView` - can view all the entries listed in Scanner results, can view a scanner result detail page; doesn't have access to Actions fields (i.e. reporting as false or true positive); doesn't have access to Scanner rules at all (i.e. trying to open a linked scanner rule will throw a 404)
1. User with `Admin:ScannersRulesView` - can view all the entries listed in Scanner rules, can view a Rule detail page; cannot add, change or delete scanner rules; doesn't have access to Scanner results at all (i.e. trying to open a linked scanner result will throw a 404)
3. User with `Admin:ScannersRulesEdit` - can view all the entries listed in Scanner rules, can view a Rule detail page, can add, change or delete scanner rules; doesn't have access to Scanner results at all

  * The above permissions can be combined to allow broader access to scanners, as needed.

4. Full Admins' access to Scanners results and rules works as before
5. Users with `Admin:Advanced` permission don't have access to Scanners anymore

@willdurand I'd like to clarify something with regards to scanner Actions (Report False positive or True Positive): the new `Admin:ScannersResultsView` permission doesn't give access to this option. However, if I'm combining this permission with `Admin:Advanced`, the Actions fields become available.
Was this intended?

willdurand:
> @willdurand I'd like to clarify something with regards to scanner Actions (Report False positive or True Positive): the new `Admin:ScannersResultsView` permission doesn't give access to this option. However, if I'm combining this permission with `Admin:Advanced`, the Actions fields become available.
> Was this intended?

Yes. I realize that we could have a `Admin:ScannersResultsEdit` instead, though (but we do not actually edit the results).

AlexandraMoga:
[...]

PR #13240, opened 2020-01-13
Add new `ADMIN_SCANNERS_*` permissions

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13239

---

This patch adds news permissions for accessing the scanners admin. Because admins have some actions to triage the scanner results, some code was added to handle permissions for those actions.

### DIFF
--- a/src/olympia/scanners/admin.py
+++ b/src/olympia/scanners/admin.py
@@ -203,12 +203,8 @@ class ScannerResultAdmin(admin.ModelAdmin):
                     reverse(
                         'reviewers.review',
                         args=[
-                            (
-                                'listed'
-                                if obj.version.channel
-                                == amo.RELEASE_CHANNEL_LISTED
-                                else 'unlisted'
-                            ),
+                            ('listed' if obj.version.channel ==
+                             amo.RELEASE_CHANNEL_LISTED else 'unlisted'),
                             obj.version.addon.id,
                         ],
                     ),
--- a/src/olympia/scanners/admin.py
+++ b/src/olympia/scanners/admin.py
@@ -13,6 +13,7 @@ from django.utils.translation import ugettext
 from urllib.parse import urljoin
 
 from olympia import amo
+from olympia.access import acl
 from olympia.addons.models import Addon
 from olympia.amo.urlresolvers import reverse
 from olympia.constants.scanners import (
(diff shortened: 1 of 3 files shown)


src/olympia/scanners/admin.py
```diff
@@ -178,6 +179,20 @@ def has_delete_permission(self, request, obj=None):
     def has_change_permission(self, request, obj=None):
         return False
 
+    def get_list_display(self, request):
+        fields = super().get_list_display(request)
+        return self._excludes_admin_fields(request=request, fields=fields)
+
+    def get_fields(self, request, obj=None):
```
bobsilverberg: I see tests for `test_list_view`, but I'm not sure if those are testing `get_fields` or `get_list_display` (or both, although that seems unlikely). I guess my question is: do we have added coverage for both of these methods?

src/olympia/scanners/tests/test_admin.py
```diff
@@ -430,13 +453,52 @@ def test_handle_revert_report(self):
...
+        assert response.status_code == 404
+
 
 class TestScannerRuleAdmin(TestCase):
     def setUp(self):
         super().setUp()
 
         self.user = user_factory()
-        self.grant_permission(self.user, 'Admin:Advanced')
+        self.grant_permission(self.user, 'Admin:*')
```
bobsilverberg: As above, why this change?

src/olympia/scanners/admin.py
```diff
@@ -178,6 +179,20 @@ def has_delete_permission(self, request, obj=None):
...
+        fields = super().get_fields(request, obj)
+        return self._excludes_admin_fields(request=request, fields=fields)
+
+    def _excludes_admin_fields(self, request, fields):
+        is_admin = acl.action_allowed(request, amo.permissions.ADMIN_ADVANCED)
+        if not is_admin:
+            return list(filter(lambda x: x != 'result_actions', fields))
```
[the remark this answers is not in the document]
willdurand: It is indeed only used to exclude this field (for now). And yes, this field contains admin links.

src/olympia/scanners/admin.py
```diff
@@ -239,7 +254,8 @@ def formatted_matched_rules(self, obj):
     formatted_matched_rules.short_description = 'Matched rules'
 
     def handle_true_positive(self, request, pk, *args, **kwargs):
-        if request.method != "POST":
+        is_admin = acl.action_allowed(request, amo.permissions.ADMIN_ADVANCED)
+        if not is_admin or request.method != "POST":
```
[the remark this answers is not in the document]
willdurand: There was no check previously because only admins could get to this page (behind a VPN). Adding a new permission made me think that now would be a good time to actually secure those actions, too.

src/olympia/scanners/tests/test_admin.py
```diff
@@ -39,7 +39,7 @@ def setUp(self):
...
```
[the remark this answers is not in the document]
willdurand: That's because most test cases assume the current user is an admin. It was working before because we only one permission but now we have more permissions. Admins have `Admin:*` (or rather `*:*`) so that should not change anything for them.

FWIW, this would be a problem for users with `Admin:Advanced` only (but I don't think we have that) and admins would need to give `Admin:ScannersResultsView`, etc. to these users.

src/olympia/scanners/admin.py
willdurand: We don't for this one specifically, but both use the same (private) method, which is tested _via_ the `test_list_diplay()` and `test_list_display_for_non_admins()`.

src/olympia/scanners/tests/test_admin.py
willdurand: Same reason, to "fix" existing test cases.

Issue #13248, opened 2020-01-14
Wrong URL to the file viewer in the scanners admin


willdurand:
The links pointing to the files matched by the rules in the scanners admin are not absolute and they are not correct in the hosted envs (-dev, etc.). The links should open the (old) file viewer.

Edit: looks like the ID is not correct either..

AlexandraMoga:
This was verified in https://github.com/mozilla/addons-server/issues/13220#issuecomment-574633703 both on -dev and stage.

PR #13249, opened 2020-01-14
Use absolute urls for the files linked in the scannerresult change view

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13248

---

This patch should fix the issue mentioned above by:

- adding the external base URL to each URL to a file
- changing the ID in the generated URL (from addon ID to file ID)

### DIFF
--- a/src/olympia/scanners/templates/admin/scanners/scannerresult/formatted_matched_rules_with_files.html
+++ b/src/olympia/scanners/templates/admin/scanners/scannerresult/formatted_matched_rules_with_files.html
@@ -13,7 +13,7 @@
     <td>
       {% for file in rule.files %}
         {% if addon_id %}
-          <a href="{% url 'files.list' addon_id 'file' file %}">{{ file }}</a>
+          <a href="{{ external_site_url }}{% url 'files.list' addon_id 'file' file %}">{{ file }}</a>
         {% else %}
           {{ file }}
         {% endif %}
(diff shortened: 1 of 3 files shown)


willdurand:
For some reasons, I cannot load the file viewer with a file in the URL locally. It works without a file, but not with one.. In addition, I cannot browse a file (fragment?) once the file viewer is loaded locally. The generated URLs look correct, though.

src/olympia/scanners/templates/admin/scanners/scannerresult/formatted_matched_rules_with_files.html
```diff
@@ -12,8 +12,8 @@
     <td><a href="{% url 'admin:scanners_scannerrule_change' rule.pk %}">{{ rule.name }}</a></td>
     <td>
       {% for file in rule.files %}
-        {% if addon_id %}
-          <a href="{% url 'files.list' addon_id 'file' file %}">{{ file }}</a>
+        {% if file_id %}
+          <a href="{{ external_site_url }}{% url 'files.list' file_id 'file' file %}">{{ file }}</a>
```
diox: We've got `|absolutify` filter for this... though I'm not sure how we can use it here, since I assume this is a django template and not a jina one ?

willdurand: I saw the `absolutify` filter but I could not make it work :/

diox: Yeah let's not worry about it for now.

Issue #13143, opened 2019-12-20
Hide matched rules in the scanner results admin


willdurand:
It is currently possible to disable a rule, which actually means disabling the effect of the rule (in other words, the execution of the action that is bound to it).

In case a rule flags too many things (inaccurately), we'd like to be able to hide the results for this rule in the admin list.

willdurand:
@diox @wagnerand thinking about this again, I guess using the `is_active` boolean attribute _could_ work here too. We could check `is_active` when binding rules to the scanner results so that:

1. a disabled rule won't be seen by default because the scanner result would have `has_matches=False`
2. a disable rule won't trigger an action (current behavior)

Implementing (1) would mean that a scanner result would not be associated to a disabled rule in the database. Given that we have all the code needed to do this anyway, we could theoretically rebuild the relationships between a result and the rules if needed.

Only disabling the effect of a rule could be done by using the "no action" action, if needed.

diox:
I like that plan, yes.

wagnerand:
Sounds good!

AlexandraMoga:
I've submitted add-ons that should have matched rules both for `yara` and `customs` and I have the following results:

With all rules disabled:
- no matched rules are present in scanners results list
- no matched rules are listed in scanner details 
- no review actions attached to any of the rules are executed 

With one or more rules enabled for either `yara` or `customs`:
- matched rules are present in scanners results list
- matched rules are listed in scanner details 
- review actions attached to the matched rules are executed 


@willdurand I have the following sitation:
With all rules disabled, in scanner _details_, [customs](https://addons-internal.dev.mozaws.net/en-US/admin/models/scanners/scannerresult/6068/change/?_changelist_filters=has_matched_rules%3Dall) and [wat](https://addons-internal.dev.mozaws.net/en-US/admin/models/scanners/scannerresult/6069/change/?_changelist_filters=has_matched_rules%3Dall) will still have the 'Results' field populated, while for [yara](https://addons-internal.dev.mozaws.net/en-US/admin/models/scanners/scannerresult/6067/change/?_changelist_filters=has_matched_rules%3Dall) this field is empty. In `customs` case, you can even see which rules were matched as `true`. I'm not sure if this was expected or not.

willdurand:
> @willdurand I have the following sitation:
[...]

PR #13252, opened 2020-01-14
Do not bind a disabled rule to a scanner result

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13143

---

Previously, we were binding a scanner rule to a scanner result even if the
scanner rule was deactivated. With this patch, the scanner rule must be active.


src/olympia/scanners/models.py
```diff
@@ -81,7 +81,10 @@ def extract_rule_names(self):
...
         rule_model = self._meta.get_field('matched_rules').related_model
         matched_rules = rule_model.objects.filter(
-            scanner=self.scanner, name__in=self.extract_rule_names()
+            scanner=self.scanner,
+            name__in=self.extract_rule_names(),
+            # See: https://github.com/mozilla/addons-server/issues/13143
+            is_active=True,
         )
         self.has_matches = bool(matched_rules)
         # Save the instance first...
```
eviljeff: I'm aware this isn't new functionality in this commit, but wouldn't this add a new `matched_rule` _each_ time a `AbstractScannerResult` instance is saved?  So the second time the instance is saved it'd create a new `ScannerMatch` record, even though it already exists - or is there some de-duplication going on elsewhere?

willdurand: Good question, I assume this is handled _via_ the ORM. I'll double-check and file an issue if needed. /cc @diox

eviljeff: There's no built-in duplication, afaik.  Either there will be duplicate many-to-many links on subsequent saves, or a database error, depending on what constraints there are in place around uniqueness.

diox: There is, in fact. Django does this for us. I've checked it when we introduced it, because I also thought this was wrong, but it's fine :) It's documented in https://docs.djangoproject.com/en/3.0/topics/db/examples/many_to_many/:
> Adding a second time is OK, it will not duplicate the relation

You can test it locally, it doesn't create additional matches for dupes.

willdurand: Thanks for confirming. That sounds like something an ORM should do (otherwise, what's the point? :P).

I think we only save the scanner results once anyway (we don't provide tools to update these models) but that was a good point to double check.

PR #13257, opened 2020-01-14
Add task to run a yara rule on all versions in database

### LINKED ISSUE #13256: Add task to run a yara rule on all versions in database
Follow-up to https://github.com/mozilla/addons-server/issues/13205, we need to implement a task that runs the task for a query rule on all existing versions that we might be interested in.

The default criteria, hardcoded for now, will be:
- Must belong to a webextension (and not be a static theme, only extensions are considered)
- The add-on must not have been disabled by Mozilla
- The version needs either to be unlisted or be the currently public listed version of the addon it belongs to

diox:
When triggered, this task creates a chord to do the work and triggers *that*.

Fixes #13256


src/olympia/scanners/admin.py
```diff
@@ -368,3 +370,29 @@ def matched_results_link(self, obj):
...
+    pass
+
+
+@admin.register(ScannerRule)
+class ScannerRuleAdmin(AbstractScannerRuleAdminMixin, admin.ModelAdmin):
+    pass
+
+
+@admin.register(ScannerQueryRule)
+class ScannerQueryRuleAdmin(AbstractScannerRuleAdminMixin, admin.ModelAdmin):
```
diox: This and the one above are not very useful right now as they don't expose a way to trigger an action yet, and what they are displaying isn't necessarily what we want, but I needed them for easier debugging.

src/olympia/lib/settings_base.py
```diff
@@ -1119,6 +1119,11 @@ def get_db_config(environ_var, atomic_requests=True):
     'olympia.scanners.tasks.run_customs': {'queue': 'devhub'},
     'olympia.scanners.tasks.run_wat': {'queue': 'devhub'},
     'olympia.scanners.tasks.run_yara': {'queue': 'devhub'},
+    'olympia.scanners.mark_yara_query_rule_as_completed': {'queue': 'devhub'},
+    'olympia.scanners.run_yara_query_rule': {'queue': 'devhub'},
+    'olympia.scanners.run_yara_query_rule_on_versions_chunk': {
+        'queue': 'devhub'
+    },
```
willdurand: Shouldn't we have a new queue for the code search (so that the regular devhub tasks are not impacted)?

src/olympia/scanners/admin.py
```diff
@@ -332,8 +334,7 @@ def result_actions(self, obj):
     result_actions.allow_tags = True
 
 
-@admin.register(ScannerRule)
-class ScannerRuleAdmin(admin.ModelAdmin):
+class AbstractScannerRuleAdminMixin(admin.ModelAdmin):
```
willdurand: An abstract mixin :p

src/olympia/scanners/tasks.py
```diff
@@ -221,29 +220,82 @@ def _run_yara_for_path(scanner_result, path, definition=None):
...
 @task
 @use_primary_db
-def run_yara_query_rule_on_version(query_rule_pk, version_pk):
+def mark_yara_query_rule_as_completed(query_rule_pk):
     """
-    Run a specific ScannerQueryRule on a Version.
+    Mark a ScannerQueryRule as completed
+    """
+    rule = ScannerQueryRule.objects.get(pk=query_rule_pk)
+    rule.update(state=COMPLETED)
```
willdurand: Should we handle non-existent rules?

willdurand:
One more thing: it's great to reuse most of the code between the scanner results/rules but I find it a bit confusing to use "query" for the "code search" feature. I get why it is like this and I don't have a better suggestion, though.

src/olympia/lib/settings_base.py
diox: Yeah, probably. I'll file a follow-up to change that, needs a discussion with ops to see what workers we prioritize (IIRC most queues outside the "priority" one go the same pool of workers anyway)

src/olympia/scanners/tasks.py
diox: I think we don't have to: if it doesn't exist somehow (and I'm using `use_primary_db` to avoid replica delay) then the task chord will fail, it's likely `run_yara_query_rule` will fail and the state will never be updated.

In a follow-up branch I'm adding ability to abort and that should take care of most error cases.

diox:
> One more thing: it's great to reuse most of the code between the scanner results/rules but I find it a bit confusing to use "query" for the "code search" feature. I get why it is like this and I don't have a better suggestion, though.

Yeah. I struggled with naming and how far to push code re-use. I think we're OK as long as we're aware that despite sharing some base code, they don't really behave the same way. I'm open to suggestions though.

diox:
I'll file follow-ups for each of your general comments, I had all that in mind and had planned to work on those things along the way but explicit issues will force us not to forget about them.

src/olympia/lib/settings_base.py
diox: https://github.com/mozilla/addons-server/issues/13267

PR #13259, opened 2020-01-14
Disable OpenSearch add-ons.

### LINKED ISSUE #12463: Disable all OpenSearch add-ons
Part of [this PRD](https://docs.google.com/document/d/1FQq47mNCRy1enLsN0giseD2tnS0OaHth3S5mZL9C8jY/edit).

We'll stop accepting submissions and reach out to devs at the beginning of the quarter. Actually disabling them should wait until the end of the quarter, to give devs time to migrate.

Depends on https://github.com/mozilla/addons-frontend/issues/8680.

EnTeQuAk:
I didn't just add it to the `delete_obsolete_addons` task to be able to
run it independently whenever we want to (afaik obsolete add-ons are
already hard deleted anyway).

Fixes #12463


src/olympia/addons/management/commands/process_addons.py
```diff
@@ -160,6 +160,14 @@ def get_recalc_needed_filters():
         ],
         'allowed_kwargs': ('with_deleted',),
     },
+    'disable_opensearch_addons': {
+        # We're re-using the `delete_addons` method but don't allow for hard
+        # deletes
```
eviljeff: Why don't we allow for hard deletes?  At some point in the next few months we will want to hard-delete all the opensearch addons we'll be soft deleting this time.  (And any soft-deleted opensearch add-ons in the database currently).

EnTeQuAk: It might be me being overly cautious but we may want to hard delete them at some point but not in the immediate future and we'll soft-delete first. That's why I separated them, not sure the cautiousness is appropriate.

Happy to change that if you think it isn't.

eviljeff: I understand why we don't want to hard-delete them now, but I don't understand why you explicitly don't want the `--with_deleted` option available, so they can be hard deleted in the future without a further patch.

EnTeQuAk:
I added the `with-deleted` option as requested.

Issue #13089, opened 2019-12-11
Dual sign-off for blocklist changes


jvillalobos:
Per requirements from Security, we'll need to implement dual sign-off for blocklist changes. This means that some changes will first go to a "pending" state that then can only be approved by a second user. Any user with access to this admin can reject a block change that is pending.

The changes that require dual sign-off are:
- Adding a new block.
- Deleting a block.
- Changing the version range for a block.

Changing the URL, Reason, or legacy state doesn't require dual sign-off.

Visiting the Edit view of a block in a "pending" state should show what the change is:

- `This is a new block, pending sign-off.`
- `This block will be deleted, pending sign-off.`
- `The version range for this block will be changed to Min version: 0 and Max version: 2.5, pending sign-off.`

And it should have an Approve Change button and Reject Change button. The history should show who requested the change. After approval or rejection, those actions should also be logged in the history.

Blocks with pending changes can't be deleted or have their version range changed again. The pending change needs to be resolved first.

We discussed some kind of notification system for this. I think it would make sense to send an email to a list if something is added to the "pending" list, and when it's approved or rejected. We could just create a Blocklist Admin user group and send the email to those users. Thoughts, @wagnerand?

eviljeff:
> Visiting the Edit view of a block in a "pending" state 

This sounds like for a multiple block action like adding new blocks you'd want the blocks created, and then someone would need to approve them individually.  The idea I had in my head was to require the _bulk create_ action to need sign-off.  So all the guids would all be listed in a single page and then the blocks would be created _after_ the sign-off.

Bulk editing existing blocks would work in a similar way.

(Adding or editing a single block would need some special handling)

jvillalobos:
Yeah, what I've been thinking is that approving a change that is part of a bulk operation would take you to the confirmation step of that operation, including all affected IDs. I'm going to update the mocks with what I'm thinking, and we can discuss if it makes sense.

eviljeff:
If the sign-off page was based around the existing multi-block model/form it would look something like: 
![signoff mockup](https://user-images.githubusercontent.com/768592/70902184-26b37d80-1ff4-11ea-9d30-bdb212a0ebdb.png)
Showing a multi-block that updates a single existing block (because the other guids I submitted were invalid or already blocked for that range)

wagnerand:
The proposal sounds good overall, we might want to change the following:

> Blocks with pending changes can't be deleted or have their version range changed again. The pending change needs to be resolved first.

[...]

PR #13266, opened 2020-01-15
blocklist dual sign off for multi-guid submission

### LINKED ISSUE #13201: implement dual sign-off for multi block add/edit
part of #13089 

The current read-only view for multiblocks needs updating to allow blocks to be confirmed.  Changes should not now be immediately applied.  In general anyway because 
> there could be a threshold for requiring dual sign-off. If all add-ons involved in a block have user counts under that threshold, the block change would be allowed without sign-off.

eviljeff:
fixes #13201 which is part of #13089 

This patch transforms the `MultiBlockSubmit` class into `BlockSubmission`, so it can handle a sign-off of the guids being submitted before the cron job handles the Blocks being created.

Because this is already quite a large series of commits I've deferred the following requested functionality to future prs/issues:
- log who signed off in Block history when the Block is created
- show/block further edits in existing Block page when a `BlockSubmission` is outstanding
- separate permissions for signoff & submission
- replace the dual sign-off waffle with a condition around addon ADU

Additionally, there was some other functionality I'd like to add that wasn't explicitly requested:
- show in GUID input too if a matching GUID is in an outstanding `BlockSubmission`
- allow rejected `BlockSubmission`'s to be edited or re-used/resubmitted easily
- rewrite `BlockSubmission.add_view` to use more vanilla `ModelAdmin` functionality like `change_view` does; use same templates and style sheets, etc.


src/olympia/blocklist/admin.py
```diff
@@ -141,14 +129,75 @@ class MultiBlockSubmitAdmin(admin.ModelAdmin):
...
+    change_form_template = 'blocklist/block_submission_change_form.html'
+
+    class Media:
+        css = {
+            'all': ('css/admin/blocklist_blocksubmission.css',)
+        }
 
     def has_delete_permission(self, request, obj=None):
+        # For now, keep all BlockSubmission records.
+        # TODO: define under what cirumstances records can be safely deleted.
```
willdurand: Is there an issue filed for this `TODO`? If not, I suggest creating one.

src/olympia/blocklist/admin.py
```diff
@@ -141,14 +129,75 @@ class MultiBlockSubmitAdmin(admin.ModelAdmin):
...
     def has_change_permission(self, request, obj=None):
-        return False
+        has = super().has_change_permission(request, obj=obj)
+        pending = obj and obj.signoff_state == BlockSubmission.SIGNOFF_PENDING
+        return has and (not obj or pending)
+
+    def has_signoff_permission(self, request, obj=None):
+        has = self.has_change_permission(request, obj=obj)
+        pending = obj and obj.signoff_state == BlockSubmission.SIGNOFF_PENDING
+        return has and obj and pending and obj.can_user_signoff(request.user)
```
willdurand: The two functions above would benefit from having comments to explain the logic behind these conditions.

eviljeff:
> The submission logs look off (it contains unrelated messages)

https://github.com/mozilla/addons-server/pull/13266/commits/146b05966c12371393c85139dbd43e3edbd83b72

src/olympia/blocklist/admin.py
eviljeff: https://github.com/mozilla/addons-server/issues/13278

src/olympia/blocklist/admin.py
eviljeff: https://github.com/mozilla/addons-server/pull/13266/commits/61755adf93508dca052d6613741371266942c175

PR #13274, opened 2020-01-16
keep track of kinto record imports with a model

eviljeff:
fixes #13105


src/olympia/blocklist/models.py
```diff
@@ -325,3 +325,24 @@ def save_to_blocks(self):
...
+
+
+class KintoImport(ModelBase):
+    OUTCOME_INCOMPLETE = 0
+    OUTCOME_BADGUID = 1
+    OUTCOME_NOTFIREFOX = 2
+    OUTCOME_BLOCK = 3
+    OUTCOME_REGEXBLOCKS = 4
+    OUTCOME_NOMATCH = 5
+    OUTCOMES = {
```
bobsilverberg: It seems unfortunate that this entire mapping (i.e., the integers to their corresponding string values) has to be duplicated in the migration. I assume that's necessary, and there's no way to import the values from this file into the migration?

src/olympia/blocklist/tasks.py
```diff
@@ -30,25 +30,30 @@ def create_blocks_from_multi_block(multi_block_submit_id, **kw):
 @task
 @use_primary_db
 def import_block_from_blocklist(record):
-    log.debug('Processing block id: [%s]', record.get('id'))
+    kinto_id = record.get('id')
+    log.debug('Processing block id: [%s]', kinto_id)
+    kin = KintoImport.objects.create(kinto_id=kinto_id, record=record)
```
bobsilverberg: This seems like an odd name for this variable (being so short and not entirely obvious that it corresponds to a `KintoImport` object), but if this matches the standards for variable naming in the project then I guess it's fine.

src/olympia/blocklist/models.py
```diff
@@ -325,3 +325,24 @@ def save_to_blocks(self):
...
+        OUTCOME_INCOMPLETE: 'Incomplete',
+        OUTCOME_BADGUID: 'Missing GUID',
```
bobsilverberg: If the meaning is "Missing GUID", would a better name for the key be `OUTCOME_MISSINGGUID`? `BADGUID` implies to me that it could be invalid and not just missing.

src/olympia/blocklist/tasks.py
```diff
@@ -60,25 +65,33 @@ def import_block_from_blocklist(record):
...
+            log.debug('Kinto %s: Added Block for [%s]', kinto_id, block.guid)
             block.update(modified=modified_date)
         else:
-            log.debug('Updated Block for [%s]', block.guid)
-    else:
+            log.debug('Kinto %s: Updated Block for [%s]', kinto_id, block.guid)
+    if addons_qs:
+        kin.update(outcome=(KintoImport.OUTCOME_REGEXBLOCKS if regex else
+                            KintoImport.OUTCOME_BLOCK))
+    if not addons_qs:
```
bobsilverberg: Could this just be an `else` paired with the `if addons_qs:` above?

src/olympia/blocklist/tests/test_commands.py
```diff
@@ -37,15 +37,22 @@ def test_empty(self):
         """
         addon_factory()
         assert Block.objects.count() == 0
+        assert KintoImport.objects.count() == 0
         call_command('import_blocklist')
         assert Block.objects.count() == 0
+        assert KintoImport.objects.count() == 6
+        assert KintoImport.objects.filter(
+            outcome=KintoImport.OUTCOME_NOTFIREFOX).count() == 1
```
bobsilverberg: How do you know to expect this? Is it just assumed that the `blocklist.json` file is set up so that one record will result in `OUTCOME_NOTFIREFOX`? Would it be possible to set this up in the test, so one can see just from looking at the test file what the valid assertions would be? Perhaps that's overly complicated and not the way these tests are generally written, but in the frontend and code-manager code bases we try as much as possible to have all of the test setup in the tests themselves so that one can see, just from the test, what the logical expectations are.

I wonder if it could be as simple as modifying the `json` after it is read from `blocklist.json`?

src/olympia/blocklist/tests/test_commands.py
```diff
@@ -64,11 +71,21 @@ def test_regex(self):
             assert block.kinto_id == '*' + this_block['id']
             assert block.include_in_legacy
             assert block.modified == datetime(2019, 11, 29, 22, 22, 46, 785000)
+        assert KintoImport.objects.count() == 6
+        assert KintoImport.objects.filter(
+            outcome=KintoImport.OUTCOME_NOMATCH).count() == 4
+        assert KintoImport.objects.filter(
+            outcome=KintoImport.OUTCOME_NOTFIREFOX).count() == 1
+        kim = KintoImport.objects.get(
```
bobsilverberg: Nit: In `src/olympia/blocklist/tasks.py` there's a variable called `kin`, and here there's one called `kim`, which both seem to refer to a `KintoImport` object.

src/olympia/blocklist/tests/test_commands.py
```diff
@@ -64,11 +71,21 @@ def test_regex(self):
             assert block.kinto_id == '*' + this_block['id']
             assert block.include_in_legacy
             assert block.modified == datetime(2019, 11, 29, 22, 22, 46, 785000)
+        assert KintoImport.objects.count() == 6
+        assert KintoImport.objects.filter(
+            outcome=KintoImport.OUTCOME_NOMATCH).count() == 4
+        assert KintoImport.objects.filter(
+            outcome=KintoImport.OUTCOME_NOTFIREFOX).count() == 1
```
bobsilverberg: If there were a way to configure the `blocklist.json` in the test, as mentioned above, then you wouldn't need to assert this in each test, where the tests are not about testing this at all.

src/olympia/blocklist/tests/test_commands.py
```diff
@@ -94,6 +111,18 @@ def test_single_guid(self):
         assert blocks[1].kinto_id == blocklist_json['data'][2]['id']
         assert blocks[1].include_in_legacy
         assert blocks[1].modified == datetime(2019, 11, 22, 16, 49, 58, 416000)
+        assert KintoImport.objects.count() == 6
+        assert KintoImport.objects.filter(
+            outcome=KintoImport.OUTCOME_NOMATCH).count() == 3
+        assert KintoImport.objects.filter(
```
bobsilverberg: As mentioned above, this test isn't really about testing this, but rather about testing the single guid functionality. If this data could configured/altered by the test we could remove this assertion.

eviljeff:
note to myself: Bump the migration number!

src/olympia/blocklist/management/commands/import_blocklist.py
```diff
@@ -23,11 +23,11 @@ def handle(self, *args, **options):
 
         # filter out the blocks we've already imported
         kinto_ids = [record['id'] for record in response.json()['data']]
-        existing_blocks = Block.objects.filter(
-            kinto_id__in=kinto_ids).values_list('kinto_id', flat=True)
+        already_imported = list(KintoImport.objects.filter(
```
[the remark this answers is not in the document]
eviljeff: No idea, tbh, but I got `TypeError: argument of type 'BaseQuerySet' is not iterable` without it during local testing.

src/olympia/blocklist/models.py
eviljeff: Unfortunately not - in a safe way anyway.  Django migrations are automatically generated and aren't really supposed to reference much outside the migration (apart from django Models and Fields) so they can be trusted to reliably work in the future when the constants might have changed.

eviljeff: Also, when you update the constant this is how django know a new migration is needed to bring the database up to date.

src/olympia/blocklist/tasks.py
eviljeff: it should!

src/olympia/blocklist/management/commands/import_blocklist.py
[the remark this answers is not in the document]
bobsilverberg: Seems odd. Is the something different about what's returned from `Block.objects.filter` as opposed to from `KintoImport.objects.filter`, and if so, what, and why?

src/olympia/blocklist/models.py
bobsilverberg: Ah, I forgot that migrations are just auto-generated files. Nevermind!

src/olympia/blocklist/management/commands/import_blocklist.py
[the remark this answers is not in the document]
eviljeff: looks like it's working now without the `list()` wrapper 🤷‍♂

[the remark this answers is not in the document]
eviljeff: https://github.com/mozilla/addons-server/pull/13274/commits/05b0c2536b4bf1b1f310a482ad84345acf39477d

src/olympia/blocklist/tasks.py
eviljeff: https://github.com/mozilla/addons-server/pull/13274/commits/05b0c2536b4bf1b1f310a482ad84345acf39477d

src/olympia/blocklist/models.py
eviljeff: https://github.com/mozilla/addons-server/pull/13274/commits/05b0c2536b4bf1b1f310a482ad84345acf39477d

src/olympia/blocklist/tasks.py
eviljeff: https://github.com/mozilla/addons-server/pull/13274/commits/05b0c2536b4bf1b1f310a482ad84345acf39477d

src/olympia/blocklist/tests/test_commands.py
eviljeff: https://github.com/mozilla/addons-server/pull/13274/commits/05b0c2536b4bf1b1f310a482ad84345acf39477d

src/olympia/blocklist/tests/test_commands.py
eviljeff: The blocklist.json file is a cut down (with guids mangled for sanitation) version of the actual blocklist so I was trying to keep it as authentic as possible.  I removed a few of the unnecessary assertions though - it was redundant to check in _every_ test that there was only one `OUTCOME_NOTFIREFOX`

PR #13277, opened 2020-01-16
Update pygit2 to 1.0.2

### LINKED ISSUE #13242: Update pygit2 to 1.0.2
There has been a few new releases looking at https://github.com/libgit2/pygit2/blob/master/CHANGELOG.rst - I don't quite know why that hasn't been handled by pyup.

At least the 1.0.0 release may need some rework of our code according to changelog but I haven't spend much time on that yet.

EnTeQuAk:
* Fix `Object.type` assertions

Fixes #13242


EnTeQuAk:
> I was surprised we weren't affected by other breaking changes, but the changes do make sense and everything I've tried seem to work, test pass, so r+.

yeah I expected much more breakage but many of the other changes are actually just implementing new interfaces and we're not reading/writing anything to the index which avoids most of the changes

Issue #13292, opened 2020-01-17
Pretty print the `definition` of a rule when viewed in readonly mode


willdurand:
When a user can access the scanner rules in readonly mode only, it would be nice to pretty-print the definition of the rules. Currently, it is displayed as text but it is hard to read.

### Example

Current view:

![Screen Shot 2020-01-17 at 17 48 37](https://user-images.githubusercontent.com/217628/72629951-9c8f6980-3951-11ea-9a4b-9714c1af0a38.png)

Expected view:

![Screen Shot 2020-01-17 at 17 48 00](https://user-images.githubusercontent.com/217628/72629945-9600f200-3951-11ea-8a98-fe0ae9048dd4.png)

AlexandraMoga:
Verified fixed on -dev. 
Rule readability has been improved with the nee format:

 
![image](https://user-images.githubusercontent.com/31961530/72737809-fc338200-3ba8-11ea-99b2-4cfb9a58adcb.png)

PR #13293, opened 2020-01-17
Format readonly definition

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13292

---

This is a minor improvement. The issue linked above contains screenshots.

### DIFF
--- a/src/olympia/scanners/admin.py
+++ b/src/olympia/scanners/admin.py
@@ -390,6 +390,15 @@ class AbstractScannerRuleAdminMixin(admin.ModelAdmin):
     )
     readonly_fields = ('created', 'modified', 'matched_results_link')
 
+    def get_fields(self, request, obj=None):
+        fields = super().get_fields(request, obj)
+        if not self.has_change_permission(request, obj):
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/admin.py
```diff
@@ -390,6 +390,15 @@ class AbstractScannerRuleAdminMixin(admin.ModelAdmin):
...
     readonly_fields = ('created', 'modified', 'matched_results_link')
 
+    def get_fields(self, request, obj=None):
+        fields = super().get_fields(request, obj)
+        if not self.has_change_permission(request, obj):
+            # Remove the 'definition' field...
+            fields = list(filter(lambda x: x != 'definition', fields))
+            # ...and add its readonly (and pretty!) alter-ego.
+            fields.append('formatted_definition')
+        return fields
```
eviljeff: this isn't the _wrong_ way of doing it, though isn't just building the list directly in this function (i.e. not defining `fields =` at the top of the class) more straightforward than a list&filter with a lambda?

Issue #13263, opened 2020-01-15
Add a new Scanners permission that would allows scanner results management


AlexandraMoga:
Follow up for #13239 

### Describe the problem and steps to reproduce it:
As discussed in https://github.com/mozilla/addons-server/issues/13239#issuecomment-574192238, none of the available `Admin:Scanners*` permissions will give access to the Actions field for reporting True or False matched rules. Currently, a user must also have `Admin:Advanced` permissions in order to access those Actions.

### What did you expect to happen?
We should have a separate rule - `Admin:ScannersResultsEdit` - which allow access to the Actions fields

### Anything else we should know?
`Edit` can be replaced with other term, like `Manage`, since scanner results are not actually editable

AlexandraMoga:
Verified fixed on -dev.
Users with `Admin:ScannersResultsEdit` can now use the Action fields in Scanner results.

PR #13294, opened 2020-01-17
Add Admin:ScannersResultsEdit permision

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13263

---

`Admin:Advanced` was used before but it seems a bit better to have a dedicated permissions.


src/olympia/constants/permissions.py
```diff
@@ -82,6 +82,8 @@
 
 # Can access the scanners results admin.
 ADMIN_SCANNERS_RESULTS_VIEW = AclPermission('Admin', 'ScannersResultsView')
+# Can use "actions" on the scanners results.
```
willdurand: and maybe more in the future..

PR #13300, opened 2020-01-20
enable dual sign-off only for ADU>100k; refactor BlockSubmission fields

### LINKED ISSUE #13285: replace the Block dual sign-off waffle with a condition around addon ADU
from https://github.com/mozilla/addons-server/issues/13089#issuecomment-572233744

> One more thing we discussed with the Security people: there could be a threshold for requiring dual sign-off. If all add-ons involved in a block have user counts under that threshold, the block change would be allowed without sign-off. We need to figure out what the right number is, but we talked about 100K being a good starting point.
>
> Since we don't have stats for unlisted extensions at the moment, any ID without stats should required dual sign-off. Once we have complete data this threshold should become more practical.

eviljeff:
fixes #13285


src/olympia/blocklist/models.py
```diff
@@ -220,14 +224,20 @@ def is_save_to_blocks_permitted(self):
         )
 
     def save(self, *args, **kwargs):
-        if self.input_guids and not self.processed_guids:
+        def serialize_block(block):
+            return {
+                'id': block.id,
+                'guid': block.guid,
+                'adu': block.addon.average_daily_users,
```
willdurand: Can we rename `adu` to `average_daily_users` everywhere? It is super confusing, especially given that Mozilla uses acronyms like DAU and ADAU (for daily active users).

src/olympia/blocklist/admin.py
```diff
@@ -317,12 +321,13 @@ def save_model(self, request, obj, form, change):
...
                 obj.signoff_state = BlockSubmission.SIGNOFF_REJECTED
-        else:
-            # TODO: something more fine-grained that looks at user counts
-            if waffle.switch_is_active('blocklist_admin_dualsignoff_disabled'):
-                obj.signoff_state = BlockSubmission.SIGNOFF_NOTNEEDED
 
         super().save_model(request, obj, form, change)
+
+        if obj.signoff_state == BlockSubmission.SIGNOFF_PENDING:
+            if not obj.check_needs_signoff():
```
willdurand: This would read better without the `check_` prefix.

eviljeff:
> I also think most of the logic in the admin should be moved in the model

Can you point to some examples?

src/olympia/blocklist/models.py
eviljeff: https://github.com/mozilla/addons-server/pull/13300/commits/c5d5002fe1ae86db2c0dbbf84bf2d2c0f919b75d

src/olympia/blocklist/admin.py
eviljeff: https://github.com/mozilla/addons-server/pull/13300/commits/d6dc7663d8f4bd45bc13d722067df570e84b0892

willdurand:
> Can you point to some examples?

Sorry, yes (although I did not mean to change anything in this patch). I see a bunch of conditions to change the state of a block and that would likely be more self-contained in the model. I was thinking about introducing some model methods like: `signed_of_by(user)`, etc.

PR #13312, opened 2020-01-21
Correctly add all parent folders for deleted files.

### LINKED ISSUE #13092: Grandparent folder of deleted file is missing in file.entries returned from Compare API
Url: https://addons-dev.allizom.org/api/v5/reviewers/addon/7560/versions/1252898/compare_to/1688305/?lang=en-US
Method: GET
### What happened?
Response.file.entries lists entries in the folder called `chrome`, but does not include `chrome` folder itself.

image

file.entries should include entry `chrome`. The file tree is expected to complete.

The folder `chrome` are entirely deleted in the new version in this case.

It looks like that we added the parent folder of the deleted file to mimic the original data structure, but did not add the grandparent path. 
https://github.com/mozilla/addons-server/blob/b2c6b7eaac6815daee682e9efbe8101ff40eca01/src/olympia/reviewers/serializers.py#L328

The web app code-manager throws an error when it find that the parent folder of any entries does not exist (https://github.com/mozilla/addons-code-manager/issues/1327).

EnTeQuAk:
This will make sure all parent folders are created properly in the
exposed file-tree for deleted files.

Fixes #13092


src/olympia/reviewers/serializers.py
```diff
@@ -348,25 +349,29 @@ def get_entries(self, obj):
...
-                    'depth': path_depth - 1,
-                    'filename': os.path.basename(parent_path),
-                    'sha256': None,
-                    'mime_category': 'directory',
-                    'mimetype': 'application/octet-stream',
-                    'path': parent_path,
-                    'size': None,
-                    'modified': None,
-                }
+            parents = [x for x in pathlib.Path(path).parents if str(x) != '.']
```
willdurand: shouldn't `x` be renamed to something else more explicit?

src/olympia/reviewers/tests/test_serializers.py
```diff
@@ -471,6 +471,33 @@ def test_do_not_recreate_parent_dir_if_it_exists(self):
         # modified timestamp.
         assert parent['modified'] is not None
 
+    def test_expose_grantparent_dir_deleted_subfolders(self):
```
willdurand:
```suggestion
    def test_expose_grandparent_dir_deleted_subfolders(self):
```

willdurand: (I guess)

src/olympia/reviewers/tests/test_serializers.py
```diff
@@ -471,6 +471,33 @@ def test_do_not_recreate_parent_dir_if_it_exists(self):
...
+            'chrome/icons/de/foo.png',
+            delete=True)
+
+        data = self.serialize(
+            new_version.current_file, parent_version=parent_version)
+
+        entries_by_file = {
+            e['path']: e for e in data['entries'].values()
+        }
+        # Check that we correctly include grant-parent folders too
```
willdurand:
```suggestion
        # Check that we correctly include grandparent folders too
```

willdurand: I would also rename the variable :)

src/olympia/reviewers/tests/test_serializers.py
EnTeQuAk: :man_facepalming:

src/olympia/reviewers/serializers.py
EnTeQuAk: Indeed, I think e25bce37f6 will actually fix this in a better way. Let's see if I broke something accidentally though.

PR #13314, opened 2020-01-21
Add separate signoff permission

### LINKED ISSUE #13284: separate permissions for Block signoff & creation
from https://github.com/mozilla/addons-server/issues/13089#issuecomment-573023301

> Another requirement that just came up yesterday is that the review workflow should have a separate permission, as we might have users who are only reviewing blocks.

eviljeff:
fixes #13284


eviljeff:
next time they'll be "split into different" permissions

PR #13320, opened 2020-01-22
Trigger content review on metadata change only for listed

### LINKED ISSUE #13319: Limit content-review trigger on metadata changes to add-ons with listed versions

Follow-up for #13044.

Only add-ons with listed versions should re-enter the content-review queue after metadata changes. For a reviewer, this is not a problem since the content-review queue filters on add-ons with listed versions, but the `statsd` signal does not and we use that for statistics in grafana.

### What happened?

`devhub.metadata_content_review_triggered` triggers for changes to unlisted-only add-ons.

`devhub.metadata_content_review_triggered` only triggers for changes in add-ons with listed versions.

wagnerand:
Fix #13319.

### DIFF
--- a/src/olympia/devhub/forms.py
+++ b/src/olympia/devhub/forms.py
@@ -171,7 +171,7 @@ class AddonFormBase(TranslationFormMixin, forms.ModelForm):
             fetch_existing_translations_from_addon(
                 self.instance, self.fields_to_trigger_content_review)
             if self.instance and
-            self.instance.addon.has_listed_versions() and
+            self.instance.has_listed_versions() and
             metadata_content_review else {})
         obj = super().save(*args, **kwargs)
         if not metadata_content_review:
--- a/src/olympia/devhub/forms.py
+++ b/src/olympia/devhub/forms.py
@@ -174,15 +174,16 @@ class AddonFormBase(TranslationFormMixin, forms.ModelForm):
                 self.instance, self.fields_to_trigger_content_review)
             if metadata_content_review else {})
         obj = super().save(*args, **kwargs)
-        if metadata_content_review:
-            new_data = (
-                fetch_existing_translations_from_addon(
-                    obj, self.fields_to_trigger_content_review)
-            )
-            if existing_data != new_data:
-                # flag for content review
-                statsd.incr('devhub.metadata_content_review_triggered')
-                AddonApprovalsCounter.reset_content_for_addon(addon=obj)
+        if not metadata_content_review:
+            return obj
+        new_data = (
+            fetch_existing_translations_from_addon(
+                obj, self.fields_to_trigger_content_review)
+        )
+        if existing_data != new_data:
+            # flag for content review
+            statsd.incr('devhub.metadata_content_review_triggered')
+            AddonApprovalsCounter.reset_content_for_addon(addon=obj)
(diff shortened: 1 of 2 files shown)


src/olympia/devhub/forms.py
```diff
@@ -178,7 +178,7 @@ def save(self, *args, **kwargs):
             fetch_existing_translations_from_addon(
                 obj, self.fields_to_trigger_content_review)
         )
-        if existing_data != new_data:
+        if obj.has_listed_versions() and existing_data != new_data:
```
eviljeff: lets move this up to line 173 so we're not even bothering fetching translations for unlisted only addons.

src/olympia/devhub/tests/test_views_edit.py
```diff
@@ -321,17 +321,24 @@ def test_metadata_change_triggers_content_review(self):
...
 
         # make the edit
         response = self.client.post(self.describe_edit_url, data)
         assert response.status_code == 200
         addon = self.addon = self.get_addon()
 
-        # last_content_review should have been reset and the metadata updated
-        assert not AddonApprovalsCounter.objects.get(
-            addon=addon).last_content_review
+        if self.listed:
```
eviljeff: Does this test definitely run with an addon with only unlisted versions at some point?

wagnerand: Yes, via the child class `TestEditDescribeUnlisted`. Without that if/else block, the test fails.

src/olympia/devhub/forms.py
```diff
@@ -170,19 +170,22 @@ def save(self, *args, **kwargs):
...
-        new_data = (
-            fetch_existing_translations_from_addon(
-                obj, self.fields_to_trigger_content_review)
-        )
-        if existing_data != new_data:
-            # flag for content review
-            statsd.incr('devhub.metadata_content_review_triggered')
-            AddonApprovalsCounter.reset_content_for_addon(addon=obj)
-        return obj
+        if obj.has_listed_versions():
```
wagnerand: @eviljeff This method looks a bit messy now, should I refactor it to clearly separate between waffle enabled/disabled and listed/unlisted?

eviljeff: Yeah, I'd change it to something like
```
metadata_content_review = waffle.switch_is_active('metadata-content-review') and self.instance and self.instance.addon.has_listed_versions()
existing_data = (
            fetch_existing_translations_from_addon(
                self.instance, self.fields_to_trigger_content_review)
            if metadata_content_review else {})
```

src/olympia/devhub/forms.py
```diff
@@ -165,24 +165,27 @@ def get_tags(self, addon):
...
+        else:
+            new_data = (
+                fetch_existing_translations_from_addon(
+                    obj, self.fields_to_trigger_content_review)
+            )
+            if existing_data != new_data:
+                # flag for content review
+                statsd.incr('devhub.metadata_content_review_triggered')
+                AddonApprovalsCounter.reset_content_for_addon(addon=obj)
+            return obj
```
eviljeff: you don't technically need any of this code churn now - `metadata_content_review` would be False so execution would have returned already.

PR #13326, opened 2020-01-22
Iterate through deltas instead of patches to generate entries.

### LINKED ISSUE #13291: Investigate pygit2 performance by iterating differently through the diffs
This is just temporary for now till I played more with it but I found https://github.com/libgit2/pygit2/issues/736 and it seems that there might be some more performant API usage that we could use to speed up the diff api.

I'll investigate and update the issue accordingly.

EnTeQuAk:
Fixes #13291

This change will use a different pygit2 API to iterate through the
metadata of a diff without accessing it's textual representation.

That'll speed up the generation of `entries` for the compare views.

There might still be some significant time spent generating the diffs
and `entries` as we do fetch some data to figure out correct mimetype
but we should see at least some speedup (according to pygit2 issue
tracker it's significant, unsure how it compares to other places we
spend lots of time on)

### DIFF
--- a/src/olympia/lib/git.py
+++ b/src/olympia/lib/git.py
@@ -564,6 +564,41 @@ class AddonGitRepository(object):
                     patch, commit, parent, pathspec))
         return changes
 
+    def get_deltas(self, commit, parent, pathspec=None):
+        """Only fetch deltas from `parent` to `commit`.
+
+        This method specifically does not render any textual changes
+        but fetches as few details as possible to use a different
+        `pygit2` API to retrieve changes and to improve performance
+        significantly.
+
+        The entries returned are fairly similar to what `get_diff`
+        returns but don't include `hunks`, `lines_deleted` / `lines_added`
+        as well as `new_ending_new_line` and `old_ending_new_line`
+
+        We also don't expose `size` and `is_binary` as it's unreliable since
+        the `deltas` iterator tries to not examine the files content if
+        possible - so they might have wrong values.
+        """
+        diff = self.get_raw_diff(
+            commit, parent=parent, include_unmodifed=pathspec is not None)
+
+        deltas = []
+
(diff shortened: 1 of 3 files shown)


bobsilverberg:
> Fixes #13291
> 
> This change will use a different pygit2 API to iterate through the
> metadata of a diff without accessing it's textual representation.
> 
> That'll speed up the generation of `entries` for the compare views.
> 
> There might still be some significant time spent generating the diffs
> and `entries` as we do fetch some data to figure out correct mimetype
> but we should see at least some speedup (according to pygit2 issue
> tracker it's significant, unsure how it compares to other places we
> spend lots of time on)

I wonder if we actually need the mimetype for the entries? ASAIU entries are just used for generating the file tree, and the file tree doesn't give an indication of what the type of a file is. Could we just not include mimetype in entries, and if we did that would it help performance?

EnTeQuAk:
> I wonder if we actually need the mimetype for the entries? ASAIU entries are just used for generating the file tree, and the file tree doesn't give an indication of what the type of a file is. Could we just not include mimetype in entries, and if we did that would it help performance?

That would indeed help performance considerably. But as discussed yesterday, afaik some frontend changes might be needed for that?

bobsilverberg:
> > I wonder if we actually need the mimetype for the entries? ASAIU entries are just used for generating the file tree, and the file tree doesn't give an indication of what the type of a file is. Could we just not include mimetype in entries, and if we did that would it help performance?
> 
> That would indeed help performance considerably. But as discussed yesterday, afaik some frontend changes might be needed for that?

That is correct, but that changes are certainly manageable. Perhaps we can coordinate this work next week?

EnTeQuAk:
> That is correct, but that changes are certainly manageable. Perhaps we can coordinate this work next week?

Sounds great.

Issue #13140, opened 2019-12-20
Add a new task to call the ML API


willdurand:
Now that we collect data during the submission process, the next step is to get some more insights about the collected data. In order to do that, we'll call a _ML API_.

The idea is to create a new task in the validation chain that will call the ML API once with the collected data. We'll store the results and continue the process.

We'll have to decide whether this task should fail gracefully or if an error should abort the submission process. FTR, the scanner tasks fail gracefully. We should probably do this too. We could store this piece of information somewhere so that we now that a given submission/version did not get "scanned" correctly.

willdurand:
> We'll have to decide whether this task should fail gracefully or if an error should abort the submission process. FTR, the scanner tasks fail gracefully. We should probably do this too.

This is not yet handled.

willdurand:
> The idea is to create a new task in the validation chain that will call the ML API once with the collected data.

I filed an issue to actually add the task to the chain: https://github.com/mozilla/addons-server/issues/13331

PR #13330, opened 2020-01-22
Add a new task to call the ML API

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13140

---

It is possible that we'll replace `ML_API_*` variables once we have a codename
for this API (but maybe not). This patch adds a new validation task to call a
ML API that is not yet up and running. This task calls an HTTP API with the
`customs` raw results and stores the JSON response in a new `ScannerResult`
object (with `scanner=ML_API`).

Because we do not have the API yet (I don't think the API has been written
yet), we cannot really test this task (unless we set up a fake API server).
I think it's fine because we'll test the integration in #13331.

If you want more info about the ML API and its contract, let me know in MP.

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1873,6 +1873,8 @@ CUSTOMS_API_URL = env('CUSTOMS_API_URL', default=None)
 CUSTOMS_API_KEY = env('CUSTOMS_API_KEY', default=None)
 WAT_API_URL = env('WAT_API_URL', default=None)
 WAT_API_KEY = env('WAT_API_KEY', default=None)
+ML_API_URL = env('ML_API_URL', default=None)
+ML_API_TIMEOUT = 10  # seconds
 # Git(Hub) repository names, e.g., `owner/repo-name`
 CUSTOMS_GIT_REPOSITORY = env('CUSTOMS_GIT_REPOSITORY', default=None)
 YARA_GIT_REPOSITORY = env('YARA_GIT_REPOSITORY', default=None)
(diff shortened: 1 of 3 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -1873,6 +1873,8 @@ def get_raven_release():
 CUSTOMS_API_KEY = env('CUSTOMS_API_KEY', default=None)
 WAT_API_URL = env('WAT_API_URL', default=None)
 WAT_API_KEY = env('WAT_API_KEY', default=None)
+ML_API_URL = env('ML_API_URL', default=None)
```
eviljeff: A longer prefix would be ideal.  A two-letter abbreviation could mean many things.

src/olympia/scanners/tasks.py
```diff
@@ -14,7 +14,17 @@
...
+    ABORTED,
+    ABORTING,
+    COMPLETED,
+    CUSTOMS,
+    ML_API,
+    NEW,
+    RUNNING,
+    SCANNERS,
+    WAT,
+    YARA
```
eviljeff: needs a trailing comma if you're going to go this style 😛

src/olympia/lib/settings_base.py
willdurand: I agree it's not the best name ever.. I left a note about that in this PR's description. I think I'll change that once we have a codename.

src/olympia/scanners/tasks.py
willdurand: can't wait to use Black.

Issue #13317, opened 2020-01-22
Create an API endpoint to expose the scanner results


willdurand:
Refs https://github.com/mozilla/addons-server/issues/13142

---

Only authenticated users with `ADMIN_SCANNERS_RESULTS_VIEW` should be able to query this new endpoint. The new endpoint will expose the scanner results with pagination (page-based?). We can start by exposing the following properties:

- `id`
- scanner name
- (raw) `results`
- ~~label (good vs bad but values TBD)~~ see #13334 

**For QA:** please ping us so that we can temporarily enable this API endpoint in -dev (it is behind a feature flag (switch) named `enable-scanner-results-api`).

willdurand:
> label (good vs bad but values TBD)

This will be done in https://github.com/mozilla/addons-server/issues/13334

AlexandraMoga:
I've verified the new [scanner results api endpoint](https://addons-server.readthedocs.io/en/latest/topics/api/scanners.html#get--api-v4-scanner-results-) on -dev and I have the following results:

- the API responds only when the `enable-scanner-results-api` is enabled
- to query this API your user needs to be either super admin or has `Admin:ScannersResultsView` permission; non-authorized users will receive a 403
- the API exposes the scanner `id`, `name` and `results` object 
- the API is paginated (25 results per page)
- the `results` object keeps a track of files scanned and their matched rules (where applicable) for each of the three scanners

@willdurand For future iterations, I think it could be useful to also expose the add-on `guid` in this API

willdurand:
@AlexandraMoga thanks! As for the `guid`, we do not want to expose it because the clients should not be able to associate scanner results with add-ons/versions (for privacy reasons).

PR #13335, opened 2020-01-23
Add a new API endpoint for ScannerResult(s)

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13317

---

This is the first iteration to provide a new API endpoint for the scanner results.

I did not update the docs (yet?) because this endpoint is restricted and will only be useful to very few people (also, it won't be enabled right away, even in -dev I guess).

The endpoint URL is: `/scanner/results/`.


src/olympia/scanners/migrations/0023_auto_20200122_1649.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 2.2.9 on 2020-01-22 16:49
+
+from django.db import migrations
+
+
+def create_waffle_switch(apps, schema_editor):
```
eviljeff: Do we really need the waffle switch if it's gated by a permission check too?  What's the scenario where we want to turn it off and just telling the handful of users with that permission not to access it isn't sufficient?

willdurand: Yes, we do because we need a confirmation from secops as well (mostly to know whether we need extra protection)

willdurand:
> I think you need to write the API docs 😛
> We document the all internal APIs, however useless to a normal user. (e.g. [addons-server.readthedocs.io/en/latest/topics/api/activity.html#incoming-mail-end-point](https://addons-server.readthedocs.io/en/latest/topics/api/activity.html#incoming-mail-end-point))

Okay!

willdurand:
> I think you need to write the API docs 😛
> We document the all internal APIs, however useless to a normal user. (e.g. [addons-server.readthedocs.io/en/latest/topics/api/activity.html#incoming-mail-end-point](https://addons-server.readthedocs.io/en/latest/topics/api/activity.html#incoming-mail-end-point))

@eviljeff Can I defer this to a different issue/patch? I really don't want to advertise this endpoint for now?

docs/topics/api/scanners.rst
```diff
@@ -0,0 +1,22 @@
...
+
+.. _scanner-results:
+
+This endpoint returns a list of labelled scanner results.
+
+.. http:get:: /api/v4/scanner/results/
+
+    :>json int id: The scanner result ID.
+    :>json string scanner: The scanner name.
+    :>json object results: The scanner (raw) results.
```
eviljeff: It would be good if we could document a breakdown of the object structure, though I guess that's a) not determined currently and b) could change from scanner to scanner (and c) maybe something we can't expose publicly)

src/olympia/api/urls.py
```diff
@@ -29,6 +29,7 @@
     url(r'^', include('olympia.signing.urls')),
     url(r'^', include(amo_api_patterns)),
     url(r'^hero/', include('olympia.hero.urls')),
+    url(r'^scanner/', include('olympia.scanners.api_urls')),
```
eviljeff: Do we want this available in /v4/ or just /v5/ ? The test specifies v5 but it's been added to v4 here.  (it's not difficult to change - just define a `v5_api_urls = v4_api_urls + [url(r'^scanner...]`)

docs/topics/api/scanners.rst
```diff
@@ -0,0 +1,22 @@
+============
+Scanners
+============
```
eviljeff: The change should be added to https://github.com/mozilla/addons-server/blob/master/docs/topics/api/overview.rst#v4-api-changelog too

docs/topics/api/scanners.rst
willdurand: it is definitely (b) and (c) :wink:

src/olympia/api/urls.py
willdurand: could be in v4+

docs/topics/api/scanners.rst
willdurand: done

docs/topics/api/scanners.rst
EnTeQuAk: I wonder if we have some potential structures documented in the scanners repositories and link there with some "this link may require specific permissions" notice or so? That way we'd document the structures of each scanner individually (doesn't have to be in-depth documentation just now) and link to specific scanners individually.

willdurand: We discussed about this and decided not to do it :)

willdurand: FWIW, each scanner and/or repo has its own docs and things are documented but we won't link to the different scanners/repos.

PR #13336, opened 2020-01-23
Fix state changes for ScannerQueryRule, add logging in tasks and change celery queue

### LINKED ISSUE #13267: Route code search tasks to a separate celery queue
https://github.com/mozilla/addons-server/issues/13256 will route the code search tasks to the `devhub` queue, but we should consider changing that so that firing up a code search doesn't affect submissions.

Need to figure out with ops what's our current queues configuration and if adding a new one makes sense or not.

diox:
Fixes #13267


src/olympia/scanners/models.py
```diff
@@ -295,6 +304,30 @@ class ScannerQueryRule(AbstractScannerRule):
     class Meta(AbstractScannerRule.Meta):
         db_table = 'scanners_query_rules'
 
+    def change_state_to(self, target):
```
diox: Fairly basic but makes it more clear what states are and what they can be changed to.

willdurand: I like it.

src/olympia/scanners/tasks.py
```diff
@@ -272,12 +278,15 @@ def run_yara_query_rule(query_rule_pk):
...
     # time, chained to a task that marks the query as completed.
     chunk_size = 250
+    chunked_tasks = create_chunked_tasks_signatures(
+        run_yara_query_rule_on_versions_chunk, list(pks), chunk_size,
+        task_args=(query_rule_pk,))
     workflow = (
-        create_chunked_tasks_signatures(
-            run_yara_query_rule_on_versions_chunk, list(pks), chunk_size,
-            task_args=(query_rule_pk,)) |
+        chunked_tasks |
```
willdurand: I find this notation quite confusing actually (the `|`) but that was there before so 🤷‍♂

src/olympia/scanners/tasks.py
```diff
@@ -235,29 +235,35 @@ def mark_yara_query_rule_as_completed_or_aborted(query_rule_pk):
...
+        elif rule.state == ABORTING:
+            rule.change_state_to(ABORTED)
+    except ImproperScannerQueryRuleStateError:
+        log.error('Not marking rule as completed or aborted for rule %s in '
+                  'mark_yara_query_rule_as_completed_or_aborted, its state is '
+                  '%s' % (rule.pk, rule.get_state_display()))
 
 
 @task
 def run_yara_query_rule(query_rule_pk):
```
willdurand: We could also add some `statsd` calls to monitor things (not in this patch obviously).

diox: I debated adding `statsd` calls, but because code search will be used infrequently and with very different rules, I wasn't sure there would be enough data to properly visualize stuff in Grafana.

Issue #13331, opened 2020-01-22
Add `call_ml_api` task to the validation chain


willdurand:
In https://github.com/mozilla/addons-server/issues/13140, we added a new task to call the ML API for the sectools. This issue is about using this task in the validation chain so that we can call this API (once it is possible).

We'll likely want to use a waffle switch to control this task (similar to what we have for the scanners).

AlexandraMoga:
@willdurand with the switch enabled, I'm currently looking for regressions in add-on submissions and scanner results linked to them. I believe this is what I'm able to test at this moment, given the fact that we are not exposing any actual `ml_api` scanner results in admin.
If you have some more specific instructions, I'll be happy to test them too.

willdurand:
Indeed, that's the only thing we have to test right now. Once we hook up the ML API, I'll file a new issue with instructions to test the integration.

AlexandraMoga:
I've submitted several add-ons, both from dev hub and the signing api (including load tests) and verified the scanner results tied to these submissions. I didn't notice anything out of the usual. 

I'm marking this issue as verified fixed.

PR #13073, opened 2019-12-09
Don't deal with multiple data types at the same time in reindex command

diox:
This simplification allows us to fix an issue with the reindex workflow being wrong with Celery > 4.1.1, which completely broke reindexing of a single index.

A new `--data` parameter is introduced for when we want to reindex stats and not addons. We'll have to fire the command twice if we want to reindex both.

Fixes #13049

(I'm hoping it would also fix #13051 but I'm not sure yet)

Issue #13049, opened 2019-12-05
elasticsearch `reindex` django command is broken


eviljeff:
`./manage.py reindex` is supposed to trigger a full re-index of the elasticsearch indexes but it exits too early.
https://sentry.prod.mozaws.net/operations/olympia-dev/issues/6733763/ is an exception thrown "NotFoundError: TransportError(404, 'index_not_found_exception', 'no such index')".

diox:
`reindex` works like this:
- Flag the db
- *then*, create a new index
- *then*, reindex all content (that will happen on both the new index and the current one)
- *then*, update aliases so that the alias points to the new index instead of the old one
- *then*, unflag the database
- *then*, delete the old index

The command is supposed to create a celery workflow with all these tasks chained, fire it up, wait with a crude `while` loop for the database to be unflagged, while celery processes all this.

Crucially, that exception we're seeing is coming from `update_aliases`, so it should *not* be happening multiple times... but it does. It also seems to indicate that we've deleted the old index way too early.

The fact that the command returns almost straight away is probably a byproduct of the tasks failing very quickly.

diox:
I'm also seeing this locally. I suspect it's been broken or a while, probably since the last celery upgrade ?

Hard to read, but the workflow looks like this:
`%(olympia.lib.es.management.commands.reindex.update_aliases([flag_database('addons-20191205162444', 'addons', 'addons') | create_new_index('addons', 'addons-20191205162444') | group([olympia.addons.tasks.index_addons([...]), index_addons([...]), index_addons([...]), ...])], [{'remove': {'index': 'addons-20191205161659', 'alias': 'addons'}}, {'add': {'index': 'addons-20191205162444', 'alias': 'addons'}}]) | unflag_database() | delete_indexes(['addons-20191205161659']))`

(I've had to manually trim it since I've got a lot of add-ons locally. Maybe I can get a pretty drawing output to see it more clearly)

diox:
Downgrading to Celery 4.2.x **does not** fix the issue.
Downgrading to Celery 4.1.1 **does** fix the issue.

I suggest the following patch to test locally:
```diff
diff --git a/src/olympia/addons/indexers.py b/src/olympia/addons/indexers.py
 index 612a4d0305..a637ae665e 100644
 --- a/src/olympia/addons/indexers.py
 +++ b/src/olympia/addons/indexers.py
 @@ -496,6 +496,6 @@ def reindex_tasks_group(index_name):
[...]

PR #13337, opened 2020-01-23
Add `call_ml_api` task to the validation chain

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13331

---

Because this task must be the [callback of the chord](https://docs.celeryproject.org/en/latest/userguide/canvas.html#chords), we have to add it to the chain all the time and therefore the switch has to live in the task itself. I think it is fine, though. I did a few tests and it should not impact the submission process.


willdurand:
(also /cc @diox)

diox:
So with that change we'd be chaining the result of a chord with another task. Need to double-check how safe it is, given the recent regressions we've seen with celery and complex workflows.

willdurand:
> So with that change we'd be chaining the result of a chord with another task. Need to double-check how safe it is, given the recent regressions we've seen with celery and complex workflows.

Yes, I was surprised that it worked so well locally.

willdurand:
I rebased this patch to avoid a migration conflict.

diox:
When I fixed [the reindexation code to work with Celery > 4.1.1](https://github.com/mozilla/addons-server/issues/13049), one of the things I did was separate in a specific method the code that returns the tasks workflow to run, and add [a test that proves the workflow looks good](https://github.com/mozilla/addons-server/pull/13073/files#diff-ad36a728cb976284cc85b48045843640R177).

Since the workflow is getting quite complex here, I think it'd be good to also have that in this case.

willdurand:
> When I fixed [the reindexation code to work with Celery > 4.1.1](https://github.com/mozilla/addons-server/issues/13049), one of the things I did was separate in a specific method the code that returns the tasks workflow to run, and add [a test that proves the workflow looks good](https://github.com/mozilla/addons-server/pull/13073/files#diff-ad36a728cb976284cc85b48045843640R177).
> 
> Since the workflow is getting quite complex here, I think it'd be good to also have that in this case. Something that would break loudly if there is a bad regression in celery, something we can't necessarily see in other unit tests that are ran in eager mode.

I like the idea but I am not sure to understand how the test case works.

We have test cases where we only mock the main `chain` so that we can have assertions on all the tasks in it, making sure we configure the chains/chords correctly.

I agree that an internal change in Celery would likely not break these tests, though but it is not clear to me how the linked test case covers that. Is it thanks to `assert isinstance(workflow, _chain)`?

diox:
Mocking `chain` tells you what you give to celery, but that doesn't tell you how celery actually translates it later. Celery automatically "upgrades" a group chained to a task to a chord for instance, and also modifies the workflow in other circumstances, trying to fit what you gave as input into something workers can then use.

I don't remember exactly what in that test was revealing the breakage, but I think you should inspect everything: the returned object, the tasks in it, etc, much like that test does.

willdurand:
Right, let me see how I can convert that into a test case.

PR #13342, opened 2020-01-23
use BlockSubmission add_view for single guids too

### LINKED ISSUE #13202: implement dual sign off for adding single guid/block
Currently single add-on add block are created immediately.  For dual-sign (#13089) off to work we need to defer the creation of the block to after sign-off.  Most likely we can do this by reusing the Multi-block action (hopefully).

eviljeff:
fixes #13202 by redirecting a single guid into `BlockSubmission.add_view` used for multiple guids.  Some special casing has been added to make it similar in features.  Because there wasn't a `Block.add_view` any more I could delete some of the code supporting non-change in `BlockAdmin` and simplify how the input view hooked in too (and delete some tests that were obsolete because it was now the same view)

I also took the opportunity to make `BlockSubmission.add_view` and `.change_view` share more functionality (field definitions, base model form, stylesheets).  End goal would be to use the same view like django admin does by default.


src/olympia/blocklist/admin.py
```diff
@@ -25,46 +24,42 @@
...
+def _get_version_choices(obj, field):
+    default = obj._meta.get_field(field).default
+    choices = [
+        (version, version) for version in (
+            [default] + list(obj.addon_versions.keys())
+        )
+    ]
+    value = getattr(obj, field)
+    if value and (value, value) not in choices:
+        choices = [(getattr(obj, field), '(invalid)')] + choices
```
EnTeQuAk: I'm not sure I fully understood what this is doing. How can the `(vaue, value) not in choices` ever be `True` here? Where is `(invalid)` coming from again, I couldn't find it.

eviljeff: ... yeah, it's become a little detached from the original implementation that was on the back of a pull request discussion.

The call would be something like `_get_version_choices(block, 'min_version')` so `value` would be a version string.  `(value, value)` wouldn't be in `choices` if the version string (`Block.min_version`) wasn't one of the versions of the addon the block was referencing.  This could either be because the block was created as part of a multiple block submission and the admin creating it chose the broadest version range that would block everything, or because it was a new block and the `min_version` was passed as a `GET` param.
(it's also used for `max_version` too)

eviljeff: https://github.com/mozilla/addons-server/pull/13342/commits/8bfabf161520ec4f1aedabcec1a3cd4e3eee0baf

Issue #13334, opened 2020-01-23
Expose `label` in the scanner results API


willdurand:
We added a new API endpoint to expose the scanner results in https://github.com/mozilla/addons-server/issues/13317. We need to expose a `label` in addition to the existing data per scanner result. In order to expose this `label`, we need to fix https://github.com/mozilla/addons-server/issues/13142 first.

**Note:** we'll probably only expose the scanner results with a `label` that is not `null` (or `None` or `Unknown`) because we're only interested in labelled results.

**For QA:** there should be a new `label` attribute in the JSON response (for each result entry) and we should only return labelled scanner results.

ioanarusiczki:
@willdurand I verified with Postman on dev the scanners result[ API](https://addons-server.readthedocs.io/en/latest/topics/api/scanners.html). In the response JSON object
- results include a "label": good - for add-ons without matched rules 
- results include a "label": bad  - add-ons reported as true positive  
- there are results which also have an extra "label":null in `results` below the `filepath` 

![the label](https://user-images.githubusercontent.com/33448286/73546538-13def600-4446-11ea-9785-809b8f19db86.png)

Let me know if it's alright. 
Thanks!

willdurand:
please don't expose the API results in public issues :)

ioanarusiczki:
@willdurand I thought I forgot to add it. You were quick!
I'll disable the switch from admin too because I finished testing.
Noted. If it's about scanners I'll be careful in the future.

willdurand:
> @willdurand I thought I forgot to add it. You were quick!

Fortunately :)

> Noted. If it's about scanners I'll be careful in the future.

Thanks.

As for your question, the second `label` belongs to one of the scanners and we don't really care (the content of the results is tied to the scanners).

Let me know what you found but without JSON output :p

ioanarusiczki:
@willdurand Ok, sure. I can explain the details without adding screenshots too.

PR #13343, opened 2020-01-23
Expose `label` in scanner results API endpoint

willdurand:
~~⚠️ depends on #13335~~
Fixes https://github.com/mozilla/addons-server/issues/13334

### DIFF
--- a/src/olympia/scanners/views.py
+++ b/src/olympia/scanners/views.py
@@ -1,10 +1,13 @@
 import waffle
 
+from django.conf import settings
+from django.db.models import CharField, Value
 from django.db.transaction import non_atomic_requests
 from django.http import Http404
 from rest_framework.generics import ListAPIView
 
 from olympia import amo
+from olympia.constants.scanners import TRUE_POSITIVE, LABEL_BAD, LABEL_GOOD
 from olympia.api.permissions import GroupPermission
 
 from .models import ScannerResult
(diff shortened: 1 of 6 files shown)


willdurand:
Rebased!

src/olympia/scanners/views.py
```diff
@@ -16,9 +19,31 @@ class ScannerResultViewSet(ListAPIView):
         GroupPermission(amo.permissions.ADMIN_SCANNERS_RESULTS_VIEW)
     ]
 
-    queryset = ScannerResult.objects.all()
     serializer_class = ScannerResultSerializer
 
+    def get_queryset(self):
```
diox: So the issue is, this is going to potentially give you duplicate results, because we don't prevent reviewers from confirming an approval that has already been confirmed. Is that a problem ? I am not completely sure I get why we have an API that lists *all* results in the first place - eventually there is going to be a ton of them to paginate through ?

willdurand: ah, I see so we would need some sort of `DISTINCT` then?

> I am not completely sure I get why we have an API that lists all results in the first place - eventually there is going to be a ton of them to paginate through ?

Mainly to provide fresh data to the ML folks without having to manually do it. In the future, we could have a filter to only return results with `id` greater than a value or something. We could also have a hard time constraint to not return too many results in the future.

diox: You can add a `distinct()` yeah. Performance is probably going to be terrible, but we can start with that.

willdurand: mmm so the duplicate results would happen because of the `VersionLog` entries, right? I tried to write a test case with multiple entries for the same user/version but it does not return duplicate results 😕

willdurand: I think we cannot have duplicate results because the main model is `ScannerResult`, no?

diox: Actually, if you add another `VersionLog` and then inspect `get_queryset()`, `good_results` will contain a duplicate if evaluated, but the `union()` will make it disappear.
```diff
diff --git a/src/olympia/scanners/tests/test_views.py b/src/olympia/scanners/tests/test_views.py
 index 2010f13ddc..944d9149c5 100644
 --- a/src/olympia/scanners/tests/test_views.py
 +++ b/src/olympia/scanners/tests/test_views.py
 @@ -78,6 +78,14 @@ def test_get(self):
              ),
              version=good_version_2,
          )
 +        VersionLog.objects.create(
 +            activity_log=ActivityLog.create(
 +                action=amo.LOG.CONFIRM_AUTO_APPROVED,
 +                version=good_version_2,
 +                user=user_factory(),
 +            ),
 +            version=good_version_2,
 +        )
          # result NOT labelled as "good" because action is not correct.
          version_3 = version_factory(addon=addon_factory())
          ScannerResult.objects.create(scanner=YARA, version=version_3)
```

```
ipdb> good_results                                                                                                                                                                   
<BaseQuerySet [<ScannerResult: ScannerResult object (518)>, <ScannerResult: ScannerResult object (519)>, <ScannerResult: ScannerResult object (519)>]>
ipdb> bad_results                                                                                                                                                                    
<BaseQuerySet [<ScannerResult: ScannerResult object (515)>]>
ipdb> good_results.union(bad_results)                                                                                                                                                
<BaseQuerySet [<ScannerResult: ScannerResult object (518)>, <ScannerResult: ScannerResult object (519)>, <ScannerResult: ScannerResult object (515)>]>
ipdb>
```

So I think that yeah, we don't need the distinct, but adding the test would be nice to make sure it never regress if/when we changed the implementation.

willdurand: ah, thanks for helping me on this! I was a bit confused..

willdurand: I updated one of the test cases.

Issue #13360, opened 2020-01-27
Add a new `state` for scanner results


willdurand:
We need a new `state` value to describe scanner results that are technically correct but still not considered "bad" (and not "good" either). This is when a rule matches for good reasons but the use case is legit.

I am not sure what a good name would be, though. I have `inconclusive` in mind but that's generic.

AlexandraMoga:
Verified fixed on -dev.
A scanner result can now be marked as `Inconclusive` from the Actions section in the scanner results list view or from a scanner result detail view. 
The Inconclusive report can also be reverted.
Users are also able to filter scanner results that have been marked with this state.

PR #13368, opened 2020-01-28
Add inconclusive state

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13360

---

This patch adds a new `state` to triage scanner results when admins cannot use true positive or false positive.


willdurand:
@diox this might conflict with your patch

PR #13380, opened 2020-01-29
Make scanner result admin template generic

### LINKED ISSUE #13372: Fix link to matching query rule in scanner query results admin
In the admin, when looking at scanner query results, the link to matching rule is broken, it links to a Scanner Rule and not a Scanner Query Rule. The reverse works fine (Scanner Query Rule correctly links to Scanner Query Results).

diox:
This allows those templates to be re-used for scanners query results admin

Fixes #13372

### DIFF
--- a/src/olympia/scanners/admin.py
+++ b/src/olympia/scanners/admin.py
@@ -268,12 +268,15 @@ class AbstractScannerResultAdminMixin(admin.ModelAdmin):
     formatted_results.short_description = 'Results'
 
     def formatted_matched_rules(self, obj):
+        rule_model = self.model.matched_rules.rel.model
+        info = rule_model._meta.app_label, rule_model._meta.model_name
+
         return format_html(
             ', '.join(
                 [
                     '<a href="{}">{}</a>'.format(
                         reverse(
-                            'admin:scanners_scannerrule_change', args=[rule.pk]
+                            'admin:%s_%s_change' % info, args=[rule.pk]
                         ),
                         rule.name,
                     )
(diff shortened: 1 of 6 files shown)


src/olympia/scanners/tests/test_models.py
```diff
@@ -238,6 +238,15 @@ def test_get_files_by_matched_rules_for_yara(self):
             rule2: [file2],
         }
 
+    def test_get_files_by_matched_rules_no_file_somehow(self):
```
willdurand: How is this actually possible?

src/olympia/scanners/admin.py
```diff
@@ -430,8 +436,21 @@ def get_urls(self):
...
+                ),
+                'handletruepositive_urlname': (
+                    'admin:%s_%s_handletruepositive' % info
+                ),
+                'handleinconclusive_urlname': (
+                    'admin:%s_%s_handleinconclusive' % info
+                ),
+                'handlerevert_urlname': 'admin:%s_%s_handlerevert' % info,
+                'obj': obj,
+            }
```
willdurand: Do we want to triage the code search results as well?

src/olympia/scanners/tests/test_models.py
diox: It was happening in tests and with manually tweaked data. Since most of the code did handle cases where some data was missing, I thought I might as well add that fallback as well.

src/olympia/scanners/admin.py
diox: I'm not sure yet. I wanted the actions to work on a basic level, like not impact regular scanner results at the very least. Whether we end up actually needing the actions is still up in the air

src/olympia/scanners/tests/test_models.py
willdurand: ok. I guess that's fine. I don't like the `???` notation but hopefully that will never appear :p

diox: I'm open to suggestions :) Didn't want to go for a blank string, wanted *something* to be shown

PR #13397, opened 2020-02-03
Switch to recaptcha global variant

### LINKED ISSUE #13389: Use global reCAPTCHA variant
We use reCAPTCHA v2 in addons-server by calling a script (falling back to iframe) from `https://www.google.com/recaptcha/`.

We've had users complaining that it doesn't work in China, and it sounds like there is [a simple solution for this](https://developers.google.com/recaptcha/docs/faq#can-i-use-recaptcha-globally):
> (...) use "www.recaptcha.net" in your code in circumstances when "www.google.com" is not accessible.

So we should try that. According to https://www.lowendtalk.com/discussion/comment/3058885/#Comment_3058885 we should also alter our CSP to not just replace `https://www.google.com/recaptcha/` with `https://www.recaptcha.net/recaptcha/` but also include `https://www.gstatic.cn/recaptcha/`.

diox:
Fixes #13389

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1381,6 +1381,7 @@ CSP_FONT_SRC = (
 CSP_CHILD_SRC = (
     "'self'",
     'https://www.google.com/recaptcha/',
+    'https://www.recaptcha.net/recaptcha/',
 )
 CSP_FRAME_SRC = CSP_CHILD_SRC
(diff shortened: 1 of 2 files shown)


diox:
Yeah I was just planning on testing it ourselves and then asking people that complained in the past in various issues to retest. As long as it doesn't regress in our countries at least we're not in a worse position.

Issue #12725, opened 2019-10-25
Browse deleted version in reviewer API


kumar303:
### Describe the problem and steps to reproduce it:

As a reviewer, use the review tools to follow a Code Manager link to a deleted version, e.g. one deleted by the developer.

Here is an example API request made by the Code Manager page:

```
https://addons.mozilla.org/api/v5/reviewers/addon/989342/versions/4887355/
```

### What happened?

The API returns a 404 and Code Manager cannot display anything.

### What did you expect to happen?

The API should not return a 404 so that Code Manager can display the deleted version.

### Anything else we should know?

The old File Viewer displays deleted versions.

See https://github.com/mozilla/addons-code-manager/issues/1159

bobsilverberg:
@EnTeQuAk Should this be fixed now?

EnTeQuAk:
I investigated this and code-wise everything does look fine and with proper permissions (Admin permissions `*:*`) you'll see deleted versions just fine.

Unfortunately I can't access the review mentioned in the original issue in https://github.com/mozilla/addons-code-manager/issues/1159 - maybe @wagnerand can help making sure I'm not missing something here?

wagnerand:
I can view the deleted file in the code-manager as well. Closing.

AlexandraMoga:
@wagnerand you can see the page with an admin user, but can you access it with a regular reviewer? My understanding of the issue is that regular reviewers should also have view access to deleted files in code manager.
See also https://github.com/mozilla/addons-code-manager/issues/1162#issuecomment-589132568 for more details.

wagnerand:
Reopening because of https://github.com/mozilla/addons-code-manager/issues/1162#issuecomment-589132568 until we know whether this is a permission/server issue, or a code-manager issue.

bobsilverberg:
@wagnerand I'm not seeing an answer here to the question of whether a non-admin should be able to see a deleted version. Should they be able to see one? If so then it sounds like there's still an issue here. @EnTeQuAk are you able to test with a non-admin account?

wagnerand:
@bsmedberg Users that have the permission `Addons:ViewDeleted` should be able to view deleted files, and that worked for me on prod. @AlexandraMoga is currently re-testing on her side to see if she can confirm.

AlexandraMoga:
[...]

PR #13398, opened 2020-02-03
Add test to verify comparing with deleted version works.

### LINKED ISSUE #12726: View a comparison for a deleted version in the reviewer API

As a reviewer, use the review tools to craft a Code Manager comparison link between a non-deleted version and a deleted version. You would have to edit the URL to craft this URL but a UI for this will be forthcoming.

Given this crafted URL: https://code.addons.mozilla.org/en-US/compare/989342/versions/3431730...3432145/

...here is the API request made by Code Manager:

```
https://addons.mozilla.org/api/v5/reviewers/addon/989342/versions/3431730/compare_to/3432145/?lang=en-US
```

### What happened?

The API returns a 404 and Code Manager cannot display anything.

The API should not return a 404 so that Code Manager can display the comparison.

The old File Viewer would display a comparison like this.

See https://github.com/mozilla/addons-code-manager/issues/1159

EnTeQuAk:
Fixes #12726

The issue from #12726 is kinda a "worksforme" - you need the `Addons:ViewDeleted` permission to do such comparisons.

Maybe that'll make #12725 clearer too but I'll add another test to verify it's working as expected too.

pinging @bobsilverberg and @wagnerand to make sure that's the expected behavior and works for everyone.

### DIFF
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -6872,6 +6872,9 @@ class TestReviewAddonVersionCompareViewSet(
 
         user = UserProfile.objects.create(username='reviewer')
         self.grant_permission(user, 'Addons:Review')
+
+        # A reviewer needs the `Addons:ViewDeleted` permission to view and
+        # compare deleted versions
         self.grant_permission(user, 'Addons:ViewDeleted')
 
         self.client.login_api(user)
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -6800,7 +6800,7 @@ class TestReviewAddonVersionCompareViewSet(
             }
         ]
 
-    def test_get_deleted_file(self):
+    def test_compare_with_deleted_file(self):
         new_version = version_factory(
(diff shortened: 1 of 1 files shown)


EnTeQuAk:
Thanks @wagnerand, I investigated further and in that specific add-on, the versions in the original report simply don't exist anymore (they seem to be hard-deleted).

Testing this further, in this specific addon…

Screenshot_2020-02-03 New Query

And running a compare between these two versions…

https://code.addons.mozilla.org/en-US/compare/989342/versions/4894835...4894865/

works perfectly with the correct permissions.

Am I missing something?

wagnerand:
@EnTeQuAk I'm not sure I can confirm that. The file seems soft-deleted. I can still download it from the reviewer tools and the old file viewer displays the file just fine.

EnTeQuAk:
So, thanks to @wagnerand we figured out that the ids used in the reported URL are the file-ids, not the version ids.

https://www.irccloud.com/pastebin/dvLkS99V/

That's the reason why we were seeing a 404.

bobsilverberg:
Thanks for taking care of this @EnTeQuAk and @wagnerand!

PR #13403, opened 2020-02-03
Return 404 if there is no file validation

### LINKED ISSUE #13374: File validation returns 500 for versions
https://addons-dev.allizom.org/en-US/developers/addon/webmail-ad-blocker/file/237880/validation.json

### What happened?
It returns 500

No errors. (Or 404 when url is incorrect)

I found it when visiting https://code.addons-dev.allizom.org/en-US/compare/7560/versions/124625...1508403/. It looks like it depends on the versions I chose. I am not sure why some versions have this problems, while others work well.

mirefly:
Fixes #13374

### DIFF
--- a/src/olympia/devhub/fixtures/devhub/addon-validation-1.json
+++ b/src/olympia/devhub/fixtures/devhub/addon-validation-1.json
@@ -159,6 +159,22 @@
         }
     },
     {
+        "pk": 100400,
+        "model": "files.file",
+        "fields": {
+            "status": 1,
+            "is_restart_required": true,
+            "hash": "sha256:a300e3943309aee1e1ee47f569f9be36276791b3e66596be4c9cc4ee238b01b6",
+            "created": "2010-12-17 13:04:57",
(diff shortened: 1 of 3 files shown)


diox:
@mirefly do you know how the code manager copes with that 404 ? I've spent some time looking deeper into this, and the non-json equivalent of this view, `file_validation()`, handles the error gracefully without a 404, simply not showing any validation data (it uses `file_.has_been_validated` property, which does the `try`/`except` you included). So I'm wondering if maybe we should return something anyway ? Or maybe just return the 404 in json to help code manager ?

mirefly:
Code manager will show an error message as follows when the HTTP status code is not [ok](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok).

> So I'm wondering if maybe we should return something anyway ? Or maybe just return the 404 in json to help code manager ?

Do you mean to include JSON data in 404 response body? The Code manager would not try to get response body when the status code is not ok (200-299), so I guess we do not need to return JSON response body(?) @diox 

image

PR #13407, opened 2020-02-04
Don't allow modifications of a Scanner Query Rule if it's not NEW

### LINKED ISSUE #13271: Don't allow editing scanner query rules once they have been triggered once
In code search admin, once a query rule status is no longer in `New` state, we shouldn't allow admins to edit them, except to abort a `Running` query.

diox:
Fixes #13271

### DIFF
--- a/src/olympia/scanners/admin.py
+++ b/src/olympia/scanners/admin.py
@@ -543,6 +543,11 @@ class ScannerQueryRuleAdmin(AbstractScannerRuleAdminMixin, admin.ModelAdmin):
         'created', 'modified', 'matched_results_link', 'state_with_actions',
     )
 
+    def has_change_permission(self, request, obj=None):
+        if obj and obj.state != NEW:
+            return False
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/admin.py
```diff
@@ -543,6 +543,11 @@ class ScannerQueryRuleAdmin(AbstractScannerRuleAdminMixin, admin.ModelAdmin):
         'created', 'modified', 'matched_results_link', 'state_with_actions',
     )
 
+    def has_change_permission(self, request, obj=None):
+        if obj and obj.state != NEW:
```
eviljeff: I know it's not introduced in this patch, but imo these state constants are too generically named and should be either be prefixed e.g. `SCANNER_STATE_NEW` or at least namespaced. e.g. `from olympia.constants import scanners`

diox: filed https://github.com/mozilla/addons-server/issues/13411

PR #13417, opened 2020-02-05
add export_blocklist command that generates a filtercascade blob

### LINKED ISSUE #13408: Test DB generation and diffing
In order to support a CRLite DB for an add-ons blocklist we need to test generating it from AMO's block data.

* [x] Test [filter-cascade](https://github.com/mozilla/filter-cascade) for blocklist db generation
* [x] Test creating diffs

## Questions to answer

- What size is the DB output?
- What sSize are the diffs?
- How quickly is the main db generated?
- How quickly are diffs generated?

eviljeff:
fixes #13408 - will probably need refactoring to be useful.


src/olympia/blocklist/management/commands/export_blocklist.py
```diff
@@ -49,50 +50,65 @@ def add_arguments(self, parser):
...
     def get_all_guids(self):
-        return Version.unfiltered.values_list('addon__guid', 'version')
+        return File.objects.values_list(
+            'version__addon__guid', 'version__version', 'cert_serial_num')
 
     def load_json(self, json_path):
+        def tuplify(record):
+            return tuple(
+                record if len(record) == 3 else
+                record + [str(random.randint(100000, 999999))])
```
diox: I wouldn't do this. We can regenerate the dummy data we already have to include fake cert ids separately if needed, I don't think it should live in this command.

eviljeff: fair enough - it was easier when I was developing it.

eviljeff: https://github.com/mozilla/addons-server/pull/13417/commits/7b839ac67fb487e0b95aa040ae52374dc4099b16

PR #13418, opened 2020-02-05
Fix scanners celery tasks routing and remove a couple obsolete routes

### LINKED ISSUE #13267: Route code search tasks to a separate celery queue

diox:
Fixes #13267

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1080,7 +1080,6 @@ CELERY_IMPORTS = (
 CELERY_TASK_QUEUES = (
     Queue('addons', routing_key='addons'),
     Queue('amo', routing_key='amo'),
-    Queue('api', routing_key='api'),
     Queue('bandwagon', routing_key='bandwagon'),
     Queue('cron', routing_key='cron'),
     Queue('crypto', routing_key='crypto'),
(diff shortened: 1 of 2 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -1060,10 +1060,21 @@ def get_db_config(environ_var, atomic_requests=True):
...
+    'olympia.files.tasks',
     'olympia.lib.crypto.tasks',
     'olympia.lib.es.management.commands.reindex',
+    'olympia.ratings.tasks',
+    'olympia.reviewers.tasks',
+    'olympia.versions.tasks',
     'olympia.stats.management.commands.index_stats',
+    'olympia.tags.tasks',
+    'olympia.users.tasks',
+    'olympia.zadmin.tasks',
```
EnTeQuAk: This is missing `celery.scanners.tasks`

diox: You mean `olympia.scanners.tasks` ? It's not needed because it ends up automatically imported anyway. This setting is not an exhaustive list, it's just everything that doesn't get automatically imported at startup (and a few more that are automatically started at startup but not in tests as well)

PR #13423, opened 2020-02-06
Optimize number of queries made by draft comment API

### LINKED ISSUE #12779: Get comments API is slow, even when cached

* Browse a version in Code Manager like https://code.addons-dev.allizom.org/en-US/browse/616959/versions/1688394/
* Enter 10 or more comments
* Open the Network panel
* Reload the page
* The app makes a GET request to load the saved comments. See https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#get--api-v4-reviewers-addon-(int-addon_id)-versions-(int-version_id)-draft_comments-

### What happened?

The GET request is pretty slow. It's about 1 second for 10 comments and 2 seconds for 25 comments.

It should be faster 🚀 

Curiously, the response is still slow when it's cached using an etag. Maybe this is a server bug? If you think the frontend is doing something wrong for cached requests, please file an issue.

Here are some screenshots showing me loading 26 comments, with and without cache. In the cached response, the server did respond with a 304 status.

<details>
<summary>Screenshots</summary>

Screenshot 2019-10-31 15 02 24
Screenshot 2019-10-31 15 02 50

</details>

diox:
Fixes #12779

### DIFF
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -6243,6 +6243,37 @@ class TestDraftCommentViewSet(TestCase):
                 cls=amo.utils.AMOJSONEncoder))
         }
 
+    def test_list_queries(self):
+        user = user_factory(username='reviewer')
+        self.grant_permission(user, 'Addons:Review')
+        self.client.login_api(user)
+        DraftComment.objects.create(
+            version=self.version, comment='test1', user=user,
+            lineno=0, filename='manifest.json')
+        DraftComment.objects.create(
+            version=self.version, comment='test2', user=user,
+            lineno=1, filename='manifest.json')
+        DraftComment.objects.create(
+            version=self.version, comment='test3', user=user,
+            lineno=2, filename='manifest.json')
+        url = reverse_ns('reviewers-versions-draft-comment-list', kwargs={
+            'addon_pk': self.addon.pk,
+            'version_pk': self.version.pk
+        })
+        with self.assertNumQueries(15):
+            # - 2 savepoints because of tests
+            # - 2 user and groups
+            # - 2 addon and translations
+            # - 2 version and translations
+            # - 1 applications versions
+            # - 2 licenses and translations
+            # - 1 files
+            # - 1 file validation
+            # - 1 count
+            # - 1 drafts
+            response = self.client.get(url, {'lang': 'en-US'})
+        assert response.json()['count'] == 3
+
     def test_create_retrieve_and_update(self):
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/views.py
```diff
@@ -1428,12 +1434,26 @@ def get_object(self, **kwargs):
...
 
+    def get_addon_object(self):
+        if not hasattr(self, 'addon_object'):
+            self.addon_object = get_object_or_404(
+                # The serializer will not need to return much info about the
+                # addon, so we can use just the translations transformer and
+                # avoid the rest.
+                Addon.objects.get_queryset().only_translations(),
+                pk=self.kwargs['addon_pk'])
+        return self.addon_object
```
EnTeQuAk: Isn't that kinda what a `cached_property` is for? We aren't re-using the `self.addon_object` and `self.version_object` in any way at this moment unless I'm missing something?

diox: I didn't want to change the semantics, it's a method and not a property and that's why I'm re-implementing a caching mechanism instead of using `@cached_property` (The fact that it's a method aligns well with `get_object()` and our other APIs).

We do re-use the version, and that was one reason we had duplicate queries: `get_version_object()` is called in `get_queryset()`, `get_extra_comment_data()` and also in `get_serializer_context()`. The other reason was that the version object queryset was not "starting" from the addon, so every call to `version.addon` later made some additional queries to load the add-on (and its transformers) instead of loading it once. Technically I didn't have to extract the `get_addon_object()` and cache it into a property, but I did for consistency.

src/olympia/reviewers/tests/test_views.py
```diff
@@ -6243,6 +6243,37 @@ def test_create_and_retrieve(self):
...
+            version=self.version, comment='test2', user=user,
+            lineno=1, filename='manifest.json')
+        DraftComment.objects.create(
+            version=self.version, comment='test3', user=user,
+            lineno=2, filename='manifest.json')
+        url = reverse_ns('reviewers-versions-draft-comment-list', kwargs={
+            'addon_pk': self.addon.pk,
+            'version_pk': self.version.pk
+        })
+        with self.assertNumQueries(15):
```
diox: Same test without my changes we were doing 90+ queries.

src/olympia/reviewers/views.py
EnTeQuAk: Right, I forgot where `get_version_object` is called. Works for me, thanks for clarifying.

PR #13432, opened 2020-02-07
Stop excluding non-current listed versions in code search

### LINKED ISSUE #13431: Stop excluding non-current listed versions in code search
For the initial implementation of Scanner query rule task, we restricted the versions it runs on, excluding non-current listed versions. Now that we have done some initial testing, we've determined we no longer want to restrict versions that much.

STR:
- Submit a listed add-on, and then submit a new version for that add-on.
- Create a Scanner Query Rule that would match something in both versions and run it

Actual Results:
- Only the currently public version is in the results

Expected Results
- The 2 versions are found

diox:
Fixes #13431

### DIFF
--- a/src/olympia/scanners/tasks.py
+++ b/src/olympia/scanners/tasks.py
@@ -238,8 +238,10 @@ def mark_yara_query_rule_as_completed_or_aborted(query_rule_pk):
     rule = ScannerQueryRule.objects.get(pk=query_rule_pk)
     try:
         if rule.state == RUNNING:
+            log.info('Marking Yara Query Rule %s as completed', rule.pk)
             rule.change_state_to(COMPLETED)
         elif rule.state == ABORTING:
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/tasks.py
```diff
@@ -238,8 +237,10 @@ def mark_yara_query_rule_as_completed_or_aborted(query_rule_pk):
     rule = ScannerQueryRule.objects.get(pk=query_rule_pk)
     try:
         if rule.state == RUNNING:
+            log.info('Marking Yara Query Rule %s as completed', rule.pk)
```
eviljeff: these two `log.info` statements seem unrelated?

diox: Yeah, it's just additional logging that I slipped in because you can never have enough logging :)

PR #13447, opened 2020-02-10
route block edits via blocksubmission too

### LINKED ISSUE #13341: implement dual sign off for editing Blocks
split from #13202

Currently single block edits are applied immediately. For dual-sign (#13089) off to work we need to defer the update of the block to after sign-off. This is more complicated than adding because we still want to be able to view a Block before deciding to edit it.  
Ideas:
 - ~interrupt the save and create a BlockSubmission to actually make the change (breaks the save flow which the django docs warn not to do)~
- ~show a read-only view of the Block and a button that allows editing via the BlockSubmission form (which would have to be changed to not filter out the Block based on current min/max versions)~
- In the end I did a mix between the two - the form is edited as normal but the Save is redirected to be processed as a `BlockSubmission`

eviljeff:
fixes #13341

### DIFF
--- a/src/olympia/blocklist/admin.py
+++ b/src/olympia/blocklist/admin.py
@@ -287,7 +287,7 @@ class BlockSubmissionAdmin(admin.ModelAdmin):
             frm_data = form.data
             # Check if the versions specified were the ones we calculated which
             # Blocks would be updated or skipped on.
-            # Ignore for a single guid because we always update it irrespetive
+            # Ignore for a single guid because we always update it irrespective
             # of whether it needs to be updated.
             # TODO: make this more intelligent and don't force a refresh when
             # we have multiple new Blocks (but no existing blocks to update)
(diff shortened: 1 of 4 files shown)


src/olympia/blocklist/admin.py
```diff
@@ -287,8 +287,8 @@ def add_view(self, request, **kwargs):
             frm_data = form.data
             # Check if the versions specified were the ones we calculated which
             # Blocks would be updated or skipped on.
-            # Ignore for a single version because we only use it for new Blocks
-            # where the versions changing can't affect anything
+            # Ignore for a single guid because we always update it irrespetive
```
diox: Typo `irrespetive`

eviljeff: https://github.com/mozilla/addons-server/pull/13447/commits/8eb16b8339c05d83b130b8d3084367d5baf15353

PR #13450, opened 2020-02-10
Fallback to alternative file path if necessary in run_yara_query_rule_on_versions

### LINKED ISSUE #13344: Handle FileNotFoundError error in run_yara_query_rule_on_versions[_chunk]
```
  File "/data/olympia/src/olympia/scanners/tasks.py", line 316, in run_yara_query_rule_on_versions_chunk
    _run_yara_query_rule_on_version(version, rule)
  File "/data/olympia/src/olympia/scanners/tasks.py", line 331, in _run_yara_query_rule_on_version
    definition=rule.definition)
  File "/data/olympia/src/olympia/scanners/tasks.py", line 214, in _run_yara_for_path
    zip_file = SafeZip(source=path)
  File "/data/olympia/src/olympia/files/utils.py", line 739, in __init__
    self.is_valid = self.initialize_and_validate()
  File "/data/olympia/src/olympia/files/utils.py", line 752, in initialize_and_validate
    zip_file = zipfile.ZipFile(self.source, self.mode)
  File "/usr/local/lib/python3.6/zipfile.py", line 1113, in __init__
    self.fp = io.open(file, filemode)
<class 'FileNotFoundError'>
FileNotFoundError(2, 'No such file or directory')
```

diox:
Fixes #13344

### DIFF
--- a/src/olympia/scanners/tasks.py
+++ b/src/olympia/scanners/tasks.py
@@ -325,9 +325,20 @@ def _run_yara_query_rule_on_version(version, rule):
     """
     file_ = version.all_files[0]
     scanner_result = ScannerQueryResult(version=version, scanner=YARA)
-    _run_yara_for_path(
-        scanner_result, file_.current_file_path,
-        definition=rule.definition)
+    try:
+        _run_yara_for_path(
+            scanner_result, file_.current_file_path,
+            definition=rule.definition)
+    except FileNotFoundError:
+        # Fallback in case the file was disabled/re-enabled and not yet moved,
+        # we try the other possible path. This shouldn't happen too often.
+        tried_path = file_.current_file_path
+        fallback_path = (
+            file_.file_path if tried_path == file_.guarded_file_path
+            else file_.guarded_file_path
+        )
+        _run_yara_for_path(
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/tasks.py
```diff
@@ -325,9 +325,20 @@ def _run_yara_query_rule_on_version(version, rule):
...
+    except FileNotFoundError:
+        # Fallback in case the file was disabled/re-enabled and not yet moved,
+        # we try the other possible path. This shouldn't happen too often.
+        tried_path = file_.current_file_path
+        fallback_path = (
+            file_.file_path if tried_path == file_.guarded_file_path
+            else file_.guarded_file_path
+        )
+        _run_yara_for_path(
+            scanner_result, fallback_path, definition=rule.definition)
```
bobsilverberg: Is it possible that this `fallback_path` will also yield a `FileNotFoundError`?

diox: Yes: there could be a race condition, or the file could simply be missing for any number of reasons (since we're running searches on potentially old/buggy versions we have in the db). In that case it would simply raise the exception as normal, the parent task catches all exceptions, logs them and continues.

PR #13479, opened 2020-02-12
prevent block updates when pending

### LINKED ISSUE #13283: prevent further updates to an existing Block when it's guid is present in a BlockSubmission
follow on from  #13201 

We should show/block further edits in existing Block page when a BlockSubmission is outstanding.  Ideally the same restriction should be shown in the GUID input view too if a matching GUID is in an outstanding BlockSubmission

eviljeff:
fixes #13283

### DIFF
--- a/src/olympia/blocklist/admin.py
+++ b/src/olympia/blocklist/admin.py
@@ -502,9 +502,19 @@ class BlockAdmin(BlockAdminAddMixin, admin.ModelAdmin):
         return format_html('<a href="{}">{}</a>', obj.url, obj.url)
 
     def block_history(self, obj):
+        submission = BlockSubmission.get_submission_from_guid(obj.guid).first()
+        submission_log = (
+            format_html(
+                '<li>{date}. <a href="{url}">Changes pending</a></li>',
+                date=submission.modified.date(),
+                url=reverse(
+                    'admin:blocklist_blocksubmission_change',
+                    args=(submission.id,)))
+            if submission else '')
         return format_block_history(
             ActivityLog.objects.for_guidblock(obj.guid).filter(
-                action__in=Block.ACTIVITY_IDS).order_by('created'))
+                action__in=Block.ACTIVITY_IDS).order_by('created'),
+            additional_logs=submission_log)
 
     def get_fieldsets(self, request, obj):
         details = (
@@ -534,6 +544,12 @@ class BlockAdmin(BlockAdminAddMixin, admin.ModelAdmin):
 
         return (details, history, edit)
 
+    def has_change_permission(self, request, obj=None):
+        if obj and BlockSubmission.get_submission_from_guid(obj.guid):
+            return False
+        else:
+            return super().has_change_permission(request, obj=obj)
+
     def save_model(self, request, obj, form, change):
         # We don't actually save via this Admin so if we get here something has
         # gone wrong.
@@ -550,10 +566,11 @@ class BlockAdmin(BlockAdminAddMixin, admin.ModelAdmin):
 
     def get_form(self, request, obj=None, change=False, **kwargs):
         form = super().get_form(request, obj=obj, change=change, **kwargs)
-        form.base_fields['min_version'].choices = _get_version_choices(
-            obj, 'min_version')
-        form.base_fields['max_version'].choices = _get_version_choices(
-            obj, 'max_version')
+        if 'min_version' in form.base_fields:
+            form.base_fields['min_version'].choices = _get_version_choices(
+                obj, 'min_version')
+            form.base_fields['max_version'].choices = _get_version_choices(
+                obj, 'max_version')
         return form
 
     def formfield_for_dbfield(self, db_field, request, **kwargs):
@@ -573,6 +590,8 @@ class BlockAdmin(BlockAdminAddMixin, admin.ModelAdmin):
                 return HttpResponseTemporaryRedirect(
                     reverse('admin:blocklist_blocksubmission_add'))
 
+        extra_context = extra_context or {}
+        extra_context['show_save_and_continue'] = False
         return super().changeform_view(
             request, object_id=obj_id, form_url=form_url,
             extra_context=extra_context)
--- a/src/olympia/blocklist/admin.py
+++ b/src/olympia/blocklist/admin.py
@@ -502,7 +502,8 @@ class BlockAdmin(BlockAdminAddMixin, admin.ModelAdmin):
         return format_html('<a href="{}">{}</a>', obj.url, obj.url)
 
     def block_history(self, obj):
-        submission = BlockSubmission.get_submission_from_guid(obj.guid).first()
+        submission = BlockSubmission.get_submissions_from_guid(
+            obj.guid).first()
         submission_log = (
             format_html(
                 '<li>{date}. <a href="{url}">Changes pending</a></li>',
@@ -545,7 +546,7 @@ class BlockAdmin(BlockAdminAddMixin, admin.ModelAdmin):
         return (details, history, edit)
 
     def has_change_permission(self, request, obj=None):
-        if obj and BlockSubmission.get_submission_from_guid(obj.guid):
+        if obj and BlockSubmission.get_submissions_from_guid(obj.guid):
             return False
         else:
             return super().has_change_permission(request, obj=obj)
--- a/src/olympia/blocklist/forms.py
+++ b/src/olympia/blocklist/forms.py
@@ -48,9 +48,9 @@ class MultiAddForm(MultiGUIDInputForm):
                     _('Addon with GUID %(guid)s does not exist'),
                     params={'guid': guid}))
         for guid in guids:
-            if BlockSubmission.get_submission_from_guid(guid):
+            if BlockSubmission.get_submissions_from_guid(guid):
                 errors.append(ValidationError(
-                    _('GUID %(guid)s is already in pending BlockSubmission'),
+                    _('GUID %(guid)s is already in a pending BlockSubmission'),
                     params={'guid': guid}))
         if errors:
             raise ValidationError(errors)
--- a/src/olympia/blocklist/forms.py
+++ b/src/olympia/blocklist/forms.py
@@ -4,7 +4,7 @@ from django.utils.translation import ugettext_lazy as _
 
 from olympia.addons.models import Addon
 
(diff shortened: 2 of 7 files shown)


src/olympia/blocklist/templates/blocklist/block_change_form.html
```diff
@@ -4,13 +4,5 @@
 {% block submit_buttons_bottom %}
   {# This is so we can submit directly to the blocksubmission add view - it expect an input_guids field #}
   <input type="hidden" name="input_guids" value="{{ original.guid }}">
-  <div class="submit-row">
```
bobsilverberg: I don't understand this change. I can see that code that is similar to (but not exactly like) this exists in `src/olympia/blocklist/templates/blocklist/block_submission_change_form.html`, but why is this being removed from here. Also, what does `{{ block.super }}` do? Where is `"admin/change_form.html"` that this extends?

src/olympia/blocklist/tests/test_admin.py
```diff
@@ -1196,6 +1208,7 @@ def test_can_not_edit_without_permission(self):
         # Try to edit the block anyway
         response = self.client.post(
             self.change_url, {
+                'input_guids': self.block.guid,
```
bobsilverberg: Why was this change necessary in this patch?

src/olympia/blocklist/admin.py
```diff
@@ -502,9 +502,19 @@ def url_link(self, obj):
         return format_html('<a href="{}">{}</a>', obj.url, obj.url)
 
     def block_history(self, obj):
+        submission = BlockSubmission.get_submission_from_guid(obj.guid).first()
```
bobsilverberg: I assume from the `first()` that it's it possible that `get_submission_from_guid` could return more than one submission. Is it sufficient to only output the first pending submission found?

src/olympia/blocklist/admin.py
```diff
@@ -502,9 +502,19 @@ def url_link(self, obj):
...
+                date=submission.modified.date(),
+                url=reverse(
+                    'admin:blocklist_blocksubmission_change',
+                    args=(submission.id,)))
+            if submission else '')
         return format_block_history(
             ActivityLog.objects.for_guidblock(obj.guid).filter(
-                action__in=Block.ACTIVITY_IDS).order_by('created'))
+                action__in=Block.ACTIVITY_IDS).order_by('created'),
+            additional_logs=submission_log)
```
bobsilverberg: Should there be a test for this new code? I don't see one anywhere.

src/olympia/blocklist/admin.py
```diff
@@ -534,6 +544,12 @@ def get_fieldsets(self, request, obj):
 
         return (details, history, edit)
 
+    def has_change_permission(self, request, obj=None):
+        if obj and BlockSubmission.get_submission_from_guid(obj.guid):
```
bobsilverberg: Should there be a test for this, or is it considered covered by `test_cannot_edit_when_guid_in_blocksubmission`?

src/olympia/blocklist/admin.py
```diff
@@ -550,10 +566,11 @@ def delete_queryset(self, request, queryset):
 
     def get_form(self, request, obj=None, change=False, **kwargs):
         form = super().get_form(request, obj=obj, change=change, **kwargs)
-        form.base_fields['min_version'].choices = _get_version_choices(
-            obj, 'min_version')
-        form.base_fields['max_version'].choices = _get_version_choices(
-            obj, 'max_version')
+        if 'min_version' in form.base_fields:
```
bobsilverberg: Why is this change necessary for this patch? I don't see any changes that would cause `form.base_fields` to not have `min_version`.

src/olympia/blocklist/admin.py
```diff
@@ -573,6 +590,8 @@ def changeform_view(self, request, obj_id=None, form_url='',
                 return HttpResponseTemporaryRedirect(
                     reverse('admin:blocklist_blocksubmission_add'))
 
+        extra_context = extra_context or {}
```
bobsilverberg: I'm not really sure what the `extra_context` does or where it comes from, but instead of this could you just default `extra_context` to `{}` instead of `None` on line 582?

src/olympia/blocklist/forms.py
```diff
@@ -39,10 +39,18 @@ class MultiAddForm(MultiGUIDInputForm):
...
             if not blk and not Addon.unfiltered.filter(guid=guid).exists():
-                raise ValidationError(
+                errors.append(ValidationError(
                     _('Addon with GUID %(guid)s does not exist'),
-                    params={'guid': guid})
+                    params={'guid': guid}))
+        for guid in guids:
+            if BlockSubmission.get_submission_from_guid(guid):
+                errors.append(ValidationError(
+                    _('GUID %(guid)s is already in pending BlockSubmission'),
```
bobsilverberg: Should the message read: "GUID %(guid)s is already in *a* pending BlockSubmission"?

src/olympia/blocklist/tests/test_models.py
```diff
@@ -64,3 +64,25 @@ def test_is_save_to_blocks_permitted(self):
         # Except when that's not enforced locally
         with override_settings(DEBUG=True):
             assert block.is_save_to_blocks_permitted
+
+    def test_get_submission_from_guid(self):
```
bobsilverberg: For safety's sake, is it worth also having a test that proves that a `guid` that does not exist in a submission returns an empty list?

src/olympia/blocklist/models.py
```diff
@@ -383,6 +387,12 @@ def save_to_blocks(self):
 
         self.update(signoff_state=self.SIGNOFF_PUBLISHED)
 
+    @classmethod
+    def get_submission_from_guid(cls, guid, states=SIGNOFF_STATES_FINISHED):
+        return (
```
bobsilverberg: Will this return an empty list if nothing is found, or `None`?

src/olympia/blocklist/admin.py
bobsilverberg: Also, I asked this elsewhere, but I assume that `get_submission_from_guid` returns `None` if none are found and not an empty list?

src/olympia/blocklist/admin.py
eviljeff: yeah, it returns a queryset.  A guid should only be in one BlockSubmission at once (because adding it would otherwise have been blocked by the changes in this patch).  `.first()` is just an easy (lazy) shortcut because it will either return one `BlockSubmission` instance or `None`.

src/olympia/blocklist/admin.py
eviljeff: https://github.com/mozilla/addons-server/pull/13479/files#diff-55adb0dc3ea4f62a69a70ccd80833668R1238-R1240

src/olympia/blocklist/admin.py
eviljeff: that covers it - when `has_change_permission` is `False` the page is read only.

eviljeff: it returns a queryset which is falsey when there are no results.  (an empty list is also falsey in python)

src/olympia/blocklist/admin.py
eviljeff: when `has_change_permission` is `False` the form is empty so `.base_fields` doesn't contain the min/max fields.

src/olympia/blocklist/admin.py
eviljeff: > I'm not really sure what the extra_context does or where it comes from, 

It's extra data that will be passed to the templates when rendering (the "context" is the dict containing all the dynamic content needed to render).

> but instead of this could you just default extra_context to {} instead of None on line 582?  

In python the default args are reused for the function so if you define a mutable object and modify it then it will still be there _next_ time that function is called.  Currently in this function we're always modifying the object in the same way and only referencing it afterwards but imo it's good practise to take that common gotcha into account (if any of the code above was changed to reference `extra_context` it'd lead to a confusing bug).

src/olympia/blocklist/models.py
eviljeff: it returns a queryset which is falsey when there are no results.

src/olympia/blocklist/models.py
```diff
@@ -383,6 +387,12 @@ def save_to_blocks(self):
...
+            cls.objects.exclude(signoff_state__in=states)
+                       .filter(to_block__contains=f'"{guid}"'))
```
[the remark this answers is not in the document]
eviljeff: the field is storing JSON data so the `__contains` is just a way of matching it - there isn't native support for JSON in the database.  The `"..."` should ensure that only full guids are matched.  (Now I see there's an edge case if a guid is entered in an invalid format, but it'd be an error thrown on invalid input so no real harm there)

src/olympia/blocklist/templates/blocklist/block_change_form.html
eviljeff: `{{ block.super }}` means use the parent definition of `{% block submit_buttons_bottom %}` (like `super()` inside a python object function).  `admin/change_form.html` is the original django admin template https://github.com/django/django/blob/2.2.10/django/contrib/admin/templates/admin/change_form.html

I removed my custom template fragment because I realised it wasn't needed any longer.

src/olympia/blocklist/tests/test_admin.py
eviljeff: It wasn't strictly necessary but I noticed it so drive-by fixed it.  The `.post` would still fail anyway but if the permission checks were broken in subsequent patch then, without `input_guids`, it would fail in a different way (it'd 200 back to the same page because the form data was broken) which could be confusing.

src/olympia/blocklist/tests/test_models.py
eviljeff: https://github.com/mozilla/addons-server/pull/13479/commits/7595cf2bf80cbc3035ebc93e7ecbe3b1947e81c3

src/olympia/blocklist/forms.py
eviljeff: https://github.com/mozilla/addons-server/pull/13479/commits/7595cf2bf80cbc3035ebc93e7ecbe3b1947e81c3

Issue #12924, opened 2019-11-19
Incorrect mime type for Privacy Badger lib/basedomain.js file


wagnerand:
### Describe the problem and steps to reproduce it:

Load this file in the code manager https://code.addons.mozilla.org/en-US/browse/506646/versions/4904550/?path=lib%2Fbasedomain.js

### What happened?

It does not have JS syntax highlighting

### What did you expect to happen?

It should have JS syntax highlighting

### Anything else we should know?

The API call to retrieve the file content is `https://addons.mozilla.org/api/v5/reviewers/addon/506646/versions/4904550/?file=lib%2Fbasedomain.js&lang=en-US`

This incorrectly returns a text mimetype for the file:

<img width="468" alt="Screenshot 2019-11-20 09 07 02" src="https://user-images.githubusercontent.com/55398/69250527-6f9f2000-0b75-11ea-8f9a-5886a540f7ef.png">

wagnerand:
Another example, if needed: https://code.addons.mozilla.org/en-US/browse/2608730/versions/4922979/?path=background.js

bobsilverberg:
We can have a whitelist of file extensions for which we'll assume a mimetype. E.g., `.js` is JavaScript, `.html` is HTML. In addition to the front-end making those assumptions, on the server we can not bother doing mimetype detection for this list of file extensions.

wagnerand:
Let's start with
* HTML
* JS
* JSON
* CSS

bobsilverberg:
@EnTeQuAk Isn't this fixed via https://github.com/mozilla/addons-server/issues/13460? And if so, shouldn't the milestone be set to `2020.02.20`?

ioanarusiczki:
@EnTeQuAk 
The file https://code.addons-dev.allizom.org/en-US/browse/503019/versions/1541869/?path=lib%2Fbasedomain.js indicates a different MIME type now on AMO dev as the API https://addons-dev.allizom.org/api/v5/reviewers/addon/503019/versions/1541869/?path=lib/basedomain.js. I also see the code as being highlighted in code manager. 
Checking the Privacy Badger on AMO dev I also see the same behavior for a .css, .html or .json file.
 

![javascript](https://user-images.githubusercontent.com/33448286/75431146-148b7f00-5955-11ea-84b1-e65822d5fde4.png)

![highlighting](https://user-images.githubusercontent.com/33448286/75431165-1bb28d00-5955-11ea-9e6c-5847a4a8258e.png)

PR #13488, opened 2020-02-13
Move to a rolling debian-testing release for libmagic and libgit2

### LINKED ISSUE #13487: Move from debian-buster (prev testing) to rolling testing
We do have one feature that relies on the most recent features to be available and that's our `libmagic` integration.

Another one depends on an up-to-date libgit2 library - see #13391 

Let's move from an explicit debian buster - which we're currently using - to simply using debian testing to have a rolling target.

EnTeQuAk:
Fixes #13487
Unblocks #13391

And at least refs #12924, #12357 (these need further testing)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,8 +20,7 @@ x-env-mapping: &env
 services:
   worker: &worker
     <<: *env
-    # image: addons/addons-server:latest
-    build: .
+    image: addons/addons-server:latest
     command: supervisord -n -c /code/docker/supervisor-celery.conf
     user: olympia
     volumes:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -20,7 +20,8 @@ x-env-mapping: &env
 services:
   worker: &worker
     <<: *env
-    image: addons/addons-server:latest
+    # image: addons/addons-server:latest
+    build: .
```
diox: mmm I assume you had that for local testing ? Once we build a new image and push it it shouldn't be necessary, so I vote for not keeping it in the PR

EnTeQuAk: Indeed, sorry I forgot to roll that change back. Fixed in 4f09ec7d39

PR #13507, opened 2020-02-17
Route each task explicitly

### LINKED ISSUE #13506: Explicitly route each celery task to a specific queue
Our `default` celery queue is used by too many tasks. It's usually fine, except last Thursday when we ran a manual git extraction of all unlisted add-ons: suddenly the `default` queue was full and some critical tasks where still incorrectly routed there, but because the queue was so full they were never executed as quickly as we wanted. See https://github.com/mozilla/addons-server/issues/13491 for more details.

To avoid this, let's try to avoid using `default` entirely, explicitly routing each task to a specific queue.

diox:
Fixes #13506


src/olympia/lib/settings_base.py
```diff
@@ -1060,21 +1060,14 @@ def get_db_config(environ_var, atomic_requests=True):
...
     'olympia.lib.crypto.tasks',
     'olympia.lib.es.management.commands.reindex',
-    'olympia.ratings.tasks',
-    'olympia.reviewers.tasks',
-    'olympia.versions.tasks',
     'olympia.stats.management.commands.index_stats',
-    'olympia.tags.tasks',
-    'olympia.users.tasks',
-    'olympia.zadmin.tasks',
+    'olympia.lib.es.tests.test_commands',
```
eviljeff: mmm, could that line be appended to the list in `settings_test` to avoid specifying test code in `settings_base`?

src/olympia/lib/settings_base.py
```diff
@@ -1150,11 +1163,36 @@ def get_db_config(environ_var, atomic_requests=True):
     'olympia.amo.tasks.delete_logs': {'queue': 'amo'},
     'olympia.amo.tasks.send_email': {'queue': 'amo'},
     'olympia.amo.tasks.set_modified_on_object': {'queue': 'amo'},
+    'olympia.amo.tasks.sync_object_to_basket': {'queue': 'amo'},
+
+    # Test tasks that will never really be triggered in prod.
+    'olympia.amo.tests.test_celery.fake_task': {'queue': 'amo'},
+    'olympia.amo.tests.test_celery.fake_task_with_result': {'queue': 'amo'},
+    'olympia.amo.tests.test_celery.sleeping_task': {'queue': 'amo'},
+    'olympia.lib.es.tests.test_commands.dummy_task': {'queue': 'amo'},
```
eviljeff: I guess all these too

src/olympia/lib/settings_base.py
diox: Yeah, didn't think about that but that's a good point.

PR #13509, opened 2020-02-17
Add option (default: False) to run a scanner query rule on disabled add-ons

### LINKED ISSUE #13490: Add option to include disabled addons in code-search
For some use cases it would be useful to have an option upfront to include disabled addons in the code-search, for example during the creation of a rule.

Since that can have significant impact on the performance, it should be off by default.

diox:
Fixes #13490


src/olympia/scanners/tasks.py
```diff
@@ -267,18 +267,19 @@ def run_yara_query_rule(query_rule_pk):
...
     ).filter(
         addon__type=amo.ADDON_EXTENSION,
         files__is_webextension=True,
-    ).exclude(
-        addon__status=amo.STATUS_DISABLED,
-    ).values_list('id', flat=True).order_by('pk')
+    )
+    if not rule.run_on_disabled_addons:
+        qs = qs.exclude(addon__status=amo.STATUS_DISABLED)
+    qs = qs.values_list('id', flat=True).order_by('pk')
```
bobsilverberg: Nit: This was called `pks` previously, I assume because it's a list of pks and not a queryset. It is still a list of pks here, so maybe it should still be called `pks` here and on line 282?

diox: I never liked to call this `pks` : it is, at this point, technically still a queryset. It just so happens that it's a *values* queryset, containing only pks, but because we can still apply filters etc on it at this point (and we do if not excluding disabled add-ons), I preferred to rename it to `qs`. If I had assigned the `list(qs)` to a variable I would have named that variable `pks`.

PR #13512, opened 2020-02-17
Try notifications.slack.room

muffinresearch:
Using this PR to test changes. It should notify if it works.


muffinresearch:
So weirdly messages have started appearing without this change. I think we can leave the config as is for now.

PR #13517, opened 2020-02-18
Fix dockerfile deploy.

EnTeQuAk:
Fixes the Dockerfile.deploy to include the correct apt-get sources.list
file.

### DIFF
--- a/Dockerfile.deploy
+++ b/Dockerfile.deploy
@@ -16,7 +16,7 @@ RUN apt-get update && apt-get install -y \
     && rm -rf /var/lib/apt/lists/*
 RUN cat /etc/pki/gpg/GPG-KEY-nodesource | apt-key add -
 ADD docker/debian-stretch-nodesource-repo /etc/apt/sources.list.d/nodesource.list
-ADD docker/debian-buster-testing-repo /etc/apt/sources.list.d/testing.list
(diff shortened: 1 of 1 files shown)


muffinresearch:
Are we safe to use test apt sources in production? Presumably this means if a newer version of any package that we use is in the test repo it will be installed.

EnTeQuAk:
> Are we safe to use test apt sources in production? Presumably this means if a newer version of any package that we use is in the test repo it will be installed.

We are only doing this explicitly for `file`, `libmagic` and libgit2-dev. We'll see failures on -dev long before we do on prod in case something happens but we're pretty safe in that regard imo.

muffinresearch:
> > Are we safe to use test apt sources in production? Presumably this means if a newer version of any package that we use is in the test repo it will be installed.
> 
> We are only doing this explicitly for `file`, `libmagic` and `libgit2-dev`. We'll see failures on -dev long before we do on prod in case something happens but we're pretty safe in that regard imo.

I see, so this is limited to specific packages by the `-t` option to `apt-get`. Despite Still makes me a little bit nervous trusting a potentially unstable dist source.

EnTeQuAk:
Closing for now to work on this a bit more

PR #13518, opened 2020-02-18
Error when existing blocks need to be recalculated rather than on all min/max changes

### LINKED ISSUE #13480: Don't force a refresh on a new BlockSubmission when changing the min/max unless nessecary
Currently if the min/max versions are changed for a multiple guid submission we force a refresh to the page to update the list of existing Blocks that would be impacted by the change.  Instead, we should make it more intelligent and don't do that if all the guids are new guids.  (i.e. no existing Blocks are referenced).   To make it clearer we might want to explicitly show a validation error to make it clearer what's happening.

eviljeff:
fixes #13480

### DIFF
--- a/src/olympia/blocklist/forms.py
+++ b/src/olympia/blocklist/forms.py
@@ -62,21 +62,40 @@ class BlockSubmissionForm(forms.ModelForm):
     existing_max_version = forms.fields.CharField(
         widget=forms.widgets.HiddenInput, required=False)
 
+    def _check_if_existing_blocks_changed(self, all_guids, v_min, v_max,
+                                          existing_v_min, existing_v_max):
+        # shortcut if the min/max versions havn't changed
+        if v_min == existing_v_min and v_max == existing_v_max:
+            return False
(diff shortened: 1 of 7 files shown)


src/olympia/blocklist/forms.py
```diff
@@ -54,3 +54,48 @@ def clean(self):
...
+        guids = splitlines(data.get('input_guids'))
+        # Ignore for a single guid because we always update it irrespective of
+        # whether it needs to be updated.
+        if len(guids) > 1:
+            blocks_have_changed = self._check_if_existing_blocks_changed(
+                guids,
+                data.get('min_version'),
+                data.get('max_version'),
+                data.get('existing_min_version'),
+                data.get('existing_max_version'))
```
EnTeQuAk: For so many arguments I usually prefer naming the arguments so that you have `all_guids=..., v_min=...` etc - I think it works here because it's only used internally but worth keeping in mind should more arguments be added in the future.

PR #13525, opened 2020-02-18
rewrite format_block_history as a template fragment for safety

### LINKED ISSUE #13489: Block history is not populated in block detail view when the add-on goes through dual sign-off
1. Log in with a user that has `Blocklist:Create` permissions
2. Open the Blocklist model and add a new add-on to the list (make sure that you choose an add-on that requires sign-off - i.e. add-on with no stats or with over 100k ADU)
3. With a user that has `Blocklist:Signoff` permissions, open the Block Submissions list and look for the new block request (it should be Pending for sign off)
4. Approve the block submission request
5. Open the block detail page and check the Block history field

### What happened?
The Block history is empty
The Block history should be populated in block detail pages when add-ons are going through the dual sign off process

- this is also valid for block edits that require sign-off
- the Block history field is populated when the add-on doesn't require dual sign-off

[Example](https://addons-internal.dev.mozaws.net/en-US/admin/models/blocklist/block/1013/change/)
image

eviljeff:
fixes #13489

### DIFF
--- a/src/olympia/blocklist/templates/blocklist/includes/logs.html
+++ b/src/olympia/blocklist/templates/blocklist/includes/logs.html
@@ -21,5 +21,10 @@
         </ul>
       </li>
     {% endfor %}
-    {{ addtional_content }}
+    {% if block_submission %}
+      <li>
+        {{ block_submission.modified|date:"Y-m-d" }}.
+        <a href="{% url 'admin:blocklist_blocksubmission_change' block_submission.id %}">Changes pending</a>
+      </li>
+    {% endif %}
 </ul>
--- a/src/olympia/blocklist/templates/blocklist/includes/logs.html
+++ b/src/olympia/blocklist/templates/blocklist/includes/logs.html
@@ -1,0 +1,25 @@
+{% load i18n %}
+
+<ul>
+    {% for log in logs %}
+      <li>
+        {% if log.details.url %}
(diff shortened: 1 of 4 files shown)


src/olympia/blocklist/templates/blocklist/includes/logs.html
```diff
@@ -0,0 +1,25 @@
...
+        {% endif %}
+        {% if log.details.include_in_legacy %}
+            {% trans 'Included in legacy blocklist.' %}
+        {% endif %}
+        <ul>
+            <li>{{ log.details.reason }}</li>
+        </ul>
+      </li>
+    {% endfor %}
+    {{ addtional_content }}
```
diox: How about moving the `format_html()` call generating this to the template as well, for consistency ? I think you only need to pass the `submission` instance to do this.

eviljeff: I guess I could too.  (I was only thinking about the increasingly convoluted code for the Block logs.)

diox: Yeah re-reading this, especially because it's inside a `<ul>` I think it's best to move everything to the template. There is no good reason to do this outside now.

eviljeff: https://github.com/mozilla/addons-server/pull/13525/commits/29a4847976c79bd14ad0d93f7c78bdcb530ba66e

PR #13531, opened 2020-02-18
Hide disabled files when delete an add-on

### LINKED ISSUE #13433: File temporarily unavailable after add-on deletion

When an add-on gets deleted, the file is unavailable in the reviewer and admin tools. I believe this is temporary, until the cronjobs have run. That can take a couple of hours tough, affecting admins.

Assuming STR:
(This might depend on the time of the hour you run this, due to the cronjobs runnin at :25 and :55 each hour)
1. Submit a new add-on
2. Delete the add-on
3. As a reviewer, try to download the file in the reviewer tools, inspect the code using the file-viewer or the code-manager
4. As an admin, visit the django file admin page for the submitted file and try to download the file using the download link. 

### What happened?

I expect 3 and 4 to fail.

At least 4 should work. It would be great if 3 worked as well.

mirefly:
Fixes #13433.

I added `file.tasks.hide_disabled_file` and let `addon.delete` fire up that task. 

As to force-disabled and user-disabled add-ons, I found that we disabled the add-on by [calling `addon.update`](https://github.com/mozilla/addons-server/blob/8785d836caecf8df192e5ccb3141dd3eb38b538b/src/olympia/addons/models.py#L507) which [triggers watcher to hide the files](https://github.com/mozilla/addons-server/blob/8785d836caecf8df192e5ccb3141dd3eb38b538b/src/olympia/addons/models.py#L1584). So I guess we do not need changes for these two cases (?).

I also updated the `docker-compose.yml` for mounting `guarded-addons` folder, and created a PR https://github.com/mozilla/addons-nginx/pull/16

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,6 +39,7 @@ services:
       - ./site-static:/srv/site-static
       - ./storage/shared_storage/uploads:/srv/user-media
       - ./storage/files:/srv/user-media/addons
+      - ./storage/guarded-addons:/srv/user-media/guarded-addons
     ports:
       - "80:80"
 
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -628,10 +628,11 @@ class Addon(OnChangeMixin, ModelBase):
             self._ratings.all().delete()
             # We avoid triggering signals for Version & File on purpose to
             # avoid extra work. Files will be moved to the correct storage
-            # location with hide_disabled_files cron later.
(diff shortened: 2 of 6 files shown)


diox:
I had forgotten about this watcher. So we already do it synchronously in some cases then. I wonder if this should be moved to a task as well... I need to investigate, test on stage with add-ons with lots of versions and see how it behaves.

mirefly:
Thanks for reviewing this patch @diox . I updated it.

PR #13549, opened 2020-02-21
Display matching filenames & authors in scanner query results changelist

### LINKED ISSUE #13206: Add admin UI to view scanner results for arbitrary queries
Once #13205 is done, we'll want to be able to display the results of the arbitrary yara queries separately from the automated scanner results that happen at upload time.

diox:
Also stop displaying action buttons, we don't want them in the query results
at this time.

Fixes #13206


diox:
We used to inherit `list_display` from the mixin class, we don't any more, that removed the buttons.

PR #13554, opened 2020-02-21
use BlockSubmission for deletes too

### LINKED ISSUE #13086: Block delete_selected action needs dual sign-off
In `BlockAdmin.delete_queryset` we add some extra ActivityLog logging before the delete is carried out.  This is fine, but it's also quite expensive - for each Block delete there will be 3 activity log records, plus 1 extra for each version the addon has.  To fk link the ActivityLog instance with VersionLog and AddonLog there also needs to be a query on Addon, and a query on Version.

At a minimum we should preload `Block.addon` and `Block.addon_versions` so there isn't a separate query for each Block; potentially we should batch the processing in a celery task too.

We won't know how much of a problem this will be unless the blocklist is actually used.  After #12461 adds kinto integration moving the delete to a celery task might be necessary anyway.

### LINKED ISSUE #13444: Block multiple versions flow for unlisted add-ons is broken
1. Log in with a super admin
2. Open the Unlisted review queue - https://reviewers.addons-dev.allizom.org/en-US/reviewers/unlisted_queue/all
3. Select an add-on from the list
4. Use the Block multiple versions option to reject one or more versions
5. When redirected to the admin `/blocksubmission/add/` view, change the version range or edit any other editable fields
6. Save your changes

### What happened?
The information changed by the user is nor saved and the block request is not processed

The information added by the user is saved and the block request is successfully processed

- if you verify the version details in Reviewer Tools, you can see that the status is Disabled by Mozilla, so only the blocklisting flow is broken 
- this looks like a regression because the above scenario used to work; maybe it was introduced when we transitioned to the dual sign off process?
- reproduced on -dev and stage

eviljeff:
fixes #13086 - sorry, it's a bit sprawling, but there needed to be lots of little changes to reuse BlockSubmission for deletes too, like Jorge wanted.  (Really, `BlockSubmission` could do with a rename but I didn't want the extra code churn in this patch).  

Follow-up to re-add individual Block deletes - #13555

-- edit I believe this should fix #13444 too


src/olympia/blocklist/admin.py
```diff
@@ -523,20 +550,15 @@ def has_change_permission(self, request, obj=None):
         else:
             return super().has_change_permission(request, obj=obj)
 
+    def has_delete_permission(self, request, obj=None):
+        # TODO: restore this functionality so bulk delete isn't the only option
```
EnTeQuAk: We should file a follow-up issue and link it here.

eviljeff: https://github.com/mozilla/addons-server/issues/13555 - its in the project

Issue #13278, opened 2020-01-16
Allow BlocklistSubmission instances to be deleted


eviljeff:
BlocklistSubmission objects are created to allow signoff of block submissions to the blocklist.  To keep during the process they can't be deleted via django admin. Once they have been approved and the blocks created they could be regarded as obsolete.  We could determine in what circumstances they should be deletable and by who. 

Also consider other statuses - rejected, pending.

PR #13564, opened 2020-02-24
in reviewer tools check for active blocksubmission too

### LINKED ISSUE #13522: The admin should not be able to create a new block from rev tools for an add-on with dual sign off
STR:
1. Block an add-on using admin tools with `Add block` which required dual sign-off
2. While the add-on is in Pending state go to rev tools to it's review page. Example: https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/shret-sheasaur-starse2
3. As an admin click the Block add-on blue button

Expected result:
The button should be in "Update" state or greyed out already.

Actual result:
Another block request is created and is available in admin tools in `pending` state. 

Notes:
This issue is reproducible on dev with FF73(Win10).

eviljeff:
fixes #13522

### DIFF
--- a/src/olympia/addons/tests/test_models.py
+++ b/src/olympia/addons/tests/test_models.py
@@ -22,7 +22,7 @@ from olympia.amo.tests import (
 from olympia.amo.tests.test_models import BasePreviewMixin
 from olympia.applications.models import AppVersion
 from olympia.bandwagon.models import Collection
-from olympia.blocklist.models import Block
+from olympia.blocklist.models import Block, BlockSubmission
 from olympia.constants.categories import CATEGORIES
 from olympia.devhub.models import RssKey
 from olympia.discovery.models import DiscoveryItem
@@ -1674,18 +1674,29 @@ class TestAddonModels(TestCase):
     def test_block_property(self):
         addon = Addon.objects.get(id=3615)
         assert addon.block is None
-        assert not addon.is_addon_blocklisted
 
         del addon.block
         block = Block.objects.create(
             guid=addon.guid, updated_by=user_factory())
-        assert addon.block == addon.block
-        assert addon.is_addon_blocklisted
+        assert addon.block == block
 
         del addon.block
         block.update(guid='not-a-guid')
         assert addon.block is None
-        assert not addon.is_addon_blocklisted
+
+    def test_blocksubmission_property(self):
+        addon = Addon.objects.get(id=3615)
+        assert addon.blocksubmission is None
+
+        del addon.blocksubmission
+        submission = BlockSubmission.objects.create(
+            input_guids=addon.guid, updated_by=user_factory())
+        assert addon.blocksubmission == submission
+
+        del addon.blocksubmission
+        submission.update(input_guids='not-a-guid')
+        submission.update(to_block=None)
+        assert addon.blocksubmission is None
 
 
 class TestShouldRedirectToSubmitFlow(TestCase):
--- a/src/olympia/reviewers/templates/reviewers/review.html
+++ b/src/olympia/reviewers/templates/reviewers/review.html
@@ -245,18 +245,18 @@
                   data-toggle-button-selector="#force_disable_addon"
                   id="force_enable_addon" type="button">{{ _('Force-enable add-on') }}</button>
         </li>
-
-        {% if not addon.block %}
         <li>
+        {% if addon.blocksubmission %}
+          <a href="{{ url('admin:blocklist_blocksubmission_change', addon.blocksubmission.id) }}" class="button"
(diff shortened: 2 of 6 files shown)


src/olympia/addons/tests/test_models.py
```diff
@@ -1674,18 +1674,29 @@ def test_addon_author_delete_not_synced_to_basket_if_addon_is_deleted(
...
+        addon = Addon.objects.get(id=3615)
+        assert addon.blocksubmission is None
+
+        del addon.blocksubmission
+        submission = BlockSubmission.objects.create(
+            input_guids=addon.guid, updated_by=user_factory())
+        assert addon.blocksubmission == submission
+
+        del addon.blocksubmission
+        submission.update(input_guids='not-a-guid')
```
bobsilverberg: Here we're updating the `submission` so it no longer includes the original `guid` (I assume). Should there also be a test where the `submission` itself is deleted, and therefore the `guid` is no longer part of a submission, or is that not something that can happen?

src/olympia/reviewers/templates/reviewers/review.html
```diff
@@ -245,18 +245,18 @@ <h3 id="history">
...
+        {% if addon.blocksubmission %}
+          <a href="{{ url('admin:blocklist_blocksubmission_change', addon.blocksubmission.id) }}" class="button"
+                  id="edit_addon_blocksubmission" type="button">{{ _('View Block Submission') }}</a>
+        {% elif not addon.block %}
           <a href="{{ url('admin:blocklist_blocksubmission_add') }}?guids={{ addon.guid }}" class="button"
                   id="block_addon" type="button">{{ _('Block add-on') }}</a>
-        </li>
         {% else %}
-        <li>
           <a href="{{ url('admin:blocklist_block_change', addon.block.id) }}" class="button"
```
bobsilverberg: Just clarifying here: This last case would occur when `addon.block = True` and ` addon.blocksubmission = False`. Does that mean that a block exists for the addon, but that a pending submission does not exist?

src/olympia/reviewers/templates/reviewers/review.html
```diff
@@ -245,18 +245,18 @@ <h3 id="history">
                   data-toggle-button-selector="#force_disable_addon"
                   id="force_enable_addon" type="button">{{ _('Force-enable add-on') }}</button>
         </li>
-
-        {% if not addon.block %}
         <li>
+        {% if addon.blocksubmission %}
+          <a href="{{ url('admin:blocklist_blocksubmission_change', addon.blocksubmission.id) }}" class="button"
+                  id="edit_addon_blocksubmission" type="button">{{ _('View Block Submission') }}</a>
```
bobsilverberg: Just wondering why the `id` is `edit_addon_blocksubmission` but the text says "View Block Submission". Should it say "Edit Block Submission"?

src/olympia/reviewers/tests/test_utils.py
```diff
@@ -1492,8 +1492,18 @@ def _test_block_multiple_unlisted_versions(self, redirect_url):
...
+            redirect_url = redirect_url % (
+                old_version.version, self.version.version)
+        assert self.helper.redirect_url == redirect_url
+
+    def test_pending_blocksubmission_multiple_unlisted_versions(self):
+        subm = BlockSubmission.objects.create(
+            input_guids=self.addon.guid, updated_by=user_factory())
+        redirect_url = (
+            reverse('admin:blocklist_blocksubmission_change', args=(subm.id,)))
+        assert Block.objects.count() == 0
```
bobsilverberg: I don't understand why this assert is here.

src/olympia/addons/tests/test_models.py
eviljeff: They can't be deleted right now (#13278) though when the submission has finished (state changed to `SIGNOFF_PUBLISHED` or `SIGNOFF_REJECTED`) they don't, by default, get returned by `get_submissions_from_guid`

src/olympia/reviewers/templates/reviewers/review.html
eviljeff: Yes.  (`Addon.block` and `Addon.blocksubmission` actually return the instance or `None` but in python that's truthy or falsey)

src/olympia/reviewers/templates/reviewers/review.html
eviljeff: Technically it's a change page (edit) but depending on which permissions you have it would either be something you could change or something you could view and signoff.

src/olympia/reviewers/tests/test_utils.py
eviljeff: it's checking that just visiting the url isn't creating a Block - you would need to take a further action to create something.

PR #13572, opened 2020-02-25
Normalize cache key before caching when caching sha256 hashes.

### LINKED ISSUE #13272: Reviewers API responds with a 500 error when trying to view a file with blank spaces in its name
1. Send the following request to the reviewers/browse API:
```
https://addons-dev.allizom.org/api/v5/reviewers/addon/618629/versions/1690063/?file=content script.js
```

### What happened?
The API responds with a 500 server error because the file name contains a blank space

No server error

- the issue is also reproducible for the compare API and will completely prevent the compare page from loading in code-manager:

https://code.addons-dev.allizom.org/en-US/compare/618629/versions/1690062...1690063/

Update: - the issue can also happen if a folder name contains spaces - see [example](https://code.addons-dev.allizom.org/en-US/browse/618821/versions/1690254/?path=more%20rules%2Fmore%20more%20rules%2Fmore%283%29%20rules%2Fcoinhive.js)

EnTeQuAk:
Fixes #13272

### DIFF
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -96,11 +96,13 @@ class FileEntriesSerializer(FileSerializer):
         commit = self._get_commit(obj)
         tree = self.repo.get_root_tree(commit)
 
-        selected_file_key = make_key(
-            selected_file, with_locale=False, normalize=True)
-        cache_key = (
+        # Normalize the key as we want to avoid that we exceed max
+        # key lengh because of selected_file.
+        cache_key = make_key(
             f'reviewers:fileentriesserializer:hashes'
-            f':{commit.hex}:{selected_file_key}')
+            f':{commit.hex}:{selected_file}',
+            with_locale=False,
+            normalize=True)
 
         def _calculate_hash():
             try:
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -96,11 +96,11 @@ class FileEntriesSerializer(FileSerializer):
         commit = self._get_commit(obj)
         tree = self.repo.get_root_tree(commit)
 
-        cache_key = make_key(
+        selected_file_key = make_key(
+            selected_file, with_locale=False, normalize=True)
+        cache_key = (
             f'reviewers:fileentriesserializer:hashes'
-            f':{commit.hex}:{selected_file}',
-            with_locale=False,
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/serializers.py
```diff
@@ -95,9 +96,11 @@ def _get_hash_for_selected_file(self, obj):
         commit = self._get_commit(obj)
         tree = self.repo.get_root_tree(commit)
 
-        cache_key = (
+        cache_key = make_key(
             f'reviewers:fileentriesserializer:hashes'
-            f':{commit.hex}:{selected_file}')
+            f':{commit.hex}:{selected_file}',
+            with_locale=False,
+            normalize=True)
```
eviljeff: I don't disagree with the solution, but wouldn't it have been enough to normalize/encode `selected_file`?

src/olympia/reviewers/tests/test_views.py
```diff
@@ -6086,6 +6086,38 @@ def test_non_existent_requested_file_returns_404(self):
...
         assert response.status_code == 404
 
+    def test_requested_file_contains_whitespace(self):
+        new_version = version_factory(
+            addon=self.addon, file_kw={'filename': 'webextension_no_id.xpi'})
+
+        repo = AddonGitRepository.extract_and_commit_from_version(new_version)
+
+        apply_changes(
+            repo, new_version, u'(function() {})\n', u'content script.js')
```
eviljeff: `u'...'` 
gpg3k75

EnTeQuAk: Damn you copy & paste :see_no_evil:

src/olympia/reviewers/serializers.py
EnTeQuAk: I think that makes sense and ensures the key stays readable. 
0ce5268

src/olympia/reviewers/tests/test_views.py
EnTeQuAk: c025ed4

src/olympia/reviewers/serializers.py
EnTeQuAk: As @diox correctly pointed out we need to avoid to exceed memcached max key lengh so `make_key` over the whole cache key makes much more sense. I added a comment.

PR #13592, opened 2020-02-27
Keep random order stable for 24 hours in search API

### LINKED ISSUE #13574: Sorting for "See more Recommended extensions"
From the AMO homepage click "See more Recommended extensions" text link within its shelf.

### What happened?
You're taken to a "search results" type page, with pagination that currently runs to four pages. 

Since the Search results page tells me there are (currently) "97 extensions found" I expect the four pages to list them in some type of consistent order in case I want to peruse the entire list of Recommended extensions. However when you click from page to page you find an inconsistent ordering and often see extensions you already saw on previous pages. User is never certain if they've seen the entire list.

diox:
Fixes #13574

### DIFF
--- a/src/olympia/search/tests/test_filters.py
+++ b/src/olympia/search/tests/test_filters.py
@@ -440,7 +440,7 @@ class TestSortingFilter(FilterTestsBase):
             {'random_score': {'seed': 737481}}
         ]
 
-    @freeze_time('2020-02-26')
+    @freeze_time('2020-02-27')
     def test_sort_random(self):
         qs = self._filter(data={'recommended': 'true', 'sort': 'random'})
         # Note: this test does not call AddonRecommendedQueryParam so it won't
@@ -448,7 +448,7 @@ class TestSortingFilter(FilterTestsBase):
         # TestCombinedFilter.test_filter_recommended_sort_random
         assert qs['sort'] == ['_score']
         assert qs['query']['function_score']['functions'] == [
-            {'random_score': {'seed': 737481}}
+            {'random_score': {'seed': 737482}}
         ]
 
     def test_sort_recommended_only(self):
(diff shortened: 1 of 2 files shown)


src/olympia/search/tests/test_filters.py
```diff
@@ -428,24 +429,26 @@ def test_sort_random_restrictions(self):
...
+    @freeze_time('2020-02-26')
     def test_sort_random_featured(self):
         qs = self._filter(data={'featured': 'true', 'sort': 'random'})
         # Note: this test does not call AddonFeaturedQueryParam so it won't
         # apply the featured filtering. That's tested below in
         # TestCombinedFilter.test_filter_featured_sort_random
         assert qs['sort'] == ['_score']
         assert qs['query']['function_score']['functions'] == [
-            {'random_score': {}}
+            {'random_score': {'seed': 737481}}
```
bobsilverberg: This seems to leave room for human error. Would it be better to declare a const for the date, and then use that date both as input to `@freeze_time()`, and also do our own calculation of `.toordinal()` on the const? Then we're only specifying the value once, and not relying on a manual calculation in the test.

diox: I altered the value in one of the tests in https://github.com/mozilla/addons-server/pull/13592/commits/e95d109407aaf549cf2485c6f8a2e956ed3abca7 to test 2 different dates.

I kinda prefer hardcoding the value: that way we don't rely on a potentially unknown computation, we know exactly what we're going to send. With the second test having a different value I think it should be a little more reliable.

bobsilverberg: I'd argue it's not unknown at all, as you know how it's being calculated in the code under test, but this is just a nit and I'm fine with it landing as it now stands.

PR #13622, opened 2020-03-02
Allow users to reject their own blocklist submission

### LINKED ISSUE #13308: A user who submitted a block request doesn't have the option to reject his own submission
Follow up for https://github.com/mozilla/addons-server/issues/13089#issuecomment-566736627

1. Open the Admin Block submissions module
2. Submit a new multi block request (which is the only one that supports the dual sign-off process currently)
3. After the request is submitted, try to Reject your own submission

### What happened?
There is no Reject button present in the `/blocksubmission/change/` form for the user who submitted the request

The user who submitted a block should have an option to roll-back his request

This has been brought into discussion in #13089 and several times and this specific behavior was outlined in https://github.com/mozilla/addons-server/issues/13089#issuecomment-566736627
> > Also, every user should be able to roll-back their changes.
> 
> Agree. That's implicit in the proposed flow (a user can reject their own change), but it's good to call out.

### LINKED ISSUE #13594: implement TestBlockSubmissionAdmin.test_cannot_reject_with_only_block_create_permission test
Looks like I forgot to implement the `test_cannot_reject_with_only_block_create_permission`
 test - it's still just `pass`.

eviljeff:
fixes #13308 and fixes #13594 along the way

### DIFF
--- a/src/olympia/blocklist/admin.py
+++ b/src/olympia/blocklist/admin.py
@@ -144,17 +144,18 @@ class BlockSubmissionAdmin(admin.ModelAdmin):
         we need to return true if the user has sign-off permission instead.
         We can override that permissive behavior with `strict=True`."""
         change_perm = super().has_change_permission(request, obj=obj)
-        signoff_perm = self.has_signoff_permission(request, obj=obj)
-        either_perm = change_perm or (signoff_perm and not strict)
+        approve_perm = self.has_signoff_approve_permission(request, obj=obj)
+        either_perm = change_perm or (approve_perm and not strict)
         return either_perm and (not obj or self.is_pending_signoff(obj))
 
     def has_view_permission(self, request, obj=None):
         return (
             super().has_view_permission(request, obj) or
-            self.has_signoff_permission(request, obj))
+            self.has_signoff_approve_permission(request, obj) or
+            self.has_signoff_reject_permission(request, obj))
 
-    def has_signoff_permission(self, request, obj=None):
-        """ This controls whether the sign-off approve and reject actions are
+    def has_signoff_approve_permission(self, request, obj=None):
+        """ This controls whether the sign-off approve action is
         available on the change form.  `BlockSubmission.can_user_signoff`
         confirms the current user, who will signoff, is different from the user
         who submitted the guids (unless settings.DEBUG is True when the check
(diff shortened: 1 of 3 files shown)


src/olympia/blocklist/admin.py
```diff
@@ -387,15 +400,16 @@ def render_change_form(self, request, context, add=False, change=False,
 
     def save_model(self, request, obj, form, change):
         if change and self.is_pending_signoff(obj):
-            is_signoff = '_signoff' in request.POST
+            is_approve = '_signoff' in request.POST
```
bobsilverberg: You're calling this `is_approve` because that is a clearer name, so should the `POST` param also be renamed from `_signoff` to `_approve`? I realize that might make for a bigger patch, but would it be worth doing in a follow-up?

src/olympia/blocklist/tests/test_admin.py
```diff
@@ -1060,7 +1060,58 @@ def test_cannot_approve_with_only_block_create_permission(self):
         assert mbs.reason != 'a reason'
 
     def test_cannot_reject_with_only_block_create_permission(self):
```
bobsilverberg: Should the test be renamed to indicate that it is also testing that a user with only `Blocklist:Create` *can* reject their own submission? If that is too much of a mouthful, should it be made into two different tests?

src/olympia/blocklist/templates/blocklist/block_submission_change_form.html
```diff
@@ -10,8 +10,10 @@
     {% if can_change_object and change %}
         <input type="submit" value="{% trans 'Update' %}" class="default" name="_save">
     {% endif %}
-    {% if has_signoff_permission and is_pending_signoff %}
+    {% if has_signoff_reject_permission and is_pending_signoff %}
```
bobsilverberg: Should there be a test for this change? I don't see one.

src/olympia/blocklist/admin.py
eviljeff: I was just being lazy https://github.com/mozilla/addons-server/pull/13622/commits/91e55bf467f9aa70128590fd5f99c65dd28f6f15

src/olympia/blocklist/tests/test_admin.py
eviljeff: https://github.com/mozilla/addons-server/pull/13622/commits/91e55bf467f9aa70128590fd5f99c65dd28f6f15

src/olympia/blocklist/templates/blocklist/block_submission_change_form.html
eviljeff: https://github.com/mozilla/addons-server/pull/13622/commits/fa46856b102546c8410258626ec4fa193d1ae974

Issue #13383, opened 2020-01-30
Track progress of a Scanner Query Rule


diox:
We'd like to be able to track the progress of a scanner query rule in the change page / change list page once it's triggered. Also being able to abort the remaining tasks themselves would be great.

What is needed to test:
```python
sc = ScannerQueryRule(scanner=YARA, state=SCHEDULED, name='my_fancy_yara_test', definition='rule my_fancy_yara_test\r\n{\r\n  // This is a stub definition that will always match.\r\n  // See: https://yara.readthedocs.io/en/latest/\r\n  condition: true\r\n')
```

Then trigger the chord (`run_yara_query_rule` is a task itself but we're bypassing that for simplicity):
`run_yara_query_rule(sc.pk)`

If you have enough versions locally, the chord will execute a bunch of tasks (chunks of 250) and then a callback. What we'd like to track is the completion rate of the tasks. Celery groups provide `completed_count()` on their `ResultSet`s, and we could store the initial number of tasks, but I've ran into the following issues:
1. `chord()` doesn't expose the inner group results
2. Even without that, I can't get it to work by running just the group alone (see how `run_yara_query_rule()` is built internally)

diox:
Also: to simulate the task taking a long time, I've tried shutting down the workers, then firing the task, but then I can't find the task(s) when running `rabbitmqctl list_queues` for some reason.

diox:
Pinging @EnTeQuAk for input.

diox:
(FWIW https://gist.github.com/charlax/5205649 sounded promising but it's old and broken in several ways)

diox:
It deserves its own follow-up issue, but adding a comment here because it may have an impact on how it's implemented: eventually we'd also like admins to be able to abort a query, revoking the whole group of tasks that are left to execute at the time the query is aborted.

AlexandraMoga:
I've created and run some scanner query rules on -dev and tracked the Completion rate while the rule was running.
- the progress is expressed in percentages and will be updated when the page is refreshed. 
- you can check the progress both on the query list and detail view. 
![image](https://user-images.githubusercontent.com/31961530/75789250-12b92580-5d72-11ea-97ff-7685f7717b29.png)

- If a run is Aborted, the Completion rate goes back to None. 

@diox when a run is Completed, the Completion rate goes back to ` - ` (list view) or  `None` (detail view). Shouldn't we show a 100%  instead?
[...]

PR #13623, opened 2020-03-02
Track progress of a scanner query rule

diox:
Fixes https://github.com/mozilla/addons-server/issues/13383


src/olympia/scanners/models.py
```diff
@@ -306,6 +306,8 @@ class ScannerQueryRule(AbstractScannerRule):
     run_on_disabled_addons = models.BooleanField(
         default=False, help_text=_('Run this rule on add-ons that have been '
                                    'force-disabled as well.'))
+    group_result_id = models.UUIDField(default=None, null=True)
```
EnTeQuAk: I feel like this either deserves a comment or should be named `celery_group_result_id` to make it clearer what this is about.

diox: b45b8c39af10c3b07d54b85ac82cccbfb249b534

Issue #13416, opened 2020-02-05
Ask add-on source code when responding yes to obfuscated code


lafonth:
When a dev is submitting an add on and is responding yes to the question of the obfuscated code, the form is proposing to submit source code. 
Obfuscated code is totally forbidden by our Add-on policy today.

saw with @kewisch

eviljeff:
@jvillalobos / @wagnerand what should the copy be revised to?

wagnerand:
Turns out, there are several strings that would need to be updated:

```
src/olympia/devhub/templates/devhub/addons/submit/done.html
53:           "if your submission includes minified, obfuscated or compiled code.") }}

src/olympia/devhub/templates/devhub/includes/source_form_field.html
3:    {{ _('If your add-on contains binary or obfuscated code other than known '

src/olympia/devhub/templates/devhub/versions/edit.html
126:              {{ tip(None, _("If your add-on contain binary or obfuscated code, make the source available here for reviewers.")) }}

static/js/common/upload-addon.js
474:                            messages.push(gettext("Compiled binaries, as well as minified or obfuscated scripts (excluding known libraries) need to have their sources submitted separately for review. Make sure that you use the source code upload field to avoid having your submission rejected."));
```

I think we should generally refer to it as "minified, concatenated or otherwise machine-generated code". @kewisch what do you think?

jvillalobos:
The proposed change look good to me.

ManishAradwad:
Would like to work on this one...
I'm supposed to change only the prompt being displayed, right??

varunbankar:
Would like to work on this
So as far I understand the word "obfuscated" needs to be changed to "minified, concatenated or otherwise machine-generated code"?

caitmuenster:
Mentor: @bobsilverberg 

@ManishAradwad, it looks like you commented on this issue first. Do you still want to work on this issue?

ManishAradwad:
Sure, I'd like to give it a try...

AlexandraMoga:
I've checked the pages where the new terminology was updated and I could find two of them:

1. In the last submission step:
![image](https://user-images.githubusercontent.com/31961530/76202771-0dd6f480-61fe-11ea-879f-0f0d1e74b598.png)

2. On the manage version page (Source code field tooltip)
![image](https://user-images.githubusercontent.com/31961530/76202794-18918980-61fe-11ea-92d0-0d080335fd30.png)

[...]

PR #13631, opened 2020-03-03
Changed 'obfucated' to 'minified, concatenated or otherwise machine-generated'

ManishAradwad:
Fixes : https://github.com/mozilla/addons-server/issues/13416
Description: Changed all occurences of 'obfuscated' to 'minified, concatenated or otherwise machine-generated'

### DIFF
--- a/src/olympia/devhub/templates/devhub/includes/source_form_field.html
+++ b/src/olympia/devhub/templates/devhub/includes/source_form_field.html
@@ -1,6 +1,6 @@
 <div class="binary-source{% if not field.errors %} edit_initially_hidden{% endif %}">
   <p>
-    {{ _('If your add-on contains binary, minified, concatenated or otherwise machine-generated code other than known '
+    {{ _('If your add-on contains minified, concatenated or otherwise machine-generated code other than known '
          'libraries, upload its sources for review.') }}<br />
     {{ _('Please include instructions on how to reproduce the final add-on file, '
          'either in a README file or the Notes to Reviewer for this version.') }}<br />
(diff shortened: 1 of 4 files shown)


src/olympia/devhub/templates/devhub/versions/edit.html
```diff
@@ -123,7 +123,7 @@ <h3>{{ title }}</h3>
           {% with field = version_form.source %}
             <th>
               <label for="{{ field.auto_id }}">{{ _("Source code") }}</label>
-              {{ tip(None, _("If your add-on contain binary or obfuscated code, make the source available here for reviewers.")) }}
+              {{ tip(None, _("If your add-on contain minified, concatenated or otherwise machine-generated code, make the source available here for reviewers.")) }}
```
bobsilverberg: While we're here, we might as well fix this to say "If your add-on contains", rather than "If your add-on contain".

src/olympia/devhub/templates/devhub/includes/source_form_field.html
```diff
@@ -1,6 +1,6 @@
 <div class="binary-source{% if not field.errors %} edit_initially_hidden{% endif %}">
   <p>
-    {{ _('If your add-on contains binary or obfuscated code other than known '
+    {{ _('If your add-on contains minified, concatenated or otherwise machine-generated code other than known '
```
bobsilverberg: I think we should keep "binary" as I believe we are only replacing instances of "obfuscated", so this should be "If your add-on contains binary, minified, concatenated or otherwise machine-generated code...".

eviljeff: @jvillalobos for copy confirmation

jvillalobos: I think it's okay without "binary", since that's something only legacy extensions could use. WebExtensions can talk to binaries, but not ones contained inside the XPI.

bobsilverberg: Okay, thanks for the clarification.

@ManishAradwad please change it back to what you had before.

ManishAradwad:
Sure @bobsilverberg, I just made the changes. Can you plz check it
Thanks for help :smile:

PR #13658, opened 2020-03-06
Remove unnecessary version properties from APIs used by code-manager

### LINKED ISSUE #13651: Code manager APIs contain full license and release notes text
The version browse and compare API contain license information and release notes. In some cases it can be quite large, take the following request:
https://addons.mozilla.org/api/v5/reviewers/addon/920304/versions/4957090/?lang=en-US
Size: 90k
Size gzipped: 20k

Without license information:
Size: 50k
Size gzipped: 6k

Does the code manager need that info ? If we were to drop both both release notes and license information, not only would the payload be reduced in size, but we would also avoid a database query or two (There might be more things we could look into dropping, like compatibility info, which doesn't increase payload but does add database queries)

diox:
Fixes #13651

### DIFF
--- a/docs/topics/api/reviewers.rst
+++ b/docs/topics/api/reviewers.rst
@@ -156,7 +156,7 @@ This endpoint allows you to browse through the contents of an Add-on version.
 
 .. http:get:: /api/v4/reviewers/addon/(int:addon_id)/versions/(int:version_id)/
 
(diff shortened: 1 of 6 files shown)


diox:
Ah I was wondering about that. We can always change it back if needed later, but right now I'll take the perf win since we don't need it.

PR #13661, opened 2020-03-06
Add information about whether a file is minified to the reviewers API.

### LINKED ISSUE #13589: Add information about whether a file is minified to the reviewers API
When retrieving information about a file via the [Browse](https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#browse) and [Compare](https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#compare) APIs, the consumer should be able to tell whether the file requested is minified, based on the determination from the linter.

### What happened?
Information about whether the file is minified is not included in the response.

Information about whether the file is minified should be included in the response.

This is a request for additional data in the API response. 
This blocks https://github.com/mozilla/addons-code-manager/issues/1462.

EnTeQuAk:
Fixes #13589

A small implementation detail that I noticed while implementing this was that it's unclear to me how we're going to handle this information in the diff view.

What I did now is, that if there's information about a minified file in either the current version or the parent one (comparing against) we set the `uses_unknown_minified_code` flag. I don't think it makes sense to expand this into separate properties for the current and the parent version.

Does that make sense @bobsilverberg?


src/olympia/reviewers/tests/test_views.py
```diff
@@ -6991,8 +6991,8 @@ def test_compare_basic(self):
             'version_pk': self.version.pk,
             'pk': new_version.pk})
 
-        with self.assertNumQueries(12):
-            # FIXME: check queries and see if there is a need to optimize, 12
+        with self.assertNumQueries(14):
+            # FIXME: check queries and see if there is a need to optimize, 14
```
EnTeQuAk: I'm simply increasing this here, as investigating all the potential ways to optimize this is far beyond this issue.

diox: I think we should make sure we're not regressing, though. Can you add the right `select_related` to prevent the number of queries from increasing ?

EnTeQuAk: No, I can't easily in this case. I need to load the file-validation which is attached via `version.files` which - unless I'm missing something - needs a `prefetch_related` so the best I can think of is something along the lines of

```diff
diff --git a/src/olympia/reviewers/serializers.py b/src/olympia/reviewers/serializers.py
index 382b869008..aa9c1b15c9 100644
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -407,8 +407,13 @@ class FileEntriesDiffSerializer(FileEntriesSerializer):
         parent = self.context['parent_version']
         selected_file = self.get_selected_file(obj)

-        validation_objects = FileValidation.objects.filter(file_id__in=(
-            (obj.pk, parent.current_file.pk)))
+        validation_objects = []
+
+        for fobj in (obj, parent.current_file):
+            try:
+                validation_objects.append(fobj.validation)
+            except FileValidation.DoesNotExist:
+                continue

         for validation in validation_objects:
             data = json.loads(validation.validation)
diff --git a/src/olympia/reviewers/tests/test_views.py b/src/olympia/reviewers/tests/test_views.py
index b2a77fb570..be7b63e592 100644
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -6991,8 +6991,8 @@ class TestReviewAddonVersionCompareViewSet(
             'version_pk': self.version.pk,
             'pk': new_version.pk})

-        with self.assertNumQueries(14):
-            # FIXME: check queries and see if there is a need to optimize, 14
+        with self.assertNumQueries(12):
+            # FIXME: check queries and see if there is a need to optimize, 12
             # seems too high.
             response = self.client.get(self.url + '?file=README.md&lang=en-US')
         assert response.status_code == 200
diff --git a/src/olympia/reviewers/views.py b/src/olympia/reviewers/views.py
index 1a7ec74d2e..deb81db804 100644
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -1504,6 +1504,9 @@ class ReviewAddonVersionDraftCommentViewSet(
 class ReviewAddonVersionCompareViewSet(ReviewAddonVersionMixin,
                                        RetrieveModelMixin, GenericViewSet):

+    def filter_queryset(self, qset):
+        return qset.prefetch_related('files__validation')
+
     def retrieve(self, request, *args, **kwargs):
         serializer = AddonCompareVersionSerializer(
             instance=self.get_object(),
```

to ensure the query count is predictable but it doesn't reduce it. I'll try to spend some more time to try to avoid a regression.

diox: If it stays at 12 it's fine. We'll look at how to reduce it separately, I'm just concerned about not making things worse for now.

diox: Once https://github.com/mozilla/addons-server/pull/13726 is merged the number of queries should stay at 10 even when using the file validation info for the second file in the compare view.

src/olympia/reviewers/serializers.py
```diff
@@ -213,6 +216,20 @@ def get_content(self, obj):
...
+        try:
+            validation = obj.validation
+        except FileValidation.DoesNotExist:
+            # We don't have any idea about whether it could be minified or not
+            # so let's assume it's not for now.
+            return False
+
+        validation_data = json.loads(validation.validation)
+
+        prop = 'unknownMinifiedFiles'
```
willdurand: Can't this be inlined instead of having a variable used one time?

willdurand:
I didn't r+ yet because I asked a question in the related issue (this is good to go otherwise)

src/olympia/reviewers/serializers.py
EnTeQuAk: I found it much easier to read that way. In-lining it would mean breaking the following statement through multiple lines and that will make it much harder to read imo.

PR #13680, opened 2020-03-09
Ignore whitespace changes in AddonGitRepository diff handling

### LINKED ISSUE #13662: EOL conversion should be ignored when comparing two versions
This compare url https://code.addons.mozilla.org/en-US/compare/920304/versions/4957090...4976235/ shows that all the codebase has been changed. This is because the dev uploaded a new version with converted EOL.

### What happened?
One example of this can be found at https://code.addons.mozilla.org/en-US/compare/920304/versions/4957090...4976235/?path=app%2Feventpages%2Fblock.html

In this file, only lines number 24 and 25 actually have real code differences. The rest is because of "0D 0A" to "0A" EOL convertions.

image

image

To have a result similar to the old review tool : https://reviewers.addons.mozilla.org/en-US/firefox/files/compare/3520605...3501462/file/app/eventpages/block.html#top

diox:
(For parity with old file viewer)

Fixes #13662

### DIFF
--- a/src/olympia/lib/tests/test_git.py
+++ b/src/olympia/lib/tests/test_git.py
@@ -861,6 +861,7 @@ def test_get_diff_newline_new_file():
     # The file has been modified, so as far as git is concerned there should
     # be one change.
     assert len(changes) == 1
+    assert changes[0]['mode'] == 'M'
     assert changes[0]['new_ending_new_line'] is True
     assert changes[0]['old_ending_new_line'] is True
 
@@ -903,6 +904,7 @@ def test_get_diff_eol_changes():
     # The file has been modified, so as far as git is concerned there should
     # be one change.
     assert len(changes) == 1
+    assert changes[0]['mode'] == 'M'
 
     # We are ignoring all whitespace modifications so there is no diff.
     assert not changes[0]['hunks']
--- a/src/olympia/lib/tests/test_git.py
+++ b/src/olympia/lib/tests/test_git.py
@@ -860,30 +860,45 @@ def test_get_diff_newline_new_file():
 
     assert len(changes) == 1
     assert changes[0]['new_ending_new_line'] is True
-    assert changes[0]['old_ending_new_line'] is False
+    assert changes[0]['old_ending_new_line'] is True
 
-    hunk_changes = changes[0]['hunks'][0]['changes']
+    assert len(changes[0]['hunks']) == 0
 
-    assert hunk_changes == [
-        {
-            'content': '{"id": "random"}',
-            'new_line_number': -1,
-            'old_line_number': 1,
-            'type': 'delete'
-        },
-        {
-            'content': '\n\\ No newline at end of file',
-            'new_line_number': -1,
-            'old_line_number': 1,
-            'type': 'insert-eofnl'
-        },
-        {
-            'content': '{"id": "random"}',
-            'new_line_number': 1,
-            'old_line_number': -1,
-            'type': 'insert'
-        }
-    ]
+
+@pytest.mark.django_db
+def test_get_diff_eol_changes():
+    addon = addon_factory(file_kw={'filename': 'webextension_no_id.xpi'})
+
+    original_version = addon.current_version
+
+    AddonGitRepository.extract_and_commit_from_version(original_version)
+
+    parent_version = version_factory(
+        addon=addon, file_kw={'filename': 'webextension_no_id.xpi'})
+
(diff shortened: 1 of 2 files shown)


src/olympia/lib/tests/test_git.py
```diff
@@ -860,30 +860,45 @@ def test_get_diff_newline_new_file():
...
+    apply_changes(
+        repo, version,
+        '{"id": "random",\n"name": "foo"\n}', 'manifest.json')
+
+    changes = repo.get_diff(
+        commit=version.git_hash,
+        parent=parent_version.git_hash)
+
+    assert len(changes) == 1
+    assert len(changes[0]['hunks']) == 0
```
diox: Does that make sense, @EnTeQuAk ? Not sure why I still get a change with no hunks.

src/olympia/lib/tests/test_git.py
```diff
@@ -860,30 +860,45 @@ def test_get_diff_newline_new_file():
...
-        }
-    ]
+
+@pytest.mark.django_db
+def test_get_diff_eol_changes():
+    addon = addon_factory(file_kw={'filename': 'webextension_no_id.xpi'})
+
+    original_version = addon.current_version
+
+    AddonGitRepository.extract_and_commit_from_version(original_version)
```
EnTeQuAk: I don't think you're making use of `original_version` since you're creating a new `parent_version` - both seem to be duplicated to me. Removing one in favor of the other will speed this test up.

src/olympia/lib/tests/test_git.py
EnTeQuAk: I think that's fine, do any other properties in the `changes` change? I wonder if this behaves differently if you're using `GIT_DIFF_IGNORE_WHITESPACE` rather than just `GIT_DIFF_IGNORE_WHITESPACE_CHANGE` (which then ignores whitespace changes at the end of line too)

diox:
```
> pprint(changes)                                                                                                                                                                
[{'hash': 'f28ce3675d7aa63548560f2883bbde2aac8c95f3',
  'hunks': [],
  'is_binary': False,
  'lines_added': 0,
  'lines_deleted': 0,
  'mode': 'M',
  'new_ending_new_line': True,
  'old_ending_new_line': True,
  'old_path': 'manifest.json',
  'parent': 'e6636d850a0b92367924745b029fad93725b4d98',
  'path': 'manifest.json',
  'size': 17}]
```

diox: It makes sense I think: the file *has* been modified, we just happen to ignore all those modifications when showing the diff, but it's still modified as far as git is concerned.

src/olympia/lib/tests/test_git.py
diox: This is copied from every other test :) It would speed things up yeah, I assume you were doing this to start with a fresh manifest you control entirely.

EnTeQuAk: yeah, I'm sure I had a good reason in the first place. Re-reading it, it doesn't make much sense. I can file an issue to remove those duplicates.

src/olympia/lib/tests/test_git.py
EnTeQuAk: > It makes sense I think: the file has been modified, we just happen to ignore all those modifications when showing the diff, but it's still modified as far as git is concerned.

I think that explains it pretty good. Mind putting that into a comment in the test for us in a few months time?

diox: It did, `size` reflects the actual size of the file contents.

src/olympia/lib/tests/test_git.py
```diff
@@ -858,32 +858,54 @@ def test_get_diff_newline_new_file():
...
+    changes = repo.get_diff(
+        commit=version.git_hash,
+        parent=parent_version.git_hash)
+
+    # The file has been modified, so as far as git is concerned there should
+    # be one change.
+    assert len(changes) == 1
+
+    # We are ignoring all whitespace modifications so there is no diff.
+    assert not changes[0]['hunks']
```
EnTeQuAk: To support that, maybe do an `assert` on `changes[0]['mode'] == 'M'` to show that we do have modifications from a git perspective but are not showing them because we ignore them.

diox: https://github.com/mozilla/addons-server/pull/13680/commits/def9761acc015f876265e205200c6801db7bf8a7

PR #13683, opened 2020-03-09
handle updating imported kinto blocks

### LINKED ISSUE #13598: Handle updating an existing block with a regex
When we import a block from the v2 blocklist that contains a regex in the guid field we expand the regex to a separate Block instance for each matching guid.  This is correct (and what we want).  What is currently undetermined is the behaviour if that block was to be updated or deleted - if changing a single block on AMO results in changes to a block on kinto that targetted multiple guids then we're making unexpected changes; if we create a new block (for updates) then the existing block remains.

Options:  (user == the admin user making blocklist changes)
 
1.  Don't make changes on kinto to blocks imported from a regex - add a warning message and expect the change to be manually made via kinto admin webui by the user.  Simplest to implement but places an extra burden on the user; leaves kinto admin webui as a dependency until v2 blocklist is decommissioned.  (If the user didn't make the change via kinto admin, or made a mistake while doing it, then the v2 blocklist wouldn't match what was on AMO and wouldn't be a subset of v3)

2. Attempt to replace the existing regex block on kinto with multiple single guid blocks.  More complex (so more edge cases and bugs) and could make adding the blocks take a lot more time; Expanding the blocks could significantly (and inadvertently) increase the size of the v2 blocklist - e.g. a deleting a block on AMO that was imported from a kinto block that was a regex covering 100 guids would mean deleting the existing kinto block and adding 98 new block records.

I don't think adjusting the existing regex automatically from AMO is a feasible option.

(1) would be the option to choose if we think editing/deleting existing blocks in the v2 blocklist will be a rare task; (2) would ease the administrative burden if it was a frequent task - but at the cost of complexity and bloating the v2 blocklist.

eviljeff:
fixes #13598

### DIFF
--- a/src/olympia/blocklist/models.py
+++ b/src/olympia/blocklist/models.py
@@ -171,7 +171,8 @@ class BlockSubmission(ModelBase):
         ACTION_DELETE: 'Delete',
     }
     FakeBlock = namedtuple(
-        'FakeBlock', ('id', 'guid', 'addon', 'min_version', 'max_version'))
+        'FakeBlock', ('id', 'guid', 'addon', 'min_version', 'max_version',
+                      'is_imported_from_kinto_regex'))
     FakeAddon = namedtuple('FakeAddon', ('guid', 'average_daily_users'))
 
     action = models.SmallIntegerField(
@@ -219,7 +220,7 @@ class BlockSubmission(ModelBase):
             # If we'd be returning too many Block objects, fake them with the
             # minimum needed to display the link to the Block change page.
             blocks = [
-                self.FakeBlock(block.id, block.guid, None, None, None)
+                self.FakeBlock(block.id, block.guid, None, None, None, None)
                 for block in blocks]
         return blocks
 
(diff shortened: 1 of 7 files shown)


src/olympia/blocklist/models.py
```diff
@@ -219,7 +220,7 @@ def get_blocks_submitted(self, load_full_objects_threshold=1_000_000_000):
             # If we'd be returning too many Block objects, fake them with the
             # minimum needed to display the link to the Block change page.
             blocks = [
-                self.FakeBlock(block.id, block.guid, None, None, None)
+                self.FakeBlock(block.id, block.guid, None, None, None, None)
```
diox: keyword arguments please :)

src/olympia/blocklist/models.py
```diff
@@ -344,7 +347,7 @@ def process_input_guids(cls, guids, v_min, v_max, load_full_objects=True,
                 blocks_to_update_dict.get(addon.guid, None) or (
                     Block(addon=addon) if load_full_objects else
                     cls.FakeBlock(
-                        None, addon.guid, addon, Block.MIN, Block.MAX)
+                        None, addon.guid, addon, Block.MIN, Block.MAX, False)
```
diox: keyword arguments here as well.

src/olympia/blocklist/models.py
```diff
@@ -311,9 +312,11 @@ def process_input_guids(cls, guids, v_min, v_max, load_full_objects=True,
             list(block_qs)
             if load_full_objects else
             [cls.FakeBlock(
-                id_, guid, addon_guid_dict.get(guid), min_version, max_version)
-             for id_, guid, min_version, max_version in block_qs.values_list(
-                'id', 'guid', 'min_version', 'max_version')])
+                id_, guid, addon_guid_dict.get(guid), min_version, max_version,
+                (kinto_id.startswith('*')))
```
diox: keyword arguments here as well. (the whole `existing_blocks = (...)` bit is hard to read, I don't think trying to squeeze all that in the same "line" helps...

eviljeff: refactored the entire method to separate the cases where we're using fake objects and full django model instances https://github.com/mozilla/addons-server/pull/13683/commits/e8b281a875fd11b96e533c8933fb63b73405aef6

Issue #13697, opened 2020-03-10
Add 'scanners' key in the MAD request body


willdurand:
The API contract has slightly changed so we need to update our request to the MAD server from:

```json
{"customs": {}}
```

to:

```json
{"scanners": {"customs": {} }}
```

PR #13698, opened 2020-03-10
Prefix mad request body with 'scanners'

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13697


src/olympia/scanners/tests/test_tasks.py
```diff
@@ -592,7 +592,8 @@ def test_run_on_chunk_fallback_path(self):
         # We avoid triggering the on_change callback that would move the file
         # when the status is updated by doing an update() on the queryset.
         File.objects.filter(pk=self.version.all_files[0].pk).update(
-            status=amo.STATUS_DISABLED)
+            status=amo.STATUS_DISABLED
+        )
```
willdurand: I applied black.. 🙈

eviljeff: 🙄 not even
```
File.objects.filter(pk=self.version.all_files[0].pk).update(
    status=amo.STATUS_DISABLED,
)
```

Issue #13695, opened 2020-03-10
Pass timestamp along with the generated filter cascade


Rob--W:
The blocklist should include a timestamp, so clients can determine whether the filter cascade can be relied upon for testing the blocklist state.

This timestamp should be synchronized with clock of the signing server, so that we do not inadvertently block add-ons that shouldn't be blocked, or allow add-ons that should have been blocked.

eviljeff:
in https://github.com/mozilla/addons-server/pull/13699 I've added the `last_sign_time` property, though currently just the dumb current time on AMO.  The value is the timestamp in milliseconds (i.e. milliseconds since the unix epoch), like kinto uses for it's `modified` record property.  Will that work as a value for your implementation @Rob--W ?

I'll replace the dumb current time with the last signing time from autograph to close this issue.

Rob--W:
Thanks, that will do for now. It will help with testing the client implementation.

I do suggest to add a `# TODO` with a reference to this issue.

The implementation should be corrected when we ship though.
Otherwise the following can happen:

- When AMO's clock is behind Autograph's clock by X seconds, an add-on's signature may be later than `last_sign_time` (when the add-on is signed X seconds before the blocklist generation). Such add-ons cannot reliably be blocked.
- When AMO's clock is ahead of the Autograph's clock by X seconds, then add-ons that were signed X seconds after the generation of the blocklist have a chance of being blocked while they should not be.

Ideally X should be zero (i.e. clocks of Autograph and AMO are perfectly in sync).

eviljeff:
so as far I can see there are two routes we can do down here:
A) accurately keep track of the signing timestamps per certificate by extracting them after being returned from autograph and storing them in a field in files table (File model)
B) just return current time - 1 minute* to account for possible differences in system datetime between autograph and AMO instances plus the lag between the certificate being signed by autograph and the signed xpi being received and processed by AMO.

Given Blocks need to manually created (and possibly then signed off) it's very unlikely any addon/version/file would be blocked within a minute* of it being signed, so I'd recommend B.

* even 5 minutes.

@muffinresearch @diox @jvillalobos

jvehent:
So

- Clocks should be in near perfect sync (to the point where deltas are irrelevant).
- Certificates do not reflect the signing time. Their validity typically span over 90 days.

[...]

PR #13699, opened 2020-03-10
add support for signing mlbf via autograph

### LINKED ISSUE #13617: sign MLBF kinto attachments
for MLBF attachments (#13616) to be trusted we need to sign them and include the signature in the kinto record submission.

eviljeff:
fixes #13617 ... assuming my implementation is correct.  

(Also contains a minimal implementation for #13695 that should be close enough for now)

### DIFF
--- a/scripts/autograph_travis_test_config.yaml
+++ b/scripts/autograph_travis_test_config.yaml
@@ -240,11 +240,25 @@ signers:
           PIDNiTxNecePOmrD+1ivAEXcoL+e1w==
           -----END PRIVATE KEY-----
 
+    - id: addon-blocklist_content-signature_mozilla_org
+      type: contentsignature
+      privatekey: |
+          -----BEGIN EC PARAMETERS-----
+          BgUrgQQAIg==
+          -----END EC PARAMETERS-----
+          -----BEGIN EC PRIVATE KEY-----
+          MIGkAgEBBDART/nn3fKlhyENdc2u3klbvRJ5+odP0kWzt9p+v5hDyggbtVA4M1Mb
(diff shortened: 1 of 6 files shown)


scripts/autograph_travis_test_config.yaml
```diff
@@ -240,11 +240,25 @@ signers:
           PIDNiTxNecePOmrD+1ivAEXcoL+e1w==
           -----END PRIVATE KEY-----
 
+    - id: addon-blocklist_content-signature_mozilla_org
```
eviljeff: copied from https://github.com/mozilla-services/autograph/blob/master/autograph.yaml#L43 tbh

src/olympia/blocklist/cron.py
```diff
@@ -20,12 +22,22 @@ def upload_mlbf_to_kinto():
     server = KintoServer(KINTO_BUCKET, KINTO_COLLECTION_MLBF)
     stats = {}
     key_format = get_mlbf_key_format()
+    signing_datetime = int(time.time() * 1000)
```
EnTeQuAk: You need to explain to an old man what this actually doing, I'm not sure I've seen time.time() being multiplied by `1000` before :thinking:

eviljeff: heh, it's just to get time in milliseconds, but as a integer.  Kinto uses the same timestamp format for the record `modified` property.

eviljeff:
On the back of a recent meeting the current thinking is that signing the bloomfilter is ineffective against the kinds of attacks we'd want to prevent so closing this for now 😞

PR #13704, opened 2020-03-11
support kinto collections with no dual sign off too

### LINKED ISSUE #13701: submit kinto records as `to-sign` rather than `to-review`
according to https://bugzilla.mozilla.org/show_bug.cgi?id=1619994#c6 when there isn't dual sign-off needed we need to set a different state to allow the /stage/ bucket records to signed by kinto and propagated to the live /blocklist/ bucket

eviljeff:
fixes #13701


src/olympia/blocklist/cron.py
```diff
@@ -17,7 +17,8 @@ def upload_mlbf_to_kinto():
         log.info('Upload MLBF to kinto cron job disabled.')
         return
     log.info('Starting Upload MLBF to kinto cron job.')
-    server = KintoServer(KINTO_BUCKET, KINTO_COLLECTION_MLBF)
+    server = KintoServer(
+        KINTO_BUCKET, KINTO_COLLECTION_MLBF, kinto_sign_off_needed=False)
```
willdurand: nit: since the class is named `KintoServer`, does the argument name need the `kinto_` prefix?

eviljeff: There's AMO sign off too so I felt it more explicit to name this arg `kinto_`

PR #13716, opened 2020-03-12
use file hash in mlbf key; drop salt

### LINKED ISSUE #13714: update bloomfilter key generation to use xpi file hash
> if you include a hash of the signed add-on in the murmer computation, you should be okay as long as you're producing a 256-bit random value per signature computation. To clarify, would this now constitute your murmer computation: murmerHash3(ID + version + sha_256(file+sig_on_file)) ?

eviljeff:
fixes #13714

### DIFF
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -1,5 +1,4 @@
 import math
-import secrets
 
 from filtercascade import FilterCascade
 
@@ -11,6 +10,7 @@ log = olympia.core.logger.getLogger('z.amo.blocklist')
 def get_blocked_guids():
     from olympia.addons.models import Addon
     from olympia.blocklist.models import Block
+    from olympia.files.models import File
 
     blocks = Block.objects.all()
     blocks_guids = [block.guid for block in blocks]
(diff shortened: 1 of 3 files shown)


src/olympia/blocklist/mlbf.py
```diff
@@ -25,28 +25,33 @@ def get_blocked_guids():
...
-    return Version.unfiltered.values_list('addon__guid', 'version')
+    return File.objects.values_list(
+        'version__addon__guid', 'version__version', 'hash')
 
 
 def hash_filter_inputs(input_list, key_format):
     return [
-        key_format.format(guid=guid, version=version)
-        for (guid, version) in input_list]
+        key_format.format(guid=guid, version=version, xpi_hash=hash)
```
willdurand: Given this line, I'd expect `hash` to be `xpi_hash` everywhere in this patch. That would make it slightly more future-proof as the `key_format` expects `xpi_hash` so potentially a further iteration might introduce `hash` a placeholder (maybe not but we never know and that would still be more consistent)

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -41,65 +41,70 @@ def setUp(self):
...
         assert len(all_guids) == File.objects.count() == 10 + 8
-        assert (self.three_ver.guid, '123.40') in all_guids
-        assert (self.three_ver.guid, '123.5') in all_guids
-        assert (self.three_ver.guid, '123.45.1') in all_guids
-        over_tuple = (self.over.guid, self.over.addon.current_version.version)
-        under_tuple = (
-            self.under.guid, self.under.addon.current_version.version)
-        assert over_tuple in all_guids
-        assert under_tuple in all_guids
+        assert (self.three_ver.guid, '123.40', '') in all_guids
```
willdurand: Why is there no xpi hash in these `assert`?

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -41,65 +41,70 @@ def setUp(self):
...
-        assert (self.three_ver.guid, '123.5') in blocked_guids
-        assert (self.three_ver.guid, '123.45.1') not in blocked_guids
-        over_tuple = (self.over.guid, self.over.addon.current_version.version)
-        under_tuple = (
-            self.under.guid, self.under.addon.current_version.version)
-        assert over_tuple not in blocked_guids
-        assert under_tuple not in blocked_guids
+        assert (self.three_ver.guid, '123.40', '') in blocked_guids
+        assert (self.three_ver.guid, '123.5', '') in blocked_guids
+        assert (self.three_ver.guid, '123.45.1', '') not in blocked_guids
```
willdurand: Same question here.

src/olympia/blocklist/mlbf.py
eviljeff: The property is `File.hash` and, imo, just `hash` is ambiguous for Firefox so I went with `xpi_hash`.  So the mismatch is going to happen somewhere.

src/olympia/blocklist/tests/test_mlbf.py
eviljeff: hash wasn't set when the addon/version was created so it's the default (`''`).

src/olympia/blocklist/tests/test_mlbf.py
eviljeff: hash wasn't set when the addon/version was created so it's the default (`''`).

src/olympia/blocklist/mlbf.py
willdurand: I see. Would it be worth changing here, though?

src/olympia/blocklist/tests/test_mlbf.py
willdurand: Can we set more realistic values? Or do we have non-default values in other test cases? (I quickly looked but could not find anything, maybe I missed something)

src/olympia/blocklist/mlbf.py
eviljeff: which change are you suggesting exactly?

willdurand: Sorry, renaming `hash` to `xpi_hash` in this file (even if the `File` property is named `hash`).

diox:
Will likely have to be dropped as including the file hash is easy for us but not for Firefox. We are looking into going back to the salt approach but with a different hash.

eviljeff:
closing this for now as it's not a good change on the Firefox side as @diox said

PR #13720, opened 2020-03-12
Allow access to review page to most reviewers while restricting actions

### LINKED ISSUE #13566: Reviewer cannot access review tools page for add-on without public or pending versions

An add-ons reviewer (with permissions `Addons:PostReview`) cannot access review pages for add-ons where all files are disabled.

### What happened?

"You're not allowed to view this page"

They should be allowed to view the page and also do write operations like sending a reviewer reply or commenting.

diox:
This kinda reverses the previous behavior, allowing users with any kind of reviewer permission to see listed review pages, but restrict the actions depending on the permissions they have.

Unlisted continue to be restricted to unlisted reviewers.

Fixes #13566 (and more)

### DIFF
--- a/src/olympia/access/acl.py
+++ b/src/olympia/access/acl.py
@@ -151,10 +151,11 @@ def is_reviewer(request, addon):
     return check_addons_reviewer(request)
 
 
-def is_user_any_kind_of_reviewer(user):
+def is_user_any_kind_of_reviewer(user, allow_viewers=False):
     """More lax version of is_reviewer: does not check what kind of reviewer
     the user is, and accepts unlisted reviewers, post reviewers, content
-    reviewers, or people with just revierwer tools view access.
+    reviewers. If allow_viewers is passed and truthy, also allows users with
+    just reviewer tools view access.
 
     Don't use on anything that would alter add-on data.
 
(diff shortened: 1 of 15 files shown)


src/olympia/reviewers/views.py
```diff
@@ -607,10 +607,9 @@ def queue_moderated(request):
 
 
 @any_reviewer_required
-@post_required
```
diox: This is to make it work with `ReviewerTools:View`, aligning it with the rest of the views. It's a fairly minor thing, I almost dropped it (it's buried in the search form, when you add an app as a criteria)

src/olympia/access/acl.py
```diff
@@ -151,10 +151,11 @@ def is_reviewer(request, addon):
     return check_addons_reviewer(request)
 
 
-def is_user_any_kind_of_reviewer(user):
+def is_user_any_kind_of_reviewer(user, allow_viewers=False):
```
willdurand: So how is this supposed to work? I don't see this argument referenced in the body of the function.

willdurand: Shouldn't we have new test cases for this change as well?

diox: Yeah I missed a bit. I'll implement that change and test for it.

diox: https://github.com/mozilla/addons-server/pull/13720/commits/a36a64bdac981e4da8a8c17ab043c49ce906aaac

Issue #13194, opened 2020-01-07
Disable files when we disable an add-on


wagnerand:
Follow-up of #13185.

There are cases when add-ons (and all their files) need to get disabled. One example is when blocking add-ons, but there have been cases of add-ons having the get disabled without blocking them.

Please add a "Disable all" button to the page that changes the add-on status and the file status of all files to `STATUS_DISABLED`.

---

**Update (for QA):** we changed the behavior of the "force disable" button to disable the add-ons as well as the files (see https://github.com/mozilla/addons-server/issues/13194#issuecomment-599491696 for more details)

eviljeff:
@diox seemed to think this functionality belongs more in the reviewer tools, like the other admin disable action.

wagnerand:
We can do that too, as long as we gate it by some admin permission (tbd).

diox:
I don't mind tbh ; I was just noting that we already have a "Force disable" action, that only disables the add-on (not the files) and it lives in reviewer tools.

The nice thing about having it in the django admin it that it would be under VPN.

willdurand:
> I don't mind tbh ; I was just noting that we already have a "Force disable" action, that only disables the add-on (not the files) and it lives in reviewer tools.

So if the "force disable" does not disable the files, how should be name the button to disable the add-on + its files then? I am thinking about adding it to the django admin, next to the "force disable" one. WDYT?

/cc @wagnerand

diox:
So, for posterity:
- We already move the files belonging to a disabled add-on to the guarded location on filesystem, and check the add-on state in all relevant code (devhub, update service, download views)
- The reason Admins need the files to be disabled in database is for consistency, that helps with raw SQL queries.
- @wagnerand confirmed that it was ok to lose the original `status` when doing this. It makes reversing a completely manual operation that has to be performed using the django admin.
- We already have a `<Addon>.force_disable()` method so we can add `File.objects.filter(version__addon=self).update(status=amo.STATUS_DISABLED)` inside. As mentioned above we already move the files on filesystem, through a `on_change` signal handler.

willdurand:
I updated the issue to reflect what we are going to do (explained in the comment above)

ioanarusiczki:
@willdurand @wagnerand 
I force-disabled an add-on from the review page https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/anonymox
[...]

PR #13739, opened 2020-03-16
Add Addon.disable_all_files() and use it in force_disable()

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13194

---

This patch makes sure we disable the files when we "force disable" an add-on.

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -506,6 +506,8 @@ class Addon(OnChangeMixin, ModelBase):
                  self.slug, amo.STATUS_DISABLED)
         self.update(status=amo.STATUS_DISABLED)
         self.update_version()
+        # See: https://github.com/mozilla/addons-server/issues/13194
+        self.disable_files()
 
     def force_enable(self):
         activity.log_create(amo.LOG.CHANGE_STATUS, self, amo.STATUS_APPROVED)
@@ -531,6 +533,10 @@ class Addon(OnChangeMixin, ModelBase):
         activity.log_create(amo.LOG.DENIED_GUID_DELETED, self)
         log.info('Allow resubmission for addon "%s"', self.slug)
         DeniedGuid.objects.filter(guid=self.guid).delete()
+
+    def disable_files(self):
+        File.objects.filter(version__addon=self).update(
+            status=amo.STATUS_DISABLED)
 
     @property
     def is_guid_denied(self):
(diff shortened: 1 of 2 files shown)


src/olympia/addons/models.py
```diff
@@ -506,6 +506,8 @@ def force_disable(self):
                  self.slug, amo.STATUS_DISABLED)
         self.update(status=amo.STATUS_DISABLED)
         self.update_version()
+        # See: https://github.com/mozilla/addons-server/issues/13194
+        self.disable_files()
 
     def force_enable(self):
         activity.log_create(amo.LOG.CHANGE_STATUS, self, amo.STATUS_APPROVED)
```
wagnerand: Would it be safer to set the add-on status to `STATUS_INCOMPLETE` here?

eviljeff: if it's `STATUS_INCOMPLETE` then we're not actually disabling the add-on, just disabling all the versions and leaving the developer free to upload more.

src/olympia/addons/models.py
```diff
@@ -532,6 +534,10 @@ def allow_resubmission(self):
         log.info('Allow resubmission for addon "%s"', self.slug)
         DeniedGuid.objects.filter(guid=self.guid).delete()
 
+    def disable_files(self):
```
eviljeff: `disable_all_files`?

src/olympia/addons/models.py
diox: This is a comment on the `force_enable` method :)

src/olympia/addons/models.py
```diff
@@ -506,6 +506,8 @@ def force_disable(self):
                  self.slug, amo.STATUS_DISABLED)
         self.update(status=amo.STATUS_DISABLED)
         self.update_version()
+        # See: https://github.com/mozilla/addons-server/issues/13194
+        self.disable_files()
 
     def force_enable(self):
```
eviljeff: I guess everyone is aware that `force_disable` won't be a reversible action any longer as `force_enable` won't re-enable the files?  (Really this change should have failed an integration test somewhere)

src/olympia/addons/models.py
willdurand: Let's file a new issue if this is a problem

eviljeff: I didn't see these replies before I submitted my --edit :)

src/olympia/addons/models.py
willdurand: Yes, and that's why I added a comment with the link to the issue (it is described there)

src/olympia/addons/models.py
wagnerand: Done: https://github.com/mozilla/addons-server/issues/13746

Issue #13199, opened 2020-01-08
Add link to review page on django add-on admin page


wagnerand:
Admins quite often need to go to the review page from a django admin add-on page they reached using a certain use case we have. Could we please add a link to the review page (listed is fine for now) to `admin/models/addons/addon/<ID/change`?

gabbyprecious:
Is anyone working on this, can I jump on it?

diox:
You can, if you're familiar with how the django admin works. Our install docs are over there: https://addons-server.readthedocs.io/en/latest/topics/install/index.html

I was thinking the link would be at the top of the add-on admin change page, next to the "History" and "View on site" links. For this, you need to have a [custom `change_form.html` template](https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#overriding-vs-replacing-an-admin-template) for the `addon` model inheriting from the django one, and add the link as a `<li>` in the `object-tools-items` block. The link will probably be something like `{% url reviewers.review object_id %}`.

willdurand:
@wagnerand you wrote "listed is fine for now"–would it make sense to have 2 links instead? Maybe we could be smart and only show 2 links if there are both listed and unlisted versions (if possible)

wagnerand:
@willdurand Yes, it would be great to show 2 links *if* there are both listed and unlisted add-ons. We should only show links to channels that we have versions for.

If we cannot determine that, it might make sense to show 2 links by default. Just keep in mind that reviewers would probably open the two links all the time, since it's cheaper to just CMD+Click on 2 links and the close one, than to click on one, realize there are no versions for that channel, go back and click the other link.

willdurand:
@willdurand Yes, it would be great to show 2 links if there are both listed and unlisted add-ons. We should only show links to channels that we have versions for.

Cool!

> On the other hand, the review pag contains a link to the other channel if it exists, so a reviewer might not even go back and click the other link. Maybe you could reuse the code from the reviewer page for determining whether there are listed/unlisted versions?

Good idea, thanks for the tip!

diox:
FYI there are `has_unlisted_versions() and `has_listed_versions()` helpers on the `Addon` instances. Do pass `include_deleted=True` argument in both cases.

ioanarusiczki:
@willdurand I can navigate from en-US/admin/models/addons/addon/1002608/change/ to both: listed and unlisted review pages.

I filed an issue here https://github.com/mozilla/addons-server/issues/13778

[...]

PR #13741, opened 2020-03-16
Add links to reviewer tools in addon admin

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13199

---

This patch adds new links in the admin, for example:

Screen Shot 2020-03-17 at 16 55 18

It adds:

- a "listed" link when the add-on has at least one listed version
- an "unlisted" link when the add-on has at least one unlisted version

### DIFF
--- a/src/olympia/addons/admin.py
+++ b/src/olympia/addons/admin.py
@@ -194,22 +194,26 @@ class AddonAdmin(admin.ModelAdmin):
                     url += '?' + request.GET.urlencode()
                 return http.HttpResponsePermanentRedirect(url)
 
-        response = super(AddonAdmin, self).change_view(
-            request, object_id, form_url, extra_context=extra_context
+        return super().change_view(request, object_id, form_url,
+                                   extra_context=extra_context)
+
+    def render_change_form(self, request, context, add=False, change=False,
+                           form_url='', obj=None):
+        context.update(
+            {
+                'has_listed_versions': obj.has_listed_versions(
+                    include_deleted=True
+                ),
+                'has_unlisted_versions': obj.has_unlisted_versions(
+                    include_deleted=True
+                )
+            }
         )
-        if response.status_code == 200:
-            obj = response.context_data.get('original', None)
-            response.context_data.update(
-                {
-                    'has_listed_versions': obj.has_listed_versions(
-                        include_deleted=True
-                    ) if obj else False,
-                    'has_unlisted_versions': obj.has_unlisted_versions(
-                        include_deleted=True
-                    ) if obj else False,
-                }
-            )
-        return response
+
+        return super().render_change_form(request=request, context=context,
+                                          add=add, change=change,
+                                          form_url=form_url, obj=obj)
+
 
     def save_model(self, request, obj, form, change):
         super().save_model(request, obj, form, change)
--- a/src/olympia/addons/admin.py
+++ b/src/olympia/addons/admin.py
@@ -194,9 +194,22 @@ class AddonAdmin(admin.ModelAdmin):
                     url += '?' + request.GET.urlencode()
                 return http.HttpResponsePermanentRedirect(url)
 
-        return super(AddonAdmin, self).change_view(
-            request, object_id, form_url, extra_context=None,
+        response = super(AddonAdmin, self).change_view(
+            request, object_id, form_url, extra_context=extra_context
         )
+        if request.method == 'GET':
+            obj = response.context_data.get('original', None)
+            response.context_data.update(
+                {
+                    'has_listed_versions': obj.has_listed_versions(
+                        include_deleted=True
+                    ) if obj else False,
+                    'has_unlisted_versions': obj.has_unlisted_versions(
+                        include_deleted=True
+                    ) if obj else False,
+                }
+            )
+        return response
 
     def save_model(self, request, obj, form, change):
         super().save_model(request, obj, form, change)
--- a/src/olympia/addons/admin.py
+++ b/src/olympia/addons/admin.py
@@ -197,7 +197,7 @@ class AddonAdmin(admin.ModelAdmin):
         response = super(AddonAdmin, self).change_view(
             request, object_id, form_url, extra_context=extra_context
         )
-        if request.method == 'GET':
+        if response.status_code == 200:
             obj = response.context_data.get('original', None)
             response.context_data.update(
                 {
--- a/src/olympia/addons/admin.py
+++ b/src/olympia/addons/admin.py
@@ -214,7 +214,6 @@ class AddonAdmin(admin.ModelAdmin):
                                           add=add, change=change,
                                           form_url=form_url, obj=obj)
 
-
     def save_model(self, request, obj, form, change):
         super().save_model(request, obj, form, change)
         if 'status' in form.changed_data:
--- a/src/olympia/addons/tests/test_admin.py
+++ b/src/olympia/addons/tests/test_admin.py
@@ -149,9 +149,8 @@ class TestAddonAdmin(TestCase):
         self.grant_permission(user, 'Addons:Edit')
         self.client.login(email=user.email)
         response = self.client.get(detail_url, follow=True)
-        assert 'Reviewer Tools (listed)' in response.content.decode('utf-8')
-        assert ('Reviewer Tools (unlisted)' not in
-                response.content.decode('utf-8'))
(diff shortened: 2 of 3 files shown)


willdurand:
@diox this patch increases the number of SQL queries from 23 to 34 😨 I think that's because of how `has_listed_versions()`/`has_unlisted_versions()` work. Would you have an idea to mitigate this problem?

I tried to prefect related versions but it did not change anything.

diox:
`has_listed_versions()` / `has_unlisted_versions()` in themselves will add 2 queries which will be difficult to avoid (it's just a `.exists()` that returns `True` or `False` - behind the scenes it does a `SELECT 1 WHERE ...` which is impossible to prefetch, we're not dealing with objects here, nor can we just add a join in the main query).

9 extra sounds bad :) I would be this is because you're calling `get_object()`, and that in turn causes you to load the `Addon` and its transformers (which we should deactivate for this page, but that's a story for another time). Can you avoid this call and try to get the instance after the django admin has already fetched it ? Maybe it's in the context somewhere if you call the `super()` implementation first ?

diox:
FWIW the instance is probably available in `adminForm.form.instance` in the context after `super().change_view()` has ran (`instance` could be `None` for additions, which we don't support, but maybe write this defensively just in case)

willdurand:
@diox I see but once `change_view()` is called, we get the http response so isn't it too late?

willdurand:
Also, note that I want to inject these new values to the context because it is not possible to call methods with arguments in Django templates. The alternative would be to create new `@property` methods in the `Addon` model (to enforce `include_disabled=True`)

diox:
All django admin code should return [`TemplateResponse`](https://docs.djangoproject.com/en/2.2/ref/template-response/) objects ; those are not rendered immediately, you can alter the context after it's been initialized, the rendering happens later, so that would work. You would need to avoid doing this for `POST` requests, though (since those would, if the form is valid, return a `HttpResponseRedirect` eventually)

willdurand:
> All django admin code should return [`TemplateResponse`](https://docs.djangoproject.com/en/2.2/ref/template-response/) objects ; those are not rendered immediately, you can alter the context after it's been initialized, the rendering happens later, so that would work. You would need to avoid doing this for `POST` requests, though (since those would, if the form is valid, return a `HttpResponseRedirect` eventually)

Ah, great. Thanks!

willdurand:
@diox thanks, that works well now :)

src/olympia/addons/admin.py
```diff
@@ -194,9 +194,22 @@ def change_view(self, request, object_id, form_url='', extra_context=None):
                     url += '?' + request.GET.urlencode()
                 return http.HttpResponsePermanentRedirect(url)
 
-        return super(AddonAdmin, self).change_view(
-            request, object_id, form_url, extra_context=None,
+        response = super(AddonAdmin, self).change_view(
```
eviljeff: nit: we don't need the class name in `super().` in modern python versions, so it's okay to strip it out when touching code that does it the old way.

src/olympia/addons/admin.py
```diff
@@ -194,9 +194,22 @@ def change_view(self, request, object_id, form_url='', extra_context=None):
...
+            request, object_id, form_url, extra_context=extra_context
         )
+        if request.method == 'GET':
```
diox: I know I'm the one who suggested it, but maybe there is a better way to do this. You would need to double-check, but maybe you can just use `response.status_code` instead ? It's a minor detail, but it would mean the links wouldn't suddenly disappear when you're doing a `POST` but there is an error to be shown so the change form is displayed back to the user.

eviljeff: mmm, tbh I don't like tinkering with `TemplateResponse`.

If you override `render_change_form` instead we have the `obj` reference passed in as a parameter and you can add it to `context`

src/olympia/addons/tests/test_admin.py
```diff
@@ -409,15 +448,15 @@ def test_query_count(self):
...
+        with self.assertNumQueries(24):
             # It's very high because most of AddonAdmin is unoptimized but we
             # don't want it unexpectedly increasing.
             response = self.client.get(self.detail_url, follow=True)
         assert response.status_code == 200
         assert addon.guid in response.content.decode('utf-8')
 
         version_factory(addon=addon)
-        with self.assertNumQueries(23):
+        with self.assertNumQueries(24):
```
eviljeff: this would be 1 higher if you added a version of a different channel but I guess that's still okay

src/olympia/addons/tests/test_admin.py
```diff
@@ -140,6 +140,45 @@ def test_can_edit_with_addons_edit_permission(self):
...
+        version_factory(addon=addon, channel=amo.RELEASE_CHANNEL_LISTED)
+        detail_url = reverse('admin:addons_addon_change', args=(addon.pk,))
+        user = user_factory()
+        self.grant_permission(user, 'Admin:Tools')
+        self.grant_permission(user, 'Addons:Edit')
+        self.client.login(email=user.email)
+        response = self.client.get(detail_url, follow=True)
+        assert 'Reviewer Tools (listed)' in response.content.decode('utf-8')
+        assert ('Reviewer Tools (unlisted)' not in
+                response.content.decode('utf-8'))
```
eviljeff: you can either `content = response.content.decode('utf-8')` once or test with `b'...'` strings so you don't have to `.decode(...)` each time.

src/olympia/addons/tests/test_admin.py
```diff
@@ -140,6 +140,45 @@ def test_can_edit_with_addons_edit_permission(self):
...
+        addon = addon_factory(guid='@foo', version_kw=version_kw)
+        detail_url = reverse('admin:addons_addon_change', args=(addon.pk,))
+        user = user_factory()
+        self.grant_permission(user, 'Admin:Tools')
+        self.grant_permission(user, 'Addons:Edit')
+        self.client.login(email=user.email)
+        response = self.client.get(detail_url, follow=True)
+        assert ('Reviewer Tools (listed)' not in
+                response.content.decode('utf-8'))
+        assert 'Reviewer Tools (unlisted)' in response.content.decode('utf-8')
```
eviljeff: same thing with `.decode(...)`

src/olympia/addons/tests/test_admin.py
```diff
@@ -140,6 +140,45 @@ def test_can_edit_with_addons_edit_permission(self):
...
+        version_factory(addon=addon, channel=amo.RELEASE_CHANNEL_LISTED)
+        version_factory(addon=addon, channel=amo.RELEASE_CHANNEL_UNLISTED)
+        detail_url = reverse('admin:addons_addon_change', args=(addon.pk,))
+        user = user_factory()
+        self.grant_permission(user, 'Admin:Tools')
+        self.grant_permission(user, 'Addons:Edit')
+        self.client.login(email=user.email)
+        response = self.client.get(detail_url, follow=True)
+        assert 'Reviewer Tools (listed)' in response.content.decode('utf-8')
+        assert 'Reviewer Tools (unlisted)' in response.content.decode('utf-8')
```
eviljeff: same thing with `.decode(...)`

src/olympia/addons/admin.py
willdurand: ah, right..

willdurand: ah yes, that's much better, thanks!

src/olympia/addons/admin.py
```diff
@@ -194,10 +194,26 @@ def change_view(self, request, object_id, form_url='', extra_context=None):
...
+                           form_url='', obj=None):
+        context.update(
+            {
+                'has_listed_versions': obj.has_listed_versions(
+                    include_deleted=True
+                ),
+                'has_unlisted_versions': obj.has_unlisted_versions(
+                    include_deleted=True
+                )
+            }
```
eviljeff: thankfully we don't support creating new `Addon` instances via django admin because this would fail in that case 😬 (obj would be None)

willdurand: I saw the same thing in the blocklist admin so I thought it was okay :D

eviljeff: If it's https://github.com/mozilla/addons-server/blob/master/src/olympia/blocklist/admin.py#L397 note the `if change:` - `change=True` when it's not a new instance.

willdurand: ha! /me writes a followup patch

PR #13744, opened 2020-03-17
Can reject a previously approved theme

### LINKED ISSUE #13072: Cannot reject a previously approved theme

A previously approved theme version needs to be rejected (there is only one version in this case).

### What happened?

There is no open to reject.

A _Reject_ or _Reject multiple versions_ button

junngo:
Fixes #13072 
This patch has been modified to allow reject a previously approved theme.

**Before**
13072_before

**After**
13072_after

### DIFF
--- a/src/olympia/reviewers/tests/test_utils.py
+++ b/src/olympia/reviewers/tests/test_utils.py
@@ -379,7 +379,8 @@ class TestReviewHelper(TestReviewHelperBase):
         # Themes reviewers get access to everything.
         self.request.user.groupuser_set.all().delete()
         self.grant_permission(self.request.user, 'Addons:ThemeReview')
-        expected = ['public', 'reject', 'reply', 'super', 'comment']
+        expected = ['public', 'reject', 'reject_multiple_versions',
+                    'reply', 'super', 'comment']
         assert list(self.get_review_actions(
             addon_status=amo.STATUS_APPROVED,
             file_status=amo.STATUS_AWAITING_REVIEW).keys()) == expected
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -386,6 +386,7 @@ class ReviewHelper(object):
                 is_appropriate_reviewer or
                 (acl.action_allowed_user(
                     request.user, amo.permissions.ADDONS_POST_REVIEW) and
+                 self.addon.type != amo.ADDON_STATICTHEME and
                  not is_admin_needed)
             )
 
@@ -467,7 +468,6 @@ class ReviewHelper(object):
                          'versions. The comments will be sent to the '
                          'developer.'),
             'available': (
-                self.addon.type != amo.ADDON_STATICTHEME and
                 addon_is_valid_and_version_is_listed and
                 can_reject_multiple
             ),
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -4922,7 +4922,8 @@ class TestReview(ReviewBase):
         doc = pq(response.content)
 
         expected_actions_values = [
-            'public|', 'reject|', 'reply|', 'super|', 'comment|']
+            'public|', 'reject|', 'reject_multiple_versions|',
+            'reply|', 'super|', 'comment|']
         assert [
             act.attrib['data-value'] for act in
             doc('.data-toggle.review-actions-desc')] == expected_actions_values
@@ -4931,7 +4932,7 @@ class TestReview(ReviewBase):
 
         assert (
             doc('.data-toggle.review-comments')[0].attrib['data-value'] ==
-            'public|reject|reply|super|comment|')
+            'public|reject|reject_multiple_versions|reply|super|comment|')
         # we don't show files and tested with for any static theme actions
         assert (
             doc('.data-toggle.review-files')[0].attrib['data-value'] ==


src/olympia/reviewers/tests/test_utils.py
```diff
@@ -371,15 +371,16 @@ def test_actions_public_static_theme(self):
         # action, but it should not be available for static themes.
         self.grant_permission(self.request.user, 'Addons:PostReview')
         self.addon.update(type=amo.ADDON_STATICTHEME)
-        expected = []
+        expected = ['reject_multiple_versions']
```
junngo: I made PR by referring to the guide.
According to the test code, It cannot be used `reject_multiple_versions` if `PostReview` and `static themes`.
Should I make an exception to limit using `reject_multiple_versions` in this case?

src/olympia/reviewers/utils.py
```diff
@@ -467,7 +467,6 @@ def get_actions(self, request):
                          'versions. The comments will be sent to the '
                          'developer.'),
             'available': (
-                self.addon.type != amo.ADDON_STATICTHEME and
```
diox: Rejecting multiple versions kinda doesn't make sense for a theme, since they should only have a single version. Could you make it work with the `reject` action instead ?

diox:
`Addons:PostReview` is for reviewers to confirm approval (or reject) previously auto-approved add-ons.
`Addons:ThemeReview` is specific to themes, which are manually approved or rejected.

src/olympia/reviewers/utils.py
junngo: Sure. I'll modify it and submit it.

eviljeff: fwiw, themes _can_ have multiple versions, but it's rare in reality.

diox: Ah, in that case, maybe we should remove that restriction and let the `reject_multiple_versions` be available for theme reviewers as well.

src/olympia/reviewers/tests/test_utils.py
diox: You should need `Addons:ThemeReview` to deal with static themes in general. Someone with just `Addons:PostReview` shouldn't be able to reject themes.

src/olympia/reviewers/utils.py
junngo: Yes, I understand.

junngo:
Only those who have `Addons:PostReview` authority should not be able to use `reject_multiple_versions`.
In this case, even those who have the `Admins` authority cannot use `reject_multiple_versions`.
So I'll revise it again and talk when it's finished.

junngo:
I submitted again.
When using `acl` function with admin authority, the results are always true.
I had to compare directly to find a `PostReview` authority.
Let me know if there's a better way.

diox:
Sorry, this is the wrong approach. Checking for `PostReview` to disallow access is the wrong approach here. The fact that you have to manually check for this permission is a good hint that something is wrong with the logic.

First, you shouldn't be adding new conditions on top of the existing ones. What you want to do is remove the check on `self.addon.type != amo.ADDON_STATICTHEME` like you're doing, and then modify `can_reject_multiple` variable that's already there. You can see it's defined like this:
```python
can_reject_multiple = (
                is_appropriate_reviewer or
                (acl.action_allowed_user(
                    request.user, amo.permissions.ADDONS_POST_REVIEW) and
                 not is_admin_needed)
            )
```

So the only thing needed is to make sure the `acl.action_allowed_user(user, amo.permissions.ADDONS_POST_REVIEW)` condition is only executed for non-static themes. If you do that, it should just work - the `is_appropriate_reviewer` should take care of the rest.

junngo:
Thanks for reviewing.
Oh, I could think a little more. I was short of thought.
I'll write it again and submit it.

junngo:
It cannot use `reject multiple versions` when It is `PostReview`.
I wrote this code(`not is_post_review`) to deal with the following cases.

https://github.com/mozilla/addons-server/blob/1dc785be605be62dd6ec9635d12428a70b633f8d/src/olympia/reviewers/tests/test_utils.py#L369-L374

src/olympia/reviewers/utils.py
```diff
@@ -382,10 +382,12 @@ def get_actions(self, request):
...
             # been auto-approved, provided that the add-on isn't marked as
             # needing admin review.
+            is_post_review = acl.action_allowed_user(
+                request.user, amo.permissions.ADDONS_POST_REVIEW)
             can_reject_multiple = (
                 is_appropriate_reviewer or
-                (acl.action_allowed_user(
-                    request.user, amo.permissions.ADDONS_POST_REVIEW) and
+                ((is_post_review if self.addon.type != amo.ADDON_STATICTHEME
+                    else not is_post_review) and
```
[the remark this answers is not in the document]
diox: Something like this probably works, please test (I removed `is_post_review` entirely):

```suggestion
                (acl.action_allowed_user(
                    request.user, amo.permissions.ADDONS_POST_REVIEW) and
                 self.addon.type != amo.ADDON_STATICTHEME and
```

junngo:
Thank you for reviewing it :)
I was too focused on the test case only.
I cleaned up all the unpretty code and rewritten it to solve the problem.

diox:
Hey sorry, this looks good now, but I ended up having to touch this action for a different issue so I'll roll a different version of this fix with my own changes (https://github.com/mozilla/addons-server/pull/13805)

PR #13749, opened 2020-03-17
expose the last version time before the filter was generated

### LINKED ISSUE #13695: Pass timestamp along with the generated filter cascade

eviljeff:
fixes #13695 

@Rob--W the field is called `last_version_time` now because its not strictly connected to signing - though obviously a version can't be signed before it's been created.  (alternate names welcome)

### DIFF
--- a/src/olympia/blocklist/cron.py
+++ b/src/olympia/blocklist/cron.py
@@ -1,5 +1,6 @@
 import json
 import tempfile
+import time
 
 import waffle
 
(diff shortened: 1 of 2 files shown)


Rob--W:
`version` + `time` looks a bit stange. How about any of those:

- `generation_time`
- `creation_time`
- `last_modified_time`

eviljeff:
> `version` + `time` looks a bit stange. How about any of those:
> 
>     * `generation_time`

https://github.com/mozilla/addons-server/pull/13749/commits/8e246af2b9d8d4bbcb4bbc3f84db2292def4e324

src/olympia/blocklist/cron.py
```diff
@@ -21,12 +22,15 @@ def upload_mlbf_to_kinto():
         KINTO_BUCKET, KINTO_COLLECTION_MLBF, kinto_sign_off_needed=False)
     stats = {}
     key_format = get_mlbf_key_format()
+    generation_time = int(time.time() * 1000)
```
Rob--W: Is this temporary (until you have a better way of finding the time), or permanent?

In any case, a comment about the significance of this time stamp would be nice. Even if it is just a reference to #13695.

eviljeff: The conclusion from #13695 was the time on AMO was for all practical purposes the same time as on autograph so this is the best way, unless I'm misunderstanding the other's comments.  If you still disagree I'd comment in the issue.  

I'll add a comment in the code.

eviljeff: https://github.com/mozilla/addons-server/pull/13749/commits/c8379bbe5dd733744597b37e1446b80f7dd1b6cd

PR #13761, opened 2020-03-18
Hide buttons if no permission in scanner query rule change list/form

### LINKED ISSUE #13461: The Run/Abort button is displayed to users with view only permissions in scanner query rules
Follow up for #13268 

1. Log in to AMO -dev with a user that has `Admin:ScannersQueryView` permissions
2. Open the Scanner query rule model 
3. Make sure that the list contains a New rule pending for Run (you will need to do that with a user that has edit permissions first)
4. Observe the options available in the State column

### What happened?
The Run button is displayed to your user even though you have only view permissions to this model. In consequence, if you try to Run the query you'll receive a 404. 
The Run/Abort button should be hidden to users with only `Admin:ScannersQueryView` permissions

image
image

diox:
Fixes #13461

### DIFF
--- a/static/css/admin/scannerrule.css
+++ b/static/css/admin/scannerrule.css
@@ -1,5 +1,10 @@
 /* note: this is shared between scannerrule and scannerqueryrule admins */
 
 .model-scannerqueryrule.hide-action-buttons .field-state_with_actions button {
+    /* 
+    Hide query rule action buttons if user has no permission (through class set
+    on body by the change view/list template). We use visibility: hidden so
+    that the view looks the same for people with or without the buttons
+    */
     visibility: hidden;
 }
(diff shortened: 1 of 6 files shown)


static/css/admin/scannerrule.css
```diff
@@ -0,0 +1,5 @@
+/* note: this is shared between scannerrule and scannerqueryrule admins */
+
+.model-scannerqueryrule.hide-action-buttons .field-state_with_actions button {
+    visibility: hidden;
```
willdurand: We might want to use `display: block` to save some space in the page.

diox: I specifically wanted to keep the same space, so that people like us who might get the view only permission in prod can compare what we see with admins that can edit.

willdurand: Oh ok, I see. Should we have a comment then?

diox: e4a14c83a5f8ec9c1e1801ac2ad1ea87099a2ba4

PR #13764, opened 2020-03-18
optimize Addon query in blocklist_import

eviljeff:


src/olympia/blocklist/tests/test_commands.py
```diff
@@ -57,8 +57,8 @@ def test_regex(self):
         assert Block.objects.count() == 3
         blocks = list(Block.objects.all())
         this_block = blocklist_json['data'][0]
-        assert blocks[0].guid == '_qdNembers_@exmys.myysarch.com'
-        assert blocks[1].guid == '_dqMNemberstst_@www.dowespedtgttest.com'
+        assert blocks[0].guid == '_dqMNemberstst_@www.dowespedtgttest.com'
+        assert blocks[1].guid == '_qdNembers_@exmys.myysarch.com'
```
diox: What made the order change ? Could we make it more stable to be sure ? (Adding a `order_by('pk')` or something)

eviljeff: I did, but removed it again - changing the queries to keep the tests the same seemed the wrong way around.

It seems it was sorting by pk before, but now is sorting by guid.  Either is fine really, and they're both consistent with the same data.

diox: There is no default ordering on `Addon` (which surprised me) so it's going to be whatever MySQL chooses... It's fine to keep it like this as the task doesn't care, but if the test starts randomly failing you'll need to add an explicit one.

PR #13773, opened 2020-03-19
12842 drop is source public property

### LINKED ISSUE #12842: Drop view_source property from Addon model
follow up from #12514

hwj0418:
Fixes #12842


junngo:
There is a lint error.
Please check the error using the following command:
`docker-compose exec web make lint-codestyle`

eviljeff:
closing this.  This isn't the correct solution - it's the `Addon.view_source`  property that needs removing and all references to it.

PR #13786, opened 2020-03-19
Expose add-on name in blocklist/block api

### LINKED ISSUE #13757: Return the add-on name in the blocklist/block API response
We need the add-on `name` in the block API response so that the frontend can display it (see: https://github.com/mozilla/addons-frontend/issues/9216). Having the `name` in the response would avoid another API call (and because add-ons are blocked, it might not be easy to retrieve such information _via_ public APIs anyway).

@jvillalobos do we need any other information on the new (upcomping) block(ed) page?

eviljeff:
fixes #13757

### DIFF
--- a/src/olympia/api/fields.py
+++ b/src/olympia/api/fields.py
@@ -102,7 +102,10 @@ class TranslationSerializerField(fields.Field):
 
     def get_attribute(self, obj):
         source = self.source or self.field_name
-        field = fields.get_attribute(obj, source.split('.'))
+        try:
+            field = fields.get_attribute(obj, source.split('.'))
+        except AttributeError:
+            field = None
 
         if not field:
             return None
--- a/docs/topics/api/blocklist.rst
+++ b/docs/topics/api/blocklist.rst
@@ -20,10 +20,12 @@ This endpoint returns an add-on Block from the blocklist, specified by guid or i
 
 .. http:get:: /api/v4/blocklist/block/(int:block_id|string:guid)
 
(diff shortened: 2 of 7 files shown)


src/olympia/api/fields.py
```diff
@@ -102,7 +102,10 @@ def fetch_single_translation(self, obj, source, field, requested_language):
 
     def get_attribute(self, obj):
         source = self.source or self.field_name
-        field = fields.get_attribute(obj, source.split('.'))
+        try:
+            field = fields.get_attribute(obj, source.split('.'))
+        except AttributeError:
+            field = None
```
EnTeQuAk: I can't really grasp in which cases that happen, mind to explain a little? Ideally adding a test to `api/tests/test_fields.py` to make sure we don't just test this change via blocklist tests since it's a more generic API that can be used by other parts too.

eviljeff: if your source has nested attributes (e.g. `source='addon.current_version.release_notes'`) but one of the parent objects is `None` (e.g. `addon.current_version is None`) there's an exception thrown.   

We need it for this patch because there's an edge case where a Block exists but the Addon guid doesn't so a source of `addon.name` can raise - I guess it's never been an issue before because we generally include entire serializers for child objects which safely handle it.

eviljeff: https://github.com/mozilla/addons-server/pull/13786/commits/e104df29ae5b230b3db3e3af9f4c052d5d04f40c

PR #13791, opened 2020-03-19
Fixes #13508

### LINKED ISSUE #13508: Deleting a group fails

STR:
1. Create a user group
2. Add some users to it
3. Try to delete the group. 
### What happened?

```
'Cannot delete or update a parent row: a foreign key constraint fails (`addons_dev_allizom_org`.`log_activity_group`, CONSTRAINT `log_activity_group_group_id_e03ab8c8_fk_groups_id` FOREIGN KEY (`group_id`) REFERENCES `groups` (`id`))'
```

Group is successfully deleted.

https://sentry.prod.mozaws.net/operations/olympia-dev/issues/7296022/?query=is:unresolved

liangyo6:
Fixes #13508 

I have to execute a custom SQL in Django migration in order to make sure mysql includes on delete cascade condition in the foreign key constraint in table log_activity_group. Below are the steps I performed to the fix:

   1. Type python manage.py makemigrations activity --empty -n added_ondelete_cascade to create a custom empty migration named “0005_added_ondelete_cascade” under activity app.
   2. Added codes in "0005_added_ondelete_cascade"
   3. Type python manage.py migrate activity to run the migration.
   4. Done

The change has been successfully run locally and the group was successfully deleted.

Screen Shot 2020-03-19 at 19 12 08
Screen Shot 2020-03-19 at 19 12 16
Screen Shot 2020-03-19 at 19 12 24


eviljeff:
closing this pr.  As @diox comments, we want to understand why the existing code/db doesn't work for this group first, and only hack around it with raw sql as a last resort.

PR #13792, opened 2020-03-19
implemented issue #8262 as required

### LINKED ISSUE #8262: "More Actions" is too far down on the review page
The whiteboard fields are blank or only contain one line most of the time. Having to scroll down 800 pixels of whiteboards gets cumbersome if I want to be notified if a new version gets submitted:

Two improvements I thought of:

- Move the "More Actions" section above the whiteboard(s)
- If a whiteboard is empty (`:empty`), hide them (`display:none`), and only show the empty whiteboard if a button is pressed/clicked on the page.

AlanXu12:
Fixes #8262 

Here is the fix process and the files that are changed:

1. static/css/zamboni/reviewers.less
Added textarea:empty for default empty style

2. static/js/zamboni/reviewers.js
Added click listener for whiteboard and private whiteboard in order to have expanding functionalities for these two boards. Also added click listener for elements other than the two boards in order to collapse back.

Here is the fixed screenshots:

1. When whiteboards are empty, whiteboards are collapsed to height of one line:
Screen Shot 2020-03-19 at 7 20 28 PM

2. When an empty whiteboard is clicked, it will be expanded to the original height. If no contents are added and clicked on somewhere else, the whiteboard will be collapsed again:
Screen Shot 2020-03-19 at 7 20 41 PM

3. When whiteboards are with content, whiteboards are expanded to the original height:
Screen Shot 2020-03-19 at 7 21 17 PM

### DIFF
--- a/static/js/zamboni/reviewers.js
+++ b/static/js/zamboni/reviewers.js
@@ -58,6 +58,37 @@
         window.location.hash = "id=" + $("#addon, #persona").attr("data-id");
       });
     }
+
+    // Toggle expansion for whiteboard
+    $("#id_whiteboard-public").click(function(){
+        $("#id_whiteboard-public").css("height", "200px");
+    });
+
+    // Toggle expansion for whiteboard private
+    $("#id_whiteboard-private").click(function(){
+        $("#id_whiteboard-private").css("height", "200px");
+    });
+
+    // When clicking on areas other than whiteboard/whiteboard private on the page
+    $('html').click(function(e) {
+        // When whiteboard is clicked
+        if (!$(e.target).is("#id_whiteboard-public")) {
+            console.log("Non-id_whiteboard-public clicked");
+            // When whiteboard is still empty, collapse
+            if (!$("#id_whiteboard-public").val()) {
+                $("#id_whiteboard-public").css("height", "20px");
+            }
+        }
+        // When whiteboard private is clicked
+        if (!$(e.target).is("#id_whiteboard-private")) {
+            console.log("Non-id_whiteboard-private clicked");
+            // When whiteboard private is still empty, collapse
+            if (!$("#id_whiteboard-private").val()) {
+                $("#id_whiteboard-private").css("height", "20px");
+            }
+        }
+    });
+
 })();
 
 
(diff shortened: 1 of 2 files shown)


static/js/zamboni/reviewers.js
```diff
@@ -58,6 +58,37 @@
         window.location.hash = "id=" + $("#addon, #persona").attr("data-id");
       });
     }
+
+    // Toggle expansion for whiteboard
+    $("#id_whiteboard-public").click(function(){
+        $("#id_whiteboard-public").css("height", "200px");
+    });
```
eviljeff: Your JS looks likes it would work (not tried locally yet) but does just using the `:active` css rule work instead?  It seems like we only want the whiteboard collapsed when it's empty and not active.

AlanXu12: I tried` :focus` css rule as well. It would do the collapsing work. However, by using `:focus` css rule, the whiteboard may still remain collapsed when there are contents on it but it is not selected. I am not very sure if remaining collapsed when there are contents on the whiteboard is what we want. If so, I could change it to use `:focus` instead.

FYI, the current JS and CSS solution will perform as follow (When the whiteboard does not have content and is not selected, it will collapse. If the whiteboard does have contents and is not selected, it will remain expanded.):
more_action2

eviljeff: A colleague suggested making use of the `:placeholder-shown` css rule instead to detect when the textarea is empty.

e.g. 
```
+        textarea:placeholder-shown {
+            height: 20px;
+        }
```

you'd just have to add define a placeholder in the `WhiteboardForm.meta` class with something like:
```
+       widgets = {
+            'private': widgets.Textarea(attrs={'placeholder': _('some placeholder text here')}),
+            'public': widgets.Textarea(attrs={'placeholder': _('some placeholder text here')}),
+        }
```

We'd rather not add extra javascript to the reviewer pages if possible - there's no test framework set up for it on addons-server so it's quite fragile.

eviljeff:
closing this PR for now.  If you have time to finish it we can re-open.

PR #13805, opened 2020-03-23
Fix reject_multiple_versions action availability in reviewer tools

### LINKED ISSUE #12413: Volunteers should not be able to review recommended extensions
I've received an "Dear reviewer, You're looking great today!" email regarding a new upload awaiting review : https://addons.mozilla.org/reviewers/review/2464 version 7.0

This extension is currently part of the Recommended Extensions program. To my understanding, volunteers shouldn't have access to do such reviews.

### What happened?
In the review page, it seems I have access to all the different options to finish the review. "Approve", "Reject" and "Save" buttons are active. Note that I didn't try to actually save the review, for oblivious reasons.

Saving the review should be disabled or should show an error to the volunteer reviewer.

### LINKED ISSUE #13072: Cannot reject a previously approved theme

diox:
Fixes #12413
Fixes #13072

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -375,7 +375,9 @@ class ReviewHelper(object):
             request.user, permission)
 
         # Special logic for availability of reject multiple action:
-        if self.content_review:
+        if (self.content_review or
+                is_recommendable or
+                self.addon.type == amo.ADDON_STATICTHEME):
             can_reject_multiple = is_appropriate_reviewer
         else:
             # When doing a code review, this action is also available to
@@ -467,7 +469,6 @@ class ReviewHelper(object):
                          'versions. The comments will be sent to the '
                          'developer.'),
             'available': (
-                self.addon.type != amo.ADDON_STATICTHEME and
                 addon_is_valid_and_version_is_listed and
                 can_reject_multiple
             ),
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/utils.py
```diff
@@ -375,7 +375,9 @@ def get_actions(self, request):
             request.user, permission)
 
         # Special logic for availability of reject multiple action:
-        if self.content_review:
+        if (self.content_review or
+                is_recommendable or
+                self.addon.type == amo.ADDON_STATICTHEME):
```
eviljeff: I can see how the change fixes the linked issues but trying to read this function from beginning to end to work out what actions would be available makes my head hurt.  

Suggestion for follow-up: each action should be split into a class and we can move the logic around if its available into there; this function can be something like  `return [action.name: action for action in [ActionClass(...) for ActionClass in self.actions]]`

diox: I'm not sure that would be a good thing. The availability conditions for the various actions share a lot of code, and that's why they have been kept in the same method. When I refactored the permission checks I started by separating the checks, and it ended up making things a lot more complicated because I still wanted to avoid copy/pasting a bunch of code around.

PR #13809, opened 2020-03-24
Don't let content reviewers review recommendable add-ons, warn about them in review page

### LINKED ISSUE #13806: Restrict content review for Recommended Extensions
Currently, all content reviewers can content review Recommended extensions. Instead, we only want this for users who also have permission to review Recommended extensions.

So, if a user has content review permission but no recommended review permission and is looking at the content review page for a Recommended extension:

* The page should not allow content review actions.
* A message should be shown: `You don't have permission to content review Recommended extensions.`

diox:
Fixes #13806

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/review.html
+++ b/src/olympia/reviewers/templates/reviewers/review.html
@@ -83,6 +83,16 @@
     </div>
   </div>
 
+  {% if is_recommendable %}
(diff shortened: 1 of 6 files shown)


diox:
https://travis-ci.org/github/mozilla/addons-server/builds/666313201 is actually green, despite github showing the check as in progress.

PR #13810, opened 2020-03-24
Add regex parsing to import_blocklist

eviljeff:
fixes the `import_blocklist` command choking on complex regular expressions by expanding them to lists of guids instead.  Only works on well-formed lists-as-regexs that start with a `^`, end with a `$`, and the guids themselves are wrapped in `(` `)`.

### DIFF
--- a/src/olympia/blocklist/tasks.py
+++ b/src/olympia/blocklist/tasks.py
@@ -75,21 +75,20 @@ def import_block_from_blocklist(record):
         # need to escape the {} brackets or mysql chokes.
         guid_regexp = bracket_open_regex.sub(r'\{', guid[1:-1])
         guid_regexp = bracket_close_regex.sub(r'\}', guid_regexp)
-        log.debug(
-            'Kinto %s: Attempting to create Blocks for addons matching [%s]',
-            kinto_id, guid_regexp)
         # we're going to try to split the regex into a list for efficiency.
         guids_list = split_regex_to_list(guid_regexp)
         if guids_list:
             log.debug(
-                'Kinto %s: Broke down regex into list of guids [%s]',
+                'Kinto %s: Broke down regex into list; '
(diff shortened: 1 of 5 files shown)


src/olympia/blocklist/tasks.py
```diff
@@ -78,8 +78,20 @@ def import_block_from_blocklist(record):
         log.debug(
             'Kinto %s: Attempting to create Blocks for addons matching [%s]',
             kinto_id, guid_regexp)
-        addons_guids_qs = Addon.unfiltered.using(using_db).filter(
-            guid__regex=guid_regexp).values_list('guid', flat=True)
+        # we're going to try to split the regex into a list for efficiency.
+        guids_list = split_regex_to_list(guid_regexp)
+        if guids_list:
+            log.debug(
+                'Kinto %s: Broke down regex into list of guids [%s]',
```
diox: Don't need the surrounding `[]` since you're logging a list as a string.

src/olympia/blocklist/tasks.py
```diff
@@ -78,8 +78,20 @@ def import_block_from_blocklist(record):
...
+                kinto_id, guids_list)
+            addons_guids_qs = Addon.unfiltered.using(using_db).filter(
+                guid__in=guids_list).values_list('guid', flat=True)
+        else:
+            log.debug(
+                'Kinto %s: Unable to break down regex into list of guids',
+                kinto_id)
```
diox: I found it helped me to debug what was going on better to add the `guid` in that log, and remove it from the `Attempting to create Blocks` log above. Makes it consistent with the case  where we did manage to break down the regexp into a list, too (where we log the final list)

eviljeff: https://github.com/mozilla/addons-server/pull/13810/commits/4a915dd0ec8991fe488100d574a6e9b42d4b76e0

Issue #13361, opened 2020-01-27
Allow to exclude results with a given matched rule in the list of scanner results


willdurand:
Some rules match more often than others. We use the list of scanner results to triage these results. We'd like to exclude results that are associated with a given matched rule so that we can better see the results for the rules that don't match too often.

**For QA:** we added a new filter ("By all but this rule") that is used to hide the scanner results (not the query results) bound to the rule selected in this filter. By default no rule is excluded.

AlexandraMoga:
I've verified the new 'By all but this rule' filter and it seems to be working as expected.
@willdurand I have one observation regarding scanner results with multiple matched rules. If one of those matched rules is filtered out, the scanner result will be hidden from view entirely. I'm not sure if we can or need to address this use case.

willdurand:
> @willdurand I have one observation regarding scanner results with multiple matched rules. If one of those matched rules is filtered out, the scanner result will be hidden from view entirely. I'm not sure if we can or need to address this use case.

I think this is the expected behavior. I am not sure to know how we can do it otherwise. This filter isn't used by default so I guess that's fine for now. We'll see once @wagnerand starts using it.

PR #13811, opened 2020-03-24
Add exclude rule filter in scanner results admin

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13361

---

We decided to fix the issue above by excluding a single rule at a time (for now).

### DIFF
--- a/src/olympia/scanners/admin.py
+++ b/src/olympia/scanners/admin.py
@@ -116,6 +116,37 @@ class ScannerRuleListFilter(admin.RelatedOnlyFieldListFilter):
         ]
 
 
+class ExcludeMatchedRuleFilter(SimpleListFilter):
+    title = ugettext('all but this rule')
+    parameter_name = 'exclude_rule'
+
+    def lookups(self, request, model_admin):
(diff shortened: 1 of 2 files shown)


willdurand:
Also requesting @diox's review because we discussed this issue during today's meeting.

src/olympia/scanners/admin.py
```diff
@@ -116,6 +116,37 @@ def field_choices(self, field, request, model_admin):
         ]
 
 
+class ExcludeMatchedRuleFilter(SimpleListFilter):
```
diox: I would have expected you to inherit from `ScannerRuleListFilter` which takes care of the choices/lookups for you, you just have to customize the queryset. Did you look into that ? Does that not work ?

willdurand: I wasn't able to customize the default entry. With this patch, it looks like this (and that will, hopefully, avoid confusions):

Screen Shot 2020-03-24 at 17 16 43

diox: Ok

Issue #13332, opened 2020-01-22
Add a `score` field to `ScannerResult`


willdurand:
We discussed that with @diox in Slack. We'd like to extract the different `score` values returned by the ML API and store them in the scanner results.

We'll have an aggregate score that we'll set on the `ScannerResult` created by the `call_ml_api` task (see #13140).

We'll have one score per scanner result sent to the API and we'll set this score to the corresponding `ScannerResult` (customs results will always have a score, for other scanners we'll see).

This issue is related to #13141.

willdurand:
I requested a change in the API contract of the mad server (because this contract has changed). Once we have that, we'll be able to extract and store the score easily.

AlexandraMoga:
There is a new 'Score' column in scanner results which is populated with `-` (yara and wat) or `n/a` (customs). By reading the conversation from the pull request, I get it that these values might change in the future.

@willdurand Is there an interpretation to this score right now or was this issue just about adding the new column?

willdurand:
> @willdurand Is there an interpretation to this score right now or was this issue just about adding the new column?

There is nothing else for now *but* as soon as the MAD server works (and therefore returns valid results), the score of `customs` should be updated with a value that is not `n/a`.

AlexandraMoga:
I'm marking this as verified for now and will revisit once the MAD server works.

PR #13820, opened 2020-03-25
Add a `score` field to `ScannerResult` (and use it)

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13332

---

This patch adds a new `score` field to the `ScannerResult` model, populate it when possible (in the task) and render it in the admin.


src/olympia/scanners/admin.py
```diff
@@ -436,6 +440,15 @@ def formatted_matched_rules_with_files(
 
     formatted_matched_rules_with_files.short_description = 'Matched rules'
 
+    def formatted_score(self, obj):
+        if obj.scanner not in [CUSTOMS, MAD]:
+            return '-'
+        if obj.score < 0:
+            return 'n/a'
```
bobsilverberg: Why differentiate between `-` and `n/a`? Wouldn't `n/a` be appropriate for the case where `obj.scanner not in [CUSTOMS, MAD]`?

willdurand: So that's because some scanners won't have a score at all whereas CUSTOMS/MAD should (at least the ML API should return something, which could be `-1`)

bobsilverberg: I gathered that from the code, but that doesn't explain why we need both `-` and `n/a`. Will a user care about the difference between `-` and `n/a`? Is there a meaningful difference between them from a user perspective?

willdurand: Maybe not. To me the distinction makes sense because I can see which results are analyzed by the ML service and if I see `n/a` next to a `customs` result, I'll have to investigate. It would likely work for "-" too, I guess.

willdurand: I keep your comment in mind as I am sure we'll have to update how we display the score once the integration is enabled in production.

PR #13831, opened 2020-03-26
only include signed webext files in mlbf blocked list

### LINKED ISSUE #13821: only include signed/webext files when generating list of guid, version to block in mlbf
split from #13813 
"not every single add-on/version/file that is in the database is also worth blocking.

If the file was not signed, we don't need to block it. Likewise, if the file is not a webextension, we don't need to block it either because all currently supported types of add-ons are some kind of webextension."

eviljeff:
fixes #13821

### DIFF
--- a/src/olympia/blocklist/tests/test_commands.py
+++ b/src/olympia/blocklist/tests/test_commands.py
@@ -214,17 +214,20 @@ class TestExportBlocklist(TestCase):
             addon_factory()
         # one version, 0 - *
         Block.objects.create(
-            addon=addon_factory(),
+            addon=addon_factory(
+                file_kw={'is_signed': True, 'is_webextension': True}),
             updated_by=user_factory())
         # one version, 0 - 9999
         Block.objects.create(
-            addon=addon_factory(),
+            addon=addon_factory(
+                file_kw={'is_signed': True, 'is_webextension': True}),
             updated_by=user_factory(),
             max_version='9999')
         # one version, 0 - *, unlisted
         Block.objects.create(
             addon=addon_factory(
-                version_kw={'channel': amo.RELEASE_CHANNEL_UNLISTED}),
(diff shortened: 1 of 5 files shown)


src/olympia/blocklist/tests/test_commands.py
```diff
@@ -214,17 +214,20 @@ def test_command(self):
             addon_factory()
         # one version, 0 - *
```
bobsilverberg: Should there be tests added for unsigned and non-webextension add-ons? We seem to have tests for the types of add-ons that should be exported, but not for those that should no longer be exported. Or are those covered by `test_mlbf.py`?

src/olympia/blocklist/utils.py
```diff
@@ -17,8 +17,8 @@ def add_version_log_for_blocked_versions(obj, al):
     from olympia.activity.models import VersionLog
 
     VersionLog.objects.bulk_create([
-        VersionLog(activity_log=al, version_id=id_chan[0])
-        for version, id_chan in obj.addon_versions.items()
+        VersionLog(activity_log=al, version_id=id_)
+        for version, (id_, _) in obj.addon_versions.items()
```
bobsilverberg: I don't really understand what this change is about.

eviljeff: ah, that was a drive-by fix.  I originally looked at changing how `Block.addon_versions` was populated and was checking where it was used and decided expanding the tuples directly was better coding practise (bad past-me!).  I didn't end up changing `addon_versions` because it impacted too many other things but left this fix in.

src/olympia/blocklist/tests/test_commands.py
eviljeff: the tests in `test_mlbf.py` should cover it.  The setup here is just creating enough `Blocks` to generate the filter - it probably doesn't need as many as we have here at the moment.

PR #13832, opened 2020-03-26
import_blocklist should skip addons that dont have webext files

### LINKED ISSUE #13813: Block webextension files only
Before an add-on version is added to the blocklist, AMO should check whether any files belonging to that version are both webextensions and signed.

If so, add the version to the list.
If not, skip it.

eviljeff:
fixes #13813

### DIFF
--- a/src/olympia/blocklist/tasks.py
+++ b/src/olympia/blocklist/tasks.py
@@ -10,6 +10,7 @@ from olympia import amo
 from olympia.addons.models import Addon
 from olympia.amo.celery import task
 from olympia.amo.decorators import use_primary_db
+from olympia.files.models import File
 from olympia.users.utils import get_task_user
 
 from .models import Block, BlocklistSubmission, KintoImport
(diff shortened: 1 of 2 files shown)


diox:
Was about to approve but... Should we actually skip unsigned files ? If they ever somehow get signed (say they were only temporarily not auto-approved, for instance), then the block won't do anything for them, even though it will look as if we imported it correctly. That seems like a risk. What's the rationale behind doing this ?

jvillalobos:
I think @wagnerand looked at the add-ons involved and they weren't stuck in a temporary state, but rather were really old add-ons that we hadn't cleaned up from the blocklist.

eviljeff:
> Was about to approve but... Should we actually skip unsigned files ? If they ever somehow get signed (say they were only temporarily not auto-approved, for instance), then the block won't do anything for them, even though it will look as if we imported it correctly. That seems like a risk. 

If they've not been signed then they can't be installed by users; if the files _are_ subsequently signed then the date of the signing will be after the bloomfilter creation date so Firefox won't check the filter anyway.  (Unless I'm missing some known edge case where the files are signed by autograph and distributed by AMO but _aren't_ marked as `is_signed=True`.)  

> What's the rationale behind doing this ?

Filter size efficiency, I believe, but https://github.com/mozilla/addons-server/issues/13813#issuecomment-603489444 was the request

diox:
It totally makes sense to exclude them from the bloom filter, that I get. But at import time, because we will consider the block as imported, I am not sure it's worth it. I'd rather have the block fully imported regardless of file state, and then filter when generating the bloom filter cascade.

eviljeff:
ugh, I got this pr mixed up with https://github.com/mozilla/addons-server/pull/13831

So, to re-answer the question:  the issue came off the back of an observation by @wagnerand that we were importing old blocks relating to legacy add-ons, which was redundant because no-one who will be using the v3 blocklist could install/use them anyway.  After #13831 they won't be in the bloomfilter too so they're just there in the database, but effectively blocking nothing.   Would just focusing on `is_webextension` be enough?

diox:
> Would just focusing on `is_webextension` be enough?

Yeah I'd like that.

eviljeff:
> > Would just focusing on `is_webextension` be enough?
> 
> Yeah I'd like that.

https://github.com/mozilla/addons-server/pull/13832/commits/33b7b03adf2f3dd52c7d258e0d2eafa5204aa948

Issue #13833, opened 2020-03-26
GUID column should be sortable in the list of scanner results


willdurand:
The `guid` column should be sortable in the list of scanner results.

ioanarusiczki:
@willdurand I verified the scanner results on dev and stage and the add-on GUID filter is applied on the GUID column. - FF74(Win10).

PR #13834, opened 2020-03-26
Allow to sort guid column again

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13833


src/olympia/scanners/admin.py
```diff
@@ -377,6 +377,7 @@ def guid(self, obj):
         return '-'
 
     guid.short_description = 'Add-on GUID'
+    guid.admin_order_field = 'version__addon__guid'
```
eviljeff: Can haz test?  (assuming it's not horrendously complicated)

PR #13844, opened 2020-03-27
Split translations per locale in Elasticsearch instead of per analyzer

### LINKED ISSUE #13591: Split translations per locale in Elasticsearch instead of per analyzer
When storing translations in Elasticsearch, and then querying them, we store them per `analyzer`. We have different language analyzers enable, like one for english and one for french for instance, and so when an add-on has an `en-US` and a `fr` translation for its name, it's stored as `name_l10n_english` and `name_l10n_french`. We also have a field without analyzer for the translation in the default locale the developer chose for their add-on.

The problem with that approach is that it merges translations for all english-based locales we support into a single field: `en-US`, `en-CA` and `en-GB` all end up in that `english` field. That is causing various issues with relevancy.

We should have one field per locale instead.

The drawback is, if an add-on only has translations in `en-US` and `fr`, and the default locale is `fr`, if I look for it in `en-GB` I might not find it at all. Maybe there should be a special case for languages like English and Portuguese where we support different variants - we could simply store a duplicate value if there is nothing, or better yet have a fallback for those locales (could be done with a `DisMax` looking in multiple locales for those languages).

diox:
This avoids storing `['foo', 'foo', 'foo']` for `english` if the add-on has translations for `en-US`, `en-CA` and `en-GB`, and in turn prevents such cases from being considered more relevant than they really are.

To allow for users searching in a language that has multiple variants in a locale different from what the add-on has provided, we still search through all variants, but we only consider the best match.

This is backwards-incompatible in a way, but the missing data shouldn't generate errors and the we should be ok after a reindex.

Fixes #13591

### DIFF
--- a/src/olympia/amo/indexers.py
+++ b/src/olympia/amo/indexers.py
@@ -2,7 +2,7 @@ from django.conf import settings
 
 import olympia.core.logger
 
-from olympia.constants.search import SEARCH_ANALYZER_MAP
+from olympia.constants.search import SEARCH_LANGUAGE_TO_ANALYZER
 
 from .models import SearchMixin
 from .utils import to_language
@@ -100,9 +100,9 @@ class BaseSearchIndexer(object):
         """
         doc_name = cls.get_doctype_name()
 
-        for analyzer in SEARCH_ANALYZER_MAP:
+        for lang, analyzer in SEARCH_LANGUAGE_TO_ANALYZER.items():
             for field in field_names:
-                property_name = '%s_l10n_%s' % (field, analyzer)
+                property_name = '%s_l10n_%s' % (field, lang)
                 mapping[doc_name]['properties'][property_name] = {
                     'type': 'text',
                     'analyzer': analyzer,
@@ -117,9 +117,9 @@ class BaseSearchIndexer(object):
         """
         doc_name = cls.get_doctype_name()
 
-        for analyzer in SEARCH_ANALYZER_MAP:
+        for lang, analyzer in SEARCH_LANGUAGE_TO_ANALYZER.items():
             for field in field_names:
-                property_name = '%s_l10n_%s' % (field, analyzer)
+                property_name = '%s_l10n_%s' % (field, lang)
                 mapping[doc_name]['properties'][property_name] = {
                     'type': 'text',
                     'analyzer': analyzer,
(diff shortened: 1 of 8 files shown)


src/olympia/amo/tests/__init__.py
```diff
@@ -895,11 +895,6 @@ def setUpClass(cls):
         # right before each test.
         stop_es_mocks()
         cls.es = amo_search.get_es(timeout=settings.ES_TIMEOUT)
-        cls._SEARCH_ANALYZER_MAP = amo.SEARCH_ANALYZER_MAP
-        amo.SEARCH_ANALYZER_MAP = {
```
diox: This has been here forever and was actively harmful to relevancy score testing in other languages. In fact, besides the new tests added around `en-CA` and `en-GB`, this is the only reason why I've had to update the expected scores in `test_search_ranking` : otherwise, my changes wouldn't affect them at all (which makes sense because none of the ranking tests used languages that had the same analyzer)

src/olympia/amo/indexers.py
```diff
@@ -165,20 +165,21 @@ def extract_field_search_translation(cls, obj, field, default_locale):
...
     def extract_field_analyzed_translations(cls, obj, field, db_field=None):
         """
-        Returns a dict containing translations for each language-specific
-        analyzer for the given field. Empty translations are skipped entirely.
+        Returns a dict containing translations for each language that we have
+        an analyzer for, for the given field.
+
+        When no translation exist for a given language+field combo, the value
+        returned is an empty string, to avoid storing the word "None" as the
+        field does not understand null values.
```
eviljeff: is just not including the language+field combination in the dict not an option?

src/olympia/constants/search.py
```diff
@@ -59,31 +59,3 @@
 for analyzer, languages in SEARCH_ANALYZER_MAP.items():
     for language in languages:
         SEARCH_LANGUAGE_TO_ANALYZER[language] = analyzer
-
-# Which stemmer to use for each language.
-#
-# Note: We use the keys of this dict for supported stop words, also, which is
-# specified as, e.g., '_english_'.
-STEMMER_MAP = {
```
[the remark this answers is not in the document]
diox: It's a drive-by fix because I was looking at the constants, we aren't using that anywhere. Each analyzer already comes with its own stemmer config, there is no need to override this.

src/olympia/amo/indexers.py
diox: It's an option, but I went with what I thought was the simplest option that stayed close to the previous behavior (previously, we would store an empty list, because we potentially stored multiple values, now we store an empty string)

PR #13853, opened 2020-03-30
Redirect content reviewers when they try to access a code review page

### LINKED ISSUE #6936: Developer reply emails for content-reviewers link to technical-review page
This is a bit of a hard bug but maybe someone has an idea for an elegant solution.

STR:
1. Do a content-review. Either content-reject or send an information request.
2. The developer replies.
3. The email that is sent to the content reviewer contains the link to the technical review page, to which they usually don't have access.

I think what needs to happen is when receiving the email from the developer (after step 2) we need to determine whether it was a content-review (rejection or information request) and adapt the link in the email we send out to reviewers (in step 3).

diox:
Fixes #6936

### DIFF
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -696,6 +696,17 @@ def review(request, addon, channel=None):
     if unlisted_only and not acl.check_unlisted_addons_reviewer(request):
         raise PermissionDenied
 
+    # Are we looking at a listed review page while only having content review
+    # permissions ? Redirect to content review page, it will be more useful.
+    if (channel == amo.RELEASE_CHANNEL_LISTED and content_review is False and
+        acl.action_allowed(request, amo.permissions.ADDONS_CONTENT_REVIEW) and
+        not any(
+            acl.action_allowed(request, perm) for perm in (
+                amo.permissions.ADDONS_REVIEW,
+                amo.permissions.ADDONS_POST_REVIEW,
+                amo.permissions.ADDONS_RECOMMENDED_REVIEW))):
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/views.py
```diff
@@ -696,6 +696,17 @@ def review(request, addon, channel=None):
     if unlisted_only and not acl.check_unlisted_addons_reviewer(request):
         raise PermissionDenied
 
+    # Are we looking at a listed review page while only having content review
+    # permissions ? Redirect to content review page, it will be more useful.
+    if (channel == amo.RELEASE_CHANNEL_LISTED and content_review is False and
+        acl.action_allowed(request, amo.permissions.ADDONS_CONTENT_REVIEW) and
+        not any(
```
eviljeff: 😿 I guess there's no other way of doing it than this?

src/olympia/reviewers/views.py
```diff
@@ -696,6 +696,17 @@ def review(request, addon, channel=None):
...
+            acl.action_allowed(request, perm) for perm in (
+                amo.permissions.ADDONS_REVIEW,
```
eviljeff: is just `.ADDONS_REVIEW` sufficient?  i.e. does anyone actually have `.POST_REVIEW` or `.ADDONS_RECOMMENDED_REVIEW` while not also having `.ADDONS_REVIEW`?

diox: I'm not sure, I went for the safest approach. Eventually we want to merge `POST_REVIEW` and `ADDONS_REVIEW` for sure, but for now since they are separated everywhere in the code I treated them separately.

Issue #13843, opened 2020-03-27
Add links to reviewer tools on django admin add-ons by author view


wagnerand:
Similar to #13199, it would be helpful to admins to add a link to the listed/unlisted reviewer tools page (as applicable) to each row in the add-ons table on `admin/models/addons/addon/?authors=<id>`.

ioanarusiczki:
@willdurand 
The listed/unlisted links are available from the mentioned page
https://addons-internal.dev.mozaws.net/en-US/admin/models/addons/addon/?authors=10641574
Verified on admin tools dev - FF74(win10).

PR #13856, opened 2020-03-30
Add reviewer tools links in addon admin list

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13843

### DIFF
--- a/src/olympia/addons/admin.py
+++ b/src/olympia/addons/admin.py
@@ -1,4 +1,4 @@
-from urllib.parse import urlencode
+from urllib.parse import urlencode, urljoin
 
 from django import http, forms
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


src/olympia/addons/admin.py
```diff
@@ -167,6 +168,32 @@ def total_ratings_link(self, obj):
             count=obj.total_ratings)
     total_ratings_link.short_description = _(u'Ratings')
 
+    def reviewer_links(self, obj):
```
diox: We haven't looked at optimizing `AddonAdmin` at all yet, I don't think there is even an `assertNumQueries` test for the `changelist` view, and this particular case is tough to optimize, but it's going to add one to two (depending on whether the add-ons have a `current_version` set or not) queries _for each add-on_ in the list :( This might make it a lot slower, even if the queries should be fast...

willdurand: ok, I filed https://github.com/mozilla/addons-server/issues/13862 to add a test case at least. Depending on the outcome, we might want to spend some time optimizing the `changelist` view.

PR #13873, opened 2020-04-01
Use `Addon.unfiltered` manager in abuse report querysets

### LINKED ISSUE #12742: Use `Addon.unfiltered` manager in abuse report querysets
In #12735, we added support for "deleted add-ons" in the scanners django admin. The optimized queryset was using the default manager (`Addon.objects`), which automatically excluded deleted add-ons, leading to entries without add-on information in the django admin.

We fixed this problem by using the `Addon.unfiltered` manager, which does not exclude the deleted add-ons.

We noticed a similar behavior in the following optimizations:
```
src/olympia/abuse/admin.py:                'addon', queryset=Addon.objects.all().only_translations()),
src/olympia/reviewers/views.py:                    'addon', queryset=Addon.objects.all().only_translations())
src/olympia/reviewers/views.py:        Prefetch('addon', queryset=Addon.objects.all().only_translations()),
```

This issue is about fixing these occurrences.

junngo:
Fixes #12742 

I changed code for showing deleted addon.

path) Admin - abuse - abuse report
**Before**
12742_before

**After**
12742_after

### DIFF
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -4837,6 +4837,14 @@ class TestReview(ReviewBase):
 
         assert doc('.abuse_reports').text().split('\n') == expected
 
+        self.addon.delete()
+        self.url = reverse('reviewers.review', args=[self.addon.id])
+        response = self.client.get(self.url)
+        assert response.status_code == 200
+        doc = pq(response.content)
+        assert doc('.abuse_reports')
+        assert doc('.abuse_reports').text().split('\n') == expected
+
     def test_abuse_reports_unlisted_addon(self):
         user = UserProfile.objects.get(email='reviewer@mozilla.com')
         self.grant_permission(user, 'Addons:ReviewUnlisted')
@@ -5217,6 +5225,13 @@ class TestAbuseReportsView(ReviewerTest):
             'Et mël mazim ludus.',
         ]
 
+        assert doc('.abuse_reports').text().split('\n') == expected
+
+        self.addon.delete()
+        self.url = reverse('reviewers.abuse_reports', args=[self.addon.id])
+        assert response.status_code == 200
+        doc = pq(response.content)
+        assert len(doc('.abuse_reports')) == 1
         assert doc('.abuse_reports').text().split('\n') == expected
 
     def test_queries(self):
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -755,7 +755,8 @@ def review(request, addon, channel=None):
                 # a ton of potentially duplicate queries with all the useless
                 # Addon transforms.
                 Prefetch(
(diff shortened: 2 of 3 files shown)


src/olympia/reviewers/views.py
```diff
@@ -1037,7 +1038,7 @@ def abuse_reports(request, addon):
         Q(addon=addon) | Q(user__in=developers)
     ).select_related('user').prefetch_related(
         # See review(): we only need the add-on objects and their translations.
-        Prefetch('addon', queryset=Addon.objects.all().only_translations()),
+        Prefetch('addon', queryset=Addon.unfiltered.all().only_translations()),
```
eviljeff, line 1041: We need a test that proves this change works.  There's no prescriptive way to do this, but I'd add another test in `TestAbuseReportsView` in `reviewers/tests/test_views.py` that deletes `self.addon` and tries to do `test_abuse_reports` again (or something like it)

src/olympia/reviewers/views.py
```diff
@@ -755,7 +755,8 @@ def review(request, addon, channel=None):
                 # a ton of potentially duplicate queries with all the useless
                 # Addon transforms.
                 Prefetch(
-                    'addon', queryset=Addon.objects.all().only_translations())
+                    'addon',
+                    queryset=Addon.unfiltered.all().only_translations())
```
eviljeff, line 759: We need a test that proves this change works.  Something in `reviewers/tests/test_views.py` `TestReview` - If you look at the various `test_abuse_reports` methods that test scenarios you should be able to copy one and change it to work on a deleted addon instead and check the abuse reports section still works.

junngo:
Thanks @eviljeff 
I'll focus more on the test code.
I'm going to submit it after the modification.

junngo:
I submitted PR again.
But CI looks like failed status. When I looked at detail, I couldn't find failed.

Please let me know if I have anything to change related test code.

src/olympia/reviewers/tests/test_views.py
```diff
@@ -4816,6 +4816,13 @@ def test_abuse_reports(self):
 
         assert doc('.abuse_reports').text().split('\n') == expected
 
+        self.addon.update(status=amo.STATUS_DELETED)
+        response = self.client.get(self.url)
```
eviljeff, line 4820: after `self.addon.delete()` you'll need to regenerate `self.url` because the url uses `slug`, and `slug` is cleared during delete - the url will have to use `id` (or `pk`) instead of `slug`.
`self.url = reverse('reviewers.review', args=[self.addon.id])`

junngo:
As always, thank you :)
I submitted PR again. If it have any problems, please let me know.

eviljeff:
@junngo can you rebase this patch against master?  (Ubuntu removed some libs so our tests started failing - it's fixed on master)

PR #13875, opened 2020-04-01
Update DevHub footer links and alignment

### LINKED ISSUE #13656: Update DevHub homepage footer links to align with the new Firefox download page
Follow up for https://github.com/mozilla/addons-frontend/issues/9089

1. Open Developer Hub homepage - https://addons.allizom.org/en-US/developers/
2. Check the links present in the page Footer (AMO footer)

### What happened?
The footer is still using the links used by the previous amo-frontend footer

The footer should be updated to use the _Browsers_ and the _Products_ links present on the [Firefox download page](https://www.mozilla.org/en-US/firefox/new/) footer. 
image

The UTM parameters used by the frontend footer links are  `utm_source=addons.mozilla.org&utm_medium=referral&utm_content=footer-link`
Also, the Footer should be aligned on three different columns.

[amo-frontend footer( _note that the Browsers column is still not updated_)]

image

kanurag94:
Fixes #13656 
Updated UI 

Screenshot from 2020-04-01 19-56-54

### DIFF
--- a/src/olympia/templates/photon-footer.html
+++ b/src/olympia/templates/photon-footer.html
@@ -21,24 +21,24 @@
         </section>
 
         <section class="Footer-browsers-links">
-            <h4 class="Footer-links-header">Browsers</a></h4>
+            <h4 class="Footer-links-header">Browsers</h4>
 
             <ul class="Footer-links">
-                <li><a href="https://www.mozilla.org/firefox/new/?utm_content=footer-link&utm_medium=referral&utm_source=addons.mozilla.org">Desktop</a></li>
-                <li><a href="https://www.mozilla.org/firefox/mobile/?utm_content=footer-link&utm_medium=referral&utm_source=addons.mozilla.org">Mobile</a></li>
+                <li><a href="https://www.mozilla.org/firefox/new/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=footer-link">Desktop</a></li>
+                <li><a href="https://www.mozilla.org/firefox/mobile/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=footer-link">Mobile</a></li>
                 <li><a href="https://mixedreality.mozilla.org/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=footer-link">Reality</a></li>
-                <li><a href="https://www.mozilla.org/firefox/enterprise/?utm_content=footer-link&utm_medium=referral&utm_source=addons.mozilla.org">Enterprise</a></li>
+                <li><a href="https://www.mozilla.org/firefox/enterprise/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=footer-link">Enterprise</a></li>
             </ul>
         </section>
 
-        <section class="Footer-product-links">
+        <section class="Footer-products-links">
             <h4 class="Footer-links-header">Products</h4>
 
             <ul class="Footer-links">
-                <li><a href="https://www.mozilla.org/firefox/lockwise/?utm_content=footer-link&utm_medium=referral&utm_source=addons.mozilla.org">Lockwise</a></li>
+                <li><a href="https://www.mozilla.org/firefox/lockwise/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=footer-link">Lockwise</a></li>
                 <li><a href="https://monitor.firefox.com/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=footer-link">Monitor</a></li>
                 <li><a href="https://send.firefox.com/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=footer-link">Send</a></li>
-                <li><a href="https://www.mozilla.org/firefox/browsers/?utm_content=footer-link&utm_medium=referral&utm_source=addons.mozilla.org">Browsers</a></li>
+                <li><a href="https://www.mozilla.org/firefox/browsers/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=footer-link">Browsers</a></li>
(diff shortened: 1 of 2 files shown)


src/olympia/templates/photon-footer.html
```diff
@@ -20,24 +20,34 @@ <h4 class="Footer-links-header"><a href="{{ url('home') }}">Add-ons</a></h4>
             </ul>
         </section>
 
-        <section class="Footer-firefox-links">
-            <h4 class="Footer-links-header"><a href="https://www.mozilla.org/firefox/">Firefox</a></h4>
+        <section class="Footer-browsers-links">
+            <h4 class="Footer-links-header">Browsers</a></h4>
```
bobsilverberg, line 24: I see a closing `</a>` tag on this line, but no matching opening tag. I believe we just need to remove this closing tag.

src/olympia/templates/photon-footer.html
```diff
@@ -20,24 +20,34 @@ <h4 class="Footer-links-header"><a href="{{ url('home') }}">Add-ons</a></h4>
...
-                <li><a href="https://www.mozilla.org/firefox/android/?utm_source=addons.mozilla.org&amp;utm_campaign=footer&amp;utm_medium=referral">Android Browser</a></li>
-                <li><a href="https://www.mozilla.org/firefox/ios/?utm_source=addons.mozilla.org&amp;utm_campaign=footer&amp;utm_medium=referral">iOS Browser</a></li>
-                <li><a href="https://www.mozilla.org/firefox/focus/?utm_source=addons.mozilla.org&amp;utm_campaign=footer&amp;utm_medium=referral">Focus Browser</a></li>
-                <li><a href="https://www.mozilla.org/firefox/desktop/?utm_source=addons.mozilla.org&amp;utm_campaign=footer&amp;utm_medium=referral">Desktop Browser</a></li>
-                <li><a href="https://www.mozilla.org/firefox/channel/desktop/?utm_source=addons.mozilla.org&amp;utm_campaign=footer&amp;utm_medium=referral">Beta, Nightly, Developer Edition</a></li>
+                <li><a href="https://www.mozilla.org/firefox/new/?utm_content=footer-link&utm_medium=referral&utm_source=addons.mozilla.org">Desktop</a></li>
```
bobsilverberg, line 27: The previous code used `&amp;` in place of `&`. Should we continue to do that instead of using just `&` here? (cc @eviljeff)

src/olympia/templates/photon-footer.html
```diff
@@ -20,24 +20,34 @@ <h4 class="Footer-links-header"><a href="{{ url('home') }}">Add-ons</a></h4>
...
+                <li><a href="https://www.mozilla.org/firefox/mobile/?utm_content=footer-link&utm_medium=referral&utm_source=addons.mozilla.org">Mobile</a></li>
```
bobsilverberg, line 28: The order of the params is inconsistent across the various links. I.e., some have `utm_content` first and some have `utm_source`, etc. It would be cleaner and easier to read if all of the links had the params in the same order.

src/olympia/templates/photon-footer.html
```diff
@@ -20,24 +20,34 @@ <h4 class="Footer-links-header"><a href="{{ url('home') }}">Add-ons</a></h4>
...
+                <li><a href="https://mixedreality.mozilla.org/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=footer-link">Reality</a></li>
+                <li><a href="https://www.mozilla.org/firefox/enterprise/?utm_content=footer-link&utm_medium=referral&utm_source=addons.mozilla.org">Enterprise</a></li>
             </ul>
+        </section>
+
+        <section class="Footer-product-links">
```
bobsilverberg, line 34: You called the class on the browsers section `Footer-browsers-links` but this class is called `Footer-product-links`. I think it should be `Footer-products-links` for consistency.

src/olympia/templates/photon-footer.html
```diff
@@ -20,24 +20,34 @@ <h4 class="Footer-links-header"><a href="{{ url('home') }}">Add-ons</a></h4>
...
+            <h4 class="Footer-links-header">Products</h4>
+
+            <ul class="Footer-links">
+                <li><a href="https://www.mozilla.org/firefox/lockwise/?utm_content=footer-link&utm_medium=referral&utm_source=addons.mozilla.org">Lockwise</a></li>
```
bobsilverberg, line 38: My comments above apply here as well. I.e., we should probably use `&amp;`, and the params should be in the same order for each link.

src/olympia/templates/photon-footer.html
kanurag94: I thought so too. But the current implementation at https://addons.mozilla.org/en-US/firefox/ follows Footer-product-links

src/olympia/templates/photon-footer.html
kanurag94: Thanks, rectified.

src/olympia/templates/photon-footer.html
kanurag94: Most of these changes were inspired from the format at https://addons.mozilla.org/en-US/firefox/ , I guess once this resolve we should be fixing those too

src/olympia/templates/photon-footer.html
bobsilverberg: @kanurag94 We discussed this and decided that it's fine to just use `&` in the links as you've done. I would like to see the order of params fixed, though.

kanurag94: Will do and update PR. :)

src/olympia/templates/photon-footer.html
kanurag94: What about this @bobsilverberg ?

bobsilverberg: I think it's better to be consistent in this repo, rather than try to reproduce what is in a different repo, so I'd still suggest `products`.

PR #13898, opened 2020-04-03
Handle long blocklistsubmission urls via fxa

### LINKED ISSUE #13752: 'Invalid parameter: state` for a login triggered by a Block action from Rev Tools
Prerequisites: User is logged out in django admin

1. Log in with an admin user on AMO
2. Open a listed add-on review page 
3. Scroll down to More actions and click on 'Block add-on'
4. Enter valid admin credentials on the FxA login page
5. Click Sign in

### What happened?
The following error is received:
image

You can also see this error in the console with a 400 status code:
```
POST https://stable.dev.lcip.org/auth/v1/oauth/authorization
```

Log in is successful 

- it will also happen for unlisted add-ons when the Block multiple version option is used
- I did not reproduce the error when I tried to log into django admin from other links from Reviewer Tools or Dev Hub

eviljeff:
fixes #13752 by:
1) adding a shortcut url for adding a single addon guid by addon pk (so we're not creating a state that's too long) via `/admin/models/block/block_add_addon/<pk>/` and
2) correctly redirect a url that has a `?next=...` param, which django admin adds when a login is forced.  (not strictly related to this issue, because we've always failed to properly redirect direct urls to /model/ pages through a login via django admin, but without it makes the fix in (1) a bit pointless)

### DIFF
--- a/src/olympia/blocklist/admin.py
+++ b/src/olympia/blocklist/admin.py
@@ -101,10 +101,10 @@ class BlockAdminAddMixin():
 
     def add_from_addon_pk_view(self, request, pk, **kwargs):
         addon = get_object_or_404(Addon.unfiltered, pk=pk or kwargs.get('pk'))
-        get_params = [f'{key}={value}' for key, value in request.GET.items()]
+        get_params = request.GET.urlencode()
         return redirect(
-            reverse('admin:blocklist_blocklistsubmission_add') + '?' +
-            '&'.join([f'guids={addon.guid}'] + get_params))
+            reverse('admin:blocklist_blocklistsubmission_add') +
+            f'?guids={addon.guid}&{get_params}')
 
 
 @admin.register(BlocklistSubmission)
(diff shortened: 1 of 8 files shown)


src/olympia/blocklist/admin.py
```diff
@@ -94,6 +99,13 @@ def _multi_input_view(self, request, *, add, form_url='',
         return TemplateResponse(
             request, 'blocklist/multi_guid_input.html', context)
 
+    def add_from_addon_pk_view(self, request, pk, **kwargs):
+        addon = get_object_or_404(Addon.unfiltered, pk=pk or kwargs.get('pk'))
+        get_params = [f'{key}={value}' for key, value in request.GET.items()]
+        return redirect(
+            reverse('admin:blocklist_blocklistsubmission_add') + '?' +
+            '&'.join([f'guids={addon.guid}'] + get_params))
```
diox, line 107: Would break if somehow passed `GET` params that need to be escaped. How about:

```suggestion
        get_params = request.GET.urlencode()
        return redirect(
            reverse('admin:blocklist_blocklistsubmission_add') +
            f'?guids={addon.guid}&{get_params}'
```

eviljeff: https://github.com/mozilla/addons-server/pull/13898/commits/d8357eca085cb7ad177281afbb897ce432360802 (the suggestion was missing the closing `)`

PR #13911, opened 2020-04-06
Show add-on name translations in add-on review page

### LINKED ISSUE #13910: Show all available translations for add-on name in review page
Split from #11516 : as a minimal workaround we want to show all translations for the add-on name in the review page. 

This issue is exclusively about the add-on name: we'll focus on other fields like summary and description later in #11516, they require a more complex solution in terms of UX (hard to balance wanting to show all translations for reviewers to see everything at a glance with not turning the review page into a big mess with several pages of content to look at).

diox:
Fixes #13910

### DIFF
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -882,6 +882,14 @@ def review(request, addon, channel=None):
         amo.LOG.REMOVE_USER_WITH_ROLE.id]
     user_changes_log = ActivityLog.objects.filter(
         action__in=user_changes_actions, addonlog__addon=addon).order_by('id')
+
+    name_translations = (
+        addon.name.__class__.objects.filter(
+            id=addon.name.id, localized_string__isnull=False).exclude(
+            localized_string='')
+        if addon.name else []
+    )
+
     ctx = context(
         actions=actions, actions_comments=actions_comments,
         actions_full=actions_full, addon=addon,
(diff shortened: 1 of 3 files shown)


eviljeff:
can you add a screenshot showing how the (section of the) page looks with an add-on with multiple localizations?

diox:
Screenshot

src/olympia/reviewers/views.py
```diff
@@ -890,8 +898,8 @@ def review(request, addon, channel=None):
         content_review=content_review, count=count,
         deleted_addon_ids=deleted_addon_ids, flags=flags,
         form=form, is_admin=is_admin, is_recommendable=is_recommendable,
-        now=datetime.now(), num_pages=num_pages,
-        pager=pager, reports=reports, show_diff=show_diff,
+        name_translations=name_translations, now=datetime.now(),
+        num_pages=num_pages, pager=pager, reports=reports, show_diff=show_diff,
```
eviljeff, line 902: ugh, these should just be all on their own lines when there are these many params.  (no requirement to change in this patch, just saying)

src/olympia/reviewers/views.py
```diff
@@ -882,6 +882,14 @@ def review(request, addon, channel=None):
         amo.LOG.REMOVE_USER_WITH_ROLE.id]
     user_changes_log = ActivityLog.objects.filter(
         action__in=user_changes_actions, addonlog__addon=addon).order_by('id')
+
+    name_translations = (
+        addon.name.__class__.objects.filter(
+            id=addon.name.id, localized_string__isnull=False).exclude(
+            localized_string='')
+        if addon.name else []
+    )
```
eviljeff, line 891: Works as a solution.  If `name_translations` was preprocessed though, it could be list of tuples of locales+string, e.g. `[(('en-US', 'en-GB'), 'English Translation'),(('fr',), 'Traduction En Français')]` that would result in a shorter list.

diox: https://github.com/mozilla/addons-server/pull/13911/commits/a4eeb3cb75eb9e84513028fe0ad832fd5008dd59 does that in template

src/olympia/reviewers/views.py
diox: Yeah, my first version of the patch did that, but then I rebased it a hundred times since and it kept conflicting so I didn't keep it... Next time...

PR #13912, opened 2020-04-06
translate /d to [:digit:] for mysql regexs

### LINKED ISSUE #13887: v2 blocklist regex doesn't match all guids with import_blocklist
Some of the blocklist regular expressions that should match guids aren't matching all the expected add-ons.  The blocks do match _some_ addon guids, just not all they would be expected to.

There may be a regex syntax incompatibility between javascript and mysql.

[missing_blocks.txt](https://github.com/mozilla/addons-server/files/4421943/missing_blocks.txt)

eviljeff:
fix #13887

### DIFF
--- a/src/olympia/blocklist/tasks.py
+++ b/src/olympia/blocklist/tasks.py
@@ -90,7 +90,7 @@ def import_block_from_blocklist(record):
                 'Kinto %s: Unable to break down regex into list; '
                 'attempting to create Blocks for guids matching [%s]',
                 kinto_id, guid_regexp)
-            # mysql doesn't support /d - only [:digit:]
+            # mysql doesn't support \d - only [:digit:]
             guid_regexp = guid_regexp.replace(r'\d', '[[:digit:]]')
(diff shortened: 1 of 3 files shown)


src/olympia/blocklist/tasks.py
```diff
@@ -90,6 +90,8 @@ def import_block_from_blocklist(record):
                 'Kinto %s: Unable to break down regex into list; '
                 'attempting to create Blocks for guids matching [%s]',
                 kinto_id, guid_regexp)
+            # mysql doesn't support /d - only [:digit:]
```
diox, line 93: `\` not `/`

eviljeff: doesn't support `\d` and doesn't support `/d` either! :)

eviljeff: https://github.com/mozilla/addons-server/pull/13912/commits/e67bdbddcdd0933eb11edf9bcafa7d6e17dd3c62

PR #13918, opened 2020-04-06
Update default need-info timeout to 14 days

### LINKED ISSUE #13915: Set default for "reply required" to 14 days
When reviewers request more information from an add-on developer, the notification email says the developer has 7 days to reply. The "7" is the default number that can be toggled by admins but not contributor reviewers. 

Due to the COVID-19 situation, we'd like to give developers a default of 14 days to respond to requests.

wagnerand:
Fix #13915.


src/olympia/activity/tests/test_utils.py
```diff
@@ -287,6 +288,8 @@ def _check_email_info_request(self, call, url, reason_text, days_text):
             self.addon.name, self.version.version)
         assert ('visit %s' % url) in body
         assert ('receiving this email because %s' % reason_text) in body
+        print(days_text)
+        print(body)
```
eviljeff, line 292: leftover `print`s

Issue #13019, opened 2019-12-03
Git: disable checking for `.keep` files


EnTeQuAk:
This is a follow-up once a new pygit/libgit release has been issued with **https://github.com/libgit2/pygit2/pull/908/ landed.

This might ease stress on EFS during extraction and followed commit.

ioanarusiczki:
@EnTeQuAk Please be more specific about " git extraction works as before.", which part of the server is affected, more exactly what needs to be tested by qa. 
Thanks!

ioanarusiczki:
@willdurand I verified on dev with several addon uploads - listed and unlisted and then checked that codemanager displayed the files. 
If there's anything else I could check on please let me know. 
Verified on dev with FF75(Win10).

willdurand:
@ioanarusiczki thanks, I guess it should be okay then..

PR #13921, opened 2020-04-07
Update pygit2 to 1.1.1

willdurand:
In theory we could rely on wheels for the `libgit2-dev` lib but this is broken
(PEP517 something something) and it is not new. That's why I decided to switch
to the more manual way of installing `libgit2-dev`.

I was trying to fix https://github.com/mozilla/addons-server/pull/13902 but
that did not work because the `libgit2-dev` package is not available in any
debian repos... So here we are, with version `1.1.1`.


willdurand:
This would allow us to land https://github.com/mozilla/addons-server/issues/13019

willdurand:
The test failure is unrelated (ADI command again)

PR #13922, opened 2020-04-07
Disable pack keep file checks

willdurand:
~~:warning: depends on #13921~~

Fixes https://github.com/mozilla/addons-server/issues/13019

---

I am not sure how to test that, though..


willdurand:
> Looks fine. You said you weren't sure how to test it - I assume that means in an automated test. Were you able to test it interactively, or is that also not seemingly possible? I guess, as it's just configuration, it's not something we'd normally test anyway.

Yeah, it is just configuration and it is a server setting so I don't know.

PR #13925, opened 2020-04-07
Update pygit2 to 1.1.1

willdurand:
In theory we could rely on wheels for the `libgit2-dev` lib but this is broken
(PEP517 something something) and it is not new. That's why I decided to switch
to the more manual way of installing `libgit2-dev`.

I was trying to fix https://github.com/mozilla/addons-server/pull/13902 but
that did not work because the `libgit2-dev` package is not available in any
debian repos... So here we are, with version `1.1.1`.


Dockerfile
```diff
@@ -56,11 +59,19 @@ RUN apt-get update && apt-get -t stretch install -y \
...
 # file-detection.
 RUN apt-get update && apt-get -t buster install -y \
-       file \
-       libmagic-dev \
+        file \
+        libmagic-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install a recent libgit2-dev version...
+RUN apt-get update && apt-get -t buster-backports install -y \
```
eviljeff, line 73: are these additional `apt-get update`s necessary when we just did it a few lines above?

Dockerfile.deploy
```diff
@@ -51,13 +54,21 @@ RUN apt-get update && apt-get -t stretch install -y \
...
+
 RUN apt-get update && apt-get -t buster install -y \
        # For an up-to-date `file` and `libmagic-dev` library for better file
        # detection.
        file \
        libmagic-dev \
     && rm -rf /var/lib/apt/lists/*
 
+# Install a recent libgit2-dev version...
+RUN apt-get update && apt-get -t buster-backports install -y \
```
eviljeff, line 68: same question about repeated `apt-get update` commands as `Dockerfile`

requirements/prod.txt
```diff
@@ -356,20 +356,20 @@ typing==3.7.4.1 \
...
-    --hash=sha256:3b60c7037454d4d4ac6ddfef5f7afc4c3115143e5c18ae1cb90a8a9e514691fa \
-    --hash=sha256:4b7ffbd09a0e5da3f24c8913498fc628a5f27e171a2e0509890403b6cbe1f333 \
-    --hash=sha256:5ac57ef32d0b7edb57e0dc49a02c52477821539827dcdac415d54c4cf66b1af2 \
-    --hash=sha256:6313c0febdf21173bb6d6088c760b0f9ec3887e8afaa14e53b8003c82cada97c \
-    --hash=sha256:9429a998cfdc76ccea8512277c6581f75a30099fe4dfbe7a5a10c541013180a9 \
-    --hash=sha256:a8cd17ca7ccf3585b031dded74a7dcb00fc3231aa1cc00d98eecbafe20c5378d \
-    --hash=sha256:b7c0abce8cfe4b570e78d16627fe44e76004a96f6cb1c77c8a3ffe8ce37dd2b0 \
-    --hash=sha256:c316248dae9f2a1d79275dc36f7d3200b217931367c6c37eda751d7ea6f441dc \
-    --hash=sha256:d6c1864da421ad41f0548a58b2e2d9d808e55bccde80a8b047aa1027f687213d
+pygit2==1.1.1 \
```
eviljeff, line 359: oh, I forgot, if pygit2 can't use 1.2 and above then we should pin the dependency to less than 1.2 (or equal to 1.1.1)

Dockerfile
willdurand: Yes because we clear the cache with `rm -rf /var/lib/apt/lists/*`

requirements/prod.txt
willdurand: isn't it already equal to 1.1.1?

eviljeff: I wasn't clear, I meant for pyup so it doesn't offer to update it every day.  See the Django requirement at the top of the file for example.

willdurand: gotcha!

PR #13934, opened 2020-04-08
Don't redirect admins that have *:* when viewing validation results

### LINKED ISSUE #13889: code-manager does not finish loading for admins for some add-ons
https://code.addons.mozilla.org/en-US/browse/2633678/versions/4995785/ is stuck showing `Loading linter information...`. In the web console, I see CORS request failures:

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://addons.mozilla.org/en-US/developers/addon/cookie-api-test-extension3/file/3540155/validation.json. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://addons.mozilla.org/en-US/developers/addon/cookie-api-test-extension3/file/3540155/validation.json. (Reason: CORS request did not succeed).
```

Loading that file validation page manually redirects me to https://addons.mozilla.org/en-US/developers/addon/cookie-api-test-extension3/submit/details

@diox has figured out the cause for this, I'll leave it up to him to add more details, as needed.

diox:
Because `check_addon_ownership()` returns `True` for admins, they were redirected to the submission flow for addons that somehow have a listed versions but are incomplete (status=0) w/ incomplete metadata. Since they also would be considered having `ReviewerTools:View`, swapping the 2 conditions will let them through.

Fixes #13889

### DIFF
--- a/src/olympia/devhub/tests/test_views_validation.py
+++ b/src/olympia/devhub/tests/test_views_validation.py
@@ -151,22 +151,22 @@ class TestFileValidation(TestCase):
     def test_only_dev_can_see_results(self):
         self.client.logout()
         assert self.client.login(email='regular@mozilla.com')
-        assert self.client.head(self.url, follow=True).status_code == 403
+        assert self.client.head(self.url, follow=False).status_code == 403
 
     def test_only_dev_can_see_json_results(self):
         self.client.logout()
         assert self.client.login(email='regular@mozilla.com')
-        assert self.client.head(self.json_url, follow=True).status_code == 403
+        assert self.client.head(self.json_url, follow=False).status_code == 403
 
     def test_reviewer_can_see_results(self):
         self.client.logout()
         assert self.client.login(email='reviewer@mozilla.com')
-        assert self.client.head(self.url, follow=True).status_code == 200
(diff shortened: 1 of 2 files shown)


src/olympia/devhub/tests/test_views_validation.py
```diff
@@ -151,62 +151,79 @@ def test_results_page(self):
     def test_only_dev_can_see_results(self):
         self.client.logout()
         assert self.client.login(email='regular@mozilla.com')
-        assert self.client.head(self.url, follow=True).status_code == 403
+        assert self.client.head(self.url, follow=False).status_code == 403
```
bobsilverberg, line 154: What are all the switches from `follow=True` to `follow=False` about?

diox: Those `follow=True` force the test client to follow redirects, but in this particular case there is no good reason to do so as it would just hide similar class of bugs. If you look at the tests I've added, but set them to follow redirect like the ones that I modified, then the added tests pass even without any chances, since what we're concerned about here *is* an erroneous redirect.

Issue #13590, opened 2020-02-27
GitError: cannot locate local branch 'unlisted'


muffinresearch:
There are 90k events in sentry for this see https://sentry.prod.mozaws.net/share/issue/8ea46d4d182144cd8c972708a2ffd633/

```
GitError: cannot locate local branch 'unlisted'
  File "celery/app/trace.py", line 385, in trace_task
    R = retval = fun(*args, **kwargs)
  File "celery/app/trace.py", line 650, in __protected_call__
    return self.run(*args, **kwargs)
  File "olympia/amo/decorators.py", line 108, in wrapper
    return f(*args, **kw)
  File "olympia/versions/tasks.py", line 130, in extract_version_to_git
    version=version, author=author, note=note)
  File "olympia/lib/git.py", line 329, in extract_and_commit_from_version
    branch = repo.find_or_create_branch(BRANCHES[version.channel])
  File "olympia/lib/git.py", line 394, in find_or_create_branch
    branch = self.git_repository.branches.get(name)
  File "pygit2/repository.py", line 1212, in get
    return self[key]
  File "pygit2/repository.py", line 1200, in __getitem__
    branch = self._repository.lookup_branch(name, GIT_BRANCH_LOCAL)
```

/cc @EnTeQuAk

diox:
See also https://github.com/mozilla/addons-server/issues/11431

willdurand:
Sentry issue: [OLYMPIA-PROD-31G](https://sentry.prod.mozaws.net/operations/olympia-prod/issues/5713574/)

willdurand:
I started to look into this error. This is an error coming from `libgit2` directly.

According to the `pygit2` lib/docs, `lookup_branch` should only raise `KeyError` internally (this error is handled [here](https://github.com/libgit2/pygit2/blob/31f27f7f6223d4d14fc79f45be5f65b72b855303/pygit2/repository.py#L1215)) or `InvalidSpecError` (cf. https://github.com/libgit2/pygit2/issues/832) when we use an invalid branch name.

In this case, this is a `GitError`, which is the most generic error we can have in `pygit2`. It is obviously not a branch name issue because the name of the branch is `unlisted` (without any special or forbidden characters).

[...]

PR #13936, opened 2020-04-08
[WIP] Remove broken git ref file when GitError is raised

willdurand:

---

After having investigated this issue, I think this is the right patch assuming
branches are not corrupted after some time. ~~The git repos from prod I reviewed
were created in April 2019 and I did not find any other commit so I believe the
branch was not created correctly the first time and no extraction was ever 
committed to this branch.~~

~~Theoretically, no git commit can be lost so if a branch is corrupted after some
time (which I don't think is the case but I cannot be 100% sure), deleting and
recreating this branch will fix the extraction issue and we could still
re-attach the commits (cherry-pick actually) to the newly created branch (or we
re-extract the missing versions).~~

**EDIT:** so I found a different command to retrieve unreachable git commits
and there were in fact other git commits :( So that invalidates my theory here.
Ugh.

We could use a waffle switch if we want to be extra careful about this change.

Other (nuclear) option: we delete the entire repository and re-extract
everything but I don't know how to do this from within the
`find_or_create_branch()` function.

WDYT?

### DIFF
--- a/src/olympia/lib/git.py
+++ b/src/olympia/lib/git.py
@@ -391,7 +391,19 @@ class AddonGitRepository(object):
 
     def find_or_create_branch(self, name):
         """Lookup or create the branch named `name`"""
-        branch = self.git_repository.branches.get(name)
+        try:
+            branch = self.git_repository.branches.get(name)
+        except pygit2.GitError:
+            # The ref (branch) is broken so we have to remove the file. In this
+            # case, there is no other way to solve the problem if we want to
+            # use the `name` branch.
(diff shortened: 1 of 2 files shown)


willdurand:
We discussed with @diox about:

> **EDIT:** so I found a different command to retrieve unreachable git commits
> and there were in fact other git commits :( So that invalidates my theory here.
> Ugh.

We think that we should re-extract all the versions of the add-on when this happens because we can't be sure when the branch ref became broken.

PR #13940, opened 2020-04-09
An attempt to fix 'GitError: cannot locate local branch'

willdurand:

---

This patch does a few things to fix the issue linked above:

1. it detects a broken ref error in `find_or_create_branch()`
2. when a broken git ref is detected, the git repo cannot be repaired so we remove it and create a new celery task to re-extract all the versions of an add-on (this is a new task!)
3. it adds a `GitExtraction` model that is used to lock an add-on for git extraction. This should prevent the task to re-extract an add-on to be executed twice (or more) in parallel.
4. this lock is also used to delay the git extraction tasks for new versions (if any)


src/olympia/lib/git.py
```diff
@@ -397,7 +406,11 @@ def get_author(self, user=None):
 
     def find_or_create_branch(self, name):
         """Lookup or create the branch named `name`"""
-        branch = self.git_repository.branches.get(name)
+        try:
+            branch = self.git_repository.branches.get(name)
+        except pygit2.GitError:
+            raise BrokenRefError('Reference for branch "{}" is '
```
diox, line 412: Probably should log here as well

src/olympia/lib/git.py
```diff
@@ -261,6 +265,11 @@ def git_repository(self):
 
         return git_repository
 
+    def delete(self):
+        if not self.is_extracted:
```
diox, line 269: This shouldn't happen, right ? Maybe we should log it

willdurand: It should not, unless we (developers) call it before the repo is initialized. I was thinking about raising an error but logging might be enough.

src/olympia/versions/tasks.py
```diff
@@ -126,12 +156,30 @@ def extract_version_to_git(version_id, author_id=None, note=None):
...
+        # Reset the git hash of each version of the add-on related to the
+        # current add-on because we want to re-extract everything.
+        versions = addon.versions.update(git_hash='')
+        # Retrieve the repo for the add-on and delete it.
+        addon_repo = AddonGitRepository(addon, package_type='addon')
+        addon_repo.delete()
+        log.warn('Deleted the git addon repository for addon_id={} because we '
+                 'detected a broken ref.'.format(addon.id))
+        # Create a task to re-extract the add-on.
+        extract_addon_to_git.delay(addon.pk)
```
diox, line 182: Might need something to avoid duplicate tasks for the same add-on. Also, before deleting the repos, you might want to add something to make sure other version extraction tasks for the same add-on won't be processed.

willdurand: Yep, those are the two questions I had in mind and that's why I opened this draft now.

> Might need something to avoid duplicate tasks for the same add-on.

This can probably be solved with some sort of _lock_.

> Also, before deleting the repos, you might want to add something to make sure other version extraction tasks for the same add-on won't be processed.

I am not sure to know how to do that yet. The other problematic scenario would be:

1. broken ref, repo deletion
2. task to re-extract add-on starts
3. new version is submitted by developer

How do we make sure this last version gets extracted?

I think this is where using rabbitmq acknowledgment would be nice because we could put the task back into the queue if its related add-on is being re-extracted. But maybe we have something to solve this problem already.

src/olympia/lib/git.py
willdurand: Yep. This error is logged in the caller code but we might also want to log it here as well.

src/olympia/versions/tasks.py
```diff
@@ -112,7 +113,36 @@ def delete_preview_files(pk, **kw):
 
 @task
 @use_primary_db
-def extract_version_to_git(version_id, author_id=None, note=None):
+def extract_addon_to_git(addon_pk):
```
willdurand, line 116: It's worth noting that once we solve the issue for the add-on code/repo, we can apply the same approach to source repo/code.

src/olympia/versions/tasks.py
```diff
@@ -126,11 +156,30 @@ def extract_version_to_git(version_id, author_id=None, note=None):
...
-        version=version, author=author, note=note)
+    try:
+        repo = AddonGitRepository.extract_and_commit_from_version(
+            version=version, author=author, note=note)
 
-    log.info('Extracted {version} into {git_path}'.format(
-        version=version_id, git_path=repo.git_repository_path))
+        log.info('Extracted {version} into {git_path}'.format(
+            version=version_id, git_path=repo.git_repository_path))
+    except BrokenRefError as err:
```
willdurand, line 165: (The fact that we only handle `BrokenRefError` here is by design)

src/olympia/versions/tasks.py
diox: Something something https://docs.celeryproject.org/en/latest/faq.html#faq-acks-late-vs-retry combined with some sort of lock might help.

I don't think we anything to retry skipped/failed extraction of a specific version at the moment. We do know about them because they are missing `source_git_hash` or `git_hash` on the `Version` instance, but I don't think we currently try to catch up later and re-extract.

willdurand: Do you think it would be practical to add a new field on `Addon` (`git_extraction_in_progress` but shorter) that we could us:

- in `extract_addon_to_git()` to avoid this task to be executed more than once
- in `extract_version_to_git()` to... re-submit itself?

That would probably solve the issues we're discussing here. Maybe we could add a small delay before re-submitting the `extract_version_to_git` task from within `extract_version_to_git`.

WDYT?

diox: That could work, yeah. I would use a separate model with a `models.OneToOneField(Addon, primary_key=True, on_delete=models.CASCADE)` instead of adding a new field to `Addon`, though.

src/olympia/versions/tasks.py
```diff
@@ -112,11 +114,94 @@ def delete_preview_files(pk, **kw):
...
     # we can access them.
     version = Version.unfiltered.get(pk=version_id)
+    addon = version.addon
+
+    if can_be_delayed and addon.git_extraction_is_in_progress:
+        log.info('Delaying task "extract_version_to_git" for version_id={} '
+                 'because the add-on is already being git-extracted.'
+                 ''.format(version_id))
+
+        time.sleep(30)  # 30 seconds
```
willdurand, line 196: I am not sure about this but that would avoid spamming the queue..

diox: You can give a [`countdown` to the `delay()` call](https://docs.celeryproject.org/en/latest/userguide/calling.html#eta-and-countdown) instead.

willdurand: oh, amazing. Thanks!

willdurand:
@diox this should be good for a review now. There are a two things that we should decide:

- are we happy with a 30s countdown?
- should we add a kill switch (with a waffle switch)?

src/olympia/addons/models.py
```diff
@@ -2099,3 +2105,14 @@ class ReusedGUID(ModelBase):
     guid = models.CharField(max_length=255, null=False)
     addon = models.OneToOneField(
         Addon, null=False, on_delete=models.CASCADE, unique=True)
+
+
+class GitExtraction(ModelBase):
```
willdurand, line 2110: mmm maybe this could be renamed to `AddonGitExtraction` to avoid potential conflicts in the future but the relation name would become `addongitextraction`...

willdurand:
I added two new test cases for the `delete()` method.

src/olympia/addons/models.py
diox: You can customize the related name... But I don't think it matters at the moment. This is fine like this.

src/olympia/versions/tasks.py
```diff
@@ -126,11 +211,29 @@ def extract_version_to_git(version_id, author_id=None, note=None):
...
+        # reference is detected in `extract_addon_to_git()`, which we call
+        # later in this block.
+        if stop_on_broken_ref:
+            raise err
 
-    log.info('Extracted {version} into {git_path}'.format(
-        version=version_id, git_path=repo.git_repository_path))
+        # Reset the git hash of each version of the add-on related to the
+        # current add-on because we want to re-extract everything.
+        addon.versions.update(git_hash='')
```
diox, line 229: You need to use `addon.versions(manager='unfiltered_for_relations')` to also include deleted versions.

src/olympia/versions/tasks.py
```diff
@@ -126,11 +211,29 @@ def extract_version_to_git(version_id, author_id=None, note=None):
...
+        # Retrieve the repo for the add-on and delete it.
+        addon_repo = AddonGitRepository(addon, package_type='addon')
+        addon_repo.delete()
```
diox, line 232: I think you should add the lock on the addon before you do this. Otherwise, if there is *another* extraction in progress for that addon that is somehow not broken (maybe because it's on a different, not broken channel) it might go through while you're deleting the repos...

In fact, I think the lock should be set super early in all extraction tasks (addon and version) and released once they are done, in order to prevent parallel extractions of multiple versions of a given add-on.

You'll also need something to remove the lock in case anything goes wrong in all tasks, otherwise you might end up locking an add-on for all eternity...

src/olympia/versions/tasks.py
diox: Maybe this should be done by the `delete()` method of the `AddonGitRepository`.

src/olympia/versions/tasks.py
```diff
@@ -112,11 +113,95 @@ def delete_preview_files(pk, **kw):
...
+
+    # "Lock" this add-on so that the git extraction can only run once and tasks
+    # for extracting related versions are delayed until this task has finished.
+    git_storage, created = GitExtraction.objects.update_or_create(
+        addon=addon,
+        defaults={'in_progress': True},
+    )
+
+    # Filter out versions that are already present in the git storage.
+    versions = addon.versions.filter(git_hash='').order_by('created')
```
diox, line 135: You need to use `addon.versions(manager='unfiltered_for_relations')` to also include deleted versions.

diox: Also, if all we need are version pks, might as well use `values_list('pk', flat=True)` to avoid fetching objects needlessly. Especially as we're looping over all existing versions for this addon.

src/olympia/versions/tasks.py
```diff
@@ -112,11 +113,95 @@ def delete_preview_files(pk, **kw):
...
+
+    for version in versions:
+        try:
+            log.info('Starting extraction of version "{}" for addon "{}" to '
+                     'git storage.'.format(version.pk, addon_pk))
+
+            extract_version_to_git(
```
diox, line 142: This should be called with `.delay()`, otherwise you're directly calling the function. You'll end up hitting time-limits if not using tasks... See comment about `chain` below.

src/olympia/versions/tasks.py
```diff
@@ -112,11 +113,95 @@ def delete_preview_files(pk, **kw):
...
+                version.pk,
+                # We do not want to trigger an infinite loop so an error will
+                # be thrown if there is a broken git reference.
+                stop_on_broken_ref=True,
+                # The `extract_version_to_git` task also checks the lock we set
+                # above so we need to tell it to skip this check here.
+                can_be_delayed=False,
+            )
+
+            log.info('Ending extraction of version "{}" for addon "{}" to '
```
diox, line 152: Can't work if using tasks, obviously. This, and the `git_storage.update(in_progress=False)`, likely need to be in a `chord` callback... Actually, a `chain` even: since we're not obeying the lock, we probably want the extractions to happen in sequence, not in parallel.

src/olympia/versions/tasks.py
```diff
@@ -112,11 +113,95 @@ def delete_preview_files(pk, **kw):
...
+        log.info('Delaying task "extract_version_to_git" for version_id={} '
+                 'because the add-on is already being git-extracted.'
+                 ''.format(version_id))
+
+        extract_version_to_git.apply_async(
+            kwargs={
+                'version_id': version_id,
+                'author_id': author_id,
+                'note': note,
+                'stop_on_broken_ref': stop_on_broken_ref
```
diox, line 200: Just in case I'd propagate `can_be_delayed` as well (in case the default value ever changes)

src/olympia/versions/tasks.py
```diff
@@ -112,11 +113,95 @@ def delete_preview_files(pk, **kw):
...
+            },
+            countdown=30,  # Executes the task in 30 seconds from now.
```
diox, line 202: Note: there is also a `retry` mechanism in celery that might work better for this. But it's not necessarily needed.

src/olympia/versions/tasks.py
diox: Should add a comment about this.

src/olympia/versions/tasks.py
willdurand: > You need to use `addon.versions(manager='unfiltered_for_relations')` to also include deleted versions.

Interesting. This line comes from the "migrate to git" command, so does that mean we didn't git-extract the deleted versions?

src/olympia/versions/tasks.py
willdurand: Again, this was borrowed from the migrate command. Calling the function directly was not a mistake here but I forgot about the soft timeout...

src/olympia/versions/tasks.py
willdurand: So we would build a gigantic chain of tasks for each version? That will probably speed up the whole process but it makes things more complicated since we'll have to handle errors in the chain (for instance, to reset the lock).

src/olympia/versions/tasks.py
willdurand: > I think you should add the lock on the addon before you do this. Otherwise, if there is _another_ extraction in progress for that addon that is somehow not broken (maybe because it's on a different, not broken channel) it might go through while you're deleting the repos...

Ah, right. That would make sense but then we'll need to pass a `can_be_delayed=False` param to the `extract_addon_to_git` task. I tried to keep the lock/unlock in the same task so that if it fails, the add-on is still unlocked at the end.

Locking in this task and unlocking in another task is going to be... interesting.

src/olympia/versions/tasks.py
diox: > Interesting. This line comes from the "migrate to git" command, so does that mean we didn't git-extract the deleted versions?

Yeah, and I don't think that's intended.

src/olympia/versions/tasks.py
diox: Yeah, a big chain is not ideal for sure, I'm open to alternatives :) We definitely can't do everything in a single task...

willdurand: > We definitely can't do everything in a single task...

That, I tend to agree but that's how we did it to migrate _via_ the command, right?

In any case, I think the big chain isn't ideal but sounds a bit better than doing everything in a single task. I can't think of any other way...

src/olympia/versions/tasks.py
diox: Interesting. I am curious to see if we did hit timeouts when we ran the migration... I don't think it's safe to bet that we'll never hit it, but maybe...

src/olympia/versions/tasks.py
diox: I can think of three alternatives... the first two I wouldn't recommend, but here for the sake of the discussion:
- Trigger a single task that processes a single or more versions and then once it's done, triggers a new task to do the next bit.
- Trigger `x` tasks but don't bother with a callback, instead check on status elsewhere in a loop.

My favorite approach:
- Trigger `x` tasks of `y` versions (essentially what you're doing now, except you're processing _all_ versions instead of a fixed `y` number). This is similar to the approach taken by reindex and code search, which both use a `chord` with a callback to signal that it's finished. The trick is to find a reasonable `y` value.

src/olympia/versions/tasks.py
diox: > Ah, right. That would make sense but then we'll need to pass a `can_be_delayed=False` param to the `extract_addon_to_git` task. I tried to keep the lock/unlock in the same task so that if it fails, the add-on is still unlocked at the end.

Yes, and also... when you trigger the task to rebuild the repos entirely for the addon, you shouldn't obey the lock that was set for the extraction of the version in which we noticed the brokenness... But at the same time, you can't release the lock too early (you don't want other version extractions to go through) and you can't completely ignore locks in the rebuild task either (cause you don't want multiple rebuilds in parallel). Maybe you need a second lock mechanism: one that says "a version is being extracted" and a different one that says "the whole addon repos is being rebuilt"... That all depends on how you implement triggering the extractions in the rebuild. 
> 
> Locking in this task and unlocking in another task is going to be... interesting.

willdurand: Yeah, I fixed the easy things so now I am looking into that.

What if we don't delete the repo in the `extract_version_to_git` task and pass `delete_repo=True` to the `extract_addon_to_git` task? Wouldn't that solve our problem here?

diox: It'd certainly be a good thing. But we probably still want the extract version task to not do anything while a full addon extraction (or another version extraction) is underway...

willdurand: So there are two things here:

> we probably still want the extract version task to not do anything while a full addon extraction is underway...

This is why we have the `git_extraction_is_in_progress` lock, no? It is set before we start the full add-on extraction and it is checked in the version task.

> we probably still want the extract version task to not do anything while a version extraction is underway...

That is different. We do not have that right now.

willdurand:
I marked this PR as draft because we want to revisit the entire git extraction process. We'll create new issue(s) to explain our plan.

PR #13941, opened 2020-04-09
Update collection modified when triggering collection_meta task

### LINKED ISSUE #13937: Collections' "Last updated" not working
I updated some collections last week (such as this [one](https://addons.mozilla.org/firefox/collections/4757633/must-have-media/?page=1&collection_sort=-added)) and discovered the "Last updated" date does not change when collections are indeed updated. 

It would be great to get this fixed so users don't feel like we're promoting abandoned collections. 

@muffinresearch

diox:
Includes drive-by micro-optimization that is probably completely unnecessary but I find it makes things easier to read.

Fixes #13937

### DIFF
--- a/src/olympia/bandwagon/tests/test_models.py
+++ b/src/olympia/bandwagon/tests/test_models.py
@@ -61,12 +61,16 @@ class TestCollections(TestCase):
         assert isinstance(c.uuid, uuid.UUID)
 
     def test_collection_meta(self):
-        c = Collection.objects.create(author=self.user)
-        assert c.addon_count == 0
-        c.add_addon(Addon.objects.all()[0])
+        some_time_ago = self.days_ago(442)
+        collection = Collection.objects.create(author=self.user)
+        collection.update(modified=some_time_ago)
+        assert collection.addon_count == 0
+        self.assertCloseToNow(collection.modified, now=some_time_ago)
+        collection.add_addon(Addon.objects.all()[0])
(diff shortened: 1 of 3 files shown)


src/olympia/bandwagon/tests/test_models.py
```diff
@@ -61,12 +61,16 @@ def test_auto_uuid(self):
...
 
     def test_collection_meta(self):
-        c = Collection.objects.create(author=self.user)
-        assert c.addon_count == 0
-        c.add_addon(Addon.objects.all()[0])
+        some_time_ago = self.days_ago(442)
+        collection = Collection.objects.create(author=self.user)
+        collection.update(modified=some_time_ago)
+        assert collection.addon_count == 0
+        self.assertCloseToNow(collection.modified, now=some_time_ago)
```
bobsilverberg, line 68: Nit: This is a cool function, but why is it needed? Wouldn't `collection.modified` be equal to `some_time_ago` at this point?

diox: It would, but I'm being paranoid because `modified` is a bit special. I initially used `Collection.objects.create(author=self.user, modified=some_time_ago)` but this does *not* work as you'd think - `modified` is set to more recent date.

diox:
I noticed a problem with my changes: if all add-ons are removed, then the task no longer resets the addon count to 0. I will revert back the drive-by optim that is causing this and add an additional test.

diox:
46e11b6ca63ba414fd325246ea14f09f671a8a9f

Issue #13851, opened 2020-03-30
Hide "wat" and "mad" options in the list of scanners to create a new rule


willdurand:
Given that we do not have rules for the "wat" and "mad" scanners, let's hide them in the list of scanners displayed in the form of a scanner result:

<img width="486" alt="Screen Shot 2020-03-30 at 11 41 53" src="https://user-images.githubusercontent.com/217628/77898529-9257e880-727b-11ea-85af-e11f6b7a9815.png">

**For QA:** we should not see "wat" and "mad" in the list displayed above when this issue is fixed.

Note: I suppose we can use this: https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.formfield_for_choice_field but I am not sure.

avgupt:
@willdurand  can you please help me with how I could navigate to this page?

willdurand:
> @willdurand can you please help me with how I could navigate to this page?

You should go to the django admin: http://olympia.test/en-US/admin/models/, then `Scanners rules` and click the `+ Add` button.

AlexandraMoga:
The select field for scanner rule creation will include only `yara` and `customs` now:

![image](https://user-images.githubusercontent.com/31961530/80087488-cf09bd80-8563-11ea-91fb-710181b5d05a.png)

@willdurand This patch introduces a small regression for scanner query rule creations.  In https://github.com/mozilla/addons-server/pull/13430, scanner query rules were restricted to show only `yara`, but now they also have `customs` in the select list, same as regular scanner rules. 

I can file a follow so that can be fixed (again).

willdurand:
> @willdurand This patch introduces a small regression for scanner query rule creations. In #13430, scanner query rules were restricted to show only `yara`, but now they also have `customs` in the select list, same as regular scanner rules.
> 
> I can file a follow so that can be fixed (again).

Yes, you can file an issue. I am not sure why this is different now, though.

PR #13949, opened 2020-04-09
Hide some scanner names in the scanner rule form

avgupt:
Fixes: https://github.com/mozilla/addons-server/issues/13851

      PRs open for the same issue.


src/olympia/scanners/models.py
```diff
@@ -192,7 +193,10 @@ class AbstractScannerRule(ModelBase):
         max_length=200,
         help_text=_('This is the exact name of the rule used by a scanner.'),
     )
-    scanner = models.PositiveSmallIntegerField(choices=SCANNERS.items())
+    select_scanner = SCANNERS
+    del select_scanner[WAT]
+    del select_scanner[MAD]
+    scanner = models.PositiveSmallIntegerField(choices=select_scanner.items())
```
eviljeff, line 199: You can't do this - `select_scanner = SCANNERS` doesn't create a copy of `SCANNERS`, you're modifying the original dict, which will affect all other code that references the same constants (and because the server instances are reused, other requests/users too).  

A generator function would work e.g. `[name: scanner for name, scanner in SCANNERS.items() if name not in [WAT, MAD]]`.  Anyway though, changing the model is different to changing `formfield_for_choice_field` like @willdurand suggested - would this work @willdurand or does the model need to stay as it is?

diox: IMHO the model need to stay as it is. Those scanners _are_ valid choices, we just don't want them to show up in the rule creation at the moment.

willdurand: > would this work @willdurand or does the model need to stay as it is?

No, as @diox said, we only want to hide these choices in the form but they are still valid choices so we shouldn't change the model.

avgupt:
@willdurand is this the right way? I tried a lot, but wasn't able to hide the choices using django so used JS instead.

willdurand:
> @willdurand is this the right way? I tried a lot, but wasn't able to hide the choices using django so used JS instead.

I think @eviljeff suggested https://docs.djangoproject.com/en/3.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.formfield_for_choice_field, so I'd start with that first and if that does not work, sure some JS would be OK.

eviljeff:
JS is last reset for django admin - and I'm almost certain it's not necessary for this.  `formfield_for_choice_field` should work, or even just override `get_form` and change `form.base_fields['scanner'].choices`.  See the other `admin.py` files in the repo for examples.

avgupt:
Thanks @willdurand , @eviljeff for pointing me to the right direction.

avgupt:
@willdurand for the test cases, I'm not able to query the form to find the number of choices. Can you please help me with that?

avgupt:
> ```diff
> +    def test_create_form_filters_list_of_scanners(self):
> +        url = reverse('admin:scanners_scannerrule_add')
> +        response = self.client.get(url)
> +        select = pq(response.content)('#id_scanner')
> +        assert len(select.children()) == 3
> ```

Ah, thanks. I was stuck with using get_form() somehow to get the form elements.

src/olympia/scanners/admin.py
```diff
@@ -612,6 +614,14 @@ class AbstractScannerRuleAdminMixin(admin.ModelAdmin):
     )
     readonly_fields = ('created', 'modified', 'matched_results_link')
 
+    def formfield_for_choice_field(self, db_field, request,  **kwargs):
+        if db_field.name == "scanner":
+            kwargs['choices'] = (("","---------"),)
+            for key, value in SCANNERS.items():
```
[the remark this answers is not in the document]
avgupt: I generally choose a more elaborate way to write code as I think it's generally easier to read and understand for other developers (atleast for me :P ).

willdurand:
Awesome, thanks @avgupt! /cc @caitmuenster

PR #13971, opened 2020-04-14
Update filtercascade from 0.3.1 to 0.4.0

eviljeff:
Not sure why pyup hasn't picked this up, but needed some test fixes anyway.

### DIFF
--- a/src/olympia/blocklist/tests/test_mlbf.py
+++ b/src/olympia/blocklist/tests/test_mlbf.py
@@ -98,7 +98,7 @@ class TestMLBF(TestCase):
     def test_get_blocked_versions(self):
         blocked_versions = MLBF.get_blocked_versions()
         blocked_guids = blocked_versions.values()
-        assert len(blocked_guids) == 5
+        assert len(blocked_guids) == 5, blocked_guids
         assert (self.five_ver.guid, '123.40') in blocked_guids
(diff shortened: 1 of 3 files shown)


src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -98,7 +98,7 @@ def test_all_guids(self):
     def test_get_blocked_versions(self):
         blocked_versions = MLBF.get_blocked_versions()
         blocked_guids = blocked_versions.values()
-        assert len(blocked_guids) == 5
+        assert len(blocked_guids) == 5, blocked_guids
```
eviljeff, line 101: A drive-by change because this is an intermittent ci failure and it'd be useful to know what the guids where when it failed.

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -155,10 +155,10 @@ def test_generate_mlbf(self):
             assert key not in bfilter
         assert stats['mlbf_version'] == 2
         assert stats['mlbf_layers'] == 1
-        assert stats['mlbf_bits'] == 21608
+        assert stats['mlbf_bits'] == 2160
```
eviljeff, line 158: fwiw, the bit counts and file sizes are decreasing because 0.4.0 reduced some of the minimum defaults and these are very small filters with only a few blocks.

PR #13986, opened 2020-04-16
upload stash files to kinto

### LINKED ISSUE #13751: `test_get_blocked_guids` intermittent test failure
In Travis-CI:

```
src/olympia/blocklist/tests/test_mlbf.py:66: in test_get_blocked_guids

    assert len(blocked_guids) == 5

E   AssertionError: assert 6 == 5
E     -6
E     +5
```

### LINKED ISSUE #13762: implement Block stashing so we don't need a new MLBF each time
Similar to what crlite is doing (mozilla/crlite#53 mozilla/crlite#48) we should implement some sort of stash list of new blocks so we don't have to generate a new bloom filter - and the client doesn't have to download an entire new filter - each time a new block is added.  

things tbd:
 - How often we should build a new filter  (every x days?  every x new blocks?)
 - Should we still look to build a new filter, or update the stash list, on a 6 hour schedule or on demand each time a Block is created/updated?
 - Format: crlite creates a binary format blob containing the new items and uploads it as an attachment every day.  IMO json is more straightforward (at the cost of only a few extra bytes for `"`s and `,`s) - we could upload that either as an attachment or directly as data on the record.  Alternatively could have a record per stashed block, at the cost of extra overhead per record.
- storage location.  crlite places both the filter and the stash blobs in the same collection so the client can just download (or diff) the entrire collection.
- stash list deletion policy.  crlite's strategy is to clear the collection - both stashed items and previous filter - each time a new filter is available.

I guess we'll have to always rebuild the filter for Block deletions (and we'll need to take care with Block updates if the min/max versions change) unless we want a un-stash list too(!)

eviljeff:
fixes #13762 (+ should hopefully fix #13751 if I'm lucky)

### DIFF
--- a/src/olympia/blocklist/tasks.py
+++ b/src/olympia/blocklist/tasks.py
@@ -32,7 +32,7 @@ MLBF_TIME_CONFIG_KEY = 'blocklist_mlbf_generation_time'
 MLBF_BASE_ID_CONFIG_KEY = 'blocklist_mlbf_base_id'
 
 BLOCKLIST_RECORD_MLBF_BASE = 'bloomfilter-base'
-BLOCKLIST_RECORD_MLBF_UPDATE = 'bloomfilter-update'
+BLOCKLIST_RECORD_MLBF_UPDATE = 'bloomfilter-full'
 
 
 @task
--- a/src/olympia/blocklist/tasks.py
+++ b/src/olympia/blocklist/tasks.py
@@ -29,6 +29,11 @@ bracket_open_regex = re.compile(r'(?<!\\){')
 bracket_close_regex = re.compile(r'(?<!\\)}')
 
 MLBF_TIME_CONFIG_KEY = 'blocklist_mlbf_generation_time'
+MLBF_BASE_ID_CONFIG_KEY = 'blocklist_mlbf_base_id'
+
+BLOCKLIST_RECORD_MLBF_BASE = 'bloomfilter-base'
+BLOCKLIST_RECORD_MLBF_UPDATE = 'bloomfilter-update'
+BLOCKLIST_RECORD_STASH = 'stash'
 
 
 @task
@@ -143,16 +148,37 @@ def import_block_from_blocklist(record):
 
 
 @task
-def upload_filter_to_kinto(generation_time):
+def upload_filter_to_kinto(generation_time, is_base=True, upload_stash=False):
     server = KintoServer(
         KINTO_BUCKET, KINTO_COLLECTION_MLBF, kinto_sign_off_needed=False)
+    mlbf = MLBF(generation_time)
+    # Deal with possible stashes first
+    if upload_stash:
+        # If we have a stash, write the attachment for that
+        stash_data = {
+            'key_format': MLBF.KEY_FORMAT,
+            'stash_time': generation_time,
+            'attachment_type': BLOCKLIST_RECORD_STASH,
+        }
+        with storage.open(mlbf.stash_path, 'r') as stash_file:
+            attachment = ('stash.json', stash_file)
+            server.publish_attachment(stash_data, attachment)
+    if is_base:
+        # clear the collection for the base - we want to be the only filter
+        server.delete_all_records()
+
+    # Then the bloomfilter
     data = {
         'key_format': MLBF.KEY_FORMAT,
         'generation_time': generation_time,
+        'attachment_type':
+            BLOCKLIST_RECORD_MLBF_BASE if is_base else
+            BLOCKLIST_RECORD_MLBF_UPDATE,
     }
-    mlbf_path = MLBF(generation_time).filter_path
-    with storage.open(mlbf_path) as filter_file:
+    with storage.open(mlbf.filter_path, 'rb') as filter_file:
         attachment = ('filter.bin', filter_file, 'application/octet-stream')
         server.publish_attachment(data, attachment)
     server.complete_session()
     set_config(MLBF_TIME_CONFIG_KEY, generation_time, json_value=True)
+    if is_base:
+        set_config(MLBF_BASE_ID_CONFIG_KEY, generation_time, json_value=True)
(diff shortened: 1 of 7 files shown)


eviljeff:
@Rob--W - the changes in tasks.py are probably the ones that most affect Firefox

src/olympia/blocklist/tasks.py
```diff
@@ -143,16 +148,37 @@ def import_block_from_blocklist(record):
...
+            'key_format': MLBF.KEY_FORMAT,
+            'stash_time': generation_time,
+            'attachment_type': BLOCKLIST_RECORD_STASH,
+        }
+        with storage.open(mlbf.stash_path, 'r') as stash_file:
+            attachment = ('stash.json', stash_file)
+            server.publish_attachment(stash_data, attachment)
+    if is_base:
+        # clear the collection for the base - we want to be the only filter
+        server.delete_all_records()
```
Rob--W, line 168: If `is_base == upload_stash == True`, then you would be uploading the stash and immediately delete it. I think that you have to delete before pushing any new records/attachments.

src/olympia/blocklist/tasks.py
```diff
@@ -143,16 +148,37 @@ def import_block_from_blocklist(record):
...
+    if upload_stash:
+        # If we have a stash, write the attachment for that
+        stash_data = {
+            'key_format': MLBF.KEY_FORMAT,
+            'stash_time': generation_time,
+            'attachment_type': BLOCKLIST_RECORD_STASH,
+        }
+        with storage.open(mlbf.stash_path, 'r') as stash_file:
+            attachment = ('stash.json', stash_file)
+            server.publish_attachment(stash_data, attachment)
```
Rob--W, line 165: Can you publish the JSON directly instead of a file?

Using multiple attachments makes the client implementation more complicated. The [client library](https://searchfox.org/mozilla-central/rev/97cb0a90bd053de87cd1ab7646d5565809166bb1/services/settings/Attachments.jsm#62-74) takes the filename and overwrites existing files with the same name, so I would have to implement my own storage mechanism to avoid this issue.

src/olympia/blocklist/tasks.py
eviljeff: Ah, you're right.  In current use `upload_stash` is always `not is_base` so it wouldn't happen but I should fix that.

src/olympia/blocklist/tasks.py
eviljeff: We could - I had them as attachments to cover the possible use-case (e.g. ESR) of turning off stash support so it wouldn't be downloading unnecessary data in the records.  If we're pretty confident that won't happen (and/or don't care about the extra download data if it does) then it can be directly in the data block.

eviljeff: https://github.com/mozilla/addons-server/pull/13986/commits/a79c82f67ed89a36986fc2acd872e95db48ffc34

src/olympia/blocklist/tasks.py
eviljeff: https://github.com/mozilla/addons-server/pull/13986/commits/a79c82f67ed89a36986fc2acd872e95db48ffc34

src/olympia/blocklist/tasks.py
```diff
@@ -143,16 +147,35 @@ def import_block_from_blocklist(record):
...
 @task
-def upload_filter_to_kinto(generation_time):
+def upload_filter_to_kinto(generation_time, is_base=True, upload_stash=False):
     server = KintoServer(
         KINTO_BUCKET, KINTO_COLLECTION_MLBF, kinto_sign_off_needed=False)
+    mlbf = MLBF(generation_time)
+    if is_base:
+        # clear the collection for the base - we want to be the only filter
+        server.delete_all_records()
+    # Deal with possible stashes first
```
Rob--W, line 157: I'm a bit confused by the implementation.

I am expecting a fixed MLBF base plus some (none, one or more) stashes. And maybe one continuously updating MLBF, if we want to disable stashes.

The current implementation always published a new MLBF. So the collection may look like.

```
{ stash_time: 1, ... }, (or maybe not, if upload_stash=False when is_base=True)
{ stash_time: 2, ... },
{ stash_time: 3, ... },
{ stash_time: 4, ... },
{ generation_time: 1, attachment_type: "bloomfilter-base", attachment: ... }, 
{ generation_time: 2, attachment_type: "bloomfilter-update", attachment: ... }, 
{ generation_time: 3, attachment_type: "bloomfilter-update", attachment: ... }, 
{ generation_time: 4, attachment_type: "bloomfilter-update", attachment: ... }, 
```

This has too many bloomfilter-update attachments. I would expect something like this:

```
{ stash_time: 2, ... },
{ stash_time: 3, ... },
{ stash_time: 4, ... },
{ generation_time: 1, attachment_type: "bloomfilter-base", attachment: ... }, 
{ generation_time: 4, attachment_type: "bloomfilter-full", attachment: ... },
```
Note: "full" instead of "update" to better reflect its meaning.

eviljeff: The pre-stash client side implementation mentioned previously (if I understood it correctly) was to look at all records that had `generation_time` and just download the attachment from the one with the latest `generation_time`.  In that implementation it wouldn't matter how many MLBF attachments there were in the collection.  Has that been changed?

We could keep track of the `bloomfilter-update`/`bloomfilter-full` records and update them I suppose, I just didn't see a need to.  (I'll do it as a follow-up issue - this pr has enough moving parts as it is)

Rob--W: > The pre-stash client side implementation mentioned previously (if I understood it correctly) was to look at all records that had `generation_time` and just download the attachment from the one with the latest `generation_time`. In that implementation it wouldn't matter how many MLBF attachments there were in the collection. Has that been changed?

From the client's perspective it doesn't matter. Avoiding duplication could indeed be done in a follow-up.

I still suggest to go with the "-full" suffix right away rather than "-update", because "-full" more accurately represents the meaning of the MLBF.

eviljeff: https://github.com/mozilla/addons-server/pull/13986/commits/f5d3fb70189c0dcdfe2704ef26516d9f8c7eb8e3 and #14013 for follow-up

PR #13987, opened 2020-04-16
Add external variables when matching yara rules 

### LINKED ISSUE #13966: Allow including or excluding files in code search
It would be helpful to admins if there was a way to include and exclude files or file patterns during the definition of a code search. This could speed up performance and reduce load.

diox:
Fixes #13966

### DIFF
--- a/src/olympia/scanners/tasks.py
+++ b/src/olympia/scanners/tasks.py
@@ -229,8 +229,12 @@ def _run_yara_for_path(scanner_result, path, definition=None):
                     scanner=YARA, is_active=True, definition__isnull=False
                 ).values_list('definition', flat=True)
             )
-
-        rules = yara.compile(source=definition)
+        # Initialize external variables so that compilation works, we'll
+        # override them later when matching.
+        externals = {
+            'filename': '',
+        }
+        rules = yara.compile(source=definition, externals=externals)
 
         zip_file = SafeZip(source=path)
         for zip_info in zip_file.info_list:
@@ -238,8 +242,11 @@ def _run_yara_for_path(scanner_result, path, definition=None):
                 file_content = zip_file.read(zip_info).decode(
                     errors='ignore'
                 )
-                for match in rules.match(data=file_content):
-                    # Add the filename to the meta dict.
+                # Add filename to externals, making it available in the rule.
+                externals['filename'] = zip_info.filename
+                for match in rules.match(
+                        data=file_content, externals=externals):
+                    # Also add the filename to the meta dict in results.
                     meta = {**match.meta, 'filename': zip_info.filename}
                     scanner_result.add_yara_result(
                         rule=match.rule,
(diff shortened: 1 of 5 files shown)


src/olympia/scanners/tasks.py
```diff
@@ -229,17 +229,24 @@ def _run_yara_for_path(scanner_result, path, definition=None):
                     scanner=YARA, is_active=True, definition__isnull=False
                 ).values_list('definition', flat=True)
             )
-
-        rules = yara.compile(source=definition)
+        # Initialize external variables so that compilation works, we'll
+        # override them later when matching.
+        externals = {
+            'filename': '',
```
willdurand, line 235: I don't think this is necessarily a good idea, especially given that the real use case is different than what's described in the issue. The use case seems to be about detecting `manifest.json` and locale files.

While the former can be retrieved easily, it is definitely tricky for locale files but we could do something a bit different. We could inject three variables instead of the filename (because we should really write rules that apply to the content of a file and not based on its name, which can be anything).

How about we inject the following instead?

```
'is_json_file': zip_info.filename.endswith('.json'),
'is_manifest': zip_info.filename == 'manifest.json',  # I guess there is a better way to compare strings in python
'is_locale_file': zip_info.filename.contains('_locales') and zip_info.filename.endswith('messages.json'),  # there is probably a better way to write that, see: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Internationalization#Localized_string_selection
```

diox: https://github.com/mozilla/addons-server/pull/13987/commits/6c5391f3a3a922b0a0163ae9e641cba4f4aa2451 does that. I am not a super fan of how rigid this is - that's why I originally went for just `filename` - but we can try this out and see if that covers our needs.

src/olympia/scanners/tasks.py
```diff
@@ -229,18 +229,33 @@ def _run_yara_for_path(scanner_result, path, definition=None):
                     scanner=YARA, is_active=True, definition__isnull=False
                 ).values_list('definition', flat=True)
             )
-
-        rules = yara.compile(source=definition)
+        # Initialize external variables so that compilation works, we'll
+        # override them later when matching.
+        externals = {
```
willdurand, line 234: It would be nice if we could export the keys of this dict and add something similar to a `help_text` in the form when creating a Yara rule. I don't know how to do it, though :p

diox: Mmm I don't know either but looking at the `models.py` code looks like I should have modified `clean_yara()` as well since it uses `compile()` without referencing the `externals`...

willdurand: oh right..

diox: b6414b6a9aed7ebad5f41692436a82678b06c86d fixes that and 8fdfc439e945b8d9e6aa10f88664cac780c4d843 adds a hint (not as a `help_text`, and not DRY, but should be good enough)

willdurand: ah yes, smart. I like it!

PR #14003, opened 2020-04-17
Upgrade docker image to latest Debian Buster

### LINKED ISSUE #13985: Upgrade docker image to latest Debian stable
Our `Dockerfile` and `Dockerfile.deploy` are built from `stretch` which is the old debian stable. We should upgrade to `buster`, which is the latest stable. That will allow us to remove this hack where we install a couple packages from `buster` while being on `stretch`:
```
# Install `file` and `libmagic` from the `buster` repositories for an up-to-date
# file-detection.
RUN apt-get update && apt-get -t buster install -y \
        file \
        libmagic-dev \
```

championshuttler:
Fixes #13985

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6-slim-stretch
+FROM python:3.6-slim-buster
 
 ENV PYTHONDONTWRITEBYTECODE=1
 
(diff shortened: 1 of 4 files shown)


diox:
This fails to build locally (`docker build . -t addons/addons-server:latest`) for at least 2 reasons:
- `buster` repos is fetched twice (also need to remove the line adding that repos, since it will already be present)
`mysql-client` has no installation candidate. Can probably be replaced by `mysql-client-5.7`.

championshuttler:
>     * `mysql-client` has no installation candidate. Can probably be replaced by `mysql-client-5.7`.

@diox I dont think thats possible https://packages.debian.org/search?searchon=names&keywords=mysql-client

Also getting hit by this error if i go with `mysql-client-5.7`

```
Package mysql-client-5.7 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  mariadb-server-core-10.3 mariadb-client-10.3
``` 

May be `default-mysql-client` we can use, https://packages.debian.org/buster/default-mysql-client, can you please confirm it?

diox:
mmm... yeah, we'd have to switch to `default-mysql-client`. It gives us `mariadb-client-10.3` which in theory should be fully compatible from a client's perspective. I think we'll want to try that out on dev for a week before making the switch to stage and then prod, so we'd have to wait until after the tag on tuesday to merge this.

bqbn:
The app itself doesn't depend on mysql client, yes? 

I remember it was only for command line need, i.e. someone running command line tool inside the container to connect to the database, but I could  be wrong that some python library depends on it.

diox:
Actually yes - we depend on `default-libmysqlclient-dev` for `mysqlclient` python library, but that was already depending on `mariadb` implementation in `stretch`. `default-mysql-client` just changes the CLI so yeah, we should be fine, even more than I thought we were :)

PR #14004, opened 2020-04-17
Footer links headers should use the same font style

### LINKED ISSUE #13970: Footer links headers should use the same font style
Follow up for #13656 

1. Open Developer Hub homepage - https://addons-dev.allizom.org/en-US/developers/
2. Scroll down to the footer section
3. Notice the font style used for all three Footer-links-headers

### What happened?
Browsers and Products are using a lighter font weight compared to Add-ons

All headers should have the same font style

[Actual]
image

[Expected]
image

championshuttler:
Fixes #13970

### DIFF
--- a/static/css/photon-site/footer.less
+++ b/static/css/photon-site/footer.less
@@ -84,6 +84,7 @@
 
 .Footer-links-header {
     margin: 40px 0 0;
+    font-weight: 700;
 
     @media @large {
       margin: 0;


static/css/photon-site/footer.less
```diff
@@ -84,6 +84,7 @@
 
 .Footer-links-header {
     margin: 40px 0 0;
+    font-weight: 700;
```
eviljeff, line 87: is the `font-weight: 700` a few lines below inside `a:link, a:visited` obsolete now?  (You'll have to try it out as css priorities are weird)

championshuttler, line 87: @eviljeff Thats not obsolete because thats for the headers with some link for example the first section header `Addons` is linked to `addons.mozilla.org`, but `Browsers` and `Products` are not linked to anything so need to add `font-weight: 700` explicitely

championshuttler: @eviljeff Thats not obsolete because thats for the headers with some link for example the first section header Addons is linked to addons.mozilla.org, but Browsers and Products are not linked to anything so need to add font-weight: 700 explicitely

championshuttler:
image

static/css/photon-site/footer.less
eviljeff: yes, I understand why we need the new statement under `.Footer-links-header`, I was just wondering if the statement below was obsolete because `a:link, a:visited` inherits from `.Footer-links-header`.  It's not worth bothering about though.

Issue #14016, opened 2020-04-20
Add a model that stores add-on IDs for git extraction


willdurand:
Supports https://github.com/mozilla/addons-server/issues/13964

---

We need a model that acts as a queue of add-on IDs so that our CRON task knows which add-ons to process.

Issue #14015, opened 2020-04-20
Add a model to lock an add-on during git-extraction


willdurand:
Supports https://github.com/mozilla/addons-server/issues/13964

---

We need a way to lock an add-on when we are extracting its files into git. Let's have a 1-1 model for that.

willdurand:
We don't need this. We can merge this model with the one for https://github.com/mozilla/addons-server/issues/14016

Issue #13964, opened 2020-04-14
Run git extraction in a CRON (similar to `auto_approve`)


willdurand:
Refs https://github.com/mozilla/addons-server/issues/13590

---

While discussing about https://github.com/mozilla/addons-server/pull/13940, @diox and I decided to step back and see how we could avoid further issues and solve existing ones.

We believe that performing the git extraction in a CRON task would allow us to:

1. make sure versions are extracted in the right order
2. more easily re-extract all the versions of an add-on
3. recover from errors like https://github.com/mozilla/addons-server/issues/13590

Similar to `auto_approve`, we'll have a command scheduled every X minutes (5 minutes maybe). We'll use a FS lock to make sure the command is not run more than once at a time.

This command will look into a "queue" of add-on IDs to git-extract. For each add-on ID, the command will retrieve the versions that are not committed in the git repository yet (using the `git_hash` field on each version) and then extract these versions sequentially. The command will lock and then unlock the add-on once these versions have been extracted.

In https://github.com/mozilla/addons-server/pull/13940, we introduced a database lock when we re-extract all the versions of an add-on. We will keep this idea of a DB lock to guarantee the order of commits in the git repo of an add-on. If something goes terribly wrong (like a Celery worker that dies), we'll be able to unlock the add-on by deleting its DB lock.

If an add-on whose ID is in the "queue" is locked, then we won't remove the ID from the queue and next time the command runs, it will try to extract the non-extracted versions for this add-on.

During the submission process, instead of performing the git extraction, we'll push the add-on ID to the queue and that's it. An add-on ID can be present more than once in this queue.

If we detect an error like https://github.com/mozilla/addons-server/issues/13590, we can delete the repository (see #13963) and push the add-on ID to the queue. Next time the command runs, the add-on will be re-extracted.

willdurand:
@diox did I miss anything? Unfortunately, I did not take as much notes as I wished..

diox:
To extract extract versions sequentially, we talked about using a celery task chain with an unlock task at the end, similar to how it's done for code search. But that's an implementation detail (that we may or may not end up doing).

We also discussing needing a simple admin tool to view the queue and what's locked, so that we can force an extraction to happen again by pushing an add-on to the queue manually, and remove a lock if something went wrong.

bobsilverberg:
> 
[...]

PR #14017, opened 2020-04-20
Add new `GitExtractionEntry` model (in new 'git' app)

willdurand:
~~Fixes https://github.com/mozilla/addons-server/issues/14015~~
Fixes https://github.com/mozilla/addons-server/issues/14016

---

This patch has been (partially) extracted from https://github.com/mozilla/addons-server/pull/13940. I created a new 'git' app because more code will be added into it for https://github.com/mozilla/addons-server/issues/13964 (new models, tasks, migrations). The existing code for the git extraction feature is located in different places so my idea would be to re-centralize all the git parts into this new app.


src/olympia/lib/settings_base.py
```diff
@@ -477,6 +477,7 @@ def get_db_config(environ_var, atomic_requests=True):
     'olympia.devhub',
     'olympia.discovery',
     'olympia.files',
+    'olympia.git',
```
eviljeff, line 480: I'm not sure about creating a new top level app for this - its more an interface to git so `olympia.lib.git` seems more appropriate to me... wdyt @diox?

src/olympia/git/models.py
```diff
@@ -0,0 +1,26 @@
...
+from olympia.amo.models import ModelBase
+
+
+class AddonGitExtraction(ModelBase):
+    """
+    This is an add-on related model that stores information related to the git
+    extraction (for code-manager).
+    """
+
+    addon = models.OneToOneField(
```
eviljeff, line 13: isn't git extraction a thing for each new version?  If so you'd need the fk to `Version` or `File` instead, right?

src/olympia/git/models.py
```diff
@@ -0,0 +1,26 @@
...
+        Addon, on_delete=models.CASCADE, primary_key=True
```
eviljeff, line 14: mmm, `primary_key`, are you sure?

src/olympia/git/models.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+    in_progress = models.BooleanField(default=False)
+
+
+class GitExtractionEntry(ModelBase):
```
eviljeff, line 19: Actually, I'm not sure why we have two models that are both fks to `Addon` and both for git extraction.  How would adding the `in_progress` field to this model instead change things?

src/olympia/git/models.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    This is a model that represents an entry in a "queue" of add-ons scheduled
+    for git extraction. When an add-on is in this queue, its versions should be
+    extracted to a git repository.
+    """
+
+    addon = models.ForeignKey(Addon, on_delete=models.CASCADE, null=True)
```
eviljeff, line 26: when would this fk be `null=True`?

src/olympia/addons/models.py
```diff
@@ -1534,6 +1534,12 @@ def blocklistsubmission(self):
         # GUIDs should only exist in one (active) submission at once.
         return BlocklistSubmission.get_submissions_from_guid(self.guid).last()
 
+    @property
+    def git_extraction_is_in_progress(self):
+        if not hasattr(self, 'addongitextraction'):
+            return False
+        return self.addongitextraction.in_progress
```
EnTeQuAk, line 1541: While I do see numerous advantages of using the database for locking, we'll have to make sure that every query checking the "lock" is going to the master to avoid replication lag and ideally it'll play together with the filesystem locking that we already set up in https://github.com/mozilla/addons-server/blob/77e8b8d41e5661029e0c8ebfed535e98f0a98308/src/olympia/files/utils.py#L1354-L1412 even though it's not fully being used by git-extraction yet.

I haven't gone through all the issues you filed for your work though so I may miss some context

willdurand:
> It's ~difficult~impossible to tell how suitable these models are without any code that uses it. I'm not sure this is a practical way of developing either - personally I always end up changing the model numerous times while I'm adding a new a feature.

Sorry, the next patch is here: https://github.com/mozilla/addons-server/pull/14021 and these models have been extracted from it.

src/olympia/lib/settings_base.py
willdurand: As I said, we'll need more code like new tasks and a new admin so I think that would make sense. And I rather see `lib.git` moved to `git.utils`.

src/olympia/git/models.py
willdurand: Git extraction extracts each version of an add-on but we want to lock the full add-on to prevent issues (like wrong order of the versions in git or concurrent extraction).

src/olympia/git/models.py
willdurand: it was suggested here https://github.com/mozilla/addons-server/pull/13940#discussion_r406134574

src/olympia/git/models.py
willdurand: The other model is there to avoid adding a new field to the `Addon` model (1-1) while this model (`GitExtractionEntry`) can have the same add-on multiple times and we create and delete entries in this table (that acts as a sort of queue).

src/olympia/git/models.py
willdurand: This is probably not needed, indeed.

src/olympia/addons/models.py
willdurand: All the tasks involving this lock will have the `@use_primary_db` decorator. We'll also use the FS lock for the CRON task (to prevent this task to run more than once at the same time).

src/olympia/git/models.py
eviljeff: There would be multiple instances of the same add-on in the table at the same time?  Why would you need that?  (surely the first time would extract all the versions)

willdurand: There is no guarantee that all the versions will be extracted by the same CRON task execution. Because the CRON is scheduled at a fixed interval and submissions are random, we could likely have a scenario where a new version is submitted after the CRON retrieved the versions to extract.

eviljeff: I've been thinking the design over and I still don't think two models are required.  If `in_progress` was added to this model then the only change (that I can see) would be change to `Addon.git_extraction_is_in_progress` to

```
    def git_extraction_is_in_progress(self):
        return GitExtractionEntry.objects.filter(addon=self, in_progress=True).exists()
```

I understand this came off the back of the pr discussions with @diox though and I've not trawled through all of that (I made it half way through but there were too many collapsed threads!)

willdurand: Here is a scenario:

1. the `git_extraction` CRON task starts 10 tasks to extract the different versions from v1 to v10, those versions are big so it takes more than 5 minutes to extract them
2. user submits new version: v11
3. v1 to v6 have been extracted so far
4. the `git_extraction` CRON task starts again

How do we prevent the second set of tasks to no run before the task for v10? To me, the answer is: locking the add-on so that the task is not created for v11. It will eventually be created once the add-on is unlocked.

So now, how do we remember that v11 needs extraction if we only use a single boolean flag?

This is why, I believe, we need two different models. The `AddonGitExtraction` model is really used to not add new columns to the `addons` table, otherwise the `git_extraction_in_progress` field would live in there. The purpose of this field is to prevent a git extraction of an add-on when one is already in progress, because we want to change our approach from: "let's git extract a version on the fly" to "let's git extract an add-on". In this case, extracting an add-on means extracting all the versions that haven't been extract yet, in the right order.

The `GitExtractionEntry` model is used to add add-on IDs in a table that will be consumed by the CRON task. If it's ok to create a rabbitmq to do this instead, I am fine with it but I don't think we do that.

With these two models, we can make sure that at step (4), the CRON task will skip the add-on because it is locked by `AddonGitExtraction.in_progress` but once this flag is set to `false`. In addition, at step (2) a new `GitExtractionEntry` will be created so that, once the add-on is unlocked, the CRONtask will create a task to extract all the not-yet-extracted versions (v11 here).

That being said, this is coming from my perspective and if we could make the scenario above work with a single model, then I'm fine with that because I don't really mind having one or more models. It's an implementation detail to me. The scenario above is important, though. WDYT?

eviljeff: > How do we prevent the second set of tasks to no run before the task for v10? To me, the answer is: locking the add-on so that the task is not created for v11. It will eventually be created once the add-on is unlocked.

> So now, how do we remember that v11 needs extraction if we only use a single boolean flag?

`GitExtractionEntry.objects.filter(addon=self, in_progress=True).exists()` would be true (via `Addon.git_extraction_is_in_progress`).  You'd only need a separate model if you intended to persist the flag beyond the time the task was waiting to be processed and being processed, but as I'm understanding the code and issues/comments the flag is set back to `in_progress=False` which is the same as not existing.

diox: Yeah I agree with @eviljeff , I think one model is enough. To ensure consistency you'd need an unique constraint on (`addon`, `in_progress`), and default `in_progress` to `null` (`NullBooleanField`) and never set it to `False`, only `True` when an extraction is in progress.

willdurand: I am sorry, I don't understand, and it's certainly on me.

Do you mean not having a field on the `Addon` model (or the 1:1 with `AddonGitExtraction`) and relying on `GitExtractionEntry` to determine when an add-on is being extracted or not?

How would we retrieve the add-ons to process in the CRON task? And how would we mark them as "consumed"? (which means removing them from the queue because the git-extraction tasks have been created and submitted)

> To ensure consistency you'd need an unique constraint on (addon, in_progress)

Wouldn't that fail (the second time) if two versions of the same add-on are submitted while this add-on is being git-extracted?

> and default in_progress to null (NullBooleanField) and never set it to False, only True when an extraction is in progress.

Why never setting it back to `False`?

diox: If you set it to `null` instead of `False` then you can have multiple entries with the same `addon`, the unique constraint would not get in the way. It would fail otherwise.

diox: > How would we retrieve the add-ons to process in the CRON task? And how would we mark them as "consumed"? (which means removing them from the queue because the git-extraction tasks have been created and submitted)

Take row by row, if `GitExtractionEntry.objects.filter(addon=self, in_progress=True).exists()` is `True` then skip it. When extraction is done remove the row you were processing (or just `GitExtractionEntry.objects.filter(addon=addon, in_progress=True).delete()` which should do the same thing since there should only be one extraction in progress per add-on at any given time)

willdurand: Ok, I think I see now. Sorry @eviljeff, if that was what you were saying from the beginning, then obviously I don't know how to read :/

src/olympia/git/migrations/0001_initial.py
```diff
@@ -0,0 +1,58 @@
...
+            options={
+                'get_latest_by': 'created',
+                'abstract': False,
+                'base_manager_name': 'objects',
+                'unique_together': {('addon', 'in_progress')},
+            },
+            bases=(
+                olympia.amo.models.SearchMixin,
+                olympia.amo.models.SaveUpdateMixin,
+                models.Model,
```
eviljeff, line 55: has this been run through a prettifier?  it looks non-django-migrations-y.  (which I guess is fine?  We exclude migrations from flake8 generally.  If you had to manually tinker with it can you call out your changes).

willdurand: Indeed, it's Black 🙈

PR #14021, opened 2020-04-20
Add a new `git_extraction` command

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13964
~~:warning: depends on https://github.com/mozilla/addons-server/pull/14017~~

---

This is the first working iteration for https://github.com/mozilla/addons-server/pull/14017. There is no django admin yet and there is no fix for broken repositories. Those two features will be added later. ~~In addition, this first iteration reuses the existing task and does not create tasks with subsets of version IDs (slicing).~~ => I added one more commit for that

### Local setup

The following waffle switch should be created and active:

```
enable-uploads-commit-to-git-storage: on
enable-git-extraction-cron: on
```

`enable-git-extraction-cron` is the switch that enabled the feature in this patch.

### Known knowns

- I am not sure why or if this is real but it looks like Celery calls the error handler twice when a task in the chain fails. One can easily replicate a task failure by emptying a branch file in a git repo, cf. https://github.com/mozilla/addons-server/issues/13590#issuecomment-611099846

- Initially, I was using the `remove` task as error handler as well but it is confusing so that's why we have two tasks now.

- In the added test cases, one of them replicates the issue mentioned in #13590. For some reason, Celery in eager mode raises the underlying error, even with `CELERY_TASK_EAGER_PROPAGATES` set to `False`. If we don't set `CELERY_TASK_EAGER_PROPAGATES` to `False`, the error handler is not called in eager mode so we have to set it to `False`.

- ~~I chose to create and then delete `GitExtractionEntries`. We could soft delete instead (or add a boolean field) but then we would need to clean up this table on a regular basis.~~ that was also suggested in https://github.com/mozilla/addons-server/pull/14017

- We lose the `author_id` and `note` information. I am not sure if it matters a lot, though. It's tricky to re-add because those values are bound to a specific version and we switch to an add-on based extraction here.


src/olympia/git/management/commands/git_extraction.py
```diff
@@ -0,0 +1,109 @@
...
+        This method takes a GitExtractionEntry object and creates a chain of
+        Celery tasks to extract each version in a git repository that haven't
+        been extracted yet (including the deleted versions).
+
+        It does not run if the add-on is locked for git extraction.
+        """
+        addon = entry.addon
+        log.info('Starting extraction of add-on "{}".'.format(addon.pk))
+
+        if addon.git_extraction_is_in_progress:
```
willdurand, line 65: I thought we could remove this and use `entry.in_progress` but that didn't work (and I added a test case for that). We could use the model directly here but I find the call to this property more readable (and we could use the property elsewhere).

willdurand:
I just rebased this patch.

src/olympia/git/management/commands/git_extraction.py
diox: Add a comment explaining why we can't use `entry.in_progress` here :)

scripts/crontab/crontab.tpl
```diff
@@ -7,6 +7,7 @@ HOME=/tmp
 
 # Every 5 minutes
 */5 * * * * %(django)s auto_approve
+*/5 * * * * %(django)s git_extraction
```
diox, line 10: We could consider using something like `1-60/5` for it not to start at exactly the same time as `auto_approve`, just to make it easier to follow logs? Maybe that's completely unnecessary though.

willdurand:
Errors are related to #14057

Issue #14020, opened 2020-04-20
Allow Browse and Compare APIs to return just file data


bobsilverberg:
This is a spin off from #13984 

Code manager only needs version specific information (e.g., file entries) the first time a version is requested, but the current APIs return version info for every request, even when code manager really only needs data for the requested file.

To deal with this we should introduce a new URL param (called `file_only`) for the Browse and Compare  endpoints that can be used to request just file data and not version data.

When `file_only` is set to `true` then only file data is returned to the client. The file data is in a top-level property called `file`, and none of the other `version` properties are returned in the API response.

willdurand:
> To deal with this we should introduce new endpoints that can be used to request just file data for both the browse and compare use cases.

Did we ever consider adding query parameters to include/exclude data instead of creating new endpoints?

ioanarusiczki:
@bobsilverberg I tested in Postman on dev, by sending some requests for the Browse and then the Compare API. 
- the response is more simplified, with `entries:null` and also general information about the addon (including the version number) will be excluded.
Two examples:
https://addons-dev.allizom.org/api/v5/reviewers/addon/502991/versions/1541836/?file_only=true&file=images%2Ficon128_gray.png
https://addons-dev.allizom.org/api/v5/reviewers/addon/619715/versions/1691176/compare_to/1691183/?file_only=true&file=autosave-ui.html

- without a specific `file` added the response will return information about manifest.json (as default)

- if I used `file_only:false` the information in the response returns the same data as if I use `file_only:true`. The complete (original) information is available only if I delete the parameter from request.

https://addons-dev.allizom.org/api/v5/reviewers/addon/502991/versions/1541836/?file_only=false&file=images%2Ficon128_gray.png
https://addons-dev.allizom.org/api/v5/reviewers/addon/619715/versions/1691176/compare_to/1691183/?file_only=false&file=autosave-ui.html

with false 
![fileonlyonfalse](https://user-images.githubusercontent.com/33448286/81410015-c232bf80-9148-11ea-8173-d6b8be03dfc7.png)

with true 
![true](https://user-images.githubusercontent.com/33448286/81410384-4ab16000-9149-11ea-9a04-dd58750c4f29.png)


  If I should file an issue about this let me know. Thanks!

bobsilverberg:
@ioanarusiczki Yes, please file an issue about the `false` param not working. We should fix that.

ioanarusiczki:
@bobsilverberg 
[...]

PR #14031, opened 2020-04-21
Add sha256, mimetype and size to File information returned by reviewers APIs

### LINKED ISSUE #14026: Add sha256, mimetype and size to File information returned by reviewers APIs
This supports #14020 

When a client requests data for a specific file, the API should return the `sha256`, `mimetype` and `size` of the requested file. This information is currently only provided in `entries`, but we want the API to be able to return data for a file without the `entries`, hence the need for this data for the requested file.

bobsilverberg:
Fixes #14026

This will also fix https://github.com/mozilla/addons-code-manager/issues/1548

### DIFF
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -54,11 +54,14 @@ class FileEntriesSerializer(FileSerializer):
     download_url = serializers.SerializerMethodField()
     entries = serializers.SerializerMethodField()
     selected_file = serializers.SerializerMethodField()
+    mimetype = serializers.SerializerMethodField()
+    sha256 = serializers.SerializerMethodField()
+    size = serializers.SerializerMethodField()
 
     class Meta:
         fields = FileSerializer.Meta.fields + (
             'content', 'entries', 'selected_file', 'download_url',
-            'uses_unknown_minified_code'
+            'uses_unknown_minified_code', 'mimetype', 'sha256', 'size'
         )
         model = File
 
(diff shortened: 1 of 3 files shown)


src/olympia/reviewers/serializers.py
```diff
@@ -174,6 +177,17 @@ def _fetch_entries():
 
         return self._entries
 
+    def get_mimetype(self, obj):
+        entries = self.get_entries(obj)
```
diox, line 181: For https://github.com/mozilla/addons-server/issues/14020 to work and help performance we'll want to avoid fetching the entries when fetching a single file, right ? This would mean this and even the `get_selected_file()` implementation would need to change to stop going through all entries every time we request a single file (it could be in cache, since the file tree is likely to be requested in parallel by code manager, but it could also not if we're unlucky)

bobsilverberg: Yes, there will be more changes needed in this area. I started a big patch and then decided to break it down into smaller pieces, and this is just the first small piece. 

I had thought that we might be able to get away with keeping this logic, because code manager will never actually just need a file without also needing the version information the first time, which would then cause this data to be cached, but then I realized that we are pre-fetching the "next" file on each page, and therefore issuing two requests in parallel (one would be for the version + file, and the other would be for just the file), and both of those could theoretically kick off the process to populate the cache.

bobsilverberg: The other helpful thing about this small patch is that it doesn't require any changes to code manager, whereas subsequent patches will need to be landed in tandem with patches in code manager.

bobsilverberg:
@diox I just submitted PR #14043, which builds on this and addresses the issue you raised. If you'd prefer to review that, which includes all of the commits from this, then please feel free to do so, and I can close this PR. Otherwise I'll wait for this to land and then rebase #14043.

src/olympia/reviewers/serializers.py
diox: Yeah I'm fine with that approach as long as you've got improvements lined up for later.

PR #14040, opened 2020-04-22
Override user when creating ActivityLog for version creation

### LINKED ISSUE #14039: Override user when creating `ActivityLog` in submission-related tasks
STR:
- Submit an add-on using the API
- Look for `ActivityLog` related to the user that did the submission

Expected results:
- You find at least one instance of `ADD_VERSION` (`{version} added to {addon}.`) linked to the user and the add-on.

Actual results:
- You don't find any instances of `ADD_VERSION` linked to the user. You do find one linked to the add-on.

This is because the `ActivityLog` is created in a task so it would be linked to the task user instead. We should pass the `user` down to `ActivityLog` creation in `Version.from_upload()`, and check if other `ActivityLog` creations are affected.

diox:
Fixes #14039

### DIFF
--- a/src/olympia/versions/models.py
+++ b/src/olympia/versions/models.py
@@ -36,6 +36,7 @@ from olympia.files.models import File, cleanup_file
 from olympia.translations.fields import (
     LinkifiedField, PurifiedField, TranslatedField, save_signal)
 from olympia.scanners.models import ScannerResult
+from olympia.users.utils import get_task_user
 
 from .compare import version_int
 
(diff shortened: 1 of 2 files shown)


src/olympia/versions/models.py
```diff
@@ -253,7 +254,9 @@ def from_upload(cls, upload, addon, selected_apps, channel,
                     'from_api': upload.source == amo.UPLOAD_SOURCE_API,
                 }
             )
-            activity.log_create(amo.LOG.ADD_VERSION, version, addon)
+            activity.log_create(
+                amo.LOG.ADD_VERSION, version, addon,
+                user=upload.user or get_task_user())
```
eviljeff, line 259: when don't we have `upload.user`?

diox: It should always be there, but some tests specifically set it to `None` to force weird situations and see what happens... so I added this fallback cause `ActivityLog` creation without a `user` silently fails...

PR #14041, opened 2020-04-22
Remove unused file after migrating to Debian Buster Image

championshuttler:
Follow up on #13985 

This file is not needed anymore I think

### DIFF
--- a/Dockerfile.deploy
+++ b/Dockerfile.deploy
@@ -16,7 +16,6 @@ RUN apt-get update && apt-get install -y \
     && rm -rf /var/lib/apt/lists/*
 RUN cat /etc/pki/gpg/GPG-KEY-nodesource | apt-key add -
 ADD docker/debian-buster-nodesource-repo /etc/apt/sources.list.d/nodesource.list
-ADD docker/debian-buster-repo /etc/apt/sources.list.d/buster.list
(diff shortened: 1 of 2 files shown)


diox:
So, it's currently used in the `Dockerfile` and `Dockerfile.deploy`, there is a `ADD docker/debian-buster-repo /etc/apt/sources.list.d/buster.list` line. 

Do check in the container, but I think it's causing a duplicate apt source and should indeed be removed. In that case the `ADD` line in the `Dockerfile` and `Dockerfile.deploy` should go away as well.

championshuttler:
> So, it's currently used in the `Dockerfile` and `Dockerfile.deploy`, there is a `ADD docker/debian-buster-repo /etc/apt/sources.list.d/buster.list` line.
> 
> Do check in the container, but I think it's causing a duplicate apt source and should indeed be removed. In that case the `ADD` line in the `Dockerfile` and `Dockerfile.deploy` should go away as well.

@diox I removed that line in `Dockerfile` last patch but `accidentally ` left in `Dockerfile.deploy` . Removed that now :)

championshuttler:
One more question, why there is a merge commit generated for my last patch https://github.com/mozilla/addons-server/commit/97f0aef666d660ac61c957e3671e4d12a1b209dc

Issue #13999, opened 2020-04-17
"TestADICommand.test_download_counts_from_file_includes_disabled_addons" test failure


eviljeff:
```
____ TestADICommand.test_download_counts_from_file_includes_disabled_addons ____

[gw0] linux -- Python 3.6.7 /home/travis/build/mozilla/addons-server/.tox/main/bin/python3

src/olympia/stats/tests/test_commands.py:219: in test_download_counts_from_file_includes_disabled_addons

    date=self.date, stats_source=self.stats_source)

.tox/main/lib/python3.6/site-packages/django/core/management/__init__.py:148: in call_command

    return command.execute(*args, **defaults)

.tox/main/lib/python3.6/site-packages/django/core/management/base.py:364: in execute

    output = self.handle(*args, **options)

src/olympia/stats/management/commands/download_counts_from_file.py:104: in handle

    if get_date(filepath, sep) != day:

src/olympia/stats/management/commands/__init__.py:21: in get_date

    with open(path) as f:

E   FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/mozilla/addons-server/src/olympia/../../src/olympia/stats/fixtures/files/2014-07-10/download_counts.hive'
```

eviljeff:
Our guess is there's a race condition because the file is copied to that location in test setup and then deleted in cleanup - if two instances are running at once then the first deletes the file before the second is finished.

eviljeff:
can be replicated locally with `pytest src/olympia/stats/tests/test_commands.py::TestADICommand -n 2` (repeat if it doesn't fail)

PR #14043, opened 2020-04-22
 Remove properties from entries, and do not require entries to calculate file-level properties

### LINKED ISSUE #14042: Do not rely on entries for file-only data
This supports #14020 and is a follow-up to #14031.

It proposes two changes:

1. Changes to the logic for calculating `mimetype`, `sha256` and `size`, so that `entries` are not required to be calculated when just retrieving data for a single file. Note that, because code won't have been introduced which would allow for just data for a file to be requested, the serializer will still calculate and return `entries`, but these changes set the stage for requesting file data without `entries`.

2. Removal of properties from `entries` that are no longer needed, which includes `mimetype`, `sha256` and `size` (which are no longer needed because they are now returned for the file), as well as `modified`, which is never used.

Note that a change to code manager will need to land alongside this.

bobsilverberg:
Fixes #14042 

This needs to land in the same push as https://github.com/mozilla/addons-code-manager/pull/1565, which is why I have marked it as "do not merge" for now.

This allows the next step, which will be to have a serializer that can return data for just a file, without having to calculate the `entries` beforehand. It also removes properties from `entries` that are no longer needed.

All data needed for a file can now be calculated and returned without having to generate `entries`. I had to add some extra logic to deal with deleted files, as they won't exist in the `tree`, but would have existed in `entries`. If we feel this is all too much, we could just allow the serializer to always need to calculate `entries` (as suggested by @diox in Matrix), knowing that those `entries` are cached for future requests. That would remove the need for a lot of this code.

### DIFF
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -4,7 +4,6 @@ import mimetypes
 import pathlib
 import json
 from collections import OrderedDict
-from datetime import datetime, timezone, timedelta
 
 import pygit2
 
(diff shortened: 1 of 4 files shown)


bobsilverberg:
I see some tests are failing. I guess some views rely on the deleted properties. I'll have to look into this and fix any problems I've caused there. I'll do that before requesting a review of this.

bobsilverberg:
I fixed the tests, but then realized that this patch doesn't actually avoid generating entries, so it needs some more work.

willdurand:
> I fixed the tests, but then realized that this patch doesn't actually avoid generating entries, so it needs some more work.

Is it ready now?

The build is failing because of https://github.com/mozilla/addons-server/issues/13999

src/olympia/reviewers/serializers.py
```diff
@@ -82,46 +81,26 @@ def get_instance(self):
 
     def _get_commit(self, file_obj):
         """Return the pygit2 repository instance, preselect correct channel."""
+        if hasattr(self, '_commit'):
+            return self._commit
```
willdurand, line 85: If this serializer is called with another `file_obj`, that will return the wrong `commit` object. Is it ok?

willdurand: It would be nice to add a comment to explain that it is used to avoid a call to `revparse`.

src/olympia/reviewers/serializers.py
```diff
@@ -82,46 +81,26 @@ def get_instance(self):
...
-        # Return the hash if we already saved it to the locally cached
-        # `self._entries` dictionary.
-        _entries = getattr(self, '_entries', {})
+        return self._commit
 
-        if _entries and _entries[selected_file]['sha256']:
-            return _entries[selected_file]['sha256']
+    def _get_tree(self, obj):
+        if hasattr(self, '_tree'):
+            return self._tree
```
willdurand, line 98: A comment to explain why this is needed would be nice here too.

src/olympia/reviewers/serializers.py
```diff
@@ -171,46 +140,117 @@ def _fetch_entries():
...
+
+        return cache.get_or_set(cache_key, _calculate_hash, 60 * 60 * 24)
 
     def get_size(self, obj):
-        entries = self.get_entries(obj)
-        return entries[self.get_selected_file(obj)]['size']
+        tree = self._get_tree(obj)
+        selected_file = self.get_selected_file(obj)
+        if selected_file in tree:
+            blob_or_tree = tree[selected_file]
```
willdurand, line 192: I saw this code before I think. Should we have a `get_blob_or_tree_for_selected_file` method?

willdurand: Actually, we are only interested in `get_blob_for_selected_file` it seems.

src/olympia/reviewers/serializers.py
```diff
@@ -171,46 +140,117 @@ def _fetch_entries():
...
+                raise NotFound('File not found')
 
         return requested_file
 
     def get_content(self, obj):
-        commit = self._get_commit(obj)
-        tree = self.repo.get_root_tree(commit)
+        tree = self._get_tree(obj)
         selected_file = self.get_selected_file(obj)
         blob_or_tree = tree[selected_file]
```
willdurand, line 255: We could even reuse `get_blob_for_selected_file` here :p

bobsilverberg:
> > I fixed the tests, but then realized that this patch doesn't actually avoid generating entries, so it needs some more work.
> 
> Is it ready now?
> 

Sorry, yes, I assigned a reviewer and assumed that was a sign it was ready to review, but you're right, I should have mentioned it here too.

src/olympia/reviewers/serializers.py
bobsilverberg: As discussed on Matrix, I don't think your first comment is an issue. I did add a comment regarding your second comment in the next commit.

bobsilverberg:
@willdurand I made a bunch of changes, including a new helper for the `blob` and using `@cached_property` for `commit` and `tree`. It seems like we're still debating the whole caching thing on Matrix, but maybe you want to take a look at the latest version for now.

bobsilverberg:
@willdurand I believe we decided to go ahead and cache things in the serializer for now, as we never need to use it for more than one file, and I'm not really sure when there might be a use case for that. This is ready to be looked at again.

willdurand:
> @willdurand I believe we decided to go ahead and cache things in the serializer for now, as we never need to use it for more than one file, and I'm not really sure when there might be a use case for that. This is ready to be looked at again.

I thought the decision was to use the view to cache things ([example of the use of `view` in a serializer](https://github.com/mozilla/addons-server/blob/963ed3f9a697d331373668ab9f158202258955c5/src/olympia/ratings/serializers.py#L34))

bobsilverberg:
This is now ready for another review @willdurand. Sorry for all the changes. I've gone back to using cached `entries` to look up some of the `file` values.

PR #14045, opened 2020-04-23
Process all emails using Django Template Language, harmonize template paths

### LINKED ISSUE #14044: Process all email templates with the same template language, harmonize their paths
Most or our email templates are being processed with Django Template Language instead of Jinja2, but not all. For consistency, and to help fix the config for extraction in https://github.com/mozilla/addons-server/issues/12879, we want to make them all use the same engine and harmonize their paths - have them all under `emails/` and not `email/`.

~~**Edit:** because of issues with localization we're going to process them all using Jinja, not Django.~~

diox:
Fixes #14044


src/olympia/lib/settings_base.py
```diff
@@ -271,28 +271,26 @@ def get_db_config(environ_var, atomic_requests=True):
...
-# Templates
+# Templates configuration.
+# List of path patterns for which we should be using Django Template Language.
 JINJA_EXCLUDE_TEMPLATE_PATHS = (
+    # All emails should be processed with Django for consistency.
+    r'^.*\/emails\/',
+
+    # ^admin\/ covers most django admin templates, since their path should
+    # follow /admin/<app>/<model>/*
     r'^admin\/',
```
eviljeff, line 282: if this is django's admin templates shouldn't this be in the "Third-party apps + django." section below?

diox: It's our own admin templates as opposed to third-party.

eviljeff: right, it's matching `admin` folder within `templates` for each app...  In that case, follow-up of moving `blocklist` and `addons` admin templates into the correct places so we don't have to special case them?

diox: Yeah I was too lazy to do this after having spent the day on various extraction-related issues, but you're right it'd be cleaner. I filed https://github.com/mozilla/addons-server/issues/14053

Issue #9112, opened 2018-08-08
Automate l10n extract/merge


EnTeQuAk:
Acceptance criteria:

* weekly Pull request is opened with extracted localization context
* manually approved and merged as a part of release push

Notes:

* use the simplest approach
* should include addons frontend and server, not linter

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-165)

EnTeQuAk:
Just for the sake of documenting my progress, here are two temporary experiments that progress on automating l10n extraction:

* An experiment that adds auto-merge capabilities to the bash script https://github.com/mozilla/addons-server/commit/4b4db2c06c12ded680399751744d2988d3f1b331
* Another experiment that converts the current extraction bash script to Python for easier testing and extension https://github.com/mozilla/addons-server/commit/6da2e9249dbf85e13f4246ae91777e7f2c0c642f

diox:
Still needed.

diox:
Note: we might want to consider running just the extract (generating the `.pot` file) on each commit / as part of the test suite. It's reasonably fast and if it runs into trouble, that should probably error out - it's likely the configuration/templates are broken in some way.

KevinMind:
I am even more in favor of implementing the proposal here after looking at this.

But here is a potential solution for the current manual step of extracting and merging locales using a git pre-commit hook.

https://github.com/mozilla/addons-server/assets/19595165/08874a08-969e-493a-815a-ad419de68b19

PR: https://github.com/mozilla/addons-frontend/pull/12775

KevinMind:
Adding this here as a WIP of how this could be solved in a scalable way that doesn't deviate too much from the current architecture.

We can utilize a github action to handle extracting and merging locales in a fully automated pipeline that can be shared between multiple repositories enabling this feature for all Addons repositories that do locale extraction.

Requirements:
- implement a generic github action that:
  - orchestrates execution of extract and merge via expected bash scripts that should be implemented in the consuming repo.
  - executes on pushes to master (a better and more automated solution would be to execute mid merge, but this would require implementing merge queues and we 1) don't have that and 2) have mixed feelings about them.
  - checks if locale files have changed
  - if they have, commits them and pushes to master (we could also open a PR but pontoon is already committing directly so there is little benefit)

[...]

PR #14049, opened 2020-04-23
Make locales extraction script work with and without token

diox:
Helps while we're waiting to have fully automated l10n extraction (https://github.com/mozilla/addons-server/issues/9112)

### DIFF
--- a/scripts/extract-l10n.sh
+++ b/scripts/extract-l10n.sh
@@ -8,10 +8,10 @@
 #   - Update your code
 #   - Extract new strings and push to the .po files
 #
-# If you're running the script manually please make sure to expose
+# For the fully automated experience make sure to expose
 # the following variables to the environment:
 #   - GITHUB_TOKEN (to the github token of addons-robot,
-#                   talk to tofumatt or cgrebs)
+#                   talk to @diox or @muffinresearch)
 #   - TRAVIS_REPO_SLUG="mozilla/addons-server"
 #   - TRAVIS_BRANCH="master"
 
@@ -72,9 +72,11 @@ function extract_locales {
     echo "done."
 }
 
-
-function commit_and_push {
+function git_commit {
     git commit -m "$MESSAGE" --author "$ROBOT_NAME <$ROBOT_EMAIL>" --no-gpg-sign locale/*/LC_MESSAGES/*.po locale/templates/
+}
+
+function git_push {
     git push -q "https://addons-robot:$GITHUB_TOKEN@github.com/$TRAVIS_REPO_SLUG/"
 }
 
@@ -98,21 +100,23 @@ function create_auto_pull_request {
     echo "auto merge pull request is created ..."
 }
 
-if [ "$TRAVIS_BRANCH" != "master" ]; then
-  echo "This commit was made against the $TRAVIS_BRANCH and not the master! No extract!"
-  exit 0
-fi
-
-if [ "GITHUB_TOKEN" == "" ]
-then
-    echo "Must provide github token"
-    exit 0
-fi
-
 init_environment
 
 extract_locales
 
-commit_and_push
+git_commit
 
-create_auto_pull_request
+if [ "GITHUB_TOKEN" = "" ]
+then
+    echo "No Github token provided, you'll have to push the branch and create the pull request yourself."
+else
+    if [ "$TRAVIS_BRANCH" != "master" ]; then
+      echo "This commit was made against the $TRAVIS_BRANCH and not the master! No extract!"
+      exit 0
+    fi
+    git_push
+
+    create_auto_pull_request
+fi
+
+


diox:
Please do test locally. It should work from the container, but do note that it will attempt to checkout to `master` first.

scripts/extract-l10n.sh
```diff
@@ -8,10 +8,10 @@
 #   - Update your code
 #   - Extract new strings and push to the .po files
 #
-# If you're running the script manually please make sure to expose
+# For the fully automated experience make sure to expose
 # the following variables to the environment:
 #   - GITHUB_TOKEN (to the github token of addons-robot,
-#                   talk to tofumatt or cgrebs)
+#                   talk to @diox or @muffinresearch)
 #   - TRAVIS_REPO_SLUG="mozilla/addons-server"
```
eviljeff, line 15: Can't we provide this as a default in the file?

scripts/extract-l10n.sh
```diff
@@ -8,10 +8,10 @@
...
 #   - TRAVIS_BRANCH="master"
```
eviljeff, line 16: This doesn't appear to be used at all (despite the erroring if it's not set).

scripts/extract-l10n.sh
```diff
@@ -8,10 +8,10 @@
...
```
eviljeff, line 17: In `init_environment` it does `git checkout master` which depends on you having an up-to-date local master.  Can we just `git checkout upstream/master` instead?   (Even better if we can always checkout from the repo in `TRAVIS_REPO_SLUG`)

scripts/extract-l10n.sh
```diff
@@ -98,21 +100,23 @@ function create_auto_pull_request {
...
 init_environment
 
 extract_locales
 
-commit_and_push
+git_commit
+
+if [ "GITHUB_TOKEN" = "" ]
+then
+    echo "No Github token provided, you'll have to push the branch and create the pull request yourself."
```
eviljeff, line 111: I understand why we can't create the pull request without the github API access but we could still push the branch like https://github.com/mozilla/addons-frontend/pull/9375/files#diff-cae1c8c05fe95719db14844a8f236ab1R76

eviljeff:
Anything that can be done to have https://github.com/mozilla/addons-frontend/pull/9375 and this work in the same way would be a big improvement - right now the two scripts seem to make a lot of assumptions about git checkouts and branches and it's hard to remember the differences.

diox:
> I tried but it fails on `jinja2.exceptions.TemplateSyntaxError: Encountered unknown tag 'load'.` - which I thought was fixed in another commit?

It was! which file was this ? Are you sure your `master` was up to date ?

eviljeff:
> 
> 
> > I tried but it fails on `jinja2.exceptions.TemplateSyntaxError: Encountered unknown tag 'load'.` - which I thought was fixed in another commit?
> 
> It was! which file was this ? Are you sure your `master` was up to date ?

Forgot to do `fetch upstream` first.  (can we do that too?).  

It completed and then said:
`
Cleaning out obsolete messages.  See bug 623634 for details.
done.

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: unable to auto-detect email address (got 'olympia@0eca485b3639.(none)')
`
- this is being run in the container so does it have all the relevant git details?

diox:
Ah. Ok, so I should probably not even do the `commit` in that mode then I think, to be safe.

scripts/extract-l10n.sh
diox: This is purely if used in travis to make sure it's not running against branches other than master

scripts/extract-l10n.sh
diox: As `TRAVIS_BRANCH` this is automatically present in travis

scripts/extract-l10n.sh
diox: mmm in the case of travis, `upstream` doesn't exist as the remote... I'm going to try to steal some logic from https://github.com/mozilla/addons-frontend/blob/master/bin/run-l10n-extraction for this, pulling from the remote directly.

diox:
Going to rework this to use the same logic we use on the frontend.

scripts/extract-l10n.sh
eviljeff: I'm thinking for local use so we don't have to define extra variables to make it run.

scripts/extract-l10n.sh
eviljeff: okay, that makes sense.

scripts/extract-l10n.sh
diox: Yeah I'll make sure no variables are needed for manual usage

PR #14052, opened 2020-04-23
add bulk_add_blocks command

### LINKED ISSUE #14050: Add command to bulk add guids to the blocklist
#14038 wasn't enough - there are bottlenecks in django and the ORM with attempting to add a BlockSubmission with a large (1000+) number of guids in single attempt. 

Instead lets add a management command to be run by ops for this one-off task.

eviljeff:
fixes #14050


src/olympia/blocklist/models.py
```diff
@@ -150,6 +149,36 @@ def review_unlisted_link(self):
...
         return BlocklistSubmission.get_submissions_from_guid(self.guid)
 
+    @classmethod
+    def get_blocks_from_guids(cls, guids):
+        """Given a list of guids, return a list of Blocks - either existing
+        instances if the guid exists in a Block, or new instances otherwise.
+        """
+        # load all the Addon instances together
+        addons = list(Addon.unfiltered.filter(
+            guid__in=guids).no_transforms())
```
diox, line 159: Probably want to force the use of the replica here and in the `existing_blocks` fetching below, for a bulk operation it becomes worth it I think.

diox: This is probably already happening in the management command case though, since there is no `use_primary_db` decorator or things like that, so carry on :)

Issue #14057, opened 2020-04-24
Error in locales causes travis/circle errors


willdurand:
We should fix the errors reported by `dennis` because it breaks pretty much everything :(

willdurand:
Some examples:

```
locale/kab/LC_MESSAGES/django.po:6732: 'msgid' and 'msgstr' entries do not both begin with '\n'
locale/kab/LC_MESSAGES/django.po:6743: 'msgid' and 'msgstr' entries do not both begin with '\n'
```

If we fix those, new ones will appear for similar strings in other locales :( I don't know why this is happening, though.

PR #14059, opened 2020-04-24
Fix sv_SE locale

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14057, maybe


willdurand:
arg, another locale is failing now. It looks like all the locales are going to be affected so my solution won't scale..

PR #14060, opened 2020-04-24
Handle BrokenRefError errors

willdurand:

---

This patch handles the git errors related to broken git references (see issue above). The strategy is to re-extract everything because we cannot repair the git repository.

### DIFF
--- a/src/olympia/git/tests/test_commands.py
+++ b/src/olympia/git/tests/test_commands.py
@@ -1,6 +1,8 @@
 from pathlib import Path
 from unittest import mock
 
+import pytest
+
 from django.test.utils import override_settings
 
 from olympia.amo.tests import (
@@ -9,7 +11,7 @@ from olympia.amo.tests import (
     create_switch,
     version_factory,
 )
-from olympia.lib.git import AddonGitRepository
+from olympia.lib.git import AddonGitRepository, BrokenRefError
 from olympia.git.models import GitExtractionEntry
 from olympia.git.tasks import (
     extract_versions_to_git,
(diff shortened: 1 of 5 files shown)


src/olympia/git/tests/test_commands.py
```diff
@@ -172,11 +172,14 @@ def test_extract_addon_with_error_during_extraction(self):
 
         try:
             self.command.extract_addon(entry)
-        except Exception:
-            pass
+        except Exception as exc:
+            assert isinstance(exc, BrokenRefError)
```
eviljeff, line 176: you can use `self.assertRaises(BrokenRefError)` instead.

src/olympia/git/tests/test_commands.py
```diff
@@ -170,13 +172,14 @@ def test_extract_addon_with_error_during_extraction(self):
         Path(f'{repo.git_repository_path}/.git/refs/heads/listed').touch()
         entry = GitExtractionEntry.objects.create(addon=addon)
 
-        try:
+        with pytest.raises(BrokenRefError):
```
eviljeff, line 175: `self.assertRaises` is the preferred solution.  We'd only use `pytest.raises` outside of a `TestCase`.

willdurand: Why is that the preferred solution? Is that becasue `assertRaises` comes from `unittest`?

I have no problem with changing `pytest.raises` to `self.assertRaises` but we should do that everywhere, no?

- https://github.com/mozilla/addons-server/blob/855464d3890d858edf9b9a6cf09dec14c7f8666a/src/olympia/accounts/tests/test_verify.py#L33
- https://github.com/mozilla/addons-server/blob/36e03763b6e90c42dff2d8144cf173c478a08a40/src/olympia/files/tests/test_utils.py#L123

eviljeff: I can't find the source code right now but iirc last time I looked `assertRaises` had some extra support for tracebacks or something.  Regardless, it's the norm in the addons-server codebase and imo it's best to pick a flavour and try to stick to it - I'm not sure it's worth specifically going through and replacing all the instances but if we were changing the code it's best to change it (that's why I suggested it 😉 at the time)

PR #14061, opened 2020-04-24
bypass signoff for submissions that don't change min/max version

### LINKED ISSUE #13477: Dual sign off should be required only when editing the version range in a Block
Follow up for #13341 

1. Go to Admin - Blocklist model
2. Open the list of existing blocks - https://addons-internal.dev.mozaws.net/en-US/admin/models/blocklist/block/
3. Select one of the blocks and open its detail page (make sure that the block you selected would normally require dual sign-off, i.e. is either a listed add-on with >100k ADU or an add-on without stats)
4. Edit any field other than the version range, i.e the Reason field
5. Save your changes

### What happened?
The block is Pending for dual sign off

The Block should be updated without requiring dual sign off. 

The requirements mentioned the following:

> The changes that require dual sign-off are:
> 
> * Adding a new block.
> * Deleting a block.
> * _Changing the version range for a block._

eviljeff:
turned out to be rabbit hole of potential implementations but ended up with something that seems sensible.  fixes #13477

### DIFF
--- a/src/olympia/blocklist/models.py
+++ b/src/olympia/blocklist/models.py
@@ -29,6 +29,7 @@ from .utils import (
 class Block(ModelBase):
     MIN = '0'
     MAX = '*'
+
     guid = models.CharField(max_length=255, unique=True, null=False)
     min_version = models.CharField(max_length=255, blank=False, default=MIN)
     max_version = models.CharField(max_length=255, blank=False, default=MAX)
@@ -155,13 +156,13 @@ class BlocklistSubmission(ModelBase):
     SIGNOFF_PENDING = 0
     SIGNOFF_APPROVED = 1
     SIGNOFF_REJECTED = 2
-    SIGNOFF_NOTNEEDED = 3
+    SIGNOFF_AUTOAPPROVED = 3
     SIGNOFF_PUBLISHED = 4
     SIGNOFF_STATES = {
         SIGNOFF_PENDING: 'Pending',
         SIGNOFF_APPROVED: 'Approved',
         SIGNOFF_REJECTED: 'Rejected',
-        SIGNOFF_NOTNEEDED: 'No Sign-off',
+        SIGNOFF_AUTOAPPROVED: 'No Sign-off',
         SIGNOFF_PUBLISHED: 'Published to Blocks',
     }
     SIGNOFF_STATES_FINISHED = (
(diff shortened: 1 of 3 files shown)


src/olympia/blocklist/models.py
```diff
@@ -256,49 +257,65 @@ def can_user_signoff(self, signoff_user):
...
         if self.input_guids and not self.to_block:
-            processed = self.process_input_guids(
-                self.input_guids,
-                self.min_version,
-                self.max_version,
-                load_full_objects=False,
-                filter_existing=(self.action == self.ACTION_ADDCHANGE))
             # serialize blocks so we can save them as JSON
-            self.to_block = self._serialize_blocks(processed.get('blocks', []))
+            self.to_block = self._serialize_blocks()
```
eviljeff, line 318: This change to move more code to `self._serialize_blocks` was a refactor for an aborted implementation but I preferred the cleaner `save` method so kept it.

src/olympia/blocklist/models.py
```diff
@@ -363,16 +380,16 @@ def _split_guids_fake(cls, guids, v_min, v_max, filter_existing):
...
-                max_version=max_version,
-                is_imported_from_kinto_regex=kinto_id.startswith('*'))
-            for id_, guid, min_version, max_version, kinto_id in block_qs]
+                id=block.id,
+                guid=block.guid,
+                addon=addon_guid_dict.get(block.guid),
+                min_version=block.min_version,
+                max_version=block.max_version,
+                is_imported_from_kinto_regex=block.kinto_id.startswith('*'))
+            for block in block_qs]
```
eviljeff, line 392: this is an unconnected refactoring too when I was experimenting with adding new properties to `FakeBlock` but the code is cleaner now so kept it.

src/olympia/blocklist/models.py
```diff
@@ -256,49 +257,65 @@ def can_user_signoff(self, signoff_user):
...
-        if self.signoff_state == self.SIGNOFF_PENDING and not any_unsafe():
-            self.update(signoff_state=self.SIGNOFF_NOTNEEDED)
+        return has_new_blocks or blocks_with_version_changes_qs.exists()
+
+    def update_if_signoff_not_needed(self):
+        is_pending = self.signoff_state == self.SIGNOFF_PENDING
+        add_action = self.action == self.ACTION_ADDCHANGE
+        if (
+            (is_pending and self.all_adu_safe()) or
+            (is_pending and add_action and not self.has_version_changes())
```
bobsilverberg, line 282: When would a block have `has_version_changes() == True` but `self.action != self.ACTION_ADDCHANGE`? I.e., which use cases are we excluding with this rule?

src/olympia/blocklist/models.py
```diff
@@ -363,16 +380,16 @@ def _split_guids_fake(cls, guids, v_min, v_max, filter_existing):
 
         # And then any existing block instances
         block_qs = Block.objects.filter(guid__in=guids).values_list(
-            'id', 'guid', 'min_version', 'max_version', 'kinto_id')
+            'id', 'guid', 'min_version', 'max_version', 'kinto_id', named=True)
```
bobsilverberg, line 383: Do we need `named=True` here, or was that just for debugging purposes?

src/olympia/blocklist/models.py
```diff
@@ -429,7 +446,7 @@ def process_input_guids(cls, guids, v_min, v_max, load_full_objects=True,
         Block.addon.average_daily_users,
         Block.min_version,
         Block.max_version,
-        Block.process_input_guids
+        Block.is_imported_from_kinto_regex
```
bobsilverberg, line 449: Was this just a typo from before?

src/olympia/blocklist/tests/test_models.py
```diff
@@ -55,7 +55,7 @@ def test_is_submission_ready(self):
         assert not block.is_submission_ready
 
         # Except when the state is NOTNEEDED.
```
bobsilverberg, line 57: Should this comment be changed to read `AUTOAPPROVED`?

src/olympia/blocklist/tests/test_models.py
```diff
@@ -107,3 +107,49 @@ def test_get_submissions_from_guid(self):
...
+
+        submission.to_block = submission._serialize_blocks()
+        # no changes to anything
+        assert not submission.has_version_changes()
+
+        block.update(min_version='999', reason='things')
+        # min_version has changed (and reason)
+        assert submission.has_version_changes()
+
+        submission.update(min_version='999')
```
bobsilverberg, line 153: Why `block.update()` on line 149 and `submission.update()` here?

src/olympia/blocklist/models.py
eviljeff: Only with `ACTION_ADDCHANGE` are the versions updated - with `ACTION_DELETE` the version properties aren't used (though do exist on the model instance) so we shouldn't consider if the values are different in any Blocks vs. this submission.

tbh, conceptually having a separate model for `ACTION_DELETE` may be clearer but model inheritance in django is tricky and can cause more problems than it solves.

src/olympia/blocklist/models.py
eviljeff: it's needed - what it does is `.values_list` then returns a `namedtuple` rather than just a `tuple` so we can iterate over `x` and reference the properties as `x.id` rather than having to destructure `x` into it's properties (and having to name them again) or reference the `tuple` elements by index (`x[0]`).

src/olympia/blocklist/models.py
eviljeff: yes!  I've no idea when it happened because `process_input_guids` has always been a `classmethod` rather than a property - find/replace fail or copypasta mistake.

src/olympia/blocklist/tests/test_models.py
eviljeff: yep

src/olympia/blocklist/tests/test_models.py
eviljeff: `block.update` so the submission would be changing the `min_version` back to the default (`0`) and `.has_version_changes()` would be true; `submission.update` then changes it so submission isn't updating the block `min_version` because it's already `'999'` so `.has_version_changes()` would be false; (but it's not the default, so testing it works with non-default properties not changing)

src/olympia/blocklist/tests/test_models.py
eviljeff: https://github.com/mozilla/addons-server/pull/14061/commits/f3445f19b778f587d9e23d3363314af5f0924d43

Issue #9588, opened 2018-10-04
Error messages in the static theme submission wizard should be displayed next to the field that generated them


AlexandraMoga:
STR:

1. Log in to Developer Hub
2. Submit a new static theme using the wizard
3. Enter some spam content in the "Name" field in order to trigger an error
4. Click on Finish Theme after completing all the submission steps
5. Check the error message received 

Actual result:
The error message is displayed at the bottom of the page

Expected result:
The error message should be displayed right after the Name field to better emphasize what generated the error 

Notes:
- the same idea is relevant for other error messages that might be triggered in the submission wizard (see 2nd screenshot as example)
- reproduced on all AMO servers with FF62, WIn10x64

[1]
![image](https://user-images.githubusercontent.com/31961530/46466484-1985e680-c7d4-11e8-8143-bdba5e5baa3a.png)

[2]
![image](https://user-images.githubusercontent.com/31961530/46466526-33272e00-c7d4-11e8-838d-1b69af0b444b.png)

eviljeff:
this depends on #9586

eviljeff:
for [1] #9586 will include the name of the field (i.e. "name") in the message and mozilla/addons#804 does something similar for [2] so I'm going to downgrade this.

caitmuenster:
This is a good issue for folks who have contributed to the add-ons server before. @eviljeff can provide guidance.

caitmuenster:
Go for it, @dsychin!

dsychin:
Hi, can I give this issue a try?

dsychin:
I am running into a problem trying to replicate the error.
Every time I fill in the fields and press Finish Theme, it just gets stuck at "Uploading Theme".

This is what I get in my terminal.
```
addons-frontend_1  | [1] [1557569375801850000] WARN  (amo/146 on 2e484ae7336d): CSP has been disabled from the config
addons-frontend_1  | [1]     amo_request_id: "01754654-3120-4937-b095-9af43fff4f9a"
addons-frontend_1  | [1] [1557569375802117600] INFO  (amo/146 on 2e484ae7336d): Replacing lang in URL en-US -> en-US
addons-frontend_1  | [1]     amo_request_id: "01754654-3120-4937-b095-9af43fff4f9a"
addons-frontend_1  | [1] [1557569375802168300] INFO  (amo/146 on 2e484ae7336d): Exception in URL found; we fallback to addons-server.
[...]

PR #14065, opened 2020-04-24
Error messages displayed next to the field error

avgupt:
Fixes: https://github.com/mozilla/addons-server/issues/9588

      PRs open for the same issue.
* [x] Add before and after screenshots (Only for changes that impact the UI).

Screenshots:
Screenshot from 2020-04-25 00-40-40
Screenshot from 2020-04-25 00-41-47

### DIFF
--- a/src/olympia/devhub/templates/devhub/addons/submit/wizard.html
+++ b/src/olympia/devhub/templates/devhub/addons/submit/wizard.html
@@ -100,8 +100,6 @@
       {{ new_addon_form.compatible_apps.errors }}
       {{ new_addon_form.upload.errors }}
       {{ new_addon_form.non_field_errors() }}
-      <ul class="errorlist validator">
-      </ul>
       <div class="submission-buttons addon-submission-field">
         <button class="button upload"
             {% if addon %}
--- a/src/olympia/devhub/templates/devhub/addons/submit/wizard.html
+++ b/src/olympia/devhub/templates/devhub/addons/submit/wizard.html
@@ -29,6 +29,8 @@
       </ul>
         {% else %}
           <input type="text" id="theme-name"/>
+          <ul class="theme-name-error errorlist validator">
+          </ul>
         {% endif %}
     </div>
     <div id="theme-header" class="row" data-existing-header="{{ existing_properties.get('images', {}).get('theme_frame','') }}">
@@ -45,6 +47,8 @@
       </ul>
       <ul class="errorlist"></ul>
       <img class="preview" alt="">
+      <ul class="header-image-error errorlist validator">
+      </ul>
       <a href="#" class="reset">
         {{ _('Select a different header image') }}</a>
     </div>
--- a/src/olympia/devhub/templates/devhub/addons/submit/wizard.html
+++ b/src/olympia/devhub/templates/devhub/addons/submit/wizard.html
@@ -29,7 +29,7 @@
       </ul>
         {% else %}
           <input type="text" id="theme-name"/>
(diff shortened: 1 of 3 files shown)


avgupt:
@eviljeff @caitmuenster can you please take a look?

static/js/zamboni/static_theme.js
```diff
@@ -224,8 +226,12 @@ $(document).ready(function() {
                     $wizard.find('#submit-describe').submit();
                 } else {
                     data.validation.messages.forEach(function(message) {
-                       $('.errorlist.validator').append($('<li>', {'html': message.message}));
-                       console.error(message);
+                       if (headerImageError) {
+                        $('.header-image-error').append($('<li>', {'html': message.message}));
+                       } else {
+                        $('.theme-name-error').append($('<li>', {'html': message.message}));
```
eviljeff, line 232: There can be non-name errors thrown too if something unexpected happens - anything the server or linter throws back as an error could be returned.  Though I agree it's most likely going to be something with the name so including it there (so at the top of the form) is better.

src/olympia/devhub/templates/devhub/addons/submit/wizard.html
```diff
@@ -29,6 +29,8 @@ <h3>{{ _('Theme name') }}<span class="req" title="{{ _('required') }}">*</span><
       </ul>
         {% else %}
           <input type="text" id="theme-name"/>
+          <ul class="theme-name-error errorlist validator">
```
eviljeff, line 32: on the screenshot the message(s) aren't on new lines so I think you need to tweak the css

static/js/zamboni/static_theme.js
eviljeff: change `theme-name-error` to `general-validation-error` everywhere so future codebase maintainers don't think it's name specific.

static/js/zamboni/static_theme.js
```diff
@@ -224,8 +226,12 @@ $(document).ready(function() {
...
+                       }
+                       console.error(data);
```
eviljeff, line 234: this is changed from `console.error(message)` - do we need the extra information from the data object?

src/olympia/devhub/templates/devhub/addons/submit/wizard.html
avgupt: Do we need them on a new line?
I deliberately made them appear next to the field.

static/js/zamboni/static_theme.js
avgupt: No we don't. I was using it for debugging purposes. Seems like I forgot to change it back.

src/olympia/devhub/templates/devhub/addons/submit/wizard.html
eviljeff: Please, do. It's what we do elsewhere on the site - we don't know how long the error messages will be.

static/js/zamboni/static_theme.js
```diff
@@ -171,6 +172,7 @@ $(document).ready(function() {
 
             zip.generateAsync({type: 'blob'}).then(function (blob) {
                 if (blob.size > MAX_STATICTHEME_SIZE) {
+                    headerImageError = true;
```
eviljeff, line 175: thinking about this, we'll need to reset `headerImageError` between each submission of the upload (earlier in the on click handler function is probably fine).  Otherwise if you get the size error, fix it, then submit and get a different error from the linter/server it will appear as a header error.

Issue #14063, opened 2020-04-24
Add an admin to view the git extraction queue


willdurand:
We have a new `GitExtractionEntry` model that is used as a git extraction queue (we add add-ons that might have versions to git-extract there). We should have a django admin to see the state of this "queue" in case something goes wrong and an add-on is locked for "git extraction" forever (we need an action to unlock the add-on in this case).

ioanarusiczki:
@willdurand the models/git/gitextractionentry/ is available with an admin account and I also created a new group for Admin:GitExtractionEdit. Users added in this group will only have access to the git queue from admin. 

What I've noticed is that after an upload and until the files for the add-ons are extracted there's a state "In progress" with a question mark and this turns green when the extraction is successful then the entry is no longer available in the queue (if I refresh the page). I think it's the expected result.

I could also delete an entry with state "In Progress", before the files were extracted which I believe it's not ok in case the extraction fails.

willdurand:
> What I've noticed is that after an upload and until the files for the add-ons are extracted there's a state "In progress" with a question mark and this turns green when the extraction is successful then the entry is no longer available in the queue (if I refresh the page). I think it's the expected result.

Yep, sounds good to me.

> I could also delete an entry with state "In Progress", before the files were extracted which I believe it's not ok in case the extraction fails.

So the main use case for the delete button is to delete old entries because entries should be automatically deleted. If we have entries that are older than... 24h for example, then we should manually delete them because it means something went wrong but also no other version can be extracted anymore for the related add-on.

ioanarusiczki:
@willdurand The idea of the new extraction method is to manually extract the versions with problems and this means it's those who remain visible in this admin queue. This means there's to be implemented a way to re-extract the problematic addons by the admin. 
I hope I am right and if I am for now this issue and https://github.com/mozilla/addons-server/issues/13964 can be marked verified because there will be issues and more testing on dev for this feature the following weeks.

willdurand:
> This means there's to be implemented a way to re-extract the problematic addons by the admin.

Yes, we are still thinking about how to do it but there will be something for that in the near future.

PR #14073, opened 2020-04-27
Add admin for GitExtractionEntry model

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14063

---

This patch adds a django admin for the `GitExtractionEntry` model. It allows to view and delete the existing entries. The main use case is to allow an add-on to be extracted again if, for some reason, a previous git-extraction process failed (and did not remove the git extraction entry correctly).

Deleting the entry will not do anything on the add-on. I believe we should have a new action at the add-on level to delete and re-extract it. WDYT?


src/olympia/git/admin.py
```diff
@@ -0,0 +1,49 @@
...
+
+from .models import GitExtractionEntry
+
+
+@admin.register(GitExtractionEntry)
+class GitExtractionEntryAdmin(admin.ModelAdmin):
+    actions = None
+    view_on_site = False
+
+    list_display = ('id', 'formatted_addon', 'in_progress', 'created')
```
diox, line 16: Maybe add `modified` ? I believe it should be updated when we set `in_progress=True`, so that could be useful info maybe ?

willdurand:
> Do we need something to delete the repos ? I am not sure at this point. It's a bit of a nuclear option, I'd prefer if we dig into the logs first to see if there are other cases we need to worry about and have a workaround for in our code first...

@ioanarusiczki fyi, this is why we do not want to add a "re-extract this add-on" button/action yet

Issue #14077, opened 2020-04-27
Add filters to the scanners API


willdurand:
We need some filters for https://addons-server.readthedocs.io/en/latest/topics/api/scanners.html because it would be useful to target a specific scanner or a label.

AlexandraMoga:
I've verified the scanners api on -dev (I've enabled the `enable-scanner-results-api`) and tested around the new filter parameters added for `label` and `scanner`.

For `label`, the only accepted values are `good` or `bad`. Any other values, including `null` or `None` will return a 400 error with the message `"invalid value for label"`
For `scanner`, the only accepted values are the four scanner names we enabled so far. Any other value will return a 400 error with the message `"invalid value for scanner"`

Unfiltered requests will return only `good` or `bad` labelled result, i.e any other scanner results are ignored.

PR #14078, opened 2020-04-28
Add filters to the scanners api endpoint

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14077

---

This patch adds some filters to the scanners API endpoint.

### DIFF
--- a/src/olympia/scanners/views.py
+++ b/src/olympia/scanners/views.py
@@ -4,17 +4,23 @@ from django.conf import settings
 from django.db.models import CharField, Value
 from django.db.transaction import non_atomic_requests
 from django.http import Http404
+from rest_framework.exceptions import ParseError
 from rest_framework.generics import ListAPIView
 
 from olympia import amo
-from olympia.constants.scanners import TRUE_POSITIVE, LABEL_BAD, LABEL_GOOD
+from olympia.constants.scanners import (
+    LABEL_BAD,
+    LABEL_GOOD,
+    SCANNERS,
+    TRUE_POSITIVE,
+)
 from olympia.api.permissions import GroupPermission
 
 from .models import ScannerResult
 from .serializers import ScannerResultSerializer
 
 
-class ScannerResultViewSet(ListAPIView):
+class ScannerResultView(ListAPIView):
     permission_classes = [
         GroupPermission(amo.permissions.ADMIN_SCANNERS_RESULTS_VIEW)
     ]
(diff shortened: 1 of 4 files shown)


willdurand:
Test failures are related to https://github.com/mozilla/addons-server/issues/14057

src/olympia/scanners/views.py
```diff
@@ -4,27 +4,50 @@
...
+class ScannerResultView(ListAPIView):
     permission_classes = [
         GroupPermission(amo.permissions.ADMIN_SCANNERS_RESULTS_VIEW)
     ]
 
     serializer_class = ScannerResultSerializer
 
     def get_queryset(self):
+        label = self.request.query_params.get('label', None)
+        scanner = next(
```
bobsilverberg, line 32: Nit: This feels like a kind of convoluted and overly complex way of doing this, but I suppose it composes it into a single statement. I would have imagined checking for `self.request.query_params.get('scanner')` first, as we don't need to check `SCANNERS` if the query param isn't there, and then, if `scanner` exists, validating that it also exists in `SCANNERS`. What you have here is more concise, but also seems harder to read/mentally parse.

src/olympia/scanners/views.py
```diff
@@ -37,21 +60,39 @@ def get_queryset(self):
...
         )
-        return good_results.union(bad_results).order_by('-created')
+
+        queryset = ScannerResult.objects.none()
+
+        if not label:
+            queryset = good_results.union(bad_results)
+        elif label == LABEL_GOOD:
+            queryset = good_results
+        elif label == LABEL_BAD:
```
bobsilverberg, line 74: So if there is a `label` and it's neither `LABEL_GOOD` nor `LABEL_BAD` we return an empty queryset? Is that the behaviour we want?

bobsilverberg: Oh, wait, I see we throw an error on `get` if the `label` is invalid, but I guess you want this behaviour as a fallback in case a `queryset` is somehow requested outside of a `get`?

willdurand: If the label is neither GOOD nor BAD, we shouldn't return anything indeed. It is less useful now that we have some checks in the `get` method but at least it will always return a queryset.

src/olympia/scanners/views.py
willdurand: Part of the complexity is retrieving a key in a dict based on the value.

Issue #14082, opened 2020-04-28
Do not call the MAD API when we don't need it


willdurand:
We identified cases where it does not really make sense to call the MAD API, including:

- static themes
- search engines
- dictionaries
- language packs

We should not call the MAD API in these cases (but other scanners should run).

willdurand:
Note: we can fix this issue by introspecting the scan map.

AlexandraMoga:
I've tested this issue on -dev, where all the scanners are currently enabled and I have the following results:
1.  submission of dictionary, language pack, static theme _with no matching rules_  => `mad` is not running, `customs` shows `n/a` for score and all other scanners are running
2. submission of dictionary, language pack, static theme _with matching rules_  => all the scanners are running, and a score is returned both for `mad` and for `customs`
3. submitting a very simple extension, like this [example](https://code.addons-dev.allizom.org/en-US/browse/619727/versions/1691196/), will not trigger the `mad` scanner and customs shows `n/a` for score

@willdurand please have a look at these results as I'm not very sure that this is what's expected (especially for 2 and 3). 
Regarding search tools, were you referring to old `xml` formats or to search extensions?

willdurand:
>     1. submission of dictionary, language pack, static theme _with no matching rules_  => `mad` is not running, `customs` shows `n/a` for score and all other scanners are running

Expected, good.

>     2. submission of dictionary, language pack, static theme _with matching rules_  => all the scanners are running, and a score is returned both for `mad` and for `customs`

What does that mean exactly? Can you share an example in private?

>     3. submitting a very simple extension, like this [example](https://code.addons-dev.allizom.org/en-US/browse/619727/versions/1691196/), will not trigger the `mad` scanner and customs shows `n/a` for score

Expected, good.

> @willdurand please have a look at these results as I'm not very sure that this is what's expected (especially for 2 and 3).
> Regarding search tools, were you referring to old `xml` formats or to search extensions?

search extensions

AlexandraMoga:
Verified that is working as expected with search extensions as well.
This can be marked as verified fixed.

PR #14089, opened 2020-04-28
Do not call the MAD API when the scan map is too small

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14082


willdurand:
I'd like to cherry-pick this patch if possible.

PR #14090, opened 2020-04-28
Allow the serializer to return file info without entries

### LINKED ISSUE #14072: Allow The FileEntries serializer to return data for a file only
Depends on ~~#14042~~
Supports #14020 

Allow the `FileEntries` serializer to return data without entries. Once this is in place it should be pretty easy to implement #14020 which would provide an API endpoint that can be used to request just file data, without entries.

bobsilverberg:
Fixes #14072 

This is just a small change to allow a serializer to return a file without entries. Once this lands work can be done to create an API endpoint that returns this limited information for a file.

### DIFF
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -191,6 +191,9 @@ class FileEntriesSerializer(serializers.ModelSerializer):
         return self._entries
 
     def get_entries(self, obj):
+        file_only = self.context.get('file_only', False)
+        if file_only:
+            return None
         entries = self._get_entries(obj)
         return self._trim_entries(entries)
 
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -191,8 +191,7 @@ class FileEntriesSerializer(serializers.ModelSerializer):
         return self._entries
 
     def get_entries(self, obj):
-        file_only = self.context.get('file_only', False)
-        if file_only:
+        if self.context.get('exclude_entries', False):
             return None
         entries = self._get_entries(obj)
         return self._trim_entries(entries)
--- a/src/olympia/reviewers/tests/test_serializers.py
+++ b/src/olympia/reviewers/tests/test_serializers.py
@@ -232,6 +232,26 @@ class TestFileEntriesSerializer(TestCase):
         data = self.serialize(fobj, file='background-script.js')
         assert not data['uses_unknown_minified_code']
 
+    def test_can_exclude_entries(self):
+        file = self.addon.current_version.current_file
+
+        data = self.serialize(file, file_only=True)
+
+        assert data['id'] == file.pk
+        assert data['selected_file'] == 'manifest.json'
+        assert data['mimetype'] == 'application/json'
+        assert data['entries'] is None
+
+    def test_can_exclude_entries_and_specify_a_file(self):
+        file = self.addon.current_version.current_file
+
+        data = self.serialize(file, file='icons/LICENSE', file_only=True)
+
+        assert data['id'] == file.pk
+        assert data['selected_file'] == 'icons/LICENSE'
+        assert data['mimetype'] == 'text/plain'
+        assert data['entries'] is None
+
 
 class TestFileEntriesDiffSerializer(TestCase):
     def setUp(self):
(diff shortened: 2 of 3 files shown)


bobsilverberg:
One test is failing on Travis, but is passing locally, which seems very strange. I'm going to see if I can rework the test, but I'm really not sure why it would pass locally and fail on Travis.

diox:
the mimetype one ? travis might not be running the exact same version of `libmagic` and associated packages than what we have in the docker image (keep in mind the docker image was recently updated, though).

bobsilverberg:
> the mimetype one ? travis might not be running the exact same version of `libmagic` and associated packages than what we have in the docker image (keep in mind the docker image was recently updated, though).

I updated my local install before running these tests, so I should have the most up to date docker image. Is it possible that Travis is still running a different version of packages than me? I can fix this in the test (by choosing a different file with a known mimetype), but is this something we should try to fix globally, or is this just something (different versions of packages) that will happen from time to time?

willdurand:
> I updated my local install before running these tests, so I should have the most up to date docker image. Is it possible that Travis is still running a different version of packages than me? I can fix this in the test (by choosing a different file with a known mimetype), but is this something we should try to fix globally, or is this just something (different versions of packages) that will happen from time to time?

Could you run the following commands in a `make shell`?

```
apt-cache show libmagic1 | grep Version
apt-cache show libmagic-dev | grep Version
```

willdurand:
Also, this comment is probably relevant:https://github.com/mozilla/addons-server/blob/bfd3b046af54c46267c06b863129180a25dd9ea8/src/olympia/reviewers/tests/test_serializers.py#L376-L377

On Travis-CI I see version 5.37 installed and I have 5.35 locally, so I think we should keep the test case as it is and see why it is not working on Travis.

willdurand:
Ok, I understand what's going on. `mimetypes.guess_type` uses multiple sources to retrieve the mimetype of a file. One of these sources is on the file system, e.g., `/etc/mime.types`.

I think we could update the travis-ci config to update this file. Let's add this step at the end of the `before_install` list.

```
sudo echo "ext/markdown                                   md markdown" >> /etc/mime.types
```

bobsilverberg:
@willdurand I tried your suggestion, but there seems to be a permissions issue:
```
/home/travis/.travis/functions: line 109: /etc/mime.types: Permission denied

The command "sudo echo "text/markdown                                   md markdown" >> /etc/mime.types" failed and exited with 1 during .
```

https://travis-ci.org/github/mozilla/addons-server/jobs/681194811

willdurand:
> @willdurand I tried your suggestion, but there seems to be a permissions issue:
> 
> ```
> /home/travis/.travis/functions: line 109: /etc/mime.types: Permission denied
> 
> The command "sudo echo "" >> /etc/mime.types" failed and exited with 1 during .
> ```
> 
> [travis-ci.org/github/mozilla/addons-server/jobs/681194811](https://travis-ci.org/github/mozilla/addons-server/jobs/681194811)

Yes.. I am daft, sorry.

It should be:

```
- echo "text/markdown                                   md markdown" | sudo tee -a /etc/mime.types
```

bobsilverberg:
@willdurand Your comment at https://github.com/mozilla/addons-server/pull/14090#issuecomment-621211916 is valid too. I should have noticed that we already had an issue with checking the mimetype of a markdown file. I could remove that assertion, or, another thing we could do would be to add `.md` files to the list of files that we identify based on extension at https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/git.py#L96, which seems like a good thing to do.

WDYT?

willdurand:
> WDYT?

The new command line I pasted above should work. The permission issue was me not thinking before leaving a comment (`sudo` was on the `echo` command but not on the file because of the `>>` redirect).

Since our docker images have this config, we should try to bring travis closer to them, hence the command to update the file. On a second though, I think we should put the `mime.types` under version control (in the `docker` folder) so that we can use the exact same file in both docker and travis.

I feel like using `SIMPLIFIED_MIMETYPE_DETECTION` isn't necessarily a good idea (otherwise, why would be even use `libmagic` in the first place?) but maybe it's fine. Maybe all this complex logic around mimetypes should be built with `SIMPLIFIED_MIMETYPE_DETECTION`?

willdurand:
> I feel like using SIMPLIFIED_MIMETYPE_DETECTION isn't necessarily a good idea (otherwise, why would be even use libmagic in the first place?) but maybe it's fine. Maybe all this complex logic around mimetypes should be built with SIMPLIFIED_MIMETYPE_DETECTION?

To expand on that, this list should probably only contain file extensions reviewers will look at all the time. It is possible to have a js file with a different file extension, like `.md`, hence the content sniffing. I think a markdown file with only js content would be detected as a js file instead of md file. If we add the `.md` extension to thw simplified list, that won't be the case anymore.

bobsilverberg:
> > I feel like using SIMPLIFIED_MIMETYPE_DETECTION isn't necessarily a good idea (otherwise, why would be even use libmagic in the first place?) but maybe it's fine. Maybe all this complex logic around mimetypes should be built with SIMPLIFIED_MIMETYPE_DETECTION?
> 
> To expand on that, this list should probably only contain file extensions reviewers will look at all the time. It is possible to have a js file with a different file extension, like `.md`, hence the content sniffing. I think a markdown file with only js content would be detected as a js file instead of md file. If we add the `.md` extension to the simplified list, that won't be the case anymore.

That's true. We added it for performance reasons, because the process to sniff the type is expensive. I don't see an issue adding it for markdown files - I just think that we didn't think to add them to the list to begin with.

Regarding the issue with detecting markdown files on Travis, you seem to have 2 suggestions:
1. Use the command you suggested to update `/etc/mime.types`.
2. Put `mime.types` under version control for the docker image.

Do you think we should proceed with one of those, rather than updating `SIMPLIFIED_MIMETYPE_DETECTION`?

src/olympia/reviewers/serializers.py
```diff
@@ -192,6 +192,9 @@ def _fetch_entries():
         return self._entries
 
     def get_entries(self, obj):
+        file_only = self.context.get('file_only', False)
+        if file_only:
```
willdurand, line 196: If you don't use the variable anywhere else, you could directly use `self.context.get` in the `if` statement.

willdurand: Given that we are only excluding entries, I would rather name the variable `exclude_entries`. That seems a bit more informative because what does "[including a] file only" mean for a "file serializer"?

src/olympia/reviewers/tests/test_serializers.py
```diff
@@ -243,6 +243,26 @@ def test_uses_unknown_minified_code(self):
         data = self.serialize(fobj, file='background-script.js')
         assert not data['uses_unknown_minified_code']
 
+    def test_can_exclude_entries(self):
```
willdurand, line 246: The names of the different test cases make me think that it would be a good idea to rename the `file_only` variable to `exclude_entries`.

Issue #14098, opened 2020-04-29
"IndexError: list index out of range" while trying to commit a version without files


willdurand:
We do not check that `all_files` is not empty [here](https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/git.py#L348) because that shouldn't be possible. That being said, old versions might not have any files so we should probably handle this edge case.

ioanarusiczki:
@willdurand If qa can test the issue, please add some STR.
Thanks!

PR #14100, opened 2020-04-29
Create an empty commit when a version has no files

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14098

---

This patch creates an empty commit when a version has no files. It shouldn't be possible anymore but we probably have versions without any file in the wild..

### DIFF
--- a/src/olympia/lib/git.py
+++ b/src/olympia/lib/git.py
@@ -345,12 +345,12 @@ class AddonGitRepository(object):
             translation.activate('en-US')
 
             repo = cls(version.addon.id, package_type='addon')
-            file_obj = version.all_files[0]
+            file_obj = version.all_files[0] if version.all_files else None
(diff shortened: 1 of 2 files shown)


src/olympia/lib/git.py
```diff
@@ -345,12 +345,13 @@ def extract_and_commit_from_version(cls, version, author=None, note=None):
             translation.activate('en-US')
 
             repo = cls(version.addon.id, package_type='addon')
-            file_obj = version.all_files[0]
+            file_obj = (version.all_files[0] if len(version.all_files) > 0 else
```
diox, line 348: `if version.all_files` is simpler and works just as well (I know, I know, I've been guilty of using `len(version.all_files)` myself in code still present in the repos :)

willdurand: oh ok, I always forget if that works in python (and I often have to think twice in JavaScript..)

PR #14103, opened 2020-04-29
Make sure folder fixtures are unique per test case

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13999

### DIFF
--- a/src/olympia/stats/tests/test_commands.py
+++ b/src/olympia/stats/tests/test_commands.py
@@ -14,6 +14,7 @@ from django.test.utils import override_settings
 from unittest import mock
 
 from olympia import amo
+from olympia.amo.storage_utils import rm_stored_dir
 from olympia.amo.tests import addon_factory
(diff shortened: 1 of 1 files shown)


src/olympia/stats/tests/test_commands.py
```diff
@@ -26,23 +26,29 @@
...
     def clean_up_files(self):
-        dirpath = os.path.join(hive_folder, self.date)
+        tmp_hive_folder = self.get_tmp_hive_folder()
+        dirpath = os.path.join(tmp_hive_folder, self.date)
         if os.path.isdir(dirpath):
             for name in os.listdir(dirpath):
                 os.unlink(os.path.join(dirpath, name))
             os.rmdir(dirpath)
+        if os.path.isdir(tmp_hive_folder):
+            os.rmdir(tmp_hive_folder)
```
eviljeff, line 45: you can use `rm_stored_dir` from `storage_utils.py` to clear up multi-level folders/files.

PR #14105, opened 2020-04-29
Remove unnecessary file properties from reviewers API responses

### LINKED ISSUE #14099: Remove unnecessary file properties from reviewers API responses
This supports #13984 

There are a number of properties for the version file which are returned via reviewers API responses that are never used by code manager. These are mostly properties of the file uploaded for the version (i.e., the xpi file), not properties of the file being reviewed (e.g., manifest.json). As code manager never uses these properties, we can eliminate them from the API response.

The properties in question are:
-  created
-  hash
-  is_mozilla_signed_extension
-  is_restart_required
-  is_webextension
-  permissions
-  platform
-  status
-  url

Note that these are documented at https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#version-detail-object.

@wagnerand None of these are currently used by code manager, but could you please confirm that there are no future use cases that might desire these properties. Bear in mind, as discussed above, these are not properties for the specific file being reviewed (e.g., an individual js file), but rather properties of the file that was uploaded for the version (the xpi file).

bobsilverberg:
Fixes #14099

### DIFF
--- a/docs/topics/api/reviewers.rst
+++ b/docs/topics/api/reviewers.rst
@@ -165,7 +165,7 @@ This endpoint allows you to browse through the contents of an Add-on version.
     :>json string validation_url: The absolute url to the addons-linter validation report, rendered as HTML.
     :>json boolean has_been_validated: ``True`` if the version has been validated through addons-linter.
     :>json object addon: A simplified :ref:`add-on <addon-detail-object>` object that contains only a few properties: ``id``, ``name``, ``icon_url`` and ``slug``.
-    :>json object file: The file attached to this version. See :ref:`version detail -> files[] <version-detail-object>` for more details.
+    :>json object file: The file attached to this version. Only includes ``id`` from :ref:`version detail -> files[] <version-detail-object>`.
     :>json string file.content: Raw content of the requested file.
     :>json string file.selected_file: The selected file, either from the ``file`` parameter or the default (manifest.json, install.rdf or package.json for Add-ons as well as the XML file for search engines).
     :>json string|null file.download_url: The download url of the selected file or ``null`` in case of a directory.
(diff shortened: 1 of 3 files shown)


docs/topics/api/reviewers.rst
```diff
@@ -165,7 +165,7 @@ This endpoint allows you to browse through the contents of an Add-on version.
     :>json string validation_url: The absolute url to the addons-linter validation report, rendered as HTML.
     :>json boolean has_been_validated: ``True`` if the version has been validated through addons-linter.
     :>json object addon: A simplified :ref:`add-on <addon-detail-object>` object that contains only a few properties: ``id``, ``name``, ``icon_url`` and ``slug``.
-    :>json object file: The file attached to this version. See :ref:`version detail -> files[] <version-detail-object>` for more details.
+    :>json object file: The file attached to this version. Only includes ``id`` from :ref:`version detail -> files[] <version-detail-object>`.
```
eviljeff, line 168: I don't understand this - it doesn't only include `id` - it includes a bunch of other fields too.

src/olympia/reviewers/serializers.py
```diff
@@ -62,11 +62,10 @@ class FileEntriesSerializer(FileSerializer):
     filename = serializers.SerializerMethodField()
 
     class Meta:
-        fields = FileSerializer.Meta.fields + (
```
eviljeff, line 65: so, maybe `FileEntriesSerializer` shouldn't inherit from `FileSerializer` any longer if it's not using any of it's fields or methods?

docs/topics/api/reviewers.rst
bobsilverberg: Are you sure about that? Which fields? The only one I can see from https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#version-detail-object other than `id` is `size`, but that's not the same field. It's not the size of the xpi file, but the size of the file requested. Maybe I'm misunderstanding what you're saying, but my intention was to not return any of the `file` fields from `version.file` other than `id`.

src/olympia/reviewers/serializers.py
bobsilverberg: Hmm, good point. Let me look into that.

docs/topics/api/reviewers.rst
eviljeff: Oh, you mean _of the fields from `version-details-object`_ it only includes `id`.  That's accurate, but, imo, confusing - better to just include the description for `id` in the block below.

src/olympia/reviewers/serializers.py
bobsilverberg: Fixed!

PR #14108, opened 2020-04-29
upgrade filtercascade to 0.4.1; fix edge when blocked>unblocked

### LINKED ISSUE #14066: Update to filtercascade >= 0.4.1 and fix the logic
`filtercascade` has had an update to fix a typo (`filter.bitarray` -> `current_filter.bitarray`), plus some extra sanity checks (https://github.com/mozilla/filter-cascade/pull/23).

The implementation in addons-server should be updated to use this new version.
Besides that, the `fprs` calculation usage should also be adjusted: Using that logic I hit a runtime error for the sample input of 1 non-blocked and 1000 blocked add-ons. The logic should be resilient enough to not fail unexpectedly.

eviljeff:
fixes #14066

### DIFF
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -76,7 +76,7 @@ class MLBF():
 
     @classmethod
     def generate_mlbf(cls, stats, blocked, not_blocked):
-        log.info('Generating filter')
+        log.info('Starting to generating bloomfilter')
 
         cascade = FilterCascade(
             defaultHashAlg=HashAlgorithm.SHA256,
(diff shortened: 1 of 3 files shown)


src/olympia/blocklist/mlbf.py
```diff
@@ -77,32 +76,28 @@ def hash_filter_inputs(cls, input_list):
 
     @classmethod
     def generate_mlbf(cls, stats, blocked, not_blocked):
-        """Originally based on:
-        https://github.com/mozilla/crlite/blob/master/create_filter_cascade/certs_to_crlite.py
-        (not so much any longer, apart from the fprs calculation)
-        """
-        salt = secrets.token_bytes(16)
+        log.info('Generating filter')
```
diox, line 79: Could go with a more specific logging message.

PR #14114, opened 2020-04-29
Configure Dependabot

### LINKED ISSUE #14088: fix/replace pyup
we're 6 days without any new pyup.io pull requests being created and my https://pyup.io/account/account/ doesn't show any mozilla organisation reps any longer.

It's unclear if it's a change on our side to restrict access (the application permissions page show ❌ next to organisation access) or on their side (the account page on pyup says "Please upgrade your account to the Business or Organization plan to allow organization access." - I don't know if that's new and we need to upgrade something or it's always said that and it only applies to private/commercial projects)

The alternative to fixing pyup is switching it out for another tool/service.

bobsilverberg:
Fixes #14088 

Note that I didn't remove the `pyup` comments from `requirements/prod.txt` yet, as it might be worth keeping them there for now, until we're sure we have a working solution.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,26 @@
+version: 1
+update_configs:
+  - package_manager: "python"
+    directory: ""/requirements""
+    update_schedule: "daily"
+    ignored_updates:
+      - match:
+          dependency_name: "Django"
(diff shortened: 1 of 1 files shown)


bobsilverberg:
I think we need to merge this to verify it works.

Issue #14127, opened 2020-04-30
GitError: reference 'refs/heads/master' not found


willdurand:
It looks like we have an extra `self` here: https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/git.py#L253. It usually works because `init_repository` creates the git repository and when we call `self.git_repository` on the next line, the repo is extracted so the `git_repository` method takes the "else" branch.

Yet, it does not work all the time as we can see in Sentry, likely because of timing issues/race conditions. We should remove this `self` call, it was a mistake.

We can highlight the bug in the test suite by using this patch:

```diff
diff --git a/src/olympia/lib/git.py b/src/olympia/lib/git.py
index fbe3e23541..fee329f5fd 100644
--- a/src/olympia/lib/git.py
+++ b/src/olympia/lib/git.py
@@ -263,6 +263,7 @@ class AddonGitRepository(object):
                 path=self.git_repository_path))
         else:
             git_repository = pygit2.Repository(self.git_repository_path)
+            git_repository.lookup_reference('refs/heads/master')
 
         return git_repository
```

Explanation: there must be a `master` branch when we return the already-initialized git repo.

Given that we can detect this error at runtime, we can probably raise a special error that would be handled like the `BrokenRefError` one.

---

https://sentry.prod.mozaws.net/operations/olympia-prod/issues/5525326/

```
GitError: reference 'refs/heads/master' not found
(1 additional frame(s) were not displayed)
...
  File "celery/app/trace.py", line 650, in __protected_call__
    return self.run(*args, **kwargs)
  File "olympia/amo/decorators.py", line 108, in wrapper
    return f(*args, **kw)
  File "olympia/versions/tasks.py", line 130, in extract_version_to_git
    version=version, author=author, note=note)
  File "olympia/lib/git.py", line 329, in extract_and_commit_from_version
    branch = repo.find_or_create_branch(BRANCHES[version.channel])
  File "olympia/lib/git.py", line 398, in find_or_create_branch
    name, self.git_repository.head.peel())

GitError: reference 'refs/heads/master' not found
```

ioanarusiczki:
@willdurand If qa can test the issue, please add some STR.
Thanks!

willdurand:
> @willdurand If qa can test the issue, please add some STR.
> Thanks!

@ioanarusiczki sorry. same as the other one, please upload some versions to verify the git extraction process. We cannot really simulate this particular issue in -dev.

[...]

PR #14130, opened 2020-04-30
Detect and handle missing master branch in git repos

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14127

---

This patch fixes an issue described in the link above. Because of a small
mistake in our code, some repositories have been created without a `master`
branch and that is a problem for git extraction.

With this patch, we detect this problem. Without the cron-based git extraction
feature, it does nothing more than before. The error in Sentry will be
different but that's it. With the feature enabled, we will handle this error by
deleting the git repository and re-extracting all the versions of the add-on.

### DIFF
--- a/src/olympia/git/management/commands/git_extraction.py
+++ b/src/olympia/git/management/commands/git_extraction.py
@@ -60,7 +60,7 @@ class Command(BaseCommand):
         It does not run if the add-on is locked for git extraction.
         """
         addon = entry.addon
-        log.info('Starting extraction of add-on "{}".'.format(addon.pk))
+        log.info('Starting git extraction of add-on "{}".'.format(addon.pk))
 
         # We cannot use `entry.in_progress` because we have to be sure of the
         # add-on state and `entry` might not reflect the most up-to-date
@@ -85,7 +85,7 @@ class Command(BaseCommand):
 
         if len(versions_to_extract) == 0:
             log.info(
-                'No version to extract for add-on "{}", '
+                'No version to git-extract for add-on "{}", '
                 'exiting.'.format(addon.pk)
             )
             # We can safely delete the entry because there is no version to
@@ -104,7 +104,7 @@ class Command(BaseCommand):
         tasks.append(remove_git_extraction_entry.si(addon.pk))
 
         log.info(
-            'Submitted {} tasks to extract {} versions for add-on '
+            'Submitted {} tasks to git-extract {} versions for add-on '
             '"{}".'.format(len(tasks), len(versions_to_extract), addon.pk)
         )
         # Attach an error handler on the chain and run it. The error
--- a/src/olympia/git/tasks.py
+++ b/src/olympia/git/tasks.py
@@ -2,7 +2,11 @@ import olympia.core.logger
 
 from olympia.amo.celery import task
 from olympia.amo.decorators import use_primary_db
-from olympia.lib.git import AddonGitRepository, BrokenRefError
+from olympia.lib.git import (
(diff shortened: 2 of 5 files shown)


src/olympia/git/management/commands/git_extraction.py
```diff
@@ -104,7 +104,7 @@ def extract_addon(self, entry):
         tasks.append(remove_git_extraction_entry.si(addon.pk))
 
         log.info(
-            'Submitted {} tasks to extract {} versions for add-on '
+            'Submitted {} tasks to git-extract {} versions for add-on '
```
willdurand, line 107: I added a few "git" in this file to ease my life in Stackdriver.

src/olympia/lib/git.py
```diff
@@ -250,7 +254,7 @@ def git_repository(self):
                 path=self.git_repository_path,
                 bare=False)
             # Write first commit to 'master' to act as HEAD
-            tree = self.git_repository.TreeBuilder().write()
+            tree = git_repository.TreeBuilder().write()
```
willdurand, line 257: That's the root cause of the issue.

src/olympia/git/tasks.py
```diff
@@ -27,18 +31,32 @@ def remove_git_extraction_entry(addon_pk):
 def on_extraction_error(request, exc, traceback, addon_pk):
     log.error('Git extraction failed for add-on "{}".'.format(addon_pk))
 
+    # We only handle *some* errors here because we cannot apply the same
+    # approach to recover from all possible errors. Our current technique to
+    # repair the errors below is to delete the git repository and let the
+    # add-on being re-extracted later.
```
bobsilverberg, line 37: Nit: s/being/be

src/olympia/git/tasks.py
```diff
@@ -27,18 +31,32 @@ def remove_git_extraction_entry(addon_pk):
...
-        # We only handle `BrokenRefError` here to recover from such errors and
-        # we cannot apply the same approach for all errors.
-        # See: https://github.com/mozilla/addons-server/issues/13590
+        delete_repo = True
+        log.warning(
+            'Deleting the git repository for add-on "{}" because we detected '
+            'a broken reference.'.format(addon_pk)
+        )
+    # See: https://github.com/mozilla/addons-server/issues/14127
+    if isinstance(exc, MissingMasterBranchError):
```
bobsilverberg, line 48: Both of these cases do the exact same thing, with the only difference being the message logged. Would it make more sense to combine them into a single case which executes if either exception occurs, and if it's really important to have different log messages then deal with that when building the log message?

src/olympia/git/tests/test_tasks.py
```diff
@@ -63,6 +67,28 @@ def test_handles_broken_ref_errors(self):
...
+        # Simulate a git extraction in progress.
+        GitExtractionEntry.objects.create(addon_id=addon.pk, in_progress=True)
+        # This is the error raised by the task that extracts a version.
+        exc = MissingMasterBranchError('cannot find master branch')
+
+        on_extraction_error(
+            request=None, exc=exc, traceback=None, addon_pk=addon.pk
+        )
+
+        # The task should remove the git repository on BrokenRefError.
```
bobsilverberg, line 85: s/BrokenRefError/MissingMasterBranchError

src/olympia/git/tasks.py
willdurand: I could not find a way to have different log messages in a cleaner way and those logs are very important for debugging.

PR #14134, opened 2020-05-01
default BlocklistSubmission list filter to signoff_state=Pending

### LINKED ISSUE #14014: Filter out non-pending Blocklistsubmission instances from the admin list view by default
`Blocklistsubmission` records are never deleted (currently, #13278), even after the records have been approved and the submission expanded to `Block` objects.  We should have a default filter so when users visit the page they just see the outstanding submissions that need approval.

These a the 5* states of a `Blocklistsubmission`
```
        SIGNOFF_PENDING: 'Pending',
        SIGNOFF_APPROVED: 'Approved',
        SIGNOFF_REJECTED: 'Rejected',
        SIGNOFF_NOTNEEDED: 'No Sign-off',
        SIGNOFF_PUBLISHED: 'Published to Blocks',
```

@wagnerand would you prefer the default to be filtering _out_ `SIGNOFF_PUBLISHED` or just filtering to `SIGNOFF_PENDING` ?

 *(`SIGNOFF_NOTNEEDED` and `SIGNOFF_APPROVED` are interim states used just after the instance has been saved; straight after a celery task is started and at the end of the task the state will be `SIGNOFF_PUBLISHED`.)

eviljeff:
fixes #14014

### DIFF
--- a/src/olympia/blocklist/admin.py
+++ b/src/olympia/blocklist/admin.py
@@ -7,6 +7,7 @@ from django.template.loader import render_to_string
 from django.template.response import TemplateResponse
 from django.urls import path
 from django.utils.html import format_html
+from django.utils.translation import ugettext
 
 from olympia.activity.models import ActivityLog
 from olympia.addons.models import Addon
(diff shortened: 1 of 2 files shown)


src/olympia/blocklist/admin.py
```diff
@@ -43,6 +44,37 @@ def _get_version_choices(block, field_name):
     return choices
 
 
+class BlocklistSubmissionStateFilter(admin.SimpleListFilter):
```
bobsilverberg, line 47: Wow, that's a lot of code just to switch from having a filter without a default to having a filter with a default!

eviljeff: yeah, I know.  I tried to write it as generically as possible so it could be reused but it'd be another issue to see if some of the other Admin classes could be rewritten to use it too (e.g. the scanners admin uses a number of ListFilter classes - but they're doing other things as well changing the default)

PR #14142, opened 2020-05-01
run import_blocklist on cron and update/delete existing blocks

### LINKED ISSUE #14094: Automate the import of v2 blocks into AMO
Running v2 from AMO's data had two potential issues: 

- Perf impact from multiple records being created rather than a single block of guids in a larger regex.
- Issues with dual-sign off being required in two places.

To workaround this and minimize the risks to the V2 blocklist during the overlap with v3, we are going to keep the existing process of setting up blocks in v2.

This means we will need to import v2 blocks into AMO at an interval rather than as a one-off.

The risks around this are if the import fails then the v3 blocklist won't get updated correctly. So if any part of this process fails errors should result in an alert so we know.

To minimize the chances of problems we will look to avoid unpredictable custom regexes and changes to the format that's currently used.

## Edits/Deletions

We'll need to consider edits and deletions in so much as how we will handle them. 

For imports I think it would make things simpler to not require the import to spot changes to historical data, and for admins handle that manually on a case by case basis in changes to AMO as they happen. This is based on these kind of changes having been fairly infrequent in the past.

It's important that if something is blocked and then that block is edited/removed later in v2 records (and assuming a manual update to AMO), that the the import won't reverse that manual change (thanks to Philipp for raising this point).

eviljeff:
fixes #14094


eviljeff:
fwiw, I can't see how the test fail is connected to my changes at all - the import_blocklist code is self contained (no link to `TestRecommendedQueue.test_nothing_recommended_filtered_out` )

willdurand:
> fwiw, I can't see how the test fail is connected to my changes at all - the import_blocklist code is self contained (no link to `TestRecommendedQueue.test_nothing_recommended_filtered_out` )

Yes, it happens on `master` as well and I filed https://github.com/mozilla/addons-server/issues/14146.

src/olympia/blocklist/tasks.py
```diff
@@ -144,9 +151,34 @@ def import_block_from_blocklist(record):
...
+
     kinto_import.save()
 
 
+@task
+@use_primary_db
+@transaction.atomic
+def delete_imported_block_from_blocklist(kinto_id):
+    existing_blocks = (
+        Block.objects.filter(kinto_id__in=(kinto_id, f'*{kinto_id}')))
```
willdurand, line 173: Why do we have to prepend the id with `*`?

src/olympia/blocklist/management/commands/import_blocklist.py
```diff
@@ -14,20 +16,40 @@
...
+        log.debug(
+            '%s new, %s modified, % deleted records from blocklist to process',
+            len(new_records), len(modified_records), len(deleted_record_ids))
+        for record in new_records + modified_records:
             import_block_from_blocklist.delay(record)
+        if deleted_record_ids:
+            log.debug(
+                'Deleting Blocks that have been removed from v2 blocklist')
+            for kinto_id in deleted_record_ids:
+                delete_imported_block_from_blocklist.delay(kinto_id)
```
willdurand, line 55: In the issue, there is this paragraph:

> It's important that if something is blocked and then that block is edited/removed later in v2 records (and assuming a manual update to AMO), that the the import won't reverse that manual change (thanks to Philipp for raising this point).

Is this guaranteed?

src/olympia/blocklist/management/commands/import_blocklist.py
```diff
@@ -14,20 +16,40 @@
...
-            'Downloading blocklist from %s', self.KINTO_JSON_BLOCKLIST_URL)
-        response = requests.get(self.KINTO_JSON_BLOCKLIST_URL)
-
+        LEGACY_BLOCKLIST_URL = (
+            settings.REMOTE_SETTINGS_API_URL +
+            'buckets/blocklists/collections/addons/records')
+        log.debug('Downloading blocklist from %s', LEGACY_BLOCKLIST_URL)
+        response = requests.get(LEGACY_BLOCKLIST_URL)
+
+        data = response.json()['data']
```
willdurand, line 26: I feel like this isn't really safe.. I ran the command locally and 💥

```
{"Timestamp": 1588694965152747264, "Type": "z.amo.blocklist", "Logger": "http_app_addons", "Hostname": "70c3b12064a3", "EnvVersion": "2.0", "Severity": 7, "Pid": 51, "Fields": {"uid": "<anon>", "remoteAddressChain": "", "msg": "Downloading blocklist from https://kinto.dev.mozaws.net/v1/buckets/blocklists/collections/addons/records"}, "severity": 100}
  File "./manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/deps/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/deps/lib/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/deps/lib/python3.6/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/deps/lib/python3.6/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/code/src/olympia/blocklist/management/commands/import_blocklist.py", line 26, in handle
    data = response.json()['data']
KeyError: 'data'
```

src/olympia/blocklist/models.py
```diff
@@ -554,3 +554,4 @@ class KintoImport(ModelBase):
     record = JSONField(default={})
     outcome = models.SmallIntegerField(
         default=OUTCOME_INCOMPLETE, choices=OUTCOMES.items())
+    timestamp = models.BigIntegerField()
```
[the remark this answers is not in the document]
eviljeff: the timestamp from kinto is an integer (the unix epoc * 1000)

src/olympia/blocklist/utils.py
```diff
@@ -56,7 +56,8 @@ def block_activity_log_save(obj, change, submission_obj=None):
     add_version_log_for_blocked_versions(obj, al)
 
 
-def block_activity_log_delete(obj, submission_obj):
+def block_activity_log_delete(obj, *, submission_obj=None, delete_user=None):
```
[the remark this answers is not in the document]
eviljeff: heh, it's the future!

src/olympia/blocklist/management/commands/import_blocklist.py
```diff
@@ -14,20 +16,40 @@
...
+            record for record in data
+            if record['id'] in already_imported and
+            record['last_modified'] != already_imported[record['id']]
+        ]
+        deleted_record_ids = [
+            kinto_id for kinto_id in already_imported
+            if kinto_id not in kinto_ids
+        ]
+        log.debug(
+            '%s new, %s modified, % deleted records from blocklist to process',
```
willdurand, line 47:
```suggestion
            '%s new, %s modified, %s deleted records from blocklist to process',
```

src/olympia/blocklist/management/commands/import_blocklist.py
eviljeff: It's not possible for that to happen.  `Block` creations, deletions, and edits don't trigger changes to the v2 collection - creating and signing off a `BlocklistSubmission` does.  We don't create `BlocklistSubmission` instances with this command.

eviljeff: * it would if the waffle switch was enabled anyway.  The code behind the waffle switch is obsolete now but we're not removing it until it's clear what's happening in September after the v2 blocklist is deprecated.  With the waffle switch disabled (the default) there are no updates to the v2 blocklist from AMO under any circumstances.

src/olympia/blocklist/tasks.py
eviljeff: We don't have to (it's an `__in` with `kinto_id` too) but it's one of the ways.  Prefixing with `*` is the magic code to indicate that the Block came from a regular expression in the guid field in a v2 block so they'll be (potentially) multiple Blocks on AMO from a single record in the v2 blocklist.

eviljeff: See `Block.is_imported_from_kinto_regex`

willdurand: ah thanks, I imagined something like this but I did not know the reasons behind that.

src/olympia/blocklist/management/commands/import_blocklist.py
eviljeff: https://github.com/mozilla/addons-server/pull/14142/commits/b145f51565b37e8bad706c535466af145dcfbb4b

src/olympia/conf/stage/settings.py
```diff
@@ -124,4 +124,5 @@
 EXTENSION_WORKSHOP_URL = env(
     'EXTENSION_WORKSHOP_URL', default='https://extensionworkshop.allizom.org')
 
-KINTO_API_URL = 'https://settings-writer.stage.mozaws.net/v1/'
+REMOTE_SETTINGS_API_URL = 'https://firefox.settings.services.mozilla.com/v1/'
```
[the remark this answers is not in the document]
eviljeff: 🙄 https://github.com/mozilla/addons-server/pull/14142/commits/b145f51565b37e8bad706c535466af145dcfbb4b

PR #14157, opened 2020-05-04
Refactor l10n extraction script to be ran in either manual or automatic mode

### LINKED ISSUE #14155: Make manual extraction of locales easier
Eventually we'd like to automate l10n extraction as much as possible in https://github.com/mozilla/addons-server/issues/9112 but first let's make it easier to run without any special environment variables. For reference, the frontend script is https://github.com/mozilla/addons-frontend/blob/master/bin/run-l10n-extraction

diox:
(Remove travis support for now as it never worked)

Fixes #14155


diox:
Yeah I preferred to explicitly require python 3 for people that might want to use this in a virtualenv (which means `git` is working with regular config) since it would likely break with `python` binary pointing to python 2.

PR #14158, opened 2020-05-04
compare blocked versions to detect block deletions too

### LINKED ISSUE #14107: Only block deletions won't trigger a new mlbf
The logic for the cron mlbf job currently relies on checking the latest modified date of the blocks in the database and skipping generation unless it's greater than (more recent) than the last generation time.  This works great for block additions and changes but wouldn't catch a block deletion because it wouldn't be in the database.

There could be some other flag we can set when blocks are deleted, though it's probably more stable to generate the list of blocked `guid:version`s and compare them to the last generation instead (at the expense of an extra database query and reading a JSON file from disk).

eviljeff:
fixes #14107 - the other solution to doing this is to load the previous blocked json and I'm not sure if that is a better idea. We'd have to load and parse an extra JSON file but we wouldn't have to keep track of a config value.

### DIFF
--- a/src/olympia/blocklist/cron.py
+++ b/src/olympia/blocklist/cron.py
@@ -8,12 +8,13 @@ from olympia.zadmin.models import get_config
 from .mlbf import MLBF
 from .models import Block
 from .tasks import (
-    MLBF_BASE_ID_CONFIG_KEY, MLBF_TIME_CONFIG_KEY, upload_filter_to_kinto)
+    MLBF_BASE_ID_CONFIG_KEY, MLBF_COUNT_CONFIG_KEY, MLBF_TIME_CONFIG_KEY,
+    upload_filter_to_kinto)
 
 log = olympia.core.logger.getLogger('z.cron')
 
 
-def _get_blocklist_last_modified_time():
+def get_blocklist_last_modified_time():
     latest_block = Block.objects.order_by('-modified').first()
     return int(latest_block.modified.timestamp() * 1000) if latest_block else 0
 
@@ -23,10 +24,7 @@ def upload_mlbf_to_kinto():
         log.info('Upload MLBF to kinto cron job disabled.')
         return
     last_generation_time = get_config(MLBF_TIME_CONFIG_KEY, 0, json_value=True)
-    if last_generation_time > _get_blocklist_last_modified_time():
-        log.info(
-            'No new/modified Blocks in database; skipping MLBF generation')
-        return
+    last_blocked_count = get_config(MLBF_COUNT_CONFIG_KEY, 0, json_value=True)
 
     log.info('Starting Upload MLBF to kinto cron job.')
 
@@ -39,6 +37,16 @@ def upload_mlbf_to_kinto():
     # https://github.com/mozilla/addons-server/issues/13695
     generation_time = int(time.time() * 1000)
     mlbf = MLBF(generation_time)
+
+    need_mlbf = (
+        last_generation_time < get_blocklist_last_modified_time() or
+        last_blocked_count != len(mlbf.blocked_versions))
+    if not need_mlbf:
+        log.info(
+            'No new/modified/deleted Blocks in database; '
+            'skipping MLBF generation')
+        return
+
     mlbf.generate_and_write_mlbf()
 
     base_filter_id = get_config(MLBF_BASE_ID_CONFIG_KEY, 0, json_value=True)
--- a/src/olympia/blocklist/cron.py
+++ b/src/olympia/blocklist/cron.py
@@ -8,8 +8,7 @@ from olympia.zadmin.models import get_config
 from .mlbf import MLBF
 from .models import Block
 from .tasks import (
-    MLBF_BASE_ID_CONFIG_KEY, MLBF_COUNT_CONFIG_KEY, MLBF_TIME_CONFIG_KEY,
-    upload_filter_to_kinto)
+    MLBF_BASE_ID_CONFIG_KEY, MLBF_TIME_CONFIG_KEY, upload_filter_to_kinto)
 
 log = olympia.core.logger.getLogger('z.cron')
 
@@ -24,7 +23,6 @@ def upload_mlbf_to_kinto():
         log.info('Upload MLBF to kinto cron job disabled.')
         return
     last_generation_time = get_config(MLBF_TIME_CONFIG_KEY, 0, json_value=True)
-    last_blocked_count = get_config(MLBF_COUNT_CONFIG_KEY, 0, json_value=True)
 
     log.info('Starting Upload MLBF to kinto cron job.')
 
@@ -37,10 +35,11 @@ def upload_mlbf_to_kinto():
     # https://github.com/mozilla/addons-server/issues/13695
     generation_time = int(time.time() * 1000)
     mlbf = MLBF(generation_time)
+    previous_filter = MLBF(last_generation_time)
 
     need_mlbf = (
         last_generation_time < get_blocklist_last_modified_time() or
-        last_blocked_count != len(mlbf.blocked_versions))
+        mlbf.blocks_changed_since_previous(previous_filter))
     if not need_mlbf:
         log.info(
             'No new/modified/deleted Blocks in database; '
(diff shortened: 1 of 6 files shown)


src/olympia/blocklist/cron.py
```diff
@@ -23,10 +24,7 @@ def upload_mlbf_to_kinto():
         log.info('Upload MLBF to kinto cron job disabled.')
         return
     last_generation_time = get_config(MLBF_TIME_CONFIG_KEY, 0, json_value=True)
-    if last_generation_time > _get_blocklist_last_modified_time():
-        log.info(
-            'No new/modified Blocks in database; skipping MLBF generation')
-        return
+    last_blocked_count = get_config(MLBF_COUNT_CONFIG_KEY, 0, json_value=True)
 
     log.info('Starting Upload MLBF to kinto cron job.')
```
bobsilverberg, line 29: Would it make sense to move this `log` statement below the new `if` check, so that it doesn't appear in the log if we choose not to upload?

src/olympia/blocklist/cron.py
```diff
@@ -23,10 +24,7 @@ def upload_mlbf_to_kinto():
         log.info('Upload MLBF to kinto cron job disabled.')
         return
     last_generation_time = get_config(MLBF_TIME_CONFIG_KEY, 0, json_value=True)
-    if last_generation_time > _get_blocklist_last_modified_time():
-        log.info(
-            'No new/modified Blocks in database; skipping MLBF generation')
-        return
+    last_blocked_count = get_config(MLBF_COUNT_CONFIG_KEY, 0, json_value=True)
```
bobsilverberg, line 27: This is only used once, so it could be inlined in the expression for `need_mlbf`, but it does make the code a bit more obvious to have this named variable, so I think it's fine either way.

src/olympia/blocklist/cron.py
eviljeff: we could, but I left it because that's the point when we define `generation_time` (and we start doing database queries)

eviljeff:
@bobsilverberg I rewrote all the things :)  It's not that expensive to read a JSON file from disk but I had to move things around and add caching so we weren't reading things twice in the normal case where we have to write a bloomfilter + stash.

src/olympia/blocklist/cron.py
```diff
@@ -39,21 +35,32 @@ def upload_mlbf_to_kinto():
...
+
     mlbf.generate_and_write_mlbf()
 
     base_filter_id = get_config(MLBF_BASE_ID_CONFIG_KEY, 0, json_value=True)
-    base_filter = MLBF(base_filter_id) if base_filter_id else None
+    # optimize for when the base_filter was the previous generation so
+    # we don't have to load the blocked JSON file twice.
+    base_filter = (
+        MLBF(base_filter_id)
+        if last_generation_time != base_filter_id else
```
bobsilverberg, line 56: Why are we comparing `last_generation_time` to `base_filter_id`? Is that correct? I wouldn't expect to compare a time to an id.

src/olympia/blocklist/mlbf.py
```diff
@@ -127,34 +142,38 @@ def not_blocked_json(self):
...
 
+    def fetch_not_blocked_json(self):
+        """A safer version of .not_blocked_json that will generate from the db
+        if the cached_property isn't set."""
+        try:
+            return self.not_blocked_json
+        except FileNotFoundError:
+            # when self._not_blocked_path doesn't exist
+            pass
+        # see fetch_blocked_json
```
bobsilverberg, line 153: This feels a bit ugly. The fact that you need this comment signals to me that these functions are coupled in a way that is non-intuitive. I wonder if a shared helper would make thus a bit cleaner?

src/olympia/blocklist/mlbf.py
```diff
@@ -127,34 +142,38 @@ def not_blocked_json(self):
...
+        return self.not_blocked_json
+
     @property
-    def stash_path(self):
+    def _stash_path(self):
         return os.path.join(
             settings.MLBF_STORAGE_PATH, self.id, 'stash.json')
 
     @cached_property
     def stash_json(self):
```
bobsilverberg, line 166: Nit: Should this maybe be `stashed_json` or `json_stash` instead of `stash_json`? `stash_json` sounds like a method, not a property.

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -185,13 +187,19 @@ def test_waffle_off_disables_publishing(self):
...
         assert not get_config(MLBF_TIME_CONFIG_KEY)
 
-    def test_no_block_changes(self):
+    @freeze_time('2020-01-01 12:34:56', as_arg=True)
+    def test_no_block_changes(frozen_time, self):
         # This was the last time the mlbf was generated
         last_time = int(
             datetime.datetime(2020, 1, 1, 12, 34, 1).timestamp() * 1000)
         # And the Block was modified just before so would be included
         self.block.update(modified=datetime.datetime(2020, 1, 1, 12, 34, 0))
```
bobsilverberg, line 196: Instead of relying on two hardcoded datetimes (and introducing the possibility of a typo), could this datetime be calculated based on `last_time`?

src/olympia/blocklist/mlbf.py
```diff
@@ -61,7 +61,7 @@ def get_blocked_versions(cls):
         return all_versions
 
     @classmethod
-    def get_all_guids(cls, excluding_version_ids=None):
+    def fetch_all_addons_from_db(cls, excluding_version_ids=None):
```
bobsilverberg, line 64: The name of the thing being fetched changed from `guids` to `addons`. Is that an improvement? I notice that tests that call this still store the returned value in a variable named, for example, `all_guids`.

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -69,7 +69,7 @@ def setUp(self):
             min_version='9999')
 
     def test_all_guids(self):
```
bobsilverberg, line 71: It looks like this test case needs to be renamed, like you did below with `test_fetch_blocked_from_db`.

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -230,19 +232,21 @@ def test_write_stash(self):
         full_stash = {
             'blocked': ['fooo@baaaa:999'],
             'unblocked': [f'{self.five_ver.guid}:123.5']}
-        with open(newer_mlbf.stash_path) as stash_file:
+        with open(newer_mlbf._stash_path) as stash_file:
             assert json.load(stash_file) == full_stash
         assert newer_mlbf.stash_json == full_stash
 
     def test_should_reset_base_filter(self):
```
bobsilverberg, line 239: Should this test case be renamed to reflect that it's also testing `blocks_changed_since_previous`?

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -185,13 +187,19 @@ def test_waffle_off_disables_publishing(self):
...
         set_config(MLBF_TIME_CONFIG_KEY, last_time, json_value=True)
+        prev_blocked_path = os.path.join(
+            settings.MLBF_STORAGE_PATH, str(last_time), 'blocked.json')
+        with storage.open(prev_blocked_path, 'w') as blocked_file:
```
bobsilverberg, line 200: What exactly is happening here?

src/olympia/blocklist/cron.py
eviljeff: `last_generation_time` is used as an id (and `base_filter_id` would be from a time too).  Should I rename `last_generation_time` to make it clearer?

src/olympia/blocklist/mlbf.py
eviljeff: before I had a cached_property around `fetch_blocked_json` but I stripped it out because cached properties relying on cached properties tends to end up with annoying end cases. (though as we don't change them after being set here that could work still...)

src/olympia/blocklist/mlbf.py
eviljeff: it's the same naming scheme as `_blocked_path`, `blocked_json` (which reads `blocked.json`).  Could though?

src/olympia/blocklist/tests/test_cron.py
eviljeff: now we're using the json of previously blocked guid:version to compare to, the file needs to exist so it can be read.

src/olympia/blocklist/mlbf.py
eviljeff: yeah, I don't know.  It's not strictly just guids, it's a list of `(guid:version)` tuples.  `fetch_all_versions_from_db` might be more accurate.

src/olympia/blocklist/cron.py
eviljeff: It got removed in the revision anyway

src/olympia/blocklist/tests/test_cron.py
eviljeff: https://github.com/mozilla/addons-server/pull/14158/commits/ecdf64d6ae42063feacab214f5f4aff24141818c

src/olympia/blocklist/tests/test_mlbf.py
eviljeff: https://github.com/mozilla/addons-server/pull/14158/commits/ecdf64d6ae42063feacab214f5f4aff24141818c

src/olympia/blocklist/tests/test_mlbf.py
eviljeff: https://github.com/mozilla/addons-server/pull/14158/commits/ecdf64d6ae42063feacab214f5f4aff24141818c

PR #14159, opened 2020-05-04
Fix ActivityLog/UserRestrictionHistory not saved for API 403/429 errors

### LINKED ISSUE #13943: Add ActivityLog when a user is rate-limited
When a user is rate-limited we should add that to an `ActivityLog` tied to the user (and the add-on, if available). We should also log those events to stackdriver.

### LINKED ISSUE #14039: Override user when creating `ActivityLog` in submission-related tasks

diox:
Fix `ActivityLog`/`UserRestrictionHistory` not saved for API 403/429 errors
     
DRF's default exception handler was rolling back the current transaction when those errors were triggered, causing any database inserts to not go through.

Fixes #14039
Fixes #13943

### DIFF
--- a/src/olympia/signing/views.py
+++ b/src/olympia/signing/views.py
@@ -137,13 +137,11 @@ class VersionView(APIView):
         """
         If request is not permitted, determine what kind of exception to raise.
 
-         (Lifted from DRF, but also passing the optional code argument to
-          the PermissionDenied exception)
+        (Lifted from DRF, but also passing the optional code argument to
+        the PermissionDenied exception)
         """
         if request.authenticators and not request.successful_authenticator:
             raise exceptions.NotAuthenticated()
-        # If I can't figure out how to properly get the code without breaking
-        # stuff, maybe I can cheat and return an entirely different class here
         raise exceptions.PermissionDenied(
             detail=message, code=code)
 
--- a/src/olympia/signing/views.py
+++ b/src/olympia/signing/views.py
@@ -3,6 +3,7 @@ import functools
 from django import forms
 from django.utils.translation import ugettext
 
+from rest_framework import exceptions
 from rest_framework import status
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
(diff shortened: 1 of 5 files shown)


src/olympia/signing/views.py
```diff
@@ -113,6 +114,39 @@ def check_throttles(self, request):
             return
         super().check_throttles(request)
 
+    # When DRF 3.12 is released, we can remove the custom check_permissions()
+    # and permission_denied() as it will contain the fix for
+    # https://github.com/encode/django-rest-framework/issues/7038
```
willdurand, line 119: Should we have an issue in addons-server for that, too?

src/olympia/signing/views.py
```diff
@@ -113,6 +114,39 @@ def check_throttles(self, request):
...
+                self.permission_denied(
+                    request, message=getattr(permission, 'message', None),
+                    code=getattr(permission, 'code', None),
+                )
+
+    def permission_denied(self, request, message=None, code=None):
+        """
+        If request is not permitted, determine what kind of exception to raise.
+
+         (Lifted from DRF, but also passing the optional code argument to
```
willdurand, line 140: Nit: wrong indentation.

src/olympia/signing/views.py
```diff
@@ -113,6 +114,39 @@ def check_throttles(self, request):
...
+          the PermissionDenied exception)
+        """
+        if request.authenticators and not request.successful_authenticator:
+            raise exceptions.NotAuthenticated()
+        # If I can't figure out how to properly get the code without breaking
+        # stuff, maybe I can cheat and return an entirely different class here
```
willdurand, line 146: s/I/we

diox: Ooops that comment shouldn't be there anymore since I did figure out a way to make the exception handler work. I'll remove it.

src/olympia/signing/views.py
diox: https://github.com/mozilla/addons-server/issues/14178

Issue #14160, opened 2020-05-04
Show scanner "feedback buttons" when there is no match


willdurand:
Currently we hide the feedback buttons in the django admin when a scanner does not have any matched rules. It would be helpful to have the buttons even when there is no match.

AlexandraMoga:
The feedback buttons are now available for each scanner result,  with or without matched rules. Their functionality remains unchanged.
Note that the action buttons apply only to `yara` and `customs`

willdurand:
>  Note that the action buttons apply only to `yara` and `customs`

Yes, that's perfect. Thanks!

Issue #14146, opened 2020-05-01
`TestRecommendedQueue.test_nothing_recommended_filtered_out` is failing


willdurand:
See: https://travis-ci.org/github/mozilla/addons-server/jobs/682003278#L3092

```
=================================== FAILURES ===================================

__________ TestRecommendedQueue.test_nothing_recommended_filtered_out __________

[gw0] linux -- Python 3.6.7 /home/travis/build/mozilla/addons-server/.tox/reviewers-and-zadmin/bin/python3

src/olympia/reviewers/tests/test_views.py:1790: in test_nothing_recommended_filtered_out

    self._test_results()

        self       = <olympia.reviewers.tests.test_views.TestRecommendedQueue testMethod=test_nothing_recommended_filtered_out>

        version    = <Version: 0.1>

src/olympia/reviewers/tests/test_views.py:1174: in _test_results

    check_links(expected, links, verify=False)

        addon      = <Addon: 3840: Pending Two>

        channel    = []

        doc        = [<html>]

        expected   = [('Nominated One 0.1', '/en-US/reviewers/review/nominated-one'),

 ('Nominated Two 0.1', '/en-US/reviewers/review/nominated-two'),

 ('Pending One 0.1', '/en-US/reviewers/review/pending-one'),

 ('Pending Two 0.1', '/en-US/reviewers/review/pending-two')]

        idx        = 3

        latest_version = <Version: 0.1>

        links      = [<a>, <a>, <a>, <a>]

        name       = 'Pending Two 0.1'

        response   = <HttpResponse status_code=200, "text/html; charset=utf-8">

        self       = <olympia.reviewers.tests.test_views.TestRecommendedQueue testMethod=test_nothing_recommended_filtered_out>

        url        = '/en-US/reviewers/review/pending-two'

src/olympia/amo/tests/__init__.py:199: in check_links

    f'At index {idx}, expected {text}, got {e.text()}')

E   AssertionError: At index 0, expected Nominated One 0.1, got Pending Two 0.1

        e          = [<a>]

        elements   = [<a>, <a>, <a>, <a>]

        expected   = [('Nominated One 0.1', '/en-US/reviewers/review/nominated-one'),

 ('Nominated Two 0.1', '/en-US/reviewers/review/nominated-two'),

 ('Pending One 0.1', '/en-US/reviewers/review/pending-one'),

 ('Pending Two 0.1', '/en-US/reviewers/review/pending-two')]

        idx        = 0

[...]

PR #14162, opened 2020-05-04
Show scanner feedback buttons even when there is no match

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14160


willdurand:
Test failure is #14146

PR #14163, opened 2020-05-04
Update Browse and Compare API endpoints to allow for `file_only`

### LINKED ISSUE #14020: Allow Browse and Compare APIs to return just file data

bobsilverberg:
Fixes #14020 

This adds a new URL param to the Browse and Compare endpoints called `file_only`. When that is set to `true` then only file data is returned to the client. The file data is in a top-level property called `file`, and none of the other `version` properties are returned in the API response.

### DIFF
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -1274,7 +1274,6 @@ class AddonReviewerViewSet(GenericViewSet):
         return Response(status=status_code)
 
 
-# TODO Some helpers for the APIs via mixins
 class ReviewAddonVersionMixin(object):
     permission_classes = [AnyOf(
         AllowReviewer, AllowReviewerUnlisted, AllowAddonAuthor,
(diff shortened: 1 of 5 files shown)


bobsilverberg:
Thanks for taking a look at this @eviljeff. I have changed everything, using a URL param with the existing endpoints, and it's made for a lot less code. I also updated the API docs. I think the result is perhaps not perfect, but good enough, considering how much work it will save both on addons-server and on code-manager.

src/olympia/reviewers/views.py
```diff
@@ -1362,6 +1363,18 @@ def list(self, request, *args, **kwargs):
         return Response(serializer.data)
 
     def retrieve(self, request, *args, **kwargs):
+        if self.request.GET.get('file_only', False):
+            version = self.get_object()
```
eviljeff, line 1367: nit: we need to do this regardless, so `version = self.get_object()` could be outside of the `if` and used below for `AddonBrowseVersionSerializer` too.

Issue #14161, opened 2020-05-04
Scanners API should return results that have been previously edited by the task user


willdurand:
We should return scanner results that have been auto-approved by the task user and later approved by a reviewer (auto approval confirmed). This likely does not work currently because we exclude scanner results with an activity log containing entries created by the task user.

We only want to exclude the activity log entries created by the task user.

AlexandraMoga:
@willdurand I'm not sure I fully understand this issue. Should I look for scanner results that are not returned by the API and then confirm approval for those versions and re-verify that the results are returned by the API this time?

willdurand:
> should I look for scanner results that are not returned by the API and then confirm approval for those versions and re-verify that the results are returned by the API this time?

Exactly, yes. Sorry that was not clear (I was tired when I filed the issue...)

Before the patch for this issue, no scanner results were returned when auto approval was confirmed by a reviewer it seems.

AlexandraMoga:
I've tested the following scenarios:
- scanner result for _listed_ add-on, auto-approved => results returned by the API after reviewer confirms approval 
- scanner result for _unlisted_ add-on, auto-approved => results returned by the API after reviewer confirms approval for selected versions (note that for unlisted add-ons, auto-approval can be confirmed for specific versions, so the API will obviously return the results only for the version for which auto-approval was confirmed)
- scanner result for theme or add-on that doesn't go through the auto-approval flow => results returned by the API after reviewer _manually approves_ the add-on/theme

@willdurand just to confirm, we want the API to return results for an add-on that doesn't go through auto-approval, but it is approved manually by reviewers (see last scenario)?

willdurand:
> @willdurand just to confirm, we want the API to return results for an add-on that doesn't go through auto-approval, but it is approved manually by reviewers (see last scenario)?

Yes.

So, everything looks good to me, thanks!

PR #14164, opened 2020-05-04
Fix query for scanner results

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14161

---

This should allow to select scanner results bound to versions that have
activity log entries created by the task user and another user.


willdurand:
> just needs some reformatting to make travis happy.

ah oops, thanks

PR #14172, opened 2020-05-05
Fix missing newline in french translation

diox:


diox:
(We should probably remove that newline from the template - but I want to fix the build first)

Issue #14084, opened 2020-04-28
Flag for review for some of the MAD results


willdurand:
We'd like to flag for review a version for which the MAD result has:

- a score equal to `0%`
- a score equal to `100%`
- a model disagreement

diox:
Note: ideally it would live in a separate model, a bit like `AddonReviewerFlags` but for versions. Cause `ALTER` on `versions` is costly, and also extra fields we don't need 99% of the time are costly since by default Django requests them all all the time...

willdurand:
> Note: ideally it would live in a separate model, a bit like `AddonReviewerFlags` but for versions. Cause `ALTER` on `versions` is costly, and also extra fields we don't need 99% of the time are costly since by default Django requests them all all the time...

@diox can't we reuse the flag for review action somehow? or maybe refactor it so that it contains a list of scanners that have set this flag?

diox:
It's currently a boolean column on `versions` table, so if we want to avoid the costly `ALTER`, while still providing the extra info, it's going to be difficult... (though long-term if we do have a separate model we'll want to do the `ALTER` once to remove the current column)

willdurand:
Ok, so I will add a `VersionReviewerFlags` (or `VersionScannerFlags`) model.

AlexandraMoga:
@willdurand I'm not sure how I should verify these new MAD flags. There is no flag visible in rev tools for a version that had a score equal to `100%`  for example. 
In any case, I've run some tests to make sure the existing scanner flags are still working.

willdurand:
@AlexandraMoga we need https://github.com/mozilla/addons-server/issues/14085 to have a UI on top of this issue.

AlexandraMoga:
In that case, I'll mark this as verified, since I did not find regressions for the other scanner flags, and the MAD flags will be verified when the UI is ready.

PR #14173, opened 2020-05-05
Add automatic action for some MAD results

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14084

---

This patch adds new scanner flags that we'll use in the reviewer tools in a next iteration.


src/olympia/scanners/models.py
```diff
@@ -387,3 +421,14 @@ class Meta(AbstractScannerResult.Meta):
 class ScannerQueryMatch(ModelBase):
     result = models.ForeignKey(ScannerQueryResult, on_delete=models.CASCADE)
     rule = models.ForeignKey(ScannerQueryRule, on_delete=models.CASCADE)
+
+
+class VersionScannerFlags(ModelBase):
+    version = models.OneToOneField(
+        'versions.Version', primary_key=True, on_delete=models.CASCADE
+    )
+    needs_human_review_by_mad = models.BooleanField(default=False)
```
diox, line 430: I don't have anything better to suggest yet, but I'm not a huge fan of this name.

willdurand: Yep. I was thinking that, given that we do not have a lot of scanners, we could have one field per scanner in the future (and stop relying on the version fields in some cases)

diox: Yeah, that's not my concern. It's mostly the `by_` that I don't like: It kinda implies the version needs a human review to be done by the mad scanner, at least that how I read it the first time and I was confused :)

willdurand: oh sorry, I misunderstood.

PR #14175, opened 2020-05-05
Remove obsolete 'es-use-classic-similarity' code and waffle switch

### LINKED ISSUE #14080: Remove usage of `classic` similarity algorithm in elasticsearch
We're using the `classic` similarity algorithm in elasticsearch at the moment, which isn't the default - it's there for compatibility with old elasticsearch versions, and we set that way back in https://github.com/mozilla/addons-server/issues/8867 to try to address issues we were having with search results after upgrading.

That setting has been deprecated for a while, but now [the documentation for 6.8](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/index-modules-similarity.html#classic-similarity) mentions this:
> The quality of the produced scores used to rely on coordination factors, which have been removed. It is advised to use BM25 instead.

I've noticed the quality drop in my branch as part of [the work to upgrade to 6.8](https://github.com/mozilla/addons-server/issues/8508). We should try to anticipate this and switch to the `default` algorithm and see if it's good enough for us, we've made a number of changes to search scoring since so my hope is that we don't actually need the `classic` algorithm anymore.

diox:
The `classic` algorithm has been deprecated for a while, will have more drawbacks in next version, it's time to use the new default instead.

Fixes #14080


src/olympia/search/migrations/0001_remove-es-use-classic-similarity-waffle-switch.py
```diff
@@ -0,0 +1,13 @@
+# Generated by Django 2.2.12 on 2020-05-05 13:39
+
+from django.db import migrations
+
+
+def remove_waffle_switch(apps, schema_editor):
```
bobsilverberg, line 6: I don't see any model changes associated with this. Just curious, how did Django know to generate this migration?

diox: I generated it myself with `makemigrations --empty` and then customized it. It's not super common but that's the way to go for data migrations like this

Issue #14176, opened 2020-05-05
Avoid git-extraction loops


willdurand:
It is possible to end up in a git-extraction (infinite) loop if a recoverable error occurs over and over for the same add-on.

ioanarusiczki:
@willdurand Should I try uploading again versions for an add-on?

ioanarusiczki:
@willdurand 
Tested on https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/noscriptcron1 version 21.0 
The admin queue did now show the extraction - which is the expected result.

Version 21.0 was broken. Right now it it visible again.
https://code.addons-dev.allizom.org/en-US/browse/619736/versions/1691239/

I also uploaded version 22.0 after 21.0 and it is also visible.
https://code.addons-dev.allizom.org/en-US/browse/619736/versions/1691240/

If results are alright I can mark it verified (you expected 21.0 to be broken).

willdurand:
(heh, wrong button)

Looks good to me

PR #14177, opened 2020-05-05
Avoid git-extraction loops based on the git repo creation time

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14176

---

Theoretically, the new git-extraction process might be subject to infinite loops if the same add-on fails with a recoverable error over and over again. In this case, we would waste resources for nothing.

This patch should prevent infinite loops by making sure we are not deleting/re-creating a git repository too "soon". In order to do this, we need to know the git repository "creation time". We cannot retrieve this time from the FS (because UNIX systems don't expose that) so this patch uses a file that is present in all git repositories and that never changes (`.git/description`).

### DIFF
--- a/src/olympia/lib/git.py
+++ b/src/olympia/lib/git.py
@@ -1,13 +1,14 @@
 # -*- coding: utf-8 -*-
-import uuid
-import os
 import io
-import shutil
-import tempfile
-import sys
 import mimetypes
+import os
 import posixpath
+import shutil
+import sys
+import tempfile
+import uuid
 
+from datetime import datetime, timedelta
 from collections import namedtuple
 
 import pygit2
@@ -203,6 +204,8 @@ class TemporaryWorktree(object):
 
 class AddonGitRepository(object):
 
+    GIT_DESCRIPTION = '.git/description'
+
     def __init__(self, addon_or_id, package_type='addon'):
         from olympia.addons.models import Addon
         assert package_type in ('addon', 'source')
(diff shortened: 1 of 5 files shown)


src/olympia/lib/git.py
```diff
@@ -246,6 +249,23 @@ def __init__(self, addon_or_id, package_type='addon'):
...
+    def is_recent(self):
+        if not self.is_extracted:
+            return False
+        # A git repository is recent when it was created less than 1 hour ago.
+        an_hour_ago = datetime.utcnow() - timedelta(hours=1)
+        mtime = datetime.utcfromtimestamp(
+            os.path.getmtime(
+                # There is no way to get the creation time of a file/folder on
+                # most UNIX systems, so we use a file that is created by
+                # default by never modified (GIT_DESCRIPTION) to determine
```
eviljeff, line 262: nit "_but_ never modified"

src/olympia/lib/tests/test_git.py
```diff
@@ -1541,3 +1553,29 @@ def test_commit_with_old_reference(settings):
...
+
+    assert repo.is_recent
+
+
+def test_is_recent_with_old_repo(settings):
+    addon_pk = 123
+    repo = AddonGitRepository(addon_pk)
+    # Force the creation of the git repository.
+    repo.git_repository
+    update_git_repo_creation_time(repo, time=datetime.datetime(2020, 1, 1))
```
eviljeff, line 1579: can we test with a timedelta _just_ over an hour?

src/olympia/lib/tests/test_git.py
```diff
@@ -1541,3 +1553,29 @@ def test_commit_with_old_reference(settings):
...
+    assert not repo.is_recent
+
+
+def test_is_recent(settings):
+    addon_pk = 123
+    repo = AddonGitRepository(addon_pk)
+    # Force the creation of the git repository.
+    repo.git_repository
+
+    assert repo.is_recent
```
eviljeff, line 1571: can we also test with a timedelta that's not immediate but under an hour?

src/olympia/lib/git.py
```diff
@@ -246,6 +249,23 @@ def __init__(self, addon_or_id, package_type='addon'):
     def is_extracted(self):
         return os.path.exists(self.git_repository_path)
 
+    @property
+    def is_recent(self):
+        if not self.is_extracted:
+            return False
+        # A git repository is recent when it was created less than 1 hour ago.
+        an_hour_ago = datetime.utcnow() - timedelta(hours=1)
```
eviljeff, line 257: so, I agree this is a good idea but the end result is we'll still be attempting extraction 24 times a day for a perpetually broken git repo when we may not look at the errors for days/weeks.  Maybe the delay should be longer?  12 hours rather than 1?

willdurand: I think we'll see how it goes. We have statsd metrics for that.

I am not sure we would end up in a situation where an add-on cannot be re-extracted at all for 24h. If we loop as you describe, it means new versions are also uploaded during that time.

Otherwise, if we have an extraction problem on a new repo, we won't re-extract the repo anymore (until a new version gets submitted).

PR #14186, opened 2020-05-06
Stop git-extraction of version sources

### LINKED ISSUE #14174: Stop git-extraction of version sources
We don't use the source files currently so we will remove the code to git-extract source files as well as the existing data, which means resetting the `Version.source_git_hash` values and deleting the "source" repositories on disk.

willdurand:
Fixes #14174

---

This patch removes the task that is used to git-extract the source files of a version and update the `AddonGitRepository.delete()` method to be able to delete `source` repos. This allows us to implement a new `delete_source_git_repositories` task for `process_addons`.

### DIFF
--- a/src/olympia/addons/management/commands/process_addons.py
+++ b/src/olympia/addons/management/commands/process_addons.py
@@ -19,9 +19,11 @@ from olympia.abuse.models import AbuseReport
 from olympia.constants.base import _ADDON_PERSONA, _ADDON_THEME, _ADDON_WEBAPP
 from olympia.amo.utils import chunked
 from olympia.devhub.tasks import get_preview_sizes, recreate_previews
+from olympia.git.tasks import delete_source_git_repositories
 from olympia.lib.crypto.tasks import sign_addons
 from olympia.reviewers.tasks import recalculate_post_review_weight
 from olympia.versions.compare import version_int
+from olympia.versions.models import Version
 
 
 firefox_56_star = version_int('56.*')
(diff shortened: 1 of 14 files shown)


src/olympia/git/tests/test_tasks.py
```diff
@@ -162,3 +163,21 @@ def test_calls_extract_version_to_git_n_times(
                 mock.call(version_id=3, force_extraction=True),
             ]
         )
+
+
+class TestDeleteSourceGitRepositories(TestCase):
+    def test_deletes_source_repositories(self):
+        deleted_addon = addon_factory()
```
bobsilverberg, line 170: I assume we're doing this to test the `with_deleted` kwarg? If so, I'm not sure how just using `1, 2, 3` as addon_ids is working without creating those addons first as well. Are they assumed to exist already, or does it not need there to be actual addons for those ids? I'm just a bit confused about how this test works. Also, does `addon_factory()` yield a deleted addon, and if not, why are we calling this `deleted_addon`?

src/olympia/addons/management/commands/process_addons.py
```diff
@@ -125,6 +127,21 @@ def get_recalc_needed_filters():
...
     },
+    'delete_source_git_repositories': {
+        'method': delete_source_git_repositories,
+        'distinct': True,
+        'qs': [
+            Q(
+                # Retrieve a list of add-on IDs that have at least one version
+                # with a source git hash (which means source files have been
+                # git-extracted).
+                pk__in=Version.unfiltered.exclude(source_git_hash__exact='')
```
willdurand, line 138: @diox this feels weird but since the queryset for `process_addons` "starts from" the `Addon` model, I don't know how to write the query differently.

src/olympia/git/tests/test_tasks.py
willdurand: mm, no you're right. This is a mistake.

The task does not really care about the add-ons being created in the database because the add-on IDs are enough to create/delete the git repositories on disk. I think `deleted_addon` should be deleted here.

src/olympia/addons/management/commands/process_addons.py
diox: It's a weird one but given what it's for I think that's fine, we shouldn't worry about spending too much time figuring out a better way

PR #14189, opened 2020-05-06
Add creation date and id to django admin pages

### LINKED ISSUE #14187: Add creation date to some django admin pages

Some django admin pages miss the creation date (and sometimes the id). They can be useful for admins, so let's add them.

I'm adding them for _UserProfile_, _Addon_, _Version_ and _File_.

### What happened?

Creation date (and sometimes even id) missing on the django admin pages for the models mentioned above.

Creation date and id should be shown.

The following django admin pages are affected:
* /admin/models/files/file/1/change/
* /admin/models/versions/version/1/change/
* /admin/models/addons/addon/1/change/
* /admin/models/users/userprofile/1/change/

(Please replace `1` by the ID of an actual instance that exists on the environment you are testing with.)

wagnerand:
Fix #14187.

### DIFF
--- a/src/olympia/versions/admin.py
+++ b/src/olympia/versions/admin.py
@@ -17,14 +17,15 @@ class VersionAdmin(admin.ModelAdmin):
         js = ('js/admin/l10n.js',)
 
     view_on_site = False
-    readonly_fields = ('version', 'channel', 'recommendation_approved')
+    readonly_fields = ('id', 'created', 'version', 'created', 'channel',
+                       'recommendation_approved')
 
     raw_id_fields = ('addon', 'license')
 
     fieldsets = (
         (None, {
             'fields': (
-                'addon', 'version', 'channel', 'release_notes',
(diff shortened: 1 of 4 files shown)


src/olympia/versions/admin.py
```diff
@@ -17,14 +17,15 @@ class Media:
         js = ('js/admin/l10n.js',)
 
     view_on_site = False
-    readonly_fields = ('version', 'channel', 'recommendation_approved')
+    readonly_fields = ('id', 'created', 'version', 'created', 'channel',
```
willdurand, line 20: Do you need `created` twice?

wagnerand: Oops, no. Copy & paste mistake.

wagnerand: Fixed in https://github.com/mozilla/addons-server/pull/14189/commits/b46b302234dfd06ddade978c0867c4b937e08f5a.

PR #14193, opened 2020-05-06
Return the version.id property for the file_only APIs

### LINKED ISSUE #14192: Return version.id with file_only data for Browse and Compare APIs
The new `file_only` version of the Browse and Compare APIs return just a `file` object, but they also need to return the `id` of the current version as well.

For QA:

After this patch, the responses to the `file_only` API requests should include both a `file` property and an `id` property, the latter of which is the `id` of the current version.

bobsilverberg:
Fixes #14192

### DIFF
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -62,10 +62,11 @@ from olympia.reviewers.models import (
     clear_reviewing_cache, get_flags, get_reviewing_cache,
     get_reviewing_cache_key, set_reviewing_cache)
 from olympia.reviewers.serializers import (
-    AddonBrowseVersionSerializer, AddonCompareVersionSerializer,
+    AddonBrowseVersionSerializer, AddonBrowseVersionSerializerFileOnly,
+    AddonCompareVersionSerializer, AddonCompareVersionSerializerFileOnly,
     AddonReviewerFlagsSerializer, CannedResponseSerializer,
     DiffableVersionSerializer, DraftCommentSerializer, FileEntriesSerializer,
-    FileEntriesDiffSerializer)
+)
 from olympia.reviewers.utils import (
     AutoApprovedTable, ContentReviewTable, ExpiredInfoRequestsTable,
     NeedsHumanReviewTable, ReviewHelper, ViewUnlistedAllListTable,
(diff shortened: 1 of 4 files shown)


src/olympia/reviewers/views.py
```diff
@@ -1374,7 +1374,7 @@ def retrieve(self, request, *args, **kwargs):
                     'request': self.request
                 }
             )
-            return Response({'file': serializer.data})
+            return Response({'id': version.id, 'file': serializer.data})
```
eviljeff, line 1377: To use DRF patterns shouldn't be building up responses like this - this is what a Serializer is for.  `AddonBrowseVersionSerializer` should be subclassed (or a parent class introduced); `get_serializer()` overridden on `ReviewAddonVersionViewSet`, and `get_serializer_context` overridden on `ReviewAddonVersionViewSet` to provide the context.  - Then we shouldn't need to redefine `retrieve` because that's what the `RetrieveModelMixin` is supposed to be doing.

src/olympia/reviewers/views.py
```diff
@@ -1511,20 +1511,21 @@ def get_objects(self):
...
                     'file': self.request.GET.get('file', None),
                     'request': self.request,
                     'parent_version': objs['parent_version'],
                 })
-            return Response({'file': serializer.data})
+            return Response({'id': version.id, 'file': serializer.data})
 
         serializer = AddonCompareVersionSerializer(
-            instance=objs['instance'],
+            instance=version,
```
eviljeff, line 1528: same comments on this class.  I know it's more work but these small quick hacks are turning into something more complicated.

src/olympia/reviewers/views.py
bobsilverberg: Thanks @eviljeff, that makes sense. I'll have to take a closer look at the `retrieve` thing, though, because this view already redefined `retrieve` so there must have been a need for that.

eviljeff: The `RetrieveModelMixin` literally just defines the `retrieve` method [1] so I'm not sure why it was used if was also defined tbh.

[1] https://github.com/encode/django-rest-framework/blob/3.11.0/rest_framework/mixins.py#L49

src/olympia/reviewers/views.py
eviljeff: actually, I think this one _does_ need a custom retrieve model because it uses `get_objects` rather than `get_object`.  For maintainability it would be best to still have a `get_serializer` and `get_serializer_context` thought.

bobsilverberg:
Thanks for the review and comments @eviljeff. I think I've made enough changes that this warrants another look, of you don't mind, please and thank you.

bobsilverberg:
The failure appears to be the intermittent filed at https://github.com/mozilla/addons-server/issues/14146.

Issue #14195, opened 2020-05-06
Test error: No module named 'wcwidth'


wagnerand:
This seems to be raised only if there are test failures.

```
Traceback (most recent call last):
  File "/home/travis/build/mozilla/addons-server/.tox/reviewers-and-zadmin/bin/pytest", line 8, in <module>
    sys.exit(main())
  File "/home/travis/build/mozilla/addons-server/.tox/reviewers-and-zadmin/lib/python3.6/site-packages/_pytest/config/__init__.py", line 125, in main
    config=config
  File "/home/travis/build/mozilla/addons-server/.tox/reviewers-and-zadmin/lib/python3.6/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/home/travis/build/mozilla/addons-server/.tox/reviewers-and-zadmin/lib/python3.6/site-packages/pluggy/manager.py", line 93, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/home/travis/build/mozilla/addons-server/.tox/reviewers-and-zadmin/lib/python3.6/site-packages/pluggy/manager.py", line 87, in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
  File "/home/travis/build/mozilla/addons-server/.tox/reviewers-and-zadmin/lib/python3.6/site-packages/pluggy/callers.py", line 208, in _multicall
    return outcome.get_result()
  File "/home/travis/build/mozilla/addons-server/.tox/reviewers-and-zadmin/lib/python3.6/site-packages/pluggy/callers.py", line 80, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/home/travis/build/mozilla/addons-server/.tox/reviewers-and-zadmin/lib/python3.6/site-packages/pluggy/callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "/home/travis/build/mozilla/addons-server/.tox/reviewers-and-zadmin/lib/python3.6/site-packages/_pytest/main.py", line 240, in pytest_cmdline_main
    return wrap_session(config, _main)
  File "/home/travis/build/mozilla/addons-server/.tox/reviewers-and-zadmin/lib/python3.6/site-packages/_pytest/main.py", line 229, in wrap_session
    session=session, exitstatus=session.exitstatus
[...]

PR #14198, opened 2020-05-06
Add wcwidth as test dep

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14195, hopefully


wagnerand:
What would be a good way to test this? Do you want to introduce a test-failure in a temporary commit to see if the error is still raised?

willdurand:
> What would be a good way to test this? Do you want to introduce a test-failure in a temporary commit to see if the error is still raised?

I was hoping to get a failure.. I guess, yes, introducing a failure is the right thing to do now that everything works as intended...

diox:
Note that it's an [explicit requirement](https://github.com/pytest-dev/pytest/blob/master/setup.py) from pytest so we should definitely include it.

willdurand:
> Note that it's an [explicit requirement](https://github.com/pytest-dev/pytest/blob/master/setup.py) [](/pytest-dev/pytest/blob/HEAD@{2020-05-06T22:47:35Z}/setup.py)

Then I guess, no need to test it further.

Issue #14085, opened 2020-04-28
Create a new (experimental) queue for MAD in the reviewer tools


willdurand:
Once we fix #14084, we'll have a list of versions that reviewers should examine to improve MAD. We should have a new queue for those versions.

willdurand:
@wagnerand what would be a good name for that queue? Should it be part of the "Flagged By Scanners" section?

willdurand:
> @wagnerand what would be a good name for that queue? Should it be part of the "Flagged By Scanners" section?

we'll go with "flagged for human review" and we'll rename the section to "security scanners".

AlexandraMoga:
@willdurand I had a look at the new MAD queue in reviewer tools and so far I've noticed the following:
- scanner results with a score of exactly `0%` or `100%` will appear in the queue
- scanners with different scores can appear or not in the queue; for example, a [result](https://addons-internal.dev.mozaws.net/en-US/admin/models/scanners/scannerresult/10187/change/?_changelist_filters=has_matched_rules%3Dall%26scanner__exact%3D4%26state%3Dall) with a score of `1%` was not in the queue, but [another](https://addons-internal.dev.mozaws.net/en-US/admin/models/scanners/scannerresult/10195/change/?_changelist_filters=has_matched_rules%3Dall%26scanner__exact%3D4%26state%3Dall) one with a score of `65%` was present in the queue (note that I ran these tests while the other scanner flags were disabled/triggered no action)
- in #14084 you also mention "a model disagreement" as a possible condition for scanner results to be present in the queue. What does this mean exactly?

There are a few issues with unlisted add-ons (one of them being that they link to the listed instead of the unlisted review page) but for these I will file a follow up.

I've also seen some empty queue entries (see screenshot). I'm guessing these are some scanner results without a version, but I don't think we should see them in the queue in any case.

![image](https://user-images.githubusercontent.com/31961530/81819005-6c3e8d00-9537-11ea-9f67-4c76390a6af9.png)

willdurand:
>     * in #14084 you also mention "a model disagreement" as a possible condition for scanner results to be present in the queue. What does this mean exactly?

You can forget about that, it's an implementation detail, sorry.

willdurand:
> I've also seen some empty queue entries (see screenshot). I'm guessing these are some scanner results without a version, but I don't think we should see them in the queue in any case.

Ah, that's interesting. I don't see how this is possible, though..

willdurand:
[...]

PR #14205, opened 2020-05-07
Add new queue for MAD in reviewer tools

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14085

---

I had to rename an existing queue because some of its methods/url names were conflicting with this new queue.


src/olympia/addons/models.py
```diff
@@ -351,6 +351,45 @@ def get_needs_human_review_queue(self, admin_reviewer=False):
                 addonreviewerflags__needs_admin_code_review=True)
         return qs
 
+    def get_mad_queue(self):
```
eviljeff, line 354: I've not actually attempted it, but this, `get_scanners_queue` and `get_auto_approved_queue` (and possibly `get_content_review_queue`) seem ripe for some refactoring because there's (seemingly) a lot of common code.  

The queryset methods can be built in any order so there could be a method that gets a base queryset for these type of functions with the common queryset modifiers and then the differences added in the individual `get_x_queue` methods.

src/olympia/addons/models.py
```diff
@@ -351,6 +351,45 @@ def get_needs_human_review_queue(self, admin_reviewer=False):
...
+            .prefetch_related(
+                '_current_version__files'
+            )
+            .order_by(
+                'created',
+            )
+            # There could be several versions matching for a single add-on so
+            # we need a distinct.
+            .distinct()
+        )
```
eviljeff, line 390: the other similar methods filter out addons that have `addonreviewerflags__needs_admin_code_review` (or content_review) - do we want to be including those to non-admin reviewers?  They can't do anything on the review page iirc, (can they even access the review page?).  Similarly with unlisted-only addons - there's a different permission needed for those reviews so can standard reviewers do anything?

If the review page can be accessed but they can't do anything that seems suboptimal, but if Product are okay with it, then sure.  Any inaccessible review page shouldn't be shown.

willdurand: Ha, so does that mean `addonreviewerflags__needs_admin_code_review` prevents regular reviewers to review add-on?

eviljeff: there's been some changes on that recently iirc, so... @diox or @wagnerand ?

diox: `needs_admin_code_review`  prevents regular reviewers from reviewing code, yes. `needs_admin_content_review` prevents content reviewers from reviewing... content.

*technically*, if `needs_admin_content_review` is `True` regular reviewers can't review either, because they are supposed to do a full review, and that includes content. The queue list doesn't care about that, but the review page does.

src/olympia/addons/models.py
willdurand: done

Issue #8508, opened 2018-06-13
Upgrade to ElasticSearch 6


EnTeQuAk:
* Add ElasticSearch 6 to Travis
* Figure out exact ElasticSearch version

EnTeQuAk:
Slightly blocked by https://github.com/travis-ci/apt-source-whitelist/issues/379 - not a hard block though, we can get around it.

EnTeQuAk:
@bqbn are there any ops requirements in regards to timing when this has to happen on our end?

bqbn:
We haven't started any work regarding ES6 upgrade yet. We can probably target this Q1/Q2 2019 from ops perspective.

We'd certainly want to do this after we are done with the UTF8mb4 upgrade (https://bugzilla.mozilla.org/show_bug.cgi?id=1479111), which is slated to happen early Q1.

diox:
Going to give this a try on travis to see where we are. I think we want do this soon, but after Python 3 and possibly after Django 2.2 too.

bqbn:
> I think we want do this soon, but after Python 3 and possibly after Django 2.2 too.

What are those going to happen the earliest?

And it looks like ops may not be able to start working on this in Q1 after all.

diox:
Python 3: We were thinking this week or the next. Sadly we missed the tag so it's probably going to be next week.
Django 2.2: When it's done. It's currently in beta, and the port seems to be going well on our side so we'll probably be ready when it's released sometime in April this year. Ideally we'd want to migrate as soon as it's released.

Maybe we can try to target early Q2 for Elasticsearch 6 ? Not sure how much work there is to do yet both on the dev and ops sides.

diox:
Good news: there aren't a lot of changes to make, so [my branch is almost done](https://github.com/diox/olympia/tree/elasticsearch-6.x), just a couple tests to adjust, but the code is ready.

Bad news: [Removal of mapping types](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/removal-of-types.html) hurt us. My branch has a refactor of stats code to deal with this, but this requires splitting the stats index into 2 separate indexes, each containing a single type (one with update counts, one with download counts). This means that we need to come up with a transition plan. A brutal approach would be to do a full reindex of stats on a new cluster with new code while serving the pages with the old cluster and old code (how long does that take?). An alternative would be to have ops figure out a way to move the mapping to the new indexes without requiring a full reindex, but I'm not sure what's possible here.

[...]

PR #14206, opened 2020-05-07
Upgrade to Elasticsearch 6.x

diox:
Because ES 6.x removed mapping types, move `update_count` and `download_count` in separate ES indexes. This requires a full reindex, including stats, but we need to create an entirely new cluster anyway.

ES 5.x compatibility is kept for now.

Fixes https://github.com/mozilla/addons-server/issues/8508


.travis.yml
```diff
@@ -70,7 +80,16 @@ before_script:
   - mysql --version
   - node --version
   - java -version
-  - curl -v http://localhost:9200/
+  - |
+    if [ $TOXENV == "es" ]; then
```
diox, line 84: This can all be removed, I was keeping it around in case tests fail.

diox:
`--wipe` fixed in https://github.com/mozilla/addons-server/pull/14206/commits/2bcc682e67d5daec88e276d7e791113729ca51e8 but we're waiting a couple more weeks before merging this to get the best time to deploy this with ops.

Issue #14086, opened 2020-04-28
Use the (new) blocklist to label scanner results as "bad"


willdurand:
The current approach to label "bad" scanner results is to rely on the `state` property. This is sub-optimal as we miss many blocked add-ons for other reasons (the primary use cases of the `state` property are: (1) triaging the scanner results and (2) assessing the soundness of the scanner rules)

AlexandraMoga:
@willdurand  I've verified this on -dev and I have the following results:

- an add-on is fully blocked (for all versions) => the scanner results for all its versions will be labelled  as `bad`
- an add-on is partially blocked (only for selected versions) => the scanner results pertaining to those versions will be labelled as `bad`
- an existing block is edited by increasing the blocked version ranges => the scanner results for the newly included versions will be labelled as `bad`
- a scanner result is marked as true positive then the add-on is blocked as well => there is only one `bad` scanner result for each scanner (i.e. there are no double entries for the same labelled result)

Potential issues:
- a scanner result label remains `bad` when a previously blocked version becomes unblocked
- a new version for a fully blocked add-on is marked as blocked automatically when submitted; however, the scanner result for this version doesn't show up in the api, i.e it is not labelled as `bad` (example: https://addons-dev.allizom.org/en-US/reviewers/review-unlisted/743bb61d623b425a8143). I guess this might be happening because there is no actual add/edit action on that last version?
- once a scanner result was labelled `good`, it won't change to `bad` if the add-on gets blocked afterwards

willdurand:
> a scanner result label remains `bad` when a previously blocked version becomes unblocked; the same happens when the block is deleted entirely ;

yep, that's a known issue and I don't think we can fix  that. We shouldn't delete blocks very often, though.



> a new version for a fully blocked add-on is marked as blocked automatically when submitted; however, the scanner result for this version doesn't show up in the api, i.e it is not labelled as `bad` (example: https://addons-dev.allizom.org/en-US/reviewers/review-unlisted/743bb61d623b425a8143). I guess this might be happening because there is no actual add/edit action on that last version?

Interesting, yeah I guess you're right. @eviljeff what could we do here?



> once a scanner result was labelled `good`, it won't change to `bad` if the add-on gets blocked afterwards

ha, really? I guess it should.

eviljeff:
[...]

PR #14208, opened 2020-05-07
Label scanner results as 'bad' when versions are blocked (according to the activity log)

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14086

---

/cc @diox @eviljeff

### DIFF
--- a/src/olympia/scanners/tests/test_views.py
+++ b/src/olympia/scanners/tests/test_views.py
@@ -9,6 +9,8 @@ from olympia.amo.tests import (
     version_factory,
 )
 from olympia.constants.scanners import YARA, CUSTOMS, WAT, TRUE_POSITIVE
+from olympia.blocklist.models import Block
+from olympia.blocklist.utils import block_activity_log_save
 from olympia.scanners.models import ScannerResult
 from olympia.scanners.serializers import ScannerResultSerializer
 from django.test.utils import override_settings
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/views.py
```diff
@@ -61,7 +61,13 @@ def get_queryset(self):
             .all()
         )
         bad_results = (
-            bad_results.filter(state=TRUE_POSITIVE)
+            bad_results.filter(
+                Q(state=TRUE_POSITIVE) | Q(
+                    version__versionlog__activity_log__action__in=(
+                        amo.LOG.BLOCKLIST_BLOCK_ADDED.id,
```
eviljeff, line 67: I think you can add `amo.LOG.BLOCKLIST_BLOCK_EDITED` too.

willdurand: added, thanks

willdurand:
I updated this patch, it should be ready now :)

src/olympia/scanners/tests/test_views.py
```diff
@@ -275,3 +277,30 @@ def test_get_by_label_and_scanner(self):
...
+        )
+        # result labelled as "bad" because the add-on is blocked.
+        blocked_addon_1 = addon_factory()
+        blocked_version_1 = version_factory(addon=blocked_addon_1)
+        ScannerResult.objects.create(scanner=YARA, version=blocked_version_1)
+        block_1 = Block.objects.create(
+            guid=blocked_addon_1.guid, updated_by=self.user
+        )
+        block_activity_log_save(block_1, change=False)
+        # result labelled as "bad" because the add-on is blocked.
```
diox, line 296: needs to say `and the block has been edited` or something.

src/olympia/scanners/tests/test_views.py
```diff
@@ -275,3 +277,30 @@ def test_get_by_label_and_scanner(self):
...
+        blocked_addon_2 = addon_factory()
+        blocked_version_2 = version_factory(addon=blocked_addon_2)
+        ScannerResult.objects.create(scanner=YARA, version=blocked_version_2)
+        block_2 = Block.objects.create(
+            guid=blocked_addon_2.guid, updated_by=self.user
+        )
+        block_activity_log_save(block_2, change=True)
```
diox, line 303: Maybe add 2 more cases:
- One with a block added *and* edited.
- One with a true positive and a block added

The idea being to guard against possible regression where we'd return the same add-on multiple times.

Issue #14046, opened 2020-04-23
Hide 'customs' in scanner query rule select field


AlexandraMoga:
Regressed by https://github.com/mozilla/addons-server/pull/13949

### Describe the problem and steps to reproduce it:
1. Open the scanner query rule creation form - https://addons-internal.stage.mozaws.net/en-US/admin/models/scanners/scannerqueryrule/add/
2. Click on the select field to see the available scanners

### What happened?
Both `yara` and `customs` are present

### What did you expect to happen?
Only `yara` should be present

### Anything else we should know?
This was fixed once in https://github.com/mozilla/addons-server/pull/13430

![image](https://user-images.githubusercontent.com/31961530/80095773-27938780-8571-11ea-82ea-3bcaffa46600.png)

diox:
Regression from https://github.com/mozilla/addons-server/issues/13851

AlexandraMoga:
Verified fixed on -dev:

For scanner query rules, `yara` is the only available select option.

![image](https://user-images.githubusercontent.com/31961530/81687827-ea306480-9461-11ea-84cb-52bbacc4b006.png)

For normal scanner rules, both `yara` or `customs` can be selected.

PR #14210, opened 2020-05-07
Fix list of scanners in query rule

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14046

---

This works because of this: https://github.com/mozilla/addons-server/pull/13430/files#diff-0ea3bc006b4f95bb1e4378265faf5829R300-R302


willdurand:
I fixed the codestyle issue.

Issue #14132, opened 2020-04-30
Refactor `get_mime_type_for_blob()`


willdurand:
The `get_mime_type_for_blob` function does content sniffing and file extension-based look-up to assign a `mimetype` to a given file: https://github.com/mozilla/addons-server/blob/1d76cfa07d7f03ce06dbe2465e06acdf9bc70d48/src/olympia/lib/git.py#L108-L158

It currently has 3 ways of finding a `mimetype`:

1. a small map named `SIMPLIFIED_MIMETYPE_DETECTION`
2. content sniffing with the `magic` lib (libmagic1)
3. it tries to find a more specific mimetype using the `mimetypes` lib (which uses some system files like `/etc/mime.types`)

That seems overly complicated and I am not quite sure to understand why we have 3 different techniques (although (2) and (3) combined makes perfectly sense).

I suggest we simplify this function by:

1. removing `SIMPLIFIED_MIMETYPE_DETECTION`
2. removing the code that uses `magic`, including the code to install `libmagic1` in the docker and travis-ci envs
3. relying on `mimetypes.guess_type()` only, which uses a system file as source. This would mean putting a `mime.types` file under version control (which would be copied during the docker build and on travis-ci as well).

Note: we updated the `/etc/mime.types` file on travis-ci to be similar to the docker image one in https://github.com/mozilla/addons-server/issues/14072

willdurand:
@diox any concern? The `mimetype` value is only used by code-manager to configure syntax highlighting and we do have this `SIMPLIFIED_MIMETYPE_DETECTION` map so I doubt we want to be hyper-accurate.

diox:
That sounds fine. I also don't think we need super accurate mime type handling for code-manager, something fast based on filename sounds good.

ioanarusiczki:
@willdurand I verified on AMO dev with FF76(Win10).

Compared with the previous behavior the policy of disabling the syntax highlighting did not change because I've been testing with the same files on AMO stage and got the same results.

**1. All Browse pages have the highlighting enabled.**

**2. Compare pages have both:**
 - syntax is enabled for files with code lines (without minified content or long blocks of code on a single line).

Examples:
https://code.addons-dev.allizom.org/en-US/compare/618525/versions/1689955...1689956/?allowSlowPages=false&path=src%2Foptions%2Fstyle.bundle.css 25 lines of code
[...]

PR #14211, opened 2020-05-07
Refactor get_mime_type_for_blob()

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14132


docker/etc/mime.types
```diff
@@ -0,0 +1,836 @@
...
+#
+#  MIME media types and the extensions that represent them.
+#
+#  The format of this file is a media type on the left and zero or more
+#  filename extensions on the right.  Programs using this file will map
+#  files ending with those extensions to the associated type.
+#
+#  This file is part of the "mime-support" package.  Please report a bug using
+#  the "reportbug" command of the "reportbug" package if you would like new
+#  types or extensions to be added.
```
eviljeff, line 11: where is this file from?  From the comments it looks like it's been copied&pasted from another project?

willdurand: it comes from our docker image (default file in debian)

willdurand: I added to our repo so that we can reuse them in travis-ci

eviljeff: _that's_ the part of the explanation I was missing 😉 

It's a pity we can't pull it in at travis build time so we don't have to keep it up to date

willdurand: Yep.. On the other hand, having it in our repo means a docker update won't break our detection (if the file gets updated in the debian image)

PR #14218, opened 2020-05-08
Move the `entries` from the `file` object into the version object for Browse and Compare APIs

### LINKED ISSUE #14212: Move the `entries` from the `file` object into the version object for Browse and Compare APIs
The `entries`, which is the tree of all files that are included in an uploaded add-on, are currently returned in the `file` object in the API responses for Browse and Compare. When the APIs are called with the `file_only` param, these `entries` are empty. This design isn't ideal, because these `entries` really apply to the `version` and not the `file`, because the `file` now represents the requested file, not the xpi file. This also makes things more difficult for types in code manager.

To resolve this, and have the data where it really belongs, we should move these `entries` from the `file` object up to the root of the response, which is where the other `version` properties are. While we're at it, it probably makes sense to rename the property from `entries`. I was thinking `file_entries`, `filetree` or maybe even just `files`. I think the latter makes sense, but it might be a bit confusing to have both a `file` property and a `files` property. If anyone has a suggestion for what to call this property once it has been moved from `file` to `version`, please let me know.

bobsilverberg:
Fixes #14212

### DIFF
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -47,41 +47,31 @@ class AddonReviewerFlagsSerializer(serializers.ModelSerializer):
         )
 
 
-# NOTE: Because of caching, this serializer cannot be reused and must be
-# created for each file. It cannot be used with DRF's many=True option.
-class FileEntriesSerializer(serializers.ModelSerializer):
-    content = serializers.SerializerMethodField()
-    uses_unknown_minified_code = serializers.SerializerMethodField()
-    download_url = serializers.SerializerMethodField()
-    entries = serializers.SerializerMethodField()
-    selected_file = serializers.SerializerMethodField()
-    mimetype = serializers.SerializerMethodField()
-    sha256 = serializers.SerializerMethodField()
-    size = serializers.SerializerMethodField()
-    mime_category = serializers.SerializerMethodField()
-    filename = serializers.SerializerMethodField()
+class FileEntriesMixin(object):
+    def _get_version(self):
+        """If neither the current instance nor the parent instance is a Version,
+        check the context for a version, otherwise raise an exception."""
+        if isinstance(self.instance, Version):
+            return self.instance
 
-    class Meta:
-        fields = ('id', 'content', 'entries', 'selected_file', 'download_url',
-                  'uses_unknown_minified_code', 'mimetype', 'sha256', 'size',
-                  'mime_category', 'filename'
-                  )
-        model = File
+        if self.parent is not None and isinstance(
+                self.parent.instance, Version):
+            return self.parent.instance
+
+        version = self.context.get('version', None)
+        if isinstance(version, Version):
+            return version
+
+        raise RuntimeError(
+            'This serialzer should not be created without a Version')
 
     @cached_property
     def repo(self):
-        return AddonGitRepository(self.get_instance().version.addon)
+        return AddonGitRepository(self._get_version().addon)
 
     @property
     def git_repo(self):
         return self.repo.git_repository
-
-    def get_instance(self):
-        """Fetch the correct instance either from this serializer or
-        it's parent"""
-        if self.parent is not None:
-            return self.parent.instance.current_file
-        return self.instance
 
     @cached_property
     def commit(self):
@@ -89,7 +79,7 @@ class FileEntriesSerializer(serializers.ModelSerializer):
         # Caching the commit to avoid calling revparse_single many times.
         try:
             return self.git_repo.revparse_single(
-                self.get_instance().version.git_hash)
+                self._get_version().git_hash)
         except pygit2.InvalidSpecError:
             raise NotFound(
                 'Couldn\'t find the requested version in git-repository')
@@ -99,13 +89,33 @@ class FileEntriesSerializer(serializers.ModelSerializer):
         # Caching the tree to avoid calling get_root_tree many times.
         return self.repo.get_root_tree(self.commit)
 
-    def _get_blob_for_selected_file(self, obj):
+    def _get_selected_file(self):
+        requested_file = self.context.get('file', None)
+        files = self._get_entries()
+
+        if requested_file is None:
+            default_files = ('manifest.json', 'install.rdf', 'package.json')
+
+            for manifest in default_files:
+                if manifest in files:
+                    requested_file = manifest
+                    break
+            else:
+                # This could be a search engine
+                requested_file = list(files.keys())[0]
+
+        if requested_file not in files:
+            raise NotFound('File not found')
+
+        return requested_file
+
+    def _get_blob_for_selected_file(self):
         """Returns the blob and filename for the selected file.
 
         Returns (None, None) if the selected file is not a blob.
         """
         tree = self.tree
-        selected_file = self.get_selected_file(obj)
+        selected_file = self._get_selected_file()
         if selected_file in tree:
             blob_or_tree = tree[selected_file]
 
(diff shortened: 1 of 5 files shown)


bobsilverberg:
I am having a bit of trouble understanding what I need to do to address the failures for `reviewers-versions-draft-comment-list`. It looks like the serializer is receiving a `list` object, but my code expects just a `DraftComment` object. I'm just having bit of difficulty understanding how this all works for the `list` view, and also why this is causing a problem now, when it wasn't before. I'm going to mark this for review, and maybe whoever gets it can help me understand what's going on here.

bobsilverberg:
@diox This is ready for a look. As we discussed on Matrix, I'm not sure if what I've done with `_get_version()` in the serializer really makes sense, but it all seems to work for every existing case (i.e., the tests all pass). I do think that we should probably stop the `DraftCommentSerializer` from returning all the `version` info, which might remove the need for the `get_version_object` hack. I just need to confirm that Code Manager doesn't actually need that data. 

I also included a quick drive-by fix for https://github.com/mozilla/addons-server/issues/14238 in here because it was so small and I was already editing the code in that area. I realize that adds to the burden of reviewing this, so I can strip those changes out if you'd prefer.

bobsilverberg:
Hmm, for some reason Travis isn't doing a build for the latest commit.

src/olympia/reviewers/views.py
```diff
@@ -1320,7 +1320,9 @@ def get_addon_object(self):
         return self.addon_object
 
     def get_version_object(self):
```
diox, line 1322: This isn't technically new, but since you made changes to it... I can't find any test hitting this method. When is it supposed to be called, do you know ?

src/olympia/reviewers/serializers.py
```diff
@@ -47,49 +47,34 @@ class Meta:
...
-        fields = ('id', 'content', 'entries', 'selected_file', 'download_url',
-                  'uses_unknown_minified_code', 'mimetype', 'sha256', 'size',
-                  'mime_category', 'filename'
-                  )
-        model = File
+        if self.parent is not None and isinstance(
+                self.parent.instance, Version):
+            return self.parent.instance
+
+        return self.context['view'].get_version_object()
```
diox, line 61: When is this ever called in actual usage ? As I mentioned in my other comment The `get_version_object()` implementation in `ReviewAddonVersionMixin` never seems to be called by tests, but this is also true of this line - some serializer tests hit it, but not views tests, which is quite suspicious.

src/olympia/reviewers/views.py
bobsilverberg: It is supposed to get called when a serializer whose instance is not a version, needs to use the mixin. I found this use case via tests, so maybe those tests are faulty if this never needs to be called in real world scenarios. I did have some concerns about what the DraftCommentSerializer is doing.

src/olympia/reviewers/serializers.py
bobsilverberg: I agree it seems suspicious. I'll have to look more carefully at the cases where it is called in serializer tests.

How are you able to tell when something is or isn't called via tests? Is there a coverage tool one can use when running tests to see which code is and is not covered?

diox: You can generate coverage with pytest (https://coverage.readthedocs.io/en/coverage-5.1/#quick-start - we disabled that a while back because it wasn't generating useful data for us, between how we split tests and how much legacy code there was with poor coverage) but in this case I simply added a breakpoint and ran the view tests, and it was never triggered.

bobsilverberg: Ah, that makes sense!

diox: FWIW I just tested `coverage run -m pytest src/olympia/reviewers/tests/test_views.py -k ViewSet` and `coverage report -m src/olympia/reviewers/views.py` and it found the same thing - line 61 of `serializers.py` was missed, and so was line 1323 of `views.py`. I would have expected both to be hit when running those tests.

bobsilverberg: Ok, I see the issue. This line is called in the tests when we are testing the `FileInfoSerializer` and the `FileInfoDiffSerializer` in isolation. In that case the serializer has an instance of a `File` object, not a `Version` object, and it doesn't have a `parent`. Hence it needs to call `self.context['view'].get_version_object()` to obtain a `Version` object.

In real life, those serializers are never called in isolation, but rather only as sub-serializers for the `file` property in their parent serializer. If we only need this code to support testing these in isolation, then we should remove it. I see two possibilities:
1. Write the tests differently so the serializer will have access to a `Version` object. I'm not sure this is actually practical, because these FileInfo... serializers are not designed to have access to a `Version` object, except via their parent. 
2. Remove the tests for the FileInfo... serializers in isolation and put those tests into the tests for the associated parent objects.

Number 2 sounds like the correct solution to me.

@diox WDYT?

diox: A possible alternative:
- Write the tests differently so that the serializer will have access to a `Version` via a parent. You'd need to set up the test in a certain way, but that should be doable.

Ultimately though, we should remove any dead code the views aren't actually using (Or, if it's not actually dead, write new views tests - normally I wouldn't necessarily test everything from the view perspective, but in this instance it's a fairly specific piece of view-serializer interaction so I think it would deserve it)

bobsilverberg: This is all fixed now.

bobsilverberg:
Some changes are required to address the tests that are failing, and I'm going to change the way the tests for the `FileInfo...` serializers are working. Please stand by...

bobsilverberg:
Ok, this is ready for another look. Sorry for the noise.

bobsilverberg:
This will break code manager, so I've labelled it "do not merge". We'll have to coordinate it with a code manager patch, for which I will open an issue.

src/olympia/reviewers/tests/test_serializers.py
```diff
@@ -33,18 +34,39 @@ def setUp(self):
...
+                'notify-link-clicks-i18n.xpi')
+        self.version = self.addon.current_version
+        self.file = self.addon.current_version.current_file
+
+    def get_serializer(self, **extra_context):
         api_version = api_settings.DEFAULT_VERSION
         request = APIRequestFactory().get('/api/%s/' % api_version)
         request.versioning_scheme = api_settings.DEFAULT_VERSIONING_CLASS()
         request.version = api_version
         extra_context.setdefault('request', request)
```
bobsilverberg, line 49: I have a question about this code, the code that creates the request, configures it, and passes it into `extra_context`. It doesn't seem to be needed by the code (which doesn't access anything in `request`), and if I remove it from the test cases they all still pass. Is this something that is needed for each of these tests, and if so, why?

diox: Looks like it was needed when `download_url` was added in 1a7217dcfad9514ce23ed00a0aaba94e82dd9272 because back then the download view was in DRF and we used `reverse()` from DRF which needs a `request`, but it was later changed in e67eeccaf848b0fc2eaa4304bd2b041bf84a9551 to call the "regular" download view which is not an API, so we removed the need for DRF's `reverse', which in turn meant we no longer need to rely on `request` from the serializer context.

DRF still provides `request` when it initializes a serializer from a view, but we can forget about all this and remove that code if we no longer depend on it.

bobsilverberg: Ok, so we can remove it from all the tests in this file?

diox: If tests pass, yes

PR #14229, opened 2020-05-11
disable v2 block editing via AMO admin

### LINKED ISSUE #14141: Prevent imported v2 blocks from being edited via AMO admin
#14094 will automate the import of the blocks from the v2 blocklist collection on remote-settings, but we also need to stop those blocks from being edited via AMO (or the changes could be overwritten during the next import)

eviljeff:
fixes #14141

### DIFF
--- a/src/olympia/blocklist/tests/test_forms.py
+++ b/src/olympia/blocklist/tests/test_forms.py
@@ -2,13 +2,39 @@ from django.contrib import admin as admin_site
 from django.core.exceptions import ValidationError
 from django.test import RequestFactory
 
+from waffle.testutils import override_switch
+
 from olympia.amo.tests import addon_factory, TestCase, user_factory
 from olympia.blocklist.admin import BlocklistSubmissionAdmin
-from olympia.blocklist.forms import MultiDeleteForm
+from olympia.blocklist.forms import MultiAddForm, MultiDeleteForm
 from olympia.blocklist.models import Block, BlocklistSubmission
 
 
-class TestBlocklistSubmissionForm(TestCase):
+class LegacySyncCheckMixin():
+    def test_v2_blocks_cant_be_edited(self):
+        data = {
+            'guids': 'any@thing\nsecond@thing',
+        }
+        anyblock = Block.objects.create(
+            guid='any@thing', updated_by=user_factory())
+        Block.objects.create(guid='second@thing', updated_by=user_factory())
+
+        form = MultiDeleteForm(data=data)
+        form.is_valid()
+        form.clean()  # would raise
+
+        # Give any@thing a .kinto_id so its sync'd with legacy blocklist
+        anyblock.update(kinto_id='123456')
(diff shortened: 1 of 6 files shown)


src/olympia/blocklist/tests/test_models.py
```diff
@@ -42,6 +43,23 @@ def test_is_imported_from_kinto_regex(self):
         block.update(kinto_id='1234567890')
         assert not block.is_imported_from_kinto_regex
 
+    def is_readonly(self):
```
willdurand, line 46:
```suggestion
    def test_is_readonly(self):
```

src/olympia/blocklist/tests/test_models.py
```diff
@@ -42,6 +43,23 @@ def test_is_imported_from_kinto_regex(self):
...
+        block = Block.objects.create(guid='foo@baa', updated_by=user_factory())
+        # not read only by default
+        assert not block.is_readonly
+        # but should be if there's an active BlocklistSubmission
+        block.active_submissions = [object()]  # just needs to be non-empty
+        assert block.is_readonly
+
+        # otherwise kinto_id being non-false means it's imported, so readonly
+        del block.active_submissions
```
willdurand, line 55: It would probably read better with an assignment to an empty list:
```suggestion
        block.active_submissions = []
```

src/olympia/blocklist/tests/test_forms.py
```diff
@@ -140,3 +164,38 @@ def test_guids_must_exist_for_block_deletion(self):
...
+        data = {
+            'guids': 'any@thing\nsecond@thing',
+        }
+
+        form = MultiAddForm(data=data)
+        addon_factory(guid='any@thing')
+        Block.objects.create(guid='any@thing', updated_by=user_factory())
+        addon_factory(guid='second@thing')
+        form.is_valid()
+        form.clean()  # would raise
```
willdurand, line 192: shouldn't it be `wouldn't raise`?

eviljeff: as in, would raise if there was an error.

src/olympia/blocklist/tests/test_models.py
eviljeff: 😬

src/olympia/blocklist/tests/test_forms.py
willdurand: ah ok, then maybe update the comment with the longer version :D

src/olympia/blocklist/tests/test_models.py
eviljeff: deleting it resets the cached property so the query runs - I want to check the query works too.

willdurand: ah ok, then maybe add a comment? I tried with empty list and the test passed too.

PR #14234, opened 2020-05-11
Prevent some activity logging from being garbage collected

wagnerand:
I went through all activity types and made a few permanent, where I believe it makes sense to keep them.


wagnerand:
@bobsilverberg We have a garbage collector for activity logs at https://github.com/mozilla/addons-server/blob/master/src/olympia/amo/cron.py#L35-L44 . Adding `keep = True` prevents entries of that type from being gc'ed.

Issue #14230, opened 2020-05-11
Change git extraction CRON task to extract 10 versions at a time (no more chain)


willdurand:
The chain of tasks isn't reliable so let's only extract 10 versions in a row. If there are more than 10 versions to extract, we'll keep the entry in the git extraction queue (setting its `in_progress` prop to `False`). Otherwise we'll remove it from the queue (that's what we're doing already).

For QA: we don't expect a lot of changes here except that it might take more time to re-extract add-ons with lots of versions.

ioanarusiczki:
@willdurand I uploaded:
 12 versions for https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/noscriptcron1 
and 15 versions for https://reviewers.addons.allizom.org/en-US/reviewers/review/cookie-autodeletetest2334?page=1#history - it took approximately 7 min for all versions to be displayed in code manager 

All versions were extracted successfully. 
Verified on AMO stage with FF76(Win10).

PR #14235, opened 2020-05-11
Only extract BATCH_SIZE versions per add-on in the CRON task

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14230

---

The idea behind this is that celery does not handle workflows very well so chains with more than a few tasks become very fragile. In production, we created chains of more than 1K tasks and that did not work very well.

This approach guarantees what the chain used to guarantee (sequential order) without the chain. It should be more robust. We shouldn't have to extract more than 1 version per add-on at the same time, except when we need to fix this add-on (either re-creating the repo and everything _or_ extracting missing versions).

### DIFF
--- a/src/olympia/git/management/commands/git_extraction.py
+++ b/src/olympia/git/management/commands/git_extraction.py
@@ -7,10 +7,10 @@ from django.core.management.base import BaseCommand
 import olympia.core.logger
 
 from olympia.amo.decorators import use_primary_db
-from olympia.amo.utils import chunked
 from olympia.files.utils import lock
 from olympia.git.models import GitExtractionEntry
 from olympia.git.tasks import (
+    continue_git_extraction,
     extract_versions_to_git,
     on_extraction_error,
     remove_git_extraction_entry,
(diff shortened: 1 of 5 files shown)


src/olympia/git/management/commands/git_extraction.py
```diff
@@ -19,9 +19,13 @@
 
 log = olympia.core.logger.getLogger('z.git.git_extraction')
 
-BATCH_SIZE = 10  # Number of versions to extract in a single task.
-LOCK_NAME = 'git-extraction'  # Name of the lock() used.
-SWITCH_NAME = 'enable-git-extraction-cron'  # Name of the waffle switch.
+# Number of versions to extract in a single task. If you change this value,
+# please adjust the soft time limit of the `extract_versions_to_git` task.
+BATCH_SIZE = 10
```
eviljeff, line 24: adjust the time limit where?  (I don't see one defined - somewhere adjacent to `extract_versions_to_git` anyway)

willdurand: right, I think we're using the default. I'll add a note.

PR #14248, opened 2020-05-12
handle duplicate versions correctly during MLBF generation

### LINKED ISSUE #14237: handle duplicate version strings during bloomfilter creation
https://sentry.prod.mozaws.net/operations/olympia-prod/issues/8362254/

If an add-on has two versions with the same version string it breaks the queryset excludes() (because only one of the version ids is recorded while building up the list)

eviljeff:
fixes #14237 - also changes the order of `generate_and_write_mlbf` so the JSON files are written first, enabling us to debug with the original inputs in case of a future similar breakage.

### DIFF
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -41,11 +41,11 @@ class MLBF():
             'version__addon__guid',
             'version__version',
             'version_id')
-        addons_versions = defaultdict(dict)
+        addons_versions = defaultdict(list)
         for file_ in file_qs:
             addon_key = file_['version__addon__guid']
-            addons_versions[addon_key][file_['version__version']] = (
-                file_['version_id'])
+            addons_versions[addon_key].append(
+                (file_['version__version'], file_['version_id']))
 
         all_versions = {}
         # collect all the blocked versions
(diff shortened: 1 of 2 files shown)


src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -36,24 +36,24 @@ def setUp(self):
                 file_kw={'is_signed': True, 'is_webextension': True}),
             updated_by=user_factory())
         # five versions, but only two within block (123.40, 123.5)
-        self.five_ver = Block.objects.create(
+        self.five_ver_block = Block.objects.create(
```
eviljeff, line 39: most of the changes in this file is code churn because `five_ver` was ambiguously named vs. `five_ver_123_45`

src/olympia/blocklist/mlbf.py
```diff
@@ -87,10 +87,11 @@ def generate_mlbf(cls, stats, blocked, not_blocked):
         cascade.set_crlite_error_rates(
             include_len=error_rates[0], exclude_len=error_rates[1])
 
-        cascade.initialize(include=blocked, exclude=not_blocked)
-
         stats['mlbf_blocked_count'] = len(blocked)
         stats['mlbf_notblocked_count'] = len(not_blocked)
+
+        cascade.initialize(include=blocked, exclude=not_blocked)
```
willdurand, line 93: Does putting the line here change anything?

eviljeff: It just means the stats dict is populated earlier. When i had the traceback on sentry i wished i had that information

PR #14251, opened 2020-05-12
When extracting a non-existing file, try its alternate location

### LINKED ISSUE #14232: Handle moved/deleted files better during git extraction
We should do something like we did for code-search.

diox:
Also delete code to handle source files that was in the way, it's obsolete since we stopped extracting source in #14174

Fixes #14232

### DIFF
--- a/src/olympia/files/models.py
+++ b/src/olympia/files/models.py
@@ -275,6 +275,17 @@ class File(OnChangeMixin, ModelBase):
             return self.file_path
 
     @property
+    def fallback_file_path(self):
+        """Fallback path in case the file was disabled/re-enabled and not yet
+        moved - sort of the opposite to current_file_path. This should only be
+        used for things like code search or git extraction where we really want
+        the file contents no matter what."""
+        return (
+            self.file_path if self.current_file_path == self.guarded_file_path
+            else self.guarded_file_path
+        )
+
+    @property
     def extension(self):
         return os.path.splitext(self.filename)[-1]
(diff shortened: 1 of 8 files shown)


src/olympia/files/models.py
```diff
@@ -274,6 +274,17 @@ def current_file_path(self):
         else:
             return self.file_path
 
+    @property
+    def fallback_file_path(self):
```
eviljeff, line 278: couldn't we just always try `self.file_path` and then `self.guarded_file_path`.  I'm not sure this extra logic gives us much (apart from one possible attempt at loading a file from the file system which would then fail.

diox: I'd rather avoid the extra filesystem calls (that would fail because it's the wrong location) as much as possible. We've got a lot of disabled/deleted files so there is no path that's always best to try first...

src/olympia/lib/git.py
```diff
@@ -421,19 +389,25 @@ def find_or_create_branch(self, name):
...
-                    source=path,
-                    dest=worktree.extraction_target_path,
-                    force_fsync=True)
+                try:
+                    extract_extension_to_dest(
+                        source=file_obj.current_file_path,
+                        dest=worktree.extraction_target_path,
+                        force_fsync=True)
+                except FileNotFoundError:
+                    extract_extension_to_dest(
```
willdurand, line 407: Should we log a message?

Issue #14233, opened 2020-05-11
Use a setting to configure routes behind VPN


willdurand:
We would like to have some routes available behind VPN only. We need a new setting configured via an environment variable.

This will be used for the scanners API (we can remove the waffle switch).

AlexandraMoga:
The [scanners api](https://addons-server.readthedocs.io/en/latest/topics/api/scanners.html#scanners) can be accessed from an internal URL - i.e https://addons-internal.dev.mozaws.net/ - by authenticated users (amo admins) and while being connected to the VPN.
Without the VPN active, the endpoint will not respond. 

The scanners api waffle switch is no longer available.

PR #14254, opened 2020-05-12
Introduce new setting for internal routes

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14233

---

~~I tried to write tests for when we have the setting set to `False` but I was not able to find how to do that correctly.. The problem is that django loads the url config once at the beginning, so either we have tests with this setting set to `True` or `False` but not both, apparently. That might quickly become a problem in the near future, though. Ideas?~~


diox:
Yeah this is always a pain to test. Does it work if you use [`override_settings()` to reset the `ROOT_URLCONF` as described in django docs](https://docs.djangoproject.com/en/2.2/topics/testing/tools/#urlconf-configuration) in addition to changing the `INTERNAL_ROUTES_ALLOWED` setting any chance ?

willdurand:
> Yeah this is always a pain to test. Does it work if you use [`override_settings()` to reset the `ROOT_URLCONF` as described in django docs](https://docs.djangoproject.com/en/2.2/topics/testing/tools/#urlconf-configuration) in addition to changing the `INTERNAL_ROUTES_ALLOWED` setting any chance ?

No.

I played with `shell_plus` and I found a way to reload the urlconf manually.

willdurand:
@diox this should be ready-ish now :)

willdurand:
I added a  new toxenv.

.travis.yml
```diff
@@ -16,6 +16,7 @@ jobs:
     - { env: TOXENV=reviewers-and-zadmin }
     - { env: TOXENV=amo-lib-locales-and-signing }
     - { env: TOXENV=main }
+    - { env: TOXENV=disable-internal-routes }
```
diox, line 19: The number of jobs we run in parallel was determined by trial and error, somewhat optimized to make the suite run the fastest possible while still keeping the envs definition sane and as logical as possible. Running an extra job just for these tests seem wasteful, especially with the amount of overhead caused by deps installation etc.

Also, the logic is kinda inverted compared to production: in production, the routes will be disabled most of the time, only enabled on internal instance, so maybe the tests should reflect that.

I realize that this makes writing the envs definition a bit wonky though... Not sure how much.

willdurand: > The number of jobs we run in parallel was determined by trial and error, somewhat optimized to make the suite run the fastest possible while still keeping the envs definition sane and as logical as possible. Running an extra job just for these tests seem wasteful, especially with the amount of overhead caused by deps installation etc.

How can we run a new toxenv otherwise?

> Also, the logic is kinda inverted compared to production: in production, the routes will be disabled most of the time, only enabled on internal instance, so maybe the tests should reflect that.

It is inverted because we'll write more test cases for the feature when it is enabled than disabled.

> I realize that this makes writing the envs definition a bit wonky though... Not sure how much.

So what do we do? Not having test coverage for the feature disabled seems risky as it means exposing internal features publicly.

willdurand: The last option I have in mind, which does not sound great but would allow us to move forward, is to do the check in the view itself (a bit like it was done with the waffle switch).

diox: I think doing it in the url patterns is good, and so is the toxenv approach for testing. But I'd like to find a way to write the toxenvs that do not cause tests to take even more time than usual.

Ideally, I think the setting should be disabled by default and only enabled for specific tests, or maybe all tests dealing with admin stuff (since the admin is supposed to run on the same internal instances where we'll have the setting enabled).

@eviljeff you played with tests grouping in the past as well, any thoughts about how we should do this ?

eviljeff: There's not a good answer here - ideally we'd isolate the relevant tests but there are only a handful of scanner result tests.  Collecting admin tests are a good call but there are few left in the zadmin/ app though so we'd need to collect the `test_admin.py` tests from all of the other apps and just run those in the admin job.  (`--ignore-glob=*/test_admin.py` in the others and `src/*/test_admin.py`).  

To get the number of jobs down again we could combine the `es` job with one of the others (run the pytest command twice in that job).  Or maybe once the test_admin.py tests are combined we'll see some reductions in the other test jobs.  Setup for a single job takes between 4 and 5 minutes but Travis only runs 5 concurrent jobs on forks - they're ordered so the earlier ones finish quickly.  The sweet spot is around 8-9 minutes a test.  Takes a lot of trial and error and it's not a given our current setup is optimum. 

I agree with @diox that the default in the tests should be the setting is disabled.  (And the double negative in `not internal_routes_not_allowed` is confusing, making it seem like we're enabling internal routes in all of the other tests)

diox: I'd keep the `es` job separate, it's fairly specific - in the future, it'll be the only job where elasticsearch is enabled.

willdurand: > making it seem like we're enabling internal routes in all of the other tests

All but one test case will require the setting to be set to `True` so it makes more sense to be able to run the test files without any option to run all the tests, and let Travis handle the case where the setting is `False` (which should be only 1 test case per protected route(s)).

Anyway, I am not sure to understand what to do next. We still need a TOXENV for the setting set to "not the default value" (whatever it is, and we can rename the pytest mark if neeeded).

diox: Technically, you don't need another env - you just need to run `pytest` with the variable / marker. That needs to be separate, but it doesn't have to be a new toxenv.

I still think the setting should be set to `False` by default because it matches the fact that it's going to be `False` for the "regular" instance. The idea is that the majority of our tests cover the "regular" instance, not the "internal" one. This would catch someone accidentally reversing the URL because they have set the setting to `True` locally, not realizing it wouldn't work then on dev/stage/prod.

willdurand: > Technically, you don't need another env - you just need to run `pytest` with the variable / marker. That needs to be separate, but it doesn't have to be a new toxenv.

mmm isn't the toxenv needed to configure the env var (which controls the setting)?

> I still think the setting should be set to False by default because it matches the fact that it's going to be False for the "regular" instance. The idea is that the majority of our tests cover the "regular" instance, not the "internal" one. This would catch someone accidentally reversing the URL because they have set the setting to True locally, not realizing it wouldn't work then on dev/stage/prod.

So `settings.py` would have the setting set to `False` as well or is that just for the test suite?

diox: > mmm isn't the toxenv needed to configure the env var (which controls the setting)?

I think `ENV_VAR=1 pytest ...` should work, but do try it out.

> So `settings.py` would have the setting set to `False` as well or is that just for the test suite?

I guess in `settings.py` you would have it set to `True` so that it works locally (not entirely satisfying, but we'll have to make compromises anyway)

willdurand:
GitHub/Travis CI integration seems broken right now. Builds pass (except one "es" job in one build) :)

tox.ini
```diff
@@ -22,17 +22,22 @@ commands =
...
     pytest -n 2 -m 'not es_tests and not needs_locales_compilation and not static_assets' -v src/olympia/devhub/ {posargs}
 
 [testenv:reviewers-and-zadmin]
+whitelist_externals =
+    {[testenv]whitelist_externals}
+    sed
 commands =
     make install_python_test_dependencies
     pytest -n 2 -m 'not es_tests and not needs_locales_compilation and not static_assets' -v src/olympia/reviewers/ src/olympia/zadmin/ {posargs}
+    sed -i 's/^INTERNAL_ROUTES_ALLOWED.*/INTERNAL_ROUTES_ALLOWED=True/' settings_test.py
```
willdurand, line 39: This is needed to allow the subsequent `pytest` call to run with the right settings.

eviljeff: can't we set the env variable here and have it consumed?  Using `INTERNAL_ROUTES_ALLOWED = os.environ.get('INTERNAL_ROUTES_ALLOWED', False)` or something similar?

willdurand: nope, unfortunately

willdurand:
(rebased)

willdurand:
(rebased)

Issue #14194, opened 2020-05-06
Intermittent test failure: TestRecommendedQueue.test_results


wagnerand:
```
______________________ TestRecommendedQueue.test_results _______________________
[gw0] linux -- Python 3.6.7 /home/travis/build/mozilla/addons-server/.tox/reviewers-and-zadmin/bin/python3
src/olympia/reviewers/tests/test_views.py:1732: in test_results
    self._test_results()
        self       = <olympia.reviewers.tests.test_views.TestRecommendedQueue testMethod=test_results>
src/olympia/reviewers/tests/test_views.py:1174: in _test_results
    check_links(expected, links, verify=False)
        addon      = <Addon: 3840: Pending Two>
        channel    = []
        doc        = [<html>]
        expected   = [('Nominated One 0.1', '/en-US/reviewers/review/nominated-one'),
 ('Nominated Two 0.1', '/en-US/reviewers/review/nominated-two'),
 ('Pending One 0.1', '/en-US/reviewers/review/pending-one'),
 ('Pending Two 0.1', '/en-US/reviewers/review/pending-two')]
        idx        = 3
        latest_version = <Version: 0.1>
        links      = [<a>, <a>, <a>, <a>]
        name       = 'Pending Two 0.1'
        response   = <HttpResponse status_code=200, "text/html; charset=utf-8">
        self       = <olympia.reviewers.tests.test_views.TestRecommendedQueue testMethod=test_results>
        url        = '/en-US/reviewers/review/pending-two'
src/olympia/amo/tests/__init__.py:199: in check_links
    f'At index {idx}, expected {text}, got {e.text()}')
E   AssertionError: At index 0, expected Nominated One 0.1, got Pending Two 0.1
        e          = [<a>]
        elements   = [<a>, <a>, <a>, <a>]
        expected   = [('Nominated One 0.1', '/en-US/reviewers/review/nominated-one'),
 ('Nominated Two 0.1', '/en-US/reviewers/review/nominated-two'),
 ('Pending One 0.1', '/en-US/reviewers/review/pending-one'),
 ('Pending Two 0.1', '/en-US/reviewers/review/pending-two')]
        idx        = 0
        item       = ('Nominated One 0.1', '/en-US/reviewers/review/nominated-one')
        link       = '/en-US/reviewers/review/nominated-one'
        selected   = None
        text       = 'Nominated One 0.1'
        verify     = False
```

diox:
The whole class is a source of intermittent failures:
[...]

PR #14255, opened 2020-05-12
Fix intermittent failures in reviewer queues tests

diox:
Fixes https://github.com/mozilla/addons-server/issues/14146
Fixes https://github.com/mozilla/addons-server/issues/14194

### DIFF
--- a/src/olympia/amo/tests/__init__.py
+++ b/src/olympia/amo/tests/__init__.py
@@ -855,7 +855,7 @@ def version_factory(file_kw=None, **kw):
     addon_type = getattr(kw.get('addon'), 'type', None)
     min_app_version = kw.pop('min_app_version', '4.0.99')
     max_app_version = kw.pop('max_app_version', '5.0.99')
-    version_str = kw.pop('version', '%.1f' % random.uniform(0, 2))
+    version_str = kw.pop('version', '%.1f' % (random.uniform(0, 99) + 0.1))
     application = kw.pop('application', amo.FIREFOX.id)
     if not kw.get('license') and not kw.get('license_id'):
         # Is there a built-in one we can use?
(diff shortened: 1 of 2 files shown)


src/olympia/amo/tests/__init__.py
```diff
@@ -855,7 +855,7 @@ def version_factory(file_kw=None, **kw):
     addon_type = getattr(kw.get('addon'), 'type', None)
     min_app_version = kw.pop('min_app_version', '4.0.99')
     max_app_version = kw.pop('max_app_version', '5.0.99')
-    version_str = kw.pop('version', '%.1f' % random.uniform(0, 2))
+    version_str = kw.pop('version', '%.1f' % (random.uniform(0, 99) + 0.1))
```
diox, line 858: Technically not required, but wanted more different version numbers (and avoid `0.0`).

eviljeff: `random.uniform(0.1, 99)`?

src/olympia/reviewers/tests/test_views.py
```diff
@@ -1491,19 +1521,17 @@ def test_results_two_versions(self):
         version2 = self.addons['Nominated Two'].versions.all()[0]
         file_ = version2.files.get()
 
-        # Versions are ordered by creation date, so make sure they're set.
-        past = self.days_ago(1)
-        version2.update(created=past, nomination=past)
-
-        # Create another version, v0.2, by "cloning" v0.1.
+        # Create another version for Nominated Two, v0.2, by "cloning" v0.1.
```
bobsilverberg, line 1524: Nit: I find this comment a bit confusing because it says "Create another version", but we already have `version2`, which we've created above, and all we're doing is updating some of its properties. Would it be more accurate to say "Update version 2..."?

diox: by setting the `pk` to `None` and saving it again, we create a different version (cloning properties of the instance we started from)

src/olympia/amo/tests/__init__.py
diox: :rofl: not sure why I did it the way that I did... technically doesn't matter though

Issue #14269, opened 2020-05-13
Unlisted add-ons are linked to the listed review page in the flagged by 'mad' review queue


AlexandraMoga:
Follow up for #14085

### Describe the problem and steps to reproduce it:
1. Log in with and admin account or an Admin reviewer
2. Open the Flagged for human review queue - https://reviewers.addons-dev.allizom.org/en-US/reviewers/queue/mad
3. Identify an unlisted add-on present in the queue (any entry that has _None_ for the version string is an unlisted add-on)
4. Click on the add-on link

### What happened?
You are redirected to the listed review page, so there will be no versions to show in the add-on review history
### What did you expect to happen?
The link should point to the unlisted review page

### Anything else we should know?
- the fact that the add-on version string is _None_ instead of the actual version number is also an issue
- the 'Flagged by scanners' queue uses separate links for the listed and unlisted review pages, which could also handle cases in which an add-on has both listed and unlisted flagged versions

willdurand:
ha oops, we should render the list just like the other existing scanner queue I guess.

willdurand:
@AlexandraMoga I just fixed this issue and we'll push it today

AlexandraMoga:
@willdurand looks good

![image](https://user-images.githubusercontent.com/31961530/81948469-73829b00-960a-11ea-9d14-a86de2787a71.png)

Both links are opened on the correct review page.

AlexandraMoga:
Stage is also fixed.

PR #14273, opened 2020-05-13
Fix mad queue

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14269

also fixes https://github.com/mozilla/addons-server/issues/14085#issuecomment-628108792

---

Not sure how I missed that given that the security queues are similar...

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -241,37 +241,46 @@ class ContentReviewTable(AutoApprovedTable):
 
 
 class ScannersReviewTable(AutoApprovedTable):
+    listed_text = _('Listed versions needing human review ({0})')
+    unlisted_text = _('Unlisted versions needing human review ({0})')
+    filters = {'needs_human_review': True}
+
     def render_addon_name(self, record):
         rval = [jinja2.escape(record.name)]
-        versions_flagged_by_scanners = record.versions.filter(
+
+        listed_versions = record.versions.filter(
             channel=amo.RELEASE_CHANNEL_LISTED,
-            needs_human_review=True).count()
-        if versions_flagged_by_scanners:
+            **self.filters,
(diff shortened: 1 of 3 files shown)


willdurand:
@diox do you think that this fix should be cherry-picked?

diox:
I'll review it properly later but it looks ok, I don't mind cherry-picking it.

src/olympia/reviewers/utils.py
```diff
@@ -271,7 +271,33 @@ def render_addon_name(self, record):
 
 
 class MadReviewTable(AutoApprovedTable):
-    pass
+    def render_addon_name(self, record):
+        rval = [jinja2.escape(record.name)]
+        listed_versions = record.versions.filter(
+            channel=amo.RELEASE_CHANNEL_LISTED,
+            versionscannerflags__needs_human_review_by_mad=True,
```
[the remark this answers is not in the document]
willdurand: ha yes, makes sense, thanks. I updated the patch.

src/olympia/reviewers/utils.py
```diff
@@ -241,37 +241,46 @@ def _get_addon_name_url(self, record):
...
             )
+
         return ''.join(rval)
 
 
-class MadReviewTable(AutoApprovedTable):
-    pass
+class MadReviewTable(ScannersReviewTable):
+    listed_text = _('Listed versions ({0})')
+    unlisted_text = _('Unlisted versions ({0})')
```
diox, line 282: Not sure this is better than the text from the parent class. Ultimately, we're only showing specific versions, not all, so without the extra `needing human review` bit it's a bit misleading, maybe.

willdurand: The name of the queue is "needs human review", that's why I removed it here.

Issue #14283, opened 2020-05-14
Improve 500 error handling when caused by django/middleware


diox:
When a 500 is caused by django request handling or middlewares, we don't go through all middlewares correctly and as a result, we might not have a session or user on the `request` object. Also, we might not have what we need to identify that it's an API request. This causes us to, for such exceptions:
- Try to return a pretty HTML 500 even if the request was made from the API
- Fail to return that pretty HTML 500, because the template wants `request.user` to be present.

Issue #14282, opened 2020-05-14
Remove ScrubRequestOnException


diox:
`ScrubRequestOnException` was imported a while ago but it's obsolete these days:
- For `POST` data it only cares about `password` field, but we haven't handled passwords for years
- For cookies it only cares about the session cookie

We're using sentry for our error handling and its data scrubbing is already a lot more efficient than this.

PR #14285, opened 2020-05-14
Cleanup 500

eviljeff:
Fixes https://github.com/mozilla/addons-server/issues/14283
Fixes https://github.com/mozilla/addons-server/issues/14282


src/olympia/amo/views.py
```diff
@@ -90,6 +90,15 @@ def handler404(request, exception=None, **kwargs):
...
 def handler500(request, **kwargs):
+    if getattr(request, 'is_api', False):
+        # API exceptions happening in DRF code would be handled with by our
+        # custom_exception_handler function in olympia.api.exceptions, but in
+        # the rare case where the exception is caused by a middleware or django
+        # itself, it might not, so we need to handle it here.
+        return HttpResponse(
+            json.dumps({'detail': 'Internal Server Error'}),
+            content_type='application/json',
+            status=500)
```
eviljeff, line 101: if using the entire code from `custom_exception_handler` isn't practical, can we can have:
```
if settings.DEBUG:
            data['traceback'] = traceback.format_exc()
```
included?

src/olympia/api/middleware.py
```diff
@@ -6,12 +6,15 @@
...
-        LocaleAndAppURLMiddleware has activated."""
+    def identify_request(self, request):
         request.is_api = re.match(settings.DRF_API_REGEX, request.path_info)
 
+    def process_request(self, request):
+        self.identify_request(request)
+
+    def process_exception(self, request, exception):
+        self.identify_request(request)
+
```
eviljeff, line 17: maybe not high priority here, but if we're rewriting IdentifyAPIRequestMiddleware we could drop the deprecated middleware style

diox: I don't want to make too many changes since I might cherry-pick this

src/olympia/amo/views.py
diox: I will re-use same function

src/olympia/reviewers/templates/reviewers/addon_details_box.html
```diff
@@ -64,12 +64,6 @@ <h2>{{ addon_name }}</h2>
               </td>
             </tr>
           {% endif %}
-          {% if version and version.compatible_apps[APP] %}
```
eviljeff, line 67: we can leave this in an replace with either `version.compatible_apps[amo.FIREFOX]` or a loop that shows all `version.compatible_apps`

diox: I did that initially but the value seems dubious at best, and I ran into problems because it's also used as an include in abuse admin... I ran out of motivation to shave yet another yak...

PR #14295, opened 2020-05-15
prevent blocked versions from being uploaded

### LINKED ISSUE #14029: Prevent signing of new versions if a guid is blocked and the version range is open-ended
If a new add-on version is uploaded for a guid that is already blocked (e.g. an AMO block already exists) and the version range blocked matches the new version, it would makes sense for that new version to not be allowed to be signed.

If we continue to allow signing we'll need to continually issue new blocks when generating new block dbs / stashes. 

/cc @jvillalobos and @wagnerand for input in-case there are any other considerations to take into account before doing this.

eviljeff:
fixes #14029

### DIFF
--- a/src/olympia/signing/tests/test_views.py
+++ b/src/olympia/signing/tests/test_views.py
@@ -18,8 +18,10 @@ from olympia.addons.models import Addon, AddonUser
 from olympia.amo.templatetags.jinja_helpers import absolutify
 from olympia.amo.tests import (
     TestCase, addon_factory, create_default_webext_appversion,
-    developer_factory, get_random_ip, reverse_ns)
+    developer_factory, get_random_ip, reverse_ns, user_factory)
+from olympia.amo.urlresolvers import reverse
 from olympia.api.tests.utils import APIKeyAuthTestMixin
+from olympia.blocklist.models import Block
 from olympia.files.models import File, FileUpload
 from olympia.signing.views import VersionView
 from olympia.users.models import (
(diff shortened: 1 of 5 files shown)


diox:
This doesn't handle the case where the version is submitted using the API (which, unfortunately, has a separate upload logic...)

eviljeff:
>This doesn't handle the case where the version is submitted using the API (which, unfortunately, has a separate upload logic...)

ugh.  I figured it reused it because the error message about existing versions wasn't duplicated (we used a different error message so I didn't find it)

eviljeff:
https://github.com/mozilla/addons-server/pull/14295/commits/b2b962d0c3134c4904100f5171ff14741c0fe9c8 adds the restriction for the signing API too

src/olympia/signing/tests/test_views.py
```diff
@@ -64,18 +66,18 @@ def create_version(self, version):
         response = self.request('PUT', self.url(self.guid, version), version)
         assert response.status_code in [201, 202]
 
-    def xpi_filepath(self, addon, version):
+    def xpi_filepath(self, guid, version):
```
eviljeff, line 69: sorry for the noise here... the guid param being called `addon` was an itch I had to scratch

src/olympia/signing/views.py
```diff
@@ -232,15 +236,30 @@ def handle_upload(self, request, addon, version_string, guid=None):
...
+            if not amo.ADDON_GUID_PATTERN.match(guid):
                 raise forms.ValidationError(
                     ugettext('Invalid Add-on ID in URL or package'),
                     status.HTTP_400_BAD_REQUEST)
 
+        block_qs = Block.objects.filter(guid=addon.guid if addon else guid)
+        if block_qs and block_qs.first().is_version_blocked(version_string):
+            msg = ugettext(
+                'Version {version} matches {block_url} for this add-on. '
+                'You can contact {amo_admins} for additional information.')
```
diox, line 254: You have different variables than the message in devhub so this would be a different string - could you change this so that we don't have 2 different translations ?

eviljeff: the string is slightly different because the devhub message is html (`<a href...` for the links/email) and the api is plain text with the url directly in the content.  I'll attempt to rephrase so it works for both places.

eviljeff: https://github.com/mozilla/addons-server/pull/14295/commits/aa2cc12040ec960ded4f14e1f6b0076b94f5a66e

Issue #14289, opened 2020-05-15
Only git-extract web extensions


willdurand:
We currently git-extract everything but this isn't needed. Let's only extract addons with `type=ADDON_EXTENSION` and `is_webextension=True`.

ioanarusiczki:
@willdurand I tested on AMO dev with FF76(Win10). 
- the static themes are not extracted anymore and won't show up in the admin queue.
- dictionaries and lang packs are no longer extracted either - are they part of the plan too?

willdurand:
>     * dictionaries and lang packs are no longer extracted either - are they part of the plan too?

Yes

ioanarusiczki:
@willdurand Alright, marking this verified on AMO dev with FF76(win10).

PR #14299, opened 2020-05-15
Only git-extract web-extensions

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14289

### DIFF
--- a/src/olympia/git/management/commands/git_extraction.py
+++ b/src/olympia/git/management/commands/git_extraction.py
@@ -6,6 +6,7 @@ from django.core.management.base import BaseCommand
 
 import olympia.core.logger
 
+from olympia import amo
 from olympia.amo.decorators import use_primary_db
 from olympia.files.utils import lock
 from olympia.git.models import GitExtractionEntry
@@ -67,6 +68,15 @@ class Command(BaseCommand):
         addon = entry.addon
         log.info('Starting git extraction of add-on "{}".'.format(addon.pk))
 
+        # See: https://github.com/mozilla/addons-server/issues/14289
+        if addon.type != amo.ADDON_EXTENSION:
+            log.info(
+                'Skipping git extraction of add-on "{}": not an '
+                'extension.'.format(addon.pk)
+            )
+            entry.delete()
+            return
+
         # We cannot use `entry.in_progress` because we have to be sure of the
         # add-on state and `entry` might not reflect the most up-to-date
         # database state here.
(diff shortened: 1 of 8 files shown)


willdurand:
ready :)

src/olympia/versions/tasks.py
```diff
@@ -124,10 +124,18 @@ def extract_version_to_git(
     # we can access them.
     version = Version.unfiltered.get(pk=version_id)
 
+    if (
+        version.addon.type != amo.ADDON_EXTENSION or
+        not version.all_files[0].is_webextension
+    ):
+        log.debug('Skipping git extraction of add-on "{}": not a '
+                  'web-extension.'.format(version.addon.id))
```
diox, line 132: I don't think this is enough.

You should also exclude those from the management command, so that we never call the extraction function on those versions, otherwise if the first `BATCH_SIZE` versions of an add-on are non-webextensions we'd never extract anything more.

You should still keep this, but maybe make the logging a `warn` instead, cause if the management command is fixed this should be unreachable, so we want to log a little louder if somehow it happens anyway.

willdurand: > You should also exclude those from the management command, so that we never call the extraction function on those versions, otherwise if the first BATCH_SIZE versions of an add-on are non-webextensions we'd never extract anything more.

heh, good catch 😥

willdurand:
Sorry about that, our discussion this morning made me double-check that we would also skip non-webextensions in the CRON task but I somehow did not realize that it would keep selecting the same versions because we don't update the `git_hash`...

src/olympia/git/management/commands/git_extraction.py
```diff
@@ -67,6 +68,15 @@ def extract_addon(self, entry, batch_size=BATCH_SIZE):
         addon = entry.addon
         log.info('Starting git extraction of add-on "{}".'.format(addon.pk))
 
+        # See: https://github.com/mozilla/addons-server/issues/14289
+        if addon.type != amo.ADDON_EXTENSION:
```
diox, line 72: Can we avoid them entering the queue at all? Even avoid entering the `extract_version_to_git` task ?

Which reminds me: let's not do this now, but ultimately, when we remove the code to extract without using the cron, we'll probably want to replace calls to `extract_version_to_git.delay` at upload/signing time entirely and replace them with just a function call doing the `GitExtractionEntry.objects.create(addon=version.addon)` bit, to avoid unnecessarily triggering tasks for just a single database insert.

willdurand: > Can we avoid them entering the queue at all? Even avoid entering the `extract_version_to_git` task ?

Yes, we could do this once we do what you said next: once we remove `extract_version_to_git.delay`, we'll have to create a `enqueue_git_extraction` or something and that's where we'll prevent non-webextensions to be added to the queue.

With this patch, we'll run `extract_version_to_git.delay` but non-webextensions won't be added to the queue anymore.

diox: Yeah that works for me.

willdurand: I filed https://github.com/mozilla/addons-server/issues/14311

Issue #14091, opened 2020-04-29
Help reduce logging volume in production


bqbn:
### Describe the problem and steps to reproduce it:
AMO generates about 600+ GB logs per month. All those go across the Internet from AWS to GCP, and incur data transfer fees.

Can we configure the app to suppress at least DEBUG level logs for production?

The celery component in particular, seems to generate a lot DEBUG level logs.

### What did you expect to happen?

Suppress at least DEBUG level logs and/or unnecessary logs to help reduce cost and possibly improve performance.

### Anything else we should know?
n/a

diox:
Agreed, we log too much. This is mostly for legacy reasons (some of the logs have always been there and we haven't bothered to touch them, and also on the other hand we've had to debug bizarre and complex legacy code), but we should fix this, it's even annoying to look at the sea of logs in stackdriver.

Before suppressing `DEBUG`, we should start by going through all `log.debug` calls and figure out if they are really debug logs.

diox:
@bqbn is there a way for you to export the most common log types and severity seen (from `jsonPayload.Type` + `jsonPayload.severity`), or specific `DEBUG` messages you've seen ? That would give a different starting point that just doing through all `debug` logs to figure out if they matter (which I still want to do, but is quite painful)

bqbn:
* for `olympia_web` component

  Please see if we can suppress this line of log

  `"download file xxxxxxx: addon/version/file not public and user None does not have relevant permissions."`

  The above line accounts for 96% of `olympia_web` logs, and it doesn't seem to be very useful to anybody. An example looks like this:

  ```
  {"Timestamp": 1589473368810223872, "Type": "z.versions", "Logger": "http_app_addons", "Hostname": "ip-172-31-44-31.us-west-2.compute.internal", "EnvVersion": "2.0", "Severity": 6, "Pid": 17, "Fields": {"uid": "<anon>", "remoteAddressChain": "159.65.67.14", "msg": "download file 563330: addon/version/file not public and user None does not have relevant permissions."}, "severity": 200}
  ```

* for `olympia_celery` component

  It looks like every task has at least 3 log messages, `Received task`, `starting task timer` and `tracking task run time`. I actually don't know how useful they are, so I'll let you decide if you want to suppress them or not, here are an example of those,
  ```
[...]

PR #14310, opened 2020-05-18
Update logging config

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14091

---

- This patch changes most `.debug()` calls to `.info()`
- Change some specific `.info()` calls to reduce the amount of logs sent to Stackdriver
- Consolidate the logging config to only have one config for the hosted envs and one for local dev

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -20,7 +20,7 @@ INSTALLED_APPS += (
 )
 
 # Override logging config to enable DEBUG logs for (almost) everything.
-LOGGING['root'] = {'handlers': ['mozlog'], 'level': logging.DEBUG}
+LOGGING['root']['level'] = logging.DEBUG
 for logger in list(LOGGING['loggers'].keys()):
     if logger not in ['filtercascade', 'mohawk.util', 'post_request_task']:
         del LOGGING['loggers'][logger]
--- a/settings.py
+++ b/settings.py
@@ -19,33 +19,11 @@ INSTALLED_APPS += (
     'debug_toolbar',
 )
 
-LOGGING['root'].update({'level': logging.DEBUG})
-
-LOGGING['loggers'].update({
-    'amo': {'level': logging.DEBUG},
-    'amqp': {'level': logging.DEBUG},
-    'caching': {'level': logging.DEBUG},
-    'caching.invalidation': {'level': logging.DEBUG},
-    'django': {'level': logging.DEBUG},
-    'django.security.csrf': {'level': logging.DEBUG},
-    'elasticsearch': {'level': logging.DEBUG},
-    # filtercascade
-    # mohawk.util
-    'newrelic': {'level': logging.DEBUG},
-    'parso': {'level': logging.DEBUG},
-    # post_request_task
-    'raven': {'level': logging.DEBUG},
-    'rdflib': {'level': logging.DEBUG},
-    'requests': {'level': logging.DEBUG},
-    'request.summary': {'level': logging.DEBUG},
-    's.client': {'level': logging.DEBUG},
-    'z': {'level': logging.DEBUG},
-    'z.addons': {'level': logging.DEBUG},
-    'z.celery': {'level': logging.DEBUG},
-    'z.es': {'level': logging.DEBUG},
-    'z.pool': {'level': logging.DEBUG},
-    'z.task': {'level': logging.DEBUG},
-})
+# Override logging config to enable DEBUG logs for (almost) everything.
+LOGGING['root'] = {'handlers': ['mozlog'], 'level': logging.DEBUG}
+for logger in list(LOGGING['loggers'].keys()):
+    if logger not in ['filtercascade', 'mohawk.util', 'post_request_task']:
+        del LOGGING['loggers'][logger]
 
 # django-debug-doolbar middleware needs to be inserted as high as possible
 # but after GZip middleware
--- a/settings.py
+++ b/settings.py
@@ -19,6 +19,33 @@ INSTALLED_APPS += (
     'debug_toolbar',
 )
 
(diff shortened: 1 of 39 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -1345,14 +1345,24 @@ def get_db_config(environ_var, atomic_requests=True):
...
+        'request': {
+            'handlers': ['mozlog'],
+            'level': logging.WARNING,
+            'propagate': False
+        },
         'request.summary': {
             'handlers': ['mozlog'],
-            'level': logging.DEBUG,
+            'level': logging.INFO,
             'propagate': False
```
willdurand, line 1366: I suppose this is the default config now (inherited from `root`) so we could probably remove this entry.

src/olympia/lib/settings_base.py
```diff
@@ -1362,7 +1372,12 @@ def get_db_config(environ_var, atomic_requests=True):
...
         'z': {
             'handlers': ['mozlog'],
-            'level': logging.DEBUG,
+            'level': logging.INFO,
+            'propagate': False
+        },
+        'z.addons': {
+            'handlers': ['mozlog'],
+            'level': logging.INFO,
             'propagate': False
```
willdurand, line 1381: I suppose this is the default config now (inherited from `root`) so we could probably remove this entry.

src/olympia/lib/settings_base.py
```diff
@@ -1382,7 +1397,7 @@ def get_db_config(environ_var, atomic_requests=True):
         },
         'z.task': {
             'handlers': ['mozlog'],
-            'level': logging.DEBUG,
+            'level': logging.INFO,
             'propagate': False
```
willdurand, line 1401: I suppose this is the default config now (inherited from `root`) so we could probably remove this entry.

src/olympia/versions/views.py
```diff
@@ -91,9 +91,9 @@ def is_appropriate_reviewer(addon, channel):
         has_permission = True
 
     if not has_permission:
-        log.info('download file {file_id}: addon/version/file not public and '
-                 'user {user_id} does not have relevant permissions.'.format(
-                     file_id=file_id, user_id=request.user.pk))
+        log.debug('download file {file_id}: addon/version/file not public and '
```
willdurand, line 94: This is on purpose.

willdurand:
cc @eviljeff @diox for a first look

src/olympia/lib/settings_base.py
eviljeff: yes

src/olympia/lib/settings_base.py
eviljeff: also yes

src/olympia/lib/settings_base.py
eviljeff: double check that `request.summary` doesn't inherit from `request`

settings.py
```diff
@@ -19,6 +19,33 @@
...
+    'requests': {'level': logging.DEBUG},
+    'request.summary': {'level': logging.DEBUG},
+    's.client': {'level': logging.DEBUG},
+    'z': {'level': logging.DEBUG},
+    'z.addons': {'level': logging.DEBUG},
+    'z.celery': {'level': logging.DEBUG},
+    'z.es': {'level': logging.DEBUG},
+    'z.pool': {'level': logging.DEBUG},
+    'z.task': {'level': logging.DEBUG},
+})
```
eviljeff, line 48: Can we just delete all these from `LOGGING['loggers']` so they just inherit from the root?  Seems cleaner than trying to redefine them.

settings.py
```diff
@@ -19,6 +19,33 @@
     'debug_toolbar',
 )
 
+LOGGING['root'].update({'level': logging.DEBUG})
```
eviljeff, line 22: yeah... I know this is what were (are!) doing in the conf/ files below but this isn't how dict updates work - 
before:
`LOGGING['root']` =  `{'handlers': ['mozlog'], 'level': logging.INFO}`
then running the `.update`
`LOGGING['root'].update({'level': logging.DEBUG})`
after
`LOGGING['root']` = `'root': {'level': logging.INFO}`
we lose the other dict elements because update isn't recursive.

src/olympia/lib/settings_base.py
willdurand: looks like this one comes from dockerflow so it shouldn't inherit from `request`

src/olympia/lib/settings_base.py
```diff
@@ -1279,16 +1279,16 @@ def get_db_config(environ_var, atomic_requests=True):
             'class': 'django_statsd.loggers.errors.StatsdHandler',
         },
     },
-    'root': {'handlers': ['mozlog'], 'level': logging.DEBUG},
+    'root': {'handlers': ['mozlog'], 'level': logging.INFO},
     'loggers': {
         'amo': {
             'handlers': ['mozlog'],
             'level': logging.DEBUG,
```
eviljeff, line 1286: are we keeping this on at `DEBUG` level on purpose?

settings.py
```diff
@@ -19,6 +19,11 @@
     'debug_toolbar',
 )
 
+# Override logging config to enable DEBUG logs for (almost) everything.
+LOGGING['root'] = {'handlers': ['mozlog'], 'level': logging.DEBUG}
```
eviljeff, line 23: this is the most explicit, but I suppose `LOGGING['root']['level'] = logging.DEBUG` works too

src/olympia/lib/settings_base.py
willdurand: Good catch, this one should also be `INFO`.

Issue #14268, opened 2020-05-13
A 'good' labelled scanner result doesn't change to 'bad' if the add-on is blocked


AlexandraMoga:
Follow up for #14086 

### Describe the problem and steps to reproduce it:

1. Send the following API request: `https://addons-dev.allizom.org/api/v4/scanner/results/?label=good`
2. Pick a scanner result and look for its ID in the list of scanner results admin
3. Identify the add-on it belongs to and block it
4. After the block is added and approved, repeat step 1
5. Search for the scanner ID belonging to the blocked add-on

### What happened?
The scanner ID is still found under the `label=good` results

### What did you expect to happen?
The scanner result label should have changed to `bad` since the add-on is blocked

### Anything else we should know?
To narrow down the search, you can send this request https://addons-dev.allizom.org/api/v4/scanner/results/?label=good&scanner=mad and look for scanner ID `9230`.
The corresponding add-on can be found in the [result detail](https://addons-internal.dev.mozaws.net/en-US/admin/models/scanners/scannerresult/9230/change/?_changelist_filters=has_matched_rules%3Dall%26has_version%3Dall%26p%3D2%26scanner__exact%3D4%26state%3Dall). I've already blocked the add-on so the label should have changed.

AlexandraMoga:
Verified fixed on -dev.
A `good` scanner result will receive a `bad` label if the add-on version is blocked. 
I've checked that there are no more entries for the same result when filtering by `label=good`
As a note, once a scanner result is labelled as `bad` it won't be changed to `good` if auto-approval is confirmed afterwards (although, in case of a blocked add-on I don't think approval will be confirmed  anymore)

PR #14313, opened 2020-05-18
Do not mark a result as good if it is also bad

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14268

---

A result can only be "good" if it is not "bad".

### DIFF
--- a/src/olympia/scanners/tests/test_views.py
+++ b/src/olympia/scanners/tests/test_views.py
@@ -8,7 +8,13 @@ from olympia.amo.tests import (
     user_factory,
     version_factory,
 )
-from olympia.constants.scanners import YARA, CUSTOMS, WAT, TRUE_POSITIVE
+from olympia.constants.scanners import (
+    CUSTOMS,
+    LABEL_BAD,
+    TRUE_POSITIVE,
+    WAT,
+    YARA,
+)
 from olympia.blocklist.models import Block
 from olympia.blocklist.utils import block_activity_log_save
 from olympia.scanners.models import ScannerResult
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/tests/test_views.py
```diff
@@ -326,3 +332,29 @@ def test_get_results_with_blocked_versions(self):
...
+        )
+        VersionLog.objects.create(
+            activity_log=ActivityLog.create(
+                action=amo.LOG.CONFIRM_AUTO_APPROVED,
+                version=version_1,
+                user=self.user,
+            ),
+            version=version_1,
+        )
+        # Oh noes! The version has been blocked.
```
bobsilverberg, line 351: Would it be worth adding a check assertion here to verify that at this point the result would be labelled `good`, to make sure that the reason it is labelled `bad` below is valid?

src/olympia/scanners/views.py
```diff
@@ -46,6 +46,13 @@ def get_queryset(self):
             bad_results = bad_results.filter(scanner=scanner)
             good_results = good_results.filter(scanner=scanner)
 
+        bad_filters = Q(state=TRUE_POSITIVE) | Q(
+            version__versionlog__activity_log__action__in=(
+                amo.LOG.BLOCKLIST_BLOCK_ADDED.id,
+                amo.LOG.BLOCKLIST_BLOCK_EDITED.id,
```
bobsilverberg, line 52: I notice that there are two different reasons that a `good` result could be labelled `bad` as a result of being blocked (i.e., `BLOCKLIST_BLOCK_ADDED` and `BLOCKLIST_BLOCK_EDITED`), but it looks like a test was only added for one of these. Is that a concern, or do you feel the test cases are sufficient to cover the code change?

willdurand: I think it is fine as it is.

src/olympia/scanners/tests/test_views.py
willdurand: It is covered by other test cases so I would say it is fine?

src/olympia/scanners/views.py
bobsilverberg: Yes, we do generally take that approach, although I must say I'm not a huge fan of it. It does cut down on test code, but I actually think it's better if one can read a single test in isolation and be able to tell that it's correctly testing what needs to be tested. This approach, which, again, I know we commonly take, means that we have to have knowledge of other tests in the file to be sure that our test setup is correct.

Personally, I think one additional check assertion in a test adds more value than it takes away (due to maintenance), but I know this isn't how we usually do things, and this PR isn't necessarily the place to change that approach. In other words, let's leave it as is, but perhaps this is a topic worth discussing further.

src/olympia/scanners/tests/test_views.py
bobsilverberg: Yes, we do generally take that approach, although I must say I'm not a huge fan of it. It does cut down on test code, but I actually think it's better if one can read a single test in isolation and be able to tell that it's correctly testing what needs to be tested. This approach, which, again, I know we commonly take, means that we have to have knowledge of other tests in the file to be sure that our test setup is correct.

Personally, I think one additional check assertion in a test adds more value than it takes away (due to maintenance), but I know this isn't how we usually do things, and this PR isn't necessarily the place to change that approach. In other words, let's leave it as is, but perhaps this is a topic worth discussing further.

src/olympia/scanners/views.py
bobsilverberg: Oops, I added my reply to the wrong comment. I moved it.

Issue #14319, opened 2020-05-18
Use git extraction "queue" as a "stack"


willdurand:
This is an idea that came from a discussion with @wagnerand (initially about mass-extraction). I thought about it and that might work but let's see what others think (hello @diox).

Regardless of the mass-extraction plan I am about to explain, we need a limit on the number of git extraction entries processed by the CRON task to not overload the workers.

The mass-extraction plan would be to create git extraction entries for all add-ons that are not correctly extracted (something like 45K but to be confirmed) at the beginning of the queue (with a date equal to "yesterday" or something to make sure we don't have recent submissions in between).

We would not consume the queue from the start but from the end, like a stack (so that would be a stack and not a queue anymore). In other words, we would process the last _X_ entries every time. This would allow us to not block the extraction of recent versions while git extracting the remaining add-ons when possible, but in a controlled manner (thanks to the limit).

For normal execution, that wouldn't change much and that would probably be better to enforce a limit. We don't have a lot of submissions per day anyway so taking entries from the start or the end would not matter a lot.

Thoughts?

diox:
Seems good, but we need to plan for spam/increase in submission. I wonder what the limit should be. The cron in itself doesn't do much, but the tasks need some breathing room to be executed without overloading the workers.

bobsilverberg:
This sounds like a good idea to me, and sounds similar to a question I asked in Matrix last week. It seemed to be that we should be processing the entries in reverse chronological order.

willdurand:
This is mainly because of the mass-extraction. Otherwise, it makes sense to git extract the versions that have been uploaded first.

ioanarusiczki:
@willdurand I verified this issue on AMO dev - FF67(Win10):
- the last submitted addons are extracted first. 
- unlike what I've noticed for submitting one new version or a couple more (up to 5) for https://github.com/mozilla/addons-server/issues/14311 this time an addon had 2 entries (after I refreshed the queue).

PR #14320, opened 2020-05-18
Limit the number of entries processed by a single execution (and change queue to stack)

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14319

---

I set the limit to `20` for now, I don't know if that's too high or okay for now (maybe too high).

With this configuration, we should only have up to 20 add-ons that are git-extracted in parallel. If we have 20 add-ons still being git extracted and no new add-on, this would mean a new execution of the CRON task wouldn't do anything but I think it is what we want, no?


willdurand:
(rebased)

PR #14333, opened 2020-05-19
Do not log 'received task' anymore

willdurand:
Refs https://github.com/mozilla/addons-server/issues/14091 

---

This will remove the "Received task" logs in Stackdriver.

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1296,6 +1296,11 @@ LOGGING = {
             'level': logging.INFO,
             'propagate': False
         },
+        'celery.worker.strategy': {
+            'handlers': ['mozlog'],
+            'level': logging.WARNING,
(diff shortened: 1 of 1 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -1296,6 +1296,11 @@ def get_db_config(environ_var, atomic_requests=True):
             'level': logging.INFO,
             'propagate': False
         },
+        'celery.worker.strategy': {
+            'handlers': ['mozlog'],
+            'level': logging.WARNING,
```
bobsilverberg, line 1301: So this indicates that only logs with a level of `WARNING` or higher would be sent?

willdurand: Yes and we see in Stackdriver that the "Received..." log message is an INFO one.

PR #14343, opened 2020-05-20
Allow django admin userprofile lookup by email

### LINKED ISSUE #14342: Allow accessing django admin user profile by email
It would be helpful for admins if they could access django admin user profile pages by email, i.e. `/admin/models/users/userprofile/<email>`

wagnerand:
Fix #14342.

### DIFF
--- a/src/olympia/users/admin.py
+++ b/src/olympia/users/admin.py
@@ -1,5 +1,6 @@
 import functools
 
+from django import http
 from django.conf.urls import url
 from django.contrib import admin, messages
 from django.contrib.admin.utils import unquote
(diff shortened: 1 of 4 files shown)


src/olympia/users/models.py
```diff
@@ -189,6 +189,15 @@ def __init__(self, *args, **kw):
     def __str__(self):
         return u'%s: %s' % (self.id, self.display_name or self.username)
 
+    @classmethod
+    def get_lookup_field(cls, identifier):
```
willdurand, line 193: Would be nice to have separate test coverage for this method.

src/olympia/users/admin.py
```diff
@@ -130,6 +131,19 @@ def change_view(self, request, object_id, form_url='', extra_context=None):
         extra_context = extra_context or {}
         extra_context['has_users_edit_permission'] = acl.action_allowed(
             request, amo.permissions.USERS_EDIT)
+
+        lookup_field = UserProfile.get_lookup_field(object_id)
+        if lookup_field != 'pk':
+            try:
+                if lookup_field == 'email':
```
willdurand, line 138: I would remove this condition so that, if `lookup_field` is something else, it should raise `UserProfile.DoesNotExist`. Otherwise you'll have to add a `else` branch to handle the case where `lookup_field` is something else.

wagnerand: Clarified via chat that this is not necessary. We should let the method of the parent class handle other `lookup_field`s.

src/olympia/users/models.py
wagnerand: Added in 909b5096b6.

PR #14347, opened 2020-05-20
change "kinto" references to "remote settings"|"legacy"

### LINKED ISSUE #14346: de-kinto our remote settings integration
as noted in https://github.com/mozilla/addons-server/issues/13701#issuecomment-603377577 it's preferred that external services reference "remote settings" rather than "kinto".  The former is the product's name, the latter the codename for the current implementation.

There may be a few cases where it's not practical, but it's a sensible change (that I've been started making ad-hoc each time there's been a need to refactor).

eviljeff:
fixes #14346 
- no functional changes in commit "kinto -> remote settings or legacy" - just find&replace;
- had to manually rewrite the django migration in the other commit to not make db changes to the tables and columns.


src/olympia/blocklist/models.py
```diff
@@ -569,9 +572,13 @@ class KintoImport(ModelBase):
         OUTCOME_REGEXBLOCKS: 'Added blocks from regex',
         OUTCOME_NOMATCH: 'No matches',
     }
-    kinto_id = models.CharField(
-        unique=True, max_length=255, null=False, default='')
+    legacy_id = models.CharField(
+        unique=True, max_length=255, null=False, default='',
+        db_column='kinto_id')
```
bobsilverberg, line 577: So the column in the db remains `kinto_id`?

eviljeff:
> Wow, that was a lot of fun to review! Thanks for the opportunity,

you didn't need to read all the find&replace changes 😄

src/olympia/blocklist/models.py
eviljeff: yes.  The ordering means none of the migrations actually cause any SQL* be generated so the database is the same as before (zero downtime/impact during the push).  Without the migration file though django will continue to think/complain the database isn't up to date vs the modesl.

*this is what will be run in reality, from `./manage.py sqlmigrate blocklist 0015`:
```
BEGIN;
--
-- Rename table for KintoImport to blocklist_kintoimport
--
--
-- Rename model KintoImport to LegacyImport
--
--
-- Alter field kinto_id on block
--
--
-- Rename field kinto_id on block to legacy_id
--
COMMIT;
```
(it's all comments)

PR #14358, opened 2020-05-21
missing blocklist.models.LegacyImport migration

eviljeff:
follow-up from #14346 - when I manually fixed the migration I forgot about the `kinto_id` property on `LegacyImport` (nee `KintoImport`).


src/olympia/blocklist/migrations/0016_auto_20200521_1710.py
```diff
@@ -0,0 +1,27 @@
...
+
+    dependencies = [
+        ('blocklist', '0015_auto_20200520_1725'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='legacyimport',
+            options={},
+        ),
```
eviljeff, line 16: don't know what this one is for tbh, but it's harmless.

PR #14361, opened 2020-05-21
Remove `version` from retrieve and list draft_comments API responses

### LINKED ISSUE #14360: Remove most version info from the response of the draft_comments API
Currently the [`draft_comments` API ](https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#get--api-v4-reviewers-addon-(int-addon_id)-versions-(int-version_id)-draft_comments-) returns a full `version` object in its response. This is a lot of data and includes the complete `file_entries` which, for add-ons with a lot of files, can be very large indeed.

Code manager does not use *any* of this data, so I suggest we remove it from the response. Although code manager doesn't need even the `versionId`, I suggest we keep that in the response as I can see it being useful, because it allows the comments to be tied back to the version to which they apply.

bobsilverberg:
Fixes #14360 

This needs to land along with https://github.com/mozilla/addons-code-manager/pull/1602, otherwise Code Manager will break.

### DIFF
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -522,8 +522,8 @@ class DraftCommentSerializer(serializers.ModelSerializer):
     user = SplitField(
         serializers.PrimaryKeyRelatedField(queryset=UserProfile.objects.all()),
         BaseUserSerializer())
-    version = serializers.PrimaryKeyRelatedField(
-        queryset=Version.unfiltered.all())
+    version_id = serializers.PrimaryKeyRelatedField(
+        queryset=Version.unfiltered.all(), source='version')
     canned_response = SplitField(
         serializers.PrimaryKeyRelatedField(
             queryset=CannedResponse.objects.all(),
@@ -536,7 +536,7 @@ class DraftCommentSerializer(serializers.ModelSerializer):
         model = DraftComment
         fields = (
             'id', 'filename', 'lineno', 'comment',
-            'version', 'user', 'canned_response'
+            'version_id', 'user', 'canned_response'
         )
 
     def get_or_default(self, key, data, default=''):
@@ -577,9 +577,4 @@ class DraftCommentSerializer(serializers.ModelSerializer):
                 {'comment': ugettext(
                     'You can\'t submit a line number without associating '
                     'it to a filename.')})
-        return data
-
-    def to_representation(self, obj):
-        data = super(DraftCommentSerializer, self).to_representation(obj)
-        data['version_id'] = data.pop('version')
         return data
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -518,10 +518,21 @@ class CannedResponseSerializer(serializers.ModelSerializer):
         return amo.CANNED_RESPONSE_CATEGORY_CHOICES[obj.category]
 
 
+class VersionIdOnlySerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = Version
+        fields = ['id']
+
+
 class DraftCommentSerializer(serializers.ModelSerializer):
     user = SplitField(
         serializers.PrimaryKeyRelatedField(queryset=UserProfile.objects.all()),
         BaseUserSerializer())
+    version = SplitField(
+        serializers.PrimaryKeyRelatedField(
+            queryset=Version.unfiltered.all()),
+        VersionIdOnlySerializer())
     canned_response = SplitField(
         serializers.PrimaryKeyRelatedField(
             queryset=CannedResponse.objects.all(),
@@ -534,7 +545,7 @@ class DraftCommentSerializer(serializers.ModelSerializer):
         model = DraftComment
         fields = (
             'id', 'filename', 'lineno', 'comment',
-            'version_id', 'user', 'canned_response'
+            'version', 'user', 'canned_response'
         )
 
     def get_or_default(self, key, data, default=''):
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -541,12 +541,6 @@ class DraftCommentSerializer(serializers.ModelSerializer):
             'version', 'user', 'canned_response'
         )
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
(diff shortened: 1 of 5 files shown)


src/olympia/reviewers/serializers.py
```diff
@@ -586,3 +580,16 @@ def validate(self, data):
                     'You can\'t submit a line number without associating '
                     'it to a filename.')})
         return data
+
+
+class DraftCommentSerializerReadOnly(DraftCommentSerializer):
+    version_id = serializers.SerializerMethodField()
```
eviljeff, line 586: you can do `version_id = serializers.IntegerField(source='version.id')` directly if you don't need any additional processing in the `get....` method.

src/olympia/reviewers/serializers.py
```diff
@@ -586,3 +580,16 @@ def validate(self, data):
                     'You can\'t submit a line number without associating '
                     'it to a filename.')})
         return data
+
+
+class DraftCommentSerializerReadOnly(DraftCommentSerializer):
```
eviljeff, line 585: I think if the serializer is named `...ReadOnly` you should make all the fields be readonly.

src/olympia/reviewers/serializers.py
```diff
@@ -586,3 +580,16 @@ def validate(self, data):
...
+    version_id = serializers.SerializerMethodField()
+
+    class Meta(DraftCommentSerializer.Meta):
+        fields = (
+            'id', 'filename', 'lineno', 'comment',
+            'version_id', 'user', 'canned_response'
+        )
+
+    def get_version_id(self, obj):
+        return obj.version.id
```
eviljeff, line 595: So thinking about the parent `DraftCommentSerializer` class - it's got a `SplitField` which would SET a version (a pk) and GET a full serialized response - but when would that full response ever be used if the get responses are handled by `DraftCommentSerializerReadOnly`?

eviljeff: Is the solution just to change `DraftCommentSerializer` instead to always be a version_id?

src/olympia/reviewers/serializers.py
bobsilverberg: Thanks!

src/olympia/reviewers/serializers.py
bobsilverberg: That was the very first thing I tried, and it seemed fine for the `list` and `retrieve` cases, but broke all the tests for `create`. I, possibly incorrectly, assumed that those cases needed access to the `version` object, instead of `version_id`, which led me to this solution. I was just trying to get a result that works, but I think I went the wrong direction. How would I tell a single serializer that I need to be able to set a `version` (which is really just a `pk`), but that I don't want to return a `version` object, and rather only want to return a `version_id`?

src/olympia/reviewers/serializers.py
bobsilverberg: That makes sense. I called it `...ReadOnly` just because it's used for the `list` and `retrieve` cases, but yeah, if I'm going to use that name I should make sure it actually is readonly. Perhaps a better overall solution, which I asked about in another comment, will eliminate the need for this entirely.

src/olympia/reviewers/serializers.py
bobsilverberg: @eviljeff I guess I should have flagged you or asked for another review, as I think you might have missed the fact that I asked a question here. Can you help?

src/olympia/reviewers/serializers.py
eviljeff: actually, doh, it's even easier - `version_id` is already a property of `DraftComment` (each foreign key is available as both `.object` and `.object_pk`)

bobsilverberg: @eviljeff I made a change, but the tests are still failing. I imagine I need to make some sort of associated change to the view, but I can't seem to figure that out. Can you help?

bobsilverberg:
@eviljeff I figured out something that will work, using `SplitField` with a new, minimal serializer, but it seems like something as simple as this might be able to work without that additional serializer. Please let me know what you think when you have a chance to look at it.

eviljeff:
after _a lot_ of playing around locally it turns out it's just:
```diff
-    version = SplitField(
-        serializers.PrimaryKeyRelatedField(
-            queryset=Version.unfiltered.all()),
-        VersionIdOnlySerializer())
+    version = serializers.PrimaryKeyRelatedField(
+        queryset=Version.unfiltered.all())
```

Just needs a few tests (and docs) updating to indicate the response is the `version.pk` directly rather than `{id: <version.pk>}`

bobsilverberg:
> after _a lot_ of playing around locally it turns out it's just:
> 
> ```diff
> -    version = SplitField(
> -        serializers.PrimaryKeyRelatedField(
> -            queryset=Version.unfiltered.all()),
> -        VersionIdOnlySerializer())
> +    version = serializers.PrimaryKeyRelatedField(
> +        queryset=Version.unfiltered.all())
> ```
> 
> Just needs a few tests (and docs) updating to indicate the response is the `version.pk` directly rather than `{id: <version.pk>}`

Nice, thanks @eviljeff! I thought I actually tried that as one of my combinations, but I must have gotten something wrong. Thanks again for your help.

src/olympia/reviewers/serializers.py
```diff
@@ -522,10 +522,8 @@ class DraftCommentSerializer(serializers.ModelSerializer):
     user = SplitField(
         serializers.PrimaryKeyRelatedField(queryset=UserProfile.objects.all()),
         BaseUserSerializer())
-    version = SplitField(
-        serializers.PrimaryKeyRelatedField(
-            queryset=Version.unfiltered.all()),
-        AddonBrowseVersionSerializer())
+    version = serializers.PrimaryKeyRelatedField(
+        queryset=Version.unfiltered.all())
```
eviljeff, line 526: you can do `version_id = serializers.PrimaryKeyRelatedField(queryset=Version.unfiltered.all(), source='version')` and then you don't have to override it in `to_representation`  (`'version_id': self.get_version_object().pk` in `ReviewAddonVersionDraftCommentViewSet.get_extra_comment_data`

bobsilverberg:
> I feel like we've spent weeks on these few lines of code 😄

There have indeed been a few iterations, and I think each one has been simpler than the previous one! It's looking pretty good now.

Issue #14344, opened 2020-05-20
Multiple results for the same scanner ID returned when filtering by 'label'


AlexandraMoga:
### Describe the problem and steps to reproduce it:
1. Send the following api request: `https://addons-dev.allizom.org/api/v4/scanner/results/?label=bad&scanner=mad`
2. Search for scanner id `10421`

### What happened?
There are three entries for the same id. This happened because the api seems to count when the block was first added and also counts each edit for that block (two edits in the case of the mentioned id)

### What did you expect to happen?
There should be only one entry for the same scanner id

### Anything else we should know?
- if you omit the `label` filter, there are no identical results returned, i.e there is only one entry per  scanner id
- the same issue can be replicated when the `label` filter is set to `good`. In this case, the same scanner id can be returned multiple times if there are multiple approval confirmations for the same version (see bug #6497 for background)
Example: `https://addons-dev.allizom.org/api/v4/scanner/results/?label=good&scanner=mad` with id `10353`

willdurand:
We have test coverage for that in theory so I am a little surprised but I can reproduce the bug in -dev. I cannot reproduce it in a test case, though.

AlexandraMoga:
Verified fixed on -dev.
There is now only one entry per scanner id when filtering by label.

PR #14368, opened 2020-05-22
Fix scanner results when filter by label

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14344

### DIFF
--- a/src/olympia/scanners/tests/test_views.py
+++ b/src/olympia/scanners/tests/test_views.py
@@ -11,6 +11,7 @@ from olympia.amo.tests import (
 from olympia.constants.scanners import (
     CUSTOMS,
     LABEL_BAD,
+    MAD,
     TRUE_POSITIVE,
     WAT,
     YARA,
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/tests/test_views.py
```diff
@@ -73,67 +74,33 @@ def test_get(self):
...
         )
-        VersionLog.objects.create(
-            activity_log=ActivityLog.create(
-                action=amo.LOG.APPROVE_VERSION,
-                version=good_version_1,
-                user=self.user,
-            ),
-            version=good_version_1,
+        ActivityLog.create(
+            amo.LOG.APPROVE_VERSION, good_version_1, user=self.user
```
eviljeff, line 78: are all these changes also connected to the issue?

willdurand: no, that's a bit of refactoring because `ActivityLog.create()` already deals with version log.

Issue #14355, opened 2020-05-21
investigate our logging config and why some ERROR are going to sentry


eviljeff:
The underlying cause of #14348 is `log.error(...)` calls are ending up in sentry now when they weren't before.  #14310 is the likely cause because it tinkered with our logging setup though it's unclear _what_ change exactly.

diox:
This looks slightly suspicious (for tasks specically) : https://github.com/mozilla/addons-server/blob/ac03717486f78393068b49d8e1e37d3eca9b34e6/src/olympia/amo/celery.py#L84-L94

eviljeff:
https://github.com/mozilla/addons-server/pull/14310/files#diff-9a87f4a997c3c97f3e8c7768e6dc00d6L1383-L1387
```diff
-  'z.task': {
-            'handlers': ['mozlog'],
-            'level': logging.DEBUG,
-            'propagate': False
-        }
``` 
touched `z.task`'s settings.  I suppose the ` 'propagate': False` in particular stopped the errors from cascading before.

PR #14380, opened 2020-05-25
Pass params to log statements in activity task in a way that allows aggregation

### LINKED ISSUE #14378: olympia.activity.tasks.process_email is logging too much
Split from https://github.com/mozilla/addons-server/issues/14355 because the `process_email` task logging in itself is an issue we might want to fix sooner.

This task is logging too much, and what's worse the logs aren't aggregated properly, so it clogs up sentry.

diox:
Fixes #14378

Wasn't sure if those were valuable to keep as `error()`, what do you think?

### DIFF
--- a/src/olympia/activity/tasks.py
+++ b/src/olympia/activity/tasks.py
@@ -27,7 +27,7 @@ def process_email(message, spam_rating, **kwargs):
         return
     _, created = ActivityLogEmails.objects.get_or_create(messageid=msg_id)
     if not created:
-        log.error('Already processed email [%s], skipping', msg_id, extra={
+        log.warning('Already processed email [%s], skipping', msg_id, extra={
             'message_obj': message
         })
         return
--- a/src/olympia/activity/tasks.py
+++ b/src/olympia/activity/tasks.py
@@ -21,15 +21,19 @@ def process_email(message, spam_rating, **kwargs):
             if header.get('Name', '').lower() == 'message-id':
                 msg_id = header.get('Value')
     if not msg_id:
-        log.error('No MessageId in message, aborting.')
-        log.error(message)
+        log.error('No MessageId in message, aborting.', extra={
+            'message_obj': message
+        })
         return
     _, created = ActivityLogEmails.objects.get_or_create(messageid=msg_id)
     if not created:
-        log.error('Already processed [%s], skipping' % msg_id)
-        log.error(message)
+        log.error('Already processed email [%s], skipping', msg_id, extra={
+            'message_obj': message
+        })
         return
     res = add_email_to_activity_log_wrapper(message, spam_rating)
 
     if not res:
-        log.error('Failed to save email.')
+        log.error('Failed to process email [%s].', msg_id, extra={
+            'message_obj': message
+        })


src/olympia/activity/tasks.py
```diff
@@ -21,15 +21,19 @@ def process_email(message, spam_rating, **kwargs):
...
+            'message_obj': message
+        })
         return
     _, created = ActivityLogEmails.objects.get_or_create(messageid=msg_id)
     if not created:
-        log.error('Already processed [%s], skipping' % msg_id)
-        log.error(message)
+        log.error('Already processed email [%s], skipping', msg_id, extra={
+            'message_obj': message
+        })
```
eviljeff, line 32: I don't think this needs to be an error - a warning at best.

src/olympia/activity/tasks.py
```diff
@@ -21,15 +21,19 @@ def process_email(message, spam_rating, **kwargs):
             if header.get('Name', '').lower() == 'message-id':
                 msg_id = header.get('Value')
     if not msg_id:
-        log.error('No MessageId in message, aborting.')
-        log.error(message)
+        log.error('No MessageId in message, aborting.', extra={
+            'message_obj': message
+        })
```
eviljeff, line 26: this is still an error, but I don't think we need to propagate it to sentry.

src/olympia/activity/tasks.py
```diff
@@ -21,15 +21,19 @@ def process_email(message, spam_rating, **kwargs):
...
+        log.error('Already processed email [%s], skipping', msg_id, extra={
+            'message_obj': message
+        })
         return
     res = add_email_to_activity_log_wrapper(message, spam_rating)
 
     if not res:
-        log.error('Failed to save email.')
+        log.error('Failed to process email [%s].', msg_id, extra={
+            'message_obj': message
```
eviljeff, line 38: same here.  I think we just need to add the config `propagate=False` to these logging statements but keep them as error (we might need a different namespace - not sure if `z.amo.activity` is used elsewhere).

src/olympia/activity/tasks.py
```diff
@@ -21,15 +21,19 @@ def process_email(message, spam_rating, **kwargs):
             if header.get('Name', '').lower() == 'message-id':
                 msg_id = header.get('Value')
     if not msg_id:
-        log.error('No MessageId in message, aborting.')
-        log.error(message)
```
eviljeff, line 25: the only downside with combing the message content in with the error text is stackdriver has a habit of losing large log messages (try searching for "TextCharSet" in the same timeperiod)

diox: Do we care ? What can we do about this ? (can't find it in stackdriver atm for some reason)

src/olympia/activity/tasks.py
diox: https://github.com/mozilla/addons-server/pull/14380/commits/562ce30d070336b18d10bd9cb426c126633a423d

diox:
For the specific problem of propagating stuff to sentry or not, there is https://github.com/mozilla/addons-server/issues/14355 - if they are aggregated properly this is slightly less urgent.

src/olympia/activity/tasks.py
eviljeff: the reason is stackdriver loses large log messages 😉

eviljeff: we probably don't care because we never look at _anything_ to do with email -> activitylog processing, but just pointing out we lose that message from stackdriver in the meantime.

diox: Ah, lol, I'm dumb :) Are you sure it's stackdriver ? Would that also happen with the message passed as `extra` ?

eviljeff: > Are you sure it's stackdriver ? 

It's somewhere in the pipeline.  I noticed it last year when I was debugging static theme problems - the log entries are in the raw logs but missing in stackdriver (the entry is there but no message content so difficult to identify among the other noise).

> Would that also happen with the message passed as extra ?

no idea tbh.  Lets find out!

Issue #14375, opened 2020-05-25
Add-on still present the mad review queue after manual approval or confirming approval (for unlisted)


AlexandraMoga:
Follow up for https://github.com/mozilla/addons-server/issues/14284#issuecomment-631952721

### Describe the problem and steps to reproduce it:
1. Log in with an admin account or an Admin Reviewer
2. Open the mad review queue - https://reviewers.addons.allizom.org/en-US/reviewers/queue/mad
3. Look for a listed/unlisted add-on that is still _Awaiting Review_ and manually Approve it, or confirm approval for an _unlisted_ version
4. Return to the mad queue and look for the approved add-on

### What happened?
The add-on is still present in the queue both when manually approving a version, or when confirming approval for an _unlisted_ version
### What did you expect to happen?
The add-on should no longer be in the queue

willdurand:
mm looks like I overlooked something else :(

AlexandraMoga:
@willdurand here are two addons that will reproduce the issues described. I've given them a suggestive name, to match the action that is expected:

https://reviewers.addons-dev.allizom.org/en-US/reviewers/queue/mad
![image](https://user-images.githubusercontent.com/31961530/82888969-b588d680-9f52-11ea-8c76-e97964587f2e.png)

AlexandraMoga:
Verified fixed on -stage.
Manual approvals and confirming approval for _unlisted_ versions will remove the entry from the mad queue.

PR #14388, opened 2020-05-26
Clear scanner flags when approving a listed version or multiple unlisted ones

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14375


src/olympia/scanners/migrations/0034_delete_version_flags_again.py
```diff
@@ -0,0 +1,20 @@
...
+
+from django.db import migrations
+
+
+def delete_version_scanner_flags(apps, schema_editor):
+    # We have to delete all the existing entries (again) because we did not
+    # have the logic to clear the flags after a review, so all versions were
+    # kept in the queue. The feature is kinda new so it is fine to delete
+    # existing entries.
+    # See: https://github.com/mozilla/addons-server/issues/14375
```
eviljeff, line 11: you could work it out based on `Version.needs_human_review`?  But if @wagnerand doesn't care then 🤷‍♂️

willdurand: It's fine, it is an experimental queue anyway..

Issue #14396, opened 2020-05-27
Generate stats data when generating fake add-ons


willdurand:
Let's generate some stats data (update/download counts) when we generate new add-ons so that we can see some metrics in the stats pages.

PR #14397, opened 2020-05-27
Generate stats data when generating fake add-ons

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14396

---

With this patch, we can run `./manage.py index_stats` and the stats pages will work locally.

### DIFF
--- a/src/olympia/landfill/tests/test_stats.py
+++ b/src/olympia/landfill/tests/test_stats.py
@@ -20,6 +20,10 @@ class TestGenerateDownloadCounts(TestCase):
         generate_download_counts(self.addon, x_days)
 
         assert DownloadCount.objects.all().count() == x_days
+        download_count = DownloadCount.objects.all()[0]
+        assert not download_count.sources
+        download_count = DownloadCount.objects.all()[1]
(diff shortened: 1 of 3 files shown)


src/olympia/landfill/tests/test_stats.py
```diff
@@ -0,0 +1,39 @@
...
+        self.addon = Addon.objects.create(type=amo.ADDON_EXTENSION)
+
+    def test_generate_update_counts(self):
+        x_days = 10
+
+        generate_update_counts(self.addon, x_days)
+
+        assert UpdateCount.objects.all().count() == x_days
+        update_count = UpdateCount.objects.all()[0]
+        assert 'Darwin' in update_count.oses
```
bobsilverberg, line 38: Just wondering why you decided to test these two things, but not `applications`, nor `sources` in the test for `generate_download_counts`?

willdurand: ha, that's because I added the `sources` after I wrote the test cases.. I added some more assertions. These tests don't really need to be ultra exhaustive so only a few things are checked.

PR #14398, opened 2020-05-27
Notify developers if their submission is pending review after a few hours

### LINKED ISSUE #13974: Send email to developer if add-on submission is pending review after a few hours
If an add-on submission hasn't been signed and is pending review after a few hours, the developer should be sent an email to let them know that the file is still pending review and it could take up to a couple of days to get a resolution.

The text of the email [can be found here](https://docs.google.com/document/d/1msQ-vvDi7GDK04bTp8P5MJgiXoejTrSZ_EOay7HCUA0/edit?ts=5e8d0852#heading=h.tvvrctr03w7t). I suggest sending the email after a 3 hour wait, since I think that should cover for any normal server-side signing delays. Happy to change it if others think there's a better range.

/cc @caitmuenster

diox:
Fixes #13974


src/olympia/reviewers/management/commands/notify_about_auto_approve_delay.py
```diff
@@ -0,0 +1,43 @@
+from datetime import datetime, timedelta
+
+from django.core.management.base import BaseCommand
+
+from olympia.reviewers.utils import ReviewHelper
+from olympia.versions.models import Version
+
+
+# Note: this not done in auto_approve.py on purpose: in case somehow
```
willdurand, line 9:
```suggestion
# Note: this isn't done in auto_approve.py on purpose: in case somehow
```

Issue #4538, opened 2017-01-31
Disable codecov, delete it once and for all


EnTeQuAk:
Codecov is hitting our nerves and it's not being used at the moment anyway. We should but we aren't and we can think about this at a later point once we have resources to take care of this.

Can we delete it? https://codecov.io/gh/mozilla/addons-server Apparently it started commenting on our pull requests again 👎 This needs an admin user unfortunately (or fortunately :D)

muffinresearch:
I've disabled the webhook - feel free to re-open this and let me know if that's not enough.

PR #14404, opened 2020-05-27
Add coverage badge

muffinresearch:
Add back the coverage badge so we can see it and track it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,6 +5,8 @@
 .. image:: https://travis-ci.org/mozilla/addons-server.svg?branch=master
     :target: https://travis-ci.org/mozilla/addons-server
 
+.. image:: https://coveralls.io/repos/github/mozilla/addons-server/badge.svg?branch=master
+    :target: https://coveralls.io/github/mozilla/addons-server?branch=master   
 
 Addons-Server
 =============


diox:
Last I checked it was broken and didn't reflect reality, that's why I pulled it out. We've had some problems in the past finding a service that was a) reliable b) worked well with the way addons-server tests matrix is set up.

muffinresearch:
Ah ok - I wondererd why it was missing. Most of our other services are using codecov rather than coveralls do you know if that would be any better in supporting the current configuration?

It doesn't have to be a super high priority, but would be good to know where coverage is at.

diox:
We tried codecov back in the day (removed in https://github.com/mozilla/addons-server/issues/4538) and the main problem IIRC is it was super noisy for what was essentially rounding errors - like it'd tell us our coverage dropped to 92.576% from 92.577% just because a file was removed, things like that.

On top of that, a problem we've had with coverage services in the past is that we have lots of tests, and our tests are slow and clunky, so they are split between different jobs that run in parallel to be faster. Generating the coverage data sometimes took a significant amount of time per job, and it also has to be reconciled between different jobs, and that also had a significant perf cost sometimes.

Not sure if there has been some improvements that would fix those 2 pain points. We can certainly try to add it back - right now coverage report generation has been completely removed from the test suite.

diox:
Screenshot_2020-05-28 Coveralls io - Test Coverage History and Statistics

None of these files or their coverage make sense. Not only are there only JavaScript files but they don't even match anything in our repos or our dependencies. That makes the coverage numbers totally bogus.

eviljeff:
Closing this because fixing coverage is a prerequisite to adding the badge back.

PR #1471, opened 2016-01-21
stop blacklisting guids & allow developers to re-use their GUIDs bug1200777

eviljeff:

PR #14406, opened 2020-05-27
 Convert ReusedGUID to AddonGUID, containing fks for all Addons 

### LINKED ISSUE #14391: Have a record of the guids and addons stored in the database in one table
Current situation:
- [upload an xpi with guid='foo@baa']
- an addon instance with id=1 and guid = 'foo@baa' exists
- [delete addon]
- [upload a new xpi with guid='foo@baa']
- an addon instance with id=2 and guid = 'foo@baa' exists
- the addon instance id=1 has now guid = None
- there is a ReusedGUID instance that fks to addon_id=1 with guid='foo@baa'.

So if we want to search for all instances of an Addon in Addon.unfiltered with have to do `Addon.unfiltered.filter(Q(guid='foo@baa') | Q(reusedguid__guid='foo@baa'))`.  Similarly to do a values_list/values we have to do `Addon.unfiltered.filter(Q(guid='foo@baa') | Q(reusedguid__guid='foo@baa')).values_list('guid', 'reused__guid')` and process the result. (Even a reference on a single instance of Addon becomes `guid = addon.guid or addon.reusedguid.guid`)

This isn't efficient for database queries and makes code unnecessarily complex. Two options:
1. Add a second `.guid` (`.permanent_guid`?) on Addon.  Straightforward and efficient for queries.  Heavy migration needed to update the Addons table though.
2. Expand and change `ReusedGUID` to contain all Addons, even those not deleted and reused.  Would result in a 1to1 match to Addon.  Queries that needed anything _on_ Addon would still be a join though.  (Including other objects that fk to Addon, like Version)

eviljeff:
fixes #14391


src/olympia/addons/migrations/0005_auto_20200527_0513.py
```diff
@@ -0,0 +1,40 @@
...
+    AddonGUID = apps.get_model('addons', 'AddonGUID')
+    Addon = apps.get_model('addons', 'Addon')
+
+    batch_size = 1000
+    addon_qs = Addon.unfiltered.exclude(guid=None).values('id', 'guid')
+    for addons in chunked(addon_qs, batch_size):
+        addonguids = [
+            AddonGUID(
+                addon_id=addon['id'], guid=addon['guid']) for addon in addons]
+        AddonGUID.objects.bulk_create(addonguids, batch_size)
```
eviljeff, line 17: @diox for second opinion on this migration

src/olympia/reviewers/tests/test_views.py
```diff
@@ -5184,14 +5184,14 @@ def test_reused_guid_from_previous_deleted_addon(self):
...
 
-        old_one = addon_factory(status=amo.STATUS_DELETED)
-        old_two = addon_factory(status=amo.STATUS_DELETED)
-        old_other = addon_factory(status=amo.STATUS_DELETED)
-        old_noguid = addon_factory(status=amo.STATUS_DELETED)
-        ReusedGUID.objects.create(addon=old_one, guid='reuse@')
-        ReusedGUID.objects.create(addon=old_two, guid='reuse@')
-        ReusedGUID.objects.create(addon=old_other, guid='other@')
-        ReusedGUID.objects.create(addon=old_noguid, guid='')
+        old_one = addon_factory(status=amo.STATUS_DELETED, guid=None)
```
bobsilverberg, line 5187: Why is it necessary to set `guid` to `None` now, when it wasn't before? Is this just making the test more explicit?

src/olympia/reviewers/views.py
```diff
@@ -851,8 +851,10 @@ def review(request, addon, channel=None):
     actions_comments = [k for (k, a) in actions if a.get('comments', True)]
 
     deleted_addon_ids = (
-        ReusedGUID.objects.filter(guid=addon.guid).values_list(
-            'addon_id', flat=True) if addon.guid else [])
+        AddonGUID.objects.filter(guid=addon.guid)
+                         .exclude(addon=addon)
```
bobsilverberg, line 855: Do we need this `exclude` now because we don't want the `id` of the current `addon` included in this list? Is there any test coverage for this change? I.e., should there be a test where the `id` for the current `addon` *is* in `AddonGUID`? Maybe that's already covered in the changes to the tests?

src/olympia/reviewers/tests/test_views.py
eviljeff: It's to match the states the addon instances would be in reality.  If they're supposed to be deleted and re-uploaded then the guid would have been reset on the old instance after the new upload.

(If we don't set the guid to `None` then we'd need to find and update the AddonGUID instance for them to point to `guid='reuse@'` instead - it'd be `AddonGUID.objects.get(addon=old_one).update(guid='reuse@')`)

src/olympia/reviewers/views.py
eviljeff: > Do we need this `exclude` now because we don't want the `id` of the current `addon` included in this list? 

Yes.  The current addon wouldn't have been in ReusedGUID before, but _is_ in AddonGUID now.

> Is there any test coverage for this change?

the existing test lines below would fail if we don't exclude the current add-on
```
expected = [
            (f'{old_one.id}', reverse('reviewers.review', args=[old_one.id])),
            (f'{old_two.id}', reverse('reviewers.review', args=[old_two.id])),
        ]
```

src/olympia/addons/migrations/0005_auto_20200527_0513.py
diox: This might be a little too much for a migration... I guess we can test that on dev/stage and see if it takes too long...

In prod, we'll likely want the site in read only to ensure that no new add-ons are created before the new code is up - Otherwise there is a window between the guids being fetched and new code being deployed where add-ons would not have a corresponding `AddonGUID`.

src/olympia/reviewers/tests/test_views.py
eviljeff: ... and the guid hasn't been set to `None` since 2016, when _someone_ changed it in #1471.

eviljeff: Now it's back to not setting `guid` to `None`.

eviljeff:
rebased to fix the conflicts and to adjust things a little after https://github.com/mozilla/addons-server/pull/14423

eviljeff:
going to merge.  If -dev fails to deploy we can reconsider the migration - though if we _don't_ have it in the migration we'll need to disable bloomfilter generation until it's been executed in some other way.

Issue #14384, opened 2020-05-26
have mlbf generation take into account previous usage of the addon guid


eviljeff:
similar problem to #14345

eviljeff:
QA:
two things to test here -
- that an addon that has been uploaded multiple times (uploaded with a fixed guid; deleted; uploaded again with the same guid), and _then_ blocked, has all it's versions blocked in the bloomfilter*, not just the versions from the latest one.
- that an addon that is blocked with a limited min-max range, and then is deleted and reuploaded with the same guid, still has the versions from the earlier upload blocked in the bloomfilter*, not just the versions from the latest one.

*inspecting the bloomfilter is hard, but the stash content will be in https://kinto.stage.mozaws.net/v1/buckets/blocklists/collections/addons-bloomfilters/records.

AlexandraMoga:
@eviljeff I've verified the stash records for stage and these are the results:

Not in stash:
[second-reused@guid](https://addons-internal.stage.mozaws.net/en-US/admin/models/blocklist/block/246/change/) - blocked for versions `0 - 1.1` -> add-on is deleted and resubmitted again with the same guid and versions `1.2` and `1.3` -> block `max` version is invalid in AMO admin
[third-reused@guid](https://addons-internal.stage.mozaws.net/en-US/admin/models/blocklist/block/247/change/) - blocked for versions `1.2 - *` -> add-on is deleted and resubmitted again with the same guid and version `1.0` -> block `min` version is invalid in AMO admin
[fourth-reused@guid](https://addons-internal.stage.mozaws.net/en-US/admin/models/blocklist/block/248/change/) - similar to the first scenario

In stash, but...:
[sixth-reused@guid](https://addons-internal.stage.mozaws.net/en-US/admin/models/blocklist/block/250/change/) - first add-on with this guid had versions `1.0` and `1.1` -> add-on was _not_ blocked -> add-on was deleted and resubmitted with same guid and version `1.0` -> add-on is fully blocked in AMO admin -> only version `1.0` appears in stash, version `1.1` from the previous add-on does not

For the first set of add-ons (not in stash), do we also depend on #14345 to be fixed or should this have worked without that?

eviljeff:
The admin hasn't been fixed/touched with this patch - that's #14345 - so what's in AMO admin can't be relied on to verify the issue.

But the versions you've mentioned aren't in the stash, _should_ be in the stash so 😞

AlexandraMoga:
New results:
[...]

PR #14423, opened 2020-05-28
Addon.guid is not None for reused guids!

eviljeff:
fix for #14384


eviljeff:
If only a young(er) naive(r) developer hadn't changed the logic in #1471 so Addon.guid wasn't set to None any more.

Issue #14412, opened 2020-05-28
Use BigQuery for the "user" (updates) stats in beta mode


willdurand:
Once we have #14402, we'll be able to use BigQuery. This issue is about using BigQuery for the "user" stats:

<img width="238" alt="Screen Shot 2020-05-28 at 09 26 25" src="https://user-images.githubusercontent.com/217628/83111731-539cae00-a0c5-11ea-967d-736df9b24e11.png">

willdurand:
For QA: please reach out to me via PM for more information about how to verify this issue. I have some instructions that shouldn't be shared publicly.

ioanarusiczki:
@willdurand Thanks for the STR.
It worked for me.
Example: 
https://addons-dev.allizom.org/en-US/firefox/addon/wollinosstest/statistics/beta/usage/?last=30

PR #14425, opened 2020-05-28
Use BigQuery for the "user" (updates) stats in beta mode

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14412

~~:warning: depends on https://github.com/mozilla/addons-server/pull/14403~~

---

This patch implements a new data source (BigQuery) but only for the "beta" stats. It is possible to test it locally but you need to:

1. fetch the credentials (a json file) in -dev (using -ssh, path is in `settings.GOOGLE_APPLICATION_CREDENTIALS`)
2. configure `GOOGLE_APPLICATION_CREDENTIALS` in `local_settings.py`
3. update a local add-on with a known GUID like the one mentioned [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1633746#c0)

To access the beta, append `beta/` to the URL of the main (overview) stats page and make sure your local account as a `@mozilla.com` email address.

Other notes:

- the fact that we call BigQuery directly was decided before and it looks like a good idea. It works quite well but we'll see in production. We might want to add some caching layer later, though.
- all the rendering logic has been kept so that's why we do not have lots of tests for the views. We'll tackle that when/if we'll change the stats pages.

### DIFF
--- a/src/olympia/stats/tests/test_views.py
+++ b/src/olympia/stats/tests/test_views.py
@@ -2,6 +2,9 @@
 import csv
 import json
 
+from datetime import date
+from unittest import mock
+
 from django.http import Http404
 from django.test.client import RequestFactory
 from django.urls.exceptions import NoReverseMatch
@@ -703,6 +706,15 @@ class TestStatsBeta(TestCase):
 
         self.user = user_factory(email='staff@mozilla.com')
         self.addon = addon_factory(users=[self.user])
+        self.start_date = date(2020, 1, 1)
+        self.end_date = date(2020, 1, 5)
+        self.series_args = [
+            self.addon.slug,
+            'day',
+            self.start_date.strftime('%Y%m%d'),
+            self.end_date.strftime('%Y%m%d'),
+            'json'
+        ]
         self.client.login(email=self.user.email)
 
     def test_stats_overview_page(self):
@@ -714,10 +726,7 @@ class TestStatsBeta(TestCase):
         assert response.context['beta']
 
     def test_beta_series_urls(self):
-        url = reverse(
-            'stats.overview_series.beta',
-            args=[self.addon.slug, 'day', '20200101', '20200105', 'json']
-        )
+        url = reverse('stats.overview_series.beta', args=self.series_args)
 
         match = resolve(url)
 
@@ -734,7 +743,7 @@ class TestStatsBeta(TestCase):
             reverse('stats.statuses.beta', args=[self.addon.slug])
 
         with self.assertRaises(NoReverseMatch):
-            reverse('tats.statuses_series.beta', args=[self.addon.slug])
+            reverse('stats.statuses_series.beta', args=self.series_args)
 
     def test_no_beta_for_non_staff(self):
         self.client.logout()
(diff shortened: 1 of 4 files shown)


src/olympia/stats/tests/test_views.py
```diff
@@ -734,7 +743,7 @@ def test_no_status_page(self):
             reverse('stats.statuses.beta', args=[self.addon.slug])
 
         with self.assertRaises(NoReverseMatch):
-            reverse('tats.statuses_series.beta', args=[self.addon.slug])
+            reverse('stats.statuses_series.beta', args=self.series_args)
```
willdurand, line 746: Bad rebase in a previous patch. The test case was still passing because we expect no route to be reversed..

src/olympia/stats/utils.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+
+    filter_by = AMO_TO_BIGQUERY_COLUMN_MAPPING[source] if source else None
+
+    select_clause = 'SELECT submission_date, dau'
+    if filter_by:
+        select_clause = f'{select_clause}, {filter_by}'
+    query = f"""
+{select_clause}
+FROM `moz-fx-data-shared-prod.telemetry.amo_stats_dau`
```
willdurand, line 46: This table name might change but it's not clear how or when. This is the production table, though so that will work in prod.

src/olympia/stats/views.py
```diff
@@ -134,7 +134,14 @@ def overview_series(request, addon, group, start, end, format, beta=False):
...
 
     dls = get_series(DownloadCount, addon=addon.id, date__range=date_range)
-    updates = get_series(UpdateCount, addon=addon.id, date__range=date_range)
+
+    if beta:
+        updates = get_updates_series(addon=addon,
+                                     start_date=date_range[0],
+                                     end_date=date_range[1])
+    else:
+        updates = get_series(UpdateCount, addon=addon.id,
```
willdurand, line 143: I decided not to change things in `get_series` because it is also used for download stats, and that's already messy enough.

src/olympia/stats/utils.py
```diff
@@ -0,0 +1,69 @@
...
+    'locales': 'dau_by_locale',
+    'os': 'dau_by_app_os',
+    'versions': 'dau_by_addon_version',
+}
+
+
+def rows_to_series(rows, filter_by=None):
+    """Transforms BigQuery rows into series items suitable for the rest of the
+    AMO stats logic."""
+    for row in rows:
```
eviljeff, line 18: what's the data structure of `row`?  Seems like it's both an object (`row.dau`) and a dict (`row[filter_by]`).

src/olympia/stats/views.py
```diff
@@ -283,8 +303,8 @@ def process_locales(series):
         if 'data' in row:
             new = {}
             for key, count in row['data'].items():
-                if key in languages:
-                    k = u'%s (%s)' % (languages[key], key)
+                if key and key.lower() in languages:
+                    k = u'%s (%s)' % (languages[key.lower()], key)
```
eviljeff, line 307: nit:
```suggestion
                    k = '%s (%s)' % (languages[key.lower()], key)
```

src/olympia/stats/utils.py
willdurand: it is a class (`Row`) which can be accessed using both notations (attributes and dict-style)

Issue #14417, opened 2020-05-28
Remove call to `events-xxxx.json` in stats pages


willdurand:
It looks like we call an endpoint ending with `events-xxxx.json` in the stats pages but there is no route for it, so we redirect to `/`. This seems unnecessary so let's remove it.

AlexandraMoga:
There is no call to `events-xxxx.json` in the stats page on -dev anymore (I've compared with stage where this call is present).
I'm not marking this issue verified yet, because the statistics page is currently broken on -dev and the root cause hasn't been identified so far.

willdurand:
@AlexandraMoga please file a new issue for this.

willdurand:
> @AlexandraMoga please file a new issue for this.

nevermind, I filed https://github.com/mozilla/addons-server/issues/14444

ioanarusiczki:
@willdurand I did not see any `events-xxxx.json` call on dev while loading stats for https://addons-dev.allizom.org/en-US/firefox/addon/view-page-archive-cache/statistics/?last=365 or https://addons-dev.allizom.org/en-US/firefox/addon/terms-of-service-didn-t-read/statistics/?last=30.

Verified on AMO dev with FF76(Win10).

PR #14426, opened 2020-05-28
Remove call to non-existing events-xxxx.json endpoint

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14417


willdurand:
> I guess there was no test for this?

Answer is no :D

Issue #14424, opened 2020-05-28
Add a check to restrict the access to the beta stats


willdurand:
Let's restrict the access to the stats beta mode to specific users first, like all users with a `mozilla.com` email address first.

ioanarusiczki:
@willdurand I verified this on AMO dev with FF76(Win10).
- an account without `mozilla.com` doesn't have access to any beta stats (even if it's an admin) public or private.
- a `mozilla.com` account has access to beta pages without being an admin to public pages. The condition to see the beta private stats is to be an admin. 

Let me know if that's the expected for now. 
Thanks!

willdurand:
Yes, I think it makes sense. In addition to that, a developer should be able to access their own stats in beta if and only if they have a `mozilla.com` email address.

ioanarusiczki:
@willdurand Yes, I tested with the developer scenario for a `mozilla.com` address.  Beta stats are visible.

PR #14429, opened 2020-05-28
Add a check to restrict the access to the beta stats

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14424

:warning: depends on #14403

---

~~The `User.is_staff` has been discussed with @diox.~~ switched to a homemade solution because `is_staff` is not working as I wished.

### DIFF
--- a/src/olympia/stats/views.py
+++ b/src/olympia/stats/views.py
@@ -131,7 +131,7 @@ def extract(dicts):
 def overview_series(request, addon, group, start, end, format, beta=False):
     """Combines downloads_series and updates_series into one payload."""
     date_range = check_series_params_or_404(group, start, end, format)
-    check_stats_permission(request, addon)
+    check_stats_permission(request, addon, beta)
 
     dls = get_series(DownloadCount, addon=addon.id, date__range=date_range)
     updates = get_series(UpdateCount, addon=addon.id, date__range=date_range)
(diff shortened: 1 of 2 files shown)


src/olympia/users/models.py
```diff
@@ -208,8 +208,7 @@ def is_staff(self):
         the django admin tools. Does not guarantee that the user will then
         be able to do anything, as each module can have its own permission
         checks. (see has_module_perms() and has_perm())"""
```
diox, line 210: We should reword that slightly. I suggest (doesn't let me do that as a github suggestion):
```
        """Property indicating whether the user is a Mozilla Corp employee or
        not. Django uses it to allow access to admin tools, but it does not
        guarantee that they would then be able to do anything as each module
        can have its own permission checks. (see has_module_perms() and
        has_perm()). On top of this, our admin tools should only be
        available over VPN for users part of a specific VPN group."""

(Pretty verbose, feel free to adjust)

Can we also double-check with @wagnerand that it's ok to restrict the admin to users with a `@mozilla.com` email ?

Comment (willdurand) [reply src/olympia/users/models.py]: ha well, that broke lots of tests :(

File: src/olympia/stats/views.py [hunk 2]
```diff
@@ -297,12 +297,21 @@ def check_series_params_or_404(group, start, end, format):
     return get_daterange_or_404(start, end)
 
 
-def check_stats_permission(request, addon):
+def check_stats_permission(request, addon, beta):
```
diox, line 300: Personal preference: I don't like booleans as positional arguments - If someone ends up calling the function with just `True` or `False` passed as the value, like in `foo(True)`, it makes it difficult to understand what the boolean represents at first glance. So I always try to remember to set them as keyword (or even keyword-only) arguments. 

It matters slightly less in this instance as you did directly pass the `beta` variable in every call, which is explicit enough, and we're unlikely to have many more calls to that function, so you can ignore that comment for this PR.

PR #14442, opened 2020-05-29
Make the code manager the default link for reviewer tools

### LINKED ISSUE #13634: Make the code manager the default link for reviewer tools
To get reviewers using the new code manager more we should make the default link point to the new code manager. 

For the escape hatch we have a couple of options. Either keep the old code manager link (e.g. swap the two) or alternatively we just link to the new code manager and provide the link to the old reviewer tools from the code-manager ui. 

The former is easier, but the latter would prevent reviewers learning to just clink the link to the old tool. 

/cc @jvillalobos for thoughts.

bobsilverberg:
Fixes #13634 

I removed all the logic for the `code-manager` waffle flag, as I don't think we need it anymore, but please correct me if I'm wrong about that. Do we need to do anything to get rid of the flag from waffle as well?


wagnerand:
Could you also update the links in the admin tools pointing to the file-viewer? I believe it's mostly scanner results.

src/olympia/devhub/views.py
```diff
@@ -661,7 +662,7 @@ def file_validation(request, addon_id, addon, file_id):
 
     validate_url = reverse('devhub.json_file_validation',
                            args=[addon.slug, file_.id])
-    file_url = reverse('files.list', args=[file_.id, 'file', ''])
+    file_url = code_manager_url('browse', addon.pk, file_.version.pk)
```
bobsilverberg, line 665: It looks like this change should be made, but I'm not sure what effect it has on the site or where to test it. After making this change no tests broke, so I assume this line is not currently covered by tests.

willdurand:
> Do we need to do anything to get rid of the flag from waffle as well?

Maybe a db migration to remove it fully?

src/olympia/reviewers/templates/reviewers/includes/files_view.html
```diff
@@ -11,38 +11,26 @@
     </div>
     <a href="{{ url('devhub.file_validation', addon.slug, file[0].id) }}">{{ _('Validation') }}</a>
     &middot;
-    <a href="{{ url('files.list', file[0].id) }}">{{ _('Contents') }}</a>
+    <a 
+      href="{{ code_manager_url('browse', addon.pk, version.pk) }}"
+      title="Browse all files in this version"
```
willdurand, line 16: We likely want to localize this string so it should probably be something like:

```suggestion
      title="{{ _('Browse all files in this version') }}"
```

bobsilverberg:
> > Do we need to do anything to get rid of the flag from waffle as well?
> 
> Maybe a db migration to remove it fully?

I added that in a commit.

src/olympia/reviewers/templates/reviewers/includes/files_view.html
```diff
@@ -11,38 +11,26 @@
...
+      >
+        {{ _('Contents') }}
+    </a>
     {% if show_diff and version == latest_version_in_channel %}
     &middot;
-    <a class="compare" href="{{ url('files.compare', file[0].id, file_compare(file[0], show_diff)) }}">{{ _('Compare') }}</a>
+    <a
+      class="compare"
+      href="{{ code_manager_url('compare', addon.pk, version.pk, show_diff.pk) }}"
+      title="Compare this version to the last reviewed version"
```
willdurand, line 25: Same here. I know it wasn't like that before but all the other strings are localized.

willdurand: `show_diff.pk`? 🤔

src/olympia/devhub/views.py
willdurand: so maybe we shouldn't touch that in this patch?

src/olympia/reviewers/migrations/0005_remove_code_manager_waffle_switch.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.12 on 2020-06-02 13:28
+
+from django.db import migrations
+
+
+def delete_waffle_switch(apps, schema_editor):
+    Switch = apps.get_model('waffle', 'Switch')
```
willdurand, line 7: Wasn't it a `Flag`? If so, then use the `Flag` model instead of the `Switch` one. Waffle has different types: https://waffle.readthedocs.io/en/stable/types/index.html

src/olympia/reviewers/templatetags/code_manager_tags.py
```diff
@@ -0,0 +1,19 @@
...
+from django.conf import settings
+
+register = template.Library()
+
+
+@register.simple_tag
+def code_manager_url(
+        page, addon_id, version_id, compare_version_id=None, file=None):
+    # Always return URLs in en-US because the Code Manager is not localized.
+    url = '{}/en-US'.format(settings.CODE_MANAGER_URL)
```
willdurand, line 11: Can't we have locale-free URLs?

src/olympia/reviewers/templatetags/code_manager_tags.py
```diff
@@ -0,0 +1,19 @@
...
+    if page == 'browse':
+        url = '{}/browse/{}/versions/{}/'.format(url, addon_id, version_id)
+    else:
+        url = '{}/compare/{}/versions/{}...{}/'.format(
+            url, addon_id, compare_version_id, version_id)
+    if file is not None:
```
willdurand, line 17: Isn't it enough to write `if file:`?

src/olympia/reviewers/templatetags/code_manager_tags.py
```diff
@@ -0,0 +1,19 @@
+from django import template
+from django.conf import settings
+
+register = template.Library()
+
+
+@register.simple_tag
+def code_manager_url(
+        page, addon_id, version_id, compare_version_id=None, file=None):
```
willdurand, line 9: Is `compare_version_id` the `base_version_id`? We use that in code-manager so would be nice to keep the same name.

src/olympia/reviewers/tests/test_code_manager_tags.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
```
willdurand, line 1: I'm not sure we need this comment anymore.

src/olympia/reviewers/tests/test_code_manager_tags.py
```diff
@@ -0,0 +1,56 @@
...
+from django.test.utils import override_settings
+
+from olympia.reviewers.templatetags import code_manager_tags
+from olympia.amo.tests import TestCase
+
+
+class TestCodeManagerUrl(TestCase):
+
+    def setUp(self):
+        self.cm_url = 'http://code-manager'
```
willdurand, line 11: We need to call `super().setUp()` here (before anything else)

src/olympia/reviewers/tests/test_code_manager_tags.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_create_a_browse_url(self):
+        addon_id = 1
```
willdurand, line 14: You probably want to add those in the `setUp` method and use `self.addon_id`, etc.

src/olympia/reviewers/tests/test_jinja_helpers.py
```diff
@@ -79,12 +78,6 @@ def test_file_review_status_handles_invalid_status_id():
...
-            'http://code-manager/en-US/some/path'
-        )
-
-
-def test_code_manager_url_expects_a_relative_path():
-    with pytest.raises(ValueError):
-        jinja_helpers.code_manager_url('http://code-manager/some/path')
+    assert jinja_helpers.code_manager_url(
+        'browse', 1, 2, 3) == code_manager_tags.code_manager_url(
+            'browse', 1, 2, 3)
```
willdurand, line 83: Can we use named parameters?

src/olympia/reviewers/tests/test_views.py
```diff
@@ -4327,7 +4329,8 @@ def test_compare_link_auto_approved_ignored(self):
         # ignoring the interim version because it was auto-approved and not
         # manually confirmed by a human.
         expected = [
-            reverse('files.compare', args=[new_file.pk, first_file.pk]),
+            code_manager_url('compare', self.addon.pk,
+                             new_version.pk, first_version_pk),
```
willdurand, line 4333: would be nice to have named parameters here too

src/olympia/reviewers/templatetags/jinja_helpers.py
```diff
@@ -221,13 +222,10 @@ def is_expired_lock(context, lock):
...
-    if not path.startswith('/'):
-        raise ValueError(
-            'Expected a relative path; got: "{}"'.format(path)
-        )
-    # Always return URLs in en-US because the Code Manager is not localized.
-    return '{}/en-US{}'.format(settings.CODE_MANAGER_URL, path)
+def code_manager_url(
+        page, addon_id, version_id, compare_version_id=None):
+    return code_manager_tags.code_manager_url(
+        page, addon_id, version_id, compare_version_id)
```
willdurand, line 228: So, no file in this case?

src/olympia/reviewers/templatetags/code_manager_tags.py
```diff
@@ -0,0 +1,19 @@
+from django import template
```
willdurand, line 1: I don't think we need the `_tags` suffix on the filename, the file lives inside `templatetags/` already.

src/olympia/reviewers/templates/reviewers/includes/files_view.html
bobsilverberg: Yeah, not a great name. I just left it as it was, but it's only used in this code so I will change it to `base_version`.

src/olympia/reviewers/templatetags/code_manager_tags.py
bobsilverberg: It looks like we cannot. If I try to go to a code manager page without the `/en-US/` I get an "Ooops, page not found!" page. That's not ideal, but also not something I think is worth working on at this point. TBH I'm not sure why we dealt with locales at all for code manager. I guess the theory was that at some point it could possibly be localizable.

src/olympia/reviewers/templatetags/jinja_helpers.py
bobsilverberg: Correct. This helper is only used on one page, and that doesn't provide links to specific files, so it doesn't need the `file` feature.

bobsilverberg:
Thanks for the review, @willdurand. I have addressed your comments and pushed a new commit.

src/olympia/reviewers/templates/reviewers/includes/files_view.html
```diff
@@ -11,38 +11,26 @@
...
     {% if file[0].is_webextension %}
     <div class="file-permissions"><strong>{{ _('Permissions:') }}</strong> {{ ', '.join(file[0].webext_permissions_list) or _('None') }}</div>
     {% endif %}
   </span>
 </li>
-{% if waffle.flag('code-manager') %}
-  <li class="code-manager-links">
-    <div>
-      <strong>Code Manager</strong>
-      <img width="25" height="9" src="{{ static('img/reviewers/new.gif') }}">
```
willdurand, line 28: I think we can delete the image in `img/reviewers/new.gif` now

src/olympia/reviewers/templatetags/code_manager.py
```diff
@@ -0,0 +1,19 @@
+from django import template
+from django.conf import settings
+
+register = template.Library()
+
+
+@register.simple_tag
+def code_manager_url(
```
willdurand, line 8: Maybe run Black on this file, it will fix the arguments.

bobsilverberg: I'm not sure what you're looking for. Running `black` on this file leaves everything as is, except for the function signature, and then it replaces all the single quotes with double quotes, which I don't think we want. Also, by placing all the function arguments on a single line it breaks a `flake8` rule (the line is too long).

What exactly is the problem you think needs fixing?

willdurand: ha, you don't have Black configured for this project then. Black would put the closing parent on its own line like this:

```
def code_manager_url(
    a, r, g, s
):

Comment (willdurand) [reply src/olympia/reviewers/templatetags/code_manager.py]: but I like the update code too, which is:

```
def code_manager_url(a, r, g,
                     s):
```

bobsilverberg: Yeah, I don't have `black` configured because when I tried it awhile back it seemed to want to make too many unwanted changes. Do you have a config that makes it play nicely with addons-server?

willdurand: Yeah. I use this vim config: https://github.com/willdurand/dotfiles/blob/71c0e7c222f624bc5332092a43e169b3a9f0950e/vim/vimrc#L323-L325

bobsilverberg: Thanks. I went and checked my VSCode settings and I already had it configured to do that. I ran a test and found that it made a bunch of changes to parentheses. The resulting format is fine, but it would mean a lot of churn in a large file where only a small amount of code has changed, so I'm not sure I should use it all the time.

PR #14443, opened 2020-05-29
We still need to call the data endpoint

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14444


willdurand:
Merging because we don't have any tests for this part of the codebase.

PR #14446, opened 2020-05-29
add Block.average_daily_users property

### LINKED ISSUE #14407: Block refactoring - add .active_daily_user and drop FakeAddon
splitting out some of the work from #14391

eviljeff:
fixes #14407

### DIFF
--- a/src/olympia/blocklist/models.py
+++ b/src/olympia/blocklist/models.py
@@ -75,6 +75,11 @@ class Block(ModelBase):
         return Addon.unfiltered.filter(
             guid=self.guid).only_translations().first()
 
+    @property
+    def average_daily_users(self):
+        addon = self.addon
+        return addon.average_daily_users if addon else 0
+
     @cached_property
     def addon_versions(self):
         # preload_addon_versions will overwrite this cached_property.
@@ -175,7 +180,6 @@ class Block(ModelBase):
             block.guid: block
             for block in cls.objects.using(using_db).filter(guid__in=guids)}
 
-        blocks = []
         for addon in addons:
             # get the existing block object or create a new instance
             block = existing_blocks.get(addon.guid, None)
@@ -184,12 +188,9 @@ class Block(ModelBase):
                 block.addon = addon
             else:
                 # otherwise create a new Block
-                block = cls(addon=addon)
-            blocks.append(block)
-        blocks.extend(
-            block for guid, block in existing_blocks.items()
-            if block not in blocks)
-        return blocks
+                block = Block(addon=addon)
+                existing_blocks[block.guid] = block
+        return list(existing_blocks.values())
 
 
 class BlocklistSubmission(ModelBase):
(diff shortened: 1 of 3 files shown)


src/olympia/blocklist/models.py
```diff
@@ -361,9 +362,11 @@ def save(self, *args, **kwargs):
     @classmethod
     def _split_guids_full(cls, guids, v_min, v_max, filter_existing):
         # load all the Addon instances together
-        addon_qs = Addon.unfiltered.filter(guid__in=guids).order_by(
-            '-average_daily_users').only_translations()
-        addons = list(addon_qs)
+        addons_qs = Addon.unfiltered.filter(guid__in=guids).only_translations()
+        addons = sorted(
```
bobsilverberg, line 366: How are these new lines different from what the old lines were doing?

src/olympia/blocklist/models.py
```diff
@@ -409,14 +412,15 @@ def _split_guids_full(cls, guids, v_min, v_max, filter_existing):
...
-            cls.FakeAddon(guid, addon_users) for guid, addon_users in addon_qs]
-        addon_guid_dict = {addon.guid: addon for addon in addons}
+        addons_qs = Addon.unfiltered.filter(guid__in=guids).values_list(
+            'guid', 'average_daily_users', named=True)
+
+        addons = sorted(
+            addons_qs,
+            key=lambda addon: addon.average_daily_users,
+            reverse=True)
+        adu_lookup = {
```
bobsilverberg, line 422: Same question as above. How is this new code different from what was there before?

src/olympia/blocklist/templates/admin/blocklist/includes/enhanced_blocks.html
```diff
@@ -5,10 +5,8 @@ <h3>{% blocktrans with count=blocks|length %}{{ count }} Add-on GUIDs{% endblock
         {% for block_obj in blocks %}
           <li>
             {{ block_obj.guid }}.
-            {% if block_obj.addon %}
-              <span class="addon-name">{{ block_obj.addon.name }}</span>
-              ({{ block_obj.addon.average_daily_users }} {% trans 'users' %}).
-            {% endif %}
+            <span class="addon-name">{{ block_obj.addon.name }}</span>
```
bobsilverberg, line 8: Why did we need an `if` statement to check for `block_obj.addon` before, but no longer need it? We're still accessing `block_obj.addon.name`.

src/olympia/blocklist/models.py
eviljeff: they're the same - the follow on patch does it this way and I brought it forward (or left it in, depending on how you think about it)

src/olympia/blocklist/models.py
eviljeff: same as above.  The `adu_lookup` is a change though

src/olympia/blocklist/templates/admin/blocklist/includes/enhanced_blocks.html
eviljeff: `block_obj.addon` won't exist now when the `block_obj` is a `FakeBlock` (i.e. when a large number of blocks are loaded).  Template parsing just ends up resolving missing attributes to `''` so it was in an `if` so it didn't look like ` users` when addon didn't exist.

eviljeff:
> I am also wondering, why are there no changes to tests? Is this just a 100% refactoring task? Are there no tests that care that the shape of FakeBlock has changed?

Yeah, it's all refactoring so make the followup patch less noisy.  That no tests need changing is a good thing :)

PR #14453, opened 2020-06-01
Block.addon_versions to be a list of Version

### LINKED ISSUE #14451: Block refactoring - addon_versions to be full versions
splitting out some of the work from #14391

`Block.addon_versions` is currently a dict of version string: tuple of some extra properties which makes it hard to maintain when the metadata is accessed in different places outside of the model and pretty unnecessary (it's only used for cases when a relatively small number of guids are being blocked.)  Lets just have it be a list of `Version` instances directly.

eviljeff:
fixes #14451

### DIFF
--- a/src/olympia/blocklist/models.py
+++ b/src/olympia/blocklist/models.py
@@ -1,10 +1,11 @@
-from collections import defaultdict, namedtuple, OrderedDict
+from collections import defaultdict, namedtuple
 
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.html import format_html
 from django.utils.functional import cached_property
+from django.utils.translation import gettext_lazy as _
 
 import waffle
 from django_extensions.db.fields.json import JSONField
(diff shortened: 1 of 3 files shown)


src/olympia/blocklist/models.py
```diff
@@ -83,20 +83,19 @@ def addon_versions(self):
...
-        qs = Version.unfiltered.filter(addon_id__in=addon_ids).order_by(
-            'id').values('addon_id', 'version', 'id', 'channel')
-        addons_versions = defaultdict(OrderedDict)
+        """Preload block.addon_versions into a list of blocks."""
+        block_guids = [block.guid for block in blocks]
+        GUID = 'addon__guid'
+        qs = (
+            Version.unfiltered.filter(**{f'{GUID}__in': block_guids})
+                              .order_by('id')
+                              .annotate(**{GUID: models.F(GUID)}))
```
diox, line 92: Do we want the default transformers to be called ? That's a few extra queries to fetch translations, files, applicationversions, and then build the related model instances.

eviljeff: https://github.com/mozilla/addons-server/pull/14453/commits/2407bf7cbd0bb62f6a0335f02e5b9aa676c39ffd

PR #14464, opened 2020-06-02
Move VersionScannerFlags to VersionReviewerFlags, adding pending_rejection field

### LINKED ISSUE #14462: Add model to track pending version rejection
For https://github.com/mozilla/addons-server/issues/14454 we need a new field to track when a version is pending rejection. Since we have `VersionScannerFlags` already, let's move it to the `versions` app under `VersionReviewerFlags`, and add a `pending_rejection` field to it.

diox:
Fixes #14462


src/olympia/versions/models.py
```diff
@@ -693,6 +693,18 @@ def can_be_disabled_and_deleted(self):
         return True
 
 
+class VersionReviewerFlags(ModelBase):
+    version = models.OneToOneField(
+        Version, primary_key=True, on_delete=models.CASCADE
+    )
+    needs_human_review_by_mad = models.BooleanField(default=False)
+    pending_rejection = models.DateTimeField(default=None, null=True)
+
```
eviljeff, line 702: We'll lose any existing flags in `VersionScannerFlags` - should we either be setting `Meta.db_table` to point to the old table or have a python migration that copies them over?

src/olympia/versions/models.py
```diff
@@ -693,6 +693,18 @@ def can_be_disabled_and_deleted(self):
         return True
 
 
+class VersionReviewerFlags(ModelBase):
+    version = models.OneToOneField(
+        Version, primary_key=True, on_delete=models.CASCADE
+    )
+    needs_human_review_by_mad = models.BooleanField(default=False)
+    pending_rejection = models.DateTimeField(default=None, null=True)
```
eviljeff, line 701: Nothing is actually using this property currently - that's expected?

src/olympia/versions/models.py
```diff
@@ -693,6 +693,18 @@ def can_be_disabled_and_deleted(self):
...
+
+    @property
+    def needs_human_review(self):
```
eviljeff, line 704: not sure I understand the need for the property if we're recreating the entire model - can we just rename the field to `needs_human_review`?

src/olympia/scanners/models.py
```diff
@@ -417,14 +417,3 @@ class Meta(AbstractScannerResult.Meta):
 class ScannerQueryMatch(ModelBase):
     result = models.ForeignKey(ScannerQueryResult, on_delete=models.CASCADE)
     rule = models.ForeignKey(ScannerQueryRule, on_delete=models.CASCADE)
-
-
-class VersionScannerFlags(ModelBase):
```
eviljeff, line 422: followup issue for the migration to drop the model/table plz.

src/olympia/versions/models.py
diox: @willdurand said it's ok to lose the data

src/olympia/versions/models.py
diox: Yes, just adding the model to unblock other work that will take more time to finish.

src/olympia/versions/models.py
diox: I'm not sure we really need it, I'll look into it.

willdurand: The property is likely not used yet but we don't want to rename the field because it needs to be specific. And we'll want to move the `needs_human_review` flag from `Version` to this model instead in the future.

eviljeff: why does it need to be specific?  And if it does, why not just name the field that.  What's the use case for having a field available under two names?

willdurand: I think we can get rid of the property for now, assuming it isnt used (but I think it is not used).

My idea was to use a property to answer the question "does this version need human review?", which can be the case for different reasons, each reason represented by a different field. In some cases, we only want to know whether a version needs human review but in other cases, we want to know if it is caused by mad or another scanner or something else. We're not there yet, so we can remove the property I think

src/olympia/scanners/models.py
diox: https://github.com/mozilla/addons-server/issues/14499

src/olympia/versions/models.py
diox: a8c0d81e654e4bc65766b161ffd2cc9de90d397d

PR #14466, opened 2020-06-02
Handle past instances of an addon reusing the same guid in the admin

### LINKED ISSUE #14345: Version range can end up 'invalid' for blocked GUIDs belonging to deleted then resubmitted add-ons
1. As an admin, block an add-on for a fixed version range - i.e `min_version` = `0` , `max_version` = `1.1`
2. As the developer of the blocked add-on, delete the add-on from devhub
3. Submit a new add-on with the same guid (which is already blocked) and set your version string in the manifest to a value that is not covered by the existing block - i.e version can be `1.2`
4. As an admin, verify the version range of the exiting block for this particular guid 

### What happened?
The `max_version` will be `(invalid)` - see [example](https://addons-internal.dev.mozaws.net/en-US/admin/models/blocklist/block/26197/change/)

We should avoid ending up with `(invalid)` versions (I guess)

proposed in https://github.com/mozilla/addons-server/issues/14029#issuecomment-631499637
>I believe a better follow-up would be to prevent resubmission of a blocked add-ons after it has been deleted. We can do that by adding it to DeniedGuid.

eviljeff:
fixes #14345 (I split out the refactoring into other prs/issues so there's not much change here)

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -1547,14 +1547,16 @@ class Addon(OnChangeMixin, ModelBase):
         from olympia.blocklist.models import Block
 
         # Block.guid is unique so it's either on the list or not.
-        return Block.objects.filter(guid=self.guid).last()
+        guid = getattr(self, 'addonguid', self).guid
+        return Block.objects.filter(guid=guid).last()
 
     @cached_property
     def blocklistsubmission(self):
         from olympia.blocklist.models import BlocklistSubmission
 
         # GUIDs should only exist in one (active) submission at once.
-        return BlocklistSubmission.get_submissions_from_guid(self.guid).last()
+        guid = getattr(self, 'addonguid', self).guid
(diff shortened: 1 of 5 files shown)


src/olympia/addons/models.py
```diff
@@ -1547,14 +1547,16 @@ def block(self):
         from olympia.blocklist.models import Block
 
         # Block.guid is unique so it's either on the list or not.
-        return Block.objects.filter(guid=self.guid).last()
+        guid = getattr(self, 'addonguid', self).guid
```
bobsilverberg, line 1550: Would it be worth having a helper for this, as it's used twice?

eviljeff: mmm, yes.  I did have one at one point along the way, I guess I must have refactored it away.  There's another follow-up refactoring so I'll look to add it there.

PR #14469, opened 2020-06-02
Some refactoring of Block & BlocklistSubmission

### LINKED ISSUE #14468: Refactor Block code
There's a fair bit of duplicate functionality in how guids are turned into Block objects that I've been putting off landing in other patches to avoid unnecessary review noise.

eviljeff:
fixes #14468

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -1542,21 +1542,24 @@ class Addon(OnChangeMixin, ModelBase):
                 lookup_field = 'slug'
         return lookup_field
 
+    @property
+    def addonguid__guid(self):
+        return getattr(self, 'addonguid', self).guid
+
     @cached_property
     def block(self):
         from olympia.blocklist.models import Block
 
         # Block.guid is unique so it's either on the list or not.
-        guid = getattr(self, 'addonguid', self).guid
-        return Block.objects.filter(guid=guid).last()
+        return Block.objects.filter(guid=self.addonguid__guid).last()
 
     @cached_property
     def blocklistsubmission(self):
         from olympia.blocklist.models import BlocklistSubmission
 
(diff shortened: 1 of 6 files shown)


src/olympia/addons/models.py
```diff
@@ -1542,21 +1542,24 @@ def get_lookup_field(cls, identifier):
                 lookup_field = 'slug'
         return lookup_field
 
+    @property
+    def addonguid__guid(self):
```
diox, line 1546: I get the idea being the name, but I find it confusing when used in a queryset like you do below...

Why do we need this anyway? Don't we have a guarantee nowadays than `self.addonguid` is going to exist ? Couldn't we use _that_ directly everywhere instead of having the property?

eviljeff: If Addon.guid is `None` it wouldn't get created (that would be invalid, and it's unlikely to ever affect blocklist code because it's all based on guid).  Plus there isn't an absolute guarantee that every Addon will have a matching AddonGUID in the database because the fk is on AddonGUID rather than on Addon.

Changing it to be `self.addonguid.guid` directly is more straightforward, at the potential risk of a traceback somewhere in the future.

eviljeff: Although maybe the name conflicting with the queryset syntax is an accident waiting to happen.

diox: Yeah I think we need a) that explanation or a variant of it in a comment or docstring and b) a new name to avoid the queryset syntax.

eviljeff: https://github.com/mozilla/addons-server/pull/14469/commits/ba4ba1778b9729cf3c290a116f35e2e11bf90b0c

Issue #14484, opened 2020-06-03
Update BigQuery dataset name


willdurand:
We need to update the BigQuery dataset because we're not using `telemetry` anymore. See: https://bugzilla.mozilla.org/show_bug.cgi?id=1633746#c13

**For QA:** the only visible change is that currently -dev does not display beta stats anymore and if you open the devtools > network panel, you'll see errors for the XHR requests. Once this is fixed, there shouldn't be any error but we won't have any data to display.

ioanarusiczki:
@willdurand For 
https://addons-dev.allizom.org/en-US/firefox/addon/view-page-archive-cache/statistics/beta/?last=30 and 
https://addons-dev.allizom.org/en-US/firefox/addon/awesome-screenshot-plus-/statistics/beta/?last=365

I can see the stats again including the pie charts which were not loaded until I checked the classic view and then went back to /beta/

willdurand:
@ioanarusiczki perfect. see my comment in the other issue about add-ons that will have more data.

PR #14485, opened 2020-06-03
Adjust bigquery dataset name

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14484

---

This should be correct if I understood the ops changes correctly.


willdurand:
@diox fwiw, I tried in local dev and it worked. I did a last minute change to make sure we're querying a table instead of a view as per https://bugzilla.mozilla.org/show_bug.cgi?id=1633746#c16

Issue #14436, opened 2020-05-29
Restriction logging stopped after May 21 push


wagnerand:
Since the push on May 21, we are no longer seeing restriction logs in stackdriver. Probably caused by #14310.

diox:
Not just restriction logging, *all* logging from `z.users` it looks like.

willdurand:
Current `master`:

```
'PIL': <logging.PlaceHolder object at 0x7f4848a56c50>,
'PIL.Image': <Logger PIL.Image (DEBUG)>,
'accounts': <Logger accounts (DEBUG)>,
'accounts.verify': <Logger accounts.verify (DEBUG)>,
'amqp': <Logger amqp (DEBUG)>,
'asyncio': <Logger asyncio (DEBUG)>,
'basket': <logging.PlaceHolder object at 0x7f4848c020f0>,
'basket.base': <Logger basket.base (DEBUG)>,
'bcdocs': <Logger bcdocs (DEBUG)>,
'boto3': <Logger boto3 (DEBUG)>,
'boto3.resources': <logging.PlaceHolder object at 0x7f4845cf84a8>,
'boto3.resources.action': <Logger boto3.resources.action (DEBUG)>,
'boto3.resources.base': <Logger boto3.resources.base (DEBUG)>,
'boto3.resources.collection': <Logger boto3.resources.collection (DEBUG)>,
'boto3.resources.factory': <Logger boto3.resources.factory (DEBUG)>,
'boto3.resources.model': <Logger boto3.resources.model (DEBUG)>,
'botocore': <Logger botocore (DEBUG)>,
'botocore.args': <Logger botocore.args (DEBUG)>,
'botocore.auth': <Logger botocore.auth (DEBUG)>,
'botocore.awsrequest': <Logger botocore.awsrequest (DEBUG)>,
'botocore.client': <Logger botocore.client (DEBUG)>,
'botocore.compat': <Logger botocore.compat (DEBUG)>,
'botocore.configprovider': <Logger botocore.configprovider (DEBUG)>,
'botocore.credentials': <Logger botocore.credentials (DEBUG)>,
'botocore.discovery': <Logger botocore.discovery (DEBUG)>,
'botocore.endpoint': <Logger botocore.endpoint (DEBUG)>,
'botocore.handlers': <Logger botocore.handlers (DEBUG)>,
'botocore.history': <Logger botocore.history (DEBUG)>,
'botocore.hooks': <Logger botocore.hooks (DEBUG)>,
[...]

PR #14486, opened 2020-06-03
Fix logging config

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14436

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -25,7 +25,9 @@ INSTALLED_APPS += (
 LOGGING['root']['level'] = logging.DEBUG
 for logger in list(LOGGING['loggers'].keys()):
     if logger not in ['filtercascade', 'mohawk.util', 'post_request_task']:
-        del LOGGING['loggers'][logger]
+        # It is important to keep the loggers configured in `settings_base.py`
+        # so we only update the level below:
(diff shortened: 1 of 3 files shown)


settings.py
```diff
@@ -25,7 +25,7 @@
 LOGGING['root']['level'] = logging.DEBUG
 for logger in list(LOGGING['loggers'].keys()):
     if logger not in ['filtercascade', 'mohawk.util', 'post_request_task']:
-        del LOGGING['loggers'][logger]
+        LOGGING['loggers'][logger]['level'] = logging.DEBUG
```
willdurand, line 28: This seems safer as we'll have the exact same logger defined in local dev and prod.

src/olympia/lib/settings_base.py
```diff
@@ -1357,6 +1358,11 @@ def get_db_config(environ_var, atomic_requests=True):
             'level': logging.WARNING,
             'propagate': False
         },
+        'z': {
+            'handlers': ['mozlog'],
+            'level': logging.INFO,
+            'propagate': True
```
willdurand, line 1364: This is needed to let `z.*` log messages too.

Issue #10003, opened 2018-11-16
The review page is not displayed for an Add-on reviewer that was manually approved by an Admin


ioanarusiczki:
Moving issue https://github.com/mozilla/addons-frontend/issues/6873 

STR:

from Rev Tools 

1. Load AMO dev with an Add-on reviewer account that had no Legacy Reviewer rights and go to Add-on Review Log.
2. Click on "Approved" link for an add-on that was manually approved by an Admin.

from and Add-on's detail page 

1. Load AMO dev with an Add-on reviewer account and go to an add-ons detail page.
2. Click the  "Review add-on code" link from More information section. 

Expected results:
The reviewer can view the detail page of the review page.

Actual results:
The details of the review page is not displayed, instead there's an "Oops! Not allowed."

Notes:
This issue is not reproducible when the user belongs to both Reviewers: Add-ons and Reviewers: Legacy Groups.
Reproducible on AMO dev and AMO stage with FF63 (Win10).

![rev tools - oops](https://user-images.githubusercontent.com/33448286/48615343-53ccd100-e999-11e8-95f3-034e44ecc5e9.gif)

![frontend - oops](https://user-images.githubusercontent.com/33448286/48615393-79f27100-e999-11e8-9066-574592765039.gif)

ioanarusiczki:
This issue is still reproducible on AMO dev - FF68(Win10).

PR #14489, opened 2020-06-03
Move addons-server to use beacon for Google Analytics requests

### LINKED ISSUE #14427: Move addons-server to use beacon for Google Analytics requests
We should move to using beacon for GA requests to ensure better reliability and tighten our CSP.

To do this we need to check if we need to update from the legacy GA implementation first (we probably do since I can only find docs about configuring the transport mechanism in the newer implementation).

Doing this requires the following changes:

- Update the GA implementation on addons-server (if required)
- Configure GA to use beacon as a transport.
- Remove the GA host entry from `img-src` in the CSP
- Add an entry for a GA host to `connect-src` in the CSP.
- Test analytics are still working.

This was done for addons-frontend a few months ago in https://github.com/mozilla/addons-frontend/commit/e82335d1a6c7e5338d275531124c9fd639de7f82

bobsilverberg:
Fixes #14427 

I was able to test a page view in devhub and see that analytics are still working, and using the `beacon` transport, but I'm not sure where the changes in  `static/js/lib/jquery.pjax.js`, in `static/js/shield_study/main.js` or in `static/js/zamboni/buttons.js` are manifested. Can anyone provide some guidance on those?


eviljeff:
> I'm not sure where the changes in static/js/lib/jquery.pjax.js,

I don't think we use that file, so it can be deleted  Though we do use the file with the same name (and similar content? But maybe doesn't touch GA?) in static/js/node-lib/jquery.pjax.js.

> in static/js/shield_study/main.js 

not used any more - the shield studies got removed in #10003 

> in static/js/zamboni/buttons.js

this file is included throughout the site, but it's not clear the buttons (that reference GA) are used anymore - they seem to be concerned with install buttons which is now a frontend thing.

bobsilverberg:
> > I'm not sure where the changes in static/js/lib/jquery.pjax.js,
> 
> I don't think we use that file, so it can be deleted Though we do use the file with the same name (and similar content? But maybe doesn't touch GA?) in static/js/node-lib/jquery.pjax.js.
> 
> > in static/js/shield_study/main.js
> 
> not used any more - the shield studies got removed in #10003
> 

I think that's the wrong link. Do you have the correct one handy so I can double check that this file can be removed?

> > in static/js/zamboni/buttons.js
> 
> this file is included throughout the site, but it's not clear the buttons (that reference GA) are used anymore - they seem to be concerned with install buttons which is now a frontend thing.

bobsilverberg:
> cleaning up the obsolete js files could be a followup

Thanks. I think it makes sense to do at least the simple cleanup for now, so I'll look at removing `static/js/lib/jquery.pjax.js` and `static/js/shield_study/main.js`.

eviljeff:
> 
> 
> > > I'm not sure where the changes in static/js/lib/jquery.pjax.js,
> > 
> > 
> > I don't think we use that file, so it can be deleted Though we do use the file with the same name (and similar content? But maybe doesn't touch GA?) in static/js/node-lib/jquery.pjax.js.
> > > in static/js/shield_study/main.js
> > 
> > 
> > not used any more - the shield studies got removed in #10003
> 
> I think that's the wrong link. Do you have the correct one handy so I can double check that this file can be removed?

#10448 though you can see for yourself by searching for "shield" in the code - the only references are the redirects we left in place.

Issue #14463, opened 2020-06-02
Make stats pages private


willdurand:
According to this [PRD](https://docs.google.com/document/d/1Db0czumd_VFLOdSqsuDvlmWumqFQo8eIgGhLQ1WUPzo/edit), we would like to remove the `public_stats` setting. Developers (and admins) will be the only users who can have access to the stats pages.

willdurand:
@jvillalobos is that accurate? Where should we display the link to the stats pages in DevHub? Do we want to keep the link in the frontend for authenticated users?

My plan would be to:

1. make sure we have links to access the stats pages
2. remove the `public_stats` setting

In the meantime, we'll have to reach out to developers who have add-ons with public stats.

willdurand:
Also, semi-related to this: for now we'll keep the existing pages as they are currently (no UI changes). Is that OK? "Moving" the pages to the DevHub look&feel will be a bit challenging but it also sounds more like a code quality issue than a feature _per se_.

jvillalobos:
> Where should we display the link to the stats pages in DevHub?

There are two links on DevHub that I know about: `View Statistics Dashboard` in the navigation sidebar in the edit pages of an add-on, and a `Statistics` link in the Manage My Submissions page (`/developers/addons`).

> Do we want to keep the link in the frontend for authenticated users?

Yes.

> In the meantime, we'll have to reach out to developers who have add-ons with public stats.

Let me know when you expect this change to happen and I'll coordinate comms with @caitmuenster 

> for now we'll keep the existing pages as they are currently (no UI changes). Is that OK?

Yes, that's okay for now. As I noted above, there are links in DevHub already, so the experience will be the same.

willdurand:
> There are two links on DevHub that I know about: `View Statistics Dashboard` in the navigation sidebar in the edit pages of an add-on, and a `Statistics` link in the Manage My Submissions page (`/developers/addons`).

Oh mm, sorry. I didn't see these links for some reasons. Great!

> Let me know when you expect this change to happen and I'll coordinate comms with @caitmuenster

@jvillalobos I think next week (2020.06.11) if that's fine?

> Yes, that's okay for now. As I noted above, there are links in DevHub already, so the experience will be the same.

Perfect, thanks

ioanarusiczki:
@willdurand 
- the "public setting" is no longer available in dev hub and addons are no longer "public". All pages I verified are displayed as "private". 

Observations:

[...]

PR #14490, opened 2020-06-03
Remove public_stats

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14463


src/olympia/amo/tests/test_models.py
```diff
@@ -68,34 +68,34 @@ def test_multiple_ignored(self):
 
     def test_change_called_on_new_instance_save(self):
         for create_addon in (Addon, Addon.objects.create):
-            addon = create_addon(public_stats=False, type=amo.ADDON_EXTENSION)
-            addon.public_stats = True
+            addon = create_addon(view_source=False, type=amo.ADDON_EXTENSION)
+            addon.view_source = True
```
eviljeff, line 72: can we fix this in a way that doesn't rely on `.view_source` - it's supposed to be removed from the model too in https://github.com/mozilla/addons-server/issues/12842

eviljeff: #12842 is a contrib:welcome issue too, fwiw, so it'll be making that more difficult for contributors to fix.

willdurand: ah yes, sure. I didnt know about this. Thanks!

Issue #14496, opened 2020-06-03
Uncaught ReferenceError: jQuery is not defined


willdurand:
### Describe the problem and steps to reproduce it:

1. Go to https://addons-dev.allizom.org/en-US/firefox/addon/some-addon/statistics/beta/usage/
2. Open the devtools > console

### What happened?

- The charts don't load on the page.
- JavaScript errors in the console

### What did you expect to happen?

The charts load.

### Anything else we should know?

```
Uncaught ReferenceError: jQuery is not defined
    <anonymous> https://addons-dev-cdn.allizom.org/static/js/impala-min.js?build=aad65070-5ed7e28e:1
impala-min.js:1:18938
    <anonymous> https://addons-dev-cdn.allizom.org/static/js/impala-min.js?build=aad65070-5ed7e28e:1
```

```
Uncaught TypeError: fe is not a function
    <anonymous> https://addons-dev-cdn.allizom.org/static/js/zamboni/stats-min.js?build=aad65070-5ed7e28e:1
    <anonymous> https://addons-dev-cdn.allizom.org/static/js/zamboni/stats-min.js?build=aad65070-5ed7e28e:1
stats-min.js:1:10460
```

willdurand:
@bobsilverberg I added the statistics label because that's where I found the error but it works in stage so it is something new-ish on `master`.

bobsilverberg:
I cannot reproduce this locally, although when I go to that page for an add-on it tells me there are no statistics, so I'm not sure if I need populated statistics to be able to reproduce this error. The page seems fine to me, the GA code is executing, and I don't see any errors in the console about `jQuery`. 

@eviljeff suggested it might be a problem with the pipeline not being able to handle the JS for the new analytics code, but I don't think I'm doing anything fancy in there. Is there somewhere that we might be able to see any build failures?

bobsilverberg:
I do see an error on the devhub landing page (on dev):

```
ReferenceError: $ is not defined js-min.js:1:1
    <anonymous> https://addons-dev-cdn.allizom.org/static/js/devhub/new-landing/js-min.js?build=aad65070-5ed7e28e:1
```

which sounds similar (`$` tends to be `jQuery`), so it's not limited to the stats pages.

bobsilverberg:
I think it's worth just putting the file I deleted back and see if that fixes it. I'll try that. I'm not sure it's testable without landing the change on dev though.

diox:
[...]

PR #14507, opened 2020-06-04
Do not use trailing comma in JS function calls

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14496


willdurand:
> trailing, not leading :)

ha yeah..

Issue #14493, opened 2020-06-03
Collect some statsd metrics for the new stats (BigQuery)


willdurand:
It would probably be useful to have some statsd metrics for BigQuery (like execution time) to see how BigQuery in production behaves. This will be useful to determine if we need to introduce caching or if we're good as is.

ioanarusiczki:
@willdurand Please add some STR for qa.

PR #14513, opened 2020-06-04
Add statsd timer for bigquery call

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14493

---

I can't think of anything else right now. @diox any ideas?


diox:
Maybe in the future we'll want to know which series are the slowest ? I'd imagine it would just tell us that the larger the date range, the slower it is...

willdurand:
> Maybe in the future we'll want to know which series are the slowest ? I'd imagine it would just tell us that the larger the date range, the slower it is...

Ha yeah, that's a good idea.

Issue #14439, opened 2020-05-29
Normalize restricted email during email restriction check


wagnerand:
For email restriction, we compare the normalized email of the user making the request to the non-normalized email pattern in the restriction records.

This can cause restrictions not to be applied because the two do not match.

For STR, please reach out to me directly.

AlexandraMoga:
I've tested this based on the instructions received from Andreas and the email addresses used in my tests were normalized as expected. The restrictions were also applied at submission.

PR #14516, opened 2020-06-04
Normalize email patterns

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14439


willdurand:
@diox quickly chatting with @wagnerand who said you'd wanted to do this during lookup. As shown in this patch, that won't work for `email_pattern` values that are in fact domains.. So I think normalizing the data before saving would make sense here?

diox:
Yeah that's fine, I like this approach.

PR #14519, opened 2020-06-04
Re-enable z.* loggers

willdurand:


src/olympia/lib/settings_base.py
```diff
@@ -1357,6 +1357,11 @@ def get_db_config(environ_var, atomic_requests=True):
             'level': logging.WARNING,
             'propagate': False
         },
+        'z': {
+            'handlers': ['mozlog'],
+            'level': logging.INFO,
+            'propagate': True
```
eviljeff, line 1363: not ` 'propagate': False` like it was before all this?

willdurand: apparently not.. I can't explain why but it does not work without `propagate=True` 😕

PR #14521, opened 2020-06-04
Remove old add-on categories

willdurand:
Fixes https://github.com/mozilla/addons-frontend/issues/9453

---

I hard deleted the old add-on categories, hopefully it's ok..


willdurand:
> I hard deleted the old add-on categories, hopefully it's ok..

@eviljeff WDYT?

src/olympia/addons/migrations/0010_auto_20200604_1319.py
```diff
@@ -0,0 +1,59 @@
+# Generated by Django 2.2.12 on 2020-06-04 13:19
+
+from django.db import migrations
+
+
+def delete_old_categories(apps, schema_editor):
+    AddonCategory = apps.get_model('addons', 'AddonCategory')
+    AddonCategory.objects.filter(
+        pk__in=(
+            # persona
```
eviljeff, line 10: can we filter this qs based on `type` instead of hard-coding the ids?  There may be other non-constant defined Category instances still in the databases.

willdurand: Ha, right. I realized that I wasn't using the right model either, although it worked.. I changed my strategy anyway and removed all categories that we aren't supporting anymore.

src/olympia/addons/migrations/0010_auto_20200604_1319.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 2.2.12 on 2020-06-04 13:19
```
bobsilverberg, line 1: Nit: Don't we tend to rename the migration files to describe what they are doing?

src/olympia/addons/tests/test_views.py
```diff
@@ -1408,12 +1407,12 @@ def get_category(type_, name):
 
         # Search for add-ons in the first category. There should be two.
         data = self.perform_search(self.url, {'app': 'firefox',
-                                              'type': 'search,statictheme',
-                                              'category': 'music'})
+                                              'type': 'extension,statictheme',
+                                              'category': 'other'})
```
bobsilverberg, line 1411: Should we use a variable to store this category slug instead of relying on it being hardcoded correctly here and above?

src/olympia/addons/migrations/0010_auto_20200604_1319.py
willdurand: I rarely do this.

src/olympia/addons/tests/test_views.py
willdurand: I guess it's fine here?

Issue #14525, opened 2020-06-04
Indexing of a newly initialized database fails


wagnerand:
```
python3 manage.py reindex --wipe --force --noinput
{"Timestamp": 1591287203964317952, "Type": "raven.contrib.django.client.DjangoClient", "Logger": "http_app_addons", "Hostname": "7f3b9e2bb260", "EnvVersion": "2.0", "Severity": 6, "Pid": 1463, "Fields": {"uid": "", "remoteAddressChain": "", "msg": "Raven is not configured (logging is disabled). Please see the documentation for more information."}, "severity": 200}
{"Timestamp": 1591287204023002112, "Type": "raven.base.Client", "Logger": "http_app_addons", "Hostname": "7f3b9e2bb260", "EnvVersion": "2.0", "Severity": 6, "Pid": 1463, "Fields": {"uid": "", "remoteAddressChain": "", "msg": "Raven is not configured (logging is disabled). Please see the documentation for more information."}, "severity": 200}
/deps/lib/python3.6/site-packages/multidb/__init__.py:68: UserWarning: [multidb] No replica databases are configured! You can configure them with the REPLICA_DATABASES setting.
  UserWarning,
Starting the reindexation for addons.
{"Timestamp": 1591287205368356864, "Type": "z.elasticsearch", "Logger": "http_app_addons", "Hostname": "7f3b9e2bb260", "EnvVersion": "2.0", "Severity": 6, "Pid": 1463, "Fields": {"uid": "<anon>", "remoteAddressChain": "", "msg": "Unflagging the database"}, "severity": 200}
--- Logging error ---
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/logging/__init__.py", line 994, in emit
    msg = self.format(record)
  File "/usr/local/lib/python3.6/logging/__init__.py", line 840, in format
    return fmt.format(record)
  File "/deps/lib/python3.6/site-packages/dockerflow/logging.py", line 138, in format
    out = self.convert_record(record)
  File "/code/src/olympia/core/logger.py", line 58, in convert_record
    out = super().convert_record(record)
  File "/deps/lib/python3.6/site-packages/dockerflow/logging.py", line 127, in convert_record
    fields["error"] = repr(record.exc_info[1])
TypeError: 'bool' object is not subscriptable
Call stack:
  File "manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
[...]

PR #14526, opened 2020-06-04
Hack for initialize failure

wagnerand:
Probably not the right approach, but @diox asked for it as a starting point for #14525.


diox:
Don't need to keep that opened, this is just a hack

PR #14527, opened 2020-06-04
Trying to get validation.json to work

bobsilverberg:
The `json_file_validation` view in devhub returns a 404 when the add-on has been deleted. 

I am trying to fix that, and I've traced the current source of the 404 to the call to https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/decorators.py#L28. I changed the `qs` that is being passed into `addon_view` via the `dev_required` decorator from `Addon.objects.all` to `Addon.unfiltered.all` (as can be seen in the attached patch), thinking that would allow it to find the deleted add-on, but I'm still getting a 404 at https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/decorators.py#L28.

I realize that I might not be able to make this change to `dev_required` without some undesirable repercussions, but at this point I'm just trying to debug it and figure out how to eliminate the 404.

### DIFF
--- a/src/olympia/devhub/decorators.py
+++ b/src/olympia/devhub/decorators.py
@@ -19,7 +19,7 @@ def dev_required(owner_for_post=False, allow_reviewers_for_read=False,
     When allow_reviewers is True, reviewers can view the page.
     """
     def decorator(f):
-        @addon_view_factory(qs=Addon.objects.all)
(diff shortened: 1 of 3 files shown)


bobsilverberg:
We figured it out. Thanks @diox and @willdurand!

Issue #14491, opened 2020-06-03
Add a waffle flag to rollout the beta stats pages


willdurand:
We'd like to rollout the beta stats for more users in the near future (like start with 5-10-25% of our user base). In order to do this, we'll need a waffle flag. This flag will replace the email address check introduced in https://github.com/mozilla/addons-server/issues/14424.

This flag will also add links to beta stats for _unlisted_ add-ons.

willdurand:
@jvillalobos thinking about it, I believe we can ship this flag next week and maybe start to rollout for 5% of the users or something and target 100% on July 2nd. WDYT?

The reason why we'd like to have a flag and a progressive rollout is that we don't know how BigQuery will behave in production. It seems fine but better safe than sorry.

willdurand:
> This flag will also add links to beta stats for _unlisted_ add-ons.

Also, I think users who will be able to see the beta stats will only see the beta stats (and not both versions of the stats). That will make the transition easier.

jvillalobos:
We want to give developers time to export their old stats, if they want to, so they need a clear date after which they shouldn't expect the old dashboard to work. I see a couple of options: (1) do the rollout, but start later (06/18) to give devs more time, or (2) give developers control over the rollout, maybe by having a switch in the stats page (it should be fine if it's a one way switch, as long as we are clear about it).

willdurand:
@jvillalobos ha, right. Maybe we could use the flag to enable the beta stats by default and add a link to the "old" stats from the beta pages?

jvillalobos:
That would work!

willdurand:
@jvillalobos so I update the messages a bit and we have two messages now:

1. When a user lands on any of the beta stats pages (if the user did not have access to the beta, it would return a 404):

<img width="1016" alt="Screen Shot 2020-06-05 at 14 21 54" src="https://user-images.githubusercontent.com/217628/83875788-07421580-a738-11ea-9438-92845cacf02d.png">

2. When a user who has access to the beta lands on any of the "old" stats pages:

<img width="1012" alt="Screen Shot 2020-06-05 at 14 22 07" src="https://user-images.githubusercontent.com/217628/83875791-08734280-a738-11ea-9b5a-fd99e5ef4bc2.png">

I'd be happy to reword these two messages if needed.. WDYT?

willdurand:
> This flag will also add links to beta stats for _unlisted_ add-ons.

[...]

PR #14542, opened 2020-06-05
Add waffle flag for the beta stats

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14491

---

In this patch, we'll give access to the beta stats to more users (controlled by a flag) and users who can access beta stats will also be able to view stats for unlisted add-ons.


settings.py
```diff
@@ -62,6 +62,7 @@ def insert_debug_toolbar_middleware(middlewares):
 # If you're not running on SSL you'll want this to be False.
 SESSION_COOKIE_SECURE = False
 SESSION_COOKIE_DOMAIN = None
+WAFFLE_SECURE = False
```
eviljeff, line 65: interesting.  Why do we need this?

src/olympia/devhub/templates/devhub/addons/listing/item_actions.html
```diff
@@ -47,12 +47,20 @@ <h5>{{ _('Actions') }}</h5>
...
     {% endif %}
   {% endif %}
-  {% if not addon.disabled_by_user and show_listed and addon.status == amo.STATUS_APPROVED %}
-    <li>
-      <a href="{{ url('stats.overview', addon.slug) }}" class="tooltip"
-         title="{{ _('Daily statistics on downloads and users') }}">
-        {{ _('Statistics') }}</a>
-    </li>
+  {% if not addon.disabled_by_user and addon.status == amo.STATUS_APPROVED %}
+    {% if waffle.flag('beta-stats') %}
```
eviljeff, line 51: Are we actually planning to segment the people who have access by user or enable it for a % of users?  Not sure why we're not just using waffle.switch.

willdurand: % of users

settings.py
willdurand: It needs to be `False` so that the cookie works in local dev. Otherwise FF rejects the cookie because of `SameSite`. I'll have to verify how it behaves in our hosted envs.

PR #14545, opened 2020-06-05
Allow code manager APIs to return data for deleted add-ons

### LINKED ISSUE #14498: The code manager files are not shown for a deleted addon
Depends on https://github.com/mozilla/addons-server/issues/14607
Followup for https://github.com/mozilla/addons-server/issues/13634#issuecomment-638219693

STR:
When an add-on is uploaded (I waited for extraction and I could open the files) and then deleted the API responds with a 404 

Examples:
https://code.addons-dev.allizom.org/en-US/browse/619895/versions/1691482/
https://code.addons-dev.allizom.org/en-US/browse/619894/versions/1691481/

Expected:
Maybe the files should be visible for the deleted addons - the legacy viewer shows them. 
Probably needs a discussion. 

This is reproducible on AMO dev with FF76(Win10).

bobsilverberg:
~~Depends on https://github.com/mozilla/addons-server/pull/14611~~
Fixes #14498

### DIFF
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -678,7 +678,7 @@ def file_validation(request, addon_id, addon, file_id):
 
 
 @csrf_exempt
-@dev_required(allow_reviewers_for_read=True)
+@dev_required(allow_reviewers_for_read=True, qs=Addon.unfiltered.all)
 def json_file_validation(request, addon_id, addon, file_id):
     file = get_object_or_404(File, version__addon=addon, id=file_id)
     try:
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -678,6 +678,8 @@ def file_validation(request, addon_id, addon, file_id):
 
 
 @csrf_exempt
+# This allows read-only access to deleted add-ons for reviewers
+# but not developers.
 @dev_required(allow_reviewers_for_read=True, qs=Addon.unfiltered.all)
 def json_file_validation(request, addon_id, addon, file_id):
     file = get_object_or_404(File, version__addon=addon, id=file_id)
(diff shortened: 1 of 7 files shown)


bobsilverberg:
@diox The conflict has been resolved.

src/olympia/devhub/views.py
```diff
@@ -677,7 +677,7 @@ def file_validation(request, addon_id, addon, file_id):
 
 
 @csrf_exempt
-@dev_required(allow_reviewers_for_read=True)
+@dev_required(allow_reviewers_for_read=True, qs=Addon.unfiltered.all)
```
diox, line 680: So, following the implications of that wasn't trivial. It looks like with that change, for deleted add-ons:
- developers still don't have access because `check_addon_ownership()` which always return `False` for deleted add-ons
- reviewers retain read-only access (the only thing required for this view) because they bypass that same check entirely, thanks to `allow_reviewers_for_read=True`.

Can you add a test covering the fact that developers don't get access, if there isn't already one ? Also, add a comment explaining who gets access and why next to that decorator call would be helpful.

src/olympia/reviewers/views.py
```diff
@@ -1329,7 +1329,7 @@ def get_addon_object(self):
         if not hasattr(self, 'addon_object'):
             # We only need translations on the add-on, no other transforms.
             self.addon_object = get_object_or_404(
-                Addon.objects.get_queryset().only_translations(),
+                Addon.unfiltered.all().only_translations(),
```
diox, line 1332: With that change and the one below we *really* want https://github.com/mozilla/addons-server/issues/14607 cause developers should never be able to see deleted add-ons. So to be safe, maybe implement #14607 first.

bobsilverberg:
I've added "do not merge" to this and commented that it depends on https://github.com/mozilla/addons-server/pull/14611.

src/olympia/devhub/views.py
bobsilverberg: I've addressed these issues in the latest commit.

bobsilverberg:
https://github.com/mozilla/addons-server/pull/14611 landed, so this should be good to go, although I see I also introduced a new conflict, which I'll resolve this morning.

bobsilverberg:
@diox Conflicts have been resolved once again.

Issue #14494, opened 2020-06-03
setup cron job(s) to delete account/addon data after a set period


eviljeff:
see #13896 for the main issue.

This is the implement the steps mentioned in the deletion policy:
> When an AMO account is deleted, all of its associated data is deleted with it. The only exception is when the account is the sole owner of an add-on, in which case:
>
> - All add-on files and their associated metadata will be retained for a total of 7 years, after which they will be permanently deleted.
> - The developer’s email address and IP addresses used to submit their add-ons will be retained for a total of 7 years, after which they will be permanently deleted.

(and other comments relating to the non-immediate deletion of data)

Data retention policy doc: https://docs.google.com/document/d/1E4rFeLhsafcOnesetDbpaXLiCkGjkCsgnZnzbso9Fls/edit#heading=h.91acyaz58wl1

eviljeff:
@jvillalobos the current implementation retains email, fxa_id, and last_login_ip on the UserProfile (along with any addons they own), for **all** users.  last_login_ip is then cleared after 6 months for all users.  

- Do we want to differentiate users with addons, and users without addons?  Or is clearing for all users after 7 years sufficient/desirable?
- does `last_login_ip` still need to be cleared after 6 months, or is that on a 7 year timescale too?
- At the 7 year mark, should we clear email and fxa_id too? 
- what about `notes` (admin notes when deleting accounts)?

diox:
> does `last_login_ip` still need to be cleared after 6 months, or is that on a 7 year timescale too?

FWIW that 6 month period was something we came up with Legal & Trust & Privacy teams. They mentioned in the email thread that we could re-assess if needed, but they should be involved in that discussion if we do.

jvillalobos:
> Do we want to differentiate users with addons, and users without addons?

Yes, the 7 year data retention only applies to users with add-ons. For all other users we should delete the data within 48 hours.

> does last_login_ip still need to be cleared after 6 months, or is that on a 7 year timescale too?

For add-on developers, the same 7 year timescale.

> At the 7 year mark, should we clear email and fxa_id too?

Yes.

> what about notes (admin notes when deleting accounts)?

That should be okay to keep indefinitely, but I don't think it'll be of much use if all other data is cleared. So, whatever's easiest.

> FWIW that 6 month period was something we came up with Legal & Trust & Privacy teams.

The new policy was cleared with those teams, so we should be okay.

eviljeff:
[...]

PR #14546, opened 2020-06-05
add delete_user_event task

### LINKED ISSUE #13896: Listen to FXA SQS queue to process account deletions
Supercedes #7922

See https://docs.google.com/document/d/1hjlWVzzrSrGzMYKmxqK76MzOVQbV7sbc7padtj8xclA/edit# for details.

Once this is ready the release will need to be coordinated with landing https://github.com/mozilla/fxa/pull/4532/

eviljeff:
fixes #13896 apart from:
- adding the post-delete data scrubbing after x years (split out to #14494 
- reconciling the differences with AMO (frontend via the API) triggered deletes (will be split out when jorge confirms)

The code is behind a waffle switch until FxA have made the changes on their end.

### DIFF
--- a/src/olympia/accounts/tasks.py
+++ b/src/olympia/accounts/tasks.py
@@ -1,3 +1,4 @@
+import functools
 from datetime import datetime
 
 import olympia.core.logger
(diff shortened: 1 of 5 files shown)


src/olympia/accounts/tasks.py
```diff
@@ -10,24 +11,45 @@
...
+                profile.id, email, changed_date))
+    else:
+        log.warning('Account pk [%s] email updated ignored, %s > %s' %
+                    (profile.id, profile.email_changed, changed_date))
+
+
+@task
+@use_primary_db
+@user_profile_from_uid
+def delete_user_event(user, deleted_date):
```
bobsilverberg, line 49: Trying to understand how the `@user_profile_from_uid` decorator works. In this case `user` is what is returned by the wrapper in the decorator, which is the same thing as `profile` in `primary_email_change_event`, above (just a different argument name)? So it's positional - i.e., the first argument to this function is whatever is returned by `@user_profile_from_uid`?

src/olympia/accounts/tests/test_utils.py
```diff
@@ -247,20 +268,39 @@ def test_success(self, primary_email_change_event):
...
+    def setUp(self):
+        self.email_changed_date = self.days_ago(42)
+        self.body = json.dumps({'Message': json.dumps(
+            {'event': 'delete',
+             'uid': 'ABCDEF012345689',
+             'ts': totimestamp(self.email_changed_date)})})
+
+    @override_switch('fxa-account-delete', active=True)
+    def test_success_integration(self):
+        user = user_factory(fxa_id='ABCDEF012345689')
```
bobsilverberg, line 286: How about setting this into `self.uid` in `setUp` and then referencing that variable here, rather than relying on accurate hardcoding?

src/olympia/accounts/tests/test_utils.py
```diff
@@ -247,20 +268,39 @@ def test_success(self, primary_email_change_event):
...
+        process_fxa_event(self.body)
+        user.reload()
+        assert user.email is None
+        assert user.deleted
+        assert user.fxa_id is None
+
+    @override_switch('fxa-account-delete', active=True)
+    @mock.patch('olympia.accounts.utils.delete_user_event.delay')
+    def test_success(self, delete_user_event_mock):
```
bobsilverberg, line 295: Do we really need this if we also have the integration test, above? I guess this ensures that the results are due to `delete_user_event` being called, and not some other side effect? Is there a way to combine the two and spy on `delete_user_event` but also have it actually execute? We can do that in JS tests but I'm not sure if `unittest` supports that.

eviljeff:
> 
> 
> Could you add a test that makes sure the email address and IP addresses are retained if the user is an add-on developer?

This is retained via other logging, isn't it?  (Otherwise just deleting your addons before you delete your account would clear the account email)

src/olympia/accounts/tests/test_utils.py
eviljeff: it's possible, but a pain to do with mock, so we generally don't do it.  (Normally we would just have either the integration test _or_ the mocked test - I must have been feeling extra cautious when I added the tests for the earlier integration, and I just copied them here)

src/olympia/accounts/tasks.py
eviljeff: `return f(profile, timestamp, *args, **kw)` is the return from the decorator, so yeah, it takes a user arg (which is then an fxa_id) and a date-time arg (which is then a timestamp) and returns the same but now as a `UserProfile` and a `datetime` instance.

user or profile... it's a `UserProfile` instance so either variable name is descriptive.

wagnerand:
> > Could you add a test that makes sure the email address and IP addresses are retained if the user is an add-on developer?
> 
> This is retained via other logging, isn't it? (Otherwise just deleting your addons before you delete your account would clear the account email)

Unfortunately, it's not logged in other places, at least not where it's accessible. The routine would need to include deleted accounts when checking if a user is a developer.

wagnerand:
(I removed some comments about a specific detail to unblock this PR)

src/olympia/accounts/tests/test_utils.py
eviljeff: https://github.com/mozilla/addons-server/pull/14546/commits/b4ec801141c93ca71daf6459ed7b73118d9182a5

PR #14564, opened 2020-06-08
Add missing period

willdurand:
See: https://github.com/mozilla/addons-server/issues/14491#issuecomment-639667412

### DIFF
--- a/src/olympia/stats/templates/stats/stats.html
+++ b/src/olympia/stats/templates/stats/stats.html
@@ -102,7 +102,7 @@
     {% set beta_stats_overview_url = url('stats.overview.beta', addon.slug) %}
     {% trans %}
     We are rolling out new statistics pages and you have access to the preview!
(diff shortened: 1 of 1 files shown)


src/olympia/stats/templates/stats/stats.html
```diff
@@ -102,7 +102,7 @@ <h1>
     {% set beta_stats_overview_url = url('stats.overview.beta', addon.slug) %}
     {% trans %}
     We are rolling out new statistics pages and you have access to the preview!
-    <a href="{{ beta_stats_overview_url }}">Take me to the new dashboard</a>
+    <a href="{{ beta_stats_overview_url }}">Take me to the new dashboard</a>.
```
bobsilverberg, line 105: Are you sure the period shouldn't be inside the `<a>` tag? I know this seems like a uber-nit, but I actually think it might look a bit weird if the entire message, other than the period, appeared like a link.

PR #14566, opened 2020-06-08
Remove stats dashboard visibility section

willdurand:
Refs. https://github.com/mozilla/addons-server/issues/14463

### DIFF
--- a/src/olympia/stats/templates/stats/stats.html
+++ b/src/olympia/stats/templates/stats/stats.html
@@ -180,6 +180,9 @@
         </div>
       </div>
     {% endblock %}
+    <div id="stats-permissions">
+      <p>{{ _('This dashboard is <b>private</b>.') }}</p>
+    </div>
     <div class="hidden">
       <div id="fieldMenuPopup" class="popup">
         <form id="fieldMenu" data-no-csrf>
--- a/src/olympia/stats/templates/stats/stats.html
+++ b/src/olympia/stats/templates/stats/stats.html
@@ -180,15 +180,6 @@
         </div>
       </div>
     {% endblock %}
-    {% if addon %}
-      <div id="stats-permissions">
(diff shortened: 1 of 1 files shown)


src/olympia/stats/templates/stats/stats.html
```diff
@@ -180,15 +180,6 @@ <h1>
         </div>
       </div>
     {% endblock %}
-    {% if addon %}
-      <div id="stats-permissions">
-        <p>{{ _('This dashboard is <b>private</b>.') }}</p>
```
eviljeff, line 185: Should this be left here but moved out of the `{% if ... ` block or is it mentioned somewhere else on the page?

src/olympia/stats/templates/stats/stats.html
```diff
@@ -180,15 +180,6 @@ <h1>
...
-
-        {% if request.check_ownership(addon) %}
-          <p><a href="{{ addon.get_dev_url() }}#edit-addon-technical">{{ _('Change settings.') }}</a></p>
```
eviljeff, line 188: No existing tests to be removed/changed? 🐼

src/olympia/stats/templates/stats/stats.html
willdurand: It is not mentioned anymore, no. There is no notion of private vs public dashboard anymore. Would you still keep such a sentence? (Or rather why would we mention this here and not on other devhub pages for instance?)

eviljeff: Because the dashboard privacy has been a setting for 10+ years so existing developers might make assumptions.  Just asking though.

willdurand: That's a good suggestion. I'll add the sentence back.

PR #14567, opened 2020-06-08
add statsd.incr to upload_mlbf_to_remote_settings

### LINKED ISSUE #14359: Add statsd metrics to upload_mlbf_to_remote_settings cron too
We added statsd metrics to the `auto_import_blocklist` job; it would be cool to have the same for `upload_mlbf_to_remote_settings` so we can add that to the dashboard too.

eviljeff:
fixes #14359

### DIFF
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -239,11 +239,12 @@ class MLBF():
             return True
 
     def blocks_changed_since_previous(self, previous_base_mlbf):
+        self.fetch_blocked_json()
         try:
             # compare base with current blocks
             extras, deletes = self.generate_diffs(
-                previous_base_mlbf.blocked_json, self.fetch_blocked_json())
-            return bool(extras) or bool(deletes)
+                previous_base_mlbf.blocked_json, self.blocked_json)
+            return len(extras) + len(deletes)
         except FileNotFoundError:
             # when previous_base_mlfb._blocked_path doesn't exist
-            return True
+            return len(self.blocked_json)
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -239,12 +239,12 @@ class MLBF():
             return True
 
     def blocks_changed_since_previous(self, previous_base_mlbf):
-        self.fetch_blocked_json()
+        blocked_json = self.fetch_blocked_json()
         try:
             # compare base with current blocks
             extras, deletes = self.generate_diffs(
-                previous_base_mlbf.blocked_json, self.blocked_json)
+                previous_base_mlbf.blocked_json, blocked_json)
             return len(extras) + len(deletes)
         except FileNotFoundError:
             # when previous_base_mlfb._blocked_path doesn't exist
-            return len(self.blocked_json)
+            return len(blocked_json)
--- a/src/olympia/blocklist/tasks.py
+++ b/src/olympia/blocklist/tasks.py
@@ -208,6 +208,7 @@ def upload_filter(generation_time, is_base=True, upload_stash=False):
     if is_base:
         # clear the collection for the base - we want to be the only filter
         server.delete_all_records()
+        statsd.incr('blocklist.tasks.upload_filter.reset_collection')
     # Deal with possible stashes first
     if upload_stash:
         # If we have a stash, write that
@@ -217,6 +218,7 @@ def upload_filter(generation_time, is_base=True, upload_stash=False):
             'stash': mlbf.stash_json,
         }
         server.publish_record(stash_data)
+        statsd.incr('blocklist.tasks.upload_filter.upload_stash')
 
     # Then the bloomfilter
     data = {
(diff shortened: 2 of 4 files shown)


src/olympia/blocklist/tests/test_cron.py
```diff
@@ -24,11 +25,16 @@
...
+
+
 @freeze_time('2020-01-01 12:34:56')
 @override_switch('blocklist_mlbf_submit', active=True)
 class TestUploadToRemoteSettings(TestCase):
     def setUp(self):
-        addon_factory()
+        addon = addon_factory()
+        version_factory(addon=addon)
+        version_factory(addon=addon)
```
bobsilverberg, line 37: Is this meant to be called twice?

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -218,6 +267,19 @@ def test_force_base_option(self, should_reset_mock):
         # no stash because we're starting with a new base mlbf
         assert not os.path.exists(os.path.join(gen_path, 'stash.json'))
 
+        self.statsd_incr_mock.assert_has_calls([
+            mock.call(f'{STATSD_PREFIX}blocked_changed', 0),
+            mock.call(f'{STATSD_PREFIX}success'),
+            # 2nd execution
+            mock.call(f'{STATSD_PREFIX}blocked_changed', 0),
```
bobsilverberg, line 274: Will this actually check that this was called twice, or will this statement also return true if it was only called once? I.e., could the same call to the mock cause line 271 and line 274 to return true?

src/olympia/blocklist/tasks.py
```diff
@@ -229,6 +231,10 @@ def upload_filter(generation_time, is_base=True, upload_stash=False):
     with storage.open(mlbf.filter_path, 'rb') as filter_file:
         attachment = ('filter.bin', filter_file, 'application/octet-stream')
         server.publish_attachment(data, attachment)
+        statsd.incr('blocklist.tasks.upload_filter.upload_mlbf')
+    statsd.incr(
+        'blocklist.tasks.upload_filter.upload_mlbf.'
+        f'{"base" if is_base else "full"}')
```
bobsilverberg, line 237: Why is this "base" and "full"? Those names seem confusing to me. Would "base" and "update" be better?

src/olympia/blocklist/mlbf.py
```diff
@@ -239,11 +239,12 @@ def should_reset_base_filter(self, previous_base_mlbf):
...
             # compare base with current blocks
             extras, deletes = self.generate_diffs(
-                previous_base_mlbf.blocked_json, self.fetch_blocked_json())
-            return bool(extras) or bool(deletes)
+                previous_base_mlbf.blocked_json, self.blocked_json)
+            return len(extras) + len(deletes)
         except FileNotFoundError:
             # when previous_base_mlfb._blocked_path doesn't exist
-            return True
+            return len(self.blocked_json)
```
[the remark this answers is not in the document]
eviljeff: I want the function to return the numbers rather than just `True` or `False`.  The only scenario where it would be Falsey rather than Truthy would be when there is no previous base (there was an exception thrown, i.e. it's the first blocklist generated) _and_ there are 0 blocked guids.  We don't particularly support this case on AMO - it's unlikely you'd enable the cron when there are no blocked addons - but returning 0 (Falsey) is probably better anyway, because it wouldn't try to generate an empty bloomfilter.

src/olympia/blocklist/tasks.py
eviljeff: It's the same as the values of the constants BLOCKLIST_RECORD_MLBF_BASE an BLOCKLIST_RECORD_MLBF_UPDATE above in the file.  Rob wanted the attachment type constants to be "base and "full" (full, as in it's a full bloomfilter rather than a stash file, would would be the update) and I've use the same for consistency with the metadata in remote settings.

src/olympia/blocklist/tests/test_cron.py
eviljeff: yeah, it's generating 2 extra versions (so addon will have 3 versions total).  Doesn't matter much for the bloomfilter itself, but it was useful to see different numbers in the statsd calls.)

src/olympia/blocklist/tests/test_cron.py
eviljeff: yeah, `assert_has_calls` checks (by default, anyway) for exactly these calls, in exactly this order.

src/olympia/blocklist/mlbf.py
```diff
@@ -239,11 +239,12 @@ def should_reset_base_filter(self, previous_base_mlbf):
             return True
 
     def blocks_changed_since_previous(self, previous_base_mlbf):
+        self.fetch_blocked_json()
```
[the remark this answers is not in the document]
eviljeff: https://github.com/mozilla/addons-server/pull/14567/commits/a75b88055e2e9dd052b70035c1eafc3cc4e116a1

PR #14568, opened 2020-06-08
Update links in validation results to point to code manager

### LINKED ISSUE #14455: Update links in validation results to point to code manager
This is a follow-up to https://github.com/mozilla/addons-server/issues/13634.

There are links in the validation results, when viewed by an administrator, that currently point to the legacy file viewer. These should be updated to point to the new code manager.

I am filing this as a separate issue as these links are created via JS code which might take a bit of work to update and I don't want it to block https://github.com/mozilla/addons-server/issues/13634.

Here's a screenshot showing a couple of links on the validation results page:
Screenshot 2020-06-01 11 18 06

bobsilverberg:
Fixes #14455

### DIFF
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -48,6 +48,7 @@ from olympia.files.models import File, FileUpload
 from olympia.files.utils import parse_addon
 from olympia.reviewers.forms import PublicWhiteboardForm
 from olympia.reviewers.models import Whiteboard
+from olympia.reviewers.templatetags.code_manager import code_manager_url
 from olympia.reviewers.templatetags.jinja_helpers import get_position
 from olympia.reviewers.utils import ReviewHelper
 from olympia.users.models import DeveloperAgreementRestriction
@@ -661,7 +662,8 @@ def file_validation(request, addon_id, addon, file_id):
 
     validate_url = reverse('devhub.json_file_validation',
                            args=[addon.slug, file_.id])
-    file_url = reverse('files.list', args=[file_.id, 'file', ''])
+    file_url = code_manager_url(
+        'browse', addon_id=addon.pk, version_id=file_.version.pk)
 
     context = {'validate_url': validate_url, 'file_url': file_url,
                'file': file_, 'filename': file_.filename,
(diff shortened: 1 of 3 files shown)


src/olympia/devhub/views.py
```diff
@@ -661,7 +662,8 @@ def file_validation(request, addon_id, addon, file_id):
 
     validate_url = reverse('devhub.json_file_validation',
                            args=[addon.slug, file_.id])
-    file_url = reverse('files.list', args=[file_.id, 'file', ''])
+    file_url = code_manager_url(
+        'browse', addon_id=addon.pk, version_id=file_.version.pk)
```
eviljeff, line 666: it's a shame you can't just pass in a `Version` instance and get the `.pk` and `.addon.pk` from there but I suppose it's used in other places where you only have ids.

static/js/zamboni/validator.js
```diff
@@ -253,7 +253,7 @@ function initValidator($doc) {
             }
 
             if (this.fileURL) {
-                var url = this.fileURL + file;
+                var url = this.fileURL + '?path=' + file;
```
eviljeff, line 256: it's lucky this works.  I assume you've tested it still works?  (`file` is the same as code manage expects)

bobsilverberg: Yes, I tested it manually and it seems to work. We actually have a Python function that will build this URL for us, including the `path` param, but this is JS, and the `file` value is only available at this point where we're processing the json file, so a Python function doesn't help us at this point.

I could have abstracted this into a function, but I figured the less I touched the JS code the better! 😉

src/olympia/devhub/views.py
bobsilverberg: That's an interesting idea. It might be possible, but we already have other code calling and using the `code_manager_url` function, and I wouldn't want to add changes to that to this patch. It's possible you're right, and we might be able to do this, but I'm not sure it's worth a follow-up. WDYT?

eviljeff: I looked and it was used in a template used for scanner results that seemed to be literal version ids so you'd need to go back into the view code and check that the version objects are available (having to do a db lookup just for this is a waste).  So, yeah, probably not worth logging for a follow-up.

PR #14569, opened 2020-06-09
Drop view_source property from addon model

### LINKED ISSUE #12842: Drop view_source property from Addon model

junngo:
Ref #12842 

Drop view_source property from addon model and related code.


src/olympia/addons/migrations/0011_auto_20200609_0522.py
```diff
@@ -0,0 +1,17 @@
...
+
+    dependencies = [
+        ('addons', '0010_auto_20200604_1319'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='addon',
+            name='view_source',
+        ),
```
junngo, line 16: I have a question.
When I made migrations file, `public_stats` filed is occurred that is Remove Field.
But I deleted line about `public_stats`.
Should I include `public_stats` property?

junngo:
Thanks for reviewing. I have learned a lot :)
I submitted PR again about step 1.

junngo:
Yes, I will.
After land this PR, I'll submit PR for step 2(`Remove the field from the models, stop using it in code`).
Thanks :)

Issue #14582, opened 2020-06-09
Developers of unlisted add-ons cannot see their stats 


ioanarusiczki:
STR:
1. I tried to access the beta stats for an unlisted addon with a developer account (the author of the unlisted addon): 
https://addons-dev.allizom.org/en-US/firefox/addon/154b2ca27eff421e93d7/statistics/beta/?last=30

Expected result:
Page should be available.

Actual result:
I receive a 404

Notes:
This is also reproducible for mozilla.com developer accounts.

ioanarusiczki:
@willdurand This issue is no longer reproducible on AMO dev - FF76(Win10). 

I've noticed that:
- "Statistics" link on the Manage my submission page is not present for unlisted addons. Maybe this should be added too. 
- On AMO stage only admins can access stats for unlisted addons.

willdurand:
> On AMO stage only admins can access stats for unlisted addons.

Only for unlisted add-ons?

ioanarusiczki:
@willdurand On AMO stage with a developer account I can see my stats for listed add-ons but for unlisted add-ons I don't have the "View Statistics Dashboard" link on the left. 
With an admin account I have access to all stats. For my unlisted addon I am redirected to the beta stats.

Issue #4092, opened 2016-11-29
Don't disable statistics dashboard for dev-disabled add-ons


dhemmert:
### Describe the problem and steps to reproduce it:
Team,  the add-on SalesforceOverlay (Developed by Nate Snow and managed by myself) has been removed from public visibility.  However, I needed to get metrics related to the add-on like "How many downloads? and Where were the downloads from?" but since it was removed the dashboard I had before is no longer accessiblie.

### What happened?
Dashboard statistics/metrics not available anymore

### What did you expect to happen?
Wasn't aware that this data would be removed immediately and would like to recover if possible.

### Anything else we should know?
https://addons.mozilla.org/en-US/firefox/addon/salesforceoverlay/versions/

jvillalobos:
You can temporarily enable your add-on, get what you need from the stats dashboard, and then disable it again. The data isn't gone, it's just inaccessible in the current state.

I'll keep this bug open in case anyone wants to take it. Seems like an easy thing to fix.

EnTeQuAk:
Should this dashboard maybe still be accessible by the add-on developer so that the enabling back and forth isn't needed anymore? Adding this permission check shouldn't be too heavy, could be a good first bug to work on in the backend views.

wagnerand:
The triage team thinks: This is probably not a good first bug

diox:
I do think it's a good first bug for python/django contributors, the changes required are minimal.

To sum up, the issue is that we'd like to let developers (or people with specific `STATS_VIEW` permission) view statistics for add-ons that have been disabled by user. However, we don't want general public to view them.

A contributor would need to look at the views in the `stats` app and do the following:

1. Figure out what is preventing user disabled (`disabled_by_user` property) add-ons from showing their statistics and change that to allow them.
2. Change the permission check so that when we display stats because the add-on is configured to show its stats publicly (`public_stats` property), we also check that the add-on itself is public (Adding a `is_public()` call for that specific condition)
3. Change tests accordingly. There is already a test covering this, and it has comments explaining what is currently tested so what needs to be changed should be straightforward.

Akhila499:
Hello Mentors, 

I am Akhila and I am outreachy applicant for March 2020. I would like to work on this issue if it is considered as firstbug/outreachy eligible issue. I am currently reading through the Mozilla addon contribution page (https://wiki.mozilla.org/Add-ons/Contribute/Code). 

@diox @eviljeff Kindly help me in getting up to speed with the issue.

[...]

PR #14583, opened 2020-06-09
Allow authors to see unlisted stats

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14582

Fixes https://github.com/mozilla/addons-server/issues/4092

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -165,6 +165,10 @@ class AddonQuerySet(BaseQuerySet):
         """Get valid, enabled add-ons only"""
         return self.filter(self.valid_q(amo.VALID_ADDON_STATUSES))
 
+    def not_disabled_by_mozilla(self):
+        """Get all add-ons not disabled by Mozilla."""
+        return self.exclude(status=amo.STATUS_DISABLED)
+
     def listed(self, app, *status):
         """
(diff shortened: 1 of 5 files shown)


src/olympia/stats/decorators.py
```diff
@@ -13,7 +13,7 @@ def wrapper(request, addon_id=None, *args, **kw):
         if acl.action_allowed(request, permissions.STATS_VIEW):
             qs = Addon.objects.all
         else:
-            qs = Addon.objects.valid
+            qs = Addon.objects.not_disabled_by_mozilla
```
eviljeff, line 16: I think I've have just done `Addon.objects.exclude(...)` here directly.

src/olympia/addons/models.py
```diff
@@ -165,6 +165,10 @@ def valid(self):
         """Get valid, enabled add-ons only"""
         return self.filter(self.valid_q(amo.VALID_ADDON_STATUSES))
 
+    def not_disabled_by_mozilla(self):
+        """Get all add-ons not disabled by Mozilla."""
+        return self.exclude(status=amo.STATUS_DISABLED)
```
eviljeff, line 170: If you're adding to AddonManager it needs a test in addons/tests/test_models.py

Issue #14430, opened 2020-05-28
Reset client-side stats cache when retrieving fresh data


willdurand:
I've found myself looking at stale stats data because of we cache data in the session storage. Let's make sure it won't be a problem for the beta.

**For QA:** this problem occurs when switching from old stats to beta stats for example. It can be reproduced with the following STRs:

1. navigate to: https://addons-dev.allizom.org/en-US/firefox/addon/some-addon/statistics/usage/os/?last=30
2. open the devtools > storage > session storage
3. there should be a key named `statscache-618396`, update its value with the following json. If there is no key, please create a new one.

```
[...]

PR #14594, opened 2020-06-10
Cache bust stats session cache

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14430

---

Note: this is a patch I'd like to cherry-pick because users might want to look at the diff between old and beta stats and the cache might show mixed data.

### DIFF
--- a/static/js/impala/stats/manager.js
+++ b/static/js/impala/stats/manager.js
@@ -14,7 +14,7 @@ z.StatsManager = (function() {
     // strictly the same and the cache might contain mixed data so we
     // cache-bust the (session) cache when users switch between the two stats
     // features.
-    var isBeta = $primary.attr("data-is-beta") === 'True';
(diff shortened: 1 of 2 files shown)


static/js/impala/stats/manager.js
```diff
@@ -8,13 +8,24 @@ z.StatsManager = (function() {
     var STATS_VERSION = '2011-12-12';
     var PRECISION = 2;
 
+    var $primary = $(".primary");
+
+    // Users can navigate between "old" and "beta" stats pages. Data isn't
+    // strictly the same and the cache might contain mixed data so we
+    // cache-bust the (session) cache when users switch between the two stats
+    // features.
+    var isBeta = $primary.attr("data-is-beta") === 'True';
```
eviljeff, line 17: `$primary.data("is-beta")`

PR #14605, opened 2020-06-11
Addon queryset tidy up of unused methods

### LINKED ISSUE #14604: AddonQuerySet cleanup
Just cleanup of some old unused code I found while looking into a different issue

eviljeff:
fixes #14604

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -153,13 +153,9 @@ class AddonQuerySet(BaseQuerySet):
             return self.filter(slug=val)
         return self.filter(id=val)
 
-    def enabled(self):
-        """Get add-ons that haven't been disabled by their developer(s)."""
-        return self.filter(disabled_by_user=False)
-
     def public(self):
-        """Get public add-ons only"""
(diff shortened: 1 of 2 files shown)


src/olympia/addons/models.py
```diff
@@ -153,13 +153,9 @@ def id_or_slug(self, val):
...
 
-    def enabled(self):
-        """Get add-ons that haven't been disabled by their developer(s)."""
-        return self.filter(disabled_by_user=False)
-
     def public(self):
-        """Get public add-ons only"""
-        return self.filter(self.valid_q([amo.STATUS_APPROVED]))
+        """Get reviewed add-ons only"""
+        return self.filter(self.valid_q(amo.REVIEWED_STATUSES))
```
bobsilverberg, line 158: The issue says this is just removing unused code, but it looks like the logic of this has changed as well. Can you explain that?

eviljeff: in the constants, `REVIEWED_STATUSES = (STATUS_APPROVED,)` so it's the same.

PR #14606, opened 2020-06-11
When notifying about auto-approval delays, only consider one version per addon

### LINKED ISSUE #13974: Send email to developer if add-on submission is pending review after a few hours

diox:
Only the latest auto-approvable version must be considered for each add-on, so that new submissions "reset" the waiting period if they aren't approved.

Fixes #13974 (again)

### DIFF
--- a/src/olympia/reviewers/management/commands/notify_about_auto_approve_delay.py
+++ b/src/olympia/reviewers/management/commands/notify_about_auto_approve_delay.py
@@ -1,6 +1,7 @@
 from datetime import datetime, timedelta
 
 from django.core.management.base import BaseCommand
+from django.db.models import Max
 
 from olympia.reviewers.utils import ReviewHelper
 from olympia.versions.models import Version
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/management/commands/notify_about_auto_approve_delay.py
```diff
@@ -17,21 +18,36 @@ def fetch_versions_waiting_for_approval_for_too_long(self):
...
-        return (
+        qs = (
+            # Base queryset with auto-approvable versions. Default ordering
+            # is reset to make the GROUP BY work.
             Version.objects
                    .auto_approvable()
-                   .filter(created__lt=waiting_period)
                    .exclude(**exclude_kwargs)
-                   .order_by('created')
-                   .distinct()
```
diox, line 33: The `distinct()` is no longer needed now that we're grouping by add-on in the subquery and then filtering by pk.

src/olympia/reviewers/management/commands/notify_about_auto_approve_delay.py
```diff
@@ -17,21 +18,36 @@ def fetch_versions_waiting_for_approval_for_too_long(self):
...
+                   .order_by()
```
eviljeff, line 37: is this `.order_by()` with no args doing anything?  There's a default order on the Version model.

src/olympia/reviewers/management/commands/notify_about_auto_approve_delay.py
```diff
@@ -17,21 +18,36 @@ def fetch_versions_waiting_for_approval_for_too_long(self):
...
         )
+        # Get only the latest version for each add-on from the auto-approvable
+        # versions.
+        latest_per_addon = qs.values('addon').annotate(latest_pk=Max('pk'))
```
eviljeff, line 41: This doesn't take into account channels - there could be an upload as a listed version and an unlisted version.  As the unlisted version doesn't supersede the listed version (or visa versa) should emails be sent for both?

diox:
> I'm not sure I entirely understand the scenario being fixed - version 1.0 is `STATUS_AWAITING_REVIEW`; version 2.0 is uploaded so version 1.0 will be set to `STATUS_DISABLED`. Are we sending emails for both version 1.0 and 2.0 currently?

I will fix the test to reflect this, but the scenario was for unlisted versions (which do not get disabled at new submission)

src/olympia/reviewers/management/commands/notify_about_auto_approve_delay.py
diox: It's precisely to avoid the default ordering, which messes up the group by.

src/olympia/reviewers/management/commands/notify_about_auto_approve_delay.py
diox: My idea was to avoid sending multiple emails, regardless of what the versions are.

src/olympia/reviewers/management/commands/notify_about_auto_approve_delay.py
eviljeff: ah, didn't know it did that.

src/olympia/reviewers/management/commands/notify_about_auto_approve_delay.py
eviljeff: I guess that fine - as long as there's nothing channel specific in the email sent.

Issue #14608, opened 2020-06-11
Remove stats pages for dictionaries and language packs


willdurand:
See: https://github.com/mozilla/addons-server/issues/14533

---

We want to remove stats pages for langpacks and dictionaries (temporarily) in beta mode only (because the beta will become the default after July 9, 2020).

The idea is therefore to:

- remove beta stats pages for langpacks and dictionaries
- hide links to beta in "Manage my submission" and the DevHub "edit" add-on pages
- link to old stats pages in "Manage my submission" and the DevHub "edit" add-on pages (because users need access to these pages before we sunset the old stats)
- hide "beta" message shown on old stats pages for langpacks and dictionaries

ioanarusiczki:
@willdurand I verified the statistics links for lang packs and dictionaries on AMO dev:
- the beta stats are not available for developers or an admin. 
- old stats can be accessed from edit and manager my submission page without messages about beta.

Is it alright that the frontend link "Visit stats dashboard" is not available for the developers or the admin?

willdurand:
> Is it alright that the frontend link "Visit stats dashboard" is not available for the developers or the admin?

Yes, because, as soon as beta becomes the default, there won't be stats for langpacks and dicts for some time.

ioanarusiczki:
@willdurand Ok, I am marking this issue verified. - FF76(Win10).

PR #14612, opened 2020-06-11
Remove stats pages for langpacks/dictionaries in beta

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14608

---

The issue above should have all the information to explain this patch. In short, we only have beta/new stats for extensions and static themes. And we have to link to old stats for other add-ons so that users have access to the old pages until we sunset them.


src/olympia/devhub/templates/devhub/addons/listing/item_actions.html
```diff
@@ -47,7 +47,7 @@ <h5>{{ _('Actions') }}</h5>
       </li>
     {% endif %}
   {% endif %}
-  {% if addon.status != amo.STATUS_DISABLED and waffle.flag('beta-stats') %}
+  {% if addon.type in [amo.ADDON_EXTENSION, amo.ADDON_STATICTHEME] and addon.status != amo.STATUS_DISABLED and waffle.flag('beta-stats') %}
```
eviljeff, line 50: can we define a constant for `[amo.ADDON_EXTENSION, amo.ADDON_STATICTHEME]` in amo and reference that instead?  I feel like we'll end up missing a check somewhere when it's expanded to other addon types and/or removed at a later point.

src/olympia/stats/views.py
```diff
@@ -336,7 +337,7 @@ def check_stats_permission(request, addon, beta):
                 user.email.endswith('@mozilla.com') or
                 waffle.flag_is_active(request, 'beta-stats')
             )
-        )
+        ) or addon.type not in [ADDON_EXTENSION, ADDON_STATICTHEME]
```
eviljeff, line 340: can you refactor this `if` out into some boolean variables to make the ending `if` more straightforward.

src/olympia/stats/views.py
```diff
@@ -355,12 +356,15 @@ def stats_report(request, addon, report, beta=False):
...
     view = get_report_view(request)
+    hide_beta = addon.type not in [ADDON_EXTENSION, ADDON_STATICTHEME]
+
     return render(
         request,
         'stats/reports/%s.html' % report,
         {
             'addon': addon,
             'beta': beta,
+            'hide_beta': hide_beta,
```
eviljeff, line 367: `hide_beta` and `link_to_old_stats` are the same, right?  Does a common variable name make sense?

willdurand: yeah, they are but they convey different intents so that's why I used two different names :/

src/olympia/devhub/templates/devhub/addons/listing/item_actions.html
willdurand: Good idea.

Issue #14549, opened 2020-06-05
Have the account delete method also delete appropriate associated content the user created


eviljeff:
Spun off from #13896 to make the changes to the API delete endpoint for accounts so we're deleteing the same thing both from account deletion requests from FxA and from AMO directly.

addons-frontend issue for copy change - mozilla/addons-frontend#9467

eviljeff:
Forgot about the restriction we specifically added in the API

AlexandraMoga:
I've verified the user account deletion from AMO and from FxA (with the `fxa-account-delete` waffle enabled) and, for both cases, the following user data is deleted*:
- profile data - display name, homepage, location, picture, biography, notification preferences
- ratings
- collections
- addons

*except for the content mentioned in issue #14701 and #14711

PR #14622, opened 2020-06-12
don't clear email and fxa on delete; reuse existing record on reregister

### LINKED ISSUE #14613: Don't clear email and fxa_id from UserProfile on delete
We currently clear the email address and fxa_id from a UserProfile instance on delete, and rely on logging to keep track of them.  It will make querying easier if we just retain the email address and fxa_id, and only clear them later in line with our data retention policies.

eviljeff:
fixes #14613 

Needs to land at the same time as the fix for #14549 or the "new" account registered by the user will contain content from the "old" account.


diox:
> Needs to land at the same time as the fix for #14549 or the "new" account registered by the user will contain content from the "old" account.

There is no PR yet for that issue right ?

eviljeff:
> 
> 
> > Needs to land at the same time as the fix for #14549 or the "new" account registered by the user will contain content from the "old" account.
> 
> There is no PR yet for that issue right ?

no

diox:
So do we really want to merge this now, knowing we aren't deleting the content yet ?

eviljeff:
> So do we really want to merge this now, knowing we aren't deleting the content yet ?

We don't.

eviljeff:
should be merged with #14678 and mozilla/addons-frontend#9477

PR #14638, opened 2020-06-15
Support page link in content review page

### LINKED ISSUE #14219: Include the Support Site at the top of the content review page

During a content review most of the information is on the content review page, but the support site is only found on the product page.

- The URL should be checked to ensure it is a suitable link, but it is easy to forget to visit the product page to check it. (Especially as most add-ons don't list a support site.)
- When I want more information about the product and there is no Website entry it is an extra step to see if there is a support site link.

### What happened?
The content review page doesn't show the support site.

If there is a support site list it on the content review page below where the Website link goes.

An example of an add-on with a support site but not website: https://reviewers.addons.mozilla.org/en-US/reviewers/review-content/ray-webpage-screenshot

linzichao:
Fixes #14219

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/addon_details_box.html
+++ b/src/olympia/reviewers/templates/reviewers/addon_details_box.html
@@ -64,6 +64,15 @@
               </td>
             </tr>
           {% endif %}
+          {% if addon.support_url %}
+            <tr>
+              <th>{{ _('Support Website') }}</th>
+              <td>
+                <strong><a href="{{ addon.support_url|external_url }}">{{
+                  addon.support_url }}</a></strong>
+              </td>
+            </tr>
+          {% endif %}
           {% if addon.has_listed_versions() %}
           <tr>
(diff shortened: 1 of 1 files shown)


src/olympia/reviewers/templates/reviewers/addon_details_box.html
```diff
@@ -64,6 +64,15 @@ <h2>{{ addon_name }}</h2>
               </td>
             </tr>
           {% endif %}
+          {% if addon.support_url %}
+            <tr>
+              <th>{{ _('Support Website') }}</th>
+              <td>
+                <strong><a href="{{ addon.support_url }}">{{
```
eviljeff, line 71: like the homepage link above, this url needs the `external_url` filter adding to it (it wraps the link in redirection service for safety)

linzichao: Sure

linzichao:
My bad. I exclude the file from the commit.

Issue #14609, opened 2020-06-11
Use download stats for updating `addon.average_daily_users` for dictionaries and langpacks


willdurand:
We won't have usage stats for langpacks and dictionaries _via_ Telemetry yet so let's update `addon.average_daily_users` with the download stats instead.

willdurand:
@diox we have a waffle switch to either use the BigQuery data or the old way of updating the `average_daily_users` values: https://github.com/mozilla/addons-server/blob/b22a64eab4f1336d4c6aa493fe364fe5041a3558/src/olympia/addons/cron.py#L40-L43

I am thinking about adding logic similar to what we have in `update_addon_download_totals()` (filtering on the addon-on `type` too) and update the `d` array right after. WDYT?

willdurand:
@eviljeff @diox now that I landed the patch for this issue and that I fixed it (twice), it works well but I realize that it might not be needed at all :|

While the `Addon.total_downloads` property isn't exposed by the API, we still compute it in a CRON task (frequently). The value for this property is almost the same as the one we now use for `addon.average_daily_users` (for add-ons that don't have stats in BigQuery). I am wondering if this is actually needed?

Instead, we could expose the `total_downloads` property and use it in addons-frontend. That would mean add-ons without stats in BigQuery would not get their `average_daily_users` updated anymore. Is that a problem? Where do we use this value in addons-server?

eviljeff:
iirc, not exposing total downloads was a deliberate decision.  Why aren't we using weekly downloads?

jvillalobos:
I believe we used total downloads for search engines before, but I could be wrong. If it's easier to use weekly downloads, that's fine too.

ioanarusiczki:
@willdurand I verified on AMO dev the following lang packs and dictionaries:
acholi-ug-language-pack
afrikaans-language-pack
chinese-simplified-zh-cn-la
afrikaans-spell-checker
albanian-dictionary

- in all cases I could see in the overview a number of downloads (which should be the total number of downloads) and ADU which are equal. The reason is [explained](https://github.com/mozilla/addons-server/issues/14609#issuecomment-644916552)

![information](https://user-images.githubusercontent.com/33448286/85008771-14162e80-b166-11ea-929d-0b0c38f5fe2d.png)

- the number I see in [Overview](https://addons-dev.allizom.org/en-US/firefox/addon/chinese-simplified-zh-cn-la/statistics/?last=30) is also shown by the frontend.

PR #14639, opened 2020-06-15
Use download stats for updating `addon.average_daily_users` for dictionaries and langpacks

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14609

### DIFF
--- a/src/olympia/addons/cron.py
+++ b/src/olympia/addons/cron.py
@@ -38,7 +38,18 @@ def update_addon_average_daily_users():
 
     kwargs = {'id_field': 'pk'}
     if waffle.switch_is_active('use-bigquery-for-addon-adu'):
-        d = get_addons_and_average_daily_users_from_bigquery()
+        # BigQuery does not have data for add-ons with type other than those in
+        # `ADDON_TYPES_WITH_STATS` so we use download counts instead.
+        # See: https://github.com/mozilla/addons-server/issues/14609
+        d = list(
+            Addon.objects
+            .exclude(type__in=amo.ADDON_TYPES_WITH_STATS)
+            .annotate(count=Sum('downloadcount__count'))
+            .values_list('guid', 'count')
+            .order_by('guid')
+            .all()
+        )
+        d += get_addons_and_average_daily_users_from_bigquery()
         # BigQuery stores GUIDs, not AMO primary keys.
         kwargs['id_field'] = 'guid'
     else:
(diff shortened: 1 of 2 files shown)


src/olympia/addons/cron.py
```diff
@@ -39,6 +39,17 @@ def update_addon_average_daily_users():
...
+        # BigQuery does not have data for add-ons with type other than those in
+        # `ADDON_TYPES_WITH_STATS` so we use download counts instead.
+        # See: https://github.com/mozilla/addons-server/issues/14609
+        d += (
+            Addon.objects
+                .exclude(type__in=amo.ADDON_TYPES_WITH_STATS)
+                .annotate(count=Sum('downloadcount__count'))
+                .values_list('guid', 'count')
+                .order_by('guid')
+                .all()
```
willdurand, line 51: @diox I don't like the fact that we retrieve all results first but I don't have a nicer way. WDYT?

diox: That's probably fine, but I'd do the `Addon.objects.[...]` call first. The reason is, you might lose the connection to the database if the big query call takes too much time, so it'd be safer if the 2 were swapped.

src/olympia/addons/cron.py
```diff
@@ -38,7 +38,18 @@ def update_addon_average_daily_users():
...
-        d = get_addons_and_average_daily_users_from_bigquery()
+        # BigQuery does not have data for add-ons with type other than those in
+        # `ADDON_TYPES_WITH_STATS` so we use download counts instead.
+        # See: https://github.com/mozilla/addons-server/issues/14609
+        d = list(
+            Addon.objects
+            .exclude(type__in=amo.ADDON_TYPES_WITH_STATS)
+            .annotate(count=Sum('downloadcount__count'))
+            .values_list('guid', 'count')
+            .order_by('guid')
```
eviljeff, line 49: does the response need to be ordered?  guid ordering is pretty arbitrary.

src/olympia/addons/cron.py
```diff
@@ -38,7 +38,18 @@ def update_addon_average_daily_users():
...
+            .all()
```
eviljeff, line 50: don't need the `.all()` after you've done other queryset operations to the model manager.

src/olympia/addons/cron.py
```diff
@@ -38,7 +38,18 @@ def update_addon_average_daily_users():
 
     kwargs = {'id_field': 'pk'}
     if waffle.switch_is_active('use-bigquery-for-addon-adu'):
-        d = get_addons_and_average_daily_users_from_bigquery()
+        # BigQuery does not have data for add-ons with type other than those in
+        # `ADDON_TYPES_WITH_STATS` so we use download counts instead.
+        # See: https://github.com/mozilla/addons-server/issues/14609
+        d = list(
```
eviljeff, line 44: Can we take the chance to make a change to have a longer variable name please?  Single letter vars can be problematic.

src/olympia/addons/cron.py
willdurand: I think we can get rid of that indeed

PR #14640, opened 2020-06-15
raven -> sentry-sdk

### LINKED ISSUE #14636: upgrade from raven to sentry-python
https://github.com/getsentry/sentry-python is the new integration from Sentry; https://github.com/getsentry/raven-python is their legacy client.  Upgrading might fix exceptions/errors not propogating properly from log.exception and log.error calls (#12446)

eviljeff:
fixes #14636 ... probably?

### DIFF
--- a/src/olympia/amo/tests/test_settings.py
+++ b/src/olympia/amo/tests/test_settings.py
@@ -6,7 +6,7 @@ import pytest
 
 from django.conf import settings
 
-from olympia.lib.settings_base import get_raven_release
+from olympia.lib.settings_base import get_sentry_release
 
 
 @pytest.mark.parametrize('key', (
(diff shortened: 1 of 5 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -1814,21 +1814,37 @@ def get_raven_release():
...
+                head = str(fp.read()).strip()
+
+            if head.startswith('ref: '):
+                head = head[5:]
+                revision_file = os.path.join(ROOT, '.git', *head.split('/'))
+            else:
+                return head
+            with open(revision_file) as fh:
+                version = str(fh.read()).strip()
+        except IOError:
```
eviljeff, line 1828: this is based on what `raven.fetch_git_sha` did

src/olympia/amo/tests/test_settings.py
```diff
@@ -48,15 +48,15 @@ def test_raven_release_config():
         os.remove(version_json)
 
     # by default, if no version.json exists it simply fetches a git-sha
-    assert len(get_raven_release()) == 40
+    assert len(get_sentry_release()) == 40
 
     # It fetches `version` from the version.json
     with open(version_json, 'w') as fobj:
         fobj.write(json.dumps({
             'version': '2018.07.19'
```
bobsilverberg, line 56: Nit: Can we make this a const so that it doesn't require a double hardcoding? The same goes for `commit = '1111111'`, below.

eviljeff:
I've been attempting write a test that would trigger sentry (the thinking being that responses would stop the connection and we can check for that) but no joy - even with a literal `capture_exception(e)` call (`from sentry_sdk import capture_exception`).  I'm out of ideas.

@diox What do you think?

Issue #14650, opened 2020-06-16
Errors while updating average daily users (when bigquery waffle is turned on)


willdurand:
While testing the cron task, I noticed two errors:

- BigQuery can return `null` (count) values
- We have add-ons without GUIDs, which should be excluded

ioanarusiczki:
@willdurand I left a comment here https://github.com/mozilla/addons-server/issues/14428#issuecomment-645288854

If there's anything else I can check for this issue please let me know.

PR #14651, opened 2020-06-16
Fix errors during ADU update

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14650

### DIFF
--- a/src/olympia/stats/tests/test_utils.py
+++ b/src/olympia/stats/tests/test_utils.py
@@ -236,3 +236,15 @@ GROUP BY addon_id"""
 
         returned_results = get_addons_and_average_daily_users_from_bigquery()
         assert returned_results == [(1, 123), (2, 456)]
+
+    @mock.patch('google.cloud.bigquery.Client')
+    def test_skips_null_values(self, bigquery_client_mock):
+        results = [
+            self.create_bigquery_row({'addon_id': 1, 'count': 123}),
+            self.create_bigquery_row({'addon_id': 2, 'count': None}),
+        ]
+        client = self.create_mock_client(results=results)
+        bigquery_client_mock.from_service_account_json.return_value = client
+
+        returned_results = get_addons_and_average_daily_users_from_bigquery()
+        assert returned_results == [(1, 123)]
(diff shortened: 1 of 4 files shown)


src/olympia/stats/tests/test_utils.py
```diff
@@ -236,3 +236,15 @@ def test_returned_results(self, bigquery_client_mock):
 
         returned_results = get_addons_and_average_daily_users_from_bigquery()
         assert returned_results == [(1, 123), (2, 456)]
+
+    @mock.patch('google.cloud.bigquery.Client')
+    def test_skips_null_values(self, bigquery_client_mock):
+        results = [
+            self.create_bigquery_row({'addon_id': 1, 'count': 123}),
+            self.create_bigquery_row({'addon_id': 2, 'count': None}),
+        ]
```
eviljeff, line 245: isn't this testing the wrong thing? It's `None` values in the database rather the bigquery that are failing.

willdurand: BigQuery can return `count=None` and the DB can return `guid=None`

willdurand: We should probably have a test case for the DB change but simply adding an add-on without a GUID does not seem to work 🤷

eviljeff: > 
> 
> BigQuery can return `count=None` and the DB can return `guid=None`

right, sure, but the change above is only related to the database query, but this test only (appears to) mock bigquery response.

eviljeff: > 
> 
> We should probably have a test case for the DB change but simply adding an add-on without a GUID does not seem to work 🤷

what do you mean doesn't work?  You can't `.update(guid=None)`?

willdurand: Yeah, that works (or `addon_factory(guid=None)`) but for some reason, we don't get an error raised by the subtask in this cron task..

willdurand: I added another test case.

PR #14658, opened 2020-06-16
Add command to fix langpacks that were erroneously set as compatible with '*'

### LINKED ISSUE #14653: Fix language packs pointing to `*` as their max AppVersion
In https://bugzilla.mozilla.org/show_bug.cgi?id=1646016 we noticed some language packs versions are linked to a max `AppVersion` of `*`. This cause the langpacks to be returned for Firefox versions they don't correspond to.

We need to write a script to fix that data by querying langpacks that have versions that are linked to `*` in the database, extract relevant XPIs, and link to the correct max `AppVersion` instead.

diox:
Fixes #14653

### DIFF
--- a/src/olympia/addons/management/commands/fix_langpacks_with_max_version_star.py
+++ b/src/olympia/addons/management/commands/fix_langpacks_with_max_version_star.py
@@ -1,0 +1,53 @@
+from django.core.management.base import BaseCommand
+
+from olympia import amo
+from olympia.core.logger import getLogger
+from olympia.versions.compare import version_dict
+from olympia.versions.models import AppVersion, Version
+
+log = getLogger('z.fix_langpacks_with_max_version_star')
+
+
+class Command(BaseCommand):
+    help = 'Fix language packs that have a max version compatibility set to *'
+
+    def find_affected_langpacks(self):
+        qs = Version.unfiltered.filter(
+            addon__type=amo.ADDON_LPAPP, apps__max__version='*').distinct()
+        return qs
+
+    def fix_max_appversion_for_version(self, version):
+        for app in (amo.FIREFOX, amo.ANDROID):
+            if app not in version.compatible_apps:
+                log.info(
+                    'Version %s for addon %s min version is not compatible '
+                    'with %s, skipping this version for that app.',
+                    version, version.addon, app.pretty)
+                continue
+            if version.compatible_apps[app].max.version != '*':
+                log.info(
+                    'Version %s for addon %s max version is not "*" for %s '
+                    'app, skipping this version for that app.',
+                    version, version.addon, app.pretty)
+                continue
+            min_appversion_str = version.compatible_apps[app].min.version
+            max_appversion_str = '%d.*' % version_dict(
+                min_appversion_str)['major']
+            log.warn(
+                'Version %s for addon %s min version is %s for %s app, '
+                'max should be changed to %s instead of *',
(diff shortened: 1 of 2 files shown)


src/olympia/addons/management/commands/fix_langpacks_with_max_version_star.py
```diff
@@ -0,0 +1,53 @@
...
+            addon__type=amo.ADDON_LPAPP, apps__max__version='*').distinct()
+        return qs
+
+    def fix_max_appversion_for_version(self, version):
+        for app in (amo.FIREFOX, amo.ANDROID):
+            if app not in version.compatible_apps:
+                log.info(
+                    'Version %s for addon %s min version is not compatible '
+                    'with %s, skipping this version for that app.',
+                    version, version.addon, app.pretty)
```
eviljeff, line 25: this is going to be a pretty noisy log

src/olympia/addons/management/commands/fix_langpacks_with_max_version_star.py
```diff
@@ -0,0 +1,53 @@
...
+                min_appversion_str)['major']
+            log.warn(
+                'Version %s for addon %s min version is %s for %s app, '
+                'max should be changed to %s instead of *',
+                version, version.addon, min_appversion_str, app.pretty,
+                max_appversion_str)
+            max_appversion = AppVersion.objects.get(
+                application=app.id, version=max_appversion_str)
+            version.compatible_apps[app].max = max_appversion
+            version.compatible_apps[app].save()
```
eviljeff, line 44: nit, but `version.compatible_apps[app].update(max=max_version)`?

src/olympia/addons/management/commands/fix_langpacks_with_max_version_star.py
```diff
@@ -0,0 +1,53 @@
...
+                    'Version %s for addon %s max version is not "*" for %s '
+                    'app, skipping this version for that app.',
+                    version, version.addon, app.pretty)
+                continue
+            min_appversion_str = version.compatible_apps[app].min.version
+            max_appversion_str = '%d.*' % version_dict(
+                min_appversion_str)['major']
+            log.warn(
+                'Version %s for addon %s min version is %s for %s app, '
+                'max should be changed to %s instead of *',
```
eviljeff, line 38: should this be "max _will_ be changed to ..."?

src/olympia/addons/management/commands/fix_langpacks_with_max_version_star.py
diox: Yeah - I figured we wanted to be extra sure we wouldn't be missing anything, and if we were what the reason was

src/olympia/addons/management/commands/fix_langpacks_with_max_version_star.py
diox: I didn't do that because... `.update()` doesn't exist on this model (!) It doesn't inherit from our `ModelBase`.

eviljeff: 🤯

src/olympia/addons/management/commands/fix_langpacks_with_max_version_star.py
diox: 221c3ec7585639dfc8fa5bd5da2b7ee572b62bd4

PR #14659, opened 2020-06-16
Provide a validation.json url via the reviewers app

### LINKED ISSUE #14655: Loading validation.json from code-manager is generating a CORS error

1. With a reviewer account, open the browse page for any add-on - see for example https://code.addons-dev.allizom.org/en-US/browse/619739/versions/1691228/

### What happened?
The code overview is not loaded - i.e. it's stuck on  _Loading linter information..._

The code overview is loaded 

- reproduced only on -dev with both the browse and the compare views
 
image

bobsilverberg:
Fixes #14655 

Note that this also removes support for the CORS headers from the `devhub.json_file_validation` view as that was only needed for code manager, and now this view is no longer used by code manager. Perhaps we should also back out the changes from https://github.com/mozilla/addons-server/commit/9e421a5c5927b8ca2de8eccb3cba60b74f779adb which allowed for access to validation.json for deleted add-ons? My understanding is that that was also only needed for code manager, so perhaps those changes should be undone for the `devhub` view?

### DIFF
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -1156,8 +1156,10 @@ def theme_background_images(request, version_id):
 
 
 @any_reviewer_required
-def json_file_validation(request, addon_id, file_id):
-    addon = get_object_or_404(Addon.unfiltered.id_or_slug(addon_id))
+@reviewer_addon_view_factory
+def json_file_validation(request, addon, file_id):
+    if not acl.is_reviewer(request, addon):
+        raise PermissionDenied
     file = get_object_or_404(File, version__addon=addon, id=file_id)
     try:
         result = file.validation
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -12,6 +12,7 @@ from django.core.exceptions import PermissionDenied
 from django.core.paginator import Paginator
 from django.db.models import Prefetch, Q
 from django.db.transaction import non_atomic_requests
+from django.http import JsonResponse
 from django.shortcuts import get_object_or_404, redirect
 from django.utils.http import urlquote
 from django.utils.translation import ugettext
@@ -50,6 +51,7 @@ from olympia.api.permissions import (
 from olympia.constants.reviewers import REVIEWS_PER_PAGE, REVIEWS_PER_PAGE_MAX
 from olympia.devhub import tasks as devhub_tasks
 from olympia.discovery.models import DiscoveryItem
+from olympia.files.models import File
 from olympia.ratings.models import Rating, RatingFlag
 from olympia.reviewers.forms import (
     AllAddonSearchForm, MOTDForm, PublicWhiteboardForm, QueueSearchForm,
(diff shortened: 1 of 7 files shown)


src/olympia/reviewers/views.py
```diff
@@ -1153,6 +1155,26 @@ def theme_background_images(request, version_id):
     return version.get_background_images_encoded(header_only=False)
 
 
+@any_reviewer_required
+def json_file_validation(request, addon_id, file_id):
```
diox, line 1159: Could you use `reviewer_addon_view_factory` ? This does the `addon = get_object_or_404(...)`  bit for you (It should already handle deleted add-ons) but more importantly adds a check for unlisted reviewer if the add-on has no listed version. It would be nice to have a test around this as well, making sure we aren't exposing unlisted data to non-unlisted reviewers.
The devhub view indirectly had that as well, through `dev_required` which was using `addon_view_factory`.

diox: Bonus: inside the function, I would call `acl.is_reviewer(request, addon)`, and bail if its false - that way we don't allow theme developers access to extensions and vice-versa.

bobsilverberg: I tried to implement this, but I'm running into something I don't understand. When I run a test that was passing before making this change, it's now returning a `301` instead of a `200`. I traced it back to [this code](https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/decorators.py#L37-L42) in `addon_view` which is being triggered because `addon.slug != addon_id`. I expect `addon.slug != addon_id` because `addon_id` is the `pk`. That code redirects to a URL using `slug` instead of `pk`, but we want to be able to use `pk` in order to support deleted add-ons, so I'm not really sure what is going on here.

diox: Which tests ? Redirecting to the slug version is normal if there is one. The only time there isn't one is for deleted add-ons (which have no slug)

bobsilverberg: Ah, I see. My basic test isn't for a deleted add-on, but I was checking for a `200`. I guess before introducing `reviewer_addon_view_factory` it did return a `200`, but now it's returning a `301`. I had an additional test which was for a deleted add-on, and it's still returning `200`, which is what had me confused. So, if there is a slug, then we redirect, which makes sense, and means I have to change the test to check for the `301` instead.

bobsilverberg: Actually, scratch that. I'll just use `slug` for all of the tests *except* the test for a deleted add-on.

diox: Note that if possible (do you know early enough ?) you should use that logic (slug except for deleted add-on) in code-manager as well.

bobsilverberg: Everything is `addonId` in code manager, I believe. The API docs say "int: addon_id", so that's what we've gone with. Are those docs inaccurate? I'm not sure we have a slug to begin with anyway.

bobsilverberg:
Thanks for the feedback and help, @diox. This is ready for another look.

src/olympia/reviewers/views.py
diox: Thinking about this more, we _need_ to avoid the redirect, because of the perf cost associated with it. 

Since we don't have the slug in code-manager at the time, that means the API needs to support add-on ids without redirecting. `reviewer_addon_view_factory` might not be the right approach then, though you still need to handle access to unlisted versions. So:
- Stop using `reviewer_addon_view_factory` and go back to `addon = get_object_or_404(Addon.unfiltered.id_or_slug(addon_id))` (well technically we only need the add-on id)
- `Move the `is_reviewer` check below the code that fetches the `File`.
- Check for `file.version.channel`, if it's unlisted, you should call `check_unlisted_addons_reviewer()` and not `is_reviewer`

This is also better because it checks if the specific version the user is accessing is listed or not before letting them access it.

bobsilverberg:
Thanks @diox. This is ready for another look.

bobsilverberg:
The failure is an ES intermittent.

PR #14660, opened 2020-06-16
Do not allow trademarks in add-on name whatsoever

### LINKED ISSUE #14656: Remove `content-optimization` waffle dependency for trademark checks.

The trademark check in https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/utils.py#L32-L40 allows "... for Firefox" if the waffle switch `content-optimization` is turned off. We'd like to change that so it's disallowed regardless of the waffle switch.

STR:
1. Have an existing add-on.
2. Go to the developer hub and change the name to "Something for Firefox" or "Something for Mozilla".

### What happened?

The add-on is successfully renamed.

An error should occur: _Add-on names cannot contain the Mozilla or Firefox trademarks._

wagnerand:
Fix #14656.

### DIFF
--- a/src/olympia/addons/tests/test_utils_.py
+++ b/src/olympia/addons/tests/test_utils_.py
@@ -9,8 +9,6 @@ from django.conf import settings
 from django.forms import ValidationError
 from django.utils.encoding import force_text
 
-from waffle.testutils import override_switch
-
 from olympia import amo
 from olympia.addons.utils import (
     build_webext_dictionary_from_legacy,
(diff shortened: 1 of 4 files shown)


src/olympia/addons/utils.py
```diff
@@ -30,16 +28,7 @@ def _check(name):
         name = normalize_string(name, strip_punctuation=True).lower()
 
         for symbol in amo.MOZILLA_TRADEMARK_SYMBOLS:
-            if waffle.switch_is_active('content-optimization'):
```
willdurand, line 33: Should we have a migration to delete this waffle switch?

src/olympia/addons/tests/test_utils_.py
```diff
@@ -46,11 +45,11 @@
     ('Firefox add-on for Firefox', False, 'foo@bar.com', False),
     ('Firefox add-on for Firefox', True, 'foo@mozilla.com', False),
     ('Foobarfor Firefox', False, 'foo@bar.com', False),
-    ('Better Privacy for Firefox!', True, 'foo@bar.com', False),
+    ('Better Privacy for Firefox!', False, 'foo@bar.com', False),
     ('Firefox awesome for Mozilla', False, 'foo@bar.com', False),
     ('Firefox awesome for Mozilla', True, 'foo@mozilla.org', False),
 
-    # And with the content optimization waffle onL
+    # And with the content optimization waffle on
```
willdurand, line 52: Is this comment still accurate?

src/olympia/addons/utils.py
wagnerand: No, there are still other features gated by that switch.

src/olympia/addons/tests/test_utils_.py
wagnerand: Yes, I was just fixing the typo, presumably.

willdurand: I dont see the waffle being used in utils.py?

wagnerand: I didn't want to change the tests too much, assuming we would have a cleanup one that waffle has been in prod for a while and gets removed. If you prefer, I can clean up this test.

willdurand: It would be nice to keep the test in sync with the function.

This comment suggests that we're using a waffle in the function but that doesn't seem to be the case.

wagnerand: 67c2d08152

wagnerand:
Unrelated test failures:

FAILED src/olympia/lib/es/tests/test_commands.py::TestIndexCommand::test_reindexation_starting_from_one_addon

FAILED src/olympia/lib/es/tests/test_commands.py::TestIndexCommand::test_reindexation_with_wipe

Issue #14637, opened 2020-06-15
Handle no `.git/description` file in `AddonGitRepository`


willdurand:
The `.git/description` file should be present because git creates it (and it's an internal git file) but we noticed a few errors with potentially old git repos that did not have this file (or maybe something else prevented this file to be written on disk).

We should handle this case to avoid "file not found" errors.

ioanarusiczki:
@willdurand If there's something qa can test for this issue, let me know.
Thanks!

willdurand:
@ioanarusiczki it is tricky to verify so I am marking the issue as "qa: not needed"

willdurand:
@ioanarusiczki actually, the only "thing" to verify is to ensure that git extraction still works as before.

PR #14667, opened 2020-06-17
Do not fail when .git/description is missing

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14637

### DIFF
--- a/src/olympia/git/utils.py
+++ b/src/olympia/git/utils.py
@@ -206,7 +206,11 @@ class AddonGitRepository(object):
 
     @property
     def is_recent(self):
-        if not self.is_extracted:
+        git_description_path = os.path.join(
+            self.git_repository_path,
+            self.GIT_DESCRIPTION
+        )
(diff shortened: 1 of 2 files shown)


src/olympia/git/utils.py
```diff
@@ -206,7 +206,11 @@ def is_extracted(self):
 
     @property
     def is_recent(self):
-        if not self.is_extracted:
+        git_description_path = os.path.join(
+            self.git_repository_path,
+            self.GIT_DESCRIPTION
+        )
+        if not self.is_extracted or not os.path.exists(git_description_path):
             return False
```
bobsilverberg, line 214: What if it is recent *and* there's no description file? You mentioned in the issue it's possible that something prevented it from being written.  What's the impact of returning `False` in that scenario?

willdurand: We use this method to determine if we can remove a git repo because of an error or if we shouldn't remove it to prevent an endless loop (think the git repo being deleted over and over again).

Issue #14652, opened 2020-06-16
Handle nested special git files


willdurand:
We should better handle nested special git files, especially when special git files are located inside other special git files:

```
/.git/.gitmodules
```

wagnerand:
Also `.git/.git/`

ioanarusiczki:
@willdurand Please add some STR for qa if we can help here.
Thanks!

willdurand:
@ioanarusiczki sorry about that, added now. thanks

ioanarusiczki:
@willdurand I deleted the extension of the file (.txt for example) and uploaded the following:

for .git/.gitattributes

https://code.addons-dev.allizom.org/en-US/browse/619945/versions/1691571/
https://code.addons-dev.allizom.org/en-US/browse/619948/versions/1691574/

for .git/.git/.gitconfig

https://code.addons-dev.allizom.org/en-US/browse/619946/versions/1691572/
https://code.addons-dev.allizom.org/en-US/browse/619947/versions/1691573/

willdurand:
@ioanarusiczki looks good

PR #14674, opened 2020-06-18
Handle nested special git files

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14652


src/olympia/git/utils.py
```diff
@@ -414,10 +414,8 @@ def _commit_through_worktree(self, file_obj, message, author, branch):
 
             # Fetch all files and strip the absolute path but keep the
             # `extracted/` prefix
-            files = get_all_files(
-                worktree.extraction_target_path,
-                worktree.path,
-                '')
+            files = get_all_files(worktree.extraction_target_path,
+                                  worktree.path, '')
```
eviljeff, line 418: is this an auto-format from black?  It's much worse!

PR #14678, opened 2020-06-18
always delete related content on UserProfile.delete

### LINKED ISSUE #14549: Have the account delete method also delete appropriate associated content the user created

eviljeff:
fixes #14549


eviljeff:
should be merged with #14622 and mozilla/addons-frontend#9477

PR #14703, opened 2020-06-23
send email on user.delete()

### LINKED ISSUE #14512: Send an email to user after their account is deleted
We've wanted to send an email to users when their account is deleted but never got around to do it. The copy for the email exists and had been validated at the time: https://docs.google.com/document/d/1Zuos9dZINSaX4vj4eNu0jT18DMc_HkB_Nm84x3iL5_g/edit

Open questions:
- Should we do this for banned users as well ? Even users banned in bulk ?
- Should we send it when we're deleting the AMO account because the FxA account is deleted after doing https://github.com/mozilla/addons-server/issues/13896 ?

eviljeff:
fixes #14512

### DIFF
--- a/src/olympia/users/models.py
+++ b/src/olympia/users/models.py
@@ -18,6 +18,7 @@ from django.core import validators
 from django.core.cache import cache
 from django.core.files.storage import default_storage as storage
 from django.db import models
+from django.template import loader
 from django.utils import timezone
 from django.utils.crypto import salted_hmac
 from django.utils.encoding import force_text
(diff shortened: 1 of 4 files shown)


src/olympia/users/models.py
```diff
@@ -515,17 +516,30 @@ def ban_and_disable_related_content_bulk(cls, users, move_files=False):
...
+        email_msg = template.render(context={
+            'site_url': site_url,
+            'name': self.name})
+        return {
+            'subject': f'Your account on {site_url} has been deleted',
+            'message': email_msg,
+            'recipient_list': email_to,
+        }
+
+    def delete(self, addon_msg='', send_delete_email=True):
```
willdurand, line 532: No more `hard` param?

eviljeff: We didn't use it anywhere so I took the opportunity to remove a code path.  (we could still do `UserProfile.objects.get(...).delete()` if we particularly wanted to hard delete in a test)

PR #14704, opened 2020-06-23
move waffle switch check to delete_user_event task

### LINKED ISSUE #14702: don't make waffle checks in monitor_fxa_sqs
we run `monitor_fxa_sqs` as a long-running command to monitor the sqs queue for FxA, looking for email changes and deletions.  The deletion monitoring is currently behind a waffle switch.  

Unfortunately django can lose it's database connection during long running commands, meaning the `waffle.switch_is_active` check errors and raises an exception. 

Long term we won't need the switch, but short term we do, so we should move the check to the task where a new database connection is opened each time.

eviljeff:
fixes #14702

### DIFF
--- a/src/olympia/accounts/tasks.py
+++ b/src/olympia/accounts/tasks.py
@@ -1,6 +1,8 @@
 import functools
 from datetime import datetime
 
+from waffle import switch_is_active
+
 import olympia.core.logger
 
 from olympia.amo.celery import task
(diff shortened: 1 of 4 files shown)


src/olympia/accounts/tasks.py
```diff
@@ -48,6 +50,11 @@ def primary_email_change_event(profile, changed_date, email):
...
-    log.info(
-        'Account pk [%s] deleted from FxA on %s' % (user.id, deleted_date))
+    if switch_is_active('fxa-account-delete'):
+        user.delete(addon_msg='Deleted via FxA account deletion')
+        log.info(
+            'Account pk [%s] deleted from FxA on %s' % (user.id, deleted_date))
+    else:
+        log.info(
+            f'Skipping deletion from FxA for account [{user.id}] because '
+            'waffle inactive')
```
diox, line 60: It matters less for things that don't go in sentry, so that doesn't block this PR, but we should try to always let `logging` to the formatting to allow for grouping - So we should do logging formatting with `log.info('foo %s', bar)` and not `log.info('foo %s' % bar)`, `log.info('foo {bar}')` or `log.info('foo {}'.format(bar))` where possible.

PR #14713, opened 2020-06-24
clear more fields on UserProfile on delete

### LINKED ISSUE #14701: The Occupation field is not cleared on account deletion
Follow up for #14549 

1. Log in to AMO -dev
2. Edit your user profile, making sure that the Occupation field is filled in too
3. Save your changes
4. Delete the profile
5. Log in with the same email again and check the content present on the profile page

### What happened?
The Occupation field still has the data from the previous profile

The Occupation field is empty 

This is an example of a deleted user as seen in the admin user profile (you can see that Occupation is not empty) - https://addons-internal.dev.mozaws.net/en-US/admin/models/users/userprofile/10641952/change/

eviljeff:
fixes #14701


src/olympia/users/models.py
```diff
@@ -536,7 +547,8 @@ def delete(self, addon_msg='', send_delete_email=True):
         log.info(f'User ({self}: <{self.email}>) is being anonymized.')
         if send_delete_email:
             email = self._prepare_delete_email()
-        self._anonymize()
+        self.anonymize_users((self, ))
```
bobsilverberg, line 550: Took me a bit to figure out what's going on here. So `(self, )` is a tuple that is being passed into the `users` argument of `anonymize_users`? Is that trailing comma necessary?

src/olympia/users/tests/test_models.py
```diff
@@ -105,7 +106,7 @@ def test_delete(self):
         assert user.auth_id
         assert user.auth_id != old_auth_id
         assert user.fxa_id == '0824087ad88043e2a52bd41f51bbbe79'
-        assert user.display_name is None
+        assert user.display_name == ''
```
bobsilverberg, line 109: Why did this change from `None` to `''`?

eviljeff:
> This looks good. I have a couple of questions. Also, I don't see a test that verifies that the specific problem (that `occupation` was not being cleared) has been fixed.

https://github.com/mozilla/addons-server/pull/14713/commits/46b2a46840f9140fc622386620666f5efc7a07d2

PR #14717, opened 2020-06-24
Change the validation.json endpoint to a true API endpoint

### LINKED ISSUE #14655: Loading validation.json from code-manager is generating a CORS error

bobsilverberg:
Fixes #14655

### DIFF
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -414,7 +414,7 @@ class AddonBrowseVersionSerializer(
     def get_validation_url_json(self, obj):
         return absolutify(drf_reverse(
             'reviewers-addon-json-file-validation',
-            request=self._context.get('request'),
+            request=self.context.get('request'),
             args=[
                 obj.addon.pk, obj.current_file.id
             ]))
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -9,6 +9,7 @@ import pygit2
 
 from rest_framework import serializers
 from rest_framework.exceptions import NotFound
+from rest_framework.reverse import reverse as drf_reverse
 
 from django.core.cache import cache
 from django.utils.functional import cached_property
@@ -411,9 +412,12 @@ class AddonBrowseVersionSerializer(
                 ) if key in entry}
 
     def get_validation_url_json(self, obj):
-        return absolutify(reverse('reviewers.json_file_validation', args=[
-            obj.addon.pk, obj.current_file.id
-        ]))
+        return absolutify(drf_reverse(
+            'reviewers-addon-json-file-validation',
+            request=self._context.get('request'),
+            args=[
+                obj.addon.pk, obj.current_file.id
+            ]))
 
     def get_validation_url(self, obj):
         return absolutify(reverse('devhub.file_validation', args=[
--- a/src/olympia/reviewers/tests/test_serializers.py
+++ b/src/olympia/reviewers/tests/test_serializers.py
@@ -43,8 +43,17 @@ class TestFileInfoSerializer(TestCase):
         self.version = self.addon.current_version
         self.file = self.addon.current_version.current_file
 
+        # Set up the request to support drf_reverse
+        api_version = api_settings.DEFAULT_VERSION
+        self.request = APIRequestFactory().get('/api/%s/' % api_version)
+        self.request.versioning_scheme = (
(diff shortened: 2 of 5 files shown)


src/olympia/reviewers/serializers.py
```diff
@@ -411,9 +412,12 @@ def _trim_entry(self, entry):
                 ) if key in entry}
 
     def get_validation_url_json(self, obj):
-        return absolutify(reverse('reviewers.json_file_validation', args=[
-            obj.addon.pk, obj.current_file.id
-        ]))
+        return absolutify(drf_reverse(
+            'reviewers-addon-json-file-validation',
+            request=self._context.get('request'),
```
diox, line 417: We have some code using `self._context` but we should change it (no idea why it was done that way), `self.context` is the supported way of accessing the context in a serializer.

src/olympia/reviewers/tests/test_serializers.py
```diff
@@ -403,7 +405,16 @@ def setUp(self):
         assert self.addon.current_version.release_notes
         self.version = self.addon.current_version
 
+        # Set up the request to support dnf_reverse
```
diox, line 408: Typo

```suggestion
        # Set up the request to support drf_reverse
```

src/olympia/reviewers/views.py
```diff
@@ -1315,6 +1290,36 @@ def allow_resubmission(self, request, **kwargs):
...
+            raise http.Http404
+        response = JsonResponse({
+            'validation': result.processed_validation,
+            'error': None,
+        })
+        # See: https://github.com/mozilla/addons-server/issues/11048
+        response['Access-Control-Allow-Origin'] = settings.CODE_MANAGER_URL
+        response['Access-Control-Allow-Methods'] = 'GET, OPTIONS'
+        response['Access-Control-Allow-Headers'] = 'Content-Type'
+        response['Access-Control-Allow-Credentials'] = 'true'
```
diox, line 1320: I don't think any of those are necessary any more ? The default for the API should be `Access-Control-Allow-Origin: *`, and with being an API now, it's going to use token-based auth w/ the `Authorization` header, not the session cookie.

bobsilverberg:
@diox This is a P1 because code manager will be broken on Thursday if we don't fix this issue.

bobsilverberg:
Note that this needs a code manager change as well, which I am currently working on.

src/olympia/reviewers/views.py
```diff
@@ -1315,6 +1290,29 @@ def allow_resubmission(self, request, **kwargs):
             status_code = status.HTTP_409_CONFLICT
         return Response(status=status_code)
 
+    @action(
+        detail=True,
+        methods=['get'],
+        permission_classes=[AllowAnyKindOfReviewer],
+        url_path=r'file/(?P<file_id>[^/]+)/validation\.json')
```
diox, line 1297: Drop the `.json` - The URL is going to end with a slash anyway.

src/olympia/reviewers/tests/test_serializers.py
```diff
@@ -58,8 +60,16 @@ def test_raises_without_version(self):
             serializer.data
 
     def test_can_access_version_from_parent(self):
+        # Set up the request to support drf_reverse
+        api_version = api_settings.DEFAULT_VERSION
+        request = APIRequestFactory().get('/api/%s/' % api_version)
+        request.versioning_scheme = (
+            api_settings.DEFAULT_VERSIONING_CLASS()
+        )
+        request.version = api_version
```
diox, line 69: Could this be done in `setUp` ? You do that below for other test classes.

bobsilverberg: It could be. I just figured because only this test needs it, I might as well just add it here, but I suppose if other tests might need it in the future it would be nice to have it in `setUp`. I'll move it.

bobsilverberg:
This needs a fix on code manager as well, which is at https://github.com/mozilla/addons-code-manager/pull/1628, but code manager is broken without this anyway, and this will allow for local testing against dev, so I've merged this.

PR #14727, opened 2020-06-25
Migrations catch up

diox:
Just the result of `manage.py makemigrations` on the whole repos.


src/olympia/hero/migrations/0007_auto_20200625_1139.py
```diff
@@ -0,0 +1,24 @@
...
+        ),
+    ]
```
eviljeff, line 24: can we drop this migration file?  Lisa's making changes to it and I don't want to add extra complications.  (Plus when it's done, both of those fields will be gone anyway)

diox: https://github.com/mozilla/addons-server/pull/14727/commits/7663e3a3213080fb1d5d36e9eb3331998231fce2

PR #14739, opened 2020-06-26
Ignore data coming from bigquery for langpacks and dictionaries for ADU

### LINKED ISSUE #14732: Invalid ADU for langpacks/dicts after task is executed
The ADU value of langpacks/dictionaries is computed based on the total number of downloads but something is wrong in -prod right now. We have very low values.

This is the django query we do for computing the total downloads in the ADU update task: https://github.com/mozilla/addons-server/blob/917825b911533a6600d79cf02a5fdb6f9a21b7d5/src/olympia/addons/cron.py#L46-L51

This is the django query used in a different task: https://github.com/mozilla/addons-server/blob/917825b911533a6600d79cf02a5fdb6f9a21b7d5/src/olympia/addons/cron.py#L82-L85

diox:
Fixes #14732

### DIFF
--- a/src/olympia/addons/cron.py
+++ b/src/olympia/addons/cron.py
@@ -42,15 +42,20 @@ def update_addon_average_daily_users():
         # BigQuery does not have data for add-ons with type other than those in
         # `ADDON_TYPES_WITH_STATS` so we use download counts instead.
         # See: https://github.com/mozilla/addons-server/issues/14609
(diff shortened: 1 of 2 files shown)


diox:
We could also just ask the data team to investigate why there is some leftover data and wipe it, but this seems safer, less likely to regress. Let me know what you think when you're back from PTO.

PR #14763, opened 2020-06-28
Remove the view_source field from the models, stop using it in code

junngo:
Ref #12842 

I removed the view_source field from the models, stop using it in code.

I'm not sure about test code[0] because I don't know Bug-1353788 history.
But I think the view_source dependent code has been deleted.
Let me know If there is anything needed modify or wrong it.

[0]
src/olympia/files/tests/test_views.py


diox:
Looks good thanks. Next week we should be able to run the migration to drop the column entirely!

PR #14764, opened 2020-06-28
RE: Add code to resize and otherwise adjust images uploaded via the admin 

### LINKED ISSUE #14475: Add code to resize and otherwise adjust images uploaded via the admin
Supports https://github.com/mozilla/addons-server/issues/14471
Depends on #14472

The code that uploads images likely needs to do some manipulation of those images in order to ensure they are of a reasonable size and are consistent. This might just be resizing, but perhaps other things will need to be done too. There was also discussion of generating thumbnails for the images at the same time.

Specific dimensions desired for the images and thumbnails can be added to this issue.

xlisachan:
Fixes #14475 

The following changes have been made with this pull request:
- Images uploaded to Primary Hero Images are resized upon saving to two dimensions: 960x640 and 150x120
- Images displayed in Primary Hero Images are thumbnails of the primary hero images
- Within Discovery Items, the primary hero images displayed are the thumbnails as well
- `/discovery/tests/test_admin.py` has been revised to check that the primary hero images exist in two sizes, as well as checks their dimensions

### DIFF
--- a/src/olympia/discovery/tests/test_admin.py
+++ b/src/olympia/discovery/tests/test_admin.py
@@ -7,6 +7,7 @@ from olympia.amo.tests import TestCase, addon_factory, user_factory
 from olympia.amo.tests.test_helpers import get_uploaded_file
 from olympia.amo.urlresolvers import django_reverse, reverse
 from django.conf import settings
+from django.core.files.images import get_image_dimensions
 from olympia.discovery.models import DiscoveryItem
 from olympia.hero.models import (
     PrimaryHero, PrimaryHeroImage, SecondaryHero, SecondaryHeroModule)
@@ -274,7 +275,8 @@ class TestDiscoveryAdmin(TestCase):
     def test_can_add_primary_hero_with_discovery_edit_permission(self):
         addon = addon_factory(name=u'BarFöo')
         item = DiscoveryItem.objects.create(addon=addon)
-        image = PrimaryHeroImage.objects.create(custom_image='foo.jpg')
+        uploaded_photo = get_uploaded_file('transparent.png')
+        image = PrimaryHeroImage.objects.create(custom_image=uploaded_photo)
         self.detail_url = reverse(
             'admin:discovery_discoveryitem_change', args=(item.pk,)
         )
@@ -577,17 +579,19 @@ class TestPrimaryHeroImageAdmin(TestCase):
         assert self.list_url in response.content.decode('utf-8')
 
     def test_can_list_with_discovery_edit_permission(self):
-        PrimaryHeroImage.objects.create(custom_image='föo.jpg')
+        uploaded_photo = get_uploaded_file('transparent.png')
+        PrimaryHeroImage.objects.create(custom_image=uploaded_photo)
         user = user_factory()
         self.grant_permission(user, 'Admin:Tools')
         self.grant_permission(user, 'Discovery:Edit')
         self.client.login(email=user.email)
         response = self.client.get(self.list_url, follow=True)
         assert response.status_code == 200
-        assert 'föo.jpg' in response.content.decode('utf-8')
+        assert 'transparent.png' in response.content.decode('utf-8')
 
     def test_can_edit_with_discovery_edit_permission(self):
-        item = PrimaryHeroImage.objects.create(custom_image='föo.jpg')
+        uploaded_photo = get_uploaded_file('transparent.png')
+        item = PrimaryHeroImage.objects.create(custom_image=uploaded_photo)
         self.detail_url = reverse(
             'admin:discovery_primaryheroimageupload_change', args=(item.pk,)
         )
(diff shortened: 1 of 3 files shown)


src/olympia/hero/models.py
```diff
@@ -97,16 +99,33 @@ class PrimaryHeroImage(ModelBase):
     def __str__(self):
         return f'{self.custom_image}'
 
+    @property
+    def preview_url(self):
+        (path, fn) = os.path.split(self.custom_image.path)
```
eviljeff, line 104: `os.path.basename`

src/olympia/hero/models.py
```diff
@@ -97,16 +99,33 @@ class PrimaryHeroImage(ModelBase):
...
+                self.preview_url)
             )
         else:
             return None
     preview_image.short_description = "Image"
     preview_image.allow_tags = True
 
+    def save(self, *args, **kwargs):
+        super().save(*args, **kwargs)
+        (path, fn) = os.path.split(self.custom_image.path)
```
eviljeff, line 120: `os.path.split` returns a bytes object rather than a unicode string?

src/olympia/hero/models.py
```diff
@@ -97,16 +99,33 @@ class PrimaryHeroImage(ModelBase):
...
+        dest_thumb = path + b'/thumbs/' + fn
+
+        size_full = (960, 640)
+        size_thumb = (150, 120)
+
+        resize_image(self.custom_image, self.custom_image.path, size_full)
```
eviljeff, line 126: the first arg here, `self.custom_image` is just the image path, isn't it?  (i.e. the same as the second arg).  Tbh I don't know why we enforce that the source and destination be different in `resize_image` - you might be introducing a bug you have to fix later 😉  I'd recommend you write to a temporary file and then move it back but I'm not going to require it.

src/olympia/discovery/tests/test_admin.py
```diff
@@ -650,6 +667,16 @@ def test_can_add_with_discovery_edit_permission(self):
         assert item.custom_image == 'hero-featured-image/transparent.png'
         assert os.path.exists(os.path.join(
             settings.MEDIA_ROOT, 'hero-featured-image', 'transparent.png'))
+        assert os.path.exists(os.path.join(
+            settings.MEDIA_ROOT, 'hero-featured-image',
+            'thumbs', 'transparent.png'))
```
eviljeff, line 672: lets use a larger image to test the resize works for the full size image too.  There are a few larger images in the same folder.

src/olympia/hero/models.py
xlisachan: Yep, I had to convert it from a bytes object to unicode when I needed it in the preview_url.

src/olympia/hero/models.py
xlisachan: The two arguments are different. The first argument, `self.custom_image`, has an `ImageFieldFile` value, while the second argument (self.custom_image.path) has a bytes object value (see example below). Please let me know if I should still write to a temporary file.
Screen Shot 2020-06-30 at 1 16 35 PM

xlisachan: Definitely don't want bugs 😬 I'm not 100% sure about my revision though, can you please check if I incorporated the temporary file correctly?

xlisachan:
Thanks for the feedback, @eviljeff! The following updates have been made:
- Moved model's `save` to `PrimaryHeroImageAdmin.save_model`
- Resized the image file to a thumbnail first
- Updated filename retrieval in `def preview_url` by utilizing `os.path.basename`
- Used `tempfile` for filename in `save_model`
- Revised `test_admin` to check that it resizes a larger image (2400 × 1800 image dimension)

As mentioned above, I'm not sure if I incorporated the temporary file correctly, can you please take a look at it? Also, one of the CI checks failed, but it passed when I went to run it in the terminal. Is there a way to refresh the checks?

src/olympia/hero/admin.py
```diff
@@ -58,6 +66,21 @@ class Media:
...
+
+        size_thumb = (150, 120)
+        size_full = (960, 640)
+
+        img = Image.open(obj.custom_image)
+        f = tempfile.NamedTemporaryFile(dir=settings.TMP_PATH)
+        img.save(f, 'png')
+
+        resize_image(f.name, dest_thumb, size_thumb)
+        resize_image(f.name, obj.custom_image.path, size_full)
```
eviljeff, line 82: How about:
```suggestion
        resize_image(obj.custom_image.path, dest_thumb, size_thumb)
        with tempfile.NamedTemporaryFile(dir=settings.TMP_PATH) as tmp:
            resize_image(obj.custom_image.path, tmp.name, size_full)
            copy_stored_file(tmp.name, obj.custom_image.path)
```
(untested!)

eviljeff:
> Also, one of the CI checks failed, but it passed when I went to run it in the terminal. Is there a way to refresh the checks?

ah, that's a common intermittently failing test.  Try logging into Travis with your github account, you may have access to restart jobs/builds?

src/olympia/hero/admin.py
xlisachan: Ah I think I see now, so I should've written `resize_image(obj.custom_image.path, f.name, size_full)` instead and then moved it with `copy_stored_file(f.name, obj.custom_image.path)`, correct?

xlisachan:
`def save_model` has been updated to correctly use `tempfile`, @eviljeff! Please review when you have a moment, thanks!

src/olympia/hero/admin.py
eviljeff: yeah.  Using `with` means that the temporary file is closed (and deleted) at the end of the `with` block so we don't need to worry about remembering to close it (and handle exceptions leaving the file open, etc).

PR #14780, opened 2020-06-29
Disable renovate for Python

### LINKED ISSUE #14778: Disable renovate for Python
We use Dependabot for Python dependencies, not Revovate, so we need to disable renovate support.

bobsilverberg:
Fixes #14778 

I decided to leave the `pip_requirements` in there in case we ever decide we want to use Renovate for Python again. Simply adding the `"python": { "enabled": false }` is supposed to be enough to disable Python support.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -20,6 +20,9 @@
       "rangeStrategy": "replace"
     }
   ],
+  "python": {
+    "enabled": false
+  },
   "pip_requirements": {
     "fileMatch": [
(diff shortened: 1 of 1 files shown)


bobsilverberg:
> For the record: [docs.renovatebot.com/python/#disabling-python-support](https://docs.renovatebot.com/python/#disabling-python-support) documents this key.

Yep, that's where I found it! 😄

PR #14781, opened 2020-06-29
Upgrade UglifyJS to 3.10.0, fixing compatibility with that version

### LINKED ISSUE #14779: Upgrade Uglify-js to 3.10.0 or higher and fix our assets tests coverage
Upgrading to uglify-js 3.10.0 caused https://github.com/mozilla/addons-server/issues/14726. It didn't break any tests.
We are reverting this upgrade, but then we should:
- Figure out why this didn't work and fix the issue
- Figure out why this didn't break any tests and fix the tests

diox:
Also added a test proving we do compress js files correctly.

Fixes #14779

### DIFF
--- a/src/olympia/amo/management/commands/compress_assets.py
+++ b/src/olympia/amo/management/commands/compress_assets.py
@@ -198,7 +198,7 @@ class Command(BaseCommand):
         """Run the proper minifier on the file."""
         if ftype == 'js' and hasattr(settings, 'UGLIFY_BIN'):
             opts = {'method': 'UglifyJS', 'bin': settings.UGLIFY_BIN}
-            run_command('{uglify} -v -o {target} {source} -m'.format(
+            run_command('{uglify} -o {target} {source} -m'.format(
                 uglify=opts['bin'],
                 target=file_out,
                 source=file_in))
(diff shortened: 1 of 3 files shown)


src/olympia/amo/management/commands/compress_assets.py
```diff
@@ -198,7 +198,7 @@ def _minify(self, ftype, file_in, file_out):
         """Run the proper minifier on the file."""
         if ftype == 'js' and hasattr(settings, 'UGLIFY_BIN'):
             opts = {'method': 'UglifyJS', 'bin': settings.UGLIFY_BIN}
-            run_command('{uglify} -v -o {target} {source} -m'.format(
+            run_command('{uglify} -o {target} {source} -m'.format(
```
bobsilverberg, line 201: You decided not to bother with `--verbose`?

diox: I tried it and there wasn't any difference in the output, weirdly. I decided I had already spent too much time on this...

src/olympia/amo/tests/test_commands.py
```diff
@@ -115,6 +125,43 @@ def test_compress_assets_correctly_fetches_static_images(settings, tmpdir):
...
+    except FileNotFoundError:
+        pass
+
+    # Capture output to avoid it being logged and allow us to validate it
+    # later if needed
+    out = io.StringIO()
+
+    # Now run compress and collectstatic
+    call_command('compress_assets', force=True, stdout=out)
+    call_command('collectstatic', interactive=False, stdout=out)
```
bobsilverberg, line 158: I assume we need to call both of these to get the files to be written?

diox: Technically, only the first one is needed to have files written, but `collecstatic` moves static files from all apps to the final location where they will be served from, so it's useful to have this here as well to test the full config/behavior.

PR #14795, opened 2020-06-30
Export CSV with null columns

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14796

### DIFF
--- a/src/olympia/stats/tests/test_views.py
+++ b/src/olympia/stats/tests/test_views.py
@@ -921,10 +921,12 @@ class TestRenderCSV(TestCase):
             {'data': {'a': 4}, 'count': 4, 'date': '2020-01-02'},
         ]
 
+        # Simulates how other views are rendering CSV content.
         stats, fields = views.csv_fields(series)
         response = views.render_csv(request=RequestFactory().get('/'),
                                     addon=addon_factory(),
                                     stats=stats,
                                     fields=fields)
 
-        assert '\n,a\r\n1,2\r\n0,4\r\n' in force_text(response.content)
+        assert ('\r\n'.join([',a', '1,2', '0,4']) in
+                force_text(response.content))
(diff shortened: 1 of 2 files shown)


src/olympia/stats/views.py
```diff
@@ -84,7 +84,7 @@ def csv_fields(series):
         row['data'].update(count=row['count'], date=row['date'])
     # Sort the fields before returning them - we don't care much about column
     # ordering, but it helps make the tests stable.
-    return rv, sorted(fields)
+    return rv, sorted(fields, key=lambda field: '' if not field else field)
```
eviljeff, line 87: How does changing the sort order fix the issue?  It doesn't change the content of the fields

src/olympia/stats/tests/test_views.py
```diff
@@ -912,3 +912,19 @@ def test_skips_none_key(self):
...
+            {'data': {'a': 4}, 'count': 4, 'date': '2020-01-02'},
+        ]
+
+        stats, fields = views.csv_fields(series)
+        response = views.render_csv(request=RequestFactory().get('/'),
+                                    addon=addon_factory(),
+                                    stats=stats,
+                                    fields=fields)
+
+        assert '\n,a\r\n1,2\r\n0,4\r\n' in force_text(response.content)
```
eviljeff, line 930: It'd be:
```
,a
1,2
0,4
```
?

willdurand: Yes. I guess I can use a `'\r\n'.join([',a', '1,2', '0,4'])` instead if the comment is about readability. Otherwise, I think using a multiline string won't work because of the line endings.

src/olympia/stats/views.py
willdurand: Indeed. The problem is that `sorted` uses the `<` sign to compare strings, which does not work with `NoneType`.

src/olympia/stats/tests/test_views.py
eviljeff: I think that's clearer, yeah.  (I wrote it out because that was the only way I could parse it - doing it in my head didn't work!)

PR #14798, opened 2020-06-30
clear more user data, after 24 hours; after 7 years for addon devs

### LINKED ISSUE #14494: setup cron job(s) to delete account/addon data after a set period

eviljeff:
fixes #14494


src/olympia/users/management/commands/clear_old_user_data.py
```diff
@@ -0,0 +1,46 @@
...
+        'and non addon-developers after 24 hours')
+
+    def handle(self, *args, **options):
+        profile_clear = {
+            'last_login_ip': '',
+            'email': None,
+            'fxa_id': None,
+        }
+        one_day_ago = datetime.now() - timedelta(days=1)
+        seven_years_ago = datetime.now() - timedelta(days=365 * 7)
```
willdurand, line 28: It's not strictly 7 years, is that ok?

eviljeff: 7 years ish 😉 
@jvillalobos does it need to be exactly 7 years or is 365*7 close enough? (it's 1 or 2 days short depending on leap years)

jvillalobos: Close enough!

Issue #14807, opened 2020-07-01
Make beta stats the default stats (and remove legacy stats)


willdurand:
Let's remove the legacy stats pages and replace them with the beta pages. We're going to get rid of the beta waffle flag and there won't be any way to access the legacy stats anymore.

The beta flag won't be used but it won't be deleted now to avoid potential issues during deployment.

ioanarusiczki:
@willdurand On dev - FF78(Win10):

- there's no longer a beta URL - it's 404 (even if the beta cookie is true). And for the [samples](https://addons-dev.allizom.org/en-US/firefox/addon/amo-info-with-extra-dirs/statistics/usage/countries/?last=365) available the stats are displayed (what used to be beta before).

- lang packs and dicts no longer have stats of any type. The stats URLs from dev hub have been removed. I don't know if that's expected to be temporary or permanent?

willdurand:
> lang packs and dicts no longer have stats of any type. The stats URLs from dev hub have been removed. I don't know if that's expected to be temporary or permanent?

Expected. There was no beta for langpacks/dicts.

PR #14810, opened 2020-07-01
Make beta stats the default stats

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14807


src/olympia/devhub/tests/test_views.py
```diff
@@ -155,10 +155,9 @@ def test_show_hide_statistics_and_new_version_for_disabled(self):
         assert 'Statistics' in links, ('Unexpected: %r' % links)
         assert 'New Version' in links, ('Unexpected: %r' % links)
 
-        # Disabled (user): hide statistics and new version links.
+        # Disabled (user): hide new version link.
         self.addon.update(disabled_by_user=True)
         links = self.get_action_links(self.addon.pk)
-        assert 'Statistics' not in links, ('Unexpected: %r' % links)
```
willdurand, line 161: stats pages are available almost all the time for the developer

src/olympia/stats/tests/test_cron.py
```diff
@@ -68,15 +59,6 @@ def test_no_addon_or_date(self, group_mock):
...
         calls = group_mock.call_args[0][0]
 
-        # There should be 3 updates, but 2 of them have a date close enough
-        # together that they'll be indexed in the same chunk, so we should have
-        # 2 calls.
-        update_counts_calls = [
-            call.args for call in calls
-            if call.task == 'olympia.stats.tasks.index_update_counts'
-        ]
-        assert len(update_counts_calls) == 2
```
willdurand, line 78: Things like this will be fully removed in a different patch. I removed it here because tests were failing since I changed the fixtures files to only have download counts.

src/olympia/stats/tests/test_views.py
```diff
@@ -7,41 +7,67 @@
...
 
-from waffle.testutils import override_flag
-
 from olympia import amo
 from olympia.access.models import Group, GroupUser
 from olympia.addons.models import Addon, AddonUser
-from olympia.amo.tests import (TestCase, version_factory, addon_factory,
-                               user_factory)
-from olympia.amo.urlresolvers import reverse, resolve
+from olympia.amo.tests import (
```
willdurand, line 15: I had to rewrite most of the file because test cases verifying things like csv/json exports, etc. were based on the legacy stats. I kept almost everything and configured a BigQuery mock in each test method.

src/olympia/stats/tests/test_views.py
```diff
@@ -163,36 +186,7 @@ def test_stats_for_user_disabled_addon(self):
         self._check_it(self.public_views_gen(format='json'), 200)
 
 
-class ESStatsTest(StatsTest, amo.tests.ESTestCase):
```
willdurand, line 166: Moved/renamed below

src/olympia/stats/tests/test_views.py
```diff
@@ -234,353 +228,428 @@ def test_addon_anonymous(self):
         self._check_it(self.public_views_gen(addon_id=5, format='json'), 403)
 
 
-class TestCSVs(ESStatsTest):
```
willdurand, line 237: Similar test cases exist in `TestExports`. I removed the class here and kept test cases that weren't duplicated.

src/olympia/stats/views.py
```diff
@@ -162,6 +155,8 @@ def zip_overview(downloads, updates):
         start_date = downloads[0]['date']
     if updates:
         d = updates[0]['date']
+        if isinstance(d, str):
+            d = datetime.strptime(d, '%Y-%m-%d').date()
```
eviljeff, line 159: is this a drive-by fix for something else?

src/olympia/stats/views.py
```diff
@@ -170,7 +165,10 @@ def iterator(series):
             item = next(series)
             next_date = start_date
             while True:
-                if item['date'] == next_date:
+                date = item['date']
+                if isinstance(date, str):
+                    date = datetime.strptime(date, '%Y-%m-%d').date()
```
eviljeff, line 170: same here.  We're now converting string to date... but only when it's a str (what other types can it be?)

src/olympia/stats/views.py
willdurand: d'oh. I meant to comment on this line too, sorry. Yes. That's because BigQuery returns datetime instances but the test suite uses dates as strings. I feel like it is safer to have that. WDYT?

eviljeff: I think we should be testing with the same format really, so the test data is datetime (at least by the time it's consumed by the view).  If this is going to be a big change feel free to put it in a follow-up.

willdurand: You're right. I updated the test file with `date` instances. I also updated the BigQuery function to format series so that it returns `date` instances too. It should be the case but better safe than sorry.

src/olympia/stats/utils.py
```diff
@@ -21,10 +23,15 @@ def rows_to_series(rows, filter_by=None):
     """Transforms BigQuery rows into series items suitable for the rest of the
     AMO stats logic."""
     for row in rows:
+        submission_date = (
+            datetime.strptime(row['submission_date'], '%Y-%m-%d').date()
+            if not isinstance(row['submission_date'], date)
+            else row['submission_date']
```
eviljeff, line 29: I thought you said BigQuery returned `date` instances already?  When are they not `date`s?  (Should there be try... except handling if the data format is incorrect?)

willdurand: right, although this is luck because I didn't specifically check the code of the third-party lib.. There is no date-related error so far so I can remove this for now, I guess.

PR #14830, opened 2020-07-02
Update logging for developer reply emails

### LINKED ISSUE #14461: Do not forward review developer reply expected behavior to sentry
The following exceptions should not be forwarded to sentry since they are expected:

https://github.com/mozilla/addons-server/blob/9a129eb7a443fe77afdf49fc439310815d8eeec9/src/olympia/activity/utils.py#L113-L115

https://github.com/mozilla/addons-server/blob/9a129eb7a443fe77afdf49fc439310815d8eeec9/src/olympia/activity/utils.py#L176-L177

wagnerand:
Fix #14461.


wagnerand:
I am not sure we specifically test logging levels. This is to stop the sentry emails, so I'll merge now and we can follow up later if needed.

PR #14833, opened 2020-07-03
RE: Delete PrimaryHeroImages files when instance is deleted 

### LINKED ISSUE #14729: Delete PrimaryHeroImages files when instance is deleted
When a `PrimaryHeroImage` instance is deleted we should delete the file(s) on disk that correspond to it too.  This would include any resized or thumbnail copies.

xlisachan:
Fixes #14729 

The following updates have been made with this pull request:
- Image files are deleted upon deleting PrimaryHeroImage instance
- `discovery/tests/test_admin.py` has been updated

I had tried to place what's in `models.py` in `admin.py` (changing `self` to `obj`, etc) and experimented with `delete_model` as well. However, I'd either (1) get an error (`'basequeryset' has not attribute custom_image`) and the files would not delete while all tests in `test_admin`would pass, or (2) the images would get deleted upon deleting the instance, but the tests would fail. 🤨 Please review when you have a moment. Thanks!

### DIFF
--- a/src/olympia/hero/models.py
+++ b/src/olympia/hero/models.py
@@ -113,6 +113,14 @@ class PrimaryHeroImage(ModelBase):
     preview_image.short_description = "Image"
     preview_image.allow_tags = True
 
+    def delete(self, *args, **kwargs):
+        (path, fn) = os.path.split(self.custom_image.path)
+        dest_thumb = path + b'/thumbs/' + fn
+
+        os.remove(dest_thumb)
+        os.remove(self.custom_image.path)
+        super().delete(*args, **kwargs)
+
 
 class PrimaryHero(ModelBase):
     select_image = models.ForeignKey(
--- a/src/olympia/hero/models.py
+++ b/src/olympia/hero/models.py
@@ -2,6 +2,7 @@ import os
 
 from django.conf import settings
 from django.core.exceptions import ValidationError
+from django.core.files.storage import default_storage as storage
 from django.db import models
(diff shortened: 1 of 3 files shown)


eviljeff:
> I had tried to place what's in models.py in admin.py (changing self to obj, etc) and experimented with delete_model as well.

I think it's better in the model's `delete()` anyway - we _always_ want the files to be tidied up, even if the delete happens through some other means (example: maybe in the future they'll be an API to delete them)

src/olympia/hero/models.py
```diff
@@ -113,6 +113,14 @@ def preview_image(self):
     preview_image.short_description = "Image"
     preview_image.allow_tags = True
 
+    def delete(self, *args, **kwargs):
+        (path, fn) = os.path.split(self.custom_image.path)
+        dest_thumb = path + b'/thumbs/' + fn
```
eviljeff, line 118: I think as we're doing this in two places now (both here and in `PrimaryHeroImage.save_model`) we could do with abstracting it out into an `@property` (`def thumbnail_path`?) and then using that here and in `save_model`.

xlisachan:
Thanks @eviljeff! Per your feedback, please see the changes made below:
- Imported `default_storage as storage` and replaced `os.remove(...)` to `storage.delete(...)`
- Added a check to see if the files existed in storage with `storage.exists(...)`
- Added `thumbnail_path` and updated `save.model` and `delete`

Issue #14454, opened 2020-06-01
Implement delayed review rejections


diox:
To give developers sufficient time to fix their add-ons before their versions are taken down and 
Minimize the time add-on listings are unavailable due to review issues, we want to implement delayed review rejection option as per [PRD](https://docs.google.com/document/d/1rh8rfMviJwEQekz-qoDT487q2mc5KsJiDWY7rDQtmBs/edit#).

Once [the other issues in the project](https://github.com/orgs/mozilla/projects/130) are done: 

- [ ] In the reviewer tools, rejections default to applying after a grace period of ~~7~~ 14 days (the same time interval used previously for info requests).
  - [ ] Like with info requests, the email sent to developers must explain this.
  - [ ] Rejections of multiple versions apply the grace period for all selected versions.
  - [ ] Auto-approval is disabled for the add-on until it gets a manual approval or an admin changes the auto-approval flag.
- [ ] Reviewers also have the option to reject immediately, for cases where it’s considered users are at real risk or there have been repeated rejections with no progress.

diox:
@jvillalobos Previously we had a cronjob that would notify developers about soon to expire info requests, now that pending rejection deadline replaces that, are we keeping that cron, but using it with the new delayed rejections ?

jvillalobos:
Yes, please.

diox:
Note: to support https://github.com/mozilla/addons-server/issues/14814, the comments from the reviewer needs to be attached to the activity log associated with a delayed review rejection. It probably needs to be a separate activity log, too, to be able to distinguish between an immediate and a delayed rejection.

AlexandraMoga:
Since this new feature covered quite a lot of functionality in Reviewer Tools, I've created a list of [test scenarios](https://docs.google.com/spreadsheets/d/13UUZ5AnmzkOs38Wb-jmJpQhzBRYgcYVWlxWy6AFgiyI/edit?usp=sharing) that I found relevant - based on the [PRD](https://docs.google.com/document/d/1rh8rfMviJwEQekz-qoDT487q2mc5KsJiDWY7rDQtmBs/edit#) and the various comments added in the other issues from [this project](https://github.com/orgs/mozilla/projects/130).

The Reject multiple versions action (for listed) includes now two options: 
-  reject versions with a delay (default is 14 days and the period can't be updated by regular reviewers) 
- reject immediately (the option selected by default).

[...]

PR #14838, opened 2020-07-03
Add cron to auto-reject versions pending rejection past grace period

### LINKED ISSUE #14799: Add management command applying the delayed reviewers rejections
Background: https://github.com/mozilla/addons-server/issues/14454 will be about implementing delayed rejections as an action for reviewers. https://github.com/mozilla/addons-server/issues/14577 is about the developers side. 

We're going to need a cron job that does the rejections, regularly looking at pending versions to reject. https://github.com/mozilla/addons-server/issues/14454 will automate this, for now we want the command without automating it, to ease testing.

From the PRD:
- [ ] If the grace period ends and there’s no version pending review, the rejection is applied and the versions are disabled.

diox:
Fixes #14799 

The command isn't added to crontab yet to ease testing. It will be in https://github.com/mozilla/addons-server/issues/14454

### DIFF
--- a/src/olympia/reviewers/management/commands/auto_reject.py
+++ b/src/olympia/reviewers/management/commands/auto_reject.py
@@ -93,7 +93,7 @@ class Command(BaseCommand):
             # Don't auto-reject something that has been locked, even by the
             # task user - wait until it's free to avoid any conflicts.
             log.info(
-                'Skipping rejections for add-on %s until lock from %s '
+                'Skipping rejections for add-on %s until lock from user %s '
                 'has expired', addon.pk, locked_by)
             return
             set_reviewing_cache(addon.pk, settings.TASK_USER_ID)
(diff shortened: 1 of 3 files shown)


src/olympia/reviewers/management/commands/auto_reject.py
```diff
@@ -0,0 +1,121 @@
...
+            log.info(
+                'Somehow no versions to auto-reject for add-on %s', addon.pk)
+            return
+        locked_by = get_reviewing_cache(addon.pk)
+        if locked_by:
+            # Don't auto-reject something that has been locked, even by the
+            # task user - wait until it's free to avoid any conflicts.
+            log.info(
+                'Skipping rejections for add-on %s until lock from %s '
+                'has expired', addon.pk, locked_by)
```
willdurand, line 97: `locked_by` returns a numeric value for me (like `1`), maybe we could rephrase this log statement like: `...until lock from reviewer %s has expired` or something.

diox: Yeah it's an user id. I adjusted the message in https://github.com/mozilla/addons-server/pull/14838/commits/e2c8b3a7c1494d9b963c69e0256e3797ae5309f2

Issue #14832, opened 2020-07-02
Expose stats by app version for Android (Fenix)


willdurand:
Thanks to https://github.com/mozilla/bigquery-etl/pull/1056 (bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1643683), we also have usage stats for Fenix (in addition to Desktop stats). There is a new column for Fenix-specific data: `dau_by_fenix_build`, which is similar to `dau_by_app_version`.

This column contains usage stats broken down by Fenix build, and it is very similar to Dekstop versions. Here is an example:

```json
[
  {
    "build": "77.0.1",
    "count": "3692627"
  },
  {
    "build": "76.0.1",
    "count": "94774"
  },
  {
    "build": "78.0",
    "count": "71458"
  },
  {
    "build": "68.9.0",
    "count": "55914"
  },
  {
    "build": "75.0",
    "count": "35954"
  },
  {
    "build": "52.9.0",
    "count": "22079"
  },
  {
    "build": "72.0.2",
    "count": "20901"
  },
  {
    "build": "77.0",
    "count": "20149"
  },
  {
    "build": "74.0",
    "count": "16903"
  },
  {
    "build": "76.0",
    "count": "13837"
  },
  {
    "build": "73.0.1",
    "count": "10328"
  },
  {
    "build": "79.0a1",
    "count": "9110"
  },
  {
    "build": "71.0",
    "count": "8788"
  },
  {
    "build": "72.0.1",
    "count": "7325"
  },
  {
    "build": "74.0.1",
    "count": "6580"
  },
  {
    "build": "70.0.1",
    "count": "6389"
  },
  {
    "build": "68.8.0",
    "count": "5650"
  },
  {
    "build": "66.0.3",
    "count": "5476"
  },
  {
    "build": "79.0",
    "count": "4852"
  },
  {
    "build": "80.0a1",
    "count": "3839"
  },
  {
    "build": "68.0.1",
    "count": "2881"
  },
  {
    "build": "48.0.2",
    "count": "2869"
  },
  {
    "build": "69.0.1",
    "count": "2680"
  },
  {
    "build": "69.0",
    "count": "2600"
  },
  {
    "build": "60.9.0",
    "count": "2570"
  },
  {
    "build": "68.0.2",
    "count": "2359"
  },
  {
    "build": "70.0",
    "count": "2157"
  },
  {
    "build": "68.7.0",
    "count": "2108"
  },
  {
    "build": "73.0",
[...]

PR #14850, opened 2020-07-06
Expose stats by app version for Android (Fenix)

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14832

---

We need data in two columns for the stats by Application. We don't have a
"Fenix" app constant and Fenix doesn't have a different app GUID so I reused
`ANDROID`.

Note: one can try this patch by updating the `AMO_TO_BIGQUERY_COLUMN_MAPPING`
value like this, which will copy the Desktop data for Fenix (because we don't
have add-ons compatible with Fenix in -dev/-stage):

```
    'apps': 'dau_by_app_version, dau_by_app_version as dau_by_fenix_build',
```

(Google credentials are needed)

### DIFF
--- a/src/olympia/stats/utils.py
+++ b/src/olympia/stats/utils.py
@@ -2,11 +2,11 @@ from django.conf import settings
 from django_statsd.clients import statsd
 from google.cloud import bigquery
 
-from olympia.constants.applications import FIREFOX
+from olympia.constants.applications import ANDROID, FIREFOX
 
 # This is the mapping between the AMO stats `sources` and the BigQuery columns.
 AMO_TO_BIGQUERY_COLUMN_MAPPING = {
-    'apps': 'dau_by_app_version',
+    'apps': 'dau_by_app_version, dau_by_fenix_build',
     'countries': 'dau_by_country',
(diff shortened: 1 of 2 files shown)


src/olympia/stats/utils.py
```diff
@@ -27,13 +27,25 @@ def rows_to_series(rows, filter_by=None):
...
             if filter_by == AMO_TO_BIGQUERY_COLUMN_MAPPING['apps']:
-                item['data'] = {FIREFOX.guid: item['data']}
+                item['data'] = {
+                    ANDROID.guid: {
+                        d['key']: d['value']
+                        for d in row.get('dau_by_fenix_build', [])
+                    },
+                    FIREFOX.guid: {
+                        d['key']: d['value']
+                        for d in row.get('dau_by_app_version', [])
```
eviljeff, line 42: I'd prefer there was an implementation that doesn't end up with `dau_by_app_version` and `dau_by_fenix_build` being hardcoded both here and in the constant above 😢 But I couldn't think of anything that wasn't horribly overcomplicated.

Issue #14849, opened 2020-07-06
Lots of UnicodeEncodeErrors going to sentry


wagnerand:
We are getting lots of UnicodeEncodeErrors going to sentry, apparently with a custom format, so each is added a separate entry: https://sentry.prod.mozaws.net/operations/olympia-prod/?query=is%3Aunresolved+UnicodeEncodeError

Can we try and identify the source of that error and update the error handling so sentry can group these correctly?

diox:
This is essentially requests with malformed data. They were previously not recorded because we didn't capture `log.exception()`. 

IMHO we should try to avoid them entirely.

diox:
This would be typical for some automated vulnerability scanning to try to inject specific characters like a null character in the query string and we have some of that, but weirdly, we also have a lot of different requests with the same `鎈` character at different positions. They look like this:
- `'currentAppVersion[1]': '3鎈\'"\\('`
-  `'reqVersion': '2鎈\'"\\(',`
- `'updateType[0]': '=2鎈\'"\\(and 3 in '`

diox:
QA: test that the update service still works ; Then play with the URL (like https://versioncheck-dev.allizom.org/update/VersionCheck.php?reqVersion=2&id=%7Bd10d0bf8-f5b5-c8b4-a8b2-2b9879e08c5d%7D&version=3.5.2&maxAppVersion=77.%2A&status=userEnabled%2Cincompatible&appID=%7Bec8030f7-c20a-464f-9b0e-13a3a9e97384%7D&appVersion=78.0.112&appOS=Linux&appABI=x86_64-gcc3&locale=en-US&currentAppVersion=78.0.112&updateType=49&compatMode=normal) adding special characters in various places, you should still get a valid json response.

ioanarusiczki:
@diox I'll try to update to the latest version some addons and themes - check they do well with manual and automatic update (after restarting the browser). (so far it's going well)

About the json above, depends very much of the "various places" because I already broke it. Could you be a bit more specific?

ioanarusiczki:
@diox Results:
- the [specified](https://github.com/mozilla/addons-server/issues/14849#issuecomment-654788608) URL works as expected, returning `{}` as a response.
- I verified the update service on AMO dev and AMO stage - works fine.

PR #14851, opened 2020-07-06
Fix database encoding and exception handling in update service

diox:
Fixes https://github.com/mozilla/addons-server/issues/14849

### DIFF
--- a/settings_test.py
+++ b/settings_test.py
@@ -48,6 +48,11 @@ CACHES = {
     },
 }
 
+# Cheat and change SERVICES_DATABASE['NAME'] to the test database name. Our
+# update service doesn't use django database connections so we need to override
+# that setting so that it points to the correct database.
+SERVICES_DATABASE['NAME'] = 'test_%s' % SERVICES_DATABASE['NAME']
+
 # Overrides whatever storage you might have put in local settings.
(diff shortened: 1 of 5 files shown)


settings_test.py
```diff
@@ -48,6 +48,11 @@
     },
 }
 
+# Cheat and change SERVICES_DATABASE['NAME'] to the test database name. Our
+# update service doesn't use django database connections so we need to override
+# that setting so that it points to the correct database.
+SERVICES_DATABASE['NAME'] = 'test_%s' % SERVICES_DATABASE['NAME']
```
diox, line 54: I knew this was naive but yeah, this isn't going to work at all. The problems are:
- `services/utils.py` builds `mypool` directly at import time using `SERVICES_DATABASES`
- `SERVICES_DATABASES` doesn't take into account the test db

Django doesn't really provide an API to retrieve the test db name either... It's going to be in `settings.DATABASES[x]['NAME']`... once tests have been initialized. Need to be smarter about this.

diox: Found a proper fix through pytest fixtures, implemented in https://github.com/mozilla/addons-server/pull/14851/commits/f3ed1cfa0a68e934efbfa2c8c9234cd2d3f2f711

Issue #14815, opened 2020-07-01
Use BigQuery to compute add-on "hotness"


willdurand:
Let's use BigQuery to update add-on hotness values.

ioanarusiczki:
@willdurand If I could verify this, please add some STR. The `hotness` parameter is available in search results such as https://addons-dev.allizom.org/en-US/firefox/search/?platform=windows&q=amo-info-with-extra-dirs&sort=hotness

PR #14852, opened 2020-07-06
Use BigQuery to compute add-on "hotness"

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14815

---

This replaces the previous CRON task with a newer version, very similar but it
uses BigQuery. The averages computation is performed in BigQuery. I tried in
Redash (with the prod data) is take around 20 seconds.

### DIFF
--- a/src/olympia/addons/cron.py
+++ b/src/olympia/addons/cron.py
@@ -1,3 +1,7 @@
+import time
+
+from datetime import date
+
 from django.db.models import F, Q, Sum
 from django.db.models.functions import Coalesce
 
@@ -8,7 +12,7 @@ from celery import group
 import olympia.core.logger
 
 from olympia import amo
-from olympia.addons.models import Addon
+from olympia.addons.models import Addon, FrozenAddon
 from olympia.addons.tasks import (
     update_addon_average_daily_users as _update_addon_average_daily_users,
     update_addon_download_totals as _update_addon_download_totals,
@@ -17,7 +21,8 @@ from olympia.amo.decorators import use_primary_db
 from olympia.amo.utils import chunked
 from olympia.files.models import File
 from olympia.stats.utils import (
-    get_addons_and_average_daily_users_from_bigquery)
+    get_addons_and_average_daily_users_from_bigquery,
+    get_averages_by_addon_from_bigquery)
 
 
 log = olympia.core.logger.getLogger('z.cron')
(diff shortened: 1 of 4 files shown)


src/olympia/addons/cron.py
```diff
@@ -199,33 +199,39 @@ def deliver_hotness():
...
+            log.debug('Add-on with GUID "%s" is not public, skipping', guid)
+            continue
+
+        # Update the hotness score but only update hotness if necessary. We
+        # don't want to cause unnecessary re-indexes.
+        threshold = 250 if addon.type == amo.ADDON_STATICTHEME else 1000
+        if this > threshold and three > 1:
+            hotness = (this - three) / float(three)
+            if addon.hotness != hotness:
+                addon.update(hotness=(this - three) / float(three))
```
bobsilverberg, line 231: Why can't this be just `addon.update(hotness=hotness)`?

src/olympia/addons/cron.py
```diff
@@ -199,33 +199,39 @@ def deliver_hotness():
...
+    averages = get_averages_by_addon_from_bigquery(today=date.today())
+
+    for average in averages:
+        # BigQuery returns GUIDs
+        guid = average['addon_id']
+        this = average['avg_this_week']
+        three = average['avg_three_weeks_before']
+
+        try:
+            addon = Addon.objects.get(guid=guid)
```
eviljeff, line 212: This is really inefficient compared to the previous implementation - we're doing a database lookup for _each_ addon.  Get a list of guids from the `averages` list and make a single query.

src/olympia/stats/utils.py
```diff
@@ -111,3 +115,76 @@ def get_addons_and_average_daily_users_from_bigquery():
...
+    """This function is used to compute the 'hotness' score of each add-on (see
+    also `deliver_hotness()` cron task)."""
+    client = bigquery.Client.from_service_account_json(
+        settings.GOOGLE_APPLICATION_CREDENTIALS
+    )
+
+    one_week_date = today - timedelta(days=7)
+    four_weeks_date = today - timedelta(days=28)
+
+    fully_qualified_table_name = '.'.join(
```
eviljeff, line 130: This is defined 3 times in this file; can we abstract it to a constant or function ?

src/olympia/addons/cron.py
```diff
@@ -199,33 +199,39 @@ def deliver_hotness():
...
-                    addon.update(hotness=(this - three) / float(three))
-            else:
-                if addon.hotness != 0:
-                    addon.update(hotness=0)
+    frozen = set(f.addon_id for f in FrozenAddon.objects.all())
+    averages = get_averages_by_addon_from_bigquery(today=date.today())
+
+    for average in averages:
+        # BigQuery returns GUIDs
+        guid = average['addon_id']
```
eviljeff, line 207: there's a risk of hitting the `guid is None` problem again here.

willdurand: The `get_averages_by_addon_from_bigquery` function omits `None` GUIDs.

eviljeff: oh, I missed it because it was still called `addon_id` at that point

willdurand: I wasn't sure about renaming this property, but looks like I should.

src/olympia/addons/cron.py
willdurand: heh, good call yeah.

Issue #14876, opened 2020-07-07
Expose creation date in scanner results


willdurand:
This came up in our ML meeting today. It would be useful to have the `created` date exposed in the API response.

AlexandraMoga:
The `created` field is now returned by the scanner results API:

`"created": "2020-07-02T12:56:19Z"`

Verified fixed on stage.

PR #14877, opened 2020-07-07
Add created field in ScannerResultSerializer

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14876

### DIFF
--- a/src/olympia/scanners/tests/test_serializers.py
+++ b/src/olympia/scanners/tests/test_serializers.py
@@ -1,3 +1,5 @@
+from django.conf import settings
+
 from olympia.amo.tests import TestCase
 from olympia.constants.scanners import CUSTOMS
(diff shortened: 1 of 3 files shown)


src/olympia/scanners/tests/test_serializers.py
```diff
@@ -13,4 +15,7 @@ def test_serialize(self):
             'scanner': result.get_scanner_name(),
             'label': None,
             'results': result.results,
+            'created': result.created.strftime(
```
bobsilverberg, line 18: I gather that it's magic built into DRF that applies this `strftime` to the `created`?

willdurand: Yeah, this is https://www.django-rest-framework.org/api-guide/fields/#datetimefield

PR #14896, opened 2020-07-09
Ignore user-disabled add-ons for user average rating

### LINKED ISSUE #14895: Ratings for "invisible" addons persist on developer profile pages
I have 2 extensions. One of them received a rating. Later, I marked that extension as "invisible". However, its rating is still displayed in my profile, under "Average rating of developer’s add-ons". I don't think this is the intended behavior.

See https://addons.mozilla.org/en-US/firefox/user/15596255/

wagnerand:
Fix #14895.


eviljeff:
> Someday we should look into moving that query to use the ORM...

ugh, yeah.

PR #14898, opened 2020-07-09
add promoted app and PromotedAddon, PromotedApproval classes

### LINKED ISSUE #14858: Add Promoted model(s)
When we expand the types of promoted add-ons from just recommended we need some more models to hold this information

eviljeff:
fixes #14858


src/olympia/addons/tests/test_models.py
```diff
@@ -1586,6 +1588,43 @@ def test_theme_is_recommended(self):
...
+    def test_promoted_group(self):
+        addon = addon_factory()
+        # default case - no group so not recommended
+        assert not addon.promoted_group
+
+        PromotedApproval.objects.create(
+            version=addon.current_version, group_id=SPOTLIGHT.id)
+        promoted = PromotedAddon.objects.create(
+            addon=addon, group_id=SPOTLIGHT.id)
+        del addon.promoted_group
```
bobsilverberg, line 1600: You need to do this because it's a `@cached_property`?

src/olympia/constants/promoted.py
```diff
@@ -0,0 +1,78 @@
...
+
+class NOT_PROMOTED(PromotedClass):
+    pass
+
+
+class RECOMMENDED(PromotedClass):
+    id = 1
+    name = _('Recommended')
+    search_ranking_bump = 1000  # TODO: confirm this bump
+    badge = 'badge'  # TODO: define what the badge is
```
bobsilverberg, line 22: Do you see this being used in an API response, to tell the frontend which badge to display?

src/olympia/promoted/models.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def is_addon_promoted(self, addon=None):
+        """Is the addon approved for promotion within the *current* promoted
+        group."""
+        addon = addon or self.addon  # the caller might already have the addon
+        if not addon or not addon.current_version:
+            return False
+        return bool(
+            self.group != NOT_PROMOTED.id and
+            addon.current_version.promoted_approvals.filter(
```
bobsilverberg, line 39: So just the existence of a `PromotedApproval` means that the `addon` was approved for this group - there isn't a boolean `approved` property?

src/olympia/promoted/models.py
```diff
@@ -0,0 +1,57 @@
...
+from olympia.constants.promoted import (
+    NOT_PROMOTED, PROMOTED_GROUPS, PROMOTED_GROUPS_BY_ID)
+from olympia.versions.models import Version
+
+
+class PromotedAddon(ModelBase):
+    GROUP_CHOICES = [(group.id, group.name) for group in PROMOTED_GROUPS]
+    group_id = models.SmallIntegerField(
+        choices=GROUP_CHOICES, default=NOT_PROMOTED.id)
+    addon = models.OneToOneField(
```
bobsilverberg, line 14: This is what enforces the rule that any `addon` can have only one `PromotedAddon`?

src/olympia/addons/tests/test_models.py
eviljeff: yes.

src/olympia/constants/promoted.py
eviljeff: I don't know at this point - the PRD calls for some groups to have a special badge but I'm not sure if that means the API would be returning a constant indicating a special badge, or if frontend would have the logic to identify a certain group needing a special badge.

src/olympia/promoted/models.py
eviljeff: Correct - the record would only be created on approval.  Strategic group promoted addons are a ❔ right now because they're auto-approved - I guess we'd just created it on auto approval.

src/olympia/promoted/models.py
eviljeff: yep.  OneToOneField are inherently unique in django.

src/olympia/constants/promoted.py
bobsilverberg: I suppose no matter what approach we take, it's going to require coordination/synchonization between the server and the frontend. I'm working on the badges right now, and there's a bunch of JS involved, so I don't think the server can return an actual badge to us. Therefore, we'll need to create a mapping between a code returned from the server and a badge type on the frontend. We already do this with all sorts of things, so it's no big deal. 

I guess the question is whether the server needs to understand the concept of a "badge" at all, or if the server can just return a code for the promotion category, and the frontend will understand whether a badge is needed, and if so which badge based on that code. The latter seems simpler from the server perspective, and we'll need some logic on the frontend anyway, so I'd probably take that approach.

eviljeff: dropped `badge` in https://github.com/mozilla/addons-server/pull/14898/commits/89826e69b29c8274a2e8c0ee8aa049774c27b41c

src/olympia/promoted/models.py
```diff
@@ -0,0 +1,57 @@
...
+        Addon, on_delete=models.CASCADE, null=False)
+
+    @property
+    def group(self):
+        return PROMOTED_GROUPS_BY_ID.get(self.group_id, NOT_PROMOTED)
+
+    def __getattr__(self, name):
+        # magic to return the fixed properties of the promoted group too
```
diox, line 22: Is this really necessary ? You could have a `@property` called... `properties` or something that would return them without adding them magically to this class.

src/olympia/promoted/models.py
```diff
@@ -0,0 +1,57 @@
...
+        try:
+            super().__getattribute__(name)
+        except AttributeError as parent_attr_error:
+            try:
+                return getattr(self.group, name)
+            except AttributeError:
+                raise parent_attr_error
+
+    def is_addon_promoted(self, addon=None):
```
diox, line 31: I like the overall strategy of how the models are set up but I am not a big fan of this API: Fetching a `PromotedAddon` only to have to call `is_addon_promoted()` to check feels weird. It also doesn't really tell you that the `current_version` is what really matters...

Also, I guess being able to pass `addon` helps avoid extra queries but that makes the API even weirder...

src/olympia/promoted/models.py
eviljeff: Its not.  I just added it on a whim because I thought it might be cool.  There already is a `@property` that returns them - `group` above.

diox: Yeah I'd remove that, playing with `__getattr__` just for this is asking for trouble later down the road, IMHO. And the `group` `@property` seems enough.

src/olympia/promoted/models.py
eviljeff: Implementing it like `Addon.is_recommended` is the alternative, but it places the logic in the Addon model instead and I was trying to avoid that again.  Maybe I should make it a classmethod that handles fetching the `PromotedAddon` too?

diox: So actually... when django instantiates the related instance from the `onetoonefield` it should automatically set the parent instance, avoiding extra queries (said differently: `addon.promotedaddon.addon` should already automatically be the same instance than `addon`, avoiding extra queries). So, I'd keep this as a regular method, but remove the ability to pass an `addon`. And maybe rename to `is_addon_current_version_promoted` or `is_addon_currently_promoted` or something...

eviljeff: >  Maybe I should make it a classmethod that handles fetching the PromotedAddon too?

/me tries that

eviljeff: https://github.com/mozilla/addons-server/pull/14898/commits/714cae363455aba4f8edf98457e145e8ed18654a

src/olympia/promoted/models.py
eviljeff: https://github.com/mozilla/addons-server/pull/14898/commits/714cae363455aba4f8edf98457e145e8ed18654a

src/olympia/constants/promoted.py
```diff
@@ -0,0 +1,73 @@
...
+    name = _('Recommended')
+    search_ranking_bump = 1000  # TODO: confirm this bump
+    warning = False
+    pre_review = True
+
+
+class VERIFIED_ONE(PromotedClass):
+    id = 2
+    name = _('Verified - Tier 1')
+    search_ranking_bump = 500  # TODO: confirm this bump
```
diox, line 28: Note: the current naive way we boost works ok for the limited number of extensions we need to boost up - worst case scenario we end up boosting a couple recommended results that were at the end of the results, almost irrelevant, but if we increase the number of things we boost, we might end up having to redesign the way it works to avoid littering the results with mostly irrelevant but verified/recommended extensions.

PR #14901, opened 2020-07-09
Use the git shortref as tag for Docker images.

smarnach:
When pushing the `latest` tag to Docker Hub, this change will tag the same image with an abbreviated version of the git SHA1 hash that triggered the build. We can use this tag in Jenkins to trigger the actual build, so the Slack messages contain the git commit hash that was deployed.

This is one step towards implementing https://bugzilla.mozilla.org/show_bug.cgi?id=1411598.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ commands:
           name: Build docker image and push to repo
           command: |
             docker version
-            docker build -t app:build -f << parameters.dockerfile >> .
+            docker build -t app:build -f << parameters.dockerfile >> --label git.commit="$CIRCLE_SHA1" .
             docker tag app:build "${DOCKERHUB_REPO}":<< parameters.image_tag >>
             docker login -u "${DOCKERHUB_USER}" -p "${DOCKERHUB_PASS}"
             docker push "${DOCKERHUB_REPO}":<< parameters.image_tag >>


smarnach:
The new code will only run _after_ this has been merged, so this hasn't been tested yet. I could mess with the filters, so we can test this before merging if you want me to. Alternatively, we just merge after review, and roll back if it breaks.

bqbn:
I'm not sure how useful it is to have the short-ref in the bot message. By fulfilling the request in bug 1411598, we add overheads to the build and push systems. The dev environment can be pushed dozens of times in a busy day, historically, knowing the short-refs of each of them has never been necessary to me.

That's just speaking for myself though, and others may have a stronger opinion to have the short-ref in the bot message. My focus is apparently mainly on the complexity and reliability of the build-push pipeline.

diox:
It's super helpful when a deploy goes wrong to figure out quickly what revision caused the problem. It's also useful when a deploy goes right to quickly be able to tell QA they can start verifying something, without having to load the website and check `__version__`.

smarnach:
I don't really need the shortref as well, but I can see how it is useful for the developers. And the overhead we add here is pretty negligible – it's basically just a call to the Docker Hub API to add another tag to the image that was already uploaded, so I'd be surprised if this adds more than a half a second to the build time.

Maybe we can try whether this causes problems? I don't expect it to, and if it does, we can still roll it back.

smarnach:
For some reason, some Python tests failed in the build. I don't think the test failures are related. I'll try to rebase on top of the main branch.

diox:
> For some reason, some Python tests failed in the build. I don't think the test failures are related. I'll try to rebase on top of the main branch.

Those are known intermittent failures

smarnach:
@diox Fair enough, then I won't bother rebasing.

willdurand:
I have the feeling that we shouldn't create more docker tags and instead use docker labels. We can set a label during the build:

```
 docker build [...] --label "git_hash=$CIRCLE_SHA1"
```

and then we can read it during deployment:

```
docker inspect -f '{{ .Config.Labels.git_hash }}' <image_name>
```

smarnach:
@willdurand I agree that using labels is nicer in theory. However, as mentioned on Bugzilla, tags are [the only information we control that's included in the webhook payload](https://docs.docker.com/docker-hub/webhooks/#example-webhook-payload). If we use labels, we need to implement a way to retrieve them in the Jenkins pipeline. Which is certainly possible, but more work than the approach chosen so far.

On the plus side, when using labels we won't need any changes to the "meta-pipeline" that decides what actual Jenkins pipeline to run when receiving the webhook data from Docker Hub, since we would still run the addons-server-dev pipeline whenever `latsest` is updated. In addition, we wouldn't accumulate useless persistent tags in Docker Hub at the same rate – currently our only labels are `latest` and the release date tags.

So all things considered, I guess using labels is the better approach. I've updated this PR accordingly.

There's an [open standard for container labels](https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys), and according to that standard we are supposed to use the label `org.opencontainers.image.revision`. I decided not to follow that standard.

smarnach:
@bqbn Are your concerns addressed here as well with the simpler approach? It comes at the price of a bit more complexity in [the pull-verify-push script](https://github.com/mozilla-services/cloudops-deployment/blob/master/bin/pull-verify-push). I also need to figure out how to get the hash from that script back into a Groovy variable.

bqbn:
Yeah, I too feel this is a better solution, especially there is no need to change the "meta-pipeline", and we can keep the standard practice of pulling `latest`.

Thanks!

smarnach:
@bqbn Thanks!

@diox @willdurand Can we merge this PR then?

smarnach:
I tried building locally with the argument, but it will only run on CircleCI once we merge it. I don't have write permissions to the repo, so you need to merge it yourself. :)

PR #14911, opened 2020-07-10
Add new reviewer flag to disable auto-approval until the next manual approval

### LINKED ISSUE #14910: Add addon reviewer flag to disable auto-approval until the next manual approval
For #14454 we need a way to disable auto-approval for an add-on until the next manual reviewer approval. I was hoping to re-use the existing flag but we want something separate to avoid re-enabling auto-approval when it's been disabled explicitly by an admin.

That new flag will, as the title suggests, disable auto-approval and automatically be reset when the next version of the add-on is manually approved. Like the other flag that disables auto-approval entirely, it will only affect/be reset by listed versions approvals.

Later, it will be set automatically in #14454 for delayed rejections.

diox:
This will be used by delayed review rejections to "pause" listed auto-approvals for an add-on until the next version that fixes the issues noted by the reviewer is approved.

Fixes #14910


src/olympia/reviewers/utils.py
```diff
@@ -828,21 +829,26 @@ def process_public(self):
...
             VersionReviewerFlags.objects.filter(
                 version=self.version
             ).update(needs_human_review_by_mad=False)
 
-        # Increment approvals counter if we have a request (it means it's a
-        # human doing the review) otherwise reset it as it's an automatic
-        # approval.
-        if self.human_review:
+            # An approval took place so we can reset this.
+            AddonReviewerFlags.objects.update_or_create(
```
willdurand, line 840: Why do we use `.update_or_create()` here? Instead of `.update()`?

src/olympia/reviewers/utils.py
```diff
@@ -879,11 +881,15 @@ def process_sandbox(self):
         self.set_files(amo.STATUS_DISABLED, self.files,
                        hide_disabled_file=True)
 
-        # Unset needs_human_review on the latest version - it's the only
-        # version we can be certain that the reviewer looked at.
         if self.human_review:
+            # Unset needs_human_review on the latest version - it's the only
+            # version we can be certain that the reviewer looked at.
             self.version.update(needs_human_review=False)
```
willdurand, line 887: This isn't related to this patch, but shouldn't we also reset the `needs_human_review_by_mad` (reviewerflags)?

src/olympia/reviewers/utils.py
```diff
@@ -1103,7 +1105,7 @@ def log_sandbox_message(self):
 
 class ReviewUnlisted(ReviewBase):
 
-    def process_public(self):
+    def approve_latest_version(self):
```
willdurand, line 1108: Not related to this patch, I filed this issue earlier https://github.com/mozilla/addons-server/issues/14943 and this renaming makes me think that this is where we want to fix the problem.

src/olympia/addons/tests/test_models.py
```diff
@@ -1365,6 +1365,18 @@ def test_auto_approval_disabled_property(self):
...
+        addon = Addon.objects.get(pk=3615)
+        # No flags: None
+        assert addon.auto_approval_disabled_until_next_approval is None
+        # Flag present, value is None (default): None.
+        flags = AddonReviewerFlags.objects.create(addon=addon)
+        assert flags.auto_approval_disabled_until_next_approval is None
+        assert addon.auto_approval_disabled_until_next_approval is None
+        # Flag present, value is True: True.
+        flags.update(auto_approval_disabled_until_next_approval=True)
+        assert addon.auto_approval_disabled_until_next_approval is True
```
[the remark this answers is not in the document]
diox: Yeah, we have to support `None` for the initial migration but technically it could be `False` as well, I'll patch the test.

src/olympia/reviewers/utils.py
diox: In this case both could work since we consider `False`, `None` and the absence of an `AddonReviewerFlags` instance as the same thing. In general though, I think `update_or_create` is slightly safer, since it will create the flags and set the explicit value you asked for no matter what. We haven't been very consistent with this, maybe we should...

src/olympia/reviewers/utils.py
diox: Probably, yes. Could be fixed when doing https://github.com/mozilla/addons-server/issues/14943 (even though we want it for listed as well)

src/olympia/reviewers/utils.py
diox: I think for #14943 we'll likely need to have a method in the `ReviewBase` class that does it and make sure it's called for a bunch of actions.

src/olympia/addons/tests/test_models.py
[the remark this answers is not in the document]
diox: 2c701375f40a4ebff804a73cd26afb87471a03cd

Issue #14900, opened 2020-07-09
Make the `deliver_hotness` task more efficient


willdurand:
We should really use something like `chunked()` to not load/instantiate 300K add-ons directly. We could also avoid loading the translations. In addition to that, we could certainly change the task to be asynchronous.

bqbn:
BTW, the manual run of the command failed with the following error.

```
$ ./manage.py cron deliver_hotness
{"Timestamp": 1594316605617230080, "Type": "z.cron", "Logger": "http_app_addons", "Hostname": "ip-172-31-25-45", "EnvVersion": "2.0", "Severity": 6, "Pid": 6, "Fields": {"uid": "<anon>", "remoteAddressChain": "", "msg": "Beginning job: deliver_hotness [] {} (start timestamp: 1594316605617.134)"}, "severity": 200}
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/mysql/base.py", line 71, in execute
    return self.cursor.execute(query, args)
  File "/usr/local/lib/python3.6/site-packages/MySQLdb/cursors.py", line 206, in execute
    res = self._query(query)
  File "/usr/local/lib/python3.6/site-packages/MySQLdb/cursors.py", line 319, in _query
    db.query(q)
  File "/usr/local/lib/python3.6/site-packages/MySQLdb/connections.py", line 259, in query
    _mysql.connection.query(self, query)
MySQLdb._exceptions.OperationalError: (2013, 'Lost connection to MySQL server during query')

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "./manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
[...]

PR #14913, opened 2020-07-10
Rewrite deliver_hotness() cron task

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14900

---

This patch updates the `deliver_hotness()` cron task to be asynchronous by
creating a Celery task for up to 300 add-ons.

I tried locally with 500K GUIDs in the `averages` dict and it does not eat a
lot memory and the command returns after a 20 seconds. We have to create small
dicts for each task because we cannot use `chunked` on a `dict`.

### DIFF
--- a/src/olympia/addons/cron.py
+++ b/src/olympia/addons/cron.py
@@ -1,5 +1,3 @@
-import time
-
 from datetime import date
 
 from django.db.models import F, Q, Sum
@@ -16,7 +14,7 @@ from olympia.addons.models import Addon, FrozenAddon
 from olympia.addons.tasks import (
     update_addon_average_daily_users as _update_addon_average_daily_users,
     update_addon_download_totals as _update_addon_download_totals,
-    update_appsupport)
+    update_appsupport, update_addon_hotness)
 from olympia.amo.decorators import use_primary_db
 from olympia.amo.utils import chunked
 from olympia.files.models import File
@@ -171,7 +169,7 @@ def unhide_disabled_files():
             file_.unhide_disabled_file()
 
 
-def deliver_hotness():
+def deliver_hotness(chunk_size=300):
     """
     Calculate hotness of all add-ons.
 
@@ -180,29 +178,15 @@ def deliver_hotness():
     threshold = 250 if addon type is theme, else 1000
     hotness = (a-b) / b if a > threshold and b > 1 else 0
     """
-    frozen = set(f.addon_id for f in FrozenAddon.objects.all())
+    frozen_ids = list(set(f.addon_id for f in FrozenAddon.objects.all()))
     averages = get_averages_by_addon_from_bigquery(today=date.today())
-    addons = (
-        Addon.objects.filter(guid__in=averages.keys())
-        .filter(status__in=amo.REVIEWED_STATUSES)
-        .exclude(id__in=frozen)
-    )
 
-    for addon in addons:
-        average = averages.get(addon.guid)
-        this = average['avg_this_week']
-        three = average['avg_three_weeks_before']
-
-        # Update the hotness score but only update hotness if necessary. We
-        # don't want to cause unnecessary re-indexes.
-        threshold = 250 if addon.type == amo.ADDON_STATICTHEME else 1000
-        if this > threshold and three > 1:
-            hotness = (this - three) / float(three)
-            if addon.hotness != hotness:
-                addon.update(hotness=hotness)
-        else:
-            if addon.hotness != 0:
-                addon.update(hotness=0)
-
-        # Let the database catch its breath.
-        time.sleep(10)
+    group(
+        [
+            update_addon_hotness.si(
+                averages={key: averages.get(key) for key in keys},
+                frozen_ids=frozen_ids,
+            )
+            for keys in chunked(averages.keys(), chunk_size)
+        ]
+    ).apply_async()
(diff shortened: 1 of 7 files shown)


src/olympia/addons/tasks.py
```diff
@@ -281,3 +281,32 @@ def delete_addons(addon_ids, with_deleted=False, **kw):
...
+        # Update the hotness score but only update hotness if necessary. We
+        # don't want to cause unnecessary re-indexes.
+        threshold = 250 if addon.type == amo.ADDON_STATICTHEME else 1000
+        if this > threshold and three > 1:
+            hotness = (this - three) / float(three)
+            if addon.hotness != hotness:
+                addon.update(hotness=hotness)
+        else:
+            if addon.hotness != 0:
+                addon.update(hotness=0)
```
eviljeff, line 312: Not a problem with this commit specifically, but can an addon suddenly drop out of the bigquery response?  e.g. guid `foo@baa` was calculated with a hotness of 500 yesterday but today it's not in the bigquery response so it's not updated to 0?

I'm wondering if we need something that mops up any addon with a non-0 hotness that wasn't in bigquery?

src/olympia/addons/cron.py
```diff
@@ -180,29 +178,15 @@ def deliver_hotness():
...
-        # Let the database catch its breath.
-        time.sleep(10)
+    group(
+        [
+            update_addon_hotness.si(
+                averages={key: averages.get(key) for key in keys},
+                frozen_ids=frozen_ids,
+            )
+            for keys in chunked(averages.keys(), chunk_size)
+        ]
```
eviljeff, line 191: as discussed on Matrix, can we use `create_chunked_tasks_signatures` here instead?

src/olympia/addons/cron.py
```diff
@@ -180,29 +178,15 @@ def deliver_hotness():
...
-            if addon.hotness != 0:
-                addon.update(hotness=0)
-
-        # Let the database catch its breath.
-        time.sleep(10)
+    group(
+        [
+            update_addon_hotness.si(
+                averages={key: averages.get(key) for key in keys},
+                frozen_ids=frozen_ids,
```
eviljeff, line 188: if we're doing the frozen query before the tasks, filter out the frozen ids beforehand?  We don't need to pass it down as an arg then either.

src/olympia/addons/tasks.py
willdurand: If an add-on isn't used at all anymore, yes this could probably happen. cc @diox

src/olympia/addons/cron.py
willdurand: yep

src/olympia/addons/tasks.py
diox: Yes that's definitely an issue that needs to be addressed. The previous version of the cron handled that by querying the ids for all add-ons in the database and then only fetching update data for each chunk when looping over them.

willdurand: Should we have something like this before creating the tasks?

```
Addon.objects.filter(status__in=amo.REVIEWED_STATUSES).filter(hotness__gt=0).exclude(guid__in=averages.keys()).update(hotness=0)
```

diox: Probably not a good idea as it would kill hotness for everything while the tasks are being processed - and depending on the order in which they are processed, we'd see a bunch of add-ons with low hotness while the rest is being processed.

willdurand:
@eviljeff I re-requested a review because addressing your comments involved lots of changes

willdurand:
I added a `reset_addon_hotness` task because we have to use `update()` on each add-on to trigger the post save stuff.

src/olympia/addons/cron.py
```diff
@@ -180,29 +179,22 @@ def deliver_hotness():
     threshold = 250 if addon type is theme, else 1000
     hotness = (a-b) / b if a > threshold and b > 1 else 0
     """
-    frozen = set(f.addon_id for f in FrozenAddon.objects.all())
-    averages = get_averages_by_addon_from_bigquery(today=date.today())
-    addons = (
-        Addon.objects.filter(guid__in=averages.keys())
-        .filter(status__in=amo.REVIEWED_STATUSES)
-        .exclude(id__in=frozen)
+    frozen_guids = list(set(fa.addon.guid for fa in FrozenAddon.objects.all()))
```
eviljeff, line 182: we have to wrap this in a `set`?  That's a bit odd.  (Even if there were duplicates, does the break something?  If so, it should probably be in `get_averages_by_addon_from_bigquery`)

eviljeff: 🤷‍♂️ meh though

src/olympia/addons/cron.py
```diff
@@ -180,29 +179,22 @@ def deliver_hotness():
...
+    create_chunked_tasks_signatures(
+        update_addon_hotness, averages.items(), chunk_size
+    ).apply_async()
+
+    # Reset add-ons that won't be returned by BigQuery.
+    addon_ids = (
+        Addon.objects.filter(status__in=amo.REVIEWED_STATUSES)
+        .filter(hotness__gt=0)
+        .exclude(guid__in=averages.keys())
+        .no_transforms()
```
eviljeff, line 195: nit: the transforms won't be executed with a `values_list` so it's not needed.

willdurand: TIL!

Issue #14467, opened 2020-06-02
Intermittent test failure in /lib/es/tests/test_commands.py:test_reindexation_starting_from_one_addon()


bobsilverberg:
I have seen this at least a couple of times on recent PR Travis runs:

```
=================================== FAILURES ===================================

__________ TestIndexCommand.test_reindexation_starting_from_one_addon __________

src/olympia/lib/es/tests/test_commands.py:165: in test_reindexation_starting_from_one_addon

    self._test_reindexation()

        self       = <olympia.lib.es.tests.test_commands.TestIndexCommand testMethod=test_reindexation_starting_from_one_addon>

src/olympia/lib/es/tests/test_commands.py:133: in _test_reindexation

    self.check_results(self.expected)

        ReindexThread = <class 'olympia.lib.es.tests.test_commands.TestIndexCommand._test_reindexation.<locals>.ReindexThread'>

        old_addons_count = 1

        old_indices = [('test_amo_addons-20200602162433', 'test_amo_addons'),

 ('test_amo_addons_stats_download_counts-20200602162433',

  'test_amo_addons_stats_download_counts'),

 ('test_amo_addons_stats_update_counts-20200602162433',

  'test_amo_addons_stats_update_counts')]

        self       = <olympia.lib.es.tests.test_commands.TestIndexCommand testMethod=test_reindexation_starting_from_one_addon>

        t          = <ReindexThread(Thread-1, stopped 140473940182784)>

        wipe       = False

src/olympia/lib/es/tests/test_commands.py:79: in check_results

    assert addon.pk in got, '%s is not in %s' % (addon.pk, got)

E   AssertionError: 1 is not in []

E   assert 1 in []

E    +  where 1 = <Addon: 1: Addôn 1935673f1f1e43a6831543b946c41cd5>.pk

        addon      = <Addon: 1: Addôn 1935673f1f1e43a6831543b946c41cd5>

        expected   = [<Addon: 1: Addôn 1935673f1f1e43a6831543b946c41cd5>,

 <Addon: 2: Addôn 1aa2878bb3604c31b44f4aba997316cc>]

        got        = []

        response   = <Response status_code=200, "application/json">

        self       = <olympia.lib.es.tests.test_commands.TestIndexCommand testMethod=test_reindexation_starting_from_one_addon>

------------------------------ Captured log call -------------------------------

INFO     z.elasticsearch:reindex.py:79 Returning reindexing group for test_amo_addons-20200602162717

[...]

PR #14917, opened 2020-07-11
Update valid system addon guid list and add new guid variations to tests

### LINKED ISSUE #14916: Add `@mozillaonline.com` to the list of restricted GUIDs
In order to have better control of internally-developed add-ons, we should add `@mozillaonline.com` to the [list of system add-on guids](https://github.com/mozilla/addons-server/blob/9f5a1a06a06f5ce11c1170cca3a8d9ca2ee5e85b/src/olympia/constants/base.py#L393).

luckyvs1:
- Fixes #14916 
- Updated valid system guides to include `@mozillaonline.com`
- Updated unit tests to include variations of `@mozillaonline.com`
- Tested by building docker image locally and running `make test` and checking the Travis CI build results

### DIFF
--- a/src/olympia/constants/blocklist.py
+++ b/src/olympia/constants/blocklist.py
@@ -1,5 +1,5 @@
 # How many guids should there be in the stashes before we make a new base.
-BASE_REPLACE_THRESHOLD = 6_000
+BASE_REPLACE_THRESHOLD = 5_000
 
 # Config keys used to track recent mlbf ids
 MLBF_TIME_CONFIG_KEY = 'blocklist_mlbf_generation_time'
--- a/src/olympia/constants/blocklist.py
+++ b/src/olympia/constants/blocklist.py
@@ -1,5 +1,5 @@
 # How many guids should there be in the stashes before we make a new base.
-BASE_REPLACE_THRESHOLD = 5_000
+BASE_REPLACE_THRESHOLD = 6_000
 
 # Config keys used to track recent mlbf ids
 MLBF_TIME_CONFIG_KEY = 'blocklist_mlbf_generation_time'
--- a/src/olympia/constants/base.py
+++ b/src/olympia/constants/base.py
@@ -392,7 +392,8 @@ SYSTEM_ADDON_GUIDS = (
     '@mozilla.org',
     '@pioneer.mozilla.org',
     '@search.mozilla.org',
-    '@shield.mozilla.org'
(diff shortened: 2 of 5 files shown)


luckyvs1:
I had a look through the two failures in Travis CI the test failure I'm unsure of is the following: https://travis-ci.org/github/mozilla/addons-server/jobs/707131541 -- it doesn't seem related to the guid addition, but unsure what caused it (`src/olympia/blocklist/tests/test_cron.py::TestUploadToRemoteSettings::test_stash_because_previous_mlbf`).

The second failure seems to be an ElasticSearch issue: https://travis-ci.org/github/mozilla/addons-server/jobs/707131537 (`src/olympia/lib/es/tests/test_commands.py::TestIndexCommand::test_reindexation_with_wipe`).

luckyvs1:
Took another look, the block list is related to the system guid, so updated the number of expected GUIDs being present the remaining failure during the Travis CI build was related to the following:

`src/olympia/lib/es/tests/test_commands.py::TestIndexCommand::test_reindexation_starting_from_one_addon`

https://travis-ci.org/github/mozilla/addons-server/jobs/707208175

Which I think is related to https://github.com/mozilla/addons-server/issues/14467

src/olympia/constants/blocklist.py
```diff
@@ -1,5 +1,5 @@
 # How many guids should there be in the stashes before we make a new base.
-BASE_REPLACE_THRESHOLD = 5_000
+BASE_REPLACE_THRESHOLD = 6_000
```
eviljeff, line 2: This change isn't connected to the issue.

luckyvs1: I think this parameter is related to the following test and an increment was needed, since system guid was incremented from 5 to 6:

https://travis-ci.org/github/mozilla/addons-server/jobs/707131541 --(src/olympia/blocklist/tests/test_cron.py::TestUploadToRemoteSettings::test_stash_because_previous_mlbf).

I will revert this change and re-verify via the TravisCI build.

eviljeff: It's not related to the test.  You're changing it from 5000 to 6000 and we certainly don't test with that many guids (we mock the threshold down to a lower amount and test with that instead)

luckyvs1: Thank you for the clarification, I have made the update and re-requested a review

PR #14919, opened 2020-07-13
ensure soft-deleted AddonUser instances are filtered out

### LINKED ISSUE #14681: Keep a permanent record of which users were an addon owner/developer
When a user is removed from an Addon we just remove it from the AddonUsers table, which means tracking down previous users means having to query via other data sources (logs, ActivityLog, etc).  

We should either soft-delete AddonUser, or add a further role of "Deleted" (and generally filter it out).  Either way needs to be done with care in regards to only exposing the non-deleted addonusers publicly, and handling a user being re-added back to an addon (which would effectively undelete).

eviljeff:
follow-up fix #14681 - I missed some places where we directly filtered on Addon/UserProfile directly.

### DIFF
--- a/src/olympia/amo/models.py
+++ b/src/olympia/amo/models.py
@@ -521,16 +521,16 @@ class FilterableManyToManyDescriptor(ManyToManyDescriptor):
         filter to the queryset returned via get_queryset."""
         class ManagerWithFiltering(manager):
             def get_queryset(self):
-                # Hook into the queryset caching django uses during these
-                # lookups, or otherwise we get a lot more queries performed.
-                try:
-                    return self.instance._prefetched_objects_cache[
-                        self.prefetch_cache_name]
-                except (AttributeError, KeyError):
-                    pass
+                # Check the queryset caching django uses during these lookups -
+                # we only want to add the q_filter the first time.
+                from_cache = (
+                    self.prefetch_cache_name in
+                    getattr(self.instance, '_prefetched_objects_cache', {}))
                 qs = super().get_queryset()
-                # Here is where we add the filter.
-                return qs.filter(q_filter or models.Q())
+                if not from_cache and q_filter:
+                    # Here is where we add the filter.
+                    qs = qs.filter(q_filter)
+                return qs
 
         return ManagerWithFiltering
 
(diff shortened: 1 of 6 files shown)


src/olympia/amo/models.py
```diff
@@ -521,16 +521,16 @@ def _get_manager_with_default_filtering(cls, manager, q_filter):
...
+                from_cache = (
+                    self.prefetch_cache_name in
+                    getattr(self.instance, '_prefetched_objects_cache', {}))
                 qs = super().get_queryset()
-                # Here is where we add the filter.
-                return qs.filter(q_filter or models.Q())
+                if not from_cache and q_filter:
+                    # Here is where we add the filter.
+                    qs = qs.filter(q_filter)
+                return qs
```
eviljeff, line 533: Not _strictly_ a fix for the issue, but it made it clearer to me what the logic actually is.

src/olympia/devhub/tests/test_views.py
```diff
@@ -57,6 +57,11 @@ def setUp(self):
         assert self.client.login(email='regular@mozilla.com')
         assert self.client.get(self.url).status_code == 200
         self.user_profile = UserProfile.objects.get(id=999)
+        not_their_addon = addon_factory(users=[user_factory()])
```
bobsilverberg, line 60: Nit: It actually *is* their addon, isn't it? It's just that the `AddonUser` record has them listed as deleted.

src/olympia/devhub/feeds.py
```diff
@@ -31,7 +30,7 @@ def items(self, key):
         if key.addon:
             addons = key.addon
         else:  # We are showing all the add-ons
-            addons = Addon.objects.filter(authors=key.user)
+            addons = key.user.addons.all()
```
bobsilverberg, line 33: Is there a test for this change? Is it covered by the change in ` src/olympia/devhub/tests/test_views.py`?

src/olympia/devhub/tests/test_views.py
eviljeff: It's not their add-on _any more_  😉 - they've been deleted from the addon owners.  (A user would never be added as deleted in reality)

src/olympia/devhub/feeds.py
eviljeff: yeah, I added to the `setUp` for the base `HubTest` which is used by the feeds tests too.

Issue #14934, opened 2020-07-14
Retry MAD API call on 50x errors


willdurand:
We should really have a MAD response so let's add a retry strategy.

For QA: all submitted versions should have a (MAD) score.

AlexandraMoga:
I've submitted several add-ons and there was a score for the `mad`* and the `customs` scanner for each submitted versions.

*excluding themes and language tools which do not produce a score

PR #14935, opened 2020-07-14
Add retry strategy in call_mad_api()

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14934

---

[We can't use responses to test the retry strategy](https://github.com/getsentry/responses/issues/135) unfortunartely, so there is no new test case. I tried locally, though.

### DIFF
--- a/src/olympia/scanners/tasks.py
+++ b/src/olympia/scanners/tasks.py
@@ -8,6 +8,8 @@ import waffle
 import yara
 
 from django_statsd.clients import statsd
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util import Retry
 
 import olympia.core.logger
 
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/tasks.py
```diff
@@ -438,13 +440,25 @@ def call_mad_api(all_results, upload_pk):
...
-            response = requests.post(
-                url=settings.MAD_API_URL,
-                json=json_payload,
-                timeout=settings.MAD_API_TIMEOUT,
-                headers={'x-request-id': request_id},
-            )
+            with requests.Session() as http:
+                adapter = HTTPAdapter(
+                    max_retries=Retry(
+                        total=1,
```
bobsilverberg, line 446: Does this mean we're retrying once, and then giving up?

willdurand: yes

Issue #14933, opened 2020-07-14
Show MAD score in reviewer tools (detail page)


willdurand:
Above the "Weight" in the add-ons history, let's show the MAD score and add some help text too.

AlexandraMoga:
The Scanner score is displayed in the add-on history*
![image](https://user-images.githubusercontent.com/31961530/87763590-b2f18180-c81d-11ea-9d61-061d4848155f.png)


*Note: versions without a score or older versions that were submitted before the score was implemented will still display the Scanner score field with `n/a` as value
Verified fixed on -dev.

PR #14944, opened 2020-07-15
Show MAD score in reviewer tools

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14933


willdurand:
I updated this patch to use 'Scanners score'.

PR #14946, opened 2020-07-15
RE: Add the ability to create, update, and delete shelves

### LINKED ISSUE #14872: Add the ability to create, update, and delete shelves
Supports #14871 

Add the ability to create, update, and delete the shelves that will appear on the AMO homepage by creating a Shelf model. Each shelf will contain fields such as title, type, criteria, footer text, footer link, etc.

xlisachan:
Fixes #14872 

This pull request creates the Shelf model, which can be found under Discovery in admin. Tests have also been added to `discovery/tests/test_admin.py`.

Per yesterday's discussion, validations will be included in a separate issue.

Screen Shot 2020-07-15 at 10 52 38 AM

Screen Shot 2020-07-15 at 10 54 31 AM


bobsilverberg:
This looks great, thanks @xlisachan! Do you think it's worth having an example under `Footer pathname`, similar to what we have under `Criteria`?

xlisachan:
`help_text` has been added for the footer pathname. I think incorporating `settings.EXTERNAL_SITE_URL`, language ('en-US'), and clientApp ('firefox') would be applied later on and isn't necessary to place within the model at this point, correct?

Screen Shot 2020-07-15 at 3 01 09 PM

src/olympia/shelves/models.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Shelf(ModelBase):
+    title = models.CharField(max_length=200, unique=True)
+    shelfType = models.CharField(
+        max_length=200, choices=CHOICE_TYPES, verbose_name='type')
+    criteria = models.CharField(
+        max_length=200,
+        help_text="e.g., search/?recommended=true&sort=random&type=extension")
+    footerText = models.CharField(
+        max_length=200, default="See more", verbose_name="footer text")
```
eviljeff, line 23: lets not specify the `default` here - it makes it annoying to change (needing database migrations) if we were to update it later from "See more" to a different string.  Just have `blank=True`.  In the serializer we can check if it's not defined and return `'See more'` there instead.

src/olympia/shelves/models.py
```diff
@@ -0,0 +1,32 @@
...
+    ('extension', 'extension'),
+    ('recommended', 'recommended'),
+    ('search', 'search'),
+    ('theme', 'theme'),
+)
+
+
+class Shelf(ModelBase):
+    title = models.CharField(max_length=200, unique=True)
+    shelfType = models.CharField(
```
eviljeff, line 17: 😄 too much javascript - python prefers snake_case rather than camelCase for it's class attributes.

eviljeff: the same for the other properties

src/olympia/shelves/models.py
```diff
@@ -0,0 +1,32 @@
...
+        max_length=200, choices=CHOICE_TYPES, verbose_name='type')
+    criteria = models.CharField(
+        max_length=200,
+        help_text="e.g., search/?recommended=true&sort=random&type=extension")
+    footerText = models.CharField(
+        max_length=200, default="See more", verbose_name="footer text")
+    footerPathname = models.CharField(
+        max_length=1000, verbose_name="footer pathname",
```
eviljeff, line 25: a 1000 max_length is little extreme.  is 255 not long enough?

eviljeff, line 25: You won't need to specify `verbose_name`  when it's `footer_pathname` either.

src/olympia/shelves/models.py
```diff
@@ -0,0 +1,32 @@
...
+from olympia.amo.models import ModelBase
+
+CHOICE_TYPES = (
+    ('category', 'category'),
+    ('collection', 'collection'),
+    ('extension', 'extension'),
+    ('recommended', 'recommended'),
+    ('search', 'search'),
+    ('theme', 'theme'),
+)
```
eviljeff, line 12: Mmm, we might need this list elsewhere, so instead lets have a flat tuple of the shelf types `SHELF_TYPES = ('category', 'collection', ...)` and then `SHELF_TYPE_CHOICES = tuple((ty, ty) for ty in SHELF_TYPES)` for the field choices

src/olympia/shelves/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class ShelvesConfig(AppConfig):
+    name = 'shelves'
```
eviljeff, line 5: We don't need this file?

src/olympia/shelves/models.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class Shelf(ModelBase):
+    title = models.CharField(max_length=200, unique=True)
```
eviljeff, line 16: don't set unique here - there's a possible scenario where Scott might want to have two shelves with the same title temporarily (maybe creating a replacement while keeping the old shelf live).

src/olympia/shelves/models.py
eviljeff: Add help_text mentioning the default.

src/olympia/discovery/tests/test_admin.py
```diff
@@ -1038,3 +1039,190 @@ def test_need_3_modules(self):
...
+                'footerText': 'See more',
+                'footerPathname': '/this/is/the/pathname'
+            },
+            follow=True)
+        assert response.status_code == 200
+        assert Shelf.objects.count() == 1
+        item = Shelf.objects.get()
+        assert item.title == 'Recommended extensions'
+
+    def test_can_not_add_with_discovery_edit_permission(self):
```
bobsilverberg, line 1155: Should this read "test_can_not_add_with*out*_discovery_edit_permission"?

src/olympia/shelves/models.py
xlisachan: 😅 - updated!

src/olympia/shelves/models.py
xlisachan: Yea I may have gotten carried away with setting the max_length on that. Updated! :)

src/olympia/discovery/tests/test_admin.py
xlisachan: Yes, yes it should. Good eyes! Updated!

xlisachan:
Thanks for the feedback :) The following updates have been made:
- Revised camelCase to snake_case for all properties
- Removed `default` and `verbose_name` from footer_text and added `help_text`
- Updated `max_length` for footer_pathname and removed `verbose_name`
- Added `SHELF_TYPES` and `SHELF_TYPE_CHOICES`; revised value for `choices=` to `SHELF_TYPE_CHOICES`
- Deleted unnecessary `shelves/apps.py` file
- Removed `unique=True` from `title` 
- Fixed test name to `test_can_not_add_without_discovery_edit_permission`

src/olympia/shelves/migrations/0002_auto_20200716_1254.py
```diff
@@ -0,0 +1,46 @@
...
+    operations = [
+        migrations.RenameField(
+            model_name='shelf',
+            old_name='shelfType',
+            new_name='shelf_type',
+        ),
+        migrations.RemoveField(
+            model_name='shelf',
+            name='footerPathname',
+        ),
```
eviljeff, line 21: It's not worth doing here because it's all being landed at once with no existing data but this migration would need manually fixing if 0001 was already deployed.

Django hasn't picked up that you've renamed the field and altered it at the same time so it thinks you've deleted the field and then added a new one.  To fix you'd change the AddField to an AlterField and then move and replace RemoveField with RenameField after it

xlisachan: Makes sense, I'll be sure to check the migrations moving forward!

PR #14951, opened 2020-07-15
add PromotedAddon.application

### LINKED ISSUE #14941: Add application support to Promoted models
from the prd comments

> We  expect to have two distinct lists of Recommended add-ons for desktop  and mobile... Since an add-on could belong to both lists and so far promoted lists are exclusive, I  think it might be better to have a single Recommended list and then a  per-recommended-add-on application setting, but that's an engineering  decision.

eviljeff:
fixes #14941


src/olympia/addons/models.py
```diff
@@ -1307,16 +1307,30 @@ def is_recommended(self):
                 addon=self).exists()
         return recommended
 
-    @cached_property
-    def promoted_group(self):
-        """If the addon is promoted then return the Promoted group details."""
+    def is_promoted(self, group=None, application=None, currently=True):
+        """Is the addon currently promoted for the specified group and
+        application.
```
willdurand, line 1312:
```suggestion
        application?
```

😁

src/olympia/addons/models.py
```diff
@@ -1307,16 +1307,30 @@ def is_recommended(self):
...
+
+        `group` is the PromotedClass class; if group=None (default) then all
+        promotion groups are considered;
+        `application` is the App class; if application=None (default) then all
+        applications are considered;
+        `currently=True` means only returns True if self.current_version is
```
willdurand, line 1318: I find this argument name not very self-explanatory.

eviljeff: I'm open to better naming.

bobsilverberg: `currently_approved` is a suggestion

src/olympia/addons/models.py
```diff
@@ -1307,16 +1307,30 @@ def is_recommended(self):
                 addon=self).exists()
         return recommended
 
-    @cached_property
-    def promoted_group(self):
-        """If the addon is promoted then return the Promoted group details."""
+    def is_promoted(self, group=None, application=None, currently=True):
```
bobsilverberg, line 1310: This method seems kind of complicated and convoluted, which is borne out in the test. I'm wondering why we need it to be this way. Can you describe the use cases you see for it needing to accept a `group`? I'm wondering why we cannot just have a function that returns the promoted group (as this function used to) and accepts the `application` as an argument? It seems to me like that would deliver the same data, and be less complicated.

As well, what are the use cases for `currently`?

src/olympia/addons/tests/test_models.py
```diff
@@ -1592,42 +1592,58 @@ def test_theme_is_recommended(self):
...
-        assert addon.promoted_group == VERIFIED_ONE
-
-        PromotedApproval.objects.all().delete()
-        del addon.promoted_group
-        # if there aren't any approvals there's no group returned
-        assert not addon.promoted_group
+        assert addon.is_promoted(group=VERIFIED_ONE)
+
+        # Application specific group membership should work too
+        # if no app is specifed in the PromotedAddon everything should match
```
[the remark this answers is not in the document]
eviljeff: It's implied it'll work but I guess only because of it's current implementation so could do with some more tests.

src/olympia/addons/models.py
eviljeff: > Can you describe the use cases you see for it needing to accept a group? 

To make it a drop in replacement for `is_recommended = addon.is_recommended` -> `is_recommended = addon.is_promoted(group=RECOMMENDED, application=amo.FIREFOX)`.  Otherwise it's `group = addon.promoted_group(application=amo.FIREFOX); is_recommended = group and group == RECOMMENDED`.  

On the other side, if we _do_ need the group details it becomes `group = addon.promotedaddon.group if addon.is_promoted() else None`.  Do you think I should re-add `def promoted_group` that just does that?

> As well, what are the use cases for currently?

mmm, it's difficult to know because I'm writing the utility functions and model code before anything that uses it.  The API wouldn't use it.  But there are some places in the reviewer/admin tools that check `DiscoveryItem.recommendable` at the moment and `addon.is_promoted(group=RECOMMENDED, currently=False)` would be a drop in for that (possibly without the group in the future).  Though I can drop it for now and add it back later if becomes needed.  WDYT?

bobsilverberg: I hadn't thought of all of the current uses of `is_recommended`. I think it's fine to leave as is for now, and enhance later if we find it isn't quite working for us.

PR #14964, opened 2020-07-16
reimplement description in PrimaryHeroShelf object

### LINKED ISSUE #14947: move/copy description_text from DiscoveryItem to PrimaryHero
in the serializer for `PrimaryHero` we return details about the addon, and `description_text` from `DiscoveryItem`.  When `PrimaryHero` links to a `PromotedAddon` instead we won't have that data so we should move/copy* it to the `PrimaryHero` model.  *not sure if it's used for the discovery serializer too currently or not.

eviljeff:
fixes #14947


src/olympia/hero/serializers.py
```diff
@@ -33,6 +34,15 @@ def to_representation(self, instance):
         rep.pop('addon' if instance.is_external else 'external')
         return rep
 
+    def get_description(self, obj):
+        if obj.description:
+            return ugettext(obj.description)
+        else:
+            addon = obj.disco_addon.addon
+            return (str(addon.summary)
```
bobsilverberg, line 42: Where does this requirement come from? Was that baked into `serializers.CharField(source='disco_addon.description_text')`?

eviljeff: yeah, check `DiscoveryItem.description_text`.

src/olympia/discovery/management/commands/extract_content_strings.py
```diff
@@ -72,6 +72,12 @@ class DiscoItemAPIParser(BaseAPIParser):
     fields = ('custom_heading', 'custom_description')
 
 
+class PrimaryHeroShelfAPIParser(BaseAPIParser):
+    api = settings.PRIMARY_HERO_EDITORIAL_CONTENT_API
+    l10n_comment = 'editorial content for the primary hero shelves.'
+    fields = ('description',)
```
eviljeff, line 78: actually, we want to be translating the raw description field on the model... mmm.
- a special view&serializer just for exposing description (like DiscoItemAPIParser)?
- a `?raw` parameter on https://addons-server.readthedocs.io/en/latest/topics/api/hero.html#get--api-v4-hero-primary- ?
- return an extra `custom_description` or `description_raw` field if `?all=True`?
- something else?

diox: I'd go with the extra special view I think, just to make it consistent with the way we do it for discovery items, but I don't care about this very strongly...

eviljeff: added the ?raw param as it seemed the least disruptive.

Issue #14942, opened 2020-07-15
Raise errors in scanner tasks to abort the submission process


willdurand:
We'd like to abort the submission process when a scanner raises an error so we should re-raise errors we handle in the different tasks.

AlexandraMoga:
@willdurand do you have some STR for this?

willdurand:
> @willdurand do you have some STR for this?

I think uploading a web extension in -dev would be a good start. It should likely fail because of one of our scanners (which seems to often fail lately). After that, you can ping me and I'll disable this scanner, and re-uploading the same extension should work.

AlexandraMoga:
I wasn't able to reproduce unexpected upload errors in -dev so I'm not sure how we can verify this.

AlexandraMoga:
I've verified this on with an add-on example and I was able to reproduce the upload error. If I'm submitting the same add-on with the problematic scanner disabled, the error is gone.

I think this can be marked as verified fixed on -dev.

PR #14965, opened 2020-07-16
Raise exceptions in scanner tasks

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14942

---

With this patch, we get a generic error:

Screen Shot 2020-07-16 at 12 34 51

I added a kill switch too, in case something goes wrong...

Also, we have a retry strategy for the MAD call, so I didn't add anything. The MAD call could also be done later so it's less of a problem I think (for now).

### DIFF
--- a/src/olympia/scanners/tasks.py
+++ b/src/olympia/scanners/tasks.py
@@ -85,12 +85,12 @@ def run_scanner(results, upload_pk, scanner, api_url, api_key):
         statsd.incr('devhub.{}.success'.format(scanner_name))
         log.info('Ending scanner "%s" task for FileUpload %s.', scanner_name,
                  upload_pk)
-    except Exception:
+    except Exception as exc:
         statsd.incr('devhub.{}.failure'.format(scanner_name))
-        # We log the exception but we do not raise to avoid perturbing the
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/tasks.py
```diff
@@ -182,6 +182,10 @@ def run_yara(results, upload_pk):
        results as first argument.
     - `upload_pk` is the FileUpload ID.
     """
+    return _run_yara(results, upload_pk)
+
+
+def _run_yara(results, upload_pk):
```
bobsilverberg, line 188: What does this change (moving all the code into a separate function) do? Does the function where the exception is raised need to be *not* a `@validation_task`?

bobsilverberg: Oh, after looking at the tests I think this is just to support the tests, or maybe not?

willdurand: yes, exactly. `@validation_task` catches all exceptions so the new test cases wouldn't show that we actually raise exceptions there.

willdurand:
> Is that why you requested review from both @diox and me?

Yes :)

Issue #14899, opened 2020-07-09
Remove obsolete views/code about expired info requests


diox:
With delayed review rejections the notion of "Information request" (which is essentially a reviewer reply with a deadline associated to it) will no longer exist. There will be some code to remove:
- The cron that notifies developers about soon to expire info requests (https://github.com/mozilla/addons-server/issues/14814 will implement the same thing for delayed rejections) and the `notified_about_expiring_info_request` field in `AddonReviewerFlags`
- The expired info requests reviewer queue (https://github.com/mozilla/addons-server/issues/14576 will implement a queue for add-ons with versions pending rejection)
- The code to create an info request (will probably be removed in https://github.com/mozilla/addons-server/issues/14454 already)
- The `pending_info_request` field in `AddonReviewerFlags`.

diox:
Reviewer tools & devhub should continue to work as normal, this change should be completely transparent to users.

AlexandraMoga:
I've verified that actions related to former Information requests are no longer present and did a bit of regression around the still available Reviewer replies:

- reviewer replies can still be performed
- developers are receiving emails when a reviewer reply is posted or they can use the version review history on the Manage versions page to communicate with the reviewer 
- the Expired information request queue has been removed from the reviewers dashboard
- the Pending information request button will no longer appear in the Admin actions section
- the Pending information request icon is not displayed next to add-ons that have an unresolved information request 

This issue is verified fixed on -dev.

PR #14968, opened 2020-07-16
Implement delayed review rejections in reviewer tools

### LINKED ISSUE #14454: Implement delayed review rejections

diox:
This removes the ability to have an info request with a deadline, and instead makes it possible to delay review rejections when using the reject multiple versions action.

reject_delay

Fixes #14454


src/olympia/reviewers/tests/test_utils.py
```diff
@@ -1584,6 +1517,72 @@ def test_reject_multiple_versions(self):
...
+        assert list(self.addon.versions.all()) == [self.version, old_version]
+        assert self.file.status == amo.STATUS_APPROVED
+
+        # The versions are now pending rejection.
+        for version in self.addon.versions.all():
+            assert version.pending_rejection
+            self.assertCloseToNow(version.pending_rejection, now=in_the_future)
+
+        assert len(mail.outbox) == 1
+        assert mail.outbox[0].to == [self.addon.authors.all()[0].email]
```
bobsilverberg, line 1560: Nit: Would a `email = mail.outbox[0]` here make the following code a bit more succinct?

src/olympia/reviewers/tests/test_utils.py
```diff
@@ -1690,6 +1689,79 @@ def test_reject_multiple_versions_content_review(self):
...
```
bobsilverberg, line 1739: Nit: Same comment as above about creating a shortcut to `mail.outbox[0]`.

src/olympia/reviewers/utils.py
```diff
@@ -1016,25 +991,65 @@ def reject_multiple_versions(self):
...
+            pending_rejection_deadline = now + timedelta(
+                days=int(self.data['delayed_rejection_days'])
+            )
+        else:
+            pending_rejection_deadline = None
+        if pending_rejection_deadline:
+            action_id = (amo.LOG.REJECT_CONTENT_DELAYED if self.content_review
+                         else amo.LOG.REJECT_VERSION_DELAYED)
+            log.info(
+                'Marking %s versions %s for delayed rejection' % (
```
bobsilverberg, line 1005: I know it wasn't there before, but is there any value in adding an indicator of whether this is a content or version rejection to this log message?

src/olympia/reviewers/forms.py
```diff
@@ -370,19 +382,19 @@ def __init__(self, *args, **kw):
...
-        deadline_widget_attributes.update({
-            'min': info_request_deadline.min_value,
-            'max': info_request_deadline.max_value,
-        })
-        info_request_deadline.widget.attrs.update(deadline_widget_attributes)
+            rejection_period.min_value = rejection_period.initial
+            rejection_period.max_value = rejection_period.initial
+            rejection_period_widget_attributes['readonly'] = 'readonly'
+        rejection_period_widget_attributes['min'] = rejection_period.min_value
+        rejection_period_widget_attributes['max'] = rejection_period.max_value
```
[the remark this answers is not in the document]
diox: If we disable the widget, then the value isn't sent via POST, plus it looks kinda weird in the UI (it will show the greyed out input which may make the value hard to read)

src/olympia/reviewers/forms.py
```diff
@@ -370,19 +382,19 @@ def __init__(self, *args, **kw):
         self.helper = kw.pop('helper')
         super(ReviewForm, self).__init__(*args, **kw)
 
-        # Info request deadline needs to be readonly unless we're an admin.
+        # Delayed rejection period needs to be readonly unless we're an admin
```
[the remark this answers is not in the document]
diox: No good reason, adding it back

src/olympia/activity/tests/test_utils.py
```diff
@@ -310,8 +309,7 @@ def _check_email(self, call, url, reason_text):
     def test_reviewer_request_for_information(self, send_mail_mock):
         AddonReviewerFlags.objects.create(
             addon=self.addon,
-            pending_info_request=datetime.now() + timedelta(
-                days=REVIEWER_NEED_INFO_DAYS_DEFAULT))
+            pending_info_request=datetime.now() + timedelta(days=14))
```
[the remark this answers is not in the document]
diox: Because it's not the same thing. I'm removing the old constant because it was meant for info requests through reviewer actions and I'm removing the ability to do that, but for now the rest of the code around info requests is kept (will be removed fully in https://github.com/mozilla/addons-server/issues/14899). Any arbitrary value would work for this test as long as it's consistent.

src/olympia/reviewers/utils.py
diox: I don't think it matters much. We can find the info later if we want in `ActivityLog`, this logging is more useful to say "we've gone through the code that sets delayed rejection flag, if it's not there something is wrong".

src/olympia/reviewers/tests/test_utils.py
diox: There were quite a few tests to patch... but I think yeah, it's worth doing. 9343bbd66aeda3c992c09d7093b9eca89a041086

src/olympia/reviewers/tests/test_utils.py
diox: 9343bbd66aeda3c992c09d7093b9eca89a041086

Issue #14931, opened 2020-07-14
Expose model version in scanners API response


willdurand:
We'd need the model version to be returned in the scanners API response.

willdurand:
cc @dzeber

AlexandraMoga:
The `model_version` field is exposed in scanners api results with value `null` for `mad`, `yara` and `wat`. The field will have a string value for `customs`. 
Note that `customs` will have a `model_version` value only for new submitted add-ons. Older `customs` results will have `model_version: null` 

Verified fixed on -dev.

PR #14970, opened 2020-07-16
Expose model_version in scanners API response

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14931


docs/topics/api/scanners.rst
```diff
@@ -28,3 +28,4 @@ This endpoint returns a list of labelled scanner results.
     :>json string label: Either ``good`` or ``bad``.
     :>json object results: The scanner (raw) results.
     :>json string created: The date the result was created, formatted with `this format <http://ecma-international.org/ecma-262/5.1/#sec-15.9.1.15>`_.
+    :>json string|null model_version: The model version when applicable, ``null`` otherwise.
```
eviljeff, line 31: this sentence doesn't really mean anything to someone who doesn't understand the whole scanner ecosystem.  What's a model, what does it's version mean, and when is it applicable?

src/olympia/scanners/tasks.py
```diff
@@ -478,13 +478,13 @@ def call_mad_api(all_results, upload_pk):
             score=data.get('ensemble', default_score),
```
eviljeff, line 478: so, you get the `model_version` from `MAD_API_URL` results and add it the `ScannerResult(scanner=CUSTOMS)` record, but not the `ScannerResult(scanner=MAD)` record?

docs/topics/api/scanners.rst
willdurand: Vagueness is a feature here.

src/olympia/scanners/tasks.py
willdurand: Yes. This `model_version` is specific to customs results. We still have the raw response stored on the mad results, though.

PR #14971, opened 2020-07-16
RE: Store uploaded hero images as JPGs 

### LINKED ISSUE #14948: Store uploaded hero images as JPGs
Store primary hero images to .jpg instead of .png, as png images may be too large and impact the size of homepage assets.

xlisachan:
Fixes #14948 

This pull request converts the uploaded image for the primary hero area to jpg by updating `resize_image` in `utils.py`.

`discovery/tests/test_admin/.py` has been updated, and I also ran other tests that include `resize_image` (`amo/tests/test_amo_utils.py` and `versions/tests/test_tasks.py`) and they passed based on this revision as well.

In addition, please see below for the difference in size between the initial uploaded and resulting jpg for a few of the existing images:
201 KB -> 156 KB
179 KB -> 169 KB
39 KB -> 38 KB

### DIFF
--- a/src/olympia/amo/utils.py
+++ b/src/olympia/amo/utils.py
@@ -624,7 +624,7 @@ def pngcrush_image(src, **kw):
     return False
 
 
-def resize_image(source, destination, size=None):
+def resize_image(source, destination, size=None, hero_image=False):
     """Resizes and image from src, to dst.
     Returns a tuple of new width and height, original width and height.
 
@@ -641,11 +641,17 @@ def resize_image(source, destination, size=None):
         original_size = im.size
         if size:
             im = processors.scale_and_crop(im, size)
-    with storage.open(destination, 'wb') as fp:
-        # Save the image to PNG in destination file path. Don't keep the ICC
-        # profile as it can mess up pngcrush badly (mozilla/addons/issues/697).
-        im.save(fp, 'png', icc_profile=None)
-    pngcrush_image(destination)
+    if hero_image:
+        with storage.open(destination, 'wb') as fp:
+            im = im.convert('RGB')
+            im.save(fp, 'JPEG')
+    else:
+        with storage.open(destination, 'wb') as fp:
+            # Save the image to PNG in destination file path.
+            # Don't keep the ICC profile as it can mess up pngcrush badly
+            # (mozilla/addons/issues/697).
+            im.save(fp, 'png', icc_profile=None)
+        pngcrush_image(destination)
     return (im.size, original_size)
 
 
--- a/src/olympia/amo/utils.py
+++ b/src/olympia/amo/utils.py
@@ -624,7 +624,7 @@ def pngcrush_image(src, **kw):
     return False
 
 
-def resize_image(source, destination, size=None, hero_image=False):
+def resize_image(source, destination, size=None, format='png'):
     """Resizes and image from src, to dst.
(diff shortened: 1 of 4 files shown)


src/olympia/amo/utils.py
```diff
@@ -624,7 +624,7 @@ def pngcrush_image(src, **kw):
     return False
 
 
-def resize_image(source, destination, size=None):
+def resize_image(source, destination, size=None, hero_image=False):
```
eviljeff, line 627: how about the kwarg is just the format, so `format=png` and if it's not `'png'` we just pass it directly to the save, like `im.save(fp, format)`?  That makes it a bit more general in case we decide to use this for other stuff beyond the hero shelves.

xlisachan:
Updated kwarg with `format='png'`. All tests mentioned above passed with this revision as well. Please review when you have a moment. Thanks!

src/olympia/amo/utils.py
```diff
@@ -641,11 +641,17 @@ def resize_image(source, destination, size=None):
...
             im = processors.scale_and_crop(im, size)
-    with storage.open(destination, 'wb') as fp:
-        # Save the image to PNG in destination file path. Don't keep the ICC
-        # profile as it can mess up pngcrush badly (mozilla/addons/issues/697).
-        im.save(fp, 'png', icc_profile=None)
-    pngcrush_image(destination)
+    if not format:
+        with storage.open(destination, 'wb') as fp:
+            im = im.convert('RGB')
+            im.save(fp, 'JPEG')
```
diox, line 647: Default quality for `JPEG` format in Pillow is 75. That's certainly going to make the file size small, but maybe it's a bit too aggressive...

src/olympia/amo/utils.py
```diff
@@ -641,11 +641,17 @@ def resize_image(source, destination, size=None):
         original_size = im.size
         if size:
             im = processors.scale_and_crop(im, size)
-    with storage.open(destination, 'wb') as fp:
-        # Save the image to PNG in destination file path. Don't keep the ICC
-        # profile as it can mess up pngcrush badly (mozilla/addons/issues/697).
-        im.save(fp, 'png', icc_profile=None)
-    pngcrush_image(destination)
+    if not format:
```
eviljeff, line 644: this will actually end up with all images being saved as png again because when `format='jpeg'` it's not, `not format`.  When `format == 'png'` you want to do something special (specify the iccprofile and pngcrush), else you want to just convert and save with the format specified.

src/olympia/amo/utils.py
xlisachan: I mostly tested the images currently in `static/img/hero/featured` that are already 960x640 jpgs. I think `quality` should be set to 80, as the majority of those file sizes remain the same upon upload with this setting. Thoughts?

src/olympia/amo/utils.py
xlisachan: That makes sense now. Apologies for the oversight!!

xlisachan:
Thanks for the feedback! Updates include correcting the conditional in `resize_image` and setting the `quality` for the `JPEG` format to 80. Please review when you have a moment. Thank you!

src/olympia/amo/utils.py
diox: 80 sounds good. We can bump to 90 if we notice issues with image quality.

Issue #14945, opened 2020-07-15
Roll back trademark restriction for add-on names


jvillalobos:
We are running into some issues with the restriction we enabled in https://github.com/mozilla/addons-server/issues/14656 and https://github.com/mozilla/addons-linter/issues/2158, so we need to roll them back.

willdurand:
@jvillalobos do we plan on re-enabling this check later? or can we remove everything in a more permanent manner?

jvillalobos:
You can remove it in a more permanent manner. We _might_ implement something like it in the future, but if we do it will be more nuanced, allowing certain things for themes and others for extensions. It's not in the plans, though.

jvillalobos:
@willdurand do we need a separate issue for the linter?

willdurand:
> @willdurand do we need a separate issue for the linter?

Probably but I was going to file one before submitting a patch.

ioanarusiczki:
@willdurand I verified on AMO dev with FF78(Win10) with a regular account (using a gmail e-mail address in my case) and it allowed `for Mozilla` and `for Firefox` in the `Name` field for add-ons or themes.

The only case I still receive the red warning is when I use both trademarks such as `Bubbles for Mozilla Firefox`.  The issue is reproducible at addon submission too. 

No problems encountered when using a mozilla account.

willdurand:
cc @wagnerand and @jvillalobos

jvillalobos:
That should be okay. I believe that's what we did before.

willdurand:
> That should be okay. I believe that's what we did before.

Theoretically, yes.

PR #14973, opened 2020-07-16
Make trademark checks less strict

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14945

### DIFF
--- a/src/olympia/addons/utils.py
+++ b/src/olympia/addons/utils.py
@@ -9,6 +9,8 @@ from django.conf import settings
 from django.forms import ValidationError
 from django.utils.translation import ugettext
 
+import waffle
+
 from olympia import amo
 from olympia.amo.utils import normalize_string, to_language
 from olympia.discovery.utils import call_recommendation_server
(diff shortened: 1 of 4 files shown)


willdurand:
Test failures: ES tests

willdurand:
Ha wait, looks like we still need some checks.

willdurand:
Patch updated. cc @wagnerand

src/olympia/addons/utils.py
```diff
@@ -28,7 +28,12 @@ def _check(name):
         name = normalize_string(name, strip_punctuation=True).lower()
 
         for symbol in amo.MOZILLA_TRADEMARK_SYMBOLS:
-            if symbol in name:
+            violates_trademark = (
+                name.count(symbol) > 1 or (
+                    name.count(symbol) >= 1 and not
+                    name.endswith(' for {}'.format(symbol))))
```
diox, line 34: I'm confused, this doesn't *remove* the trademark restriction code, but patches it ?

willdurand: Yeah, looks like we need to remove some restrictions but not everything. I reverted the previous patch that added stricter checks, and removed the waffle switch that was used to enable/disable the stricter checks.

PR #14988, opened 2020-07-17
rewrite Addon.is_recommended to use PromotedAddon.

### LINKED ISSUE #14859: Refactor recommended add-ons to be a PromotedAddon (rather than a type of DiscoveryItem)
>Initial backend work to refactor DiscoveryItem.recommendable + Version.recommendation_approved into separate models with different promotion states (both in database+elasticsearch)

depends on #14858 to add the new models

eviljeff:
fixes #14859 - sorry, I couldn't see a way of splitting it up any further!  It aims to not change any current behaviour, just how the recommended system is implemented in the models.

This patch removes the inline admin for PrimaryHero shelves so until a follow-up is landed there won't be any way to add/view/edit/delete PrimaryHero shelves (the patch was already too complicated).  Ideally the patch to re-implement the admin for PrimaryHero shelves will be implemented and landed in the same tag!


src/olympia/hero/admin.py
```diff
@@ -22,14 +22,15 @@ def label_from_instance(self, obj):
                 obj.preview_url))
 
 
-class PrimaryHeroInline(admin.StackedInline):
+class PrimaryHeroAdmin(admin.ModelAdmin):
```
eviljeff, line 25: This isn't used (or tested) anywhere yet as per the PR summary

src/olympia/versions/migrations/0009_auto_20200717_1137.py
```diff
@@ -0,0 +1,18 @@
...
+    dependencies = [
+        ('versions', '0008_auto_20200625_1114'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='version',
+            name='recommendation_approved',
+            field=models.BooleanField(null=True),
+        ),
```
eviljeff, line 17: just realised this is missing a RunPython to add a `PromotedApproval` if `version.recommendation_approved`

diox: This is still missing ?

eviljeff: https://github.com/mozilla/addons-server/pull/14988/commits/d5856a3046e1a08679cecff4b80e7c4eed21278d

Issue #14982, opened 2020-07-17
Enable stats for all add-on types we currently support


willdurand:
Once https://bugzilla.mozilla.org/show_bug.cgi?id=1653499 is fixed, we'll have data for all add-ons in BigQuery, which would allow us to enable the stats feature for all add-ons and not just web-extensions and themes.

willdurand:
For QA: I am still working on getting langpack data in -dev/-stage. For now, we'll have to update the GUID of an existing langpack (and/or dict) with [one of those](https://github.com/mozilla/bigquery-etl/blob/c8d08ace85cba1b2fad445daa6a67ec74bc3a57f/sql/amo_dev/amo_stats_dau_v2/query.sql#L14-L24)

ioanarusiczki:
@willdurand test results:
- I used an id from the [samples above](https://github.com/mozilla/bigquery-etl/blob/c8d08ace85cba1b2fad445daa6a67ec74bc3a57f/sql/amo_dev/amo_stats_dau_v2/query.sql#L14-L24) on dev for a [lang pack](https://addons-dev.allizom.org/en-US/firefox/addon/lang-pack-with-statistics/) and a [dictionary ](https://addons-dev.allizom.org/en-US/firefox/addon/release-dictionary-with-stats/). In this case I can see data including the average ADU which is now displayed on the frontend too instead of the "Downloads".

- with guids from prod for lang packs I checked these:
https://addons-dev.allizom.org/en-US/firefox/addon/asturianu-language-pack/statistics/?last=30
https://addons-dev.allizom.org/en-US/firefox/addon/czech-cz-language-pack/statistics/?last=30

They show data except that the ADU (the blue link) is 0, which is also shown as "no users" on the frontend. I don't know why.

- with guids from prod for dictionaries I did not find data though. I looked for stats into the following:
https://addons-dev.allizom.org/en-US/firefox/addon/afrikaans-spell-checker/
https://addons-dev.allizom.org/en-US/firefox/addon/bulgarian-dictionary/

willdurand:
> with guids from prod for dictionaries I did not find data though. I looked for stats into the following:

Oh right, we don't have data for dictionaries.

ioanarusiczki:
@willdurand Yes. Are the lang packs with guids from prod alright with that ADU 0?

willdurand:
> @willdurand Yes. Are the lang packs with guids from prod alright with that ADU 0?

[...]

PR #14994, opened 2020-07-20
Enable stats for dictionaries and langpacks

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14982


src/olympia/addons/cron.py
```diff
@@ -28,34 +27,36 @@
...
 
-    ts = [
-        _update_addon_average_daily_users.subtask(args=[chunk])
-        for chunk in chunked(counts, 250)
-    ]
-    group(ts).apply_async()
+    log.info('Preparing update of `average_daily_users` for %s add-ons.',
+             len(counts))
+
+    create_chunked_tasks_signatures(
```
willdurand, line 57: This is an improvement to have some more logs.

src/olympia/addons/cron.py
```diff
@@ -28,34 +27,36 @@
...
         .values_list('guid', 'count')
         # Just to make order predictable in tests, we order by id. This
         # matches the GROUP BY being generated so it should be safe.
         .order_by('id')
     )
-    counts = dict(get_addons_and_average_daily_users_from_bigquery())
-    counts.update(amo_counts)
+    # Update the `counts` with values from BigQuery.
+    counts.update(dict(get_addons_and_average_daily_users_from_bigquery()))
     counts = list(counts.items())
```
willdurand, line 52: If this new strategy works, I think we could apply it to the `hotness` cron task. We wouldn't have 2 different tasks anymore, we wouldn't have a giant `IN` clause and we the number of add-ons to process/tasks would be roughly the same.

PR #14998, opened 2020-07-20
RE: Add validations to Shelf model

### LINKED ISSUE #14952: Add validations to Shelf model
Supports #14871
Dependent on #14872 

Add validators to Shelf model that checks that `criteria` does not return 404s upon saving the instance.

Another issue will cover creating a validator for `footerPathname`.

xlisachan:
Fixes #14952 

I wanted to see if I was headed in the right direction in terms of creating a validator (and its tests) for the criteria before I add a validator for the footer pathname.

With this draft, the validator catches if the criteria is invalid or whether the parameters are incorrect (returning the error from `results`):
- E.g., Invalid "recommended" parameter
Screen Shot 2020-07-20 at 11 11 01 AM

- E.g., Invalid "sort" parameter
Screen Shot 2020-07-20 at 11 11 23 AM

- E.g., Invalid "type" parameter
Screen Shot 2020-07-20 at 11 11 50 AM

- E.g., In this case, the `type` parameter was `theme` instead of `statictheme` and it returned a status code of 200
Screen Shot 2020-07-20 at 11 12 03 AM

- E.g. Invalid criteria
Screen Shot 2020-07-20 at 11 58 23 AM

I revised `mock_basket()` within `conftest.py` as it was throwing errors for `test_can_edit_with_discovery_edit_permission` and `test_can_edit_with_discovery_edit_permission` in `test_admin.py` (see below). In addition, I added tests to check whether the criteria passed validations, or failed and returned either a 404, 400, or empty 'results' list. All tests currently pass :)
Screen Shot 2020-07-20 at 11 31 04 AM


conftest.py
```diff
@@ -79,6 +79,31 @@ def mock_basket(settings):
         responses.POST,
         settings.BASKET_URL + '/news/unsubscribe/{}/'.format(USER_TOKEN),
         json={'status': 'ok', 'token': USER_TOKEN})
+    responses.add(
```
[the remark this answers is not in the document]
xlisachan: I wasn't sure if I should've went this route either, the errors in the tests stated the response did not match any registered mocks that led me here. I agree that a new fixture should be created if this is the way to go though.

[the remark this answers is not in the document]
eviljeff: Yeah, don't add them here - this is for urls that are pinged site-wide (if we're randomly accessing the API elsewhere on the site it's a bug and we certainly want those tests to error).  Add the to the TestXxx setup for tests that deal with shelves.

src/olympia/discovery/tests/test_admin.py
```diff
@@ -1076,7 +1079,7 @@ def test_can_edit_with_discovery_edit_permission(self):
         item = Shelf.objects.create(
             title='Recommended extensions',
             shelf_type='extension',
-            criteria='/this/is/the/criteria',
```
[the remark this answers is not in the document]
xlisachan: I feel like there should be too. I made these changes because with the `validate_criteria` validator, `test_can_edit_with_discovery_edit_permission` and `test_can_add_with_discovery_edit_permission` do not pass with just a placeholder in `criteria`. It led to the error message above about the call not matching a registered mock.

src/olympia/discovery/tests/test_admin.py
```diff
@@ -1226,3 +1233,36 @@ def test_can_not_delete_without_discovery_edit_permission(self):
...
+
+    def test_criteria_does_pass_validation(self):
+        criteria = 'search/?recommended=true&sort=random&type=extension'
+        try:
+            validate_criteria(criteria)
+        except ValidationError as e:
+            self.assertRaises(
+                '404 Not Found - Invalid criteria' in e.message_dict)
+
+    def test_criteria_does_not_pass_validation_returns_404(self):
```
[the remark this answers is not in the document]
xlisachan: The values for `criteria` are different to test that it returns a 404.
- `test_criteria_does_pass_validation`:
`criteria = 'search/?recommended=true&sort=random&type=extension'`
- `test_criteria_does_not_pass_validation_returns_404`
`criteria = 'recommended=true&sort=users&type=extension'`

src/olympia/shelves/validators.py
```diff
@@ -0,0 +1,23 @@
+import requests
+
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext_lazy as _
+
+
+def validate_criteria(value):
+    url = "https://addons.mozilla.org/api/v4/addons/{}"
```
eviljeff, line 8: as discussed in the meeting, we don't want this as a constant inside the method.  If we're running against the current server we can do `drf_reverse('v4:addon-search')` (`from rest_framework.reverse import reverse as drf_reverse`) to get the addon search api endpoint.  Note this would mean a change in the criteria because you wouldn't need the "search/" part of it.

Not strictly necessary here, but to get an absolute url, append `settings.EXTERNAL_SITE_URL` to the start of the url.

conftest.py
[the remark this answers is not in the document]
eviljeff: If we're accessing the api in the test instance we don't need any of these responses adding now - we can just add the addons (using `addon_factory()`) in the test setup instead.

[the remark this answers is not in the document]
eviljeff: Though for the rest of the tests that don't _specifically_ test validation it would be easier to just mock the whole validation, like @bobsilverberg suggested.  You can use `mock.patch` on the clean method (see other comments where I talk about using clean over the validator method)

xlisachan:
Thanks for all the feedback! The following updates have been made:
- Created `mock.patch` for `test_can_edit_with_discovery_edit_permission` and `test_can_edit_with_discovery_edit_permission` in `test_admin.py`
- Replaced validator with `clean_criteria` method in a form
- Removed responses previously added to `conftest.py`

I am currently working on the following:
- Utilizing `settings.EXTERNAL_SITE_URL` as I receive the error `HTTPConnectionPool: Max retries with url`
- Adding tests for `clean_criteria` method

src/olympia/shelves/forms.py
```diff
@@ -0,0 +1,51 @@
...
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext_lazy as _
+
+from olympia.shelves.models import Shelf
+
+
+class ShelfForm(forms.ModelForm):
+    class Meta:
+        model = Shelf
+        fields = '__all__'
```
eviljeff, line 15: define fields in `ShelfAdmin` instead and explicitly list them rather than `__all__` - we probably don't want to see created or modified.

src/olympia/shelves/forms.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def clean_criteria(self):
+        data = self.cleaned_data
+        criteria = data['criteria']
+        if (data['shelf_type'] == 'extension' or
+                data['shelf_type'] == 'theme' or
+                data['shelf_type'] == 'search'):
```
eviljeff, line 22: `data['shelf_type'] in ('extension', 'theme', 'search')`

src/olympia/shelves/forms.py
```diff
@@ -0,0 +1,51 @@
...
+                "api/v4/addons/recommendations/" + criteria
+
+        response = requests.get(url)
+        results = response.json()
+
+        if response.status_code == 404:
+            raise ValidationError(
+                _("Check criteria field: 404 Not Found"),
+                params={'criteria': criteria},)
+        if response.status_code == 400:
```
eviljeff, line 42: this is probably better as `if response.status_code != 200:` because other non-200 responses would be errors too.

src/olympia/shelves/forms.py
```diff
@@ -0,0 +1,51 @@
...
+            raise ValidationError(
+                _("Check criteria field: %s" % results[0]),
+                params={'criteria': criteria},)
+        if response.status_code == 200 and 'results' in results:
+            if len(results['results']) == 0:
```
eviljeff, line 47: now the status code is dealt with if in the previous `if` this can become a single `if results.get('results', []) == 0:`

src/olympia/shelves/forms.py
```diff
@@ -0,0 +1,51 @@
...
+    def clean_criteria(self):
+        data = self.cleaned_data
+        criteria = data['criteria']
+        if (data['shelf_type'] == 'extension' or
+                data['shelf_type'] == 'theme' or
+                data['shelf_type'] == 'search'):
+            url = "https://addons.mozilla.org" + \
+                drf_reverse('v4:addon-search') + criteria
+        elif data['shelf_type'] == 'categories':
+            url = "https://addons.mozilla.org/api/v4/addons/categories/" + \
```
eviljeff, line 26: do the same `drf_reverse` lookup` with these urls too.

eviljeff: If you ever can't find the view name for a url then `./manage.py show_urls` lists them - the name is in red on the right hand side.

src/olympia/shelves/forms.py
```diff
@@ -0,0 +1,51 @@
...
+                criteria
+        elif data['shelf_type'] == 'collections':
+            url = "https://addons.mozilla.org/api/v4/accounts/account/" + \
```
eviljeff, line 29: and the other urls in this method

src/olympia/shelves/forms.py
```diff
@@ -0,0 +1,51 @@
...
+                criteria
+        elif data['shelf_type'] == 'recommended':
+            url = "https://addons.mozilla.org/" + \
+                "api/v4/addons/recommendations/" + criteria
```
eviljeff, line 33: it's a project style thing, but we don't use `\` for continuing lines in addons-server python code, we always wrap in `(` and `)` instead.

src/olympia/discovery/tests/test_admin.py
```diff
@@ -1089,15 +1093,16 @@ def test_can_edit_with_discovery_edit_permission(self):
         content = response.content.decode('utf-8')
         assert 'Recommended extensions' in content
 
+        clean_criteria_mock.return_value = {
+            'title': 'Popular extensions',
+            'shelf_type': 'extension',
+            'criteria': '?recommended=true&sort=users&type=extension'
+        }
```
eviljeff, line 1100: I don't think you need this `return_value`?  The `clean_criteria` method wouldn't usually return anything - it just raises exceptions for problems and returns nothing otherwise.

This is the kind of return you'd expect from the responses.add though.

src/olympia/shelves/forms.py
xlisachan: *`if len(results.get('results')) == 0` 👍🏼

eviljeff: `if len(results.get('results', [])) == 0` - `len(None)` is a TypeError.

xlisachan: * `if results.get('results', []) == []` ?

src/olympia/discovery/tests/test_admin.py
xlisachan: The `clean_criteria` method returns the criteria. When I removed `return criteria`, the `criteria` field errors to `This field cannot be null` when there is a valid criteria in the field. I had it as just a `clean` method before which did not return anything, but I anticipate adding a `clean_footer_pathname` later and did not one huge `clean` method. Thoughts?

src/olympia/shelves/forms.py
xlisachan: I totally forgot about 'count'! 🤦🏻‍♀️ - `if results.get('count', 0) == 0`?

xlisachan:
Please see updates below regarding my latest push:
- Made a minor update to `SHELF_TYPES` in `models.py`
- Refactored `clean_criteria` in `forms.py` - as mentioned above, the value of the results variable could either be a dict or list depending on the endpoint so I added `if 'count' in results:`
- Updated `mock.patch` in `test_admin.py`, as `clean_criteria` only returns the criteria and not the rest of the fields
- Added tests/responses to check `clean_critera` in 'test_admin.py`
- Added messages in `admin.py` to display the total number of addons returned for the shelf after successfully creating/editing a shelf

I was not able to utilize `settings.EXTERNAL_SITE_URL`, as it'd error with `HTTPConnectionPool: Max retries with url`.

src/olympia/discovery/tests/test_admin.py
```diff
@@ -1,17 +1,23 @@
 # -*- coding: utf-8 -*-
 from pyquery import PyQuery as pq
+from unittest import mock
 
 import os
+import responses
+
+from rest_framework.reverse import reverse as drf_reverse
```
eviljeff, line 8: you can use `reverse_ns` in tests (and trim the `v4:` from the start of the view names)

src/olympia/discovery/tests/test_admin.py
```diff
@@ -1226,3 +1277,58 @@ def test_can_not_delete_without_discovery_edit_permission(self):
         assert response.status_code == 403
         assert Shelf.objects.filter(pk=item.pk).exists()
         assert item.title == 'Recommended extensions'
+
+    def test_clean_criteria_search(self):
```
eviljeff, line 1281: because we're explicitly testing the form here, rather than the admin view, these tests should be in `test_forms.py` instead.

src/olympia/discovery/tests/test_admin.py
```diff
@@ -1226,3 +1277,58 @@ def test_can_not_delete_without_discovery_edit_permission(self):
...
+
+    def test_clean_criteria_returns_404(self):
+        with self.assertRaises(Exception) as context:
+            form = ShelfForm({
+                'title': "Popular themes",
+                'shelf_type': 'theme',
+                'criteria': 'sort=users&type=statictheme'})
+            assert form.is_valid()
+            self.assertTrue(
+                'No data found - check criteria' in str(context.exception))
```
eviljeff, line 1314: `assert 'No data found - check criteria' in str(context.exception)` though you should be able to do `assert 'No data found - check criteria' in context.exception.message`, right?

src/olympia/discovery/tests/test_admin.py
```diff
@@ -1226,3 +1277,58 @@ def test_can_not_delete_without_discovery_edit_permission(self):
...
+    def test_clean_criteria_recommendations(self):
+        form = ShelfForm({
+            'title': "Recommended Add-ons",
+            'shelf_type': 'recommendations',
+            'criteria': '?recommended=true'})
+        assert form.is_valid(), form.errors
+        assert form.cleaned_data['criteria'] == '?recommended=true'
+
+    def test_clean_criteria_returns_404(self):
+        with self.assertRaises(Exception) as context:
```
eviljeff, line 1307: catch django's `ValidationError` rather than the generic `Exception`

eviljeff: same for the following tests

src/olympia/discovery/tests/test_admin.py
eviljeff: same for the following tests

src/olympia/shelves/admin.py
```diff
@@ -1,6 +1,45 @@
...
 
 
 class ShelfAdmin(admin.ModelAdmin):
     list_display = ('title', 'shelf_type')
     actions = ['delete_selected']
+    form = ShelfForm
+
+    def message_user(self, request, message, level=messages.INFO,
+                     extra_tags='', fail_silently=False):
+        pass
```
eviljeff, line 17: what's this method for?  (it doesn't do anything)

src/olympia/shelves/forms.py
```diff
@@ -0,0 +1,58 @@
...
+class ShelfForm(forms.ModelForm):
+    class Meta:
+        model = Shelf
+        fields = ('title', 'shelf_type', 'criteria',
+                  'footer_text', 'footer_pathname',)
+
+    def clean_criteria(self):
+        data = self.cleaned_data
+        criteria = data['criteria']
+        baseUrl = 'https://addons.mozilla.org'
```
eviljeff, line 21: This certainly shouldn't be a hardcoded url to the production amo site.  If using `settings.EXTERNAL_URL` is causing an error perhaps it's set to an unexpected value?  (either use python debugger or throw in some temporary `print()` statements to see what's going on)

src/olympia/shelves/admin.py
```diff
@@ -1,6 +1,45 @@
...
+        elif obj.shelf_type == 'categories':
+            api = drf_reverse('v4:category-list')
+        elif obj.shelf_type == 'collections':
+            api = drf_reverse('v4:collection-list')
+        elif obj.shelf_type == 'recommendations':
+            api = drf_reverse('v4:addon-recommendations')
+
+        url = baseUrl + api + obj.criteria.lower()
+        response = requests.get(url)
+        results = response.json()
```
eviljeff, line 33: this is all duplicated with the code in forms, and would mean two identical requests, so we don't want that.  I'm guessing this is just because you need the response to include the count in the messages.  

If so, lets just hack around that - store the response on the form instance `self.response = response.json()`, then add it to the Shelf instance (in the form's save method), then it'll be available as `obj.response` in `save_model`.

src/olympia/discovery/tests/test_admin.py
eviljeff: I guess you can move the responses.add in the test setup there too.

src/olympia/shelves/models.py
```diff
@@ -2,8 +2,8 @@
 
 from olympia.amo.models import ModelBase
 
-SHELF_TYPES = (
-    'category', 'collection', 'extension', 'recommended', 'search', 'theme')
+SHELF_TYPES = ('categories', 'collections', 'extension',
+               'recommendations', 'search', 'theme')
```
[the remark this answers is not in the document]
xlisachan: I wanted to revise the types to match the name of the API being used to avoid any confusion when creating shelves. I would've left it alone if it was already in use, but since it's not, I figured now's the time to make the update.

src/olympia/shelves/admin.py
xlisachan: It overrides the automated success message after a shelf is created/updated.

Without it :
Screen Shot 2020-07-27 at 4 33 35 PM

With it:
Screen Shot 2020-07-27 at 4 34 11 PM

eviljeff: okay, can you add a code comment in the method saying that.

xlisachan:
This latest push includes the following revisions:
- Created `test_forms.py`
- Utilized `reverse-ns` and trimmed `v4:`
- Updated `self.assertTrue` to `assert...`
- Revised `Exception` to use `ValidationError`
- Included a note that disables the automated success message
- Changed the `baseUrl` to `settings.EXTERNAL_SITE_URL`

I'm having trouble with saving the results in the form's save method. It errors to `ShelfForm object does not have attribute 'results'`. The `test_can_add_with_discovery_edit_permission tests` and `test_can_edit_with_discovery_edit_permission` in `discovery/tests/test_admin.py` fails with the same error, but should that be mock patched since it's form related? Hints would be appreciated :)

src/olympia/shelves/forms.py
```diff
@@ -0,0 +1,56 @@
...
+                        'Check criteria parameters - e.g., "type"')
+            elif len(self.results) == 0:
+                raise forms.ValidationError('Check criteria - No data found')
+        except requests.exceptions.ConnectionError as err:
+            print('Error connecting:', err)
+
+    def save(self, commit=True):
+        instance = super().save(commit=False)
+        instance.results = self.results
+        instance.save()
```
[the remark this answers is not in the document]
xlisachan: > this is all duplicated with the code in forms, and would mean two identical requests, so we don't want that. I'm guessing this is just because you need the response to include the count in the messages.
> 
> If so, lets just hack around that - store the response on the form instance `self.response = response.json()`, then add it to the Shelf instance (in the form's save method), then it'll be available as `obj.response` in `save_model`.

I'm not following how `ShelfAdmin.save_model` gets the add-ons count from the form's `clean` method if the `save_model`'s duplicate call to the API is eliminated...

src/olympia/shelves/forms.py
```diff
@@ -0,0 +1,56 @@
...
+
+            # Value of results is either dict or list depending on the endpoint
+            if 'count' in self.results:
+                if self.results.get('count', 0) == 0:
+                    raise forms.ValidationError(
+                        'Check criteria parameters - e.g., "type"')
+            elif len(self.results) == 0:
+                raise forms.ValidationError('Check criteria - No data found')
+        except requests.exceptions.ConnectionError as err:
+            print('Error connecting:', err)
```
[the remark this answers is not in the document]
xlisachan: I suppose I'll log it. It won't save if there's a connection error and it just keeps being raised if set as a validation error.

src/olympia/shelves/forms.py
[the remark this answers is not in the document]
eviljeff: this `save` method is supposed to return `instance` (I missed that).

So, django admin does it (roughly) in this order:
- form.valid(), which calls form.clean()
- Admin.save_form(), which calls form.save() - that should return the new/update instance, but _doesn't_ save it to the database (if commit=False, which is what it passes in Admin.save_form)
- Admin.save_model(), which calls obj.save() and actually saves the data to the database.

https://github.com/django/django/blob/2.2.14/django/contrib/admin/options.py#L1745:L1773  (it's a bit overly complicated, but if you're interested, take a look)

xlisachan:
The latest push includes updating the `settings` files in order to call the API for shelf validations. As discussed in the meeting, `results` have been removed for this push and will be added at a later time.

settings.py
```diff
@@ -76,12 +76,13 @@ def insert_debug_toolbar_middleware(middlewares):
...
 DOMAIN = SERVICES_DOMAIN = urlparse(SITE_URL).netloc
 SERVICES_URL = SITE_URL
+INTERNAL_SITE_URL = 'http://nginx'
 EXTERNAL_SITE_URL = SITE_URL
 
 CODE_MANAGER_URL = (
     os.environ.get('CODE_MANAGER_URL') or 'http://localhost:3000')
 
-ALLOWED_HOSTS = ALLOWED_HOSTS + [SERVICES_DOMAIN]
+ALLOWED_HOSTS = ALLOWED_HOSTS + [SERVICES_DOMAIN] + [EXTERNAL_SITE_URL]
```
eviljeff, line 85:
```suggestion
ALLOWED_HOSTS = ALLOWED_HOSTS + [SERVICES_DOMAIN, EXTERNAL_SITE_URL]
```

src/olympia/lib/settings_base.py
```diff
@@ -31,6 +31,7 @@
     '.mozilla.com',
     '.mozilla.net',
     '.mozaws.net',
+    'nginx'
```
eviljeff, line 34: this shouldn't be nessecary

src/olympia/conf/prod/settings.py
```diff
@@ -103,3 +103,6 @@
 
 # See: https://bugzilla.mozilla.org/show_bug.cgi?id=1633746
 BIGQUERY_AMO_DATASET = 'amo_prod'
+
+INTERNAL_SITE_URL = env(
+    'INTERNAL_SITE_URL', default='https://addons.mozilla.org')
```
eviljeff, line 108: can we move this further up the file so it's next/near to `EXTERNAL_SITE_URL`.

eviljeff: same with dev and stage files

src/olympia/lib/settings_base.py
xlisachan: Without it, it errors out to a 400 response - `DisallowedHost at /api/v4/addons/search/, Invalid HTTP_HOST header: 'nginx'. You may need to add 'nginx' to ALLOWED_HOSTS.`

eviljeff: Ah, `ALLOWED_HOSTS = ALLOWED_HOSTS + [SERVICES_DOMAIN, EXTERNAL_SITE_URL]` is incorrect then.  It's because we're adding the full _url_ rather than just the domain. Add `'nginx'` to the list rather than `EXTERNAL_SITE_URL`

PR #15005, opened 2020-07-21
Clear pending rejection flag on version deletion

### LINKED ISSUE #15002: The Pending rejection status is not removed from deleted versions
Follow up for #14454 

1. With a reviewer account, submit a delayed review rejection for an add-on version
2. As the add-on developer, delete the version that is under Pending rejection
3. Verify the version status in reviewer tools

### What happened?
The deleted version still has an active Pending rejection - see [example](https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/webmail-ad-blocker2), second version

Since the version is deleted, the Pending rejection status should be removed as well

image

diox:
Fixes #15002

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -48,7 +48,8 @@ from olympia.translations.hold import translation_saved
 from olympia.translations.models import Translation
 from olympia.users.models import UserProfile
 from olympia.versions.compare import version_int
-from olympia.versions.models import Version, VersionPreview, inherit_nomination
+from olympia.versions.models import (
+    Version, VersionPreview, VersionReviewerFlags, inherit_nomination)
 
 from . import signals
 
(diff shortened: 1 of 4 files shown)


src/olympia/addons/models.py
```diff
@@ -638,6 +639,8 @@ def delete(self, msg='', reason='', send_delete_email=True):
             file_tasks.hide_disabled_files.delay(addon_id=self.id)
 
             self.versions.all().update(deleted=True)
+            VersionReviewerFlags.objects.filter(version__addon=self).update(
+                pending_rejection=None)
```
eviljeff, line 643: Do we need something similar for addon disable, or is that handled a different way?  (I guess the versions get disabled at that point and... somehow the flags get removed...?)

diox: The cron will eventually reject them even if they are disabled, so we don't need to do anything. For deletions we explicitly clear it mostly for convenience, because it's not something that is supposed to be reverted so we know we can get rid of the flag right away.

diox: (Also, yes, rejecting versions would clear the flag)

Issue #14981, opened 2020-07-17
Not all "hotness" tasks seem to be submitted in the `deliver_hotness` CRON task


willdurand:
It does not seem to affect -dev/-stage but it was very obvious in -prod after yesterday's push. We see logs for `reset` sub-tasks but nothing for `update` sub-tasks. Theoretically, there should be 1K `update` tasks and very few `reset` ones.

I don't see any errors in Sentry and the BQ call seems fine because the `reset` sub-tasks looked good (and we need the BQ data to prepare and submit those tasks).

willdurand:
We found the root cause today and we were able to unlock the CRON task.

willdurand:
> We found the root cause today and we were able to unlock the CRON task.

We found that one `FrozenAddon` had a GUID set to `None`, because this add-on was deleted and we apparently don't keep the list of frozen add-ons in sync (see: https://github.com/mozilla/addons-server/issues/15004).

We also found an error related to a BigQuery `addon_id` (GUID) with an invalid case like `something@moz.Org` instead of `something@moz.org`. MySQL is happy to return results but then the dict lookup fails (`averages`).

ioanarusiczki:
@willdurand In case there's something qa can verify please let me know. 
Thanks!

PR #15007, opened 2020-07-21
Refactor `deliver_hotness()` task

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/14981

### DIFF
--- a/src/olympia/addons/cron.py
+++ b/src/olympia/addons/cron.py
@@ -13,7 +13,7 @@ from olympia.addons.models import Addon, FrozenAddon
 from olympia.addons.tasks import (
     update_addon_average_daily_users as _update_addon_average_daily_users,
     update_addon_download_totals as _update_addon_download_totals,
-    update_appsupport, update_addon_hotness, reset_addon_hotness)
+    update_appsupport, update_addon_hotness)
 from olympia.amo.celery import create_chunked_tasks_signatures
 from olympia.amo.decorators import use_primary_db
 from olympia.amo.utils import chunked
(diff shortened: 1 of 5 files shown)


src/olympia/addons/cron.py
```diff
@@ -180,27 +180,36 @@ def deliver_hotness(chunk_size=300):
...
     log.info('Found %s frozen add-on GUIDs.', len(frozen_guids))
 
-    averages = get_averages_by_addon_from_bigquery(
-        today=date.today(), exclude=frozen_guids
+    amo_guids = (
+        Addon.objects.filter(status__in=amo.REVIEWED_STATUSES)
+        .exclude(guid__in=frozen_guids)
+        .exclude(guid__isnull=True)
+        .exclude(guid__exact='')
+        .exclude(hotness=0)
```
wagnerand, line 193: The more I think about this, the more I think we should exclude _inactive_ add-ons. They are not visible on any public page. We also include other reasons for an add-on being - temporarily or permanently - being hidden from public pages, by filtering on `REVIEWED_STATUSES`, so why wouldn't we want to exclude _inactive_ ones as well?

diox: I think the fewer add-ons we exclude, the better. Excluding based on things the developers can trivially change themselves like user-disabled is risky, because then they might keep an old `hotness` value and suddenly show up as trending later when they are re-enabled if the value they had before being disabled was high.

There is no drawback in resetting hotness for add-ons that are not public, except for the fact that it makes us fetch more in the cron.

wagnerand: The same would apply to add-ons that have status _Incomplete_ because there is currently no public version (might have been rejected by a reviewer or disabled by the developer). At any time, the developer could upload a new listed version and in the vast majority of cases, it would be auto-approved and the add-on would become visible again.

diox: Yes, and I think we should also not exclude those.

diox: @willdurand with that new logic, I don't see a reason to filter by `status` anymore. We do want to continue excluding `hotness=0` to avoid unnecessary updates, excluding empty `guids` because that won't lead to anything, and excluding frozen guids because they should never be updated, but I don't see why we'd continue filtering by status. In any case if there is some data in bigquery those add-ons will get their hotness updated.

willdurand: afaik that was there before? (and it'd be consistent with `average_daily_users`?

willdurand: I removed the `.filter()`

PR #15025, opened 2020-07-22
Don't default to delayed rejections in unlisted reviews

### LINKED ISSUE #15023: Reject multiple versions for unlisted add-on reviews applies a delayed review rejection
Follow up for #14454 

1. Submit an unlisted add-on
2. Open its unlisted review page with an Admin reviewer _before_ the add-on gets auto-approved
3. Use the Reject multiple versions option to reject the version
4. Observe the results

### What happened?
The version ends up with a delayed review rejection

Reject multiple versions for unlisted should default to immediate rejection (this was specifically requested in #14024)

- the Reject multiple versions UI for unlisted has not been updated to the new model, since delayed review rejections have not been implemented for unlisted versions (yet)
- example: https://reviewers.addons.allizom.org/en-US/reviewers/review-unlisted/c42623a07d7c445c9416

image

diox:
The choice isn't even present for unlisted reviews, so we should default to immediate rejections in that case.

Fixes #15023

### DIFF
--- a/src/olympia/reviewers/forms.py
+++ b/src/olympia/reviewers/forms.py
@@ -382,6 +382,17 @@ class ReviewForm(forms.Form):
         self.helper = kw.pop('helper')
         super(ReviewForm, self).__init__(*args, **kw)
 
+        if self.helper.version:
+            channel = self.helper.version.channel
+        else:
+            channel = amo.RELEASE_CHANNEL_LISTED
+
+        # We default to delayed rejection for listed, but for unlisted the
+        # input will never be shown for any action, so it should default to
+        # False.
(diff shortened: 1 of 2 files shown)


diox:
Actually, we'll want to default even listed to be immediate for the time being, I'll update this.

wesinator:
> Actually, we'll want to default even listed to be immediate for the time being, I'll ~update~ close this and make a separate PR.

@diox does it make sense to change this now, so that listed addons will default to a delayed rejection? 
this is the most common path that reviewers take it seems, and therefore seems like the saner default.

diox:
That's more of a @wagnerand question. Feel free to open a new issue to discuss this.

Issue #15030, opened 2020-07-22
Use BigQuery for download stats (behind a Flag)


willdurand:
Once we have a way to pull data, let's display it. We'll use a waffle flag to determine whether to show new data (from BigQuery) or legacy data (download count). This issue is about using BigQuery for the existing download stats only.

willdurand:
I think in -dev/-stage, we'd likely have no data when flag is disabled and some data from BQ when flag is active.

ioanarusiczki:
@willdurand On dev:
- the admin has access to data as before without being added as a user for `bigquery-download-stats`.
- I added then removed a regular user that has an uploaded [dictionary](https://addons-dev.allizom.org/en-US/developers/addon/release-dictionary-with-stats/edit) with BigQuery data but I don't see changes, the same data is available in both cases. 
Perhaps I'm doing something wrong but I think the waffle is enabled so I did not change anything else except for the user.
- a single user can be added or removed from `bigquery-download-stats`. Can't tell what's expected.

willdurand:
> a single user can be added or removed from `bigquery-download-stats`. Can't tell what's expected.

mm, no you can have more than one I think. I just added myself.


> the admin has access to data as before without being added as a user for `bigquery-download-stats`.

Yeah, the pages remain the same. Only the data is different and that's depending on the waffle flag.

Here is an add-on with download data: https://addons-dev.allizom.org/en-US/firefox/addon/facebook-container/statistics/downloads/?last=30. Last few days should be 0 because we didn't automate the import of new data in the BigQuery table.

> I added then removed a regular user that has an uploaded dictionary with BigQuery data but I don't see changes, the same data is available in both cases.

Really? Did you try to hard refresh?

ioanarusiczki:
@willdurand 
 Something is not working on my default profile but I tested on a new profile and I did not have problems, no need to hard refresh. 
No problems with adding more users either, for other admin models each user has a new field where it can be added but in this case multiple users can be added using the same field, separated by `,`  (so I got confused)

1. With flag disabled and an admin account - I see the data displayed for all the add-ons shown in the [search results](https://addons-dev.allizom.org/en-US/firefox/search/?platform=windows&sort=users). Indeed, the downloads are `0` since 21th of July. 
2. With flag enabled and the admin added into the `users` group the data from 1. + the new download categories from https://github.com/mozilla/addons-server/issues/15012 are displayed.
[...]

PR #15031, opened 2020-07-22
Use BigQuery for download stats (behind a Flag)

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15030

---

~~:warning: depends on https://github.com/mozilla/addons-server/pull/15029~~

### DIFF
--- a/src/olympia/stats/tests/test_views.py
+++ b/src/olympia/stats/tests/test_views.py
@@ -8,6 +8,7 @@ from unittest import mock
 from django.http import Http404
 from django.test.client import RequestFactory
 from django.utils.encoding import force_text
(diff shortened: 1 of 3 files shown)


src/olympia/stats/tests/test_views.py
```diff
@@ -279,36 +289,25 @@ def csv_eq(self, response, expected):
 
 
 class TestViews(ESStatsTestCase):
-    def _test_no_cache(self):
```
willdurand, line 282: I removed this test case. It wasn't running anyway and we have another test class for cache.

willdurand:
rebased!

Issue #15012, opened 2020-07-21
Add new "By" download stats pages 


willdurand:
According to the [PRD](https://docs.google.com/document/d/1TuMqhEWNlCNDQRgqFywpJvDHTTNjnAq7-3BAXtoFLQ4/edit#), we want the following pages for the new download stats:

- By Medium
- By Content
- By Campaign

We'll keep the "By Source" page.

ioanarusiczki:
@willdurand 

I tested with an admin and a regular account on dev - the waffle active. 

When the users mentioned above are added to `bigquery-download-stats` the [new](https://github.com/mozilla/addons-server/issues/15012#issue-663276455) classification is displayed, including `by Source` and the same data from the general "Downloads" is everywhere, with "unknown" on the right side of the detailed daily stats.

Issue #15013, opened 2020-07-21
Update the documentation in the “About tracking external sources” popup


willdurand:
According to the [PRD](https://docs.google.com/document/d/1TuMqhEWNlCNDQRgqFywpJvDHTTNjnAq7-3BAXtoFLQ4/edit#), the link should appear in all download stats pages.

willdurand:
We'll have this popup in all "By XYZ" download stats pages (see #15012). We would need a new copy for each page I suppose or have one copy that would work for each page (i.e. each tracking dimension).

willdurand:
> We'll have this popup in all "By XYZ" download stats pages (see #15012). We would need a new copy for each page I suppose or have one copy that would work for each page (i.e. each tracking dimension).

cc @jvillalobos

jvillalobos:
@caitmuenster What do you think? We could either have something in EW explaining how to use UTM params or link to a generic doc about it.

caitmuenster:
@jvillalobos I like the idea of having a page on EW explaining how to use the UTM params! 

Would we be getting rid of the popups on those pages and directly linking to the EW article?

jvillalobos:
> Would we be getting rid of the popups on those pages and directly linking to the EW article?

Yes, the link would open the EW page in a new tab.

caitmuenster:
Ok, cool. I'll get a draft of the page for your review soon.

willdurand:
> Ok, cool. I'll get a draft of the page for your review soon.

Awesome, thanks. Let me know when we have something to review/link to. Also, we might need to display a banner saying that we're not yet collecting the attribution data using UTM parameters. That's the plan but not something we are currently doing (and we'll likely enable the new download stats without a lot of attribution data at the very beginning).

AlexandraMoga:
All the Downloads by... options will include the "About tracking external sources" link which points to https://extensionworkshop-dev.allizom.org/documentation/manage/monitoring-extension-usage-statistics/

![2020-08-07_18-09-07](https://user-images.githubusercontent.com/31961530/89659986-58f25200-d8d9-11ea-9f7b-a569f354b66e.gif)

caitmuenster:
Not sure if this needs to be called out but `/documentation/manage/monitoring-extension-usage-statistics/` won't be live on EW prod until we release the new version, which will hopefully be this week. The prod URL will be `https://extensionworkshop.com/documentation/manage/monitoring-extension-usage-statistics/`.

PR #15042, opened 2020-07-23
Add new "By" download stats pages

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15012

~~:warning: depends on #15031~~

---

This patch adds new "By" pages behind the same flag introduced in #15031. The data for each new page still needs to be collected in Firefox, but BigQuery does return data for each anyway (all with an `unknown` key).

There is some more work to document how data is collected, see https://github.com/mozilla/addons-server/issues/15013

### DIFF
--- a/src/olympia/stats/templatetags/jinja_helpers.py
+++ b/src/olympia/stats/templatetags/jinja_helpers.py
@@ -1,11 +1,10 @@
 from django.template import loader
 
 import jinja2
+import waffle
 
 from django_jinja import library
 
-from olympia import amo
-from olympia.access import acl
 from olympia.addons.models import Addon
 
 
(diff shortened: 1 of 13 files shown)


src/olympia/stats/templatetags/jinja_helpers.py
```diff
@@ -14,15 +13,11 @@
...
-        has_privs = False
-        if request.user.is_authenticated and (
-            acl.action_allowed(request, amo.permissions.STATS_VIEW) or
-            obj.has_author(request.user)
-        ):
-            has_privs = True
         tpl = loader.get_template('stats/addon_report_menu.html')
         ctx = {
             'addon': obj,
-            'has_privs': has_privs,
```
bobsilverberg, line 26: We no longer need to check for privileges? Or has that logic been moved elsewhere?

src/olympia/stats/tests/test_views.py
```diff
@@ -918,6 +920,234 @@ def test_downloads_sources_csv(self):
             2009-06-01,1000,550,450""",
         )
 
+    def test_download_by_content_json(self):
+        self.get_download_series_mock.return_value = [
```
bobsilverberg, line 924: Most of this setup is repeated in each test. I.e., the `date`, `end` and `count` are identical in each test, with only the `data` changing. Do you see any value in creating a helper function that returns this, and thereby being able to remove a lot of code from each test. Or do you think it's better to have everything explicit in each test?

src/olympia/stats/templatetags/jinja_helpers.py
willdurand: This wasn't used anymore.

src/olympia/stats/tests/test_views.py
willdurand: I tried to refactor a bit but I could get a parametrized test working correctly. This is repetitive for sure but it is also very explicit.. Maybe some sort of helper would help (hehe) but I also like that we have everything in each test case.

PR #15043, opened 2020-07-23
drop Block.include_in_legacy and derive legacy state from legacy_id

### LINKED ISSUE #14262: refactor Block to derive `include_in_legacy` from `legacy_id`
follow-up from #14222 - really, we don't want to be holding `include_in_legacy` property on `Block`, when instead we could be inspecting `kinto_id`.  If it's set it means there's a matching record in the legacy blocklist so it's included (`include_in_legacy is True`); if it's not set there isn't a matching record in the legacy blocklist (`include_in_legacy is False`).

It's easy enough to drop the model field and replace it with a `@property`, but when the `blocklist_legacy_submit` waffle switch is enabled we also want to be able to change it.  This means the `BlocklistSubmission` object still needs to store the property value.  Added complication is we cheat with the change page for a `Block` and submit the form directly into `BlocklistSubmission.add_view`, which relies on the Block change form having the same fields.

eviljeff:
fixes #14262


src/olympia/blocklist/models.py
```diff
@@ -261,12 +264,18 @@ def __str__(self):
         trimmed = [rep if len(rep) < 40 else rep[0:37] + '...' for rep in repr]
         return f'{self.get_signoff_state_display()}: {"; ".join(trimmed)}'
 
+    @property
+    def in_legacy_blocklist(self):
+        # This is for consitency with Block.  Should be a boolean anyway.
```
willdurand, line 269: Why is there two spaces between the two sentences? Also, shouldn't this comment be added to the same method above (in the other class)?

src/olympia/blocklist/utils.py
```diff
@@ -104,52 +107,41 @@ def legacy_publish_blocks(blocks):
...
+            log.info(
+                f'Block [{block.guid}] was imported from a regex guid so '
+                'can\'t be safely updated.  Skipping.')
+            continue
+        data = {
+            'guid': block.guid,
+            'details': {
+                'bug': block.url,
+                'why': block.reason,
+                'name': str(block.reason).partition('.')[0],  # required
```
willdurand, line 122: Can `.partition()` return nothing (or empty tuple)?

src/olympia/blocklist/models.py
eviljeff: > Why is there two spaces between the two sentences? 

https://www.theatlantic.com/science/archive/2018/05/two-spaces-after-a-period/559304/ 😝 (In other words, I was taught to type in such a bygone era that consideration of typewriter mechanics was still a possible thing)

> Also, shouldn't this comment be added to the same method above (in the other class)?

It's not a boolean in `Block` so it's an actually useful function.  So I'd lean "no" ... but maybe?

src/olympia/blocklist/utils.py
eviljeff: `''.partition('.')` = `('', '', '')`

PR #15044, opened 2020-07-23
Clear needs human review reviewer flags on unlisted approval by humans

### LINKED ISSUE #14943: Clear reviewer flags when reviewer approves an unlisted version
It's possible for a reviewer to approve an unlisted version (not confirm approval) if the version was not auto-approved because of a scanner rule. We currently don't clear the flags (`needs_human_review` and `needs_human_review_by_mad`) for this version but we should.

diox:
Fixes #14943

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -1139,6 +1139,14 @@ class ReviewUnlisted(ReviewBase):
         subject = u'Mozilla Add-ons: %s %s signed and ready to download'
         self.log_action(amo.LOG.APPROVE_VERSION)
 
+        if self.human_review:
+            if self.version.needs_human_review:
+                self.version.update(needs_human_review=False)
+
+            VersionReviewerFlags.objects.filter(
+                version=self.version
+            ).update(needs_human_review_by_mad=False)
+
         self.notify_email(template, subject, perm_setting=None)
 
         log.info(u'Making %s files %s public' %
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -1127,11 +1127,6 @@ class ReviewFiles(ReviewBase):
 
 class ReviewUnlisted(ReviewBase):
 
-    def clear_all_needs_human_review_flags(self):
-        """Clear needs_human_review flags, but unlike listed review, only do
-        it on the latest version, not all past versions."""
-        self.clear_specific_needs_human_review_flags(self.version)
-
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/utils.py
```diff
@@ -1139,6 +1139,14 @@ def approve_latest_version(self):
...
+            if self.version.needs_human_review:
+                self.version.update(needs_human_review=False)
+
+            VersionReviewerFlags.objects.filter(
+                version=self.version
+            ).update(needs_human_review_by_mad=False)
+
         self.notify_email(template, subject, perm_setting=None)
 
         log.info(u'Making %s files %s public' %
```
willdurand, line 1152: Might be a different issue but in `block_multiple_versions`, shouldn't we also clear the `VersionReviewerFlags`?

diox: Oh yeah I was wondering about that and forgot to ask. @wagnerand , should we clear human review/human review by mad flags in the block multiple versions action as well ?

wagnerand: I think it makes sense to clear the flags in that case, yes.

diox: I added https://github.com/mozilla/addons-server/pull/15044/commits/d221ea9855805e96ab1cb7f6b4dee9a6b699d28e which clears the flag in `block_multiple_versions` action but also refactors things so that we have centralized methods to clear the needs human review flags.

diox:
@eviljeff re-requested a review for https://github.com/mozilla/addons-server/commit/d221ea9855805e96ab1cb7f6b4dee9a6b699d28e

src/olympia/reviewers/utils.py
```diff
@@ -1124,6 +1127,11 @@ def log_sandbox_message(self):
 
 class ReviewUnlisted(ReviewBase):
 
+    def clear_all_needs_human_review_flags(self):
+        """Clear needs_human_review flags, but unlike listed review, only do
+        it on the latest version, not all past versions."""
+        self.clear_specific_needs_human_review_flags(self.version)
```
eviljeff, line 1133: I don't understand why it's necessary to override this to change the behaviour - all of the methods that set the flags are overridden anyway?  (maybe just `raise` if it's just for safety)

diox: Good catch. I removed the override, simplifies things further 4792137f946c3f337b1f2bc1734cbfc78a9c04c3

PR #15047, opened 2020-07-23
Expose optional permissions in the Addon API

### LINKED ISSUE #15036: Expose optional permissions in the Addon API
This supports the quarterly goal "Display Optional Permissions" ([PRD](https://docs.google.com/document/d/1px08vsGEbI3dggO8xj-6CVAJsh1MAk9IFROBfv3mT6Y/edit?usp=sharing)). In order to display optional permissions on AMO, we'll need them to be returned in the API response. 

We currently return `permissions` in [`current_version.files[].permissions`](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#version-detail), so it might make sense to return optional permissions in a similar way. A couple of options are:
- We could expand `permissions` to be an object with `required` and `optional` keys, each of which contains an array of permissions.
- We could rename `permissions` to `required_permissions` and include a new `optional_permissions` array as well.

bobsilverberg:
Fixes #15036


docs/topics/api/addons.rst
```diff
@@ -362,6 +362,7 @@ This endpoint allows you to fetch a single version belonging to a specific add-o
     :>json boolean files[].is_mozilla_signed_extension: Whether the file was signed with a Mozilla internal certificate or not.
     :>json boolean files[].is_restart_required: Whether the file requires a browser restart to work once installed or not.
     :>json boolean files[].is_webextension: Whether the file is a WebExtension or not.
+    :>json array files[].optional_permissions[]: Array of the optional webextension permissions for this File, as strings.  Empty for non-webextensions.
```
willdurand, line 365: Why is there two spaces between the two sentences?

src/olympia/files/models.py
```diff
@@ -354,6 +358,23 @@ def webext_permissions_list(self):
         except WebextPermission.DoesNotExist:
             return []
 
+    @cached_property
+    def optional_permissions_list(self):
```
willdurand, line 362: Could this be renamed `webext_optional_permissions` or simply `optional_permissions`?

src/olympia/files/utils.py
```diff
@@ -652,6 +655,7 @@ class FSyncMixin(object):
     We need this to make sure that on EFS / NFS all data is immediately
     written to avoid any data loss on the way.
     """
+
```
willdurand, line 658: 🐼

src/olympia/files/utils.py
```diff
@@ -587,6 +589,7 @@ class SigningCertificateInformation(object):
     extension, so is signed already with a special certificate.  We want to
     know this so we don't write over it later, and stop unauthorised people
     from submitting them to AMO."""
+
```
willdurand, line 592: 🐼

src/olympia/files/tasks.py
```diff
@@ -35,14 +35,17 @@ def extract_webext_permissions(ids, **kw):
...
             parsed_data = parse_xpi(file_.current_file_path, user=user)
             permissions = parsed_data.get('permissions', [])
+            optional_permissions = parsed_data.get('optional_permissions', [])
             # Add content_scripts host matches too.
             for script in parsed_data.get('content_scripts', []):
                 permissions.extend(script.get('matches', []))
-            if permissions:
+            if permissions or optional_permissions:
                 log.info('Found %s permissions for: %s' %
                          (len(permissions), file_.pk))
```
willdurand, line 44: We should probably update this log message to include optional permissions.

src/olympia/files/models.py
```diff
@@ -183,9 +185,11 @@ def from_upload(cls, upload, version, platform, parsed_data=None):
             # Add content_scripts host matches too.
             for script in parsed_data.get('content_scripts', []):
```
willdurand, line 186: I see that we extend `permissions` with other manifest properties here. Is there any optional permission elsewhere in the `manifest` file? cc @Rob--W

Rob--W: Only `optional_permissions` contains permissions.

Does the code here behave correctly when passed an invalid value for the key or array?

And since it's related (to mandatory permissions), does the code also account for the devtools_page manifest key (which triggers a permission warning)?

willdurand: > Does the code here behave correctly when passed an invalid value for the key or array?

Would you have an example in mind?

> And since it's related (to mandatory permissions), does the code also account for the devtools_page manifest key (which triggers a permission warning)?

yeah, we handle that.

Rob--W:
```
// Invalid, rejected by Firefox:
"optional_permissions": 123,
"optional_permissions": { },

// Accepted by Firefox, but unsupported values are ignored:
"optional_permissions": null,
"optional_permissions": [1234],
"optional_permissions": [{"__proto__": {}}],
```

willdurand: Oh I see now, thanks! We'll likely add some more test cases to make sure that we can handle those values one way or another.

docs/topics/api/addons.rst
bobsilverberg: Because it was copy/pasted from the next line about `permissions`, so the real question is why did *that* line have two spaces? 😛 I'll change it in both places.

src/olympia/files/models.py
bobsilverberg: It seemed unnecessary and overly long to include `webext` in this property name, but I kept the `_list` suffix to keep it consistent with the `webext_permissions_list` property, which is why it ended up as `optional_permissions_list`. 

I could change it to just `optional_permissions`, which is in fact what I would probably have called it in a vacuum, but I still see some value in maintaining the consistency with `webext_permissions_list`. Therefore, my inclination is to just keep it as is, unless you feel strongly about it.

src/olympia/files/tasks.py
bobsilverberg: Taking another look at this, I have a feeling that the `extract_permissions` command, and this related `extract_webext_permissions` are obsolete, so rather than update the code and tests to address optional permissions, perhaps we should just delete the command and task. WDYT @diox?

src/olympia/files/models.py
willdurand: The `webext` prefix is useful to know that this is coming from a related model I think. I'd go for `optional_permissions` then because, if the `webext` prefix is not needed, then the `_list` is definitely useless.

src/olympia/files/utils.py
bobsilverberg: That's my auto-formatter doing the wrong thing. Let me get rid of those.

src/olympia/files/models.py
bobsilverberg: Code has been added to ignore any invalid optional permissions: https://github.com/mozilla/addons-server/pull/15047/files#diff-ecb8df04bee7aac11123f3729e23cfafR366-R373, and a test was added as well. It covers all the cases above except for `null`, so I'll add that to the test.

src/olympia/files/models.py
diox: FWIW I think consistency is important too - I don't particularly care about keeping the `webext_` prefix or the `_list` suffix, but whatever we choose we should do the same on the other property, IMHO.

bobsilverberg: In this case I think just having `permissions` and `optional_permissions` is the best way to go. I'll make that change.

src/olympia/files/tasks.py
willdurand: ha yeah, it seems to be called in a management command but maybe we don't need it anymore

bobsilverberg: We discussed this on Matrix, and we do in fact need this to extract the optional permissions to backfill all the add-ons, but some additional changes will be needed. I'm going to defer that to a follow-up issue.

bobsilverberg: I opened https://github.com/mozilla/addons-server/issues/15062. All outstanding issues for this PR have been addressed. I'll merge it once Travis is happy.

bobsilverberg:
Oops, just noticed that this never got an r+ from @willdurand. I just assumed it had one. So I will wait for an approval before merging.

src/olympia/addons/indexers.py
```diff
@@ -170,7 +170,9 @@ def get_mapping(cls):
                         'strict_compatibility': {
                             'type': 'boolean', 'index': False},
                         'status': {'type': 'byte'},
-                        'webext_permissions_list': {
+                        'permissions': {
+                            'type': 'keyword', 'index': False},
+                        'optional_permissions': {
```
willdurand, line 175: @diox does that require a full reindex?

src/olympia/addons/tests/test_indexers.py
```diff
@@ -63,7 +63,7 @@ def expected_fields(cls, include_nullable=True):
         _indexed_translated_fields = ('name', 'description', 'summary')
         analyzer_fields = list(chain.from_iterable(
             [['%s_l10n_%s' % (field, lang) for lang, analyzer
-             in SEARCH_LANGUAGE_TO_ANALYZER.items()]
+              in SEARCH_LANGUAGE_TO_ANALYZER.items()]
```
willdurand, line 66: 🐼

src/olympia/addons/tests/test_serializers.py
```diff
@@ -31,6 +31,7 @@
 class AddonSerializerOutputTestMixin(object):
     """Mixin containing tests to execute on both the regular and the ES Addon
     serializer."""
+
```
willdurand, line 34: 🐼

src/olympia/addons/tests/test_indexers.py
bobsilverberg: Searching and replacing in VSCode seems to also do some automatic formatting which is not desired. 😞

src/olympia/addons/indexers.py
diox: Yes. It's also not backwards-compatible, which is a problem.

bobsilverberg: I fixed this up in the serializer, as suggested by @diox on Matrix.

willdurand:
We'll have to trigger an ES reindex in -dev, then -stage and -prod (and add instructions in the push doc for that).

bobsilverberg:
https://github.com/mozilla/addons-server/pull/15076 is going to be needed as well, but it depends on this, so I'm going to merge this to make reviewing that simpler.

PR #15060, opened 2020-07-24
Drop obsolete lightweight theme update endpoint

### LINKED ISSUE #15059: Drop obsolete lightweight themes update endpoint
The ` /<locale>/themes/update-check/` endpoint (`services/theme_update.py`) is no longer necessary. We may want to keep the `MigratedLWT` model a little longer, but we can drop the code serving the updates.

diox:
Fixes #15059


services/theme_update.py
```diff
@@ -1,120 +0,0 @@
...
-            start_response('404 Not Found', [])
-            return ['']
-
-        try:
-            query_string = environ.get('QUERY_STRING')
-            update = MigratedUpdate(locale, id_, query_string)
-            is_migrated = update.is_migrated
-            if is_migrated:
-                output = (
-                    update.get_json() if settings.MIGRATED_LWT_UPDATES_ENABLED
```
willdurand, line 107: Could we remove this setting? (`MIGRATED_LWT_UPDATES_ENABLED`)

diox: 66e4d1a770

PR #15066, opened 2020-07-27
RE: Adjust the hero API code to deliver uploaded images

### LINKED ISSUE #14473: Adjust the hero API code to deliver uploaded images
Supports https://github.com/mozilla/addons-server/issues/14471
Depends on #14472, #14474 

After #14472 and #14474 land, adjust the code for the hero API to deliver images from the location where the uploaded files are stored, as opposed to from the CDN where they are currently stored.

Question: Will that new location also be served via a CDN, and if not are there any issues with not serving the images from a CDN?

xlisachan:
Fixes #14473 

The `image_url` property has been updated in the `PrimaryHero` model (`hero/models.py`), which is what the serializer was using for the hero image in `hero/serializers.py`. I also revised the following tests: 
- `discovery/tests/test_admin.py`
- `hero/tests/test_models.py`
- `hero/tests/test_serializers.py`

It failed the CI check for `test_translations_reading_from_multiple_db_using` in `translations/tests/test_models.py` (`reason='Needs django-queryset-transform patch to work'`). I'm not sure how to go about solving this...


src/olympia/hero/models.py
```diff
@@ -157,7 +157,10 @@ def __str__(self):
 
     @property
     def image_url(self):
-        return f'{FEATURED_IMAGE_URL}{self.image}' if self.image else None
+        if self.select_image:
+            return f'{self.select_image.custom_image.url}'
+        else:
+            None
```
eviljeff, line 163: This code is a little weird - I guess you meant to write `return None`? 
(You didn't like the `return xxx if yyy else zzz` style?)

src/olympia/hero/admin.py
```diff
@@ -30,7 +30,6 @@ class Media:
     model = PrimaryHero
     fields = (
         'description',
-        'image',
```
eviljeff, line 33: You can drop the field from the model too.   (but we don't want to actually remove the database field until a following deploy to production.  steps:
- add `null=True` kwarg to `image` in `PrimaryHero`
- make migrations
- delete `image` from `PrimaryHero`
- (a few days later so it's not in the same tag;) run makemigrations again to actually delete the field.

eviljeff:
> It failed the CI check for test_translations_reading_from_multiple_db_using in translations/tests/test_models.py (reason='Needs django-queryset-transform patch to work'). I'm not sure how to go about solving this.

I wouldn't worry about it if it doesn't fail on travis-ci.  Some of the translation related tests need extra setup to work locally.

src/olympia/hero/models.py
xlisachan: No I like the style! flake8 was giving me problems about the line being too long. I'll revert to the original style and add parenthesis.

src/olympia/hero/admin.py
xlisachan: > * add `null=True` kwarg to `image` in `PrimaryHero`
> * make migrations
> * delete `image` from `PrimaryHero`
> * (a few days later so it's not in the same tag;) run makemigrations again to actually delete the field.

Should I migrate after the first `makemigrations`?

eviljeff: yes.  The last step can be done whenever - we're pretty lax about cleaning up the database once it's not referenced in the code.

Issue #15070, opened 2020-07-27
Add 'Build #' prefix to Fenix versions


willdurand:
Usage stats for Fenix have build IDs instead of more user friendly **app versions**. It's not clear whether we can rely on a more meaningful version string so let's make it clear that those `202023343` strings are build IDs while we figure out how to retrieve more user-friendly version strings.

We're not going to add this prefix in the exported data.

For QA: I don't think you'll be able to verify this issue because of the lack of Fenix data in -dev/-stage. Firefox Desktop data shouldn't be affected.

ioanarusiczki:
@willdurand 
- I actually found an add-on on dev:
https://addons-dev.allizom.org/en-US/firefox/addon/firefox-color/statistics/usage/applications/?last=90
It's "Android Build #2015752521" displayed instead of  "Android 2015752521"

- other data such as usage by date doesn't seem to be affected.  I took a look at dev versus stage. 

Verified on AMO dev - FF78(Win10).

willdurand:
>     * I actually found an add-on on dev:
>       [addons-dev.allizom.org/en-US/firefox/addon/firefox-color/statistics/usage/applications/?last=90](https://addons-dev.allizom.org/en-US/firefox/addon/firefox-color/statistics/usage/applications/?last=90)
>       It's "Android Build #2015752521" displayed instead of  "Android 2015752521"

That's unexpected but good for QA :p

ioanarusiczki:
@willdurand Yes, a nice surprise there :)

PR #15071, opened 2020-07-27
Add 'Build #' prefix to Fenix app versions (usage stats)

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15070

---

This is a UI change only, controlled by a waffle switch in case we're able to change the data in BigQuery.

## Screenshot

Note that Fenix versions are `20200534` and not `78.0`, etc. but that's the prefix this patch adds:

Screen Shot 2020-07-27 at 11 15 39


static/js/impala/stats/manager.js
```diff
@@ -518,9 +520,14 @@ z.StatsManager = (function() {
...
 
+        // Fenix app versions are build IDs so we need a prefix for end users.
+        var versionPrefix = metric === 'apps' &&
+            key === '{aa3c5121-dab2-40e2-81ca-7ea25febc110}' &&
+            useFenixBuildIDs ? 'Build #' : '';
+
         if (metric in csv_keys) {
             if (key in csv_keys[metric]) {
-                return csv_keys[metric][key] + ' ' + parts.join(' ');
+                return csv_keys[metric][key] + ' ' + versionPrefix + parts.join(' ');
```
eviljeff, line 530: what's this actually doing?  The end result in the UI?

willdurand: sorry, added a screenshot now

static/js/impala/stats/manager.js
```diff
@@ -518,9 +520,14 @@ z.StatsManager = (function() {
             key = parts[0];
         parts = parts.slice(1);
 
+        // Fenix app versions are build IDs so we need a prefix for end users.
+        var versionPrefix = metric === 'apps' &&
+            key === '{aa3c5121-dab2-40e2-81ca-7ea25febc110}' &&
```
eviljeff, line 525: code-coding-tastic .... if this is just a temporary fix then fine.  If it's longer term I'd be concerned about this - would we even look in js code for something fenix application specific?

src/olympia/stats/migrations/0005_create_switch_for_fenix_build_ids.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.13 on 2020-07-27 08:43
+
+from django.db import migrations
+
+
+def create_waffle_switch(apps, schema_editor):
+    Switch = apps.get_model("waffle", "Switch")
+
+    Switch.objects.create(
+        name="use-fenix-build-ids",
```
eviljeff, line 10: It doesn't hurt, but why the switch ?  It's not a big feature we're launching and/or liable to break is it?

willdurand: The switch is needed because I am working on the data as well. For now, we didn't identify a better value than the build ID (which seemed to be documented as the "app version" similar to what we have for FF Desktop). If we find a better app version for Fenix, then we'll update the BigQuery data and we'll need to disable this UI change on AMO.

Having the switch lets us work in parallel and won't require an AMO push to revert the small change. In the end, we'd like to have app versions for Fenix that look like Desktop (so `78.0`, `79.1`, etc.)

Issue #11980, opened 2019-07-31
Upgrade Python to 3.8


EnTeQuAk:
We are currently using Python 3.6 in our docker environment. We should fairly easy be able to upgrade to ~~3.7~~ 3.8.

Python 3.6 is already in "security fix only"-mode and doesn't receive any further bugfixes should they be needed.

Aside from many useful features, Python 3.7 is supposed to be one of the fastest releases of the CPython environment yet.

diox:
We were (are?) blocked by https://github.com/celery/celery/issues/4849 when we migrated to Python 3 for the record, and that's why we chose Python 3.6.

EnTeQuAk:
I know, all this can theoretically tackled once we upgraded to 4.3 as part of #9068

eviljeff:
Celery wasn't the only python3.7 blocker during the 2.7->3.6 upgrade iirc. Have you tried 3.7 on a branch yet?

EnTeQuAk:
> Celery wasn't the only python3.7 blocker during the 2.7->3.6 upgrade iirc. Have you tried 3.7 on a branch yet?

No, I only filed the issue :laughing:

tapaswenipathak:
Hi folks: Can I PR for the ticket?

EnTeQuAk:
> Hi folks: Can I PR for the ticket?

This issue isn't really suitable for contributors. It requires lots of testing and can't be solved before  #9068 has been solved.

Please take a look at other issues that have the [`contrib: welcome`](https://github.com/mozilla/addons-server/labels/contrib%3A%20welcome) label.

diox:
#9068 is solved, so we should be able to try the upgrade to 3.7 again (maybe we'll discover other blockers)

tapaswenipathak:
Heyo folks - unsubscribing.

diox:
Made https://github.com/diox/olympia/tree/python3.7 let's see if that works... Follow along in https://travis-ci.org/diox/olympia/builds/589973239

diox:
Ahah, it fails because `jingo_minify_helpers` is using an undocumented property that has been removed in Python 3.7:
```
src/olympia/lib/jingo_minify_helpers.py:54: in ensure_path_exists
    if e.errno != os.errno.EEXIST:
E   AttributeError: module 'os' has no attribute 'errno'
```

From https://docs.python.org/3/whatsnew/3.7.html :
[...]

PR #15074, opened 2020-07-27
Switch to Python 3.8

diox:
Fixes https://github.com/mozilla/addons-server/issues/11980


diox:
@eviljeff and @willdurand you might also be interested in having a look at this. It touches bad zip file handling in validation task decorator, but that seems ok.

src/olympia/devhub/tasks.py
```diff
@@ -180,11 +180,12 @@ def wrapper(task, results, pk, *args, **kwargs):
...
-        except BadZipfile:
+        except BadZipFile:
+            # If we raised a BadZipFile we can return a single exception with
+            # a generic message indicating the zip is invalid or corrupt.
             results = deepcopy(amo.VALIDATOR_SKELETON_EXCEPTION_WEBEXT)
-            annotations.insert_validation_message(
-                results, type_='error',
-                message=ugettext('Invalid or corrupt add-on file.'))
+            results['messages'][0]['message'] = ugettext(
+                'Invalid or corrupt add-on file.')
```
eviljeff, line 188: Why can't we use annotations here any longer? (it's still used a few lines above)

diox:
> it's mainly just changing `BadZipfile` to `BadZipFile`?

I didn't have to do that, in fact. Both still work, but while I was looking at the documentation I noticed it was deprecated so I figured I might as well take care of it now.

src/olympia/devhub/tasks.py
diox: Because I don't want to insert a new error, I want to replace the base one. 

The main difference with before is that before, we ended up catching some invalid zip files later, when validating the files inside, and we raised a `ValidationError`. That was turned by this decorator into a generic linter error using the code just below this:
```python
        except Exception as exc:
            log.exception('Unhandled error during validation: %r' % exc)
            results = deepcopy(amo.VALIDATOR_SKELETON_EXCEPTION_WEBEXT)
            return results
```

Note that in that case, there is only one error returned. 

Now, with Python 3.8, `BadZipFile` is raised earlier than before for the same invalid zips. So we end up in that `except` clause, where we previously returned *two* errors: `Invalid or corrupt add-on file` *and* the generic ` coming from `VALIDATOR_SKELETON_EXCEPTION_WEBEXT`, `Sorry, we couldn't load your WebExtension.`. To stay consistent with what we were doing before for those zips, I'm changing the way we deal with `BadZipFile` and only return *one* error in that case, the `Invalid or corrupt add-on file`. This means I can no longer use `insert_validation_message`, as that would add a new error message instead of replacing the generic one.

src/olympia/files/utils.py
```diff
@@ -742,16 +742,12 @@ def __init__(self, source, mode='r', force_fsync=False):
...
         self.force_fsync = force_fsync
-        self.is_valid = self.initialize_and_validate()
+        self.initialize_and_validate()
 
     def initialize_and_validate(self):
         """
         Runs some overall archive checks.
         """
-        # Shortcut to avoid expensive check over and over again
-        if getattr(self, 'is_valid', False):
```
bobsilverberg, line 752: Why did you remove this optimization? Did it never get used?

diox: It wasn't - if you look at the whole patch, there is only one place where we used `is_valid` left (that I removed), and even there it was dead code, because an exception would have be thrown before we got to that point.

diox:
I'll merge this after the tag just to be safe.

src/olympia/files/utils.py
bobsilverberg: Right, but wasn't the idea of this that if we had to call `initialize_and_validate` on the same object more than once we could shortcut it? I assume there's nowhere in the code where `initialize_and_validate` might be called more than once on the same object?

diox: It's called at `__init__` time so there is no way around it (I think historically it wasn't always the case, and that's why this property made sense)

PR #15078, opened 2020-07-27
re-add the PrimaryHero admin tool inline in PromotedAddon

### LINKED ISSUE #15028: Reimplement PrimaryHero admin tools
The PR for #14859 removes the PrimaryHero from DiscoveryItem.  We need it back!

eviljeff:
fixes #15028 - the queryset lookup is unoptimized (just copied from DiscoveryItemAdmin to skip some of the default transforms on Addon) because addressing that turned out to be a rabbit hole so will be in a follow-up.

### DIFF
--- a/src/olympia/promoted/admin.py
+++ b/src/olympia/promoted/admin.py
@@ -1,8 +1,10 @@
 from django.contrib import admin
+from django.db.models import Prefetch
 from django.forms.models import modelformset_factory
 
 from olympia.addons.models import Addon
 from olympia.discovery.admin import SlugOrPkChoiceField
+from olympia.hero.admin import PrimaryHeroInline
 
 from .forms import AdminBasePromotedApprovalFormSet
 from .models import PromotedAddon, PromotedApproval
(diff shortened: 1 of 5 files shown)


src/olympia/promoted/admin.py
```diff
@@ -81,3 +109,9 @@ def formfield_for_foreignkey(self, db_field, request=None, **kwargs):
             return SlugOrPkChoiceField(**kwargs)
         return super().formfield_for_foreignkey(
             db_field, request, **kwargs)
+
+    def has_delete_permission(self, request, obj=None):
+        qs = PrimaryHeroInline.model.objects.filter(enabled=True)
```
bobsilverberg, line 114: Just wondering, why is this using `PrimaryHeroInline.model.objects` rather than just `PrimaryHero.objects`?

src/olympia/promoted/admin.py
```diff
@@ -81,3 +109,9 @@ def formfield_for_foreignkey(self, db_field, request=None, **kwargs):
...
+        if obj and list(qs) == [getattr(obj, 'primaryhero', None)]:
```
bobsilverberg, line 115: Will this be true if there is only one enabled `primaryhero`, and it is the same as the current `primaryhero`? If so, couldn't the same thing be done by just counting the number of enabled `primaryhero`s if the current one is enabled, and checking that there is only one?

src/olympia/promoted/tests/test_admin.py
```diff
@@ -44,7 +62,8 @@ def test_can_see_in_admin_with_discovery_edit(self):
         assert self.list_url in response.content.decode('utf-8')
 
     def test_can_list_with_discovery_edit_permission(self):
-        PromotedAddon.objects.create(addon=addon_factory(name='FooBâr'))
+        pro = PromotedAddon.objects.create(addon=addon_factory(name='FooBâr'))
+        PrimaryHero.objects.create(promoted_addon=pro)
```
bobsilverberg, line 66: Why is this needed when the test does nothing afterwards related to the `PrimaryHero`?

src/olympia/promoted/admin.py
eviljeff: I thought not having to import the `PrimaryHero` model was better, but I dunno?

src/olympia/promoted/admin.py
eviljeff: Yes.  I did it this way in DiscoveryItemAdmin so I just copied it over, but I can change it if you want?

src/olympia/promoted/admin.py
bobsilverberg: I dunno either. I was just asking. If there is no difference then I guess just leave it as is.

src/olympia/promoted/admin.py
bobsilverberg: It seems like the code will be a bit simpler with my suggestion. I had to stop and think about this one. Simpler is usually better, imo.

src/olympia/promoted/tests/test_admin.py
eviljeff: https://github.com/mozilla/addons-server/pull/15078/commits/c3759260619cb9d23bded23584fb801b2834f92c

src/olympia/promoted/admin.py
eviljeff: https://github.com/mozilla/addons-server/pull/15078/commits/c3759260619cb9d23bded23584fb801b2834f92c

Issue #12052, opened 2019-08-07
"TypeError: i.renderTo is undefined" appears after a graphic from "Statistics" page has "No data available"


madalincm:
### STR:
1. Access AMO homepage: https://addons.allizom.org/en-US/firefox/.
2. Click on an add-on that has the "Visit stats dashboard" link in the "More information" section. (e.g. uBlock Origin: https://addons.allizom.org/en-US/firefox/addon/ublock-origin/?src=homepage-collection-featured)
3. Click on the "Visit stats dashboard" link in the "More information" section.
4. Once on the "Statistics" page click on "Custom..." from the "For last:" section and select the present day.
5. Click on any buttons from the "Group by:" or "For last:" section.

### Expected result:
The statistical graphics should be displayed.

### Actual result:
A "TypeError: i.renderTo is undefined" error will appear after a graphic has "No data available", which causes the following graphics to not load anymore.

### Notes:
The bug is reproduced only after having a graphic with "No data available". The bug is not reproduced if graphics display data.
Reproduced in amo-dev, amo-stage and amo-prod with Fx68, Win10x64.
After refresh the page loads the graphics again.

![2019-08-07_11h50_50](https://user-images.githubusercontent.com/15685441/62609214-07039e80-b90a-11e9-8917-96bbb5d64a29.gif)

ioanarusiczki:
I see this issue reproducible on AMO dev - FF76(Win10). 

![reproducible](https://user-images.githubusercontent.com/33448286/85020937-c1467200-b179-11ea-8322-25e600f0ff3f.png)

willdurand:
I am not able to reproduce locally. I can reproduce in -dev but not locally, even with assets compiled so I dunno. I spent way too much time on this issue already.

willdurand:
I am wondering if this isn't some sort of timing issue like we load data and quickly ask the app to load new data (which is empty). I cannot reproduce locally or in a test suite so I am a bit hopeless.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-15

PR #15086, opened 2020-07-28
Fix dates in stats export links (csv/json)

### LINKED ISSUE #14826: csv reports are generated within the chosen interval plus one extra day
From https://github.com/mozilla/addons-server/issues/14796#issuecomment-652378499

STR:
1. Go to an addon with stats such as https://addons-dev.allizom.org/en-US/firefox/addon/some-addon/statistics/usage/?last=30
2. Generate a report with a custom date range  let's say 12.04.2020 - 05.06.2020 and export it in .csv format. 
3. Open the report and observe the intervals.

Actual result:
 the csv report is generated with an extra day 12.04 -06.06 and if 06.06 has data this is displayed too. 

Expected:
report should be for 12.04.2020 - 05.06.2020

Notes:
That's reproducible for both beta and legacy - FF77(win10).

willdurand:
Fixes #14826 

---

This patch also includes some JS tests. Things would be a bit better if we could use ES6 in our source files.

### DIFF
--- a/static/js/impala/stats/helpers.js
+++ b/static/js/impala/stats/helpers.js
@@ -29,7 +29,7 @@ function WorkerPool(size) {
                 worker.addEventListener('message', function(e) {
                     if (job.cb.call(job.ctx, e.data, worker)) {
                         worker.terminate();
-                        delete worker;
+                        worker = null;
                         workers--;
                         nextJob();
                     };
--- a/static/js/impala/stats/stats.js
+++ b/static/js/impala/stats/stats.js
@@ -1,104 +1,138 @@
-(function() {
+(function (jQuery, window) {
+  'use strict';
 
-    $(function() {
-        "use strict";
+  // `$` is passed by jQuery itself when calling `jQuery(stats_stats)`.
+  var stats_stats = function ($, injectedSessionStorage) {
+    var internalSessionStorage =
+      injectedSessionStorage || window.sessionStorage;
 
-        // Modify the URL when the page state changes, if the browser supports pushState.
-        if (z.capabilities.replaceState) {
-            $(window).on('changeview', function(e, view) {
-                var queryParams = {},
-                    range = view.range;
-                if (range) {
-                    if (typeof range == 'string') {
-                        queryParams.last = range.split(/\s+/)[0];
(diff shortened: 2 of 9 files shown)


static/js/impala/stats/stats.js
```diff
@@ -1,104 +1,122 @@
...
+      baseURL = $('.primary').attr('data-base_url');
+
+    $(window).on('changeview', function (e, newView) {
+      _.extend(view, newView);
+      var metric = view.metric;
+
+      var range = normalizeRange(view.range);
+
+      // See: https://github.com/mozilla/zamboni/commit/4263102
+      if (
```
willdurand, line 85: The actual fix is this `if`.

static/js/impala/stats/helpers.js
```diff
@@ -29,7 +29,7 @@ function WorkerPool(size) {
                 worker.addEventListener('message', function(e) {
                     if (job.cb.call(job.ctx, e.data, worker)) {
                         worker.terminate();
-                        delete worker;
+                        worker = null;
```
willdurand, line 32: We can't `delete` in strict mode.

willdurand:
I'm going to see if I can reproduce/fix https://github.com/mozilla/addons-server/issues/12052 with this test suite. If that's the case, this test suite would be very useful. If not, well it's less useful but still interesting because we could at least execute the code once in Travis.

tests/js/zamboni/stats.spec.js
```diff
@@ -0,0 +1,127 @@
...
+    dateMock.clear();
+  });
+
+  describe('stats/stats.js', () => {
+    const report = 'apps';
+
+    let stats_stats;
+
+    beforeEach(() => {
+      stats_stats = require('../../../static/js/zamboni/stats-all.js')
```
willdurand, line 38: Using the compiled (or just concatenated?) version is easier because of the various dependencies. Fortunately, all deps are not global. We could use the minified version but that makes reading errors tricky.

willdurand: It needs to be `require`'ed that way because of the few global mocks declared above.

willdurand:
@diox updated the patch to use the `assets` toxenv

PR #15089, opened 2020-07-28
Update log messges for `extract_optional_permissions`

bobsilverberg:
Just a quick follow-up to #15076 to update the log messages

### DIFF
--- a/src/olympia/files/management/commands/extract_optional_permissions.py
+++ b/src/olympia/files/management/commands/extract_optional_permissions.py
@@ -19,7 +19,7 @@ class Command(BaseCommand):
         pks = files.values_list('pk', flat=True)
 
         log.info('Using %s file pks to extract permissions, max pk: %s' %
-                 (pks.count(), pks[len(pks) - 1]))
+                 (pks.count(), pks[pks.count() - 1]))
         if pks:
             chunked_tasks = create_chunked_tasks_signatures(
                 extract_optional_permissions, list(pks), chunk_size=100)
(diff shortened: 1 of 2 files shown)


src/olympia/files/management/commands/extract_optional_permissions.py
```diff
@@ -18,7 +18,8 @@ def handle(self, *args, **options):
         ).order_by('pk')
         pks = files.values_list('pk', flat=True)
 
-        log.info('Using %s file pks to extract permissions' % pks.count())
+        log.info('Using %s file pks to extract permissions, max pk: %s' %
+                 (pks.count(), pks[len(pks) - 1]))
```
eviljeff, line 22: `pks[-1]` is the same as `pks[len(pks) - 1]`

src/olympia/files/tasks.py
```diff
@@ -22,8 +22,13 @@
 @task
 @use_primary_db
 def extract_optional_permissions(ids, **kw):
-    log.info('[%s@%s] Extracting permissions from Files, starting at id: %s...'
-             % (len(ids), extract_optional_permissions.rate_limit, ids[0]))
+    log.info(
+        '[%s@%s] Extracting permissions from Files, from id: %s to id: %s...'
+        % (
+            len(ids), extract_optional_permissions.rate_limit, ids[0],
+            ids[len(ids) - 1]
```
eviljeff, line 29: same here.

src/olympia/files/management/commands/extract_optional_permissions.py
bobsilverberg: I tried that. I got "Negative indexing is not supported" :-(

eviljeff: ah, must be a queryset thing.  I feel `len(pks)` should be `pks.count()` again though, or both `len` - seems inconsistent to have the same thing two different ways in one statement.

Issue #15101, opened 2020-07-29
Code Manger fails on .ts


erosman:
Example:
```
https://code.addons.mozilla.org/en-US/compare/764197/versions/2266957...4859565/?allowSlowPages=true&path=scripts%2FInputArea%2FContentEditable.ts
https://reviewers.addons.mozilla.org/en-US/firefox/files/browse/3403941/file/scripts/InputArea/ContentEditable.ts#top
```

Code manger displays:
```
Syntax highlighting was disabled for performance
+++ 0--- 0
```

Legacy File Viewer displays:

```
module GhostText.InputArea {
--
  | /**
  | * Implementation for a contenteditable element.
  | *
  | * @licence The MIT License (MIT)
  | * @author Guido Krömer <mail 64 cacodaemon 46 de>
  | */
  | export class ContentEditable implements IInputArea {

.....
```

Same file without compare:
```
https://code.addons.mozilla.org/en-US/browse/764197/versions/4859565/?path=scripts%2FInputArea%2FContentEditable.ts
https://reviewers.addons.mozilla.org/en-US/firefox/files/browse/3403941/file/scripts/InputArea/ContentEditable.ts
```

Code manger displays: (nothing, blank)
```
  
```

Legacy File Viewer displays:

```
module GhostText.InputArea {
--
  | /**
  | * Implementation for a contenteditable element.
  | *
  | * @licence The MIT License (MIT)
  | * @author Guido Krömer <mail 64 cacodaemon 46 de>
  | */
  | export class ContentEditable implements IInputArea {

.....
```

wagnerand:
In the API response `hunks` is empty. Also
```
mimetype | "video/MP2T"
mime_category | "binary"
```

ioanarusiczki:
@willdurand I could view the .ts files on dev - FF78(Win10). 
[Browse](https://code.addons-dev.allizom.org/en-US/browse/620063/versions/1691753/?path=scripts%2FInputArea%2FAceCodeEditor.ts) view
[Compare](https://code.addons-dev.allizom.org/en-US/compare/620063/versions/1691753...1691754/?path=scripts%2FInputArea%2FCodeMirror.ts) view

The [API response](https://addons-dev.allizom.org/api/v5/reviewers/addon/620063/versions/1691753/?file_only=true&file=scripts%2FInputArea%2FContentEditable.ts) returns MIME type: text/javascript and so does code-manager as information. 
Also: "mime_category": "text"

PR #15102, opened 2020-07-29
Fix wrong mimetype for ts files

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15101


willdurand:
I was able to reproduce locally. This is enough to fix the issue fully.

Issue #15105, opened 2020-07-29
Add support for `ActivityLog.objects.for_versions()`


diox:
We need to be able to build an `ActivityLog` queryset for multiple versions easily. We already have `for_addons()` so let's replace `for_version()` with `for_versions()`.

Blocks https://github.com/mozilla/addons-server/issues/14814

PR #15106, opened 2020-07-29
Simplify ActivityLog.objects.for_*() methods and implement for_versions()

diox:
The simplification makes us do JOINs instead of an extra query and avoids a costly evaluation of the results just to build a list of pks when there are many logs returned.

`for_versions()` replaces `for_version()`, works the same way as `for_addons()`, if a single `Version` instance is passed it converts it to a tuple of one item behind the scenes.

Fixes https://github.com/mozilla/addons-server/issues/15105

### DIFF
--- a/src/olympia/activity/models.py
+++ b/src/olympia/activity/models.py
@@ -218,18 +218,13 @@ class ActivityLogManager(ManagerBase):
         if isinstance(addons, Addon):
             addons = (addons,)
 
-        vals = (AddonLog.objects.filter(addon__in=addons)
-                .values_list('activity_log', flat=True))
+        return self.filter(addonlog__addon__in=addons)
 
-        if vals:
-            return self.filter(pk__in=list(vals))
-        else:
-            return self.none()
+    def for_versions(self, versions):
+        if isinstance(versions, Version):
+            versions = (versions,)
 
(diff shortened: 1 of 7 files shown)


src/olympia/activity/models.py
```diff
@@ -238,19 +233,13 @@ def for_groups(self, groups):
...
-        vals = (BlockLog.objects.filter(block=block)
-                .values_list('activity_log', flat=True))
-        return self.filter(pk__in=list(vals))
+        return self.filter(blocklog__block=block)
 
     def for_guidblock(self, guid):
-        vals = (BlockLog.objects.filter(guid=guid)
-                .values_list('activity_log', flat=True))
-        return self.filter(pk__in=list(vals))
+        return self.filter(blocklog__block__guid=guid)
```
bobsilverberg, line 242: Why is the one above `blocklog__block=block` but this is `blocklog__block__guid=guid`? Is `blocklog__guid` not a thing? I ask because the old versions use `BlockLog.objects.filter(block=block)` and `BlockLog.objects.filter(guid=guid)`.

diox: Good catch, I went a little too quickly on this one. It shouldn't change functionality but does save us a join.

Issue #12334, opened 2019-09-11
document best practices for creating django migrations


eviljeff:
moved from #12333, part of #3705

@diox said

> It's also worth coming up with some documentation about how to customize migrations to make them backwards-compatible - something we do right now when writing SQL migrations. Maybe something about what strategy to use in which case (delay committing the migration itself when removing a field, add new fields as nullable, etc) ?

diox:
Things to cover:
- How to customize migration that adds a new field (maybe we need to always add fields as nullable, not doing anything special with the migration, then remove the null later, adding a second migration to do that?)
- When to commit a migration that removes a column / table
- What to check for when we add a migration (in particular, was there some migrations that hadn't been generated before, and what would be the consequences of applying those)
- How to deal with big tables, how to test suspected long migrations with ops, etc.

diox:
Something worth noting regarding adding new columns: If you don't add a default, and the field is not nullable, django migrations asks if you want to set a one-off

Regardless of your choice, here is what it does in the migration:
- `ALTER` with a database default matching your choice (causing all existing data to get that default value)
- `ALTER` to remove that database default

It does this because Django handles default values in python and never leaves a database one.

This is a clever trick... that we never want to use! During deployment, we run on the old code for a limited time before promoting the new tag, so if were were to run that second `ALTER`, old code would never provide a default value and any insertions it'd make would fail.

So, instead, we want to do this 99% of the time:
- Add all new field as nullable (create & commit the migration for this) 
- Wait to push to happen
- Make the field non-nullable (create & commit migration for this)

diox:
https://github.com/3YOURMIND/django-migration-linter might be useful

ChetanKhanna:
Hi! :)
Is this issue something that I can take on? If so, where exactly should the docs go?

diox:
This is unlikely to be a good bug for external contributors, no. https://addons-pm.herokuapp.com/contrib/good-first-bugs/?dir=desc&sort=updatedAt has a list of good first bugs across AMO repositories.

PR #15122, opened 2020-07-31
hold for pre-review all Promoted addons in applicable groups, not just recommended

### LINKED ISSUE #14862: Pre-review for Promoted addons
> When an add-on is added to either of the Verified, Spotlight, or Line lists, it won’t be fully added until its next version passes review, like it’s done with Recommended Add-ons.

> [P2] Make auto-approval a configuration option per list.

eviljeff:
fixes #14862


src/olympia/discovery/models.py
```diff
@@ -11,10 +11,6 @@
 
 
 class DiscoveryItem(OnChangeMixin, ModelBase):
-    RECOMMENDED = 'Recommended'
-    PENDING_RECOMMENDATION = 'Pending Recommendation'
-    NOT_RECOMMENDED = 'Not Recommended'
-
```
eviljeff, line 17: drive-by clean-up of some now unused constants.

src/olympia/reviewers/migrations/0007_auto_20200730_1335.py
```diff
@@ -0,0 +1,23 @@
...
+
+    dependencies = [
+        ('reviewers', '0006_auto_20200609_1655'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='autoapprovalsummary',
+            name='is_promoted_prereview',
+            field=models.BooleanField(default=False, help_text='Is in a promoted addon group that requires pre-review'),
```
diox, line 16: That won't do it. Here and in the model you need a `NullBooleanField` (or `BooleanField` with `null=True`) for backwards-compatibility with previous code. Otherwise there won't be a database-level default (*) and code still running the old tag will fail.

(*) When migrating, django generates one to set a value in the column for every row, and then immediately throws it away, considering that you're only running new code, which isn't the case. This is also the reason why I prefer to add those new fields with `default=None` - this avoid the triple `ALTER` dance django is doing with the database default - we can then later fix the data by changing the default to `False` and then at a later push remove the `null=True` + add a migration to convert `None` to `False`.

eviljeff: https://github.com/mozilla/addons-server/pull/15122/commits/7cbd7eade4fb5a4ea2f39360c2dfea5cf6fa8a22

diox: Please change the `default` as well to avoid multiple unnecessary `ALTER TABLE`

eviljeff: https://github.com/mozilla/addons-server/pull/15122/commits/5e388c98e17677ff71f502cd1d64e4c2394410b4 but we need https://github.com/mozilla/addons-server/issues/12334 to be done really.

Issue #13682, opened 2020-03-09
Switch to Terser for JS minification


willdurand:
This will prevent errors like this one I think:

```
Parse error at /code/src/olympia/../../static/js/debug-all.js:9180,14
              function hue(h) {
              ^
ERROR: In strict mode code, functions can only be declared at top level or immediately within another function.
    at JS_Parse_Error.get (eval at <anonymous> (/deps/node_modules/uglify-js/tools/node.js:18:1), <anonymous>:71:23)
    at fatal (/deps/node_modules/uglify-js/bin/uglifyjs:312:27)
    at run (/deps/node_modules/uglify-js/bin/uglifyjs:255:9)
    at Object.<anonymous> (/deps/node_modules/uglify-js/bin/uglifyjs:173:5)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
Minifying /code/src/olympia/../../static/js/debug-all.js (using UglifyJS)
Unchanged files skipped for minification: 18
```

But the main benefit is to allow newer JS syntax (and Terser is better maintained that Uglify).

willdurand:
This is an error in the `less` (js) library. Since it is only used in debug mode (it seems), I lowered the priority. I don't think we can do much right now.

diox:
Maybe we should pin a lower version ? It used to work fine...

willdurand:
I can bisect to find what has changed. I am not sure it is the less code and the error is a bit "dumb" because it is still valid js code...

eviljeff:
> I am not sure it is the less code 

I can't link to it because it's copied over at build time, but it'd be in /static/js/node_lib/less.js if you look locally.

> and the error is a bit "dumb" because it is still valid js code...

so it _is_ valid to have a function declaration in the middle of another function?  
https://github.com/mishoo/UglifyJS/issues/3286 was closed as not a bug.

AlexandraMoga:
@willdurand are we still pushing the Terser upgrade this week or has it been rescheduled for the next milestone? I'm looking at #15193

willdurand:
[...]

PR #8694, opened 2018-06-29
 Fold jingo_minify into our code

EnTeQuAk:
This will simplify the upgrade to Django 1.11 and allow us to keep this
under control, upstream isn't properly maintained.

* Copies jingo-minify into our code
* Removed all parts we aren't using (yui, git integration, `{{ inline_css }}` template tag etc)
* Try to clean up how subprocess calls are being made
* Cleaned up tests

This doesn't change anything regarding how jingo-minify works. It's still using good old `cat` and other ugly things but the code now looks nicer.

See https://github.com/mozilla/addons-server/pull/8694/commits/742297fef9d75ade43ce0b416d9b645d59a71e61 for more specific details about the cleanup from the original jingo-minify dump.

Fixes #8532

PR #15129, opened 2020-07-31
Switch to Terser for JS minification

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/13682

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -85,10 +85,11 @@ LESS_BIN = env(
 CLEANCSS_BIN = env(
     'CLEANCSS_BIN', default='node_modules/less/bin/lessc')
 
-# Path to uglifyjs (our JS minifier).
+# Path to our JS minifier.
 # Set as None to use YUI instead (at your risk).
-UGLIFY_BIN = env(
-    'UGLIFY_BIN', default='node_modules/uglify-js/bin/uglifyjs')
+JS_MINIFIER_BIN = env(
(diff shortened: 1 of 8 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -85,10 +85,11 @@ def path(*folders):
 CLEANCSS_BIN = env(
     'CLEANCSS_BIN', default='node_modules/less/bin/lessc')
 
-# Path to uglifyjs (our JS minifier).
+# Path to our JS minifier.
 # Set as None to use YUI instead (at your risk).
```
bobsilverberg, line 89: Is this comment still valid? I didn't see anything in `_minify` that suggests this, nor anywhere else that `JS_MINIFIER_BIN` is used.

willdurand: Looks like this was indeed removed in https://github.com/mozilla/addons-server/pull/8694

static/js/impala/stats/stats.js
```diff
@@ -2,7 +2,7 @@
   'use strict';
 
   // `$` is passed by jQuery itself when calling `jQuery(stats_stats)`.
-  var stats_stats = function ($, injectedSessionStorage) {
+  const stats_stats = function ($, injectedSessionStorage) {
```
eviljeff, line 5: 😱 `const`!

bobsilverberg: When's the PR coming to convert everything to `const`? 😉

willdurand: https://github.com/mozilla/addons-server/pull/15131

PR #15131, opened 2020-07-31
Introduce Prettier

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15148


package.json
```diff
@@ -6,8 +6,8 @@
     "node": ">= 10.13"
   },
   "scripts": {
-    "prettier": "prettier --write '/code/static/**/*.js'",
-    "prettier-ci": "prettier --list-different '/code/static/**/*.js' || (echo '\n\nThis failure means you did not run `make format` before committing\n\n' && exit 1)"
+    "prettier": "prettier --write '/code/**/*.js'",
+    "prettier-ci": "prettier --list-different '/code/**/*.js' || (echo '\n\nThis failure means you did not run `make format` before committing\n\n' && exit 1)"
```
muffinresearch, line 10: Any reason to not use `prettier -c ...`

willdurand: Probably didn't exist when we added that in addons-frontend. Thanks!

willdurand:
So this is currently failing because we don't rely on the docker-based env in travis-ci, mmm

willdurand:
I am going to land this now to avoid conflicts

PR #15168, opened 2020-08-04
Remove views/code about info requests

### LINKED ISSUE #14899: Remove obsolete views/code about expired info requests

diox:
Info requests are no longer possible, replaced with delayed rejections  simple reviewer reply with no deadline).

Fixes #14899


src/olympia/devhub/forms.py
```diff
@@ -788,39 +787,14 @@ class VersionForm(WithSourceMixin, forms.ModelForm):
...
-    clear_pending_info_request = forms.BooleanField(required=False)
 
     class Meta:
         model = Version
-        fields = ('release_notes', 'clear_pending_info_request',
-                  'approval_notes', 'source',)
+        fields = ('release_notes', 'approval_notes', 'source',)
 
     def __init__(self, *args, **kwargs):
         self.request = kwargs.pop('request')
```
eviljeff, line 796: do we still need to pass and store the request on the form now?  (it was just used for the `log_and_notify` in save, right?)

src/olympia/devhub/views.py
```diff
@@ -1038,31 +1037,19 @@ def version_edit(request, addon_id, addon, version_id):
...
         if 'version_form' in data:
-            # VersionForm.save() clear the pending info request if the
-            # developer specifically asked for it, but we've got additional
-            # things to do here that depend on it.
-            had_pending_info_request = bool(addon.pending_info_request)
             data['version_form'].save()
 
             if 'approval_notes' in version_form.changed_data:
-                if had_pending_info_request:
-                    log_and_notify(amo.LOG.APPROVAL_NOTES_CHANGED, None,
```
eviljeff, line 1049: do we need to `log_and_notify` when there's a delayed rejection pending instead?

src/olympia/devhub/forms.py
diox: I originally removed it, but then changed my mind because it meant looking at all the callers, but there seems to be only one view that uses it, so yeah I'll go ahead and remove it.

src/olympia/devhub/views.py
diox: I don't think so. Delayed rejections can't be cancelled by regular reviewers, a developer needs to submit a new version to fix the issues (or appeal to an admin). So notifying here doesn't help much. (Reviewers will be subscribed to notifications about new versions for that add-on when they submit a delayed rejection, though)

src/olympia/devhub/forms.py
diox: 319cee5deb59538bdc1097f40f855cc2b85589db

PR #15180, opened 2020-08-04
Remove legacy file viewer

### LINKED ISSUE #15171: Remove legacy file viewer
The legacy file viewer is no longer used. Let's remove it.

wagnerand:
Fix #15171.

I might have removed too much, or not enough. Please give it a close look. Thank you!


src/olympia/files/file_viewer.py
```diff
@@ -1,512 +0,0 @@
-import codecs
```
eviljeff, line 1: you've left the file there as 0 byte file?

src/olympia/files/forms.py
```diff
@@ -1,139 +0,0 @@
-from collections import defaultdict
```
eviljeff, line 1: same with this one.

src/olympia/files/file_viewer.py
wagnerand: Ah, thanks. That was unintentional. I kept deleting the code, then deleted all empty files and apparently missed two.

src/olympia/lib/cache.py
```diff
@@ -82,53 +80,6 @@ def get(self, delete=False):
         return res
 
 
-class Token(object):
```
diox, line 85: `Message` above can probably be removed as well, but this can be done as a follow-up (looks like most of this class can go really, `assert_cache_requests` and `CacheStatTracker` are only used in like one test, should be possible to replace with something even simpler)

src/olympia/files/cron.py
```diff
@@ -1,40 +0,0 @@
...
-            log.info('Removing extracted files: %s, from %sd.' % (full, day))
-
-            # Remove all files.
-            # No need to remove any caches since we are deleting files from
-            # yesterday or before and the cache-keys are only valid for an
-            # hour. There might be a slight edge-case but that's reasonable.
-            shutil.rmtree(full)
-
-
-def cleanup_validation_results():
```
diox, line 35: That's a scary function!

eviljeff: at the time, validation would be re-run in the reviewer tools on demand if there wasn't a validation report already.

wagnerand: Are you saying we should keep that function? I was part of `CRON_JOBS` (in settings_base.py) but not referenced anywhere else

src/olympia/lib/cache.py
wagnerand: Removed `Message` class in f5122de500.

src/olympia/files/cron.py
diox: Definitely don't keep this. We depend on `FileValidation` existing in a couple places now.

PR #15190, opened 2020-08-05
restrict deleting the only approved version of promoted addons

### LINKED ISSUE #15121: restrict developers from deleting current versions when approved for promotion
We currently prevent a developer from deleting the current version of a recommended add-on if by deleting it the addon becomes not recommended (i.e. the previous version isn't approved for recommendation).  I'm assuming we need to generalise this to other promoted add-ons..

@jvillalobos which other classes of promoted add-on, if any, should have the same restriction applied?

eviljeff:
fixes #15121

### DIFF
--- a/src/olympia/versions/models.py
+++ b/src/olympia/versions/models.py
@@ -31,7 +31,7 @@ from olympia.amo.urlresolvers import reverse
 from olympia.amo.utils import sorted_groupby, utc_millesecs_from_epoch
 from olympia.applications.models import AppVersion
 from olympia.constants.licenses import LICENSES_BY_BUILTIN
-from olympia.constants.promoted import PROMOTED_GROUPS_BY_ID, RECOMMENDED
+from olympia.constants.promoted import PROMOTED_GROUPS_BY_ID
 from olympia.constants.scanners import MAD
 from olympia.files import utils
 from olympia.files.models import File, cleanup_file
@@ -728,10 +728,14 @@ class Version(OnChangeMixin, ModelBase):
                 file_obj, theme_data=None, header_only=header_only).items()}
 
     def can_be_disabled_and_deleted(self):
+        # see https://github.com/mozilla/addons-server/issues/15121#issuecomment-667226959  # noqa
+        # "It should apply to the <groups> that require a review to be badged"
         from olympia.promoted.models import PromotedApproval
 
-        if (self != self.addon.current_version or
-                not self.addon.promoted_group(group=RECOMMENDED)):
+        if self != self.addon.current_version or (
+            not (group := self.addon.promoted_group()) or
+            not (group.badged and group.pre_review)
+        ):
             return True
 
         previous_ver = (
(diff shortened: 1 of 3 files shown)


src/olympia/versions/models.py
```diff
@@ -728,18 +728,22 @@ def get_background_images_encoded(self, header_only=False):
...
 
     def can_be_disabled_and_deleted(self):
+        # see https://github.com/mozilla/addons-server/issues/15121#issuecomment-667226959  # noqa
+        # "It should apply to the <groups> that require a review to be badged"
         from olympia.promoted.models import PromotedApproval
 
-        if (self != self.addon.current_version or
-                not self.addon.promoted_group(group=RECOMMENDED)):
+        if self != self.addon.current_version or (
+            not (group := self.addon.promoted_group()) or
```
bobsilverberg, line 736: Won't this condition always equate to `True`? Isn't this saying "if the group that is currently returned by `self.addon.promoted_group()` is `self.addon.promoted_group()`"? If this is the case, it seems overly complex to code it this way, rather than simply assigning `group = self.addon.promoted_group()` above the `if` and removing this clause from the `if`.

src/olympia/versions/tests/test_models.py
```diff
@@ -600,7 +601,7 @@ def test_version_deleted_not_synced_to_basket(
         new_version.delete()
         assert sync_object_to_basket_mock.delay.call_count == 0
 
-    def test_can_be_disabled_and_deleted(self):
+    def test_promoted_can_be_disabled_and_deleted(self):
         addon = Addon.objects.get(id=3615)
         # A non-recommended addon can have it's versions disabled.
```
bobsilverberg, line 606: I wonder if these comments should be updated to say "promoted" instead of "recommended", and whether it would be worth using different promoted groups in the test, as opposed to just recommended?

src/olympia/versions/tests/test_models.py
```diff
@@ -661,6 +662,27 @@ def test_can_be_disabled_and_deleted(self):
         assert not version_d.can_be_disabled_and_deleted()
         assert addon.is_recommended
 
+    def test_unbadged_promoted_can_be_disabled_and_deleted(self):
```
bobsilverberg, line 665: Nit: It's actually unbadged and also not pre-review.

src/olympia/versions/models.py
eviljeff: > Won't this condition always equate to True?

how so?

> Isn't this saying "if the group that is currently returned by self.addon.promoted_group() is self.addon.promoted_group()"? 

it's assigning `group` to the value of `self.addon.promoted_group()` and then testing the group is truthy (`not` truthy in this case)

> If this is the case, it seems overly complex to code it this way, rather than simply assigning group = self.addon.promoted_group() above the if and removing this clause from the if.

we don't want to run the database query in `Addon.promoted_group` if it's not the current version of the add-on.  There are certainly other implementations, but this seemed like the ideal use for the new (for 3.8) operator.

src/olympia/versions/tests/test_models.py
eviljeff: it kinda needs to be the same group throughout the test to test all the edge cases.  I thought about changing it to another promoted group but it doesn't give us anything new in terms of test coverage.  We could repeat it x more times to cover all the groups?  My thinking was this test would continue to cover all the edge cases and the next test double check the basic with some different groups.

src/olympia/versions/models.py
bobsilverberg: I misinterpreted the code. I thought it was both assigning and testing. It's both assigning and returning.

Issue #15189, opened 2020-08-05
Store hash of add-on GUIDs in the database


willdurand:
For the download stats, we'll be using hashed add-on GUIDs as identifiers. While we could compute this value on the fly in BQ queries, we need the hashes in the CRON job that updates the `weekly_downloads` values. Therefore, let's add the hash of each add-on GUID, likely in the `AddonGUID` model.

willdurand:
For QA, the issue is related to the "AddonGUID"/"ReusedGUID" feature. Things should work as before when creating a new add-on or updating the GUID if that is possible.

ioanarusiczki:
@willdurand 
- Updating a GUID is no longer possible because of https://github.com/mozilla/addons-server/issues/14700
- The only way is to submit an addon with a guid that uses BQ data. 
- Right now all guids from the [list ](https://github.com/mozilla/bigquery-etl/blob/3cefa96e15cebbded340633bed240db36bb8cff5/sql/amo_dev/amo_stats_dau_v2/query.sql) have been used on AMO dev and AMO Stage.

willdurand:
@ioanarusiczki I think you can delete an add-on and re-submit a new one with the same GUID

ioanarusiczki:
@willdurand And to be the owner of the addon too (which I keep forgetting) 
It worked for me. I deleted `@testpilot-containers` and submitted again an addon and it also loaded stats data. 

https://addons-dev.allizom.org/en-US/firefox/addon/statistics-reloaded/statistics/?last=30

Verified on dev with FF80(win10).

PR #15191, opened 2020-08-05
Store hash of add-on GUIDs in the database

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15189


src/olympia/addons/models.py
```diff
@@ -2170,6 +2171,11 @@ class AddonGUID(ModelBase):
     guid = models.CharField(max_length=255, null=False, db_index=True)
     addon = models.OneToOneField(
         Addon, null=False, on_delete=models.CASCADE, unique=True)
+    hashed_guid = models.CharField(max_length=100, null=True)
```
willdurand, line 2174: We could maybe lower the `max_length` but 🤷 

Also, I think we have to use `null=True` because we're adding a new field to an existing table and we'll run the backfill after the creation of this new column in the DB. @diox should we file an issue to set this field to `null=False` later?

diox: Yes.

src/olympia/addons/migrations/0019_auto_20200805_1350.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("addons", "0018_auto_20200803_1311"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="addonguid",
+            name="hashed_guid",
+            field=models.CharField(max_length=100, null=True),
```
eviljeff, line 16: I thought the problem occurred if the guid length was above 80?

eviljeff: actually, isn't the hash always 64 chars long?

src/olympia/addons/migrations/0020_backfill_hashed_guids.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 2.2.15 on 2020-08-05 13:57
+
+from django.db import migrations
+from django.db.models import Func, F
+
+
+def backfill_hashed_guids(apps, schema_editor):
+    AddonGUID = apps.get_model("addons", "AddonGUID")
+    AddonGUID.objects.filter(hashed_guid=None).update(
+        hashed_guid=Func(F("guid"), 256, function="SHA2")
```
eviljeff, line 10: call me suspicious, but are we 100% sure this is producing the same hash as `hashlib.sha256(self.guid.encode()).hexdigest()`?

willdurand: Yeah, I checked manually a few GUIDs against the FF implementation, an online too to compute SHA256 hashes, the MySQL function, and the python implementation.

src/olympia/addons/migrations/0019_auto_20200805_1350.py
willdurand: correct, fixed.

src/olympia/addons/migrations/0020_backfill_hashed_guids.py
diox: I'm not worried about any differences since guids are supposed to be ascii so there shouldn't be any implementation surprises, but I would rather do this in small chunks using a dedicated `adhoc` task (computing the hashes in python to leave the database some time to breathe). Computing and storing millions of hashes in one query is probably not very nice on the db...

Could you check with ops ? If they say it's fine I don't mind, but otherwise, let's go the task route.

willdurand: @bqbn what do you think? This query should perform an `UPDATE` on `addons_reusedguid` and will update all rows.

bqbn: If there are millions of hashes to be processed by the db, then definitely take the task route!

I don't think no one wants to sit in the push window and wait too long for this to complete. Besides, db migrations on the Addon table often cause metadata locks in the replicas.

Thanks for catching this @diox. :)

willdurand: @bqbn That would be on the `addons_reusedguid` table. How many rows do we have on prod?

bqbn: 1.9 millions.

willdurand: Okay, that's a lot. I will update this PR to remove the backfill migration, and I'll work on a new backfill solution when I am back.

willdurand: done.

diox (approved):

PR #15198, opened 2020-08-05
Add internal API to create AppVersions

### LINKED ISSUE #14649: Add API to create new `AppVersion` instances
We need to automate `AppVersion` creation. Our idea is to provide an API that would be called by Firefox release management when a new version (beta or release) is released. That API would require a specific permission we would give their user.

diox:
Fixes #14649

### DIFF
--- a/src/olympia/applications/tests.py
+++ b/src/olympia/applications/tests.py
@@ -5,7 +5,7 @@ from unittest import mock
 
 from olympia import amo
 from olympia.amo.templatetags.jinja_helpers import url
-from olympia.amo.tests import TestCase
+from olympia.amo.tests import APITestClient, reverse_ns, TestCase, user_factory
 from olympia.applications.models import AppVersion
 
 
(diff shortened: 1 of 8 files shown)


src/olympia/applications/tests.py
```diff
@@ -29,6 +29,158 @@ def test_appversions_feed(self):
...
+        assert response.status_code == 201
+        assert AppVersion.objects.count() == 3
+        assert AppVersion.objects.filter(
+            application=amo.FIREFOX.id, version='42.0').exists()
+        assert AppVersion.objects.filter(
+            application=amo.FIREFOX.id, version='42.0a1').exists()
+        assert AppVersion.objects.filter(
+            application=amo.FIREFOX.id, version='42.*').exists()
+
+    def test_alpha_star_already_exists(self):
```
bobsilverberg, line 144: This name suggests that this is a test for when the star version already exists, but you don't seem to include the setup code that pre-creates the star version. 

This also suggests that there may not be sufficient assertions in these tests if it's critical to know that *only* the non-pre-created versions have been created, which is why this test currently passes. Having said that, I'm not sure it's terribly important to know that.

diox: You're right, I forgot to pre-create the star variant here. I think we do have enough assertions though: the main goal of those tests is to ensure that a) we don't accidentally create more app versions that necessary and b) we don't somehow raise an exception when one already exists. So pre-creating the missing `AppVersion` instance like I do in the tests above and below that one should be enough.

PR #15204, opened 2020-08-06
RE: Update shelf model/form

### LINKED ISSUE #15202: Update shelf model/form
Supports #14871

1. As the shelf types categories, extension, recommendations, search, and themes all use the `search` endpoint, revise `shelf_type` to  `endpoint`. Endpoints to only include `collections` and `search`.
2. Fix `search` endpoint to raise a validation error if `criteria` is incorrectly set (e.g., `recommended=true` - missing `?`)
3. Fix `collections` endpoint
  - Successly saves when all fields are correct (`criteria` set to a collection slug e.g., `password-managers`)
  - Raise validation error if `criteria` is incorrectly set (e.g, `passwordmanager`)

xlisachan:
Fixes #15202 

This pull request includes the following updates:
- Revises `recommendations` to `recommended` in `SHELF_TYPES`, and changes its endpoint to `search`
- Fixes form to raise a validation error for the criteria field when it returns a 404 response
- Updates `collections` endpoint

As the `categories` endpoint returns categories (not add-ons), we may need to revisit this later and remove it if it is not necessary (or remove it now and add it in later?)


src/olympia/shelves/forms.py
```diff
@@ -19,27 +19,29 @@ def clean(self):
         criteria = data['criteria']
         baseUrl = settings.INTERNAL_SITE_URL
 
-        if data['shelf_type'] in ('extension', 'search', 'theme'):
+        if data['shelf_type'] in ('extension', 'recommended', 'search',
+                                  'theme'):
             api = drf_reverse('v4:addon-search')
+            url = baseUrl + api + criteria
         elif data['shelf_type'] == 'categories':
             api = drf_reverse('v4:category-list')
```
eviljeff, line 27: I didn't look too closely at the actual api endpoints defined before - the categories shelves don't access the category list (check the network panel on AMO frontend) - it's an addons/search query.    They're _all_ addons/search queries apart from the collection shelves.

src/olympia/shelves/models.py
```diff
@@ -3,7 +3,7 @@
 from olympia.amo.models import ModelBase
 
 SHELF_TYPES = ('categories', 'collections', 'extension',
-               'recommendations', 'search', 'theme')
+               'recommended', 'search', 'theme')
```
eviljeff, line 6: what's the difference between `extension`, `theme`, `search`, and `recommended` (and `categories`)?  It looks like they all do exactly the same thing?

src/olympia/shelves/forms.py
```diff
@@ -19,27 +19,29 @@ def clean(self):
...
-            results = response.json()
-
             if response.status_code == 404:
                 raise forms.ValidationError('Check criteria - No data found')
             if response.status_code != 200:
                 raise forms.ValidationError(
-                    'Check criteria - %s' % results[0])
+                    'Check criteria - %s' % response.json()[0])
 
+            results = response.json()
```
eviljeff, line 44: not sure what this refactoring is for.

src/olympia/shelves/models.py
xlisachan: I initially thought they all went to separate endpoints, but now that it's really just two, it seems redundant to list out types. Can we rename it to endpoints and just keep `search` and `collections`?

src/olympia/shelves/forms.py
xlisachan: To help retrieve the number of add-ons returned. It's kind of redundant now that `categories` uses the `search` endpoint.

src/olympia/shelves/forms.py
xlisachan: I'll update this.

src/olympia/shelves/models.py
eviljeff: yeah I think that's best.

xlisachan:
This latest push eliminates the redundancies of shelf_type and replaces it with the endpoints of `collections` and `search`. Please review when you have a moment. Thanks!

src/olympia/shelves/migrations/0005_auto_20200807_1354.py
```diff
@@ -0,0 +1,27 @@
...
+            model_name='shelf',
+            name='endpoint',
+            field=models.CharField(choices=[('collections', 'collections'), ('search', 'search')], max_length=200),
+        ),
+        migrations.AlterField(
+            model_name='shelf',
+            name='criteria',
+            field=models.CharField(help_text='e.g., ?sort=users&statictheme?recommended=true&sort=random&type=extensionpassword-managers', max_length=200),
+        ),
+    ]
```
eviljeff, line 27: we don't really want to do this.  It'll delete the existing column for shelf_type and add a new one for endpoint - it's similar enough that we can just add `db_column='shelf_type'` in the `CharField` kwargs.  Then delete both these migrations and regenerate and there should be just a single migration with two AlterFields inside.

src/olympia/shelves/models.py
```diff
@@ -2,25 +2,24 @@
...
 class Shelf(ModelBase):
     title = models.CharField(max_length=200)
-    shelf_type = models.CharField(
-        max_length=200, choices=SHELF_TYPE_CHOICES, verbose_name='type')
+    endpoint = models.CharField(
+        max_length=200, choices=ENDPOINT_CHOICES)
     criteria = models.CharField(
         max_length=200,
-        help_text="e.g., ?recommended=true&sort=random&type=extension")
+        help_text='<i>?recommended=true&sort=random&type=extension</i>')
```
eviljeff, line 16: The html isn't really necessary in the help_text - if we really want it italic we can do it in .css (I wouldn't bother though)

xlisachan:
Thanks for the feedback! The latest push removes the html in `help_text`, and revises the migration file to include `db_column='shelf_type'` in `CharField` kwargs and rename the field from `shelf_type` to `endpoint`.

PR #15211, opened 2020-08-06
(Hopefully) fix test_reindexation_with_wipe() intermittent failure

### LINKED ISSUE #15103: FAILED src/olympia/lib/es/tests/test_commands.py::TestIndexCommand::test_reindexation_with_wipe
See: https://travis-ci.org/github/mozilla/addons-server/jobs/712851866

diox:
When using `wipe=True`, the previous index will be destroyed, so the test shouldn't try to perform a search while reindexing is taking place. 

When it worked before it was because we accidentally were running the search on the new index.

Fixes #15103

### DIFF
--- a/src/olympia/lib/es/tests/test_commands.py
+++ b/src/olympia/lib/es/tests/test_commands.py
@@ -37,11 +37,6 @@ class TestIndexCommand(ESTestCase):
 
         self.addons = []
         self.expected = self.addons[:]
-        # Monkeypatch Celerys ".get()" inside async task error
-        # until https://github.com/celery/celery/issues/4661 (which isn't just
-        # about retries but a general regression that manifests only in
-        # eager-mode) fixed.
-        self.patch('celery.app.task.denied_join_result')
 
     # Since this test plays with transactions, but we don't have (and don't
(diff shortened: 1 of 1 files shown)


src/olympia/lib/es/tests/test_commands.py
```diff
@@ -121,20 +116,22 @@ def run(self):
...
+                self.expected.append(addon_factory())
+                connection._commit()
+                connection.clean_savepoints()
+                self.refresh()
+                self.check_results(self.expected)
+
+            if len(self.expected) == old_addons_count:
+                raise AssertionError(
+                    'Could not index objects in foreground while reindexing '
+                    'in the background.')
```
eviljeff, line 134: So... what happens if you try to index a new document at the same time as a new index is being built when the old one is being wiped?  Should that work?  (or is it something we never do on live systems because it doesn't)

diox: It would try to create the index (which doesn't exist yet) automatically, which would leave a mess - I think creating the alias at the end would then fail. It should never be used in production with a live cluster unless absolutely necessary, it's more of an initial setup thing.

PR #15214, opened 2020-08-06
Update documentation about tracking external sources

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15013

### DIFF
--- a/src/olympia/stats/templates/stats/reports/campaigns.html
+++ b/src/olympia/stats/templates/stats/reports/campaigns.html
@@ -1,3 +1,11 @@
 {% extends 'stats/report.html' %}
 
 {% set title = _('Download campaigns by Date') %}
+
+{% block stats_note_link %}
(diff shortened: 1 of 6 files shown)


src/olympia/stats/templates/stats/reports/campaigns.html
```diff
@@ -1,3 +1,11 @@
 {% extends 'stats/report.html' %}
 
 {% set title = _('Download campaigns by Date') %}
+
+{% block stats_note_link %}
```
bobsilverberg, line 5: As this is added identically to a few templates, would it be worth having a jinja/django (whichever these use) helper that outputs all/most of this?

willdurand: I thought about this too but it's kinda trivial so I dunno..

willdurand:
I'll cleanup later, once we fully switched to the new download stats.

PR #15219, opened 2020-08-06
Add assertNumQueries() for the MAD queue

willdurand:


willdurand:
This is a very small patch to show no regression in an upcoming patch.

willdurand:
> Could you add a comment breaking that down? 30 is a lot, being more explicit will help.

I have no idea what we have besides test savepoints and users..

diox:
You can dump the queries with `from django.db import connection` and `print(connection.queries)` inside the `assertNumQueries` block to figure it out. Or use a trick: add `assertNumQueries(0)`, when failing the test will list everything.

willdurand:
ha well, I don't understand the queries and I am about to EOM so..

diox:
Note that just a rough description is enough, at least to start. See https://github.com/mozilla/addons-server/blob/1c3e5bbec64e3a76fb2b01fcac57068e16e379ef/src/olympia/reviewers/tests/test_views.py#L2169-L2181 for example. It's all about recognizing what is displayed on the page and matching the queries with the querysets we know we have. Django Debug Toolbar can also help give you some context when loading the page locally.

willdurand:
I see, I'll revisit this patch if I can then.

PR #15222, opened 2020-08-06
Update API code to return data about promoted add-ons

### LINKED ISSUE #14923: Update API code to return data about promoted add-ons
As part of the Promoted Add-Ons feature, an add-on can now have a promoted add-on category assigned to it, which could be one of a number of values including recommended, verified tier 1, verified tier 2, line, etc. Once these categories are assignable to add-ons, and being used, we will need the API that returns add-on data to be updated to return the promoted category of the add-on. This will effectively replace the `is_recommended` property currently returned. We will need to decide whether to remove that property, and just provide a promoted category property, or whether to return both.

The Search API will also need to be updated to deal with promoted categories, as it also currently just has a `recommended` flag that can be used both as criteria and also as a sort sequence. A separate issue (or issues) will be opened for that once we better understand the requirements.

bobsilverberg:
Fixes #14923

### DIFF
--- a/docs/topics/api/addons.rst
+++ b/docs/topics/api/addons.rst
@@ -107,7 +107,7 @@ for autocomplete though, there are a couple key differences:
     :query string platform: Filter by :ref:`add-on platform <addon-detail-platform>` availability.
     :query string tag: Filter by exact tag name. Multiple tag names can be specified, separated by comma(s).
     :query string type: Filter by :ref:`add-on type <addon-detail-type>`.
-    :>json array results: An array of :ref:`add-ons <addon-detail-object>`. Only the ``id``, ``icon_url``, ``is_recommended``, ``name``, ``type`` and ``url`` fields are supported though.
+    :>json array results: An array of :ref:`add-ons <addon-detail-object>`. Only the ``id``, ``icon_url``, ``is_recommended``, ``name``, ``promoted``, ``type`` and ``url`` fields are supported though.
 
 
 ------
@@ -175,6 +175,9 @@ This endpoint allows you to fetch a specific add-on by id, slug or guid.
     :>json string previews[].image_url: The URL (including a cachebusting query string) to the preview image.
     :>json int previews[].thumbnail_size[]: width, height dimensions of of the preview image thumbnail.
     :>json string previews[].thumbnail_url: The URL (including a cachebusting query string) to the preview image thumbnail.
+    :>json object|null promoted: Object holding promotion information about the add-on. Null if the add-on is not currently promoted.
+    :>json string promoted.category: The name of the promoted category for the add-on.
+    :>json array promoted.applications[]: Array of the application names for which the add-on is promoted.
     :>json boolean public_stats: Boolean indicating whether the add-on stats are public or not.
     :>json object ratings: Object holding ratings summary information about the add-on.
     :>json int ratings.count: The total number of user ratings for the add-on.
--- a/docs/topics/api/addons.rst
+++ b/docs/topics/api/addons.rst
@@ -176,8 +176,8 @@ This endpoint allows you to fetch a specific add-on by id, slug or guid.
     :>json int previews[].thumbnail_size[]: width, height dimensions of of the preview image thumbnail.
     :>json string previews[].thumbnail_url: The URL (including a cachebusting query string) to the preview image thumbnail.
     :>json object|null promoted: Object holding promotion information about the add-on. Null if the add-on is not currently promoted.
-    :>json string promoted.category: The name of the promoted category for the add-on.
-    :>json array promoted.applications[]: Array of the application names for which the add-on is promoted.
+    :>json string promoted.category: The name of the :ref:`promoted category <addon-detail-promoted-category>` for the add-on.
+    :>json array promoted.apps[]: Array of the :ref:`applications <addon-detail-application>` for which the add-on is promoted.
     :>json boolean public_stats: Boolean indicating whether the add-on stats are public or not.
     :>json object ratings: Object holding ratings summary information about the add-on.
     :>json int ratings.count: The total number of user ratings for the add-on.
@@ -267,6 +267,20 @@ This endpoint allows you to fetch a specific add-on by id, slug or guid.
        statictheme  Theme (Static Theme)
     ==============  ==========================================================
 
+.. _addon-detail-promoted-category:
+
+    Possible values for the ``promoted.category`` field:
+
+    ==============  ==========================================================
+             Value  Description
+    ==============  ==========================================================
+              line  Line category
+       recommended  Recommended category
+         sponsored  Sponsored category
+         spotlight  Spotlight category
+         strategic  Strategic category
+          verified  Verified category
+    ==============  ==========================================================
 
 -----------------------------
 Add-on and Version Submission
--- a/src/olympia/addons/indexers.py
+++ b/src/olympia/addons/indexers.py
@@ -271,8 +271,8 @@ class AddonIndexer(BaseSearchIndexer):
                     'promoted': {
                         'type': 'object',
                         'properties': {
-                            'category': {'type': 'keyword'},
-                            'applications': {'type': 'byte'}
+                            'group_id': {'type': 'byte'},
+                            'application_id': {'type': 'byte'}
                         }
                     },
                     'ratings': {
@@ -430,6 +430,12 @@ class AddonIndexer(BaseSearchIndexer):
         data['previews'] = [{'id': preview.id, 'modified': preview.modified,
                              'sizes': preview.sizes}
                             for preview in obj.current_previews]
+
+        data['promoted'] = {
+            'application_id': obj.promoted.application_id,
+            'group_id': obj.promoted.group_id,
+        } if obj.promoted else None
+
         data['ratings'] = {
             'average': obj.average_rating,
             'count': obj.total_ratings,
--- a/src/olympia/addons/indexers.py
+++ b/src/olympia/addons/indexers.py
@@ -391,7 +391,7 @@ class AddonIndexer(BaseSearchIndexer):
                  'default_locale', 'guid', 'hotness', 'icon_hash', 'icon_type',
                  'is_disabled', 'is_experimental', 'is_recommended',
                  'last_updated',
-                 'modified', 'promoted', 'requires_payment', 'slug',
+                 'modified', 'requires_payment', 'slug',
                  'status', 'type', 'weekly_downloads')
         data = {attr: getattr(obj, attr) for attr in attrs}
 
--- a/src/olympia/addons/indexers.py
+++ b/src/olympia/addons/indexers.py
@@ -268,6 +268,13 @@ class AddonIndexer(BaseSearchIndexer):
                             },
                         },
                     },
+                    'promoted': {
+                        'type': 'object',
+                        'properties': {
+                            'category': {'type': 'text'},
+                            'applications': {'type': 'keyword'}
+                        }
+                    },
                     'ratings': {
                         'type': 'object',
                         'properties': {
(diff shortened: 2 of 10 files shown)


willdurand:
Sorry Bob, I'm out now.

bobsilverberg:
> Sorry Bob, I'm out now.

Np. We'll need to remove you from the rotation. @diox are you able to do that? If not, then @muffinresearch?

bobsilverberg:
I had mentioned on Matrix that I was unable to get the Autocomplete API returning values for `promoted` when testing locally. I'm not sure if I missed something in my code or if it's something else.

docs/topics/api/addons.rst
```diff
@@ -175,6 +175,9 @@ This endpoint allows you to fetch a specific add-on by id, slug or guid.
     :>json string previews[].image_url: The URL (including a cachebusting query string) to the preview image.
     :>json int previews[].thumbnail_size[]: width, height dimensions of of the preview image thumbnail.
     :>json string previews[].thumbnail_url: The URL (including a cachebusting query string) to the preview image thumbnail.
+    :>json object|null promoted: Object holding promotion information about the add-on. Null if the add-on is not currently promoted.
+    :>json string promoted.category: The name of the promoted category for the add-on.
+    :>json array promoted.applications[]: Array of the application names for which the add-on is promoted.
```
eviljeff, line 180: Link to the table that lists the application names (like `app` does above)

eviljeff: ... maybe this field should be called `apps` for parity with the `app` query field?  Not sure.

docs/topics/api/addons.rst
```diff
@@ -175,6 +175,9 @@ This endpoint allows you to fetch a specific add-on by id, slug or guid.
     :>json string previews[].image_url: The URL (including a cachebusting query string) to the preview image.
     :>json int previews[].thumbnail_size[]: width, height dimensions of of the preview image thumbnail.
     :>json string previews[].thumbnail_url: The URL (including a cachebusting query string) to the preview image thumbnail.
+    :>json object|null promoted: Object holding promotion information about the add-on. Null if the add-on is not currently promoted.
+    :>json string promoted.category: The name of the promoted category for the add-on.
```
eviljeff, line 179: I wonder if we should list the possible values of `category` in a table like we do with the other constants the API returns (status, app, platform, etc)

src/olympia/addons/indexers.py
```diff
@@ -268,6 +268,13 @@ def get_mapping(cls):
                             },
                         },
                     },
+                    'promoted': {
+                        'type': 'object',
+                        'properties': {
+                            'category': {'type': 'text'},
```
eviljeff, line 274: `keyword` - it's a constant rather than a full text string.

src/olympia/addons/indexers.py
```diff
@@ -268,6 +268,13 @@ def get_mapping(cls):
...
+                            'applications': {'type': 'keyword'}
```
eviljeff, line 275: we store application compatibility as byte elsewhere so I think we should do the same here.  wdyt?

(The serializer would need some changes though.  I think you'd need to make a new serializer just for the promoted field - the field for applications would be `ReverseChoiceField` inside it)

src/olympia/addons/models.py
```diff
@@ -1323,6 +1324,28 @@ def promoted_group(self, *, group=None, application=None,
             not currently_approved or promoted.is_addon_currently_promoted)
         return promoted.group if is_promoted else NOT_PROMOTED
 
+    def promoted_applications(self):
+        """Returns a list of applications for which the addon is currently
+        promoted, or None if the addon is not currently promoted.
+        """
+        promoted_group = self.promoted_group()
+        if promoted_group != NOT_PROMOTED:
```
eviljeff, line 1332: you can just do `if promoted_group` - `NOT_PROMOTED` is falsey.  (Or just `if self.promoted_group()`)

src/olympia/addons/models.py
```diff
@@ -1323,6 +1324,28 @@ def promoted_group(self, *, group=None, application=None,
...
+            application = self.promotedaddon.application
+            if not application:
+                return [app.short for app in APP_USAGE]
+            return [application.short]
+        return None
+
+    @cached_property
+    def promoted(self):
+        promoted_group = self.promoted_group()
+        if promoted_group != NOT_PROMOTED:
```
eviljeff, line 1342: same here.  You don't really need to check for `NOT_PROMOTED` group explicitly

src/olympia/addons/models.py
```diff
@@ -1323,6 +1324,28 @@ def promoted_group(self, *, group=None, application=None,
...
+            return {
+                'category': promoted_group.api_name,
+                'applications': self.promoted_applications(),
```
eviljeff, line 1345: if `Addon.promoted_applications` is only used here maybe we should fold it all together in the same method so we don't have to get `self.promoted_group` twice.

`'applications': [self.promotedaddon.application] if self.promotedaddon.application else [app.short for app in APP_USAGE]` or similar?

src/olympia/constants/promoted.py
```diff
@@ -44,6 +47,7 @@ def __bool__(self):
 VERIFIED_ONE = PromotedClass(
     id=2,
     name=_('Verified - Tier 1'),
+    api_name='verified_tier_1',
```
eviljeff, line 50: can we confirm this `api_name`? `sponsored` has been used for the codename elsewhere so maybe use it here?  (`sponsored` will be the signing state constant too for verified tier 1)

eviljeff: https://github.com/mozilla/addons-server/pull/15221 for the signing changes.

src/olympia/constants/promoted.py
```diff
@@ -53,13 +57,15 @@ def __bool__(self):
 VERIFIED_TWO = PromotedClass(
     id=3,
     name=_('Verified - Tier 2'),
+    api_name='verified_tier_2',
```
eviljeff, line 60: `verified`? (for the same reason as VERIFIED_ONE)

docs/topics/api/addons.rst
bobsilverberg: A few places where `application` is referenced in the doc for a field it's called `app_name`. I do notice that it's never explicitly called `application`, so changing it to `apps` or `app_names` makes sense. The latter seem cumbersome, so I'll just use `apps`.

src/olympia/addons/indexers.py
bobsilverberg: I'll give that a try. I was wondering how to specify that this is a list, and not just a single value. Is that not important to do? Would it just be `'applications': {'type': 'byte'}`?

diox: Everything is technically a list (of the specified type) in ES as any field can hold multiple values.

bobsilverberg: Does this need to match what's returned by the API to a caller? The API will be returning `promoted: {category: 'recommended', apps: ['firefox', 'android']}`, but as I understand the suggestion here, this actually wouldn't result in a list, but rather a single number representing what's stored in `promotedaddon.application_id`, so that seems like a mismatch. Is that an issue? I don't fully understand exactly how all this stuff is used.

bobsilverberg: I'm trying to figure out how to implement what @eviljeff suggested and I'm a bit stuck. If the field for `applications` is to be a `ReverseChoiceField`, wouldn't the `Model` need to be `PromotedAddon` and not `Addon`? If we do that, then it seems like we'd have to duplicate some of the logic that currently exists in `Addon.promoted`. I tried just creating a new serializer for `promoted`, as suggested, and including inline code that calculates what I need, but tests are failing because it's causing a query to be run. I'm wondering if a better approach would be to create a new model property on `Addon` which can return the `promoted` dict in the format that ES wants, and then just include that in the ES serializer?

src/olympia/addons/serializers.py
```diff
@@ -481,6 +498,7 @@ class ESAddonSerializer(BaseESSerializer, AddonSerializer):
     current_version = ESCurrentVersionSerializer()
     previews = ESPreviewSerializer(many=True, source='current_previews')
     _score = serializers.SerializerMethodField()
+    promoted = ESAddonPromotedSerializer()
```
eviljeff, line 501: you need to define this in `AddonSerializer` instead.

src/olympia/addons/tests/test_serializers.py
```diff
@@ -470,6 +474,31 @@ def test_is_recommended(self):
...
+        # With a recommended theme.
+        self.addon = addon_factory(type=amo.ADDON_STATICTHEME)
+        featured_collection, _ = Collection.objects.get_or_create(
+            id=settings.COLLECTION_FEATURED_THEMES_ID)
+        featured_collection.add_addon(self.addon)
+
+        result = self.serialize()
+        promoted = result['promoted']
+        assert promoted['category'] == RECOMMENDED.api_name
+        assert promoted['apps'] == [app.short for app in amo.APP_USAGE]
```
bobsilverberg, line 500: For some reason this test is failing, but only for the `ESAddonSerializer` and not for the plain `AddonSerializer`. It's failing because rather than finding all of the apps in `promoted['apps']`, it's just finding `['firefox']`, which seems to be leftover from the previous step in the test. I don't know why this is happening for just the `ESAddonSerializer` test.

bobsilverberg:
Thanks for all your help, @eviljeff. This is ready for another look. There's an issue with one of the tests that I mentioned in a comment.

src/olympia/addons/tests/test_serializers.py
eviljeff: Ah, `TestESAddonSerializerOutput` assumes there's only one addon; the second `addon_factory` is creating another one , but the test just selects `[0]`, i.e the extension instead of the theme. (look for the "there should only be one addon in the index" comment.)

So, you can either change the test to modify the existing self.addon (change the addon type; delete the PromotedAddon) or split the theme test into a separate method.

src/olympia/addons/tests/test_serializers.py
```diff
@@ -1233,14 +1262,16 @@ def test_basic(self):
...
+            {'id', 'name', 'icon_url', 'is_recommended', 'type', 'url',
+             'promoted'}
         )
         assert result['id'] == self.addon.pk
         assert result['name'] == {'en-US': str(self.addon.name)}
         assert result['icon_url'] == absolutify(self.addon.get_icon_url(64))
         assert result['is_recommended'] == self.addon.is_recommended is False
         assert result['type'] == 'extension'
         assert result['url'] == self.addon.get_absolute_url()
+        assert result['promoted'] == self.addon.promoted
```
eviljeff, line 1274: can we add on the assertion that self.addon.promoted is `None`  (that's the only time that `self.addon.promoted` should equal the serialized JSON)

src/olympia/addons/tests/test_serializers.py
bobsilverberg: Great, thanks!

src/olympia/addons/tests/test_models.py
```diff
@@ -1673,6 +1674,56 @@ def test_promoted_group(self):
...
+        featured_collection, _ = Collection.objects.get_or_create(
+            id=settings.COLLECTION_FEATURED_THEMES_ID)
+        featured_collection.add_addon(addon)
+        del addon.promoted
+        # it's in the collection, so is now promoted.
+        promoted = addon.promoted
+        assert promoted
+        assert promoted.addon == addon
+        assert promoted.group_id == RECOMMENDED.id
+        assert promoted.application_id is None
```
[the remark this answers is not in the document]
bobsilverberg: I mentioned this on Matrix as well, but should `PromotedAddon.id` be `0`, or should it be `None`? It seems to be `None` at this point in the test.

PR #15223, opened 2020-08-06
Include deleted add-ons when calculating average_daily_users

### LINKED ISSUE #15220: Include deleted add-ons when calculating `average_daily_users`
We want to include deleted add-ons when calculating `average_daily_users` so we have usage statistics for deleted add-ons that might still be used or distributed outside of AMO.

wagnerand:
Fix #15220.

### DIFF
--- a/src/olympia/addons/cron.py
+++ b/src/olympia/addons/cron.py
@@ -42,7 +42,7 @@ def update_addon_average_daily_users(chunk_size=250):
         # In order to reset the `average_daily_users` values of add-ons that
         # don't exist in BigQuery, we prepare a set of `(guid, 0)` for most
         # add-ons.
-        Addon.objects
+        Addon.unfiltered
         .filter(type__in=amo.ADDON_TYPES_WITH_STATS)
         .exclude(guid__isnull=True)
         .exclude(guid__exact='')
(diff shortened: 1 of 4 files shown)


src/olympia/addons/cron.py
```diff
@@ -42,7 +42,7 @@ def update_addon_average_daily_users(chunk_size=250):
         # In order to reset the `average_daily_users` values of add-ons that
         # don't exist in BigQuery, we prepare a set of `(guid, 0)` for most
         # add-ons.
-        Addon.objects
+        Addon.unfiltered
```
eviljeff, line 45: This is going to return a lot more addons than currently - all the `reused-by-xxx` addons.  Though after the first run the adu will be set to `0` so it's a one time problem.  (it's fine if it works once!)

wagnerand: @eviljeff Yes, @willdurand mentioned that as well. Do we need to let ops know about that?

willdurand: We should run thw cron task as a one-off command during next push so that ops or us can monitor it.

wagnerand: https://github.com/mozilla/addons/commit/a4f2faa3a8bf3622c49ac2b3d6c1d0394616ee26

PR #15224, opened 2020-08-06
Update mad queue to exclude older listed versions

### LINKED ISSUE #15200: Update "Flagged for human review" queue
The "Flagged for human review" queue should be updated to show add-ons the following, and the following only:

* Any unlisted version that has the `needs_human_review_by_mad` flag
* If listed the latest auto-approved version (if it has the `needs_human_review_by_mad` flag). (This should be `addon.current_version` if I am not mistaken?) It can also be generally just the latest approved version, but in practice, the flag is set before a human looks at it, and cleared when a human has reviewed it.

/cc @willdurand

wagnerand:
Fix #15200.

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -329,13 +329,16 @@ class AddonManager(ManagerBase):
             # to have a version since they got the needs_human_review_by_mad
             # flag, so returning incomplete ones is acceptable.
             .filter(
-                status__in=[
+                Q(status__in=[
                     amo.STATUS_APPROVED, amo.STATUS_NOMINATED, amo.STATUS_NULL
-                ],
-                versions__files__status__in=[
+                ]),
+                Q(versions__files__status__in=[
                     amo.STATUS_APPROVED, amo.STATUS_AWAITING_REVIEW,
-                ],
-                versions__reviewerflags__needs_human_review_by_mad=True
+                ]),
+                (Q(versions__reviewerflags__needs_human_review_by_mad=True,
+                    versions__channel=amo.RELEASE_CHANNEL_UNLISTED) |
+                    Q(_current_version__reviewerflags__needs_human_review_by_mad=(  # noqa
+                        True)))
             )
             .order_by('created')
             # There could be several versions matching for a single add-on so
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -8135,6 +8135,18 @@ class TestMadQueue(QueueTest):
         assert len(links) == len(expected)
         check_links(expected, links, verify=False)
 
+        # Make sure the query count is stable
+        VersionReviewerFlags.objects.create(
+            version=version_factory(addon=addon_factory()),
+            needs_human_review_by_mad=True
+        )
+        VersionReviewerFlags.objects.create(
+            version=version_factory(addon=addon_factory()),
+            needs_human_review_by_mad=True
+        )
+        with self.assertNumQueries(30):
+            response = self.client.get(self.url)
+
     def test_only_viewable_with_specific_permission(self):
         # Post-review reviewer does not have access.
         self.user.groupuser_set.all().delete()  # Remove all permissions
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -8109,23 +8109,23 @@ class TestMadQueue(QueueTest):
             needs_admin_code_review=True,
         )
 
-        # # Mixed listed and unlisted versions. Should not show up in queue.
-        # mixed_addon = addon_factory()
-        # VersionReviewerFlags.objects.create(
-        #     version=version_factory(addon=mixed_addon,
-        #                             channel=amo.RELEASE_CHANNEL_UNLISTED),
-        #     needs_human_review_by_mad=False
-        # )
-        # VersionReviewerFlags.objects.create(
-        #     version=version_factory(addon=mixed_addon,
-        #                             channel=amo.RELEASE_CHANNEL_LISTED),
-        #     needs_human_review_by_mad=True
-        # )
-        # VersionReviewerFlags.objects.create(
-        #     version=version_factory(addon=mixed_addon,
-        #                             channel=amo.RELEASE_CHANNEL_LISTED),
-        #     needs_human_review_by_mad=False
-        # )
+        # Mixed listed and unlisted versions. Should not show up in queue.
+        mixed_addon = addon_factory()
+        VersionReviewerFlags.objects.create(
+            version=version_factory(addon=mixed_addon,
+                                    channel=amo.RELEASE_CHANNEL_UNLISTED),
+            needs_human_review_by_mad=False
+        )
+        VersionReviewerFlags.objects.create(
+            version=version_factory(addon=mixed_addon,
+                                    channel=amo.RELEASE_CHANNEL_LISTED),
+            needs_human_review_by_mad=True
+        )
+        VersionReviewerFlags.objects.create(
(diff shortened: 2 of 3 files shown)


wagnerand:
`src/olympia/lib/es/tests/test_commands.py::TestIndexCommand::test_reindexation_starting_from_one_addon` keeps failing, I gave up retrying after the third time.

src/olympia/reviewers/utils.py
```diff
@@ -255,12 +256,24 @@ def render_addon_name(self, record):
...
+            if self.single_listed_version:
+                rval.append(
+                    '<a href="%s">%s</a>' % (
+                        url,
+                        self.listed_text.format(listed_versions)
+                    )
+                )
+            else:
+                rval.append(
+                    '<a href="%s">%s</a>' % (
```
bobsilverberg, line 272: I'm confused, and maybe I'm just blind, but this statement looks identical to the one above it, in the other path of the `if`.

wagnerand: You are right. I initially added this because if `single_listed_version` is true, we don't have nothing to format `self.listed_text`, but it seems `format` is just ignored if there is no placeholder in the string.

wagnerand: .a2388487cc

src/olympia/reviewers/tests/test_views.py
```diff
@@ -8098,14 +8112,15 @@ def setUp(self):
         self.expected_addons = [listed_addon, unlisted_addon]
 
     def test_results(self):
-        response = self.client.get(self.url)
+        with self.assertNumQueries(30):
```
eviljeff, line 8115: 30 seems a lot.  Does it at least scale? (i.e. it's always 30 regardless of whether there's 1 or 100 items in the list)  Can you increase the number of addons (or whatever) and call it again to confirm?

src/olympia/addons/models.py
```diff
@@ -329,13 +329,16 @@ def get_mad_queue(self, admin_reviewer=False):
...
             .filter(
-                status__in=[
+                Q(status__in=[
                     amo.STATUS_APPROVED, amo.STATUS_NOMINATED, amo.STATUS_NULL
-                ],
-                versions__files__status__in=[
+                ]),
+                Q(versions__files__status__in=[
                     amo.STATUS_APPROVED, amo.STATUS_AWAITING_REVIEW,
-                ],
```
eviljeff, line 337: this would probably read better if you had two `.filter()` - the second with the `Q`s  in - so you didn't have wrap the `__in` filters in `Q` too.

src/olympia/addons/models.py
```diff
@@ -329,13 +329,16 @@ def get_mad_queue(self, admin_reviewer=False):
...
-                versions__reviewerflags__needs_human_review_by_mad=True
+                ]),
+                (Q(versions__reviewerflags__needs_human_review_by_mad=True,
+                    versions__channel=amo.RELEASE_CHANNEL_UNLISTED) |
```
eviljeff, line 339: This will be filtering and returning the addon if _any_ version is unlisted and any there are _any_ needs_human_review_by_mad set on the versions.  I.e. it's not testing that the same version is both unlisted and flagged. (if I'm remembering my queryset logic correctly). 

This addon would be in the queryset:
version_a: unlisted, needs_human_review_by_mad=False
version_b: listed, needs_human_review_by_mad=True
version_c (current_version): listed, needs_human_review_by_mad=False

 Just checking this is what you intended.  Can you mix in an unlisted version into the tests?

src/olympia/reviewers/tests/test_views.py
wagnerand: The patch actually just increases the queries by 1. It's 24 before this patch, 25 with it, and 30 in tests because safepoints.

src/olympia/addons/models.py
wagnerand: That doesn't sound like what we want. What would the correct filter look like?

src/olympia/reviewers/tests/test_views.py
eviljeff: that's fine.  Can you just add a second response.get with more addons to confirm it doesn't increase to 31 (or more)

wagnerand: How do I do that without duplicating the test?

src/olympia/addons/models.py
wagnerand: Doesn't `Q(A, B)` mean `A AND B`?

eviljeff: it does.  But it's `<any version of that addon>__reviewerflags__needs_human_review_by_mad` `AND` `<any version of that addon>__channel=amo.RELEASE_CHANNEL_UNLISTED`

wagnerand: I see. How do you fix that?

src/olympia/reviewers/tests/test_views.py
eviljeff: Add another addon (and add any matching flags) then `with self.assertNumQueries(30):  self.client.get(self.url)` again.  Put it at the end of the test if it'll disrupt the other asserts.

src/olympia/addons/models.py
eviljeff: does forcing it into a subquery work?
```
.filter(
    status__in=[
        amo.STATUS_APPROVED, amo.STATUS_NOMINATED, amo.STATUS_NULL
    ],
    versions__files__status__in=[
        amo.STATUS_APPROVED, amo.STATUS_AWAITING_REVIEW,
    ],
    versions__in=Version.unfiltered.filter(
        Q(reviewerflags__needs_human_review_by_mad=True, channel=amo.RELEASE_CHANNEL_UNLISTED) |
        Q(addon___current_version_id=id)
    )
```

src/olympia/reviewers/tests/test_views.py
wagnerand: Got it, thank you. Fixed in aa9505b3ec.

src/olympia/addons/models.py
wagnerand: I added this scenario in 3b16c96f1a and the current code seems to work based on that.

src/olympia/reviewers/tests/test_views.py
```diff
@@ -8095,17 +8109,39 @@ def setUp(self):
...
+                                    channel=amo.RELEASE_CHANNEL_LISTED,
+                                    created=self.days_ago(1)),
+            needs_human_review_by_mad=False
+        )
+
         self.expected_addons = [listed_addon, unlisted_addon]
 
     def test_results(self):
-        response = self.client.get(self.url)
+        with self.assertNumQueries(30):
```
diox, line 8136:
```suggestion
        with self.assertNumQueries(30):
            # 30 queries is a lot. Some of them are unfortunately scaling with
            # the number of add-ons in the queue.
            # - 2 for savepoints because we're in tests
            # - 2 for user/groups
            # - 11 for various queue counts, including current one
            #      (unfortunately duplicated because it appears in two
            #       completely different places)
            # - 3 for the addons in the queue and their files (regardless of
            #     how many are in the queue - that's the important bit)
            # - 2 for config items (motd / site notice)
            # - 2 for my add-ons / my collection in user menu
            # - 4 for reviewer scores and user stuff displayed above the queue
            # - 2 queries for first add-on to get listed/unlisted count of
            #     versions with needs human review flag
            # - 2 queries for second add-on to get listed/unlisted count of
            #     versions with needs human review flag
```

src/olympia/addons/models.py
wagnerand: According to Mat, that might have unintended consequences. Let's keep the current structure.

PR #15242, opened 2020-08-10
RE: Add the ability to reorder or change the shelves

### LINKED ISSUE #14874: Add the ability to reorder or change the shelves
Supports #14871

Add the ability to manage the shelves that will appear below the hero area on the AMO homepage. This will includes determining which shelves will appear on the homepage as well as reordering the order of each shelves if necessary.

xlisachan:
Fixes #14874 
** Currently failing Travis, as it is dependent on the migration file in #15204 **

This pull request adds the ability to assign a shelf module a position/enables its appearance on the homepage. It also allows the admin user to reorder the shelves.

The tests `discovery/tests/test_admin.py` and `shelves/tests/test_forms.py` have been updated.


src/olympia/shelves/admin.py
```diff
@@ -1,9 +1,30 @@
...
     actions = ['delete_selected']
     form = ShelfForm
+
+
+class ShelfManagementAdmin(admin.ModelAdmin):
+    list_display = ('position', 'shelf', 'enabled')
+    list_display_links = ('shelf',)
+    list_editable = ('position', 'enabled')
+    actions = ['delete_selected']
+    ordering = [F('position').asc(nulls_last=True)]
```
eviljeff, line 18: Interesting, I've not seen `F` objects in ordering before!

src/olympia/shelves/admin.py
```diff
@@ -1,9 +1,30 @@
...
+    form = ShelfManagementForm
+
+    def save_model(self, request, obj, form, change):
```
eviljeff, line 21: Mmm, I'm not a fan of defining this validation twice.  I suggest moving the `def clean` from the form to `ShelfManagement` model.  (You probably don't need `ShelfManagementForm` at that point either - just define `fields` on `ShelfManagementAdmin`)

src/olympia/shelves/models.py
```diff
@@ -27,3 +27,15 @@ class Meta:
 
     def __str__(self):
         return self.title
+
+
+class ShelfManagement(ModelBase):
+    shelf = models.OneToOneField(Shelf, on_delete=models.CASCADE)
+    enabled = models.BooleanField(default=False)
+    position = models.PositiveIntegerField(blank=True, null=True)
```
eviljeff, line 35: when is it necessary to have `position` is `None`?   Wouldn't you just delete the `ShelfManagement` object?

src/olympia/shelves/models.py
```diff
@@ -27,3 +27,15 @@ class Meta:
...
+
+    def __str__(self):
+        return f'{self.shelf}'
```
eviljeff, line 38:
```suggestion
        return str(self.shelf)
```

src/olympia/shelves/tests/test_forms.py
```diff
@@ -116,3 +117,32 @@ def test_clean_returns_empty(self):
...
+
+    def test_clean_position_required_if_enabled(self):
+        form = ShelfManagementForm({
+            'enabled': True,
+            'shelf': self.shelf.pk})
+        assert not form.is_valid(), form.errors
+        with self.assertRaises(ValidationError) as exc:
+            form.clean()
+        assert exc.exception.message == (
+            u'Position field is required to enable shelf.')
```
eviljeff, line 148: No need for `u'...'` strings - we're way past python2 compatibility 😄  (they only still exist in the codebase because it will be a huge patch to remove them all)

src/olympia/discovery/admin.py
```diff
@@ -133,6 +133,12 @@ def previews(self, obj):
         return mark_safe(''.join(translations))
 
 
+class HomepageShelves(ShelfManagement):
+    class Meta:
+        proxy = True
+        verbose_name_plural = 'homepage shelves'
```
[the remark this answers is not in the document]
xlisachan: "Sentence Case" for `verbose_name_plural`? It didn't seem so. [https://docs.djangoproject.com/en/3.0/ref/models/options/#verbose-name-plural](url)

src/olympia/shelves/models.py
xlisachan: Possibly. With the enabled field, the user has the option of keeping shelf/shelves without enabling them on the homepage. My concern was to catch the case where the admin is reordering enabled shelves and just forgets to set the position before saving.

eviljeff: it doesn't default to `0`? (or some other non-None number)

xlisachan: No, default appears to be None. Should I set it to 0?

eviljeff: yes, i think so (it might happen automatically if `null=True` wasn't specified)

xlisachan: It didn't happen automatically, but `def clean` in models would no longer be needed if default is set to `0`, correct?

eviljeff: correct.  If the field didn't allow `None` as a value then it would trigger the field's validation instead.

xlisachan:
Thanks for the feedback! This latest push includes the following updates:
- Revised the `ShelfManagement` model to set the default of `position` to 0, which eliminates the need for `def clean`
- Removed the `u'...'` strings
- Updated the return of `def __str__` for the `ShelfManagement` model

src/olympia/shelves/admin.py
eviljeff: you might not need the `F` object now `position` doesn't support null/None.

PR #15251, opened 2020-08-10
RE: Build an API endpoint for shelves

### LINKED ISSUE #14873: Build an API endpoint for shelves
Supports #14871

Build an API endpoint (`/api/v4/shelves/`) that represents a list of shelves from the previous issue ( #14872).

xlisachan:
Fixes #14873 
** Currently failing Travis, as it is dependent on the migration file in #15242 **

This pull request builds an API endpoint (`/api/v4/shelves/` - or should this be `/api/shelves`?) that represents a list of shelves to appear on the AMO homepage.

I'm not 100% sure about the value for `permission_classes` within `shelves/views.py`.

### DIFF
--- a/src/olympia/api/urls.py
+++ b/src/olympia/api/urls.py
@@ -32,6 +32,7 @@ v4_api_urls = [
     url(r'^', include('olympia.signing.urls')),
     url(r'^', include(amo_api_patterns)),
     url(r'^scanner/', include('olympia.scanners.api_urls')),
+    url(r'^shelves/', include('olympia.shelves.urls')),
 ]
 
 urlpatterns = [
--- a/src/olympia/shelves/serializers.py
+++ b/src/olympia/shelves/serializers.py
@@ -14,13 +14,16 @@ class ShelfSerializer(serializers.ModelSerializer):
         fields = ['title', 'url', 'footer_text', 'footer_pathname']
 
     def get_url(self, obj):
-        baseUrl = settings.INTERNAL_SITE_URL
         if obj.endpoint == 'search':
-            api = drf_reverse('v4:addon-search')
-            url = baseUrl + api + obj.criteria
+            api = drf_reverse(
+                'addon-search',
+                request=self.context.get('request'))
+            url = api + obj.criteria
         elif obj.endpoint == 'collections':
-            api = drf_reverse('v4:collection-addon-list', kwargs={
-                'user_pk': settings.TASK_USER_ID,
-                'collection_slug': obj.criteria})
-            url = baseUrl + api
+            url = drf_reverse(
+                'collection-addon-list',
+                request=self.context.get('request'),
+                kwargs={
+                    'user_pk': settings.TASK_USER_ID,
+                    'collection_slug': obj.criteria})
         return url
--- a/src/olympia/shelves/serializers.py
+++ b/src/olympia/shelves/serializers.py
@@ -1,18 +1,26 @@
 from rest_framework import serializers
+from rest_framework.reverse import reverse as drf_reverse
 
-from olympia.shelves.models import Shelf, ShelfManagement
+from django.conf import settings
+
+from olympia.shelves.models import Shelf
 
 
 class ShelfSerializer(serializers.ModelSerializer):
+    url = serializers.SerializerMethodField()
+
     class Meta:
         model = Shelf
-        fields = ['id', 'title', 'endpoint', 'criteria',
-                  'footer_text', 'footer_pathname']
+        fields = ['title', 'url', 'footer_text', 'footer_pathname']
 
-
-class HomepageSerializer(serializers.ModelSerializer):
-    shelf = ShelfSerializer(required=True)
-
-    class Meta:
-        model = ShelfManagement
-        fields = ['position', 'shelf', 'enabled']
+    def get_url(self, obj):
+        baseUrl = settings.INTERNAL_SITE_URL
+        if obj.endpoint == 'search':
+            api = drf_reverse('v4:addon-search')
+            url = baseUrl + api + obj.criteria
+        elif obj.endpoint == 'collections':
+            api = drf_reverse('v4:collection-addon-list', kwargs={
+                'user_pk': settings.TASK_USER_ID,
+                'collection_slug': obj.criteria})
+            url = baseUrl + api
+        return url
--- a/src/olympia/shelves/serializers.py
+++ b/src/olympia/shelves/serializers.py
@@ -1,0 +1,18 @@
+from rest_framework import serializers
+
+from olympia.shelves.models import Shelf, ShelfManagement
+
+
+class ShelfSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Shelf
+        fields = ['id', 'title', 'endpoint', 'criteria',
+                  'footer_text', 'footer_pathname']
+
+
+class HomepageSerializer(serializers.ModelSerializer):
+    shelf = ShelfSerializer(required=True)
+
+    class Meta:
+        model = ShelfManagement
+        fields = ['position', 'shelf', 'enabled']
--- a/src/olympia/shelves/tests/test_serializers.py
+++ b/src/olympia/shelves/tests/test_serializers.py
@@ -1,6 +1,10 @@
+from rest_framework.reverse import reverse as drf_reverse
+
+from django.conf import settings
+
 from olympia.amo.tests import TestCase
-from olympia.shelves.models import Shelf, ShelfManagement
-from olympia.shelves.serializers import ShelfSerializer, HomepageSerializer
+from olympia.shelves.models import Shelf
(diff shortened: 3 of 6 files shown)


src/olympia/shelves/views.py
```diff
@@ -0,0 +1,13 @@
...
+
+from olympia.shelves.models import ShelfManagement
+from olympia.shelves.serializers import HomepageSerializer
+
+
+class ShelfViewSet(viewsets.ModelViewSet):
+    queryset = ShelfManagement.objects.all()
+    permission_classes = [
+        permissions.AllowAny
+    ]
```
eviljeff, line 11:
```suggestion
    permission_classes = []
```

src/olympia/shelves/serializers.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class ShelfSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Shelf
+        fields = ['id', 'title', 'endpoint', 'criteria',
+                  'footer_text', 'footer_pathname']
+
+
+class HomepageSerializer(serializers.ModelSerializer):
```
eviljeff, line 13: So, I don't think we actually need this Serializer at all - we won't be returning any data from it directly.  The `ShelfSerializer` has everything we need.

src/olympia/shelves/serializers.py
```diff
@@ -0,0 +1,18 @@
+from rest_framework import serializers
+
+from olympia.shelves.models import Shelf, ShelfManagement
+
+
+class ShelfSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Shelf
+        fields = ['id', 'title', 'endpoint', 'criteria',
```
eviljeff, line 9: frontend won't need `id`.

src/olympia/shelves/views.py
```diff
@@ -0,0 +1,13 @@
...
+
+    serializer_class = HomepageSerializer
```
eviljeff, line 13: `ShelfSerializer`

src/olympia/shelves/views.py
```diff
@@ -0,0 +1,13 @@
+from rest_framework import permissions, viewsets
+
+from olympia.shelves.models import ShelfManagement
+from olympia.shelves.serializers import HomepageSerializer
+
+
+class ShelfViewSet(viewsets.ModelViewSet):
+    queryset = ShelfManagement.objects.all()
```
eviljeff, line 8: So, really here we want a queryset of Shelf, but filtered and ordered by some of the ShelfManagement properties.  Something like `Shelf.objects.filter(shelfmanagement__enabled=True).order_by('shelfmanagement__position')`

src/olympia/shelves/serializers.py
```diff
@@ -0,0 +1,18 @@
...
+                  'footer_text', 'footer_pathname']
+
+
+class HomepageSerializer(serializers.ModelSerializer):
+    shelf = ShelfSerializer(required=True)
+
```
eviljeff, line 15: The next issue will be to actually include the contents of the shelves in the response.

src/olympia/shelves/serializers.py
eviljeff: And `endpoint` and `criteria` won't mean much either - they should be combined into a url field.  (url can be a `SerializerMethodField` that returns a full api url - similar to how the url is being built in the clean methods)

xlisachan:
This latest push includes the following updates:
- Removed `HomepageSerializer`
- Combined `endpoint` and `criteria` into a `url` field
- Updated `serializer_class` to `ShelfSerializer` and `queryset` in `ShelfViewSet`
- Revised `test_serializers.py`

src/olympia/shelves/serializers.py
```diff
@@ -0,0 +1,26 @@
...
+    url = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Shelf
+        fields = ['title', 'url', 'footer_text', 'footer_pathname']
+
+    def get_url(self, obj):
+        baseUrl = settings.INTERNAL_SITE_URL
+        if obj.endpoint == 'search':
+            api = drf_reverse('v4:addon-search')
```
eviljeff, line 19: because this is in the API we want to always use the matching API version in the reverses too.  So strip off the `v4:` at the start of the url name and include `request=self.context.get('request')` as a kwarg to pass down the request so the api version can be resolved from that by DRF.

src/olympia/shelves/serializers.py
```diff
@@ -0,0 +1,26 @@
...
+            url = baseUrl + api + obj.criteria
+        elif obj.endpoint == 'collections':
+            api = drf_reverse('v4:collection-addon-list', kwargs={
```
eviljeff, line 22: same here.

src/olympia/shelves/tests/test_serializers.py
```diff
@@ -0,0 +1,27 @@
...
+            endpoint='search',
+            criteria='?sort=users&type=statictheme',
+            footer_text='See more populâr themes')
+
+    def test_shelf_serializer(self):
+        serializer = ShelfSerializer(instance=self.shelf)
+        assert serializer.data == {
+            'title': 'Populâr themes',
+            'url': (settings.INTERNAL_SITE_URL +
+                    drf_reverse('v4:addon-search') +
```
eviljeff, line 23: use `reverse_ns` in API related tests - it handles the api versioning.

src/olympia/shelves/views.py
```diff
@@ -0,0 +1,12 @@
...
+from olympia.shelves.models import Shelf
+from olympia.shelves.serializers import ShelfSerializer
+
+
+class ShelfViewSet(viewsets.ModelViewSet):
+    queryset = Shelf.objects.filter(
+        shelfmanagement__enabled=True).order_by('shelfmanagement__position')
+    permission_classes = []
+
+    serializer_class = ShelfSerializer
```
eviljeff, line 12: can we have a test for the view too?  Doesn't need to do anything complicated - just check that it does indeed return a couple of Shelf instances, in the correct order, and without disabled shelves.

xlisachan:
Thanks for the feedback! This latest push includes the following revisions:
- Updated `serializers.py` to match the API version in the reverse
- Use `reverse_ns` in `test_serializers.py` (I think I have to update `test_forms.py` 👀 as I used `drf_reverse` in there?)
- Added `test_views.py` - I got a `NoReverseMatch` error stating `Reverse for 'addon-search' not found. 'addon-search' is not a valid view function or pattern name` when I tried to use `ShelfSerializer(instance=shelf_a).data`. I'm not sure if lines 64-68 are acceptable.

src/olympia/shelves/views.py
xlisachan: I added a `test_views.py`, but I'm not 100% sure about it. I got a `NoReverseMatch` error stating `Reverse for 'addon-search' not found. 'addon-search' is not a valid view function or pattern name` when I tried to use `ShelfSerializer(instance=shelf_a).data`, so I wrote lines 64-68 instead. Please take a look when you have a moment.

src/olympia/shelves/tests/test_views.py
```diff
@@ -0,0 +1,68 @@
+import json
+
+from olympia.amo.tests import TestCase, reverse_ns
+from olympia.shelves.models import Shelf, ShelfManagement
+
+
+class TestShelfViewSet(TestCase):
+    def setUp(self):
+        self.url = reverse_ns('shelves-list', api_version='v4')
```
eviljeff, line 9: do you need to specify the `api_version` here?  (it should default to 'v5' which should work okay too)

src/olympia/shelves/views.py
eviljeff: yes, that's fine - ShelfSerializer now relies on a request being passed in to resolve the api url correctly so `ShelfSerializer(instance=shelf_a).data` won't work

src/olympia/shelves/tests/test_views.py
xlisachan: I'll remove `api_version` as it did work without it. I had made the endpoint for the shelves API `/api/v4/shelves/`, should it be changed?

eviljeff: > I had made the endpoint for the shelves API /api/v4/shelves/, should it be changed?

where did you do that?

src/olympia/api/urls.py
```diff
@@ -32,6 +32,7 @@
     url(r'^', include('olympia.signing.urls')),
     url(r'^', include(amo_api_patterns)),
     url(r'^scanner/', include('olympia.scanners.api_urls')),
+    url(r'^shelves/', include('olympia.shelves.urls')),
```
xlisachan, line 35: I wasn't sure where to put this.

src/olympia/shelves/tests/test_views.py
xlisachan: Please see below.

PR #15252, opened 2020-08-10
RE: Remove previous featured hero images

### LINKED ISSUE #15176: Remove previous featured hero images
Supports  #14471

After #14473 lands to have the uploaded primary hero images associated with the enabled add-ons, remove the hero images that were previously utilized.

xlisachan:
Fixes #15176 
Dependent on issue #14473 

This pull request removes the path/url in `models.py` and the previously featured hero images that will no longer be used once issue stated above is closed.


eviljeff:
@xlisachan I should have noticed at the time, but this is failing most of the tests.  Probably because `FEATURED_IMAGE_URL` is still referenced in `ImageChoiceWidget` (it can be combined with `IconChoiceWidget`)

xlisachan:
@eviljeff, the latest push updates `hero/models.py` - as `ImageChoiceWidget` was not being used elsewhere, I renamed it `IconChoiceWidget`, updated its `image_url_base` variable, and removed the original `ImageChoiceWidget`.

A test in the shelves' `test_serializers.py` failed, which is strange as the two issues aren't related...? I don't think I have access to restart the CI build anymore (the restart button is missing on my end); however, please note that test is being updated in #15881 .

eviljeff:
@xlisachan I restarted the build - it passes now.  I've seen that failure a few times - it's a)n intermittent fail we'll need to fix at some point (I'm guessing, from the type of test, due to a race condition with ES indexing

PR #15263, opened 2020-08-11
move promoted admin under discovery (for now, at least)

### LINKED ISSUE #15194: re-organise other promoted-type admin tools
We show all the shelf related admin tools under /discovery, but the PrimaryHero shelves have moved to /promoted, inside PromotedAddon, so it will be cleaner to move them all in the same place,

eviljeff:
fixes #15194


src/olympia/discovery/migrations/0007_auto_20200811_1319.py
```diff
@@ -0,0 +1,25 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('promoted', '0005_auto_20200803_1214'),
+        ('discovery', '0006_auto_20200807_2051'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
```
bobsilverberg, line 14: I didn't realize that we'd need a migration for this. This doesn't actually alter the db in any way, does it?

eviljeff: no, it's just how django keeps track of the models

PR #15264, opened 2020-08-11
Fix mad queue display

### LINKED ISSUE #15200: Update "Flagged for human review" queue

wagnerand:
Fix #15200 (again).

### DIFF
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -8228,7 +8228,7 @@ class TestMadQueue(QueueTest):
 
     def test_queue_layout(self):
         self._test_queue_layout('Flagged for Human Review', tab_position=2,
-                                total_addons=2, total_queues=3, per_page=1)
+                                total_addons=3, total_queues=3, per_page=1)
 
     def test_queue_layout_admin(self):
         # Admins should see the extra add-on that needs admin content review.
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -8110,51 +8110,59 @@ class TestMadQueue(QueueTest):
         )
 
         # Mixed listed and unlisted versions. Should not show up in queue.
-        mixed_addon = addon_factory(created=self.days_ago(2))
+        mixed_addon = addon_factory(created=self.days_ago(5))
         VersionReviewerFlags.objects.create(
             version=version_factory(addon=mixed_addon,
-                                    channel=amo.RELEASE_CHANNEL_UNLISTED,
-                                    created=self.days_ago(5)),
+                                    channel=amo.RELEASE_CHANNEL_UNLISTED),
             needs_human_review_by_mad=False
         )
         VersionReviewerFlags.objects.create(
             version=version_factory(addon=mixed_addon,
-                                    channel=amo.RELEASE_CHANNEL_LISTED,
-                                    created=self.days_ago(3)),
+                                    channel=amo.RELEASE_CHANNEL_LISTED),
             needs_human_review_by_mad=True
         )
         VersionReviewerFlags.objects.create(
             version=version_factory(addon=mixed_addon,
-                                    channel=amo.RELEASE_CHANNEL_LISTED,
-                                    created=self.days_ago(1)),
+                                    channel=amo.RELEASE_CHANNEL_LISTED),
             needs_human_review_by_mad=False
         )
 
         # Mixed listed and unlisted versions. Only the unlisted should show up.
-        mixed_addon2 = addon_factory(created=self.days_ago(1))
+        mixed_addon2 = addon_factory(created=self.days_ago(4))
         VersionReviewerFlags.objects.create(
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/utils.py
```diff
@@ -244,14 +245,14 @@ def _get_addon_name_url(self, record):
 class ScannersReviewTable(AutoApprovedTable):
     listed_text = _('Listed versions needing human review ({0})')
     unlisted_text = _('Unlisted versions needing human review ({0})')
-    filters = {'needs_human_review': True}
+    filter_listed = filter_unlisted = Q(needs_human_review=True)
```
eviljeff, line 248: hah, all the queries are single items in the end so we don't even need Q objects (having different `filters` dicts for listed and unlisted for `MadReviewTable` is enough).

src/olympia/reviewers/tests/test_views.py
```diff
@@ -8150,6 +8171,8 @@ def test_results(self):
             #     versions with needs human review flag
             # - 2 queries for second add-on to get listed/unlisted count of
             #     versions with needs human review flag
+            # - 2 queries for third add-on to get listed/unlisted count of
+            #     versions with needs human review flag
```
eviljeff, line 8175: whaaa?  Does this query count really increase by 2 per add-on?  So what's it like for 100 items?  126 queries?  I'd really recommend a follow-up that prefetches if that's the case.

src/olympia/reviewers/tests/test_views.py
```diff
@@ -8166,6 +8189,12 @@ def test_results(self):
             'Unlisted versions (2)',
             reverse('reviewers.review', args=['unlisted', addon.slug])
         ))
+        # mixed, only unlisted flagged
+        addon = self.expected_addons[2]
+        expected.append((
+            'Unlisted versions (1)',
+            reverse('reviewers.review', args=['unlisted', addon.slug])
+        ))
```
eviljeff, line 8197: Do we have any that have both unlisted and listed versions flagged to test that case?

src/olympia/reviewers/tests/test_views.py
wagnerand: Yeah, because for each add-on, we fetch the number of listed/unlisted. The other scanners queue suffers from the same. Mat noted it might be unavoidable.

src/olympia/reviewers/tests/test_views.py
wagnerand: Hm, no. I can add one, that will increase the number of queries by 2 again. Is that ok?

eviljeff: the number of queries in the test doesn't concern me, it's what the number of queries will be on prod for each view of the page.

wagnerand: I agree. Although this patch doesn't add to the impact, it's already on prod today.

src/olympia/reviewers/tests/test_views.py
eviljeff: I don't believe it's unavoidable, but it's probably not easy - it'll take some work and trial and error.  If a simple prefetch on where ever we get the list of records isn't sufficient a custom transformer could work.  And that may lead to changes to the queries in render_addon_name.

src/olympia/reviewers/tests/test_views.py
wagnerand: Filed https://github.com/mozilla/addons-server/issues/15269

src/olympia/reviewers/tests/test_views.py
```diff
@@ -8166,24 +8189,18 @@ def test_results(self):
...
-            version=version_factory(addon=addon_factory()),
-            needs_human_review_by_mad=True
-        )
-        VersionReviewerFlags.objects.create(
-            version=version_factory(addon=addon_factory()),
-            needs_human_review_by_mad=True
-        )
-        with self.assertNumQueries(30):
-            response = self.client.get(self.url)
-
```
wagnerand, line 8186: This is clearly not working for some reason, therefore I removed it.

src/olympia/reviewers/tests/test_views.py
wagnerand: Yes, it might require some larger changes, because currently the queries are separate and defined in two different places:
* To build the queue, https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/models.py#L331-L342 fetches each add-on (without its versions). 
* To render the add-on name (and the versions), https://github.com/mozilla/addons-server/blob/master/src/olympia/reviewers/utils.py#L252-L268 counts the versions per add-on in the queue

PR #15283, opened 2020-08-13
RE: The /shelfmodule/add/ page throws a 500 error if the Endpoint is not specified

### LINKED ISSUE #15282: The /shelfmodule/add/ page throws a 500 error if the Endpoint is not specified
1. Open https://addons-internal.dev.mozaws.net/en-US/admin/models/discovery/shelfmodule/add/ with an admin account
2. Add a Title and a valid Criteria but leave the Endpoint empty
3. Save the form

### What happened?
A 500 server error is received

There should be a validation for the Endpoint field to mark it as mandatory

xlisachan:
Fixes #15282 

This pull requests fixes the error when the endpoint is not specified.

### DIFF
--- a/src/olympia/shelves/forms.py
+++ b/src/olympia/shelves/forms.py
@@ -16,18 +16,22 @@ class ShelfForm(forms.ModelForm):
 
     def clean(self):
         data = self.cleaned_data
-        criteria = data['criteria']
+
+        criteria = data.get('criteria')
+        endpoint = data.get('endpoint')
         baseUrl = settings.INTERNAL_SITE_URL
 
-        if data['endpoint'] == 'search':
+        if endpoint == 'search':
             api = drf_reverse('v4:addon-search')
             url = baseUrl + api + criteria
-        elif data['endpoint'] == 'collections':
+        elif endpoint == 'collections':
             api = drf_reverse('v4:collection-addon-list', kwargs={
                 'user_pk': settings.TASK_USER_ID,
                 'collection_slug': criteria
             })
             url = baseUrl + api
+        else:
+            return
 
         try:
             response = requests.get(url)
--- a/src/olympia/shelves/tests/test_forms.py
+++ b/src/olympia/shelves/tests/test_forms.py
@@ -1,11 +1,9 @@
 import responses
 
-from rest_framework.reverse import reverse as drf_reverse
-
 from django.conf import settings
 from django.core.exceptions import ValidationError
 
-from olympia.amo.tests import TestCase
+from olympia.amo.tests import TestCase, reverse_ns
 from olympia.shelves.forms import ShelfForm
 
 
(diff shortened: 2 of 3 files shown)


src/olympia/shelves/forms.py
```diff
@@ -15,14 +15,19 @@ class Meta:
...
     def clean(self):
-        data = self.cleaned_data
-        criteria = data['criteria']
+        cleaned_data = super(ShelfForm, self).clean()
+
+        if any(self.errors):
+            return self.errors
+
+        criteria = cleaned_data.get('criteria')
+        endpoint = cleaned_data.get('endpoint')
```
eviljeff, line 24: these couple of lines that replace `data['endpoint']` with `data.get('endpoint')` (and same with `criteria`) are the fix for the issue, aren't they?  I don't understand what the rest of the change in the patch is for.

xlisachan: There would still be a `KeyError` if any of the required fields were left blank when attempting to save the form. The fix of the issue was to return any errors first with:
`if any(self.errors):`
 `            return self.errors`

I guess I didn't necessarily have to replace `data['endpoint']` with `data.get('endpoint')`, but I read that was a better way to write it. The rest of the changes were just to move the error message above the field, as opposed to the top of the form, which is more of a preference and not necessary for the the patch. Please let me know if you want me to revert those revisions.

eviljeff:
You don't need all of this code and custom handling for errors.  It just needs to only try to ping the endpoint if endpoint is specified.  That endpoint is _required_ is already handled by django forms code and will raise appropriate error messages.

```diff
------------------------- src/olympia/shelves/forms.py -------------------------
index 4d3d9a4ae0..6da6d5d41a 100644
@@ -16,18 +16,21 @@ class ShelfForm(forms.ModelForm):
 
     def clean(self):
         data = self.cleaned_data
-        criteria = data['criteria']
+        criteria = data.get('criteria')
+        endpoint = data.get('endpoint')
         baseUrl = settings.INTERNAL_SITE_URL
 
-        if data['endpoint'] == 'search':
+        if endpoint == 'search':
             api = drf_reverse('v4:addon-search')
             url = baseUrl + api + criteria
-        elif data['endpoint'] == 'collections':
+        elif endpoint == 'collections':
             api = drf_reverse('v4:collection-addon-list', kwargs={
                 'user_pk': settings.TASK_USER_ID,
                 'collection_slug': criteria
             })
             url = baseUrl + api
+        else:
+            return
 
         try:
             response = requests.get(url)
```
seems to fix the issue.

src/olympia/shelves/tests/test_forms.py
```diff
@@ -17,44 +14,39 @@ def setUp(self):
         self.criteria_404 = 'sort=users&type=statictheme'
         self.criteria_not_200 = '?sort=user&type=statictheme'
         self.criteria_empty = '?sort=users&type=theme'
-        baseUrl = settings.INTERNAL_SITE_URL
 
         responses.add(
             responses.GET,
-            baseUrl + drf_reverse('v4:addon-search') +
-            self.criteria_sea,
+            reverse_ns('addon-search') + self.criteria_sea,
```
eviljeff, line 20: We're specifically pinging the v4 api in the forms code above, because that's the version of the api that addons-frontend is currently using.  If you want to use `reverse_ns`, specify `api_version`.

src/olympia/shelves/tests/test_forms.py
```diff
@@ -17,44 +14,39 @@ def setUp(self):
...
             status=200,
             json={'count': 103})
         responses.add(
             responses.GET,
-            baseUrl + drf_reverse('v4:collection-addon-list', kwargs={
-                                  'user_pk': settings.TASK_USER_ID,
-                                  'collection_slug': self.criteria_col}),
+            reverse_ns('collection-addon-list', kwargs={
+                'user_pk': settings.TASK_USER_ID,
+                'collection_slug': self.criteria_col}),
```
eviljeff, line 27: same here with `api_version`

eviljeff: and etc in this file.

src/olympia/shelves/tests/test_forms.py
```diff
@@ -81,44 +73,40 @@ def test_clean_search_returns_404(self):
             'endpoint': 'search',
             'criteria': self.criteria_404}
         form = ShelfForm(data)
-        assert not form.is_valid()
-        with self.assertRaises(ValidationError) as exc:
-            form.clean()
-        assert exc.exception.message == (
-            'Check criteria - No data found')
+        form.is_valid()
```
eviljeff, line 76: so, now the form _is_ valid?  Are there tests that prove that these changes still result in an error here prevent the Shelf from being saved?

xlisachan:
This latest push includes the following updates:
- Removed the unneccessary changes
- Added the `api_version` to the tests
- Added a test to confirm that `def clean` checks required fields

src/olympia/shelves/tests/test_forms.py
xlisachan: I removed the unnecessary changes to the form/tests, and added a test to confirm that `def clean` checks if there are missing required fields.

PR #15293, opened 2020-08-14
Update file link in activity log

### LINKED ISSUE #15291: Broken admin activity log

1. Open `/admin/models/activity/activitylog/`

### What happened?

Oops! We had an error.

No oops!

https://sentry.prod.mozaws.net/operations/olympia-prod/issues/9394553/

Likely a regression from #15180.

wagnerand:
Fix #15291.

### DIFF
--- a/src/olympia/activity/models.py
+++ b/src/olympia/activity/models.py
@@ -9,7 +9,6 @@ from datetime import datetime
 from django.apps import apps
 from django.conf import settings
 from django.db import models
-from django.urls import reverse
 from django.utils import timezone
 from django.utils.functional import cached_property
 from django.utils.translation import ugettext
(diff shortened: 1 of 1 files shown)


src/olympia/activity/models.py
```diff
@@ -538,7 +538,7 @@ def to_string(self, type_=None):
                     validation = 'ignored'
 
                 file_ = self.f(u'<a href="{0}">{1}</a> (validation {2})',
-                               reverse('files.list', args=[arg.pk]),
+                               arg.get_url_path(arg.pk),
```
eviljeff, line 541: are you sure about this?  File.get_url_path seems to be an instance method that doesn't require the id, and is the download link (rather than an admin manage page)

wagnerand: I decided against the admin manage page because none of the other types in this method link to their admin manage page, so I will just keep it consistent. As for the parameter, it is required, but it's not the pk, but the `src` one. Updated in 100faa13abbdfe85f245b05d129107c015c5d629.

wagnerand:
Also, that whole rendering could use a do-over. Rendering html links as plaintext is not that helpful.

PR #15294, opened 2020-08-14
RE: Include the contents of the shelves in the API

### LINKED ISSUE #15284: Include the contents of the shelves in the API
Supports #14871

Add the contents of the shelves to the API endpoint that represents the list of shelves that will appear on the AMO homepage.

* Please note the linked pull request only returns the addons based on `search` criteria. The addons based on `collections` will be addressed in #15614

xlisachan:
Fixes #15284 
Dependent on #15283 (`settings_test.py`)

This pull requests updates the homepage shelves with the endpoint `search` by including the add-ons based on the shelves' crtieria within the API.

I am not sure if I am on the right track - I updated the serializer to add a field for the returned addons. As the container is not configured to call the search API, I had to repeat the code in `def clean` from `forms.py` - would it be better to save the url so it can be reused here?

### DIFF
--- a/src/olympia/shelves/serializers.py
+++ b/src/olympia/shelves/serializers.py
@@ -1,17 +1,22 @@
+from urllib import parse
+
 from rest_framework import serializers
 from rest_framework.reverse import reverse as drf_reverse
 
 from django.conf import settings
 
+from olympia.addons.views import AddonSearchView
 from olympia.shelves.models import Shelf
 
 
 class ShelfSerializer(serializers.ModelSerializer):
     url = serializers.SerializerMethodField()
+    addons = serializers.SerializerMethodField()
 
     class Meta:
         model = Shelf
-        fields = ['title', 'url', 'footer_text', 'footer_pathname']
+        fields = ['title', 'url', 'endpoint', 'criteria', 'footer_text',
+                  'footer_pathname', 'addons']
 
     def get_url(self, obj):
         if obj.endpoint == 'search':
@@ -30,3 +35,14 @@ class ShelfSerializer(serializers.ModelSerializer):
             url = None
 
         return url
+
+    def get_addons(self, obj):
+        if obj.endpoint == 'search':
+            criteria = obj.criteria.strip('?')
+            params = dict(parse.parse_qsl(criteria))
+            request = self.context.get('request', None)
+            request.GET = request.GET.copy()
+            request.GET.update(params)
+            return AddonSearchView(request=request).data
+        else:
+            return None
--- a/src/olympia/shelves/tests/test_serializers.py
+++ b/src/olympia/shelves/tests/test_serializers.py
@@ -6,13 +6,45 @@ from rest_framework.test import APIRequestFactory
 from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
 
-from olympia.addons.views import AddonSearchView
-from olympia.amo.tests import TestCase, reverse_ns
+from olympia import amo
+from olympia.amo.tests import addon_factory, ESTestCase, reverse_ns
+from olympia.constants.promoted import RECOMMENDED
+from olympia.promoted.models import PromotedAddon
 from olympia.shelves.models import Shelf
 from olympia.shelves.serializers import ShelfSerializer
 
 
-class TestShelvesSerializer(TestCase):
+class TestShelvesSerializer(ESTestCase):
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+
+        # Shouldn't be necessary, but just in case.
+        cls.empty_index('default')
+
+        addon_factory(
+            name='test addon test01', type=amo.ADDON_EXTENSION,
+            average_daily_users=46812, weekly_downloads=132, summary=None)
+        addon_factory(
+            name='test addon test02', type=amo.ADDON_STATICTHEME,
+            average_daily_users=18981, weekly_downloads=145, summary=None)
+        addon_ext = addon_factory(
+            name='test addon test03', type=amo.ADDON_EXTENSION,
+            average_daily_users=482, weekly_downloads=506, summary=None)
+        addon_theme = addon_factory(
+            name='test addon test04', type=amo.ADDON_STATICTHEME,
+            average_daily_users=8838, weekly_downloads=358, summary=None)
+
+        PromotedAddon.objects.create(
+            addon=addon_ext, group_id=RECOMMENDED.id
+        ).approve_for_version(version=addon_ext.current_version)
+
+        PromotedAddon.objects.create(
+            addon=addon_theme, group_id=RECOMMENDED.id
+        ).approve_for_version(version=addon_theme.current_version)
+
+        cls.refresh()
+
     def setUp(self):
         self.search_shelf = Shelf.objects.create(
             title='Populâr themes',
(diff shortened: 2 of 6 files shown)


src/olympia/shelves/serializers.py
```diff
@@ -27,3 +31,16 @@ def get_url(self, obj):
...
+    def get_addons(self, obj):
+        if obj.endpoint == 'search':
+            endpoint = drf_reverse('v4:addon-search') + obj.criteria
+        elif obj.endpoint == 'collections':
+            endpoint = drf_reverse('v4:collection-addon-list', kwargs={
+                'user_pk': settings.TASK_USER_ID,
+                'collection_slug': obj.criteria})
+
+        url = settings.INTERNAL_SITE_URL + endpoint
+        response = requests.get(url)
```
eviljeff, line 44: we don't want to use `requests` at all really- we want to call the APIView directly and get the results before it's been turned into a response using `AddonSearchView(request=request).data` - the tricky part is creating a fake request that at least has the correct GET parameters (i.e. the criteria) so it returns the correct data.  (Best case would be cloning the original `self.context.get('request')` and modifying it, somehow, but whatever works)

https://docs.djangoproject.com/en/2.2/ref/request-response/#httprequest-objects is the request object, though it's wrapped by Django Rest Framework in https://www.django-rest-framework.org/api-guide/requests/

xlisachan:
This push adds `class HomepageView(APIView)`, which includes the shelves' data (title, url, footer text/pathname) from `class ShelfViewSet` and will include the shelves' contents.

I am no longer getting errors when `class HomepageView` includes data from `class ShelfViewSet`, but it looks to have failed a Travis check with the error `Reverse for 'shelves-list' not found. 'shelves-list' is not a valid view function or pattern name.` I used `hero/views.py` and `hero/urls.py` as a guide in creating `shelves/views.py` and `shelves/urls.py`, and I don't believe `hero-shelves-list` exists either?

I'm going to continue working through including the shelves' contents from your guidance above. I'll keep you updated. :)

eviljeff:
> it looks to have failed a Travis check with the error `Reverse for 'shelves-list' not found. 'shelves-list' is not a valid view function or pattern name.

you changed the viewset basename in the Router in urls.py to `shelves-info` from `shelves` so all the generated view names the Router adds are updated too.

src/olympia/shelves/views.py
```diff
@@ -1,12 +1,25 @@
...
+        return self.get_serializer(self.queryset, many=True)
+
+
+class HomepageView(APIView):
+    def get(self, request, format=None):
+        output = {
+            'info': ShelfViewSet(
+                request=request).get_shelves().data,
+        }
+        return Response(output)
```
eviljeff, line 25: So, we will probably need this kind of combination view in the future when we want to include _different_ types of object in the response, but at the moment all the results will be `Shelf`s, so it doesn't bring us much closer.

eviljeff: The addons will be per Shelf, so really the best place for them is inside the Shelf, so that's leads to us including the viewset data inside the Serializer, like you were doing before via requests.

eviljeff (commented):
Adding this to `AddonSearchView` (and `CollectionAddonViewSet`, but I'd focus on `AddonSearchView` for this patch because it's more straightforward) should make the ShelfSerializer code a bit more understandable - it just needs to do `AddonSearchview(request=request).data`, then:
```
    @property
    def data(self):
        self.initial(self.request)
        queryset = self.filter_queryset(self.get_queryset())
        serializer = self.get_serializer(queryset, many=True)
        return serializer.data
```

src/olympia/shelves/serializers.py
```diff
@@ -30,3 +35,14 @@ def get_url(self, obj):
...
         return url
+
+    def get_addons(self, obj):
+        if obj.endpoint == 'search':
+            criteria = obj.criteria.strip('?')
+            params = dict(parse.parse_qsl(criteria))
+            request = self.context.get('request', None)
+            request.GET = request.GET.copy()
+            request.GET.update(params)
+            return AddonSearchView(request=request).data
```
xlisachan, line 46: + @bobsilverberg 
I got it to return the add-ons based on `criteria`, but I don't believe I did it correctly. `tests/test_serializers.py` fails with `AttributeError: 'WSGIRequest' object has no attribute 'query_params'`. When `query_params` is set, it goes on to fail with AttributeError: 'WSGIRequest' object has no attribute 'user'`.

eviljeff: Somewhere along the way `request` has been unwrapped - DRF has it's own `Request` [1] object that wraps django's `HttpRequest` (`WSGIRequest`) and adds among other things, `query_params` and `user`.  I can't see the problem immediately so I'll have to have a look around.

[1] https://www.django-rest-framework.org/api-guide/requests/#query_params

eviljeff: Oh, so `APIRequestFactory` that DRF provides doesn't actually return a DRF `Request` - it generates a normal django request.  Functional tests in test_views.py should work (the view creates the DRF `Request`).  We'll probably have to something similar to what `initialize_request` does for the serializer tests - maybe just `self.request = Request(APIRequestFactory().get('/api/%s/' % api_version))` is enough?

[1] https://github.com/encode/django-rest-framework/blob/3.11.1/rest_framework/views.py#L381

xlisachan: The request is currently `<WSGIRequest: GET '/api/v4/'>`, but should the request that goes into `AddonSearchView(request=request).data` look like `<WSGIRequest: GET '/api/v4/search/?sort=users&type=extension'>`?

eviljeff: mmm, good question.  I don't _think_ it matters - iirc the path to the api isn't checked within the view itself, just the version at the start.  But it's something to play around with if it doesn't work as expected (especially in tests)

xlisachan:
`+` @bobsilverberg 
I updated the requests within `test_serializers.py` to have `query_params` and `user` attributes, which seems to fix the errors from before. I set `self.request.user` to `AnonymousUser()`, as I saw it being used in other tests, please let me know if it should be changed.

For `test_views.py`, I broke up `def test_basic` into separate tests just so it's clear that it checks that (1) the view is empty when there are no enabled shelves and (2) only enabled shelves appear in the view.

Please review when you have a moment. Thank you!!

src/olympia/shelves/tests/test_views.py
```diff
@@ -55,14 +53,34 @@ def test_basic(self):
...
+            {
+                'title': 'Recommended extensions',
+                'url': self.search_url,
+                'endpoint': 'search',
+                'criteria': '?promoted=recommended&sort=random&type=extension',
+                'footer_text': 'See more recommended extensions',
+                'footer_pathname': '',
+                'addons': []
+            },
+        ]
```
eviljeff, line 86: Can we have a test that includes at least one addon to test the complete output?  

As `AddonSearchView` calls elasticsearch the `TestCase` will need to change to `ESTestCase` - which is a little different.   Have a look through the tests that use it, but main difference is the addon setup is in `setUpTestData` (a `@classmethod`) and you have to do `cls.refresh()` after all the setup is done to trigger an elasticsearch index update.

src/olympia/shelves/views.py
```diff
@@ -7,6 +7,4 @@
 class ShelfViewSet(viewsets.ModelViewSet):
     queryset = Shelf.objects.filter(
         shelfmanagement__enabled=True).order_by('shelfmanagement__position')
-    permission_classes = []
```
eviljeff, line 10: I think this is needed.

src/olympia/shelves/tests/test_serializers.py
```diff
@@ -43,8 +51,12 @@ def test_shelf_serializer_search(self):
         assert data == {
             'title': 'Populâr themes',
             'url': search_url,
+            'endpoint': self.search_shelf.endpoint,
+            'criteria': self.search_shelf.criteria,
             'footer_text': 'See more populâr themes',
-            'footer_pathname': ''}
+            'footer_pathname': '',
+            'addons': AddonSearchView(request=self.request).data
```
eviljeff, line 58: this is fine if we're testing `AddonSearchView` elsewhere but I don't see we are.  So either have a test for `AddonSearchView` or explicitly show the output here.  (I think it's fine to `mock` `AddonSearchView` and set `.data` if there's a complete test in test_views (see other comment))

eviljeff (changes requested):

xlisachan:
Thanks for the feedback! `views.py` and `test_views.py` have been updated as advised. `test_serializers.py` has been updated similar to `test_views.py` to show some of the output. Please review when you have a moment. Thanks again!!

eviljeff (approved):
r+ lgtm!

eviljeff:
@bobsilverberg do you have any further feedback before I merge?

bobsilverberg:
This looks good to me. Sorry, I haven't been keeping totally up to date with where we are with this feature. Does this only include the `addons` in the API response for `search` criteria? Does it not address collections? If so, do we need a follow-up for that?

Also, should we include updates to the API docs for this, or is that premature at this point?

eviljeff:
> Does this only include the `addons` in the API response for `search` criteria? Does it not address collections? If so, do we need a follow-up for that?
> 
> Also, should we include updates to the API docs for this, or is that premature at this point?

yes, yes, yes & yes* 😄 *ish.  @xlisachan can you file an issue for implementing the collection type and finishing the functionality?  The api documentation could have been started with this patch but could be done when it's finished - i.e. when the collection shelves work.

xlisachan:
I have filed issues to implement the collections type (#15614) and add the API documentation (#15615). Please assign me to both when you have a moment, thank you!

bobsilverberg:
> I have filed issues to implement the collections type (#15614) and add the API documentation (#15615). Please assign me to both when you have a moment, thank you!

Thanks Lisa, and done!

PR #15295, opened 2020-08-14
rm custom addon and user search fields on /admin landing page

### LINKED ISSUE #15292: rm addon and user search fields from /admin landing page
the admin page currently has two obsolete search fields on the landing page.  

Addon search does a simple name query via the Addon manager and returns the results on a page (that then link to the django admin Addon change page).  Alternative is using the search field on the Addon changelist.

User search does an ajax request to get users (the endpoint actually django admin's UserProfile changelist search) and displays an inline div using javascript to show matches.  Clicking on the result appears to go to a broken link to a frontend page so I'm pretty sure it's not used.  Alternative is using the search field on the UserProfile changelist (that returns the same results, but with working links).

We should remove them both.

eviljeff:
fixes #15292


src/olympia/zadmin/admin.py
```diff
@@ -40,6 +44,21 @@ def related_single_content_link(obj, related_field):
         return ''
 
 
+# Hijack the admin's login to use our pages.
```
bobsilverberg, line 47: Just wondering what this change is about. It doesn't seem to be mentioned at https://github.com/mozilla/addons-server/issues/15292.

src/olympia/zadmin/tests/test_views.py
```diff
@@ -141,89 +124,6 @@ def test_django_admin_logout(self):
...
-        assert res.status_code == 302
-
-
-class TestAddonAdmin(TestCase):
-    fixtures = ['base/users', 'base/addon_3615']
-
-    def setUp(self):
-        super(TestAddonAdmin, self).setUp()
-        assert self.client.login(email='admin@mozilla.com')
-        self.url = reverse('admin:addons_addon_changelist')
```
bobsilverberg, line 216: What was this testing? I'm not seeing how it's related to the search features that were removed.

src/olympia/zadmin/admin.py
eviljeff: I just moved the code out of urls.py (it was an odd place to have it - I should have moved it previously)

src/olympia/zadmin/tests/test_views.py
eviljeff: It was a random single test checking that the addon changelist page works, but we have a whole class of them here already https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/tests/test_admin.py

PR #15312, opened 2020-08-17
Fallback to summary for all addon types for PrimaryHeroShelf

### LINKED ISSUE #15134: Allow fallback to Addon.summary for all addon types for PrimaryHeroShelf
We currently only fall back to the addon.summary for extensions, so in the absence of a defined description for the primary hero shelf, it returns `""`.  It doesn't really make sense any longer so we'll remove that condition.

eviljeff:
fixes #15134

### DIFF
--- a/src/olympia/hero/serializers.py
+++ b/src/olympia/hero/serializers.py
@@ -2,7 +2,6 @@ from django.utils.translation import ugettext
 
 from rest_framework import serializers
 
-from olympia import amo
 from olympia.addons.models import Addon
 from olympia.addons.serializers import AddonSerializer
 from olympia.amo.templatetags.jinja_helpers import absolutify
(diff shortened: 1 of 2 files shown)


src/olympia/hero/serializers.py
```diff
@@ -40,10 +39,7 @@ def get_description(self, obj):
         elif obj.description:
             return ugettext(obj.description)
         else:
-            addon = obj.promoted_addon.addon
-            return (str(addon.summary)
-                    if addon.summary and addon.type == amo.ADDON_EXTENSION
-                    else '')
+            return str(obj.promoted_addon.addon.summary or '')
```
muffinresearch, line 42: Presumably this entire chain of props always exists?

eviljeff: yes - both PrimaryHero.promoted_addon, and PromotedAddon.addon are defined with `null=False` so at the database level they must exist.

PR #15324, opened 2020-08-18
promoted addon search api filter

### LINKED ISSUE #15272: Add `promoted` filter to addons api
the addons api can filter by recommended status - we need to expand that to allow addon searches to be filtered by any promoted class.

eviljeff:
fixes #15272 - possibly one for a post-PTO re-review from @diox because it's ES queries (not particularly complicated ES queries though) and API additions.

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -26,7 +26,7 @@ from olympia.amo.tests import (
 from olympia.amo.urlresolvers import get_outgoing_url, reverse
 from olympia.bandwagon.models import CollectionAddon
 from olympia.constants.categories import CATEGORIES, CATEGORIES_BY_ID
-from olympia.constants.promoted import RECOMMENDED
+from olympia.constants.promoted import LINE, RECOMMENDED
 from olympia.discovery.models import DiscoveryItem
 from olympia.users.models import UserProfile
 from olympia.versions.models import ApplicationsVersions, AppVersion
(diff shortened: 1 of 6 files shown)


src/olympia/addons/tests/test_views.py
```diff
@@ -1236,6 +1236,61 @@ def test_filter_by_recommended(self):
...
+        # That will filter out for a different app
+        data = self.perform_search(
+            self.url, {'promoted': 'recommended', 'app': 'android'})
+        assert data['count'] == 1
+        assert len(data['results']) == 1
+        assert data['results'][0]['id'] == addon.pk
+        # addon2 was for Firefox only
+
+        # test with other other promotions
+        self.make_addon_promoted(addon, LINE, approve_version=True)
```
muffinresearch, line 1286: Should these tests cover more than `RECOMMENDED` and `LINE`?

src/olympia/addons/tests/test_views.py
```diff
@@ -1236,6 +1236,61 @@ def test_filter_by_recommended(self):
...
+
+        data = self.perform_search(
+            self.url, {'promoted': 'recommended'})
+        assert data['count'] == 2
+        assert len(data['results']) == 2
+        assert {res['id'] for res in data['results']} == {addon.pk, addon2.pk}
+
+        # And with app filtering too
+        data = self.perform_search(
+            self.url, {'promoted': 'recommended', 'app': 'firefox'})
```
muffinresearch, line 1272: Should this use constants instead of strings?

eviljeff: we tend to just write the strings in the api tests (if they're human understandable strings) so things would loudly fail if an api value was inadvertently changed.

src/olympia/addons/tests/test_views.py
eviljeff: It's just laziness to test a couple of the groups.  Could loop through them all, sure.

eviljeff: https://github.com/mozilla/addons-server/pull/15324/commits/8c7e0ec6f33070f0e223c236970aa333abc60552

src/olympia/search/tests/test_filters.py
```diff
@@ -907,6 +908,24 @@ def test_search_by_recommended(self):
...
+    def test_search_by_promoted(self):
+        qs = self._filter(data={'promoted': 'recommended'})
+        filter_ = qs['query']['bool']['filter']
+        assert {'term': {'promoted.group_id': RECOMMENDED.id}} in filter_
+
+        with self.assertRaises(serializers.ValidationError) as context:
+            self._filter(data={'promoted': 'foo'})
+        assert context.exception.detail == ['Invalid "promoted" parameter.']
+
+        qs = self._filter(data={'promoted': 'verified', 'app': 'firefox'})
```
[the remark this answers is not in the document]
eviljeff: https://github.com/mozilla/addons-server/pull/15324/commits/8c7e0ec6f33070f0e223c236970aa333abc60552

eviljeff:
I also slightly rewrote the filter because when I added the loop over all the groups I realised we would regard `not_promoted` as a valid group and it wouldn't work (without special handling... which _could_ be added but would be a follow-up if there's agreement it's actually useful)

src/olympia/search/filters.py
```diff
@@ -335,6 +336,32 @@ class AddonRecommendedQueryParam(AddonQueryParam):
...
+
+    def get_es_query(self):
+        query = [Q(
+            self.operator,
+            **{'promoted.group_id': self.get_value()})]
+
+        if app := self.get_app():
+            query.append(
+                Q(self.operator, **{'promoted.application_id': app}) |
+                ~Q('exists', field='promoted.application_id'))
```
diox, line 360: I assume this is here for a good reason, but why ?

eviljeff: When application_id is `None` it seems to just not get stored (unless I'm missing something)

diox: Ah, yes. Deserves a comment I think.

eviljeff: https://github.com/mozilla/addons-server/commit/23a5ef9c6a2bf61a6714df48ab406d13666233b7

PR #15331, opened 2020-08-19
migration to set current Recommended promoted addons to desktop Firefox

### LINKED ISSUE #15313: set all the existing recommended addons to be desktop Firefox only (?)
The previous recommendation implementation was application agnostic on addons-server, but assumed to be desktop Firefox specific on addons-frontend.  When recommendation add-ons were transitioned to the promoted addons models they were similarly set without a specific application.  I believe all the existing addons were assessed for desktop only so we should update them to be just for desktop Firefox so frontend can fully support promoted add-ons for different applications.

@jvillalobos @devaneymoz can you confirm if all the existing recommended add-ons are indeed only supposed to be recommended for desktop Firefox?  If so we can run a db migration to fix them (if it's a partial thing you might have to do it manually)

eviljeff:
fixes #15313 

seems to work locally.


muffinresearch (approved):
Looks sane, though this is another one that's worth getting a quick check from Mat when he's back.

PR #15345, opened 2020-08-21
merge Addons:PostReview permission into Addons:Review

### LINKED ISSUE #14782: Merge `Addons:Review` and `Addons:PostReview` permissions together
A significant source of complexity in reviewer tools is the different kind of actions a reviewer can perform depending on whether they have `Addons:PostReview` and `Addons:Review`. The former allows only confirming auto-approval and multiple rejections of auto-approved add-ons, the latter allows for manual approval/rejection of add-ons that haven't been auto-approved.

Do we still need that distinction ? I'm convinced there are subtle bugs lying around depending on the add-on state limiting what actions are available to users that only have one of the two in some situations. We could simplify the code around permissions quite a bit if it was just a single permission.

cc @jvillalobos @wagnerand for input.

eviljeff:
fixes #14782

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -346,7 +346,7 @@ class ReviewHelper(object):
         # Default permissions / admin needed values if it's just a regular
         # code review, nothing fancy.
         permission = amo.permissions.ADDONS_REVIEW
-        permission_post_review = amo.permissions.ADDONS_POST_REVIEW
+        permission_post_review = amo.permissions.ADDONS_REVIEW
         is_admin_needed = (
             self.addon.needs_admin_content_review or
             self.addon.needs_admin_code_review)
--- a/src/olympia/access/acl.py
+++ b/src/olympia/access/acl.py
@@ -130,7 +130,6 @@ def check_addon_ownership(request, addon, dev=False, admin=True,
 def check_addons_reviewer(request, allow_content_reviewers=True):
     permissions = [
         amo.permissions.ADDONS_REVIEW,
-        amo.permissions.ADDONS_POST_REVIEW,
         amo.permissions.ADDONS_RECOMMENDED_REVIEW
     ]
     if allow_content_reviewers:
@@ -177,7 +176,6 @@ def is_user_any_kind_of_reviewer(user, allow_viewers=False):
         amo.permissions.ADDONS_REVIEW,
         amo.permissions.ADDONS_REVIEW_UNLISTED,
         amo.permissions.ADDONS_CONTENT_REVIEW,
-        amo.permissions.ADDONS_POST_REVIEW,
         amo.permissions.ADDONS_RECOMMENDED_REVIEW,
         amo.permissions.STATIC_THEMES_REVIEW,
     ]
--- a/src/olympia/access/fixtures/initial.json
+++ b/src/olympia/access/fixtures/initial.json
@@ -40,16 +40,6 @@
     }
   },
   {
-    "pk": 7,
-    "model": "access.group",
-    "fields": {
-        "name": "Reviewers: Add-ons",
-        "rules": "Addons:PostReview",
-        "created": "2017-10-26 15:00:00",
-        "modified": "2017-10-26 15:00:00"
-    }
-  },
-  {
     "pk": 8,
     "model": "access.group",
     "fields": {
--- a/src/olympia/access/tests.py
+++ b/src/olympia/access/tests.py
@@ -229,19 +229,6 @@ class TestCheckReviewer(TestCase):
         assert check_static_theme_reviewer(request)
         assert is_user_any_kind_of_reviewer(request.user)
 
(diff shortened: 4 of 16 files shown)


src/olympia/reviewers/utils.py
```diff
@@ -346,7 +346,7 @@ def get_actions(self, request):
         # Default permissions / admin needed values if it's just a regular
         # code review, nothing fancy.
         permission = amo.permissions.ADDONS_REVIEW
-        permission_post_review = amo.permissions.ADDONS_POST_REVIEW
+        permission_post_review = amo.permissions.ADDONS_REVIEW
```
diox, line 349: This can be simplified further, the whole `permission_post_review` variable is now redundant, as well as `is_appropriate_reviewer_post_review` below.

There are also at least one comment mentioning `Addons:PostReview` that should be reworded as it no longer makes sense now that the 2 permissions are merged together.

eviljeff: I tried that initially but the differentiation between the two variables seemed to be important later because it broke the logic (it's not easy to follow the logic through 250 lines of code in a single function)

eviljeff: I just tried this again - it's not just removing `permission_post_review`, but `is_admin_needed_post_review` and `is_appropriate_reviewer_post_review` too and doing that results in the wrong actions being available when a version is pending rejection.

```diff
------------------------ src/olympia/reviewers/utils.py ------------------------
index 26066d7bf8..20f813e250 100644
@@ -346,43 +346,37 @@ class ReviewHelper(object):
         # Default permissions / admin needed values if it's just a regular
         # code review, nothing fancy.
         permission = amo.permissions.ADDONS_REVIEW
-        permission_post_review = amo.permissions.ADDONS_REVIEW
         is_admin_needed = (
             self.addon.needs_admin_content_review or
             self.addon.needs_admin_code_review)
-        is_admin_needed_post_review = is_admin_needed
 
         # More complex/specific cases.
         if promoted_group == RECOMMENDED:
             permission = amo.permissions.ADDONS_RECOMMENDED_REVIEW
-            permission_post_review = permission
         elif version_is_unlisted:
             is_admin_needed = self.addon.needs_admin_code_review
             permission = amo.permissions.ADDONS_REVIEW_UNLISTED
-            permission_post_review = permission
         elif promoted_group.admin_review:
-            is_admin_needed = is_admin_needed_post_review = True
+            is_admin_needed = True
         elif self.content_review:
             is_admin_needed = self.addon.needs_admin_content_review
             permission = amo.permissions.ADDONS_CONTENT_REVIEW
         elif self.addon.type == amo.ADDON_STATICTHEME:
             is_admin_needed = self.addon.needs_admin_theme_review
             permission = amo.permissions.STATIC_THEMES_REVIEW
-            permission_post_review = permission
 
         # In addition, if the latest (or current for post-review) version is
         # pending rejection, an admin is needed.
-        if self.version and self.version.pending_rejection:
+        if (
+            self.version and self.version.pending_rejection) or (
+                self.addon.current_version and
+                self.addon.current_version.pending_rejection
+        ):
             is_admin_needed = True
-        if (self.addon.current_version and
-                self.addon.current_version.pending_rejection):
-            is_admin_needed_post_review = True
 
         # Whatever permission values we set, we override if an admin is needed.
         if is_admin_needed:
             permission = amo.permissions.REVIEWS_ADMIN
-        if is_admin_needed_post_review:
-            permission_post_review = amo.permissions.REVIEWS_ADMIN
 
         # Is the current user a reviewer for this kind of add-on ?
         is_reviewer = acl.is_reviewer(request, self.addon)
@@ -392,8 +386,6 @@ class ReviewHelper(object):
         # impactful actions).
         is_appropriate_reviewer = acl.action_allowed_user(
             request.user, permission)
-        is_appropriate_reviewer_post_review = acl.action_allowed_user(
-            request.user, permission_post_review)
 
         addon_is_complete = self.addon.status not in (
             amo.STATUS_NULL, amo.STATUS_DELETED)
@@ -434,8 +426,7 @@ class ReviewHelper(object):
             # been auto-approved, provided that the add-on isn't marked as
             # needing admin review.
             can_reject_multiple = addon_is_valid_and_version_is_listed and (
-                is_appropriate_reviewer or
-                is_appropriate_reviewer_post_review
+                is_appropriate_reviewer
             )
 
         # Definitions for all actions.
@@ -498,7 +489,7 @@ class ReviewHelper(object):
                 not self.content_review and
                 addon_is_valid_and_version_is_listed and
                 current_version_is_listed_and_auto_approved and
-                is_appropriate_reviewer_post_review
+                is_appropriate_reviewer
             ),
         }
         actions['reject_multiple_versions'] = {

```

diox (approved):
As discussed on Matrix, it would have been nice to be able to refactor the permission handling in `reviewer/utils.py` but it's tricky as we still have some conditions that are specific to post-review actions. Let's leave that aside for now and merge what we have.

PR #15351, opened 2020-08-24
Catch form criteria edge cases

### LINKED ISSUE #15349: The search endpoint allows saving a shelf when hash or slash are used for criteria
@xlisachan 
Followup for https://github.com/mozilla/addons-server/issues/14952

1. Go to the shelves module to "+add shelf module"
2. Give, a title, select the `search` endpoint 
3. Add "#" or "/" into the Criteria field

Expected result:
Check criteria - No data found

Actual result:
The shelf is saved

Notes:
This is reproducible with: "#?recommended=true&sort=random&type=extension" or "/?recommended=true&sort=random&type=extension" also with multiple "#####" or "/////////////" and combinations of "#/" or "#/text".
I could not reproduce it with the collections endpoint.
I could not reproduce this issue using other special chars.???????????//

### LINKED ISSUE #15350: Two dots used in the criteria field with the search endpoint produce a server error
@xlisachan 
Followup for https://github.com/mozilla/addons-server/issues/14952

STR:
1. Go to the new shelf model 
2. Add a new shelf, select the search endpoint and use for Criteria two dots before a valid criteria "..?recommended=true&sort=random&type=extension"or simply two dots in this field ".."

result:
internal server error - 500

Notes:
I could not reproduce a server error for the collection endpoint with this scenario. Also I tested other special char combinations and I did not reproduce the error. - FF79(Win10).

xlisachan:
Fixes #15349 
Fixes #15350 

This pull request updates `shelves/forms.py` to check that the search criteria field starts with a `?` and cannot include multiple `?`.

The following tests have been added to `test_forms.py`:
- `def test_clean_search_criteria_does_not_start_with_qmark`
- `def test_clean_search_criteria_has_multiple_qmark`

`test_clean_search_returns_404`  has been removed since the criteria example used to return the 404 started without a `?` and therefore is now caught by the conditional that it needs to start with `?`. I could not figure out a combo that would return a 404 for search.

In addition, I am not sure how to prevent the admin user from entering in a criteria, such as `?#####`, `?//////`, `?hi`, which allows a shelf module to be saved as it returns valid results. 

### DIFF
--- a/src/olympia/shelves/forms.py
+++ b/src/olympia/shelves/forms.py
@@ -27,8 +27,11 @@ class ShelfForm(forms.ModelForm):
 
         try:
             if endpoint == 'search':
-                api = drf_reverse('v4:addon-search')
-                url = baseUrl + api + criteria
+                if not criteria.startswith('?') or criteria.count('?') > 1:
+                    raise forms.ValidationError('Check criteria field.')
+                else:
+                    api = drf_reverse('v4:addon-search')
+                    url = baseUrl + api + criteria
             elif endpoint == 'collections':
                 api = drf_reverse('v4:collection-addon-list', kwargs={
                     'user_pk': settings.TASK_USER_ID,
--- a/src/olympia/shelves/tests/test_forms.py
+++ b/src/olympia/shelves/tests/test_forms.py
@@ -93,6 +93,28 @@ class TestShelfForm(TestCase):
         assert not form.is_valid()
         assert form.errors == {'criteria': ['This field is required.']}
 
+    def test_clean_search_criteria_does_not_start_with_qmark(self):
+        form = ShelfForm({
+            'title': 'Recommended extensions',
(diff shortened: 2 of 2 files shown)


eviljeff (approved):
> In addition, I am not sure how to prevent the admin user from entering in a criteria, such as ?#####, ?//////, ?hi, which allows a shelf module to be saved as it returns valid results.

I really wouldn't bother about it.  The tool isn't user/developer facing - it's only going to be used by Scott and potentially another admin so catching stupid mistakes is enough - he's not likely to enter garbage inputs.

xlisachan:
Thanks for the feedback! Just a heads-up, I no longer have access to merge (or request reviewers). 😛

eviljeff:
Because you've been removed from the MoCo org on github you'll need to comment in the issues before I can assign them to you too.

PR #15352, opened 2020-08-24
Block change page can be loaded with guid in the url rather than Block pk

### LINKED ISSUE #15344: Enable admin block view access for an add-on using its GUID
Similar to django admin views for add-ons instances, admins need to be able to access an add-on block given an add-on GUID.

There is a workaround to go through the reviewer tools page, but for convenience, important and often used pages should have a way to directly access them using the official identifier, which is the add-on GUID.

eviljeff:
fixes #15344

### DIFF
--- a/src/olympia/blocklist/admin.py
+++ b/src/olympia/blocklist/admin.py
@@ -1,3 +1,4 @@
+from django import http
 from django.contrib import admin, auth, contenttypes, messages
 from django.core.exceptions import PermissionDenied
 from django.forms.fields import ChoiceField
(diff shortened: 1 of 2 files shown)


src/olympia/blocklist/admin.py
```diff
@@ -696,7 +697,17 @@ def formfield_for_dbfield(self, db_field, request, **kwargs):
...
-        obj = self.get_object(request, obj_id) if obj_id else None
+        if obj_id:
+            obj = (
+                self.get_object(request, obj_id) or
+                # if we can't find the obj_id maybe it's a guid instead
+                self.get_object(request, obj_id, 'guid'))
+            if obj and str(obj_id) != str(obj.id):
+                # we found it from the guid if the obj_id != obj.id so redirect
+                url = request.path.replace(obj_id, str(obj.id), 1)
+                return http.HttpResponsePermanentRedirect(url)
```
wagnerand, line 708: Could we avoid that redirect if we just set

```
obj_id = obj.id
```

here?

eviljeff: we could, but we want the redirect so there are no weird edge cases.

wagnerand (approved):

PR #15364, opened 2020-08-26
sign promoted addons for specific applications, where available

### LINKED ISSUE #15315: support autograph signing for promoted for specific applications
#14863 added support for signing other non-recommend promoted groups, but without considering applications.  So the possible states are `['recommended', 'verified', 'sponsored', 'line']`.  

After talking with @Rob--W and @mixedpuppy we're going to interpret `['recommended', 'verified', 'sponsored', 'line']` as being just for desktop Firefox, and add `['recommended-android', 'verified-android', 'sponsored-android', 'line-android']` to support Fenix.

eviljeff:
fixes #15315

### DIFF
--- a/scripts/autograph_travis_test_config.yaml
+++ b/scripts/autograph_travis_test_config.yaml
@@ -131,7 +131,7 @@ signers:
         path: "mozilla-recommendation.json"
         states:
           recommended: true
-          sponsored: true
+          recommended-android: true
           verified: true
           line: true
         relative_start: 0h
--- a/src/olympia/lib/crypto/tests/test_signing.py
+++ b/src/olympia/lib/crypto/tests/test_signing.py
@@ -339,14 +339,7 @@ class TestSigning(TestCase):
             'Digest-Algorithms: SHA1 SHA256\n'
         )
 
-    @override_switch('autograph_promoted_signer', active=False)
-    def test_call_signing_promoted_recommended_only(self):
-        # This is the usual process for recommended add-ons, they're
-        # in "pending recommendation" and only *after* we approve and sign
-        # them they will become "recommended". If their promoted group changes
-        # we won't sign further versions as recommended.
-        self.make_addon_promoted(self.file_.version.addon, RECOMMENDED)
-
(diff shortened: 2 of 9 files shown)


scripts/autograph_travis_test_config.yaml
```diff
@@ -131,7 +131,7 @@ signers:
         path: "mozilla-recommendation.json"
```
muffinresearch, line 131: Not part of this change, but is the right path or just used for tests? Autograph has `recommendation.json` in their docs https://github.com/mozilla-services/autograph/blob/515ce3638d1b8a8cab1543baf6f33f2223c12831/signer/xpi/README.rst#id1

src/olympia/lib/crypto/tests/test_signing.py
```diff
@@ -364,7 +357,19 @@ def test_call_signing_promoted_recommended_only(self):
         recommendation_data = _get_recommendation_data(
             self.file_.current_file_path)
         assert recommendation_data['addon_id'] == 'xxxxx'
-        assert recommendation_data['states'] == ['recommended']
+        assert sorted(recommendation_data['states']) == states
```
muffinresearch, line 360: Could this use sets so order doesn't matter, or should `states` be sorted too?

muffinresearch (approved):
r+wc

scripts/autograph_travis_test_config.yaml
eviljeff: looks like their readme is wrong
https://github.com/mozilla-services/autograph/blob/master/autograph.yaml#L987

(and yeah, it's just for tests and local docker dev)

src/olympia/lib/crypto/tests/test_signing.py
eviljeff: Casting the JSON into a set would lose the case where a state was included twice in the list, and we don't want that.  States could be sorted too, though it's a kwarg we pass in inside the tests so it's just a matter of remembering to define the kwarg value in alphabetical order or the test fails (we only pass a list of more than 1 item in one place at the moment).

Issue #15271, opened 2020-08-12
drop `is_recommended` from addon api; shim in v4.


eviljeff:
We don't need `is_recommended` any longer - `is_promoted` has the same information as recommended addons are now a subset of promoted addons.  Let's remove it from the api and add a shim for the property in api/v4 (it could exist in v3 too, but it was only added after v4 was launched so is undocumented in v3).  Or maybe just drop entirely once addons-frontend has switched over?

Dropping the data from the ES index depends on us replicating the the `recommended` filter and sort, or at least rewriting them to rely on promoted data in the ES index instead.

--edited after https://github.com/mozilla/addons-server/issues/15368 dropped the model property

eviljeff:
blocked by #15272 and #15274 that rely on `is_recommended` in the ES index.

eviljeff:
@bobsilverberg https://github.com/mozilla/addons-frontend/issues/9579 removed references to `is_recommended` in the API, right?  So we're okay to drop the property from the api.

bobsilverberg:
> @bobsilverberg [mozilla/addons-frontend#9579](https://github.com/mozilla/addons-frontend/issues/9579) removed references to `is_recommended` in the API, right? So we're okay to drop the property from the api.

Actually, no. As I was working on the next patch I realized it’s still being used by the code that calls the autocomplete API.

eviljeff:
@bobsilverberg issue link?

bobsilverberg:
> @bobsilverberg issue link?

Oops, I lied. That's what I get for replying to a GitHub issue first thing in the morning on my day off. Even the code for the autocomplete API has been moved away from `is_recommended`. We no longer rely on `is_recommended` in the API response. Sorry for the confusion.

ioanarusiczki:
@eviljeff 'is_recommended` property which had boolean true/false values, is no longer present in the [Addon API](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#detail). I verified for an addon, lang pack, dictionary and static theme submitted on AMO dev. 

An example:
https://addons-dev.allizom.org/api/v3/addons/addon/495924/
https://addons-dev.allizom.org/api/v4/addons/addon/495924/
https://addons-dev.allizom.org/api/v5/addons/addon/495924/

I've noticed `is_featured` is still present in the API but maybe there's a reason for that.

PR #15369, opened 2020-08-27
drop Addon.is_recommended property

### LINKED ISSUE #15368: drop Addon.is_recommended
it's unnecessary now we have Addon.promoted_groups (and Addon.promoted cached_property for ES)

eviljeff:
fixes #15368

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -1292,21 +1292,12 @@ class Addon(OnChangeMixin, ModelBase):
                     collection_id=settings.COLLECTION_FEATURED_THEMES_ID,
                     addon=self).exists())
 
-    @cached_property
-    def is_recommended(self):
-        return (bool(self.promoted_group(group=RECOMMENDED)) or
-                self._is_recommended_theme())
-
-    def promoted_group(self, *, group=None, application=None,
-                       currently_approved=True):
-        """Is the addon currently promoted for the specified group and
-        application?
+    def promoted_group(self, *, application=None, currently_approved=True):
+        """Is the addon currently promoted for the specified application?
 
         Returns the group constant, or NOT_PROMOTED (which is falsey)
         otherwise.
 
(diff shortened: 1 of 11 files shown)


src/olympia/addons/models.py
```diff
@@ -1292,21 +1292,12 @@ def _is_recommended_theme(self):
...
-    @cached_property
-    def is_recommended(self):
-        return (bool(self.promoted_group(group=RECOMMENDED)) or
-                self._is_recommended_theme())
-
-    def promoted_group(self, *, group=None, application=None,
-                       currently_approved=True):
-        """Is the addon currently promoted for the specified group and
-        application?
+    def promoted_group(self, *, application=None, currently_approved=True):
```
eviljeff, line 1295: I also dropped `group` from this method - I realised there were two ways of doing the same thing, `if addon.promoted_group(group=FOO)` and `if addon.promoted_group() == FOO` which seemed redundant (and the group parameter was only used in tests anyway)

muffinresearch:
Does the frontend need to been updated to account for this change first?

muffinresearch (approved):
r+ Looks sane to me, but I think it'll be good add this to @diox's post landing review queue, in-case I've missed something that needs more solid domain knowledge.

src/olympia/addons/serializers.py
```diff
@@ -644,6 +653,11 @@ def fake_object(self, data):
             for data_author in data_authors
         ]
 
+        if 'is_recommended' in data:
+            # if we have this property in the index, store it so we can expose
+            # it in get_is_recommended.
+            obj._is_recommended = data['is_recommended']
```
diox, line 659: Do we have a follow-up issue to remove this (and the code using it above) later ?

eviljeff: https://github.com/mozilla/addons-server/issues/15271

PR #15386, opened 2020-08-31
Remove obsolete PyOpenSSL dependencies

### LINKED ISSUE #15370: evaluate drop/replace use of pyOpenSSL/cryptography
We don't use https://github.com/pyca/cryptography directly, apart from in a test (just to check the version)

It's a dependency of https://pypi.org/project/pyOpenSSL/ ... that we don't seem to use directly anywhere either.  (There's a reference in https://github.com/mozilla/addons-server/blob/master/src/olympia/core/apps.py#L30 but it's loading a package from `urllib3`.  urllib3, however, only mentions having to use cryptography in reference to python2 https://urllib3.readthedocs.io/en/latest/user-guide.html#certificate-verification-in-python-2 so... maybe we don't need it any more in python3?

Another oddity is the warning at the top of the pyOpenSSL project saying " Note: The Python Cryptographic Authority strongly suggests the use of pyca/cryptography where possible. If you are using pyOpenSSL for anything other than making a TLS connection you should move to cryptography and drop your pyOpenSSL dependency."

diox:
Fix #15370

### DIFF
--- a/requirements/system.txt
+++ b/requirements/system.txt
@@ -1,6 +1,6 @@
 ## Requirements needed to be installed *before* any other requirements.
 ## This makes sure we have the latest version of pip, setuptools and relevant
-## OpenSSL packages to ensure our actual downloads are as safe as possible.
+## system packages first.
 
 # appdirs is required by setuptools
 appdirs==1.4.4 \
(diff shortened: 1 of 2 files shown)


eviljeff (approved):
we still need pyOpenSSL itself then?

diox:
Uhu no, I had it removed originally but looks like I messed up the merge. I'll fix.

PR #15402, opened 2020-09-02
RE: Build an API endpoint for shelves

### LINKED ISSUE #14873: Build an API endpoint for shelves

xlisachan:
Fixes #14873 

This pull request updates `def get_url`. An `else` statement was added to prevent the error of referencing the `url` variable before assignment, in cases that `obj.endpoint` does not match either `search` or `collections`.


eviljeff (approved):
Should really have a test that proves its fixed if the same bad data happened again...

eviljeff:
But it's a small fix so meh

PR #15435, opened 2020-09-08
disable affected versions on Block creation

### LINKED ISSUE #13124: A listed version block does not set a Disabled status for the blocked version
1. Add a new block for an add-on with listed versions - see [this example](https://addons-internal.dev.mozaws.net/en-US/admin/models/blocklist/block/889/change/)
2. Add a min - max version range that will also block at least one listed version
3. Verify the listed version status - in Rev Tools or DevHub - after the block is saved

### What happened?
The listed version is Blocked but not Disabled - see [here](https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/view-page-archive-cache)
The listed version should be disabled automatically when a block is submitted 

- unlisted versions are automatically disabled when a block is submitted 
- reproduced on -dev and stage

listed version with status 'Approved'  after a block was submitted
image

eviljeff:
fixes #13124

### DIFF
--- a/src/olympia/blocklist/templates/admin/blocklist/includes/logs.html
+++ b/src/olympia/blocklist/templates/admin/blocklist/includes/logs.html
@@ -21,7 +21,7 @@
         </ul>
       </li>
     {% endfor %}
-    {% if blocklistsubmission %}
+    {% if blocklistsubmission|default_if_none:False %}
       <li>
         {% url 'admin:blocklist_blocklistsubmission_change' blocklistsubmission.id as change_url%}
         {{ blocklistsubmission.modified|date:"Y-m-d" }}.
--- a/src/olympia/blocklist/utils.py
+++ b/src/olympia/blocklist/utils.py
@@ -193,43 +193,29 @@ def split_regex_to_list(guid_re):
     return GUID_SPLIT.split(trimmed)
 
 
-def disable_addon_for_block(block, timestamp):
+def disable_addon_for_block(block):
     """Disable appropriate addon versions that are affected by the Block, and
     the addon too if 0 - *."""
     from .models import Block
+    from olympia.reviewers.utils import ReviewFiles
 
-    for version in block.addon_versions:
-        if not block.is_version_blocked(version.version):
-            continue
-        if version.addon != block.addon:
-            # We don't need to reject versions from older deleted instances
-            continue
+    review = ReviewFiles(
+        request=None,
+        addon=block.addon,
+        version=None,
+        review_type='pending',
+        user=block.updated_by)
+    review.set_data({'versions': [
+        ver for ver in block.addon_versions
+        # We don't need to reject versions from older deleted instances
+        if ver.addon == block.addon and block.is_version_blocked(ver.version)]
+    })
+    review.reject_multiple_versions()
+
+    for version in review.data['versions']:
         # Clear needs_human_review on rejected versions, we consider that
         # the admin looked at them before blocking.
-        if version.needs_human_review:
-            version.update(needs_human_review=False)
-        if version.needs_human_review_by_mad:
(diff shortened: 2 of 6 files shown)


src/olympia/blocklist/templates/admin/blocklist/includes/logs.html
```diff
@@ -21,7 +21,7 @@
         </ul>
       </li>
     {% endfor %}
-    {% if blocklistsubmission %}
+    {% if blocklistsubmission|default_if_none:False %}
```
eviljeff, line 24: driveby to attempt to stop complaints about undefined variables in django templates

src/olympia/blocklist/utils.py
```diff
@@ -193,12 +193,55 @@ def split_regex_to_list(guid_re):
     return GUID_SPLIT.split(trimmed)
 
 
+def disable_addon_for_block(block, timestamp):
+    """Disable appropriate addon versions that are affected by the Block, and
+    the addon too if 0 - *."""
+    from .models import Block
```
diox, line 199: Preliminary review: Not a big fan of the duplication of code this causes. You have to copy/paste a bunch of things the `ReviewUnlisted` class did, and in a completely different app too. Also, this is missing the python logging that ReviewUnlisted` did.

I wonder if it wouldn't be better to instantiate the correct review object (can this happen for listed add-ons?) and add a method there that would do what we need - keeping the reviewer specific stuff in the reviewer object, re-using the existing code to clear flags, set status, log activity etc.

eviljeff: I'm not sure there's a review action we can re-use directly but I'll look into it.  (my first iteration called the helper methods on ReviewBase directly but circular import error and having to make too many changes led me to create a specialised version here).

And yes, it works on listed versions the same as unlisted versions (no such thing as listed add-ons 😉) - the blocklist is channel agnostic.

eviljeff: https://github.com/mozilla/addons-server/pull/15435/commits/3a6f29517ba1e6a1ae4dc20aea63f8d53b6976eb

src/olympia/blocklist/utils.py
```diff
@@ -193,6 +193,34 @@ def split_regex_to_list(guid_re):
     return GUID_SPLIT.split(trimmed)
 
 
+def disable_addon_for_block(block):
+    """Disable appropriate addon versions that are affected by the Block, and
+    the addon too if 0 - *."""
+    from .models import Block
+    from olympia.reviewers.utils import ReviewFiles
+
+    review = ReviewFiles(
```
diox, line 202: At the moment it doesn't hurt to use `ReviewFiles`, but this is kinda dangerous. The rest of addons-server uses the `ReviewHelper` wrapper to get the right class, passing `version` parameter (and not just `None`).

diox (approved):

src/olympia/blocklist/utils.py
eviljeff: I looked at ReviewHelper but all it'd be used for is `set_review_handler` so it didn't seem much point.   The difficulty with passing in a `version` is knowing which version to provide - and `reject_multiple_versions` doesn't use `self.version` anyway for non-`human_review`.
Could if you want though.

diox: My worry is that if for instance we start overriding `reject_multiple_versions` in `ReviewUnlisted` for some reason, or maybe start having a `channel` property in the review classes to implement channel-specific behavior, this wouldn't get it. So I think this is worth fixing, but I agree it might be difficult. This could be a follow-up.

eviljeff: https://github.com/mozilla/addons-server/pull/15435/commits/8136660bc6349399f90d97c04ccab358125c8c76

PR #15436, opened 2020-09-08
drop is_recommended from api addon detail

### LINKED ISSUE #15271: drop `is_recommended` from addon api; shim in v4.

eviljeff:
fixes #15271

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -324,7 +324,6 @@ class AddonSerializer(serializers.ModelSerializer):
     icons = serializers.SerializerMethodField()
     is_source_public = serializers.SerializerMethodField()
     is_featured = serializers.SerializerMethodField()
-    is_recommended = serializers.SerializerMethodField()
     name = TranslationSerializerField()
     previews = PreviewSerializer(many=True, source='current_previews')
     promoted = PromotedAddonSerializer()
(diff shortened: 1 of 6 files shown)


src/olympia/addons/serializers.py
```diff
@@ -775,9 +758,10 @@ class AddonBasketSyncSerializer(AddonSerializerWithUnlistedData):
     # We want to send all authors to basket, not just listed ones, and have
     # the full basket-specific serialization.
     authors = UserProfileBasketSyncSerializer(many=True)
-    name = serializers.SerializerMethodField()
-    latest_unlisted_version = VersionBasketSerializer()
     current_version = VersionBasketSerializer()
+    is_recommended = serializers.SerializerMethodField()
```
bobsilverberg, line 762: Why does this need to be added now?

bobsilverberg (approved):
Looks good, I just have one question.

r+wc

src/olympia/addons/serializers.py
eviljeff: it's been removed above from super class `AddonSerializer`

Issue #15440, opened 2020-09-09
Use weekly downloads in stats overview


willdurand:
See: https://discourse.mozilla.org/t/downloads-statistics/66949

This change was planned but we had to stop the execution of the cron task that updates total downloads to fix https://github.com/mozilla/addons-server/issues/15135 last month, so let's hide total downloads.

ioanarusiczki:
@willdurand From the [Bigquery samples](https://github.com/mozilla/bigquery-etl/blob/3cefa96e15cebbded340633bed240db36bb8cff5/sql/amo_dev/amo_stats_dau_v2/query.sql) I verified, all show in the [Overview ](https://addons-dev.allizom.org/en-US/firefox/addon/firefox-color/statistics/?last=90)below the chart a blue link with "Weekly Downloads" but there is no daily download data for them in the last 30 days. There is some data for the last 90 days.  

![weekly](https://user-images.githubusercontent.com/33448286/92714414-ddb30e80-f364-11ea-80bb-1f0fdd8d6d49.png)

willdurand:
@ioanarusiczki ok, thanks. -stage doesn't use BQ data yet but we should now have "Weekly downloads" on this page as well.

ioanarusiczki:
@willdurand I checked for both dev and stage using the same guids and it's the same data displayed on both servers.

PR #15441, opened 2020-09-09
Use weekly downloads in stats overview

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15440

### DIFF
--- a/src/olympia/stats/templates/stats/reports/overview.html
+++ b/src/olympia/stats/templates/stats/reports/overview.html
@@ -6,11 +6,7 @@
     <section class="island two-up">
       <div>
         <a href="downloads/">
-        {% if waffle.flag('bigquery-download-stats') %}
           {{ _('<b>{0}</b> Weekly Downloads')|format_html(addon.weekly_downloads|numberfmt) }}
-        {% else %}
-          {{ _('<b>{0}</b> Downloads')|format_html(addon.total_downloads|numberfmt) }}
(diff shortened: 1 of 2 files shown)


src/olympia/stats/templates/stats/reports/overview.html
```diff
@@ -6,11 +6,7 @@
     <section class="island two-up">
       <div>
         <a href="downloads/">
-        {% if waffle.flag('bigquery-download-stats') %}
```
eviljeff, line 9: So, is the flag `bigquery-download-stats` on already in prod?  If it is then this doesn't seem an urgent enough change to divorce from the wider removal of the code that sets/updates weekly downloads.  If it _isn't_ enabled on prod, then is weekly downloads being processed from bigquery (even with the waffle off)?

willdurand: No, the flag isn't used/enabled/active in production.

This is a consequence of the early removal of the total downloads cron task. This template currently shows the total downloads but, because we're not updating the value anymore, the value is stale and inaccurate.

Given that we were planning on showing the weekly downloads here instead, and that weekly downloads are still updated (using the legacy infra for stats), this patch removes the condition to always show the weekly downloads.

eviljeff (approved):
If there isn't already an issue logged, can we have one for removing `total_downloads` (couldn't see it in the project.

Issue #15444, opened 2020-09-09
Remove `Addon.total_downloads` field


willdurand:
We can remove the `total_downloads` field now that it isn't updated/used anymore.

ioanarusiczki:
@willdurand I think I can mark this verified since I did not find anything broken after the code removal and look in more detail on downloads for https://github.com/mozilla/addons-server/issues/15213

PR #15445, opened 2020-09-09
Remove code related to total_downloads

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15444


willdurand:
cc @diox

src/olympia/addons/migrations/0020_auto_20200909_1302.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("addons", "0019_auto_20200901_1459"),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
bobsilverberg, line 13: Interesting. So removing a field from a model doesn't cause a migration that deletes the field from the db, but rather just makes it nullable?

bobsilverberg (approved):
Looks good, thanks @willdurand. There sure are a lot of fixtures, aren't there?

src/olympia/addons/migrations/0020_auto_20200909_1302.py
willdurand: No, I had to generate the migration for that before I removed the field in the class. When deleting a field in the class, it creates a migration that performs the column deletion in the DB but I don't want it here. It will be done in a different push (see #15446)

Issue #15213, opened 2020-08-06
Use `hashed_guid`/`hashed_addon_id` for the download stats


willdurand:
BQ will expose a `hashed_addon_id` column instead of the currently exposed `addon_id`. We introduced a `hashed_guid` field on the `AddonGUID` model in #15189. This issue is about using hashes to query the download stats (in `stats/utils.py`) as well as updating the CRON job for the weekly downloads.

For QA: this issue should make the new download stats work again in -dev and it would be great to have extensive verification so that we can enable the new download stats in prod by the end of this month (September 2020).

ioanarusiczki:
@willdurand I found 4 testing samples where I can see daily downloads and the weekly downloads in the overview. 
https://addons-dev.allizom.org/en-US/firefox/addon/facebook-container/statistics/?last=7
https://addons-dev.allizom.org/en-US/firefox/addon/firefox-color/statistics/?last=30
https://addons-dev.allizom.org/en-US/firefox/addon/private-relay/statistics/?last=90

The forth one below is the one I deleted for https://github.com/mozilla/addons-server/issues/15189 and resubmitted. There was 0 weekly downloads but now it's been updated, ADU is still 0 but I think that will be updated too after a cron task run. (i'll re-check)
https://addons-dev.allizom.org/en-US/firefox/addon/statistics-reloaded/statistics/?last=365

ioanarusiczki:
@willdurand What is the difference between the weekly downloads and the last 7 days? Shouldn't they show the same numbers?
It's similar for ADU too.

An example, https://addons-dev.allizom.org/en-US/firefox/addon/facebook-container/statistics/?last=7

![7 days](https://user-images.githubusercontent.com/33448286/93177752-f132f080-f73b-11ea-8fc4-2b7af4cf898a.png)

ioanarusiczki:
@willdurand the [PRD](https://docs.google.com/document/d/1TuMqhEWNlCNDQRgqFywpJvDHTTNjnAq7-3BAXtoFLQ4/edit) says that the tracking is not desired for the unlisted addons. [The stats](https://addons-dev.allizom.org/en-US/firefox/addon/a636b7e141c14ccca7d4/statistics/?last=7) for unlisted add-ons also display the Downloads and it's sub-categories. Is that expected? (because it will never provide any information)

willdurand:
> What is the difference between the weekly downloads and the last 7 days? Shouldn't they show the same numbers?
> It's similar for ADU too

[...]

PR #15447, opened 2020-09-09
Use `hashed_guid`/`hashed_addon_id` for the download stats

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15213

---

This patch updates the stats code to make it compatible with the latest version of the "installs" view in BigQuery.

### DIFF
--- a/src/olympia/addons/tasks.py
+++ b/src/olympia/addons/tasks.py
@@ -291,11 +291,11 @@ def update_addon_weekly_downloads(data):
 
     for hashed_guid, count in data:
         try:
-            addon = AddonGUID.objects.get(hashed_guid=hashed_guid).addon
-        except AddonGUID.DoesNotExist:
+            addon = Addon.objects.get(addonguid__hashed_guid=hashed_guid)
+        except Addon.DoesNotExist:
             # The processing input comes from metrics which might be out of
             # date in regards to currently existing add-ons.
-            log.info('Got a weekly_downloads update (%s) but the add-on GUID '
+            log.info('Got a weekly_downloads update (%s) but the add-on '
                      'doesn\'t exist (hashed_guid=%s).', count, hashed_guid)
             continue
 
--- a/src/olympia/addons/tasks.py
+++ b/src/olympia/addons/tasks.py
@@ -289,14 +289,14 @@ def update_addon_hotness(averages):
 def update_addon_weekly_downloads(data):
     log.info('[%s] Updating add-ons weekly downloads.', len(data))
 
-    for addon_guid, count in data:
+    for hashed_guid, count in data:
         try:
-            addon = Addon.objects.get(guid=addon_guid)
-        except Addon.DoesNotExist:
+            addon = AddonGUID.objects.get(hashed_guid=hashed_guid).addon
+        except AddonGUID.DoesNotExist:
             # The processing input comes from metrics which might be out of
             # date in regards to currently existing add-ons.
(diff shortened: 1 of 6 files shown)


src/olympia/addons/tasks.py
```diff
@@ -289,14 +289,14 @@ def update_addon_hotness(averages):
 def update_addon_weekly_downloads(data):
     log.info('[%s] Updating add-ons weekly downloads.', len(data))
 
-    for addon_guid, count in data:
+    for hashed_guid, count in data:
         try:
-            addon = Addon.objects.get(guid=addon_guid)
-        except Addon.DoesNotExist:
+            addon = AddonGUID.objects.get(hashed_guid=hashed_guid).addon
```
eviljeff, line 294: You'll need some extra filtering on this - `AddonGUID` contains _all_ instances of a `GUID` being used and you just want the most recent addon instance rather than previous deleted ones.

I guess for _downloads_ you can just filter out deleted addons - all of the addons might have been deleted - but deleted addons don't get downloads.  (unlike ADU when you do want to keep updating the stats for admin purposes).

src/olympia/addons/tests/test_tasks.py
```diff
@@ -164,8 +164,8 @@ def test_update_addon_weekly_downloads():
 def test_update_addon_weekly_downloads_skips_non_existent_addons():
     addon = addon_factory(weekly_downloads=0)
     count = 123
-    invalid_addon_guid = 'does.not@exist'
-    data = [(invalid_addon_guid, 0), (addon.guid, count)]
+    invalid_hashed_guid = 'does.not@exist'
+    data = [(invalid_hashed_guid, 0), (addon.addonguid.hashed_guid, count)]
     assert addon.weekly_downloads == 0
```
eviljeff, line 170: Can we have a test that includes previous deleted instances of the same guid in AddonGUID to make sure the correct Addon instance is updated.

src/olympia/addons/tasks.py
eviljeff: `.get()` would have failed when there were multiple AddonGUID with the same guid (hashed_guid)

eviljeff (commented):
Is this just use for download stats (not ADU?)

willdurand:
> Is hashed_addon_id just used for download stats (not ADU?)

Yes.

src/olympia/addons/tasks.py
willdurand: Do I simply have to exclude `addon=None`?

willdurand: I am confused, how can we retrieve the last/current add-on?

src/olympia/addons/tests/test_tasks.py
willdurand: If I can understand what that means in practice, yeah. See above.

src/olympia/addons/tasks.py
eviljeff: The easiest way is probably just to rely on the Addon objects manager (like now) that filters out non-deleted addons.  So it'd be `Addon.objects.get(addonguid__hashed_guid=hashed_guid)`

src/olympia/addons/tests/test_tasks.py
eviljeff: create and delete an addon before `addon` above that has the same guid

eviljeff (approved):

PR #15448, opened 2020-09-09
Extend ?promoted= addon api filter to support multiple groups

### LINKED ISSUE #15432: allow API to filter by multiple promoted groups at once
from https://github.com/mozilla/addons-server/issues/15272#issuecomment-685924200

> In order to be able to search for "Reviewed", as discussed above, we would need to be able to specify ?promoted=recommended,verified,line

Note it'd be `?promoted=recommended,sponsored,verified,line` because the sponsored and verified are still differentiated in the api (if there is a request to actually combine the groups rather than just hide the difference in addons-frontend for now please file an issue)

There was an alternative suggestion from @bobsilverberg too:
> We could probably implement a ?promoted=reviewed argument, which the server would understand to be the same as ?promoted=recommended,verified,line

Effectively that's allowing filtering by promoted group _feature_ rather than the group itself (all groups where `.pre_review is True`) which is an interesting idea.  It's doable?  Not sure how much benefit there would be for consumers of the API though - they wouldn't need a change if the groups which were reviewed was changed... but there would probably be other changes needed so 🤷 

@jvillalobos in the issue (and PRD) you've said "Reviewed" is Recommended, Verified (& Sponsored), and Line, but Spotlight is also reviewed, it's just not badged.  Is this an oversight, or is "Reviewed" more "Badged"?

eviljeff:
fixes #15432 + refactors search/filters.py a little so all the comma separated value filters use the same logic.

### DIFF
--- a/src/olympia/search/tests/test_filters.py
+++ b/src/olympia/search/tests/test_filters.py
@@ -14,7 +14,7 @@ from olympia import amo
 from olympia.amo.tests import TestCase
 from olympia.constants.categories import CATEGORIES
 from olympia.constants.promoted import (
-    ENABLED_PROMOTED_GROUPS_BY_ID, RECOMMENDED)
+    VALID_PROMOTED_GROUPS_BY_ID, RECOMMENDED, LINE, SPOTLIGHT)
 from olympia.search.filters import (
     ReviewedContentFilter, SearchParameterFilter, SearchQueryFilter,
     SortingFilter)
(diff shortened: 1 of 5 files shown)


src/olympia/search/tests/test_filters.py
```diff
@@ -838,7 +838,7 @@ def test_search_by_author(self):
         filter_ = qs['query']['bool']['filter']
         assert len(filter_) == 1
         should = filter_[0]['bool']['should']
-        assert {'terms': {'listed_authors.id': ['123', '456']}} in should
+        assert {'terms': {'listed_authors.id': [123, 456]}} in should
```
eviljeff, line 841: I note this change in the ES query but using the integers seems more "correct" anyway?

diox: Both should work equally

diox (approved):

PR #15449, opened 2020-09-09
Update 0006_auto_20200819_0951.py

eviljeff:
follow up for #15331 to force affected addons to be reindexed


src/olympia/promoted/migrations/0006_auto_20200819_0951.py
```diff
@@ -14,6 +15,20 @@ def make_recommended_firefox_only(apps, schema_editor):
     for promo in qs:
         promo.application_id = amo.FIREFOX.id
         promo.save()
+    if qs.count():
```
diox, line 18: `.exists()` would have been enough, or even storing `[promoted.addon_id for promoted in qs]` into a variable and checking that variable, but this should be good enough.

diox (approved):

src/olympia/promoted/migrations/0006_auto_20200819_0951.py
eviljeff: it would, but the for loop above would have cached the query response so `.count()` is pretty much the same as `.exists()` then.

Issue #15454, opened 2020-09-10
Remove `?src=` in download urls (files)


willdurand:
We append a `src` query parameter to all file urls. We don't need this anymore as we are going to use UTM query params instead but not on the file urls themselves.

AlexandraMoga:
I was able to download files from DevHub, Rev Tools and Admin tools, which, I think, are the areas affected by this issue.
I filed #15479 as a follow up, but I'm not sure that it is indeed a regression from this patch.

PR #15457, opened 2020-09-10
Remove src query parameter in download links

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15454


bobsilverberg (approved):
Looks good. Just wondering, why was `src/olympia/files/fixtures/files/jetpack.xpi` added?

r+wc

willdurand:
> why was `src/olympia/files/fixtures/files/jetpack.xpi` added?

ugh, I removed it. Maybe that was removed recently?

PR #15469, opened 2020-09-11
allow random addon sort to be used with promoted filter

### LINKED ISSUE #15462: Update the logic for random sort to use promoted rather than recommended
There is some logic in [`filter_queryset` in `SortingFilter`](https://github.com/mozilla/addons-server/blob/master/src/olympia/search/filters.py#L900-L926) which is designed to protect against a sort of `random` being requested without an accompanying `featured` or `recommended` param. This logic should be updated to look for the `promoted` param, rather than the `recommended` param.

eviljeff:
fixes #15462


diox (approved):

bobsilverberg:
Thanks for taking care of this so quickly, @eviljeff!

PR #15471, opened 2020-09-12
Added GitHub Sponsor to contributors domain

conema:


eviljeff:
@jvillalobos to approve.  It's a little annoying that there isn't a specific (sub)domain - I guess we'll just have to trust that developers won't just stick their main github repo url in the contributions url setting.

jvillalobos:
Yeah, I am also concerned that this is so open ended. The link could be pointing to a Readme or some other file that attempts to phish users. I think in order to accept this we would need to at least validate the URL is in `https://github.com/sponsors/`, which would make an exception on how we deal with this validation now.

jvillalobos (changes requested):
I think the next step here is to file an issue to figure out how to make this exception.

eviljeff:
yeah.  Issues can be filed here: https://github.com/mozilla/addons-server/issues/new @conema - I don't know how much work it'll be to expand the simplistic list of domains to require specific paths (for some domains?).  Anyway, discussion in that issue.  Closing this PR.

PR #15478, opened 2020-09-14
add ?promoted=badged search filter

### LINKED ISSUE #15468: add meta promoted=badged group in addons api
Currently if a consumer of the addons api (e.g. addons-frontend) wants to filter by all add-ons that are supposed to badged (exposed as being reviewed by Mozilla) it has to know which promoted groups have that property and then specify all of those groups as `?promoted=recommended,verified,sponsored,line`.  If we have a `?promoted=badged` filter then the consumer of the api can make a simpler query that isn't dependent on identical logic being maintained on addons-server and addons-frontend.

eviljeff:
fixes #15468


bobsilverberg (approved):
Looks good, thanks for taking this on @eviljeff!

PR #15480, opened 2020-09-14
Remove/fix more instances of `src` being passed to file download url

### LINKED ISSUE #15479: Admin activity log is broken (again)
Open the Admin activity log at https://addons-internal.dev.mozaws.net/en-US/admin/models/activity/activitylog/

### What happened?
A 500 server error

The activity log page is opened

Last time the activity log was broken it had something to do with files and the removal of the legacy file viewer - see #15291. The only closed issue that touched files now is #15454 I think?

diox:
Fixes #15479

### DIFF
--- a/src/olympia/devhub/templates/devhub/includes/version_file.html
+++ b/src/olympia/devhub/templates/devhub/includes/version_file.html
@@ -1,6 +1,6 @@
 <tr>
   <td>
-    <a href="{{ file.get_absolute_url('devhub') }}">
+    <a href="{{ file.get_absolute_url() }}">
       {{ file.pretty_filename() }}</a>
   </td>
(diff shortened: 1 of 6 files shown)


src/olympia/devhub/templates/devhub/includes/version_file.html
```diff
@@ -1,6 +1,6 @@
 <tr>
   <td>
-    <a href="{{ file.get_absolute_url('devhub') }}">
```
diox, line 3: This and below didn't fail, but caused the files to be sent as attachment instead... (because it was then passed as `attachment='devhub'`, which evaluates to `True`.)

willdurand (approved):
:facepalm:

PR #15485, opened 2020-09-15
delete old mlbf folders after 6 months

### LINKED ISSUE #14749: Optimize the storage in mlbf directory
Currently dumped blocked/notblocked json files are saved in `<storage_root>/mlbf` directory. In chatting with @eviljeff [[1]](https://mozilla.slack.com/archives/C4MTGNZ7S/p1589560549015900), we think that there are two possible issues with the current design.

First is that there is no hierarchy under `mlbf/`, thus this directory can grow very big, and cause the same management headache as the `files/` directory does (e.g. nearly impossible to do `ls`; hard to find files, and etc.).

Second is that the directory can grow very big given each dump is over 150 MB, and at a maximum rate of 28 dumps per week. This puts pressure on costs and backup operations.

### What happened?
As above.

As mentioned in the slack thread [[1]](https://mozilla.slack.com/archives/C4MTGNZ7S/p1589560549015900), there are two things we can do to improve this:

One is to delete old dumps periodically and keep only dumps that are X days/months old on the disk.

Two is to implement a hierarchical directory structure under `mlbf/` similar to  what's in `<storage root>/git-storage/` directory. That will make management of this directory much easier.

Out of the two, if we don't keep too many days worth of dumps, then I think the second measure is probably optional. Because if we prune the `mlbf/` directory periodically, then the sub-directories won't grow to a unmanageable level anyways.

 
n/a

eviljeff:
fixes #14749

### DIFF
--- a/src/olympia/blocklist/cron.py
+++ b/src/olympia/blocklist/cron.py
@@ -1,5 +1,3 @@
-import time
-
 from django.core.management import call_command
 from django.core.management.base import CommandError
 
@@ -13,7 +11,8 @@ from olympia.zadmin.models import get_config
 
 from .mlbf import MLBF
 from .models import Block
-from .tasks import upload_filter
+from .tasks import cleanup_old_files, upload_filter
+from .utils import datetime_to_ts
 
 
 log = olympia.core.logger.getLogger('z.cron')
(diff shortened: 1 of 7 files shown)


bobsilverberg (approved):
Looks good, nice work @eviljeff!

Issue #15487, opened 2020-09-15
Hide download stats links when add-on is purely unlisted


willdurand:
We don't have download stats data for purely unlisted add-ons (or any unlisted versions). Let's hide those links when add-on is purely unlisted because they are not very useful. When an add-on has some listed versions (at least one), then we should display the links.

ioanarusiczki:
@willdurand 

2 aspects noticed:

- for an unlisted add-on there are no downloads shown but there's still the URL which redirects to them in the overview. 

![no url](https://user-images.githubusercontent.com/33448286/93325725-68da4b80-f820-11ea-8f74-89c5ae2d2260.gif)

- the scenario: when the listed version is deleted and the unlisted remains the downloads will be shown. 

Example: https://addons-dev.allizom.org/en-US/firefox/addon/statistics-reloaded/statistics/?last=7

willdurand:
> for an unlisted add-on there are no downloads shown but there's still the URL which redirects to them in the overview.

yeah, that's normal. It does not really matter in this case.

> the scenario: when the listed version is deleted and the unlisted remains the downloads will be shown.

Yes because there could be data from the past.

ioanarusiczki:
@willdurand Ok, then it is verified on dev - FF80(Win10).

PR #15488, opened 2020-09-15
Hide download stats links in menu when add-on is purely unlisted

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15487

### DIFF
--- a/src/olympia/stats/templates/stats/addon_report_menu.html
+++ b/src/olympia/stats/templates/stats/addon_report_menu.html
@@ -3,7 +3,7 @@
     <li data-report="overview" data-layout="overview">
       <a href="{{ url('stats.overview', addon.slug) }}">{{ _('Overview') }}</a>
     </li>
-    {% if addon.has_listed_versions() %}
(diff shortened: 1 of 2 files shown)


src/olympia/stats/templates/stats/addon_report_menu.html
```diff
@@ -3,6 +3,7 @@
     <li data-report="overview" data-layout="overview">
       <a href="{{ url('stats.overview', addon.slug) }}">{{ _('Overview') }}</a>
     </li>
+    {% if addon.has_listed_versions() %}
```
eviljeff, line 6: Do we show stats for old versions after they've been deleted?  if so we might want `include_deleted=True` here.

eviljeff (approved):

src/olympia/stats/templates/stats/addon_report_menu.html
willdurand: > Do we show stats for old versions after they've been deleted?

Yeah, I think that's possible so let's included deleted versions, that probably does not hurt.

Issue #15192, opened 2020-08-05
Set `null=False` to `AddonGUID.hashed_guid` once it lives in production


willdurand:
We added a new field to the `AddonGUID` model in https://github.com/mozilla/addons-server/issues/15189. Once this field exists in production and we run a backfill script, we'll want to change the `null` attribute to make this field non-nullable.

ioanarusiczki:
@willdurand Please add some STR for qa.
Thanks!

willdurand:
@ioanarusiczki added, sorry

AlexandraMoga:
I've submitted listed and unlisted extensions, themes, language tools and I didn't find any issues.

I'm marking this as verified fixed on stage.

PR #15499, opened 2020-09-16
Make AddonGUID.hashed_guid required

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15192

---

I removed the test case because it's not possible to create the test data due to the DB constraint. Maybe we could delete the command itself too?


eviljeff (approved):
Yeah, delete the command and the task too

PR #15549, opened 2020-09-22
Upgrade node version to v12

### LINKED ISSUE #15513: Upgrade to Node 12
We'd like to use the new Node LTS (12) whenever it's possible.

bqbn:
Fix #15513

* [X] This PR relates to an existing open issue and there are no existing
      PRs open for the same issue.
* [X] Add `Fixes #ISSUENUM` at the top of your PR.
* [X] The change has been successfully run locally.

### DIFF
--- a/docker/debian-buster-nodesource-repo
+++ b/docker/debian-buster-nodesource-repo
@@ -1,2 +1,2 @@
-deb https://deb.nodesource.com/node_10.x buster main
-deb-src https://deb.nodesource.com/node_10.x buster main
+deb https://deb.nodesource.com/node_12.x buster main
+deb-src https://deb.nodesource.com/node_12.x buster main


Zenix27 (approved):

willdurand (commented):
There is one additional file that shouldn't be there. I don't why, though. I'll update the patch!

bqbn:
> There is one additional file that shouldn't be there. I don't why, though. I'll update the patch!

Thanks for catching that!

It turns out I need to set `core.precomposeunicode` to `false` in order for my local git to treat the unicoded file name of [this file](https://github.com/mozilla/addons-server/blob/master/src/olympia/files/fixtures/files/je%CC%81tpack.xpi) the same as the repository does, or my local git will think the existing file is new and tries to add it.

PR #15558, opened 2020-09-23
use curlylint (and existing fix template problems)

### LINKED ISSUE #15535: Integrate curlylint for template linting
https://pypi.org/project/curlylint/

We don't have any linting for templates currently so we're dependant on functional tests not returning template errors in their response which is fine... but something specific would be better.  Curlylint also claims to check for accessibility which is worthwhile to add.

I gave it a quick try and it found 15 errors in our templates - mainly in devhub.

eviljeff:
fixes #15535 - most of the fixes were minor, but some needed a bit of reworking to not have fragments of tags spread over `{% if .... endif %}` blocks.   compat_app_input_option.html was a ... challenge (it's django template rather than jinja too which means a more restrictive syntax)

### DIFF
--- a/src/olympia/devhub/templates/devhub/addons/submit/done.html
+++ b/src/olympia/devhub/templates/devhub/addons/submit/done.html
@@ -66,12 +66,13 @@
 {% endif %}
 
 <p>
-  {% if addon.type == amo.ADDON_STATICTHEME %}
-    <a class="button" href="{{ url('devhub.themes') }}">
-  {% else %}
-    <a class="button" href="{{ url('devhub.addons') }}">
-  {% endif %}
-    {{ _("Go to My Submissions") }}</a>
+  {% with link_text=_("Go to My Submissions") %}
+    {% if addon.type == amo.ADDON_STATICTHEME %}
+      <a class="button" href="{{ url('devhub.themes') }}">{{ link_text }}</a>
+    {% else %}
+      <a class="button" href="{{ url('devhub.addons') }}">{{ link_text }}</a>
+    {% endif %}
+  {% endwith %}
 </p>
 
 {% endblock %}
--- a/src/olympia/devhub/templates/devhub/addons/submit/done.html
+++ b/src/olympia/devhub/templates/devhub/addons/submit/done.html
@@ -66,12 +66,8 @@
 {% endif %}
 
 <p>
-  {% with link_text=_("Go to My Submissions") %}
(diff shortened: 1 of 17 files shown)


src/olympia/devhub/templates/devhub/addons/submit/done.html
```diff
@@ -66,12 +66,13 @@ <h3>{{ _("Version Signature Pending") }}</h3>
...
-    <a class="button" href="{{ url('devhub.themes') }}">
-  {% else %}
-    <a class="button" href="{{ url('devhub.addons') }}">
-  {% endif %}
-    {{ _("Go to My Submissions") }}</a>
+  {% with link_text=_("Go to My Submissions") %}
+    {% if addon.type == amo.ADDON_STATICTHEME %}
+      <a class="button" href="{{ url('devhub.themes') }}">{{ link_text }}</a>
+    {% else %}
+      <a class="button" href="{{ url('devhub.addons') }}">{{ link_text }}</a>
```
diox, line 73: Tempted to just put the `{% if %}` inside the `href` attribute here... What do you think?

src/olympia/devhub/templates/devhub/versions/list.html
```diff
@@ -134,7 +134,7 @@
               <form class="dev-review-reply-form" action="{{ drf_url('version-reviewnotes-list', addon.id, version.id) }}"
                     data-token="{{ token }}" data-history="#{{ version.id }}-review-history" data-no-csrf>
                   <textarea name="comments" placeholder="{{ _('Leave a reply') }}"></textarea>
-                  <button type="submit" class="submit" >{{ _('Reply') }}</button ty>
```
diox, line 137: !

diox (approved):
Love it!

src/olympia/devhub/templates/devhub/addons/submit/done.html
eviljeff: yeah.  As it's jinja we could either write out the full `{% if ... }` or use it inline instead, like `{% with link=url('devhub.themes') if addon.type == amo.ADDON_STATICTHEME else url('devhub.addons') %}`

eviljeff: https://github.com/mozilla/addons-server/pull/15558/commits/5de6dcaef2e9aed7bb51bb244f6c157ede7d59b9

PR #15566, opened 2020-09-24
upgrade to django 3.1's JSONField

### LINKED ISSUE #15559: Upgrade to django's JSONField
django 3.1 introduces a built-in JSONField (replacing the many JSONField packages out there, including the one we use -`django_extensions.db.fields.json.JSONField`).  It's available as back-port https://github.com/laymonage/django-jsonfield-backport so we can upgrade this feature now before we upgrade addons-server to the next django LTS (3.2).

eviljeff:
fixes #15559 - I did a quick scan of the usage of the affected fields and we mainly just set and get them as model attributes.  `blocklist.models.BlockSubmission.get_submissions_from_guid` was the only place I found that was actually filtering on them (which we can do properly now!).

I also ran a few of these migrations locally without errors, but I don't have the range of weird data that's on -dev that could potentially trip up the db field conversions.


diox (commented):
I'm concerned about the performance of the `ALTER TABLE` needed, particularly for `download_stats`. Since we're planning to get rid of that table in a couple weeks, maybe wait until then ?

eviljeff:
> I'm concerned about the time needed for some of the `ALTER TABLE`, particularly for `download_stats`. Since we're planning to get rid of that table in a couple weeks (actually a bit more than that since we won't get rid of it the second we switch to the new download stats), maybe wait until then ? Or simply don't migrate that one.

we can leave that one using the existing JSONField.  Are there any other tables you see as problematic? ActivityLog could be but it's fields aren't currently defined as any type of JSONField so I didn't touch it in this patch.

diox:
`webext_permissions` is another big one with millions of rows, the rest have hundreds of thousands entries but hopefully should be ok. Ideally we'd test just the `ALTER` on stage beforehand to see how much time it takes and go back to `varchar` before landing this.

eviljeff:
https://github.com/mozilla/addons-server/pull/15566/commits/556c4b5705051438e074f62ec4cc005460d4a480 excludes DownloadCount

Ops (@bqbn) estimated the migrations in total may take around 30 minutes (from manually running the SQL on some of the larger tables on a database copy).  To be safe (and in case it ends up being longer) we could put the site into read-only mode.

diox (approved):

PR #15576, opened 2020-09-25
Group scanner query results by add-on and channel

### LINKED ISSUE #12947: Group scanner results per add-on
Scanner results should be grouped per add-on. Note that it should still be possible to triage the individual results.

diox:
The way django generates the changelist makes it a bit tricky to customize, so we run some javascript once the dom is loaded to tweak the rowspans, remove duplicate cells and add relevant css classnames.

Fixes #12947

Screenshot_2020-09-25 Select scanner query result to view Django site admin(1)

### DIFF
--- a/src/olympia/scanners/admin.py
+++ b/src/olympia/scanners/admin.py
@@ -390,7 +390,9 @@ class AbstractScannerResultAdminMixin(admin.ModelAdmin):
     channel.short_description = 'Channel'
 
     def formatted_created(self, obj):
-        return obj.created.strftime('%Y-%m-%d %H:%M:%S')
+        if obj.version:
+            return obj.version.created.strftime('%Y-%m-%d %H:%M:%S')
+        return '-'
 
     formatted_created.short_description = 'Created'
 
@@ -406,11 +408,12 @@ class AbstractScannerResultAdminMixin(admin.ModelAdmin):
         return format_html(
             ', '.join(
                 [
-                    '<a href="{}">{}</a>'.format(
+                    '<a href="{}">{} ({})</a>'.format(
                         reverse(
                             'admin:%s_%s_change' % info, args=[rule.pk]
                         ),
                         rule.name,
+                        rule.get_scanner_display(),
                     )
                     for rule in obj.matched_rules.all()
                 ]
@@ -678,15 +681,15 @@ class ScannerResultAdmin(AbstractScannerResultAdminMixin, admin.ModelAdmin):
 class ScannerQueryResultAdmin(
         AbstractScannerResultAdminMixin, admin.ModelAdmin):
     raw_id_fields = ('version',)
+    list_display_links = None
     list_display = (
-        'id',
         'formatted_addon',
-        'guid',
+        'channel',
+        'version_number',
+        'formatted_created',
         'authors',
-        'scanner',
         'formatted_matched_rules',
         'matching_filenames',
-        'formatted_created',
     )
     list_filter = (
         ('matched_rules', ScannerRuleListFilter),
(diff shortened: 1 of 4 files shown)


src/olympia/scanners/admin.py
```diff
@@ -697,6 +700,51 @@ class ScannerQueryResultAdmin(
         ('version__files__is_signed', FileIsSigned),
     )
 
+    ordering = ('version__addon_id', 'version__channel', 'version__created')
```
diox, line 703: We don't have an index on `Version` doing that, it might help this and the `AddonVersionViewSet`, I'll investigate.

diox: I've tested it on prod and played with it locally and it looks like this wouldn't help, at least with the version of MySQL we're using (the table that would need better indexes is `scanners_scannerquerymatch`, we'd need some serious denormalization for this to work). It's slow when there are several hundred thousands results (worst case) but should still be bearable.

diox: Also, because the problem is the ordering of results with `scanners_scannerquerymatch` this actually doesn't seem to make it any worse than the current ordering on `created`.

static/js/admin/scannerqueryresult.js
```diff
@@ -0,0 +1,65 @@
...
+    // This is only for the scanners query result change list page.
+    return;
+  }
+
+  let current_addon = {
+    cell: null,
+    link: null,
+  };
+
+  for (row of result_list.tBodies[0].rows) {
```
willdurand, line 12: Where is `result_list` coming from?

static/js/admin/scannerqueryresult.js
```diff
@@ -0,0 +1,65 @@
...
+    let new_addon = {
+      cell: row.querySelector('.field-formatted_addon'),
+      link: row.querySelector('.field-formatted_addon a'),
+    };
+
+    // FIXME: for performance, it might be better to store the elements we
+    // want to change in a list and change them in one go outside of this
+    // loop ?
```
willdurand, line 20: I suppose it's fine unless you have more than 1K elements probably?

willdurand (approved):
r+wc, I wish I could test that feature more

static/js/admin/scannerqueryresult.js
diox: Uh, there used to be a selector to get that element by id... I must have deleted it by accident... but I was saved by the fact that browsers auto-create the global variable since it's the exact id... I'll add an explicit call.

static/js/admin/scannerqueryresult.js
diox: Yeah - and we paginate for 100, so it's likely fine (appears ok on my machine), but I left the comment for now just in case we want to revisit the pagination someday.

static/js/admin/scannerqueryresult.js
```diff
@@ -1,5 +1,9 @@
 document.addEventListener('DOMContentLoaded', () => {
-  if (!document.body.matches('.change-list.model-scannerqueryresult')) {
+  let result_list = document.querySelector(
```
willdurand, line 2: If you don't change `result_list`, you may want to use `const` instead.

diox: https://github.com/mozilla/addons-server/pull/15576/commits/40844610cccf8a5e1120328ae59bbb1f4d897ec7

static/js/admin/scannerqueryresult.js
diox: https://github.com/mozilla/addons-server/pull/15576/commits/3f0ce7b9fa0ea3e5f205fee1d9ab14a267485beb

PR #15589, opened 2020-09-29
change url to path in olympia.urls.py

Sparsh-Bansal:
The old django.conf.urls.url() function is replaced with new django.urls.path() function for routing .
The old django.conf.urls.include() function is now importable from django.urls so it is also changed.
The new django.urls.path() function allows a simpler, more readable URL routing syntax.
The django.urls.re_path() is used for complex url mapping.
For Django-2.x , Its more efficient to use the latest path() and re_path() instead of url().
The django.conf.urls.url() function from previous versions is now available as django.urls.re_path(). The old location remains for backwards compatibility, without an imminent deprecation.

### DIFF
--- a/src/olympia/urls.py
+++ b/src/olympia/urls.py
@@ -1,5 +1,18 @@
+#
+# from django.contrib import admin
+# from django.urls import path
+# from django.conf.urls import include
+# from firstapp import views
+#
+# urlpatterns = [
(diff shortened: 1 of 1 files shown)


diox:
Please don't open random pull requests with no associated issue - we like to discuss changes before starting to make them :) Also, it's likely there would be a lot more files to change that just this one.

Issue #15560, opened 2020-09-23
The download hits total metric is decreasing


bqbn:
### Describe the problem and steps to reproduce it:
We monitor the elasticsearch cluster's  `/addons_stats_download_counts_prod/_search` endpoint for the `hits.total` metric, and we notice that the value of it has dropped below 1000 since Sunday (9/20/2020), as shown below,

<img width="747" alt="Screen Shot 2020-09-23 at 11 59 11 AM" src="https://user-images.githubusercontent.com/2209111/94057375-6170f000-fd94-11ea-9b9d-ce8ef780d994.png">

The logs show that the 2 related daily cron jobs, `download_counts_from_file` and `index_stats`, finished successfully. We also tried to re-ran the jobs manually, and they both completed without errors.

### What happened?
We are not sure why the `hits.total` metric started to drop since 9/20/2020, and if it is expected or it is an indication that something is wrong.

### What did you expect to happen?
To understand if this is expected or not, and in the case that it is not, we need to fix the issue.

### Anything else we should know?
Note that the metric started to drop to below 1000 on Sunday, which was two days after last push day (which was on Thursday).

willdurand:
The removal of the `src` parameter unexpectedly broke the `download_counts_from_file` command because it seems to require that `src` parameter. We can probably fix the command to allow no `src` param and maybe re-run this command for all days since September 18th?

willdurand:
Marking as p1 because we asked devs to backup their download stats before October 8th.

bqbn:
> ... maybe re-run this command for all days since September 18th?

If we're to re-run the commands, I believe they are

```
./manage.py download_counts_from_file --date YYYY-MM-DD
./manage.py index_stats --date YYYY-MM-DD
```

for each affected date.

ioanarusiczki:
@willdurand In case I can test on the issue, please add some STR
Thanks!

willdurand:
You need a user on -dev who does not have the bigquery waffle flag enabled. One way to do it is to visit a stats page like this one: https://addons-dev.allizom.org/en-US/firefox/addon/private-relay/statistics/?dwft_bigquery-download-stats=0 (note the query param). This should set a cookie, if not clear all cookies/caches and retry.

Loading this page without the bigquery flag enabled should show no download stats and only a "By source" page in the left menu (under "Downloads"). If you can still see some download stats, clear your session cache manually in the devtools.

[...]

PR #15591, opened 2020-09-29
Handle download stats without source

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15560

---

Thanks to ops, I was able to review some hive files from production. We have a lot more NULL than before:

```
cat ~/Desktop/2020-09-16_prod.hive |grep -v '\N' |wc -l
   47796

~/projects/mozilla/addons-server fix-download-counts*
❯ cat ~/Desktop/2020-09-28_prod.hive |grep -v '\N' |wc -l
    1506
```

We need to handle those and that's what the patch does.

I created a `DownloadCount` instance manually:

```
DownloadCount.objects.create(addon=addon, count=3, date='2020-09-28', sources={'search': 2, 'null': 1})
```

and ran the `index_stats` command. After that I was able to see the following:

Screenshot 2020-09-29 at 17 41 18

### DIFF
--- a/src/olympia/stats/fixtures/files/src/download_counts.hive
+++ b/src/olympia/stats/fixtures/files/src/download_counts.hive
@@ -8,3 +8,4 @@
 2014-07-10	1	 a3615	search
 2014-07-10	1	 disabled-addon	search
 2014-07-10	1	 incomplete-addon	search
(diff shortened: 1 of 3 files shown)


diox (approved):
Makes sense.

PR #15593, opened 2020-09-29
Mark devhub/forms/widgets/compat_app_input_option.html as a django template

diox:
Otherwise l10n extraction is broken as it tries to load it as a django template.

Django itself uses `settings.FORM_RENDERER` or the form `default_renderer`/`renderer` properties, which in the end defaults to the django template loader, so it should be considered as a django template at all times.


eviljeff (approved):
still not sure if the template shouldn't just be rewritten to not use any django template specific tags, but if this works 🤷

diox:
If we rewrote it to avoid django-specific tags it would still be incredibly brittle - it's a django template, so it should be marked as such.

Issue #15583, opened 2020-09-28
The custom date range is not shown correctly under the weekly downloads and average daily users


ioanarusiczki:
As mentioned in https://github.com/mozilla/addons-server/issues/15446#issuecomment-698267720

STR:
1. Load the overview page for https://addons-dev.allizom.org/en-US/firefox/addon/facebook-container/statistics/
or https://addons-dev.allizom.org/en-US/firefox/addon/firefox-color/statistics/
2. Select a custom data range 
3. Observe the dates below the weekly downloads and average daily users

Expected result:
The selected dates are displayed. 

Actual result:
There is an extra day added at the end of the interval. 

Example:

![testing](https://user-images.githubusercontent.com/33448286/94416391-792ae880-0187-11eb-98a8-cad09482b5f5.png)

Reproducible on AMO dev and stage - FF80(Win10).

ioanarusiczki:
@willdurand This is fixed on dev - FF81(Win10). 

![correct dates](https://user-images.githubusercontent.com/33448286/94914829-6af50900-04b4-11eb-93d0-0f577c8bb558.png)

PR #15604, opened 2020-09-30
Fix wrong date on stats overview page

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15583

---

I tried to write some tests but it was a lot more code than I expected. I don't
know if I'll continue to add tests like this, although it's probably better
than nothing..

### DIFF
--- a/static/js/impala/stats/overview.js
+++ b/static/js/impala/stats/overview.js
@@ -1,20 +1,12 @@
-$(function () {
-  if ($('.primary').attr('data-report') != 'overview') return;
+(function (jQuery, window) {
+  'use strict';
 
-  // set up topcharts (defined in topchart.js)
-  $('.toplist').topChart();
+  // This function is called once we have stats data and we get aggregates for
+  // daily users and weekly downloads.
+  const stats_overview_make_handler = function ({ view }) {
+    const range = normalizeRange(view.range);
 
-  $(window).on('changeview', function (e, view) {
-    $('.two-up').addClass('loading');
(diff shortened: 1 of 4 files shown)


static/js/impala/stats/overview.js
```diff
@@ -28,6 +20,13 @@ $(function () {
           endString = range.end.iso(),
           downloadFormat,
           userFormat;
+
+        // Trim end date by one day if custom range.
+        if (view.range.custom) {
+          const msDay = 24 * 60 * 60 * 1000; // One day in milliseconds.
+          endString = new Date(range.end.getTime() - msDay).iso();
```
willdurand, line 27: That's the actual "fix", which is what we're using elsewhere.

tests/js/zamboni/stats.spec.js
```diff
@@ -124,4 +124,110 @@ describe(__filename, () => {
...
+      const createMinimalHTML = () => {
+        return `
+<div class="primary" data-report="overview">
+  <div id="downloads-in-range"></div>
+  <div id="users-in-range"></div>
+</div>`;
+      };
+
+      beforeEach(() => {
+        const date = new Date(2019, 10 - 1, 14);
```
bobsilverberg, line 147: Why do we need this date setup? It doesn't seem like any of the tests care about this date.

bobsilverberg (approved):
Looks good, thanks @willdurand. I just have a question about the test setup.

r+wc

PR #15608, opened 2020-09-30
GitHub Sponsor path validation & tests

### LINKED ISSUE #15497: Add GitHub Sponsor to contributors domain
As discussed in PR #15471, it will be a nice thing if GitHub Sponsor was added to the list of accepted contributors domain. As noted by @eviljeff and @jvillalobos , it's annoying that there isn't a specific (sub)domain for GitHub Sponsor, since it could lead to phishing attempt. 
So, other than adding the domain to the accepted contributors list, there is also the need to expand the list of domains to require specific paths and not only specific domains.

conema:
Fixes #15497
The PR implements a special-case path validator for GitHub Sponsor as described by @diox's [reply](https://github.com/mozilla/addons-server/issues/15497#issuecomment-693619605) and adds two tests. 

immagine

### DIFF
--- a/src/olympia/devhub/forms.py
+++ b/src/olympia/devhub/forms.py
@@ -342,11 +342,20 @@ class AdditionalDetailsForm(AddonFormBase):
 
     def clean_contributions(self):
         if self.cleaned_data['contributions']:
-            hostname = urlsplit(self.cleaned_data['contributions']).hostname
+            parsed_url = urlsplit(self.cleaned_data['contributions'])
+            hostname = parsed_url.hostname
+            path = parsed_url.path
+            
             if not hostname.endswith(amo.VALID_CONTRIBUTION_DOMAINS):
                 raise forms.ValidationError(ugettext(
                     'URL domain must be one of [%s], or a subdomain.'
                 ) % ', '.join(amo.VALID_CONTRIBUTION_DOMAINS))
+            elif hostname == 'github.com' and not path.startswith('/sponsors/'):
+                # Issue 15497, validate path for GitHub Sponsor
+                raise forms.ValidationError(ugettext(
+                    'URL path for GitHub Sponsor must contain /sponsors/.'
+                ))
+
         return self.cleaned_data['contributions']
 
     def clean(self):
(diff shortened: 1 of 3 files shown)


eviljeff:
you've got some codestyle fails around line length https://travis-ci.org/github/mozilla/addons-server/jobs/731753477#L555 (you can replicate locally with `make lint`)

src/olympia/devhub/forms.py
```diff
@@ -342,11 +342,21 @@ def __init__(self, *args, **kw):
...
-            hostname = urlsplit(self.cleaned_data['contributions']).hostname
+            parsed_url = urlsplit(self.cleaned_data['contributions'])
+            hostname = parsed_url.hostname
+            path = parsed_url.path
+
             if not hostname.endswith(amo.VALID_CONTRIBUTION_DOMAINS):
                 raise forms.ValidationError(ugettext(
                     'URL domain must be one of [%s], or a subdomain.'
                 ) % ', '.join(amo.VALID_CONTRIBUTION_DOMAINS))
+            elif (hostname == 'github.com' and
```
bobsilverberg, line 353: I wonder if it's worth making `github.com` into a constant, so that it can be used here, in `src/olympia/constants/base.py`, and also in the tests, without having to hardcode it multiple times? It would also allow for it to be changed easily if for some reason the GitHub hostname changed.

@eviljeff | @diox WDYT?

bobsilverberg (commented):
This looks great! Outstanding job, @conema! I had one suggestion for a possible improvement, but I'm seeking feedback from the primary server developers on that. Great work so far, and stay tuned!

src/olympia/devhub/forms.py
diox: I don't think it's really worth the effort.

bobsilverberg: Fair enough. This is good to land, then. Thanks for the patch @conema!

caitmuenster:
Thanks so much for the patch, @conema! 🙌 Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

Welcome onboard -- we look forward to seeing you around the project! 😎

Issue #15628, opened 2020-10-02
Code-search: Separate add-on name and guid and add link to XPI


wagnerand:
The recent changes to the code-search styling are a huge improvement, thank you @diox for doing that!

Can we undo combining add-on name and guid, and have them in separate columns again, please? This helps admins select and copy the guids for further processing.

Also, can we please add a separate "XPI" column (at the end) that holds a link to the XPI file of the match?
Nit: The link text could be something static like "XPI", or we could show something potentially slightly more helpful like the file id. (The full link as text might ideal from a copy-paste perspective, but actual usefulness and usability might be limited as some links might not be accessible easily (need to pass your auth cookie outside the browser), and it would probably take quite some space (horizontally or vertically, depending on the column width).)

AlexandraMoga:
The code search layout looks fine to me. The add-on name and the guid are split on different columns and a link to download the `.xpi` has been added.
Just a note that the download link will trigger an installation instead of an actual file download when clicked, so you'll have to right click to save the file.

![image](https://user-images.githubusercontent.com/31961530/95103345-2f20a480-073d-11eb-9ce6-0bcf34963a5d.png)

wagnerand:
@diox is there a way to force-trigger a download instead of installation?

diox:
Mmm, we don't seem to do that for reviewer tools, but it should be possible yes. Can you open a new issue ?

wagnerand:
Great! Yes, it's been a long-standing issue in the reviewer tools as well, see #8426.

AlexandraMoga:
@diox since you are working on the download links, you might also want to look at #9303

PR #15629, opened 2020-10-02
Further tweaks  to code search results display

diox:
- Review link is now on the channel column, always repeated
- Guid is now back to being a separate column (with a monospace font to differentiate it from the add-on name)
- Name and guid duplicates are now always eliminated
- Download link column has been added

Fixes https://github.com/mozilla/addons-server/issues/15628


bobsilverberg (approved):
LGTM!

PR #15645, opened 2020-10-05
add fix for empty strings to migration addons 0024

eviljeff:
migration choked on bad JSON (empty string) in the database (from 2012)


eviljeff:
tested locally by setting a record to `''` in db shell and then running the migration

diox (approved):

PR #15670, opened 2020-10-07
add sponsored shelf that uses Adzerk for placement

### LINKED ISSUE #15617: Add API to return sponsored add-ons for the homepage shelf
We need an API to return sponsored add-ons for the homepage shelf. It needs to call Adzerk API to decide which add-ons to show, using the `metadata` returned to find add-on ids.

On top of the regular add-on serializer info returned, we need to return 2 different additional URLs:
- The impression URL (proxied through https://github.com/mozilla/addons-server/issues/15618)
- The click URL (proxied through https://github.com/mozilla/addons-server/issues/15619)

Both are returned by adzerk but needs to be modified to be proxied through our servers (hiding client IPs from Adzerk and letting us implement rate limiting, integrating with the fraud pipeline etc later if needed).

PRD: https://docs.google.com/document/d/19LgWE3Bs_UUE6uV18rwhJOKaMHyPGEoPvg5zJ7gsZwc/edit?pli=1

eviljeff:
fixes #15617

### DIFF
--- a/src/olympia/shelves/utils.py
+++ b/src/olympia/shelves/utils.py
@@ -1,0 +1,79 @@
+from urllib.parse import urlparse
+
+from django.conf import settings
+
+import requests
+from django_statsd.clients import statsd
+
+import olympia.core.logger
+
+
+log = olympia.core.logger.getLogger('z.shelves')
+
+
+def call_adzerk_server(placeholders):
+    """Call adzerk server to get sponsored addon results.
+
+    `placeholders` is a list of arbitrary strings that we pass so we can
+    identify the order of the results in the response dict."""
+    site_id = settings.ADZERK_SITE_ID
+    network_id = settings.ADZERK_NETWORK_ID
+    placements = [
+        {"divName": ph,
+         "networkId": network_id,
+         "siteId": site_id,
+         "adTypes": [5]} for ph in placeholders]
+
+    try:
+        log.info('Calling adzerk')
+        with statsd.timer('services.adzerk'):
+            response = requests.post(
+                settings.ADZERK_URL,
+                json={'placements': placements},
+                timeout=settings.ADZERK_TIMEOUT)
+        if response.status_code != 200:
+            raise requests.exceptions.RequestException()
+    except requests.exceptions.RequestException as e:
+        log.exception('Calling adzerk failed: %s', e)
+        statsd.incr('services.adzerk.fail')
+        return None
+    else:
+        statsd.incr('services.adzerk.success')
+    try:
+        return response.json()
+    except ValueError:
+        return {}
+
+
+def process_adzerk_result(decision):
+    contents = (decision.get('contents') or [{}])[0]
+    return {
+        'impression': urlparse(decision.get('impressionUrl', '')).query,
+        'click': urlparse(decision.get('clickUrl', '')).query,
+        'addon_id':
+            contents.get('data', {}).get('customData', {}).get('id', None)
+    }
+    return
+
+
+def process_adzerk_results(response, placeholders):
+    response_decisions = response.get('decisions', {})
+    decisions = [response_decisions.get(ph, {}) for ph in placeholders]
+    results_dict = {}
+    for decision in decisions:
+        result = process_adzerk_result(decision)
+        addon_id = str(result['addon_id'])
+        if addon_id in results_dict or not addon_id.isdigit():
+            continue  # no duplicates or weird/missing ids
+        results_dict[addon_id] = result
+    return results_dict
+
+
+def get_addons_from_adzerk(count):
+    placeholders = [f'div{i}' for i in range(count)]
+    response = call_adzerk_server(placeholders)
+    results_dict = (
+        process_adzerk_results(response, placeholders) if response else {})
+    log.debug(f'{results_dict=}')
+    # results_dict['16'] = list(results_dict.values())[0]  # TESTING HACK
+    return results_dict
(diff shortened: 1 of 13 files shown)


src/olympia/shelves/utils.py
```diff
@@ -0,0 +1,78 @@
...
+    except requests.exceptions.RequestException as e:
+        log.exception('Calling adzerk failed: %s', e)
+        statsd.incr('services.adzerk.fail')
+        return None
+    else:
+        statsd.incr('services.adzerk.success')
+    try:
+        return response.json()
+    except ValueError:
+        return {}
```
bobsilverberg, line 45: Should we log in this case so we can track if/when this happens?

bobsilverberg (approved):
Looks good to me. Nice work @eviljeff!

bobsilverberg:
@eviljeff In trying to write the code to consume this on the frontend, I've run into something. The way our code works with loading add-ons from an API response, it expects the `addon` in the response to contain the exact same set of fields as any other `addon` in any other response (e.g., a search). I knew we had something similar with collections, where the add-ons returned contained an extra field for `notes`. Looking at our code, and the API docs for [Collection Add-ons Detail](https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#collection-add-ons-detail), it looks like in that case we returned the `addon` data and the extra fields separately (i.e., each "addon" record actually looks like `{addon: addon, notes: notes}`).

It does feel a bit odd to be returning the extra fields separately from the rest of the data for the add-on, but I'm wondering if the reason we chose to do that at the time was the same as what I am now encountering.

I'm not really sure if this is the best way to address this. Another option is for me to add some special case code to deal with this new "type" of add-on, but that might not be particularly pretty. Of course, all of this has to do with Flow and our use of types.

I'm wondering what everyone thinks about this. Feedback is also welcome from @willdurand and @diox.

bobsilverberg:
Actually, nevermind. I can actually work around this quite easily by cheating a bit on the types.

eviljeff:
> @eviljeff In trying to write the code to consume this on the frontend, I've run into something. The way our code works with loading add-ons from an API response, it expects the `addon` in the response to contain the exact same set of fields as any other `addon` in any other response (e.g., a search). 

That's interesting because we subclass `AddonSerializer` and it's the serailizers used for it's fields and sub-fields in quite a few places to add & remove fields (notably, the serialized response for the addon /search and detail endpoints are slightly different).  Do we have to re-define the types in each place?

> I knew we had something similar with collections, where the add-ons returned contained an extra field for `notes`. Looking at our code, and the API docs for [Collection Add-ons Detail](https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#collection-add-ons-detail), it looks like in that case we returned the `addon` data and the extra fields separately (i.e., each "addon" record actually looks like `{addon: addon, notes: notes}`).
> 
> It does feel a bit odd to be returning the extra fields separately from the rest of the data for the add-on, but I'm wondering if the reason we chose to do that at the time was the same as what I am now encountering.

That may be more good luck than planning - the only reason the collection addon detail endpoint returns the addon and notes separately is just that that's how the `CollectionAddon` model is defined (the notes are a field on the model; addon is a foreign key), so I defaulted to writing the `CollectionAddonSerializer` the same way.  IIRC I added the api before the frontend work was complete.

> I'm not really sure if this is the best way to address this. Another option is for me to add some special case code to deal with this new "type" of add-on, but that might not be particularly pretty. Of course, all of this has to do with Flow and our use of types.

typing causing us extra work? We know the fix! 🤣

src/olympia/shelves/utils.py
eviljeff: https://github.com/mozilla/addons-server/pull/15670/commits/b06bd9ea4ba5671682a6637738e75fc48a569690

src/olympia/shelves/utils.py
```diff
@@ -0,0 +1,79 @@
...
+        results_dict[addon_id] = result
+    return results_dict
+
+
+def get_addons_from_adzerk(count):
+    placeholders = [f'div{i}' for i in range(count)]
+    response = call_adzerk_server(placeholders)
+    results_dict = (
+        process_adzerk_results(response, placeholders) if response else {})
+    log.debug(f'{results_dict=}')
```
diox, line 78: Do we want to keep this ?

src/olympia/shelves/views.py
```diff
@@ -1,11 +1,63 @@
...
+
+    def filter_queryset(self, qs):
+        qs = super().filter_queryset(qs)
+        count = self.paginator.get_page_size(self.request)
+        results = get_addons_from_adzerk(count)
+        ids = list(results.keys())
+        results_qs = qs.query(query.Bool(must=[
+            Q('terms', id=ids),
+            Q('term', **{'promoted.group_id': VERIFIED_ONE.id})]))
+        return results_qs
```
diox, line 55: Probably a question for product, but do we need to obey adzerk's ordering ? Or randomize ? In any case the default ES order is probably not what we want.

src/olympia/shelves/utils.py
eviljeff: I wasn't sure if we'd want to debug later, but I can remove.

src/olympia/shelves/views.py
eviljeff: I'm sure Jorge said we need to obey adzerk's ordering.  I forgot to add the code to do that to the ES query though.

diox (approved):
Could handle sorting in a follow-up

eviljeff:
> 
> 
> Could handle sorting in a follow-up

https://github.com/mozilla/addons-server/issues/15678

PR #15686, opened 2020-10-09
Replace email based checks for add-ons with permission checks

### LINKED ISSUE #15655: Replace email based checks for add-ons with permission checks
We have 3 different checks done at submission time that are tied to the email address of the user uploading the add-on:
- `system_addon_submission_allowed()`, which allows you to submit an add-on with a `guid` ending with a reserved suffix
- `mozilla_signed_extension_submission_allowed()`, which allows you to submit an add-on already signed with an internal Mozilla certificate
- `skip_trademark_check` inside `verify_mozilla_trademark()` which allows you to submit an add-on with a name that violates certain rules about trademarks around Mozilla and Firefox.

All these are based on the user having a @mozilla.com (or, for trademarks, @mozilla.org) email. We want to replace them with permission checks that we would then give to different groups of people that may need them.

diox:
- `SystemAddon:Submit` allows you to submit add-ons with reserved guids and xpis already signed with internal certificate
- `Trademark:Bypass` allows you to bypass trademark checks

Fixes #15655

### DIFF
--- a/src/olympia/devhub/tests/test_views.py
+++ b/src/olympia/devhub/tests/test_views.py
@@ -1380,9 +1380,9 @@ class TestUploadDetail(BaseUploadTest):
 
     @mock.patch('olympia.devhub.tasks.run_addons_linter')
     def test_system_addon_allowed(self, mock_validator):
-        user = user_factory(email='redpanda@mozilla.com')
+        user = user_factory()
         self.grant_permission(user, 'SystemAddon:Submit')
-        assert self.client.login(email='redpanda@mozilla.com')
+        assert self.client.login(email=user.email)
         mock_validator.return_value = json.dumps(self.validation_ok())
         self.upload_file(
             '../../../files/fixtures/files/mozilla_guid.xpi')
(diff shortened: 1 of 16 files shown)


src/olympia/devhub/tests/test_views.py
```diff
@@ -1380,7 +1380,8 @@ def test_experiment_xpi_not_allowed(self, mock_validator):
 
     @mock.patch('olympia.devhub.tasks.run_addons_linter')
     def test_system_addon_allowed(self, mock_validator):
-        user_factory(email='redpanda@mozilla.com')
+        user = user_factory(email='redpanda@mozilla.com')
```
bobsilverberg, line 1383: Would there be any value in using an email address that doesn't end with `mozilla.com`, just to prove that that is not what is allowing this test to pass?

src/olympia/devhub/tests/test_views.py
```diff
@@ -1414,8 +1415,9 @@ def test_system_addon_not_allowed_not_mozilla(self, mock_validator):
     @mock.patch('olympia.devhub.tasks.run_addons_linter')
     @mock.patch('olympia.files.utils.get_signer_organizational_unit_name')
     def test_mozilla_signed_allowed(self, mock_get_signature, mock_validator):
-        user_factory(email='redpanda@mozilla.com')
+        user = user_factory(email='redpanda@mozilla.com')
```
bobsilverberg, line 1418: As above, would it make sense to change the domain of the email address?

bobsilverberg (approved):
Looks good, nice job @diox! Just a couple of comments about tests.

r+wc

src/olympia/devhub/tests/test_views.py
diox: Yeah - I removed the email pretty much everywhere else to prove just that, but for some reason here I changed my strategy and kept the email but re-used it also in the test below where it's *not* allowed (because it doesn't have the permission). I guess being consistent and never having mozilla.com emails is best, I'll change it.

diox: 8c8f9a8a9f6980c7dc702160d08d47e2c32ee42e

src/olympia/devhub/tests/test_views.py
diox: 8c8f9a8a9f6980c7dc702160d08d47e2c32ee42e

PR #15687, opened 2020-10-09
implement impression api endpoint for sponsored shelf

### LINKED ISSUE #15618: Add proxy for Adzerk impression URLs
As mentioned in https://github.com/mozilla/addons-server/issues/15617 we'll want to proxy Adzerk's impression and click URLs through addons-server views we control, hiding client IPs from Adzerk.

eviljeff:
fixes #15618

### DIFF
--- a/docs/topics/api/shelves.rst
+++ b/docs/topics/api/shelves.rst
@@ -114,6 +114,6 @@ When the sponsored shelf is displayed for the user this endpoint can be used to 
 The current implemenation forwards these impression pings to Adzerk.
 
 
-.. http:post:: /api/v4/shelves/sponsored/
+.. http:post:: /api/v4/shelves/sponsored/impression/
 
-    :query string impression_data: the signed data payload that was sent in the :ref:`sponsored shelf <sponsored-shelf>` response.
+    :form string impression_data: the signed data payload that was sent in the :ref:`sponsored shelf <sponsored-shelf>` response.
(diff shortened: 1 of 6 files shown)


docs/topics/api/shelves.rst
```diff
@@ -101,4 +101,19 @@ Current implementation relies on Adzerk to determine which addons are returned a
...
+
+.. _sponsored-shelf-impression:
+
+When the sponsored shelf is displayed for the user this endpoint can be used to record the impressions.
+The current implemenation forwards these impression pings to Adzerk.
+
+
+.. http:post:: /api/v4/shelves/sponsored/
+
+    :query string impression_data: the signed data payload that was sent in the :ref:`sponsored shelf <sponsored-shelf>` response.
```
bobsilverberg, line 119: We're going to use the exact same endpoint for requesting shelf data, sending impression pings, and sending click pings, and just use the associated data to determine what type of API call it is? I suppose that's fine, but it seems a bit complicated from a caller perspective to me. Also, this data will be sent in a POST. Is it accurate to call this `:query`?

bobsilverberg (approved):
Looks good, @eviljeff. I just have a question about the endpoints, and also the doc.

r+wc

docs/topics/api/shelves.rst
eviljeff: copypasta fail 🙄 - should be `/shelves/sponsored/impression/'.  

And I guess it isn't `query` exactly - I'll look into how POST, but non-JSON, data is supposed to be specified.

eviljeff: https://github.com/mozilla/addons-server/pull/15687/commits/698b50d3d3d791a7ce33e9782f4aaca82b3009ba

PR #15691, opened 2020-10-09
Fix reviewer subscription

### LINKED ISSUE #15605: Allow reviewers to subscribe to unlisted version submissions
Reviewers can already subscribe to new listed submissions on the reviewer tools page. 

They should also have a way to subscribe to new unlisted submissions (if they have permissions to access the unlisted reviewer tools page).

wagnerand:
Fix #15605 (again).

I rewrote the entire `TestReviewerSubscription` class because the old one didn't take listed/unlisted into account consistently and was too hard to maintain (one of the reasons why I missed some edge-cases in the first try).

Note that I couldn't test this on my local instance since logging in is currently broken.

### DIFF
--- a/src/olympia/reviewers/models.py
+++ b/src/olympia/reviewers/models.py
@@ -427,16 +427,26 @@ def send_notifications(sender=None, instance=None, signal=None, **kw):
     if not subscribers:
         return
 
+    listed_perms = [
+        amo.permissions.ADDONS_REVIEW,
+        amo.permissions.ADDONS_CONTENT_REVIEW,
+        amo.permissions.ADDONS_RECOMMENDED_REVIEW,
+        amo.permissions.STATIC_THEMES_REVIEW,
+        amo.permissions.REVIEWER_TOOLS_VIEW
+    ]
+
     for subscriber in subscribers:
         user = subscriber.user
         is_active_user = user and not user.deleted and user.email
         is_reviewer_and_listed_submission = (
-            acl.is_user_any_kind_of_reviewer(user, allow_viewers=True) and
-            instance.channel == amo.RELEASE_CHANNEL_LISTED)
+            subscriber.channel == amo.RELEASE_CHANNEL_LISTED and
+            instance.channel == amo.RELEASE_CHANNEL_LISTED and
+            any(acl.action_allowed_user(user, perm) for perm in listed_perms))
         is_unlisted_reviewer_and_unlisted_submission = (
+            subscriber.channel == amo.RELEASE_CHANNEL_UNLISTED and
+            instance.channel == amo.RELEASE_CHANNEL_UNLISTED and
             acl.action_allowed_user(user,
-                                    amo.permissions.ADDONS_REVIEW_UNLISTED) and
-            instance.channel == amo.RELEASE_CHANNEL_UNLISTED)
+                                    amo.permissions.ADDONS_REVIEW_UNLISTED))
         if is_active_user and (
             is_reviewer_and_listed_submission or
             is_unlisted_reviewer_and_unlisted_submission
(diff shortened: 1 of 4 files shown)


src/olympia/reviewers/models.py
```diff
@@ -427,16 +427,26 @@ def send_notifications(sender=None, instance=None, signal=None, **kw):
     if not subscribers:
         return
 
+    listed_perms = [
+        amo.permissions.ADDONS_REVIEW,
+        amo.permissions.ADDONS_CONTENT_REVIEW,
+        amo.permissions.ADDONS_RECOMMENDED_REVIEW,
+        amo.permissions.STATIC_THEMES_REVIEW,
+        amo.permissions.REVIEWER_TOOLS_VIEW
+    ]
```
eviljeff, line 436: how does this differ from `is_user_any_kind_of_reviewer` - is it just it doesn't include unlisted review permission?

src/olympia/reviewers/templates/reviewers/review.html
```diff
@@ -223,21 +223,23 @@ <h3 id="history">
...
-      </li>
-      {% if action_allowed(amo.permissions.ADDONS_REVIEW_UNLISTED) %}
-      <li>
-        <input type="checkbox" id="notify_new_unlisted_versions"
-          data-api-url-subscribe-unlisted="{{ drf_url('reviewers-addon-subscribe-unlisted', addon.pk) }}"
-          data-api-url-unsubscribe-unlisted="{{ drf_url('reviewers-addon-unsubscribe-unlisted', addon.pk) }}"
-          {% if subscribed_unlisted %}checked="checked"{% endif %} autocomplete="off" />
-          <label for="notify_new_unlisted_versions">{{ _('Notify me about new unlisted versions') }}</label>
-      </li>
+        {% if acl_is_reviewer or action_allowed(amo.permissions.REVIEWER_TOOLS_VIEW) %}
```
eviljeff, line 226: can a user who isn't `acl_is_reviewer` even get to this page?

src/olympia/reviewers/tests/test_models.py
```diff
@@ -386,98 +386,176 @@ class TestReviewerSubscription(TestCase):
...
-            name='Reviewers: Unlisted', rules='Addons:ReviewUnlisted')
-        GroupUser.objects.create(
-            group=self.reviewer_group, user=self.user_one)
-        version_uploaded.send(sender=Version, instance=self.version)
-        assert len(mail.outbox) == 2
-        assert mail.outbox[0].to == [u'del@icio.us']
+    def test_listed_subscription(self):
+        version_uploaded.send(sender=Version, instance=self.listed_version)
+        assert len(mail.outbox) == 1
+        assert mail.outbox[0].to == [u'listed@reviewer']
```
eviljeff, line 459: `u'`  :whatyearisthis:

eviljeff (approved):

src/olympia/reviewers/tests/test_models.py
wagnerand: I took that from the old test. I guess it's not needed anymore?

src/olympia/reviewers/templates/reviewers/review.html
wagnerand: Maybe not, but we need some way to limit this part to listed reviewers only, since technically, unlisted and listed reviewers are distinct.

src/olympia/reviewers/models.py
wagnerand: Correct.

wagnerand: I did consider writing a new acl helper method, but it would have been a super generic one and so far we'd only need it for this particular use-case. I also can't re-use much of what's already there since I need to use methods that don't rely on a `request`.

src/olympia/reviewers/tests/test_models.py
eviljeff: we've not bothered to strip it out of all the old code, but it's jarring to see it in new code.  That said, meh, I wouldn't bother with a new commit _just_ for that.

wagnerand: Not a big deal, I will remove it.

Issue #9080, opened 2018-08-03
Change local nginx to route appropriate requests to localhost:3000 (frontend)


eviljeff:
`http://olympia.test/` is always legacy frontend right now, which is bad already because deprecated 😞 but will be extra bad once we remove all the legacy views 😨.  We should change it to do the appropriate routing for frontend pages to localhost:3000 instead, mirroring what's done on prod(+stage/dev).

~Note: potential blocker on how we'd do the reverting to legacy frontend switch - would we have to do the cookie thing too?  Can we locally?~

(Bonus if we can move all the routing logic from the ops deployment files into addons-server?)

EnTeQuAk:
I got this working after some fiddling, the only problem I am running into right now is that at least the start page requires quite a lot of add-ons that already exist on addons-server.

The ui-tests already have some code for it, I guess we may just re-use that?

eviljeff:
landfill has some addon generation code - is that not enough?  Otherwise I suppose it's fine to move some code from the ui-tests to `src/olympia/` to do it.

EnTeQuAk:
> landfill has some addon generation code - is that not enough? 

No, since the startpage from addons-frontend requires a certain set of existing add-ons, similar to discopane.

PR #15702, opened 2020-10-12
Add special URL to bypass authentication locally

### LINKED ISSUE #15701: Provide user creation and authentication methods that do not depend on FxA for local environments
Our local addons-server environments depend on FxA for user creation and authentication flow just like dev/stage/prod does, except it uses https://stable.dev.lcip.org/. It's useful to have the exact same flow when implementing things around authentication, but most of the time it gets in the way as it prevents full offline development. It's also annoying when https://stable.dev.lcip.org/ is down or broken (like it is right now, https://jira.mozilla.com/browse/FXA-2595)

We need to:
- Stop calling FxA during `make initialize`. This was added for the ui tests but they should use an explicit, different call for this.
- Provide an alternative way of creating users locally (when `DEBUG` is `True`)
- Provide an alternative way of logging in as a user locally (when `DEBUG` is `True`)

diox:
Fixes #15701


Makefile-docker
```diff
@@ -87,7 +87,12 @@ populate_data: ## populate a new database
 	# These add-ons are specifically useful for the addons-frontend
 	# homepage. You may have to re-run this, in case the data there
 	# changes.
-	$(PYTHON_COMMAND) manage.py generate_default_addons_for_frontend
+	@if test "$(UITEST_FXA_EMAIL)" != ""; then \
+		echo 'UITEST_FXA_EMAIL is present, running generate_default_addons_for_frontend';\
+        $(PYTHON_COMMAND) manage.py generate_default_addons_for_frontend;\
+    else\
+		echo 'No UITEST_FXA_EMAIL, skipping generate_default_addons_for_frontend';\
+    fi
```
eviljeff, line 96: weird indentation here

eviljeff (commented):
I was hoping this could be more transparent tbh - like it would kick in by default with `settings.DEBUG == True` unless a setting/waffle was enabled to specifically use real FxA.  But it's better than the current roadblock.

Makefile-docker
```diff
@@ -87,7 +87,12 @@ populate_data: ## populate a new database
 	# These add-ons are specifically useful for the addons-frontend
 	# homepage. You may have to re-run this, in case the data there
 	# changes.
-	$(PYTHON_COMMAND) manage.py generate_default_addons_for_frontend
+	@if test "$(UITEST_FXA_EMAIL)" != ""; then \
+		echo 'UITEST_FXA_EMAIL is present, running generate_default_addons_for_frontend';\
+        $(PYTHON_COMMAND) manage.py generate_default_addons_for_frontend;\
```
eviljeff, line 92: wait... don't we _need_ this or the local frontend is entirely empty so fails?  It's the thing that populates the database/ES with enough addons.

eviljeff: It's just the fxa account creation part that can be skipped.

diox: We already generate a bunch of add-ons and themes a few lines above this one, so I'd say no ?

diox: That was added for https://github.com/mozilla/addons-server/issues/9080 ... but I doubt it's still up to date anyway, so I don't know ? I must admit I usually don't particularly care about the homepage content on my local env...

eviljeff: it's the command that creates the promoted addons and collections.  IIRC frontend fails badly when nothing is available.  I don't care that much about the homepage being fully populated, though I feel if it raises an error and doesn't work at all we've just moved where a contributor would be blocked to a different part in the process.

Can we just bypass the part of the command that creates the user on FxA instead?  (skip `self.fxa_id = self._create_fxa_user()` if `UITEST_FXA_EMAIL` is missing or something)?

diox: Yeah, I initially started with that and then starting to question why we even called this command in the first place... I'll go back to having the condition in the command.

diox:
> I was hoping this could be more transparent tbh - like it would kick in by default with `settings.DEBUG == True` unless a setting/waffle was enabled to specifically use real FxA. But it's better than the current roadblock.

Had not thought of making it that transparent, but I see the appeal... I'll give it a try.

src/olympia/accounts/utils.py
```diff
@@ -71,8 +73,12 @@ def fxa_login_url(config, state, next_path=None, action=None,
...
             query['prompt'] = 'none'
             query['id_token_hint'] = id_token
-    return '{host}/authorization?{query}'.format(
-        host=settings.FXA_OAUTH_HOST, query=urlencode(query))
+    if use_fake_fxa():
+        base_url = reverse('fake-fxa-authorization')
+    else:
+        base_url = '{host}/authorization'.format(host=settings.FXA_OAUTH_HOST)
+    return '{base_url}?{query}'.format(
+        base_url=base_url, query=urlencode(query))
```
diox, line 81: Not super pretty but the reason I'm doing this instead of overriding `FXA_CONFIG` for local envs is that I wanted to make it work regardless of the `config` used, and also make it easy to switch back to the real config if needed.

src/olympia/amo/templates/amo/fake_fxa_authorization.html
```diff
@@ -0,0 +1,13 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h2> {{ _('Register or Log in') }} </h2>
+  <form id="fake_fxa_authorization" method="get" action="{{ url('auth:accounts.authenticate') }}">
+      {% for key, value in request.GET.items() %}
+        <input type="hidden" name="{{ key }}" value="{{ value }}" />
+      {% endfor %}
+      <input type="hidden" name="code" value="fakecode" />
```
eviljeff, line 9: what's `fakecode` for?

eviljeff (approved):

src/olympia/amo/templates/amo/fake_fxa_authorization.html
diox: The authenticate callback view checks that a `code` is passed, we never enter the function that uses its value but it needs to be present in the query for it to work. I didn't want to add yet another check to bypass that.

PR #15711, opened 2020-10-13
Change UserProfile.is_staff to be True for all mozilla.com emails

### LINKED ISSUE #15601: Change UserProfile.is_staff to be True for all mozilla.com emails
`is_staff` is a `UserProfile` property django uses to gate login to the django admin. It doesn't give any particular access, if you don't have any permissions but that property returns `True`, then you can log in to the admin but you'd just see an empty admin (provided you can even reach the page in the first place, since we have nginx rules that force you to be part of a specific VPN group for the admin to work).

Typically it's implemented as a column in the database, but in our case it's a `@property` returning`acl.action_allowed_user(self, amo.permissions.ANY_ADMIN)` - essentially returning `True` as long as the user has any `Admin:<Something>` permission.

I'd like to change it to be `True` for all mozilla employees - anyone with a `mozilla.com` email. We'd then be able to use waffle flag's `staff` property, which enables a given flag for all `staff` members, and be able to enable a flag for all mozilla employees.

diox:
Remove `Admin:Tools` in the process as it was only really used to ensure a given user could log in to the admin.

Fixes #15601

### DIFF
--- a/src/olympia/users/models.py
+++ b/src/olympia/users/models.py
@@ -223,12 +223,21 @@ class UserProfile(OnChangeMixin, ModelBase, AbstractBaseUser):
 
     @property
     def is_staff(self):
-        """Property indicating whether the user should be able to log in to
-        the django admin tools. Does not guarantee that the user will then
-        be able to do anything, as each module can have its own permission
-        checks. (see has_module_perms() and has_perm())"""
-        from olympia.access import acl
-        return acl.action_allowed_user(self, amo.permissions.ANY_ADMIN)
+        """Property indicating whether the user is considered to be a Mozilla
+        Employee.
+
+        Django admin uses this to allow logging in, though it doesn't give
+        access to the individual admin pages: each module has their own
+        permission checks (see has_module_perms() and has_perm() below). In
+        addition we also force users to use the VPN to access the admin.
+
+        It's also used by waffle Flag `staff` property, which allows a feature
+        behind a flag to be enabled just for users for which this property
+        returns True. This shouldn't be used as a replacement to a permission
+        check, but only for progressive rollouts of features that are intended
+        to eventually be enabled globally.
+        """
+        return self.email and self.email.endswith('@mozilla.com')
 
     def has_perm(self, perm, obj=None):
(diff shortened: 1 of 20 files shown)


src/olympia/users/models.py
```diff
@@ -223,12 +223,21 @@ def is_superuser(self):
...
+        permission checks (see has_module_perms() and has_perm() below). In
+        addition we also force users to use the VPN to access the admin.
+
+        It's also used by waffle Flag `staff` property, which allows a feature
+        behind a flag to be enabled just for users for which this property
+        returns True. This shouldn't be used as a replacement to a permission
+        check, but only for progressive rollouts of features that are intended
+        to eventually be enabled globally.
+        """
+        return self.email and self.email.endswith('@mozilla.com')
```
wagnerand, line 240: Do we want to include MoFo (`@mozillafoundation.org`)?

diox: I'd say no, the more limited the better. `is_staff` use cases are very limited anyway.

src/olympia/users/tests/test_models.py
```diff
@@ -1111,7 +1126,7 @@ def test_create_superuser(self):
         )
         assert user.pk is not None
         Group.objects.get(name="Admins") in user.groups.all()
-        assert user.is_staff
+        assert not user.is_staff  # Not a mozilla.com email...
```
eviljeff, line 1129: so... if you don't have a @mozilla.com email address you can't be a superuser, even locally?  Doesn't that lock out any contribution that involves django admin?

eviljeff (commented):
I wonder if we shouldn't include full superusers as `is_staff`, at least locally.

src/olympia/users/tests/test_models.py
diox: If you don't have a mozilla.com email you can be a superuser, but you won't have `is_staff` so you won't have access to the admin. With https://github.com/mozilla/addons-server/issues/15701 you can easily create a mozilla.com user to gain access the admin locally, so I don't think it's a big problem.

eviljeff (approved):

PR #15731, opened 2020-10-14
Add a `PromotedSubscription` model

### LINKED ISSUE #15723: Add a django admin to onboard candidates
From the PRD:

> The admin enters the ID of the add-on that will be onboarded and the Promoted group (Verified or Sponsored), and that generates an AMO URL that will be sent to the developer to set up payment.

For QA: this should allow admins (or users with the `Discovery:Edit` permission) to create subscriptions for an add-on and a promoted group. The generated link will point to nowhere until we close #15726.

Note: this admin will be inlined in the existing promoted admin and only revealed (in read only mode) once a promoted addon has been created (verified or sponsored only).

### LINKED ISSUE #15724: Add a model to store onboarding/payment/subscription information
We need a model/db table to store information about payment for promoted add-ons.

willdurand:
Fixes #15724


willdurand:
@diox I updated this patch with the model only. I commented in the issue to explain why we want to use `PromotedAddon` instead of `Addon`. I'll create a PR on top of this one with the admin, which is a bit simpler.

willdurand:
See: https://github.com/mozilla/addons-server/pull/15750

src/olympia/promoted/models.py
```diff
@@ -127,3 +129,47 @@ def update_es_for_promoted(sender, instance, **kw):
...
+        null=True,
+        help_text=(
+            "This date is set when the developer has visited the onboarding "
+            "page.",
+        ),
+    )
+    stripe_session_id = models.CharField(
+        default=None, null=True, max_length=100
+    )
+    payment_cancelled_at = models.DateTimeField(
```
diox, line 150: Both `DateTimeField`s are probably a bit too simplistic in the long run - we'll have to define exactly what happens when developers cancel and come back, for instance but I reckon this is enough to get us started. (And we might not need to store too much ourselves, ultimately Stripe does have that data available through their API)

diox (approved):

src/olympia/promoted/models.py
willdurand: Yes, I agree.

According to the PRD, we want to offer the developer to pay again (and again and again..) when they cancel on Stripe Checkout (which is only possible with the "back" arrow/button). It's a bit different than cancelling the subscription after the first payment, but this is not covered by the PRD.

PR #15739, opened 2020-10-15
changes local dev and stage adzerk site ids

### LINKED ISSUE #15738: Define seperate siteid for dev & stage (& local dev) for adzerk
we currently use the same site_id in our settings for all env (though it can be overridden by ops in the env files).  To be able to accurately test and develop for different platforms it be best to define specific site ids (and flights that are targetted to those site ids)

eviljeff:
fixes #15738

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1928,7 +1928,7 @@ BIGQUERY_AMO_DATASET = 'amo_dev'
 
 ADZERK_TIMEOUT = 5  # seconds
 ADZERK_NETWORK_ID = env('ADZERK_NETWORK_ID', default=10521)
-ADZERK_SITE_ID = env('ADZERK_SITE_ID', default=1131244)
+ADZERK_SITE_ID = env('ADZERK_SITE_ID', default=1133496)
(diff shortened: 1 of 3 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -1928,7 +1928,7 @@ def get_sentry_release():
 
 ADZERK_TIMEOUT = 5  # seconds
 ADZERK_NETWORK_ID = env('ADZERK_NETWORK_ID', default=10521)
-ADZERK_SITE_ID = env('ADZERK_SITE_ID', default=1131244)
+ADZERK_SITE_ID = env('ADZERK_SITE_ID', default=1133496)
```
bobsilverberg, line 1931: What does this default site id correspond to?

bobsilverberg (approved):
Looks good, thanks @eviljeff. Just a question about the default site id.

r+wc

src/olympia/lib/settings_base.py
eviljeff: local dev

eviljeff: once this is live on dev we can delete all the non-dev creatives from https://app.adzerk.com/#!/flight/12435798/edit/ - I set up separate flights for dev, stage and local dev and targetted the different sites at each.

PR #15742, opened 2020-10-15
Support all badged add-ons in Return To AMO

### LINKED ISSUE #15732: Support RTAMO for all badged add-ons
RTAMO now works with add-ons that have a `DiscoveryItem`, but we want it to work will all badged add-ons (Recommended, By Firefox, Verified, Sponsored). RTAMO is being brought back to Firefox soon. Related bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1670908.

diox:
Slight behavior change: we now avoid the database query and return directly some results or an empty list if the parameter is not correct. Callers should always have expected the empty list, because the add-on could be in DiscoveryItems without necessarily being public at the time of the request.

Fixes #15732

### DIFF
--- a/src/olympia/search/filters.py
+++ b/src/olympia/search/filters.py
@@ -16,7 +16,6 @@ from olympia.api.utils import is_gate_active
 from olympia.constants.categories import CATEGORIES, CATEGORIES_BY_ID
 from olympia.constants.promoted import (
     BADGED_API_NAME, PROMOTED_API_NAME_TO_IDS, PROMOTED_GROUPS)
-from olympia.discovery.models import DiscoveryItem
 from olympia.versions.compare import version_int
 
 
--- a/src/olympia/search/filters.py
+++ b/src/olympia/search/filters.py
@@ -28,11 +28,11 @@ class AddonQueryParam(object):
     valid_values = None
     es_field = None
 
-    def __init__(self, request):
-        self.request = request
+    def __init__(self, query_data):
+        self.query_data = query_data
 
     def get_value(self):
-        value = self.request.GET.get(self.query_param, '')
+        value = self.query_data.get(self.query_param, '')
         try:
             # Try the int first.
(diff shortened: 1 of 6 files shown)


src/olympia/search/filters.py
```diff
@@ -212,20 +213,19 @@ def get_values(self):
...
-                    ugettext(
-                        'Invalid Return To AMO guid (not a curated add-on)'
-                    )
-                )
-            return [value] if value else []
+            # Filter on the now decoded guid param as normal, then add promoted
+            # filter on top to only return "safe" add-ons for return to AMO.
+            # We don't care about the app param - we just want to ensure the
+            # add-ons are "safe".
+            filters = super().get_es_query([value])
```
eviljeff, line 220: mmm, did leaving all the decoding `rta:...` logic in `get_values` not work?

eviljeff (approved):
interesting changes

src/olympia/search/filters.py
diox: It did, but `get_values()` cannot return the additional filter for promoted add-ons, so I would have had to implement that elsewhere anyway. I figure I might as well override the whole `get_es_query()` then. I'm open to better ways of doing this, didn't want to refactor too far without a better plan...

eviljeff: I don't know if there's a "more correct" way, I just saw you had to change `AddonQueryMultiParam.get_es_query` to make it work because the code was moved from `get_values` to this method and I was curious why.

diox: Ah yeah, I still had to modify the `get_es_query()` method to allow an arbitrary value to be passed.

One alternative way of doing this would be to override `get_values()` to do the decoding, store a flag indicating that we had to decode something, and add the promoted condition in `get_es_query()` if that flag is present. I thought it wasn't that much better.

PR #15749, opened 2020-10-16
pin celery & it's dependants to < 5.0

### LINKED ISSUE #15677: Stick to Celery 4.x LTS
Celery 5 has been released, and the 4.x branch has been designated a LTS https://docs.celeryproject.org/en/stable/whatsnew-5.0.html#long-term-support-policy. Celery 5 has breaking changes from 4.x, like most releases (and some point releases), so it will be easier, at least in the short term, to stick with Celery 4.x. 

The docs state the 4.x branch will be supported until August 2021 (by which time I guess the next release of celery, 6, will be available? - it's not clear they have a scheduled release policy). Note, as their LTS policy seems a new thing too, we will have to wait to see how security/bug fixes are backported from the 5.x release branch.  There are also dependent packages (kombu and vine) that we wouldn't be able to upgrade - they also managed by the celery project so although they don't have a LTS policy specifically we can probably assume they would have a 4.x fix if required.

eviljeff:
fixes #15677 closes #15419 closes #15420 closes #15573

Billiard is called out as another celery dependant that was previously pinned due to celery versions, but it doesn't seem to have been updated for a while so I'm not sure if we need to remove the version pinning, or change it to <5.0 like the others. @diox

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,10 @@ updates:
   ignore:
   - dependency-name: billiard
     versions:
-    - ">= 4.0"
+    - ">= 5"
+  - dependency-name: celery
+    versions:
+    - ">= 5"
   - dependency-name: Django
     versions:
     - ">= 3.0"
@@ -21,6 +24,12 @@ updates:
   - dependency-name: elasticsearch-dsl
     versions:
     - ">= 7"
+  - dependency-name: kombu
+    versions:
+    - ">= 5"
   - dependency-name: pygit2
     versions:
     - ">= 1.2.0"
+  - dependency-name: vine
+    versions:
+    - ">= 5"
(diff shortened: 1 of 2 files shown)


willdurand (approved):

diox:
We aren't using `pyup` anymore, you need to do that in `.github/dependabot.yml` using `ignore` directives to ignore versions above a specific threshold.

eviljeff:
> We aren't using `pyup` anymore, you need to do that in `.github/dependabot.yml` using `ignore` directives to ignore versions above a specific threshold.

oh, heh, I should have remembered that!  (we should remove all the pyup comments at some point I guess?)

eviljeff:
force pushed an amendment commit (merge conflicts)

willdurand (approved):
sprry, I forgot about dependabot too.

PR #15750, opened 2020-10-16
Add admin for promoted subscriptions

### LINKED ISSUE #15723: Add a django admin to onboard candidates

willdurand:
Fixes #15723

---

We refined the onboarding flow with Product and here it is:

> (1) developer requests to be added by a sign-up form, (2) we choose the next
> add-on to be added (from the sign-up list) and pre-review the add-on to make
> sure it meets our standards, (3) if the add-on passes we add it to the
> promoted group and email the dev with the payment link, (4) if that succeeds,
> then next version approval should finalize the process and show the badge

As we can see in (3), admins will add an add-on to the promoted group before
sending the email to the developer so it makes sense to inline the subscription
admin in the existing promoted add-on admin: there is no reason to have a
different admin for subscriptions.

### Screenshots

When creating a new promoted add-on, we don't show the subscription info:

Screenshot 2020-10-16 at 12 31 19

When editing an existing promoted add-on that does not require payment, we don't show the subscription info:

Screenshot 2020-10-16 at 12 38 08

When editing an existing promoted add-on that does require payment, we show the subscription info:

Screenshot 2020-10-16 at 12 37 14

### DIFF
--- a/src/olympia/addons/tests/test_models.py
+++ b/src/olympia/addons/tests/test_models.py
@@ -25,7 +25,7 @@ from olympia.bandwagon.models import Collection
 from olympia.blocklist.models import Block, BlocklistSubmission
 from olympia.constants.categories import CATEGORIES
 from olympia.constants.promoted import (
-    NOT_PROMOTED, RECOMMENDED, SPOTLIGHT, VERIFIED_ONE)
+    NOT_PROMOTED, RECOMMENDED, SPOTLIGHT, SPONSORED)
 from olympia.devhub.models import RssKey
 from olympia.files.models import File
 from olympia.files.tests.test_models import UploadTest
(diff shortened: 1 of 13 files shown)


willdurand:
I just added a second commit that renames the `VERIFIED_*` constants. I thought about opening a different PR for that but the diff is small so 🤷

willdurand:
(I opened a draft PR with the Stripe integration built on top of this patch: https://github.com/mozilla/addons-server/pull/15753)

diox (approved):

PR #15751, opened 2020-10-16
Update link to ratings in reviewer tools for disabled addons.

### LINKED ISSUE #10113: Update link to ratings in reviewer tools

The current link for ratings in the reviewer tools points to the public ratings page, which is only available if the add-on itself is public. If it's disabled, it will respond with 404.

We also have an admin ratings page (`/admin/models/ratings/rating`) that will show ratings even for disabled add-ons, but only for admins.

What we should probably do is only linkify the text `X reviews` if the user has the `Ratings:Moderate` ~and `Admin:Tools`~ permission, pointing to `/admin/models/ratings/rating/?addon=<id>`. For users that don't have that permission (ie regular reviewers, they would still see that text, but without a link.

If that is too much effort, we can linkify the text always, but then should add a suffix like `(admin only)` so non-admins will be aware that the link is not supposed to work for them.

VishalCR7:
Fixes #10113 

If the addon is public, `X reviews` points to the public ratings page
If the addon is not public and user has the `Ratings:Moderate` permission, pointing to /admin/models/ratings/rating/?addon=<id>. For users that don't have that permission (ie regular reviewers, they would still see that text, but without a link.

      PRs open for the same issue.

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/addon_details_box.html
+++ b/src/olympia/reviewers/templates/reviewers/addon_details_box.html
@@ -95,15 +95,12 @@
                   {{ addon.average_rating|float|stars }}
                   {% with count='<span itemprop="count">{0}</span>'|format_html(num|numberfmt) %}
                     {# Using num=count so we don't change an L10n string. #}
+                    {% set review_url_text = ngettext('{num} review', '{num} reviews', num)|format_html(num=count) %}
                     {% if acl_is_review_moderator %}
                       {% set review_url = url('admin:ratings_rating_changelist')|urlparams(addon=addon.pk) %}
-                      <a href="{{ review_url|absolutify }}"><strong>{{
-                        ngettext('{num} review', '{num} reviews', num)|format_html(num=count)
-                      }}</strong></a>
+                      <a href="{{ review_url|absolutify }}"><strong>{{ review_url_text }}</strong></a>
                     {% else %}
-                      <strong>{{
-                        ngettext('{num} review', '{num} reviews', num)|format_html(num=count)
-                      }}</strong>
+                      <strong>{{ review_url_text }}</strong>
                     {% endif %}
                   {% endwith %}
                   {% else %}
--- a/src/olympia/reviewers/templates/reviewers/addon_details_box.html
+++ b/src/olympia/reviewers/templates/reviewers/addon_details_box.html
@@ -106,7 +106,7 @@
                     {% else %}
                       <strong>{{
                         ngettext('{num} review', '{num} reviews', num)|format_html(num=count)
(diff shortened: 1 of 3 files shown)


VishalCR7:
@eviljeff Please review

src/olympia/reviewers/templates/reviewers/addon_details_box.html
```diff
@@ -88,7 +88,33 @@ <h2>{{ addon_name }}</h2>
...
+            <td>
+              <p class="addon-rating">
+                {% with num=addon.total_ratings %}
+                  {% if num %}
+                  {{ addon.average_rating|float|stars }}
+                  {% with count='<span itemprop="count">{0}</span>'|format_html(num|numberfmt) %}
+                    {# Using num=count so we don't change an L10n string. #}
+                    {% if not addon.is_disabled or action_allowed(amo.permissions.RATINGS_MODERATE) %}
+                      {% set review_url = url('addons.ratings.list', addon.slug) %}
+                      {% if addon.is_disabled %}
```
eviljeff, line 100: we want this always when the user has the permission

eviljeff: see the original comment in the issue - the two alternatives are having the permission and not.

VishalCR7: We want to only linkify if the user has permission `Ratings:Moderate`, otherwise leave it unlinked. Is it right ?

VishalCR7:
@eviljeff  Some tests are failing after recent updates. Having a hard time to figure out what is wrong looking at those long stack trace. Can you please help me identify the issue here if possible

eviljeff (commented):
we also need some tests that prove your changes work.

`TestReview.test_user_ratings` can be added to, perhaps, to check the link is there.  And there should be a new test to check the link _isn't_ there when the user doens't have the correct permission.

eviljeff (approved):
This is looking good, r+

eviljeff:
@wagnerand before I merge this, is it really only `Ratings:Moderate` permission that's needed to access the admin tool?  (That's what the issue says now, but no special admin perm/checks?)

wagnerand:
@eviljeff I believe so (https://github.com/mozilla/addons-server/blob/master/src/olympia/constants/permissions.py#L215:L215), unless I am missing something?

eviljeff:
> @eviljeff I believe so (https://github.com/mozilla/addons-server/blob/master/src/olympia/constants/permissions.py#L215:L215), unless I am missing something?

I thought maybe a check that the user was able to actually access django admin (`is_staff`?) was needed too but okay

PR #15753, opened 2020-10-16
Implement payment flow using Stripe Checkout

### LINKED ISSUE #15725: Add logic to create and retrieve a Stripe checkout session
This is needed for the payment integration and subscription flow.

### LINKED ISSUE #15726: Implement payment flow using Stripe Checkout
The PRD has more info about the exact flow: https://docs.google.com/document/d/1cohRp8JRlv7h41I0MxFUK1oBk70th71Fa7g4sGruzpM/edit#

In short, we have a landing page with a big button to go to Stripe. Once payment has been made, we return to a similar page. In case of an error, it should be possible to go to Stripe again (and there will be an error message displayed).

When the payment is successful, there are changes to make to the add-on but this will be tackled in a different issue. This issue is about implementing the views to have a working payment integration.

willdurand:
Fixes #15725
Fixes #15726 

---

This patch provides the integration with Stripe Checkout. Once a "promoted subscription" is created in the admin, we can use the link generated and start the payment flow.

At the end of the flow, the subscription is active in Stripe but we didn't update the current add-on version. This will be done in a different patch.

The content might change too so that's why assertions are pretty basic. The idea is to have the flow implemented so that we can adjust things with Product.

## Screenshots

The link generated in the admin points to this page:

Clicking the "continue to stripe checkout" button redirects the user to this page (Stripe Checkout, not AMO domain anymore):

Clicking the left arrow that we can see above redirects to AMO:

Clicking the "continue to stripe checkout" button again, that time with valid card:

And now Stripe Checkout redirects to AMO and we see this:

Screenshot 2020-10-20 at 18 04 02


src/olympia/promoted/utils.py
```diff
@@ -0,0 +1,49 @@
...
+from olympia.amo.templatetags.jinja_helpers import absolutify
+from olympia.amo.urlresolvers import reverse
+from olympia.constants.promoted import SPONSORED, VERIFIED
+
+
+def create_or_retrieve_stripe_checkout_session(subscription, customer_email):
+    if subscription.stripe_session_id:
+        return retrieve_stripe_checkout_session(subscription)
+
+    price_id = {
```
bobsilverberg, line 14: I wonder if it makes sense to have this dict as a constant somewhere? I think similar things exist for other pieces of data.

src/olympia/promoted/utils.py
```diff
@@ -0,0 +1,49 @@
...
+from django.conf import settings
+
+from olympia.amo.templatetags.jinja_helpers import absolutify
+from olympia.amo.urlresolvers import reverse
+from olympia.constants.promoted import SPONSORED, VERIFIED
+
+
+def create_or_retrieve_stripe_checkout_session(subscription, customer_email):
+    if subscription.stripe_session_id:
+        return retrieve_stripe_checkout_session(subscription)
```
bobsilverberg, line 12: What happens if this doesn't return a session? I.e., if `stripe.checkout.Session.retrieve(subscription.stripe_session_id)` doesn't return a session? Would we want to go ahead and create a new one in that case?

src/olympia/promoted/utils.py
```diff
@@ -0,0 +1,49 @@
...
+            )
+        ),
+        line_items=[{"price": price_id, "quantity": 1}],
+        customer_email=customer_email,
+    )
+
+
+def retrieve_stripe_checkout_session(subscription):
+    stripe.api_key = settings.STRIPE_API_SECRET_KEY
+    return stripe.checkout.Session.retrieve(subscription.stripe_session_id)
```
bobsilverberg, line 49: What happens if a session isn't found? Might it cause an exception, or just return `None`, or something else?

src/olympia/devhub/templates/devhub/addons/onboarding_subscription.html
```diff
@@ -0,0 +1,79 @@
+{% extends "devhub/base_impala.html" %}
+
+{% set title = _('Set Up Payment') %}
+
+{% block js %}
+{{ super() }}
+<script src="https://js.stripe.com/v3/"></script>
```
bobsilverberg, line 7: Just wondering why this is here, rather than in our `stripe.js`?

src/olympia/promoted/tests/test_utils.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+    sub = PromotedSubscription.objects.filter(
+        promoted_addon=promoted_addon
+    ).get()
+    customer_email = "some-email@example.org"
+
+    with mock.patch(
+        "olympia.promoted.utils.stripe.checkout.Session.create"
+    ) as stripe_create:
+        stripe_create.return_value = "fake session"
```
bobsilverberg, line 44: Nit: "fake session" could be in a const to avoid hardcoding it twice.

src/olympia/promoted/models.py
```diff
@@ -177,11 +178,20 @@ class PromotedSubscription(ModelBase):
...
-        )
+
+        url = reverse('devhub.addons.onboarding_subscription',
+                      args=[self.promoted_addon.addon.slug])
+        if absolute:
+            url = urljoin(settings.EXTERNAL_SITE_URL, url)
+        return url
+
+    @property
+    def stripe_checkout_completed(self):
```
bobsilverberg, line 192: Should we have unit tests for these properties?

src/olympia/devhub/views.py
```diff
@@ -1837,3 +1842,101 @@ def logout(request):
...
+    FRAME_SRC=["https://js.stripe.com", "https://hooks.stripe.com"],
+)
+def onboarding_subscription(request, addon_id, addon):
+    sub = get_promoted_subscription_or_404(addon=addon)
+
+    fields_to_update = {}
+    if addon.has_author(request.user) and not sub.link_visited_at:
+        fields_to_update['link_visited_at'] = datetime.datetime.now()
+
+    session = create_or_retrieve_stripe_checkout_session(
```
bobsilverberg, line 1872: It looks like `create_or_retrieve_stripe_checkout_session` can throw an exception. What happens in this case?

src/olympia/devhub/tests/test_views_subscription.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+class TestOnboardingSubscription(OnboardingSubscriptionTestCase):
+    url_name = "devhub.addons.onboarding_subscription"
+
+    @override_switch("enable-subscription", active=False)
+    def test_returns_404_when_switch_is_disabled(self):
+        assert self.client.get(self.url).status_code == 404
+
+    def test_returns_404_when_subscription_is_not_found(self):
```
bobsilverberg, line 36: Should we have a test where the subscription has a `stripe_session_id`, but `stripe.checkout.Session.retrieve()` cannot find it? It seems like we are assuming that a `stripe_session_id` will always correspond to a valid session, but isn't it possible it might not?

bobsilverberg (commented):
This looks good, nice work @willdurand! I've added a bunch of comments, mostly small, but I do have one concern about a case I think we're not catching.

src/olympia/devhub/urls.py
```diff
@@ -1,4 +1,5 @@
-from django.urls import include, re_path
+from django.conf.urls import include
```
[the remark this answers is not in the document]
willdurand: ha ok, good catch. I didnt see it during the rebase.

src/olympia/promoted/utils.py
willdurand: it's only useful in this file so I don't know. Later, it could be moved to a function, though

src/olympia/promoted/utils.py
willdurand: It should throw an error but I'll check again

src/olympia/promoted/utils.py
willdurand: Exception

src/olympia/devhub/templates/devhub/addons/onboarding_subscription.html
willdurand: this requires a js file from stripe

src/olympia/devhub/views.py
willdurand: 500 :p

src/olympia/devhub/tests/test_views_subscription.py
willdurand: maybe yes, although the current behavior is to return a 500. Maybe we should create a new session indeed.

src/olympia/promoted/utils.py
willdurand: So I read the stripe docs again and they say:

> We recommend creating a new Session each time your customer attempts to pay.

I didn't see this the first time so I guess we should only have a `create_stripe_checkout_session` here.

src/olympia/devhub/templates/devhub/addons/onboarding_subscription.html
diox: And stripe themselves ask you specifically require their js from their site rather than copying it into your bundle

src/olympia/promoted/migrations/0011_create_switch_for_subscription.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 2.2.16 on 2020-10-16 14:48
+
+from django.db import migrations
+
+
+def create_waffle_switch(apps, schema_editor):
+    Switch = apps.get_model("waffle", "Switch")
+
+    Switch.objects.create(
+        name="enable-subscription",
```
diox, line 10: That name isn't very descriptive, maybe `enable-subscriptions-for-promoted-addons` instead ? Anything that indicates what kind of subscription we're talking about.

src/olympia/devhub/views.py
```diff
@@ -1837,3 +1842,101 @@ def logout(request):
...
+@csp_update(
+    SCRIPT_SRC="https://js.stripe.com",
+    CONNECT_SRC="https://api.stripe.com",
+    FRAME_SRC=["https://js.stripe.com", "https://hooks.stripe.com"],
+)
+def onboarding_subscription(request, addon_id, addon):
+    sub = get_promoted_subscription_or_404(addon=addon)
+
+    fields_to_update = {}
+    if addon.has_author(request.user) and not sub.link_visited_at:
```
diox, line 1869: `@dev_required` should ensure `has_author()` already. But I'm thinking we probably want to check that the user is an owner and not just a developer ?

willdurand: I use `has_author()` to only set the `link_visited_at` when the person who is supposed to open the link has opened it. That way, admins can browse the page too.

src/olympia/devhub/views.py
willdurand: I changed to always create a new session, unless the payment has been successful.

src/olympia/devhub/tests/test_views_subscription.py
willdurand: I added a test for that.

willdurand:
I addressed most comments and added some more logs. This should be ready for another review.

bobsilverberg (approved):
Looks good, thanks @willdurand

PR #15770, opened 2020-10-19
RE: "addons": [] is empty for theme categories in the shelf API 

### LINKED ISSUE #15688: "addons": [] is empty for theme categories in the shelf API
STR:
1. Enable the following homepage shelves from[ admin](https://addons-internal.dev.mozaws.net/en-US/admin/models/discovery/homepageshelves/) 
"See more recommended extensions" with criteria ?promoted=recommended&sort=random&type=extension
with a theme category:
"Scenery"(theme category) criteria: ?category=scenery&sort=recommended%2Cusers&type=statictheme&app=firefox
or 
"Holiday (theme category)" criteria: ?category=holiday&sort=recommended%2Cusers&type=statictheme&app=firefox

2. Make an API request
https://addons-dev.allizom.org/api/v4/shelves/

Expected result:
there are results for "addons" in both shelves.

Actual result:
The theme shelf won't display results. It's "addons": []

This issue is also reproducible if one of the theme shelves from above is associated with 
[Recommended extensions sorted by rating](https://addons-internal.dev.mozaws.net/en-US/admin/models/discovery/shelfmodule/14/change/) : ?promoted=recommended&sort=rating&type=extension

I wrote this as a followup for https://github.com/mozilla/addons-server/issues/15284
The issue mentioned as a [comment](https://github.com/mozilla/addons-server/issues/15284#issuecomment-704231174) is no longer reproducible.

### LINKED ISSUE #15689: The shelf API responds with a 400 Bad Request when a recommended and a query shelf are enabled
Followup for: 
https://github.com/mozilla/addons-server/issues/15284

STR:
1. Enable the following shelves:
[See more recommended extensions ](https://addons-internal.dev.mozaws.net/en-US/admin/models/discovery/shelfmodule/34/change/) ?promoted=recommended&sort=random&type=extension
or
[See more recommended themes](https://addons-internal.dev.mozaws.net/en-US/admin/models/discovery/shelfmodule/37/change/) ?promoted=recommended&sort=random&type=statictheme

with the following shelf:
[Just a query](https://addons-internal.dev.mozaws.net/en-US/admin/models/discovery/shelfmodule/39/change/) with criteria: ?platform=windows&q=roboform

2. make an API request to https://addons-dev.allizom.org/api/v4/shelves/

Expected result:
the content of the 2 active shelves should be displayed.

Actual result:
The API responds with "The \"sort\" parameter \"random\" can only be specified when the \"featured\" or \"promoted\" parameter is also present, and the \"q\" parameter absent."

xlisachan:
Fixes #15688 
Fixes #15689 

This pull request revises the serializer to reset `request` before returning `get_addons`. 

The addons are now displayed for each shelf instead of an empty array or  `The "sort" parameter "random" can only be specified when the "featured" or "promoted" parameter is also present, and the "q" parameter absent.` error.

### DIFF
--- a/src/olympia/shelves/serializers.py
+++ b/src/olympia/shelves/serializers.py
@@ -43,11 +43,11 @@ class ShelfSerializer(serializers.ModelSerializer):
             criteria = obj.criteria.strip('?')
             params = dict(parse.parse_qsl(criteria))
             request = self.context.get('request', None)
+            tmp = request.GET
             request.GET = request.GET.copy()
             request.GET.update(params)
             addons = AddonSearchView(request=request).data
-            for key in request.GET.dict():
-                del request.GET[key]
+            request.GET = tmp
             return addons
         else:
             return None
--- a/src/olympia/shelves/serializers.py
+++ b/src/olympia/shelves/serializers.py
@@ -45,7 +45,10 @@ class ShelfSerializer(serializers.ModelSerializer):
             request = self.context.get('request', None)
             request.GET = request.GET.copy()
             request.GET.update(params)
-            return AddonSearchView(request=request).data
+            addons = AddonSearchView(request=request).data
+            for key in request.GET.dict():
+                del request.GET[key]
+            return addons
         else:
             return None
 
--- a/src/olympia/shelves/tests/test_serializers.py
+++ b/src/olympia/shelves/tests/test_serializers.py
@@ -51,11 +51,25 @@ class TestShelvesSerializer(ESTestCase):
         cls.refresh()
 
     def setUp(self):
-        self.search_shelf = Shelf.objects.create(
+        self.search_pop_thm = Shelf.objects.create(
             title='Populâr themes',
             endpoint='search',
             criteria='?sort=users&type=statictheme',
             footer_text='See more populâr themes')
+
+        self.search_hol_thm = Shelf.objects.create(
+            title='Holidây themes',
+            endpoint='search',
+            criteria=(
+                '?category=holiday&sort=recommended%2Cusers' + \
+                '&type=statictheme&app=firefox'),
+            footer_text='See more holidây themes')
+
+        self.search_rec_thm = Shelf.objects.create(
+            title='Recommended themes',
+            endpoint='search',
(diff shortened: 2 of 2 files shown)


src/olympia/shelves/serializers.py
```diff
@@ -45,7 +45,10 @@ def get_addons(self, obj):
             request = self.context.get('request', None)
             request.GET = request.GET.copy()
             request.GET.update(params)
-            return AddonSearchView(request=request).data
+            addons = AddonSearchView(request=request).data
+            for key in request.GET.dict():
+                del request.GET[key]
```
eviljeff, line 50: `request.GET = QueryDict()` would do to get an empty GET querydict instead of looping through the keys.

But it'd miss the case where we want to pass on some parameters from the `/shelves/` endpoint (e.g. `/shelves/?lang=fr` to enable French localisation) down to each of the requests.  So we've got to keep a copy of the unmodified query params from the original request object before we start changing it, so we're starting with those parameters each time.

eviljeff (changes requested):

eviljeff:
Also, github doesn't let issues be assigned to people outside of the org until they comment in that issue - opening a PR isn't enough.

xlisachan:
With the latest push, a copy of the unmodified `request.GET` is stored before updating the query_params and used to 'reset' the request object after retrieving the add-ons. Please review when you have a moment. Thanks!

eviljeff (approved):
👍

src/olympia/shelves/serializers.py
```diff
@@ -43,9 +43,12 @@ def get_addons(self, obj):
             criteria = obj.criteria.strip('?')
             params = dict(parse.parse_qsl(criteria))
             request = self.context.get('request', None)
+            tmp = request.GET
             request.GET = request.GET.copy()
             request.GET.update(params)
-            return AddonSearchView(request=request).data
+            addons = AddonSearchView(request=request).data
+            request.GET = tmp
```
eviljeff, line 50: if we had to do this in a few different places we could write it as a decorator to be used with a `with` statement (it would make the changes to request and revert them at the end) - not worth the extra code/complexity for one occurrence though.

PR #15778, opened 2020-10-20
Add logic to create/retrieve a Stripe Checkout session

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15725

---

This will be needed for #15726, which will be a large patch (because 3 new views + tests).


willdurand:
ha well, no, this isn't going to work.

Issue #15786, opened 2020-10-20
Handle already promoted add-ons after payment


willdurand:
Let's adjust the content for add-ons part of the pilot program.

**For QA:** if an add-on is already promoted as "sponsored" (or verified) and the current version has been approved, the payment flow should say something like "You're done".

willdurand:
I think we can start with a different message like `You're done!`.

ioanarusiczki:
@willdurand For all the verified/sponsored add-ons I verified for https://github.com/mozilla/addons-server/issues/15726, that went through the payment step when I try to access the URL again I receive "You're done" and the Manage add-on button that redirect to edit addon page.

![you're done](https://user-images.githubusercontent.com/33448286/97148127-44548600-1773-11eb-8cb3-40f57e05bfbf.png)

Those that are sponsored - already on the sponsored shelf (approved before introducing the payment flow) can be onboarded - the flow starts with "Continue to Stripe Checkout". 

I believe that's the expected behavior. 

Verified on dev - FF82(Win10).

PR #15788, opened 2020-10-20
Handle already promoted add-ons after initial payment

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15786

### DIFF
--- a/src/olympia/devhub/tests/test_views_subscription.py
+++ b/src/olympia/devhub/tests/test_views_subscription.py
@@ -76,9 +76,8 @@ class TestOnboardingSubscription(OnboardingSubscriptionTestCase):
         ]
 
         # Get the page.
-        queries = 31
+        queries = 36
         with self.assertNumQueries(queries):
-            # 31 queries:
             # - 3 users + groups
             # - 2 savepoints (test)
             # - 3 addon and its translations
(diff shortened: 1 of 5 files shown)


src/olympia/devhub/tests/test_views_subscription.py
```diff
@@ -76,9 +76,8 @@ def test_get(self, create_mock):
         ]
 
         # Get the page.
-        queries = 31
+        queries = 36
```
willdurand, line 79: I don't quite get why we have some many queries.. It seems we're fetching `previews`, `versions`, `files`, etc. many times for nothing.

diox: `previews`, `versions`, `files` etc come for the `transformer` that is called automatically when you fetch one or more add-ons. It's a relic from ancient times that is still quite useful in some places, and completely wasteful in others, so we've been scared to touch it.

That being said, this should happen *once*. If it happens multiple times, it means somewhere you're re-fetching an instance instead of re-using one, which is probably worth looking into.

bobsilverberg: This is just an unknowledgeable drive-by, but I seem to recall seeing code that allows us to avoid using the transformer if we don't need it. Is that not possible in this case?

diox: We have a `no_transforms()` method, yes. But before adding that, we should figure out why there are duplicate queries for the add-on here.

willdurand: I tried and couldn't see anything obvious..

diox: I suspect this is not tied to your code at all and just a widespread problem in devhub. I'll take a look.

willdurand:
I rebased this patch, it should be good to review.

src/olympia/devhub/tests/test_views_subscription.py
diox: The duplicates are caused by the my add-ons menu (that list all your add-ons) and some extra ownership checks in templates for the most part - We can deal with that in a follow-up, it affects all devhub, not just this page.

diox (approved):

PR #15793, opened 2020-10-21
Remove contrib:welcome from stalebot exclusions

muffinresearch:
Removes the contrib:welcome label from stalebot exclusions.


eviljeff:
I tried to approve this for consistency but got github errors so just merged it.

Issue #15795, opened 2020-10-21
500 error when making an addon sponsored/verified


ioanarusiczki:
I was trying to add back the promoted shelf on the frontend from the promoted model - admin, I have to change the Group from "Not promoted" into "Sponsored" and Save.

Result:
I receive a 500 server error

Notes:
I could reproduce the issue - when changing the group to Sponsored or Verified. It is not reproducible for the other groups.
Ids in this situation: 496379, 620026, 620031, 502855

willdurand:
@ioanarusiczki once this is verified on -dev, we will cherry-pick the patch for it because -stage is affected too.

ioanarusiczki:
@willdurand Alright! first thing tomorrow morning.

ioanarusiczki:
@willdurand I checked on dev and I can't reproduce the issue. - FF82(Win10).

ioanarusiczki:
Verified on AMO stage - FF82(Win10).

PR #15797, opened 2020-10-21
Avoid 500s when changing the group of a promoted add-on

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15795

---

Not sure if this is the right way :tm: but it works and given that we don't
allow subscription data updates on this admin page, it should be fine.

This issue linked above should be fixed before this week's push.

### DIFF
--- a/src/olympia/promoted/admin.py
+++ b/src/olympia/promoted/admin.py
@@ -102,7 +102,8 @@ class PromotedAddonAdmin(admin.ModelAdmin):
 
     def get_inline_instances(self, request, obj=None):
         inlines = self.inlines
-        if obj and obj.group in PROMOTED_GROUPS_FOR_SUBSCRIPTION:
+        if (obj and obj.group in PROMOTED_GROUPS_FOR_SUBSCRIPTION and
+                request.method != "POST"):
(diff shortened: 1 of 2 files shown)


eviljeff:
what's the actual error this is trying to "fix"?  I'm concerned we're hacking around a genuine bug rather than addressing it.

willdurand:
> what's the actual error this is trying to "fix"? I'm concerned we're hacking around a genuine bug rather than addressing it.

The error says:

```
django.core.exceptions.ValidationError: ['ManagementForm data is missing or has been tampered with']
```

It's captured by the test case added in this patch.

eviljeff:
For reference (because this doesn't seem to be spelled out in the issue either), the failure is because the inline form is only shown for sponsored addons but if the request changes the addon _to_ sponsored the inline form isn't shown when page was submitted but its now needed to be a complete submission.

Like you said, it only works when there's nothing to change on the inline form.  Once the subscription can be changed at all on the form it'll fail again.  It's the right way... but it's not a bad workaround in the interim.

eviljeff (approved):
acceptable fix 😬

PR #15808, opened 2020-10-22
Tweak PromotedSubscription fields

willdurand:
This is a bit unfortunate but after having playing with the current code, I feel like this is needed.


src/olympia/promoted/migrations/0013_auto_20201022_0906.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.16 on 2020-10-22 09:06
```
eviljeff, line 1: you could have combined this into a single migration file but 🤷

eviljeff (approved):

Issue #15785, opened 2020-10-20
Do not automatically release a version after payment if there are versions pending for review


willdurand:
This is a different scenario, related to #15784.

eviljeff:
after discussion in a meeting we're still going to re-sign the previously approved version even if there are new versions pending review

PR #15822, opened 2020-10-23
reapprove the current version when payment has been made, or for spotlight

### LINKED ISSUE #15784: Release a new version of an add-on after payment
Once the payment flow has been completed (successfully) and iff there is no version pending for review, we should create a new version automatically and sign it so that the add-on becomes promoted (if not done already).

### LINKED ISSUE #15801: Do not require version approval for an add-on to be added to the Spotlight group
Because Spotlight add-ons generally come up unexpectedly and we don't necessarily have a fast connection with the developer, it's best if we don't require a new version to be approved for the changes in a Spotlight add-on to apply. Operationally, we will always perform a pre-review on the current version before making the change in the admin, and nothing needs to change in the file, so it should be okay to apply the change right away.

/cc @devaneymoz @kewisch

eviljeff:
fixes #15784 (but ignores anything to do with pending versions - #15785 will deal with that) and fixes #15801 at the same time


src/olympia/constants/promoted.py
```diff
@@ -18,6 +18,8 @@
         'badged',
         'autograph_signing_states',
         'can_primary_hero',
+        'subscription',
```
willdurand, line 21: Shouldn't this be something like `require_subscription`

src/olympia/devhub/views.py
```diff
@@ -1939,6 +1939,7 @@ def onboarding_subscription_success(request, addon_id, addon):
             payment_completed_at=datetime.datetime.now(),
         )
         log.info('PromotedSubscription %s has been completed.', sub.pk)
+        sub.promoted_addon.approve_for_addon()
```
willdurand, line 1942: I guess we need to update the content of the template (`onboarding_subscription.html`) too.

src/olympia/promoted/models.py
```diff
@@ -80,16 +79,26 @@ def approve_for_version(self, version):
...
-                promoted_addon=self
-            ).exists()
-        ):
-            PromotedSubscription.objects.create(promoted_addon=self)
+        if self.group.subscription:
+            if not hasattr(self, 'promotedsubscription'):
+                PromotedSubscription.objects.create(promoted_addon=self)
+        elif (self.group.immediate_approval and
+              self.approved_applications != self.all_applications):
+            self.approve_for_addon()
```
willdurand, line 101: Is this backward compatible with the current -prod behavior? (i.e. the subscription feature is only enabled on -dev)

willdurand (commented):
I only took a quick look and I have a few comments.

src/olympia/devhub/views.py
eviljeff: Oh, I didn't read the text.  Yep.

src/olympia/promoted/models.py
eviljeff: The changes I made would be backwards compatible - it only wouldn't be if `SPONSORED` or `VERIFIED` were `immediate_approval = True` but it's only `SPOTLIGHT` (currently).

That said, looking again at the existing code you landed a few days ago - that would create `PromotedSubscription` instances if an existing `VERIFIED` or `SPONSORED` was edited/added regardless of the waffle state.  Is that a problem?  (I guess not because we don't check for promotedsubscription anywhere currently)

src/olympia/devhub/views.py
willdurand: I think you'll have to rebase your patch too

eviljeff: indeed, and `addon_already_promoted` needed taking into account now that's a thing https://github.com/mozilla/addons-server/pull/15822/commits/2fed0781911f2da5ede75bc9268215dad9461bc6

src/olympia/constants/promoted.py
eviljeff: `subscription` would be fine... though I see we've used that attribute name a lot already so yeah, it's confusing.  https://github.com/mozilla/addons-server/pull/15822/commits/8a03ccd787ff685e6e1bd14eabafa872e59fe21b

willdurand:
Looks like this shouls be rebased again (sorry). The main change should be the use of `strong` instead of `em` in the template.

Talking about the template, I am not sure about passing the group name as a l10n variable, that might make it harder to translate the sentence.

eviljeff:
> Looks like this should be rebased again (sorry). The main change should be the use of `strong` instead of `em` in the template.

I will rebase once the review is finished.

> Talking about the template, I am not sure about passing the group name as a l10n variable, that might make it harder to translate the sentence.

How so?  The alternative is worse - the name of a group (which could be changed later) localised in many different places.

willdurand:
> How so? The alternative is worse - the name of a group (which could be changed later) localised in many different places.

mm in addons-frontend we couldn't concatenate strings like that because that doesn't work for all languages (that is why we have translated strings like "This (addon|extension|theme) is nice" 3 times instead of "This %(addonType)s is nice".

willdurand:
> I will rebase once the review is finished.

I get that rebasing the patch one more time isn't fun but it makes it hard to review as the final patch won't be exactly the same as the one I am seeing right now.

eviljeff:
> > I will rebase once the review is finished.
> 
> I get that rebasing the patch one more time isn't fun but it makes it hard to review as the final patch won't be exactly the same as the one I am seeing right now.

If there is a material change I will ask for another review. Which there is unlikely to be as you already said it was a minor adjustment to add `<strong>`

eviljeff:
> > How so? The alternative is worse - the name of a group (which could be changed later) localised in many different places.
> 
> mm in addons-frontend we couldn't concatenate strings like that because that doesn't work for all languages (that is why we have translated strings like "This (addon|extension|theme) is nice" 3 times instead of "This %(addonType)s is nice".

It's not concatenation, it's placeholder substitution. I'm surprised this is necessary in addons-frontend either but I'm a lot less familiar with it.

src/olympia/devhub/views.py
```diff
@@ -1940,6 +1940,8 @@ def onboarding_subscription_success(request, addon_id, addon):
             payment_completed_at=datetime.datetime.now(),
         )
         log.info('PromotedSubscription %s has been completed.', sub.pk)
+        if not sub.addon_already_promoted:
+            sub.promoted_addon.approve_for_addon()
```
willdurand, line 1944: Not sure how/if we should solve this but paid subscriptions on -dev won't have the add-ons auto-approved. The success page will say "add-on will be re-signed" but that won't happen. (I am sure QA will verify that)

WDYT?

src/olympia/promoted/models.py
willdurand: >  Is that a problem? (I guess not because we don't check for promotedsubscription anywhere currently)

I don't think it's a problem because we cannot do anything without the waffle enabled.

src/olympia/devhub/views.py
willdurand: On Element, we decided to truncate the promoted subscription in -dev. It won't be possible to trigger such a scenario in stage/prod.

willdurand (approved):
r+, let me know if you need a quick look after rebase

PR #15840, opened 2020-10-26
make Version.version a VersionString, and VersionString a complete object

### LINKED ISSUE #15798: make Version.version (and other version fields) a type of VersionString
We recently refactored version string comparison to fix some edge cases but it would be good if Version.version was naturally able to be compared as a version string everywhere.  This will lead to more straightforward code where version is compared to, and will remove any edge case where we accidentally forget to convert it first

eviljeff:
fixes  #15798 - supersedes #15800

### DIFF
--- a/src/olympia/blocklist/admin.py
+++ b/src/olympia/blocklist/admin.py
@@ -423,6 +423,7 @@ class BlocklistSubmissionAdmin(admin.ModelAdmin):
             'save_as': False,
             'block_history': self.block_history(
                 self.model(input_guids=guids_data)),
+            'submission_complete': False,
         }
         context.update(**self._get_enhanced_guid_context(request, guids_data))
         return TemplateResponse(
--- a/src/olympia/versions/compare.py
+++ b/src/olympia/versions/compare.py
@@ -1,7 +1,4 @@
 import re
-from functools import total_ordering
-
-from django.utils.encoding import force_text
 
 
 BIGINT_POSITIVE_MAX = 2 ** 63 - 1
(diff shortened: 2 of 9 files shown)


src/olympia/blocklist/admin.py
```diff
@@ -423,6 +423,7 @@ def add_view(self, request, **kwargs):
             'save_as': False,
             'block_history': self.block_history(
                 self.model(input_guids=guids_data)),
+            'submission_complete': False,
```
bobsilverberg, line 426: How is this change related to the issue?

src/olympia/versions/compare.py
```diff
@@ -99,22 +108,22 @@ def __gt__(self, other):
...
 
 
 def version_int(version):
     """This is used for converting an app version's version string into a
     single number for comparison.  To maintain compatibility the minor parts
     are limited to 99 making it unsuitable for comparing addon version strings.
     """
-    vstr = VersionString(version)
-    vdict = dict(vstr.full_dict)
+    vdict = dict(VersionString(version).parts)
```
bobsilverberg, line 126: I might be missing something here, but isn't this patch about making `version` always be a `VersionString`? Why do we need `VersionString(version)`?

bobsilverberg (approved):
Looks good, thanks @eviljeff. I have a couple of questions.

r+wc

src/olympia/blocklist/admin.py
eviljeff: it was a drive-by fix for a warning about undefined template variables.

src/olympia/versions/compare.py
eviljeff: The changes are related to add-on versions, but `version_int` is used by _application_ version comparison code - that doesn't use `VersionString` (yet).

src/olympia/blocklist/admin.py
eviljeff: I could remove it again.

bobsilverberg: No need to do that. I was just curious.

diox (approved):

Issue #15864, opened 2020-10-29
Log in to dockerhub at every step in circleci config


diox:
[Docker Hub is soon going to enforce pull rate limits on anonymous users](https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/). This will impact the CircleCI jobs that need to pull images to fetch/build stuff.

The solution is to log in to Docker Hub at the beginning of every job that deals with docker in Circle CI config, not just before pushing images like we do now.

PR #15865, opened 2020-10-29
Login to dockerhub before building the docker image

diox:
Fixes https://github.com/mozilla/addons-server/issues/15864


diox:
I don't think we need anything else ? We don't rely on integrated build command, so we don't need an auth block like it was done for https://github.com/mozilla-services/Dockerflow/blob/master/.circleci/config.yml

bqbn (approved):

PR #15870, opened 2020-10-29
Support CRX version 3

### LINKED ISSUE #15849: Add support for CRX version 3
Looks like we support CRX version 3 in addons-linter so let's add support for it in addons-server too, see: https://github.com/mozilla/addons-scanner-utils/pull/106#discussion_r513070587

diox:
Fixes #15849


src/olympia/files/tests/test_models.py
```diff
@@ -1065,6 +1065,36 @@ def test_webextension_crx_large(self):
...
+        explicit exception otherwise (doesn't have to be caught, showing a 500
+        error is fine)."""
+        data = b'Cr24\x04\x00\x00\x00&\x01\x00\x00\x00\x01\x00\x00'
+        with self.assertRaises(InvalidOrUnsupportedCrx):
+            FileUpload.from_post([data], filename='test.crx', size=1234)
+
+    def test_webextension_crx_version_cant_unpack(self):
+        """Test to ensure we raise an explicit exception when we can't unpack
+        a crx (doesn't have to be caught, showing a 500 error is fine)."""
+        data = b'Cr24\x02\x00\x00\x00&\x00\x00\x00\x01\x00\x00'
```
eviljeff, line 1094: Can you differentiate the exceptions theses blobs of data are raising?  They could all be just bad data raising in the same place rather than good data that's the wrong version, etc, raising for different reasons.

src/olympia/files/tests/test_models.py
```diff
@@ -1065,6 +1065,36 @@ def test_webextension_crx_large(self):
         assert upload.path.endswith('.xpi')
         storage.delete(upload.path)
 
+    def test_webextension_crx_version_3(self):
+        """Test to ensure we accept CRX uploads (version 3), but convert them
+        into XPI files ASAP to keep things simple.
+        """
+        upload = self.get_upload('webextension_crx3.crx')
```
eviljeff, line 1072: Is it practical to test this in the same way as the following tests, with `FileUpload.from_post`?

eviljeff (approved):

src/olympia/files/tests/test_models.py
diox: It's possible but I wanted to follow what the original tests for successful scenarios did

src/olympia/files/tests/test_models.py
diox: Yeah I can look in the message, I'll do that.

diox: c1cf299ed8e2250d3584ed2e493acba14a8e57af

Issue #15857, opened 2020-10-28
Allow setting specific subscription rates for promoted add-ons


jvillalobos:
There may be cases where certain promoted add-ons get different rates (like a discount, for instance), so we want to allow some flexibility in the onboarding process.

What I'm thinking we should do is have a field in the admin for the subscription rate that can be changed by the admin and will determine what rate to show to the developer during onboarding. So, if the admin sets the add-on as Sponsored, the onboarding link is generated and the rate is set for the default for that tier, and that can be changed to something else by the admin. I think we should call this "Onboarding rate" so it's clear that it's only used for that. You wouldn't be able to change the rate from AMO after onboarding.

willdurand:
> I think we should call this "Onboarding rate" so it's clear that it's only used for that. You wouldn't be able to change the rate from AMO after onboarding.

Ok, this is doable. The rate will be an amount (strictly positive integer) **in cents**.

willdurand:
For QA:

1. if we set an onboarding rate, we should get that price displayed in Stripe
2. if we have paid already, we shouldn't be able to change the price

ioanarusiczki:
@willdurand I verified on AMO dev:
- all promoted addon types will show an empty Promoted Subscriptions section. 
- Verified and Sponsored addons will show all details for the Promoted Subscription section.
- the "Onboarding rate" is available and it can be set to a value as described. In case it is not set for a custom value, the standard rates ( 5 euros for Verified and 10 euros for Sponsored) will be displayed while making the payment.
- the "Onboarding rate" for paid addons will display the value in read only.

![value](https://user-images.githubusercontent.com/33448286/97991172-bbc88c00-1de9-11eb-8369-cf795bba6bbf.png)

- the already paid addons (before the actual implementation) won't show a value, it's "-" but it also read only. 
- the onboarding fee can be edited in admin until the user won't complete the payment. In case the user is on the Stripe page already but goes back and does a page refresh then clicks again "Continue to Stripe checkout", a different value is displayed in case the admin changed the value.
 
Other notes:

- the [testing cards](https://stripe.com/docs/testing#cards) all worked for me except the last 6 (from Discover below). I've done many payments using the first 3 card brands.

PR #15871, opened 2020-10-29
Allow setting specific subscription rates for promoted add-ons

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15857


bobsilverberg (approved):
This looks good but I'm wondering, is simply making the field `readonly` in the admin enough, or should there also be some code enforcing the fact that the field cannot be updated, if somehow the form is altered? Or perhaps setting it as `readonly` in the admin already takes care of that?

r+wc

willdurand:
> This looks good but I'm wondering, is simply making the field `readonly` in the admin enough, or should there also be some code enforcing the fact that the field cannot be updated, if somehow the form is altered? Or perhaps setting it as `readonly` in the admin already takes care of that?

I suppose django admin takes care of that?

diox:
Yes, putting the field in `readonly_fields` should ensure if somehow the data is submitted, it will be ignored.

PR #15881, opened 2020-11-02
RE: Include add-ons based on the collections endpoint/criteria in the homepage shelves API

### LINKED ISSUE #15614: Include add-ons based on the collections endpoint/criteria in the homepage shelves API
Supports #14871

Implement collections type in the homepage shelves API to include add-ons based on the collections endpoint/criteria

xlisachan:
Fixes #15614 

This pull request updates the shelves' serializer to include add-ons to shelves with a `collections` endpoint. The update to `serializers.py` populates the add-ons for collections in the API, but I am having trouble with updating `test_serializers.py`.

I am getting the error, `AttributeError: 'WSGIRequest' object has no attribute 'query_params'`, but `query_params` were not needed when updating the serializer for collections. If `query_params` are added for collections in the test though, it leads to the following error, `django.http.response.Http404: No UserProfile matches the given query.

 I'm not sure how to resolve the error, can you please review and provide guidance when you have a moment? Thank you!

### DIFF
--- a/src/olympia/shelves/serializers.py
+++ b/src/olympia/shelves/serializers.py
@@ -33,7 +33,7 @@ class ShelfSerializer(serializers.ModelSerializer):
                 'collection-addon-list',
                 request=self.context.get('request'),
                 kwargs={
-                    'user_pk': settings.TASK_USER_ID,
+                    'user_pk': str(settings.TASK_USER_ID),
                     'collection_slug': obj.criteria})
         else:
             url = None
--- a/src/olympia/shelves/serializers.py
+++ b/src/olympia/shelves/serializers.py
@@ -8,6 +8,7 @@ from rest_framework.reverse import reverse as drf_reverse
 
 from olympia.addons.serializers import ESAddonSerializer
 from olympia.addons.views import AddonSearchView
+from olympia.bandwagon.views import CollectionAddonViewSet
 
 from .models import Shelf
 
@@ -50,6 +51,13 @@ class ShelfSerializer(serializers.ModelSerializer):
             addons = AddonSearchView(request=request).data
             request.GET = tmp
             return addons
+        elif obj.endpoint == 'collections':
+            request = self.context.get('request', None)
+            kwargs = {
+                'user_pk': str(settings.TASK_USER_ID),
+                'collection_slug': obj.criteria}
+            return CollectionAddonViewSet(request=request, action='list',
+                                          kwargs=kwargs).data
         else:
             return None
 
--- a/src/olympia/shelves/serializers.py
+++ b/src/olympia/shelves/serializers.py
@@ -44,7 +44,7 @@ class ShelfSerializer(serializers.ModelSerializer):
         if obj.endpoint == 'search':
             criteria = obj.criteria.strip('?')
             params = dict(parse.parse_qsl(criteria))
-            request = self.context.get('request', None)
+            request = self.context.get('request')
             tmp = request.GET
             request.GET = request.GET.copy()
             request.GET.update(params)
@@ -52,7 +52,7 @@ class ShelfSerializer(serializers.ModelSerializer):
             request.GET = tmp
             return addons
         elif obj.endpoint == 'collections':
-            request = self.context.get('request', None)
+            request = self.context.get('request')
             kwargs = {
                 'user_pk': str(settings.TASK_USER_ID),
                 'collection_slug': obj.criteria}
--- a/src/olympia/shelves/tests/test_serializers.py
+++ b/src/olympia/shelves/tests/test_serializers.py
@@ -6,13 +6,14 @@ from rest_framework.test import APIRequestFactory
 from freezegun import freeze_time
 
 from django.conf import settings
+from django.contrib.auth.models import AnonymousUser
 
 from olympia import amo
 from olympia.addons.models import Addon
 from olympia.addons.tests.test_serializers import (
     AddonSerializerOutputTestMixin)
 from olympia.amo.tests import (
-    addon_factory, collection_factory, ESTestCase, reverse_ns, user_factory)
+    addon_factory, collection_factory, ESTestCase, reverse_ns)
 from olympia.bandwagon.models import CollectionAddon
 from olympia.constants.promoted import RECOMMENDED
 from olympia.promoted.models import PromotedAddon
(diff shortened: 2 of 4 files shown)


xlisachan:
I had forgot to include a collection in `def setUpTestData`, but I'm not sure if I added it correctly in this latest push. I am getting the error `django.http.response.Http404: No Collection matches the given query.`

I also updated `user_pk` in `shelves/serializers.py` - I had it set to `settings.TASK_USER_ID` and I believe it should be `request.user.pk` instead?

@eviljeff , can you please advise when you have a moment? Thank you!

eviljeff:
> I also updated `user_pk` in `shelves/serializers.py` - I had it set to `settings.TASK_USER_ID` and I believe it should be `request.user.pk` instead?

no, the task user is Mozilla, the account that also "owns" the official collections.  So the reverse is getting the collection under that account.

> I had forgot to include a collection in `def setUpTestData`, but I'm not sure if I added it correctly in this latest push. I am getting the error `django.http.response.Http404: No Collection matches the given query.`

Changing it from task user would do that - it'd be looking for that collection name under the _current_ user that called the API (which, unless you authenticated, would be anonymous), rather than the Mozilla user.

xlisachan:
@eviljeff, `serializers.py` has been updated to set `user_pk` back to `settings.TASK_USER_ID` per your feedback.

In addition, the following changes have been made in `test_serializers.py`:
- `def setUpTestData`: updated the collection test data by adding in `author` variable,
- `def test_url_and_addons_collections`: revised to check the url and addons in the collection shelf, 
- `def test_url_and_addons_search`: I tried to make this test a little clearer by separating the two shelves with `search` as their endpoints

`test_views.py` was also updated to include the collection test data. The number of queries changed from 4 to 26, but I am not sure if 26 should be the expected number of queries. Can you please take a look at the captured queries and let me know if it looks right- [https://pastebin.com/rnq6CzG0](url)

src/olympia/shelves/tests/test_serializers.py
```diff
@@ -78,85 +79,69 @@ def setUp(self):
...
 
     def serialize(self, instance, **extra_context):
+        self.request.query_params = dict(parse.parse_qsl(
+            instance.criteria))
+        extra_context['request'] = self.request
+        if instance.endpoint == 'collections':
+            extra_context['action'] = 'list'
+            extra_context['kwargs'] = {
+                'user_pk': self.request.user.pk,
+                'collection_slug': instance.criteria}
```
eviljeff, line 92: what's this extra_context for collections used for?  I still only see `request['context']` being used in the serializer.

src/olympia/shelves/serializers.py
```diff
@@ -50,6 +51,13 @@ def get_addons(self, obj):
             addons = AddonSearchView(request=request).data
             request.GET = tmp
             return addons
+        elif obj.endpoint == 'collections':
+            request = self.context.get('request', None)
```
eviljeff, line 55: nit: you don't need to specify `None` as the default with a dict `.get` call in python - it's the default anyway.

src/olympia/shelves/tests/test_views.py
```diff
@@ -100,7 +107,7 @@ def test_only_enabled_shelves_in_view(self):
         self.hpshelf_b.update(enabled=True)
         # don't enable shelf_c
 
-        with self.assertNumQueries(4):
+        with self.assertNumQueries(26):
```
eviljeff, line 110: 😨 22 extra queries?  That's all from a single collection? (with one addon and a sole author)?

eviljeff (approved):

src/olympia/shelves/tests/test_views.py
xlisachan: I just realized the link above that outlined the captured queries did not go to [pastebin](https://pastebin.com/rnq6CzG0), sorry! I was looking through the extra queries, but I wasn't sure which were necessary and/or where it needs to be revised to correct it.

xlisachan:
The latest push updates `extra_context` to `context` in `test/test_serializers.py` and removes the unnecessary `None` in `serializers.py`.
In addition, I corrected the link to pastebin that outlines the increase in the number of queries captured in `test/test_views.py`.
There is a conflicting file, but I'm hesitant to resolve it in case it merges this branch before you review the updates just made, as it has already been marked approved 😬🙈

src/olympia/shelves/tests/test_views.py
eviljeff: coincidentally, this week we started looking at the amazing (in a bad way!) number of queries the collection endpoints make (#16081) so 22 queries doesn't seem so exceptional now.

src/olympia/shelves/tests/test_serializers.py
eviljeff: argh, I wrote that the wrong way around - I should have written `context['request']`.  
My question was that you're passing `action`, and `kwargs` in as extra context (or just `context` as renamed now) but neither appear to be used in the serializer code itself, so I was wondering if that was an oversight or they were used in some hidden way (if so, please add a comment in the code explaining why they need to be passed in)

xlisachan: Aah, sorry it was an oversight! I had included that when I was trying to resolve the errors when running the tests and before realizing I had not added a collection in `setUpTestData`. I didn't check if it was necessary when all the tests had finally passed.

xlisachan:
The latest push removes the unnecessary lines in `def serialize` in `test_serializers.py`. Please review when you have a moment. Thanks!

src/olympia/shelves/tests/test_serializers.py
```diff
@@ -86,85 +85,64 @@ def setUp(self):
             api_settings.DEFAULT_VERSIONING_CLASS()
         )
         self.request.version = api_version
-        self.request.user = AnonymousUser()
+        self.request.user = user_factory()
```
eviljeff, line 88: I missed this - is it a new change?  Does AnonymousUser no longer work?  (the API should work when not logged in)

eviljeff (approved):
Just a question on the request user in test_serializers.py - and the branch needs a rebase like you noted.

src/olympia/shelves/tests/test_serializers.py
xlisachan: AnonymousUser works, I'll change it back.

xlisachan:
The latest push updates `user_factory()` back to `AnonymousUser()`. In addition, when I rebased, the lines that created the PromotedAddons in `setUpTestData` were removed, so I added them back in. Please review when you have a moment. Thank you!

eviljeff:
> In addition, when I rebased, the lines that created the PromotedAddons in `setUpTestData` were removed, so I added them back in. 

yeah, I did that a few weeks ago - the `recommended=True` kwarg in `addon_factory` does the same thing as manually creating the `PromotedAddon` instance in a RECOMMENDED group - in an attempt to fix intermittent test failures.  (`addon_factory` suspends the ES reindexing while it does all it's set-up so we don't end up with multiple indexing happening at once and potentially causing inconsistent data in elasticsearch).  Can you change it back.

xlisachan:
Thanks for the insight, I thought I hadn't rebased correctly - I have changed it back! One of the CI builds failed, but when I ran it locally it passed, can you restart it?

eviljeff (approved):
I think it's good now, r+ (!)

Issue #15889, opened 2020-11-02
Changing an archive extension to .crx will trigger a server error at validation


AlexandraMoga:
### Describe the problem and steps to reproduce it:
1. Have a valid add-on archive - an `.xpi` or `.zip` 
2. Change the archive extension to `.crx`
3. Submit the archive on AMO -dev

### What happened?
A 500 validation error:
![image](https://user-images.githubusercontent.com/31961530/97870130-2fe82e80-1d1b-11eb-8a5b-2d8ded18aa24.png)

### What did you expect to happen?
The validation should not fail with a server error

### Anything else we should know?
- probably a regression from #15849 
- this used to work before as a workaround to avoid the `.crx` validation error - see https://github.com/mozilla/addons-server/issues/14036#issuecomment-609663818

diox:
We should not 500, but this was never supposed to work. It should raise a proper error indicating to the user that it's not a valid crx file.

AlexandraMoga:
I've verified this issue on dev and the 500 error for `.zip` archives renamed as `.crx` is no longer reproducible. 

@diox I was under the impression that the validation should now error with an "Invalid archive" message, but from what I've seen, there was no error raised. The renamed archive produces the same validation results as the original one.

diox:
That depends on the file.
- Renaming `.zip` to `.crx` should now work (we attempt to parse it as a CRX, fail, fall back to normal zip)
- Renaming a random file like a `.png` to `.crx` should raise `Invalid or corrupt add-on file.`.
- Submitting a `.png` or other non-extension file type directly should raise `Unsupported file type, please upload a supported file (...)` with the accepted extensions list.

Obviously, submitting a valid .zip, .xpi or .crx should still work as normal.

AlexandraMoga:
Yes, when a non-supported file is being renamed as an apparently valid file (i.e .`zip`, `.xpi`, `.crx`) the "Invalid or corrupt add-on file" error message is received. Directly submitting a non-supported file type (i.e `.png`, `.xml` etc) will result in an "Unsupported file type" message. 

Valid add-on files are behaving the same as before.

This isssue is verified fixed on -dev.

PR #15890, opened 2020-11-02
Don't let the exception be raised for invalid/corrupt crx files

diox:
Fixes https://github.com/mozilla/addons-server/issues/15889

The idea is to catch the exception and let the error happen later, in `parse_addon()` (showing a generic message about the archive being invalid or corrupt)

### DIFF
--- a/src/olympia/files/models.py
+++ b/src/olympia/files/models.py
@@ -527,12 +527,16 @@ class FileUpload(ModelBase):
         # Filename we'll expose (but not use for storage).
         self.name = force_text('{0}_{1}'.format(self.uuid.hex, filename))
 
-        # Final path on our filesystem. Will always end in .xpi no matter what
-        # is uploaded. CRX files are converted beforehand. If somehow this is
-        # not a valid archive parse_addon() will eventually complain at
-        # validation time.
+        # Final path on our filesystem. If it had a valid extension we change
+        # it to .xpi (CRX files are converted before validation, so they will
+        # be treated as normal .xpi for validation). If somehow this is
+        # not a valid archive or the extension is invalid parse_addon() will
+        # eventually complain at validation time or before even reaching the
+        # linter.
+        if ext in amo.VALID_ADDON_FILE_EXTENSIONS:
+            ext = '.xpi'
         self.path = os.path.join(
-            user_media_path('addons'), 'temp', uuid.uuid4().hex, '.xpi')
+            user_media_path('addons'), 'temp', uuid.uuid4().hex, ext)
 
         log.info('UPLOAD: %r (%s bytes) to %r' % (self.name, size, self.path),
                  extra={'email': (self.user.email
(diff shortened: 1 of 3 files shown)


diox:
@eviljeff before I fix the tests (essentially just some breakage around opensearch, so I will likely just delete them), could you take a look to tell me what you think about the patch ? With it, an invalid crx should be rejected with `Invalid or corrupt add-on file` before even reaching the linter, since it would have failed to be converted to a zip, so `parse_addon()` would raise `BadZipFile`, which we catch to show that generic error.

Screenshot_2020-11-03 Upload Add-on Developer Hub Add-ons for Firefox

src/olympia/files/models.py
```diff
@@ -519,47 +520,45 @@ def save(self, *args, **kw):
...
-        self.name = filename
-        self.hash = 'sha256:%s' % hash_func.hexdigest()
+
+        hash_obj = None
+        if was_crx:
+            try:
+                hash_obj = write_crx_as_xpi(chunks, self.path)
+            except InvalidOrUnsupportedCrx:
+                # We couldn't convert the crx file. Write it to the filesystem
+                # normally, the linter should reject this with a proper error.
```
eviljeff, line 557: in the pr comments it says the error happens before file gets to the linter

src/olympia/files/tests/test_models.py
```diff
@@ -1071,36 +1086,47 @@ def test_webextension_crx_version_3(self):
...
         storage.delete(upload.path)
 
     def test_webextension_crx_not_a_crx(self):
         """Test to ensure we raise an explicit exception when a .crx file isn't
         a true crx (doesn't have to be caught, showing a 500 error is fine)."""
         data = b'Cr42\x02\x00\x00\x00&\x01\x00\x00\x00\x01\x00\x00'
-        with self.assertRaises(InvalidOrUnsupportedCrx) as exc:
-            FileUpload.from_post([data], filename='test.crx', size=1234)
-        assert str(exc.exception) == 'CRX file does not start with Cr24'
+        upload = FileUpload.from_post([data], filename='test.crx', size=1234)
```
eviljeff, line 1100: so, with the refactoring we've lost the ability to check that the individual exceptions are thrown correctly at this level - are there tests for `write_crx_as_xpi` directly?

eviljeff (commented):
I guess its fine to rely on later in the process to throw the error.  

We wouldn't show a specific error message about _what_ is wrong with the file though (or even that we know it's a crx and have failed to parse it as that, rather than just failed to parse it as a zip file).  That's a shame but I can't think of an easy way off the top of my head to pass that information down.

src/olympia/files/tests/test_models.py
diox: Not yet but I plan to add some

src/olympia/files/models.py
diox: Yeah I need to fix the comment

eviljeff (approved):

PR #15891, opened 2020-11-02
show new version number on subscription confirmation page

### LINKED ISSUE #15867: Expose the version we're re-signing to the developer in the subscription payment page
We tell the developer we'll resign the current version of the addon but we don't make clear which version that is (and that'll be version bumped slightly)

eviljeff:
fixes #15867

### DIFF
--- a/src/olympia/devhub/templates/devhub/addons/onboarding_subscription.html
+++ b/src/olympia/devhub/templates/devhub/addons/onboarding_subscription.html
@@ -18,19 +18,30 @@
     <h2>{{ _('Set Up Payment') }}</h2>
 
     {% if stripe_checkout_completed %}
-    <p>
-      {% trans %}
-      You're done!
-      {% endtrans %}
-    </p>
-      {% if not already_promoted %}
+      {% if already_promoted %}
       <p>
-        {% trans addon_name=addon.name, group_name=promoted_group.name %}
-        The latest approved version of your add-on <strong>{{ addon_name }}</strong>
-        will be re-signed and published as {{ group_name }}, and should become
-        available shortly. If you have any questions, don't hesitate to contact
-        us.
+        {% trans %}
+        You're done!
         {% endtrans %}
+      </p>
+      {% else %}
+      <p>
+        {% if new_version_number %}
+          {% trans addon_name=addon.name, group_name=promoted_group.name %}
+          You're done! A new version <strong>{{ new_version_number }}</strong> of your add-on
+          <strong>{{ addon_name }}</strong> will be published as {{ group_name }}
+          and should become available shortly.
+          {% endtrans %}
+          {% if existing_version_pending %}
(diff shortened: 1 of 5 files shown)


src/olympia/devhub/templates/devhub/addons/onboarding_subscription.html
```diff
@@ -18,20 +18,31 @@
     <h2>{{ _('Set Up Payment') }}</h2>
 
     {% if stripe_checkout_completed %}
-    <p>
-      {% trans %}
-      You're done!
-      {% endtrans %}
-    </p>
```
willdurand, line 25: Why not keeping this sentence like this? It's displayed in both cases below.

willdurand (approved):
r+wc

src/olympia/devhub/tests/test_views_subscription.py
```diff
@@ -76,7 +78,7 @@ def test_get(self, create_mock):
         ]
 
         # Get the page.
-        queries = 36
+        queries = 37
```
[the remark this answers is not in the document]
eviljeff: 🐼

src/olympia/devhub/templates/devhub/addons/onboarding_subscription.html
eviljeff: it added some non-trivial logic for a few lines of text - it's displayed in cases when `already_promoted` or `new_version_number` (now I write this... it's not complicated though) but not for new promotions where we don't have the version number (i.e. it failed).  The PRD comments from Jorge seemed like he wanted it as a single paragraph anyway so I simplified it.  Can keep it separate if you think that's best?

willdurand: not a big deal

src/olympia/devhub/templates/devhub/addons/onboarding_subscription.html
```diff
@@ -18,20 +18,31 @@
...
+        {% if new_version_number %}
+          {% trans addon_name=addon.name, group_name=promoted_group.name %}
+          You're done! A new version <strong>{{ new_version_number }}</strong> of your add-on
+          <strong>{{ addon_name }}</strong> will be published as {{ group_name }}
+          and should become available shortly.
+          {% endtrans %}
+          {% if existing_version_pending %}
+            {{ _('A newer version of your add-on is currently pending review.') }}
+          {% endif %}
+          {{ _('If you have any questions, don’t hesitate to contact us.') }}
```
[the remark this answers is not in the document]
eviljeff: (I just moved it from the `{% trans %}` block above).  `’` is the alternate (correct?) character for abbreviations - in this case it's probably just brought over from a google doc where it's been auto-corrected from `'` when Jorge typed it.  `’` has the added benefit of not breaking single quoted strings too!

Issue #15879, opened 2020-10-30
Add an endpoint to receive Stripe events (webhook)


willdurand:
We'll need to listen to some specific Stripe events in the near future so let's create an endpoint for that, which should at least log the events we configured in the Stripe dashboard.

Issue #15805, opened 2020-10-22
Handle subscription cancellation after initial payment


willdurand:
We should determine how to handle the following scenario:

1. developer has a promoted add-on that requires payment
2. developer successfully sets up payment _via_ Stripe Checkout
3. ...after some time...
4. developer wants to cancel their subscription

I thought Stripe would allow the developer to cancel the subscription on Stripe directly, but it does not seem possible, which is good for us because it means we have to implement the cancellation feature and we can design it the way we want.

willdurand:
Cancellation will be handled manually in Stripe. AMO should listen to events in order to update the promoted group after cancellation. We should also add a warning when an AMO admin wants to update a promoted add-on in the django admin.

jvillalobos:
One more thing: when an add-on is removed from a paid promoted group (either through Stripe or an AMO override), a new version should be generated so that it is signed without any promoted flags.

willdurand:
> One more thing: when an add-on is removed from a paid promoted group (either through Stripe or an AMO override), a new version should be generated so that it is signed without any promoted flags.

cc @eviljeff

eviljeff:
> One more thing: when an add-on is removed from a paid promoted group (either through Stripe or an AMO override), a new version should be generated so that it is signed without any promoted flags.

mmm, why?  If it's not in the promoted group (or group is not promoted) it won't show as promoted in the api/frontend.

jvillalobos:
So that users are updated to a version that isn't badged anymore and the action on AMO is consistent in Firefox.

eviljeff:
We don't do this for the other promoted groups that have special signing - should those be treated similarly?

imo, the version *is* verified - stopping making further payments doesn't rewrite history so it wasn't manually reviewed by us and verified at the time it was installed.  Is "consistency" the only user benefit to this?

jvillalobos:
> We don't do this for the other promoted groups that have special signing - should those be treated similarly?

I think so, yes, I just didn't want to add extra work to this request.

> imo, the version is verified - stopping making further payments doesn't rewrite history so it wasn't manually reviewed by us and verified at the time it was installed. Is "consistency" the only user benefit to this?

Cancellation can stem from a number of reasons, including a developer being uncooperative, an add-on being taken over by someone less responsible than the previous owner, etc. I expect lack of payment to be the less likely one. I would prefer that we have a way to remove add-ons from the program that brings them back to their previous state in every way.

eviljeff:
[...]

PR #15901, opened 2020-11-03
Add basic code to receive Stripe events (webhook)

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15879

---

I followed these to guides:

- https://stripe.com/docs/webhooks/build
- https://stripe.com/docs/webhooks/signatures

We only log received events so far. We'll have to see which events we want to listen to and what to do when we receive them. Webhooks don't listen to _all_ events by default, we have to select which ones are interesting to us, depending on the needs. This work is needed to handle cancellation in the future (see: https://github.com/mozilla/addons-server/issues/15805)


src/olympia/promoted/views.py
```diff
@@ -0,0 +1,35 @@
...
+    except stripe.error.SignatureVerificationError:
+        log.exception("received stripe event with invalid signature")
+        return HttpResponse(status=400)
+    except ValueError:
+        log.exception("received stripe event with invalid payload")
+        return HttpResponse(status=400)
+
+    log.info('Received stripe event type: "%s".', event.type)
+
+    return HttpResponse(status=200)
```
eviljeff, line 35: return a DRF `Response` object everywhere in the function.

src/olympia/api/urls.py
```diff
@@ -31,6 +31,7 @@
     re_path(r'^reviewers/', include('olympia.reviewers.api_urls')),
     re_path(r'^', include('olympia.signing.urls')),
     re_path(r'^', include(amo_api_patterns)),
+    re_path(r'^promoted/', include('olympia.promoted.api_urls')),
```
eviljeff, line 34: it's unfortunate, imo, that there's not some other endpoint to place this under

docs/topics/api/promoted.rst
```diff
@@ -0,0 +1,21 @@
...
+
+.. _stripe-webhook:
+
+This endpoint receives `event notifications
+<https://stripe.com/docs/webhooks>`_ from Stripe.
+
+    .. note::
+        Requests are signed by Stripe and verified by the server.
+
+.. http:get:: /api/v4/promoted/stripe-webhook
```
eviljeff, line 21: `http:get` vs. `@post_required`

eviljeff, line 21: is there no standard about what's in the payload we can document?

src/olympia/promoted/views.py
```diff
@@ -0,0 +1,35 @@
...
+def stripe_webhook(request):
+    payload = request.body
+    sig_header = request.META.get("HTTP_STRIPE_SIGNATURE")
+
+    try:
+        event = create_stripe_webhook_event(
+            payload=payload, sig_header=sig_header
+        )
+    except stripe.error.SignatureVerificationError:
+        log.exception("received stripe event with invalid signature")
```
eviljeff, line 27: `.exception` might end up being too extreme if miscreants start pinging the unauthenticated endpoint - they'll all turn up in sentry as a well as stackdriver.

src/olympia/promoted/views.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+log = olympia.core.logger.getLogger("z.promoted")
+
+
+@post_required
+@csrf_exempt
+def stripe_webhook(request):
+    payload = request.body
+    sig_header = request.META.get("HTTP_STRIPE_SIGNATURE")
```
eviljeff, line 20: I think I need to add a lint rule about using single quotes in addons-server python 😛

src/olympia/promoted/views.py
```diff
@@ -0,0 +1,35 @@
...
+
+from olympia.amo.decorators import post_required
+
+from .utils import create_stripe_webhook_event
+
+
+log = olympia.core.logger.getLogger("z.promoted")
+
+
+@post_required
```
eviljeff, line 16: This is written as a standard django view, but we want a Rest Framework api view because its in the api. 

`@api_view(['POST'])`

src/olympia/promoted/views.py
```diff
@@ -0,0 +1,35 @@
...
+@csrf_exempt
+def stripe_webhook(request):
+    payload = request.body
```
eviljeff, line 19: looks like the in stripe docs the payload is parsed as JSON before passing it to `construct_event`.

src/olympia/promoted/views.py
```diff
@@ -0,0 +1,35 @@
...
+            payload=payload, sig_header=sig_header
+        )
+    except stripe.error.SignatureVerificationError:
+        log.exception("received stripe event with invalid signature")
+        return HttpResponse(status=400)
+    except ValueError:
+        log.exception("received stripe event with invalid payload")
+        return HttpResponse(status=400)
+
+    log.info('Received stripe event type: "%s".', event.type)
```
eviljeff, line 33: so we're not actually doing anything with the event at this time?  Would the functionality sit in `create_stripe_webhook_event`?  I'm wondering how the processing flow is going to work here.

docs/topics/api/index.rst
```diff
@@ -41,6 +41,7 @@ using the API.
    categories
    collections
    discovery
+   promoted
```
eviljeff, line 44: needs an entry in the changelog in overview.rst too

docs/topics/api/promoted.rst
willdurand: ah oops :/

src/olympia/promoted/views.py
willdurand: tell Black about it :)

src/olympia/promoted/views.py
willdurand: ha ok

src/olympia/promoted/views.py
willdurand: not for signed requests

src/olympia/promoted/views.py
willdurand: My idea would be to submit a Celery task, which we would have to create for each event we want to listen to and handle. This only receives and validates notifications and once we have an `event` object, we can do whatever we want (which is the complicated part, heh).

We'll see shortly with payment failures, which we should handle by sending an email to amo admins.

docs/topics/api/promoted.rst
willdurand: Probably documented somewhere in the Stripe docs. It's not like we can fake such requests, we need to use their CLI, which can forward events locally, simulate all kind of notifications, etc. I don't think we should document anything else given that we cannot create such requests manually.

src/olympia/promoted/tests/test_utils.py
```diff
@@ -149,3 +150,24 @@ def test_create_stripe_checkout_session_with_custom_rate():
...
+
+@override_settings(STRIPE_API_WEBHOOK_SECRET="webhook-secret")
+def test_create_stripe_webhook_event():
+    fake_event = "fake-event"
+    payload = "some payload"
+    sig_header = "some sig_header"
+
+    with mock.patch(
+        "olympia.promoted.utils.stripe.Webhook.construct_event"
+    ) as stripe_construct_event:
```
eviljeff, line 163: Is it possible to have one test that works with some real event data that stripe returns, so it'd be decoded correctly?  i.e. no mocking  (might have to freezegun the time if the signing is time stamped)

docs/topics/api/promoted.rst
```diff
@@ -0,0 +1,21 @@
...
+
+
+--------------
+Stripe Webhook
+--------------
+
+.. _stripe-webhook:
+
+This endpoint receives `event notifications
+<https://stripe.com/docs/webhooks>`_ from Stripe.
```
eviljeff, line 16: random comment that doesn't need action: this could all be on one line (the docs don't have line limits strictly enforced)

eviljeff (approved):

src/olympia/promoted/tests/test_utils.py
willdurand: mm it's tricky because a shared secret is also used to sign the requests.

Issue #15878, opened 2020-10-30
Add Stripe Customer Portal to manage billing


willdurand:
We could use [Stripe Customer Portal](https://stripe.com/docs/billing/subscriptions/customer-portal) to offload billing management to Stripe. This portal would allow us to have a nice webpage to update the card information or email for example (we can do more).

@jvillalobos I think we need this to allow our customers to update their card information in the future. If there is a payment issue, we could tell them to go back to the onboarding page and click the "Manage billing" button:

update: see below (link/button moved to DevHub edit page)

Then they would see this page (on Stripe):

<img width="1619" alt="Screenshot 2020-10-30 at 15 53 11" src="https://user-images.githubusercontent.com/217628/97720018-0aafb200-1ac8-11eb-9dd1-c4de5c846f07.png">

Stripe Customer Portal requires 2 URLs, though:

- a Privacy Policy URL
- a Terms of Service URL

---

For QA: only add-ons that are verified or sponsored should have the link to "manage billing on Stripe". In addition to that, the link should only appear when the subscription is active, which means it has been paid once (when the onboarding flow has been completed).

It shouldn't be possible to change the subscription or cancel it. It should be possible to change the email address or the billing information (credit card).

willdurand:
Yes, it is configurable. The second screenshot shows no way to cancel as we discussed.

jvillalobos:
Okay, that sounds good.

> If there is a payment issue, we could tell them to go back to the onboarding page and click the "Manage billing" button:

I would add a link to the sidebar in the extension edit page instead, below Manage Status & Versions.

> Stripe Customer Portal requires 2 URLs, though:
>    a Privacy Policy URL

https://www.mozilla.org/privacy/websites/

>    a Terms of Service URL:

https://extensionworkshop.com/documentation/publish/firefox-add-on-distribution-agreement/

willdurand:
> I would add a link to the sidebar in the extension edit page instead, below Manage Status & Versions.

Okay, something like this then?

<img width="1046" alt="Screenshot 2020-11-03 at 14 31 32" src="https://user-images.githubusercontent.com/217628/97991251-7b651000-1de1-11eb-92f3-735d7b7031de.png">

jvillalobos:
Yes, but with an upper case "B" in "Billing", please.

ioanarusiczki:
@willdurand 

[...]

PR #15905, opened 2020-11-03
Implement Stripe Customer Portal

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15878

### DIFF
--- a/src/olympia/devhub/templates/devhub/includes/addons_edit_nav.html
+++ b/src/olympia/devhub/templates/devhub/includes/addons_edit_nav.html
@@ -29,6 +29,15 @@
         <li {% if url in request.path|urlencode %}class="selected"{% endif %}>
           <a href="{{ url }}">{{ title }}</a></li>
       {% endfor %}
+
+      {% if addon.promoted_subscription and addon.promoted_subscription.stripe_checkout_completed %}
+      <li class="stripe-customer-portal">
+        <form method="POST" action="{{ url('devhub.addons.subscription_customer_portal', addon.slug) }}">
+          {% csrf_token %}
+          <button class="link" type="submit">{{ _('Manage billing on Stripe') }}</button>
+        </form>
+      </li>
+      {% endif %}
     </ul>
     <ul class="refinements">
       {% if show_listed_fields %}
--- a/src/olympia/devhub/templates/devhub/includes/addons_edit_nav.html
+++ b/src/olympia/devhub/templates/devhub/includes/addons_edit_nav.html
@@ -34,7 +34,7 @@
       <li class="stripe-customer-portal">
         <form method="POST" action="{{ url('devhub.addons.subscription_customer_portal', addon.slug) }}">
           {% csrf_token %}
-          <button class="link" type="submit">{{ _('Manage billing on Stripe') }}</button>
+          <button class="link" type="submit">{{ _('Manage Billing on Stripe') }}</button>
         </form>
       </li>
       {% endif %}
--- a/src/olympia/devhub/tests/test_views_edit.py
+++ b/src/olympia/devhub/tests/test_views_edit.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import datetime
 import json
 import os
 from unittest import mock
@@ -20,7 +21,9 @@ from olympia.amo.tests.test_helpers import get_image_path
 from olympia.amo.urlresolvers import reverse
 from olympia.amo.utils import image_size
 from olympia.constants.categories import CATEGORIES_BY_ID
+from olympia.constants.promoted import VERIFIED
 from olympia.devhub.forms import DescribeForm
+from olympia.promoted.models import PromotedAddon
 from olympia.tags.models import AddonTag, Tag
 from olympia.users.models import UserProfile
 from olympia.versions.models import VersionPreview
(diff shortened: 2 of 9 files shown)


willdurand:
(the error in one of the travis jobs is unrelated)

willdurand:
second commit to address a minor change (replaced a letter)

src/olympia/devhub/views.py
```diff
@@ -1978,3 +1979,33 @@ def onboarding_subscription_cancel(request, addon_id, addon):
...
+    except Exception:
+        log.exception(
+            "error while trying to retrieve a stripe checkout session for "
+            "PromotedSubscription %s (customer_portal).",
+            sub.pk
+        )
+        raise http.Http404()
+
+    try:
+        portal = create_stripe_customer_portal(
```
bobsilverberg, line 2000: Just wondering, why not have `create_stripe_customer_portal` do the `retrieve_stripe_checkout_session`, rather than doing it above?

src/olympia/promoted/utils.py
```diff
@@ -77,3 +77,15 @@ def retrieve_stripe_checkout_session(subscription):
     error when the session does not exist or the API call has failed."""
     stripe.api_key = settings.STRIPE_API_SECRET_KEY
     return stripe.checkout.Session.retrieve(subscription.stripe_session_id)
+
+
+def create_stripe_customer_portal(customer_id, addon):
+    """This function creates a Stripe Customer Portal object or raises an error
+    when the session does not exist or the API call has failed."""
+    stripe.api_key = settings.STRIPE_API_SECRET_KEY
```
bobsilverberg, line 85: I notice that this isn't covered in the test. Is it worth covering, or is it just considered configuration?

src/olympia/promoted/tests/test_utils.py
```diff
@@ -149,3 +150,23 @@ def test_create_stripe_checkout_session_with_custom_rate():
...
+
+def test_create_stripe_customer_portal():
+    addon = addon_factory()
+    customer_id = "some-customer-id"
+    fake_portal = "fake-return-value"
+
+    with mock.patch(
+        "olympia.promoted.utils.stripe.billing_portal.Session.create"
+    ) as create_portal_mock:
+        create_portal_mock.return_value = fake_portal
```
bobsilverberg, line 163: You define a return value here, but never check in the test that it's actually returned.

src/olympia/devhub/tests/test_views_subscription.py
```diff
@@ -358,3 +358,61 @@ def test_get_does_not_set_payment_cancelled_date_when_already_paid(
...
+
+@override_switch("enable-subscriptions-for-promoted-addons", active=True)
+class TestSubscriptionCustomerPortal(SubscriptionTestCase):
+    url_name = "devhub.addons.subscription_customer_portal"
+
+    @override_switch("enable-subscriptions-for-promoted-addons", active=False)
+    def test_returns_404_when_switch_is_disabled(self):
+        assert self.client.post(self.url).status_code == 404
+
+    def test_allows_post_requests(self):
```
bobsilverberg, line 371: This test name doesn't seem to match the contents. The test seems to be checking that the view disallows `get` requests, as opposed to allows `post` requests.

src/olympia/devhub/tests/test_views_edit.py
```diff
@@ -726,6 +729,34 @@ def test_edit_description_does_not_affect_privacy_policy(self):
         assert addon.description_id
         assert addon.description == u'Sométhing descriptive.'
 
+    def test_no_manage_billing_when_no_subscription(self):
+        response = self.client.get(self.url)
+
+        assert pq(response.content)('.stripe-customer-portal').length == 0
+
+    def test_no_manage_billing_when_subscription_process_not_completed(self):
```
bobsilverberg, line 737: The case has no subscription existing. Do you want to have a case where a subscription exists, but it isn't marked as completed? Or will that case never happen?

src/olympia/devhub/templates/devhub/includes/addons_edit_nav.html
```diff
@@ -29,6 +29,15 @@
         <li {% if url in request.path|urlencode %}class="selected"{% endif %}>
           <a href="{{ url }}">{{ title }}</a></li>
       {% endfor %}
+
+      {% if addon.promoted_subscription and addon.promoted_subscription.stripe_checkout_completed %}
+      <li class="stripe-customer-portal">
+        <form method="POST" action="{{ url('devhub.addons.subscription_customer_portal', addon.slug) }}">
```
bobsilverberg, line 35: Why does this need to be a POST, when all the other links are just `<a>` tags?

bobsilverberg (approved):
Looks good, nice work @willdurand. I've added a few comments, but mostly small things.

r+wc

src/olympia/devhub/views.py
willdurand: This helps me return the right response/log the error differently.

src/olympia/promoted/utils.py
willdurand: It's configuration I think

src/olympia/promoted/tests/test_utils.py
willdurand: good call

src/olympia/devhub/tests/test_views_subscription.py
willdurand: Yeah, I see how it can be confusing. I'll update

src/olympia/devhub/templates/devhub/includes/addons_edit_nav.html
willdurand: This is what Stripe recommends to initiate the portal creation.

src/olympia/devhub/tests/test_views_edit.py
willdurand: So, it does have one actually. When we create a `PromotedAddon` with a group like `VERIFIED`, a subscription is automatically created.

PR #15911, opened 2020-11-04
Upgrade Travis and local envs to MySQL 8.0.x, fixing broken tests

### LINKED ISSUE #15898: Upgrade travis and local environments to use MySQL 8.0
We want to prepare the upgrade to MySQL 8.0, starting with travis and local envs. See https://bugzilla.mozilla.org/show_bug.cgi?id=1666431

diox:
Fixes #15898

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,8 +45,8 @@ services:
   - redis
 
 before_install:
-  - wget https://dev.mysql.com/get/mysql-apt-config_0.8.15-1_all.deb
-  - sudo dpkg -i mysql-apt-config_0.8.15-1_all.deb
+  - curl -sL https://dev.mysql.com/get/mysql-apt-config_0.8.15-1_all.deb --output mysql-apt-config.deb
+  - sudo dpkg -i mysql-apt-config.deb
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -41,11 +41,16 @@ cache:
     - $HOME/.gimme
 
 services:
-  - mysql
   - memcached
   - redis
 
 before_install:
+  - wget https://dev.mysql.com/get/mysql-apt-config_0.8.15-1_all.deb
```
willdurand, line 48: Maybe we want an env var for `mysql-apt-config_0.8.15-1_all.deb`. It's referenced twice and that might help with bumping the version up in the future.

willdurand (approved):

.travis.yml
diox: I kinda like that it's not a variable tbh, it's less likely to be accidentally overwritten. I came up with a middle ground approach (and I also applied to elasticsearch downloading, which had the same issue) : f53b71a22eab2e03645600d12a646e2a0315ee66

PR #15915, opened 2020-11-04
Don't allow promoted addons with pending subscriptions to be approved

### LINKED ISSUE #15868: Don't approve versions for promoted when a subscription is in progress
When an add-on has been added to the Verified/Sponsored promoted group but there is a subscription pending we shouldn't auto approve any new versions created in the meantime (or currently pending approval at the time the add-on was added to the group)

We also shouldn't allow addons to be manually approved for verified/sponsored while the subscription is pending.

eviljeff:
fixes #15868

### DIFF
--- a/src/olympia/promoted/models.py
+++ b/src/olympia/promoted/models.py
@@ -83,15 +83,16 @@ class PromotedAddon(ModelBase):
         except AttributeError:
             pass
 
-    def check_subscription_for_approval(self):
+    @property
+    def has_pending_subscription(self):
         """Checks if there is a subscription needed for this promotion, and if
-        so, if it has been completed.  Returns True if there isn't a
-        subscription needed or the subscription is okay."""
+        so, if it has been completed.  Returns True if there is outstanding
+        payment needed."""
         return (
-            not self.group.require_subscription or
-            not (subscr := getattr(self, 'promotedsubscription', None)) or
-            subscr.stripe_checkout_completed or
-            subscr.addon_already_promoted)
+            self.group.require_subscription and
+            (subscr := getattr(self, 'promotedsubscription', None)) and
+            not subscr.stripe_checkout_completed and
+            not subscr.addon_already_promoted)
 
     def approve_for_addon(self):
         """This sets up the addon as approved for the current promoted group.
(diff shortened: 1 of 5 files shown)


src/olympia/promoted/models.py
```diff
@@ -78,6 +78,20 @@ def approve_for_version(self, version):
                 version=version,
                 group_id=self.group_id,
                 application_id=app.id)
+        try:
+            del version.approved_for_groups
+        except AttributeError:
+            pass
+
+    def check_subscription_for_approval(self):
```
bobsilverberg, line 86: Nit: This returns a boolean, but from the name it isn't clear (to me anyway), what True vs. False actually means.

src/olympia/promoted/tests/test_models.py
```diff
@@ -163,6 +163,66 @@ def test_get_resigned_version_number(self):
         assert addon.current_version is None
         assert promo.get_resigned_version_number() is None
 
+    def test_check_subscription_for_approval(self):
```
bobsilverberg, line 166: I'll repeat my comment above and promote it to more than a nit. ;-)  I find these test cases really difficult to parse not knowing what `check_subscription_for_approval` is expected to return. I.e., when is it True and when is it False?

bobsilverberg (approved):
Looks good. I just have an issue with the confusing function name, which made it really hard for me to parse the tests and even some of the code.

r+wc

src/olympia/promoted/models.py
eviljeff: I tried to explain in the comment (just below).  Any suggestions for a better name?  (This is after a few iterations!)

bobsilverberg: I think the big disconnect for me is that this is a boolean, so I think it would be clearer if that name was of the format `isXx` or `hasXx`. So maybe something like `hasPendingSubscription`?

eviljeff: https://github.com/mozilla/addons-server/pull/15915/commits/a85ee788b2834b3504ba3c1f9b84679a014db9e2 - changing the logic around helped, 

> thanks!

Issue #15837, opened 2020-10-26
Handle failed subscription payments


eviljeff:
https://github.com/mozilla/addons-server/issues/15783#issuecomment-715562845 said that users cannot unsubscribe from subscription payments on Stripe but the payments could still end.  Eg. if the payment method expired (credit card expiry date passed), or was cancelled on the bank side (card cancelled) or otherwise failed (payment bounced).

Not sure how this needs to be handled or if it's possible to automate, so logging it while I remember.

willdurand:
We'll handle that manually assuming Stripe sends emails to Stripe admins. If that's not the case, we'll have to listen to some Stripe events and send notifications _via_ AMO.

willdurand:
> We'll handle that manually assuming Stripe sends emails to Stripe admins. If that's not the case, we'll have to listen to some Stripe events and send notifications _via_ AMO.

FWIW, we listen to `charge.failed` Stripe events and send an email to admins.

ioanarusiczki:
@willdurand ID:620297, Sponsored addon: https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/a-test-of-failed-payment

ioanarusiczki:
@willdurand I received an e-mail 
From: Mozilla Add-ons <nobody@mozilla.org> 
To: verified-addons@mozilla.com

Content: 

![failed subscription](https://user-images.githubusercontent.com/33448286/99230522-1b666480-27f8-11eb-816a-a99090b614a4.png)

ioanarusiczki:
@willdurand I'll mark this verified but please take a look at the e-mail content, looks incomplete at the end.

willdurand:
@ioanarusiczki no, it's alright

PR #15918, opened 2020-11-04
Store Stripe subscription ID

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15917

PR #15927, opened 2020-11-05
Listen to "charge.failed" Stripe events and send emails

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15837
~~:warning: depends on https://github.com/mozilla/addons-server/pull/15918~~

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1128,6 +1128,7 @@ CELERY_TASK_ROUTES = {
     'olympia.devhub.tasks.validate_file': {'queue': 'devhub'},
     'olympia.devhub.tasks.validate_upload': {'queue': 'devhub'},
     'olympia.files.tasks.repack_fileupload': {'queue': 'devhub'},
+    'olympia.promoted.tasks.on_stripe_charge_failed': {'queue': 'devhub'},
     'olympia.scanners.tasks.run_customs': {'queue': 'devhub'},
     'olympia.scanners.tasks.run_wat': {'queue': 'devhub'},
     'olympia.scanners.tasks.run_yara': {'queue': 'devhub'},
--- a/src/olympia/promoted/tasks.py
+++ b/src/olympia/promoted/tasks.py
@@ -1,0 +1,108 @@
+import olympia.core.logger
+
+from django.conf import settings
+from django.template import loader
+
+from olympia.amo.celery import task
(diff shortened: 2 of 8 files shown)


willdurand:
rebased!

willdurand:
(rebased)

src/olympia/promoted/tasks.py
```diff
@@ -0,0 +1,108 @@
...
+            event_type,
+            event_id,
+            charge_id,
+        )
+        return
+
+    addon = sub.promoted_addon.addon
+
+    # Create the Stripe URL pointing to the Stripe subscription.
+    stripe_sub_url = "https://dashboard.stripe.com"
```
diox, line 85: Nit: we don't enforce it so it's fine to keep as-is, but addons-server is using single quotes wherever possible

src/olympia/promoted/tasks.py
```diff
@@ -0,0 +1,108 @@
...
+    if not stripe_sub["livemode"]:
+        stripe_sub_url = stripe_sub_url + "/test"
+    stripe_sub_url = stripe_sub_url + f"/subscriptions/{subscription_id}"
```
diox, line 88: Nit:

```suggestion
    stripe_sub_url = f"{stripe_sub_url}/subscriptions/{subscription_id}"
```

Should work (and above as well)

src/olympia/lib/settings_base.py
```diff
@@ -1128,6 +1128,7 @@ def get_db_config(environ_var, atomic_requests=True):
     'olympia.devhub.tasks.validate_file': {'queue': 'devhub'},
     'olympia.devhub.tasks.validate_upload': {'queue': 'devhub'},
     'olympia.files.tasks.repack_fileupload': {'queue': 'devhub'},
+    'olympia.promoted.tasks.on_stripe_charge_failed': {'queue': 'devhub'},
```
diox, line 1131: I wonder if that's the best queue for it... I suppose it's fine for now - ultimately I hope admins will not rely on this notification only but also periodically visit the Stripe dashboard...

diox (approved):
r+, just nits.

src/olympia/lib/settings_base.py
willdurand: Yeah, I don't know either.. I didn't want to create a new queue for that right now, though.

PR #15928, opened 2020-11-05
log when azderk results aren't valid sponsored addons in ES

### LINKED ISSUE #15919: Add logging around discrepancies between Adzerk and Elasticsearch in sponsored helves API
The [Sponsored Shelves API](https://addons-server.readthedocs.io/en/latest/topics/api/shelves.html#sponsored-shelf) uses Adzerk to retrieve which add-ons to show (Adzerk returns the ids) and then fetches the data from Elasticsearch, forcing add-ons that are returned to be promoted and public.

If somehow an add-on is present in the Adzerk data but not returned when we perform the Elasticsearch query, something weird is going on and we should log that to investigate later.

eviljeff:
fixes #15919

### DIFF
--- a/src/olympia/shelves/tests/test_views.py
+++ b/src/olympia/shelves/tests/test_views.py
@@ -231,7 +231,9 @@ class TestSponsoredShelfViewSet(ESTestCase):
                     'impression': '',
                     'click': ''},
             }
-            data = self.perform_search()
+            with mock.patch('django_statsd.clients.statsd.incr') as incr_mock:
+                data = self.perform_search()
+            incr_mock.assert_any_call('services.adzerk.elasticsearch_miss', 3)
             get.assert_called_with(6)
         # non sponsored are ignored
         assert data['count'] == 2
--- a/src/olympia/shelves/tests/test_views.py
+++ b/src/olympia/shelves/tests/test_views.py
@@ -213,6 +213,9 @@ class TestSponsoredShelfViewSet(ESTestCase):
     def test_adzerk_returns_none_sponsored(self):
         self.populate_es()
         signer = TimestampSigner()
+        log_msg = (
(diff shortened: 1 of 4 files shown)


src/olympia/shelves/tests/test_views.py
```diff
@@ -231,7 +231,9 @@ def test_adzerk_returns_none_sponsored(self):
                     'impression': '',
                     'click': ''},
             }
-            data = self.perform_search()
+            with mock.patch('django_statsd.clients.statsd.incr') as incr_mock:
+                data = self.perform_search()
+            incr_mock.assert_any_call('services.adzerk.elasticsearch_miss', 3)
```
bobsilverberg, line 236: Do we need to assert about the log call as well? I realize we probably don't assert about log calls much, but this is specifically being added to address an issue from the RRA, as opposed to just support possible debugging. We wouldn't want it accidentally removed.

bobsilverberg (approved):
This looks good, thanks @eviljeff!  Just one question about test coverage.

r+wc

bobsilverberg:
Also, we should make sure a Grafana alert is set up as per https://github.com/mozilla/addons-server/issues/15919#issuecomment-721927738.

eviljeff:
> Also, we should make sure a Grafana alert is set up as per [#15919 (comment)](https://github.com/mozilla/addons-server/issues/15919#issuecomment-721927738).

grafana dashaboards (let alone alerts) are difficult to set up, and impossible to test, without the metrics being generated first by the code being deployed.  It'll have to be done later.

src/olympia/shelves/tests/test_views.py
eviljeff: https://github.com/mozilla/addons-server/pull/15928/commits/18eccf2fce952c87ff04d63ca84c70972fe02a1a

PR #15932, opened 2020-11-05
Use MySQL client & libs, not MariaDB

### LINKED ISSUE #15929: Setting up local developer environment from scratch fails with mysql 8
When setting up a local dev enviromnent, setup fails during `python3 manage.py migrate --noinput`:

`MySQLdb._exceptions.OperationalError: (1156, 'Plugin caching_sha2_password could not be loaded: /usr/lib/x86_64-linux-gnu/mariadb19/plugin/caching_sha2_password.so: cannot open shared object file: No such file or directory')`

This is due to the changed default auth method in mysql 8: https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password

A workaround is to manually connect to the db and `ALTER` the user root as follows:

```
$ docker-compose exec mysqld mysql
mysql> alter user 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY '';
mysql> alter user 'root'@'%' IDENTIFIED WITH mysql_native_password BY '';
```

diox:
Fixes #15929

We're using MySQL 8.0 from MySQL, not MariaDB, so client libs should follow for consistency - especially as MariaDB doesn't support the default auth method MySQL 8 uses.

Also contains a slight refactor of the Dockerfile-s to have fewer steps and be more consistent with each other.


Dockerfile
```diff
@@ -11,27 +11,32 @@ ARG USER_ID=1000
...
 RUN apt-get update && apt-get install -y \
         apt-transport-https              \
         gnupg2                           \
     && rm -rf /var/lib/apt/lists/*
-RUN cat /etc/pki/gpg/GPG-KEY-nodesource | apt-key add -
-ADD docker/debian-buster-nodesource-repo /etc/apt/sources.list.d/nodesource.list
-ADD docker/debian-buster-backports-repo /etc/apt/sources.list.d/buster-backports.list
+# Add keys and repos for node and mysql
+COPY docker/*.gpg.key /etc/pki/gpg/
+RUN APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=DontWarn \
```
willdurand, line 21: This is new, right?

Dockerfile.deploy
```diff
@@ -8,59 +8,66 @@ ENV SWIG_FEATURES="-D__x86_64__"
...
 ADD docker/debian-buster-backports-repo /etc/apt/sources.list.d/buster-backports.list
+ADD docker/debian-buster-mysql-repo /etc/apt/sources.list.d/buster-mysql.list
 
 # IMPORTANT: When editing this list below, make sure to also update
-# `Dockerfile`.
-RUN apt-get update && apt-get -t buster install -y \
+# `Dockerfile.deploy`.
+# Allow scripts to detect we're running in our own container and install
+# packages.
+RUN touch /addons-server-docker-container \
```
willdurand, line 31: Do we want that file in the image?

willdurand (approved):
r+wc

Dockerfile.deploy
diox: Yes. The `Makefile` uses that to figure out if we're inside a container or not.

Dockerfile
diox: This is yak shaving : I'm adding a new gpg key, but I didn't want to do that using 2 docker commands, so I passed it using a file instead of stdin, but I was annoyed at the output of `apt-key`, so I use that to shut it down.

Dockerfile.deploy
eviljeff: mmm, but why are we adding this in `Dockerfile.deploy` though?  The deployment pipeline doesn't use `make` (or if it does it specifies the correct makefile explicitly)

diox: Just consistency, and ensuring if we do ask ops to run `make <something>` manually later on it works as expected.

Issue #15933, opened 2020-11-05
Use `stripe_subscription_id` whenever possible


willdurand:
Now that we store the `stripe_subscription_id` (see #15917), we can and should use it in most cases. The Customer Portal session is created using the Checkout session, which is a bit odd now that we have a stripe subscription ID so let's change that.

For QA: there shouldn't be any visible changes.

ioanarusiczki:
@willdurand I don't see anything changed since Friday when I tested it https://github.com/mozilla/addons-server/issues/15878#issuecomment-723131999

I've noticed that the admin can visit Manage Billing on Stripe URL for developers from Edit Product Page. Should this be expected? It's personal data about cards and so on.

willdurand:
> I've noticed that the admin can visit Manage Billing on Stripe URL for developers from Edit Product Page. Should this be expected? It's personal data about cards and so on

Good question. Technically, amo admins will likely be stripe admins too. They'll have access to cards in the stripe dashboard as well. In addition, the customer portal does not reveal all the information. We'll restrict this page and other onboarding pages to owners, though.

@jvillalobos WDYT?

jvillalobos:
Most admins also have access to the django admin and the DB, where the subscription ID is stored. I don't think we can do much to protect those URLs against admins, and as Will mentioned, we will have access to that information in other ways.

PR #15934, opened 2020-11-05
Use stripe_subscription_id whenever possible

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15933


willdurand:
The failure seems unrelated (and more frequent than before)

diox (approved):

Issue #15921, opened 2020-11-04
support versions being un-signed as promoted when they're remove from the promotional group


eviljeff:
see https://github.com/mozilla/addons-server/issues/15805#issuecomment-718994437 and earlier comments

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-47

PR #15947, opened 2020-11-06
Handle subscription cancellation

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15805

---

WIP but that's how I imagine it will work.


src/olympia/promoted/tasks.py
```diff
@@ -106,3 +109,65 @@ def on_stripe_charge_failed(event):
...
+        return
+
+    sub.update(
+        cancelled_at=datetime.datetime.fromtimestamp(
+            subscription["canceled_at"]
+        )
+    )
+    log.info("PromotedSubscription %s has been marked as cancelled.", sub.id)
+
+    # TODO: this should trigger resigning somehow, see:
```
willdurand, line 168: @eviljeff FYI but also a question: should we try to land https://github.com/mozilla/addons-server/issues/15921 first?

src/olympia/promoted/tasks.py
```diff
@@ -106,3 +109,65 @@ def on_stripe_charge_failed(event):
...
+    )
+
+    try:
+        sub = PromotedSubscription.objects.get(
+            stripe_subscription_id=subscription_id
+        )
+    except PromotedSubscription.DoesNotExist:
+        log.error(
+            'received a "%s" event (event_id=%s) for a non-existent'
+            " promoted subscription (subscription_id=%s).",
```
eviljeff, line 145: nits: space at the end of lines rather than start; mixed `"` and `'` for no apparent reason.

willdurand: Quotes are added/adjusted by Black, which I run locally on new files I create.

src/olympia/promoted/tasks.py
eviljeff: It doesn't seem a strict prerequisite so I'd say no.

src/olympia/promoted/tasks.py
willdurand: I am still hoping to adopt Black soon-ish so I let you folks prepare for that given that Black isn't going to support single quotes (it enforces double quotes most of the time)

eviljeff: I half read through that issue before getting bored - I summarise it as "we made a decision and won't back down out of stubbornness".  It's clear that single quotes are the python default:
```python
>>> foo = "foo"
>>> foo
'foo'
```
😛 (in addition to the many other examples in the issue)

I strongly suspect if we do ever deploy Black on this repo, it'll be with `--skip-string-normalization`.

willdurand: > I summarise it as "we made a decision and won't back down out of stubbornness". 

I agree.

> I strongly suspect if we do ever deploy Black on this repo, it'll be with --skip-string-normalization.

I'd prefer not to skip that, otherwise why having a code formatter if we don't use it fully? Anyway, we'll see if I can make another random act of kindness this quarter :)

src/olympia/devhub/templates/devhub/includes/addons_edit_nav.html
```diff
@@ -30,7 +30,7 @@
           <a href="{{ url }}">{{ title }}</a></li>
       {% endfor %}
 
-      {% if addon.promoted_subscription and addon.promoted_subscription.stripe_checkout_completed %}
+      {% if addon.promoted_subscription and addon.promoted_subscription.is_active == True %}
```
eviljeff, line 33:
```suggestion
      {% if addon.promoted_subscription and addon.promoted_subscription.is_active %}
```
(it's only `False` and `None` that need disambiguating)

src/olympia/devhub/views.py
```diff
@@ -1858,7 +1858,12 @@ def get_promoted_subscription_or_404(addon):
         raise http.Http404()
 
     qs = PromotedSubscription.objects.select_related('promoted_addon__addon')
-    return get_object_or_404(qs, promoted_addon__addon=addon)
+    subscription = get_object_or_404(qs, promoted_addon__addon=addon)
```
eviljeff, line 1861: for an efficiency gain we could filter by subscriptions that are `stripe_checkout_completed` and not `cancelled_at` directly here (would need `Q` objects I would have thought).  Would make it a little less readable though.

src/olympia/devhub/templates/devhub/includes/addons_edit_nav.html
eviljeff: or, if you really want to be explicit, `is True`

src/olympia/promoted/tasks.py
```diff
@@ -106,3 +110,66 @@ def on_stripe_charge_failed(event):
...
+            event_type,
+            event_id,
+            subscription_id,
+        )
+        return
+
+    if sub.cancelled_at:
+        log.info(
+            "PromotedSubscription %s already cancelled, ignoring event"
+            " (event_id=%s).",
```
eviljeff, line 157:
```suggestion
            "PromotedSubscription %s already cancelled, ignoring event "
            "(event_id=%s).",
```
assuming Black doesn't also mis-format strings over multiple lines 😛

src/olympia/devhub/views.py
eviljeff: This will be needed anyway if there's a future where we have multiple `PromotedSubscription` (one is cancelled, and later a new one starts).

src/olympia/promoted/models.py
```diff
@@ -92,7 +92,7 @@ def has_pending_subscription(self):
         return (
             self.group.require_subscription and
             (subscr := getattr(self, 'promotedsubscription', None)) and
-            not subscr.stripe_checkout_completed and
+            not subscr.is_active and
```
eviljeff, line 95: can you add something to `TestPromotedAddon.test_has_pending_subscription` that covers the subscription being cancelled?  (or change the asserts on `stripe_checkout_completed` to `is_active`)

src/olympia/promoted/tasks.py
```diff
@@ -106,3 +110,66 @@ def on_stripe_charge_failed(event):
...
+            subscription["canceled_at"]
+        )
+    )
+    log.info("PromotedSubscription %s has been marked as cancelled.", sub.id)
+
+    # TODO: this should trigger resigning somehow, see:
+    # https://github.com/mozilla/addons-server/issues/15921
+    sub.promoted_addon.update(group_id=NOT_PROMOTED.id)
+    log.info(
+        "PromotedAddon %s is not promoted anymore.", sub.promoted_addon.id
```
eviljeff, line 174: I'd lean towards saying we should be logging `Addon.id` here, rather then `PromotedAddon.id`, but is that consistent with other logging you've added?

eviljeff (approved):

src/olympia/devhub/views.py
willdurand: I think that wouldn't work for the `cancel` endpoint, which wouldn't have `stripe_checkout_completed` set yet.

> This will be needed anyway if there's a future where we have multiple PromotedSubscription (one is cancelled, and later a new one starts).

I agree. This is something I'd like to investigate in the near future, but I am not sure how to do that so I'll start by creating an issue.

src/olympia/promoted/tasks.py
willdurand: I added the addon id, the promoted addon id and the promoted sub id

src/olympia/devhub/views.py
willdurand: see: https://github.com/mozilla/addons-server/issues/15967

eviljeff: > I think that wouldn't work for the cancel endpoint, which wouldn't have stripe_checkout_completed

I was meaning: replicate the exact logic, so it would be not stripe_checkout_completed or stripe_checkout_completed and not cancelled_at.

PR #15959, opened 2020-11-09
Add ability to restrict add-on detail/version API and downloads in specific regions

### LINKED ISSUE #15942: Restrict add-on detail/versions API and download views for add-ons in specific regions
addons-server companion to https://github.com/mozilla/addons-frontend/issues/9842, we need some add-ons to be restricted in some specific regions.

addons-frontend and nginx (*) will pass a `X-Country-Code` header to API requests to the add-on detail and versions API requests indicating the region the client request is originating from, in [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) format. We need to add a table containing add-on region exclusions for a given add-on, probably just stored as flat json list of country codes. When an add-on is requested from a restricted country, we need to return a HTTP 451 response instead of the normal detail page response.

This same logic also needs to be added to download views.

As a follow-up, we'll likely need something to be added to be able to bypass this if logged in as an admin, while adding something to the response to indicate to the frontend that we bypassed the restriction, but this doesn't have to be implemented right away.

(*) We will have to let clients override the header, because of addons-frontend server side rendering. We may have some config in nginx to only allow that under specific circumstances though, to prevent clients from playing with this.

diox:
Fixes #15942


diox:
Closing for now as I'm investigating a completely different alternative.

Issue #15948, opened 2020-11-06
Limit access to Manage Billing link and Onboarding to developers with an Owner role


jvillalobos:
Following up on https://github.com/mozilla/addons-server/issues/15878, I think we need to tighten access restrictions to it. The sidebar link should only show up if you're a developer for the add-on with an Owner role.

Talking to @willdurand about this, it would also be good to limit access to the onboarding page to the Owner role. This one I'm less concerned about since it's not a permanent fixture like the sidebar link, but it's a good idea to maintain consistency.

ioanarusiczki:
@willdurand 

- to the onboarding URL: owners and developers still have access to it, I added users to this [addon](https://addons-internal.dev.mozaws.net/en-US/admin/models/addons/addon/620287/change/). Same for Billing on Stripe URL when I added other owners/developers to a verified addon [addon](https://addons-internal.dev.mozaws.net/en-US/admin/models/addons/addon/620282/change/).

ioanarusiczki:
@willdurand I verified this issue again on AMO dev - FF82(Win10):

The issue mentioned above is no longer reproducible meaning: developers cannot access the onboarding URL or manager customer portal URL. 

Only the owners (no matter how many added, they can be more than one) can access both URLs, but reading the [comment](https://github.com/mozilla/addons-server/issues/15948#issue-737940928) I am not sure it's the expected.

PR #15961, opened 2020-11-09
Restrict subscription pages to owners in DevHub

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15948


bobsilverberg (approved):
Looks good. I'm not a huge fan of `owner_for_get` as an argument name as it doesn't seem to really explain what it's for, but I see we already have `owner_for_post`, so it's consistent at least.

PR #15965, opened 2020-11-10
Add ability to restrict add-on detail/version API and downloads in specific regions

### LINKED ISSUE #15942: Restrict add-on detail/versions API and download views for add-ons in specific regions

diox:
Fixes #15942

Depends on nginx config change to automatically pass the `X-Country-Code` header (and addons-frontend changes to pass that same header when doing server side rendering requests, as well as handling the 451 errors)


src/olympia/api/middleware.py
```diff
@@ -2,16 +2,22 @@
...
+class APIRequestMiddleware(MiddlewareMixin):
     def identify_request(self, request):
         request.is_api = re.match(settings.DRF_API_REGEX, request.path_info)
 
     def process_request(self, request):
         self.identify_request(request)
 
+    def process_response(self, request, response):
+        if request.is_api:
+            patch_vary_headers(response, ['X-Country-Code'])
```
eviljeff, line 18: _all_ api requests?

src/olympia/api/exceptions.py
```diff
@@ -4,13 +4,20 @@
...
 from rest_framework import exceptions
 from rest_framework import status
 from rest_framework.response import Response
 from rest_framework.views import set_rollback
 
 
+class UnavailableForLegalReasons(exceptions.APIException):
+    status_code = status.HTTP_451_UNAVAILABLE_FOR_LEGAL_REASONS
+    default_detail = _('Unavailable for legal reasons.')
+    default_code = 'unavailable_for_legal_reasons'
```
eviljeff, line 18: add the Link header as suggested?  https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/451  (it's a "SHOULD" in https://tools.ietf.org/html/rfc7725#section-4)

src/olympia/addons/views.py
```diff
@@ -224,6 +230,13 @@ def check_object_permissions(self, request, obj):
                 'is_disabled_by_mozilla': obj.status == amo.STATUS_DISABLED,
             }
             raise exc
+        region_code = (
+            self.request and self.request.META.get(
+                'HTTP_X_COUNTRY_CODE', None))
+        if region_code and AddonRegionalRestrictions.objects.filter(
+                addon=obj,
+                excluded_regions__contains=region_code.upper()).exists():
```
eviljeff, line 238: Once we start getting this for all requests it'll probably be more efficient to use `select_related` on `'regional_restrictions'` vs. a separate query (as a follow-up if so)

src/olympia/addons/tests/test_views.py
```diff
@@ -251,6 +252,17 @@ def test_get_not_listed(self):
         assert data['is_disabled_by_developer'] is False
         assert data['is_disabled_by_mozilla'] is False
 
+        AddonRegionalRestrictions.objects.filter(
+            addon=self.addon).update(excluded_regions=['AB', 'CD', 'FR'])
+        # Regional restrictions should be processed after other permission
+        # handling, so something that would return a 401/403/404 without
+        # region restrictions would still do that.
+        response = self.client.get(self.url)
```
bobsilverberg, line 260: Just a drive-by comment, but where is one of `['AB', 'CD', 'FR']` set as `X-Country-Code` for the request? I.e., what reason do we have to believe the test will care that these have been added to `excluded_regions`?

src/olympia/versions/tests/test_views.py
```diff
@@ -179,6 +188,14 @@ def test_download_for_unlisted_addon_owner(self):
         self.assert_served_internally(
             self.client.get(url), False, attachment=True)
 
+        # Even allowed to bypass georestrictions.
```
bobsilverberg, line 191: Another drive-by question: where are all of the exclusions tested for in this file defined, in terms of requirements?

src/olympia/api/exceptions.py
willdurand: +1

src/olympia/versions/views.py
```diff
@@ -98,14 +99,24 @@ def is_appropriate_reviewer(addon, channel):
...
-        loc = urlparams(file_.get_file_cdn_url(attachment=attachment),
-                        filehash=file_.hash)
-        response = http.HttpResponseRedirect(loc)
-        response['X-Target-Digest'] = file_.hash
+        # When serving the file for the general public through the CDN, we need
+        # to obey regional restrictions
+        region_code = request.META.get('HTTP_X_COUNTRY_CODE', None)
+        if region_code and AddonRegionalRestrictions.objects.filter(
+                addon=addon,
+                excluded_regions__contains=region_code.upper()).exists():
```
willdurand, line 107: Shouldn't we use `icontains` here as well?

src/olympia/addons/views.py
willdurand: Also, shouldn't we use `icontains` here?

src/olympia/versions/views.py
```diff
@@ -98,14 +99,24 @@ def is_appropriate_reviewer(addon, channel):
...
-        # Content-Disposition: attachment for attachments. To work around this,
-        # if attachment=True, get_file_cdn_url() changes the path to something
-        # we recognize in the nginx config.
-        loc = urlparams(file_.get_file_cdn_url(attachment=attachment),
-                        filehash=file_.hash)
-        response = http.HttpResponseRedirect(loc)
-        response['X-Target-Digest'] = file_.hash
+        # When serving the file for the general public through the CDN, we need
+        # to obey regional restrictions
+        region_code = request.META.get('HTTP_X_COUNTRY_CODE', None)
```
willdurand, line 104: It's interesting that we call the header "country code" but the variable is "region code".

willdurand (approved):
r+wc

src/olympia/api/middleware.py
diox: It's just a `Vary` on a header that shouldn't be sent very often, so I think it's safe

diox: It's a lot easier to add there - we might return the 451 for anything that depend on an add-on, but we don't have a centralized way of returning the `Vary` header otherwise

src/olympia/addons/views.py
diox: Doing `icontains` would make the query a ` LIKE %` (which could lead to false positives) instead of a `JSON_CONTAINS()`

src/olympia/addons/tests/test_views.py
diox: It's a good point, I manually passed the header for `test_addon_regional_restrictions` below but forgot it there (and in a couple other places). I'll fix that.

src/olympia/versions/tests/test_views.py
diox: What do you mean ?

src/olympia/versions/views.py
diox: https://github.com/mozilla/addons-server/pull/15965/files#r520608452

src/olympia/versions/tests/test_views.py
bobsilverberg: I mean, who says that we should be doing this exclusion? Who says that it's okay for an owner in a restricted region to be able to download their add-on? (Or maybe I'm just not understanding what's being tested here)

src/olympia/versions/views.py
diox: Yeah. We called the header country code because it follows https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2 but I didn't want to call the model/field country because it was a bit of an oversimplification IMHO - technically, a different country code would be sent for special collectivities and territories, even if it's the same "country".

For instance, Guadeloupe has its own ISO 3166-1 alpha-2 country code, despite being part of France.

src/olympia/addons/views.py
willdurand: Interesting. how do we ensure that entries will be correct in the JSON then?

eviljeff: a big stick gestured at whoever updates the data 😜

src/olympia/api/exceptions.py
diox: I can look into that, but I'm not sure where to link to. https://www.mozilla.org/en-US/about/policy/transparency/ ?

eviljeff: Sounds good.

src/olympia/versions/tests/test_views.py
diox: It's not defined anywhere so I made the call myself, mostly to make it easier to implement a bypass for admins & reviewers. I don't think it matters much, and it matches what we allow for disabled, so that makes sense.

src/olympia/addons/views.py
diox: Something to ensure in https://github.com/mozilla/addons-server/issues/15968 in the long run

src/olympia/api/exceptions.py
diox: ffb10974610e186d7f8cfeff99de2f88d4bfd01b

src/olympia/addons/tests/test_views.py
diox: 7226d9cd3e49743f4c28b8286a7f8e4ca6aff569

Issue #15966, opened 2020-11-10
Review Tools showing files as blank


erosman:
I have noticed a number of files that show up as blank in comparison.

![image](https://user-images.githubusercontent.com/6458876/98673555-bda3c980-236c-11eb-9c56-afd961bb1139.png)

Downloading the add-on, shows that the files were not blank.

Example:
Add-on:
`https://addons.mozilla.org/en-US/firefox/addon/pepega-catch/`
Compare:
`https://code.addons.mozilla.org/en-US/compare/2592637/versions/5004720...5104530/?path=src%2Fcontent%2Fcontent.js`

Other files showing as blank in above add-on:

battleBreakdown.html
fusionRecipes.html
popup.html

willdurand:
The problem comes from whitespace changes. Looks like the file is only reformatted and we ignore whitespace changes sometimes. We aren't able to reconstruct the entire file because the status of the delta is not marked as unmodified (it is actually modified).

The good news is that we should be able to write a test case for that. The bad news is that I am not sure how to fix the issue.

willdurand:
Looks like the change related to whitespaces has been introduced in https://github.com/mozilla/addons-server/commit/4f7a5f7e2c174463f6cb585cb206c48d5dbdb782. This is a consequence of https://github.com/mozilla/addons-server/pull/13680.

erosman:
The situation is not limited to one add-on. Most files in the following also show as blank.

`https://code.addons.mozilla.org/en-US/compare/2666547/versions/5129470...5131839/?path=src%2Fjs%2Fbackground.js`

ioanarusiczki:
@willdurand I verified on AMO dev, FF82(Win10) - the files mentioned by the issue (for both addons) in comparison with AMO stage.

Examples:
https://code.addons-dev.allizom.org/en-US/compare/620286/versions/1692283...1692284/?path=src%2Fpopup%2FfusionRecipes%2FfusionRecipes.html
https://code.addons-dev.allizom.org/en-US/compare/620286/versions/1692283...1692284/?path=src%2Fpopup%2Fpopup.html
https://code.addons-dev.allizom.org/en-US/compare/620286/versions/1692283...1692284/?path=src%2Fcontent%2Fcontent.js

For the other addon: many .css and .js files were not visible in compare view but now they're displayed on -dev 

https://code.addons-dev.allizom.org/en-US/compare/620301/versions/1692301...1692302/?path=src%2Fcss%2Findeed.css
[...]

PR #15969, opened 2020-11-10
Return fake diff when entire file has whitespace changes only

willdurand:
fixes https://github.com/mozilla/addons-server/issues/15966


willdurand:
@diox here is an alternative solution

diox:
I haven't tried it yet but that looks promising!

willdurand:
I just tried this patch fully by re-uploading an add-on and two version from -dev locally and that does what it is supposed to do.

bobsilverberg (approved):
Looks good. Thanks for taking care of this, @willdurand!

Issue #15982, opened 2020-11-11
Add link to Stripe subscription in django admin


willdurand:
We should display a link to easily access the Stripe subscription (on Stripe) linked to a promoted subscription in the django admin.

ioanarusiczki:
@willdurand On AMO dev: Stripe information fields and "View subscription on Stripe" URL is available after a subscription is completed. 

I get redirected to https://dashboard.stripe.com/login?redirect=%2Fsubscriptions%2Fsub_IOslKBwgsHKJPt -> login page, since I do not have access to Stripe admin.

PR #15983, opened 2020-11-11
Add link to Stripe subscription in django admin

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/15982

### DIFF
--- a/src/olympia/promoted/admin.py
+++ b/src/olympia/promoted/admin.py
@@ -65,7 +65,7 @@ class PromotedSubscriptionInline(admin.StackedInline):
     onboarding_url.short_description = "Onboarding URL"
 
     def stripe_information(self, obj):
-        if not obj or (obj and not obj.stripe_subscription_id):
+        if not obj or not obj.stripe_subscription_id:
             return '-'
 
         stripe_sub_url = '/'.join([
--- a/src/olympia/promoted/admin.py
+++ b/src/olympia/promoted/admin.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.contrib import admin
 from django.db.models import Prefetch
 from django.forms.models import modelformset_factory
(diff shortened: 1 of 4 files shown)


src/olympia/promoted/admin.py
```diff
@@ -32,13 +33,15 @@ class PromotedSubscriptionInline(admin.StackedInline):
         'checkout_cancelled_at',
         'checkout_completed_at',
         'cancelled_at',
+        'stripe_information',
```
willdurand, line 36: I'm thinking that we might have more info to display later, hence the generic custom field

src/olympia/promoted/admin.py
```diff
@@ -61,6 +64,23 @@ def onboarding_url(self, obj):
 
     onboarding_url.short_description = "Onboarding URL"
 
+    def stripe_information(self, obj):
+        if not obj or (obj and not obj.stripe_subscription_id):
```
eviljeff, line 68:
```suggestion
        if not obj or not obj.stripe_subscription_id:
```

src/olympia/promoted/admin.py
```diff
@@ -61,6 +64,23 @@ def onboarding_url(self, obj):
...
+            return '-'
+
+        stripe_sub_url = '/'.join([
+            settings.STRIPE_DASHBOARD_URL,
+            'subscriptions',
+            obj.stripe_subscription_id
+        ])
```
eviljeff, line 75: there's some extra stuff with test mode in tasks.py - do we need to take that into account here too?

eviljeff (approved):

src/olympia/promoted/admin.py
willdurand: I don't think so. In `tasks.py` that's because we have an API response, which we don't have here and issuing a request for test mode seems overkill to me. In addition, I noticed our -dev Stripe account shows a page to decide whether to go to the test mode or not (in case it does not find the subscription in the default mode)

PR #15993, opened 2020-11-12
pass on new version number in session; replace addon_already_promoted

### LINKED ISSUE #15867: Expose the version we're re-signing to the developer in the subscription payment page

eviljeff:
fixes #15867

### DIFF
--- a/src/olympia/devhub/tests/test_views_subscription.py
+++ b/src/olympia/devhub/tests/test_views_subscription.py
@@ -69,7 +69,7 @@ class TestOnboardingSubscription(SubscriptionTestCase):
 
     @mock.patch("olympia.devhub.views.create_stripe_checkout_session")
     def test_get_for_the_first_time(self, create_mock):
-        create_mock.return_value = mock.MagicMock(id="session-id")
+        create_mock.return_value = dict(id="session-id")
 
         assert not self.subscription.link_visited_at
         assert not self.subscription.stripe_session_id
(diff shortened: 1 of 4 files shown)


src/olympia/devhub/tests/test_views_subscription.py
```diff
@@ -99,12 +99,12 @@ def test_get_for_the_first_time(self, create_mock):
...
         create_mock.side_effect = [
-            mock.MagicMock(id="session-id-1"),
-            mock.MagicMock(id="session-id-2"),
+            dict(id="session-id-1"),
+            dict(id="session-id-2"),
         ]
 
         # Get the page.
-        queries = 37
+        queries = 36
```
willdurand, line 107: 🏆

src/olympia/devhub/views.py
```diff
@@ -1903,21 +1903,19 @@ def onboarding_subscription(request, addon_id, addon):
...
-    if sub.stripe_session_id != session.id:
-        fields_to_update['stripe_session_id'] = session.id
+    if sub.stripe_session_id != session['id']:
+        fields_to_update['stripe_session_id'] = session['id']
 
     if len(fields_to_update) > 0:
         sub.update(**fields_to_update)
 
-    existing_version_pending = addon.versions.filter(
+    new_version_number = request.session.get('resigned_version_string')
```
willdurand, line 1912: Shouldn't we clear the session after that? I think we should because otherwise the page will always display content like:

```
 You're done! A new version 1.5.1-signed of your add-on textmarkerpro will be published as Sponsored and should become available shortly. If you have any questions, don’t hesitate to contact us.
```

I'd expect to see the "A new version [...]" message once. It should be gone after page refresh.

willdurand (approved):
r+wc

src/olympia/devhub/views.py
eviljeff: I wasn't really sure how long the session would persist for - and how long we'd want to keep the message around for - but I guess it's best to clear it to tidy it up

eviljeff: https://github.com/mozilla/addons-server/pull/15993/commits/811d8a62e84e95d847add46d5f4f03693aeca0e9

PR #16012, opened 2020-11-16
create VERSION_RESIGNED log as task user

### LINKED ISSUE #15981: Show the version being resigned by Mozilla
As a followup for https://github.com/mozilla/addons-server/issues/15869#issuecomment-725437106

1. versions being resigned after a subscription should show "By Mozilla" instead of the name of the developer who submitted the version. 

An example: 
the developer

Another example: manually approved by and admin then resigned https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/a-cancel-button

eviljeff:
fixes #15981

### DIFF
--- a/src/olympia/devhub/tests/test_views_subscription.py
+++ b/src/olympia/devhub/tests/test_views_subscription.py
@@ -321,6 +321,8 @@ class TestOnboardingSubscriptionSuccess(SubscriptionTestCase):
 
     @mock.patch("olympia.devhub.views.retrieve_stripe_checkout_session")
     def test_get_records_payment_once(self, retrieve_mock):
+        self.subscription.promoted_addon.approve_for_version(
+            self.subscription.promoted_addon.addon.current_version)
         stripe_subscription_id = "some-subscription-id"
         retrieve_mock.return_value = {
             "id": "session-id",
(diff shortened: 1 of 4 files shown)


src/olympia/devhub/tests/test_views_subscription.py
```diff
@@ -321,6 +321,8 @@ def test_get_redirects_to_main_page(self, retrieve_mock):
 
     @mock.patch("olympia.devhub.views.retrieve_stripe_checkout_session")
     def test_get_records_payment_once(self, retrieve_mock):
+        self.subscription.promoted_addon.approve_for_version(
```
bobsilverberg, line 324: Why are the changes in this file needed for this patch?

bobsilverberg (approved):
Looks good @eviljeff, thanks!

Just added a question as I wasn't clear why those statements needed to be added to `test_views_subscription.py` for this patch.

r+wc

src/olympia/devhub/tests/test_views_subscription.py
eviljeff: Those tests were triggering a resign once that was added in an earlier patch - it just wasn't apparent until now.  I could have created the task user for the tests to pass, but they didn't need to do the resign for the functionality being tested, so I pre-approved the version to skip the resigning instead (mocking the resigning task would have worked too).

PR #16019, opened 2020-11-17
geo restrict bypass for ReviewNotesViewSet

### LINKED ISSUE #16005: Review History in Manage versions page is broken when viewed from a region where the add-on is restricted
Follow up for #15942 

1. Sign in to AMO-dev with a developer account that has some add-ons submitted
2. With an admin account, restrict one of the developer's add-ons in a specific country
3. With the developer account, open the add-on Manage versions page in Dev-hub while being connected from the region in which the add-on has been restricted
4. Click on the version Review History

### What happened?
An error message with a broken layout is being displayed (see screenshot). The message displays the 451 error response returned by the `/api/v4/addons/addon/620244/versions/1692213/reviewnotes/` endpoint. 

We can either show the review notes to add-on owners and reviewers, or display a nicer error message.

- Reviewer notes are being displayed on the add-on review page, but I think that's because we do not rely on the version detail API there to retrieve those notes.
- This [addon](https://addons-dev.allizom.org/en-US/developers/addon/search-extensionggff/versions) is restricted in FR and the issue should be visible when logged in with an admin account

image

eviljeff:
fixes #16005

### DIFF
--- a/src/olympia/addons/views.py
+++ b/src/olympia/addons/views.py
@@ -22,13 +22,12 @@ import olympia.core.logger
 
 from olympia import amo
 from olympia.access import acl
-from olympia.addons.models import AddonRegionalRestrictions
 from olympia.amo.urlresolvers import get_outgoing_url
 from olympia.api.exceptions import UnavailableForLegalReasons
 from olympia.api.pagination import ESPageNumberPagination
 from olympia.api.permissions import (
     AllowAddonAuthor, AllowReadOnlyIfPublic, AllowRelatedObjectPermissions,
-    AllowReviewer, AllowReviewerUnlisted, AnyOf, GroupPermission)
+    AllowReviewer, AllowReviewerUnlisted, AnyOf, GeoAvailable, GroupPermission)
 from olympia.constants.categories import CATEGORIES_BY_ID
 from olympia.search.filters import (
     AddonAppQueryParam, AddonAppVersionQueryParam, AddonAuthorQueryParam,
@@ -149,6 +148,7 @@ def find_replacement_addon(request):
 
 class AddonViewSet(RetrieveModelMixin, GenericViewSet):
     permission_classes = [
+        GeoAvailable,
         AnyOf(AllowReadOnlyIfPublic, AllowAddonAuthor,
               AllowReviewer, AllowReviewerUnlisted),
     ]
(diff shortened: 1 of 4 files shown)


bobsilverberg (commented):
I don't really understand what this patch is doing. How is it that an exception is being made for `ReviewNotesViewSet` when the only change I see is for `AddonViewSet`?

eviljeff:
@bobsilverberg it moves the functionality into a permission, used in `permission_classes`, so other viewsets that use `AddonViewSet` indirectly and override `permission_classes` (e.g. `ReviewNotesViewSet`) won't get restricted by it.

src/olympia/addons/views.py
```diff
@@ -209,13 +209,11 @@ def check_object_permissions(self, request, obj):
...
         has been disabled by their developer or something else.
-
-        On top of this, can also raise a 451 if the add-on is not available
-        because of regional restrictions - no additional detail is available
-        in that case.
         """
         try:
             super(AddonViewSet, self).check_object_permissions(request, obj)
+        except UnavailableForLegalReasons as exec_451:
+            raise exec_451
```
eviljeff, line 216: This is because `UnavailableForLegalReasons` subclasses `exceptions.APIException` but we don't need the extra `is_disabled` data when it's a 451.

bobsilverberg:
> @bobsilverberg it moves the functionality into a permission, used in `permission_classes`, so other viewsets that use `AddonViewSet` indirectly and override `permission_classes` (e.g. `ReviewNotesViewSet`) won't get restricted by it.

Ok, I understand, but that makes me a bit nervous. How do we know that there aren't any views that are overriding `permission_classes` but that we'd still want to be protected against showing disallowed add-ons?

eviljeff:
That's just how the permission system is supposed to work - permission_classes can be overridden to do anything - already they could be changed to allow any type of user access to unlisted addons, for example.   

The current reality is we don't have any `AddonViewSet` subclasses, so overriding is limited to where `permission_classes` is specified in the `AddonViewSet` initialization - two places - one of those being `get_addon_object` where `permission_classes` is passed in as an optional arg, and again is specified in two places.  (so three occurances total; one is ReviewNotesViewSet; the others are RatingViewSet that allows all access to public addons, and AddonAbuseViewSet that is open to all addons)

PR #16021, opened 2020-11-17
add admin for AddonRegionalRestrictions

### LINKED ISSUE #15968: Add admin for georestrictions
https://github.com/mozilla/addons-server/issues/15942 will initially land without an admin, but we should have one. It should be implemented as a regular django admin, with an email notification going out for every addition/modification/deletion that includes the details of the changes because it's so sensitive.

eviljeff:
fixes #15968 - minimally, it doesn't email yet.

### DIFF
--- a/src/olympia/addons/tests/test_admin.py
+++ b/src/olympia/addons/tests/test_admin.py
@@ -10,7 +10,8 @@ from django.contrib.messages.storage import (
 from olympia import amo, core
 from olympia.activity.models import ActivityLog
 from olympia.addons.admin import AddonAdmin, ReplacementAddonAdmin
-from olympia.addons.models import Addon, ReplacementAddon
+from olympia.addons.models import (
+    Addon, AddonRegionalRestrictions, ReplacementAddon)
 from olympia.amo.tests import (
     TestCase, addon_factory, collection_factory, user_factory, version_factory)
 from olympia.amo.urlresolvers import django_reverse, reverse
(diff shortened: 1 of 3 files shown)


bobsilverberg (approved):
Looks good, nice work @eviljeff. How is access to this admin function defined? Can anyone who can access the admin make these changes, or do they need to belong to a particular group?

r+wc

eviljeff:
> Looks good, nice work @eviljeff. How is access to this admin function defined? Can anyone who can access the admin make these changes, or do they need to belong to a particular group?

The default for access to admin modules is only full admins (`*:*`) can access them.

src/olympia/addons/tests/test_admin.py
```diff
@@ -726,3 +727,86 @@ def test_can_list_with_admin_curation_permission(self):
...
+        addon = addon_factory()
+        self.add_url = reverse('admin:addons_addonregionalrestrictions_add')
+
+        response = self.client.get(self.add_url, follow=True)
+        assert response.status_code == 200
+        assert pq(response.content)('#id_addon')  # addon input is editable
+
+        response = self.client.post(
+            self.add_url, {
+                'excluded_regions': '["de", "pt-BR"]',
```
willdurand, line 769: Region code should be uppercase values only and I dont see that documented or enforced.

eviljeff: I couldn't see that documented anywhere either - can you log an issue?

eviljeff: I logged an issue https://github.com/mozilla/addons-server/issues/16023

PR #16026, opened 2020-11-18
Force AddonRegionalRestrictions.excluded_regions to be uppercase

### LINKED ISSUE #16023: Force AddonRegionalRestrictions.excluded_regions to be uppercase
follow-up from #15968 - due to how we match the results we need the country codes to be uppercase but we don't enforce (or inform about) that requirement.

eviljeff:
fixes #16023


src/olympia/addons/models.py
```diff
@@ -1744,6 +1744,14 @@ class AddonReviewerFlags(ModelBase):
         default=None)
 
 
+def uppercase_items_json_validator(value):
```
bobsilverberg, line 1747: Rather than validate, couldn't we just uppercase any values we receive?

bobsilverberg (approved):
The code looks fine, but why wouldn't we just convert any values we receive to uppercase before storing them, rather than validate and return an error?

r+wc

eviljeff:
mmm, we don't generally do that with the admin, but can do.  I'll close this pr because it's a completely different implementation

PR #16030, opened 2020-11-19
clean excluded_regions to uppercase the county codes

### LINKED ISSUE #16023: Force AddonRegionalRestrictions.excluded_regions to be uppercase

eviljeff:
fixes #16023 (in a different way)


bobsilverberg (approved):
Looks good, thanks for making this change!

PR #16031, opened 2020-11-19
allow regional restrictions to be combined with other permssions

### LINKED ISSUE #16014: Ignore georestrictions in api for certain authenticated users
follow-up for #15942 
> As a follow-up, we'll likely need something to be added to be able to bypass this if logged in as an admin, while adding something to the response to indicate to the frontend that we bypassed the restriction, but this doesn't have to be implemented right away.

eviljeff:
fixes #16014

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -425,6 +425,14 @@ class AddonAndVersionViewSetDetailMixin(object):
         data = response.json()
         assert data == {'detail': 'Unavailable for legal reasons.'}
 
+        # But admins can still access:
+        user = user_factory()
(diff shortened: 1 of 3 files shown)


bobsilverberg (approved):
Looks good, but just wondering, where did we decide that `Addons:Edit` is the permission that will allow this?

r+wc

eviljeff:
> Looks good, but just wondering, where did we decide that `Addons:Edit` is the permission that will allow this?

Addons:Edit is the pseudo-admin permission we used in a number of places where full admin is too restrictive.  I assumed that's what @jvillalobos wanted but he can contradict me (by replying to this 😄).

jvillalobos:
sgtm

PR #16059, opened 2020-11-27
Remove uwsgi-dogstatsd plugin

bqbn:
We no longer use datadog for monitoring and this plugin is no longer needed.
The PR essentially undoes https://github.com/mozilla/addons-server/pull/8143.

* [X] Add a description of the changes introduced in this PR.
* [X] The change has been successfully run locally.

### DIFF
--- a/Dockerfile.deploy
+++ b/Dockerfile.deploy
@@ -32,7 +32,6 @@ RUN touch /addons-server-docker-container \
         build-essential \
         curl \
         libjpeg-dev \
-        libpcre3-dev \
         libsasl2-dev \
         libxml2-dev \
         libxslt-dev \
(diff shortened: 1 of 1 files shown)


Dockerfile.deploy
```diff
@@ -41,7 +40,6 @@ RUN touch /addons-server-docker-container \
         zlib1g-dev \
         libffi-dev \
         libssl-dev \
-        libpcre3-dev \
```
eviljeff, line 44: this is listed twice fwiw - are you trying to drop it or removing the duplicate?

eviljeff (approved):
We also install the same deps (including `libpcre3-dev` twice) in Dockerfile (i.e. local docker).  I'm guessing they aren't needed / were never needed too?

Dockerfile.deploy
bqbn: I intend to drop it because https://github.com/mozilla/addons-server/pull/8143 added it and this PR is to undo that PR.

bqbn: Now it's dropped.

Issue #6548, opened 2017-09-29
Return (paginated) add-ons along with the detail of a collection


willdurand:
The API exposes a detail endpoint to fetch a single collection: https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#detail. Yet, it does not expose the add-ons part of this collection, only metadata about it. There is a second endpoint to fetch the add-ons for a given collection though: https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#collection-add-ons-list.

The problem is that to retrieve and display a collection, the frontend has to send two queries. Would it be possible to have a query parameter to include the add-ons along with the metadata?

Something like this or with `?with-addons`:

```
GET /api/v3/accounts/account/<user>/collections/<collection>/?includes=addons
```

It should return the same JSON as the [collection add-ons list](https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#collection-add-ons-list) in a new attribute (`items`?). I also think this endpoint should also support pagination.

Thanks!

eviljeff:
copy of the irc chat pre- the issue being filed:

> 17:41 @willdurand eviljeff: would it make sense to have a "?with-addons" or "?includes=addons" in the collection detail API? https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#detail. I have to send 2 requests to display a collection page right now. I did not see this pattern in the API though, so just asking
> 
> 18:19 @eviljeff Andrew Williamson willdurand: it's possible. Generally we don't try to combine detail responses with lists of stuff that could need pagination but if it's a performance bottleneck it's doable.
> 18:19 that said, I guess there are a number of pages that need multiple requests (add-on page, for example) so is it a problem? Hmm. mat, cgrebs, kumar: opinions?
> 
> 18:21 @kumar303  eviljeff: willdurand: there are a couple other pages where we need to load data with two API requests per page. We should probably wait until we finish our initial release then take a step back and look at all the pages that require multiple API requests.

eviljeff:
Is this still a post-release task, or do you need this adding before the MVP in November?

There are a few different ways we could add this, depending on whether it's purely single collection (detail view, not list); if only read-only is needed, or write too.   Whether this is going to be a pattern used for other API endpoints that have a similar hierachy (Addons, Accounts, etc)

willdurand:
I think it can be done post release. I have implemented the feature on the frontend calling the two endpoints.

AlexandraMoga:
[...]

PR #16064, opened 2020-11-30
Add some caching for anonymous requests to the collection add-ons API endpoint

### LINKED ISSUE #16062: Collections Add-ons list API endpoint is slow
The [collection add-ons list API endpoint](https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#collection-add-ons-list) is slow. This is not new, but now that it's used by Fenix (actually android-components) to fetch add-ons available for Fenix, and that the collection(s) used for that are growing larger and larger, this is a problem. It can be seen on this graph:

Screenshot_2020-11-30 AMO Prod frontend APIs usage performance - Grafana

It's been slower over time because add-ons kept being added, and around November 17th a more severe slowdown can be observed, maybe because [android-components started fetching more add-ons per request](https://github.com/mozilla-mobile/android-components/issues/8076) or just because the number of add-ons in the collection got even bigger ?

We want Fenix (and android-components) to stop using this endpoint altogether, replacing it with the search endpoint w/ the parameter to request recommended extensions only, but in the meantime, we likely need to find a solution to make the collection add-ons list endpoint faster.

diox:
Sort of fixes #16062

### DIFF
--- a/src/olympia/bandwagon/views.py
+++ b/src/olympia/bandwagon/views.py
@@ -1,12 +1,13 @@
 from django.conf import settings
+from django.utils.decorators import method_decorator
 
 from rest_framework import serializers
 from rest_framework.viewsets import ModelViewSet
 
-
 from olympia import amo
 from olympia.accounts.views import AccountViewSet
 from olympia.addons.models import Addon
+from olympia.amo.utils import cache_page_if_anonymous
 from olympia.api.filters import OrderingAliasFilter
 from olympia.api.permissions import (
     AllOf, AllowReadOnlyIfPublic, AnyOf, PreventActionPermission)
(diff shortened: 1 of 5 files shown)


src/olympia/bandwagon/views.py
```diff
@@ -110,6 +111,12 @@ class CollectionAddonViewSet(ModelViewSet):
                               'added': 'created'}
     ordering = ('-addon__weekly_downloads',)
 
+    # This endpoint can be quite slow so it's cached for one hour for all
+    # anonymous users.
+    @method_decorator(cache_page_if_anonymous(60 * 60 * 1))
+    def list(self, *args, **kwargs):
+        return super().list(*args, **kwargs)
```
eviljeff, line 118: CollectionViewSet.detail with `?with_addons` returns a very similar response to this if we want to cover all the bases.

eviljeff (approved):

src/olympia/bandwagon/views.py
diox: I started looking at that, mumbling about the risk of extending the cache to collection metadata, and then realized... `with_addons` doesn't seem to be used anywhere ? So I don't think this is worth doing.

eviljeff: https://github.com/mozilla/addons-server/issues/6548 it was requested for addons-frontend to use.  If it's not used we could arguably drop it... (though I think it's a quite useful feature really)

PR #16082, opened 2020-12-02
optimize CollectionAddonViewset with transforms and attach_trans_dict

### LINKED ISSUE #16081: optimize CollectionAddonViewSet queryset performance
follow on from #8831 (that added caching).

A single list request for CollectionAddonViewSet makes a large number of queries, and some of them are per addon, so it doesn't scale well.  We should change the queries so that at least all the results are retrieved in a single query - a collection with 100 addons should have the same number of queries as a collection with 1 addon inside.

eviljeff:
fixes #16081

### DIFF
--- a/src/olympia/bandwagon/views.py
+++ b/src/olympia/bandwagon/views.py
@@ -6,11 +6,12 @@ from rest_framework.viewsets import ModelViewSet
 
 from olympia import amo
 from olympia.accounts.views import AccountViewSet
-from olympia.addons.models import Addon
-from olympia.amo.utils import cache_page_if_anonymous
+from olympia.addons.models import Addon, attach_tags
+from olympia.amo.utils import attach_trans_dict, cache_page_if_anonymous
 from olympia.api.filters import OrderingAliasFilter
 from olympia.api.permissions import (
     AllOf, AllowReadOnlyIfPublic, AnyOf, PreventActionPermission)
+from olympia.versions.models import License, Version
 from olympia.translations.query import order_by_translation
 
 from .models import Collection, CollectionAddon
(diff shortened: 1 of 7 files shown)


bobsilverberg (approved):
I can't say that I really understand what's going on in some of these files, but nothing jumps out at me as a problem, and I see that Mat has also been asked to review.

src/olympia/bandwagon/views.py
```diff
@@ -138,11 +139,38 @@ def get_object(self):
...
+        attach_trans_dict(CollectionAddon, objs)
+        attach_trans_dict(Addon, addons)
+        attach_trans_dict(License, [ver.license for ver in current_versions])
+
     def get_queryset(self):
         qs = (
             CollectionAddon.objects
             .filter(collection=self.get_collection())
-            .prefetch_related('addon'))
+            .prefetch_related('addon__promotedaddon', 'addon__addon_tags')
```
diox, line 168: Why do you need `addon__addon_tags` here, shouldn't it be handled by your call to `attach_tags` in the `_transformer` ?

diox (commented):
Needs some tests to ensure the transformer are fetching the right data and setting it correctly

src/olympia/bandwagon/views.py
eviljeff: you're right, that's one more query down https://github.com/mozilla/addons-server/pull/16082/commits/aa96ec8673353d9bc86611d02f2c5679167cf3fe

eviljeff:
Added a test that checks the localisations were loaded correctly https://github.com/mozilla/addons-server/pull/16082/commits/aa96ec8673353d9bc86611d02f2c5679167cf3fe

diox (approved):
:1st_place_medal:

PR #16085, opened 2020-12-02
Add list of interesting accounts in local dev login page

### LINKED ISSUE #16084: Add list of interesting accounts in local dev login page
For people like me who have a bunch of different accounts locally with different sets of permissions, it's always a pain to remember which account has what and what email address I used.

If `settings.DEBUG` and `settings.USE_FAKE_FXA_AUTH` are `True` we have a completely fake login page anyway, so we should suggest interesting accounts there, for instance listing those that have any special permissions, making it easier to juggle through different accounts locally.

diox:
Fixes #16084

### DIFF
--- a/static/js/zamboni/users.js
+++ b/static/js/zamboni/users.js
@@ -3,11 +3,11 @@ $('#admin-login').click(function () {
   window.location = $(this).attr('data-url');
 });
 
-$('#fake_fxa_authorization .accounts a').click(function(e) {
+$('#fake_fxa_authorization .accounts a').click(function (e) {
   $('#id_email').prop('value', $(this).prop('href').replace('mailto:', ''));
   e.preventDefault();
   $('#fake_fxa_authorization').submit();
-})
+});
 
 // Recaptcha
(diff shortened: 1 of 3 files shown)


static/js/zamboni/users.js
```diff
@@ -3,6 +3,12 @@ $('#admin-login').click(function () {
   window.location = $(this).attr('data-url');
 });
 
+$('#fake_fxa_authorization .accounts a').click(function (e) {
+  $('#id_email').prop('value', $(this).prop('href').replace('mailto:', ''));
```
willdurand, line 7: I was wondering how a `mailto` would submit a form... Maybe it should not be `mailto` but a data attribute?

willdurand (approved):
r+wc

static/js/zamboni/users.js
diox: Maybe... it was already in the text and I prefer having actual links but maybe I don't need to...

willdurand: (it doesn't matter much and I dont have a better idea right now fwiw)

PR #16093, opened 2020-12-03
Update hint regarding allowed extensions for uploads to remove .xml

### LINKED ISSUE #16090: .xml should be removed from the linter message
As a followup for https://github.com/mozilla/addons-server/issues/13729

STR:
1. The message displayed below Select a file button when uploading an addon is:  Your add-on should end with .zip, .xpi, .crx or .xml

Expected result:
I think that .xml could be removed from the message since opensearch plugins are no longer supported for upload.

Notes:
This issue is reproducible on all AMO servers. 

linter message

diox:
Opensearch add-ons are gone.

Fixes #16090

### DIFF
--- a/static/js/common/upload-addon.js
+++ b/static/js/common/upload-addon.js
@@ -76,7 +76,7 @@
         }),
         ui_details = $('<div>', {
           class: 'upload-details',
-          text: gettext('Your add-on should end with .zip, .xpi, .crx or .xml'),
+          text: gettext('Your add-on should end with .zip, .xpi, or .crx'),
         });
 
       $upload_field.prop('disabled', false);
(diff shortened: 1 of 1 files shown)


static/js/common/upload-addon.js
```diff
@@ -76,7 +76,7 @@
         }),
         ui_details = $('<div>', {
           class: 'upload-details',
-          text: gettext('Your add-on should end with .zip, .xpi, .crx or .xml'),
+          text: gettext('Your add-on should end with .zip, .xpi, or .crx'),
```
bobsilverberg, line 79: Nit: previous version did not use an Oxford comma. I'm guessing this is a typo and should be removed here as well.

bobsilverberg (approved):
Looks good, just a grammar-nerd nit.

r+wc

static/js/common/upload-addon.js
diox: Good catch, fixed in 0ec5a367758bd1b2d8d6280786e297f0e9964b04

PR #16105, opened 2020-12-04
Optimize review page slightly by grouping and removing some queries

### LINKED ISSUE #16096: Optimize calls to version.is_ready_for_auto_approval property and addon.find_latest_version() in reviewer tools
There are 2 obvious sets of duplicate queries in reviewer tools at the moment:

- `<Version>.is_ready_for_auto_approval` is a property we only use in a couple places (in reviewer tools an devhub) to show/avoid showing some extra info to reviewers / developers about a version depending on whether it's auto approvable or not.
It's implemented in a DRY way, it calls `Version.objects.auto_approve()` under the hood, restricting with the `pk` of the version we're looking at. Unfortunately, this means it's causing a database query for every version it's called. This matters in reviewer tools where we display 10 versions on each page.
We should optimize this, maybe making this a `@cached_property` that we could fill for all instances in a queryset in one go, or by finding a way to still share the code with the manager method while using the `Addon` and `File` instances that are already loaded.
- `files_view.html`, which is rendered for each version through `all_distinct_files()` template tag, does `{% set latest_version_in_channel=addon.find_latest_version(channel=version.channel) %}` repeating the two queries that method executes for every version. It should be set before in context, but we should be careful since it's a template tag : nothing else is using it at the moment, but we may want to explicitly pass the `latest_version` to the tag so that it doesn't fail suddenly if we change the context or use the tag in a different manner in the future.

Doing ~ 30 fewer queries per page seems like a nice perf win...

diox:
Around 39 fewer queries for an add-on with 10 versions being displayed.

Fixes #16096

### DIFF
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -3340,7 +3340,76 @@ class TestReview(ReviewBase):
             # Add versions 1.0 to 1.9
             version_factory(
                 addon=addon, version=f'1.{i}', created=self.days_ago(365 - i))
-        response = self.client.get(self.url)
+        with self.assertNumQueries(60):
+            # FIXME: 60 is obviously still too high, but it's a starting point.
+            # Potential further optimizations:
+            # - Remove trivial... and not so trivial duplicates
+            # - Group similar queries
+            # - Try to do counts of things on different page in a single query
+            # - Remove useless things like user add-ons and collection
+            # - Make some joins
+            #
+            # 1. user
+            # 2. savepoint
+            # 3. groups
+            # 4. add-on by slug
+            # 5. add-on translations
+            # 6. categories
+            # 7. current version
+            # 8. version translations
+            # 9. applications versions
+            # 10. files
+            # 11. authors
+            # 12. previews
+            # 13. autoapprovalsummary for current version
+            # 14. promoted info for the add-on
+            # 15. latest version
+            # 16. latest version translations
+            # 17. latest version (repeated)
+            # 18. latest version translations (repeated)
+            # 19. addon reviewer flags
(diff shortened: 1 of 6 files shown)


src/olympia/versions/models.py
```diff
@@ -675,6 +675,22 @@ def transformer_license(cls, versions):
...
+        ids = set(v.id for v in versions)
+        if not ids:
+            return
+
+        auto_approvable = dict(
+            Version.objects.auto_approvable()
+                   .annotate(
+                       is_ready_for_auto_approval=Value(True, BooleanField()))
+                   .values_list('pk', 'is_ready_for_auto_approval')
+        )
```
eviljeff, line 689: So, this is just a dict of `pk: True`?  Why not go with the more simple solution of getting a list of pks and setting true if the version pk is in that list?

src/olympia/reviewers/tests/test_views.py
```diff
@@ -3340,7 +3340,76 @@ def test_item_history_pagination(self):
             # Add versions 1.0 to 1.9
             version_factory(
                 addon=addon, version=f'1.{i}', created=self.days_ago(365 - i))
-        response = self.client.get(self.url)
+        with self.assertNumQueries(60):
```
eviljeff, line 3343: 🐼

eviljeff (approved):
it's something!

src/olympia/versions/models.py
diox: Yeah, also, I should filter by `pk__in=ids` ... I'll change this.

diox: https://github.com/mozilla/addons-server/pull/16105/commits/5263dc2c5789318325bb4b5fce314a09bc21a3bd

PR #16113, opened 2020-12-07
Remove icon selector from devhub +(js/css fixes for 128 size default)

### LINKED ISSUE #10090: The 2000s wants its graphics back - drop/update the selection of default addon icons
We current allow developers to either upload a custom addon icon or select from 18 icons.  Which are, imho, quite dated - unsurprisingly as they've not been updated since they were added over 8 years [ago](https://github.com/mozilla/addons-server/commit/b8665d4d640c10fd83f9c6a9d5704347771c1a9e#diff-e8edcb422a92f23e11fd0aab389e2ec4) as category icons.

The trigger for this is #5805 makes 128x128 the new standard but we only have 32 and 64 size for them, so as a workaround I has to upscale the 64 to 128 which makes them look even worse.

~So we should either drop them, or if someone wants to replace them with a more current selection I guess that works too.~  We're going to drop the multiple icon selection and replace them with the single default icon as per https://github.com/mozilla/addons-server/issues/10090#issuecomment-533190650 (the developer can still upload their own)

@jvillalobos

eviljeff:
Fixes #10090 - the first step anyway, removing the selection.  Follow-ups are to migrate existing add-ons using the icons (copying the chosen icon to that particular add-on) (#16114); then deleting the supporting code for the `icon/xxx` `image_type` property.

### DIFF
--- a/src/olympia/devhub/forms.py
+++ b/src/olympia/devhub/forms.py
@@ -7,13 +7,11 @@ from urllib.parse import urlsplit
 
 from django import forms
 from django.conf import settings
-from django.core.files.storage import default_storage as storage
 from django.core.validators import MinLengthValidator
 from django.db.models import Q
 from django.forms.formsets import BaseFormSet, formset_factory
 from django.forms.models import BaseModelFormSet, modelformset_factory
 from django.forms.widgets import RadioSelect
-from django.utils.encoding import force_text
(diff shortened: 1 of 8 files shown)


bobsilverberg (approved):
Looks good, thanks @eviljeff

PR #16117, opened 2020-12-07
Return an HttpResponse and not a DRF's Response before caching collections API

### LINKED ISSUE #16102: Investigate cache issues in production
We deployed 2020.12.03 yesterday as normal, expecting the [cache added for the Collections Add-ons API endpoint](https://github.com/mozilla/addons-server/issues/16062) to help with the [scaling issues](https://github.com/mozilla/addons-server/issues/15995) we noticed recently.

Unfortunately, that did not work as expected. We can see on this graph the endpoint remains slow:
Screenshot_2020-12-04 AMO Prod frontend APIs usage performance - Grafana

When testing manually via `curl`, we can see we're being served a cache response when the `Expires` points to a date that's less than one hour in the future. Here is what it can look like on stage, where it does work as expected:

```
% curl -s https://addons.allizom.org/api/v4/accounts/account/4760859/collections/f2b87cb7-9a00-1c0d-64d0-0911e7/addons/ -I
# Wait 2 minutes...
% curl -s https://addons.allizom.org/api/v4/accounts/account/4760859/collections/f2b87cb7-9a00-1c0d-64d0-0911e7/addons/ -I | egrep 'Date|Expires'
Date: Fri, 04 Dec 2020 10:46:46 GMT
Expires: Fri, 04 Dec 2020 11:44:21 GMT
```

We did see that happening in prod, but not often as we'd like. Most of the time we appear to be served fresh content. We asked Wei for some sample URLs from nginx and they all looked normal.
```
"GET /api/v4/accounts/account/mozilla/collections/7dfae8669acc4312a65e8ba5553036/addons/?page_size=50&sort=-popularity HTTP/1.1" 200 210472 "-" "Mozilla/5.0 (Android 10; Mobile; rv:83.0) Gecko/83.0 Firefox/83.0"
"GET /api/v4/accounts/account/mozilla/collections/7dfae8669acc4312a65e8ba5553036/addons/?page_size=50&sort=-popularity HTTP/1.1" 200 210339 "-" "Mozilla/5.0 (Android 9; Mobile; rv:83.0) Gecko/83.0 Firefox/83.0"
"GET /api/v4/accounts/account/mozilla/collections/7dfae8669acc4312a65e8ba5553036/addons/?page_size=50&sort=-popularity HTTP/1.1" 200 210339 "-" "Mozilla/5.0 (Android 8.1.0; Mobile; rv:83.0) Gecko/83.0 Firefox/83.0"
"GET /api/v4/accounts/account/mozilla/collections/7dfae8669acc4312a65e8ba5553036/addons/?page_size=50&sort=-popularity HTTP/1.1" 200 210447 "-" "Mozilla/5.0 (Android 10; Mobile; rv:83.0) Gecko/83.0 Firefox/83.0"
"GET /api/v4/accounts/account/mozilla/collections/7dfae8669acc4312a65e8ba5553036/addons/?page_size=50&sort=-popularity HTTP/1.1" 200 210447 "-" "Mozilla/5.0 (Android 10; Mobile; rv:83.0) Gecko/83.0 Firefox/83.0"
"GET /api/v4/accounts/account/mozilla/collections/7dfae8669acc4312a65e8ba5553036/addons/?page_size=50&sort=-popularity HTTP/1.1" 200 210315 "-" "Mozilla/5.0 (Android 10; Mobile; rv:83.0) Gecko/83.0 Firefox/83.0"
"GET /api/v4/accounts/account/mozilla/collections/7dfae8669acc4312a65e8ba5553036/addons/?page_size=50&sort=-popularity HTTP/1.1" 200 210315 "-" "Mozilla/5.0 (Android 10; Mobile; rv:83.0) Gecko/83.0 Firefox/83.0"
"GET /api/v4/accounts/account/mozilla/collections/7dfae8669acc4312a65e8ba5553036/addons/?page_size=50&sort=-popularity HTTP/1.1" 200 210321 "-" "Mozilla/5.0 (Android 5.0.1; Mobile; rv:83.0) Gecko/83.0 Firefox/83.0"
```

There are some minor variations in the `User-Agent`, which would cause the cache key to be different (we `Vary` on it, which we shouldn't do but that's a topic for another day), but not that many.

Because local envs, dev & stage appear to be working normally, we suspect there is something wrong with prod.

diox:
This makes it more likely the response will actually be cached, otherwise when it's pickled the response might be over memcached's size limit

Fixes #16102

### DIFF
--- a/src/olympia/bandwagon/views.py
+++ b/src/olympia/bandwagon/views.py
@@ -1,3 +1,4 @@
+from django.http import HttpResponse
 from django.conf import settings
 from django.utils.decorators import method_decorator
 
@@ -115,8 +116,21 @@ class CollectionAddonViewSet(ModelViewSet):
     # This endpoint can be quite slow so it's cached for one hour for all
     # anonymous users.
     @method_decorator(cache_page_if_anonymous(60 * 60 * 1))
-    def list(self, *args, **kwargs):
-        return super().list(*args, **kwargs)
+    def list(self, request, *args, **kwargs):
+        # Swap DRF's Response with a HttpResponse so that it's smaller in size
+        # when pickling (data attribute is lost, we only keep the rendered
+        # content), which matters because by default, memcached doesn't
+        # accept values over 1 MB.
+        # Note that this needs to happen before cache_page_if_anonymous() has
+        # seen the response, because it attaches the callback that caches the
+        # response to it. That's why we manually call finalize_response() and
+        # render() here.
+        response = super().list(*args, *kwargs)
+        response = self.finalize_response(request, response, *args, **kwargs)
+        response.render()
+        return HttpResponse(
+            response.content, status=response.status_code,
+            content_type='application/json')
 
     def get_collection(self):
         if not hasattr(self, 'collection'):
--- a/src/olympia/bandwagon/tests/test_views.py
+++ b/src/olympia/bandwagon/tests/test_views.py
@@ -4,6 +4,7 @@ import json
 import django.test
 
 from django.conf import settings
+from django.http import HttpResponse
 from django.test.utils import override_settings
 
 from rest_framework.fields import empty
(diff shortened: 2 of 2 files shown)


diox:
In case anyone is interested, the approaches I've discarded while working on this:
- Swap the response in `finalize_response()`. Doesn't work, because `list()` is where the decorator is applied, so when we would finalize the response to it to be swapped, we'd run the post render callback the decorator has attached, and attempt to cache the original DRF response instead of the swapped one
- Swap the response in the decorator. Would work, but requires copy/pasting a bunch of the code from django's `UpdateCacheMiddleware` (which is what `cache_page` is using behind the scenes) because there is no entry point for that.
- Don't swap, return an `HttpResponse` or `JsonResponse` directly in `list()`. Pagination makes it annoying. The `Response` we return is actually buried deeper in the paginator, so we'd need to override a bunch of code.
- Drop caching entirely, relying on optimizations done in https://github.com/mozilla/addons-server/issues/16081. I figured caching would still be worth it even with the optimizations, given how much traffic we're seeing to that endpoint because of Fenix.

src/olympia/bandwagon/views.py
```diff
@@ -115,8 +116,21 @@ class CollectionAddonViewSet(ModelViewSet):
...
+    def list(self, request, *args, **kwargs):
+        # Swap DRF's Response with a HttpResponse so that it's smaller in size
+        # when pickling (data attribute is lost, we only keep the rendered
+        # content), which matters because by default, memcached doesn't
+        # accept values over 1 MB.
+        # Note that this needs to happen before cache_page_if_anonymous() has
+        # seen the response, because it attaches the callback that caches the
+        # response to it. That's why we manually call finalize_response() and
+        # render() here.
+        response = super().list(*args, *kwargs)
```
eviljeff, line 128: it's a shame we can't do this in `cache_page_if_anonymous` so it could be reused for other endpoints without having to copy this chunk of code into each one.

eviljeff (approved):

src/olympia/bandwagon/views.py
diox: Yeah. Alternatively we could write our own, but this one is fully battle-tested so... We might end up having a base mixin for all API classes to handle this, your optimizations to the translations, etc.

Issue #9885, opened 2018-11-05
move binary operators to start of the line in our python code


eviljeff:
Since Flake8 3.6.0 started enforcing W504 (#9795) binary operators can't be on the end of the line so we need to reformat to be compliant.  W504 is currently on the ignore list because there's 500+ places that fail otherwise.

EnTeQuAk:
:see_no_evil: #8617 :see_no_evil:

eviljeff:
when black stops emitting broken line lengths I'm for it 😛

willdurand:
See also: https://www.flake8rules.com/rules/W503.html

PR #16123, opened 2020-12-08
"Black" all the things!

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/8617
Fixes https://github.com/mozilla/addons-server/issues/9885

---

This is going to be fun!


willdurand:
I excluded the migrations and re-applied black so that it does not conflict. This is ready to go 😈

willdurand:
mm I am confused. It says `toml` is not found but it is installed via `requirements/travis_base.txt`

pyproject.toml
```diff
@@ -0,0 +1,21 @@
...
+   | media
+   | storage
+   | logs
+   | site-static
+   | static
+   | \.git
+   | \.npm
+   | \.tox
+  )/
+  | src/olympia/wsgi\.py
```
eviljeff, line 17: 😕 mmm?

src/olympia/access/acl.py
```diff
@@ -63,41 +64,50 @@ def langpack_submission_allowed(user, parsed_addon_data):
     See https://github.com/mozilla/addons-server/issues/11788 and
     https://github.com/mozilla/addons-server/issues/11793
     """
-    return (
-        not parsed_addon_data.get('type') == amo.ADDON_LPAPP or
-        action_allowed_user(user, amo.permissions.LANGPACK_SUBMIT))
+    return not parsed_addon_data.get("type") == amo.ADDON_LPAPP or action_allowed_user(
+        user, amo.permissions.LANGPACK_SUBMIT
+    )
```
eviljeff, line 69: this doesn't seem like an improvement, but I'm not going to fight Black on it.

Makefile-docker
```diff
@@ -192,6 +192,7 @@ lint: ## lint the code
 	flake8 src/ services/ tests/
 	$(shell npm $(NPM_ARGS) bin)/prettier --check '**'
 	curlylint src/
+	black src/ services/ tests/
```
eviljeff, line 195: this needs to be a command that would return the problems with the code, rather than auto-fix them.  I'd also be happier if it was at the top, before flake8.

Makefile-docker
```diff
@@ -255,6 +256,7 @@ watch_js_tests: ## Run+watch the JavaScript test suite (requires compiled/compre
 .PHONY: format
 format: ## Autoformat our codebase.
 	$(shell npm $(NPM_ARGS) bin)/prettier --write '**'
+	black src/
```
eviljeff, line 259: should we format `/services` too?

eviljeff (commented):
I _do_ wonder how much of this is `'` to `"` (so how much smaller if would be if string normalization was skipped) 🤷‍♂️

pyproject.toml
willdurand: this was in Chris' patch. I meant to check why but I forgot obviously.

Makefile-docker
willdurand: ah oops, right. there was `--check` before

willdurand: ah oops, right. there was `--check` before

Makefile-docker
willdurand: yep, sorry about that

willdurand:
Green!

willdurand:
> I _do_ wonder how much of this is `'` to `"` (so how much smaller if would be if string normalization was skipped) 🤷‍♂️

This would be smaller but we don't have a way to enforce the use of single quotes everywhere, which is why this patch uses black defaults.

willdurand:
FTR with double quotes:

Screenshot 2020-12-09 at 16 07 27

src/olympia/scanners/admin.py
```diff
@@ -48,41 +51,38 @@ def _is_safe_url(url, request):
...
         for lookup, title in self.lookup_choices:
             yield {
-                'selected': self.value() == lookup,
-                'query_string': cl.get_query_string(
-                    {self.parameter_name: lookup}, []
-                ),
-                'display': title,
+                "selected": self.value() == lookup,
+                "query_string": cl.get_query_string({self.parameter_name: lookup}, []),
+                "display": title,
```
eviljeff, line 63: the "switch back to single quotes " commit should have reverted these changes?  (I just picked a few files to check - I don't know how widespread it is)

willdurand: Yes, I am working on fixing that.

willdurand: I am going to re-apply black entirely to fix conflicts so this shuld be fixed.

willdurand:
So I changed the black config as we discussed on Element: black skips string normalization (for now). I'll file an issue about that later. I re-applied black on our codebase. I added another commit to fix errors reported by flake8. I re-ran `black --check` + `flake8` and everything was ok.

I'll take care of the existing double quotes in a follow-up patch.

diox (approved):
:black_heart:

services/utils.py
```diff
@@ -55,15 +55,18 @@ def user_media_url(what):
...
+    r"""(?P<major>\d+)         # major (x in x.y)
                             \.(?P<minor1>\d+)      # minor1 (y in x.y)
                             \.?(?P<minor2>\d+|\*)? # minor2 (z in x.y.z)
                             \.?(?P<minor3>\d+|\*)? # minor3 (w in x.y.z.w)
                             (?P<alpha>[a|b]?)      # alpha/beta
                             (?P<alpha_ver>\d*)     # alpha/beta version
                             (?P<pre>pre)?          # pre release
                             (?P<pre_ver>\d)?       # pre release version
-                          """, re.VERBOSE)
+                          """,
```
eviljeff, line 67: nit: the formatting for this is weird now (not that it matters after it's compiled)

src/olympia/abuse/models.py
```diff
@@ -50,8 +50,11 @@ class AbuseReport(ModelBase):
     REASONS = APIChoicesWithNone(
         ('DAMAGE', 1, 'Damages computer and/or data'),
         ('SPAM', 2, 'Creates spam or advertising'),
-        ('SETTINGS', 3, 'Changes search / homepage / new tab page '
-                        'without informing user'),
+        (
+            'SETTINGS',
+            3,
+            'Changes search / homepage / new tab page ' 'without informing user',
```
eviljeff, line 56: nit: weird string concatenation

eviljeff (approved):
From a spot check, found a few places where `'abc' 'def'` could be collapsed into a single string.  I wouldn't bother fixing and making a new commit+ci+etc, just noting it down while I saw it.
r+ 🚢

willdurand:
I'll fix the issues you've found in a follow-up, thanks

willdurand:
> From a spot check, found a few places where `'abc' 'def'` could be collapsed into a single string. I wouldn't bother fixing and making a new commit+ci+etc, just noting it down while I saw it.

I filed https://github.com/mozilla/addons-server/issues/16139

PR #16125, opened 2020-12-08
always returned urls wrapped in outgoing + originals

### LINKED ISSUE #11206: Expose the original URLs when the addons API returns outgoing links
The URLs received in [the addons API](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#detail) response use the outgoing proxy. In order to fix https://github.com/mozilla/addons-frontend/issues/3580, we need to know the original URLs (`homepage` and `support_url` fields, at least).

Can we have that?

### LINKED ISSUE #16268: correctly handle outgoing'ing translated fields in api/v5
looks like if we specify a lang, e.g. `/api/v5/addons/addon/tasty-sushi/?lang=en-GB` then the response ends up as:
```json
"homepage": {
    "url": {
      "en-US": "https://goog.gl",
      "en-GB": null,
      "_default": "en-US"
    },
    "outgoing": {
      "en-US": "https://outgoing.prod.mozaws.net/v1/19edd172a6b04a284719f9ba8bded9e04f10bc5fa13a91c019997c0c72d996/https%3A//goog.gl",
      "en-GB": null,
      "_default": "/"
    }
  },
```
Because we try to outgoingify the `_default` key too.  We shouldn't try to pass `_default` through outgoing because it's not a url.

eviljeff:
fixes #11206 - maybe `url` should be the outgoing url and there should be a `raw` (or a better name)?  That'd be a little more breaking, but we're already changing the data structure so maybe it doesn't matter.

Before merging this we'd need to have patches lined up for addons-code-manager (and addons-frontend if it switches to v5 in the meantime).   Adding the gate to `v5` is possible as a short-term workaround but it'd need more test changes and probably a `v6` adding (some of the tests need a real url that wouldn't have the gate enabled)

fixes #16268


bobsilverberg:
I've got a huge patch ready which implements support for `v5` of the API for addons-frontend, but not with this change. I'll need to either add these changes to that patch as well, or write another patch, and then block this on that patch. Do you think there's any value in making this a `v6` to simplify this situation a bit?

willdurand:
> I've got a huge patch ready which implements support for `v5` of the API for addons-frontend, but not with this change. I'll need to either add these changes to that patch as well, or write another patch, and then block this on that patch. Do you think there's any value in making this a `v6` to simplify this situation a bit?

We've done that in the past without necessarily introducing a new API version. I think it's fine to add support for current v5 in addons-frontend (that patch is huge, as you said). Then we'll see what needs to be done in addons-frontend and code-manager before landing this server patch. I don't think we need a `v6` right now for the reasons @eviljeff mentioned in the PR description.

eviljeff:
> I've got a huge patch ready which implements support for `v5` of the API for addons-frontend, but not with this change. I'll need to either add these changes to that patch as well, or write another patch, and then block this on that patch. 

I certainly wouldn't want work in `addons-frontend` to be blocked on this.  I was saying merging _this_ patch would be blocked on changes that would be needed in `addons-frontend`. (And `addons-code-manager` but I wouldn't be surprised if it didn't use any of the affected apis)

> Do you think there's any value in making this a `v6` to simplify this situation a bit?

There's value, but at a cost of adjusting tests and documenting it.  I don't think it's worth it.

bobsilverberg:
> > I've got a huge patch ready which implements support for `v5` of the API for addons-frontend, but not with this change. I'll need to either add these changes to that patch as well, or write another patch, and then block this on that patch.
> 
> I certainly wouldn't want work in `addons-frontend` to be blocked on this. I was saying merging _this_ patch would be blocked on changes that would be needed in `addons-frontend`. (And `addons-code-manager` but I wouldn't be surprised if it didn't use any of the affected apis)
> 
> > Do you think there's any value in making this a `v6` to simplify this situation a bit?
> 
> There's value, but at a cost of adjusting tests and documenting it. I don't think it's worth it.

Fair enough.

willdurand:
> fixes #11206 - maybe `url` should be the outgoing url and there should be a `raw` (or a better name)? That'd be a little more breaking, but we're already changing the data structure so maybe it doesn't matter.

Or we'd have `original_url` and `outgoing_url`? I guess `url` and `outgoing` is good enough, though.

willdurand (commented):
That looks good to me. If we are okay with the naming then it's a r+ otherwise let's discuss.

bobsilverberg:
> > fixes #11206 - maybe `url` should be the outgoing url and there should be a `raw` (or a better name)? That'd be a little more breaking, but we're already changing the data structure so maybe it doesn't matter.
> 
> Or we'd have `original_url` and `outgoing_url`? I guess `url` and `outgoing` 

> That looks good to me. If we are okay with the naming then it's a r+ otherwise let's discuss.

I’m fine with either.

eviljeff:
Found a bug while testing the frontend patch.  (which i logged as #16268 before remembering I hadn't merged this yet)

willdurand:
>  ** continuous-integration/travis-ci/push ** — The Travis CI build failed

Why is this running travis-ci? Could you please rebase your patch to use circle? Thanks!

eviljeff:
> Why is this running travis-ci? 

To mention the obvious: the branch is 28 days old - we didn't have circle-ci test CI back then 😛 

> Could you please rebase your patch to use circle? Thanks!

It creates *a lot* of merge conflicts so it seems like a lot of pain for no gain.  Having the tests run on travis-ci for this patch (while it's still working) seems okay to me.

willdurand:
> It creates _a lot_ of merge conflicts so it seems like a lot of pain for no gain. Having the tests run on travis-ci for this patch (while it's still working) seems okay to me.

The branch rebased is available here in case you want to use it: https://github.com/mozilla/addons-server/compare/11206-always-wrap-outgoing-links-2?expand=1

What I do when a simple rebase doesn't work is the following:

1. make sure `master` is up to date: `git checkout master; git pull origin master`
2. rename `branch` to `branch-2`: `git checkout branch; git checkout -b branch-2`
3. delete `branch`: `git branch -D branch`
4. create branch `branch` based on `master`: `git checkout master; git checkout -b branch`
5. merge with squash: `git merge --squash branch-2`
6. commit and force push `branch` to update the PR

willdurand:
I didn't approve this patch fwiw...

eviljeff:
> That looks good to me. If we are okay with the naming then it's a r+ otherwise let's discuss.

sorry, I misunderstood the status.  You said it was an r+ previously and then commented after the r? without mentioning any problems.

Issue #16132, opened 2020-12-09
Add transformer to prefill has_listed_versions / has_unlisted_versions for a queryset


diox:
In a few places we call `has_listed_versions` / `has_unlisted_versions` for each add-on in a list. This is super inefficient because it will generate 1 or 2 queries (that just do a `.exists()`) for each add-on (for listed we have a shortcut that'll return `True` when there is a `current_version`, but we don't have that possibility for unlisted)

We should have a transformer to do that on a list, supporting the optional `include_deleted` argument as well (which the admin/reviewer tools need). It would prefill some kind of cache the method could then use.

diox:
This is less useful than I'd like - in reviewer tools, this doesn't help the Scanners & MAD queues, because they used a different query, filtering also on the presence of an additional field (`needs_human_review` or `reviewerflags__needs_human_review_by_mad`) and displaying a count. On the review page, we're only dealing with a single add-on so it doesn't matter.

The one place that would benefit though, is the add-on changelist admin. Not as important, but it's used in popup mode to select an add-on so it does matter a bit.

diox:
QA: addon list/detail in admin pages and reviewer tools queues (particularly flagged by scanners/need human review) should still work correctly after this. It's unlikely you'd notice the perf difference on dev/stage.

ioanarusiczki:
@diox
I verified on -dev - FF84(Win10) and it looks good to me.
- the django admin list has an extra column "Authors" but it's probably expected.
- rev tools pages look alright, Flagged for Human Review takes some time to load on both dev and stage (-dev is loading a bit faster but still takes a while).

PR #16133, opened 2020-12-09
Expose authors (and their role, included deleted) in add-on admin list

diox:
In user admin:
Screenshot_2020-12-09 Change user profile Django site admin
In add-on admin changelist (regardless of whether there is any filtering applied):
Screenshot_2020-12-09 Select addon to change Django site admin


src/olympia/addons/admin.py
```diff
@@ -194,10 +198,38 @@ def wrapper(*args, **kwargs):
...
+                    (
+                        urljoin(
+                            settings.EXTERNAL_SITE_URL,
+                            reverse(
+                                'admin:users_userprofile_change',
+                                args=(author.pk,)
+                            ),
+                        ),
+                        author.email,
+                        dict(amo.AUTHOR_CHOICES_UNFILTERED)[author.role],
```
eviljeff, line 220: `author.get_role_display()`?

eviljeff: Oh, you can't do that because it's not an `AddonUser` instance, hmm.

src/olympia/users/admin.py
```diff
@@ -287,18 +287,28 @@ def has_active_api_key(self, obj):
...
-    addons_created.short_description = _('Addons')
+    collections_authorship.short_description = _('Collections')
+
+    def addons_authorship(self, obj):
+        non_deleted_authorship_count = Addon.unfiltered.filter(
+            addonuser__user=obj).exclude(
+            addonuser__role=amo.AUTHOR_ROLE_DELETED).count()
+        deleted_authorship_count = Addon.unfiltered.filter(
+            addonuser__user=obj,
+            addonuser__role=amo.AUTHOR_ROLE_DELETED).count()
```
eviljeff, line 300: I don't suppose there's a way to do this in a single query?  Something like `.values_list('addonuser__role', flat=True)` and count the roles in the response?

src/olympia/addons/admin.py
```diff
@@ -177,8 +180,9 @@ class Media:
         }))
     actions = ['git_extract_action']
 
-    def queryset(self, request):
-        return models.Addon.unfiltered
+    def get_queryset(self, request):
+        return Addon.unfiltered.all().only_translations().transform(
+            Addon.attach_all_authors)
```
eviljeff, line 185: This is going to change the queryset for the `change_view` too and we don't need to attach_all_authors there (... right??) 

I'm assuming it's not a terrible thing anyway, including bypassing the usual Addon transforms, as the tests pass and the query count seems to have gone down for the `change_view`.

src/olympia/addons/tests/test_admin.py
```diff
@@ -497,20 +497,24 @@ def test_query_count(self):
         self.grant_permission(user, 'Addons:Edit')
         self.grant_permission(user, 'Admin:Advanced')
         self.client.login(email=user.email)
-        with self.assertNumQueries(26):
+        with self.assertNumQueries(20):
```
eviljeff, line 500: Can we have a test that confirms `assertNumQueries` for the list view?

eviljeff (approved):

src/olympia/addons/admin.py
diox: Yeah...

src/olympia/users/admin.py
diox: I'd like to avoid doing this in case there are a lot of results. But there might be a clean way with some aggregation trickery, I'm not sure (it might not work in MySQL). I'll check.

src/olympia/addons/tests/test_admin.py
diox: That's my FIXME below - it will still be terrible while https://github.com/mozilla/addons-server/issues/16132 isn't fixed (next on my todo list) but I'll add that yeah.

src/olympia/users/admin.py
eviljeff: `.aggregate(Count('role'))` maybe?

src/olympia/addons/admin.py
diox: Yeah. Ideally we'd either be able to change the queryset for a given view or the inline for the authors would be able to use `all_authors` (or we'd be able to use `prefetch_related()`). But I don't think you can do either of these things in a clean/satisfying/safe way, so I figured I would live it like that for now.

src/olympia/users/admin.py
diox: ae19826171ff10f1aed282c53f351b75e58b26d4

src/olympia/addons/tests/test_admin.py
diox: e99376e1578c0b9a1f2eb6f8aaf1d059aa5ca05b

Issue #16139, opened 2020-12-10
Fix weird string concatenation


willdurand:
Follow-up: https://github.com/mozilla/addons-server/issues/8617

---

@eviljeff found some weird string concatenation occurrences like:

```python
foo = 'abc' 'def'
```

instead of:

```python
foo = 'abcdef'
```

Let's fix the ones we can find!

PR #16140, opened 2020-12-10
Fix weird string concatenation

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/16139

### DIFF
--- a/src/olympia/amo/monitors.py
+++ b/src/olympia/amo/monitors.py
@@ -48,7 +48,7 @@ def memcache():
 
             memcache_results.append((ip, port, result))
         if not using_twemproxy and len(memcache_results) < 2:
-            status = ('2+ memcache servers are required.' '%s available') % len(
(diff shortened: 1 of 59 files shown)


src/olympia/amo/monitors.py
```diff
@@ -48,7 +48,7 @@ def memcache():
 
             memcache_results.append((ip, port, result))
         if not using_twemproxy and len(memcache_results) < 2:
-            status = ('2+ memcache servers are required.' '%s available') % len(
+            status = ('2+ memcache servers are required. %s available') % len(
```
willdurand, line 51: This changes the string slightly but I think that's what we want?

eviljeff (approved):
I saw one with `"` strings too - promoted/models.py

willdurand:
> I saw one with `"` strings too - promoted/models.py

I will fix them in https://github.com/mozilla/addons-server/issues/16142

Issue #16144, opened 2020-12-10
Can't install or update local environement when there is a pyproject.toml file


diox:
When there is a `pyproject.toml` file in the repos we can't install or update addons-server locally. `make update_docker` eventually runs `python3 -m pip install -e .` and it fails with:

```
btaining file:///code
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Installing collected packages: olympia
  Running setup.py develop for olympia
    ERROR: Command errored out with exit status 1:
     command: /usr/local/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/code/setup.py'"'"'; __file__='"'"'/code/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps --user --prefix=
         cwd: /code/
    Complete output (28 lines):
    running develop
    WARNING: The user site-packages directory is disabled.
    error: can't create or remove files in install directory
    
    The following error occurred while trying to add or remove files in the
    installation directory:
    
        [Errno 13] Permission denied: '/usr/local/lib/python3.8/site-packages/test-easy-install-376.write-test'
    
    The installation directory you specified (via --install-dir, --prefix, or
    the distutils default setting) was:
    
        /usr/local/lib/python3.8/site-packages/
    
    Perhaps your account does not have write access to this directory?  If the
    installation directory is a system-owned directory, you may need to sign in
    as the administrator or "root" account.  If you do not have administrative
    access to this machine, you may wish to choose a different installation
    directory, preferably one that is listed in your PYTHONPATH environment
    variable.
    
    For information on other options, you may wish to consult the
    documentation at:
    
      https://setuptools.readthedocs.io/en/latest/easy_install.html
    
    Please make the appropriate changes for your system and try again.
    
    ----------------------------------------
[...]

PR #16145, opened 2020-12-10
Downgrade pip to 20.2.4

diox:
Works around https://github.com/mozilla/addons-server/issues/16144


diox:
mmm actually I'm still reproducing even with pip 20.2.4...

PR #16150, opened 2020-12-10
handle licenses (and other objects) being missing for translations

### LINKED ISSUE #16129: Collection add-ons list returns a 500 error when certain add-ons are included in the collection

Make the following request:
https://addons-dev.allizom.org/api/v4/accounts/account/10642146/collections/500/addons/

### What happened?
You receive a 500 error

Collection add-ons are returned 

There are some particularities to reproducing this issue, at least from what I've noticed after trying different scenarios:
- I encountered the issue only with some dictionaries; in the first example, the dictionary that causes the 500 is https://addons-dev.allizom.org/api/v4/accounts/account/addons/addon/3627/
- a few other collection examples where the issue can be reproduced:
  - https://addons-dev.allizom.org/api/v4/accounts/account/10640485/collections/collection-create-10/addons/ => caused by add-on 204312, also a dictionary
  -  https://addons-dev.allizom.org/api/v4/accounts/account/10642146/collections/500-error/addons/ => caused by add-on 10391, also a dictionary
- if you add a `lang=` parameter to these requests, the error disappears, so this could be somehow related to translated fields, although, the addon detail API for the problematic add-ons doesn't seem to be affected

eviljeff:
fixes #16129

### DIFF
--- a/src/olympia/amo/tests/test_utils.py
+++ b/src/olympia/amo/tests/test_utils.py
@@ -101,7 +101,7 @@ class TestAttachTransDict(TestCase):
             'es': 'Spanish 2 Name',
         }
         addon2.save()
-        attach_trans_dict(Addon, [addon, addon2])
+        attach_trans_dict(Addon, [addon, addon2, None])
         assert set(addon.translations[addon.description_id]) == (
             set([('en-us', 'English Description'), ('fr', 'French Description')])
         )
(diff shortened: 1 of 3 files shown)


willdurand (approved):

willdurand:
I'm going to merge this patch. It's fairly small but still.

PR #16161, opened 2020-12-11
Add link to scanner result on flagged versions in review page for admins

### LINKED ISSUE #15903: Link from reviewer page to scanner result page for scanner result admins
On the reviewer tools page, if a version was flagged by a scanner, we see a _Flagged by scanners_ label.

If the reviewer has the permission `Admin:ScannersResultsView` and `is_staff` is `true`, let's linkify that label and point to the scanners result page for that version.

diox:
Fixes #15903

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/includes/version.html
+++ b/src/olympia/reviewers/templates/reviewers/includes/version.html
@@ -60,7 +60,16 @@
       {% endif %}
 
       {% if version.needs_human_review %}
-        <div><strong class="risk-high">{{ _('Flagged by scanners') }}</strong></div>
+        <div><strong class="risk-high">
+          {{ _('Flagged by scanners') }}
+          {%- if request.user.is_staff and action_allowed(amo.permissions.ADMIN_SCANNERS_RESULTS_VIEW) %}
+            {% for result in version.scannerresults.all() -%}
+              {%- if loop.first %}({% endif -%}
+              <a href="{{ url('admin:scanners_scannerresult_change', result.pk ) }}">{{ result.get_scanner_display() }}</a>
+              {%- if loop.last %}){% else %}, {% endif -%}
+            {%- endfor %}
+          {%- endif %}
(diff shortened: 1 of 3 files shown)


src/olympia/reviewers/templates/reviewers/includes/version.html
```diff
@@ -60,7 +60,16 @@
       {% endif %}
 
       {% if version.needs_human_review %}
-        <div><strong class="risk-high">{{ _('Flagged by scanners') }}</strong></div>
+        <div><strong class="risk-high">
+          {{ _('Flagged by scanners') }}
+          {%- if request.user.is_staff and action_allowed(amo.permissions.ADMIN_SCANNERS_RESULTS_VIEW) %}
```
willdurand, line 65: Why do we need the `is_staff` restriction?

src/olympia/reviewers/views.py
```diff
@@ -781,12 +781,12 @@ def review(request, addon, channel=None):
...
         .select_related('reviewerflags')
-        # Prefetch scanner results, but we only care about pk, scanner and
-        # score, not the results themselves.
+        # Prefetch scanner results... but without the results json as we don't
+        # need it.
         .prefetch_related(
             Prefetch(
                 'scannerresults',
-                queryset=ScannerResult.objects.only('pk', 'scanner', 'score'),
+                queryset=ScannerResult.objects.defer('results'),
```
willdurand, line 789: ah that's cool, I didn't know about `defer`

willdurand (approved):
r+wc

src/olympia/reviewers/templates/reviewers/includes/version.html
diox: Requested in the issue. It's unlikely to happen in production, but locally or on dev, users might have the permission while not having access to the admin in general because they don't have `is_staff`. The extra condition ensures those users wouldn't see a broken link.

PR #16173, opened 2020-12-14
Optimize queries made by reviewer & admin tools to figure out if add-ons have listed/unlisted versions in a list

### LINKED ISSUE #16132: Add transformer to prefill has_listed_versions / has_unlisted_versions for a queryset

diox:
Fixes #16132

### DIFF
--- a/src/olympia/addons/admin.py
+++ b/src/olympia/addons/admin.py
@@ -6,6 +6,7 @@ from django.conf import settings
 from django.conf.urls import url
 from django.contrib import admin
 from django.core import validators
+from django.db.models import Exists, OuterRef
 from django.forms.models import modelformset_factory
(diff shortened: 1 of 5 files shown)


bobsilverberg (approved):
Looks good, nice work!

PR #16187, opened 2020-12-16
Only apply collections API endpoint caching to mozilla collections

### LINKED ISSUE #16151: Drop collection add-ons API endpoint caching if collection author isn't mozilla
The caching we added in https://github.com/mozilla/addons-server/pull/16064 for the collection add-ons API endpoint is now working, but it's a little dumb, as there is no invalidation mechanism. We only really need it for the official Fenix collections, so let's try droping it if the author isn't the mozilla user.

diox:
Fixes #16151

### DIFF
--- a/src/olympia/bandwagon/views.py
+++ b/src/olympia/bandwagon/views.py
@@ -1,6 +1,7 @@
 from django.http import HttpResponse
 from django.conf import settings
 from django.utils.decorators import method_decorator
+from django.views.decorators.cache import cache_page
 
 from rest_framework import serializers
 from rest_framework.viewsets import ModelViewSet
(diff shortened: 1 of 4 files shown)


willdurand (approved):

src/olympia/bandwagon/views.py
```diff
@@ -150,6 +150,18 @@ def list(self, request, *args, **kwargs):
...
+        # This endpoint can be quite slow so we cache the most popular
+        # collections - those from mozilla - for all anonymous users for one
+        # hour.
+        if (
+            self.kwargs['user_pk'] in ('mozilla', str(settings.TASK_USER_ID))
+            and not request.user.is_authenticated
+        ):
+            return self._cached_list(request, *args, **kwargs)
+        else:
+            return super().list(*args, *kwargs)
```
eviljeff, line 163: Why do we want to strip `request` from the args/kwargs?

diox: Good eye, that's a mistake. It fortunately doesn't have any consequences, DRF's default implementation for `list()` doesn't use that parameter or any other `args`, but I'll fix it.

PR #16189, opened 2020-12-16
Migrate to Circle CI

willdurand:
That will be awesome once it's ready.


willdurand:
As I mentioned on slack, there is no cache yet because I am not sure what to cache. It seems fast enough to me (faster than travis).

diox:
The autograph and elasticsearch steps seem rather slow, and have this at the end:
> Build was canceled

Any idea what's up with that ?

willdurand:
> Any idea what's up with that ?

They last for the duration of the job. At the end of the job, circle kills the containers, hence the "build was canceled". It's the same for all services.

diox:
Ah, confusing, but I get it.

diox:
Follow-up task for later: GitHub seems to order the jobs in the workflow alphabetically by name. Also, since it displays all of them, and CircleCI seems faster, maybe we should 1) Redispatch the tests to limit to 6 jobs to avoid scrolling and 2) Change their names to add a prefix to fix the ordering.

.circleci/config.yml
```diff
@@ -1,7 +1,321 @@
+# These environment variables must be set in CircleCI UI
+#
+# DOCKERHUB_REPO - docker hub repo, format: <username>/<repo>
+# DOCKER_USER - Login user for docker hub
+# DOCKER_PASS - Login password for docker hub user
```
diox, line 5: So when the jobs say `Warning: No authentication provided, using CircleCI credentials for pulls from Docker Hub.`, it means they are using our credentials, but that they were set up in CircleCI ?

willdurand:
> Follow-up task for later: GitHub seems to order the jobs in the workflow alphabetically by name. Also, since it displays all of them, and CircleCI seems faster, maybe we should 1) Redispatch the tests to limit to 6 jobs to avoid scrolling and 2) Change their names to add a prefix to fix the ordering.

One question about the order: does it matter? I kept the same order as before because I copied the travis config. I thought about sorting the toxenvs but I forgot, otherwise I think it would have been sorted alphabetically as well :D

.circleci/config.yml
```diff
@@ -1,7 +1,321 @@
...
+    #
+    #               The autograph configuration ends here.
+    #
+    ###########################################################################
+
+  defaults: &defaults
+    working_directory: ~/addons-server
+    docker:
+      # This is the python version we run in production.
+      - image: circleci/python:3.8
```
diox, line 282: Should we be using the so-called "next-gen" images instead ? See https://circleci.com/docs/2.0/circleci-images/ - there should be one for `python`, `redis` and `mysql` at least.

.circleci/config.yml
```diff
@@ -1,7 +1,321 @@
...
+      # services, we also need autograph, which is started in the `test` job
+      # because we need to pass a configuration file to it and it's not
+      # possible in this section.
+      #
+      # Most settings below should be kept in sync with `docker-compose.yml`.
+      - image: redis:2.8
+      - image: memcached:1.4
+      - image: mysql:8.0
+        # This is needed to avoid an error with an authentication plugin.
+        command: --default-authentication-plugin=mysql_native_password
```
diox, line 293: Mmm that is surprising. Are we using the right MySQL client libs in CircleCI ?

.circleci/config.yml
```diff
@@ -1,7 +1,321 @@
...
+        environment:
+          # Disable all xpack related features to avoid unrelated logging in
+          # docker logs. https://github.com/mozilla/addons-server/issues/8887
+          xpack.security.enabled: false
+          xpack.monitoring.enabled: false
+          xpack.graph.enabled: false
+          xpack.watcher.enabled: false
+          discovery.type: single-node
+          cluster.name: default-cluster
+          ES_JAVA_OPTS: -Xms256m -Xmx256m
```
diox, line 307: Travis had a different set of options but I guess being closer to our local env setup is better, as long as we have enough resources on the host.

.circleci/config.yml
```diff
@@ -27,34 +341,93 @@ commands:
...
+          command: dockerize -wait tcp://localhost:11211 -timeout 1m
+      - run:
+          name: Wait for elasticsearch
+          command: dockerize -wait tcp://localhost:9200 -timeout 1m
+      - run: pip install tox
+      - run: pip install --no-deps -r requirements/ci_base.txt
+      - run:
+          command: tox -e << parameters.toxenv >>
+          environment:
+            RUNNING_IN_CI: True
```
diox, line 387: Do we have anything that still depends on this ? Maybe we can remove it.

.circleci/config.yml
```diff
@@ -27,34 +341,93 @@ commands:
...
+          name: Wait for mysql
+          command: dockerize -wait tcp://localhost:3306 -timeout 1m
+      - run:
+          name: Wait for memcached
+          command: dockerize -wait tcp://localhost:11211 -timeout 1m
+      - run:
+          name: Wait for elasticsearch
+          command: dockerize -wait tcp://localhost:9200 -timeout 1m
+      - run: pip install tox
+      - run: pip install --no-deps -r requirements/ci_base.txt
```
diox, line 383: Do we need to keep `tox` and `ci_base` separated ? I think we did that because `tox` was already installed on travis or something

.circleci/config.yml
willdurand: These credentials are used for pushing the generated docker images in the `release-*` jobs. We don't need to provide our own credentials for pulling because Circle is allowed to pull lots of images by Docker.

.circleci/config.yml
willdurand: it's still referenced in our code:

```
rg RUNNING_IN_CI
src/olympia/amo/search.py
29:    if os.environ.get('RUNNING_IN_CI'):
```

.circleci/config.yml
willdurand: I think we can remove the `pip install tox` step and keep `ci_base`.

.circleci/config.yml
willdurand: I don't see `cimg` for redis or mysql so I think using the official docker ones should be good enough. I'll switch to `cimg/python`, though.

.circleci/config.yml
willdurand: I am not sure. It's using whatever is the default on Ubuntu I guess.

.circleci/config.yml
diox: I think it should be useless - it's just to set up some fake auth ?

.circleci/config.yml
diox: That's a problem, we specifically updated the MySQL client libs to use MySQL's in our image. We were also installing them in travis (See `before_install` first line).

.circleci/config.yml
willdurand: lets see if removing it does anything

.circleci/config.yml
diox: Ah, yeah. Still for `redis` and `mysql` I was wondering if using circleci ones would give us better chance of having it cached

willdurand: I used the `circleci` image for `mysql`, there is no `redis:2.8`

.circleci/config.yml
willdurand: fixed

.circleci/config.yml
```diff
@@ -1,7 +1,319 @@
...
+            duration: 26298h
+          # RSA key gen is slow and CPU intensive, so we can optionally
+          # pregenerate and cache keys with a worker pool
+          rsacacheconfig:
+            numkeys: 25
+            numgenerators: 2
+            generatorsleepduration: 1m
+            fetchtimeout: 100ms
+            statssamplerate: 1m
+          certificate: |
```
muffinresearch, line 165: Would there be a way to reference these as external files?

willdurand: no

willdurand:
As discussed on slack, I added a custom command for `checkout` that passes `--depth` to `git`. It is configured with `1` for `clone` and `10` for `fetch` as seen in existing orbs (I guess those are the common/default values then?).

scripts/autograph_travis_test_config.yaml
```diff
@@ -1,254 +0,0 @@
-# Note: Most of the configuration here got copied from
-# https://github.com/mozilla-services/autograph/blob/master/autograph.yaml
-
```
diox, line 3: This file is used by `docker-compose.yml` for `autograph` on local envs as well, I don't think it should be removed (ideally it would be included by circle somehow but as you mentioned that doesn't seem to be possible). It should be renamed though.

willdurand: awww, sorry about that. The name was misleading and I probably grep'ed the wrong filename later.

diox (approved):
r+ once you fix the autograph file for local envs. Nice job!

.circleci/config.yml
```diff
@@ -27,34 +341,163 @@ commands:
             docker build -t app:build -f << parameters.dockerfile >> --label git.commit="$CIRCLE_SHA1" .
             docker tag app:build "${DOCKERHUB_REPO}":<< parameters.image_tag >>
             docker push "${DOCKERHUB_REPO}":<< parameters.image_tag >>
+
+  better_checkout:
+    description: circle ci checkout step on steroids
```
muffinresearch, line 346: 🤣

PR #16196, opened 2020-12-17
Sync users with basket when they are banned

### LINKED ISSUE #16188: Banning a user isn't reflected in Salesforce
When we delete a user normally (they delete their account on AMO or via FxA) we trigger a sync with Basket which updates the data in Salesforce.

It doesn't happen when a user is banned, because it happens in bulk bypassing post save signals. So we need to manually add the sync trigger.

diox:
Also avoids queuing the task needlessly while the waffle is off and reduce logging 

Fixes #16188


diox:
(Forgot to add a test proving banning actually syncs the banned users, I'll add that)

willdurand (approved):

PR #16200, opened 2020-12-18
Fix checkout command

willdurand:


src/olympia/core/tests/test_certifi.py
```diff
@@ -2,8 +2,6 @@
 import pytest
 import requests
 
-from cryptography.hazmat.backends.openssl.backend import backend
```
diox, line 5: Ah yeah that's me, sorry about that, thought I had fixed that and merged without re-running the codestyle job since I knew it would fail once rebased with circleci

diox (approved):

PR #16202, opened 2020-12-18
Handle forks in better_checkout command

willdurand:


diox (approved):
:+1:  checkout appears to be correct now, `HEAD is now at 87600ab Handle forks in better_checkout command`

PR #16204, opened 2020-12-18
Only use services that are needed in the different circle jobs

willdurand:
This patch optimizes our circle ci configuration to only pull and start services we actually need for each `toxenv`.


diox (approved):
:hand

Issue #16224, opened 2020-12-23
New listed version submissions do not trigger an add-on sync with Salesforce 


AlexandraMoga:
**Prerequisites:**
AMO developer account exists in Salesforce

### Describe the problem and steps to reproduce it:
1. Log into AMO -dev with a synced developer account (see Prerequisites)
2. Upload a new listed add-on 
3. Approve the add-on with a reviewer account
4. Make sure the new add-on is listed under the developer's account info in Salesforce - make a note of the 'AMO Current Version' in the add-on details
5. Upload a new listed version for this add-on on AMO
6. Once the version is auto-approved, verify if the new version string is reflected in Salesforce 'AMO Current Version' field

### What happened?
Salesforce is still showing the previous version number

### What did you expect to happen?
Salesforce should be displaying the current listed version number

### Anything else we should know?
- the issue does not apply to unlisted add-ons/unlisted versions
- the new listed version number is updated only when certain add-on fields (name, summary etc) are manually updated on AMO

Example: https://addons-dev.allizom.org/en-US/firefox/addon/test-for-listed-sf/

1. Add-on details on AMO:
![image](https://user-images.githubusercontent.com/31961530/103010112-6e8eae00-4540-11eb-8724-d732b1fa53e1.png)

2. Add-on details in Salesforce:
![image](https://user-images.githubusercontent.com/31961530/103010150-7e0df700-4540-11eb-93c8-042aaa982263.png)

AlexandraMoga:
@diox I've reverified the issue on -dev but it doesn't look fixed.
Example:
https://addons-dev.allizom.org/en-US/developers/addon/second-test-fix-version-sync/versions

Current version on AMO:
![image](https://user-images.githubusercontent.com/31961530/103655193-9255e900-4f6f-11eb-92a0-dc9573f6a316.png)

Current version in Salesforce:
![image](https://user-images.githubusercontent.com/31961530/103655244-a1d53200-4f6f-11eb-8cf2-97aacc4b5f93.png)

There is one thing that I did not observe in my previous tests, so I'm not sure if this was the case even before the current patch: with the [new add-on example](https://addons-dev.allizom.org/en-US/developers/addon/second-test-fix-version-sync/versions), I've submitted several subsequent versions. As in my initial tests, the second listed version was not synced, but with each new version, a sync seems to be triggered, but Salesforce is still one version behind AMO (as seen in the comparison screenshots).

diox:
[...]

PR #16248, opened 2021-01-04
Trigger a sync with basket on current version change through version_changed signal

diox:
Fixes https://github.com/mozilla/addons-server/issues/16224

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -976,7 +976,7 @@ class Addon(OnChangeMixin, ModelBase):
             try:
                 self.update(**updated)
                 if send_signal and _signal:
-                    signals.version_changed.send(instance=self, sender=self.__class__)
+                    signals.version_changed.send(sender=self.__class__, instance=self)
                 log.info(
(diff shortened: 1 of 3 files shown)


src/olympia/addons/models.py
```diff
@@ -1704,10 +1704,14 @@ def git_extraction_is_in_progress(self):
 
 
 @receiver(signals.version_changed, dispatch_uid='version_changed')
-def version_changed(sender, **kw):
+def version_changed(sender, instance, **kw):
```
eviljeff, line 1707: possibly missing something, but why are we changing `sender` to be the Addon class and then passing the instance separately?  Is this just clean-up because `sender` is typically the class?

eviljeff (approved):
Is this going to add even more basket related logging in tests? 😬

src/olympia/addons/models.py
diox: Yes - consistency because I made the mistake thinking sender was a class initially when testing when it wasn't at the time :)

Issue #16272, opened 2021-01-07
TypeError in `generate_default_addons_for_frontend` command


willdurand:
The `generate_default_addons_for_frontend` command is used for the addons-frontend UI tests. It's broken currently, let's fix it.

```
Traceback (most recent call last):
  File "manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/deps/lib/python3.8/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/deps/lib/python3.8/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/deps/lib/python3.8/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/deps/lib/python3.8/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/code/src/olympia/landfill/management/commands/generate_default_addons_for_frontend.py", line 82, in handle
    serializer.create_generic_featured_addons()
  File "/code/src/olympia/landfill/serializers.py", line 97, in create_generic_featured_addons
    addon = addon_factory(
  File "/code/src/olympia/amo/tests/__init__.py", line 726, in addon_factory
    addon = Addon.objects.create(type=type_, **kwargs)
  File "/deps/lib/python3.8/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/deps/lib/python3.8/site-packages/django/db/models/query.py", line 420, in create
    obj = self.model(**kwargs)
  File "/code/src/olympia/addons/models.py", line 526, in __init__
    super(Addon, self).__init__(*args, **kw)
  File "/code/src/olympia/amo/models.py", line 188, in __init__
    super(OnChangeMixin, self).__init__(*args, **kw)
  File "/deps/lib/python3.8/site-packages/django/db/models/base.py", line 501, in __init__
    raise TypeError("%s() got an unexpected keyword argument '%s'" % (cls.__name__, kwarg))
TypeError: Addon() got an unexpected keyword argument 'recommended'
make: *** [Makefile-docker:171: setup-ui-tests] Error 1
```

PR #16273, opened 2021-01-07
Fix generate_default_addons_for_frontend

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/16272

### DIFF
--- a/src/olympia/landfill/serializers.py
+++ b/src/olympia/landfill/serializers.py
@@ -29,6 +29,8 @@ from olympia.ratings.models import Rating
 from olympia.users.models import UserProfile
 from olympia.devhub.tasks import create_version_for_upload
 from olympia.hero.models import PrimaryHero, SecondaryHero
+from olympia.promoted.models import PromotedAddon
+from olympia.constants.promoted import RECOMMENDED
 
 from .version import generate_version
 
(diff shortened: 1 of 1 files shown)


willdurand:
I "manually" tested this fix by patching addons-server in the addons-frontend test suite, see: https://app.circleci.com/pipelines/github/mozilla/addons-frontend/3452/workflows/5e6b763d-6f4b-464f-b9e0-e8223d4a684f/jobs/24872

bobsilverberg (approved):
Looks good to me. I gather there isn't anything in `addon_factory` that would allow us to pass a argument, like `promoted=RECOMMENDED`, which would do the `PromotedAddon.objects.create` for us? If not, might that be something worth adding?

r+wc

eviljeff:
> 
> 
> Looks good to me. I gather there isn't anything in `addon_factory` that would allow us to pass a argument, like `promoted=RECOMMENDED`, which would do the `PromotedAddon.objects.create` for us? If not, might that be something worth adding?
> 
> r+wc

yes, there is.  It's exactly that -  `promoted=RECOMMENDED`.  The `recommended` parameter was removed in https://github.com/mozilla/addons-server/pull/16130 (I must have overlooked the landfill usage, somehow)

Issue #16259, opened 2021-01-05
Extension approved, NOT published for over a week?


PELock:
### Describe the problem and steps to reproduce it:

1. I have published a new version of my extension https://addons.mozilla.org/en-US/firefox/addon/social-media-bot-auto-clicker/ (v1.3.0)
2. It's been approved
3. I have updated the descriptions manually at the developer panel

### What happened?

1. The submitted version is marked as v1.3.0
2. It's listed here https://addons.mozilla.org/en-US/firefox/addon/social-media-bot-auto-clicker/versions/ but it's listed as an older version than v1.2.9 (with the older date)?
3. In dev panel I can see:
 
Listed Version: 1.2.9 ?
Next Listed Version: 1.3.0 ? (with the tooltip that it isn't live on the website yet)

In recent activity in dev console I see:

Social Media Bot - Instagram, Twitter, SoundCloud properties edited. (descriptions updated)
1 week, 2 days ago by PELock

Social Media Bot - Instagram, Twitter, SoundCloud Version 1.3.0 approved.
1 week, 2 days ago by Mozilla Firefox

### What did you expect to happen?

If the extension was approved I would expect it to be available as the current version on the main site

### Anything else we should know?

Screenshots from dev panel console:

https://addons.mozilla.org/en-US/firefox/addon/social-media-bot-auto-clicker/

wagnerand:
All statuses look correct to me, @diox mentioned this is a known but unidentified rare bug. I don't recall it at the moment, so I am not sure when we've seen it last time or if there is a related bug.

I believe mat suggested to take a look at logs to see if we can spot anything.

diox:
From the activity log, I have a good idea of what happened:
- The version was approved at 2020-12-27 01:00 exactly
- The developer mades changes in devhub at exactly the same time (and also one minute, two minutes and four minutes before - it was basically approved just as they were updating reviewer notes and uploading source code)

So my guess is that one of the forms (or all of them...) in devhub that modifies the add-on saves all fields (django's default behavior) and it overrode `_current_version` with the value from before the approval.

PELock:
I did upgrade the descriptions like I always do (with version changes detail description), right after it was approved, maybe its some concurrency bug?

diox:
Yes. Between the database transactions, replication, and the fact that devhub forms potentially override data that has since been modified, it's possible to end up in this situation if everything happens in the same super short period of time.

I am looking into fixing this, but this is not trivial...

[...]

PR #16275, opened 2021-01-07
Only send updated fields to the database UPDATE queries where possible

diox:
This avoids accidentally overwriting fields that have been modified in parallel

Fixes https://github.com/mozilla/addons-server/issues/16259

### DIFF
--- a/src/olympia/addons/tasks.py
+++ b/src/olympia/addons/tasks.py
@@ -31,12 +31,6 @@ log = olympia.core.logger.getLogger('z.task')
 @task
 @use_primary_db
 def version_changed(addon_id, **kw):
-    update_last_updated(addon_id)
-    update_appsupport([addon_id])
-
-
-def update_last_updated(addon_id):
-    queries = Addon._last_updated_queries()
     try:
         addon = Addon.objects.get(pk=addon_id)
     except Addon.DoesNotExist:
@@ -44,18 +38,25 @@ def update_last_updated(addon_id):
             '[1@None] Updating last updated for %s failed, no addon found' % addon_id
         )
         return
+    log.info('[1@None] Updating last updated for %s.' % addon.pk)
+    addon.update(last_updated=compute_last_updated(addon))
+    update_appsupport([addon.pk])
 
-    log.info('[1@None] Updating last updated for %s.' % addon_id)
 
+def compute_last_updated(addon):
+    queries = Addon._last_updated_queries()
     if addon.status == amo.STATUS_APPROVED:
         q = 'public'
     else:
         q = 'exp'
-    qs = queries[q].filter(pk=addon_id).using('default')
-    res = qs.values_list('id', 'last_updated')
-    if res:
-        pk, t = res[0]
-        Addon.objects.filter(pk=pk).update(last_updated=t)
+    values = (
+        queries[q]
+        .filter(pk=addon.pk)
+        .using('default')
+        .values_list('last_updated', flat=True)
+    )
+    if values:
+        addon.update(last_updated=values[0])
 
 
 @task
--- a/src/olympia/addons/tasks.py
+++ b/src/olympia/addons/tasks.py
@@ -16,6 +16,7 @@ from olympia.addons.models import (
     attach_tags,
     attach_translations_dict,
 )
+from olympia.addons.utils import compute_last_updated
 from olympia.amo.celery import task
 from olympia.amo.decorators import use_primary_db
 from olympia.amo.utils import LocalFileStorage, extract_colors_from_image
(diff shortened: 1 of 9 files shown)


src/olympia/devhub/forms.py
```diff
@@ -859,7 +859,7 @@ def get_context(self, name, value, attrs):
 class VersionForm(WithSourceMixin, forms.ModelForm):
     release_notes = TransField(widget=TransTextarea(), required=False)
     approval_notes = forms.CharField(
-        widget=TranslationTextarea(attrs={'rows': 4}), required=False
+        widget=forms.Textarea(attrs={'rows': 4}), required=False
```
diox, line 862: Drive-by because some test was initially failing because of approval notes so I was digging around there. Approval notes is just a `CharField` in the model, there shouldn't be any reason to use a `TranslationTextarea`.

src/olympia/users/models.py
```diff
@@ -1059,8 +1059,12 @@ def watch_changes(old_attr=None, new_attr=None, instance=None, sender=None, **kw
     }
 
     # Log email changes.
-    if 'email' in changes and new_attr['email'] is not None:
-        log.info('Creating user history for user: %s' % instance.pk)
+    if (
+        'email' in changes
+        and new_attr['email'] is not None
+        and old_attr.get('email') is not None
```
diox, line 1065: Technically can happen now that the initial attributes reflects initial state correctly for new instances - `email` has no default, so you start with a `UserProfile` instance with no email, and then set one, it would fire and `old_attr['email']` would be empty. A couple tests were broken because of it because then we would attempt to create a `UserHistory` with an empty email.

src/olympia/addons/tasks.py
```diff
@@ -31,31 +31,31 @@
...
             '[1@None] Updating last updated for %s failed, no addon found' % addon_id
         )
         return
+    log.info('[1@None] Updating last updated for %s.' % addon.pk)
+    addon.update(last_updated=compute_last_updated(addon))
+    update_appsupport([addon.pk])
 
-    log.info('[1@None] Updating last updated for %s.' % addon_id)
 
+def compute_last_updated(addon):
```
eviljeff, line 46: I'm going to have a proper look at this next week but just noting down one thing before I forget: imo `compute_last_updated` should be in utils.py (or elsewhere) as it's not actually a task.

diox: Yeah I agree, I'll move it

src/olympia/addons/tasks.py
```diff
@@ -31,31 +31,31 @@
...
+    queries = Addon._last_updated_queries()
     if addon.status == amo.STATUS_APPROVED:
         q = 'public'
     else:
         q = 'exp'
```
eviljeff, line 51: I'm aware that you're not touching this functionality but these `_last_updated_queries` are weird.  If the addon status isn't `STATUS_APPROVED` (e.g. `STATUS_NOMINATED`) then you get the "exp" query ("experimental"??) which excludes `STATUS_NOMINATED`.

src/olympia/amo/models.py
```diff
@@ -238,29 +248,77 @@ def _send_changes(self, old_attr, new_attr_kw):
...
+            kwargs['update_fields'] = list(
+                dict(
+                    {(k, self.__dict__[k]) for k in initial_attrs}
+                    - set(initial_attrs.items())
+                    # Never include primary key field - it might be set to None
+                    # initially in initial_attrs right after a call to create()
+                    # even though self.pk is set.
+                    - set(((self._meta.pk.name, self.pk),))
+                ).keys()
+            ) + [f.name for f in self._meta.fields if getattr(f, 'auto_now', False)]
```
eviljeff, line 291: Although clever, imo this chunk of code (single statement!) is difficult to parse - it's tuples inside a set inside a dict inside a list.  Could it be rewritten slightly? (maybe bring the tuples + sets out into an `updated_values` variable or something?)

eviljeff (approved):

src/olympia/amo/models.py
diox: Yeah, it grew bit by bit and spiraled out of control... I'll split it into variables

src/olympia/addons/tasks.py
diox: Yeah. I filed https://github.com/mozilla/addons-server/issues/16295

src/olympia/amo/models.py
diox: b86edcaa0d7026a2fa82068cdf8fc14267ef8486

src/olympia/addons/tasks.py
diox: 222cee1be4af80fac89b0f4abbb98101290ab0a4

PR #16284, opened 2021-01-08
Do not disable deleted add-on when blocking

### LINKED ISSUE #16283: Do not override add-on status when blocking a deleted add-on.

When an add-on gets blocked with range 0-*, we set it's status to `DISABLED_BY_MOZILLA` to prevent further submissions.
However, when an already deleted add-on gets blocked, we shouldn't change the status but instead deny resubmission (add the GUID to `DeniedGuid`)

STR:
1. Delete an add-on.
2. Block it.

### What happened?

Add-on status is `DISABLED_BY_MOZILLA`

Add-on status is `DELETED`, and add-on GUID has been added to `DeniedGuid`.

wagnerand:
Fix #16283.

~Could probably use an additional quick check by @eviljeff.~ Hah, the magic wheel picked you already. :)

### DIFF
--- a/src/olympia/blocklist/tests/test_admin.py
+++ b/src/olympia/blocklist/tests/test_admin.py
@@ -6,6 +6,7 @@ from unittest import mock
 from django.conf import settings
 from django.contrib.admin.models import LogEntry, ADDITION
 from django.contrib.contenttypes.models import ContentType
+from olympia.addons.models import DeniedGuid
 from olympia.constants.activity import BLOCKLIST_SIGNOFF
 
 from pyquery import PyQuery as pq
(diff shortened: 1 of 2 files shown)


src/olympia/blocklist/tests/test_admin.py
```diff
@@ -1729,6 +1729,45 @@ def test_list_filters(self):
...
+                '_save': 'Save',
+            },
+            follow=True,
+        )
+        assert response.status_code == 200
+        assert Block.objects.count() == 1
+        block = Block.objects.first()
+        assert block.addon == deleted_addon
+        deleted_addon.reload()
+        assert deleted_addon.status == amo.STATUS_DELETED  # Should stay deleted
```
eviljeff, line 1769: can you check there's an entry for the guid in `DeniedGuid` for completeness?

eviljeff (approved):
looks good, r+

src/olympia/blocklist/tests/test_admin.py
wagnerand: Yes, just noticed myself! [`395d9ec` (#16284)](https://github.com/mozilla/addons-server/pull/16284/commits/395d9eca89d7699f6c16a1e45c6cae47f85128d4)

PR #16285, opened 2021-01-08
migrate addons using prefined icons to custom icons

### LINKED ISSUE #16114: migrate the addons that use icons from the preselected list
follow up for #10090 - the icon selection has been removed, but to completely remove the feature we need to give each add-on that chose one of the icons a copy of it.

A quick redash query shows 471 approved and enabled add-ons (and another 2446 in various other states that are potentially usuable by the developer, i.e. not deleted/disabled, if we want to fully support them)

eviljeff:
fixes #16114

### DIFF
--- a/src/olympia/addons/management/commands/process_addons.py
+++ b/src/olympia/addons/management/commands/process_addons.py
@@ -9,6 +9,7 @@ from olympia import amo
 from olympia.addons.models import Addon
 from olympia.addons.tasks import (
     add_dynamic_theme_tag,
+    create_custom_icon_from_predefined,
     delete_addons,
     extract_colors_from_static_themes,
     find_inconsistencies_between_es_and_db,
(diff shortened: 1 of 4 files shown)


bobsilverberg (approved):
Looks good, thanks @eviljeff. I have a couple of questions:

1. After this has done, can/should we delete the files for the pre-defined icons? 
2. When does the `process_addons` command run? It seems like we're adding this to a command that will be run more than once in the future, but isn't this something that only ever needs to be run once? If so, would a data migration make more sense?

r+wc

eviljeff:
>     1. After this has done, can/should we delete the files for the pre-defined icons?

yep! And the supporting code in django.  And eventually the JS in devhub that handles the icon selection.
 
>     2. When does the `process_addons` command run? It seems like we're adding this to a command that will be run more than once in the future, but isn't this something that only ever needs to be run once? If so, would a data migration make more sense?

It's a one off command but we don't know how long it'll take, and as it's not blocking a DB migration, it's easier to have it as a command we can have ops run after the push at our leisure.

Issue #16301, opened 2021-01-12
Add "kill-switch" for BigQuery


willdurand:
We don't have a waffle switch to temporarily disable BigQuery but we probably should in case GCP goes crazy again.

For QA: there should be a new `disable-bigquery` switch. When this switch is turned **on**, the stats pages should be empty (no data available) and there should be a banner.

ioanarusiczki:
@willdurand 
- I made `disable-bigquery` active -> the data is no longer available - only numbers  from the blue URLs in the overview pages are available (in case that's unexpected, let me know). 

![numbers are available](https://user-images.githubusercontent.com/33448286/105045744-efbe5f80-5a70-11eb-8a49-a4bb319c3039.png)

- After I disabled the switch, the banner is no longer present but I could not get the data back. 
Do you see anything for :

https://addons-dev.allizom.org/en-US/firefox/addon/amo-info-with-extra-dirs/statistics/
https://addons-dev.allizom.org/en-US/firefox/addon/facebook-container/statistics/
https://addons-dev.allizom.org/en-US/firefox/addon/firefox-color/statistics/
https://addons-dev.allizom.org/en-US/firefox/addon/private-relay/statistics/

I opened the pages in a new profile but I still don't see data displayed.

willdurand:
> I opened the pages in a new profile but I still don't see data displayed.

Possibly a different issue: current -dev JavaScript is broken

ioanarusiczki:
@willdurand Ok, then I'll re-check this later.

ioanarusiczki:
@willdurand Looks like it's broken on both -dev and -stage, I get 503 Service Unavailable errors, when the switch is enabled. 

![js errors on both servers](https://user-images.githubusercontent.com/33448286/105179872-899a1100-5b32-11eb-9a1c-7bada3632353.gif)

I don't get errors when the switch is disabled.

willdurand:
@ioanarusiczki By "broken", do you mean the loading indicators aren't hidden? I noticed this too, which is odd because that's not what I saw locally. It's not a huge problem but you can certainly create a new p4 issue for that.

ioanarusiczki:
@willdurand  
Sometimes they even show data so I filed https://github.com/mozilla/addons-server/issues/16366

PR #16304, opened 2021-01-12
Add switch to enable/disable BigQuery

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/16301


src/olympia/stats/templates/stats/stats.html
```diff
@@ -89,6 +89,14 @@ <h1>
 {% endblock %}
 
 {% block content %}
+  {% if waffle.switch('disable-bigquery') %}
+    <div class="notification-box info">
+      {% trans %}
+        Sorry, we had to disable the statistics temporarily.
```
diox, line 95: Feels weird to use "we" on public pages like this. Probably not a huge deal since it's a panic switch though.

src/olympia/stats/views.py
```diff
@@ -149,12 +120,15 @@ def overview_series(request, addon, group, start, end, format):
...
-    series = zip_overview(downloads, updates)
+    series = []
+    if not switch_is_active('disable-bigquery'):
+        downloads = get_download_series(
+            addon=addon, start_date=start_date, end_date=end_date
+        )
+        updates = get_updates_series(
+            addon=addon, start_date=start_date, end_date=end_date
+        )
+        series = zip_overview(downloads, updates)
```
diox, line 132: Here and below I think we need to indicate that there was a temporary issue in the response as well. Probably return a 503 response somehow.

src/olympia/stats/templates/stats/stats.html
willdurand: I'll rephrase a bit

src/olympia/stats/views.py
willdurand: Do you think 503 is accurate? I am wondering how that will work with the js :p It's certainly better that returning empty data so I'll see what I can do

diox (approved):

static/js/impala/stats/manager.js
```diff
@@ -409,7 +409,12 @@ z.StatsManager = (function () {
       error: errorHandler,
     });
 
-    function errorHandler() {
+    function errorHandler(response, unused1, unused2) {
+      if (response.status === 503) {
```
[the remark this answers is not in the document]
willdurand: yeah, that's tricky tbh

PR #16317, opened 2021-01-14
Include hero shelves in /shelves/ response

### LINKED ISSUE #16309: return hero shelves data under /shelves/
We have a new /shelves/ api endpoint but to get the primary and secondary hero shelves you have to make a separate call to /hero/.  Given the end objective to have a single API request to populate the homepage with content it makes sense to return everything at once in a single call.

eviljeff:
fixes #16309 @bobsilverberg does the api response seem reasonable?

### DIFF
--- a/src/olympia/shelves/tests/test_views.py
+++ b/src/olympia/shelves/tests/test_views.py
@@ -179,7 +179,7 @@ class TestShelfViewSet(ESTestCase):
         assert result['results'][1]['addons'][0]['type'] == 'extension'
 
     # If we delete HeroShelvesView move all the TestHeroShelvesView tests here
-    def test_hero_shelves_in_response(self):
+    def test_only_hero_shelves_in_response(self):
         phero = PrimaryHero.objects.create(
             promoted_addon=PromotedAddon.objects.create(addon=addon_factory()),
             enabled=True,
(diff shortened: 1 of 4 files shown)


src/olympia/shelves/tests/test_views.py
```diff
@@ -168,6 +178,37 @@ def test_only_enabled_shelves_in_view(self):
         )
         assert result['results'][1]['addons'][0]['type'] == 'extension'
 
+    # If we delete HeroShelvesView move all the TestHeroShelvesView tests here
+    def test_hero_shelves_in_response(self):
+        phero = PrimaryHero.objects.create(
```
bobsilverberg, line 183: I read that as a single word "phero", like "fero". Not a comment to address. Just something I found funny. Same with "shero", below (is that like a female hero)?

bobsilverberg (approved):
The API looks fine to me, thanks for asking! One question about the patch: should we have a test where we have data for both homepage shelves and hero shelves? It looks like the test added for heroes doesn't include any homepage shelves.

eviljeff:
> One question about the patch: should we have a test where we have data for both homepage shelves and hero shelves? It looks like the test added for heroes doesn't include any homepage shelves.

Yeah... I thought it'd be unnecessary as both scenarios were tested but that kind of thing has bitten us in this api before so I better add one.

bobsilverberg:
> > One question about the patch: should we have a test where we have data for both homepage shelves and hero shelves? It looks like the test added for heroes doesn't include any homepage shelves.
> 
> Yeah... I thought it'd be unnecessary as both scenarios were tested but that kind of thing has bitten us in this api before so I better add one.

Bitten us in the api. 🤣

PR #16319, opened 2021-01-14
Add missing migrations

wagnerand:
I believe @eviljeff would like to take a look at least at the migration for regional restrictions.

I am not sure what changes caused the other migration, it looks faily generic.


eviljeff:
they're both no-ops at the database level (so arguably pointless).  The activity is just choices changing (because the list of activity type constants changed); the addon regional restriction is `'.'` being added to the `help_text`

wagnerand:
Ok, I am happy to close this if it's pointless. How do we tell django to ignore the changes so it stops nagging about missing migrations?

eviljeff:
I don't know of a way to silence the checks, though I've only noticed django actually warning you if you were adding new migrations so, personally, I'd just add them at that point so there's no extra work.
At this point you've already done the work so 🤷

wagnerand:
I noticed them during running `make update_docker`, which eventually ends up at https://github.com/mozilla/addons-server/blob/master/Makefile-docker#L136:L136.

So, are you ok with merging this?

eviljeff (approved):
🤷 when tests are passing

wagnerand:
Test failure is unrelated.

PR #16326, opened 2021-01-15
drop AppSupport model

### LINKED ISSUE #12315: Remove AppSupport model
`AppSupport` is used to denormalize application support info, making it easier to find out what apps an add-on support from the `Addon` instance. It was used before in various parts of legacy frontend, when making sql queries listing add-ons compatible with the current app.

The new frontend doesn't rely on it, instead using the search API for all its "browse" queries, and therefore depending on elasticsearch, not sql, making that model obsolete.

Let's remove it. A few things still depend on it but it should be possible to work around that.

eviljeff:
fixes #12315 (and changes the language tools api along the way)


src/olympia/addons/migrations/0029_delete_appsupport.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 2.2.17 on 2021-01-14 18:05
```
eviljeff, line 1: Actually, we'll have to land this in a following tag.  I'll put it out before merging.

willdurand: Could you move that file to another PR then?

eviljeff: I will.

willdurand (approved):

willdurand:
Is there an issue filed for the migration file?

PR #16327, opened 2021-01-15
differentiate search shelves for themes

### LINKED ISSUE #16312: Add `themes` property to homepage shelf data
When the homepage starts consuming the [API for homepage shelves](https://addons-server.readthedocs.io/en/latest/topics/api/shelves.html#homepage-shelves), it will need to know whether the shelf is for themes or extensions. This is required because theme shelves are handled differently (e.g., we show 3 themes, but for extensions we show 4 extensions).

This means that we need a property added to the model to indicate if the shelf is for themes, we need the admin updated to allow this property to be specified, and we need the API updated to expose this property.

We did discuss a possible hack which would use the `criteria` returned from the API and inspect it to guess whether it is returning themes or not, but this feels like a hack and probably not the best way to go to solve this.

@eviljeff @diox What do you think about the hack idea, and do you agree it's not a good solution?

eviljeff:
fixes #16312

### DIFF
--- a/docs/topics/api/shelves.rst
+++ b/docs/topics/api/shelves.rst
@@ -103,13 +103,26 @@ hero shelf is returned, along with the enabled shelves.
     :>json array results: The array of shelves displayed on the AMO Homepage.
     :>json string results[].title: The title of the shelf.
     :>json string results[].url: The configured URL using the shelf's endpoint and criteria; links to the shelf's returned add-ons.
-    :>json string results[].endpoint: The endpoint selected for the shelf.
+    :>json string results[].endpoint: The :ref:`endpoint type <shelf-endpoint-type>` selected for the shelf.
     :>json string results[].criteria: The criteria for the addons in the shelf.
     :>json string|null results[].footer_text: The optional text in the footer of the shelf.
     :>json string|null results[].footer_pathname: The optional pathname of the URL for the footer's text.
     :>json array results[].addons: An array of :ref:`add-ons <addon-detail-object>` or :ref:`collections <collection-detail-object>`.
     :>json object primary: A :ref:`primary hero shelf <primary-hero-shelf>`.
     :>json object secondary: A :ref:`secondary hero shelf <secondary-hero-shelf>`.
+
+.. _shelf-endpoint-type:
+
+    Possible values for the ``endpoint`` field:
+
+    ==============  ====================================================================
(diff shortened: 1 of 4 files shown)


eviljeff:
note: whichever one of this or #16321 is merged first is going to cause a merge conflict in the other.

bobsilverberg (approved):
Thanks for this @eviljeff. The code looks good, but I'm not 100% on adding another value to `endpoint` for this. Sure, it works, and it's a pretty simple way to solve it, but it feels like it's not accurate. The shelf isn't actually using a different endpoint, it's still using the `search` endpoint, but with different criteria. I was thinking more of a checkbox for "contains themes", but I realize that's more complex as it's a new property.

I'm fine with it the way it is, if you think it's best, but I wanted to at least raise this as something to think about.

r+wc

eviljeff:
> I'm not 100% on adding another value to `endpoint` for this. Sure, it works, and it's a pretty simple way to solve it, but it feels like it's not accurate. The shelf isn't actually using a different endpoint, it's still using the `search` endpoint, but with different criteria. I was thinking more of a checkbox for "contains themes", but I realize that's more complex as it's a new property.
> 
> I'm fine with it the way it is, if you think it's best, but I wanted to at least raise this as something to think about.

Yeah, I saw you called that approach a hack in the issue but `endpoint` is really more of a "shelf type" in the implementation (that have to be handled differently), so it didn't seem too bad to expand the types from 2 to 3. And yes it was easier to change the possible values of an existing property than add a new property to the model, admin, form,and serializer (& test each of those parts) 😄 

We can revisit if this simple fix turns out to be a hassle to handle.

xlisachan:
@eviljeff, @bobsilverberg -`shelves/serializers.py` would need to be updated (line 33 and 51) to include the new `search-themes` endpoint, should I make this revision?
Also, I am unable to save homepage shelves in admin - an error is raised (`No data found - check criteria parameters.`), which is from the NoReverseMatch exception in `shelves/forms.py`. It states `addon-search` and `collection-addon-list` are no longer valid view function or pattern names. I'm not sure how to resolve this.

eviljeff:
😐  I'll fix it.  I shouldn't trust the tests passing meaning everything is okay.

eviljeff:
regression should be fixed by #16347

PR #16328, opened 2021-01-15
Order auto-approved queue by MAD scanner score

### LINKED ISSUE #16263: Add scanner score to auto-approval queue table and order by it
To help with prioritization efforts, let's add the security scanner score to the auto-approval review queue and also order by it (descending).

We should keep the current weight column as well. Optional: If it's easy to do, could we provide a way to load that queue ordered by weight, e.g. by passing a GET parameter? If not, no big deal, we can skip that.

diox:
Score is copied at auto-approval time in a denormalized field for perf and to make the query simpler.

Fixes #16263


src/olympia/reviewers/migrations/0013_backfill_autoapprovalsummary_score.py
```diff
@@ -0,0 +1,56 @@
...
+        # we have to cast and multiply by 100 to match the way we store scores
+        # in AutoApprovalSummary.
+        score=Cast(
+            Subquery(
+                ScannerResult.objects.filter(
+                    version=OuterRef('version'),
+                    scanner=MAD,
+                ).values('score')[:1]
+            )
+            * 100,
```
willdurand, line 35: Should we do this here or rather in the view (template) layer? We're loosing some information here (maybe not important, though)

src/olympia/reviewers/migrations/0013_backfill_autoapprovalsummary_score.py
```diff
@@ -0,0 +1,56 @@
...
+            IntegerField(),
+        )
+    )
+
+
+def reverse_func(apps, schema_editor):
```
willdurand, line 41: mm is this called when we want to downgrade a migration?

src/olympia/reviewers/models.py
```diff
@@ -1142,6 +1143,14 @@ def calculate_static_analysis_weight_factors(self):
             }
         return factors
 
+    def calculate_score(self):
+        """Compute maliciousness score for this version."""
+        # Some precision is lost but we don't particularly care that much, it's
```
willdurand, line 1148: Ha! I guess that answers my question above :D

willdurand (approved):

src/olympia/reviewers/migrations/0013_backfill_autoapprovalsummary_score.py
diox: Yeah. We typically don't bother (we should!) but I did a bunch of runs of that migration while writing it and so I added that to quickly go back to the previous test before doing another run

src/olympia/reviewers/models.py
diox: Yeah, I don't think we care. It's a denormalized field, only really there because otherwise the query is nasty. And even if we were to end up using it more, I don't think the precision would matter for what we're doing with it.

PR #16336, opened 2021-01-18
cleaning up remaining templates from #10918

### LINKED ISSUE #16294: tidy up MDN search field in devhub base templates that's no longer used
When the base templates were [cleaned up](https://github.com/mozilla/addons-server/pull/10926/files#diff-520fc674741702badfb6599cea9cce0f5a93e5bec543bf24caeb834d99989aa3L100 (edited) ) as part of the legacy frontend removal the search field was (knowingly) dropped from the devhub pages too.  But the code in the devhub base templates wasn't.  That should be removed.

ankushduacodes:
Fixes #16294

Changes Introduced:
-- Minor Template cleanup

### DIFF
--- a/src/olympia/amo/tests/test_csp_headers.py
+++ b/src/olympia/amo/tests/test_csp_headers.py
@@ -94,6 +94,7 @@ class TestCSPHeaders(TestCase):
         assert "'self'" in base_settings.CSP_FRAME_SRC
         assert "'self'" in base_settings.CSP_IMG_SRC
         assert "'self'" in base_settings.CSP_STYLE_SRC
+        assert "'self'" in base_settings.CSP_FORM_ACTION
 
     def test_not_self_in_script_src(self):
         """script-src should not need 'self' or a.m.o for services.a.m.o"""
--- a/src/olympia/amo/tests/test_csp_headers.py
+++ b/src/olympia/amo/tests/test_csp_headers.py
@@ -74,11 +74,6 @@ class TestCSPHeaders(TestCase):
         for val in base_settings.CSP_IMG_SRC:
             assert not val.startswith('http:')
 
-    def test_http_protocol_not_in_form_action(self):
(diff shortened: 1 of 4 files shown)


ankushduacodes:
@eviljeff One of the tests are failing I am not sure why. But can confirm that this same error about `local_settings` import came up when I was trying to setup the dev env locally.

Any insight?

eviljeff (approved):
thanks r+

src/olympia/lib/settings_base.py
```diff
@@ -1286,10 +1286,6 @@ def get_db_config(environ_var, atomic_requests=True):
     ANALYTICS_HOST,
     PROD_CDN_HOST,
 )
-CSP_FORM_ACTION = (
-    "'self'",
-    'https://developer.mozilla.org',
```
eviljeff, line 1291: actually, can we change this to just drop the `https:// ....` line, and keep the rest of the tuple as now.

src/olympia/amo/tests/test_csp_headers.py
```diff
@@ -97,7 +92,6 @@ def test_self_in_common_settings(self):
         assert "'self'" in base_settings.CSP_CONNECT_SRC
         assert "'self'" in base_settings.CSP_CHILD_SRC
         assert "'self'" in base_settings.CSP_FRAME_SRC
-        assert "'self'" in base_settings.CSP_FORM_ACTION
```
eviljeff, line 100: and revert this line too

eviljeff (changes requested):
sorry, we looked further into how the dependency (https://django-csp.readthedocs.io/en/latest/configuration.html) uses these settings and it appears that `CSP_FORM_ACTION` _does_ need to be defined, even when it's `"'self'"`.

eviljeff (approved):
thanks! r+

PR #16337, opened 2021-01-18
Enable gzip middleware for everything again

diox:
This was previously disabled to mitigate [BREACH](https://en.wikipedia.org/wiki/BREACH) attacks. Nowadays we are using `SameSite=Lax` on all session/auth cookies (and `CSRF_USE_SESSIONS` is enabled) so it should be safe to re-enable gzip for everything.

Ref mozilla/addons#1217

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -403,9 +403,9 @@ MIDDLEWARE = (
     # Test if it's an API request first so later middlewares don't need to.
     # Also add relevant Vary header to API responses.
     'olympia.api.middleware.APIRequestMiddleware',
-    # Gzip (for API only) middleware needs to be executed after every
-    # modification to the response, so it's placed at the top of the list.
-    'olympia.api.middleware.GZipMiddlewareForAPIOnly',
+    # Gzip middleware needs to be executed after every modification to the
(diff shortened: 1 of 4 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -425,8 +425,8 @@ def get_db_config(environ_var, atomic_requests=True):
     'corsheaders.middleware.CorsMiddleware',
     # Enable conditional processing, e.g ETags.
     'django.middleware.http.ConditionalGetMiddleware',
-    'olympia.amo.middleware.CommonMiddleware',
     'olympia.amo.middleware.NoVarySessionMiddleware',
+    'olympia.amo.middleware.CommonMiddleware',
```
eviljeff, line 429: deliberate move?

eviljeff (approved):

src/olympia/lib/settings_base.py
diox: Yeah, https://docs.djangoproject.com/en/2.2/ref/middleware/#middleware-ordering mentions the right place to put `CommonMiddleware` is:
> After `SessionMiddleware` if you’re using `CSRF_USE_SESSIONS`

PR #16343, opened 2021-01-19
drop Category model

### LINKED ISSUE #5981: Get rid of the Category model
We shouldn't need the `Category` model, the `StaticCategory` constants should be enough, together with, say, a simple `JSONField` to store categories on `Addon`s.

eviljeff:
fixes #5981 

The migration is a challenge - without it the tests failed because things like the fk constraint prevent an id from being referenced if there wasn't a category record (we generate Category instances in the tests - though don't in production use). 

 In _theory_ the data is the same as before (it's still an integer column called `category_id`) so won't cause any problems during the deploy with the migration... and wouldn't cause issues if it wasn't there either.  In practise 🤷


bobsilverberg (approved):
Looks good, thanks for the patch, @eviljeff. I take it that this patch doesn't address the issue of unifying the categories across apps as was discussed in the related issue?

eviljeff:
> 
> 
> Looks good, thanks for the patch, @eviljeff. I take it that this patch doesn't address the issue of unifying the categories across apps as was discussed in the related issue?

no, that's a different issue.

src/olympia/addons/migrations/0029_auto_20210118_1527.py
```diff
@@ -0,0 +1,50 @@
...
+        ),
+        migrations.AlterField(
+            model_name='addoncategory',
+            name='category',
+            field=models.SmallIntegerField(),
+        ),
+        migrations.RenameField(
+            model_name='addoncategory',
+            old_name='category',
+            new_name='category_id',
```
diox, line 37: Can you avoid this ? This isn't backwards-compatible, that would cause loading/saving of addons to fail as soon as the migration is applied until the new code is deployed.

src/olympia/addons/migrations/0029_auto_20210118_1527.py
```diff
@@ -0,0 +1,50 @@
...
+        ),
+        migrations.AddIndex(
+            model_name='addoncategory',
+            index=models.Index(fields=['category_id', 'addon'], name='category_id_addon_idx'),
+        ),
+        migrations.AddConstraint(
+            model_name='addoncategory',
+            constraint=models.UniqueConstraint(fields=('addon', 'category_id'), name='category_id_addon_id'),
+        ),
+        migrations.DeleteModel(
```
diox, line 47: Wouldn't that fail with the old code as soon as the migration is applied too ?

src/olympia/addons/migrations/0029_auto_20210118_1527.py
eviljeff: the field in the database is already `category_id` because it's a foreign key, but the `AlterField` would rename it to `category` (when the field type changes), so this just changes it back.  There would be a moment during the migration where it database queries would fail though - I couldn't see a way to have AlterField not change the name and have it all be done in one step.

src/olympia/addons/migrations/0029_auto_20210118_1527.py
eviljeff: Actually yeah, this step would break lookups during the deploy.  If we opt for leaving the rest of the migration 2021-01-28 in  this would have to be in later tag.

eviljeff: It's a big mess (in pre-django-migration days we could just drop the constraint and that would be all that was needed!)

eviljeff: The alternative is we add a _new_ column for the category id, and copy the value from `category_id` during the migration.  But that seemed wasteful (and probably really slow)

src/olympia/addons/models.py
```diff
@@ -1910,17 +1905,28 @@ class Meta:
 class AddonCategory(models.Model):
     id = PositiveAutoField(primary_key=True)
     addon = models.ForeignKey(Addon, on_delete=models.CASCADE)
-    category = models.ForeignKey('Category', on_delete=models.CASCADE)
+    category_id = models.SmallIntegerField()
```
eviljeff, line 1908: Thinking: if we change this for whatever size integer is used for foreign keys in mysql we could save the database field size conversion in the migration.

diox: Yeah. I was thinking of doing something like `category = models.IntegerField(db_column='category_id')` but it doesn't seem to work...

diox: Another trick I was thinking about would be to do keep `ForeignKey` with `db_constraint=False`... And then change it to an `IntegerField` later, but the intermediary step would help convince migrations that it's ok. (Haven't completely thought this through yet)

diox: One final option: use `RunSQL`, providing `state_operations` to help migrations system understand what we're doing: https://docs.djangoproject.com/en/2.2/ref/migration-operations/#runsql ; then we'd be free to do the `ALTER` we want and leave things in a state migrations would understand.

eviljeff: https://github.com/mozilla/addons-server/pull/16343/commits/c35b82fbac587e09b33df9696a621f40213e7c1a results in just one `ALTER` to drop the fk constraint.  It's difficult to know if that's enough, but I played around locally with tests from the master branch running on a post-migration database and they passed (addons/tests/test_models.py anyway - I didn't run the entire suite)

The drop model for the `Category` model will have to be in a later tag, obvs.

eviljeff: (The field type was changed to PositiveIntegerField too so the column type doesn't have to change)

diox (approved):
:+1: working nicely, good job, one fewer model to worry about!

- [x] database schema with and without migrations is identical.
- [x] `sqlmigrate` shows only the `DROP FOREIGN KEY` is ran

PR #16349, opened 2021-01-19
Fix rating URL showing for non staff reviewers

### LINKED ISSUE #15930: Don't redirect to django users that don't have a mozilla e-mail address
Followup for https://github.com/mozilla/addons-server/issues/10113#issuecomment-722292203

1. I've added to the Add-ons Reviewer group a user with an e-mail address different than mozilla.com and added the Ratings:Moderate permission. 
2. Go to a disabled addon review page -> https://reviewers.addons.allizom.org/en-US/reviewers/review/facebook-container88
3. Click the rating URL

Expected result:
it should be read only 

Actual result:
It's throws a 403 Forbidden error

Notes:
reproducible on AMO stage - FF82(Win10).

renancleyson-dev:
Fixes #15930

The ```acl_is_review_moderator``` value now checks if the user is of the staff(have a Mozilla e-mail address).

### DIFF
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -5637,8 +5637,7 @@ class TestReview(ReviewBase):
         response = self.client.get(self.url)
         assert response.status_code == 200
         doc = pq(response.content)
-        assert doc('.addon-rating strong')
-        assert doc('.addon-rating strong').text() == '1 review'
+        assert not doc('.addon-rating a')
 
     def test_user_ratings_unlisted_addon(self):
         user = UserProfile.objects.get(email='reviewer@mozilla.com')
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -6096,6 +6096,25 @@ class TestReview(ReviewBase):
 
         self.assertRedirects(response, self.url)
 
+    def test_review_moderator_addon_rating_read_only(self):
+        user = user_factory()
+        Rating.objects.create(
+            body='Lorem ipsum dolor',
+            rating=3,
+            ip_address='10.5.6.7',
+            addon=self.addon,
+            user=user,
+        )
+
+        self.grant_permission(self.reviewer, 'Rating:Modarate')
+        self.reviewer.update(email='reviewer@nonmozilla.com')
+
+        response = self.client.get(self.url)
+        assert response.status_code == 200
+        doc = pq(response.content)
+        assert doc('.addon-rating strong')
+        assert doc('.addon-rating strong').text() == '1 review'
+
 
 class TestAbuseReportsView(ReviewerTest):
     def setUp(self):
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -5632,6 +5632,14 @@ class TestReview(ReviewBase):
         )
         assert doc('.addon-rating a').attr['href'] == rating_url
 
+        self.reviewer.update(email='reviewer@nonmozilla.com')
+
+        response = self.client.get(self.url)
+        assert response.status_code == 200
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/tests/test_views.py
```diff
@@ -6096,6 +6096,25 @@ def test_redirect_after_review_unlisted(self):
...
+        user = user_factory()
+        Rating.objects.create(
+            body='Lorem ipsum dolor',
+            rating=3,
+            ip_address='10.5.6.7',
+            addon=self.addon,
+            user=user,
+        )
+
+        self.grant_permission(self.reviewer, 'Rating:Modarate')
```
eviljeff, line 6109: it's `'Ratings:Moderate'`

src/olympia/reviewers/tests/test_views.py
```diff
@@ -6096,6 +6096,25 @@ def test_redirect_after_review_unlisted(self):
 
         self.assertRedirects(response, self.url)
 
+    def test_review_moderator_addon_rating_read_only(self):
```
eviljeff, line 6099: Lets have this just after the `test_review_moderator_addon_rating_present` test as it's testing the same functionality

eviljeff (changes requested):

src/olympia/reviewers/tests/test_views.py
diox: The fact that tests passed despite this typo is suspicious...

wagnerand: I think we should test with and without a mozilla.com address and specifically check whether the element is a link or not.

eviljeff: I was suspicious too, but in this case both the permission and the email address need to be correct so it was functioning correctly (though doesn't make a great test)

eviljeff: we do test with a mozilla email address in `test_review_moderator_addon_rating_present`.  It might be safer though, to have a single function that sets everything up, then renders it once with a @mozilla address and once with a non mozilla email.

renancleyson-dev: So it's better to merge ```test_review_moderator_addon_rating_present``` and the one I created on only one test, and then verify specifically the link instead of checking for ```assert doc('.addon-rating strong')``` as I did when rendering with the non mozilla email?

eviljeff: Add to the existing `test_review_moderator_addon_rating_present` function - after the existing code change the email address of the user to one that doesn't end in `@mozilla.com`,  continue with `response = self.client.get(self.url)` and so on like you did in `test_review_moderator_addon_rating_read_only`.  That way the rest of the setup is identical, and only the email is different.

src/olympia/reviewers/tests/test_views.py
```diff
@@ -5632,6 +5632,14 @@ def test_review_moderator_addon_rating_present(self):
         )
         assert doc('.addon-rating a').attr['href'] == rating_url
 
+        self.reviewer.update(email='reviewer@nonmozilla.com')
+
+        response = self.client.get(self.url)
+        assert response.status_code == 200
+        doc = pq(response.content)
+        assert doc('.addon-rating strong')
```
eviljeff, line 5640: I just tried this locally and unfortunately this would pass anyway without your changes - there's a `<strong>` inside the link too.  We need something that checks for the `<a..>` really - e.g. `assert not doc('.addon-rating a')`

eviljeff (approved):
thanks for the changes r+

caitmuenster:
Thanks so much, @renancleyson-dev! 🎉 Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 

Hope to see you around the project in the future!

PR #16365, opened 2021-01-21
Split auto approval weight into code related and non-code related

### LINKED ISSUE #16274: Split review weight into code-related and non-code related
We'd like to split the current compound review weight into two values:
* one containing all code-related signals
* one containing all non-code-related signals.

On the review page, let's expose the the two values instead of the single combined one.

In the auto-approval queue, if we can, let's keep the combined weight and also add 2 further columns, each containing one of the new values. If that is complicated to do, we can drop the combined weight from the queue page and just expose the new values in separate columns.

diox:
Rename some fields/columns to go with the new data being displayed

Also simplify the task/command used to recalculate post review weights since we'll need to run it on the current queue, and optimize the weight breakdown we store in the db by filtering before saving (we don't care about storing the keys that don't have an impact on the final weight)

Auto Approved Queue

Fixes #16274


src/olympia/reviewers/models.py
```diff
@@ -1023,10 +1025,25 @@ def calculate_weight(self):
         The weight value is then used in reviewer tools to prioritize add-ons
         in the auto-approved queue, the weight_info shown to reviewers in the
         review page."""
+        non_code_weight_factors = self.calculate_non_code_weight_factors()
+        code_weight_factors = self.calculate_static_analysis_weight_factors()
```
willdurand, line 1029: maybe we should aim for consistency here and therefore use `code_weight` instead of `static_analysis`

src/olympia/reviewers/templates/reviewers/includes/version.html
```diff
@@ -50,13 +50,16 @@
...
+      <div class="maliciousness-score">
+        <strong>{{ _('Maliciousness Score:') }}</strong> {{ version.maliciousness_score|format_score }}
+        <span class="tip tooltip" title="{{ _('The Maliciousness Score represents the likelihood of this version being malicious, based on the scanners findings.') }}">?</span>
       </div>
 
       {% if version.autoapprovalsummary %}
-        <div class="file-weight" title="{{ version.autoapprovalsummary.get_pretty_weight_info()|join('\n') }}"><strong>{{ _('Weight:') }}</strong> {{ version.autoapprovalsummary.weight }}</div>
+        <div class="file-weight" title="{{ version.autoapprovalsummary.get_pretty_weight_info()|join('\n') }}">
+          <strong>{{ _('Code Weight:') }}</strong> {{ version.autoapprovalsummary.code_weight }}<br>
+          <strong>{{ _('Metadata Weight:') }}</strong> {{ version.autoapprovalsummary.non_code_weight }}
```
willdurand, line 61: I am wondering if we shouldn't adopt `metadata` (`_weight`) in the code as well

willdurand (approved):
r+wc

PR #16370, opened 2021-01-21
drop pagination from shelves api

### LINKED ISSUE #16342: Drop pagination, count for the shelf API
Making the https://addons-dev.allizom.org/api/v4/shelves/ API request, it returns page_size, page_cont (with next/previous) - which became irrelevant and it should be dropped. 

 "page_size": 25,
    "page_count": 1,
    "count": 2,
    "next": null,
    "previous": null,

expected:
only return the results.

eviljeff:
fixes #16342 also drops all the /shelves/ api endpoints from /v4 so we don't have to worry about future incompatibilities.


src/olympia/shelves/tests/test_views.py
```diff
@@ -223,9 +213,9 @@ def test_full_response(self):
 
         self.hpshelf_a.update(enabled=True)
```
diox, line 215: Can you also test that pagination stuff isn't in this response ?

diox (approved):

src/olympia/shelves/tests/test_views.py
eviljeff: https://github.com/mozilla/addons-server/pull/16370/commits/06d05bc2ee189cf6bf953fb48aeac65ea2c19003

PR #16377, opened 2021-01-22
Add back logging statement used by foxsec-pipeline, document others

### LINKED ISSUE #16369: Add logging statement at login time back
In e10f985cb2344e0e61140ee32a704c1e0304d7ac we removed the `Got profile {...}` log statement, but it turns out it's actually useful for the foxsec-pipeline, so let's add it back.

diox:
Fixes #16369

### DIFF
--- a/src/olympia/accounts/views.py
+++ b/src/olympia/accounts/views.py
@@ -303,6 +303,8 @@ def with_user(format):
                     request, ERROR_NO_PROFILE, next_path=next_path, format=format
                 )
             else:
+                # The following log statement is used by foxsec-pipeline.
+                log.info('Logging in FxA user %s', identity['email'])
                 user = find_user(identity)
                 # We can't use waffle.flag_is_active() wrapper, because
                 # request.user isn't populated at this point (and we don't want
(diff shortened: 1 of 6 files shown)


src/olympia/accounts/views.py
```diff
@@ -303,6 +303,8 @@ def inner(self, request):
                     request, ERROR_NO_PROFILE, next_path=next_path, format=format
                 )
             else:
+                # The following log statement is used by foxsec-pipeline.
+                log.info('Logging in FxA user %s', identity['email'])
```
diox, line 307: Note: this is a reworded version of what was there before, with only the relevant info needed by foxsec-pipeline.

eviljeff (approved):
Relying on precise strings and comments just still seems fragile to me, but I don't have a better idea so 👍

src/olympia/accounts/views.py
diox: https://github.com/mozilla-services/foxsec-pipeline/issues/510 goes with that change

willdurand: oops sorry about that, good call on adding comments for the future!

Issue #16405, opened 2021-01-27
potential issues with django 3.2


eviljeff:
### Breaking changes/removals

* [ ] ForeignKey.validate() now uses _base_manager rather than _default_manager to check that related instances exist.
  * See comments below for discussion

* [x] The compatibility import of django.core.exceptions.FieldDoesNotExist in django.db.models.fields is removed.

* [x] The compatibility import of django.forms.boundfield.BoundField in django.forms.forms is removed.

* [x] The STATIC_URL and MEDIA_URL settings set to relative paths are now prefixed by the server-provided value of SCRIPT_NAME (or / if not set). This change should not affect settings set to valid URLs or absolute paths.
  * In dev, stage, prod, (and testing) we set these to a fully qualified URL; in settings_base (i.e. local dev) we use an absolute path `/static/` and `/user-media/` so, (unless I'm misintepretting the docs), we arne't affected.

* [x] Meta.ordering and Meta.unique_together options on models in django.contrib modules that were formerly tuples are now lists.
  * I don't think having models declared with tuples will cause a problem unless we have code that explicitly expects, for example, the super class `.ordering` to be a list. 

* [x] Model.save() when providing a default for the primary key. Model.save() no longer attempts to find a row when saving a new Model instance and a default value for the primary key is provided, and always performs a single INSERT query. In older Django versions, Model.save() performed either an INSERT or an UPDATE based on whether or not the row exists.

* [x] urlencode() now encodes iterable values as they are when doseq=False, rather than iterating them, bringing it into line with the standard library urllib.parse.urlencode() function.

### Deprecations (not removals but potentially worth fixing while we're at it)

* [x] The smart_text() and force_text() aliases (since Django 2.0) of smart_str() and force_str() are deprecated

* [x] django.utils.translation.ugettext(), ugettext_lazy(), ugettext_noop(), ungettext(), and ungettext_lazy() are deprecated in favor of the functions that they’re aliases for: django.utils.translation.gettext(), gettext_lazy(), gettext_noop(), ngettext(), and ngettext_lazy().

* [ ] To avoid possible confusion as to effective scope, the private internal utility is_safe_url() is renamed to url_has_allowed_host_and_scheme().
  * https://github.com/mozilla/addons-server/issues/16491

[...]

PR #16407, opened 2021-01-27
replace some obsolete/deprecated functions/classes in django

### LINKED ISSUE #16413: Make the easy/bulk django3.2 changes in the codebase
part of #16405 - make the following easy find&replace-ish changes:

* [x] The compatibility import of django.core.exceptions.FieldDoesNotExist in django.db.models.fields is removed.

* [x] The compatibility import of django.forms.boundfield.BoundField in django.forms.forms is removed.

* [x] The smart_text() and force_text() aliases (since Django 2.0) of smart_str() and force_str() are deprecated

* [x] django.utils.translation.ugettext(), ugettext_lazy(), ugettext_noop(), ungettext(), and ungettext_lazy() are deprecated in favor of the functions that they’re aliases for: django.utils.translation.gettext(), gettext_lazy(), gettext_noop(), ngettext(), and ngettext_lazy().

* [x] Calling django.utils.crypto.get_random_string() without a length argument is deprecated.

* [x] django.conf.urls.url() alias of django.urls.re_path() is deprecated.

eviljeff:
part of #16405, fixes #16413 - just the straightforward find&replace for functions/classes that work equally well in django2.2

### DIFF
--- a/src/olympia/addons/admin.py
+++ b/src/olympia/addons/admin.py
@@ -3,7 +3,6 @@ from urllib.parse import urlencode, urljoin
 
 from django import http, forms
 from django.conf import settings
-from django.conf.urls import url
 from django.contrib import admin
 from django.core import validators
 from django.forms.models import modelformset_factory
@@ -13,10 +12,10 @@ from django.http.response import (
     HttpResponseRedirect,
 )
 from django.shortcuts import get_object_or_404
-from django.urls import resolve
-from django.utils.encoding import force_text
+from django.urls import re_path, resolve
+from django.utils.encoding import force_str
 from django.utils.html import format_html, format_html_join
-from django.utils.translation import ugettext, ugettext_lazy as _
+from django.utils.translation import gettext, gettext_lazy as _
 
 import olympia.core.logger
 from olympia import amo
@@ -298,7 +297,7 @@ class AddonAdmin(admin.ModelAdmin):
 
         urlpatterns = super(AddonAdmin, self).get_urls()
         custom_urlpatterns = [
-            url(
+            re_path(
                 r'^(?P<object_id>.+)/git_extract/$',
                 wrap(self.git_extract_view),
                 name='addons_git_extract',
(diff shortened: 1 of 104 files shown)


src/olympia/addons/admin.py
```diff
@@ -298,7 +297,7 @@ def wrapper(*args, **kwargs):
 
         urlpatterns = super(AddonAdmin, self).get_urls()
         custom_urlpatterns = [
-            url(
+            re_path(
```
willdurand, line 300: This isn't just a replacement, right? `url` and `re_path` don't seem to be equivalent.

willdurand (commented):
Is there anything that we could tell QA to verify once this patch lands? (even if it's not exhaustive?) I'd suggest you file an issue for the functions that have been changed here too.

willdurand:
while reviewing this patch I saw some changes so I submitted what I had in mind so far. let's me know when it's ready :)

src/olympia/addons/admin.py
eviljeff: https://docs.djangoproject.com/en/2.2/ref/urls/#django.conf.urls.url

eviljeff:
> while reviewing this patch I saw some changes so I submitted what I had in mind so far. let's me know when it's ready :)

it's finished, I just forgot to actually push the codestyle fix (duplicate import) before opening the PR

eviljeff:
> 
> 
> Is there anything that we could tell QA to verify once this patch lands? (even if it's not exhaustive?) I'd suggest you file an issue for the functions that have been changed here too.

There isn't anything for QA to test - in this PR it's just bulk renames of functions that were already aliased and/or deprecated so 3.2 will (hopefully) minimally run.  I just checked all the checkboxes in #16405 to indicate what's been addressed but I can copy those into a new issue if you think it's necessary.

willdurand:
> There isn't anything for QA to test - in this PR it's just bulk renames of functions that were already aliased and/or deprecated so 3.2 will (hopefully) minimally run.

Not even a quick sanity check?

> I just checked all the checkboxes in #16405 to indicate what's been addressed but I can copy those into a new issue if you think it's necessary.

Yep, that would be nice. #16405 looks like a meta-issue that can stay opened as long as it is needed.

eviljeff:
> > There isn't anything for QA to test - in this PR it's just bulk renames of functions that were already aliased and/or deprecated so 3.2 will (hopefully) minimally run.
> 
> Not even a quick sanity check?

Nothing specific I can think of, but feel free to suggest something in the issue

> > I just checked all the checkboxes in #16405 to indicate what's been addressed but I can copy those into a new issue if you think it's necessary.
> 
> Yep, that would be nice. #16405 looks like a meta-issue that can stay opened as long as it is needed.

I logged #16413

willdurand (approved):
Thanks!

PR #16421, opened 2021-01-29
Upgrade django-tables2 to 2.x

### LINKED ISSUE #9532: Update django-tables2 to 2.0+
see #9416
django-tables2 2.0 (yay naming) is a breaking change: at the very least as it does a sanity check on `__init__` and realises the model we specify in `Meta` is not actually a django `Model`, but our own `RawSQLModel`.  (Possibly more fixes needed once the code runs).

Chosen fix is "rewrite the review queues to use standard django Model or the standard AMO ModelBase subclass."

diox:
Fixes #9532

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,9 +15,6 @@ updates:
   - dependency-name: Django
     versions:
     - ">= 3.0"
-  - dependency-name: django-tables2
-    versions:
-    - ">= 2"
   - dependency-name: elasticsearch
     versions:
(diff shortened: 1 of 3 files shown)


eviljeff (approved):
Oh, not that many changes needed in the end

diox:
Yeah, the key thing was avoiding the `model =` declaration since it's not a real django model (it was too hard to convince django-tables2 otherwise, and we didn't really need it in the end). Rest is just a bonus.

PR #16424, opened 2021-01-31
Remove yara-read-binary waffle switch

### LINKED ISSUE #16308: Remove yara-read-binary waffle switch
`yara-read-binary` waffle switch should always be on, we should remove that waffle switch and the code that tries to decode file content if the switch is not active.

osamamagdy:
`Fixes #16308`
Here we try to Improve Code Quality in issue https://github.com/mozilla/addons-server/issues/16308
- [x] It is needed to remove the unnecessary check on the `yara-read-binary` waffle switch in ` src/olympia/scanners/tasks.py` 
- [x] Removing unnecessary tests in `src/olympia/scanners/tests/test_tasks.py` as the switch is always on and we don't need to override it in `test_run_in_binary_mode` and also removing the whole `test_run_in_non_binary_mode_with_binary`
- [x] adding a migration to remove waffle switch


diox:
Can you add the migration to delete the switch ?

osamamagdy:
Can you help me with the migration? I didn't use it before and need some documentation about it

diox:
You'll find some additional documentation in https://docs.djangoproject.com/en/2.2/topics/migrations/#data-migrations, but  following the instructions I gave in the issue should be enough:
> add a migration to remove the waffle switch (create an empty migration using `makemigrations --empty`, change the migration to delete the switch like `src/olympia/scanners/migrations/0033_delete_waffle_switch.py` does.

(You can use `docker-compose exec web ./manage.py` to run management commands)

osamamagdy:
I created a new migration like the `src/olympia/scanners/migrations/0033_delete_waffle_switch.py`. I had problems with installing Django so I created the file manually and copied it from `0033_delete_waffle_switch.py` with only changing names but the checks keep failing.

src/olympia/scanners/migrations/0042_remove_waffle_switch.py
```diff
@@ -0,0 +1,15 @@
...
+
+def remove_waffle_switch(apps, schema_editor):
+    Switch = apps.get_model('waffle', 'Switch')
+
+    Switch.objects.filter(name='enable-scanner-results-api').delete()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [('scanners', '0032_auto_20200519_1438')]
```
eviljeff, line 13: this needs to depend on the previous migration (i.e. 41) - that's why all the tests are failing in CI

osamamagdy: Got it. Thank you

osamamagdy: What about now?

src/olympia/scanners/migrations/0042_remove_waffle_switch.py
```diff
@@ -0,0 +1,15 @@
+# Generated by Django 2.2.12 on 2021-02-01 16:11
+from django.db import migrations
+
+
+def remove_waffle_switch(apps, schema_editor):
+    Switch = apps.get_model('waffle', 'Switch')
+
+    Switch.objects.filter(name='enable-scanner-results-api').delete()
```
diox, line 8: This is not the switch we want to remove.

osamamagdy: my bad! the name supposed to be yara-read-binary, right?

diox: Yes

osamamagdy: Modified now

osamamagdy:
Are there any other problems or the code can be merged?
@diox @eviljeff

diox:
Please be patient, we'll review the PR shortly

Issue #16149, opened 2020-12-10
consolidate setup config (and tox.ini) into pyproject.toml


eviljeff:
Since https://www.python.org/dev/peps/pep-0518/ `pyproject.toml` is a standard-ish for specifying python project config.  We still have some tool settings in setup.cfg however.   And some more in tox.ini (tox apparently supports `pyproject.toml` too) It would be better if everything could be specified in one place (I've not investigated if every tool can support `pyproject.toml`, though pytest and tox state they can at least, and that's a big chunk of config).

Bonus if we can get rid of any other random config files!

willdurand:
Looks like flake8 does not support `pyproject.toml` https://gitlab.com/pycqa/flake8/-/issues/428

eviljeff:
> Looks like flake8 does not support `pyproject.toml` https://gitlab.com/pycqa/flake8/-/issues/428

🐼 your regular reminder that 3rd party projects are dictatorships rather than democracies.

I still think it's worth moving as much as possible into pyproject.toml so it's the default source of config.

eviljeff:
isort can use pyproject.toml too

Though it's not clear to me how we actually use isort these days (it's not referenced in the Makefiles or in another config), or if any of the config is actually needed beyond enabling `profile = "black` https://pycqa.github.io/isort/docs/configuration/black_compatibility/

willdurand:
> Though it's not clear to me how we actually use isort these days (it's not referenced in the Makefiles or in another config), or if any of the config is actually needed beyond enabling `profile = "black` [pycqa.github.io/isort/docs/configuration/black_compatibility](https://pycqa.github.io/isort/docs/configuration/black_compatibility/)

I don't think we're actually using it. It was introduced in https://github.com/mozilla/addons-server/commit/57915e248080bafdf5b3a3d3f95b78d64bd0c278, without any tool. Maybe we should remove it then?

PR #16443, opened 2021-02-02
Get rid of tox

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/16149
Fixes https://github.com/mozilla/addons-server/issues/16445


.circleci/config.yml
```diff
@@ -488,55 +497,134 @@ commands:
             - run:
                 name: Wait for memcached
                 command: dockerize -wait tcp://localhost:11211 -timeout 1m
+      - run: pip install --no-deps -r requirements/ci_base.txt
```
diox, line 500: With the removal of `tox`, we shouldn't need `ci_base.txt` anymore. Running `make install_python_test_dependencies` should install `system.txt` and `tests.txt` which should contain everything we need.

willdurand: So, we should move `toml` to `system.txt` I guess?

willdurand: Do we still need `wheels` and `setuptools`?

diox: I would move `wheel` to `system.txt`.

diox: and `toml` is already in `tests.txt` so it should be ok.

willdurand: > I would move `wheel` to `system.txt`.

looks like it's not even needed?

diox: Looks like you're right!

diox (approved):

PR #16453, opened 2021-02-03
make addon count customizable per shelf

### LINKED ISSUE #16397: allow shelf addon count to be customizable
#16367 will limit the addons returned to 4 per shelf (3 for search-themes shelves) but admins may want some shelves to return a different number for some reason.  So there should be an admin setting in the Shelf to override the default to another value.

eviljeff:
fixes #16397


bobsilverberg (approved):
Looks good, thanks for the feature @eviljeff!

PR #16456, opened 2021-02-03
remove isort

willdurand:
fixes https://github.com/mozilla/addons-server/issues/16149


diox:
Maybe we should keep it ? Back in the day, running isort inside tests had nasty consequences because the module was messing with internal stuff or something, I don't remember exactly. So we deactivated it and just ran it manually from time to time. I think it'd go nicely with our `black` strategy if we can find a way to include it in the test suite.

willdurand:
>  I think it'd go nicely with our `black` strategy if we can find a way to include it in the test suite.

I'd rather try to add it correctly then (which would start by removing it first, then filing an issue to investigate isort with/without black and make sure it is automated and verified by CI)

diox:
WFM

bobsilverberg (approved):
LGTM!

Issue #16467, opened 2021-02-04
calculate grouped ratings in addon_rating_aggregates task?


eviljeff:
We currently calculate grouped ratings (counts of ratings with a particular score) on demand and store the result in cache per addon.  If this is something we regularly need it may be a better solution to calculate once per add-on asynchronously in a task and store in the database (and elasticsearch?) like we do with average rating, count, and bayesian average.

diox:
I think that's a good idea. It's a better solution than what we do right now, and it makes sense since we already do that for the averages.

eviljeff:
model options:

1. a JSONField on Addon with a dict for the counts
2. ~a new model, with fk to Addon, containing a single JSONField~
3. a new model, with fk to Addon, containing score_1, score_2, score_3, score_4, score_5
4. ~a new model with fk to Addon, containing score, count - i.e. up to 5 records per addon~

I'm discounting 2 as wasteful for only a single field and 4 as vast overkill for our use-case.  Leaning towards 3 - we could move over `average_rating`, `bayesian_rating`, `total_ratings`, and `text_ratings_count` in a follow-up if we wanted.  1. would be easiest to implement at the cost of an extra field on all addons.

eviljeff:
#16522 went with 3.

AlexandraMoga:
@eviljeff I've verified #16507 and the rating process (submitting, editing, deleting reviews, developer replies). I didn't find issues with any of these and the ratings count.
I've also had a quick look at the Ratings admin model.

Let me know if there is anything else to be verified.

PR #16468, opened 2021-02-04
return grouped ratings with addon detail api

### LINKED ISSUE #16459: Return grouped ratings with addon detail response
The addon detail page calls https://addons.mozilla.org/api/v5/addons/addon/{slug}/ but then has to call https://addons.mozilla.org/api/v5/ratings/rating/?addon={addonId}&show_grouped_ratings=true just for the group rating counts. Given this is a relatively small amount of data we could (optionally) return it with the add-on detail response.

something like:
https://addons.mozilla.org/api/v5/addons/addon/{slug}/?show_grouped_ratings=true

eviljeff:
fixes #16459 - uses the exact same method as the ratings list api does currently.  Filed #16467 to consider an alternative.


bobsilverberg (approved):
Looks good, thanks @eviljeff. Just wanted to confirm: the reason we aren't storing grouped ratings in ES is because they are only needed for the detail API, and not for the search API - is that accurate?

eviljeff:
> Looks good, thanks @eviljeff. Just wanted to confirm: the reason we aren't storing grouped ratings in ES is because they are only needed for the detail API, and not for the search API - is that accurate?

It's more we don't store the grouped ratings anywhere right now - it's an on-demand database query that's cached.  #16467 is going to pre-calculate them like we do with the other addon rating numbers and we could start to index them after that if we wanted to (not sure if there's a particular need?  But it'll be much easier).

PR #16469, opened 2021-02-04
handle django3.2 using fancy quotes; enable devhub job for 3.2

eviljeff:
part of #16405 - a test fix so the devhub job passes on django3.2

### DIFF
--- a/src/olympia/amo/tests/__init__.py
+++ b/src/olympia/amo/tests/__init__.py
@@ -90,9 +90,6 @@ ES_INDEX_SUFFIXES = {key: timestamp_index('') for key in settings.ES_INDEXES.key
 SQUOTE_ESCAPED = escape("'")
 # Unfortunately `escape` returns `&quot;` rather than the code so workaround:
 DQUOTE_ESCAPED = '&#x22;' if 'x' in SQUOTE_ESCAPED else '&#34;'
-# django3.2 uses fancy double quotes in its error strings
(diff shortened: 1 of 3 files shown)


src/olympia/amo/tests/__init__.py
```diff
@@ -90,6 +90,9 @@
 SQUOTE_ESCAPED = escape("'")
 # Unfortunately `escape` returns `&quot;` rather than the code so workaround:
 DQUOTE_ESCAPED = '&#x22;' if 'x' in SQUOTE_ESCAPED else '&#34;'
+# django3.2 uses fancy double quotes in its error strings
+STRING_QUOTE_OPEN = '“' if 'x' in SQUOTE_ESCAPED else "'"
+STRING_QUOTE_CLOSE = '”' if 'x' in SQUOTE_ESCAPED else "'"
```
diox, line 95: IMHO those should use a check on the django version, not on the value of `SQUOTE_ESCAPED` which has nothing to do with it.

diox (approved):
r+wc

src/olympia/amo/tests/__init__.py
eviljeff: yeah I ended up doing that in the subsequent fix prs like #16481

eviljeff: https://github.com/mozilla/addons-server/pull/16469/commits/0c543ce57ea30c08052d879b3122252f6e737b07

PR #16481, opened 2021-02-05
django3.2 fixes for main test job

eviljeff:
part of #16405 - almost all test fixes, apart from the template tag library change affecting users admin and an optional kwarg to a collections inline admin module.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -681,7 +681,7 @@ workflows:
             parameters:
               djangoversion:
                 - django22
-#                - django32 # django3.2 tests still failing
+                - django32
       - reviewers-and-zadmin:
           matrix:
             parameters:
--- a/src/olympia/blocklist/tests/test_admin.py
+++ b/src/olympia/blocklist/tests/test_admin.py
@@ -3,6 +3,7 @@ import json
 
 from unittest import mock
 
+import django
 from django.conf import settings
 from django.contrib.admin.models import LogEntry, ADDITION
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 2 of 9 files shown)


eviljeff:
https://app.circleci.com/pipelines/github/eviljeff/olympia?branch=16405-django32-fixes-main because github doesn't seem to pick up CI for commits that started before the PR was opened.

.circleci/config.yml
```diff
@@ -681,7 +681,7 @@ workflows:
             parameters:
               djangoversion:
                 - django22
-#                - django32 # django3.2 tests still failing
```
bobsilverberg, line 684: Did you mean to remove this?

src/olympia/blocklist/tests/test_admin.py
```diff
@@ -21,6 +22,17 @@
 from ..models import Block, BlocklistSubmission
 
 
+IS_DJANGO_32 = django.VERSION[0] == 3
```
bobsilverberg, line 25: Just wondering if you should actually call this `IS_DJANGO_3`, because that's all you're checking?

src/olympia/blocklist/tests/test_admin.py
```diff
@@ -1245,14 +1258,18 @@ def test_edit_with_blocklist_create(self):
...
+        change_json[0]['changed']['fields'] = [
+            field.lower() for field in change_json[0]['changed']['fields']
+        ]
+        assert change_json == [{'changed': {'fields': ['url', 'reason']}}]
 
         response = self.client.get(multi_url, follow=True)
         assert (
-            b'Changed &quot;Pending: guid@, invalid@, second@invalid; '
-            b'new.url; a new reason thats longer than 40 cha...' in response.content
+            f'Changed {FANCY_QUOTE_OR_ENTITY_OPEN}Pending: guid@, invalid@, '
```
bobsilverberg, line 1270: Because you're using an `in` in this assert, perhaps you could just leave the open quote out of the assertion, as I can see the closed quote is left out? I'm not sure if that compromises the integrity of the test or not, but it would read a bit easier.

bobsilverberg (approved):
Looks good, thanks @eviljeff. I just added a few small comments.

r+wc

.circleci/config.yml
eviljeff: yep! That's what enables the `main` job

src/olympia/blocklist/tests/test_admin.py
eviljeff: It _would_ be more accurate to name it `IS_DJANGO_3`.  But we're only ever going to run this on 3.2 so doesn't really make any difference (we're skipping over 3.0 and 3.1 because they aren't LTS; the next release after 3.2 is 4.0).  Plus we can just drop this workaround once prod is running django3.2 and we drop the 2.2 tests.

src/olympia/blocklist/tests/test_admin.py
eviljeff: We kind of want the `'Changed '` at the start too.

PR #16496, opened 2021-02-08
settings.LANGUAGES to a list; refactor use of AMO_LANGUAGES/LANGUAGE_MAPPING

### LINKED ISSUE #16479: settings.LANGUAGES should always have been a list of tuples, not a dict (django3.2 fix)
`settings.LANGUAGES` is a list* of tuples https://docs.djangoproject.com/en/2.2/ref/settings/#languages and we've been setting it to dict [forever](https://github.com/mozilla/addons-server/commit/723ab7a6ec39ad41b2f7d1c02178cdc4018c65da)  https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/settings_base.py#L185:L186 😐️

In django3.2 there's a system check that errors because the  `(x, y) in LANGUAGES` doesn't work for dicts like it would for list/tuple.

*it was a tuple of tuples back in django1.8 and before, which is fine as lists and tuples are pretty compatible; dict, not so much.

eviljeff:
fixes #16479 - replaces where the dict was used with workarounds using other related settings.

### DIFF
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -1005,6 +1005,7 @@ def review(request, addon, channel=None):
         form=form,
         has_versions_pending_rejection=has_versions_pending_rejection,
         is_admin=is_admin,
+        language_dict=dict(settings.LANGUAGES),
         latest_not_disabled_version=latest_not_disabled_version,
         latest_version_is_unreviewed_and_not_pending_rejection=(
             version
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -873,8 +873,10 @@ class Addon(OnChangeMixin, ModelBase):
         return Rating.objects.filter(addon=self, reply_to=None)
 
     def language_ascii(self):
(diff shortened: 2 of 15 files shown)


eviljeff:
https://app.circleci.com/pipelines/github/eviljeff/olympia?branch=16479-languages-should-be-list 🙄

src/olympia/reviewers/views.py
```diff
@@ -1005,6 +1005,7 @@ def review(request, addon, channel=None):
         form=form,
         has_versions_pending_rejection=has_versions_pending_rejection,
         is_admin=is_admin,
+        language_dict=dict(settings.LANGUAGES),
```
diox, line 1008: Maybe also needed in `AbuseReportAdmin:addon_card()` and `abuse_reports()` views ? Those re-use (directly or indirectly) the template

eviljeff: The use in the template is inside a `{% if name_translations %}` and we only add `name_translations` to the context in the reviewer view.

diox (approved):

Issue #16504, opened 2021-02-09
Add new scanner action that delays indefinitely and restrict the authors


diox:
We already have a scanner rule action that delays indefinitely, we want one that also automatically restrict the associated users by email and IP.

AlexandraMoga:
I've verified this issue on stage by setting a scanner rule action to 'Delay auto-approval indefinitely and add restrictions' and then submitting an add-on that would match that rule. Both the user email and the IP have been added to the Email user restrictions, respectively IP network user restrictions. The user is blocked from submitting further add-ons from the same email address. Trying to submit an add-on from the same IP but with a different user is also not allowed.

If there are multiple scanner rules active, which also have a Delay approval action, the 'Delay auto-approval indefinitely and add restrictions' takes precedence when an add-on matches all those rules.

@diox I've noticed the following behavior for IP restrictions and I don't remember if it is expected or not:
- user logs in from a restricted IP
- user tries to submit an add-on => submission is blocked
- user changes the IP address (using a VPN) but doesn't log out
- user tries to submit an add-on again => submission is blocked

I suppose this happens because the user's Last Known IP address recorded at our end is still the restricted one, even if the user changed it in the meantime. The IP restriction is removed only after the user logs out and logs in again.

diox:
Yes, that's on purpose.

PR #16505, opened 2021-02-09
Add scanner action that adds restrictions on top of delaying auto-approval

diox:
Contains a couple changes that go with that:
- Keep stale add-ons and fileuploads for a little while longer
- Prevent using `FileUpload` belonging to another add-on to create a `Version`
- Link the `FileUpload` to the `Addon` after creating a Version (if none was set)

Fixes https://github.com/mozilla/addons-server/issues/16504

### DIFF
--- a/src/olympia/scanners/actions.py
+++ b/src/olympia/scanners/actions.py
@@ -1,6 +1,7 @@
 from datetime import datetime, timedelta
 
 from olympia.constants.scanners import MAD
+from olympia.users.models import EmailUserRestriction, IPNetworkUserRestriction
 
 
 def _no_action(version):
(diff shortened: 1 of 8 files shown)


src/olympia/scanners/actions.py
```diff
@@ -37,10 +38,37 @@ def _delay_auto_approval_indefinitely(version):
...
+
+    for ip in ips:
+        IPNetworkUserRestriction.objects.get_or_create(network=f'{ip}/32')
+
+
 def _flag_for_human_review_by_scanner(version, scanner):
     from olympia.versions.models import VersionReviewerFlags
 
-    if scanner is not MAD:
+    if scanner != MAD:
```
diox, line 71: Small drive-by, this was wrong and just worked by accident, identity != equality

willdurand (approved):

PR #16522, opened 2021-02-10
replace GroupedRating with RatingAggregate model 

### LINKED ISSUE #16467: calculate grouped ratings in addon_rating_aggregates task?

### LINKED ISSUE #16507: Total ratings count includes ratings with a score of 0 for some old addons
1. Send the following request to the ratings api:
https://addons.mozilla.org/api/v5/ratings/rating/?addon=3006&show_grouped_ratings=true
2. Make a note of the total ratings `count` and then add up the `grouped_ratings`
3. Send the request again, but this time add a `score=0` parameter
4. Compare your results

### What happened?
At the time of this request: total ratings count is 26115; sum of grouped ratings is 26102; number of ratings with `score=0` is 12 (one rating still unaccounted for in this example). This discrepancy between total ratings and sum of grouped ratings is visible on the frontend.

Total ratings and sum of `grouped_ratings` should always match. Ratings with `score = 0` should be excluded.

Another example:
https://addons.mozilla.org/api/v5/ratings/rating/?addon=adblock-plus&show_grouped_ratings=true
Total ratings=13,623; sum of grouped ratings=13612; ratings with `score=0`=11

For these examples, the `score=0` ratings are very old. I don't think this issue will happen with more recent addons/ratings.

eviljeff:
fixes #16467 and fixes #16507 too


src/olympia/addons/migrations/0031_delete_category.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 2.2.18 on 2021-02-09 13:26
```
bobsilverberg, line 1: This doesn't seem to have anything to do with this patch.

bobsilverberg (approved):
Looks good, thanks @eviljeff. Just one comment about what appears to be a stray migration.

r+wc

src/olympia/addons/migrations/0031_delete_category.py
eviljeff: I only generated for the `ratings` app but django wanted to generate this within the `addons` app too (and then made `0002_ratingaggregate.py` depend on it), so I'd either need to manually tinker with them or land it beforehand in a separate PR.  I figured the migration needed generating and committing at _some point_ (after the 2021-02-11 tag) anyway so 🤷 
What do you think?

PR #16531, opened 2021-02-11
return the hero shelves with full localised fields 

### LINKED ISSUE #16515: Return translated fields for the Primary and Secondary hero shelves
The data returned from the [`hero`](https://addons-server.readthedocs.io/en/latest/topics/api/shelves.html#get--api-v5-hero-) endpoint does contain translated data, but it is returned in simple strings, as opposed to most other translated fields which are returned as objects. This should be updated in the API, which will also require a change on the frontend.

eviljeff:
fixes #16515 + refactors some of the field functionality now we're using in 2 different api endpoints (bonus actual tests!)


diox (approved):

docs/topics/api/overview.rst
```diff
@@ -412,7 +412,8 @@ These are `v5` specific changes - `v4` changes apply also.
 * 2021-02-04: dropped /shelves/sponsored endpoint https://github.com/mozilla/addons-server/issues/16390
 * 2021-02-11: removed Stripe webhook endpoint https://github.com/mozilla/addons-server/issues/16391
 * 2021-02-11: added ``show_grouped_ratings`` to addon detail endpoint. https://github.com/mozilla/addons-server/issues/16459
-* 2021-02-12: added ``versions_url`` to addon detail endpoint. https://github.com/mozilla/addons-server/issues/16534
+* 2021-02-18: made ``description``, ``headline``, and ``cta.text`` in shelves/hero endpoint translated fields in the response. (They were always localized, we just didn't return them as such). https://github.com/mozilla/addons-server/issues/16515
+* 2021-02-18: added ``versions_url`` to addon detail endpoint. https://github.com/mozilla/addons-server/issues/16534
```
willdurand, line 416: Why did you change the date?

eviljeff: we usually date the change for the production push as that's when it'll be available for people to use.  (it's not a strict rule, I just changed it because it was the next line)

PR #16553, opened 2021-02-15
Use task user to reject during block

### LINKED ISSUE #16551: Do not expose who blocked an add-on

When an add-on gets blocked, we trigger a rejection for all affected files in order to disable them. 

### What happened?

The reviewer name of this rejection action is exposed in the developer hub.

We should use the "task user" for the rejection.

wagnerand:
Fix #16551.

### DIFF
--- a/src/olympia/blocklist/tests/test_admin.py
+++ b/src/olympia/blocklist/tests/test_admin.py
@@ -1802,7 +1802,9 @@ class TestBlockAdminEdit(TestCase):
         )
         self.change_url = reverse('admin:blocklist_block_change', args=(self.block.pk,))
         self.submission_url = reverse('admin:blocklist_blocklistsubmission_add')
-        self.task_user = user_factory(id=settings.TASK_USER_ID)
+        # We need the task user because some test cases eventually trigger
+        # `disable_addon_for_block()`.
+        user_factory(id=settings.TASK_USER_ID)
 
     def _test_edit(self, user, signoff_state):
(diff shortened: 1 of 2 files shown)


src/olympia/blocklist/tests/test_admin.py
```diff
@@ -1801,6 +1802,7 @@ def setUp(self):
         )
         self.change_url = reverse('admin:blocklist_block_change', args=(self.block.pk,))
         self.submission_url = reverse('admin:blocklist_blocklistsubmission_add')
+        self.task_user = user_factory(id=settings.TASK_USER_ID)
```
willdurand, line 1805: Why is this needed here?

wagnerand: A few tests in that class call code that eventually passes by `disable_addon_for_block`, for example: https://github.com/mozilla/addons-server/blob/72d5bbaba243c88b13bad9e3d2bf410ec8134054/src/olympia/blocklist/tests/test_admin.py#L1895:L1895

willdurand: Interesting. I guess you don't need `self.task_user`?
```suggestion
        # We need the task user because some test cases eventually trigger
        # `disable_addon_for_block()`.
        user_factory(id=settings.TASK_USER_ID)
```

wagnerand: Ok, thanks

willdurand (approved):

PR #16555, opened 2021-02-15
Add file signed status to scanner query result page

### LINKED ISSUE #16554: Expose file signed status on scanner query result page
For quicker triage, it would be helpful if the scanner query result page showed whether the file that matched was signed or not.

wagnerand:
Fix #16554.

### DIFF
--- a/src/olympia/scanners/admin.py
+++ b/src/olympia/scanners/admin.py
@@ -690,6 +690,7 @@ class ScannerQueryResultAdmin(AbstractScannerResultAdminMixin, admin.ModelAdmin)
         'formatted_channel',
         'version_number',
         'formatted_created',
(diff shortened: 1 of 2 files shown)


bobsilverberg (approved):
Looks good to me. I would think we'd want to add a test for this, especially as it is a custom field.

r+wc

wagnerand:
> Looks good to me. I would think we'd want to add a test for this, especially as it is a custom field.
> 
> r+wc

@diox I couldn't find tests for the scanner query result page. Do we have any?

diox:
`TestScannerQueryResultAdmin` ?

wagnerand:
Oh weird, I somehow missed that when searching. Thanks!

PR #16569, opened 2021-02-17
Allow clients to cache non-authenticated API responses for 3 minutes

### LINKED ISSUE #16311: Cache all API non-authenticated responses for 3 minutes
We'd like for consumers of the api to be able to (optionally) return cached results for some of the api endpoints for a short period on the server side for performance.  The collections and addons/search endpoints are the most accessed (because they're shown on the AMO landing page).

diox:
This will also be the duration that nginx will use to cache the responses.

Fixes #16311

### DIFF
--- a/src/olympia/api/middleware.py
+++ b/src/olympia/api/middleware.py
@@ -40,5 +40,5 @@ class APICacheControlMiddleware:
             and get_max_age(response) is None
         )
         if request_conditions and response_conditions:
-            patch_cache_control(response, max_age=3 * 60)
+            patch_cache_control(response, max_age=settings.API_CACHE_DURATION)
         return response
--- a/src/olympia/api/middleware.py
+++ b/src/olympia/api/middleware.py
@@ -1,7 +1,7 @@
 import re
 
 from django.conf import settings
-from django.utils.cache import patch_vary_headers
+from django.utils.cache import get_max_age, patch_cache_control, patch_vary_headers
 from django.utils.deprecation import MiddlewareMixin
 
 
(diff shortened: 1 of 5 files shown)


src/olympia/api/middleware.py
```diff
@@ -19,3 +19,26 @@ def process_response(self, request, response):
...
+    def __call__(self, request):
+        response = self.get_response(request)
+        request_conditions = (
+            request.is_api
+            and request.method in ('GET', 'HEAD')
+            and 'HTTP_AUTHORIZATION' not in request.META
+            and 'disable_caching' not in request.GET
+        )
+        response_conditions = (
+            not response.cookies
```
eviljeff, line 37: do we set cookies for API calls?

src/olympia/api/middleware.py
```diff
@@ -19,3 +19,26 @@ def process_response(self, request, response):
...
+            and response.status_code >= 200
+            and response.status_code < 400
+            and get_max_age(response) is None
+        )
+        if request_conditions and response_conditions:
+            patch_cache_control(response, max_age=3 * 60)
```
eviljeff, line 43: `3 * 60` as a setting (or constant)?

eviljeff (approved):
r+wc

src/olympia/api/middleware.py
diox: Yes, in the authentication flow (and logout, but that's a `DELETE`)

src/olympia/api/middleware.py
```diff
@@ -19,3 +19,26 @@ def process_response(self, request, response):
...
+            request.is_api
+            and request.method in ('GET', 'HEAD')
+            and 'HTTP_AUTHORIZATION' not in request.META
+            and 'disable_caching' not in request.GET
+        )
+        response_conditions = (
+            not response.cookies
+            and response.status_code >= 200
+            and response.status_code < 400
+            and get_max_age(response) is None
```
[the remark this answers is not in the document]
diox: `get_max_age()`  would return a number of seconds if it's set - which can be 0. It will be `None` if it isn't set, which is what we want to check for: we don't want to override if it was explicitly set already by the view.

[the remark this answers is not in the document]
eviljeff: Do we override it anywhere currently?  Or are you just planning for the possibility we might want to in the future?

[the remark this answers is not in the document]
diox: Language Tools and Collection endpoint both set it - they use django's `cache_page` decorator which both caches in memcached _and_ sets `Cache-Control`, with a `max-age` of 24 hours and 1 hour respectively.

src/olympia/api/middleware.py
diox: https://github.com/mozilla/addons-server/pull/16569/commits/b8baceea795b51bc1cf853979ec426c54bcdb2da

Issue #16597, opened 2021-02-22
`app` query parameter should be supported (like `lang`)


willdurand:
The `app` query parameter isn't supported but the frontend passes it as we expect links to be configured with the right `app` (`firefox` or `android`). It isn't a problem in real life because we don't really have any difference between the two supported apps but it actually breaks Fenix installs from AMO because Fenix expects (rightly?) the download URL to contain "android" and we return download URLs with (random) apps.. Given that the frontend passes the right `app` value to the server, we only need to honor this query parameter to fix Fenix and return correct URLs.

In other words, we should probably support `app` in the query string like we do for `lang`.

ioanarusiczki:
@willdurand Making a request with the addon detail API https://addons-dev.allizom.org/api/v5/addons/addon/gmailnoadabc@mywebber.com/?app=android or addons recommendations https://addons-dev.allizom.org/api/v5/addons/recommendations/?guid=gmailnoadabc@mywebber.com&recommended=true&lang=en-US (which I see exposed on the frontend when loading an addon from the android recommended shelf) and I pass the parameter app=android I get the following for the file download url:
"url": "https://addons-dev.allizom.org/android/downloads/file/409056/ublock_origin-2.3.4-an+fx.xpi",

If there's anything else I should be looking for let me know but I think that this should be the expected.

willdurand:
> If there's anything else I should be looking for let me know but I think that this should be the expected.

@ioanarusiczki this issue is kinda related to Fenix so maybe you could try to verify https://github.com/mozilla/addons-frontend/issues/10100

ioanarusiczki:
@willdurand Yes, I was expecting the fix for the frontend issue. I'll try to check it on -dev first.

PR #16598, opened 2021-02-22
Allow urlresolver to set application based on the app query parameter

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/16597

---

This patch does what it is supposed to do, although the original is kinda obscure and not easy to navigate.. I did some manual testing too and it looks good but I dunno, maybe something is missing.

### DIFF
--- a/src/olympia/amo/middleware.py
+++ b/src/olympia/amo/middleware.py
@@ -60,15 +60,16 @@ class LocaleAndAppURLMiddleware(MiddlewareMixin):
         ):
             return redirect_type(request.path.replace('/mobile', '/android'))
 
-        if 'lang' in request.GET and not re.match(
+        if ('lang' in request.GET or 'app' in request.GET) and not re.match(
             settings.SUPPORTED_NONAPPS_NONLOCALES_REGEX, prefixer.shortened_path
         ):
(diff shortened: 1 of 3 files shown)


src/olympia/amo/urlresolvers.py
```diff
@@ -124,6 +124,11 @@ def get_app(self):
         Return a valid application string using the User Agent to guess.  Falls
         back to settings.DEFAULT_APP.
```
diox, line 125: Need to update that comment

src/olympia/amo/middleware.py
```diff
@@ -60,15 +60,16 @@ def process_request(self, request):
         ):
             return redirect_type(request.path.replace('/mobile', '/android'))
 
-        if 'lang' in request.GET and not re.match(
+        if ('lang' in request.GET or 'app' in request.GET) and not re.match(
```
diox, line 63: Do we really need / want this change ? Is it just for parity with how we deal with `?lang` ?

willdurand: I added this change when I found some existing tests for `lang`. Do we need it? No. I thought it would be slightly better to have the same behavior for both `lang` and `app` query params.

diox (approved):

PR #16611, opened 2021-02-23
Remove support for filtering by `platform` in the search/autocomplete API

### LINKED ISSUE #16463: Remove ability to filter by platform from the search/autocomplete APIs
Part of the removal of the `platform` field. https://github.com/mozilla/addons-server/issues/16448

We need to remove ability to filter by platform from the search/autocomplete APIs and decide what to do about backwards-compatibility.

diox:
Also add a migration to ensure all Files have platform=all set now that versions no longer have multiple files in our database.

Fixes #16463


diox:
The `urlparams` failing test is due to https://github.com/mozilla/addons-server/issues/16612

src/olympia/files/migrations/0006_auto_20210223_1215.py
```diff
@@ -0,0 +1,25 @@
...
+from olympia.constants.platforms import PLATFORM_ALL
+
+
+def set_all_files_platform_to_all(apps, schema_editor):
+    File = apps.get_model('files', 'File')
+    # At this point we should no longer have more than one File per Version.
+    # We already set platform=ALL on Files on Versions that did have duplicate
+    # Files, and we've been setting it for new Files for a while now as well,
+    # now we need to do it for all the rest of them.
+    File.objects.exclude(platform=PLATFORM_ALL.id).update(platform=PLATFORM_ALL.id)
```
eviljeff, line 14: to confirm, there's no chance of this failing because there are some rogue versions that still have multiple files in dev or stage? (even deleted addons/versions)

src/olympia/files/migrations/0006_auto_20210223_1215.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 2.2.18 on 2021-02-23 12:15
```
eviljeff, line 1: Any idea how long this will take to execute on prod?

eviljeff (approved):

src/olympia/files/migrations/0006_auto_20210223_1215.py
diox: There was about 14K files to update on prod, but this should be fairly straightforward. I don't expect more than a minute

src/olympia/files/migrations/0006_auto_20210223_1215.py
diox: No there shouldn't be any duplicates left anywhere, and I don't think there was a unique constraint anyway.

PR #16613, opened 2021-02-23
Don't fail on deletion of disabled add-ons with an empty guid

### LINKED ISSUE #16602: Don't try to insert empty guids in DeniedGuid
https://sentry.prod.mozaws.net/operations/olympia-prod/issues/10662093/?query=is%3Aunresolved

Note that there are 2 paths from `delete_addons` task:
- The first one is when `with_deleted` is `True`, we do a hard-delete. In that case, we exclude `guid=None` before doing a `bulk_create` of `DeniedGuid`s. That one should be fine.
- The second one is when we soft-delete. It's just a loop calling `addon.delete()`, which in turn calls `deny_resubmission()`. This method doesn't check that the `guid` is `None`, because it's a little surprising, but I guess we have some weird data floating around...

diox:
Fixes #16602

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -555,6 +555,8 @@ class Addon(OnChangeMixin, ModelBase):
         self.update_status()
 
     def deny_resubmission(self):
+        if not self.guid:
+            raise RuntimeError('No GUID on this add-on')
         if self.is_guid_denied:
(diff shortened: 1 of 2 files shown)


willdurand (approved):
CI fails, though

diox:
Fails are unrelated, fixed since by https://github.com/mozilla/addons-server/pull/16614

PR #16624, opened 2021-02-24
load urlresolvers before sentry_sdk

### LINKED ISSUE #16623: Fix sentry-sdk integration in 0.20.0+
See https://github.com/mozilla/addons-server/pull/16582 for failing tests.

I presume getsentry/sentry-python#992 is causing the test fails as it changes how middleware is patched - it's used for (normal) sync as well as async middleware, despite the description.

eviljeff:
fixes #16623 and closes #16582

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1695,6 +1695,11 @@ SENTRY_CONFIG = {
     'send_default_pii': True,
 }
 
+
+# We need to load this before sentry_sdk.init or our reverse replacement is too late.
+from olympia.amo import urlresolvers  # noqa
+
+
 sentry_sdk.init(
     integrations=[DjangoIntegration(), CeleryIntegration()],
     **SENTRY_CONFIG,
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1697,7 +1697,7 @@ SENTRY_CONFIG = {
 
 
 # We need to load this before sentry_sdk.init or our reverse replacement is too late.
-from olympia.amo import urlresolvers  # noqa
+from olympia.amo import reverse  # noqa
 
 
 sentry_sdk.init(
--- a/conftest.py
+++ b/conftest.py
@@ -129,7 +129,7 @@ def default_prefixer(settings):
     prefixer = amo.urlresolvers.Prefixer(request)
     prefixer.app = settings.DEFAULT_APP
     prefixer.locale = settings.LANGUAGE_CODE
-    amo.urlresolvers.set_url_prefix(prefixer)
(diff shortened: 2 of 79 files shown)


diox:
https://app.circleci.com/pipelines/github/eviljeff/olympia/150/workflows/fe20eeeb-1c79-4807-ab1a-fd89de05b146

src/olympia/lib/settings_base.py
```diff
@@ -1695,6 +1695,11 @@ def get_sentry_release():
     'send_default_pii': True,
 }
 
+
+# We need to load this before sentry_sdk.init or our reverse replacement is too late.
+from olympia.amo import urlresolvers  # noqa
```
eviljeff, line 1700: this could be at the top?  Don't know if that's helpful or not.

diox: I would do the monkeypatching at the top, yes. 

One thing that worries me though is importing something like that in a settings file - typically, settings limit the django-related imports they do to avoid accidentally triggering the initialization of something too early. It's difficult to see what consequences this has, but I feel it's a little bit risky.

eviljeff: it's unfortunate, but we can't really avoid it unless we move the sentry initialization out of settings_base too

eviljeff: we could split olympia.amo.urlresolvers into just `reverse` (and the supporting functions) and the other stuff too, to limit what's imported.

diox: Did we try loading sentry via the `AppConfig` when we integrated the sdk ? https://github.com/mozilla/addons-server/blob/master/src/olympia/core/apps.py

eviljeff: no, sentry-sdk was just swapped in for raven so it was set-up in the same way.  (I don't know if raven supported loading via AppConfig as that was added before my time)

eviljeff: their docs recommend initialization in settings https://docs.sentry.io/platforms/python/guides/django/

diox: Sentry docs and https://github.com/getsentry/sentry-python/issues/408#issuecomment-507667574 do seem to insist in doing sentry init in settings, so we probably want to keep it that way. Could you give the split into 2 different python modules a try ?

eviljeff: https://github.com/mozilla/addons-server/pull/16624/commits/6ab82b0258772cf010135d12f2b6d7f64cfcf969 - this moves the functions (and refactors everywhere we import `reverse` because we were inconsistently importing it from `django.urls` in some places and `amo.urlresolvers` in others)

diox (approved):

PR #16661, opened 2021-03-03
Make it clear reminder email does not relate to new issue

### LINKED ISSUE #11758: Mark reminder email about expiring information request as such

It is not obvious to developers that the reminder email we send one day before information request expiry is indeed a reminder and not a separate issue.

### What happened?

The reminder email looks exactly like the original email (except the remaining number of days to reply).

We should make it clear that the email is a reminder.

dbxnr:
Fixes #11758 - Makes it clear pending rejection email doesn't relate to a new issue


eviljeff (approved):
r+ thanks.

PR #6803, opened 2017-11-01
expose preview image dimensions in API

eviljeff:
fixes #6802

PR #16663, opened 2021-03-03
adjust amo theme thumbnail size; save as jpg instead

### LINKED ISSUE #16639: generate smaller thumbnail for theme previews
We generate a thumbnail for theme previews of 501w x 65h, but addons_frontend* displays the image at 375w x 48h (-ish - there are a few other places that have different -smaller- sizes because we generally scale the image to the space available in the layout).  If we scaled the thumbnail to 48h there would be file size savings. Bonus would be to generate the thumbnail directly from the svg xml template at a smaller scale so UI elements would be cleanly scaled and look better (and also save file size because less antialiasing means better compression)

*(currently we don't know of any api consumer that displays thumbnails for theme previews, after mozilla/addons-frontend#10095 closes addons_frontend will)

eviljeff:
fixes #16639 - also renames some things - referring to: the different renderings (full, list, single), the different image sizes (original, full, thumbnail), and the actual image dimensions (width*height), all as "sizes" started to get very confusing.

### DIFF
--- a/src/olympia/constants/base.py
+++ b/src/olympia/constants/base.py
@@ -226,31 +226,33 @@ VALID_CONTRIBUTION_DOMAINS = (
 # Icon upload sizes
 ADDON_ICON_SIZES = [32, 64, 128]
 
-_size_tuple = namedtuple('SizeTuple', 'width height')
+_dimensions = namedtuple('SizeTuple', 'width height')
 # Preview upload sizes - see mozilla/addons-server#9487 for background.
 ADDON_PREVIEW_SIZES = {
-    'thumb': _size_tuple(640, 480),
-    'min': _size_tuple(1000, 750),
-    'full': _size_tuple(2400, 1800),
+    'thumb': _dimensions(640, 480),
+    'min': _dimensions(1000, 750),
+    'full': _dimensions(2400, 1800),
 }
 
-# Static theme preview sizes
-THEME_PREVIEW_SIZES = {
+# Static theme preview renderings, with different dimensions
+THEME_PREVIEW_RENDERINGS = {
     'header': {
-        'thumbnail': _size_tuple(473, 64),
-        'full': _size_tuple(680, 92),
+        'thumbnail': _dimensions(473, 64),
+        'full': _dimensions(680, 92),
         'position': 0,
+        'thumbnail_format': 'png',
     },
     'list': {
-        'thumbnail': _size_tuple(529, 64),
-        'full': _size_tuple(760, 92),
+        'thumbnail': _dimensions(529, 64),
+        'full': _dimensions(760, 92),
         'position': 1,
+        'thumbnail_format': 'png',
     },
-    # single is planned to be the new default size in 2019 Q1.
     'single': {
-        'thumbnail': _size_tuple(501, 64),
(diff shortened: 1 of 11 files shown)


src/olympia/constants/base.py
```diff
@@ -226,31 +226,33 @@
...
+THEME_PREVIEW_RENDERINGS = {
     'header': {
-        'thumbnail': _size_tuple(473, 64),
-        'full': _size_tuple(680, 92),
+        'thumbnail': _dimensions(473, 64),
+        'full': _dimensions(680, 92),
         'position': 0,
+        'thumbnail_format': 'png',
     },
     'list': {
```
diox, line 245: Where do we still use that one ?

eviljeff: see #16662

src/olympia/versions/tasks.py
```diff
@@ -80,30 +80,37 @@ def generate_static_theme_preview(theme_manifest, version_pk):
...
         svg = tmpl.render(context).encode('utf-8')
         if write_svg_to_png(svg, preview.image_path):
-            resize_image(preview.image_path, preview.thumbnail_path, size['thumbnail'])
+            resize_image(
+                preview.image_path,
+                preview.thumbnail_path,
+                rendering['thumbnail'],
+                format=rendering['thumbnail_format'],
+            )
             pngcrush_image(preview.image_path)
```
diox, line 104: Shouldn't do this for formats other than `png`

eviljeff: we don't - `image_path` is always a png because `write_svg_to_png`

diox: Ah, right. And `resize_image` already avoids calling `pngcrush_image` on non-pngs, so we're good. Carry on!

bobsilverberg (approved):
LGTM. Does this mean that the API response will now include `image_size` and `thumbnail_size`? Is it expected that frontend will do something with those?

eviljeff:
> Does this mean that the API response will now include `image_size` and `thumbnail_size`? Is it expected that frontend will do something with those?

The API has included those fields for 3 years 🙂 since #6803 - frontend already uses them (well, `image_size` at least) in `getPreviewImage` and possible elsewhere.

bobsilverberg:
> > Does this mean that the API response will now include `image_size` and `thumbnail_size`? Is it expected that frontend will do something with those?
> 
> The API has included those fields for 3 years 🙂 since #6803 - frontend already uses them (well, `image_size` at least) in `getPreviewImage` and possible elsewhere.

Oh, well, that's good then. 😄 I noticed they were added to the serializer, so I assumed they were new fields.

eviljeff:
> 
> 
> > > Does this mean that the API response will now include `image_size` and `thumbnail_size`? Is it expected that frontend will do something with those?
> > 
> > 
> > The API has included those fields for 3 years 🙂 since #6803 - frontend already uses them (well, `image_size` at least) in `getPreviewImage` and possible elsewhere.
> 
> Oh, well, that's good then. 😄 I noticed they were added to the serializer, so I assumed they were new fields.

they were already defined in `Meta.fields` so get auto-added to the serializer but I had to explicitly define them as I changed the underlying model property names.

PR #16664, opened 2021-03-03
Create pre-commit hooks

### LINKED ISSUE #14249: Addtion of pre-commit hooks

### What happened?
Integrate pre-commit hooks having standard python lint tools like flake8, isort, black, pep8auto.

This will help us to keep a standard code style and compliance with best python practices.

https://pre-commit.com/

dbxnr:
Fixes #14249

This adds [pre-commit](https://pre-commit.com/) as a dependency to requirements/codestyle.txt and .pre-commit-config.yaml in the root directory.

Pre-commit will run a selection of black, prettier, flake8 and curlylint from within the docker container on specific files being committed, depending on type, and will hold them in staging if they don't pass.

e.g. `docker-compose exec -T web curlylint <changed files>`

Requires a developer to have pre-commit installed in their local environment, and to manually install the hook with `pre-commit install`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,24 @@
 repos:
   - repo: local
     hooks:
-      - id: format
-        name: format with black and prettier
+      - id: black
+        name: format python with black
         language: system
-        entry: docker-compose exec -T web make format
-      - id: lint
-        name: check code formatting
+        entry: docker-compose exec -T web black
+        types: [python]
+      - id: prettier
+        name: format javascript with prettier
         language: system
-        entry: docker-compose exec -T web make lint
-fail_fast: true
+        entry: docker-compose exec -T web $(shell npm $(NPM_ARGS) bin)/prettier
+        args: [--write]
+        types: [javascript]
+      - id: flake8
+        name: lint python with flake8
+        language: system
+        entry: docker-compose exec -T web flake8
+        types: [python]
+      - id: curlylint
+        name: lint html with curlylint
+        language: system
+        entry: docker-compose exec -T web curlylint
+        types: [html]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,13 @@
+repos:
+  - repo: local
+    hooks:
+      - id: format
+        name: format with black and prettier
+        language: system
+        entry: docker-compose exec -T web make format
+      - id: lint
+        name: check code formatting
+        language: system
+        entry: docker-compose exec -T web make lint
+fail_fast: true
+default_stages: [push]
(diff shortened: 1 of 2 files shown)


dbxnr:
That failing test seems to be a CI bug?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - id: flake8
+        name: lint python with flake8
+        language: system
+        entry: docker-compose exec -T web flake8
+        types: [python]
+      - id: curlylint
+        name: lint html with curlylint
+        language: system
+        entry: docker-compose exec -T web curlylint
+        types: [html]
```
eviljeff, line 24: would `docker-compose exec -T web make format` then `docker-compose exec -T web make lint` replace this?  (It'd be consistent)

requirements/codestyle.txt
```diff
@@ -129,3 +129,6 @@ typed-ast==1.4.2 \
     --hash=sha256:9fc0b3cb5d1720e7141d103cf4819aea239f7d136acf9ee4a69b047b7986175a
 flake8-quotes==3.2.0 \
     --hash=sha256:3f1116e985ef437c130431ac92f9b3155f8f652fda7405ac22ffdfd7a9d1055e
+pre-commit==2.10.1 \
+    --hash=sha256:16212d1fde2bed88159287da88ff03796863854b04dc9f838a55979325a3d20e
```
eviljeff, line 133: what is this for, exactly?  The python requirements are installed within the web/worker containers so wouldn't be accessible outside.

eviljeff (commented):
So, a general question - after seeing the patch, does using the pre-commit tool give us anything different, compared to defining this as a native .git pre-commit hook?  The pros of using it are that arbitrary packages and requirements can be installed to do the linting (we don't need that - everything is setup inside the docker container); and it's multi-language (also something we don't need).

requirements/codestyle.txt
dbxnr: An error, will remove.

.pre-commit-config.yaml
dbxnr: The advantage of pre-commit is being is being able to easily share the hooks. It's not possible to add .git/hooks to a repo unfortunately, the workaround is configuring a custom path and locally updating git config.

It is possible to call `make lint` and `make format` although they process the entire repo rather than just the files which are being committed, which is time consuming.

I could add optional args to lint and format in makefile-docker, and the pre-commit hook would pass the changed filenames?

eviljeff:
After discussing this (see issue) we've decided that we don't want to go ahead with this approach because we aren't convinced it will help enough.  Thanks for your effort on this.

PR #16673, opened 2021-03-04
actually install django3.2 when we ask to install django3.2

eviljeff:
we accidentally "upgraded" django3.2b1 to django2.2.19 in #16607


diox:
oops, my bad!

eviljeff:
well, I didn't realise either!

PR #16694, opened 2021-03-08
set jpeg thumbnail quality to 35

### LINKED ISSUE #16692: set jpeg thumbnail quality
In #16639 we were supposed to set the jpeg quality to 35 to get the file size vs. quality at half scale trick to work, but that detail got lost along the way in the slack discussions.  We need to do that.

eviljeff:
fixes #16692

### DIFF
--- a/src/olympia/amo/utils.py
+++ b/src/olympia/amo/utils.py
@@ -666,12 +666,14 @@ def pngcrush_image(src, **kw):
     return False
 
 
-def resize_image(source, destination, size=None, format='png'):
+def resize_image(source, destination, size=None, *, format='png', quality=80):
     """Resizes and image from src, to dst.
     Returns a tuple of new width and height, original width and height.
 
     When dealing with local files it's up to you to ensure that all directories
     exist leading up to the dst filename.
+
+    quality kwarg is only valid for jpeg format - it's ignored for png.
     """
     if source == destination:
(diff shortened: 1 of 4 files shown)


src/olympia/amo/utils.py
```diff
@@ -666,12 +666,14 @@ def pngcrush_image(src, **kw):
     return False
 
 
-def resize_image(source, destination, size=None, format='png'):
+def resize_image(source, destination, size=None, *, format='png', quality=80):
```
championshuttler, line 669: Sorry if i am wrong but why you setting it to 80?

championshuttler, line 669: Sorry if i am wrong but why you setting it to 80? is it going to be ignored for jpg (quality=80)?

eviljeff: 80 is the default value of `quality` if the kwarg isn't provided.  As to "why 80?" - 80 is generally the best compromise between filesize and not losing any noticeable quality. Mainly though, it's been the default we've used for years for jpegs so I didn't want to regress the use of `resize_image` elsewhere.

championshuttler: thank you so much @eviljeff got it, sorry for noise

willdurand (approved):

PR #16695, opened 2021-03-08
Log IP address for each Version creation

### LINKED ISSUE #16636: Store IP address for submissions
At the moment, we store the IP address for submissions temporarily. For security reasons, it should be attached to the submission permanently.

The same clearing rules that apply for other IP address retaining should apply here as well.

This was approved by the Mozilla Legal and Privacy teams.

diox:
Fixes #16636


src/olympia/activity/tests/test_models.py
```diff
@@ -230,6 +231,35 @@ def test_user_log_as_argument(self):
         entries = ActivityLog.objects.for_user(user)
         assert len(entries) == 1
 
+    def test_ip_log(self):
+        addon = Addon.objects.get()
+        assert IPLog.objects.count() == 0
+        # Creating an activity log for an action without store_ip=True doesn't
+        # create an IPLog.
+        assert not getattr(amo.LOG.REJECT_VERSION, 'store_ip', False)
```
bobsilverberg, line 239: Nit: You could put `amo.LOG.REJECT_VERSION` into a variable and use it both here and in the call to `ActivityLog.create` to guarantee they are the same. The same goes for `amo.LOG.ADD_VERSION` below.

src/olympia/activity/tests/test_models.py
```diff
@@ -230,6 +231,35 @@ def test_user_log_as_argument(self):
...
+                amo.LOG.REJECT_VERSION,
+                addon,
+                addon.current_version,
+                user=self.request.user,
+            )
+        assert IPLog.objects.count() == 0
+        # Creating an activity log for an action *with* store_ip=True *does*
+        # create an IPLog.
+        assert getattr(amo.LOG.ADD_VERSION, 'store_ip', False)
+        with core.override_remote_addr('15.16.23.42'):
```
bobsilverberg, line 251: Nit: You could also use a variable for this IP address for the same reason.

src/olympia/devhub/tests/test_feeds.py
```diff
@@ -37,7 +37,8 @@ def log_updates(self, num, version_string='1'):
         version = Version.objects.create(version=version_string, addon=self.addon)
 
         for i in range(num):
-            ActivityLog.create(amo.LOG.ADD_VERSION, self.addon, version)
+            with core.override_remote_addr('127.0.0.1'):
```
bobsilverberg, line 40: Why is this override needed if there is no assertion about the value of the IP address?

src/olympia/users/tests/test_commands.py
```diff
@@ -75,39 +76,54 @@ def test_adds_supergroup(self):
...
 
 class TestClearOldUserData(TestCase):
+    def create_ip_log(self, user):
+        # Note: this is a dummy log that doesn't have store_ip=True on
+        # purpose, to ensure we directly use IPLog when it comes to
+        # deletion - so even if we change our minds about storing ips for
+        # a given action after a while, we'll still correctly clear the old
+        # data when it's time to do so.
+        activity = ActivityLog.create(amo.LOG.CUSTOM_TEXT, 'hi', user=user)
+        IPLog.objects.create(activity_log=activity, ip_address='127.0.0.56')
```
bobsilverberg, line 86: Nit: Making the `ip_address` a const would avoid having to hardcode it in so many places in the tests below.

bobsilverberg (approved):
Looks good, nice work @diox. I just added a few comments about some tests.

r+wc

src/olympia/devhub/tests/test_feeds.py
diox: By default there is no value, but since this creates an `ActivityLog` with `ADD_VERSION` this would trigger the code I've added and fail. The overrides just ensures the call goes through.

src/olympia/users/tests/test_commands.py
diox: Yeah, I wanted to follow the style of the tests I was modifying which does repeat the value.

src/olympia/activity/tests/test_models.py
diox: 76a5f405685625eb007c9521981f1d7483755b2c

src/olympia/activity/tests/test_models.py
diox: 76a5f405685625eb007c9521981f1d7483755b2c

PR #16700, opened 2021-03-09
enhance recreate_theme_previews task to selectively regenerate sizes

### LINKED ISSUE #16691: recreate theme preview thumbnails for existing content
#16639 will start generating jpeg thumbnails (the thumbnail is actually full size, but should be displayed at half size) so we should do something about the existing themes too.  There's possibly a few existing tasks that can be reused/adapted.

eviljeff:
fixes #16691


src/olympia/addons/tasks.py
```diff
@@ -191,23 +193,50 @@ def recreate_theme_previews(addon_ids, **kw):
         '[%s@%s] Recreating previews for themes starting at id: %s...'
         % (len(addon_ids), recreate_theme_previews.rate_limit, addon_ids[0])
     )
-    addons = Addon.objects.filter(pk__in=addon_ids).no_transforms()
+    version_ids = Addon.objects.filter(pk__in=addon_ids).values_list('_current_version')
+    versions = Version.objects.filter(pk__in=version_ids)
```
diox, line 197: Probably want a `no_transforms()` in there still (`Version` does have its own default `transformer` that brings unnecessary stuff)

src/olympia/addons/tests/test_tasks.py
```diff
@@ -58,38 +58,82 @@ def test_recreate_theme_previews():
...
         sizes={'image': [123, 456], 'thumbnail': [34, 45]},
     )
 
-    # addon has 2 complete previews already so skip when only_missing=True
-    with mock.patch('olympia.addons.tasks.generate_static_theme_preview') as p:
+    # addon has all the complete previews already so skip when only_missing=True
+    assert VersionPreview.objects.count() == 3
+    with mock.patch(f'{PATCH_PATH}.generate_static_theme_preview') as p, mock.patch(
+        f'{PATCH_PATH}.resize_image'
+    ) as r:
```
diox, line 95: one letter variables are the devil (even if it was there before :)

src/olympia/addons/tasks.py
```diff
@@ -191,23 +193,50 @@ def recreate_theme_previews(addon_ids, **kw):
...
+        render['full']: {
+            'thumb_size': render['thumbnail'],
+            'thumb_format': render['thumbnail_format'],
+        }
+        for render in amo.THEME_PREVIEW_RENDERINGS.values()
+    }
+
+    for version in versions:
         try:
             if only_missing:
```
diox, line 210: Just so that we're clear: you only need to do the search for which things to recreate when `only_missing` is `True` because in the other case, we recreate _everything_ using `generate_static_theme_preview()` ?

diox (approved):

src/olympia/addons/tasks.py
eviljeff: yes.  It's for when we change the preview svg and need to rebuild everything.  (It's perhaps a use-case we don't need any longer, but I left it there)

src/olympia/addons/tasks.py
eviljeff: if we're doing a full regeneration (`only_missing` is False, or there's one of more render sizes missing) then we do actually use the some of the stuff Version.transformer sets - `.all_files` so it depends.  I could add in `.no_transforms()` but it'd get more inefficient for the full regeneration use-case.

On the other hand, we don't even need the `Version` instance unless we're regenerating the preview (just resizing with `VersionPreview` can be filtered with `version_id`. (Changing _that_ to only load the Version inside the loop would make it even more inefficient for that case though.)

diox: Ok, let's leave it as-is

PR #16723, opened 2021-03-12
Add IP throttling to Ratings and Abuse Reports

### LINKED ISSUE #16722: Add IP throttling to Ratings and Abuse Reports
We already rate-limit ratings and abuse reports per user (which falls back to IP for non-identified requests), we decided in https://bugzilla.mozilla.org/show_bug.cgi?id=1502206 to apply the same limit (1 / minute for ratings, 20 / day for abuse reports) per IP as well.

diox:
Fixes #16722

### DIFF
--- a/src/olympia/abuse/views.py
+++ b/src/olympia/abuse/views.py
@@ -12,7 +12,7 @@ from olympia.addons.views import AddonViewSet
 from olympia.api.throttling import GranularIPRateThrottle, GranularUserRateThrottle
 
 
-class AbuseThrottle(GranularUserRateThrottle):
+class AbuseUserThrottle(GranularUserRateThrottle):
     rate = '20/day'
     scope = 'user_abuse'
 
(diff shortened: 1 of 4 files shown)


src/olympia/abuse/views.py
```diff
@@ -9,18 +9,23 @@
 )
 from olympia.accounts.views import AccountViewSet
 from olympia.addons.views import AddonViewSet
-from olympia.api.throttling import GranularUserRateThrottle
+from olympia.api.throttling import GranularIPRateThrottle, GranularUserRateThrottle
 
 
 class AbuseThrottle(GranularUserRateThrottle):
```
bobsilverberg, line 15: Would it make sense to rename this to `AbuseUserThrottle` to be consistent with the new `AbuseIPThrottle`?

src/olympia/ratings/views.py
```diff
@@ -78,7 +89,7 @@ class RatingViewSet(AddonChildMixin, ModelViewSet):
     ]
     reply_serializer_class = RatingSerializerReply
     flag_permission_classes = [AllowNotOwner]
-    throttle_classes = (RatingThrottle,)
+    throttle_classes = (RatingThrottle, RatingIPThrottle)
```
bobsilverberg, line 92: Similar to my comment above, should we rename `RatingThrottle` to `RatingUserThrottle`?

bobsilverberg (approved):
Looks good, nice work @diox. I just had a couple of questions/suggestions about naming.

r+wc

src/olympia/abuse/views.py
diox: I agree - 93c84173fb548e4be310032a61d51d878a02c0f8

src/olympia/ratings/views.py
diox: 93c84173fb548e4be310032a61d51d878a02c0f8

Issue #16740, opened 2021-03-15
Remove the platform field from the Files model


diox:
Once https://github.com/mozilla/addons-server/issues/16466 is done we can remove the `platform` field from the `Files` model. Then, in a follow-up push, remove the column entirely when closing https://github.com/mozilla/addons-server/issues/16448

diox:
Note: double-check `landfill` code still works after this.

diox:
Maybe it would also be a good occasion to replace any code looping over `<Version>.all_files` or `<Version>.files` with `<Version>.current_file`, since we've removed the ability for a version to have multiple files.

diox:
> Maybe it would also be a good occasion to replace any code looping over `<Version>.all_files` or `<Version>.files` with `<Version>.current_file`, since we've removed the ability for a version to have multiple files.

I ended up not doing that because the patch was already pretty big.

diox:
QA: this should be a transparent change. Please perform a quick sanity check around uploads and review page.

AlexandraMoga:
I've verified add-on and version uploads and I did  not find issues. I've also checked the Reviewer Tools (queues, review pages, logs) and didn't find regressions.

I'm marking this issue as verified fixed on stage.

PR #16741, opened 2021-03-15
Remove platform from API v5 and stop filtering on it in downloads

### LINKED ISSUE #16466: Remove platform field from API v5, shim it in previous versions
Part of the removal of the `platform` field. #16448

We want to remove `platform` entirely in API v5, and shim it as `all` in earlier versions.

diox:
Also prepare the removal of the field on the model by making it nullable and removing the index

Fixes #16466


willdurand (approved):
Looks good, thanks.

I see `platform` entries in some fixture files (`src/olympia/(amo|reviewers)/fixtures/`) as well as some code in `src/olympia/amo/tests/__init__.py`. I am wondering if we want to do some more cleanups here.

diox:
Yeah that'll be in https://github.com/mozilla/addons-server/issues/16740

Issue #16717, opened 2021-03-11
scale down extension preview thumbnails to the correct size


eviljeff:
Currently we create extension preview thumbnails as pngs at 480h*640w (#10069) but frontend only displays them at 200h.  

Recreating them at 400h is a no-brainer for a 10-20% size reduction.  We could also consider:
- dropping support for higher pixel density (retina) displays and sizing to 200h (~75% size reduction)
- saving as jpg in 400h (~75% size reduction at normal image quality.  More reduction at a lower quality)

diox:
We ran some tests comparing the different approaches for the themes previews, I think we need to do the same here with a variety of previews.

eviljeff:
I ran some tests, results here: http://aka-andy.com/extension_preview_comparison.html (page will not be available forever) - tl;dr we think encoding the thumbnails as jpeg, quality 80, is the best compromise.

There are other options we could consider for further improvements:
- selectively choosing the image format based on file size - for original files that are small pngs its unlikely there will be much file size reduction (and sometimes the file size is actually greater) for the quality loss.
- stick with pngs and using additional tools in the pipeline to optimize them, e.g. `optipng` (we already use `pngcrush`)
- use the mozjpeg encoder to save the jpeg (smaller file sizes, and the possibility of fewer image artefacts)

ioanarusiczki:
@eviljeff 

I uploaded a couple of images for https://addons.allizom.org/en-US/firefox/addon/ublock-origin-perms/

- the screenshot viewer displays them at a 200h with a variable weight.
- the thumbnail previews differ a bit from 640x480 ( a classic landscape image is 640x480)
- thumbnails are in .jpeg format at 80 quality. 
- only new uploaded images have a .jpeg preview available.

I think that's all expected - verified on AMO stage with FF86(Win10).

eviljeff:
ugh!  After all the investigation and changes to save as jpeg, I overlooked the "no-brainer" change - making the constant 533x400

eviljeff:
@ioanarusiczki I just landed a fix (and cherry-picked it to stage) - the thumbnails were supposed to be max 533w x 400h

PR #16770, opened 2021-03-17
actually change the thumbnail size to 400h

eviljeff:
followup fix for #16717


eviljeff:
we might want to cherry-pick this if we want to be able to optimize this week

willdurand (approved):

PR #16771, opened 2021-03-17
Add total ADU to blocklist add page

### LINKED ISSUE #16674: Show the total amount of affected users on the blocklist page

When blocking several add-ons in bulk, there is a nice list of add-ons involved in the block, along with usage numbers for each.

### What happened?

I can't easily see how many affected users there are in total.

That page could also show the total number of users.

wagnerand:
Fixes #16674.

### DIFF
--- a/src/olympia/blocklist/tests/test_admin.py
+++ b/src/olympia/blocklist/tests/test_admin.py
@@ -490,6 +490,8 @@ class TestBlocklistSubmissionAdmin(TestCase):
         content = response.content.decode('utf-8')
         # meta data for new blocks and existing ones needing update:
         assert 'Add-on GUIDs (one per line)' not in content
+        total_adu = addon_adu + addon_adu - 1
+        assert '2 Add-on GUIDs with {:,} users:'.format(total_adu) in content
         assert 'any@new' in content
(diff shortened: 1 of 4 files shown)


src/olympia/blocklist/tests/test_admin.py
```diff
@@ -490,6 +490,8 @@ def _test_add_multiple_submit(self, addon_adu):
         content = response.content.decode('utf-8')
         # meta data for new blocks and existing ones needing update:
         assert 'Add-on GUIDs (one per line)' not in content
+        total_adu = addon_adu + addon_adu - 1
```
willdurand, line 493: mm, where does that come from?

wagnerand: That's being defined a few lines above, line 458 and 471

willdurand: I suggest we introduce variables for these then, like `self.new_addon_adu`.

wagnerand: d40eb14dc3

willdurand (approved):

PR #16780, opened 2021-03-18
Remove MYSQL_ALLOW_EMPTY_PASSWORD env var in circle config

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/16781 (when CI is fixed)

---

It's already set in the base image we're using.


eviljeff (approved):
if it's already set upstream, sure.

PR #16782, opened 2021-03-18
Add new flag to disallow *unlisted* versions auto-approval for an add-on

### LINKED ISSUE #16560: Allow unlisted auto-approval to be manually disabled
Reviewers can manually disable listed auto-approval for any add-on in the reviewer tools.

It would be helpful for admins if there was a separate option for disabling unlisted auto-approval per add-on as well.

diox:
Fixes #16560


src/olympia/reviewers/models.py
```diff
@@ -1338,17 +1338,24 @@ def check_is_locked(cls, version):
...
+        - Delayed until a future date by scanners (applies to both listed and
+          unlisted)
         """
         addon = version.addon
         is_listed = version.channel == amo.RELEASE_CHANNEL_LISTED
-        auto_approval_disabled = is_listed and bool(
-            addon.auto_approval_disabled
-            or addon.auto_approval_disabled_until_next_approval
-        )
+        if is_listed:
```
bobsilverberg, line 1352: Can an add-on move from listed to unlisted and vice versa? If so, and an add-on was unlisted, and switches to be listed, would the `auto_approval_disabled_unlisted` no longer be valid? The reason I am asking these questions is because I am wondering if, for this logic, you even need to check `is_listed` at all, or whether it could simply be an `or` between `auto_approval_disabled_unlisted`, `auto_approval_disabled` and `auto_approval_disabled_until_next_approval`?

bobsilverberg: I guess the test below answers my question, but still curious about this.

bobsilverberg (approved):
Looks good, thanks @diox. I asked a question, but I think the test I looked at afterwards basically answers my question.

r+wc

src/olympia/reviewers/models.py
diox: An add-on, by itself, has no channel - it only has versions that do. You can upload listed versions, or unlisted versions, or a mix of both, but can never change the channel for an existing version.

The flags we look at here are set on the add-on, but this is a check for a single version - and it's that version that is either listed or unlisted.

PR #16799, opened 2021-03-22
Remove the platform field from the Files model

### LINKED ISSUE #16740: Remove the platform field from the Files model

diox:
Fixes #16740


eviljeff (approved):
what about constants/platforms.py?  They aren't used any longer, right?  (If it's just that single test that checks platform is ignored, I'd just define that constant - or any random int value - in the test)

diox:
Yeah I agree let's remove `platforms.py` as well. I ignored it initially because it was even included from a migration, but I can just update that as well to hardcode the value.

PR #16801, opened 2021-03-22
Add read-only mode for unlisted reviewer tools pages

### LINKED ISSUE #14580: Implement read-only mode for unlisted reviewer tools pages
In order to limit the amount of users who can make changes to unlisted add-ons through the reviewer tools, we need to add a read-only permission for unlisted reviewer tools pages.

wagnerand:
Fix #14580.

### DIFF
--- a/src/olympia/access/acl.py
+++ b/src/olympia/access/acl.py
@@ -175,6 +175,14 @@ def check_unlisted_addons_reviewer(request):
     return action_allowed(request, amo.permissions.ADDONS_REVIEW_UNLISTED)
 
 
+def check_unlisted_addons_viewer(request):
+    unlisted_perms = [
+        amo.permissions.ADDONS_REVIEW_UNLISTED,
+        amo.permissions.REVIEWER_TOOLS_UNLISTED_VIEW,
+    ]
+    return any(action_allowed(request, perm) for perm in unlisted_perms)
+
+
 def check_static_theme_reviewer(request):
     return action_allowed(request, amo.permissions.STATIC_THEMES_REVIEW)
 
@@ -214,6 +222,11 @@ def is_user_any_kind_of_reviewer(user, allow_viewers=False):
         amo.permissions.STATIC_THEMES_REVIEW,
     ]
     if allow_viewers:
-        permissions.append(amo.permissions.REVIEWER_TOOLS_VIEW)
+        permissions.extend(
+            [
+                amo.permissions.REVIEWER_TOOLS_VIEW,
+                amo.permissions.REVIEWER_TOOLS_UNLISTED_VIEW,
+            ]
+        )
     allow_access = any(action_allowed_user(user, perm) for perm in permissions)
     return allow_access
--- a/src/olympia/access/acl.py
+++ b/src/olympia/access/acl.py
@@ -160,7 +160,7 @@ def check_addon_ownership(request, addon, dev=False, admin=True, ignore_disabled
     return addon.addonuser_set.filter(user=request.user, role__in=roles).exists()
 
 
-def check_addons_reviewer(request, allow_content_reviewers=True):
+def check_listed_addons_reviewer(request, allow_content_reviewers=True):
     permissions = [
         amo.permissions.ADDONS_REVIEW,
         amo.permissions.ADDONS_RECOMMENDED_REVIEW,
@@ -171,16 +171,20 @@ def check_addons_reviewer(request, allow_content_reviewers=True):
     return allow_access
 
 
+def check_listed_addons_viewer_or_reviewer(request, allow_content_reviewers=True):
+    return action_allowed(
+        request, amo.permissions.REVIEWER_TOOLS_VIEW
+    ) or check_listed_addons_reviewer(request, allow_content_reviewers)
+
+
 def check_unlisted_addons_reviewer(request):
     return action_allowed(request, amo.permissions.ADDONS_REVIEW_UNLISTED)
 
 
-def check_unlisted_addons_viewer(request):
-    unlisted_perms = [
-        amo.permissions.ADDONS_REVIEW_UNLISTED,
-        amo.permissions.REVIEWER_TOOLS_UNLISTED_VIEW,
-    ]
-    return any(action_allowed(request, perm) for perm in unlisted_perms)
+def check_unlisted_addons_viewer_or_reviewer(request):
+    return action_allowed(
+        request, amo.permissions.REVIEWER_TOOLS_UNLISTED_VIEW
+    ) or check_unlisted_addons_reviewer(request)
 
 
 def check_static_theme_reviewer(request):
@@ -197,7 +201,7 @@ def is_reviewer(request, addon, allow_content_reviewers=True):
     """
     if addon.type == amo.ADDON_STATICTHEME:
         return check_static_theme_reviewer(request)
-    return check_addons_reviewer(
+    return check_listed_addons_reviewer(
         request, allow_content_reviewers=allow_content_reviewers
     )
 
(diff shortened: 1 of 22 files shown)


src/olympia/access/acl.py
```diff
@@ -175,6 +175,14 @@ def check_unlisted_addons_reviewer(request):
     return action_allowed(request, amo.permissions.ADDONS_REVIEW_UNLISTED)
 
 
+def check_unlisted_addons_viewer(request):
+    unlisted_perms = [
+        amo.permissions.ADDONS_REVIEW_UNLISTED,
```
diox, line 180: While I can see how it makes things easier, I'm not sure I like the idea of having a function named `check_unlisted_addons_viewer` check if  you're a viewer... or more than that. It can be confusing, and it's not consistent with the rest (mainly because in current code we don't have a function that checks if you're a viewer, the check is done in various places, which I agree is not ideal).

src/olympia/addons/decorators.py
```diff
@@ -7,12 +7,12 @@
 from olympia.addons.models import Addon
 
 
-def owner_or_unlisted_reviewer(request, addon):
+def owner_or_unlisted_viewer(request, addon):
```
diox, line 10: As above, the changes you're proposing would subtly change the meaning of this function: with your changes it reads to me as "the user needs to be an owner or an unlisted viewer" when it would actually do "the user needs to be an owner or an unlisted viewer or an unlisted reviewer (with full write access)".

src/olympia/devhub/decorators.py
```diff
@@ -32,11 +31,18 @@ def wrapper(request, addon, *args, **kw):
...
+            ]
+
             if request.method in ('HEAD', 'GET'):
                 # Allow reviewers for read operations.
                 if allow_reviewers_for_read and (
                     acl.is_reviewer(request, addon)
-                    or acl.action_allowed(request, permissions.REVIEWER_TOOLS_VIEW)
+                    or any(
+                        acl.action_allowed(request, perm) for perm in read_permissions
+                    )
```
diox, line 45: This would not be ideal. It would allow read access in devhub to users with the unlisted view permission and not the other one, regardless of what channels the add-on has versions in... This might be ok, but it can be a bit surprising...

diox (commented):
The general idea is fine, but the naming of the functions is going to be confusing. I'm also not a fan of applying that new permission to devhub.

Suggested paths to move forward:
- Find better naming for the `check_unlisted_addons_viewer()` and `owner_or_unlisted_viewer()` functions ?
- Or, do we even need a new permissions, could we get away with re-using `REVIEWER_TOOLS_VIEW` for unlisted as well ? Do we need the extra granularity ?

wagnerand:
> Suggested paths to move forward:
> 
>     * Find better naming for the `check_unlisted_addons_viewer()` and `owner_or_unlisted_viewer()` functions ?

Do you have any suggestions? The idea here was that a reviewer will always be a viewer as well, because read operations are a subset of read-write operations.

> 
>     * Or, do we even need a new permissions, could we get away with re-using `REVIEWER_TOOLS_VIEW` for unlisted as well ? Do we need the extra granularity ?

I wasn't sure we should fold it into `REVIEWER_TOOLS_VIEW`. It would have been inconsistent in many ways as well and we'd lose some flexibility. Can we keep it as a separate permission, also to avoid rewriting the patch again?

diox:
Can you check if there is any code still checking `REVIEWER_TOOLS_VIEW` manually ? If those don't exist or are fixed to use the functions that check both this permission and the regular reviewer permission, then we would be consistent and the naming wouldn't be an issue and we could have a different permission for unlisted.

Devhub access worries me though. Anything we can do about that ?

src/olympia/devhub/decorators.py
wagnerand: Do you know why we have `REVIEWER_TOOLS_VIEW` at all here? What is that needed for?

wagnerand: Do you know why we have `REVIEWER_TOOLS_VIEW` at all here? What is that needed for?

wagnerand: Ah, this is just used to access the file validation page or json, which is part of the devhub. This is controlled by `allow_reviewers_for_read`

diox: Ah, yeah. We were already inconsistent (previously allowing `REVIEWER_TOOLS_VIEW` users access to unlisted) so I suppose this is not worse with your changes, but I'm concerned about the potential for misusing this in the future...

Can you update the docstring? Instead of referencing `allow_reviewers` which is no longer a thing, it should mention what `allow_reviewers_for_read` does, updated with your changes. Hopefully that should be enough.

src/olympia/access/acl.py
wagnerand: I renamed it to `check_unlisted_addons_viewer_or_reviewer` in eeec85729e.

src/olympia/addons/decorators.py
wagnerand: I renamed it to `owner_or_unlisted_viewer_or_reviewer` in eeec85729e.

src/olympia/devhub/decorators.py
wagnerand: I updated the decorator as per the Matrix chat in eeec85729e.

src/olympia/reviewers/decorators.py
```diff
@@ -35,18 +37,19 @@ def wrapper(request, *args, **kw):
...
+    - have the permission 'ReviewerTools:ViewUnlisted' (unlisted viewer)
+    - have the permission 'Addons:ReviewUnlisted' (unlisted reviewer)
     """
 
     @login_required
     @functools.wraps(f)
     def wrapper(request, *args, **kw):
-        if acl.check_unlisted_addons_reviewer(request):
+        view_on_get = _view_on_get(request)
+        if view_on_get or acl.check_unlisted_addons_reviewer(request):
```
diox, line 52: This would allow someone with just `REVIEWER_TOOLS_VIEW` to view a page decorated by `unlisted_addons_reviewer_required` since you modified `_view_on_get` to allow this. I think you shouldn't change `_view_on_get` and do something specific here for `REVIEWER_TOOLS_UNLISTED_VIEW`.

wagnerand: Fixed in 200de0dddb403472977c3edb75bed9e91730f791.

diox:
`AttributeError: module 'olympia.access.acl' has no attribute 'check_unlisted_addons_viewer'`

wagnerand:
> `AttributeError: module 'olympia.access.acl' has no attribute 'check_unlisted_addons_viewer'`

Weird, I don't remember changing [`200de0d` (#16801)](https://github.com/mozilla/addons-server/pull/16801/commits/200de0dddb403472977c3edb75bed9e91730f791#diff-618357ff4449192d1596d4f7435a86639740f00d6b42c1e937fddfda2ec5a8c6). Maybe a search-and-replace mistake.

src/olympia/addons/decorators.py
```diff
@@ -7,12 +7,12 @@
...
 
-def owner_or_unlisted_reviewer(request, addon):
+def owner_or_unlisted_viewer_or_reviewer(request, addon):
     return (
-        acl.check_unlisted_addons_reviewer(request)
+        acl.check_unlisted_addons_viewer_or_reviewer(request)
         # We don't want "admins" here, because it includes anyone with the
         # "Addons:Edit" perm, we only want those with
-        # "Addons:ReviewUnlisted" perm (which is checked above).
+        # "ReviewerTools:ViewUnlisted" perm (which is checked above).
```
diox, line 15: Both permissions work

diox (approved):

src/olympia/addons/decorators.py
wagnerand: 0d194078e5

PR #16813, opened 2021-03-24
Stop displaying status changes in review page

### LINKED ISSUE #16808: Filter out auto-approvals in "Add-on important changes history" section

The "Add-on important changes history" on the reviewer tools page contains an entry for each auto-approval, which can cause lots of entries for add-ons with many versions. Let's hide them.

### What happened?

Many auto-approval entries shown like: "... status changed to Approved."

Auto-approvals should probably be hidden.

diox:
We currently record an `ActivityLog` everytime we go through `update_status()` roughly, even when the status doesn't actually change (like going from 'Approved' to... 'Approved'). So displaying them all while this is happening is not a good idea.

This also attempts to make the tests around query counts more stable for the future if we ever re-implement displaying the status changes, by forcing those tests to either do more activity log related stuff (`test_item_history_pagination()`) or ignore them completely (`test_versions_that_are_flagged_by_scanners_are_highlighted()`)

Fixes #16808

### DIFF
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -3576,8 +3576,18 @@ class TestReview(ReviewBase):
             version_factory(
                 addon=addon, version=f'1.{i}', created=self.days_ago(365 - i)
             )
-        with self.assertNumQueries(62):
-            # FIXME: 62 is obviously still too high, but it's a starting point.
+        # Since we're testing queries, also include an author change that will
+        # be displayed in the "important changes" log.
+        author = self.addon.addonuser_set.get()
+        core.set_user(author.user)
+        ActivityLog.create(
+            amo.LOG.ADD_USER_WITH_ROLE,
+            author.user,
+            str(author.get_role_display()),
+            self.addon,
+        )
+        with self.assertNumQueries(63):
+            # FIXME: obviously too high, but it's a starting point.
             # Potential further optimizations:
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/tests/test_views.py
```diff
@@ -5326,29 +5336,24 @@ def test_important_changes_log(self):
...
-        # normally be getting for an approved add-on, because it forces the
-        # status - that's fine, in this case we only want to make sure the
-        # change status ones are included in the log.
-        assert ' status changed to Incomplete' in important_changes[0].text_content()
-        assert 'class' not in important_changes[1].attrib
-        assert '(Owner) added to ' in important_changes[1].text_content()
+        assert len(important_changes) == 5
+        assert 'class' not in important_changes[0].attrib
+        assert '(Owner) added to ' in important_changes[0].text_content()
+        assert 'class' not in important_changes[0].attrib
```
bobsilverberg, line 5342: This seems to be the exact same statement as on line 5340.

bobsilverberg (approved):
Looks good, thanks @diox. I think there's a minor copy/paste error in one of the tests.

r+wc

src/olympia/reviewers/tests/test_views.py
diox: Good catch, fixed in bcf464c3cb5c1667ca74a808b92099a52bfbc213

Issue #16800, opened 2021-03-22
Delete versions without files


wagnerand:
There are a few versions in the database that don't have any files, so we should just hard-delete them.

```
select count(v.id) from versions v left join files f on f.version_id = v.id where f.id is null;
+-------------+
| count(v.id) |
+-------------+
|        4748 |
+-------------+
```

wagnerand:
We should be able to do this in MySQL via the following transaction:

```
START TRANSACTION;

DELETE FROM
    applications_versions
WHERE
    version_id IN (
        SELECT
            v.id
        FROM
            versions v
            LEFT JOIN files f ON f.version_id = v.id
        WHERE
            f.id IS NULL
    );

DELETE FROM
    editors_autoapprovalsummary
WHERE
    version_id IN (
        SELECT
            v.id
        FROM
            versions v
            LEFT JOIN files f ON f.version_id = v.id
        WHERE
            f.id IS NULL
    );

DELETE FROM
    log_activity_comment_draft
WHERE
    version_id IN (
        SELECT
            v.id
        FROM
            versions v
            LEFT JOIN files f ON f.version_id = v.id
        WHERE
            f.id IS NULL
    );

DELETE FROM
    log_activity_version
WHERE
    version_id IN (
        SELECT
            v.id
        FROM
            versions v
            LEFT JOIN files f ON f.version_id = v.id
        WHERE
            f.id IS NULL
    );

DELETE FROM
    promoted_promotedapproval
WHERE
    version_id IN (
        SELECT
            v.id
        FROM
            versions v
            LEFT JOIN files f ON f.version_id = v.id
        WHERE
            f.id IS NULL
    );

DELETE FROM
    reviewer_scores
WHERE
    version_id IN (
        SELECT
            v.id
        FROM
            versions v
            LEFT JOIN files f ON f.version_id = v.id
        WHERE
            f.id IS NULL
    );

DELETE FROM
    reviews_moderation_flags
WHERE
    review_id IN (
        SELECT
            id
        FROM
            reviews
        WHERE
            version_id IN (
                SELECT
                    v.id
                FROM
                    versions v
                    LEFT JOIN files f ON f.version_id = v.id
                WHERE
                    f.id IS NULL
            )
    );

UPDATE
    reviews
SET
    reply_to = NULL
WHERE
    reply_to IN (
        SELECT
            id
        FROM
            (
                SELECT
                    id
                FROM
                    reviews
                WHERE
                    version_id IN (
[...]

PR #16818, opened 2021-03-25
Add unique constraint for add-on version string

### LINKED ISSUE #14332: Resolve duplicate versions for an add-on to allow the version string to be unique per add-on
To date, there are about 2800 submissions with a duplicate version string (the version number). 16 of these affected webextensions [1].

We should resolve these duplicates so we can make the version string unique (per add-on).

[1] Query for reference:
```
select
  v.addon_id, v.version, count(v.id)
  from versions v
  where v.id in (
    select f.version_id
    from files f
    where f.is_signed = 1 and f.is_webextension = 1)
  group by v.addon_id, v.version having count(v.id) > 1;
```

### LINKED ISSUE #16901: Add unique database constraint to prevent duplicate versions for a given add-on
`version` string should be unique per add-on. We enforce that in code, but we should have a database constraint, when the [task to remove existing duplicates](https://github.com/mozilla/addons-server/issues/14332) has been run in prod.

wagnerand:
Fix #16901.

This depends on some preparation work (see issue), so it can't be merged right now.


src/olympia/versions/models.py
```diff
@@ -215,6 +215,9 @@ class Meta(ModelBase.Meta):
             models.Index(fields=('addon',), name='addon_id'),
             models.Index(fields=('license',), name='license_id'),
         ]
+        constraints = [
+            models.UniqueConstraint(fields=('addon', 'version'), name='version_id'),
```
eviljeff, line 219: I know this is WIP, but can we have a more descriptive name for this constraint? - it's got nothing to do with a version_id field. 

(the indexes above are also poorly named, but they are least match the column's name - i.e the index for `addon` is actually for `addon_id` in the database table, so the name matches)

wagnerand: What do you suggest? I was looking at several other constraints defined in the code and they had similar or even "worse" naming (in terms of meaningfulness), so I went with something that aligns with what's being used in other places.

diox: Note that the name needs to be unique. I would suggest trying to follow the name `unique_together` would generate, which is `versions_addon_id_version_5a2e75b6_uniq`.

wagnerand: dc9177812d

diox (approved):
Should not be merged before https://github.com/mozilla/addons-server/issues/16800 is fired in all environments (so it will likely land in **2021.04.15** and not before)

wagnerand:
We decided not to create a new task in #14332 and instead run it manually on Monday, so this can and shall land for 2021.04.15.

diox:
We've added the fix command to [this week's push notes](https://github.com/mozilla/addons/blob/main/releases/2021/04/15.md) in the "before we start" section and notified ops about it.

PR #16821, opened 2021-03-25
Add command to hard-delete versions without files

### LINKED ISSUE #16800: Delete versions without files

diox:
Also includes a commit that refactors `process_addons` command to allow us to create other `process_xxx` commands for other types of objects while sharing most of the logic

Fixes #16800


src/olympia/amo/management/__init__.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class ProcessObjectsCommand(BaseCommand):
+    """
+    A generic command base class to run a task on objects.
+
+    Inherit from it and implement `get_model()` and `get_tasks()`.
+    """
+
+    def get_model(self):
```
eviljeff, line 15: nit, but I'd just had this as a property in the subclass declaration as it doesn't need to be dynamically determined.

eviljeff (approved):

src/olympia/amo/management/__init__.py
diox: I originally had this as a property but I wanted to be able to raise `NotImplementedError` to get an explicit exception if someone tried to use that class directly...

PR #16831, opened 2021-03-26
theme previews can be generated as svgs

### LINKED ISSUE #16755: investigate serving full size theme previews in jpg (or other optimizations)
addons-frontend displays the full size theme preview in search results, so will display up to 25 of them per page.  Jpegs are generally smaller so would result in a significant download saving for the user.  Drawback - potentially ugly artefacts that could put users off installing the theme, because they think that's what their Firefox would look like too.

In addition to just saving as jpeg we have some other options to consider (copied over from #16717):
* selectively choosing the image format based on file size - for original files that are small pngs its unlikely there will be much file size reduction (and sometimes the file size is actually greater) for the quality loss.
* stick with pngs and using additional tools in the pipeline to optimize them, e.g. optipng (we already use pngcrush)
* use the mozjpeg encoder to save the jpeg (smaller file sizes, and the possibility of fewer image artefacts)

eviljeff:
fixes #16755 - for new uploads at least.  I'll look into migrating existing previews after, though I think `process_addons --task=recreate_theme_previews` will work as is (`task=create_missing_theme_previews` wouldn't)

### DIFF
--- a/src/olympia/addons/management/commands/process_addons.py
+++ b/src/olympia/addons/management/commands/process_addons.py
@@ -93,7 +93,7 @@ tasks = {
         'qs': [
             Q(
                 type=amo.ADDON_STATICTHEME,
-                status__in=[amo.STATUS_APPROVED, amo.STATUS_AWAITING_REVIEW],
+                status__in=[amo.STATUS_APPROVED, amo.STATUS_NOMINATED],
             )
         ],
         'kwargs': {'only_missing': False},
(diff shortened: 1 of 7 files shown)


bobsilverberg (approved):
Looks good to me, nice work @eviljeff. I can't say I exactly follow everything that is going on, but nothing in the code or tests jumps out at me as needing improvement.

PR #16852, opened 2021-03-31
convert background image to jpg directly in render_to_svg

### LINKED ISSUE #16755: investigate serving full size theme previews in jpg (or other optimizations)

eviljeff:
fixes #16755 (again)

- it seems that when you pass a `BytesIO` into `Image.save` it's then closed once the `Image` instance is closed/discarded, which clears the buffer (or it might have been the inner `open` - I didn't debug further once I realised what was happening).  Fix was to do the image resize without using `resize_image` (which is written to work with file paths rather File instances).
- all the tests passed before because the code is tolerant enough to deal with broken images, and we didn't make any attempt to validate the background image in the final svg (now we do)


src/olympia/amo/utils.py
```diff
@@ -673,8 +673,6 @@ def resize_image(source, destination, size=None, *, format='png', quality=80):
     When dealing with local files it's up to you to ensure that all directories
     exist leading up to the destination filename.
 
-    destination can be a string filename, or a file object.
```
willdurand, line 676: Maybe add that it should only be a filename.

willdurand (approved):
r+wc

src/olympia/amo/utils.py
eviljeff: (merged so it's there for QA to test - I'll address comments in a follow-up commit tomorrow)

eviljeff: so, this comment was only added in the previous PR #16831 and it's still kind of true in some cases - a `file` object (file-like-object) will work, but only a real file object that you don't care about being closed and where closing the file-like-object doesn't discard the data, like with `BytesIO`.  And only for jpeg format (`pngcrush_image` will fail for png files).  

Basically I realised my earlier comment was naive, and only applied for certain cases, and I either needed a) to write a longer explanation, and add some guard rails (`assert`, etc) or use different functions and contexts depending the type of the input, or b) just remove the comment so if someone was to try to reuse `resize_image` again in the future they would have to read and understand the code rather than rely on the comments.  We always use a filepath in the other uses anyway so it'd be a lot of work for an unused code path.

PR #16893, opened 2021-04-08
Optimize & reduce number of queries made by Developer Hub Submissions Dashboard

### LINKED ISSUE #16772: Developer dashboard page is slow for most users since the switch to MySQL 8.0.x
We've not tried to particularly optimize https://addons.mozilla.org/developers/addons (it was never fast, but it was fast enough for it's purpose) but recently (from around January 25th, according to stats) has become very slow (~15 seconds).  

We should investigate if any code changes caused the slowdown (possible, but pushes were on January 21st and February 4th) or something else in the infrastructure.

diox:
**Query optimizations:**
- Make query to fetch `ActivityLogs` for add-ons the user is an author of faster  by fetching the add-ons first to avoid the double subquery that MySQL 8.0.17+  has trouble with

**Queries reductions:**
- Attach the `Addon` instance to other models that might need it when building  `ActivityLog` arguments
- Avoid `Addon` transformers where not needed
- Avoid fetching the `License` just to figure out if the add-on has complete metadata

~~**Bonus drive-by queries reduction:** avoid fetching the `Addon` to link to a custom `License`~~

Fixes #16772

### DIFF
--- a/src/olympia/devhub/templates/devhub/includes/addon_details.html
+++ b/src/olympia/devhub/templates/devhub/includes/addon_details.html
@@ -67,7 +67,7 @@
     {% if addon.current_version %}
       <li>
         <strong>{{ _('Listed Version:') }}</strong>
-        <a href="{{ addon.versions_url }}">{{ addon.current_version.version }}</a>
+        <a href="{{ addon.current_version.get_url_path() }}">{{ addon.current_version.version }}</a>
         <span class="tip tooltip" title="{{ _('This is the version of your add-on that will '
                                               'be installed if someone clicks the Install '
                                               'button on addons.mozilla.org') }}">?</span>
--- a/src/olympia/devhub/templates/devhub/includes/addon_details.html
+++ b/src/olympia/devhub/templates/devhub/includes/addon_details.html
@@ -67,7 +67,7 @@
     {% if addon.current_version %}
       <li>
         <strong>{{ _('Listed Version:') }}</strong>
-        <a href="{{ addon.current_version.get_url_path() }}">{{ addon.current_version.version }}</a>
+        <a href="{{ addon.versions_url }}">{{ addon.current_version.version }}</a>
         <span class="tip tooltip" title="{{ _('This is the version of your add-on that will '
                                               'be installed if someone clicks the Install '
                                               'button on addons.mozilla.org') }}">?</span>
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -123,7 +123,7 @@ def addon_listing(request, theme=False):
 def index(request):
     ctx = {}
     if request.user.is_authenticated:
-        recent_addons = request.user.addons.order_by('-modified')[:3]
+        recent_addons = request.user.addons.all().order_by('-modified')[:3]
         ctx['recent_addons'] = []
         for addon in recent_addons:
             ctx['recent_addons'].append(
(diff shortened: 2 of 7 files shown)


src/olympia/devhub/templates/devhub/includes/addon_details.html
```diff
@@ -67,7 +67,7 @@
     {% if addon.current_version %}
       <li>
         <strong>{{ _('Listed Version:') }}</strong>
-        <a href="{{ addon.current_version.get_url_path() }}">{{ addon.current_version.version }}</a>
+        <a href="{{ addon.versions_url }}">{{ addon.current_version.version }}</a>
```
eviljeff, line 70: this is just a drive-by change, right?  (it's obviously not saving accessing `current_version`)

src/olympia/versions/models.py
```diff
@@ -399,7 +399,7 @@ def from_upload(cls, upload, addon, selected_apps, channel, parsed_data=None):
         return version
 
     def license_url(self, impala=False):
-        return reverse('addons.license', args=[self.addon.slug, self.version])
+        return reverse('addons.license', args=[self.addon_id, self.version])
```
eviljeff, line 402: Not sure about this - we expose this in the API response so we're now returning non-canonical frontend urls

src/olympia/devhub/views.py
```diff
@@ -254,6 +254,15 @@ def _get_activities(request, action):
...
 def _get_items(action, addons):
+    if not isinstance(addons, (list, tuple)):
+        # MySQL 8.0.21 (and maybe higher) doesn't optimize the join with
+        # double # subquery the ActivityLog.objects.for_addons(addons) below
+        # would generate if addons is not transformed into a list first. Since
+        # some people have a lot of add-ons, we only take the last 100.
+        addons = list(
+            addons.all().order_by('-modified').values_list('pk', flat=True)[:100]
+        )
+
```
eviljeff, line 265: I read this as the same scenario we were talking about on chat (that I thought you'd decided against):

>  So it wouldn't really be recent activity on the addons, it'd be activity on addons that have recently been updated.

expect that now rather being add-ons that have been updated in the past 6 months, it's the last 100 most recently modified add-ons (which for at least one prolific user is likely a lot less than 6 months).

If this is really a necessary restriction I think we need to call it out somewhere in the copy on the page.

src/olympia/devhub/templates/devhub/includes/addon_details.html
diox: The issue I was trying to solve is that `<Version>.get_url_path()` is accessing the `addon`, which may not have been loaded already, so I was hunting for all calls to `<Version>.get_url_path()` to figure out where the problematic ones were coming from. But that was before I ended up setting `addon` in arguments loaded by `ActivityLog`, and this one in particular should be safe, since `current_version` should be coming from the `Addon` transformer. I'll try removing it and see if that changes anything.

src/olympia/devhub/views.py
diox: For 99% users there will be no change in behavior since they have less than 100 add-ons. For the most prolific ones I doubt the recent changes list was useful to them anyway (too many add-ons to care) so I don't think they'll notice, and I don't think we should draw attention to it on the page.

src/olympia/devhub/templates/devhub/includes/addon_details.html
eviljeff: perhaps out of scope for this patch, but `Version.get_url_path` should probably just call `Addon.get_url_path` anyway, rather than replicate the `reverse`

src/olympia/versions/models.py
diox: reverted in f896d03c704fefc4abb1cf471fdf7955a2a3e4fe, it's not necessary.

src/olympia/devhub/templates/devhub/includes/addon_details.html
diox: reverted in f896d03c704fefc4abb1cf471fdf7955a2a3e4fe, it's not necessary.

eviljeff (approved):

PR #16900, opened 2021-04-09
Fix repack_fileupload() to clean up temporary directories once it's done

### LINKED ISSUE #16879: Ensure devhub worker tasks are cleaning up local temporary files
We had an issue where `/` filed up in production on devhub celery workers. 

`repack_fileupload` is the most likely suspect as it writes to the local tempdir (normally `/tmp`) which isn't on a separate filesystem. It _should_ cleanup after itself, but maybe there is an issue with that. Or maybe it's a completely different task that's the problem and we just happened to notice it because uploads were broken and we were left with errors about repack.

We should verify which task is actually responsible and fix it.

diox:
~~Also remove unnecessary timers~~

Fixes #16879

### DIFF
--- a/src/olympia/files/tasks.py
+++ b/src/olympia/files/tasks.py
@@ -7,7 +7,6 @@ import olympia.core.logger
 from olympia.amo.celery import task
 from olympia.amo.decorators import use_primary_db
 from olympia.amo.storage_utils import move_stored_file
-from olympia.amo.utils import StopWatch
 from olympia.devhub.tasks import validation_task
 from olympia.files.models import File, FileUpload
 from olympia.files.utils import extract_zip, get_sha256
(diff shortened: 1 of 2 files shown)


src/olympia/files/tasks.py
```diff
@@ -26,32 +25,30 @@ def repack_fileupload(results, upload_pk):
...
-                'Could not extract upload %s for repack.', upload_pk, exc_info=exc
-            )
-            raise
-        timer.log_interval('1.extracted')
-        log.info('Zip from upload %s extracted, repackaging', upload_pk)
-        file_ = tempfile.NamedTemporaryFile(suffix='.zip', delete=False)
-        shutil.make_archive(os.path.splitext(file_.name)[0], 'zip', tempdir)
+        # tempdir must *not* be on TMP_PATH, we want local fs instead. It will be
+        # deleted automatically once we exit the context manager.
+        with tempfile.TemporaryDirectory(prefix='repack_fileupload_extract') as tempdir:
```
diox, line 30: Fancy Python 3.2+ feature I didn't know about! https://docs.python.org/3/library/tempfile.html?highlight=tempfile#tempfile.TemporaryDirectory

src/olympia/files/tasks.py
```diff
@@ -26,32 +25,30 @@ def repack_fileupload(results, upload_pk):
...
+            file_ = tempfile.NamedTemporaryFile(suffix='.zip', delete=False)
+            shutil.make_archive(os.path.splitext(file_.name)[0], 'zip', tempdir)
         with open(file_.name, 'rb') as f:
             upload.hash = 'sha256:%s' % get_sha256(f)
-        timer.log_interval('2.repackaged')
         log.info('Zip from upload %s repackaged, moving file back', upload_pk)
         move_stored_file(file_.name, upload.path)
-        timer.log_interval('3.moved')
         upload.save()
-        timer.log_interval('4.end')
```
willdurand, line 54: uh, one of my dashboards rely on these metrics...

diox: Ah, it didn't seem very useful, but I can add it back.

diox: https://github.com/mozilla/addons-server/pull/16900/commits/1459bb6ff5c05082f2440577889180326c62deea

bobsilverberg (approved):

PR #16903, opened 2021-04-09
Ignore django.security.DisallowedHost in Sentry

### LINKED ISSUE #16902: Stop forwarding django.security.DisallowedHost errors to sentry
Non-actionable `django.security.DisallowedHost` errors are polluting sentry.

Quoting Mat:

> we can filter out django.security.DisallowedHost by specifying a null handler in the config (or downgrading it to a warning so that it shows in logs still ?)

Which option do we prefer?

wagnerand:
Fix #16902.

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1216,6 +1216,12 @@ LOGGING = {
             'level': logging.WARNING,
             'propagate': True,
         },
+        # Downgrade DisallowedHost so it doesn't show up in Sentry
+        'django.security.DisallowedHost': {
+            'handlers': ['mozlog'],
+            'level': logging.WARNING,
+            'propagate': False,
+        },
         'elasticsearch': {
             'handlers': ['null'],
             'level': logging.INFO,
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -9,7 +9,7 @@ import socket
 
 from datetime import datetime
 
-from django.security import DisallowedHost
+from django.core.exceptions import DisallowedHost
 
 import sentry_sdk
 from kombu import Queue
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -9,12 +9,11 @@ import socket
 
 from datetime import datetime
 
-from django.core.exceptions import DisallowedHost
-
 import sentry_sdk
 from kombu import Queue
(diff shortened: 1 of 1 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -1216,6 +1216,12 @@ def get_db_config(environ_var, atomic_requests=True):
             'level': logging.WARNING,
             'propagate': True,
         },
+        # Downgrade DisallowedHost so it doesn't show up in Sentry
+        'django.security.DisallowedHost': {
+            'handlers': ['mozlog'],
+            'level': logging.WARNING,
+            'propagate': False,
```
wagnerand, line 1223: Not sure if it should propagate or not.

diox (changes requested):
That doesn't look correct - I don't think it'd work ? Nothing about this downgrades anything, it wouldn't have any impact.

I think if we want to go the downgrade route the proper way to do this would be a [python logging filter](https://docs.python.org/3/library/logging.html#logging.Filter.filter) attached to the config, but I'm not sure how this would interact with sentry integration, and it's a little annoying to write.

But maybe we don't need to! Instead, I think we should leave the logging config intact, don't downgrade the log, and just configure sentry: either using its [filtering callback](https://docs.sentry.io/platforms/python/guides/django/configuration/filtering/) or its API to [ignore specific loggers](https://docs.sentry.io/platforms/python/guides/logging/#ignoring-a-logger).

wagnerand:
You are right, I misread what `level` means.

Django specifically covers this behavior in the [documentation](https://docs.djangoproject.com/en/2.2/topics/logging/#django-security). Alternatively to use a python logging filter, we can also use a [django logging filter](https://docs.djangoproject.com/en/2.2/topics/logging/#id5).

Let me know what approach exactly we should pursue here.

diox:
I think the more robust solution is going to be the sentry approach, because ultimately that's the thing we care about.

wagnerand:
I found that the [python package](https://getsentry.github.io/sentry-python/api.html) supports passing `ignore_errors`, which looks simpler than using the filtering callback.

diox (changes requested):
This doesn't seem to work.

From the docs, the recommended way is to use `ignore_logger`, which does seem to work.

```python
from sentry_sdk.integrations.logging import ignore_logger
(...)
ignore_logger('django.security.DisallowedHost')
```

(Placing the `ignore_logger(...)` after the `sentry_sdk.init` call)

diox (approved):

PR #16904, opened 2021-04-09
implement sitemap.xml with basic site and addon support

### LINKED ISSUE #16508: Create a sitemap
We'd like to expose an XML [sitemap](https://developers.google.com/search/docs/advanced/sitemaps/build-sitemap) to help search engines discover all listed content on AMO.

Several considerations need to be discussed:
- Performance. It likely needs to be pre-generated, or cached on disk.
- Refresh period, how fresh the contents needs to be
- Figuring out what pages go in it, how to deal with translations/languages in general

Django has [a framework for this](https://docs.djangoproject.com/en/3.0/ref/contrib/sitemaps/) we'd likely build upon. It supports paginating the sitemap and most basic things we'd need.

eviljeff:
minimally fixes #16508 though there is more to do before I would consider the sitemap ready to be exposed to search engines.

I'll log follow-ups for the following acknowledged deficiencies:
- we don't have all the generic AMO pages - addons-server doesn't "know about" every frontend page so I just added the ones we already had defined in urls.pys (the vast majority of the urls in addons-server are behind auth, or apis)
- only the add-on detail pages are there.  Other add-on specific pages need adding
- collections (I trivially added them, but dropped it as the patch was already big enough for now)
- localized urls and non-Firefox app urls
- job isn't in crontab.tpl yet

Needs a patch to cloudops-deployment scripts to add sitemap.xml to ngnix 

If you decide to test this locally you'll hit the bug I tried to fix in mozilla/addons-frontend#10308 but got backed out in master.


src/olympia/lib/settings_base.py
```diff
@@ -516,11 +519,15 @@ def get_db_config(environ_var, atomic_requests=True):
     'django.contrib.humanize',
     'django.contrib.messages',
     'django.contrib.sessions',
+    'django.contrib.sitemaps',
+    # 'django.contrib.sites',
```
bobsilverberg, line 523: What's this?

src/olympia/lib/settings_base.py
```diff
@@ -516,11 +519,15 @@ def get_db_config(environ_var, atomic_requests=True):
...
     'django.contrib.staticfiles',
     # Has to load after auth
     'django_statsd',
 )
 
+# SITE_ID = 1
```
bobsilverberg, line 529: Leftover from testing?

bobsilverberg (approved):
Looks good, nice work @eviljeff. Just some nits.

r+wc

src/olympia/amo/tests/sitemaps/sitemap-addons-2.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml">
+<url><loc>http://testserver/en-US/firefox/addon/delicious-pierogi/</loc><lastmod>2020-10-02</lastmod><changefreq>daily</changefreq><priority>1</priority></url><url><loc>http://testserver/en-US/firefox/addon/swanky-curry/</loc><lastmod>2020-10-01</lastmod><changefreq>daily</changefreq><priority>1</priority></url><url><loc>http://testserver/en-US/firefox/addon/spicy-pierogi/</loc><lastmod>2020-09-30</lastmod><changefreq>daily</changefreq><priority>1</priority></url>
```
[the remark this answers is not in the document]
eviljeff: django's [template](https://github.com/django/django/blob/2.2.20/django/contrib/sitemaps/templates/sitemap.xml) wraps the output in `{% spaceless %}` so it strips out all the line returns - as I'm doing a byte-by-byte comparison in the tests I didn't prettify it.

src/olympia/lib/settings_base.py
eviljeff: 🙈

src/olympia/lib/settings_base.py
eviljeff: 😬

PR #16923, opened 2021-04-12
Fixes #16744 Extension Workshop URLS must end with a trailing slash to not lose query params

### LINKED ISSUE #16744: Extension Workshop URLS must end with a trailing slash to not lose query params
Related to https://github.com/mozilla/extension-workshop/issues/935 and https://github.com/mozilla/addons-server/issues/16649 

Query params are being lost in the redirection to add a slash. Since there doesn't appear to be a way to fix the redirection on s3, for now our best mitigation will be to ensure all extension workshop URLS have a trailing slash to avoid the redirection.

To fix this we'll need to search for extension workshop urls and update accordingly.

Here's a [repo search](https://github.com/mozilla/addons-server/search?q=extensionworkshop.com), but a code search locally should also be checked.

siddhi523:
Fixes #16744 

Added **/** before **?** in three links inside src/olympia/devhub/templates/devhub/nav.html

### DIFF
--- a/src/olympia/devhub/templates/devhub/nav.html
+++ b/src/olympia/devhub/templates/devhub/nav.html
@@ -34,11 +34,11 @@
     <li>
       <a href="#" class="controller">{{ _('Documentation') }}</a>
       <ul>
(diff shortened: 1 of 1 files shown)


eviljeff (approved):
r+ thanks

PR #16961, opened 2021-04-16
add some more pages to sitemap

### LINKED ISSUE #16905: add all (remaining) frontend pages to sitemap
follow up for #16508

addons-server doesn't "know about" every frontend page so the pr for #16508 just added the ones we already had defined in urls.pys (the vast majority of the urls in addons-server are behind auth, or apis).  

~Pages with pagination/filters are a problem - e.g. category pages - because we a) don't know how many pages there are without either attempting to replicate the same query that the API would run in elasticsearch, or actually calling the API, and b) filters cause a potential explosion in the number of pages (if the default filter is effectively filtering nothing out I think we can avoid the filtered results).~

~As we exclude search result this wasn't previously a concern, but category pages are effectively search pages and inherit the same difficulties that caused us to exclude them.~
(spun out category pages to #16959)

Also, a future problem will be keeping everything in sync on addons-server when pages are added/removed to addons-frontend.

eviljeff:
fixes #16905 - this turned out to be a small patch after the category result pages were spun out to a different issue.

### DIFF
--- a/src/olympia/amo/sitemap.py
+++ b/src/olympia/amo/sitemap.py
@@ -83,12 +83,6 @@ sitemaps = {
     'amo': AMOSitemap(),
     'addons': AddonSitemap(),
     'collections': CollectionSitemap(),
-    # category pages (wrapped search results of addons in a category)
-    # blocked-addons?
-    # user profiles
-    # error pages?
-    # addon review pages (+ direct?)
-    # addon privacy, eula, license, versions
 }
 
 
--- a/src/olympia/amo/sitemap.py
+++ b/src/olympia/amo/sitemap.py
@@ -40,10 +40,18 @@ class AMOSitemap(Sitemap):
 
     def items(self):
         return [
+            # frontend pages
+            'home',
             'pages.about',
-            'contribute.json',
-            'devhub.index',
             'pages.review_guide',
+            'browse.extensions',
+            'browse.extensions.categories',
+            'browse.themes',  # TODO: when we add /android, .themes are /firefox only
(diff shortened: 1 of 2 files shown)


src/olympia/amo/sitemap.py
```diff
@@ -75,6 +83,12 @@ def location(self, item):
     'amo': AMOSitemap(),
     'addons': AddonSitemap(),
     'collections': CollectionSitemap(),
+    # category pages (wrapped search results of addons in a category)
+    # blocked-addons?
+    # user profiles
+    # error pages?
+    # addon review pages (+ direct?)
+    # addon privacy, eula, license, versions
```
willdurand, line 91: These comments look like your own todo list :p Do we really need to commit them?

willdurand (commented):
So there is no test cases?

eviljeff:
> 
> 
> So there is no test cases?

No additional test cases, no.  `test_amo_sitemap` already trivially checks all the items and testing that `reverse` works is beyond the scope of the patch.

src/olympia/amo/sitemap.py
eviljeff: heh, maybe?  There's a splattering of `# TODO:` all over this file.  But I'll remove.

willdurand:
> No additional test cases, no. `test_amo_sitemap` already trivially checks all the items and testing that `reverse` works is beyond the scope of the patch.

ah ok, makes sense.

willdurand (approved):
r+wc

My only comment is about the comments (hehe).

PR #16963, opened 2021-04-16
collection code cleanup

### LINKED ISSUE #16935: remove some cruft from bandwagon/models.py
Most of the custom `CollectionManager` class appears unused.

eviljeff:
fixes #16935 - the pr (finally) removes the concept of collection "types" and per application collections, which we've ignored ever since collections on frontend became a thing (the API, by design, doesn't filter/differentiate by `type` or `application`, so they were irrelevant)

Also we get a bonus 1 less query on all django served pages from not checking it the user had an old "favorites" collection on every request (noticeable by the reviewers tests that counted queries)


src/olympia/bandwagon/models.py
```diff
@@ -19,76 +17,27 @@
 from olympia.users.models import UserProfile
 
 
-SPECIAL_SLUGS = amo.COLLECTION_SPECIAL_SLUGS
-
-
-class CollectionQuerySet(BaseQuerySet):
-    def with_has_addon(self, addon_id):
```
bobsilverberg, line 26: Are we just removing this because it isn't currently used? I know there are still some issues opened on addons-frontend that might make use of this if we ever get around to addressing them. It's unlikely we will - they are issues that keep becoming stale and then are commented on by an interested user, but I wonder if it's worth keeping this functionality around in case a contributor wants to fix those issues themselves? Just a thought.

src/olympia/bandwagon/models.py
```diff
@@ -166,13 +103,6 @@ def remove_addon(self, addon):
     def owned_by(self, user):
         return user.id == self.author_id
 
-    def can_view_stats(self, request):
```
bobsilverberg, line 169: Do collection stats no longer exist?

src/olympia/landfill/serializers.py
```diff
@@ -316,7 +316,7 @@ def create_featured_theme(self):
             promoted=RECOMMENDED,
         )
         addon.save()
-        generate_collection(addon, app=FIREFOX, type=amo.COLLECTION_RECOMMENDED)
+        generate_collection(addon, app=FIREFOX)
         print('Created Theme {0} for testing successfully'.format(addon.name))
 
     def create_featured_collections(self):
```
bobsilverberg, line 322: Do we still need this? Are featured collections still a thing?

bobsilverberg (approved):
Looks good, nice cleanup @eviljeff! I added some questions and also have a general question: What is `listed` used for with respect to collections? Do we still need it?

r+wc

eviljeff:
> I added some questions and also have a general question: What is `listed` used for with respect to collections? Do we still need it?

`listed` is if the collection is publicly available.  We didn't intend to keep it around after the switch to frontend/apis but the problem of existing private collections was never really resolved so it's still supported.

src/olympia/bandwagon/models.py
eviljeff: The whole issue was removing code because it wasn't currently used 😉 

For frontend to be able to do anything with this it would need to be exposed via the API... and if _that_ work was done it'd be trivial to reimplement it there.

src/olympia/bandwagon/models.py
eviljeff: nope, and if anything called this method (or it was tested) it would have failed because the `COLLECTION_STATS_VIEW` permission is already gone.

src/olympia/landfill/serializers.py
eviljeff: They aren't.  I didn't touch landfill more than I had to... though that's been our strategy for years so it's an unmaintained mess.  I'll delete this function - it's likely a lot of the other code in here that generates collections is unnecessary too.

eviljeff: https://github.com/mozilla/addons-server/pull/16963/commits/f4160ce7401b10b606fc725c0ed582a84855756a

Issue #10357, opened 2019-01-10
Update autograph to 2.7.0


EnTeQuAk:
To support #9308 and to make sure AMO works with the latest additions we need to update autograph for our local development systems and travis to 2.7.0.

PR #16967, opened 2021-04-16
RE: Add addon_type property to shelf module, remove search-themes shelf endpoint type

### LINKED ISSUE #16950: Adjust size of themes in collections within homepage shelves
Per [#10357](https://github.com/mozilla/addons-frontend/issues/10357), the size of the themes within collection shelves are 266x34 instead of 375x48. Similar to `search` and `search-themes`, adding a `collections-themes` shelf endpoint type will help differentiate between collections with themes and collections with extensions.

114522413-0a593c80-9c4c-11eb-8db3-3395359c3312

xlisachan:
Fixes #16950
In relation to [#10357](https://github.com/mozilla/addons-frontend/issues/10357)
Companion to [#10439](https://github.com/mozilla/addons-frontend/pull/10439)

This pull request removes `search-themes` as a shelf endpoint type and adds `addon_type` property to `shelves`.

The following files have been updated:
- `api/shelves.rst`
- `shelves/forms.py`
- `shelves/models.py`
- `shelves/serializers.py`
- `shelves/tests/test_forms.py`
- `shelves/tests/test_serializers.py`
- `shelves/migrations/0014_auto_20210416_2225.py`
- `discovery/tests/test_admin.py`


docs/topics/api/shelves.rst
```diff
@@ -100,6 +100,7 @@ small number of shelves this endpoint is not paginated.
     :>json object|null results[].headline: The title of the shelf. (See :ref:`translated fields <api-overview-translations>`.  Note: when ``lang`` is not specified, not all locales will be returned, unlike other translated fields).
     :>json string results[].url: The configured URL using the shelf's endpoint and criteria; links to the shelf's returned add-ons.
     :>json string results[].endpoint: The :ref:`endpoint type <shelf-endpoint-type>` selected for the shelf.
+    :>json string results[].addon_type: The type of the addons selected for the shelf (e.g., extensions or themes).
```
eviljeff, line 103: link to `<addon-detail-type>`

src/olympia/shelves/models.py
```diff
@@ -29,9 +33,12 @@ class Shelf(ModelBase):
     )
     addon_count = models.PositiveSmallIntegerField(
         default=0,
-        help_text='0 means the default number (4, or 3 for search-themes) of add-ons '
+        help_text='0 means the default number (4, or 3 for themes) of add-ons '
         'will be included in shelf responses. Set to override.',
     )
+    addon_type = models.CharField(
+        max_length=200, choices=ADDON_TYPE_CHOICES, default='extensions'
```
eviljeff, line 40: To be consistent, this should be the same addon type we use for `Addon` (it's an int).

src/olympia/shelves/forms.py
```diff
@@ -30,24 +31,15 @@ def clean(self):
...
-            params = criteria[1:].split('&')
-            if endpoint == 'search' and 'type=statictheme' in params:
-                raise forms.ValidationError(
-                    'Use "search-themes" endpoint for type=statictheme.'
-                )
-            elif endpoint == 'search-themes' and 'type=statictheme' not in params:
-                raise forms.ValidationError(
-                    'Don`t use "search-themes" endpoint for non themes. '
-                    'Use "search".'
-                )
```
eviljeff, line 50: Does this validation need to go?  I would have thought it'd be pretty simple to change it to rely on `addon_type` instead.

eviljeff (changes requested):
changing the type of addon_type is going to be quite disruptive so I'll have another look after that change has been made.

bobsilverberg:
Just a note that there needs to be a companion issue on addons-frontend for this, I believe. If we just land this change it may cause a regression on addons-frontend.

There were some discussions recently about landing these types of changes, where a change is needed both on addons-server and addons-frontend. We often deal with that by just making sure both issues are fixed in the same milestone, but the problem with that is that it's possible that production will regress during deployment as one of them is updated before the other.

To be sure that doesn't happen we would need to make sure the front-end code can deal with both the current API response and the new API response, and then, in a future milestone, remove the code that deals with the "old" API response.

Does all of that make sense @eviljeff @xlisachan ?

eviljeff:
it does @bobsilverberg - something we need to be careful of now this code is actually used on AMO production.

eviljeff:
@bobsilverberg thinking - do we need that frontend change to land before (i.e. in a previous milestone) to the api patch?

bobsilverberg:
> @bobsilverberg thinking - do we need that frontend change to land before (i.e. in a previous milestone) to the api patch?

Yes, I suppose we do, otherwise we still run the risk of the server change being deployed prior to the frontend change.

@xlisachan This means you'll need to land the frontend changes first, and it will need to deal with both the current API response and also the new one, based on this patch. Do we have a frontend issue for that already?

xlisachan:
Thanks for the feedback, @eviljeff! The latest push updates the API doc, model, and adds the validation back in `forms.py`. Please note, in `test_forms.py`, I removed `def test_clean_returns_empty`, as it was initially used to check that the admin did not use `type=theme` instead of `type=statictheme` in the criteria field. This is now prevented with the add_on type validation. Please review when you have a moment. Thank you!

@bobsilverberg I will open a frontend issue and start working on the fix now. I'll let you know if I have any questions!

src/olympia/discovery/tests/test_admin.py
```diff
@@ -881,6 +881,7 @@ def test_can_edit_with_discovery_edit_permission(self):
         item = Shelf.objects.create(
             title='Popular extensions',
             endpoint='search',
+            addon_type=1,
```
eviljeff, line 884: change the literal `1` to `amo.ADDON_EXTENSION` everywhere in this file.

src/olympia/shelves/models.py
```diff
@@ -40,7 +46,7 @@ def __str__(self):
         return self.title
 
     def get_count(self):
-        return self.addon_count or (3 if self.endpoint in ('search-themes',) else 4)
+        return self.addon_count or (3 if self.addon_type in (10,) else 4)
```
eviljeff, line 49:
```suggestion
        return self.addon_count or (3 if self.addon_type == amo.ADDON_STATICTHEME else 4)
```

src/olympia/discovery/tests/test_admin.py
```diff
@@ -1090,6 +1102,7 @@ def setUp(self):
         self.shelf = Shelf.objects.create(
             title='Populâr themes',
             endpoint='search',
+            addon_type=10,
```
eviljeff, line 1105: `ADDON_STATICTHEME`

src/olympia/shelves/serializers.py
```diff
@@ -34,13 +34,14 @@ class Meta:
             'title',
             'url',
             'endpoint',
+            'addon_type',
```
eviljeff, line 37: specify the field above like `AddonSerializer` does https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/serializers.py#L444

eviljeff: (that will do the conversion from int to a string value)

src/olympia/shelves/tests/test_forms.py
```diff
@@ -51,18 +51,13 @@ def setUp(self):
...
-            status=200,
-            json={'count': 0},
-        )
 
     def test_clean_search(self):
         form = ShelfForm(
             {
                 'title': 'Recommended extensions',
                 'endpoint': 'search',
+                'addon_type': 1,
```
eviljeff, line 60: same thing with using `ADDON_EXTENSION` in this file

src/olympia/shelves/tests/test_forms.py
```diff
@@ -246,17 +232,36 @@ def test_clean_returns_not_200(self):
             form.clean()
         assert exc.exception.message == ('Check criteria - Invalid "sort" parameter.')
 
-    def test_clean_returns_empty(self):
+    def test_clean_themes_addontype_used_for_statictheme_type(self):
+        data = {
+            'title': 'Recommended extensions',
+            'endpoint': 'search',
+            'addon_type': 10,
```
eviljeff, line 239: use the theme constant here

src/olympia/shelves/tests/test_serializers.py
```diff
@@ -98,7 +98,8 @@ def setUpTestData(cls):
     def setUp(self):
         self.search_pop_thm = Shelf.objects.create(
             title='Populâr themes',
-            endpoint='search-themes',
+            endpoint='search',
+            addon_type=10,
```
eviljeff, line 102: use the addon type constant when creating the `Shelf` object (and elsewhere in this file)

src/olympia/shelves/tests/test_serializers.py
```diff
@@ -149,6 +152,7 @@ def test_basic(self):
         data = self.serialize(self.search_rec_ext)
         assert data['title'] == {'en-US': 'Recommended extensions'}
         assert data['endpoint'] == 'search'
+        assert data['addon_type'] == 1
```
eviljeff, line 155: but the serialized result should be a string - it's okay to literally test for `'extension'` here (no need to do any lookup for the string constant for an extension)

eviljeff: And the same elsewhere in the file - Shelf model should be using `ADDON_EXTENSION` (or whatever); testing `data['addon_type']` should be checking for the literal string value.

src/olympia/shelves/forms.py
```diff
@@ -30,24 +31,27 @@ def clean(self):
...
+        addon_type = data.get('addon_type')
         criteria = data.get('criteria')
 
-        if criteria is None:
+        if None in (addon_type, criteria):
             return
 
-        if endpoint in ('search', 'search-themes'):
+        params = criteria[1:].split('&')
+        if addon_type == 1 and 'type=statictheme' in params:
```
eviljeff, line 41: `ADDON_EXTENSION`

src/olympia/shelves/forms.py
```diff
@@ -30,24 +31,27 @@ def clean(self):
...
+            raise forms.ValidationError(
+                'Use "Theme (Static)" in Addon type field for type=statictheme.'
+            )
+        elif addon_type == 10 and 'type=statictheme' not in params:
```
eviljeff, line 45: `ADDON_STATICTHEME`

src/olympia/shelves/forms.py
```diff
@@ -30,24 +31,27 @@ def clean(self):
         base_url = settings.INTERNAL_SITE_URL
 
         endpoint = data.get('endpoint')
+        addon_type = data.get('addon_type')
         criteria = data.get('criteria')
 
-        if criteria is None:
+        if None in (addon_type, criteria):
```
eviljeff, line 37: it's correct but I think `if addon_type is None or criteria is None:` is more explicit.

eviljeff: But is that what we actually want?  I.e. can `addon_type` ever be `None` and the form be valid?

eviljeff (changes requested):

src/olympia/shelves/forms.py
xlisachan: Ahh, the default is set to 'extension' and the field will always have a selected option, so `addon_type` can never be set to `None`.

xlisachan:
This latest push updates the `addon_type` to reflect the name instead of its integer.

src/olympia/discovery/tests/test_admin.py
```diff
@@ -912,13 +915,15 @@ def test_can_edit_with_discovery_edit_permission(self):
             assert Shelf.objects.count() == 1
             assert item.title == 'Recommended extensions'
             assert item.endpoint == 'search'
+            assert item.addon_type == 1
```
eviljeff, line 918: missed one

src/olympia/discovery/tests/test_admin.py
```diff
@@ -1006,6 +1014,7 @@ def test_can_add_with_discovery_edit_permission(self):
             item = Shelf.objects.get()
             assert item.title == 'Recommended extensions'
             assert item.endpoint == 'search'
+            assert item.addon_type == 1
```
eviljeff, line 1017: another one here

eviljeff (approved):
r+wc just a couple of fixes and then we can merge

bobsilverberg:
> r+wc just a couple of fixes and then we can merge

But are we going to merge it? Does it need to land in a milestone *after* the change to addons-frontend, or will this server code also work with both the new and current versions of frontend code?

eviljeff:
> 
> 
> > r+wc just a couple of fixes and then we can merge
> 
> But are we going to merge it? Does it need to land in a milestone _after_ the change to addons-frontend, or will this server code also work with both the new and current versions of frontend code?

that will depend on the existing frontend code and how the homepage is currently set up - the extra field is fine, but the endpoint for a shelf that's based on search and contains themes will change from `search-themes` to `search` and I assume _that_ would be a problem.  But really only for moment during the deploy window so maybe that's fine.  

If it's not fine then either the frontend patch, or this patch, needs to be compatible with both styles; that patch would need to deploy the previous tag; and then the following week the compat code would be deleted (so over 3 weeks)

xlisachan:
@eviljeff, the latest push updates the `addon_type` values to the constants instead of integer in `test_admin.py`. Please review when you have a moment. Thank you!

bobsilverberg:
> > > r+wc just a couple of fixes and then we can merge
> > 
> > 
> > But are we going to merge it? Does it need to land in a milestone _after_ the change to addons-frontend, or will this server code also work with both the new and current versions of frontend code?
> 
> that will depend on the existing frontend code and how the homepage is currently set up - the extra field is fine, but the endpoint for a shelf that's based on search and contains themes will change from `search-themes` to `search` and I assume _that_ would be a problem. But really only for moment during the deploy window so maybe that's fine.
> 
> If it's not fine then either the frontend patch, or this patch, needs to be compatible with both styles; that patch would need to deploy the previous tag; and then the following week the compat code would be deleted (so over 3 weeks)

It's a good point about whether the code will break entirely, or if there will just be a minor blip. 

I believe that if the server patch lands before the frontend patch, then the only problem would be that the shelves that contain themes would look a bit different, and that would only be for a brief period. If the frontend code lands first, it shouldn't matter at all, as it deals with both versions of the server code. So I think we're fine to land both of these in the same milestone. Do you agree @xlisachan?

xlisachan:
@bobsilverberg, yep! If the server patch lands first, then the shelves that had a `search-theme` endpoint will have the appearance of extensions until the frontend code lands. If that's not a problem, then it should be fine.

bobsilverberg:
> @bobsilverberg, yep! If the server patch lands first, then the shelves that had a `search-theme` endpoint will have the appearance of extensions until the frontend code lands. If that's not a problem, then it should be fine.

Thanks for the confirmation @xlisachan! Ok, it sounds like each of these patches can be merged independently.

eviljeff:
We forgot about existing shelves - I filed #17005 to fix them.

PR #16983, opened 2021-04-20
Fix reviewers review log filtering to include force disable/enable for listed reviewers

### LINKED ISSUE #16743: The 'Forced disabled/enabled' link in Add-on review log redirects to the unlisted review page for all add-ons
Follow up for #16637 

1. Log in with an Admin reviewer
2. Open the review page for a listed add-on 
3. Force-disable the add-on
4. Open the Add-on review log and look for the Forced-disable action recorded there
5. Click on the 'Force disabled' link in the Event description 

### What happened?
You land on the add-on unlisted review page

You should land on the add-on listed review page

- reviewers that do not have Review:Unlisted permissions will not see the Force disable/enable action in the Add-on review log because of this issue
- the same applies to the Force enable link 

Add-on used in this example:
https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/webext-test-1ab1c1

9HmItyOT7p

diox:
Flip the logic around: if the user doesn't have unlisted permission, instead of filtering to only show activities linked to a version on listed channel, exclude those on unlisted channel. That makes activities not linked to a particular version, like force disable/enable, visible for those users.

Same idea applied to the review link URL: instead of defaulting to unlisted and changing that if the activity is linked to a listed version, default to listed and only link to unlisted review page if the activity is linked to an unlisted version.

Fixes #16743

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/reviewlog.html
+++ b/src/olympia/reviewers/templates/reviewers/reviewlog.html
@@ -50,10 +50,10 @@
                   {% endif %}
                   {% if item.action == amo.LOG.REJECT_CONTENT.id or item.action == amo.LOG.APPROVE_CONTENT.id %}
                     {% set review_url = url('reviewers.review', 'content', item.arguments[0].slug ) %}
-                  {% elif channel == amo.RELEASE_CHANNEL_LISTED %}
-                    {% set review_url = url('reviewers.review', item.arguments[0].slug) %}
-                  {% else %}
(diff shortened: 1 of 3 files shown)


eviljeff (approved):
every time I look at or write code referencing `RELEASE_CHANNEL_UNLISTED` I think: "I should have chosen a more concisely named constant!"

diox:
It's so tempting to rename the constant and drop the `RELEASE_` bit every time I use it...

PR #16987, opened 2021-04-20
in recreate_theme_previews run generate_static_theme_preview with adhoc queue

### LINKED ISSUE #16973: run the theme preview regeneration tasks in a different task queue
follow up for #16920

we generate theme previews in the priority queue because normally we want to display them to the theme developer straight away. But if we do this for `recreate_theme_previews` the priority queue ends up full of themes, which blocks other priority tasks from completing (including elasticsearch reindex tasks leading to a stale index).

eviljeff:
fixes #16973

### DIFF
--- a/src/olympia/addons/tasks.py
+++ b/src/olympia/addons/tasks.py
@@ -28,11 +28,8 @@ from olympia.devhub.tasks import resize_image
 from olympia.files.utils import get_filepath, parse_addon
 from olympia.lib.es.utils import index_objects
 from olympia.tags.models import Tag
-from olympia.versions.models import (
-    generate_static_theme_preview,
-    Version,
-    VersionPreview,
-)
+from olympia.versions.models import Version, VersionPreview
+from olympia.versions.tasks import generate_static_theme_preview
 
 
 log = olympia.core.logger.getLogger('z.task')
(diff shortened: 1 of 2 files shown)


src/olympia/addons/tasks.py
```diff
@@ -243,7 +240,9 @@ def recreate_theme_previews(addon_ids, **kw):
             log.info('Recreating previews for theme: %s' % version.addon_id)
             xpi = get_filepath(version.all_files[0])
             theme_data = parse_addon(xpi, minimal=True).get('theme', {})
-            generate_static_theme_preview(theme_data, version.id)
+            generate_static_theme_preview.apply_async(
+                args=(theme_data, version.id), queue='adhoc'
```
eviljeff, line 244: despite our warnings in settings_base, I guess this is okay @diox ?

bobsilverberg (approved):
Looks fine to me, but I see there is also a question for @diox.

src/olympia/addons/tasks.py
diox: I guess so. It's worth trying out (though maybe confirm with ops if the queues are also separated on dev)

eviljeff: ops confirmed the queues are separate on addons-dev too

PR #16994, opened 2021-04-21
Fix upload path generation and log it when running through the linter

### LINKED ISSUE #14865: Please log the file name before it's linted
Related to https://bugzilla.mozilla.org/show_bug.cgi?id=1650878, it sounds like it's a good idea to log the XPI file name before it is linted.

This way, it is easier to find what XPI may have caused OOM for node for example.

It is also hinted in the bug ticket that we may have logged "a little too much when we repack". So maybe it is a good idea to remove some of the "too much" logging when you are there.

### What happened?
This request came from the discussion in the above bugzilla ticket.

The name of the file being linted is logged before the linting happens.

diox:
Fixes #14865

### DIFF
--- a/src/olympia/devhub/tasks.py
+++ b/src/olympia/devhub/tasks.py
@@ -298,6 +298,7 @@ def validate_file_path(path, channel):
             path=path, results=results, parsed_data=data, error=not is_mozilla_signed
         )
         return json.dumps(results)
+    log.info('Running linter on %s', path)
     return run_addons_linter(path, channel=channel)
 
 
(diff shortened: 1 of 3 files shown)


src/olympia/devhub/tasks.py
```diff
@@ -307,7 +308,6 @@ def forward_linter_results(results, upload_pk):
     linter results to `handle_upload_validation_result()` (the callback of the
     chord).
     """
-    log.info('Called forward_linter_results() for upload_pk = %d', upload_pk)
```
willdurand, line 310: That was kinda useful to me, but maybe it could be a `.debug()`?

willdurand (approved):
r+wc

src/olympia/devhub/tasks.py
diox: 76311339e9f67e0ea992ab2e466713c94945bc65

PR #16998, opened 2021-04-22
RE: Update shelves - collections shelves with themes

### LINKED ISSUE #16997: Update shelves - collections shelves with themes
Admin is unable to create collection homepage shelves when 'Theme (Static)' is selected in `addon_type`.

The admin receives a validation error directing them to check the form fields as 'Theme (Static)' is selected for `addon_type`. However, as the criteria field for collections is the slug, it would not include `type=statictheme` in its field. 

Note: admin is able to create the following shelves:
- `search` endpoint with `addon_type` extension
- `search` endpoint with `addon_type` statictheme
- `collections` endpoint with `addon_type` extension

xlisachan:
Fixes #16997

This pull request updates `forms.py` to check the params only when `search` is selected as the endpoint, as `collections` would not have `type=` in its criteria slug.

To be more thorough with the tests, there are now four distinct tests in `test_forms.py` to check each combination:
-`test_clean_search_extensions`,
-`test_clean_search_themes`,
-`test_clean_collections_extensions`,
-`test_clean_collections_themes`

In addition, the last two tests within `test_forms.py` were also updated. I have also updated the titles for clarification:
- `test_clean_cannot_use_theme_addontype_without_type_statictheme`
- `test_clean_cannot_use_extensions_addontype_with_type_statictheme`

Note, with this update, the admin is able to now create a shelf with the `collections` endpoint using either addon type. The code, however, is not checking whether the slug will return extensions or themes. Should this be a concern?

@eviljeff, please review when you have a moment. Thank you!

### DIFF
--- a/src/olympia/shelves/forms.py
+++ b/src/olympia/shelves/forms.py
@@ -38,21 +38,22 @@ class ShelfForm(forms.ModelForm):
         if criteria is None:
             return
 
-        params = criteria[1:].split('&')
-        if addon_type == amo.ADDON_EXTENSION and 'type=statictheme' in params:
-            raise forms.ValidationError(
-                'Use "Theme (Static)" in Addon type field for type=statictheme.'
-            )
-        elif addon_type == amo.ADDON_STATICTHEME and 'type=statictheme' not in params:
-            raise forms.ValidationError(
-                'Check fields - for "Theme (Static)" addon type, use type=statictheme. '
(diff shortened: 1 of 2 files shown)


eviljeff (approved):
I don't think we've tried to validate the actual results with other shelves work, so it's consistent.  The admin using it will be carefully choosing the collection, so it's fine I think.

Issue #16910, opened 2021-04-09
Include /android/ pages in the sitemap


eviljeff:
followup for #16508 

We currently generate all urls with the app in them as `/firefox/`.  We should also include `/android/` pages too - the landing pages are different but the detail pages are essentially the same (slightly different header, with some extra warnings about compatibility)

Questions:
- Should add-ons that only support desktop Firefox have an `/android/` page?  Should purely Android compatible add-ons have a `/firefox/` page?  They do exist if navigated to but we don't link to them as they're invalid, really.
- If we have both /firefox/ and /android/ urls, should one be an "alternative" of the other or included separately?

eviljeff:
@rraue ^ should we generate /android/ and /firefox for all addon detail pages?

rraue:
It can be in one or two split sitemaps. Doesnt matter. Important is only that all pages we want indexed are in the sitemaps.

willdurand:
See: https://github.com/mozilla/addons-server/pull/17004#discussion_r620967007

bobsilverberg:
@eviljeff Before I review this, did anything from today's meeting change anything about how we want to handle `android` URLs? It sounds like we only need to include pages that we actually want indexed, so we should probably exclude categories, and it also seems like it would make sense to only include detail pages for add-ons that actually work on Android.

willdurand:
> @eviljeff Before I review this, did anything from today's meeting change anything about how we want to handle `android` URLs? It sounds like we only need to include pages that we actually want indexed, so we should probably exclude categories, and it also seems like it would make sense to only include detail pages for add-ons that actually work on Android.

Is it really about "all pages we want indexed" or is it "all pages that are actually indexed"? Because I am not sure those two sets are the same.

eviljeff:
> @eviljeff Before I review this, did anything from today's meeting change anything about how we want to handle `android` URLs? It sounds like we only need to include pages that we actually want indexed, so we should probably exclude categories, 

we do exclude categories - there's a code comment about it.

> and it also seems like it would make sense to only include detail pages for add-ons that actually work on Android.

I asked that above, but I suppose the answer wasn't explicit.  The patch is blocking remaining work so any changes can be in a follow-up.

The general answer from the meeting - that I understood - was that the sitemap should reflect the pages we want indexed, but it's the meta tags in the page itself that defines it.  I.e. _just_ dropping pages from the sitemap doesn't help if they happen to be linked elsewhere; the meta tags in the html have to exclude the page.

[...]

Issue #16909, opened 2021-04-09
include localized urls in the sitemap


eviljeff:
followup for #16508 

We currently generate all pages for the sitemap as `/en-US/` but we should have localized urls too.  

There is some level of support in this already in the django sitemaps app but in django2.2 it's a bit too naive - it doesn't support the alternatives syntax to group the urls together, and mis-paginates them meaning we'd end up with more than 50,000 urls per file (the google search limit) unless we manually calculate how much lower it should be (`50000/<however many languages we support>`).  We need the django3.2 implementation.  If we really need this before we're ready to upgrade we can copy over the code from django3.2 to support django2.2 (it's not that much)

Other comments:
- we should prioritize/default to the addon's default locale, rather than `en-US`
- It's unclear what "list" of languages/locales we want to use - I think supported languages differs on addons-frontend and addons-server, and some of those locales have quite minimal coverage.

eviljeff:
ref for canonical/alternate url discussion https://github.com/mozilla/addons-frontend/issues/6644

AlexandraMoga:
@eviljeff I've noticed two things since this issue was closed:

- the sitemap https://addons-dev.allizom.org/sitemap.xml?debug has become super slow - it took for me ~ 2 minutes to load
- sitemap sections are no longer organized as an `.xml`? For example, when I'm loading https://addons-dev.allizom.org/sitemap.xml?section=amo&debug, I'm seeing only a sequence of URLs.

eviljeff:
>     * the sitemap https://addons-dev.allizom.org/sitemap.xml?debug has become super slow - it took for me ~ 2 minutes to load

I've logged #17178 to address this - the way django iterates through the locales is inefficient

>     * sitemap sections are no longer organized as an `.xml`? For example, when I'm loading https://addons-dev.allizom.org/sitemap.xml?section=amo&debug, I'm seeing only a sequence of URLs.

it's still valid sitemap xml (at least from the validators I could find on the web), but Firefox doesn't seem to like the extra `xhtml` tags (or gets confused) and tries to render it as html.

AlexandraMoga:
> it's still valid sitemap xml (at least from the validators I could find on the web), but Firefox doesn't seem to like the extra `xhtml` tags (or gets confused) and tries to render it as html.

so does Chrome too, which makes sections even harder to analyze now.

eviljeff:
[...]

PR #17004, opened 2021-04-22
include category pages, with pagination in sitemap

### LINKED ISSUE #16959: add category pages to sitemap
spun out from #16905

Pages with pagination/filters are a problem - e.g. category pages - because we a) don't know how many pages there are without either attempting to replicate the same query that the API would run in elasticsearch, or actually calling the API, and b) filters cause a potential explosion in the number of pages (if the default filter is effectively filtering nothing out I think we can avoid the filtered results).

### LINKED ISSUE #16993: Extension and theme categories links included in the sitemap should end with a trailing slash
follow up for #16508

1. Navigate to https://addons-dev.allizom.org/sitemap.xml?section=amo&debug
2. Locate the extension and theme categories links among the items

### What happened?
Extension and theme categories links do not end with a trailing slash, which will trigger a 301:
```
<loc>https://addons-dev.allizom.org/en-US/firefox/extensions/categories</loc>
```

Extension and theme categories links need to end with a trailing slash since that's the URL served by the frontend:
https://addons-dev.allizom.org/en-US/firefox/extensions/categories/

eviljeff:
fixes #16959 (fixes #16993 too)

### DIFF
--- a/src/olympia/amo/sitemap.py
+++ b/src/olympia/amo/sitemap.py
@@ -1,16 +1,21 @@
 import datetime
+import math
 import os
 from collections import namedtuple
 from urllib.parse import urlparse
 
+from django.db.models import Count
 from django.conf import settings
 from django.contrib.sitemaps import Sitemap
 from django.db.models import F
 from django.template import loader
 from django.urls import reverse
 
-from olympia.addons.models import Addon
+from olympia import amo
+from olympia.addons.models import Addon, AddonCategory
+from olympia.amo.reverse import get_url_prefix
 from olympia.amo.templatetags.jinja_helpers import absolutify
+from olympia.constants.categories import CATEGORIES
 from olympia.bandwagon.models import Collection
 from olympia.versions.models import License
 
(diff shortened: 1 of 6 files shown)


src/olympia/amo/sitemap.py
```diff
@@ -92,6 +97,45 @@ def location(self, item):
...
+        def additems(type):
+            items = []
+            for category in CATEGORIES[current_app.id][type].values():
+                items.append((category, 1))
+                pages_needed = math.ceil(addon_counts.get(category.id, 1) / page_size)
+                for page in range(2, pages_needed + 1):
+                    items.append((category, page))
+            return items
+
+        page_size = settings.REST_FRAMEWORK['PAGE_SIZE']
```
willdurand, line 116: I don't think that's the case for category pages but the frontend is free to use a different `page_size` value. I took a quick look and we probably don't set it very often in addons-frontend but that's not unlikely.

If we start to rely on this value here, we'll probably want tests in addons-frontend to avoid any changes around `page_size`.

src/olympia/amo/sitemap.py
```diff
@@ -92,6 +97,45 @@ def location(self, item):
         return reverse(item)
 
 
+class CategoriesSitemap(Sitemap):
+    priority = 0.7
+    # i18n = True  # TODO: support all localized urls
```
willdurand, line 102: I know it's a TODO but some localized URLs are excluded, see: https://github.com/mozilla/addons-frontend/issues/6644. It'd be interesting to know exactly which URLs should be part of the sitemap because I think it should be all URLs for which we have canonical URLs (which is a subset of _all possible URLs_)

src/olympia/amo/tests/test_sitemap.py
```diff
@@ -66,6 +70,43 @@ def test_amo_sitemap():
...
+    # should be 3 addons in shopping (one not public, so 2 public), and 3 in bookmarks
+
+    patched_drf_setting = dict(settings.REST_FRAMEWORK)
+    patched_drf_setting['PAGE_SIZE'] = 2
+    with override_settings(REST_FRAMEWORK=patched_drf_setting):
+        cats_with_addons = list(sitemap.items())
+    # only one extra url, for a second bookmarks category page, because PAGE_SIZE = 2
+    extra = (bookmarks_category, 2)
+    assert extra in cats_with_addons
+    assert set(cats_with_addons) - set(empty_cats) == {extra}
```
willdurand, line 107: we should probably have some data for Android categories, no? https://addons.mozilla.org/en-US/android/extensions/category/experimental/

src/olympia/amo/sitemap.py
eviljeff: #16909 for localized urls, but I'd not really worked out _which_ urls.  (I didn't know we'd already had that discussion in 2018!)

src/olympia/amo/tests/test_sitemap.py
eviljeff: #16910 for android - tbh I'm still not 100% on which urls should be included and which should be alternates (I see https://github.com/mozilla/addons-frontend/issues/6644 mentions android urls too).  For categories specifically, we do have category data, but we don't expose category pages on frontend so it's unclear (discussion in #16910 I guess).

src/olympia/amo/sitemap.py
eviljeff: > I don't think that's the case for category pages but the frontend is free to use a different `page_size` value. I took a quick look and we probably don't set it very often in addons-frontend but that's not unlikely.

So, what change are you proposing exactly?  We obviously need to assume a page_size to calculate how many pages.

> If we start to rely on this value here, we'll probably want tests in addons-frontend to avoid any changes around `page_size`.

perhaps, but that equally applies to any url that frontend serves.  Is the ask here to log an issue to add a test to frontend?

willdurand: Writing tests to prevent regressions is what I am suggesting.

willdurand (approved):

src/olympia/amo/sitemap.py
eviljeff: mozilla/addons-frontend#10498

Issue #13729, opened 2020-03-13
Clean up unused code around opensearch plugins


diox:
We [stopped accepting](https://github.com/mozilla/addons-server/issues/12462) and then [disabled all opensearch plugins](https://github.com/mozilla/addons-server/issues/12463).

To clean things up we should now remove all dead code related to opensearch plugins: there is a lot of special handling since they weren't extensions. 

To get started one can [set up addons-server locally](https://addons-server.readthedocs.io/en/latest/topics/install/index.html) and then try renaming `ADDON_SEARCH` constant in `src/olympia/constants/base.py` to `_ADDON_SEARCH` and see what tests break. Ultimately all code using that constant should probably be changed. Some classes, functions, tests can be entirely removed.

ajinkabeer:
Hey @diox, I am outreachy applicant. I would like to take work on this issue if that's alright. :)

junngo:
Can I work on it?

ajinkabeer:
Hello, I am an outreachy applicant. If this issue is still unassigned, I would like to take a look at it, please?

bhushan-borole:
Hello, I would like to work on this issue.

diox:
Feel free to start working on it and submit a PR when you're ready. It should be relatively straightforward as it's mostly about removing code.

bhushan-borole:
For running the tests, the command is `make test` right ?

diox:
Yes.

bhushan-borole:
> Yes.

Basically removing the part of the code, which uses this variable right?

diox:
Per description:

> Ultimately all code using that constant should probably be changed. Some classes, functions, tests can be entirely removed.

This is a `contrib: welcome` issue and not a `contrib: good first bug` issue so while I'll be happy to review a pull request I won't go into too much more details, you'll need to exercise judgement to figure out what can be removed exactly.

diox:
This is getting in my way so I'm taking it.

AlexandraMoga:
@diox do you have any specific STR for verifying this issue? I was planning to verify various add-on submissions and the Developer Hub mostly.

diox:
> I was planning to verify various add-on submissions and the Developer Hub mostly.

Yeah, just a general sanity check around add-on & versions submissions should be enough.

AlexandraMoga:
I haven't found issues with add-ons and version submissions.

I'm marking this issue as verified fixed on stage.

PR #17029, opened 2021-04-27
fix: remove "xml" as allowed filetype from uploader

Rob--W:
Follow-up to #13729 and #16090

I came across this small issue when I investigated the current state for https://github.com/mozilla/extension-workshop/pull/990

### DIFF
--- a/static/js/common/upload-addon.js
+++ b/static/js/common/upload-addon.js
@@ -32,7 +32,7 @@
 
   $.fn.addonUploader = function (options) {
     var settings = {
-      filetypes: ['zip', 'xpi', 'crx', 'xml'],
+      filetypes: ['zip', 'xpi', 'crx'],
       getErrors: getErrors,
       cancel: $(),
       maxSize: 200 * 1024 * 1024, // 200M


eviljeff:
@diox do we want this?  We still have `annotate_search_plugin_restriction` which gives an error message explaining why they aren't supported - a developer won't get this if we prevent them from uploading it in the first place.

diox:
That's the reason why this is still supported by the uploader, yeah. https://github.com/mozilla/addons-server/issues/16090 made it so we no longer advertise `.xml` upfront, which was possibly confusing to developers of WebExtensions - a more important point, IMHO.

That being said, maybe it's time to remove everything, including that error message ? In that case we would want not only to remove the `.xml` from allowed filetypes client side, but also remove `annotate_search_plugin_restriction` and related code server side. @jvillalobos to decide if we care about people that want to upload search plugins.

jvillalobos:
Since OpenSearch still works on Firefox, I think it's useful to continue giving useful information to prospective developers. I wouldn't remove if it isn't causing any problems.

PR #17037, opened 2021-04-28
sitemap user account profiles

### LINKED ISSUE #16960: Expose (some) account profile pages to sitemap
users who are developers (have >0 public addons) have a visible account profile page that links to their extensions/themes, so we should expose those pages in the sitemap too.

(double!) pagination for users who have many addons is a complication (which could be added in a follow-up)

eviljeff:
fixes #16960


bobsilverberg (approved):
Looks good, nice work @eviljeff!

PR #17048, opened 2021-04-29
add versions pages to sitemap

### LINKED ISSUE #17019: add versions to sitemap
split #16986 in 2

we need to add pages for versions for addons

eviljeff:
fixes #17019 - this is the last of the classes of pages we want to include in the sitemap (remaining work is to include localised and android versions of some/most of the urls)

### DIFF
--- a/src/olympia/amo/sitemap.py
+++ b/src/olympia/amo/sitemap.py
@@ -4,10 +4,9 @@ import os
 from collections import namedtuple
 from urllib.parse import urlparse
 
-from django.db.models import Count
 from django.conf import settings
 from django.contrib.sitemaps import Sitemap
-from django.db.models import F
+from django.db.models import Count, F, Q
 from django.template import loader
 from django.urls import reverse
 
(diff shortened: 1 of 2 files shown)


src/olympia/amo/sitemap.py
```diff
@@ -69,6 +68,51 @@ def location(self, item):
...
+                'last_updated',
+                'slug',
+                'versions_count',
+                named=True,
+            )
+        )
+        items = []
+        page_size = settings.REST_FRAMEWORK['PAGE_SIZE']
+        for addon in addons:
+            pages_needed = math.ceil(addon.versions_count / page_size)
```
bobsilverberg, line 100: I'm sorry this wasn't caught sooner, but the All Versions page on frontend does not support pagination, so I don't think we need to (or should) worry about multiple pages for this `Sitemap`.

bobsilverberg (changes requested):
Nice job, but I don't think we need to worry about pagination for this one. See my comment.

src/olympia/amo/sitemap.py
eviljeff: d'oh! I should have remembered that mozilla/addons-frontend#7542 was wontfix'd.  I'm going to close this PR and start again because it doesn't need any of this complexity now.

PR #17057, opened 2021-05-02
Issue #16498 Replace NullBooleanField with BooleanField(null=True)

### LINKED ISSUE #16498: Replace NullBooleanField with BooleanField(null=True) (django3.2)
`BooleanField(null=True)` has been an alternative for `NullBooleanField` for a while, and is deprecated in django3.2 (to be removed in 4.0).  Because of its upcoming deprecation there's a(n annoying) system check warning about it whenever you launch a django shell, etc.  As it's already supported in django2.2 and a no-op at the database level we should just make the change to stop the complaining. 

https://docs.djangoproject.com/en/2.2/ref/models/fields/#nullbooleanfield

ssudheerk99:
a `Fixes #16498 ` replaced Replace NullBooleanField with BooleanField(null=True) as mentioned in the issue summary


diox (changes requested):
- Please don't commit that huge change to `package-lock.json`
- You're missing a bunch of `default=True` and `default=False`, which in some cases is breaking tests. Please add them back carefully.

eviljeff:
Also, don't change references to `NullBooleanField` in existing migration files.

ssudheerk99:
Hi @diox added the default values wherever missing, but bit sure why code-lint is failing, I just changed 5 lines in that, below is the output from the circleCI

```
#!/bin/bash -eo pipefail
make lint-codestyle

black --check src/ services/ tests/
would reformat src/olympia/addons/models.py
Oh no! 💥 💔 💥
1 file would be reformatted, 648 files would be left unchanged.
make: *** [Makefile-docker:190: lint] Error 1

Exited with code exit status 2

CircleCI received exit code 2
```

eviljeff:
> Hi @diox added the default values wherever missing, but not sure why code-lint is failing, I just changed 5 lines in that, below is the output from the circleCI
> 
> ```
> #!/bin/bash -eo pipefail
> make lint-codestyle
> 
> black --check src/ services/ tests/
> would reformat src/olympia/addons/models.py
> Oh no! 💥 💔 💥
> 1 file would be reformatted, 648 files would be left unchanged.
> make: *** [Makefile-docker:190: lint] Error 1
> ```

probably the extra space in one of the field declarations, but you can fix automatically with `make format` locally.

PR #17078, opened 2021-05-05
generate android pages for sitemap too

### LINKED ISSUE #16910: Include /android/ pages in the sitemap

eviljeff:
fixes #16910


src/olympia/amo/sitemap.py
```diff
@@ -112,34 +116,46 @@ class AMOSitemap(Sitemap):
...
-            'browse.themes.categories',
-            'browse.language-tools',  # TODO: when we add /android this is /firefox only
+            ('home', amo.FIREFOX),
+            ('home', amo.ANDROID),
+            ('pages.about', None),
+            ('pages.review_guide', None),
+            ('browse.extensions', amo.FIREFOX),
+            ('browse.extensions', amo.ANDROID),
+            ('browse.extensions.categories', amo.FIREFOX),
+            ('browse.extensions.categories', amo.ANDROID),
```
bobsilverberg, line 131: Did we discuss excluding categories for Android, or is this something else?

bobsilverberg (approved):
Looks good, sorry for the delay @eviljeff. I just have one question.

r+wc

src/olympia/amo/sitemap.py
eviljeff: ah, this is the page that shows the _list_ of categories https://addons.mozilla.org/en-US/firefox/extensions/categories/ though I'm not sure we need that at all (or why it exists as a page).  

The addons in a particular category is handled in `CategoriesSitemap` (I thought that's what you were referring to in the issue)

eviljeff: I'll drop this url for Firefox too.

PR #17090, opened 2021-05-06
save transparent pngs with white background when converting to jpeg

### LINKED ISSUE #17028: Transparency in user generated content should be converted to the same colour as the details page background

If you upload a `png` screenshot which is nothing but a black or dark outline with transparency, when the image is converted to jpeg the png's transparent background becomes solid black by default. 

### What happened?

The preview shows the image uploaded as a png with a transparent background. However post conversion this turns into a completely black image.

Ideally we'd change the conversion to use a background for the jpeg that is equivalent to the transparency. Which should be possible by "pasting" the image onto a fill of our chosen colour via the image api.

An alternative would be to have the preview match the output of the jpeg conversion, but I would think a light background for transparency would  better reflect what's been happening prior to the png -> jpeg conversion.

Screenshot of edit screen after selecting image:

Edit_Product_Page____favicon-customizer____Add-ons_for_Firefox

Screenshot of details page post conversion:

favicon-customizer_–_Get_this_Extension_for_🦊_Firefox__en-US_

eviljeff:
fixes #17028


willdurand (approved):
Thanks!

Issue #16445, opened 2021-02-02
Remove tox


willdurand:
We don't use `tox` the way we should :tm: so let's get rid of it.

PR #17103, opened 2021-05-07
Ci jobs - move amo lib signing

### LINKED ISSUE #17102: amo-lib-locales-and-signing CI job is super slow, holding up other test verification
the amo-lib-locales-and-signing job became super slow after we moved to circleci - the locale compilation step alone takes 10+ (15+ is not uncommon) minutes, which blocks you from seeing that your patch hasn't broken tests under `amo` or `lib` that don't require locale extraction.  

The actual solution is to work out why the extraction become so slow, and fix it, if possible.  In the interim, let's move the tests that don't need to wait for the extraction to other jobs.

eviljeff:
fixes #17102 - moves the non-locales tests into other jobs (signing related tests to reviewers, which already set up autograph; the rest to main)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -529,15 +529,11 @@ jobs:
       - run:
           command: pytest -n 2 -m 'not es_tests and not needs_locales_compilation and not static_assets' -v src/olympia/addons/ src/olympia/versions/ src/olympia/files/ src/olympia/ratings/
 
-  amo-lib-locales-and-signing:
-    <<: *defaults-with-autograph
+  locales:
+    <<: *defaults-with-services
     steps:
       - setup_container:
           install_node_dependencies: true
-      - run:
-          command: pytest -n 2 -m 'not es_tests and not needs_locales_compilation and not static_assets' -v src/olympia/amo/ src/olympia/lib/ src/olympia/signing
-          environment:
-            AUTOGRAPH_SERVER_URL: http://127.0.0.1:5500
       - run: bash ./locale/compile-mo.sh ./locale/
       - run:
           command: pytest -n 2 -m 'needs_locales_compilation' -v src/olympia/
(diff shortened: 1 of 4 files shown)


eviljeff:
> I'd like to understand why we see a regression from after [the tox removal](https://github.com/mozilla/addons-server/issues/16445) before updating the CI config. 

I'd like to understand too I don't see why it's a blocker on deploying a workaround to speed up our workflow in the meantime.

> Also, I see a fix for a different issue I believe (mysql max signed int)

It's connected - see matrix chat.  The way we were testing the max value of an unsigned column broke subsequent tests that also used the `Group` model - it just wasn't noticeable when the `amo` tests were run in isolation.

willdurand:
> I'd like to understand too I don't see why it's a blocker on deploying a workaround to speed up our workflow in the meantime.

The "tox removal" likely impacted all the jobs and they are all slower since then, with one or two _very_ slow. We'd probably have a better impact by trying to fix the root cause now that we know it.

> It's connected - see matrix chat. The way we were testing the max value of an unsigned column broke subsequent tests that also used the `Group` model - it just wasn't noticeable when the `amo` tests were run in isolation.

Does that mean moving the non-locales tests in a different job requires the mysql max signed int change? I tried to find the discussion on Matrix but search isn't great. I don't find any mention of that in the issue linked above either.

diox:
Before the move, we never noticed how creating a group with a max int id was problematic, because there were no tests that ran after that one that created groups. So things worked by accident.

Moving the test revealed the problem: MySQL doesn't reset the auto increment value, even after a rollback. So if we create a group with a max int id... We can never create another group after that!

eviljeff:
> > I'd like to understand too I don't see why it's a blocker on deploying a workaround to speed up our workflow in the meantime.
> 
> The "tox removal" likely impacted all the jobs and they are all slower since then, with one or two _very_ slow. We'd probably have a better impact by trying to fix the root cause now that we know it.

None of the other jobs are unacceptably slow (at least historically).  The problem with the locales setup in particular isn't the test being generally slower - previously the tests were 4 minutes for the entire tox step (including the locale extraction _and_ the tests); now the locale extraction step _alone_ takes 14 minutes.

willdurand:
> None of the other jobs are unacceptably slow (at least historically). The problem with the locales setup in particular isn't the test being generally slower - previously the tests were 4 minutes for the entire tox step (including the locale extraction _and_ the tests); now the locale extraction step _alone_ takes 14 minutes.

I took a closer look and this seems to be a problem with pyenv shims. Pyenv is used by Circle to install python, and it looks like pyenv creates a shim for every python CLI. If we skip the pyenv shim and use the `dennis-cmd` binary directly, it is a lot more acceptable, see: https://github.com/mozilla/addons-server/pull/17136

eviljeff:
#17136 fixes the issue in a different way, so we don't need to reorganise the jobs any longer.  I'll move the mysql maxint fix into a different pr.

PR #17113, opened 2021-05-10
drop changefreq and priority from sitemap categories

### LINKED ISSUE #17093: drop changefreq and priority from sitemap entries
@rraue says we don't need them these days.

eviljeff:
fixes #17093


bobsilverberg (approved):
Bye bye arbitrary priorities! 😉 

Looks good!

PR #17150, opened 2021-05-13
Updated error message.

### LINKED ISSUE #17128: Update localized message when GUID in `manifest.json` does not match the AMO one
The error message returned when there is a GUID mismatch mentions `install.rdf`, which is quite old now and we don't support add-ons with such a file anymore: https://github.com/mozilla/addons-server/blob/74f256f8d91c8c6b016dc9bf660ce4fa6758eaed/src/olympia/files/utils.py#L997-L1002

We should update the message to:

```
The add-on ID in your manifest.json (%s) does not match the ID of your add-on on AMO (%s)
```

NegiAkash890:
Updated the old error message in utils.py

Fixes #17128

### DIFF
--- a/src/olympia/files/utils.py
+++ b/src/olympia/files/utils.py
@@ -996,7 +996,7 @@ def check_xpi_info(xpi_info, addon=None, xpi_file=None, user=None):
 
         if addon and addon.guid != guid:
             msg = gettext(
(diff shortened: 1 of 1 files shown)


eviljeff (approved):
thanks r+

eviljeff:
if you comment in the issue github will allow us to assign it to you

Issue #17163, opened 2021-05-17
Stop exposing license text in versions API


diox:
The versions list API exposes the license text for each version returned, which is wasteful. We should only return it for the version detail endpoint. This would speed up the add-on versions page endpoint as mentioned in https://github.com/mozilla/addons-frontend/issues/10446.

For backwards-compatibility we should only do this in v5, at least for now. Ideally we should also `defer` the field in the ORM to avoid even fetching the text from the database.

bobsilverberg:
Do we need a related issue for addons-frontend to deal with the fact that this will be missing?

diox:
I *think* it should already work:
And the various types defined have `text` as optional, and we have this:
https://github.com/mozilla/addons-frontend/blob/8ad42e516ff9dd81c5a05d35b44b36f16b644ff8/src/amo/reducers/versions.js#L138-L141

In addition, it looks like the `text` property is never accessed by the `AddonVersions` page or its components (which makes sense, given we only seem to display it when the user visits the special link to display the text of a custom license).

Let me know if I'm wrong though.

bobsilverberg:
> I _think_ it should already work:
> And the various types defined have `text` as optional, and we have this:
> [mozilla/addons-frontend@`8ad42e5`/src/amo/reducers/versions.js#L138-L141](https://github.com/mozilla/addons-frontend/blob/8ad42e516ff9dd81c5a05d35b44b36f16b644ff8/src/amo/reducers/versions.js#L138-L141)
> 
> In addition, it looks like the `text` property is never accessed by the `AddonVersions` page or its components (which makes sense, given we only seem to display it when the user visits the special link to display the text of a custom license).
> 
> Let me know if I'm wrong though.

It does look like we indeed do not use `text` on the `AddonVersions` page, so that's good. I notice, however, that that code in the reducer is specifically looking for `undefined`. Will we be getting `undefined` back from the API now, or might it be `null`?

diox:
You'll get `undefined`, the key won't even exist in the object in the JSON.

bobsilverberg:
> You'll get `undefined`, the key won't even exist in the object in the JSON.

Prefect! Thanks.

diox:
QA:
- License `text` should be gone for the [version list API](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#versions-list) in v5
- It should still be present in v3/v4
- It should still be present in the version detail API (single version as opposed to a list of multiple versions) for v3/v4/v5
[...]

PR #17166, opened 2021-05-17
Drop license text from version list API endpoint

diox:
Don't even fetch the license text from the database when it's not needed: defer it, and ensure the translations app doesn't try to fetch translations for deferred fields.

Fixes https://github.com/mozilla/addons-server/issues/17163

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -293,6 +293,12 @@ class VersionSerializer(SimpleVersionSerializer):
         )
 
 
+class VersionSerializerForListing(VersionSerializer):
+    # When we're listing versions, we don't want to include the full license
+    # text every time: we only do this for the version detail endpoint.
+    license = CompactLicenseSerializer()
+
+
 class CurrentVersionSerializer(SimpleVersionSerializer):
     def to_representation(self, obj):
         # If the add-on is a langpack, and `appversion` is passed, try to
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -293,7 +293,7 @@ class VersionSerializer(SimpleVersionSerializer):
         )
 
 
-class VersionSerializerForListing(VersionSerializer):
+class VersionListSerializer(VersionSerializer):
     # When we're listing versions, we don't want to include the full license
     # text every time: we only do this for the version detail endpoint.
     license = CompactLicenseSerializer()
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -3,6 +3,7 @@ import json
 
 from unittest import mock
 
+from django.conf import settings
 from django.core.cache import cache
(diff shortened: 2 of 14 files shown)


diox:
mmm I realized afterwards that I should gate this only for v5 API. I'll add that.

src/olympia/addons/tests/test_views.py
```diff
@@ -908,10 +916,9 @@ def _set_tested_url(self, param):
         self.url = reverse_ns('addon-version-list', kwargs={'addon_pk': param})
 
     def test_queries(self):
-        with self.assertNumQueries(13):
-            # 13 queries:
+        with self.assertNumQueries(11):
+            # 11 queries:
             # - 2 savepoints because of tests
-            # - 2 user and its groups
```
diox, line 914: This wasn't true: there was no query dealing with user & groups. Instead, the 2 extra queries that I've removed were in fact caused by license & license translation being fetched for each object. In other words, we had a massive scaling issue on that API.

eviljeff:
> (Fixes #17165 too, in a commit that can be reviewed separately in #17165)

rebase plz?

diox:
rebased

src/olympia/amo/utils.py
```diff
@@ -965,19 +965,31 @@ def path(self, name):
...
 
-    # Get translations in a dict, ids will be the keys. It's important to
-    # consume the result of sorted_groupby, which is an iterator.
-    qs = Translation.objects.filter(id__in=ids, localized_string__isnull=False)
+    if ids:
+        # Get translations in a dict, ids will be the keys. It's important to
+        # consume the result of sorted_groupby, which is an iterator.
+        qs = Translation.objects.filter(id__in=ids, localized_string__isnull=False)
+    else:
+        qs = []
```
eviljeff, line 992: this is just an optimization I assume?  (a filter with `id__in=[]` would resolve to an empty queryset anyway)

eviljeff (approved):

src/olympia/addons/serializers.py
```diff
@@ -293,6 +293,12 @@ class Meta:
         )
 
 
+class VersionSerializerForListing(VersionSerializer):
```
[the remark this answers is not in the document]
diox: Sounds better yeah.. I'll change it.

src/olympia/addons/tests/test_views.py
diox: They were duplicated and miscounted before (you had one `licenses` and one `licenses translations` *for each version*. It just so happened so that were only 2 versions, and the comment misled us into thinking the 2 added queries were for users & groups).

They aren't anymore (you now have one `licenses` and one `licenses translations` *period*, regardless of how many versions are returned, thanks to the transformer), hence the lower number of queries.

src/olympia/amo/utils.py
diox: Yes, avoiding the query in that case.

PR #17167, opened 2021-05-17
save all resized jpegs with a white background

### LINKED ISSUE #17028: Transparency in user generated content should be converted to the same colour as the details page background

eviljeff:
fixes #17028 again


src/olympia/amo/utils.py
```diff
@@ -726,13 +726,12 @@ def resize_image(source, destination, size=None, *, format='png', quality=80):
...
         else:
-            if source_fileext == '.png':
-                # Create a white rgba background for transparency
-                white_background = Image.new('RGBA', im.size, 'WHITE')
-                white_background.paste(im, (0, 0), im)
-                im = white_background
-            im = im.convert('RGB')
-            im.save(dest_file, 'JPEG', quality=quality)
+            # Create a white rgba background for transparency
+            white_background = Image.new('RGBA', im.size, 'WHITE')
```
diox, line 730: Could you rename this, now that it's not just the white background but also the "final" image ?

diox (approved):

src/olympia/amo/utils.py
eviljeff: https://github.com/mozilla/addons-server/pull/17167/commits/eccfab2acad9933634f64a11b92d6534241e9276

PR #17182, opened 2021-05-19
delete fake emails after 3 months

### LINKED ISSUE #17122: Delete fake emails after X months
On -dev, I noticed that we still have fake emails from 2017. We should probably delete them after a few months (maybe 2-3 months or even 1?). @diox suggested to remove these fake emails in the `gc` CRON.

iamvaishnav:
Fixes #17122


diox (changes requested):
Please don't change `jétpack.xpi` and also include a test for your changes.

src/olympia/amo/tests/test_cron.py
```diff
@@ -55,6 +56,16 @@ def test_file_uploads_deletion_oserror(self, storage_mock):
...
 
+    def test_delete_fake_emails(self, storage_mock):
+        fe_old = FakeEmail.objects.create(message='This is the oldest fake email')
+        fe_old.update(created=self.days_ago(360))
+        fe_new = FakeEmail.objects.create(message='This is the newest fake email')
+        fe_new.update(created=self.days_ago(45))
+
+        gc()
+        # FakeEmail which are older than 90 were deleted.
+        assert FakeEmail.objects.count() == 1
```
diox, line 67: Suggestion:

```suggestion
        assert FakeEmail.objects.count() == 1
        assert FakeEmail.objects.filter(pk=fe_new.pk).count() == 1
```

That way we also prove we deleted the right one.

diox (commented):
Could you avoid renaming this file: 
`olympia/files/fixtures/files/jétpack.xpi → src/olympia/files/fixtures/files/jétpack.xpi`

It has a weird unicode filename (combining e and acute accent instead of a straight `é`) on purpose. Some git clients might have trouble with that, you might need to rebase and explicitly tell git to ignore that change.

iamvaishnav:
> Could you avoid renaming this file:
> `olympia/files/fixtures/files/jétpack.xpi → src/olympia/files/fixtures/files/jétpack.xpi`
> 
> It has a weird unicode filename (combining e and acute accent instead of a straight `é`) on purpose. Some git clients might have trouble with that, you might need to rebase and explicitly tell git to ignore that change.

Yeah, I've trouble with that

diox (approved):

PR #17193, opened 2021-05-20
update theme preview to match photon UI redesign

### LINKED ISSUE #10028: Add tab_line picker to theme editor
From: https://github.com/mozilla/addons-frontend/issues/6963

---

Right now, there's no tab_line picker in the theme editor, which seems to be encouraging theme authors to change the accentcolor to change the tab line color, which produces inaccessible themes, since the accentcolor is used as window background.

Here's an example of a bug that this causes: bugzilla.mozilla.org/show_bug.cgi?id=1506913

See also: bugzilla.mozilla.org/show_bug.cgi?id=1508081

@eviljeff Would it be possible to add a "Tab line color" field that changes the tab_line property?

### LINKED ISSUE #16887: Update theme previews to support new Firefox design
There are some changes coming to Firefox that will impact themes. We need to update the theme tool in DevHub and the theme previews we show on the site to follow these changes.

So far, it appears these changes are:
* The tab is no longer attached to the toolbar.
* The tab shape is a rounded rectangle.
* The "tab line" at the top of the tab is no longer there. For themes that declare that property, `tab_line will` define the border color for the tab. The appearance of this border is still being worked on in [bug 1702329](https://bugzilla.mozilla.org/show_bug.cgi?id=1702329).

The new design is expected to go out with Firefox 89 on May 18, but that could change. It'd be great if we can coordinate this update with the release, or slightly earlier.

/cc @caitmuenster

eviljeff:
fixes #16887 + bonus fixes #10028 (it was helpful when debugging to be able to see the tab border highlight Firefox now has)


src/olympia/devhub/views.py
```diff
@@ -1397,6 +1397,17 @@ def submit_version_distribution(request, addon_id, addon):
         ),
         False,
     ),
+    (
+        'tab_line',
+        _('Tab color highlight'),
```
jvillalobos, line 1402: Let's just call it 'Tab highlight', since we don't use 'color' for any of the others (and they're all colors).

src/olympia/devhub/views.py
```diff
@@ -1397,6 +1397,17 @@ def submit_version_distribution(request, addon_id, addon):
...
+        _(
+            'The highlight color of the active tab. Implemented as a '
+            'border around the tab on Firefox 89+ and line above the tab on older '
```
jvillalobos, line 1405: Nit: 'and a line'

src/olympia/devhub/views.py
eviljeff: https://github.com/mozilla/addons-server/pull/17193/commits/1c913fb3ad44201d3190a67f66d064d7a0b755ec

PR #17198, opened 2021-05-21
correctly show account pages per app (android promoted filtering)

### LINKED ISSUE #17142: filter userprofile extension and theme counts in sitemap by addon compat
follow-up for #16960 - addons listed are filtered by app compatibility per page but the sitemap is using an overall count which is incorrect.

### LINKED ISSUE #17153: Some user accounts without actual content are listed in the sitemap
Follow up for #16960

On the last [sitemap page with users](https://addons-dev.allizom.org/sitemap.xml?section=users&app_name=firefox&p=43&debug), the last entries do not have a `lastmod` date and when I've tried to open some of those profiles I've found:
1. Users without content (extensions or themes), profile page is loaded:
https://addons-dev.allizom.org/en-US/firefox/user/10641975/
https://addons-dev.allizom.org/en-US/firefox/user/7725178/
2. Profile page returns a 404 on the frontend(*):
https://addons-dev.allizom.org/en-US/firefox/user/10642071/
https://addons-dev.allizom.org/en-US/firefox/user/10642003/

(*) for this set, django admin shows these users as Deleted, but the 'Is Public' flag is still on. I'm not sure why the flag hasn't changed when those users were deleted, or banned.

eviljeff:
fixes #17142 (+ fixes #17153)

more efficient suggestions to the two query solution welcome

### DIFF
--- a/src/olympia/amo/sitemap.py
+++ b/src/olympia/amo/sitemap.py
@@ -18,7 +18,9 @@ from olympia.addons.models import Addon, AddonCategory
 from olympia.amo.reverse import get_url_prefix, override_url_prefix
 from olympia.amo.templatetags.jinja_helpers import absolutify
 from olympia.constants.categories import CATEGORIES
+from olympia.constants.promoted import RECOMMENDED
 from olympia.bandwagon.models import Collection
+from olympia.promoted.models import PromotedAddon
 from olympia.users.models import UserProfile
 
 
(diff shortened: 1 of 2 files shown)


diox (approved):
This shouldn't generate an extra query, just an extra join that makes the query rather large and complex. 

In prod the query takes 10-15 seconds (+ a count that takes the same amount, when paginating), but doesn't seem to get worse as we go to the end of the result list (with like `LIMIT 1000 OFFSET 234537).

eviljeff:
> + a count that takes the same amount, when paginating)

mmm, why would there be an extra query for a count when paginating, when `_cached_items` is a `@cached_property`?

Issue #4161, opened 2016-12-20
svg support for Add-on icon


Jan02:
add svg support for Add-on icon

EnTeQuAk:
We were planning on implementing a better image handling some time. Marking as "needs more triage" to open the question if SVG support should be something we should do and maybe open a tracking issue for issues like that?

muffinresearch:
Adding SVG support would be nice, we'd need to be careful around sanitization though.

andymckay:
With WebExtensions its pretty cool because you can use an SVG everywhere, browserAction, pageAction, icons and so on. Everywhere except AMO where I have to make a special PNG out of my image.

Also SVG icons scale a lot better, so presumably would make a lot of sense for usability.

jvillalobos:
For the Q4 investigation, here are the questions I'd like answered:
* What is necessary to support SVG icons on AMO, both from the XPI or separate upload. Is it just a matter of using DOM Purify, or is there more to it?
* Are there any restrictions developers need to know about for their SVG files to work?

diox:
> Adding SVG support would be nice, we'd need to be careful around sanitization though.

https://github.com/cloudflare/svg-hush might be interesting to look at if we ever want to do this.

PR #17200, opened 2021-05-21
update default icon to simplified photon style graphic

### LINKED ISSUE #17080: Update generic extension icon
Firefox 89 will update the generic icon for add-ons that don't define an icon. We should update the equivalent icons on AMO so Firefox and AMO are in sync. The AMO icons are https://addons.cdn.mozilla.net/static/img/addon-icons/default-64.png and friends. 

Please see https://phabricator.services.mozilla.com/D114008 for context. The new icon is [here](https://searchfox.org/mozilla-central/source/toolkit/themes/shared/extensions/extensionGeneric.svg). That icon is optimized for display at 16px. We also have a slightly different icon for display at 20px:
```
<!-- This Source Code Form is subject to the terms of the Mozilla Public
   - License, v. 2.0. If a copy of the MPL was not distributed with this
   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" width="20" height="20" fill="context-fill" fill-opacity="context-fill-opacity">
  <path d="M15.5 5 13 5l0-2.5a2.5 2.5 0 1 0-5 0L8 5 4.5 5A2.5 2.5 0 0 0 2 7.5l0 1.9.6.6 1.893 0a2 2 0 0 1 0 4L2.6 14l-.6.6 0 2.9A2.5 2.5 0 0 0 4.5 20l11 0a2.5 2.5 0 0 0 2.5-2.5l0-10A2.5 2.5 0 0 0 15.5 5z"/>
</svg>
```
cc @muffinresearch

eviljeff:
fixes #17080 - I included the source .svg too and if we get around to #4161 we could serve it (which would be better overall)


bobsilverberg (approved):
I don't see any code changes that suggest that we don't need `-128`, `-32`, and `-64` anymore, and that `default.svg` will be used instead, but I trust that all of that is correct.

eviljeff:
> I don't see any code changes that suggest that we don't need `-128`, `-32`, and `-64` anymore, and that `default.svg` will be used instead, but I trust that all of that is correct.

It's the other way around - we _do_ still need the -32, -64, and -128 files, and `default.svg` isn't used (yet).

Issue #17143, opened 2021-05-12
only include firefox compatible addons under /firefox/ in sitemap? (& for android)


eviljeff:
follow-on from #16910

the sitemap currently includes both /firefox/addon/<slug> and /android/addon/<slug> for every add-on, regardless of compatibility.  Both urls work, and render a slightly different page - you get a warning banner if the add-on is not compatible with desktop Firefox on /firefox/, and similarly with /android/.

We _don't_ mark an incompatible page as an alternative of the compatible one; or tag it with `<meta name="robots" content="noindex, follow" />` to de-index it; or link to the other page.  It's assumed you landed on the page intentionally as you visited the link from AMO where the app would have been auto-detected. 

But, should we exclude from the sitemap /firefox/ versions of add-ons that aren't compatible with desktop Firefox?  And should we exclude /android/ versions of add-ons that aren't compatible with Firefox on Android?

back in 2018 @rraue cautioned against excluding android urls in https://github.com/mozilla/addons-frontend/issues/6644#issuecomment-430215919 but maybe this isn't still valid (or isn't relevant to the sitemap)

eviljeff:
current proposal is:
- for addons that are only compatible with desktop Firefox, only include the `/en-US/firefox/addon/<slug>/` urls
  - don't include `/en-US/android/addon/<slug>/`
- for addons that are only compatible with Android Firefox - and are recommended - only include `/en-US/android/addon/<slug>/`
  - don't include `/en-US/firefox/addon/<slug>/`
  - don't include the addon for either `/firefox/` or `/android/` if the add-on is Android only but is _not_ recommended, because only recommended add-ons are installable on release Android Firefox.
- for addons that are compatible with desktop Firefox, and Android Firefox, and are recommended for Android, include both `/en-US/firefox/addon/<slug>/` and `/en-US/android/addon/<slug>/`

does this seem reasonable @rraue ?

(I'm only mentioning `/en-US/` urls for brevity; this would equally apply to urls under other locales too)

rraue:
This seems exactly right. We want only urls that are useful for a user indexed and getting traffic to. And our sitemap is basically the reference for that. So I think you got it right here.

AlexandraMoga:
I've created the following set of add-ons in order to verify this issue:

1. [Add-on compatible only with Firefox for Desktop](https://addons-dev.allizom.org/en-US/firefox/addon/only-with-firefox-for-desktop/)
[...]

PR #17247, opened 2021-06-01
enable sitemap generation cron

### LINKED ISSUE #16906: add `write_sitemaps` to crontab
follow-up for #16508 

when we're ready and happy to have the sitemaps generated on a schedule we need to add `write_sitemaps` to `crontab.tpl`

eviljeff:
fixes #16906 - the timing is arbitrary, I just picked a time after the stats crons finish.  (Might be too late in the day then?)

### DIFF
--- a/scripts/crontab/crontab.tpl
+++ b/scripts/crontab/crontab.tpl
@@ -28,7 +28,7 @@ HOME=/tmp
 30 6 * * * %(z_cron)s update_addon_hotness
 30 9 * * * %(z_cron)s update_user_ratings
 0 22 * * * %(z_cron)s gc
-45 14 * * * %(z_cron)s write_sitemaps
+0 23 * * * %(z_cron)s write_sitemaps
 
 # Once per day after metrics import is done
 30 12 * * * %(z_cron)s update_addon_weekly_downloads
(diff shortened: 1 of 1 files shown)


diox:
Since is this going to take a while, might want to give ops a heads up about it (and see if the time matters for them)

eviljeff:
@bqbn to r+ the time the cron runs

eviljeff:
We probably want the PR for #17143 to be merged before this, too (or we'll be generating a lot more sitemap than necessary)

bqbn:
> @bqbn to r+ the time the cron runs

For my info, does it need to run during a certain time window in a day, or any time in a day works?

One comment is to put this line above  the `0 22 * * * %(z_cron)s gc` line, so these few lines are ordered by time when they run.

eviljeff:
> > @bqbn to r+ the time the cron runs
> 
> For my info, does it need to run during a certain time window in a day, or any time in a day works?

any time of day - ideally when there's no much else going on.

bqbn:
> any time of day - ideally when there's no much else going on.

Then I'd recommend 23:00 UTC, which is a good low traffic time.

bobsilverberg (approved):

PR #17257, opened 2021-06-02
Add DailyUserAddonUploadThrottle

willdurand:
fixes https://github.com/mozilla/addons-server/issues/17261

I renamed some classes given that we want hourly and daily limits.


diox (approved):
r+issue :)

PR #17263, opened 2021-06-02
Add query parameter to cachebust default icon

### LINKED ISSUE #17262: Cache bust default icon
I don't see the new icon on -dev (or -stage):

Screenshot 2021-06-02 at 18 13 53

This is because the default icon URL does not have a `modified` query parameter (or any other param) for cache busting. We should have something because even a hard refresh does not fix this problem.

willdurand:
Fixes #17262

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -1090,7 +1090,9 @@ class Addon(OnChangeMixin, ModelBase):
             return jinja_helpers.user_media_url('addon_icons') + path
 
     def get_default_icon_url(self, size):
-        return '{0}img/addon-icons/{1}-{2}.png'.format(
+        # We don't update the default icon very frequently so there is no
+        # automated query parameter for cachebusting...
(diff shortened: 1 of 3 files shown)


src/olympia/addons/models.py
```diff
@@ -1090,7 +1090,9 @@ def get_icon_url(self, size, use_default=True):
             return jinja_helpers.user_media_url('addon_icons') + path
 
     def get_default_icon_url(self, size):
-        return '{0}img/addon-icons/{1}-{2}.png'.format(
+        # We don't update the default icon very frequently so there is no
+        # automated query parameter for cachebusting...
+        return '{0}img/addon-icons/{1}-{2}.png?v=20210602'.format(
```
eviljeff, line 1095: nit, it was merged yesterday so would be `20210601`

eviljeff (approved):
one of the CI jobs has failed - likely an intermittent (haven't looked)

PR #17272, opened 2021-06-03
Add '@mozillafoundation.org' to patterns for system guids

### LINKED ISSUE #17271: Add some more GUID patterns to system guid list
There is at least one privileged extension with an ID ending in "@mozillafoundation.org", so we should add that pattern to the `SYSTEM_ADDON_GUIDS` list.

Edit: We need to add `@rally.mozilla.org` as well for future rally studies.

wagnerand:
Fix #17271.

### DIFF
--- a/src/olympia/constants/base.py
+++ b/src/olympia/constants/base.py
@@ -371,6 +371,8 @@ SYSTEM_ADDON_GUIDS = (
     '@shield.mozilla.com',
     '@shield.mozilla.org',
     '@mozillaonline.com',
+    '@mozillafoundation.org',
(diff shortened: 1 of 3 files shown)


eviljeff (approved):
yeah... I'd argue that adding new domains to the list of system guids isn't helpful for anyone involved - there's no reason @mozilla.org couldn't have been used - but if the privileged addon is already out there then there isn't much we (addons) can do about it so 🤷‍♂️

wagnerand:
I agree with you. We are now also putting safeguards in place that protect this list from growing further (without knowing beforehand).

PR #17278, opened 2021-06-04
Add ability to set restrictions that affect approval

diox:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1713805


src/olympia/scanners/tests/test_actions.py
```diff
@@ -82,15 +87,25 @@ def test_delay_auto_approval_indefinitely_and_restrict(self):
...
-        assert EmailUserRestriction.objects.filter(email_pattern=user3.email).exists()
+        assert EmailUserRestriction.objects.filter(
+            email_pattern=user1.email, restriction_type=RESTRICTION_TYPES.SUBMISSION
+        ).exists()
+        assert EmailUserRestriction.objects.filter(
+            email_pattern=user2.email, restriction_type=RESTRICTION_TYPES.SUBMISSION
+        ).exists()
+        assert EmailUserRestriction.objects.filter(
+            email_pattern=user3.email, restriction_type=RESTRICTION_TYPES.SUBMISSION
+        ).exists()
```
bobsilverberg, line 98: Should we also be checking that we don't have an `EmailUserRestriction` with a `restriction_type` of `RESTRICTION_TYPES.APPROVAL`, or is that unnecessary for this test? The same goes for `IPNetworkUserRestriction`.

src/olympia/users/tests/test_models.py
```diff
@@ -795,26 +795,83 @@ def test_blocked_user_login_ip(self):
...
         IPNetworkUserRestriction.objects.create(network='192.168.1.1/32')
-        assert not IPNetworkUserRestriction.allow_request(request)
+        assert not IPNetworkUserRestriction.allow_submission(request)
+
+    def test_allowed_approval_while_blocking_submission(self):
+        request = RequestFactory(REMOTE_ADDR='192.168.0.1').get('/')
+        request.user = user_factory(last_login_ip='10.0.0.1')
+        IPNetworkUserRestriction.objects.create(network='192.168.0.0/28')
+        # Submission is allowed.
+        assert not IPNetworkUserRestriction.allow_submission(request)
```
bobsilverberg, line 805: The comments say `# Submission is allowed.` and `# Approval is blocked.`, but the assertions seems to be the opposite, and the name of the test also indicates that we expect the opposite. I am guessing the comments are wrong?

src/olympia/users/tests/test_models.py
```diff
@@ -795,26 +795,83 @@ def test_blocked_user_login_ip(self):
...
+        # Approval is blocked.
+        upload = FileUpload.objects.create(ip_address='192.168.0.1', user=request.user)
+        assert IPNetworkUserRestriction.allow_auto_approval(upload)
+
+    def test_blocked_approval_last_login_ip(self):
+        request = RequestFactory(REMOTE_ADDR='192.168.0.1').get('/')
+        request.user = user_factory(last_login_ip='192.168.0.1')
+        IPNetworkUserRestriction.objects.create(
+            network='192.168.0.0/28', restriction_type=RESTRICTION_TYPES.APPROVAL
```
bobsilverberg, line 814: Does the restriction on `192.168.0.0` also encompass `192.168.0.1`? If not, I don't follow how adding the restriction on `192.168.0.0` actually blocks this at all? If it does, then the test data on line 820 seems invalid because `192.168.0.2` would be blocked anyway.

src/olympia/users/tests/test_models.py
```diff
@@ -795,26 +795,83 @@ def test_blocked_user_login_ip(self):
...
+        # Submission remains allowed.
+        assert DisposableEmailDomainRestriction.allow_submission(request)
+        # Approval is blocked.
+        upload = FileUpload.objects.create(ip_address='192.168.0.1', user=request.user)
+        assert not DisposableEmailDomainRestriction.allow_auto_approval(upload)
+
+    def test_allowed_approval(self):
+        request = RequestFactory().get('/')
+        request.user = user_factory(email='foo@bar.com')
+        DisposableEmailDomainRestriction.objects.create(domain='bar.com')
```
bobsilverberg, line 869: Nit: I realize that this shows up in a number of other places, but for some reason it hit me when reading this test: It seems a bit unintuitive that adding a restriction by default creates a submission restriction, and that in order to create an approval restriction one has to specify `restriction_type=RESTRICTION_TYPES.APPROVAL`. I know it's convenient, but it seems like it would be better/safer to always require a `restriction_type` when creating a restriction so it's obvious and explicit what's being done.

bobsilverberg (approved):
Phew, that was a bit patch! I guess a lot of it was just moved from one place to another, but I didn't realize that when I started reviewing the files from top to bottom. 😛 

I added a bunch of comments, but I think they are mostly things that you can ignore if you don't agree, so I'm calling this an r+wc.

src/olympia/addons/tests/test_utils_.py
```diff
@@ -128,3 +139,234 @@ def test_is_outcome_recommended(self):
...
+class TestRestrictionChecker(TestCase):
+    def setUp(self):
+        self.request = RequestFactory(REMOTE_ADDR='10.0.0.1').get('/')
+        self.request.is_api = False
+        self.request.user = user_factory(read_dev_agreement=self.days_ago(0))
+        self.request.user.update(last_login_ip='192.168.1.1')
+
+    def test_is_submission_allowed_pass(self, incr_mock):
+        checker = RestrictionChecker(request=self.request)
+        assert checker.is_submission_allowed()
```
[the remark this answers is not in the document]
diox: I try to strike a good balance between reducing duplication while staying explicit. We have a bunch of older tests where we go through a number of common methods/function/classes to set up stuff and it becomes hard to read... So in general for newer stuff I will often duplicate the test stuff, and sometimes the initialization of the class/method I'm testing, and move the rest to a `setUp` or helper.

src/olympia/addons/tests/test_utils_.py
```diff
@@ -128,3 +139,234 @@ def test_is_outcome_recommended(self):
...
+        checker = RestrictionChecker(request=self.request)
+        assert checker.is_submission_allowed(check_dev_agreement=False)
+        assert incr_mock.call_count == 1
+        assert incr_mock.call_args_list[0][0] == (
+            'RestrictionChecker.is_submission_allowed.success',
+        )
+        assert not UserRestrictionHistory.objects.exists()
+
+    def test_user_is_allowed_to_bypass_restrictions(self, incr_mock):
+        IPNetworkUserRestriction.objects.create(network='10.0.0.0/24')
```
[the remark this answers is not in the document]
diox: Technically it's also missing a test against the reputation stuff, but yeah, the test is assuming the restrictions all work the same. I'll add an `EmailUserRestriction`, it doesn't hurt (the reputation stuff start requiring more mocks)

src/olympia/addons/tests/test_utils_.py
```diff
@@ -128,3 +139,234 @@ def test_is_outcome_recommended(self):
...
+            'RestrictionChecker.is_submission_allowed.failure',
+        )
+        assert UserRestrictionHistory.objects.count() == 1
+        history = UserRestrictionHistory.objects.get()
+        assert history.get_restriction_display() == 'EmailUserRestriction'
+        assert history.user == self.request.user
+        assert history.last_login_ip == self.request.user.last_login_ip
+        assert history.ip_address == '10.0.0.1'
+
+    def test_is_submission_allowed_bypassing_read_dev_agreement_restricted(
```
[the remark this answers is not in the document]
diox: It's a subtle case: as it tries to explain (poorly) in the comment, what we care about is that bypassing the dev agreement check, regardless of what `read_dev_agreement` is for that user, doesn't affect the *other* restrictions. That we're only bypassing that single check, and are correctly hitting the other restriction.

There is an other test (referenced in that comment) specifically about the `read_dev_agreement` being `None` when we use `check_dev_agreement=False` (`test_is_submission_allowed_bypassing_read_dev_agreement`)

src/olympia/addons/utils.py
```diff
@@ -123,3 +127,123 @@ def compute_last_updated(addon):
...
+        if self.user and self.user.bypass_upload_restrictions:
+            return True
+
+        if check_dev_agreement is False:
+            restriction_choices = filter(
+                lambda item: item[1] != DeveloperAgreementRestriction,
+                self.restriction_choices,
+            )
+        else:
+            restriction_choices = None
```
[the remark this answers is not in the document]
diox: Yes, I was treating that variable as a `const` in my mind :)

src/olympia/scanners/tests/test_actions.py
diox: Might as well

src/olympia/users/tests/test_models.py
diox: Yes, I'll fix

src/olympia/users/tests/test_models.py
diox: `192.168.0.0/28` should affect `192.168.0.1` to `192.168.0.14`. You're right that `192.168.0.2` should be changed..

src/olympia/users/tests/test_models.py
diox: Yeah, I did that for backwards-compatibility, and was too lazy to go back and change all the existing tests

Issue #17302, opened 2021-06-08
Normalize `manifest.json` file


willdurand:
We would like to normalize the `manifest.json` file at upload time to make sure it is correctly formatted. The normalization will remove comments, too. We think that this step could be performed in the "repack" task.

ioanarusiczki:
@willdurand I uploaded [Search by image ](https://addons-dev.allizom.org/en-US/firefox/addon/search-by-image-test3/)- as a test on AMO dev - and it passed validation.

The manifest.json format contained all the information in a row https://code.addons-dev.allizom.org/en-US/browse/620598/versions/1692763/ plus I added a comment "_comment": "comment text goes here..."

Nothing changed in manifest.json as format and the comment is available as shown by codemanager.

I don't know if that's expected. If I should be testing more around it , let me know.

willdurand:
@ioanarusiczki could you retry now? The feature was not deployed and not enabled.

ioanarusiczki:
@willdurand I tried again and the manifest is looking now like this: https://code.addons-dev.allizom.org/en-US/browse/620599/versions/1692765/ which is different than before - looks good

I can see the comment on line 2.

ioanarusiczki:
@willdurand 
1. Switch is active now on stage - I tested again with //comment. These are deleted after uploads.

Example: 
https://code.addons-dev.allizom.org/en-US/browse/620603/versions/1692771/

2. I tried adding a comment into a manifest with all data in a row:

The linter on stage throws this error: 

![message on stage](https://user-images.githubusercontent.com/33448286/121500003-91c1f280-c9e6-11eb-8f0d-d69c27b69abb.png)

On production the message is:

![linter on prod](https://user-images.githubusercontent.com/33448286/121500170-bcac4680-c9e6-11eb-95f5-62fecb4d431c.png)

I believe it is not expected to pass but the message has changed.

willdurand:
> I tried adding a comment into a manifest with all data in a row:

Can you share the content of the file? I don't understand.

ioanarusiczki:
@willdurand 

I modified the json like that: 

![comment](https://user-images.githubusercontent.com/33448286/121507950-ca18ff00-c9ed-11eb-870e-032446645477.png)

willdurand:
ah ok, well this is clearly not valid so I guess that makes sense.

ioanarusiczki:
@willdurand Yes, what I said is that the linter message is different for the same error on stage versus prod. 

[...]

PR #17303, opened 2021-06-08
Normalize manifest.json file in the repack task

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/17302


src/olympia/files/tests/test_tasks.py
```diff
@@ -76,6 +78,68 @@ def test_repacking_xpi_files(self):
...
+        fake_results = {'errors': 0}
+        with zipfile.ZipFile(upload.path, 'w') as z:
+            manifest_with_comments = """
+            {
+                // Required
+                "manifest_version": 2,
+                "name": "My Extension",
+                "version": "versionString",
+                // Recommended
+                "description": "haupt\u005fstra\u00dfe"
```
diox, line 117: Same here. In fact, as it is, this test is checking that entering `haupt_straße` normalizes to `haupt_stra\u00dfe` in the JSON. Ideally we'd test instead that entering `haupt\u005fstra\u00dfe` in the JSON normalizes to `haupt_stra\u00dfe`, removing the superfluous encoding for the ascii character. Doubling the backslashes should do that.

src/olympia/files/tests/test_tasks.py
```diff
@@ -76,6 +78,68 @@ def test_repacking_xpi_files(self):
...
```
diox, line 93: I don't think that's a good test. Python will parse those unicode escapes, you gotta double backslash them to keep them as-is in JSON.

eviljeff: or use `r"..."` strings

diox (approved):

Issue #17311, opened 2021-06-09
investigate speeding up generation of a sitemap files


eviljeff:
currently to create and write a single file for addons in the sitemap takes ~0.9 seconds.  This sounds a small number but ends up taking hours when there are thousands in total.  

Possible changes:
- ~write directly to a file object in `Sitemap.render_xml`~ not possible
- [x] make template loading more efficient - PR #17320 (didn't much difference on addons-dev)
- [x] drop some items by not including some urls (e.g. #17289 drops /versions/ items)
- [x] bump up the 1000 limit per file to, say, 2000 -  this would obviously mean a render would take longer for a single file, but likely not double, so overall a speed increase.
- [ ] Reduce the number of languages we're supporting - currently we're generating links for the top 10 locales. If we reduced this to, say, 5, we'd have 50% less items, but each item would also have less 5 alternates (i.e. only 6 urls in the xml rather than 11) , so the actual filesize would decrease by 72.5%

eviljeff:
Other choices:
- Reduce the number of languages we're supporting - currently we're generating links for the top 10 locales.  If we reduced this to, say, 5, we'd have 50% less items, but each item would also have less 5 alternates (i.e. only 6 urls in the xml rather than 11) , so the actual filesize would decrease by 72.5%

eviljeff:
(reworded the description to summarize our options.)

eviljeff:
QA: We didn't need to reduce the number of languages in the end, so the changes this issue represents should be transparent, apart from increasing the number of items per xml file to 2000.  (https://github.com/mozilla/addons-server/issues/17289 was separately verified)

AlexandraMoga:
I browsed throughout the various sitemap sections and I didn't find other changes, apart from the new 2000 item limit.

PR #17320, opened 2021-06-10
attempt to optimize sitemap generation by caching loading template

eviljeff:
This is part of #17311 - there wasn't a way to directly write to file (that I could see) so I just optimized the template loading (now we do it once for each Sitemap subclass, rather than for each page) for some presumed speed increase.


diox (approved):
That sounds promising...

PR #17323, opened 2021-06-10
Fix ES reindex tests with Django 3.2 

diox:


eviljeff:
I don't think we want to merge this and disable all the other CI jobs 😬

diox:
Good point :)

PR #17342, opened 2021-06-14
only reverse urls once per sitemap item

### LINKED ISSUE #17311: investigate speeding up generation of a sitemap files

eviljeff:
fixes #17311 - this patch moves the `reverse` into `_cached_items` for `AddonSitemap` and `AccountSitemap` (the two costly sitemaps) and cheats with locales by just doing a string replace on the locale in the url, avoiding `override_url_prefix`.  

Both of these things are fine normally but at scale (22000 times for a page of 2000 items) take ~90% of the time for the data for a page to be generated (at least during experimentation)

### DIFF
--- a/src/olympia/amo/sitemap.py
+++ b/src/olympia/amo/sitemap.py
@@ -234,11 +234,13 @@ class Sitemap(DjangoSitemap):
         super().__init__(*args, **kwargs)
 
     def _location(self, item, force_lang_code=None):
+        # modified from Django implementation - we don't rely on locale for urls
         if self.i18n:
             obj, lang_code = item
-            # modified from Django implementation - we don't rely on locale for urls
-            with override_url_prefix(locale=(force_lang_code or lang_code)):
-                return self.location(obj)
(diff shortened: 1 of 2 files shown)


bobsilverberg (approved):
Looks good. Nice job figuring out how to speed up this task!

PR #17350, opened 2021-06-15
enhance cron logging to show time taken

eviljeff:


src/olympia/amo/management/commands/cron.py
```diff
@@ -48,14 +48,17 @@ def handle(self, *args, **options):
 
         module = import_module(path)
 
-        current_millis = datetime.now().timestamp() * 1000
-
+        start_millis = datetime.now().timestamp() * 1000
```
bobsilverberg, line 51: Nit: I would have just called it `start_ms` but 🤷

bobsilverberg (approved):
Just a small nit.

r+wc

src/olympia/amo/management/commands/cron.py
eviljeff: I adjusted based on the current naming but I agree

PR #17352, opened 2021-06-15
Do not normalize already signed add-ons

### LINKED ISSUE #17344: Manifest normalization breaks signature of privileged add-ons
Versions of privileged add-ons (and other add-ons that are already signed when submitted to AMO) might be broken because of the manifest normalization we introduced in #17302. We should skip normalization for these add-ons.

willdurand:
Fixes #17344

### DIFF
--- a/src/olympia/files/tasks.py
+++ b/src/olympia/files/tasks.py
@@ -15,7 +15,12 @@ from olympia.amo.storage_utils import move_stored_file
 from olympia.amo.utils import StopWatch
 from olympia.devhub.tasks import validation_task
 from olympia.files.models import File, FileUpload
-from olympia.files.utils import extract_zip, get_sha256, ManifestJSONExtractor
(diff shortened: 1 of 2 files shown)


src/olympia/files/tasks.py
```diff
@@ -40,9 +45,13 @@ def repack_fileupload(results, upload_pk):
                 extract_zip(upload.path, tempdir=tempdir)
 
                 if waffle.switch_is_active('enable-manifest-normalization'):
+                    xpi_data = parse_xpi(upload.path, minimal=True)
```
diox, line 48: That will cause us to parse the manifest twice, which is a bit unfortunate. We could fix this with a lot of refactoring (ideally `parse_xpi` would give us access to the manifest data so we don't have to call `ManifestJSONExtractor` later), but it doesn't block deploying this.

diox (approved):

PR #17354, opened 2021-06-15
render sitemap.xml with jinja2 for speed

### LINKED ISSUE #17353: render template xml with jinja2
We currently force the sitemap.xml to be rendered with django template engine as the file is the exact version in django3.2 and there are minor syntax differences between jinja2 and django flavor templates.  Jinja2 is faster though, and the template rendering time is the last big chunk of time in sitemap generation we would like to optimize (down from ~0.35s per 2000 item addon sitemap file)

eviljeff:
fixes #17353

### DIFF
--- a/src/olympia/amo/templatetags/jinja_helpers.py
+++ b/src/olympia/amo/templatetags/jinja_helpers.py
@@ -10,12 +10,16 @@ from django.forms import CheckboxInput
 from django.template import defaultfilters, Library, loader
 from django.urls import reverse
 from django.utils.encoding import smart_str
-from django.utils.html import format_html as django_format_html
+from django.utils.html import (
+    format_html as django_format_html,
+    strip_spaces_between_tags,
+)
 from django.utils.safestring import mark_safe
 from django.utils.translation import get_language, to_locale, trim_whitespace, gettext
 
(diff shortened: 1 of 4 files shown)


src/olympia/amo/templatetags/jinja_helpers.py
```diff
@@ -447,3 +451,20 @@ def format_datetime(value, format='DATETIME_FORMAT'):
 def class_selected(a, b):
     """Return ``'class="selected"'`` if ``a == b``."""
     return mark_safe('class="selected"' if a == b else '')
+
+
+class Spaceless(Extension):
```
diox, line 456: A couple basic tests covering this extension in `src/olympia/amo/tests/test_helpers.py` would be nice

diox (approved):

src/olympia/amo/templatetags/jinja_helpers.py
eviljeff: https://github.com/mozilla/addons-server/pull/17354/commits/a0e7556a6055449ec35da523b24114605c266c67 its just one test, but I tried to cover what it does and the edge cases.

Issue #17370, opened 2021-06-17
'Unexpected error' during validation if manifest is not at the root of the extension 


AlexandraMoga:
### Describe the problem and steps to reproduce it:
1. Prepare an extension archive where the the `manifest.json` file is not at the root - for example by creating an archive from the extension directory rather than archiving the files directly
2. Submit the add-on on -dev or stage

### What happened?
The following validation error is triggered: "Error: Validation was unable to complete successfully due to an unexpected error."

### What did you expect to happen?
The following error: "Error: No manifest.json was found at the root of the extension. The package file must be a ZIP of the extension's files themselves, not of the containing directory. See: https://mzl.la/2r2McKv for more on packaging."

### Anything else we should know?
Not reproduced on prod

[manifest-root.zip](https://github.com/mozilla/addons-server/files/6668109/manifest-root.zip)

willdurand:
> Not reproduced on prod

mm so that rules out the manifest normalization I guess?

willdurand:
Oh wait, no. It's because we're using `parse_xpi()` but it should only be used if there is a manifest file at the root of the xpi.

AlexandraMoga:
This issue is verified fixed on -dev:

![image](https://user-images.githubusercontent.com/31961530/122929901-211abe80-d374-11eb-98e5-d88941ffdd4a.png)

PR #17372, opened 2021-06-17
Skip manifest normalization when it fails

### LINKED ISSUE #17369: JSON syntax errors in the manifest will trigger an "unexpected error" during validation
1. Edit an add-on manifest to purposely introduce a syntax error (like an extra `,` , `}` or a missing `"` etc)
2. Submit the add-on on AMO

### What happened?
The following validation error is triggered: Error: Validation was unable to complete successfully due to an unexpected error.

Validation results point to the JSON syntax error

- issue reproduced in all AMO environments 
- most likely a regression from the manifest normalization feature

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/17370
Fixes #17369

### DIFF
--- a/src/olympia/files/tasks.py
+++ b/src/olympia/files/tasks.py
@@ -45,19 +45,18 @@ def repack_fileupload(results, upload_pk):
                 extract_zip(upload.path, tempdir=tempdir)
 
                 if waffle.switch_is_active('enable-manifest-normalization'):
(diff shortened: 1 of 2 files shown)


diox (approved):
It's unfortunate we have to parse the manifest multiple times, but it's probably not worth refactoring the `parse_xpi` stuff to avoid that, at least not yet...

PR #17377, opened 2021-06-17
force truncate column contents in shelves migrations 0016

eviljeff:
fix migration in #17376


willdurand (approved):

willdurand:
merging now to fix -dev

PR #17437, opened 2021-07-06
Store collections description as text, not HTML

### LINKED ISSUE #7648: Collection API returns encoded HTML entities

STR: Create a collection with the `&` character in the description. See https://github.com/mozilla/addons-frontend/issues/4375

### What happened?

The API endpoint returns HTML encoded entities such as `&amp;`. Example:

```
curl 'https://addons-dev.allizom.org/api/v3/accounts/account/kumar-krsti%C4%87/collections/small-collection/?lang=en-US' | jq .description
"Things &amp; stuff and “quotes”\n\n£1,000\n\n&lt;not-a-tag&gt;"
```

Well, I'm not sure. If this is expected then we can deal with it on the frontend. I thought to file a bug because the reviews endpoint is not returning encoded entities. Example:

```
$ curl 'https://addons-dev.allizom.org/api/v3/reviews/review/?addon=sound-control&api=&filter=without_empty_body' | jq '.results[0].reply.body'
"This is a reply & stuff"
```

Maybe the reviews endpoint should be returning encoded HTML entities?

diox:
URLs should still be stripped, but we should store (and return) text, and let clients escape it as needed - it's not supposed to contain
any HTML.

Includes a command to fix existing translations in the database.

Fixes #7648

### DIFF
--- a/src/olympia/bandwagon/models.py
+++ b/src/olympia/bandwagon/models.py
@@ -30,9 +30,9 @@ class Collection(ModelBase):
     name = TranslatedField(require_locale=False)
     slug = models.CharField(max_length=30, blank=True, null=True)
 
-    # biography can (and sometimes does) contain html and other unsanitized
+    # description can (and sometimes does) contain html and other unsanitized
     # content. It must be cleaned before display - NoURLsField just strips the
-    # URL without escaping.
+    # URL without doing any escaping.
     description = NoURLsField(require_locale=False)
     default_locale = models.CharField(
         max_length=10, default='en-US', db_column='defaultlocale'
--- a/src/olympia/bandwagon/models.py
+++ b/src/olympia/bandwagon/models.py
@@ -10,7 +10,7 @@ from olympia.amo.fields import PositiveAutoField
 from olympia.amo.models import ManagerBase, ModelBase
 from olympia.translations.fields import (
     LinkifiedField,
-    NoLinksNoMarkupField,
+    NoURLsField,
     TranslatedField,
     save_signal,
 )
@@ -30,7 +30,10 @@ class Collection(ModelBase):
     name = TranslatedField(require_locale=False)
     slug = models.CharField(max_length=30, blank=True, null=True)
 
-    description = NoLinksNoMarkupField(require_locale=False)
+    # biography can (and sometimes does) contain html and other unsanitized
+    # content. It must be cleaned before display - NoURLsField just strips the
+    # URL without escaping.
+    description = NoURLsField(require_locale=False)
     default_locale = models.CharField(
         max_length=10, default='en-US', db_column='defaultlocale'
     )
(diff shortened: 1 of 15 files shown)


src/olympia/bandwagon/models.py
```diff
@@ -30,7 +30,10 @@ class Collection(ModelBase):
     name = TranslatedField(require_locale=False)
     slug = models.CharField(max_length=30, blank=True, null=True)
 
-    description = NoLinksNoMarkupField(require_locale=False)
+    # biography can (and sometimes does) contain html and other unsanitized
```
bobsilverberg, line 33: Should this comment say "description" as opposed to "biography"?

src/olympia/translations/tests/test_commands.py
```diff
@@ -0,0 +1,70 @@
...
+        addon.name.save()
+        # Translation belonging to a collection description that we don't need to bother
+        # with (because its localized string clean should already match its localized
+        # string): should not be touched.
+        extra_collection = Collection.objects.create(
+            name='foo_collection_name2', description='foo_collection_description2'
+        )
+        extra_collection.description.update(modified=self.days_ago(42))
+
+        # The command should fix this one.
```
bobsilverberg, line 44: Just for ease of readability of the test, I would suggest moving this block up right below where you create `self.collection`. Otherwise one has to scroll up to verify that `localized_string_clean` is in fact different from `description` in `self.collection`.

src/olympia/translations/tests/test_models.py
```diff
@@ -734,38 +734,103 @@ def test_forbidden_tags(self):
...
+            translation.localized_string_clean == translation.localized_string == 'foo'
+        )
+
+        translation = NoURLsTranslation.objects.create(
+            id=1003,
+            localized_string='foo http://example.com',
+            localized_string_clean='foo',
+        )
+        assert translation.localized_string == 'foo http://example.com'
+        assert translation.localized_string_clean == 'foo'
```
bobsilverberg, line 827: Is this testing that `localized_string` has been cleaned, and the URL removed, or is this testing that the value provided for `localized_string_clean` ends up in `translation.localized_string_clean`? I believe it is the former, and if that's the case, perhaps it would be better if the original value passed into `localized_string_clean` is something other than `'foo'`.

bobsilverberg (approved):
Looks good, nice work @diox. I added a few comments, but they are pretty small/simple things, so r+wc.

src/olympia/bandwagon/models.py
diox: Oops, yeah

src/olympia/translations/tests/test_models.py
diox: 4313e1e2adb4fb2e69fc176537a2b31026783000

src/olympia/translations/tests/test_commands.py
diox: 7afa8a5dbaf2b10dd69842e110a44c2b35fe11e7

src/olympia/bandwagon/models.py
diox: 649f7fb8b508b50e05c1644fba2e69124a565769

PR #17448, opened 2021-07-07
Stop filtering ES queries on doc type

### LINKED ISSUE #17447: Stop filtering elasticsearch queries with doc type
We currently pass `doc_type=AddonIndexer.get_doctype_name()` to our Elasticsearch queries, and also pass it in the mapping at index creation time. We only have a single type in our indices right now : addons.

But document types are a deprecated feature, and [Elasticsearch 7.x will remove types completely](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html), so that would break for us. We need to  stop passing it at index creation in the mapping and at indexing  time, and stop passing it when searching.

Because we are still currently passing it to search queries, if we were to remove it right away from index creation/bulk indexing,    search would break and never find any results, since the version we're currently using still supports filtering by type. So the first step to remove it from search queries (since we only have one type it shouldn't have any impact), and then later remove it from everywhere else too.

Prerequisite for #17389

diox:
doc type is deprecated and will eventually be removed. We need to stop passing it at index creation in the mapping and at indexing time, and stop passing it when searching.

Because we are still currently passing it to search queries, if we were to remove it right away from index creation/bulk indexing, search would break and never find any results. So the first step is to remove it from search queries, and then later remove it from everywhere else too.

Fixes #17447

### DIFF
--- a/src/olympia/addons/views.py
+++ b/src/olympia/addons/views.py
@@ -472,7 +472,6 @@ class AddonSearchView(ListAPIView):
             Search(
                 using=amo.search.get_es(),
                 index=AddonIndexer.get_index_alias(),
-                doc_type=AddonIndexer.get_doctype_name(),
             )
             .extra(_source={'excludes': AddonIndexer.hidden_fields})
             .params(search_type='dfs_query_then_fetch')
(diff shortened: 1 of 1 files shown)


willdurand (approved):
r+ if CI agrees

PR #17452, opened 2021-07-08
default local dev to django3.2; deploy to addons-dev with 3.2

### LINKED ISSUE #17449: enable django3.2 on addons-dev
CI tests pass and locally during development we've found no problems.

eviljeff:
fixes #17449 - this keeps the stage/prod deploys using django22 (I hope) but could change that too?


diox (approved):
I'm not sure targeting dev specifically is worth the added complexity, we might go straight to stage & prod, but let's see how it behaves on dev...

PR #17457, opened 2021-07-08
Elasticsearch 7.x compatibility & cleanups

### LINKED ISSUE #17389: Upgrade Elasticsearch to 7.1.x
- https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html
- https://www.elastic.co/guide/en/elasticsearch/reference/current/rolling-upgrades.html

Some changes on our side are required, need to figure out how much and if we can make them in a backwards-compatible way.

diox:
- Remove legacy `SearchMixin` and ES classes (they were obsolete,   everything in addons-server should be using elasticsearch-dsl now)
- Remove `'standard'` no-op filter
- Remove doc types, keep backwards-compatible `_doc` for now, passing `include_type_name=True` to support that with ES 6.x

This doesn't change the Elasticsearch version yet, so it should be fully backwards-compatible.

Fixes #17389


diox:
Blocked until https://github.com/mozilla/addons-server/issues/17447 goes to prod

src/olympia/lib/es/utils.py
```diff
@@ -47,7 +45,7 @@ def index_objects(ids, model, extract_func, index=None, transforms=None, objects
                 {
                     '_source': data,
                     '_id': ob.id,
-                    '_type': ob.get_mapping_type(),
+                    '_type': '_doc',
```
diox, line 48: Unfortunately, that doesn't work on existing indexes:
`{'type': 'illegal_argument_exception', 'reason': 'Rejecting mapping update to [addons-20210707143803] as the final mapping would have more than 1 type: [_doc, addons]'}`

Need to look if I can remove it entirely while keeping ES 6 compatibility. Might have to do something differently depending on the version to make the change transparent.

diox: https://github.com/mozilla/addons-server/pull/17457/commits/7060b61a3da3691aadf8e6c83a1133f54aad2138 should address this

eviljeff:
closing this for now

PR #17471, opened 2021-07-12
add waffle switch to enable mv3 submissions

### LINKED ISSUE #17450: add waffle switch to enable mv3 validation in linter
Experimental support for mv3 was added in mozilla/addons-linter#3721 but we need to pass `max-manifest-version=3` to enable it.  

Behind a waffle switch because it's still experimental and until Firefox support is more advanced we don't want mv3 extensions uploaded to AMO right now even if they _do_ have a valid manifest.

eviljeff:
fixes #17450


src/olympia/devhub/tests/test_tasks.py
```diff
@@ -539,6 +539,43 @@ def test_xpi_autoclose_is_enabled(self, subprocess_mock):
...
+    @override_switch('enable-mv3-submissions', active=True)
+    def test_mv3_submission_enabled(self):
+        with mock.patch('olympia.devhub.tasks.subprocess') as subprocess_mock:
+            subprocess_mock.Popen = self.FakePopen
+
+            tasks.run_addons_linter(
+                path=self.valid_path, channel=amo.RELEASE_CHANNEL_LISTED
+            )
+
+            assert '--max-manifest-version=3' in self.FakePopen.get_args()
```
bobsilverberg, line 567: Should we also be asserting that `'--max-manifest-version=2' not in self.FakePopen.get_args()`, similar to what's being done in the test above?

bobsilverberg (approved):
Looks good, thanks @eviljeff. I just added a couple of questions.

r+wc

src/olympia/devhub/tests/test_tasks.py
```diff
@@ -539,6 +539,43 @@ def test_xpi_autoclose_is_enabled(self, subprocess_mock):
...
+            tasks.run_addons_linter(
+                path=self.valid_path, channel=amo.RELEASE_CHANNEL_LISTED
+            )
+
+            assert '--max-manifest-version=3' not in self.FakePopen.get_args()
+            assert '--max-manifest-version=2' in self.FakePopen.get_args()
+
+        mv3_path = get_addon_file('webextension_mv3.xpi')
+        result = tasks.run_addons_linter(mv3_path, channel=amo.RELEASE_CHANNEL_LISTED)
+        assert json.loads(result or '{}').get('errors') == 1
```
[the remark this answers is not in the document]
eviljeff: if the test passes then the result will be valid.  But if the test doesn't pass then it means the error message is correct (rather than the message be a complaint that `none` can't be parsed as valid json).

src/olympia/devhub/tests/test_tasks.py
eviljeff: Not really sure testing for it in either test is necessary (I added the line when debugging) but I suppose it makes sense to do it in both tests consistently. https://github.com/mozilla/addons-server/pull/17471/commits/b15d08d6cb3b569da8f27609d09c2f43382783f3

Issue #17479, opened 2021-07-13
drop unused properties from Tag model


eviljeff:
We currently define a `restricted` (only settable by admins) and a `denied` (hidden, for abusive tag text) properties for a `Tag`.  

Now all tags are staff defined we don't need `denied`, but do we need `restricted` still? We currently only have 2 restricted tags - `restartless` (from many years ago where most legacy add-ons needed to restart Firefox to work); and `dynamic theme` (added automatically when a developer uploads a dynamic theme).

Do we want to continue to add `dynamic theme` automatically?  If we drop support for `restricted` then a developer would be able to add that tag to arbitrary non-dynamic-theme add-ons (or remove it from an add-on that was a dynamic theme).

Either way, if we _do_ want to continue to support `restricted` in some way, are there preferences how this is exposed?  
- Just hidden from devhub entirely (i.e. not settable by admins either).  It would be possible to set via django admin but a bit messy
- settable by admins in devhub but hidden from non-admins
- visible by normal developers but with the checkbox disabled (or similar)

eviljeff:
@jvillalobos to opine ^

bobsilverberg:
Looking at the existing list of proposed tags at https://docs.google.com/document/d/1QJ5PSHJJn5ruvhITlTMongePpZggsrTYKdmmBO0Wbk4/edit#heading=h.kn3nkelyc86, it doesn't seem like we'd need `restricted` anymore.

eviljeff:
I asked because the use case isn't called out in the doc either way, and `dynamic theme` may have been overlooked (I forgot all about it, and I added it!)

diox:
Note: we have the ability to find dynamic themes in database by looking at the `theme` permission, so we don't need to keep the `dynamic theme` tag if that's just for our own benefit (and even for the outside world, we could expose some kind of search on permissions in the future)

eviljeff:
it was primarily to be able to point users to them https://github.com/orgs/mozilla/projects/31

jvillalobos:
I think it should be fine to drop `restricted`. Dynamic themes used to be a complicated subject because we thought they could become a big deal and we weren't sure how we would classify them on the site. Turns out it's not a huge deal and they're fine as extensions.

eviljeff:
[...]

PR #17481, opened 2021-07-13
replace tag selection in devhub with checkboxes

### LINKED ISSUE #17474: replace existing tags with new, editorial defined, tags
https://docs.google.com/document/d/1QJ5PSHJJn5ruvhITlTMongePpZggsrTYKdmmBO0Wbk4/edit# for full details on background, but tl;dr: we want to replace the existing tags that a developer can define with a fixed list of tags AMO's staff define.  

* existing tags will be reset (deleted)
* devhub edit form needs changing to allow a selection from the tags available (multi-select list?  checkboxes like categories?) rather than a free text box

(and more, but this is the main change in addons-server and probably needs to be done first and in one go)

eviljeff:
fixes #17474 - will need some follow-up either way depending on the outcome of #17479 but this keeps open the option of continuing to tag with the `dynamic theme` tag.


diox:
While testing I noticed something weird/annoying: when you save, you get back to the edit product page, and the tags always show `None`, even if there were some before the edit. It doesn't behave like this for the other fields like homepage. It doesn't appear to be a new bug, so it doesn't block this PR, but it would be good to fix it.

eviljeff:
> 
> 
> While testing I noticed something weird/annoying: when you save, you get back to the edit product page, and the tags always show `None`, even if there were some before the edit. It doesn't behave like this for the other fields like homepage. It doesn't appear to be a new bug, so it doesn't block this PR, but it would be good to fix it.

that's weird! I didn't notice that during development

eviljeff:
> While testing I noticed something weird/annoying: when you save, you get back to the edit product page, and the tags always show `None`, even if there were some before the edit. It doesn't behave like this for the other fields like homepage. It doesn't appear to be a new bug, so it doesn't block this PR, but it would be good to fix it.

I can't replicate this - it always updates correctly for me.  When you say it doesn't appear to be a new bug, do you mean you can replicate this now on addons-dev?  Or just locally?  (maybe you have something weird locally?)

diox:
I had reproduced locally (on `master`) but after investigating, it's because my add-on is unlisted. `addons_section` has a `and show_listed` condition so the XHR that does the edit will not display tags for unlisted... But the section is present, and the form is displayed, and when submitted the data is saved, it's just not displayed at submission time, which is kinda weird.

diox (approved):

eviljeff:
> I had reproduced locally (on `master`) but after investigating, it's because my add-on is unlisted. `addons_section` has a `and show_listed` condition so the XHR that does the edit will not display tags for unlisted... But the section is present, and the form is displayed, and when submitted the data is saved, it's just not displayed at submission time, which is kinda weird.

Oh, good spot.  I guess we never noticed.  Tags are pretty useless for unlisted but dropping the field can be for another issue (if we can be bothered - probably isn't worth it) so I'll just drop the condition in `addons_section`

PR #17503, opened 2021-07-16
add tags shelf endpoint

### LINKED ISSUE #17482: implement "rotating" tag shelf
We want to have a shelf for tags on the homepage, but we want to "rotate" (randomly select) which tag shelf is shown. 

There is currently the ability to create arbitrary shelves that would show content from any valid API response that returns add-ons,  and in some ways rotating is similar to the primary hero shelf, which shows a randomly selected promoted add-on, but we don't want to have to create 44 shelves, each one defining a different `&tag=xxx` value.

eviljeff:
minimally fixes #17482 - plus some refactoring to make the endpoint values constants.  

limiting the shelf addons to certain thresholds will be covered in another issue.

### DIFF
--- a/src/olympia/shelves/forms.py
+++ b/src/olympia/shelves/forms.py
@@ -40,7 +40,7 @@ class ShelfForm(forms.ModelForm):
         if criteria is None:
             return
 
-        if endpoint in (Shelf.Endpoints.SEARCH, Shelf.Endpoints.TAGS):
+        if endpoint in (Shelf.Endpoints.SEARCH, Shelf.Endpoints.RANDOM_TAG):
             if not criteria.startswith('?') or criteria.count('?') > 1:
                 raise forms.ValidationError('Check criteria field.')
             params = dict(parse.parse_qsl(criteria.strip('?')))
@@ -60,7 +60,7 @@ class ShelfForm(forms.ModelForm):
                     'type=statictheme. For non theme addons, use "Extension" in Addon '
                     'type field, not "Theme (Static)".'
                 )
-            if endpoint == Shelf.Endpoints.TAGS:
+            if endpoint == Shelf.Endpoints.RANDOM_TAG:
                 if 'tag' in params:
                     raise forms.ValidationError(
                         'Omit `tag` param for tags shelf - a random tag will be chosen.'
--- a/src/olympia/shelves/forms.py
+++ b/src/olympia/shelves/forms.py
@@ -17,6 +17,9 @@ log = olympia.core.logger.getLogger('z.admin.shelves')
 
 
 class ShelfForm(forms.ModelForm):
+    # It's required in the model, but we set it to a default selectively per endpoint.
+    criteria = forms.CharField(required=False)
+
     class Meta:
         model = Shelf
         fields = (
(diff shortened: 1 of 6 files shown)


src/olympia/shelves/forms.py
```diff
@@ -38,25 +40,38 @@ def clean(self):
...
             ):
                 raise forms.ValidationError(
                     'Check fields - for "Theme (Static)" addon type, use '
                     'type=statictheme. For non theme addons, use "Extension" in Addon '
                     'type field, not "Theme (Static)".'
                 )
+            if endpoint == Shelf.Endpoints.TAGS:
+                if 'tag' in params:
+                    raise forms.ValidationError(
+                        'Omit `tag` param for tags shelf - a random tag will be chosen.'
```
bobsilverberg, line 66: Would it be worth allowing for a `tag` to be specified? Or is that use case dealt with by simply using the `search` endpoint?

src/olympia/shelves/forms.py
```diff
@@ -38,25 +40,38 @@ def clean(self):
...
+                    )
+                params['tag'] = Tag.objects.first().tag_text
```
bobsilverberg, line 68: What does this do?

src/olympia/shelves/models.py
```diff
@@ -1,22 +1,28 @@
...
 from olympia.amo.models import ModelBase
-
-ENDPOINTS = ('collections', 'search')
-
-ENDPOINT_CHOICES = tuple((ty, ty) for ty in ENDPOINTS)
+from olympia.tags.models import Tag
 
 
 class Shelf(ModelBase):
+    Endpoints = namedtuple('Endpoints', ['COLLECTIONS', 'SEARCH', 'TAGS'])(
```
bobsilverberg, line 13: Do you think we should call it `RANDOM_TAG` (or something like that), as opposed to `TAGS`, as the latter suggests to be that it might be used for getting a specific tag?

src/olympia/shelves/models.py
```diff
@@ -1,22 +1,28 @@
...
+
     title = models.CharField(max_length=70, help_text='Will be translated.')
     endpoint = models.CharField(
         max_length=200, choices=ENDPOINT_CHOICES, db_column='shelf_type'
     )
     criteria = models.CharField(
         max_length=200,
         help_text='e.g., "?promoted=recommended&sort=random&type=extension" '
-        'or the collection slug',
+        'or the collection slug. For non-collection shelves, for no criteria enter "?"',
```
bobsilverberg, line 25: Should it be necessary for a user to enter `?` for no criteria? Couldn't we deal with that for them if they leave it blank?

src/olympia/shelves/models.py
```diff
@@ -50,6 +56,20 @@ def get_count(self):
             3 if self.addon_type == amo.ADDON_STATICTHEME else 4
         )
 
+    @cached_property
```
bobsilverberg, line 59: What are the implications of this being a `@cached_property`? Are we still always going to get a new, random tag each time the API is called? Or maybe I don't understand exactly how this property is used.

src/olympia/shelves/tests/test_serializers.py
```diff
@@ -199,6 +209,28 @@ def test_url_and_addons_collections(self):
...
+        assert data['endpoint'] == 'tags'
+        assert data['addon_type'] == 'extension'
+        assert data['criteria'] == '?'
+        assert data['footer']['text'] == {'en-US': 'See similar add-ons?'}
+        assert data['footer']['url'] is None
+        assert data['footer']['outgoing'] is None
+        assert data['url'] == self._get_result_url(tag_shelf, tag='foo')
+        assert len(data['addons']) == 2
+        assert data['addons'][0]['name'] == {'en-US': 'test addon test04'}
+        assert data['addons'][0]['name'] == {'en-US': 'test addon test04'}
```
bobsilverberg, line 232: I think the second one of these asserts should be checking `data['addons'][1]`.

bobsilverberg (commented):
Thanks @eviljeff, it looks good, but I am a bit confused by some of it. Does this code allow for a shelf where the tag is randomized every time the API is called, or just one that is randomized when the shelf is created. I'm probably just misunderstanding the code, but it looks like the latter to me.

eviljeff:
>  Does this code allow for a shelf where the tag is randomized every time the API is called, or just one that is randomized when the shelf is created. I'm probably just misunderstanding the code, but it looks like the latter to me.

The former.  (I'll reply to each comment separately so hopefully it'll make sense but follow-up if it needs further clarification/ code comments)

src/olympia/shelves/forms.py
eviljeff: It'd make more sense to use the search endpoint, at least as it's currently written (the _only_ additional functionality is the random tag value so if you're overriding that it's pointless).  That might change depending on how the other issues end up being implemented, but I'll change/drop the error then.

src/olympia/shelves/forms.py
eviljeff: the code tries to call the API with the criteria set to validate it works, and normally we'd be calling the API with a random tag value.  To replicate this - in a consistent way - it sets the tag to the first Tag instance we have in the database.

src/olympia/shelves/models.py
eviljeff: I had that thought too but never got around to changing it.

src/olympia/shelves/models.py
eviljeff: it's possible... but it makes the validation tricky (especially at a database level) because we require a value for collection shelves.  I'll look into it.

src/olympia/shelves/models.py
eviljeff: `cached_property` stores the value on the object instance - i.e. in this case the `Shelf` object.  `ShelfViewSet` creates the queryset (which is an iterable of `Shelf` objects) each time the API is called so the `cached_property` value will only persist for the life cycle of the API request.

bobsilverberg (approved):
Thanks for all of the clarifications, @eviljeff. It's all good now except for my few comments.

r+wc

PR #17505, opened 2021-07-16
Add searching by ip address in user admin

### LINKED ISSUE #14518: Add ips from user restriction history in knowns ips field in the user admin
The user admin has a knowns ips fields listing, well, known ips. The source for this are `Ratings` and `last_login_ip` on the `UserProfile`. There is another source available though, and that's `UserRestrictionHistory`, so it would be good to use it.

Maybe we should do a `UNION DISTINCT` in MySQL to deal with that while we're at it. Something like (need to see if there is a less verbose way of doing this with the ORM, and also exclude empty string / nulls):
```python
Rating.objects.filter(user=obj).values_list('ip_address', flat=True).order_by().union(
UserRestrictionHistory.objects.filter(user=obj).values_list('last_login_ip', flat=True).order_by(),
UserRestrictionHistory.objects.filter(user=obj).values_list('ip_address', flat=True).order_by(),
UserProfile.objects.filter(user=obj).values_list('last_login_ip', flat=True).order_by(),
)

### LINKED ISSUE #14547: Admin interface to find users by IP address
Admins could use a view that allows them to find all users accounts that match a certain IP address.

### PR DESCRIPTION
**Pull request by diox**
Fixes #14547
Fixes #14518

### DIFF
--- a/src/olympia/users/tests/test_admin.py
+++ b/src/olympia/users/tests/test_admin.py
@@ -654,17 +654,15 @@ class TestUserAdmin(TestCase):
         doc = pq(model_admin.known_ip_adresses(self.user))
         result = doc('ul li').text().split()
         assert len(result) == 7
-        assert set(result) == set(
-            [
-                '130.1.2.4',
-                '128.1.2.3',
-                '129.1.2.4',
-                '127.1.2.3',
-                '15.16.23.42',
-                '172.0.0.2',
-                '4.8.15.16',
-            ]
-        )
+        assert set(result) == {
+            '130.1.2.4',
+            '128.1.2.3',
+            '129.1.2.4',
+            '127.1.2.3',
+            '15.16.23.42',
+            '172.0.0.2',
+            '4.8.15.16',
+        }
 
         # Duplicates are ignored
         Rating.objects.create(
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/olympia/users/tests/test_admin.py [hunk 1]
```diff
@@ -83,6 +83,150 @@ def test_search_for_multiple_user_ids(self):
...
+                self.user.email,
+            ]
+        )
+        # Make sure last login is now displayed, and has the right values.
+        assert (
+            doc('.field-last_login_ip').text()
+            == '127.0.0.3 127.0.0.3 127.0.0.2 127.0.0.2'
+        )
+        # Same for the others that match.
+        assert (
```
diox, line 209: That test is unfortunately not stable, there is no guarantee in which order the "duplicates" entry for the same user (caused by multiple matching user restriction history entries) are returned. If I add an explicit `order_by()` django seems to mess with the ordering more than I'd like, need to investigate more.

static/js/admin/userprofile.js
```diff
@@ -0,0 +1,113 @@
...
+  const search_bar = document.querySelector('#searchbar');
+  const original_search_terms = get_search_bar_terms();
+
+  const ip_fields = document.querySelectorAll(
+    '.field-last_login_ip, \
+    .field-restriction_history__last_login_ip, \
+    .field-restriction_history__ip_address, \
+    .field-_ratings_all__ip_address, \
+    .field-activitylog__iplog__ip_address',
+  );
```
eviljeff, line 93: can we move `search_bar` and `ip_fields` `const`ants to the top of the file if they're going to referenced inside the functions - it makes the code hard to follow

src/olympia/users/tests/test_admin.py
```diff
@@ -516,11 +646,53 @@ def test_known_ip_adresses(self):
...
+        assert set(result) == set(
+            [
+                '130.1.2.4',
+                '128.1.2.3',
+                '129.1.2.4',
+                '127.1.2.3',
+                '15.16.23.42',
+                '172.0.0.2',
+                '4.8.15.16',
+            ]
```
eviljeff, line 666: nit, but can replace `set([...])` with set literal syntax.

eviljeff (approved):
nits all the way down, r+

static/js/admin/userprofile.js
diox: I wanted to initialize them only if needed, so after the `result_list` check. And being `const` I need to initialize them right away... The code is short & simple enough that I can make them `var`s and move the declaration up though.

PR #17514, opened 2021-07-19
parse manifest_version from xpi; store on File

### LINKED ISSUE #17472: store manifest version on File model
we may need to filter or get information on compatibility later.

eviljeff:
fixes #17472


src/olympia/files/models.py
```diff
@@ -156,6 +158,7 @@ def from_upload(cls, upload, version, parsed_data=None):
 
         file_.hash = file_.generate_hash(upload_path)
         file_.original_hash = file_.hash
+        file_.manifest_version = parsed_data.get('manifest_version', 2)
```
willdurand, line 161: could we have a constant for this value? I see it used twice in the same file (the constant could be local to this file)

willdurand (approved):
r+wc

src/olympia/files/models.py
eviljeff: a constant for `2`?

willdurand: Yeah, `DEFAULT_MANIFEST_VERSION`

eviljeff: I think I'm just going to take it out in `from_upload` - there's no valid case where the `manifest_version` isn't defined and available in `parsed_data` (after parsing rdf was removed).  
It just needs to stay as a field default while we have a bunch of code that tests with legacy fixtures, etc, or they'll break.

eviljeff: meh, even dropping from `from_upload` broke many tests.  Constant it is.  https://github.com/mozilla/addons-server/pull/17514/commits/ad046679153d40ccf58d423560b74f3e94903128

PR #17540, opened 2021-07-21
drop unused tag properties + dynamic theme tag

### LINKED ISSUE #17479: drop unused properties from Tag model

eviljeff:
fixes #17479


src/olympia/tags/migrations/0004_auto_20210721_1206.py
```diff
@@ -0,0 +1,27 @@
...
+        migrations.AlterField(
+            model_name='tag',
+            name='denied',
+            field=models.BooleanField(default=False, null=True),
+        ),
+        migrations.AlterField(
+            model_name='tag',
+            name='restricted',
+            field=models.BooleanField(default=False, null=True),
+        ),
```
willdurand, line 26: Do we have an issue filed to drop these columns in a later push?

willdurand (approved):
More red than green award, congratulations!

src/olympia/tags/migrations/0004_auto_20210721_1206.py
eviljeff: https://github.com/mozilla/addons-server/issues/17545

PR #17550, opened 2021-07-22
generate a fallback for footer.url if footer_pathname not set

### LINKED ISSUE #17544: always return shelf footer url - generate fallback
Currently, a Shelf has an optional footer.url, which is the link that is displayed on the homepage to see more addons from that Shelf, and a footer.text to accompany it.  Both are optional, and if not provided default to `See more {shelf.title}` for the latter, and either `/collections/{MOZILLA_USER_ID}/{shelf.criteria}/` for collections or `/search/{shelf.criteria}` otherwise for the former.  This means that frontend needs to duplicate the logic about how each shelf works to be able to generate the footer if not provided.  

Instead, lets generate this in the API and always return the footer content so the logic only needs to be in once place, and adding new types in server doesn't (necessarily) mean adjusting the supporting code in frontend.

eviljeff:
fixes #17544

### DIFF
--- a/src/olympia/shelves/serializers.py
+++ b/src/olympia/shelves/serializers.py
@@ -30,35 +30,36 @@ class ShelfFooterField(serializers.Serializer):
         is_flat_url = request and is_gate_active(request, 'wrap-outgoing-parameter')
 
         url = data.get('url')
-        if not url and obj.endpoint in (
-            Shelf.Endpoints.SEARCH,
-            Shelf.Endpoints.RANDOM_TAG,
-        ):
-            search_url = reverse('search.search')
-            query_string = '&'.join(
-                f'{key}={value}' for key, value in obj.get_param_dict().items()
-            )
-            fallback = absolutify(f'{search_url}?{query_string}')
-            url = (
-                {'url': fallback, 'outgoing': fallback} if not is_flat_url else fallback
-            )
-        elif not url and obj.endpoint == Shelf.Endpoints.COLLECTIONS:
-            fallback = absolutify(
-                reverse(
-                    'collections.detail',
-                    kwargs={
-                        'user_id': str(settings.TASK_USER_ID),
-                        'slug': obj.criteria,
-                    },
+        if not url:
+            if obj.endpoint in (Shelf.Endpoints.SEARCH, Shelf.Endpoints.RANDOM_TAG):
+                search_url = reverse('search.search')
+                query_string = '&'.join(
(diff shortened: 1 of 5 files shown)


src/olympia/shelves/serializers.py
```diff
@@ -6,21 +6,66 @@
...
     text = GetTextTranslationSerializerField(source='footer_text')
 
+    def to_representation(self, obj):
+        data = super().to_representation(obj)
+        request = self.context.get('request', None)
+        # when 'wrap-outgoing-parameter' is on url is a flat string already
+        is_flat_url = request and is_gate_active(request, 'wrap-outgoing-parameter')
+
+        url = data.get('url')
+        if not url and obj.endpoint in (
```
bobsilverberg, line 33: Nit: I'm not sure if it would actually be better because of nesting, but looking at this it feels like you could start with just `if not url:`, because that is part of both of the following `if` / `elif` clauses.

src/olympia/shelves/serializers.py
```diff
@@ -6,21 +6,66 @@
...
+            url = (
+                {'url': fallback, 'outgoing': fallback} if not is_flat_url else fallback
+            )
+        # text = data.get('text')
+
+        if is_flat_url:
+            return url
+        else:
+            return {
+                **data,
```
bobsilverberg, line 64: What would this `**data` be? If we don't need to worry about it when `is_flat_url` is `True`, why do we need to worry about it here? I guess it's just being safe in case there is something there?

bobsilverberg (approved):
Looks good, nice work!

r+wc

src/olympia/shelves/serializers.py
eviljeff: yeah I was just avoid the extra layer of indentation but I guess it's more "correct" to have a two levels.

src/olympia/shelves/serializers.py
eviljeff: `data` will contain `text` too (and other fields if they're added to the serializer later) but I see now I forgot to do the same when `flat_url` is true above!

eviljeff: fixed in https://github.com/mozilla/addons-server/pull/17550/commits/332ed63a8aa3d52f06a8fa8db6fc9ead32858507 - and changed the tests to better check that the output is what we expect (before `ShelfSerializer` was using `CTAField` so we partly relied on those tests to cover all the edge cases with the more unusual api gates)

PR #17560, opened 2021-07-23
Throttle requests using PATCH / DELETE against the ratings API

### LINKED ISSUE #17535: Add rate limiting for user review edits
We should add some rate limiting for editing existing user reviews.

I'd suggest 5 edits per minute. Should that be also counted against the review, i.e. 5 edits  (of the same review) per minute?

diox:
Bonus: also give a rate limit for flagging (same as abuse reporting)

Fixes #17535

### DIFF
--- a/src/olympia/ratings/tests/test_views.py
+++ b/src/olympia/ratings/tests/test_views.py
@@ -1276,29 +1276,60 @@ class TestRatingViewSetDelete(TestCase):
         assert response.status_code == 404
         assert Rating.objects.count() == 1
 
-    def test_no_throttle(self):
-        # Add two reviews for different versions.
+    @override_settings(CACHES=locmem_cache)
+    def test_throttle(self):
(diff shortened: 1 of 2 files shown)


src/olympia/ratings/tests/test_views.py
```diff
@@ -2274,12 +2332,36 @@ def test_no_throttle(self):
...
+            assert RatingFlag.objects.count() == 2
+
+            for x in range(0, 18):
+                # You can keep flagging up to 20 a day.
+                response = self.client.post(
+                    url_b, data={'flag': 'review_flag_reason_spam'}
+                )
+                assert response.status_code == 202
+
+            # Then it becomes too much
```
willdurand, line 2352: The fact that you sometimes use full stops but not all the time triggered my OCD. Can we have one here (and above)? 😺

willdurand (approved):

PR #17572, opened 2021-07-27
combine ShelfManagement->Shelf; relax enabled+position constraint

### LINKED ISSUE #16559: The homepage shelves list can't be saved because of an incorrectly detected position overlap error
1. Open the [/homepageshelves/](https://addons-internal.stage.mozaws.net/en-US/admin/models/discovery/homepageshelves/) admin and create two sets of shelves to alternate between - i.e. two shelves for each position, one shelf Enabled and another one Disabled (as seen in the stage example linked)
2.  Disable the shelf displayed in a certain position and Enable the other shelf available for the same position
3. Save the form

### What happened?
The form can't be saved because of two field validation errors that detect an overlap between the two shelves.

There shouldn't be an overlap detected, since there aren't two shelves with the same position Enabled at the same time

To avoid the errors, you need to delete one of the shelves

xqqJQqbfrz

### LINKED ISSUE #17573: drop ShelfManagement model
We currently have a `ShelfManagement` model, that contains enabled state and shelf position, and `Shelf`, that contains all of the details about the Shelf.  The difference between the two would enable Editorial to build up a collection of `Shelf` modules that would only be added to a `ShelfManagement` when needed.  

But that's not how it's ended up being used so the two models just add extra complexity when creating a new shelf (you have to create the `Shelf` fix; then the `ShelfManagement` object) and in code, with little benefit operationally - any unused shelves can just be left as `enabled=False`.  The code and tests are also larger with two models.

eviljeff:
fixes #17573 and fixes #16559 - it drops the ShelfManagement model that has the constraint.


bobsilverberg (approved):
Looks good, thanks for taking this on @eviljeff. I agree that just dropping the constraint makes sense.

Issue #17561, opened 2021-07-23
Screenshot section has lots of 404s in the network until page is refreshed


ioanarusiczki:
STR:
I've noticed this being reproducible on all AMO servers - I simply load some screenshots into the Image section and save

Expected:
the thumbnails are loaded without errors

Actual: 
there's a loading state and the network returns 404

![Screenshot section](https://user-images.githubusercontent.com/33448286/126791191-8faa8b25-6ae2-4ab6-85f6-ecc7589085af.png)

After a page refresh the errors are no longer displayed.

diox:
It can take seconds or minutes for the thumbnails to be regenerated. After you have uploaded a preview, we try to load it with the expected URL without knowing if it's ready yet in a loop (with a different query string to avoid any caching), hence the 404s. It's a bit aggressive and not very clever but eventually it should get a 200 and stop refreshing.

ioanarusiczki:
@diox Ok, I'll try again on Monday, see how long it takes. I remember this being there before.

ioanarusiczki:
@diox It seems to me it is taking too long. I waited 10 minutes and the loading state did not change.  After a page refresh it's gone.

diox:
I haven't been able to reproduce every time, not sure why. But I think I see the problem: we're trying to load a png when it should be a jpg...

diox:
So, it does come down to a timing issue:
If you save too quickly after uploading, then the task that generates the thumbnail hasn't time to update the `Preview` in the database to update the `sizes` object with the thumbnail format, so `self.get_format('thumbnail')` falls back to the default, which is `png`, so when `thumbnail_url` is called it gives the URL of a PNG, then the template is rendered with that, and the JavaScript notices it's an error and tries to refresh it over and over again... Sometime during this the thumbnail is generated, or at least the `Preview` in the database gets the right `sizes` object, fixing the problem when you reload.

`get_format()` should probably have smarter defaults or something like that to avoid the issue, then it would all work as expected, first failing to load the URL because it hasn't been generated yet, then eventually succeeding after a few automatic reloads by the JavaScript.

It's probably a regression caused by https://github.com/mozilla/addons-server/issues/16639 or one of the issues we worked on when we messed with the thumbnails optimizations.

ioanarusiczki:
@diox I verified on AMO dev, it's a lot faster than before, with 6 screenshots (I could reproduce it easily with only 3 screenshots before) I did not get 404s in the network.

[...]

PR #17577, opened 2021-07-27
Fix default preview format for extensions

diox:
The hardcoded `png` was wrong since we changed the format for the thumbnails to `jpg`. This caused an issue when uploading previews in devhub, the URL was wrong while the Preview was still being resized.

Fixes https://github.com/mozilla/addons-server/issues/17561


eviljeff (approved):
+ Add `'image_format': 'png',` to `ADDON_PREVIEW_SIZES` for completeness?

PR #17589, opened 2021-07-28
Remove queue position from Dev Hub

### LINKED ISSUE #17451: Remove queue position from Dev Hub
The Dev Hub displays an extension's position in the review queue. Since we don't always review the queue by positional order, we should remove this display from the Dev Hub so we don't set false expectations around estimated wait time.

bhushan-borole:
Fixes #17451


bhushan-borole:
Hey @diox 
I will have to remove these tests `test_in_queue` and `test_static_themes_in_queue` right?

diox:
Yes

diox (changes requested):
Nice work. Could you please also remove the `get_position()` function entirely ? I don't think it's used anymore.

bhushan-borole:
Sure I will, No problem.

src/olympia/devhub/views.py
```diff
@@ -126,9 +125,7 @@ def index(request):
         recent_addons = request.user.addons.all().order_by('-modified')[:3]
         ctx['recent_addons'] = []
         for addon in recent_addons:
-            ctx['recent_addons'].append(
-                {'addon': addon, 'position': get_position(addon)}
-            )
+            ctx['recent_addons'].append({'addon': addon})
```
diox, line 128: If you're up for going a little further, I just noticed this is now useless... 

The whole loop could be completely removed, and `recent_addons` assigned to `ctx['recent_addons']` directly. Then, in the template (`src/olympia/devhub/templates/devhub/new-landing/components/my-addons.html`), you would modify the loop over `recent_addons` to do `{% for addon in recent_addons %}`, removing the `{% set addon = item.addon %}` bit.

Does that make sense ?

bhushan-borole: Oh okay, yes I got it.
Will make the changes.

bhushan-borole:
@diox Can you tell me the command to run all the tests?

diox:
If you have docker etc set up locally, you can use `make test` to run the whole suite, but it takes a while (and there might be a few unrelated failures locally). You can also do something like `make test ARGS=something` to pass arguments down. That can be used to restrict which file/directory containing the tests to use for instance (and access any `pytest` command-line arguments)

bhushan-borole:
I was actually thinking of running this command, I copied this from the CI pipeline checks
`pytest -m "es_tests and not needs_locales_compilation and not static_assets" -v src/olympia/`
I have installed docker, but everything is yet to be set.

diox:
That's fairly close to what `make test` does, but it would take a while to run them all locally as we have a bunch of tests.

I just noticed passing `ARGS` seems broken. Instead you can just add `docker-compose exec web` to your command-line (I personally use a small shell script that does exactly this)

diox:
Side note: that `-m es_tests` would run the elasticsearch tests only, which are unlikely to be affected here. I would rather do something like `docker-compose exec web pytest src/olympia/devhub` to run all devhub tests, for instance. Maybe adding the `-m not needs_locales_compilation` to avoid tests needing locales.

bhushan-borole:
Oh okay, it seems all the tests have passed, so I think I don't have to go through that hassle.

bhushan-borole:
> Side note: that `-m es_tests` would run the elasticsearch tests only, which are unlikely to be affected here. I would rather do something like `docker-compose exec web pytest src/olympia/devhub` to run all devhub tests, for instance. Maybe adding the `-m not needs_locales_compilation` to avoid tests needing locales.

Ohh I see, that's actually a nice idea!

diox (approved):
Nice work, thanks!

caitmuenster:
Thank you so much, @bhushan-borole! 😍

bhushan-borole:
@caitmuenster Most Welcome!!!

PR #17594, opened 2021-07-28
missing migrations to tidy up deleted columns/tables

### LINKED ISSUE #17545: add migration to drop restricted and denied columns from `tags` db table
follow on from #17540 - needs to be merged in a subsequent tag than https://github.com/mozilla/addons-server/pull/17540

### LINKED ISSUE #17581: 500 errors when I try to delete shelves that have been enabled then disabled
STR:
1. From the homepage shelves module I tried to delete some shelves but I run into 500 errors every time:

["Random tag test2" ](https://addons-internal.dev.mozaws.net/en-US/admin/models/discovery/homepageshelves/70/change/) which currently is disabled 

server error at delete2

2. Another example with recommended extensions, themes - I tried with the shelf enabled then disabled, same result -> 500 

server error at delete

Note:
I could delete other enabled/disabled shelves while I tried to reproduce this issue - I can't conclude what's causing this.

eviljeff:
fixes #17581 + fixes #17545


willdurand:
mm these migrations broke pretty much all CI jobs.

eviljeff:
> 
> 
> mm these migrations broke pretty much all CI jobs.

yeah, I saw - super weird when I literally just ran `./manage.py makemigrations` and committed the files.  I need to investigate locally because the CI log exceeded 10MB so was cut!

eviljeff:
fixed in https://github.com/mozilla/addons-server/pull/17594/commits/8c84e1359669d5ba5157c5a016b3f0d08802a4c4 - it seems django migrations don't handle changes relating to proxy models well.

src/olympia/discovery/migrations/0006_auto_20200807_2051.py
```diff
@@ -11,16 +11,20 @@ class Migration(migrations.Migration):
...
-            ],
-            options={
-                'verbose_name_plural': 'homepage shelves',
-                'proxy': True,
-                'indexes': [],
-                'constraints': [],
-            },
-            bases=('shelves.shelfmanagement',),
-        ),
+        # Have to disable this because it broke the migration chain when we deleted the
```
willdurand, line 14: mm does it make sense to keep this file then?

eviljeff: If we delete the file we have to adjust any migrations that depend on it and if would cause problems locally if anyone has migrated up to that migration but not beyond.  So we could just delete all the code inside and have it an empty migration file.  Not sure if that's better than commenting it out.

willdurand: oh, mmm it is an old migration. Well, why would that fail now? We have 11 migrations for `discovery` and this `0006` seems to be at least 12 months old.

eviljeff: Not sure, but, afaik, django parses the migrations each time and builds an internal data structure of what the models would be from the migration files, and then compares it to the reality defined in models.py.  In this case it can't resolve the base model for the proxy model defintion any longer so fails.  

It certainly seems like a django bug but some of the renaming I did when I replaced the proxy models for Shelf and ShelfManagement with a single proxy model might have fooled it somehow.

Proxy models are a bit weird anyway - I'm not really certain why they even need to be in migration files when they don't have any database representation

willdurand: Ok, I guess that's fine then

src/olympia/discovery/migrations/0006_auto_20200807_2051.py
```diff
@@ -11,16 +11,20 @@ class Migration(migrations.Migration):
...
+        # base shelves.shelfmanagement - django does seem to be able detect it changing.
+        # Its okay though - it's only a proxy model (no database table backing it).
```
willdurand, line 16: Maybe add a date to indicate when we commented out this migration (we can retrieve it with `git blame` but it is simpler to have it in the comment IMO)

willdurand (approved):
r+wc

PR #17595, opened 2021-07-29
Forbid submission of @temporary-addon guids

### LINKED ISSUE #17579: Deny new add-ons that use an ID ending in @temporary-addon

IDs ending in `@temporary-addon` are reserved for temporarily loaded add-ons in Firefox. We should not accept such IDs for signing (and thus submission).

Since we already have a few add-ons on AMO with that ID, we can't deny any submissions yet, so we are limiting this issue to new add-ons first.

Submitting a new add-on with an ID ending in `@temporary-addon`...

### What happened?

...is accepted by AMO.

...should not be accepted by AMO.

As mentioned above, submission of updates to existing add-ons using such IDs should not be denied for now.

diox:
Fixes #17579

### DIFF
--- a/src/olympia/constants/base.py
+++ b/src/olympia/constants/base.py
@@ -375,6 +375,7 @@ RESERVED_ADDON_GUIDS = (
     '@mozillaonline.com',
     '@mozillafoundation.org',
     '@rally.mozilla.org',
+    # Temporary add-ons as defined in Firefox. Should not be submitted to AMO.
     '@temporary-addon',
 )
 
--- a/src/olympia/constants/base.py
+++ b/src/olympia/constants/base.py
@@ -362,8 +362,9 @@ ADDON_GUID_PATTERN = re.compile(
     re.IGNORECASE,
 )
 
-# Changes to this list need to be synced with the privileged add-ons signing pipeline.
-# Please reach out to the Operations Team (awagner) before making any changes!
+# Changes to this list need to be coordinated with the Operations team to make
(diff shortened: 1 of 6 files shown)


src/olympia/constants/base.py
```diff
@@ -364,7 +364,7 @@
 
 # Changes to this list need to be synced with the privileged add-ons signing pipeline.
 # Please reach out to the Operations Team (awagner) before making any changes!
```
eviljeff, line 366: add to this comment as it's not just for privileged add-ons now?

diox: Yeah, I wasn't sure exactly what to say here instead. Any suggestion @wagnerand ?

wagnerand (approved):

src/olympia/constants/base.py
wagnerand: I think we can say something like

```
# Changes to this list need to be coordinated with the Operations team to make sure other systems like the privileged add-ons signing pipeline are in sync. Please reach out to them before making changes.
```

src/olympia/constants/base.py
```diff
@@ -374,6 +374,7 @@
     '@mozillaonline.com',
     '@mozillafoundation.org',
     '@rally.mozilla.org',
+    '@temporary-addon',
```
wagnerand, line 378: Maybe also add a specific comment for this line, like
```
# Temporary add-ons as defined in Firefox. They should not be submitted/signed.
```

PR #17602, opened 2021-07-29
implement ratings and users threshold search filters

### LINKED ISSUE #17497: limit the results for a (tag) shelf to minimum thresholds
For tag shelves (that will be randomly displayed on the homepage) we want to limit the add-ons shown to ones that meet certain thresholds.

> Sorting by users sounds good, but we should also set some minimum thresholds. I  think something like 500 users or more, and 3.5-star rating or more.

Implementation details are if this is hardcoded just for tag shelves, or available for other shelf endpoints too, or added as an API feature (e.g. something like a param of `?threshold=rating>3.5`)

eviljeff:
fixes #17497


docs/topics/api/addons.rst
```diff
@@ -76,6 +78,18 @@ This endpoint allows you to search through public add-ons.
...
+.. _addon-threshold-param:
+
+    Threshold style parameters allow queries against numeric values using comparison operators.
+
+    The following is supported (examples for query parameter `foo`):
+    * greater than > (example query: ?foo=>10.1)
+    * less than < (example query: ?foo=<10.1)
+    * greater than or equal to >= (example query: ?foo=>=10.1)
+    * less than or equal to <= (example query: ?foo=<=10.1)
+    * equal to == (example query: ?foo==10.1)
```
eviljeff, line 90: tbh I'm not 100% happy with the syntax.  Because queries kinda have to be of the format xxx=yyy theres always an `=` in there.  E.g. `=>` seems like it should be greater than or equal to, but then what is just greater than? (If that's actually a useful thing for comparing to a decimal value - could just only have greater|less than or equal to operators).

diox: could go with django's ORM style - `gt`, `lt`, `gte`, `lte`, and `=` ?

eviljeff: hmm, what would an example url look like in django ORM style?

diox: `?foo__gt=10.1`, `?foo__lt=10.1`, `?foo__gte=10.1`, `?foo__lte=10.1`, `?foo=10.1`

eviljeff: Could do.  I _did_ like the style of writing `>` directly in the url but `__gt` is more explicit.

diox: Could also look at how https://django-filter.readthedocs.io/en/stable/ does it

eviljeff: it'd need 5 `AddonQueryParam` classes per field but that's fine.

bobsilverberg: I agree that the double equals sign is ugly and potentially confusing. I don't love the django style either, but agree it's better.

At the risk of suggesting removing functionality, it seems like we actually only need a threshold of gte currently (i.e., must have at least x users, must have at least x rating), so maybe we don't even need all of the variations?

eviljeff: its true we only need gte right now, but if it's straightforward enough (it's really little more than a lightweight layer over the ES query) I'd prefer to fully implement it - we never know when it might be useful in the future

bobsilverberg: Fair enough. Are you going to update the patch with the django approach? If so I'll want for that to do a full review.

willdurand: When it comes to filters in json API, I've seen the following in the past (exclude OData):

```
/api/?filters[foo][eq]=100&filters[bar][gt]=50
```

and

```
/api/?foo=eq(100)&bar=gt(50)
```

eviljeff: https://github.com/mozilla/addons-server/pull/17602/commits/fd2aabad7d4951dc27e75d29165ca9fdbb977e75 changes the PR to use orm __style comparison parameters.

docs/topics/api/addons.rst
```diff
@@ -76,6 +78,18 @@ This endpoint allows you to search through public add-ons.
...
+    Threshold style parameters allow queries against numeric values using comparison.
+
+    The following is supported (examples for query parameter `foo`):
+    * greater than ``foo__gt`` (example query: ?foo__gt=10.1)
+    * less than ``foo__lt`` (example query: ?foo__lt=10.1)
+    * greater than or equal to ``foo__gte`` (example query: ?foo__gte=10.1)
+    * less than or equal to ``foo__lte`` (example query: ?foo__lte=10.1)
+    * equal to ``foo`` (example query: ?foo=10.1)
+
+
```
diox, line 92: This is missing the indentation to make a list, it would squish everything on a single line.

diox (approved):

src/olympia/search/filters.py
```diff
@@ -459,6 +518,16 @@ def get_es_query(self):
         return [Q('nested', path='colors', query=query.Bool(filter=clauses))]
 
 
+class AddonRatingQueryParam(AddonThresholdQueryParam):
+    query_param = 'ratings'
+    es_field = 'ratings.average'
```
diox, line 523: Actually I just realized, this field is not indexed, so that wouldn't work properly. You need to change that in the indexer and force a reindex.

diox: Alternatively we could search on `bayesian_rating` instead, which is indexed and usually better for search purposes, even though it's not the exact same value users see.

eviljeff: it seemed to work locally?

diox: mmm if it worked locally then you probably don't have a real mapping set up (it can happen if you trigger the indexing of an add-on before running `reindex`), I would run `reindex --wipe` to ensure you do. It shouldn't be possible in ES to query/filter against non-indexed fields.

diox: You should get an Internal Server Error with an Elasticsearch traceback looking like this:
```
failed to create query: {
  "bool" : {
    "filter" : [
      {
        "terms" : {
          "status" : [
            4
          ],
          "boost" : 1.0
        }
      },
      {
        "exists" : {
          "field" : "current_version",
          "boost" : 1.0
        }
      },
      {
        "term" : {
          "is_disabled" : {
            "value" : false,
            "boost" : 1.0
          }
        }
      },
      {
        "range" : {
          "ratings.average" : {
            "from" : 1.0,
            "to" : null,
            "include_lower" : false,
            "include_upper" : true,
            "boost" : 1.0
          }
        }
      }
    ],
    "adjust_pure_negative" : true,
    "boost" : 1.0
  }
}
```

docs/topics/api/addons.rst
eviljeff: https://github.com/mozilla/addons-server/pull/17602/commits/c202cf7f70a2874cf635f4d20c48054a43ebb75f

src/olympia/search/filters.py
eviljeff: https://github.com/mozilla/addons-server/pull/17602/commits/ae527dcb9002a0860538c7cf85f49c02340ca341 adds ratings.average to the ES index

PR #17603, opened 2021-07-29
run pyupgrade

### LINKED ISSUE #17571: use pyupgrade to freshen our code?
https://github.com/asottile/pyupgrade can rewrite python code to use newer syntax.  Doesn't really help any more than any linting tool does, but may as well have a consistent newer style of python rather than a mishmash depending on how many years ago it was originally written.

I've played around with it and it's not a drop in command line tool because it seems to expect to be used as a [pre-commit](https://github.com/pre-commit/pre-commit) hook mainly - e.g. you can't just give it a directory and it'll walk through the structure - it needs single files.  I hacked something together to work around that to be used as one-off though, it would just need a little polish to use permanently (and possibly submitted as a patch upstream)

eviljeff:
This is what it looks like after running pyupgrade (via the script in the first commit) + black to fix formatting.
fixes #17571

### DIFF
--- a/src/olympia/constants/activity.py
+++ b/src/olympia/constants/activity.py
@@ -1,6 +1,5 @@
 # The absolute import feature is required so that we get the root celery
 # module rather than `amo.celery`.
-from __future__ import absolute_import
 
 from collections import namedtuple
 from inspect import isclass
@@ -15,7 +14,7 @@ __all__ = (
 )
 
 
-class _LOG(object):
+class _LOG:
     action_class = None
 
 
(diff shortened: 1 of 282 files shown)


willdurand:
Looks good and useful to me!

eviljeff:
The hacky script to run pyupgrade has been changed a little, and I stuck it in `/scripts`.  I figured this would be at most an occasional script to be run when we upgrade python versions so didn't add the pyupgrade dependency.  (Though it _could_ be added to `make format` if we wanted?  Or go the opposite direction and not include it in the repo at all).

willdurand:
> Though it _could_ be added to `make format` if we wanted?

If it does not take too long to complete, we could run it in `make format` indeed.

eviljeff:
> 
> 
> > Though it _could_ be added to `make format` if we wanted?
> 
> If it does not take too long to complete, we could run it in `make format` indeed.

it takes a good few minutes, though so does running `black` and `prettier` on all the code.  If people are regularly doing `make format`, like before every push to origin, I don't think it's worth the extra time it adds on; if it's an occasional fix then yeah?

src/olympia/constants/activity.py
```diff
@@ -1,7 +1,3 @@
-# The absolute import feature is required so that we get the root celery
-# module rather than `amo.celery`.
```
eviljeff, line 2: After reading the original commit I still have no idea what the `absolute_import` in this file had to do with celery as it doesn't reference celery at all and these constants aren't related to it either.

diox: It made sense when it imported celery a while ago but that was removed in 133827601c02a5f6c3aca3c1db3c86f40f2ca685

diox (approved):
I think it's fine to keep as something we run once in a while for now and not in `make format` or `pre-commit`.

Issue #17539, opened 2021-07-21
Upgrade Celery to 5.x


diox:
We chose to stay on Celery LTS in https://github.com/mozilla/addons-server/issues/15677 but no new LTS has been announced and the version we're using will be EOL soon, so we need to switch to 5.x.

This means unblocking pinned dependencies, ensuring compatibility of things like https://github.com/mozilla/django-post-request-task/, and of course getting everything to work and tests to pass.

diox:
To prepare for that I've released django-post-request-task 0.5 to PyPI. I've only tested it locally with Celery 4 for the moment but its unit tests pass with Celery 5 now.

diox:
Bumping priority since the EOL of celery 4 is near.

diox:
QA: test that add-on upload and reindexing for search after changes work.

ioanarusiczki:
@diox I tested today the reindex for search after making changes: 
- when I'm authorized and changed: name, summary, ratings, icons, authors, deleted/disabled/invisible - the changes are immediately visible into the search results.  
- unauthorized users might see the changes with a delay (except for icons and authors). For example, if an addon was deleted, it will take a while until it's not displayed on the search results page (I had to make several page refreshes and wait a while). I think it's a result of the cache feature.

ioanarusiczki:
@diox I uploaded listed/unlisted versions of an extension, theme, dicts and lang pack without errors. 

I verified this issue on AMO dev - FF90(win10).

PR #17624, opened 2021-08-02
Update Celery to 5.x

diox:
Fixes https://github.com/mozilla/addons-server/issues/17539

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,10 +8,10 @@ updates:
   ignore:
   - dependency-name: billiard
     versions:
-    - ">= 5"
+    - ">= 4"
   - dependency-name: celery
     versions:
-    - ">= 5"
+    - ">= 6"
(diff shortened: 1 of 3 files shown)


eviljeff (approved):
... this is it?  I'm suspicious 😐

diox:
The import issue (which was also a problem in django-post-request-task) + the upgrades seemed to do the trick, yeah.

diox:
We probably need to update cloudops deployment configuration though:
https://docs.celeryproject.org/en/stable/whatsnew-5.1.html says:
> The global options can no longer be positioned after the sub-command. Instead, they must be positioned as an option for the celery command like so:
> `celery --app path.to.app worker`

We use the `--app` / `-A` parameter in our dev/stage/prod deploys but it's after the "command" (`worker`). It's not an issue in addons-server local configuration.

PR #17635, opened 2021-08-03
set mv3 minimum app version to 100

### LINKED ISSUE #17473: set min version compatibility for mv3 addons
We want to set a minimum version compatibility for add-ons that use manifest version 3, equal to the version of Firefox that supports it.  (Need a target version of Firefox, or a guess for now - won't matter if it moves until we enable the submission switch)

@mixedpuppy want to estimate a target release?  98?

eviljeff:
fixes #17473

### DIFF
--- a/src/olympia/constants/base.py
+++ b/src/olympia/constants/base.py
@@ -355,6 +355,12 @@ DEFAULT_STATIC_THEME_MIN_VERSION_ANDROID = '65.0'
 # Dicts are not compatible with Firefox for Android, only desktop is relevant.
 DEFAULT_WEBEXT_DICT_MIN_VERSION_FIREFOX = '61.0'
 
+# The default version of Firefox that (is planned to) support manifest version 3 (MV3)
+# This will probably have to be adjusted later.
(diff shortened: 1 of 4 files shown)


src/olympia/constants/base.py
```diff
@@ -355,6 +355,12 @@
 # Dicts are not compatible with Firefox for Android, only desktop is relevant.
 DEFAULT_WEBEXT_DICT_MIN_VERSION_FIREFOX = '61.0'
 
+# The default version of Firefox that (is planned to) support manifest version 3 (MV3)
+# This will probably have to be adjusted later.
+DEFAULT_WEBEXT_MIN_VERSION_MV3_FIREFOX = '100.0'
+# We don't know if the Android min version will be different, but assume it might be.
+DEFAULT_WEBEXT_MIN_VERSION_MV3_ANDROID = '100.0'
```
willdurand, line 362: So maybe (if that's possible):

```suggestion
# We don't know if the Android min version will be different, but assume it might be.
DEFAULT_WEBEXT_MIN_VERSION_MV3_ANDROID = DEFAULT_WEBEXT_MIN_VERSION_MV3_FIREFOX
```

willdurand (approved):

PR #17669, opened 2021-08-09
Make AMO ID the canonical identifier in reviewer tools

### LINKED ISSUE #17644: Make the AMO add-on ID the canonical identifier in the reviewer tools
Currently, the slug is the canonical identifier for various parts on AMO. For the reviewer tools, we want to change that to the numerical AMO add-on ID to avoid loss of access when add-ons get deleted (which clears the slug) or the developer changes the slug.

It also removes ambiguity for add-ons that are reusing the GUID across several AMO add-on instances.

Accessing the page using GUID/slug should redirect to the numerical ID.

wagnerand:
Fix #17644.

### DIFF
--- a/src/olympia/reviewers/decorators.py
+++ b/src/olympia/reviewers/decorators.py
@@ -98,17 +98,17 @@ def any_reviewer_or_moderator_required(f):
     return wrapper
 
 
-def reviewer_addon_view(
-    f, qs=Addon.objects.all, include_deleted_when_checking_versions=False
-):
-    """Use a separate view decorator here so we can make the AMO ID the canonical
-    identifier in reviewer tools urls."""
+def reviewer_addon_view(f):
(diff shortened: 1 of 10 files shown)


src/olympia/reviewers/decorators.py
```diff
@@ -90,3 +96,44 @@ def wrapper(request, *args, **kw):
         raise PermissionDenied
 
     return wrapper
+
+
+def reviewer_addon_view(
+    f, qs=Addon.objects.all, include_deleted_when_checking_versions=False
+):
+    """Use a separate view decorator here so we can make the AMO ID the canonical
+    identifier in reviewer tools urls."""
```
diox, line 105: This doesn't describe what the decorator is doing very well

diox (approved):

PR #17673, opened 2021-08-09
Allow users with Ratings:BypassThrottling permission to bypass ratings API throttling

### LINKED ISSUE #17659: Add the possibility to bypass API throttling for specific users on stage
We need a way to bypass the API throttling for specific users to avoid test failures in the automated release tests - https://github.com/mozilla/addons-release-tests.
Since tests are running quite fast on CI, the ratings test suite is currently failing as it hits the API throttling.
According to @diox this should be a pretty straightforward implementation.

diox:
Fixes #17659

### DIFF
--- a/src/olympia/ratings/tests/test_views.py
+++ b/src/olympia/ratings/tests/test_views.py
@@ -1510,7 +1510,9 @@ class TestRatingViewSetEdit(TestCase):
         with freeze_time('2021-08-09') as frozen_time:
             for x in range(1, 6):
                 response = self.client.patch(
-                    self.url, {'score': x, 'body': f'Blâh {x}'}
+                    self.url,
+                    {'score': x, 'body': f'Blâh {x}'},
+                    REMOTE_ADDR='127.0.0.42',
                 )
                 assert response.status_code == 200
                 self.rating.reload()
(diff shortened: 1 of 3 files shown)


src/olympia/ratings/tests/test_views.py
```diff
@@ -1503,6 +1503,60 @@ def test_throttle(self):
...
+
+            # Now with the permission we should be ok.
+            self.grant_permission(self.user, 'Ratings:BypassThrottling')
+            response = self.client.patch(self.url, {'score': 1, 'body': 'Eheheh'})
+            assert response.status_code == 200
+            self.rating.reload()
+            assert str(self.rating.body) == 'Eheheh'
+            assert self.rating.rating == 1
+
+            # Someone else with the same IP would still be stuck.
```
willdurand, line 1535: Same IP because it is the default in `user_factory()`?

willdurand (approved):

src/olympia/ratings/tests/test_views.py
diox: Not even related to the user, this is using the (default) client IP. I'll make it more explicit (need to set `REMOTE_ADDR` in the `client.<verb>()` call)

PR #17675, opened 2021-08-09
drop is_restart_required from File (and api/v5)

### LINKED ISSUE #17390: Remove is_restart_required from API responses
Depends on https://github.com/mozilla/addons-frontend/issues/10716

After https://github.com/mozilla/addons-frontend/issues/10716 is fixed, addons-frontend will no longer be using ~either~ `is_restart_required` ~or `is_webextension`~ from API responses. Therefore, we can remove them from `v5` of the API.

--edited to limit scope of the issue to `is_restart_required`.  `is_webextension` to be dealt with in #17658

eviljeff:
fixes #17390


diox (approved):
Needs follow-up issue to drop the column

Issue #17674, opened 2021-08-09
Add `Access-Control-Allow-Origin: *` to `/__version__`


willdurand:
Similar to https://github.com/mozilla/addons-frontend/issues/10822, adding the header `access-control-allow-origin: *` to the `/__version__` endpoint would allow anyone to fetch the content of this file.

AlexandraMoga:
The `access-control-allow-origin:  *` response header is now present in https://addons-dev.allizom.org/__version__

PR #17676, opened 2021-08-09
Set 'Access-Control-Allow-Origin: *' to '/__version__' endpoint

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/17674

---

I am not sure how we can test that locally, `/__version__` is served by addons-frontend it seems 🤷

### DIFF
--- a/src/olympia/amo/views.py
+++ b/src/olympia/amo/views.py
@@ -135,7 +135,9 @@ def version(request):
     contents['django'] = '{major}.{minor}'.format(
         major=django.VERSION[0], minor=django.VERSION[1]
     )
-    return HttpResponse(json.dumps(contents), content_type='application/json')
+    res = HttpResponse(json.dumps(contents), content_type='application/json')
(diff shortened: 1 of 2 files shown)


src/olympia/amo/views.py
```diff
@@ -135,7 +135,9 @@ def version(request):
     contents['django'] = '{major}.{minor}'.format(
         major=django.VERSION[0], minor=django.VERSION[1]
     )
-    return HttpResponse(json.dumps(contents), content_type='application/json')
+    res = HttpResponse(json.dumps(contents), content_type='application/json')
+    res.headers['Access-Control-Allow-Origin'] = '*'
```
eviljeff, line 139: you can set headers on creating the `HttpResponse` in django3.2

eviljeff (approved):
r+wc

you can edit the ngnix config in the container if you really want to test locally (it's hassle; I wouldn't bother and just wait for it to deploy to addons-dev)

PR #17672, opened 2021-08-09
Update "report a bug" link in footer

bhushan-borole:
Fixes #16884 

@willdurand

PR #17683, opened 2021-08-10
Revert "Update "report a bug" link in footer"

willdurand:
Reverts mozilla/addons-server#17672

---

This is needed because the wrong footer template was edited...

### DIFF
--- a/src/olympia/templates/photon-footer.html
+++ b/src/olympia/templates/photon-footer.html
@@ -14,7 +14,7 @@
                 <li><a href="{{ url('devhub.index') }}">Developer Hub</a></li>
                 <li><a href="https://developer.mozilla.org/docs/Mozilla/Add-ons/AMO/Policy">Developer Policies</a></li>
                 <li><a href="https://discourse.mozilla-community.org/c/add-ons">Forum</a></li>
-                <li><a class="Footer-bug-report-link" href="https://addons.mozilla.org/about">Report a bug</a></li>
+                <li><a class="Footer-bug-report-link" href="https://developer.mozilla.org/docs/Mozilla/Add-ons/Contact_us">Report a bug</a></li>
                 <li><a href="{{ url('pages.review_guide') }}">Review Guide</a></li>
                 <li><a href="https://status.mozilla.org/">Site Status</a></li>
             </ul>


diox:
Don't think this should be reverted ? We happen to have 2 different templates because only the devhub "landing" page was ever revamped, but it's still used and should also point to the right URL

willdurand:
> Don't think this should be reverted ? We happen to have 2 different templates because only the devhub "landing" page was ever revamped, but it's still used and should also point to the right URL

@diox the photon footer is also the AMO frontend one, which has the same link, which is why I reverted the patch.

diox:
If you revert this the devhub landing page will have the wrong link while the rest of devhub will have the right one. I think we're probably just missing an issue in addons-frontend to also point to the about page to be 100% consistent, but we shouldn't revert this.

eviljeff:
> 
> 
> > Don't think this should be reverted ? We happen to have 2 different templates because only the devhub "landing" page was ever revamped, but it's still used and should also point to the right URL
> 
> @diox the photon footer is also the AMO frontend one, which has the same link, which is why I reverted the patch.

That explanation just confuses me further 😆 

Can I start with: what is the correct link?

willdurand:
> Can I start with: what is the correct link?

There is no "correct link" IMO. Only 1 link that was 404. Other links work.

willdurand:
> If you revert this the devhub landing page will have the wrong link while the rest of devhub will have the right one. I think we're probably just missing an issue in addons-frontend to also point to the about page to be 100% consistent, but we shouldn't revert this.

Yeah, maybe. Although "report a bug" on AMO pointing to the about page that is linked above seems weird. Maybe we should remove the "report a bug" link in addons-frontend?

eviljeff:
> 
> 
> > Can I start with: what is the correct link?
> 
> There is no "correct link" IMO. Only 1 link that was 404. Other links work.

So, is https://addons.mozilla.org/about a 404?  If not then this revert doesn't make things any more correct?  Though I think the actual way forward is someone (Product?) deciding which link is correct.

willdurand:
> > > Can I start with: what is the correct link?
> > 
> > 
> > There is no "correct link" IMO. Only 1 link that was 404. Other links work.
> 
> So, is [addons.mozilla.org/about](https://addons.mozilla.org/about) a 404? If not then this revert doesn't make things any more correct? Though I think the actual way forward is someone (Product?) deciding which link is correct.

The link that was 404 has been fixed in https://github.com/mozilla/addons-server/pull/17684 but there was a previous patch (which is the one reverted here) trying to fix this 404 except that the wrong template was edited.

willdurand:
Both a.m.o/about and https://developer.mozilla.org/docs/Mozilla/Add-ons/Contact_us are valid links. The latter is used on AMO frontend and the "photo-footer" template here is based on AMO frontend. Reverting the link in this patch keep both footers similar but as @diox said, the "report a bug" link in DevHub would point to a.m.o/about whereas the other "report a bug" link will point to https://developer.mozilla.org/docs/Mozilla/Add-ons/Contact_us

willdurand:
I guess we can NOT revert the change here (i.e. close this PR). "Report a bug" in addons-frontend will either stay as it is currently OR we probably want to remove it given that we have a link to the About page already:

Screenshot 2021-08-10 at 16 21 05

eviljeff:
+1 to closing this PR.  imo it's more important to have the same "Report a bug" link in all the places on devhub than to match frontend's footer (and that may be changed too in a future issue)

PR #17688, opened 2021-08-10
Fix review url redirect

wagnerand:
Fix mozilla/addons#8453.

### DIFF
--- a/src/olympia/reviewers/tests/test_decorators.py
+++ b/src/olympia/reviewers/tests/test_decorators.py
@@ -16,7 +16,6 @@ class TestReviewerAddonView(TestCase):
         self.func.__name__ = 'mock_function'
         self.view = dec.reviewer_addon_view(self.func)
         self.request = mock.Mock()
-        self.slug_path = f'http://testserver/addon/{self.addon.slug}/reviews'
         self.request.path = self.id_path = (
             'http://testserver/addon/%s/reviews' % self.addon.id
         )
@@ -27,7 +26,7 @@ class TestReviewerAddonView(TestCase):
         assert res == mock.sentinel.OK
 
     def test_301_by_slug(self):
-        self.request.path = self.slug_path
+        self.request.path = f'http://testserver/addon/{self.addon.slug}/reviews'
         res = self.view(self.request, self.addon.slug)
         self.assert3xx(res, self.id_path, 301)
 
@@ -41,6 +40,8 @@ class TestReviewerAddonView(TestCase):
         self.request.path = f'http://testserver/addon/{slug}/reviews/{slug}/path'
 
         res = self.view(self.request, self.addon.slug)
+        # We only replace the part of the URL that matters to look up the
+        # add-on (in this case, the slug, only once).
         redirection = (
             f'http://testserver/addon/{self.addon.id}/reviews/{quote(slug)}/path'
         )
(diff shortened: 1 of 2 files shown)


diox:
The `FIXME` comment needs to be updated. It should probably just say something like:
```python
            # FIXME: this replace() is fragile, if the add-on slug appears
            # elsewhere in the URL it will break. Instead, we should probably
            # use `request.resolver_match` to rebuild the URL, replacing
            # `kwargs['addon_id']` if present by `addon.pk`
```

src/olympia/reviewers/tests/test_decorators.py
```diff
@@ -34,16 +32,18 @@ def test_301_by_slug(self):
...
 
     def test_slug_replace_no_conflict(self):
-        slug = quote(self.addon.slug.encode('utf8'))
+        slug = self.addon.slug
         self.request.path = f'http://testserver/addon/{slug}/reviews/{slug}/path'
 
         res = self.view(self.request, self.addon.slug)
-        redirection = f'http://testserver/addon/{self.addon.id}/reviews/{slug}/path'
+        redirection = (
+            f'http://testserver/addon/{self.addon.id}/reviews/{quote(slug)}/path'
```
eviljeff, line 45: why is it quoted here?

src/olympia/reviewers/tests/test_decorators.py
```diff
@@ -16,9 +16,7 @@ def setUp(self):
         self.func.__name__ = 'mock_function'
         self.view = dec.reviewer_addon_view(self.func)
         self.request = mock.Mock()
-        self.slug_path = 'http://testserver/addon/%s/reviews' % quote(
-            self.addon.slug.encode('utf-8')
-        )
+        self.slug_path = f'http://testserver/addon/{self.addon.slug}/reviews'
```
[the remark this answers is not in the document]
eviljeff: I guess not related to the changes in this PR, but made it hard to follow.

src/olympia/reviewers/tests/test_decorators.py
```diff
@@ -34,16 +32,18 @@ def test_301_by_slug(self):
         self.assert3xx(res, self.id_path, 301)
 
     def test_301_by_guid(self):
-        self.request.path = f'http://testserver/addon/{quote(self.addon.guid)}/reviews'
+        self.request.path = f'http://testserver/addon/{self.addon.guid}/reviews'
         res = self.view(self.request, self.addon.guid)
         self.assert3xx(res, self.id_path, 301)
 
     def test_slug_replace_no_conflict(self):
```
eviljeff, line 39: what's this tests supposed to be checking for?  the slug doesn't seem to be replaced, and no conflict with what?

eviljeff (approved):
r+ ... because we know the fix will work and is needed.  But I have opinions on the tests.

src/olympia/reviewers/tests/test_decorators.py
diox: We receive the unquoted path but we quote it (via `urlencode()`) when returning the new URL (the redirection)

src/olympia/reviewers/tests/test_decorators.py
diox: If the slug is somehow present twice in the URL, we don't replace twice - said differently, we only replace the part of the URL that matters to look up the add-on.

As I've noted in the `FIXME` in the decorator (and in my comment above that rewords that `FIXME`) the whole replace  method, copied from the original `addons` decorator, is still not great and can lead to incorrect URLs depending on the slug, though.

willdurand:
I updated this PR a bit so that we can land it and cherry-pick it. Hopefully the minor changes I made are enough.

PR #17691, opened 2021-08-10
switch memcache backend to pymemcache

### LINKED ISSUE #16489: Replace MemcachedCache backend with PyMemcacheCache or PyLibMCCache (post django3.2 fix)
from django release notes:
> The django.core.cache.backends.memcached.MemcachedCache backend is deprecated as python-memcached has some problems and seems to be unmaintained. Use django.core.cache.backends.memcached.PyMemcacheCache or django.core.cache.backends.memcached.PyLibMCCache instead.

eviljeff:
fixes #16489 ... I guess?  Tests all pass (we _do_ test cache functionality in tests, right?)


diox (approved):
We do have a few tests depending on the default cache backend working, yes. And it seems to work locally as well.

PR #17695, opened 2021-08-11
Allow {tag} in Shelf title and footer text for random-tag shelves

### LINKED ISSUE #17689: Allow the current tag to be referenced in shelf copy
For a homepage shelf with an endpoint type `random-tag` we might want to craft a title or footer_text referencing the random tag returned, but this isn't possible.  We should add support for specifying a `{tag}` token or similar in strings.

eviljeff:
fixes #17689


src/olympia/shelves/models.py
```diff
@@ -27,7 +31,9 @@ class Shelf(ModelBase):
     footer_text = models.CharField(
         max_length=70,
```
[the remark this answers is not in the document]
eviljeff: The tag value isn't stored in the shelf model, so the 70 max length of the field isn't relevant.  Editorial define the tags and define the footer text so it's up to them to take how it may be displayed on the homepage into account.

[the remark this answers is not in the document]
willdurand: That's what I am saying I believe. This wasn't a problem before because there was no placeholder but now there is `{tag}`.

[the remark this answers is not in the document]
eviljeff: I'm not clear what you're suggesting to change, or how the problem is significantly different to now when l10n can result in strings of an unknown value being displayed for non en-US locales.

src/olympia/shelves/tests/test_serializers.py
```diff
@@ -295,6 +295,12 @@ def test_tags_shelf(self):
         assert data['addons'][0]['name'] == {'en-US': 'test addon test04'}
         assert data['addons'][1]['name'] == {'en-US': 'test addon test01'}
 
+    def test_tag_token_substitution(self):
+        self.tag_shelf.update(title='Title for {tag}', footer_text='footer with {tag}')
+        data = self.serialize(self.tag_shelf)
+        assert data['title'] == {'en-US': 'Title for foo'}
+        assert data['footer']['text'] == {'en-US': 'footer with foo'}
```
willdurand, line 302: Could we have a test case with `{tag}` in a footer text/title for a shelf that isn't for "random tag"?

src/olympia/shelves/serializers.py
```diff
@@ -134,10 +150,9 @@ def get_addons(self, obj):
         return addons
 
 
-class ShelfEditorialSerializer(ShelfSerializer):
+class ShelfEditorialSerializer(serializers.ModelSerializer):
     title = serializers.CharField()
     footer_text = serializers.CharField()
-    addons = None  # we don't the results for this serializer
```
willdurand, line 140: Why is this not needed anymore?

src/olympia/shelves/models.py
```diff
@@ -27,7 +31,9 @@ class Shelf(ModelBase):
...
         blank=True,
-        help_text='e.g., See more recommended extensions. Will be translated.',
+        help_text='e.g., See more recommended extensions. Will be translated. '
+        '`random-tag` shelves can use {tag} in the text, which will be substituted for '
+        'the random tag selected.',
```
willdurand, line 36: Isn't django generating a migration file for this kind of changes?

src/olympia/shelves/models.py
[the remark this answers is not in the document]
willdurand: I am asking questions more than suggesting any change. It's true that we have a potential problem with non en-US locales already.

src/olympia/shelves/serializers.py
eviljeff: ShelfEditorialSerializer now isn't subclassing another serializer that had defined an `addons` field.

src/olympia/shelves/models.py
eviljeff: it would if the migration command was run, but as it's a no-op at the database level we don't generally do it.  I could though?

willdurand: We usually commit these migration files eventually so maybe we could add it here to avoid having to add it later yeah

willdurand (approved):
thanks

PR #17696, opened 2021-08-11
replace use of django's is_safe_url; consolidate use in scanners

### LINKED ISSUE #16491: replace is_safe_url with url_has_allowed_host_and_scheme (after django2.2 suport is dropped)
from django release notes:
> To avoid possible confusion as to effective scope, the private internal utility is_safe_url() is renamed to url_has_allowed_host_and_scheme().

This is just a rename from 3.x - it'll be an easy fix to do once 3.2 is out and we remove support for 2.2 (it'd be an unnecessary hassle beforehand imo)

eviljeff:
fixes #16491 + bit of a rewrite so we don't have the same-ish code in two places.


src/olympia/scanners/tests/test_admin.py
```diff
@@ -1764,25 +1763,3 @@ def test_matching_filenames_in_changelist(self):
             ),
         ]
         assert [link.attrib['href'] for link in links] == expected
-
-
-class TestIsSafeUrl(TestCase):
-    def test_enforces_https_when_request_is_secure(self):
```
willdurand, line 1770: shouldn't we have these tests for amo/utils.py?

eviljeff: I assumed `_is_safe_url `was already tested 😬 (I guess not)

eviljeff: https://github.com/mozilla/addons-server/pull/17696/commits/dab2a4fed49b0adaee2b598c2a708798c16be34e

diox (approved):

PR #17704, opened 2021-08-13
define (necessary) unique index and constraint names

### LINKED ISSUE #16497: Change index & constraint names to be unique amongst models (post django3.2)
in django3.2 there are additional system checks which enforce that constraint and indexes have unique names, even accross different apps/models.

```
SystemCheckError: System check identified some issues:

ERRORS:
?: (models.E030) index name 'addon_id' is not unique among models: addons.Preview, bandwagon.CollectionAddon, ratings.Rating, tags.AddonTag, versions.Version.
?: (models.E030) index name 'application_id' is not unique among models: applications.AppVersion, bandwagon.Collection.
?: (models.E030) index name 'created_idx' is not unique among models: abuse.AbuseReport, activity.ActivityLog, addons.Addon, bandwagon.Collection, bandwagon.CollectionAddon, files.File.
?: (models.E030) index name 'listed' is not unique among models: addons.AddonUser, bandwagon.Collection.
?: (models.E030) index name 'status' is not unique among models: addons.Addon, files.File.
?: (models.E030) index name 'user_id' is not unique among models: bandwagon.CollectionAddon, users.UserNotification.
?: (models.E032) constraint name 'addon_id' is not unique among models: addons.AddonCategory, addons.AddonUser.
```

eviljeff:
fixes #16497 so we can stop silencing the system checks.

I changed the index/constraint names on the smaller tables where possible, but there were some duplicates among addons and files models so some were unavoidable (I changed addons over files)


diox:
Before merging this, we need to gather the SQL changes (with `manage.py sqlmigrate`) and try to apply them on a replica to see how long it would take. I suspect it's going to be quite painful...

eviljeff:
Thinking about these changes more, in mysql5.7+ there's a `ALTER TABLE RENAME INDEX foo TO baa` statement that we can swap out for django's generated `CREATE` and `DROP`, which I hope will be much quicker.

diox:
Yeah, we could try that. Another thing to try would be to group the statements in a single `ALTER` for each table, which should be faster. We could then use django migrations `state_operations` argument to `RunSQL` to reconciliate custom SQL with the operations django migrations wants to make.

eviljeff:
I played around with a few different ways of doing this that didn't require a massive amount of refactoring to what django generates and https://github.com/mozilla/addons-server/pull/17704/commits/b963dd4c46cccf416180cf60d13b32c8b026f266 seems to work locally (I reversed the addons app migrations and that seemed to work)

diox:
That looks like a good path forward, we might need this again later. Could you add some tests for the 2 migration operations ?

diox (approved):
:+1: so nice to see

diox:
Please call out the migrations in the push doc so that we can monitor the deploy on all envs more closely in case it takes a while.

PR #17707, opened 2021-08-13
import developer blog posts as json (replace rdf)

### LINKED ISSUE #16724: change blog import to use JSON api rather than xml feed
spun off #16711 

>>> Also, behind the scenes, we fetch blog entries to show them in devhub, and that feedparser.parse() on the Atom feed

> >Maybe we can replace that with something json based soon with the new blog?

> We could do this now, in fact - the Rest API is currently active with no authentication for read-only purposes: https://blog.mozilla.org/addons/wp-json/wp/v2/posts

eviljeff:
fixes #16724 

This is also the only place the `dateutil` package is used in live code, but it's (annoyingly) still used in tests.  Could remove in a follow-up (if it's worth the effort for the benefit - doubtful but GFB perhaps).


diox:
FWIW I think this should be done in a separate issue/PR but `dateutil` is used in `assertCloseToNow` to allow us to pass a string as the `dt` argument instead of a `datetime`, but I'm not sure we actually use that anywhere... So it might be easy to remove.

willdurand (approved):
Looks good, thanks!

PR #17722, opened 2021-08-16
Replace RawSQLModel usage in reviewer tools queues with pure ORM queries

### LINKED ISSUE #15431: Remove reviewers/sql_models.py / RawSQLModel
For the longest time reviewer tools in AMO have depended on a custom pseudo-ORM implemented in 
https://github.com/mozilla/addons-server/blob/master/src/olympia/reviewers/sql_model.py.

The main purpose was to work around limitations of the Django ORM 10 years ago, but now this is a pain to maintain and can make things harder to optimize (like counts) or force code duplication. We should remove it and replace it with nice manager methods and querysets to fetch reviewer queues that are still using it.

diox:
Fixes #15431


src/olympia/addons/models.py
```diff
@@ -237,57 +238,137 @@ def not_disabled_by_mozilla(self):
...
         self,
         admin_reviewer=False,
-        admin_content_review=False,
-        show_pending_rejection=False,
-        listed=True,
+        content_review=False,
+        theme_review=False,
+        exclude_listed_pending_rejection=True,
+        select_related_fields_for_listed=True,
+        select_related=True,
```
eviljeff, line 245: `select_related` is unused?

src/olympia/addons/models.py
```diff
@@ -237,57 +238,137 @@ def not_disabled_by_mozilla(self):
...
                 _current_version__reviewerflags__pending_rejection__isnull=True
             )
         if not admin_reviewer:
-            if admin_content_review:
+            if content_review:
                 qs = qs.exclude(reviewerflags__needs_admin_content_review=True)
+            elif theme_review:
+                qs = qs.exclude(reviewerflags__needs_admin_theme_review=True)
             else:
                 qs = qs.exclude(reviewerflags__needs_admin_code_review=True)
```
eviljeff, line 290: is it any slower to just always exclude all of the`reviewerflags__needs_admin_xxx_review.` if not `admin_reviewer`?  If so, less complexity and we don't need to pass in `content_review` and `theme_review`

src/olympia/addons/models.py
```diff
@@ -237,57 +238,137 @@ def not_disabled_by_mozilla(self):
...
+        return (
+            # We passed select_related_fields_for_listed=False but there are
+            # still base select_related() fields in that queryset that are
+            # applied in all cases that we don't want.
+            qs.select_related(None)
+            .filter(filters)
+            .select_related('reviewerflags')
+            .annotate(
+                first_version_nominated=Min('versions__nomination'),
+                latest_version=Func(F('versions__version'), function='LOWER'),
```
eviljeff, line 351: why we do we need to apply the `LOWER` function?

src/olympia/addons/models.py
```diff
@@ -237,57 +238,137 @@ def not_disabled_by_mozilla(self):
...
+    def get_addons_with_unlisted_versions_queue(self, admin_reviewer=False):
+        qs = self.get_base_queryset_for_queue(
+            select_related_fields_for_listed=False,
+            exclude_listed_pending_rejection=False,
+            admin_reviewer=admin_reviewer,
+        )
+        return (
+            qs.filter(versions__channel=amo.RELEASE_CHANNEL_UNLISTED)
+            .exclude(status=amo.STATUS_DISABLED)
+            .select_related(None)
```
eviljeff, line 364: what does `.select_related(None)` do?

eviljeff (approved):
\o/ finally gone!

src/olympia/addons/models.py
diox: Ah yes, I refactored this method a few times and forgot to remove that parameter

src/olympia/addons/models.py
diox: It's a trick I should document in a comment. 

- If you just do `annotate(latest_version=F('versions__version'))` with no other annotations, everything is fine.
- If you just do `annotate(first_version_nominated=Min('versions__nomination'))`, then Django detects you want a `GROUP BY addon.id`, everything is fine.
- If you combine them with `.annotate(latest_version=F('versions__version'), first_version_nominated=Min('versions__nomination'))`, then Django detects you want a `GROUP BY addon.id`, but also adds the other field to it, causing a `GROUP BY addons.id, versions.version` which we don't want at all!
- Adding the `Func()` prevents that (Django can no longer detect what to add to the `GROUP BY`, as you're not annotating with a field directly), but you can't have an empty `Func()`, so I used `LOWER` as it's harmless here.

src/olympia/addons/models.py
diox: It removes the previous `select_related()` that have been added to the queryset so far by `get_base_queryset_for_queue()`. I'll add a comment.

src/olympia/addons/models.py
diox: 4175920ae1076a4d47f1f70b7f899a8359892ebd adds some comments and also a test for the query being generated

src/olympia/addons/models.py
diox: probably not that much, but it wouldn't match the behavior we have - the fields were separated on purpose, so I didn't want to touch that logic.

PR #17724, opened 2021-08-16
Remove reviewer pseudonyms

### LINKED ISSUE #17318: Remove reviewer pseudonyms and no longer expose reviewer names

The reviewer pseudonyms shown in review emails and on devhub were a nice idea but revealed some issues that at least in some parts don't really address the root problem and can (and has) cause additional confusion for developers.

Let's remove the pseudonyms altogether.

The pseudonym is not helpful to the developer. Also, some reviewers have been mistaken as bots due to their pseudonym. There are more issues on the reviewer privacy side.

### What happened?

Reviewer pseudonyms are shown on devhub and in review emails, potentially in other places as well.

Don't expose reviewer pseudonyms (or real names).

Review email can just refer to "a reviewer", or "a human reviewer".

Attributions to reviewers on devhub can say "Add-ons Review Team" pr similar.

wagnerand:
Fix #17318.

### DIFF
--- a/docs/topics/api/v4_frozen/accounts.rst
+++ b/docs/topics/api/v4_frozen/accounts.rst
@@ -58,7 +58,6 @@ If you have `Users:Edit` permission you will see these extra fields for all user
 
     :>json boolean deleted: Is the account deleted.
     :>json string|null display_name: The name chosen by the user.
-    :>json string|null reviewer_name: The reviewer name chosen by the user, if set. Only available for users with any kind of reviewer permission.
     :>json string email: Email address used by the user to login and create this account.
     :>json string fxa_edit_email_url: The configured URL for editing the user's email on FxA.
     :>json string last_login: The date of the last successful log in to the website.
--- a/src/olympia/activity/models.py
+++ b/src/olympia/activity/models.py
@@ -14,6 +14,7 @@ from django.utils.functional import cached_property
 from django.utils.translation import gettext
 
 import markupsafe
+from olympia.access.acl import is_user_any_kind_of_reviewer
 
 import olympia.core.logger
 
@@ -655,11 +656,11 @@ class ActivityLog(ModelBase):
     def author_name(self):
         """Name of the user that triggered the activity.
 
-        If it's a reviewer action that will be shown to developers, the
-        `reviewer_name` property is used if present, otherwise `name` is
-        used."""
+        If it's a reviewer action that will be shown to developers, use a generic
+        Review team name"""
         if self.action in constants.activity.LOG_REVIEW_QUEUE_DEVELOPER:
-            return self.user.reviewer_name or self.user.name
+            if is_user_any_kind_of_reviewer(self.user):
+                return 'Firefox Add-on Review Team'
         return self.user.name
 
     @classmethod
--- a/src/olympia/activity/tests/test_views.py
+++ b/src/olympia/activity/tests/test_views.py
@@ -244,9 +244,9 @@ class TestReviewNotesViewSetList(ReviewNotesViewSetDetailMixin, TestCase):
             'fiiiine', amo.LOG.REVIEWER_REPLY_VERSION, self.days_ago(0)
         )
         self._login_developer()
-        with self.assertNumQueries(17):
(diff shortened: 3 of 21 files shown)


docs/topics/api/v4_frozen/accounts.rst
```diff
@@ -58,7 +58,6 @@ If you have `Users:Edit` permission you will see these extra fields for all user
 
     :>json boolean deleted: Is the account deleted.
     :>json string|null display_name: The name chosen by the user.
-    :>json string|null reviewer_name: The reviewer name chosen by the user, if set. Only available for users with any kind of reviewer permission.
```
diox, line 61: MMM normally since it's frozen I'd say this needs a shim for backwards-compatibility... But given it was only restricted to those specific users, maybe that's fine...

src/olympia/activity/tests/test_views.py
```diff
@@ -244,9 +244,9 @@ def test_queries(self):
             'fiiiine', amo.LOG.REVIEWER_REPLY_VERSION, self.days_ago(0)
         )
         self._login_developer()
-        with self.assertNumQueries(17):
+        with self.assertNumQueries(18):
             # - 2 savepoints because of tests
-            # - 2 user and groups
+            # - 3 user and groups
```
diox, line 249: Why ?

diox: mmm yeah. It scales with the number of items returned, so that would be bad and should be avoided.

src/olympia/activity/models.py
```diff
@@ -655,11 +656,11 @@ def __html__(self):
...
         """Name of the user that triggered the activity.
 
-        If it's a reviewer action that will be shown to developers, the
-        `reviewer_name` property is used if present, otherwise `name` is
-        used."""
+        If it's a reviewer action that will be shown to developers, use a generic
+        Review team name"""
         if self.action in constants.activity.LOG_REVIEW_QUEUE_DEVELOPER:
-            return self.user.reviewer_name or self.user.name
+            if is_user_any_kind_of_reviewer(self.user):
```
diox, line 662: Per comment below, you should remove this check, if will just create a bunch of additional queries (~ one per item).

Instead, I would suggest to just return the generic name always in this property for reviewer actions, and use something completely different when you know ahead of time you're showing content for a reviewer.

wagnerand: I am not sure I follow this. As far as I understand, this method is called both for reviewer and developer replies. If the reply comes from a reviewer, we want to use _Firefox Add-on Review Team_ instead, but if the sender is a developer, it should use the developer name.

diox: You can know that with the `action` and display the reviewer name for reviewer actions and the user name for developer actions. If that's all you think we need, you can use that (I think `review_action` property on the action should be enough) and stop there.

What I thinking about is, in reviewer or admin contexts, you might want to see which reviewer it actually was. If that's the case you would need to use something different when displaying the activity, in those contexts, to avoid the generic name and show the true user name.

wagnerand: Ah, but is `author_name` even used in reviewer or admin contexts?

diox: I don't know, maybe not (which would help), you'll have to find out :)

bobsilverberg:
@wagnerand @eviljeff I am planning to work on #17318 and I am curious, why was this closed without merging. Was a decision made not to fix this issue, were there major problems with this patch, was it a question of @wagnerand not having the time to work on the patch, or was it something else?

eviljeff:
> @wagnerand @eviljeff I am planning to work on #17318 and I am curious, why was this closed without merging? Was a decision made not to fix this issue, were there major problems with this patch, was it a question of @wagnerand not having the time to work on the patch, or was it something else?

I can't recall, but given it was closed about 2 weeks after the review it was likely because I asked on slack/matrix if it was being worked on, and then closed as @wagnerand didn't have time to fix it up.  (I could search through logs to confirm)

bobsilverberg:
Thanks @eviljeff. I spoke to Andreas on Slack about it and he confirms the same.

Issue #17727, opened 2021-08-17
Extract client IP from request headers when behind the CDN


diox:
When behind the CDN, we can no longer use `REMOTE_ADDR` directly as nginx is passing it, because it will be the CDN's. Instead we need to use `HTTP_X_FORWARDED_FOR`, but only if the request actually came from the CDN - if it was a direct request to the origin (which can happen as the origin is public, though we don't advertise its domain) we want to continue using `REMOTE_ADDR`.

This is important for rate limiting and proper ip addresses logging to work (geolocation should already be ok thanks to us using the headers provided by the CDN).

diox:
Quick & dirty summary for future reference:

When we're behind the CDN, the actual client IP is in `HTTP_X_FORWARDED_FOR`... second to last. The last being the IP of the CDN server that made the request to the origin, because we also have a load balancer between nginx and the CDN.

When we're dealing with the origin directly, we only go through the load balancer, so the client IP is in  `HTTP_X_FORWARDED_FOR`, in the last position. In that particular case though, `REMOTE_ADDR` is already the client IP.

diox:
QA: verify that your IP is recorded properly when you make requests to endpoints that record it (add-on upload, ratings, logging in)

diox:
(Re-opening temporarily because the secret necessary for this to work isn't on dev yet. Note that this shouldn't affect stage or prod at the moment even when it lands there.)

AlexandraMoga:
I've verified the IP records on -dev and these are my results (I've switched IPs to make sure that the changes are properly recorded in such scenarios as well):

- the IP is correctly recorded when a user logs in - I've checked the Last login ip filed in users admin
- the IP is correctly recorded when a submission restriction is hit, like `DeveloperAgreementRestriction` or `IPNetworkUserRestriction`
- the IP is correctly recorded when a rating is posted

There are still some activity logs recorded when an add-on upload rate limit is reached, but rate limiting is not active on dev, so I'll have to check those later on stage.

I've also made a quick check on add-on regional restrictions to make sure those continue to work as well.

PR #17728, opened 2021-08-17
Add view to debug client information (IP, detected geolocation, user agent)

diox:
Supports #17727

### DIFF
--- a/src/olympia/amo/views.py
+++ b/src/olympia/amo/views.py
@@ -6,7 +6,7 @@ import django
 from django import http
 from django.conf import settings
 from django.contrib.sitemaps.views import x_robots_tag
-from django.core.exceptions import ViewDoesNotExist
+from django.core.exceptions import PermissionDenied, ViewDoesNotExist
 from django.core.paginator import EmptyPage, PageNotAnInteger
 from django.db.transaction import non_atomic_requests
 from django.http import Http404, HttpResponse, JsonResponse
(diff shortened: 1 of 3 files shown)


src/olympia/amo/views.py
```diff
@@ -55,6 +55,20 @@ def monitor(request):
...
 
+@never_cache
+@non_atomic_requests
+def client_info(request):
+    keys = (
+        'REMOTE_ADDR',
+        'HTTP_USER_AGENT',
+        'HTTP_X_FORWARDED_FOR',
+        'HTTP_X_REQUEST_VIA_CDN',
+        'HTTP_X_COUNTRY_CODE',
```
eviljeff, line 66: Is any of this potentially secret-ish, to the extent we wouldn't want it exposed on an unauthenticated url?

eviljeff (approved):

src/olympia/amo/views.py
diox: Eventually yes - `HTTP_X_REQUEST_VIA_CDN` will be once the CDN is in production. I could leave it for now but actually, that's a good call, let me hide it by just exposing whether the value matches the expected secret.

diox: Actually... even just exposing that makes me nervous. Instead of redacting some of the values, I'm going to only allow this on dev, that should be enough. Once the logic is in place and tested we can probably remove this view anyway, it just makes it easier to develop.

src/olympia/amo/views.py
```diff
@@ -55,6 +55,22 @@ def monitor(request):
     return http.HttpResponse(json.dumps(status_summary), status=status_code)
 
 
+@never_cache
+@non_atomic_requests
+def client_info(request):
+    if getattr(settings, 'ENV', None) != 'dev':
```
eviljeff, line 61: having a setting that defaulted to `True` in the dev settings file but was settable via env files would give us more flexibility if it needed to be tested on stage, etc, too but if it's a short term thing then this works.

PR #17764, opened 2021-08-23
Add indexes against columns containing ip addresses

### LINKED ISSUE #17734: IP search times out

Follow up of #14547.

### What happened?

`models/users/userprofile/?q=<IP>` times out with 504.

No timeout, results are displayed.

diox:
Fixes #17734

But wait until this has been tested on a replica to see how long it would take to migrate.


src/olympia/users/migrations/0006_auto_20210823_1247.py
```diff
@@ -0,0 +1,25 @@
...
+    ]
+
+    operations = [
+        migrations.AddIndex(
+            model_name='userprofile',
+            index=models.Index(fields=['last_login_ip'], name='users_last_login_ip_2cfbbfbd'),
+        ),
+        migrations.AddIndex(
+            model_name='userrestrictionhistory',
+            index=models.Index(fields=['ip_address'], name='u_restrict_hist_ip_4376df32'),
```
eviljeff, line 19: we have `LongNameIndex` to avoid these abbreviated names, fwiw.  (applies to the other cryptically named indexed below too)

diox: Ah, I was trying to remember how we handled that... That's much better, I'll use that.

eviljeff (approved):

src/olympia/ratings/migrations/0004_auto_20210823_1255.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dependencies = [
+        ('ratings', '0003_auto_20210813_0941'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='rating',
+            name='ip_address',
+            field=models.CharField(default='0.0.0.0', max_length=45),
```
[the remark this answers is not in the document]
diox: Yeah, I checked all envs and it's fine. In fact, I don't think we even need 45 - I used that for consistency with the rest of our code but Django's `GenericIPAddressField`, which we should consider using, uses 39 (a maximum of 8 times 4 characters separated with a  "." for IPv6)

diox:
https://mozilla-hub.atlassian.net/browse/SVCSE-65 for the production replica run. Ops are ok with it, worst was the `users` table but still ok for a few minutes.

PR #17765, opened 2021-08-23
implement new uploads to the addon api, for listed too.

eviljeff:
Basic flow:
1) Upload the file to be validated via POSTing to `api/v5/addons/submission/` to create a FileUpload (returns a serialized response that contains the upload UUID).  The upload is where `channel` is set.
2) optional - you can check on validation status at `api/v5/addons/submission/<uuid>`
3) with a valid upload, submit it as a new addon to `api/v5/addons/addon/` with json like `{"categories": {"firefox": ["bookmarks"]}, "version": {"upload": "<uuid>", "license": 7}}`
4) or can be submitted as a new version to an existing addon via `api/v5/addons/addon/<addon slug or id or guid>/versions/` with json like `{"upload": "<uuid>", "license": 7}`

Known limitations:
- you can only select from pre-defined licenses - and only by specifying a numeric license id.  So worked needed to:
  - select from predefined licenses by slug or some other mnemonic 
  - maybe make the field optional (for listed versions only?) and default to all rights reserved, or the previous license for subsequent versions
- `FileUploadViewSet` needs refactoring to move most/all of the creation logic into `FileUploadSerializer`
- categories are required to create an add-on but aren't really needed unless the add-on has listed versions
- For the first listed version we want to require all the necessary metadata (name, summary, categories, version license) being defined if not available in the manifest, but not really for unlisted versions.  This will necessitate some refactoring - probably moving some validation from `AddonSerializer` to `VersionSerializer`.
- source code submission isn't implemented
- there are some fields on `AddonSerializer` that should be writable but are currently either `SerializerMethodField`s or custom `Field`s/`Serializer`s that don't have a working write implementation.
- setting `default_locale` did unexpected things during Addon creation with respect to data in the manifest so left it read only.
- Only `CreateModelMixin` has been added to the views to limit the scope of the changes/tests.  The serializers should handle updates without a massive amount of extra work (in theory!).
- probably many other things

when the feature is launched #7811 will be fixed, but it's a way off at this point


eviljeff:
How we handle accepting developer agreement is an interesting problem - I guess we implement the ability to upload the flag on the accounts api?  (a problem for the internal auth only; web token auth users will have already accepted to get the API key)

eviljeff:
I'm considering moving all of the random fixes/changes I made elsewhere into a separate patch, so this can be focused on new code relating to the api

eviljeff:
@diox changed my mind - I was just looking for an initial review to confirm it makes sense but I'm going to split out the refactoring and re-open this as a new PR.

PR #17767, opened 2021-08-24
Pass version channel to submit source page as well (needed to forward it properly)

### LINKED ISSUE #16562: "Cancel and Disable Version" from the source code page should be removed for unlisted addons
1. A developer uploads a file - using the unlisted channel (a new addon or version) and clicks Sign add-on
2. The Source Code page is the next step where there's "Continue" or "Cancel and Disable Version" buttons
3. Click "Cancel and Disable Version" and confirm "Yes, cancel"

Actual result:
The submitted version is in awaiting review state - it will be automatically signed. 

Expected result:
The red button should not be displayed - it has no role since the add-on is already signed. 

Notes:
This is reproducible on all AMO servers - FF85(Win10).

diox:
Fixes #16562

### DIFF
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -1441,7 +1441,10 @@ def _submit_upload(request, addon, channel, next_view, wizard=False):
                 user=request.user,
             )
             version = addon.find_latest_version(channel=channel)
-            url_args = [addon.slug]
+            url_args = [
+                addon.slug,
+                'listed' if channel == amo.RELEASE_CHANNEL_LISTED else 'unlisted',
+            ]
 
         check_validation_override(request, form, addon, version)
         if (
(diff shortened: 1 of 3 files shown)


src/olympia/devhub/views.py
```diff
@@ -1441,7 +1441,10 @@ def _submit_upload(request, addon, channel, next_view, wizard=False):
                 user=request.user,
             )
             version = addon.find_latest_version(channel=channel)
-            url_args = [addon.slug]
+            url_args = [
+                addon.slug,
+                'listed' if channel == amo.RELEASE_CHANNEL_LISTED else 'unlisted',
```
eviljeff, line 1446: you can use `amo.CHANNEL_CHOICES_API[channel]` to get the string from the channel constant

eviljeff (approved):

src/olympia/devhub/views.py
diox: I know, but that feels wrong - it's not an API value at all. That same logic is used in the templates and tests already.

eviljeff: heh, we use the opposite constant, `CHANNEL_CHOICES_LOOKUP` throughout this file, it's just not got `API` in it's name.

PR #17794, opened 2021-08-26
add enable_for_random_shelf field; expose admin to DISCOVERY_EDIT

### LINKED ISSUE #17772: Allow tags to be enabled/disabled for shelf rotation
We want to be able create tags for developers to choose from, but stop them from being randomly chosen for homepage inclusion (until there are enough add-ons tagged with the tag, etc)

### LINKED ISSUE #17784: Add Tags admin to discovery edit permission
Scott needs to be able to add/remove tags as part of the homepage editorial process.

eviljeff:
fixes #17784; fixes #17772


src/olympia/tags/migrations/0007_tag_enable_for_random_shelf.py
```diff
@@ -0,0 +1,29 @@
...
+            name='enable_for_random_shelf',
+            field=models.BooleanField(default=True, null=True),
+        ),
+        migrations.RunPython(set_tag_default),
+        migrations.AlterField(
+            model_name='tag',
+            name='enable_for_random_shelf',
+            field=models.BooleanField(default=True),
+        ),
+    ]
```
diox, line 29: The reason we generally make new columns nullable at first and then remove the null in a following push is to handle creations of new model instances with the old code but with the new SQL schema *during the push period*. There is no benefit in trying to do both in the same migration: the potential for failure is the same.

By default, Django behaves exactly in the same way, and would fail in the same way if a tag is added during the push, except the migration it generates is a lot more simple:
```python
migrations.AddField(model_name='tag', enable_for_random_shelf', models.BooleanField(default=True))
```

Produces the following SQL:
```sql
ALTER TABLE `tags` ADD COLUMN `enable_for_random_shelf` bool DEFAULT b'1' NOT NULL;
ALTER TABLE `tags` ALTER COLUMN `enable_for_random_shelf` DROP DEFAULT;
```

For a model like `Tag`, where we now fully control the creations, it's fine to not bother with that, so you could just let Django generate the migration without doing anything custom.

diox (changes requested):

src/olympia/tags/migrations/0007_tag_enable_for_random_shelf.py
eviljeff: I was just a little distrustful of django's ability to backfill the existing records with the new column default.  I'll replace it.

src/olympia/shelves/models.py
```diff
@@ -67,9 +67,10 @@ def get_count(self):
 
     @cached_property
     def tag(self):
+        tag_qs = Tag.objects.filter(enable_for_random_shelf=True).order_by('?')
         return (
-            Tag.objects.order_by('?').first().tag_text
-            if self.endpoint == self.Endpoints.RANDOM_TAG
+            tag.tag_text
```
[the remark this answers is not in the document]
eviljeff: https://github.com/mozilla/addons-server/pull/17794/commits/c2f1b0505507104b24800ccc95e5d6865e446ce9

diox (approved):

PR #17795, opened 2021-08-26
Version.files -> Version.file

### LINKED ISSUE #16465: Expose File as a single field instead of a list for each Version
Once https://github.com/mozilla/addons-server/issues/16464 is done, `Version` will have a single `File`, all the time. To further clean it up, we'll want to make `File` a `OneToOneField` or even merge `File` and `Version` together... while ensuring that 1) we can deploy without errors (living on the old tag for a short while) 2) the API backwards-compatibility promises are still kept.

eviljeff:
fixes #16465 ~(or will when the remaining 5 tests pass - assuming they don't point to some larger problems)~

There is follow-up clean-up needed to drop `Version.all_files`, `Version.current_file`, and possibly some of the other helper functions, too (the diff was already huge)


src/olympia/files/models.py
```diff
@@ -42,9 +42,7 @@ class File(OnChangeMixin, ModelBase):
     id = PositiveAutoField(primary_key=True)
     STATUS_CHOICES = amo.STATUS_CHOICES_FILE
 
-    version = models.ForeignKey(
-        'versions.Version', related_name='files', on_delete=models.CASCADE
-    )
+    version = models.OneToOneField('versions.Version', on_delete=models.CASCADE)
```
eviljeff, line 45: Something I noticed during fixing the tests was `SaveUpdateMixin.reload()` doesn't handle reverse relations for `OneToOneField`s - i.e. `version.reload` doesn't reload `version.file` - you have to specifically do `version.file.reload()`.  We should look into dropping our special sauce and just use django's `.refresh_from_db`

src/olympia/files/migrations/0014_auto_20210824_1722.py
```diff
@@ -0,0 +1,20 @@
...
+        ('files', '0013_auto_20210830_1424'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='file',
+            name='version',
+            field=models.OneToOneField(on_delete=django.db.models.deletion.CASCADE, to='versions.version'),
+        ),
+    ]
```
eviljeff, line 20: This generates:
```sql
ALTER TABLE `files` DROP FOREIGN KEY `files_version_id_44ed234d_fk_versions_id`;
DROP INDEX `files_version_id_44ed234d_fk_versions_id` ON `files`;
ALTER TABLE `files` ADD CONSTRAINT `files_version_id_44ed234d_uniq` UNIQUE (`version_id`);
ALTER TABLE `files` ADD CONSTRAINT `files_version_id_44ed234d_fk_versions_id` FOREIGN KEY (`version_id`) REFERENCES `versions` (`id`);
```
which seems partially redundant? - only adding the `UNIQUE` constraint is necessary?

eviljeff:
(test fail in main job is unrelated)

src/olympia/files/migrations/0014_auto_20210824_1722.py
diox: That's a bit annoying... I'm guessing django has no specific migration path from `ForeignKey` to `OneToOneField` so it has to drop everything to re-create... You could consider doing the relevant operation in a `RunSQL` with a `state_operations=[migrations.AlterField(...)]` to avoid this ?

eviljeff: btw, I just double checked the sql statements don't do anything else by running locally and comparing the outputs of `show create table files` before and after.  
Confirming that the end result is
```KEY `files_version_id_44ed234d_fk_versions_id` (`version_id`)``` being replaced with 
```UNIQUE KEY `files_version_id_44ed234d_uniq` (`version_id`)```

eviljeff:
Filed follow-ups:
https://github.com/mozilla/addons-server/issues/17838 (generally replacing `reload()`
https://github.com/mozilla/addons-server/issues/17839 (api work)
https://github.com/mozilla/addons-server/issues/17837 (everything else)

PR #17829, opened 2021-09-01
Fix for test_temporary_worktree test failure

### LINKED ISSUE #17828: fix failing `test_temporary_worktree` test
Since https://github.com/mozilla/addons-server/commit/fdfc355fa18237752285c99d6628b369138a72fd CI has been failing with a single test - `test_temporary_worktree`.  

First thoughts were that the pytest bump from 6.2.4 to 6.2.5 was to blame but when that reverting that didn't fix it, further investigation pointed to a python upgrade from 3.8.11 to 3.8.12 (we don't specify the minor python version in our `Dockerfile` and circleci config - just 3.8 - and the latest version in the prebuilt images gets used).

https://github.com/eviljeff/olympia/commit/4927df46a86c1c06c1ae82ed90bf9441fd34eac5 confirmed that specifying 3.8.11 fixed the test fail. Upgrading to the latest 3.9 python didn't fix it either - there was a 3.9.6 release around the same time that contained all the same fixes as 3.8.12 had.

diox:
Wait until we've initialized the worktree before running git worktree list, otherwise the behavior is undefined because there could be leftover data in the directory.

Fixes #17828 (hopefully)

### DIFF
--- a/src/olympia/git/tests/test_utils.py
+++ b/src/olympia/git/tests/test_utils.py
@@ -99,10 +99,10 @@ def apply_changes(repo, version, contents, path, delete=False):
 def test_temporary_worktree(settings):
     repo = AddonGitRepository(1)
 
-    output = _run_process('git worktree list', repo)
-    assert output.startswith(repo.git_repository.path)
-
     with TemporaryWorktree(repo.git_repository) as worktree:
+        output = _run_process('git worktree list', repo)
+        assert output.startswith(repo.git_repository.path)
+
         assert worktree.temp_directory.startswith(settings.TMP_PATH)
         assert worktree.path == os.path.join(worktree.temp_directory, worktree.name)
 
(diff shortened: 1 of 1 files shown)


src/olympia/git/tests/test_utils.py
```diff
@@ -99,9 +99,6 @@ def apply_changes(repo, version, contents, path, delete=False):
 def test_temporary_worktree(settings):
     repo = AddonGitRepository(1)
 
-    output = _run_process('git worktree list', repo)
-    assert output.startswith(repo.git_repository.path)
```
eviljeff, line 103: mmm, don't we want this to be correct?  Does it work _after_ the extraction instead?

diox: There are other tests ensuring add-ons are extracted in the right place, this one is just about `TemporaryWorktree`, so :shrug: . Not sure what we gain by testing the output of `git worktree list` here compared to a few lines below.

eviljeff: but we don't test `repo.git_repository.path` is in the output further down, just the name

PR #17832, opened 2021-09-02
Fix test_temporary_worktree test with newer versions of git

diox:
- `repo.git_repository.path` contains the full path with `/.git/`
- `repo.git_repository_path` doesn't

With newer versions of git the path returned doesn't include`.git/`, so avoid checking for that part in the test.

Fixes #https://github.com/mozilla/addons-server/issues/17828


eviljeff (approved):
I'm surprised that kind of change doesn't break more things.  r+

diox:
We're generally not dependent on the behavior of the `git` cli, it's more of a test thing, and only this test really looked at the full repository path

PR #17841, opened 2021-09-03
rm import (legacy) blocklist feature

### LINKED ISSUE #17782: rm blocklist legacy import integration
We have integration that imports the legacy, v2, blocklist from remote settings, Now the import cron job is disabled and no longer used we can remove all the code.

eviljeff:
fixes #17782


bobsilverberg (approved):
Looks good, thanks @eviljeff!

For some reason it seems to be impossible to find the "more red than green" ribbon image using Google, otherwise I would have added it.

PR #17856, opened 2021-09-06
Add waffle switch enabling return to amo for all listed add-ons

### LINKED ISSUE #17826: Open RTAMO flow to all listed add-ons
RTAMO allows an user installing Firefox to be prompted to install the add-on they installed Firefox from (on Windows). This is limited to badged add-ons at the moment, and we want to extend it to all listed.

This needs to be guarded by a separate waffle switch so that the feature can be turned off. The switch needs to be different from the one guarding the RTAMO feature itself, so that RTAMO can be turned off independently. The switch needs to be `off` by default, we'll turn it on when ready.

diox:
Fixes #17826


src/olympia/addons/tests/test_views.py
```diff
@@ -1200,8 +1201,12 @@ class TestAddonSearchView(ESTestCase):
     def setUp(self):
         super().setUp()
         self.url = reverse_ns('addon-search')
+        # Create return to AMO waffle switches used for rta: guid search, then
+        # fetch them once to get them in the cache.
         self.create_switch('return-to-amo', active=True)
+        self.create_switch('return-to-amo-for-all-listed', active=False)
         switch_is_active('return-to-amo')
+        switch_is_active('return-to-amo-for-all-listed')
```
eviljeff, line 1209: heh, this `switch_is_active` statement fooled me - I thought it was an assert (that should have failed!)

eviljeff (approved):

src/olympia/addons/tests/test_views.py
diox: Yeah, that's why I added the comment. We don't really care about the return value, but we do need to trigger a check once so that the result is cached (avoiding database queries in the tests)

PR #17859, opened 2021-09-06
replace xml parsing in encode_header with regex; drop lxml in prod

### LINKED ISSUE #16711: DeprecationWarning: defusedxml.lxml is no longer supported and will be removed in a future release
The following `DeprecationWarning` is thrown when you start addons-server:
```
DeprecationWarning: defusedxml.lxml is no longer supported and will be removed in a future release
```
This is because we import `defusedxml.lxml` in [`src/olympia/lib/safe_lxml_etree.py`](https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/safe_lxml_etree.py) which triggers the deprecation notice because of https://github.com/tiran/defusedxml/issues/38.

Need to find out if we can remove the `defusedxml.lxml` import to get rid of that `DeprecationWarning` and what should we do about it. A couple pointers:
- In `safe_lxml_etree.py` (which is imported from [`src/olympia/lib/safe_xml.py`](https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/safe_xml.py), which is imported in `src/olympia/__init__.py` to hopefully quite early) we already provide a patched `XMLParser` where the default is not to resolve entities. But would that be active when using `fromstring()` as well if we didn't have `defusedxml` monkeypatch ?
- Beside tests, we only really use `lxml` via `fromstring()` in [`src/olympia/versions/utils.py`'](https://github.com/mozilla/addons-server/blob/master/src/olympia/versions/utils.py)s `encode_header()` function to find the `width` and `height` of SVG backgrounds in static themes. Maybe it can all be replaced by something else, although even if we did that we'd need to be careful to stay protected in the future if we introduce more stuff depending on XML parsing later

### LINKED ISSUE #16725: replace use of lxml in encode_header
spun off #16711

we use lxml in [src/olympia/versions/utils.py](https://github.com/mozilla/addons-server/blob/master/src/olympia/versions/utils.py)::encode_header() to parse svg images but it's overkill - we only care about 2 integer values so we could use a much more simplistic parser.

eviljeff:
fixes #16725 (and fixes #16711 too) by dropping the last remaining place we parse xml, in favor of a trivial regex.


src/olympia/versions/utils.py
```diff
@@ -52,12 +52,15 @@ def write_svg_to_png(svg_content, out):
...
+SVG_DIMENSIONS_REGEX = rb'(?=.* width="(?P<width>\d+)")(?=.* height="(?P<height>\d+)")'
+
+
 def encode_header(header_blob, file_ext):
     try:
         if file_ext == '.svg':
-            tree = lxml.etree.fromstring(header_blob)
-            width = int(tree.get('width'))
-            height = int(tree.get('height'))
+            dimensions = re.search(SVG_DIMENSIONS_REGEX, header_blob).groupdict()
```
willdurand, line 61: do we have test coverage for this?

eviljeff: well, some tests failed when I made a mistake 😁

willdurand: Ok. I was wondering what would happen if a svg has more than one tag with `width`/`height` attributes but `re.search()` only returns the first occurrence, so that's good: https://docs.python.org/3/library/re.html#re.search

src/olympia/versions/utils.py
```diff
@@ -52,12 +52,15 @@ def write_svg_to_png(svg_content, out):
         return convert_svg_to_png(temporary_svg.name, out)
 
 
+SVG_DIMENSIONS_REGEX = rb'(?=.* width="(?P<width>\d+)")(?=.* height="(?P<height>\d+)")'
```
willdurand, line 55: What would happen if the SVG is declared like this:

```
<svg height="92" width="720" ...>
```

(width and height attrs inverted)

willdurand (commented):
One more question.

src/olympia/versions/utils.py
eviljeff: added some tests https://github.com/mozilla/addons-server/pull/17859/commits/f395f45f48fb21d6a85baa2971c3360097dce3f5

willdurand (approved):

Issue #17880, opened 2021-09-09
Access easier deleted addon statistics 


ioanarusiczki:
Followup for https://github.com/mozilla/addons-server/issues/17858#issuecomment-916075951

1. An admin should be able to access the stats for a deleted add-on when going from the overview page to another one. For now this is possible by replacing manually  into the URL `None` with ID

ioanarusiczki:
@willdurand 
With an admin account all links from https://addons-dev.allizom.org/en-US/firefox/addon/620747/statistics/?last=30 are accessible now. 
Same for an unlisted deleted addon https://addons-dev.allizom.org/en-US/firefox/addon/620741/statistics/?last=30

Verified on AMO dev - FF91(Win10).

PR #17882, opened 2021-09-10
Fix links in statistics pages for deleted add-ons

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/17880

---

This PR updates the links in the menu of the statistics pages.

### DIFF
--- a/src/olympia/stats/templatetags/jinja_helpers.py
+++ b/src/olympia/stats/templatetags/jinja_helpers.py
@@ -15,6 +15,7 @@ def report_menu(context, request, report, obj):
     if isinstance(obj, Addon):
         tpl = loader.get_template('stats/addon_report_menu.html')
         ctx = {
-            'addon': obj,
+            'has_listed_versions': obj.has_listed_versions(include_deleted=True),
+            'slug_or_id': obj.id if obj.is_deleted else obj.slug,
         }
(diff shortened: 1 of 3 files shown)


src/olympia/stats/templatetags/jinja_helpers.py
```diff
@@ -15,6 +15,7 @@ def report_menu(context, request, report, obj):
     if isinstance(obj, Addon):
         tpl = loader.get_template('stats/addon_report_menu.html')
         ctx = {
-            'addon': obj,
+            'has_listed_versions': obj.has_listed_versions(include_deleted=True),
+            'slug_or_id': obj.id if obj.is_deleted else obj.slug,
```
bobsilverberg, line 19: Why not just use `obj.slug` if it exists, otherwise use `obj.id`, rather than specifically checking `obj.is_deleted`? Or must we use `obj.id` if the add-on has been deleted?

bobsilverberg (approved):
Looks good, thanks @willdurand. I just have one question.

r+wc

src/olympia/stats/templatetags/jinja_helpers.py
willdurand: we use `obj.id` only when the add-on is deleted, indeed.

PR #17889, opened 2021-09-10
rm all legacy blocklist update integration

### LINKED ISSUE #17840: rm blocklist legacy update integration
We have integration that imports that synchronises back Blocks created/edited on AMO that are marked as legacy to the legacy v2 blocklist. The waffle switch to enable the feature has been disabled for a while now and the feature is no longer needed so we can remove the code.

eviljeff:
fixes #17840


bobsilverberg (approved):
Looks good. That's a lot of code removed!

PR #17908, opened 2021-09-15
refactoring to aid new addon submission api

eviljeff:
part of the work for #7811


src/olympia/constants/base.py
```diff
@@ -78,10 +78,12 @@
...
 
 UPLOAD_SOURCE_DEVHUB = 1
-UPLOAD_SOURCE_API = 2
+UPLOAD_SOURCE_SIGNING_API = 2
+UPLOAD_SOURCE_ADDON_API = 3
 UPLOAD_SOURCE_CHOICES = (
     (UPLOAD_SOURCE_DEVHUB, _('Developer Hub')),
-    (UPLOAD_SOURCE_API, _('API')),
+    (UPLOAD_SOURCE_SIGNING_API, _('Signing API')),
+    (UPLOAD_SOURCE_ADDON_API, _('Addon API')),
```
willdurand, line 86: What's the difference between the two?

willdurand (commented):
That looks fine but I don't quite understand why these changes are needed. Could you add a few lines describing why the PR is useful? E.g., why do we have two constants for an "upload source api"? why do we need a mixin instead of a base class for tests?

src/olympia/constants/base.py
eviljeff: `UPLOAD_SOURCE_SIGNING_API` is for uploads that came from the existing signing API, `UPLOAD_SOURCE_ADDON_API` is for uploads that will come from the new add-on submission api

eviljeff:
> That looks fine but I don't quite understand why these changes are needed. 

that's the downside of splitting out a few hundred lines of refactoring from a larger patch to make the new code easier to review - there's no obvious need for the changes.

> Could you add a few lines describing why the PR is useful? E.g., why do we have two constants for an "upload source api"? 

Because there will be two apis for uploading xpis - it's part of #7811

> why do we need a mixin instead of a base class for tests?

We don't _need_ it, but it's functionality is like mixin - to add extra functionality to certain test clases - and most of classes subclass `TestCase` also anyway.  Test classes can subclass different base classes (e.g. TransactionTestCase`) without confusing inheritance hierarchies now.  (And bonus: we don't need to `import UploadTest` as `BaseUploadTest` because it's not going to conflict)

src/olympia/api/tests/__init__.py
```diff
@@ -0,0 +1,46 @@
+from datetime import datetime, timedelta
+
+from django.conf import settings
+
+import jwt
+
+from olympia.api.models import SYMMETRIC_JWT_TYPE, APIKey
+
+
+class JWTAuthKeyTester:
```
eviljeff, line 10: this was moved to avoid circular import conflicts

src/olympia/amo/tests/__init__.py
```diff
@@ -355,6 +356,30 @@ def logout_api(self):
         self.defaults.pop('HTTP_AUTHORIZATION', None)
 
 
+class JWTAPITestClient(JWTAuthKeyTester, APIClient):
```
eviljeff, line 359: I added this so we can switch out `APITestClient` for the JWT version without having to rewrite tests.

src/olympia/devhub/views.py
```diff
@@ -649,22 +649,16 @@ def handle_upload(
...
         filedata.size,
         addon=addon,
-        automated_signing=automated_signing,
-        ip_address=(core.get_remote_addr() or '')[:max_ip_length],
+        channel=channel,
         source=source,
         user=user,
     )
-    # The following log statement is used by foxsec-pipeline.
-    log.info('FileUpload created: %s' % upload.uuid.hex)
```
eviljeff, line 667: moved up to `FileUpload.from_post` (the new addon submission api won't call devhub views)

src/olympia/devhub/views.py
```diff
@@ -738,7 +732,6 @@ def file_validation(request, addon_id, addon, file_id):
         'filename': file_.filename,
         'timestamp': file_.created,
         'addon': addon,
-        'automated_signing': file_.automated_signing,
```
eviljeff, line 741: not actually used in the template

src/olympia/files/models.py
```diff
@@ -94,12 +94,6 @@ def has_been_validated(self):
         else:
             return True
 
-    @property
-    def automated_signing(self):
-        """True if this file is eligible for automated signing. This currently
-        means that either its version is unlisted."""
-        return self.version.channel == amo.RELEASE_CHANNEL_UNLISTED
```
eviljeff, line 101: unused

src/olympia/files/models.py
```diff
@@ -607,6 +611,14 @@ def pretty_name(self):
             return parts[1]
         return self.name
 
+    @property
+    def channel(self):
+        return (
+            amo.RELEASE_CHANNEL_UNLISTED
+            if self.automated_signing
+            else amo.RELEASE_CHANNEL_LISTED
+        )
```
eviljeff, line 620: will be used in the addon submission api upload serializer

src/olympia/versions/models.py
```diff
@@ -511,29 +511,6 @@ def is_compatible_app(self, app):
             )
         return False
 
-    def compat_override_app_versions(self):
```
eviljeff, line 514: unused

src/olympia/signing/serializers.py
```diff
@@ -11,7 +11,7 @@
 from olympia.files.models import FileUpload
 
 
-class FileUploadSerializer(serializers.ModelSerializer):
+class SigningFileUploadSerializer(serializers.ModelSerializer):
```
eviljeff, line 14: I'll be adding a new serializer for `FileUpload` instances (in files/serializers) so having this named the same thing but in a different place was potentially confusing

willdurand:
> that's the downside of splitting out a few hundred lines of refactoring from a larger patch to make the new code easier to review - there's no obvious need for the changes.

Yeah, this is a good thing but a bit more context usually helps, either by linking to a draft PR that contains all the changes or describing important parts that have changed like you just did by leaving comments. Thank you!

src/olympia/files/models.py
```diff
@@ -607,6 +611,14 @@ def pretty_name(self):
             return parts[1]
         return self.name
 
+    @property
+    def channel(self):
+        return (
+            amo.RELEASE_CHANNEL_UNLISTED
+            if self.automated_signing
```
diox, line 618: Ugh, we should file an issue to rename + migrate that to `channel`...

willdurand (approved):

src/olympia/files/models.py
eviljeff: filed https://github.com/mozilla/addons-server/issues/17911

PR #17913, opened 2021-09-15
expose a single file in the version api serializer

### LINKED ISSUE #17839: Expose a single File for each Version in the API
follow-up from #16465 - we've dropped multiple files per version at the model/database level and should do the same in the API.  
Note:
- it will be tricky to deploy without breaking anything. Might want to continue exposing files and file initially, deploy that, then following push start using file, then following push remove files.
- v3/v4 must continue to expose a list of files for backwards compatibility.

eviljeff:
fixes #17839 - exposes both `file` and `files` in v5 for now - we'll drop `files` once frontend is updated.


src/olympia/landfill/management/commands/fetch_prod_addons.py
```diff
@@ -68,18 +68,15 @@ def _get_addons_from_page(self, page, params=None):
...
-
-        file_kw = {}
 
         try:
             file_kw = {
-                'hash': files[0]['hash'],
-                'status': amo.STATUS_CHOICES_API_LOOKUP[files[0]['status']],
-                'size': files[0]['size'],
-                'is_webextension': files[0]['is_webextension'],
+                'hash': version['file']['hash'],
```
bobsilverberg, line 74: Nit: before there was a shortcut `files`, which made these statements a bit shorter. Would it be worthwhile to define a `file = version['file']` above and use that shortcut in all of these statements?

src/olympia/landfill/management/commands/fetch_prod_versions.py
```diff
@@ -73,7 +73,7 @@ def _handle_version(self, data):
             print('Skipping %s (version already exists' % data['version'])
             return
 
-        files_data = data['files'][0]
+        files_data = data['file']
```
bobsilverberg, line 76: Nit: should this be `file_data` now?

bobsilverberg (approved):
Looks good, nice cleanup @eviljeff!

I just added a couple of nits.

r+wc

src/olympia/landfill/management/commands/fetch_prod_addons.py
eviljeff: doesn't really make much difference - especially in landfill code (that's only executed once when a local dev setup is installed) but can do

PR #17914, opened 2021-09-15
implement new uploads to the addon api, for listed too.

### LINKED ISSUE #7811: Upload/submit a new listed add-on version through the API

eviljeff:
fixes #7811 ... or the base work for it at least.  (draft PR was #17765 - copied text over from it)

Basic flow:
1) Upload the file to be validated via POSTing to `api/v5/addons/submission/` to create a FileUpload (returns a serialized response that contains the upload UUID).  Upload was added in #17919.
2) with a valid upload, submit it as a new addon to `api/v5/addons/addon/` with json like `{"categories": {"firefox": ["bookmarks"]}, "version": {"upload": "<uuid>", "license": 7}}`
3) or can be submitted as a new version to an existing addon via `api/v5/addons/addon/<addon slug or id or guid>/versions/` with json like `{"upload": "<uuid>", "license": 7}`

Known limitations:
- you can only select from pre-defined licenses - and only by specifying a numeric license id.  So worked needed to:
  - select from predefined licenses by slug or some other mnemonic 
  - maybe make the field optional (for listed versions only?) and default to all rights reserved, or the previous license for subsequent versions
- categories are required to create an add-on but aren't really needed unless the add-on has listed versions
- For the first listed version we want to require all the necessary metadata (name, summary, categories, version license) being defined if not available in the manifest, but not really for unlisted versions.  This will necessitate some refactoring - probably moving some validation from `AddonSerializer` to `VersionSerializer`.
- source code submission isn't implemented
- there are some fields on `AddonSerializer` that should be writable but are currently either `SerializerMethodField`s or custom `Field`s/`Serializer`s that don't have a working write implementation.
- setting `default_locale` did unexpected things during Addon creation with respect to data in the manifest so left it read only.
- Only `CreateModelMixin` has been added to the views to limit the scope of the changes/tests.  The serializers should handle updates without a massive amount of extra work (in theory!).
- probably many other things

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -6,7 +6,8 @@ from django.urls import reverse
 
 from rest_framework import exceptions, serializers
 
-from olympia import amo
+import olympia.core.logger
+from olympia import activity, amo
 from olympia.accounts.serializers import (
     BaseUserSerializer,
     UserProfileBasketSyncSerializer,
@@ -18,17 +19,20 @@ from olympia.api.fields import (
     OutgoingTranslationField,
     OutgoingURLField,
     ReverseChoiceField,
+    SplitField,
     TranslationSerializerField,
 )
 from olympia.api.serializers import BaseESSerializer
 from olympia.api.utils import is_gate_active
 from olympia.applications.models import AppVersion
 from olympia.bandwagon.models import Collection
-from olympia.constants.applications import APPS_ALL, APP_IDS
+from olympia.blocklist.models import Block
+from olympia.constants.applications import APPS, APPS_ALL, APP_IDS
 from olympia.constants.base import ADDON_TYPE_CHOICES_API
-from olympia.constants.categories import CATEGORIES_BY_ID
+from olympia.constants.categories import CATEGORIES, CATEGORIES_BY_ID
 from olympia.constants.promoted import PROMOTED_GROUPS, RECOMMENDED
-from olympia.files.models import File
+from olympia.files.models import File, FileUpload
+from olympia.files.utils import parse_addon
 from olympia.promoted.models import PromotedAddon
 from olympia.search.filters import AddonAppVersionQueryParam
 from olympia.ratings.utils import get_grouped_ratings
@@ -40,7 +44,7 @@ from olympia.versions.models import (
     VersionPreview,
 )
 
-from .models import Addon, Preview, ReplacementAddon, attach_tags
+from .models import Addon, AddonCategory, Preview, ReplacementAddon, attach_tags
 
 
 class FileSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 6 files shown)


eviljeff:
#17919 is just the upload part - I'll rebase this patch after that one lands.

eviljeff:
rebased and now just focused on creating addons/versions from fileuploads

src/olympia/addons/views.py
```diff
@@ -183,6 +187,7 @@ class AddonViewSet(RetrieveModelMixin, GenericViewSet):
             AllowUnlistedViewerOrReviewer,
         ),
     ]
+    authentication_classes = [JWTKeyAuthentication, WebTokenAuthentication]
```
diox, line 190: It's probably safe to do (ultimately this doesn't let you do that much), but this makes the whole API available with external auth and should be documented.

src/olympia/addons/views.py
```diff
@@ -322,14 +339,19 @@ def get_addon_object(
...
+    ListModelMixin,
+    GenericViewSet,
 ):
     # Permissions are always checked against the parent add-on in
     # get_addon_object() using AddonViewSet.permission_classes so we don't need
     # to set any here. Some extra permission classes are added dynamically
     # below in check_permissions() and check_object_permissions() depending on
     # what the client is requesting to see.
     permission_classes = []
+    authentication_classes = [JWTKeyAuthentication, WebTokenAuthentication]
```
diox, line 354: As above

diox (approved):
Good first stab at it! We should capture all those TODOs and known limitations in an issue (or several...) for follow-up.

src/olympia/addons/views.py
```diff
@@ -285,6 +295,13 @@ def eula_policy(self, request, pk=None):
         )
         return Response(serializer.data)
 
+    def create(self, request, *args, **kwargs):
+        from olympia.signing.views import VersionView  # circular import
+
+        # TODO: consolidate/replicate this behaviour.
+        VersionView().check_throttles(request)
```
eviljeff, line 302: https://github.com/mozilla/addons-server/issues/18023

src/olympia/addons/serializers.py
```diff
@@ -315,8 +346,74 @@ class Meta:
...
+        upload = validated_data.get('upload')
+        parsed_and_validated_data = {
+            **self.parsed_data,
+            **validated_data,
+            'license_id': validated_data['license'].id,
+        }
+        version = Version.from_upload(
+            upload=upload,
+            addon=self.addon or validated_data.get('addon'),
+            # TODO: change Version.from_upload to take the compat values into account
```
eviljeff, line 410: https://github.com/mozilla/addons-server/issues/18024

src/olympia/addons/views.py
eviljeff: Documented in the API docs?  Or elsewhere

diox: In the API docs. For addons (and many others), we currently say `The only authentication method available at the moment is the internal one.` and this wouldn't be true anymore after these changes.

diox: The only thing needed is to remove that whole sentence for addons API I think.

eviljeff:
Filed follow-ups and created a project to hold them: https://github.com/orgs/mozilla/projects/174

PR #17918, opened 2021-09-16
Allow excluding more than one rule on scannerresult admin page

### LINKED ISSUE #17898: Allow excluding more than one rule on scannerresult admin page
The _scannerresult_ admin page currently allows to exclude one scanner rule from the view. Due to changed needs, admins need to be able to exclude more than one rule now.

diox:
Fixes #17898

### DIFF
--- a/src/olympia/scanners/admin.py
+++ b/src/olympia/scanners/admin.py
@@ -148,7 +148,7 @@ class ExcludeMatchedRulesFilter(SimpleListFilter):
                 'selected': selected,
                 'value': lookup,
                 'display': title,
-                'params': cl.get_filters_params()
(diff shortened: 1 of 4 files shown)


willdurand (approved):
Looks good, thanks. An improvement could be to be able to deselect ALL rules in the select box because once you start excluding a rule or two, there is no way to fully clear this filter.

wagnerand:
> Looks good, thanks. An improvement could be to be able to deselect ALL rules in the select box because once you start excluding a rule or two, there is no way to fully clear this filter.

Fwiw, a workaround to clear is to select a random rule, then CTRL+Click (CMD+Click on Mac) that rule to deselect it

PR #17919, opened 2021-09-16
implement fileupload api, part of new addon submisison api

eviljeff:
part of #7811 - just the file upload

1). Upload the file to be validated via POSTing to `api/v5/addons/upload/` to create a FileUpload (returns a serialized response that contains the upload UUID). The upload is where channel is set.
2) optional - you can check on validation status at `api/v5/addons/upload/<uuid>`

### DIFF
--- a/src/olympia/api/urls.py
+++ b/src/olympia/api/urls.py
@@ -2,7 +2,7 @@ from django.urls import include, re_path
 
 from olympia.accounts.urls import accounts_v3, accounts_v4, auth_callback_patterns
 from olympia.amo.urls import api_patterns as amo_api_patterns
-from olympia.addons.api_urls import addons_v3, addons_v4
+from olympia.addons.api_urls import addons_v3, addons_v4, addons_v5
 from olympia.ratings.api_urls import ratings_v3, ratings_v4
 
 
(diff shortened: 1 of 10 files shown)


src/olympia/api/urls.py
```diff
@@ -33,7 +33,20 @@
     re_path(r'^scanner/', include('olympia.scanners.api_urls')),
 ]
 
-v5_api_urls = v4_api_urls + [
+v5_api_urls = [
```
bobsilverberg, line 36: Why this change?

src/olympia/files/tests/test_serializers.py
```diff
@@ -0,0 +1,50 @@
...
+
+        file_upload.update(
+            automated_signing=True,
+            validation=json.dumps([{'something': 'happened'}]),
+        )
+        data = FileUploadSerializer(
+            instance=file_upload, context={'request': self.request}
+        ).data
+        assert data['channel'] == 'listed'
+        assert data['validation'] == [{'something': 'happened'}]
```
bobsilverberg, line 50: Should this include a check for `valid` as well?

src/olympia/files/tests/test_views.py
```diff
@@ -43,3 +54,128 @@ def test_returns_410_when_upload_path_is_falsey(self):
...
+            response = self.client.post(
+                self.list_url,
+                data,
+                format='multipart',
+                REMOTE_ADDR='127.0.3.1',
+            )
+
+        assert response.status_code == 201
+
+        assert FileUpload.objects.count() != upload_count_before
```
bobsilverberg, line 130: Why assert it's not equal? Why not assert it's `upload_count_before + 1`?

src/olympia/files/views.py
```diff
@@ -37,3 +48,48 @@ def serve_file_upload(request, uuid):
...
+        IsSubmissionAllowedFor,
+    ]
+    authentication_classes = [JWTKeyAuthentication, WebTokenAuthentication]
+    lookup_field = 'uuid'
+
+    def get_queryset(self):
+        return super().get_queryset().filter(user=self.request.user)
+
+    def create(self, request):
+        # TODO: move most/all of this logic to the serializer
```
bobsilverberg, line 68: Why not just address this `# TODO` now? Is it something that you are planning as part of a follow-up?

bobsilverberg (approved):
Looks good, nice work @eviljeff! I just have a few comments.

r+wc

src/olympia/files/tests/test_serializers.py
eviljeff: valid won't have changed.  They're tests for the serializer anyway, and there isn't anything special in particular about the `valid` field (`get_validation` has some trivial logic, and `channel` has the choices.)

src/olympia/files/views.py
eviljeff: I'm still not sure if I should make that change - doing things in the serializer is generally the rest framework way, but in this case there is very little that can set or we need to validate, and update functionality isn't appropriate for FileUpload instances. I'll remove the TODO comment, because it's not really a definite TODO any longer.

PR #17934, opened 2021-09-20
Reverse code-search versions order

wagnerand:
Fix mozilla/addons#8495.

### DIFF
--- a/src/olympia/scanners/tasks.py
+++ b/src/olympia/scanners/tasks.py
@@ -336,7 +336,7 @@ def run_yara_query_rule(query_rule_pk):
     )
     if not rule.run_on_disabled_addons:
         qs = qs.exclude(addon__status=amo.STATUS_DISABLED)
-    qs = qs.values_list('id', flat=True).order_by('pk')
+    qs = qs.values_list('id', flat=True).order_by('-pk')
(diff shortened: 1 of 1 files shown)


src/olympia/scanners/tasks.py
```diff
@@ -336,7 +336,7 @@ def run_yara_query_rule(query_rule_pk):
     )
     if not rule.run_on_disabled_addons:
         qs = qs.exclude(addon__status=amo.STATUS_DISABLED)
-    qs = qs.values_list('id', flat=True).order_by('pk')
+    qs = qs.values_list('id', flat=True).order_by('-pk')
```
willdurand, line 339: Do you think this should be an option when creating the query?

wagnerand: No. For a complete result, you need to wait for the task to finish anyway, this just helps me to get a head start.

willdurand:
Why would the `release-master` job run in this PR? 🤔

eviljeff:
> Why would the `release-master` job run in this PR? 🤔

@wagnerand is trying to merge from wagnerand:master so it triggered.

willdurand (approved):
If that works for you, that works for me.

Issue #17941, opened 2021-09-21
Update API docs to indicate that the Recommendations endpoint supports the `app` parameter


willdurand:
The [Recommendations API](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#recommendations) supports the `app` parameter although it isn't documented yet. We should document it because addons-frontend should be passing it!

See also: https://github.com/mozilla/addons-frontend/issues/10912

PR #17942, opened 2021-09-21
Update Recommendations API docs

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/17941

### DIFF
--- a/docs/topics/api/addons.rst
+++ b/docs/topics/api/addons.rst
@@ -492,7 +492,7 @@ Four recommendations are fetched, but only valid, publicly available addons are 
 
 .. http:get:: /api/v5/addons/recommendations/
 
-    :query string app: Set the :ref:`add-on application <addon-detail-application>` for that query. This won't filter the results.
+    :query string app: Set the :ref:`add-on application <addon-detail-application>` for that query. This won't filter the results. Defaults to `firefox`.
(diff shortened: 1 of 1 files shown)


willdurand:
> if we're really recommending users on Android install an add-on that's desktop that seems wrong.

I suppose we could also NOT show the recommendations on Android... I don't think TAAR is supposed to work for Android anyway but I dunno. We could ask Product, maybe.

> But just sticking to the documentation, I think we should mention the default if app isn't specified as a query param (I'm guessing: the urls will default to /firefox/ if the add-on supports both desktop and android)

Ah yes, thanks

docs/topics/api/addons.rst
```diff
@@ -492,6 +492,7 @@ Four recommendations are fetched, but only valid, publicly available addons are
 
 .. http:get:: /api/v5/addons/recommendations/
 
+    :query string app: Set the :ref:`add-on application <addon-detail-application>` for that query. This won't filter the results. Defaults to `firefox`.
```
eviljeff, line 495: It default to `firefox` if firefox compatible but presumably it'd default to `android` if it wasn't?  Meh, I suppose it's not that important.

eviljeff: ``` ``firefox`` ```<- double back tick for literals.

eviljeff (approved):

PR #17944, opened 2021-09-21
upgrade to python39

eviljeff:
fixes mozilla/addons#8130 - seems harmless


eviljeff:
bad base branch selection

Issue #17946, opened 2021-09-21
drop `is_webextension` from File model


eviljeff:
All add-ons have been webextensions for years on AMO prod, and we dropped support for parsing installing rdfs in #16726 a few months back, effectively stopping any new uploads even by admins.  We should drop the actual model field, and any lingering supporting code that differentiates between webextension files and non-webextension files.

https://sql.telemetry.mozilla.org/queries/82037/source shows 43 files remaining that are webextension=False.  (They will either need to be deleted beforehand, or in the migration)

---

For QA:

- please verify that git extraction continues to exclude non web-extension files (like themes, langpacks, dicts) as per https://github.com/mozilla/addons-server/issues/14289#issuecomment-630783979
- please re-verify the following issue: https://github.com/mozilla/addons-server/issues/14082

wagnerand:
Those files are signed with the privileged cert. Since they won't run in any recent version of Firefox, I think it's safe to delete them.

ioanarusiczki:
@eviljeff 

I talked with @willdurand and filed a followup for MAD https://github.com/mozilla/addons-server/issues/18065

Since AMO stage looks good I think this can be marked verified - FF92(Win10).

PR #17947, opened 2021-09-21
drop is_webextension from api

### LINKED ISSUE #17658: Remove is_webextension from API responses
split from #17390 

After mozilla/addons-frontend#10716 is fixed, addons-frontend will no longer be using is_webextension from API responses. Therefore, we can remove it from v5 of the API.

eviljeff:
fixes #17658


eviljeff:
> but I am interested in knowing what we'll do about the 50-ish entries in the DB with `is_webextension=False`

covered in #17946

PR #18003, opened 2021-09-30
Extract app & locale info from next_path before redirecting in FxA callback

### LINKED ISSUE #17995: User lands on desktop view instead of /android/ after creating a new account
STR:
1. Create a new account on an android device 
2. Observe the landing page 

Expected result:
https://addons.allizom.org/en-US/android/users/edit is displayed

Actual result:
User lands on https://addons.allizom.org/en-US/firefox/users/edit/ - the desktop view

prod bug

Notes:
- editing something on the page and then updating it with "Create My Profile" lands back on android view.
- not editing the profile and navigating to the homepage using the logo or to any other page - the desktop view will persist. 
- the issue is reproducible on all AMO servers.

diox:
Fixes #17995

### DIFF
--- a/src/olympia/accounts/views.py
+++ b/src/olympia/accounts/views.py
@@ -407,11 +407,11 @@ class AuthenticateView(FxAConfigMixin, APIView):
             # on, we extract that information from the next_path if present
             # and set locale/app on the prefixer instance that reverse() will
             # use automatically.
-            prefixer = get_url_prefix()
-            if prefixer and next_path:
-                splitted = prefixer.split_path(next_path)
-                prefixer.locale = splitted[0]
-                prefixer.app = splitted[1]
+            if next_path:
+                if prefixer := get_url_prefix():
+                    splitted = prefixer.split_path(next_path)
+                    prefixer.locale = splitted[0]
+                    prefixer.app = splitted[1]
             edit_page = reverse('users.edit')
             if next_path:
                 next_path = f'{edit_page}?to={quote_plus(next_path)}'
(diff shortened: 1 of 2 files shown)


src/olympia/accounts/views.py
```diff
@@ -397,8 +398,22 @@ def get(self, request, user, identity, next_path):
...
+            if not is_safe_url(next_path, request):
+                next_path = None
+            # If we just reverse() directly, we'd use a prefixer instance
+            # initialized from the current view, which would not contain the
+            # app information since it's a generic callback, the same for
+            # everyone. To ensure the user stays on the app/locale they were
+            # on, we extract that information from the next_path if present
+            # and set locale/app on the prefixer instance that reverse() will
+            # use automatically.
+            prefixer = get_url_prefix()
```
bobsilverberg, line 410: It seems like the logic on lines 410-414 are only needed if `next_path` is truthy, and we set it to `None` above after the `if` on line 401, so should all this logic also be gated on that `if` statement?

bobsilverberg (approved):
Looks good, thanks for the quick fix @diox! I just had one comment about the logic in the view.

r+wc

src/olympia/accounts/views.py
```diff
@@ -397,8 +398,22 @@ def get(self, request, user, identity, next_path):
                 user = register_user(identity)
             else:
                 reregister_user(user)
+            if not is_safe_url(next_path, request):
+                next_path = None
+            # If we just reverse() directly, we'd use a prefixer instance
+            # initialized from the current view, which would not contain the
+            # app information since it's a generic callback, the same for
+            # everyone. To ensure the user stays on the app/locale they were
+            # on, we extract that information from the next_path if present
```
willdurand, line 407: Does this assume that `next_path` will always have valid app/locale?

PR #18006, opened 2021-09-30
Load addons-server statics from the main domain on dev

### LINKED ISSUE #17988: Load addons-server static files from /static-server on the main domain on dev
Essentially https://github.com/mozilla/addons-frontend/issues/10785 but for addons-server statics. We are currently loading addons-server statics on dev from https://addons-dev-cdn.allizom.org/static/, we want to load them from https://addons-dev.allizom.org/static-server/ (which is served from the CDN as the rest of the main domain) instead.

Once it's verified on dev we can enable it on stage & prod as well.

diox:
Fixes #17988


eviljeff (approved):
n.b while I was looking around for where `STATIC_URL` was used, the `static_url` context processor function seems a bit pointless.

diox:
Agreed, I filed https://github.com/mozilla/addons-server/issues/18012 for this

PR #18007, opened 2021-09-30
Avoid using .pop() on a tuple to disable 'addon-submission-api' gate in prod

diox:


willdurand (approved):
I tried to simulate this in a python console, that looks OK.

Issue #6111, opened 2017-08-04
Catch zips using '\' in file-paths


EnTeQuAk:
See https://github.com/mozilla/addons-linter/issues/1426 for much more details.

tldr; zip doesn't allow paths to contain backslashes but windows did that in the past and users do submit extensions like that. I don't know current behavior and couldn't find anything related to that in sentry so potentially we simply accept zips like that.

Should add at least a test though.

muffinresearch:
>  I don't know current behavior and couldn't find anything related to that in sentry so potentially we simply accept zips like that.

We should *not* accept zips using backslashes in paths and should flag it to developers.

EnTeQuAk:
> We should not accept zips using backslashes in paths and should flag it to developers.

I was just stating my guess of our current behavior, if we do that, we indeed should change it.

muffinresearch:
> I was just stating my guess of our current behavior, if we do that, we indeed should change it. 

Aha ok - it makes more sense now - I was confused by "so potentially we simply accept zips like that" which sounded like a future direction which conflicted with the linter issue. 

Feel free to clarify the original description and delete my comments.

yfdyh000:
Due to use of [`path`](https://nodejs.org/docs/latest/api/path.html). I tried to changes to `path.posix.` or [path-posix](https://www.npmjs.com/package/path-posix), but encountered a variety of errors for build or debugging, did not succeed.

willdurand:
Let's reopen this issue because that could be useful to avoid bugs like https://bugzilla.mozilla.org/show_bug.cgi?id=1732997

willdurand:
The idea would be to reject submissions with such invalid paths during (or before) the `repack` task so that we don't build completely different XPIs.

diox:
The check should probably be added in https://github.com/mozilla/addons-server/blob/2f9f85d6b05256e145c765a4553605b64a4f9a39/src/olympia/files/utils.py#L515. Would need to double-check if somehow `\` isn't used as an escape character in some cases too (which could be valid)

willdurand:
For an invalid zipfile (as per this bug), the backslash in a path looks escaped in the list of filenames:

```
>>> z.namelist()
['path\\to\\file.txt']
```

AlexandraMoga:
[...]

PR #18009, opened 2021-10-01
Raise an error when backslashes are found in zip filenames

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/6111

---

That looks good but I don't know if this is too naive or if this is fine.


src/olympia/files/utils.py
```diff
@@ -513,7 +513,7 @@ def archive_member_validator(archive, member):
 
 
 def _validate_archive_member_name_and_size(filename, filesize):
-    if filename is None or filesize is None:
+    if filename is None or filesize is None or '\\' in filename:
```
diox, line 516: I would add that check a few lines below instead (line 531, where `filename` is being checked). That gives a better, more accurate error message.

willdurand: Heh. I didn't read the function correctly and I thought it would end right below... Yeah, that makes sense.

diox (approved):
Probably fine. We can also add more info about this on extensionworkshop - but the error message should make it clear there is something wrong with the filenames, and the backslashes would be shown here, so I think that's cool.

PR #18031, opened 2021-10-04
Avoid database queries in 500 error page

### LINKED ISSUE #17861: 500 page should not query the database

When a 500 page is displayed, it queries the database (looking for a site notice
in `Config`) but if the 500 is initially caused by MySQL, then we'll end up with
a many errors until the stack is exhausted.

We should not query the database when we display a 500 page, which means:

- no configuration should be retrieved from the DB
- no authentication stuff should be involved
- the 500 page should be as static as possible

The issue with `Config` is that the 500 template inherits a "base" template,
which is where the site notice is displayed.

diox:
Fixes #17861

### DIFF
--- a/src/olympia/amo/tests/test_views.py
+++ b/src/olympia/amo/tests/test_views.py
@@ -92,22 +92,23 @@ class Test404(TestCase):
 
 
 class Test500(TestCase):
-    def test_500_logged_in(self):
-        self.client.login(email=user_factory().email)
-        response = self.client.get('/services/500')
-        assert response.status_code == 500
-        self.assertTemplateUsed(response, 'amo/500.html')
-        content = response.content.decode('utf-8')
-        assert 'data-anonymous="false"' in content
-        assert 'Log in' not in content
-
-    def test_500_logged_out(self):
-        response = self.client.get('/services/500')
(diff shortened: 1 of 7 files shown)


src/olympia/amo/tests/test_views.py
```diff
@@ -92,22 +92,23 @@ def test_404_api_v4(self):
...
+        # We don't even want to show the log in link.
+        assert 'Log in' not in content
+        return content
+
+    def test_500_renders_correctly_with_no_queries_or_auth_even_when_logged_in(self):
+        # Being logged in shouldn't matter for the 500 page.
+        user = user_factory()
+        self.client.login(email=user.email)
+        content = self.test_500_renders_correctly_with_no_queries_or_auth()
+        assert user.email not in content
```
willdurand, line 111: I am wondering if we could use `assertNumQueries()` here.

willdurand (approved):
r+wc, thanks

src/olympia/amo/tests/test_views.py
diox: Since I'm calling `test_500_renders_correctly_with_no_queries_or_auth()` that should be automatically done

PR #18051, opened 2021-10-05
refactor so we're using the exact same throttles for both apis

### LINKED ISSUE #18023: consolidate/refactor submission throttles
currently some of our logic for throttling submissions is contained in VersionView().check_throttles itself, which makes it difficult to reuse in other classes.  We should rewrite it into either a utility function, or entirely within DRF throttle classes that can be reused more easily.

eviljeff:
fixes #18023 - not much actual changes in the end

nb. not replicating the langpack bypass was intentional - releng will need a special endpoint anyway

### DIFF
--- a/src/olympia/addons/views.py
+++ b/src/olympia/addons/views.py
@@ -38,6 +38,7 @@ from olympia.api.permissions import (
     GroupPermission,
     RegionalRestriction,
 )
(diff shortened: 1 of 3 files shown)


willdurand (approved):
I haven't tried the patch locally but that looks good and tests seem to pass, so r+

PR #18071, opened 2021-10-07
Allow app compatibility to be specified at submission time 

eviljeff:
fixes mozilla/addons#8512 for addon/version creation.  

When we expand the allow to allow updates it'll need some more work to reconcile the value of `compatibility`|`compatible_apps` with the existing `ApplicationVersion` instances.  (Or maybe a full rewrite with a serializer for the `ApplicationVersion` instances directly?)


src/olympia/versions/models.py
```diff
@@ -222,7 +222,16 @@ def __str__(self):
...
     @classmethod
-    def from_upload(cls, upload, addon, selected_apps, channel, parsed_data=None):
+    def from_upload(
+        cls,
+        upload,
+        addon,
+        channel,
+        *,
+        selected_apps=None,
+        compatiblity=None,
```
diox, line 232: typo `compatiblity` - which I think doesn't cause issues cause you seem to have made it everywhere you use that argument, but would be good to fix.

src/olympia/versions/models.py
```diff
@@ -222,7 +222,16 @@ def __str__(self):
...
+        addon,
+        channel,
+        *,
+        selected_apps=None,
+        compatibility=None,
+        parsed_data=None,
+    ):
         """
         Create a Version instance and corresponding File(s) from a
         FileUpload, an Addon, a list of compatible app ids, a channel id and
```
diox, line 237: Could you reword that doc to match the parameters ?

eviljeff: https://github.com/mozilla/addons-server/pull/18071/commits/b6f27f0a754fe5214e2cdc474c3af2478c4dfd74

Issue #18012, opened 2021-10-01
Clean up / tighten addons-server CSP config


diox:
When #18010 is fixed and the dust has settled, we can clean up and tighten down the CSP a little bit:
- Remove `https://addons.mozilla.org` from `CSP_BASE_URI`
- Remove references to `PROD_CDN_HOST`
- Remove (hardcoded) reference to `https://static.addons.mozilla.net`
- Remove `CSP_CONNECT_SRC` customization in each env setting file (just need to connect to self, sentry, google analytics)
- Set `CSP_DEFAULT_SRC` to `'none'`
- Remove `'self'` from CSP_CHILD_SRC ? I don't think we need it
- Remove `data:` from `CSP_IMG_SRC ? I don't think we need it either nowadays (`blob:` is required for image uploads)
- Remove `'self'` from `CSP_STYLE_SRC` ? Having the static URL should be enough.

diox:
In the CSP config, we can also:
- Remove `https://addons.mozilla.org` from `CSP_BASE_URI`
- Remove references to `PROD_CDN_HOST`
- Remove (hardcoded) reference to `https://static.addons.mozilla.net`

AlexandraMoga:
I couldn't find any other regressions in Dehub, Revtools, Statistics page caused by this CSP change. 

I'm marking this issue as verified fixed on stage.

PR #18110, opened 2021-10-13
Load addons-server static files and user-uploaded images from the main domain on stage and prod

diox:
Fixes mozilla/addons#8508


bobsilverberg (approved):
This looks good, thanks @diox. I just have one question because I am a bit confused by the relationship between all the settings files. Does this config mean that `PROD_STATIC_URL` and `PROD_MEDIA_URL` will be added to the CSP for *all* environments?

diox:
Yes. I originally changed that too, but realized it's by design, so that the CSP from the default settings is what we see in prod. I filed  https://github.com/mozilla/addons-server/issues/18012 to do more CSP cleanups, and we could consider removing the prod URLs from other envs when implementing that issue.

PR #18141, opened 2021-10-19
enable patches for Version detail api

eviljeff:
fixes mozilla/addons#8527

I went backwards and forwards with a few different ways of handling the problem we have with writing to fields that are `cached_property`s wrapped around one-to-many fields - we have it here with `compatible_apps`, and did with Addon with categories. Implementing full serializers for the underlying models is an option but: we can't set the values with the contents of our ES index, (and there's no in-built support in DRF for creating new instances anyway, so it's not a complete solution).  

I've gone with defining a separate `set_compatible_apps` method, like with categories, and creating the unsaved instances in the field.  

Other options I played around with: could change the `compatible_apps` setter to actually save the instances (anywhere that intended to just set a cached value, such as the ES serializer initialization, would have to be changed so we don't actually try to save the values); still have the setter property as now but also flag (with `ModelState`) when the cached value has been set outside of the model transformer/initialization, and then override the model `save` method to also save the `compatible_apps` instances if they've been flagged (added extra complexity that could end up being pretty confusing to future us).

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -1369,16 +1369,6 @@ class TestVersionViewSetCreate(UploadMixin, TestCase):
             self.url,
             data={
                 **self.minimal_data,
-                # 99 doesn't exist as an appversion
-                'compatibility': {'firefox': {'min': '99.0'}},
-            },
-        )
-        assert response.data == {'compatibility': ['Unknown app version specified']}
-
-        response = self.client.post(
-            self.url,
-            data={
-                **self.minimal_data,
                 # DEFAULT_STATIC_THEME_MIN_VERSION_ANDROID is 65.0 so it exists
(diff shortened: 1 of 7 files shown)


src/olympia/addons/tests/test_views.py
```diff
@@ -1412,6 +1423,214 @@ class TestVersionViewSetCreateJWTAuth(TestVersionViewSetCreate):
...
+            response = self.client.patch(
+                self.url,
+                data={'upload': upload.uuid},
+            )
+            parse_addon_mock.assert_not_called()
+
+        assert response.status_code == 200, response.content
+        self.addon.reload()
+        self.version.reload()
+        assert self.version.version == '123.b4'
```
bobsilverberg, line 1547: I take it from this that if the upload was successful the version number would be something other than `123.b4`?

Issue #18104, opened 2021-10-13
Clean up usage of STATIC_URL


diox:
After #18010 is fixed:
- Hunt down any reamining usage of `STATIC_URL` - we should be using `static()` instead most of the time
- Remove `static_url` context processor

diox:
QA: URL for the default icon in all sizes should still point to a valid image and have a unique filename set (`https://addons-dev.allizom.org/static/addon-icons/default-32.1366fe4b2b33.png` for instance) but not query string parameters.

ioanarusiczki:
@diox 
I see the following formats:
https://addons-dev.allizom.org/static-server/img/addon-icons/default-64.d144b50f2bb8.png
https://addons-dev.allizom.org/static-server/img/addon-icons/default-32.2b26add4b394.png 

This view - in edit profile page image section has `?1638195039982` at the end of the URLs
such as 
https://addons-dev.allizom.org/static-server/img/addon-icons/default-64.d144b50f2bb8.png?1638195039982 

![with query](https://user-images.githubusercontent.com/33448286/143882962-983c764c-c034-40e6-a77c-c6b6bc06a6fb.png)

Are these alright?

diox:
Ah yes, I forgot Devhub was doing this. As long as the base URL are `.../addon-icons/default-64.d144b50f2bb8.png` and `.../addon-icons/default-32.2b26add4b394.png` and the added query string doesn't create errors that's fine.

ioanarusiczki:
@diox I filed https://github.com/mozilla/addons-server/issues/18438 but it is not related to the changes made by this patch.

PR #18154, opened 2021-10-20
Fix static URL for no-preview thumbnail URL

diox:
Fixes mozilla/addons#8552

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -9,6 +9,7 @@ from datetime import datetime
 from urllib.parse import urlsplit
 
 from django.conf import settings
+from django.contrib.staticfiles.storage import staticfiles_storage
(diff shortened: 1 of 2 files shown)


src/olympia/addons/models.py
```diff
@@ -1390,7 +1391,7 @@ def thumbnail_url(self):
             preview = self._all_previews[0]
             return preview.thumbnail_url
         except IndexError:
-            return settings.STATIC_URL + '/img/icons/no-preview.png'
+            return staticfiles_storage.url('img/icons/no-preview.png')
```
eviljeff, line 1394: should probably use this function in `get_default_icon_url` too

diox: Yeah I'll do that in https://github.com/mozilla/addons-server/issues/18104

PR #18206, opened 2021-10-27
Add a ReviewActionReason model

bobsilverberg:
Fixes mozilla/addons#8554 

I've chosen to use the name `ReviewActionReason`, but obviously that is up for debate. There's no urgency to land this, but I wanted to get started on the feature. Requesting review from Andrew as he's the only other one here!

PR #18214, opened 2021-10-28
Add a ReviewActionReasonLog model

### LINKED ISSUE #18191: Allow for categories/reasons to be assigned to ActivityLog entries
This is a spin-off from https://github.com/mozilla/addons-server/issues/18168, and depends on https://github.com/mozilla/addons-server/issues/18190. 

After https://github.com/mozilla/addons-server/issues/18190 lands, we will have a category/reason entity which could be connected to an `ActivityLog` entry. This issue is about creating that relationship.

@eviljeff suggests, at https://github.com/mozilla/addons-server/issues/18168#issuecomment-952017150, that we could have, for example, a `ReviewCategoryLog` with fks to activity_log, version and review_category. We should therefore be able to create this as an empty table, with all three fks required and then create the routines to populate it in a follow-up issue.

I gather from the above that `ActivityLog` does not include `version`, or if that's not the case then I'm unclear on why we need an fk for version in the relationship table.

bobsilverberg:
~~Depends on https://github.com/mozilla/addons-server/pull/18206~~
Fixes #18191


src/olympia/activity/models.py
```diff
@@ -150,6 +151,18 @@ class Meta:
...
+    """
+    This table allows ReviewActionReasons to be assigned to ActivityLog entries.
+    """
+
+    activity_log = models.ForeignKey('ActivityLog', on_delete=models.CASCADE)
+    reason = models.ForeignKey(ReviewActionReason, on_delete=models.CASCADE)
+
+    class Meta:
+        ordering = ('-created',)
+
```
eviljeff, line 164: Activity log models are one of the few places we still specify `db_table`, to have them consistently all named `log_activity_xxx`

eviljeff (approved):
looks good - is there anything that's going to be consuming/displaying these events though?  Or is it just something that'll be queried via redash?

bobsilverberg:
> looks good - is there anything that's going to be consuming/displaying these events though? Or is it just something that'll be queried via redash?

AFAIU, for now it's just about being able to query the data.

PR #18266, opened 2021-11-04
Link ReviewActionReasonLog to ActivityLog

### LINKED ISSUE #18265: Link ReviewActionReasonLog to ActivityLog
This is a follow-up to https://github.com/mozilla/addons-server/issues/18191.

The original implementation for `ReviewActionReasonLog` joined to `VersionLog`, but really it should have joined to `ActivityLog`. This issue is about making that change.

bobsilverberg:
Fixes #18265


eviljeff (approved):
Like Mat was saying on Slack, we'd need to write a manual migration to backfill `activity_log` with the contents of `version_log.activity_log` if we have any log entries on prod, but as it's not in use yet it's fine.

PR #18278, opened 2021-11-05
Add basic admin for install origins

### LINKED ISSUE #18242: Add basic admin for InstallOrigin
We want a basic admin set up for QA/admins to look at `InstallOrigin` records.

diox:
Fixes #18242

### DIFF
--- a/src/olympia/versions/admin.py
+++ b/src/olympia/versions/admin.py
@@ -1,6 +1,6 @@
 from django.contrib import admin
 
-from .models import License, Version, VersionReviewerFlags
+from .models import InstallOrigin, License, Version, VersionReviewerFlags
 
 
 class VersionReviewerFlagsInline(admin.StackedInline):
@@ -24,6 +24,8 @@ class VersionAdmin(admin.ModelAdmin):
 
     view_on_site = False
     readonly_fields = ('id', 'created', 'version', 'channel')
+    list_display = ('id', 'addon_guid', 'version', 'deleted', 'channel')
+    list_select_related = ('addon',)
 
     raw_id_fields = ('addon', 'license')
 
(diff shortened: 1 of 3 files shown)


src/olympia/versions/admin.py
```diff
@@ -48,6 +50,27 @@ class Media:
...
+    def addon_guid(self, obj):
+        return obj.addon.guid
 
+    addon_guid.short_description = 'Add-on GUID'
+
+
+class InstallOriginAdmin(admin.ModelAdmin):
+    view_on_site = False
+    raw_id_fields = ('version',)
+    readonly_fields = ('id', 'version', 'origin', 'base_domain')
```
eviljeff, line 62: with `has_change_permission` always `False` this is redundant?

src/olympia/versions/admin.py
```diff
@@ -48,6 +50,27 @@ class Media:
     )
     inlines = (VersionReviewerFlagsInline,)
 
+    def addon_guid(self, obj):
+        return obj.addon.guid
 
+    addon_guid.short_description = 'Add-on GUID'
+
+
+class InstallOriginAdmin(admin.ModelAdmin):
```
eviljeff, line 59: what's the use case for this admin?  Admins can't add, delete, or change the records, and the table will be a long list of ids next to origins and domains (without filtering), which appears of limited use.

eviljeff: is it only for QA?

src/olympia/versions/admin.py
```diff
@@ -24,6 +24,8 @@ class Media:
 
     view_on_site = False
     readonly_fields = ('id', 'created', 'version', 'channel')
+    list_display = ('id', 'addon_guid', 'version', 'deleted', 'channel')
+    list_select_related = ('addon',)
```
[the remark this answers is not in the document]
diox: I'm only displaying the add-on guid and not the add-on str representation/name so it shouldn't cause extra queries, and that's also why I didn't bother going with the `prefetch_related` route, since a join is enough.

I verified manually that it doesn't cause extra queries but I should add a test.

src/olympia/versions/admin.py
diox: Mostly for QA/debugging at the moment, yeah. There hasn't been anything specced for admin needs yet.

eviljeff (approved):

src/olympia/versions/admin.py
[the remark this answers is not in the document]
diox: Ugh I totally messed up, mixed together `VersionAdmin` (which I changed for convenience when testing) and `InstallOriginAdmin`. Will fix and add tests.

PR #18244, opened 2021-11-02
Add a select field for ReviewActionReasons

bobsilverberg:
Fixes #18233 

Submitting this for feedback to make sure I am on the right track. There are still a couple of questions to be answered about this.

Here's a screenshot of what the new field looks like in the form:
Screen Shot 2021-11-02 at 2 48 58 PM

Issue #18233, opened 2021-11-01
Add a field for selecting review action reasons


bobsilverberg:
This is a spin-off from https://github.com/mozilla/addons-server/issues/18168.

When a reviewer is performing certain actions (e.g., rejecting a version, rejecting multiple versions, replying to a review), they should be able to select one or more reasons for the action. This will be enabled via a simple multi-select for the MVP.

This behaviour will generally be present on screens which currently allow for selection of a canned response, so it might make sense to add the multi-select for the action reason to the same template as the canned response select, which currently exists at https://github.com/mozilla/addons-server/blob/master/src/olympia/reviewers/templates/reviewers/review.html#L119.

A couple of questions (for @kewisch / @diox):
1. Can we confirm precisely on which screens this select should be available?
2. Should there be a maximum number of reasons that can be selected? Would at least one reason be required to perform an action, or is it considered optional?
3. Does this need to be behind a flag while we are developing it, or does the fact that it is only visible to reviewers mean that we can implement it in pieces and just tell them it's not ready yet?

bobsilverberg:
One more question for @kewisch, do you envision this as a multi-select input, or as multiple check boxes?

tublitzed:
@bobsilverberg:

2. No maximum. This should be a required field.

re:multi-select or multiple check boxes - whatever is simplest on your end, given that the end result is the same.

bobsilverberg:
> @bobsilverberg:
> 
>     1. No maximum. This should be a required field.
> 
> 
> re:multi-select or multiple check boxes - whatever is simplest on your end, given that the end result is the same.

I think they are pretty much equivalent in terms of simplicity, so it comes down to what's the best UX for the reviewers. Using check boxes would mean all options would be visible at all times, but if there are a lot of options it might be ugly. Using a multi-select I would imagine that one would have to scroll through the list to see all of the options, so that's a tradeoff with that.

Perhaps @kewisch has a particular preference, or maybe @wagnerand has an opinion?

kewisch:
> This is a spin-off from #18168.
> 
> When a reviewer is performing certain actions (e.g., rejecting a version, rejecting multiple versions, replying to a review), they should be able to select one or more reasons for the action. This will be enabled via a simple multi-select for the MVP.
> 
[...]

PR #18289, opened 2021-11-08
Only display reasons select when needed

### LINKED ISSUE #18288: Only display the reason select where is it needed
This is a follow-up from https://github.com/mozilla/addons-server/issues/18233.

A reason is only required (and relevant) for the following actions:
* `public`
* `reject`
* `reject_multiple_versions`
* `reply`

However, the comment field is also displayed for:
* `super`
* `comment`

The implementation from #18233 displays the reason select on all forms which include a comment field, but we need to update it so it only displays for the actions in the top list above.

bobsilverberg:
~~Depends on #18244~~
Fixes #18288

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/review.html
+++ b/src/olympia/reviewers/templates/reviewers/review.html
@@ -117,7 +117,8 @@
             {{ form.comments }}
             {{ form.comments.errors }}
           </div>
-          <div class="review-actions-reasons">
+          <div class="data-toggle review-actions-reasons"
+            data-value="{{ actions_reasons|join('|') }}|">
             <label for="id_reasons">{{ form.reasons.label }}</label>
             <div class="review-actions-reasons-select">
               {{ form.reasons }}
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -5602,6 +5602,12 @@ class TestReview(ReviewBase):
             doc('.data-toggle.review-comments')[0].attrib['data-value']
             == 'reject_multiple_versions|reply|super|comment|'
         )
+
+        assert (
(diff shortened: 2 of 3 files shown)


bobsilverberg:
This PR build on top of https://github.com/mozilla/addons-server/pull/18244, so it probably makes sense to hold off reviewing it until that is merged and I rebase it. All I've added here is the code to hide the `reasons` select for actions for which it is not applicable. It's just a few lines of code, but I'm not sure how to test it. I tried to find tests for similar code, but wasn't successful. I tried to see if I could do something with the `content = str(form); doc = pq(content)` pattern, but it seemed that the element to which I added the `data-value` attribute doesn't appear in that content.

Any suggestions for what I can to do test this new code would be appreciated.

bobsilverberg:
This is ready for review, and should ideally land this week as it complements #18233. Please see the comment above about being unsure how to add a test for this.

diox:
You can't use `str(form)` because the template is separate, and completely unrelated to the `Form` instance (sadly: it's something we could/should fix, but it's a lot of work).

We do have a few tests in `src/olympia/reviewers/tests/test_views.py` that test similar things by rendering the full review page, and looking for the right element in there.

bobsilverberg:
> You can't use `str(form)` because the template is separate, and completely unrelated to the `Form` instance (sadly: it's something we could/should fix, but it's a lot of work).
> 
> We do have a few tests in `src/olympia/reviewers/tests/test_views.py` that test similar things by rendering the full review page, and looking for the right element in there.

Thanks @diox. I actually found a place where similar things are already being tested, and it makes sense to just update those tests. I'm not sure why I didn't find them the first time around.

diox (approved):

bobsilverberg:
Thanks @diox. I think I'm going to cherry-pick this so that the form doesn't appear for actions where it is not valid.

PR #18290, opened 2021-11-08
Assign reasons to actions when the ActivityLog entry is written

### LINKED ISSUE #18234: Allow for recording ReviewActionReasons when logging activity
This is a spin-off from https://github.com/mozilla/addons-server/issues/18168.

In order to record the reason(s) for a review action, the routines that create `ActivityLog` entries when a user performs an action on a review will need to be able to accept one or more `ReviewActionReason`s, and then record those reasons in the database.

It looks like the places where this occurs are limited to:
- [`log_action`](https://github.com/mozilla/addons-server/blob/master/src/olympia/reviewers/utils.py#L831) in `reviewers/utils.py`
- [`log_and_notify`](https://github.com/mozilla/addons-server/blob/master/src/olympia/activity/utils.py#L223) in `activity/utils.py`

There are also a few places where we explicitly create an instance of `ActivityLog`, but they don't seem to be cases where we'd need to assign a reason, so we can probably ignore them. I am curious about why we are doing that and not just calling `log_action` though, but that's just me being curious.
- [`sign_file`](https://github.com/mozilla/addons-server/blob/b0db9bb7ea3093dc6bfe1625685bcb226603e703/src/olympia/reviewers/utils.py#L943) in `reviewers/utils.py`
- [`approve_latest_version`](https://github.com/mozilla/addons-server/blob/b0db9bb7ea3093dc6bfe1625685bcb226603e703/src/olympia/reviewers/utils.py#L1337) in `reviewers/utils.py`

This issue is about updating `log_action` and `log_and_notify` to accept one or more reasons and to record them in the database when recording the `ActivityLog`. As it only appears that `log_and_notify` is used in one place, it might be worth investigating whether we can discontinue its use in `reviewers/utils.py` and instead use `log_action`.

bobsilverberg:
Depends on https://github.com/mozilla/addons-server/pull/18244
Fixes #18234


src/olympia/reviewers/utils.py
```diff
@@ -851,6 +851,10 @@ def log_action(self, action, version=None, file=None, timestamp=None):
         if timestamp is None:
             timestamp = datetime.now()
 
+        reasons = self.data.get('reasons', [])
+        for reason in reasons:
+            args = args + (reason,)
+
```
eviljeff, line 857:
```suggestion
        args = (*args, *self.data.get('reasons', []))
```

src/olympia/reviewers/utils.py
```diff
@@ -919,6 +923,7 @@ def get_context_data(self):
             ),
             'comments': self.data.get('comments'),
             'SITE_URL': settings.SITE_URL,
+            'reasons': self.data.get('reasons'),
```
eviljeff, line 926: `get_context_data` is used for the email (I don't see any changes to the email template)

eviljeff (approved):

src/olympia/activity/models.py
```diff
@@ -754,6 +754,12 @@ def create(cls, action, *args, **kw):
                     guid=arg.guid,
                     created=kw.get('created', timezone.now()),
                 )
+            elif class_ == ReviewActionReason:
+                ReviewActionReasonLog.objects.create(
+                    reason_id=id_,
+                    activity_log=al,
+                    created=kw.get('created', timezone.now()),
+                )
```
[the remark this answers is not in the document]
bobsilverberg: Sure, I'll do that.

src/olympia/reviewers/utils.py
bobsilverberg: I thought it was used for the form, so when validations fail the form can be re-displayed with the data. Is that not the case?

bobsilverberg: A quick test shows me that I was wrong. :-)

I'll remove this.

eviljeff: not as far as I can see - it's the context for the email template - but reviewer tools is a sprawling chunk of code so I wouldn't be too surprised if I was wrong.

PR #18295, opened 2021-11-09
Make our render() util use TemplateResponse, keeping rendering lazy

### LINKED ISSUE #18263: Modernize/replace amo.utils.render
We have our own render function that is used by several views in addons-server:
https://github.com/mozilla/addons-server/blob/7a68c0fedf98420088fb4d60dfab151fda92ed0a/src/olympia/amo/utils.py#L73-L75

It's obsolete, because 1) https://docs.djangoproject.com/en/3.2/topics/http/shortcuts/#render exists and 2) `TemplateResponse()` is a lot better (keeps template rendering lazy, allows for debugging of context data...)

Something like the following patch should be a good replacement:
```diff
diff --git a/src/olympia/amo/utils.py b/src/olympia/amo/utils.py
index d1b5741061..d0c91d9e31 100644
--- a/src/olympia/amo/utils.py
+++ b/src/olympia/amo/utils.py
@@ -34,6 +34,7 @@ from django.forms.fields import Field
 from django.http import HttpResponse
 from django.http.response import HttpResponseRedirectBase
 from django.template import engines, loader
+from django.template.response import TemplateResponse
 from django.urls import reverse
 from django.utils import translation
 from django.utils.encoding import force_bytes, force_str
@@ -71,8 +72,9 @@ log = getLogger('z.amo')
 
 
 def render(request, template, ctx=None, status=None, content_type=None):
-    rendered = loader.render_to_string(template, ctx, request=request)
-    return HttpResponse(rendered, status=status, content_type=content_type)
+    return TemplateResponse(
+        request, template, context=ctx, status=status, content_type=content_type
+    )
 
 
 def from_string(string):
```

Need to test that out.

### LINKED ISSUE #18296: Read Only mode 500s when processing API requests resulting in a OperationalError
Shortly after read-only mode was introduced in 9dfcf92c096e8e502413c468a39a912169f54a91, it was simplified to remove the `Retry-After` header, in 18c1cc943809a83f77ba83c168dfb983454c28cf

Unfortunately, that created a bug in a small, super rare condition that wasn't covered by tests:
https://github.com/mozilla/addons-server/blob/6b0ee7ca941430ba9dcc1a4f6067f62042493a23/src/olympia/amo/middleware.py#L244-L251

When:
- We're in read-only mode
- We're processing an exception that comes from a previous middleware in the stack (can't come from a view, because that would be caught in `process_request()` before)
- That exception is an `OperationalError` from MySQL
- We're dealing with an API request

We call `self._render_api_error()`, which no longer exists, resulting in a 500 instead of the intended 503.

In addition the middleware docstrings mentions we `Supports issuing `Retry-After` header` which isn't true anymore.

diox:
Fixes #18263
Fixes #18296


src/olympia/amo/middleware.py
```diff
@@ -239,7 +239,9 @@ def process_request(self, request):
             if writable_method:
                 return JsonResponse({'error': self.ERROR_MSG}, status=503)
         elif request.method == 'POST':
-            return render(request, 'amo/read-only.html', status=503)
+            response = render(request, 'amo/read-only.html', status=503)
+            response.render()  # Force render to return an HttpResponse
+            return response
```
eviljeff, line 244: this reads weird - what's going on with the read-only case that we have to force `.render()` on the response before returning?

diox: Middlewares that return a response instead of letting the request be processed by the view need to directly return an `HttpResponse`. If you return a `TemplateResponse` without rendering it, it will never be rendered, because that normally happens in view handling code that is being bypassed.

eviljeff: Where is the view handling code that is being bypassed that would ordinarily do `render()`?  django code?

(I sometimes think we should rewrite all our old middlewares to use modern style callables rather than using `MiddlewareMixin`)

diox: It's in django yes. 

This is alluded to in the docs, it might be surprising at first glance but makes sense given what happens behind the scenes:
> (...) the `get_response` callable for the last middleware in the chain won’t be the actual view but rather a wrapper method from the handler which takes care of applying view middleware, calling the view with appropriate URL arguments, and applying template-response and exception middleware.
> (...)
> If one of the layers decides to short-circuit and return a response without ever calling its get_response, none of the layers of the onion inside that layer (including the view) will see the request or the response. The response will only return through the same layers that the request passed in through.

Returning a response at this step is allowed, but [it bypasses `get_response()` entirely](https://github.com/django/django/blob/99532fdadf0b2bd065b1f50b4ec14e1b0e53f8a3/django/utils/deprecation.py#L117), and that function is [what does the rendering](https://github.com/django/django/blob/99532fdadf0b2bd065b1f50b4ec14e1b0e53f8a3/django/core/handlers/base.py#L204).

If we don't call `render()` ourselves, all middlewares that we went through that need to do something with the response content (`CommonMiddleware`, `GZipMiddleware`) will fail as soon as they try to access it as rendering should have taken place by now, but didn't because we bypassed it. If we remove them, it will fail at the very end, when the test is accessing the response content, for the same reason.

diox: I'm open to suggestions to reword that comment since evidently it's not very clear as it stands now :)

eviljeff: it's unfortunate the function is named `render` when it doesn't actually render... but something like "render the lazy response into a HttpResponse"?  I'm assuming that the response **always** has a `render` method too?

src/olympia/amo/middleware.py
```diff
@@ -248,7 +250,9 @@ def process_exception(self, request, exception):
         if isinstance(exception, mysql.OperationalError):
             if request.is_api:
                 return self._render_api_error()
-            return render(request, 'amo/read-only.html', status=503)
+            response = render(request, 'amo/read-only.html', status=503)
+            response.render()  # Force render to return an HttpResponse
+            return response
```
eviljeff, line 255: if you were in the mood for refactoring, these lines are a bit DRYy with `process_request`

eviljeff (approved):

src/olympia/amo/middleware.py
diox: That's a good point. Maybe I could get rid of that `render()` function altogether, and use `TemplateResponse` directly everywhere ? We'd still need to call `response.render()` here, but it would be less confusing.

diox: https://github.com/mozilla/addons-server/pull/18295/commits/20b612358178e8fb6d5a55212327c8c69f06e52b

src/olympia/amo/middleware.py
diox: https://github.com/mozilla/addons-server/pull/18295/commits/20b612358178e8fb6d5a55212327c8c69f06e52b

PR #18298, opened 2021-11-09
makes writeable: is_experimental, requires_payment, is_disabled

### LINKED ISSUE #18269: allow basic properties on addon to be set via the api
part of #18036 

These properties are all just boolean properties so it should be easy enough to implement write on them at the same time.
- 'is_disabled',
- 'is_experimental'
- 'requires_payment',

eviljeff:
fixes #18269

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -806,6 +806,8 @@ class TestAddonViewSetCreate(UploadMixin, TestCase):
             'description': {'en-US': 'new description'},
             'developer_comments': {'en-US': 'comments'},
             'homepage': {'en-US': 'https://my.home.page/'},
+            'is_experimental': True,
+            'requires_payment': True,
             # 'name'  # don't update - should retain name from the manifest
             'slug': 'addon-slug',
             'summary': {'en-US': 'new summary'},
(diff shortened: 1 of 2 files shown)


src/olympia/addons/tests/test_views.py
```diff
@@ -1024,6 +1052,32 @@ def test_set_extra_data(self):
...
+        data = response.data
+        assert data['is_disabled'] is True
+        assert addon.is_disabled is True
+        assert addon.disabled_by_user is True  # sets the user property
+
+        # Confirm that a STATUS_DISABLED can't be overriden
+        addon.update(status=amo.STATUS_DISABLED)
+        response = self.client.patch(
+            self.url,
+            data={'is_disabled': False},
```
bobsilverberg, line 1072: Should this actually raise an exception? It seems like an invalid API call.

bobsilverberg (approved):
Looks good. Just one question about trying to use the API to enable an already disabled add-on.

r+wc

src/olympia/addons/tests/test_views.py
bobsilverberg: I was thinking because they are trying to un-disable a disabled add-on and we don't allow that. So it seems like maybe we should return an error rather then just not allowing it to happen?

eviljeff: There's a wider lack of enforcement of the restriction around disabled add-ons, so I filed #18330 to cover everything.  (The test is really more about checking a write to that property is writing to the correct model field; once #18330 is done we'll likely have to drop this test because it won't be possible anyway)

PR #18299, opened 2021-11-09
validate slug in addon api

### LINKED ISSUE #18201: slug accepts special chars when sending patch requests
Followup for 

https://github.com/mozilla/addons-server/issues/18037#issuecomment-948586765

Send a request https://addons-dev.allizom.org/api/v5/addons/addon/620866/ with {"slug": "!@!#!@##@$$%$#%#%$^^%&%"} 

Result:
The response is 200. On dev hub a slug such as '75b25b33bf9c4cf9bcda' is generated.  

Another example: {"slug": "~~~~~~~~"}  -> this slug is accepted. The frontend page :https://addons-dev.allizom.org/en-US/firefox/addon/~~~~~~~~~~~~~~~~~~~~/

Expected result:
A 400 response with a warning message.

eviljeff:
fixes #18201

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -853,6 +853,7 @@ class TestAddonViewSetCreate(UploadMixin, TestCase):
         assert response.status_code == 201, response.content
 
     def test_set_slug(self):
+        # Check for slugs with invalid characters in it
         response = self.client.post(
             self.url,
             data={
@@ -868,6 +869,7 @@ class TestAddonViewSetCreate(UploadMixin, TestCase):
             ]
         }
 
+        # Check for a slug in the DeniedSlug list
         DeniedSlug.objects.create(name='denied-slug')
         response = self.client.post(
             self.url,
@@ -881,6 +883,7 @@ class TestAddonViewSetCreate(UploadMixin, TestCase):
             'slug': ['This slug cannot be used. Please choose another.']
         }
 
+        # Check for all numeric slugs - DeniedSlug.blocked checks for these too.
         response = self.client.post(
             self.url,
             data={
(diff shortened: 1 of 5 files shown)


src/olympia/amo/utils.py
```diff
@@ -490,7 +490,7 @@ def normalize_string(value, strip_punctuation=False):
 def slug_validator(
     s,
     ok=SLUG_OK,
-    lower=True,
+    lower=False,
```
diox, line 493: That function follows `slugify()` closely, only adding arguments that matter for validator purposes. `slugify()` defaults to `lower=True` so I don't think we should change that.

eviljeff: I did it because we always passed `lower=False` everywhere we used `slug_validator` but if we want those redundant kwargs I can change it back?

eviljeff: I was tempted to drop all the kwargs apart from `message` so we're directly relying on slugify defaults.

diox: If you mean drop all kwargs except `message` and always pass `lower=False` to `slugify` then yes, that sounds ok.

src/olympia/addons/tests/test_views.py
```diff
@@ -852,6 +852,47 @@ def test_too_many_categories(self):
...
+        assert response.status_code == 400, response.content
+        assert response.data == {
+            'slug': ['This slug cannot be used. Please choose another.']
+        }
+
+        response = self.client.post(
+            self.url,
+            data={
+                **self.minimal_data,
+                'slug': '1234',
```
diox, line 888: I forgot that `DeniedSlug` was the bit of code responsible for denying pure digits slugs and was initially struggling to figure out what caused that error message to be returned! Could you add a small comment ?

src/olympia/addons/tests/test_views.py
```diff
@@ -852,6 +852,47 @@ def test_too_many_categories(self):
...
+            },
+        )
+        assert response.status_code == 400, response.content
+        assert response.data == {
+            'slug': ['This slug cannot be used. Please choose another.']
+        }
+
```
diox, line 895: Could you add a test for the `addon with this slug already exists` case ? (I assume this is already handled by DRF because it's unique, but it's good to be sure)

diox (approved):

src/olympia/addons/tests/test_views.py
eviljeff: https://github.com/mozilla/addons-server/pull/18299/commits/f8f35438a6ffef4e8e5184f8ee67d4d44294f064

src/olympia/addons/tests/test_views.py
eviljeff: https://github.com/mozilla/addons-server/pull/18299/commits/f8f35438a6ffef4e8e5184f8ee67d4d44294f064

PR #18302, opened 2021-11-09
Add an admin for ReviewActionReasonLog

### LINKED ISSUE #18301: Add an admin for ReviewActionReasonLog
A feature request is to be able to change the reason assigned to a review action if the original reason selected was incorrect. Possibly the simplest way to address this is to add an admin for `ReviewActionReasonLog` which will allow a user to edit the `reason`. I did a quick test of this, and it seems to be doable, although we'd need a way for users to be able to search through the log entries to find the one they want.

bobsilverberg:
Fixes #18301 

This is my first stab at this. I think it will be fairly helpful in terms of being able to find the log that needs to be fixed. There is a filter for `reason`, and I added a search that can be used for the reviewer's email address. It's not perfect, but it might do for now, or maybe someone has some suggestions for improvements.

~~One problem that does exist, that I was unable to figure out how to deal with, is the fact that the select that can be used to change the `reason` includes all reasons, not just active reasons, and that is less than ideal. I imagine there is a way to define a queryset to be used to populate this select, but I couldn't find anything in the basic Django admin docs. Suggestions appreciated!~~
- I figured it out! (It *was* in the docs)

### DIFF
--- a/src/olympia/activity/admin.py
+++ b/src/olympia/activity/admin.py
@@ -65,18 +65,15 @@ class ReviewActionReasonLogAdmin(admin.ModelAdmin):
     def has_delete_permission(self, request, obj=None):
         return False
 
-    def formfield_for_foreignkey(self, db_field, request, **kwargs):
-        if db_field.name == 'reason':
-            kwargs['queryset'] = ReviewActionReason.objects.filter(
-                is_active__exact=True
-            )
-        return super().formfield_for_foreignkey(db_field, request, **kwargs)
-
     def get_form(self, request, obj=None, **kwargs):
         form = super(ReviewActionReasonLogAdmin, self).get_form(request, obj, **kwargs)
         form.base_fields['reason'].widget.can_add_related = False
         form.base_fields['reason'].widget.can_change_related = False
         form.base_fields['reason'].empty_label = None
+        form.base_fields['reason'].choices = [
+            (reason.id, reason.labelled_name())
+            for reason in ReviewActionReason.objects.all()
+        ]
         return form
 
 
--- a/src/olympia/activity/admin.py
+++ b/src/olympia/activity/admin.py
@@ -37,26 +37,26 @@ class ReviewActionReasonLogAdmin(admin.ModelAdmin):
     fields = (
         'created',
         'activity_log',
-        'user_email',
+        'activity_log__user__email',
         'reason',
     )
     list_display = (
         'created',
         'activity_log',
         'reason',
-        'user_email',
+        'activity_log__user__email',
     )
     list_filter = ('reason',)
+    list_select_related = ('activity_log__user',)
     readonly_fields = (
         'created',
         'activity_log',
-        'user_email',
+        'activity_log__user__email',
     )
     search_fields = ('activity_log__user__email',)
     view_on_site = False
 
-    @admin.display()
-    def user_email(self, obj):
+    def activity_log__user__email(self, obj):
         return obj.activity_log.user.email
 
     def has_add_permission(self, request):
@@ -76,6 +76,7 @@ class ReviewActionReasonLogAdmin(admin.ModelAdmin):
         form = super(ReviewActionReasonLogAdmin, self).get_form(request, obj, **kwargs)
         form.base_fields['reason'].widget.can_add_related = False
         form.base_fields['reason'].widget.can_change_related = False
+        form.base_fields['reason'].empty_label = None
         return form
 
 
--- a/src/olympia/activity/admin.py
+++ b/src/olympia/activity/admin.py
@@ -67,7 +67,7 @@ class ReviewActionReasonLogAdmin(admin.ModelAdmin):
 
     def formfield_for_foreignkey(self, db_field, request, **kwargs):
         if db_field.name == 'reason':
-            kwargs["queryset"] = ReviewActionReason.objects.filter(
+            kwargs['queryset'] = ReviewActionReason.objects.filter(
                 is_active__exact=True
             )
         return super().formfield_for_foreignkey(db_field, request, **kwargs)
--- a/src/olympia/activity/admin.py
+++ b/src/olympia/activity/admin.py
@@ -1,7 +1,7 @@
 from django.contrib import admin
 
 from .models import ActivityLog, ReviewActionReasonLog
-from olympia.reviewers.models import CannedResponse, ReviewActionReason
+from olympia.reviewers.models import ReviewActionReason
 
 
 class ActivityLogAdmin(admin.ModelAdmin):
@@ -71,6 +71,12 @@ class ReviewActionReasonLogAdmin(admin.ModelAdmin):
                 is_active__exact=True
             )
         return super().formfield_for_foreignkey(db_field, request, **kwargs)
+
+    def get_form(self, request, obj=None, **kwargs):
+        form = super(ReviewActionReasonLogAdmin, self).get_form(request, obj, **kwargs)
+        form.base_fields['reason'].widget.can_add_related = False
+        form.base_fields['reason'].widget.can_change_related = False
+        return form
 
 
 admin.site.register(ActivityLog, ActivityLogAdmin)
--- a/src/olympia/activity/admin.py
+++ b/src/olympia/activity/admin.py
@@ -1,6 +1,7 @@
 from django.contrib import admin
 
 from .models import ActivityLog, ReviewActionReasonLog
+from olympia.reviewers.models import CannedResponse, ReviewActionReason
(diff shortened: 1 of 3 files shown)


src/olympia/activity/admin.py
```diff
@@ -31,4 +32,52 @@ def has_delete_permission(self, request, obj=None):
         return False
 
 
+class ReviewActionReasonLogAdmin(admin.ModelAdmin):
+    date_hierarchy = 'created'
+    fields = (
+        'created',
+        'activity_log',
+        'user_email',
```
eviljeff, line 40: you can specify the field directly here like you have in `search_fields` without adding a ` def user_email` method.

src/olympia/activity/admin.py
```diff
@@ -31,4 +32,52 @@ def has_delete_permission(self, request, obj=None):
...
+        'reason',
+    )
+    list_display = (
+        'created',
+        'activity_log',
+        'reason',
+        'user_email',
+    )
```
eviljeff, line 48: define `list_select_related` to efficiently preload all the `UserProfile` instances for the emails (or you end up with an query per row in the results table)

src/olympia/activity/admin.py
```diff
@@ -31,4 +32,52 @@ def has_delete_permission(self, request, obj=None):
...
+    def has_add_permission(self, request):
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        return False
+
+    def formfield_for_foreignkey(self, db_field, request, **kwargs):
+        if db_field.name == 'reason':
+            kwargs['queryset'] = ReviewActionReason.objects.filter(
+                is_active__exact=True
```
eviljeff, line 71: I don't think you need the `__exact`?

src/olympia/activity/admin.py
```diff
@@ -31,4 +32,52 @@ def has_delete_permission(self, request, obj=None):
...
+            )
+        return super().formfield_for_foreignkey(db_field, request, **kwargs)
+
+    def get_form(self, request, obj=None, **kwargs):
+        form = super(ReviewActionReasonLogAdmin, self).get_form(request, obj, **kwargs)
+        form.base_fields['reason'].widget.can_add_related = False
+        form.base_fields['reason'].widget.can_change_related = False
```
eviljeff, line 78: what's this for?

src/olympia/activity/admin.py
```diff
@@ -31,4 +32,52 @@ def has_delete_permission(self, request, obj=None):
...
+        return form
+
```
eviljeff, line 80: can we have some tests?  (it's not a standard ModelAdmin - there's specific code for the form field)

eviljeff: tests at least for:
- the correct reasons been available for selection (excludes inactive reasons)
- that viewing an instance that used a now inactive reason still shows the original reason

src/olympia/activity/admin.py
```diff
@@ -31,4 +32,52 @@ def has_delete_permission(self, request, obj=None):
...
+        'user_email',
+    )
+    search_fields = ('activity_log__user__email',)
+    view_on_site = False
+
+    @admin.display()
+    def user_email(self, obj):
+        return obj.activity_log.user.email
+
+    def has_add_permission(self, request):
```
eviljeff, line 62: Might be worth checking what users will be expected to have access to this tool.  By default only superpowered admins can even view it (`*:*`) - if any other admin classes will need access to view/change then permissions need defining in olympia.constants/permissions.py

eviljeff (commented):
tests please 😉

src/olympia/activity/admin.py
bobsilverberg: I tried that first, but I get an error: `Unknown field(s) (activity_log__user__email) specified for ReviewActionReasonLog. Check fields/fieldsets/exclude attributes of class ReviewActionReasonLogAdmin.` I don't really understand why it works for `search_fields` but not for `fields` or `list_display`.

eviljeff: mmm, I forget what's allowed.  Okay, well elsewhere we've used the full name and then defined a method with that name.  e.g. `activity_log__user__email` in `fields` and `def activity_log__user__email` in the model class.

src/olympia/activity/admin.py
bobsilverberg: I added `list_select_related = ('activity_log__user',)`. Is that what you mean here?

src/olympia/activity/admin.py
bobsilverberg: Without them you get a little pencil icon and a plus-sign icon beside the select for `reason`, which allow you to edit the `reason` and add a new `reason`, and I don't think those functions are required from this screen, and might in fact be confusing to users. This is what it looks without these:
Screen Shot 2021-11-10 at 12 00 08 PM

src/olympia/activity/admin.py
eviljeff: yes

src/olympia/activity/admin.py
eviljeff: Ah. I think there's a different way to do this (we don't set those properties elsewhere) but if this works then 👍

src/olympia/activity/admin.py
bobsilverberg: I was able to use the same name in `fields` and `search_fields`, by renaming my method in `admin.py` to `activity_log__user__email`, but I cannot seem to get it to work by adding a method to the model, which seems to be what you are suggesting above. I can't seem to find an example like this in any of our model files, but I do see a few examples of it being done the way I did it here, in the `admin.py` file. When I push a new commit you can take a look and let me know what you think.

eviljeff: whoops, I meant `ModelAdmin` rather than `Modal`

src/olympia/activity/admin.py
bobsilverberg: This question has spurred some discussion about the validity of using only the admin for this feature, so for the purposes of this patch, I am going to keep it as is, and only allow super users to access the feature. We can have a follow-up if needed.

src/olympia/activity/admin.py
bobsilverberg: I added a test for "the correct reasons been available for selection (excludes inactive reasons)", but in trying to add a test for the second use case, I found that it does not in fact show the original reason. I have code in place limiting the source queryset for the select to only those reasons that are active, which explains why the inactive reason is excluded, but I'm not sure how to go about making it should the current reason if it is excluded. I thought maybe I could add the current reason to the queryset using `union`, but it doesn't seem like I have access to the current instance of the model object from `formfield_for_foreignkey`. Do you have any suggestions for how I might get this to work?

bobsilverberg:
Thanks for the review, @eviljeff. I have addressed all of your comments, and I added a test which is now failing because the admin form does not in fact include the current reason in the select if the current reason is inactive. This is tricky because we wouldn't want the user to be able to select the current reason if it is inactive, but it's also confusing if we do not show the current reason if it is inactive, so I'm not sure what to do here. Maybe we should include the current reason as a read-only field in addition to the select to change the reason?

eviljeff:
> This is tricky because we wouldn't want the user to be able to select the current reason if it is inactive, but it's also confusing if we do not show the current reason if it is inactive, so I'm not sure what to do here. Maybe we should include the current reason as a read-only field in addition to the select to change the reason?

I don't really like that solution in general as, imo, it's bad UX to have a field that silently changes it's value without the user changing it, (especially as it'd just be whatever the first reason on the list would be, rather than a sensible alternative).  If you save the instance it should only update what you changed, and no other fields.  Though admittedly, this isn't a likely scenario for this form as reason is the only thing you can change anyway.

src/olympia/activity/admin.py
bobsilverberg: As discussed on Slack, I'm now including all of the reasons in the select, but I have added a label of `(** inactive **)` to inactive reasons which hopefully should discourage admins from choosing them.

src/olympia/reviewers/models.py
```diff
@@ -1132,6 +1132,9 @@ class ReviewActionReason(ModelBase):
     )
     name = models.CharField(max_length=255)
 
+    def labelled_name(self):
+        return self.name + ' (** inactive **)' if not self.is_active else self.name
```
eviljeff, line 1136: I'd go with prepending `(inactive) ` rather than appending, thinking about it - `name` can be up to 255 long so it could be easy to overlook.

eviljeff (approved):

PR #18332, opened 2021-11-16
make TranslationSerializerField inherit from CharField; add URLTranslationField

### LINKED ISSUE #18202: Warning messages should be displayed for an invalid e-mail address or URL
Followup for https://github.com/mozilla/addons-server/issues/18037#issuecomment-948586765

support_email, suport_url and homepage accept invalid emails and urls. Example: {"support_email": {"ro":"Lorem ipsum dolor sit amet, consectetuer adipiscin", "it":"Lorem ipsum dolor sit amet, consectetuer adipiscin"}, "support_url": {"ro":"#%^%&&%^&^&^*"}}

Expected result:
a 400 response with a warning message "Enter a valid email address" "Enter a valid URL"

eviljeff:
Most of the changes here were adjusting our `TranslationSerializerField` class to work in a "rest framework"-y way with regards to validation on fields, so we could inherit from `CharField`.  Then subclassing `URLField` or `EmailField` was an easy change to get the validation associated with those classes.  Fixes #18202

### DIFF
--- a/src/olympia/api/fields.py
+++ b/src/olympia/api/fields.py
@@ -48,7 +48,7 @@ class ReverseChoiceField(fields.ChoiceField):
         return super().to_internal_value(value)
 
 
-class TranslationSerializerField(fields.Field):
+class TranslationSerializerField(fields.CharField):
     """
     Django-rest-framework custom serializer field for our TranslatedFields.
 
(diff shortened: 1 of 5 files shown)


src/olympia/api/fields.py
```diff
@@ -387,11 +362,19 @@ def to_representation(self, value):
...
+class EmailTranslationField(serializers.EmailField, TranslationSerializerField):
+    pass
+
+
 class OutgoingURLField(OutgoingSerializerMixin, serializers.URLField):
     pass
 
 
-class OutgoingTranslationField(OutgoingSerializerMixin, TranslationSerializerField):
+class OutgoingURLTranslationField(OutgoingSerializerMixin, URLTranslationField):
```
diox, line 377: Can you rename the ES variant of this field accordingly ?

diox (approved):

src/olympia/api/fields.py
eviljeff: The ES variant technically doesn't inherit from a `URLTranslationField`/`URLField` so I didn't change it (doesn't need to - the only difference between a `CharField` and a `URLField` is the validation, which isn't used for read-only fields) but I can for naming consistency, yep.

PR #18335, opened 2021-11-16
Enable ENABLE_FEATURE_REVIEW_ACTION_REASON on stage and prod

### LINKED ISSUE #18334: Enable the `ENABLE_FEATURE_REVIEW_ACTION_REASON` setting in all environments
We would like to enable the feature in stage and prod, as well as dev. Once this setting is changed, the a user will be able to specify one or more reasons when they are performing certain actions.

bobsilverberg:
Fixes #18334


eviljeff (approved):
So, this effectively enables the feature everywhere and launches it.  Can we have a follow-up issue to remove the setting entirely?

PR #18350, opened 2021-11-17
Extract install_origins from the root of the manifest

diox:
We aren't going to have it in `browser_specific_settings` > `gecko` because at the moment the schema for gecko specific properties doesn't allow new properties, so this wouldn't be backwards-compatible.

Fixes mozilla/addons#8588


willdurand (approved):

willdurand:
(this PR turned out to be my "next tab" and it was green so r+)

PR #18362, opened 2021-11-18
Add logging and timing to _submit_source

### LINKED ISSUE #18337: Add logging and statsd timers around each important step of source code submission
We want to add logging and statsd timing around the various steps in the source code submission views, to narrow down issues we've seen in #18105 when submitting large files.

Note that there are 2 different paths to upload source code: one is at submission, after you've uploaded your add-on or version, and the other is in the version edit view. We could maybe focus on just the submission one (/{locale}/developers/addon/{slug}/versions/submit/{version_id}/source, /{locale}/developers/addon/mante-armorne-sherrew/submit/source-listed and /{locale}/developers/addon/{slug}/submit/source-unlisted) which should be slightly less complex (see also https://github.com/mozilla/addons-server/issues/18336).

bobsilverberg:
Fixes #18337

### DIFF
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -43,7 +43,14 @@ from olympia.amo import messages, utils as amo_utils
 from olympia.amo.decorators import json_view, login_required, post_required
 from olympia.amo.templatetags.jinja_helpers import absolutify, urlparams
 from olympia.amo.reverse import get_url_prefix
-from olympia.amo.utils import MenuItem, escape_all, is_safe_url, render, send_mail
+from olympia.amo.utils import (
+    MenuItem,
+    StopWatch,
+    escape_all,
+    is_safe_url,
+    render,
+    send_mail,
+)
 from olympia.api.models import APIKey, APIKeyConfirmation
 from olympia.devhub.decorators import dev_required, no_admin_disabled
 from olympia.devhub.models import BlogPost, RssKey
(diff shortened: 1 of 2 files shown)


src/olympia/devhub/views.py
```diff
@@ -1576,8 +1592,21 @@ def _submit_source(request, addon, version, submit_page, next_view):
...
+            timer.log_interval('2.form saved')
 
-        return redirect(next_view, *redirect_args)
+        result = redirect(next_view, *redirect_args)
+        log.info(
+            '_submit_source, redirecting to next view, addon.slug: %s, version.pk: %s',
+            addon.slug,
+            version.pk,
+        )
+        timer.log_interval('3.redirecting to next view')
```
eviljeff, line 1608: 2. and 3. are always going to be same time - `redirect` is a shortcut to just building a 301 response object so returns very quickly.

src/olympia/devhub/views.py
```diff
@@ -1576,8 +1592,21 @@ def _submit_source(request, addon, version, submit_page, next_view):
             )
             VersionLog.objects.create(version_id=version.id, activity_log=activity_log)
             form.save()
+            log.info(
+                '_submit_source, form saved, addon.slug: %s, version.pk: %s',
+                addon.slug,
+                version.pk,
+            )
+            timer.log_interval('2.form saved')
```
eviljeff, line 1600: My suspicion is if there is any time spent between 1. and 2. it could be in either the validation (`is_valid`) or the saving (the rest) so maybe there should be timer straight after `is_valid`?

eviljeff (approved):
I'd be interested to know if you get anything useful out of this logging because (as you note yourself in the issue) the view is called after the form post data has worked its way through the django stack so there isn't much to do at that point 🤷

src/olympia/devhub/views.py
bobsilverberg: So I should just remove this one, as it won't be at all useful?

src/olympia/devhub/views.py
bobsilverberg: Ok, I added logging for immediately after `is_valid`.

src/olympia/devhub/views.py
eviljeff: I wouldn't have thought it'd be useful but stranger things have happened

Issue #18361, opened 2021-11-18
add license slug


eviljeff:
for the addon submission api license implementation #18034 we need some way for an api client to be able to specify a pre-defined license.  We have a license `id` (the license record pk), and a `builtin` integer (if `>0` used to indicate the license is pre-defined, and to enumerate the choices), but neither is particularly user-friendly.  Ideally we'd have a slug, for example `mpl2`, `bsd`, `cc`, etc, for the non-custom licenses.

ioanarusiczki:
@eviljeff 

I tried to test this using the info for license slugs from the [docs](https://addons-server.readthedocs.io/en/latest/topics/api/licenses.html). I ran quickly into 429 Too Many Requests while making PATCH requests on versions endpoint, sending license slugs. I don't know what's the new rule on throttles and I don't think I can override it as an admin. 

- at addon create I could send the request with a license slug `"license": "MPL-2.0"` 
- PATCH requests for static themes -> I tested with all license slugs 
- PATCH requests for non-themes -> I got some error for "LGPL-3.0-or-later" and I did not understand the reason why yet, I did have a problem with "all-rights-reserved" but once I associated it with builtinid:18 (in admin) I could use it. Because of "Request was throttled" I'm going to re-check this tomorrow.

eviljeff:
@ioanarusiczki it was the fix for #18122 (that just got reverted)

ioanarusiczki:
@eviljeff I tried to send {"license": "LGPL-3.0-or-later"} but I get a 400 "Invalid value". I checked the admin and I see "GNU Lesser General Public License, version 3.0" with the URL http://www.gnu.org/licenses/lgpl-3.0.html being available. 

The other similar license with "LGPL-2.1-or-later" is alright. Maybe the slug for 3.0 is not correct?

eviljeff:
> @eviljeff I tried to send {"license": "LGPL-3.0-or-later"} but I get a 400 "Invalid value". I checked the admin and I see "GNU Lesser General Public License, version 3.0" with the URL http://www.gnu.org/licenses/lgpl-3.0.html being available.
> 
> The other similar license with "LGPL-2.1-or-later" is alright. Maybe the slug for 3.0 is not correct?

eek, you're right - the slug is wrong.  I'll have to fix it.

PR #18368, opened 2021-11-19
allow custom licenses to be created and updated per version

### LINKED ISSUE #18034: implement full license selection/definition for addon submission api
follow-up for #7811 

currently you can only select from pre-defined licenses - and only by specifying a numeric license id. 

So worked needed to:
- ~select from predefined licenses by slug or some other mnemonic~ #18361
- ~maybe make the field optional (for listed versions only?) and default to all rights reserved, or the previous license for subsequent versions~ #18367

eviljeff:
fixes #18034 - note you still need to know the license id unless you are specifying a custom license, which makes it a little impractical for general use (see #18361)

### DIFF
--- a/docs/topics/api/addons.rst
+++ b/docs/topics/api/addons.rst
@@ -472,6 +472,9 @@ This endpoint allows a submission of an upload to an existing add-on to create a
         production environments. See :ref:`the older signing API <upload-version>` for
         how to submit new add-ons/versions on AMO today.
 
+    .. note::
+        This API requires :doc:`authentication <auth>`.
+
 .. http:post:: /api/v5/addons/addon/(int:addon_id|string:addon_slug|string:addon_guid)/versions/
 
     .. _version-create-request:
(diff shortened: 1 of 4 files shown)


docs/topics/api/addons.rst
```diff
@@ -482,7 +482,9 @@ This endpoint allows a submission of an upload to an existing add-on to create a
         where min/max versions from the manifest, or defaults, will be used.  See :ref:`examples <version-compatibility-examples>`.
     :<json string compatibility[app_name].max: Maximum version of the corresponding app the version is compatible with. Should only be enforced by clients if ``is_strict_compatibility_enabled`` is ``true``.
     :<json string compatibility[app_name].min: Minimum version of the corresponding app the version is compatible with.
-    :<json int license: The license id.
+    :<json int license: The id of a non-custom license. The license must match the add-on type. The license must match the add-on type. Either provide ``license`` or ``custom_license``, not both.
```
bobsilverberg, line 485: It looks like you duplicated "The license must match the add-on type."

bobsilverberg (approved):
This looks good, nice work @eviljeff. I may have missed it, but are there tests for cases where someone submits both a license id and a custom license? Should the API call fail in that case?

r+wc

docs/topics/api/addons.rst
eviljeff: :copypasta: 😬

eviljeff:
> This looks good, nice work @eviljeff. I may have missed it, but are there tests for cases where someone submits both a license id and a custom license? Should the API call fail in that case?

It's a bit of a edge case - because both serializer fields have the same source (`source='license'`, which is implicit for the field also named `license`) then the latter (`custom_license`) overrides the former if you specify both.  I was going to add a validation check for it but it's a little to difficult to even detect as by the time you get to `def validate` the data has already been combined.  I guess I'll call it out in the docs and add a test to confirm my assertion.

eviljeff:
I added a check for both license and custom_license being provided.

PR #18383, opened 2021-11-23
prevent changes from being made to STATUS_DISABLED addons

### LINKED ISSUE #18330: Prevent updates to disabled addons via addons api
In devhub you're prevented from updating most (all?) meta data for a mozilla-disabled add-on (i.e. an add-on we've disabled, rather than user disabled), and from uploading new versions.  We should consistently enforce the same restrictions for the new addon submission/edit apis.

eviljeff:
fixes #18330 - also contains some refactoring of permission handling so the permissions used aren't defined in multiple places.

I toyed with the idea of just changing `AllowAddonAuthor` to always check for `amo.STATUS_DISABLED` (for unsafe http verbs) but unintended consequences on the other API endpoints that also use `AllowAddonAuthor`?


src/olympia/api/tests/test_permissions.py
```diff
@@ -169,6 +170,19 @@ def test_has_object_permission_anonymous(self):
         )
 
 
+class TestAllowAddonAuthorIfNotMozillaDisabled(TestAllowAddonAuthor):
+    def setUp(self):
```
bobsilverberg, line 174: Is it worth having the `setUp` if there's only one test case? Or is it better for have it for consistency?

src/olympia/addons/views.py
```diff
@@ -187,13 +187,18 @@ def find_replacement_addon(request):
 class AddonViewSet(
     CreateModelMixin, RetrieveModelMixin, UpdateModelMixin, GenericViewSet
 ):
+    _write_permissions = [
```
bobsilverberg, line 190: Why the leading underscore in this variable name? How is it different from, for example, `permission_classes` below?

src/olympia/addons/tests/test_views.py
```diff
@@ -1776,6 +1776,14 @@ def test_check_blocklist(self):
         assert 'Version 0.0.1 matches ' in str(response.data['non_field_errors'])
         assert self.addon.reload().versions.count() == 1
 
+    def test_cant_update_disabled_addon(self):
+        self.addon.update(status=amo.STATUS_DISABLED)
+        response = self.client.post(
+            self.url,
+            data=self.minimal_data,
+        )
+        assert response.status_code == 403
```
bobsilverberg, line 1785: I notice in other tests we also assert about `response.data`. Is that worth doing here as well?

bobsilverberg (approved):
Looks good. I just have a few questions.

r+wc

src/olympia/api/tests/test_permissions.py
eviljeff: The class doesn't have only one test - it inherits from `TestAllowAddonAuthor`

src/olympia/addons/views.py
eviljeff: it was supposed to be a private class property (well, not really, because python doesn't have those except by convention) but it ended up being referenced in other classes so I guess that isn't true any longer.  
`permission_classes` is a property on `GenericViewSet` we're overriding/setting btw.

src/olympia/addons/tests/test_views.py
eviljeff: added extra assertions in https://github.com/mozilla/addons-server/pull/18383/commits/777df1f9cd22b2b061c4e2675b2327f4c7f95b99

src/olympia/addons/views.py
eviljeff: renamed in https://github.com/mozilla/addons-server/pull/18383/commits/777df1f9cd22b2b061c4e2675b2327f4c7f95b99

PR #18387, opened 2021-11-23
Do not display or require reasons for actions for themes

### LINKED ISSUE #18380: Don't show or require reasons for theme reviews
For the #18168 Reviewer Statistics Improvement project, we are less interested in review decision stats for themes. For the time being, we should not require a reason, nor show the reasons, when a theme is being reviewed. 

If this is difficult to do I'm happy to talk about alternatives:
* Allow selecting what reasons apply to different kinds of add-ons, e.g. some reasons only apply to themes
* Accept doing nothing until at least Q1, accepting that theme reviewers will need to select specific reasons from the list that we tailor towards themes.

bobsilverberg:
Fixes #18380

### DIFF
--- a/src/olympia/reviewers/forms.py
+++ b/src/olympia/reviewers/forms.py
@@ -212,10 +212,7 @@ class ReviewForm(forms.Form):
                 self.fields['comments'].required = False
             if action.get('versions', False):
                 self.fields['versions'].required = True
-            if (
-                not action.get('requires_reasons', False)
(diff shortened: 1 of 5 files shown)


diox (changes requested):
IMHO for this kind of requirement it's better to make the properties depend on the addon in `ReviewHelper.get_actions()` itself, to avoid having checks all over the place. That's how it's done for other properties like `available` or `delayable`.

diox (approved):

PR #18392, opened 2021-11-24
Add ability to fail validation for denied install_origins patterns entered by admins

### LINKED ISSUE #18331: Add ability to prevent submission of add-ons whose `install_origins` contain specific origins
It should be possible through an admin tool to add a pattern that would prevent certain origins from being used in `install_origins` when `record-install-origins` waffle switch is enabled.

- Shell-style (`*` to match everything) patterns should be supported.
- IDN origins should be supported, both in unicode or punycode format. The pattern can also be in either format, and would block both variants if they appear in the `install_origins`.
  - For example, a pattern of `*.栃木.jp` would block `https://foo.bar.栃木.jp` as well as `https://foo.xn--4pvxs.jp` origins. Similarly, a pattern of `*.xn--4pvxs.jp` would block `https://bar.栃木.jp` and `https://bar.xn--4pvxs.jp`
- When adding new patterns in the admin, an option needs to be present in the admin to automatically add a variant of the input pattern that matches subdomains.
  - For example, if a pattern of `foo.example.com` is provided, if that option is checked, behind the scenes that pattern would also block `*.foo.example.com`. Admins would then be free to change that option later if they wish (that wouldn't affect past uploads that have already been validated)
- If `install_origins` for an add-on being uploaded contains one or more origins that match one if the disallowed patterns, that submission should be considered invalid at upload time.

In addition, invalid `install_origins` values don't need to be blocked by a pattern, since they would cause validation to fail anyway thanks to the linter, but shouldn't cause 500/unknown errors.

(See https://github.com/mozilla/addons-server/issues/18134 for more details about `install_origins`)

diox:
Fixes #18331


src/olympia/devhub/tests/test_file_validation_annotations.py
```diff
@@ -0,0 +1,72 @@
...
+        assert results['messages'][0] == {
+            'tier': 1,
+            'type': 'error',
+            'id': ['validation', 'messages', ''],
+            'message': 'The install origin https://foo.com is not permitted.',
+            'description': [],
+            'compatibility_type': None,
+        }
+
+    def test_invalid_origins_multiple_matches(self):
```
bobsilverberg, line 57: If multiple origins are specified which fail validation, would that result in multiple validation failure messages? If so, should we have a test for that?

bobsilverberg (approved):
Look good, nice work @diox! I just had one question about a test.

r+wc

PR #18400, opened 2021-11-25
Clean-up / tighten CSP config

### LINKED ISSUE #18012: Clean up / tighten addons-server CSP config

diox:
Fixes #18012

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1270,8 +1270,7 @@ CSP_EXCLUDE_URL_PREFIXES = ()
 
 # NOTE: CSP_DEFAULT_SRC MUST be set otherwise things not set
 # will default to being open to anything.
-CSP_DEFAULT_SRC = ("'self'",)
-CSP_BASE_URI = ("'self'",)
+CSP_DEFAULT_SRC = ("'none'",)
 CSP_CONNECT_SRC = (
     "'self'",
     'https://sentry.prod.mozaws.net',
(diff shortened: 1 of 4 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -1309,7 +1305,6 @@ def get_db_config(environ_var, atomic_requests=True):
     PROD_STATIC_URL,
 )
 CSP_STYLE_SRC = (
-    "'self'",
```
eviljeff, line 1312: don't we need this for admin pages that have some django css?

eviljeff (approved):
Testing on -dev to see what breaks is much more valuable than (my) code review, but seems gtm (one question on css)

src/olympia/lib/settings_base.py
diox: django admin css should be served from `STATIC_URL` (`https://a.m.o./static/`)

bobsilverberg (approved):
I don't see any issues with it.

PR #18401, opened 2021-11-25
Replace last use of STATIC_URL with staticfile's storage.url()

### LINKED ISSUE #18104: Clean up usage of STATIC_URL

diox:
Fixes #18104

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -1177,11 +1177,7 @@ class Addon(OnChangeMixin, ModelBase):
             return jinja_helpers.user_media_url('addon_icons') + path
 
     def get_default_icon_url(self, size):
-        # We don't update the default icon very frequently so there is no
-        # automated query parameter for cachebusting...
-        return '{}img/addon-icons/{}-{}.png?v=20210601'.format(
-            settings.STATIC_URL, 'default', size
-        )
+        return staticfiles_storage.url(f'img/addon-icons/default-{size}.png')
 
     @use_primary_db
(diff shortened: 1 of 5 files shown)


src/olympia/addons/models.py
```diff
@@ -1177,11 +1177,7 @@ def get_icon_url(self, size, use_default=True):
             return jinja_helpers.user_media_url('addon_icons') + path
 
     def get_default_icon_url(self, size):
-        # We don't update the default icon very frequently so there is no
-        # automated query parameter for cachebusting...
-        return '{}img/addon-icons/{}-{}.png?v=20210601'.format(
-            settings.STATIC_URL, 'default', size
-        )
+        return staticfiles_storage.url(f'img/addon-icons/default-{size}.png')
```
eviljeff, line 1180: does this mean we've dropped the cachebusting?  Or `staticfiles_storage.url` handles it for us?  (I assumed the latter but the tests below don't include any params?)

diox: The latter, but in tests we don't want to run collectstatic, so `STATICFILES_STORAGE` is not set to `ManifestStaticFilesStorage`, so there is no cachebusting. There is in prod (same for django's `static() templatetag, which calls `staticfiles_storage.url()` behind the scenes)

eviljeff (approved):

PR #18408, opened 2021-11-26
Show some helpful text in the blocklist admin (#18407)

### LINKED ISSUE #18407: Show some helpful text in the blocklist admin
image

kewisch:
Fixes #18407
Adds some helpful for admins current and future.


eviljeff (commented):
can you generate a migration file (django will want one because you updated the kwargs on the fields - irregardless of if they result in underlying database changes)

PR #18413, opened 2021-11-28
Drop is_webextension field from files table in database.

### LINKED ISSUE #18409: Drop `is_webextension` field from `files` table in database
We removed `is_webextension` from the model (https://github.com/mozilla/addons-server/pull/18018) but it looks like we haven't dropped it from the database table yet. It should just be a matter of generating the migration.

tejaram15:
Fixes #18409


src/olympia/files/migrations/0017_auto_20211128_0808.py
```diff
@@ -0,0 +1,22 @@
...
+    operations = [
+        migrations.RemoveField(
+            model_name='file',
+            name='is_webextension',
+        ),
+        migrations.AlterField(
+            model_name='fileupload',
+            name='source',
+            field=models.PositiveSmallIntegerField(choices=[(1, 'Developer Hub'), (2, 'Signing API'), (3, 'Addon API')], default=None, null=True),
+        ),
```
diox, line 21: Please remove that part as it's not related (we'll need to have it, but it doesn't cause any changes to the database and can be done separately)

tejaram15: Hi @diox removed the above operation in the new commit.

diox (approved):

Issue #18427, opened 2021-11-30
add api endpoint that exposes available tags


eviljeff:
The addon api will soon allow tags to be edited #18268 and the tag selection is limited to what admin/curation staff have chosen.  But we need some way for clients to know the available tags (documenting it in the API isn't practical because the tags are defined in the database so can change at any time without code changes)

ioanarusiczki:
@eviljeff  https://addons-dev.allizom.org/api/v5/addons/tags/ is responding with a 200 and returns the list of tags for AMO dev.

PR #18428, opened 2021-11-30
set tags property via addon api

### LINKED ISSUE #18268: allow tags to be writeable via the addon api
part of #18036 - addon tags should be writeable.  The tags are only allowed from a pre-selected list (how to expose that list via the API?), and limited to how many can be selected per addon.

eviljeff:
fixes #18268 - filed #18427 for an endpoint to list valid tags

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -1793,6 +1793,23 @@ class Addon(OnChangeMixin, ModelBase):
 
         return GitExtractionEntry.objects.filter(addon=self, in_progress=True).exists()
 
+    @cached_property
+    def tag_list(self):
+        attach_tags([self])
+        return self.tag_list
+
+    def set_tag_list(self, new_tag_list):
+        tag_list_to_add = set(new_tag_list) - set(self.tag_list)
+        tag_list_to_drop = set(self.tag_list) - set(new_tag_list)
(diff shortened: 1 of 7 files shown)


src/olympia/addons/models.py
```diff
@@ -1793,6 +1793,23 @@ def git_extraction_is_in_progress(self):
 
         return GitExtractionEntry.objects.filter(addon=self, in_progress=True).exists()
 
+    @cached_property
+    def tag_list(self):
+        attach_tags([self])
+        return self.tag_list
```
diox, line 1799: I... never thought of doing it that way (returning the same property in a `@cached_property` method) - before. Neat.

src/olympia/api/fields.py
```diff
@@ -548,3 +548,27 @@ def to_representation(self, value):
             if rep:
                 return rep
         return rep
+
+
+class LazyChoiceField(fields.ChoiceField):
```
diox, line 553: We should have dedicated tests for this

diox (approved):

src/olympia/api/fields.py
eviljeff: https://github.com/mozilla/addons-server/pull/18428/commits/93d7e9848d1399e626c346931765fef3a1773a7a

Issue #18405, opened 2021-11-26
Make site permission addons go through auto-approve


diox:
Site Permission add-ons from #18135 will need to be signed after they have been generated, going through `auto_approve` as normal.  We'll probably want to customize the approval email.

diox:
I... don't think I need to change anything for them to go through `auto_approve`, since they are unlisted and we don't restrict add-on types for unlisted versions:
https://github.com/mozilla/addons-server/blob/36b320b1beeb9f2032d3f02d76f754eaaee056ab/src/olympia/versions/models.py#L119-L126

That being said, customizing the approval email might still be worth it, I'll look into it.

diox:
Default email is https://github.com/mozilla/addons-server/blob/master/src/olympia/reviewers/templates/reviewers/emails/unlisted_to_reviewed_auto.ltxt which looks fine for this, at least for now.

PR #18429, opened 2021-11-30
Add backend to generate site permission add-ons

### LINKED ISSUE #18135: Add ability to generate site permission add-ons
Add ability to generate and sign unlisted add-ons that enable a list of permissions for an origin, built on top of https://github.com/mozilla/addons-server/issues/18134

- The add-on will appear as an add-on with a single unlisted version
- Everything about it is automatically generated, including name, description but also manifest and xpi package itself
- It will belong to the developer account that requested it
- We need to record what permissions it enables in a new table

The `.xpi` will contain only a manifest, in which, besides mandatory extension fields, we'll have something like (property names subject to change):
```json
{
    "install_origins": ["https://sub.foo.com"],
    "site_permissions": ["permission"]
}
```

Follow-ups:
- They will be signed through `auto_approve` as normal, though we may want to customize the approval email. https://github.com/mozilla/addons-server/issues/18405

diox:
When Firefox implementation is done, they'll allow some specific APIs to be used by the site from which they are installed.

In AMO:
- A new type is added for them, `ADDON_SITE_PERMISSION`
- They are automatically generated by the class added in this commit. They can't be submitted through regular add-on upload (a special form will eventually trigger the generation for a given origin/permission/user)
- They have one or more unlisted versions as normal (which will eventually go through auto_approve)
- What they enable is recorded in the new `FileSitePermission` model, and each origin attached to a version is recorded like regular distribution control mechanism, through the `InstallOrigin` model
- ~~They can't be edited through devhub or the API~~

Fixes #18135


diox:
> They can't be edited through devhub or the API

We'll need a better implementation, but for now that is treated as a mozilla-disabled add-on, which is not great cause you see the link in devhub, but you get a 404 as soon as you click on it. It's probably good enough to begin with, though.

src/olympia/devhub/tasks.py
```diff
@@ -118,6 +118,7 @@ def create_version_for_upload(addon, upload, channel):
             'Skipping Version creation for {upload_uuid} that would '
             ' cause duplicate version'.format(upload_uuid=upload.uuid)
         )
+        return None
```
eviljeff, line 121: obsolete line?

src/olympia/devhub/tasks.py
```diff
@@ -106,7 +106,7 @@ def submit_file(addon_pk, upload_pk, channel):
 
 
 @transaction.atomic
-def create_version_for_upload(addon, upload, channel):
+def create_version_for_upload(addon, upload, channel, parsed_data=None):
```
eviljeff, line 109: 😬 I forgot we have this complicated non-task in tasks.py ... should it be move to devhub/utils.py?

src/olympia/addons/utils.py
```diff
@@ -247,3 +259,130 @@ def get_error_message(self):
...
+
+class PermissionEnablerCreator:
+    """Helper class to create new permission enabler add-ons and versions.
+
+    Assumes parameters have already been validated beforehand."""
+
+    SITE_PERMISSIONS = {
+        'webmidi': 'WebMIDI',
+    }
+    MIN_VERSION = '96.0'
```
eviljeff, line 272: I think we should move these to constants (a specific file maybe), if they're going to grow to a longer list of permissions.

src/olympia/addons/utils.py
```diff
@@ -247,3 +259,130 @@ def get_error_message(self):
...
+            }
+        return manifest_data
+
+    def _create_zipfile(self, manifest_data):
+        # Create the xpi containing the manifest, in memory.
+        raw_buffer = BytesIO()
+        with zipfile.ZipFile(raw_buffer, 'w') as zip_file:
+            zip_file.writestr('manifest.json', json.dumps(manifest_data, indent=2))
+        raw_buffer.seek(0)
+        filename = 'automatic.xpi'
```
eviljeff, line 320: a more descriptive file name? (containing hostname or permission, etc)

src/olympia/addons/utils.py
```diff
@@ -247,3 +259,130 @@ def get_error_message(self):
...
+        manifest_data = self._create_manifest(version_number)
+        file_obj = self._create_zipfile(manifest_data)
+
+        # Parse the zip we just created. The user needs to be the Mozilla User
+        # because regular submissions of this type of add-on is forbidden to
+        # normal users.
+        parsed_data = parse_addon(
+            file_obj,
+            addon=addon,
+            user=UserProfile.objects.get(pk=settings.TASK_USER_ID),
```
eviljeff, line 352: nit: `from olympia.users.utils import get_task_user`

src/olympia/addons/utils.py
```diff
@@ -247,3 +259,130 @@ def get_error_message(self):
...
+        else:
+            # If the user isn't an author, something is wrong.
+            if not addon.authors.filter(pk=self.user.pk).exists():
+                raise ImproperlyConfigured(
+                    'PermissionEnablerCreator was instantiated with a bogus addon/user'
+                )
+            # FIXME: if we already have an add-on, we'll likely want to update
+            # its name and description so that it matches the latest manifest
+            # values (which we have in parsed_data).
+
```
eviljeff, line 372: Are we checking anywhere that the addon is already an `ADDON_PERMISSION_ENABLER` to prevent an addon ending up with a mix of types?  (I would hope that would raise elsewhere though)

eviljeff (approved):

diox:
Note to self: I need to add the ability for people with a special permission to upload such add-ons themselves, to help the WebExt team generate test add-ons

diox:
Another thing to add: we don't want to allow multiple origins at all right now, so we can restrict it.

src/olympia/addons/utils.py
```diff
@@ -247,3 +259,130 @@ def get_error_message(self):
...
         return msg
+
+
+class PermissionEnablerCreator:
+    """Helper class to create new permission enabler add-ons and versions.
+
+    Assumes parameters have already been validated beforehand."""
+
+    SITE_PERMISSIONS = {
+        'webmidi': 'WebMIDI',
```
diox, line 270: New name is actually `midi-sysex` (there are 2 permissions behind the scenes, but the only one we'll care about to start is that new one)

src/olympia/addons/utils.py
diox: Yep, I'll do that. And the min version needs to be 97 actually.

src/olympia/access/acl.py
```diff
@@ -149,8 +149,13 @@ def check_addon_ownership(request, addon, dev=False, admin=True, ignore_disabled
     # Users with 'Addons:Edit' can do anything.
     if admin and action_allowed(request, amo.permissions.ADDONS_EDIT):
         return True
-    # Only admins can edit admin-disabled addons.
-    if addon.status == amo.STATUS_DISABLED and not ignore_disabled:
+    # Only admins can edit admin-disabled addons or permission-enabler add-ons,
+    # unless the caller determined it was ok (for specific scenarios like
+    # read-only requests).
+    if not ignore_disabled and (
```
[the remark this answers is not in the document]
diox: I was afraid you'd say that :) I didn't want to touch the signature because `ignore_disabled` is also a parameter in the more generic `check_ownership()` function (and methods on some objects like `Collection`), but yeah, I need to do something about that cause it's all terrible...

src/olympia/addons/utils.py
diox: Ultimately it doesn't really matter that much as it's all in memory, it won't get written to the filesystem with that name. I just need _a_ name so that I can use a `File` object, which I need because that's the kind of object `parse_addon()` expects.

src/olympia/addons/utils.py
diox: Good point, it doesn't hurt to throw an exception if the `type` is wrong

src/olympia/constants/base.py
```diff
@@ -175,6 +181,7 @@
     _ADDON_PERSONA: 'background-theme',
     ADDON_PLUGIN: 'plugin',
     ADDON_STATICTHEME: 'static-theme',
+    ADDON_PERMISSION_ENABLER: 'permission-enabler',
```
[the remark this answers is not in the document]
diox: Ah, yes, you're right. I'll double check with Shane but we probably want `extension` in there.

src/olympia/devhub/tasks.py
diox: It's just to be explicit, with the `else` clause returning `version` now.

src/olympia/files/utils.py
```diff
@@ -255,6 +255,8 @@ def type(self):
             if 'theme' in self.data
             else amo.ADDON_DICT
             if 'dictionaries' in self.data
+            else amo.ADDON_PERMISSION_ENABLER
+            if 'site_permissions' in self.data
```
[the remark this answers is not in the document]
diox: I'm not sure if that's even allowed by the schema. If it is, it shouldn't be.

src/olympia/constants/base.py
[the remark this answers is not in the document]
diox: Actually, we no longer expose that in the update response. See https://versioncheck.addons.mozilla.org/update/versionCheck.php?reqVersion=1&id=uBlock0@raymondhill.net&version=1.38.0&maxAppVersion=96.0&appID={ec8030f7-c20a-464f-9b0e-13a3a9e97384}&appVersion=96.0&appOS=WINNT&appABI=x86-msvc for an example. In `get_success_output()` in the update code, we don't use `data['type']` at all.



I'll remove it and adjust tests that depend on it.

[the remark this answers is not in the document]
diox: https://github.com/mozilla/addons-server/issues/18442 for the removal actually since it turned to be a little more involved.

src/olympia/devhub/tasks.py
eviljeff: I'm surprised there isn't a lint rule about having `return` rather than `return None`

src/olympia/files/utils.py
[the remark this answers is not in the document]
diox: FWIW, Firefox has similar logic:
```javascript
   if (this.manifest.theme) {
      this.type = "theme";
    } else if (this.manifest.langpack_id) {
      this.type = "langpack";
    } else if (this.manifest.dictionaries) {
      this.type = "dictionary";
    } else if (this.manifest.site_permissions) {
      this.type = "sitepermission";
    } else {
      this.type = "extension";
    }
```

The ordering isn't completely the same, but the idea is similar.

src/olympia/constants/base.py
```diff
@@ -162,6 +167,7 @@
     ADDON_LPAPP: 'language-tools',
     _ADDON_SEARCH: 'search-tools',
     ADDON_STATICTHEME: 'themes',
+    ADDON_SITE_PERMISSION: 'sitepermission',
```
eviljeff, line 170:
```suggestion
    ADDON_SITE_PERMISSION: 'site-permission',
```
? (and for `ADDON_TYPE_CHOICES_API`)

src/olympia/files/migrations/0019_rename_sitepermission_filesitepermission.py
```diff
@@ -0,0 +1,17 @@
...
+    dependencies = [
+        ('files', '0018_auto_20211130_1539'),
+    ]
+
+    operations = [
+        migrations.RenameModel(
+            old_name='SitePermission',
+            new_name='FileSitePermission',
+        ),
+    ]
```
eviljeff, line 17: just squash these two migration together, imo (apologies to your local db migration state 😛 )

src/olympia/versions/models.py
```diff
@@ -114,7 +114,12 @@ def auto_approvable(self):
                 channel=amo.RELEASE_CHANNEL_LISTED,
                 addon__status__in=(amo.STATUS_NOMINATED, amo.STATUS_APPROVED),
                 addon__disabled_by_user=False,
-                addon__type__in=(amo.ADDON_EXTENSION, amo.ADDON_LPAPP, amo.ADDON_DICT),
+                addon__type__in=(
+                    amo.ADDON_EXTENSION,
+                    amo.ADDON_PERMISSION_ENABLER,
+                    amo.ADDON_LPAPP,
+                    amo.ADDON_DICT,
```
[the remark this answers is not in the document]
diox: I'll also revert this for now and fully deal with it (including testing) in https://github.com/mozilla/addons-server/issues/18405

src/olympia/access/acl.py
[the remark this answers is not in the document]
diox: https://github.com/mozilla/addons-server/issues/18444 is the follow-up

src/olympia/constants/base.py
diox: This is my overly optimistic attempt to bring us closer to what Firefox uses (we're already diverging since they use `theme` for `statictheme`, and `langpack` and not `language`)

eviljeff: I look forward to the future patch to update those in the v5 api 😉

src/olympia/constants/site_permissions.py
```diff
@@ -0,0 +1,4 @@
+# List of site_permissions names we know about.
+SITE_PERMISSIONS = {'midi-sysex'}  # WebMIDI
```
eviljeff, line 2: nit: comment says "list" but it's defined as a set

diox: mmm now that I don't generate descriptions (Shane told me that wasn't necessary as about:addons will display the hosts/permissions in a better way) I don't need this until I start working on the form to validate the permissions we're asked to generate. So I'll remove for now.

PR #18435, opened 2021-12-01
Remove obsolete version information from docs

diox:


eviljeff (approved):
r+ if you meant to just comment out the version settings rather the delete the lines entirely

diox:
I figured commenting out was fine since that's how the rest of the conf is set up already

PR #18443, opened 2021-12-02
Remove dead code from update service

### LINKED ISSUE #18442: Removed unused columns/code in update service database query
The update service does a `SELECT` with a bunch of columns:
https://github.com/mozilla/addons-server/blob/7bc09ac344ae9a2756acdfc88d6308e449df2f37/services/update.py#L91-L104

Then when building the results, they are referenced again and part of the data used to build the update:
https://github.com/mozilla/addons-server/blob/7bc09ac344ae9a2756acdfc88d6308e449df2f37/services/update.py#L165-L178

But if we look at the code below, in `get_success_output()` and `get_no_updates_output()`, only a few are actually needed, so we could remove some. We can also remove the awkward `ADDON_SLUGS_UPDATE` constant.

In addition, there is also a check using `applications.D2C_MIN_VERSIONS` that is completely obsolete since we no longer have legacy extensions.

diox:
Fixes #18442

### DIFF
--- a/services/update.py
+++ b/services/update.py
@@ -58,7 +58,7 @@ class Update:
 
         data['app_id'] = app.id
 
-        sql = """SELECT `id`, `status`, `addontype_id`, `guid` FROM `addons`
+        sql = """SELECT `id`, `status`, `guid` FROM `addons`
                  WHERE `guid` = %(guid)s AND
                        `inactive` = 0 AND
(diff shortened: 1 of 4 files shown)


eviljeff (approved):
it's a gnarly chunk of code and I could easily have overlooked something but r+

PR #18445, opened 2021-12-03
add /addon/tags/ endpoint to list available tags

### LINKED ISSUE #18427: add api endpoint that exposes available tags

eviljeff:
fixes #18427 - I used the same cache control override as the category list view

### DIFF
--- a/docs/topics/api/addons.rst
+++ b/docs/topics/api/addons.rst
@@ -113,7 +113,7 @@ for autocomplete though, there are a couple key differences:
     :query string author: Filter by exact (listed) author username. Multiple author names can be specified, separated by comma(s), in which case add-ons with at least one matching author are returned.
     :query string category: Filter by :ref:`category slug <category-list>`. ``app`` and ``type`` parameters need to be set, otherwise this parameter is ignored.
     :query string lang: Activate translations in the specific language for that query. (See :ref:`translated fields <api-overview-translations>`)
-    :query string tag: Filter by exact tag name. Multiple tag names can be specified, separated by comma(s).  See :ref:`available tags <tag-list>`
+    :query string tag: Filter by exact tag name. Multiple tag names can be specified, separated by comma(s), in which case add-ons containing *all* specified tags are returned. See :ref:`available tags <tag-list>`
     :query string type: Filter by :ref:`add-on type <addon-detail-type>`.
     :>json array results: An array of :ref:`add-ons <addon-detail-object>`. Only the ``id``, ``icon_url``, ``name``, ``promoted``, ``type`` and ``url`` fields are supported though.
 
--- a/docs/topics/api/addons.rst
+++ b/docs/topics/api/addons.rst
@@ -33,7 +33,7 @@ This endpoint allows you to search through public add-ons.
     :query string promoted: Filter to add-ons in a specific :ref:`promoted category <addon-detail-promoted-category>`.  Can be combined with `app`.   Multiple promoted categories can be specified, separated by comma(s), in which case any add-ons in any of the promotions will be returned.
     :query string ratings: Filter to add-ons that have average ratings of a :ref:`threshold value <addon-threshold-param>`.
     :query string sort: The sort parameter. The available parameters are documented in the :ref:`table below <addon-search-sort>`.
-    :query string tag: Filter by exact tag name. Multiple tag names can be specified, separated by comma(s), in which case add-ons containing *all* specified tags are returned.
+    :query string tag: Filter by exact tag name. Multiple tag names can be specified, separated by comma(s), in which case add-ons containing *all* specified tags are returned. See :ref:`available tags <tag-list>`
     :query string type: Filter by :ref:`add-on type <addon-detail-type>`.  Multiple types can be specified, separated by comma(s), in which case add-ons that are any of the matching types are returned.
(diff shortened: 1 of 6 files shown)


src/olympia/tags/tests/test_views.py
```diff
@@ -0,0 +1,25 @@
+from olympia.amo.tests import APITestClient, TestCase, reverse_ns
+
+from ..models import Tag
+
+
+class TestStaticCategoryView(TestCase):
```
bobsilverberg, line 6: Should this be called `TestStaticTagView`?

src/olympia/tags/tests/test_views.py
```diff
@@ -0,0 +1,25 @@
...
+    client_class = APITestClient
+
+    def setUp(self):
+        super().setUp()
+        self.url = reverse_ns('tag-list')
+
+    def test_basic(self):
+        response = self.client.get(self.url)
+        assert response.status_code == 200
+        tags = list(Tag.objects.all())
```
bobsilverberg, line 16: Will there be any existing tags in the test environment when this is run?

src/olympia/tags/views.py
```diff
@@ -0,0 +1,24 @@
...
+from rest_framework.views import APIView
+
+from .models import Tag
+
+
+class TagListView(APIView):
+    authentication_classes = []
+
+    def get(self, request, *args, **kwargs):
+        return Response(list(Tag.objects.values_list('tag_text', flat=True)))
```
bobsilverberg, line 14: Should we return these sorted, or is that done automatically based on a model annotation?

docs/topics/api/addons.rst
```diff
@@ -113,7 +113,7 @@ for autocomplete though, there are a couple key differences:
     :query string author: Filter by exact (listed) author username. Multiple author names can be specified, separated by comma(s), in which case add-ons with at least one matching author are returned.
     :query string category: Filter by :ref:`category slug <category-list>`. ``app`` and ``type`` parameters need to be set, otherwise this parameter is ignored.
     :query string lang: Activate translations in the specific language for that query. (See :ref:`translated fields <api-overview-translations>`)
-    :query string tag: Filter by exact tag name. Multiple tag names can be specified, separated by comma(s).
+    :query string tag: Filter by exact tag name. Multiple tag names can be specified, separated by comma(s).  See :ref:`available tags <tag-list>`
```
bobsilverberg, line 116: The description for the `search` endpoint includes ", in which case add-ons containing *all* specified tags are returned.", but this one doesn't specify that. Is the behaviour in fact different, or should that be added to this description?

src/olympia/tags/views.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @classmethod
+    def as_view(cls, **kwargs):
+        view = super().as_view(**kwargs)
+        return non_atomic_requests(view)
```
bobsilverberg, line 19: Just curious, what does this do and why is it needed?

bobsilverberg (commented):
This looks generally good, but I added a bunch of comments/questions.

src/olympia/tags/tests/test_views.py
eviljeff: You can see where I copied the `TestCase` from 😬

src/olympia/tags/tests/test_views.py
eviljeff: yeah there's a migration that's run while setting up the test db that adds all tags that were defined a few months back.

src/olympia/tags/views.py
eviljeff: the model defines a default order (by `tag_text`, which is what we want)

docs/topics/api/addons.rst
eviljeff: Not sure, I can investigate (I didn't touch anything to do with filters or the search views, I just thought it helpful to link to the new tag page while I was at it)

src/olympia/tags/tests/test_views.py
bobsilverberg: Would it be worth asserting that `len(response.data) > 0`, just to make me feel more secure about this (as well as any future reader of the code)? I added the comment because I notice that we often generate test data in the test itself, but we're not doing that here.

src/olympia/tags/views.py
eviljeff: we have [atomic](https://docs.djangoproject.com/en/3.2/topics/db/transactions/#django.db.transaction.atomic) requests enabled by default - that adds database queries at the start and end of the request cycle for transactions. This just disables that specifically for this view -because we know it won't ever result in database writes- for efficiency.

docs/topics/api/addons.rst
eviljeff: I poked around in the code and the autocomplete endpoint actually supports the same parameters as the full search endpoint (apart from `sort`, in v4+) despite the docs not mentioning them all.  I'm not going to update everything (that's veering off track for this pr) but I'll update the `tags` entry.

PR #18500, opened 2021-12-13
(optionally) force Outgoing url fields to deny setting internal urls

### LINKED ISSUE #18450: support_url and homepage should only accept external websites
While checking for https://github.com/mozilla/addons-server/issues/18330 I send PATCH requests and used https://addons-dev.allizom.org/en-US/ for support_url and homepage: 

`{"slug": "thisisa_slug_7", "categories": {"firefox": ["search-tools"]}, "description": {"en-US": "testareinromana"}, "developer_comments": {"en-US": "testez aici cu de toate"}, "homepage": { "en-US":"https://addons-dev.allizom.org/"},"name": {"en-US":"apply changes"}, "summary": {"en-US": "a Festivus for the rest of us"}, "support_email": {"en-US": "change@ro.com"}, "support_url": {"en-US":"https://addons-dev.allizom.org/en-US/"}}`

This should throw an error such as the one below: 

not allowed

eviljeff:
fixes #18450

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -1082,6 +1082,7 @@ class TestAddonViewSetCreate(UploadMixin, TestCase):
         assert addon.is_disabled is True
         assert addon.disabled_by_user is True  # sets the user property
 
+    @override_settings(EXTERNAL_SITE_URL='https://amazing.site')
     def test_set_homepage_support_url_email(self):
         data = {
             **self.minimal_data,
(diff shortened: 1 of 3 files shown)


bobsilverberg (approved):
Looks good. I know this has nothing to do with this patch, but it strikes me as odd that `settings.EXTERNAL_SITE_URL` is actually an *internal* URL, unless I am totally misunderstanding this.

eviljeff:
> Looks good. I know this has nothing to do with this patch, but it strikes me as odd that `settings.EXTERNAL_SITE_URL` is actually an _internal_ URL, unless I am totally misunderstanding this.

e.g. for the admin instance it's https://addons.mozilla.org/ vs. https://addons-internal.prod.mozaws.net/ - the external facing site url of the instance vs. the site url you may be connecting to (although there are other more "internal" urls too - `HOSTNAME` and `DOMAIN` could be different).  For the public instances it'll be the same but we default to `EXTERNAL_SITE_URL` when building absolute urls, so I thought it safer/consistent. The naming of the settings could be improved I suppose!

PR #18512, opened 2021-12-15
add heartbeat urls

### LINKED ISSUE #18510: Please add healthcheck URLs

Per https://mana.mozilla.org/wiki/display/SVCOPS/New+Service+Guide#NewServiceGuide-StandardEndpoints, all applications running in containers should support a few standard endpoints.

Although the addons-server supports the `__version__` endpoint, it does not support the following health check URLs,

* `/__heartbeat__`
* `/__lbheartbeat__`

Can we add them please?

### What happened?

The addons-frontend recently had an incident that has to do with using TCP port to monitor app health. We filed https://github.com/mozilla/addons-frontend/issues/11059 for it, and would like the addons-server to support the above two health check endpoints as well, so that we can start to using them.

The addons-server should support the above said two health check URLs (per Mozilla standard).

n/a

eviljeff:
fixes #18510 
- repurposes `/services/monitor.json` as `/__heartbeat__` (but leaves the url alone as an alias for now)
- adds `/__lbheartbeat__`
- adds a database check for replica and primary databases.

### DIFF
--- a/src/olympia/amo/middleware.py
+++ b/src/olympia/amo/middleware.py
@@ -11,13 +11,19 @@ from django.contrib.sessions.middleware import SessionMiddleware
 from django.core.exceptions import ImproperlyConfigured
 from django.db import transaction
 from django.http import (
+    HttpResponse,
     HttpResponsePermanentRedirect,
     HttpResponseRedirect,
     JsonResponse,
 )
 from django.middleware import common
 from django.template.response import TemplateResponse
-from django.utils.cache import get_max_age, patch_cache_control, patch_vary_headers
+from django.utils.cache import (
+    add_never_cache_headers,
+    get_max_age,
+    patch_cache_control,
+    patch_vary_headers,
+)
 from django.utils.crypto import constant_time_compare
 from django.utils.deprecation import MiddlewareMixin
 from django.utils.encoding import force_str, iri_to_uri
@@ -352,3 +358,19 @@ class CacheControlMiddleware:
         elif max_age_from_response is None:
             patch_cache_control(response, s_maxage=0)
         return response
+
+
+class LBHeartbeatMiddleware:
+    """Middleware to capture request to /__lbheartbeat__ and return a 200.
+    Must be placed above CommonMiddleware to work with ELB.
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if request.path == '/__lbheartbeat__':
+            response = HttpResponse(status=200)
+            add_never_cache_headers(response)
+            return response
+        return self.get_response(request)
--- a/src/olympia/amo/monitors.py
+++ b/src/olympia/amo/monitors.py
@@ -13,6 +13,7 @@ from PIL import Image
 import olympia.core.logger
 
 from olympia.amo import search
+from olympia.amo.models import use_primary_db
 from olympia.amo.templatetags.jinja_helpers import user_media_path
 
 
@@ -177,3 +178,24 @@ def signer():
         signer_results = False
 
     return status, signer_results
+
+
+def database():
+    # check database connection
+    from olympia.addons.models import Addon
+
+    status = ''
+    try:
+        Addon.unfiltered.exists()
+    except Exception as e:
+        status = f'Failed to connect to replica database: {e}'
+        monitor_log.critical(status)
+    else:
+        with use_primary_db():
+            try:
+                Addon.unfiltered.exists()
+            except Exception as e:
+                status = f'Failed to connect to primary database: {e}'
+                monitor_log.critical(status)
+
+    return status, None
--- a/src/olympia/amo/tests/test_middleware.py
+++ b/src/olympia/amo/tests/test_middleware.py
@@ -66,6 +66,18 @@ class TestMiddleware(TestCase):
         AuthenticationMiddlewareWithoutAPI().process_request(req)
         assert process_request.call_count == 3
 
+    def test_lbheartbeat_middleware(self):
+        # /__lbheartbeat__ should be return a 200 from middleware, bypassing later
+        # middleware and view code.
+
+        with self.assertNumQueries(0):
+            response = test.Client().get('/__lbheartbeat__', SERVER_NAME='elb-internal')
+        assert response.status_code == 200
+        assert response.content == b''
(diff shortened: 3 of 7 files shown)


eviljeff:
Note for these urls to work locally they either need adding to https://github.com/mozilla/addons-frontend/blob/master/config/default.js or https://github.com/mozilla/addons-nginx/blob/master/addons.conf (I did the latter locally).  That's probably not necessary in general (they're useless locally) and if we keep services/monitor.json around, redundant.

bqbn:
> fixes #18510
> 
> * repurposes `/services/monitor.json` as `/__heartbeat__` (but leaves the url alone as an alias for now)

I like the "alias" idea too. It provides flexibility while we transition monitoring to the other endpoint. We can remove it later.

bqbn:
The load balancer is not able to send custom headers to the application, in particular, for value of the `HOST` header, it'll use the instance's internal IP. I'm not sure if you can make the the application recognize that.

Here is an example of the headers the ELB will send,

```
GET /__version__ HTTP/1.1
host: 172.31.14.247:4000
User-Agent: ELB-HealthChecker/1.0
Accept: */*
Connection: keep-alive
```

The above example currently works well with addons-frontend, but Django may be different.

eviljeff:
> The load balancer is not able to send custom headers to the application, in particular, for value of the `HOST` header, it'll use the instance's internal IP. I'm not sure if you can make the the application recognize that.
> 
> Here is an example of the headers the ELB will send,
> 
> ```
> GET /__version__ HTTP/1.1
> host: 172.31.14.247:4000
> User-Agent: ELB-HealthChecker/1.0
> Accept: */*
> Connection: keep-alive
> ```

I'm not sure what you're asking/saying here @bqbn - you _want_ `/__heartbeat__` to take the host/headers into account?  Or you think (or know) that the checks taking place will fail or be inaccurate because of the host will be internal IP?

bqbn:
> I'm not sure what you're asking/saying here @bqbn - you _want_ `/__heartbeat__` to take the host/headers into account? Or you think (or know) that the checks taking place will fail or be inaccurate because of the host will be internal IP?

I see the following error when mimicking the ELB checking the health check endpoint (I used `/services/monito.json` in this case).

```
Dec 16 17:04:31 ip-172-31-26-77.ec2.internal docker-addons[15414]: {"Timestamp": 1639674271885841152, "Type": "django.security.DisallowedHost", "Logger": "http_app_addons_dev", "Hostname": "ip-172-31-26-77.ec2.internal", "EnvVersion": "2.0", "Severity": 3, "Pid": 9, "Fields": {"status_code": 400, "request": "<WSGIRequest: HEAD '/services/monitor.json'>", "uid": "", "remoteAddressChain": "", "msg": "Invalid HTTP_HOST header: '172.31.26.77'. You may need to add '172.31.26.77' to ALLOWED_HOSTS."}, "severity": 500}
```

bobsilverberg (approved):
LGTM

eviljeff:
> > I'm not sure what you're asking/saying here @bqbn - you _want_ `/__heartbeat__` to take the host/headers into account? Or you think (or know) that the checks taking place will fail or be inaccurate because of the host will be internal IP?
> 
> I see the following error when mimicking the ELB checking the health check endpoint (I used `/services/monito.json` in this case).
> 
> ```
> Dec 16 17:04:31 ip-172-31-26-77.ec2.internal docker-addons[15414]: {"Timestamp": 1639674271885841152, "Type": "django.security.DisallowedHost", "Logger": "http_app_addons_dev", "Hostname": "ip-172-31-26-77.ec2.internal", "EnvVersion": "2.0", "Severity": 3, "Pid": 9, "Fields": {"status_code": 400, "request": "<WSGIRequest: HEAD '/services/monitor.json'>", "uid": "", "remoteAddressChain": "", "msg": "Invalid HTTP_HOST header: '172.31.26.77'. You may need to add '172.31.26.77' to ALLOWED_HOSTS."}, "severity": 500}
> ```

Ah, I feared it might be something like that.  [`settings.ALLOWED_HOSTS`](https://docs.djangoproject.com/en/3.2/ref/settings/#allowed-hosts) is what hosts are allowed to make a request and it's checked in the base `HttpRequest` class, then enforced in django middleware (`CommonMiddleware` makes a `request.get_host()` call, which [triggers](https://github.com/django/django/blob/3.2.10/django/middleware/common.py#L48) the check).  

I.e. we can't simply add an `if ...` in our code - we either need to:
- override the `ALLOWED_HOSTS` setting for certain defined paths (hacky, and no support for temporarily overriding settings, outside of tests)
- monkey-patch `HttpRequest.get_host()` (or `django.http.request.validate_host`) to not raise for certain defined paths (even more hacky)
- drop `ALLOWED_HOSTS` setting  (set it to `['*']`); define our own `ALLOWED_HOSTS_ISH` setting; then create our own middleware that enforces that hosts are in `ALLOWED_HOSTS_ISH`, apart from certain defined paths.

bqbn:
For the `__lbheartbeat__` endpoint only, can we add a health check middleware early in the middleware chain? What's suggested in https://stackoverflow.com/a/64623669 seems to be a reasonable solution IMHO.

For the `__heartbeat__` endpoint, it can be treated as any other endpoints because ELB won't need to access it.

eviljeff:
@diox I implemented @bqbn's suggestion - does this seem ... okay to you?  (we don't listen for random paths in other middleware and return a response, to my knowledge)

diox:
Not ideal, but that works. It's worth noting though, that python-dockerflow already provides that functionality through its `DockerflowMiddleware`. We should look into what it would take to use it instead, but it can be a separate issue - maybe even a good first bug. (Among other things it also implements `__version__`, so we would need to customize `DOCKERFLOW_VERSION_CALLBACK`, since our implementation is a little custom)

eviljeff:
> Not ideal, but that works. It's worth noting though, that python-dockerflow already provides that functionality through its `DockerflowMiddleware`. We should look into what it would take to use it instead, but it can be a separate issue

agreed

> maybe even a good first bug.

Hah, hacking around in middleware wouldn't be my idea of a good first bug 

> (Among other things it also implements `__version__`, so we would need to customize `DOCKERFLOW_VERSION_CALLBACK`, since our implementation is a little custom)

Also our __heartbeat__ does different things and we may want to keep some of that.

PR #18520, opened 2021-12-17
Only log when source has been submitted

### LINKED ISSUE #18489: Update source submit logging to only log when source is actually being submitted
This is a follow-up to https://github.com/mozilla/addons-server/issues/18412#issuecomment-989873194.

We should only log, in both `_submit_source` and `version_edit`, when source is actually being uploaded.

bobsilverberg:
Fixes #18489


bobsilverberg:
Requesting review from Mat (in the new year) so that he can verify whether this new logging scheme is what he is looking for. I had to change it around a bit.

diox (approved):

Issue #18556, opened 2022-01-03
Prevent session cookie from being sent when it's not needed


diox:
2 things are causing us to return a `Set-cookie: sessionid=...` on almost every anonymous template-based response in addons-server:
1. The register and login links for FxA, on pages like the developer hub landing page, which add a `fxa_state` in the session. Modifying the session causes it to be created/saved and the cookie is sent as a result.
2. The CSRF handling. When a csrf token is used, django flips a flag in the request  (`META['CSRF_COOKIE_USED']`) which is then used by `CsrfViewMiddleware.get_response()`
  to return the cookie or add it to the session (when `CSRF_USE_SESSIONS` is `True` as it's
  the case with our setup), which in turns causes the session cookie to be sent as above.

For 1) the solution is to link to `/api/authenticate/start` or whatever it's called instead of having the true links to FxA (which contain the state parameter) in our HTML.

For 2) we need to find why Django thinks `csrf_token` is being used. On most pages not behind auth it shouldn't be the case, Django's default context processor that adds csrf_token to template contexts is supposed to be lazy, yet I'm seeing this even on the `robots.txt` page. Maybe we are bypassing that somehow ? Could be related to our Jinja2 setup.

diox:
2) is a red-herring, it's caused by... Django Debug Toolbar, only locally.

diox:
QA:
- Once the fix for this lands, going to `/en-US/developers/` anonymously without a `sessionid` cookie should not create one.
- Logging in from that page or elsewhere should still work and redirect you to where you where when you clicked on the link to log in

AlexandraMoga:
@diox I've verified this issue on -dev an these are my results:

- open https://addons-dev.allizom.org/en-US/firefox/ while being logged out =>  `sessionid` cookie not set
- open https://addons-dev.allizom.org/en-US/developers/ directly while being logged out => `sessionid` cookie not set
- use client side navigation to open various pages on the site while being logged out => `sessionid` cookie not set
- logging in in from any page on the site will set the `sessionid` cookie and users will be redirected back to the page where they started the log in process
- the  `sessionid` cookie will remain active on the site until the user logs out, at which point the cookie will be deleted

There weren't any other issues identified concerning the login functionalities.

PR #18557, opened 2022-01-03
Don't return direct link to FxA for login/register, use our API endpoint

diox:
This allows us to avoid modifying the session on pages that just want to display a login/register link, which in turn avoids creating a new session and potentially returning a session cookie to anonymous users.

Fixes https://github.com/mozilla/addons-server/issues/18556

### DIFF
--- a/src/olympia/accounts/templatetags/jinja_helpers.py
+++ b/src/olympia/accounts/templatetags/jinja_helpers.py
@@ -1,3 +1,5 @@
+from urllib.parse import quote_plus
+
 from django_jinja import library
 import jinja2
 
@@ -10,7 +12,7 @@ from olympia.amo.templatetags.jinja_helpers import drf_url
 def login_link(context):
     next_path = path_with_query(context['request'])
     login_start = drf_url(context, 'accounts.login_start')
-    return f'{login_start}?to={next_path}'
+    return f'{login_start}?to={quote_plus(next_path)}'
 
 
 @library.global_function
--- a/src/olympia/accounts/templatetags/jinja_helpers.py
+++ b/src/olympia/accounts/templatetags/jinja_helpers.py
@@ -13,9 +13,3 @@ def login_link(context):
     next_path = path_with_query(context['request'])
     login_start = drf_url(context, 'accounts.login_start')
     return f'{login_start}?to={quote_plus(next_path)}'
-
-
-@library.global_function
-@jinja2.pass_context
-def register_link(context):
-    return login_link(context)
--- a/src/olympia/accounts/templatetags/jinja_helpers.py
+++ b/src/olympia/accounts/templatetags/jinja_helpers.py
@@ -1,23 +1,19 @@
+from django_jinja import library
 import jinja2
 
-from django_jinja import library
-
-from olympia.accounts import utils
+from olympia.accounts.utils import path_with_query
+from olympia.amo.templatetags.jinja_helpers import drf_url
 
 
 @library.global_function
 @jinja2.pass_context
 def login_link(context):
-    return utils.default_fxa_login_url(context['request'])
+    next_path = path_with_query(context['request'])
+    login_start = drf_url(context, 'accounts.login_start')
+    return f'{login_start}?to={next_path}'
 
 
 @library.global_function
 @jinja2.pass_context
 def register_link(context):
-    return utils.default_fxa_register_url(context['request'])
-
-
-@library.global_function
-@jinja2.pass_context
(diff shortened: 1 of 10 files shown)


src/olympia/accounts/templatetags/jinja_helpers.py
```diff
@@ -1,23 +1,21 @@
...
 @jinja2.pass_context
 def register_link(context):
-    return utils.default_fxa_register_url(context['request'])
-
-
-@library.global_function
-@jinja2.pass_context
-def fxa_config(context):
-    return utils.fxa_config(context['request'])
+    return login_link(context)
```
bobsilverberg, line 21: If `register_link` just returns `login_link` now, do we need `register_link` at all? Could we just call `login_link` from anywhere we're currently calling `register_link`? If so, should it perhaps be called `fxa_link`?

src/olympia/accounts/tests/test_helpers.py
```diff
@@ -1,23 +1,31 @@
...
 
-@mock.patch(
-    'olympia.accounts.templatetags.jinja_helpers.utils.default_fxa_login_url',
-    lambda c: 'http://auth.ca',
-)
 def test_login_link():
     request = RequestFactory().get('/en-US/firefox/addons')
-    assert jinja_helpers.login_link({'request': request}) == ('http://auth.ca')
+    assert jinja_helpers.login_link({'request': request}) == (
+        'http://testserver/api/v5/accounts/login/start/'
```
bobsilverberg, line 9: Would it be worth putting `'http://testserver/api/v5/accounts/login/start/'` into a constant as it's used in each of the three assertions below? Also, how do we know that is the correct URL to expect? Can that be imported from elsewhere to be sure it's the correct value?

src/olympia/accounts/utils.py
```diff
@@ -18,28 +18,6 @@
 from .tasks import clear_sessions_event, delete_user_event, primary_email_change_event
 
 
-def fxa_config(request):
```
bobsilverberg, line 21: Just curious, with the removal of this, what was it ever used for?

src/olympia/amo/decorators.py
```diff
@@ -181,22 +181,3 @@ def wrapper(request, *args, **kw):
         return response
 
     return wrapper
-
-
-def allow_mine(f):
```
bobsilverberg, line 186: What does this have to do with this patch?

bobsilverberg (approved):
Looks good. I just have a few questions and a suggestion.

r+wc

src/olympia/amo/decorators.py
diox: Just a drive-by cleanup while I was looking at this file for stuff to change around auth

src/olympia/accounts/utils.py
diox: It was used in the `fxa_config` helper which was called from the base template to add a `data-` attribute containing that info. That would presumably have been useful a long time ago with the legacy site.

src/olympia/accounts/templatetags/jinja_helpers.py
diox: We technically don't need it anymore, I wasn't sure if we'd need the distinction. I guess I can remove it and rename the other.

src/olympia/accounts/tests/test_helpers.py
diox: If you don't hardcode the expected result, then you end up re-creating the logic from the function being tested (which is super short - it's just concatenating the result of `drf_url()` with `path_with_query()` encoded).

I could add a constant, but I'm not super convinced it's necessary since it's the same block of code and it's in a test.

bobsilverberg: > If you don't hardcode the expected result, then you end up re-creating the logic from the function being tested (which is super short - it's just concatenating the result of `drf_url()` with `path_with_query()` encoded).
> 
> I could add a constant, but I'm not super convinced it's necessary since it's the same block of code and it's in a test.

Fair enough. We tend to use constants whenever possible in tests for frontend, but I know that's not the case for server, so this is consistent with the way a lot of other tests are written.

src/olympia/accounts/templatetags/jinja_helpers.py
diox: 450ea6c7e8870a81750ccd50e94849541d96634c (didn't end up doing the renaming, I don't think it's that much clearer)

PR #18562, opened 2022-01-04
Prevent site permission add-ons from being edited except for deletion/ownership

### LINKED ISSUE #18444: Prevent developers from editing site permission add-ons in devhub
Site Permission add-ons from #18135 are automatically generated and should be visible in devhub, but the only thing you should be able to do is add/change/remove authors.

They shouldn't be allowed to be uploaded by developers either, unless they have a specific permission.

diox:
Includes a renaming of the arguments for `check_addon_ownership()` to make it more clear and consistent, and an extra `@readonly_if_site_permission` added to all pages we don't want to be editable in devhub for site permission add-ons.

Fixes #18444


diox:
@eviljeff This is missing tests for the stuff I've added but can you give me your opinion on the approach ?

src/olympia/reviewers/views.py
```diff
@@ -1140,7 +1140,7 @@ def download_git_stored_file(request, version_id, filename):
         raise http.Http404
 
     if version.channel == amo.RELEASE_CHANNEL_LISTED:
-        is_owner = acl.check_addon_ownership(request, addon, dev=True)
+        is_owner = acl.check_addon_ownership(request, addon, allow_developer=True)
```
eviljeff, line 1143: do we want to allow_site_permission here too?

src/olympia/versions/views.py
```diff
@@ -99,7 +99,9 @@ def is_appropriate_reviewer(addon, channel):
         use_cdn = False
         has_permission = is_appropriate_reviewer(
             addon, channel
-        ) or acl.check_addon_ownership(request, addon, dev=True, ignore_disabled=True)
+        ) or acl.check_addon_ownership(
+            request, addon, allow_developer=True, allow_mozilla_disabled_addon=True
```
eviljeff, line 103: do we want to allow_site_permission here too?

src/olympia/devhub/decorators.py
```diff
@@ -56,16 +56,28 @@ def fun():
...
                     if not submitting and addon.should_redirect_to_submit_flow():
                         return redirect('devhub.submit.details', addon.slug)
                     return fun()
-            # Require an owner or dev for POST requests (if the add-on status
-            # is disabled that check will return False).
+            # Require an owner or deveveloper for POST requests (if the add-on
+            # status is disabled that check will return False). Allow site
+            # permission add-ons, an extra check will be made for pages that
+            # don't make sense if the add-on is a site permission add-on
+            # through the @readonly_if_site_permission decorator below.
```
eviljeff, line 73: I don't like the way `@dev_required` now defaults to allowing writes to site_permission add-ons, and you have to remember to add a second decorator to make it safe - the ideal situation is it defaults to preventing writes to those add-ons.

IMO add a `allow_site_permission=False` kwarg to the decorator and pass that here.

src/olympia/devhub/decorators.py
```diff
@@ -80,6 +92,30 @@ def fun():
         return decorator
 
 
+def readonly_if_site_permission(f):
+    """If the add-on is a site permission add-on, only allow readonly requests.
+
+    Add this decorator on top of dev_required() (which would normally allow
+    editing site permission add-ons) to disallow edition of specific devhub
+    pages for site permission add-ons."""
```
eviljeff, line 100: maybe we still need this for some views?  But I don't think we should require it to be used too for any `@dev_required` decorated view.

eviljeff:
> @eviljeff This is missing tests for the stuff I've added but can you give me your opinion on the approach ?

The overall approach seems good, but I don't like the having to always use the new decorator with `@dev_required` in all those devhub views.  An extra kwarg for `dev_required` would make it safer by default.

src/olympia/reviewers/views.py
diox: I don't think so, since they'll be auto-generated it's not really useful to look at them in code manager (I'm not even convinced we need to allow add-on developers to access the code manager APIs, btw...)

src/olympia/addons/decorators.py
```diff
@@ -7,14 +7,17 @@
 from olympia.addons.models import Addon
 
 
+# FIXME: name is incorrect, since it allows developers and not just owners.
```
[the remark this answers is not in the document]
diox: Yeah, something like that. I'll rename and move.

src/olympia/versions/views.py
diox: Yes, we do - I'll change it (They will be pure unlisted add-ons, we'll want developers to be able to download them).

src/olympia/devhub/decorators.py
diox: That's one of the reasons I went with the separate decorator route, but actually, that turned out to be false. So yeah, I agree, no good reason for a separate decorator.

diox:
Thanks for the feedback, I agree the separate decorator is probably ill-advised. I'll try the separate keyword argument to `dev_required` approach.

src/olympia/devhub/templates/devhub/addons/owner.html
```diff
@@ -8,11 +8,6 @@
 {{ dev_page_title(title, addon) }}
 {% endblock %}
 
-{% block bodyclass %}
-{{ super() }}
-{% if not check_addon_ownership(request, addon) %} no-edit{% endif %}
-{% endblock %}
```
diox, line 14: This is done in the view now, and thanks to the other change to `devhub/base.html`, the check is only done once.

src/olympia/devhub/tests/test_views_edit.py
```diff
@@ -742,6 +742,17 @@ class TestEditDescribeUnlisted(BaseTestEditDescribe, L10nTestsMixin):
     listed = False
     __test__ = True
 
+    def test_no_edit_if_disabled(self):
+        pass
```
eviljeff, line 746: hmm?   Why is this overridden?

src/olympia/devhub/tests/test_views_edit.py
```diff
@@ -1472,25 +1491,38 @@ class TestEditTechnicalUnlisted(BaseTestEdit):
...
         addon = self.get_addon()
         assert addon.whiteboard.public == ''
 
+    def test_site_permission(self):
+        self.addon.update(type=amo.ADDON_SITE_PERMISSION)
+        response = self.client.get(self.technical_url)
+        assert response.status_code == 200
+
+        response = self.client.post(self.technical_edit_url, {})
+        assert response.status_code == 403
```
eviljeff, line 1524: I guess the whiteboard is redundant when there aren't any reviews, but it's not really altering metadata about the site permission add-on so it'd be fine (expected) that whiteboard would still be available.

src/olympia/devhub/tests/test_views_submit.py
```diff
@@ -1900,6 +1900,13 @@ def test_has_read_agreement(self):
             response, reverse('devhub.submit.version.agreement', args=[self.addon.slug])
         )
 
+    def test_site_permission_not_allowed(self):
+        self.addon.update(type=amo.ADDON_SITE_PERMISSION)
+        response = self.client.get(self.url)
+        assert response.status_code == 403
+        response = self.client.post(self.url)
+        assert response.status_code == 403
```
eviljeff, line 1908: nit: a mixin for this?  Seems a bit un-DRY-y to have the same chunk of code 4 times.

src/olympia/devhub/templates/devhub/base.html
```diff
@@ -7,9 +7,9 @@
 {% endblock %}
 
 {% if addon %}
-  {% set editable_body_class = "no-edit" if not check_addon_ownership(request, addon, dev=True) %}
+  {% set default_body_class = "no-edit" if not check_addon_ownership(request, addon, allow_developer=True) %}
```
eviljeff, line 10: we should probably make this change in base_impala.html too, for consistency.

eviljeff (approved):

src/olympia/devhub/tests/test_views_edit.py
diox: mmm I think it's a test I noticed could be useful while I was looking at those tests, I intended to implement it but forgot. I'll look into it.

src/olympia/devhub/tests/test_views_edit.py
diox: It's a bit annoying to allow though, cause it goes through the generic `addons_section` view. I don't think it's worth it for the moment.

src/olympia/devhub/tests/test_views_edit.py
diox: I moved the test up in d2f1a0cbc546d2a3676fa050732393b1746044c8 and clarified the comment in 135528f68f19450e61d49a9cf2fba6f5e9b21cb2. Now I remember why I added that test: the behavior is slightly surprising. If the add-on is pure unlisted, mozilla-disabling it completely hides it from developers, they get a 404. If it had listed versions, they can see it (but of course can't edit it).

This is because of https://github.com/diox/olympia/blob/d2f1a0cbc546d2a3676fa050732393b1746044c8/src/olympia/addons/decorators.py#L35-L44 : they fail the `has_listed_version` part, but *also* the `acl.author_or_unlisted_viewer_or_reviewer()`, despite being an author, because inside that function, we call `check_addon_ownership()` without `allow_mozilla_disabled_addon=True`, so despite being an author, they are denied access.

As far as I can tell, this is intentional, just maybe a little bit surprising.

PR #18563, opened 2022-01-04
Stop exposing reviewer names to developers in review emails

### LINKED ISSUE #18565: Stop exposing reviewer names to developers in review emails
This is a spin-off from #18564 and is just about removing the names of reviewers from emails sent specifically about reviews. The templates for these emails can be found in `src/olympia/reviewers/templates/reviewers/emails/`. Code used to send emails about activity log entries, as well as code that exposes reviewer names on devhub pages will be dealt with in a separate issue.

bobsilverberg:
Fixes #18565

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/emails/extension_nominated_to_rejected.ltxt
+++ b/src/olympia/reviewers/templates/reviewers/emails/extension_nominated_to_rejected.ltxt
@@ -1,9 +1,6 @@
 {% extends "reviewers/emails/base.ltxt" %}{% block content %}
 Your add-on, {{ name }} {{ number }}, has been reviewed and did not meet the criteria for being hosted in our gallery.
 
-Reviewer:
-{{ reviewer }}
-
 Comments:
 {{ comments }}
 
--- a/src/olympia/reviewers/templates/reviewers/emails/extension_pending_to_rejected.ltxt
+++ b/src/olympia/reviewers/templates/reviewers/emails/extension_pending_to_rejected.ltxt
@@ -1,8 +1,6 @@
 {% extends "reviewers/emails/base.ltxt" %}{% block content %}
 Your add-on, {{ name }} {{ number }}, has been reviewed and did not meet the criteria for being hosted in our gallery.
 {% include "reviewers/emails/files.ltxt" %}
-Reviewer:
-{{ reviewer }}
 
 Comments:
 {{ comments }}
--- a/src/olympia/reviewers/templates/reviewers/emails/reject_multiple_versions.ltxt
+++ b/src/olympia/reviewers/templates/reviewers/emails/reject_multiple_versions.ltxt
@@ -7,10 +7,6 @@ Details:
 
 Version(s) affected and disabled:
 {{ version_numbers }}
-
-Reviewer:
-{{ reviewer }}
-
 ********
 
 Please address the issues raised in the reviewer's notes and inquire about any unclear items. Afterwards, please upload a new version of your add-on at {{ dev_versions_url }}.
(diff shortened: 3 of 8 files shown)


bobsilverberg:
> One of the main problems with the original PR was figuring out what to do with the activity logs, that can appear in various places and might need the author information depending on who is looking at it. So we need to do a little digging for that.

I saw that, and as I was looking at the code base regarding this change (to just stop exposing reviewer names), I noticed exactly that - the fact that we are also exposing them with activity notifications. I was going to include some changes to those templates as well in the final version of this PR. 

I guess it's a question of seeing whether we can identify templates where we can remove the reviewer/note author entirely, vs. those where we need to keep the reviewer/note author for cases where the communication is not with an add-on developer. Does that summarize the issue in the context of this PR (which is only about stopping exposing the info, and not about removing the `reviewer_name` field)?

bobsilverberg:
As a side note, I was surprised to see that these changes resulted in no test failures whatsoever. That makes me think that none of the code I changed is covered by tests (including the `get_context_data` method). I certainly wasn't able to find any tests for it when I looked, but I assumed that running the full suite in CI would result in some failures. Is this expected, or are we in fact missing some coverage that should be there?

diox:
> I guess it's a question of seeing whether we can identify templates where we can remove the reviewer/note author entirely, vs. those where we need to keep the reviewer/note author for cases where the communication is not with an add-on developer. Does that summarize the issue in the context of this PR 

Yes. In some cases it might be tricky - activity log code is fairly generic, and as noted in the original PR we want to avoid scaling issues with database queries.

> As a side note, I was surprised to see that these changes resulted in no test failures whatsoever. That makes me think that none of the code I changed is covered by tests (including the `get_context_data` method). I certainly wasn't able to find any tests for it when I looked, but I assumed that running the full suite in CI would result in some failures. Is this expected, or are we in fact missing some coverage that should be there?

This should be under coverage but we're probably missing some specific assertions - we do have reviewer tests checking that the email is sent, with some checks against the content, just not the reviewer name I would imagine.

bobsilverberg:
> > I guess it's a question of seeing whether we can identify templates where we can remove the reviewer/note author entirely, vs. those where we need to keep the reviewer/note author for cases where the communication is not with an add-on developer. Does that summarize the issue in the context of this PR
> 
> Yes. In some cases it might be tricky - activity log code is fairly generic, and as noted in the original PR we want to avoid scaling issues with database queries.
> 

Right. I'll start looking at the activity log code and see if I can figure out the different cases. One thing I am assuming, and maybe this is an incorrect assumption; if it's exposed in our tools (e.g., the reviewer tools screens), then it's okay, because those are all "trusted" users. This leads me to a second assumption that is that we only need to worry about exposing the names in emails, as that is the only way we currently expose this information to "untrusted" users. But, having said that, I realize I am probably wrong - there are probably places in devhub where we expose information about add-on reviews to the add-on author, and perhaps that information includes the reviewer's name. Is that correct? If so, then I guess we need to remove/anonymize the reviewer names in those places too.

bobsilverberg:
Thanks for all the feedback @diox and @wagnerand. I think to simplify things I'll break this down into even smaller pieces. That way we can land the changes just to the email templates in the `reviewers` app, and save the more complicated stuff related to activity log and devhub for a different issue. I'll open a new issue and attach this patch to it.

eviljeff (approved):
r+ for these minimal changes to the email templates.

bobsilverberg:
> r+ for these minimal changes to the email templates.

Thanks Andrew. I think it will be easier to address the whole issue eventually with smaller patches like this one.

PR #18581, opened 2022-01-07
update our Storage class and django 3.2.11

### LINKED ISSUE #18570: Update LocalFileStorage to work with django 3.2.11
https://www.djangoproject.com/weblog/2022/jan/04/security-releases/ includes a fix (https://github.com/django/django/commit/6d343d01c57eb03ca1c6826318b652709e58a76e) for directory traversal.  

Our LocalFileStorage does a few things that conflict with it:
- uses bytestrings rather the str for paths
- allows us to access paths other than under `MEDIA_ROOT`

eviljeff:
fixes #18570 

I originally intended to change all the references to `default_storage` to a specific version locked down to the path we wanted but it turned out to be difficult in many cases - most of the time we use storage we're moving files around or reading/writing to temp and another location, and `STORAGE_ROOT` ends up being the only path that both locations are a subset of (see blocklist and user profile pics for examples where it did work as I wanted).  There may be some cases I missed where we could also restrict the location, at least in some methods.

Test classes needed even less restriction where they were reading from fixtures (under `/code`) so `root_storage` effectively gives tests the same freedom to read files from wherever they had before.


src/olympia/amo/tests/__init__.py
```diff
@@ -91,6 +91,11 @@
 SQUOTE_ESCAPED = escape("'")
 
 
+# A Storage instance for the filesystem root to be used during tests that read fixtures
+# and/or try to copy them under settings.STORAGE_ROOT.
+root_storage = SafeStorage(location='/')
```
bobsilverberg, line 96: Just curious, is there a reason this needs to be set here, as opposed to just doing this at line 469?

src/olympia/devhub/tests/test_views.py
```diff
@@ -1899,7 +1898,7 @@ def test_header_image(self):
         zip_file = os.path.join(
             settings.ROOT, 'src/olympia/devhub/tests/addons/static_theme.zip'
         )
-        copy_stored_file(zip_file, destination)
+        self.root_storage.copy_stored_file(zip_file, destination)
```
bobsilverberg, line 1901: I was confused about why sometimes you import `root_storage`, and other times you use `self.root_storage`. I can see that when a test class inherits from `TestCase`, then you can use `self.root_storage`, but why do some of our tests live in classes that inherit from `TestCase` while others do not?

src/olympia/files/tasks.py
```diff
@@ -5,13 +5,15 @@
 
 from pathlib import Path
 
+from django.conf import settings
+from django.core.files.storage import default_storage as storage
```
bobsilverberg, line 9: Why does this one use `default_storage` instead of `SafeStorage`? I notice this in a few files and I don't quite understand it.

src/olympia/hero/models.py
```diff
@@ -115,15 +123,15 @@ def __str__(self):
...
-        return path + b'/thumbs/' + fn
+        return path + '/thumbs/' + fn
 
     @property
     def image_url(self):
         return f'{self.custom_image.url}?modified={int(self.modified.timestamp())}'
 
     @property
     def preview_url(self):
-        fn = os.path.basename(self.custom_image.path).decode('utf-8')
```
bobsilverberg, line 126: Why did we remove the `decode`?

src/olympia/users/tests/test_models.py
```diff
@@ -85,6 +85,7 @@ def test_delete(self):
         user = UserProfile.objects.get(pk=4043307)
 
         # Create a photo so that we can test deletion.
+        storage = SafeStorage(user_media='userpics')
```
bobsilverberg, line 88: Could we add this into a `setUp` as it seems to be added to a number of tests?

bobsilverberg (approved):
Looks good to me. I have a few questions, mostly just so I can understand the code better.

r+wc

src/olympia/amo/tests/__init__.py
bobsilverberg: I think I answered my own question after reading more of the code. Sometimes you want to import this, and sometimes it's available in `TestCase`. I did ask a different, related question below.

eviljeff: yep, it's just for the two different use cases.  There's technically nothing stopping us from always importing it, or accessing it as `TestCase.root_storage` outside of a `TestCase` subclass, but the former introduced more unnecessary `import`s and the latter is a little abnormal in our tests (though acceptable if it was, like, one occurrence)

src/olympia/devhub/tests/test_views.py
eviljeff: No particular reason, it's just the two ways of writing python tests.  Generally if you need a single test for a stand-alone function, a function based test is fine; if there are multiple tests needed, or you need some of the extra setup/functionality available within a `TestClass`, (e.g. you have to mark each function based test as needing database access or it's blocked), then subclass.  Conversely there are a few things that are difficult to do inside classes, like `@pytest.mark.parametrize`

eviljeff: Oh, so after re-reading your question it seems maybe you're talking about tests that live in classes that inherit from `TestCase` vs. _tests that live in classes that don't inherit from `TestCase`_.  If so, can you point them out because I don't recall them.

src/olympia/files/tasks.py
eviljeff: django provides a `default_storage` instance that uses `settings.DEFAULT_FILE_STORAGE` - that's the normal way (it's not initialized with any args so it ends up with a base_location of `settings.ROOT_STORAGE`).  If we want a non-standard storage location we can create our own instance.

src/olympia/hero/models.py
eviljeff: `.path` was a bytestring previously, so needed to be decoded into a unicode string so it could be used in the `f` string.  Now all our paths are unicode strings.

diox (approved):

PR #18597, opened 2022-01-11
Add form to generate site permission add-ons

diox:
Fixes #18136

Note that there are no links pointing to that page yet. This is by design, we don't want to officially launch it yet (and even when we do, there might not be many links pointing to it from AMO, it could be from an external page like MDN instead).


src/olympia/devhub/forms.py
```diff
@@ -1396,3 +1396,38 @@ def save(self):
...
+        )
+        try:
+            parsed = urlparse(value)
+        except ValueError:
+            raise forms.ValidationError(error_message)
+        if (
+            not parsed.scheme
+            or parsed.scheme not in ('https', 'http')
+            or not parsed.netloc
+            or not actual_value.startswith(parsed.scheme)
```
bobsilverberg, line 1426: I don't understand this one. If we accept the value, we'll be using `cleaned_data`, will we not? Why are we concerned with checking `actual_value`?

src/olympia/devhub/tasks.py
```diff
@@ -742,3 +744,20 @@ def send_api_key_revocation_email(emails):
         use_deny_list=False,
         perm_setting='individual_contact',
     )
+
+
+@task
+@use_primary_db
+def create_site_permission_version(
```
bobsilverberg, line 751: Why are we using a task for this, as opposed to just doing it synchronously when the form is submitted? Is it because we have to do signing asynchronously?

src/olympia/devhub/tests/test_tasks.py
```diff
@@ -886,3 +886,27 @@ def test_returns_received_results(self):
...
+    def test_pass_down_to_creator_util(self, SitePermissionVersionCreator_mock):
+        user = user_factory()
+        tasks.create_site_permission_version(
+            user_pk=user.pk,
+            remote_addr='127.0.0.42',
+            install_origins=['https://example.com'],
+            site_permissions=['foo'],
+        )
+        assert SitePermissionVersionCreator_mock.call_count == 1
+        assert SitePermissionVersionCreator_mock.call_args[0] == ()
```
bobsilverberg, line 902: What is this? Is the the `self` that is always passed into functions?

bobsilverberg (approved):
Looks good, I just have a few questions.

r+wc

src/olympia/devhub/forms.py
diox: I'll clarify in a comment. Django's `URLField` accepts `foo` as actual value, translating it to `http://foo` in the cleaned value. So I have to check the actual value here.

src/olympia/devhub/tasks.py
diox: signing would take place in `auto_approve` asynchronously no matter what - it will never be immediate.

The reason why I switched to a task was because behind the scenes I'm reusing the regular primitives we se for uploads, the zip is created the zip in memory, written to the temporary location for file uploads, then to the final location, and there are a few database calls in between to set up everything. All trivial operations, but it can add up under load and I don't want the http servers to wait for that.

All in all it's consistent with everything else in addons-server: things that can take time/resources in a significant way are in a task or cronjob.

src/olympia/devhub/tests/test_tasks.py
diox: Regular (non-keyword) arguments. There are none, everything is passed as keyword arguments.

src/olympia/devhub/forms.py
```diff
@@ -1396,3 +1396,38 @@ def save(self):
...
+    site_permissions = forms.MultipleChoiceField(
+        label=_('Permissions'), choices=(('midi-sysex', 'WebMIDI'),)
+    )
+
+    def clean_origin(self):
+        actual_value = str(self.data.get('origin'))
+        value = self.cleaned_data.get('origin')
+        # Note that URLField should already ensure it's an URL.
+        error_message = _(
+            'Origin should be made of a scheme (protocol), a hostname (domain) and an '
```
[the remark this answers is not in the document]
diox: Yes, but I'm not a native speaker and I thought the sentence looked a bit weird with `only` added. Where would you add the `only` in the sentence ?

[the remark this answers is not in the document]
bobsilverberg: Here are a couple of suggestions:

- "Origin should be exclusively made up of a scheme (protocol), a hostname (domain) and an optional port"
- "Origin should include only a scheme (protocol), a hostname (domain) and an optional port"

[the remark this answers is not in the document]
diox: c9c19c79b570f741f0dabd756bfa82c07089ffec

PR #18620, opened 2022-01-14
Move CREATE_ADDON activity log to always log addon creation regardless of source

### LINKED ISSUE #14190: No activity log for add-on creation when submitting from the API
Follow up for https://github.com/mozilla/addons-server/issues/14039#issuecomment-624653339

1. Submit an add-on using the API
2. Look for `ActivityLog` related to this new submission

### What happened?
You can see an `ActivityLog` for the `{id:username} | {version} added to the {addon}` but not one for the add-on creation itself

An `ActivityLog` of the format `{id:username} | {addon-name} was created` should be created when submitting a new add-on from the API

diox:
Fixes #14190

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -892,6 +892,8 @@ class Addon(OnChangeMixin, ModelBase):
             )
         if user:
             AddonUser(addon=addon, user=user).save()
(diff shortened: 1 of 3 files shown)


eviljeff (approved):
Just think of the bright future we'll have when FutureDevhub™️ uses the api and there's only a single path for things to happen in.

PR #18621, opened 2022-01-14
Clean up wsgi.py

### LINKED ISSUE #18276: Clean up wsgi.py so that it's less custom
Our `wsgi.py` script contains this:
https://github.com/mozilla/addons-server/blob/ddffab7fe9ca0ad84f58eab6e055a9376ecd5257/src/olympia/wsgi.py#L28-L37

That looks completely obsolete. We can probably return `django_app` directly (unless we need `wsgi.loaded`?).

diox:
Fixes #18276


src/olympia/wsgi.py
```diff
@@ -10,27 +6,11 @@
...
 
-# This is what mod_wsgi runs.
-django_app = get_wsgi_application()
-
 
-# Normally we could let WSGIHandler run directly, but while we're dark
-# launching, we want to force the script name to be empty so we don't create
-# any /z links through reverse.  This fixes bug 554576.
+# This is referenced in docker/uwsgi.ini: module = olympia.wsgi:application
 def application(env, start_response):
```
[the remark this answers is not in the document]
diox: The early `activate()` call is one of addons-server oldest bits of code. It was added in https://github.com/jbalogh/zamboni/commit/9a2e5be8d8982d2beab99190030120d00b8f39ea and then tweaked several times. Back then this was with Apache and mod_wsgi, which we have replaced by nginx+uwsgi a long time ago.

Bottom line, I don't think that's useful these days. the only benefit I can see would be to ensure we have a language set in the thread super early before the first request goes through `LocaleAndAppURLMiddleware()`, but I can't see a benefit in doing that. Any queryset that includes translations is patched to depend on the current language, but by the time you're doing such queries, we should have gone through the middleware, or we're in trouble...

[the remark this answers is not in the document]
diox: I'm not brave enough to remove that call right now. It works with it, let's keep it for now.

src/olympia/wsgi.py
```diff
@@ -10,27 +6,11 @@
...
-    if 'HTTP_X_ZEUS_DL_PT' in env:
-        env['SCRIPT_URL'] = env['SCRIPT_NAME'] = ''
-    env['wsgi.loaded'] = wsgi_loaded
```
eviljeff, line 33: what about the `loaded` view that relies on this env value?  If we're removing the code that sets it should the view be removed at the same time?

src/olympia/wsgi.py
```diff
@@ -10,27 +6,11 @@
...
-    env['hostname'] = django.conf.settings.HOSTNAME
-    env['datetime'] = str(datetime.now())
-    return django_app(env, start_response)
+    return get_wsgi_application()(env, start_response)
```
eviljeff, line 16: I'm guessing/hoping there was no reason why `django_app` was declared outside of the function?

eviljeff (approved):

src/olympia/wsgi.py
diox: Wei confirmed that the view was useless but I forgot to remove it, I'll do that now.

src/olympia/wsgi.py
diox: mmm, just in case I'm going to replace that with `application = get_wsgi_application()` to be closer to django's default implementation.

diox: https://github.com/mozilla/addons-server/pull/18621/commits/f5814670990ce54609b41be74363cac740d52dfe

src/olympia/wsgi.py
diox: https://github.com/mozilla/addons-server/pull/18621/commits/136bec8a8552350a3179bd395e6793175dda9427

PR #18622, opened 2022-01-14
Stop exposing reviewer names to developers in activity emails

### LINKED ISSUE #18566: Stop exposing reviewer names to developers in activity emails
This is a spin-off from #18564 and is about removing and/or anonymizing the names of reviewers from emails sent to developers as a result of activities. The templates for these emails can be found in `src/olympia/activity/templates/activity/emails/`. Depending on how these templates are used, we may be able to remove the reviewer's name, or we may have to add logic to just anonymize the reviewer's name when the email is being sent to an add-on developer.

bobsilverberg:
Fixes #18566

### DIFF
--- a/src/olympia/activity/templates/activity/emails/from_reviewer.txt
+++ b/src/olympia/activity/templates/activity/emails/from_reviewer.txt
@@ -1,6 +1,6 @@
 Hello,
 
-{{ author }}, a reviewer at addons.mozilla.org, found an issue when reviewing version {{ number }} of the add-on {{ name }} and requests additional information. You are receiving this email because {{ email_reason }}.
+A reviewer at addons.mozilla.org, found an issue when reviewing version {{ number }} of the add-on {{ name }} and requests additional information. You are receiving this email because {{ email_reason }}.
 
 *********
 {{ author }} wrote:
--- a/src/olympia/activity/templates/activity/emails/from_reviewer.txt
+++ b/src/olympia/activity/templates/activity/emails/from_reviewer.txt
@@ -1,6 +1,6 @@
 Hello,
 
-A reviewer at addons.mozilla.org, found an issue when reviewing version {{ number }} of the add-on {{ name }} and requests additional information. You are receiving this email because {{ email_reason }}.
+A reviewer at addons.mozilla.org found an issue when reviewing version {{ number }} of the add-on {{ name }} and requests additional information. You are receiving this email because {{ email_reason }}.
 
 *********
 {{ author }} wrote:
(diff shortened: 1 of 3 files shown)


src/olympia/activity/tests/test_utils.py
```diff
@@ -303,7 +304,15 @@ def _recipients(self, email_mock):
             assert reply_to.endswith(settings.INBOUND_EMAIL_DOMAIN)
         return recipients
 
-    def _check_email(self, call, url, reason_text):
+    def _check_email(
+        self,
+        call,
+        url,
+        reason_text,
+        author,
```
diox, line 312: So many non-named arguments make the calls annoying to read. Could you use keyword-only arguments instead (at least for the last 3 you've added), and modify the calls accordingly ?

src/olympia/activity/utils.py
```diff
@@ -33,6 +33,9 @@
 ACTIVITY_MAIL_GROUP = 'Activity Mail CC'
 NOTIFICATIONS_FROM_EMAIL = 'notifications@%s' % settings.INBOUND_EMAIL_DOMAIN
 SOCKETLABS_SPAM_THRESHOLD = 10.0
+# Types of users who might be sending or receiving emails.
+ADDON_AUTHOR = 'An add-on author'
```
diox, line 37: I find it a little confusing that it's both a constant to identify what type of user it is and the string that will be used as the sender name... but only if they are a reviewer. I think I would rather have different constants: `ADDON_AUTHOR` and `ADDON_REVIEWER` integers to denote the type of user, then another would like `ADDON_REVIEWER_NAME` or something for the name string sent to developers.

What do you think?

src/olympia/activity/utils.py
```diff
@@ -297,6 +313,10 @@ def notify_about_activity_log(
             task_user = {get_task_user()}
         except UserProfile.DoesNotExist:
             task_user = set()
+        # Update the author and from_email to use the real name in all cases.
```
diox, line 316: Could you modify this comment to indicate why ? (i.e. because it's for the email sent to reviewers)

Also the comment a few lines below mentions building the context for reviewers but it's now built here so it needs to be adjusted as well.

src/olympia/activity/templates/activity/emails/from_reviewer.txt
```diff
@@ -1,6 +1,6 @@
 Hello,
 
-{{ author }}, a reviewer at addons.mozilla.org, found an issue when reviewing version {{ number }} of the add-on {{ name }} and requests additional information. You are receiving this email because {{ email_reason }}.
+A reviewer at addons.mozilla.org, found an issue when reviewing version {{ number }} of the add-on {{ name }} and requests additional information. You are receiving this email because {{ email_reason }}.
```
diox, line 3: Probably remove the comma after `addons.mozilla.org` ?

src/olympia/activity/utils.py
bobsilverberg: I also changed the consts to be `USER_TYPE_ADDON_AUTHOR` and `USER_TYPE_ADDON_REVIEWER` as that's more descriptive of why these consts exist.

bobsilverberg:
Thanks for the review @diox. I have addressed all of the issues.

diox (approved):

PR #18624, opened 2022-01-16
Replace "Occitan (Lengadocian)" with "Occitan"

Rob--W:
Fixes https://github.com/mozilla/addons-server/issues/18626

---

Lengadocian is just a variant of Occitan. In practice, the `oc` category on AMO includes dictionaries from other dialects such as Gascon and Aranese, so the "(Lengadocian)" qualifier is inaccurate.

See https://github.com/mdn/content/issues/11868

### DIFF
--- a/src/olympia/core/languages.py
+++ b/src/olympia/core/languages.py
@@ -198,7 +198,7 @@ ADDITIONAL_PRODUCT_LANGUAGES = {
     'ne-NP': {'english': 'Nepali', 'native': '\u0928\u0947\u092a\u093e\u0932\u0940'},
     'nr': {'english': 'Ndebele, South', 'native': 'isiNdebele'},
     'nso': {'english': 'Northern Sotho', 'native': 'Sepedi'},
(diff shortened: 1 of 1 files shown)


willdurand:
That looks reasonable but the AMO language used on the language tools page is implemented in addons-frontend so we probably should update this file too: https://github.com/mozilla/addons-frontend/blob/e24d5643b49f50e38fadb2ee947ba429456d3795/src/amo/languages.js#L444

Rob--W:
> That looks reasonable but the AMO language used on the language tools page is implemented in addons-frontend so we probably should update this file too: https://github.com/mozilla/addons-frontend/blob/e24d5643b49f50e38fadb2ee947ba429456d3795/src/amo/languages.js#L444

I've submitted a PR here: https://github.com/mozilla/addons-frontend/pull/11095

willdurand:
I edited the PR description to include the issue: https://github.com/mozilla/addons-server/issues/18626

willdurand (approved):

PR #18627, opened 2022-01-17
Prevent generation of site permission add-ons for denied install origins

### LINKED ISSUE #18608: Check whether origin is allowed in site permission add-on generator
https://github.com/mozilla/addons-server/issues/18136 will allow developers to generate site permissions add-ons for a given origin. Because the add-on is automatically generated, it bypasses the linter, it also currently bypasses the mechanism to deny origins.

We need to run `annotate_validation_results()` or equivalent to benefit from `DeniedInstallOrigin.find_denied_origins()`, which will ensure we don't generate site permission add-ons for origins we've denied.

diox:
Fixes #18608

### DIFF
--- a/src/olympia/devhub/forms.py
+++ b/src/olympia/devhub/forms.py
@@ -62,6 +62,7 @@ from olympia.users.models import (
 from olympia.versions.models import (
     VALID_SOURCE_EXTENSIONS,
     ApplicationsVersions,
+    DeniedInstallOrigin,
     License,
     Version,
 )
@@ -1433,4 +1434,8 @@ class SitePermissionGeneratorForm(forms.Form):
             or parsed.fragment
         ):
             raise forms.ValidationError(error_message)
+        if DeniedInstallOrigin.find_denied_origins([value]):
+            raise forms.ValidationError(
+                str(DeniedInstallOrigin.ERROR_MESSAGE).format(origin=value)
+            )
         return value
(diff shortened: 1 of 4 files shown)


src/olympia/devhub/forms.py
```diff
@@ -1433,4 +1434,8 @@ def clean_origin(self):
             or parsed.fragment
         ):
             raise forms.ValidationError(error_message)
+        if DeniedInstallOrigin.find_denied_origins([value]):
+            raise forms.ValidationError(
+                str(DeniedInstallOrigin.ERROR_MESSAGE).format(origin=value)
```
eviljeff, line 1439: I guess that doesn't get cast into a string anyway along the way?

eviljeff (approved):

src/olympia/devhub/forms.py
```diff
@@ -1433,4 +1434,8 @@ def clean_origin(self):
             or parsed.fragment
         ):
             raise forms.ValidationError(error_message)
+        if DeniedInstallOrigin.find_denied_origins([value]):
```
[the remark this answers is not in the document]
diox: The check for the validation is for regular add-ons that specify an install origin in their manifests. We aren't currently enforcing the presence of that key, but the idea was that we would eventually do that in some circumstances - that's the Distribution Control project.

Site permissions also need an install origin to be specified, so it's re-using the same mechanism, but because they are generated and bypass validation, the check needs to be added here.

src/olympia/devhub/forms.py
diox: Looks like `format()` does it actually, so the explicit cast is not needed. I'll remove.

PR #18633, opened 2022-01-17
Prevent an user from generating a site permission add-on twice for the same origin & permissions

### LINKED ISSUE #18595: Prevent an user from generating a site permission add-on twice for the same origin & permissions
In the form built for #18136, if a site permission add-on already exists for a given origin + permissions *from the same developer*, throw an error. 

If a site permission add-on already exists for the same origin+permissions but the user is different, let it through: we're not planning to verify origin ownership of the origin, so we have to allow different users to generate an add-on for the same origin.

Note: Someone could still end up as the author of multiple site permission add-ons through authorship changes, race conditions with submitting the form twice while the task is running, etc. The goal with this is to prevent mistakes, it's not meant as a security feature - we have denied origins if we want to completely prevent submissions for a given origin.

diox:
Fixes #18595

### DIFF
--- a/src/olympia/devhub/tests/test_forms.py
+++ b/src/olympia/devhub/tests/test_forms.py
@@ -27,7 +27,7 @@ from olympia.amo.tests.test_helpers import get_image_path
 from olympia.amo.utils import rm_local_tmp_dir
 from olympia.applications.models import AppVersion
 from olympia.devhub import forms
-from olympia.files.models import FileUpload
+from olympia.files.models import FileSitePermission, FileUpload
 from olympia.signing.views import VersionView
 from olympia.tags.models import AddonTag, Tag
 from olympia.versions.models import ApplicationsVersions, DeniedInstallOrigin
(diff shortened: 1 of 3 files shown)


src/olympia/devhub/tests/test_forms.py
```diff
@@ -929,25 +932,71 @@ def test_site_permission_generator_origin_invalid(origin):
...
+    version.installorigin_set.create(origin='https://example.com')
+
+    # User already has a site permission, but for a different origin/permissions.
+    form = forms.SitePermissionGeneratorForm(
+        {'site_permissions': _DEFAULT_SITE_PERMISSIONS, 'origin': 'https://foo.com'},
+        request=request,
+    )
+    assert form.is_valid()
+
+    # User already has a site permission, but for different permissions
```
bobsilverberg, line 980: There's a test for different origin/different permissions, and a test for same origin/different permissions, but there isn't a test for different origin/same permissions. Should there be?

bobsilverberg (approved):
Looks good, just a question about test coverage.

r+wc

src/olympia/devhub/tests/test_forms.py
diox: Yeah, was a bit lazy, added in fe9f7bfef82ad38741f9fca7bf896cba00f6bc6c

PR #18642, opened 2022-01-19
Add readthedocs config file to force Python 3.9 to be used

### LINKED ISSUE #18641: Fix documentation build
Sphinx 4.4.0 broke our RTD builds, because it added importlib-metadata dependency, which we don't install for docs building on readthedocs.

It was added to `docs.txt` in https://github.com/mozilla/addons-server/pull/18640 but that revealed another longstanding issue: RTD builds the doc with its own special process that is subtly different from ours, so sometimes CI will pass for `docs` when the RTD build would fail. Notably, it uses Python 3.7 when we are using 3.9.

We should try to make the docs building more consistent with the way we do CI, or at least fix this the immediate build issue (adding `typing-extensions` dependency to `docs.txt` as `importlib-metadata` requires it with Python < 3.8)

diox:
Fixes #18641 (hopefully)

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,19 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+build:
+  os: ubuntu-20.04
+  tools:
+    python: "3.9"
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+   configuration: docs/conf.py
+
+python:
+   install:
+   - requirements: requirements/docs.txt


.readthedocs.yml
```diff
@@ -0,0 +1,19 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+build:
+  os: ubuntu-20.04
```
eviljeff, line 9: I guess this is the ubuntu version circleci uses in it's docker containers?  (which is annoyingly different than we use in our docker containers - debian buster vs ubuntu 20.04, based on debian bullseye)

eviljeff (approved):
🤞

.readthedocs.yml
diox: Yeah, it matches CI (which annoyingly has its own docker images, but at least it's the same Ubuntu version). I wish our docker images were based on the same version, but at least this is consistent with CI.

PR #18649, opened 2022-01-19
fxa notification webhook

### LINKED ISSUE #14730: Migrate FxA notifications from SQS to webhooks
bbangert mentions webhooks is the new thing in #13896 - we should migrate to them if we can (we probably need a full example of usage to build a proof of concept first - there may be limitations).  This will remove a dependency on AWS infrastructure.

eviljeff:
fixes #14730 - on the code side, anyway.  We will need to liaise with FxA to enable notifications to the web-hook on dev, stage, and prod; AMO ops will need to stop the SQS monitoring command.  (Then clean-up)

Note: this can't be tested locally - it relies on FxA sending signed JWTs to a specific url - so we'll only really find out on addons-dev if it's working as expected.

### DIFF
--- a/src/olympia/accounts/views.py
+++ b/src/olympia/accounts/views.py
@@ -3,6 +3,7 @@ import binascii
 import functools
 import os
 
+from datetime import datetime
 from urllib.parse import quote_plus
 
 from django.conf import settings
@@ -17,6 +18,8 @@ from django.utils.html import format_html
 from django.utils.translation import gettext_lazy as _
 from django.views.decorators.cache import never_cache
 
+import jwt
+import requests
 import waffle
 
 from corsheaders.conf import conf as corsheaders_conf
@@ -28,10 +31,10 @@ from corsheaders.middleware import (
     ACCESS_CONTROL_MAX_AGE,
 )
 from django_statsd.clients import statsd
+from rest_framework import exceptions
 from rest_framework import serializers
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.decorators import action
-from rest_framework.exceptions import PermissionDenied
 from rest_framework.generics import GenericAPIView
 from rest_framework.mixins import (
     DestroyModelMixin,
@@ -73,6 +76,7 @@ from .serializers import (
     UserNotificationSerializer,
     UserProfileSerializer,
 )
+from .tasks import clear_sessions_event, delete_user_event, primary_email_change_event
 from .utils import fxa_login_url, generate_fxa_state
 
 
@@ -134,7 +138,7 @@ def find_user(identity):
             # Alternatively if someone tried to log in as the task user then
             # prevent that because that user "owns" a number of important
             # addons and collections, and it's actions are special cased.
-            raise PermissionDenied()
+            raise exceptions.PermissionDenied()
         return user
     except UserProfile.DoesNotExist:
         return None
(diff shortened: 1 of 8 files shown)


bqbn:
Once this is landed we can stop running the sqs listener docker container, is that correct?

eviljeff:
> Once this is landed we can stop running the sqs listener docker container, is that correct?

Yes. Well, once the webhook has been registered on the FxA side and it's sending notifications, yes.  (I'm going to request just for addon-dev initially)

src/olympia/accounts/views.py
```diff
@@ -774,3 +778,88 @@ def post(self, request):
...
+
+        if not self.fxa_verifying_keys:
+            raise exceptions.AuthenticationFailed(
+                'FXA verifying keys are not available.'
+            )
+
+        return self.fxa_verifying_keys
+
+    def get_jwt(self, request):
+        request_jwt = request.headers['Authorization'].split('Bearer ')[1]
```
diox, line 810: Probably want a 40x when there is no `Authorization` header, or when the split value length is not 2.

src/olympia/accounts/views.py
```diff
@@ -774,3 +778,88 @@ def post(self, request):
...
+        client_id = self.get_fxa_config(request)['client_id']
+
+        for verifying_key in self.get_fxa_verifying_keys():
+            if verifying_key['alg'] != 'RS256':
+                # we only support RS256
+                continue
+            algorithm = jwt.algorithms.RSAAlgorithm.from_jwk(verifying_key)
+            authenticated_jwt = jwt.decode(
```
diox, line 818: These could raise a number of exceptions we likely want to catch and raise `AuthenticationFailed` for

src/olympia/accounts/views.py
```diff
@@ -774,3 +778,88 @@ def post(self, request):
...
+                primary_email_change_event.delay(uid, timestamp, new_email)
+        if event_key == self.FXA_DELETE_EVENT:
+            log.info(f'Fxa Webhook: Got delete event for {uid}')
+            delete_user_event.delay(uid, timestamp)
+        if event_key == self.FXA_PASSWORDCHANGE_EVENT:
+            log.info(f'Fxa Webhook: Got password-change event for {uid}')
+            clear_sessions_event.delay(uid, timestamp, 'password-change')
+
+    def post(self, request):
+        jwt = self.get_jwt(request)
```
diox, line 851: I would call this `payload` rather than `jwt` to avoid confusion with the `jwt` module (and to be consistent with how we call this in our existing auth code)

src/olympia/accounts/views.py
```diff
@@ -774,3 +778,88 @@ def post(self, request):
                 _('Notification [%s] does not exist') % notification_name
             )
         return Response(serializer.data)
+
+
+class FxaNotificationView(FxAConfigMixin, APIView):
+    authentication_classes = []  # TODO: reimplement FxA auth in authentication_class?
```
diox, line 784: The one thing it could bring us would be sharing some code with the auth class that already deals with JWT... though I don't think there would be much besides parsing of the header, since we have special considerations for jwts crafted by developers with API keys which is completely different from dealing with FxA

src/olympia/accounts/views.py
```diff
@@ -774,3 +778,88 @@ def post(self, request):
...
+        if event_key == self.FXA_PASSWORDCHANGE_EVENT:
+            log.info(f'Fxa Webhook: Got password-change event for {uid}')
+            clear_sessions_event.delay(uid, timestamp, 'password-change')
+
+    def post(self, request):
+        jwt = self.get_jwt(request)
+        events = jwt.get('events', {})
+        uid = jwt.get('sub')
+        for event_key, event_data in events.items():
+            log.debug(
```
diox, line 855: Not sure we want to log all of this, it may contain private/sensitive info. I know it's just a `debug` but it may be risky to leave.

diox (approved):
First pass, this looks good but I want to play around with it a bit more

bqbn:
> > Once this is landed we can stop running the sqs listener docker container, is that correct?
> 
> Yes. Well, once the webhook has been registered on the FxA side and it's sending notifications, yes. (I'm going to request just for addon-dev initially)

Thank you for confirming that. Please file a bug with me to remove the docker container when the time comes.

eviljeff:
https://github.com/mozilla/addons-server/pull/18649/commits/3cdf04e86937f5dfcf495a1f6400d88fdc5fc5b4 addresses the review comments @diox

diox (approved):

PR #18667, opened 2022-01-21
use modern syntax in gen-cron.py; format consistently

### LINKED ISSUE #18666: gen-cron.py script isn't python3
When adding a crontab entry, and double checking how the template was parsed I noted that `gen-cron.py`, that's used by ops deployment scripts to generate the cron file, is using syntax that was removed in python 3.  That should be updated.

eviljeff:
fixes #18666

### DIFF
--- a/scripts/crontab/gen-cron.py
+++ b/scripts/crontab/gen-cron.py
@@ -29,10 +29,12 @@ def main():
     if not opts.deprecations:
         opts.python += ' -W ignore::DeprecationWarning'
 
-    django_cmd = f'{opts.user or ""} cd {opts.zamboni}; {opts.python} manage.py'.strip()
+    django_cmd = '{u} cd {z}; {p} manage.py'.format(
+        u=(opts.user or ''), z=opts.zamboni, p=opts.python
+    ).strip()
     ctx = {
         'django': django_cmd,
-        'z_cron': f'{django_cmd} cron',
+        'z_cron': '{django_cmd} cron'.format(django_cmd=django_cmd),
         'python': opts.python,
     }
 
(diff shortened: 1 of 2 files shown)


eviljeff:
@bqbn one thing that occurs to me, https://github.com/mozilla-services/cloudops-deployment/blob/master/projects/addons-server/ansible/playbooks/tasks/deploy_commands_admin.yml#L44 runs this script and it's (presumably) not been failing despite using syntax that fails in modern python.  So what version of python is being used to run that?

(if the answer is python2.7, and we can't/don't want to change that, I need to drop the `f'...'` strings because they don't exist in python2.7)

diox:
The ansible playbook always explicitly passes the `-p` argument when generating the cron, so we avoiding the issue.

eviljeff:
> The ansible playbook always explicitly passes the `-p` argument when generating the cron, so we avoiding the issue.

the invalid in python3 syntax is when `-u` is passed

diox:
Oh right, there were 2 issues here, I thought you meant the `python2.7` binary used by default would not exist...

diox (approved):
A simple test proving that the script can be run (the way we do it right now, at least) would be nice.

Need @bqbn to weigh in on the python executable question, but IMHO we should ensure everything uses Python 3.

bqbn:
> So what version of python is being used to run that?
> 

Python v2.7. 

The `gen-cron.py` script is executed on the admin instance not in a container thus it uses the Python that comes with the OS. And the reason for running `gen-cron.py` on the instance not in a container is because we need to put those cron jobs on the instance's cron table and run them in containers.

We aren't upgrading Python before moving to GCP so please drop the f-string usage.

PR #18670, opened 2022-01-21
return 'ERROR_AUTHENTICATION_EXPIRED', when there's a session auth mismatch

### LINKED ISSUE #18669: return 'ERROR_AUTHENTICATION_EXPIRED' when session_auth mismatches
When a user logs out or ends up creating a new session the session auth hash is cleared or regenerated, leading to frontend auth tokens now being invalid.  We don't return a code for that case, but if we do return `ERROR_AUTHENTICATION_EXPIRED` then frontend can handle it and show the appropriate error page to prompt the user to re-authenticate.

eviljeff:
fixes #18669 - not sure on the messaging but frontend isn't going to show them anyway.

### DIFF
--- a/src/olympia/api/authentication.py
+++ b/src/olympia/api/authentication.py
@@ -142,7 +142,11 @@ class WebTokenAuthentication(BaseAuthentication):
                 'User tried to authenticate with invalid auth hash in'
                 'payload {}'.format(payload)
             )
-            raise exceptions.AuthenticationFailed()
+            msg = {
+                'detail': gettext('Auth hash mismatch. Session is likely expired.'),
+                'code': 'ERROR_AUTHENTICATION_EXPIRED',
+            }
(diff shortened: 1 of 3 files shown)


src/olympia/api/authentication.py
```diff
@@ -142,7 +142,11 @@ def authenticate_credentials(self, payload):
                 'User tried to authenticate with invalid auth hash in'
                 'payload {}'.format(payload)
             )
-            raise exceptions.AuthenticationFailed()
+            msg = {
+                'detail': gettext('Auth hash mismatch. Session is likely expired.'),
```
diox, line 146: Not a big fan of that messaging, but if it's going to be replaced by something more generic on the frontend maybe that's fine... It's tough to pick something given it's a fairly technical issue...

diox (approved):

PR #18671, opened 2022-01-21
Stop exposing reviewer names on devhub

### LINKED ISSUE #18564: Stop exposing reviewer names to developers
This is a spin-off from https://github.com/mozilla/addons-server/issues/17318 and is just about removing the names of reviewers from any screens that developers can see, which should be exclusively devhub pages.

bobsilverberg:
Fixes #18564

### DIFF
--- a/src/olympia/activity/models.py
+++ b/src/olympia/activity/models.py
@@ -420,6 +420,13 @@ class ActivityLog(ModelBase):
         return markupsafe.Markup(self.formatter.format(*args, **kw))
 
     @classmethod
+    def transformer_hide_user_from_developer(cls, logs):
+        """Remove user from actions where it shouldn't be shown to a developer."""
+        for log in logs:
+            if log.action not in constants.activity.LOG_SHOW_USER_TO_DEVELOPER:
+                log.user = None
+
+    @classmethod
     def arguments_builder(cls, activities):
         def handle_renames(value):
             # Cope with renames of key models (use the original model name like
@@ -673,13 +680,7 @@ class ActivityLog(ModelBase):
 
     @property
     def author_name(self):
-        """Name of the user that triggered the activity.
-
-        If it's a reviewer action that will be shown to developers, the
-        `reviewer_name` property is used if present, otherwise `name` is
-        used."""
-        if self.action in constants.activity.LOG_REVIEW_QUEUE_DEVELOPER:
-            return self.user.reviewer_name or self.user.name
+        """Name of the user that triggered the activity."""
         return self.user.name
 
     @classmethod
--- a/src/olympia/activity/models.py
+++ b/src/olympia/activity/models.py
@@ -435,9 +435,11 @@ class ActivityLog(ModelBase):
         """Replace the user with a generic user in actions where it shouldn't
         be shown to a developer.
         """
+        generic_user = GenericMozillaUser()
+
         for log in logs:
             if log.action not in constants.activity.LOG_SHOW_USER_TO_DEVELOPER:
-                log.user = GenericMozillaUser.objects.create()
+                log.user = generic_user
 
     @classmethod
     def arguments_builder(cls, activities):
--- a/src/olympia/activity/models.py
+++ b/src/olympia/activity/models.py
@@ -39,6 +39,17 @@ log = olympia.core.logger.getLogger('z.amo.activity')
 # Number of times a token can be used.
 MAX_TOKEN_USE_COUNT = 100
 
+GENERIC_USER_NAME = gettext('Add-ons Review Team')
+
+
+class GenericMozillaUser(UserProfile):
+    class Meta:
+        proxy = True
+
+    @property
+    def name(self):
+        return GENERIC_USER_NAME
+
 
 class ActivityLogToken(ModelBase):
     id = PositiveAutoField(primary_key=True)
@@ -420,11 +431,13 @@ class ActivityLog(ModelBase):
         return markupsafe.Markup(self.formatter.format(*args, **kw))
 
     @classmethod
-    def transformer_hide_user_from_developer(cls, logs):
-        """Remove user from actions where it shouldn't be shown to a developer."""
+    def transformer_anonymize_user_for_developer(cls, logs):
+        """Replace the user with a generic user in actions where it shouldn't
+        be shown to a developer.
+        """
         for log in logs:
             if log.action not in constants.activity.LOG_SHOW_USER_TO_DEVELOPER:
-                log.user = None
+                log.user = GenericMozillaUser.objects.create()
 
     @classmethod
     def arguments_builder(cls, activities):
@@ -677,11 +690,6 @@ class ActivityLog(ModelBase):
 
     def __html__(self):
         return self
-
-    @property
-    def author_name(self):
-        """Name of the user that triggered the activity."""
-        return self.user.name
 
     @classmethod
     def create(cls, action, *args, **kw):
--- a/src/olympia/activity/views.py
+++ b/src/olympia/activity/views.py
@@ -19,7 +19,10 @@ from rest_framework.viewsets import GenericViewSet
 import olympia.core.logger
 
 from olympia import amo
-from olympia.activity.models import ActivityLog
+from olympia.access import acl
+from olympia.activity.models import (
+    ActivityLog,
+)
 from olympia.activity.serializers import ActivityLogSerializer
 from olympia.activity.tasks import process_email
 from olympia.activity.utils import (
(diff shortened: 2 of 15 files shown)


src/olympia/activity/models.py
```diff
@@ -673,13 +680,7 @@ def __html__(self):
 
     @property
     def author_name(self):
```
bobsilverberg, line 682: This is only still is use at https://github.com/mozilla/addons-server/blob/master/src/olympia/amo/feeds.py, and I'm not exactly sure what it's doing there, and if we could just replace it with `user.name` as I have done in a number of other places.

src/olympia/activity/models.py
```diff
@@ -419,6 +419,13 @@ def f(self, *args, **kw):
         # SafeFormatter escapes everything so this is safe.
         return markupsafe.Markup(self.formatter.format(*args, **kw))
 
+    @classmethod
+    def transformer_hide_user_from_developer(cls, logs):
```
bobsilverberg, line 423: I was hoping to do the actual anonymization in here, rather than just making the `user` `None`, but it didn't seem happy with me trying to change `log.user.username` in here. I am guessing we can only change things that exist on the `log` instance itself? Maybe there is a different way of doing it in here?

src/olympia/devhub/templates/devhub/addons/activity.html
```diff
@@ -30,7 +30,7 @@ <h2>{{ _('Recent Activity for My Add-ons') }}</h2>
             {{ item }}
           </p>
           <p class="timestamp">
-          {% trans user=item.author_name,
+          {% trans user=item.user.name if item.user else generic_name,
```
bobsilverberg, line 33: This is done in a couple of templates. If I cannot figure out a way to do this at a higher level (e.g., in the transformer), then maybe this duplication could be addressed via a jinja helper?

bobsilverberg:
As discussed on Matrix, I haven't addressed the case of the Manage Versions screen which calls the `reviewnotes` API. I think I need a better understanding of what we should do with that.

eviljeff (commented):
We _do_ want to address the review notes API before this can be merged though - otherwise we're un-anonymizing the reviewers.  I may have missed something on Matrix but we can check the user's permissions (to see if they are a reviewer) and selectively apply the transform in `VersionReviewNotesViewSet`?

src/olympia/devhub/views.py
```diff
@@ -80,6 +80,7 @@
 # We use a session cookie to make sure people see the dev agreement.
 
 MDN_BASE = 'https://developer.mozilla.org/en-US/Add-ons'
+GENERIC_USER_NAME = _('Add-ons Review Team')
```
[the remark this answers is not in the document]
eviljeff: oh, I see you mention it causes an error doing that.  See my suggestion about `GenericMozillaUser`

src/olympia/activity/models.py
bobsilverberg: In `feeds.py` it looks like a list of keys, two of which are `author_name` and `item_author_name`. I'm not sure what to replace those with. Can you advise?

src/olympia/constants/activity.py
```diff
@@ -67,12 +73,14 @@ class USER_DISABLE(_LOG):
     id = 8
     format = _('{addon} disabled.')
     keep = True
+    show_user_to_developer = True
```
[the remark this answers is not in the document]
bobsilverberg: It's based on which actions we think a developer should be able to see the user for. It started with my best guess and then @diox added a bunch as well. I am going to ask @wagnerand to review the final PR for completeness of this list.

src/olympia/activity/models.py
eviljeff: Seems like the use in feeds.py is unconnected - it predates the reviewer names feature

eviljeff: And it's probably unnecessary code anyway - at the time we had rss feeds that contained user submitted content (add-on names, etc) - but now the only rss feed that uses that class is for app versions, that's entirely generated from internal models.

bobsilverberg: Cool, thanks. I wasn't sure if it was related or not. That means we can remove `author_name` from this class.

src/olympia/devhub/templates/devhub/addons/activity.html
bobsilverberg: Dealt with via `GenericMozillaUser`

bobsilverberg:
> We _do_ want to address the review notes API before this can be merged though - otherwise we're un-anonymizing the reviewers. I may have missed something on Matrix but we can check the user's permissions (to see if they are a reviewer) and selectively apply the transform in `VersionReviewNotesViewSet`?

Your suggestion is to check the permissions of the current user, the one calling the API, correct? And then if they are not a reviewer we call the transformer to anonymize, but if they are a reviewer we do not call it?

eviljeff:
> > We _do_ want to address the review notes API before this can be merged though - otherwise we're un-anonymizing the reviewers. I may have missed something on Matrix but we can check the user's permissions (to see if they are a reviewer) and selectively apply the transform in `VersionReviewNotesViewSet`?
> 
> Your suggestion is to check the permissions of the current user, the one calling the API, correct? And then if they are not a reviewer we call the transformer to anonymize, but if they are a reviewer we do not call it?

yes.  By the time that code has been called the user's groups will have been checked and cached anyway by the permission and authentication classes.

bobsilverberg:
@wagnerand The `show_user_to_developer attribute in `src/olympia/constants/activity.py` is meant to indicate for which actions we will show the actual user name to a developer. Any actions without this attribute will have the names of the users anonymized. Can you please look through the list and verify if I have captured all of the actions needed, and also whether I have marked anything that should not have been marked?

src/olympia/constants/activity.py
```diff
@@ -55,24 +60,28 @@ class REMOVE_USER_WITH_ROLE(_LOG):
...
     format = _('{0.name} ({1}) removed from {addon}.')
     keep = True
+    show_user_to_developer = True
 
 
 class EDIT_CONTRIBUTIONS(_LOG):
     id = 7
     action_class = 'edit'
     format = _('Contributions for {addon}.')
+    show_user_to_developer = True
```
wagnerand, line 70: Fwiw, this activity is no longer used.

src/olympia/constants/activity.py
```diff
@@ -629,6 +650,7 @@ class DEVELOPER_CLEAR_INFO_REQUEST(_LOG):
     short = _('Information request removed')
     keep = True
     review_queue = True
+    show_user_to_developer = True
```
wagnerand, line 653: This activity looks unused as well.

wagnerand:
> @wagnerand The `show_user_to_developer attribute in `src/olympia/constants/activity.py` is meant to indicate for which actions we will show the actual user name to a developer. Any actions without this attribute will have the names of the users anonymized. Can you please look through the list and verify if I have captured all of the actions needed, and also whether I have marked anything that should not have been marked?

The list looks good overall, thanks!

wagnerand (approved):

src/olympia/activity/models.py
```diff
@@ -39,6 +39,17 @@
...
+
+
+class GenericMozillaUser(UserProfile):
+    class Meta:
+        proxy = True
+
+    @property
+    def name(self):
+        return GENERIC_USER_NAME
+
```
eviljeff, line 52: not required for this patch, but if we start to use this class more widely we should override the save method (and possibly others) to raise if it's inadvertently written to the database or used in unexpected ways.

src/olympia/activity/views.py
```diff
@@ -48,6 +51,8 @@ class VersionReviewNotesViewSet(
 
     def get_queryset(self):
         alog = ActivityLog.objects.for_versions(self.get_version_object())
+        if not acl.is_user_any_kind_of_reviewer(self.request.user):
```
eviljeff, line 54: pass `allow_viewers=False` too

eviljeff (approved):
Looks good!  We might find some more edge cases - especially around the api - but this seems to cover everything obvious.

src/olympia/activity/models.py
bobsilverberg: Good idea. Should I file an issue about that, or shall we just wait to see if we end up using it for something else and address it at that time?

eviljeff: I'd just wait

src/olympia/activity/views.py
bobsilverberg: Confirmed via Matrix that you meant `allow_viewers=True`. 😄

PR #18681, opened 2022-01-24
Always pass user and ip address from upload to Addon/Version creation

### LINKED ISSUE #18650: Activity logs for site-permission add-on generation are incomplete
Follow up for https://github.com/mozilla/addons-server/issues/18638#issuecomment-1016578702

1. Request a site-permission add-on through the [devhub form](https://addons.allizom.org/en-US/developers/site_permission_generator/)
2. Open the admin [Activity logs](https://addons-internal.stage.mozaws.net/en-US/admin/models/activity/activitylog/) page and search for entries corresponding to the site-permission add-on generation

### What happened?
You can see an `ActivityLog` for the `{id:username} | {version} added to the {addon}` but not one for the add-on creation itself

An `ActivityLog` of the format `{id:username} | {addon-name} was created` should be created when requesting a new site-permission add-on

image

diox:
Also add constraints to forbid creation of `FileUpload`s without a user/ip_address/source.

Fixes #18650


src/olympia/addons/utils.py
```diff
@@ -385,7 +385,12 @@ def create_version(self, addon=None):
                 user=self.user,
                 source=amo.UPLOAD_SOURCE_GENERATED,
             )
+
         # And finally create the Version instance from the FileUpload.
-        return create_version_for_upload(
-            addon, upload, amo.RELEASE_CHANNEL_UNLISTED, parsed_data=parsed_data
```
diox, line 390: `create_version_for_upload` is an unnecessary wrapper in this case.

src/olympia/addons/models.py
```diff
@@ -892,19 +892,19 @@ def initialize_addon_from_upload(cls, data, upload, channel, user):
...
 
     @classmethod
     def from_upload(
         cls,
+        *,
         upload,
         selected_apps,
         channel=amo.RELEASE_CHANNEL_LISTED,
-        parsed_data=None,
-        user=None,
```
diox, line 907: Removed this because this method receives a `FileUpload`, so we should already have a `user`.

src/olympia/addons/models.py
```diff
@@ -892,7 +892,7 @@ def initialize_addon_from_upload(cls, data, upload, channel, user):
             )
         if user:
             AddonUser(addon=addon, user=user).save()
-        activity.log_create(amo.LOG.CREATE_ADDON, addon)
+        activity.log_create(amo.LOG.CREATE_ADDON, addon, user=user)
```
diox, line 895: This is the crucial bit I needed to solve, but getting to that point involved making sure `FileUpload`s always had a `user` set, hence the big refactor

eviljeff: "+557 −231" 😿

diox: Mostly tests though !

src/olympia/devhub/views.py
```diff
@@ -699,15 +699,14 @@ def handle_upload(
     submit=False,
     source=amo.UPLOAD_SOURCE_DEVHUB,
 ):
-    user = request.user if request.user.is_authenticated else None
```
eviljeff, line 702: I assumed this was for the standalone upload case (just submitting to test validation) - I forget, do we support that unauthenticated these days?

eviljeff (commented):
I mainly have questions

src/olympia/devhub/views.py
diox: No, it requires auth.

src/olympia/addons/models.py
```diff
@@ -892,7 +892,7 @@ def initialize_addon_from_upload(cls, data, upload, channel, user):
             )
         if user:
```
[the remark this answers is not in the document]
diox: mmm it shouldn't. If it is I'd prefer an explicit exception being raised, so I'll remove this `if`

src/olympia/addons/tests/test_models.py
```diff
@@ -2714,26 +2725,36 @@ def test_webext_resolve_translations_unknown_locale(self):
...
         # Normalized from `en` to `en-US`
         assert addon.default_locale == 'en-US'
 
     def test_activity_log(self):
-        core.set_user(self.user)
-        self.upload = self.get_upload('webextension.xpi')
+        # Set current user as the task user, but use an upload that belongs to
+        # another, making sure the activity log belongs to them and not the
+        # task user.
+        core.set_user(UserProfile.objects.get(pk=settings.TASK_USER_ID))
```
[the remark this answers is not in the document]
diox: `base/users` fixture is loaded for this test class

src/olympia/versions/models.py
```diff
@@ -283,7 +282,11 @@ def from_upload(
         if upload.addon and upload.addon != addon:
             raise VersionCreateError('FileUpload was made for a different Addon')
 
-        if not upload.user or not upload.ip_address or not upload.source:
+        if (
+            not getattr(upload, 'user', None)
```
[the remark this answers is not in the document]
diox: It always does, but when `upload.user` is `None`, django raises an exception (because I'm no longer allowing that in the model). The `getattr` is just to avoid this so I can raise my own (it's really a use case that shouldn't be happening any more, but I left this condition to be a bit paranoid)

src/olympia/files/models.py
```diff
@@ -557,14 +555,19 @@ def get_authenticated_download_url(self):
         return f'{absolute_url}?access_token={self.access_token}'
 
     @classmethod
-    def from_post(cls, chunks, filename, size, **params):
+    def from_post(
+        cls, chunks, *, filename, size, addon=None, version='', user, source, channel
```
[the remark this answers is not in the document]
diox: mmm it should probably be `None` instead

src/olympia/files/models.py
```diff
@@ -464,10 +464,8 @@ class FileUpload(ModelBase):
     version = models.CharField(max_length=255, null=True)
     addon = models.ForeignKey('addons.Addon', null=True, on_delete=models.CASCADE)
     access_token = models.CharField(max_length=40, null=True)
-    ip_address = models.CharField(max_length=45, null=True, default=None)
-    source = models.PositiveSmallIntegerField(
-        choices=amo.UPLOAD_SOURCE_CHOICES, default=None, null=True
-    )
+    ip_address = models.CharField(max_length=45)
+    source = models.PositiveSmallIntegerField(choices=amo.UPLOAD_SOURCE_CHOICES)
```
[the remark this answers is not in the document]
diox: @wagnerand confirmed there are no file uploads with `ip_address`, `user` or `source` being `NULL` in production. I checked dev&stage and there weren't any there either. We can thank the `gc` cron for that - any old file uploads from before the change to pass those fields was introduced have been deleted

src/olympia/devhub/forms.py
```diff
@@ -1019,6 +1019,7 @@ def _clean_upload(self):
         if not (
             self.cleaned_data['upload'].valid
             or self.cleaned_data['upload'].validation_timeout
+            or self.cleaned_data['upload'].user != self.request.user
```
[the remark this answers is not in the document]
diox: I refactored that and added a test in 19a843bc76a32154495a3dceff12dfff6576ee4b, I think it should be enough. QA should verify that uploading and bypassing validation works though.

eviljeff (approved):

PR #18706, opened 2022-01-27
Remove lib/cache.py, simplify CACHES settings

### LINKED ISSUE #18085: Remove `lib/cache.py`
We have a few functions in `lib/cache.py` that are almost useless nowadays:
- `make_key()` is a function to normalize cache keys. The only use left is in `reviewers/serializers.py` to make a key for `_get_hash_for_selected_file` that is hashed to avoid being too long. This could be replaced by doing something like that in `KEY_FUNCTION` https://docs.djangoproject.com/en/3.2/ref/settings/#key-function in our `CACHES` setting instead, doing that for all keys all the time. The `with_locale` part is not used anymore.
- `memoize_key()` is only used by `memoize()` below
- `memoize()` is a caching decorator that we no longer use

diox:
Fixes #18085


settings.py
```diff
@@ -106,21 +98,24 @@ def insert_debug_toolbar_middleware(middlewares):
         'client_id': env('FXA_CLIENT_ID', default='a25796da7bc73ffa'),
         'client_secret': env(
             'FXA_CLIENT_SECRET',
-            default='4828af02f60a12738a79c7121b06d42b481f112dce1831440902a8412d2770c5'),  # noqa
+            default='4828af02f60a12738a79c7121b06d42b481f112dce1831440902a8412d2770c5',
+        ),  # noqa
```
eviljeff, line 102: I'm surprised these `# noqa` still work on the following line?

src/olympia/reviewers/serializers.py
```diff
@@ -141,10 +141,13 @@ def _get_hash_for_selected_file(self):
...
         # Normalize the key as we want to avoid that we exceed max
         # key lengh because of selected_file.
-        cache_key = make_key(
-            f'reviewers:fileentriesserializer:hashes' f':{commit.hex}:{selected_file}',
-            with_locale=False,
-            normalize=True,
+        cache_key = force_str(
+            hashlib.sha256(
+                force_bytes(
+                    f'reviewers:fileentriesserializer:hashes'
```
eviljeff, line 147: nit: this isn't an f-string.

eviljeff (approved):

settings.py
diox: Ah, I think those are no longer necessary ? black reformatted this for me... I'll remove them

PR #18718, opened 2022-01-28
Allow Fake FxA login page to redirect to :7000

### LINKED ISSUE #18637: Allow Fake FxA login page to redirect to :7000
When testing locally with `yarn amo:olympia`, the site is hosted at `olympia.test:7000`. After logging in via the Fake FxA screen, you are redirected to  `olympia.test`, without the port. This is quite annoying as you have to remember to update the URL to include the `7000` port, which is fine for the most part, but if it's important to test something without reloading the site then you are out of luck.

It would be nice if the Fake FxA page could somehow notice that you have come from port 7000, and then redirect you back to that port. If we can do that conditionally then it shouldn't interfere with the running of addons-frontend from within the docker environment.

I can work on this, but if either @diox or @eviljeff have any helpful tips they would be welcome.

bobsilverberg:
Fixes #18637

Note that this will be used with a change to addons-frontend that will pass the fully qualified URL (including the port) as the `to` when logging in from a local development environment.

Note also that this seemed like a simpler/better change than changing `is_safe_url` to specifically allow URLs that we might use with fake FxA. Looking at all the uses of `is_safe_url`, it seems fine to just return `True` if we are currently using fake FxA.


diox:
> Note also that this seemed like a simpler/better change than changing `is_safe_url` to specifically allow URLs that we might use with fake FxA. Looking at all the uses of `is_safe_url`, it seems fine to just return `True` if we are currently using fake FxA.

It's fine but a bit surprising. It's also used by devhub and scanners admin completely outside of FxA at the moment. Maybe those two should use a different function and this one should be renamed & moved to indicate that it's specific to the auth redirects.

bobsilverberg:
> > Note also that this seemed like a simpler/better change than changing `is_safe_url` to specifically allow URLs that we might use with fake FxA. Looking at all the uses of `is_safe_url`, it seems fine to just return `True` if we are currently using fake FxA.
> 
> It's fine but a bit surprising. It's also used by devhub and scanners admin completely outside of FxA at the moment. Maybe those two should use a different function and this one should be renamed & moved to indicate that it's specific to the auth redirects.

That's a good idea!

willdurand:
Why exactly is this change needed? My understanding is that `olympia.test` should be already safe to use because that should be already configured in `settings.DOMAIN`.

bobsilverberg:
> Why exactly is this change needed? My understanding is that `olympia.test` should be already safe to use because that should be already configured in `settings.DOMAIN`.

I think it's because of the port.

diox:
With the current implementation on `master`, `olympia.test` is considered safe, `olympia.test:7000` isn't. Behind the scenes, django's `url_has_allowed_host_and_scheme()` compares the full netloc (hostname+port) and we only pass the hostname.

bobsilverberg:
@diox This is ready for another look.

settings.py
```diff
@@ -67,6 +67,7 @@ def insert_debug_toolbar_middleware(middlewares):
 
 SITE_URL = os.environ.get('OLYMPIA_SITE_URL') or 'http://localhost:8000'
 DOMAIN = SERVICES_DOMAIN = urlparse(SITE_URL).netloc
+ADDONS_FRONTEND_PROXY_PORT = '7000'
```
diox, line 70: Since `settings_test.py` loads settings from `settings.py`, we're going to end up with ``ADDONS_FRONTEND_PROXY_PORT` being defined in tests. Can you set it to `None` in `settings_tests.py` as well ?

src/olympia/amo/utils.py
```diff
@@ -1182,6 +1182,9 @@ def is_safe_url(url, request, allowed_hosts=None):
     allowed_hosts = allowed_hosts or (
         settings.DOMAIN,
         urlparse(settings.CODE_MANAGER_URL).netloc,
+        f'{settings.DOMAIN}:{settings.ADDONS_FRONTEND_PROXY_PORT}'
+        if settings.ADDONS_FRONTEND_PROXY_PORT
+        else None,
```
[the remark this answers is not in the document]
bobsilverberg: I went ahead and made this change

src/olympia/amo/utils.py
```diff
@@ -1182,10 +1182,12 @@ def is_safe_url(url, request, allowed_hosts=None):
...
         settings.DOMAIN,
         urlparse(settings.CODE_MANAGER_URL).netloc,
-        f'{settings.DOMAIN}:{settings.ADDONS_FRONTEND_PROXY_PORT}'
-        if settings.ADDONS_FRONTEND_PROXY_PORT
-        else None,
     )
+    if settings.ADDONS_FRONTEND_PROXY_PORT:
+        allowed_hosts = allowed_hosts + (
+            f'{settings.DOMAIN}:{settings.ADDONS_FRONTEND_PROXY_PORT}',
+        )
```
diox, line 1189: I don't think this is better as it adds the `domain:port` version in every case, even when a custom `allowed_hosts` is passed, which is probably a bit surprising.

Using a separate variable for the "default" allowed hosts or doing `allowed_hosts or (...) + ((f'{settings.DOMAIN}:{settings.ADDONS_FRONTEND_PROXY_PORT}',) if settings.ADDONS_FRONTEND_PROXY_PORT else ())` would fix this. The default variable path is probably going to be a lot easier to read

bobsilverberg: I updated the code to address the issue. I did it a bit differently from how you suggested, but I think it works and is reasonable.

diox (approved):

PR #18734, opened 2022-02-01
Return an empty and cacheable 404 for legacy API requests

### LINKED ISSUE #18733: Raise an explicit error (avoiding regular 404 html page) for obsolete API
We still have some clients hitting URLs like `/en-US/firefox/api/1.5/search/` which have been removed a long time ago (on services.addons.mozilla.org domain, since this is a Firefox request). We serve the generic HTML 404 for those, which can add unnecessary load (it handles auth if necessary, displays the site notice which causes a database request, etc).

We could consider making the generic 404 lighter, but specifically for this case, I think we should raise an explicit error early and don't even return HTML. We could also cache this for a long time, though I suspect clients would all come with different query parameters.

diox:
Fixes #18733

### DIFF
--- a/src/olympia/amo/views.py
+++ b/src/olympia/amo/views.py
@@ -10,7 +10,7 @@ from django.contrib.sitemaps.views import x_robots_tag
 from django.core.exceptions import PermissionDenied, ViewDoesNotExist
 from django.core.paginator import EmptyPage, PageNotAnInteger
 from django.db.transaction import non_atomic_requests
-from django.http import Http404, HttpResponse, JsonResponse
+from django.http import Http404, HttpResponse, HttpResponseNotFound, JsonResponse
 from django.template.response import TemplateResponse
 from django.utils.cache import patch_cache_control
 from django.views.decorators.cache import never_cache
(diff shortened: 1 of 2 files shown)


src/olympia/amo/views.py
```diff
@@ -105,8 +106,15 @@ def handler403(request, exception=None, **kwargs):
...
 def handler404(request, exception=None, **kwargs):
     if getattr(request, 'is_api', False):
-        # It's a v3+ api request
+        # It's a v3+ api request (/api/vX/ or /api/auth/)
         return JsonResponse({'detail': str(NotFound.default_detail)}, status=404)
+    elif re.match(r'^/api/\d\.\d/', getattr(request, 'path_info', '')):
+        # It's a legacy API request in the form of /api/X.Y/. We use path_info,
+        # which is set in LocaleAndAppURLMiddleware, because there might be a
+        # locale and app prefix we don't care about in the URL.
+        response = HttpResponse(status=404)
```
diox, line 115: I'm not brave enough to return a 410 but I considered it...

eviljeff: yeah, a 410 is more accurate but 🤷‍♂️

src/olympia/amo/views.py
```diff
@@ -105,8 +106,15 @@ def handler403(request, exception=None, **kwargs):
...
+        patch_cache_control(response, max_age=60 * 60 * 48)
```
eviljeff, line 116: Could be a (much) longer cache time really - /api/1.5 isn't not coming back!

src/olympia/amo/views.py
eviljeff, line 115: `HttpResponseNotFound`

src/olympia/amo/views.py
diox: I don't think it would matter much because of the sheer number of different URLs (since it's coming from Firefox, it's passing a number of things in the URL...)

diox:
I don't think ops would be too happy with defining yet another exception in nginx. It makes more sense here, the key being avoiding the template rendering + database query that goes with it

PR #18752, opened 2022-02-03
Fix typo in django admin

wagnerand:
This fixes a typo in django admin. Note that I also updated the help text in the inital migration, I hope this doesn't break anything, and I don't think it warrants writing a migration to change the text in existing environments. It would only have an effect for environments being set up after this has landed.

cc @diox for the migration change.


diox (approved):
Note that you could have generated a migration normally : it would have been a no-op, there are  no database changes.

PR #18755, opened 2022-02-03
Remove unused/obsolete JavaScript

### LINKED ISSUE #18142: Remove as much as possible of jquery-ui
Why do we use jquery-ui for these days in addons-server, and could we get rid of it ? 

Off the top of my head I can only think of:
- ~~[ ] [`.sortable()`](https://jqueryui.com/sortable/) in devhub which we could easily find a lightweight replacement for.~~
- [x] [`.datepicker()`](https://api.jqueryui.com/datepicker/) in stats which we could probably replace with an `<input type=date>`: https://github.com/mozilla/addons-server/issues/18738

### LINKED ISSUE #18143: Remove obsolete js code and libraries
On top of https://github.com/mozilla/addons-server/issues/18142 ... we can probably remove:
- `jqmodal`, which doesn't seem to be referenced anywhere
- `pjax`, which might still be referenced in some templates but shouldn't be used anymore
- Almost everything in `static/js/impala/` except for `stats/` and maybe a couple other things the statistics page would use.

diox:
Could do more, but that's a good chunk already...

Fixes #18142
Fixes #18143


src/olympia/lib/settings_base.py
```diff
@@ -706,153 +706,44 @@ def get_db_config(environ_var, atomic_requests=True):
...
             'js/zamboni/helpers.js',
-            'js/common/banners.js',
             'js/zamboni/global.js',
-            'js/amo2009/global.js',
-            'js/common/ratingwidget.js',
-            'js/node_lib/jqModal.js',
             'js/zamboni/l10n.js',
-            'js/zamboni/debouncer.js',
-            # Homepage
-            'js/zamboni/homepage.js',
```
wagnerand, line 749: I don't see that file being removed in this PR. Am I missiing something?

diox: I need to go back to check, it looks like there are a few files I forgot to remove in zamboni/. That's why this PR is a draft :) (It's also broken right now, need to tweak things out a bit)

diox: https://github.com/mozilla/addons-server/pull/18755/commits/f1e136a1ea0fb7242597aede0aed79325c6a1bb2

diox:
Depends on #18768 to fix a few issues with the removal of impala bundle.

diox:
rebasing with https://github.com/mozilla/addons-server/pull/18768 fixed the devhub issues, but the stats page is broken differently, need to sort that out as well... probably more impala stuff to clean up.

eviljeff (approved):
all the 🟥

PR #18757, opened 2022-02-03
rm auth token in api response & document api auth changes

### LINKED ISSUE #18742: stop adding the auth token to the api response
We add the auth token to the api response during login, https://github.com/mozilla/addons-server/blob/master/src/olympia/accounts/views.py#L357
but
a) it's only set when there is a `data` attribute in the response (i.e. a JSON response, which we don't have any current code path to, afaik), and
b) it's not actually consumed by addons-frontend.

### LINKED ISSUE #18743: document new internal authentication methods
Our API docs detail how to use the existing internal auth methods, but not the new session ID based authentication added in #18149.

eviljeff:
fixes #18742 and fixes #18743

### DIFF
--- a/docs/topics/api/auth_internal.rst
+++ b/docs/topics/api/auth_internal.rst
@@ -10,34 +10,31 @@ are looking for how to authenticate with the API from an external client, using
 your API keys, read the :ref:`documentation for external authentication
 <api-auth>` instead.
 
-When using this authentication mechanism, the server is responsible for
-creating an API Token when the user logs in, and sends it back in
-the response. The clients must then include that token as an ``Authorization``
-header on requests that need authentication. The clients never generate JWTs
-themselves.
+When using this authentication mechanism, the server creates a session and stores the
+session id in the ``sessionid` cookie when the user logs in.
+The client must then include that session id in an ``Authorization`` header on requests
+that need authentication.
+The clients never generate tokens or sessions themselves.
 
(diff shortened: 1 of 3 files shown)


docs/topics/api/auth_internal.rst
```diff
@@ -10,34 +10,31 @@ are looking for how to authenticate with the API from an external client, using
...
-When using this authentication mechanism, the server is responsible for
-creating an API Token when the user logs in, and sends it back in
-the response. The clients must then include that token as an ``Authorization``
-header on requests that need authentication. The clients never generate JWTs
-themselves.
+When using this authentication mechanism, the server creates a session and stores the
+session id in the ``sessionid` cookie when the user logs in.
+The client must then include that session id in an ``Authorization`` header on requests
+that need authentication.
+The clients never generate tokens or sessions themselves.
```
bobsilverberg, line 17: Why do you mention "tokens" here when they aren't mentioned anywhere else? (Or maybe they are and I just don't see it in this diff)

bobsilverberg (approved):
Looks good, just one comment.

r+wc

docs/topics/api/auth_internal.rst
eviljeff: External auth uses JWTs, but perhaps the sentence is redundant anyway? (the previous version said something similar)

PR #18758, opened 2022-02-03
enable addon submission api on prod; remove warnings in api docs

### LINKED ISSUE #18371: Enable addon submission api for users on prod
Once all the blocking (MVP) issues in https://github.com/orgs/mozilla/projects/174 are finished we should enable the API on production AMO so it can be used by users.  It's expected they'll be other bugs found before we would want to announce it as "finished" but we won't find those until it's actually in use.

eviljeff:
fixes #18371 - we might decide to gate merging this on some sort of announcement?  But I'm not sure it's a blocker

### DIFF
--- a/docs/topics/api/addons.rst
+++ b/docs/topics/api/addons.rst
@@ -296,11 +296,6 @@ Create
 This endpoint allows a submission of an upload to create a new add-on and setting other AMO metadata.
 
     .. note::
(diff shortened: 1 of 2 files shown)


diox (approved):
I redid a pass over the entire API to make sure there is no glaring problem and that we have good test coverage before enabling that in prod, haven't found any blockers. I filed https://github.com/mozilla/addons-server/issues/18779 as a follow-up.

PR #18768, opened 2022-02-04
Remove devhub/base_impala.html, re-implement missing dashboard CSS as a result

### LINKED ISSUE #18760: Remove devhub/base_impala.html
DevHub has layers upon layers of CSS, templates that override older templates... With subtle differences everywhere. Let's try to harmonize everything by getting rid of `devhub/base_impala.html`.

It's only used by the API keys page and the DevHub dashboard. The biggest issue is that the dashboard has a couple design elements specific to impala style: the sidebar on the right instead of the left, and the pagination/sort links/items. So we'll need to make some minor adjustments.

diox:
In the dashboard, there are a few minor alignment differences of 1px, as well as:
- A slightly darker shade of gray for text, to improve readability
- A slightly larger font size for the buttons
- A different pagination style to match the rest of devhub

Fixes #18760


src/olympia/addons/templatetags/jinja_helpers.py
```diff
@@ -4,9 +4,9 @@
 
 
 @library.global_function
-@library.render_with('addons/impala/listing/sorter.html')
+@library.render_with('devhub/includes/listing_header.html')
 @jinja2.pass_context
-def impala_addon_listing_header(
+def addon_listing_header(
```
eviljeff, line 9: should this helper be moved under /devhub too?

eviljeff:
> In the dashboard, there are a few minor alignment differences of 1px, as well as:

can haz screenshot?

diox:
Here are a few screenshots, but it's probably best if you try it yourself because I didn't replicate the same cropping every time...

**Before**

Screenshot from 2022-02-04 19-01-40
Screenshot from 2022-02-04 19-01-44
Screenshot from 2022-02-04 19-01-58

**After**

Screenshot from 2022-02-04 19-01-11
Screenshot from 2022-02-04 19-01-17
Screenshot from 2022-02-04 19-00-57

eviljeff (approved):
r+wc - just the comment about the location of the `addon_listing_header` jinja helpers.

PR #18769, opened 2022-02-04
add statsd pings for various methods of addon submission

### LINKED ISSUE #18484: Record a statsd ping to understand breakdown of listed/unlisted API/AMO Devhub submissions.
We'd like to know how many uploads are: unlisted or listed and whether they come via the API or through AMO's devhub.

To be able to know this we'd like to record a metric for this at the point a File is created from a FileUpload.

_Note, it would be worth also thinking about potential internal future use-cases of the API and being able to tell them apart from external API usage if possible._

eviljeff:
fixes #18484

### DIFF
--- a/src/olympia/devhub/tests/test_utils.py
+++ b/src/olympia/devhub/tests/test_utils.py
@@ -639,9 +639,7 @@ class TestCreateVersionForUpload(UploadMixin, TestCase):
         )
         assert self.mocks['parse_addon'].call_count == 0
         self.mocks['Version.from_upload'].assert_called()
-        self.mocks['statsd.incr'].assert_any_call(
-            'signing.submission.addon.listed'
-        ), self.mocks['statsd.incr'].mock_calls
+        self.mocks['statsd.incr'].assert_any_call('signing.submission.addon.listed')
 
     def test_statsd_logging_new_version(self):
         file_ = get_addon_file('valid_webextension.xpi')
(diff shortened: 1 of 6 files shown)


src/olympia/devhub/tests/test_utils.py
```diff
@@ -621,12 +621,43 @@ def setUp(self):
...
+        )
+        parsed_data = mock.Mock()
+        utils.create_version_for_upload(
+            empty_addon, upload, amo.RELEASE_CHANNEL_LISTED, parsed_data=parsed_data
+        )
+        assert self.mocks['parse_addon'].call_count == 0
+        self.mocks['Version.from_upload'].assert_called()
+        self.mocks['statsd.incr'].assert_any_call(
+            'signing.submission.addon.listed'
+        ), self.mocks['statsd.incr'].mock_calls
```
bobsilverberg, line 644: I don't understand what's happening here. What does this part after the comma do?

bobsilverberg (approved):
This looks good. I just have a question about a test.

r+wc

src/olympia/devhub/tests/test_utils.py
eviljeff: in a normal `assert`, `assert foo, baa` would print out `baa` in the console if the assert failed... but I've just realised that this isn't a normal assert so it's not going to do anything 😞

PR #18771, opened 2022-02-04
Attribute delayed rejection to the user who scheduled the rejection

### LINKED ISSUE #18403: Automatic rejection after a delayed rejection should be attributed to the user that sent the delayed rejection
Currently, a delayed rejection is attributed to the reviewer who sent it, but the final rejection when the time expires is attributed to the Mozilla user. I would like to have a way to make this connection without having to piece together activity log entries.

From my view it looks like it would be difficult to just set the original reviewer as the user taking the action, because this would cause the email sent out to note the original reviewer instead of the Mozilla user and the usual `Automatic rejection after grace period ended`. I'd be fine changing the rejection template accordingly in case a simple solution to this would be to not show the user if it is an automatic rejection (or the Mozilla user in general). Happy to provide copy for any template changes.

I'm also fine solving this in a completely different way.

bobsilverberg:
Fixes #18403


src/olympia/versions/tests/test_models.py
```diff
@@ -920,6 +920,45 @@ def test_pending_rejection_property(self):
...
+        assert version.pending_rejection_by is None
+        # Flag present, value is None (default): None.
+        flags = VersionReviewerFlags.objects.create(version=version)
+        assert flags.pending_rejection_by is None
+        assert version.pending_rejection_by is None
+        # Flag present, value is a user.
+        flags.update(pending_rejection_by=user)
+        assert version.pending_rejection_by == user
+
+    def test_pending_rejection_by_cleared_when_pending_rejection_cleared(self):
```
bobsilverberg, line 937: I considered going through and adding assertions to every test where we are clearing `pending_rejection` (e.g., [`test_delete_clear_pending_rejection`](https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/tests/test_models.py#L624)), but I think this test, for the `save()` method, should suffice.

src/olympia/versions/models.py
```diff
@@ -896,6 +904,14 @@ class VersionReviewerFlags(ModelBase):
     pending_rejection = models.DateTimeField(
         default=None, null=True, blank=True, db_index=True
     )
+    pending_rejection_by = models.ForeignKey(
+        UserProfile, null=True, on_delete=models.CASCADE
+    )
+
+    def save(self, *args, **kwargs):
+        if not self.pending_rejection:
+            self.pending_rejection_by = None
```
eviljeff, line 913: Ugh, I wish we had conditional database constraints so we could enforce "if pending_rejection is null, then pending_rejection_by has to be null", but mysql 😞

src/olympia/versions/tests/test_models.py
```diff
@@ -920,6 +920,45 @@ def test_pending_rejection_property(self):
...
+            pending_rejection=self.days_ago(1),
+            pending_rejection_by=user,
+        )
+        assert flags.pending_rejection
+        assert flags.pending_rejection_by == user
+        assert not flags.needs_human_review_by_mad
+
+        # Update, but do not clear pending_rejection. Both should remain.
+        flags.update(needs_human_review_by_mad=True)
+        flags.save()
```
eviljeff, line 952: mmm, but what if there is _only_ a `flags.update(...)`?  You don't need to `.save()` after an update - it writes to the database.

eviljeff (approved):
Just a question about the (normal) behaviour of using `.update` rather than (or in addition to) `.save`.

src/olympia/versions/tests/test_models.py
bobsilverberg: I'm glad you asked about this, because I don't quite understand it myself. I originally wrote the test with just `update` and no `save` afterwards, but the test was failing. It didn't seem that `update` was triggering the custom `save` method. This led me to believe that `update` doesn't save the record, but if it does then maybe there's a problem with what I've done here. I assumed that `save` was always called (for either a create or an update), but maybe it isn't? Or maybe it's related to this running in a test? Can you advise?

eviljeff: `update` ends up calling the queryset's `update` method, which doesn't call save on the individual instances.  You could override `update` on the `VersionFlags` model too, but that wouldn't address using any of the queryset update methods directly - which I see is typically how it's used.

Adding stuff to `save` is (unfortunately) not a complete solution - that'd be database constraints.  There is some limited support for constraints now in mysql 8.0.16, and apparently in django3.2 they're supported, so you could try adding a `CheckConstraint` to the model? https://docs.djangoproject.com/en/3.2/ref/models/constraints/#django.db.models.CheckConstraint

(And change everywhere in the code where we currently clear `pending_rejection`; clearing `pending_rejection_user` too.)

bobsilverberg: Ok, so my assumption that overriding `save` would "just work" and get fired any time an instance is changed and saved to the database is incorrect. I guess that means there is no point in having that code in `save`, and, as you say, we need to make sure we manually clear `pending_rejection_by` anytime we are also clearing `pending_rejection`. The database constraint would protect us from writing invalid data to the database, but we'd still need to manually clear both fields to avoid falling afoul of that constraint, so I'm not sure how much value it really adds.

@eviljeff  Is what I have written above accurate?

diox: To ensure your custom behavior would apply, you can put it in a [`pre_save`](https://docs.djangoproject.com/en/3.2/ref/signals/#pre-save) callback. `save()` triggers it, and our custom `update()` method does too, unless you really don't want it to (you can pass `_signal=False` to prevent that).

eviljeff: I forgot about pre_save - yeah, that'd be an alternative to changing all the places we currently set `pending_rejection`.  A constraint would still be ideal though.

diox: Yeah `pre_save` would make the code work, the constraint would ensure the database is always in a clean state. Both could be useful here.

bobsilverberg: Thanks @diox. I changed it to use `pre_save` and it works well. I can remove the `save` from my test and just doing an `update` triggers it. I will look into adding the constraint as well.

Do you both feel that the test I have for `test_pending_rejection_by_cleared_when_pending_rejection_cleared` is sufficient for this new `pre_save` code, or should I add assertions to other tests where `pending_rejection` is cleared to assert that `pending_rejection_by` is also cleared?

bobsilverberg:
Ok, I switched to using `pre_save` and added a `CheckConstraint`. Re-requesting review to make sure I did all of this properly.

eviljeff (approved):

PR #18772, opened 2022-02-04
correctly parse changeTime sent in fxa webhook payload

### LINKED ISSUE #18770: Fix timestamp parsing with fxa webhook events
For some fxa webhook events a `changeTime` property is sent in the payload https://github.com/mozilla/fxa/blob/main/packages/fxa-event-broker/README.md#example-password-change-event and we use that as the timestamp for the event (rather than `datetime.now()`).  Unfortunately it's not a unix timestamp (like the `iat` property), but a timestamp * 1000 - i.e. in millseconds rather than seconds, so python fails to parse it and raises an exception.  The exception is caught and logged as `year 54065 is out of range`

eviljeff:
fixes #18770

### DIFF
--- a/src/olympia/accounts/views.py
+++ b/src/olympia/accounts/views.py
@@ -2,8 +2,8 @@ import base64
 import binascii
 import functools
 import os
+import time
 
-from datetime import datetime
 from urllib.parse import quote_plus
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


src/olympia/accounts/views.py
```diff
@@ -852,7 +852,11 @@ def get_jwt_payload(self, request):
         return authenticated_jwt
 
     def process_event(self, uid, event_key, event_data):
-        timestamp = event_data.get('changeTime') or datetime.now().timestamp()
+        timestamp = (
```
diox, line 855: Not new, but do we want to `int()` that to be consistent ? `time.time()` (and `datetime.now().timestamp()`) returns a float, not an integer.

eviljeff: imo a float is better - we get millisecond precision when comparing the datetimes.  It's more consistent now we're doing `xxxx / 1000` as that'll be a float too.

diox (approved):

PR #18781, opened 2022-02-07
Add logging and statsd pings for source code submission using the new API

### LINKED ISSUE #18753: Add logging and statsd pings for source code submission using the new API
When source code is submitted through the new API (https://github.com/mozilla/addons-server/issues/9913) we should do logging and record statsd pings detailing the time the various steps took, like we implemented in version edit and source submission views.

See https://github.com/mozilla/addons-server/issues/18337 and its follow-ups https://github.com/mozilla/addons-server/issues/18461 and https://github.com/mozilla/addons-server/issues/18489 that made sure we only do this when there is source code being submitted.

bobsilverberg:
Fixes #18753

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -5,6 +5,8 @@ from django.urls import reverse
 from django_statsd.clients import statsd
 from rest_framework import exceptions, serializers
 
+import olympia.core.logger
+
 from olympia import amo
 from olympia.accounts.serializers import (
     BaseUserSerializer,
@@ -12,7 +14,7 @@ from olympia.accounts.serializers import (
 )
 from olympia.activity.utils import log_and_notify
 from olympia.amo.templatetags.jinja_helpers import absolutify
-from olympia.amo.utils import slug_validator
+from olympia.amo.utils import StopWatch, slug_validator
 from olympia.api.fields import (
     EmailTranslationField,
     LazyChoiceField,
@@ -54,6 +56,9 @@ from .fields import (
     VersionCompatabilityField,
 )
 from .models import Addon, AddonReviewerFlags, DeniedSlug, Preview, ReplacementAddon
+
+
+log = olympia.core.logger.getLogger('z.addons')
 
 
 class FileSerializer(serializers.ModelSerializer):
@@ -491,6 +496,18 @@ class DeveloperVersionSerializer(VersionSerializer):
             )
 
     def create(self, validated_data):
+        has_source = 'source' in validated_data and validated_data['source']
+        slug = self.addon.slug if self.addon else validated_data.get('addon').slug
+
+        if has_source:
+            timer = StopWatch('addons.serializers.developer_version_serializer.create.')
+            timer.start()
+            log.info(
+                'create, source upload received, addon.slug: %s',
+                slug,
+            )
+            timer.log_interval('1.source_received')
+
         upload = validated_data.get('upload')
         parsed_and_validated_data = {
             **self.parsed_data,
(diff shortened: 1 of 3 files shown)


src/olympia/addons/serializers.py
```diff
@@ -477,6 +482,18 @@ def _update_admin_review_flag_and_logging(self, version):
...
+        slug = self.addon.slug if self.addon else validated_data.get('addon').slug
+
+        if has_source:
+            timer = StopWatch('addons.serializers.developer_version_serializer.create.')
+            timer.start()
+            log.info(
+                'create, source upload received, addon.slug: %s',
+                slug,
+            )
+            timer.log_interval('1.source_received')
```
eviljeff, line 495: I suspect this is too late to capture the timing for the source code upload - it will already have been validated (including checking the file validity) by the point `create` is called so a bunch of the work will have been done.

bobsilverberg: Ok, so do we need to have the logging in the view?

diox: Yeah, this should probably be done in the view so that you can start the timer before calling `serializer.is_valid()`, then log an internal when you call `.save()`...

The view code is currently the [generic DRF implementation](https://github.com/encode/django-rest-framework/blob/master/rest_framework/mixins.py) though (`CreateModelMixin` and `UpdateModelMixin`), so you'd need to copy that over in our viewset... (And like in devhub, you'd want to check for `source` in `request.data`, before validation)

bobsilverberg: Thanks Mat. I thought it was using the generic DRF implementation, so I guess we need to override that for both create and update, unless there are insertion points where we can add logging. I can check for that but I'm guessing there are not.

bobsilverberg: I moved the logging code into the view.

src/olympia/addons/tests/test_views.py
```diff
@@ -1731,7 +1731,9 @@ def _generate_source_zip(
         source.seek(0)
         return source
 
-    def test_source_zip(self):
+    @mock.patch('olympia.addons.views.log')
```
eviljeff, line 1734: what about statsd logging? Should we test for that too?

src/olympia/addons/views.py
```diff
@@ -525,6 +526,96 @@ def get_queryset(self):
...
+                addon.slug,
+            )
+            timer.log_interval('4.data_saved')
+
+        headers = self.get_success_headers(serializer.data)
+        return Response(
+            serializer.data, status=status.HTTP_201_CREATED, headers=headers
+        )
+
+    def update(self, request, *args, **kwargs):
```
eviljeff, line 570: A thought I had while looking at these two chunks of code (that copy DRF source exactly, and are near duplicates in terms of logging - both generally bad things) is whether it would be better to wrap some of the individual functions that both `create` and `update` call instead.

e.g.
``` python
    def get_serializer(self, *args, **kwargs):
        has_source = request.data.get('source')
        if has_source:
            addon = self.get_addon_object()
            self.timer = StopWatch('addons.views.AddonVersionViewSet.update.')
            self.timer.start()
            log.info(
                'update, source upload received, addon.slug: %s, version.id: %s',
                addon.slug,
                instance.id,
            )
            self.timer.log_interval('1.source_received')
        serializer = super().get_serializer(*args, **kwargs)
        if has_source:
            log.info(
                'update, serializer loaded, addon.slug: %s, version.id: %s',
                addon.slug,
                instance.id,
            )
            self.timer.log_interval('2.serializer_loaded')
```
(If differentiating create from update, or logging the version id, was strictly necessary you'd need to tinker with the logging strings to be slightly different if there was an instance or not)

I'm really not sure if that'd actually be any better - the amount of code won't be far off because of the extra conditionals, but it'd all only be defined once and without copying the code in create or update.  Just throwing it out there 🤷‍♂️

eviljeff (approved):

src/olympia/addons/tests/test_views.py
bobsilverberg: When I was writing the patch to log and add timings to source code uploads via the views, I asked about writing tests for the `statsd` timings, as I wasn't sure how to do it, and I believe we decided it was unnecessary, so I didn't do it here either.

src/olympia/addons/views.py
bobsilverberg: I had originally through about trying not to override `create` and `update` and instead add timings to functions that they call, as you suggest, but it turned out that there were enough differences, and places where I needed to add timings to `create` and `update` themselves that this just seemed like a better route. I could do what you suggest above, and I could also override `serializer.is_valid` and `serializer.save` to add a timings to those, but I am worried that that will be an equivalent amount of code and will make everything more convoluted, especially if some of the code is in the view and some is in the serializer. 

So, while it's not 100% ideal, I do think what I've done in this patch is probably as good as the alternative. WDYT?

PR #18795, opened 2022-02-08
Remove impala templates, pages and css we don't use

### LINKED ISSUE #18792: Remove impala templates
Few traces of the legacy "impala" templates remain, they should get removed to limit duplication, reducing confusion and maintenance cost.

diox:
Also rework the few remaining things that depended on those templates and styles

Fixes #18792


diox:
This integrates statistics into devhub a little more. It looks like this with this patch:
Screenshot 2022-02-08 at 17-51-55 startle-crose-eagonfly Statistics Dashboard Add-ons for Firefox

(The initial reason was because statistics needed a new base template, and I thought this was a good occasion to roll them inside DevHub now that these pages are not public anymore)

src/olympia/templates/header_title.html
```diff
@@ -1,8 +1,10 @@
 <h1 class="site-title">
-{% macro heading(text, icon) -%}
+{% macro heading(app, text, icon) -%}
```
bobsilverberg, line 2: What purpose does this `app` serve? I don't see it referenced anywhere.

static/css/zamboni/stats.less
```diff
@@ -1,4 +1,4 @@
-@import 'lib';
+@import '../impala/lib';
```
bobsilverberg, line 1: Ok, so we're moving this file from `/impala/` to `/zamboni/`, but still importing a file from `/impala/lib`. Should we be looking at moving that also into `/zamboni/`, or is that a much bigger task?

bobsilverberg (approved):

r+wc

static/css/zamboni/stats.less
diox: It's imported in a bunch of stuff still and I wanted to stop before going too far. We can certainly do it later.

src/olympia/templates/header_title.html
diox: I copied that over from the duplicate file, but it seems useless yeah. I'll try removing it

PR #18802, opened 2022-02-09
add USER_DELETED activity class

### LINKED ISSUE #18773: Add DELETE_USER action
When a user gets deleted, we should log an action to the activity log. Note that the deletion request can originate either from AMO or from FxA, so we might actually want 2 separate actions, e.g. `DELETE_USER_AMO`, `DELETE_USER_FXA`.

The deletion through fxa is done via a task, but it would be good if we log entry would be attributed to the deleted user, so we have all activity for a user attributed to them.

eviljeff:
fixes #18773

fwiw, `USER_EDITED` doesn't seem to be used any longer - we should probably add those events to user profile edits (in the serializer) - or comment that it's unused.

### DIFF
--- a/src/olympia/accounts/tests/test_tasks.py
+++ b/src/olympia/accounts/tests/test_tasks.py
@@ -10,6 +10,7 @@ from olympia.accounts.tasks import (
     primary_email_change_event,
 )
 from olympia.accounts.tests.test_utils import totimestamp
+from olympia.activity.models import UserLog
 from olympia.amo.tests import addon_factory, collection_factory, TestCase, user_factory
 from olympia.bandwagon.models import Collection
 from olympia.ratings.models import Rating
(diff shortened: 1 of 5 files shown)


src/olympia/accounts/tests/test_tasks.py
```diff
@@ -80,6 +81,12 @@ def test_success_basic(self, delete_picture_mock):
         assert not Collection.objects.filter(id=collection.id).exists()
         assert not another_addon.ratings.all().exists()
         delete_picture_mock.assert_called()
+        # Two UserLog instances get created - one for request.user, one for the instance
```
diox, line 84: Uh ? Why ? I'd understand if the users were different (and even then, it could be confusing to admins looking at the data later, so we should make sure we can easily tell the 2 apart), but that's not the case here ?

src/olympia/accounts/tests/test_tasks.py
```diff
@@ -80,6 +81,12 @@ def test_success_basic(self, delete_picture_mock):
...
+        assert UserLog.objects.filter(user=self.user).count() == 2
+        alog = UserLog.objects.filter(user=self.user).first().activity_log
+        # For a user deleting their own account, they're just duplicates
+        assert alog == UserLog.objects.filter(user=self.user).last().activity_log
+        assert alog.arguments == [self.user]
```
diox, line 89: This is missing a test on the `action` (same in tests below)

src/olympia/accounts/tests/test_tasks.py
eviljeff: One `UserLog` instance is always created for the user adding the activity log; because this is related to a `UserProfile` object another `UserLog` is created for that (it'd be an `AddonLog` if it related to an `Addon`, etc).  

It's existing activity log functionality but I could change it?

eviljeff: https://github.com/mozilla/addons-server/issues/18811 to drop `UserLog` entirely

PR #18809, opened 2022-02-10
make blog post cron more efficient and check status code is 200

### LINKED ISSUE #18763: Developer Blog import cron is failing
https://sentry.prod.mozaws.net/operations/olympia-prod/issues/17705281/?referrer=github_plugin

```
TypeError: unhashable type: 'slice'
(2 additional frame(s) were not displayed)
...
  File "django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "olympia/amo/management/commands/cron.py", line 52, in handle
    getattr(module, name)(*args, **kwargs)
  File "olympia/devhub/cron.py", line 23, in update_blog_posts
    for item in items[:5]:
```

The actual issue is a 401 on the wordpress API though:
`GET https://blog.mozilla.org/addons/wp-json/wp/v2/posts [401]`

eviljeff:
fixes #18763


src/olympia/devhub/cron.py
```diff
@@ -14,17 +14,27 @@
 
 def update_blog_posts():
     """Update the blog post cache."""
-    items = requests.get(settings.DEVELOPER_BLOG_URL, timeout=10).json()
-    if not items:
+    response = requests.get(settings.DEVELOPER_BLOG_URL, timeout=10)
+    items = response.json()
```
diox, line 18: Maybe handle the case where the response is not json at all ? It would throw here.

eviljeff: it'd raise and the job would stop (without deleting the records) and get logged in sentry.  Is there something else we'd do though? Or it shouldn't raise?

diox: I guess it's fine to raise. Although maybe raising for the non-200 would make sense as well then

bobsilverberg (approved):

PR #18830, opened 2022-02-15
remove UserLog model

### LINKED ISSUE #18811: drop `UserLog` model/table
Our `ActivityLog` model includes a number of models that, in theory, lead to simpler code/queries when referencing a subset of those activity logs.  One of these models is `UserLog` which links `UserProfile`s to the activity log records (we have others such as `AddonLog`, `GroupLog`, etc.).  A `UserLog` instance is created for the user who created the `ActivityLog`; and in the case of an ActivityLog that is related to a `UserProfile`, another one is created for that.

But it's only actually used in one place for calculating [`last_known_activity_time`](https://github.com/mozilla/addons-server/blob/master/src/olympia/users/admin.py#L468) - and we can calculate that directly from `ActivityLog` instead.

eviljeff:
fixes #18811

### DIFF
--- a/src/olympia/users/admin.py
+++ b/src/olympia/users/admin.py
@@ -22,7 +22,7 @@ from django.utils.translation import gettext, gettext_lazy as _
 from olympia import amo
 from olympia.abuse.models import AbuseReport
 from olympia.access import acl
-from olympia.activity.models import ActivityLog, IPLog, UserLog
+from olympia.activity.models import ActivityLog, IPLog
 from olympia.addons.models import Addon, AddonUser
 from olympia.amo.admin import CommaSearchInAdminMixin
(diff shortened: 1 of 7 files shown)


src/olympia/users/admin.py
```diff
@@ -465,7 +465,9 @@ def last_known_activity_time(self, obj):
         # We sort by -created by default, so first() gives us the last one, or
         # None.
         user_log = (
```
bobsilverberg, line 467: Nit: Because we no longer have such a thing as a `UserLog`, should we rename this variable to just `log` or `activity_log`?

bobsilverberg (approved):
Looks good, thanks @eviljeff. I just had one nit. Also, I assume a separate patch will follow with the next push to include the migration for removing the table?

r+wc

PR #18848, opened 2022-02-17
fix statsd logging for new addon submissions via signing api

eviljeff:
Fixes the statsd logging for the signing api add-on submissions - the test whether the addon had any versions or not was too later (_after_ we created the version instance).  Wasn't picked up by testing because we mocked the function that would have created the database entry.


eviljeff:
> Looks good, but if this was not picked up by a test, shouldn't we also be adding a test that would have picked this up?
> 
> r+wc

meh, we wouldn't have added a test if I'd written it correctly in the first place, and it's only statsd pings after all.

PR #18875, opened 2022-02-21
insert guid to manifest before signing

### LINKED ISSUE #11559: Insert add-on ID into manifest if it's missing
One of the issues that came up during Armagadd-on is that it's problematic for add-ons to not have an ID set in their manifest. A broken signature means there's no way for Firefox to figure out their ID.

AMO should add the [ID in the manifest](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings) if the uploaded add-on doesn't have it, before signing.

The reason we don't want to require developers to add the ID themselves is that we don't want to make it harder for them to maintain a Firefox and Chrome version of their add-on. In some cases it's possible to submit the same code for both browsers, so we should try to keep things that way if possible.

/cc @muffinresearch @ddurst

eviljeff:
fixes #11559

This tries to be fairly flexible about what keys are used - if the manifest is using `"applications": {"gecko": ...}` it uses that object to insert the id into, knowing Firefox's logic to only look in there if `browser_specific_settings` isn't used.


diox (commented):
mmmm I wonder if just before signing isn't a bit too late. I was thinking we'd have it much earlier, when repacking the upload (*), so that we could do stuff depending on the guid before, like in scanner rules. /cc @wagnerand for input.

(*) which might require some refactor depending on the submission method used

eviljeff:
> mmmm I wonder if just before signing isn't a bit too late. I was thinking we'd have it much earlier, when repacking the upload (*), so that we could do stuff depending on the guid before, like in scanner rules. /cc @wagnerand for input.
> 
> (*) which might require some refactor depending on the submission method used

That was my original idea, hence https://github.com/mozilla/addons-server/pull/18874, but it's not practical.  
- We don't normally know if the upload is a new version or a new add-on at the point we do the manifest normalization (only with the obsolete signing api)
- So we could implement it there _just_ for the signing API, but then it'd need to be replicated in another place devhub and the new addon submission API (possibly differently in each place - so 3 different integration points)
- For devhub/submission api, if we do it at the point we know it's a new version, and what the guid should be, then where/how?  Might be unexpectedly slow to do it in the request/response cycle; a task _could_ be triggered to do the rewrite, at the risk of introducing some race conditions with other things that could be happening (xpis being downloaded by the developer; auto signing; plus caching.)

So I went with the cleanest solution, (that happens to fit exactly the wording of the issue "AMO should add the ID in the manifest if the uploaded add-on doesn't have it, before signing")

wagnerand:
This could be a problem for submissions that never get signed. @eviljeff are you saying it's not possible to do this earlier?

eviljeff:
> This could be a problem for submissions that never get signed. @eviljeff are you saying it's not possible to do this earlier?

It depends.  And that is part of the problem - there are many different paths for xpis to come through (vs. a single place we sign the xpi).  

For the old signing api, we do know up front what the guid is - and allow it to be specified as a parameter - so it's possible the manifest could be rewritten during those tasks.  

For devhub, at the point the upload is handled and sent for async validation we don't pass along any details of the add-on or if it's a new version.  With the URL structure of the pages it's possible (for some currently unknown level of difficultly) to pass that down to the relevant tasks so I would assume a rewrite would be possible.  

For new submission api, we (in the current design) don't connect the upload to an add-on until it's passed validation and ready to create so it's not possible.  There would need to be new upload endpoints added for new versions vs. new add-ons.

For devhub and the submission api, in all cases for new add-ons without a guid already in the manifest it'd just be a randomly generated string anyway.  And for new versions of existing add-ons the non-edge case is that the guid either wasn't specified when the add-on was created, so it's random, or was in the manifest and is still in the manifest for the new version.

The ask for rewrite the manifest **during upload validation** is quite different than described in #11559 really and would be an entirely different PR.  If #11559 is no longer valid, it should be updated with the current requirements (or closed and re-filed)

eviljeff:
the existing repack is done before we run the linter.  It's roughly:
1) repack of the xpi
2) linter
3) yara&customs&wat&mad

Then if they pass the upload is considered valid and in the case of devhub and addons submission api, it takes a developer action to create an add-on and/or version with that upload, which is when we know about any existing guid that should be used.  
(The signing API does things slightly differently and creates an add-on instance first - why we end up with so many empty add-on instances - and then creates the version immediately after the upload is valid.)

Some time later (minutes to hours for auto signing; longer for manual reviews) the signing task starts which is when the code in this PR would do it's thing.  If there is a requirement to rewrite the manifest before the signing task then it either needs to happen somewhere in the upload validation (repack; linter; yara,etc), or afterwards when a version is created from the upload.

src/olympia/lib/crypto/signing.py
```diff
@@ -60,15 +62,53 @@ def use_promoted_signer(file_obj, promo_group):
     )
 
 
+def add_guid(file_obj):
+    with storage.open(file_obj.current_file_path) as fobj:
+        # Get the file data and add the guid to the manifest if waffle switch is enabled
+        if waffle.switch_is_active('add-guid-to-manifest'):
+            with zipfile.ZipFile(fobj, mode='r') as existing_zip:
+                manifest_json = json.loads(existing_zip.read('manifest.json'))
```
bobsilverberg, line 70: What happens if `manifest.json` does not exist in the `existing_zip`? Or are we guaranteed that this code will never run if that is the case?

bobsilverberg (approved):
Looks good. I just have one question.

r+wc

src/olympia/lib/crypto/signing.py
eviljeff: It wouldn't have gotten through the linter and other validation with a `manifest.json`, and if it did then I think we're okay erroring - autograph isn't going to sign it anyway.

PR #18882, opened 2022-02-22
Remove extra logging when multiple user profiles are found

### LINKED ISSUE #18881: Remove extra logging when multiple user profiles are found
https://github.com/mozilla/addons-server/blob/fc77fd9b7b75055d685bfc65f2a0b82018f5b729/src/olympia/accounts/views.py#L147-L152

We are both raising and logging here, which is redundant. Let's remove the extra logging (which is also dangerous because it logs PII)

diox:
Fixes #18881

### DIFF
--- a/src/olympia/accounts/views.py
+++ b/src/olympia/accounts/views.py
@@ -144,12 +144,9 @@ def find_user(identity):
         return user
     except UserProfile.DoesNotExist:
         return None
-    except UserProfile.MultipleObjectsReturned:
-        # This shouldn't happen, so let it raise.
-        log.error(
-            'Found multiple users for %s and %s', identity['email'], identity['uid']
-        )
-        raise
+    # UserProfile.MultipleObjectsReturned might be raised here. This is
+    # expected, we don't know what to do in this case so we let the exception
+    # happen loudly.
 
 
 def register_user(identity):


src/olympia/accounts/views.py
```diff
@@ -144,12 +144,9 @@ def find_user(identity):
...
         return None
-    except UserProfile.MultipleObjectsReturned:
-        # This shouldn't happen, so let it raise.
-        log.error(
-            'Found multiple users for %s and %s', identity['email'], identity['uid']
-        )
-        raise
+    # UserProfile.MultipleObjectsReturned might be raised here. This is
+    # expected, we don't know what to do in this case so we let the exception
+    # happen loudly.
```
eviljeff, line 149: Might it still be better to raise a custom error/exception with useful information in the string?

eviljeff (approved):
r+wc - the exception could contain PII anyway?

src/olympia/accounts/views.py
wagnerand: +1. We are still getting inquiries about this from affected developers occasionally. Could we include the two user ids here?

diox:
Fully scrubbing the PII is part of a larger effort in https://github.com/mozilla/addons-server/issues/18885 - notably, we should scrub email in local variables (dictionaries values for `email` key or straight `email` variable)

src/olympia/accounts/views.py
diox: The traceback already contains this:
```clone: <BaseQuerySet [<UserProfile: 1234567: Foo>, <UserProfile: 1234568: Bar>]>```

wagnerand: Ah, thanks, that wfm.

Issue #18885, opened 2022-02-22
Configure Sentry to filter out PII information


diox:
We currently have a custom on premises Sentry, and we have configured the client to send full user information, through the `send_default_pii` key in `SENTRY_CONFIG` setting.

As we're moving to a SaaS external Sentry instance, we need to turn that option off to avoid sharing PII information with the outside world.

We do want to keep user IDs in the tracebacks for logged in users to help debugging however, so we might need to include it ourselves through before_send if `send_default_py=False` doesn't include that.

In addition, we should ensure we don't include PII in logging messages, even for exceptions.

diox:
https://github.com/willkg/kent/ might be useful for local testing

eviljeff:
Is there a list of the data we store/handle than could be regarded as PII?

diox:
> Is there a list of the data we store/handle than could be regarded as PII?

I've asked, but I think that doesn't block us from deploying my PR

eviljeff:
> > Is there a list of the data we store/handle than could be regarded as PII?
> 
> I've asked, but I think that doesn't block us from deploying my PR

You have a list: https://github.com/mozilla/addons-server/pull/18893/files#diff-9349f4d8607c31c4ec557e0f3fa841a30c2fcf98ceab6ed8c85e7b9d598833d3R1531 😛

diox:
Yeah, that was my best guess by looking at a basic event

PR #18893, opened 2022-02-23
Scrub sensitive data before sending to Sentry

diox:
Fixes https://github.com/mozilla/addons-server/issues/18885

We might want to go even further, but that should be enough for now.


src/olympia/lib/settings_base.py
```diff
@@ -1496,15 +1496,39 @@ def get_sentry_release():
...
+                data[key] = _scrub_sensitive_data_recursively(value, name=key)
+        elif isinstance(data, str) and name in SENTRY_SENSITIVE_FIELDS:
+            data = '*** redacted ***'
+        return data
+
+    try:
+        event = _scrub_sensitive_data_recursively(event)
+    except Exception:
+        pass
+    return event
```
eviljeff, line 1514: I know this isn't a new thing, but I don't like how we have random functions in the middle of the settings file.  Perhaps they should be at the top at least?

src/olympia/lib/settings_base.py
```diff
@@ -1496,15 +1496,39 @@ def get_sentry_release():
...
 
 
-# This is the DSN to the Sentry service.
+def scrub_sensitive_data(event, hint):
+    def _scrub_sensitive_data_recursively(data, name=None):
+        # This only works with lists or dicts but we shouldn't need anything else.
+        if isinstance(data, (list, dict)):
+            items = data.items() if isinstance(data, dict) else enumerate(data)
+            for key, value in items:
+                data[key] = _scrub_sensitive_data_recursively(value, name=key)
```
eviljeff, line 1505: mmm, for `list`s `key` would just be the list index integer, which is never going to match `SENTRY_SENSITIVE_FIELDS`. So should we be passing down the list name instead?  

Example:
```python
{"email": ["a@foo.com", "b@baa.com"]}
```
In the code above wouldn't redact any of the email addresses because `name` would be `0` and `1`.

eviljeff (approved):

src/olympia/lib/settings_base.py
diox: To be fair if we had something like that we'd likely call the key `emails` and it would never match `SENTRY_SENSITIVE_FIELDS` in its current state either :stuck_out_tongue: 

I think the logic breaks down quite quickly for cases like that and it's always going to be a best effort kind of thing. If we had `{'email': [{'value': 'a@foo.com'}, {'value': 'b@foo.com'}]}` for instance what should we do...

eviljeff: We could make the match to `SENTRY_SENSITIVE_FIELDS` "sticky" and redact all values under it... that may be a bit too extreme though (and very prone to false positives)

src/olympia/lib/settings_base.py
diox: I agree in theory, but it references `SENTRY_SENSITIVE_FIELDS` which is defined below with the rest of the sentry stuff. So I'm not sure moving it to the top is better...

PR #18901, opened 2022-02-24
Downgraded logging level to warning

### LINKED ISSUE #18880: Downgrade 'No MessageId in message, aborting.' logging from error to warning
Follow-up for https://github.com/mozilla/addons-server/issues/18764

Currently, when we process an incoming email for activitylog replies, we log an error if we can't find a message id:
https://github.com/mozilla/addons-server/blob/fc77fd9b7b75055d685bfc65f2a0b82018f5b729/src/olympia/activity/tasks.py#L23-L25

This results in the following traceback in sentry: https://sentry.prod.mozaws.net/operations/olympia-prod/issues/8507150/?query=is%3Aunresolved

We don't need to error there, for the most part this is going to be just noise. A warning would be just fine.

NegiAkash890:
Fixes #18880 
Downgraded logging level to warning


diox:
Hi,

Please run `black` on your code to reformat it (line is now too long)

diox (approved):
Thanks!

PR #18902, opened 2022-02-24
drop all support for internal WebTokenAuthentication

### LINKED ISSUE #18854: remove WebTokenAuthentication support
#18475 added a new internal authentication method, and once frontend (https://github.com/mozilla/addons-frontend/issues/11110) and code-manager (https://github.com/mozilla/addons-code-manager/issues/2156) support it we can remove it as a method of authentication, and stop generating the `frontend_auth_token` token during login.  

There are no other supported webapps that use this internal method of authentication (we provide JWT auth keys for external developers to use); any other random tool that uses the method can switch easily to use SessionIDAuthenticatoin.

eviljeff:
fixes #18854 - it ends up being much wider than I imagined.  

In addition to deleting `APITestClientWebToken`, `WebTokenAuthentication` and the code in the login view that added the frontend cookie, there was quite a lot more:
- updating docs that mentioned internal auth, including v3/v4 docs.  I considered just removing the auth_internal pages but we do link to the sections in a few other places, and there are a number of documented apis under v3 and v4 that are internal auth only.
- changing devhub & reviewer tools use of the API - including the JS that does the actual xhr.
- updating tests to use a different test auth class - on the whole very straightforward, apart from `TestSessionView` - debugging that lead me to define `AUTHENTICATION_BACKENDS` in settings_base rather than as a class property of `UserProfile` because the real-ish authenticate didn't set the correct `_auth_user_backend` key in the session.


diox:
> updating tests to use a different test auth class - on the whole very straightforward, apart from `TestSessionView` - debugging that lead me to define `AUTHENTICATION_BACKENDS` in settings_base rather than as a class property of `UserProfile` because the real-ish authenticate didn't set the correct `_auth_user_backend` key in the session.

I'm confused by this. Isn't this backend the default ? Furthermore, tests have a specific backend set in `settings_tests.py`. So why is setting `AUTHENTICATION_BACKENDS` in `settings_base` necessary? Tests seem to work fine for me without doing that, and in a shell `settings.AUTHENTICATION_BACKENDS` returns `['django.contrib.auth.backends.ModelBackend']` as I'd expect since it's the default.

eviljeff:
> > updating tests to use a different test auth class - on the whole very straightforward, apart from `TestSessionView` - debugging that lead me to define `AUTHENTICATION_BACKENDS` in settings_base rather than as a class property of `UserProfile` because the real-ish authenticate didn't set the correct `_auth_user_backend` key in the session.
> 
> I'm confused by this. Isn't this backend the default ? Furthermore, tests have a specific backend set in `settings_tests.py`. So why is setting `AUTHENTICATION_BACKENDS` in `settings_base` necessary? Tests seem to work fine for me without doing that, and in a shell `settings.AUTHENTICATION_BACKENDS` returns `['django.contrib.auth.backends.ModelBackend']` as I'd expect since it's the default.

maybe specifying it in `AUTHENTICATION_BACKENDS` isn't necessary if it's already the default, but _not_ specifying it in `UserProfile` is.  With that property set login was always using the auth backend defined in `UserProfile` even in tests - we don't seem to do a real login call and then try to use that session to do follow-up calls anywhere else than `SessionView`.

diox:
:+1: I get it now.

Related to this... I think we should stop using the custom auth backend in tests. It's necessary because we do a lot of `client.login(email=email)` in tests, but nowadays Django has [`force_login()`](https://docs.djangoproject.com/en/2.2/topics/testing/tools/#django.test.Client.force_login) for this use case. I'll file an issue.

diox (approved):

PR #18906, opened 2022-02-25
Change git storage directory structure (behind a waffle switch, with a migration command)

### LINKED ISSUE #18896: Change file structure for add-on git extractions
Each add-on is extracted and stored through git on EFS for reviewing purposes. The directory structure is: `/git-storage/<last addon_id digit>/<last 2 addon_id digits>/<addon_id>/`.

With only 2 intermediary levels of 10 possibilities each (the second level has 2 digits but one of them doesn't change for a given first level), the second level directories can have tens of thousands of subdirectories, which make listing them almost impossible.

We want to move to 3 intermediary levels, much like what we'll have to do for add-ons soon in https://github.com/mozilla/addons-server/issues/11510. The new structure should be:
`/git-storage/<last addon_id digit>/<last 2 addon_id digits>/<last 3 addon_id digits>/<addon_id>/`.

We'd want to avoid re-extracting all add-ons from scratch, so we should have a migration plan.

diox:
Fixes #18896


src/olympia/files/utils.py
```diff
@@ -82,19 +82,23 @@ def get_filepath(fileorpath):
...
-def id_to_path(pk):
+def id_to_path(pk, depth=2):
     """
     Generate a path from an id, to distribute folders in the file system.
     1 => 1/1/1
     12 => 2/12/12
     123456 => 6/56/123456
+
+    depth can be specified through the corresponding parameter, defaults to 2,
+    levels deep, minimum is 1. depth=3 would return 6/56/456/123456 for 123456.
```
eviljeff, line 93: With only 10 folders at each level we do end up with many sparsely populated interim folders, followed by a lot in the end level.  

For example, if we have 1 million end folders (add-on ids are trending upwards towards that): then for `depth=2` it would be 10 x 10 x 10000, and for `depth=3` we would have 10 x 10 x 10 x 1000.  

But if we group by 2 digits then its 123456 => 56/3456/123456 for `depth=2` (100 x 100 x 100) which is manageable at all the levels, and higher depths wouldn't be needed until we're beyond 10 million (so just fix it at `depth=2`).

src/olympia/git/management/commands/migrate_git_storage_to_new_structure.py
```diff
@@ -0,0 +1,131 @@
...
+                # id is 60, let's add it manually.
+                zrange = list(range(0, 10))
+                if x == 0 and y == 6:
+                    zrange.append(60)
+                for z in zrange:
+                    path = os.path.join(
+                        new_file_storage_path,
+                        f'{x}',
+                        f'{y}{x}',
+                        f'{z}{y}{x}' if z < 10 else f'{z}',
```
eviljeff, line 83: this is special casing, _just_ for id=60, right?  why not just create it with `os.makedirs(os.path.join(new_file_storage_path, '0', '60', '60'))` directly first?

eviljeff (approved):
r+ nit: maybe a bit late after you've written it all but we should be using [pathlib](https://docs.python.org/3/library/pathlib.html) these days for fie path manipulation really.

src/olympia/activity/tasks.py
```diff
@@ -21,7 +21,9 @@ def process_email(message, spam_rating, **kwargs):
             if header.get('Name', '').lower() == 'message-id':
                 msg_id = header.get('Value')
     if not msg_id:
-        log.error('No MessageId in message, aborting.', extra={'message_obj': message})
+        log.warning(
+            'No MessageId in message, aborting.', extra={'message_obj': message}
+        )
```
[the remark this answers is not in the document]
diox: Ugh yes, I'll remove, it's a separate PR I was testing, didn't notice which branch I was and then it got mistakenly pushed with the rest.

src/olympia/git/management/commands/migrate_git_storage_to_new_structure.py
diox: In hindsight... Yeah that's better :) I didn't know how many special cases like that I'd have to handle when I started, so I figured I would handle them all in a generic way, but since we have only one... I'll change it

src/olympia/files/utils.py
diox: Yeah, I thought about that and then decided to stick to the original plan because it was easier, but I'm going to look into that. My main concern is around low ids - there is only one edge case with the implementation I went for, there would be a bit more with that approach, but maybe that's not a big deal.

src/olympia/git/management/commands/migrate_git_storage_to_new_structure.py
diox: Special casing is gone with the ne 100x100 approach in https://github.com/mozilla/addons-server/pull/18906/commits/26e45049b229182ed35217fae485da6123137cd2

src/olympia/files/utils.py
```diff
@@ -87,26 +87,38 @@ def id_to_path(pk, breadth=1):
...
-    It defaults to 1, creating 10 * 10 * x directories. A value of 2 would
-    create 100 * 100 * x directories, resulting in:
-    1 => 1/1/1
-    12 => 12/12/12
-    123 => 23/123/123
+
+    At breadth=2, there should be 100 * 100 * x directories, resulting in:
+    1 => 01/0001/1
+    12 => 12/0012/12
+    123 => 23/0123/0123
```
eviljeff, line 104: `23/0123/123`, no?

src/olympia/files/utils.py
```diff
@@ -87,26 +87,38 @@ def id_to_path(pk, breadth=1):
...
+        path = [pk[-1]]
+        if len(pk) >= 2:
+            path.append(pk[-2:])
+        else:
+            path.append(pk)
     else:
-        path.append(pk)
+        # 'New' algorithm, padded, uses more digits depending on breadth.
+        padded_pk = pk.zfill(2 * breadth)
+        path = [padded_pk[-breadth:], padded_pk[-breadth * 2 :]]
```
eviljeff, line 120: I understand if you want to differentiate the "original" algorithm separate for safety, but the new one would work the same as the old for `breadth=1` too, apart from with single digit add-on ids (which we don't have).

eviljeff (approved):

src/olympia/files/utils.py
diox: Correct !

src/olympia/files/utils.py
diox: Oh, yeah, let me simplify then. https://github.com/mozilla/addons-server/pull/18906/commits/fba9715b94cf6791238f0b222b1d331ea5b54e32

PR #18918, opened 2022-03-01
Remove ip_address from sentry events completely instead of redacting it

### LINKED ISSUE #18917: Remove `user.ip_address` key from sentry events
In https://github.com/mozilla/addons-server/issues/18885 we redacted anything in sentry events that look like PII from the name. For `user.ip_address` specifically, because of its special nature, we should remove it from the event entirely - sentry is going to discard it anyway because `** redacted **` is not a valid IP address :)

diox:
Fixes #18917

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1515,6 +1515,8 @@ def scrub_sensitive_data(event, hint):
         event = _scrub_sensitive_data_recursively(event)
     except Exception:
         pass
+    if 'ip_address' in event.get('payload', {}).get('user', {}):
+        event['payload']['user'].pop('ip_address')
     return event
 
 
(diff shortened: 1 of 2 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -1515,6 +1515,8 @@ def _scrub_sensitive_data_recursively(data, name=None):
         event = _scrub_sensitive_data_recursively(event)
     except Exception:
         pass
+    if 'ip_address' in event.get('payload', {}).get('user', {}):
+        event['payload']['user'].pop('ip_address')
```
eviljeff, line 1519: this wouldn't strip out the other ip addresses fields - `'remote_addr', 'remoteaddresschain', 'x-forwarded-for'` - so I guess we're okay with that?

eviljeff (approved):

PR #18925, opened 2022-03-02
Add more version comparison tests

### LINKED ISSUE #17574: Check support for Firefox versions over 100
As Firefox versions > 100 are not too far off it would be good to check if there are any issues from using Firefox versions that are >= 100 and ensure we have tests covering those areas. Feel free to add details to this bug of what those areas are, or create new bugs as needed.

Hopefully this will be a non-issue but it wouldn't hurt to be sure.

diox:
Fixes #17574

Tests are cheap.

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -2192,6 +2192,26 @@ class TestVersionViewSetCreate(UploadMixin, SubmitSourceMixin, TestCase):
         }
         assert list(version.compatible_apps.keys()) == [amo.FIREFOX, amo.ANDROID]
 
+    def test_compatibility_dict_100(self):
+        response = self.client.post(
+            self.url,
+            data={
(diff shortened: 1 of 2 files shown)


eviljeff (approved):
r+ but are there more places to test?  Seems a little too easy to add a couple of lines to this one file.

diox:
There is already https://github.com/mozilla/addons-server/blob/ce0a50f72001a729b7d6d0762105ffc6827cd0b2/src/olympia/files/tests/test_utils.py#L256 (and  https://github.com/mozilla/addons-server/blob/ce0a50f72001a729b7d6d0762105ffc6827cd0b2/src/olympia/files/tests/test_utils.py#L280 in a way, since `amo.DEFAULT_WEBEXT_MIN_VERSION_MV3_FIREFOX` is `100.0`)

I'll add an int something in `addons/tests/test_views.py` just to be sure.

PR #18935, opened 2022-03-03
Update CircleCI executor config to use an explicitly supported ubuntu image

### LINKED ISSUE #18843: Move away from `machine: true` in CircleCI config
We received the following email from CircleCI:
> Hello,
>
> We are deprecating Ubuntu 14.04-based machine images on CircleCI in preparation for an EOL on Tuesday, May 31, 2022 to ensure your builds remain secure. For a detailed overview of how this will affect your workflow, [read the blog article here](https://go.circleci.com/NDg1LVpNSC02MjYAAAGCndaubP1lrB_xKEs9K3iqzYYxUBqxzrHoMsjM7oN8hTLoyfIucsihEKChdSM-NVzcCvkPtwk=).
>
> We will also be conducting temporary brownouts on Tuesday, March 29, 2022, and again on Tuesday, April 26, 2022 during which these images will be unavailable.
> 
> We are contacting you because one or more of your projects has a job that either:
>
>    does not specify an image (uses machine: true in config)
>    explicitly uses an Ubuntu 14.04-based image
>
> 
> If you have specified an Ubuntu 14.04-based image or you are using machine: true in your config file, please see our [migration guide](https://go.circleci.com/NDg1LVpNSC02MjYAAAGCndaubAkpQkmFHfDYu-fp4EmnV8zWBJhxf8MPSWnVgraetk-HLjJopo1dxVGRR4O4citj3UQ=) to upgrade to a newer version of Ubuntu image in order to avoid any service disruption during the brownout & subsequent EOL.
>
> We will also be releasing a CircleCI Ubuntu 22.04 image on April 22nd offering the flexibility to upgrade to the latest LTS version of Ubuntu image before we remove older versions permanently. A beta version of the image will be available March 21st.

So we need to update our config to avoid disruption. We seem to be using `machine: true` only for the release build jobs, do we even need a machine executor for that or could we get away with a docker one ? And in any case, which image should we pick ?

(We have the same problem in other repos, but we can focus on fixing one and then applying the same fix elsewhere)

diox:
Fixes #18843

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -344,7 +344,8 @@ references:
             AUTOGRAPH_CONFIG: *autograph_config
 
   defaults-release: &defaults-release
-    machine: true
+    machine:
+      - image: ubuntu-2004:202201-02
     working_directory: &working-directory
 
 commands:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -345,7 +345,7 @@ references:
 
   defaults-release: &defaults-release
     machine:
-      - image: ubuntu-2004:202201-02
+      image: ubuntu-2004:202201-02
     working_directory: &working-directory
 
 commands:


diox:
Note: I went with a specific tag instead of `current` to avoid surprises: I don't want a build for stage to suddenly use a build different from the one used for dev.

bobsilverberg (approved):

PR #18949, opened 2022-03-08
Use sub directories to store sitemap xml files

### LINKED ISSUE #18895: Move sitemap files in subdirectories on filesystem
Our sitemap is backed by XML files generated by a cron every day and stored on EFS. The files are all stored inside the same `sitemaps/` directory. That directory contains over 17K files in prod right now, which is making basic operations like `ls` difficult.

We need to update the code to spread the files in sub-directories, with each sub-directory ideally not containing more than 1000 files.

Changing the code in this way could break the sitemap temporarily if we deploy code using the files with the new structure while the files haven't been regenerated yet in the new location. Generating the full sitemap takes ~45 minutes in production, so we could simply accept a 45 minutes potential breakage and regenerate the sitemap right away after deploy, or be smarter about backwards-compatibility, depending on how easy it is.

diox:
Fixes #18895

### DIFF
--- a/src/olympia/amo/sitemap.py
+++ b/src/olympia/amo/sitemap.py
@@ -464,7 +464,7 @@ def get_sitemap_path(section, app, page=1):
         endpath = id_to_path(page)
     return os.path.join(
         settings.SITEMAP_STORAGE_PATH,
-        section if section else '',
-        app if app else '',
+        section or '',
+        app or '',
         f'{endpath}.xml',
     )
--- a/src/olympia/amo/sitemap.py
+++ b/src/olympia/amo/sitemap.py
@@ -18,6 +18,7 @@ from olympia.amo.templatetags.jinja_helpers import absolutify
 from olympia.constants.categories import CATEGORIES
 from olympia.constants.promoted import RECOMMENDED
 from olympia.bandwagon.models import Collection
+from olympia.files.utils import id_to_path
 from olympia.promoted.models import PromotedAddon
 from olympia.tags.models import AddonTag, Tag
 from olympia.users.models import UserProfile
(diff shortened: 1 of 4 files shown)


src/olympia/amo/sitemap.py
```diff
@@ -454,11 +455,16 @@ def render_index_xml(sitemaps):
 
 
 def get_sitemap_path(section, app, page=1):
+    if section is None or app is None:
+        # If we don't have a section or app, we don't need a complex directory
+        # structure and we can call the first page 'sitemap' for convenience
+        # (it's likely going to be the only page).
+        endpath = str(page) if page != 1 else 'sitemap'
+    else:
+        endpath = id_to_path(page)
```
eviljeff, line 464: If we're already creating subdirectories for section and app do we even need a further breakdown into subdirectories?
I.e. what's the maximum number of files a given section/app has?  (which I'm guessing is `/addons/firefox/`; I should know this already but forgot)

src/olympia/amo/sitemap.py
```diff
@@ -454,11 +455,16 @@ def render_index_xml(sitemaps):
...
     return os.path.join(
         settings.SITEMAP_STORAGE_PATH,
-        'sitemap'
-        + (f'-{section}' if section else '')
-        + (f'-{app}' if app else '')
-        + (f'-{page}' if page != 1 else '')
-        + '.xml',
+        section if section else '',
+        app if app else '',
```
eviljeff, line 468:
```suggestion
        section or '',
        app or '',
```

src/olympia/amo/sitemap.py
diox: I had the same thought but I looked it up and it's ~ 5000 for addons/firefox. Which is bordeline manageable, but wouldn't leave space to grow.

eviljeff (approved):

PR #18956, opened 2022-03-09
Replace circleci/mysql image with the official mysql docker image

### LINKED ISSUE #18951: Update CircleCI docker images
Follow-up from https://github.com/mozilla/addons-server/issues/18843 we're also using on addons-server some legacy convenience images:
https://github.com/mozilla/addons-server/blob/fde1bb0f1dfce82a6a9459e68ab9f566f44bf8e1/.circleci/config.yml#L11
https://github.com/mozilla/addons-server/blob/fde1bb0f1dfce82a6a9459e68ab9f566f44bf8e1/.circleci/config.yml#L288

Those are [deprecated](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034) and we need to update them to supported images. The modern CircleCI convenience images have the `cimg/` prefix (but we need to check if the ones we need exist or need to be replaced with some upstream image instead)

diox:
Downgrading (!) the version too to match prod.

Fixes #18951

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -290,6 +290,7 @@ references:
 
   mysql-config: &mysql-config
     environment:
+      MYSQL_ALLOW_EMPTY_PASSWORD: true
       MYSQL_DATABASE: olympia
 
   defaults: &defaults
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,9 +6,10 @@
 version: 2.1
 
 orbs:
-  # This is needed to be able to install Node.js in the primary container,
-  # which provides python only.
-  node: circleci/node@4.7.0
+  # This is needed to be able to install Node.js 14 in the primary container,
+  # which provides python only (the convenience cimg/python:3.9-node provides
+  # node 16).
+  node: circleci/node@5.0.0
 
 references:
   # We declare the autograph configuration here to be able to fully leverage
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -6,9 +6,10 @@
 version: 2.1
 
 orbs:
-  # This is needed to be able to install Node.js in the primary container,
-  # which provides python only.
-  node: circleci/node@4.7.0
+  # This is needed to be able to install Node.js 14 in the primary container,
+  # which provides python only (the convenience cimg/python:3.9-node provides
+  # node 16).
+  node: circleci/node@5.0.0
```
diox, line 12: This is just an orb, so it doesn't need to change. I updated the version but it shouldn't be any different, we're explicitly installing a specific nodejs version.

eviljeff: _could_ it change though?  Using the deprecated images seems undesirable, even if it's not an immediate problem.

diox: It's (confusingly) not an image, just a (very fancy) bit of config. It isn't deprecated like the `circleci/` images are, and doesn't have a replacement.

eviljeff: Oh! got it

bobsilverberg (approved):
Seems fine to me. I assume it works or CI would have failed.

PR #18960, opened 2022-03-10
Force elasticsearch-py to send queries with body as POST

diox:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1758026

### DIFF
--- a/src/olympia/amo/search.py
+++ b/src/olympia/amo/search.py
@@ -14,7 +14,7 @@ DEFAULT_HOSTS = ['localhost:9200']
 DEFAULT_TIMEOUT = 5
 
 
-def get_es(hosts=None, timeout=None, **settings):
+def get_es(hosts=None, timeout=None):
     """Create an ES object and return it."""
     # Cheap way of de-None-ifying things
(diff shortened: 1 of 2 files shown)


src/olympia/amo/search.py
```diff
@@ -14,7 +14,7 @@
 DEFAULT_TIMEOUT = 5
 
 
-def get_es(hosts=None, timeout=None, **settings):
```
diox, line 17: `**settings` was never passed anywhere that I could see

eviljeff: we don't ever appear to pass `hosts` either, fwiw - and the one place we pass `timeout` as an arg it's `settings.ES_TIMEOUT` which is the default anyway

eviljeff: And now I'm looking, the `getattr(...` fallbacks with the constants in this file are also redundant - we define `ES_HOSTS` and `ES_TIMEOUT` in settings_base so those settings are also present.

eviljeff (approved):
r+wc nits about redundant kwargs and fallbacks

PR #18961, opened 2022-03-10
Remove legacy SearchMixin, Addon.search() and ES class

### LINKED ISSUE #18962: Remove legacy SearchMixin and ES classes and related code
`SearchMixin`, `Addon.search()`, `Addon.index()`, the whole `ES` class... all of this is legacy code that calls Elasticsearch directly but shouldn't be used anymore: we should be using `elasticsearch_dsl` everywhere instead. Remove it!

diox:
We shouldn't have code using this anymore, it should all be using the API and `elasticsearch_dsl`.

Fixes #18962


src/olympia/addons/migrations/0026_addonregionalrestrictions.py
```diff
@@ -26,6 +26,6 @@ class Migration(migrations.Migration):
                 'abstract': False,
                 'base_manager_name': 'objects',
             },
-            bases=(olympia.amo.models.SearchMixin, olympia.amo.models.SaveUpdateMixin, models.Model),
+            bases=(olympia.amo.models.SaveUpdateMixin, models.Model),
```
eviljeff, line 29: I wonder if it's a problem changing all these old migrations?  maybe we should just squash them?

eviljeff (approved):

src/olympia/addons/tasks.py
```diff
@@ -69,21 +71,29 @@ def index_addons(ids, **kw):
...
+        log.info('Removing addon [%s] from search index.' % addon_id)
+        try:
+            es.delete(
+                AddonIndexer.get_index_alias(),
+                AddonIndexer.get_doctype_name(),
+                addon_id,
+            )
+        except TransportError:
+            # We ignore already deleted add-ons.
+            pass
```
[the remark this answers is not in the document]
diox: I agree, though I would rather move all functionality to the indexer class tbh, but I wasn't willing to go that far yet...

src/olympia/addons/migrations/0026_addonregionalrestrictions.py
diox: We've done this before, it should be fine (it's just to allow the schema editor to create realistic fake versions of those classes anyway)

diox: (I'd be up for a squash in the future though)

src/olympia/addons/tasks.py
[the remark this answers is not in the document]
diox: I think it deserves to be part of a larger effort so I filed https://github.com/mozilla/addons-server/issues/18966

PR #18969, opened 2022-03-11
Remove obsolete pfs and theme_update wsgi endpoints

### LINKED ISSUE #18968: Remove obsolete pfs and theme_update wsgi endpoints
Both of these endpoints are obsolete, they no longer work and are not referenced anywhere:
- https://github.com/mozilla/addons-server/blob/master/services/wsgi/theme_update.py
- https://github.com/mozilla/addons-server/blob/master/services/wsgi/pfs.py

We should remove them.

diox:
Fixes #18968


eviljeff (approved):
I can't even remember how long ago we deleted the `theme_update.py` and `pfs.py` files that these scripts would try to load `application` from!

PR #18977, opened 2022-03-14
Introduce a truly supported release notes API endpoint for Firefox

### LINKED ISSUE #18973: Point Firefox to new update info endpoint and add caching headers
Few things I realized looking at the update info (release notes) endpoint from the Firefox side while looking into https://github.com/mozilla/addons-server/issues/18971:
- Firefox gets that URL from the update service, which currently still returns the legacy URL (formatting the string manually with no `reverse()` one of the parameters being dynamic and dealt with by Firefox in the returned URL). This can be seen by calling the update service directly, as demonstrated by example [for ublock origin upgrading from 1.38.0 to 1.48.1](https://versioncheck.addons.mozilla.org/update/versionCheck.php?reqVersion=1&id=uBlock0@raymondhill.net&version=1.38.0&maxAppVersion=98.0&appID={ec8030f7-c20a-464f-9b0e-13a3a9e97384}&appVersion=96.0&appOS=WINNT&appABI=x86-msvc) - notice the `update_info_url` URL, which points to `https://addons.mozilla.org/versions/updateInfo/...`
- The URL returned is on the main domain instead of `services.addons.mozilla.org` as it should be
- There are 4 requests - 3 redirects before landing on the actual URL. They can be observed by opening `about:addons` for an add-on with release notes set, opening devtools to monitor network activity, and clicking at the `Release Notes` tab. For instance, for ublock-origin again:
  ```
  GET https://addons.mozilla.org/versions/updateInfo/5368975/en-US/
  GET https://addons.mozilla.org/en-US/firefox/versions/updateInfo/5368975/en-US/
  GET https://addons.mozilla.org/en-US/firefox/addon/607454/versions/1.41.8/updateinfo/
  GET https://addons.mozilla.org/en-US/firefox/addon/ublock-origin/versions/1.41.8/updateinfo/
  ```
  The first redirect is caused by calling the legacy URL with the locale being a suffix instead of a prefix, the second by the add-on id being used instead of the slug.

We should fix the URL directly [at the source](https://github.com/mozilla/addons-server/blob/d1e29458111842d98e48d8e5202a484dece78214/services/update.py#L205-L208), removing the need for those redirects and setting the right domain as well.

In addition, we serve those URLs without caching headers, we should add some, caching for 1 hour seems reasonable (matches the duration the XPIs and the updates are cached for).

diox:
Document it, return it directly in the update service (redirecting previous undocumented endpoints), add a bit of caching to it.

Fixes #18973

### DIFF
--- a/docs/topics/api/addons.rst
+++ b/docs/topics/api/addons.rst
@@ -457,6 +457,19 @@ This endpoint allows you to fetch a single version belonging to a specific add-o
     :>json string|null source: The (absolute) URL to download the submitted source for this version. This field is only present for authenticated users, for their own add-ons.
     :>json string version: The version number string for the version.
 
+---------------------
+Version Release Notes
+---------------------
+
+.. _version-release_notes:
+
+This endpoint allows you to fetch release notes for a version belonging to a specific add-on. The response will be an XHTML document. It is not meant for third-party consumption - use the version detail endpoint instead.
+
+.. http:get:: /api/v5/addons/addon/(int:addon_id|string:addon_slug|string:addon_guid)/versions/(int:id)/release_notes/
+
+    .. _version-detail-release_notes:
+
+    :query string lang: Activate translations in the specific language for that query. (See :ref:`translated fields <api-overview-translations>`)
 
 --------------
 Version Create
--- a/docs/topics/api/v4_frozen/addons.rst
+++ b/docs/topics/api/v4_frozen/addons.rst
@@ -394,6 +394,20 @@ This endpoint allows you to fetch a single version belonging to a specific add-o
     :>json boolean is_strict_compatibility_enabled: Whether or not this version has `strictCompatibility <https://developer.mozilla.org/en-US/Add-ons/Install_Manifests#strictCompatibility>`_. set.
     :>json string version: The version number string for the version.
 
+---------------------
+Version Release Notes
+---------------------
+
+.. _v4-version-release_notes:
+
+This endpoint allows you to fetch release notes for a version belonging to a specific add-on. The response will be an XHTML document. It is not meant for third-party consumption - use the version detail endpoint instead.
+
+.. http:get:: /api/v5/addons/addon/(int:addon_id|string:addon_slug|string:addon_guid)/versions/(int:id)/release_notes/
+
+    .. _v4-version-detail-release_notes:
+
+    :query string lang: Activate translations in the specific language for that query. (See :ref:`translated fields <api-overview-translations>`)
+
 
 ------------------------------
 Add-on EULA and Privacy Policy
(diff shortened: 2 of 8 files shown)


eviljeff (commented):
Code itself is r+ but I'm really struggling with the idea of serving a html document (albeit a small one) via our JSON based api.  Especially when it's integrated into an existing viewset that's currently entirely JSON too.

If it really needs to be under `/api/` perhaps we can have a specific path & viewset for it?

eviljeff:
Tbh, I'd prefer Firefox change to be able to consume the version detail JSON output directly so it can just use release notes from that without a need for a speical endpoint.  I don't know how pressing the issue is though - that's a larger (and slower) change that making a change on the AMO side.  (With the self-hosting question too)

diox:
I'm fine with having it elsewhere. The reason behind having this under `/api/` was to avoid all the madness about locale & app prefixing and ensure it would be properly documented. When I started looking into this endpoint in my previous PR I was confused as to what was the actual URL used by Firefox, until I realized this was exposed through the update service. That motivated me to introduce a documented endpoint. But maybe we can achieve that differently - I just thought `/api/` kinda made sense, but I don't feel strongly about it.

> Tbh, I'd prefer Firefox change to be able to consume the version detail JSON output directly so it can just use release notes from that without a need for a speical endpoint. I don't know how pressing the issue is though - that's a larger (and slower) change that making a change on the AMO side. (With the self-hosting question too)

Yeah, different, larger and more complex conversation. I thought about that for sure but that'd be a longer term thing I'm not sure I want to deal with right now.

diox:
I remembered an extra twist: https://github.com/mozilla/addons-server/issues/18973#issuecomment-1066614665
> Slight complication here: the url is not under `/api/` to the services.a.m.o. configuration will reject it as a 403 at the moment. But it could be under `/api/` ... we would need to pick an arbitrary API version though.

diox:
Because of the services.a.m.o. stuff I think we should really use `/api/` to keep the config sane.

> If it really needs to be under `/api/` perhaps we can have a specific path & viewset for it?

If you feel strongly about this I can do that. Essentially exposing the old existing endpoint under `/api/` instead of making a new `@action` inside the `AddonVersionViewSet`.

eviljeff:
> Because of the services.a.m.o. stuff I think we should really use `/api/` to keep the config sane.
> 
> > If it really needs to be under `/api/` perhaps we can have a specific path & viewset for it?
> 
> If you feel strongly about this I can do that. Essentially exposing the old existing endpoint under `/api/` instead of making a new `@action` inside the `AddonVersionViewSet`.

yeah I don't have an ideal place for the url to sit.  It's just that it's not an API endpoint at all, so exposing it as one seemed wrong.

diox:
IMHO it is an API - just one that returns (X)HTML. The document returned isn't usable on its own, it's not something you would browse to, you need to parse it or iframe it in some way.

PR #18978, opened 2022-03-14
Serve xpis through addons server

### LINKED ISSUE #18844: Serve xpis from addons-server through `X-Accel-Redirect`
Currently, and since pretty much forever, `.xpi` downloads go through the following process:
1) `GET https://addons.mozilla.org/firefox/downloads/file/{file_id}/{filename}`
  That request is handled by addons-server. It considers authentication & permissions, and is not cached. It redirects to...
2) `GET https://addons.cdn.mozilla.net/user-media/addons/{addon_id}/{filename}?filehash={hash}`
  That request is handled by nginx from EFS directly, bypassing addons-server. It does not consider authentication & permissions, and is cached for 24 hours.

That second URL depends on nginx having some configuration to directly serve the file from where it's stored on EFS. The idea behind it is to avoid hitting addons-server, for perf/scalability concerns.

Because of this process, the whole `guarded-addons` concept was invented back in the day, where deleted/disabled files are moved away from the normal place where files are stored, to prevent nginx from being able to serve them. This forces us to maintain 2 directory structures, and move files around when deleting/disabling, which in turn is responsible for a number of quirks/bugs where files are not always at the right place at the right time, and awful cronjobs regularly go through our files to resume any move that failed.

Ahead of the migration to GCP, we want to move away from this architecture to make file storage simpler to change. The idea would be to:
- No longer move files to `guarded-addons/`
- Replace the step 2) in the process above: Instead of serving the file by nginx from EFS, we would go through addons-server, do a database request, determine whether the file can be served (is not disabled/deleted/etc), and let nginx serve the `.xpi` with `X-Accel-Redirect`.

We might completely remove the step 2) and just do everything in the first response (avoiding the redirect) later, but I am not sure we want to do this now yet. Note that the update endpoint that serves add-on updates to Firefox clients directly links to the URL from step 2).

Hidden benefit: currently the URL directly maps to the path on the filesystem. We'll want to change the directory structure in https://github.com/mozilla/addons-server/issues/11510 so this would help.

diox:
Serve xpis through X-Accel-Redirect instead of redirecting to dedicated domain

The so called "CDN URLs" for files are less useful now that even AMO main domain is behind a CDN, and we want to move away from it because internally it exposes paths on EFS that nginx is serving directly, making directory structure changes difficult.

This commit focuses on the download_file/download_latest endpoints addons-server exposes, and leaves the old dedicated CDN domain intact for the moment - it is still used by the update service but will go away in a future commit.

Fixes #18844

### DIFF
--- a/src/olympia/versions/urls.py
+++ b/src/olympia/versions/urls.py
@@ -19,7 +19,7 @@ download_patterns = [
     # .* at the end to match filenames.
     # /file/:id/type:attachment
     re_path(
-        r'^file/(?P<file_id>\d+)(?:/type:(?P<type>\w+))?(?:/.*)?',
+        r'^file/(?P<file_id>\d+)(?:/type:(?P<type_>\w+))?(?:/.*)?',
         views.download_file,
         name='downloads.file',
     ),
@@ -28,7 +28,7 @@ download_patterns = [
     ),
     # /latest/1865/type:xpi/platform:5
     re_path(
-        r'^latest/%s/(?:type:(?P<type>\w+)/)?'
+        r'^latest/%s/(?:type:(?P<type_>\w+)/)?'
         r'(?:platform:(?P<platform>\d+)/)?.*' % ADDON_ID,
         views.download_latest,
         name='downloads.latest',
--- a/src/olympia/versions/views.py
+++ b/src/olympia/versions/views.py
@@ -1,11 +1,9 @@
-import os
-
 from django import http
 from django.db.transaction import non_atomic_requests
 from django.shortcuts import get_object_or_404, redirect
 from django.template.response import TemplateResponse
 from django.urls import reverse
-from django.utils.cache import patch_vary_headers
+from django.utils.cache import patch_cache_control, patch_vary_headers
 
 import olympia.core.logger
 
@@ -13,7 +11,7 @@ from olympia import amo
 from olympia.access import acl
 from olympia.addons.decorators import addon_view_factory
 from olympia.addons.models import Addon, AddonRegionalRestrictions
-from olympia.amo.utils import HttpResponseXSendFile, urlparams
+from olympia.amo.utils import HttpResponseXSendFile
 from olympia.files.models import File
 from olympia.versions.models import Version
 
@@ -59,11 +57,9 @@ def update_info_redirect(request, version_id):
 
 # Should accept junk at the end for filename goodness.
 @non_atomic_requests
-def download_file(request, file_id, type=None, file_=None, addon=None):
+def download_file(request, file_id, type_=None):
     """
-    Download given file.
-
-    `addon` and `file_` parameters can be passed to avoid the database query.
+    Download the file identified by `file_id` parameter.
 
     If the file is disabled or belongs to an unlisted version, requires an
     add-on developer or appropriate reviewer for the channel. If the file is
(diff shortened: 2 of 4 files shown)


src/olympia/versions/tests/test_views.py
```diff
@@ -566,43 +585,134 @@ def setUp(self):
...
+                kwargs={'addon_id': self.addon.slug},
+            )
+            == '/firefox/downloads/latest/better-gcal-5299/'
+        )
+        assert (
+            reverse(
+                'downloads.latest',
+                kwargs={'addon_id': self.addon.slug, 'type_': 'attachment'},
+            )
+            == '/firefox/downloads/latest/better-gcal-5299/type:attachment/'
```
eviljeff, line 604: 😨 it actually puts `type:attachment/` - that's pretty strange!

src/olympia/versions/views.py
```diff
@@ -157,27 +144,30 @@ def is_appropriate_reviewer(addon, channel):
...
+    approved listed version.
     """
-    try:
-        file_ = addon.current_version.file
-    except IndexError:
-        raise http.Http404()
-    return download_file(request, file_.id, type=type, file_=file_, addon=addon)
+    file_ = addon.current_version.file
+    response = http.HttpResponseRedirect(
+        file_.get_absolute_url(attachment=type_ == 'attachment')
```
eviljeff, line 166: this reads weirdly - can we have `(...)` - around the boolean check at least?

src/olympia/versions/views.py
```diff
@@ -57,13 +57,10 @@ def update_info_redirect(request, version_id):
     )
 
 
-# Should accept junk at the end for filename goodness.
 @non_atomic_requests
-def download_file(request, file_id, type=None, file_=None, addon=None):
+def download_file(request, file_id, type_=None, filename=None):
```
eviljeff, line 61: I had to re-read over the patch a few times to confirm to myself that we don't actually do anything with the `filename` parameter even though we pass it down everywhere.  Do we need to specify the kwarg? (could just be `**kwargs` if we are forced to capture it in the regex to name it).

eviljeff (approved):

src/olympia/versions/views.py
diox: We don't need it, we are forced to capture it to name it and pass it to `reverse()` without doing the weird string concatenation we were doing before, but you're right I could just accept `kwargs` instead here.

My initial idea was to use it to ensure the filename received was not garbage and matched the database, but that is a somewhat backwards-incompatible change as https://addons-server.readthedocs.io/en/latest/topics/api/signing.html#downloading-signed-files mentions filename `(...) is just a convenience for clients so that they write meaningful file names to disk`.

src/olympia/versions/tests/test_views.py
diox: The whole thing with optional captured arguments inside an optional non-capturing group drove me crazy. At some point I had the opposite behavior, where we'd get `/firefox/downloads/latest/better-gcal-5299/attachment`. That's why I added this test (there are more in `files/tests/test_models.py` - I added this one here for convenience when debugging views.

src/olympia/versions/urls.py
```diff
@@ -16,20 +16,26 @@
 ]
 
 download_patterns = [
-    # .* at the end to match filenames.
-    # /file/:id/type:attachment
+    # /file/<id>/type:attachment/lol.xpi - everything after the file id is
+    # ignored though.
     re_path(
-        r'^file/(?P<file_id>\d+)(?:/type:(?P<type>\w+))?(?:/.*)?',
+        r'^file/(?P<file_id>\d+)/(?:type:(?P<type_>\w+)/)?(?:(?P<filename>[\w+.-]*))?$',
```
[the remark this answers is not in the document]
diox: I went with `download_type` as it's not really a file type

PR #18982, opened 2022-03-16
Handle malformed multipart/form-data errors without raising a 500

### LINKED ISSUE #18981: Exception about request.session not being set raised for invalid multipart requests
We've seen this weird exception in Sentry sometimes:

```
CSRF_USE_SESSIONS is enabled, but request.session is not set. SessionMiddleware must appear before CsrfViewMiddleware in MIDDLEWARE.
```

But in our settings, our custom child of `SessionMiddleware`, `NoVarySessionMiddleware`, does appear
before `CsrfViewMiddleware`... 

The key is that it happens when the request itself is malformed. Specifically, I've been able to reproduce it with bad multipart/form-data requests like these:

```sh
% curl https://addons-dev.allizom.org/api/v5/addons/upload/ -H 'Content-Type: multipart/form-data' -d '' -H Authorization: Session xxxxx
```

We expect the following 400 error response:
```json
["Missing \"upload\" key in multipart file data."]
```

We get a 500 instead with the aforementioned exception logged in Sentry. The reason why is that for this request, we go through our `LocaleAndAppURLMiddleware`, and it does this:
https://github.com/mozilla/addons-server/blob/129f8c1291f4c78ead877b5d97117c64b706e905/src/olympia/amo/middleware.py#L120

Which itself does this:
https://github.com/mozilla/addons-server/blob/129f8c1291f4c78ead877b5d97117c64b706e905/src/olympia/amo/urlresolvers.py#L84

And that triggers Django post data and files loading from the request, which it cannot do because the request is invalid. It wants to trigger an exception (`MultiPartParserError`) for this, but this is supposed to happen from a view (where all middlewares would be loaded already), not from inside a middleware, especially not as early in the chain as we are.

Django error handling could be smarter about this and fix this, but we could also work around this on our own: maybe we can avoid loading `request.POST` like that, at least for all requests, or maybe we can catch the problem in `Prefixer.get_language()`, or maybe rework our middlewares order to ensure `request.session` is loaded early enough to avoid this entirely.

diox:
- ~~Since we try to load `request.POST` in a middleware, catch any multipart parsing errors there and don't let them raise~~ Stop loading `lang` from `request.POST` in middleware
- Remove custom obsolete DRF `MultiPartParser` that doesn't expose the underlying error when it happens

Fixes #18981


src/olympia/amo/urlresolvers.py
```diff
@@ -77,11 +78,23 @@ def get_app(self):
...
-        Return a locale code that we support on the site using the
-        user's Accept Language header to determine which is best.  This
-        mostly follows the RFCs but read bug 439568 for details.
+        Return a locale code that we support on the site using `lang` from GET
+        or POST, falling back to the user's Accept Language header if
+        necessary (mostly following the RFCs but read bug 439568 for details).
         """
-        data = self.request.GET or self.request.POST
+        try:
+            data = self.request.GET or self.request.POST
```
eviljeff, line 86: Not strictly related to this issue/patch but I didn't think we supported `lang` setting the language as POST data- only as a url parameter?  If it's the `request.POST` that raises then only checking `request.GET` would mean we wouldn't need this handling.

eviljeff (approved):
r+wc but we might not need some of this?

src/olympia/amo/urlresolvers.py
diox: Actually even outside of storing translations, having a way for a POST API request to specify the language outside of just `Accept-Language` header might be useful... it would be used to translate error messages for instance

diox: ... but we do that through the query string, i.e. for ratings we `POST` to `https://addons.mozilla.org/api/v5/ratings/rating/?lang=<lang>`, and the language is not part of the post data. So we could probably remove this...

eviljeff: There are writeable apis available under v3/v4 that have translations but they were all internal auth (collections, accounts) - it's unclear if anyone used them with the session cookie and flat strings. 

We don't appear to document how you're supposed to set the "current" language in v3/v4 where we mention you can specify a single flat string for localized values so we don't know if people would have thought to specify it in the POST or accept language header (or as a URL param)

diox: There are 2 kind of API error messages: those tied to the client misusing parameters in some way (invalid or missing parameters) and those tied to user-supplied data. The former we tend not to translate, but I would expect us to always translate the latter.

That being said, we support and use `lang` from the query string even in POSTS as mentioned above, we support `Accept-Language`, and we haven't really documented (or used ourselves) that behavior where we look for `lang` in `POST` data, so I think it's fine to remove. I'll do that.

PR #18991, opened 2022-03-17
Write FileUploads to a directory under MEDIA_ROOT (/uploads/)

### LINKED ISSUE #18871: Move FileUpload storage to `shared_storage/uploads/fileuploads`
`FileUpload` files are stored in a `temp` directory under `files/` (where permanent add-ons are stored):

https://github.com/mozilla/addons-server/blob/1ac05a5454f04b0985d1b9b540ca3af601882276/src/olympia/files/models.py#L514-L516

This is a bit inconsistent with everything else: `files/` should only really contain permanent xpis (i.e. `File`). We already have a `/shared_storage/tmp/` path for really temporary stuff, and a bunch of other uploads are stored under `shared_storage/uploads/<directory>/`.

Since `FileUpload`s are short-lived but not quite as short-lived as other temporary files that are discarded right after uploads, I think we should use `shared_storage/uploads/fileuploads` instead.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-80)

diox:
Fixes #18871

### DIFF
--- a/src/olympia/files/models.py
+++ b/src/olympia/files/models.py
@@ -508,7 +508,7 @@ class FileUpload(ModelBase):
         if ext in amo.VALID_ADDON_FILE_EXTENSIONS:
             ext = '.xpi'
         self.path = os.path.join(
-            user_media_path('addons'), 'temp', f'{uuid.uuid4().hex}{ext}'
+            settings.MEDIA_ROOT, 'fileuploads', f'{uuid.uuid4().hex}{ext}'
         )
 
         hash_obj = None
--- a/src/olympia/files/tests/test_models.py
+++ b/src/olympia/files/tests/test_models.py
@@ -738,7 +738,9 @@ class TestFileUpload(UploadMixin, TestCase):
         assert upload.name == f'{force_str(upload.uuid.hex)}_filenamé.xpi'
         # Actual path on filesystem is different, fully random
         assert upload.name not in upload.path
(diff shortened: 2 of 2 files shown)


src/olympia/files/models.py
```diff
@@ -508,7 +508,7 @@ def add_file(self, chunks, filename, size):
         if ext in amo.VALID_ADDON_FILE_EXTENSIONS:
             ext = '.xpi'
         self.path = os.path.join(
-            user_media_path('addons'), 'temp', f'{uuid.uuid4().hex}{ext}'
+            settings.MEDIA_ROOT, 'fileuploads', f'{uuid.uuid4().hex}{ext}'
```
eviljeff, line 511:
```suggestion
            user_media_path('fileuploads'), f'{uuid.uuid4().hex}{ext}'
```

eviljeff (approved):

src/olympia/files/models.py
diox: mmm yes, but... I was kinda thinking about removing that utility function entirely. I find it more confusing that helpful...

eviljeff: it _is_ a bit "magic"

diox: https://github.com/mozilla/addons-server/issues/18999

PR #19001, opened 2022-03-21
Update redis docker image

willdurand:
This is needed to run addons-server with Apple silicons. I don't know we
want to do that upfront given I don't know which version we're using on
production but this PR can serve as "documentation" in the meantime.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -78,7 +78,7 @@ services:
     mem_limit: 2g
 
   redis:
-    image: redis:2.8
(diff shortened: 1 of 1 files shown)


diox (approved):
Tested locally and it seems to be working, which is not super surprising as it's just the celery result backend - we don't use it for anything else anymore.

We can merge it now, there is a Jira ticket filed for ops but given that it's a fairly transparent change we don't need to wait.

PR #19004, opened 2022-03-21
support setting/updating addon icon via the addons api

### LINKED ISSUE #18232: Allow icons to be updated via the API
split from #18036

Some options here:
 - Allow the addon.icons property to be writeable and for an upload blob to be specified (`null` for delete, I guess) <- going with this option
 - ~Have a specified sub-endpoint under addons to upload an icon and/or delete it~
 - ~Support updating the addon icon by uploading a new version of the xpi - i.e. extract the icon from the xpi for each version.  (if we're going to do this automatically, we probably don't want to allow other ways for the icon to be updated)~

eviljeff:
fixes #18232


src/olympia/addons/serializers.py
```diff
@@ -898,6 +904,30 @@ def validate(self, data):
...
+        if uploaded_icon:
+            original = f'{destination}-original.png'
+
+            storage = SafeStorage(user_media='addon_icons')
+            with storage.open(original, 'wb') as original_file:
+                for chunk in uploaded_icon.chunks():
+                    original_file.write(chunk)
+
+            self.instance.update(icon_type=uploaded_icon.content_type)
+            # note: we're calling the task without .delay() to resize immediately
```
diox, line 920: Why ?

eviljeff: we return the links to the resized image files in the serialized response so I thought the files should actually exist at that point or they'd 404 if they were loaded immediately (or there was celery task queue traffic).  It should be quick enough to resize a single file to 32, 64, and 128.

diox: It might be quick enough in most cases, but it also calls `pngcrush_image`, which in turn calls a (potentially slow) binary, something I think we should avoid doing in a web process.

Current devhub works around the same problem in 2 ways:
- Before you click save, the preview is generated locally
- After you click save, it calls the icon with a `modified=` parameter in a loop until they stop 404ing.

This might not be super elegant but I think it's still better than executing the resize+crush code in a web process.

eviljeff: I know, but how's that going to work with the api though?  Or are you suggesting we just return the urls as potentially 404s and expect the client to know what to do (and/or document how devhub does it in the api docs as a suggestion)?

diox: Yes, I think we can return the URLs as potentially 404s. They should be served without caching (and if we decide to add caching to errors, we could do a very short expiration) so they shouldn't even need devhub's trick and can use the URL as-is.

diox (approved):

PR #19006, opened 2022-03-22
limit contributions_url to strict domains with https.

### LINKED ISSUE #19000: The /addons/create/ API returns a 500 error if the `contributions_url` field contains some invalid values
Follow up for #18267

1. Create a new add-on using the `/addons/create/ ` API
2. Add a `contributions_url` in the JSON request body with either a random string or number value, i.e. `"foobar"` or `12345`
3. Send the request

### What happened?
A 500 server error is returned

A 400 error asking for a valid URL

Empty strings , `bool` and `null` values produce correct validation messages with a 400 response

### LINKED ISSUE #19005: drop (now unneeded) support for subdomains in contributions url
At the time the[ contributions url was added](https://github.com/mozilla/addons-server/issues/6070) there was a specific need for subdomains, but that's no longer the case.

eviljeff:
fixes #19005 and seems to also fix #19000 along the way.
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1760296


src/olympia/addons/migrations/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    'paypal.com': 'www.paypal.com',
+    'paypal.me': 'www.paypal.com',
+}
+
+BROKEN_URLS = (
+    'https://www.fdfgd@paypal.me',
+    'https://gist.github.com/miry/2a7635beaa7078e0d7af',
+    'https://thezenithpoint.co.uk@donate.mozilla.org',
+    'https://bunny.github.com',
+)
```
eviljeff, line 25: these are the known broken urls that I could see a way to sensibly fix

src/olympia/addons/migrations/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def fix_contribitions_url(url):
```
eviljeff, line 28: btw, this code is in here rather than the migration file itself because I wanted it to have test coverage but importing from a filename that starts with numbers is a giant pain in python.

src/olympia/constants/base.py
```diff
@@ -212,16 +212,16 @@
...
     'ko-fi.com',
     'liberapay.com',
-    'micropayment.de',
+    'www.micropayment.de',
     'opencollective.com',
-    'patreon.com',
-    'paypal.com',
+    'www.patreon.com',
+    'www.paypal.com',
     'paypal.me',
```
eviljeff, line 225: to work out which was the canonical domain I just loaded up each site without the `www.` and saw if it redirected.

src/olympia/addons/migrations/__init__.py
diox: typo in the name of the function `contribitions` vs `contributions` (which you seem to have copy/pasted so at least it doesn't break...)

src/olympia/addons/fields.py
```diff
@@ -94,19 +94,23 @@ def to_representation(self, value):
...
         elif parsed_url.hostname == 'github.com' and not parsed_url.path.startswith(
             '/sponsors/'
         ):
             # Issue 15497, validate path for GitHub Sponsors
-            raise exceptions.ValidationError(
-                'URL path for GitHub Sponsors must contain /sponsors/.'
-            )
+            errors.append('URL path for GitHub Sponsors must contain /sponsors/.')
+        if parsed_url.scheme != 'https':
+            errors.append('URLs must start https://.')
```
diox, line 110: Missing `with` (also below and in tests)

eviljeff: Read okay to me as is, but can do.

src/olympia/addons/tests/test_views.py
```diff
@@ -709,8 +709,56 @@ def test_with_grouped_ratings(self):
...
+            'contributions_url': [
+                f'URL domain must be one of [{domains}].',
+                'URLs must start with https://.',
+            ]
+        }
+
+        valid_url = 'https://flattr.com/xxx'
+        response = self.request(contributions_url=valid_url)
+        assert response.status_code == self.SUCCESS_STATUS_CODE, response.content
+        assert response.data['contributions_url']['url'].startswith(valid_url)
```
bobsilverberg, line 738: I think we can now use `==` instead of `startswith`.

src/olympia/addons/tests/test_views.py
```diff
@@ -709,8 +709,56 @@ def test_with_grouped_ratings(self):
...
+        assert response.data == {
+            'contributions_url': [
+                'URL path for GitHub Sponsors must contain /sponsors/.',
+            ]
+        }
+
+        valid_url = 'https://github.com/sponsors/xxx'
+        response = self.request(contributions_url=valid_url)
+        assert response.status_code == self.SUCCESS_STATUS_CODE, response.content
+        assert response.data['contributions_url']['url'].startswith(valid_url)
```
bobsilverberg, line 754: I think we can now use `==` instead of `startswith`.

src/olympia/constants/base.py
bobsilverberg, line 225: In the migration, you're converting `paypal.me` to `www.paypal.com`, so should we be keeping `paypal.me` here? And if so, why are we converting it in the migration?

bobsilverberg (approved):
Looks good. Just a few comments.

r+wc

src/olympia/constants/base.py
diox: We shouldn't convert it: paypal started redirecting `https://paypal.me/foo` to `https://www.paypal.com/paypalme/foo` at some point, but I don't think we should be trying to do that conversion ourselves. And just replacing `https://paypal.me/foo` with `https://paypal.com/foo` is wrong.

eviljeff: whoops (it's a mistake in the dict in the migration)

src/olympia/addons/tests/test_views.py
eviljeff: we do `startswith` because we append utm params to the end of the url

src/olympia/addons/tests/test_views.py
eviljeff: same - it's because there are utm parameters on the end

src/olympia/constants/base.py
eviljeff: https://github.com/mozilla/addons-server/pull/19006/commits/082e8b7ee99b05f7e54b8c18c338bd6a6a470f83

PR #19011, opened 2022-03-23
Remove code dedicated domain CDN - download xpis through the main domain

### LINKED ISSUE #18998: Remove usage of addons.cdn.mozilla.net
Remove any code dealing with addons.cdn.mozilla.net in addons-server. Links to download `.xpi` files should use the endpoint on the main domain from https://github.com/mozilla/addons-server/issues/18844 instead (including in the update service).

When this is deployed, we'll need an ops ticket to check usage of `addons.cdn.mozilla.net` and eventually decommission it, along with removing the ability to serve `.xpi` files directly from nginx (those downloads should go through addons-server and be served through `X-Accel-Redirect` and `internal` nginx directive instead). Then, we'll be able to do https://github.com/mozilla/addons-server/issues/18869

diox:
Fixes #18998


diox:
This needs to be reworked to avoid passing `filehash` at all times, given currently Fenix would not support it - see https://github.com/mozilla/addons-server/issues/19013

eviljeff (approved):

PR #19017, opened 2022-03-24
Fix addons/migration/0037 to not set "" to "https://"

eviljeff:
fixes #19012

### DIFF
--- a/src/olympia/addons/migrations/0037_auto_20220321_1545.py
+++ b/src/olympia/addons/migrations/0037_auto_20220321_1545.py
@@ -12,7 +12,7 @@ def contributions_url_to_from_www(apps, schema_editor):
     Addon = apps.get_model('addons', 'Addon')
     to_index = []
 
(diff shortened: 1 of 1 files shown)


src/olympia/addons/migrations/0037_auto_20220321_1545.py
```diff
@@ -32,3 +32,4 @@ class Migration(migrations.Migration):
     ]
 
     operations = [migrations.RunPython(contributions_url_to_from_www)]
+19012-fix-migration-addons-0037
```
wagnerand, line 35: accidental add?

eviljeff: stray drag'n'drag of the branch name

bobsilverberg (approved):
I r+'d this last time because just looking at it I assumed it was fine, but evidently it was not. I'm not intimate with the syntax of `Q`, but I assume this definitely fixes it.

eviljeff:
> but I assume this definitely fixes it.

4th time lucky for me definitely writing correct code 🤞  😄

Issue #11510, opened 2019-05-24
New directory structure for xpis storage


bqbn:
### Describe the problem and steps to reproduce it:
Per the comment in https://github.com/mozilla/addons-server/issues/11464#issuecomment-495544507, we'd like to request change of directory structure under `guarded-addons/` and `files/` directory.

It is suggested that instead of a flat directory structure, we should have a 3-level deep structure under `guarded-addons`/`files/` that is based on the add-on id, for example, something like `/guarded-addons/9/89/789/123456789/`.

 
### What happened?
There are too many sub-directories under `guarded-addons/` and `files/`, currently more than 327K of them and keeps growing.

### What did you expect to happen?
A new directory structure for guarded-addons directory.

### Anything else we should know?
There is also https://github.com/mozilla/addons-server/issues/3598. If it can be done first, then we can probably close this one by that time.

**Edited** by @diox to reflect latest directory structure plans

bqbn:
Assuming the work for this hasn't started, I just updated the request, changing it from a 2-level deep to 3-level deep. The reason is that given 1 million addon ids, 2-level directory structure means there could be 10,000 sub-directories at the third level, which is still too many for `ls` to handle. Besides, our site keeps growing, I suspect it'll grow into multi-million addon ids in a few years.

bqbn:
Any chance we can make this ticket into a project and try to tackle it in year 2022?

The original request was for the `guarded-addons` directory only, however, I think the `files` and `shared_storage` directories should also be considered for using the hierarchical directory layout because they suffer the same problem as the `guarded-addons` directory and it is largely due to the flat directory layout.

diox:
Started working on a [document](https://docs.google.com/document/d/1fn-C0nl5wFAur4yEsPiEoXAKznAxl4u1TGcvaqLD_N4/edit#) detailing the issue and potential solutions.

diox:
Nice side-benefit of doing https://github.com/mozilla/addons-server/issues/18844 first: once files are served through addons-server instead of nginx, their path can change independently of their URL. Currently the 2 are inextricably linked, the URL is directly built from the path on EFS - With urls like `https://addons.cdn.mozilla.net/user-media/addons/<addonid>/<filename.xpi>?filehash=<hash>` pointing to a file under `files/addons/<addonid>/<filename.xpi>`.

diox:
Additional thing to do while we're renaming files:
[...]

PR #19019, opened 2022-03-24
Change file extension to .xpi for signed files, .zip for unsigned

### LINKED ISSUE #19013: Cachebust download URLs after signing
We recently changed the way we do downloads, [removing a redirect and serving them directly](https://github.com/mozilla/addons-server/issues/18844). One thing that is gone as a result though, is the `filehash` parameter ([bug 1155813](https://bugzilla.mozilla.org/show_bug.cgi?id=1155813), [bug 1158738](https://bugzilla.mozilla.org/show_bug.cgi?id=1158738)). That parameter was applied to the CDN URLs to ensure that we cached pre and post signing separately, in case the URL was somehow accessed before signing.

Now that we only have a single download URL and no redirect, we should add the parameter to that download URL to ensure the following STRs work:
- Submit an xpi (new or existing add-on)
- Grab the download URL for that xpi **before it gets signed**.
- Access it a few times without authentication (in a private window for instance). You should get an error, and eventually, the responses should be cached by the CDN (`X-Cache-Status: Hit`, `Cache-Control: max-age=<24 hours>`)
- Get it approved and signed
- Visit the frontend page without authentication again, install the add-on from there
- Your add-on should be downloaded and installed correctly instead of errorring out.

I intended to do this in #18998, but learned that [Fenix would break if we starting adding the parameter to our download URLs](https://github.com/mozilla-mobile/fenix/issues/24407), so this is a separate issue. We'll need to wait until Fenix fixes the problem and it rides the train to fully do this. Maybe in the meantime we should not do it for android ?

diox:
Fixes #19013

### DIFF
--- a/src/olympia/files/models.py
+++ b/src/olympia/files/models.py
@@ -12,11 +12,11 @@ from django.conf import settings
 from django.core.files.storage import default_storage as storage
 from django.db import models
 from django.dispatch import receiver
-from django.template.defaultfilters import slugify
 from django.urls import reverse
 from django.utils.crypto import get_random_string
 from django.utils.encoding import force_bytes, force_str
 from django.utils.functional import cached_property
+from django.utils.text import slugify
 
 from django_statsd.clients import statsd
 
@@ -122,8 +122,6 @@ class File(OnChangeMixin, ModelBase):
 
         file_ = cls(version=version)
         upload_path = force_str(nfd_str(upload.path))
-        ext = force_str(os.path.splitext(upload_path)[1])
-        file_.filename = file_.generate_filename(extension=ext or '.xpi')
         # Size in bytes.
         file_.size = storage.size(upload_path)
         file_.strict_compatibility = parsed_data.get('strict_compatibility', False)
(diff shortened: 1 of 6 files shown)


src/olympia/files/models.py
```diff
@@ -182,18 +182,21 @@ def generate_hash(self, filename=None):
...
+    def generate_filename(self):
         """
         Files are in the format of:
         {addon_name}-{version}-{apps}
         (-{platform} for some of the old ones from back when we had multiple
          platforms)
+
+        By convention, newly signed files after 2022-03-31 get a .xpi
+        extension, unsigned get .zip. This helps ensure CDN cache is busted
+        when we sign something.
```
eviljeff, line 194: how much of a concern are files that were uploaded prior to 2022-03-31, so have an `.xpi` extension even though they aren't signed, but may be then signed?  I understand it may not be an immediate problem until there other changes to how files are served but will there be a point where it is?  And if so is there a migration strategy for those older files?

eviljeff (approved):

src/olympia/files/models.py
diox: I've just added that to https://github.com/mozilla/addons-server/issues/11510 - I don't think it's a big concern but let's tackle that during the big file directory structure migration.

PR #19043, opened 2022-03-30
Make the update service use the regular logging settings

### LINKED ISSUE #19042: Prevent update service tests from breaking other tests
The update service currently calls [`log_configure()`](https://github.com/mozilla/addons-server/blob/a4cb2ba4313582e1938475983901cdfed0b0bb13/services/utils.py#L26-L45) [as soon as it's imported](https://github.com/mozilla/addons-server/blob/a4cb2ba4313582e1938475983901cdfed0b0bb13/services/update.py#L23).

It does this because it's a completely separate wsgi application, that doesn't depend on django: so if it wasn't configuring logging on its own, it wouldn't have any config set, the logs wouldn't be emitted properly.

Unfortunately, we run the [update service tests](https://github.com/mozilla/addons-server/blob/a4cb2ba4313582e1938475983901cdfed0b0bb13/src/olympia/addons/tests/test_update.py) as part of the rest of the test suite, and that call to configure logging is messing up the logging configuration for all tests running after that !

A couple possible fixes/workarounds:
- Separate the test runs completely. This is a bit annoying and not a 100% guarantee, since the problem occurs as soon as we import stuff from `services/`. OTOH we shouldn't do that...
- Move the `log_configure()` call to happen at runtime inside `application()`
  - On its own it's not enough, we'd also need to mock it when we do call `application()` in tests
  - A variation of this, move the call but also replace `log_configure` by something ensuring the logging config is shared with the one we use for the django app (remember though, the update service is not django, so it can't call `django.setup()`. It can depend on settings though)

diox:
Also configure logging at startup instead of import time to minimize impact on other bits of code, notably tests from the rest of addons-server

Fixes #19042


services/utils.py
```diff
@@ -21,25 +16,3 @@ def getconn():
 
 
 mypool = pool.QueuePool(getconn, max_overflow=10, pool_size=5, recycle=300)
```
wagnerand, line 18: Could also resolve the remainder of this file since it's only used in the update service and the corresponding test, or give the 90s their "My X" back.

diox: I was waiting for somebody to suggest that! Did it in 2ef813a587cbf4d23fb6f307f921ba4ac99ba384

diox:
https://github.com/mozilla/addons-server/pull/19045 is useful to test locally (you still need public listed add-ons and to craft the URL, though)

bobsilverberg (approved):

PR #19054, opened 2022-03-31
Link add-on acitivity on django admin add-on page

### LINKED ISSUE #19048: Add add-on activity to django admin add-on page
The django admin add-on page misses a link to the add-on activity.

wagnerand:
Fix #19048.

### DIFF
--- a/src/olympia/addons/tests/test_admin.py
+++ b/src/olympia/addons/tests/test_admin.py
@@ -96,6 +96,12 @@ class TestAddonAdmin(TestCase):
         self.admin_home_url = reverse('admin:index')
         self.list_url = reverse('admin:addons_addon_changelist')
 
+    def _call_related_content_method(self, method, addon):
+        model_admin = AddonAdmin(Addon, admin.site)
+        result = getattr(model_admin, method)(addon)
+        link = pq(result)('a')[0]
+        return link.attrib['href'], link.text
+
     def test_can_see_addon_module_in_admin_with_addons_edit(self):
         user = user_factory(email='someone@mozilla.com')
         self.grant_permission(user, 'Addons:Edit')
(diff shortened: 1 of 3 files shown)


src/olympia/addons/tests/test_admin.py
```diff
@@ -96,6 +96,12 @@ def setUp(self):
         self.admin_home_url = reverse('admin:index')
         self.list_url = reverse('admin:addons_addon_changelist')
 
+    def _call_related_content_method(self, method, addon):
```
eviljeff, line 99: Is this is only used in one place?  It seems overwritten for that (you wouldn't need the `getattr` stuff either)

src/olympia/addons/tests/test_admin.py
```diff
@@ -516,7 +522,7 @@ def test_query_count(self):
         self.grant_permission(user, 'Addons:Edit')
         self.grant_permission(user, 'Admin:Advanced')
         self.client.login(email=user.email)
-        with self.assertNumQueries(20):
+        with self.assertNumQueries(21):
```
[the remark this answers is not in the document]
wagnerand: Because we are fetching the count of activity logs to display on that page (see screenshot in the issue).

src/olympia/addons/tests/test_admin.py
eviljeff: IMO, we should either generalize that function and move it to `TestCase` (or just in `amo/tests/__init__.py`) so it'd be `def _call_admin_related_content_method(self, admin, method, instance)` (we can get which model with `admin.model`).  Or (easier) just put the functionality in just this class directly into `test_activity`.

eviljeff (approved):
r+ when tests pass

PR #19058, opened 2022-04-01
log all metadata updates via addon api

### LINKED ISSUE #19008: Make sure API calls for add-on and version changes emit a log activity.
https://github.com/mozilla/addons-server/issues/18036 and friends make several fields for add-ons and versions writable through the API. We should make sure those actions emit a log activity.

eviljeff:
fixes #19008 - this turned out to be a lot more involved than I thought.  

It renames `CHANGE_ICON` but otherwise replicates the logging that devhub has (largely - it at least logs the metadata fields that have changed, which devhub doesn't).

There's certainly an argument for fixing the inconsistencies in logging overall, but that will have to be a follow-up.  Inconsistencies found so far:
- icons and media have their own activity log type
- `EDIT_PROPERTIES` is used for all other add-on meta-data changes, but without specifying what's actually been changed (the patch includes the fields that have changes in the api - but is that enough?)
- tags have logging _per tag_ that's added or removed, and it's buried in the model's functions rather than the views.
- version changes don't have a generic `EDIT_PROPERTIES` style activity so only certain changes are logged
- tags and source code upload have activity logs on _create_, but all the others are only logged on update.
- `MAX_APPVERSION_UPDATED` is super specific compared to the other activity logs

### DIFF
--- a/src/olympia/addons/fields.py
+++ b/src/olympia/addons/fields.py
@@ -1,3 +1,4 @@
+import copy
 import os
 import tarfile
 import zipfile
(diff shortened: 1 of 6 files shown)


src/olympia/addons/fields.py
```diff
@@ -240,27 +241,27 @@ def to_internal_value(self, data):
...
                 internal = {}
                 for app_name, min_max in data.items():
                     app = amo.APPS[app_name]
-                    apps_versions = existing.get(
-                        app, ApplicationsVersions(application=app.id)
-                    )
+                    # we need to copy() to avoid changing the instance before save
+                    apps_versions = (
+                        (ex := existing.get(app)) and copy.copy(ex)
+                    ) or ApplicationsVersions(application=app.id)
```
eviljeff, line 250: this was kind of a bug in the existing code, but harmless - `to_internal_value` is supposed to only return the new value, and isn't supposed to change anything on the version instance, but it was, subtly.  It was harmless because it was only changing a cached property and the instance wasn't saved in any case.

eviljeff: It only became a problem when I needed to work out what had changed compared to the original instance for logging.

src/olympia/addons/serializers.py
```diff
@@ -535,15 +536,32 @@ def create(self, validated_data):
...
+                        amo.LOG.MAX_APPVERSION_UPDATED,
+                        instance.addon,
+                        instance,
+                        user=user,
+                        details={
+                            'version': instance.version,
+                            'target': appver.version.version,
+                            'application': appver.application,
+                        },
+                    )
```
eviljeff, line 564: fwiw, I don't like how the logging in the Version serializer has turned out - it's all over the place.  `source` has it's own logging because we send an email too; and `license` and the max version change needs to know information from both before the update and after 😞

bobsilverberg (approved):
This looks fine to me. I agree that having the logging all over the place seems less than ideal. Are you going to file a follow-up to clean that up and try to make it more centralized?

eviljeff:
> This looks fine to me. I agree that having the logging all over the place seems less than ideal. Are you going to file a follow-up to clean that up and try to make it more centralized?

I'll file a follow-up to the issue.

PR #19064, opened 2022-04-01
addons api metadata checks

### LINKED ISSUE #19018: Check all the functionality in devhub forms is replicated in the addon submission api
https://bugzilla.mozilla.org/show_bug.cgi?id=1761263

eviljeff:
fixes #19018 & https://bugzilla.mozilla.org/show_bug.cgi?id=1761263


src/olympia/addons/tests/test_views.py
```diff
@@ -1733,6 +1784,84 @@ def test_delete_icon(self, remove_icons_mock):
...
+        )
+        assert response.status_code == 200
+
+    @override_switch('metadata-content-review', active=False)
+    @mock.patch('olympia.addons.serializers.fetch_translations_from_addon')
+    def test_metadata_content_review_waffle_off(self, fetch_mock):
+        self._test_metadata_content_review()
+
+        fetch_mock.assert_not_called()
+        with self.assertRaises(AssertionError):
```
bobsilverberg, line 1800: Just a comment: This seems like an ugly way to assert "not called", but I assume a simpler version does not exist or we'd use it.

bobsilverberg (approved):
Looks good. I just have a comment on an assertion method, but I don't imagine it can be changed.

r+wc

src/olympia/addons/tests/test_views.py
eviljeff: yep.  You can `assert_not_called` on a mock, but asserting a _specific_ call wasn't made, while other calls were, means either doing this (ugly but at least clear, imo) or iterating through all the calls, extracting the call kwargs and args, and asserting that none of the calls have this arg.

PR #19091, opened 2022-04-06
 enable previews to be created, edited, deleted via addons api

### LINKED ISSUE #18236: Enable addon previews (screenshots) to be added/deleted via the addon api
split from #18036

We want to be able to add and remove (and edit) the addon previews/screenshots via the api, like we can in devhub.  Because there can be many we'll likley want a sub-endpoint under addon dedicated to creating, listing, deleting, etc, the previews.

eviljeff:
fixes #18236 with some refactoring (I organised the end patch into separate commits to make it easier to see what the core changes are)


eviljeff:
(circleci is currently down)

src/olympia/addons/tasks.py
```diff
@@ -358,3 +358,36 @@ def resize_icon(source, dest_folder, target_sizes, **kw):
...
+def resize_preview(src, preview_pk, **kw):
+    """Resizes preview images and stores the sizes on the preview."""
+    preview = Preview.objects.get(pk=preview_pk)
+    preview.sizes = {'thumbnail_format': amo.ADDON_PREVIEW_SIZES['thumbnail_format']}
+    thumb_dst, full_dst, orig_dst = (
+        preview.thumbnail_path,
+        preview.image_path,
+        preview.original_path,
+    )
+    log.info('[1@None] Resizing preview and storing size: %s' % thumb_dst)
```
bobsilverberg, line 374: Just curious, what is this `[1@None]`?

eviljeff: I just moved this function from devhub/tasks.py wholesale, but I believe it's to match the logging format of tasks that work on multiple object instances... not that I could find any other logging for tasks that worked with multiple objects that was a matching format.  @diox can you remember anything more?

eviljeff: thanks @diox.  It doesn't seem like any priority to change these log statements so I'm not going to file a follow-up to change them.

PR #19097, opened 2022-04-07
Remove puente: add necessary babel configuration instead

### LINKED ISSUE #19076: puente: end-of-life
I'm the current maintainer of [puente](https://github.com/mozilla/puente). I'm maintaining too many things and I haven't used Puente in years, so I'm looking at either:

* passing maintenance to someone else
* or, ending the project

Puente was built to make it possible for projects to transition off of Tower. It was always intended to be a temporary stop-gap to using existing tooling provided in Django, Babel, and Jinja2. It's possible it might not be needed anymore.

If you're interested in taking over maintenance, that'd be great! If no one takes over, I'll shut down the project in May 2022.

diox:
Fixes #19076


src/olympia/core/tests/test_babel.py
```diff
@@ -0,0 +1,26 @@
...
+    ],
+)
+def test_generate_option(value, expected):
+    assert generate_option(value) == expected
+
+
+def test_extract_jinja_no_options():
+    # Doesn't actually extract, just tests that extract_jinja() doesn't fail
+    # when converting our settings in options for the underlying function.
+    extract_jinja(BytesIO(), DEFAULT_KEYWORDS, ['L10n:'], {})
```
diox, line 26: It's out of scope for this but I think we should consider run a true extract (without merging the locale files) as part of the test suite. This would ensure we can actually run the extraction at all times on `master` branch.

locale/templates/LC_MESSAGES/django.pot
```diff
@@ -1,10 +1,14 @@
-
+# Translations template for addons-server.
+# Copyright (C) 2022 ORGANIZATION
+# This file is distributed under the same license as the addons-server project.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2022.
```
eviljeff, line 4: should these have non-default values?

locale/templates/LC_MESSAGES/djangojs.pot
```diff
@@ -1,10 +1,14 @@
-
+# Translations template for addons-server.
+# Copyright (C) 2022 ORGANIZATION
+# This file is distributed under the same license as the addons-server project.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2022.
```
eviljeff, line 4: same question as for django.pot

babel.cfg
```diff
@@ -0,0 +1,6 @@
+[python: src/olympia/**.py]
+[jinja2_custom: src/olympia/discovery/strings.jinja2]
+[django: src/olympia/**/templates/**/emails/**.*]
+[django: **/templates/admin/**.html]
+[django: **/templates/devhub/forms/widgets/compat_app_input_option.html]
```
eviljeff, line 5: should these also start `/src/olympia`, at least so there's no attempt to parse files elsewhere?

src/olympia/lib/settings_base.py
```diff
@@ -551,52 +555,10 @@ def get_db_config(environ_var, atomic_requests=True):
     'https://addons.mozilla.org/api/v5/shelves/editorial'
 )
 
-# Filename where the strings will be stored. Used in puente config below.
+# Filename where the strings will be stored. Used in extract_content_strings
+# management command, but note that the filename is hardcoded in babel.cfg.
```
eviljeff, line 559: it's a shame we can't define/generate all the babel config in one place 😞

eviljeff (approved):

locale/templates/LC_MESSAGES/django.pot
diox: I think we might want to change the `ORGANIZATION` bit but leave that line - it's supposed to be the "last translator" but it doesn't make sense to change in the template

(It's all autogenerated by `pybabel extract`)

src/olympia/lib/settings_base.py
diox: We could generate the babel config, but that's one more cog in the machine... Alternatively we could do what puente used to do and provide our own extract command, but I thought it'd be better if we tried to stay closer to how babel works...

src/olympia/core/tests/test_babel.py
diox: Commented in https://github.com/mozilla/addons-server/issues/9112#issuecomment-1094975106

PR #19109, opened 2022-04-12
Remove git storage migration code

### LINKED ISSUE #19108: Remove git storage migration code
The migration command added for https://github.com/mozilla/addons-server/issues/18896 can now be removed (it will potentially serve as inspiration for future storage migrations, but it doesn't have to stay in `master` branch).

diox:
We successfully ran the migration over a month ago.

Fixes #19108


src/olympia/git/migrations/0006_auto_20220412_1144.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 3.2.12 on 2022-04-12 11:44
+
+from django.db import migrations
+
+
+def delete_waffle_switch(apps, schema_editor):
+    Switch = apps.get_model('waffle', 'Switch')
+    Switch.objects.filter(name='git-storage-broader-structure').delete()
```
eviljeff, line 8: nit, but in the interests of DRY should we have a generic utility function that we could reuse for each time we want to delete a waffle switch?

eviljeff (approved):

eviljeff (approved):

PR #19119, opened 2022-04-13
addons api raise when empty app compatibility is specified for version create/update

### LINKED ISSUE #19093: Submitting an add-on with  `compatibility` set as an empty array or object will create an add-on without versions/ version compat
1. Upload a new listed add-on through the upload API
2. Set `compatibility` as an empty array  or object in the request JSON body, i.e. `"compatibility": []` or `"compatibility": {}` 
3. Submit the add-on

### What happened?
A 500 error is received. If I'm looking in DevHub after submission, I can see that the add-on was created without a version - https://addons-dev.allizom.org/en-US/developers/addon/my_addon-slugish25/versions
image

The API should return a 400 error and ask for `compatibility` to include at least one compatible application (i.e. firefox or android)

- In a PATCH version request, if I send `compatibility` as an empty array or object, the request is successful (200) and the version will end up without any compatibility set. This shouldn't be possible either:
image
- This issue happens even when a compatibility is specified in the add-on manifest.

eviljeff:
fixes #19093 - the change (raising when `len(data) == 0`) is much more minimal than the diff implies but I rearranged some of the code and took the opportunity to de-duplicate some of the tests in `TestVersionViewCreate` and `TestVersionViewUpdate` that were doing the same thing.

### DIFF
--- a/src/olympia/addons/fields.py
+++ b/src/olympia/addons/fields.py
@@ -238,42 +238,43 @@ class VersionCompatabilityField(serializers.Field):
             if isinstance(data, list):
                 # if it's a list of apps, normalize into a dict first
                 data = {key: {} for key in data}
-            if isinstance(data, dict):
-                version = self.parent.instance
-                existing = version.compatible_apps if version else {}
-                internal = {}
-                for app_name, min_max in data.items():
-                    app = amo.APPS[app_name]
-                    # we need to copy() to avoid changing the instance before save
-                    apps_versions = (
-                        (ex := existing.get(app)) and copy.copy(ex)
-                    ) or ApplicationsVersions(application=app.id)
-
-                    app_version_qs = AppVersion.objects.filter(application=app.id)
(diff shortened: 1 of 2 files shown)


src/olympia/addons/fields.py
```diff
@@ -238,43 +238,44 @@ def to_internal_value(self, data):
...
-                        apps_versions.min = app_version_qs.get(version=min_max['min'])
-                    elif version:
-                        apps_versions.min = app_version_qs.get(
-                            version=amo.DEFAULT_WEBEXT_MIN_VERSIONS[app]
-                        )
-
-                    internal[app] = apps_versions
-                return internal
-            else:
+            if not isinstance(data, dict) or len(data) == 0:
```
diox, line 241: Or maybe 
```suggestion
            if not isinstance(data, dict) or data == {}:
```

?

diox (approved):

PR #19121, opened 2022-04-13
prevent extension previews from being created for themes via the addons api

### LINKED ISSUE #19118: Throw an error for static themes while attempting api preview operations
Followup for https://github.com/mozilla/addons-server/issues/18236

 I ran into a situation: after I tried PATCH and DELETE on a theme id -> I get no response, Status and Time are blank. 
POST  create preview is returning status created but sending a preview for a theme has no impact in dev hub. 
I replaced the id of the theme with an extension after this but I still get no response for PATCH and DELETE  so I made new requests (into a new tab) which will work as expected. 

PATCH/DELETE after using a theme id/guid:

does not respond

Expected result:
POST, PATCH or DELETE should throw a warning when a static theme id/guid is used. 

API docs: 
https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#preview-create

eviljeff:
fixes #19118

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -56,7 +56,12 @@ from olympia.files.utils import parse_addon
 from olympia.files.tests.test_models import UploadMixin
 from olympia.tags.models import Tag
 from olympia.users.models import UserProfile
-from olympia.versions.models import ApplicationsVersions, AppVersion, License
+from olympia.versions.models import (
+    ApplicationsVersions,
+    AppVersion,
+    License,
+    VersionPreview,
+)
 
 from ..models import (
     Addon,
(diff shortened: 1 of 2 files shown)


src/olympia/addons/tests/test_views.py
```diff
@@ -5563,9 +5593,15 @@ def test_update(self):
         )
         data = {'caption': {'en-US': 'a thing', 'fr': 'un thíng'}, 'position': 1}
 
+        # can't patch if not authenticated
         response = self.client.patch(url, data=data)
         assert response.status_code == 401
 
+        # can't patch if not your add-on
```
bobsilverberg, line 5600: These assertions don't seem to be related to the patch. Are they just things that you noticed were missing, or am I misunderstanding? Also, should there be tests that previews for themes cannot be updated, or is that pointless because they cannot be created?

bobsilverberg (approved):

r+wc

src/olympia/addons/tests/test_views.py
eviljeff: I should have added them earlier. They're kind of redundant because it's the same auth+permission setup used for Addon and other sub-Addon views, so there's no reason it wouldn't work, but it should be tested to be sure.

PR #19138, opened 2022-04-21
Fix spelling of add-on(s) in strings

flodolo:
See guideline for add-ons in [our wordlist](https://mana.mozilla.org/wiki/pages/viewpage.action?spaceKey=FIREFOX&title=Word+List#WordList-add-ons).

Feel free to use this as a starting point, not sure how many tests I broke in the process. I've only traced down the strings from the .pot file.


diox (approved):
:+1:

PR #19142, opened 2022-04-21
Pass user instead of request in most acl checks

diox:


bobsilverberg (approved):
Looks good. Not that I want the patch to get even bigger (so maybe it could be a follow-up if you think it makes sense), but I was wondering if you want to change the name of `check_addon_ownership` as well, as you renamed a bunch of `check...` functions in this patch.

r+wc

diox:
> I was wondering if you want to change the name of `check_addon_ownership` as well, as you renamed a bunch of `check...` functions in this patch.

I would love to, maybe as a follow-up, but I'm not sure what to name it. It checks a bunch of things, not just ownership really... so it's hard to come up with a better name. It's sort of a `is_owner_of_non_deleted_non_disabled_addon_or_user_with_addons_edit_permission_or_pass_parameters_allowing_regular_authors_or_disabled_addons()` :)

wagnerand:
> > I was wondering if you want to change the name of `check_addon_ownership` as well, as you renamed a bunch of `check...` functions in this patch.
> 
> I would love to, maybe as a follow-up, but I'm not sure what to name it. It checks a bunch of things, not just ownership really... so it's hard to come up with a better name. It's sort of a `is_owner_of_non_deleted_non_disabled_addon_or_user_with_addons_edit_permission_or_pass_parameters_allowing_regular_authors_or_disabled_addons()` :)

That sounds like that method is doing too much and should be split? :)

diox:
It does, but I don't have the energy for that refactor.

Issue #19055, opened 2022-03-31
Name and Summary can be `null` in the add-on submission API


AlexandraMoga:
### Describe the problem and steps to reproduce it:
1. Upload a new listed add-on through the upload API
2. Set the `name` and `summary` to `null` in the request JSON body - i.e. `"name"/"summary": {"en-US": null}`
3. Submit the add-on

### What happened?
The submission is successful. The add-on created will not have a name and summary. 

### What did you expect to happen?
A 400 error with the following message:  "This field may not be blank."

### Anything else we should know?
If `name` or `"summary"` is set as an empty string, we get a 400 error, as expected.

AlexandraMoga:
I've updated the issue to also include the `name` property, which shows the same behavior.

ioanarusiczki:
@eviljeff I'm also checking on https://github.com/mozilla/addons-server/issues/8816 and I tried this scenario with POST - create listed addon and PATCH on name and summary. 
- when sending "summary":{"en-US":null},"name":{"en-US":null} at addon creation I still get a 201 Created, the submission can be resumed from dev hub since there's no name or summary for the default locale. 
- PATCH with "summary":{"en-US":null},"name":{"en-US":null} throws an error message         "A value in the default locale of \"en-US\" is required."

eviljeff:
> @eviljeff I'm also checking on #8816 and I tried this scenario with POST - create listed addon and PATCH on name and summary.
> 
>     * when sending "summary":{"en-US":null},"name":{"en-US":null} at addon creation I still get a 201 Created, the submission can be resumed from dev hub since there's no name or summary for the default locale.

what is the upload xpi you're submitting - maybe there's something about the manifest structure that's breaking some check.

ioanarusiczki:
[kanji.zip](https://github.com/mozilla/addons-server/files/8610228/kanji.zip)

@eviljeff it's just a dummy extension I'm using.

eviljeff:
So, effectively what's happening is the validation code is ignoring the `null` value but seeing the `name` as being present because there's a value in the manifest; whereas `name` dict (with the `null` value(s)) is then overriding the `name` from the manifest later.  I.e. there's a disconnect between what the validation step is doing and what the actual add-on creation step does.

ioanarusiczki:
@eviljeff 
when I'm  sending "summary":{"en-US":null},"name":{"en-US":null} at addon creation I get a 400 Bad Request,with the error messages:
{
    "name": [
[...]

PR #19150, opened 2022-04-22
fix deleting localized field values

### LINKED ISSUE #8816: Allow translations of localised API fields to be deleted
Our current API implementation has trouble clearing translations of localised fields:
* if a field is required then an update that sends a translation locale=None is treated as an invalid input.
* if the field isn't required then just sending None (i.e. a flat string style input, but None) is ignored by DRF (will be fixed in  /v4/ see #8794)
* you can set a translation for the default_locale of the object to be None but you shouldn't be able to if there are other translations in other locales (or ever if it's a required field)

From testing during #8794 it's not clear if sending locale=None to delete locales is working properly at all, even in the general case (the `Translation` objects still exist, for example).  But needs more investigation.

eviljeff:
fixes #8816 - I added tests for the field and in AddonViewSetUpdate for real world use.  Let me know if there are any more edge cases that needs tests adding to cover.

(this would also address #19055 but `name` and `summary` are not set as `required=True` because we want to fall back to the manifest values so further work is needed there.)

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -1381,6 +1381,61 @@ class TestAddonViewSetUpdate(AddonViewSetCreateUpdateMixin, TestCase):
         assert alog.action == amo.LOG.EDIT_PROPERTIES.id
         assert alog.details == ['summary']
 
+    @override_settings(API_THROTTLING=False)
+    def test_translated_fields(self):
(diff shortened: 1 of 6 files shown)


bobsilverberg (approved):
Looks good. Thanks for taking care of this. I recall this being an issue that has been discussed for quite some time.

Issue #19152, opened 2022-04-25
Remove code related to the "wat" scanner


willdurand:
We never enabled _wat_ in prod so let's remove the code for it.

**For QA:** please verify that:

1. there is no scanner result for "wat" in the admin
2. the  `enable-wat` waffle switch no longer exists
3. other scanner results still exist (customs/yara should still be listed in the admin)
4. other waffle switches should still exist (only the waffle switch mentioned in (2) should be removed)

willdurand:
We should also get rid of the following env vars in all envs:

- `WAT_API_URL`
- `WAT_API_KEY`

willdurand:
I filed a JIRA ticket for the ops side (deployment config)

AlexandraMoga:
I've verified the scanners on -dev and I haven't found any issues.
There are no `wat` scanners results available and the other scanners are running as expected.

The only place where `wat` is still mentioned are the filters, but this can be done in another patch:
![image](https://user-images.githubusercontent.com/31961530/166238037-ae4f38cc-4713-4fde-9cf4-fad5b6e5ba39.png)

willdurand:
> The only place where `wat` is still mentioned are the filters, but this can be done in another patch:

that is expected

PR #19153, opened 2022-04-25
Remove code related to the wat scanner

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/19152

---

I haven't removed the constants in `src/olympia/constants/scanners.py`
since it is used in the model definition. ~~I suppose we could add a
migration to delete all the scanner results with `scanner=WAT` and
remove the constant + update the model in a follow-up patch?~~


diox:
> I haven't removed the constants in `src/olympia/constants/scanners.py` since it is used in the model definition. I suppose we could add a migration to delete all the scanner results with `scanner=WAT` and remove the constant + update the model in a follow-up patch?

I'd keep the constant, prefix it with an underscore + comment explaining not to use it (just to avoid confusion in the future), and add a cleanup migration. Could be done in this patch or a follow-up.

willdurand:
> I'd keep the constant, prefix it with an underscore + comment explaining not to use it (just to avoid confusion in the future), and add a cleanup migration. Could be done in this patch or a follow-up.

done

src/olympia/scanners/migrations/0046_delete_waffle_switch.py
```diff
@@ -0,0 +1,14 @@
+from django.db import migrations
+
+
+def remove_waffle_switch(apps, schema_editor):
+    Switch = apps.get_model('waffle', 'Switch')
+
+    Switch.objects.filter(name='enable-wat').delete()
```
diox, line 7: We can't do that until after the code has been removed, otherwise during the push we would potentially enable wat for a very short moment during which the old code is still running on the migrated db.

So this migration (and maybe the other one, it matters less) should be in a follow-up issue scheduled for the following push after the one in which this PR lands.

diox: (Sorry, should have caught this earlier when you mentioned running a clean-up migration)

willdurand: > We can't do that until after the code has been removed, otherwise during the push we would potentially enable wat for a very short moment during which the old code is still running on the migrated db.

How could that happen? I am not sure the pref has been created on -prod for instance.

diox: Err you're right, disregard me. The default would be for the switch not to be active, so we're fine...

diox (approved):

PR #19168, opened 2022-04-26
Reload version in auto_approve before determining auto-approval verdict

### LINKED ISSUE #19136: Reload version in `auto_approve` before determining whether or not it should be auto-approved
`auto_approve` fetches versions that are auto-approvable but then can take a long time to go through them all, so instead, it should fetch version ids and instantiate each version from the database at the last minute, just before determining whether or not it should be auto-approved.

diox:
The general idea is to improve consistency checks and avoid accidentally auto-approving an add-on that isn't in the right state for it to happen.

In order to do that, on top of loading the version at the last possible time, a refactor of `ReviewHelper` was necessary to make non human reviews more explicit and check add-on status before making the approval action available, which allows auto_approve to check for action availability instead of bypassing that check as it did before.

Fixes #19136

### DIFF
--- a/src/olympia/reviewers/management/commands/auto_approve.py
+++ b/src/olympia/reviewers/management/commands/auto_approve.py
@@ -36,6 +36,10 @@ log = olympia.core.logger.getLogger('z.reviewers.auto_approve')
 LOCK_NAME = 'auto-approve'  # Name of the lock() used.
 
 
+class ApprovalNotAvailableError(Exception):
+    pass
+
+
 class Command(BaseCommand):
     help = 'Auto-approve add-on versions based on predefined criteria'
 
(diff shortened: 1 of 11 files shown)


src/olympia/reviewers/management/commands/auto_approve.py
```diff
@@ -51,9 +55,13 @@ def add_arguments(self, parser):
         )
 
     def fetch_candidates(self):
```
bobsilverberg, line 57: Nite: Would it make sense to rename this to `fetch_candidate_ids`? Maybe that's not necessary.

src/olympia/reviewers/utils.py
```diff
@@ -755,17 +765,17 @@ def process(self):
 
 class ReviewBase:
     def __init__(
-        self, request, addon, version, review_type, content_review=False, user=None
+        self,
+        *,
```
bobsilverberg, line 769: What is the `*` used for in the class?

bobsilverberg (approved):
Looks good. I just have one question.

r+wc

src/olympia/reviewers/utils.py
diox: It's to force the arguments that follow to be keyword-only arguments. Prevents you from accidentally using positional arguments, making calls more explicit.

src/olympia/reviewers/management/commands/auto_approve.py
diox: With the adjusted docstring I don't think it's necessary. It's unlikely that method would ever be used outside of this class anyway.

PR #19171, opened 2022-04-27
When a reindex is taking place, only index on old and new indexes when necessary

### LINKED ISSUE #19165: Rendered license name does not match the actual addon's license
The rendered "License" is not the actual license. It says MPL but the hyperlink is GPL

### What happened?
I went to https://addons.mozilla.org/addon/ublock-origin/
The "License" section under "More Information" says MPL

I expected the "License" section under "More Information" to say GPL 3.0

diox:
2 types of indexing calls can be made while a reindex is taking place:
- Regular indexing from modifications made to an object. Those need to happen on both the old and new index so that changes are reflected immediately on the old index we're currently using, and available on the new one when we switch to it at the end of the reindex.
- indexing triggered by the reindex task group. Those don't need to happen on the old index: we only care about the new one.

Before this patch, we would index the objects on both the old and new index even for indexing tasks triggered by reindex. On top of being unnecessary, this was a problem if the extraction code is no longer compatible with the old mapping.

Fixes #19165 (unblocks the reindex)


src/olympia/amo/tests/__init__.py
```diff
@@ -1013,6 +964,56 @@ def setUpClass(cls):
         # right before each test.
         stop_es_mocks()
         cls.es = amo_search.get_es()
+        # Make sure ES cluster is in a good state, resetting the index if
```
eviljeff, line 967: what does refactoring this code from `setup_es_data` (that was called from `setUpTestData` instead) change?

eviljeff (approved):

src/olympia/amo/tests/__init__.py
diox: `setUpTestData` was not called in `TestIndexCommand` because that's a `TransactionTestCase`, and `setUpTestData` is a `TestCase` thing. Moving it to `setUpClass` ensures it is called, which in turns is useful to ensure the tests start from a clean state.

The problem wasn't visible in CI, but locally if you ran the tests multiple times and had some test failures while developing, you could end up in a state where the `TestIndexCommand` would repeatedly fail because an existing index was set up already. Before that change, the only way to get out of that was to execute _another_ test with elasticsearch enabled that wasn't a `TransactionTestCase`, or manually clean up your ES indexes...

PR #19185, opened 2022-05-02
Remove easy-thumbnails dependency #18679

### LINKED ISSUE #18679: Remove easy-thumbnails dependency
We depend on `easy-thumbnails` only for the following:
https://github.com/mozilla/addons-server/blob/f7db392878567b7332f362534c90fc7ca84c5ca1/src/olympia/amo/utils.py#L648-L649

So, a single function: `scale_and_crop` from https://github.com/SmileyChris/easy-thumbnails/blob/7ee6219f180c2f73b22b6f4a91c1647fff4696b4/easy_thumbnails/processors.py#L115-L270

Notably, we always pass a `size`, and we never pass `crop`, `upscale`, `zoom` or `target` argument. So we should be able to replace `scale_and_crop` entirely by:
- A check on the requested size being too small compared to the image size:
  - If the image is already small, do nothing
  - If it's not:
    - `im = im.convert('RGBA')`
    - `im = im.resize(...)` call with the right parameters for the x and y size, passing `resample=Image.ANTIALIAS` to keep doing antialiasing.

We should already have enough tests covering this. To manually test it, upload various images to devhub (in various sizes, aspect ratios and formats) and check that the thumbnails are generated correctly.

Vish-Yadlapalli:
Fixes #18679

Calculate scale based on a target to source size ratio and use the scale to calculate output dimensions. Used `Lanczos` filter as `Antialias` is renamed to `Lanczos` in PIL.

Kept the code similar to the `easy-thumbnails` code and it passed the tests in my local dev environment.

### DIFF
--- a/requirements/system.txt
+++ b/requirements/system.txt
@@ -97,9 +97,9 @@ idna==2.10 \
 packaging==21.3 \
     --hash=sha256:ef103e05f519cdc783ae24ea4e2e0f508a9c99b2d4969652eed6a2e1ea5bd522 \
     --hash=sha256:dd47c42927d89ab911e606518907cc2d3a1f38bbd026385970643f9c5b8ecfeb
-pip==22.1.1 \
-    --hash=sha256:8dfb15d8a1c3d3085a4cbe11f29e19527dfaf2ba99354326fd62cec013eaee81 \
-    --hash=sha256:e7bcf0b2cbdec2af84cc1b7b79b25fdbd7228fbdb61a4dca0b82810d0ba9d18b
+pip==22.1.2 \
+    --hash=sha256:6d55b27e10f506312894a87ccc59f280136bad9061719fac9101bdad5a6bce69 \
+    --hash=sha256:a3edacb89022ef5258bf61852728bf866632a394da837ca49eb4303635835f17
 # pycparser is required by cffi
 pycparser==2.21 \
     --hash=sha256:8ee45429555515e1f6b185e78100aea234072576aa43ab53aefcae078162fca9 \


Vish-Yadlapalli:
Where can I find more details on why the `codestyle` test is failing?

diox:
You need to run `black` on `src/olympia/amo/utils.py`

src/olympia/amo/utils.py
```diff
@@ -642,7 +641,21 @@ def resize_image(source, destination, size=None, *, format='png', quality=80):
...
+                im = im.convert('RGBA')
+                im = im.resize(
+                    (
+                        int(round(source_width * scale)),
+                        int(round(source_height * scale)),
+                    ),
+                    resample=Image.LANCZOS,
+                )
+        except ValueError as value_error:
+            print(value_error)
```
diox, line 658: We don't want to leave `print()`s around. In fact, we probably don't want to `try/except` at all here: if a `ValueError` is raised from the code above something is wrong in the logic and should be handled there rather than hide the error.

src/olympia/amo/utils.py
```diff
@@ -642,7 +641,21 @@ def resize_image(source, destination, size=None, *, format='png', quality=80):
...
+            source_width, source_height = [float(v) for v in im.size]
+            scale = min(target_width / source_width, target_height / source_height)
+            if scale <= 1:
+                im = im.convert('RGBA')
+                im = im.resize(
+                    (
+                        int(round(source_width * scale)),
+                        int(round(source_height * scale)),
+                    ),
+                    resample=Image.LANCZOS,
```
diox, line 655: Could you add a comment explaining that it's actually antialias ?

src/olympia/amo/utils.py
```diff
@@ -642,7 +641,21 @@ def resize_image(source, destination, size=None, *, format='png', quality=80):
             im.load()
     original_size = im.size
     if size:
-        im = processors.scale_and_crop(im.convert('RGBA'), size)
+        try:
+            target_width, target_height = [int(v) for v in size]
+            source_width, source_height = [float(v) for v in im.size]
```
eviljeff, line 646: You don't need to `int()` the `size` values - they're from a constant.  
`float()` on the im.size values seems unnecessary also as python automatically returns a float from a `/` operation even if both numbers are integers.

eviljeff: This will be enough to expand
```suggestion
            target_width, target_height = size
            source_width, source_height = im.size
```

Vish-Yadlapalli:
@eviljeff I'll finish it this week.

PR #19191, opened 2022-05-02
addons api (pending) authors create and confirm endpoint

### LINKED ISSUE #19164: Allow addon authors to be added via addon api
split from https://github.com/mozilla/addons-server/issues/18231 (which was itself spun off from https://github.com/mozilla/addons-server/issues/18036)

You should be able to add new add-on owners/developers via the addon api

Complexity here is 
- not actually creating AddonUser, but creating AddonUserPendingConfirmation.
- we need to expose any AddonUserPendingConfirmation somewhere.  It would _seem_ like the ideal place is the same `/authors/` sub-endpoint but the models themselves are different and not (easily) exposed with a single queryset (there is django magic but probably complete overkill)
- how do users confirm the pending confirmation?

eviljeff:
fixes #19164 
- includes a slight reorg of the /addon/ api docs because they were getting loooong.


docs/topics/api/authors.rst
```diff
@@ -0,0 +1,226 @@
...
+
+
+---------------------
+Pending Author Create
+---------------------
+
+.. _addon-pending-author-create:
+
+This endpoint allows an owner to invite a user to become an author of an add-on.
+A pending author is created for the add-on, and once they confirm the invitation, they will an author.
```
bobsilverberg, line 113: I think you're missing a word. At the end here it says "they will an author."

src/olympia/addons/serializers.py
```diff
@@ -697,25 +699,23 @@ class Meta(BaseUserSerializer.Meta):
 
 
 class AddonAuthorSerializer(serializers.ModelSerializer):
-    user_id = serializers.IntegerField(source='user.id', read_only=True)
```
bobsilverberg, line 700: I'm curious about why this was deleted.

src/olympia/addons/tests/test_views.py
```diff
@@ -5846,3 +5848,189 @@ def test_delete(self):
...
+
+        self.client.login_api(user_factory())
+        assert self.client.delete(self.detail_url).status_code == 403
+
+        self.client.login_api(self.pending_author.user)
+        assert self.client.get(self.detail_url).status_code == 403
+
+        self.client.login_api(self.user)
+        response = self.client.delete(self.detail_url)
+        assert response.status_code == 204
```
bobsilverberg, line 5930: Is it worth asserting the pending author is no longer there?

bobsilverberg (approved):
Looks good. I just have a couple of questions and a comment.

r+wc

src/olympia/addons/serializers.py
eviljeff: I realised from tinkering around it wasn't necessary - `user_id` would be serialized into an integer anyway by the `ModelSerializer` code.

Issue #19207, opened 2022-05-05
Replace `File.filename` with a `FileField` to manipulate files the django way


diox:
Should be completely transparent to users, but make it easier to manipulate paths behind the scenes.

Dependency of #11510

diox:
QA: It should be a transparent change. Do check that uploading, downloading files still work as normal, as well as code search.

PR #19208, opened 2022-05-05
Simplify `pretty_filename` implementation ; use it when we display filenames

diox:
Make `File.pretty_filename` a property, and make it the main way to display filenames in URLs & UI, to prepare for the replacement of the filename field later.

See #19207 (this is split from a bunch of other changes to fix this issue)

### DIFF
--- a/src/olympia/files/models.py
+++ b/src/olympia/files/models.py
@@ -99,7 +99,7 @@ class File(OnChangeMixin, ModelBase):
         # See https://github.com/mozilla-mobile/fenix/blob/
         # 07d43971c0767fc023996dc32eb73e3e37c6517a/app/src/main/java/org/mozilla/fenix/
         # AppRequestInterceptor.kt#L173
-        kwargs = {'file_id': self.pk, 'filename': self.filename}
+        kwargs = {'file_id': self.pk, 'filename': self.pretty_filename}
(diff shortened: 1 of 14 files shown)


src/olympia/files/models.py
```diff
@@ -203,19 +202,10 @@ def generate_filename(self):
...
-        m = self._pretty_filename.match(self.filename)
-        if not m:
-            return self.filename
-        if len(m.group('slug')) < maxlen:
-            return self.filename
-        return '{}...{}'.format(m.group('slug')[0 : (maxlen - 3)], m.group('suffix'))
+    @property
+    def pretty_filename(self):
+        """Displayable filename."""
+        return os.path.basename(self.filename) if self.filename else ''
```
eviljeff, line 208: It's quite a lot less functional than the existing `pretty_filename` - it doesn't sanitize or trim to a max length.  I guess that's okay though? (we do the sanitizing earlier and wrap in css?)

eviljeff (approved):

src/olympia/files/models.py
diox: Yeah. Ultimately it was only used in a single place - devhub - everywhere else was already using the raw filename...

PR #19201, opened 2022-05-04
Add a `FileField` to our `File` class to manipulate files the django way

diox:
Fixes https://github.com/mozilla/addons-server/issues/19207

PR #19209, opened 2022-05-05
Remove platforms suffix in generated filenames

diox:
We no longer need to differentiate files per platform as there is only a single file per version now.

See https://github.com/mozilla/addons-server/issues/19207 (this is split from a bunch of other changes to fix this issue)


diox:
Splitting this from https://github.com/mozilla/addons-server/pull/19201 doesn't buy us much in terms of diff size since this diff is super small and the filename generation code and json fixtures have to be changed again in the main PR, so I'm closing this.

PR #19211, opened 2022-05-05
In tests, automatically copy over files when using file_factory() with a filename

diox:
See https://github.com/mozilla/addons-server/issues/19207 (this is split from a bunch of other changes to fix this issue)


src/olympia/files/models.py
```diff
@@ -116,7 +116,7 @@ def from_upload(cls, upload, version, parsed_data=None):
         assert parsed_data is not None
 
         file_ = cls(version=version)
-        upload_path = force_str(nfd_str(upload.path))
```
diox, line 119: Drive-by change, but `upload.path` nowadays only contains ascii - plus we're not running Python 2 anymore, unicode paths just work.

src/olympia/zadmin/tests/test_views.py
```diff
@@ -114,38 +113,29 @@ class TestRecalculateHash(TestCase):
...
-        version = Version.objects.create(addon_id=3615)
-        file = File.objects.create(filename='https-everywhere.xpi', version=version)
+        file = version_factory(
+            addon=self.addon, file_kw={'filename': 'https-everywhere.xpi'}
+        ).file
 
-        r = self.client.post(reverse('zadmin.recalc_hash', args=[file.id]))
-        assert json.loads(r.content)['success'] == 1
+        response = self.client.post(reverse('zadmin.recalc_hash', args=[file.id]))
+        assert json.loads(response.content)['success'] == 1
```
eviljeff, line 125: nit, but `response.json`

eviljeff (approved):
question: what about tidy up after each test?  If we're copying over fixture files each time a `file_factory` is passed a `filename` are they removed somewhere?

src/olympia/addons/tests/test_serializers.py
```diff
@@ -1438,20 +1438,15 @@ def test_basic_dict(self):
         assert 'current_compatible_version' not in result
 
     def test_current_compatible_version(self):
-        # Set a filename to make sure the file actually exists.
-        # file_factory (used via addon_factory) copies files that exists
-        # as fixtures in src/olympia/files/fixtures/files to their rightful
-        # place.
-        file_kw = {'filename': 'langpack-localepicker.xpi'}
-        self.addon = addon_factory(type=amo.ADDON_LPAPP, file_kw=file_kw)
+        self.addon = addon_factory(type=amo.ADDON_LPAPP)
```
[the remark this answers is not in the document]
diox: It doesn't actually exist on disk and wasn't necessary for the test.

PR #19220, opened 2022-05-11
send emails & log when authors change

### LINKED ISSUE #19216: Send the invitation email to the new author
As explained into https://github.com/mozilla/addons-server/issues/19164#issuecomment-1122219093

The expected is:
that there should be an invitation sent on email when an author(owner) of an addon is creating it using the [API](https://addons-server.readthedocs.io/en/latest/topics/api/authors.html#pending-author-create)

eviljeff:
fixes #19216

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -1,8 +1,10 @@
 import os
 import re
 
+from django.conf import settings
 from django.core.exceptions import ValidationError as DjangoValidationError
 from django.core.files import File as DjangoFile
+from django.template import loader
 from django.urls import reverse
 from django.utils.translation import gettext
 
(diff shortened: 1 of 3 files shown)


bobsilverberg (approved):
Looks good. I find it curious that some of the logic lives in a serializer (create and update), while some lives in a view (delete), but I guess that's because there is no code for create and update in the view (it just uses the default DRF code)?

eviljeff:
> Looks good. I find it curious that some of the logic lives in a serializer (create and update), while some lives in a view (delete), but I guess that's because there is no code for create and update in the view (it just uses the default DRF code)?

yep.  DRF steers you to putting all the logic in the serializer, but there's no such thing as a `.delete` function that's called in the serializer (by default... I suppose I could add one).  Or may be _all_ the emailing and logging should in the viewsets so it's consistent?  mmm.  Refactoring ahead.

PR #19221, opened 2022-05-11
Use slug and not name when generating filenames

diox:
Slugs don't depend on translations and have a db-enforced max length

Follow-up to https://github.com/mozilla/addons-server/issues/11510 to ensure filenames are a bit more predictable & safe.

### DIFF
--- a/src/olympia/files/models.py
+++ b/src/olympia/files/models.py
@@ -51,7 +51,7 @@ def files_upload_to_callback(instance, filename):
     called saved yet.
 
     For new uploads, the returned path is:
-    {directories}/addon_name}-{version}.{extension}, where {directories} is
+    {directories}/addon_slug}-{version}.{extension}, where {directories} is
     {addon_id last 2 digits}/{addon_id last 4 digits}/{addon_id}
 
     For older uploads, the returned path used to be in the format of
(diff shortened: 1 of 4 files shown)


wagnerand:
Wouldn't it be better to use something immutable, e.g. ID or GUID?

diox:
Ultimately the filename is stored in the database, so we don't have to have something immutable. And the slug is going to make for a prettier filename than the id or guid (which matters since it will be returned to users... I suppose we could override that, but I'm not sure it's worth it...)

src/olympia/files/models.py
```diff
@@ -68,9 +68,9 @@ def files_upload_to_callback(instance, filename):
     a filename, but the filename is completely ignored here (it's meant to
     represent the user-provided filename in user uploads).
     """
-    # slugify drops unicode so we may end up with an empty string.
-    # Apache did not like serving unicode filenames (bug 626587).
-    name = slugify(instance.addon.name).replace('-', '_') or 'addon'
+    # Start with the add-on slug, but make it go through django's slugify() to
+    # drop unicode characters.
```
eviljeff, line 72: _is_ there still a problem with filename's with unicode characters in them?  The original comment referenced Apache and the bug talks about release mirrors, both of which aren't a thing in our current stack, iirc.

eviljeff (approved):

src/olympia/files/models.py
diox: It's probably fine, and we do have a few tests that prove we know how to deal with unicode filenames if necessary, but then again, non-ascii in things like `Content-Disposition` header is a mess, so I don't want to change that now...

Issue #19243, opened 2022-05-14
Addons update won't install - failed: 404 Not Found


Juraj-Masiar:
### Describe the problem and steps to reproduce it:
I think something's wrong with the addons update server because my Firefox ESR, Developer Edition and Nightly can't update any of the addons.
1) open browser console with "Ctrl + Shift + J"
1) open `about:addons` page
1) click gear icon / Check for Updates

### What happened?
It says it's updating, it even puts a blue dot next to addons that has updates, but addons are not updated and in the browser console are warnings:
```
1652511844168	addons.xpi	WARN	Download of https://addons.mozilla.org/firefox/downloads/file/3948477/38/1438/481438/i_dont_care_about_cookies-3.4.0.xpi failed: 404 Not Found
1652511844169	addons.xpi	WARN	Download of https://addons.mozilla.org/firefox/downloads/file/3948722/80/4480/1474480/movie_meter-3.0.xpi failed: 404 Not Found
1652511844181	addons.xpi	WARN	Download of https://addons.mozilla.org/firefox/downloads/file/3948047/65/1865/1865/adblock_plus-3.13.xpi failed: 404 Not Found 
```

### What did you expect to happen?
Update addons

### Anything else we should know?
(Please include a link to the page, screenshots and any relevant files.)

diox:
The update service is returning incorrect URLs after https://github.com/mozilla/addons-server/issues/11510

diox:
(The fix has been committed, but not deployed yet)

diox:
This can be seen by requesting updates for an add-on that has had new versions approved since the push, like adblock plus:
https://versioncheck.addons.mozilla.org/update/?reqVersion=1&id={d10d0bf8-f5b5-c8b4-a8b2-2b9879e08c5d}&version=3.12&appID={ec8030f7-c20a-464f-9b0e-13a3a9e97384}&appVersion=100.0

The expected update_link here should be `https://addons.mozilla.org/firefox/downloads/file/3948047/adblock_plus-3.13.xpi`
The actual update_link is `https://addons.mozilla.org/firefox/downloads/file/3948047/65/1865/1865/adblock_plus-3.13.xpi`

diox:
We just deployed a fix to production. It might take up to one hour for caching to expire though.

diox:
Thanks a lot for reporting this, @Juraj-Masiar !

PR #19244, opened 2022-05-14
Fix update_link with new directory structure in update service

diox:
Fixes https://github.com/mozilla/addons-server/issues/19243


willdurand:
(FTR: I has silently read the patch before it landed and got deployed over the weekend)

PR #19257, opened 2022-05-18
add api auth to download_file view

### LINKED ISSUE #19237: The addons api doesn't provide an API accessible download API for non-public files
The addons api (in particular the [version detail](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#version-detail) output) only exposes a url (`file.url`) that relies on browser based cookie auth - i.e. the standard add-ons server view also used for devhub (and other internal tooling). We _don't_ expose a file url that can be used by the api with [api auth](https://addons-server.readthedocs.io/en/latest/topics/api/auth.html#authentication-external).  This only applies to non-public files (not signed and/or not approved) as publicly accessible downloads don't require auth.

The older signing api has it's own [download endpoint](https://addons-server.readthedocs.io/en/latest/topics/api/signing.html#downloading-signed-files) that's a light wrapper around the same add-ons view but that's (afaik) only referenced within the signing api responses.

Options:
A) enhance the standard download view to support API auth in addition to django auth.  This could be non-trivial as there are many assumptions about api auth applying to urls under /api/ and cookie auth for others.  Essentially it'd be a hack to force checking api auth in a non-api view.
B) link to the signing api's download endpoint in the version detail serializer response (for authenticated developers).
C) re-implement a new download view/sub-view somewhere under the addons api that would possibly also reimplement some of the standard download view (and/or refactor).

eviljeff:
fixes #19237

### DIFF
--- a/src/olympia/amo/decorators.py
+++ b/src/olympia/amo/decorators.py
@@ -6,6 +6,9 @@ from django import http
 from django.conf import settings
 from django.core.exceptions import PermissionDenied
 
+from rest_framework import exceptions as drf_exceptions
+from rest_framework.settings import api_settings
+
 import olympia.core.logger
 
 from . import models as context
@@ -188,10 +191,6 @@ def api_authentication(f):
     already have been attempted by this point so api auth will only be tried for
     anonymous (unauthenticated) requests."""
 
-    from rest_framework import exceptions as drf_exceptions
-    from rest_framework.settings import api_settings
-    from rest_framework.views import exception_handler as drf_exception_handler
-
     from olympia.api.authentication import (
         get_authorization_header,
         SessionIDAuthentication,
@@ -212,7 +211,7 @@ def api_authentication(f):
             except drf_exceptions.AuthenticationFailed as exc:
                 # We have to set some props DRF would usually set in the APIView
                 exc.auth_header = api_auth.authenticate_header(request)
-                response = drf_exception_handler(exc, None)
+                response = api_settings.EXCEPTION_HANDLER(exc, None)
                 response.accepted_renderer = api_settings.DEFAULT_RENDERER_CLASSES[0]()
                 response.accepted_media_type = response.accepted_renderer.media_type
                 response.renderer_context = {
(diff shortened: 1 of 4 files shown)


src/olympia/amo/decorators.py
```diff
@@ -181,3 +181,48 @@ def wrapper(request, *args, **kw):
...
+            # if user isn't authenticated with standard auth, try the API auth methods
+            try:
+                for api_auth in (SessionIDAuthentication, JWTKeyAuthentication):
+                    api_auth = api_auth()
+                    result = api_auth.authenticate(request)
+                    if result:
+                        request.user, _ = result
+                        break
+            except drf_exceptions.AuthenticationFailed as exc:
+                # We have to set some props DRF would usually set in the APIView
```
diox, line 213: Do we need that/care ?

eviljeff: the comment or setting those properties?

diox: setting those properties - given that the underlying view will _not_ be an API view

eviljeff: it's because we're explicitly returning an API style error response rather than just re-raising and returning a normal django view response - I figured that was more "correct" as any client ideally doesn't want to have to handle an entirely different auth error response just for that view.

diox:
I want to do a pass at it since I was the one pushing for this approach. You can leave the full review to me if you want.

src/olympia/amo/decorators.py
```diff
@@ -181,3 +181,48 @@ def wrapper(request, *args, **kw):
...
+                exc.auth_header = api_auth.authenticate_header(request)
+                response = drf_exception_handler(exc, None)
```
diox, line 215:
```suggestion
                response = api_settings.EXCEPTION_HANDLER(exc, None)
```

That seems more correct ? That way you can also remove the `from rest_framework.views import exception_handler as drf_exception_handler` import, and move the other 2 `rest_framework` imports at the top of the file.

diox (approved):

Issue #19193, opened 2022-05-02
Migrate files to new directory structure


diox:
Once https://github.com/mozilla/addons-server/issues/11510 is implemented, make a script to migrate existing files to the new directory structure. This needs to work transparently/in a backwards-compatible way.

Logic is likely going to look like:
- Build a list of all known files excluding files that have been migrated already (new files etc)
- Go through the filesystem `files/` directory contents : for each directory:
  - For all files (and just files, not sub-directories) inside that directory:
    - If that file is not in the known list, ignore it
    - If that file is in the list:
      - Create directories for new path if necessary (makedirs)
      - Try linking from old to new path (skip if fail)
      - Once we know the file is linked, update file definition in database, marking the file as migrated

Once we're 100% done with that, all files in the old location (`files/<directory>/<file>` as opposed to `files/<directory>/<directory>/<directory>/<file>`) will be obsolete and we'll be able to remove them.

diox:
Note: we should use `generate_filename()` to generate the new name, if it has been properly updated in https://github.com/mozilla/addons-server/issues/11510 that will be helpful to fix any old inconsistencies, removing any artefacts like the `platform` in the name we used to have.

bqbn:
> * Try linking from old to new path (skip if fail)

What kind of linking is this going to be?

diox:
A hard link, for backwards-compatibility during the migration - the old path will eventually be removed, leaving no trace of the link (that's the `Once we're 100% done with that...` part at the end)

ioanarusiczki:
@diox Should I do a re-check of https://github.com/mozilla/addons-server/issues/19207#issuecomment-1123796442 ?

diox:
I'm currently running the command to migrate the files on dev (*); when it's over I'll do stage. Once it's done, you should pick any old version uploaded before https://github.com/mozilla/addons-server/issues/11510 landed on that env, and see if you can download it. Repeat for different states (deleted, disabled, approved...)

(*) It's likely going to take several hours. I'll then do stage. If it takes too long then we won't run it on prod this week.

diox:
@ioanarusiczki I ran the migration on dev and stage successfully, so you can try inspecting various versions and see if everything is working.

ioanarusiczki:
@diox I looked on both servers, what I've noticed:
- rev tools pages for themes - at click on the file -> it triggers an install , for the rest of the addons it's a download. 
[...]

PR #19266, opened 2022-05-19
Use addon id instead of slug in filenames if slug is empty

diox:
Follow-up for https://github.com/mozilla/addons-server/issues/19193

### DIFF
--- a/src/olympia/files/models.py
+++ b/src/olympia/files/models.py
@@ -70,7 +70,9 @@ def files_upload_to_callback(instance, filename):
     """
     # Start with the add-on slug, but make it go through django's slugify() to
     # drop unicode characters.
-    name = slugify(instance.addon.slug).replace('-', '_') or 'addon'
+    name = (
+        slugify(instance.addon.slug or instance.addon.pk).replace('-', '_') or 'addon'
+    )
     parts = (name, instance.version.version)
     file_extension = '.xpi' if instance.is_signed else '.zip'
     return os.path.join(
--- a/src/olympia/files/models.py
+++ b/src/olympia/files/models.py
@@ -51,8 +51,10 @@ def files_upload_to_callback(instance, filename):
     called saved yet.
 
     For new uploads, the returned path is:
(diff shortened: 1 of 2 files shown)


src/olympia/files/models.py
```diff
@@ -70,7 +70,9 @@ def files_upload_to_callback(instance, filename):
     """
     # Start with the add-on slug, but make it go through django's slugify() to
     # drop unicode characters.
-    name = slugify(instance.addon.slug).replace('-', '_') or 'addon'
+    name = (
+        slugify(instance.addon.slug or instance.addon.pk).replace('-', '_') or 'addon'
+    )
```
eviljeff, line 75: maybe we should always fall back to the addon id, rather than `addon`?
```suggestion
    name = (
         (instance.addon.slug and slugify(instance.addon.slug).replace('-', '_')) or str(instance.addon.pk)
    )
```
?

src/olympia/files/tests/test_models.py
```diff
@@ -216,6 +216,17 @@ def test_generate_filename_ja(self):
             == '42/4242/4242/addon-0.1.7.xpi'
         )
 
+    def test_generate_filename_no_slug(self):
+        file_ = File()
+        file_.version = Version(version='0.1.7')
+        file_.version.compatible_apps = {amo.FIREFOX: None}
+        file_.version.addon = Addon(slug='', pk=4242)
```
eviljeff, line 223: can we test when `slug=None` also?

eviljeff (approved):

src/olympia/files/models.py
diox: Yeah let's do that. d8607db353764e70a7dd0070505443630388ecad

src/olympia/files/tests/test_models.py
diox: https://github.com/mozilla/addons-server/pull/19266/commits/08c1b191997b6e6c75c31f3bdd3755187e6f05a3

PR #19275, opened 2022-05-23
Tell dependabot to ignore google-cloud-bigquery >= 3.0.0 for now

diox:
See https://github.com/mozilla/addons-server/pull/19272#issuecomment-1133171403


diox:
I think those are good - they should have been done before.

PR #19290, opened 2022-05-25
Allow default_locale to be set/changed via the API

### LINKED ISSUE #18235: allow default_locale to be set/updated via the addon api
split from #18036

We'd like to be able to set during create, and update later, the default locale for the addon's translated strings.  This will probably need a lot of validation because it's easy to end up with an add-on with (apparently) entirely missing metadata because those strings aren't present in the newly selected default locale.

eviljeff:
fixes #18235 - ended up quite a beast of a patch!

- it also introduces two subtle changes to all `TranslatedField`s when used with a new instance (creating): 
  - to disallow `{}` as a value for required fields (it was always kind of a bug that we considered an empty dict as a non-empty value);  
  - and to disallow `{locale: null}` (`allow_null`) by default (for new instances there is no legitimate reason to send `null` as a locale value - because it's used to clear _existing_ locale values - as it made validation extra complicated to have to consider that edge case too)


src/olympia/addons/serializers.py
```diff
@@ -231,8 +233,8 @@ def fake_object(self, data):
 
 class LicenseSerializer(serializers.ModelSerializer):
     is_custom = serializers.SerializerMethodField()
-    name = LicenseNameSerializerField()
-    text = TranslationSerializerField()
+    name = LicenseNameSerializerField(allow_null=True)
+    text = TranslationSerializerField(allow_null=True)
```
eviljeff, line 237: These were an edge case in how we use `LicenseSerializer` for custom licences - in effect the instance was seen as a `None` for an update too - so providing the `kwarg` just changes it back to the previous behavior.

src/olympia/addons/tests/test_views.py
```diff
@@ -1364,6 +1365,107 @@ def test_set_tags(self):
...
+            },
+        )
+        assert response.status_code == 400, response.data
+        assert response.data == {
+            'name': [error_string],
+            'support_url': [error_string],
+        }
+
+        # success cases:
+        # field is provided with a value in new default
```
bobsilverberg, line 1444: Same question here.

src/olympia/addons/tests/test_views.py
```diff
@@ -1975,6 +2077,59 @@ def test_metadata_required(self):
...
+        error_string = 'A value in the default locale of "fr" is required.'
+        assert response.data == {
+            'name': [error_string],
+            'summary': [error_string],
+            'description': [error_string],
+            'support_url': [error_string],
+        }
+
+        self.addon.update(description=None)
+        # success cases:
```
bobsilverberg, line 2116: Ok, I see these success cases are tested here, but why mention them in the comments above if they are not being tested in those test cases?

bobsilverberg (approved):
Looks good. I just have a couple of questions.

r+wc

src/olympia/addons/tests/test_views.py
```diff
@@ -1364,6 +1365,107 @@ def test_set_tags(self):
...
+            },
+        )
+        assert response.status_code == 400, response.data
+        error_string = 'A value in the default locale of "fr" is required.'
+        assert response.data == {
+            'description': [error_string],
+        }
+
+        # success cases:
+        # field is provided with a value in new default
```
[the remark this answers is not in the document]
eviljeff: they're all tested at once - `name` is given a value in post; `summary` is is defined in the manifest (and confirmed with `assert addon.summary == ...`);  `description` doesn't have a value (`assert addon.description is None`)

src/olympia/addons/tests/test_views.py
eviljeff: same here, the only difference is `summary` has a localized value in the _locales files in the xpi in `de` rather than just a string directly in the manifest.

src/olympia/addons/tests/test_views.py
eviljeff: similarly to the addon create tests, they are being all tested at once.  The summary translation is created earlier in the test on line 2085; description is set to `None` just above

eviljeff: I can add to the test comments?  What do you think would clarify what's going on?

bobsilverberg: In this test it seems like they are all called out, whereas in the others they are not, so maybe you can do the same in the others, but ultimately I don't think it's a big deal. Do whatever you like (or nothing at all).

PR #19300, opened 2022-05-30
Fail earlier in signing API if add-on is deleted

### LINKED ISSUE #19285: bug: "Internal Server Error" while upload not new version
### Is this a feature request or a bug?
Bug

### What is the current behavior?

When i test the "web-ext", i did successfully upload addon 2 times with version 4.0.0 and 4.0.1
Then i did removed addon from AMO.
Then i tried to upload an addon with same id (as removed addon) and i got unclear error on screenshot.
Then i tried to upload addon with no specify id, but "web-ext" wrote "Using previously auto-generated extension ID" and i got unclear error again.

It looks that server return unclear error when i try to upload not new version of removed addon, because when i tried to upload 4.0.2, process will starting successfully.

We should improve error code and handle this case correctly.

### What is the expected or desired behavior?

Return error with details of problem

diox:
Fixes #19285

### DIFF
--- a/src/olympia/signing/tests/test_views.py
+++ b/src/olympia/signing/tests/test_views.py
@@ -959,6 +959,20 @@ class TestUploadVersion(BaseUploadVersionTestMixin, TestCase):
         )
         assert response.status_code == 201
 
+    def test_deleted_webextension(self):
+        guid = '@webextension-with-guid'
+        addon = addon_factory(guid=guid, users=[self.user])
+        addon.delete()
+        response = self.request(
+            'POST',
+            guid=guid,
+            version='1.0',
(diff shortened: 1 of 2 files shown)


src/olympia/signing/tests/test_views.py
```diff
@@ -959,6 +959,20 @@ def test_addon_blocked_guid_in_xpi(self):
         )
         assert response.status_code == 201
 
+    def test_deleted_webextension(self):
```
diox, line 962: This test (creating a new add-on, not a new version) was already passing with the existing code FYI, but I figured we might as well make sure.

bobsilverberg (approved):

PR #19308, opened 2022-05-31
Add migration to insert missing DeniedGuids from blocked add-ons

### LINKED ISSUE #19289: Some fully blocked deleted add-ons are not in `DeniedGuid`s
It looks like some previously deleted blocked add-ons (with range `0-*`) have not been denied for resubmission (`DeniedGuid` entry). We should add those.

```sql
SELECT
    COUNT(DISTINCT a.id)
FROM
    addons a
    JOIN blocklist_block b ON b.guid = a.guid COLLATE utf8mb4_unicode_ci
    LEFT JOIN denied_guids d on d.guid = a.guid
WHERE
    min_version = "0" AND max_version = "*" AND a.status = 11 AND d.guid IS NULL;
```

diox:
Fixes #19289


src/olympia/addons/migrations/0039_auto_20220531_0953.py
```diff
@@ -0,0 +1,35 @@
...
+
+sql_query = """SELECT DISTINCT `a`.`guid`
+FROM
+    `addons` `a`
+    JOIN `blocklist_block` `b` ON `b`.`guid` = `a`.`guid` COLLATE utf8mb4_unicode_ci
+    LEFT JOIN `denied_guids` `d` on `d`.`guid` = `a`.`guid`
+WHERE
+    `min_version` = "0" AND `max_version` = "*" AND `a`.`status` = 11
+AND
+    `d`.`guid` IS NULL;"""
```
eviljeff, line 13:
```python
Addon.unfiltered.filter(
    guid__in=Block.objects.filter(min_version=Block.MIN, max_version=Block.MAX).values(
        guidc=Collate('guid', 'utf8mb4_unicode_ci')
    ),
    status=amo.STATUS_DELETED,
).exclude(guid__in=DeniedGuid.objects.values('guid'))
``` 
?

diox: I originally stayed with raw SQL because I couldn't do the arbitrary join with the ORM, but didn't think of using subqueries. Did that in https://github.com/mozilla/addons-server/pull/19308/commits/d598621dffaacf61eebb242f0ad1decd9d25c17a

eviljeff: I think you need a `values_list` on the end if you want just guids rather than `Addon` instances.

eviljeff (approved):

PR #19309, opened 2022-05-31
Remove sync_objects_to_basket and associated code

### LINKED ISSUE #19238: Remove sync_objects_to_basket and associated code
`sync_objects_to_basket` and associated code is behind a waffle switch, `basket-amo-sync`, that was never turned on in production. It was intended, as the name suggests, to sync objects (add-ons, users) data to basket which would then maintain them in Salesforce, but we moved away from Salesforce before this was enabled.

It's unlikely we'll ever pick up that work again as we've changed strategy with the new contact management system and instead of doing the queries there, we run a query against STMO and the AMO database and then export the ids into the marketing tool we use.

So let's remove all that code...

diox:
It was never enabled in production and we have no plans to ever resume that work.

Fixes #19238


bobsilverberg (approved):
Looks good. Nice cleanup!

PR #19342, opened 2022-06-08
Implement addon PUT as full-update or create

### LINKED ISSUE #15353: Add an API for releng to submit entirely new locales
The current signing API does not support submission of the first listed add-on version, due to required extra meta-data. To remove the obstacles and avoid breakage in the release process, we could add an API that support this specifically for langpack submissions from the releng account.

When a the first listed version gets submitted through the API, AMO will set the required metadata to some hard-coded defaults, for example setting the license etc.

/cc @escapewindow

eviljeff:
fixes #15353

This allows submission of all types of add-on, not just langpacks (as #15353 talks about), but should we limit PUT on `addon/<slug>/` to just langpacks as per issue?  Pros: The existing signing API allowed this style PUT to create or update so it'd be restoring that functionality. Cons: it's generally not a "good idea" to have an API where you don't know what the request will do - it _could_ create a new version for an existing add-on, or it _could_ create a new add-on - both outcomes could be a surprise if you weren't expecting it.  There's also some subtle differences around metadata in the manifest (the name, description, and homepage would be extracted for a new add-on; they wouldn't be for a new version) which are non-obvious.

Another feature this patch restores from the signing API (partially) is the ability to specify the GUID via the url if it's not in the manifest.  Pros: It's a convenient feature; Cons: it's against the direction of travel regarding future changes to require the manifest specify the guid explicitly so it's possibly unwise to encourage that use.  

Note: the issue talks about auto-generating the summary + license + category for a langpack - the patch doesn't do that as it's now trivial for the client to supply this in the POST data.

### DIFF
--- a/src/olympia/addons/views.py
+++ b/src/olympia/addons/views.py
@@ -383,6 +383,7 @@ class AddonViewSet(
             return super().update(request, *args, **kwargs)
 
         # otherwise we create a new add-on
+        self.action = 'create'
         serializer = self.get_serializer(data=request.data)
         serializer.is_valid(raise_exception=True)
         serializer.save(guid=self.kwargs['guid'])
--- a/src/olympia/addons/views.py
+++ b/src/olympia/addons/views.py
@@ -381,14 +381,10 @@ class AddonViewSet(
         if instance:
             # if the add-on exists we can use the standard update
             return super().update(request, *args, **kwargs)
-
-        # otherwise we create a new add-on
-        self.action = 'create'
-        serializer = self.get_serializer(data=request.data)
-        serializer.is_valid(raise_exception=True)
-        serializer.save(guid=self.kwargs['guid'])
-
-        return Response(serializer.data, status=status.HTTP_201_CREATED)
+        else:
+            # otherwise we create a new add-on
+            self.action = 'create'
+            return self.create(request, *args, **kwargs)
 
 
 class AddonChildMixin:
--- a/src/olympia/addons/views.py
+++ b/src/olympia/addons/views.py
@@ -80,7 +80,6 @@ from .indexers import AddonIndexer
 from .models import Addon, AddonUser, AddonUserPendingConfirmation, ReplacementAddon
 from .serializers import (
     AddonEulaPolicySerializer,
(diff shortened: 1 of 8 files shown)


src/olympia/addons/views.py
```diff
@@ -362,6 +363,32 @@ def destroy(self, request, *args, **kwargs):
...
+        if instance:
+            # if the add-on exists we can use the standard update
+            return super().update(request, *args, **kwargs)
+
+        # otherwise we create a new add-on
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save(guid=self.kwargs['guid'])
+
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
```
diox, line 390: Couldn't you use `self.create(request, *args, **kwargs)` here instead ?

diox: And, related to that... wouldn't `self.action` be incorrect here with your current code ?

eviljeff: It is possible to use `self.create` (and I almost did) but we'd need to override the `save` method to include `guid` (and either check the action in there so we only did it for PUT) so I didn't think it helped much.  I can though?

eviljeff: I'm not following what you mean about `self.action` - can you expand on it and/or include an example scenario?

diox: When you reach this code, `self.action` is still `update`, but we're doing a creation

diox: > It is possible to use `self.create` (and I almost did) but we'd need to override the `save` method to include `guid` (and either check the action in there so we only did it for PUT) so I didn't think it helped much. I can though?

No, you're right, it's not worth it.

diox (approved):

src/olympia/addons/views.py
eviljeff: https://github.com/mozilla/addons-server/pull/19342/commits/4445dd36723a9b9c65df34918f93c8464e8647bc

PR #19365, opened 2022-06-15
Optimize Version queries around file/permissions

### LINKED ISSUE #18047: Fix Version queries that don't scale when webext permissions are loaded
When writing the patch for #17946 I found the querysets in AddonVersionViewSet, AddonVersionViewSet, and reviewers.views.review end up with webext_permission queries for each Version instance returned.  Adding a `prefetch_related` didn't help (there were more queries instead).  I suspect the fix will have to made with, or inside, the version transformers.

diox:
- Add a `select_related()` on `file` in Version's default queryset instead of making another query in the transformer to fetch the files.
- Add a `select_related(None)` on querysets that really don't need the `file` for the versions they are fetching (most do)
- Add a `select_related()` on `file__webext_permissions` in `review()` and `AddonVersionViewSet` now that the files are no longer fetched separately
- Avoid loading permissions entirely for `LanguageToolView` - we don't need them there.
- Stop loading the entire base version in `review()` - we just need the pk

This significantly reduces the number of queries made by language tools API, addon versions API and the review page.

Fixes #18047

### DIFF
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -5626,7 +5626,7 @@ class TestReview(ReviewBase):
         # Delete ActivityLog to make the query count easier to follow. We have
         # other tests for the ActivityLog related stuff.
         ActivityLog.objects.for_addons(self.addon).delete()
-        with self.assertNumQueries(67):
+        with self.assertNumQueries(51):
             # See test_item_history_pagination() for more details about the
             # queries count. What's important here is that the extra versions
             # and scanner results don't cause extra queries.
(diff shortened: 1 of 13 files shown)


src/olympia/reviewers/tests/test_views.py
```diff
@@ -3082,7 +3082,7 @@ def test_item_history_pagination(self):
             str(author.get_role_display()),
             self.addon,
         )
-        with self.assertNumQueries(71):
+        with self.assertNumQueries(55):
```
eviljeff, line 3085: 🥳

eviljeff (approved):

PR #19397, opened 2022-06-23
Always convert images to RGBA before processing them

### LINKED ISSUE #19396: Image uploads in devhub stuck on "Image changes being processed"

I'm trying to upload a new screenshot for the add-on: https://addons.mozilla.org/en-US/firefox/addon/add-torrent-to/

Steps to reproduce: 
1. Open "Edit Product Page" of an add-on.
2. Press edit button in the Images section.
3. Add a new screenshot.
4. Press the "Save Changes" button.

The uploaded image is stuck in the "Image changes being processed" stage.

### What happened?

After I press "Save Changes" button at the Images section, the uploaded image is stuck in the "Image changes being processed" stage.
being processed

The uploaded images are in the archive below.
[images.zip](https://github.com/mozilla/addons/files/8965671/images.zip)

Images being loaded normally.

diox:
Otherwise we can end up trying to paste a RGB image onto a RGBA background, which fails.

Fixes #19396


eviljeff:
> Otherwise we can end up trying to paste an RGBA background on an RGB image, which fails.

We paste the image (which may be RBG or RGBA) onto a RGBA background, rather than the other way around
 https://github.com/mozilla/addons-server/blob/master/src/olympia/amo/utils.py#L668:L669

diox:
Fixed the description and rebased to avoid failures from old `master` branch.

PR #19432, opened 2022-07-05
Only consider extensions in the `mad`/`scanner` reviewer queues and scanner actions

### LINKED ISSUE #19430: Only consider extensions in the `mad` and `scanners` reviewer queues
The `mad` and `scanners` queues in reviewer tools currently consider all add-on types. Their queries would likely be twice as fast if they only had to consider extensions (because they need to look at all non-disabled/deleted add-ons - scanners can flag any version of any public/non public add-on, in theory). The drawback would be, obviously, that non extensions would no longer show up in these queues. IF that's not a problem, let's do it to improve performance of these queues.

diox:
This allows the query to look at fewer rows, making it faster - the scanner actions also avoid touching non-extensions to make things consistent and avoid the situation where something would trigger an action we wouldn't notice because it's not shown in the queues.

In addition, in the reviewer queues, the `COUNT()` is optimized using the same method already used for dashboard counts, by not returning columns we won't need and avoiding sorting in the subquery.

Fixes #19430

### DIFF
--- a/src/olympia/scanners/tests/test_actions.py
+++ b/src/olympia/scanners/tests/test_actions.py
@@ -377,6 +377,29 @@ class TestRunAction(TestCase):
         assert _delay_auto_approval_indefinitely_mock.called
         _delay_auto_approval_indefinitely_mock.assert_called_with(self.version)
 
+    @mock.patch('olympia.scanners.models._delay_auto_approval_indefinitely')
+    def test_returns_for_non_extension_addons(
+        self, _delay_auto_approval_indefinitely_mock
+    ):
+        self.scanner_rule.update(action=DELAY_AUTO_APPROVAL_INDEFINITELY)
+        self.version.addon.update(type=amo.ADDON_DICT)
+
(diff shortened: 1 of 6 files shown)


src/olympia/scanners/tests/test_actions.py
```diff
@@ -377,6 +377,29 @@ def test_runs_delay_auto_approval_indefinitely(
...
+
+        ScannerResult.run_action(self.version)
+
+        assert not _delay_auto_approval_indefinitely_mock.called
+
+        self.version.addon.update(type=amo.ADDON_STATICTHEME)
+
+        ScannerResult.run_action(self.version)
+
+        assert not _delay_auto_approval_indefinitely_mock.called
```
bobsilverberg, line 401: It might be unnecessary, but would it be worth updating the addon type one final time to be an extension and then asserting that the mock is in fact called?

bobsilverberg (approved):
Looks good. I have one question about a test.

r+wc

PR #19441, opened 2022-07-08
Run CI/local envs with ES 7 and update client libs

### LINKED ISSUE #17389: Upgrade Elasticsearch to 7.1.x

diox:
Also drive-by update CI to have the right versions of MySQL/Redis

Fixes #17389

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -284,9 +284,9 @@ references:
   docker-images:
     # These versions should try to match what we run in production.
     - &python "cimg/python:3.9"
-    - &redis "redis:2.8"
+    - &redis "redis:6.2"
     - &memcached "memcached:1.4"
-    - &mysql "mysql:8.0.21"
+    - &mysql "mysql:8.0.28"
 
   mysql-config: &mysql-config
     environment:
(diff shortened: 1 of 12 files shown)


.circleci/config.yml
```diff
@@ -284,9 +284,9 @@ references:
   docker-images:
     # These versions should try to match what we run in production.
     - &python "cimg/python:3.9"
-    - &redis "redis:2.8"
+    - &redis "redis:6.2"
```
eviljeff, line 287: that's quite an upgrade!

eviljeff (approved):

.circleci/config.yml
diox: Note that it was already done in production and in our local images (https://github.com/mozilla/addons-server/pull/19001) - this is just for consistency. Ultimately we only use redis as a celery result backend so the impact is fairly limited.

PR #19465, opened 2022-07-15
show version status in reject multiple versions list

### LINKED ISSUE #18121: Recognize the status of add-on versions when selecting for a multiple version rejection
Hello, 
as admin reviewers, when we want to reject multiple versions of an add-on we must select which are concerned.

In this list, unreviewed, delayed rejected versions appear but also the last confirmed version.
It appears that except in rare cases we must not reject approved versions.

Because of this, we must always verify if there are versions approved in the list. When there is a lot, it's a not negligible time consumed.
The admin review team would save a lot of time if we would be able to see the status next to the number in the version selecting list.

Maybe with a colour code? 
Green - Human approved (+ confirmed auto approval)
Orange - Delayed rejected
Yellow - Awaiting review
Grey - Auto approved (not reviewed yet)

Here a screenshot:
image

eviljeff:
fixes #18121

### DIFF
--- a/src/olympia/reviewers/forms.py
+++ b/src/olympia/reviewers/forms.py
@@ -19,6 +19,7 @@ from olympia.ratings.permissions import user_can_delete_rating
 from olympia.reviewers.models import CannedResponse, ReviewActionReason, Whiteboard
 from olympia.versions.models import Version
 
+import markupsafe
 
 log = olympia.core.logger.getLogger('z.reviewers.forms')
 
@@ -114,6 +115,28 @@ class VersionsChoiceWidget(forms.SelectMultiple):
         amo.STATUS_AWAITING_REVIEW: ('reject_multiple_versions',),
     }
 
+    def get_extended_status(self, obj):
+        if not obj or not obj.file:
+            return
+        # This is the default status
+        status = obj.file.get_status_display()
+        # But we override with more a specific status if available
+        if (
+            reviewer_flags := getattr(obj, 'reviewerflags', None)
+        ) and reviewer_flags.pending_rejection:
+            status = 'Delayed Rejection'
+        elif obj.file.status == amo.STATUS_APPROVED:
+            summary = getattr(obj, 'autoapprovalsummary', None)
+            if summary and summary.verdict == amo.AUTO_APPROVED:
+                status = (
+                    'AutoApproved, Confirmed'
+                    if summary.confirmed is True
+                    else 'AutoApproved, not Confirmed'
+                )
+            else:
+                status = 'Approved, Manual'
+        return status
+
     def create_option(self, *args, **kwargs):
         option = super().create_option(*args, **kwargs)
         # label_from_instance() on VersionsChoiceField returns the full object,
@@ -131,7 +154,9 @@ class VersionsChoiceWidget(forms.SelectMultiple):
             )
         # Just in case, let's now force the label to be a string (it would be
         # converted anyway, but it's probably safer that way).
-        option['label'] = str(obj)
+        option['label'] = str(obj) + markupsafe.Markup(
+            f' - {extended}' if (extended := self.get_extended_status(obj)) else ''
+        )
         return option
 
 
@@ -254,6 +279,8 @@ class ReviewForm(forms.Form):
                 .filter(channel=channel, file__status__in=statuses)
                 .no_transforms()
                 .select_related('file')
+                .select_related('autoapprovalsummary')
+                .select_related('reviewerflags')
                 .order_by('created')
             )
             # Reset data-value depending on widget depending on actions
--- a/src/olympia/reviewers/tests/test_forms.py
+++ b/src/olympia/reviewers/tests/test_forms.py
@@ -1,12 +1,15 @@
+from datetime import datetime
+
 from django.utils.encoding import force_str
 
+import pytest
 from pyquery import PyQuery as pq
 
 from olympia import amo
 from olympia.addons.models import Addon, AddonReviewerFlags
 from olympia.amo.tests import TestCase, addon_factory, version_factory
 from olympia.constants.reviewers import REVIEWER_DELAYED_REJECTION_PERIOD_DAYS_DEFAULT
-from olympia.reviewers.forms import ReviewForm
+from olympia.reviewers.forms import ReviewForm, VersionsChoiceWidget
 from olympia.reviewers.models import (
     AutoApprovalSummary,
     CannedResponse,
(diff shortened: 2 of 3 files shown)


eviljeff:
I didn't go with the colours like the reporter suggested - it's not particularly accessible and you have to have a key somewhere to decode colour -> status.  Suggestions welcome for changes to text though.

src/olympia/reviewers/forms.py
```diff
@@ -114,6 +115,28 @@ class VersionsChoiceWidget(forms.SelectMultiple):
...
+    def get_extended_status(self, obj):
+        if not obj or not obj.file:
+            return
+        # This is the default status
+        status = obj.file.get_status_display()
+        # But we override with more a specific status if available
+        if (
+            reviewer_flags := getattr(obj, 'reviewerflags', None)
+        ) and reviewer_flags.pending_rejection:
+            status = 'Delayed Rejection'
```
wagnerand, line 127: Does this mean the rejection has not been executed yet?

src/olympia/reviewers/forms.py
```diff
@@ -114,6 +115,28 @@ class VersionsChoiceWidget(forms.SelectMultiple):
         amo.STATUS_AWAITING_REVIEW: ('reject_multiple_versions',),
     }
 
+    def get_extended_status(self, obj):
+        if not obj or not obj.file:
+            return
+        # This is the default status
+        status = obj.file.get_status_display()
```
wagnerand, line 122: I assume this will show "Disabled by Mozilla" for rejected files? If I understand and remember correctly, we also (wrongfully) use that label when the developer disables a version. Can we distinguish the two here?

If so, let's use "Rejected" and "Disabled by Developer".
If not, we'll probably need go to with something like "Rejected or Disabled by Developer".

wagnerand (commented):
Thanks, I added a few questions, and will leave the actual code-review to Bob.

src/olympia/reviewers/forms.py
eviljeff: yes, `pending_rejection` should be a (truthy) date when it's been set but not yet rejected.

wagnerand: Maybe renaming it to the following might make it more clear:

_Delay-rejected, scheduled for <DATE\>_

src/olympia/reviewers/forms.py
```diff
@@ -114,6 +115,28 @@ class VersionsChoiceWidget(forms.SelectMultiple):
...
+        # But we override with more a specific status if available
+        if (
+            reviewer_flags := getattr(obj, 'reviewerflags', None)
+        ) and reviewer_flags.pending_rejection:
+            status = 'Delayed Rejection'
+        elif obj.file.status == amo.STATUS_APPROVED:
+            summary = getattr(obj, 'autoapprovalsummary', None)
+            if summary and summary.verdict == amo.AUTO_APPROVED:
+                status = (
+                    'AutoApproved, Confirmed'
```
wagnerand, line 132:
```suggestion
                    'Auto-approved, Confirmed'
```

src/olympia/reviewers/forms.py
```diff
@@ -114,6 +115,28 @@ class VersionsChoiceWidget(forms.SelectMultiple):
...
+                    if summary.confirmed is True
+                    else 'AutoApproved, not Confirmed'
```
wagnerand, line 134:
```suggestion
                    else 'Auto-approved, not Confirmed'
```

src/olympia/reviewers/forms.py
eviljeff: sure, we can... but as I understood it the purpose of the issue was to notate the version ids to know which should/could be rejected.  Does knowing the scheduled date add anything to that?

src/olympia/reviewers/forms.py
eviljeff: afaik we can reliably know if a version has been disabled by the developer (`version.is_user_disabled`) but differentiating normal rejection from being superseded by a later upload (or manually changed in the admin) is tricky.  We attempt it in elsewhere on the review page - if there's a review date we say "Rejected", otherwise "Rejected or Unreviewed" - so we could do the same here.

eviljeff: tbh it's annoying this extra status information logic is buried in obscure reviewer tools jina helper functions!

wagnerand: I see. It might be an easier approach then to say "Disabled by Developer" if that happened and "Disabled by Mozilla" otherwise. Thoughts?

eviljeff: I was going to reuse the same code we use for display file status on the review page for consistency (but refactor it out of jinja_helpers)

wagnerand: Ah hm. My concern is that the file status display on the review page might not be accurate. It's possible it might be missing `is_user_disabled`. It sounded logical to me not to repeat (or reuse) this bug, but we can also fix in a followup if you prefer.

src/olympia/reviewers/forms.py
eviljeff: https://github.com/mozilla/addons-server/pull/19465/commits/a9e9d769fac6b178db0b8ecaf6e594307d4e4ac7

src/olympia/reviewers/forms.py
eviljeff: (this will be in another commit - WIP)

eviljeff: https://github.com/mozilla/addons-server/pull/19465/commits/c4e2ceb03bf09abdf5fa8fafa9f6847503c23b47 adds these status functions, and moves the logic out of forms.py too so it's all in one place, in the model.

eviljeff:
@bobsilverberg this is good to review now

src/olympia/files/models.py
```diff
@@ -291,6 +292,19 @@ def optional_permissions(self):
         except WebextPermission.DoesNotExist:
             return []
 
+    def get_review_status_display(self):
+        # Like .get_file_status_display but with logic to make the status more accurate
+        if self.status == amo.STATUS_DISABLED:
+            return (
+                gettext('Rejected')
+                if self.reviewed is not None
```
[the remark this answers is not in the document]
eviljeff: yeah, that's why the else is "Rejected or Unreviewed" (and the comment).  I'd kinda hoped that had improved since but the text covers our bases.

wagnerand (approved):
Overall logic looks good to me, thanks! Will leave in-depth code review to Bob.

PR #19467, opened 2022-07-18
redirect all (non-themes) back to review page after submission

### LINKED ISSUE #19437: After submitting a listed review, load/redirect to the review page
When submitting a listed review, we should load/redirect to the (same) review page. 

### What happened?
We currently load a queue page, I believe it's the "Auto-approved add-ons" page, which is slow to load and probably not used much at the moment.

Load the review page we submitted the review for.

We are already doing that for unlisted review submissions.

eviljeff:
fixes #19437

### DIFF
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -59,7 +59,6 @@ from olympia.api.permissions import (
     AnyOf,
     GroupPermission,
 )
-from olympia.constants.promoted import RECOMMENDED
 from olympia.constants.reviewers import REVIEWS_PER_PAGE, REVIEWS_PER_PAGE_MAX
 from olympia.devhub import tasks as devhub_tasks
 from olympia.files.models import File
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/views.py
```diff
@@ -650,26 +643,23 @@ def review(request, addon, channel=None):
...
+    approvals_info = None
+    if (
+        channel == amo.RELEASE_CHANNEL_LISTED
+        and addon.current_version
+        and addon.current_version.was_auto_approved
+    ):
+        try:
+            approvals_info = addon.addonapprovalscounter
+        except AddonApprovalsCounter.DoesNotExist:
+            pass
```
diox, line 655: nit: Not a new thing, but we can move this block way down, to avoid doing this when we're redirecting - it seems to be only useful when displaying the template.

diox (approved):

src/olympia/reviewers/views.py
eviljeff: https://github.com/mozilla/addons-server/pull/19467/commits/534a011218f67ea9af1a6a25b6ebce47b540612f

PR #19469, opened 2022-07-18
Don't inherit nomination time from delay-rejected versions

### LINKED ISSUE #19439: Don't inherit nomination time from delay-rejected versions

In the reviewer tools, the nomination time is inherited from previous versions if they have not been reviewed. It seems we never updated the logic when we introduced delayed rejections?

### What happened?
When a version is delay-rejected, the nomination time, should be reset.

As far as I can see, when a version is delay-rejected, the nomination time, subsequent submissions inherit the nomination time from the delay-reject version.

diox:
Fixes #19439

### DIFF
--- a/src/olympia/versions/models.py
+++ b/src/olympia/versions/models.py
@@ -792,16 +792,18 @@ class Version(OnChangeMixin, ModelBase):
             self.update(nomination=nomination, _signal=False)
 
     def inherit_nomination(self, from_statuses=None):
-        last_ver = (
+        qs = (
             Version.objects.filter(addon=self.addon, channel=amo.RELEASE_CHANNEL_LISTED)
             .exclude(nomination=None)
             .exclude(id=self.pk)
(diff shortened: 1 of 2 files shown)


src/olympia/versions/models.py
```diff
@@ -792,16 +792,18 @@ def reset_nomination_time(self, nomination=None):
...
+            .values_list('nomination', flat=True)
             .order_by('-nomination')
         )
         if from_statuses:
-            last_ver = last_ver.filter(file__status__in=from_statuses)
-        if last_ver.exists():
-            self.reset_nomination_time(nomination=last_ver[0].nomination)
+            qs = qs.filter(file__status__in=from_statuses)
+        if qs.exists():
+            self.reset_nomination_time(nomination=qs[0])
```
eviljeff, line 806:
```suggestion
        if nomination := qs.first():
            self.reset_nomination_time(nomination=nomination)
```

eviljeff (approved):

PR #19481, opened 2022-07-20
Refactor search code into a simpler set of modules and disable ES verification

### LINKED ISSUE #18966: Refactor search code in a simpler set of modules
Search code in addons-server is spread across multiple files:
- `amo/search.py` contains the code to create an `Elasticsearch` instance from our base settings
- `addons/indexers.py` contains the `AddonIndexer` class (*)
- `lib/es/utils.py` contains the actual indexing function
- `lib/es/models.py`, `lib/es/management/commands/reindex.py` contains the reindex command and associated model
- `search/filters.py` contains logic to make queries for the search API (kind of a bridge between DRF and `elasticsearch_dsl`).

(*) We used to have more searchable models - stats, users, collections... but that's all gone now, only add-ons are left.

It's time to clean it all up and do a big refactor. Maybe all of this should be moved to the  `olympia.search` app ?

### LINKED ISSUE #19463: Number of requests for the ES cluster doubled right after 2022.07.14 push
We see that the number of requests for the ES cluster doubled almost immediately after the 2022.07.14 push. 

Screen Shot 2022-07-14 at 9 49 27 AM

### What happened?
The ES cluster was upgraded to v7 on 7/11/2022. Right after today's push (tag 2022.07.14), we see the number of requests for the ES cluster doubled. And meanwhile, the number of requests for the website hadn't increased, so this wasn't caused by increased amount of traffic to the website.

The number of requests for the ES cluster does not increase when the traffic to the website hasn't increased.

n/a

diox:
- `olympia.search` should contain everything to interact with ES
- `olympia.addons.indexers` continue to contain the mapping for `Addon`

Fixes #18966
Fixes #19463


bobsilverberg (approved):
Looks good. Clever way at getting around that ridiculous code they added. :-)

PR #19490, opened 2022-07-21
after delayed rejection, correctly reject as content_review or not

### LINKED ISSUE #19336: There is no difference between content rejection and code rejection
When we content delay reject after two weeks, when it gets auto rejected it's counted as a code rejection, not a content one.

### What happened?
When a content delayed rejected add-on gets auto rejected, after 2 weeks, in database the action is set as rejected instead of reject content.

A content delayed add-on should get auto rejected with the action for reject content, not rejected.

eviljeff:
fixes #19336


src/olympia/reviewers/tests/test_utils.py
```diff
@@ -2318,19 +2329,18 @@ def test_reject_multiple_versions_delayed_uses_original_user(self):
         self.addon.reload()
         assert self.addon.status == amo.STATUS_NULL
 
-        # The versions are not pending rejection.
-        for version in self.addon.versions.all():
-            assert version.pending_rejection is None
-            assert version.pending_rejection_by is None
```
eviljeff, line 2324: These lines didn't work if the aim of the test was to emulate how it would have been called via the `auto_reject` command - the pending rejection flags are cleared in the `auto_reject` command itself rather than here, and when `human_review` is `False` (which it would be via `auto_reject`) then they wouldn't be cleared either.

Unless I'm misinterpreting what behaviour the test was originally validating.

diox: I think this is just to prove the rejection wasn't delayed - multiple tests have that, here it's just as a safeguard. The test is way too long though so it's hard to follow regardless...

src/olympia/versions/migrations/0026_versionreviewerflags_pending_content_rejection.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('versions', '0025_auto_20220614_1653'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='versionreviewerflags',
+            name='pending_content_rejection',
+            field=models.BooleanField(default=False),
```
diox, line 16: Needs to be a `NullBooleanField` for backwards-compatibility since Django won't generate a boolean `DEFAULT` in database (or rather it will but then drop it, so it's not 100% backwards-compatible)

If you want to avoid multiple migrations afterwards, don't modify the model, only change the migration to make it a `NullBooleanField`, keeping `default=False`.

src/olympia/versions/models.py
```diff
@@ -985,6 +985,7 @@ class VersionReviewerFlags(ModelBase):
     pending_rejection_by = models.ForeignKey(
         UserProfile, null=True, on_delete=models.CASCADE
     )
+    pending_content_rejection = models.BooleanField(default=False, db_index=False)
```
diox, line 988: `db_index=False` is redundant

diox (changes requested):
For consistency you should clear the flag when the version is approved/rejected like we do for the other delayed rejection bits. This is done in `auto_reject.py` + for various actions in the `ReviewBase` children classes.

IMHO this is also a good reason to make it a `NullBooleanField(default=None)`, and not just for backwards-compatibility: `None` would indicate that it hasn't been set at all (no delayed rejection), and `True`/`False` would correspond to a content/regular delayed rejection

src/olympia/versions/migrations/0026_versionreviewerflags_pending_content_rejection.py
eviljeff: Not following here - we want the django behaviour don't we?  It'd set all the existing records to `False` by defining the default.  Are we just concerned about the incompatibility during the push window?  (if so, could just ask reviewers not to submit during that 30 minute window?)

eviljeff:
> For consistency you should clear the flag when the version is approved/rejected like we do for the other delayed rejection bits. This is done in `auto_reject.py` + for various actions in the `ReviewBase` children classes.

To clarify - there isn't any change in behaviour here.  We currently only clear the flag when `self.human_review` is True in the review helper, and currently clear the flag in `auto_reject`.  None of this is changing.  What are you requesting be changed exactly?

eviljeff:
> IMHO this is also a good reason to make it a `NullBooleanField(default=None)`, and not just for backwards-compatibility: `None` would indicate that it hasn't been set at all (no delayed rejection), and `True`/`False` would correspond to a content/regular delayed rejection

`pending_rejection` having a null value indicates delayed rejection hasn't been set.   I figured the only reason that `reject_multiple_versions` would ever be called with `human_review == False` would be via the `auto_reject` command, where only versions with a `pending_rejection` value would be filtered into the qs.  But we could explicitly check that too in the helper I suppose?

src/olympia/versions/migrations/0026_versionreviewerflags_pending_content_rejection.py
diox: I'm worried about the incompatibility during the push window and lack of backwards-compatibility if we need to go back to previous tag/cancel the push midway.

diox:
> To clarify - there isn't any change in behaviour here. We currently only clear the flag when `self.human_review` is True in the review helper, and currently clear the flag in `auto_reject`. None of this is changing. What are you requesting be changed exactly?

Currently you are not clearing the `pending_content_rejection` like we do for `pending_rejection` or `pending_rejection_by` in [`auto_reject.py`](https://github.com/mozilla/addons-server/blob/75d779b4fc9f34f9215360318b778d5f9d29ab83/src/olympia/reviewers/management/commands/auto_reject.py#L75-L76), [`approve_latest_version` action](https://github.com/mozilla/addons-server/blob/75d779b4fc9f34f9215360318b778d5f9d29ab83/src/olympia/reviewers/utils.py#L1028-L1031) and [`confirm_auto_approved` action](https://github.com/mozilla/addons-server/blob/75d779b4fc9f34f9215360318b778d5f9d29ab83/src/olympia/reviewers/utils.py#L1191-L1197)

diox:
> > IMHO this is also a good reason to make it a `NullBooleanField(default=None)`, and not just for backwards-compatibility: `None` would indicate that it hasn't been set at all (no delayed rejection), and `True`/`False` would correspond to a content/regular delayed rejection
> 
> `pending_rejection` having a null value indicates delayed rejection hasn't been set. I figured the only reason that `reject_multiple_versions` would ever be called with `human_review == False` would be via the `auto_reject` command, where only versions with a `pending_rejection` value would be filtered into the qs. But we could explicitly check that too in the helper I suppose?

I'm worried about the inconsistency if we add that flag more in the future in different actions etc. Having 3 states and properly clearing it would make things more explicit and avoid potential inconsistencies - same reason why I like having `pending_rejection_by` cleared at the same time as `pending_rejection`.

eviljeff:
> > To clarify - there isn't any change in behaviour here. We currently only clear the flag when `self.human_review` is True in the review helper, and currently clear the flag in `auto_reject`. None of this is changing. What are you requesting be changed exactly?
> 
> Currently you are not clearing the `pending_content_rejection` like we do for `pending_rejection` or `pending_rejection_by` in [`auto_reject.py`](https://github.com/mozilla/addons-server/blob/75d779b4fc9f34f9215360318b778d5f9d29ab83/src/olympia/reviewers/management/commands/auto_reject.py#L75-L76), [`approve_latest_version` action](https://github.com/mozilla/addons-server/blob/75d779b4fc9f34f9215360318b778d5f9d29ab83/src/olympia/reviewers/utils.py#L1028-L1031) and [`confirm_auto_approved` action](https://github.com/mozilla/addons-server/blob/75d779b4fc9f34f9215360318b778d5f9d29ab83/src/olympia/reviewers/utils.py#L1191-L1197)

right, yeah.  It's just the code/content state of the last time `pending_rejection` had a value.  

If it was changed to be a boolean+null field then we could clear it, but it's not particularly obvious what we'd do if the value was null yet `pending_rejection` had a date defined.  We can treat null as `False` (i.e. default to code review, like now)... though in that case we may as well just "clear" it by setting the value to `False`?

eviljeff:
> I'm worried about the inconsistency if we add that flag more in the future in different actions etc. Having 3 states and properly clearing it would make things more explicit and avoid potential inconsistencies - same reason why I like having `pending_rejection_by` cleared at the same time as `pending_rejection`.

I'd argue having 3 states would mean the potential for _more_ inconsistency 😅 - when `pending_content_rejection` is `null` and `pending_rejection` has a value - but if you want, sure.

bobsilverberg:
@eviljeff Does this still need my review, or is @diox taking care of it?

diox:
I'll take the whole review.

diox:
> I'd argue having 3 states would mean the potential for _more_ inconsistency sweat_smile - when `pending_content_rejection` is `null` and `pending_rejection` has a value - but if you want, sure.

This kind of thing would be solved by a database constraint - not letting `pending_content_rejection` be `null` if `pending_rejection` is set to anything. We have a `CheckConstraint` doing this for `pending_rejection_by` already.

src/olympia/versions/migrations/0026_versionreviewerflags_pending_content_rejection.py
eviljeff: https://github.com/mozilla/addons-server/pull/19490/commits/0b4e06f7609601d195cd90f1653016d9ad94096b

diox (approved):

src/olympia/versions/models.py
```diff
@@ -985,22 +987,23 @@ class VersionReviewerFlags(ModelBase):
...
-                name='pending_rejection_both_none',
+                name='pending_rejection_all_none',
                 check=(
                     models.Q(
                         pending_rejection__isnull=True,
                         pending_rejection_by__isnull=True,
+                        pending_content_rejection__isnull=True,
                     )
                     | models.Q(
                         pending_rejection__isnull=False,
```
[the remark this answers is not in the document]
eviljeff: https://github.com/mozilla/addons-server/issues/19498

PR #19504, opened 2022-07-26
Remove addon field from Abuse Reports

### LINKED ISSUE #17629: The addon detail card is no longer displayed in admin Abuse reports
Follow up for #17507

1. Go to the Abuse Reports model in admin tools
2. Select an abuse report for a public add-on from the list - see https://addons-internal.dev.mozaws.net/en-US/admin/models/abuse/abusereport/25580/change/
3.  Notice the layout and contents displayed in the first section of the page

### What happened?
The section is called "Unknown Add-on" and is missing some details

The section should contain and add-on info card, similar to what we have on the add-on review page, and list all the Abuse reports submitted for the same add-on. This is an example of an older abuse report and how the page should look like: https://addons-internal.dev.mozaws.net/en-US/admin/models/abuse/abusereport/25556/change/

[current layout]
image

[before]
image

### LINKED ISSUE #17630: The Abuse Reports table on the add-on review page no longer displays the add-on name
Follow up for #17507

1. Open the review page for an add-on which has some abuse reports submitted - see https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/re-re-reused-guid-abuse
2. Examine the first column in the Abuse Reports table

### What happened?
The column has an entry for the add-on version only 

The column should also display the add-on name

[current layout]
image

[before]
image

### LINKED ISSUE #18879: Make sure AbuseReport.addon is unused and drop it
We used to link abuse reports to the add-on instance via `AbuseReport.addon`. For a while now, we have stopped doing that at least for abuse reports submitted from Firefox. Those are linked via `AbuseReport.guid`. We should make sure nothing else uses `addon` or migrate them to using `guid` (abuse reports from AMO possibly), then drop that field.

diox:
Fixes #18879
Fixes #17630
Fixes #17629

### DIFF
--- a/src/olympia/abuse/tests/test_admin.py
+++ b/src/olympia/abuse/tests/test_admin.py
@@ -32,39 +32,40 @@ class TestAbuse(TestCase):
         cls.addon1.name.__class__.objects.create(
             id=cls.addon1.name_id, locale='fr', localized_string='Elu'
         )
-        cls.addon2 = addon_factory(guid='@guid2', name='Two')
-        cls.addon3 = addon_factory(guid='@guid3', name='Three')
+        cls.addon2 = addon_factory(guid='@guid2', name='Owt')
+        cls.addon3 = addon_factory(guid='@guid3', name='Eerth')
         cls.user = user_factory(email='someone@mozilla.com')
         grant_permission(cls.user, 'AbuseReports:Edit', 'Abuse Report Triage')
         # Create a few abuse reports.
         cls.report1 = AbuseReport.objects.create(
-            addon=cls.addon1,
-            guid='@guid1',
+            addon_name='The One',
+            guid=cls.addon1.guid,
             message='Foo',
             state=AbuseReport.STATES.VALID,
             created=days_ago(98),
         )
         AbuseReport.objects.create(
-            addon=cls.addon2,
-            guid='@guid2',
+            addon_name='The Two',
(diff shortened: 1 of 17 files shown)


src/olympia/abuse/tests/test_admin.py
```diff
@@ -32,39 +32,40 @@ def setUpTestData(cls):
         cls.addon1.name.__class__.objects.create(
             id=cls.addon1.name_id, locale='fr', localized_string='Elu'
         )
-        cls.addon2 = addon_factory(guid='@guid2', name='Two')
-        cls.addon3 = addon_factory(guid='@guid3', name='Three')
+        cls.addon2 = addon_factory(guid='@guid2', name='Owt')
+        cls.addon3 = addon_factory(guid='@guid3', name='Eerth')
```
bobsilverberg, line 36: Did you mean this to be `Eerht`? 😉

bobsilverberg (approved):

src/olympia/abuse/tests/test_admin.py
diox: Any string would do, as it's only really useful for debugging, but it should totally have been `Eerht` !

diox: https://github.com/mozilla/addons-server/pull/19504/commits/a44b303c5d7dd9b5318572f7b2d4f5e2e9787113

PR #19506, opened 2022-07-26
Stop indexing non public addons in ES

### LINKED ISSUE #19445: Don't include non-public add-ons in the ES index
We currently index all add-ons in our ES index, regardless of whether there are any approved listed versions, if the addon is enabled or disabled, or its deleted status; then filter the non-public add-ons out before returning the results.  This was done for `<`historical reasons`>`.  Although it makes the indexing code a little simpler - index everything without working out if we should - it's inflating the ES index size for no good reason so lets stop doing that.

eviljeff:
fixes #19445

### DIFF
--- a/src/olympia/addons/tasks.py
+++ b/src/olympia/addons/tasks.py
@@ -79,14 +79,24 @@ def delete_preview_files(id, **kw):
 @use_primary_db
 def index_addons(ids, **kw):
     log.info(f'Indexing addons {ids[0]}-{ids[-1]}. [{len(ids)}]')
-    transforms = (attach_tags, attach_translations_dict)
-    index_objects(
-        ids=ids,
-        indexer_class=AddonIndexer,
-        index=kw.pop('index', None),
-        transforms=transforms,
-        manager_name='unfiltered',
+
+    queryset = (
+        Addon.objects.public()
+        .filter(id__in=ids)
+        .transform(attach_tags)
+        .transform(attach_translations_dict)
     )
+    ids_in_qs = list(queryset.values_list('pk', flat=True))
+    excluded_ids = [pk for pk in ids if pk not in ids_in_qs]
+
+    if queryset:
+        index_objects(
+            queryset=queryset,
+            indexer_class=AddonIndexer,
+            index=kw.pop('index', None),
+        )
+    if excluded_ids:
+        unindex_addons(excluded_ids)
 
 
 @task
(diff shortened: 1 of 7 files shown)


src/olympia/addons/tests/test_serializers.py
```diff
@@ -1054,6 +1062,22 @@ def test_grouped_ratings(self):
...
+        # deleted add-ons aren't in the ES index, so won't ever be serialized
+        pass
+
+    def test_no_current_version(self):
+        # add-ons with no current version aren't in the ES index, so won't be serialized
+        pass
+
+    def test_static_theme_preview_no_current_version(self):
+        # add-ons with no current version aren't in the ES index, so won't be serialized
+        pass
```
eviljeff, line 1079: I had to disable these tests because the changes meant that the ES search returned 0 results so there was nothing to serialize.

src/olympia/addons/tasks.py
```diff
@@ -96,11 +106,7 @@ def unindex_addons(ids, **kw):
...
         try:
-            es.delete(
-                AddonIndexer.get_index_alias(),
-                AddonIndexer.get_doctype_name(),
-                addon_id,
-            )
+            es.delete(index=AddonIndexer.get_index_alias(), id=addon_id)
         except TransportError:
             # We ignore already deleted add-ons.
             pass
```
diox, line 112: I can't find tests covering this bit (*), and it seems important - not only we want to guarantee that `unindex_addons()` works and that the document(s) are gone from the ES index, but also we want to make sure it works for already deleted/disabled add-ons that are modified.

eviljeff: https://github.com/mozilla/addons-server/pull/19506/commits/dadfd88cd86b95a19d7708387a7f7fec21501281 adds a test to check `unindex_objects` works and doesn't fail.  It's a bit basic - it just counts records - but it works and doesn't raise.

src/olympia/addons/tasks.py
```diff
@@ -79,14 +79,24 @@ def delete_preview_files(id, **kw):
 @use_primary_db
 def index_addons(ids, **kw):
     log.info(f'Indexing addons {ids[0]}-{ids[-1]}. [{len(ids)}]')
```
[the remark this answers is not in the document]
eviljeff: I refactored unindex_addons into a utility function, unindex_objects, to have the codepath be similar for both adding/updating and removing addons.
https://github.com/mozilla/addons-server/pull/19506/commits/dadfd88cd86b95a19d7708387a7f7fec21501281

eviljeff:
The complete CI doesn't seem to have shown up for the latest commit: https://app.circleci.com/pipelines/github/eviljeff/olympia/883/workflows/40d5bc8f-8734-4fc5-8c7e-da3ae5191661

diox (approved):

PR #19513, opened 2022-07-27
Run watchmedo through supervisor in local envs instead of through uwsgi

### LINKED ISSUE #19512: uwsgi autoreload sometimes spawns too many `watchmedo` processes in local envs
Our local envs are configured to automatically reload `uwsgi` when a file changed, through `watchmedo`. The way it's configured though, it relies on `uwsgi` to start and kill `watchmedo`, which is wasteful since `watchmedo` doesn't have to be killed, and sometimes spiral out of control as we're triggering new watchers while the others aren't dead yet, taking a lot of CPU, sometimes hitting system max watchers threshold.

Instead, we should have a single long-lived `watchmedo` process at, ran through `supervisor`, that triggers the reload without being killed.

diox:
This should prevent a storm of watchmedo processes when there are changes, as this should keep the same long-lived watchmedo process instead of spawning new ones.

Also, watch a specific file instead of the whole directory to avoid reloads when just opening a shell (since bash_history is written to the artifacts dir)

Fixes #19512

### DIFF
--- a/docker/supervisor-celery.conf
+++ b/docker/supervisor-celery.conf
@@ -3,10 +3,9 @@ logfile=/code/logs/supervisord-celery.log
 
 [program:olympia-worker]
 # Local env setup for celery. This is similar to prod, but with only two workers
-# and auto-restart through watchmedo (which may not work in 100% of the cases)
 # Prod config for reference:
 # https://github.com/mozilla-services/cloudops-deployment/blob/master/projects/amo/puppet/modules/olympia/manifests/celery.pp
-command=watchmedo auto-restart -d src/ -p '*.py' -- celery -A olympia.amo.celery:app worker -E -c 2 --loglevel=INFO
+command=watchmedo auto-restart --directory /code/src --pattern '*.py' --recursive -- celery -A olympia.amo.celery:app worker -E -c 2 --loglevel=INFO
 directory=/code
 stopasgroup=true
 autostart=true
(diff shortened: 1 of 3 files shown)


diox:
A few things to look into:
- That `touch` doesn't seem to work, even after changing it to `/code/docker/artifacts/uwsgi-reload-monitor`. I can't seem to get the trigger to work properly from `supervisord`, but it works fine from a shell though...
- `uwsgi` wants the file to exist at startup. Should we add it to the repos ?
- It's difficult to test, I think because `uwsgi` is running in lazy mode, so all workers might not have the app loaded at any time, so you might get accidentally the updated code without needing autoreload sometimes.
- `watchmedo` triggers twice for every modification, so ideally we'd have some sort of debouncer. There are `--drop` or `--wait` option but neither does the right thing.

eviljeff (approved):
They seems like reasonable changes, but file watching generally doesn't work for me (because the file system I'm using is NTFS and it's running via Docker)

eviljeff (changes requested):
Looks like this has broken `make reload`
```
uwsgi --reload /code/docker/artifacts/addons-server-uwsgi-master.pid
signal_pidfile()/kill(): No such process [core/uwsgi.c line 1695]
make: *** [Makefile-docker:198: reload] Error 255
```

diox:
Ah, I probably want to keep the `safe-pidfile` then.

eviljeff (approved):
`make reload` works now.

PR #19519, opened 2022-08-01
rm user_media_url and user_media_path jinja helpers

### LINKED ISSUE #18999: Remove `user_media_url` and `user_media_path`
Those 2 util functions are a little bit too magical. They will use the argument to look up a setting (`FOO_URL` | `FOO_PATH`) or fallback to a completely different logic (`MEDIA_URL/foo` | `MEDIA_PATH/foo`).

Ultimately they don't save you from having to concatenate the path anyway, since it's only really a prefix, so I think they are doing more harm than good. Let's remove them.

eviljeff:
fixes #18999 - also needed a reworking of `SafeStorage` because it used `user_media_path` (which affected a lot more things than the jinja helpers directly)

### DIFF
--- a/src/olympia/amo/utils.py
+++ b/src/olympia/amo/utils.py
@@ -861,21 +861,37 @@ class SafeStorage(FileSystemStorage):
 
     DEFAULT_CHUNK_SIZE = 64 * 2**10  # 64kB
 
-    def __init__(self, *args, **kwargs):
-        """If `user_media` arg is provided, location will be lazily resolved under the
-        appropriate user_media_path. If `user_media` is falsey MEDIA_ROOT is used - see
-        user_media_path for the resolution logic."""
-        if 'user_media' in kwargs:
-            self._user_media = kwargs.pop('user_media')
+    root_setting = 'STORAGE_ROOT'
+
+    def __init__(self, *args, rel_location=None, **kwargs):
+        """Default storage root can be changed with `root_setting` kwarg.
+
+        If `location` kwarg is provided then that will be used instead of default root.
+        If `rel_location` kwarg is provided then a relative path under `root_setting` is
+        used. It is an error to provide both `location` and `rel_location`.
+        """
+        if 'root_setting' in kwargs:
+            self.root_setting = kwargs.pop('root_setting')
+        self.rel_location = rel_location
+        assert not (
+            rel_location is not None and kwargs.get('location') is not None
+        ), "Don't provide both location and rel_location"
         super().__init__(*args, **kwargs)
+
+    def _clear_cached_properties(self, setting, **kwargs):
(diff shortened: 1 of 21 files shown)


src/olympia/amo/utils.py
```diff
@@ -861,21 +861,37 @@ class SafeStorage(FileSystemStorage):
 
     DEFAULT_CHUNK_SIZE = 64 * 2**10  # 64kB
 
-    def __init__(self, *args, **kwargs):
-        """If `user_media` arg is provided, location will be lazily resolved under the
-        appropriate user_media_path. If `user_media` is falsey MEDIA_ROOT is used - see
-        user_media_path for the resolution logic."""
-        if 'user_media' in kwargs:
-            self._user_media = kwargs.pop('user_media')
+    root_setting = 'STORAGE_ROOT'
```
eviljeff, line 864: btw, the reason we need to pass in the setting name and lazily resolve - i.e. can't just pass in a `location` kwarg with `path(settings.MEDIA_PATH, xxx)` or whatever, in it - is that it gets resolved too early for the testing setup where the paths are overridden to per-test temporary directories.

src/olympia/users/tasks.py
```diff
@@ -15,12 +14,8 @@
 def delete_photo(dst, **kw):
     task_log.info('[1@None] Deleting photo: %s.' % dst)
 
-    if not dst.startswith(user_media_path('userpics')):
```
bobsilverberg, line 18: Why is it safe to remove this? Will a `dst` that doesn't match `MEDIA_ROOT` + `userpics` simply not be deleted now?

bobsilverberg (approved):
This looks good, but I am wondering; it looks like a large number of the code changes result in prepending either `settings.MEDIA_ROOT` or `settings.MEDIA_URL` to a string (which is sometimes accomplished via a template string and other times via `os.path.join`).

Would it be worthwhile still having a helper function to do that, so we don't have to continually specify `settings.MEDIA_ROOT` or `settings.MEDIA_URL` in all of these files? I realize that might not actually save us any lines of code, but it just sticks out a bit to me that we are doing the same thing over and over again in different files.

r+wc

eviljeff:
> Would it be worthwhile still having a helper function to do that, so we don't have to continually specify `settings.MEDIA_ROOT` or `settings.MEDIA_URL` in all of these files? I realize that might not actually save us any lines of code, but it just sticks out a bit to me that we are doing the same thing over and over again in different files.

That's exactly what `user_media_url` and `user_media_path` were 🙂 and the issue was to remove them and just use the settings directly.

> (which is sometimes accomplished via a template string and other times via `os.path.join`).

urls are best with a template string; paths have to use `os.path.join` to be os-aware.

PR #19528, opened 2022-08-03
File.filename -> File.file.name; File.file_path -> File.file.path

### LINKED ISSUE #19215: Remove `File.filename`, `File.file_path`
Following https://github.com/mozilla/addons-server/issues/19207 we can remove `File.filename` and `File.file_path` properties - they are there for backwards-compatibility, but we don't have to keep them, we can switch our code to use `File.file.name` and `File.file.path` instead.

Note that those properties do a `self.file` check that might still be useful in some cases, particularly in tests that create dummy `File` instances without an actual file on the filesystem.

eviljeff:
fixes #19215

### DIFF
--- a/src/olympia/addons/indexers.py
+++ b/src/olympia/addons/indexers.py
@@ -539,7 +539,7 @@ class AddonIndexer:
                     {
                         'id': version_obj.file.id,
                         'created': version_obj.file.created,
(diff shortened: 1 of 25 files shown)


diox (approved):
Let's file a separate issue for the fixtures that need to be updated.

PR #19568, opened 2022-08-15
upgrade to python 3.10

### LINKED ISSUE #19560: Upgrade to python 3.10
We're currently running python 3.9, which will receive security fixes for another ~3 years, but had it's final bugfix release back in May (3.9.13, release on 2022-05-17).  To keep up to date we should upgrade to python 3.10 which was released last year.

At this stage none of our dependencies have been checked to see if they support 3.10.

eviljeff:
Could do with some more testing and perhaps an inspection of the stated python compatibility of significant dependencies (celery, elasticsearch, etc) but tests pass - fixes #19560

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -283,7 +283,7 @@ references:
 
   docker-images:
     # These versions should try to match what we run in production.
-    - &python "cimg/python:3.9"
+    - &python "cimg/python:3.10"
     - &redis "redis:6.2"
     - &memcached "memcached:1.4"
     - &mysql "mysql:8.0.28"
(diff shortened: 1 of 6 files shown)


eviljeff:
(I just chose pygit1.7 as it was the lower upgrade that supported python3.10 to minimize compatibility issues)

diox:
> (I just chose pygit1.7 as it was the lower upgrade that supported python3.10 to minimize compatibility issues)

I think we should upgrade pygit first separately. We used to need to compile it ourselves because the wheels were broken/not available for our config, and we include libgit2-dev in our CI/Docker images for that reason. Nowadays that shouldn't be necessary, 1.4.0 (the version we're currently using on `master`) fixed that so we should be able to remove that cruft and upgrade to a more recent one.

eviljeff:
Rebased against master with the pygit2 upgraded landed already.

diox (approved):

Issue #19578, opened 2022-08-18
Return add-on slug in abuse report API responses


diox:
https://github.com/mozilla/addons-server/issues/18879 dropped the `addon` foreign key from `AbuseReport` and in doing so, stopped returning `slug` and `id` of the add-on being reported in the response.

When an abuse report is submitted using the `slug` or `id` though, we _should_ return the `slug` and `id` in the response. And might as well do it at all times if the add-on is public (even when a `guid` was used). addons-frontend is using that in https://github.com/mozilla/addons-frontend/issues/11780 when it's not using the Firefox API (which submits by `guid`) to submit the abuse report.

ioanarusiczki:
@diox With https://addons-dev.allizom.org/api/v5/abuse/report/addon/ I sent several abuse reports using id, guid or slug -> the response returned "id" and "slug" every time (unlike -stage where I've noticed the issue still reproducible).

PR #19580, opened 2022-08-18
Add add-on slug and id back in abuse report responses

diox:
Fixes https://github.com/mozilla/addons-server/issues/19578

### DIFF
--- a/src/olympia/abuse/serializers.py
+++ b/src/olympia/abuse/serializers.py
@@ -1,5 +1,6 @@
 from rest_framework import serializers
 
+from django.http import Http404
 from django.utils.translation import gettext_lazy as _
 
 import olympia.core.logger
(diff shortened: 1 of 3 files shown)


src/olympia/abuse/views.py
```diff
@@ -52,19 +52,28 @@ def get_addon_object(self):
...
+            guid = self.kwargs['addon_pk']
         else:
             # At this point the parameter is a slug or pk. For backwards-compatibility
             # we accept that, but ultimately record only the guid.
             self.get_addon_object()
             if self.addon_object:
-                data['guid'] = self.addon_object.guid
-        return data
+                guid = self.addon_object.guid
+        return guid
```
eviljeff, line 76: kind of a nit, but may as well just return the `guid` value directly in the `if` and `else` rather than creating a `guid` variable.

eviljeff, line 76: kind of a nit, but may as well just return the `guid` value directly in the `if` and `else` rather than creating a `guid` variable.

src/olympia/abuse/tests/test_serializers.py
```diff
@@ -16,8 +16,45 @@
 
 
 class TestAddonAbuseReportSerializer(TestCase):
-    def serialize(self, report, **extra_context):
-        return AddonAbuseReportSerializer(report, context=extra_context).data
+    def serialize(self, report, context=None):
+        return dict(AddonAbuseReportSerializer(report, context=context or {}).data)
```
eviljeff, line 20: Why the wrapping in `dict`?  I've not seen this pattern elsewhere in our tests.

src/olympia/abuse/serializers.py
```diff
@@ -153,19 +154,30 @@ def to_internal_value(self, data):
...
     def get_addon(self, obj):
-        return {
+        output = {
             'guid': obj.guid,
-            'id': None,  # For backwards-compatibility.
-            'slug': None,  # For backwards-compatibility.
+            'id': None,
+            'slug': None,
         }
+        if view := self.context.get('view'):
```
eviljeff, line 170: I'm not following how this is connected to the pr title - won't it always include the `pk` and `slug` if there's a view, irregardless of how the lookup was made?

src/olympia/abuse/tests/test_serializers.py
diox: Convenience - `<serializer.data>` is an `OrderedDict` and so if the `assert` fails, it doesn't have as nice diffs as if it was a `dict` comparison - it shows the diff as a tuple of tuples instead IIRC. I had an issue with the patch initially and that helped narrow it down.

src/olympia/abuse/serializers.py
diox: It would - if the add-on is public (so it shouldn't cause issues). I could have relied on the `view.addon_object` property we used for caching, but I thought it was a little too magical, and using `get_addon_object()` would be more consistent.

eviljeff: so... this pr doesn't do what the issue/pr is titled (i.e. return the slug/id **if** it was used in the request) but just always returns the slug and id if the add-on is public?  And I guess it also wouldn't return the slug/id for a non-public add-on, even if the id/slug was used in the abuse report?

Should the issue be updated to reflect the change in functionality?  (And will this new operation still fix the issue on frontend?)

diox: Yes - essentially it reverts to the original behavior (except without storing the `addon` fk). I'll update the issue and PR title.

eviljeff (approved):

PR #19582, opened 2022-08-19
update file fixtures to use the path id_to_path would return

### LINKED ISSUE #19577: Updating fixtures to use modern directory structure for `File.file` property
We still have many fixtures using the old directory structure (`<addon_id>/<filename.xpi>` instead of `<last 2 digits of addon_id>/<last 4 digits of addon_id>/<addon_id>/<filename.xpi>`). This is dangerous: it hid a problem when we changed the directory structure that had to be hotfixed (https://github.com/mozilla/addons-server/issues/19243) - the code at the time wasn't handling new directory structure in the update service, and tests were passing because the fixtures used by the corresponding tests were all using the old structure.

Example: https://github.com/mozilla/addons-server/blob/43591247351c2c62c415fac0b383db2005200bfa/src/olympia/amo/fixtures/base/addon_3615.json#L204

So we should switch them all to the new structure.

eviljeff:
fixes #19577 - note, manual changes ahead 😬


diox (approved):
We should also be able to remove `test_url_new_directory_structure()` in `addons/tests/test_update.py` with that change.

PR #19583, opened 2022-08-19
Don't store locale&app prefixes when saving SecondaryHero/Shelves CTA URLs

### LINKED ISSUE #19258: Secondary shelf module CTA URLs are being wrongly transformed and lead to broken pages
**STR:**
- In the admin, set up a secondary shelf with a secondary hero module that has a Cta Url pointing to an URL on AMO with an app (i.e. firefox) and locale (i.e. en-US) in it.
- Once cache clears on the frontend (*), click on the link for that module on the homepage

**Expected results:**
- The page linked originally loads

**Actual results:**
- You end up loading a page with the app and locale re-added, which errors out. See examples below.

**Examples of broken CTA urls:**
- https://addons-dev.allizom.org/en-US/firefox/addon/spellcheck-enabler/ if you're visiting the homepage in `fr`, you'll end up loading https://addons-dev.allizom.org/fr/firefox/en-US/firefox/addon/spellcheck-enabler/ instead (+ some query string for utm parameters). It would work in `en-US`, though.
- https://addons-dev.allizom.org/firefox/addon/spellcheck-enabler/ would be transformed to 
https://addons-dev.allizom.org/en-US/firefox/firefox/addon/spellcheck-enabler/ (+ some query string for utm parameters).

The CTA url in the API is correct: the problem lies on the frontend. We should fix it, but note that we should probably prevent linking to an URL in a specific locale in the admin, because if we obey that, users would end up on a page translated in a different language than the one they were on initially.

---

(*) include a query string in the CTA url when you're changing it : you'll now cache has cleared and you're looking at fresh data when you see that query string in the CTA url on the homepage. This applies to both broken and working scenarios.

diox:
Fixes #19258

### DIFF
--- a/src/olympia/hero/tests/test_models.py
+++ b/src/olympia/hero/tests/test_models.py
@@ -198,6 +198,22 @@ class TestSecondaryHero(TestCase):
         hero.cta_url = 'http://goo.gl'
         hero.clean()  # it raises if there's an error
 
+    def test_clean_cta_remove_prefixes(self):
+        hero = SecondaryHero.objects.create()
+
+        with self.activate(locale='en-US', app='firefox'):
+            hero.cta_url = '/en-US/firefox/addon/foo'
+            hero.clean()
+            assert hero.cta_url == '/addon/foo/'
+
+            hero.cta_url = '/en-US/firefox/addon/foo/'
+            hero.clean()
+            assert hero.cta_url == '/addon/foo/'
+
+            hero.cta_url = '/fr/android/collections/4757633/privacy-matters'
+            hero.clean()
+            assert hero.cta_url == '/collections/4757633/privacy-matters/'
+
     def test_clean_only_enabled(self):
         hero = SecondaryHero.objects.create(
             headline='Its a héadline!', description='description'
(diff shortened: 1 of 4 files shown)


src/olympia/hero/tests/test_models.py
```diff
@@ -198,6 +198,22 @@ def test_clean_cta(self):
...
+
+        with self.activate(locale='en-US', app='firefox'):
+            hero.cta_url = '/en-US/firefox/addon/foo'
+            hero.clean()
+            assert hero.cta_url == '/addon/foo/'
+
+            hero.cta_url = '/en-US/firefox/addon/foo/'
+            hero.clean()
+            assert hero.cta_url == '/addon/foo/'
+
```
eviljeff, line 212: Do we need tests for a url that's missing the app or locale?

eviljeff (approved):

src/olympia/hero/tests/test_models.py
diox: Yeah. Added a few more tests in 9bf8b1251a2a22b9bcadeb3bf142c8207ccdbbd3

PR #19587, opened 2022-08-19
Allow REDIRECT_URL to be set through an environement variable

### LINKED ISSUE #19581: Allow REDIRECT_URL to be overridden by env variables
Currently `REDIRECT_URL` is hardcoded [[1]](https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/settings_base.py#L714), and we'd like it to be overridden by the same named env variable. Just like how `REDIRECT_SECRET_KEY` works.

### What happened?
When we migrate the outgoing service to a different cloud provider, this allows us to use a native domain name provided by that provider and at the same time not have to worry about manual SSL certificate renewal.

We expect that the `REDIRECT_URL` setting is able to be overridden by an env variable. If the same named env variable is not present, then the setting should use its current value as its default.

n/a

diox:
Fixes #19581

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -711,7 +711,7 @@ CACHES['default']['BACKEND'] = 'django.core.cache.backends.memcached.PyMemcacheC
 CACHES['default']['KEY_PREFIX'] = 'amo:%s:' % BUILD_ID
 
 # Outgoing URL bouncer
-REDIRECT_URL = 'https://outgoing.prod.mozaws.net/v1/'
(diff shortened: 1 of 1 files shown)


eviljeff (approved):

bqbn:
@diox What about https://github.com/mozilla/addons-server/blob/master/src/olympia/conf/dev/settings.py#L68 and https://github.com/mozilla/addons-server/blob/master/src/olympia/conf/stage/settings.py#L65? Do we need to do anything about them? 

We should be able to override the values of these too.

diox:
Ugh missed those, will fix, thanks for spotting that.

PR #19590, opened 2022-08-22
Also allow REDIRECT_URL to be set through an env variable on dev/stage

diox:
Fixes https://github.com/mozilla/addons-server/issues/19581


eviljeff (approved):
r+

(Doing a quick search to verify no more other use of `REDIRECT_URL` does expose the tests where we directly assign to `settings` rather than use the `override_settings` helper though 😬)

PR #19594, opened 2022-08-22
Limit image upload size in devhub like we do in the API

### LINKED ISSUE #19137: Limit an upload for screenshots to max. 4 MB
As discussed https://github.com/mozilla/addons-server/issues/19007#issuecomment-1104185647
 
When using the API https://addons-dev.allizom.org/api/v5/addons/addon/621486/previews/ with an image >4 MB the error message is:  "Images must be smaller than 4MB"

The same error should be displayed when making the upload from dev hub because currently the upload is not limited:

https://addons-dev.allizom.org/en-US/developers/addon/unlisted-example-05-272/edit

diox:
Fixes #19137

### DIFF
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -1035,7 +1035,7 @@ def upload_image(request, addon_id, addon, upload_type):
         else:
             max_size = settings.MAX_IMAGE_UPLOAD_SIZE
 
-        if max_size and upload_preview.size > max_size:
+        if upload_preview.size > max_size:
             errors.append(
                 gettext('Please use images smaller than %dMB.')
                 % (max_size // 1024 // 1024)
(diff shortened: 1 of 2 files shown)


src/olympia/devhub/views.py
```diff
@@ -1033,14 +1033,13 @@ def upload_image(request, addon_id, addon, upload_type):
         if is_icon:
             max_size = settings.MAX_ICON_UPLOAD_SIZE
         else:
-            max_size = None
+            max_size = settings.MAX_IMAGE_UPLOAD_SIZE
 
         if max_size and upload_preview.size > max_size:
```
eviljeff, line 1038: nit: we don't need the `max_size and` now `max_size` will always have a non-falsey value.

eviljeff (approved):

PR #19598, opened 2022-08-23
Also fix Hero shelves/modules cta_urls that start with EXTERNAL_SITE_URL

### LINKED ISSUE #19258: Secondary shelf module CTA URLs are being wrongly transformed and lead to broken pages

diox:
In the admin, `SITE_URL` is different from `EXTERNAL_SITE_URL`, let's fix both.

Fixes #19258

### DIFF
--- a/src/olympia/hero/tests/test_models.py
+++ b/src/olympia/hero/tests/test_models.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 
 from django.core.exceptions import ValidationError
+from django.test.utils import override_settings
 
 from olympia.amo.tests import addon_factory, TestCase
 from olympia.amo.tests.test_helpers import get_uploaded_file
@@ -231,6 +232,27 @@ class TestSecondaryHero(TestCase):
             hero.clean()
             assert hero.cta_url == '/en-US/firefox/something/weird/'
 
+    def test_clean_cta_remove_prefixes_with_site_url(self):
+        hero = SecondaryHero.objects.create()
+
+        with self.activate(locale='en-US', app='firefox'):
+            hero.cta_url = 'http://testserver/en-US/firefox/addon/foo'
+            hero.clean()
+            assert hero.cta_url == '/addon/foo/'
+
+        with self.activate(locale='en-US', app='firefox'):
+            # With default test settings, this isn't transformed because
+            # neither the SITE_URL nor the EXTERNAL_SITE_URL match.
+            hero.cta_url = 'https://addons.mozilla.org/en-US/firefox/addon/foo'
+            hero.clean()
+            assert hero.cta_url == 'https://addons.mozilla.org/en-US/firefox/addon/foo'
+
+            # When overriding the EXTERNAL_SITE_URL however, it works.
+            with override_settings(EXTERNAL_SITE_URL='https://addons.mozilla.org'):
+                hero.cta_url = 'http://testserver/en-US/firefox/addon/foo'
+                hero.clean()
+                assert hero.cta_url == '/addon/foo/'
+
     def test_clean_only_enabled(self):
         hero = SecondaryHero.objects.create(
             headline='Its a héadline!', description='description'
@@ -292,8 +314,9 @@ class TestSecondaryHeroModule(TestCase):
             shelf=SecondaryHero.objects.create()
         )
 
+        module.cta_text = 'something'
+
         with self.activate(locale='en-US', app='firefox'):
-            module = SecondaryHero.objects.create()
             module.cta_url = '/en-US/firefox/addon/foo'
             module.clean()
             assert module.cta_url == '/addon/foo/'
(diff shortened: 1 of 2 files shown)


src/olympia/hero/tests/test_models.py
```diff
@@ -231,6 +232,27 @@ def test_clean_cta_remove_prefixes(self):
...
+            # neither the SITE_URL nor the EXTERNAL_SITE_URL match.
+            hero.cta_url = 'https://addons.mozilla.org/en-US/firefox/addon/foo'
+            hero.clean()
+            assert hero.cta_url == 'https://addons.mozilla.org/en-US/firefox/addon/foo'
+
+            # When overriding the EXTERNAL_SITE_URL however, it works.
+            with override_settings(EXTERNAL_SITE_URL='https://addons.mozilla.org'):
+                hero.cta_url = 'http://testserver/en-US/firefox/addon/foo'
+                hero.clean()
+                assert hero.cta_url == '/addon/foo/'
```
eviljeff, line 254: This test isn't too clear to me - we're changing `hero.cta_url` as well to another url that isn't `addons.mozilla.org` and then asserting it works.  Maybe we keep `hero.cta_url` the same?

src/olympia/hero/tests/test_models.py
```diff
@@ -231,6 +232,27 @@ def test_clean_cta_remove_prefixes(self):
...
+
```
eviljeff, line 255: Also, can we repeat with `SITE_URL`?

src/olympia/hero/tests/test_models.py
```diff
@@ -323,6 +346,33 @@ def test_clean_cta_remove_prefixes(self):
...
+            module.clean()
+            assert (
+                module.cta_url == 'https://addons.mozilla.org/en-US/firefox/addon/foo'
+            )
+
+            # When overriding the EXTERNAL_SITE_URL however, it works.
+            with override_settings(EXTERNAL_SITE_URL='https://addons.mozilla.org'):
+                module.cta_url = 'http://testserver/en-US/firefox/addon/foo'
+                module.clean()
+                assert module.cta_url == '/addon/foo/'
```
eviljeff, line 374: same comments here

eviljeff (approved):

src/olympia/hero/tests/test_models.py
diox: I should have used hero.cta_url = 'https://addons.mozilla.org/en-US/firefox/addon/foo' in the line above. I'll fix.

src/olympia/hero/tests/test_models.py
diox: The initial assertion at the top of this method is doing that, since `SITE_URL is `http://testserver` in tests.

src/olympia/hero/tests/test_models.py
```diff
@@ -231,6 +232,29 @@ def test_clean_cta_remove_prefixes(self):
...
+        with self.activate(locale='en-US', app='firefox'):
+            # With default test settings, this isn't transformed because
+            # neither the SITE_URL nor the EXTERNAL_SITE_URL match.
+            hero.cta_url = 'https://addons.mozilla.org/en-US/firefox/addon/foo'
+            hero.clean()
+            assert hero.cta_url == 'https://addons.mozilla.org/en-US/firefox/addon/foo'
+
+            # When overriding the EXTERNAL_SITE_URL however, it works.
+            with override_settings(EXTERNAL_SITE_URL='https://addons.mozilla.org'):
+                hero.cta_url = 'https://addons.mozilla.org/en-US/firefox/addon/foo'
```
eviljeff, line 254: nit: `hero.cta_url` was already that value - it was `assert`ed that it was on line 250.

diox: I know, but I figured it was a bit more explicit and easier to cut&paste that way.  (I almost made a bunch of different methods - I would have loved to use `pytest.mark.parametrize` but it doesn't play nice with methods vs functions)

PR #19616, opened 2022-08-25
Record an ActivityLog and IPLog for new ratings, logins, and restrictions events

### LINKED ISSUE #19602: Log IPs from where ratings are posted in IPLog
We already log the IP address when a user rating is posted, but we do that in the `Rating` model (`reviews` table). For consistency and to help https://github.com/mozilla/addons-server/issues/17504 I'd like to store that as an `IPLog` instance linked to the corresponding `ActivityLog` instead.

### LINKED ISSUE #19615: Log IPs from where users are logging in IPLog
We already log the IP address a user logs in with and, but we do that in the UserProfile model (`users` table, `last_login_ip` column). For consistency and to help #17504 I'd like to store that as an `IPLog` instance linked to an `ActivityLog` instead (that activity action doesn't exist yet).

On top of that, when users hit a restriction, we store their current ip address and `last_login_ip` in `UserRestrictionHistory`. Like above, let's store that in an `IPLog` linked to the activity log instead. Only the current `ip_address` would need to be stored, since we'd already have a record of the `last_login_ip` one through the logging in action.

See also #19602 which does the same for `Rating`.

diox:
Fixes #19602 
Fixes #19615

Logins for users that logged in since 2019 and their restrictions are backfilled. In addition, keep non-permanent activity logs (like new ratings) for 180 days instead of 90 and store the ip address as binary in `IPLog` for future use in database queries.


src/olympia/activity/migrations/0016_auto_20220826_1055.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 3.2.15 on 2022-08-26 10:55
+
+import ipaddress
+from django.db import migrations
+
+
+def backfill_iplog_ip_address_binary(apps, schema_editor):
+    IPLog = apps.get_model('activity', 'IPLog')
+    for log in IPLog.objects.filter(ip_address_binary__isnull=True).iterator():
```
diox, line 9: Could be done in pure SQL with `UPDATE log_activity_ip SET ip_address_binary = INET6_ATON(ip_address) WHERE ip_address_binary IS NULL;`. Doing it in Python is slower, but the bet is that this slowness is useful to let the replicas catch up. I could also do a backfill command that spawns tasks if we think we need it.

src/olympia/activity/models.py
```diff
@@ -733,13 +748,15 @@ def create(cls, action, *args, **kw):
                 ReviewActionReasonLog.objects.create(reason_id=id_, **create_kwargs)
 
         if getattr(action, 'store_ip', False):
+            ip_address = core.get_remote_addr()
```
eviljeff, line 751:
```suggestion
        if getattr(action, 'store_ip', False) and (ip_address := core.get_remote_addr()):
```
(and then you don't need the next change)

src/olympia/activity/models.py
```diff
@@ -212,6 +213,16 @@ class IPLog(ModelBase):
 
     activity_log = models.ForeignKey('ActivityLog', on_delete=models.CASCADE)
     ip_address = models.CharField(max_length=45)
+    ip_address_binary = VarBinaryField(max_length=16, null=True)
```
eviljeff, line 216: I guess there isn't a django field that stores the ip address in an SQL compatible way and also has a nice string representation/setter in django?  (It'd be nice to have a single ip address field)

src/olympia/constants/activity.py
```diff
@@ -826,6 +826,24 @@ class FORCE_ENABLE(_LOG):
     short = _('Force enabled')
 
 
+class LOG_IN(_LOG):
+    id = 169
+    # Note: clear_old_user_data cron would delete the IPLog when needed, so we
+    # can keep the activity log, it just won't have any IP associated with it.
+    keep = True
```
eviljeff, line 833: what are we keeping the activity log for, after the IP address log has been deleted?

src/olympia/devhub/tests/test_views_edit.py
```diff
@@ -873,7 +873,7 @@ def test_edit_media_uploadedicon(self):
 
     def test_edit_media_icon_log(self):
         self.test_edit_media_uploadedicon()
-        log = ActivityLog.objects.all()
+        log = ActivityLog.objects.exclude(action=amo.LOG.LOG_IN.id)
```
eviljeff, line 876: if we have to do this a lot maybe we need a manger/queryset shortcut for objects without `LOG_IN`?

eviljeff: also, `LOG.LOG_IN` 😬

src/olympia/users/management/commands/process_users.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class Command(ProcessObjectsCommand):
+    def get_model(self):
+        return UserProfile
+
+    def get_tasks(self):
+        return {
+            # Don't forget to use the --with_deleted argument when triggering
+            # this task!
```
eviljeff, line 19: why?

src/olympia/users/management/commands/process_users.py
```diff
@@ -0,0 +1,26 @@
...
+            'backfill_activity_and_iplog': {
+                'task': backfill_activity_and_iplog,
+                'queryset_filters': [
+                    ~Q(last_login_ip='') & Q(last_login__gte=datetime(2019, 1, 1))
```
eviljeff, line 23: I thought we cleared the logs after 90 days (increasing to 180 days) so would there be anything after `datetime(2022, 5, 1)`?  (Or, if there will be, would those logs have been deleted long ago if this code was in place in 2019?)

src/olympia/activity/models.py
eviljeff: If there isn't, maybe we could implement one on top of `VarBinaryField` as a follow-up?

eviljeff (approved):

src/olympia/activity/models.py
diox: There isn't. That's not a bad idea...

src/olympia/constants/activity.py
diox: I guess it's not always that useful, but IMHO it's still useful to know an action took place, and what it was, even if you don't have the IP, so you can reconstruct what they did.

src/olympia/users/management/commands/process_users.py
diox: I need to remove this - I wanted to make sure to run this even on deleted users, but `UserProfile` doesn't have `unfiltered`, so it will do that out of the box (in fact `--with_deleted` doesn't work)

src/olympia/users/management/commands/process_users.py
diox: We're keeping `LOG_IN` (and `RESTRICTED`) activities forever. The 90/180 days is for `ADD_RATING` (and a bunch of others that existed before) which I'm not backfilling (see https://github.com/mozilla/addons-server/issues/19602#issuecomment-1227360838)

src/olympia/devhub/tests/test_views_edit.py
diox: IMHO it'd not worth having a shortcut for it - We could make each of those assertions look for specific actions instead of excluding but I didn't have the motivation to do that...

src/olympia/activity/migrations/0016_auto_20220826_1055.py
diox: https://github.com/mozilla/addons-server/pull/19616/commits/cdd90d91db5a030a4bcf1703484a2e09d33f8aae

src/olympia/activity/models.py
diox: https://github.com/mozilla/addons-server/pull/19616/commits/7043bce4e8ee855c1e43b9b58566f9d4b66072ea

src/olympia/users/management/commands/process_users.py
diox: https://github.com/mozilla/addons-server/pull/19616/commits/f8fd73507f33a1acd6a99e1677c768a309a82d2d

PR #19626, opened 2022-08-30
Use IPLog only for IP searches in the admin

### LINKED ISSUE #17504: Optimize IP address query performance
Searching for users by ip address is currently a bit slow, the data is scattered on different tables, IPs are stored in string form as a `varchar(45)`, and not always indexed.

One fix to investigate would be to store them as `blob` or `varbinary`, using `INET6_ATON()` and friends in MySQL, or `ipaddress.ip_address().packed` in python) and indexing that.  That's blocked by https://github.com/mozilla/addons-server/issues/14547 because we currently do manual SQL queries with `FIND_IN_SET()`, which would definitely not work anymore with that approach.

diox:
Using `IPLog` only makes the query a lot more efficient, previously the multiple joins + OR were forcing MySQL to do a full table scan.

In addition:
- Leverage `ip_address_binary` (and most importantly, for faster searches (and in the future, ranges)
- Use `GROUP_CONCAT()` to find and display all other IPs for each user in results
- Refactor existing frontend code to dynamically add/remove IPs from results to use the new layout

Fixes #17504

### DIFF
--- a/src/olympia/users/admin.py
+++ b/src/olympia/users/admin.py
@@ -185,13 +185,11 @@ class UserAdmin(CommaSearchInAdminMixin, admin.ModelAdmin):
     def get_search_results(self, request, queryset, search_term):
         ips = self.ip_addresses_if_query_is_all_ip_addresses(search_term)
         if ips:
-            q_objects = Q()
-            for ip in ips:
-                q_objects |= Q(
-                    activitylog__iplog__ip_address_binary=ipaddress.ip_address(
-                        ip
-                    ).packed
-                )
+            condition = Q(
+                activitylog__iplog__ip_address_binary__in=[
+                    ipaddress.ip_address(ip).packed for ip in ips
+                ]
+            )
             # We want to duplicate the joins against activitylog + iplog so
             # that one is used for the search, and the other for the group
             # concat showing all IPs for activities of that user. Django
@@ -202,7 +200,7 @@ class UserAdmin(CommaSearchInAdminMixin, admin.ModelAdmin):
                     'activitylog__iplog__ip_address', distinct=True
                 ),
                 'activitylog_filtered': FilteredRelation(
-                    'activitylog__iplog', condition=q_objects
+                    'activitylog__iplog', condition=condition
                 ),
             }
             # ...and then add the most simple filter to "activate" the join
--- a/src/olympia/users/admin.py
+++ b/src/olympia/users/admin.py
@@ -422,7 +422,7 @@ class UserAdmin(CommaSearchInAdminMixin, admin.ModelAdmin):
     picture_img.short_description = _('Profile Photo')
 
     def known_ip_adresses(self, obj):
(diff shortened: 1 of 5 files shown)


src/olympia/users/tests/test_admin.py
```diff
@@ -93,35 +93,70 @@ def test_search_ip_as_int_isnt_considered_an_ip(self):
...
-        assert not doc('.column-_ratings_all__ip_address')
+        assert not doc('.column-known_ip_adresses')
 
     def test_search_for_single_ip(self):
         user = user_factory(email='someone@mozilla.com')
         self.grant_permission(user, 'Users:Edit')
         self.client.force_login(user)
-        user_factory(last_login_ip='127.0.0.1')  # Extra user that shouldn't match
-        self.user.update(email='foo@bar.com', last_login_ip='127.0.0.2')  # Will match
+        with core.override_remote_addr('127.0.0.2'):
```
eviljeff, line 102: can we have some `assertNumQueries` in these tests?  It'd be good to know how we're doing currently, at least.

src/olympia/users/admin.py
```diff
@@ -190,21 +186,33 @@ def get_search_results(self, request, queryset, search_term):
...
-            # a feature: it will happen when the IPs returned are different
-            # (so technically the rows are not duplicates), since the
-            # annotations are part of the distinct().
-            queryset = queryset.distinct()
+            for ip in ips:
+                q_objects |= Q(
+                    activitylog__iplog__ip_address_binary=ipaddress.ip_address(
+                        ip
+                    ).packed
+                )
```
eviljeff, line 194: I'd natively assume this could be replace with a single `__in` query:
`q_objects = Q(activitylog__iplog__ip_address_binary__in=[ipaddress.ip_address(ip).packed for ip in ips])`
or does that not work?

src/olympia/users/admin.py
```diff
@@ -438,30 +424,40 @@ def picture_img(self, obj):
...
-            Rating.objects.filter(user=obj)
-            .values_list('ip_address', flat=True)
-            .order_by()
-            .distinct()
-        )
-        ip_adresses.update(
-            itertools.chain(
-                *UserRestrictionHistory.objects.filter(user=obj)
-                .values_list('last_login_ip', 'ip_address')
+        # activity_ips is a, annotation added by get_search_results() above
```
eviljeff, line 427: weird comment phrasing.  "an annotation"?

eviljeff (approved):

src/olympia/users/admin.py
diox: I naively assumed it wouldn't work because it's all binary strings, but actually, you're probably right, it might work...

src/olympia/users/tests/test_admin.py
diox: https://github.com/mozilla/addons-server/pull/19626/commits/39ce33266b5430775385dc78e02d756e62a722e1

PR #19646, opened 2022-09-01
work around django4.0's extra elide_empty kwarg to Query.get_compiler

### LINKED ISSUE #19638: Fix `TranslationQuery.get_compiler() got an unexpected ... 'elide_empty'` in django4.x
the [main CI job](https://app.circleci.com/pipelines/github/eviljeff/olympia/960/workflows/a1adc3e4-e6c4-40d8-bcd0-b75214890a85/jobs/10839?invite=true#step-112-6206) running under django4.1 raises the error `TypeError: TranslationQuery.get_compiler() got an unexpected keyword argument 'elide_empty'` for some tests.

eviljeff:
fixes #19638 + a bit of random cleanup from pre-django2.2

### DIFF
--- a/src/olympia/translations/query.py
+++ b/src/olympia/translations/query.py
@@ -90,10 +90,16 @@ class TranslationQuery(models.sql.query.Query):
         c.translation_aliases = self.translation_aliases
         return c
 
-    def get_compiler(self, using=None, connection=None):
+    def get_compiler(self, using=None, connection=None, **kwargs):
         # Call super to figure out using and connection.
-        c = super().get_compiler(using, connection)
-        return SQLCompiler(self, c.connection, c.using)
+        compiler = super().get_compiler(using, connection, **kwargs)
+        # slightly weird kwargs because django4.0 adds an extra elide_empty kwarg
+        return SQLCompiler(
+            self,
+            compiler.connection,
+            compiler.using,
+            **{kw: getattr(compiler, kw, None) for kw in kwargs if kw == 'elide_emtpy'},
+        )
 
 
 class SQLCompiler(compiler.SQLCompiler):
@@ -102,13 +108,9 @@ class SQLCompiler(compiler.SQLCompiler):
     def get_from_clause(self):
         # Temporarily remove translation tables from query.tables so Django
         # doesn't create joins against them.
-        # self.query.tables was removed in django2.0+
-        old_tables = list(self.query.tables) if hasattr(self.query, 'tables') else None
         old_map = OrderedDict(self.query.alias_map)
         for table in itertools.chain(*self.query.translation_aliases.values()):
             if table in self.query.alias_map:
-                if old_tables:
-                    self.query.tables.remove(table)
                 del self.query.alias_map[table]
 
         joins, params = super().get_from_clause()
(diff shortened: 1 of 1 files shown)


src/olympia/translations/query.py
```diff
@@ -90,10 +90,16 @@ def clone(self):
...
-        c = super().get_compiler(using, connection)
-        return SQLCompiler(self, c.connection, c.using)
+        compiler = super().get_compiler(using, connection, **kwargs)
+        # slightly weird kwargs because django4.0 adds an extra elide_empty kwarg
+        return SQLCompiler(
+            self,
+            compiler.connection,
+            compiler.using,
+            **{kw: getattr(compiler, kw, None) for kw in kwargs if kw == 'elide_emtpy'},
+        )
```
diox, line 102: Instead of all of this, I suggest:
```diff
diff --git a/src/olympia/translations/query.py b/src/olympia/translations/query.py
index 5e9c48c072..a029e803aa 100644
--- a/src/olympia/translations/query.py
+++ b/src/olympia/translations/query.py
@@ -2,7 +2,7 @@ import itertools
 from collections import OrderedDict
 
 from django.conf import settings
-from django.db import models
+from django.db import connections, models
 from django.db.models.sql import compiler
 from django.db.models.sql.constants import LOUTER
 from django.db.models.sql.datastructures import Join
@@ -90,10 +90,16 @@ class TranslationQuery(models.sql.query.Query):
         c.translation_aliases = self.translation_aliases
         return c
 
-    def get_compiler(self, using=None, connection=None):
-        # Call super to figure out using and connection.
-        c = super().get_compiler(using, connection)
-        return SQLCompiler(self, c.connection, c.using)
+    def get_compiler(self, using=None, connection=None, **kwargs):
+        # Unfortunately even though there is a mechanism to override the
+        # compiler class through `self.compiler`, that needs to point to a
+        # member of the django.db.models.sql.compiler module (unless you're
+        # also customizing the database backend). So instead we do that
+        # here, returning our own custom SQLCompiler class. The parent
+        # implementation can set connection though, so we do that as well.
+        if using:
+            connection = connections[using]
+        return SQLCompiler(self, connection, using, **kwargs)
 
 
 class SQLCompiler(compiler.SQLCompiler):
```

src/olympia/translations/query.py
```diff
@@ -102,13 +108,9 @@ class SQLCompiler(compiler.SQLCompiler):
     def get_from_clause(self):
         # Temporarily remove translation tables from query.tables so Django
         # doesn't create joins against them.
```
diox, line 110: That comment needs to be updated well - just mentioning `query.alias_map` instead of `query.tables` should be enough.

src/olympia/translations/query.py
eviljeff: I looked into overriding the compiler class too - it's not that onerous as we already have a custom database backend for our `PositiveAutoField` field - the annoying part was that it wanted extra classes for Insert (i.e. `SQLInsertCompiler`), Delete and Update too, and I felt like the alternative ended up way more complicated than our current code.

diox: Ah, yeah, forgot we already had a custom backend. My comment should be reworded then to reflect what you just said, but I think the approach is still a bit better than trying to keep up with which parameters are present depending on the django version

diox (approved):

PR #19648, opened 2022-09-01
Remove unused dependencies

### LINKED ISSUE #19647: Remove more unused depencencies
Spotted a few unused dependencies while investigating https://github.com/mozilla/addons-server/issues/19643 , let's clean them up.

diox:
Fixes #19647


diox:
Ah, we're still using `asn1crypto` to parse the signature, but we should be able to do that with just `cryptography`.

diox:
> Ah, we're still using `asn1crypto` to parse the signature, but we should be able to do that with just `cryptography`.

Actually, we can load the certificates using `cryptography.hazmat.primitives.serialization .pkcs7.load_der_pkcs7_certificates()` (*), but I couldn't find a nice way to read signer info, which we need to grab the right certificate. So let's just keep using `asn1crypto` for now.

(*) What about COSE though ? We don't seem to bother with that one...

eviljeff (approved):
is it worth filing a follow-up issue for removing `asn1crypto`?  If we're just using it for one minor thing we might be able to work around it - or implement it directly?  Then again, maybe it's not worth the effort to remove it 🤷

diox:
I've spent like an hour on it and failed - it's not that trivial, so I don't think it's worth spending more time on it now. Ideally, autograph would return the info we want so that we wouldn't have to extract it ourselves, but it appears to be in maintenance mode right now...

PR #19668, opened 2022-09-07
Trim whitespace in admin search when query contains a comma

diox:
Makes is less likely to return '0 results' just because you accidentally added a space or a trailing comma in IP search.

Fixes mozilla/addons#8856


src/olympia/amo/admin.py
```diff
@@ -2,14 +2,40 @@
...
+        self.cleaned_data = super().clean()
+        search_term = self.cleaned_data[SEARCH_VAR]
+        if ',' in search_term:
+            self.cleaned_data[SEARCH_VAR] = ','.join(
+                [
+                    term.strip()
+                    for term in search_term.strip().split(',')
+                    if term.strip()
+                ]
+            )
```
eviljeff, line 24:
```suggestion
        self.cleaned_data[SEARCH_VAR] = ','.join(
            term.strip() for term in self.cleaned_data[SEARCH_VAR].split(',')
        )
```
?

I couldn't see what all the extra `strip()`s were achieving.  (And, sure, always doing it on the string even if no `,` is wasteful but simple code tho.)

eviljeff (approved):

src/olympia/amo/admin.py
diox: I want to remove whitespace around the values but I also don't want to keep empty values: passing `a, ,b` should result in `a,b`. Similarly I want to ignore trailing space: `a,b,` should result in `a,b`, though I guess just doing the `if term.strip()` is enough for that and I can drop the `.strip()` on `search_term`.

src/olympia/amo/admin.py
```diff
@@ -2,14 +2,36 @@
...
 from .models import FakeEmail
 
 
+class CommaSearchInAdminChangeListSearchForm(ChangeListSearchForm):
+    def clean(self):
+        self.cleaned_data = super().clean()
+        search_term = self.cleaned_data[SEARCH_VAR]
+        if ',' in search_term:
+            self.cleaned_data[SEARCH_VAR] = ','.join(
+                [term.strip() for term in search_term.split(',') if term.strip()]
```
eviljeff, line 19: aware we're getting into small nits here, but you don't need the `[...]` for the generator - `string.join` will accept it directly.

diox: :+1: I'll fix that in the next PR that also touches this code

diox:
Test failure is from `master` and was fixed in https://github.com/mozilla/addons-server/pull/19673

PR #19674, opened 2022-09-08
Support ranges and networks in IP admin search

### LINKED ISSUE #19663: Add range search to django admin IP address search
The IP address search on the django admin user profile page allows to search for one or several (comma-separated) IP addresses. 
It would be helpful if I could also search for a range of IPs (e.g. 205.251.192.0 - 205.251.255.255 or 205.251.192.0/18), or even multiple ranges, since especially larger cloud providers often have several, non-consecutive ranges. We can start with one if multiple is a challenge.

Ideally, there would be a way to go from one IP to its corresponding provider range (like we have the `+` button now to add a single IP to the search query). That might require some lookup of IP to range (which I usually do through a manual `whois` on the console). I am not sure if that is feasible on the server-side.

diox:
Fixes #19663

This adds support for searching for `192.168.0.1-192.168.0.254` or `192.168.0.0/24`, and mixing it up with multiple networks/IPs separated by commas. The JavaScript on the frontend doesn't have nice UI to remove networks, but won't suggest the addition of a single IP when it's part of a network already in the search query.

### DIFF
--- a/Makefile-docker
+++ b/Makefile-docker
@@ -36,6 +36,7 @@ jquery.cookie/jquery.cookie.js \
 jszip/dist/jszip.js \
 timeago/jquery.timeago.js \
 underscore/underscore.js \
+netmask/lib/netmask.js \
 
 NODE_LIBS_JQUERY_UI := \
 jquery-ui/ui/data.js \
--- a/src/olympia/users/admin.py
+++ b/src/olympia/users/admin.py
@@ -230,7 +230,8 @@ class UserAdmin(CommaSearchInAdminMixin, admin.ModelAdmin):
                 for network in ips_and_networks['networks']:
                     condition |= Q(
                         activitylog__iplog__ip_address_binary__range=(
-                            network[0], network[-1],
+                            network[0],
+                            network[-1],
                         )
                     )
             # We want to duplicate the joins against activitylog + iplog so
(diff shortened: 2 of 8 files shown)


Makefile-docker
```diff
@@ -36,6 +36,7 @@ jquery.cookie/jquery.cookie.js \
 jszip/dist/jszip.js \
 timeago/jquery.timeago.js \
 underscore/underscore.js \
+netmask/lib/netmask.js \
```
eviljeff, line 39: do you need to update `settings.MINIFY_BUNDLES` too?

static/js/exports.js
```diff
@@ -0,0 +1,2 @@
+// Minimal hack to be able to use JavaScript modules client-side
+var exports = {};
```
eviljeff, line 2: Does this file also need to be added to `settings.MINIFY_BUNDLES`?

src/olympia/users/admin.py
```diff
@@ -171,37 +171,69 @@ def get_list_display(self, request):
...
+                networks.append(ipaddress.ip_network(term))
+                continue
+            except ValueError:
+                pass
+            # Is the search term an IP range ?
+            if term.count('-') == 1:
+                try:
+                    networks.extend(
+                        ipaddress.summarize_address_range(
+                            *(ipaddress.ip_address(i.strip()) for i in term.split('-'))
```
eviljeff, line 208: ... I can't help thinking there is a way to write this that doesn't need all the iterator plus expansion magic for exactly 2 args but this works...

eviljeff (approved):

Makefile-docker
diox: No, we don't have a bundle for admin js/css - they get included through the `Media` class on the `ModelAdmin`, which is good enough - we don't get minification but we do get a unique filename thanks to `staticfiles`.

PR #19678, opened 2022-09-08
return the url to the license for the addon - not version- on frontend

### LINKED ISSUE #19122: `custom_license.url` links to a non-existent page
1. GET the details of an add-on with a custom license - see https://addons-dev.allizom.org/api/v5/addons/addon/621471/
2. Open the link present in the `custom_license.url` field

### What happened?
The `url` links to a 404 page - see https://addons-dev.allizom.org/en-US/firefox/addon/custom_license_url/license/10.222

The `url` has the version number appended, which is what causes this issue. The `url` should link instead to https://addons-dev.allizom.org/en-US/firefox/addon/custom_license_url/license/

Reproduces with add-ons created through DevHub and with the API

eviljeff:
fixes #19122


bobsilverberg (approved):
Looks good. Just one comment about a comment.

r+wc

PR #19695, opened 2022-09-15
Prevent pk parameters in reviewers API to be anything other than digits

### LINKED ISSUE #19694: Raise 404 for reviewers API when the pk in the URL is not a valid number
Sentry Issue: [ADDONS-SERVER-PROD-7Y](https://sentry.io/organizations/mozilla/issues/3574399919/?referrer=github_integration)

```
ValueError: invalid literal for int() with base 10: 'NaN'
  File "django/db/models/fields/__init__.py", line 1823, in get_prep_value
    return int(value)

ValueError: Field 'id' expected a number but got 'NaN'.
(12 additional frame(s) were not displayed)
...
  File "django/db/models/sql/query.py", line 1370, in build_filter
    condition = self.build_lookup(lookups, col, value)
  File "django/db/models/sql/query.py", line 1216, in build_lookup
    lookup = lookup_class(lhs, rhs)
  File "django/db/models/lookups.py", line 25, in __init__
    self.rhs = self.get_prep_lookup()
  File "django/db/models/lookups.py", line 77, in get_prep_lookup
    return self.lhs.output_field.get_prep_value(self.rhs)
  File "django/db/models/fields/__init__.py", line 1825, in get_prep_value
    raise e.__class__(

ValueError: Field 'id' expected a number but got 'NaN'.
```

diox:
Fixes #19694

### DIFF
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -6562,6 +6562,12 @@ class TestAddonReviewerViewSet(TestCase):
         response = self.client.post(self.subscribe_url_listed)
         assert response.status_code == 404
 
+        self.subscribe_url_listed = self.subscribe_url_listed.replace(
+            f'{self.addon.pk + 42}', 'NaN'
+        )
+        response = self.client.post(self.subscribe_url_listed)
+        assert response.status_code == 404
+
     def test_subscribe_already_subscribed_listed(self):
         ReviewerSubscription.objects.create(
             user=self.user, addon=self.addon, channel=amo.CHANNEL_LISTED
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/tests/test_views.py
```diff
@@ -6808,11 +6824,15 @@ def test_disable_addon_does_not_exist(self):
         self.grant_permission(self.user, 'Reviews:Admin')
         self.client.login_api(self.user)
         self.disable_url = reverse_ns(
-            'reviewers-addon-enable', kwargs={'pk': self.addon.pk + 42}
```
bobsilverberg, line 6811: I take it this was a bug in the existing version of the test (pointing to `enable` instead of `disable`)?

bobsilverberg (approved):

src/olympia/reviewers/tests/test_views.py
diox: Yeah, copy/paste error in the previous version of the test

PR #19701, opened 2022-09-19
Add error when developer goes through submission process twice with same upload

### LINKED ISSUE #19700: Add explicit error when a developer goes through the submission process twice
- Log in and go to submit a new add-on
- Select a distribution mechanism, then pick an add-on file
- Hit continue to go to the source code step, but then, use your browser back button to go back to the previous step
- You should see the already validated add-on from the previous step, with the progress bar full
- Hit continue to go to the source code step again

### What happened?
- You see a generic error `Could not parse the manifest file`. Behind the scenes, a sentry error is generated (https://sentry.io/organizations/mozilla/issues/3179698028/?project=6310819&query=is%3Aunresolved in prod)

- A specific error message indicating that the submission process has already started for that add-on, and no sentry error behind the scenes.

diox:
Fixes #19700

### DIFF
--- a/src/olympia/files/utils.py
+++ b/src/olympia/files/utils.py
@@ -106,6 +106,12 @@ def id_to_path(pk, breadth=1):
 def get_file(fileorpath):
     """Get a file-like object, whether given a FileUpload object or a path."""
     if hasattr(fileorpath, 'path'):  # FileUpload
+        if not fileorpath.path:
+            # This upload has already been used to create a Version. Note that
+            # we say "processed" but it's different than the `processed`
+            # property on FileUpload, it's just used here as a generic term to
+            # inform the developer this FileUpload has already been used.
+            raise AlreadyUsedUpload(gettext('This upload is already being processed.'))
         return storage.open(fileorpath.path, 'rb')
     if hasattr(fileorpath, 'name'):
         return fileorpath
@@ -136,6 +142,13 @@ class InvalidManifest(forms.ValidationError):
 
 class InvalidZipFile(forms.ValidationError):
     """This error is raised when we attempt to open an invalid file with SafeZip."""
+
+    pass
+
+
+class AlreadyUsedUpload(forms.ValidationError):
+    """Error raised when trying to use a FileUpload that has already been used
+    to make a Version+File, its path will be empty."""
 
     pass
 
@@ -822,7 +835,7 @@ def parse_xpi(xpi, addon=None, minimal=False, user=None):
                 zip_file.read('manifest.json'), certinfo=certificate_info
             ).parse(minimal=minimal)
         else:
-            raise NoManifestFound('No manifest.json found')
+            raise NoManifestFound(gettext('No manifest.json found'))
 
     except forms.ValidationError:
         raise
--- a/src/olympia/files/utils.py
+++ b/src/olympia/files/utils.py
@@ -111,7 +111,7 @@ def get_file(fileorpath):
             # we say "processed" but it's different than the `processed`
             # property on FileUpload, it's just used here as a generic term to
             # inform the developer this FileUpload has already been used.
-            raise AlreadyUsedUpload(gettext('This upload is already being processed.'))
+            raise AlreadyUsedUpload(gettext('This upload has already been submitted.'))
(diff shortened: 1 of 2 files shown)


src/olympia/files/utils.py
```diff
@@ -140,6 +146,13 @@ class InvalidZipFile(forms.ValidationError):
     pass
 
 
+class AlreadyUsedUpload(forms.ValidationError):
+    """Error raised when trying to use a FileUpload that has already been used
+    to make a Version+File, its path will be empty."""
+
+    pass
```
eviljeff, line 153: I see this is a common pattern in our codebase, but why don't we specify an error string in the class directly for a specific error like this, rather than have to do it where it's used?

src/olympia/files/utils.py
```diff
@@ -106,6 +106,12 @@ def id_to_path(pk, breadth=1):
 def get_file(fileorpath):
     """Get a file-like object, whether given a FileUpload object or a path."""
     if hasattr(fileorpath, 'path'):  # FileUpload
+        if not fileorpath.path:
```
eviljeff, line 109: I don't know which is more "correct", but we have a `FileUpload.submitted` `@property` that checks if `self.addon` is set to determine if the `FileUpload` has been "used" to create a version.

src/olympia/files/utils.py
```diff
@@ -106,6 +106,12 @@ def id_to_path(pk, breadth=1):
...
+            # This upload has already been used to create a Version. Note that
+            # we say "processed" but it's different than the `processed`
+            # property on FileUpload, it's just used here as a generic term to
+            # inform the developer this FileUpload has already been used.
+            raise AlreadyUsedUpload(gettext('This upload is already being processed.'))
```
eviljeff, line 114: `This upload has already been used` or `This upload has already been submitted` (or if we really want to say "processed", even though we have a `FileUpload.processed` property that refers to validation instead, `This upload has already been processed`)

src/olympia/files/utils.py
eviljeff: (we should probably have a single method of determining whether a `FileUpload` has already been used.  Could be a follow-up)

src/olympia/files/utils.py
diox: `ValidationError` requires a message to be passed when instantiated anyway, so we'd have to make something really custom to avoid that... We could, bu :shrug:

src/olympia/files/utils.py
diox: mmm, I don't think we're using that property anywhere. It's wrong/misleading, since nowadays we set `addon` all the time. I guess I could replace it with something like `self.pk and self.path` and use it instead here... but ultimately, what we really care about here is that the path exists...

src/olympia/files/utils.py
eviljeff: something something `__init__` something something kwarg defaults but 🤷 too - just thinking out loud at us creating specific subclasses of `ValidationError` for exceptions but then not adding anything which would render differently than a vanilla `ValidationError`

src/olympia/files/utils.py
eviljeff: :/ looks like it's only uploads created via the addon submission api that don't have `addon` set until they're consumed to create a `Version`

src/olympia/files/utils.py
wagnerand: Can we add a sentence on what the developer can do about this issue? What is the next step? Wait? Proceed somewhere else? Upload a new version?

diox: We don't know their intent, but there is probably nothing to do on their side : the `FileUpload` has already been processed, there is likely already a `Version`. 

If we knew what they did we could be a little more specific but it's difficult to be sure. One way to reproduce, that I described in the issue, is going back after submission and trying to submit again from the same upload. Another would be to manage to quickly submit twice so that the first request gets processed before the second request goes through.

bobsilverberg (approved):
The code looks fine to me. I agree with Andrew's suggestion of changing the message to use "has already been" instead of "is already being" (although both are proper English).

PR #19704, opened 2022-09-19
Validate incoming client_ids and guids more thoroughly

### LINKED ISSUE #19702: Validate incoming client ids/guids more throrougly
Some of our APIs accept "client ids" that Firefox sends. Those are SHA256 hashes, so they should always be a 64 characters hexadecimal value. Some also accept "add-on guids", which are either a proper GUID with 32 hexadecimal characters, 4 hyphens and opening & closing brackets, or a email-like identifier. We have regular expressions constants available to validate both.

Although this doesn't seem to be causing any problems (worst case scenario, we just store garbage useless data - we always use the ORM or an external service which is not susceptible to SQL injections for instance), in some APIs we haven't been validating the values properly before storing them or sending them to other services.

We should fix that. Affected APIs/parameters that I could find:
- Abuse Report API (`client_id`)
- Add-on Recommendations API (`guid`)

Non-affected APIs/parameters (relevant for QA to test the fix wouldn't cause regressions):
- Abuse Report API (`addon` should consider invalid guids as slugs)
- Discovery Recommendations API (`telemetry-client-id`, we already validate it)

diox:
Fixes #19702

### DIFF
--- a/src/olympia/addons/views.py
+++ b/src/olympia/addons/views.py
@@ -1189,6 +1189,8 @@ class AddonRecommendationView(AddonSearchView):
     def filter_queryset(self, qs):
         qs = super().filter_queryset(qs)
         guid_param = self.request.GET.get('guid')
+        if not guid_param or not amo.ADDON_GUID_PATTERN.match(guid_param):
+            raise exceptions.ParseError('Invalid guid parameter')
         taar_enable = self.request.GET.get('recommended', '').lower() == 'true'
         guids, self.ab_outcome, self.fallback_reason = get_addon_recommendations(
             guid_param, taar_enable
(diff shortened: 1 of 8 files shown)


src/olympia/addons/views.py
```diff
@@ -1189,6 +1189,8 @@ def get_results_count(self, results):
     def filter_queryset(self, qs):
         qs = super().filter_queryset(qs)
         guid_param = self.request.GET.get('guid')
+        if not guid_param or not amo.ADDON_GUID_PATTERN.match(guid_param):
+            raise exceptions.ParseError('Invalid guid parameter')
```
eviljeff, line 1193: imho, it'd be more consistent to return fallback results with a `TAAR_LITE_FALLBACK_REASON_xxx` reason, rather than raising with this API. 🤷

eviljeff (approved):

src/olympia/addons/views.py
diox: I almost went that route, but decided that if the `guid` is wrong, it's really a problem with the caller, where as fallback is intended for when something goes wrong on the server.

PR #19712, opened 2022-09-20
Instrument get_addon_recommendations to record result of calling TAAR LITE API

### LINKED ISSUE #19711: Instrument get_addon_recommendations to record result of calling TAAR LITE API
There has been a bug in the TAAR LITE API for quite some time, which we didn't notice because our API simply returns fallback results when it receives no results from TAAR LITE, so we don't interpret that as an error. While some manual and/or automated testing could be added by QA to verify this feature in the future, we should also instrument our `get_addon_recommendations` to record data about TAAR LITE API results so we can set up an alert which would fire in cases like this.

bobsilverberg:
Fixes #19711

### DIFF
--- a/src/olympia/addons/utils.py
+++ b/src/olympia/addons/utils.py
@@ -97,15 +97,15 @@ TAAR_LITE_FALLBACK_REASON_INVALID = 'invalid_results'
 def get_addon_recommendations(guid_param, taar_enable):
     guids = None
     fail_reason = None
-    taar_lite_outcome = 'success'
     if taar_enable:
         guids = call_recommendation_server(
             settings.TAAR_LITE_RECOMMENDATION_ENGINE_URL, guid_param, {}
         )
-        outcome = (
-            TAAR_LITE_OUTCOME_REAL_SUCCESS if guids else TAAR_LITE_OUTCOME_REAL_FAIL
-        )
-        if not guids:
+        if guids:
+            outcome = TAAR_LITE_OUTCOME_REAL_SUCCESS
+            taar_lite_outcome = 'success'
+        else:
+            outcome = TAAR_LITE_OUTCOME_REAL_FAIL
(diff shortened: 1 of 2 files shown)


src/olympia/addons/utils.py
```diff
@@ -110,6 +111,8 @@ def get_addon_recommendations(guid_param, taar_enable):
                 if guids == []
                 else TAAR_LITE_FALLBACK_REASON_TIMEOUT
             )
+            taar_lite_outcome = fail_reason
```
diox, line 114: Maybe use the occasion to rewrite the conditions above into something a bit easier to follow? There are a lot of tests around `guids`... Maybe this:
```python
def get_addon_recommendations(guid_param, taar_enable):
    guids = None
    fail_reason = None
    if taar_enable:
        guids = call_recommendation_server(
            settings.TAAR_LITE_RECOMMENDATION_ENGINE_URL, guid_param, {}
        )
        if guids:
            outcome = TAAR_LITE_OUTCOME_REAL_SUCCESS
            taar_lite_outcome = 'success'
        else:
            outcome = TAAR_LITE_OUTCOME_REAL_FAIL
            fail_reason = (
                TAAR_LITE_FALLBACK_REASON_EMPTY
                if guids == []
                else TAAR_LITE_FALLBACK_REASON_TIMEOUT
            )
            taar_lite_outcome = fail_reason
    else:
        outcome = TAAR_LITE_OUTCOME_CURATED
    if not guids:
        guids = TAAR_LITE_FALLBACKS
    return guids, outcome, fail_reason
```

diox (approved):

src/olympia/addons/utils.py
bobsilverberg: That is a good suggestion, thanks. I have updated the patch.

PR #19722, opened 2022-09-22
Validate tar files more thoroughly and automatically like we do for zip files

### LINKED ISSUE #19721: Validate tar files more thoroughly
Though we already do validate members of `.tar.gz` or `.tar.bz2` files (which you can only use for source code submission at the moment), we do it manually, so it's error prone, you need to remember to call `archive_member_validator()`. Instead, let's create a `SafeTar` class like we have `SafeZip` for zipfiles and use that everywhere. It should validate the file contents when opened.

Validation should remain the same (forbid filenames with `..` or starting with `/`) but we can also add checks against symlinks, device files etc.

diox:
We were already validating "manually", at upload time, this creates a wrapper around `TarFile.open()` to do it automatically at all times, much like we do with `SafeZip` for zip files.

Fixes #19721


src/olympia/addons/fields.py
```diff
@@ -214,15 +214,17 @@ def to_internal_value(self, data):
...
             else:
-                # For tar files we need to do a little more work.
+                # For tar files, opening them through SafeTar() checks that we
+                # can accept it.
                 mode = 'r:bz2' if ext == '.bz2' else 'r:gz'
-                with tarfile.open(mode=mode, fileobj=data) as archive:
-                    for member in archive.getmembers():
-                        archive_member_validator(archive, member)
+                with SafeTar.open(mode=mode, fileobj=data):
+                    pass
```
eviljeff, line 227: Is it possible (practical) to have a similar method for validating for `SafeTar` and `SafeZip`?  (i.e. not `if SafeZip(data).zip_file.testzip() is not None` for zips and an entirely different `with SafeTar.open(mode=mode, fileobj=data)` for tars.)

src/olympia/devhub/forms.py
```diff
@@ -754,17 +754,18 @@ def clean_source(self):
...
-                    # For tar files we need to do a little more work.
+                    # For tar files, opening them through SafeTar() checks that
+                    # we can accept it.
                     mode = 'r:bz2' if source.name.endswith('bz2') else 'r:gz'
-                    with tarfile.open(mode=mode, fileobj=source) as archive:
-                        archive_members = archive.getmembers()
-                        for member in archive_members:
-                            archive_member_validator(archive, member)
+                    with SafeTar.open(mode=mode, fileobj=source):
+                        pass
```
eviljeff, line 768: same question here as for addons/fields.py

src/olympia/addons/fields.py
diox: Possibly - we could do the `testzip()` on opening for all zip files, but we'd need to add more code to ignore any errors about that in the scanners code, since we might already have such broken zip files in storage right now...

diox: I looked at this and I'm not convinced it's a good idea to do that everytime we use `SafeZip`, as it forces us to read through all the contents of all the files in the archive. 

At add-on upload, if a developer uploads a broken archive, their extension won't work, so it's their responsibility, so we don't bother. We check at source code upload because they might not realize then that easily.

diox: Both `SafeZip` and `SafeTar` perform cheap validation checks on init (stuff around filenames).

`testzip()` is a special thing added on top to check CRC, just for source code uploads: it goes through all the files inside the archive and tries to read them, which takes some time & memory to do.

eviljeff: "For tar files, opening them through SafeTar.open() checks that we"

eviljeff (approved):

PR #19728, opened 2022-09-23
implement admin reviewer unreject versions action

### LINKED ISSUE #19683: Allow admin reviewers to unreject a version
In the Reviewer Tools, users with permission `Reviews:Admin` should be able to change the status of a version from `STATUS_DISABLED` to `STATUS_AWAITING_REVIEW`.
Doing so might have an impact on the add-on status, the system should determine the right status automatically.

As a security precaution, the ability to unreject should only be available if the add-on status is not `STATUS_DISABLED`. Admin reviewers should first manually "force enable" an add-on which changes the add-on status to `STATUS_NULL`, to express the explicit intent to enable a previously manually disabled add-on.

eviljeff:
fixes #19683

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -561,9 +561,11 @@ class ReviewHelper:
         is_appropriate_reviewer_post_review = acl.action_allowed_for(
             self.user, permission_post_review
         )
+        is_admin_reviewer = is_appropriate_reviewer and acl.action_allowed_for(
+            self.user, amo.permissions.REVIEWS_ADMIN
+        )
 
-        addon_is_complete_and_not_disabled = self.addon.status not in (
-            amo.STATUS_NULL,
+        addon_is_not_disabled_or_deleted = self.addon.status not in (
             amo.STATUS_DELETED,
             amo.STATUS_DISABLED,
         )
@@ -571,9 +573,8 @@ class ReviewHelper:
             self.addon.status == amo.STATUS_NULL and version_is_unlisted
         )
         addon_is_reviewable = (
-            addon_is_complete_and_not_disabled
-            or addon_is_incomplete_and_version_is_unlisted
-        )
+            addon_is_not_disabled_or_deleted and self.addon.status != amo.STATUS_NULL
+        ) or addon_is_incomplete_and_version_is_unlisted
         version_is_unreviewed = self.version and self.version.is_unreviewed
         addon_is_valid = self.addon.is_public() or self.addon.is_unreviewed()
         addon_is_valid_and_version_is_listed = (
@@ -689,7 +690,7 @@ class ReviewHelper:
             'method': self.handler.approve_multiple_versions,
             'label': _('Approve Multiple Versions'),
             'minimal': True,
-            'versions': True,
+            'multiple_versions': True,
             'details': _(
                 'This will approve the selected versions. '
                 'The comments will be sent to the developer.'
@@ -708,7 +709,7 @@ class ReviewHelper:
                 # or (unlisted and) awaiting review
                 or self.version.file.status == amo.STATUS_AWAITING_REVIEW
             ),
-            'versions': True,
+            'multiple_versions': True,
             'details': _(
                 'This will reject the selected versions. '
                 'The comments will be sent to the developer.'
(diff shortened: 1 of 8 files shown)


src/olympia/reviewers/utils.py
```diff
@@ -1360,6 +1373,61 @@ def reject_multiple_versions(self):
...
+                    'unrejecting versions',
+                )
+            )
+            self.addon.update(status=amo.STATUS_NOMINATED)
+            ActivityLog.create(
+                amo.LOG.CHANGE_STATUS, self.addon, self.addon.status, user=self.user
+            )
+            self.addon.update_version()
+
+        # The reviewer should be automatically subscribed to any new versions posted to
```
wagnerand, line 1425: I am not sure we need to subscribe the reviewer. What scenario did you have in mind?

eviljeff: 🤷 up to you really.  It's standard for the other reviewer actions but we can omit it for this one.

eviljeff: https://github.com/mozilla/addons-server/pull/19728/commits/a9a7883bbaed4f0a26b2954cae6d94c85a333341

wagnerand: Yes, let's omit here. Putting it back into `AWAITING_REVIEW` is usually followed by another action (primarily _Approve_), so there isn't much of a need to subscribe reviewers automatically.

src/olympia/reviewers/utils.py
```diff
@@ -1360,6 +1373,55 @@ def reject_multiple_versions(self):
...
+                    self.addon.status,
+                    amo.STATUS_NOMINATED,
+                    'unrejecting versions',
+                )
+            )
+            self.addon.update(status=amo.STATUS_NOMINATED)
+            ActivityLog.create(
+                amo.LOG.CHANGE_STATUS, self.addon, self.addon.status, user=self.user
+            )
+            self.addon.update_version()
```
diox, line 1423: I don't love doing all this manually... Why `STATUS_NOMINATED` in particular and not `STATUS_APPROVED` ? Edge cases around add-on not actually ready to be public ?

eviljeff: We don't have a method in the model that supports status being recalculated "up" currently (from `STATUS_NULL` to a higher status).  We could move this functionality into a new method on `Addon` if you prefer?  (it'd still be the same "manual" code though).

> Why STATUS_NOMINATED in particular and not STATUS_APPROVED ? Edge cases around add-on not actually ready to be public ?

The versions are changed to `STATUS_AWAITING_REVIEW`, so `STATUS_NOMINATED` is the correct status when there is a listed version with that status but no approved versions. (If you're asking why we aren't changing the versions to `STATUS_APPROVED`: the ask, as I understood it, was for an action to change the versions back to a pending state so they could then be manually approved by the reviewer - otherwise the request would have been to have allowed the existing approve action to be applied to `STATUS_DISABLED` versions instead)

diox: Do you think we could roll that into `update_status()` ? Wondering if there is any reason to keep updating the add-on status manually in reviewer tools/devhub/API these days. It should be possible to determine it purely by looking at the attached versions right ?

eviljeff: I think the reason historically was were were afraid of weird file statuses coming from historic bad code, changes to what statuses mean, or just incomplete manual admin changes, so we didn't want an add-on that was status incomplete to "magically" change to "nominated" (or worse - change to "approved").  

Even today I'm cautious of doing it automatically in `update_status` - an add-on is only set to `STATUS_NOMINATED` in devhub and the api if all it's metadata is set - it's not just based on the file status (in the reviewer tools we're making assumptions that the action would only be used on add-ons that were previously okay to be in the `STATUS_NOMINATED` state and/or the reviewer knows what they're doing).  And we have a very limited number of places where an add-on could possibly change status to `STATUS_NOMINATED` (rather than every single time we call `update_status` - a lot)

That said, we could.  Options:
- another function like `update_status` that just handles this case
- an optional kwarg to `update_status` that unlocks an "upgrade" from `STATUS_NULL` to `STATUS_NOMINATED` if appropriate.
- no kwarg but an additional check on the metadata being valid too.

diox: I think even in reviewer tools, it shouldn't be possible to update the status to `STATUS_NOMINATED` (or `STATUS_APPROVED`...) when required metadata is absent. We should fail loudly instead... So I would vote for no kwarg...

But it should probably be done separately, as changing `update_status()` has the potential to reveal more edge cases unrelated to this feature. So in the meantime, probably just extract that code to some method that we would eventually refactor away, and file a follow-up issue to clean everything up ?

eviljeff: https://github.com/mozilla/addons-server/pull/19728/commits/4b8bfdc64ee95100d6e5865033f18492adf84066 moves the nominated status setting up to `Addon` model.  I'll file an issue to refactor it into `update_status` (in some way)

diox (approved):

src/olympia/reviewers/utils.py
eviljeff: https://github.com/mozilla/addons-server/issues/19773

PR #19731, opened 2022-09-23
ManifestStaticFilesStorage with sourceMappingURL

### LINKED ISSUE #19607: collectstatic chokes when linked map files aren't in node_libs in django4.x
part of https://github.com/mozilla/addons-server/issues/19595 to support newer 4.x django - collectstatic now looks for linked files in js and css files, and (at least some of) the js dependencies include links to the xxx.js.map file.  

We don't currently copy those files over and there is no obvious command line switch to disable the feature.  (I tried `--ignore '*.map'` but didn't seem to make a difference)

eviljeff:
fixes #19607 by effectively changing the patterns back to django3.2's regexs and ignoring linked map files.

(I looked into copying the map files but they need to be in certain locations and it just didn't seem worth it)

### DIFF
--- a/src/olympia/lib/storage.py
+++ b/src/olympia/lib/storage.py
@@ -6,11 +6,14 @@ class ManifestStaticFilesStorageNotMaps(ManifestStaticFilesStorage):
         (
             '*.css',
             (
+                # These regexs are copied from HashedFilesMixin in django4.1+
                 r"""(?P<matched>url\(['"]{0,1}\s*(?P<url>.*?)["']{0,1}\))""",
                 (
                     r"""(?P<matched>@import\s*["']\s*(?P<url>.*?)["'])""",
                     """@import url("%(url)s")""",
                 ),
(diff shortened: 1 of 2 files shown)


bobsilverberg (approved):

src/olympia/lib/storage.py
```diff
@@ -0,0 +1,16 @@
+from django.contrib.staticfiles.storage import ManifestStaticFilesStorage
+
+
+class ManifestStaticFilesStorageNotMaps(ManifestStaticFilesStorage):
+    patterns = (
+        (
+            '*.css',
+            (
+                r"""(?P<matched>url\(['"]{0,1}\s*(?P<url>.*?)["']{0,1}\))""",
```
diox, line 9: We should document that this is copied over from Django 3.2 and why we're doing it.

eviljeff: actually from Django4.1 - but compatible with django3.2 as the only changes are to name the captured groups.

PR #19745, opened 2022-09-26
Migrate user photos to new directory structure, removing orphans along the way

### LINKED ISSUE #19709: Migrate user photos to new directory structure and delete orphaned files from EFS
Follow-up to https://github.com/mozilla/addons-server/issues/19690 : we likely have some orphaned photos on the filesystem we can delete.

diox:
Also move `id_to_path()` to `amo.utils` to avoid circular import when using it from `users/models.py` (it makes sense anyway since it's a generic util).

Fixes #19709

### DIFF
--- a/src/olympia/users/management/commands/migrate_user_photos.py
+++ b/src/olympia/users/management/commands/migrate_user_photos.py
@@ -1,0 +1,53 @@
+import os
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+import olympia.core.logger
+from olympia.users.models import UserProfile
+
+
+log = olympia.core.logger.getLogger('z.users.migrate_user_photos')
+
+
+class Command(BaseCommand):
+    help = 'Migrate user photos to new directory structure and remove orphaned ones'
+
+    # Note: because avatars are only displayed on developer profiles and
+    # heavily cached, we're directly migrating in-place without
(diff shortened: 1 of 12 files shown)


src/olympia/users/management/commands/migrate_user_photos.py
```diff
@@ -0,0 +1,53 @@
...
+
+log = olympia.core.logger.getLogger('z.users.migrate_user_photos')
+
+
+class Command(BaseCommand):
+    help = 'Migrate user photos to new directory structure and remove orphaned ones'
+
+    # Note: because avatars are only displayed on developer profiles and
+    # heavily cached, we're directly migrating in-place without
+    # backwards-compatibility - UserProfile.picture_dir has already been
```
bobsilverberg, line 18: I think there is a word missing from this line of the comment. Should it say "has already been *moved*"?

bobsilverberg (approved):
Looks good, just one comment about a comment. ;-) 

r+wc

src/olympia/users/management/commands/migrate_user_photos.py
diox: indeed (well, technically it shouldn't say "moved", because we're talking about the implementation here), reworded in 2922dff8e021cd7889594804e34632ab5ca01871

Issue #19771, opened 2022-10-10
Remove site permissions code from AMO


diox:
All code related to site permission generation/handling should be removed from AMO. It was never fully enabled in prod, and Firefox is going to generate those add-ons differently (see https://bugzilla.mozilla.org/show_bug.cgi?id=1790620 and related bugs).

diox:
The entire implementation will be removed from Firefox soon in https://bugzilla.mozilla.org/show_bug.cgi?id=1789718, and doesn't do much anyway, so I think it's fine. I'll file a linter issue to prevent uploads just in case.

PR #19772, opened 2022-10-11
Remove all site permissions code and hard-delete corresponding add-ons

diox:
Fixes https://github.com/mozilla/addons-server/issues/19771


src/olympia/addons/migrations/0040_auto_20221011_1104.py
```diff
@@ -0,0 +1,20 @@
...
+
+from django.db import migrations
+from olympia.constants.base import _ADDON_SITE_PERMISSION
+
+
+def delete_site_permissions(apps, schema_editor):
+    Addon = apps.get_model('addons', 'Addon')
+
+    # Hard delete the now obsolete site permission add-ons.
+    Addon.unfiltered.filter(type=_ADDON_SITE_PERMISSION).delete()
```
eviljeff, line 11: Aren't there files, etc, that need to be tidied up too?

diox: Yeah... I'm not too worried about a few < 400 bytes files though. I figured the main issue was ensuring the database didn't contain anything the code couldn't handle properly anymore...

eviljeff: Is there a plan for tidying up those files manually somehow, or are we okay with ignoring them?  I would have thought it'd be considerably harder to identify which are the obsolete files once we've nuked all the database references to them.

diox: We've had obsolete files on the filesystem before, I think it's fine to ignore them since they are so small and there are so few of them. We could technically identify them from their name (`site_permissions_for_*`) if we really need to find them and remove them later.

eviljeff (approved):

PR #19785, opened 2022-10-12
update outgoing server to GCP

### LINKED ISSUE #19733: Update REDIRECT_URL default to point to GCP
Once #19636 is done we should update the `REDIRECT_URL` setting default values to point to the new GCP URL. This should be a completely transparent change as we're already overriding it through environment variables everywhere.

eviljeff:
fixes #19733 + drops the per env settings because the servers are defined in .env files so it's redundant.

### DIFF
--- a/src/olympia/search/tests/test_search_ranking.py
+++ b/src/olympia/search/tests/test_search_ranking.py
@@ -725,7 +725,7 @@ class TestRankingScenarios(ESTestCase):
             'Open Image in New Tab',
             (
                 ['Open Image in New Tab', 5553],
-                ['Open image in a new tab', 1735],
+                ['Open image in a new tab', 1736],
             ),
         )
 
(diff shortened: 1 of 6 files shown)


src/olympia/search/tests/test_search_ranking.py
```diff
@@ -725,7 +725,7 @@ def test_scenario_open_image_new_tab(self):
             'Open Image in New Tab',
             (
                 ['Open Image in New Tab', 5553],
-                ['Open image in a new tab', 1735],
+                ['Open image in a new tab', 1736],
```
bobsilverberg, line 728: Why did this PR require these changes?

bobsilverberg (approved):
Looks good, I just have one question.

r+wc

src/olympia/search/tests/test_search_ranking.py
eviljeff: it's the magic of ElasticSearch... the match score is calculated based on all the text content and that includes fields that contain an external url, so have outgoing server url applied to them.  As the outgoing server has changed the text is different so the score can be different.  A small numeric difference (that doesn't change the search ordering) is fine.

PR #19786, opened 2022-10-12
Convert IP addresses to strings in MySQL before calling `GROUP_CONCAT()`

### LINKED ISSUE #19781: IP search chokes on some IP addresses in results
In the user admin IP search, some user known IP addresses can cause a 500 error.

https://sentry.io/organizations/mozilla/issues/3664586752/events/9caea18145cc4ca6b4f9703f2bc86b48/?project=6310819

This is because when the IPs are coming from the database in binary, the are grouped together with `GROUP_CONCAT` and then split by the python code, but if the binary string happened to contain a `,` character (which will be the case if a `44` is part of an IP address returned), we try to split on that, resulting in an invalid IP being parsed by the code to display it.

We either need to find a safe separator or find a way to keep the binary IP in hexadecimal form (or even pre-convert it to the IP in string format) until the last minute to avoid this.

diox:
This avoids issues with commas inside the bytestring representation of the binary value, which can happen for some IPs if they contain a `44`.

Fixes #19781

### DIFF
--- a/src/olympia/users/admin.py
+++ b/src/olympia/users/admin.py
@@ -6,7 +6,6 @@ from django import http
 from django.contrib import admin, messages
 from django.contrib.admin.utils import unquote
 from django.db.models import (
-    CharField,
     Count,
     F,
     FilteredRelation,
--- a/src/olympia/users/admin.py
+++ b/src/olympia/users/admin.py
@@ -37,7 +37,7 @@ from olympia.amo.admin import (
     SEARCH_VAR,
 )
 from olympia.amo.fields import IPAddressBinaryField
-from olympia.amo.models import GroupConcat
+from olympia.amo.models import GroupConcat, Inet6Ntoa
 from olympia.api.models import APIKey, APIKeyConfirmation
 from olympia.bandwagon.models import Collection
 from olympia.ratings.models import Rating
@@ -242,7 +242,7 @@ class UserAdmin(CommaSearchInAdminMixin, admin.ModelAdmin):
             # FilteredRelation we can force it...
             annotations = {
                 'activity_ips': GroupConcat(
-                    'activitylog__iplog__ip_address_binary',
+                    Inet6Ntoa('activitylog__iplog__ip_address_binary'),
                     distinct=True,
                     # You can't have multiple ip addresses in an IPv[4|6]Address object
                     # so we store the binary values in a CharField instead.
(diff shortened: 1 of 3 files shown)


wagnerand:
I'd like to nominate this patch for cherry-picking once approved. It looks encapsulated to admin tools and low risk?

src/olympia/users/admin.py
```diff
@@ -242,7 +242,7 @@ def get_search_results(self, request, queryset, search_term):
             # FilteredRelation we can force it...
             annotations = {
                 'activity_ips': GroupConcat(
-                    'activitylog__iplog__ip_address_binary',
+                    Inet6Ntoa('activitylog__iplog__ip_address_binary'),
                     distinct=True,
                     # You can't have multiple ip addresses in an IPv[4|6]Address object
                     # so we store the binary values in a CharField instead.
```
eviljeff, line 248: this comment needs updating because we're no longer storing binary values.  (the `output_field` kwarg is potentially redundant too if `Inet6Ntoa` specifies an `output_field` but dunno)

src/olympia/users/admin.py
```diff
@@ -482,11 +482,12 @@ def known_ip_adresses(self, obj):
         # extra queries for each row of results), otherwise, look everywhere
         # we can.
         activity_ips = getattr(obj, 'activity_ips', None)
-        to_ipaddress = IPAddressBinaryField().to_python
         if activity_ips is not None:
-            # The GroupConcat value is a comma seperated string of the binary values.
-            ip_addresses = {to_ipaddress(ip) for ip in activity_ips.split(b',')}
+            # The GroupConcat value is a comma seperated string of the ip
+            # addresses (already converted to string thanks to INET6_NTOA).
+            ip_addresses = set(activity_ips.split(','))
```
eviljeff, line 488: We still want the set to be `IPv4Address`|`IPv6Address` objects, so it's the same (apart from splitting on `','` instead)
```suggestion
            ip_addresses = {to_ipaddress(ip) for ip in activity_ips.split(',')}
```

diox: Do we ? AFAICT it's just used to format the string which is then returned and displayed in the results.

eviljeff (approved):

src/olympia/users/admin.py
diox: Ah, because `IPLog` will return such objects, so we need the `set` contents to be consistent...

PR #19800, opened 2022-10-18
Remove M2M between ScannerQueryResult and ScannerQueryRule, use FK instead

### LINKED ISSUE #19797: Code search returns wrong data if it's the last rule
STR:
- Create and execute a ScannerQueryRule matching a bunch of add-ons
- Delete all ScannerQueryRules
- Create and execute a ScannerQueryRule matching any number of add-ons (can be 0 to make it easier to reproduce)
- Visit that rule page and click on the link leading to its results

Expected results:
- You see the results for that rule (nothing if the rule matched 0 add-ons)

Actual results:
- You see a bunch of results with no matching rule (and the page can take a while to load, too!), pagination count doesn't match what you'd expect

This is because of the combination of 2 things:
- We have a M2M relationship between `ScannerQueryRule` and `ScannerQueryResults`. When a rule is deleted, the intermediary model instances between the 2 are deleted, but the `ScannerQueryResults` remain, now orphaned
- When Django detects that a filter has only one choice, it skips it (both in the UI and in the query)
 
So when you're down to the last `ScannerQueryRule` because all other rules have been deleted, that filter is skipped even if the corresponding parameter was present in the query string, and because we have some orphaned `ScannerQueryResults`, they are displayed...

We should remove the M2M and see if we can force the application of the filter at all times (that last part may be tricky...)

diox:
For queries, there should be one rule per result only. This allows deletion of query rules to properly delete query results, and in turn avoids the issue where orphaned results would stay around and mess up filtering when only one rule was remaining.

Because it's in the way, also move things specific to `ScannerResult` and `ScannerRule` that `ScannerQueryResult` / `ScannerQueryRule` are not using outside of the common base classes.

Fixes #19797


src/olympia/scanners/models.py
```diff
@@ -402,17 +418,25 @@ def completion_rate(self):
...
 class ScannerQueryResult(AbstractScannerResult):
-    # Has to be overridden, because the parent refers to ScannerMatch.
-    matched_rules = models.ManyToManyField(
-        'ScannerQueryRule', through='ScannerQueryMatch', related_name='results'
+    # Note: ScannerResult uses a M2M called 'matched_rules', but here we don't
+    # need a M2M because there will always be a single rule for each result, so
+    # we have a single FK called 'matched_rule' (singular).
+    matched_rule = models.ForeignKey(
+        ScannerQueryRule, on_delete=models.CASCADE, related_name='results'
     )
```
diox, line 426: This is the key change - there is a lot of code being moved around from the abstract classes to the more specific classes in this PR, but this change is the most important part, changing the M2M to a FK.

src/olympia/scanners/migrations/0050_auto_20221018_1106.py
```diff
@@ -0,0 +1,21 @@
...
+    operations = [
+        migrations.RemoveField(
+            model_name='scannerqueryrule',
+            name='action',
+        ),
+        migrations.RemoveField(
+            model_name='scannerqueryrule',
+            name='is_active',
+        ),
+    ]
```
eviljeff, line 21: Not that it matters much, but do we want to combine these migrations into a single file?  0049 and 0050 look pretty innocuous at least.

src/olympia/scanners/models.py
```diff
@@ -324,6 +231,115 @@ class Meta(AbstractScannerResult.Meta):
...
+        super().save(*args, **kwargs)
+        # ...then add the associated rules.
+        for scanner_rule in matched_rules:
+            self.matched_rules.add(scanner_rule)
+
+    def can_report_feedback(self):
+        return self.state == UNKNOWN and self.scanner not in [MAD]
+
+    def can_revert_feedback(self):
+        return self.state != UNKNOWN and self.scanner not in [MAD]
```
eviljeff, line 260: these `not in [MAD]` statements are weird

src/olympia/scanners/models.py
```diff
@@ -402,17 +418,25 @@ def completion_rate(self):
...
+    @classproperty
+    def rule_model(cls):
+        return cls.matched_rule.field.related_model
 
-class ScannerQueryMatch(ModelBase):
-    result = models.ForeignKey(ScannerQueryResult, on_delete=models.CASCADE)
-    rule = models.ForeignKey(ScannerQueryRule, on_delete=models.CASCADE)
+    def save(self, *args, **kwargs):
+        if self.results:
+            self.matched_rule = self.get_rules_queryset().get()
```
eviljeff, line 441: So, this is never going to be a queryset of length `<1` or `>1`? No need for defensive code?  e.g. `.first()`?

src/olympia/scanners/migrations/0050_auto_20221018_1106.py
diox: Yeah I kinda changed my mind on this as I was writing the patch. Started with bundling a bunch of things into migration 48, then did it piece by piece. The (admittedly weak) logic behind keeping 49 and 50 separate is that they are for different models.

src/olympia/scanners/models.py
diox: Similarly, if no rule is found, something is super weird and we shouldn't have created the result in the first place - the FK doesn't allow `null`, since I'd also want to fail loudly rather than let this happen.

src/olympia/scanners/migrations/0050_auto_20221018_1106.py
diox: https://github.com/mozilla/addons-server/pull/19800/commits/d7fdcf4c2baebbb3576eb26a857b7c05a17c4acb merged 49 and 50

PR #19811, opened 2022-10-20
refactor version_dict usage; update blocklist tests

### LINKED ISSUE #19809: Make sure AMO accepts version strings that follow the "new format" for both MV2 and MV3
Once https://github.com/mozilla/addons-linter/issues/4507 has been merged, the linter will start to emit warnings to encourage users to adopt a "simpler version format" (documented on EW later today ([see -stage](https://extensionworkshop.allizom.org/documentation/develop/manifest-v3-migration-guide/))).

The linter will only emit errors for MV3 extensions but that isn't really a problem at the moment because AMO does not accept MV3 submissions.

With the linter PR linked above, there is a small but important change for MV2 extensions: previously, when an extension was not using a toolkit version (e.g. any version string with some letters), each "part" (or number) of the version string had to be between 0 and 65535. That is going to change for consistency purposes. MV2 extensions will be able to use numbers with up to 9 digits (the other rules about the number of parts and leading zeros were already enforced before so that will not change).

This issue is about making sure that AMO will accept MV2 extensions with a version string like `123456789.0` or `999999999.999999999.999999999.999999999` for instance.

eviljeff:
closes #19809 - there weren't any fixes needed, but I updated the blocklist tests to prove it, and tidied up / simplified some usage.

### DIFF
--- a/src/olympia/blocklist/tests/test_models.py
+++ b/src/olympia/blocklist/tests/test_models.py
@@ -2,7 +2,6 @@ from django.core.exceptions import ValidationError
 from django.test.utils import override_settings
 
 from olympia.amo.tests import addon_factory, TestCase, user_factory
-from olympia.versions.compare import MAX_VERSION_PART
 
 from ..models import Block, BlocklistSubmission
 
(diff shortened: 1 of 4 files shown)


diox:
> there weren't any fixes needed

wDOEbo0

PR #19820, opened 2022-10-25
enforce a higher than current version number for new listed versions

### LINKED ISSUE #19788: For listed versions, do not allow submissions with a version number lower than the current public version

### What happened?

eviljeff:
fixes #19788 - the fun of having three different ways to create a new version 😬


src/olympia/addons/serializers.py
```diff
@@ -513,24 +514,41 @@ def _check_for_existing_versions(self, version_string):
...
-                    and self.addon.disabled_by_user
-                ):
-                    raise exceptions.ValidationError(
-                        gettext(
-                            'Listed versions cannot be submitted while add-on is '
-                            'disabled.'
+                self._check_for_existing_versions(version_string)
+
+                if data['upload'].channel == amo.CHANNEL_LISTED:
+                    self._check_is_greater_version_number(version_string)
```
bobsilverberg, line 543: Why are we limiting it to listed versions? Wouldn't this be a valuable check for unlisted versions too? Or do we just not care about those?

src/olympia/devhub/forms.py
```diff
@@ -1062,31 +1062,40 @@ def check_blocklist(self, guid, version_string):
...
-                else:
-                    msg = gettext('Version {version} already exists.')
-                raise forms.ValidationError(msg.format(version=version_string))
+                )
+            else:
+                msg = gettext('Version {version} already exists.')
+            raise forms.ValidationError(msg.format(version=version))
+
+    def check_for_greater_version_number(self, version_string):
+        if (
```
bobsilverberg, line 1091: I know it's just a couple of lines of code, but this is the same check that was added into `src/olympia/addons/serializers.py`, so should we maybe have a shared helper for the logic?

src/olympia/signing/views.py
```diff
@@ -235,6 +235,23 @@ def handle_upload(self, request, addon, version_string, guid=None):
                     ),
                     status.HTTP_400_BAD_REQUEST,
                 )
+        if (
+            addon
+            and channel == amo.CHANNEL_LISTED
+            and (previous_version := addon.current_version)
```
bobsilverberg, line 241: Same comment here about the logic being duplicated.

bobsilverberg (approved):
Looks good, I'm just wondering whether we should have a shared helper for the `is_greater_version_number` logic, as it appears in 3 places.

r+wc

src/olympia/addons/serializers.py
eviljeff: the issue requested just listed versions have the restriction applied.

src/olympia/devhub/forms.py
eviljeff: I thought the same when I was writing it.  The logic is literally just the one line: `previous_version.version >= version_string` (and then we throw a slightly different exception in each place) so I wasn't sure it would make a difference.  I'll try a helper function and see how that looks.

bobsilverberg: Yeah, it just being a single line of code it does seem like it could be overkill, but because it's business logic it's nice to have it in once place if for some reason in the future the logic needs to change.

eviljeff: https://github.com/mozilla/addons-server/pull/19820/commits/d29edd74bf304c0163fd4fbc1ca7125b0d6e2997  not bad

PR #19827, opened 2022-10-28
Bump max result window to 30000

### LINKED ISSUE #19794: Bump search max result window (to 30000
For performance/scalability reasons, we set max_result_window in elasticsearch to 25000 - if there more results for a given query, we don't allow returning them.

25000 was chosen back in the day because it was above the total number of public extensions available, but we're now past that number. Let's try bumping it to 30000, hopefully that should still work and allow us to see all public extensions through the API.

The setting is `ES_MAX_RESULT_WINDOW`.

diox:
Fixes #19794

### DIFF
--- a/src/olympia/api/tests/test_pagination.py
+++ b/src/olympia/api/tests/test_pagination.py
@@ -74,7 +74,7 @@ class TestESPageNumberPagination(TestCustomPageNumberPagination):
         response = view(request)
         assert response.data == {
             'page_size': page_size,
-            'page_count': int(ceil(settings.ES_MAX_RESULT_WINDOW / page_size)),
+            'page_count': page + 1,  # We know there should be one more.
             'results': mock.ANY,
             'previous': f'http://testserver/?page={page - 1}&page_size={page_size}',
             'next': f'http://testserver/?page={page + 1}&page_size={page_size}',
@@ -88,7 +88,7 @@ class TestESPageNumberPagination(TestCustomPageNumberPagination):
         response = view(request)
         assert response.data == {
             'page_size': page_size,
-            'page_count': int(ceil(settings.ES_MAX_RESULT_WINDOW / page_size)),
+            'page_count': page,  # We know it should be the last one.
             'results': mock.ANY,
             'previous': f'http://testserver/?page={page - 1}&page_size={page_size}',
             'next': None,
--- a/src/olympia/api/tests/test_pagination.py
+++ b/src/olympia/api/tests/test_pagination.py
@@ -1,4 +1,3 @@
-from math import ceil
 from unittest import mock
 
 from django.conf import settings
--- a/src/olympia/api/tests/test_pagination.py
+++ b/src/olympia/api/tests/test_pagination.py
@@ -1,4 +1,7 @@
+from math import ceil
 from unittest import mock
+
+from django.conf import settings
 
 from rest_framework import generics, serializers, status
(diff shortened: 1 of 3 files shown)


src/olympia/api/tests/test_pagination.py
```diff
@@ -54,36 +57,43 @@ def test_metadata_with_default_page_size(self):
...
+        # Request the last page that still has a `next`.
+        page_size = 5
+        page = int(settings.ES_MAX_RESULT_WINDOW / page_size) - 1
+        request = self.factory.get('/', {'page_size': page_size, 'page': page})
         response = view(request)
         assert response.data == {
-            'page_size': 5,
-            'page_count': 5000,
+            'page_size': page_size,
+            'page_count': int(ceil(settings.ES_MAX_RESULT_WINDOW / page_size)),
```
eviljeff, line 77: the discrepancy here makes me nervous - there's an extra `ceil` - shouldn't it be just `page + 1`?

src/olympia/api/tests/test_pagination.py
```diff
@@ -54,36 +57,43 @@ def test_metadata_with_default_page_size(self):
...
-            'count': 30000,
+            'previous': f'http://testserver/?page={page - 1}&page_size={page_size}',
+            'next': f'http://testserver/?page={page + 1}&page_size={page_size}',
+            'count': total,
         }
 
-        request = self.factory.get('/', {'page_size': 5, 'page': 5000})
+        # Request the page that doesn't have a `next` because it's over the
+        # max result window.
+        page = int(settings.ES_MAX_RESULT_WINDOW / page_size)
```
eviljeff, line 86:
```suggestion
        page = page + 1
```
?

src/olympia/api/tests/test_pagination.py
```diff
@@ -54,36 +57,43 @@ def test_metadata_with_default_page_size(self):
...
+        request = self.factory.get('/', {'page_size': page_size, 'page': page})
         response = view(request)
         assert response.data == {
-            'page_size': 5,
-            'page_count': 5000,
+            'page_size': page_size,
+            'page_count': int(ceil(settings.ES_MAX_RESULT_WINDOW / page_size)),
```
eviljeff, line 91: this should be just `'page_count': page`?

src/olympia/api/tests/test_pagination.py
eviljeff: Or if there's some case I'm missing where there's a difference, we should have a test for that scenario?

eviljeff (approved):
r+wc just nits with the tests

src/olympia/api/tests/test_pagination.py
diox: It was just to make the test generic and copy/pasteable

src/olympia/api/tests/test_pagination.py
diox: 89ee89001f692455fd71a4c50744afe52be2d1fc

src/olympia/api/tests/test_pagination.py
diox: 89ee89001f692455fd71a4c50744afe52be2d1fc

src/olympia/api/tests/test_pagination.py
diox: committed another simplification in 89ee89001f692455fd71a4c50744afe52be2d1fc

PR #19836, opened 2022-10-31
Add canned_response to ReviewActionReason

### LINKED ISSUE #19835: Store a canned response for each review action reason
This contributes to #19818

When a reviewer chooses a review action reason, a particular canned response will be added to the comments input for the review. We should therefore update the `ReviewActionReason` model to also store a canned response text, so that it will be available for use in the UI.

bobsilverberg:
Fixes #19835


wagnerand:
The new field would also have to be exposed in django admin so I can add canned responses.

bobsilverberg:
> The new field would also have to be exposed in django admin so I can add canned responses.

This is done automatically. I didn't include the canned response field in the listing page in the admin, as it will be long and I'm not sure that adds any value. @wagnerand Would you like to see the canned response on the listing page, or is it sufficient to be able to see, and edit, it on the detail page?

diox (approved):

PR #19853, opened 2022-11-02
increase check_xpi_info check for version string field length to 255

### LINKED ISSUE #19842: The upload errors for a version with more than 32chars
Followup https://github.com/mozilla/addons-server/issues/19809#issuecomment-1298318644

I re-tested some samples which I used for the linter issue.  

One thing is different on AMO Dev and reproducible for MV2 and MV3: 

when I tried with a longer string such as "123456789.123456789.123456789.123456789" I will get an error 

Version numbers should have fewer than 32 characters.  -> [full validation report](https://addons-dev.allizom.org/en-US/developers/upload/695693cb39bb493fb20731b938843291) would not indicate this error

So it is obvious this comes from the server side -> the linter does not throw this error when I test it locally. 

error

eviljeff:
fixes #19842

### DIFF
--- a/src/olympia/files/utils.py
+++ b/src/olympia/files/utils.py
@@ -874,7 +874,7 @@ def check_xpi_info(xpi_info, addon=None, xpi_file=None, user=None):
     if len(xpi_info['version']) > version_field_max_length:
         raise forms.ValidationError(
             gettext(
-                'Version numbers should have fewer than {max_length} characters.'
+                'Version numbers should have at most {max_length} characters.'
             ).format(max_length=version_field_max_length)
         )
     if not VERSION_RE.match(xpi_info['version']):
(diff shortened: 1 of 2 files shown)


src/olympia/files/utils.py
```diff
@@ -868,9 +869,13 @@ def check_xpi_info(xpi_info, addon=None, xpi_file=None, user=None):
...
             raise forms.ValidationError(gettext('Duplicate add-on ID found.'))
-    if len(xpi_info['version']) > 32:
+
+    version_field_max_length = Version.version.field.max_length
+    if len(xpi_info['version']) > version_field_max_length:
         raise forms.ValidationError(
-            gettext('Version numbers should have fewer than 32 characters.')
+            gettext(
+                'Version numbers should have fewer than {max_length} characters.'
+            ).format(max_length=version_field_max_length)
```
diox, line 878: Technically we allow up to `max_length`, so should add a `+1` here when formatting the message or tweak it.

diox (approved):

src/olympia/files/utils.py
eviljeff: So the message was wrong for 12 years and no-one noticed...

Issue #19776, opened 2022-10-11
Remove `_FileNameFieldFile` and `FilenameFileField`


diox:
We added  `_FileNameFieldFile` and `FilenameFileField` as part of the migration away from `File.filename` in https://github.com/mozilla/addons-server/issues/19207 but now that the migration is complete and we no longer have old paths in the database we should use a regular django `FileField` instead and remove that code.

PR #19856, opened 2022-11-04
Set path to files missing on filesystem to '', skip them in code search/downloads

### LINKED ISSUE #19815: Scanners should ignore files that don't have a `filename`.
A few files are missing on disk (see also #17885), and code-search keeps tripping over them. Let's clear the `filename` field for those, and make code-search ignore `File`s with an empty `filename`.

SQL query to find them: `SELECT id, filename, created FROM files WHERE filename NOT LIKE '%/%';`.

diox:
Fixes #19815
Fixes https://github.com/mozilla/addons-server/issues/19776


src/olympia/scanners/tasks.py
```diff
@@ -291,7 +291,9 @@ def run_yara_query_rule(query_rule_pk):
         return
     log.info('Fetching versions for run_yara_query_rule on rule %s', rule.pk)
     # Build a huge list of all pks we're going to run the tasks on.
-    qs = Version.unfiltered.filter(addon__type=amo.ADDON_EXTENSION)
+    qs = Version.unfiltered.filter(
+        addon__type=amo.ADDON_EXTENSION, file__file__isnull=False
```
eviljeff, line 295: is `null` a possible value for the `File.file` field?  (the field definition doesn't include `null=True`)

eviljeff: Should it be `file__isnull=False` to filter out missing files instead?

eviljeff (approved):

src/olympia/scanners/tasks.py
diox: It's not a possible value no. I'll fix it.

diox: f141214c4108ee0fe0f23cb82d05eaaac0764841

PR #19858, opened 2022-11-04
Enhance VersionString to support full toolkit version strings

### LINKED ISSUE #19788: For listed versions, do not allow submissions with a version number lower than the current public version

eviljeff:
fixes #19788 

references were:
https://github.com/mozilla/addons-moz-compare/blob/main/src/index.js and then original
https://searchfox.org/mozilla-central/source/xpcom/base/nsVersionComparator.cpp but the way the comparisons are done is different because `VersionString` builds up a tuple of `Part` objects representing each part first, and delegates the part comparisons to that class instead.  `parseVersionPart` is the key function in addons-moz-compare, and it's implemented in `Part.__init__` here.

### DIFF
--- a/src/olympia/versions/compare.py
+++ b/src/olympia/versions/compare.py
@@ -1,5 +1,7 @@
 import re
 
+from django.utils.functional import cached_property
+
 
 BIGINT_POSITIVE_MAX = 2**63 - 1
 APP_MAJOR_VERSION_PART_MAX = 2**16 - 1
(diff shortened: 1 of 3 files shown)


eviljeff:
it may be easier to review compare.py as a whole file, rather than via the diff, because `VersionString` is completely different.

(`version_int` should be functionally the same, but I had to move some of the old code from `VersionString` into it)

bobsilverberg (approved):

PR #19859, opened 2022-11-04
drop redundant string limit in VERSION_RE

### LINKED ISSUE #19842: The upload errors for a version with more than 32chars

eviljeff:
fixes #19842 again 🙃

The length limit in the regex was redundant as we only ever use it just after a string length check that would raise if the string length exceeded 255 (and the linter before that would return an error for a string length above 100)

### DIFF
--- a/src/olympia/files/utils.py
+++ b/src/olympia/files/utils.py
@@ -43,7 +43,7 @@ class ParseError(forms.ValidationError):
     pass
 
 
-VERSION_RE = re.compile(r'^[-+*.\w]{,32}$')
+VERSION_RE = re.compile(r'^[-+*.\w]*$', re.ASCII)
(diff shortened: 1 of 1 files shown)


diox (approved):
Maybe add a quick comment mentioning that we [should always] check the length when we use this regexp

PR #19871, opened 2022-11-07
Replace memcached caching of pages by HTTP caching with the CDN/clients

### LINKED ISSUE #18083: Remove page caching that uses memcached
We cache 2 views using `@cache_page` (collections add-ons list and language tools APIs), which uses the default cache backend (memcached in our case). With the CDN and nginx caching that shouldn't be necessary anymore, just returning `Cache-Control` header should be enough.

diox:
Fixes #18083

### DIFF
--- a/src/olympia/addons/views.py
+++ b/src/olympia/addons/views.py
@@ -5,8 +5,6 @@ from django.db.models import Max, Prefetch
 from django.db.transaction import non_atomic_requests
 from django.shortcuts import redirect
 from django.utils.cache import patch_cache_control
-from django.utils.decorators import method_decorator
-from django.views.decorators.cache import cache_page
 
 from elasticsearch_dsl import Q, query, Search
 from rest_framework import exceptions, serializers, status
--- a/src/olympia/addons/views.py
+++ b/src/olympia/addons/views.py
@@ -1127,17 +1127,16 @@ class LanguageToolsView(ListAPIView):
             .distinct()
         )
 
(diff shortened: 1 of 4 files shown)


src/olympia/addons/views.py
```diff
@@ -1127,17 +1125,16 @@ def get_language_packs_queryset_with_appversions(self, application, appversions)
...
         # Ignore pagination (return everything) but do wrap the data in a
         # 'results' property to mimic what the default implementation of list()
         # does in DRF.
         queryset = self.filter_queryset(self.get_queryset())
         serializer = self.get_serializer(queryset, many=True)
-        return Response({'results': serializer.data})
+        response = Response({'results': serializer.data})
+        # Cache on the CDN/clients for 24 hours.
+        patch_cache_control(response, max_age=60 * 60 * 24)
+        return response
```
eviljeff, line 1137: kind of a nit, but it's irritating (to me) that we always have to do this in 3 separate statements - if only `patch_cache_control` returned the `response` it could be written as a single statement 🤷

eviljeff (approved):

src/olympia/addons/views.py
diox: Yeah... the better path would be to use the `cache_control` decorator, but any time that header needs a dynamic value or be added or not depending on something else, it doesn't work... I'm not a big fan of `patch_cache_control` either but also not really a fan or writing our own helper just for 3 lines of code.

PR #19888, opened 2022-11-09
sanitize sitemap parameters

### LINKED ISSUE #19887: santize sitemap.xml parameters
Even though any invalid values for `app`, `section` or `page` would just result in a 404 when nginx tries to serve the file, we should sanitize parameters beforehand to raise for any known invalid values.

eviljeff:
fixes #19887

### DIFF
--- a/src/olympia/amo/sitemap.py
+++ b/src/olympia/amo/sitemap.py
@@ -7,6 +7,7 @@ from urllib.parse import urlparse
 
 from django.conf import settings
 from django.contrib.sitemaps import Sitemap as DjangoSitemap
+from django.core.paginator import PageNotAnInteger
 from django.db.models import Count, Max, Q
 from django.template import loader
 from django.utils.functional import cached_property
@@ -42,6 +43,10 @@ FRONTEND_LANGUAGES = [
     'ru',
     'zh-CN',
 ]
+
+
+class InvalidSection(Exception):
+    pass
 
 
 class LazyTupleList:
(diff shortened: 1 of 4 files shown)


src/olympia/amo/sitemap.py
```diff
@@ -469,16 +474,23 @@ def render_index_xml(sitemaps):
...
-    return os.path.join(
-        settings.SITEMAP_STORAGE_PATH,
-        section or '',
-        app or '',
-        f'{endpath}.xml',
-    )
+        if (section, amo.APPS.get(app, '') if app else app) not in get_sitemaps():
+            raise InvalidSection
+        try:
+            int(page)
```
diox, line 483: I think you want to assign `page` to the result of that `int()` call to ensure we are only dealing with an integer in the code later. `int()` accepts a bunch of things : floats, integer strings prefixed/suffixed by whitespace or `+`/`-` signs, etc (the following strings do not raise `ValueError`: ` 42`, `42 `, `+42`).

Like now, ultimately we would fail the load the page, but we would still try to pass down an invalid path to nginx.

FWIW my suggestion would be consistent with what [django paginator does](https://github.com/django/django/blob/d6fbfea08d9bae4165df6a0cfbd4520b5e2242b2/django/core/paginator.py#L48).

eviljeff: https://github.com/mozilla/addons-server/pull/19888/commits/7d038816b23460486eb110df8131571d09bf1df7

diox (approved):

PR #19889, opened 2022-11-09
enforce a higher than current version number for new listed versions

### LINKED ISSUE #19788: For listed versions, do not allow submissions with a version number lower than the current public version

eviljeff:
fixes #19788 (again...)

This patch is very similar to https://github.com/mozilla/addons-server/pull/19820 that was later reverted.  The differences are:
- added tests for the utility function specifically
- renamed the utility function
- removed an unnecessary cast to `VersionString`


src/olympia/addons/serializers.py
```diff
@@ -516,21 +519,24 @@ def _check_for_existing_versions(self, version_string):
...
-                            'disabled.'
+                self._check_for_existing_versions(version_string)
+
+                if data['upload'].channel == amo.CHANNEL_LISTED:
+                    if error_message := is_version_number_not_greater_than_current(
+                        self.addon, version_string
+                    ):
+                        raise exceptions.ValidationError({'version': error_message})
+                    if self.addon.disabled_by_user:
+                        # Also check for submitting listed versions when disabled.
```
bobsilverberg, line 533: Nit: I think this comment makes sense above line 532 (in a similar spot to where it was before).

bobsilverberg (approved):
Looks good. Just one nit.

r+wc

PR #19894, opened 2022-11-11
replace `|` in reviewer actions lists with space padding

### LINKED ISSUE #19807: The Un-reject version info text is also displayed for the 'Reject Multiple Versions' action
1. Open an extension review page  - see https://reviewers.addons.allizom.org/en-US/reviewers/review/1006140
2. Select the 'Reject Multiple Versions' action
3. Notice the information messages displayed for the selected action

### What happened?
The information message for the Un-reject action is also displayed for the 'Reject Multiple Versions' action:
image

With error messages:
image

Only the message specific to 'Reject Multiple Versions' should be displayed

eviljeff:
fixes #19807 - took a few attempts playing around with prefixing `|` or ` ` and edge cases with empty lists but this works - and the extra spaces read much better than extra `|` characters.

### DIFF
--- a/src/olympia/reviewers/forms.py
+++ b/src/olympia/reviewers/forms.py
@@ -149,8 +149,8 @@ class VersionsChoiceWidget(forms.SelectMultiple):
             # Add our special `data-toggle` class and the right `data-value` depending
             # on status.
             option['attrs']['class'] = 'data-toggle'
-            option['attrs']['data-value'] = ''.join(
-                f' {a} ' for a in self.actions_filters[obj.channel].get(status, ())
+            option['attrs']['data-value'] = ' '.join(
+                self.actions_filters[obj.channel].get(status, ())
             )
         # Just in case, let's now force the label to be a string (it would be
         # converted anyway, but it's probably safer that way).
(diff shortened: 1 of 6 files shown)


src/olympia/reviewers/forms.py
```diff
@@ -149,8 +149,8 @@ def create_option(self, *args, **kwargs):
             # Add our special `data-toggle` class and the right `data-value` depending
             # on status.
             option['attrs']['class'] = 'data-toggle'
-            option['attrs']['data-value'] = '|'.join(
-                self.actions_filters[obj.channel].get(status, ()) + ('',)
+            option['attrs']['data-value'] = ''.join(
+                f' {a} ' for a in self.actions_filters[obj.channel].get(status, ())
```
diox, line 153: As mentioned on matrix, if you use `~=` without adding spaces in the JS (*) then you can drop the extra whitespace before and after each value here and below, joining with a space instead. It would look much better and will be less confusing IMHO.

(*) Assuming jQuery supports `~=`. That needs to be double-checked.

bobsilverberg (approved):
The code changes look fine, but I think Mat made a comment that should be tried/addressed.

eviljeff:
> The code changes look fine, but I think Mat made a comment that should be tried/addressed.

@bobsilverberg The code is pretty different after applying Mat's suggestion - don't know if you want to re-review?

src/olympia/reviewers/forms.py
eviljeff: https://github.com/mozilla/addons-server/pull/19894/commits/8c59f0a4651974249f6224a9af19d960670b4770

PR #19971, opened 2022-11-14
Only force greater version numbers against approved versions.

### LINKED ISSUE #19964: Throw the version error only when the previous submitted version is approved
For https://github.com/mozilla/addons-server/issues/19849#issuecomment-1311858536

The error that a version must be higher -> is displayed even if the submitted version has not been approved yet (it is in awaiting review) 

Expected result:
An error is thrown only if the previous version is approved.

eviljeff:
fixes #19964 (includes yet-another-rename of the utility function)

### DIFF
--- a/src/olympia/addons/utils.py
+++ b/src/olympia/addons/utils.py
@@ -302,14 +302,19 @@ def webext_version_stats(request, source):
     log.info('webext_version_stats no match')
 
 
-def validate_version_number_is_greater(addon, version_string):
+def validate_version_number_is_gt_latest_signed_listed_version(addon, version_string):
     """Returns an error string if `version_string` isn't greater than the current
     approved listed version."""
     if (
         addon
-        and (previous_version := addon.current_version)
-        and previous_version.file.status == amo.STATUS_APPROVED
-        and previous_version.version >= version_string
+        and (
+            latest_version_string := addon.versions(manager='unfiltered_for_relations')
+            .filter(channel=amo.CHANNEL_LISTED, file__is_signed=True)
+            .order_by('created')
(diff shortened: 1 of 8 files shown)


src/olympia/addons/utils.py
```diff
@@ -302,10 +302,13 @@ def webext_version_stats(request, source):
     log.info('webext_version_stats no match')
 
 
-def is_version_number_not_greater_than_current(addon, version_string):
+def validate_version_number_is_greater(addon, version_string):
```
diox, line 305: I like the `validate_` part, but the fact that it doesn't say `greater` than what is weird...

src/olympia/addons/utils.py
```diff
@@ -302,10 +302,13 @@ def webext_version_stats(request, source):
...
+    """Returns an error string if `version_string` isn't greater than the current
+    approved listed version."""
     if (
         addon
         and (previous_version := addon.current_version)
+        and previous_version.file.status == amo.STATUS_APPROVED
```
wagnerand, line 311: We should probably check for signed versions here, instead of approved versions, see also https://github.com/mozilla/addons-server/issues/19964#issuecomment-1313684964

src/olympia/addons/utils.py
eviljeff: https://github.com/mozilla/addons-server/pull/19971/commits/55d76e55d6dc3ab7ed6aceac7f1dec98ef4538ae `validate_version_number_is_gt_latest_signed_listed_version` is the longest that fits

diox (approved):

PR #19989, opened 2022-11-21
Lower minimum version for MV3 to 109.0a1, allowing Nightly

diox:


src/olympia/devhub/file_validation_annotations.py
```diff
@@ -136,7 +136,7 @@ def add_manifest_version_messages(*, results, channel):
                     'compatible with manifest version 2 (MV2) extensions. We recommend '
                     'uploading Manifest V3 extensions as self-hosted for now to not '
                     'break compatibility for your users.'
-                ).format(version=amo.DEFAULT_WEBEXT_MIN_VERSION_MV3_FIREFOX),
+                ).format(version='109.0'),
```
eviljeff, line 139: There's a way to do this programmatically with `version_dict` - something like `mv3_min = version_dict(amo.DEFAULT_WEBEXT_MIN_VERSION_MV3_FIREFOX); f'{mv3_min["major"]}.{mv3_min["minor1"]}'`

eviljeff (approved):
just a suggestion to avoid hardcoding 109.0 somewhere instead of relying on the existing constant

src/olympia/devhub/file_validation_annotations.py
diox: Oh sorry, misread the context. I'll write a follow-up patch for the message, you're right.

diox: https://github.com/mozilla/addons-server/issues/19990

PR #19992, opened 2022-11-21
Avoid sending bounce emails to invalid addresses

### LINKED ISSUE #19892: Don't try bouncing to an invalid email in response to an activity log email
We can sometimes end up trying to send an email to an invalid address when bouncing in response to an activity log email we can't process. We should validate it first.

That bounce attempt is invisible, but it clutters sentry. QA should focus on checking that the fix for this doesn't break sending bounces entirely (*)

(*) Typical bounce email we send:
> 
> Hello,
>
> An email was received, apparently from you. Unfortunately we couldn't process it because of:
> (...)
> Please visit https://addons.mozilla.org to leave a reply instead.
>
> Mozilla Add-ons
> https://addons.mozilla.org

diox:
Fixes #19892

### DIFF
--- a/src/olympia/activity/utils.py
+++ b/src/olympia/activity/utils.py
@@ -5,6 +5,7 @@ from email.utils import formataddr
 from html import unescape
 
 from django.conf import settings
+from django.core.mail.message import sanitize_address
 from django.forms import ValidationError
 from django.template import loader
 from django.urls import reverse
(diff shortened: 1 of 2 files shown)


src/olympia/activity/utils.py
```diff
@@ -441,25 +442,33 @@ def filter_queryset_to_pending_replies(queryset, log_type_ids=NOT_PENDING_IDS):
...
-    recipient = (
+    recipient_header = (
         None
         if not isinstance(message, dict)
         else message.get('From', message.get('ReplyTo'))
     )
-    if not recipient:
-        log.error(
+    if not recipient_header:
+        log.warning(
```
diox, line 451: We don't need a hard error for this either.

bobsilverberg (approved):

PR #20006, opened 2022-11-23
"Upgrade" datetime.strptime()/dateutil.parser.parse() where possible

### LINKED ISSUE #20003: Replace `strptime` with `fromisoformat` where possible
From this [excellent post about profiling with Sentry](https://blog.sentry.io/2022/11/21/profiling-for-python/#working-with-date-times):
```python
>>> from dateutil import parser
>>> from datetime import datetime
>>> from timeit import timeit
>>>
>>> ts = "2022-11-09T16:15:00+00:00"
>>>
>>> number = 1_000_000
>>>
>>> print(timeit("parser.parse(ts)", number=number, globals=globals()))
85.08067758600001
>>>
>>> print(timeit("datetime.strptime(ts, '%Y-%m-%dT%H:%M:%S+00:00')", number=number, globals=globals()))
9.526757356999994
>>>
>>> print(timeit("datetime.fromisoformat(ts)", number=number, globals=globals()))
0.1656539809999913
```

`datetime.fromisoformat()` is ~ 60 times faster than `strptime()` to parse iso 8601 dates. 

On a typical search API response on AMO, we currently need to convert 3 date fields per add-on from a string to a `datetime`, times 25 for each add-on if using the default pagination that's probably one or two milliseconds gained per response. Doesn't look like much but it's free and easy to do.

diox:
- `strptime()` is faster than `dateutile.parser.parse()` to parse a fixed format
- `fromisoformat()` is faster than `strptime()` to parse an iso 8601 date

Fixes #20003

### DIFF
--- a/src/olympia/api/serializers.py
+++ b/src/olympia/api/serializers.py
@@ -63,8 +63,7 @@ class BaseESSerializer(serializers.ModelSerializer):
         # Don't be picky about microseconds here. We get them some time
         # so we have to support them. So let's strip microseconds and handle
         # the datetime in a unified way.
-        value = value.partition('.')[0]
-        return datetime.strptime(value, '%Y-%m-%dT%H:%M:%S')
+        return datetime.fromisoformat(value.partition('.')[0])
 
     def _attach_fields(self, obj, data, field_names):
         """Attach fields to fake instance."""
--- a/src/olympia/amo/tests/__init__.py
+++ b/src/olympia/amo/tests/__init__.py
@@ -30,7 +30,6 @@ from django.utils.encoding import force_str
 from django.utils.html import escape
 
 import pytest
-from dateutil.parser import parse as dateutil_parser
 from rest_framework.reverse import reverse as drf_reverse
 from rest_framework.settings import api_settings
 from rest_framework.test import APIClient, APIRequestFactory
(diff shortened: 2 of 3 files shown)


src/olympia/api/serializers.py
```diff
@@ -63,8 +63,7 @@ def handle_date(self, value):
         # Don't be picky about microseconds here. We get them some time
         # so we have to support them. So let's strip microseconds and handle
         # the datetime in a unified way.
-        value = value.partition('.')[0]
-        return datetime.strptime(value, '%Y-%m-%dT%H:%M:%S')
+        return datetime.fromisoformat(value.partition('.')[0])
```
eviljeff, line 66: That we're stripping off microseconds makes me wonder - _what_ format exactly does ES guarantee to return it's datetimes in?  (And is that format always something that `fromisoformat` will parse?)

eviljeff (approved):

src/olympia/api/serializers.py
diox: The [docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.1/date.html) specify how to deal with the input a lot more than output... it just says
> Dates will always be rendered as strings, even if they were initially supplied as a long in the JSON document.

As far as I can see by making requests, it's always in ISO 8601 with the full time including microseconds... but that precision may be lost by MySQL. Stripping them is an attempt to keep the 2 consistent IIRC.

PR #20014, opened 2022-11-24
Add a message to the comments box when Approve or a Reject action is selected.

### LINKED ISSUE #19981: Prepopulate empty review message in reviewer tools for certain actions
On a reviewer tools review page, if the review text is empty (!) and

* The reviewer selects "Approve" as an action, prepopulate the text area with

"Thank you for your contribution."

* The reviewer selects "Reject" or "Reject multiple versions", prepopulate the text area with

"This add-on didn't pass review because of the following problems:

1)"

bobsilverberg:
Fixes #19981 

This is a quick and dirty way of implementing this. If we don't need these messages to be translated this might be good enough. Otherwise we probably have to get tricky with the Django form.

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -614,8 +614,8 @@ class ReviewHelper:
             can_approve_multiple = False
 
         # Definitions for all actions.
-        boilerplate_for_approve = _('Thank you for your contribution.')
-        boilerplate_for_reject = _(
+        boilerplate_for_approve = 'Thank you for your contribution.'
+        boilerplate_for_reject = (
             "This add-on didn't pass review because of the following problems:\n\n1) "
         )
 
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -614,6 +614,11 @@ class ReviewHelper:
             can_approve_multiple = False
 
         # Definitions for all actions.
+        boilerplate_for_approve = _('Thank you for your contribution.')
+        boilerplate_for_reject = _(
+            "This add-on didn't pass review because of the following problems:\n\n1) "
+        )
+
         actions['public'] = {
             'method': self.handler.approve_latest_version,
             'minimal': False,
@@ -632,6 +637,7 @@ class ReviewHelper:
             ),
             'allows_reasons': not is_static_theme,
             'requires_reasons': False,
+            'boilerplate_text': boilerplate_for_approve,
         }
         actions['reject'] = {
             'method': self.handler.reject_latest_version,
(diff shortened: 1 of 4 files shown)


static/js/zamboni/reviewers.js
```diff
@@ -60,13 +60,27 @@ function initReviewActions() {
...
+    // Input message into empty comments textbox.
+    if ($comments.val().length === 0) {
+      if ($element.text().trim() === 'Approve') {
+        insertAtCursor($comments, 'Thank you for your contribution.');
+      } else if (
+        ['Reject', 'Reject Multiple Versions'].includes($element.text().trim())
+      ) {
+        insertAtCursor(
+          $comments,
+          "This add-on didn't pass review because of the following problems:\n\n1)",
```
wagnerand, line 80:
```suggestion
          "This add-on didn't pass review because of the following problems:\n\n1) ",
```

wagnerand: Nit, which I missed in the issue. Thanks!

diox (changes requested):
It's okay to hardcode what text to insert in JavaScript, but I wouldn't rely on element text for this, as it could change or be translated, breaking the feature.

I would pass `data-` attributes instead as we've done for other reviewers features. In fact you could have the whole string you're adding set on the relevant action in `ReviewHelper` and then pass that as a `data-` attribute.

bobsilverberg:
> It's okay to hardcode what text to insert in JavaScript, but I wouldn't rely on element text for this, as it could change or be translated, breaking the feature.
> 
> I would pass `data-` attributes instead as we've done for other reviewers features. In fact you could have the whole string you're adding set on the relevant action in `ReviewHelper` and then pass that as a `data-` attribute.

I have changed the code to store the boilerplate text in `ReviewHelper` and then pass it through as `data-value`.

src/olympia/reviewers/utils.py
```diff
@@ -612,6 +612,11 @@ def get_actions(self):
             can_approve_multiple = False
 
         # Definitions for all actions.
+        boilerplate_for_approve = _('Thank you for your contribution.')
+        boilerplate_for_reject = _(
+            "This add-on didn't pass review because of the following problems:\n\n1) "
+        )
```
diox, line 618: @wagnerand do we want to translate those ? Even if we did, that would be in the reviewer's language and not the developer, so I'm thinking no.

diox (approved):
:+1: but double-check with Andreas about translating the text, I don't think we want that.

src/olympia/reviewers/utils.py
bobsilverberg: That makes sense to me too. I will remove the translations of those strings, but we can also see if Andreas disagrees with that.

wagnerand: We don't want to translate these. All reviewer communication is in English.

PR #20015, opened 2022-11-24
Allow reasons for themes

### LINKED ISSUE #19982: Replace canned responses for theme review by review reasons
For extensions, we introduced review reasons a while ago and are in the process of phasing out the old, separate canned responses system, by attaching a canned response to each review reason.

Themes don't use review reasons yet, but they could be. If we want to get rid of the old canned responses system entirely, we would need to introduce review reasons for themes as well. If so, the review reasons should show for the "Reject" and "Reject multiple versions" action.

bobsilverberg:
Fixes #19982 

This will enable a reviewer to choose reasons when doing a Reject or Reject Multiple Versions for a theme.


eviljeff (commented):
So, this would enable the reasons selection to be available for themes too... but that's it?  Seems like part of (most of?) the patch is missing unless we're offering the same rejection reasons.

bobsilverberg:
> So, this would enable the reasons selection to be available for themes too... but that's it? Seems like part of (most of?) the patch is missing unless we're offering the same rejection reasons.

That’s it! I asked Andreas to review to make sure this meets his expectations. There has not been any talk of having different reasons for themes vs. extensions. I assume either the reasons are the same, or both sets will be added and always available.

eviljeff:
> > So, this would enable the reasons selection to be available for themes too... but that's it? Seems like part of (most of?) the patch is missing unless we're offering the same rejection reasons.
> 
> That’s it! I asked Andreas to review to make sure this meets his expectations. There has not been any talk of having different reasons for themes vs. extensions. I assume either the reasons are the same, or both sets will be added and always available.

Interesting - I thought the set of theme rejections and the set of extension rejections were entirely different currently.

bobsilverberg:
> Hm, I thought I could add them on the admin?

You can add reasons, but it’s one big set. Not different reasons for different types of add-ons.

eviljeff (approved):
r+wc - the code in the patch is fine, just I was expecting there to be some filtering so that only theme reasons were shown for themes, and extension reasons shown for non-themes, because there isn't really any overlap with the common reasons for rejection (nor the reviewers doing the reviews).  That's how the existing canned responses work.

If @wagnerand is expecting all the reasons to be shown together and is okay with theme reviewers selecting from the correct theme related reasons only, and visa versa, then r+ from me.

bobsilverberg:
> r+wc - the code in the patch is fine, just I was expecting there to be some filtering so that only theme reasons were shown for themes, and extension reasons shown for non-themes, because there isn't really any overlap with the common reasons for rejection (nor the reviewers doing the reviews). That's how the existing canned responses work.
> 
> If @wagnerand is expecting all the reasons to be shown together and is okay with theme reviewers selecting from the correct theme related reasons only, and visa versa, then r+ from me.

Ok, let's see what @wagnerand's final answer is on this. If it should be different reasons for different add-on types than I will add that to this PR and not land this code yet.

wagnerand:
> > r+wc - the code in the patch is fine, just I was expecting there to be some filtering so that only theme reasons were shown for themes, and extension reasons shown for non-themes, because there isn't really any overlap with the common reasons for rejection (nor the reviewers doing the reviews). That's how the existing canned responses work.
> > If @wagnerand is expecting all the reasons to be shown together and is okay with theme reviewers selecting from the correct theme related reasons only, and visa versa, then r+ from me.
> 
> Ok, let's see what @wagnerand's final answer is on this. If it should be different reasons for different add-on types than I will add that to this PR and not land this code yet.

Oh, I see what you mean here. Thank you for being patient. I agree while there might be some overlap for a few reasons, we would need to separate them in general, as many of the code-related issues are not applicable for themes. Themes will likely need to have less than a dozen reasons, specific to appearance and listing content.

bobsilverberg:
> > > r+wc - the code in the patch is fine, just I was expecting there to be some filtering so that only theme reasons were shown for themes, and extension reasons shown for non-themes, because there isn't really any overlap with the common reasons for rejection (nor the reviewers doing the reviews). That's how the existing canned responses work.
> > > If @wagnerand is expecting all the reasons to be shown together and is okay with theme reviewers selecting from the correct theme related reasons only, and visa versa, then r+ from me.
> > 
> > 
> > Ok, let's see what @wagnerand's final answer is on this. If it should be different reasons for different add-on types than I will add that to this PR and not land this code yet.
> 
> Oh, I see what you mean here. Thank you for being patient. I agree while there might be some overlap for a few reasons, we would need to separate them in general, as many of the code-related issues are not applicable for themes. Themes will likely need to have less than a dozen reasons, specific to appearance and listing content.

Ok, this makes this a much bigger task. Let's continue the discussion in the issue.

bobsilverberg:
> 
> Ok, this makes this a much bigger task. Let's continue the discussion in the issue.

Shows how well I know this code. What I thought would be a "much bigger" task was really only a little bit bigger, unless I missed something.

src/olympia/constants/reviewers.py
```diff
@@ -210,3 +210,14 @@
     (AUTO_APPROVED, 'Was auto-approved'),
     (NOT_AUTO_APPROVED, 'Was *not* auto-approved'),
 )
+
+# Types of Add-ons for Reasons.
+REASON_ADDON_TYPE_ALL = 1
+REASON_ADDON_TYPE_EXTENSION = 2
+REASON_ADDON_TYPE_THEME = 3
```
eviljeff, line 217:
```suggestion
REASON_ADDON_TYPE_ALL = 0
REASON_ADDON_TYPE_EXTENSION = 1
REASON_ADDON_TYPE_THEME = 2
```
Or, if you changed the theme constant to `10` they'd actually be the same as base `ADDON_` constants. (0 is `ANY`; 1 is `EXTENSION`)

src/olympia/reviewers/forms.py
```diff
@@ -360,6 +361,17 @@ def __init__(self, *args, **kw):
             (k, v['label']) for k, v in self.helper.actions.items()
         ]
 
+        # Set the queryset for reasons based on the add-on type.
+        self.fields['reasons'].queryset = ReviewActionReason.objects.filter(
+            is_active__exact=True,
```
eviljeff, line 366: out of interest, what's the `__exact` for?  (I missed it before)

src/olympia/reviewers/tests/test_forms.py
```diff
@@ -194,6 +194,37 @@ def test_reasons(self):
...
+            addon_type=amo.REASON_ADDON_TYPE_EXTENSION,
+        )
+        self.reason_theme = ReviewActionReason.objects.create(
+            name='A theme only reason',
+            is_active=True,
+            addon_type=amo.REASON_ADDON_TYPE_THEME,
+        )
+        form = self.get_form()
+        choices = form.fields['reasons'].choices
+        # By default the addon is an extension.
```
eviljeff, line 215: add an `assert self.addon.type == amo.ADDON_EXTENSION` - it's not set up in this function.

eviljeff (approved):

src/olympia/reviewers/forms.py
bobsilverberg: I don't recall. Maybe just me misunderstanding the requirements of Django. I take it from your comment that I could just have used `is_active=True`?

src/olympia/constants/reviewers.py
bobsilverberg: Thanks. I think it makes sense to just reuse `ADDON_ANY`, etc., so I've done that.

src/olympia/reviewers/forms.py
diox: Yeah you don't need `__exact` most of the time

bobsilverberg: I tried it and it seems to work. 😄

Issue #20013, opened 2022-11-24
Make more models depend on `ModelBase`


diox:
Our `ModelBase` provides a `created` and `modified` date field, as well as a few useful methods. The following models do not currently inherit from it:

- `Reindexing`
- `TranslationSequence`
- `AddonCategory`
- `AddonUser`
- `AddonUserPendingConfirmation`
- `FrozenAddon`
- `GroupUser`
- `ApplicationsVersions`
- `RssKey`
- `Config`

Some of these would benefit from inheriting from `ModelBase`, especially as we might add more features to it in the future. At the very least for `Config` it would be really useful.

diox:
Note: when implementing this, make sure  `ESAddonSerializer` passes `created=None` when creating new instances of these models, because of https://github.com/mozilla/addons-server/issues/20022.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-135

PR #20026, opened 2022-11-28
Set fields with an expensive default when initializing models in `ESAddonSerializer`

### LINKED ISSUE #20022: Avoid calling expensive default callables when creating fake objects for search
Our serializers meant for search create fake model instances from Elasticsearch data, in order to share code with the database-backed serializers and use methods & properties from our models.

But there is an interesting gotcha there: calling `__init__()` on a model instance will resolve the `default` value for each field we don't provide, including those with potentially expensive callables!

This is visible on Sentry now that we have enabled the performance and profiling features on dev (https://github.com/mozilla/addons-server/issues/20004).

Fixing that should result in a ~~minor but~~ significant easy performance win for the search endpoints.

diox:
When a model `__init__()` is called, fields are set to their default values, which can sometimes have a significant cost when they are callables. This is notably the case for `UserProfile.auth_id`, and to a lesser extend `created`.

While we're at it also provide a cheaper default for `auth_id` in tests.

Fixes #20022

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -1291,7 +1291,7 @@ class ESAddonSerializer(BaseESSerializer, AddonSerializer):
         # We might not have position in ES; if not fake it with the list position.
         position = data.get('position', idx)
         preview = model_class(
-            id=data['id'], sizes=data.get('sizes', {}), position=position
+            id=data['id'], created=None, sizes=data.get('sizes', {}), position=position
         )
         preview.addon = obj
         preview.version = obj.current_version
@@ -1329,6 +1329,7 @@ class ESAddonSerializer(BaseESSerializer, AddonSerializer):
         if data:
             version = Version(
                 addon=obj,
+                created=self.handle_date(data['files'][0]['created']),
                 id=data['id'],
                 reviewed=self.handle_date(data['reviewed']),
                 version=data['version'],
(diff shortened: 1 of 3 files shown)


src/olympia/addons/serializers.py
```diff
@@ -1342,8 +1343,12 @@ def fake_version_object(self, obj, data, channel):
...
                 app_name = APPS_ALL[int(app_id)]
                 compatible_apps[app_name] = ApplicationsVersions(
-                    min=AppVersion(version=compat_dict.get('min_human', '')),
-                    max=AppVersion(version=compat_dict.get('max_human', '')),
+                    min=AppVersion(
+                        created=None, version=compat_dict.get('min_human', '')
+                    ),
+                    max=AppVersion(
+                        created=None, version=compat_dict.get('max_human', '')
+                    ),
```
eviljeff, line 1351: re-use the file creation date like we do with `Version` above?

eviljeff: But I guess if we don't expose any of these dates in the serializer respones it's redundant anyway...

eviljeff (approved):
r+wc about using alternative real dates rather than None but likely not relevant.

src/olympia/addons/serializers.py
diox: For `AppVersion` that'd be wrong IMHO, because they are not tied to the `Version` at all - `AppVersion`s are created by us when a new Firefox versions is created... 

I wanted to do it for `ApplicationsVersions`, but it turns out this model doesn't inherit from `ModelBase` and therefore doesn't have a `created` date (https://github.com/mozilla/addons-server/issues/20013#issuecomment-1328862068).

PR #20028, opened 2022-11-28
add RatingLog

### LINKED ISSUE #20023: Add a RatingLog model
To ease lookups of ActivityLogs which store some data in json blobs we have additional models for Addons (AddonLog); Versions (VersionLog) etc which link the ActivityLog instance with the related object.  We need to add one for `Rating` too.

eviljeff:
fixes #20023


src/olympia/activity/models.py
```diff
@@ -317,7 +331,7 @@ def review_log(self):
             action__in=constants.activity.LOG_REVIEWER_REVIEW_ACTION
         ).exclude(user__id=settings.TASK_USER_ID)
 
-    def total_ratings(self, theme=False):
+    def total_reviews(self, theme=False):
```
diox, line 334: ahah oops!

src/olympia/activity/tests/test_commands.py
```diff
@@ -67,3 +68,36 @@ def test_with_token_and_version_ignores_version(self):
...
+    other_action_log = ActivityLog.create(amo.LOG.CHANGE_STATUS, addon, user=user)
+    assert other_action_log.arguments == [addon]
+    other_action_log.set_arguments((addon, other_action))
+    other_action_log.save()
+
+    assert RatingLog.objects.count() == 1
+    call_command('backfill_ratinglog')
+
+    assert RatingLog.objects.count() == 2
+    ratinglog1, ratinglog2 = list(RatingLog.objects.all())
```
diox, line 99: Add a `.order_by('-id')` to guarantee ordering (default ordering on `created` might lead to intermittent failures if both are created at the exact same time - it's unlikely in this scenario but you never know)

diox (approved):

src/olympia/activity/models.py
diox: is that even used though ? I wonder if we could remove it (and the other one below)

eviljeff: I don't think it's used.  I'll remove it.
Though that likely applies to a lot of the utility functions on `ActivityManager`,

diox: Could be in a separate issue/PR if you find too much stuff to remove

eviljeff: https://github.com/mozilla/addons-server/pull/20028/commits/dda5e2a38fb7466113c770d8a7bbe8e7449f89a6

PR #20035, opened 2022-11-29
refactor addon update nominated status

### LINKED ISSUE #19773: refactor `Addon.update_nominated_status` into `Addon.update_status`
follow-up for  #19728 - `update_nominated_status` was added to change the add-on status to `STATUS_NOMINATED` if there are any listed versions that are `STATUS_AWAITING_REVIEW` and the addon status is `STATUS_NULL`, but we should be able to **carefully** fold this functionality into `update_status` directly...  Also see thread https://github.com/mozilla/addons-server/pull/19728#discussion_r984674335

eviljeff:
fixes #19773 - but restating what I noted in the issue:
- we've got to stop changing the add-on status manually for this to be a complete fix, really.  Which is a massive task when a) our tests do it **everywhere** and b) we still need to set the status manually for `STATUS_DELETED` and `STATUS_DISABLED`.  That said, with the signals what pretty much happens is: if the status is manually set to the wrong state it gets reset to the correct state which just means extra database writes and unnecessary logging in our tests.
- the case of setting `STATUS_NOMINATED` and requiring complete metadata was unclear.  I went with the closest scenario to now, where an existing `STATUS_NULL` requires complete metadata to upgrade, but other statuses will downgrade to `STATUS_NOMINATED` without requiring complete metadata.

### DIFF
--- a/src/olympia/addons/tests/test_models.py
+++ b/src/olympia/addons/tests/test_models.py
@@ -2120,7 +2120,7 @@ class TestUpdateStatus(TestCase):
 
         version_factory(addon=addon, file_kw={'status': amo.STATUS_AWAITING_REVIEW})
         addon.reload()
-        # Because it don't have complete metadata the status isn't changed
+        # Because it doesn't have complete metadata the status isn't changed
         assert not addon.has_complete_metadata(), addon.get_required_metadata()
         addon.update_status()
         assert addon.status == amo.STATUS_NULL
--- a/src/olympia/addons/tests/test_models.py
+++ b/src/olympia/addons/tests/test_models.py
@@ -1987,24 +1987,20 @@ class TestAddonNomination(TestCase):
         )
         assert addon.versions.latest().nomination == nomination
 
-    def test_nomination_not_reset_if_adding_new_versions_and_files(self):
+    def test_nomination_not_reset_if_adding_new_versions(self):
         """
         When under review, adding new versions and files should not
         reset nomination.
         """
         addon, nomination = self.setup_nomination()
-        # Switching it to a public status.
-        version = Version.objects.create(addon=addon, version='0.2')
-        File.objects.create(
-            status=amo.STATUS_APPROVED, version=version, manifest_version=2
-        )
-        assert addon.versions.latest().nomination == nomination
+
         # Adding a new unreviewed version.
         version = Version.objects.create(addon=addon, version='0.3')
         File.objects.create(
             status=amo.STATUS_AWAITING_REVIEW, version=version, manifest_version=2
         )
         assert addon.versions.latest().nomination == nomination
+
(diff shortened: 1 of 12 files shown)


src/olympia/addons/tests/test_models.py
```diff
@@ -2104,28 +2100,52 @@ def test_unlisted_versions_ignored(self):
...
+        addon = addon_factory(
+            status=amo.STATUS_NULL,
+            summary=None,
+            file_kw={'status': amo.STATUS_DISABLED},
+        )
+        assert addon.status == amo.STATUS_NULL
+
+        version_factory(addon=addon, file_kw={'status': amo.STATUS_AWAITING_REVIEW})
+        addon.reload()
+        # Because it don't have complete metadata the status isn't changed
```
bobsilverberg, line 2123: Nit: s/don't/doesn't

src/olympia/amo/tests/__init__.py
```diff
@@ -644,6 +644,7 @@ def _get_created(created):
 
 def addon_factory(status=amo.STATUS_APPROVED, version_kw=None, file_kw=None, **kw):
     version_kw = version_kw or {}
+    file_kw = file_kw or {}
```
bobsilverberg, line 647: How is this different from specifying `file_kw={}` in the function definition?

src/olympia/devhub/utils.py
```diff
@@ -350,6 +350,5 @@ def create_version_for_upload(addon, upload, channel, parsed_data=None):
         # The add-on's status will be STATUS_NULL when its first version is
         # created because the version has no files when it gets added and it
         # gets flagged as invalid. We need to manually set the status.
```
bobsilverberg, line 352: Should this comment be changed in some way to indicate that we're not actually manually setting the status, but rather requesting that it be re-evaluated?

bobsilverberg (approved):
This looks good, but to be honest I am a bit dicey on all of the status setting rules/logic, so perhaps @diox should take a look at it as well? I also added some minor comments, so r+wc from me.

src/olympia/amo/tests/__init__.py
eviljeff: it's a python gotcha - you can't use mutable defaults in the kwargs as they're shared with future calls to the function.

diox (approved):
> the case of setting `STATUS_NOMINATED` and requiring complete metadata was unclear. I went with the closest scenario to now, where an existing `STATUS_NULL` requires complete metadata to upgrade, but other statuses will downgrade to `STATUS_NOMINATED` without requiring complete metadata.

I'm not 100% sure what should happen in this case, but I think your approach is the safest. There might be a edge-case I'm not seeing, we'll see if that ever comes up...

PR #20044, opened 2022-12-01
Distinguish between FxA logins and API token logins in ActivityLog

### LINKED ISSUE #20043: Have a way to distinguish FxA logins from API logins
The [`LOG_IN` activity](https://github.com/mozilla/addons-server/blob/0768664e5c409dec1a7e993be680498076d06bc2/src/olympia/constants/activity.py#L829) records when a user logged in, but we don't know if the login happened through FxA or an API token.

Let's add this information somehow, maybe through a different action, or some other way.

diox:
Fixes #20043

### DIFF
--- a/src/olympia/accounts/tests/test_views.py
+++ b/src/olympia/accounts/tests/test_views.py
@@ -213,6 +213,12 @@ class TestLoginUserAndRegisterUser(TestCase):
             ActivityLog.objects.filter(user=self.user, action=amo.LOG.LOG_IN.id).count()
             == 1
         )
+        assert (
+            ActivityLog.objects.filter(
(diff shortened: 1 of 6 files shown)


wagnerand:
Do we want to rename `LOG_IN` to something more descriptive, like `LOG_IN_FXA`?

diox:
I'm not sure - technically it's a catch-all, we just don't really have another log in path (except for the special super-create user for tests).

eviljeff (approved):
It's kinda weird the way that an api request that used session auth would get a single `LOG_IN` event when the FxA auth process finished and then no more for up to 30 days; but if token auth is used there is a login event for every single api request.  But I see that's not a new implementation detail so 🤷

PR #20056, opened 2022-12-05
Add a max length on TextFields

diox:
A database constraint will be added in a future issue, but we can still make our forms & serializers fields enforce according to what is set on the models.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1788117


src/olympia/activity/migrations/0019_merge_0018_auto_20221122_1313_0018_auto_20221128_1706.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 3.2.16 on 2022-12-05 11:46
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('activity', '0018_auto_20221122_1313'),
+        ('activity', '0018_auto_20221128_1706'),
```
eviljeff, line 10: could have just renamed the migration to 0019?

src/olympia/addons/serializers.py
```diff
@@ -346,7 +346,10 @@ class SimpleVersionSerializer(MinimalVersionSerializer):
         source='file.strict_compatibility', read_only=True
     )
     license = CompactLicenseSerializer(read_only=True)
-    release_notes = TranslationSerializerField(required=False, read_only=True)
+    release_notes = TranslationSerializerField(
+        required=False,
+        read_only=True,
+    )
```
eviljeff, line 352: no change

src/olympia/addons/serializers.py
```diff
@@ -870,24 +872,31 @@ def get_apps(self, obj):
...
-class AddonSerializer(serializers.ModelSerializer):
+class AddonSerializer(AMOModelSerializer):
     authors = UserSerializerWithPictureUrl(
         many=True, source='listed_authors', read_only=True
     )
     categories = CategoriesSerializerField(source='all_categories', required=False)
     contributions_url = ContributionSerializerField(
-        source='contributions', required=False
+        source='contributions',
+        required=False,
```
eviljeff, line 882: no changes here too...

eviljeff: I see a few more changes in this file that you can revert but I won't comment on each line.

src/olympia/reviewers/migrations/0021_merge_20221205_1147.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 3.2.16 on 2022-12-05 11:47
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('reviewers', '0020_auto_20221114_1835'),
+        ('reviewers', '0020_reviewactionreason_addon_type'),
```
eviljeff, line 10: same question for these migrations with clashing numbers - just rename them?

src/olympia/api/fields.py
```diff
@@ -6,6 +6,7 @@
 from django.utils.encoding import smart_str
 from django.utils.translation import get_language, gettext, gettext_lazy as _, override
 
+
```
eviljeff, line 9: random extra new line

src/olympia/amo/utils.py
```diff
@@ -44,6 +45,7 @@
     url_has_allowed_host_and_scheme,
 )
 
+
```
eviljeff, line 48: another random extra line

src/olympia/activity/migrations/0019_merge_0018_auto_20221122_1313_0018_auto_20221128_1706.py
diox: I was too lazy to check which migrations I had already applied or not locally and revert them to get to the proper state, so I figured I'd use django migration's merge feature, that's why it exists after all!

PR #20057, opened 2022-12-05
Allow LIKE searches for email in user admin, drop search by username

### LINKED ISSUE #20050: Improve email searches in user admin
- Make searches faster for exact email/id matches
- Allow basic pattern matching (`foo*@example.com`, `foo@*` etc)

diox:
Fixes #20050


src/olympia/users/tests/test_admin.py
```diff
@@ -47,20 +47,68 @@ def setUp(self):
...
 
-    def test_search_for_multiple_users(self):
+    def test_search_by_email_simple(self):
         user = user_factory(email='someone@mozilla.com')
         self.grant_permission(user, 'Users:Edit')
         self.client.force_login(user)
         another_user = user_factory()
         response = self.client.get(
             self.list_url,
-            {'q': f'{self.user.pk},{another_user.pk},foobaa'},
```
diox, line 57: Searching by ids _and_ email at the same time is no longer supported for performance reasons.

src/olympia/amo/models.py
```diff
@@ -26,6 +27,21 @@
 log = olympia.core.logger.getLogger('z.addons')
 
 
+@Field.register_lookup
+class Like(Lookup):
```
eviljeff, line 31: As this lookup is going to be available to all fields can we have a generic test for it under `amo/tests/`?

eviljeff (approved):

PR #20060, opened 2022-12-06
rewrite review reports to use ActivityLog records

### LINKED ISSUE #20041: Update reviewer reports to use ActivityLog records
We have automated reviewer activity reports but they currently use the `ReviewerScore` records.  With that model/feature going away in #18049 the report sql needs updating to use something different, such as the ActivityLog records we store for every review action.

eviljeff:
fixes #20041 
I didn't bother updating the quarterly reports (I just deleted them) as they were solely volunteer based and we don't have any non-theme volunteers.
The new sql scripts I developed using redash and, in my tests, returned almost the same results as the current ones that use the reviewer incentives.  The differences (there were a few more reviews) I attribute to the additional review types that are captured.


src/olympia/reviewers/management/commands/review_reports_sql/addon/breakdown.sql
```diff
@@ -2,30 +2,46 @@ SELECT 'All Reviewers' AS `Group`,
...
-JOIN `users` `u` ON `u`.`id` = `rs`.`user_id`
-WHERE DATE(`rs`.`created`) BETWEEN @WEEK_BEGIN AND @WEEK_END
-  AND `u`.`deleted` = 0
+FROM (
+  SELECT DISTINCT `created`, `user_id`, `action`, MAX(`id`) AS `id`
+  FROM `log_activity`
+  WHERE DATE(`created`) BETWEEN @WEEK_BEGIN AND @WEEK_END
+  /* The type of review, see constants/activity.py */
+  AND `action` IN (21, 43, 144, 165)
+  GROUP BY `created`, `user_id`, `action`
```
diox, line 11: `DISTINCT` *and* `GROUP BY` ? 

Also, we should add a comment here documenting that this only works because we explicitly set `created` to be the exact same date for actions that record multiple `ActivityLog`s.

src/olympia/reviewers/management/commands/review_reports_sql/addon/breakdown.sql
```diff
@@ -2,30 +2,46 @@ SELECT 'All Reviewers' AS `Group`,
...
-LEFT JOIN `editors_autoapprovalsummary` `aa` ON `aa`.`version_id` = `rs`.`version_id`
-JOIN `users` `u` ON `u`.`id` = `rs`.`user_id`
-WHERE DATE(`rs`.`created`) BETWEEN @WEEK_BEGIN AND @WEEK_END
-  AND `u`.`deleted` = 0
+FROM (
+  SELECT DISTINCT `created`, `user_id`, `action`, MAX(`id`) AS `id`
+  FROM `log_activity`
+  WHERE DATE(`created`) BETWEEN @WEEK_BEGIN AND @WEEK_END
+  /* The type of review, see constants/activity.py */
+  AND `action` IN (21, 43, 144, 165)
```
diox, line 10: I know this was present in the original through `note_key`, but I think this should be a parameter instead, so we could use `%s` here and pass `amo.LOG_REVIEWER_REVIEW_ACTION` (which contains more ids but I think that's just because it has obsolete constants as well - we could clean it up) to the `execute()` call and not worry about maintaining this list.

diox (commented):
My comments also apply to the other files. @wagnerand should look at the queries as well.

src/olympia/reviewers/management/commands/review_reports_sql/addon/breakdown.sql
eviljeff: https://github.com/mozilla/addons-server/pull/20060/commits/853aa0071b02f8254638682381d2c7a1046f2d9d

src/olympia/reviewers/management/commands/review_reports_sql/addon/breakdown.sql
eviljeff: added as a parameter https://github.com/mozilla/addons-server/pull/20060/commits/b796def1381f3218806680ff9303ed91a069c9c8 - extra fun because mysql doesn't support `SET` with a list of values so you need to convert the ids to a comma-seperated-list and use a function to check the id is in it.

Can't use `amo.LOG_REVIEWER_REVIEW_ACTION` because it's a mix of content and non-content review actions and these reports split them out, so I just defined the appropriate actions instead.

src/olympia/reviewers/management/commands/review_reports_sql/addon/breakdown.sql
eviljeff: and https://github.com/mozilla/addons-server/pull/20060/commits/8618d2287f584529a2163cb7c6890175b690c2ee for the comments

diox (approved):

src/olympia/reviewers/management/commands/review_reports_sql/addon/breakdown.sql
```diff
@@ -2,30 +2,50 @@ SELECT 'All Reviewers' AS `Group`,
...
-  AND `u`.`deleted` = 0
+FROM (
+  /* NOTE: This works to reduce reviews that acted on multiple reviews to a single review
+     because we set created to be the exact same datetime. */
+  SELECT `created`, `user_id`, `action`, MAX(`id`) AS `id`
+  FROM `log_activity`
+  WHERE DATE(`created`) BETWEEN @WEEK_BEGIN AND @WEEK_END
+  /* The type of review, see constants/activity.py */
+  AND FIND_IN_SET(`action`, @ACTIVITY_ID_LIST) > 0
+  GROUP BY `created`, `user_id`, `action`
```
wagnerand, line 13: Wouldn't this fail if a user submits two separate reviews at the same time?

src/olympia/reviewers/management/commands/review_reports_sql/addon/breakdown.sql
```diff
@@ -2,30 +2,50 @@ SELECT 'All Reviewers' AS `Group`,
        IFNULL(FORMAT(SUM(aa.weight), 0), '-') AS `Total Risk`,
        IFNULL(FORMAT(AVG(aa.weight), 2), '-') AS `Average Risk`,
        FORMAT(COUNT(*), 0) AS `Add-ons Reviewed`
-FROM `reviewer_scores` `rs`
```
wagnerand, line 5: I suppose we can delete the separate stats for volunteers as well (=this entire file).

It would be great if the total for all reviewers recorded here could be shown in the "Weekly Add-on Reviews" section instead. To make it easier to understand, we can drop the "5 Reviews or more" constraint.

The same applies to content reviews.

src/olympia/reviewers/management/commands/review_reports_sql/addon/breakdown.sql
eviljeff: if they were submitted at _exactly_ the same time (i.e. same millisecond), yes.  Seemed like an edge case that wouldn't affect the numbers in a significant way.

src/olympia/reviewers/management/commands/review_reports_sql/addon/breakdown.sql
eviljeff: @wagnerand to confirm, we'd just end up with 3 sql files:
- addon/weekly.sql (without the >=5 constraint)
- addon/risk.sql 
- content/weekly.sql (without the >=10 constraint)

src/olympia/reviewers/management/commands/review_reports_sql/addon/breakdown.sql
wagnerand: Sounds ok. I wanted to make sure I got it right and we are aware of it.

src/olympia/reviewers/management/commands/review_reports_sql/addon/breakdown.sql
wagnerand: Yes. Can we the weekly section include the totals?

eviljeff: https://github.com/mozilla/addons-server/pull/20060/commits/9ac2259f4f75ff81533c2cd193acfb1e6d9fead1

wagnerand (approved):

PR #20061, opened 2022-12-06
replace chardet with charset-normalizer

### LINKED ISSUE #20059: Replace `chardet` by `charset_normalizer`
`chardet` used to be required by `requests`, but they switched to `charset_normalizer` instead (though retaining some compatibility with `chardet`). We should follow, we are not using `chardet` ourselves, it's only there because of `requests`.

eviljeff:
fixes #20059 (note, we also install `chardet` in `docs.txt`, but I left that one as I figured something in the docs stack still needs it)


requirements/prod.txt
```diff
@@ -256,10 +256,96 @@ bleach==5.0.1 \
 celery==5.2.7 \
     --hash=sha256:138420c020cd58d6707e6257b6beda91fd39af7afde5d36c6334d175302c0e14 \
     --hash=sha256:fafbd82934d30f8a004f81e8f7a062e31413a23d444be8ee3326553915958c6d
-# chardet is required by requests
-chardet==5.1.0 \
-    --hash=sha256:0d62712b956bc154f85fb0a266e2a3c5913c2967e00348701b32411d6def31e5 \
-    --hash=sha256:362777fb014af596ad31334fde1e8c327dfdb076e1960d1694662d46a6917ab9
+# charset-normalizer is required by requests
+charset-normalizer==3.0.1 \
```
diox, line 260: mmm technically, latest release of `requests` is not marked as compatible with `charset-normalizer` 3.x. OTOH, there was a recent change to allow it (https://github.com/psf/requests/commit/c57f1f0ca10e61771b459c857182c23626607312) without any code changes, plus we likely don't even need it for the requests we're making, so we should be fine...

diox: `requests` does warn at import though:
```
/deps/lib/python3.10/site-packages/requests/__init__.py:109: RequestsDependencyWarning: urllib3 (1.26.13) or chardet (None)/charset_normalizer (3.0.1) doesn't match a supported version!
  warnings.warn(
```

So maybe we should err on the safe side and stay on 2.x until they make a new `requests` release...

Issue #20071, opened 2022-12-07
The "Developer reply" field from version Review history doesn't show an error message when the character limit is exceeded


AlexandraMoga:
### Describe the problem and steps to reproduce it:
1. As a developer, open the Manage versions page for an addon
2. Click on the Review History to expand the form
3. Post a developer reply which exceeds the character limit for that field

### What happened?
There is a 400 error returned by the `https://addons.allizom.org/api/v5/addons/addon/<addon_id>/versions/<version_id>/reviewnotes/` visible in the browser console but there is no actual error message displayed  in the field.

### What did you expect to happen?
There should be an error message displayed for the Developer reply field when the character limit is exceeded.

diox:
Added a limit so the widget so you won't be able to enter more data than the field allows. This should be cherry-picked once verified on dev.

AlexandraMoga:
The "Developer reply" field from version Review history has a fixed length now of 100.000 characters. 

This issue is verified fixed on -dev.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-179

PR #20072, opened 2022-12-07
Add maxlength to textarea/input widgets and display missing errors in devhub

### LINKED ISSUE #20069: Some Devhub text fields won't throw an error message when the character limit is exceeded
#### STR for new addon uploads:
1. Upload a new listed addon on AMO dev or stage
2. On the addon submission form, add some text in the "Notes to Reviewer"  field that exceeds the character limit for this textarea

#### STR for new listed version submissions:
1. Upload a new listed version for an addon
2. On the final submission step, add some text in the "Release Notes" and the "Notes to Reviewer" fields that exceeds the character limit for these textareas

#### STR for editing the Whiteboard from Devhub:
1. Open an add-on Edit page in Devhub
2. Add some text in the "Whiteboard"  field that exceeds the character limit for this textarea

### What happened?
The submission/edit is not successful and there is no error indicating what caused the submission/edit to fail

The mentioned textareas should display an error message when the character limit is exceeded

Both the Release notes and Notes to reviewer fields are displaying an error message when the fields are edited from the Manage versions page in Devhub

diox:
Fixes #20069
Fixes https://github.com/mozilla/addons-server/issues/20071

We could add more tests but I feel that's enough ?

### DIFF
--- a/src/olympia/amo/utils.py
+++ b/src/olympia/amo/utils.py
@@ -1281,13 +1281,17 @@ class BaseModelSerializerAndFormMixin:
                     field.max_length = max_length
                     # Normally setting max_length on the field would be enough,
                     # but because we're late, after the field's __init__(), we
-                    # also need to convert that into a validator ourselves if
-                    # the field requires it. Unfortunately some fields
-                    # (FileField) do not work like that and instead deal with
-                    # max_length dynamically, with a custom error message that
-                    # is generated dynamically, so we need to avoid those.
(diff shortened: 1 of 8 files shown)


eviljeff (approved):
r+
yeah I feel like it'd be difficult to cover everything in tests because we don't know where any other bugs might be lurking in the spaghetti of devhub views, templates, and ajax requests.

PR #20105, opened 2022-12-16
Introduce a base ModelAdmin class, use it everywhere

### LINKED ISSUE #19801: Don't execute a full `COUNT(*)` query for objects in changelist for admin pages that don't display it
When displaying lists of items for a particular model in the admin, if `show_full_result_count` on the `ModelAdmin` is `True` (the default), Django execute a full unfiltered `COUNT(*)` query to determine the total number of items in the database. Then, if the number of items is different than the number of items with filters etc applied, this is displayed in the search template in parenthesizes:
Screenshot 2022-10-18 at 18-08-36 Select addon to change Django site admin

But, if the search template is not shown, or filters not applied, this full unfiltered `COUNT(*)` is executed but is never displayed.

Let's stop wasting time and avoid that query completely when we know the search bar is never going to be displayed and set `show_full_result_count = False` on relevant `ModelAdmin`s.

diox:
Set `show_full_result_count = False` on it to avoid a useless `COUNT(*)` query (Used by django to show the full total of objets in the db with no filters/search applied, which we don't need most of the time)

Fixes #19801

### DIFF
--- a/src/olympia/abuse/admin.py
+++ b/src/olympia/abuse/admin.py
@@ -12,7 +12,7 @@ from rangefilter.filter import DateRangeFilter as DateRangeFilterBase
 from olympia import amo
 from olympia.access import acl
 from olympia.addons.models import Addon, AddonApprovalsCounter
-from olympia.amo.admin import CommaSearchInAdminMixin
+from olympia.amo.admin import AMOModelAdmin
 from olympia.ratings.models import Rating
 from olympia.translations.utils import truncate_text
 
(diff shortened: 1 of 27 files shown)


eviljeff (approved):
I wonder if there is anywhere we _do_ want to show the total count?

diox:
> I wonder if there is anywhere we _do_ want to show the total count?

I don't think so, which is why I set it by default. We can always override it on specific admin classes if this comes up, but my reasoning was that admins generally don't care about the full count in the UI, as we are using those tools for specific tasks like looking up a particular add-on, searching for users, etc. When someone does care about the full count, they can remove the filters, or do a query on STMO.

PR #20149, opened 2023-01-06
Remove renovate, use dependabot for npm too

### LINKED ISSUE #20146: Please use Dependabot instead of Renovate
We want to use Dependabot consistently across all our (add-ons) repos. Please update the config to use Dependabot instead of Renovate. Thanks!

diox:
Fixes #20146


.github/dependabot.yml
```diff
@@ -1,5 +1,10 @@
 version: 2
 updates:
+- package-ecosystem: "npm"
+  directory: "/"
+  schedule:
+    interval: weekly
```
diox, line 6: I figured that was good enough for addons-server

eviljeff: generally, but for updates to our own packages like the linter we usually like it quicker

diox: Good point

diox: e1620da134273d95cbb8733014934ae604039452

eviljeff (approved):

PR #20151, opened 2023-01-09
replace Version.nomination with Version.due_date

### LINKED ISSUE #19985: Replace review wait time by due date
The reviewer tools that have versions pending manual review currently have a "wait time" column which is also the order key.

We want to introduce a new order key and column called _Due date_. The default due date for an add-on is 3 business days after its nomination for review. The wait time column should be removed from the queues.

This can happen after #19984 if it's easier.

eviljeff:
fixes #19985 - apart from admins being able to set the due date because the PR is large enough. I'll split that into a follow-up issue.


src/olympia/versions/utils.py
```diff
@@ -135,3 +136,16 @@ def process_color_value(prop, value):
...
+    starting = starting or datetime.now()
+    # if starting falls on the weekend, move it to Monday morning
+    if starting.weekday() in (5, 6):
+        starting = starting.replace(hour=9) + timedelta(days=(7 - starting.weekday()))
+
+    due = starting + timedelta(days=REVIEWER_STANDARD_REVIEW_TIME)
+    # if due falls on a weekend, or passes over a weekend, add on 2 days
+    if due.weekday() in (5, 6) or due.weekday() < starting.weekday():
+        due += timedelta(days=2)
+    return due
```
eviljeff, line 151: @wagnerand The issue said 3 business days so this code is to work around pesky weekends.  Effectively, if now is a weekend (i.e. when we're trying to set the date) then it's as if it was being set 9am Monday; and if the resulting due date lands on a weekend, or the period passes over a weekend, an extra 2 days is added on.

diox: Do we need to handle holidays (if so, for which country) ?

eviljeff: that was why I didn't handle holidays 😬

src/olympia/versions/migrations/0031_rename_nomination_version_due_date.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('versions', '0030_auto_20221122_1312'),
+    ]
+
+    operations = [
+        migrations.RenameField(
```
diox, line 13: This will break all `Version` queries during the push - I'm not sure how fast the migrations will be, so that seems problematic. Since we're rewriting the values anyway, we should probably add a new field instead, doing the usual dance of removing the old field from models, then doing a SQL migration to remove the column next week.

eviljeff: https://github.com/mozilla/addons-server/pull/20151/commits/d1fcaa1b583df3b4df866ef50a6d458cc533c9f4

src/olympia/reviewers/utils.py
```diff
@@ -172,7 +172,7 @@ def render_score(self, value):
 class PendingManualApprovalQueueTable(AddonQueueTable):
     addon_type = tables.Column(verbose_name=_('Type'), accessor='type', orderable=False)
     waiting_time = tables.Column(
```
diox, line 174: Rename this column and the associated `_get_waiting_time()`, `render_waiting_time()` methods (and update `default_order_by`), now that it's not really about waiting time ?

diox (approved):

src/olympia/reviewers/utils.py
eviljeff: https://github.com/mozilla/addons-server/pull/20151/commits/9fe51c57b89ed6af368a93ee03e67571aae2ea54

PR #20182, opened 2023-01-12
Make double-escaping tests for _TransWidget more robust

diox:
The existing tests were misleading: the content _is_ escaped at some point, it has to before being displayed to the user (pyquery was buggy and unescaping...)

What we want to make sure is that we aren't rendering already escaped content, escaping it twice (once per the translation field, once per the django form widget render method)

Should fix tests in https://github.com/mozilla/addons-server/pull/20121

### DIFF
--- a/src/olympia/devhub/tests/test_views_edit.py
+++ b/src/olympia/devhub/tests/test_views_edit.py
@@ -139,25 +139,6 @@ class BaseTestEditDescribe(BaseTestEdit):
         doc = pq(response.content)
         assert doc('form').attr('action') != old_edit
 
-    def test_edit_summary_escaping(self):
-        data = self.get_dict()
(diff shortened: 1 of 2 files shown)


src/olympia/devhub/tests/test_views_edit.py
```diff
@@ -139,25 +139,6 @@ def test_edit_slug_valid(self):
         doc = pq(response.content)
         assert doc('form').attr('action') != old_edit
 
-    def test_edit_summary_escaping(self):
```
diox, line 142: This test was a bit redundant, there is already an `xss` test below in the same file, and the test I fixed regarding the widget in `translations/tests/test_widgets.py`

eviljeff (approved):

Issue #20152, opened 2023-01-09
Make `Version.due_date` settable by admins


eviljeff:
spun off from #19985

An admin reviewer (Reviews:Admin) should be able to set a new and override or clear an existing due date on the review tools page. This action should be recorded in the Add-on important changes history section.

diox:
Note that this should be available even for add-ons not currently waiting for review - that should make them appear in the queue from https://github.com/mozilla/addons-server/issues/19984 even if they are not waiting for review.

diox:
@wagnerand :

Currently we have set up due dates to be automatic, being dropped as soon as the version is auto-approvable, essentially (caveats apply, but the logic is mostly this: version needs to be waiting for review, without flags preventing auto-approval). This is mostly how things worked with the nomination date.

With this issue we are going to have due dates explicitly set by admins (please double-check the permission required to set/clear that btw), on potentially already approved versions. I assume the logic regarding automatically clearing the due date is going to be different in that case. So, should we clear a manually set due date it when:
- Status of the file is changed to disabled ?
- Version is deleted ?
- Version is disabled by developer ?
- Add-on itself is disabled ? made invisible ? deleted ?
- Other scenarios I'm not thinking about ?

(The answers can be different depending on existing status and channel - if so, specify)

wagnerand:
> So, should we clear a manually set due date it when:

(Correct me if you see a case I'm missing but I believe we'd want the following cases to apply to all due dates, regardless of whether they were set automatically or manually?)

>     * Status of the file is changed to disabled ?
* If disabled by developer, only if it was not signed.
* If disabled by an admin, yes.
If that condition adds to much complexity, then we should not clear it.

>     * Version is deleted ?
Only if it was not signed. If that condition adds to much complexity, then we should not clear it.
 
>     * Version is disabled by developer ?
Only if it was not signed. If that condition adds to much complexity, then we should not clear it.
 
>     * Add-on itself is disabled ? made invisible ? deleted ?
* I assume you mean mozilla-disabled (aka force-disabled) for the first case. There we should clear the due-dates for all versions.
* When the developer changes their add-on to invisible or deletes it, we should only clear if the version was not signed. If that condition adds to much complexity, then we should not clear it.
 
>     * Other scenarios I'm not thinking about ?
Some of them might already be done or we are doing them anyway, I'm listing them here to seek completeness:
* When the version gets manually approved, approval confirmed, rejected or delay-rejected.
* When the version gets blocked.
[...]

PR #20184, opened 2023-01-13
Only set due date for versions that will be manually reviewed

### LINKED ISSUE #20162: Only set `due_date` (nee `nominated`) for add-ons waiting for a human review
We currently set `due_date`/`nominated` on all versions that are pending approval, including versions that will be automatically approved.  This doesn't make much sense - auto approval takes place on it's own schedule, not based on the value of the field - and means we can't rely on the field being set _only_ when a manual review is going to take place.

The extra complication is we need to ensure the date is set whenever we're changing flags or other values that would cause a version that would have been auto approved to now need human review.

eviljeff:
fixes #20162

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -72,6 +72,7 @@ from olympia.versions.models import (
     VersionReviewerFlags,
     inherit_due_date,
 )
+from olympia.versions.utils import get_review_due_date
 
 from . import signals
 
@@ -1775,6 +1776,21 @@ class Addon(OnChangeMixin, ModelBase):
                 tag.remove_tag(self)
         self.tag_list = new_tag_list
 
+    def update_all_due_dates(self):
+        for version in Version.objects.should_have_due_date().filter(
+            addon=self, due_date__isnull=True
+        ):
+            due_date = get_review_due_date()
+            log.info(
+                'Version %r (%s) due_date set to %s', version, version.id, due_date
+            )
+            version.update(due_date=due_date, _signal=False)
+        for version in Version.objects.should_have_due_date(not_=True).filter(
+            addon=self, due_date__isnull=False
+        ):
+            log.info('Version %r (%s) due_date cleared', version, version.id)
+            version.update(due_date=None, _signal=False)
+
 
 dbsignals.pre_save.connect(save_signal, sender=Addon, dispatch_uid='addon_translations')
 
@@ -1861,6 +1877,13 @@ class AddonReviewerFlags(ModelBase):
     notified_about_expiring_delayed_rejections = models.BooleanField(
         default=None, null=True
     )
+
+
+@receiver(
+    dbsignals.post_save, sender=AddonReviewerFlags, dispatch_uid='addon_review_flags'
+)
+def update_due_date_for_auto_approval_changes(sender, instance=None, **kwargs):
+    instance.addon.update_all_due_dates()
 
 
 class AddonRegionalRestrictions(ModelBase):
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -1777,16 +1777,16 @@ class Addon(OnChangeMixin, ModelBase):
         self.tag_list = new_tag_list
 
     def update_all_due_dates(self):
-        for version in Version.objects.should_have_due_date().filter(
-            addon=self, due_date__isnull=True
+        for version in self.versions.should_have_due_date().filter(
+            due_date__isnull=True
         ):
             due_date = get_review_due_date()
             log.info(
                 'Version %r (%s) due_date set to %s', version, version.id, due_date
             )
             version.update(due_date=due_date, _signal=False)
-        for version in Version.objects.should_have_due_date(not_=True).filter(
-            addon=self, due_date__isnull=False
+        for version in self.versions.should_have_due_date(negate=True).filter(
+            due_date__isnull=False
         ):
             log.info('Version %r (%s) due_date cleared', version, version.id)
             version.update(due_date=None, _signal=False)
--- a/src/olympia/amo/tests/__init__.py
+++ b/src/olympia/amo/tests/__init__.py
@@ -43,6 +43,7 @@ from olympia.addons.models import (
     Addon,
     AddonCategory,
     AddonGUID,
+    AddonReviewerFlags,
     update_search_index as addon_update_search_index,
 )
 from olympia.amo.reverse import get_url_prefix, set_url_prefix
@@ -674,6 +675,7 @@ def addon_factory(status=amo.STATUS_APPROVED, version_kw=None, file_kw=None, **k
         slug = name.replace(' ', '-').lower()[:30]
 
     promoted_group = kw.pop('promoted', None)
+    reviewer_flags = kw.pop('reviewer_flags', None)
 
     kwargs = {
         # Set artificially the status to STATUS_APPROVED for now, the real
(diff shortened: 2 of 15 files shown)


src/olympia/amo/tests/__init__.py
```diff
@@ -943,6 +944,7 @@ def version_factory(file_kw=None, **kw):
         file_factory(version=ver, **file_kw)
     if promotion_approved:
         kw['addon'].promotedaddon.approve_for_version(version=ver)
+    kw['addon'].update(status=kw['addon'].status)  # forcing an update
```
diox, line 947: `addon_factory()` takes great care to set up `save` calls in a very specific order, so I'm not sure this is a good idea to force an update when creating a version

src/olympia/versions/models.py
```diff
@@ -796,10 +796,15 @@ def disable_old_files(self):
                 f.update(status=amo.STATUS_DISABLED)
 
     def reset_due_date(self, due_date=None):
-        if not self.due_date or due_date:
-            due_date = due_date or get_review_due_date()
-            # We need signal=False not to call update_status (which calls us).
-            self.update(due_date=due_date, _signal=False)
+        if self.should_have_due_date:
```
diox, line 799: Something to keep in mind - which maybe could help tests if we need an override mechanism there - is that we're going to need something to "keep" the `due_date` if it was set explicitly by an admin in https://github.com/mozilla/addons-server/issues/20152

src/olympia/addons/models.py
```diff
@@ -1775,6 +1775,12 @@ def set_tag_list(self, new_tag_list):
                 tag.remove_tag(self)
         self.tag_list = new_tag_list
 
+    def update_all_due_dates(self):
+        if self.current_version:
```
diox, line 1779: That doesn't seem sufficient - that handles removing `due_date` once the add-on has been approved, assuming this is executed after the new `current_version` has been set, but wouldn't work for any other case with listed versions, right ? In particular I'm thinking about rejections, or flags changing while the add-on is still waiting for review

src/olympia/amo/tests/__init__.py
eviljeff: This was something I added as a fix for one of the tests - I can't remember which now, or if it was actually necessary in th end

src/olympia/versions/models.py
eviljeff: I was thinking there would be a "human review requested" flag/date that could be added or reused.

diox: Yeah that makes sense.

src/olympia/addons/models.py
diox: Which doesn't seem ideal - we don't want to have to go through all versions every time.
- When we are adding a due date, it only needs to happen to a single version (the latest waiting for review in the right channel(s)).
- When we are removing a due date (which I guess we need to go through when adding one too, to remove the old one from the old versions), we only need to remove it to versions that had one

Right ?

eviljeff: Right.  Currently the logic to clear or set is located in `reset_due_date`, that only works on a single version, so that'd need changing.

src/olympia/versions/models.py
```diff
@@ -168,6 +169,32 @@ def auto_approvable(self):
         )
         return qs
 
+    def should_have_due_date(self, not_=False):
+        return getattr(self, 'exclude' if not_ else 'filter')(
```
eviljeff, line 173: I've a feeling there's a cleaner way to do this but it works.

src/olympia/versions/models.py
```diff
@@ -168,6 +169,33 @@ def auto_approvable(self):
...
+            Q(file__status=amo.STATUS_AWAITING_REVIEW)
+            & Q(reviewerflags__pending_rejection__isnull=True)
+            & Q(
+                Q(addon__type__in=amo.GROUP_TYPE_THEME)
+                | Q(
+                    addon__promotedaddon__group_id__in=(g.id for g in PRE_REVIEW_GROUPS)
+                )
```
diox, line 180: We currently ignore promotions for unlisted (which allows pseudo-beta versions to be auto-approval, so I assume we want to continue doing that).

eviljeff: I can see pros and cons with that but if that's what we currently do then I'll keep it the same

eviljeff: https://github.com/mozilla/addons-server/pull/20184/commits/15ac129c2b2179e5aec0fd93ecc76cf8815461dd

src/olympia/addons/models.py
```diff
@@ -1775,6 +1776,21 @@ def set_tag_list(self, new_tag_list):
                 tag.remove_tag(self)
         self.tag_list = new_tag_list
 
+    def update_all_due_dates(self):
+        for version in Version.objects.should_have_due_date().filter(
```
diox, line 1780: Can use `self.versions` instead of `Version.objects.(...).filter(addon=self)` here and below

src/olympia/versions/models.py
diox: How about something like:
```python
method = getattr(self, 'exclude' if not_ else 'filter')
is_theme = Q(addon__type__in=amo.GROUP_TYPE_THEME)
has_auto_approval_delayed = Q(addon__reviewerflags__auto_approval_delayed_until__isnull=False)
requires_manual_listed_approval_and_is_listed = Q(
    Q(addon__reviewerflags__auto_approval_disabled=True)
    | Q(
        addon__reviewerflags__auto_approval_disabled_until_next_approval=True  # noqa
    )
    | Q(
        addon__promotedaddon__group_id__in=(
            g.id for g in PRE_REVIEW_GROUPS
        )
    ),
    channel=amo.CHANNEL_LISTED,
)
requires_manual_unlisted_approval_and_is_unlisted = Q(
    Q(addon__reviewerflags__auto_approval_disabled_unlisted=True)
    | Q(
        addon__reviewerflags__auto_approval_disabled_until_next_approval_unlisted=True  # noqa
    ),
    channel=amo.CHANNEL_UNLISTED,
),
return method(
    Q(file__status=amo.STATUS_AWAITING_REVIEW)
    & Q(reviewerflags__pending_rejection__isnull=True)
    & Q(is_theme | has_auto_approval_delayed | requires_manual_listed_approval_and_is_listed)
)
```

?

src/olympia/versions/models.py
```diff
@@ -796,14 +826,21 @@ def disable_old_files(self):
                 f.update(status=amo.STATUS_DISABLED)
 
     def reset_due_date(self, due_date=None):
```
diox, line 828: Only use of the parameter - which is confusing me - seems to be in tests ?

eviljeff: it's used by `inherit_due_date` just below

src/olympia/versions/models.py
```diff
@@ -168,6 +169,35 @@ def auto_approvable(self):
         )
         return qs
 
+    def should_have_due_date(self, not_=False):
```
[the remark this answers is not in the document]
eviljeff: django calls it `negate` when it's building queries so I'm going to change it to that.

src/olympia/versions/models.py
diox: Ah, I had missed that because it was part of the previous patch, and it was simpler back then... I would love a docstring on here as well to make what it's doing clearer

eviljeff: add some docstrings

diox:
Merging cause it's my way and it has been approved.

PR #20209, opened 2023-01-19
Remove translations from reviewer/admin tools

### LINKED ISSUE #20208: Stop translating reviewer tools/admin tools
Translating admin tools/reviewer tools is not very useful: all extensions reviews are done by staff/contractors these days, and they have to understand English as all communication with developers or staff will happen in English. It's also annoying to translate because translators might not be able to see the strings in action.

Therefore, let's stop doing that. Let's remove all gettext/etc calls around strings in admin tools/reviewer tools.

diox:
Fixes #20208

### DIFF
--- a/src/olympia/abuse/tests/test_admin.py
+++ b/src/olympia/abuse/tests/test_admin.py
@@ -522,7 +522,7 @@ class TestAbuse(TestCase):
         assert 'Neo' in doc('.addon-info-and-previews h2').text()
         assert doc('.addon-info-and-previews .meta-abuse td').text() == '2'
         assert doc('.addon-info-and-previews .meta-rating td').text() == (
-            'Rated 2 out of 5 stars 1 review'
+            'Rated 2 out of 5 stars 1 review(s)'
         )
         assert doc('.addon-info-and-previews .last-approval-date td').text()
         assert doc('.reports-and-ratings')
--- a/src/olympia/reviewers/templates/reviewers/addon_details_box.html
+++ b/src/olympia/reviewers/templates/reviewers/addon_details_box.html
@@ -22,23 +22,22 @@
       <p{{ addon.summary|locale_html }}>{{ addon.summary|nl2br }}</p>
       <table itemscope itemtype="http://data-vocabulary.org/Review-aggregate">
         {% with addon_name='<span itemprop="itemreviewed">{0}</span>'|format_html(addon.name) %}
-          {# L10n: This is a caption for a table. {0} is an add-on name. #}
-          <caption>{{ _('Add-on Information for {0}')|format_html(addon_name) }}</caption>
+          <caption>Add-on Information for {{ addon_name }}</caption>
         {% endwith %}
         <tbody>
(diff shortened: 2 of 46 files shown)


src/olympia/abuse/tests/test_admin.py
```diff
@@ -522,7 +522,7 @@ def test_detail_addon_report(self):
         assert 'Neo' in doc('.addon-info-and-previews h2').text()
         assert doc('.addon-info-and-previews .meta-abuse td').text() == '2'
         assert doc('.addon-info-and-previews .meta-rating td').text() == (
-            'Rated 2 out of 5 stars 1 review'
+            'Rated 2 out of 5 stars 1 review(s)'
```
bobsilverberg, line 525: So removing the translation also removed the logic that would make this "review" instead of "review(s)"?

src/olympia/reviewers/templates/reviewers/addon_details_box.html
```diff
@@ -87,27 +86,26 @@ <h2>{{ addon_name }}</h2>
...
+            <th>Rating</th>
             <td>
               <p class="addon-rating">
                 {% with num=addon.total_ratings %}
                   {% if num %}
                   {{ addon.average_rating|float|stars }}
                   {% with count='<span itemprop="count">{0}</span>'|format_html(num|numberfmt) %}
-                    {# Using num=count so we don't change an L10n string. #}
-                    {% set review_url_text = ngettext('{num} review', '{num} reviews', num)|format_html(num=count) %}
+                    {% set review_url_text = '{num} review(s)'|format_html(num=count) %}
```
bobsilverberg, line 108: We decided we don't need "review"/"reviews" vs "review(s)"?

bobsilverberg (approved):
Looks good. After looking through all of it I am a bit unsure about why in some cases you kept the `with` (and added an `endwith` where you removed the `blocktrans`), but in other cases it seemed like you could remove both the `with` and the `blocktrans`. It might be worth taking a quick look at the former cases to see if we actually still need the `with`/`endwith`.

r+wc

src/olympia/abuse/tests/test_admin.py
diox: Yes, because we no longer have a `ngettext`

src/olympia/reviewers/templates/reviewers/addon_details_box.html
diox: I don't think it's worth it in reviewer tools

diox:
> Looks good. After looking through all of it I am a bit unsure about why in some cases you kept the `with` (and added an `endwith` where you removed the `blocktrans`), but in other cases it seemed like you could remove both the `with` and the `blocktrans`. It might be worth taking a quick look at the former cases to see if we actually still need the `with`/`endwith`.

Yes, that's because I started doing it manually, then realized I could just make a search&replace with `with` and simplify the diff. I filed a follow-up to take care of them: https://github.com/mozilla/addons-server/issues/20215

Issue #20235, opened 2023-01-24
Improve links between admins


wagnerand:
On the add-on detail admin, we have a `Files` section, but no way to access the version admin. The add-on detail admin should show a list of versions, along with the version id, file id, creation date of the version, version number, channel, file status (dropdown as is currently), block status, deleted (read-only), needs_human_review (read-write). 

The version id and file id columns should link to the respective admin detail pages.

diox:
@wagnerand Have you ever used the `Recalc Hash` link btw ? I think these days it's completely obsolete.

wagnerand:
> @wagnerand Have you ever used the `Recalc Hash` link btw ? I think these days it's completely obsolete.

Not in years, it's also why I didn't list it above. I think we can remove that functionality, especially if it's obsolete.

diox:
> needs_human_review (read-write)

Because that flag lives on the version, not the file, it's a bit annoying to allow modifying it there. Is that shortcut worth it for you ? It's already possible to modify it on the version admin, and we'd link to that. I'd rather not build code specifically for this small use case if we can avoid it.

wagnerand:
> > needs_human_review (read-write)
> 
> Because that flag lives on the version, not the file, it's a bit annoying to allow modifying it there. Is that shortcut worth it for you ? It's already possible to modify it on the version admin, and we'd link to that. I'd rather not build code specifically for this small use case if we can avoid it.

Ok, let's omit that for now, then. I thought it would be the same complexity as the file status, which we already support changing here.

diox:
Yeah unfortunately status is already on the file model so we get it for free, but for versions fields we need extra specific work. In the glorious distant future where the 2 models are merged this won't be a problem...

diox:
Behold my UX sk1llz:
![Screenshot 2023-01-25 at 13-33-59 115 casaur-gultur-raccock Change addon Django site admin](https://user-images.githubusercontent.com/187006/214564583-cb009d5c-e969-454e-9e42-bc959864a7f5.png)

AlexandraMoga:
I've verified this issue on AMO -dev and the requested fields/functionalities have been added to the addon admin model:

![image](https://user-images.githubusercontent.com/31961530/215486606-d2490169-8e28-437f-bc4a-2c1b2542616b.png)

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-224

PR #20244, opened 2023-01-25
Add links and more useful information to inline files in addon admin

diox:
Fixes https://github.com/mozilla/addons-server/issues/20235


src/olympia/addons/admin.py
```diff
@@ -103,12 +117,11 @@ def version__is_blocked(self, obj):
 
     version__is_blocked.short_description = 'Block status'
 
-    def hash_link(self, obj):
-        url = reverse('zadmin.recalc_hash', args=(obj.id,))
-        template = '<a href="{}" class="recalc" title="{}">Recalc Hash</a>'
-        return format_html(template, url, obj.hash)
```
eviljeff, line 109: If this is the only place `recalc_hash` is referenced (and I think it is) we should remove that view + associated tests and js

eviljeff (approved):

src/olympia/addons/admin.py
diox: Removed in https://github.com/mozilla/addons-server/pull/20244/commits/f49220dcb0fc74a9893096119c8d37e57473366d

PR #20246, opened 2023-01-25
Tweak ActivityLog to make activity admin pages more useful

### LINKED ISSUE #20245: Improve activity log admin
The activity log admin listing page should show creation date, user, action and arguments (ideally denormalized), especially when the action is about a particular class we have more information about, like addons, versions, ratings etc.

The activity log admin details page should should show the same with the additional fields provided by the type of the activity log, e.g. addons, versions, IP etc.

diox:
Fixes #20245


src/olympia/activity/admin.py
```diff
@@ -2,30 +2,35 @@
...
+        'pretty_arguments',
+        'ip_address',
     )
     date_hierarchy = 'created'
     fields = (
         'user',
         'created',
-        '__str__',
+        'pretty_arguments',
+        'ip_address',
```
diox, line 29: TODO: while we're at it, also display if it's an activity we are going to keep forever or not.

src/olympia/activity/models.py
```diff
@@ -577,13 +570,18 @@ def to_string(self, type_=None):
...
-                file_ = self.f(
+                file_ = format_html(
                     '<a href="{0}">{1}</a> (validation {2})',
-                    arg.get_absolute_url(),
+                    get_absolute_url(arg),
                     arg.pretty_filename,
                     validation,
                 )
                 arguments.remove(arg)
+            if isinstance(arg, UserProfile) and not user:
```
diox, line 580: `{user}` in format strings used to be the user that the `ActivityLog` was attached to. In other words, the user that did the activity. But in several cases the format string was incorrect, we expected it to be the target of the activity, which is not the same thing. For example, for `ADMIN_USER_EDITED`, `{user}` was the admin, and the user being edited was actually `{0}`.

Since this was super confusing and not aligned with the way the rest of the arguments are handled, instead I made `{user}` be treated like a regular argument, introduced `{user_responsible}` which is the user that did the activity, and fixed all the format strings.

src/olympia/activity/admin.py
```diff
@@ -38,9 +46,33 @@ def has_add_permission(self, request):
     def has_delete_permission(self, request, obj=None):
         return False
 
+    @admin.display(description='Arguments')
+    def pretty_arguments(self, obj):
+        return obj.to_string(type_='admin')
```
diox, line 51: TODO: Need to add a test proving that this works and that it doesn't escapes the HTML content

diox: https://github.com/mozilla/addons-server/pull/20246/commits/314ba482fcef817940fd285a15116ee40fcc9361

src/olympia/constants/activity.py
```diff
@@ -119,7 +120,7 @@ class DELETE_VERSION(_LOG):
 class ADD_FILE_TO_VERSION(_LOG):
     id = 19
     action_class = 'add'
-    format = _('File {0.name} added to {version} of {addon}.')
+    format = _('File {file} added to {version} of {addon}.')
```
eviljeff, line 123: This looks to be a File rather than a UserProfile instance?

eviljeff (approved):
r+wc

src/olympia/constants/activity.py
diox: Yes, but the problem was similar. `{0}` didn't exist, the file argument had already be consumed into `{file}`

PR #20247, opened 2023-01-25
Allow sorting by IP addresses in the admin

### LINKED ISSUE #20230: Allow sorting by IP address in ratings admin
In the ratings admin, the _IP address_ column should be sort be sortable. ~If possible, there should also be a way to group ratings by IP address, including a count of rating per IP.~

diox:
Fixes #20230

### DIFF
--- a/src/olympia/ratings/tests/test_admin.py
+++ b/src/olympia/ratings/tests/test_admin.py
@@ -257,6 +257,30 @@ class TestRatingAdmin(TestCase):
         assert addon.guid in addon_filter_options_text
         assert second_addon.guid in addon_filter_options_text
 
+        # Sort by IP address using django admin built-in sort:
+        # parameter is `o`, value is -5.3 because we're sorting by IP (the 5th
+        # column) desc and then created (3rd column) asc.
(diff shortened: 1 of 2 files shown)


src/olympia/ratings/tests/test_admin.py
```diff
@@ -257,6 +257,30 @@ def test_search_by_ip(self):
         assert addon.guid in addon_filter_options_text
         assert second_addon.guid in addon_filter_options_text
 
+        # Sort by IP address using django admin built-in sort:
+        # parameter is `o`, value is -5.3 because we're sorting by IP (the 5th
```
eviljeff, line 261: Wow, that's some opaque data formatting for the sort parameter.

eviljeff (approved):

src/olympia/ratings/tests/test_admin.py
diox: Yeah... My guess is that's a super old legacy thing that never got changed... It's weird that it's not stable if columns ordering or contents change.

PR #20253, opened 2023-01-26
Avoid expensive annotations in admin where possible

### LINKED ISSUE #20248: Django admin add-ons changelist page times out
On prod, the django admins addons changelist page loads for a minute, then times out with a 504 error.

diox:
This adds a separate method available on model admins to add annotations to the queryset that won't be used when computing the `count(*)` for a changelist.

This also avoids annotating the queryset with activity and IP information (avoiding expensive joins) unless necessary for search/display purposes.

Fixes #20248

### DIFF
--- a/src/olympia/addons/admin.py
+++ b/src/olympia/addons/admin.py
@@ -5,6 +5,7 @@ from django import http, forms
 from django.conf import settings
 from django.contrib import admin
 from django.core import validators
+from django.db.models import Exists, OuterRef
 from django.forms.models import modelformset_factory
 from django.http.response import (
     HttpResponseForbidden,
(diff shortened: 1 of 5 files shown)


src/olympia/addons/admin.py
```diff
@@ -258,44 +259,25 @@ class Media(AMOModelAdmin.Media):
...
-        #     _listed_versions_exists=Exists(
-        #         sub_qs.filter(channel=amo.CHANNEL_LISTED)
-        #     ),
-        # )
-        # But while this works, the subquery is a lot less optimized (it does a full
-        # query instead of the SELECT 1 ... LIMIT 1) and to make things worse django
-        # admin doesn't know it's only for displayed data (it doesn't realize we aren't
-        # filtering on it, and even if it did can't remove the annotations from the
-        # queryset anyway) so it uses it for the count() queries as well, making them a
-        # lot slower.
```
diox, line 279: That comment was obsolete: not only does Django optimizes the subquery nowadays, it was also executing the `extra()` because the way `.count()` works has changed since that comment was written.

Which means there was no reason to do it without using annotations anymore. Moving it to the newly created `get_queryset_annotations()` allows us to avoid it for the `COUNT(*)` query.

eviljeff (approved):

src/olympia/scanners/tests/test_admin.py
```diff
@@ -263,7 +263,7 @@ def test_list_queries(self):
         deleted_addon.delete()
 
         with self.assertNumQueries(13):
-            # 14 queries:
+            # 13 queries:
```
eviljeff, line 266: The breakdown below needs updating (?)

diox: Nah, I checked it, it was correct, this comment was just out of date.

PR #20261, opened 2023-01-27
Don't open new tab for false positive scanner result report

### LINKED ISSUE #20260: Don't open new tab when reporting a scanner result as false positive

1. Have a submission being flagged by a scanner.
2. On a scanner result page in django admin, click "False positive" for a result.

### What happened?

The same page opens in a new tab.

No new tab should be opened (and the report should still be applied correctly).

wagnerand:
Fix #20260.

### DIFF
--- a/static/js/scanners/scannerresult_actions.js
+++ b/static/js/scanners/scannerresult_actions.js
@@ -1,7 +1,0 @@
-document.querySelectorAll('.report-false-positive-button').forEach((button) => {
-  button.addEventListener('click', () => {
-    setTimeout(() => {
-      window.location.reload();
-    }, 2000); // 2 seconds
-  });
-});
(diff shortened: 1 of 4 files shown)


static/js/scanners/scannerresult_actions.js
```diff
@@ -1,7 +0,0 @@
-document.querySelectorAll('.report-false-positive-button').forEach((button) => {
-  button.addEventListener('click', () => {
-    setTimeout(() => {
-      window.location.reload();
```
eviljeff, line 4: I've not reviewed this code before, but isn't this js to allow the page to reflect the updated result?  Is that not needed now too?  (the issue didn't mention about page reloads)

wagnerand: The patch is changing the form target from "_blank" (opening a new tab) "_self"  (default, unspecified), which will cause the original tab to be reloaded.

eviljeff (approved):

PR #20267, opened 2023-01-30
Add flag to delay auto-approval for unlisted, set it at version creation

### LINKED ISSUE #20250: Unlisted submission flag
For new unlisted add-ons, we want to set an internal flag during submission.

If that flag is set, we delay auto-approval for X hours from the creation of the add-on.
The flag should be exposed as a filter on the django admin addons changelist. The corresponding admin detail page should have a way to set or unset the flag.

X should be a zadmin-config.

diox:
Fixes #20250


src/olympia/reviewers/models.py
```diff
@@ -565,29 +565,28 @@ def check_has_auto_approval_disabled(cls, version):
...
-            addon.auto_approval_delayed_until
-            and datetime.now() < addon.auto_approval_delayed_until
+        flag_suffix = '_unlisted' if version.channel == amo.CHANNEL_UNLISTED else ''
+        auto_approval_disabled = bool(
+            getattr(addon, f'auto_approval_disabled{flag_suffix}')
+        )
+        auto_approval_disabled_until_next_approval = getattr(
+            addon, f'auto_approval_disabled_until_next_approval{flag_suffix}'
+        )
+        auto_approval_delayed = getattr(
```
bobsilverberg, line 582: Nit: This sounds like a boolean to me, but I think it's actually a date, so maybe change the name to make that more obvious?

src/olympia/addons/admin.py
```diff
@@ -38,6 +39,14 @@
 log = olympia.core.logger.getLogger('z.addons.admin')
 
 
+class AddonReviewerFlagsInline(admin.TabularInline):
+    model = AddonReviewerFlags
+    # fields = ('pending_rejection', 'needs_human_review_by_mad')
```
bobsilverberg, line 44: Does this comment serve a purpose?

bobsilverberg (approved):
Just a nit and a question.

r+wc

src/olympia/addons/admin.py
diox: No, that was a copy/paste I need to remove - we don't need to define `fields` as we want to show all (non readonly) fields on the model anyway.

diox: https://github.com/mozilla/addons-server/pull/20267/commits/ea686f344a16818220baf5834366ad3e73fbf4c2

src/olympia/reviewers/models.py
diox: https://github.com/mozilla/addons-server/pull/20267/commits/4060db1d8815831942342dc3bef32bfe7522c835 and https://github.com/mozilla/addons-server/pull/20267/commits/21b46d320e16984622d581258a4331410c2bf96d should help a bit

PR #20280, opened 2023-01-31
Exclude versions with due date from Needs Human Review queue

### LINKED ISSUE #20279: Exclude versions with a due date from the Needs human review queue
With https://github.com/mozilla/addons-server/pull/20252 when a `needs_human_review` flag is set, it will set a due date, so in the future versions needing human review will just be in the manual review queue. Existing versions however, don't have a due date and we decided to not do a migration for those and instead have reviewers clear the existing needs human review queue. To help them do that we should temporarily add a filter to exclude versions with a due date from the needs human review queue.

Once the queue is cleared, we'll then remove it completely in https://github.com/mozilla/addons-server/issues/20238.

wagnerand:
Fix #20279.

### DIFF
--- a/src/olympia/addons/tests/test_models.py
+++ b/src/olympia/addons/tests/test_models.py
@@ -3031,14 +3031,17 @@ class TestExtensionsQueues(TestCase):
         assert all(addon in addons for addon in expected_addons.values())
         assert not any(addon in addons for addon in unexpected_addons.values())
 
-    def test_scanners_queue(self):
-        expected_addons = {'flagged_addon': addon_factory()}
+    def test_human_review_queue(self):
+        expected_addons = {
+            'flagged_addon': addon_factory(
+                version_kw={'needs_human_review': True, 'due_date': None}
+            )
+        }
         unexpected_addons = {}
-        version_factory(addon=expected_addons['flagged_addon'], needs_human_review=True)
         unexpected_addons['other_addon'] = addon_factory()
         version_factory(addon=unexpected_addons['other_addon'])
         unexpected_addons['addon_pending_rejection'] = addon_factory(
-            version_kw={'needs_human_review': True}
+            version_kw={'needs_human_review': True, 'due_date': None}
         )
         version_review_flags_factory(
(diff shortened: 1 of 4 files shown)


bobsilverberg (approved):
The code looks fine to me, but I have not really been keeping on top of all of these conversations about these requirements, so perhaps @diox or @eviljeff should take a look as well?

src/olympia/addons/tests/test_models.py
```diff
@@ -3178,6 +3178,19 @@ def test_pending_queue(self):
         addons = Addon.objects.get_pending_review_queue(admin_reviewer=True)
         assert set(addons) == set(expected_addons)
 
+    def test_human_review_queue_entries_have_no_due_date(self):
+        no_due_date = addon_factory(version_kw={'needs_human_review': True})
```
diox, line 3182: If you add an assertion I think you'll find that it does have a due date - because the `needs_human_review` forces it. You'd have to explicitly pass `'due_date': None` for that not to happen. I'm surprised the test is passing...

diox: Your branch is missing https://github.com/mozilla/addons-server/commit/d960b4e3a7e0044e8d8559691ef84019dc3dc107 which has the logic to set `due_date` if `needs_human_review` is also set, which is why your test is passing despite not doing what I suggest. I recommend updating it with `master` and fixing the test before merging, just in case.

wagnerand: Is that because of https://github.com/mozilla/addons-server/blob/60fdf84e050974524443979145ca6ea20a0ea362/src/olympia/versions/models.py#L1081-L1090? I am happy to update the test, the fact that it is passing without it is concerning? How can we make sure we're testing the right thing?

wagnerand: Yep, that was it, and adding an explicit `None` due date makes it pass. Thanks!

wagnerand:
hm, and this now unhid a bunch of other failures :(

diox (approved):

PR #20281, opened 2023-02-01
replace File|Version.reviewed with approval_date and human_review_date

### LINKED ISSUE #20147: Clean up `Version.reviewed` and `File.reviewed`
Both the `Version` and the `File` model have a `reviewed` field, with the logic of neither being really clear and likely not consistently applied.
We should clean that logic up or remove those fields and start with a clean approach.

Eventually, we might want to have a completion date field or something that stores when a review was completed. The activity log might already do this in some form, but that is not trivial to analyze, so we might want to consider an extra field (probably attached to `Version`?) for easier querying in reports and data analysis.

eviljeff:
fixes #20147 - the replacement fields are not quite the same (because otherwise why would we bother):
- Version.human_review_date is _only_ for when a version has been reviewed by a reviewer
- File.approval_date is set when the file is approved (and signed, if not already signed)
- if both fields are `None` then the version/file hasn't been reviewed or auto_approved

Does this seem like it fits the issue? @wagnerand  It turned out to be quite a complicated set of changes.

What's missing is a migration for existing data.  I'm not sure if we want to (the issue did say "a clean approach") but if we do then we could probably backfill `human_review_date` with similar code to `has_been_human_reviewed`, and probably backfill `approval_date` if the file is `status.AMO_APPROVED`.

(I did think after writing this that maybe we don't need `approval_date` at all?)


wagnerand:
> Does this seem like it fits the issue? @wagnerand It turned out to be quite a complicated set of changes.

I didn't look at the code but the description sounds good to me.

wagnerand:
> Version.human_review_date is only for when a version has been reviewed by a reviewer

I assumed "reviewed" here means any of approved, confirmed-approval, rejected, delay-rejected?

eviljeff:
> > Version.human_review_date is only for when a version has been reviewed by a reviewer
> 
> I assumed "reviewed" here means any of approved, confirmed-approval, rejected, delay-rejected?

I basically replaced anywhere that set `version.reviewed`, and set `human_review_date` instead (when `human_review is True`).  Currently it set `reviewed` when an unlisted version has a reviewer_reply and comment so I did that too (I guess that's still desired?)
So, yes to approved and rejected; it's set too for delayed rejection; I missed out the confirm actions so I'll add those.

wagnerand:
>  Currently it set reviewed when an unlisted version has a reviewer_reply and comment so I did that too (I guess that's still desired?)

Let's delete those.

src/olympia/addons/serializers.py
```diff
@@ -322,6 +323,11 @@ def to_representation(self, instance):
             repr['files'] = [repr.pop('file')]
         return repr
 
+    def get_reviewed(self, instance):
+        return serializers.DateTimeField().to_representation(
+            instance.file.approval_date or instance.human_review_date
+        )
```
diox, line 329: Do we want to potentially expose the human review date ? Could that reveal information that wasn't available before ? (either publicly or through API token from a developer)

wagnerand: I don't think that would be necessary at this point. What matters more to developers is the approval/signing date, which I think is public already?

eviljeff: It should return essentially the same value as `reviewed` was - if it was auto approved it'd be date that happened; otherwise it'd be when it was manually rejected.

I'm not sure we need to expose either date but it was already exposed as `reviewed` and this avoids an API change.

diox (approved):

PR #20297, opened 2023-02-06
Show & Allow removing Unlisted Auto-Approval Extra Delay in review tools

### LINKED ISSUE #20251: Show auto approval delay unlisted flag in reviewer tools
Followup from #20250.

- Add-ons with temporary (not infinite) auto-approval delay (unlisted or listed) should only be shown in the Manual Review queue if the user has `Addons:TriageDelayed` permission
- Such add-ons should be shown in the queue with a distinct background and have a specific flag
- Bonus: if a version shown in the queue for an add-on is marked as needing human review or auto-approval explicitly and permanently disabled for a channel, show a specific flag
- It should be possible for users with `Reviews:Admin` to clear the unlisted auto-approval delay like it is for listed

**Edited** by @diox for clarity.

diox:
Also refactor how flags are displayed, adding a couple new ones and hiding the ones that are not relevant to the version shown in the review page because of the channel.

Fixes #20251


src/olympia/addons/models.py
```diff
@@ -1652,18 +1675,23 @@ def auto_approval_delayed_until_unlisted(self):
...
             and self.auto_approval_delayed_until > datetime.now()
-        ) or (
+        )
+
+    @property
+    def auto_approval_delayed_indefinitely_unlisted(self):
+        return self.auto_approval_delayed_until_unlisted == datetime.max
+
+    @property
+    def auto_approval_delayed_temporarily_unlisted(self):
```
eviljeff, line 1697: Random thought for another time - we should consider changing this kind of `@property`s to a function that takes a channel parameter so we don't need duplicates when we have additional channels than listed and unlisted.

eviljeff (approved):

src/olympia/addons/models.py
diox: It's a little annoying because it's directly a "flag" for reviewers `get_flag()` function, which generates css classes...

eviljeff (approved):

src/olympia/addons/tests/test_models.py
```diff
@@ -3173,8 +3173,10 @@ def test_pending_queue(self):
                     ),
                     file_kw={'status': amo.STATUS_AWAITING_REVIEW},
                     channel=amo.CHANNEL_UNLISTED,
+                    due_date=datetime.now() + timedelta(hours=24),
                 ).addon,
                 file_kw={'status': amo.STATUS_AWAITING_REVIEW},
+                due_date=datetime.now() + timedelta(hours=48),
```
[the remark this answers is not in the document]
diox: It's to force a specific ordering since I'm testing which version is returned as the first pending for each add-on.

PR #20299, opened 2023-02-06
still show deleted add-ons and versions in review queues

### LINKED ISSUE #20256: Deleted add-ons and versions should show up in the Manual Review queue
Per title.

Change `objects` to `unfiltered` in:
https://github.com/mozilla/addons-server/blob/d821a0bb86094ee2f2fed7ac5e5c125d0caa8978/src/olympia/reviewers/utils.py#L161
https://github.com/mozilla/addons-server/blob/d821a0bb86094ee2f2fed7ac5e5c125d0caa8978/src/olympia/addons/models.py#L292

eviljeff:
fixes #20256 - after a lot of debugging, it turned out to need a few more subtle changes than the 3 places mentioned.  

Andreas in other issues said that we only want the `due_date` for deleted add-ons that have already been signed, so I added that restriction at the same time.

### DIFF
--- a/src/olympia/versions/models.py
+++ b/src/olympia/versions/models.py
@@ -196,6 +196,7 @@ class VersionManager(ManagerBase):
         )
         is_pre_review_version = Q(
             Q(file__status=amo.STATUS_AWAITING_REVIEW)
+            & ~Q(addon__status=amo.STATUS_DELETED)
             & Q(reviewerflags__pending_rejection__isnull=True)
             & Q(
                 is_theme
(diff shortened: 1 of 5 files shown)


src/olympia/versions/models.py
```diff
@@ -196,16 +196,18 @@ def should_have_due_date(self, negate=False):
         )
         is_pre_review_version = Q(
             Q(file__status=amo.STATUS_AWAITING_REVIEW)
+            & ~Q(addon__status=amo.STATUS_DELETED)
```
eviljeff, line 199: this _shouldn't_ be necessary as deleting an add-on should set the file statuses to disabled but it's clearer that the `is_pre_review_version` `Q` is restricted to non-deleted add-ons.

eviljeff:
(rebased to fix merge conflicts)

diox (approved):

Issue #20236, opened 2023-01-24
Store and show a reason why a restriction was added


wagnerand:
We have different kinds of restrictions. Many of those can get added manually or automatically by some other action. We should store why a particular restriction was added and show that in the restriction admin.

diox:
Restriction classes should all inherit from the base `RestrictionAbstractBaseModel` abstract class, so it should be just a matter of:
- Adding a `CharField` there
- Generating migrations
- Adding that field to the admins

Then, at the moment I believe the only code that automatically adds restrictions is the `_delay_auto_approval_indefinitely_and_restrict` scanner action (and the related `_delay_auto_approval_indefinitely_and_restrict_future_approvals` which calls that function). Modify that to say something like `Automatically added for scanner rule {rule} on {addon}`.

ioanarusiczki:
@diox

On -dev I've noticed that the disposable domain, email and Ip network restrictions have a "reason" field. 

As an example, I added some text for every module. That's for the [domain restriction](https://addons-internal.dev.mozaws.net/en-US/admin/models/users/disposableemaildomainrestriction/). 

If there's anything else I should be checking, let me know.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-225

PR #20327, opened 2023-02-09
Add reason to user restrictions stored in database

diox:
Fixes https://github.com/mozilla/addons-server/issues/20236

- Automatically fill the reason when a scanner action adds a restriction.
- Add simple hack to make the text inputs larger on the restriction admins


eviljeff (approved):
Should we backfill something for existing restrictions?

diox:
The problem with backfilling is that we don't know: they may have set or modified by admins. We could try to infer this from activity logs posted at the same time, or `LogEntry` from the django admin, but I fear this would be inaccurate in too many cases.

PR #20329, opened 2023-02-10
command to return folders with (potentially) changed files instead

### LINKED ISSUE #20289: Please provide an API endpoint or a command that returns a list of changed files for the past X hours
When we sync files between AWS EFS and GCP Filestore, the tool GCP provides (i.e. the file transfer agent) takes more than a day to finish scanning the whole directory tree. As a result, the delta sync between the two filesystems takes that amount of time to finish. This is not useful in reality because we would like to finish the delta sync within the migration window.

What GCP proposes is to generate a "manifest" [[1](https://cloud.google.com/storage-transfer/docs/manifest)] and provide that to the file transfer service to skip the directory tree scanning step.

Thus this ticket is to request to provide an API endpoint or a command that returns a definitive list of changed files for the past X hours, and once we get the list of files, we'll feed that to the file transfer tool and have it transfer all the files.

Another idea is to return a definitive list of changed sub-directories, and we can use the `rsync` command to sync those. This way if files are removed from the source, `rsync` knows remove them from the destination automatically. At the moment, I'm not sure if the GCP file transfer tool supports this or not, but it's worth testing it out.

eviljeff:
fixes #20289 ?  I'm not 100% on if this what we need, or if directories are going to be useful (rather than individual files) but first pass on solving the issue.

### DIFF
--- a/src/olympia/amo/management/commands/get_changed_files.py
+++ b/src/olympia/amo/management/commands/get_changed_files.py
@@ -1,0 +1,158 @@
+import os
+
+from datetime import datetime, timedelta
+from os import scandir
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+import olympia.core.logger
+
+from olympia.addons.models import Addon, Preview
+from olympia.amo.utils import id_to_path
+from olympia.blocklist.utils import datetime_to_ts
+from olympia.files.models import File
+from olympia.hero.models import PrimaryHeroImage
+from olympia.users.models import UserProfile
+from olympia.versions.models import Version, VersionPreview
+
+
+def get_modified_folders(since, *, path, manager, datefield, id_to_path_breadth):
+    qs = manager.filter(**{f'{datefield}__gt': since}).values_list('id', flat=True)
+    return [
+        os.path.join(path, id_to_path(id_, breadth=id_to_path_breadth), '')
+        for id_ in qs
+    ]
+
+
+def collect_user_pics(since):
+    return get_modified_folders(
+        since,
+        path=os.path.join(settings.MEDIA_ROOT, 'userpics'),
+        manager=UserProfile.objects,
+        datefield='modified',
+        id_to_path_breadth=2,
+    )
+
+
+def collect_files(since):
+    path = settings.ADDONS_PATH
+    qs = File.objects.filter(modified__gt=since).values_list(
+        'version__addon_id', flat=True
+    )
+    return list({os.path.join(path, id_to_path(id_, breadth=2), '') for id_ in qs})
+
+
+def collect_sources(since):
+    return get_modified_folders(
+        since,
+        path=os.path.join(settings.MEDIA_ROOT, 'version_source'),
+        manager=Version.unfiltered,
+        datefield='modified',
+        id_to_path_breadth=1,
+    )
+
+
+def collect_xpi_uploads(since):
+    # all the files are in a single folder so return it
+    return [os.path.join(settings.ADDONS_PATH, 'temp')]
+
+
+def get_previews(since, path, manager):
+    out = set()
+    qs = manager.filter(created__gt=since).values_list('id', flat=True)
+    for preview_id in qs:
+        subdir = str(preview_id // 1000)
+        out = out | {
+            os.path.join(path, 'thumbs', subdir, ''),
+            os.path.join(path, 'full', subdir, ''),
+            os.path.join(path, 'original', subdir, ''),
+        }
+    return list(out)
+
+
+def collect_addon_previews(since):
+    return get_previews(
+        since, os.path.join(settings.MEDIA_ROOT, 'previews'), Preview.objects
+    )
+
+
+def collect_theme_previews(since):
+    return get_previews(
+        since,
+        os.path.join(settings.MEDIA_ROOT, 'version-previews'),
+        VersionPreview.objects,
+    )
+
+
+def collect_addon_icons(since):
+    path = os.path.join(settings.MEDIA_ROOT, 'addon_icons')
+    qs = Addon.unfiltered.filter(modified__gt=since).values_list('id', flat=True)
+    return list({os.path.join(path, str(preview_id // 1000), '') for preview_id in qs})
+
+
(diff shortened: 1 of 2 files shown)


src/olympia/amo/management/commands/get_changed_files.py
```diff
@@ -0,0 +1,158 @@
...
+        if file_.is_dir() and file_.name.isdigit() and int(file_.name) >= since_ts
+    ]
+
+
+class Command(BaseCommand):
+    help = (
+        'Get folders containing files that have changed on the filesystem in the past '
+        'X seconds'
+    )
+    log = olympia.core.logger.getLogger('z.appversions')
```
diox, line 134: Wrong logger name

src/olympia/amo/management/commands/get_changed_files.py
```diff
@@ -0,0 +1,158 @@
...
+def collect_editoral(since):
+    return (
+        [os.path.join(settings.MEDIA_ROOT, 'hero-featured-image')]
+        if PrimaryHeroImage.objects.filter(modified__gt=since).exists()
+        else []
+    )
+
+
+def collect_sitemaps(since):
+    return [settings.SITEMAP_STORAGE_PATH]  # sitemaps change each day
```
diox, line 103: We should probably skip them for now and just re-create them on the other side when doing the migration.

src/olympia/amo/management/commands/get_changed_files.py
```diff
@@ -0,0 +1,158 @@
...
+        path=os.path.join(settings.MEDIA_ROOT, 'version_source'),
+        manager=Version.unfiltered,
+        datefield='modified',
+        id_to_path_breadth=1,
+    )
+
+
+def collect_xpi_uploads(since):
+    # all the files are in a single folder so return it
+    return [os.path.join(settings.ADDONS_PATH, 'temp')]
```
diox, line 58: I don't think we should migrate `FileUpload`s. They are temporary content, we'll have to put the site in read-only for the duration of the migration, so we probably don't care about them.

diox (commented):
In general, while I think it's a good idea not to return full instances for every type of object, I'm not super happy about the duplication of the code that determines the folder paths. 

Do you think we could go for a middle ground approach and, for each model, instantiating a single dummy instance and using it to determine the directory for each `id` returned ? (using `dirname` or similar for objects that don't directly expose the directory)

eviljeff:
> Do you think we could go for a middle ground approach and, for each model, instantiating a single dummy instance and using it to determine the directory for each `id` returned ? (using `dirname` or similar for objects that don't directly expose the directory)

For models that have a `get_file_path` method or similar, yes.  Models that use a django FileField (e.g. File and Version) store the full path once the file field is set, so we'll have to build the path using the callback and storage functions.

eviljeff:
> > Do you think we could go for a middle ground approach and, for each model, instantiating a single dummy instance and using it to determine the directory for each `id` returned ? (using `dirname` or similar for objects that don't directly expose the directory)
> 
> For models that have a `get_file_path` method or similar, yes. Models that use a django FileField (e.g. File and Version) store the full path once the file field is set, so we'll have to build the path using the callback and storage functions.

but... (for example) `files_upload_to_callback` will need the full file instance because it accesses the addon and version too, to build the filename - that we will then strip off because we only care about the directory name...

diox:
Mmmm, yeah. We could fetch the file path with a `File.objects.only('file')` then we are beginning to deal with lots of instances... Let's keep building manually for now, it should hopefully just be a one-off...

src/olympia/amo/management/commands/get_changed_files.py
```diff
@@ -0,0 +1,115 @@
...
+from olympia.files.models import File
+from olympia.git.utils import AddonGitRepository
+from olympia.hero.models import PrimaryHeroImage
+from olympia.users.models import UserProfile
+from olympia.versions.models import Version, VersionPreview
+
+
+def collect_user_pics(since):
+    qs = UserProfile.objects.filter(modified__gt=since).values_list('id', flat=True)
+    return [UserProfile(id=id_).picture_dir for id_ in qs]
```
diox, line 21: Going one step further... Since we're building new instances... Maybe we should just use `UserProfile....only('id')` instead in the queryset ? And adding some `assertNumQueries()` to make sure we are not doing extra database queries.

On top of this, we should add [`.iterator()`](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#iterator) when evaluating the queryset, to bypass queryset caching and avoid extra memory consumption.

src/olympia/amo/management/commands/get_changed_files.py
```diff
@@ -0,0 +1,115 @@
...
+
+def collect_sources(since):
+    path = os.path.join(settings.MEDIA_ROOT, 'version_source')
+    qs = Version.unfiltered.filter(modified__gt=since).values_list('id', flat=True)
+    return [os.path.join(path, id_to_path(id_, breadth=1)) for id_ in qs]
+
+
+def _get_previews(since, Model):
+    out = set()
+    qs = Model.objects.filter(created__gt=since).values_list('id', flat=True)
```
diox, line 40: Same comment as `UserProfile` above

src/olympia/amo/management/commands/get_changed_files.py
```diff
@@ -0,0 +1,115 @@
...
+def collect_addon_previews(since):
+    return _get_previews(since, Preview)
+
+
+def collect_theme_previews(since):
+    return _get_previews(since, VersionPreview)
+
+
+def collect_addon_icons(since):
+    qs = Addon.unfiltered.filter(modified__gt=since).values_list('id', flat=True)
```
diox, line 60: Same comment as `UserProfile` above

src/olympia/amo/management/commands/get_changed_files.py
eviljeff: https://github.com/mozilla/addons-server/pull/20329/commits/984f94393ce089b2d827451866aeb1ce2627b9cd

diox (approved):

Issue #20344, opened 2023-02-14
Provide arm64 docker images


wagnerand:
On Macs using the Apple Silicon arm-based architecture, Docker raises warnings for several containers about performance impact.

I can confirm that running tests on a M2 Pro are very slow. Can we provide arm64 images for Apple Silicon users?

<img width="705" alt="docker" src="https://user-images.githubusercontent.com/873258/218821946-1525c974-afd4-47b7-bba9-95fece1796cc.png">

diox:
As I commented in the pull request, the problem is that MySQL doesn't provide arm64 deb packages, and that using the Debian built-in mysql packages actually pulls `mariadb` ones, which have slight incompatibilities that we ran into in the past.

We could try using the non-deb binaries, or build the deb packages ourselves... 

An alternative to try though, is on the host, enabling the (beta) feature Rosetta for x86/amd64 emulation in docker. It should enable us to keep AMD64 images while still getting the perf improvements you seek.

![vroooooooum-docker-rosetta](https://user-images.githubusercontent.com/187006/219009824-31f68f7e-eb49-40c4-8d83-72263c51b91a.png)

wagnerand:
With that, `docker compose up -d` fails with:

```
 ⠏ memcached The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested      
 ⠏ addons-frontend The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
 ⠏ web The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested            
 ⠏ autograph The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested      
 ⠇ worker The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested         
```

Would you accept a patch that explicitly sets `platform: linux/amd64` for those images in docker-compose.yaml?

diox:
Would just setting `DOCKER_DEFAULT_PLATFORM` on the host instead work ? And more importantly, does that solve your perf issues ?

wagnerand:
I'm not sure this is desirable for several reasons: It would pull amd64 images even if arm64 images are available. Also, every contributor would have to know and set this flag on their machine. We can document it on readthedocs, but it still leaves a visibility problem.

[...]

PR #20345, opened 2023-02-14
Update Dockerfile for arm64

wagnerand:
Supports #20344.

This runs fine on my arm64 Mac, we probably also want to test this on amd64, but from what I can see, the packages should be available there as well.

Prettier auto-formatted this file, let me know if you'd like me to revert that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,8 +13,8 @@ USER root
 
 # Add support for https apt repos and gpg signed repos
 RUN apt-get update && apt-get install -y \
-        apt-transport-https              \
-        gnupg2                           \
+    apt-transport-https                  \
+    gnupg2                               \
     && rm -rf /var/lib/apt/lists/*
 # Add keys and repos for node and mysql
 COPY docker/*.gpg.key /etc/pki/gpg/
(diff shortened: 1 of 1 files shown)


diox (changes requested):
We specifically avoid `default-mysql-client` and `default-libmysqlclient-dev` because those packages pull `mariadb` which has slight incompatibilities with the official MySQL packages (that we actually ran into in production at some point).

wagnerand:
hm, https://packages.debian.org does not provide arm64 packages afaics.

diox:
Yes. That is a problem...

wagnerand:
Hm, https://bugs.mysql.com/bug.php?id=101233 is fixed.

wagnerand:
> We specifically avoid `default-mysql-client` and `default-libmysqlclient-dev` because those packages pull `mariadb` which has slight incompatibilities with the official MySQL packages (that we actually ran into in production at some point).

Iirc, prod does not use docker, so this would only affect local development environment, right? Is that already creating too much risk for incompatibilities?

diox:
Prod definitely uses docker. It doesn't use _that_ image, but I don't want to risk the inconstancy if we can avoid it - I'd much prefer merging our deployment and local dev images instead of having them grow even more different than they are now.

On Matrix I suggested an alternative to all this, which is to try the new (beta) Rosetta feature in Docker for Mac. It should give you huge performance improvements while keeping amd64-only images.

wagnerand:
Thanks.

PR #20346, opened 2023-02-14
Configure addons-server to start sending data to GA4

### LINKED ISSUE #20222: Configure addons-server to start sending data to GA4
Our plan is to run GA4 in parallel with UA until such a time as UA is discontinued. We will need to make some decisions about what data to send to GA4, but as a starting point we should set up addons-server so the default data (e.g., page views) is being sent to GA4. This issue is about doing that.

bobsilverberg:
Fixes #20222


bobsilverberg:
I spent about an hour trying to get a test to work that confirmed that the GA-related commands were also being called. I tried mocking `window.ga` (which I can understand why it wouldn't work, as `window.ga` is created in the code we are calling), as well as `window.dataLayer` (which should have worked, and debugging showed that the mock was being called), but in both cases, when I went to do the assertions, the mocks reported that they had not been called at all.

I figured that an hour was enough time to spend on that, considering that these tests didn't exist before. I did add tests for the presence of the `script` tags, and they work.

static/js/zamboni/analytics.js
```diff
@@ -42,4 +42,20 @@ if (isDoNotTrackEnabled() === false) {
...
+  newTag.src = `https://www.googletagmanager.com/gtag/js?id=${ga4Id}`;
+  firstTag.parentNode.insertBefore(newTag, firstTag);
+
+  // Set up GA4.
+  window.dataLayer = window.dataLayer || [];
+  function gtag() {
+    dataLayer.push(arguments);
+  }
+  gtag('js', new Date());
+  gtag('config', ga4Id);
```
eviljeff, line 60: out of interest, is this their code we're including here?  Just seems oddly written (obsolete `gtag` function, when we can push directly to `window.dataLayer`, and use of `arguments` instead of modern rest parameters)

eviljeff (approved):

static/js/zamboni/analytics.js
bobsilverberg: Yes, and when I was trying to get this to work for addons-frontend I discovered that it's best to just use their snippet and not try to get fancy (e.g., with rest parameters).

PR #20364, opened 2023-02-17
Prevent potential race condition in AddonReviewerFlags API

### LINKED ISSUE #20362: Potential race condition with reviewer flags API
A typical problem with django instances is the following pattern:
```
obj = Model.objects.get(...)
obj.foo = 'bar'
obj.save()
```

This triggers an `UPDATE` for *all* fields on model, even those not modified. Which means there can be a race condition between 2 things modifying different properties of the same object: one could override the changes made by the other. This is notably present in the django admin.

We have our own `.update()` method to work around that, and in some cases we call `update_or_create()` which also avoids that problem. Django also has `select_for_update()` to lock, and you can pass `update_fields` to `save` calls.

I noticed that we don't do anything special for the `AddonReviewerViewSet`'s `flags` API, which means it uses a regular `.save()` call and could suffer from this issue. Accidentally overriding reviewer flags is pretty bad, so we should add a workaround there.

See also https://github.com/encode/django-rest-framework/issues/2648 (which mentions http://chibisov.github.io/drf-extensions/docs/#partialupdateserializermixin as a possible fix), https://github.com/encode/django-rest-framework/issues/4675, https://code.djangoproject.com/ticket/27477

diox:
Fixes #20362

### DIFF
--- a/src/olympia/reviewers/serializers.py
+++ b/src/olympia/reviewers/serializers.py
@@ -54,7 +54,7 @@ class AddonReviewerFlagsSerializer(AMOModelSerializer):
         )
 
     def update(self, instance, validated_data):
-        # Only update fields that changed. Note that this only support basic
+        # Only update fields that changed. Note that this only supports basic
         # fields.
         if self.partial and instance:
             for attr, value in validated_data.items():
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/serializers.py
```diff
@@ -53,6 +53,17 @@ class Meta:
             'needs_admin_theme_review',
         )
 
+    def update(self, instance, validated_data):
+        # Only update fields that changed. Note that this only support basic
```
bobsilverberg, line 57: Nit: s/support/supports

bobsilverberg (approved):
r+ wc

PR #20366, opened 2023-02-17
Specify platform docker images where necessary

wagnerand:
Supports #20344.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,6 +22,7 @@ services:
   worker: &worker
     <<: *env
     image: addons/addons-server:latest
+    platform: linux/amd64
     command: supervisord -n -c /code/docker/supervisor-celery.conf
     user: olympia
     volumes:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -55,6 +57,8 @@ services:
 
   memcached:
     image: memcached:1.4
+    # Remove this once we upgrade to a version that provides multi-platform images
```
eviljeff, line 60: We're _only_ removing this `platform` once there's an upgrade to a new version?  Or does this apply to each service defined in this file?

eviljeff (approved):
just a question on the comment

docker-compose.yml
diox: Technically it would be true for each, but `memcached` is the one where it's most likely to happen soon because newer versions that are multi-platform already exist, where as, as we proved in the issue, it's not trivial to make that change for our images.

wagnerand: Yes, that's why I decided not to add this comment to the other sections. I can do that though if you prefer.

PR #20367, opened 2023-02-17
Expose the version used at creation/update in add-on API

### LINKED ISSUE #20365: Expose `version` field with the version used at submission in the add-on API for `POST`/`PUT` requests
The new add-on submission API doesn't expose the `version` you used at submission in its response. 

You can sort of guess it yourself by looking at either `current_version` (for listed) or `latest_unlisted_version` (for unlisted) but it's ideal, and there could be edge cases where it's not the version that was submitted (less likely for `POST`).

Let's expose during `POST` and `PUT` the version that was submitted in a `version` field.

diox:
Fixes #20365

### DIFF
--- a/docs/topics/api/addons.rst
+++ b/docs/topics/api/addons.rst
@@ -208,6 +208,7 @@ This endpoint allows you to fetch a specific add-on by id, slug or guid.
     :>json array tags: List containing the tag names set on the add-on.
     :>json string type: The :ref:`add-on type <addon-detail-type>`.
     :>json string url: The (absolute) add-on detail URL.
+    :>json object version: For create or update requests that included a ``version`` only. Object holding the :ref:`version <version-detail-object>` that was submitted.
     :>json string versions_url: The URL to the version history page for the add-on.
     :>json int weekly_downloads: The number of downloads for the add-on in the last week. Not present for lightweight themes.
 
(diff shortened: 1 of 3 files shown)


src/olympia/addons/serializers.py
```diff
@@ -1196,11 +1196,12 @@ def create(self, validated_data):
...
         addon.set_tag_list(validated_data.get('tag_list', []))
-
-        self.fields['version'].create(
+        addon.version = self.fields['version'].create(
             {**validated_data.get('version', {}), 'addon': addon}
         )
-
+        # When creating, always return the version we just created in the
+        # representation. It uses <instance>.version.
+        self.fields['version'].write_only = False
```
eviljeff, line 1204: Add a comment next to the field declaration too, where we define it as `write_only=True`?

docs/topics/api/addons.rst
```diff
@@ -208,6 +208,7 @@ This endpoint allows you to fetch a specific add-on by id, slug or guid.
     :>json array tags: List containing the tag names set on the add-on.
     :>json string type: The :ref:`add-on type <addon-detail-type>`.
     :>json string url: The (absolute) add-on detail URL.
+    :>json object version: For create or update requests that included a ``version`` only. Object holding the :ref:`version <version-detail-object>` that was submitted.
```
eviljeff, line 211:
```suggestion
    :>json object version: For create or update requests that included ``version`` only. Object holding the :ref:`version <version-detail-object>` representing :ref:`version <version-create-request>`that was submitted.
```

eviljeff (approved):

PR #20377, opened 2023-02-21
Backfill human review date more aggressively

### LINKED ISSUE #20370: Backfill human_review_date more aggressively
We introduced `human_review_date` in https://github.com/mozilla/addons-server/issues/20147 and going forward, all rejections/approvals should set it.

We backfilled the data for existing versions from `File.reviewed`. Unfortunately this isn't good enough, a bunch of rejected versions don't have a `human_review_date` right now as a result - they didn't have a `reviewed` date set before either.

We should try to backfill from the activity logs. Something like:

```
UPDATE `versions` 
INNER JOIN `addons` ON (`versions`.`addon_id` = `addons`.`id`)
INNER JOIN `files` ON (`versions`.`id` = `files`.`version_id`)
INNER JOIN `log_activity_version` ON (`versions`.`id` = `log_activity_version`.`version_id`)
INNER JOIN `log_activity` ON (`log_activity_version`.`activity_log_id` = `log_activity`.`id`)
SET `human_review_date`=`log_activity`.`created` 
WHERE (`addons`.`addontype_id` = 1
AND `files`.`status` = 5
AND `versions`.`human_review_date` IS NULL
AND `log_activity`.`action` IN (43, 148))
```

This isn't strictly accurate - in the case of delayed rejections, this would use the actual rejection date instead of the date the reviewer performed their review - but it would ensure there is a `human_review_date` set, which is important.

I originally had a filter on `addons.status = 0` because this is the most important case, but removed it to cover more cases. This should affect ~ 60K versions.

diox:
Fixes #20370


src/olympia/reviewers/migrations/0024_backfill-human-review-date-more-agressively.py
```diff
@@ -0,0 +1,43 @@
...
+INNER JOIN `log_activity`
+  ON `log_activity_version`.`activity_log_id` = `log_activity`.`id`
+SET `human_review_date`=`log_activity`.`created`
+WHERE `addons`.`addontype_id` = %s
+AND `files`.`status` = %s
+AND `versions`.`human_review_date` IS NULL
+AND `log_activity`.`action` IN (%s, %s)""",
+                    [
+                        ADDON_EXTENSION,
+                        STATUS_AWAITING_REVIEW,
```
eviljeff, line 35: Not `STATUS_DISABLED`?

eviljeff (approved):
It's updating the `versions` table so should be in the versions migrations imo.

src/olympia/reviewers/migrations/0024_backfill-human-review-date-more-agressively.py
diox: Aaaand that's why we do reviews ! Definitely meant `STATUS_DISABLED` here.

diox:
re-review request since I messed up the first time...

eviljeff (approved):
Looks reasonable...

PR #20389, opened 2023-02-22
Return a HTTP 409 for conflicts when creating add-ons/versions through the API

### LINKED ISSUE #20387: Return HTTP 409 Conflict error in add-on API when submitting a duplicate version
In the new add-on API, submitting a version that already exists returns a 400 error with `Version {version_string} was uploaded before and deleted.` or `Version {version_string} already exists.`.

It should return a 409 instead to make it easier for clients to programmatically differentiate between a duplicate submission and more general bad request errors.

diox:
Also add a few more add-on creation tests

Fixes #20387

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -20,6 +20,7 @@ from olympia.amo.validators import (
     OneOrMoreLetterOrNumberCharacterValidator,
     PreventPartialUpdateValidator,
 )
+from olympia.api.exceptions import Conflict
 from olympia.api.fields import (
     EmailTranslationField,
     LazyChoiceField,
@@ -39,7 +40,7 @@ from olympia.constants.categories import CATEGORIES_BY_ID
 from olympia.constants.promoted import PROMOTED_GROUPS, RECOMMENDED
 from olympia.core.languages import AMO_LANGUAGES
 from olympia.files.models import File, FileUpload
-from olympia.files.utils import parse_addon
+from olympia.files.utils import DuplicateAddonID, parse_addon
 from olympia.promoted.models import PromotedAddon
 from olympia.search.filters import AddonAppVersionQueryParam
 from olympia.ratings.utils import get_grouped_ratings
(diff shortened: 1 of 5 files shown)


src/olympia/addons/serializers.py
```diff
@@ -486,7 +487,10 @@ def validate_upload(self, value):
         if not own_upload or not value.valid or value.validation_timeout:
             raise exceptions.ValidationError(gettext('Upload is not valid.'))
         # Parse the file to get and validate package data with the addon.
-        self.parsed_data = parse_addon(value, addon=self.addon, user=request.user)
+        try:
+            self.parsed_data = parse_addon(value, addon=self.addon, user=request.user)
+        except DuplicateAddonID as exc:
+            raise Conflict({self.field_name: exc.messages})
```
eviljeff, line 493: I guess you need to manually wrap the `exc.message` in a dict here (exceptions from a `validate_xxx` function should already get wrapped as `{xxx: exception}`) because an `APIException` subclass is being raised rather than a `ValidationError`?

src/olympia/addons/tests/test_views.py
```diff
@@ -3395,8 +3490,8 @@ def test_duplicate_version_number_error(self):
             self.url,
             data=self.minimal_data,
         )
-        assert response.status_code == 400, response.content
-        assert response.data == {
+        assert response.status_code == 409, response.content
+        assert response.json() == {
```
eviljeff, line 3494: out of interest, why the change to `response.json()` from `response.data`? `response.data` seems to be - by far- the most common use in our tests.

eviljeff (approved):
One drawback of this change is if any api consumer is directly checking for `400` for validation errors their code will now break... but agree it's an overall good idea.

src/olympia/addons/tests/test_views.py
diox: `response.data` is the python representation of what _would_ be rendered, but can still contain objects like `ErrorDetail`. `response.json()` is the rendered json converted back to Python. The latter is easier to compare to deep dicts.

src/olympia/addons/serializers.py
diox: Yes. That is annoying, but there is no method to call to do that, the `ValidationError` conversion happens in DRF `is_valid()` code (and there is something else for authentication errors or 500 exceptions).

I have raised a non-dict, but then it wouldn't have been consistent with the way errors are typically returned.

PR #20394, opened 2023-02-23
Allow looking up versions by version number in the add-on versions API

### LINKED ISSUE #20388: Allow looking up versions by version number in the add-on versions API
The add-on versions API lets you look up versions by their `pk`, we should allow looking up by the version number (with an add-on identifier still present) too. That will make handling of conflicts (https://github.com/mozilla/addons-server/issues/20387) easier, as the clients will be able to load the existing version without going through the versions list.

diox:
Fixes #20388

### DIFF
--- a/docs/topics/api/addons.rst
+++ b/docs/topics/api/addons.rst
@@ -504,8 +504,7 @@ Version Detail
 This endpoint allows you to fetch a single version belonging to a specific add-on.
 
     .. note::
-        Version numbers without a dot (``.``) character are not supported in
-        this API URL. Use the version ``id`` instead to fetch those versions.
+        This API accepts both version ids and version numbers in the URL. If the version number passed does not contain any dot characters (``.``) it would be considered an ``id``. To avoid this and force a lookup by version number, add a ``v`` prefix to it.
 
 .. http:get:: /api/v5/addons/addon/(int:addon_id|string:addon_slug|string:addon_guid)/versions/(int:id|string:version_number)/
 
--- a/docs/topics/api/addons.rst
+++ b/docs/topics/api/addons.rst
@@ -690,7 +690,7 @@ Note: as form-data can not be nested as objects it's not possible to set ``sourc
     :reqheader Content-Type: multipart/form-data
 
 
-.. http:patch:: /api/v5/addons/addon/(int:addon_id|string:addon_slug|string:addon_guid)/versions/(int:id)/
+.. http:patch:: /api/v5/addons/addon/(int:addon_id|string:addon_slug|string:addon_guid)/versions/(int:id|string:version_number)/
 
     .. _version-sources-request-edit:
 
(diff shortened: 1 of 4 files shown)


docs/topics/api/addons.rst
```diff
@@ -503,7 +503,11 @@ Version Detail
 
 This endpoint allows you to fetch a single version belonging to a specific add-on.
 
-.. http:get:: /api/v5/addons/addon/(int:addon_id|string:addon_slug|string:addon_guid)/versions/(int:id)/
+    .. note::
+        Version numbers without a dot (``.``) character are not supported in
+        this API URL. Use the version ``id`` instead to fetch those versions.
+
+.. http:get:: /api/v5/addons/addon/(int:addon_id|string:addon_slug|string:addon_guid)/versions/(int:id|string:version_number)/
```
diox, line 510: Just realized I need to document this for `Version Edit` as well. I'll add it.

docs/topics/api/addons.rst
```diff
@@ -705,7 +709,10 @@ This endpoint allows the metadata for an existing version to be edited.
     .. note::
         This API requires :doc:`authentication <auth>`, and for the user to be an author of the add-on.
 
-.. http:patch:: /api/v5/addons/addon/(int:addon_id|string:addon_slug|string:addon_guid)/versions/(int:id)/
+    .. note::
+        This API accepts both version ids and version numbers in the URL. If the version number passed does not contain any dot characters (``.``) it would be considered an ``id``. To avoid this and force a lookup by version number, add a ``v`` prefix to it.
```
willdurand, line 713: Nit: in Firefox, we call it the version "string", not "number". https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/version

diox: Ah... let's try to be consistent... I can see a couple places in our docs where we call that `version number`, or even `version number string` that I can fix... 

OTOH that page you linked to also has a header titled `Access the version number in code` ...

willdurand: > OTOH that page you linked to also has a header titled `Access the version number in code` ...

😭

docs/topics/api/addons.rst
```diff
@@ -503,7 +503,11 @@ Version Detail
 
 This endpoint allows you to fetch a single version belonging to a specific add-on.
 
-.. http:get:: /api/v5/addons/addon/(int:addon_id|string:addon_slug|string:addon_guid)/versions/(int:id)/
+    .. note::
+        Version numbers without a dot (``.``) character are not supported in
```
bobsilverberg, line 507: Why are we allowing the `v` prefix to force the param to be recognized as a version number in other endpoints, but not in this endpoint?

bobsilverberg: Actually, looking at the code and tests, it looks like we *do* allow it here as well, so perhaps this note needs to be updated to explain that.

bobsilverberg (approved):
Looks good, just a note about the docs. Also, are the tests for the detail endpoint sufficient considering that this change also applies to other endpoints? I assume from a unit testing perspective they are, but I figured I should ask.

r+wc

docs/topics/api/addons.rst
diox: Yeah I should have updated it as well.

docs/topics/api/addons.rst
diox: https://github.com/mozilla/addons/issues/1465

diox:
> Also, are the tests for the detail endpoint sufficient considering that this change also applies to other endpoints? I assume from a unit testing perspective they are, but I figured I should ask.

Yeah I think that's enough.

PR #20396, opened 2023-02-23
add is_disabled to DeveloperVersionSerializer and update

### LINKED ISSUE #20142: Add the ability to disable versions through the API
split from https://github.com/mozilla/addons-server/issues/19784

Developers have the option to disable versions in DevHub. We should make this action available in the [version edit](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#version-edit) API as well.

eviljeff:
fixes #20142


src/olympia/addons/serializers.py
```diff
@@ -493,6 +496,25 @@ def validate_upload(self, value):
             raise Conflict({self.field_name: exc.messages})
         return value
 
+    def validate_is_disabled(self, disable):
+        # The Version.is_user_disabled setter just ignores any change it doesn't like,
+        # but we'd like to raise instead if the value wouldn't be accepted.
+        if disable and (version := self.instance):
+            if (
+                not version.is_user_disabled
```
diox, line 504: Do we need this part of the condition ? It seems fine but I don't understand why it's there.

src/olympia/addons/tests/test_views.py
```diff
@@ -4048,6 +4048,75 @@ def test_cant_upload_source_for_reviewed_file(self):
...
+        assert response.data['is_disabled'] is False
+        self.version.reload()
+        self.version.file.reload()
+        assert not self.version.is_user_disabled
+        alog = ActivityLog.objects.exclude(
+            action__in=(amo.LOG.LOG_IN.id, amo.LOG.LOG_IN_API_TOKEN.id)
+        ).get()
+        assert alog.user == self.user
+        assert alog.action == amo.LOG.ENABLE_VERSION.id
+
```
diox, line 4119: We should probably also have tests for when setting `is_user_disabled` does not modify the `file` status, to make sure we return an accurate `is_disabled` in that case.

src/olympia/addons/serializers.py
eviljeff: We do - to only raise when there is an attempt to change disabled value (i.e. it wasn't already user disabled).
I used `version.is_user_disabled` to keep the logic about whether the version is user disabled in one place but, looking now, this could be replaced with:
`if version.file.status == amo.STATUS_DISABLED and version.file.original_status == amo.STATUS_NULL:`
if that's clearer?

diox: But if the `file.status` is `amo.STATUS_DISABLED`, you shouldn't be able to flip `is_user_disabled`, right ? So that part of the condition seems superfluous.

diox: Ok, I see it now.

I was wondering about modifiying the version in general when it's `STATUS_DISABLED` already. It is currently possible in devhub, so I guess that's fine...

diox (approved):

src/olympia/addons/tests/test_views.py
eviljeff: As far as I can see that's not a possibility - you can't set `is_user_disabled` when the file is already `amo.STATUS_DISABLED` but setting `is_user_disabled` to true ends up setting file status to `amo.STATUS_DISABLED`

PR #20419, opened 2023-03-01
Store host_permissions in the webext_permissions table

### LINKED ISSUE #20417: Store host permissions in the `webext_permissions` table
MV3 has a new manifest.json key named `host_permissions` with a list of, well, host permissions. More information at: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/host_permissions. Basically, in MV2, permissions could contain both actual permission names and hosts. In MV3+, hosts are moved to `host_permissions`.

We are already parsing and storing `permissions` and `optional_permissions`. It'd be interesting to also store the host permissions. It might be handy for data exploration.

Product might even be interested in surfacing those hosts in the frontend. That would require more work in this project, though (e.g. exposing the new data in the API).

bobsilverberg:
Fixes #20417


src/olympia/files/migrations/0029_auto_20230301_1843.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('files', '0028_set_approval_date'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
bobsilverberg, line 13: `makemigrations` added this, even though my patch does not include this change. I wasn't sure if I should just keep it, because it's probably something that is missing from another migration, or if I should remove it from this file, as I didn't make this change.

bobsilverberg: Well, that broke everything on CI, so I removed it. Not sure why `makemigrations` wanted to do that.

diox: It's a dependency issue - migration `0035` in `versions` uses that field, so if you want to remove it, you need to add it as dependency of your migration to force the migration engine to execute migrations in an order that isn't broken.

We are usually deliberately not removing the fields right away from the database for backwards-compatility with existing code when we deploy, but in this case it was also useful to still have the database data laying around to backfill some more data (in fact that's what migration `0035` in `versions` did) but it's probably ok to remove now.

src/olympia/addons/models.py
```diff
@@ -1604,7 +1604,11 @@ def should_show_permissions(self, version=None):
             self.type == amo.ADDON_EXTENSION
             and version
             and version.file
-            and (version.file.permissions or version.file.optional_permissions)
+            and (
```
diox, line 1607: This entire method appears to be unused, so you can just remove it instead. (If it was used, we'd likely not want to add `host_permissions` support to it yet, since we're not displaying those permissions anywhere at the moment)

src/olympia/files/migrations/0029_auto_20230301_1843.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('files', '0028_set_approval_date'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='webextpermission',
+            name='host_permissions',
+            field=models.JSONField(default=dict),
```
diox, line 16: We usually add new columns with `null=True, default=None` for backwards-compatibility with the previous tag, because Django doesn't add database defaults in non-null cases, but it looks like it does for `JSONField`, TIL. That generates:
```
ALTER TABLE `webext_permissions` ADD COLUMN `host_permissions` json DEFAULT ('{}') NOT NULL;
```

Which _is_ backwards-compatible, so that's fine. Carry on.

src/olympia/addons/indexers.py
```diff
@@ -391,6 +391,7 @@ def get_mapping(cls):
                         'status': {'type': 'byte'},
                         'permissions': {'type': 'keyword', 'index': False},
                         'optional_permissions': {'type': 'keyword', 'index': False},
+                        'host_permissions': {'type': 'keyword', 'index': False},
```
diox, line 394: Not strictly needed since we aren't exposing that in the search API yet, but it can't hurt to start recording it in ES I guess... Will require a reindex.

diox (approved):

PR #20423, opened 2023-03-02
add add-ons without a current_version to Notable group too

### LINKED ISSUE #20422: Add listed addons with non-public version to Notable group once they cross user threshold
related to #20405 

Listed addons with non-public versions should still be added to the Notable group once the user threshold is reached. 

### What happened?
We had a case with `869387`, where an addon version got approved and crossed the threshold for notable before the addon version was listed. 
The next version they uploaded 8 days later had a due date set from the previous version, making the due date the same day as the newest version upload date. 

Due date for the new version should have been the typical week due date. 

eviljeff:
fixes #20422

~This patch (currently) doesn't change how existing versions are handled - only `Addon.current_version` is flagged with `needs_human_review` so if there isn't a current version (i.e. there aren't any listed versions that are non-deleted and not status=amo.STATUS_DISABLED) then no existing version is flagged.  But because the add-on is then Notable a subsequent new version would be held for pre-review.~

### DIFF
--- a/src/olympia/reviewers/models.py
+++ b/src/olympia/reviewers/models.py
@@ -600,13 +600,11 @@ class AutoApprovalSummary(ModelBase):
     @classmethod
     def check_is_promoted_prereview(cls, version):
         """Check whether the add-on is a promoted addon group that requires
-        pre-review.
-
-        Only applies to listed versions."""
-        if not version.channel == amo.CHANNEL_LISTED:
-            return False
-        promo_group = version.addon.promoted_group(currently_approved=False)
-        return bool(promo_group and promo_group.pre_review)
+        pre-review."""
+        return bool(
+            version.channel == amo.CHANNEL_LISTED
+            and version.addon.promoted_group(currently_approved=False).pre_review
+        )
 
     @classmethod
     def check_should_be_delayed(cls, version):
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -789,7 +789,7 @@ class ReviewBase:
 
     def set_promoted(self):
         group = self.addon.promoted_group(currently_approved=False)
-        if group and group.pre_review:
+        if self.version.channel == amo.CHANNEL_LISTED and group.pre_review:
             # These addons shouldn't be be attempted for auto approval anyway,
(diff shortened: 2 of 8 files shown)


src/olympia/reviewers/models.py
```diff
@@ -600,13 +600,11 @@ def check_has_auto_approval_disabled(cls, version):
...
         """Check whether the add-on is a promoted addon group that requires
-        pre-review.
-
-        Only applies to listed versions."""
-        if not version.channel == amo.CHANNEL_LISTED:
-            return False
-        promo_group = version.addon.promoted_group(currently_approved=False)
-        return bool(promo_group and promo_group.pre_review)
+        pre-review."""
+        return bool(
```
bobsilverberg, line 604: I'm sorry but I cannot spot the difference in the logic between the old code and the new code. It looks like they will both return `False` if `not version.channel == amo.CHANNEL_LISTED`.

src/olympia/reviewers/utils.py
```diff
@@ -789,7 +789,7 @@ def set_file(self, status, file):
 
     def set_promoted(self):
         group = self.addon.promoted_group(currently_approved=False)
-        if group and group.pre_review:
+        if self.version.channel == amo.CHANNEL_LISTED and group.pre_review:
```
bobsilverberg, line 792: The `if group` is no longer needed?

bobsilverberg (approved):
The code looks fine to me. Not being 100% aware of all the logic intricacies that have been discussed around this feature, perhaps @diox should take a quick peek as well.

I also asked a few questions.

r+wc

src/olympia/reviewers/models.py
eviljeff: Yeah, there wasn't any difference in the end - I left it in because it read better to me to have it all in a single statement

src/olympia/reviewers/utils.py
eviljeff: `Addon.promoted_group()` always returns a `PromotedClass` so it's safe.

(The old code worked because if the add-on isn't promoted `NOT_PROMOTED` is returned, which is falsey.)

PR #20428, opened 2023-03-03
Record a single activity log for reviewer actions on multiple versions

### LINKED ISSUE #20045: Group add-on review log by reviewer "actions"
The add-on review log currently shows one entry per version reviewed, even if multiple versions were reviewed in one review. We should group those entries together so that the log shows one entry per reviewer "action" (i.e. the submission of a review, regardless of how many versions are affected).

The log currently also shows the version affected, we can drop the version info since we are going to aggregate version-actions as per the above.

diox:
Adjust reviewlog page accordingly, using the `to_string()` method with a `type_='reviewlog'` parameter to do any formatting specific to the reviewlog.

Fixes #20045

### DIFF
--- a/src/olympia/activity/models.py
+++ b/src/olympia/activity/models.py
@@ -7,10 +7,11 @@ from copy import copy
 from django.apps import apps
 from django.conf import settings
 from django.db import models
+from django.urls import reverse
 from django.utils import timezone
 from django.utils.functional import cached_property
-from django.utils.html import format_html
-from django.utils.translation import gettext
+from django.utils.html import format_html, mark_safe
+from django.utils.translation import gettext, ngettext
 
 import olympia.core.logger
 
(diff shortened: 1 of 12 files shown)


src/olympia/activity/models.py
```diff
@@ -595,6 +605,32 @@ def get_absolute_url(obj):
...
+            version = format_html(
+                ngettext('Version {0}', 'Versions {0}', len(_versions)),
+                # We're only joining already escaped/safe content.
+                mark_safe(', '.join(_versions)),
+            )
+
+        if channel is None and self.details and 'channel' in self.details:
+            channel = self.details['channel']
+
+        if type_ == 'reviewlog' and addon and addon_pk and addon_name:
```
eviljeff, line 622: it's an edge case, but we can easily get the addon pk if we don't have an `addon` but have at least one version (and the name too if we want)

src/olympia/reviewers/utils.py
```diff
@@ -842,22 +844,27 @@ def log_action(
...
+        elif self.file:
+            details['files'] = [self.file.id]
+
         if version is not None:
             details['version'] = version.version
             args = (self.addon, version)
+        elif versions is not None:
+            details['versions'] = [v.version for v in versions]
+            details['files'] = [v.file.id for v in versions]
+            args = (self.addon,) + tuple(versions)
```
eviljeff, line 861:
```suggestion
            args = (self.addon, *versions)
```

src/olympia/reviewers/utils.py
```diff
@@ -842,22 +844,27 @@ def log_action(
...
-            file = self.file
-        if file is not None:
-            details['files'] = [file.id]
         if version is None and self.version:
             version = self.version
+
+        if file is not None:
+            details['files'] = [file.id]
+        elif self.file:
+            details['files'] = [self.file.id]
```
eviljeff, line 853: nit, and probably out of scope for this patch, but I can't see why we'd need to pass `file` to this function these days - it will always be `version.file`

eviljeff (approved):

src/olympia/activity/models.py
diox: Yeah... but at this point we're fighting the way `ActivityLog` builds that string... It doesn't keep instances around unless explicitly adding code for that, `addon` and `versions` are strings here, not instance(s) (that are built if one of the `arguments` is an `Addon` or `Version`, respectively).

src/olympia/reviewers/utils.py
diox: https://github.com/mozilla/addons-server/issues/20442

PR #20437, opened 2023-03-06
Create a task to extract host permissions

### LINKED ISSUE #20434: Host permissions of the already signed MV3 add-ons should be stored in `webext_permissions`
We started to store the MV3 host permissions in https://github.com/mozilla/addons-server/issues/20417. We should backfill the data for existing add-ons. This shouldn't be too complicated since only MV3 add-ons can define this manifest key.

bobsilverberg:
Fixes #20434

### DIFF
--- a/src/olympia/files/management/commands/extract_host_permissions.py
+++ b/src/olympia/files/management/commands/extract_host_permissions.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+from django.core.management.base import BaseCommand
+
+import olympia.core.logger
+from olympia import amo
+from olympia.amo.celery import create_chunked_tasks_signatures
+from olympia.files.models import File
+from olympia.files.tasks import extract_host_permissions
+
+
+class Command(BaseCommand):
+    help = 'Extract host permissions from manifests in stored xpis.'
+
+    def handle(self, *args, **options):
+        log = olympia.core.logger.getLogger('z.files')
+        files = File.objects.filter(version__addon__type=amo.ADDON_EXTENSION).order_by(
+            'pk'
+        )
+        pks = list(files.values_list('pk', flat=True))
+
+        log.info(
+            'Using %s file pks to extract host permissions, max pk: %s'
+            % (len(pks), pks[-1])
+        )
+        if pks:
+            chunked_tasks = create_chunked_tasks_signatures(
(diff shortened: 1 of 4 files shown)


src/olympia/files/tests/test_commands.py
```diff
@@ -0,0 +1,53 @@
...
+        super(TestExtractHostPermissions, self).setUp()
+        self.addon = Addon.objects.create(
+            guid='mv3@extension', type=amo.ADDON_EXTENSION, name='xxx'
+        )
+        self.version = Version.objects.create(addon=self.addon)
+        UserProfile.objects.create(pk=settings.TASK_USER_ID)
+
+    def test_extract(self):
+        upload = self.get_upload('webextension_mv3.xpi')
+        parsed_data = parse_addon(upload, user=mock.Mock(groups_list=[]))
```
eviljeff, line 40: `parse_addon` needs an addon passing to it if the xpi you're trying to parse has a guid defined in it's manifest.  If `addon` is not passed as an arg then the guid in the manifest is validated against all existing guids in `Addon.unfiltered` to make sure it's unique - it's failing because you're creating `self.addon` with the same guid.

Ideas: Passing `self.addon` to `parse_addon` might work?  Otherwise create `self.addon` after you've called `parse_addon`.

bobsilverberg: Thanks!

src/olympia/files/management/commands/extract_host_permissions.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Command(BaseCommand):
+    help = 'Extract host permissions from manifests in stored xpis.'
+
+    def handle(self, *args, **options):
+        log = olympia.core.logger.getLogger('z.files')
+        files = File.objects.filter(version__addon__type=amo.ADDON_EXTENSION).order_by(
+            'pk'
+        )
```
eviljeff, line 18: Not sure we particularly need to extract permissions from files that are already disabled but I guess it doesn't hurt.

eviljeff (approved):

src/olympia/files/management/commands/extract_host_permissions.py
bobsilverberg: It's basically the same code we ran when populating `optional_permissions`, so I didn't really consider that.

PR #20438, opened 2023-03-06
raise validation error on bad lang request values

eviljeff:
fixes https://github.com/mozilla/addons-server/issues/20452, on the back of https://bugzilla.mozilla.org/show_bug.cgi?id=1820308

### DIFF
--- a/docs/topics/api/overview.rst
+++ b/docs/topics/api/overview.rst
@@ -193,6 +193,11 @@ For example, for a request ``?lang=de``:
         }
     }
 
+
+.. warning::
+    ``lang`` must only contains alphanumeric characters (plus ``-`` and ``_``).
+
+
 For ``POST``, ``PATCH`` and ``PUT`` requests you submit an object containing
 translations for any languages needing to be updated/saved.  Any language not
 in the object is not updated, but is not removed.
@@ -452,7 +457,7 @@ These are `v5` specific changes - `v4` changes apply also.
 * 2023-03-02: added specific HTTP 409 status code for add-on/version submissions that already exist
 * 2023-03-02: added support for calling the version detail endpoint using a version number instead of an ``id``.
 * 2023-03-09: added ``is_disabled`` to version detail and update endpoints, for authenticated developers and revieweers. https://github.com/mozilla/addons-server/issues/20142
-
+* 2023-03-08: restricted ``lang`` parameter to only alphanumeric, ``_``, ``-``. https://bugzilla.mozilla.org/show_bug.cgi?id=1820308
 
 .. _`#11380`: https://github.com/mozilla/addons-server/issues/11380/
 .. _`#11379`: https://github.com/mozilla/addons-server/issues/11379/
--- a/docs/topics/api/overview.rst
+++ b/docs/topics/api/overview.rst
@@ -457,7 +457,7 @@ These are `v5` specific changes - `v4` changes apply also.
 * 2023-03-02: added specific HTTP 409 status code for add-on/version submissions that already exist
 * 2023-03-02: added support for calling the version detail endpoint using a version number instead of an ``id``.
 * 2023-03-09: added ``is_disabled`` to version detail and update endpoints, for authenticated developers and revieweers. https://github.com/mozilla/addons-server/issues/20142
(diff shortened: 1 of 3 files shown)


diox (commented):
Could we validate them and not even include them instead ? We don't accept any random language code for translations, right ?

eviljeff:
> Could we validate them and not even include them instead ? We don't accept any random language code for translations, right ?

we don't... but I'm not sure the frontend list of languages and the server list are always in-sync, so we don't want to raise if the code isn't on the server list. We could return a generic "unsupported" language code?  (I'm assuming frontend won't baulk at a different code in the response than the request, because it's going to fallback to the default anyway)

src/olympia/api/fields.py
```diff
@@ -108,8 +112,10 @@ def flat(self):
 
     def get_requested_language(self):
         request = self.context.get('request', None)
-        if request and request.method == 'GET' and 'lang' in request.GET:
-            return request.GET['lang']
+        if request and request.method == 'GET' and (lang := request.GET.get('lang')):
+            if not self.flat and not re.fullmatch(LANGUAGE_CODE_REGEX, lang):
+                self.fail('unknown_locale', lang_code=escape(lang))
```
diox, line 117: Still not sure there is any value in escaping that for HTML. @bobsilverberg what does the frontend typically do with error messages coming from the API ? Are they run through `DOMPurify` ? I see we're returning the same kind of error, with the `lang_code` not escaped at all, when writing translations.

eviljeff: 🤷‍♂️ given the entire point of the bug was us returning the value as-is in the response I figured it we should at least escape it in the error message.

diox: I think we should be consistent and stick to one way of handling parameters in error messages across the API - As I said in the bug, I'm not convinced we care about the escaping at all tbh.

eviljeff: Agree on consistency, but the escaping was the only thing that could be seen as "fixing" the bug.  If we're no longer doing that, then this pr is fixing an adjacent issue - validating the locale code (I can file an issue).  I suppose the bug is a wontfix then.

(imo a client is as likely to include a json value in html without sanitizing as a json key)

diox: I suppose it doesn't hurt to escape it, but then we should also escape other error messages and not just this one, and we're back to being blocked on what the frontend does to avoid accidentally double-escaping

bobsilverberg: In at least the standard case, where we are using `errorHandler.renderError()` we do not appear to be doing anything at all, other than outputting the message from the error. I can check for other instances where we might be explicitly outputting errors.

bobsilverberg: Looking through the codebase, I am confident that we always render error messages via `renderError()`, which means we are not doing any processing of the message at all.

diox: With https://github.com/mozilla/addons-frontend/issues/12040 handling that on our side, I believe we can wontfix the escaping but still do the validation. I personally don't think it's very useful to include the broken locale in the error message, but it shouldn't be escaped IMHO - it's the client's responsibility to deal with content as they see fit.

eviljeff: https://github.com/mozilla/addons-server/pull/20438/commits/9173f9f195a01f92c5a28ce08cf92b0184d366b9 drops the sanitizing entirely

docs/topics/api/overview.rst
```diff
@@ -452,7 +457,7 @@ These are `v5` specific changes - `v4` changes apply also.
 * 2023-03-02: added specific HTTP 409 status code for add-on/version submissions that already exist
 * 2023-03-02: added support for calling the version detail endpoint using a version number instead of an ``id``.
 * 2023-03-09: added ``is_disabled`` to version detail and update endpoints, for authenticated developers and revieweers. https://github.com/mozilla/addons-server/issues/20142
-
+* 2023-03-08: restricted ``lang`` parameter to only alphanumeric, ``_``, ``-``. https://bugzilla.mozilla.org/show_bug.cgi?id=1820308
```
diox, line 460: You can just link https://github.com/mozilla/addons-server/issues/20452 instead

diox (approved):

PR #20451, opened 2023-03-08
set needs_human_review if auto approved after auto_approval_delayed_until_unlisted

eviljeff:
fixes #20421 ?  

I think this matches what the issue specified, but I'm still a little unclear where (and/or in what circumstances) `auto_approval_delayed_until_unlisted` is cleared.  If it's _not_ cleared at some point then _every_ unlisted version after that datetime will be flagged `needs_human_review` which I doubt is desirable.

### DIFF
--- a/src/olympia/reviewers/tests/test_utils.py
+++ b/src/olympia/reviewers/tests/test_utils.py
@@ -1098,6 +1098,32 @@ class TestReviewHelper(TestReviewHelperBase):
         # Not changed this this is not a human approval.
         assert addon_flags.auto_approval_disabled_until_next_approval_unlisted
 
+    def _unlisted_approve_flag_if_passed_auto_approval_delayed_setup(self, delay):
+        self.setup_data(
+            amo.STATUS_NULL, channel=amo.CHANNEL_UNLISTED, human_review=False
+        )
+        AddonReviewerFlags.objects.create(
+            addon=self.addon, auto_approval_delayed_until_unlisted=delay
+        )
+        assert not self.review_version.needs_human_review
+
+        self.helper.handler.approve_latest_version()
+        self.addon.reload()
+        self.review_version.reload()
+        self.file.reload()
+        assert self.addon.status == amo.STATUS_NULL
+        assert self.file.status == amo.STATUS_APPROVED
+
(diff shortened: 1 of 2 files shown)


diox:
> I think this matches what the issue specified, but I'm still a little unclear where (and/or in what circumstances) `auto_approval_delayed_until_unlisted` is cleared. If it's _not_ cleared at some point then _every_ unlisted version after that datetime will be flagged `needs_human_review` which I doubt is desirable.

Reviewers - notably those with the `Addons:TriageDelayed` permission, who see add-ons with a temporary delay in the queue - are supposed to drop the delay.

I think to go with that patch, we need to show the button to clear the delay even if that delay is in the past - so that reviewers can always clear it.

eviljeff:
> > I think this matches what the issue specified, but I'm still a little unclear where (and/or in what circumstances) `auto_approval_delayed_until_unlisted` is cleared. If it's _not_ cleared at some point then _every_ unlisted version after that datetime will be flagged `needs_human_review` which I doubt is desirable.
> 
> Reviewers - notably those with the `Addons:TriageDelayed` permission, who see add-ons with a temporary delay in the queue - are supposed to drop the delay.
> 
> I think to go with that patch, we need to show the button to clear the delay even if that delay is in the past - so that reviewers can always clear it.

Even with the button, relying on a manual action from a certain subset of reviewers seems to me unworkable - if we land this patch I could see it resulting in a huge number of extra versions flagged needs_human_review, filling up the queue with noise. Can we clear `auto_approval_delayed_until_unlisted` after the first human review?  @wagnerand also, is there some extra criteria to distinguish which versions are flagged?

bobsilverberg (approved):
The code looks good. I saw some discussions around this on Slack. Not sure if any logic changes are needed,  but for what it does this patch looks good to me.

diox:
@wagnerand can you confirm here and in the slack thread that this is what you want?

eviljeff:
> we need to show the button to clear the delay even if that delay is in the past - so that reviewers can always clear it.

https://github.com/mozilla/addons-server/pull/20451/commits/692a5159bfa4ef0522ba07f95960dd5961fa4011

bobsilverberg (approved):
I still approve. 😄

PR #20454, opened 2023-03-09
pass --enable-background-service-worker to linter on dev

### LINKED ISSUE #20001: Add a setting to enable service worker support in the linter
In https://github.com/mozilla/addons-linter/releases/tag/5.23.0, we added a new CLI flag to enable the background service worker support (MV3+). We should have a waffle switch on AMO to toggle this flag, which will be useful eventually.

The CLI flag to pass to the linter is: `--enable-background-service-worker` and its default value is `false`.

eviljeff:
fixes #20001

### DIFF
--- a/src/olympia/conf/dev/settings.py
+++ b/src/olympia/conf/dev/settings.py
@@ -64,6 +64,8 @@ CEF_PRODUCT = STATSD_PREFIX
 NEW_FEATURES = True
 
 ADDONS_LINTER_BIN = 'node_modules/.bin/addons-linter'
+# --enable-background-service-worker linter flag value
+ADDONS_LINTER_ENABLE_SERVICE_WORKER = True
 
 ALLOW_SELF_REVIEWS = True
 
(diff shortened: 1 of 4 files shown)


willdurand:
I know _I_ requested this change but that might not be something we need at the moment. @rpl WDYT?

eviljeff:
> I know _I_ requested this change but that might not be something we need at the moment. @rpl WDYT?

🤷 patch is written and issue is open.  If it's no longer required, or the requirements have changed can someone update/close the issue.

rpl:
> I know _I_ requested this change but that might not be something we need at the moment. @rpl WDYT?

:+1: green light on proceeding with adding the setting and enabling it only on dev for now.

To be fair it may change slightly the validation on dev but not a big deal in the end, given that in stage it is going to stay disabled (and we can decide to disable it on -dev in a followup if it turns out we'd want to keep it disabled even on -dev while we followup on some corner case that we may notice by having enabled on -dev).

diox (approved):

PR #20457, opened 2023-03-10
Don't create DeniedGuid for `gc` addon deletes

### LINKED ISSUE #20447: Don't create DeniedGuid when gc (garbage collection) is hard-deleting add-ons without versions
[Langpacks are submitted automatically](https://moz-releng-docs.readthedocs.io/en/latest/addons/langpacks.html) when Firefox (or Nightly, or Dev Edition...) is released.

Only users with the specific permission can submit them:
https://github.com/mozilla/addons-server/blob/9b453173c2bd5c6495eca8800dabeec835250bff/src/olympia/access/acl.py#L46-L56

In addition, they all have a special `guid` ending in `@firefox.mozilla.org` or `@devedition.mozilla.org`. Interestingly, these guids are not currently in the `RESERVED_ADDON_GUIDS` as this was meant for the privileged add-ons that are signed by https://github.com/mozilla-extensions/xpi-manifest.

On stage, it appears that some langpack add-ons were created without versions - probably using the old signing API - were then disabled, and then automatically deleted. When they got deleted, their guids entered the `DeniedGuid` table, and their `guid` became impossible to re-use.

This is a problem for langpacks submitted automatically, as it's impossible for releng script to deal with that error.

Specifically for langpacks, for users with the permission, we should bypass `DeniedGuid` entirely. This mechanism was meant to avoid the possibility of a disabled-but-deleted malicious add-on re-using a guid, becoming able to ship updates to existing users, but that shouldn't be a concern here, it does us more harm than good for langpacks.

eviljeff:
fixes #20447

### DIFF
--- a/src/olympia/addons/tests/test_commands.py
+++ b/src/olympia/addons/tests/test_commands.py
@@ -442,13 +442,13 @@ class TestDeleteObsoleteAddons(TestCase):
         assert Addon.unfiltered.get(id=self.dictionary.id)
 
         # check the guids have been added to DeniedGuid
-        assert DeniedGuid.objects.all().count() == 4
+        assert DeniedGuid.objects.all().count() == 5
         assert DeniedGuid.objects.get(guid=self.xul_theme.guid)
         assert DeniedGuid.objects.get(guid=self.lpaddon.guid)
         assert DeniedGuid.objects.get(guid=self.plugin.guid)
         # no self.lwt as it didn't have a guid
         assert DeniedGuid.objects.get(guid=self.webapp.guid)
-        assert not DeniedGuid.objects.filter(guid=self.no_versions.guid).exists()
+        assert DeniedGuid.objects.filter(guid=self.no_versions.guid)
 
     def test_hard_with_already_deleted(self):
         Addon.unfiltered.update(status=amo.STATUS_DELETED)
@@ -462,6 +462,7 @@ class TestDeleteObsoleteAddons(TestCase):
         assert Addon.objects.get(id=self.extension.id)
         assert Addon.objects.get(id=self.static_theme.id)
         assert Addon.objects.get(id=self.dictionary.id)
+        # Addon.delete will hard delete this add-on because it's got no versions.
         assert not Addon.unfiltered.filter(guid=self.no_versions.guid).exists()
 
 
--- a/src/olympia/addons/tests/test_commands.py
+++ b/src/olympia/addons/tests/test_commands.py
@@ -424,8 +424,10 @@ class TestDeleteObsoleteAddons(TestCase):
         self.lwt = addon_factory(type=9, guid=None)  # _ADDON_PERSONA
         self.webapp = addon_factory()
         self.webapp.update(type=11)  # webapp
+        self.no_versions = addon_factory(type=6)  # _ADDON_PLUGIN
+        self.no_versions.current_version.delete(hard=True)
 
-        assert Addon.unfiltered.count() == 8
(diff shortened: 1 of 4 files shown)


eviljeff:
@diox I think this works as a solution without an extra kwarg... I can't think of any reason generally why we'd want to block a guid that hadn't been "used"

diox:
I can think of one: when we deleted legacy xpis, we hard-deleted versions, leaving some add-ons versionless. But we did want to block their guids when deleting the add-ons themselves later, because they had been used.

eviljeff:
> I can think of one: when we deleted legacy xpis, we hard-deleted versions, leaving some add-ons versionless. But we did want to block their guids when deleting the add-ons themselves later, because they had been used.

Does that still apply? Those add-ons would have been gc'd long ago if they had no versions.

diox:
Those add-ons are long gone, but the general idea still applies: we could have add-ons that are versionless because they used to have versions that have since been hard-deleted.

I think for that reason, we shouldn't try to guess whether or not to add something to `DeniedGuid` in `delete_addons` task. Make it a keyword argument that defaults to `True`, and set it to `False` when calling from `gc()` for versionless add-ons. That way, commands and tasks we write in the future will add to `DeniedGuid` by default, and we only have an exception for the specific use-case we know is ok.

eviljeff:
> Those add-ons are long gone, but the general idea still applies: we could have add-ons that are versionless because they used to have versions that have since been hard-deleted.
> 
> I think for that reason, we shouldn't try to guess whether or not to add something to `DeniedGuid` in `delete_addons` task. Make it a keyword argument that defaults to `True`, and set it to `False` when calling from `gc()` for versionless add-ons. That way, commands and tasks we write in the future will add to `DeniedGuid` by default, and we only have an exception for the specific use-case we know is ok.

https://github.com/mozilla/addons-server/pull/20457/commits/94a7b5398d3e41ca89ba10ad8c7f0791f5b65115 

But I note if we _did_ add a future command/task that hard-deleted all the versions from add-ons, then the `gc` cron would capture them anyway and delete without creating a `DeniedGuid`, unless we also deleted the add-on straight afterwards - and we'd likely end up creating a specialized task to do it all in one go, rather then inefficiently chaining `delete_versions`(*) and `delete_addons`.

(*) `delete_versions` is the only place (outside of tests) that calls `Version.delete` with `hard=True`, and we don't call/use `delete_versions` anywhere.

diox:
> But I note if we _did_ add a future command/task that hard-deleted all the versions from add-ons, then the `gc` cron would capture them anyway and delete without creating a `DeniedGuid`, unless we also deleted the add-on straight afterwards - and we'd likely end up creating a specialized task to do it all in one go, rather then inefficiently chaining `delete_versions`(*) and `delete_addons`.

Yeah. At least seeing that parameter in `delete_addons` would force us to think about it...

src/olympia/addons/tests/test_commands.py
```diff
@@ -424,8 +424,10 @@ def setUp(self):
         self.lwt = addon_factory(type=9, guid=None)  # _ADDON_PERSONA
         self.webapp = addon_factory()
         self.webapp.update(type=11)  # webapp
+        self.no_versions = addon_factory(type=6)  # _ADDON_PLUGIN
```
bobsilverberg, line 427: Should this comment read `# _ADDON_LPADDON`?

bobsilverberg (approved):
Looks good, just a comment about a comment.

r+wc

PR #20480, opened 2023-03-16
Merge deployment and development Dockerfiles

### LINKED ISSUE #20472: Merge `Dockerfile.deploy` and `Dockerfile`, using the same image for development and production
We're currently building and using 2 different docker images:
- One built by CircleCI from `Dockerfile.deploy`, available over mozilla/addons-server on Docker Hub
- One built manually from `Dockerfile`, available over addons/addons-server on Docker Hub

This is annoying, and especially so as Docker has notified us that the `addons` org was using a legacy free plan that they have deprecated.

We should find a way to merge the 2, possibly fixing the long-standing issues around permissions (like https://github.com/mozilla/addons-server/issues/20263) caused by users not using the same `uid` as the development image expects.

Then we'll be able to remove the development image completely, as well as the entire `addons` org on Docker Hub.

diox:
Fixes #20472


diox:
To try this locally:
- Stop the `docker-compose` services
- Build the image with `docker build . -t mozilla/addons-server`
- Either start from scratch with `make initialize` (or equivalent noted in the docs for Windows) or emulate on an existing install with:
  ```
    touch deps/package.json
    touch deps/package.json
    make create_env_file
    docker-compose run --rm --user ${UID}:${GID} web make update_deps```
- Start the `docker-compose` services as normal.

diox:
@bobsilverberg if you have time to try this on Mac... Either updating or reinstalling from scratch following docs.

bobsilverberg:
I tested both an update and a fresh install, on OSX, using the instructions above, and it all seems to work fine.

Makefile-os
```diff
@@ -11,21 +16,44 @@ help_submake:
...
+.PHONY: rootshell
 rootshell: ## connect to a running addons-server docker shell with root user
 	docker-compose exec --user root web bash
 
+.PHONY: create_env_file
+create_env_file:
+	echo "UID=${UID}\nGID=${GID}" > .env
+
+.PHONY: initialize
+initialize: create_env_file
```
eviljeff, line 37: I'm not super happy about us overriding the name of `initialize` recipe, as it'll mean it does different things depending on whether you're inside or outside of docker.  For updates we have `update_docker` (in Makefile-os) and `update` (in Makefile-docker) - maybe this should be called `initialize_docker`?  (alternatively, have `update` and `initialize` here, and the container specific ones be called `update_container` and `initializer_container` or similar?)

Makefile-os
```diff
@@ -11,21 +16,44 @@ help_submake:
...
+# instead.
+	touch deps/package.json
+	touch deps/package.json
+# Note that this is running with --user ${UID}:${GID} because the user olympia
+# would be uid 9500 regardless of host at this point (this is only fixed when
+# the container is up, through the command defined in docker-compose.yml),
+# which is wrong for local development.
+	docker-compose run --rm --user ${UID}:${GID} web make update_deps
+	docker-compose up -d
+	docker-compose exec --user olympia web make $(MAKECMDGOALS) ARGS=$(ARGS)
```
eviljeff, line 53:
```suggestion
	docker-compose exec --user olympia web make initialize
```
? (we don't pass on the full args in the other recipes, apart from the fallback)

docs/topics/install/docker.rst
```diff
@@ -68,17 +68,24 @@ on your host machine::
...
     # Download the containers
     docker-compose pull  # Can take a while depending on your internet bandwidth.
-    # Start up the containers
-    docker-compose up -d
     make initialize  # Answer yes, and create your superuser when asked.
-    # On Windows you can substitute `make initialize` for the command:
-    docker-compose exec web make initialize
+    # On Windows you can substitute `make initialize` by the following commands:
+    docker-compose run --rm --user olympia web make initialize
+    docker-compose up -d
```
eviljeff, line 74: needs updating.
I did
```
docker-compose run --rm --user 9500:9500 web make update_deps
docker-compose up -d
docker-compose exec --user olympia web make initialize
```
but maybe `--user 9500:9500` can be substituted for `--user olympia` too?

docs/topics/install/docker.rst
```diff
@@ -68,17 +68,24 @@ on your host machine::
...
 
 .. note::
 
    Docker requires the code checkout to exist within your home directory so
    that Docker can mount the source-code into the container.
 
+   Because the containers need to match the user/group permissions from your
+   host machine, on non-windows machines make sure to run ``make initialize``
+   once before running ``docker-compose up -d`` for the first time. That will
```
eviljeff, line 83:
```suggestion
   On Mac and Linux, because the containers need to match the user/group permissions from your
   host machine, make sure to run ``make initialize``
   once before running ``docker-compose up -d`` for the first time. That will
```

eviljeff (approved):
One annoyance (for me at least) - previously, `docker-compose exec` ran as `olympia` user by default (i.e. without specifying `--user`, now it defaults to `root` 😞)

docs/topics/install/docker.rst
diox: Yeah in your case `olympia` should be `9500` (it's the default value, and on Windows we are not overriding that with the `uid` from the host)

docs/topics/install/docker.rst
eviljeff, line 1: we suggest using `docker-compose` in a few other places in this file - they probably need updating with `--user` too

PR #20489, opened 2023-03-21
Allow confirmation of auto-approval of a disabled/deleted add-on

### LINKED ISSUE #20481: "Confirm auto-approval" should be available for auto-approved versions of invisible or deleted add-ons

"Confirm auto-approval" should be available for auto-approved versions of invisible or deleted add-ons.

### What happened?

Such add-ons might get added to the queue when they get added to the Notable group, but we don't have a way to get them out of the queue.

With confirmation being available, the add-on can be removed from the queue. The alternative to that would be blocking which is already available. Rejecting is not an option and can't be one.

diox:
Fixes #20481

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -412,6 +412,7 @@ class ReviewHelper:
         version_is_unlisted = (
             self.version and self.version.channel == amo.CHANNEL_UNLISTED
         )
+        version_is_listed = self.version and self.version.channel == amo.CHANNEL_LISTED
         promoted_group = self.addon.promoted_group(currently_approved=False)
         is_static_theme = self.addon.type == amo.ADDON_STATICTHEME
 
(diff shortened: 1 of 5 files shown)


src/olympia/reviewers/utils.py
```diff
@@ -486,16 +487,13 @@ def get_actions(self):
...
-            and self.version.channel == amo.CHANNEL_LISTED
-            and self.addon.current_version
-            and self.addon.current_version.was_auto_approved
+        addon_is_valid_and_version_is_listed = addon_is_valid and version_is_listed
+        current_or_latest_listed_version_was_auto_approved = version_is_listed and (
+            (
+                self.addon.current_version
+                and self.addon.current_version.was_auto_approved
+            )
+            or self.version.was_auto_approved
```
eviljeff, line 496: This sets up an inconsistency in an add-on that has a current_version that wasn't auto_approved but `self.version` _was_ auto approved, because later in `confirm_auto_approved` we only drop back to `self.version` if `current_version` is falsey.  I'm not sure if this case can happen, but perhaps it should be changed to something similar to:
```suggestion
        current_or_latest_listed_version_was_auto_approved = version_is_listed and (
            (
                self.addon.current_version
                and self.addon.current_version.was_auto_approved
            )
            or (not self.addon.current_version and self.version.was_auto_approved)
```

eviljeff (approved):

src/olympia/reviewers/utils.py
diox: Yeah. But if `current_version` is not falsey, then that is the version we should care about. If looking at the listed channel, it's only really absent when the add-on is deleted.

diox: Ah yeah I see. I'll tweak the condition.

PR #20490, opened 2023-03-21
Make BlockSubmission delayable

### LINKED ISSUE #20478: Allow a BlockSubmission to be created with a delay
part of https://mozilla-hub.atlassian.net/browse/AMO-43

We want to be able to create a block submission that won't turn into a Block on on the blocklist (or be sent for approval either) until x days have passed.  x should be definable.

eviljeff:
fixes #20478 - this adds the core functionality in BlocklistSubmission, and the admin.

Possible follow-ups would be to expose the delay status/date in the changelist view, and to add integration in the reviewer tools (if we wanted to have a widget there to set the days instead), plus the rest of the jira story obviously.


scripts/crontab/crontab.tpl
```diff
@@ -15,6 +15,7 @@ HOME=/tmp
 10 * * * * %(z_cron)s update_blog_posts
 15 * * * * %(django)s send_pending_rejection_last_warning_notifications
 20 * * * * %(z_cron)s addon_last_updated
+30 * * * * %(z_cron)s process_blocklistsubmissions
```
eviljeff, line 18: Could be more frequent but I don't know how often they're be created and how timely the block will need to be 🤷

diox (approved):

scripts/crontab/crontab.tpl
diox: Once per hour is probably fine IMHO.

PR #20495, opened 2023-03-23
Compute hotness for non-public add-ons as well

### LINKED ISSUE #20494: Include non-public add-ons when (re)setting hotness
The `hotness` value in AMO is a denormalized field representing user growth of an add-on. It's exposed on the frontend as the "Trending" sort. It's reset each day but only with public add-ons, so for everything else the data isn't necessarily accurate in the database, so let's change that.

diox:
Fixes #20494

### DIFF
--- a/src/olympia/addons/tasks.py
+++ b/src/olympia/addons/tasks.py
@@ -290,11 +290,7 @@ def update_addon_hotness(averages):
     log.info('[%s] Updating add-ons hotness scores.', (len(averages)))
 
     averages = dict(averages)
-    addons = (
-        Addon.objects.filter(guid__in=averages.keys())
-        .filter(status__in=amo.APPROVED_STATUSES)
-        .no_transforms()
-    )
+    addons = Addon.unfiltered.filter(guid__in=averages.keys()).no_transforms()
 
     for addon in addons:
         average = averages.get(addon.guid)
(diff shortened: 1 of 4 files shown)


src/olympia/addons/tasks.py
```diff
@@ -290,11 +290,7 @@ def update_addon_hotness(averages):
     log.info('[%s] Updating add-ons hotness scores.', (len(averages)))
 
     averages = dict(averages)
-    addons = (
-        Addon.objects.filter(guid__in=averages.keys())
-        .filter(status__in=amo.APPROVED_STATUSES)
-        .no_transforms()
-    )
+    addons = Addon.unfiltered.filter(guid__in=averages.keys()).no_transforms()
```
bobsilverberg, line 293: This also removed the filter for `status__in=amo.APPROVED_STATUSES`. Is that intentional?

src/olympia/addons/tests/test_tasks.py
```diff
@@ -208,13 +208,15 @@ def test_update_addon_hotness():
     addon1 = addon_factory(hotness=0, status=amo.STATUS_APPROVED)
     addon2 = addon_factory(hotness=123, status=amo.STATUS_APPROVED)
     addon3 = addon_factory(hotness=123, status=amo.STATUS_AWAITING_REVIEW)
+    addon4 = addon_factory(hotness=123, status=amo.STATUS_AWAITING_REVIEW)
```
bobsilverberg, line 211: Why are we adding a 4th add-on that is the same as the 3rd add-on?

bobsilverberg (approved):

r+wc

src/olympia/addons/tasks.py
diox: Yes. We want to set `hotness` for all kinds of add-ons, regardless of status

src/olympia/addons/tests/test_tasks.py
diox: I meant to make it deleted but forgot... I'll fix that.

PR #20543, opened 2023-03-30
Centralize code adding flags when source code is provided

### LINKED ISSUE #20508: Set `needs_human_review` on relevant version for some developer actions
https://mozilla-hub.atlassian.net/browse/AMO-47

- When a developer leaves a developer reply, the version the reply was for should receive the `needs_human_review` flag and a due date of `<reply-time>` if that version does not have a due date already.

- When a developer submits source code for a version pending rejection, that version should receive the `needs_human_review` flag and a due date of `<review-time>` if it does not have one already.

diox:
Fixes #20508

### DIFF
--- a/src/olympia/versions/models.py
+++ b/src/olympia/versions/models.py
@@ -27,6 +27,7 @@ from publicsuffix2 import get_sld
 import olympia.core.logger
 
 from olympia import activity, amo, core
+from olympia.activity.utils import log_and_notify
 from olympia.amo.decorators import use_primary_db
 from olympia.amo.fields import PositiveAutoField
 from olympia.amo.models import (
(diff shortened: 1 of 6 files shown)


src/olympia/versions/models.py
```diff
@@ -734,6 +734,22 @@ def is_unreviewed(self):
     def sources_provided(self):
         return bool(self.source)
 
+    def flag_if_sources_were_provided(self, user):
```
eviljeff, line 737: Do we need a test in test_models.py for this?  ("no we don't" is an acceptable answer 😄)

eviljeff (approved):

src/olympia/versions/models.py
diox: https://github.com/mozilla/addons-server/pull/20543/commits/e2d9d44b9376cf04116115a79eacb0992ec9f6b2

PR #20574, opened 2023-04-08
replace git protocol with https

ueokande:
Cloning via git:// is no longer supported on GitHub.com. 
Use https:// protocol on `git clone` procedure instead.


wagnerand:
Why not SSH?

diox:
Doesn't really matter to suggest using ssh, since it doesn't mention to use your own fork... It could be improved, but that's already better than mentioning the no longer supported `git://` protocol.

PR #20581, opened 2023-04-12
Add UserRestrictions for Rating submission and auto moderation

### LINKED ISSUE #20569: Add rating user restriction
part of https://mozilla-hub.atlassian.net/browse/AMO-49

> For every pattern, I want to decide whether ratings for matching users should be denied or flagged for moderation

eviljeff:
fixes #20569 (threw in support for bulk email restrictions too because it was easier than making an exception for that type)

I had to reword the restriction error messages to make them more generic... I think they still read okay.  If not we need to complicate how `error_message` and `get_error_message` works to pass the action as an arg each time.]

### DIFF
--- a/src/olympia/ratings/models.py
+++ b/src/olympia/ratings/models.py
@@ -319,7 +319,7 @@ class RatingFlag(ModelBase):
     )
     FLAGS = (
         *USER_FLAGS,
-        (AUTO, _('Auto-flagged due to word match')),
+        (AUTO, _('Auto-flagged due to word match or user restriction')),
     )
 
     rating = models.ForeignKey(Rating, db_column='review_id', on_delete=models.CASCADE)
--- a/src/olympia/users/models.py
+++ b/src/olympia/users/models.py
@@ -642,6 +642,8 @@ class DeniedName(ModelBase):
 RESTRICTION_TYPES = Choices(
     ('SUBMISSION', 1, 'Submission'),
     ('APPROVAL', 2, 'Approval'),
+    ('RATING', 3, 'Rating'),
+    ('RATING_MODERATE', 4, 'Rating Flag for Moderation'),
 )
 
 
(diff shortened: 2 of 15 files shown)


src/olympia/ratings/models.py
```diff
@@ -319,7 +319,7 @@ class RatingFlag(ModelBase):
     )
     FLAGS = (
         *USER_FLAGS,
-        (AUTO, _('Auto-flagged due to word match')),
+        (AUTO, _('Auto-flagged due to word match or user restriction')),
```
diox, line 322: Could/should we use a specific flag instead ?

src/olympia/users/models.py
```diff
@@ -642,6 +642,8 @@ def fetch_names():
 RESTRICTION_TYPES = Choices(
     ('SUBMISSION', 1, 'Submission'),
     ('APPROVAL', 2, 'Approval'),
```
diox, line 644: Maybe we should rename these to include `addon` in there for clarity.

src/olympia/users/models.py
```diff
@@ -642,6 +642,8 @@ def fetch_names():
...
+    ('RATING', 3, 'Rating'),
+    ('RATING_MODERATE', 4, 'Rating Flag for Moderation'),
 )
```
diox, line 647: Should we add another type that would disallow both submissions and ratings ?

diox (commented):
`RestrictionChecker` should probably move out of `addons` as well. In `amo` maybe ? It used to be in `devhub` IIRC, but that wasn't a great fit...

src/olympia/users/models.py
eviljeff: Mmm, well, it'd complicate the model somewhat because there would no longer be a one-to-one between the restriction type and the `allow_xxx` action.  As it wasn't specified in the ticket as a requirement I'd lean towards adding this in a follow-up if needed.

diox:
> > `RestrictionChecker` should probably move out of `addons` as well. In `amo` maybe ? It used to be in `devhub` IIRC, but that wasn't a great fit...
> 
> They're user restrictions so `users`?

Yeah, probably the best option.

src/olympia/users/models.py
eviljeff: https://github.com/mozilla/addons-server/pull/20581/commits/ac9b4f704e9c1a88e7b14dc2815325833a99ffab

src/olympia/ratings/models.py
eviljeff: https://github.com/mozilla/addons-server/pull/20581/commits/431a7c1f1f95b3fcb10c0959db706b964ee88d02

diox (approved):
:shipit:

PR #20588, opened 2023-04-17
Compute hotness over a shorter period and ignore week-ends

### LINKED ISSUE #20564: Make hotness value smoother and computed from the last 2 weeks instead of 3
Hotness is our denormalized field capturing add-on user growth.

- Avoid weekends to make data smoother
- Adjust duration from 3 weeks to 2

diox:
Fixes #20564

### DIFF
--- a/src/olympia/stats/utils.py
+++ b/src/olympia/stats/utils.py
@@ -200,10 +200,10 @@ WITH
     EXTRACT(DAYOFWEEK FROM submission_date) <> 7
   GROUP BY
     addon_id),
-  last_week AS (
+  previous_week AS (
   SELECT
     addon_id,
-    AVG(dau) AS avg_last_week
+    AVG(dau) AS avg_previous_week
   FROM
     `{get_amo_stats_dau_view_name()}`
   WHERE
@@ -219,7 +219,7 @@ SELECT
 FROM
   this_week
 JOIN
-  last_week
+  previous_week
 USING
   (addon_id)
 """
@@ -242,7 +242,7 @@ USING
     return {
         row['addon_id']: {
             'avg_this_week': row['avg_this_week'],
-            'avg_last_week': row['avg_last_week'],
+            'avg_previous_week': row['avg_previous_week'],
         }
         for row in rows
         if row['addon_id']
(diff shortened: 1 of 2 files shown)


src/olympia/stats/utils.py
```diff
@@ -192,30 +194,38 @@ def get_averages_by_addon_from_bigquery(today, exclude=None):
...
   WHERE
     submission_date >= @one_week_date
+  AND
+    EXTRACT(DAYOFWEEK FROM submission_date) <> 1
+  AND
+    EXTRACT(DAYOFWEEK FROM submission_date) <> 7
   GROUP BY
     addon_id),
-  three_weeks_before_this_week AS (
+  last_week AS (
```
eviljeff, line 203: I understand what you mean in context, but `this_week` _is_ the last week, so having a `last_week` also is ambiguous..

src/olympia/stats/utils.py
```diff
@@ -192,30 +194,38 @@ def get_averages_by_addon_from_bigquery(today, exclude=None):
     `{get_amo_stats_dau_view_name()}`
   WHERE
     submission_date >= @one_week_date
+  AND
+    EXTRACT(DAYOFWEEK FROM submission_date) <> 1
+  AND
+    EXTRACT(DAYOFWEEK FROM submission_date) <> 7
```
eviljeff, line 200: Lets hope there aren't any weekend focused extensions! 😬

eviljeff (approved):

src/olympia/stats/utils.py
diox: Since this is just data about which users have the add-on activated and using Firefox, I guess the real problem is week-end focused _users_. If some users are only using Firefox during the week-end, and never pinging during the rest of the week at all, they would be ignored. We're trying to avoid the opposite problem (low week-ends activity skewing the data) so hopefully that's ok.

src/olympia/stats/utils.py
diox: Yeah I don't love it. Any suggestions ? Maybe `this_week` and `previous_week` ?

eviljeff: > Yeah I don't love it. Any suggestions ? Maybe `this_week` and `previous_week` ?

wfm

PR #20593, opened 2023-04-17
chore(remote-settings): update dev & stage urls after gcp migration

ahoneiser:
This is updating remote-settings dev & stage URLs after GCP migration, writer & reader endpoints have changed.


diox:
Thanks!

Issue #20493, opened 2023-03-22
Upgrade to django4.2 LTS


eviljeff:
The RC1 for the next LTS release of django is [available](https://www.djangoproject.com/weblog/2023/mar/20/django-42-rc1/) [now](https://pypi.org/project/Django/4.2rc1/) so it's time to get addons-server up to spec.  No rush though - 3.2 LTS is supported for another year.

eviljeff:
Most of the heavy work should have been in https://github.com/mozilla/addons-server/issues/19595 already

eviljeff:
This is the only thing that jumps out at me from the 4.2 release notes:

https://docs.djangoproject.com/en/dev/releases/4.2/#setting-update-fields-in-model-save-may-now-be-required
> In order to avoid updating unnecessary columns, [QuerySet.update_or_create()](https://docs.djangoproject.com/en/dev/ref/models/querysets/#django.db.models.query.QuerySet.update_or_create) now passes update_fields to the [Model.save()](https://docs.djangoproject.com/en/dev/ref/models/instances/#django.db.models.Model.save) calls. As a consequence, any fields modified in the custom save() methods should be added to the update_fields keyword argument before calling super(). See [Overriding predefined model methods](https://docs.djangoproject.com/en/dev/topics/db/models/#overriding-model-methods) for more details.

eviljeff:
re: https://docs.djangoproject.com/en/dev/releases/4.2/#setting-update-fields-in-model-save-may-now-be-required

None of our tests fail currently so this is likely not going to impact us immediately... but it's a glaring bug that may bite us at some later point when we don't understand why a `update_or_create` call isn't causing a particular automatically-set field to be saved to the database while it _is_ set in the in-memory instance.

Using [IPLog.save](https://github.com/mozilla/addons-server/blob/master/src/olympia/activity/models.py#L231:L235) as an example, this is the existing code:
```python
class IPLog(ModelBase):
    # ....
    def save(self, *args, **kwargs):
        self._ip_address = str(self.ip_address_binary)
        return super().save(*args, **kwargs)
```

At it's most simple we just always set the field and add the field to `update_fields`:
```python
class IPLog(ModelBase):
    # ....
    def save_(self, *, update_fields=None, **kwargs):
        self._ip_address = str(self.ip_address_binary)
        if update_fields is not None and '_ip_address' is not in update_fields:
            update_fields = ('_ip_address', *update_fields)
[...]

PR #20612, opened 2023-04-24
Enable django4.2 deployment

### LINKED ISSUE #19595: support django 4.0+
We're currently running django 3.2, and as a policy stick to LTS releases of django, so the next release we'd upgrade to is django 4.2.  But it doesn't hurt to get ahead of the 4.2 alpha release and where possible make our codebase compatible with newer django.  First step will be any deprecation warnings raised in django 3.2.

eviljeff:
fixes #19595 and part of #20493

There were two fixes since the last round of compatibility work - the `timesince` tests needed a tweak (we already changed the `timeuntil` tests previously); and I noticed the reviewer tools css/js was broken because the form is rendering with `<div><div>...` in 4.2 rather than `<ul><li>...` (I took a look at devhub and reviewer tools to see if there were any other changes but couldn't see anything obvious).

The pr will allow django 4.2 to be deployed wherever we want - SRE will need to add `DJANGO_VERSION` = `django42` to the dev scripts, and it'll continue to default to 3.2 on stage and prod until we've had more chance to pick up any problems. I didn't default the local makefile to `django42` - I guess I could/should?

It also adds both 3.2 and 4.2 to the test matrixes while we're deploying both between dev/stage/prod.


Makefile-docker
```diff
@@ -1,4 +1,5 @@
 export PYTHON_COMMAND=python3
+export DJANGO_VERSION?=django32
```
diox, line 2: > I didn't default the local makefile to `django42` - I guess I could/should?

Yeah, let's default to 4.2 here.

diox (commented):
Simpler proposal: upgrade to 4.2 unconditionally but merge after today's tag. That way we get 4.2 on dev this week, then automatically on stage and prod next week ?

eviljeff:
Default is now `django42`, but the test matrixes, and the ability to switch back, are still there.

diox (approved):

PR #20636, opened 2023-05-02
Stop doing phrase matching for single-term queries in search

### LINKED ISSUE #20635: Stop doing phrase matching for single-term queries in search
When searching, we currently perform 4 kind of full text search against add-on names:
- a `MatchPhrase`
- a `Match`
- a `MatchPrefix`
- a `FuzzyMatch` (combined with a `trigrams` search, whichever returns the highest score)

The `MatchPhrase` seems to be performing poorly in some cases, being a little too lenient, allowing matches that are not very relevant. This is particularly noticeable for single-term queries, and this is also where a `MatchPhrase` makes the least sense anyway, the other matches should be enough in that case, so let's stop doing it when there is whitespace in the search query.

diox:
Also stop decompounding the word "all" to avoid decompounding "wallet" and "paywall". In most names where "all" is present with no whitespace (already a rare occurrence), it's at the end (so our prefix matching is enough) and/or with a word we are already decompounding anyway (download, browser, tab, etc).

Fixes #20635

### DIFF
--- a/src/olympia/search/tests/test_search_ranking.py
+++ b/src/olympia/search/tests/test_search_ranking.py
@@ -30,7 +30,6 @@ class TestRankingScenarios(ESTestCase):
         response = self.client.get(url, params)
         assert response.status_code == 200
         data = json.loads(force_str(response.content))
-        assert data['count']
         results = data['results']
 
         if expected_lang is None:
@@ -44,6 +43,7 @@ class TestRankingScenarios(ESTestCase):
             query,
             [x['name'][expected_lang] for x in results],
         )
+        assert data['count'] == len(results)
 
         for idx, addon in enumerate(expected):
             expected_name = addon[0]
@@ -66,9 +66,10 @@ class TestRankingScenarios(ESTestCase):
             # sense though!)
             # if found_score != expected_score:
             #     filename = 'src/olympia/search/tests/test_search_ranking.py'
-            #     with open('/code/tmp/sed_me.sh', 'a+') as f:
-            #         f.write('sed -i s/%s/%s/ %s\n' % (
-            #             expected_score, found_score, filename))
+            #     with open('/data/olympia/tmp/sed_me.sh', 'a+') as f:
+            #         f.write(
+            #             'sed -i s/%s/%s/ %s\n' % (expected_score, found_score, filename)
+            #         )
             #     continue
             assert found_score == expected_score, (
                 'Expected "{}" to be on position {} with score {} but '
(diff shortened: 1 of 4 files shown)


eviljeff (commented):
Should we have some tests that check how well a search for "down them all" works with respect to returning the "DownTheAll" extension?

src/olympia/search/tests/test_search_ranking.py
```diff
@@ -97,10 +98,10 @@ def setUpTestData(cls):
             name='Tab Center Redux',
             slug='tab-center-redux',
             summary='Move your tabs to the side of your browser window.',
-            weekly_downloads=915,
+            weekly_downloads=305,
         )
         amo.tests.addon_factory(
-            average_daily_users=468126,
+            average_daily_users=468127,
```
eviljeff, line 104: These `average_daily_users` and `weekly_downloads` changes are unexpected - are the search results excessively wrong now? (Presumably searching for "download" and "down them all" exposed the problem?)

src/olympia/search/tests/test_search_ranking.py
```diff
@@ -993,22 +1005,49 @@ def test_score_boost_exact_match_in_right_language(self):
...
+                ['Tab Center Redux', 692],
+                ['Open Bookmarks in New Tab', 638],
+                ['FoxyTab', 551],
+                ['Open image in a new tab', 524],
+                ['Open Image in New Tab', 410],
+            ),
+        )
+
+    def test_scenario_wallet(self):
+        # Shouldn't be found: some ad-ons have the word "all" which is
```
eviljeff, line 1028:
```suggestion
        # Shouldn't be found: some add-ons have the word "all" which is
```

src/olympia/search/tests/test_search_ranking.py
```diff
@@ -993,22 +1005,49 @@ def test_score_boost_exact_match_in_right_language(self):
...
+        # close but not enough.
+        self._check_scenario('wallet', ())
+
+    def test_downloadthemall(self):
+        self._check_scenario(
+            'down them all',
+            (
+                ['DownThemAll!', 3891],
+                ['All Downloader Professional', 40],
```
eviljeff, line 1037: interesting that this doesn't return any of the other "download" add-ons

eviljeff (approved):

src/olympia/search/tests/test_search_ranking.py
diox: Just an overzealous search&replace on my part. I'll revert them, I don't think it should matter.

src/olympia/search/tests/test_search_ranking.py
diox: It can't guess that "down them all" is about "download". It matches enough words (and I assume trigrams) because "all" and "down" are present.

PR #20641, opened 2023-05-03
limit review queues to upcoming due_dates

### LINKED ISSUE #20622: Reduce the visible depth of the manual review Queue
The current manual review queue displays all extensions that have a due date including those that have a due date in several days.

In order to reduce the mental overload of the reviewers we would like to restrict the number of extensions visible to the reviewer and display only those that are due for the review within next two business days. The total number of extensions should be displayed properly.

eviljeff:
fixes #20622

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -5,7 +5,7 @@ import re
 import time
 import uuid
 
-from datetime import datetime
+from datetime import datetime, timedelta
 from urllib.parse import urlsplit
 
 from django.conf import settings
@@ -81,6 +81,7 @@ from olympia.versions.models import (
     inherit_due_date_if_nominated,
 )
 from olympia.versions.utils import get_review_due_date
+from olympia.zadmin.models import get_config
 
 from . import signals
 
@@ -91,6 +92,7 @@ log = olympia.core.logger.getLogger('z.addons')
 MAX_SLUG_INCREMENT = 999
 SLUG_INCREMENT_SUFFIXES = set(range(1, MAX_SLUG_INCREMENT + 1))
 GUID_REUSE_FORMAT = 'guid-reused-by-pk-{}'
+UPCOMING_DUE_DATE_CUT_OFF_CONFIG_KEY = 'upcoming-due-date-cut-off'
 
 
 class GuidAlreadyDeniedError(RuntimeError):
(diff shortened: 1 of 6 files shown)


wagnerand:
@eviljeff Assuming the following scenario:

A reviewer has permission `Addons:TriageDelayed` but not `Addons:AllDueDates`. Config `upcoming-due-date-cut-off-days` is set to 1.

Will that reviewer see unlisted add-ons to be triaged with a due date further out than 1 day?

src/olympia/addons/models.py
```diff
@@ -91,6 +92,7 @@
 MAX_SLUG_INCREMENT = 999
 SLUG_INCREMENT_SUFFIXES = set(range(1, MAX_SLUG_INCREMENT + 1))
 GUID_REUSE_FORMAT = 'guid-reused-by-pk-{}'
+UPCOMING_DUE_DATE_CUT_OFF_DAYS_CONFIG_KEY = 'upcoming-due-date-cut-off-days'
```
diox, line 95: Do we have to have a config ? It adds a SQL query...

eviljeff: That was the request from @wagnerand as far I understood it.

eviljeff:
> @eviljeff Assuming the following scenario:
> 
> A reviewer has permission `Addons:TriageDelayed` but not `Addons:AllDueDates`. Config `upcoming-due-date-cut-off-days` is set to 1.
> 
> Will that reviewer see unlisted add-ons to be triaged with a due date further out than 1 day?

It's not explicitly tested, but if the `due_date`s of the delayed add-ons (or any version in the queue) is further out than the cut off then no.

If this isn't what was expected can you explain what should happen? (The more different permissions, the more this is a possibility...)

src/olympia/addons/models.py
wagnerand: Yes please. 2 days is too much at the moment, and this allows us to be flexible without a code change and deploy.

diox: We used to have caching for the `get_config` stuff but it was often inconsistent and not that useful so we removed it. I don't think it's particularly worth doing for single-key lookups like that. It's fine we can live with the extra (short) query.

kplotnik:
@wagnerand @eviljeff Ideally we want to show the `Addons:TriageDelayed` extensions independent from the `upcoming-due-date-cut-off-days` but I think for the first iteration we can use the current solution. And then we will see how it evolves.

PR #20642, opened 2023-05-03
Update Remote Settings URLs after GCP migration

leplatrem:
Follow-up of #20593


leplatrem:
@bqbn up?

leplatrem:
@diox up?

diox (approved):

diox:
@leplatrem is this needed in prod right away ? This missed this week's push window, so unless we do an out of schedule push it would be deployed in 2 weeks.

leplatrem:
No that's ok, we'll maintain the domains for a while.

PR #20652, opened 2023-05-05
switch from flake8 to ruff

### LINKED ISSUE #20643: replace flake8 & plugins with ruff
https://github.com/charliermarsh/ruff is a rust based python linter that's much faster than flake8.  We should switch to it for the speeds.

eviljeff:
fixes #20643


eviljeff:
The list of exclusions is long because `flake8-bugbear` had a much more extensive set of rules (`B` rules) than `flake8-mutable` and the additional rules added a lot of fails.  Some/most/all(?) of them are sensible rules though so it'd be worth considering making the necessary changes in a follow-up.

FYI these are the counts of the fails without the `B` rules excluded ([`*`] means `ruff` could auto-fix them):
> olympia@cfbb12810725:~$ ruff check --statistics .
> 98      B904    [` `] Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling
> 55      B015    [` `] Pointless comparison. This comparison does nothing but waste CPU instructions. Either prepend `assert` or remove it.
> 38      B007    [`*`] Loop control variable `addonname` not used within loop body
> 30      B018    [` `] Found useless attribute access. Either assign it to a variable or remove it.
>  8      B010    [`*`] Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
>  5      B026    [` `] Star-arg unpacking after a keyword argument is strongly discouraged
>  4      B017    [` `] `assertRaises(Exception)` should be considered evil
>  4      B905    [` `] `zip()` without an explicit `strict=` parameter
>  3      B023    [` `] Function definition does not bind loop variable `response`
>  1      B004    [` `] Using `hasattr(x, '__call__')` to test if x is callable is unreliable. Use `callable(x)` for consistent results.
>  1      B028    [` `] No explicit `stacklevel` keyword argument found
>  1      B009    [`*`] Do not call `getattr` with a constant attribute value. It is not any safer than normal property access.

Makefile-docker
```diff
@@ -169,7 +169,7 @@ perf-tests: setup-ui-tests
 .PHONY: lint
 lint: ## lint the code
 	black --check src/ tests/
-	flake8 src/ tests/
+	ruff check .
```
eviljeff, line 172: Changed it to check everything because it's super quick

setup.cfg
```diff
@@ -1,19 +0,0 @@
-[flake8]
-# E203: https://github.com/psf/black/issues/315
-# W503: https://www.flake8rules.com/rules/W503.html
-ignore = F999,F405,W503,E203
```
eviljeff, line 4: F999,W503,E203 aren't supported by ruff anyway

src/olympia/addons/tests/test_views.py
```diff
@@ -5598,7 +5598,9 @@ def test_prevent_too_complex_to_determinize_exception(self):
         for i in range(0, 10):
             addon_factory()
         self.refresh()
-        query = '남포역립카페추천 ˇjjtat닷컴ˇ ≡제이제이♠♣ 남포역스파 남포역op남포역유흥≡남포역안마남포역오피 ♠♣'
+        query = (
+            '남포역립카페추천 ˇjjtat닷컴ˇ ≡제이제이♠♣ 남포역스파 남포역op남포역유흥≡남포역안마남포역오피 ♠♣'  # noqa: E501
+        )
```
eviljeff, line 5603: This (and the other 2 changes) are due to https://github.com/charliermarsh/ruff/issues/3825 where ruff's line length is counted by unicode width rather than string length.  However, when I added the `(` and/or split the string over multiple lines Black would auto-format them back again. I will follow-up in that issue - as alignment with Black was one of the justifications - but in the meantime the line exclusion works.

pyproject.toml
```diff
@@ -18,10 +18,37 @@ exclude = '''
...
+    ".npm",
+    "build*.py",
+    "media",
+    "storage",
+    "logs",
+    "site-static",
+    "static",
+    ".git",
+    "*/migrations/*.py",
+]
```
diox, line 38: We should also ignore `deps` and `tmp`, but OTOH this should already be taken care of by `.gitignore`, which `ruff` respects by default (a big reason why it's so fast even without thoes directories in the explicit `exclude` list).

So maybe instead simplify this list to only contain things not in `gitignore` already ?

pyproject.toml
```diff
@@ -18,10 +18,37 @@ exclude = '''
...
+ignore = ["F405", "B004", "B007", "B009", "B010", "B015", "B017", "B018", "B023", "B026", "B028", "B904", "B905"]
```
diox, line 39: We should have an issue to take remove as many of these as possible

pyproject.toml
```diff
@@ -18,10 +18,37 @@ exclude = '''
...
+line-length = 88
+select = [
+    "B",
+    "E",
+    "F",
+    "Q",
+    "W",
```
diox, line 46: We should also have an issue to enable isort (with custom config matching our setup). It was previously disabled because of some weird behavior in the `isort` library.

diox (approved):
I'm generally more conservative when it comes to integrating new shiny tools like that, especially as `flake8` isn't _that_ slow in the grand scheme of things, but `ruff` integrates so many useful features that it's hard to ignore. :shipit:

src/olympia/addons/tests/test_views.py
eviljeff: fwiw, tl;dr from https://github.com/charliermarsh/ruff/issues/3825 is that Black only supports calculating by unicode width in preview mode - and it's currently buggy - so we'll likely need the line exclusion until at least the start of 2024 when the feature moves out of preview in Black.

pyproject.toml
eviljeff: https://github.com/mozilla/addons-server/pull/20652/commits/85f7df9e71ff9d03a84c92df200fe7612b975a3d

Seems like black can also respect gitignore too
https://black.readthedocs.io/en/stable/usage_and_configuration/file_collection_and_discovery.html#gitignore

eviljeff: https://github.com/mozilla/addons-server/pull/20652/commits/8c42181e7312e1b371b4db990a0b938e830ef796 to reduce black's config too

PR #20688, opened 2023-05-11
Record why a version was flagged for human review when that happens

### LINKED ISSUE #20687: Record why a version was flagged for human review
We have various bits of code that can flag a version for review, setting `needs_human_review` boolean to `True`. We should add 2 things when that happens:
- A model recording why that happened, so we could then later make stats on that data
- An `ActivityLog` that could then be displayed in the review page with a human-readable comment explaining why that happened.

The information would be kept even after a reviewer reviews the version, for historical purposes.

Possible reasons (more to come):
- Unknown (not-automated)
- Scanner action
- Add-on entered a promoted group that flags already approved versions (At the moment, only Notable does that)
- Inherited from a previous version
- Sources were provided on a version pending rejection
- Developer reply

diox:
Fixes #20687


src/olympia/reviewers/models.py
```diff
@@ -718,3 +718,52 @@ class Meta:
...
+
+class NeedsHumanReviewHistory(ModelBase):
+    """Model holding information about why a version was flagged for human
+    review. Never intended to be cleared, so that we have historical data even
+    after the version is eventually reviewed by a human."""
+
+    REASON_UNKNOWN = 0
+    REASON_SCANNER_ACTION = 1
+    REASON_PROMOTED_GROUP = 2
+    REASON_HOTNESS_THRESHOLD = 3
```
diox, line 731: Preparing for the future...

src/olympia/addons/models.py
```diff
@@ -669,6 +669,33 @@ def allow_resubmission(self):
     def disable_all_files(self):
         File.objects.filter(version__addon=self).update(status=amo.STATUS_DISABLED)
 
+    def set_needs_human_review_on_latest_versions(self, *, due_date=None, reason):
```
eviljeff, line 672: I know it's valid python... but can we have fields with defaults after fields with no defaults?

src/olympia/addons/models.py
```diff
@@ -669,6 +669,33 @@ def allow_resubmission(self):
...
+        self._set_needs_human_review_on_latest_signed_version(
+            channel=amo.CHANNEL_LISTED, due_date=due_date, reason=reason
+        )
+        self._set_needs_human_review_on_latest_signed_version(
+            channel=amo.CHANNEL_UNLISTED, due_date=due_date, reason=reason
+        )
+
+    def _set_needs_human_review_on_latest_signed_version(
+        self, *, channel, due_date=None, reason
```
eviljeff, line 681: 😬

src/olympia/files/tests/test_models.py
```diff
@@ -197,12 +197,12 @@ def test_generate_filename(self):
         file_ = File.objects.get(id=67442)
         assert (
             file_._meta.get_field('file').upload_to(file_, None)
-            == '15/3615/3615/a3615-2.1.072.zip'  # zip extension.
+            == '15/3615/3615/a3615-2.1.072.xpi'  # xpi extension.
         )
-        file_.is_signed = True
+        file_.is_signed = False
```
eviljeff, line 202: I think I'd have just added a `file_.is_signed = False` above the assert 🤷

src/olympia/addons/tests/test_views.py
```diff
@@ -4182,7 +4211,11 @@ def test_enable_disabled_version(self):
         self.version.file.reload()
         assert not self.version.is_user_disabled
         alog = ActivityLog.objects.exclude(
-            action__in=(amo.LOG.LOG_IN.id, amo.LOG.LOG_IN_API_TOKEN.id)
+            action__in=(
+                amo.LOG.LOG_IN.id,
+                amo.LOG.LOG_IN_API_TOKEN.id,
+                amo.LOG.CHANGE_STATUS.id,
```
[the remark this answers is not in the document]
diox: Previously the status change activity caused by the version creation was "hidden" because the task user didn't exist. I needed the task user to be present for the needs human review activity to be logged, so suddenly this caused the status change activity to be recorded...

src/olympia/files/tests/test_models.py
diox: Doesn't matter in the end, but that would have been smart :sweat_smile:

PR #20689, opened 2023-05-11
add logging for review replies + test

### LINKED ISSUE #20168: Emit `ADD_RATING` for developer replies too (or a specific one)
For new ratings replies, we don't emit an `ActivityLog`:

https://github.com/mozilla/addons-server/blob/d338d271d92c400887f1e3ccf6aea6488cb4d5dc/src/olympia/ratings/models.py#L290-L295

I've added a comment when I refactored the code a few years ago now, but this has always been the case. There is no good reason though, we might be interested in tracking this as well. Possibly through a different activity class to differentiate easily between regular ratings and replies.

GregSherman:
Fixes #20168 

Remove the blocking of ADD_RATING logging for review replies. Update the associated test name & assertion.

### DIFF
--- a/src/olympia/ratings/models.py
+++ b/src/olympia/ratings/models.py
@@ -279,12 +279,10 @@ class Rating(ModelBase):
             else:
                 log.info(f'{action} rating: {instance.pk}')
 
-            # For new ratings - not replies - and all edits (including replies
-            # this time) by users we want to insert a new ActivityLog.
-            new_rating_or_edit = not instance.reply_to or not created
-            if new_rating_or_edit:
-                action = amo.LOG.ADD_RATING if created else amo.LOG.EDIT_RATING
-                activity.log_create(action, instance.addon, instance)
+            # For new ratings, replies, and all edits by users we want 
+            # to insert a new ActivityLog.
+            action = amo.LOG.ADD_RATING if created else amo.LOG.EDIT_RATING
+            activity.log_create(action, instance.addon, instance)
 
             # For new ratings and new replies we want to send an email.
             if created:
--- a/src/olympia/ratings/tests/test_models.py
+++ b/src/olympia/ratings/tests/test_models.py
@@ -229,7 +229,7 @@ class TestRatingModel(TestCase):
         assert email.to == [addon_author.email]
         assert email.from_email == 'Mozilla Add-ons <nobody@mozilla.org>'
 
-    def test_reply_triggers_email_but_no_logging(self):
+    def test_reply_triggers_email_and_logging(self):
         rating = Rating.objects.get(id=1)
         user = user_factory()
         core.set_user(user)
@@ -240,7 +240,7 @@ class TestRatingModel(TestCase):
             body='Rêply',
         )
 
-        assert not ActivityLog.objects.exists()
+        assert ActivityLog.objects.exists()
         assert len(mail.outbox) == 1
         email = mail.outbox[0]
         reply_url = jinja_helpers.absolutify(


src/olympia/ratings/models.py
```diff
@@ -279,12 +279,10 @@ def post_save(sender, instance, created, **kwargs):
...
-            # For new ratings - not replies - and all edits (including replies
-            # this time) by users we want to insert a new ActivityLog.
-            new_rating_or_edit = not instance.reply_to or not created
-            if new_rating_or_edit:
-                action = amo.LOG.ADD_RATING if created else amo.LOG.EDIT_RATING
-                activity.log_create(action, instance.addon, instance)
+            # For new ratings, replies, and all edits by users we want
+            # to insert a new ActivityLog.
+            action = amo.LOG.ADD_RATING if created else amo.LOG.EDIT_RATING
+            activity.log_create(action, instance.addon, instance)
```
eviljeff, line 285: Lets log a different action for replies like mat [suggested](https://github.com/mozilla/addons-server/issues/20168#issue-1528827756) - you can just add a new class to activity.py for `REPLY_RATING` (or whatever)

eviljeff (changes requested):

src/olympia/ratings/models.py
GregSherman: Is there any structure for the log IDs in activity.py or any free one will do?

eviljeff: There probably should be, but there isn't any particular structure. Just add it at the end and pick the next id number based on the last id + 1.

GregSherman:
With dev replies, there is no log for deleting them. If a dev deletes a reply and then re-replies to that same comment, it's considered an edit, is this normal behaviour?

eviljeff:
> With dev replies, there is no log for deleting them. If a dev deletes a reply and then re-replies to that same comment, it's considered an edit, is this normal behaviour?

It's a little unintuitive I agree, but the reasoning is that we a) want to keep deleted replies in general, but b) want at most 1 reply per rating, so a reply would just edit (and undelete) the instance.

eviljeff:
btw, you've got a conflict now because mat just landed a patch that also added an activity action - he's taken id=142

eviljeff:
there's still a conflict

src/olympia/ratings/tests/test_models.py
```diff
@@ -240,7 +240,7 @@ def test_reply_triggers_email_but_no_logging(self):
             body='Rêply',
         )
 
-        assert not ActivityLog.objects.exists()
+        assert ActivityLog.objects.exists()
```
eviljeff, line 243: lets check the object is correct here too (see ` test_edit_triggers_logging_but_no_email` below for roughly what to check)

src/olympia/reviewers/tests/test_views.py
```diff
@@ -3269,7 +3269,8 @@ def test_extra_actions_admin_disable_enable(self):
         assert response.status_code == 200
         doc = pq(response.content)
         assert doc('#clear_auto_approval_delayed_until')  # Is still there.
-        assert doc('#clear_auto_approval_delayed_until_unlisted')  # Is still there.
+        # Is still there.
+        assert doc('#clear_auto_approval_delayed_until_unlisted')
```
eviljeff, line 3273: we don't need this change - have you got the line length set correctly?  (we have line-length set at 88 for black & ruff in pyproject.toml)

eviljeff (changes requested):

PR #20707, opened 2023-05-16
add json endpoint to return per-version ADU stats

### LINKED ISSUE #20700: create view/endpoint to return per-version ADU for reviewer tools
part of https://mozilla-hub.atlassian.net/browse/AMO-70

We want a json endpoint that will return ADU per version for a given addon id.  This is to be consumed by the reviewers tools review page to asynchronously show on the page.  Data should be queried directly from bigquery statistics.

eviljeff:
fixes #20700

### DIFF
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -7,6 +7,7 @@ from django.core.cache import cache
 from django.core.exceptions import PermissionDenied
 from django.core.paginator import Paginator
 from django.db.models import Q
+from django.db.transaction import non_atomic_requests
 from django.http import JsonResponse
 from django.shortcuts import get_object_or_404, redirect
 from django.template.response import TemplateResponse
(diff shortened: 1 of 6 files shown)


src/olympia/reviewers/views.py
```diff
@@ -1517,3 +1520,12 @@ def retrieve(self, request, *args, **kwargs):
             instance=version, data={'parent_version': objs['parent_version']}
         )
         return Response(serializer.data)
+
+
+@bigquery_api_view(json_default=dict)
+@any_reviewer_required
+@reviewer_addon_view_factory
+@non_atomic_requests
+def usage_per_version(request, addon):
```
diox, line 1529: So this is not an API, it uses regular session auth. Is that going to work for us, given that our session auth cookie is sent with `httpOnly` ?

diox: Mmmm yeah that should be fine actually - we shouldn't be trying to _access_ the cookie, and we're on the same domain.

diox (approved):

Issue #20708, opened 2023-05-17
Store a mapping of AMO add-ons and extension IDs from other browsers


willdurand:
For instance, if an add-on is published on both AMO and the Chrome Store, we'd like to have that stored in the AMO database. But not necessarily for all add-ons, only for some add-ons and only AMO admins should be able to edit this mapping for now.

(we'll need to think about admin permissions later, potentially)

AlexandraMoga:
The new tool is located at https://addons-internal.stage.mozaws.net/en-US/admin/models/addons/addonbrowsermapping/
I can add and delete an addon browser mapping  through the new tool with an admin account.

I'm marking this issue as verified fixed on stage.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-359

PR #20709, opened 2023-05-17
Add a AddonBrowserMapping model and admin

willdurand:
Fixes https://github.com/mozilla/addons-server/issues/20708

---

This PR adds both a model and the django admin for it. As mentioned in
the issue linked above, there is no specific permission for this admin
at the moment so admins should be the only ones allowed to access it.


src/olympia/addons/tests/test_admin.py
```diff
@@ -1118,3 +1124,101 @@ def test_can_delete(self):
             f"[{restriction.addon.id}] deleted: ['FR']"
         )
         assert mail.outbox[0].to == ['amo-admins@mozilla.com']
+
+
+class TestAddonBrowserMappingAdmin(TestCase):
+    def setUp(self):
+        user = user_factory(email='someone@mozilla.com')
+        self.grant_permission(user, '*:*')
```
diox, line 1132: We should start with a more specific permission. I suggest re-using `ADMIN_CURATION` (need to edit `DJANGO_PERMISSIONS_MAPPING`).

src/olympia/addons/admin.py
```diff
@@ -649,6 +649,22 @@ def save_model(self, request, obj, form, change):
...
 
+@admin.register(models.AddonBrowserMapping)
+class AddonBrowserMappingAdmin(AMOModelAdmin):
+    list_display = ('addon__name', 'browser', 'extension_id', 'created', 'modified')
+    fields = ('addon', 'browser', 'extension_id')
+    raw_id_fields = ('addon',)
+    readonly_fields = ('created', 'modified')
+
+    def get_readonly_fields(self, request, obj=None):
+        return self.readonly_fields + (('addon',) if obj else ())
```
diox, line 660: Do we really care ? I understand this is not a change to be done lightly but ff there is a mistake it'd be nice to be able to correct it without having to delete and re-create the entry.

src/olympia/addons/models.py
```diff
@@ -2385,3 +2386,24 @@ class Meta:
     def save(self, *args, **kwargs):
         self.hashed_guid = hashlib.sha256(self.guid.encode()).hexdigest()
         super().save(*args, **kwargs)
+
+
+class AddonBrowserMapping(ModelBase):
+    """
+    Mapping between a Firefox add-on on AMO and an extension ID for a different
+    browser.
```
diox, line 2394: Nit: It's actually the other way around, right ? A mapping between an extension ID for a different browser to a Firefox Add-on on AMO

src/olympia/addons/models.py
```diff
@@ -2385,3 +2386,24 @@ class Meta:
...
+        help_text='Add-on id this item will point to. If you do not know the '
+        'id, paste the slug instead and it will be transformed automatically '
+        'for you. You can also use the magnifying glass to see all the '
+        'available add-ons if you have access to the add-on admin page.',
+    )
+    browser = models.PositiveSmallIntegerField(choices=BROWSERS.items())
+    extension_id = models.CharField(max_length=255, null=False, db_index=True)
+
+    class Meta:
+        unique_together = ('addon', 'browser', 'extension_id')
```
diox, line 2409: I think the actual unique constraint we want is `browser`, `extension_id`. We shouldn't need to include the `addon` in there, we never want the same browser + extension id combo to point to multiple add-ons.

(See also nitpicking above for the docstring)

willdurand:
It's green again, ready for another review, thanks!

diox (approved):

PR #20726, opened 2023-05-22
show version adu in review page

### LINKED ISSUE #20701: Load and show per-version ADU inline in the review page versions block
part of https://mozilla-hub.atlassian.net/browse/AMO-70 ; depends on https://github.com/mozilla/addons-server/issues/20700

### LINKED ISSUE #20702: Show the highest ADU versions in a block in the review page
part of https://mozilla-hub.atlassian.net/browse/AMO-70

> Display the block that shows the top 10 most used versions
> -    The individual versions in the list have links to the review section that contains the version

eviljeff:
fixes #20701 and fixes #20702

Limitation: as bigquery query doesn't know anything about the versions, only version strings, the js doesn't have a way to know if a version string in the top 10 versions that _isn't_ on the page is in another channel or on a different page. So this patch just linkifies the versions that _are_ on the current page.  (If we do need to know this we can address it in a follow-up issue - it would involve some db querying to match up every adu result with the channel and/or which page it would be shown on when paginated)

Screenshot 2023-05-23 10 20 05

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/addon_details_box.html
+++ b/src/olympia/reviewers/templates/reviewers/addon_details_box.html
@@ -12,7 +12,7 @@
           </a>
         {% endwith %}
       {% else %}
-        <img class="screenshot thumbnail" src="{{ addon.thumbnail_url }}"
+        <img class="screenshot thumbnail" src="{{ addon.get_icon_url(150) }}"
           alt="" width="200" height="150" />
       {% endif %}
     </div>
--- a/static/js/zamboni/reviewers.js
+++ b/static/js/zamboni/reviewers.js
@@ -46,6 +46,10 @@
     initQueue();
   }
 
+  if ($('.version-adu').length > 0) {
+    initVersionsADU();
+  }
+
   // Show add-on ID when icon is clicked
   if ($('#addon[data-id], #persona[data-id]').length) {
     $('#addon .icon').click(function () {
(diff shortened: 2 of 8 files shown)


src/olympia/reviewers/templates/reviewers/addon_details_box.html
```diff
@@ -12,7 +12,7 @@ <h2>{{ addon_name }}</h2>
           </a>
         {% endwith %}
       {% else %}
-        <img class="screenshot thumbnail" src="{{ addon.thumbnail_url }}"
+        <img class="screenshot thumbnail" src="{{ addon.get_icon_url(150) }}"
```
eviljeff, line 15: this is just a drive-by fix because the error showed up in the console each time I re-loaded the page

static/js/zamboni/reviewers.js
```diff
@@ -497,3 +501,60 @@ function initScrollingSidebar() {
...
+
+  function loadVersionsADU() {
+    const aduUrl = $('#addon').data('versions-adu-url');
+    $.get(aduUrl, function (data) {
+      const versionAduPairs = Object.entries(data);
+      fillVersionsTable(versionAduPairs);
+      fillTopTenBox(versionAduPairs);
+    });
+  }
+  setTimeout(loadVersionsADU, 100);
```
diox, line 559: You could call it directly if the stuff at the top was wrapped into `$(document).ready(function () {}` instead. More reliable that way (and more consistent with our other bits of js)

diox (commented):
I wonder if we should move that `Average Daily Users` section to the sidebar on the right to avoid a big page shift when the XHR goes through ?

static/js/zamboni/reviewers.js
eviljeff: our js is far from consistent 😬 but https://github.com/mozilla/addons-server/pull/20726/commits/da22955d13e2272e769e1fee8dccb9e860e96862

diox (approved):

Issue #20714, opened 2023-05-17
Banner for the notable extensions


kplotnik:
We want to show the standard banner to the notable extension group, with the text predefined and set by the operations team. The banner can be displayed permanently till the campaign is over (current estimate 4-6 weeks)

diox:
The banner will use the same spot as the `submit_notification_warning` banner that we already have, but be limited to being displayed for new versions of a notable add-on.

Content will come from the database through a `Config` object like the regular submission warning banner, but with a different key (probably `submit_notification_warning_notable`).

diox:
Yeah, that's expected/on purpose, didn't want to show both at once and I figured the `submit_notification_warning` one was the most important to show, so if it's set it will take precedence.

AlexandraMoga:
Ok, so we don't want/need to show any  banner for regular addons then?

diox:
- If `submit_notification_warning`: show it to everyone, both at add-on and version upload (actually even at channel selection during submission), don't show anything else
- If `submit_notification_warning` is not set, but `submit_notification_warning_pre_review` is set: show it on version upload of add-ons belonging to a group that does human review for the corresponding channel (for notable, it'd be both channels - for recommended, only listed)

AlexandraMoga:
These are my final results:

- with `submit_notification_warning` : the banner is displayed both on the distribution and version upload pages, for all new addon or new version submissions (including the theme submission wizard)
[new addon]
![image](https://github.com/mozilla/addons-server/assets/31961530/81eeb95d-f0b2-479f-ba25-10fe943cfeeb)
[new version]
![image](https://github.com/mozilla/addons-server/assets/31961530/d27953fa-09b3-4ab7-9906-a8d3b97e48d4)

- with both `submit_notification_warning` and `submit_notification_warning_pre_review `: same as above (i.e. `submit_notification_warning` takes precedence)
- with `submit_notification_warning_pre_review ` only: 
  - regular addons: no banner is displayed for new addon or new version uploads
  - notable addons - pre review banner is displayed on the version upload page for both listed and unlisted distribution channels; the banner will not be displayed on the channel selection page in this case
  ![image](https://github.com/mozilla/addons-server/assets/31961530/c01bb638-059d-4154-8620-660fd61938ae)

  - recommended and by Firefox addons: pre review banner is displayed on the version upload page for new _listed_ versions only; the banner will not be displayed on the channel selection page in this case  
[...]

PR #20727, opened 2023-05-22
Show submission warning banner to add-ons belonging to promoted groups

diox:
Fixes https://github.com/mozilla/addons-server/issues/20714

### DIFF
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -1470,7 +1470,7 @@ def _submit_upload(request, addon, channel, next_view, wizard=False):
         else []
     )
     submit_notification_warning = get_config('submit_notification_warning')
-    if not submit_notification_warning:
+    if not submit_notification_warning and addon:
         # If we're not showing the generic submit notification warning, show
         # one specific to pre review if the developer would be affected because
         # of its promoted group.
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -1469,6 +1469,18 @@ def _submit_upload(request, addon, channel, next_view, wizard=False):
         if existing_properties
         else []
     )
+    submit_notification_warning = get_config('submit_notification_warning')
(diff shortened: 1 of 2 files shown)


src/olympia/devhub/views.py
```diff
@@ -1469,14 +1469,26 @@ def _submit_upload(request, addon, channel, next_view, wizard=False):
         if existing_properties
         else []
     )
+    submit_notification_warning = get_config('submit_notification_warning')
```
wagnerand, line 1472: Do we need to support showing both the general and the new banner?

src/olympia/devhub/views.py
```diff
@@ -1469,14 +1469,26 @@ def _submit_upload(request, addon, channel, next_view, wizard=False):
...
+        # If we're not showing the generic submit notification warning, show
+        # one specific to pre review if the developer would be affected because
+        # of its promoted group.
+        promoted_group = addon.promoted_group(currently_approved=False)
+        if (channel == amo.CHANNEL_LISTED and promoted_group.listed_pre_review) or (
+            channel == amo.CHANNEL_UNLISTED and promoted_group.unlisted_pre_review
+        ):
+            submit_notification_warning = get_config(
+                'submit_notification_warning_pre_review'
+            )
```
eviljeff, line 1483: The wording of this config is going to have to be a little tricky if it has to cover both groups that have `listed_pre_review` but not `unlisted_pre_review`, and groups that have both.  If we do want to have this for all groups (and I can see why that's a good idea rather than special casing it to Notable) then perhaps we want two new configs so the `listed_pre_review` only group can mention listed versions specifically?

diox: I figured since we are only showing it at the upload step, then it was ok to use a generic "pre_review" key that didn't specify the channel: we know the channel already from the previous step, so we're only showing the banner where relevant.

PR #20742, opened 2023-05-24
Add missing migrations

willdurand:
These migrations have been generated when I ran `makemigrations` for a different thing.


willdurand:
hum, looks like it will require a bit more work that I thought..

willdurand:
I removed the migration that removes the `reviewed` field on `File` for now.

src/olympia/addons/migrations/0044_remove_addonreviewerflags_notified_about_auto_approval_delay_and_more.py
```diff
@@ -0,0 +1,184 @@
...
+    dependencies = [
+        ('translations', '0001_initial'),
+        ('addons', '0043_addonreviewerflags_auto_approval_delayed_until_unlisted'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='addonreviewerflags',
+            name='notified_about_auto_approval_delay',
+        ),
```
diox, line 18: Note: this is the only one that will do anything in the database (which is fine)

diox (approved):

PR #20743, opened 2023-05-24
Clarify that authentication is not mandatory in API calls

diox:
The previous wording sort of implied you had to make authenticated calls using JWT at all time.

### DIFF
--- a/docs/topics/api/auth.rst
+++ b/docs/topics/api/auth.rst
@@ -4,8 +4,10 @@
 Authentication (External)
 =========================
 
-To execute authenticated calls against the API as an external consumer, you need to
-include a `JSON Web Token (JWT)`_ in the ``Authorization`` header for every request.
+To execute authenticated calls against the API as an external consumer (See the
+documentation for individual endpoints where it will indicate if that endpoint
+requires, or would benefit from, authentication), you need to include a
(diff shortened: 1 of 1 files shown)


docs/topics/api/auth.rst
```diff
@@ -4,8 +4,8 @@
 Authentication (External)
 =========================
 
-To access the API as an external consumer, you need to include a
```
diox, line 7: That might not be enough... Maybe we should have a paragraph like `Each endpoint should specify whether it needs or would benefit from authenticated calls in some way.` before that ?

eviljeff: Yeah, add a follow-on sentence saying something like that.  Suggestion: `See the documentation for individual endpoints where it will indicate if that endpoint requires, or would benefit from, authentication.`

diox: It felt a little awkward after the sentence, so added that bit inside parenthesis. Let me know what you think.

docs/topics/api/auth.rst
```diff
@@ -4,7 +4,9 @@
...
+To execute authenticated calls against the API as an external consumer (See the
```
eviljeff, line 7: nit
```suggestion
To execute authenticated calls against the API as an external consumer, (see the
```

eviljeff (approved):

PR #20747, opened 2023-05-24
Handle moderation queue case for reviewer queues

### LINKED ISSUE #20721: Make moderation review queue consistent with the others
Follow up to #20717 

Moderation reviewer queue is constructed differently from the others as it does not have a table class in utils.py. It's currently treated as an edge case, but it makes sense to treat it like the other queues.

GregSherman:
Fixes #20721 

Make a table object in utils with the attributes required to be consistent with other queues. Put it in the registry so that the mod queue case is handled everywhere else.

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -370,7 +370,7 @@ class MadReviewTable(AddonQueueTable):
         )
 
 
-class ModerationQueueFields:
+class ModerationQueueTable:
     title = 'Rating Reviews'
     urlname = 'queue_moderated'
     url = r'^reviews$'
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -369,6 +369,14 @@ class MadReviewTable(AddonQueueTable):
         )
 
 
+class ModerationQueueFields:
+    title = 'Rating Reviews'
+    urlname = 'queue_moderated'
+    url = r'^reviews$'
+    permission = amo.permissions.RATINGS_MODERATE
+    show_count_in_dashboard = False
+
+
 class ReviewHelper:
     """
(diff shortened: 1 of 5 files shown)


GregSherman:
@diox @eviljeff Requesting review

src/olympia/reviewers/utils.py
```diff
@@ -66,6 +66,7 @@ class AddonQueueTable(tables.Table):
         accessor='_current_version__autoapprovalsummary__score',
     )
     show_count_in_dashboard = True
+    view = None
```
eviljeff, line 69: Can we define `view = queue` directly here? Or do we come up against circular reference errors?

GregSherman: importing the functions from views leads to a total crash so I assume it's circular reference.

eviljeff: idea: replace `views.reviewer_tables_registry[queue].view` in urls.py with `getattr(views, reviewer_tables_registry[queue].view_name)()` and change `view` in the table classes to be a string called `view_name`.

(other hacks may be available!)

src/olympia/reviewers/views.py
```diff
@@ -305,6 +306,9 @@ def save_motd(request):
 
 
 def queue(request, tab):
+    # if tab == 'moderated':
+    #     return queue_moderated(request, tab)
+
```
diox, line 311: Can be removed

src/olympia/reviewers/utils.py
```diff
@@ -369,6 +370,15 @@ def get_queryset(cls, request, **kw):
         )
 
 
+class ModerationQueueFields:
```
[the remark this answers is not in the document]
eviljeff: agree

PR #20754, opened 2023-05-26
Add a view to redirect a given version to the correct page on /review

### LINKED ISSUE #20732: Link top 10 version ADU to other review pages
In https://github.com/mozilla/addons-server/issues/20702 a block was added to the review page to show the top 10 versions by ADU.  But the underlying bigquery doesn't know anything about the versions, only version strings, and we just return it to the client side JS.  So the js doesn't have a way to know if a version string in the top 10 versions that isn't on the page is in another channel or on a different page. 

We need to add to the response so the js is able to match up every adu result with the channel and/or which page it would be shown on when paginated.

eviljeff:
fixes #20732 

Screenshot 2023-05-25 at 17-34-16 facebook-container – Add-ons for Firefox


src/olympia/reviewers/views.py
```diff
@@ -1519,3 +1524,34 @@ def usage_per_version(request, addon):
     )
     patch_cache_control(response, max_age=5 * 60)
     return response
+
+
+@any_reviewer_required
+@reviewer_addon_view_factory
+@non_atomic_requests
+def review_version_redirect(request, addon, version):
+    addon_versions = list(
```
diox, line 1533: Shouldn't need to do cast into a `list()`. Avoiding it means we do once fewer loop on all the versions.

src/olympia/reviewers/tests/test_views.py
```diff
@@ -8779,3 +8780,83 @@ def test_bigquery_disabled(self, get_adu_per_version_mock):
...
+
+
+class TestReviewVersionRedirect(ReviewerTest):
+    def login_as_reviewer(self):
+        self.grant_permission(
+            UserProfile.objects.get(email='reviewer@mozilla.com'),
+            'Addons:ReviewUnlisted',
+        )
+        return super().login_as_reviewer()
+
```
diox, line 8810: Needs a additional (probably separate for clarify) 404 test.

src/olympia/reviewers/tests/test_views.py
```diff
@@ -8779,3 +8780,83 @@ def test_bigquery_disabled(self, get_adu_per_version_mock):
...
+    def test_responses(self):
+        addon = addon_factory()
+        listed = addon.current_version
+        unlisted = version_factory(addon=addon, channel=amo.CHANNEL_UNLISTED)
+        deleted = version_factory(addon=addon)
+        deleted.delete()
```
diox, line 8816: Add another add-on to make sure we only take the requested add-on into consideration

diox (approved):

src/olympia/reviewers/views.py
eviljeff: https://github.com/mozilla/addons-server/pull/20754/commits/afc49f9ac3ae4435d08cdd4acec0bc5b4e727243 + I changed the list generator into a generator function too, for some possible efficiency with large result sets.

PR #20769, opened 2023-05-29
Support browser_specific_settings.gecko_android manifest key

### LINKED ISSUE #20530: Determine Android compatibility range with `browser_specific_settings.gecko_android` if present in manifest
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1824237

The new `browser_specific_settings.gecko_android` should determine Android compatibility. It accepts the optional `strict_min_version` and `strict_max_version`, overriding values from `browser_specific_settings.gecko`.

diox:
Note: if an add-on uses `gecko_android`, AMO will force the  minimum Firefox for Android version it's compatible with to be 113.0 (or higher if it was specified). That matches the first Firefox for Android version that was released with support for that manifest key.

Fixes #20530

### DIFF
--- a/src/olympia/files/utils.py
+++ b/src/olympia/files/utils.py
@@ -229,13 +229,14 @@ class ManifestJSONExtractor:
     def get_strict_version_for(self, *, key, application):
         strict_key = f'strict_{key}_version'
         if application == amo.FIREFOX:
-            return get_simple_version(self.gecko.get(strict_key))
+            strict_version_value = self.gecko.get(strict_key)
         elif application == amo.ANDROID:
-            return get_simple_version(
-                self.gecko_android.get(strict_key, self.gecko.get(strict_key))
+            strict_version_value = self.gecko_android.get(
+                strict_key, self.gecko.get(strict_key)
             )
         else:
-            return get_simple_version(None)
+            strict_version_value = None
+        return get_simple_version(strict_version_value)
 
     @property
     def install_origins(self):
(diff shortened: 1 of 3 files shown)


src/olympia/files/utils.py
```diff
@@ -229,13 +226,16 @@ def type(self):
...
+    def get_strict_version_for(self, *, key, application):
+        strict_key = f'strict_{key}_version'
+        if application == amo.FIREFOX:
+            return get_simple_version(self.gecko.get(strict_key))
+        elif application == amo.ANDROID:
+            return get_simple_version(
+                self.gecko_android.get(strict_key, self.gecko.get(strict_key))
+            )
+        else:
+            return get_simple_version(None)
```
eviljeff, line 238: I found this function a little difficult to understand with the multiple layers of `get`.  Do you think something more like the following works? 
```suggestion
    def get_strict_version_for(self, *, key, application):
        strict_key = f'strict_{key}_version'
        if application == amo.FIREFOX:
            strict_version_string = self.gecko.get(strict_key)
        elif application == amo.ANDROID:
            strict_version_string = self.gecko_android.get(strict_key, self.gecko.get(strict_key))
        else:
            strict_version_string = None
        return get_simple_version(strict_version_string)
```
(might need reformatting.  I looked for a way to avoid doing `self.gecko.get(strict_key)` twice but there wasn't an obvious way to avoid it)

eviljeff (approved):

PR #20770, opened 2023-05-29
Rating sort param consistency in addons search

### LINKED ISSUE #20763: Inconsistency in rating sort & filter query parameters

On the [`/api/v5/addons/search`](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#get--api-v5-addons-search-) API endpoint, one can filter & sort add-ons by rating.
For example, a query used for searching add-ons with an average rating of 4 and sorting the results by rating would like this:
- `/api/v5/addons/search/?ratings__gte=4&sort=rating`

### What happened?

The query parameters for filtering & sorting add-ons by ratings aren't named consistently which can be confusing & error-prone, especially since, in the endpoint documentation, sort parameters & threshold style filter parameters are presented next to each other.

<details>
<summary>View docs screenshot</summary>

Screenshot from 2023-05-27 23-53-47

</details>

The ratings related object property & sort/filter query parameters should be named consistently.

`ratings` being the property actually returned in the [addon details object](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#addon-detail-object), it might be preferable to:
- support the `sort=ratings` query parameter.
- keep the `sort=rating` query parameter as an alias.
- update documentation to mention only `sort=ratings` query parameter for consistency / DX.
- make the "sorting parameters" & "threshold style filtering parameters" sections in the endpoint documentation stand out more from each other.

Additionally, glancing at [src/olympa/search/filters.py](https://github.com/mozilla/addons-server/blob/5d36354b061a79a1762592fd137051b3fe9e5d5f/src/olympia/search/filters.py#L979), it appears the filtering & sorting logic currently do not leverage the same ratings field: the `ratings` filter relies on the `ratings.average` field, whereas the `rating` sort relies on `bayesian_rating`, which might or might not be intentional, may lead to odd behavior.

ziir:
- Add `sort=ratings` sorting parameter for consistency
- Add relevant test & update existing tests:
  - Ensure "sort=ratings" is supported, preferred
  - Ensure "sort=rating" is still supported, for backwards compatibility
  - Update existing tests using "sort=rating" to use the new preferred parameter
 - Update add-ons search endpoint ratings sort documentation

Fixes #20763


docs/topics/api/addons.rst
```diff
@@ -53,7 +53,7 @@ This endpoint allows you to search through public add-ons.
            hotness  Hotness (average number of users progression), descending.
             random  Random ordering. Only available when no search query is
                     passed and when filtering to only return promoted add-ons.
-            rating  Bayesian rating, descending.
+           ratings  Bayesian rating, descending.
```
diox, line 56: I think it's confusing enough that we can document both (indicating that old one is deprecated/only there for backwards-compatibility), in case people stumble upon old code using `rating`.

diox (commented):
Looks good. 

We've got a manually curated changelog in https://github.com/mozilla/addons-server/blob/master/docs/topics/api/overview.rst that we should use to document this. Something like:
```
rename sort by ratings parameter in add-on search API to ``ratings``, keeping support ``rating`` for backwards-compatibility.
```

docs/topics/api/addons.rst
ziir: Agreed, fixed:

```diff
-            rating  Bayesian rating, descending.
+           ratings  Bayesian rating, descending (preferred).
+            rating  Bayesian rating, descending (deprecated, backwards-compatibility only).
```

diox: The deprecated warning is fine, but I don't think we need the `(preferred)` bit on the other one.

ziir: Done

PR #20784, opened 2023-06-01
Lock compatibility information if it came from `gecko_android`

### LINKED ISSUE #20531: Disallow changing compatibility info of a version in the Developer Hub if compatibility info was present in manifest
https://docs.google.com/document/d/1Uou3WEI6W0fwLpzipLMEZHycWOUGWiDQAbVTAPemaYo/edit#heading=h.c75pefc0iki4

- When the `browser_specific_settings.gecko_android` key is present in the manifest, AMO MUST automatically mark this add-on as compatible with Android. The compatibility range will be determined by the `strict_min_version` and `strict_max_version` specified in both `browser_specific_settings.gecko` and `browser_specific_settings.gecko_android`, if any (`gecko_android` takes precedence)
- When the `browser_specific_settings.gecko_android` key is present in the manifest, AMO SHOULD display the values in devhub, but MUST no longer allow modification from what is in the manifest.
- When the `browser_specific_settings.gecko_android` key is *not* present in the manifest, AMO SHOULD make no assumptions about the compatibility of the add-on with Android. Therefore, it MUST show the checkbox to mark the add-on as compatible with Android during submission.

This should probably be done by adding a new field in `ApplicationsVersions` specifying that the min and max values were imported from the manifest automatically, so that they cannot be changed. This should only apply to Android, as just the presence of the `gecko` key isn't enough to determine whether something is compatible with Desktop at this point.

diox:
Manifest should be the source of truth for compatibility data for Android when `gecko_android` if set.

Fixes #20531


src/olympia/versions/models.py
```diff
@@ -457,19 +457,34 @@ def from_upload(
...
-                # which will have this add-on then be listed only for
-                # Firefox specifically.
+        for avs_from_parsed_data in parsed_data.get('apps', []):
+            application = APP_IDS.get(avs_from_parsed_data.application)
+            if not application:
                 continue
-            avs = compatibility[parsed_app.appdata]
+            elif avs_from_parsed_data.locked_from_manifest:
+                # In that case, the manifest takes precedence.
+                avs = avs_from_parsed_data
```
eviljeff, line 466: I can't do this as a suggestion because there's a deleted line in the middle, but 
```python
        for avs_from_parsed_data in parsed_data.get('apps', []):
            if avs_from_parsed_data.locked_from_manifest:
                # In that case, the manifest takes precedence.
                avs = avs_from_parsed_data
            application = APP_IDS.get(avs_from_parsed_data.application)
            if application not in compatibility:
                # In that case, the developer didn't include compatibility
                # information we have to follow, and they didn't select this
                # app at upload time, so we ignore that app.
                continue
            else:
```

eviljeff (approved):

src/olympia/addons/serializers.py
```diff
@@ -344,11 +344,14 @@ class SimpleVersionSerializer(MinimalVersionSerializer):
     compatibility = VersionCompatibilityField(
         # default to just Desktop Firefox; most of the times developers don't develop
         # their WebExtensions for Android.  See https://bit.ly/2QaMicU
+        # Note that if the manifest contains `gecko_android`, this will be
+        # overridden and an ApplicationsVersions will be created for Android at
+        # upload.
```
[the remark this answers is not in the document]
diox: That's in addition, I will clarify

src/olympia/devhub/templates/devhub/versions/edit.html
```diff
@@ -34,9 +34,7 @@ <h3>{{ title }}</h3>
           {% if compat_form %}
           <tr>
             <th>{{ tip(_("Compatibility"),
-                       _("Adjusting application information here will allow users to install your "
-                         "add-on even if the install manifest in the package indicates that the "
-                         "add-on is incompatible.")) }}</th>
+                       _("Which versions of what applications your add-on is compatible with. It is not editable if you specified compatibility information in your manifest already.")) }}</th>
```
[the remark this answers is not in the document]
diox: Actually I was thinking of reverting this label entirely. It's only locked if you used `gecko_android`, which you should know if you used it... It's not a great place to add links because of the way we display the tooltip on hover only (it would disappear as you try to click on the link)

static/js/common/upload-addon.js
```diff
@@ -474,6 +474,12 @@
             $('.binary-source').hide();
             $('.compatible-apps').hide();
           } else {
+            // Pre-check Android or not depending on what we detected in the
+            // manifest.
+            $('#id_compatible_apps .android input').prop(
+              'checked',
+              results.explicitly_compatible_with_android,
```
willdurand, line 481: Can we uncheck this checkbox?

src/olympia/devhub/templates/devhub/versions/edit.html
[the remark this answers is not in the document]
diox: b26ea73ccd363a587c1ab11812f04aa28e9ab6c3

src/olympia/versions/models.py
diox: bc7178d6d999e9cd73e9422fcccc60e9f8b787b5

static/js/common/upload-addon.js
diox: Ugh I forgot that `readonly` doesn't apply to checkboxes. I'm not sure `disabled` would be the right fix, it needs more testing. I's not blocking this PR since the checkbox will be completely ignored for, it's purely to be nice to the developer.

diox: 81910a335def5a2449a22bf0541db2798238fac7

PR #20797, opened 2023-06-06
Replace NOTABLE_ADU_LIMIT_CONFIG_KEY with a UsageTier

diox:
Fixes mozilla/addons#9184

### DIFF
--- a/src/olympia/promoted/tests/test_tasks.py
+++ b/src/olympia/promoted/tests/test_tasks.py
@@ -13,7 +13,7 @@ from olympia.reviewers.models import UsageTier
 from olympia.versions.utils import get_staggered_review_due_date_generator
 from olympia.zadmin.models import set_config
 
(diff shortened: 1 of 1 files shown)


src/olympia/promoted/tasks.py
```diff
@@ -6,30 +6,33 @@
...
 
 
 @task
 @use_primary_db
 def add_high_adu_extensions_to_notable():
     """Add add-ons with high ADU to Notable promoted group."""
-    adu_limit = get_config(NOTABLE_ADU_LIMIT_CONFIG_KEY)
-    if not adu_limit:
+    try:
+        lower_adu_threshold = UsageTier.objects.get(slug='notable').lower_adu_threshold
```
eviljeff, line 25: Should have picked this up at the time, but we're not using the constant `NOTABLE_TIER_SLUG` here 🙃

PR #20800, opened 2023-06-06
Add an API endpoint for the browser mapping

### LINKED ISSUE #20798: Add an API endpoint for the browser mapping
Once we have https://github.com/mozilla/addons-server/issues/20708, we should add a new API endpoint for Firefox to consume. This API should accept a browser identifier and return the mapping for said browser. We probably want a reduced "view" of an add-on for each entry (at least the `guid` and the download URL of the `current_version`).

willdurand:
Fixes #20798

### DIFF
--- a/docs/topics/api/addons.rst
+++ b/docs/topics/api/addons.rst
@@ -974,3 +974,23 @@ Four recommendations are fetched, but only valid, publicly available addons are 
     :>json string next: The URL of the next page of results.
     :>json string previous: The URL of the previous page of results.
     :>json array results: An array of :ref:`add-ons <addon-detail-object>`. The following fields are omitted for performance reasons: ``release_notes`` and ``license`` fields on ``current_version`` and ``current_beta_version``, as well as ``picture_url`` from ``authors``.
+
+
+----------------
+Browser Mappings
+----------------
+
+.. _addon-browser-mappings:
+
+This endpoint provides browser mappings of non-Firefox and Firefox extensions.  Added to support the extensions import feature in Firefox.
+
+.. http:get:: /api/v5/addons/browser-mappings/
+
+    .. note::
+        This endpoint uses a larger ``page_size`` than most other API endpoints.
+
+    :query string browser: The browser identifier for this query (required). Must be one of these: ``chrome``.
+    :query int page_size: Maximum number of results to return for the requested page. Defaults to 100.
+    :>json array results: An array containing a mapping of non-Firefox and Firefox extension IDs for a given browser.
+    :>json string results[].extension_id: A non-Firefox extension ID.
+    :>json string results[].addon_guid: The corresponding Firefox add-on ``guid``.
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -65,6 +65,7 @@ from .fields import (
 from .models import (
     Addon,
     AddonApprovalsCounter,
+    AddonBrowserMapping,
     AddonUser,
     AddonUserPendingConfirmation,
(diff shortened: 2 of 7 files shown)


src/olympia/addons/serializers.py
```diff
@@ -1648,3 +1649,14 @@ def get_replacement(self, obj):
                 coll_match.group('user_id'), coll_match.group('coll_slug')
             )
         return []
+
+
+class AddonBrowserMappingSerializer(AMOModelSerializer):
+    addon_guid = serializers.CharField(source='addon__guid')
```
diox, line 1655: Add a comment explaining that this is coming from an annotation (since it can be a little surprising)

src/olympia/addons/views.py
```diff
@@ -1239,3 +1247,34 @@ def filter_queryset(self, qs):
...
+            (
+                key
+                for key in BROWSERS
+                if BROWSERS.get(key) == self.request.query_params.get('browser')
+            ),
+            None,
+        )
+        return AddonBrowserMapping.objects.filter(
+            browser=browser, addon__status__in=amo.APPROVED_STATUSES
+        ).annotate(addon__guid=F('addon__guid'))
```
diox, line 1275: Add a comment mentioning this annotation is used in the serializer and allows us to avoid useless SQL queries since we only care about the `guid`

docs/topics/api/addons.rst
```diff
@@ -974,3 +974,19 @@ Four recommendations are fetched, but only valid, publicly available addons are
...
     :>json array results: An array of :ref:`add-ons <addon-detail-object>`. The following fields are omitted for performance reasons: ``release_notes`` and ``license`` fields on ``current_version`` and ``current_beta_version``, as well as ``picture_url`` from ``authors``.
+
+
+----------------
+Browser Mappings
+----------------
+
+.. _addon-browser-mappings:
+
+This endpoint provides browser mappings of non-Firefox and Firefox extensions.  Added to support the extensions import feature in Firefox.
```
diox, line 985: We should document the special `page_size` (and max) that this endpoint uses.

src/olympia/addons/views.py
```diff
@@ -1239,3 +1247,34 @@ def filter_queryset(self, qs):
...
+    authentication_classes = []
+    serializer_class = AddonBrowserMappingSerializer
+    pagination_class = LargePageNumberPagination
+
+    def get(self, request, format=None):
+        browser = self.request.query_params.get('browser', None)
+        if browser not in list(BROWSERS.values()):
+            raise exceptions.ParseError('Invalid browser parameter')
+
+        return super().get(request, format)
```
diox, line 1262: Before returning the response from `super().get()`, add some HTTP caching. I think 24 hours should be fine ? Also add an assertion for that in `test_get` (and maybe one checking that we don't have that caching when the parameter is invalid).

willdurand:
Thanks for the review. PR updated.

src/olympia/addons/tests/test_views.py
```diff
@@ -7272,3 +7274,70 @@ def test_developer_role(self):
...
+
+            results = self.assert_json_results(res, 2)
+            assert results[0] == {
+                'extension_id': extension_id_1,
+                'addon_guid': addon_1.guid,
+            }
+            assert results[1] == {
+                'extension_id': extension_id_2,
+                'addon_guid': addon_2.guid,
+            }
```
diox, line 7343: nit: only the `res = self.client.get(...` needs to be nested inside the `with self.assertNumQueries()` context manager.

diox (approved):

PR #20803, opened 2023-06-07
Add an hidden input to indicate android compatibility when it's forced at upload

### LINKED ISSUE #20531: Disallow changing compatibility info of a version in the Developer Hub if compatibility info was present in manifest

diox:
Fixes #20531

### DIFF
--- a/static/js/common/upload-addon.js
+++ b/static/js/common/upload-addon.js
@@ -495,6 +495,9 @@
             });
           }, 1000);
         } else {
+          // Remove hidden android compatibility input if present (it will
+          // be re-added if necessary)
+          $('#id_compatible_apps_hidden_android').remove();
           if (results.addon_type != 1) {
             // No source or compat app selection for themes/dicts/langpacks.
             $('.binary-source').hide();
@@ -502,7 +505,8 @@
           } else {
             // Pre-check Android or not depending on what we detected in the
             // manifest.
-            $('#id_compatible_apps .android input')
+            let $checkbox = $('#id_compatible_apps .android input');
+            $checkbox
               .prop('checked', results.explicitly_compatible_with_android)
               .prop('disabled', results.explicitly_compatible_with_android)
               .parent()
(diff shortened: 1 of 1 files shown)


static/js/common/upload-addon.js
```diff
@@ -514,6 +518,19 @@
                     )
                   : '',
               );
+            // In addition, if we automatically ticked and disabled the Android
+            // checkbox, the browser won't submit the value. It's fine if
+            // Firefox was also checked, but if not then we'd not submit
+            // anything and the validation would fail server-side, so we
+            // add/remove an hidden input to compensate.
```
diox, line 525: Let me know if that's not clear. The disabled checkbox is really just for showing the developer that the add-on is going to be marked as compatible with Android, it's not submitted since it's disabled. Unfortunately `readonly` is not supported for checkboxes by design in the spec...

eviljeff: it's a hack but I can't think of another way

eviljeff (approved):

PR #20805, opened 2023-06-07
Add two trusted GCP domains

### LINKED ISSUE #20804: Add a couple of trusted domains

We'd like to add two trusted domains to the ALLOWED_HOSTS list, aka `amo.nonprod.webservices.mozgcp.net` and `amo.prod.webservices.mozgcp.net`. These are domains managed and controlled by us and in our GCP projects.

### What happened?
We're moving AMO to GCP, thus adding those two domains.

Add the said two domains to the ALLOWED_HOSTS list.

n/a

bqbn:
Fixes #20804 

Since we're moving AMO to GCP, we'd like to add two trusted domains to the ALLOWED_HOSTS list. These are domains managed and controlled by us and in our GCP projects.

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -27,6 +27,8 @@ if os.path.exists(ENVIRON_SETTINGS_FILE_PATH):
 
 ALLOWED_HOSTS = [
     '.allizom.org',
+    '.amo.nonprod.webservices.mozgcp.net',
+    '.amo.prod.webservices.mozgcp.net',
     '.mozilla.org',
     '.mozilla.com',
     '.mozilla.net',


diox (approved):

src/olympia/lib/settings_base.py
```diff
@@ -27,6 +27,8 @@
 
 ALLOWED_HOSTS = [
     '.allizom.org',
+    '.amo.nonprod.webservices.mozgcp.net',
+    '.amo.prod.webservices.mozgcp.net',
```
eviljeff, line 31: I'm aware we don't do this currently, but maybe we should define these per env?  i.e. don't allow the non-prod (`.allizom.org`, `.amo.nonprod.webservices.mozgcp.net`) hosts for the prod environment?

diox: Yeah, we should have a follow-up to tighten that up a bit per env.

bqbn: I think that's a good idea in general and worth an issue of its own. @eviljeff would you create one for it?

Meanwhile, I'm going to merge this to get us unblocked on a few tests.

eviljeff: #20806

PR #20828, opened 2023-06-14
Lock Android compatibility when gecko_android is present in manifest,…

### LINKED ISSUE #20827: Lock Android compatibility when `gecko_android` is present in manifest, even if empty
Follow-up for https://github.com/mozilla/addons-server/issues/20531

When `browser_specific_settings` -> `gecko_android` is present in the manifest, even if empty, compatibility for Firefox for Android should be set and locked for that add-on.

At the moment this only works if the `gecko_android` object contains anything, but it should also work when it's `{}`.

diox:
… even if empty

Fixes #20827


src/olympia/files/utils.py
```diff
@@ -198,14 +204,16 @@ def gecko(self):
...
+            'browser_specific_settings', self.get('applications', EMPTY_FALLBACK_DICT)
         )
-        return parent_block.get('gecko', {})
+        return parent_block.get('gecko', EMPTY_FALLBACK_DICT)
 
     @property
     def gecko_android(self):
         """Return `browser_specific_settings.gecko_android` if present."""
-        return self.get('browser_specific_settings', {}).get('gecko_android', {})
+        return self.get('browser_specific_settings', EMPTY_FALLBACK_DICT).get(
```
willdurand, line 214: The fallback of `gecko_android` should be `gecko` first.

diox: There is specific code for that when it comes to min/max version handling. We don't want to return `gecko` here because we need to distinguish from the case where `gecko_android` is present.

PR #20829, opened 2023-06-14
per version blocks

### LINKED ISSUE #20748: Allow a non-continuous range of versions to be added to the blocklist

eviljeff:
fixes #20748 

Known deficiencies:
- You can't edit directly from the Block admin page currently, but it was just a shortcut to creating a BlocklistSubmission anyway.  I aim to restore that functionality in a follow-up [#20851]
- admin UI could do with some polish, especially for viewing completed BlocklistSubmission instances
- there aren't enough tests of the new functionality - i.e. being able to create multiple submissions for a single guid at the same time, with differing delays, etc.  I changed the existing tests so I'm pretty confident it works but there should be more so I'll add some more in a follow-up.
- frontend needs updating with the new serializer output, and the dropped fields removed from v5 (behind an API gate) [mozilla/addons-frontend#12252]
- the reviewer page UI when it comes to Blocks and BlockSubmissions is a little inconsistent - but all of that will need changing anyway (again, in a follow-up).


docs/topics/api/blocklist.rst
```diff
@@ -26,7 +26,6 @@ This endpoint returns an add-on Block from the blocklist, specified by guid or i
     :>json string modified: The date the block was last updated.
     :>json object|null addon_name: The add-on name, if we have details of an add-on matching that guid (See :ref:`translated fields <api-overview-translations>`).
     :>json string guid: The guid of the add-on being blocked.
-    :>json string min_version: The minimum version of the add-on that will be blocked.  "0" is the lowest version, meaning all versions up to max_version will be blocked.  ("0" - "*" would be all versions).
-    :>json string max_version: The maximum version of the add-on that will be blocked.  "*" is the highest version, meaning all versions from min_version will be blocked.  ("0" - "*" would be all versions).
+    :>json string versions[]: The versions of this add-on that are blocked.
```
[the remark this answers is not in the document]
eviljeff: yeah, I just edited the docs - the fields are still present in v5 until I (or someone else) updates frontend and removes them from the API in a follow-up.  Once there are blocks with gaps in the ranges of versions the fields are inaccurate though so I removed them.  I can leave them in the docs for now if you prefer?

eviljeff:
Another set of undefined/sub-optimal behaviour is around block metadata:
- when a guid is submitted in a **multiple** guid blocklistsubmission, but none of that guids versions are selected to be blocked (i.e. all the checkboxes are deselected; or the versions are already present in another blocklistsubmission so not available) currently the reason + url would be updated on that guid's block.
- On the other hand, when a guid is submitted in a **single** guid blocksubmission and none of the versions are selected to be blocked (or all are unavailable), because you _want_ to update the metadata a validation error is raised because at least one version must be selected.
- when there are multiple blocklistsubmissions affecting a guid each submission that is published last will overwrite the previous metadata - e.g. if you _don't_ want to change the reason/url you have to be careful to copy to have the reason/url be the same in each submission, and copy from the existing block

I'm not sure what the solution is to metadata really - maybe a property in the blocklistsubmission instance that indicates that metadata should be updated for that block?  Definitely needs addressing in follow-up.

docs/topics/api/blocklist.rst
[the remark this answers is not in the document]
diox: I had not read the code yet so I was a bit surprised. I guess maybe leave them while it's there, with a warning that we have to migrate? We'll probably do it sooner than later but it might be better to be consistent

src/olympia/blocklist/migrations/0031_auto_20230612_1447.py
```diff
@@ -0,0 +1,28 @@
...
+
+from django.db import migrations
+
+
+def convert_min_max_to_block_version(apps, schema_editor):
+    Block = apps.get_model('blocklist', 'Block')
+    BlockVersion = apps.get_model('blocklist', 'BlockVersion')
+    File = apps.get_model('files', 'File')
+
+    for block in Block.objects.all():
```
diox, line 11: That call will likely fail without `.iterator()` since Django would try to cache all the rows for that queryset. And even with it... That's a lot of rows to process in a migration.

Starting to think we should split this PR into 2:
1. Introduce `BlockVersion`, maintain it as blocks are created/edited/deleted, but do not sure it as the source of truth yet, keep relying on min/max range. Migrate existing blocks using a command that we'd run after the push.
2. Replace min/max range with `BlockVersion` entirely next push (the bulk of what this PR currently contains)

What do you think ?

src/olympia/blocklist/migrations/0031_auto_20230612_1447.py
```diff
@@ -0,0 +1,28 @@
...
+        block_versions = [
+            BlockVersion(version_id=version_id, block=block)
+            for version_str, version_id in File.objects.filter(
```
diox, line 14: As currently written, if the migration fails mid-way we're in trouble, because it can't pick up where it left off - the `bulk_create()` would fail on already created `BlockVersion`. Could you avoid them in this queryset ?

eviljeff: if we assume that `bulk_create` doesn't fail in a way that some of the `BlockVersion`s are created and not others, then yeah we could filter out any `Block` that already has at least one `BlockVersion`.

eviljeff: Or, as I now think you mean, filter them out in the `File` query.

src/olympia/blocklist/migrations/0031_auto_20230612_1447.py
eviljeff: > That call will likely fail without .iterator() since Django would try to cache all the rows for that queryset. 

could also make it a `values_list` and only return `id`, `guid`, `min_version`, and `max_version`.

> Starting to think we should split this PR into 2:
>
> 1.     Introduce BlockVersion, maintain it as blocks are created/edited/deleted, but do not sure it as the source of truth yet, keep relying on min/max range. Migrate existing blocks using a command that we'd run after the push.
> 2.     Replace min/max range with BlockVersion entirely next push (the bulk of what this PR currently contains)

It's possible (at some currently unquantifiable amount of effort) - I doubt the second pr will be significantly smaller than this pr is now though.

src/olympia/blocklist/migrations/0031_auto_20230612_1447.py
diox: Yeah sorry, I meant filter them out of the `File` query.

src/olympia/blocklist/migrations/0031_auto_20230612_1447.py
diox: > > That call will likely fail without .iterator() since Django would try to cache all the rows for that queryset.
> 
> could also make it a `values_list` and only return `id`, `guid`, `min_version`, and `max_version`.

`.iterator()` also internally does chunking to avoid fetching too much from the database driver layer at the same time, so we likely need it. We could `values_list()` on top of that yeah, but I'm not sure it's that important with `.iterator()`.

diox:
> * On the other hand, when a guid is submitted in a **single** guid blocksubmission and none of the versions are selected to be blocked (or all are unavailable), because you _want_ to update the metadata a validation error is raised because at least one version must be selected.

Yeah I encountered that and it was confusing:
- I had an existing `Block` with 2 versions blocked
- I tried to create a new `BlocklistSubmission` for that same `guid`. It showed me the 2 versions, unchecked. Neither of them are valid choices (and the message about that is confusing - it doesn't tell you why they are not valid choices) so I couldn't submit.

eviljeff:
> > * On the other hand, when a guid is submitted in a **single** guid blocksubmission and none of the versions are selected to be blocked (or all are unavailable), because you _want_ to update the metadata a validation error is raised because at least one version must be selected.
> 
> Yeah I encountered that and it was confusing:
> 
>     * I had an existing `Block` with 2 versions blocked
> 
>     * I tried to create a new `BlocklistSubmission` for that same `guid`. It showed me the 2 versions, unchecked. Neither of them are valid choices (and the message about that is confusing - it doesn't tell you why they are not valid choices) so I couldn't submit.

It's supposed to show you the 2 versions disabled, because they're already blocked, but I guess I introduced a bug along the way.  For a create blocksubmission anyway - a delete blocksubmission is supposed to show existing blocked versions not disabled, so you can unblock them.

src/olympia/blocklist/models.py
```diff
@@ -298,26 +284,34 @@ def get_changes_from_block(self, block):
...
 
-    def clean(self):
-        if self.min_version > self.max_version:
-            raise ValidationError(_('Min version can not be greater than Max version'))
-
     def get_blocks_submitted(self, load_full_objects_threshold=1_000_000_000):
-        blocks = self.block_set.all().order_by('id')
-        if blocks.count() > load_full_objects_threshold:
+        blocks_qs = self.block_set.all().order_by('id')
+        load_fakes = blocks_qs.count() > load_full_objects_threshold
```
diox, line 289: Follow-up: I wonder if we still need that logic?

src/olympia/blocklist/utils.py
```diff
@@ -153,24 +151,77 @@ def save_guids_to_blocks(guids, submission, *, fields_to_set):
...
+                    block.addon.deny_resubmission()
+                except GuidAlreadyDeniedError:
+                    pass
+            else:
+                block.addon.update(status=amo.STATUS_DISABLED)
+
+    return blocks
+
+
+def delete_versions_from_blocks(guids, submission, *, fields_to_set):
```
diox, line 190: Is this logic covered by any tests ? or is it part of the things you were considering in a follow-up ?

diox (commented):
Overall this looks good, so well done on the titanic work :) 

I said on slack I was on the fence regarding landing this in 2 stages but now that I'm through reviewing this I think we have to. We will be a lot safer if the first batch we land is completely backwards-compatible, especially if we can do the backfill outside of the migration step of the push.

> there aren't enough tests of the new functionality - i.e. being able to create multiple submissions for a single guid at the same time, with differing delays, etc. I changed the existing tests so I'm pretty confident it works but there should be more so I'll add some more in a follow-up.

I think we'll want those extra tests before merging this, to avoid the potential for edge cases triggered by (accidental or intended) use of the new functionality.

src/olympia/blocklist/admin.py
```diff
@@ -419,40 +332,27 @@ def add_view(self, request, **kwargs):
...
-                    'existing_min_version': initial.get('min_version', Block.MIN),
-                    'existing_max_version': initial.get('max_version', Block.MAX),
-                }
-            )
+            initial.update(**{'input_guids': guids_data})
             if 'action' in request.POST:
                 initial['action'] = request.POST['action']
             form = MultiBlockForm(initial=initial)
         elif request.method == 'POST':
+            form_data = request.POST.copy()
```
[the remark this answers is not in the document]
eviljeff: Not in this iteration, no.  (Originally I was trying to get the functionality where you edit and save a Block and it creates a  BlockSubmission working again, but I stripped that out because it was too much extra and a nice to have)

src/olympia/blocklist/admin.py
```diff
@@ -655,6 +532,11 @@ def addon_updated(self, obj):
     def users(self, obj):
         return obj.average_daily_users_snapshot
 
+    def blocked_versions(self, obj):
+        return ', '.join(
+            sorted(obj.blockversion_set.values_list('version__version', flat=True))
```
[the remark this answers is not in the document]
eviljeff: The follow-up should be restoring the edit functionality on the page, so they'll be checkboxes, but (as I just wrote on the other comment) I had to pull that functionality for now.

src/olympia/blocklist/models.py
eviljeff: We could maybe switch it out for full models, but use `.only()` to load a subset of the data.  Or always load full Block instances and always have a list of `FakeBlockAddonVersion` for addon versions?  Having to handle both cases _does_ add some extra complexity.

src/olympia/blocklist/utils.py
eviljeff: TestBlockAdminDelete tests this, by deleting one block completely (all blocked versions), and just removing a single blocked version from the other.

src/olympia/blocklist/migrations/0031_auto_20230612_1447.py
eviljeff: resolved in #20837

eviljeff:
merged; addressed tests; added some code and tests! Good news is with the `BlockVersion` changes landing previously there isn't a hairy migration, and some of the already obsolete code isn't cluttering up _this_ PR; Bad news is the extra tests and other changes have _increased_ the changed line cont.

src/olympia/blocklist/models.py
eviljeff: (Note, I didn't address this.  I still could in this PR rather than in a follow-up.)

src/olympia/blocklist/admin.py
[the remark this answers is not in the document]
eviljeff: https://github.com/mozilla/addons-server/issues/20851

src/olympia/blocklist/models.py
diox: I was wondering if that was often reached - but I guess it is because `GUID_FULL_LOAD_LIMIT` is not too high.

diox: Let's keep it like that.

diox (approved):

PR #20830, opened 2023-06-15
add domain name support to deniedratingword + model validation

### LINKED ISSUE #20710: Denied Rating Words should support domains

Adding a denied rating word like `example.com` should match strings like `Get my great app at app.example.com!`

### What happened?

According to @eviljeff, patterns containing a `.` currently don't match at all.

See above.

eviljeff:
fixes #20710

### DIFF
--- a/src/olympia/ratings/models.py
+++ b/src/olympia/ratings/models.py
@@ -366,7 +366,7 @@ ALPHANUMERIC_REGEX = r'[^\w]+|_+'
 def word_validator(value):
     if DOT not in value and list(re.split(ALPHANUMERIC_REGEX, value)) != [value]:
         raise ValidationError(
-            _('%(value)s contains a non-alphanumeric charactor.'),
+            _('%(value)s contains a non-alphanumeric character.'),
             params={'value': value},
         )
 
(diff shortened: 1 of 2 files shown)


src/olympia/ratings/models.py
```diff
@@ -358,10 +359,29 @@ class RatingAggregate(ModelBase):
...
 
 
+DOT = '.'
+ALPHANUMERIC_REGEX = r'[^\w]+|_+'
+
+
+def word_validator(value):
+    if DOT not in value and list(re.split(ALPHANUMERIC_REGEX, value)) != [value]:
+        raise ValidationError(
+            _('%(value)s contains a non-alphanumeric charactor.'),
```
diox, line 369:
```suggestion
            _('%(value)s contains a non-alphanumeric character.'),
```

src/olympia/ratings/models.py
```diff
@@ -358,10 +359,29 @@ class RatingAggregate(ModelBase):
...
 class DeniedRatingWord(ModelBase):
     """Denied words in a rating body."""
 
-    word = models.CharField(max_length=255, unique=True)
+    word = models.CharField(
+        max_length=255,
+        unique=True,
+        help_text='Can only contain alphanumeric charactors ("\\w", exc. "_"). '
+        'If contains a "." it will be interpretted as a doman name instead, '
+        'and can contain any charactor',
```
diox, line 382:
```suggestion
        help_text='Can only contain alphanumeric characters ("\\w", exc. "_"). '
        'If contains a "." it will be interpreted as a domain name instead, '
        'and can contain any character',
```

diox (approved):

PR #20845, opened 2023-06-19
Record set/clear needs human review and clear pending rejection by version

### LINKED ISSUE #20813: Record some reviewer actions by version rather than by addon
https://docs.google.com/document/d/1ZSHEX0h74lcZvnJbmOcziEL8XNLGtJGTHtK8Jgl_QpA/edit

Certain reviewer actions are available as a single action affecting all versions at once, where in fact the underlying properties are per-version. These actions should be changed to be per-version.

Applicable actions
- Clear pending rejections
- Set "Needs Human Review”
- Clear Needs Human Review
- Clear Admin Static Theme Review Flag (split into different issue, https://github.com/mozilla/addons-server/issues/20844)

Each of these actions should be per-version (re-using the versions dropdown UI already used by other actions). In addition, the Set "Needs Human Review” needs to record a comment with the reason. The action and text if any, should be then shown in the version log on the review page with other activity.

diox:
Rather than applying them through the API per add-on for all versions in the channel, treat them as an action applied to the selected versions.

Fixes #20813

### DIFF
--- a/src/olympia/constants/activity.py
+++ b/src/olympia/constants/activity.py
@@ -869,6 +869,7 @@ class LOG_IN_API_TOKEN(_LOG):
     format = '{user_responsible} authenticated through an API token.'
 
 
+# Obsolete now that this is done per version.
 class CLEAR_NEEDS_HUMAN_REVIEWS(_LOG):
     id = 173
     format = '{addon} no longer flagged for human review.'
@@ -878,7 +879,7 @@ class CLEAR_NEEDS_HUMAN_REVIEWS(_LOG):
     reviewer_review_action = True
 
 
-class NEEDS_HUMAN_REVIEW(_LOG):
+class NEEDS_HUMAN_REVIEW_AUTOMATIC(_LOG):
     id = 174
     format = '{version} flagged for human review.'
     short = 'Flagged for human review'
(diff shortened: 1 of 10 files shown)


src/olympia/constants/activity.py
```diff
@@ -894,6 +895,36 @@ class REPLY_RATING(_LOG):
...
+    admin_event = True
+    review_queue = True
+    reviewer_review_action = True
+    hide_developer = True
+
+
+class CLEAR_PENDING_REJECTION(_LOG):
+    id = 177
+    format = _('{version} pending rejection cleared.')
+    short = _('Pending rejection cleared')
```
eviljeff, line 911: Are these two localized for a reason?  (when the other actions you've added aren't)

eviljeff (approved):

src/olympia/constants/activity.py
wagnerand: If that action appears in devhub, it would be good to have it localized. If not, we don't need localization.

eviljeff: @wagnerand should it show in devhub, do you think?

wagnerand: Are we showing a pending (delayed) rejection in the devhub? If so, where?
If we were to show the clearing, where would we show it?

eviljeff: Yes, we expose the activity log for `REJECT_VERSION_DELAYED` ("{addon} {version} reject scheduled.") in the activity section in devhub.  This action would be exposed in the same place.

p.s. having looked at the action wording, maybe it should be `'{version} scheduled rejection cleared.` to match? `</bikeshed>`

wagnerand: If we expose `REJECT_VERSION_DELAYED` and we persist that even beyond `CLEAR_PENDING_REJECTION` (which @eviljeff confirmed in chat), then we should also expose `CLEAR_PENDING_REJECTION` to show the full trail of that activity to the developer. And as such, localizing the activity would be good.

PR #20887, opened 2023-06-26
change version checkboxes to make it clearer between blocking and already blocked

### LINKED ISSUE #20882: correctly expose which versions are in blocklistsubmission
follow-up for #20748 

the versions that are changed in a blocklist submission should be clear in a blocklistsubmission both before and and after it's been published

eviljeff:
fixes #20882 - the UI isn't great still, imo, but it's better than now, where we mix up a version that's _already_ blocked, with a version that the submission will be blocking. 

Two things I'd like to improve on at some point:
a) a way to better indicate what the checkbox will do than write "Block" or "Unblock" next to it, like a custom toggle button with block and unblock on it (but started falling down the CSS rabbit hole)
b) the red hexagon and green circle visually are just coloured blobs at the normal font size and don't incontrovertibly indicate "blocked version" and "not blocked version"

### DIFF
--- a/src/olympia/blocklist/templates/admin/blocklist/includes/enhanced_blocks.html
+++ b/src/olympia/blocklist/templates/admin/blocklist/includes/enhanced_blocks.html
@@ -25,20 +25,21 @@
       <ul>
         {% for version in block_obj.addon_versions %}
           <li data-version-id="{{ version.id }}">
+          {% if version.id in form.changed_version_ids_choices or version.id in instance.changed_version_ids %}
             <label><input
               type="checkbox"
               name="changed_version_ids"
               value="{{ version.id }}"
-              {% if version.id in form.changed_version_ids_choices %}
-                {% if version.id in form.changed_version_ids.value %}checked{% endif %}
-              {% else %}
-                disabled
-                {% if version.is_blocked %}checked{% endif %}
+              {% if version.id in form.changed_version_ids.value %}checked{% endif %}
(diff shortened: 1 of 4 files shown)


src/olympia/blocklist/templates/admin/blocklist/includes/enhanced_blocks.html
```diff
@@ -25,20 +25,21 @@ <h3>{{ form.blocks|length|intcomma }} Add-on GUIDs with {{ form.total_adu|intcom
...
-                {% if version.id in form.changed_version_ids.value %}checked{% endif %}
-              {% else %}
-                disabled
-                {% if version.is_blocked %}checked{% endif %}
+              {% if version.id in form.changed_version_ids.value %}checked{% endif %}
+              {% if version.id in instance.changed_version_ids %}checked disabled{% endif %}
+            > {% if is_delete %}Unblock{% else %}Block{% endif %} {{ version.version }}</label>
+          {% else %}
+            <!-- Red Hexagonal stop sign for Blocked; Green cirle for not blocked -->
+            {% if version.is_blocked %}&#x1F6D1;{% else %}&#x1F7E2;{% endif %}{{ version.version }}
```
diox, line 38: Maybe wrap that in a `<span>` to add a title explaining what the icon means ? Or add that `title` to the whole `<label>` ?

diox (approved):

Issue #20859, opened 2023-06-20
The "cron gc" command fails to run in the GCP dev environment


bqbn:
### Describe the problem and steps to reproduce it:
The "cron gc" command fails to run in the GCP dev environment with the following error,

```
$ python -W ignore::DeprecationWarning manage.py cron gc
{"Timestamp": 1687293338029215232, "Type": "z.cron", "Logger": "http_app_addons_dev", "Hostname": "addons-server-v1-deploy-uwsgi-amo-web-d7b8fd9d7-ktcb5", "EnvVersion": "2.0", "Severity": 6, "Pid": 25, "Fields": {"uid": "<anon>", "remoteAddressChain": "", "msg": "Beginning job: gc [] {} (start timestamp: 1687293338029.1392)"}, "severity": 200}
{"Timestamp": 1687293338030695936, "Type": "z.cron", "Logger": "http_app_addons_dev", "Hostname": "addons-server-v1-deploy-uwsgi-amo-web-d7b8fd9d7-ktcb5", "EnvVersion": "2.0", "Severity": 6, "Pid": 25, "Fields": {"uid": "<anon>", "remoteAddressChain": "", "msg": "Collecting data to delete"}, "severity": 200}
{"Timestamp": 1687293338338046720, "Type": "z.cron", "Logger": "http_app_addons_dev", "Hostname": "addons-server-v1-deploy-uwsgi-amo-web-d7b8fd9d7-ktcb5", "EnvVersion": "2.0", "Severity": 6, "Pid": 25, "Fields": {"uid": "<anon>", "remoteAddressChain": "", "msg": "[FileUpload:2039366e-ec37-4e05-8a37-a9dc7161bf83] Removing file: /mnt/efs/addons-dev.allizom.org/files/temp/096c5ae3b0234efb82ceb18edf244ba1.zip"}, "severity": 200}
Traceback (most recent call last):
  File "/data/olympia/manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/deps/lib/python3.10/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/deps/lib/python3.10/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/deps/lib/python3.10/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
[...]

PR #20897, opened 2023-06-28
Store FileUpload paths as relative and not absolute

diox:
Helps #20859

We can't migrate existing paths without breaking compatibility during the push window, so this doesn't migrate existing fileuploads, and instead handles both absolute and relative paths with a compatibility shim (`file_path`).

A database update will be needed specifically for dev (since it has old absolute paths invalid on GCP) but after 2 weeks (period after which the `gc` cron kicks in) everything should be clean again and in the future.

### DIFF
--- a/src/olympia/amo/cron.py
+++ b/src/olympia/amo/cron.py
@@ -55,7 +55,7 @@ def gc(test_result=True):
     for file_upload in stale_uploads:
         log.info(
             '[FileUpload:{uuid}] Removing file: {path}'.format(
-                uuid=file_upload.uuid, path=file_upload.path
+                uuid=file_upload.uuid, path=file_upload.file_path
             )
         )
         if file_upload.file_path:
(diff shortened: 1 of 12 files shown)


eviljeff:
> A database update will be needed specifically for dev (since it has old absolute paths invalid on GCP) but after 2 weeks (period after which the `gc` cron kicks in) everything should be clean again and in the future

So, what about the database migration for stage and then prod when GCP moves onto those environments?

eviljeff (approved):

diox:
When stage and prod move to GCP, they're already have that patch. Assuming stage/prod happens at least 2 weeks or more after this patch lands in those environnements, they'll already no longer have `FileUpload` with absolute paths.

bqbn:
To confirm, we ran the following query in -stage and -prod environments today and both returned 0 count.

```
mysql> select count(*) from file_uploads where path like '/mnt%';
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.01 sec)
```

PR #20932, opened 2023-07-04
store more sensible logs for [un]blocked versions

### LINKED ISSUE #20899: The message generated when unblocking a version is incorrect
Followup for https://github.com/mozilla/addons-server/issues/20748

1. This add-on has all versions blocked https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/625728
2. From Edit block I clicked on Delete and selected only one version to be un-blocked

The result:
I expected to see in the Block History a message about one version being unblocked -> there's a [block edited](https://addons-internal.dev.mozaws.net/en-US/admin/models/blocklist/blocklistsubmission/add/?guids=%7B145dd373-de00-4e47-a0e9-f8e32c17ada0%7D&action=1) on all versions

The review page displays on all the versions a Block edited: 5 versions unblocked. 

Expected result: 
1 version unblocked (version 1.0)

eviljeff:
fixes #20899

### DIFF
--- a/src/olympia/blocklist/utils.py
+++ b/src/olympia/blocklist/utils.py
@@ -4,22 +4,10 @@ import olympia.core.logger
 from olympia import amo
 from olympia.activity import log_create
 from olympia.users.utils import get_task_user
+from olympia.versions.models import Version
 
 
 log = olympia.core.logger.getLogger('z.amo.blocklist')
-
-
-def add_version_log_for_blocked_versions(obj, al, submission_obj=None):
-    from olympia.activity.models import VersionLog
-
-    VersionLog.objects.bulk_create(
-        [
-            VersionLog(activity_log=al, version=version)
-            for version in obj.addon_versions
-            if version.is_blocked
-            or (submission_obj and version.id in submission_obj.changed_version_ids)
-        ]
-    )
 
 
 def block_activity_log_save(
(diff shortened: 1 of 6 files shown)


src/olympia/blocklist/utils.py
```diff
@@ -4,68 +4,92 @@
...
         details['signoff_state'] = submission_obj.SIGNOFF_STATES.get(
             submission_obj.signoff_state
         )
         if submission_obj.signoff_by:
             details['signoff_by'] = submission_obj.signoff_by.id
-    addon = obj.addon
-    al = log_create(action, addon, obj.guid, obj, details=details, user=obj.updated_by)
+
+    log_create(action, obj.addon, obj.guid, obj, details=details, user=obj.updated_by)
+    for version_id in changed_version_ids:
```
diox, line 47: I wonder if it's worth trying to create a single `ActivityLog` with multiple versions arguments like we do in reviewer tools for actions dealing with multiple versions. That still results in multiple `VersionLog` (which may be an issue when it comes to performance - the `bulk_create()` that was previously there was probably helping, by making a single query for all).

Maybe a follow-up.

diox (approved):
r+test fixes

src/olympia/blocklist/utils.py
eviljeff: I looked at the `reject_multiple_versions` action - it seemed the most analogous - and it just loops through and called `log_create` individually.

diox: It loops over `user_and_versions` (and `action_to_record`) because it processes delayed rejections into actual ones, and there might be a mix of content/code rejections made by several reviewers. But in the general case, it only records a single `ActivityLog` (it passes `versions` to `self.log_action()` in the review helper). Inside `ActivityLog.create()` it then creates as many `VersionLog` as necessary (*), but only a single `ActivityLog` for multiple versions.

(*) We could `bulk_create()` them too ...

eviljeff: https://github.com/mozilla/addons-server/issues/20948

PR #20956, opened 2023-07-07
optimize mlbf fetch_blocked_from_db query

### LINKED ISSUE #20955: optimize MLBF generation so it can actually finish
`fetch_blocked_from_db `executes a query that returns `BlockVersion` instances, but then accesses fks to `Block` and `Version` for each instance - effectively executing an extra 2 queries for every blocked guid.  On dev and stage there were sufficiently few blocks that we didn't notice; on prod not so much.  The generation cron usually takes 2-3 minutes (according to statsd pings recorded in grafana); last night it took over an hour.  Considering we run this cron 4 times a day this is way too slow and puts an unacceptable strain on the database - if it doesn't entirely timeout.

eviljeff:
fixes #20955

### DIFF
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -49,12 +49,15 @@ def generate_mlbf(stats, blocked, not_blocked):
 def fetch_blocked_from_db():
     from olympia.blocklist.models import BlockVersion
 
+    qs = BlockVersion.objects.filter(version__file__is_signed=True).values_list(
+        'block__guid', 'version__version', 'version_id', named=True
+    )
     all_versions = {
         block_version.version_id: (
-            block_version.block.guid,
-            block_version.version.version,
+            block_version.block__guid,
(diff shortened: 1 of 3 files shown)


src/olympia/blocklist/mlbf.py
```diff
@@ -49,12 +49,15 @@ def generate_mlbf(stats, blocked, not_blocked):
 def fetch_blocked_from_db():
     from olympia.blocklist.models import BlockVersion
 
+    qs = BlockVersion.objects.filter(version__file__is_signed=True).values_list(
+        'block__guid', 'version__version', 'version_id', named=True
```
diox, line 53: I keep forgetting that `named=True` exists for `values_list()`, that's handy

diox (approved):
As I said on Matrix, not loving the high `assertNumQueries()`, didn't expect them to be that high, but I've double-checked that it's only for `Config` stuff like you said, so it should be ok.

PR #20960, opened 2023-07-07
Use task user for inherited needs human review log activity

### LINKED ISSUE #20953: Attribute "Previous version ... had needs human review set" activity to task user

In the reviewer tools, version review history...

### What happened?

...the activity "Previous version in channel had needs human review set" is attributed to the developer account.

It should be attributed to the Mozilla Firefox task user.

wagnerand:
Fix #20953.

### DIFF
--- a/src/olympia/versions/tests/test_models.py
+++ b/src/olympia/versions/tests/test_models.py
@@ -2076,6 +2076,13 @@ class TestExtensionVersionFromUpload(TestVersionFromUpload):
             == NeedsHumanReview.REASON_INHERITANCE
         )
 
+        activity_log = (
+            ActivityLog.objects.for_versions(upload_version)
+            .filter(action=amo.LOG.NEEDS_HUMAN_REVIEW_AUTOMATIC.id)
+            .first()
+        )
+        assert activity_log.user == get_task_user()
+
     def test_dont_inherit_due_date_far_in_future(self):
         standard_due_date = get_review_due_date()
         due_date = datetime.now() + timedelta(days=15)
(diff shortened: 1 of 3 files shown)


src/olympia/versions/tests/test_models.py
```diff
@@ -2076,6 +2076,13 @@ def test_inherit_needs_human_review_with_due_date(self):
             == NeedsHumanReview.REASON_INHERITANCE
         )
 
+        activity_log = (
+            ActivityLog.objects.for_versions(upload_version)
+            .filter(action=amo.LOG.NEEDS_HUMAN_REVIEW_AUTOMATIC.id)
+            .first()
+        )
+        assert activity_log.user == get_task_user()
```
eviljeff, line 2084: mmm, but we fall back to the task user anyway if `core.get_user()` is falsey - which it often is in tests.  So can you either assert `core.get_user()` is non-falsey, or set it if needed?

eviljeff (changes requested):

eviljeff:
btw, mat fixed the intermittent that's causing the main job to fail, and you're getting the `release-master` job (that fails) because the commit isn't on a branch

eviljeff (approved):

wagnerand:
> btw, mat fixed the intermittent that's causing the main job to fail, and you're getting the release-master job (that fails) because the commit isn't on a branch

Yea that was a fail on my side. I did create a feature branch locally but didn't switch to it. I only realized after opening the PR, and I don't think you can change the source branch in a PR so I would have had to close this one and open a new one.

Issue #20966, opened 2023-07-07
Prevent uploading extensions using the theme-specific submission flow


diox:
Follow-up for https://github.com/mozilla/addons-server/issues/20942

If you upload a non-theme on the theme page (or a theme to the extension page? less problematic, maybe we don't care), display an error linking to the right page instead.

diox:
> or a theme to the extension page? less problematic, maybe we don't care

I left that working. It could be useful as a backwards-compatibility measure for people who bookmarked old URLs, and it doesn't cause the issues submitting an extension using the theme page would.

ioanarusiczki:
@diox 

1. The error message is displayed when extensions, dicts or lang packs are submitted from 
https://addons-dev.allizom.org/en-US/developers/addon/submit/theme/upload-listed

![not a theme](https://github.com/mozilla/addons-server/assets/33448286/c996de51-0ca2-4b7c-afca-0000f4c0eef1)

"Submit a new add-on" redirects to https://addons-dev.allizom.org/en-US/developers/addon/submit/upload-listed or upload-unlisted, depending on what the developer has chosen. 


2. Themes can be submitted from extension's (old) url

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-449

Issue #20942, opened 2023-07-05
Split themes and extensions submission initial page


diox:
At the moment the themes and extensions (+ dicts/langpacks) use the same initial page, where you can choose to either upload a xpi or use the theme wizard. This is super flexible to users, albeit slightly confusing perhaps.

We want to make changes to the extension submission flow without affecting theme developers, so we'll need to split the themes from the rest. The 2 cases (themes / rest) will have different URLs, slightly different "landing" page, which will look similar but will allow us to customize separately.

- On the extension page, the theme wizard should not be displayed, only the xpi upload should be present.
- On the theme page, the theme wizard and xpi upload should be present
- For backwards-compatibility, the old URLs should still work and be used for the extension (+ dict/langpack) flow.
- See also https://github.com/mozilla/addons-server/issues/20966

ioanarusiczki:
@diox 

On AMO Dev: 

- the themes wizard has been removed when submission is made from "Submit a New Add-on" 
- theme submission has a new url https://addons-dev.allizom.org/en-US/developers/addon/submit/theme/upload-listed or 
https://addons-dev.allizom.org/en-US/developers/addon/submit/theme/upload-unlisted and have both options available: upload using an xpi and the theme wizard.
- the old urls are still being used for extension, lang pack and dictionary submissions 
https://addons-dev.allizom.org/en-US/developers/addon/submit/upload-listed or
https://addons-dev.allizom.org/en-US/developers/addon/submit/upload-unlisted

- I've checked the submission buttons on frontend 
and dev hub:
https://addons-dev.allizom.org/en-US/developers/
https://addons-dev.allizom.org/en-US/developers/addons ( on the right side and in Tools)

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-440

PR #20968, opened 2023-07-07
Split New Theme submission from the rest

diox:
Fixes https://github.com/mozilla/addons-server/issues/20942
Fixes https://github.com/mozilla/addons-server/issues/20966

### DIFF
--- a/src/olympia/devhub/templates/devhub/addons/submit/upload.html
+++ b/src/olympia/devhub/templates/devhub/addons/submit/upload.html
@@ -48,10 +48,7 @@
     {% endif %}
   {% endif %}
   >
-  {% if theme_specific %}
-    <input type="hidden" name="theme_specific" id="upload-addon-theme-specific" value="1" />
-  {% endif %}
-
+  {{ new_addon_form.theme_specific }}
   {{ new_addon_form.non_field_errors() }}
 
     <div class="compatible-apps">
--- a/src/olympia/devhub/templates/devhub/addons/submit/upload.html
+++ b/src/olympia/devhub/templates/devhub/addons/submit/upload.html
@@ -48,6 +48,9 @@
     {% endif %}
   {% endif %}
   >
+  {% if theme_specific %}
+    <input type="hidden" name="theme_specific" id="upload-addon-theme-specific" value="1" />
+  {% endif %}
 
   {{ new_addon_form.non_field_errors() }}
 
--- a/src/olympia/devhub/templates/devhub/addons/submit/upload.html
+++ b/src/olympia/devhub/templates/devhub/addons/submit/upload.html
@@ -69,11 +69,6 @@
     </div>
     {% endif %}
 
-    {% if addon and addon.type == amo.ADDON_STATICTHEME %}
-        {% set wizard_url = url('devhub.submit.version.wizard', addon.slug, channel_param) %}
-    {% elif not addon %}
-        {% set wizard_url = url('devhub.submit.wizard', channel_param) %}
-    {% endif %}
     {% if wizard_url %}
     <div class="submission-buttons addon-create-theme-section">
       <p></p>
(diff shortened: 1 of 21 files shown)


diox:
Need to hunt down every test that submits a theme to make sure it uses the new flow.

src/olympia/devhub/templates/devhub/addons/submit/upload.html
```diff
@@ -48,6 +48,9 @@ <h3>{{ _('Upload Version') }}</h3>
     {% endif %}
   {% endif %}
   >
+  {% if theme_specific %}
+    <input type="hidden" name="theme_specific" id="upload-addon-theme-specific" value="1" />
```
eviljeff, line 52: magic number as a string 😬  - is this defined anywhere as a constant ?

src/olympia/devhub/urls.py
```diff
@@ -241,7 +257,7 @@
         # Old LWT Theme submission.
         re_path(
             r'^theme/submit/?$',
-            lambda r: redirect('devhub.submit.agreement'),
+            lambda r: redirect('devhub.submit.theme.agreement'),
             name='devhub.themes.submit',
```
eviljeff, line 261: Not entirely the fault of this patch, but we now have `devhub.themes.submit` and `devhub.submit.theme.*` urls.  Rename the old LWT one to something obviously deprecated, but under the same naming scheme?

src/olympia/devhub/utils.py
```diff
@@ -195,6 +196,25 @@ def __init__(self, file_, addon=None, listed=None, final_task=None):
             try:
                 addon_data = parse_addon(file_, minimal=True)
                 is_mozilla_signed = addon_data.get('is_mozilla_signed_extension', False)
+                if theme_specific and addon_data['type'] != amo.ADDON_STATICTHEME:
```
eviljeff, line 199: Where is the opposite case handled (`not theme_specific and addon_type['type'] == amo.ADDON_STATICTHEME`)?  Can we have a comment if there's a reason why it's not in the same function?

src/olympia/devhub/templates/devhub/addons/submit/upload.html
diox: Any truthy value would work, we convert it to a `bool` in the `upload` view

src/olympia/devhub/utils.py
diox: The opposite case is not handled, because we don't care - if someone submits a theme through the non-theme workflow, that's fine - they'll just get the 2FA restrictions when it's added later, but it should work.

Because the non-theme workflow uses the current URLs (that have been used for a while for all add-ons) it's probably best not to restrict that flow at all.

src/olympia/devhub/urls.py
diox: 722f7d6038b545be69edb900f74f9837bce24bbe

src/olympia/devhub/templates/devhub/addons/submit/upload.html
diox: Refactored to use a form field in 97f8197627d6175dd761e5a3ee459d2da239afd3

src/olympia/devhub/utils.py
diox: 94a0173fda47feca3c3c56cfddafa47bf4816f08

eviljeff (approved):

src/olympia/devhub/views.py
```diff
@@ -584,13 +586,25 @@ def handle_upload(
...
+        request.POST.get('theme_specific')
     )
+    try:
+        upload = handle_upload(
+            filedata=filedata,
+            request=request,
+            addon=addon,
+            is_standalone=is_standalone,
+            channel=channel,
+            theme_specific=bool(theme_specific),
```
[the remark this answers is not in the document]
diox: Totally, I should remove that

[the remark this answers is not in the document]
diox: https://github.com/mozilla/addons-server/pull/20968/commits/31134e3b5be0cbd3e268c4f74fa0cc7416ded7c2

Issue #20498, opened 2023-03-23
integrate delayed blocklist submission with reviewer tools


eviljeff:
part of https://mozilla-hub.atlassian.net/browse/AMO-43

The blocklist submission tool now supports delayed blocks, so we want to integrate this with the reviewer tools so, for example, the delay days amount can be set with the review.

eviljeff:
Unblocked and I started working on this.  Current flow is:
- `reject_latest_version` and `reject_multiple_versions` actions will do the usual things, but then also create a `BlocklistSubmission` object directly within the reviewer tools (no admin redirect) if the "Block" option is enabled on the form. 
- If the block would require sign-off (high ADU) it's held until that sign-off has been given.
- if the block is delayed (and this is where it starts to get tricky) we create a `BlocklistSubmission` with a delay, but we can't just have it proceed as a normal delayed submission as a) delayed_rejections can be cancelled and b) delayed_rejections can be held back if there is a new version submitted in the mean-time.
  - so when the delayed time limit comes up and the cron picks it up it checks if all the versions are still marked as delayed rejection; if they're not but not rejected the submission is cancelled; if they're not but rejected the blocksubmission goes ahead. (and if they are still marked as delayed rejection the submission is skipped until the next time the cron runs)

I think this covers all the cases - ideally we'd rely on signals for the reviewer flags but there are classes of undefined behavior because the auto_reject task doesn't "group" rejections like the blocklist submission does - it just rejects whatever is pending for that add-on that fits the other criteria.  It's still an option though.

@wagnerand what permission (if any beyond normal reviewer perms) will be needed to create the blocklist submission from the reviewer tools like this?

wagnerand:
> reject_latest_version and reject_multiple_versions actions will do the usual things, but then also create a BlocklistSubmission object directly within the reviewer tools (no admin redirect) if the "Block" option is enabled on the form.

Can we put that behind a waffle flag, please?

> what permission (if any beyond normal reviewer perms) will be needed to create the blocklist submission from the reviewer tools like this?

Since it will be the default for rejections at some point, I don't think we'll need a special permission here.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-296

PR #20978, opened 2023-07-10
add canned reasons to blocksubmission admin

eviljeff:
fixes #20908 
Screenshot 2023-07-10 15 35 31


eviljeff:
looking at it, a big chunk of the non-migration part of the patch is refactoring - in admin.py how the available fields are built (because there were too many different subtle ways to define what fields were present); and in forms.py how the fields were initialized

diox (commented):
Why the new model ? I would have expected us to re-use the `ReviewActionReason` model, especially if down the line we intend to put blocking directly in reviewer tools.

eviljeff:
> Why the new model ? I would have expected us to re-use the `ReviewActionReason` model, especially if down the line we intend to put blocking directly in reviewer tools.

I assumed that would be what @wagnerand wanted - it's the public facing reason for the block rather than the private rejection response we send to the developer.  But @wagnerand to confirm.

wagnerand:
Yes, it's the public facing block reason, but the `ReviewActionReason` is also kind of public (the canned response is sent to developers). Also @diox has a good point about future integration of blocking into the reviewer tools.

eviljeff:
okay, https://github.com/mozilla/addons-server/pull/20978/commits/c75f1c8e156fb32ac89ce0d69f774248585a3f04 switches to use `ReviewActionReason`, and a matching block reason can be added to each along with the rejection text.  It works like the reviewer tools - multiple reasons can be selected which will add to the textarea. When #20498 is done it you'll be able to use the checkboxes to prefill both at the same time.

src/olympia/reviewers/models.py
```diff
@@ -692,6 +692,7 @@ class ReviewActionReason(ModelBase):
     )
     name = models.CharField(max_length=255)
     canned_response = models.TextField(blank=True)
+    canned_block_reason = models.TextField(blank=True)
```
diox, line 695: missing `null=True` since it's a new field ? Also in the migration

static/js/admin/blocklist_blocklistsubmission.js
```diff
@@ -17,33 +17,60 @@ document.addEventListener('DOMContentLoaded', () => {
...
+      // Attempt to remove the canned response related to the reason.
+      reasonField.value = reasonField.value.replace(cannedReason, '');
+    }
+  };
+
+  document
+    .querySelectorAll('#id_canned_reasons input')
+    .forEach((a) => a.addEventListener('click', fillCannedReason));
+  // Move the cursor to the end of the textarea
+  document.getElementById('id_reason').setSelectionRange(100000, 100000);
```
diox, line 75: eh well, that's one way to do it...

diox (approved):

static/js/admin/blocklist_blocklistsubmission.js
eviljeff: heh, I guess I could have got the length of the field value but this worked too

src/olympia/reviewers/models.py
diox: For models that can get new rows inserted during the push window that wouldn't do it (the database default isn't kept after the migration is done, so there is a potential window of time where we have insertions without a default, hence why we typically use `null=True`) but for this specific case it's fine yes.

PR #20995, opened 2023-07-13
handle edits to submission metadata

### LINKED ISSUE #20979: Metadata is not updated in delayed sign-offs when there's already another version blocked
Followup https://github.com/mozilla/addons-server/issues/20930#issuecomment-1629022874

 An update on a [delayed sign-off](https://addons-internal.dev.mozaws.net/en-US/admin/models/blocklist/blocklistsubmission/639/change/) but the add-on had already a blocked version together with `url` and `reason` edited-> I cannot update the fields, I'll keep seeing the input belonging to the blocked version. 

Another example: an update on a [pending sign-off ](https://addons-internal.dev.mozaws.net/en-US/admin/models/blocklist/blocklistsubmission/635/change/ ) In this case only the `reason` field won't update because the `url `wasn't edited for the block created on versions before this one.

eviljeff:
fixes #20979 - it wasn't that values weren't updated (as per the issue) but that each time the form loaded we prefilled them with the values from the block rather than the values in the instance.


diox:
I'm not sure I'm resolving the conflict correctly locally, could you fix it before I can review that PR ?

eviljeff:
> I'm not sure I'm resolving the conflict correctly locally, could you fix it before I can review that PR ?

fixed.

diox (approved):

PR #20996, opened 2023-07-13
Record IP addresses from the 3rd HTTP_X_FORWARDED_FOR value from the right on GCP

### LINKED ISSUE #20925: IP addresses are not recorded correctly on GCP dev env
STR:
- Submit an add-on
- Look up the activity log for that add-on in the admin - there should be one "Version X.X added to YYYY".
- Check the IP address associated with that activity

Expected results:
- It's your IP

Actual results:
- It's an IP that belongs to Cloudfront

(Also affects to every other place where we record/use IPs, like rate limiting, but those STR was the easiest way to check I could think of)

---

In order for the client IP to be recorded properly, a few things are assumed by the code. One of these is likely wrong on dev:
- Cloudfront must pass it through `X-Forwarded-For` header in the last position
- Our load balancer must append the CDN IP to that header, moving the client IP to the second to last position
- Cloudfront must pass an additional `X-Request-Via-Cdn` header with a value matching `SECRET_CDN_TOKEN` env variable - that value is then stored in a django setting of the same name.

diox:
(or second if the request is not from the CDN)

Fixes #20925

### DIFF
--- a/src/olympia/amo/middleware.py
+++ b/src/olympia/amo/middleware.py
@@ -1,4 +1,5 @@
 import contextlib
+import os
 import re
 import uuid
 from urllib.parse import quote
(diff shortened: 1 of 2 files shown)


src/olympia/amo/middleware.py
```diff
@@ -279,38 +280,70 @@ def process_exception(self, request, exception):
...
+    On top of that, on AWS, the classic ELB we're using does not make any
+    alterations to X-Forwarded-For but on GCP, GKE Ingress appends its own IP,
+    so the client IP would move from second to last to third to last in the CDN
+    case
+
+    If the request didn't come from the CDN and is a direct origin request, on
+    AWS the client IP will be in last position in HTTP_X_FORWARDED_FOR but in
+    this case nginx already sets REMOTE_ADDR so we don't need to use it. On GCP
+    though, that won't be true and we'll need to extract it from the second to
+    last position in HTTP_X_FORWARDED_FOR.
```
eviljeff, line 314: This is super confusing - the request flow section above was clear, so can you write out where what the order of the IPs will be in the same way for the 3(?) scenarios?

src/olympia/amo/middleware.py
```diff
@@ -279,38 +280,70 @@ def process_exception(self, request, exception):
...
-            try:
-                value = x_forwarded_for_header.rsplit(sep=',', maxsplit=2)[-2].strip()
-                if not value:
-                    raise IndexError
-                request.META['REMOTE_ADDR'] = value
-            except IndexError:
-                # Shouldn't happen, must be a misconfiguration, raise an error
-                # rather than potentially use/record incorrect IPs.
-                raise ImproperlyConfigured('Invalid HTTP_X_FORWARDED_FOR')
+        if x_forwarded_for_header:
```
eviljeff, line 328: idea (that may not be an improvement, but putting it out there): split the header into `ips` and then `.pop()` off items off the end for each of the `if`s rather than iterating a numeric position variable

src/olympia/amo/middleware.py
diox: Reworded in dec7f275bee19d5893628a55c8160dbe00d207b0 let me know if that works for you

src/olympia/amo/middleware.py
diox: I played around with this but it didn't really feel any better. I could do a `.split(',').pop(-position)` but that doesn't seem worth it.

src/olympia/amo/middleware.py
eviljeff: Yeah, it's clearer.  Though it looks like it's always just the first IP address?  If it's more complicated than that we should probably mention why it's not just the first IP address in the list.

diox: That's just in general because the beginning of the contents of the header can't be trusted, it's potentially user input. I'll add a note to that.

src/olympia/amo/middleware.py
eviljeff: this is what I was thinking:
```python
def process_request(self, request):
        x_forwarded_for_header = request.META.get('HTTP_X_FORWARDED_FOR')
        if x_forwarded_for_header:
            ips = x_forwarded_for_header.split(',')
            try:
                if from_cdn := self.is_request_from_cdn(request):
                    ips.pop()
                if from_gcp := self.is_request_from_gcp_environement(request):
                    ips.pop()
                # If it's from the cdn or gcp we need to fix REMOTE_ADDR by
                # looking into HTTP_X_FORWARDED_FOR, otherwise we should ignore it.
                if from_cdn or from_gcp:
                    if remote_addr := ips.pop().strip():
                        request.META['REMOTE_ADDR'] = remote_addr
                    else:
                        raise IndexError
            except IndexError:
                # Shouldn't happen, must be a misconfiguration, raise an error
                # rather than potentially use/record incorrect IPs.
                raise ImproperlyConfigured('Invalid HTTP_X_FORWARDED_FOR')
```

src/olympia/amo/middleware.py
diox: 21bd46f5a6ca4a377b07a68b1a1655e6273a3d9f

src/olympia/amo/middleware.py
diox: Not a fan - you have to keep track of how many `pop()` you did if you want to follow the code. I see the appeal of `from_gcp` and `from_cdn` though, but I'm not sure it's worth doing.

eviljeff: The thinking was it was doing the reverse of what the CDN and/or GCP did but 🤷

PR #21001, opened 2023-07-17
Enforce 2FA on extension submission flow, api keys page

### LINKED ISSUE #20943: Enforce 2FA at extension upload time
A while ago we started [enforcing 2FA at login for developers](https://github.com/mozilla/addons-server/issues/10046). This is however a bit confusing because it's only enforced on the first login after they have become an extension developer, which could be a long time after their upload. We want to start doing that directly when someone begins the extension upload flow (*) so that it's all consistent and all developers eventually have 2FA enabled from the start, and not just after their second login.

Theme developers (that don't also have/try to submit extensions) continue to be exempted from this.

Depends on https://github.com/mozilla/addons-server/issues/20942 (note: even once that issue is implemented the upload stage will likely be shared between extensions and themes, so we'll probably have to check for 2FA on the pages before but also before validation, in the `handle_upload()` code or nearby)

(*) Or requests API keys.

### LINKED ISSUE #21002: Retry authentication with FxA when we get a error in the callback
We currently don't handle FxA errors, the user is redirected back to us, we notice that we can't log them in but do nothing about it.

FxA docs suggest to try again with a new `state` (and without `prompt=none`): https://mozilla.github.io/ecosystem-platform/reference/oauth-details#handling-errors

If FxA is returning bogus errors for some reason (it has happened once or twice in the past) then the user would be stuck in a loop, but it's not worse than being thrown back to AMO with no explanation as to why this happened. Removing the `prompt=none` in this cases gives us a chance for the next attempt to work.

diox:
Fixes #20943
Fixes #21002

Note: when trying this, make sure `2fa-enforcement-for-developers-and-special-users` flag is active (and note that it's a flag, not a switch - to activate it, enable it for `everyone`). This should work both with fake and actual FxA implementations.

This could be split before review into 3 parts:
- FxA error handling
- All 2FA refactor bits (all `accounts` stuff)
- All `devhub` 2FA enforcement bits (essentially just using the decorator + condition in `upload()` + tests)

### DIFF
--- a/src/olympia/accounts/utils.py
+++ b/src/olympia/accounts/utils.py
@@ -3,6 +3,8 @@ import os
 from base64 import urlsafe_b64encode
 from urllib.parse import urlencode
 
+import olympia.core.logger
+
 from django.conf import settings
 from django.http import HttpResponseRedirect
 from django.urls import reverse
@@ -11,11 +13,13 @@ from django.utils.encoding import force_str
 from olympia.amo.utils import is_safe_url, use_fake_fxa
 
 
+log = olympia.core.logger.getLogger('accounts')
+
+
 def fxa_login_url(
     config,
     state,
     next_path=None,
-    action=None,
     enforce_2fa=False,
     request=None,
     id_token_hint=None,
@@ -30,8 +34,6 @@ def fxa_login_url(
         'state': state,
         'access_type': 'offline',
     }
-    if action is not None:
-        query['action'] = action
     if enforce_2fa is True:
         # Specifying AAL2 will require the token to have an authentication
         # assurance level >= 2 which corresponds to requiring 2FA.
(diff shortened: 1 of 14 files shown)


diox:
Blocked by the fact that our oauth client config - at least the one for local development - aren't allowed to use `prompt=none` with `login_hint`. See https://mozilla.github.io/ecosystem-platform/reference/oauth-details#requesting-promptnone-during-authorization

src/olympia/accounts/utils.py
```diff
@@ -32,18 +32,37 @@ def fxa_login_url(
...
+        # have our oauth client configs be explicitly allowed.
+        # https://mozilla.github.io/ecosystem-platform/reference/oauth-details
+        # #promptnone-support
+        # FIXME: this highlights the fact that we need to handle error cases
+        # and regenerate a new state to start from scratch - right now we're
+        # ignoring them, so the user can't recover. I'm not sure what's
+        # supposed to happen if the user goes to FxA with prompt=none and
+        # login_hint=email if they no longer have a valid FxA session (ideally
+        # FxA would handle that gracefully), but if an error is raised we'll
+        # need to handle it.
```
diox, line 59: We should follow https://mozilla.github.io/ecosystem-platform/reference/oauth-details#recovering-from-errors and log the error code, remove `prompt=none`, generate a new state and try again to handle errors.

diox:
> Blocked by the fact that our oauth client config - at least the one for local development - aren't allowed to use `prompt=none` with `login_hint`. See https://mozilla.github.io/ecosystem-platform/reference/oauth-details#requesting-promptnone-during-authorization

https://mozilla-hub.atlassian.net/browse/SVCSE-1358 should have unblocked this.

src/olympia/accounts/utils.py
diox: Filed https://github.com/mozilla/addons-server/issues/21002 to do this

src/olympia/accounts/utils.py
```diff
@@ -8,17 +8,20 @@
...
 
 
+log = olympia.core.logger.getLogger('accounts')
+
+
 def fxa_login_url(
     config,
     state,
     next_path=None,
-    action=None,
```
diox, line 18: I removed `action` because we were only using `signin` and this does nothing - FxA only supports `email` or `force_auth` [according to their documentation](https://mozilla.github.io/ecosystem-platform/relying-parties/reference/query-parameters#action) (and we want neither of these).

PR #21015, opened 2023-07-18
override ALLOWED_HOSTS for prod to disallow non-production hosts

### LINKED ISSUE #20806: Split ALLOWED_HOSTS per env. We shouldn't allow non-prod hosts to have privileged access to prod addons-server.

eviljeff:
fixes #20806 ? 
Though we could maybe tighten it up further - I can't think where we'd run addons-server on a `.mozilla.com` or `.mozilla.net` domain (or a gcp domain right now, but it'd be annoying to have to change it later).

### DIFF
--- a/src/olympia/conf/prod/settings.py
+++ b/src/olympia/conf/prod/settings.py
@@ -1,6 +1,13 @@
 # ruff: noqa: F405
 from olympia.lib.settings_base import *  # noqa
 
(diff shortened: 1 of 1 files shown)


diox (approved):
I don't love prod-specific settings, as it's not covered by tests or even QA on dev/stage... but it looks reasonable.

PR #21023, opened 2023-07-19
Remove comments before processing reviewer actions that shouldn't have them

### LINKED ISSUE #20980: Comment is persisted when clearning pending rejections even if the reviewer doesn't see the comment
When clearing rejections on an add-on, if the reviewer first selects a different tab (like `approve`) before moving to clearing pending rejections, the clearing of pending rejections gets the message from the `approve` tab. 

### What happened?
- Go to review details page of an add-on that has rejection pending
- Go to the `approve` tab. Add message "XYZ" 
- Go to the `clear pending rejections` tab and select the versions to clear rejections
- Click Send
image

- The `clear pending rejections` should have either no comments (if there's no comment field, like is the case now) or its own comment field. It shouldn't persist comments from other tabs which are not visible when the reviewer clears pending rejections

### Additional details
I guess it doesn't matter all that much if the developer doesn't see this message, still thought I'd write this down here as it is a bit confusing.

diox:
Fixes #20980

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -808,10 +808,17 @@ class ReviewHelper:
         )
 
     def process(self):
-        action = self.handler.data.get('action', '')
-        if not action:
+        requested_action = self.handler.data.get('action', '')
+        if not requested_action or requested_action not in self.actions:
             raise NotImplementedError
-        return self.actions[action]['method']()
+        action = self.actions[requested_action]
+        # Clear comments in data before processing if the action isn't supposed
+        # to have any, because the reviewer might have submitted some by
+        # accident after switching between tabs, and the logging methods will
+        # automatically include them if some are present.
+        if not action.get('comments', True):
+            self.handler.data['comments'] = ''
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/utils.py
```diff
@@ -808,10 +808,17 @@ def get_actions(self):
         )
 
     def process(self):
-        action = self.handler.data.get('action', '')
-        if not action:
+        requested_action = self.handler.data.get('action', '')
+        if not requested_action or requested_action not in self.actions:
```
eviljeff, line 812: depending on your love for walruses, 
```suggestion
        if not (action := self.actions.get(self.handler.data.get('action'))):
```
(and replaces `action = self.actions[requested_action]` too)

eviljeff (approved):

src/olympia/reviewers/utils.py
diox: I like that, 849a04ab2d717a3c6b0fa445e1ab2a078286ff36

PR #21095, opened 2023-08-15
Fix API 403 error for "all_with_unlisted" when there are no unlisted.

### LINKED ISSUE #21063: API: Filtering addons versions for unlisted leads to 403, even with admin permissions.
1. As someone who has the permissions to filter by all_with_unlisted, use your api key and secret to generate a JWT.
2. Use this JWT to access `addons/addon/ublock-origin/versions?filter=all_with_unlisted`.

### What happened?
You are met with a 403 forbidden error, meaning you do not have the proper permissions.

The response should not error. Try this same endpoint (as someone who has viewing deleted versions permissions) `addons/addon/ublock-origin/versions?filter=all_with_deleted`. This will work. However, if you have the deleted perms, you also have the unlisted perms. So there must be something wrong with the handling of the logic for the specific `all_with_unlisted` permission.

Since this does not always happen, only with certain addons, it's possible that the issue lies when the addon actually doesn't have any unlisted versions, or if any of the versions are blocked.

GregSherman:
Fixes #21063 

Previously, there would be a 403 error no matter the permission classes when requesting an addon with no unlisted versions with the filter of all_with_unlisted. This fixes that and returns whatever listed versions exist.

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -4587,6 +4587,49 @@ class TestVersionViewSetList(AddonAndVersionViewSetDetailMixin, TestCase):
 
         self._test_url_contains_all(filter='all_with_unlisted')
 
+    def test_all_with_unlisted_when_no_unlisted_versions(self):
+        user = UserProfile.objects.create(username='reviewer')
+        self.grant_permission(user, 'Addons:Review')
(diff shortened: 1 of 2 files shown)


GregSherman:
@eviljeff Requesting review (I don't have permissions here)

diox (approved):

PR #21097, opened 2023-08-16
refactor to create NeedsHumanReview in log_and_notify if needed

### LINKED ISSUE #21087: Developer reply does not make a version get a due date
We have seen at least two examples where a developer reply does not make the corresponding version get a due date and thus appear in the review queue.

The code that handles this is:
https://github.com/mozilla/addons-server/blob/185aa864d7f6651ef6dd9e77838508d1c6f3360d/src/olympia/activity/views.py#L103-L115

After an investigation, @eviljeff noted the following:
_"that code ... is in the activity API viewset, which is used in devhub.  But emails don't use that viewset and there's no equivalent code there to set needs human review."_

eviljeff:
fixes #21087

### DIFF
--- a/src/olympia/activity/tests/test_utils.py
+++ b/src/olympia/activity/tests/test_utils.py
@@ -225,11 +225,15 @@ class TestEmailBouncing(TestCase):
 class TestAddEmailToActivityLog(TestCase):
     def setUp(self):
         self.addon = addon_factory(name='Badger', status=amo.STATUS_NOMINATED)
-        version = self.addon.find_latest_version(channel=amo.CHANNEL_LISTED)
+        self.version = self.addon.find_latest_version(channel=amo.CHANNEL_LISTED)
         self.profile = user_factory()
-        self.token = ActivityLogToken.objects.create(version=version, user=self.profile)
+        self.token = ActivityLogToken.objects.create(
+            version=self.version, user=self.profile
+        )
         self.token.update(uuid='5a0b8a83d501412589cc5d562334b46b')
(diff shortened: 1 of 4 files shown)


src/olympia/activity/tests/test_views.py
```diff
@@ -431,6 +431,7 @@ def _test_developer_reply(self):
 
     def test_developer_reply_listed(self):
         self._test_developer_reply()
+        print(self.version.due_date, get_review_due_date())
```
diox, line 434: Stray print

src/olympia/activity/tests/test_utils.py
```diff
@@ -225,25 +225,33 @@ def test_exception_but_spammy(self):
...
+        assert not self.version.due_date
 
     def test_developer_comment(self):
         self.profile.addonuser_set.create(addon=self.addon)
         note = add_email_to_activity_log(self.parser)
         assert note.log == amo.LOG.DEVELOPER_REPLY_VERSION
         self.token.refresh_from_db()
         assert self.token.use_count == 1
+        assert self.version.needshumanreview_set.exists()
+        assert self.version.reload().due_date
```
diox, line 245: Could test it's the right one too since the delay is custom (maybe including another test where it already had one, making sure we aren't overriding it)

diox (approved):

src/olympia/activity/tests/test_utils.py
eviljeff: we have those tests for the viewset, but we could duplicate them here too

PR #21105, opened 2023-08-17
Enforce 2FA at upload time

diox:
Fixes https://github.com/mozilla/addons-server/issues/20943

Note: when trying this, make sure `2fa-enforcement-for-developers-and-special-users` flag is active (and note that it's a flag, not a switch - to activate it, enable it for `everyone`). This should work both with fake and actual FxA implementations.

### DIFF
--- a/src/olympia/accounts/decorators.py
+++ b/src/olympia/accounts/decorators.py
@@ -1,0 +1,24 @@
+import functools
+
+import olympia.core.logger
+from olympia.accounts.utils import redirect_for_login_with_2fa_enforced
+
+
+# Needs to match accounts/views.py
+log = olympia.core.logger.getLogger('accounts')
+
+
+def two_factor_auth_required(f):
+    """Require the user to be authenticated and have 2FA enabled."""
+
+    @functools.wraps(f)
+    def wrapper(request, *args, **kw):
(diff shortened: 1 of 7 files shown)


wagnerand:
This also covers submissions via the API, right?

diox:
2FA is enforced on the API key page with this change. The API submissions themselves no, it's bypassing FxA entirely.

eviljeff (approved):

eviljeff:
> 2FA is enforced on the API key page with this change. The API submissions themselves no, it's bypassing FxA entirely.

Do we need to do anything about existing API keys generated by users who don't have 2FA enabled?  (perhaps not as 2FA protects the account from compromised logins, but API requests don't need a log in.)

diox:
> Do we need to do anything about existing API keys generated by users who don't have 2FA enabled? (perhaps not as 2FA protects the account from compromised logins, but API requests don't need a log in.)

The [doc](https://docs.google.com/document/d/1GcYrhucCBvlRYU4aXG2HZjW4FrYiLyorX3OpKTKg_k4/edit) highlights that as a future consideration - not just for API keys, but also existing developers that never logged in again after their submission. We may want to do some targeted email campaign or something to reach them.

PR #21108, opened 2023-08-18
Allow langpacks to have multiple listed versions awaiting review at the same time

diox:
This should be fine for langpacks because we don't care about their `current_version`.


eviljeff (commented):
Can we have a test to check both versions are picked to be signed by the auto approve cron?

PR #21112, opened 2023-08-18
don't subscribe reviewers with a reject_multiple_versions action

### LINKED ISSUE #21100: Default reviewer subscription to off
part of https://mozilla-hub.atlassian.net/browse/AOP-102
> When one or more versions are rejected, the reviewer will automatically subscribe to emails for when new versions are uploaded to the same channel.

Too many emails, so lets default it to not subscribing.

eviljeff:
fixes #21100 

(I was surprised to find we _only_ auto subscribed reviewers in the `reject_multiple_versions` action!)

### DIFF
--- a/src/olympia/reviewers/tests/test_utils.py
+++ b/src/olympia/reviewers/tests/test_utils.py
@@ -44,7 +44,6 @@ from olympia.reviewers.models import (
     AutoApprovalSummary,
     NeedsHumanReview,
     ReviewActionReason,
-    ReviewerSubscription,
 )
 from olympia.reviewers.utils import (
     ReviewAddon,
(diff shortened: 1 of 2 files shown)


diox (approved):
> (I was surprised to find we _only_ auto subscribed reviewers in the `reject_multiple_versions` action!)

Not _entirely_ surprising, I think the idea for that subscription was to automatically follow rejections and only rejections because that's when we're more interested in what happens next - we wanted to make sure that the developer doesn't try to work around the rejection by re-uploading the same thing, for instance...

eviljeff:
> > (I was surprised to find we _only_ auto subscribed reviewers in the `reject_multiple_versions` action!)
> 
> Not _entirely_ surprising, I think the idea for that subscription was to automatically follow rejections and only rejections because that's when we're more interested in what happens next - we wanted to make sure that the developer doesn't try to work around the rejection by re-uploading the same thing, for instance...

That's part of what I meant - it's only for reject _multiple_, not listed rejection.    Anyway, it's gone now.

PR #21152, opened 2023-09-05
Limit compatibility for Firefox for Android

### LINKED ISSUE #21139: Limit compatibility for non recommended/line extensions on Android to either Fennec or Fenix in DevHub
To prepare for the [upcoming general availability of extensions on Firefox for Android](https://blog.mozilla.org/addons/2023/08/10/prepare-your-firefox-desktop-extension-for-the-upcoming-android-release/):

- In the version edit page, continue to allow developers to set compatibilty as they see fit, however non recommended/line extensions for Android have to choose:
  - Either they are compatible with Firefox for Android `68.*` max (with whatever min below that). If they choose that option, `strict_compatibility` should be enabled.
  - Or, they are compatible with Firefox for Android `119.0` or `120.0` min (with whatever max above that, probably `*`) and `strict_compatibility` should be disabled if it was enabled.
- On upload (API or DevHub), if Firefox for Android is selected, force non recommended/line extension to be compatible with min `119.0` and max `*` (unless they had strictly chosen `68.*` as their max through `gecko.strict_max_version`, in which case obey that but set `strict_compatibility` like above)
  
This is all to ensure that for now, only users of Fenix `119.0` Nightly/Beta get access to all Android add-ons for testing while non-recommended/line (1) extensions remain non-installable/searchable in Fenix Release versions, until we eventually enable general availability of Android extensions.

(1) Specifically non-recommended/line extensions *for Android*. If an extension is recommended but only for desktop, they get the same restrictions regarding Android compatibility as everybody else.

diox:
- Extensions compatible with Android are forced to choose between being compatible with either Fennec or Fenix >= 119.0, unless they are recommended/line.
- If they are compatible with Fennec, strict compatibility is enabled (so that we don't allow them to be installed with Fenix).

Migration of existing data will happen later in a separate commit.

Fixes #21139


src/olympia/versions/models.py
```diff
@@ -1423,6 +1438,90 @@ def __str__(self):
...
+            # If the range max is below the first Fenix version, then we need
+            # to set strict_compatibility to True on the File to prevent that
+            # version from being installed in Fenix. If it's not, then we need
+            # to clear it if it was set.
+            if self.max.version_int < version_int(amo.MIN_VERSION_FENIX):
+                strict_compatibility = True
+            else:
+                strict_compatibility = False
+            if self.version.file.strict_compatibility != strict_compatibility:
+                self.version.file.update(strict_compatibility=strict_compatibility)
```
diox, line 1518: Just realized a couple issues with that I need to address:
- If someone chooses to be compatible with Fennec and gets `strict_compatibility` set to `True`, it would also apply to potential users on Desktop. So if their `max` for `FIREFOX` is set to something other than `*`, we'd obey it... I think that's an edge case we're fine with, but needs to be documented.
- I forgot to add support for deleting Firefox for Android compatibility entirely, which should drop the `strict_compatibility` if it's an extension.

eviljeff:
> If they are compatible with Fennec

Why are we bothering with Fennec compatibility?  The last Fennec release has to have been years ago... so lets just disable those add-ons?

diox:
Well most add-ons are going to also be compatible with Firefox Desktop, so we can't disable them. We could remove compatibility with Firefox for Android, but I wanted to err on the side of caution and keep the info that they originally were marked as compatible with Android.

You would have completely removed the `ApplicationsVersions` (and even the `AppVersion` ?) for Firefox for Android before Fenix ?

src/olympia/versions/models.py
```diff
@@ -1423,6 +1438,103 @@ def __str__(self):
...
+            self.application == amo.ANDROID.id
+            and self.version.addon.type == amo.ADDON_EXTENSION
+        ):
+            # Regardless of whether the compat information came from the
+            # manifest, on Android we currently only allow non-promoted/line
+            # extensions to be compatible with Fennec or Fenix that has general
+            # availability, so we might need to override the compat data before
+            # saving it.
+            if self.version_range_contains_forbidden_compatibility():
+                # Arbitrarily pick a range that makes sense and is valid.
```
eviljeff, line 1498: Should we only override the min or max down/up if it needs to be?  i.e. if the developer sets a max that is above `MIN_VERSION_FENIX_GENERAL_AVAILABILITY` don't override it

src/olympia/versions/models.py
```diff
@@ -1423,6 +1438,103 @@ def __str__(self):
...
+            del self.version._compatible_apps
+        return rval
+
+    def delete(self, *args, **kwargs):
+        if (
+            self.application == amo.ANDROID.id
+            and self.version.addon.type == amo.ADDON_EXTENSION
+            and self.version.file.strict_compatibility
+        ):
+            self.version.file.update(strict_compatibility=False)
```
eviljeff, line 1535: There's an edge case here where `strict_compatibility` was `True` anyway (we didn't override it) but now we're setting it wrongly to `False`.

eviljeff: (That we probably don't care about, just calling it out)

eviljeff:
> Well most add-ons are going to also be compatible with Firefox Desktop, so we can't disable them. We could remove compatibility with Firefox for Android, but I wanted to err on the side of caution and keep the info that they originally were marked as compatible with Android.
> 
> You would have completely removed the `ApplicationsVersions` (and even the `AppVersion` ?) for Firefox for Android before Fenix ?

Yeah, if that was the cleanest solution.

eviljeff (approved):

src/olympia/versions/models.py
diox: Yeah. But there shouldn't be any. The only extension in prod with `strict_compatibility` enabled is one of ours, an old hotfix add-on from 2019 that is currently disabled. Was the comment I put not enough ? I can reword it if needed.

src/olympia/versions/models.py
diox: From looking at the database `max` is rarely set to something other than `*` (which makes sense) so I figured this wasn't worth the extra complexity.

src/olympia/versions/models.py
```diff
@@ -1423,6 +1438,103 @@ def __str__(self):
...
+
+        if addon.type != amo.ADDON_EXTENSION:
+            # We don't care about non-extension compatibility for this.
+            return False
+
+        # Recommended/line for Android are allowed to set compatibility in the
+        # limited range.
+        if (
+            addon.promoted
+            and addon.promoted.group in (RECOMMENDED, LINE)
```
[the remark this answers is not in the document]
diox: :+1: did that in 0a6407bf7f9e9557399795e65b1d173d18267bef

src/olympia/versions/models.py
```diff
@@ -1423,6 +1438,103 @@ def __str__(self):
...
+        is_range_bigger_than_limited_range = self.min.version_int < version_int(
+            self.ANDROID_LIMITED_COMPATIBILITY_RANGE[0]
+        ) and self.max.version_int >= version_int(
+            self.ANDROID_LIMITED_COMPATIBILITY_RANGE[1]
+        )
+        return (
+            is_min_in_limited_range
+            or is_max_in_limited_range
+            or is_range_bigger_than_limited_range
+        )
```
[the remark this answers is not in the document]
diox: I pushed e10591b5167a33d441a0065e6f52f449886fa74b to make things easier to read.

The 3 different cases are important - ultimately that's what inside the range that matters.

diox:
>> You would have completely removed the `ApplicationsVersions` (and even the `AppVersion` ?) for Firefox for Android before Fenix ?
>>
> Yeah, if that was the cleanest solution.

Ok I see. The reason I didn't go with that is that there are still a lot of uncertainty around which add-ons are truly compatible, which add-ons are not, so completely losing the data was problematic. By keeping it around (both in this code and the migration PR I've yet to submit) we ensure the data isn't accidentally lost if we need to force a different value later.

src/olympia/versions/models.py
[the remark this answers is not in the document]
eviljeff: The code suggestion is (supposed to be) a replacement for all three of the different cases - albeit written in a non-obvious way.

src/olympia/versions/models.py
eviljeff: The comment was in a different function (save), but if it's in reality an edge case that doesn't happen it doesn't matter anyway.

src/olympia/versions/models.py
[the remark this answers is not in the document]
diox: Ah, what confused me is that the first condition in your suggestion should be a `>` (or we should reverse the ordering and use `<`) then. Otherwise it would not allow `119.0a1` as the `min`.

I'm not sure which is better tbh.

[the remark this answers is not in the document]
diox: Ended up using your suggestion with that fix and a couple more more tests in ef6b2428d4a17e84d3bd16cbaef0a4eeff5d18b8

Issue #5989, opened 2017-07-24
Unify categories


muffinresearch:
We've talked about this here and there before and I think the consensus was that unifying the add-on categories in AMO is something we want to do.

To that end do we can we sort out what the mappings should be and discuss what it would take to make the change?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-149)

jvillalobos:
Just to clarify scope, we're referring to the categories for extensions on desktop and Android, right?

Other categories, which presumably are out of scope:
* Thunderbird and SeaMonkey: we plan to move to a separate site.
* Dictionaries and Language Packs: they have a single category per application, but that's just an implementation detail for not having categories.
* Themes and Complete Themes: they have their own categories, and the latter are going away with legacy add-ons.
* Search Tools: they have different categories, would be hard to consolidate because they relate more to content than features. Not sure if categories are useful for them.

muffinresearch:
> Just to clarify scope, we're referring to the categories for extensions on desktop and Android, right?

I think themes would be included in this since that's currently part of the issue - see #4766 for more info.

eviljeff:
Also clarifying scope: this should be able unifying categories for the different _applications_, not the different add-on types.  Lang packs, themes, etc, are different types to Extensions, so will have different categories still.

jvillalobos:
Agree that themes (lightweight and future static) should share one set of categories that work both on desktop and Android.

eviljeff:
@jvillalobos can you (or @devaneymoz or as appropriate) provide mappings from the old to new combined categories?

Desktop Firefox has: 
```
'alerts-updates': StaticCategory(id=72, name=_(u'Alerts & Updates')),
'appearance': StaticCategory(id=14, name=_(u'Appearance')),
'bookmarks': StaticCategory(id=22, name=_(u'Bookmarks')),
'download-management': StaticCategory(id=5, name=_(u'Download Management')),
'feeds-news-blogging': StaticCategory(id=1, name=_(u'Feeds, News & Blogging')),
'games-entertainment': StaticCategory(id=142, name=_(u'Games & Entertainment')),
'language-support': StaticCategory(id=37, name=_(u'Language Support')),
'photos-music-videos': StaticCategory(id=38, name=_(u'Photos, Music & Videos')),
'privacy-security': StaticCategory(id=12, name=_(u'Privacy & Security')),
[...]

PR #21160, opened 2023-09-06
don't have categories for static themes on Android

### LINKED ISSUE #20780: Firefox for Android as category should not be visible for static themes in dev hub
Followup for https://github.com/mozilla/addons-server/issues/20757#issuecomment-1569974139

There's still information left for Categories in dev hub this is displayed:

categories

eviljeff:
fixes #20780 - we don't actually need to clear up the `AddonCategory` instances - just dropping the constants is enough to fix the issue but I figured i was cleaner to do so.


src/olympia/addons/migrations/0047_auto_20230906_1220.py
```diff
@@ -0,0 +1,37 @@
...
+    'seasonal': 412,
+    'solid': 418,
+    'sports': 404,
+    'websites': 416,
+}.values()
+
+
+def delete_android_static_theme_addon_category(apps, schema_editor):
+    AddonCategory = apps.get_model('addons', 'AddonCategory')
+    AddonCategory.objects.filter(category_id__in=ANDROID_STATICTHEME_CAT_IDS).delete()
```
diox, line 27: mmm that's a 500K rows `DELETE`... I think I would rather have that as a cleanup command that does it in chunks than something that blocks the push.

eviljeff: It's harmless to keep them there but we could spin it out into a general clean-up for all old categories (or do it with/after #5989)

diox: Yeah drop a note in that issue, leave them for now.

diox (approved):

PR #21167, opened 2023-09-07
still have a link in the queue for addons that have an empty name

### LINKED ISSUE #20785: Link add-ons without names to their review page from the review queue
Add-ons without names show an unlinked `—` in the review queue

image

While it is possible to get the add-on ID from the review queue's HTML table row element (`tr#addon-<id>`) and get to its review page, it would be more convenient to just link it like other add-ons with names. 

Personally, I don't see why add-ons without names (or whitespaces for names) would be allowed, but that doesn't have to change for this issue to be fixed.

### What happened?
Add-on without name (or with whitespace name) doesn't get linked from review queue to its review page

Add-on without name should be linked from the review queue to its review page

eviljeff:
This was quite an investigation to find out where the em-dash was coming from, but it seems, basically, that an empty string triggered a default value that _didn't_ call the render function for the add-on name.

fixes #20785 - it display the add-on id instead for the link text when the name is empty
Screenshot 2023-09-07 18 11 45

### DIFF
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -1526,6 +1526,25 @@ class TestExtensionQueue(QueueTest):
             '🛠️ Manual Review', tab_position=0, total_addons=4, total_queues=2
         )
 
+    def test_empty_name(self):
+        self.get_expected_addons_by_names(
+            ['Nominated One'],
+            auto_approve_disabled=True,
+        )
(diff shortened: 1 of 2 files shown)


willdurand:
> fixes #5989

are you sure about that?

diox:
correct issue is https://github.com/mozilla/addons-server/issues/20785

diox (approved):

Issue #21162, opened 2023-09-06
Alter Android compatibilty of existing extensions to prepare for general availability of extensions on Fenix


diox:
We need to add a set of commands to run some data migrations to prepare for general availability of extensions on Fenix.

This is going to happen in multiple steps as we gradually expand the set of add-ons we're allowing in Beta/Nightly before fully opening up general availability.

For this step, which should land before Fenix 119 Beta is released:
- Force specific extensions as compatible with Firefox for Android 119.0a1 or higher from a csv file.
  - Should happen regardless of whether or not they were already marked as compatible with another version of Firefox for Android or not
  - Should drop `strict_compatibility` on relevant files
  - Should *not* alter compatibility of the current limited set of recommended/line extensions at any point, so that they continue to be installable by current Fenix release. If those end up in the csv file, they should be ignored.
- After that migration has been executed, force **all** versions in the database marked as compatible with Firefox for Android lower than 119.0a1 and not recommended/line for Android/All applications to be marked as compatible up to 68.* max instead
  - Should enable `strict_compatibility` on those files as well

diox:
QA: when this lands, it won't do anything by default, you'll need to build a test csv with some add-on ids to test the various states described in the description, and share it with me so that I can run the commands. Then you should be able to check the compatibility data in devhub/admin tools (or API).

ioanarusiczki:
@diox I prepared a list of IDs for AMO dev.

diox:
For reference, using [ids.csv](https://github.com/mozilla/addons-server/files/12673807/ids.csv) on dev (which contains a mix of everything prepared by @ioanarusiczki):

Skipping add-on id 328839 because of its promoted group.
Skipping add-on id 491253 because of its promoted group.
Skipping add-on id 491817 because of its promoted group.
Skipping add-on id 497311 because of its promoted group.
Skipping add-on id 618623 because of its promoted group.
Skipping add-on id 620138 because of its promoted group.
Skipping add-on id 620247 because of its promoted group.
Skipping add-on id 620395 because of its promoted group.
Skipping add-on id 621034 because of its promoted group.
Skipping add-on id 621035 because of its promoted group.
Skipping add-on id 621036 because of its promoted group.
Skipping add-on id 625216 because of its promoted group.
Done forcing Android compatibility on 34 add-ons (12 skipped)
Done forcing max Android compatibility: Dropped compat on 494 versions, Updated compat for 6101 versions, Set strict compatibility on 6101 files

[...]

PR #21190, opened 2023-09-13
Commands to alter Android compatibilty to prepare for general availability

diox:
Fixes https://github.com/mozilla/addons-server/issues/21162

### DIFF
--- a/src/olympia/constants/base.py
+++ b/src/olympia/constants/base.py
@@ -359,6 +359,9 @@ DEFAULT_WEBEXT_MIN_VERSION_GECKO_ANDROID = '113.0'
 # First version we consider as "Fenix".
 MIN_VERSION_FENIX = '79.0a1'
 
+# Last version we consider as "Fennec"
+MAX_VERSION_FENNEC = '68.*'
+
 # The minimum version of Fenix where extensions are all available. Expect this
 # to be bumped to 120.0 later.
 MIN_VERSION_FENIX_GENERAL_AVAILABILITY = '119.0a1'
--- a/src/olympia/versions/management/commands/force_min_android_compatibility.py
+++ b/src/olympia/versions/management/commands/force_min_android_compatibility.py
@@ -1,0 +1,86 @@
+import csv
+
+from django.core.management.base import BaseCommand
+from django.db.transaction import atomic
+
+import olympia.core.logger
+from olympia import amo
+from olympia.addons.models import Addon
+from olympia.applications.models import AppVersion
+from olympia.versions.models import ApplicationsVersions
+
+
+log = olympia.core.logger.getLogger('z.versions.force_min_android_compatibility')
+
+
+class Command(BaseCommand):
+    """
+    Force current version of add-ons in the specified csv to be compatible with
+    Firefox for Android <MIN_VERSION_FENIX_GENERAL_AVAILABILITY> and higher.
+
+    Should not affect compatibility of add-ons recommended/line for Android.
+    """
+
+    help = (
+        'Force add-ons to be compatible with Firefox for Android '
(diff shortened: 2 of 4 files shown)


src/olympia/constants/base.py
```diff
@@ -359,6 +359,9 @@
 # First version we consider as "Fenix".
 MIN_VERSION_FENIX = '79.0a1'
 
+# Last version we consider as "Fennec"
+MAX_VERSION_FENNEC = '68.*'
```
eviljeff, line 363: If we consider the minimum version of Fenix to be 79, and the maximum version of Fennec to be 68, what do we consider versions from 69 to 78 to be?

src/olympia/versions/tests/test_commands.py
```diff
@@ -0,0 +1,329 @@
...
+            addon_factory(
+                name='Normal for Android 119',
+                version_kw={
+                    'application': amo.ANDROID.id,
+                    'min_app_version': '119.0a1',
+                    'max_app_version': '*',
+                },
+                promoted=RECOMMENDED,
+            ),
+        ]
```
eviljeff, line 188: Can we have some add-ons that are promoted for both Android and Desktop, other promoted groups than Line and Recommended so test they are reset as expected.

src/olympia/versions/management/commands/force_min_android_compatibility.py
```diff
@@ -0,0 +1,86 @@
...
+            'CSVFILE',
+            help='Path to CSV file containing add-on ids.',
+        )
+
+    def read_csv(self, path):
+        with open(path) as file_:
+            csv_reader = csv.reader(file_)
+            # Format should be a single column with the add-on id.
+            # Ignore non-numeric to avoid the column header.
+            return [int(row[0]) for row in csv_reader if row[0] and row[0].isnumeric()]
```
eviljeff, line 40: how is this going to handle trailing and leading spaces? (i.e. should we `trim` `row[0]`)?

eviljeff (approved):

src/olympia/versions/management/commands/force_min_android_compatibility.py
diox: `int()` shouldn't care. I'll add that in the tests

eviljeff: `int()` may not care but afaik `isnumeric()` returns false

src/olympia/constants/base.py
diox: One thing I could do for consistency is remove `MIN_VERSION_FENIX` completely, and use `MAX_VERSION_FENNEC` to find the beginning of the forbidden range instead. The reason I didn't is because `68.*` is an invalid min version (since it has a `*`) so I couldn't do a simple `.get()` to find the beginning of the forbidden range. But I could work around that and pick the first version after... (it would likely be `69.0`, even though that version doesn't really exist for Android)

src/olympia/versions/management/commands/force_min_android_compatibility.py
diox: Good point. (`is_numeric()` is also super permissive but it shouldn't be a problem - if wrong data is given we want to blow up anyway...)

diox: https://github.com/mozilla/addons-server/pull/21190/commits/536a85359359250fbcae3e6433dc548859bbd0d9

src/olympia/versions/tests/test_commands.py
diox: https://github.com/mozilla/addons-server/pull/21190/commits/ea2458488f142484f59763c679464dea23779af4

PR #21198, opened 2023-09-14
Raise error when specifying compatibility inside forbidden Android range in the API

### LINKED ISSUE #21195: Throw an error message when using the endpoint in case of invalid compatibility range
Followup for https://github.com/mozilla/addons-server/issues/21139#issuecomment-1717636641

On dev hub: 
For non-recommended when the interval 78.0 - 119.* is picked from edit version page it would throw an error Invalid version range. 

invalid range

With the API: 

If I sent a PATCH request such as {"compatibility": {"android": {"min":"78.0","max": "119."}}} I'll get a 200 response but the with "compatibility": {
"android": {
"min": "119.0a1",
"max": ""
}

diox:
Only do it when the full compatibility object is specified, we silently rewrite invalid ranges anyway.

Fixes #21195

### DIFF
--- a/src/olympia/addons/fields.py
+++ b/src/olympia/addons/fields.py
@@ -271,9 +271,7 @@ class VersionCompatibilityField(serializers.Field):
                 if 'max' in min_max:
                     apps_versions.max = app_version_qs.get(version=min_max['max'])
                 elif version:
-                    apps_versions.max = app_version_qs.get(
-                        version=amo.DEFAULT_WEBEXT_MAX_VERSION
-                    )
+                    apps_versions.max = apps_versions.get_default_maximum_appversion()
 
             except AppVersion.DoesNotExist:
                 raise exceptions.ValidationError(
@@ -285,24 +283,21 @@ class VersionCompatibilityField(serializers.Field):
                 if 'min' in min_max:
                     apps_versions.min = app_version_qs.get(version=min_max['min'])
                 elif version:
-                    apps_versions.min = app_version_qs.get(
-                        version=amo.DEFAULT_WEBEXT_MIN_VERSIONS[app]
-                    )
+                    apps_versions.min = apps_versions.get_default_minimum_appversion()
             except AppVersion.DoesNotExist:
                 raise exceptions.ValidationError(
                     gettext('Unknown min app version specified')
                 )
 
-            # For creations, at this point we have a potentially incomplete
-            # ApplicationsVersions instance, which could be missing its min
-            # and/or max. In Version.from_upload() we'll fill in the gaps with
-            # the manifest data, but we can't do that here, so we only validate
-            # the range for Android if we had both a min and a max provided.
-            # ApplicationsVersions.save() will silently correct min/max before
-            # saving if it's part of the forbidden range for Android anyway.
+            # We want to validate whether or not the android range contains
+            # forbidden versions. At this point the ApplicationsVersions
+            # instance might be partial and completed by Version.from_upload()
+            # which looks at the manifest. We can't do that here, so if either
+            # the min or max provided would lead to a problematic range on its
+            # own, we force the developer to explicitly set both, even though
+            # what's in the manifest could resulted in a valid range.
             if (
-                'min' in min_max
-                and 'max' in min_max
+                ('min' in min_max or 'max' in min_max)
                 and apps_versions.application == amo.ANDROID.id
                 and apps_versions.version_range_contains_forbidden_compatibility()
             ):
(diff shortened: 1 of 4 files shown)


src/olympia/addons/fields.py
```diff
@@ -291,6 +293,28 @@ def to_internal_value(self, data):
...
                 )
 
+            # For creations, at this point we have a potentially incomplete
+            # ApplicationsVersions instance, which could be missing its min
+            # and/or max. In Version.from_upload() we'll fill in the gaps with
+            # the manifest data, but we can't do that here, so we only validate
+            # the range for Android if we had both a min and a max provided.
+            # ApplicationsVersions.save() will silently correct min/max before
+            # saving if it's part of the forbidden range for Android anyway.
+            if (
```
eviljeff, line 303: So, if the request specifies `'min': '48.0` but no max, so the default of `'*'` is used, this exception won't raise, and it'll instead be silently corrected to X? 
What would X be?  And would `min` be respected?

eviljeff: I can see the justification that if you don't specify a min or max the default isn't defined so correcting is fine (though still a bit magical).  Silently overriding the manifest min/max and having different behaviour if the developer makes a decision to explicitly define min or max in the manifest, compared to if the developer makes a decision to explicitly define min or max via the API... seems inconsistent.

eviljeff: Idea: if we move the validation into the serializer - where the manifest data is available - we could calculate the resulting range based on the API, manifest, and defaults, and error if the resulting range is forbidden (perhaps generalizing `version_range_contains_forbidden_compatibility` into a utility function).

diox: I have been looking into this. The manifest data is available, but we would need to move the logic in `Version.from_upload()` that merges `parsed_data` with `compatibility`/`selected_apps`... That'd be quite the refactor.

diox: @eviljeff I don't think it's worth spending too much time/effort on this. Once Fenix allows all extensions this won't really matter (we won't actively support older Fenix versions for long).

eviljeff: I'm still not a fan of the inconsistent behaviour between min/max provided in the API request JSON and the manifest.  That `'min': '48.0', 'max': '*'` in the JSON would raise, but `'min': '48.0'` in the API JSON and `'strict_max_version': '*'` in the manifest JSON wouldn't. (right?).  And just `'min': '48.0'`, leaving the default as '*', wouldn't raise either.

Would `'min' in min_max or 'max' in min_max` be more consistent, I wonder?  (it wouldn't take into account the manifest data, but it would force you to specify a matching compliant `max` and/or `min` if you provided one of them.)

src/olympia/addons/fields.py
```diff
@@ -291,6 +293,28 @@ def to_internal_value(self, data):
...
+                'min' in min_max
+                and 'max' in min_max
+                and apps_versions.application == amo.ANDROID.id
```
eviljeff, line 306: nit, `version_range_contains_forbidden_compatibility` checks application so this condition is redundant (if we don't care about the overhead of an extra function call, which we generally don't in python)

src/olympia/addons/fields.py
eviljeff: (I agree that refactoring `from_upload` is too much)

eviljeff (approved):
I don't disagree enough to block approval 🙂

src/olympia/addons/fields.py
diox: Yes. `version_range_contains_forbidden_compatibility()` is technically not Android-specific, through its current implementation is. I was worried that if it was modified later to add additional more generic checks, we'd raise the wrong error message here (since the message _is_ Android-specific)

src/olympia/addons/fields.py
diox: I like the idea of doing `'min' in min_max or 'max' in min_max`, yes. Forcing you to specify a matching compliant `max` and/or `min` if you provided one is appealing. Going to look into that solution.

src/olympia/addons/tests/test_views.py
```diff
@@ -3158,7 +3158,26 @@ def test_compatibility_forbidden_range_android(self):
         }
 
         self.make_addon_promoted(self.addon, RECOMMENDED, approve_version=True)
-        response = self.request(compatibility={'android': {'min': '48.0', 'max': '*'}})
+        response = self.request(
+            compatibility={'android': {'min': '119.0a1', 'max': '*'}}
```
eviljeff, line 3162: I see you've changed this request so it's different than the non-recommended request.  Does it work differently for recommended add-ons too now?

diox: No, you're right, I shouldn't have made that change. Fixed in c50baf291088ffb80c924ad91c24096b3d14b06a

PR #21202, opened 2023-09-15
Firefox Accounts to Mozilla accounts

### LINKED ISSUE #21161: Rename Firefox Accounts to Mozilla Accounts everywhere (server)
part of https://mozilla-hub.atlassian.net/browse/AOP-119

eviljeff:
fixes #21161 - apart from the few mentions in the docs, which we can deal with separately (if even necessary)


src/olympia/constants/activity.py
```diff
@@ -405,9 +407,21 @@ class USER_DELETED(_LOG):
...
+class _USER_AUTO_DELETED_METACLASS(type):
+    """Using metaclass is a workaround because we need a class property,
+    (and chaining @classmethod and @property is deprecated)"""
+
+    @property
+    def format(cls):
+        return (
+            _('Account {user} deleted, from Mozilla accounts event.')
+            if amo.is_mza_branding()
+            else _('Account {user} deleted, from Firefox Accounts event.')
```
diox, line 419: Given what it is about, this will never be exposed to end-users (we can add `admin_event = True` to be certain), it's purely for admins.

Which means it shouldn't need to be translated, and we could word it in a different way to not have to deal with branding issues, avoiding the need for the metaclass workaround.

eviljeff: https://github.com/mozilla/addons-server/pull/21202/commits/994e5b3000f9a8f13049d76327c7882aa2e93776

src/olympia/constants/activity.py
```diff
@@ -407,7 +407,8 @@ class USER_DELETED(_LOG):
 
 class USER_AUTO_DELETED(_LOG):
     id = 62
-    format = _('Account {user} deleted, from Firefox Accounts event.')
+    format = _('Account {user} deleted, from FxaNotificationView event.')
```
diox, line 410: Remove the `_()`

src/olympia/users/templates/users/emails/user_deleted.ltxt
```diff
@@ -1,9 +1,17 @@
-{% load i18n %}{# L10n: This is an email. Whitespace matters! #}{% blocktrans %}Hello,
+{% load i18n %}{% if is_mza_branding %}{# L10n: This is an email. Whitespace matters! #}{% blocktrans %}Hello,
+
+You’re receiving this message because your user account {{ name }} on {{ site_url }} has been deleted. This could have been done automatically if you recently deleted your Mozilla account (we renamed Firefox Accounts to Mozilla accounts).
```
diox, line 3: The "we renamed" bit might need to be validated by someone. Do we really need that ? It's on extra thing to remove later...

diox (approved):
r+wc

src/olympia/users/templates/users/emails/user_deleted.ltxt
eviljeff: It's "strongly recommended" to have messaging about the name transition - the figma link on https://mozilla-hub.atlassian.net/browse/FXA-7993 shows messaging that says "We've renamed Firefox accounts to Mozilla Accounts.  You'll still sign in with the same username and password, and there are no other changes to the products you use. [Learn more](https://support.mozilla.org/kb/firefox-accounts-renamed-mozilla-accounts)".  

I stripped that down because this is an email you get after you've deleted your account so I didn't think telling them about how to log in or our products made much sense.  Maybe we should include the url to find out more though.

diox: Ok. I think that's good enough as-is. It's a fairly specific email, no need to be over concerned about it.

PR #21204, opened 2023-09-18
Switch remote settings URL to new GCP-specific endpoint on stage & prod

diox:
Reverts mozilla/addons-server#20729

### DIFF
--- a/src/olympia/conf/prod/settings.py
+++ b/src/olympia/conf/prod/settings.py
@@ -68,7 +68,9 @@ EXTENSION_WORKSHOP_URL = env(
 )
 
 REMOTE_SETTINGS_API_URL = 'https://firefox.settings.services.mozilla.com/v1/'
-REMOTE_SETTINGS_WRITER_URL = 'https://remote-settings.mozilla.org/v1/'
+REMOTE_SETTINGS_WRITER_URL = env(
+    'REMOTE_SETTINGS_WRITER_URL', default='https://remote-settings.mozilla.org/v1/'
+)
 REMOTE_SETTINGS_WRITER_BUCKET = 'staging'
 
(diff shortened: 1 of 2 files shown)


src/olympia/conf/prod/settings.py
```diff
@@ -68,7 +68,7 @@
 )
 
 REMOTE_SETTINGS_API_URL = 'https://firefox.settings.services.mozilla.com/v1/'
-REMOTE_SETTINGS_WRITER_URL = 'https://settings-writer.prod.mozaws.net/v1/'
+REMOTE_SETTINGS_WRITER_URL = 'https://remote-settings.mozilla.org/v1/'
```
eviljeff, line 71: (I'm aware _just_ landing this patch is what I suggested but) maybe we should add `env(...)` so SRE can override in the future?  ... Then again, the more settings that can be overridden by SRE in obscure scripts the more divergence between what we think are the settings are and what they actually are....

eviljeff (approved):

src/olympia/conf/prod/settings.py
diox: Yeah, I had an internal debate about this... It does feel like it should be overridable by the environement, being a completely internal setting, but I wasn't sure we wanted to to add some more potential for confusion right now... The env variable doesn't currently exist in dev/stage/prod though, so I guess it's safe...

diox: @eviljeff the real question: should we default to the AWS URL, and wait for Wei to add the env variables for GCP, or default to GCP, and consider that if we need to switch back to AWS, we'll override using the env variables ?

eviljeff: default to GCP, imo

PR #21212, opened 2023-09-20
Merge per-application categories, add command to merge and clean up obsolete ones

### LINKED ISSUE #5989: Unify categories

diox:
Fixes #5989


diox:
Need to fix a bunch of tests, but also the API: we currently expose categories per app in the API, which no longer makes sense in this patch...

src/olympia/devhub/forms.py
```diff
@@ -1448,31 +1417,3 @@ def clean(self):
         if not checker.is_submission_allowed(check_dev_agreement=False):
             raise forms.ValidationError(checker.get_error_message())
         return self.cleaned_data
-
-
-class SingleCategoryForm(forms.Form):
```
eviljeff, line 1453: Possibly intentional, but pointing out that we're dropping the special class for themes which currently only allows a single category (it's a radio select) for one that allows two categories (multiple select).

eviljeff (approved):

src/olympia/devhub/forms.py
diox: We no longer need `SingleCategoryForm` because `CategoryForm` has the logic to only allow a single category for themes, and we no longer need `CategoryFormSet` because a single `CategoryForm` allows developers to select all the categories for their add-on, since we don't have per-application categories anymore.

PR #21218, opened 2023-09-25
implement cinder reporting

### LINKED ISSUE #21193: Add cinder.co API integration
part of [https://mozilla-hub.atlassian.net/browse/AOP-121](https://mozilla-hub.atlassian.net/browse/AOP-121)

AMO needs to be able to forward reports of abusive content to cinder.co, so we need to add the API integration.

eviljeff:
fixes #21193 - implements basic integration with cinder, with add-on and user reporting.  Follow-up issues will refine what metadata should be sent for each content type, and add rating & collection content reporting.

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -1,0 +1,127 @@
+from django.conf import settings
+
+import requests
+
+
+class Cinder:
+    QUEUE = 'amo-content-infringement'
+    type = None  # Needs to be defined by subclasses
+
+    @property
+    def id(self):
+        # Ideally override this in subclasses to be more efficient
+        return str(self.get_attributes().get('id', ''))
+
+    def get_attributes(self):
+        raise NotImplementedError
+
+    def get_context(self):
+        raise NotImplementedError
+
+    def get_entity_data(self):
+        return {'entity_type': self.type, 'attributes': self.get_attributes()}
+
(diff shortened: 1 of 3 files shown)


src/olympia/abuse/cinder.py
```diff
@@ -0,0 +1,127 @@
...
+    def get_attributes(self):
+        return {
+            'id': self.id,
+            'name': self.user.display_name,
+            'email': self.user.email,
+            'fxa_id': self.user.fxa_id,
+        }
+
+    def get_context(self):
+        addons = [CinderAddon(addon) for addon in self.user.addons.all()]
```
diox, line 92: Just a note for later: Might want to avoid non-translations transforms here? Depends on whether or not we'll need extra data later that the transforms are bringing us.

diox (approved):
Since you depend on the add-on name, you might want to make sure when calling this that you're always in the `en-US` locale (if this is strictly called from a task then it should already default to `en-US` and only be overridden in specific cases by a context manager)

eviljeff:
> Since you depend on the add-on name, you might want to make sure when calling this that you're always in the `en-US` locale (if this is strictly called from a task then it should already default to `en-US` and only be overridden in specific cases by a context manager)

My thinking was that add-on's name/metadata in the locale of the reporter would be best, because that's possibly what they are reporting.  Although we also probably want to include _all_ locales.  Exact metdata tbd.

PR #21221, opened 2023-09-25
add logging for changed slug on devhub

### LINKED ISSUE #21174: Add log activity when slug is changed
When the add-on slug is changed, we should log an activity containing:
* Add-on the slug was changed for
* User account that changed the slug
* Old slug
* New slug

GregSherman:
Fixes #21174 

When the slug in devhub is changed, create an activity log.

### DIFF
--- a/src/olympia/constants/activity.py
+++ b/src/olympia/constants/activity.py
@@ -953,6 +953,12 @@ class CLEAR_ADMIN_REVIEW_THEME(_LOG):
     reviewer_review_action = True
     admin_event = True
 
+class ADDON_SLUG_CHANGED(_LOG):
+    id = 182
+    format = _('Addon {addon} slug changed from {old} to {new}.')
+    short = _('Addon slug changed')
+    keep = True
+    show_user_to_developer = True
 
 LOGS = [x for x in vars().values() if isclass(x) and issubclass(x, _LOG) and x != _LOG]
 # Make sure there's no duplicate IDs.
--- a/src/olympia/devhub/tests/test_views_edit.py
+++ b/src/olympia/devhub/tests/test_views_edit.py
@@ -137,6 +137,11 @@ class BaseTestEditDescribe(BaseTestEdit):
         doc = pq(response.content)
         assert doc('form').attr('action') != old_edit
 
+        activity_log = ActivityLog.objects.latest('pk')
+        assert activity_log.action == amo.LOG.ADDON_SLUG_CHANGED.id
+        assert activity_log.user == self.user
+        assert activity_log.arguments == [self.addon, self.addon.slug, 'valid']
+
     def test_edit_as_developer(self):
         self.client.force_login(UserProfile.objects.get(email='regular@mozilla.com'))
         data = self.get_dict()
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -906,7 +906,10 @@ def addons_section(request, addon_id, addon, section, editable=False):
                 else:
                     ActivityLog.create(amo.LOG.EDIT_PROPERTIES, addon)
 
+                if valid_slug != addon.slug:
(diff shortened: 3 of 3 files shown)


GregSherman:
Is `olympia/addons/views.py::AddonVersionViewSet::update()` another area the slug is being changed? I see no tests related to that

GregSherman:
@diox requesting review

src/olympia/constants/activity.py
```diff
@@ -954,6 +954,14 @@ class CLEAR_ADMIN_REVIEW_THEME(_LOG):
     admin_event = True
 
 
+class ADDON_SLUG_CHANGED(_LOG):
+    id = 182
+    format = _('Addon {addon} slug changed from {old} to {new}.')
```
wagnerand, line 959: Should we include the user here?

diox (commented):
> Is `olympia/addons/views.py::AddonVersionViewSet::update()` another area the slug is being changed? I see no tests related to that

Yes, that should be changed too. Though the actual place you'll want to modify is the serializer, which is called by the view. It has an `update()` method as well. More importantly, it already has a `log()` method that you can add your code to.

The tests for this are all in `olympia/addons/tests/test_views.py`, see `TestAddonViewSetUpdate`.

src/olympia/constants/activity.py
diox: We should. Use `{user_responsible}` which is automatically added from the `user` we record on the `ActivityLog` when formatting (and that `user` itself is automatically added from current thread, if you don't explicitly set one when creating an `ActivityLog`)

eviljeff:
> Is `olympia/addons/views.py::AddonVersionViewSet::update()` another area the slug is being changed?

it's via `AddonViewSet` (`AddonVersionViewSet` is for versions), but the API data manipulation is handled in the serializers instead - `AddonSerializer`.

src/olympia/addons/tests/test_views.py
```diff
@@ -1996,6 +1996,14 @@ def test_set_slug_denied(self):
         )
         assert response.status_code == 200, response.content
 
+        # get the most recent ADDON_SLUG_CHANGED log entry
+        log_entry = ActivityLog.objects.filter(
+            action=amo.LOG.ADDON_SLUG_CHANGED.id
+        ).latest('pk')
+        assert log_entry.user == self.user
+        assert log_entry.arguments == [self.addon, 'denied-slug', 'denied-slug~']
+        assert 'slug from denied-slug to denied-slug~' in str(log_entry)
```
diox, line 2005: Let's move that to a completely separate test - we don't want to lose coverage of this if we ever refactor away the denied slug logic.

src/olympia/devhub/tests/test_views_edit.py
```diff
@@ -137,6 +137,12 @@ def test_edit_slug_valid(self):
         doc = pq(response.content)
         assert doc('form').attr('action') != old_edit
 
+        activity_log = ActivityLog.objects.latest('pk')
+        assert activity_log.action == amo.LOG.ADDON_SLUG_CHANGED.id
+        assert activity_log.user == self.user
+        assert activity_log.arguments == [self.addon, self.addon.slug, 'valid']
+        assert f'slug from {self.addon.slug} to valid' in str(activity_log)
```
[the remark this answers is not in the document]
eviljeff: (btw, just testing the end of the string was my suggestion - the string renders with html `<a>` tags around the user and add-on which made it messy to test. Could though if you think it's necessary @diox )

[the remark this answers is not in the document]
diox: Ah yeah. Maybe just add _another_ `assert ... in ...` then, ensuring we have the user representation (id + name).

diox (approved):

PR #21235, opened 2023-09-27
Make signing API create versions compatible with Firefox only (not Android)

### LINKED ISSUE #21229: Signing API should only create versions compatible with Firefox by default
STR:
- Submit an extension using `web-ext` (without `--use-submission-api`) or the signing API directly

Expected result:
- The extension is only compatible with Firefox, *not* Firefox for Android

Actual result:
- The extension is compatible with both Firefox and Firefox for Android

The signing API doesn't allow you to choose compatibility. To submit an extension compatible with Android, developers should add `browser_specific_settings` > `gecko_android` in their manifest.

Because the signing API has been behaving like this for years, this is a breaking change that cannot land until documentation and developer outreach has been done. Those tasks are tracked separately in JIRA. (**Edit:** [announcement](https://blog.mozilla.org/addons/2023/10/05/changes-to-android-extension-signing/) is now live)

diox:
Versions created with the signing API should only be compatible with Firefox for Android if the manifest contains `browser_specific_settings.gecko_android` key.

Also add a command to clean up any recently affected add-ons (using a csv).

Fixes #21229

### DIFF
--- a/src/olympia/devhub/tests/test_utils.py
+++ b/src/olympia/devhub/tests/test_utils.py
@@ -586,7 +586,7 @@ class TestCreateVersionForUpload(UploadMixin, TestCase):
             newer_upload,
             self.addon,
             amo.CHANNEL_LISTED,
-            selected_apps=[amo.FIREFOX.id, amo.ANDROID.id],
+            selected_apps=[amo.FIREFOX.id],
             parsed_data=self.mocks['parse_addon'].return_value,
         )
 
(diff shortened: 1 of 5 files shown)


diox:
https://blog.mozilla.org/addons/2023/10/05/changes-to-android-extension-signing/ is now live, we want to ship this in the October 12th push.

eviljeff (approved):

PR #21236, opened 2023-09-27
Handle TypeError in abuse report serializer for non-string values

### LINKED ISSUE #19286: The abuse report API throws a 500 error if some field values are not strings
Follow up for #19080

Send a request to the  [/api/v5/abuse/report/addon/](https://addons-server.readthedocs.io/en/latest/topics/api/abuse.html#post--api-v5-abuse-report-addon-) with any of the following fields set as: 
- `"addon": 621556` - value as `int` instead of `str`
- `"addon_install_source": {"value": "app:profile"}`
- `"addon_install_method": {"value": "rtamo"}`

### What happened?
A 500 server error

A 400 bad request error

shribyte:
Fixes #19286

This PR addresses an issue where the abuse report API would throw a TypeError when certain fields received unexpected data types. The solution involves wrapping the problematic code block with a try-except to handle the TypeError and raising a more specific ValidationError. Tests will be pushed in a subsequent commit.

The change has been successfully run and tested (on existing tests) locally.

### DIFF
--- a/src/olympia/abuse/serializers.py
+++ b/src/olympia/abuse/serializers.py
@@ -143,9 +143,9 @@ class AddonAbuseReportSerializer(BaseAbuseReportSerializer):
                 field_name,
                 str(data[field_name])[:255],
             )
-            raise serializers.ValidationError({
-                field_name: _('Invalid type for value submitted.')
-            })
+            raise serializers.ValidationError(
+                {field_name: _('Invalid type for value submitted.')}
+            )
 
         if is_value_unknown:
             log.warning(
--- a/src/olympia/abuse/serializers.py
+++ b/src/olympia/abuse/serializers.py
@@ -143,9 +143,7 @@ class AddonAbuseReportSerializer(BaseAbuseReportSerializer):
                 field_name,
                 str(data[field_name])[:255],
             )
-            raise serializers.ValidationError(
-                {field_name: _('Invalid value')}
-            )
+            raise serializers.ValidationError({field_name: _('Invalid value')})
 
         if is_value_unknown:
             log.warning(
--- a/src/olympia/abuse/serializers.py
+++ b/src/olympia/abuse/serializers.py
@@ -133,7 +133,21 @@ class AddonAbuseReportSerializer(BaseAbuseReportSerializer):
     def handle_unknown_install_method_or_source(self, data, field_name):
         reversed_choices = self.fields[field_name].reversed_choices
         value = data[field_name]
-        if value not in reversed_choices:
+
+        try:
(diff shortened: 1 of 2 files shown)


diox (changes requested):
Please include a test for this change.

It can be a unit test in `src/olympia/abuse/tests/test_serializers.py` (in `TestAddonAbuseReportSerializer` class) or an integration test in `src/olympia/abuse/tests/test_views.py` (in `AddonAbuseViewSetTestBase`, which already has tests around method/source so it's probably the easiest)

src/olympia/abuse/serializers.py
```diff
@@ -133,7 +133,21 @@ def validate_client_id(self, value):
...
+            is_value_unknown = value not in reversed_choices
+        except TypeError:
+            # Log the invalid type and raise a validation error.
+            log.warning(
+                'Invalid type for abuse report %s value submitted: %s',
+                field_name,
+                str(data[field_name])[:255],
+            )
+            raise serializers.ValidationError(
+                {field_name: _('Invalid type for value submitted.')}
```
diox, line 147: `'Invalid value'` is fine IMHO (we're already using that string elsewhere)

diox (commented):
This looks good, but as @eviljeff pointed out in the issue, if all the logic from `handle_unknown_install_method_or_source`() was moved into a `to_internal_value()` method on a custom class inheriting from `ReverseChoiceField` (that we would use instead of the normal `ReverseChoiceField` in the serializer when declaring `addon_install_method` and `addon_install_source` fields) we would be able to validate both separately, it would be slightly cleaner. Here because we're in the parent serializer, when the validation error is raised, we only return the first error (you can see that in the testcase you've added).

Bonus point, you could use `self.fail('invalid_choice', input=value)` instead of raising the exception and coming up with an error message yourself.

Let's get this PR merged and then you can do a follow-up separate PR for that refactor.

shribyte:
> This looks good, but as @eviljeff pointed out in the issue, if all the logic from `handle_unknown_install_method_or_source`() was moved into a `to_internal_value()` method on a custom class inheriting from `ReverseChoiceField` (that we would use instead of the normal `ReverseChoiceField` in the serializer when declaring `addon_install_method` and `addon_install_source` fields) we would be able to validate both separately, it would be slightly cleaner. Here because we're in the parent serializer, when the validation error is raised, we only return the first error (you can see that in the testcase you've added).
> 
> Bonus point, you could use `self.fail('invalid_choice', input=value)` instead of raising the exception and coming up with an error message yourself.

Got it, will split the refactoring into a future PR.

diox (approved):

PR #21244, opened 2023-10-02
Harmonize footers across devhub (and frontend pages a bit more)

### LINKED ISSUE #18215: src/olympia/templates/photon-footer.html is not localized
This is a follow-up to https://github.com/mozilla/addons-server/issues/18207.

When working on that issue, @eviljeff noted that none of the strings in the footer template are localized. That seems unintentional and not a good thing.

We should fix it.

### LINKED ISSUE #21243: Harmonize footers
We have different footers content & design between AMO frontend, AMO devhub landing pages, and the rest of addons-server pages (like the rest of devhub). This is inconsistent and confusing, let's fix that.

diox:
Changes:
- Copy over fonts from addons-frontend, use Inter everywhere
- Use photon-footer everywhere, adding a light blue margin before to make the transition less aggressive and match header
- Fix error pages block to align content correctly
- Translate strings in footer

Fixes #18215
Fixes #21243


diox:
Devhub:
Screenshot 2023-10-02 at 10-44-30 Manage My Submissions Developer Hub Add-ons for Firefox

Error pages:
Screenshot 2023-10-02 at 10-45-03 Oops
Screenshot 2023-10-02 at 10-44-50 Not Found

Reviewer tools:
Screenshot 2023-10-02 at 10-45-42 Pending Rejection - Add-ons for Firefox

diox:
This makes any future pages living in addons-server more consistent with the rest of the site, particularly if they are not devhub-related.

static/css/legacy/main.css
```diff
@@ -942,7 +937,7 @@ form.favorite {
     max-width: 90em;
     min-width: 960px;
     position: relative;
-    min-height: 55em;
+    min-height: 768px; /* arbitrary high number to force the footer down */
```
eviljeff, line 940: mmm?  How would this be affected if the user had a screen height bigger than 768px (and etc)

eviljeff (approved):

static/css/legacy/main.css
diox: FWIW `55em`, the previous value, is already arbitrary and once computed without messing with the font size would be roughly equivalent to `715px`. I just increased it a bit so that it matches the footer better visually (since that new footer is longer in size).

With that new value, a typical page would be 768+65+1+606 pixels long with the default font size on my machine. 

If your window height is bigger than that, it would display the default white background after the footer (I didn't want to mess with that style). Again not a new issue, just looks a bit weirder now that the footer background is black-ish.

eviljeff: oh, I read that as `55px` so it seemed like a big increase.  Changing the page background to black would probably address that but I guess it doesn't matter.

diox: I added a background to `html` to match the footer, but kept the `min-height` because I think it's better visually. 

Here is what it would look like without the `min-height`: Screenshot 2023-10-02 at 15-54-21 Not allowed

(Without the matching background color you'd also see white instead of black after the footer's content if the `min-height` is absent or not enough compared to the window height)

PR #21260, opened 2023-10-03
add new fields to AbuseReport model to support DSA/Feedback flow

### LINKED ISSUE #21255: Make model changes to AbuseReport to handle cinder/DSA requirements
We will be repurposing the existing `AbuseReport` model to also track feedback and and content violation reports we will be handing to satisfy our DSA requirements.

Broadly we need to:
 - add decision category (disable/ban/etc)
 - probably* add the policy name and/or uuid the decision was made under
 - add reporter email and name (the user may be authenticated also)
 - change categories ... (personally) I lean towards entirely new categories and then we can translate the old categories in the serializer where appropriate while we're still getting the old reports
 - appeal state is tricky - I suppose if we're limited to just one appeal each per reporter and target then a flag on the model is sufficient. Otherwise we'd need an intermediate model.

*there is a bunch of extra metadata sent with the webhook response, but unless we _need_ it I guess we don't need to store it in the AMO DB.

eviljeff:
fixes #21255 ?  -ish, anway. ~It doesn't add the decision model the issue talks about, and~ the categories are still TBC so I just added a few?


eviljeff:
I've no idea what the codestyle fail is about - it doesn't list any files and it succeeds locally (seems to be black)

src/olympia/abuse/models.py
```diff
@@ -74,6 +75,17 @@ class AbuseReport(ModelBase):
         # previous one. We avoid re-using the value.
         ('UNWANTED', 9, "Wasn't wanted / impossible to get rid of"),
         # `10` was previously "Other". We avoid re-using the value.
+
```
diox, line 78: Black complains about this extra newline here

eviljeff: Weird - it doesn't locally for me!

src/olympia/abuse/models.py
```diff
@@ -171,6 +186,8 @@ class AbuseReport(ModelBase):
     state = models.PositiveSmallIntegerField(
         default=STATES.UNTRIAGED, choices=STATES.choices
     )
+    # if this report has been submitted to cinder this is the job reference there
+    cinder_job_id = models.CharField(max_length=36, default=None, null=True)
```
diox, line 189: Not sure this should live in this model. In the issue I mentioned adding a `AbuseReportDecision` but maybe this should be `AbuseReportCinder` or whatever and it would hold everything related to Cinder ?

eviljeff: https://github.com/mozilla/addons-server/pull/21260/commits/07227894165ae930a0dcb2433c7e1b90e8af71c3 adds a `CinderReport` model.

src/olympia/abuse/models.py
```diff
@@ -279,3 +293,23 @@ def type(self):
...
+        ('AMO_DISABLE_ADDON', 2, 'Add-on disable'),
+        ('AMO_ESCALATE_ADDON', 3, 'Escalate add-on to reviewers'),
+        ('AMO_ESCALATE_USER', 4, 'Escalate add-on to reviewers'),
+        ('AMO_DELETE_RATING', 5, 'Rating delete'),
+        ('AMO_DELETE_COLLECTION', 6, 'Collection delete'),
+        ('AMO_APPROVE', 7, 'Approved (no action)'),
+    )
+
+    job_id = models.CharField(max_length=36)
+    abuse_report = models.ForeignKey(AbuseReport, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
eviljeff: https://github.com/mozilla/addons-server/pull/21260/commits/3a1a2b8d7ae5faf7c48b9185f948086b7c7bd3d6

diox (approved):

PR #21303, opened 2023-10-11
Update serializers.py

eviljeff:
Needed fix for the frontend patch (it was an advertent change when I specified `EmailField` for `reporter_email` - when the field was generated from the model it would have allowed null because the underlying model field does)


diox:
For a second I was worried this was affecting regular abuse reports as well... I get it now but it needs a test.

eviljeff:
> For a second I was worried this was affecting regular abuse reports as well... I get it now but it needs a test.

it doesn't affect current frontend (or Firefox) because the field is just omitted, but in the new frontend it's easier to send a value, but null, when name/email/reason is not available (e.g. when the existing button is used).

diox (approved):

Issue #21268, opened 2023-10-03
Make API changes to abuse api to handle cinder/DSA requirements


eviljeff:
Depends on the model changes in https://github.com/mozilla/addons-server/issues/21255

We will be repurposing the existing /abuse api endpoint to also track feedback and and content violation reports we will be handing to satisfy our DSA requirements.

We need to:
- add fields for name and email
- call `CinderAddon.report` for appropriate (DSA covered) reasons
- `Cinder` class will need updating to handle non-authenticated users with name and email
- create a `CinderReport` instance with the job id returned from the API

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-512

PR #21316, opened 2023-10-17
add/change abuse report reason for current dsa proposal

eviljeff:
follow-up for #21268 (I didn't file a separate issue because I'm assuming everything will be QA'd together, but could?)


wagnerand (changes requested):
Looks overall, a couple of suggestions to be more accurate. Thanks!

wagnerand (approved):
r+wc

PR #21363, opened 2023-10-30
Fix typo and locale

diox:


locale/zh_TW/LC_MESSAGES/django.po
```diff
@@ -4598,7 +4598,7 @@ msgstr "Return To AMO GUID 無效（base64 網址格式不正確？）"
 
 #, python-format
 msgid "Invalid combination of %s\" and \"%s\" parameters."
-msgstr "「％s」和「％s」的參數組合無效。"
+msgstr "「%s」和「%s」的參數組合無效。"
```
diox, line 4601: (that string is now an orphaned one that will go away next extraction but I chose to fix it rather than remove it in case it would be re-added again before we extract)

eviljeff (approved):

PR #21364, opened 2023-10-30
Allow Abuse reports to be sent against Rating

diox:
Includes refactors to share some logic between different kinds of abuse reports

Fixes mozilla/addons#9324

### DIFF
--- a/src/olympia/urls.py
+++ b/src/olympia/urls.py
@@ -29,8 +29,6 @@ urlpatterns = [
     # Home.
     re_path(r'^$', frontend_view, name='home'),
     # Abuse.
-    # FIXME: modify nginx config to route /<locale>/abuse/ to addons-server
-    # This will need to not clash with new frontend reporting pages !
     re_path(r'abuse/', include('olympia.abuse.urls')),
     # Add-ons.
     re_path(r'', include('olympia.addons.urls')),
@@ -109,7 +107,7 @@ urlpatterns = [
     ),
 ]
 
-if settings.DEBUG:
+if settings.DEBUG and 'debug_toolbar' in settings.INSTALLED_APPS:
     import debug_toolbar
 
     # Remove leading and trailing slashes so the regex matches.


diox:
Not sure where the test failure is coming from `FAILED src/olympia/accounts/tests/test_views.py::TestLoginStartView::test_redirect_url_fake_fxa_auth - ModuleNotFoundError: No module named 'debug_toolbar'` I guess some of the imports changes caused a ripple effect that triggered this ? I'll look into it, but in the meantime this should be ready for review.

src/olympia/abuse/models.py
```diff
@@ -316,16 +331,34 @@ def delete(self, *args, **kwargs):
...
+        creation = kwargs.get('force_insert', False) or bool(self.pk)
+        rval = super().save(*args, **kwargs)
+        if (
+            creation
+            and waffle.switch_is_active('enable-cinder-reporting')
+            and self.reason in AbuseReport.REASONS.REPORTABLE_REASONS
+        ):
+            # call task to fire off cinder report
+            report_to_cinder.delay(self.pk)
+        return rval
```
eviljeff, line 346: I'm not sure about this change - it'll be reporting to cinder automatically ever time an instance is created which seems extreme (even in tests!).  What was the problem in firing the task in the serializer?

docs/topics/api/abuse.rst
```diff
@@ -251,3 +255,33 @@ so reports can be responded to if necessary.
...
+
+.. http:post:: /api/v5/abuse/report/rating/
+
+    .. _ratingabusereport-create-request:
+
+    :<json string rating: The id of the rating to report for abuse (required).
+    :<json string message: The body/content of the abuse report (required).
+    :<json string|null reporter_name: The provided name of the reporter, if not authenticated.
+    :<json string|null reporter_email: The provided email of the reporter, if not authenticated.
+    :>json object|null reporter: The user who submitted the report, if authenticated.
```
eviljeff, line 278: `reason` should be in here (and the response) too?

src/olympia/ratings/migrations/0009_alter_deniedratingword_word.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.6 on 2023-10-27 10:17
+
+from django.db import migrations, models
+import olympia.ratings.models
+
+
```
eviljeff, line 6: What triggered this migration?

src/olympia/abuse/models.py
diox: At the minimum I needed to move that out of the addon-specific serializer, so I thought I'd do this at the model level while I'm at it, which is cleaner - ultimately we shouldn't care that it's an API request, we have to do this for every abuse report.

I should address tests, but also we are likely going to need something along those lines when we implement https://mozilla-hub.atlassian.net/browse/AMO-116 (though I'm not sure we'll have `AbuseReport` instances for that)

src/olympia/ratings/migrations/0009_alter_deniedratingword_word.py
diox: Catching up on `ratings` migrations. I don't have to include it, but it's a no-op.

src/olympia/abuse/models.py
eviljeff: IMO the model save just seems too low level to trigger a request to a remote service that creates new records.  If we in the future (for whatever reason) create an AbuseReport instance to backfill anything, or in the admin tools that would also trigger a (hidden) cinder request.

https://mozilla-hub.atlassian.net/browse/AMO-116 I expect would need special handling, because we're going to want to immediately resolve the job after it's created.

eviljeff: we could move all the logic to a utility function if that's cleaner?

diox: I'll go back to a serializer implementation for now, I don't think a utility function helps much.

eviljeff (approved):

PR #21382, opened 2023-11-01
replace force disable/enable buttons with actions

### LINKED ISSUE #21356: DSA: Move force-disable button into reviewer action
The existing force-disable button for add-ons needs to be moved into the reviewer actions, so that a reviewer can select a reason and a write a response or use a canned one.
This means that for the force-disable action, the reason and the response are mandatory fields.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-544)

eviljeff:
fixes #21356


src/olympia/reviewers/templates/reviewers/emails/force_disable_addon.ltxt
```diff
@@ -0,0 +1,13 @@
...
+Please see the reviewer's comments below for more information.
+
+********
+Details:
+{{ comments }}
+********
+
+Thank you for your attention.
+
+{% endblock %}
```
eviljeff, line 13: @wagnerand is this email copy okay?  I took one of the existing templates for rejections and slightly modified it.

wagnerand: I forwarded it to Ashley and Lindsay

src/olympia/reviewers/views.py
```diff
@@ -1116,26 +1116,6 @@ def unsubscribe_unlisted(self, request, **kwargs):
         ).delete()
         return Response(status=status.HTTP_202_ACCEPTED)
 
-    @drf_action(
```
diox, line 1119: docs need to be updated

src/olympia/reviewers/utils.py
```diff
@@ -799,6 +800,34 @@ def get_actions(self):
             'comments': False,
             'available': is_appropriate_admin_reviewer and is_static_theme,
         }
+        actions['enable_addon'] = {
```
diox, line 803: Shouldn't we have `'comments': False,` for this action ?

src/olympia/addons/models.py
```diff
@@ -637,10 +637,12 @@ def clean_slug(self, slug_field='slug'):
...
         clean_slug(self, slug_field)
 
-    def force_disable(self):
+    def force_disable(self, user=None):
         from olympia.reviewers.models import NeedsHumanReview
 
-        activity.log_create(amo.LOG.FORCE_DISABLE, self)
+        activity.log_create(
+            amo.LOG.FORCE_DISABLE, self, **({'user': user} if user else {})
+        )
```
diox, line 645: This is inconsistent with other reviewer actions, we should be recording the `reason` and `details` (which includes `comments`) in that activity.

src/olympia/addons/models.py
```diff
@@ -654,8 +656,10 @@ def force_disable(self):
         # https://github.com/mozilla/addons-server/issues/13194
         self.disable_all_files()
 
-    def force_enable(self):
-        activity.log_create(amo.LOG.FORCE_ENABLE, self)
+    def force_enable(self, user=None):
+        activity.log_create(
+            amo.LOG.FORCE_ENABLE, self, **({'user': user} if user else {})
+        )
```
diox, line 662: As above, we're missing at least `details` in there, which would record the comment if any.

diox (changes requested):
In addition to my comments... `FORCE_DISABLE` and `FORCE_ENABLE` are both currently set to `hide_developer=True`, which means the developer won't have any trace of them in their activity feed, which is inconsistent with other reviewer actions...

I think we should change this, but it has some potentially unintended consequences, we would need to stop showing `{user_responsible}` in the `format` string for starters.

src/olympia/addons/models.py
eviljeff: I resolved both of these comments by skipping the logging in the Addon function, and relying on the reviewer tools `log_action`

src/olympia/reviewers/templates/reviewers/emails/force_disable_addon.ltxt
eviljeff: From the slack thread it seemed like no changes were needed so resolving

docs/topics/api/overview.rst
```diff
@@ -1,474 +1,475 @@
...
-* 2023-07-06: added ``is_all_versions`` to blocklist block endpoint. https://github.com/mozilla/addons-server/issues/20857
-* 2023-10-12: added ``reporter_name`` and ``reporter_email`` as two optional alternatives to an authenticated reporter in the abuse api. https://github.com/mozilla/addons-server/issues/21268
-* 2023-10-26: added ``location`` to abuse api. https://github.com/mozilla/addons-server/issues/21330
-* 2023-11-02: removed ``application`` from categories endpoint, flattened ``categories`` in addon detail/search endpoint. https://github.com/mozilla/addons-server/issues/5989
-
-.. _`#11380`: https://github.com/mozilla/addons-server/issues/11380/
-.. _`#11379`: https://github.com/mozilla/addons-server/issues/11379/
-.. _`#11378`: https://github.com/mozilla/addons-server/issues/11378/
-.. _`#11374`: https://github.com/mozilla/addons-server/issues/11374/
+.. _api-overview:
```
diox, line 1: Looks like you've added Windows-style line ending (\r\n) instead of Unix ones (\n) in that doc ?

src/olympia/constants/activity.py
```diff
@@ -814,20 +814,20 @@ class VERSION_RESIGNED(_LOG):
 class FORCE_DISABLE(_LOG):
     id = 167
     keep = True
-    # We don't want to notify developers, this is not a regular rejection - the
-    # add-on is likely malicious.
-    hide_developer = True
     reviewer_review_action = True
-    format = '{addon} force-disabled by {user_responsible}.'
+    format = '{addon} force-disabled'
```
diox, line 818: `format` now needs to be translated for both of these actions (not needed for `reviewer_format` nor `admin_format`)

src/olympia/reviewers/utils.py
```diff
@@ -799,6 +800,34 @@ def get_actions(self):
...
+                addon_is_not_deleted
+                and not addon_is_not_disabled
+                and is_appropriate_admin_reviewer
+            ),
+        }
+        actions['disable_addon'] = {
+            'method': self.handler.disable_addon,
+            'label': 'Force disable',
+            'details': (
+                "This will force disable this add-on, and all it's versions. "
```
diox, line 821:
```suggestion
                "This will force disable this add-on, and all its versions. "
```

?

diox (approved):

docs/topics/api/overview.rst
eviljeff: Gah! git set up problems on the loaner I'm using

Issue #21375, opened 2023-10-31
Put add-ons in the queue when reported for Policy Violation in the add-on


diox:
When an Abuse Report is created for `POLICY_VIOLATION`, and the location reported is `ADDON` or `BOTH`, it needs to be put in the reviewer queue through `NeedsHumanReview`, with a specific reason explaining to the reviewer an abuse report was filed against it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-549)

eviljeff:
Note: we'll need a different queue for reports that we want to log in Cinder but we don't expect T&S/TaskUs to handle (i.e. reports we'll handle in AMO reviewer tools)

ioanarusiczki:
@diox 

I tried with "it violates the law ... " and location Inside the add-on or Both locations but I don't see them flagged for human review in the Manual Review queue. Tried with "It violates Add-on policies" too , same result.

diox:
@ioanarusiczki sorry, forgot to add: you need to turn on the `enable-cinder-reporting` waffle switch first (it does more behind the scenes that we're not ready to let you verify, but it should unblock testing of this particular issue)

ioanarusiczki:
@diox 

I tried again after enabling the waffle switch but it's the same result. Or should I see these reports elsewhere? I got access to stage.cinder.nonprod 

I see there some queues but they're empty.

eviljeff:
> @diox
> 
> I tried again after enabling the waffle switch but it's the same result. Or should I see these reports elsewhere? I got access to stage.cinder.nonprod
> 
> I see there some queues but they're empty.

The cinder integration should _just now_ be working.  But it seemsyYou've hit a bug with authenticated reports - https://mozilla.sentry.io/issues/4619894006/

eviljeff:
> > @diox
> > I tried again after enabling the waffle switch but it's the same result. Or should I see these reports elsewhere? I got access to stage.cinder.nonprod
> > I see there some queues but they're empty.
> 
> The cinder integration should _just now_ be working. But it seems you've hit a bug with authenticated reports - https://mozilla.sentry.io/issues/4619894006/

https://github.com/mozilla/addons-server/issues/21421

ioanarusiczki:
@eviljeff 

[...]

PR #21376, opened 2023-11-01
consume webhook decisions

eviljeff:
fixes #21194

includes the basic functionality for processing add-on and user decisions for proof of concept; the full implementation will be in #21352 and #21355 (and notifications in subsequent tickets)

Issue #21234, opened 2023-09-27
Send all relevant metadata for all content types to Cinder when we report


eviljeff:
See https://docs.google.com/document/d/1k1iyC1TMdyaTsOGlJLJpDyXr4l03O9bGemOSHrbKbnE/edit#heading=h.dca2g7q1cvvo for what metadata needs to be sent with each content type.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-506)

wagnerand:
@eviljeff I chatted with Lindsay about this, she'll include that in her proposal.

eviljeff:
Should we also include the version for add-ons, if available?

diox:
Note: the question about whether or not to send all translations also applies to collections, which potentially have name & description in several locales

ioanarusiczki:
@diox 

I checked on -dev for extensions and themes so far. For both I don't think I've noticed the Author in Cinder. 

Both have also ID, slug and guid in Cinder (the doc does not mention those but I think these should be present)

Themes don't have release notes or theme icon - there is only a preview. (the doc mentions them both as required)

ioanarusiczki:
@diox 

- I found details about reported/reporter in Relashionships

- ratings/developer replies, I see for both type: amo rating

- collections: I did not find collection items, name, link, author, summary or item rating. Docs metion a badge but AMO collections don't have badges

- user profiles: I did not see biography field in Cinder,  I reported id 10641572. Average ratings in Cinder: 3. Average ratings on amo: 3.8. Could not check for other users after it because the reports did not arrive in Cinder , I'll re-check tomorrow.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-506

PR #21385, opened 2023-11-01
flag abuse reports reviewers will handle; send to seperate queue

eviljeff:
WIP for #21375 (sans tests 😬)


src/olympia/abuse/cinder.py
```diff
@@ -171,3 +172,26 @@ def get_context(self):
...
+
+    def flag_for_human_review(self, appeal=False):
+        from olympia.reviewers.models import NeedsHumanReview
+
+        reason = (
+            NeedsHumanReview.REASON_ABUSE_ADDON_VIOLATION_APPEAL
+            if appeal
+            else NeedsHumanReview.REASON_ABUSE_ADDON_VIOLATION
+        )
+        self.addon.set_needs_human_review_on_latest_versions(reason=reason)
```
diox, line 189: Like https://github.com/mozilla/addons-server/pull/21376, this could be the wrong versions...

src/olympia/reviewers/models.py
```diff
@@ -755,6 +755,8 @@ class NeedsHumanReview(ModelBase):
     REASON_MANUALLY_SET_BY_REVIEWER = 7
     REASON_AUTO_APPROVED_PAST_APPROVAL_DELAY = 8
     REASON_ABUSE_REPORTS_THRESHOLD = 9
+    REASON_ABUSE_ADDON_VIOLATION = 11
+    REASON_ABUSE_ADDON_VIOLATION_APPEAL = 12
```
diox, line 759: Like https://github.com/mozilla/addons-server/pull/21376, those need to be added to the `choices` below. (Someday we should use `extended-choices` like we do in `AbuseReport` to get all of this for free here too...)

src/olympia/abuse/utils.py
```diff
@@ -61,9 +61,13 @@ def process(self):
...
             if version_obj:
-                NeedsHumanReview.objects.create(version=version_obj, reason=reason)
+                NeedsHumanReview.objects.create(
+                    version=version_obj, reason=reason, is_active=True
+                )
             else:
-                addon.set_needs_human_review_on_latest_versions(reason=reason)
+                addon.set_needs_human_review_on_latest_versions(
+                    reason=reason, ignore_reviewed=False, unique_reason=True
+                )
```
eviljeff, line 70: Included the same fix for the escalation action too.

src/olympia/abuse/cinder.py
```diff
@@ -151,8 +151,9 @@ def appeal(self, *args, **kwargs):
 class CinderAddon(CinderEntity):
     type = 'amo_addon'
 
-    def __init__(self, addon):
+    def __init__(self, addon, version=None):
```
eviljeff, line 154: Note, we don't do anything with this version in general, but we should (very likely) send it - something to consider in https://github.com/mozilla/addons-server/issues/21234

src/olympia/abuse/cinder.py
```diff
@@ -202,3 +204,33 @@ def get_context(self):
                 cinder_user.get_relationship_data(self, 'amo_rating_author_of')
             ],
         }
+
+
+class CinderAddonByReviewers(CinderAddon):
```
diox, line 209: nit: Ignore if you think that's not necessary/consistent with what you had in mind

```suggestion
class CinderAddonHandledByReviewers(CinderAddon):
```

Or something like that ?

diox (approved):

src/olympia/abuse/cinder.py
eviljeff: hadn't really thought that deeply about the naming, so sure

Issue #21282, opened 2023-10-05
Don't associate reporter IP with rating


wagnerand:
It seems in some cases the IP address of an account flagging a rating could be associated with that rating. We shouldn't do that.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-516)

wagnerand:
@diox mentioned there might be another (very fringe) scenario: "if you go as an admin to the rating detail page and click save". We don't do that since this page has no editable information (everything is read-only), but to be safe, we should hide the save buttons or in other words make the view read-only.

diox:
Why this is happening: we record the `ActivityLog` in the `post_save()` method of the `Rating` model. While this is useful to have models as the place where the business logic lives, this is annoying here because we record a `EDIT_RATING` activity everytime we save an instance... And we happen to do that when a rating is flagged, to set `editorreview`.

For flagging, a simple workaround, that is already used elsewhere to update other denormalized fields, would be to pass `_signal=False` to the `update()` call that sets `editorreview=True` when flagging. 

In addition, we need to make the admin page read-only: although the admin doesn't let you modify any fields, it is still technically a `change` page instead of a `view` page, so it has `save` buttons, which would trigger an `EDIT_RATING` activity when used.

A more involved fix would be figuring out a cleaner way to only record `EDIT_RATING` when there is actually an edit made by a user, as opposed to other kind of modifications that are more internal. Maybe we could use `OnChangeMixin` or similar and look at what fields were changed.

wagnerand:
There is also a way on the admin rating listing page that allows multiple ratings to be deleted. I assume this is handled correctly already as well, but would be good to double check.

diox:
> There is also a way on the admin rating list page that allows multiple ratings to be deleted. I assume this is handled correctly already as well, but would be good to double check.

Yes, that should be handled correctly. It goes through the deletion path that sets `deleted=True` and is ignored by `post_save()`.

GregSherman:
commenting for assignment @diox

ioanarusiczki:
@GregSherman 

There is no activity about flagging a rating in Activity Log on -stage.  I can see written and updated review, written and updated reply.

@diox Activity log is loading slow (known already).

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-516

PR #21389, opened 2023-11-03
Remove edit logs for rating flags

GregSherman:
Fixes issue #21282 

Removes the change permission for admins on the ratings page.

Adds `_signal=False` to the flag view to not call `post_save()`


diox:
We should also add a data migration containing `ActivityLog.objects.filter(action=amo.LOG.EDIT_RATING.id).exclude(ratinglog__rating__user=F('user')).delete()` to delete the bogus activity.

diox (approved):

PR #21391, opened 2023-11-03
Allow reports against users with a reason

### LINKED ISSUE #21386: Call Cinder for Abuse Reports filed against users
It's already possible to file abuse reports against users and some Cinder integration will already be there automatically following refactors from https://github.com/mozilla/addons-server/issues/21349, but it needs to be checked and completed as necessary.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-551)

diox:
Fixes #21386

### DIFF
--- a/docs/topics/api/abuse.rst
+++ b/docs/topics/api/abuse.rst
@@ -235,7 +235,7 @@ Submitting a user abuse report
 .. _`userabusereport-create`:
 
 The following API endpoint allows an abuse report to be submitted for a user account
-on https://addons.mozilla.org.  Authentication is not required, but is recommended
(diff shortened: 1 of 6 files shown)


eviljeff (approved):

willdurand:
@diox I think `feedback_spam` should also be supported here

diox:
@willdurand Ugh, yeah, that's what I get for trying to share code between them... `Ratings` doesn't have that because of how it's supposed to be handled separately, so it's missing from the `reason` choices that are shared between `UserProfile` and `Ratings` reports. I'll fix it.

PR #21408, opened 2023-11-08
support webhook addon decisions

### LINKED ISSUE #21352: Handle decisions from Cinder about extensions
In the webhook for Cinder (https://github.com/mozilla/addons-server/issues/21194) we need to handle decisions about extensions.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-540)

eviljeff:
fixes #21352 - most of the patch ended up being about having a way to accurate record the file status before it was changed to disabled, and why it was changed.


src/olympia/files/models.py
```diff
@@ -76,6 +77,13 @@ def files_storage():
 class File(OnChangeMixin, ModelBase):
     id = PositiveAutoField(primary_key=True)
     STATUS_CHOICES = amo.STATUS_CHOICES_FILE
+    STATUS_DISABLED_REASONS = Choices(
+        ('NONE', 0, 'None'),
+        ('DEVELOPER', 1, 'Developer disabled'),
+        ('ADDON_DISABLE', 2, 'Add-on disabled'),
+        ('ADDON_DELETE', 3, 'Add-on deleted'),
+        ('VERSION_DELETE', 4, 'Version deleted'),
```
eviljeff, line 85: we don't really need to know about deletes for this patch, but it seemed consistent to record all the ways an approved or awaiting review file could change to disabled.  
(At one point I was recording _all_ the reasons a file would be STATUS_DISABLED - rejection and obsolete too - but it made the patch overly complicated and didn't add any functionality we actually need)

src/olympia/users/models.py
```diff
@@ -498,10 +497,6 @@ def ban_and_disable_related_content_bulk(cls, users, move_files=False):
         addons_sole = Addon.unfiltered.filter(id__in=addon_ids - addon_joint_ids)
         # set the status to disabled - using the manager update() method
         addons_sole.update(status=amo.STATUS_DISABLED)
-        # collect Files that need to be disabled now the addons are disabled
-        File.objects.filter(version__addon__in=addons_sole).update(
-            status=amo.STATUS_DISABLED
-        )
```
eviljeff, line 504: This was obsolete - they get disabled anyway

diox: Wasn't it done here to force doing it by bulk? Doing it here in a single query is a lot more efficient if the user has a lot of add-ons.

src/olympia/files/migrations/0031_file_status_disabled_reason.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('files', '0030_remove-add-guid-to-manifest-waffle-switch'),
+    ]
+
+    operations = [
+        migrations.AddField(
```
diox, line 13: To add to push notes: that is going to be an expensive migration, and since it touches `File` we'll need to disconnect the update service while it's running.

diox:
Note: This will need careful testing by QA to ensure developers can't re-enable something they shouldn't be able to, and making sure the reason is properly recorded.

src/olympia/users/models.py
eviljeff: I can't remember why I did it now 🙃 but sounds plausible.

eviljeff: https://github.com/mozilla/addons-server/pull/21408/commits/86b2560befafcb336a2d410758a7898fd67c0ef6

diox (approved):

PR #21411, opened 2023-11-08
return 200 for an unhandled webhook decison, with the reason

### LINKED ISSUE #21402: return distinct response codes with cinder webhook
Currently we return 201 in the response for all webhook requests.  After discussing with ashwin on slack we will consistently return 201 for a request we've handled, and 200 for other requests.  If the request is malformed - e.g. multiple actions, which we can't handle, we will return an error code.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-556)

eviljeff:
fixes #21402 and notably also fixes a bug in #21376 - it wasn't being converted from the api value to the int value, and tests didn't pick it up because the calls were mocked 😐

### DIFF
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -1,5 +1,6 @@
 from datetime import datetime, timedelta
 
+from django.core.exceptions import ValidationError
 from django.db import models
 
 from extended_choices import Choices
@@ -475,10 +476,18 @@ class CinderReport(ModelBase):
         self.update(job_id=job_id)
 
     def process_decision(self, *, decision_id, decision_date, decision_actions):
+        if (
+            not decision_actions
+            or len(decision_actions) > 1
+            or not self.DECISION_ACTIONS.has_api_value(decision_actions[0])
+        ):
+            raise ValidationError('decision_actions malformed')
         self.update(
             decision_id=decision_id,
             decision_date=decision_date,
-            **({'decision_action': decision_actions[0]} if decision_actions else {}),
+            decision_action=self.DECISION_ACTIONS.for_api_value(
+                decision_actions[0]
+            ).value,
         )
         self.get_action_helper().process()
 
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -1,6 +1,5 @@
 from datetime import datetime, timedelta
 
-from django.core.exceptions import ValidationError
 from django.db import models
 
 from extended_choices import Choices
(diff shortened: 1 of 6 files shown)


src/olympia/abuse/models.py
```diff
@@ -475,10 +476,18 @@ def report(self):
         self.update(job_id=job_id)
 
     def process_decision(self, *, decision_id, decision_date, decision_actions):
+        if (
+            not decision_actions
+            or len(decision_actions) > 1
+            or not self.DECISION_ACTIONS.has_api_value(decision_actions[0])
```
diox, line 482: IMHO this method shouldn't be dealing with api values, that should be done by the view, much like we process the timestamp before passing `decision_date`. Let the view worry about API stuff, and when it comes to doing the processing we should be dealing with values that are already valid choices for each field only.

eviljeff: https://github.com/mozilla/addons-server/pull/21411/commits/0c2b29e5ad53546681d6ef6f9f3004f9066a1f00
(and refactored the view to be flatter with exceptions raised to escape because the level of indentation was getting a little OTT)

src/olympia/abuse/views.py
```diff
@@ -167,34 +171,62 @@ def process_datestamp(date_string):
...
+        cinder_report.process_decision(
+            decision_id=source.get('decision', {}).get('id'),
+            decision_date=process_datestamp(payload.get('timestamp')),
+            decision_action=CinderReport.DECISION_ACTIONS.for_api_value(
+                actions[0]
+            ).value,
         )
 
-    return Response(data={'amo-received': True}, status=status.HTTP_201_CREATED)
+    except ValidationError as exc:
```
diox, line 215: mmm, I would have expected us to keep returning 400s for errors in requests meant for us, and just return 200s when it's for another queue.

eviljeff: > If the receiving server returns an error response, Cinder will continue retrying to send that webhook indefinitely (with expotential backoff). That way, even if your endpoint is down, the webhooks will come in once you're back online. Cinder expects a server response within 10 seconds, otherwise the request will timeout and be retried.
>
>HTTP response codes between 400 and 599 are all considered errors.

I'm reading that as if we return an error code cinder will just retry again which doesn't seem helpful.

diox: Ugh ok. Could you add a comment in that `except` block explaining that (maybe more succinctly) ?

diox (approved):

eviljeff:
> ## Suspect Issues
> 
> This pull request was deployed and Sentry observed the following issues:
> 
>     * ‼️ **Cinder webhook request failed: bad enforcement_actions [["amo-disable-addon","delete-status","fre...** `/api/v5/abuse/response/` [View Issue](https://mozilla.sentry.io/issues/4630508892/?referrer=github-pr-bot)

This is because we're raising each time we encounter any unknown actions.  The structure of the policies/actions is still up in the air but, imo, we need to change to only raise if we get a decision where we don't support _any_ of the actions (and probably if we support more than one of them)

eviljeff:
> > ## Suspect Issues
> > This pull request was deployed and Sentry observed the following issues:
> > ```
> > * ‼️ **Cinder webhook request failed: bad enforcement_actions [["amo-disable-addon","delete-status","fre...** `/api/v5/abuse/response/` [View Issue](https://mozilla.sentry.io/issues/4630508892/?referrer=github-pr-bot)
> > ```
> 
> This is because we're raising each time we encounter any unknown actions. The structure of the policies/actions is still up in the air but, imo, we need to change to only raise if we get a decision where we don't support _any_ of the actions (and probably if we support more than one of them)

Filed https://github.com/mozilla/addons-server/issues/21436

PR #21417, opened 2023-11-09
Allow Abuse Reports to be filed against Collections

### LINKED ISSUE #21361: Allow Abuse Reports to be filed against Collections
Abuse Reports against Collections need to be possible and be passed to Cinder.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-545)

diox:
Fixes #21361


src/olympia/bandwagon/views.py
```diff
@@ -76,10 +76,16 @@ def get_serializer_class(self):
             CollectionSerializer if not with_addons else CollectionWithAddonsSerializer
         )
 
+    def get_object(self):
+        if not hasattr(self, 'instance') and 'pk' in self.kwargs:
+            self.lookup_field = 'pk'
+        return super().get_object()
```
eviljeff, line 82: Is this exposing extra functionality to the collection api endpoint itself?  iirc it was a deliberate design to only have collections accessible by slug.

diox: No, it's not exposed: When generating the URL for `CollectionViewSet`, we use `NestedSimpleRouter` with `lookup='user'` (so `kwargs['user_pk']` is passed) + `.register()` (so `kwargs[CollectionViewSet.lookup_field]` is passed - `lookup_field` is set to `slug` by default).

That being said, I can make it safer - I can pass a custom `action` and only do the `lookup_field = 'pk'` bit if we are using that action - making it truly impossible to use that path even if we were to accidentally pass `'pk'` in `self.kwargs`.

eviljeff: I don't want to add a whole bunch of extra complexity for a scenario that can't happen... but it was unexpected that we'd be adding another way to lookup instances to CollectionViewset, for abuse reports.  Up to you 🤷‍♂️

src/olympia/abuse/cinder.py
```diff
@@ -206,6 +206,36 @@ def get_context(self):
...
+            'description': str(self.collection.description),
+            'comments': self.collection.get_all_comments(),
+        }
+
+    def get_context(self):
+        cinder_user = CinderUser(self.collection.author)
+        return {
+            'entities': [cinder_user.get_entity_data()],
+            'relationships': [
+                cinder_user.get_relationship_data(self, 'amo_collection_author_of')
```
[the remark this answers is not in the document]
diox: Yeah... I tried to follow the established convention used by `addon` (which I also followed when implementing `rating`).

I'm not sure where/how to document that though, given we can't export (or import...) schemas into a file...

src/olympia/bandwagon/views.py
diox: 29c454ea38ae82c026cc9b90ac6fc3a5290c6c5e simplifies the handling and makes it harder to accidentally end up in this scenario, while still allowing reports by pk

eviljeff (approved):

Issue #21445, opened 2023-11-14
Pass images to Cinder moderators and allow restoring them on successful appeal after deletion


diox:
- Blocks https://github.com/mozilla/addons-server/issues/12625
- Blocks https://github.com/mozilla/addons-server/issues/21362
- Blocks full metadata handling for extensions reports as well

When a report is made against content containing images (*), we need to:
- Be able to pass those images to Cinder moderators
- Be able to completely prevent access of those images to end-users when the content is disabled/deleted
- Be able to revert the decision and restore access to those images

(*) The images are:
- For users, the avatar (currently permanently deleted on user ban)
- For extensions, the previews & the icon
- For themes, the previews

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-568)

diox:
Note: in Cinder, if we pass the image as a an attribute with the correct `mime_type` value, it will be displayed as an image https://docs.cinder.co/docs/schema#media-attributes

diox:
QA: https://mozilla-hub.atlassian.net/browse/SVCSE-1619 blocks the full implementation. In the meantime, user reporting & banning/unbanning should work without that new feature.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-568

PR #21454, opened 2023-11-15
Store relationships we delete on ban to be able to restore them later on unban

### LINKED ISSUE #21453: Soft-delete content user was the author of on ban, keeping track of those relationships
To make it possible to unban users, when we ban them, we need to:
- Soft-delete instead of hard-delete their content
- Store the relationships somewhere
- Provide ability to restore the content & relationships

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-572)

diox:
Fixes #21453


src/olympia/users/models.py
```diff
@@ -1193,3 +1249,43 @@ def watch_changes(old_attr=None, new_attr=None, instance=None, sender=None, **kw
...
+    addons_users = models.ManyToManyField('addons.AddonUser')
+    ratings = models.ManyToManyField('ratings.Rating')
+
+    def restore(self):
+        from olympia.addons.models import Addon, AddonUser
+        from olympia.bandwagon.models import Collection
+        from olympia.ratings.models import Rating
+
+        # FIXME: logging
+        Collection.unfiltered.filter(bannedusercontent=self.pk).update(deleted=False)
```
diox, line 1283: Note: here and below, if all these models had `unfiltered_for_relations` like `Addon` (and `Rating` soon...) have that'd be cleaner, but I don't think it's worth introducing just for this narrow use-case

diox: Actually, this helps in tests so I'm going to add that

src/olympia/users/models.py
```diff
@@ -479,27 +479,57 @@ def ban_and_disable_related_content_bulk(cls, users, move_files=False):
...
-        # First addons who have other authors we aren't banning
+        # First addons who have other authors we aren't banning - we are
+        # keeping the add-ons up, but soft-deleting the relationships.
         addon_joint_ids = set(
             AddonUser.objects.filter(addon_id__in=addon_ids)
             .exclude(user__in=users)
             .values_list('addon_id', flat=True)
         )
-        AddonUser.objects.filter(user__in=users, addon_id__in=addon_joint_ids).delete()
+        qs = AddonUser.objects.filter(user__in=users, addon_id__in=addon_joint_ids)
```
eviljeff, line 502: nit: `qs` is an ambiguous variable name in a function that has many querysets.  And below as well. (unless this was intentional to avoid keeping many different querysets around longer than was needed?)

src/olympia/users/models.py
```diff
@@ -479,27 +479,57 @@ def ban_and_disable_related_content_bulk(cls, users, move_files=False):
         from olympia.ratings.models import Rating
         from olympia.users.tasks import delete_photo
```
eviljeff, line 480: just checking that changing `delete_photo` to soft-delete is planned in a follow-up.

eviljeff (approved):

src/olympia/users/models.py
diox: Yeah, that's part of https://github.com/mozilla/addons-server/issues/21445 - I didn't want to block the relationship soft-deletion (or even unbanning itself) on the image handling

src/olympia/addons/models.py
```diff
@@ -2072,7 +2069,20 @@ def category(self):
...
 
+class AddonUserQuerySet(models.QuerySet):
+    def delete(self, hard_delete=False):
+        if hard_delete:
+            return super().delete()
+        else:
+            return self.update(original_role=F('role'), role=amo.AUTHOR_ROLE_DELETED)
+
+    def undelete(self):
+        return self.update(role=models.F('original_role'))
```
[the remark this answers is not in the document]
diox: I thought about it but wasn't sure what - reset back to `AUTHOR_ROLE_DEV` ? I didn't want to introduce a new special value, and I kinda like the idea of keeping a valid value in there in case we accidentally call `undelete()` twice...

src/olympia/users/models.py
diox: Shouldn't make that much difference to keep the different querysets - I'll do that in https://github.com/mozilla/addons-server/pull/21472 which moves that entire method elsewhere

src/olympia/addons/models.py
```diff
@@ -2126,7 +2150,7 @@ class Meta:
 
     def delete(self):
```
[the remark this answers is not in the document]
diox: f79abbfe25e6a1896709eb76c0d3187c0d349b16

src/olympia/addons/models.py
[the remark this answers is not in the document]
diox: 22737a424fe1323271592cdfaaf605017129a7d2 sets it to the `default` value of the field (currently `AUTHOR_ROLE_DEV`)

src/olympia/addons/models.py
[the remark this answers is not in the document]
eviljeff: it's probably something that should be addressed in other models that have a `hard_delete` parameter too - though we need that in some places for tests I suppose.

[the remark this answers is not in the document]
diox: Surprisingly we don't seem to currently used or need it, even in tests, that's why I removed it. We can always add it back later if needed.

Issue #10416, opened 2019-01-16
Prevent rating the same version multiple times


wagnerand:
### Describe the problem and steps to reproduce it:

Unfortunately I don't have STR but it is possible for a user to rate the same version multiple times (not a race condition, the ratings were 6 hours apart).

Looking at the database, there are lots of example for such ratings even after we fixed #6952.

### What happened?

Multiple ratings per user per version.

### What did you expect to happen?

Only one rating per user per version.

### Anything else we should know?
(Please include a link to the page, screenshots and any relevant files.)

eviljeff:
I think we're going to have to bite the bullet and (soft) delete the older duplicate reviews per version, then add a database constraint like we talked about in #6952.  

Complication is legacy pages still allowing multiple ratings per version so we either wait (block) until legacy review pages are deleted or accept that using the legacy page is going to throw a 500 if you try to rate the same version.

wagnerand:
I believe the frontend prevents this but I don't think the API does (unless something has changed since I filed this issue).

diox:
I was unable to find if it was just introduced by 33eba49767f8e3032eae7671ec8531d9d1b56859 or if it's older, but we do have a database constraint nowadays:
https://github.com/mozilla/addons-server/blob/12301a60c69d301283a6e03a89e534bb17cc92e0/src/olympia/ratings/models.py#L142-L145

Since `version` and `user` are foreign keys, this does prevents you from submitting multiple reviews from the same version (while allowing multiple replies, for developers).

diox:
That constraint mentioned above is insufficient: for most users, `reply_to` will be `NULL`, so they'd be considered different values for the database, so the constraint would let you create multiple ratings from the same user on a given version. It does prevent a developer from replying twice to the same review, but that's useless, there is another simpler constraint for this on `reply_to` alone.

What saves us is that for non-deleted ratings, the API has an additional check for this:
https://github.com/mozilla/addons-server/blob/e0e28ebdccf53b231e1d4e340137d0130029ffaf/src/olympia/ratings/serializers.py#L227-L242

In addition, to guard against race conditions, we have throttling (1/minute):
https://github.com/mozilla/addons-server/blob/e0e28ebdccf53b231e1d4e340137d0130029ffaf/src/olympia/ratings/views.py#L35-L43

Still, it's not ideal, we should make a conditional constraint on `user`, `version` instead, only applying it when `deleted` is `0` and `reply_to` is `NULL`.

[...]

PR #21455, opened 2023-11-15
handle webhook rating decisions

### LINKED ISSUE #21351: Handle decisions from Cinder for Ratings
In the [webhook for Cinder](https://github.com/mozilla/addons-server/issues/21194) we need to handle decisions about ratings.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-539)

eviljeff:
fixes #21351


src/olympia/ratings/migrations/0011_remove_rating_one_review_per_user_and_more.py
```diff
@@ -0,0 +1,31 @@
...
+            name='one_review_per_user',
+        ),
+        migrations.AlterField(
+            model_name='rating',
+            name='deleted',
+            field=models.IntegerField(default=0),
+        ),
+        migrations.AddConstraint(
+            model_name='rating',
+            constraint=models.UniqueConstraint(fields=('version', 'user', 'reply_to', 'deleted'), name='one_review_per_user'),
```
eviljeff, line 24: not this pr/issue, but I think we either need to make `version` non-nullable or fix https://github.com/mozilla/addons-server/issues/10416 and drop version from ratings entirely, because null versions make this constraint pointless.  (Although it only works for replies in any case because null values again)

diox: I'm all for it - IIRC the problem is we have data to clean up (duplicate ratings) before we can do that, but we should bite the bullet and do it at some point.

diox (approved):

PR #21464, opened 2023-11-20
Add warning message for when unreviewed versions are in another queue

### LINKED ISSUE #20855: Inform reviewers when another version of an add-on is awaiting review in a different channel
When an add-on submits both listed and unlisted versions (with overlapping review periods), only one of them shows up in the queue. If the reviewer doesn't manually go check the other channel or if the version in the manual review queue is stuck for any reason, they might miss reviewing the version in the other channel

### What happened?
For an add-on with versions pending review in both listed and unlisted channel, only version from either channel (is it always consistently unlisted?) shows in the queue. To discover the other channel's version, the reviewer currently has to either:
- process the review on the version in the manual review queue
- manually check the other channel and make sure there's no version that's also waiting for a review

AMO informs reviewers that there are multiple versions pending review, **AND that some versions might be in a different channel**.

- Ideally, it would be nice to get fully visibility into the multiple versions awaiting review directly through the manual review queue page. However, if there are technical challenges with it, a banner on the review details page alerting the reviewer of the presence of versions in other channels could still be very useful.
- (If not already this way) The version with a earlier review date shows up in the queue, with the other version discoverable through the review details page.

Something like this could be re-utilized for this purpose
image

/cc @wagnerand

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-405)

GregSherman:
Fixes #20855 

Screenshot 2023-11-21 at 12 25 48 AM

This adds a simple red message stating that there are versions with a due date in another queue if there are any.

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/review.html
+++ b/src/olympia/reviewers/templates/reviewers/review.html
@@ -111,6 +111,13 @@
       {% include "reviewers/includes/paginator_history.html" %}
     </div>
   </div>
+  <div id="unreviewed-other-queue">
+    {% if has_versions_with_due_date_in_other_queue %}
+      <p class="unreviewed-versions-warning" style="color: red;">
+        This add-on has 1 or more versions with a due date in other pages.
+      </p>
+    {% endif %}
+  </div>
   <div id="versions-history" class="results">
     <div class="results-inner">
       <table class="review-files item-history">
--- a/src/olympia/reviewers/templates/reviewers/review.html
+++ b/src/olympia/reviewers/templates/reviewers/review.html
@@ -112,9 +112,9 @@
     </div>
   </div>
   <div id="unreviewed-other-queue">
-    {% if has_versions_with_due_date_in_other_queue %}
+    {% if has_versions_with_due_date_in_other_channel %}
       <p class="unreviewed-versions-warning" style="color: red;">
-        This add-on has 1 or more versions with a due date in other pages.
+        This add-on has 1 or more versions with a due date in another channel.
       </p>
     {% endif %}
   </div>
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -725,12 +725,11 @@ def review(request, addon, channel=None):
         flags=flags,
         form=form,
         format_matched_rules=formatted_matched_rules_with_files_and_data,
-        has_versions_with_due_date_in_other_queue=addon.versions.exclude(
-            channel=channel
+        has_versions_with_due_date_in_other_channel=addon.versions(
+            managers='unfiltered_for_relations'
         )
-        .filter(
-            due_date__isnull=False,
-        )
+        .exclude(channel=channel)
+        .filter(due_date__isnull=False)
         .exists(),
         important_changes_log=important_changes_log,
         is_admin=is_admin,
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -726,7 +726,7 @@ def review(request, addon, channel=None):
         form=form,
         format_matched_rules=formatted_matched_rules_with_files_and_data,
         has_versions_with_due_date_in_other_channel=addon.versions(
-            managers='unfiltered_for_relations'
+            manager='unfiltered_for_relations'
         )
         .exclude(channel=channel)
         .filter(due_date__isnull=False)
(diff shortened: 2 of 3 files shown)


GregSherman:
Unsure of

1. If `is_unreviewed` captures all cases of versions with a due date
2. Time to load if no versions have a due date (one passthrough of each version worst case each time). Maybe we check until we find one with a due date instead?

GregSherman:
@eviljeff requesting review

src/olympia/reviewers/views.py
```diff
@@ -439,6 +439,19 @@ def determine_channel(channel_as_text):
...
+    """Return True if the addon has unreviewed versions in another queue."""
+    versions = addon.versions.filter(
+        channel=amo.CHANNEL_LISTED
+        if channel == amo.CHANNEL_UNLISTED
+        else amo.CHANNEL_UNLISTED,
+    )
+    for version in versions:
+        if version.is_unreviewed:
+            return True
+    return False
```
eviljeff, line 452: File reviewed status isn't really the same thing as having a due_date - there can be unreviewed versions that don't need attention (e.g. because the auto approve task is just about to pick them up); and also versions with a reviewed status that have been flagged for needs human review.

Filter on `due_date` instead.

src/olympia/reviewers/views.py
```diff
@@ -439,6 +439,19 @@ def determine_channel(channel_as_text):
     return channel, content_review
 
 
+def other_queue_has_unreviewed_versions(addon, channel):
+    """Return True if the addon has unreviewed versions in another queue."""
+    versions = addon.versions.filter(
+        channel=amo.CHANNEL_LISTED
+        if channel == amo.CHANNEL_UNLISTED
+        else amo.CHANNEL_UNLISTED,
+    )
```
eviljeff, line 448: I guess you can use `exclude` instead
```suggestion
    versions = addon.versions.exclude(channel=channel)
```

src/olympia/reviewers/templates/reviewers/review.html
```diff
@@ -111,6 +111,13 @@ <h3 id="history">
       {% include "reviewers/includes/paginator_history.html" %}
     </div>
   </div>
+  <div id="unreviewed-other-queue">
+    {% if has_versions_with_due_date_in_other_queue %}
+      <p class="unreviewed-versions-warning" style="color: red;">
+        This add-on has 1 or more versions with a due date in other pages.
```
eviljeff, line 117: Can you include a screenshot of what the page looks like?

eviljeff (changes requested):

src/olympia/reviewers/views.py
diox: Also, doing a loop on `versions` like that, you would be making a database query against all versions for the add-on, loading them all, then building a `Version` instance for each loop iteration, just to check that there was something.

Filter on `due_date__isnull=False`, and use `.exists()` on that queryset, no need to loop.

Finally, use `addon.versions(manager='unfiltered_for_relations')`, otherwise you won't see deleted versions in your queryset, and you want to have them in case we need to review something the developer deleted.

src/olympia/reviewers/templates/reviewers/review.html
GregSherman: image

eviljeff, line 117: Can we change it to "in another channel"

src/olympia/reviewers/views.py
```diff
@@ -725,6 +725,13 @@ def review(request, addon, channel=None):
         flags=flags,
         form=form,
         format_matched_rules=formatted_matched_rules_with_files_and_data,
+        has_versions_with_due_date_in_other_queue=addon.versions.exclude(
+            channel=channel
+        )
+        .filter(
+            due_date__isnull=False,
+        )
```
eviljeff, line 733:
```suggestion
        .filter(due_date__isnull=False)
```

src/olympia/reviewers/views.py
```diff
@@ -725,6 +725,13 @@ def review(request, addon, channel=None):
         flags=flags,
         form=form,
         format_matched_rules=formatted_matched_rules_with_files_and_data,
+        has_versions_with_due_date_in_other_queue=addon.versions.exclude(
```
eviljeff, line 728: nit: can we change this naming to "other_channel" too

eviljeff (changes requested):
r+wc lgtm after these small changes

src/olympia/reviewers/views.py
eviljeff, line 728: oh, forget to add this comment: see mat's previous review - we need `addon.versions(manager='unfiltered_for_relations')` not just `addon.versions` to include deleted  versions too.

eviljeff (approved):

PR #21472, opened 2023-11-20
Implement unbanning user(s) functionality in admin/cinder webhook action

### LINKED ISSUE #12625: Make unbanning possible
Banning someone though the admin currently executes `ban_and_disable_related_content()`, which does the following:
- Deletes all their collections (can't be undone)
- Force-disables all add-ons they are the sole author of (can be undone, although it's difficult to separate what had already been force-disabled before banning)
- Removes them from all add-ons they are not the sole author of without touching the add-on (which can't be undone without digging through activity logs, and even then I'm not sure you can get the role back)
- Deletes all their ratings (can be undone, although it's difficult to separate what had already been deleted before banning)
- Deletes their picture (can't be undone)
- Deletes the user, which blanks a few fields - `biography`, `homepage`, `location` - and anonymizes their username (can't be undone)
- Sets a banned date on the account

Because many of these operations can't be undone, unbanning someone is technically possible, but doesn't restore the account back to their original state. We should fix this.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-20)

### LINKED ISSUE #21355: Handle decisions from Cinder about users
In the webhook for Cinder (https://github.com/mozilla/addons-server/issues/21194) we need to handle decisions about users

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-543)

diox:
Fixes #12625
Fixes #21355
~~Depends on https://github.com/mozilla/addons-server/issues/21453~~

Picture restoration on unban is not supported yet, it will be in https://github.com/mozilla/addons-server/issues/21445

### DIFF
--- a/src/olympia/users/models.py
+++ b/src/olympia/users/models.py
@@ -32,7 +32,13 @@ from olympia import activity, amo, core
 from olympia.access.models import Group, GroupUser
 from olympia.amo.decorators import use_primary_db
 from olympia.amo.fields import CIDRField, PositiveAutoField
-from olympia.amo.models import LongNameIndex, ManagerBase, ModelBase, OnChangeMixin
+from olympia.amo.models import (
+    BaseQuerySet,
+    LongNameIndex,
+    ManagerBase,
+    ModelBase,
+    OnChangeMixin,
+)
 from olympia.amo.utils import id_to_path
 from olympia.amo.validators import OneOrMorePrintableCharacterValidator
 from olympia.files.models import File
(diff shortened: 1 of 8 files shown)


src/olympia/users/models.py
```diff
@@ -95,7 +101,150 @@ def build_widget_attrs(self, *args, **kwargs):
...
 
+class UserQuerySet(BaseQuerySet):
+    def ban_and_disable_related_content(self):
+        """Admin method to ban multiple users and disable the content they
+        produced.
+
+        Similar to deletion, except that the content produced by the user is
+        forcibly soft-disabled instead of being deleted where possible, and the
+        user is not anonymized: we keep their data until hard-deletion kicks in
+        (see clear_old_user_data), including fxa_id and email so that they are
```
eviljeff, line 112: Not a change in this pr... but we need `clear_old_user_data` to be able to hard delete instances but the queryset delete methods have been changed to soft-delete instead. (Not applicable to UserProfile or Addon either because we ban and disable those respectively rather than delete)

eviljeff (approved):

src/olympia/users/models.py
diox: Yeah, I'm surprised no tests broke because of that. I've filed https://github.com/mozilla/addons-server/issues/21482 for that.

PR #21484, opened 2023-11-21
move ActivityLog.create to ActivityLogManager

### LINKED ISSUE #21471: Prevent `ActivityLog.objects.create()` from being called, or replace it with `ActivityLog.create()`
Calling `ActivityLog.create()` (custom `classmethod` on `ActivityLog` model) creates the `ActivityLog`, doing some magic to set the `_arguments`, the `user` responsible for the activity and creates the various `*Log` instances like `AddonLog`, `VersionLog` etc.

`ActivityLog.objects.create()` (default django creation mechanism) does not do any of that, so calling it is a trap that should be avoided! Fortunately we aren't currently using it in any production code, just a few tests.

We should:
- Look into whether we can create the various `*Log` instances automatically on creation through `post_save` or custom `save()` method
- Prevent `ActivityLog.objects.create()` from being used or move `ActivityLog.create()` into `ActivityLogManager()`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-577)

eviljeff:
fixes #21471 - a lot of change here but the only changes of note are in ActivityLog and ActivityLogManager - and there it's largely just a straight move.  

I didn't move the associated object creation into a post_save or similar as it'd be the same code in a different place, pretty much, but I could?

### DIFF
--- a/src/olympia/activity/models.py
+++ b/src/olympia/activity/models.py
@@ -312,6 +312,90 @@ class ActivityLogManager(ManagerBase):
         )
         return qs
 
+    def create(self, action, *args, **kw):
+        """
+        e.g. ActivityLog.objects.create(amo.LOG.CREATE_ADDON, addon),
+             ActivityLog.objects.create(amo.LOG.ADD_FILE_TO_VERSION, file, version)
+        In case of circular import you can use `olympia.activity.log_create()`
+        """
+        from olympia import core
+
+        # We might get action as an int, as that's what the model field is defined as
+        action = amo.LOG_BY_ID[action] if isinstance(action, int) else action
+        user = kw.get('user', core.get_user())
+
+        if not user:
+            log.warning('Activity log called with no user: %s' % action.id)
+            return
+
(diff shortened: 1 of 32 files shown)


src/olympia/activity/models.py
```diff
@@ -312,6 +312,90 @@ def review_log(self):
         )
         return qs
 
+    def create(self, action, *args, **kw):
```
diox, line 315: The problem I have with this approach is that `<Model>.<manager>.create()` is a standard Django method and we're overriding it with a slightly incompatible method with a different signature (because of the `action` as first arg)

The code does support passing the `action` as an `int` directly though, which _is_ compatible with the standard Django method, so maybe the only thing to do here is to stop making it the first mandatory argument, just having `*args, **kwargs` as the signature, and ensure all callers pass `action` as a keyword argument ?

eviljeff: https://github.com/mozilla/addons-server/pull/21484/commits/4d1994af818665eee32102b49e4e51fb3da331e1

Issue #21491, opened 2023-11-22
Replace CinderReport with CinderJob


eviljeff:
Cinder consolidates multiple unresolved reports for the same entity under a single job, which will have a single decision_id.  Our current models have a `CinderReport` per `AbuseReport`, so we potentially have to update multiple `CinderReport`s for a single webhook request from Cinder, with the same information (and then try to perform the same action on a single entity).  This is inefficient and prone to complications if each CinderReport is handled seperately, when there is only a single decision (and enforcement action) taking place.

Instead, we should have a CinderJob model (many AbuseReport to a single CinderJob).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-584)

eviljeff:
QA: This issue changes a core model of our Cinder integration, to reflect that cinder combines multiple abuse reports into a single "job", which is resolved with a decision. So, broadly:
- check that you can file multiple reports for the same content
- check that resolving that job (that contains the multiple reports) in cinder takes the action (e.g. disable the addon)
- check that the appeal form works with the new url (that now contains both the abuse report id and the decision id)
  - https://github.com/mozilla/addons-server/issues/21500 adds an email notification for the owner/target of a report with the appeal url in it
  - https://github.com/mozilla/addons-server/issues/21486 adds email notifications for all the reporters with unique appeal urls for each reporter in them.

ioanarusiczki:
@eviljeff 

I only tested multiple reports part for an add-on and noticed that these are gathered in Cinder when the same content is reported multiple times anonymously or by the same reporter (when user is authenticated) There's a reasoning section in Cinder without details yet 

![reasoning](https://github.com/mozilla/addons-server/assets/33448286/16479e0e-d30a-466e-8f21-14abec0ef8e2)

Let me know if I should file an issue. I think this should show what's the report's reason (what the reporter filed from AMO)

eviljeff:
@ioanarusiczki `reasoning` should be there - it's the "details" text box.  If you wrote some text in the abuse reports there's a bug, so can you file an issue please.

ioanarusiczki:
@eviljeff 

Re-tested this:

- I made consecutive reports on the same content, from 3 different browsers being authenticated or anonymous (it doesn't matter) and the reports are gathered in Reports tab under 1 job in Addon Content Infringement 

[...]

PR #21492, opened 2023-11-22
handle the same job_id for multiple reports

### LINKED ISSUE #21485: handle CinderReport.job_id being non-unique
We made the assumption the generated `job_id`s that cinder sends in response to a `create_report` are unique, but it appears that sometimes that is not true.  @diox speculates that it may be because multiple reports for the same entity (e.g. an add-on with a given guid) are being consolidated into a single item in the queue they also share a single job_id.  If that speculation is correct we need to handle that correctly

eviljeff:
this is a quick fix for fixes #21485 but really until https://github.com/mozilla/addons-server/issues/21491 is done they'll be other problems still - like appeals not working because there will be duplicate decision_ids


eviljeff:
#21497 would fix #21491, and make this PR obsolete, but is a more extensive change

eviljeff:
superseded by https://github.com/mozilla/addons-server/pull/21497

Aminkavoos (commented):
Uploading VID-20231127-WA0001.mp4…

PR #21496, opened 2023-11-23
Copy user picture to cloud storage when banning, restore it on unban

### LINKED ISSUE #21445: Pass images to Cinder moderators and allow restoring them on successful appeal after deletion

diox:
Fixes #21445

TODO:
- [ ] Pass the image to Cinder for users (w/ `media_type`)
- [ ] Tests
- [ ] Condition to avoid calling google storage if user or bucket settings are not set up
- [ ] Stop providing defaults for the settings for user/bucket
- [ ] Set up bucket & user on dev/stage/prod


src/olympia/abuse/cinder.py
```diff
@@ -107,12 +113,21 @@ def id(self):
...
     def get_attributes(self):
-        return {
+        data = {
             'id': self.id,
             'name': self.user.display_name,
             'email': self.user.email,
             'fxa_id': self.user.fxa_id,
         }
+        if self.user.picture_type and backup_storage_enabled():
+            filename = copy_file_to_backup_storage(
```
eviljeff, line 123: is there a check we can make before this, to ensure we're not uploading multiple times?  (check a database field, etc).  Just seems unexpected that repeated calling `get_attributes` - that builds an attribute dict - would also hash and upload a file.

src/olympia/amo/utils.py
```diff
@@ -1279,3 +1284,49 @@ def set_max_length_on_fields_where_necessary(self):
...
+def create_signed_url_for_file_backup(backup_file_name_remote):
+    blob = backup_storage_blob(backup_file_name_remote)
+    return blob.generate_signed_url(
+        expiration=datetime.timedelta(days=REPORTED_MEDIA_BACKUP_EXPIRATION_DAYS)
+    )
+
+
+def download_file_contents_from_backup_storage(backup_file_name_remote):
+    blob = backup_storage_blob(backup_file_name_remote)
+    return blob.download_as_string()
```
eviljeff, line 1332: I guess this can fail if the file isn't there (expired or just not uploaded), so do we want some rudimentary exception handling in here?

src/olympia/users/models.py
```diff
@@ -1303,6 +1307,19 @@ class BannedUserContent(ModelBase):
     addons = models.ManyToManyField('addons.Addon')
     addons_users = models.ManyToManyField('addons.AddonUser')
     ratings = models.ManyToManyField('ratings.Rating')
+    picture_backup_name = models.CharField(
+        max_length=75, default=None, null=True, blank=True
+    )
+    picture_type = models.CharField(max_length=75, default=None, null=True, blank=True)
+
+    def restore_avatar(self):
```
eviljeff, line 1315: I prefer avatar to picture too but we should be consistent - i.e. either rename it to avatar everywhere or call this `restore_picture` (or `restore_user_picture`)

src/olympia/users/tests/test_models.py
```diff
@@ -8,10 +8,13 @@
 from django.test.client import RequestFactory
 from django.test.utils import override_settings
 from ipaddress import IPv4Address
+import os.path
+from olympia.amo.tests.test_helpers import get_uploaded_file
```
eviljeff, line 12: I guess all the imports for this file are ignored by ruff because of the `# noqa`s but can we have these in roughly the correct place 😉

eviljeff (approved):

src/olympia/amo/utils.py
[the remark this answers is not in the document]
diox: I didn't realize there was a `download_as_bytes()` method, it's probably better to use that, I'll look into it.

src/olympia/abuse/cinder.py
diox: `copy_file_to_backup_storage()` will avoid the upload if the target filename already exists. We're still hashing and making a request every time though yeah... Not sure how to avoid that.

The problem is, if we instead generate a backup copy every time the user changes their photo, and store the backup name to then send it when it gets reported, we have to potentially keep the backup copy forever instead of limiting its life to ~ 6 months.

Maybe I could, as a follow-up, store the hash on change instead, so that it doesn't have to be regenerated everytime ? That would still leave the request to potentially store the backup copy. I'm not sure it's worth the effort.

src/olympia/amo/utils.py
diox: We should yeah. Not sure what should happen, probably just log ?

src/olympia/users/tests/test_models.py
diox: Oh wow that's weird. I'm used to `ruff` autofixing the imports so now I just stuff them wherever and hit save to have them correctly set. I'll fix this one manually but we'll want to look at why we have these `# noqa` and if we can avoid them, having `ruff` ignore the whole import block is bad...

src/olympia/amo/utils.py
eviljeff: I don't think we can do more than log... maybe log.exception at first?  And then reduce the severity if we're getting a lot for understandable reasons

src/olympia/abuse/cinder.py
eviljeff: storing the hash (or an update datetime?) would mean we wouldn't be attempted a remote call at least (because I guess the check to see if the target filename exists involves at least one call to google storage)

src/olympia/abuse/cinder.py
```diff
@@ -107,12 +113,21 @@ def id(self):
...
 
     def get_attributes(self):
-        return {
+        data = {
             'id': self.id,
             'name': self.user.display_name,
             'email': self.user.email,
             'fxa_id': self.user.fxa_id,
         }
+        if self.user.picture_type and backup_storage_enabled():
```
eviljeff, line 122: Oh, just realised, this is going to run each time a `CinderUser` is sent in a request, but we only need the avatar when the `CinderUser` is being reported, not when they're the reporter or a connected user (author, etc)... right?

src/olympia/abuse/cinder.py
diox: Just storing the hash doesn't avoid the remote calls - we still need to make the remote call to ensure the file exists or upload it if it doesn't, and another to create the signed URL

src/olympia/abuse/cinder.py
diox: Ah, yes. Maybe I should introduce a `minimal` argument to only include the image when we're gathering attributes for reporting a `CinderUser` ? Or do you see a cleaner way of doing it ?

eviljeff: Would a `CinderUser` subclass (e.g. `CinderUserProfile`) make is clearer?  (Would it still be the same entity name in cinder?  I think ...yes?)

diox: Yeah that sounds good. Same entity, just more attributes (from my testing they are optional)

src/olympia/amo/utils.py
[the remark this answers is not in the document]
diox: `download_as_string()` is deprecated and actually return `bytes`, which is what
we want (`SimpleUploadedFile()` uses `bytes` only). `download_as_bytes()` is the correct API to use.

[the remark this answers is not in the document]
diox: 59f3018f3250cb415606cae7b69a3d4f91f080a9

diox: 1f1d0deb024dd3b4fa8f21bee4d957771f1458c5 but I did this at a higher level, because that's where we'll know if it makes sense to ignore the error. This function is a bit too low level to do that.

src/olympia/users/models.py
diox: 3a95da5eba17fd57ad6db2943743ac076d5d0731

src/olympia/users/tests/test_models.py
diox: c86eb51faaffc90839a78851e0ea0eeef9922089

src/olympia/abuse/cinder.py
diox: d53fefd1e68fa1d9d205490c1e348332114784de

PR #21497, opened 2023-11-24
Replace CinderReport with CinderJob

### LINKED ISSUE #21491: Replace CinderReport with CinderJob

eviljeff:
would fix #21491 - does this seem reasonable?  

I didn't write a migration from CinderReport to CinderJob... but I could if we care about the data on dev.


eviljeff:
Note this would make https://github.com/mozilla/addons-server/pull/21492 obsolete

diox:
Just did a quick pass, couple things I'm not sure I like:
- Not having any constraint in `CinderJob` forcing you to have an `AbuseReport`. In the long run it could be something for reviewer decisions that do not come from a report, but then how do you figure out the `target` for the job ?
- Related, calling Cinder first and then creating the `CinderJob` instance and linking it to the `AbuseReport` means we could end up with a job sent to Cinder but not linked to any abuse report if the database call fails (timeout issue or other) after we have made the request...

Maybe this is fine though, need to think about it a bit more.

eviljeff:
> Just did a quick pass, couple things I'm not sure I like:
> 
>     * Not having any constraint in `CinderJob` forcing you to have an `AbuseReport`. In the long run it could be something for reviewer decisions that do not come from a report, but then how do you figure out the `target` for the job ?

it's a concern, but is it even possible to have constraint that in a one to many relationship, many is >=1?

>     * Related, calling Cinder first and then creating the `CinderJob` instance and linking it to the `AbuseReport` means we could end up with a job sent to Cinder but not linked to any abuse report if the database call fails (timeout issue or other) after we have made the request...

We don't know if the `job_id` will be unique or an existing one before we make the call to cinder.  We _could_ create the instance beforehand, and delete it afterwards if an existing `CinderJob` was referenced, but it'd be wasteful.  (Does having a CinderJob instance without any `job_id` if the request fails help anyway?)

diox (approved):

PR #21505, opened 2023-11-24
Fix Cinder reporting when the reporter is also the author/target of the report

### LINKED ISSUE #21502: Cinder reports fails when the authenticated reporter is also the author of the content
As a followup for https://github.com/mozilla/addons-server/issues/21194#issuecomment-1825770324

 I did not receive Cinder reports for dictionaries, the following guids: 
af-ZAAPI@dictionaries.addons.mozilla.org and azdev2@dictionaries.addons.mozilla.org. I've sent ~5 reports for each but I found only one being present in Cinder

Notes:
This is only reproducible for dictionaries, feedbacks for extensions, themes, users, collections or ratings are working as expected.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-588)

diox:
Fixes #21502

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -39,7 +39,9 @@ class CinderEntity:
     def build_report_payload(self, *, report_text, category, reporter):
         context = self.get_context()
         if reporter:
-            context['entities'] += [reporter.get_entity_data()]
+            reporter_entity_data = reporter.get_entity_data()
+            if reporter_entity_data not in context['entities']:
+                context['entities'] += [reporter.get_entity_data()]
             context['relationships'] += [
                 reporter.get_relationship_data(self, 'amo_reporter_of')
             ]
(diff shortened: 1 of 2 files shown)


src/olympia/abuse/cinder.py
```diff
@@ -39,7 +39,9 @@ def get_relationship_data(self, to, relationship_type):
     def build_report_payload(self, *, report_text, category, reporter):
         context = self.get_context()
         if reporter:
-            context['entities'] += [reporter.get_entity_data()]
+            reporter_entity_data = reporter.get_entity_data()
+            if reporter_entity_data not in context['entities']:
+                context['entities'] += [reporter.get_entity_data()]
```
eviljeff, line 44: something something `set`?

eviljeff (approved):

src/olympia/abuse/cinder.py
diox: I thought about using `set` but then I'd need to convert back into a `list` before passing it down to Cinder...

PR #21521, opened 2023-11-28
chore(): add swagger UI to document our apis

KevinMind:

* [ ] This PR relates to an existing open issue and there are no existing
      PRs open for the same issue.
* [ ] Add `Fixes #ISSUENUM` at the top of your PR.
* [ ] The change has been successfully run locally.


requirements/system.txt
```diff
@@ -120,3 +120,12 @@ urllib3==1.26.18 \
     --hash=sha256:f8ecc1bba5667413457c529ab955bf8c67b45db799d159066261719e328580a0
 uWSGI==2.0.21 \
     --hash=sha256:35a30d83791329429bc04fe44183ce4ab512fcf6968070a7bfba42fc5a0552a9
+drf-yasg==1.21.7 \
```
KevinMind, line 123: I had to add the dependencies here or CI would fail.

settings.py
```diff
@@ -150,3 +150,8 @@ def insert_debug_toolbar_middleware(middlewares):
 # Will show the widget but no captcha, verification will always pass.
 RECAPTCHA_PUBLIC_KEY = '6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI'
 RECAPTCHA_PRIVATE_KEY = '6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe'
+
+SWAGGER_SETTINGS = {
+    'USE_SESSION_AUTH': False,
+    'DEEP_LINKING': True,
```
KevinMind, line 156: Allows deep linking to tag/operation/method

eviljeff:
Some general questions:
- are we going to embed it in the api docs like the screenshot suggests, or have a html view directly available?
- for the api docs, is it a build-time embed - static html - or dynamically inserted each time someone views the RTD page?
- for the html view, we it's problematic having it under /api/v5/ etc, as those urls are identified as "api urls" and treated slightly differently - see `is_api` added to the `request`.  (Not really a question. And we could change the regex I suppose)

KevinMind:
> Some general questions:
> 
>     * are we going to embed it in the api docs like the screenshot suggests, or have a html view directly available?
> 
>     * for the api docs, is it a build-time embed - static html - or dynamically inserted each time someone views the RTD page?
> 
>     * for the html view, we it's problematic having it under /api/v5/ etc, as those urls are identified as "api urls" and treated slightly differently - see `is_api` added to the `request`.  (Not really a question. And we could change the regex I suppose)

@eviljeff Great point about the versioning part. I didn't realize we don't have all API endpoints under the API app. I can mount swagger on the root as well.

What exactly would be the issue resulting from  this

`
for the html view, we it's problematic having it under /api/v5/ etc, as those urls are identified as "api urls" and treated slightly differently 
`

For how we embed, that's a great question. Right now it is served directly from our django app. The examples in the description are supposed to be just that, examples. I think the first step would be to get the basic integration setup, make the documentation browsable, make any tweaks. Next steps could be investigating how we balance between manually documentation and automated.

I would have a slight preference for totally automating, but there might be some info in our existing docs that we cannot easily surface directly in swagger. I think that's something we would have to investigate and I would need support with someone who is more familiar with the content of the docs.

diox:
> for the html view, we it's problematic having it under /api/v5/ etc, as those urls are identified as "api urls" and treated slightly differently 

API requests in addons-server:
- Don't use auth through the regular middleware, and don't use session cookies - DRF handles authentication according to `authentication_classes` for the view
- Render errors as json
- Are cached by default through the CDN with `API_CACHE_DURATION` (currently 3 minutes) + get specific headers added to the `Vary` header in their responses

The auth+caching is fine, the errors (404/500) are more annoying, so we'd need to change the regexp that identifies api URLS (`DRF_API_REGEX`) or find some other workaround if we leave swagger here.

src/olympia/abuse/models.py
```diff
@@ -379,7 +390,9 @@ class AbuseReport(ModelBase):
     collection = models.ForeignKey(
         Collection, null=True, related_name='abuse_reports', on_delete=models.SET_NULL
     )
-    message = models.TextField(blank=True)
+    message = models.TextField(
+        blank=True, help_text='The body/content of the abuse report (required).'
```
diox, line 394: It's not required if a reason is sent

src/olympia/api/urls.py
```diff
@@ -1,11 +1,50 @@
...
+        re_path(
+            r'^swagger/$',
+            schema_view.with_ui('swagger', cache_timeout=0),
+            name='schema-swagger-ui',
+        ),
+        re_path(
+            r'^redoc/$',
+            schema_view.with_ui('redoc', cache_timeout=0),
+            name='schema-redoc',
+        ),
```
diox, line 41: Maybe we should do this on local envs/dev only for now ? This isn't completely ready/useful for production for all APIs...

src/olympia/lib/settings_base.py
```diff
@@ -329,6 +329,9 @@ def get_db_config(environ_var, atomic_requests=True):
     r'^registration\/',
     # Django's sitemap_index.xml template uses some syntax that jinja doesn't support
     r'sitemap_index.xml',
+    # Swagger URLs are for the API docs, and are not meant to be rendered by Django.
```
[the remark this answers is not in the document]
KevinMind: This is needed to force using the django template engine as that is what the swagger UI templates are written for. Django is rendering theme. (The comment is confusing, I will update)

src/olympia/abuse/models.py
KevinMind: I copy pasted from the existing docs. So I guess those docs are incorrect?

[image](https://addons-server.readthedocs.io/en/latest/topics/api/abuse.html#post--api-v5-abuse-report-addon-)

src/olympia/lib/settings_base.py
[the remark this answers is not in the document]
eviljeff: I think it's the "not meant to be rendered by Django" which is confusing, if they are rendered by Django.

requirements/system.txt
KevinMind: done

src/olympia/lib/settings_base.py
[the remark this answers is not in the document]
KevinMind: Fixed.

src/olympia/api/urls.py
KevinMind: Added settings.DEBUG for now for simplicity.

src/olympia/lib/settings_base.py
```diff
@@ -329,6 +329,9 @@ def get_db_config(environ_var, atomic_requests=True):
     r'^registration\/',
     # Django's sitemap_index.xml template uses some syntax that jinja doesn't support
     r'sitemap_index.xml',
+    # Swagger URLs are for the API docs, use some syntax that jija doesn't support
```
eviljeff, line 332:
```suggestion
    # Swagger URLs are for the API docs, use some syntax that jinja doesn't support
```

src/olympia/abuse/serializers.py
```diff
@@ -98,7 +103,9 @@ def create(self, validated_data):
 
 
 class AddonAbuseReportSerializer(BaseAbuseReportSerializer):
-    addon = serializers.SerializerMethodField()
+    addon = serializers.SerializerMethodField(
+        help_text='The id, slug, or guid of the add-on to report for abuse (required).'
```
eviljeff, line 107: Is this correct? - `SerializerMethod` is a read-only field so a value can't be provided for it

src/olympia/api/urls.py
```diff
@@ -1,11 +1,56 @@
...
+            description='The official API for addons.mozilla.org.',
+        ),
+        public=True,
+        permission_classes=(permissions.AllowAny,),
+    )
+
+    routes = url_patterns
+
+    # For now, this feature is only enabled in dev mode
+    if settings.DEBUG:
```
eviljeff, line 30: this means it will only be available locally, not even on addons-dev.allizom.org.  That's what you intended?

eviljeff (approved):
with the fix to address the CORS failing tests, looks good to me

src/olympia/abuse/serializers.py
KevinMind: You are right, good catch. I copied the request field and not the response field. Updated

src/olympia/api/urls.py
KevinMind: I think it's okay like that for now. We can always change it. I believe it will be useful for prod, but we can figure out where it fits and change it then rather than creating a new setting.

PR #21549, opened 2023-12-04
Don't try to backup/restore from storage bucket if it's not enabled

### LINKED ISSUE #21546: Don't copy file to backup storage if backup storage is not enabled
Follow-up from https://github.com/mozilla/addons-server/issues/21506

https://mozilla.sentry.io/issues/4697455318/?project=6224627

Looks like at banning time, we don't check whether backup storage is enabled or not, resulting in this error in `delete_all_addon_media_with_backup` task. The ban would still go through, but we should fix this.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-607)

diox:
Fixes #21546

### DIFF
--- a/src/olympia/addons/tasks.py
+++ b/src/olympia/addons/tasks.py
@@ -24,6 +24,7 @@ from olympia.addons.utils import compute_last_updated, remove_icons
 from olympia.amo.celery import task
 from olympia.amo.decorators import set_modified_on, use_primary_db
 from olympia.amo.utils import (
+    backup_storage_enabled,
     copy_file_to_backup_storage,
     download_file_contents_from_backup_storage,
     extract_colors_from_image,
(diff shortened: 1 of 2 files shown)


src/olympia/addons/tasks.py
```diff
@@ -125,7 +127,7 @@ def delete_all_addon_media_with_backup(id, **kwargs):
 def restore_all_addon_media_from_backup(id, **kwargs):
     addon = Addon.unfiltered.all().get(pk=id)
     disabled_addon_content = DisabledAddonContent.objects.filter(addon=addon).last()
```
eviljeff, line 129: Could approve efficiency when backup_storage is not enabled by only doing this query if it's enabled.
`qs = DisabledAddonContent.objects.filter(addon=addon)`
and then `if backup_storage_enabled() and (disabled_addon_content := qs.last()):`

(if we care about this case)

eviljeff (approved):

PR #21556, opened 2023-12-05
ADDSRV-601

KevinMind:

* [ ] This PR relates to an existing open issue and there are no existing
      PRs open for the same issue.
* [ ] Add `Fixes #ISSUENUM` at the top of your PR.
* [ ] The change has been successfully run locally.


src/olympia/users/tests/test_tasks.py
```diff
@@ -172,3 +178,138 @@ def test_resize_photo_poorly():
...
+        responses.add(
+            responses.GET,
+            f"{settings.SOCKET_LABS_HOST}/servers/{settings.SOCKET_LABS_SERVER_ID}/suppressions/download?sortField=suppressionLastUpdate&sortDirection=dsc",
+            status=500,
+        )
+
+        with raises(HTTPError):
+            sync_blocked_emails()
+
+        # Asser that task is retried
```
KevinMind, line 204: Have not figured out how to implement this test, expecting a retry.. It might actually be that the task is not retrying... I'm not 100% sure.

I tried replaceing line 201 with

```python
with raises(Retry):
```

And this test failed claiming the exception raised was an HTTPError, but that is the exception that should trigger the retry.. so not totally sure what to do here.

src/olympia/users/tests/test_tasks.py
```diff
@@ -172,3 +178,138 @@ def test_resize_photo_poorly():
...
+Test scenarios:
+2. recieve bad response from api
+3. cannot reach API, timeout
+5. CSV is empty
+6. CSV contains existing email
+7. CSV contains unique email
+10. Deletes file if error occurs while processing recores.
+
+
+1. environment variables are undefined.
```
KevinMind, line 311: These are test scenarios I could think of but not sure if we actually want to test. Some would be bleeding into the territory of testing the contract of socketlabs, making sure we can handle breaches in that contract, others are testing things like setttings/env variables.. not sure it's worth it. Finally, we will likely run this task with large CSV files and I didn't implement any kind of test for handling this. But since the code is batched, it should be relatively performant.

KevinMind:
@eviljeff you mentioned at one point in one of the planning documents not being the _biggest_ fan of the name `EmailBlock`. Can we address that here? Do you have a better name in mind?

src/olympia/users/tasks.py
```diff
@@ -65,3 +77,49 @@ def update_user_ratings_task(data, **kw):
...
+
+    # Convert bytes to a file-like object
+    file_like_object = io.BytesIO(response.content)
+    # Create a Django ContentFile from the file-like object
+    django_file = ContentFile(file_like_object.read())
+    # Get 10 character uinique ID for file
+    file_id = str(uuid.uuid4())[:10]
+
+    # save csv to disk
+    file_path = default_storage.save(f'tmp/suppressions-{file_id}.csv', django_file)
```
eviljeff, line 106: Can you use `tempfile.NamedTemporaryFile` or `tempfile.mkdtemp`? (both with `dir=settings.TMP_PATH` so it's stored in the appropriate place).

src/olympia/users/tasks.py
```diff
@@ -65,3 +77,49 @@ def update_user_ratings_task(data, **kw):
...
+
+    task_log.info(f'Downloaded suppression list to {file_path}')
+
+    with default_storage.open(file_path, 'r') as csv_file:
```
eviljeff, line 110: if you use tempfile with the `with` it'll clean-up without having to manually `delete` at the end too.

src/olympia/users/tests/test_tasks.py
```diff
@@ -172,3 +179,136 @@ def test_resize_photo_poorly():
...
+4. File is deleted before able to read it.
+8. CSV contains same email twice
+9. CSV contains many records.
+"""
```
eviljeff, line 314: I'm not sure what this comment is - is it explaining what the tests above do?  Or an extended #TODO?

eviljeff:
> @eviljeff you mentioned at one point in one of the planning documents not being the _biggest_ fan of the name `EmailBlock`. Can we address that here? Do you have a better name in mind?

`SurpressedEmail` was one idea; `SocketLabsBlockedEmail` or `SocketLabsSurpressedEmail` is explicit but verbose; I dunno really 🤷‍♂️

src/olympia/users/tasks.py
```diff
@@ -65,3 +77,49 @@ def update_user_ratings_task(data, **kw):
...
+
+
+BATCH_SIZE = 100
+
+
+@task(autoretry_for=(HTTPError, Timeout), max_retries=5, retry_backoff=True)
+def sync_blocked_emails(batch_size=100, **kw):
+    url = f"""
+        {settings.SOCKET_LABS_HOST}/servers/{settings.SOCKET_LABS_SERVER_ID}/suppressions/download?sortField=suppressionLastUpdate&sortDirection=dsc
+    """.strip()
```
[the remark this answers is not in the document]
KevinMind: What's the logic here? How is one better than the other?

src/olympia/users/tasks.py
KevinMind: Could you help me with a code snippet of exactly what you mean? :)

src/olympia/users/tasks.py
KevinMind: Nice, very cool.

src/olympia/users/tests/test_tasks.py
KevinMind: It's an extended todo. I added a [comment](https://github.com/mozilla/addons-server/pull/21556#discussion_r1415568442) explaining why it is there for now. I'm still working on making the tests for this PR right, and the bottom chunk are things that could be tested, but maybe shouldn't. Wanted to get some feedback as to how to handle them.

eviljeff:
Any synonym for (Email)Block(edEmail) that doesn't literally include the words "Block"/"Blocked" is fine with me 😄

src/olympia/users/tasks.py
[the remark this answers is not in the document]
eviljeff: It's just coding norms in this repo - we rarely use multiline comments (other than function descriptions), and very rarely use multiline f strings - preferring single line strings with `'`.  Plus we don't need the `strip()` to hack around the extra whitespace and line breaks.

diox:
"Suppressed" is a fairly common term in the industry to describe this, it's not just a Socketlabs thing. Since we're syncing up Socketlabs suppression list locally, I think this makes sense ? I agree about not using `block` because of how loaded that term is internally (and definitely agree with not using `socketlabs` in the name - we need to stay provider-independant)

src/olympia/users/tasks.py
eviljeff: it's difficult to write a complete snippet because it changes how you've written the rest of the function - you're using django storage rather than tempfile right now. 

Something like `with tempfile.NamedTemporaryFile(dir=settings.TMP_PATH, delete=not settings.DEBUG) as file_obj` and then write to the file_obj from `file_like_object`.

Or can you just use `csv.reader` to read directly from `file_like_object`? 🤔

src/olympia/users/tasks.py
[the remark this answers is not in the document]
KevinMind: I only have it as multiline as the linter complained for a single line. 

Is it possible to enforce this kind of a style rule programattically?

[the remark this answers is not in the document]
eviljeff: I couldn't find anything in the `ruff` rules/config, but imo it'd be great if we could prevent multi-line comments entirely, unless overridden with specific rule.

src/olympia/users/tasks.py
KevinMind: Okay so I tried both approaches. I ended up with using tempfile and then opening the tempfile inside the block. The original code was performing against the 21mb production csv at 10seconds. Trying to read the file_like_object directly tripled that. using tempfile brought it back down.

I would assume that reading the file directly in memory is less memory performant than reading from an actual file where we can stream chunks via the csv reader, but I don't know for sure. Added the change to a tmp commit that I'll squash before merging. LMK what you think.

KevinMind:
Let's go with `SuppressedEmail` then as a middle ground between officially linking our model to socket labs and confusing with "block". I'll update the code.

eviljeff (approved):

PR #21573, opened 2023-12-06
Cinder metadata tweaks: make sure we send data that passes validation

### LINKED ISSUE #21570: Remove or stop exposing invalid support_url / support_email / homepage from the database
While validation should prevent that, we have some (old?) data with invalid `support_url`, `support_email` and/or `homepage` values in our database, at least on dev.

On the frontend this is not too much of an issue for URLs, because they get through the outgoing proxy and so worst case scenario they point to `/` instead of whatever non-URL garbage the field contains. But for things like Cinder, an error will be raised when we send a report when those fields are invalid, preventing the report against the add-on from being recorded.

If possible we should clean the translations from our database, and maybe stop exposing invalid values in both our API and Cinder reporting code.

diox:
To that end, also clear old invalid `support_url` and `homepage` data so that it's never exposed.

Fixes #21570


src/olympia/abuse/cinder.py
```diff
@@ -288,12 +288,11 @@ def get_extended_attributes(self):
...
             )
-        # In addition, the URL/email fields can't be sent if blank as they
-        # would not be considered valid by Cinder.
-        for field in ('homepage', 'support_email', 'support_url'):
-            if value := getattr(self.addon, field):
-                data[field] = self.get_str(value)
-
+        # The URL/email fields can't be sent as empty strings as they would not
+        # be considered valid by Cinder.
+        data['homepage'] = self.get_str(self.addon.homepage) or None
```
eviljeff, line 293: should the `CinderUser` homepage also be handled the same?

src/olympia/addons/migrations/0049_clear_bad_url_data.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('addons', '0048_deletedpreviewfile_alter_addonuser_original_role_and_more'),
+        ('translations', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
```
eviljeff, line 13: You manually tested this (I hope)?

src/olympia/addons/migrations/0049_clear_bad_url_data.py
```diff
@@ -0,0 +1,29 @@
...
+            'UPDATE `translations` '
+            'JOIN `addons` ON `translations`.`id` = `addons`.`homepage` '
+            'SET `localized_string`="" '
+            'WHERE `translations`.`localized_string` != "" '
+            'AND NOT `translations`.`localized_string` LIKE "http://%" '
+            'AND NOT `translations`.`localized_string` LIKE "https://%" ',
```
eviljeff, line 19: Do we need to fix `localized_string_clean` too?

eviljeff (approved):

src/olympia/addons/migrations/0049_clear_bad_url_data.py
diox: Yes. You can verify this by manually creating bad `homepage`/`support_url` data in a shell and running the migration.

src/olympia/abuse/cinder.py
diox: Good call. Looks like we also need a migration there!

diox: And while fixing the API for users I also discovered the add-ons API wasn't validating properly (it allows URLs even if they don't have an http/https protocol) so I'll be fixing that as well...

diox: https://github.com/mozilla/addons-server/pull/21573/commits/069d04965ffc25d9ef0d4495ae1461e6444cdedf and https://github.com/mozilla/addons-server/pull/21573/commits/d35d776495f1886f8bfe2e87c1e89ad488fc7725

diox:
This deserves a new review now that I've taken care of `homepage` for users and added URL validation to the API

src/olympia/amo/validators.py
```diff
@@ -74,3 +77,20 @@ class PreventPartialUpdateValidator:
...
             raise fields.SkipField()
+
+
+def HttpHttpsURLValidator(message=None, code=None):
+    return URLValidator(message=message, code=code, schemes=('http', 'https'))
+
+
+def NoAMODomainValidator():
+    return RegexValidator(
+        regex=r'%s' % re.escape(settings.DOMAIN),
```
eviljeff, line 88: Using `settings.EXTERNAL_SITE_URL` seems safer then `settings.DOMAIN` - if was used on reviewers.m.o, services.a.m.o, or the admin instances it'd consistently be the primary addons.mozilla.org domain.

src/olympia/addons/tests/test_views.py
```diff
@@ -1299,12 +1299,12 @@ def test_set_disabled(self):
...
 
-    @override_settings(EXTERNAL_SITE_URL='https://amazing.site')
+    @override_settings(DOMAIN='amazing.site')
     def test_set_homepage_support_url_email(self):
         data = {
-            'homepage': {'ro': '#%^%&&%^&^&^*'},
+            'homepage': {'en-US': '#%^%&&%^&^&^*'},
             'support_email': {'en-US': '#%^%&&%^&^&^*'},
-            'support_url': {'fr': '#%^%&&%^&^&^*'},
+            'support_url': {'en-US': '#%^%&&%^&^&^*'},
```
eviljeff, line 1307: this mix of locales was on purpose to make sure l10n was working, so doesn't changing it to 'en-US` reduce our coverage?

src/olympia/addons/tests/test_views.py
```diff
@@ -1316,14 +1316,28 @@ def test_set_homepage_support_url_email(self):
         }
 
         data = {
-            'homepage': {'ro': settings.EXTERNAL_SITE_URL},
-            'support_url': {'fr': f'{settings.EXTERNAL_SITE_URL}/foo/'},
+            'homepage': {'en-US': f'https://{settings.DOMAIN}'},
+            'support_url': {'en-US': f'https://{settings.DOMAIN}/foo/'},
```
eviljeff, line 1320: same question here on changing to `en-US` (and so on in this file)

eviljeff (approved):
r+wc

src/olympia/addons/tests/test_views.py
diox: As above

src/olympia/amo/validators.py
diox: Fair. We weren't consistent before, with the validator used in devhub using `DOMAIN` and the one in the API using `EXTERNAL_SITE_URL`. I've picked `DOMAIN` but I agree maybe `EXTERNAL_SITE_URL` is better here. I'll update.

Issue #21578, opened 2023-12-07
Display Suppressed Email status in devhub


KevinMind:
### 

```[tasklist]
### Tasks
- [ ] api to check if an  email is suppressed
- [ ] update devhub view to display email suppressed status
``` 

Note: until we implement the apis for verifying suppressed emails, we can not offer a CTA for the user (other than changing their email address which we probably don't want to do without offering also to verify. That would be misleading about their options.)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-620)

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-620

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-620

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-620

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-620

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-620

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-620

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-620

PR #21579, opened 2023-12-07
feat(): add api to check suppressed email

KevinMind:
Related: #21578 

1/2 introduces an api for checking the if a given email address is suppressed.

```[tasklist]
### Tasks
- [x] add the api view and serializer
- [x] add tests
- [x] add documentation
```

### DIFF
--- a/docs/topics/api/index.rst
+++ b/docs/topics/api/index.rst
@@ -47,4 +47,5 @@ using the API.
    reviewers
    scanners
    shelves
+   suppressed_emails
    tags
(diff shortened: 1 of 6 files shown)


KevinMind:
I just realized, I probably won't need this API (at least right now) We will not need this in AMO frontend as the devhub is rendered by django.. for now. @diox I'll leave it up to you if you think we should just kill it or merge it.

diox (commented):
Let's kill it for now and not expose it to the outside world

PR #21582, opened 2023-12-08
Introduce new CinderReport entity, attach reason and reporter to it

### LINKED ISSUE #21571: Pass abuse report reason to Cinder though a dedicated entity
We need to pass the report reason to Cinder - currently it's only used to determine what queue a report goes to.

The suggested approach by MoSo folks is to use a dedicated entity and add a relationship to that entity. Later, we could have a custom renderer in Cinder to display the reason more prominently (otherwise it's going to be buried in the relationships tab)

diox:
This replaces the relationship between a reported entity and a reporter by two relationships: one between the reported entity and a new report entity, and one between that new report entity and the reporter.

Fixes #21571

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -56,27 +56,35 @@ class CinderEntity:
         # are typically returned there instead of in get_attributes().
         return {}
 
-    def build_report_payload(self, *, report_text, category, reporter):
+    def build_report_payload(self, *, report, reporter):
         context = self.get_context()
-        if reporter:
-            reporter_entity_data = reporter.get_entity_data()
-            if reporter_entity_data not in context['entities']:
-                context['entities'] += [reporter.get_entity_data()]
+        if report:
+            context['entities'] += [report.get_entity_data()]
             context['relationships'] += [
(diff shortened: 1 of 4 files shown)


src/olympia/abuse/cinder.py
```diff
@@ -56,27 +56,35 @@ def get_extended_attributes(self):
...
         return {}
 
-    def build_report_payload(self, *, report_text, category, reporter):
+    def build_report_payload(self, *, report, reporter):
         context = self.get_context()
-        if reporter:
-            reporter_entity_data = reporter.get_entity_data()
-            if reporter_entity_data not in context['entities']:
-                context['entities'] += [reporter.get_entity_data()]
+        if report:
```
eviljeff, line 61: I'm not sure about the use-case of `report` not being provided but 🤷‍♂️

eviljeff (approved):

src/olympia/abuse/cinder.py
diox: Just didn't want to fail completely initially when I was making the changes. I kept it because I figured we might want to be able to build a payload without a `report` for reviewer actions later

PR #21605, opened 2023-12-12
Change randomness of search results every hour instead of every day

### LINKED ISSUE #21604: Change sort=random seed every hour instead of every day
On the AMO homepage and other places we routinely point users to the full list of desktop Recommended extensions via this link:

https://addons.mozilla.org/firefox/search/?promoted=recommended&sort=random&type=extension

Note the "sort=random" element of the string. This is intended so that users won't see the same static list of Recommended extensions each time they hit the link; we want to give more equitable surfacing to all Recommended extensions. 

However it seems to not be working as intended. I'm consistently seeing our single most popular Recommended extension — uBlock Origin — appear on either the last or second to last page each time. 

How often is the "random" supposed to work? With each refresh of the page? Every hour? 24 hours? But more curiously, is there any reason I'm seeing uBO consistently bumped down the list?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-626)

diox:
Fixes #21604

### DIFF
--- a/src/olympia/search/filters.py
+++ b/src/olympia/search/filters.py
@@ -1,4 +1,4 @@
-from datetime import date
+from datetime import datetime
 
 from django.utils import translation
 from django.utils.encoding import force_str
@@ -1011,14 +1011,15 @@ class SortingFilter(BaseFilterBackend):
                     or AddonPromotedQueryParam.query_param in request.GET
                 ) and not search_query_param
                 if is_random_sort_available:
-                    # We want randomness to change only once every 24 hours, so
-                    # we use a seed that depends on the date.
+                    # Third, we want randomness to change only once every hour,
+                    # so we use a seed that depends on the date + hour.
+                    now = datetime.now()
                     qs = qs.query(
                         'function_score',
                         functions=[
                             query.SF(
                                 'random_score',
-                                seed=date.today().toordinal(),
+                                seed=now.toordinal() * 24 + now.hour,
                                 field='id',
                             )
                         ],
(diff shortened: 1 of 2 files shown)


src/olympia/search/filters.py
```diff
@@ -1011,14 +1011,15 @@ def filter_queryset(self, request, qs, view):
...
+                    # Third, we want randomness to change only once every hour,
+                    # so we use a seed that depends on the date + hour.
+                    now = datetime.now()
                     qs = qs.query(
                         'function_score',
                         functions=[
                             query.SF(
                                 'random_score',
-                                seed=date.today().toordinal(),
+                                seed=now.toordinal() * 24 + now.hour,
```
eviljeff, line 1022: Just pointing out that the greater the frequency of the seed changing, the more chance that a user will be browsing a list of search results sorted by random and the order suddenly changes when they're on x page of y results.

eviljeff (approved):
r+wc ... but really this doesn't fix the original complaint - that super popular add-ons are just as likely to be on page 5 than on page 1 with a random sort 😐

src/olympia/search/filters.py
diox: I agree, I'm following what was asked though - we can revisit if this cause problems. I think this is still a long enough period of time that it's not a huge issue, and also the recommended shelf (only search that is likely to be using random sort) is small enough that users won't be paginated through the results a lot.

PR #21611, opened 2023-12-12
rm AbuseReport.state

### LINKED ISSUE #21610: Drop `AbuseReport.state`
`AbuseReport.state` is used to store the resolution state of an abuse report - if it has been reviewed by Mozilla and determined to be a valid report, (or suspicious) or deleted because it's invalid.  Now abuse reports will be resolved either in Cinder or in the reviewer tools directly, and the decision record in a `CinderJob` instance, so the field is obsolete.  From https://sql.telemetry.mozilla.org/queries/96519/source it doesn't seem to have been used anyway.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-629)

eviljeff:
fixes #21610


src/olympia/abuse/migrations/0020_alter_abusereport_addon_install_method_and_more.py
eviljeff, line 1: these were your changes @KevinMind 😜

src/olympia/abuse/migrations/0019_alter_abusereport_drop_state_options_alter_abusereport_managers.py
```diff
@@ -0,0 +1,29 @@
...
+        ),
+        migrations.AlterModelManagers(
+            name='abusereport',
+            managers=[
+            ],
+        ),
+        migrations.RemoveField(
+            model_name='abusereport',
+            name='state',
+        ),
```
diox, line 28: Field was still there in the model before this commit so if you drop the database column now all `create()` and `save()` on `AbuseReport` will fail during the push.

eviljeff: forgot about that - https://github.com/mozilla/addons-server/pull/21611/commits/69fb65a5cc1a66f04a80daafcc4f7c0ec0398ab7

src/olympia/abuse/migrations/0019_alter_abusereport_drop_state_options_alter_abusereport_managers.py
```diff
@@ -0,0 +1,30 @@
...
+        migrations.AlterModelManagers(
+            name='abusereport',
+            managers=[
+            ],
+        ),
+        migrations.AlterField(
+            model_name='abusereport',
+            name='state',
+            field=models.BooleanField(default=None, null=True)
```
diox, line 28: It wasn't a `BooleanField` before, so if you do this that will result in 2 `ALTER TABLE`

```suggestion
            field=models.PositiveSmallIntegerField(default=None, null=True)
```

diox (approved):

PR #21612, opened 2023-12-13
chore(swagger): add session auth settings for authenticated requests in swagger ui

KevinMind:
Fixes: #21613

Quick fix for swagger UI to allow setting "Authorization": "Session <session-id>" headers to make authenticated requests from the UI.

https://github.com/mozilla/addons-server/assets/19595165/448c7436-3b36-4039-891b-045cae1b1956

## Testing

Follow the video more or less

1. open http://olympia.test/api/v5/swagger/#/accounts/accounts_account_read
2. click "try it out"
3. type a valid account ID (1 is probably a safe bet)
4. should get 404 and no authorization header
5. now click on the little lock
image
6. copy your sessionId from cookies and paste it (include the "Session" prefix)
7. now try to execute the api again.
8. you should now get back your user account in the swagger UI response box.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -151,7 +151,33 @@ SITEMAP_DEBUG_AVAILABLE = True
 RECAPTCHA_PUBLIC_KEY = '6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI'
 RECAPTCHA_PRIVATE_KEY = '6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe'
 
+ADDONS_SERVER_DOCS_URL = 'https://addons-server.readthedocs.io/en/latest'
+
+
 SWAGGER_SETTINGS = {
     'USE_SESSION_AUTH': False,
     'DEEP_LINKING': True,
+    'SECURITY_DEFINITIONS': {
+        'Session ID': {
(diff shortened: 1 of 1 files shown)


settings.py
```diff
@@ -154,4 +154,12 @@ def insert_debug_toolbar_middleware(middlewares):
 SWAGGER_SETTINGS = {
     'USE_SESSION_AUTH': False,
     'DEEP_LINKING': True,
+    'SECURITY_DEFINITIONS': {
+        'Session': {
```
eviljeff, line 158: Can we add support for JWT auth too?  https://addons-server.readthedocs.io/en/latest/topics/api/auth.html or is that not possible?  Thinking: we write these docs to be used by 3rd party developers, and JWT auth is what we encourage for them (because it doesn't require a browser login to get the session id, and doesn't expire)

eviljeff (approved):

settings.py
KevinMind: It's a little bit _odd_. So actually all this does is add a UI box to authenticate adding an `Authorization` header. It will let you put `JWT <token>` but I can see how the naming is a bit confusing.

I updated it to have session and jwt as separate options to make it clear there are both.
I added descriptions to both, and included how to format the header.

Does that help?

image

PR #21621, opened 2023-12-13
Fix recording of statsd timings for class-based views

### LINKED ISSUE #21620: Statsd timings for class-based-views are broken
For class-based-views that are _not_ DRF viewsets, we are recording an incorrect view name. Not sure yet when this started, but this is visible now in https://earthangel-b40313e5.influxcloud.net/d/kiTC7XDZz/amo-prod-frontend-apis-usage-performance?orgId=1 : all non-viewsets timings are missing.

Logic in django-statsd middleware is pretty basic: https://github.com/django-statsd/django-statsd/blob/a781b213296663fc8ca34e370c160b04ca95c91a/django_statsd/middleware.py#L43-L51

This returns `view` for the affected views instead of their class name... This used to work, but broke during Django 4.2 deployment earlier this year.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-632)

diox:
Fixes #21620

### DIFF
--- a/src/olympia/amo/middleware.py
+++ b/src/olympia/amo/middleware.py
@@ -1,6 +1,7 @@
 import contextlib
 import os
 import re
+import time
 import uuid
 from urllib.parse import quote
 
@@ -34,6 +35,9 @@ from django.utils.translation import activate, gettext_lazy as _
 
 import MySQLdb as mysql
 from django_statsd.clients import statsd
+from django_statsd.middleware import (
+    GraphiteRequestTimingMiddleware as GraphiteRequestTimingMiddlewareUpstream,
+)
 from rest_framework import permissions
 
 import olympia.core.logger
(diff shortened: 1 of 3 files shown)


src/olympia/amo/middleware.py
```diff
@@ -458,3 +462,22 @@ def process_response(self, request, response):
...
+        view = view_func
+        if hasattr(view, 'view_class'):
+            view = view.view_class
+        if not hasattr(view, '__name__'):
+            # A class-based view
+            view = view.__class__
+
+        request._view_module = view.__module__
+        request._view_name = view.__name__
+        request._start_time = time.time()
```
eviljeff, line 483: we don't have to `super()` I guess?

eviljeff (approved):

src/olympia/amo/middleware.py
diox: We could call `super()` before our stuff but we'd end up overriding `request._view_module` and `request._view_name` so I didn't think it was worth it but maybe?

eviljeff: 🤷‍♂️ I just noticed it wasn't there. I guess there could be more in the upstream function in the future, but on the other hand it might be incompatible so... it's fine.

PR #21622, opened 2023-12-13
Send relationships to Cinder past the first 25 items in batches

### LINKED ISSUE #21523: Unable to properly report (and therefore ban) a user that has lots of add-ons through Cinder
Followup for 
https://github.com/mozilla/addons-server/issues/21513#issue-2014107172

Expected result: 

I sent from Cinder decisions using each policy available and it should ban a user (except for "Ignore") 

Actual result: 
User 10640485 is not banned. The user profile is still public.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-595)

diox:
Fixes #21523

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -9,9 +9,14 @@ import requests
 from olympia import amo
 from olympia.amo.utils import (
     backup_storage_enabled,
+    chunked,
     copy_file_to_backup_storage,
     create_signed_url_for_file_backup,
 )
+
+
+# Number of relationships to send by default in each Cinder request.
+RELATIONSHIPS_BATCH_SIZE = 25
 
 
 class CinderEntity:
@@ -34,8 +39,11 @@ class CinderEntity:
     def get_attributes(self):
         raise NotImplementedError
 
-    def get_context(self):
+    def get_context_generator(self):
         raise NotImplementedError
+
+    def get_empty_context(self):
+        return {'entities': [], 'relationships': []}
 
     def get_entity_data(self):
         return {'entity_type': self.type, 'attributes': self.get_attributes()}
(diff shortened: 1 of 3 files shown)


src/olympia/abuse/models.py
```diff
@@ -134,6 +134,7 @@ def report(cls, abuse_report):
         )
         cinder_job, _ = cls.objects.get_or_create(job_id=job_id)
         abuse_report.update(cinder_job=cinder_job)
+        cls.get_entity_helper(abuse_report).report_additional_context()
```
eviljeff, line 137: should we get the entity helper once and then use it for `report` and `report_additional_context`, or does that mess up the generators?

eviljeff (approved):

src/olympia/abuse/cinder.py
```diff
@@ -101,6 +108,24 @@ def report(self, *, report, reporter):
...
+        next(context_generator, {})
+        for data in context_generator:
+            # Note: Cinder URLS are inconsistent. Per their documentation, that
+            # one needs a trailing slash.
+            url = f'{settings.CINDER_SERVER_URL}graph/'
+            headers = {
+                'accept': 'application/json',
+                'content-type': 'application/json',
+                'authorization': f'Bearer {settings.CINDER_API_TOKEN}',
+            }
```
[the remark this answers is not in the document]
diox: Good idea, c381fb9212b3ebeb2ba295c6c02973b5702b0197

src/olympia/abuse/models.py
diox: I've implemented that in f6992c0938acf8c197c4cd08a928cbcaabb541ef and stored the queryset the generator is using in a property. That way we still get a new generator every time, but the queryset is cached by Django when evaluated, so re-using the same property (computed at entity init time) avoids the duplicated query.

PR #21647, opened 2024-01-02
Increase default caching duration for services.a.m.o. API requests

### LINKED ISSUE #21628: Increase cache duration for services.a.m.o. API requests
Currently, all `GET` non-authenticated API requests default to being cached by the CDN for 180 seconds (with some specific endpoints having a custom higher value). The value was chosen arbitrarily to keep data fresh while still providing enough caching to reduce infrastructure costs.

We'd like to increase that cache duration for API requests on `services.addons.mozilla.org` domain (and `services.addons-dev.allizom.org`, `services.addons-dev.allizom.org`), which serves requests for Firefox. That domain receives a lot of traffic but the freshness of the data for those requests doesn't seem as important.

The main impact would be that information that Firefox pulls from the API to show in about:addons (notably number of ratings, average score) would be slightly less up to date.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-634)

diox:
Fixes #21628

### DIFF
--- a/src/olympia/amo/middleware.py
+++ b/src/olympia/amo/middleware.py
@@ -406,7 +406,11 @@ class CacheControlMiddleware:
             and not max_age_from_response
         )
         if request_conditions and response_conditions:
-            patch_cache_control(response, max_age=settings.API_CACHE_DURATION)
+            if request.META['SERVER_NAME'] == settings.SERVICES_DOMAIN:
+                max_age = settings.API_CACHE_DURATION_SERVICES
+            else:
+                max_age = settings.API_CACHE_DURATION
+            patch_cache_control(response, max_age=max_age)
         elif max_age_from_response is None:
             patch_cache_control(response, s_maxage=0)
         return response
--- a/src/olympia/amo/tests/test_middleware.py
+++ b/src/olympia/amo/tests/test_middleware.py
@@ -336,6 +336,16 @@ class TestCacheControlMiddleware(TestCase):
             )(request)
             assert response['Cache-Control'] == 'max-age=180'
 
+    def test_services_amo_should_cache_for_one_hour(self):
+        request = self.request_factory.get('/api/v5/foo')
+        request.is_api = True
+        request.META['SERVER_NAME'] = settings.SERVICES_DOMAIN
+        for status_code in (200, 201, 202, 204, 301, 302, 303, 304):
+            response = CacheControlMiddleware(
+                lambda x, status=status_code: HttpResponse(status=status)
+            )(request)
+            assert response['Cache-Control'] == 'max-age=3600'
+
     def test_functional_should_cache(self):
         response = self.client.get(reverse_ns('amo-site-status'))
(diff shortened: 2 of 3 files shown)


KevinMind (approved):
LGTM. 🚢 

I didn't test, as I assume it would be somewhat difficult? I suppose you could curl locally and modify the server name? Did you try doing this?

Nit: It would be nice if the PR could include some information about what a reviewer should be looking for, how to test, or that it should/can not be tested etc.

diox:
It is annoying to test locally, yeah: we don't have a different host set up locally for services, everything just uses the same host. I did test it by manually setting `uwsgi_param   SERVER_NAME           olympia.test;` in the `@olympia` location of https://github.com/mozilla/addons-server/blob/master/docker/nginx/addons.conf ; since `settings.SERVICES_DOMAIN` locally is `olympia.test`, that forces all requests to be considered as "services" requests using that snippet of code I use.

diox:
I stole that snippet from existing code but looking at how we define `SERVER_NAME` in dev/stage/prod I am not 100% sure it'll work, as we also have some internal server names to consider sometimes... I'll double-check.

diox:
I pushed https://github.com/mozilla/addons-server/pull/21656 which sadly confirms what I suspected: at the moment, `SERVER_NAME` is not reliable for this.

We'll have to ask SRE to inject some environment variable to reliably detect that the container is a services one.

diox:
https://mozilla-hub.atlassian.net/browse/SVCSE-1676 would fix that.

KevinMind:
Just out of curiosity, have you inspected the headers we actually get from firefox to see if there is some unique value that could be relied on?

Even bolder question, could we ask them to send such a header?

x-cache-ttl or something like that? Could let them determine how fresh they want the data.

I didn't suggest it at first as your solution seemed very simple, but it doesn't seem so anymore. Just a though. I don't have a strong opinion on this.

diox:
Firefox doesn't send any particular headers that would help distinguish that from any other request made to AMO, besides the `Host` header. We could use `request.get_host()`, but I'm not sure we wouldn't run into the same issue.

I would prefer not depending on an extra header, cause then it would also make things more complex as it would need to enter the cache key. We could add a magic query string key (since the full query string is always part of the cache key) but that seems fragile, and we'd still need to ask Firefox to use it, wait for that to ride the trains etc.

Ultimately waiting for SRE to fix `SERVER_NAME` is valuable anyway, if only for consistency (on other domains, like the main a.m.o. one, it's already correct).

diox:
Wei unblocked this on dev:

```
% curl -s https://addons-dev.allizom.org/api/v5/services/client_info | jq .SERVER_NAME
"addons-dev.allizom.org"

% curl -s https://services.addons-dev.allizom.org/api/v5/services/client_info | jq .SERVER_NAME
"services.addons-dev.allizom.org"
```

PR #21656, opened 2024-01-03
Duplicate services/ URLs under /api/v5/services/

diox:
This makes it easier to reach those views for testing purposes, enabling us to reach `client_info/` from `services.a.m.o.` domain (which only accepts specific URL patterns like `/api/`) or test the various 40x/500 handlers from an API perspective.

Supports https://github.com/mozilla/addons-server/issues/21628 (will help test that my implementation for it is correct)

This enables:
- http://olympia.test/api/v5/services/500
- http://olympia.test/api/v5/services/404
- http://olympia.test/api/v5/services/403
- http://olympia.test/api/v5/services/client_info (if `settings.ENV` is `dev`)

### DIFF
--- a/src/olympia/amo/urls.py
+++ b/src/olympia/amo/urls.py
@@ -9,6 +9,13 @@ services_patterns = [
         r'^__heartbeat__$', views.services_heartbeat, name='amo.services_heartbeat'
     ),
     re_path(r'^monitor\.json$', views.services_heartbeat, name='amo.services_monitor'),
+]
+
+shared_services_api_patterns = [
+    # These patterns are duplicated under /services/ and /api/v5/services/,
+    # which is useful to reach those services from an API request, to make them
+    # work with services.a.m.o. or test their behavior with request.is_api
+    # being True.
     re_path(r'^client_info', views.client_info, name='amo.client_info'),
     re_path(r'^403', views.handler403),
     re_path(r'^404', views.handler404),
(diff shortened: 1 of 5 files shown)


src/olympia/amo/urls.py
```diff
@@ -17,6 +17,11 @@
 
 api_patterns = [
     re_path(r'^site/$', views.SiteStatusView.as_view(), name='amo-site-status'),
+    # This duplicates services patterns under /api/vX/services/, which is
+    # useful to reach those services from an API request, to make them work
+    # with services.a.m.o. or test their behavior with request.is_api being
+    # True.
+    re_path(r'^services/', include(services_patterns)),
```
eviljeff, line 24: some of the patterns - the 403, 403, 500 views - are not JSON so should they really be exposed under /api?  (😬)

eviljeff (commented):
can haz docs

src/olympia/amo/urls.py
diox: `404` and `500` are. Only `403` is not. I could implement a quick one.

diox: 44ce9f42fbd206bdf7341d8b1d744e41e17b573f adds docs, limits to the views we care about, adds 403 API support, and tests.

eviljeff (approved):

Issue #21576, opened 2023-12-07
Expose outstanding abuse reports in reviewer tools


eviljeff:
part of https://mozilla-hub.atlassian.net/browse/AMO-114

We need to show the unresolved abuse reports for the add-on in it's review page, and allow a review to resolve those abuse reports.  The reviewer will have to select a Policy to resolve the reports under, so the reasoning can be fed back in a consistent way to the add-on developers.  (Expectation is to re-use the rejection reasons)  

Possibly in a follow-up issue: Emails should be similar to the cinder triggered emails, and include notifications to the reporter(s) too.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-618)

ioanarusiczki:
@eviljeff 

I tested with the "AMO Add-on Escalation" action from Cinder 

For extensions the version reported lands in Manual Review Queue. 
For themes it goes into Updates. 

They're flagged for human review with "Escalated for an abuse report, via cinder" reason. 

A reviewer can take actions such as reject, delay reject, confirm approval, clear needs human review, block or force-disable. 

There is a text below the "Comments box Outstanding DSA related reports to resolve:". I've never seen content after ":" so I am not sure what should be displayed here.

![outstanding dsa related reports](https://github.com/mozilla/addons-server/assets/33448286/e7832bec-a5e1-4ae3-b3b9-32b6b09f5527)


User, collection or rating reports also have action nr7 AMO Add-on Escalation but I'm not sure where these should go on AMO

eviljeff:
> There is a text below the "Comments box Outstanding DSA related reports to resolve:". I've never seen content after ":" so I am not sure what should be displayed here.

mmm, *that's* a problem - there should be a list of checkboxes for all the outstanding `CinderJob`s for this add-on, i.e. at least including the escalation.

eviljeff:
@ioanarusiczki can you link to the add-on review page so I can investigate why the jobs aren't showing up?

ioanarusiczki:
@eviljeff 

At a quick search I found these 2

https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/620562 
https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/629003

eviljeff:
> @eviljeff
> 
> At a quick search I found these 2
> 
[...]

PR #21661, opened 2024-01-04
send job decision to cinder via their api

### LINKED ISSUE #21660: Implement cinder decision resolution from AMO
prerequisite of #21576 (that I thought initially would be trivial enough not to need to land separately).  We need the ability to create a decision on abuse reports that have been processed via amo reviewer tools, and notify cinder of that decision.

It turns out the at a decision on an entity does not automatically resolve a job - and a job can't be specified in the decision - so the job will be closed manually also.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-641)

eviljeff:
fixes #21660 - needed for #21576

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -149,11 +149,6 @@ class CinderEntity:
             # type needs to be defined by subclasses
             raise NotImplementedError
         url = f'{settings.CINDER_SERVER_URL}create_decision'
-        headers = {
-            'accept': 'application/json',
-            'content-type': 'application/json',
-            'authorization': f'Bearer {settings.CINDER_API_TOKEN}',
-        }
         data = {
             'queue_slug': self.queue,
             'entity_type': self.type,
@@ -161,7 +156,7 @@ class CinderEntity:
             'reasoning': review_text,
             'policy_uuids': policy_uuids,
         }
-        response = requests.post(url, json=data, headers=headers)
+        response = requests.post(url, json=data, headers=self.get_cinder_http_headers())
         if response.status_code == 201:
             return response.json().get('uuid')
         else:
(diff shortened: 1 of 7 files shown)


src/olympia/abuse/models.py
```diff
@@ -44,6 +45,10 @@ class CinderJob(ModelBase):
         ('AMO_DELETE_COLLECTION', 6, 'Collection delete'),
         ('AMO_APPROVE', 7, 'Approved (no action)'),
     )
+    DECISION_ACTIONS.add_subset(
+        'UNRESOLVED',
+        ('NO_DECISION', 'AMO_ESCALATE_ADDON'),
```
diox, line 50: That looks weird. Is that what you meant ?

diox (approved):

src/olympia/abuse/models.py
eviljeff: It was - does it not make sense? These are reports that need a decision still.

diox: You're right, misread that

src/olympia/abuse/cinder.py
```diff
@@ -144,6 +144,42 @@ def appeal(self, *, decision_id, appeal_text, appealer):
...
+    def create_decision(self, *, review_text, policy_uuids):
+        if self.type is None:
+            # type needs to be defined by subclasses
+            raise NotImplementedError
+        url = f'{settings.CINDER_SERVER_URL}create_decision'
+        headers = {
+            'accept': 'application/json',
+            'content-type': 'application/json',
+            'authorization': f'Bearer {settings.CINDER_API_TOKEN}',
+        }
```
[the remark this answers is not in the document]
eviljeff: https://github.com/mozilla/addons-server/pull/21661/commits/9d058f219981e88e70e042e10e5c290f1d958aaf

PR #21663, opened 2024-01-05
Add a separate "Something else" reason for new feedback form abuse reports

### LINKED ISSUE #21662: Use separate "Other" category ID for new (DSA) abuse report flow
Between the old and the new abuse report categories, we are only reusing "Other" internally. We should make that a new category ID as well to have a clean cut between pre-DSA and post-DSA.

This issue complements https://github.com/mozilla/addons-frontend/issues/12752.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-642)

diox:
Fixes #21662

It needs to be deployed together with the fix for https://github.com/mozilla/addons-frontend/issues/12752, otherwise new feedback using the `other` category would be considered invalid for non add-on types.


src/olympia/abuse/models.py
```diff
@@ -233,6 +233,7 @@ class AbuseReport(ModelBase):
             'DSA: It violates the law or contains content that violates the law',
         ),
         ('POLICY_VIOLATION', 13, "DSA: It violates Mozilla's Add-on Policies"),
+        ('SOMETHING_ELSE', 14, 'DSA: Something else'),
```
eviljeff, line 236: Should we use a high number like 127 again, given this is another "other" option?

eviljeff (approved):

src/olympia/abuse/models.py
diox: I thought about it but I don't see the point - grouping with other DSA values also has merit...

PR #21664, opened 2024-01-06
expose and resolve reviewer_handled reports in tools

### LINKED ISSUE #21576: Expose outstanding abuse reports in reviewer tools

eviljeff:
Will fix #21576 - and works with the `disable_addon` and `public`/`approve_latest_version` actions already - but is currently flawed:
- Where we show the checkbox to clear abuse reports, and which actions _do_ actually clear abuse reports isn't in sync - i.e. there are some actions that show the checkbox but don't actually clear the reports; and there may be some actions that will try to clear the abuse reports but don't show the checkbox.
- we require a reason to be selected every time the checkbox is checked, but this isn't necessary for AMO_APPROVE decision action - it'll always be the same policy.
- email notifications to the developer don't use the correct templates.  (possibly/probably a follow-up issue)
- email notifications to the reporter(s) aren't sent.  (possibly/probably a follow-up issue)


src/olympia/reviewers/templates/reviewers/review.html
```diff
@@ -188,6 +188,28 @@ <h3 id="history">
...
+          <details>
+            <summary>
+              Show reports
+            </summary>
+            <ul>
+            {% for report in unresolved_abuse_reports %}
+              <li>{{ report.message }}</li>
+            {% endfor %}
+            </ul>
+          </details>
```
diox, line 207: Indent

src/olympia/reviewers/templates/reviewers/review.html
```diff
@@ -188,6 +188,28 @@ <h3 id="history">
...
+          {% else %}
+          No reports to resolve
```
diox, line 209: Indent

src/olympia/reviewers/tests/test_views.py
```diff
@@ -2616,7 +2616,7 @@ def test_item_history_pagination(self):
             str(author.get_role_display()),
             self.addon,
         )
-        with self.assertNumQueries(53):
+        with self.assertNumQueries(54):
```
diox, line 2619: :grimacing:

diox (approved):

src/olympia/reviewers/tests/test_views.py
eviljeff: yeah... the incremental drip drip of extra queries.  We could move some of the information to be be async on page load (like the ADU numbers) - I'm not sure there's much scope for actually reducing the query count without dropping functionality unfortunately.

src/olympia/reviewers/templates/reviewers/review.html
eviljeff: https://github.com/mozilla/addons-server/pull/21664/commits/09a86b181e5cda4c5571005de4e24c87d65a1dc7

Issue #21575, opened 2023-12-06
Implement what's missing for abuse report appeal process (for Cinder decisions)


diox:
Base scenarios to test:
- [x] Content **not** taken down, reporter appeals, we confirm the decision. The content stays up, the reporter can no longer appeal.
- [x] Content **not** taken down, reporter appeals, we change the decision and disable. The content goes down, the author can appeal (the only time an appeal decision can be appealed)
  - [ ] From there, in that second appeal we can change the decision to approve or confirm the decision off the back of the reporter appeal and leave the content disabled. Neither the reporter nor the author can appeal that final decision.
- [x] Content taken down, author appeals, we change the decision and bring the content back up
- [x] Content taken down, author appeals, we confirm the decision
- [x] All scenarios above but with multiple reporters involved: all get the chance to appeal individually until an appeal decision takes place, but in Cinder that would be collapsed into a single job. When the decision is made they can no longer appeal.

In addition, we need to ensure emails to reporters and content authors are sent properly for those cases. https://github.com/mozilla/addons-server/issues/21558 notably should be investigated.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-617)

eviljeff:
Also, afaik, we still include the blurb and link about being able to appeal the decision on emails on the back of their own appeal (i.e. you can't re-appeal, but we say you can).

ioanarusiczki:
@diox 

I think the scenarios above are working, I'll try to highlight a few parts:

At scenario 1,  e-mail is sent: the reporter is notified can no longer appeal

![finalemailscenario1](https://github.com/mozilla/addons-server/assets/33448286/a4b82edc-d97c-4729-9739-fd62182a4f66)

Scenario2 

content is reinstated 

![scenario2reinstate](https://github.com/mozilla/addons-server/assets/33448286/7312854c-c731-4442-b4d8-017dada95f27)

decision is confirmed 

![scenario2confirmsdecision](https://github.com/mozilla/addons-server/assets/33448286/a933cf97-2d18-4224-a3d2-e3848706d1cc)

These were emails that I did not see yet after appeals but now they're sent. 

I did not use "Revert decision" as I tried for https://github.com/mozilla/addons-server/issues/21558 For example, if  I wanted to disable content after an appeal made by the reporter when previously the decision/action was "Ignore", I clicked the Change policy and chose another action and submit. As a result content is taken down from AMO. 

[...]

PR #21676, opened 2024-01-09
Record appeals as their own CinderJob

diox:
This implements the following:
- Allowing appeals by different reporters for the same job, once per reporter. (will get consolidated in a single new appeal job in Cinder, with a single decision)
- Executing decision of the back of an appeal decision (since the appeal is just a new `CinderJob`, its decision is executed like any other by the webhook.
- Allowing appeals by authors off the back of a reporter appeal (again since the appeal is a new `CinderJob`)
- Not including the appeal URL in emails if appeals are not possible for this decision

Fixes https://github.com/mozilla/addons-server/issues/21575


src/olympia/abuse/templates/abuse/emails/reporter_restore.txt
eviljeff, line 1: this email isn't need anyway, according to legal

src/olympia/abuse/urls.py
```diff
@@ -5,6 +5,14 @@
...
-        'appeal/<str:abuse_report_id>/<str:decision_id>/', appeal, name='abuse.appeal'
+        'appeal/<str:decision_id>/',
+        appeal,
+        kwargs={'abuse_report_id': None},
+        name='abuse.appeal_author',
+    ),
+    path(
+        'appeal/<str:abuse_report_id>/<str:decision_id>/',
+        appeal,
+        name='abuse.appeal_reporter',
```
eviljeff, line 16: I feel you can probably do this with a single regex url path but this is fine

src/olympia/abuse/views.py
```diff
@@ -267,45 +267,61 @@ def cinder_webhook(request):
     )
 
 
-def appeal(request, *, decision_id, abuse_report_id, **kwargs):
-    abuse_report = get_object_or_404(AbuseReport.objects, id=abuse_report_id)
+def appeal(request, *, abuse_report_id, decision_id, **kwargs):
+    cinder_job = get_object_or_404(
+        CinderJob.objects.exclude(
+            decision_action=CinderJob.DECISION_ACTIONS.NO_DECISION
```
eviljeff, line 273: we probably want to exclude DECISION_ACTIONS.AMO_ESCALATE_ADDON too (can use `UNRESOLVED`)

eviljeff (approved):

src/olympia/abuse/views.py
diox: It would fail later anyway but I agree.

Any suggestion that reads better than this ? :

```python
appealable_decisions = list(
    CinderJob.DECISION_ACTIONS.APPEALABLE_BY_AUTHOR.values.keys()
) + list(CinderJob.DECISION_ACTIONS.APPEALABLE_BY_REPORTER.values.keys())
cinder_job = get_object_or_404(
    CinderJob.objects.exclude(decision_action__in=appealable_decisions),
    decision_id=decision_id,
)
```

PR #21677, opened 2024-01-09
Activity "Approval notes" to "Notes to Reviewers"

### LINKED ISSUE #21609: "Approval notes changed" log entry doesn't reflect the actual action that took place: Notes for reviewers changed
When "Notes for reviewers" section changes, the version history logs it as "Approval notes changed" in reviewer tools. That doesn't match the reality and may create confusion for reviewers who might think the approval status of the add-on changed (approved -> rejected etc)

When "Notes for reviewers" section updates, a corresponding "Notes for reviewers changed" entry is logged for the version in reviewer tools.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-628)

eviljeff:
fixes #21609


src/olympia/constants/activity.py
```diff
@@ -596,10 +596,10 @@ class REVIEWER_REPLY_VERSION(_LOG):
     review_queue = True
 
 
-class APPROVAL_NOTES_CHANGED(_LOG):
```
KevinMind, line 599: Does this mean that there is no valid use case for `APPROVAL_NOTES_CHANGED`?

KevinMind (approved):
Code looks good to me. There's only one confusing thing which I left a comment for.

NIT: in general I feel like this PR could have use a short description that would have resolved the question preemptively and would additionally make it easier to understand what this PR is all about. Not asking for your to re-write the issue, that's fine. But just a short 1 or 2 sentence explanation of:

1. how did you solve the problem? What changes caused the state to change from pending to fixed.
2. what gotchas are there? Anything in particular I should look at?
3. Does this need testing? How would I test this PR? CI is a valid answer.

PR #21693, opened 2024-01-12
Fix handling of appeals/overrides of appeals in the cinder webhook

### LINKED ISSUE #21575: Implement what's missing for abuse report appeal process (for Cinder decisions)

diox:
Follow-up fixes #21575

### DIFF
--- a/src/olympia/abuse/tests/assets/cinder_webhook.json
+++ b/src/olympia/abuse/tests/assets/cinder_webhook.json
@@ -1,38 +1,6 @@
 {
     "event": "decision.created",
     "payload": {
-        "appeal": {
-            "appealed_decision": {
-                "enforcement_actions": [
-                    "amo-disable-addon"
-                ],
-                "id": "ae106e36-700c-4668-9618-851451dc5c2c",
-                "notes": "",
-                "policies": [
-                    {
-                        "enforcement_actions": [
-                            "amo-disable-addon"
-                        ],
-                        "id": "f73ad527-54ed-430c-86ff-80e15e2a352b",
-                        "is_illegal": false,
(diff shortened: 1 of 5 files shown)


src/olympia/abuse/tests/assets/cinder_webhook.json
```diff
@@ -1,38 +1,6 @@
 {
     "event": "decision.created",
     "payload": {
-        "appeal": {
```
diox, line 4: This only makes sense in appeals, and this json is used to test "original" decisions.

eviljeff (approved):

Issue #21705, opened 2024-01-15
Use CinderAction emails to notify developers about reviewer tools originated decisions


eviljeff:
Currently we send out different emails when an add-on is disabled, etc, depending on whether the action was by a cinder moderator via Cinder, or by an AMO reviewer via the reviewer tools.    Duplication leads to inconsistency - even if we update the reviewer tools emails to be compliant and include the appeal link, etc. - so it would be the best case to send the same emails in both scenarios.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-652)

ioanarusiczki:
@eviljeff 


When extensions/themes are disabled from rev tools with Reject there is a default text which is going into the email (can be manually removed by a reviewer)

![default text](https://github.com/mozilla/addons-server/assets/33448286/f8d71778-31e4-47ae-a27b-d700c293397e)

This is present in the email and has some layout issues too. 

![revtoolsdisabling](https://github.com/mozilla/addons-server/assets/33448286/7ce4a3f7-f97c-4127-bc3e-c7e314020e98)

Themes when disabled from rev tools , email has "affected versions". That line is not present when theme is disabled from Cinder. 

![rejectedtheme](https://github.com/mozilla/addons-server/assets/33448286/889130f4-2981-409e-909a-b6297c7c223c)

Disabled with Cinder policies email sent to developer 

![themedisabledcinder](https://github.com/mozilla/addons-server/assets/33448286/fcb7691f-948f-4d72-b876-045d632449ce)

ioanarusiczki:
@eviljeff Themes disabled from rev tools also have 
"You may upload a new version which addresses the policy violation(s)." which is not present in Cinder e-mails sent to developers.

eviljeff:
@ioanarusiczki the default review-text and mismatched copy for reviewer tools canned responses is being fixed in https://github.com/mozilla/addons-server/issues/21816

I think you may be comparing an add-on disable (the only open from Cinder) with a version rejection from the reviewer tools - a Force disable action should have the same copy (apart from the policy/review-text section)

ioanarusiczki:
@eviljeff Oh I see, I checked with force-disable for themes and they are identical indeed (leaving policies aside).

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-652

PR #21707, opened 2024-01-15
Add AMO_REJECT_VERSION_ADDON to CinderJob.DECISION_ACTIONS

### LINKED ISSUE #21671: Add additional decision actions for reviewer tools specific actions
We currently just have Disable Add-on as an action that can resolve cinder reports, but in the reviewer tools there are other actions that could/should be taken on the back of a report, such as rejecting specific version(s).  We need to add support for recording those in `CinderJob.decision` so they can be shown as resolving abuse reports too.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-645)

eviljeff:
fixes #21671 - though there will be more changes once [emails are aligned](https://github.com/mozilla/addons-server/issues/21705).

### DIFF
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -45,6 +45,7 @@ class CinderJob(ModelBase):
         ('AMO_DELETE_RATING', 5, 'Rating delete'),
         ('AMO_DELETE_COLLECTION', 6, 'Collection delete'),
         ('AMO_APPROVE', 7, 'Approved (no action)'),
+        ('AMO_REJECT_VERSION_ADDON', 8, 'Add-on version reject'),
     )
     DECISION_ACTIONS.add_subset(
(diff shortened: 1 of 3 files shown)


src/olympia/abuse/models.py
```diff
@@ -53,6 +54,7 @@ class CinderJob(ModelBase):
             'AMO_DISABLE_ADDON',
             'AMO_DELETE_RATING',
             'AMO_DELETE_COLLECTION',
+            'AMO_REJECT_VERSION_ADDON',
```
diox, line 57: Add a comment explaining that this action (and its appeal) is stricly handled by reviewers on AMO and not by moderators on Cinder, so it should never show up in the webhook payload

diox (approved):

PR #21711, opened 2024-01-16
chore(github): update PR template.

KevinMind:
Thanks for opening a Pull Request (PR), here's a few guidelines as to what we need in your PR before we review it.

Fixes: N/A

## How did you solve the problem?

I updated the github PR template to include specific sections for required information and clear descriptions. Each section includes examples and links to help the PR author and specifies which sections are required.

## What should the reviewer focus on?

If we want to maintain a separate pull request template for community PRs, that is also [possible](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository). We can create a directory of PR templates and open them with query parameters or selecting in the Web UI.

If we want to extend this template to other addon repositories, we [could](https://www.rhysmills.com/post/2021/09/07/set-a-default-pr-template-for-a-github-organisation.html) do so, but would require moving the addon repos to a separate org. Probably not worth it.

## How to test this PR?

No testing required.

PR #21714, opened 2024-01-17
Fix SERVICES_DOMAIN to be correct on dev/stage/prod

### LINKED ISSUE #21628: Increase cache duration for services.a.m.o. API requests

diox:
Fixes #21628

`SERVICES_DOMAIN` was only set in `olympia/lib/settings_base.py`, so it was evaluated early, end up with the domain set by:
https://github.com/mozilla/addons-server/blob/6c4fc0d2fc4198f154f538f950fa7f86ea37a41f/src/olympia/lib/settings_base.py#L197-L204

Which means that on dev for instance, it was `services.addons-server-v1-deploy-uwsgi-amo-web-<id>`, which obviously is completely wrong. `SERVICES_URL` _was_ overridden in the relevant settings files and so was correct.

This doesn't impact production code (both settings are mostly unused - there is one check in `robots` function that was broken, but it's unreachable as nginx doesn't route `robots.txt` for the services domain, and the other use is what I want to fix here, [the recent introduction of a specific caching duration for the services domain](https://github.com/mozilla/addons-server/issues/21628) which isn't in prod yet).

### DIFF
--- a/src/olympia/conf/dev/settings.py
+++ b/src/olympia/conf/dev/settings.py
@@ -17,11 +17,12 @@ RAISE_ON_SIGNAL_ERROR = True
 API_THROTTLING = True
 
 DOMAIN = env('DOMAIN', default='addons-dev.allizom.org')
+SERVICES_DOMAIN = env('SERVICES_DOMAIN', default='services.addons-dev.allizom.org')
 SERVER_EMAIL = 'zdev@addons.mozilla.org'
 SITE_URL = 'https://' + DOMAIN
 INTERNAL_SITE_URL = env('INTERNAL_SITE_URL', default='https://addons-dev.allizom.org')
 EXTERNAL_SITE_URL = env('EXTERNAL_SITE_URL', default='https://addons-dev.allizom.org')
-SERVICES_URL = env('SERVICES_URL', default='https://services.addons-dev.allizom.org')
+SERVICES_URL = 'https://' + SERVICES_DOMAIN
 CODE_MANAGER_URL = env(
     'CODE_MANAGER_URL', default='https://code.addons-dev.allizom.org'
 )
--- a/src/olympia/conf/prod/settings.py
+++ b/src/olympia/conf/prod/settings.py
@@ -26,10 +26,11 @@ API_THROTTLING = True
 
 DOMAIN = env('DOMAIN', default='addons.mozilla.org')
 SERVER_EMAIL = 'zprod@addons.mozilla.org'
+SERVICES_DOMAIN = env('SERVICES_DOMAIN', default='services.addons.mozilla.org')
 SITE_URL = 'https://' + DOMAIN
 INTERNAL_SITE_URL = env('INTERNAL_SITE_URL', default='https://addons.mozilla.org')
 EXTERNAL_SITE_URL = env('EXTERNAL_SITE_URL', default='https://addons.mozilla.org')
-SERVICES_URL = env('SERVICES_URL', default='https://services.addons.mozilla.org')
+SERVICES_URL = 'https://' + SERVICES_DOMAIN
 CODE_MANAGER_URL = env('CODE_MANAGER_URL', default='https://code.addons.mozilla.org')
 STATIC_URL = PROD_STATIC_URL
 MEDIA_URL = PROD_MEDIA_URL
--- a/src/olympia/conf/stage/settings.py
+++ b/src/olympia/conf/stage/settings.py
@@ -16,11 +16,12 @@ ENV = env('ENV')
 API_THROTTLING = True
 
 DOMAIN = env('DOMAIN', default='addons.allizom.org')
+SERVICES_DOMAIN = env('SERVICES_DOMAIN', default='services.addons.allizom.org')
 SERVER_EMAIL = 'zstage@addons.mozilla.org'
 SITE_URL = 'https://' + DOMAIN
 INTERNAL_SITE_URL = env('INTERNAL_SITE_URL', default='https://addons.allizom.org')
 EXTERNAL_SITE_URL = env('EXTERNAL_SITE_URL', default='https://addons.allizom.org')
-SERVICES_URL = env('SERVICES_URL', default='https://services.addons.allizom.org')
+SERVICES_URL = 'https://' + SERVICES_DOMAIN
 CODE_MANAGER_URL = env('CODE_MANAGER_URL', default='https://code.addons.allizom.org')
 STATIC_URL = '%s/static-server/' % EXTERNAL_SITE_URL
 MEDIA_URL = '%s/user-media/' % EXTERNAL_SITE_URL


diox:
Those settings files are not covered by tests. They can be tested manually by doing, in a `web` container:
```
$ echo 'from olympia.conf.dev.settings import *' > settings_specific.py
$ DEVELOPMENT_FXA_CLIENT_SECRET=S DEVELOPMENT_FXA_CLIENT_ID=X ENV=dev EMAIL_URL='smtp://example.com' DJANGO_SETTINGS_MODULE=settings_local ./manage.py shell_plus
```

And then checking whatever `settings` property you care about. Same for `stage` and `prod`.

src/olympia/conf/dev/settings.py
```diff
@@ -17,11 +17,12 @@
 API_THROTTLING = True
 
 DOMAIN = env('DOMAIN', default='addons-dev.allizom.org')
+SERVICES_DOMAIN = env('SERVICES_DOMAIN', default='services.addons-dev.allizom.org')
 SERVER_EMAIL = 'zdev@addons.mozilla.org'
 SITE_URL = 'https://' + DOMAIN
 INTERNAL_SITE_URL = env('INTERNAL_SITE_URL', default='https://addons-dev.allizom.org')
 EXTERNAL_SITE_URL = env('EXTERNAL_SITE_URL', default='https://addons-dev.allizom.org')
-SERVICES_URL = env('SERVICES_URL', default='https://services.addons-dev.allizom.org')
```
diox, line 24: `SERVICES_URL` environment variable is no longer used as part of this but it was redundant anyway. Still, to keep things consistent, I've introduced `SERVICES_DOMAIN` env variable in case we need that escape hatch.

src/olympia/conf/dev/settings.py
```diff
@@ -17,11 +17,12 @@
...
+SERVICES_URL = 'https://' + SERVICES_DOMAIN
```
KevinMind, line 25: I'm surprised that python doesn't seem to have support for generating URLs including the sechema? Feels wrong to use simple string concat here, especially since it is split across files.

Can you think of a neater solution?

KevinMind:
Flagging as an example of why we should [update the PR template](https://github.com/mozilla/addons-server/pull/21711).

I think this is a perfect example of how the PR template suggestion I made would work in reality.

- You actually added the test information in a comment, exactly what the test section is meant for 💪 
- There was a lot of information in the description, but it was actually still missing an explanation of exactly the change that you made. I gave an example of a short description in the example below.
- Some of what you put in the description could actually have been in the what to focus on section. You could have even put less had there been a concise description.

 ---- ---- ---- ---- ---- ---- ----   ---- **EXAMPLE BELOW** ---- ---- ---- ---- ---- ---- ---- ----

Fixes https://github.com/mozilla/addons-server/issues/21628

## How did you solve the problem?

Explicitly set SERVICES_URL settings based on the value an existing environment variable "SERVICES_DOMAIN" in order to prevent using the explicitly set "SERVICES_URL" which is not actually what we want to use.

## What should the reviewer focus on?

`SERVICES_DOMAIN` was only set in `olympia/lib/settings_base.py`, so it was evaluated early, end up with the domain set by:

https://github.com/mozilla/addons-server/blob/6c4fc0d2fc4198f154f538f950fa7f86ea37a41f/src/olympia/lib/settings_base.py#L197-L204

Which means that on dev for instance, it was `services.addons-server-v1-deploy-uwsgi-amo-web-<id>`, which obviously is completely wrong. `SERVICES_URL` _was_ overridden in the relevant settings files and so was correct.

This doesn't impact production code (both settings are mostly unused - there is one check in `robots` function that was broken, but it's unreachable as nginx doesn't route `robots.txt` for the services domain, and the other use is what I want to fix here, [the recent introduction of a specific caching duration for the services domain](https://github.com/mozilla/addons-server/issues/21628) which isn't in prod yet).

## How to test this PR?

Those settings files are not covered by tests. They can be tested manually by doing, in a web container:

```bash
$ echo 'from olympia.conf.dev.settings import *' > settings_specific.py
$ DEVELOPMENT_FXA_CLIENT_SECRET=S DEVELOPMENT_FXA_CLIENT_ID=X ENV=dev EMAIL_URL='smtp://example.com' DJANGO_SETTINGS_MODULE=settings_specific ./manage.py shell_plus
```

And then checking whatever settings property you care about. Same for stage and prod. Delete the settings_specific.py file afterwards.

src/olympia/conf/dev/settings.py
```diff
@@ -17,11 +17,12 @@
 API_THROTTLING = True
 
 DOMAIN = env('DOMAIN', default='addons-dev.allizom.org')
+SERVICES_DOMAIN = env('SERVICES_DOMAIN', default='services.addons-dev.allizom.org')
```
KevinMind, line 20: Are we actually setting environment variables for "SERVICES_DOMAIN" and for "SERVICES_URL"

I see that before we read the SERVICES_URL directly, so assuming it is set, but just incorrect, per your description. But are we also setting the `SERVICES_DOMAIN` env or will we always use the default?

Is the SERVICES_DOMAIN always just `services.{DOMAIN}`, if so why not just set that for the SERVICES_URL since it is already being formatted anyway.

Are we actually using the SERVICES_DOMAIN setting anywhere? I ask because it did not exist in the file before? If it is only used in order to be added to SERVICES_URL and given my previous statement, wouldn't it be cleaner to just set

```python
SERVICES_URL = f"https://services.{DOMAIN}"
```

Or is domain also incorrect based on the environment variables?

Finally, why is the SERVICES_URL incorrect in the environment variables? Could we fix the problem there and remove the need for this PR entirely?

diox: - Currently, we are only passing `SERVICES_URL` environment variable, not `SERVICES_DOMAIN`. As mentioned below I'm introducing `SERVICES_DOMAIN` as an environment variable and removing the support for `SERVICES_URL`.

- `SERVICES_URL` was/is correct. The problem was purely `SERVICES_DOMAIN`, which was only set in `lib/settings_base.py` without being based on an environment variable at all as explained in the PR description.

- As I said in the description, both settings are barely used, the only real impact is for the caching duration change which does require `SERVICES_DOMAIN`.

I could change to `SERVICES_DOMAIN = env('SERVICES_DOMAIN', default=f'services.{DOMAIN}')` if you prefer.

src/olympia/conf/dev/settings.py
diox: IMHO it doesn't matter at all here. Could use `f` strings if you prefer, I went with the same style used in the file already for `SITE_URL`. Using the standard library we could do `urlunsplit(('https', SERVICES_DOMAIN, '', '', ''))` which I don't think is an improvement (or `urlunparse()`, but same problem: we would need to path empty strings for the `path`, `query` and `hash` components of the URL)

KevinMind (approved):
LGTM

PR #21720, opened 2024-01-17
Fix appeals from content authors (they don't pass an abuse report)

diox:
Fixes https://github.com/mozilla/addons-server/issues/21575 (again)

### DIFF
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -94,11 +94,17 @@ class CinderJob(ModelBase):
     def target(self):
         # this works because all abuse reports for a single job and all appeals
         # for a single job are for the same target.
+        if initial_abuse_report := self.initial_abuse_report:
+            return initial_abuse_report.target
+        return None
+
+    @property
+    def initial_abuse_report(self):
         if first_abuse_report := self.abusereport_set.first():
             # Early return to avoid extra queries.
-            return first_abuse_report.target
+            return first_abuse_report
         appealed_job = self.appealed_jobs.first()
-        return appealed_job.target if appealed_job else None
+        return appealed_job.initial_abuse_report if appealed_job else None
 
     @property
     def abuse_reports(self):
@@ -261,6 +267,11 @@ class CinderJob(ModelBase):
     def appeal(self, *, abuse_report, appeal_text, user, is_reporter):
         if not self.can_be_appealed(is_reporter=is_reporter, abuse_report=abuse_report):
             raise CantBeAppealed
+        # Author appeals can be done without an abuse report. Unfortunately
+        # though, at the moment we still need an abuse_report to call
+        # get_entity_helper(), so we grab the first one.
+        if abuse_report is None:
+            abuse_report = self.initial_abuse_report
         if user:
             appealer = CinderUser(user)
         else:
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -266,8 +266,6 @@ class CinderJob(ModelBase):
         self.get_action_helper(existing_decision, override=override).process()
 
     def appeal(self, *, abuse_report, appeal_text, user, is_reporter):
-        if not self.can_be_appealed(is_reporter=is_reporter, abuse_report=abuse_report):
-            raise CantBeAppealed
         appealer_entity = None
         if is_reporter:
             if not abuse_report:
@@ -293,6 +291,10 @@ class CinderJob(ModelBase):
                 abuse_report = self.initial_abuse_report
         if user:
             appealer_entity = CinderUser(user)
+
+        if not self.can_be_appealed(is_reporter=is_reporter, abuse_report=abuse_report):
+            raise CantBeAppealed
+
         appeal_id = self.get_entity_helper(abuse_report).appeal(
             decision_id=self.decision_id,
             appeal_text=appeal_text,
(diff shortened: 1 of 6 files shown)


src/olympia/abuse/models.py
```diff
@@ -261,6 +267,11 @@ def process_decision(
     def appeal(self, *, abuse_report, appeal_text, user, is_reporter):
         if not self.can_be_appealed(is_reporter=is_reporter, abuse_report=abuse_report):
             raise CantBeAppealed
+        # Author appeals can be done without an abuse report. Unfortunately
+        # though, at the moment we still need an abuse_report to call
+        # get_entity_helper(), so we grab the first one.
+        if abuse_report is None:
+            abuse_report = self.initial_abuse_report
```
eviljeff, line 274: mmm, is this safe though - we also use `abuse_report` for `get_cinder_reporter` if `not user` and additionally then update the `abuse_report` instance when `is_reporter`.  From carefully reading through it does looks like `abuse_report` won't be accessed in those cases if `abuse_report` is `None` but also, imo, we're risking a bug the next time this function (or upstream) gets slightly refactored.

Maybe `self.get_entity_helper(abuse_report or self.initial_abuse_report)` so it won't be used by accident?

eviljeff (approved):

src/olympia/abuse/tasks.py
```diff
@@ -58,9 +58,7 @@ def flag_high_abuse_reports_addons_according_to_review_tier():
 @task
 @use_primary_db
 def report_to_cinder(abuse_report_id):
-    abuse_report = AbuseReport.objects.filter(id=abuse_report_id).first()
-    if not abuse_report:
-        return
+    abuse_report = AbuseReport.objects.get(pk=abuse_report_id)
```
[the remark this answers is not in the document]
diox: Yeah, that was the point. I didn't want to fail silently in that case.

src/olympia/abuse/models.py
diox: I'll implement that suggestion yeah.

Ultimately it will need to be refactored to allow omitting the `abuse_report` entirely to forward reviewer decisions to Cinder that are not made off the back of an abuse report...

diox: Alternative proposal (on top of the PR) to make the method more robust:
```diff
diff --git a/src/olympia/abuse/models.py b/src/olympia/abuse/models.py
index 19c6832ae1..8a37e0b867 100644
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -1,6 +1,7 @@
 from datetime import datetime, timedelta
 from itertools import chain
 
+from django.core.exceptions import ImproperlyConfigured
 from django.db import models
 from django.db.models import Q
 from django.db.transaction import atomic
@@ -267,17 +268,35 @@ class CinderJob(ModelBase):
     def appeal(self, *, abuse_report, appeal_text, user, is_reporter):
         if not self.can_be_appealed(is_reporter=is_reporter, abuse_report=abuse_report):
             raise CantBeAppealed
-        # Author appeals can be done without an abuse report. Unfortunately
-        # though, at the moment we still need an abuse_report to call
-        # get_entity_helper(), so we grab the first one.
-        if abuse_report is None:
-            abuse_report = self.initial_abuse_report
-        if user:
-            appealer = CinderUser(user)
+        appealer_entity = None
+        if is_reporter:
+            if not abuse_report:
+                raise ImproperlyConfigured(
+                    'CinderJob.appeal() called with is_reporter=True without an '
+                    'abuse_report'
+                )
+            if not user:
+                appealer_entity = self.get_cinder_reporter(abuse_report)
         else:
-            appealer = self.get_cinder_reporter(abuse_report)
+            if not user:
+                # If the appealer is not an original reporter, we have to
+                # provide an authenticated user that is the author of the
+                # content.
+                raise ImproperlyConfigured(
+                    'CinderJob.appeal() called with is_reporter=False without a user'
+                )
+            if not abuse_report:
+                # Author appeals can be done without an abuse report.
+                # Unfortunately though, at the moment we still need an
+                # abuse_report to call get_entity_helper(), so we grab the
+                # first one.
+                abuse_report = self.initial_abuse_report
+        if user:
+            appealer_entity = CinderUser(user)
         appeal_id = self.get_entity_helper(abuse_report).appeal(
-            decision_id=self.decision_id, appeal_text=appeal_text, appealer=appealer
+            decision_id=self.decision_id,
+            appeal_text=appeal_text,
+            appealer=appealer_entity,
         )
         with atomic():
             appeal_job, _ = self.__class__.objects.get_or_create(job_id=appeal_id)
```

What do you think ?

eviljeff: It's more robust... And makes some of the hidden complexity/assumptions clear. Go for it.

src/olympia/abuse/models.py
```diff
@@ -261,6 +267,11 @@ def process_decision(
     def appeal(self, *, abuse_report, appeal_text, user, is_reporter):
         if not self.can_be_appealed(is_reporter=is_reporter, abuse_report=abuse_report):
             raise CantBeAppealed
+        # Author appeals can be done without an abuse report. Unfortunately
```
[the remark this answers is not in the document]
diox: That only moves the problem to the task (which I'd like to avoid - the tasks are fairly lightweight, not much code, with the logic in the `CinderJob` methods, and it's probably best to keep them that way)

PR #21728, opened 2024-01-19
Update PULL_REQUEST_TEMPLATE.md

eviljeff:
for https://github.com/mozilla/addons-server/pull/21711 - doesn't needing reviewing separately.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,46 +3,21 @@ Thanks for opening a Pull Request (PR), here's a few guidelines as to what we ne
 <!-- REQUIRED  https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue -->
 Fixes: `<issue-number>`
 
-## How did you solve the problem?
+## Are there more details and context that will help us understand and review this pull request?
 
 <!-- REQUIRED
 
-Short description of the changes you made and how they address the linked issue.
+Further description of the changes you made if the pull request title is insufficient, and how they address the linked issue.
 This should be written directly in the PR so reviewers and testers understand a basic overview of the change.
 It also ensures this PR is self documenting to viewers from the future.
+If appropriate, help the reviewer by pointing to parts of the PR that might need special attention or where you would like explicit feedback.
 
 -->
-
-## What should the reviewer focus on?
-
-<!-- OPTIONAL
-
-Help the reviewer by pointing to parts of the PR that might need special attention or where you would like explicit feedback.
-
--->
-
-## How to test this PR?
-
-<!-- REQUIRED
-
-Include explicit steps to test this PR. Valid answers include:
-- filling the task list below with steps to execute the code and expected outcomes.
-- tested in CI, no explicit testing needed (for PRs where this is true)
-- link to test scenarios in the issue or some other document. No need to repeat yourself
-
--->
-
-### Test template (replace with a block for each scenario)
-
-1. Open http://olympia.test
-2. Navigate to X
-3. Click Button
-
-**expected**: You see a notification telling you the current time.
 
 ## Check list
 
 - [ ] This PR relates to an existing open issue and there are no existing PRs open for the same issue.
 - [ ] The change has been successfully run locally.
 - [ ] Add tests to cover the changes added in this PR.
+- [ ] Add a comment to the issue that describes how to manually test the changes you have made in this PR.
 - [ ] Add before and after screenshots (Only for changes that impact the UI).
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,7 +5,7 @@ Fixes: `<issue-number>`
 
 ## Are there more details and context that will help us understand and review this pull request?
 
-<!-- REQUIRED
+<!-- OPTIONAL
 
 Further description of the changes you made if the pull request title is insufficient, and how they address the linked issue.
 This should be written directly in the PR so reviewers and testers understand a basic overview of the change.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,46 +3,21 @@ Thanks for opening a Pull Request (PR), here's a few guidelines as to what we ne
 <!-- REQUIRED  https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue -->
 Fixes: `<issue-number>`
 
-## How did you solve the problem?
+## Are there more details and context that will help us understand and review this pull request?
 
 <!-- REQUIRED
```
diox, line 8: Shouldn't be "REQUIRED"

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,46 +3,21 @@ Thanks for opening a Pull Request (PR), here's a few guidelines as to what we ne
...
 
-Short description of the changes you made and how they address the linked issue.
+Further description of the changes you made if the pull request title is insufficient, and how they address the linked issue.
```
diox, line 10: Maybe we should have a separate paragraph about making good titles

.github/PULL_REQUEST_TEMPLATE.md
eviljeff, line 8:
```suggestion
<!-- OPTIONAL
```

.github/PULL_REQUEST_TEMPLATE.md
eviljeff: as one of the checklist items?

diox: I don't have a strong opinion on this. @KevinMind any thoughts ?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,46 +3,21 @@ Thanks for opening a Pull Request (PR), here's a few guidelines as to what we ne
...
-3. Click Button
-
-**expected**: You see a notification telling you the current time.
-
 ## Check list
 
 - [ ] This PR relates to an existing open issue and there are no existing PRs open for the same issue.
 - [ ] The change has been successfully run locally.
 - [ ] Add tests to cover the changes added in this PR.
+- [ ] Add a comment to the issue that describes how to manually test the changes you have made in this PR.
```
KevinMind, line 22: I would push back on this. If this information is needed, it should just be in the description. Adding as a comment is just splitting information in two places and I'd rather avoid that.

If the info is not needed, it just isn't added, if it is needed it's added in the description. Wdyt?

.github/PULL_REQUEST_TEMPLATE.md
KevinMind: I think it makes sense especially given how we are implicitly expecting the title to be the longest living and most prominent part of the PR in terms of future references.

I've experimented with a simpler syntax in a few PRs, wdyt?

1. Description
2. Context
3. Testing

All optional (begrudgingly) wdyt? I'd like to keep the titles short, even shorter than I originally conceived them.

Ex:
- https://github.com/mozilla/addons-server/pull/21787
- https://github.com/mozilla/addons-server/pull/21784
- https://github.com/mozilla/addons-server/pull/21773
- https://github.com/mozilla/addons-server/pull/21734

@diox could you add a suggestion in the original PR for a section on titles. I think that would be good to explain in the PR template that when you merge the title of the PR will be the commit and how it should be formatted. All great context for a contributor. 

IF we agree on the simplified section names then I would add a commit on the original PR and then @eviljeff could you add your changes as suggestions? Then we will have a finallized version on the original PR and can squash and merge it.

.github/PULL_REQUEST_TEMPLATE.md
diox: The thinking here is that ultimately, testing should be done by QA to verify an issue has been fixed. QA should never need to look at PRs, so if you have manual testing steps (*)... they really should be in the issue.


(*) Which should be the exception and not the rule, for PRs. We want everything to be covered by tests ran by CI anyway.

eviljeff: yeah, this was about the instructions for QA to test the patch, which should be in the issue.  Agree that if there are instructions for the _reviewer_ they should be in the description.

diox: Maybe reword this with something like "Make sure the issue has steps to reproduce or a description of how to test it" ? Or do we want this out of this template and word something here differently for any (optional) supplemental manual testing that would be required for the PR itself ?

KevinMind: The problem that I have with this approach is that we currently only QA on master. So a reviewer is basically reading the code, and potentially looking at the automated tests.

This seems like a total misordering of the steps. I might be resisting the overall dev -> review -> merge -> QA flow and that might be why we are not aligned on where the QA / Tests steps should go.

In principle, I think it is a good idea to have the high level QA test scenarios (navigate to this page, click this button, expect "foo") on the issue. That is fine.

But ... should the reviewer need to run those steps theirself? How can we be happy merging a PR that we have not played around with and verified, at least to a minimum degree. And the question is, are the steps you give to QA for "full" testing the same as you give to a reviewer for "minimal" testing? I don't have an answer for that but it seems relevant.

I would be.... somewhat okay..  adding this as a check that "The issue linked in `fixes` has verifiable steps to QA this PR <link to how to write test cases>" Assuming we have a place where we document how to write test cases

The trick here is, sometimes merging a PR is not completely finishing a feature. Sometimes it's just one part. How do we handle those cases? That is why I had originally in mind a test section which is "how to test this chunk of the feature that maybe be anywhere from <0-1 / 1 of the total feature scope.

Thoughts?

PR #21729, opened 2024-01-19
Implement throttling on abuse appeal form

### LINKED ISSUE #21302: Add rate limiting to abuse appeal form
Abuse appeal form should be rate-limited to a reasonable value (20 / day like reporting? more?), both for `GET` and `POST`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-522)

diox:
This reuses the existing implementation that we have in the `NewUploadForm`, refactoring it in a generic mixin that no longer depends on the viewset being passed, we pass the list of throttle classes instead.

It purposefully changes the existing behavior to prevent further validation from taking place, to avoid potential credentials stuffing : we don't want to validate data (which, in the case of appeals, could be the email of a reporter) when rate-limiting the user. It shouldn't have an impact for `NewUploadForm` though.

Fixes #21302

### DIFF
--- a/src/olympia/abuse/forms.py
+++ b/src/olympia/abuse/forms.py
@@ -1,4 +1,5 @@
 from django import forms
+from django.core.exceptions import ImproperlyConfigured
 from django.utils.translation import gettext_lazy as _
 
 from olympia.api.throttling import (
@@ -45,14 +46,15 @@ class AbuseAppealEmailForm(CheckThrottlesFormMixin, forms.Form):
 
     def __init__(self, *args, **kwargs):
         self.request = kwargs.pop('request')
-        self.expected_email = kwargs.pop('expected_email', None)
+        self.expected_email = kwargs.pop('expected_email')
+        if not self.expected_email:
+            raise ImproperlyConfigured(
+                'AbuseAppealEmailForm called without an expected_email'
+            )
         return super().__init__(*args, **kwargs)
 
     def clean_email(self):
-        if (
-            self.expected_email is None
-            or (email := self.cleaned_data['email']) != self.expected_email
-        ):
+        if (email := self.cleaned_data['email']) != self.expected_email:
             raise forms.ValidationError(_('Invalid email provided.'))
         return email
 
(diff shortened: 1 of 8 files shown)


src/olympia/abuse/forms.py
```diff
@@ -1,23 +1,79 @@
...
     def __init__(self, *args, **kwargs):
-        self.expected_email = kwargs.pop('expected_email')
+        self.request = kwargs.pop('request')
+        self.expected_email = kwargs.pop('expected_email', None)
         return super().__init__(*args, **kwargs)
 
     def clean_email(self):
-        if (email := self.cleaned_data['email']) != self.expected_email:
+        if (
+            self.expected_email is None
```
eviljeff, line 53: Given this can only happen when the kwarg to `__init__` is missing or `None` shouldn't we be erroring further up or raising as a server error?

src/olympia/api/throttling.py
```diff
@@ -99,6 +100,38 @@ def allow_request(self, request, view):
...
+    Mixin to add to a django form to check DRF throttles when running clean().
+
+    Requires self.request and self.throttle_classes to be set on the form
+    instance. self.throttled_error_message can be used to customize the error
+    message on throttled requests.
+    """
+
+    throttled_error_message = _(
+        'You have submitted too many uploads recently. '
+        'Please try again after some time.'
```
eviljeff, line 114: nit: this is, imo, an odd default message for a re-usable class to be used for django forms - the message form the abuse form is more generic.

eviljeff (approved):

src/olympia/abuse/forms.py
diox: Yeah, let's do that 75d0dfb7039b711a8b059430f63e65d438448adb

src/olympia/api/throttling.py
diox: Totally agree, cf68c6c54b944edc6e52f4a0ee63922919e502ea

PR #21734, opened 2024-01-23
Add localization job to CircleCI config

KevinMind:
Relates to #9112 

### Description

This PR "extracts" the logic for extracting and merging l10n strings to an executable script that can be run in CI. 

### Context:

Previously the script was designed to run locally and did more than extract, this script just extracts and can be a part of a job that synchronizes our l10n automatically.

This PR does not implement the automation, but just the extraction. A follow up PR will include automatically committing the updated locales once we have proven it works as expected.

In the future, we can consider executing pybabel via [setup.py](https://babel.pocoo.org/en/stable/setup.html) or directly [calling the functions](https://babel.pocoo.org/en/latest/api/messages/extract.html#babel.messages.extract.extract_from_dir) in a django command, but implementing this was not straight forward and thus outside of scope.

### Testing:

You can test this script in CI by checking the "localization" job. You can also run the script locally running:

```bash
make extract_locales
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -493,7 +493,6 @@ commands:
       - run:
           name: Set environment variables
           command: |
-            echo export PYTHONPATH=src >> $BASH_ENV
             echo export CPUCOUNT=2 >> $BASH_ENV
             echo export NPM_CONFIG_PREFIX=/deps/ >> $BASH_ENV
             echo export CC=\"`python -c 'import sysconfig; print(sysconfig.get_config_var("CC"))'`\" >> $BASH_ENV
@@ -574,6 +573,17 @@ jobs:
       - run: make install_python_codestyle_dependencies
       - run: pyenv rehash
       - run: make lint-codestyle
+
+  localization:
+    <<: *defaults
+    steps:
+      - setup_container:
+          wait_services: false
+          install_node_dependencies: true
+          install_python_dev_dependencies: true
+      - run : make extract_locales
+      - run: git diff
+
 
   devhub:
     <<: *defaults-with-services
@@ -666,6 +676,7 @@ workflows:
       - main
       - reviewers-and-zadmin
       - es-tests
+      - localization
       - release-master:
           filters:
             branches:
--- a/Makefile-docker
+++ b/Makefile-docker
@@ -80,12 +80,17 @@ cleanup_python_build_dir:
 	# Work arounds "Multiple .dist-info directories" issue.
 	rm -rf /deps/build/*
 
+.PHONY: install_python_olympia_module
+install_python_olympia_module:
+	# pep 517 mode (the default) breaks editable install in our project. https://github.com/mozilla/addons-server/issues/16144
+	$(PIP_COMMAND) install --no-use-pep517 -e .
+
 .PHONY: install_python_codestyle_dependencies
 install_python_codestyle_dependencies:
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/codestyle.txt
 
 .PHONY: install_python_test_dependencies
-install_python_test_dependencies:
+install_python_test_dependencies: install_python_olympia_module
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/system.txt
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/prod.txt
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/tests.txt
@@ -94,8 +99,6 @@ install_python_test_dependencies:
 install_python_dev_dependencies: install_python_test_dependencies install_python_codestyle_dependencies
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/dev.txt
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/docs.txt
-	# pep 517 mode (the default) breaks editable install in our project. https://github.com/mozilla/addons-server/issues/16144
-	$(PIP_COMMAND) install --no-use-pep517 -e .
 
 .PHONY: install_node_dependencies
 install_node_dependencies: install_node_js copy_node_js
@@ -234,6 +237,10 @@ format: ## Autoformat our codebase.
 	NODE_PATH=$(NODE_MODULES) npm exec $(NPM_ARGS) -- prettier --write '**'
 	ruff check --fix-only .
 	ruff format .
+
+.PHONY: extract_locales
+extract_locales: ## extracts and merges translation strings
+	./scripts/run_l10n_extraction.sh
 
 .PHONY: help_submake
 help_submake:
--- a/scripts/extract-l10n.sh
+++ b/scripts/extract-l10n.sh
@@ -19,11 +19,6 @@ set -o pipefail
 # Treat unset variables as an error an exit immediately.
 set -u
 
-# Extraction needs our django settings for jinja, so we need a django settings
-# module set. Since this command is meant to be run in local envs, we use
-# "settings".
-DJANGO_SETTINGS_MODULE=settings
-
 INITIAL_GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 GIT_CHANGES=$(git status --porcelain)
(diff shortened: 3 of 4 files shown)


.circleci/config.yml
```diff
@@ -491,9 +491,9 @@ commands:
                 name: Wait for memcached
                 command: dockerize -wait tcp://localhost:11211 -timeout 1m
       - run:
-          name: Set environment variables
+          name: Set environment variables and install olympia
           command: |
-            echo export PYTHONPATH=src >> $BASH_ENV
+            pip install -e .
```
eviljeff, line 496: This... seems like an unexpected change.  Previously we were exporting `PYTHONPATH` and now we're doing `pip install` one every job`?  If it's necessary for the localization job surely we don't need it for all the other jobs too?

diox: I am not sure we should be doing this for every job, but the `PYTHONPATH=src` was a hack that we should aim to get rid of for consistency.

diox: I think ideally we should do something like:
```diff
diff --git a/Makefile-docker b/Makefile-docker
index 53ed608c40..7495ea57bb 100644
--- a/Makefile-docker
+++ b/Makefile-docker
@@ -80,12 +80,17 @@ cleanup_python_build_dir:
 	# Work arounds "Multiple .dist-info directories" issue.
 	rm -rf /deps/build/*
 
+.PHONY: install_python_olympia_module
+install_python_olympia_module:
+	# pep 517 mode (the default) breaks editable install in our project. https://github.com/mozilla/addons-server/issues/16144
+	$(PIP_COMMAND) install --no-use-pep517 -e .
+
 .PHONY: install_python_codestyle_dependencies
 install_python_codestyle_dependencies:
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/codestyle.txt
 
 .PHONY: install_python_test_dependencies
-install_python_test_dependencies:
+install_python_test_dependencies: install_python_olympia_module
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/system.txt
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/prod.txt
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/tests.txt
@@ -94,8 +99,6 @@ install_python_test_dependencies:
 install_python_dev_dependencies: install_python_test_dependencies install_python_codestyle_dependencies
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/dev.txt
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/docs.txt
-	# pep 517 mode (the default) breaks editable install in our project. https://github.com/mozilla/addons-server/issues/16144
-	$(PIP_COMMAND) install --no-use-pep517 -e .
 
 .PHONY: install_node_dependencies
 install_node_dependencies: install_node_js copy_node_js
```

+ remove `echo export PYTHONPATH=src >> $BASH_ENV` from `.circleci/config.yml` and don't add the `pip install -e .`.

KevinMind: Just curious, if the result of running the `pip install -e .` is _essentially_ to set the PYTHONPATH appropriately, and we were already doing that export in the setup_container part of our pipeline, then what exactly is the downside of running `pip install` here?

I'm not opposed to doing it somewhere else, it just seems like the alternative is harder to find and thus harder to understand.

scripts/extract-l10n-merge.sh
```diff
@@ -0,0 +1,64 @@
+#! /bin/bash
```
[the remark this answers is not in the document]
KevinMind: Curious, is it not explained in the context section of the PR description? The reason I ask is since we are discussing how relevant that section is recently and I'd like to ask if it answers your question before answering it here.

If that section does not answer, can you tell me what is still unclear?

.circleci/config.yml
diox: `pip install -e` installs a special file in the regular `PYTHONPATH` that helps `python` understand where the `olympia` module is.

My proposal removes the duplicate `pip install -e .` call that your patch introduces, and leaves only a single place where we do this, so I think it's valuable on top of removing the `PYTHONPATH=src` (which had a lot more consequences, since it forced _all_ modules to be looked at in that directory in addition to the regular directories where modules are found)

scripts/extract-l10n-merge.sh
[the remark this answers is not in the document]
diox: I was expected the existing script to be modified to allow only extraction, or a new script doing only extraction to be made and called from the existing script, instead of having the 2 live in parallel.

.circleci/config.yml
KevinMind: Ah I didn't even see the sneaky duplicated call there. Okay, this makes more sense. I'll apply this change to the PR.

scripts/extract-l10n-merge.sh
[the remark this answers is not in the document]
KevinMind: I can update the existing script to call the new script, but according to the plan I laid out in the ticket, the existing script will eventually be replaced entirely by the new script plus the outer CI. In the end all we will need in "the script" is the extraction/merge bit.

But if you prefer I can update the existing script to call this script and remove that code. I don't have a strong opinion on that.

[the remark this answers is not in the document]
KevinMind: Added the change. In addition I added a DRY_MODE flag for the extract script. I was needing to test it locally to make sure everything works, but it kept changing branches and it seemed like a pretty valid use case to want to run that script without committing anything, which is essentially the same as running the new script.

Like I said, if all goes to plan, the existing script will be deleted, but for now it is more easily executed without consequence.

.circleci/config.yml
```diff
@@ -666,6 +676,7 @@ workflows:
       - main
       - reviewers-and-zadmin
       - es-tests
+      - localization
```
diox, line 679: Is this meant to be temporary because we have no other way of testing the extraction process ? I am not sure it's worth running the extraction as part of the general test suite - or even the release process: it's an extra job that will slow things down.

KevinMind: FWIW this job is not slowing anything down. Also defining it as a job gives us the flexibility to run it how and whenever we want. If the goal here is to automate the extraction process, it has to be a job. The only thing left to define is the trigger.

If you want. I can remove it from the list of jobs that run on PRs, but per the screenshots below, it will NEVER be the bottleneck and yes, for now, gives us visibility into what is happening.

Localization:
image

Main:
image

LMK if you still think I should remove it and I will.

KevinMind: IMO it would be good to have this run on PRs so a dev understands what changes to localization content are present in a PR. The current "git diff" is a bit too dull of a knife for that but this can easily be improved. I just don't see why that has to happen now.

KevinMind: More so, in the event that a PR does something (we will find a way, we always do) that makes it impossible to extract locales... don't we want to know about it?

diox: Jobs run in parallel with a fixed number of jobs running at the same time, so any job can have an impact, we have to be careful. It seems fine with the current setup though.

There is a cost/CPU impact still, but I guess we can ignore that at least for now.

I am still not convinced about the value in PRs/commits though. At least until we can fix the formatting of our `.po` files once extracted to be more like Pontoon so that the `git diff` can be readable

diox: > More so, in the event that a PR does something (we will find a way, we always do) that makes it impossible to extract locales... don't we want to know about it?

That is a more convincing argument. I still don't like the overhead (maybe we should reorganize our CircleCI jobs) but let's try it.

.circleci/config.yml
```diff
@@ -575,6 +574,17 @@ jobs:
...
 
+  localization:
+    <<: *defaults
+    steps:
+      - setup_container:
+          wait_services: false
+          install_node_dependencies: true
+          install_python_dev_dependencies: true
+      - run : make extract_locales
+      - run: git diff
```
diox, line 585: We probably want a follow-up to make that useful - at the moment, because of differences in how pontoon formats the po files, it's a mess to read most of the time.

KevinMind: Agreed.

KevinMind: Baby steps.

PR #21737, opened 2024-01-23
notify reporters that made an appeal only

### LINKED ISSUE #21719: Notify only the right reporters when resolving abuse report appeals
When resolving abuse report appeals, we currently notify all original reporters that are attached to the original job. That's wrong: we should only notify those who appealed.

STR:
- Report a content (leave your email address or be logged in)
- Report that same content with a different user/email
- Resolve the job in Cinder (should be a single job for that content)
- Appeal from reporter $1 using the email you received
- Resolve the appeal job in Cinder

Expected results:
- Reporter $1 should be notified, reporter $2 should not.

Actual results:
- Both reporters are notified

Note: at the moment, once an appeal decision has been made, further appeals from others reporters are not possible. This will change soon, and that will add another layer of complexity to this issue: if the initial appeal is resolved and there is another one from reporter $2, then resolving that one should not notify the other reporter.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-659)

eviljeff:
fixes #21719 

Some of the refactoring in models.pu wasn't strictly necessary, I just brought it over from #21730 anyway. And I couldn't quite tell what the two `AlterField`s were for - neither field was altered in this patch - but I guess they're harmless.

One thing I didn't come to a conclusion on was where we needed `CinderJob.abuse_reports` and where `CinderJob.appellants` - i.e. where we need the full list of abuse reports for an appeal, and where we just want the abuse reports to know which reporters to notify (perhaps for reporter appeals we only ever need `appellants`, and only need the full list for target appeals?)


diox:
> One thing I didn't come to a conclusion on was where we needed `CinderJob.abuse_reports` and where `CinderJob.appellants` - i.e. where we need the full list of abuse reports for an appeal, and where we just want the abuse reports to know which reporters to notify (perhaps for reporter appeals we only ever need appellants, and only need the full list for target appeals?)

For appeals that seems right. The only place where I can see it mattering is in `CinderAction.notify_reporters()`, where you've already made that check though ?

src/olympia/abuse/migrations/0022_rename_appeal_date_abusereport_reporter_appeal_date_and_more.py
```diff
@@ -0,0 +1,34 @@
...
+        migrations.AlterField(
+            model_name='cinderjob',
+            name='appeal_job',
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, related_name='appealed_jobs', to='abuse.cinderjob'),
+        ),
+        migrations.AlterField(
+            model_name='cinderjob',
+            name='decision_action',
+            field=models.PositiveSmallIntegerField(choices=[(0, 'No decision'), (1, 'User ban'), (2, 'Add-on disable'), (3, 'Escalate add-on to reviewers'), (5, 'Rating delete'), (6, 'Collection delete'), (7, 'Approved (no action)'), (8, 'Add-on version reject')], default=0),
+        ),
```
diox, line 33: Those are no-op caused by old changes in `related_name` and `choices` respectively, that weren't covered by existing migrations.

eviljeff:
> > One thing I didn't come to a conclusion on was where we needed `CinderJob.abuse_reports` and where `CinderJob.appellants` - i.e. where we need the full list of abuse reports for an appeal, and where we just want the abuse reports to know which reporters to notify (perhaps for reporter appeals we only ever need appellants, and only need the full list for target appeals?)
> 
> For appeals that seems right. The only place where I can see it mattering is in `CinderAction.notify_reporters()`, where you've already made that check though ?

Also the reviewer tools, where for an appeal made by a reporter we'd show all the reports rather than just the one(s) that were being appealed.  But perhaps we should show them all for completeness - and reviewer tools appeals still needs some work anyway.

diox:
> Also the reviewer tools, where for an appeal made by a reporter we'd show all the reports rather than just the one(s) that were being appealed. But perhaps we should show them all for completeness - and reviewer tools appeals still needs some work anyway.

Yeah I think there we should show them all, it's potentially useful context

diox (approved):

Issue #21723, opened 2024-01-18
Local env: warning about unsupported engine


wagnerand:
On local environments, during updating the instance, I see the following warning:

```
npm install --prefix /deps/
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'addons-server@0.0.17',
npm WARN EBADENGINE   required: { node: '>= 18.18' },
npm WARN EBADENGINE   current: { node: 'v18.17.1', npm: '9.6.7' }
npm WARN EBADENGINE }
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-661)

KevinMind:
Linked: https://github.com/nodesource/distributions/issues/1706

diox:
This is not just about local dev: all our addons-server envs are stuck on 18.17 but they shouldn't.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-661

PR #21765, opened 2024-01-29
Broaden Node version requirement

KevinMind:
FIxes: #21723

## Description

This pull request updates the Node version requirement to be ">= 18" instead of ">= 18.18". This change ensures compatibility with a broader range of Node versions.

## Context

We have previously [updated](https://github.com/mozilla/addons-server/pull/21557/files)  our required version of node to 18.18.

When we install node on our docker container, it is installing version 18.17.

This could be related to issue: https://github.com/nodesource/distributions/issues/1706. Since we are only specifying 18.x in our nodesource.list file it seemed appropriate to use a similarly broad version in package.json.

We could instead opt for higher specificity in our nodelist

## Testing

You can test this by running 

```bash
make update_deps
```

You should expect not to see the warning mentioned in the ticket.


diox:
We can do this but I don't think this is the proper fix for https://github.com/mozilla/addons-server/issues/21723. At the moment, all our envs should be using 18.18 and are not because the way the nodesource repo was changed and we haven't adapted our setup.

KevinMind:
```At the moment, all our envs should be using 18.18```

What's the reason? Just want to understand.

**Comment by diox** [comment]
It's not about that specific version - we want to stay on the latest maintained 18.x version to keep up to date with any potential security fixes (in fact, we should be on 18.19 - addons-frontend already is).

The fix to the release list file is trivial, so let's do that.

File: Dockerfile [hunk 1]
```diff
@@ -26,6 +24,7 @@ RUN mkdir /opt/apt-local-repository
 COPY docker/mysql/* /opt/apt-local-repository/
 RUN /opt/apt-local-repository/local-mysql-repos.sh
 
+RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash -
```
diox, line 27: We shouldn't run random bash scripts from the internet through bash if we can help it. Updating the nodesource gpg key and fixing the source list should be enough.

diox (changes requested):

Dockerfile
KevinMind: Just following their own [instructions](https://github.com/nodesource/distributions?tab=readme-ov-file#installation-instructions) where they are strongly recommending we use this approach.

https://github.com/nodesource/distributions/wiki/Repository-Manual-Installation#-simplified-installation-using-scripts

KevinMind: I'm also not sure I would call an officially supported setup script that is the recommended way to install the project that is shipped by the project itself a "random bash script from the internet".

KevinMind: Other Mozilla projects are doing the same thing as well: https://github.com/mozilla/glean.js/pull/1807/files In case social credit means anything here.

KevinMind: BTW this gets us on 18.19.

diox:
Superseded by https://github.com/mozilla/addons-server/pull/21770

Issue #21747, opened 2024-01-25
Add management command to sync policies from cinder to AMO


eviljeff:
We need to get the policies from https://docs.cinder.co/reference/kosa_api_api_policies and fill the `CinderPolicy` model with all the policies returned.

a policy object is like
```json
{
    "id": 4,
    "name": "Harassment",
    ...
    "uuid": "f73ad527-54ed-430c-86ff-80e15e2a352b",
    "description": "We do not allow posts that appear to be shared for the purpose of diminishing, degrading or shaming others, causing them fear, or otherwise tormenting them.\n\n",
    "nested_policies": []
  },
```
`nested_policies` can contain other policies (and so on) - we want those too.  `CinderPolicy` doesn't support nesting currently, so needs a `parent` fk.

Ideally it would be implemented as a django management command that can be run on demand by SRE (and potentially linked up to a cron in the future)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-669)

wagnerand:
Note that a `CinderPolicy` is linked (1:1) to a `ReviewActionReason`, which holds the `canned_response` for that reason and policy. So before sync, the existing _CinderPolicies_ must not be cleared and during sync, for items that already exist on AMO, the internal id of a `CinderPolicy` must not change, so that the existing link is not broken.

KevinMind:
@Andreas Wagner can you clarify what does “cleared” mean and what does “internal id” mean? Is internal id the uuid as in the ticket description? And by cleared you mean:

given a policy with uuid ‘234’ already exists in our database

when we sync policies and encounter this policy based on the uuid 234

then we should have only 1 instance and the uuid for that instance should be unchanged. Also the relation between the policy and the review action reason should be unbroken

Can you clarify that is what you mean? And if not, could you help me understand?

eviljeff:
(channelling my inner @wagnerand )
> @andreas Wagner can you clarify what does “cleared” mean 

deleted

> and what does “internal id” mean? Is internal id the uuid as in the ticket description? 
yes

> And by cleared you mean:
> given a policy with uuid ‘234’ already exists in our database
> when we sync policies and encounter this policy based on the uuid 234
> then we should have only 1 instance and the uuid for that instance should be unchanged. Also the relation between the policy and the review action reason should be unbroken

yes.  (uuid should be `unique=True` if it's not already).

wagnerand:
[...]

PR #21773, opened 2024-01-31
Add CinderPolicy model and sync_cinder_policies management command

KevinMind:
Fixes: #21747

## Description

1. update CinderPolicy to make uuid unique and add parent field for nesting. On a separate commit to be cherry pickable if we want to push early/separately

2. add a task/command to sync cinder policies to our database.

## Context

Since we want to be able to trigger this either via manage.py or via cron a task seemed appropriate. There are many opinions about when to use tasks so happy to hear feedback.

## Testing

1. you need a CINDER_API_TOKEN (talk to @eviljeff). set it in settings_local
2. ```make shell```
3. ```./manage.py sync_cinder_policies```
4. Check your local database and expect CinderPolicy records to exists. Policies are related via nested_policies as child policies.


src/olympia/abuse/management/commands/sync_cinder_policies.py
```diff
@@ -0,0 +1,15 @@
...
+
+class Command(BaseCommand):
+    help = (
+        'Sync all cinder policies to our local database. Syntax: \n'
+        '    ./manage.py sync_cinder_policies'
+    )
+    log = olympia.core.logger.getLogger('z.abuse')
+
+    def handle(self, *args, **options):
+        sync_cinder_policies.apply()
```
diox, line 15: `apply()` will run the task locally synchronously, not through a celery worker. Was that your intention ? We typically use `.delay()` or `.apply_async()` to queue the task.

src/olympia/abuse/management/commands/sync_cinder_policies.py
```diff
@@ -0,0 +1,15 @@
...
+import olympia.core.logger
+from olympia.abuse.tasks import sync_cinder_policies
+
+
+class Command(BaseCommand):
+    help = (
+        'Sync all cinder policies to our local database. Syntax: \n'
+        '    ./manage.py sync_cinder_policies'
+    )
+    log = olympia.core.logger.getLogger('z.abuse')
```
diox, line 12: This is unused/pointless.

src/olympia/abuse/management/commands/sync_cinder_policies.py
```diff
@@ -0,0 +1,15 @@
+from django.core.management.base import BaseCommand
+
+import olympia.core.logger
+from olympia.abuse.tasks import sync_cinder_policies
+
+
+class Command(BaseCommand):
+    help = (
+        'Sync all cinder policies to our local database. Syntax: \n'
+        '    ./manage.py sync_cinder_policies'
```
diox, line 10: Usage syntax is already automatically included by Django when help is printed, you don't need to add it to the help string.

src/olympia/abuse/tasks.py
```diff
@@ -97,3 +99,36 @@ def resolve_job_in_cinder(*, cinder_job_id, review_text, decision, policy_ids):
...
+                defaults={
+                    'name': policy['name'],
+                    'text': policy['description'],
+                    'parent_id': parent_id,
+                },
+            )
+
+            nested = policy.get('nested_policies')
+
+            if nested:
```
diox, line 131: nit:

```suggestion
            if nested := policy.get('nested_policies'):
```

src/olympia/abuse/tasks.py
```diff
@@ -97,3 +99,36 @@ def resolve_job_in_cinder(*, cinder_job_id, review_text, decision, policy_ids):
     cinder_job = CinderJob.objects.get(id=cinder_job_id)
     policies = CinderPolicy.objects.filter(id__in=policy_ids)
     cinder_job.resolve_job(review_text, decision, policies)
+
+
+@task
+@use_primary_db
+def sync_cinder_policies():
```
diox, line 106: Add an explicit celery route in settings for this task (see failing test)

src/olympia/abuse/management/commands/sync_cinder_policies.py
KevinMind: Good catch. Only had it that way for local testing.

src/olympia/abuse/management/commands/sync_cinder_policies.py
KevinMind: Yep, artifact of copy paste. Removing.

src/olympia/abuse/management/commands/sync_cinder_policies.py
KevinMind: TIL. Again, copied from existing command.

src/olympia/abuse/management/commands/sync_cinder_policies.py
KevinMind: Actually ended up adding a log, otherwise the command prints nothing to the console which I find a bit unsettling. lmk what you think about that.

src/olympia/abuse/management/commands/sync_cinder_policies.py
diox: Local envs should have a working celery worker (tests don't, but they run with `CELERY_TASK_ALWAYS_EAGER = True`)

src/olympia/abuse/management/commands/sync_cinder_policies.py
diox: Works for me. Maybe to be more accurate the log could say something like "Triggered policy sync task" or something like that but it doesn't really matter.

diox (approved):

diox:
Oh, while you're at it, to help debugging, consider adding a simple `__str__` to `CinderPolicy` like this one:
```python
    def __str__(self):
        return self.name
````

src/olympia/abuse/management/commands/sync_cinder_policies.py
eviljeff: if this is only run as a command/cron we don't really need the task anyway, imo - the request should complete quickly and there would be immediate feedback in the console/logs.

eviljeff: (just an observation - I'm not requesting a change)

KevinMind: I think you can argue either side. Why I chose a task in this case:
- I typically prefer retry able mechanisms for things that depend on external APIs as they can randomly fail. (I should actually add a retry on non 401/3 network errors
- this is work that can run asynchronously as no one directly waits for it to finish and putting it in a task emphasizes that mental model.

Where would you put it outside of a task out of curiosity? In a util method or somewhere specific? I'm assuming we want to either trigger this via management command or via cron so it cannot be implemented directly as either.

src/olympia/abuse/management/commands/sync_cinder_policies.py
KevinMind: Updated.

src/olympia/abuse/management/commands/sync_cinder_policies.py
eviljeff: Either a function in utils that the management command calls, or directly in the management `handle` method.  Or in `cron.py` - if the function is defined in `settings.CRON_JOBS` it is callable via `./manage.py cron xxx`.  

But it's fine as it is.

eviljeff (commented):
I don't have anything to add beyond Mat's review.  My merging of another PR has caused a merge conflict in this patch though.

src/olympia/abuse/management/commands/sync_cinder_policies.py
KevinMind: TIL. Cool.

KevinMind:
There is a conflicting migration as well. Recreating mine.

KevinMind:
How did this not trigger a CI error? @diox any ideas? I suppose our linter isn't actually tracking dependencies? Should that be the case? Feels wrong.

diox:
You didn't have a test covering the management command, so that didn't break CI

PR #21783, opened 2024-01-31
Enable rejection action for unlisted approved versions

### LINKED ISSUE #21308: Rejecting unlisted auto-approved versions
As a reviewer I should be able to send a delay rejection / rejection to an unlisted auto-approved version. 

Even though it doesn't change anything for the developer we will have the correct status recorded for each version.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-524)

wagnerand:
Fix #21308.

### DIFF
--- a/src/olympia/reviewers/forms.py
+++ b/src/olympia/reviewers/forms.py
@@ -117,6 +117,7 @@ class VersionsChoiceWidget(forms.SelectMultiple):
             amo.STATUS_APPROVED: [
                 'block_multiple_versions',
                 'confirm_multiple_versions',
+                'reject_multiple_versions',
             ],
             amo.STATUS_AWAITING_REVIEW: [
                 'approve_multiple_versions',
--- a/src/olympia/reviewers/tests/test_forms.py
+++ b/src/olympia/reviewers/tests/test_forms.py
@@ -575,6 +575,7 @@ class TestReviewForm(TestCase):
             # That version is approved.
             'block_multiple_versions',
             'confirm_multiple_versions',
+            'reject_multiple_versions',
             'set_needs_human_review_multiple_versions',
         ]
         assert option1.attrib.get('value') == str(self.version.pk)


KevinMind (approved):
Read the issue, have zero idea how this PR relates to to the issue so I don't think I can review this without some additional context. But I trust you :) so fire away. 🎸 🚀

wagnerand:
This patch enables the "reject_multiple_versions" action for unlisted versions that have been approved, allowing reviewers to reject unlisted (auto-)approved versions. Does that help?

KevinMind:
Reading the issue I can latch onto "send a delay rejection / rejection" and "to an unlisted auto-approved version"

First, I don't actually know what some of that means and it seems like the ticket assumes that I do. I can pretty confidently guess what a rejection is in this context, and I happen to know what a version means from prior context, but a delay rejection vs rejection has no meaning and I can guess an auto-approved version is based on common sense. It is dangerous to rely on that though so I prefer not to when possible.

So it is possible to grok what the ticket is asking for with some context, but requires some mental effort to do so. And still there is guess work.

Now trying to map that onto the pr, which is super simple, we see that we are adding "reject_multiple_versions" to amo.STATUS_APPROVED in a reviewer form.

Again, I can intuite that the "reject_multiple_versions" is somehow related to the desired action in the issue because there is really no logically alternative, but really only because the words "reject" and "versions" are in there.

Questions that are still unanswered for me include:

where does "reject_multiple_versions" string actually come from? Is this some external property is it a reference to a field in our db somewhere?
How does adding this attribute actually impact the UX flow for a reviewer, screenshot would help here.
Why is the "delay" rejection relevant here? It's mentioned in the issue but not in the PR.
Why does the approved status mention rejecting "multiple_versions" when the issue mentions rejecting a single "auto approved" version. Are those concepts related? How? If not, why does it mentione multiple versions

wagnerand:
> How does adding this attribute actually impact the UX flow for a reviewer, screenshot would help here.

The UX impact is not super visible or meaningful in a screenshot (if you don't know what add-on and its version states it is about). Basically, when a reviewer rejects one or more versions, there is select box, from which they have to choose which versions the reject applies to. Before this patch, the reject action itself is already there and shows some versions to select from (because a reviewer can already reject an unlisted version that has *not* been approved yet). With the patch, none of that will change, except that there will be more entries in that select box. However, in order to evaluate what should or should not be shown, you need to know the version states).

> Why is the "delay" rejection relevant here? It's mentioned in the issue but not in the PR.

Yes, you are right that delayed rejections are not changed for unlisted approved versions. I should have clarified that in the issue, but the team has not yet reached consensus on whether to also enable delayed rejections (beyond instant rejections in this patch). Why this matters would make this reply much longer, I am happy to expand in a meeting, but the short version is that rejections don't impact the developer of an unlisted add-on/version overall, so an instant vs delayed rejection does not make a difference. Nevertheless the review team will talk about this probably next week and if we want to enable delayed rejections as well, I will submit a followup patch.

> Why does the approved status mention rejecting "multiple_versions" when the issue mentions rejecting a single "auto approved" version. Are those concepts related? How? If not, why does it mentione multiple versions

There is a long history with rejecting one vs multiple versions. Initially reviewers only had (and needed) the ability to reject one version (using the `reject` action), which was the version that was pending review/approval, when we didn't have auto-approval. With auto-approval, multiple versions can be submitted (and get approved) before a single one is reviewed. If an issue is found in one of these auto-approved versions, it is likely the same issue is in others as well, since it hadn't been caught before. That's why reviewers can reject (one or) multiple (auto-)approved versions at once, which is done using the separate `reject_multiple_versions` action.

KevinMind:
Thanks for the clarification @wagnerand

Issue #21687, opened 2024-01-11
Enable suppressed emails flow in production


KevinMind:
Relating to the implementation of #21538 and #21537 we need to test the feature in staging and enable the feature flag in production.

This is currently blocked by https://mozilla-hub.atlassian.net/browse/SVCSE-1673

Once that is shipped we need to:

* enable the cron job to sync suppressed emails to our database
* enable the waffle switch "suppressed-email"
* test the flow following instructions specified in the two linked tickets.

Once the feature is found to be working, we can enable the waffle switch in all environments and it will be available for users.

NOTE: replace the server id (2301) with the relevant server ID of your desired environment. You can find server IDs by going to portal.socketlabs.com

Suppressed email list: https://portal.socketlabs.com/2301/suppression

A suppressed email is an email added to the socketlabs email suppression list. Socketlabs will not attempt to send emails to these addresses as they have previously proven unreachable.

### Testing
1. Ensure the waffle flag is enabled
2. Ensure your user is on the socketlabs suppression list
3. Ensure your user is in the group "force send email" in django admin
4. Visit devhub
5. Expect to see the suppressed email banner
6. Click it
7. Click to verify email
8. Expect to receive email
9. Click on email link
10. Expect email to be unsupressed

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-648)

vcarciu:
Thank you, we will follow this issue in order to see when it is ready for testing

ioanarusiczki:
@KevinMind 

Hello, I've been adding on [dev](https://portal.socketlabs.com/2301/suppression) and [stage](https://portal.socketlabs.com/16176/suppression) some emails, they've been also added into admin - Force Send Mail group, and tried to repeat the flow. 

The problem is that dev hub won't display the yellow banner (step5). I tried this last week with same results.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-648

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-648

PR #21784, opened 2024-02-01
Add management command for sync suppressed emails

KevinMind:
relates to: #21687

## Description

This pull request adds a new management command for syncing suppressed emails from Socket Labs to the database. The command fetches the suppression list from Socket Labs, downloads it, and then bulk creates the suppressed emails in the database. The pull request also includes a task for sending suppressed email confirmations.

## Context

This PR exposes the [sync_blocked_emails](https://github.com/mozilla/addons-server/blob/2f89e2033f47131676e60720ecef6b4a0faab40a/src/olympia/users/tasks.py#L105) task via a management command. 

This can be useful for a few reasons:
- testing the email suppression feature will likely require manually adding emails to the suppression list and then syncing them to our DB so we can run through the verification steps to unsuppress the email.
- in general if we need, for operational reasons, to resync the suppression list we can do so via an easy to execute command.

In general the sync_blocked_emails task will be run periodically. This is not in production yet as we have only just recently been given the necessary credentials to communicate with socketlabs via API and having that job executing until now would have simply erred indefinitely.

### DIFF
--- a/src/olympia/users/management/commands/sync_suppressed_emails.py
+++ b/src/olympia/users/management/commands/sync_suppressed_emails.py
@@ -1,0 +1,10 @@
+from django.core.management.base import BaseCommand
+
+from olympia.users.tasks import sync_blocked_emails
+
+
+class Command(BaseCommand):
+    """Sync Socket labs suppression list to database."""
+
+    def handle(self, *args, **options):
+        sync_blocked_emails.apply()
--- a/src/olympia/users/tasks.py
+++ b/src/olympia/users/tasks.py
@@ -105,6 +105,8 @@ def assert_socket_labs_settings_defined():
 def sync_blocked_emails(batch_size=BATCH_SIZE, **kw):
     assert_socket_labs_settings_defined()
 
+    task_log.info('fetching suppression list from socket labs...')
+
     path = f'servers/{settings.SOCKET_LABS_SERVER_ID}/suppressions/download'
     params = {'sortField': 'suppressionLastUpdate', 'sortDirection': 'dsc'}
     url = (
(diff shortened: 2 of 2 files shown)


src/olympia/users/tasks.py
```diff
@@ -139,6 +146,9 @@ def sync_blocked_emails(batch_size=BATCH_SIZE, **kw):
 
                 email_blocks = [SuppressedEmail(email=record[3]) for record in batch]
                 SuppressedEmail.objects.bulk_create(email_blocks, ignore_conflicts=True)
+                count += len(batch)
```
eviljeff, line 149: maybe we want some logging every batch to show progress?

src/olympia/users/management/commands/sync_suppressed_emails.py
eviljeff, line 1: The issue might explain further, but if this is something that is expected to be run as a cron, with no args or options, it would probably be better implemented as a simple function in cron.py instead?

eviljeff (commented):
looks good so far but needs issue to compare against to see if it satisfies.

src/olympia/users/management/commands/sync_suppressed_emails.py
KevinMind: This PR is a simple convenience for testing and if we ever want to manually resync the suppression list. We should run this task on a periodic schedule. The issue will likely not explain much unfortunately.

src/olympia/users/tasks.py
KevinMind: I don't think it's necessary, at least not yet. Only added the logging here because the download part can take some time to run and without any logs it can be confusing and made me wonder if something was wrong with the code. I added the count at the end to just verify that we have synced "something". The syncing is really fast due to the batching so it wouldn't add much value IMO.

src/olympia/users/management/commands/sync_suppressed_emails.py
eviljeff: You can do that with a simple function in cron.py too 🙂

eviljeff (approved):

src/olympia/users/management/commands/sync_suppressed_emails.py
KevinMind: Can I trigger a function in cron in both a management command and periodically? Just curious? Also would there be retry logic if implemented that way?

Issue #21790, opened 2024-02-01
Add Group for sending mail on dev/stage


KevinMind:
## Problem 

Currently on dev/stage by default we do not send real mail but add a FakeMail record and log the mail to the console. We do this to prevent sending mails for real users that have data in dev/stage and to prevent devs and other employees from getting spammed too often.

Sometimes (ex: ticket) we need to actually send mails on non prod environments. Until now we have been using an environment variable EMAIL_QA_ALLOW_LIST that provides a list of emails for which we should bypass the fake mail and send a real mail. You can also specify at runtime that you want a given mail to send an actual mail.

This environment variable is static and requires SRE as well as a deployment to update, but often we want to quickly enable real mail for N users to test something and then disable again.

## Solution

We will use a `Group` to manage users that should have mail sent which can be updated on the fly via the admin or terminal. This gives us a much faster and more convenient way to update this list when we need it and removes the SRE dependency.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-684)

ioanarusiczki:
@KevinMind 

I added emails to [dev](https://addons-internal.dev.mozaws.net/en-US/admin/models/access/group/50112/change/) and [stage ](https://addons-internal.stage.mozaws.net/en-US/admin/models/access/group/50112/change/) and I tried a couple of things
- someone is rating an add-on 
- reviewer reply
- add-on submission, add-on approval
- abuse report decisions

For users from ForceSend Mail group, I received the emails in my inboxes and they continue to be present in fakemail too.

KevinMind:
@ioanarusiczki this is expected. ForceSend will have both, fake and real emails.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-684

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-684

PR #9377, opened 2018-09-07
unsigned, PositiveAutoFields

eviljeff:
tests pass so must be okay!  fixes #8853

Issue #3705, opened 2016-10-10
[tracker] Switch to django-migrations


EnTeQuAk:
This is the overall tracking issue to manage progress on switching to Django's builtin migrations framework.

We are currently using https://github.com/mozilla/schematic/ which is a very simplified migrations script which leads to writing plain SQL migrations and having a hard time with data migrations and lots of one-off scripts.

(copied from https://github.com/mozilla/addons-server/issues/3705#issuecomment-471957170)
* [x] Make sure the dev, stage and prod database schema are mostly identical (#10916)
* [x] Then make sure we are able to generate a schema locally that looks like -dev, or go the other way around and implement a migration that migrates -dev, -stage and -prod to something like -local. I haven't looked too deep into this so I can't really say which may be more complicated (#10917) 
* [x] Then switch to django-migrations :) (#12333)
  * Add initial migrations
* [x] Remove old migrations files  (#12390)

EnTeQuAk:
Some updates on this. There are a few steps required to actually get there.

* Make sure the dev, stage and prod database schema are mostly identical
* Then make sure we are able to generate a schema locally that looks like -dev, or go the other way around and implement a migration that migrates -dev, -stage and -prod to something like -local. I haven't looked too deep into this so I can't really say which may be more complicated
* Then switch to django-migrations :)
 * Add initial migrations and maybe even use that to migrate schema changes required in step 2 
* Remove old migrations files

Attached you'll find schema exports from Nov 30th 2018 which give you a hunch of what differences there are. Most are already fixed at this time afaik.

With us being on Django 1.11 we can add [custom indexes](https://docs.djangoproject.com/en/1.11/ref/models/indexes/) which hopefully cover most of our use-cases.

[prod_schema.txt](https://github.com/mozilla/addons-server/files/2956506/prod_schema.txt)
[stage_schema.txt](https://github.com/mozilla/addons-server/files/2956507/stage_schema.txt)
[dev_schema.txt](https://github.com/mozilla/addons-server/files/2956508/dev_schema.txt)

eviljeff:
copied the steps to the top comment with checkboxes and filed https://github.com/mozilla/addons-server/issues/10916 and https://github.com/mozilla/addons-server/issues/10917 for the first two steps

diox:
[...]

PR #21791, opened 2024-02-01
Refactor email sending logic to use a group for QA email recipients

KevinMind:
Fixes: #21790

## Description

This pull request refactors the email sending logic to use a group for QA email recipients instead of a static list. This allows for easier management of QA email addresses and improves code maintainability.

## Context

The name is maybe up for debate but it shouold be relatively clear what the group is for.

What is unclear right now, how to manage this group? Do we have a migration that creates it and adds users? Do we have a management command to add/remove users?

## Testing

Run the migrations to create the group and add users, try to send mail with other settings disabled and you should see it send an email, an actual email.

### DIFF
--- a/src/olympia/access/migrations/0003_add_force_send_mail_group.py
+++ b/src/olympia/access/migrations/0003_add_force_send_mail_group.py
@@ -1,0 +1,29 @@
+from django.db import migrations
+
+from olympia.amo.mail import DevEmailBackend
+
+def create_group(apps, schema_editor):
+    Group = apps.get_model('access', 'Group')
+
+    Group.objects.create(
+        name=DevEmailBackend.force_send_mail_group,
+        notes=(
+            'UserProfiles belonging to this group will have real emails sent'
+            ' to them in dev/stage environments. This is useful for testing email flows.'
+        )
+    )
+
+def delete_group(apps, schema_editor):
+    Group = apps.get_model('access', 'Group')
+
+    Group.objects.filter(name=DevEmailBackend.force_send_mail_group).delete()
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('access', '0002_give_api_bypass_throttling_permission'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_group, delete_group)
+    ]
--- a/src/olympia/amo/mail.py
+++ b/src/olympia/amo/mail.py
@@ -1,9 +1,9 @@
-from django.conf import settings
 from django.core import mail
 from django.core.mail.backends.base import BaseEmailBackend
 
 import olympia.core.logger
 from olympia.amo.models import FakeEmail
+from olympia.users.models import UserProfile
 
 
 log = olympia.core.logger.getLogger('z.amo.mail')
@@ -17,21 +17,26 @@ class DevEmailBackend(BaseEmailBackend):
     `settings.SEND_REAL_EMAIL` is disabled.
 
     BUT even if `settings.SEND_REAL_EMAIL` is disabled, if the targeted
-    email address is in the `settings.EMAIL_QA_ALLOW_LIST` list,
-    the email will be sent.
+    email address is in `dev_email` group users, the email will be sent.
     """
+
+    force_send_mail_group = 'Force Send Mail'
 
     def send_messages(self, messages):
         """Save a `FakeEmail` object viewable within the admin.
 
         If one of the target email addresses is in
-        `settings.EMAIL_QA_ALLOW_LIST`, it send a real email message.
+        group named `Group.force_send_mail_group`, it send a real email message.
         """
         log.info('Sending dev mail messages.')
         qa_messages = []
+        force_send_emails = UserProfile.objects.filter(
+            groups__name=self.force_send_mail_group
+        ).values_list('email', flat=True)
+
         for msg in messages:
             FakeEmail.objects.create(message=msg.message().as_string())
-            qa_emails = set(msg.to).intersection(settings.EMAIL_QA_ALLOW_LIST)
+            qa_emails = set(msg.to).intersection(force_send_emails)
             if qa_emails:
                 if len(msg.to) != len(qa_emails):
                     # We need to replace the recipients with the QA
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -752,8 +752,6 @@ DEFAULT_FROM_EMAIL = ADDONS_EMAIL
 # Email goes to the console by default.  s/console/smtp/ for regular delivery
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
-# Please use all lowercase for the QA allow list.
-EMAIL_QA_ALLOW_LIST = env.list('EMAIL_QA_ALLOW_LIST', default=())
 # Please use all lowercase for the deny_list.
 EMAIL_DENY_LIST = env.list('EMAIL_DENY_LIST', default=('nobody@mozilla.org',))
 
--- a/src/olympia/amo/tests/test_fields.py
+++ b/src/olympia/amo/tests/test_fields.py
@@ -96,6 +96,7 @@ class TestPositiveAutoField(TestCase):
             assert extra == 'auto_increment'
 
     def test_unsigned_int_limits(self):
+        self.ClassUsingPositiveAutoField.objects.all().delete()
         self.ClassUsingPositiveAutoField.objects.create(id=1)
         mysql_max_signed_int = 2147483647
         self.ClassUsingPositiveAutoField.objects.create(id=mysql_max_signed_int + 10)
--- a/src/olympia/amo/tests/test_pytest_fixtures.py
+++ b/src/olympia/amo/tests/test_pytest_fixtures.py
@@ -8,8 +8,7 @@ from olympia.access.models import Group
 
 
 def test_admin_group(admin_group):
-    assert Group.objects.count() == 1
-    admin_group = Group.objects.get()
+    admin_group = Group.objects.get(name='Admins')
     assert admin_group.name == 'Admins'
     assert admin_group.rules == '*:*'
 
--- a/src/olympia/amo/tests/test_send_mail.py
+++ b/src/olympia/amo/tests/test_send_mail.py
@@ -9,8 +9,13 @@ from django.utils import translation
 
 from celery.exceptions import Retry
 
+from olympia.access.models import Group
+from olympia.amo.mail import DevEmailBackend
 from olympia.amo.models import FakeEmail
-from olympia.amo.tests import TestCase
+from olympia.amo.tests import (
+    TestCase,
+    user_factory,
+)
 from olympia.amo.utils import send_html_mail_jinja, send_mail
 from olympia.users import notifications
 from olympia.users.models import UserNotification, UserProfile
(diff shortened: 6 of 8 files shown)


src/olympia/amo/mail.py
```diff
@@ -17,30 +17,39 @@ class DevEmailBackend(BaseEmailBackend):
...
 
     BUT even if `settings.SEND_REAL_EMAIL` is disabled, if the targeted
-    email address is in the `settings.EMAIL_QA_ALLOW_LIST` list,
-    the email will be sent.
+    email address is in `dev_email` group users, the email will be sent.
     """
 
+    @staticmethod
+    def force_send_mail_group():
+        return 'Force Send Mail'
```
diox, line 25: Nit: I don't think we need an email, you can make it a simple property instead
```suggestion
    force_send_mail_group = 'Force Send Mail'
```

src/olympia/amo/mail.py
```diff
@@ -17,30 +17,39 @@ class DevEmailBackend(BaseEmailBackend):
...
+
     def send_messages(self, messages):
         """Save a `FakeEmail` object viewable within the admin.
 
         If one of the target email addresses is in
-        `settings.EMAIL_QA_ALLOW_LIST`, it send a real email message.
+        `Group('force_send_mail_group')`, it send a real email message.
```
diox, line 31:
```suggestion
        group named `Group.force_send_mail_group`, it send a real email message.
```

(Or something like that, to make it clearer - I'm not particularly happy with my suggestion either)

diox (commented):
Just some drive-by nits

src/olympia/amo/mail.py
eviljeff, line 25: Probably best we keep the name descriptive and similar to the existing setting
```suggestion
        return 'QA Real Emails'
```
but @wagnerand because he'll be the one who has to remember what its for in the future 😄

src/olympia/access/migrations/0003_add_force_send_mail_group.py
```diff
@@ -0,0 +1,20 @@
+from django.db import migrations
+
+from olympia.access.models import Group
+from olympia.amo.mail import DevEmailBackend
+
+def create_group(apps, schema_editor):
+    Group.objects.create(name=DevEmailBackend.force_send_mail_group())
```
eviljeff, line 7: Lets add some `notes` too.

src/olympia/amo/mail.py
```diff
@@ -17,30 +17,39 @@ class DevEmailBackend(BaseEmailBackend):
...
-                if len(msg.to) != len(qa_emails):
-                    # We need to replace the recipients with the QA
-                    # emails only prior to send the message for real.
-                    # We don't want to send real emails to people if
-                    # they happen to also be in the recipients together
-                    # with white-listed emails
-                    msg.to = list(qa_emails)
-                qa_messages.append(msg)
+            try:
+                dev_email_group = Group.objects.get(name=self.force_send_mail_group())
```
eviljeff, line 38: move this out of the for loop so we aren't `get`ting for each `msg`, and flatten it to a set once.

src/olympia/lib/settings_base.py
```diff
@@ -752,8 +752,6 @@ def get_db_config(environ_var, atomic_requests=True):
 # Email goes to the console by default.  s/console/smtp/ for regular delivery
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
-# Please use all lowercase for the QA allow list.
-EMAIL_QA_ALLOW_LIST = env.list('EMAIL_QA_ALLOW_LIST', default=())
```
eviljeff, line 756: We could have a `RunPython` migration that (attempts to) prefill the `Group` with the current values?

src/olympia/amo/mail.py
```diff
@@ -17,30 +17,39 @@ class DevEmailBackend(BaseEmailBackend):
     `settings.SEND_REAL_EMAIL` is disabled.
 
     BUT even if `settings.SEND_REAL_EMAIL` is disabled, if the targeted
-    email address is in the `settings.EMAIL_QA_ALLOW_LIST` list,
-    the email will be sent.
+    email address is in `dev_email` group users, the email will be sent.
     """
 
+    @staticmethod
```
eviljeff, line 23: Stick it as a property on the class?  (Not sure why it needs to be a method if it just returns a fixed string)

eviljeff (approved):

src/olympia/amo/mail.py
eviljeff: dupe to mat's comment above

src/olympia/amo/mail.py
diox: I voted force `Force Send Emails` or `Bypass Fake Email` in slack so that it would describe what the group does rather than who the group contains. I think it's better to do the former, it's not just about QA.

src/olympia/lib/settings_base.py
diox: From a look at the current data it seems all worthless

src/olympia/amo/mail.py
eviljeff: I'd vote for "Bypass Fake Email" then, as it's the most descriptive.

wagnerand: This basically wouldn't impact prod, since we don't use `FakeMail` there, correct? I guess we could even just not create that group on prod, if so.

eviljeff: Ah, yes, you're right @wagnerand - it won't impact you.

src/olympia/access/migrations/0003_add_force_send_mail_group.py
KevinMind: Done.

src/olympia/amo/mail.py
KevinMind: Done.

KevinMind: I'm not sure what I'm supposed to do at this point. Are we changing the name? Do we need to modify the migration script to check for prod/non-prod. I think that latter is more code complexity to remove a harmless outcome. The former seems to be a mix of opinions so unless someone can make a definitive decision I would leave as is as the easiest path forward.

src/olympia/amo/mail.py
KevinMind: Good catch.

src/olympia/amo/mail.py
KevinMind: Done.

src/olympia/access/migrations/0003_add_force_send_mail_group.py
```diff
@@ -0,0 +1,26 @@
+from django.db import migrations
+
+from olympia.access.models import Group
```
diox, line 3: Never import an actual django model in a migration: you would get the actual model as currently defined instead of the proxy for the historical state it's supposed to be at the time of the migration, causing potential failures.

This is why the callbacks receive an `apps` argument, you want to do (inside the functions themselves):
```python
Group = apps.get_model('access', 'Group')
```

diox (commented):
Please fix tests - Notably,  `TestPositiveAutoField.test_unsigned_int_limits`, which is is trying to force the creation of a `Group` with `id` 1 but that migration is preventing it from doing that, need to find a workaround for that... Maybe we can use another simple model with no data migration, or some other trick.

KevinMind:
> Please fix tests - Notably, `TestPositiveAutoField.test_unsigned_int_limits`, which is is trying to force the creation of a `Group` with `id` 1 (it wants to test lower and upper bounds for `PositiveAutoField`) but that migration is preventing it from doing that, need to find a workaround for that... Maybe we can use another simple model with no data migration, or some other trick.

It's not totally clear to me why we have this test? Are we testing the behaviour of django/mysql here? Tried tracking back to this [PR](https://github.com/mozilla/addons-server/pull/17140) and ultimately this [ticket](https://github.com/mozilla/addons-server/issues/3705) ultimately it seems this test might have originated in spirit from this [comment](https://github.com/mozilla/addons-server/pull/9377#issuecomment-422176529).

```
You could even add a custom test model just for this.
```

Looks like that part was not implemented and we went with `Group`.

I'm happy to fix the test, one idea could be that we use a update_or_create instead of a create. Another is to implement a dummy model to use for this test that is only used there. Finally, we could make this test delete all the records before running to make sure it is clean at time of execution. Happy to do whatever you want here.

I'm wondering though if this test should be run on a clean DB though. Do we have a concept of pure/unpure DB tests? I've seen this in the past so wondering if it's relevant.

src/olympia/access/migrations/0003_add_force_send_mail_group.py
eviljeff: whoops! missed that!

diox:
> It's not totally clear to me why we have this test? Are we testing the behaviour of django/mysql here?

We're testing the behavior of our custom field (making sure that  under the hood it generates the right MySQL column, with the correct behavior we expect)

> I'm happy to fix the test, one idea could be that we use a update_or_create instead of a create. Another is to implement a dummy model to use for this test that is only used there. Finally, we could make this test delete all the records before running to make sure it is clean at time of execution. Happy to do whatever you want here.

The custom model is probably the cleanest but I'm fine with either of these options. We have several of these already, see the various `testapp` apps that are installed only in `settings_test.py`.

> I'm wondering though if this test should be run on a clean DB though. Do we have a concept of pure/unpure DB tests? I've seen this in the past so wondering if it's relevant.

We don't, all tests using the db are using the full migrations.

src/olympia/access/migrations/0003_add_force_send_mail_group.py
KevinMind: Done.

KevinMind:
@diox is it possible that we should be using this initial.json file for defining the new group?

There's a serializer that is searching for a group based on pk=1 (broken by the migration) and that is runtime code. So our app has very specific opinions about which groups exist, and what their IDs are.

https://github.com/mozilla/addons-server/blob/04c983890c1fb5f2d8a487df634b30213d483530/src/olympia/landfill/serializers.py#L321-L324

Scope here is creeping so not totally sure how to proceed. I would suggest we use whatever approach we have used in the past for creating groups. It seems that is not migration. Or?

diox:
The problem with the `initial.json` file is that they are only executed once for a given env (unless you do a full reset). And unfortunately, they are executed after migrations (including data migrations), which makes sense but is annoying here.

Maybe you could stick that group in `initial.json` (for the tests / new envs), and keep the data migration for existing envs, using `get_or_create` with the exact same `name` ? It's a bit fragile, but...

At the same time, we should move away from any hardcoded ids for Groups. AFAIK we only do this for the admin group in landfill (which is also only used for one-time setup for new envs), so you could just change that broken code to `Group.objects.get(name="Admins")`.

KevinMind:
Tests seem to be passing with the changes here so far. Going to reset review to see if we're still happy with this approach or not.

src/olympia/amo/mail.py
```diff
@@ -17,21 +17,31 @@ class DevEmailBackend(BaseEmailBackend):
...
     """
 
+    force_send_mail_group = 'Force Send Mail'
+
+    def __force_send_emails(self):
+        try:
+            dev_email_group = Group.objects.get(name=self.force_send_mail_group)
+            return dev_email_group.users.values_list('email', flat=True)
+        except Group.DoesNotExist:
+            return []
```
diox, line 30: Simpler version that does a single query (need to import `UserProfile`):
```suggestion
        return UserProfile.objects.filter(
            groups__name=self.force_send_mail_group
        ).values_list('email', flat=True)
```

diox (approved):

src/olympia/amo/mail.py
KevinMind: Done.

PR #21789, opened 2024-02-01
monitor for possible abuse reporting fails and email alerts

eviljeff:
WIP for #21769 - based on the initial proposal to email people directly, which is now probably not what we want...

Issue #21769, opened 2024-01-30
Add abuse reporting health monitoring task


eviljeff:
part of https://mozilla-hub.atlassian.net/browse/AMO-158

requested health checks:

* AMO-Cinder integration issue
Once a day, check for any abuse reports that have been created more than an hour ago and don’t have a cinder job yet
* Add-ons operations issue
Once a day, check for any abuse reports created more than X business days ago that have a cinder job ID but doesn’t have an AMO decision. 
* Cinder operations issue
Once a day, check for any abuse reports created more than Y business days ago that have a cinder job ID but don’t have a cinder decision.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-679)

diox:
If going the redash route (would need the `abuse_cinderjob` table exposed in redash), we can use the following queries (generated from the ORM, tweaked manually to use `DATE_SUB`)

> AMO-Cinder integration issue
> Once a day, check for any abuse reports that have been created more than an hour ago and don’t have a cinder job yet

```sql
SELECT `abuse_reports`.`id`
FROM `abuse_reports`
WHERE (`abuse_reports`.`cinder_job_id` IS NULL
       AND `abuse_reports`.`created` < DATE_SUB(NOW(), INTERVAL 1 HOUR)
       AND `abuse_reports`.`reason` IN (11,
                                        12,
                                        13,
                                        14));
``` 

> Add-ons operations issue
> Once a day, check for any abuse reports created more than X business days ago that have a cinder job ID but doesn’t have an AMO decision.

```sql
SELECT `abuse_cinderjob`.`id`
FROM `abuse_cinderjob`
LEFT OUTER JOIN `abuse_reports` ON (`abuse_cinderjob`.`id` = `abuse_reports`.`cinder_job_id`)
LEFT OUTER JOIN `abuse_cinderjob` T3 ON (`abuse_cinderjob`.`id` = T3.`appeal_job_id`)
LEFT OUTER JOIN `abuse_reports` T4 ON (T3.`id` = T4.`cinder_job_id`)
WHERE (`abuse_cinderjob`.`decision_action` IN (0,
                                               3)
  AND NOT (EXISTS
             (SELECT 1 AS `a`
              FROM `abuse_cinderjob` U0
              LEFT OUTER JOIN `abuse_reports` U1 ON (U0.`id` = U1.`cinder_job_id`)
              WHERE (U1.`guid` IS NULL
[...]

PR #21805, opened 2024-02-05
Add statsd pings around abuse/mail related tasks

diox:
Relates to #21769

This adds some statsd pings around abuse and email tasks, so we can build graphs of the volume of calls. Alerting will be done separately though some redash queries.

Supersedes #21789


diox:
I was thinking we'd only alert through redash anyway... But I can add failures too, might come in handy someday (and should always be at 0 in normal conditions)

diox:
https://github.com/mozilla/addons-server/pull/21805/commits/14bdfb35185644a224e1ce5fa908aef966bc9d6e adds the failure pings

PR #21822, opened 2024-02-07
Add CI workflow and update Makefile-docker

KevinMind:
fixes: mozilla/addons#5872

## Description

Using a github action to extract l10n strings on push to master. It sets up the environment, and runs the extract and push scripts. This would result in a push on master if there are new locales.

## Context

The action approach bypasses the token issues we had in previous attempts made with CircleCI as action runners can push to their repositories automatically. It's also safe because this job can only run on a push to master in our repository, so we do not need any additional potentially vulnerable credentials and modifications to our actions are in our control already.

Right now, there are two approaches, 1 using docker compose and make commands to execute. This works but is CPU expensive. The other approach attempts to install python and our bare minimum needed dependencies to execute. We will probably need to refactor this further to make it useful as it's currently not working.

I wanted to open this up for review already to make sure we align on the initial idea and if we do I can make those changes and get the PR ready to merge.

## Testing

In the long term this action can be tested outside of master by pushing to a branch called `1extract-locales`, that way we can freely tweak the code without needing pushes to master.


Makefile-os
```diff
@@ -50,7 +50,13 @@ initialize_docker: create_env_file
 # which is wrong for local development.
 	docker compose run --rm --user ${UID}:${GID} web make update_deps
 	docker compose up -d
-	docker compose exec --user olympia web make initialize
+	docker compose exec -u 0 web make initialize
```
diox, line 53: See comment above that in the Makefile: this change would subtly break local environments in some scenarios, because `initialize` runs `update_deps`, deps which are updated with the user `olympia` everywhere else (as they live in the cloned directory locally). You'd possibly end up with `root` owning some files/directories, making it impossible to update with `olympia` later.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
...
+      CI_COMMIT_AUTHOR: Continuous Integration
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+
+    - name: Start containers
+      run: |
+        docker compose pull
+        make initialize_docker
```
diox, line 22: Can you avoid this and run the extraction on the host ? We run it inside a fully built local env because it's done manually by us, but it doesn't need a database, memcached, celery or even nginx, so that's just a lot of wasted CPU, especially with initialize creating base data as well.

Makefile-os
KevinMind: I know, I just needed a workaround cause I didn't want to spend too much time. Don't plan to merge this.

.github/workflows/ci.yml
KevinMind: I will try it. I also wanted to see if I could run tests, which i can.

KevinMind:
```
service "web" is not running
```

KevinMind: Will try starting without initializing.

diox: mmm the script currently assumes it's running on a host where docker containers run, so it would need some tweaking, yeah. Avoiding initializing is a good step, but ideally we wouldn't even have docker at all here.

KevinMind: I think there is a lot of room for improvement there, but if it takes 10 minutes but runs successfully and doesn't require any kind of auth support and also doesn't have security holes for forks, then I would bet on this. The upside is there and it has none of the downsides AFAICT.

KevinMind:
https://github.com/mozilla/addons-server/assets/19595165/4067ceb4-dd3e-4e0a-8faf-e21fb4218231


You can confirm that the second commit which was generated by this branch is not triggering the job to create another commit, thus avoiding infinite loops ([link](https://joht.github.io/johtizen/build/2022/01/20/github-actions-push-into-repository.html#the-easiest-solution))

KevinMind:
Example run that did not push https://github.com/mozilla/addons-server/actions/runs/7819322737/job/21331523440?pr=21822

KevinMind:
@diox we should probably add a filter on our circleci master release to ignore pushes made by this job. No need to build a container for strings.

KevinMind:
Job, pushing changes https://github.com/mozilla/addons-server/actions/runs/7834008833/job/21376296237?pr=21822

PR #21829, opened 2024-02-08
expose addon_version in details of cinder jobs being resolved

### LINKED ISSUE #21797: expose reported addon version for abuse reports in reviewer tools
If a user has an add-on installed in their Firefox when they report abuse for it we record the version in the abuse report instance.  We expose this information in the full list of abuse reports available in the reviewer tools review page, but don't for the checkboxes that allow jobs to be resolved with the review.  (We do, however, make sure to flag that version specifically for human review).

We should include the reported version in the `<details>` for the abuse reports for each job.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-685)

eviljeff:
fixes #21797 in a very (very) minimal way - just changes it to be "v[version_number]: abuse.message" e.g. "v[1.23]: This add-on does bad thing" if there is an `addon_version` in the AbuseReport instance; keeps it to just the abuse message otherwise.

### DIFF
--- a/src/olympia/reviewers/forms.py
+++ b/src/olympia/reviewers/forms.py
@@ -214,14 +214,20 @@ class CinderJobChoiceField(ModelMultipleChoiceField):
         reasons_set = {
             (report.REASONS.for_value(report.reason).display,) for report in reports
         }
-        messages_gen = ((report.message,) for report in reports)
+        messages_gen = (
+            (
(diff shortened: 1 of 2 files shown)


eviljeff:
this is low priority but it was done so 🤷

diox (approved):

PR #21838, opened 2024-02-09
Merge requirements files and document dependency management

KevinMind:
fixes: mozilla/addons#9461

## Description

This PR merges many of our existing requirements files to simplify installation of python/node dependencies. I've extracted our installation logic to a bash script, and consolidated the requirements files to 3 pip, prod, and dev.

## Context

We have to keep pip.txt as a singleton so that:
- we ensure the correct version of pip is always installed before installing anything else
- pip can be upgraded via dependabot

## Testing

We do not build our docker containers in PRs so we won't be able to verify anything is broken. But you can build locally by pruning your local containers, if you have them, and rerunning the build following our docs.


scripts/update_deps.sh
eviljeff, line 1: Does having this extracted out into a separate file give us anything v.s. having it in the Makefile?

docs/topics/development/dependencies.md
```diff
@@ -0,0 +1,62 @@
...
+
+Dev dependencies are used by our django app in development or by tools we use for linting, testing, etc.
+
+To ensure a standard and sane process for installing packages,
+use the `make install` command to install a new package.
+This will add your specified version to one of our requirements/(dev|prod).txt files and
+install the package in your local container.
+
+```bash
+make install [package]==[version] --dev --prod
```
eviljeff, line 26: 😕 where is this make recipe defined?

docs/topics/development/dependencies.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+Note: This script is strict. You must specify a version, and you must specify either dev or prod.
+
+### Upgrading Python Dependencies
+
+TBD
```
eviljeff, line 33: missing content?

docs/topics/development/dependencies.md
KevinMind: It's not. I realized that now, I was imagining this would be good to have cause I needed to install several packages while working on this and I never remember how to do it.

But then I moved on and didn't actually implement it. I'll remove for now and keep the older manual process.

docs/topics/development/dependencies.md
KevinMind: Yep, how do we normally handle upgrades? is it exclusively dependabot? Do we have a process for managing dependabots typically? IDK what to put but knew there should be something, hence the empty section.

scripts/update_deps.sh
KevinMind: Initially working on this there was a lot more going on so I extracted, but now it's much simpler. 

TLDR; Not really at this point no. But I kind of like it.

I think this is better than what we had, which was a lot of make commands that hook onto each other. This is a single script. I could copy this into the make file directly, but I think having the logic encapsulated in a script is kind of nice, makes the makefile smaller and lets you focus on this specific chunk, no dependencies getting jumbled and scrambled together. Wdyt?

eviljeff: It's in the makefile already so if it doesn't give us anything I'd not copy it out. No point in extra change

docs/topics/development/dependencies.md
eviljeff: Yeah, dependabot. But ... well, there are always exceptions

Dockerfile
```diff
@@ -97,7 +97,7 @@ ENV PATH $PYTHONUSERBASE/bin:$PATH
 ENV NPM_CONFIG_PREFIX=/deps/
 RUN ln -s ${HOME}/package.json /deps/package.json \
     && ln -s ${HOME}/package-lock.json /deps/package-lock.json \
-    && make update_deps
+    && make update_deps --prod
```
KevinMind, line 100: should be 

```bash
make update_deps ARGS="--prod"
```

KevinMind: I also noticed that because we are copying the root directory before running update_deps, we might not be allowing docker the ability to cache deps across builds. I would just fix this now, unless you think it really should go to a different PR?

diox: > I also noticed that because we are copying the root directory before running update_deps, we might not be allowing docker the ability to cache deps across builds. I would just fix this now, unless you think it really should go to a different PR?

Could you expand on this? It's not immediately clear what you're proposing.

KevinMind: The way we organize our dockerfile is not the most efficient when it comes to taking advantage of how docker caches layers. We copy our source code too early. 

I tried getting this to work, it can be done but it is definitely too much for this PR. I'll open a separate ticket for this explaining more.

KevinMind: Also just noting, I wonder if adding those "shell" dependencies to prod.txt is the best solution long term. I think having a dedicated instance of addons-server which has prod/dev installed might be better since we don't need to bloat all the prod images with stuff that only needs to be available in one place.

WE could even not build a separate image but in the deployment install the dev depdencies. Then if you want to shell, you go to the "shell" pod and not web or worker. Just a thought but I think it would save us space on the images and time on the builds.

KevinMind: Probably better from a ergonomics / security perspective as we never ssh into web or worker pods, but only the dedicated shell pods.

diox: Yes, typically in prod SRE will only spawn django shells from a dedicated admin pod I think - we're less strict in dev/stage. Down the line it's a good idea (that's why `dev.txt` was separate in the first place) but out of scope for now.

KevinMind: I get this:

```bash
ipython==8.20.0
├── decorator [required: Any, installed: 5.1.1]
├── exceptiongroup [required: Any, installed: 1.2.0]
├── jedi [required: >=0.16, installed: 0.19.1]
│   └── parso [required: >=0.8.3,<0.9.0, installed: 0.8.3]
├── matplotlib-inline [required: Any, installed: ?]
├── pexpect [required: >4.3, installed: 4.9.0]
│   └── ptyprocess [required: >=0.5, installed: 0.7.0]
├── prompt-toolkit [required: >=3.0.41,<3.1.0, installed: 3.0.43]
│   └── wcwidth [required: Any, installed: 0.2.13]
├── pygments [required: >=2.4.0, installed: 2.17.2]
├── stack-data [required: Any, installed: 0.6.3]
│   ├── asttokens [required: >=2.1.0, installed: 2.4.1]
│   │   └── six [required: >=1.12.0, installed: 1.16.0]
│   ├── executing [required: >=1.2.0, installed: 2.0.1]
│   └── pure-eval [required: Any, installed: 0.2.2]
└── traitlets [required: >=5, installed: 5.14.1]
```

I moved everything that already existed into prod.txt, several deps were not even specified in either .txt (Iguess that's fine?>..) 

managing python dependencies is like playing piñata without the candy.

requirements/pip.txt
```diff
@@ -0,0 +1,6 @@
+## Requirements needed to be installed *before* any other requirements.
+## This makes sure we have the latest version of pip, setuptools and relevant
+## system packages first.
```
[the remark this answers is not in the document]
KevinMind: I just copied the text as is, I can remove the others or should I add setuptools and "relevant system packages" what does that include? I'd rather keep it just pip tbh. Don't even really like that but I understand we need to make dependabot happy.

scripts/update_deps.sh
KevinMind: I'm not sure I understand. Is your argument that because it is in the Makefile, that it should stay in the Makefile? 

```
 think having the logic encapsulated in a script is kind of nice, makes the makefile smaller and lets you focus on this specific chunk, no dependencies getting jumbled and scrambled together.
```

I think this is a pretty good argument after looking at it again. When working with the Makefile(s) I already found it incredibly difficult to organize which commands relied on which, which were meant to be called internally vs externally. I see this as a valuable change that makes it simpler and safer. 

Flipping the question around, do you have an argument why that is a) not true or b) some additional benefit we get from "having" it in the makefile?

KevinMind: Also we have existing scripts that are encapsulated in .sh files instead of directly in make so this isn't a new pattern or anything we haven't seen before.

docs/topics/development/dependencies.md
KevinMind: IF you write some of the exceptions you can think of in this thread, I can add them to the doc.

scripts/update_deps.sh
KevinMind: Some additional context: I'd actually like to move the npm installation bits into that same script as well but the change is a bit too big for this PR.

Dockerfile
diox: We've purposefully ignored `matplotlib-inline` because we shouldn't need it for our usage. Everything else is already specified in requirements.

scripts/update_deps.sh
```diff
@@ -0,0 +1,25 @@
...
+# Cleanup python build directory
+rm -rf /deps/build/*
+
+$PIP_COMMAND install --progress-bar=off --no-deps --exists-action=w -r "requirements/pip.txt"
+
+info "Installing prod dependencies"
+$PIP_COMMAND install --progress-bar=off --no-deps --exists-action=w -r "requirements/prod.txt"
+
+# Install dev dependencies unless --prod is passed
+if [[ ! "$@" == *"--prod"* ]]; then
```
KevinMind, line 16: @eviljeff @willdurand how would I do this in the make file? Just curious as it wasn't obvious that I could add conditions like this at least not in a way that didn't seem much simpler in a bash script.

diox: You can evaluate args in Makefiles as well but that is generally not the recommended way of doing this. If this was still in a `Makefile` the easiest way to do it IMHO would be to have different targets, with the "prod" target depending on the dev one, which is more or less what we had before.

diox (changes requested):
Can you keep using `Makefile` targets for now to bring the changes count down? We can re-evaluate moving to a script in the next iteration of the changes we want to do to deps handling

scripts/update_deps.sh
KevinMind: Refactored to use the makefile.

scripts/update_deps.sh
KevinMind: The only way to make this really work the way we want was to extract a `update_deps_base` target. This allows us to install the pip and node modules, install olympia, clean the build dir etc.

Then we have the original `update_deps` which acts as a "dev" dependencies which installs python dev deps and finally a `update_deps_prod` which prunes the node modules dev dependencies.

diox (approved):

PR #21839, opened 2024-02-09
Don't record parent CinderPolicy on top of a child policy

### LINKED ISSUE #21827: Parent policy is shown separately in a different line in the target/author emails. -DEV
Parent policy is shown separately in a different line in the target/author emails. -DEV

### What happened?
image
Parent policy is shown in a separate line
For the policy section to contain only this:
- Acceptable Use, specifically Acceptable Use: Harassment
This is related to issue: https://github.com/mozilla/addons-server/issues/21742

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-697)

diox:
We don't want to include the parent when the child policy has already been selected, it's redundant and would mess up the notification.

Fixes #21827

### DIFF
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -843,10 +843,13 @@ class CantBeAppealed(Exception):
 
 class CinderPolicyQuerySet(models.QuerySet):
     def without_parents_if_their_children_are_present(self):
+        """Evaluates the queryset into a list, excluding parents of any child
+        policy if present.
+        """
         parents_ids = set(
             self.filter(parent__isnull=False).values_list('parent', flat=True)
         )
-        return self.exclude(pk__in=parents_ids)
+        return list(self.exclude(pk__in=parents_ids))
 
 
 class CinderPolicy(ModelBase):
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -340,7 +340,10 @@ class CinderJob(ModelBase):
                 : self._meta.get_field('decision_notes').max_length
             ],
         )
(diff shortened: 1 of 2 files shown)


src/olympia/abuse/tests/test_models.py
```diff
@@ -1153,3 +1252,27 @@ def test_create_cinder_policy_with_duplicate_uuid(self):
...
+            name='Policy 2',
+            text='Policy 2 Description',
+            uuid='some-uuid',
+        )
+        qs = CinderPolicy.objects.all()
+        assert set(qs) == {parent, child_policy, lone_policy}
+        assert set(qs.without_parents_if_their_children_are_present()) == {
+            child_policy,
+            lone_policy,
+        }
```
eviljeff, line 1278: can we add a `.exclude(id=child_policy.id)` after to confirm that `parent` would be there if it's children weren't?

src/olympia/abuse/models.py
```diff
@@ -840,6 +841,14 @@ class CantBeAppealed(Exception):
     pass
 
 
+class CinderPolicyQuerySet(models.QuerySet):
+    def without_parents_if_their_children_are_present(self):
+        parents_ids = set(
+            self.filter(parent__isnull=False).values_list('parent', flat=True)
+        )
```
eviljeff, line 848: I've got a concern that by evaluating `self.filter` at this point it will mean that `without_parents_if_their_children_are_present` _has_ to be last in a chain of `exclude`|`filter`or `parents_ids` will be wrong.  Is there a way to write this without the it? 😬

eviljeff (approved):

PR #21843, opened 2024-02-12
Improve logging for suppressed email confirmation task + fix relative URL in verification email

KevinMind:
FIxes: #21687

## Description

This pull request improves the logging for the suppressed email confirmation task. It adds the site URL to the verify email template and updates the check_suppressed_email_confirmation function to provide more detailed error messages.

## Context

In testing this feature on dev, it is sometimes difficult to understand the reason why `check_suppressed_email_confirmation` is retrying. It could be because of bad network response, or that the response returned no results, or that the response returned results and non matched our target email. This PR should make distinguishing those cases easier.

## Testing

Test in dev following the linked issue.

### DIFF
--- a/src/olympia/devhub/templates/devhub/emails/verify-email-requested.ltxt
+++ b/src/olympia/devhub/templates/devhub/emails/verify-email-requested.ltxt
@@ -3,7 +3,7 @@
 You're receiving this message because you requested to verify your email.
 To finish this process, please click on the link below:
 
-{{ confirmation_link }}
+{{ settings.SITE_URL }}{{ confirmation_link }}
 
 If you didn't request this, please notify amo-admins@mozilla.com.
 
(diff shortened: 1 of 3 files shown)


src/olympia/devhub/templates/devhub/emails/verify-email-requested.ltxt
```diff
@@ -3,7 +3,7 @@
 You're receiving this message because you requested to verify your email.
 To finish this process, please click on the link below:
 
-{{ confirmation_link }}
+{{ settings.SITE_URL }}{{ confirmation_link }}
```
diox, line 6: We have a `absolutify` filter that takes care of this, it handles the URL already starting with `http://` or `https://`* and works even when `SITE_URL` has been overridden (shouldn't happen in your specific case, it's for addons-server containers meant for admin tools behind VPN, but still)

It can be called as a function when generating context - which might be better in this case, cause emails are processed with Django Template Langage, and I'm not sure whether or not we register that filter outside of Jinja2

KevinMind: Done.

diox (approved):

Issue #21850, opened 2024-02-12
Sync Cinder Policies Failing


KevinMind:
Relates to: #21747

We ran the command to sync cinder policies in production on 2024.02.08 around 6:47 CET.

```bash
olympia@addons-server-v1-deploy-uwsgi-amo-web-5ffcf86c56-qdxgd:~$ ./manage.py sync_cinder_policies
{"Timestamp": 1707414460025349376, "Type": "z.abuse", "Logger": "http_app_addons", "Hostname": "addons-server-v1-deploy-uwsgi-amo-web-5ffcf86c56-qdxgd", "EnvVersion": "2.0", "Severity": 6, "Pid": 68, "Fields": {"uid": "<anon>", "remoteAddressChain": "", "msg": "Triggered policy sync task."}, "severity": 200}
```

However when we inspected for cinder policies, we could not find any.

We do see that a statsd [ping incremeneted](https://earthangel-b40313e5.influxcloud.net/explore?orgId=1&left=%7B%22datasource%22:%22000000006%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22influxdb%22,%22uid%22:%22000000006%22%7D,%22resultFormat%22:%22time_series%22,%22orderByTime%22:%22ASC%22,%22tags%22:%5B%5D,%22groupBy%22:%5B%7B%22type%22:%22time%22,%22params%22:%5B%22$__interval%22%5D%7D,%7B%22type%22:%22fill%22,%22params%22:%5B%22null%22%5D%7D%5D,%22select%22:%5B%5B%7B%22type%22:%22field%22,%22params%22:%5B%22value%22%5D%7D,%7B%22type%22:%22mean%22,%22params%22:%5B%5D%7D%5D%5D,%22policy%22:%22autogen%22,%22measurement%22:%22addons-prod.abuse.tasks.sync_cinder_policies.failure%22%7D%5D,%22range%22:%7B%22from%22:%22now-30d%22,%22to%22:%22now%22%7D%7D) a failure of this job around when it was executed. 

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-704)

KevinMind:
Corresponding sentry error: https://mozilla.sentry.io/issues/4957875686/?project=6310819

KevinMind:
Agreed to add monitor for cinder in the scope of this task to ensure we have a stable, authenticated connection in prod.

[...]

PR #21851, opened 2024-02-12
Add logs to sync_cinder_policies + trigger sync in management command

KevinMind:
Fixes: #21850

This pull request adds logs to the `sync_cinder_policies` function and triggers the sync in the management command.


eviljeff:
you've got CI fails

diox:
I don't think we want/need this. Failures are reported in statsd and exposed in logs+sentry. Task start/stop are also logged as with all our other tasks.

PR #21856, opened 2024-02-13
Improve Cinder Policy / Review Action Reasons admin

### LINKED ISSUE #21855: Improve Cinder Policy / Review Action Reason admin
In the admin:

- Expose id, parent and text in Cinder Policy list view 
- Make it easier to find the right policy in review action reason admin

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-706)

diox:
- Make more fields visible to distinguish between the policies
- Use raw id field for Cinder policy in Review Action Reason admin to allow for a more efficient lookup + easy copy/paste of data

Fixes #21855

### DIFF
--- a/src/olympia/abuse/admin.py
+++ b/src/olympia/abuse/admin.py
@@ -353,17 +353,32 @@ class AbuseReportAdmin(AMOModelAdmin):
 
 class CinderPolicyAdmin(AMOModelAdmin):
     fields = (
+        'id',
         'created',
         'uuid',
+        'parent',
         'name',
         'text',
     )
     list_display = (
+        'id',
         'uuid',
+        'parent',
         'name',
+        'text',
     )
-    readonly_fields = ('created',)
+    ordering = ('parent__name', 'name')
+    list_select_related = ('parent',)
     view_on_site = False
+
+    def has_add_permission(self, request):
+        return False
+
+    def has_change_permission(self, request, obj=None):
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        return False
 
 
 admin.site.register(AbuseReport, AbuseReportAdmin)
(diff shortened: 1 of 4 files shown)


src/olympia/abuse/admin.py
```diff
@@ -353,18 +353,33 @@ def message_excerpt(self, obj):
...
+        'parent',
         'name',
         'text',
     )
     list_display = (
+        'id',
         'uuid',
+        'parent',
         'name',
+        'text',
```
eviljeff, line 368: might be a become a bit busy with the text, but 🤷

src/olympia/abuse/models.py
```diff
@@ -873,10 +873,10 @@ class CinderPolicy(ModelBase):
     objects = CinderPolicyQuerySet.as_manager()
 
     def __str__(self):
-        return self.name
+        return self.full_text('')
 
     def full_text(self, canned_response_text=None):
-        if not canned_response_text:
+        if canned_response_text is None:
             canned_response_text = self.text
```
eviljeff, line 880: This would mean that a policy from cinder would never be overridden by a ReviewerReason canned response, even if the canned response was blank.  (If we don't want that it'd need a tweak to ReasonsChoiceWidget)

eviljeff (approved):

src/olympia/abuse/admin.py
diox: It will, but no worse than in Cinder, and I think it's helpful to distinguish between all the noise

src/olympia/abuse/models.py
diox: > I'm assuming all active `ReviewActionReason` do have a canned response anyway though.

This is true in prod atm.

PR #21858, opened 2024-02-14
Add a delayed rejection action and send appropriate emails

eviljeff:
fixes #21811 

- a new decision action - `AMO_REJECT_VERSION_WARNING_ADDON` is send for a delayed rejection
  - if the rejection goes ahead the decision of `AMO_REJECT_VERSION_ADDON` will be sent as now
- the reminder email is sent as now, but the correct DSA one
- An email is sent out when the content is actually rejected (new requirement)
- reporters get an email when the delayed rejection is scheduled, and when (if) the content is rejected. (not a reminder)
- It relies on a M2M between `VersionReviewerFlags` and `CinderJob` to indicate there is an outstanding cinder job to be resolved.
  - (could be `Version` directly? - But `VersionReviewerFlags` is where the other pending_rejection properties lived)


src/olympia/abuse/templates/abuse/emails/CinderActionRejectVersion.txt
eviljeff, line 1: this is based on the latest email copy in the doc, but there are still unresolved comments

src/olympia/abuse/models.py
```diff
@@ -345,14 +346,16 @@ def process_decision(
...
             existing_decision != decision_action
             or decision_action == self.DECISION_ACTIONS.AMO_APPROVE
         ):
             action_helper.notify_reporters()
-        if action_occured:
-            action_helper.notify_owners()
+        if log_entry_or_true:
+            action_helper.notify_owners(
+                log_entry_id=getattr(log_entry_or_true, 'id', None)
+            )
```
diox, line 358: Can you think of a better way of doing this ? Returning either a boolean, or an activity log, or None is quite confusing (setting a property out of nowhere isn't that much better...)

eviljeff: mmm, well I had `log_entry_id` be a property on the `CinderAction` instance before but setting it was hacky with `resolve_job`.  I could still do that and have something like `log_entry_id = log_entry_id or self.log_entry_id` in `notify_owners`?

eviljeff: I'm thinking eventually that all the `process_action`s will return an activity log instance, but that'll need some rewriting where the functionality is already delegated somewhere else.  Really, it's only useful to know the log id for add-on targets anyway - it's used for the activitylog emails, and that's only a thing where there is an add-on involved.

Could have a tuple, yep.

diox (approved):

PR #21866, opened 2024-02-15
encode certain characters that Cinder chokes on

### LINKED ISSUE #21863: Email is not sent after version reject when the DSA related reports are checked
I tried to test https://github.com/mozilla/addons-server/issues/21816 with Reject versions

When I select the Outstanding[ DSA related reports to resolve ](https://reviewers.addons.allizom.org/en-US/reviewers/review-listed/1008599#review-actions) and reject e-mail is not sent to the author. Also the reports continue to be displayed in the section

examples
https://reviewers.addons.allizom.org/en-US/reviewers/review-listed/1008599#review-actions
https://reviewers.addons.allizom.org/en-US/reviewers/review-listed/1022919#review-actions

When the version is rejected without checking the Cinder reports the email is sent to the author. 

I tested this on stage.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-708)

eviljeff:
fixes #21863

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -20,6 +20,7 @@ class CinderEntity:
     type = None  # Needs to be defined by subclasses
     # Number of relationships to send by default in each Cinder request.
     RELATIONSHIPS_BATCH_SIZE = 25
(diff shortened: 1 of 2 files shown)


eviljeff:
There's other ways of doing it, but I was originally thinking about swapping out characters, rather than prepending, so I went with `translate`

diox (approved):
(The fact that we have to do) This is terrible but it will do...

PR #21878, opened 2024-02-16
Use specific waffle switch to gate reviewer tools cinder integration

### LINKED ISSUE #21877: Gate reviewer tools reports resolving feature behind a specific switch
We're currently gating the ability to see and resolve related reports in reviewer tools review page behind the `enable-cinder-reporting` switch. We discovered scalability issues with that feature though, so we'd like to turn it off while potentially keeping cinder integration live, so this should be behind a separate switch.

A follow-up task would be to find reports that were "ignored" as a result of turning that feature off.

diox:
This allows us to enable cinder full integration while not showing reports in reviewer tools.

Fixes #21877


diox:
(Merging without a review to get that on dev/stage sooner)

willdurand (commented):
r+

PR #21882, opened 2024-02-19
Improve performance of `CinderJob.resolvable_in_reviewer_tools()` / `CinderJob.for_addon()`

### LINKED ISSUE #21880: Improve performance of CinderJob.resolvable_in_reviewer_tools()
To differentiate between `CinderJob`s that are handled by T&S moderators and AMO Reviewers, we have introduced a `resolvable_in_reviewer_tools()` method that is used in the AMO review page to find jobs that can be resolved as part of the review of a given add-on.

The query that this method produces contains multiple `JOIN`s on `abuse_reports` and itself (because of appeals - which are a job on top of another job). This is made worse in the review page where we also need to filter by target, with `for_addon()`. The result scales very poorly in production where there are a lot of abuse reports, to the point where we had to [temporarily disable showing the jobs on the review page](https://github.com/mozilla/addons-server/issues/21877).

We should implement that query differently to solve this problem. Couple ideas to look into:
- Decouple from `abuse_reports` by storing the target on the `CinderJob` when it's created (this would later help to implement forwarding decisions to Cinder even if they are made on our own initiative without a report - though fixing all existing code to allow that is out of scope here)
- Decouple from `abuse_reports` by storing a boolean on the `CinderJob` indicating that this is meant for AMO Reviewers when it's created or escalated from Cinder.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-716)

diox:
Fixes #21880

### Description
Add denormalized fields to help figure out whether a `CinderJob` is for a particular add-on and whether it's meant for AMO reviewers, which greatly simplifies the corresponding queries, improving their performance.

- `target_addon` is set at reporting time and never changes after
- `resolvable_in_reviewer_tools` is also set at reporting time but can be flipped when processing escalations from Cinder.

No extra index is added besides the one automatically added for the new `target_addon` foreign key - that should be enough for now.

A command is added to backfill the existing data (it was getting a bit too hairy to write as a data migration, since you can't use custom model methods & properties there).

This should be transparent with no visible change anywhere.

### Context

When `enable-cinder-reporting` switch is enabled, each abuse report is linked to a `CinderJob`. Those jobs are processed by either moderators in Cinder or AMO reviewers in reviewer tools depending on various factors (`AbuseReport.is_handled_by_reviewers`). A job can also be escalated to AMO Reviewers (`CinderJob.decision_action == CinderJob.DECISION_ACTIONS.AMO_ESCALATE_ADDON`).

In Reviewer Tools review page, when `enable-cinder-reviewer-tools-integration`  switch is enabled, for certain actions like `Force Disable`, reviewers have the option to resolve jobs for abuse reports associated with the add-on they are reviewing. That is the query that was breaking prod that this PR aims to solve.

### Testing

Automated tests added should cover the fact that the denormalized fields are set correctly when needed.

To test manually, make sure to enable `enable-cinder-reporting` and `enable-cinder-reviewer-tools-integration` waffle switches, set a valid `CINDER_API_TOKEN`, then report a public listed add-on that exists in your local environment for "It violates Add-on Policies" reason and "inside the add-on" as the place of violation using the frontend: the add-on should then appear in reviewer tools queue, and the report be displayed towards the bottom of the review page next to " Outstanding DSA related reports to resolve" for relevant actions, such as "Force Disable").

You won't be able to test escalations coming from Cinder without calling the webhook URL directly since it requires Cinder to call a webhook on our end.

Performance can't easily be tested outside of production, but you can look at the queries the ORM would generate before/after these changes for the affected methods, particularly when chaining them with `CinderJob.objects.for_addon(self.helper.addon).unresolved().resolvable_in_reviewer_tools()` [like we do in reviewer tools](https://github.com/mozilla/addons-server/blob/5c021235c668a2d7ce69d0e8aa08f8584a09c5b1/src/olympia/reviewers/forms.py#L418-L429).


src/olympia/addons/migrations/0036_add_allow_deleted_guid_reuse_switch.py
```diff
@@ -16,6 +16,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ('addons', '0035_auto_20210812_1726'),
+        ('waffle', '0001_initial'),
```
diox, line 19: This is an old migration that created a waffle `Switch`. It previously worked despite not requiring `waffle` because of the way the dependencies resolver ordered the migrations, `waffle` ended up being required first anyway. 

Because the migration I'm adding in `abuse` depends on `addons`, the order ends up being different and could fail if the dependency on `waffle` wasn't added explicitly.

src/olympia/addons/migrations/0039_auto_20220531_0953.py
```diff
@@ -30,6 +30,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ('addons', '0038_auto_20220530_1639'),
+        ('blocklist', '0027_auto_20220530_1639'),
```
diox, line 33: As above

src/olympia/blocklist/migrations/0024_delete_legacyimport.py
```diff
@@ -12,6 +12,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ('blocklist', '0023_auto_20201103_1108'),
+        ('waffle', '0001_initial'),
```
diox, line 15: As above

diox:
TODO: need an extra migration for existing `CinderJob` unfortunately, since cinder integration went live last Thursday.

src/olympia/abuse/models.py
```diff
@@ -146,11 +122,17 @@ class CinderJob(ModelBase):
         # appeal for multiple previous decisions (jobs).
         related_name='appealed_jobs',
     )
+    target_addon = models.ForeignKey(
+        to=Addon, blank=True, null=True, on_delete=models.deletion.SET_NULL
```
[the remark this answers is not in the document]
KevinMind: Why blank **and** null? If blank, then why not `editable=False` also?

src/olympia/abuse/models.py
```diff
@@ -390,7 +379,11 @@ def appeal(self, *, abuse_report, appeal_text, user, is_reporter):
             appealer=appealer_entity,
         )
         with atomic():
-            appeal_job, _ = self.__class__.objects.get_or_create(job_id=appeal_id)
+            appeal_job, _ = self.__class__.objects.get_or_create(
```
KevinMind, line 382: I guess the reason why you don't separate `taget_addon` and `resolvable_in_reviewer_tools` to defaults in this query (where you do in the report method) is that an "appeal" operation is something that happens on an abuse report that has a decision, and should already have a specified target_addon/resolvable_in_reviewer_tools fields. A "report" operation is something that happens when we initially send an abuse report to cinder and so want to exclude those fields from the query, but set them if the cinder job doesn't already exists. Is that right?

src/olympia/abuse/models.py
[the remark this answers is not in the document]
KevinMind: I think I answered my own question looking at the serializer. We can create cinder jobs for users/addons/collections/ratings and only addons would have a target_addon? Is that right?

src/olympia/abuse/management/commands/fill_cinderjobs_denormalized_fields.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class Command(BaseCommand):
+    help = (
+        'One-off command to fill CinderJob.target_addon and '
+        'CinderJob.resolvable_in_reviewer_tools fields'
+    )
+
+    def handle(self, *args, **options):
+        jobs = CinderJob.objects.all()
```
KevinMind, line 14: Could this get crazy if there are a lot of jobs? Can this be batched?

src/olympia/abuse/management/commands/fill_cinderjobs_denormalized_fields.py
```diff
@@ -0,0 +1,33 @@
...
+
+        for job in jobs:
+            if abuse_report := job.initial_abuse_report:
+                job.target_addon = (
+                    # If the abuse report is against a guid, set target_addon
+                    # on the job accordingly. Handle the add-on not existing
```
KevinMind, line 20: I'm not sure I follow "Handle the add-on not existing in the db just in case."

src/olympia/abuse/management/commands/fill_cinderjobs_denormalized_fields.py
```diff
@@ -0,0 +1,33 @@
...
+        'One-off command to fill CinderJob.target_addon and '
+        'CinderJob.resolvable_in_reviewer_tools fields'
+    )
+
+    def handle(self, *args, **options):
+        jobs = CinderJob.objects.all()
+
+        for job in jobs:
+            if abuse_report := job.initial_abuse_report:
+                job.target_addon = (
```
KevinMind, line 18: Why is the logic here different than [here](https://github.com/mozilla/addons-server/pull/21882/files#diff-a1926bd1323775380703cbe2a00da0de6668413b9afab7815b337f5779d61c24R294-R298)

```python
target = abuse_report.target
job.target_addon = target is addon .... blah blah
```

src/olympia/abuse/management/commands/fill_cinderjobs_denormalized_fields.py
```diff
@@ -0,0 +1,33 @@
+from django.core.management.base import BaseCommand
+
+from olympia.abuse.models import CinderJob
+from olympia.addons.models import Addon
+
+
+class Command(BaseCommand):
```
KevinMind, line 7: Would the plan be to remove this code after running it in prod? Or would we need this command later? Feels like adding code that will just hang around without a plan to remove it.

src/olympia/addons/migrations/0036_add_allow_deleted_guid_reuse_switch.py
KevinMind: This confuses and scares me.

> could fail

Is doing most of the heavy lifting. How did you discover that this _could_ be a problem? How would you have known? Especially if it is not guaranteed to fail, do you just have to model the dependency array like a react hook? Feels very brittle.

KevinMind (commented):
Code seems to make sense as far as I can tell. Would like to test it on my environment.

KevinMind (approved):
Comments:
- use defaults for creating appeals
- use the abuse_report.target property in the command
- nit: consider batching, probably not needed


Tested and verified the queries work in reviewer tools

src/olympia/addons/migrations/0036_add_allow_deleted_guid_reuse_switch.py
diox: Tests did fail (for instance https://app.circleci.com/pipelines/github/mozilla/addons-server/29054/workflows/3b52f3cc-d3b1-4bdf-af4b-caf34e58476b/jobs/190061)

I said "could" because it only reproduced on an empty db: on an already existing one, migration engine would only apply the missing migration and we wouldn't have known.

src/olympia/abuse/models.py
diox: Fixed in https://github.com/mozilla/addons-server/pull/21882/commits/64eac2be7f479da0d1069810aea21b6d58256b2a

Issue #21888, opened 2024-02-19
Various Improvements to our Docker build


KevinMind:
### Description

While investigating/implementing several issues that rely on our docker build setup:
- https://github.com/mozilla/addons-server/issues/21837
- https://github.com/mozilla/addons-server/issues/9112

I found various improvements that can be made to our docker build setup. These improvements can be split into 2 categories:
- incremental improvements
- step-wise improvements

Where the former are small improvements to the speed or bandwidth requirements of the existing overall build strategy, and the latter are larger architectural changes which imply both bigger gains and bigger cost, risk, etc.

This issue is aiming to track PRs for the former. These are changes which we can safely ship today without seriously risking our overall production/ci/local development pipeline.

KevinMind:
Some resources:

- https://github.com/vinodsai-a/github-reusable-workflow-sample/tree/main/actions
- https://docs.docker.com/build/ci/github-actions/cache/
- https://dev.doroshev.com/blog/docker-mount-type-cache/
- https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#run---mounttypecache

KevinMind:
Issues with caching buildkit

- https://github.com/moby/buildkit/issues/1512
- https://github.com/docker/build-push-action/issues/716

diox:
https://docs.docker.com/build/guide/mounts/

KevinMind:
> https://docs.docker.com/build/guide/mounts/

Is there something specific to see here, or just wanting to link the documentation?

diox:
Sorry I just wanted to link to the docs about `--mount` somewhere - and I didn't want to edit your own comment.

KevinMind:
Bug with COPY over caching https://github.com/moby/moby/issues/33107

KevinMind:
https://github.com/docker/build-push-action/issues/225

PR #21890, opened 2024-02-19
Create empty local_settings.py when none exists to avoid erronious "missing local_settings" warning

KevinMind:
Relates to: #21888

### Description

modify settings.py to create a local_settings.py with base_settings import if none exists. This ensures the git ignored local_settings file is always present without polutting it with any contents.

### Context

If you don't have a local_settings.py file, any command that relies on settings.py (almost anything) will print a warning that it could not find the file.

### Testing

You will need to build the container locally and shell into it. I was able to do this with 

```bash
docker build -t test_locale .   
```

Our image doesn't have a command or entry point so you need to define one when running it.

```bash
docker run -d --name test_locale_container test_locale tail -f /dev/null
```

ANd finally shell

```
docker exec -it test_locale_container bash 
```

NOTE: There is no volume, so if you change the code on your machine, it will not be reflected in the container

Now you can run any command that uses the settings.py file

```bash
make update_assets
```

You should not see any warnings about missing local settings file, even if you delete the file and rerun. If you have a local_settings file with a setting, it should not update it.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -137,8 +137,7 @@ local_settings_path = path('local_settings.py')
 if not os.path.exists(local_settings_path):
     with open(local_settings_path, 'w') as file:
         file.write(
-            '# If you have settings you want to overload,'
-            'put them in a local_settings.py.\n'
+            '# Put settings you want to overload in this file.\n'
         )
 
 from local_settings import *  # noqa
--- a/settings.py
+++ b/settings.py
@@ -136,9 +136,7 @@ local_settings_path = path('local_settings.py')
 
 if not os.path.exists(local_settings_path):
     with open(local_settings_path, 'w') as file:
-        file.write(
-            '# Put settings you want to overload in this file.\n'
-        )
+        file.write('# Put settings you want to overload in this file.\n')
 
 from local_settings import *  # noqa
 
--- a/settings.py
+++ b/settings.py
@@ -132,17 +132,16 @@ CUSTOMS_API_KEY = 'customssecret'
 
 REMOTE_SETTINGS_IS_TEST_SERVER = True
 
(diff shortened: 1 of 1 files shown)


diox (commented):
This looks fine but I'm not exactly sure what this solves - at the very least the PR title should be more descriptive.

`local_settings.py` was always optional - we had a warning if it didn't exist, but that's it. 

`settings.py` is the default setting file, you can see it referenced by the `os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'settings')` call in `manage.py`. 
- dev/stage/prod override that `DJANGO_SETTINGS_MODULE` environment variable with their own and don't rely on it at all, importing `settings_base.py` directly. 
- local envs don't do anything special to `settings.py` is the de-facto development settings file, as mentioned in the docstring at the top.
- tests pass `DJANGO_SETTINGS_MODULE=settings_tests` but `settings_tests.py` does include `settings` for convenience.

settings.py
```diff
@@ -132,17 +132,16 @@ def insert_debug_toolbar_middleware(middlewares):
...
-        'Could not import local_settings module. {}'.format(traceback.format_exc()),
-        stacklevel=1,
-    )
+local_settings_path = path('local_settings.py')
+
+if not os.path.exists(local_settings_path):
+    with open(local_settings_path, 'w') as file:
+        file.write(
+            '# If you have settings you want to overload,'
+            'put them in a local_settings.py.\n'
```
diox, line 141:
```suggestion
            '# Put settings you want to overload in this file.\n'
```

diox (approved):
This PR needs a better title to reflect what it does

diox (approved):
r+ with the fix to codestyle (probably caused by my suggestion, sorry about that)

PR #21897, opened 2024-02-20
Disable escaping when sending abuse emails

### LINKED ISSUE #21881: The name of the rating is between: &quot; instead of: " -STAGE
The name of the rating is between: &quot ; instead of: "" -STAGE
### What happened?
1. Job is landing in AMO stage Ratings, moderator chooses Ignore/Approve action Submit
2. An email is sent to the reporter who has the right to appeal 
3. Reporter appeals
4. Appeal is sent
5. Appeal is available in AMO stage Users queue
6. Adjust Decision - Ignore/Approve and Submit
7. Observe in the sent email that the name of the rating is between: &quot ; instead of: "

The name of the rating should be between: "
Test case used: https://testrail.stage.mozaws.net/index.php?/cases/view/2499416&group_by=cases:section_id&group_id=481321&group_order=asc
image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-717)

diox:
Fixes #21881

### DIFF
--- a/src/olympia/abuse/tests/test_utils.py
+++ b/src/olympia/abuse/tests/test_utils.py
@@ -86,6 +86,8 @@ class BaseTestCinderAction:
         assert f'[ref:ab89/{self.abuse_report_auth.id}]' in mail.outbox[1].body
         assert 'After reviewing' not in mail.outbox[0].body
         assert 'After reviewing' not in mail.outbox[0].body
+        assert '&quot;' not in mail.outbox[0].body
+        assert '&quot;' not in mail.outbox[1].body
 
     def _test_reporter_ignore_email(self, subject):
         assert mail.outbox[0].to == ['email@domain.com']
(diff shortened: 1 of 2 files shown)


src/olympia/abuse/tests/test_utils.py
```diff
@@ -222,7 +238,7 @@ class TestCinderActionUser(BaseTestCinderAction, TestCase):
 
     def setUp(self):
         super().setUp()
-        self.user = user_factory()
+        self.user = user_factory(display_name='<b>Bad Hørse</b>')
```
KevinMind, line 241: Should you add assertion that the <b> is not escaped?

src/olympia/abuse/utils.py
```diff
@@ -70,13 +76,18 @@ def notify_owners(self, *, policy_text=None):
         owners = self.get_owners()
         if not owners:
             return
-        name = self.get_target_name()
+        with no_jinja_autoescape():
```
KevinMind, line 79: Could this be solved by escaping the quotes in the linked issue? Why disable all escaping?

KevinMind (approved):
I don't have a strong opinion (especially if it's true we send all these emails as plain text only) but why not simply escape characters in the template that should be escaped?

I don't see any real issues though.

src/olympia/abuse/utils.py
diox: It wasn't just the quotes, and potentially could happen to other things than just the name / ratings.

PR #21906, opened 2024-02-21
Enable HTTP compression for Elasticsearch (but not in prod yet)

### LINKED ISSUE #21905: Experiment with HTTP compression for Elasticsearch
We're not currently using compression (in requests or responses) for Elasticsearch. Let's try it, in theory these days it should just require passing `http_compress=True` when instantiating the `Elasticsearch` object.

Need to try on dev/stage first, updating the [perf graphs]( https://earthangel-b40313e5.influxcloud.net/d/kiTC7XDZz/amo-prod-frontend-apis-usage-performance?orgId=1) to show data from dev/stage, as it might be more detrimental than beneficial to performance.

This should be 100% transparent to users - no visible changes, so if search/homepage works with no noticeable performance difference then it's successful.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-727)

diox:
Fixes #21905

### Context

It's a documented feature but disabled by default so we need to explicitly turn it on (https://elasticsearch-py.readthedocs.io/en/7.x/#compression) to try it.

### Testing

It's particularly annoying to test as logging on ES doesn't seem to expose request headers, so we can't easily confirm it's sending the headers except by replacing the ES server or mocking in tests.

Ultimately it's mostly an experiment, and the setting should be fully supported, if it doesn't make performance worse (we'll see that in the [API perf grafana dashboard](https://earthangel-b40313e5.influxcloud.net/d/kiTC7XDZz/amo-prod-frontend-apis-usage-performance?orgId=1)) then we'll enable in prod and see if it makes a noticeable difference in bandwidth.


src/olympia/conf/prod/settings.py
```diff
@@ -53,6 +53,7 @@
 ES_TIMEOUT = 60
 ES_INDEXES = {k: f'{v}_{ENV}' for k, v in ES_INDEXES.items()}
 ES_DEFAULT_NUM_SHARDS = 10
+ES_COMPRESS = False
```
KevinMind, line 56: TIL about this file. Is this our production settings?

src/olympia/lib/settings_base.py
```diff
@@ -1175,6 +1175,7 @@ def read_only_mode(env):
 ES_TIMEOUT = 30
 ES_DEFAULT_NUM_REPLICAS = 2
 ES_DEFAULT_NUM_SHARDS = 5
+ES_COMPRESS = True
```
KevinMind, line 1178: Per the description, it might be better to have the default be Off, and turn it on in specific environments? Is that not possible currently?

KevinMind:
Is there a way to test this locally? Can you add the to the description?

src/olympia/conf/prod/settings.py
diox: Yes

src/olympia/lib/settings_base.py
diox: We could but I expect to turn it on by default very soon or completely revert - so I figured might as well enable it by default, rather that having to enable it in local/dev/stage.

diox:
Well if indexing and searching works, then it's working... And we have a copious amount of automated tests around those things.

src/olympia/lib/settings_base.py
KevinMind: Fair enough.

KevinMind (approved):
> Well if indexing and searching works, then it's working... And we have a copious amount of automated tests around those things.

That proves we didn't break elastic search but not that the compression works. Per our discussion could you update the description with what you said about testing and link to the client docs so we can reference back in the future?

diox:
Description updated w/ more context

Issue #21837, opened 2024-02-09
Merge requirements.txt files


KevinMind:
We have 5 requirements files, but really we only need to distinguish between production and development dependencies. This distinction can be documented and the files merged.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-702)

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-702

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-702

PR #21907, opened 2024-02-21
Include needed requirements files for read the docs

KevinMind:
Fixes: #21932

### Description

In #21837 we merged our requirements.txt files but forgot to update the readthedocs config. Because this build doesn't run on PRs it was not caught before merging.

Use our needed req files in readthedocs build.

### Context

This is a bit of a hack, but rtd has strong opinions about how they want to install pip dependencies and only let us specify files to include. This breaks the build because our .txt files include loose versions, which they don't allow.

The solution, to keep it simple is to just install pip.txt which gives us the correct pip version, and then install via our make command. This gives us total control over the installation flow and ensures compatibility with the Makefile.

FWIW this patch will likely be made redundant by https://github.com/mozilla/addons-server/pull/21937 which entirely replaces rtd with github pages.

### Testing

You can enable rtd deployments on PRs and I did enable it to verify this fixes the build.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -9,6 +9,11 @@ build:
   os: ubuntu-20.04
   tools:
     python: "3.10"
+    nodejs: "18"
+  jobs:
+    post_install:
+      - make -f Makefile-docker update_deps
+
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
@@ -16,4 +21,4 @@ sphinx:
 
 python:
    install:
-   - requirements: requirements/docs.txt
+   - requirements: requirements/pip.txt


KevinMind:
Build seems to be fixed:
site: https://addons-server--21907.org.readthedocs.build/en/21907/
build: https://readthedocs.org/projects/addons-server/builds/23574392/

diox:
> This breaks the build because our .txt files include loose versions, which they don't allow.

We do ? Where ? That shouldn't be the case, we should only specify exact versions, with hashes, and ignore dependencies as we also specify those ourselves.

KevinMind:
Guess we don't need this?

diox:
If we are replacing rtd we don't need this yeah.

PR #21914, opened 2024-02-22
Efficiently cache docker build

KevinMind:
relates to : #21888

### Description

Using more efficient caching within our docker build to speed up successive builds.

### Context

This opens up faster build times both locally and in CI and makes the idea of building images more frequently a lot more palletable.

### Testing

This PR adds a make command to build a local docker image

```bash
make build_docker_image
```

This will build an image. If you want to really see the cache benefit, then first prune your local layer cache

```bash
docker system prune -a --volumes
```

If you have already run `make build_docker_image` you should prune the builder cache as well 

```bash
docker buildx prune --all
```

This is like runnig for the first time. It will take a while.

Then modify the package.json or one of the requirements files.

Run the build again.

You will see that pip is using "cached" versions of dependencies which have already been installed. That is the more aggressive layer caching kicking in.

You'll also see that if you change other files in the code base, it won't necessarily reinstall dependencies. You can compare that behavior to running the same commands on master.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -89,7 +89,24 @@ ENV PIP_SRC=/deps/src/
 ENV PYTHONUSERBASE=/deps
 ENV PATH $PYTHONUSERBASE/bin:$PATH
 ENV NPM_CONFIG_PREFIX=/deps/
-RUN ln -s ${HOME}/package.json /deps/package.json \
+ENV NPM_CACHE_DIR=/deps/cache/npm
+ENV NPM_DEBUG=true
+
+RUN \
+    # Files needed to run the make command
+    --mount=type=bind,source=Makefile,target=${HOME}/Makefile \
+    --mount=type=bind,source=Makefile-docker,target=${HOME}/Makefile-docker \
+    # Files required to install pip dependencies
+    --mount=type=bind,source=setup.py,target=${HOME}/setup.py \
+    --mount=type=bind,source=./requirements,target=${HOME}/requirements \
+    # Files required to install npm dependencies
+    --mount=type=bind,source=package.json,target=${HOME}/package.json \
+    --mount=type=bind,source=package-lock.json,target=${HOME}/package-lock.json \
+    # Mounts for caching dependencies
+    --mount=type=cache,target=${PIP_CACHE_DIR},uid=${OLYMPIA_UID},gid=${OLYMPIA_GID} \
+    --mount=type=cache,target=${NPM_CACHE_DIR},uid=${OLYMPIA_UID},gid=${OLYMPIA_GID} \
+    # Command to install dependencies
+    ln -s ${HOME}/package.json /deps/package.json \
     && ln -s ${HOME}/package-lock.json /deps/package-lock.json \
     && make update_deps_prod
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10-slim-buster
+FROM python:3.10-slim-buster as base
 
 # Should change it to use ARG instead of ENV for OLYMPIA_UID/OLYMPIA_GID
 # once the jenkins server is upgraded to support docker >= v1.9.0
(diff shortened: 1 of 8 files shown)


KevinMind:
Note: currently the cache is not working very effectively for apt packages. I'm looking into that. Second npm seems to be getting cache misses once running the container. That might be due to how the volume mounting works when running the container. I'm not sure that /deps/npm/cache is being added to the final container.

I will likely make further improvements to this once we enable multi stage builds but for now this is still a LOT faster than before.

Makefile-docker
```diff
@@ -84,29 +92,32 @@ update_deps_base: ## update the python and node dependencies
...
-
 .PHONY: update_deps
 update_deps: update_deps_base ## update the python and node dependencies for development
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/dev.txt
+	npm install $(NPM_ARGS)
 
 .PHONY: update_deps_prod
 update_deps_prod: update_deps_base ## update the python and node dependencies for production
-	npm prune --omit=dev
+	npm install $(NPM_ARGS) --omit=dev --no-save --no-package-lock
```
diox, line 102: `--no-package-lock` seems extremely dangerous ? How else are we ensuring the hashes are checked ? Also, I assume node of the JS libs we care about in production have dependencies themselves but if we ever start to introduce some, that `--no-package-lock` would make our builds not reproducible...

Dockerfile
```diff
@@ -88,18 +93,35 @@ ENV PIP_SRC=/deps/src/
...
+    --mount=type=bind,source=package-lock.json,target=${HOME}/package-lock.json \
+    # Mounts for caching pip/npm dependencies
+    --mount=type=cache,target=${PIP_CACHE_DIR},uid=${OLYMPIA_UID},gid=${OLYMPIA_GID} \
+    --mount=type=cache,target=${NPM_CACHE_DIR},uid=${OLYMPIA_UID},gid=${OLYMPIA_GID} \
+    # Command to install pip/npm dependencies
+    ln -s ${HOME}/package.json /deps/package.json \
     && ln -s ${HOME}/package-lock.json /deps/package-lock.json \
     && make update_deps_prod
 
+# Only copy our source files after we have installed all dependencies
```
diox, line 117: What difference does that make ?

Makefile-docker
KevinMind: You are right, I refactored this PR already before this comment to make the changes more isolated and discrete. I've added a commit that replaces install for `npm ci` which achieves what I was trying to but in a better way.

I wanted to ensure we rely only on the package-lock.json and that we don't save any files as both reduce reproducibility but you're right I misunderstood what --no-package-lock does and it does the opposite.

This should be much faster/better/secure.

Dockerfile
KevinMind: The difference here is subtle and not totally visible yet but the further up in the build chain we copy "." the more likely it is that we invalidate steps which don't actually depend on 100% of files.

By moving this instruction to the final and bottom most step we can allow docker to more efficiently cache individual layers which means we are much more likely to get a cache hit.

For example, in the updated version of this pr you can see that the locale compilation step is in a separate stage with only the files it needs copied, this means the layer will only rebuild if you change files in that specific folder.

We will get the absolute most out of this change once we are able to extract the `update_assets` command from the final stage as this takes about 30-45 seconds and because it runs after copying will re-run no matter what files change. But the work to do that is way too big for this PR so we can do that later.

locale/requirements.txt
```diff
@@ -0,0 +1,3 @@
+dennis==1.1.0 \
```
diox, line 1: Why is this in this PR ? I don't see it being required anywhere. Also, requirements files are expected to live under `requirements/` top level directory to be picked up by our (current) [dependabot config](https://github.com/mozilla/addons-server/blob/master/.github/dependabot.yml).

KevinMind:
57 seconds 💪 https://github.com/mozilla/addons-server/actions/runs/8055458292/job/22002361112?pr=21914

locale/requirements.txt
KevinMind: You are so very right. Somehow at some point I was having the compile-mo.sh script failing and so I thought I had to install dennis, but that makes no sense, it's already in prod.txt so I removed this.

Build: https://github.com/mozilla/addons-server/actions/runs/8055526462/job/22002582564?pr=21914

KevinMind:
Comparison of master to this branch with different scenarios, changing different types of files that would invalidate different stages of the build.

Running `make build_docker_image` with that commit cherry-picked onto master.

| Operation                             | Before  | After  | % Difference | Absolute Difference |
|---------------------------------------|---------|--------|--------------|---------------------|
| Cold (empty docker cache)             | 398.9s  | 386.8s | -3.0%        | 12.1s               |
| Warm (filled cache no changes)        | 6.8s    | 6.2s   | -8.8%        | 0.6s                |
| Modify prod.txt (add a single dep.)   | 179.1s  | 88.7s  | -50.4%       | 90.4s               |
| Modify src/** (add a comment to file) | 170.2s  | 34.5s  | -79.7%       | 135.7s              |

diox:
So trying to use this... After running `make build_docker_image`, you get:
```
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
```

What am I supposed to run ? I tried `docker build --cache-from addons-server-test:latest -t test_amo .` but got: 
```
=> ERROR importing cache manifest from addons-server-test:latest
````

And it looks like it's rebuilding from the beginning - with layers caching, but it's not as fast as I'd expect given there are no changes (it's rebuilding dependencies, re-running locale compilation, redoing the copy and assets, taking 60 sec total)

**Comment by KevinMind** [comment]
You can ignore the warning as the make command is using the cache type=locale and saving to your local file system instead of pushing to a registry.

You can look at what the make command executes to see how you'd repeat with raw docker commands. TLDR; you have to make sure to use buildkit and you have to make sure to use the docker-container driver to enable the full scope of caching.

docker build as you have it will likely miss out on several opportunities because of one or the other or both things. part of the reason why I added the make command is to normalize that flow. I don't want to have to remember what exact magic to pass, just make it work.

Though it is probably a good thing to add --load so that the built image is added to the docker daemon, since we are using a custom builder I'm not 100% sure if it is automatically available or we have to pass load. I can check that and update if it doesn't pass it up the chain.

**Comment by KevinMind** [comment]
Also, to clarify the error you got, when you run`--cache-from addons-server-test:latest` you are asking docker to pull a cache manifest from an image called addons-server-test:latest` if that image doesn't exist on your machine or in a hub docker is setup to look in, then it will fail. In the make file we specify --cache-from... that points to the ./docker-cache directory in the local file system which you should see after running.

**Comment by diox** [comment]
Cool, adding `--load` to the build command is what I was looking for. Now I can manipulate my built image normally, I see it in `docker images` etc.

**Comment by KevinMind** [comment]
There needs to be some docks written and mostly pointing to other places for this I think :)

**Review by diox — APPROVED** [review]
Please update the PR title to something more descriptive before merging.

**Comment by KevinMind** [comment]
I'd like to merge this first https://github.com/mozilla/addons-server/pull/21930 The actions will be super useful in testing this. I would like to do a bit more testing before merging.

File: Makefile-docker [hunk 4]
```diff
@@ -84,25 +91,26 @@ update_deps_base: ## update the python and node dependencies
...
 update_db: ## run the database migrations
 	$(PYTHON_COMMAND) manage.py migrate --noinput
 
 .PHONY: update_assets
 update_assets:
+	# Copy files required in compress_assets to the static folder
+	mkdir -p $(STATIC_CSS) $(STATIC_JS) $(STATIC_JQUERY_UI)
+	for dest in $(NODE_LIBS_CSS) ; do cp $(NODE_MODULES)$$dest $(STATIC_CSS) ; done
+	for dest in $(NODE_LIBS_JS) ; do cp $(NODE_MODULES)$$dest $(STATIC_JS) ; done
+	for dest in $(NODE_LIBS_JQUERY_UI) ; do cp $(NODE_MODULES)$$dest $(STATIC_JQUERY_UI) ; done
```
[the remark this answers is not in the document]
KevinMind: Done.

PR #21930, opened 2024-02-26
github action to efficiently build docker image on PRs to verify build

KevinMind:
Relates to: #21888 

### Description

This PR introduces a github composite action that can build and (optionally) push a docker image to a registry. There are also two workflows:

- `verify-docker-image` that runs on PRs and builds (without pushing).
- `build-docker-image` the runs on workflow dispatch and can arbitrarily build and or push an image to the registry.

This ensures PRs do not merge code which results in an unbuildable docker image (as has happened recently) Also allows sharing built images.

### Context

Why build a composite action instead of only a workflow? The docker build involves several steps and can be reused across any number of workflows, from actually pushing production images, testing PRs with a real container, to deploying on temporary environments or sharing containers over a registry for testing. 

I already have several such use-cases prototyped but did not want to make this PR to big, but at the same time it is fine to implement the composite action that is already clearly useful.

Prior art: Examples of repos pushing images in CI, majority pushing, majority using dockerhub.

- https://github.com/mozilla/neqo/blob/main/.github/workflows/qns.yml
- https://github.com/mozilla/neqo/blob/7f6fc8bdf60d026c87980c9fe6974166fc2455c5/.github/workflows/qns.yml#L51
- https://github.com/mozilla/hubs-ops/blob/53b43ea7326499d324707a08159d77323f9b8dfa/.github/workflows/turkeyGitops.yml#L85
- https://github.com/mozilla/jira-bugzilla-integration/blob/c6cb0b70aad5f6207bea6f7f729e251379d29383/.github/workflows/build-publish.yaml#L53
- https://github.com/mozilla/classify-client/blob/a7bd632a63e5bcf4c6eb090a2618f1f57aa52d85/.github/workflows/publish.yml#L56
- https://github.com/mozilla/remote-settings/blob/6169a51ec1b9ec04ead5198cdcfb9dc7b9cafb53/.github/workflows/publish.yaml#L44
- https://github.com/mozilla/remote-settings/blob/6169a51ec1b9ec04ead5198cdcfb9dc7b9cafb53/.github/workflows/ci.yaml#L124
- https://github.com/mozilla/mentoring/blob/345a8bd2a66e4456c2c0338ccbc5e4e483fe8e5f/.github/workflows/docker.yaml#L33
- https://github.com/mozilla/ssh_scan/blob/f3256f29e1a8e596f37824ae063c1f05e00fc32f/.github/workflows/docker-deploy.yml#L23
- https://github.com/mozilla/ssh_scan/blob/f3256f29e1a8e596f37824ae063c1f05e00fc32f/.github/workflows/docker-test.yml#L20

### Testing

You can test the workflow on this PR or by branching and adding commits. Keep in mind, there will not be very efficient caching on the build until https://github.com/mozilla/addons-server/pull/21914 is merged, but actually, this PR is a great way to test that one so we should merge it first and make sure we don't break the build.

### DIFF
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -60,7 +60,7 @@ runs:
         buildkitd-flags: --debug
 
     # Login to a registry to push the image
-    - name: Login to GitHub Container Registry
+    - name: Login to Container Registry
       # Only login if we are pushing the image
       if: ${{ inputs.push == 'true' }}
       uses: docker/login-action@v3
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -1,0 +1,106 @@
+name: 'Docker Build'
+description: 'Builds `addons-server` docker image'
+inputs:
+  load:
+    required: false
+    description: "Build and load image to local docker daemon. (cannot be used together with push)"
+    default: "false"
+  no_cache:
+    required: false
+    description: "Build fresh image without cache"
+    default: "false"
+  password:
+    required: false
+    description: "Docker registry password"
+    default: "invalid"
+  pull:
+    required: false
+    description: "Pull all referenced images before building"
+    default: "false"
+  push:
+    required: false
+    description: "Build and push image to registry (cannot be used together with load)"
+    default: "false"
+  python_version:
+    required: false
+    description: "Python version"
+    default: "3.10"
+  username:
+    required: false
+    description: "Docker registry username"
+    default: "invalid"
+
+outputs:
+  tags:
+    description: "The Docker tags for the image"
+    value: ${{ steps.meta.outputs.tags }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Validate inputs
+      shell: bash
+      run: |
+        if [[ ${{ inputs.load == 'true' }} == ${{ inputs.push == 'true' }} ]]; then
+          echo "Cannot use load and push together. Must choose only one of them."
+          exit 1
+        fi
+
+        if [[ "${{ inputs.push  }}" == "true" && "${{ github.ref }}" == "refs/heads/master" ]]; then
+          echo "Cannot push to registry from master branch unless we migrate our master build job to GHA."
+          exit 1
+        fi
+    # Setup docker to build for multiple architectures
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+      with:
+        version: latest
+        buildkitd-flags: --debug
+
+    # Login to a registry to push the image
+    - name: Login to GitHub Container Registry
+      # Only login if we are pushing the image
+      if: ${{ inputs.push == 'true' }}
+      uses: docker/login-action@v3
+      with:
+        username: ${{ inputs.username }}
+        password: ${{ inputs.password }}
+
+    - name: Set commit sha
+      id: sha
+      shell: bash
+      run: echo "sha_short=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+
+    # Determine the tags for the image
+    - name: Docker meta
+      id: meta
+      uses: docker/metadata-action@v5
+      with:
+        # Hard coding our dockerhub imnage name
+        images: mozilla/addons-server
+        tags: |
+          type=schedule
+          type=ref,event=tag
+          type=ref,event=branch,suffix=-${{ steps.sha.outputs.sha_short }}
+          type=ref,event=pr,suffix=-${{ steps.sha.outputs.sha_short }}
+          # set latest tag for default branch
+          # Disabled for now as we do not use this action for
+          # The production build
+          # type=raw,value=latest,enable={{is_default_branch}}
+
+    - name: Build Image
+      uses: docker/build-push-action@v5
+      with:
+        context: .
+        platforms: linux/amd64
+        pull: ${{ inputs.pull }}
+        push: ${{ inputs.push }}
+        load: ${{ inputs.load }}
+        no-cache: ${{ inputs.no_cache }}
+        tags: ${{ steps.meta.outputs.tags }}
+        cache-from: type=gha
+        cache-to: type=gha,mode=max
+        build-args: |
+            PYTHON_VERSION=${{ inputs.python_version }}
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -17,7 +17,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  bulid_docker_image:
+  build_docker_image:
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -1,0 +1,35 @@
+name: Build Docker image
+
+on:
+  workflow_dispatch:
+    inputs:
+      push:
+        description: 'Push the image to registry?'
+        default: "false"
+        required: false
+      ref:
+        description: 'Branch, PR, tag or commit to build'
+        default: "master"
+        required: false
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.inputs.ref }}-${{ github.event.inputs.push }}
+  cancel-in-progress: true
+
+jobs:
+  bulid_docker_image:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.ref }}
+          fetch-depth: 1
+
+      - name: Build container
+        id: build_container
+        uses: ./.github/actions/build-docker
+        with:
+          push: ${{ github.event.inputs.push }}
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PASS }}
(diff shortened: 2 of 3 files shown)


KevinMind:
Verification that smoke test is running on the locally build image. https://github.com/mozilla/addons-server/actions/runs/8055095166/job/22001145649?pr=21930 "Hello world" print was introduced in this pr and hasn't been push to ghcr.

Makefile-docker
eviljeff, line 1: What's this change for?  We need the deps installed before our source?  (And we remove the `/deps/build/` directory in between)

.github/actions/build-docker/action.yml
```diff
@@ -0,0 +1,82 @@
...
+    required: false
+    description: "Build fresh image without cache"
+    default: "false"
+  push:
+    required: false
+    description: "Build and push image to registry (cannot be used together with load)"
+    default: "false"
+  pull:
+    required: false
+    description: "Build and push image to registry"
```
eviljeff, line 14: This description is correct?

.github/actions/build-docker/action.yml
eviljeff, line 1: so am I right in gathering that of the 6 inputs we're only using `load`, and `push`

.github/actions/build-docker/action.yml
```diff
@@ -0,0 +1,82 @@
...
+      uses: docker/setup-buildx-action@v1
+      with:
+        version: latest
+        buildkitd-flags: --debug
+
+    # Login to a registry to push the image
+    - name: Login to GitHub Container Registry
+      uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
```
eviljeff, line 50: We're not using `ghcr.io` at the moment are we?  I thought we used dockerhub?  Is the intent to move away from dockerhub or duplicate builds? (why, in either case)

eviljeff (commented):
I may be missing some context, but I'm not clear what the aim of some of this PR is.  Verifying builds, yep; pushing and pulling to ghcr.io, not so sure.

.github/actions/build-docker/action.yml
KevinMind: Good catch, copy pasta.

.github/actions/build-docker/action.yml
KevinMind: I've set it up with ghcr.io for now for 2 reasons:
1. don't pollute our main registry we actually use for production
2. it is easy to setup because auth is implicit.

We can switch this action to use dockerhub if we want,it just means we would need to add credentials as secrets in the repo. Didn't seem important yet to do so.

Makefile-docker
KevinMind: The way it was actually breaks non docker environments. trying to install with --no-use-pep517 requires us to have setuptools and wheel installed which can only happen if we run the requirements files first. I imagine we never ran into this before because our base docker image might already have them installed? 🤷 not totally sure.

Why is it here? Originally this PR also introduced an action for running a build "locally" on the github action runner host itself, because it will be useful with l10n automation amongst other things, but I wanted to keep the PR smaller.

I'm happy to remove this commit and solve it in another PR or later, but it just seemed like a quick win. Don't have a strong opinion either way. LMK which you prefer.

KevinMind:
> I may be missing some context, but I'm not clear what the aim of some of this PR is. Verifying builds, yep; pushing and pulling to ghcr.io, not so sure.

Updated the description.

Makefile-docker
eviljeff: lets drop it for now then - one less file change

.github/actions/build-docker/action.yml
eviljeff: if there's a plan to use them and they're a few lines extra in the new file may as well keep them I guess

.github/actions/build-docker/action.yml
eviljeff: My instinct is that we should stick with a single integration - either ghcr.io or dockerhub - but I don't have an opinion right now which. @diox do you?

KevinMind: I think long term yes, just don't want to block this PR on decisions that are not urgent to make and we should carefully consider.

Makefile-docker
KevinMind: I'll just open a new PR with it :)

.github/actions/build-docker/action.yml
diox: Do we have to push the image at all ?

KevinMind: If you want to be able to run and test it outside of CI then yes. I did this several times when working on implementing this PR and can see myself doing it very regularly. You build an Image and want to run that image locally to get "exactly" what the CI built in your environment.

It also has huge implications for this [decision](https://mozilla-hub.atlassian.net/wiki/spaces/FDPDT/pages/564627681/Everything+on+Master+must+be+Deployable)

KevinMind: Though the default behaviour on the `verify-docker-image` is to [load](https://github.com/mozilla/addons-server/pull/21930/files#diff-c54769fa60b5344b018413c90790e136a0ab1d23ef81b6dfaf68135c220317a2R19) instead of push. You can build it and test it, but it won't push.

If you want to push an image you can run the workflow to push the image. I imagine we may find different use cases for pushing but for now this is flexible.

eviljeff: Personally, I don't care that much if we're building to different registry in this PR, but I do think there should be a follow-up issue to consolidate to one or the other.

KevinMind: Done. https://github.com/mozilla/addons-server/issues/21940

diox: Both registries have potential costs involved so we shouldn't push to both if not actually required IMHO.

KevinMind: I think we are answering the wrong question right now. Do we want to block this PR on that decision? Or is it fine to merge and make the registry decision in a separate iteration. In order to push anything to ghcr you'd have to manually initiate a job, so I think the risk is low, but if you'd rather keep it all on dockerhub for now then we need to set the credentials in the repo and I can update the PR to login there.

I would recommend we keep it as is and tackle the registry question in the linked ticket.

.github/workflows/verify-docker-image.yml
```diff
@@ -0,0 +1,32 @@
+name: Verify Docker Image
+
+on:
+  pull_request:
```
KevinMind, line 4: @diox to clarify a concern you had about over-pushing. This job runs on pull requests, it loads the image so it can be "tested" in CI. right now a minimal smoke test. It does not and cannot push because you must either load or push, cannot do both with the underlying action.

.github/workflows/build-docker-image.yml
```diff
@@ -0,0 +1,33 @@
+name: Build Docker image
+
+on:
```
KevinMind, line 3: @diox further clarification. This job only runs on workflow_dispatch, meaning you have to manually trigger it. It also only pushes if you explicitly opt in and the default value is false.

.github/actions/build-docker/action.yml
```diff
@@ -0,0 +1,82 @@
...
+        username: ${{ github.actor }}
+        password: ${{ github.token }}
+
+    # Determine the tags for the image
+    - name: Docker meta
+      id: meta
+      uses: docker/metadata-action@v5
+      with:
+        images: ghcr.io/${{ github.repository }}
+        tags: |
```
KevinMind, line 60: We can remove the last type to make it impossible to tag latest from the action. That would improve security of this action a lot.

.github/actions/build-docker/action.yml
KevinMind: dockerhub credentials were added so we can proceed with 1 registry.

.github/actions/build-docker/action.yml
KevinMind: Done.

KevinMind:
Push job working:
build: https://github.com/mozilla/addons-server/actions/runs/8070011683/job/22046351996#step:3:1409
image: https://hub.docker.com/layers/mozilla/addons-server/ADDSRV-720-verify-docker-image-b3599707aa9f4ab65b6c3866e0b324b3515977cf/images/sha256-7e8449530b0556f3f521712e4ba0075cdc2c28ead0297830d1d523af2b4ec448?context=explore

Pulled the image and ran locally, working as expected.

.github/actions/build-docker/action.yml
```diff
@@ -0,0 +1,106 @@
...
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+      with:
+        version: latest
+        buildkitd-flags: --debug
+
+    # Login to a registry to push the image
+    - name: Login to GitHub Container Registry
```
diox, line 63:
```suggestion
    - name: Login to Container Registry
```

.github/workflows/build-docker-image.yml
```diff
@@ -0,0 +1,35 @@
...
+        description: 'Branch, PR, tag or commit to build'
+        default: "master"
+        required: false
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.inputs.ref }}-${{ github.event.inputs.push }}
+  cancel-in-progress: true
+
+jobs:
+  bulid_docker_image:
```
diox, line 20:
```suggestion
  build_docker_image:
```

diox (approved):

.github/workflows/build-docker-image.yml
KevinMind: I had to look reallly closely to catch that even in your diff.

PR #21936, opened 2024-02-27
drop escaping of certain characters when sent to cinder api

### LINKED ISSUE #21926: Revert special characters handling for cinder API integration
https://github.com/mozilla/addons-server/pull/21866 added escaping for certain characters that cinder errored on, if they were the first character in the string.  https://mozilla-hub.atlassian.net/browse/TSPLAT-152 indicates this has now been fixed in cinder so we should be okay to revert that change (probably not a straight git revert)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-731)

eviljeff:
fixes #21926 - I kept the trimming because it seemed useful still.

### DIFF
--- a/src/olympia/abuse/tests/test_cinder.py
+++ b/src/olympia/abuse/tests/test_cinder.py
@@ -160,11 +160,6 @@ class BaseTestCinderCase:
         assert instance.get_str(123) == '123'
         assert instance.get_str(None) == ''
         assert instance.get_str(' ') == ''
-        assert instance.get_str('----') == r'\----'
-        assert instance.get_str('@@@') == r'\@@@'
-        assert instance.get_str('==') == r'\=='
-        assert instance.get_str('_') == r'\_'
-        assert instance.get_str(' _ ') == r'\_'
 
 
 class TestCinderAddon(BaseTestCinderCase, TestCase):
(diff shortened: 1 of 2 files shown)


src/olympia/abuse/tests/test_cinder.py
```diff
@@ -198,7 +193,7 @@ def test_build_report_payload(self):
             privacy_policy='Söme privacy policy',
             version_kw={'release_notes': 'Søme release notes'},
         )
-        message = '- bad addon!'
```
KevinMind, line 201: What's the point of removing the characters in the test inputs? If they are allowed, shouldn't we keep them as examples of "working" features?

KevinMind (approved):
This could be tested against the cinder staging API before merging right? Do you think it would be worth doing? Otherwise we could merge this, find out it isn't fixed and just have to revert?

Do you know how you would go about testing this? There are no test scenarios listed in the ticket or the PR.

eviljeff:
> This could be tested against the cinder staging API before merging right? Do you think it would be worth doing? 
> 
> Do you know how you would go about testing this? There are no test scenarios listed in the ticket or the PR.

it's possible to manually construct the queries and make the requests to their API, with appropriate auth credentials, but it's a pain so I didn't expect anyone to manually test it.

> Otherwise we could merge this, find out it isn't fixed and just have to revert?

That's a risk, but so far our history with their developers is their statements about something being addressed are accurate so I'm assuming it is actually fixed.  If QA test this and it turns out to not actually be fixed, and we have to revert, then I'll certainly reverse that assumption next time.

PR #21938, opened 2024-02-27
Don't include empty additional reasoning in cinder appeal decline emails if there isn't any

### LINKED ISSUE #21873: Author's appeal is denied - a phrase has . . at the end
Followup for https://testrail.stage.mozaws.net/index.php?/cases/view/2498871

Step 5: ". ." left in email's content

2dots in the email

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-713)

diox:
Fixes #21873


eviljeff (approved):
Afaik, it shouldn't happen in reality, because the T&S moderators should always be giving an extended reasoning for the denial, but doesn't hurt.

PR #21944, opened 2024-02-28
Gate enqueueing reports in reviewer tools behind enable-cinder-reviewer-tools-integration switch

diox:
Fixes mozilla/addons#9488

### DIFF
--- a/src/olympia/abuse/tests/test_utils.py
+++ b/src/olympia/abuse/tests/test_utils.py
@@ -325,6 +325,7 @@ class TestCinderActionUser(BaseTestCinderAction, TestCase):
         self._test_owner_affirmation_email(f'Mozilla Add-ons: {self.user.name}')
 
 
+@override_switch('enable-cinder-reviewer-tools-integration', active=True)
 class TestCinderActionAddon(BaseTestCinderAction, TestCase):
     ActionClass = CinderActionDisableAddon
 
@@ -334,7 +335,6 @@ class TestCinderActionAddon(BaseTestCinderAction, TestCase):
         self.addon = addon_factory(users=(self.author,), name='<b>Bad Addön</b>')
         ActivityLog.objects.all().delete()
         self.cinder_job.abusereport_set.update(guid=self.addon.guid)
-        self.create_switch('enable-cinder-reviewer-tools-integration', active=True)
 
     def _test_disable_addon(self):
         self.cinder_job.update(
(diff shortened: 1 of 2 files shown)


eviljeff (commented):
imo, to handle escalations we need a change in `CinderActionEscalateAddon` too, to prevent the flagging there.  Otherwise the those add-ons will be flagged for human review but `enable-cinder-reviewer-tools-integration` will prevent the `CinderJob` from showing up to be resolved.

Or we could use a different waffle switch to allow `enable-cinder-reviewer-tools-integration` to be enabled again to process escalations, while preventing non-escalations from enqueuing

diox:
https://github.com/mozilla/addons-server/pull/21944/commits/954b308c6ecf7e768f5ab4402bf04a1b4b5eb333 will prevent escalations as well

src/olympia/abuse/tests/test_utils.py
```diff
@@ -332,6 +334,7 @@ def setUp(self):
         self.addon = addon_factory(users=(self.author,), name='<b>Bad Addön</b>')
         ActivityLog.objects.all().delete()
         self.cinder_job.abusereport_set.update(guid=self.addon.guid)
+        self.create_switch('enable-cinder-reviewer-tools-integration', active=True)
```
eviljeff, line 337: could decorate the class with  `@override_switch('enable-cinder-reviewer-tools-integration', active=True)` instead

src/olympia/abuse/tests/test_utils.py
```diff
@@ -457,6 +460,41 @@ def test_escalate_addon(self):
...
+        # Escalation when the waffle switch is off is essentially a no-op on
+        # AMO side.
+        listed_version = self.addon.current_version
+        listed_version.file.update(is_signed=True)
+        unlisted_version = version_factory(
+            addon=self.addon, channel=amo.CHANNEL_UNLISTED, file_kw={'is_signed': True}
+        )
+        ActivityLog.objects.all().delete()
+        action = CinderActionEscalateAddon(self.cinder_job)
+        with override_switch('enable-cinder-reviewer-tools-integration', active=False):
```
eviljeff, line 473: if we're making all the calls with the override, decorate the method instead?

src/olympia/abuse/utils.py
```diff
@@ -245,8 +250,19 @@ class CinderActionEscalateAddon(CinderAction):
     def process_action(self):
         """This will return always return a falsey value because we've not taken any
         action at this point, just flagging for human review."""
+        self.flag_for_human_review()
+
+    def flag_for_human_review(self):
```
eviljeff, line 255: ... not sure what the optimization is here.

eviljeff (approved):
just personal preference nits; r+

src/olympia/abuse/utils.py
diox: Not really an optimization, just thought it'd be clearer that way

PR #21953, opened 2024-02-29
monitor cinder's /health endpoint in our services monitor

### LINKED ISSUE #21931: Update monitor.json to check Cinder
split from https://github.com/mozilla/addons-server/issues/21853

We should add a check for cinder api to monitor.py to ensure they are up and we can make authenticated requests to their servers.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-734)

eviljeff:
fixes #21931 - the /health endpoint is available without auth

If you want to test this locally you can load `http://olympia.test/services/monitor.json` (some of the other services might fail depending on your local settings)

### DIFF
--- a/src/olympia/amo/monitors.py
+++ b/src/olympia/amo/monitors.py
@@ -2,6 +2,7 @@ import io
 import os
 import socket
 import traceback
+from urllib.parse import urljoin
 
 from django.conf import settings
 
@@ -224,3 +225,31 @@ def remotesettings():
         status = f'Failed to execute task in time: {e}'
         monitor_log.critical(status)
     return status, None
+
+
+def cinder():
+    # check cinder connectivity
+    signer_results = None
+    status = ''
+
+    cinder_api_url = settings.CINDER_SERVER_URL
+    if cinder_api_url:
+        try:
+            health_url = urljoin(cinder_api_url, '/health', allow_fragments=False)
+            response = requests.get(health_url, timeout=5)
+            if response.status_code != 200:
+                status = 'Failed to chat with cinder. Invalid HTTP response code.'
+                monitor_log.critical(status)
+                signer_results = False
+            else:
+                signer_results = True
+        except Exception as exc:
+            status = 'Failed to chat with cinder: %s' % exc
+            monitor_log.critical(status)
+            signer_results = False
+    else:
+        status = 'CINDER_SERVER_URL is not set'
+        monitor_log.critical(status)
+        signer_results = False
+
+    return status, signer_results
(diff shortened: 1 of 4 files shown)


src/olympia/amo/tests/test_monitor.py
```diff
@@ -126,3 +126,27 @@ def test_remotesettings_fail(self):
         )
         obtained, _ = monitors.remotesettings()
         assert '503 Server Error: Service Unavailable' in obtained
+
+    def test_cinder_success(self):
+        responses.add(
+            responses.GET,
+            'https://stage.cinder.nonprod.webservices.mozgcp.net/health',
```
KevinMind, line 133: Any reason not to use the settings.CINDER_WHATEVER_URL here?

src/olympia/amo/monitors.py
```diff
@@ -224,3 +225,31 @@ def remotesettings():
...
+    # check cinder connectivity
+    signer_results = None
+    status = ''
+
+    cinder_api_url = settings.CINDER_SERVER_URL
+    if cinder_api_url:
+        try:
+            health_url = urljoin(cinder_api_url, '/health', allow_fragments=False)
+            response = requests.get(health_url, timeout=5)
+            if response.status_code != 200:
```
KevinMind, line 240: Why do we care about 200 literally? Why not `raise_for_status()` and remove the if else? Just curious about the rationale.

src/olympia/amo/tests/test_monitor.py
eviljeff: we have to manipulate it to strip off the `/api/...` so I wanted to make sure the correct url is actually pinged (or we'd be testing the wrong url)

src/olympia/amo/monitors.py
eviljeff: I was adopting a similar style to the other monitor functions but we can use `raise_for_status`

src/olympia/amo/tests/test_monitor.py
KevinMind: Isn't the actual monitor using [settings.CINDER_SERVER_URL](https://github.com/mozilla/addons-server/pull/21953/files#diff-d1476e7a6ba82b3532601f6d7014f0a01fbbc7fef12f82cdd74ed0702b031067R238)?

KevinMind: I see, you have the `/` on the path. That is very subtle. I almost would rather see something like

```
f'{settings.CINDER_SERVER_URL.replace('/api/v1', '')}/health'
```

in the test.

That makes it very clear we are removing the api part. Or split the variable to a base url and an api url. But I get it, that second option is probably too big for this PR.

Wdyt though? I find it very not obvious how the actual URL is built in the monitor and also don't really like copy pasting URLs all over the codebase.

I'm happy to approve and let you decide. If you're happy enough, then let's go, but if either of the ideas I've shared or you have another idea how to address that, also great.

KevinMind (approved):
Tested on local branch. All good 🚀

Issue #21959, opened 2024-03-01
Fix Docker build


KevinMind:
### Describe the problem and steps to reproduce it:

#21914 broke docker build in circle CI. Parsing errors with dockerfile using mount cache.

Build: https://app.circleci.com/pipelines/github/mozilla/addons-server/29495/workflows/6b18f12a-8a0e-4967-a852-162350a78234/jobs/194514

### What happened?

Build failed

### What did you expect to happen?

Build passed

### Anything else we should know?
(Please include a link to the page, screenshots and any relevant files.)

Need to verify the docker version and see if enabling buildkit fixes the build.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-744)

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-744

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-744

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-744

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-744

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-744

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-744

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-744

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-744

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-744

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-744

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-744

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-744

KevinMind:
[...]

PR #21960, opened 2024-03-01
Build docker image with buildx to support new Dockerfile syntax

KevinMind:
Fixes: #21959

### Description

Explicitly use `docker buildx build` to ensure buildkit flags and syntax are available during local/ci builds. We run the make command in CI to ensure the same command is executed in both environments. This is also _essentially_ compatible with that is executed in the github action build.

### Context

#21914 broke the docker build in circleci, this went unnoticed because the build only runs on master and the command was simple and using the same docker build as we have in the locale make command `build_docker_image`. There were differences in arguments, but it was assumed that docker build would support buildx context... that was wrong.

### Testing

You can test by adding the `-build-image` job to a workflow. It is enabled in this commit: [d8a8420731e2b4ca7d83fbb98befc8f8817e5478](https://github.com/mozilla/addons-server/pull/21960/commits/d8a8420731e2b4ca7d83fbb98befc8f8817e5478)

Which will be removed before merging.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -350,6 +350,9 @@ commands:
   make_release:
     description: "Builds and pushes a Docker image"
     parameters:
+      push:
+        type: boolean
+        default: false
       image_tag:
         type: string
         default: "latest"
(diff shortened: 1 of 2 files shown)


KevinMind:
Successful build/push: https://hub.docker.com/layers/mozilla/addons-server/test/images/sha256-7a44b6a10568ca48fcd06a6640dcf7696de956565e3cc7172ad11592c76d4b78?context=explore

KevinMind:
Build: https://app.circleci.com/pipelines/github/mozilla/addons-server/29510/workflows/af972e53-4962-4161-9213-f97c421aa3de/jobs/194667

.circleci/config.yml
```diff
@@ -635,6 +658,8 @@ workflows:
       - devhub
       - docs
       - main
+      # Uncomment if you want to test the docker build
+      - build-image
```
eviljeff, line 662: it's already uncommented though. - is this comment accurate?

KevinMind: image

KevinMind: Wanted to confirm it actually works. I'll remove the commit reverting back to commented out before merging.

eviljeff (approved):

Issue #21943, opened 2024-02-28
Fix `check_suppressed_email_confirmation` 


KevinMind:
### Describe the problem and steps to reproduce it:
(Please include as many details as possible.)

API calls to socketlabs are sometimes returning invalid statuses which prevents the confirmation from being "acceptable"

If a user requests email verification code, and receives the code, if anything goes wrong with the check to socketlabs then the code cannot be used to verify the email.

### What happened?

User sees timeout screen. Even if they click the link with the code, they are not able to verify the email because we have not confirmed from socketlabs.

### What did you expect to happen?

User should be able to verify regardless of the status check from socketlabs.

### Anything else we should know?
(Please include a link to the page, screenshots and any relevant files.)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-737)

```[tasklist]
### Tasks
- [ ] Fix: allow valid confirmation code to verify email, regardless of socketlabs
- [ ] Remove status from email verification (not needed) ??
- [ ] https://github.com/mozilla/addons-server/issues/21965
```

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-737

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-737

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-737

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-737

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-737

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-737

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-737

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-737

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-737

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-737

[...]

PR #21964, opened 2024-03-04
Fix allow verification of valid code regardless of socketlabs connection

KevinMind:
Fixes: #21943

### Description

This pull request fixes an issue where the provided code for email verification was not being validated correctly. It ensures that the code is verified regardless of the socketlabs connection status.

### Context

In discussion with @theone we have decided to simplify some error states as well. Invalid codes and unauthorized codes should result in one single message.

### Testing

Make sure the waffle switch is on

```bash
./manage.py waffle_switch 'suppressed-email' on
```

Case: verify code before socketlabs
- create an email verification for your user in the DB. 

```python
SuppressedEmailVerification.objects.create(
  email_suppression=(
    SuppressedEmail.objects.create(email=<your-email>)
  )
)
```

- Set status to failed or pending (default is pending)
- Visit the email verify page with the code attached to the search params.

```python
email_verification.confirmation_code
```

Attach the code to the URL

`/developers/verify-email?code=<code>`

- expect email to be verified.

### DIFF
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -2089,7 +2089,7 @@ VERIFY_EMAIL_STATE = {
     'verification_failed': 'verification_failed',
     'verification_timedout': 'verification_timedout',
     'confirmation_pending': 'confirmation_pending',
-    'confirmation_unauthorized': 'confirmation_unauthorized',
+    'confirmation_invalid': 'confirmation_invalid',
 }
 
 RENDER_BUTTON_STATES = [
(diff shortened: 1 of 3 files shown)


eviljeff:
I'm not overly familiar with the code, but I can't discern in views.py the change that fixes the issue among the refactoring.  Can you highlight it?

src/olympia/devhub/views.py
```diff
@@ -2130,44 +2130,38 @@ def email_verification(request):
...
-                    ).exists():
-                        data['state'] = VERIFY_EMAIL_STATE['confirmation_unauthorized']
-                    else:
-                        return redirect('devhub.email_verification')
-                else:
-                    data['state'] = VERIFY_EMAIL_STATE['confirmation_pending']
-        else:
+        if email_verification.is_expired:
             data['state'] = VERIFY_EMAIL_STATE['verification_expired']
+        elif code := request.GET.get('code'):
```
KevinMind, line 2135: @eviljeff this is the change. I've extracted the check for a code the URL to be independent of the status of the email verification.

Before: we only check for a valid code if we confirm the email verification is sent by socketlbabs

Now: if a valid code is present then we validate it regardless of the status.

eviljeff (approved):

PR #21966, opened 2024-03-04
Automate L10n extract/merge with reusable docker run github action

KevinMind:
Relates to: #21951 

### Description

PR introduces 2 changes.
1. a reusable docker-run action that can be used to build and run arbitrary commands against our docker container in CI.
2. modifies the verify-docker-image workflow to run extract/push locales to test that this pattern works for extracting l10n.

### Context

There are 3 problems with this PR.

1. Currently using `docker build/run` instead of docker compose means we cannot rely effectively on running make commands in the same way we would running from a local environment. So long as you have the build and run inside the container it all works, but once it comes to running services for testing etc we will either have to reimplement the other services in github action services or switch to using docker compose in the docker-run job.
2. We have to rely on existing hacks to get file permissions working. In this case I had to copy the mounts we use for our docker-compose setup locally to ensure we have access to /deps and /data/olympia in the container. This works, but its fragile and means we have to keep these configs in sync with each other.
3. Running any make commands that rely on dev dependencies require installing dev depdencies as a part of the docker run step. This cannot be cached and wastes time and resources but our docker build doesn't currently support allowing to install dev dependencies.

All in all, this PR is probably fine to go ahead, it works, it does what we want. But as we move forward with GHA ci we will have to address the best way to run tests and do more complex workflows that require dependencies like database,nginx, etc. 

I have attempted using docker compose for the run command in ci, it kind of works, but not really. Problem numbe 2 is totally unimpacted and problem 1 reveals new and different problems.

### Testing

Only impacts CI. Possible but not good to test the run on master branch before merging. Though it is possible with the github CLI.

### DIFF
--- a/Makefile-docker
+++ b/Makefile-docker
@@ -94,7 +94,7 @@ update_deps_base: ## update the python and node dependencies
 .PHONY: update_deps
 update_deps: update_deps_base ## update the python and node dependencies for development
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/dev.txt
-	npm install $(NPM_ARGS)
+	npm install $(NPM_ARGS) --no-save
 
 .PHONY: update_deps_prod
 update_deps_prod: update_deps_base ## update the python and node dependencies for production
--- a/scripts/extract-l10n.sh
+++ b/scripts/extract-l10n.sh
@@ -90,7 +90,6 @@ EOF
 
 
 function create_pull_request {
-    info "Pushing the branch..."
     info "Creating the auto merge pull request for $BRANCH_NAME ..."
     curl --verbose -H "Authorization: token $GITHUB_TOKEN" --data "$(generate_post_data)" "https://api.github.com/repos/mozilla/addons-server/pulls"
     info "Pull request created."
--- a/scripts/extract-l10n.sh
+++ b/scripts/extract-l10n.sh
@@ -25,8 +25,6 @@ GIT_REMOTE="https://github.com/mozilla/addons-server.git"  # Upstream.
 REV=""
 MESSAGE=""
 BRANCH_NAME=""
-ROBOT_EMAIL="addons-dev-automation+github@mozilla.com"
-ROBOT_NAME="Mozilla Add-ons Robot"
 
 DRY_RUN=false
 
@@ -79,13 +77,6 @@ function init_environment {
     make -f Makefile-docker update_deps
 }
 
-function commit {
-    info "Committing..."
-    git -c user.name="$ROBOT_NAME" -c user.email="$ROBOT_EMAIL" commit -m "$MESSAGE" --author "$ROBOT_NAME <$ROBOT_EMAIL>" --no-gpg-sign locale/*/LC_MESSAGES/*.po locale/templates/
-    info "Committed locales extraction to local branch."
-}
-
-
 function generate_post_data()
 {
   cat <<EOF
@@ -100,7 +91,6 @@ EOF
 
 function create_pull_request {
     info "Pushing the branch..."
-    git push -q "https://addons-robot:$GITHUB_TOKEN@github.com/mozilla/addons-server/"
     info "Creating the auto merge pull request for $BRANCH_NAME ..."
     curl --verbose -H "Authorization: token $GITHUB_TOKEN" --data "$(generate_post_data)" "https://api.github.com/repos/mozilla/addons-server/pulls"
     info "Pull request created."
@@ -117,7 +107,7 @@ fi
 ./scripts/run_l10n_extraction.sh
 
 if [[ $DRY_RUN == false ]]; then
-  commit
+  ./scripts/push_l10n_extraction.sh
 
   # This script is meant to be run inside a virtualenv or inside our docker
   # container. If it's the latter, it doesn't necessarily have access to the ssh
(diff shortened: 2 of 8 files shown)


KevinMind:
Attempts:

- running fix_olympia_user in docker container as olympia doesn't work. olympia user cannot be modified if it is running the script
- running as root works, but then switching to olympia fails because /bin/nologin user
- changing host file permissions before mounting the volume fails with `Operation not permitted`

Makefile-docker
```diff
@@ -94,7 +94,7 @@ update_deps_base: ## update the python and node dependencies
 .PHONY: update_deps
 update_deps: update_deps_base ## update the python and node dependencies for development
 	$(PIP_COMMAND) install --progress-bar=off --no-deps --exists-action=w -r requirements/dev.txt
-	npm install $(NPM_ARGS)
+	npm install $(NPM_ARGS) --no-save
```
eviljeff, line 97: I guess this is harmless because it's just `npm install` without any package specified to install ourself but why would we need it adding?

scripts/run_l10n_extraction.sh
```diff
@@ -18,6 +18,7 @@ DJANGO_SETTINGS_MODULE=settings
 CLEAN_FLAGS="--no-obsolete --width=200 --no-location"
 MERGE_FLAGS="--update --width=200 --backup=none --no-fuzzy-matching"
 UNIQ_FLAGS="--width=200"
+DIFF_WITH_ONE_LINE_CHANGE="2 files changed, 2 insertions(+), 2 deletions(-)"
```
eviljeff, line 21: unused?

scripts/extract-l10n.sh
```diff
@@ -117,7 +108,7 @@ fi
 ./scripts/run_l10n_extraction.sh
 
 if [[ $DRY_RUN == false ]]; then
-  commit
+  ./scripts/push_l10n_extraction.sh
```
eviljeff, line 111: the new .sh doesn't just replace what `commit` does - it also replaces part of what `create_pull_request` does, so that needs adjusting too.

Makefile-docker
KevinMind: It prevents any changes to package.json. If for some reason running npm install would lead to a change then it's possible for l10n extraction to commit changes to package.json. I ran into this edge case before and that's why I added it. 

This way in order to update you have to actually npm install something, kind of like we need to hashin to update a requirements file.

scripts/run_l10n_extraction.sh
KevinMind: Done.

scripts/extract-l10n.sh
KevinMind: Good catch. Seems the typical flow requires the user to push manually anyway, but I removed the line which is pushing and only kept the line that is creating the PR.

I could actually just remove that script in this PR just didn't want to do it yet until we know the new flow is working well.

eviljeff: So have you removed the script? Or you're going to remove it before you merge? Or 😕 ...

KevinMind: I removed the git push on line 103

https://github.com/mozilla/addons-server/pull/21966/files#diff-9d8c85c350642e45c3ffb7863b8d8a995351b815fb2a2db0dcfc6bb00748682eL103

This way we can push via the push_l10n_extraction.sh script.

I would merge this, make sure it's working. And then sometime early next week open a PR removing the extract-l10n.sh script as a redundant piece of code.

eviljeff: oh, the `extract-l10.sh` script - I took "that script" to relate to the commented upon line - `push_l10n_extraction.sh`

KevinMind: I try to avoid pronouns but everyone and a while I get lazy. Almost always immediately explodes. That should never be so vague. (see what I did there 😆 ) I'll resolve this (did it again)

eviljeff (approved):
I think... this is fine for l10n extraction.  The hacks in action.yml may end up being problematic as it's used for other automation... but I guess that's a future problem.
I'd be interested in @diox's opinion too.

KevinMind:
> The hacks in action.yml may end up being problematic as it's used for other automation... but I guess that's a future problem.

Couldn't have said it better myself. There are 2 "hacks" currently:

- cat bash scripts into the container. This is to allow the `run` property in the action configuration to pass multiline scripts and not have to escape each line

```yaml
- uses: /.github/actions/run
  with:
    run: |
      echo "one"
      echo "one"
      echo "one"
      echo "one"
```

Looks much nicer than

```yaml
- uses: /.github/actions/run
  with:
    run: |
      echo "one" \
      && echo "one" \
      && echo "one" \
      && echo "one"
```

The other hack is that we have to run the container as "root", then run a script to change the UID/GID of the "olympia" user in the container, and then execute the rest of the logic as the olympia user. THis is essentially the same "hack" that we have in our docker-compose.yml for local development and AFAICT and have tried, the only way around this is to build our image based on the UID/GID of the host user or to run the container as root, both are kind of non starters at least right now.

The bigger issue I think we will run into, and I know we will cause I've already tried, is that much of our makefile logic is built around the idea of running via docker compose and not via a single docker container.

This action is totally useless for that setup. I think there are many places we can go with this and like you said, it's kind of a next step problem, but this is actually going to work for a large chunk of what we currently do in CI.

diox (approved):

Issue #21976, opened 2024-03-05
Run collect static after all files are generated


KevinMind:
After merging https://github.com/mozilla/addons-server/pull/21960 we realized there was another latent bug in the build resulting in generated i18n files being excluded from the static files directory in the final container.

STR
1. run docker build
2. run docker container
3. inspect site-static dir

expected:

generated js files are in the directory

actual:

they are not.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-748)

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-748

[...]

PR #21977, opened 2024-03-05
Collect static files after generating all static files

KevinMind:
Fixes: #21976

### Description

This pull request ensures that static files are collected after generating all static files. This prevents missing files and ensures that the process is executed in the correct order.

### Context

The commands for updating static assets were split between the makefile and the dockerfile before, and in the dockerfile the order was correct.

IN https://github.com/mozilla/addons-server/pull/21914 that entire operation was moved to the makefile and the ordering was included incorrectly.

### Testing

This bug can only be reproduced at runtime so we would need to include build time checks to inspect static assets are present to verify this. We should do this eventually.


diox (approved):
(It's possible to try this locally with a `DEBUG=False` env but I haven't had time to try this)

Issue #21981, opened 2024-03-07
Fix hidden bugs in docker build


KevinMind:

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-750

[...]

PR #21983, opened 2024-03-07
Fix small bugs in the docker build that are not obvious most of the time

KevinMind:
Fixes: #21981

### Description

2 changes to scripts executed during our docker build

- add the npm prefix flag to `prun` to ensure we prune the correct directory and not the default ./node_modules which is already empty

- hard failure on the compile locales script when `dennis-cmd` is not installed. Removing the venv fallback

### Context

These issues can cause failure in certain contexts, or not result in failure in other contexts. This PR ensures we have consistent and correct behavior.

### Testing

You can run a build and inspect the image before and after and see the pruned /deps/node_modules directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -138,6 +138,6 @@ COPY --chown=olympia:olympia . ${HOME}
 # This will shave nearly 1 minute off the best case build time
 RUN echo "from olympia.lib.settings_base import *" > settings_local.py \
     && DJANGO_SETTINGS_MODULE="settings_local" make update_assets \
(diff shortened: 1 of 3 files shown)


eviljeff:
can we have a descriptive PR title so the eventual merged commit will be comprehensible

KevinMind:
> can we have a descriptive PR title so the eventual merged commit will be comprehensible

Keep forgetting to do that.

eviljeff (approved):

PR #21995, opened 2024-03-11
Allow banned users to appeal by grabbing the user from the cinderjob target

### LINKED ISSUE #21949: Appeals sent by a user after being reported didn't land in Cinder
I could no longer reproduce this issue but I remember testing with rusiczki.ioana+new1@gmail.com in -stage and -dev

-> currently appeals are sent https://stage.cinder.nonprod.webservices.mozgcp.net/job/20cb683b-c7f5-459f-88c7-987429a93051

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-741)

diox:
Fixes #21949


src/olympia/abuse/models.py
```diff
@@ -354,12 +354,21 @@ def appeal(self, *, abuse_report, appeal_text, user, is_reporter):
...
+                user = self.target
             if not user:
-                # If the appealer is not an original reporter, we have to
-                # provide an authenticated user that is the author of the
-                # content.
+                # If we still don't have a user at this point there is nothing
+                # we can do, something was wrong in the call chain.
                 raise ImproperlyConfigured(
-                    'CinderJob.appeal() called with is_reporter=False without a user'
+                    'CinderJob.appeal() called with is_reporter=False without a ' 'user'
```
eviljeff, line 371: unnecessary change

src/olympia/abuse/models.py
```diff
@@ -354,12 +354,21 @@ def appeal(self, *, abuse_report, appeal_text, user, is_reporter):
...
+            # If the appealer is not an original reporter, we have to provide
+            # an authenticated user that is the author of the content.
+            # User bans are a special case, since the user can't log in any
+            # more at the time of the appeal, so we let that through using the
+            # target of the job that banned them.
+            if (
+                not user
+                and self.decision_action == CinderJob.DECISION_ACTIONS.AMO_BAN_USER
+            ):
+                user = self.target
```
eviljeff, line 366: should we verify that self.target is actually a UserProfile instance at this point, or are we trusting that the data won't ever be that bad

eviljeff (approved):

src/olympia/abuse/models.py
diox: Let's be a little more robust just in case b38d7b00f7d60fd2cee009cb916bf407980f4c4e

PR #21998, opened 2024-03-12
Sync suppressed emails daily

KevinMind:
FIxes: # <TBD>

### Description

Rename sync_suppressed_emails to reflect similar usage

sync_suppressed_emails(_cron|task|command)

Added a cron job to execute the task daily.

Added some docs describing how and when to use different execution types

### Context

I would be very surprised if there weren't opinions about the docs I wrote. That was kind of the point to solicit feedback and improvements so fire away. I'm more than happy to split that commit to a separate PR to unblock this one but wanted to get the ball rolling.

### Testing

You should be able to sync the email suppression list via the management console, via triggering the cron with a fake time (not sure if that's possible) or by manually triggering the task in a shell. 

It is idempotent so you should be able to run over and over and end up with a synced DB reflecting the current state of the socketlabs DB.

To test locally you will need to get a SOCKET_LABS_TOKEN from the socketlabs portal. Also don't forget to set the SOCKET_LABS_SERVER_ID in your local settings to the corresponding server ID.


docs/topics/development/django.md
eviljeff, line 1: yeah... lets split this file out into another PR - it's only tangentially connected to the rest of this patch and I have comments...

eviljeff (approved):
r+wc - strip out the new doc into a separate PR

src/olympia/users/management/commands/sync_suppressed_emails.py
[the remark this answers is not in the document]
KevinMind: Which command is obsolete?

docs/topics/development/django.md
KevinMind: Figured would be the case. Given other comments it is clear the doc is needed, but we can iterate.

PR #22000, opened 2024-03-12
"rename" CinderJob.decision_id to decision_cinder_id

### LINKED ISSUE #21999: rename CinderJob.decision_id column
https://github.com/mozilla/addons-server/issues/21990 will add a `decision` foreign key to `CinderJob` ... which means a `decision_id` column to hold the pk.  But we already have a different (type charfield) column called `decision_id`.  Lets rename it beforehand (actually create a new column and then copy the values across)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-756)

eviljeff:
fixes #21999 - just renames the column from decision_id to decision_cinder_id.


KevinMind (approved):
Assuming we test this via CI but looks good to me.

eviljeff:
> Assuming we test this via CI but looks good to me.

yes, apart from the `RunPython` migration - as the test setup starts with an empty database it doesn't fully test the case of migrating existing data.  But it's super trivial so it doesn't concern me.  

If you did want to test this locally:
- you'd need at least one `CinderJob` with a `decision_id` set (would need either manually setting via django admin before the patch was checked out, or via db admin after)
- then run `./manage.py migrate abuse` (or `make update_db` if you wanted to run all the migrations)
- check that `CinderJob.decision_cinder_id` now has the value that `decision_id` had.

PR #22005, opened 2024-03-12
Simplify socketlabs check for status and messaging when something goes wrong.

KevinMind:
Relates to: https://github.com/mozilla/addons-server/issues/21687

### Description

This PR simplifies the existing logic for checking the socketlabs recipient search API for delivery status of email verification sent via AMO. Instead of checking for specific statuses, we check for "Delivered" or anything else. 

If we find at least one email that is "Delivered" we mark the verification as sent, otherwise we return a list of all emails we found and their associated meta data to display to the user to help them debug why they didn't receive the email.

Finally, there are some additional copy changes that are included in this PR.

### Context

We need to review this UX change with @wagnerand but I believe this is a better UX and also simpler code without totally removing the feature.

### Testing

You can follow the flow in the attached videos that display a successful and an unsuccessful delivery. 

### Standard attempt

This example demonstrates a successful attempt where an email is sent and found to be delivered shortly after refreshing the screen.

https://github.com/mozilla/addons-server/assets/19595165/55d5b628-3429-4f84-96ba-22a2fb24358e

#### Failed attempts

This example demonstrates the UX if an email is sent but not "delivered", in this case "suppressed".
It also demonstrates the UX of using an invalid code with the new UX (already merged a while ago but just to demonstrate)

https://github.com/mozilla/addons-server/assets/19595165/35ea08c0-12b9-431e-b282-ae2524cbd793

> NOTE: UI for non delivered emails is now a table

Screenshot 2024-03-19 at 08 53 35

### DIFF
--- a/src/olympia/devhub/templates/devhub/verify_email.html
+++ b/src/olympia/devhub/templates/devhub/verify_email.html
@@ -20,21 +20,28 @@
   Could not verify email address. The verification link has expired.
   {% endtrans %}
 {% elif state == "verification_pending" %}
-  {% trans %}
-  Working... Please be patient.
-  {% endtrans %}
   <div class="loader"></div>
-{% elif state == "verification_failed" %}
-  {% trans %}
-  Failed to send confirmation email. Please try again.
-  If you no longer have access to your email address, please update your mozilla account email address.
-  <a
-    href="https://support.mozilla.org/en-US/kb/change-primary-email-address-firefox-accounts"
-    target="_blank"
-  >
-    Change email address
-  </a>
-  {% endtrans %}
+  {% if found_emails|length > 0 %}
+    <p>
+      {% trans %}
+      We have attempted to send your verification email but ran into an issue.
+      Below are the records we found and the associated delivery status.
+      This page will refresh automatically in 10 seconds to check again.
+      You can click 'try again' to attempt sending another email.
+      {% endtrans %}
+    </p>
+    <ul>
+    {% for email in found_emails %}
+      <li>
+        Subject: {{ email.subject }} ({{ email.status}})
+      </li>
+    {% endfor %}
+    </ul>
+  {% else %}
+    {% trans %}
+    Working... Please be patient.
+    {% endtrans %}
+  {% endif %}
 {% elif state == "verification_timedout" %}
   {% trans %}
   This is taking longer than expected. Try again.
--- a/src/olympia/devhub/templates/devhub/verify_email.html
+++ b/src/olympia/devhub/templates/devhub/verify_email.html
@@ -35,7 +35,7 @@
   {% endtrans %}
 {% elif state == "confirmation_invalid" %}
   {% trans email=request.user.email %}
-  The provided code is invalid, unauthorized, expired or incomplete. Please use the link in the email sent to your email address {{ email }}. If the code is still not working please request a new email.
+  The provided code is invalid, unauthorized, expired or incomplete. Please use the link in the email sent to your email address {{ email }}. If the code is still not working, please request a new email.
 
 
   {% endtrans %}
--- a/src/olympia/devhub/templates/devhub/verify_email.html
+++ b/src/olympia/devhub/templates/devhub/verify_email.html
@@ -7,6 +7,7 @@
 {% block content %}
 <h1>{{ title }}</h1>
 <div id="{{ state }}">
+<div class="verify-email-text">
 {% if state == "email_verified" %}
   {% trans %}
   Your email address is verified.
@@ -21,27 +22,9 @@
   {% endtrans %}
 {% elif state == "verification_pending" %}
   <div class="loader"></div>
-  {% if found_emails|length > 0 %}
-    <p>
-      {% trans %}
-      We have attempted to send your verification email but ran into an issue.
-      Below are the records we found and the associated delivery status.
-      This page will refresh automatically in 10 seconds to check again.
-      You can click 'try again' to attempt sending another email.
-      {% endtrans %}
-    </p>
-    <ul>
-    {% for email in found_emails %}
-      <li>
-        Subject: {{ email.subject }} ({{ email.status}})
-      </li>
-    {% endfor %}
-    </ul>
-  {% else %}
     {% trans %}
     Working... Please be patient.
     {% endtrans %}
-  {% endif %}
 {% elif state == "verification_timedout" %}
   {% trans %}
   This is taking longer than expected. Try again.
@@ -57,6 +40,49 @@
     - could be expired
     - could be a mistake
   {% endtrans %}
+{% endif %}
+</div>
+{% if found_emails|length > 0 %}
+<div class="verify-email-table">
+  <p>
+    {% trans %}
+      We have attempted to send your verification email.
+      Below are the confirmation records we found and their associated delivery statuses.
+    {% endtrans %}
+  </p>
+  <table border=1 frame=void rules=rows>
+    <thead>
+      <tr>
+        <th>{{ _('Date') }}</th>
+        <th>{{ _('From') }}</th>
+        <th>{{ _('To') }}</th>
+        <th>{{ _('Subject') }}</th>
+        <th>{{ _('Status') }}</th>
+      </tr>
+    </thead>
+    <tbody>
+    {% for email in found_emails %}
+      <tr>
+        <td>
+          {{ email.statusDate }}
+        </td>
+        <td>
+          {{ email.from }}
+        </td>
+        <td>
+          {{ email.to }}
+        </td>
+        <td>
+          {{ email.subject }}
+        </td>
+        <td>
+          {{ email.status }}
+        </td>
+      </tr>
+    {% endfor %}
+    </tbody>
+  </table>
+</div>
 {% endif %}
 {% if render_button %}
   {% with submit_text=button_text %}
--- a/src/olympia/devhub/templates/devhub/verify_email.html
+++ b/src/olympia/devhub/templates/devhub/verify_email.html
@@ -23,11 +23,11 @@
 {% elif state == "verification_pending" %}
   <div class="loader"></div>
     {% trans %}
-    Working... Please be patient.
+    We are sending an email to you, this might take a minute. The page will automatically refresh.
     {% endtrans %}
 {% elif state == "verification_timedout" %}
   {% trans %}
-  This is taking longer than expected. Try again.
+  It is taking longer than expected to confirm delivery of your verification email. Please try again.
   {% endtrans %}
 {% elif state == "confirmation_pending" %}
   {% trans email=request.user.email %}
@@ -35,10 +35,9 @@
   {% endtrans %}
 {% elif state == "confirmation_invalid" %}
   {% trans email=request.user.email %}
-    The provided code is invalid. Please use the link in the email sent to your email address {{ email }}.
-    - could be unauthorized
-    - could be expired
-    - could be a mistake
+  The provided code is invalid, unauthorized, expired or incomplete. Please use the link in the email sent to your email address {{ email }}. If the code is still not working please request a new email.
+
+
   {% endtrans %}
 {% endif %}
 </div>
--- a/src/olympia/users/utils.py
+++ b/src/olympia/users/utils.py
@@ -385,8 +385,11 @@ def check_suppressed_email_confirmation(verification, page_size=5):
             if code_snippet in item['subject']:
                 found_emails.append(
                     {
+                        'from': item['from'],
+                        'to': item['to'],
                         'subject': item['subject'],
                         'status': item['status'],
+                        'statusDate': item['statusDate'],
                     }
                 )
 
--- a/src/olympia/users/utils.py
+++ b/src/olympia/users/utils.py
@@ -1,6 +1,8 @@
 import base64
+import datetime
 import hashlib
 import hmac
+import urllib
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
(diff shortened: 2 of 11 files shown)


wagnerand:
Could you provide screenshots or screencasts for the things you'd like UX reviewed?

KevinMind:
> Could you provide screenshots or screencasts for the things you'd like UX reviewed?

Added them.

wagnerand:
@KevinMind yes, unfortunately the video resolution is too low to properly follow what is going on. Also, what exact UX are you seeking review for?

KevinMind:
Specifically for the change in the pending screen where we have attempted to send the email but have not yet confirmed. 

Because there are so many edge cases and testing is essentially only possible in production, I simplified the flow. 

If you are waiting for the email to be sent and we have found emails that have been attempted to be sent via the socket labs API. And if neither of those are delivered then we just show a list of all the email subjects and statuses to the user. This is simpler than trying to infer the state based on the status and mapping it to a field in our database. 

If we go with this approach we only care if an email is delivered and everything else is treated as unknown and just displayed to the user. 

It also simplifies the implementation by fetching the data every time you request the verify email page instead of buying asynchronous task. Data is always fresh because it's fetched on demand. 

The downside is rendering. The page is a bit slower but I don't think this is a big deal because the users are expecting to be waiting for something anyway. 

There are a lot of assumptions in there so that's why I'm asking for the review.

wagnerand:
The changes and implications are not clear to me. Could you describe the flow, the assumptions you made, implications, and the edge cases you came across?

More questions specific to your comment:

> If you are waiting for the email to be sent and we have found emails that have been attempted to be sent via the socket labs API.

This sentence seems to be missing a _then_?

> And if neither of those are delivered then we just show a list of all the email subjects and statuses to the user.

What does _neither_ refer to here?
What does an subject look like? What value does showing this hold?
Do we need to show other data in addition, for example a timestamp?
How can there be multiple emails?
Do we expect the user to have an understanding what the statuses we show mean?

src/olympia/users/utils.py
```diff
@@ -285,3 +288,108 @@ def upload_picture(user, picture):
...
+
+
+utils_log = olympia.core.logger.getLogger('z.users')
+
+
+def check_suppressed_email_confirmation(verification=None, page_size=5):
+    assert_socket_labs_settings_defined()
+
+    if not verification:
+        raise Exception('Verification is not defined')
```
eviljeff, line 311: Not sure what this is for - why have `verification` be an optional parameter if we want to raise for it not being provided?

src/olympia/users/utils.py
```diff
@@ -285,3 +288,108 @@ def upload_picture(user, picture):
...
+
+        headers = {
+            'authorization': f'Bearer {settings.SOCKET_LABS_TOKEN}',
+        }
+
+        utils_log.info(f'checking for {code_snippet} with params {params}')
+
+        response = requests.get(url, headers=headers)
+        response.raise_for_status()
+        json = response.json()
```
eviljeff, line 365: `json` here is overriding a built in python library 😬

src/olympia/users/utils.py
```diff
@@ -285,3 +288,108 @@ def upload_picture(user, picture):
...
+        'sortDirection': 'dsc',
+    }
+
+    is_first_page = True
+
+    found_emails = []
+
+    while current_count < total or is_first_page:
+        if not is_first_page:
+            params['pageNumber'] = params['pageNumber'] + 1
```
eviljeff, line 349: I don't want to bikeshed on optimal algorithms, but you could just initialize `total` to `-1`, always update it to the value of `json['total']`, and move `params['pageNumber'] = params['pageNumber'] + 1` to the end of the loop, to get rid of this `is_first_page` variable along with all the special handling of that case.

eviljeff (commented):
Just so I understand what the major change is with this PR: now every time a developer tries to validate their email address a synchronous call is made to socketlabs to get the all the suppressed emails and then we show the emails for that developer's email address on screen?

src/olympia/users/utils.py
KevinMind: Fair point. That part of the code hasn't changed since December. I'd rather handle that change in a separate PR since this already includes quite a large diff.

src/olympia/users/utils.py
KevinMind: Would something like this make more sense? Making it a required argument and assert it is a suppressed email verification?

    assert isinstance(verification, SuppressedEmailVerification)

src/olympia/users/utils.py
KevinMind: Odd that this is possible to commit. Vscode is also happy to see this.

KevinMind:
@wagnerand updated the videos, should be legible now. To your questions, I think we both know that I nor probably anyone really has answers but, let's try:

**How can there be multiple emails?**

Socketlabs can attempt to send an email any number of times, it might soft bounce and then get through. We ask them to send a mail, they do their best to make it happen.

**What does an subject look like? What value does showing this hold?**

Subject's are all templates so just "Verify your email ({{code_snippet}})"

**Do we need to show other data in addition, for example a timestamp?**

Good question, maybe? 

What does neither refer to here?

If non of the emails have a "delivered" status, then we show them all.

**Do we expect the user to have an understanding what the statuses we show mean?**

I don't really think so. But if a user actually get's into a snag, and they reach out to support. Having the info might be helpful. I also tend to think showing something rather than nothing is probably helpful. IF the user is seeing "suppressed" and they attempt to "fix" their email account then maybe it helps them. If they see it was bounced maybe it also helps them realize their email account is broken.. idk.

wagnerand:
Thank you for updating the videos.

> **Do we expect the user to have an understanding what the statuses we show mean?**
> 
> I don't really think so. But if a user actually get's into a snag, and they reach out to support. Having the info might be helpful. I also tend to think showing something rather than nothing is probably helpful. IF the user is seeing "suppressed" and they attempt to "fix" their email account then maybe it helps them. If they see it was bounced maybe it also helps them realize their email account is broken.. idk.

I wonder how many people even understand what the words "bounced" or "suppressed" mean. Maybe we should at least link to the socketlabs documentation about statuses, if we use their terminology.

KevinMind:
@wagnerand I accidnetally edited your comment instead of quote replying..... did not know that was possible.

I'm happy to do that, or update copy, anything you want. I think it's also important to think about this as an opportunity cost.

Current state: only catching specific types of "errors" that socketlabs encounters. If we run into a new type of error, we have to update the code to handle it. Also have to decide how to communicate that to the user. This really doesn't scale well and testing is next to impossible.

Alternative state: basically just show what we got from socketlabs. We might be communicating things in an imperfect way and it's unclear if the information is useful. It scales well because we just pass info from socketlabs on to the user and they decide what to do with it.

I don't think either is great, but the second scales better, is more resilient and the "ambiguity" that we would get from this approach is still there in the other approach, just hidden behind our own interpretation of what is going on. If the email was "suppressed" or "bounced" in the current flow we just say "something went wrong, try again"

Can you have a look at the [template](https://github.com/mozilla/addons-server/pull/22005/files#diff-2b394f74ad411dcbe97757ed27c9c57e20333091158b2210c44b7a11dce32a8eR24-R44) and add a comment with how you would update the text?

src/olympia/users/utils.py
eviljeff: If you were feeling particularly paranoid, I guess.  But unless the `verification` value passed in somehow happened to have a `.suppressed_email.email` attribute it'd raise on the next line anyway.

eviljeff (approved):
r+wc
- I'd like @wagnerand to give his 👍 to the changes
- I assume we've considered the volume of API calls this change will lead to and we're not going to hit some account limit on the socketlabs side for pinging them for every blocked developer?

wagnerand:
I think we should add the timestamp of the emails (and maybe the "From" header value) so that the user has an understanding of timing, as well as link to the socketlabs documentation for statuses if we are exposing them as is.

From the screencasts, it's not very easy to follow and understand the flow, so I am not sure no more follow ups are needed after this has been enabled on prod.

One thing I would do is to rename the "Try again" button to use something a bit more descriptive what it is, we are re-trying. Maybe "Send another email" or similar. Does this have a cooldown period so that it can't be abused?

KevinMind:
I think we should try to get this to prod as quickly as possible. We are very late on getting the feature out at all and we can learn from how users interact with the feature what changes should be made. I'll add the changes you suggested here and then hopefully we can merge and do another iteration post release.

src/olympia/devhub/templates/devhub/verify_email.html
```diff
@@ -20,21 +21,10 @@ <h1>{{ title }}</h1>
...
-  >
-    Change email address
-  </a>
-  {% endtrans %}
+    {% trans %}
+    Working... Please be patient.
+    {% endtrans %}
 {% elif state == "verification_timedout" %}
   {% trans %}
   This is taking longer than expected. Try again.
```
wagnerand, line 30: Can we clarify what "This" is and what we want users to "Try again"?

src/olympia/devhub/templates/devhub/verify_email.html
```diff
@@ -51,6 +41,49 @@ <h1>{{ title }}</h1>
     - could be a mistake
```
wagnerand, line 41: In public facing communication, we should use complete sentences. How about the following: 

_The provided code is invalid, unauthorized, expired or incomplete. Please use the link in the email sent to your email address {{ email }}._

If you'd like to keep the list format, it should start with something like "It could be". Still I would move the CTA below the list, then.

src/olympia/devhub/templates/devhub/verify_email.html
```diff
@@ -20,21 +21,10 @@ <h1>{{ title }}</h1>
...
-  If you no longer have access to your email address, please update your mozilla account email address.
-  <a
-    href="https://support.mozilla.org/en-US/kb/change-primary-email-address-firefox-accounts"
-    target="_blank"
-  >
-    Change email address
-  </a>
-  {% endtrans %}
+    {% trans %}
+    Working... Please be patient.
```
wagnerand, line 26: I am not sure what "working" means, is this when we are triggering a confirmation email? If so:

_We are sending an email to you, this might take a minute. The page will automatically refresh._

Or is this while the page is loading and we are fetching all the data? Then:

_We are currently fetching records for your email address, the page will refresh automatically._

src/olympia/devhub/templates/devhub/verify_email.html
KevinMind: Updated

src/olympia/devhub/templates/devhub/verify_email.html
KevinMind: It is both. Updated with the first one.

src/olympia/devhub/templates/devhub/verify_email.html
KevinMind: @@flodolo requested the list format as a more readable version. I'm not sure why we shouldn't be allowed to use bullet lists also.

KevinMind: Updated to sentence.

src/olympia/devhub/templates/devhub/verify_email.html
KevinMind: Updated.

src/olympia/devhub/templates/devhub/verify_email.html
```diff
@@ -20,37 +21,68 @@ <h1>{{ title }}</h1>
...
   {% trans email=request.user.email %}
   An email with a confirmation link has been sent to your email address: {{ email }}. Please click the link to confirm your email address. If you did not receive the email, please check your spam folder.
   {% endtrans %}
 {% elif state == "confirmation_invalid" %}
   {% trans email=request.user.email %}
-    The provided code is invalid. Please use the link in the email sent to your email address {{ email }}.
-    - could be unauthorized
-    - could be expired
-    - could be a mistake
+  The provided code is invalid, unauthorized, expired or incomplete. Please use the link in the email sent to your email address {{ email }}. If the code is still not working please request a new email.
```
flodolo, line 38:
```suggestion
  The provided code is invalid, unauthorized, expired or incomplete. Please use the link in the email sent to your email address {{ email }}. If the code is still not working, please request a new email.
```

Probably worth a comma in the last sentence.

wagnerand (approved):

src/olympia/devhub/templates/devhub/verify_email.html
```diff
@@ -20,37 +21,68 @@ <h1>{{ title }}</h1>
...
   An email with a confirmation link has been sent to your email address: {{ email }}. Please click the link to confirm your email address. If you did not receive the email, please check your spam folder.
   {% endtrans %}
 {% elif state == "confirmation_invalid" %}
   {% trans email=request.user.email %}
-    The provided code is invalid. Please use the link in the email sent to your email address {{ email }}.
-    - could be unauthorized
-    - could be expired
-    - could be a mistake
+  The provided code is invalid, unauthorized, expired or incomplete. Please use the link in the email sent to your email address {{ email }}. If the code is still not working, please request a new email.
+
```
flodolo, line 39: Are the extra blank lines expected/wanted?

src/olympia/users/utils.py
```diff
@@ -285,3 +288,112 @@ def upload_picture(user, picture):
...
+            'authorization': f'Bearer {settings.SOCKET_LABS_TOKEN}',
+        }
+
+        utils_log.info(f'checking for {code_snippet} with params {params}')
+
+        response = requests.get(url, headers=headers)
+        response.raise_for_status()
+        json_data = response.json()
+
+        utils_log.info(f'recieved data {json_data} for {code_snippet}')
```
flodolo, line 368:
```suggestion
        utils_log.info(f'received data {json_data} for {code_snippet}')
```

flodolo (approved):

PR #22006, opened 2024-03-12
Document types of work in our app (utils/cron/tasks/model methods)

KevinMind:
Fixes: https://github.com/mozilla/addons/issues/14846

### Description

Document where and how we add logic to the code base.

### Context

Need for this has come up in several related PRs. I think we need some info covering what types of execution contexts we support, and general patterns and best practices for implementing those as well as when to prefer one or the other.

### Testing

Docs.

### DIFF
--- a/docs/topics/development/django.md
+++ b/docs/topics/development/django.md
@@ -1,0 +1,39 @@
+# Django (tips and best practices)
+
+## Types of work
+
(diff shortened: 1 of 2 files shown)


KevinMind:
https://www.github.com/mozilla/addons/14844

PR #22007, opened 2024-03-12
Create github action to generate a tag/release from master

KevinMind:
Fixes: Manual release process

### Description

Github action to create tags for release. You can dispatch via the action web interface or via cli. One day, maybe we will create a tag on every push to master, who knows...

### Context

This is a good first step in automating the deployment phase of our release process. We already deploy to stage when a tag is pushed and this automates the generation of the tag, meaning you can deploy to stage via a button 💯 

### Testing

You can run this action via the github cli

```bash
gh workflow run --ref auto-tag
```

Then select the appropriate workflow

```bash
> Tag Repository for Release to Staging (tag-release.yml)
```

And select your options for `push` and `tag`

Test cases

#### Invalid tag

1. pass "banana" as tag
2. expect error "does not match the format YYYY.MM.DD"

1. pass "2024-03-12"
2. expect same error as above.

#### Invalid date

1. Pass "1999.12.31"
2. expect error "is earlier than current date"

You must provide a valid date string which is greater or equal to today's date.

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -1,0 +1,109 @@
+name: Tag Repository for Release to Staging
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Tag (YYYY.MM.DD) to push. Defaults to current day"
+        required: false
+      push:
+        description: "Should the tag be pushed to github. (false means dry run)"
+        required: false
+        default: 'false'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.ref }}
+  cancel-in-progress: true
+
+jobs:
+  tag:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          #always tag from master
+          ref: 'master'
+          # get whole history (not great but we need tags)
+          fetch-depth: '0'
+          # get tags so we can diff latest
+          fetch-tags: 'true'
+          # show progress for debugging
+          show-progress: true
+      - name: Get and validate date
+        id: date
+        shell: bash
+        run: |
+          CURRENT_DATE=$(date '+%Y.%m.%d')
+          DATE=$CURRENT_DATE
+
+          if [[ -n "${{ inputs.tag }}" ]]; then
+            DATE="${{ inputs.tag }}"
+
+            regex='^([0-9]{4})\.(0[1-9]|1[0-2])\.(0[1-9]|[12][0-9]|3[01])$'
+
+            if [[ $DATE =~ $regex ]]; then
+                input_date="${BASH_REMATCH[1]}.${BASH_REMATCH[2]}.${BASH_REMATCH[3]}"
+                if [[ $DATE < $CURRENT_DATE ]]; then
+                    echo "Error: Input date: \"$DATE\" is earlier than current date: \"$CURRENT_DATE\""
+                    exit 1
+                fi
+            else
+                echo "Error: Input date: \"$DATE\" does not match the format YYYY.MM.DD"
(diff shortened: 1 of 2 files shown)


eviljeff:
> Fixes: Manual release process

still needs an issue 😛 

> One day, maybe we will create a tag on every push to master, who knows...

I'd argue if you were tagging on every commit you don't need a tag because they're the same thing (and you wouldn't need a stage server deploying off the tags because it'd be the same code as dev) but I won't get bogged down in hypothetical future scenarios.

.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,122 @@
+name: Tag Repository for Release to Staging
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Tag (YYYY.MM.DD) to push. Defaults to current day"
+        required: false
```
eviljeff, line 8: I think this should be required - tagging for the current day is an atypical use case - the typical case is today+2, but I think even it's clearer to require it

.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,122 @@
...
+      push:
+        description: "Should the tag be pushed to github. (false means dry run)"
+        required: false
+        default: 'false'
```
eviljeff, line 12: I lean towards pushing being the default?  (otherwise what's the purpose of the action?) with maybe a `dry-run` parameter for testing?

.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,122 @@
...
+
+      - name: Get next tag
+        id: to_tag
+        shell: bash
+        run: |
+          from_tag="${{ steps.from_tag.outputs.tag }}"
+          to_tag="${{ steps.date.outputs.date }}"
+
+          revision=""
+          is_cherry_picked=false
```
eviljeff, line 76: doesn't seem like this is used?

eviljeff (changes requested):
needs issue.

Also, I wasn't clear on if it's possible to execute this via the GH web UI when there are parameters?

KevinMind:
It is possible. You'll see a form to provide inputs. Agree this solves the current use case and in a more continuous deployment strategy the benefit of tags does dissipate.

I could imagine a scenario where you create tags that don't get auto deployed to stage and instead you manually trigger a staging deploy and get to select the tag you want. That is much closer to what we do now and still benefits from automation but yeah, it's a next next step in any case. 

I'll add an issue.

.github/workflows/tag-release.yml
eviljeff: The tag date is the date it will go to production so we can investigate historically when features actually went live.  Naming it as tagging date doesn't help with that, whereas you can find out in git history when a tag was created.

eviljeff: Changing out tagging convention seems out of scope for this PR.  So, currently, we pretty much never create a tag named for the current day, so it's a footgun to default to it.

.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,112 @@
...
+        description: "Tag (YYYY.MM.DD) to push. Defaults to current day"
+        required: true
+      push:
+        description: "Should the tag be pushed to github. (false means dry run)"
+        required: true
+        default: 'false'
+      release:
+        description: "Should a draft release be created in gitub."
+        required: true
+        default: 'false'
```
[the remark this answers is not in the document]
KevinMind: Oddly it is.

KevinMind:
https://jkrsp.com/deploy-and-rollback-releases-with-github-actions/

PR #22023, opened 2024-03-15
Refactor decision related fields into CinderDecision

### LINKED ISSUE #21990: Add models to allow cinder decisions to exist without a cinder job or abuse report
part of https://mozilla-hub.atlassian.net/browse/AMO-116

Currently our models expect an abuse report to be file against a target (either add-on, rating, collection, or user); that report to be submitted creating a job, and then a decision made on the job.  But we also need to make decisions on content that didn't come from an abuse report and without the abuse report we don't have a target, and without a report we don't have a job.  We need to create new models and migrate some data from the old models.

eviljeff:
fixes #21990 

... but this is only the first step.  Everything appeal related, and probably the `CinderAction` classes too, need untangling from `CinderJob` in favour of `CinderDecision` - and anything that assumes there is always an abuse report at the beginning of any chain.  That's the follow-up.

Most of the changes outside of abuse/models.py are minimal though, and the test changes are mainly creating extra objects and being forced to properly set up (e.g. `CinderDecision` has a compulsory `action` and must have a target fk)

iirc you weren't in favour of using a migration to backfill but I wrote it anyway to see how practical it was under the artificial migration environment.  I could swap it out for a manual migration if you like, but we lose the ability to access the deleted fields values (`CinderJob.decision_cinder_id` no longer exists on the `CinderJob` model in real code, but does in the django migration environment) so we'd have to keep the deleted fields on the model as `_decision_cinder_id`, etc


src/olympia/abuse/migrations/0028_cinderdecision_backfill.py
```diff
@@ -0,0 +1,67 @@
...
+        # also track the new m2m relationships we'll need the policies, and the back links to the job
+        policies_per_job[cinder_job.decision_cinder_id] = list(cinder_job.policies.all())
+        cinder_jobs[cinder_job.decision_cinder_id] = cinder_job
+
+    # Bulk create the CinderDecisions
+    CinderDecision.objects.bulk_create(decisions_to_create)
+
+    m2m_policies_to_create = []
+    # Loop through the CinderDecisions, to generate the m2m list and to set cinderjob.decision
+    for decision in CinderDecision.objects.all():
```
eviljeff, line 45: Having to do a query to get the decisions was a little annoying, but it appears mysql doesn't return auto generated ids so (unlike, say, postgres) `bulk_create` doesn't return objects that have been saved, so they can't be used for foreign keys.

src/olympia/abuse/migrations/0028_cinderdecision_backfill.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('abuse', '0027_add_cinderdecision'),
+    ]
+
+    operations = [
+        migrations.RunPython(backfill_cinder_decisions_from_cinder_jobs)
```
diox, line 67: Add a reverse function that deletes all `CinderDecision` to make the data migration reversible, it might come handy...

diox (approved):
I went back and forth in my head regarding management command vs migration for the backfill, and in the end even though I don't like the lack of tests I think we can get away with that with the amount of test data on dev & stage (it might be one of the rare cases where we actually have more data to migrate on dev/stage than we have in prod... hell I might have more decisions locally that we have in prod), so :+1:

src/olympia/abuse/migrations/0028_cinderdecision_backfill.py
eviljeff: it'd also delete any `CinderDecision` created organically in the meantime, but if we're okay with that, then can do.

src/olympia/abuse/migrations/0027_add_cinderdecision.py
```diff
@@ -0,0 +1,63 @@
...
+import olympia.amo.models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        # migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        # ('bandwagon', '0008_alter_collection_options_alter_collection_managers_and_more'),
+        # ('addons', '0049_clear_bad_url_data'),
+        # ('ratings', '0011_remove_rating_one_review_per_user_and_more'),
```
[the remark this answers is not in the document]
eviljeff: the migration failed locallywhen I ran it, seemingly because of one of those migrations, and I forgot to go back to investigate _where_ it was failing.  I'll poke around.  Really, the 0001 of any of them should be enough - we only need the pk.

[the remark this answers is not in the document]
eviljeff: Worked fine this time so must have been an unrelated problem (or something I fixed along the way)
https://github.com/mozilla/addons-server/pull/22023/commits/d0ea08851d8dc3eb7919c27c5f8ca38c5dea5fee

PR #22026, opened 2024-03-18
Fix extract locales workflow to use correct syntax

KevinMind:
Follow up of: <TBD>

The action was incorrectly configured inspecting github.event which is "Object" instead of "github.event_name" which is a string that can match "push" on push to master.

Causes the push to always run as "dry-run"

https://github.com/mozilla/addons-server/actions/runs/8325203311/job/22778327690#step:5:3


eviljeff (approved):
r+ once CI passes

PR #22029, opened 2024-03-18
Fix extract locales split dry/wet runs to separate steps.

KevinMind:
Follow up of: https://github.com/mozilla/addons-server/pull/21966

The action was incorrectly configured inspecting github.event which is "Object" instead of "github.event_name" which is a string that can match "push" on push to master.

Causes the push to always run as "dry-run"

https://github.com/mozilla/addons-server/actions/runs/8326351811/job/22781858480

### DIFF
--- a/.github/workflows/extract-locales.yml
+++ b/.github/workflows/extract-locales.yml
@@ -29,11 +29,15 @@ jobs:
           run: |
             make update_deps
             make extract_locales
-      - name: Push Locales
+
+      - name: Push Lcoales (dry-run)
+        if: github.event_name == 'pull_request'
         shell: bash
         run: |
-          ARGS="--dry-run"
-          if [[ ${{ github.event_name }} == 'push' && ${{ github.ref }} == 'refs/head/master' ]]; then
-            ARGS=""
-          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/extract-locales.yml
```diff
@@ -29,11 +29,15 @@ jobs:
...
             make extract_locales
+
+      - name: Push Lcoales (dry-run)
+        if: github.event_name == 'pull_request'
+        shell: bash
+        run: |
+          make push_locales ARGS="--dry-run"
+
       - name: Push Locales
+        if: github.event_name == 'push'
```
eviljeff, line 40: we've lost the requirement that this be for the master branch along the way?

KevinMind: branches:
      - master

We are already filtering on the workflow level. I could introduce it also here but it would be redundant (It was there before and was redundant there as well.)

Wdyt?

eviljeff: okay, didn't check elsewhere in the file, just noticed it wasn't there any longer after your change.

PR #22032, opened 2024-03-18
Bump max number of categories to 3 & add new Accessibility category

diox:
Fixes mozilla/addons#1576

### DIFF
--- a/src/olympia/addons/management/commands/populate_accessibility_category.py
+++ b/src/olympia/addons/management/commands/populate_accessibility_category.py
@@ -1,0 +1,39 @@
+from django.core.management.base import BaseCommand
+
+import olympia.core.logger
+from olympia.addons.models import Addon
+from olympia.constants.base import ADDON_EXTENSION
+from olympia.constants.categories import CATEGORIES
+
+
+log = olympia.core.logger.getLogger('z.addons.populate_accessibility_category')
+
+guids = [
+    'addon@darkreader.org',
+    '{830f38bd-efc5-45dc-a5a6-064d9a638806}',
+    'jid1-QoFqdK4qzUfGWQ@jetpack',
+    '{ddc62400-f22d-4dd3-8b4a-05837de53c2e}',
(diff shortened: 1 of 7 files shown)


src/olympia/constants/categories.py
```diff
@@ -63,6 +63,13 @@ def __hash__(self):
 # any unused id.
 CATEGORIES = {
     ADDON_EXTENSION: {
+        'accessibility': StaticCategory(
+            id=74,
+            name=_('Accessibility'),
+            description=_(
+                'Download Firefox extensions that are helpful for accessibility'
```
diox, line 70: Need the actual description before merging.

src/olympia/addons/management/commands/populate_accessibility_category.py
```diff
@@ -0,0 +1,39 @@
...
+
+import olympia.core.logger
+from olympia.addons.models import Addon
+from olympia.constants.base import ADDON_EXTENSION
+from olympia.constants.categories import CATEGORIES
+
+
+log = olympia.core.logger.getLogger('z.addons.populate_accessibility_category')
+
+guids = [
```
KevinMind, line 11: Just a thought, but I would imagine it would be easier to make this an argument with a default value? What if we need to run it again in prod. Now another PR and another push and another sre ticket...

diox: I started going down that route, but ultimately we don't expect to need to do this more than once for a given category we're adding, since it's just to provide some base content, developers can and should tweak categories on their own outside of this specific case. And when we want to add new categories in the future, we'll need to make a change to the constants in the code anyway...

diox:
Yeah the bump to 3 is technically a separate thing but it was asked as part of the addition of the category, in order to allow developers more flexibility - we didn't want them to have to drop an existing category to add the accessibility one, and that limit has always been an arbitrary product decision anyway...

PR #22033, opened 2024-03-18
create specific activity for version flagged by cinder activty

### LINKED ISSUE #21871: Display flagged for human review + reasons for a first listed version in awaiting review
Followup for https://github.com/mozilla/addons-server/issues/21865
 
A first listed version in awaiting review was reported with flagged for human review:reported for abuse within the add-on and Escalated for abuse report, via cinder 

Expected:
in version's details these flags should be displayed 

cherrypick2

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-711)

eviljeff:
fixes #21871 by using `_no_automatic_activity_log` and manually creating the ActivityLog with a specific type ourselves.  Manually creating the ActivityLog means we can associate a single log entry with multiple versions for db efficiency too.

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -8,7 +8,7 @@ import requests
 import waffle
 
 import olympia
-from olympia import amo
+from olympia import activity, amo
 from olympia.amo.utils import (
     backup_storage_enabled,
     chunked,
@@ -456,13 +456,29 @@ class CinderAddonHandledByReviewers(CinderAddon):
             if appeal
             else NeedsHumanReview.REASON_ABUSE_ADDON_VIOLATION
         )
+        nhr_object = NeedsHumanReview(
+            version=self.version, reason=reason, is_active=True
+        )
         if self.version:
-            NeedsHumanReview.objects.get_or_create(
+            if not NeedsHumanReview.objects.filter(
                 version=self.version, reason=reason, is_active=True
-            )
+            ).exists():
+                nhr_object.save(_no_automatic_activity_log=True)
+                versions_to_log = [self.version]
+            else:
+                versions_to_log = []
         else:
-            self.addon.set_needs_human_review_on_latest_versions(
-                reason=reason, ignore_reviewed=False, unique_reason=True
+            versions_to_log = self.addon.set_needs_human_review_on_latest_versions(
+                reason=reason,
+                ignore_reviewed=False,
+                unique_reason=True,
+                skip_activity_log=True,
+            )
+        if versions_to_log:
+            activity.log_create(
+                amo.LOG.NEEDS_HUMAN_REVIEW_CINDER,
+                *versions_to_log,
+                details={'comments': nhr_object.get_reason_display()},
             )
 
     def report(self, *args, **kwargs):
--- a/src/olympia/abuse/utils.py
+++ b/src/olympia/abuse/utils.py
@@ -10,7 +10,7 @@ from django.utils.translation import gettext_lazy as _
 import waffle
 
 import olympia
-from olympia import amo
+from olympia import activity, amo
 from olympia.activity import log_create
 from olympia.addons.models import Addon
 from olympia.amo.templatetags.jinja_helpers import absolutify
@@ -285,6 +285,10 @@ class CinderActionEscalateAddon(CinderAction):
                 set(
                     self.target.versions(manager='unfiltered_for_relations')
                     .filter(version__in=reported_versions)
+                    .exclude(
+                        needshumanreview__reason=reason,
+                        needshumanreview__is_active=True,
+                    )
                     .no_transforms()
                 )
                 if reported_versions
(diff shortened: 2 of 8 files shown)


src/olympia/abuse/cinder.py
```diff
@@ -456,13 +456,29 @@ def flag_for_human_review(self, appeal=False):
             if appeal
             else NeedsHumanReview.REASON_ABUSE_ADDON_VIOLATION
         )
+        nhr_object = NeedsHumanReview(
+            version=self.version, reason=reason, is_active=True
+        )
         if self.version:
-            NeedsHumanReview.objects.get_or_create(
+            if not NeedsHumanReview.objects.filter(
```
KevinMind, line 463: Nit: This double negative kind of reads wonky to me? Doesn't [get_or_create](https://docs.djangoproject.com/en/dev/ref/models/querysets/#get-or-create) return a boolean if the record was created?

Could you do something like 

```python
versions_to_log=[]
nhr_object, was_created = NeedsHumanReview.objects.get_or_create(...)

if was_created:
   versions_to_log.apend(self.versions)

...
```

eviljeff: Is `not .... exists()` a _double_ negative?  🤔   How would you write it differently?

The PR changes it from `get_or_create` because we need to be able to pass `_no_automatic_activity_log=True` to `NeedsHumanReview.save`

src/olympia/abuse/cinder.py
```diff
@@ -456,13 +456,29 @@ def flag_for_human_review(self, appeal=False):
...
-            )
+            ).exists():
+                nhr_object.save(_no_automatic_activity_log=True)
+                versions_to_log = [self.version]
+            else:
+                versions_to_log = []
         else:
-            self.addon.set_needs_human_review_on_latest_versions(
-                reason=reason, ignore_reviewed=False, unique_reason=True
+            versions_to_log = self.addon.set_needs_human_review_on_latest_versions(
```
KevinMind, line 471: What is the point of skipping activity log here? If the activity log is implicit in the method, why not let it be handeld there?

The if versions_to_log could be handled in the existing version branch above? Seems like a smaller change.

src/olympia/reviewers/models.py
```diff
@@ -836,7 +836,7 @@ def set_on_addons_latest_signed_versions(cls, addons, reason):
         used_generator_last_iteration = None
         due_date = None
         for addon in addons:
-            if used_generator_last_iteration is not False:
+            if used_generator_last_iteration != []:
```
KevinMind, line 839: I understand that `set_needs_human_review_on_latest_versions` is supposed to return a list, but why was it checking if not false in the first place? Is this just some bug or why does it need to change now to checking for an empty array?

src/olympia/addons/models.py
```diff
@@ -708,23 +708,31 @@ def disable_all_files(cls, addons, reason):
         )
 
     def set_needs_human_review_on_latest_versions(
```
KevinMind, line 710: This method has a return type that doesn't match the signature (in my internal mental model of how set/get methods are supposed to work).

Previously this method did what it says and then returned a boolean telling you if something changed. That makes sense.

But now it sets the flag and returns all the modified versions? Might be worht thinking about.

src/olympia/abuse/cinder.py
KevinMind: I see. I guess I'm wondering why all of the acrobatics with skipping automatic activity? If the problem is that we want sometimes to save with an automatic activity and sometimes with another activity, and we already have that logic implemented on the save method, then why not make that configurable in the save method? Just seems like I had to do a lot of reading around what was going on to figure it out. If you modified the save method to accept a type of activity log, then you could just 

```python
nhr_object.save(_activity_log=amo.LOG.NEEDS_HUMAN_REVIEW_CINDER)
```

And that could be encapsulated in the _set_blah_blah method on the model, skipping the passing around this flag and then doing the same thing back up in this class.

src/olympia/abuse/utils.py
```diff
@@ -285,6 +285,10 @@ def flag_for_human_review(self):
                 set(
                     self.target.versions(manager='unfiltered_for_relations')
                     .filter(version__in=reported_versions)
+                    .exclude(
```
KevinMind, line 288: The point of this exclude is to prevent adding a NeedsHumanReview record for versions which already have an active one, right? Why do we need that additional filter now? Would it have made sense to already have it?

The changes below here seem to change specifically what type of activity log is produced, but this looks like it would change which versions get a NHR record. Am I reading that right?

src/olympia/abuse/cinder.py
eviljeff: We'd lose the second advantage of a single activity log for multiple versions (the reviewer tools is the other place that passes `_no_automatic_activity_log`, for that same reason, to be able to save a single log and link it to multiple versions)

src/olympia/abuse/utils.py
eviljeff: Yes, it should have skipped versions that already had a NeedsHumanReview flag (for the same reason) before - like `set_needs_human_review_on_latest_versions` does.

KevinMind (approved):
I think you should add in the description some detail as to why using one activity log associated with multiple versions is a desirable thing, it's not super clear and will be confusing without context.

eviljeff:
> ## Suspect Issues
> 
> This pull request was deployed and Sentry observed the following issues:
> 
>     * ‼️ **ValueError: max() arg is an empty sequence** `/api/v5/abuse/response/` [View Issue](https://mozilla.sentry.io/issues/5089687497/?referrer=github-pr-bot)
> 
> 
> Did you find this useful? React with a 👍 or 👎

Note: we should (obviously) fix this error, but if it's happening its because there are no versions to flag anyway - and nothing is emailed for an escalation - so I don't think there's any impact.

Issue #22039, opened 2024-03-19
Fix extract l10n commiting non-changed locales


KevinMind:
### Describe the problem and steps to reproduce it:
(Please include as many details as possible.)

Script to commit l10n string changes is considering a timestamp change when it should not.

### What happened?

Commits the timestamp change

### What did you expect to happen?

Should not commit if only timestamp changes

### Anything else we should know?
(Please include a link to the page, screenshots and any relevant files.)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-767)

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-767

KevinMind:
[...]

PR #22043, opened 2024-03-20
only git status on push command

KevinMind:
FIxes: #22039

### Description

This PR moves all of the git operations to the push extraction script which runs on a stable host environment where secure git ownership should be reliably known.

It also changes the way we specify the author using the `--author` flag instead of setting global config. This prevents overwriting the host committer on local environments.

### Context

The previous attempt at this fix missed the fact that there are 2 extraction commands, running in different environments.

1: extract runs in the container and has a dubious user
2. push runs on the host and should have the user who authoered the PR

There's no guarantee these are the same user and running git reset in the container could make weird things happen on the host.

It seems like running git status --porcelain on the host after running git reset --hard on the container resulted in changes detected, but then when we tried to commit, no changes were found because no files actually remained changed.

### Testing

I'm not sure why I wasn't able to encounter the issue that is now found in CI locally when running the previous PR, but by moving all of the git commands to the host, it should prevent that class of errors entirely.

We will just have to run this on master and see I think...

### DIFF
--- a/.github/workflows/extract-locales.yml
+++ b/.github/workflows/extract-locales.yml
@@ -30,7 +30,7 @@ jobs:
             make update_deps
             make extract_locales
 
-      - name: Push Lcoales (dry-run)
+      - name: Push Locales (dry-run)
         if: github.event_name == 'pull_request'
         shell: bash
         run: |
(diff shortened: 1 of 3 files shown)


diox:
Unrelated: can you fix this typo while you're at it:
```
.github/workflows/extract-locales.yml:      - name: Push Lcoales (dry-run)
```

KevinMind:
I've testing running this command locally with and without changes, and verified the committer is not overridden persistently on the host.

diox (approved):

KevinMind:
🤞 please. I want this to be over.

PR #22044, opened 2024-03-20
Upgrade base image to Debian bookworm

### LINKED ISSUE #22041: Upgrade base image from debian buster to bullseye or bookworm
Although debian buster is technically still supported as a LTS we should upgrade to bullseye or even bookworm.

Notably, official docker python images no longer include buster since https://github.com/docker-library/python/commit/5d908b0477c003712550c84ac4d133367b912f78

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-768)

diox:
Fixes #22041

### Description

This PR updates the base docker image addons-server from Debian buster to bookworm

 ### Context

Debian buster is in long-term support mode, but that will end soon. On top of this, the python docker images for buster are not longer receiving updates. We could switch to bullseye but bookworm is the current stable version and will be supported for longer

### Testing

- Run `make clean_docker` or equivalent to make sure you're rebuilding your deps
- Build the image with `make build_docker_image`
- If it builds correctly:
  - Stop your containers
  - Tag the newly built image to replace `mozilla/addons-server`
  - Restart the containers to pick it up
- Test your env by clicking around, running tests etc. 
- `make shell` followed by `cat /etc/debian_version` should return `12.5`

### DIFF
--- a/src/olympia/core/apps.py
+++ b/src/olympia/core/apps.py
@@ -1,12 +1,33 @@
 import logging
+import subprocess
 import warnings
 
 from django.apps import AppConfig
 from django.conf import settings
+from django.core.checks import Error, Tags, register
 from django.utils.translation import gettext_lazy as _
 
 
 log = logging.getLogger('z.startup')
+
+
+class CustomTags(Tags):
+    custom_setup = 'setup'
+
+
+@register(CustomTags.custom_setup)
+def uwsgi_check(app_configs, **kwargs):
+    errors = []
+    command = ['uwsgi', '--version']
+    result = subprocess.run(command, capture_output=True)
+    if result.returncode != 0:
+        errors.append(
+            Error(
+                f'{" ".join(command)} returned a non-zero value',
+                id='setup.E001',
+            )
+        )
+    return errors
 
 
 class CoreConfig(AppConfig):
--- a/src/olympia/core/apps.py
+++ b/src/olympia/core/apps.py
@@ -17,6 +17,8 @@ class CustomTags(Tags):
 
 @register(CustomTags.custom_setup)
 def uwsgi_check(app_configs, **kwargs):
+    """Custom check triggered when ./manage.py check is ran (should be done
+    as part of verifying the docker image in CI)."""
     errors = []
     command = ['uwsgi', '--version']
     result = subprocess.run(command, capture_output=True)
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,7 @@ RUN touch /addons-server-docker-container \
         # Git, because we're using git-checkout dependencies
         git \
         # Dependencies for mysql-python (from mysql apt repo, not debian)
+        pkg-config \
         mysql-client \
         libmysqlclient-dev \
         swig \
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # Read the docs/topics/development/docker.md file for more information about this Dockerfile.
 ####################################################################################################
 
-FROM python:3.10-slim-buster as base
+FROM python:3.10-slim-bookworm as base
 
 # Should change it to use ARG instead of ENV for OLYMPIA_UID/OLYMPIA_GID
 # once the jenkins server is upgraded to support docker >= v1.9.0
--- a/docker/mysql.list
+++ b/docker/mysql.list
@@ -1,1 +1,1 @@
-deb http://repo.mysql.com/apt/debian/ buster mysql-8.0
+deb http://repo.mysql.com/apt/debian/ bookworm mysql-8.0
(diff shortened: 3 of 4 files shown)


diox:
Weird, it built just fine locally...

KevinMind:
```
Exception: Can not find valid pkg-config name.
#21 7.197       Specify MYSQLCLIENT_CFLAGS and MYSQLCLIENT_LDFLAGS env vars manually
```

This was solved on Mac by changing to `-default` for the mysql packages. Might also have to manually install pkg-config? We should check but this seems like a platform issue.

diox:
I expected `pkg-config` to be installed by default, but I guess it's not... I've added it in https://github.com/mozilla/addons-server/pull/22044/commits/a27fbf96c34b2b13b980236ea357671fa3e29988. I suspect because of the platform it's somehow not needed on my machine but might here, as a fallback of some kind ?

KevinMind:
Interestingly enough, pkg-config is not installed in either buster or bookworm. I determined this by running 

```bash
container-diff analyze python:3.10-slim-buster --type=apt
```

If you change the image to `mozilla/addons-server:latest` you'll see it's installed.

Tried it for both, and did not find it. So it must be getting installed via one of the other packages we install. You've changed the mysql package target.

diox:
Since `uwsgi` is broken locally with this I wanted to look into adding a system check for it (which might prove tricky as I don't really want to launch the command... but anyway) and noticed the smoke test step in the verify docker image looks weird: why is it displaying pip/npm installs ?

diox:
We'll need to add a test to our verify docker image step that tries to run `uwsgi` (possibly celery as well), since that's currently failing locally...

`pip install -I uwsgi --no-cache` fixes it (without `--no-cache` it doesn't!) but I'm not yet sure what's going on.

diox:
Our theory is that locally, we're using dependencies from `/deps` mounted from the local volume and so, if you don't explicitly rebuild deps, you might end up with old ones (on top of having a potentially broken pip cache). We'll need to find a way to address this that doesn't completely negate the benefits of caching...

KevinMind:
Proposal:

`initialize_docker` should have a dependent target `clean_docker`

This should 
- remove ./deps/*
- prune docker layer/volume/image cache
- remove ./docker-cache/*

This means if you run initialize_docker or clean_docker you kill your cache and the next run should be totally fresh.

diox:
I agree, but that wouldn't have been enough here, I wanted to keep my database contents so I wouldn't have run `make initialize_docker`

KevinMind:
We could have the docker clean part skip the database? You could also say that the expectation is you try update_docker, if that doesn't work you can try nuking with initialize_docker.

I would actually have expected initialize_docker to run "fresh" by default.

KevinMind:
That's why I think it should just be a command you can run arbitrarily. If you run clean_docker it will blow up everything, but for now at least we can not automatically trigger it.

KevinMind:
@diox something like this? https://github.com/mozilla/addons-server/pull/22050

diox:
> @diox something like this? #22050

Yes, that sounds great, I'll try this. Meanwhile I've added a simple system check that we're able to run `uwsgi` in this PR: https://github.com/mozilla/addons-server/pull/22044/commits/146fc15f99a922fb6397020cc64d933e8c516234

I still need to file an issue for the debug-toolbar bug but it's a regression from https://github.com/mozilla/addons-server/pull/21795, not this PR, so I won't address it here.

diox:
Rebased (now contains https://github.com/mozilla/addons-server/issues/22053 debug-toolbar fix as well)

KevinMind:
@diox can you include how we should consider verifying this locally? I'm sure you have established a set of steps after going over and over yourself. How do I know this is working?

src/olympia/core/apps.py
```diff
@@ -1,14 +1,35 @@
...
 
 
 log = logging.getLogger('z.startup')
 
 
+class CustomTags(Tags):
+    custom_setup = 'setup'
+
+
+@register(CustomTags.custom_setup)
```
KevinMind, line 18: Just to clarify, this check is going to be executed when we run

```python
./manage.py check
```

Probably worth specifying that in the PR description as it is less than obvious, and I'm not 100% sure it is true even now.

diox: I added a docstring in https://github.com/mozilla/addons-server/pull/22044/commits/c8c6a3f816ffd124570395ba51d910f1b6a923f0 - the test I had added before should prove it's ran as part of `manage.py check`, which we trigger in the verify docker image GitHub Action. I've also made sure the check does fail if `uwsgi` isn't rebuilt.

KevinMind (approved):
Verified locally.

PR #22050, opened 2024-03-21
Add clean_docker command and fix local build_docker_image when run with empty cache

KevinMind:
Relates to: #21888, PR https://github.com/mozilla/addons-server/pull/22044

### Description

This PR achieves 2 aims.

1) Add a `clean_docker` command to enable pristine builds by removing local cache directories that can be "incorrectly" mounted to the build context causing unexpected issues. (see linked PR)

2) fix local docker build errors when there is no cache directory present. (first build or pristine build)

Fix local `build_docker_image` command when there is a purged cache.

Only include the `---cache-from` argument to `docker build` when the index.json for a previous docker build cache exists. if it doesn't exist and you pull as a cache it can lead to docker build errors.

### DIFF
--- a/Makefile-os
+++ b/Makefile-os
@@ -71,6 +71,13 @@ build_docker_image: create_docker_builder ## Build the docker image
 	rm -rf $(DOCKER_CACHE_DIR)
 	mv $(DOCKER_CACHE_DIR)-new $(DOCKER_CACHE_DIR)
 
+.PHONY: clean_docker
+clean_docker: ## Clean up docker containers, images, caches, volumes and local cache directories. Use with caution. To restart the app run make initialize_docker after this commandUse with caution.
+	docker compose down --rmi all --volumes
+	docker rmi $(DOCKER_TAG) || true
+	rm -rf $(DOCKER_CACHE_DIR)
+	rm -rf ./deps/**
+
 .PHONY: initialize_docker
 initialize_docker: create_env_file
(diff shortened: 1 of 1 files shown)


Makefile-os
```diff
@@ -51,21 +51,33 @@ create_docker_builder: ## Create a custom builder for buildkit to efficiently bu
...
-		--cache-from=type=local,src=$(DOCKER_CACHE_DIR),mode=max \
-		--builder=$(DOCKER_BUILDER) \
-		--label git.commit=$(DOCKER_COMMIT) \
-		.
+	DOCKER_BUILDKIT=1 docker buildx build $(DOCKER_BUILD_ARGS) .
 	rm -rf $(DOCKER_CACHE_DIR)
 	mv $(DOCKER_CACHE_DIR)-new $(DOCKER_CACHE_DIR)
 
+.PHONY: clean_docker
+clean_docker: ## Clean up docker containers, images, caches, volumes and local cache directories. Use with caution.
```
diox, line 75: This is a bit more aggressive than what I had in mind - it might be fine though... Maybe add a comment that this is meant for a full reset of your local env ? Essentially after this your only choice is to start again with `initialize_docker`

diox: Yeah that comment should be enough

diox (approved):

Makefile-os
KevinMind: Done.

PR #22054, opened 2024-03-21
Fix error on devhub pages when debug toolbar is enabled

### LINKED ISSUE #22053: Devhub fails to load with django-debug-toolbar enabled
STR:
- Load a devhub page other that the index locally with django-debug-toolbar (4.3.0) enabled

Expected result:
- Page loads successfully

Actual result:
- Page fails to load with a traceback
- 
Screenshot 2024-03-21 at 12-35-56 UndefinedError at _developers_addon_lobra-eellant-wallamst_ownership

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-773)

diox:
Normally, doing an inline if with no else in Jinja raises the special undefined error similar to what happens with missing variables etc, and that error just prints an empty string, completely ignoring it.

In django-debug-toolbar 4.3.0 however, in some specific cases (seemingly only when the inline if contains a function call, but there may be more cases), with the templates panel enabled, that causes a hard exception to be thrown instead. This changes works around that problem by simply providing an else clause.

Fixes #22053

### DIFF
--- a/src/olympia/devhub/templates/devhub/base.html
+++ b/src/olympia/devhub/templates/devhub/base.html
@@ -7,7 +7,7 @@
 {% endblock %}
 
 {% if addon %}
-  {% set default_body_class = "no-edit" if not check_addon_ownership(request.user, addon, allow_developer=True) %}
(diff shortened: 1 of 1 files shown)


KevinMind (approved):
Verified locally.

Issue #22058, opened 2024-03-21
Extract locales fails with missing git commiter info


KevinMind:
### Describe the problem and steps to reproduce it:
(Please include as many details as possible.)

Example [build](https://github.com/mozilla/addons-server/actions/runs/8373848466/job/22927805193) is failing due to 

```bash
Committer identity unknown

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: empty ident name (for <runner@fv-az801-737.jplelvhbra5e3n4vadtwxrmqqe.phxx.internal.cloudapp.net>) not allowed
make: *** [Makefile-os:27: push_locales] Error 128
```

### What happened?

the --author flag only specifies the author of the commit but not the "committer" apparently these are separate entities in a commit and both should be specified.

Git allows specifying these values globally, locally, or via the environment. GHA do not set any of them so we have to set them.

### What did you expect to happen?

Expected --author to set the committer, but that's not how it works.

### Anything else we should know?
(Please include a link to the page, screenshots and any relevant files.)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-774)

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-774

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-774

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-774

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-774

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-774

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-774

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-774

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-774

KevinMind:
[...]

PR #22059, opened 2024-03-21
Fix git commit error missing committer identity

KevinMind:
Fixes: #22058

### Description

Adds a function robot_commit that runs git commands as the robot user using environment variables.

Also cleans up some logging and exposes commands executed for better debugging.

IMPORTANT: this PR modifies how the script accepts and interprets arguments. Instead of implementing our own --dry-mode flag, i've exposed the script to accept any arguments and pass them to the git comands. git already supports --dry-mode both for commit and push so this should allow the same (and more) flexibility with less code.

### Context

The trick here is to get the commit to work in CI without potentially changing git config information if run locally.

### Testing

You have to modify some strings and commit them. Then run

```bash
make extract_locales
```

and try 

```bash
make push_locales ARGS="--dry-run"
```

To test the output.

### DIFF
--- a/.github/workflows/extract-locales.yml
+++ b/.github/workflows/extract-locales.yml
@@ -14,6 +14,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref || github.ref }}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
 
       - name: Build Docker image
         id: build
--- a/scripts/push_l10n_extraction.sh
+++ b/scripts/push_l10n_extraction.sh
@@ -9,38 +9,52 @@ set -o pipefail
 # Treat unset variables as an error an exit immediately.
 set -u
 
+set -x
+
+info() {
+  local message="$1"
+
+  echo ""
+  echo "INFO: $message"
+  echo ""
+}
+
 ROBOT_EMAIL="addons-dev-automation+github@mozilla.com"
 ROBOT_NAME="Mozilla Add-ons Robot"
+
+# Set git committer/author to the robot.
+export GIT_AUTHOR_NAME="$ROBOT_NAME"
+export GIT_AUTHOR_EMAIL="$ROBOT_EMAIL"
+export GIT_COMMITTER_NAME="$ROBOT_NAME"
+export GIT_COMMITTER_EMAIL="$ROBOT_EMAIL"
 
 DATE=$(date -u +%Y-%m-%d)
 REV=$(git rev-parse --short HEAD)
 MESSAGE="Extracted l10n messages from $DATE at $REV"
 DIFF_WITH_ONE_LINE_CHANGE="2 files changed, 2 insertions(+), 2 deletions(-)"
 
-DRY_RUN=true
-
-if [[ "${1:-}" != "--dry-run" ]]; then
-  DRY_RUN=false
-fi
-
 git_diff_stat=$(git diff --shortstat locale/templates/LC_MESSAGES)
 
-echo "git_diff_stat: $git_diff_stat"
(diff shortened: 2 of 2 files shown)


KevinMind:
Successfuly run (actually pushing) https://github.com/mozilla/addons-server/actions/runs/8376408855

.github/workflows/extract-locales.yml
```diff
@@ -14,6 +14,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
diox, line 18: Should we have that in all our workflows btw ?

scripts/push_l10n_extraction.sh
```diff
@@ -9,38 +9,50 @@ set -o pipefail
...
-  git commit --author="$ROBOT_NAME <$ROBOT_EMAIL>" -a -m "$MESSAGE"
-  git push
-fi
+info """
+GIT_AUTHOR_NAME: $GIT_AUTHOR_NAME
+GIT_AUTHOR_EMAIL: $GIT_AUTHOR_EMAIL
+GIT_COMMITTER_NAME: $GIT_COMMITTER_NAME
+GIT_COMMITTER_EMAIL: $GIT_COMMITTER_EMAIL
+
+ARGS: $@
```
diox, line 54: Can you add a comment here repeating what you said in the PR description, about how we use args to do the dry run ?

scripts/push_l10n_extraction.sh
```diff
@@ -9,38 +9,50 @@ set -o pipefail
...
+"""
+
+git commit -am "$MESSAGE" "$@"
+git push "$@"
```
diox, line 58: This is going to avoid committing because of the `dry-run`, so even without the `dry-run` the push wouldn't work, as there would be nothing to push (the commit never happened for real). The end result is the same I guess though...

diox (approved):

.github/workflows/extract-locales.yml
KevinMind: I don't think so. It's only in this one so we can commit on the branch (even dry mode requires a writeable branch) but PR checkout by default in a headless branch.

In other workflows we don't need to commit so it's fine to checkout the PR. Or is there some other reason we'd want the branch to be not headless?

scripts/push_l10n_extraction.sh
KevinMind: I think it's fine to keep it consistent. The one downside with this is the acceptable flags for commit and push overlap but not entirely so you can get some weird end states.

I don't think we will play around with this much it's mostly just the dry-mode we want, it's just easier to pass directly because it is literally the same flag and already does what we want even better than before where we just logged. dry mode on the command actually verifies it "would have" worked.

scripts/push_l10n_extraction.sh
KevinMind: Done.

.github/workflows/extract-locales.yml
diox: Sorry I was thinking bout specifying the depth, not setting it to 0 for the other workflows - but looking at the docs, the default is 1, so we don't need to do anything elsewhere, ignore me

diox (approved):
Only tested in dry-run locally (and we've been bitten by that exact problem before), but it makes sense, especially now with the way `dry-run` is now passed as a `git` argument. :shipit:

KevinMind:
> Only tested in dry-run locally (and we've been bitten by that exact problem before), but it makes sense, especially now with the way `dry-run` is now passed as a `git` argument. :shipit:

FWIW I tested by removing dry mode on a commit with text changes. It committed and pushed the changes. It also is running on a different github event so not 100% the same.

KevinMind:
Push worked https://github.com/mozilla/addons-server/commit/42c25e922617307b7aaa21f1cae535707966cce3

Issue #22060, opened 2024-03-22
Run `compress_assets` without management command (faster, faster)


KevinMind:

KevinMind:
If we used something like https://django-compressor.readthedocs.io/en/stable/ then we could simply remove the compress assets management command and totally internalize the compression process.

There are a few compatibility issues though, so would need to verify this or some other equivalent package would even work.

KevinMind:
We could also consider bundling the assets using a javascript/web bundler like webpack/rollup/vite.

Instead of managing the bundles via MINIFIED_BUNDLES in django, we could simply reference build artifacts and let the npm tool do the building. This being isolated from django could be run in a separate stage and only depend on the static folder.

diox:
I am django-compressor maintainer, and I would recommend against using it. It's not meant for big/complex websites, and is not really in a place where it can compete with modern bundlers (I suspect all django packages for assets managements suffer from the comparison with modern npm bundlers as well).

KevinMind:
TIL 👋

eviljeff:
> We could also consider bundling the assets using a javascript/web bundler like webpack/rollup/vite.
> 
> Instead of managing the bundles via MINIFIED_BUNDLES in django, we could simply reference build artifacts and let the npm tool do the building. This being isolated from django could be run in a separate stage and only depend on the static folder.

https://github.com/MrBin99/django-vite fwiw (not checked out beyond the readme)

KevinMind:
One method I had played around with but not fully verified is something like

1. replace `MINIFY_BUNDLES` dictionary with a manifest file, either yaml or json (this so we don't need django/python to read the assets
2. define a rollup/webpack/vite build based on those configured bundles (like we have now with compress_assets just replacing the tool) 
3. make sure the path of the output bundles can be inferred by the input path, via a manifest or some heuristic
4. update our jingo_minify_helpers to read the output file path via the output manifest.

This seems a lot like withat django-vite is doing, maybe? I'm fairly confident something like this means we can delete the compress_assets file and run the bundlificiationismination as a standalone stage and cache it to the depths of hell, unless you change an asset.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-775

PR #22061, opened 2024-03-22
Verify static files in docker container

KevinMind:
Relates to: #22060

### Description

This PR introduces checks for static files in our docker container. I achieve this with 3 changes:

- extracting the logic for copying static assets from node_modules into a shell script
- adding a check in the shell script to verify the assets are in the correct location in the file system
- allowing the script to execute in a --check-only mode (useful for CI) where we rely on the docker build to copy and only use the check to verify the work
- setting the NODE_PATH environment variable in the docker build so it can be sed across makefile scripts and executables.

### Context

The script expects the NODE_PATH environment variable to be set so if you run outside a docker container, you have to set it and likely you don't have the correct node_modules installed so it should ideally only be run via the make command.

### Testing 

You can run the command locally via 

```bash
make update_assets
```

to copy files and 

```bash
./scripts/copy_node_assets.sh --check-only
```

to verify the assets.


KevinMind:
This is totally unnecessary as the docker build actually fails when there are any missing but expected npm static file dependencies.

the `compress_assets` command imports specific files declared in our `MINIFY_BUNDLES` map, if any file is missing, it will raise a not found error.

PR #22072, opened 2024-03-25
Enable delay-rejections for unlisted versions

wagnerand:
Fix https://github.com/mozilla/addons/issues/1727.

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -657,12 +657,7 @@ class ReviewHelper:
             'method': self.handler.reject_multiple_versions,
             'label': 'Reject Multiple Versions',
             'minimal': True,
-            'delayable': (
-                # Either the version is listed
-                not version_is_unlisted
-                # or (unlisted and) awaiting review
-                or self.version.file.status == amo.STATUS_AWAITING_REVIEW
-            ),
+            'delayable': True,
             'multiple_versions': True,
             'details': (
                 'This will reject the selected versions. '
--- a/src/olympia/reviewers/tests/test_views.py
+++ b/src/olympia/reviewers/tests/test_views.py
@@ -5065,10 +5065,8 @@ class TestReview(ReviewBase):
         # data-value.
         assert doc('.data-toggle.review-files')[0].attrib['data-value'] == ''
         assert doc('.data-toggle.review-tested')[0].attrib['data-value'] == ''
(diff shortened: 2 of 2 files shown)


src/olympia/reviewers/utils.py
```diff
@@ -657,12 +657,7 @@ def get_actions(self):
             'method': self.handler.reject_multiple_versions,
             'label': 'Reject Multiple Versions',
             'minimal': True,
-            'delayable': (
-                # Either the version is listed
-                not version_is_unlisted
-                # or (unlisted and) awaiting review
-                or self.version.file.status == amo.STATUS_AWAITING_REVIEW
-            ),
+            'delayable': True,
```
KevinMind, line 660: The PR title says enable delayed rejections for unlisted versions, and the change seems to enable globally. I'm guessing The goal here is to enable delayed rejections for "all" versions.  We no longer care about a file status of "STATUS_AWAITING_REVIEW"?

KevinMind:
@wagnerand can you add some steps to verify this change? I would suggest you do this by adding "Testing" section to the pull request description with steps how to reproduce the change you expect to happen. In this case I would guess it is how to execute a delayed rejection. Since you are enabling something previously unavailable, you should indicate which part of the UI in reviewer tools should be available and what would you have expected before this change.

src/olympia/reviewers/utils.py
wagnerand: This change removes the condition to only allow delay-rejecting pending unlisted versions (on top of listed versions), so that all unlisted versions can be delay-rejected.

wagnerand:
> @wagnerand can you add some steps to verify this change? I would suggest you do this by adding "Testing" section to the pull request description with steps how to reproduce the change you expect to happen. In this case I would guess it is how to execute a delayed rejection. Since you are enabling something previously unavailable, you should indicate which part of the UI in reviewer tools should be available and what would you have expected before this change.

Added the STR to the issue.

KevinMind:
For the STR, I was expecting something a little bit more detailed and specific. After walking through the steps with @diox something like this emerged. We tested earlier today.

1) submit unlisted version
2) make sure it is approved
3) go to the unlisted review page for that addon reviewers/review-unlisted/<addon-id>
4) When you select the reject multiple versions (or single version)

Expect:

To see an option to immediately reject or to reject with a specific delay.

KevinMind (approved):

PR #22084, opened 2024-03-27
Improve usability of verify email page when verification is pending or timedout

KevinMind:
Fixes: mozilla/addons/issues/1734

### Description

This PR introduces usability improvements for users trying to verify their email that are not recieving the email or have not yet recieved the email.

### Context

Socketlabs uses queues to process email requests and so we cannot know if socketlabs will have sent the email within a reasonable window for the user.

If the user decides to request another email during a verification, a new code will be created and so the filter we use for socketlabs should be wider and include potentially failed emails and previous verification attempts.

### DIFF
--- a/src/olympia/devhub/templates/devhub/verify_email.html
+++ b/src/olympia/devhub/templates/devhub/verify_email.html
@@ -6,28 +6,29 @@
 
 {% block content %}
 <h1>{{ title }}</h1>
-<div id="{{ state }}">
-<div class="verify-email-text">
+<div class="verify-email" id="{{ state }}">
+<strong>Status:</strong>
 {% if state == "email_verified" %}
   {% trans %}
   Your email address is verified.
   {% endtrans %}
 {% elif state == "email_suppressed" %}
   {% trans %}
-  Please verify your email by clicking "Verify email" above.
+  Please verify your email.
   {% endtrans %}
 {% elif state == "verification_expired" %}
   {% trans %}
   Could not verify email address. The verification link has expired.
   {% endtrans %}
 {% elif state == "verification_pending" %}
-  <div class="loader"></div>
     {% trans %}
-    We are sending an email to you, this might take a minute. The page will automatically refresh.
+    We are sending an email to you, please be patient...
+    <div class="loader"></div>
+    The page will automatically refresh.
     {% endtrans %}
 {% elif state == "verification_timedout" %}
   {% trans %}
-  It is taking longer than expected to confirm delivery of your verification email. Please try again.
+  It is taking longer than expected to confirm delivery of your verification email. Please try again later.
   {% endtrans %}
 {% elif state == "confirmation_pending" %}
   {% trans email=request.user.email %}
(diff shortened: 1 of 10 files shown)


KevinMind:
@diox @eviljeff I've added STR for this PR in the issue. I'd like to continue our discussion about STRs using this as an example of "waaaayyyy tooo much/detailed" information. This is the archetype and I think we can work backwords to a middle ground between the current state and this that is feasible to include on a regular basis.

src/olympia/devhub/templates/devhub/verify_email.html
```diff
@@ -36,57 +37,74 @@ <h1>{{ title }}</h1>
...
+            {{ email.statusDate }}
+          </td>
+          <td>
+            {{ email.from }}
+          </td>
+          <td>
+            {{ email.to }}
+          </td>
+          <td>
+            {{ email.subject }}
```
diox, line 76: Should we truncate that for potential display issues with long subjects ?

diox (commented):
Still playing around with the timeout/refresh locally, but some things I found in the code.

src/olympia/users/models.py
```diff
@@ -1385,7 +1385,7 @@ def is_expired(self):
 
     @property
     def is_timedout(self):
-        return self.created + timedelta(seconds=30) < datetime.now()
+        return self.created + timedelta(minutes=1) < datetime.now()
```
diox, line 1388: Per our discussion, let's bump to 10 minutes. We know from testing it can take around 7 minutes for the email to appear in the list, even if it has been sent faster than that.

static/js/zamboni/devhub.js
```diff
@@ -6,7 +6,7 @@ $(document).ready(function () {
   $('#edit-addon').exists(initEditAddon);
 
   // Poll for suppressed email removal updates.
-  $('#verification_pending').exists(function () {
+  $('.verify-email#verification_pending').exists(function () {
     setTimeout(function () {
       window.location.reload();
     }, 10_000);
```
diox, line 12: Let's bump to 30 seconds to go with the adjusted timeout suggested above

src/olympia/lib/settings_base.py
```diff
@@ -1552,3 +1552,5 @@ def read_only_mode(env):
 SOCKET_LABS_HOST = env('SOCKET_LABS_HOST', default='https://api.socketlabs.com/v2/')
 SOCKET_LABS_TOKEN = env('SOCKET_LABS_TOKEN', default=None)
 SOCKET_LABS_SERVER_ID = env('SOCKET_LABS_SERVER_ID', default=None)
+
+SUPPORT_EMAIL = 'amo-admins@mozilla.org'
```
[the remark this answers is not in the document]
KevinMind: No, good call out though. @wagnerand wdyt?

src/olympia/devhub/templates/devhub/verify_email.html
KevinMind: I was going to say it doesn't matter since we control the subject, but now we are displaying all emails so yeah maybe it's a good point. I can add a css truncation.

static/js/zamboni/devhub.js
KevinMind: Done.

src/olympia/users/models.py
KevinMind: Done.

src/olympia/lib/settings_base.py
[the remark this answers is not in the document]
KevinMind: I've added +amo_devhub which would allow us to associated all emails sent via devhub. wdyt?

src/olympia/devhub/templates/devhub/verify_email.html
KevinMind: Done.

diox (approved):

PR #22085, opened 2024-03-27
AMO-238

KevinMind:
Relates to: AMO-238

### Descripton

This PR is the first iteration of migrating our CircleCI jobs to github action workflows, specifically the "assets" job.

### Context

If we model our CI workflows falling into 4 phases of the CI/CD process, the phase of "test" is where most of our circleci jobs and the existing "verify-docker-image" workflow fall.

### Testing

I will break the assets job, and we should expect both circleci and github action to fail for the same reason.

### DIFF
--- a/.github/actions/run-docker/action.yml
+++ b/.github/actions/run-docker/action.yml
@@ -22,6 +22,7 @@ runs:
         cat <<EOF > exec.sh
         #!/bin/bash
         whoami
+        set -x
         ${{ inputs.run }}
         EOF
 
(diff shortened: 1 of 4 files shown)


KevinMind:
Files created in the docker build still have 9500 owner, when we run fix_olympia_user we modify the `olympia` user and group but not files owner by the 9500 user/group

https://github.com/mozilla/addons-server/actions/runs/8453429444/job/23156072525?pr=22085

PR #22095, opened 2024-03-29
Expose to Cinder whether the add-on is part of any promoted group, not just badged ones

diox:
This replaces the old `promoted_badge` field, introducing a new `promoted` instead since "Notable" promoted add-ons don't get a badge, re-using the old field would be misleading.

Cinder has a custom renderer that shows only "Promoted badge" in the main area of the UI (OLD in the screenshot below), that will need to be updated to use the new field. In the meantime, the new field is shown in the side area on the left of the UI without special treatment (NEW in the screenshot below).

Screenshot 2024-03-29 at 11-52-43 Queues Cinder

Fixes https://github.com/mozilla/addons/issues/1738


src/olympia/abuse/cinder.py
```diff
@@ -290,6 +291,11 @@ def get_attributes(self):
             'name': self.get_str(self.addon.name),
             'slug': self.addon.slug,
             'summary': self.get_str(self.addon.summary),
+            'promoted': self.get_str(
+                promoted_group.name if promoted_group != NOT_PROMOTED else ''
```
eviljeff, line 295: `NOT_PROMOTED` is falsey so we can do
```suggestion
                promoted_group.name if promoted_group else ''
```
(and drop the import above)

eviljeff (approved):
r+

If this needs a renderer change in cinder remember to file a feature request via lindie

src/olympia/abuse/cinder.py
diox: https://github.com/mozilla/addons-server/pull/22095/commits/c9270b04f989f34ff7fe68149e6e7c5aaaea5fc7

Issue #22093, opened 2024-03-28
Remove `docs` job from CircleCI now that we have a GitHub Action for that


diox:
As part of https://github.com/mozilla/addons-server/issues/21932 we added a GitHub action to build docs. That means the `docs` job in CircleCI is entirely obsolete now and we should remove it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-789)

```[tasklist]
### Tasks
- [ ] remove circleCI docs job
- [ ] Migrate docs github action workflow to docker container build
- [ ] Remove readthedocs configurations
```

KevinMind:
We should also migrate the github action docs build job to the docker container version. I have a commit for that I can open a PR.

diox:
> Finally, we should remove the configs for readthedocs from our repository.

I guess it's fine since we're no longer automatically triggering builds via webhook, but do we need to keep some configuration in place for the redirects to work ?

PR #22101, opened 2024-04-02
Build docs in docker container

KevinMind:
Relates to: #22093 

### Description

Migrates the new github action workflow for building our docs site to use the docker container build.

### Context

We added the github docs action workflow before the docker container build action was available. Now that it is, we can migrate the workflow and remove the setup-environment action so all our CI will operate in the same context.

### Testing

CI should continue passing on PRs, no implication for merging because we are only publishing the artifact then.


diox:
> ... (now failing from the npm check)

What do you mean by that?

KevinMind:
> > ... (now failing from the npm check)
> 
> What do you mean by that?

I discovered this bug while adding a check in our package.json that would fail if running npm install outside of our docker container.

diox (approved):

PR #22102, opened 2024-04-02
TMP: test the node_modules named volume approach

KevinMind:
Relates to: ADDSRV-720

### Description

TMP: testing out using a docker managed volume for node_modules directory to resolve the docker/host permissions issue.

### Context

This should allow us to simplify our npm setup, using the default directory, while avoiding the host/container permissions issue by makin the node_modules exist in a docker managed volume.

### Testing

TBD: just testing out now.


KevinMind:
Note: This [build](https://github.com/mozilla/addons-server/actions/runs/8520545411/job/23336886197?pr=22102) is 
1) not failing when a test fails, this is a problem, needs to be fixed.
2) the reason it is failing is because the files `static/js/zamboni/stats-all.js` is created during the docker build, but does not exist in the host when we run the container. Docker does not merge bind mounts from the host with the container file system, but replaces the path so if the file doesn't exist in the host file system when we run, it is effectively "removed" in the container. 

This needs to be addressed, there are a few potential solutions:
- make sure built files are placed in an ignored directory on the container. like "node_modules" is in the docker ignore, so it is safe for the container to write files there, the host will not overwrite them. We could put our build assets in a ./dist folder or something and then ignore that directory from the host.
- we could make sure to re-build assets when running the contianer (inefficient, boo)
- we could have a smarter mount volume when running, only mounting what and when we need to in CI.

KevinMind:
I actually got the build to be 37 seconds. That's redonkulous https://github.com/mozilla/addons-server/actions/runs/8572559945/job/23495758215

KevinMind:
Finally 🟢 and 🚤 🔥 fast

Old time: 9 minutes 51 seconds
New time: 6 minutes 31 seconds

Percent: 151% faster
Absolute: 3 minutes 20 seconds Runtime saved.

src/olympia/core/sentry.py
```diff
@@ -29,21 +29,11 @@ def get_sentry_release():
...
-                head = head[5:]
-                revision_file = os.path.join(root, '.git', *head.split('/'))
-            else:
-                return head
-            with open(revision_file) as fh:
-                version = str(fh.read()).strip()
-        except OSError:
-            version = None
+    # sentry is loaded before django so we have to read the env directly
+    ensure_version = os.environ.get('REQUIRE_SENTRY_VERSION', False)
```
KevinMind, line 33: We will set this to true in production environments.

KevinMind:
First commit:

- verified that version.json is created in docker build on [CircleCI ](https://app.circleci.com/pipelines/github/mozilla/addons-server/30786/workflows/2be75a7a-249c-4d21-a0aa-db2d726e542d/jobs/207250/parallel-runs/0/steps/0-102)
- verified that version.json exists in local container via the repo volume
- all tests are passing

KevinMind:
Commit 2: 
- verified local `make build_docker_image` successful
- all tests passing
- circle/github docker build successful

PR #22103, opened 2024-04-02
fix test_escalate_addon_no_versions_to_flag

eviljeff:
https://github.com/mozilla/addons-server/pull/22052 broke a test because of a bad-ish merge (CinderAction classes changed to be created with a CinderDecision rather than a CinderJob instance, but in the meantime a different PR added `test_escalate_addon_no_versions_to_flag` that still used a CinderJob)

- https://github.com/mozilla/addons-server/commit/e2e90aed8c9123fb637b2e78d6b9ea4a07f70f37 added test_escalate_addon_no_versions_to_flag
- https://github.com/mozilla/addons-server/pull/22052 changed the classes from CinderJob to CinderDecision
- https://github.com/mozilla/addons-server/commit/c1b35673acb4860d4e1e25d0950cca1a9fe373cf was the merge commit of #22052

### DIFF
--- a/src/olympia/abuse/tests/test_utils.py
+++ b/src/olympia/abuse/tests/test_utils.py
@@ -4,7 +4,7 @@ from django.urls import reverse
 
 from waffle.testutils import override_switch
 
-from olympia import amoth
+from olympia import amo
 from olympia.activity.models import ActivityLog
 from olympia.amo.tests import (
     TestCase,
--- a/src/olympia/abuse/tests/test_utils.py
+++ b/src/olympia/abuse/tests/test_utils.py
@@ -4,7 +4,7 @@ from django.urls import reverse
 
 from waffle.testutils import override_switch
 
-from olympia import amo
+from olympia import amoth
 from olympia.activity.models import ActivityLog
 from olympia.amo.tests import (
     TestCase,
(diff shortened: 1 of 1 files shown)


KevinMind:
Could you link the 3 relevant commits in the PR description? The one introducing the test, the one updating the class and the merge commit. Would be nice to see all three diffs when looking at this.

eviljeff:
> Could you link the 3 relevant commits in the PR description? The one introducing the test, the one updating the class and the merge commit. Would be nice to see all three diffs when looking at this.

added to the description - and reworded because I had the order that things happened the opposite way around.

KevinMind (approved):
Main job is now passing. Just wondering how this could have been avoided and could be avoided in the future.

I guess rebasing the PR 22052 would have added the commit from 22075 and the test would have failed, right?

I think we can setup a check on github to ensure that PRs are rebased before merging. @diox do we have linear history required currently?

I'm fairly confident a merge queue would've prevented this as the queue would have added the PRs commit onto the latest master and re-ran CI checks but outside of that I wonder if there is a way to check for this class of bugs?

eviljeff:
> Main job is now passing. Just wondering how this could have been avoided and could be avoided in the future.
> 
> I guess rebasing the PR 22052 would have added the commit from 22075 and the test would have failed, right?

It would have, but we don't typically rebase after every commit unless Github highlights a merge conflict.  

> I think we can setup a check on github to ensure that PRs are rebased before merging. @diox do we have linear history required currently?

We only allow squash&merge commits.  Requiring branches to _always_ be up to date with master would be a significant additional burden with a repo that gets as many commits per day as addons-server (it's had 9 in the last 2 hours as I write this) - in the time the CI runs, on average, another commit will have been merged!

> I'm fairly confident a merge queue would've prevented this as the queue would have added the PRs commit onto the latest master and re-ran CI checks but outside of that I wonder if there is a way to check for this class of bugs?

If it doesn't require manual work, and only alerts when something conflicts, then maybe?  Something to be discussed outside of this PR.

PR #22109, opened 2024-04-02
Update task that resigns add-ons without a COSE signature

### LINKED ISSUE #21996: Update task to resign add-ons without a COSE signature as part of SHA-1 deprecation
To prepare for Firefox deprecating (phase 1) and eventually (phase 2, not in scope for now) removing support for old PKCS#1 SHA-1 signatures in add-ons, we need to resign the limited number of add-ons that don't have a COSE signature so that they don't break.

We already have a task for this, `sign_addons`, that can be triggered with the `process_addons --task=resign_addons_for_cose` management command. This task was written years ago and never executed as the deprecation was pushed back.

We need to check that the existing code still makes sense and make a number of ajustements to this task, or at least double-check if it behaves like we want, notably regarding:
- [x] Which add-ons it's considering. At the moment it's processing all publicly available add-ons on AMO whose latest public listed version was created before April 5th 2019
  - [x] Do we want to deal with deleted add-ons ? Add-ons without a publicly available listed version currently ?
- [x] Which versions it's considering. At the moment it's re-signing and bumping the latest public listed version of each add-on it's processing.
- [x] How it bumps versions. At the moment it's doing something a bit unexpected, bumping the version number _on the existing version_ and overwriting the xpi. We likely don't want this, and want to create entirely new versions instead.
- [x] Email copy (needs to be rewritten entirely)
- [x] ~~Bonus: add a one-liner release notes to the version to explain what it is~~ (too annoying with locales)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-754)

diox:
### Context

This is the main task we are going to use to resign add-ons without a COSE signature, to prepare for SHA-1 deprecation.

When ran, it will:
- Pick up publicly available add-ons for which the latest listed version (`current_version` internally) was created before the threshold date, then for each of them:
  - Clone this version into an entirely new one with a new version number considered higher than the existing one
  - Approve and sign that cloned version, replacing the existing one as the latest public listed version for each add-on
  - Notify affected developers

### Testing

The unit tests should cover a bunch of different scenarios. If you want to test it manually:

- Create some add-ons and versions depending on what you want to test. The script should only care about the `current_version` of public, non-invisible add-ons (`disabled_by_user=False`, `status=amo.STATUS_APPROVED`), if that version `created` date is from before 2019-04-05 and the add-on is not a dictionary.
- Run `manage.py --task=bump_and_resign_addons` (this will return and leave the celery workers to do the work when the tasks have been created, check your celery logs to see when it's finished)
- For add-ons matching the criteria described above, they should get a new `current_version` replacing the previous one. It should have a new version number higher than the one it's replacing (and with a `resigned1` suffix to distinguish from version numbers used by the developers), including in the manifest, and be signed.
- If there are developers/owners attached to the affected add-ons they should receive an email (locally, a `FakeEmail`).

Fixes #21996

### DIFF
--- a/src/olympia/addons/management/commands/process_addons.py
+++ b/src/olympia/addons/management/commands/process_addons.py
@@ -83,7 +83,6 @@ class Command(ProcessObjectsCommand):
                         disabled_by_user=False,
                         type__in=(
                             amo.ADDON_EXTENSION,
-                            amo.ADDON_LPAPP,
                             amo.ADDON_STATICTHEME,
                         ),
                     )
--- a/src/olympia/addons/management/commands/process_addons.py
+++ b/src/olympia/addons/management/commands/process_addons.py
@@ -20,7 +20,7 @@ from olympia.constants.base import (
     _ADDON_WEBAPP,
 )
 from olympia.devhub.tasks import get_preview_sizes, recreate_previews
-from olympia.lib.crypto.tasks import sign_addons
+from olympia.lib.crypto.tasks import bump_and_resign_addons
 from olympia.ratings.tasks import addon_rating_aggregates
 from olympia.reviewers.tasks import recalculate_post_review_weight
 from olympia.versions.tasks import delete_list_theme_previews
@@ -72,8 +72,8 @@ class Command(ProcessObjectsCommand):
                 'task': recalculate_post_review_weight,
                 'queryset_filters': get_recalc_needed_filters(),
             },
-            'resign_addons_for_cose': {
-                'task': sign_addons,
+            'bump_and_resign_addons': {
+                'task': bump_and_resign_addons,
                 'queryset_filters': [
                     # Only resign public add-ons where the latest version has been
                     # created before the 5th of April
--- a/src/olympia/addons/management/commands/process_addons.py
+++ b/src/olympia/addons/management/commands/process_addons.py
@@ -80,6 +80,12 @@ class Command(ProcessObjectsCommand):
                     Q(
                         status=amo.STATUS_APPROVED,
                         _current_version__created__lt=datetime(2019, 4, 5),
+                        disabled_by_user=False,
+                        type__in=(
+                            amo.ADDON_EXTENSION,
+                            amo.ADDON_LPAPP,
+                            amo.ADDON_STATICTHEME,
+                        ),
                     )
                 ],
             },
--- a/src/olympia/lib/crypto/tasks.py
+++ b/src/olympia/lib/crypto/tasks.py
@@ -20,38 +20,24 @@ from olympia.versions.models import Version
 log = olympia.core.logger.getLogger('z.task')
 
 
-MAIL_COSE_SUBJECT = (
-    'Your Firefox extension has been re-signed with a stronger signature'
-)
+MAIL_COSE_SUBJECT = 'Your Firefox add-on has been re-signed'
 
 MAIL_COSE_MESSAGE = """
-Hello,
+Greetings from the Firefox Add-ons team!
 
-Mozilla has recently upgraded the signing [1] for Firefox extensions, themes,
-dictionaries, and langpacks to provide a stronger signature. All add-on
-versions uploaded to addons.mozilla.org after April 5, 2019 have this
-signature. We plan to stop accepting the old signature with Firefox 70 [2].
+Per our previous communication, this email is to confirm that the most recent
+publicly available version of your add-on has now been re-signed with a
+stronger signature for a more secure add-ons ecosystem. It will remain
+backwards compatible with previous versions of Firefox.
 
-The current version of your add-on, {addon}, listed on addons.mozilla.org has
-been automatically re-signed with the stronger signature. Your add-on will
-remain backwards compatible with previous versions of Firefox, including ESR 68
-[3], and will continue working when your users upgrade to Firefox 70.
+Please be aware that to re-sign add-ons automatically we had to clone the
+latest public version of your add-on, bump the version number and then re-sign,
+so don’t be alarmed if you see that your add-on’s version number has increased.
 
-You do not need to take any action at this time.
+Please feel free to reply to this email if you have any questions.
 
 Regards,
-
-The Add-ons Team
-
----
-[1] https://extensionworkshop.com/documentation/publish/signing-and-distribution-overview/
-[2] https://wiki.mozilla.org/Release_Management/Calendar
-[3] https://www.mozilla.org/firefox/enterprise/
---
-
-You have received this email because you are a registered developer of a
-Firefox add-on. If you do not want to receive these updates regarding your
-add-on, please sign in to addons.mozilla.org and delete your add-on(s).
+Firefox Add-ons team
 """  # noqa: E501
 
 
--- a/src/olympia/lib/crypto/tasks.py
+++ b/src/olympia/lib/crypto/tasks.py
@@ -1,5 +1,6 @@
+import json
 import os
-import shutil
+import zipfile
 
 import olympia.core.logger
 from olympia import amo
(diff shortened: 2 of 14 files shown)


src/olympia/lib/crypto/tasks.py
```diff
@@ -50,20 +50,18 @@
...
+    # Parse existing version number, increment the last part, and replace
+    # the suffix with "resigned1", in order to get a new version number that
+    # would force existing users to upgrade.
+    # Example: '1.0' would return '1.1resigned1'.
+    vs = VersionString(version)
+    parts = vs.vparts
+    parts[-1].a += 1
+    parts[-1].b = 'resigned'
+    parts[-1].c = 1
+    return str(VersionString.from_vparts(parts))
```
diox, line 64: @eviljeff you're the resident `VersionCompare` expert. Does this (and the couple new methods I had to add to `VersionCompare` to make that work) makes sense to you?

eviljeff: It _should_ produce a valid increased version string, but `vparts` isn't guaranteed to store all the same information that is in the underlying version string - it's there only for comparison between two strings.  I.e. there is no certainty that `VersionString.from_vparts(VersionString('foo').vparts) == VersionString('foo')` so the new string after it's been adjusted may not "look like" the existing string.  Maybe that's fine, though.

eviljeff: (btw you don't need to cast `VersionString` back to a `str` - it subclasses `str`)

src/olympia/versions/compare.py
```diff
@@ -136,10 +136,19 @@ def asdict(self):
         def __repr__(self):
             return f'{self.asdict()}'
 
+        def __str__(self):
+            return f'{self.a}{self.b}{self.c}{self.d}'
```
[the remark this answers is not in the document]
eviljeff:
```suggestion
            return f'{part.a}{part.b}{(part.c or "")}{part.d}'
```

src/olympia/lib/crypto/tasks.py
eviljeff: mmm, I guess it would be cleaner to live in `get_new_version_string`.  It wouldn't need a function imo, just something like:

```suggestion
    vs = VersionString(version)
    parts = vs.vparts
    parts[-1].a += 1
    parts[-1].b = 'resigned'
    parts[-1].c = 1
    return VersionString('.'.join(str(part) for part in parts)
```

(with the `__str__` suggestion below)

src/olympia/versions/compare.py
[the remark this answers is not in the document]
diox: Actually... I don't need `__str__` at all ? I'm never going to convert back to `str` now... With your suggestion above, we're rebuilding a new `VersionString` from our changes, not updating `parts`.

src/olympia/promoted/models.py
```diff
@@ -96,26 +96,11 @@ def approve_for_version(self, version):
...
-        The current version will be signed for approval, and if there's special
-        signing needed for that group the version will be resigned."""
-        from olympia.lib.crypto.tasks import sign_addons
+        The current version will be signed for approval."""
 
         if not self.addon.current_version:
             return
         self.approve_for_version(self.addon.current_version)
-        if self.group.autograph_signing_states:
-            sign_addons([self.addon.id], send_emails=False)
```
diox, line 107: This was - thankfully - dead code. The only place outside tests that calls `approve_for_addon()` is in `save()` (a few lines below if you expand the diff). It only calls it if the promoted group has `immediate_approval`, and in here we checked if the group has some `autograph_signing_states`.

There is no promoted group that has both of these set. We have groups for which the promotion is immediate (`Spotlight`, because add-ons are added to that group after review anyway) and groups that cause the signature to be special (`Recommended`/`Line` and the obsolete `Verified`/`Sponsored`) but we never combine the two - we decided long ago that to "gain" a promotion with a special signature you need a new version to be explicitly uploaded, we never want to resign existing ones.

src/olympia/promoted/models.py
```diff
@@ -96,26 +96,11 @@ def approve_for_version(self, version):
...
-    def get_resigned_version_number(self):
-        """Returns what the new version number would be if approved_for_addon
-        was called.  If no version would be signed return None."""
-        from olympia.lib.crypto.tasks import get_new_version_number
-
-        version = self.addon.current_version
-        if version and not version.is_unreviewed:
-            return get_new_version_number(version.version)
-        else:
-            return None
```
diox, line 118: Only existed to test the dead code mentioned above.

src/olympia/lib/crypto/tests/test_signing.py
```diff
@@ -602,21 +602,7 @@ def test_sign_bump_non_ascii_filename(self, mock_sign_file):
...
         assert mock_sign_file.called
         self.version.reload()
-        assert self.version.version == '0.0.1.1-signed'
-        self.file_.reload()  # Otherwise self.file_.file doesn't get re-opened
-        assert file_hash != self.file_.generate_hash()
-        self.assert_backup()
-
-    @mock.patch('olympia.lib.crypto.tasks.sign_file')
-    def test_sign_bump_non_ascii_version(self, mock_sign_file):
-        """Sign versions which have non-ascii version numbers."""
```
diox, line 612: We don't have such versions in the database. Version numbers have to be ascii by definition, this is enforced by both [addons-server](https://github.com/mozilla/addons-server/blob/9585c60ec916ad9db58d9c9943ed042123c48e7c/src/olympia/files/utils.py#L46) and addons-linter.

src/olympia/addons/management/commands/process_addons.py
```diff
@@ -72,14 +72,20 @@ def get_recalc_needed_filters():
...
                 'queryset_filters': [
                     # Only resign public add-ons where the latest version has been
                     # created before the 5th of April
                     Q(
                         status=amo.STATUS_APPROVED,
                         _current_version__created__lt=datetime(2019, 4, 5),
+                        disabled_by_user=False,
+                        type__in=(
+                            amo.ADDON_EXTENSION,
+                            amo.ADDON_LPAPP,
```
diox, line 86: FIXME: We don't need langpacks

eviljeff: Aren't they signed too?

diox: https://github.com/mozilla/addons-server/pull/22109/commits/c60c833ee3cd1b543ea8f7c2cc9c141c1d54fff4 - we realized during our meeting about the comms for this that we didn't need (and really shouldn't!) bump langpacks, they are each meant for a specific version of Firefox from years ago that will still accept SHA-1 signatures...

eviljeff (approved):
r+wc nothing blocking merging

src/olympia/lib/crypto/tasks.py
```diff
@@ -1,159 +1,218 @@
...
-The current version of your add-on, {addon}, listed on addons.mozilla.org has
-been automatically re-signed with the stronger signature. Your add-on will
-remain backwards compatible with previous versions of Firefox, including ESR 68
-[3], and will continue working when your users upgrade to Firefox 70.
+Please be aware that to re-sign add-ons automatically we had to clone the
+latest public version of your add-on, bump the version number and then re-sign,
+so don’t be alarmed if you see that your add-on’s version number has increased.
 
-You do not need to take any action at this time.
+Please feel free to reply to this email if you have any questions.
```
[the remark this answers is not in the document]
diox: I don't think supporting replies would help, cause I'm not sure reviewers are the best persons to raise concerns to

diox:
> Actually, do we need the `type__` at all? Everything is signed these days, right?

- Langpacks we definitely don't want to bump. If somehow there are langpacks affected we shouldn't touch them, they are meant for a specific (very old at this point) Firefox version that should still support SHA-1.
- Dictionaries we want to avoid purely because they don't need to be signed. We could include them, we might need to care about the future, but for now we decided to exclude them to reduce the scope of things we need to worry about.

eviljeff:
>     * Dictionaries we want to avoid purely because they don't need to be signed. We could include them, we might need to care about the future, but for now we decided to exclude them to reduce the scope of things we need to worry about.

My concern would be existing dictionaries, which are signed already with a SHA-1 signature.  Firefox may not _require_ dictionaries to be signed, but if they are, and the signing is broken because its SHA-1, does it still accept them?

PR #22134, opened 2024-04-11
Update address we use to send non-activity emails

### LINKED ISSUE #22127: AMO should use a sender of its own domain for outbound emails

Outbound emails from AMO sometimes appear to be sent from domains other than its own. For instance, some outbound emails are observed as originating from `nobody@mozilla.org`. Meanwhile, AMO uses a different email provider than `mozilla.org` does for sending emails. This arrangement works because the DMARC policy of `mozilla.org` is relatively permissive. Presently, the DMARC policy of mozilla.org is as follows:

```
$ dig +short _dmarc.mozilla.org TXT
"v=DMARC1; p=none; adkim=r; aspf=r; rua=mailto:dmarc_agg@vali.email,mailto:dmarc@mozilla.com"
```

This situation may lead to email reception issues in the long term.

I propose that we explore the possibility of sending emails using the appropriate domain for each environment. For example, in the staging environment, we could use `nobody@addons.allizom.org` as the sender, while in production, we would use `nobody@addons.mozilla.org`.

This approach would enable us to configure distinct DMARC policies for each environment, without being dependent on the DMARC policies established by our parent domain.

As for rollout, we should proceed env by env, ensuring that email sending functions correctly for one env before moving on to the next.

diox:
Fixes #22127

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -746,7 +746,7 @@ LOGOUT_REDIRECT_URL = '/'
 MAX_GEN_USERNAME_TRIES = 50
 
 # Email settings
-ADDONS_EMAIL = 'Mozilla Add-ons <nobody@mozilla.org>'
+ADDONS_EMAIL = 'Mozilla Add-ons <nobody@addons.mozilla.org>'
 DEFAULT_FROM_EMAIL = ADDONS_EMAIL
 
 # Email goes to the console by default.  s/console/smtp/ for regular delivery


src/olympia/lib/settings_base.py
```diff
@@ -746,7 +746,7 @@ def get_db_config(environ_var, atomic_requests=True):
 MAX_GEN_USERNAME_TRIES = 50
 
 # Email settings
-ADDONS_EMAIL = 'Mozilla Add-ons <nobody@mozilla.org>'
+ADDONS_EMAIL = 'Mozilla Add-ons <nobody@addons.mozilla.org>'
```
diox, line 749: FIXME: We actually want a per-env value

diox:
Parking this for now.

Issue #11991, opened 2019-07-31
Fix `make debug` with uwsgi


EnTeQuAk:
Follow-up of #6668, where I broke `make debug`

Either find a good solution for remote ipdb or use uwsgi's `--honour-stdin` option, which does sound promising.

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-14

PR #22135, opened 2024-04-11
Set ipdb as the breakpoint callback

diox:
[`breakpoint()`](https://docs.python.org/3/library/functions.html#breakpoint) is a built-in python function that drops a debugger where you put it in the code, like `debugger;` in JavaScript, which is very useful to quickly debug an issue, since from there you can navigate up and down the stack, inspect locals etc. While our current local environments setup doesn't allow this to work with uwsgi/celery, because there is no terminal to attach to, it works great in tests.

This PR makes python use `ipdb` instead of `pdb` as the debugger callback for `breakpoint()`, which is a lot nicer: it has tab-completion, syntax highlighting, etc. We already have `ipdb` in our requirements because of how useful it is, this makes it easier to call (no longer have to drop `import ipdb; ipdb.set_trace()` in the code, you can just use `breakpoint()`)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,7 @@ x-env-mapping: &env
     - OLYMPIA_SITE_URL=http://olympia.test
     - PYTHONDONTWRITEBYTECODE=1
     - PYTHONUNBUFFERED=1
+    - PYTHONBREAKPOINT=ipdb.set_trace
     - TERM=xterm-256color
     - CIRCLECI=${CIRCLECI}
     - HISTFILE=/data/olympia/docker/artifacts/bash_history


KevinMind:
Could you add some information in the PR or even better in the docs of how you would use this?

diox:
Reworded the PR description.

KevinMind:
`because there is no terminal to attach to` should we consider just giving olympia a terminal?

KevinMind (approved):
I didn't test this, I assume you have.

diox:
The issue to make debugging work locally is https://github.com/mozilla/addons-server/issues/11991 - I haven't looked at it in years, I remember briefly looking at it and it was tricky. I usually debug more in tests than with an actual server so it's never been a priority.

PR #22151, opened 2024-04-16
show special flag for nhr from abuse reports in review queue

### LINKED ISSUE #22145: Distinguish add-ons flagged for abuse/jobs from ones that are queued for another reason
fulfils https://mozilla-hub.atlassian.net/browse/AMO-224

When an add-on has an abuse report (or appeal) outstanding that AMO reviewers need to review we create a NeedsHumanReview instance (commonly called “flagging” it for human review).  When a NeedsHumanReview instance is created it assigns a “due date” for the add-on, which causes it to show up in the manual review queue in the reviewer tools.  We store the reason a NeedsHumanReview is created.  

The ask from operations is that the manual review queue indicate with icons on the row what the reason of the NeedsHumanReview object is.  There will be separate icons for each reason: appeals, abuse reports, and jobs forwarded from cinder.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-797)

eviljeff:
fixes #22145 - the first commit is where all the actual fix is - the second is just refactoring.


wagnerand:
Please do not merge for the moment, as we discuss the intended scope of this PR. Thank you.

eviljeff:
scope clarified in issue and upstream jira story, and further commit added to separate icons

src/olympia/addons/models.py
```diff
@@ -357,13 +360,35 @@ def get_queryset_for_pending_queues(
                     & ~Q(**{listed_delay_flag_field: datetime.max})
                 )
             )
+        needs_human_review_kw = {
```
diox, line 363: I can't easily submit this as a suggestion because it touches lines outside of your patch, but what do you think of this:

```diff
diff --git a/src/olympia/addons/models.py b/src/olympia/addons/models.py
index 0667e5fd37..4524fe57df 100644
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -321,7 +321,7 @@ class AddonManager(ManagerBase):
             select_related_fields_for_listed=False,
         )
         versions_due_qs = (
-            Version.unfiltered.filter(due_date__isnull=False)
+            Version.unfiltered.filter(due_date__isnull=False, addon=OuterRef('pk'))
             .no_transforms()
             .order_by('due_date')
         )
@@ -360,32 +360,26 @@ class AddonManager(ManagerBase):
                     & ~Q(**{listed_delay_flag_field: datetime.max})
                 )
             )
-        needs_human_review_kw = {
-            'addon_id': OuterRef('pk'),
-            'needshumanreview__is_active': True,
-        }
         qs = (
             qs.filter(**filters)
             .annotate(
                 first_version_due_date=Min('versions__due_date'),
-                first_version_id=Subquery(
-                    versions_due_qs.filter(addon=OuterRef('pk')).values('pk')[:1]
-                ),
+                first_version_id=Subquery(versions_due_qs.values('pk')[:1]),
                 needs_human_review_from_cinder=Exists(
                     versions_due_qs.filter(
-                        **needs_human_review_kw,
+                        needshumanreview__is_active=True,
                         needshumanreview__reason=NeedsHumanReview.REASONS.CINDER_ESCALATION,
                     )
                 ),
                 needs_human_review_from_abuse=Exists(
                     versions_due_qs.filter(
-                        **needs_human_review_kw,
+                        needshumanreview__is_active=True,
                         needshumanreview__reason=NeedsHumanReview.REASONS.ABUSE_ADDON_VIOLATION,
                     )
                 ),
                 needs_human_review_from_appeal=Exists(
                     versions_due_qs.filter(
-                        **needs_human_review_kw,
+                        needshumanreview__is_active=True,
                         needshumanreview__reason=NeedsHumanReview.REASONS.ABUSE_ADDON_VIOLATION_APPEAL,
                     )
                 ),

```

It avoids repeating the `OuterRef` and removes the need for the weird dict (unfortunately we can't simplify this further and build a `needs_human_review_qs` ahead of time since the `reason` need to live in the same `filter()` to be in the same `JOIN`).

diox:
The extra subqueries it forces us to make look kinda bad but seem pretty efficient - I've tested this in production and it performs well, no speed difference at least.

src/olympia/addons/models.py
eviljeff: >  we can't simplify this further and build a needs_human_review_qs ahead of time since the reason need to live in the same filter() to be in the same JOIN

that answers my question as to why my attempts at optimizing by having a qs object that had all the common kwargs failed 🥲

eviljeff:
> The extra subqueries it forces us to make look kinda bad but seem pretty efficient - I've tested this in production and it performs well, no speed difference at least.

yeah, the `exists` seem to be pretty optimal.  I considered getting a single list of all the NHR values for all the versions and working it out in python but I feared for the edge case of add-ons with a very large number of versions it'd perform badly.

diox (approved):

PR #22152, opened 2024-04-17
Update appeal flag description in reviewer tools

wagnerand:
This is just changing an inaccurate string displayed to add-on reviewers as the reason for when an add-on is enqueued due to a developer appeal.

### DIFF
--- a/src/olympia/reviewers/models.py
+++ b/src/olympia/reviewers/models.py
@@ -802,7 +802,7 @@ class NeedsHumanReview(ModelBase):
             ),
             (
                 REASON_ABUSE_ADDON_VIOLATION_APPEAL,
-                'Appeal against a reviewer decision about a policy violation',
+                "Appeal of a reviewer's decision about a policy violation",
             ),
         ),
         editable=False,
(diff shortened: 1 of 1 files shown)


src/olympia/reviewers/models.py
```diff
@@ -802,7 +802,7 @@ class NeedsHumanReview(ModelBase):
             ),
             (
                 REASON_ABUSE_ADDON_VIOLATION_APPEAL,
-                'Appeal about a decision on abuse reported within the add-on',
+                'Appeal against a reviewer decision about a policy violation',
```
KevinMind, line 805: nit: "Appeal _of_ a reviewer's decision about a policy violation" sounds a bit more neutral.

KevinMind (approved):
Missing an issue but LGTM.

src/olympia/reviewers/models.py
wagnerand: Ah, thanks, will update. I had looked it up on dict.leo.org and one example there mentioned _against_.

PR #22158, opened 2024-04-18
Update due date after updating file status during force_disable

### LINKED ISSUE #22156: Some versions still have a due date even when the add-on was force-disabled
We noticed [on prod](https://reviewers.addons.mozilla.org/en-US/reviewers/review-unlisted/2834759) that a force-disabled add-on still had a version with a due date. While we don't know for sure, our assumption is that it is related to the `Unlisted Auto-approval delayed indefinitely` flag and the ordering of actions when an add-on with that flag gets force-disabled.

According to @diox:

My guess is that it comes down to ordering:
* The add-on has the Unlisted Auto-approval delayed indefinitely flag
* It gets force disabled, which:
  1. Removes all existing NHR
  2. Updates all due dates (at this point the version is still pending review and the flag still present. The version keeps its due date)
  3. Disables all versions

### What happened?
This causes the due date to get updated before the versions are disabled, while they are still in a state that allows a due date.

The new due date for the versions should be calculated **after** they are in the right state (i.e. disabled)
 
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-800)

wagnerand:
Fix #22156.

### DIFF
--- a/src/olympia/addons/models.py
+++ b/src/olympia/addons/models.py
@@ -633,14 +633,14 @@ class Addon(OnChangeMixin, ModelBase):
             'Addon "%s" status force-changed to: %s', self.slug, amo.STATUS_DISABLED
         )
         self.update(status=amo.STATUS_DISABLED)
+        # https://github.com/mozilla/addons-server/issues/13194
+        Addon.disable_all_files([self], File.STATUS_DISABLED_REASONS.ADDON_DISABLE)
         self.update_version()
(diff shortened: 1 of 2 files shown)


wagnerand:
Requesting review from Mat because he had already looked into the issue and we discussed it over slack.

I've become a little rusty, please give the test an extra look. Thanks.

diox (approved):

PR #22169, opened 2024-04-19
Don't require 2FA for theme developers versions

diox:
This is achieved by overriding `theme_specific` in `_submit_upload()` if an add-on is passed (or if we're using the theme wizard). That variable is passed down to the upload flow (and checked against the xpi being uploaded later) to bypass the 2FA requirement.

### Context

We enforce 2FA at submission time if the `2fa-enforcement-for-developers-and-special-users` flag is active for the user, but we want that for extensions (and langpacks/dicts, technically), not themes. The bug was that we accidentally required it for all new versions of existing add-ons, even themes.

### Testing

Automated tests should cover this but if you want to try locally you can follow the STR in the issue. You can use the theme wizard to create the initial theme, and then take the zip and bump the version number. 

Note that you need to create the waffle flag (not a switch! it can be potentially active for a limited set of users) with: `manage.py waffle_flag 2fa-enforcement-for-developers-and-special-users --everyone`. Then, when logging in with the fake auth, check or uncheck the "two-factor authentication" checkbox before logging in depending on what you want to test.

Fixes https://github.com/mozilla/addons/issues/1956

### DIFF
--- a/src/olympia/constants/base.py
+++ b/src/olympia/constants/base.py
@@ -134,7 +134,7 @@ ADDON_TYPE = {
     _ADDON_LPADDON: _('Deprecated Language Pack (Add-on)'),
     _ADDON_PLUGIN: _('Deprecated Plugin'),
     _ADDON_PERSONA: _('Deprecated LWT'),
-    ADDON_STATICTHEME: _('Theme (Static)'),
+    ADDON_STATICTHEME: _('Theme'),
     _ADDON_SITE_PERMISSION: _('Deprecated Site Permission'),
 }
 
@@ -148,7 +148,7 @@ ADDON_TYPES = {
     _ADDON_LPADDON: _('Deprecated Language Packs (Add-on)'),
     _ADDON_PLUGIN: _('Deprecated Plugins'),
     _ADDON_PERSONA: _('Deprecated LWTs'),
-    ADDON_STATICTHEME: _('Themes (Static)'),
+    ADDON_STATICTHEME: _('Themes'),
     _ADDON_SITE_PERMISSION: _('Deprecated Site Permissions'),
 }
 
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -1467,6 +1467,12 @@ def _submit_upload(
     form = forms.NewUploadForm(
         request.POST or None, request.FILES or None, addon=addon, request=request
     )
+    if wizard or (addon and addon.type == amo.ADDON_STATICTHEME):
+        # If using the wizard or submitting a new version of a theme, we can
+        # force theme_specific to be True. If somehow the developer is not
+        # uploading a theme, validation will reject it just like if they had
+        # tried to use the theme submission flow for an entirely new add-on.
+        theme_specific = True
     form.fields['theme_specific'].initial = theme_specific
     channel_text = amo.CHANNEL_CHOICES_API[channel]
     if request.method == 'POST' and form.is_valid():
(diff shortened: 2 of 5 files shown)


src/olympia/files/utils.py
```diff
@@ -1043,7 +1043,13 @@ def parse_addon(
             msg = gettext(
                 'The type (%s) does not match the type of your add-on on AMO (%s)'
             )
-            raise forms.ValidationError(msg % (parsed['type'], addon.type))
```
diox, line 1046: This is a drive-by fix for this fairly obscure/rare error message which used to return the constant instead of the "pretty" name. See corresponding test fix. 

Note with 2FA enforcement enabled, it can be difficult to trigger anyway: if your user doesn't have 2FA, and you're trying to upload an extension as a version of an existing theme, you'd get the `Please add two-factor authentication to your account to submit extensions` error message instead, as devhub would detect you're using a theme specific flow to upload an extension. It's still relevant if doing the opposite (uploading a theme as a version of an extension) or when using other types though.

src/olympia/constants/base.py
```diff
@@ -134,7 +134,7 @@
     _ADDON_LPADDON: _('Deprecated Language Pack (Add-on)'),
     _ADDON_PLUGIN: _('Deprecated Plugin'),
     _ADDON_PERSONA: _('Deprecated LWT'),
-    ADDON_STATICTHEME: _('Theme (Static)'),
+    ADDON_STATICTHEME: _('Theme'),
```
eviljeff, line 137: These are also unrelated changes too, though coming way past time!

src/olympia/devhub/views.py
```diff
@@ -1467,6 +1467,12 @@ def _submit_upload(
     form = forms.NewUploadForm(
         request.POST or None, request.FILES or None, addon=addon, request=request
     )
+    if wizard or (addon and addon.type == amo.ADDON_STATICTHEME):
+        # If using the wizard or submitting a new version of a theme, we can
+        # force theme_specific to be True. If somehow the developer is not
+        # uploading a theme, validation will reject it just like if they had
+        # tried to use the theme submission flow for an entirely new add-on.
+        theme_specific = True
```
eviljeff, line 1475: submissions via the wizard already set `theme_specific` kwarg, right?  The patch is just covering the case for existing themes having new versions submitted using the xpi upload flow?

eviljeff (commented):
Is the case of using the non-wizard upload flow to submit a _new_ theme covered already?  (Or is that impossible to handle because we don't know at the point we'd need to enforce 2FA that they're submitting a theme xpi)

src/olympia/constants/base.py
diox: Yes, goes with the drive-by fix mentioned above, mainly to avoid saying `'The type (Extension) does not match the type of your add-on on AMO (Theme (Static))` for instance, which looks weird.

src/olympia/devhub/views.py
diox: Yes. I figured it didn't hurt to make sure `theme_specific` is set even if using the wizard.

diox:
> Is the case of using the non-wizard upload flow to submit a new theme covered already? (Or is that impossible to handle because we don't know at the point we'd need to enforce 2FA that they're submitting a theme xpi)

It's in fact the only new case we're covering in this PR. We know the existing add-on is a theme so we set `theme_specific` in https://github.com/mozilla/addons-server/pull/22169/files/7d0fb5dd599c8697f04648066b5a3a752218ced6#diff-da2e9cbdc57a9d4adb11cb41193b9e44b26cbf1e702d61c3bc0b2775897f1a20. As mentioned in my other comment the wizard was technically already working because the wizard view passes `theme_specific=True` kwarg already.

eviljeff (approved):

PR #22175, opened 2024-04-22
Update Reply-To in SHA1 resign script and send mails separately

diox:
Operations asked to change the Reply-To.

The emails were grouped by add-on, which would potentially leak the emails of other owners of the add-on when they are more than one. That information is already available in the authors page for the add-on in devhub but since the email is not encrypted it's best to avoid that problem entirely by sending separate emails.

Another follow-up for #21996

### Context and Testing

You can find context and testing instructions for the whole script in the original PR, https://github.com/mozilla/addons-server/pull/22109 below but you probably don't need to test manually. The changes to the unit tests should cover the changes to the reply to and the split between owners of the same add-on.

### DIFF
--- a/src/olympia/lib/crypto/tasks.py
+++ b/src/olympia/lib/crypto/tasks.py
@@ -256,6 +256,6 @@ def bump_addon_version(old_version):
         amo.utils.send_mail(
             subject,
             message,
-            reply_to=['firefox-android-addon-support@mozilla.com'],
+            reply_to=['mozilla-add-ons-community@mozilla.com'],
             recipient_list=[email],
         )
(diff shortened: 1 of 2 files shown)


src/olympia/lib/crypto/tasks.py
```diff
@@ -250,12 +250,12 @@ def bump_addon_version(old_version):
...
-    emails = qs.values_list('user__email', flat=True)
     subject = mail_subject.format(addon=addon.name)
     message = mail_message.format(addon=addon.name)
-    amo.utils.send_mail(
-        subject,
-        message,
-        recipient_list=emails,
-        headers={'Reply-To': 'amo-admins@mozilla.com'},
-    )
+    for email in qs.values_list('user__email', flat=True).order_by('user__pk'):
```
KevinMind, line 255: why order by pk? just curious. doesn't seem relevant.

PR #22176, opened 2024-04-22
Dont give incomplete listed themes versions a due date

diox:
Fixes mozilla/addons#1762


src/olympia/versions/models.py
```diff
@@ -185,7 +184,8 @@ def should_have_due_date(self, negate=False):
                 addon__promotedaddon__group_id__in=(
                     g.id for g in PROMOTED_GROUPS if g.listed_pre_review
                 )
-            ),
+            )
+            | Q(addon__type__in=amo.GROUP_TYPE_THEME),
             addon__status__in=(amo.VALID_ADDON_STATUSES),
```
diox, line 189: `VALID_ADDON_STATUS` contains only `(STATUS_NOMINATED, STATUS_APPROVED)` so that's what makes this work.

src/olympia/versions/tests/test_models.py
```diff
@@ -969,6 +969,47 @@ def test_should_have_due_date_listed(self):
         )
         assert version.should_have_due_date
 
+    def test_should_have_due_date_listed_theme(self):
+        addon = Addon.objects.get(id=3615)
```
KevinMind, line 973: I've seen different patterns in our tests, but wouldn't creating an addon in a factory make this test more atomic and therefore less flaky? Specifically:

- susceptible to test order failures
- susceptible to failure caused by changes in other tests

KevinMind: Reading the tests I think I see another reason to use a factory pattern. 80% of the test is just setting up a version in the right state, this is done through modification of the model in the test. Is it not possible to create the model in the state you want to test against?

Couldn't you create a factory to create exactly that state and then you just pass a few args and assert against the returned model?

KevinMind:
How is 

> remove the due date for the themes in that state that already have one.

Addressed in this PR?

src/olympia/versions/models.py
KevinMind: What does "this" refer to?

KevinMind (approved):
Approving with nits. I didn't verify myself but since we are aiming to ship this in a cherry pick anyway, probably faster to have QA test on dev to speed up the process.

src/olympia/versions/tests/test_models.py
diox: Yes. We have such a factory - `addon_factory()`. We also have a bunch of old tests using fixed fixtures, and that infamous `3615` add-on is the most common of them. Typically when adding tests I try to copy the style of the tests nearby and that's why I went with re-using `3615` here, but I agree it's not ideal. I'll change it for the tests I've added here.

diox:
>> remove the due date for the themes in that state that already have one.
> Addressed in this PR?

Good shout-out: it's not addressed in this PR, we ran a one-off SQL query (that I tested on dev) manually on Monday because we figured it would be simpler & faster to ship. I'll update the issue.

src/olympia/versions/tests/test_models.py
diox: Tests refactored to use `addon_factory()` in https://github.com/mozilla/addons-server/pull/22176/commits/b1add2b705c29223b0fbcf41806c32d184986dd6

PR #22186, opened 2024-04-25
Generate version.json as part of docker build:

KevinMind:
Relates to: https://github.com/mozilla/addons-server/pull/22182

This commit can be merged independently of the others. Let's make the PRs smaller

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -357,20 +357,15 @@ commands:
         type: string
         default: "latest"
     steps:
-      - run: >
-          printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n'
-          "$CIRCLE_SHA1"
-          "$CIRCLE_TAG"
-          "$CIRCLE_PROJECT_USERNAME"
-          "$CIRCLE_PROJECT_REPONAME"
-          "$CIRCLE_BUILD_URL"
-          > version.json
       - run:
           name: Build docker image and push to repo
           command: |
             docker version
             docker login -u "${DOCKERHUB_USER}" -p "${DOCKERHUB_PASS}"
-            make build_docker_image DOCKER_TAG="app:build" DOCKER_COMMIT="$CIRCLE_SHA1"
+            make build_docker_image \
+              DOCKER_TAG="app:build" \
+              DOCKER_VERSION=<< parameters.image_tag >> \
+              DOCKER_COMMIT="$CIRCLE_SHA1"
             docker images
       - when:
           condition: << parameters.push >>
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -498,6 +493,10 @@ commands:
             echo export NPM_CONFIG_PREFIX=/deps/ >> $BASH_ENV
             echo export CC=\"`python -c 'import sysconfig; print(sysconfig.get_config_var("CC"))'`\" >> $BASH_ENV
             cat $BASH_ENV
+      - run:
+          name: Set version
+          command: |
+            make -f Makefile-os version DOCKER_VERSION="$CIRCLE_BRANCH" DOCKER_COMMIT="$CIRCLE_SHA1"
```
diox, line 499: Here and below in the `image_tag` (and in the Github Action) I wonder if we should use something more unique and also safer (since branches can contain any characters). Maybe `build-${CIRCLE_BUILD_NUM}` or something like that instead ? And try to keep the GitHub Action consistent

KevinMind: Can you describe how it should be unique? Like you cannot push the same tag from circle and GHA? Or unique across builds of the same branch? And how safe? What is the thing we need to be safe from?

The GHA and the CircleCI job are consistent in that if you run the build-docker-image job it will tag the branch name, and in circleCI it does the same.

For PRs github uses special detached head branches, AFAIK circleci doesn't really have this. I also think it's somewhat unimportant to have 1:1 parity as the GHA will replace CircleCI soon enough.

Can you clarify so I know more about what specific changes you'd like to see?

.circleci/config.yml
```diff
@@ -620,7 +619,7 @@ jobs:
       - checkout
       - make_release:
           # use invalid tag
```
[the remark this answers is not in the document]
KevinMind: Done.

.circleci/config.yml
diox: Regarding uniqueness: contributors re-using a branch name already used for some reason. Including `master` if coming from a fork (not sure if the `CIRCLE_BRANCH` contains the fork name)

Regarding safety: git branch names can contain... any character. So I was worried about potential shell escape issues.

KevinMind: secrets don't pass to forks so they cannot push to our docker registry no matter the branch name. But it is a vulnerability for us. I added a `ci-` prefix in the GHA build. Branch builds in Circle never push so I just re-hardcoded the TAG.

KevinMind: We also can, and probably should forbid branches "latest" "master" "main" etc.. no reason to allow them to exist. https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule

KevinMind: https://github.com/mozilla/addons-server/actions/runs/8848824075/job/24299408303

.circleci/config.yml
```diff
@@ -651,7 +653,7 @@ workflows:
       - devhub
       - main
       # Uncomment if you want to test the docker build
```
diox, line 655: I assume you are you going to re-comment the job before merging ?

diox (approved):

.circleci/config.yml
KevinMind: I actually would rather leave it for now. The next few PRs will benefit from comparison against Circle build, but happy to re=comment and un-comment in each PR if you prefer.

KevinMind: Re-commented, will just enable when needed.

PR #22192, opened 2024-04-26
absolutify devhub urls in the dsa emails

### LINKED ISSUE #1764: HTML tags visible on Editor Tools landing page
## steps to reproduce:
1.  Load Editor tools
2. Approve an add-on awaiting review
3. Notice the section where your  "All Time" points are listed.

## observed behavior:
HTML tags visible on Editor Tools landing page

## screenshot: 
https://www.dropbox.com/s/gepb0hcz5tw6qdm/Screenshot%202016-02-24%2013.30.09.png?dl=0

eviljeff:
fixes mozilla/addons#1764 - I forgot to absolutify the url

STR:
- reject/approve an unlisted version in the reviewer tools
  - if you don't have an unlisted version locally you'll need to upload one (api or devhub)
- look at the email and see the developer hub url in the email is an absolute one.
  - emails can be viewed in django admin FakeEmail or a django shell

### DIFF
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -11,6 +11,7 @@ from django.utils.functional import cached_property
 from olympia import amo
 from olympia.addons.models import Addon
 from olympia.amo.models import BaseQuerySet, ManagerBase, ModelBase
+from olympia.amo.templatetags.jinja_helpers import absolutify
 from olympia.api.utils import APIChoicesWithNone
 from olympia.bandwagon.models import Collection
 from olympia.constants.abuse import APPEAL_EXPIRATION_DAYS, DECISION_ACTIONS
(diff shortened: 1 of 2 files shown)


src/olympia/abuse/models.py
```diff
@@ -1044,7 +1045,11 @@ def notify_reviewer_decision(
         )
         # override target_url if this decision related to unlisted versions
         target_url_override = (
-            {'target_url': reverse('devhub.addons.versions', args=[self.target.id])}
+            {
```
KevinMind, line 1048: What is the use case for not absolutifying a URL? Like when would we want just /banana/foo ? Is there some benefit to that sometime?

KevinMind (approved):
Approving with a question that is non blocking for merge.

src/olympia/abuse/models.py
eviljeff: In emails, there isn't a use-case.  In page templates we normally don't want absolute urls.

Issue #21225, opened 2023-09-26
Store webext version/UA and "source" for each Version


diox:
Follow-up for #18803:

For each `Version` we create, let's store:

* Whether it was created from devhub, signing API, new API, automation (we already record this on the `FileUpload` but currently don't keep it afterwards)
* The `web-ext` version if applicable

We'll want that table to be exposed in redash afterwards for easy querying.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/ADDSRV-502)

diox:
untested prototype for when we have time to pick this up again https://github.com/mozilla/addons-server/pull/21233/files

ioanarusiczki:
With dev hub 

https://addons-internal.dev.mozaws.net/en-US/admin/models/versions/version/1801150/change/

With API - using 2 different clients 

https://addons-internal.dev.mozaws.net/en-US/admin/models/versions/version/1801153/change/
https://addons-internal.dev.mozaws.net/en-US/admin/models/versions/version/1801152/change/

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDSRV-502

PR #22199, opened 2024-04-29
Add basic readonly admin for VersionProvenance

diox:
\+ some minor tweaks around the changelist for Version which we never really cared for, add refactoring of associated tests

Supports #21225 

### Context/Testing

This is all to help us/QA play with/test `VersionProvenance` from that same issue. It can be tested locally with a superuser (need at least a `mozilla.com` user with `Admin:Advanced` permission) going to `/en-US/admin/models/versions/` and noticing the differences (if you have uploaded a `Version` after the changes from a3b6d38804e8994c23b09d5d147d311b87b94ca8 you'll see the version provenance info in the detail admin page for that `Version`).


src/olympia/versions/tests/test_admin.py
```diff
@@ -5,47 +5,32 @@
...
-
-    def test_unauthorized_user_has_no_access(self):
-        addon = addon_factory()
-        version = addon.current_version
-        detail_url = reverse('admin:versions_version_change', args=(version.pk,))
-        user = user_factory(email='someone@mozilla.com')
-        self.client.force_login(user)
-        response = self.client.get(detail_url, follow=True)
-        assert response.status_code == 403
+    def setUp(self):
```
KevinMind, line 15: NIT: more readable to split the refactoring here to a separate commit so one can distinguish the refactor from the addition. No change needed just mentioning.

KevinMind:
> It can be tested locally with a superuser (need at least a mozilla.com user with Admin:Advanced permission)

Does it mean the normal superuser created during `make initialize` will work or does that user not per default have Admin:Advanced permissions.. If not, how to add those permissions.

> if you have uploaded a Version

Do you "need" to upload a new version after running this branch or what are the concrete steps? Should I be able to see something useful without uploading or that is a true pre-requisite?

diox:
> Does it mean the normal superuser created during make initialize will work or does that user not per default have Admin:Advanced permissions.. If not, how to add those permissions.

It will work - superusers have `*:*` which is everything. It's not typically used outside of local dev though. If you want to test with the actual permission, you need to create a `Group` (in a django shell or admin) with `Admin:Advanced` as part of its `rules` (separated by commas, if adding multiple) and then adding your user to that group.

> if you have uploaded a Version
> Do you "need" to upload a new version after running this branch or what are the concrete steps? Should I be able to see something useful without uploading or that is a true pre-requisite?

You wouldn't see the `VersionProvenance` for old versions as we didn't record that before https://github.com/mozilla/addons-server/commit/a3b6d38804e8994c23b09d5d147d311b87b94ca8, so that wouldn't be very useful (though I suppose you can test that it doesn't fail with a `Version` for which we don't have a `VersionProvenance` - the existing admin tests should do that though)

KevinMind (approved):
Code looks good. I haven't tested.

PR #22205, opened 2024-04-29
Add "build" to version.json

KevinMind:
Fixes: https://github.com/mozilla/addons/issues/1785

### Description

Adds "build" prop to version.json with a test to verify it exists in the container on CI.

### Context

we don't rely on this property locally, but our deployment jobs expect this value to contain a link to the job which produced the image to be deployed

### Testing

You can build the container locally

```bash
make build_docker_image
```

and then inspect the file

```bash
docker run --rm -it --platform linux/amd64 <image> cat version.json
```

You can run the test locally by copying the bash script from the verify workflow. This essentially creates a new version.json with a different file, mounts it, and compares the container and host versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -500,7 +500,8 @@ commands:
               # We can use a hardcoded tag here because we're not pushing
               # an image but just testing the build
               DOCKER_VERSION="branch" \
-              DOCKER_COMMIT="$CIRCLE_SHA1"
+              DOCKER_COMMIT="$CIRCLE_SHA1" \
+              VERSION_BUILD_URL="$CIRCLE_BUILD_URL"
       - run:
             name: Install Python and Node dependencies
             command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -365,7 +365,8 @@ commands:
             make build_docker_image \
               DOCKER_TAG="app:build" \
               DOCKER_VERSION=<< parameters.image_tag >> \
-              DOCKER_COMMIT="$CIRCLE_SHA1"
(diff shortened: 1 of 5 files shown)


.github/workflows/verify-docker-image.yml
```diff
@@ -26,3 +26,26 @@ jobs:
...
+
+          # Extract the version JSON from the image
+          touch version-container.json
+          docker run \
+            -v $(pwd)/version-container.json:/data/olympia/version-container.json \
+            --rm -u 0 ${{ steps.build.outputs.tags }} \
+            cp version.json version-container.json
+
+          # expect them to be the same
+          if [[ "$(cat version.json)" != "$(cat version-container.json)" ]]; then
```
diox, line 46: FYI `cmp --quiet file1 file2` would report 0 or 1 depending on whether the files are (exactly) the same or differ. It's probably fine to `cat` them though.

diox (approved):

.circleci/config.yml
```diff
@@ -653,7 +655,7 @@ workflows:
       - devhub
       - main
       # Uncomment if you want to test the docker build
-      # - build-image
+      - build-image
```
diox, line 658: Need to re-flip that before merging

.github/workflows/verify-docker-image.yml
KevinMind: TIL. bash has everything.

.circleci/config.yml
KevinMind: Done.

PR #22206, opened 2024-04-30
Use consistent and robust mapping of UID between host and container: 

KevinMind:
Partially fulfills: https://github.com/mozilla/addons-server/pull/22182

### Description

Use consistent and robust mapping of UID between host and container:

- collapse UID/GID to one value on host and container
- rename host value to HOST_UID for clarity and debugability
- rename fix_olympia_user to entrypoint and set as docker compose entrypoint
- uwsgi config to use runtime value of olympia UID/GID

### Context

This change might not seem very relevant right now. That's okay, it is an enabling change.

We are allowing our docker compose config to opt-in to setting the olympia user id on startup. This enables:

- running the same flow on the container with or without the host ID (useful in scenarios where you want to run with container ID .e.g produciton or CI)
- run the fix olympia user script as an entrypoint on the container so it does not need to be specified manually allows consistent execution for better reproducability because the script is always run the same way.

### Testing

Test locally

```bash
make clean_docker
make initialize_docker
```

Smoke test, everything works.

Check the ID

```bash
make shell
# in the shell
id -u # expect your host ID matching the HOST_UID var in .env
id -g # expect 9500 it is not changed by the script anymore as user permissions are sufficient
```

Try a number of different combinations, running this command

```bash
docker compose up web --force-recreate
```

HOST_UID= -> `Running command as olympia 9500:9500`
HOST_UID=0 -> `usermod: UID '0' already exists`
HOST_UID=20 -> `olympia UID: 9500 -> 20` (would have failed previously as GID is used by a group in the container)
HOST_UID=65534 -> `usermod: UID '65534' already exists`

If the ID is taken, you will get a failing container, if it isn't taken or is already set to 9500 or not set, the container will run.


KevinMind:
@diox in this PR we are only modifying the UID. GID is basically ignored. We should only need to assign user permissions.

https://github.com/mozilla/addons-server/pull/22206/files#diff-c9edf2a20913b0e79fc7d0a0786c49c813375c6173dd7396c951ff584cd6665fR10

diox:
You've removed the `groupmod -g ${GID} olympia` and stopped passing the `GID` so the `olympia` user no longer has the right group id. Any files it creates will potentially have an incorrect group id and be inconsistent with other files on the host.

KevinMind:
> You've removed the `groupmod -g ${GID} olympia` and stopped passing the `GID` so the `olympia` user no longer has the right group id. Any files addons-server creates will potentially have an incorrect group id and be inconsistent with other files on the host.

Fun fact, the behavior has been to set both UID/GID to the value of UID in the shell when running docker compose up for some time.

https://github.com/mozilla/addons-server/blame/808c55b243e164aff8d7f0fe5acf4fc5baa764e4/docker-compose.yml#L26

we are specifying both values in the docker compose right now on master, if you inspect the GID of olympia you will see it is the UID and not the GID.

Now, what to do about it... we could keep it how I have it as it doesn't seem to cause any issues. The user level file ownership seems to be sufficient to prevent any issues in the local container and there are no implications for production either way.

The alternative is to fix this issue and use the actual GID specified on the host. The problem with that is, you end up with "group with id <X> already exists". So you actually have to add some error handling to that to make it really work.

I have added a commit doing this, if you think it is worth the extra code, we can keep it... I recommend we keep the first commit as it is much mnuch simpler and more explicit about what is currently happening.

diox:
Interesting, wrong assumption on my part, thanks for digging into this. On my own machine (and most Linux default installs I suppose) the `UID` and `GID` are the same so I never noticed. I think it's better to ensure it's at least set to the same value as the `UID`, or better yet since you've written the code to fix it, use the proper value from the host. That way we're consistent and not at risk of creating weird edge-cases with the files...

KevinMind:
> I think it's better to ensure it's at least set to the same value as the UID

That's equivalent of the first commit

> use the proper value from the host.

That is equivalent of the second commit, if you meant the proper value being the value of the GID on the host.


The only issue I see with the current approach is it will fail in cases where your GID or UID already exist on the container... I'm not really sure how to fix that. I've not seen it happen on the UID side, but it is affecting me currently with GID 20 on my machine. There's no reason this "couldn't" happen for UID as well, I just have never seen it.

But merging this will break dev on my machine. I think the alternatives are:
- only use the UID (if we run into the issue with UID then we have to reconsider the whole approach anyway but right now not an issue)
- modify the behavior so that if the GID already exists, we either:
  - don't change it at all
  - change it to the value of the UID

I don't really like either variant of the second option because it is just yet more complexity, it was working with UID only so I would just keep it that way and "fix" it by making it explicit.

docker-compose.yml
```diff
@@ -19,45 +19,37 @@ x-env-mapping: &env
...
     - HISTCONTROL=erasedups
-    # Note: docker compose uses the values exported from .env for GID/UID if
-    # they exist. ./docker/fix_olympia_user.sh uses those variables to fix
-    # the uid/gid of the user to match the host if necessary.
-    - UID=${UID:-9500}
-    - GID=${UID:-9500}
+    # Note: docker compose uses the values exported from .env for HOST_UID if
+    # it exists. ./docker/entrypoint.sh uses this variable to fix
+    # the uid of the olympia user to match the host user id if necessary.
+    - HOST_UID=${HOST_UID:-}
```
KevinMind, line 25: It is actually "somewhat" better IMO to use a unique name here for a few reasons:
- debugging the value set during docker build and runtime is easier, no collision
- understanding which is which is also easier
- it is clear when setting the .env file that you are tracking the host UID and not the olympia user id.

docker-compose.yml
```diff
@@ -19,45 +19,37 @@ x-env-mapping: &env
...
-        supervisord -n -c /data/olympia/docker/supervisor-celery.conf
+      - supervisord -n -c /data/olympia/docker/supervisor-celery.conf
     volumes:
       - .:/data/olympia
       - ./deps:/deps
       - ./package.json:/deps/package.json
       - ./package-lock.json:/deps/package-lock.json
     extra_hosts:
      - "olympia.test:127.0.0.1"
+    restart: on-failure:5
```
KevinMind, line 46: Makes compose fail the container if the process exits non 0 more than 5 times.

KevinMind:
Working GID/UID fix, ish https://github.com/mozilla/addons-server/pull/22206/commits/778b220a3fcebaa704bf0e5ae673f9bf6d7eab0b

Makefile-os
```diff
@@ -101,7 +99,7 @@ initialize_docker: create_env_file build_docker_image
 # would be uid 9500 regardless of host at this point (this is only fixed when
 # the container is up, through the command defined in docker-compose.yml),
 # which is wrong for local development.
-	docker compose run --rm --user ${UID}:${GID} web make update_deps
+	docker compose run --rm web make update_deps
```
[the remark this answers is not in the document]
diox: The comment above explains what this isn't true: the script is not an actual entrypoint, just a default command we run on local envs. It's bypassed when using `docker compose run` with a custom command like we are here. So that would cause `make initialize_docker` to run that initial `make update_deps` with the wrong UID/GID...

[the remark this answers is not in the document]
KevinMind: It is in fact an entrypoint now. https://github.com/mozilla/addons-server/pull/22206/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R36

I guess the comment is now redundant or outdated at least..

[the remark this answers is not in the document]
diox: Ah, I missed that. Yeah the comment needs to be updated then. I'll test locally now that the other PR has been approved and that one rebased.

Makefile-os
```diff
@@ -46,7 +43,7 @@ rootshell: ## connect to a running addons-server docker shell with root user
 
 .PHONY: create_env_file
 create_env_file:
-	echo "UID=${UID}\nGID=${GID}" > .env
+	echo "HOST_UID=${HOST_UID}" > .env
```
diox, line 46: Is there a drawback to leaving `UID` in there, to keep compatibility with older `docker-compose.yml` ? I'm thinking it could help if we go back to an earlier commit to try something out (maybe as part of a `bisect`) and we don't want to re-initialize the env file every time.

docker/entrypoint.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
```
diox, line 2: Now that this is a "real" entrypoint, we should probably have a comment in there explaining that this entrypoint is only used for CI/local envs.

diox (approved):

Makefile-os
KevinMind: I tested and it works... but I'd be wary of that. Setting a value to UID in the first place was kind of sketchy. the default value should be the UID of the current user, but by setting it we are then changing that. by keeping it in, we run the risk of that kind of a value collision in the future.

How long should we support it? I might prefer the set of error cases that you described which should reduce over time whereas keeping it keeps open a set of potential new error cases with no real end date.

docker/entrypoint.sh
KevinMind: in the compose file or the entrypoint script?

diox: The entrypoint script itself

Makefile-os
diox: Ok, let's not do that, we'll see if we run into those kind of issues.

KevinMind: Also just found out, we'd have to map it in the docker-compose.yml too. I'm seeing a pattern here that seems to stem from the idea that we have long running local environments that need to be supported or maintained over time.

I don't really see how you can have that in a codebase that is changing. When you run in production you build a new env and blow away the old one. Local should work like that too. If you are checking out a super old branch, and you don't recreate the local environment based on the state of the app at that time, you aren't really reproducing the same environment but just loading old code on to your current env.

I'm not sure this is a solvable problem and I would advocate we move to "blow it away" and add a feature for backing up specific data you want to hold onto across builds.

docker/entrypoint.sh
KevinMind: Done.

PR #22210, opened 2024-05-03
Add django check to verify docker image

KevinMind:

### Description

- add version.json check
- extract version.json loading to util
- add tests for version check
- add more sensible fallback for docker commit and version build

### Context

This change adds the django check to a make command allowing extensibility. We can (and will immeditaly after this PR) introduce further checks that can all be executed via `make check`.

Here we are also refactoring the logic of the version.json check. It is not important that the values inside and outside a container exactly match but that the container has a version.json file and it has all the required properties.

Finally, extracting the logic for loading the version.json to a shared util makes the values uniform in the now 2 places that require it.

### Testing

This is tested in CI so you can rely on that, but
you can test locally by building the image

```bash
make build_docker_image
```

Now shell into the image (not with make shell)

> NOTE: make sure you still have the version.json file crated after the build, this will overwrite the one created in the container.

```bash
docker run --rm -it -v $(pwd):/data/olympia --platform linux/amd64 addons-server-test:latest bash
```

Now in the container, run `make check`, it should pass. You can also cat the file to inspect visually. `cat version.json`, it should look like this

```json
{
  "commit": "14136defd176e3fbb0c8c2be1f9a600d073ee8c5",
  "version": "local",
  "build": "build",
  "source": "https://github.com/mozilla/addons-server"
}
```

Now you can rerun the check command with:

- removing 1 property from the file
- removing multiple properties from the file
- deleting the file

Rerun the command and it should fail. You can play around with adding removing properties to verify it works.

Required fields are specified in the tests.

### DIFF
--- a/Makefile-os
+++ b/Makefile-os
@@ -8,12 +8,11 @@ export DOCKER_BUILDER=container
 DOCKER_TAG := addons-server-test
 DOCKER_PLATFORM := linux/amd64
 DOCKER_PROGRESS := auto
-DOCKER_COMMIT := $(shell git rev-parse HEAD)
+export DOCKER_COMMIT := $(shell git rev-parse HEAD || echo "commit")
 DOCKER_CACHE_DIR := docker-cache
 
 export DOCKER_VERSION ?= local
-export DOCKER_COMMIT ?= $(shell git rev-parse --short HEAD)
-export VERSION_BUILD_URL ?= local
+export VERSION_BUILD_URL ?= build
 
 .PHONY: help_redirect
 help_redirect:
--- a/src/olympia/core/sentry.py
+++ b/src/olympia/core/sentry.py
@@ -1,7 +1,8 @@
-import json
 import os
 import re
 import urllib.parse
+
+from olympia.core.utils import get_version_json
 
 
 # List of fields to scrub in our custom sentry_before_send() callback.
@@ -16,18 +17,8 @@ SENTRY_SENSITIVE_FIELDS = (
 
 
 def get_sentry_release():
-    root = os.path.join(os.path.dirname(os.path.dirname(__file__)), '..', '..')
-    version_json = os.path.join(root, 'version.json')
-    version = 'unknown'
-
-    if os.path.exists(version_json):
-        try:
-            with open(version_json) as fobj:
-                contents = fobj.read()
-                data = json.loads(contents)
-                version = data.get('version') or data.get('commit')
-        except (OSError, KeyError):
-            version = None
+    version_json = get_version_json() or {}
+    version = version_json.get('version') or version_json.get('commit')
 
     # sentry is loaded before django so we have to read the env directly
     ensure_version = os.environ.get('REQUIRE_SENTRY_VERSION', False)
--- a/src/olympia/core/utils.py
+++ b/src/olympia/core/utils.py
@@ -1,0 +1,19 @@
+import json
+import os
+
+
+def get_version_json():
+    root = os.path.join(os.path.dirname(os.path.dirname(__file__)), '..', '..')
+    version_json = os.path.join(root, 'version.json')
+    version = None
+
+    if os.path.exists(version_json):
+        try:
+            with open(version_json) as fobj:
+                contents = fobj.read()
+                version = json.loads(contents)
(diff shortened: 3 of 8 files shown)


src/olympia/core/utils.py
```diff
@@ -0,0 +1,18 @@
...
+    root = os.path.join(os.path.dirname(os.path.dirname(__file__)), '..', '..')
+    version_json = os.path.join(root, 'version.json')
+    version = None
+
+    if os.path.exists(version_json):
+        try:
+            with open(version_json) as fobj:
+                contents = fobj.read()
+                version = json.loads(contents)
+        except Exception:
```
eviljeff, line 15: we were catching particular exceptions before, but it's just `Exception` - are there more possible (but acceptable) exceptions?

src/olympia/core/sentry.py
```diff
@@ -16,18 +17,11 @@
...
-                contents = fobj.read()
-                data = json.loads(contents)
-                version = data.get('version') or data.get('commit')
-        except (OSError, KeyError):
-            version = None
+    version = None
+    version_json = get_version_json()
+
+    if version_json:
+        version = version_json.get('version') or version_json.get('commit')
```
eviljeff, line 24:
```suggestion
    version_json = get_version_json() or {}
    version = version_json.get('version') or version_json.get('commit')
```

Makefile-os
```diff
@@ -8,12 +8,11 @@ export DOCKER_BUILDER=container
...
 DOCKER_PLATFORM := linux/amd64
 DOCKER_PROGRESS := auto
-DOCKER_COMMIT := $(shell git rev-parse HEAD)
+export DOCKER_COMMIT := $(shell git rev-parse HEAD || echo "commit")
 DOCKER_CACHE_DIR := docker-cache
 
 export DOCKER_VERSION ?= local
-export DOCKER_COMMIT ?= $(shell git rev-parse --short HEAD)
-export VERSION_BUILD_URL ?= local
+export VERSION_BUILD_URL ?= build
```
eviljeff, line 15: mmm, what does this change do?

eviljeff (approved):

Makefile-os
KevinMind: Makes it clear that the default value for version is "local" and "build is "build"

When you don't have any specified arguments

old:

```json
{
"version": "local",
"build": "local"
}
```

new:

```json
{
"version": "local",
"build": "build"
}
```

It just removes that tiny little edge case where you don't know if both values are set to the same value or you are only using one variable.

For the removal of DOCKER_COMMIT, it's actually a dupe. the end result would be if git isn't available (sometimes the case in the container) your commit would be "commit"

src/olympia/core/utils.py
KevinMind: There are a lot of exceptions that `could` be raised when reading a file and loading json.. It didn't make sense to me to try and collect all of them so I think there are 2 options:

1. don't catch any errors here, just let callers to the get_version_json handle them
2. catch all errors and return empty when something goes wrong.
3. Technically could revert to just the 2 specific previous exceptions, but that leads to inconsistent behavior.

Maybe we could log them at least though, to prevent silent errors from occurring.

src/olympia/core/sentry.py
KevinMind: Done

src/olympia/core/utils.py
eviljeff: I'm struggling to think of any legitimate errors (i.e. that we'd expect in normal operation) that aren't either the file isn't accessible or isn't valid json.  But if we want this to _always_ succeed as to not bork the wider checking function then I guess logging the exact error instead.

KevinMind: Actually looked at it again, and probably you are right to keep specific errors and just add the json decoding error as that is the most likely to occur if invalid json is recieved.

I realized we no longer need the KeyError as we aren't reading keys from the returned dictionary. I also realised we can catch the json errors and print those and os errors while returning a none object to prevent manage.py from totally failing because the version.json is missing or invalid.

https://github.com/mozilla/addons-server/pull/22210/files#diff-a99bef1b3c2386995edaab55833c3ac9a0f1a97e9e2679bb9555d99c332082d5R15

PR #22215, opened 2024-05-03
add AMO_IGNORE cinder action, and reporter email template

eviljeff:
fixes mozilla/addons#1787

It adds the AMO_IGNORE action, and a `CinderActionIgnore` action class that doesn't do anything (except define the reporter email template).  Along the way I renamed some of the exiting approve templates and tests that were named ignore to not confuse the two outcomes, and refactored some of the the more basic CinderAction subclasses to use mixins.

Note: the reporter email (unlike the other reporter emails) is not localized because we're including the policy text in it (to differentiate the alternate reasons for ignoring the report).  There may be follow-ups to address this flaw.


src/olympia/abuse/templates/abuse/emails/reporter_invalid_ignore.txt
```diff
@@ -0,0 +1,11 @@
+Hello,
+
+Thank you for your report about {{ name }}, at {{ target_url }}.
+
+Based on the information you submitted, we were unable to identify a violation of Mozilla’s policies  ({{ policy_document_url }}). The {{ type }} will therefore not be subject to any moderation action.
```
diox, line 5:
```suggestion
Based on the information you submitted, we were unable to identify a violation of Mozilla’s policies ({{ policy_document_url }}). The {{ type }} will therefore not be subject to any moderation action.
```

diox (approved):

PR #22223, opened 2024-05-06
Fix trigrams search for words of 5 characters

diox:
Fixes mozilla/addons#14773

### DIFF
--- a/src/olympia/search/filters.py
+++ b/src/olympia/search/filters.py
@@ -701,8 +701,12 @@ class SearchQueryFilter(BaseFilterBackend):
                                 }
                             }
                         },
-                        # For the trigrams query, we require at least 66% of the
-                        # trigrams to be present.
+                        # For the trigrams query, we require at least 67% of
+                        # the trigrams to be present. minimum_should_match is
+                        # rounded down to the lowest integer, so picking 66%
+                        # would mean 3 trigrams in the query would only require
+                        # 1 matching trigram in the name: int(3*66/100) = 1.
+                        # 67% ensures 2 need to match.
                         {
                             'match': {
                                 'name.trigrams': {
--- a/src/olympia/search/filters.py
+++ b/src/olympia/search/filters.py
@@ -707,7 +707,7 @@ class SearchQueryFilter(BaseFilterBackend):
                             'match': {
                                 'name.trigrams': {
                                     'query': search_query,
-                                    'minimum_should_match': '66%',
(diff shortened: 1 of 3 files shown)


src/olympia/search/filters.py
```diff
@@ -707,7 +707,7 @@ def primary_should_rules(self, search_query, lang):
                             'match': {
                                 'name.trigrams': {
                                     'query': search_query,
-                                    'minimum_should_match': '66%',
+                                    'minimum_should_match': '67%',
```
diox, line 710: `3*66/100 = 1.98`
`3*67/100 = 2.01`

Since 5 character words like "Adobe" have 3 trigrams, and `minimum_should_match` is always rounded down to the lowest integer, that change is enough to fix the problem and ensure "Dobromat" doesn't match "Adobe".

src/olympia/search/tests/test_search_ranking.py
```diff
@@ -709,7 +709,7 @@ def test_scenario_tabbicat(self):
         self._check_scenario(
             'tabbicat',
             (
-                ['Tabby Cat', 3255],
```
diox, line 712: That was the only result which had its score affected by this change, because it's not matching as well as before. Still high enough to be considered the best result for that query in our test db though.

src/olympia/search/tests/test_search_ranking.py
```diff
@@ -1063,3 +1063,10 @@ def test_scenario_promoted(self):
                 ['Stripy Dog 4', 584],  # not promoted
             ),
         )
+
+    def test_scenario_minimum_should_match_trigrams(self):
```
KevinMind, line 1067: This test scenario is hard to follow. Is it possible to mock the config for 66 and 67 and have an assert for both to verify modifying that value specifically modifies the expected outcome? This test is probably something that will randomly fail in the future due to some other partially related change and it will end up in a PR similar to Tabby Cat.

diox: These tests care about scoring so the way we have set them up is to have an entire set of add-ons that is used for each of the tests. That's fairly slow, so that's why it's the same for each, it's only created once for the entire test class.

Ultimately it's expected that changes to scoring/search algorithm will cause updates to multiple tests, that's fine. There is a script provided in this file to do bulk changes when required (In fact I expected to need to do more changes here, but ultimately didn't have to because of how much an edge case it is)

KevinMind (approved):
LGTM

PR #22233, opened 2024-05-09
send enforcement action to cinder & use jobs decision endpoint

eviljeff:
fixes mozilla/addons#14778 and fixes mozilla/addons#xxx

Fixed these two issues together as they touched the same code (but as two commits).  Now we send the enforcement action in addition to the policies a decision was made under, and if there's a job we use the new job specific endpoint for creating the decision (also means we don't need to close the job manually afterwards)


diox (approved):
Looks good. Haven't ran it locally but it matches the docs AFAICT so let's see it on dev.

PR #22242, opened 2024-05-13
Replace email link with EW page for suppressed email troubleshooting

wagnerand:
Fix https://github.com/mozilla/addons/issues/14786 by replacing email link with link to EW page.

### DIFF
--- a/src/olympia/devhub/templates/devhub/verify_email.html
+++ b/src/olympia/devhub/templates/devhub/verify_email.html
@@ -98,8 +98,8 @@
 {% with submit_text=button_text %}
   {% include 'devhub/verify_email_form.html' %}
 {% endwith %}
-{% trans %}
-  Having trouble? Please <a href="mailto:{{ support_email }}?subject={{ support_subject }}&body={{ support_body }}">contact AMO Admins</a> if you need help.
+{% trans a_attrs='target="_blank" rel="noopener noreferrer" href="%s/documentation/publish/developer-accounts/#email-issues?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=devhub"'|format(settings.EXTENSION_WORKSHOP_URL)|safe%}
+  If you encounter issues, please see our <a {{ a_attrs }}>troubleshooting suggestions</a> on Firefox Extension Workshop.
 {% endtrans %}
 </div>
 {% endif %}
--- a/src/olympia/devhub/tests/test_views.py
+++ b/src/olympia/devhub/tests/test_views.py
@@ -2361,6 +2361,12 @@ class TestVerifyEmail(TestCase):
             assert 'Send another email' in doc.text()
 
             assert 'If you encounter issues' in doc.text()
+            support_link = doc('a:contains("troubleshooting suggestions")')
+            assert (
+                '/documentation/publish/developer-accounts/#email-issues'
+                '?utm_source=addons.mozilla.org&utm_medium=referral'
+                '&utm_content=devhub' in support_link.attr('href')
+            )
 
     @mock.patch('olympia.devhub.views.check_suppressed_email_confirmation')
     def test_get_verification_delivered(self, mock_check_suppressed):
--- a/src/olympia/devhub/tests/test_views.py
+++ b/src/olympia/devhub/tests/test_views.py
@@ -2360,12 +2360,7 @@ class TestVerifyEmail(TestCase):
             assert 'It is taking longer than expected' in doc.text()
             assert 'Send another email' in doc.text()
 
-            assert 'Having trouble?' in doc.text()
-            support_link = doc('a:contains("contact AMO Admins")')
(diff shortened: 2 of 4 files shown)


wagnerand:
@KevinMind I'd like to get this in before the tag. Thank you.

KevinMind:
@wagnerand do you have this running on your local env? I would love to see a screenshot verifying the change, would mean I could just 👍 the PR.

wagnerand:
I tested this locally. I don't see how a screenshot would be sufficient for approving this PR, since you won't see if the link works (or the general HTML isn't broken), but here you go. 
Screenshot 2024-05-14 at 10 18 28

src/olympia/devhub/templates/devhub/verify_email.html
```diff
@@ -98,8 +98,8 @@ <h3>{{ _('Success!') }}</h3>
 {% with submit_text=button_text %}
   {% include 'devhub/verify_email_form.html' %}
 {% endwith %}
-{% trans %}
-  Having trouble? Please <a href="mailto:{{ support_email }}?subject={{ support_subject }}&body={{ support_body }}">contact AMO Admins</a> if you need help.
+{% trans a_attrs='target="_blank" rel="noopener noreferrer" href="%s/documentation/publish/developer-accounts/#email-issues?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=devhub"'|format(settings.EXTENSION_WORKSHOP_URL)|safe%}
```
KevinMind, line 101: Nit: If you defined the link href at least in the view, we could test it against an absolute link. This would also make verifying easier. Could also use urllib.join for a more "python knows the way" construction of the url itself.

```python
    path = 'documentation/publish/developer-accounts/#email-issues'
    query_params = {
        'utm_source': 'addons.mozilla.org',
        'utm_medium': 'referral',
        'utm_content': 'devhub'
    }
    
    query_string = urlencode(query_params)
    full_path = f"{path}?{query_string}"
    help_link = urljoin(settings.EXTENSION_WORKSHOP_URL, full_path)
```

src/olympia/devhub/tests/test_views.py
```diff
@@ -2360,12 +2360,7 @@ def test_get_verification_timedout(self, mock_check_emails):
             assert 'It is taking longer than expected' in doc.text()
             assert 'Send another email' in doc.text()
 
-            assert 'Having trouble?' in doc.text()
-            support_link = doc('a:contains("contact AMO Admins")')
-            assert f'mailto:{settings.SUPPORT_EMAIL}' in support_link.attr('href')
-            assert '?subject=Suppressed Email verification' in support_link.attr('href')
-            assert 'I have a suppressed email' in support_link.attr('href')
-            assert self.user_profile.email in support_link.attr('href')
+            assert 'If you encounter issues' in doc.text()
```
KevinMind, line 2363: Please add an assertion for the full relative url.

"/the-entire#url?with=query,params=foo"

src/olympia/devhub/templates/devhub/verify_email.html
wagnerand: I copy pasted this from other places in AMO. If we want to update that, we should update it everywhere (which would be out of scope for this PR), rather than having yet another different solution for adding a link to EW. If you grep for `_blank`, you'll see that we already have a bunch of different implementations, including some where we even hard-code the hostname.

src/olympia/devhub/tests/test_views.py
wagnerand: 605c1ac00bd

src/olympia/devhub/templates/devhub/verify_email.html
KevinMind: FWIW, I don't think many examples of bad code is a reason for more bad code. That is a pretty weak argument. Also, we don't have to update 100% of a bad pattern at once or not at all.

This is not a must have, just makes the code more reliant on standards based tools for building the URL and IMO more readable.

KevinMind (approved):
This PR is missing a testing section and the issue similarly has no section for how to test or verify the feature. I was able to figure out the flow and what flags are needed etc. but for other and or posterity This is what a testing section might look like

### Testing

Enable the waffle flag to view the suppressed email pages

```bash
make shell
```
in the shell

```bash
./manage.py waffle_switch suppressed-email on
```

Create a suppression record for your user.

Enter the djshell

```bash
make djshell
```

```python
SuppressedEmail.objects.create(email='kmeinhardt@mozilla.com')
```

Now you should be able. to see your suppressed email visiting http://olympia.test/en-US/developers/verify-email on your local environment.

Expect

1. the link points to the correct page https://extensionworkshop.allizom.org/documentation/publish/developer-accounts/#email-issues?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=devhub
2. the link opens in a new tab by default.
3. the url contains the correct UTM parameters to track the click back to this page.

PR #22244, opened 2024-05-14
Bump rabbitmq to 3.12 on local envs.

diox:
Note: you _need_ to run `docker compose down rabbitmq` at some point to start fresh: https://www.rabbitmq.com/blog/2022/07/20/required-feature-flags-in-rabbitmq-3.11

Fixes mozilla/addons#2014

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -109,7 +109,7 @@ services:
     image: redis:6.2
 
   rabbitmq:
-    image: rabbitmq:3.8
+    image: rabbitmq:3.12
     hostname: olympia
     expose:
       - "5672"


KevinMind:
I'm guessing there is no incompatibility between .8 and .12 but is there any specific way to verify this? I've never actually looked to what extend we use rabbitmq and how the dependency looks. Which set of tests would matter if something were to break.

KevinMind (approved):
LGTM, left a question about testing/verifying

diox:
We only need it for celery, which should be compatible (the feature set it uses has been constant). [Celery themselves run integration tests on rabbitmq latest](https://github.com/celery/celery/blob/4cf6ba3858779a3b71c87001133506c286aec0cd/.github/workflows/python-package.yml#L99).

PR #22246, opened 2024-05-14
"Fix makefile env variables definition"

KevinMind:
Fixes: https://github.com/mozilla/addons/issues/14787

### Description

The makefile env variables were not defined correctly, causing issues with the configuration of the containers. This pull request fixes the definition of the makefile env variables to ensure proper configuration.

### Context

We have 4 possible values of critical environment variables.

1. the default value define in the makefile
2. the value specified in the .env file
3. the value on the shell running the make command
4. the argument passed to the make command.

Critical environment variables are exactly those which might produce different behavior if running a command via a make target, or directly.

E.g. we have variables defined in our compose files which should be define on the environment. If docker compose <whatever> is run via make, values will be managed appropriately, but unless the value is in the .env or specified in the environment running docker compose directly will use a different or no value.

### Testing

I've added a CI job to verify that the make create_env_file will produce expected results given all 8 possible combinations of the 4 possible sources

You can also manually test 

1. default value.

Remove .env file

run 

```bash
make create_env_file
```

Expect all values to be the value defined in the default parameter via make file.

2. ..env file

Modify one of the values in the produced in the .env and rerun the command.

Expect the value is persisted from the .env

3. env on the shell

Keep the .env file,  and set the value in the shell before running the command. Re-run and expect the env to take priority.

```bash
HOST_UID=env make create_env_file
```

4. args on the command

Now adding to this, if you further specify an argument and expect it to take precedence over all values, regardless of if env/file/default are specified.

```bash
HOST_UID=env make create_env_file HOST_UID=arg
```

Expect value to be arg now.

The CI check runs through effectively this set of steps for every variable that is in this critical list.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -667,7 +667,7 @@ workflows:
       - devhub
       - main
       # Uncomment if you want to test the docker build
-      - build-image
+      # - build-image
       - reviewers-and-zadmin
       - es-tests
       - localization
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -369,6 +369,12 @@ commands:
             echo 'export DOCKER_PUSH=<< parameters.push >>' >> $BASH_ENV
             echo 'export DOCKER_OUTPUT=<< parameters.output_file >>' >> $BASH_ENV
       - run:
+          name: Create .env file
+          command: |
+            # We must defined a .env file for the docker-compose config
+            # The values will be inferred by what is set in the bash env above.
+            make create_env_file
+      - run:
           name: Docker build config
           command: |
             make docker_compose_config
@@ -661,7 +667,7 @@ workflows:
       - devhub
       - main
       # Uncomment if you want to test the docker build
-      # - build-image
+      - build-image
       - reviewers-and-zadmin
       - es-tests
       - localization
--- a/.github/workflows/verify-docker-image.yml
+++ b/.github/workflows/verify-docker-image.yml
@@ -11,42 +11,80 @@ jobs:
     strategy:
       matrix:
         include:
-          - expected: { version: local, push: false }
-            input: { version: '', push: '' }
-          - expected: { version: version, push: true }
-            input: { version: version, push: true }
+          - expected: { version: 'local', push: false, uid: ''}
+            input: { version: '', push: '', uid: '' }
+          - expected: { version: version, push: true, uid: '1' }
+            input: { version: version, push: true, uid: '1' }
 
     steps:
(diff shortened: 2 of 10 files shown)


.github/workflows/verify-docker-image.yml
```diff
@@ -11,42 +11,80 @@ jobs:
...
-          DOCKER_PUSH: ${{ matrix.input.push }}
         shell: bash
         run: |
-          set -xue
-          config=$(make docker_compose_config)
+          set -ue
 
-          # Expect image tag is correct
-          echo $config | grep -q "image: mozilla/addons-server:${{ matrix.expected.version }}"
+          export SUPERUSER_EMAIL="email"
```
KevinMind, line 28: I could probably reimplement this in python... wdyt?

.circleci/config.yml
```diff
@@ -597,6 +597,14 @@ jobs:
               --ignore src/olympia/versions/ \
               --ignore src/olympia/zadmin
 
+  make_tests:
```
KevinMind, line 600: This should run as a separate test because it modifies the environment and .env files... it's theoretically possible to clean up that data after, but probably not worth the inherent risk.

.github/workflows/verify-docker-image.yml
diox: Probably worth extracting somewhere (in Python or not) cause there is a lot going on and it's hard to read...

KevinMind: I've extracted the exiisting bash test, and the new tests introduced here to a JS `make.spec.js` 

The reason for JS is that the test MUST run on the host, it relies on docker commands and we don't (probably don't want to) install docker in our container. no need for the bloat.

Theoretically I could write it in python, but I felt it easier to run the js tests on the host.

docker-compose.yml
```diff
@@ -1,4 +1,9 @@
 x-env-mapping: &env
+  # https://docs.docker.com/compose/environment-variables/envvars-precedence/
+  env_file:
```
KevinMind, line 3: This is an important, but I think good change. This means if there is no .env file, docker compose will just fail. It would prevent someone from running anything via docker compose before running `make create_env_file`.

There is not really a scenario where you should run docker compose without it, it is critical for configuring the compose project.

happy to hear push back if you think this is too strong enforcement.

KevinMind: Boo, docker compose version in gha is not compatible with this syntax. Not sure it's worth the effort to upgrade.

diox (approved):

PR #22247, opened 2024-05-15
Upgrade to python 3.11

diox:
Fixes mozilla/addons#1965

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # Read the docs/topics/development/docker.md file for more information about this Dockerfile.
 ####################################################################################################
 
-FROM python:3.10-slim-bookworm as olympia
+FROM python:3.11-slim-bookworm as olympia
 
 ENV OLYMPIA_UID=9500
 RUN <<EOF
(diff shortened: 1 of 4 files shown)


diox:
You can test this by... well... doing anything in the container really. CI has been updated to run with the new version.

setup.py
```diff
@@ -25,7 +25,7 @@
         'Framework :: Django',
         'Topic :: Internet :: WWW/HTTP :: Browsers',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.10',
```
KevinMind, line 28: See above, if PYTHON_VERSION is specified in the container, you can substitute here, maybe it is defined in settings? or straight up environment.

Dockerfile
```diff
@@ -2,7 +2,7 @@
 # Read the docs/topics/development/docker.md file for more information about this Dockerfile.
 ####################################################################################################
 
-FROM python:3.10-slim-bookworm as olympia
```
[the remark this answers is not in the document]
KevinMind: If we'd like to keep it hardcoded, could you remove the input from the build action (it's been there for a while with nothing to do)

setup.py
diox: Well since we're not publishing the package anyway I'm thinking of completely removing this classifier actually. It shouldn't matter.

KevinMind (approved):
LGTM

PR #22248, opened 2024-05-16
Run github action (run-docker) action using docker compose

KevinMind:
Fixes: https://github.com/mozilla/addons/issues/14792

### Description

Convert our reusable action run-docker to use docker compose exec instead of simply running the single container via docker run.

### Context

Enabling run-docker to use compose means we run our containers the same way in local and ci environments. This makes the flow much simpler and the fidelity between running a single command in local or ci much higher.

Furthermore, running via docker compose allows us to reuse existing compose definitions for other services that are required for some CI jobs.

### Testing

CI

### DIFF
--- a/.github/workflows/verify-docker-image.yml
+++ b/.github/workflows/verify-docker-image.yml
@@ -35,7 +35,7 @@ jobs:
         id: failure
         uses: ./.github/actions/run-docker
         with:
-          image: ${{ steps.build.outputs.tags }}
+          version: ${{ steps.build.outputs.version }}
           run: |
             exit 1
         continue-on-error: true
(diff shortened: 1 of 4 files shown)


.github/workflows/verify-docker-image.yml
```diff
@@ -59,8 +59,27 @@ jobs:
...
         uses: ./.github/actions/run-docker
         with:
-          image: ${{ steps.build.outputs.tags }}
-          options:
+          version: ${{ steps.build.outputs.version }}
           run: |
-            make update_deps
             make check
+
+  test:
```
diox, line 66: Consider removing that for now - we've proven that it works, but until we do the full transition (PR in the very near future) we don't need to run these.

diox (approved):

.github/workflows/verify-docker-image.yml
KevinMind: Done.

PR #22253, opened 2024-05-17
TMP: create development/production stages to build dependencies

KevinMind:
TBD

PR #22254, opened 2024-05-17
verify pip dependencies in CI

KevinMind:
### Description

Add check to CI verifying that pip packages are installed in the container.

- add script to determine all dependencies specified in a set of requirements.txt files are installed via `pip show`
- add pip_development stage to docker build to install dev related dependencies (like dennis previously in it's own requirements file) (not included in final image)
- merge locale.txt into dev.txt dependencies (we have dev stage now, don't need it)
- add ci check to verify pip. packages are installed on PRs

### Context

We rely on some (mildly) complex logic to install our dependencies. For this reason, it makes sense to verify that dependencies that are "required" actually get installed in the final container. This PR adds such a check.

Having this check is very useful (for example) in this PR that wants to modify how dependencies are finally added to the image https://github.com/mozilla/addons-server/pull/22253

### Testing

You can verify this check works, by adding a package to a requirements.txt file, NOT installing it, and running the check command specifying that file.

1. add a package (any package) to prod.txt
2. run `make check`
3. should exit 1 that the package you added is "<package": Missing"


diox:
No deal breakers, but:
- Because it's also printing packages that were found, it's a bit annoying to find the missing package(s) when that happens. Do we need to print the found packages ?
- It's rather slow - around 25 to 30 seconds on my machine. Looks like because the individual calls to `pip show` are slow. Maybe we can address this ? It takes `pip freeze -l` under a second to list all installed packages...
- Should we check for versions as well ?

diox:
> It's rather slow - around 25 to 30 seconds on my machine. Looks like because the individual calls to pip show are slow. Maybe we can address this ? It takes `pip freeze -l` under a second to list all installed packages...

`pip freeze --all --exclude-editable | sort -u` is almost comparable to `cat requirements/dev.txt requirements/prod.txt| grep -E -v '^\s*(#|--hash)' | cut -d ' ' -f 1 | sort -u` using `diff` FYI. There are only 3 things to deal with using this approach:
- The special `package[foo]` case
- `bitarray == 2.9.2` in `requirements/prod.txt` should be specified without spaces
- Some packages are specified with underscores instead of dashes and should be fixed as well

As a bonus this approach would deal with the version numbers being different.

KevinMind:
Sweet. I'll add this.

diox (approved):
I wonder if dependabot will recreate package names with underscore or capital letters... I guess we'll find out, we can easily adapt the script if it does.

(remove the TMP: from the PR title before merging)

PR #22256, opened 2024-05-20
Use the latest delayed-rejection comments when auto-rejecting, not the first

diox:
Fixes mozilla/addons#14793

### DIFF
--- a/src/olympia/reviewers/management/commands/auto_reject.py
+++ b/src/olympia/reviewers/management/commands/auto_reject.py
@@ -77,6 +77,7 @@ class Command(BaseCommand):
                     amo.LOG.REJECT_VERSION_DELAYED.id,
                 )
             )
+            .order_by('created')
             .last()
         )
         log_details = getattr(relevant_activity_log, 'details', {})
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/management/commands/auto_reject.py
```diff
@@ -77,6 +77,7 @@ def reject_versions(self, *, addon, versions, latest_version):
                     amo.LOG.REJECT_VERSION_DELAYED.id,
                 )
             )
+            .order_by('created')
```
eviljeff, line 80: the `Meta` ordering is `-created` so changing `.last()` to `.first()` would do the same thing, right?  No preference is this is clearer though.

eviljeff (approved):

src/olympia/reviewers/management/commands/auto_reject.py
diox: Yes. I figured it would be less confusing to keep using `last()` and force the ordering since we do want the "last created".

PR #22250, opened 2024-05-16
Implement All CircleCI tests in github actions

KevinMind:
Fixes: [AMOENG-599](https://mozilla-hub.atlassian.net/browse/AMOENG-599)

### Description

This is the first PR in the final phase of the gha migration. It implements parity between circle ci and github actions for tests. All tests run in circle ci are also run in github actions.

### Context

After merging this there are follow ups for finalizing the migration and for improving performance. Though we are at near parity with circle CI (while building the docker container so actually faster already).

This PR includes 2 sub PRs that could be merged before or with this one:

- https://github.com/mozilla/addons-server/pull/22280
- https://github.com/mozilla/addons-server/pull/22299

PR #22260, opened 2024-05-20
Docker digest

KevinMind:
Fixes: https://github.com/mozilla/addons/issues/14797

### Description

This PR adds support for running our docker compose services via the build digest instead of the version. Version is still supported in cases where you want to run on "latest" or the latest build of a tag for example.

The PR achieves several discrete and related things:
- refactors the test suite for docker/make configurations 
- refactors the script to create version.json and .env into a single python script.
- adds the ability to specify DOCKER_VERSION as a tag or a sha256:string formatting correctly to the docker compose yml
- removes the mysql volume name environment variable as it is static and can be verified to be correct via tests alone.
- updates existing CI jobs to use image digest instead of version

### Context

This PR implements a fix for a bug that was discovered during the preparation of https://github.com/mozilla/addons-server/pull/22250 and should be a prerequisite for continuing.

Coincidentally and (I believe) beneficially, we now totally decouple the specification of .env values used in docker-compose substitution and the make file. This prevents over-reliance of make commands on these values as they are "special" and shouldn't be used outside of substituting values in compose.

An important edge case that is covered is that we need to prefix DOCKER_VERSION. This is because the way you specify an image version is different for tags or digests. 

mozilla/addons-server@sha2456:<id> or mozilla/addons-server:latest

Tags use `:` and digests use `@`. This behavior is tested and should work regardless of how you specify the DOCKER_VERSION at run time.

### Testing

You can test locally running the test suite or play around with the specifications.

```bash
<name>=<value> make setup
```

Relevant values are noted in the test suite here (https://github.com/mozilla/addons-server/pull/22260/files#diff-1904f45ef3db1fec0d686e0545e32db0dca4c7c4151c56ca56a4f803f96e194bR115-R169)


Makefile-os
```diff
@@ -2,18 +2,6 @@
 # Our makefile makes use of docker compose commands. Our config files rely on environment variables
 # both for passing configuration to the containers as well as configuring the compose file itself.
 # Variables referenced in docker-compose*.yml should be read from .env, exported and saved in .env
-
-define ENV_VAR
```
KevinMind, line 6: This logic was already very hard to follow, and now with the requirement to additionally prefix the DOCKER_VERSION value, it makes a lot of sense to move this to a python script.

tests/make/make.spec.js
```diff
@@ -2,207 +2,157 @@ const { spawnSync } = require('child_process');
...
 
-    const composeFiles = globSync('docker-compose*.yml', { cwd: rootPath });
-    const variableDefinitions = [];
+  runSetup(values);
 
-    for (let file of composeFiles) {
-      const fileContent = fs.readFileSync(path.join(rootPath, file), {
-        encoding: 'utf-8',
-      });
+  const { stdout: rawConfig } = spawnSync(
```
KevinMind, line 48: We can call docker explicitly here to test the scenario where a user doesn't use make and must rely only on .env or env variables.

tests/make/make.spec.js
```diff
@@ -2,207 +2,157 @@ const { spawnSync } = require('child_process');
...
+    {
+      name,
+      file: 'file',
+      env: 'env',
+      expected: 'env',
+    },
+  ];
+}
+
+const testCases = [
```
KevinMind, line 115: This format for the env variable test is much more flexible and simple. Also results in less code.

diox (approved):
More descriptive PR/final commit title please

PR #22272, opened 2024-05-22
Remove "accessibility" category

diox:
It will be replaced by a curated collection instead. Existing add-ons with only that category are migrated to the "Other" category instead.

Removing it from the constants should automatically make it disappear from the API (and in turn the frontend) and DevHub.

Fixes https://github.com/mozilla/addons/issues/14801


src/olympia/devhub/tests/test_views_edit.py
```diff
@@ -487,24 +488,38 @@ class TestEditDescribeListed(BaseTestEditDescribe, L10nTestsMixin):
     __test__ = True
 
     def test_edit_categories_add(self):
-        assert [c.id for c in self.get_addon().all_categories] == [22]
-        self.cat_initial['categories'] = [22, 1]
```
diox, line 491: Here and below I took the opportunity to clean up and remove hardcoded IDs in tests

src/olympia/addons/migrations/0051_remove_accessibility.py
```diff
@@ -0,0 +1,44 @@
...
+from django.db import migrations, models
+from olympia import amo
+from olympia.constants.categories import CATEGORIES
+
+
+def remove_accessibility_category(apps, schema_editor):
+    Addon = apps.get_model('addons', 'Addon')
+    AddonCategory = apps.get_model('addons', 'AddonCategory')
+    other_category_id = CATEGORIES[amo.ADDON_EXTENSION]['other'].id
+    old_accessibility_category_id = 74
```
KevinMind, line 12: Speaking of old ids, could probably derive this ID from a query. What if there is ever another category with id 74 and then this migration is run again... not saying that is likely, but weirder things have happened.

KevinMind:
Is there any expectation or possibility to verify this? How would I do so locally?

src/olympia/addons/migrations/0051_remove_accessibility.py
```diff
@@ -0,0 +1,44 @@
+# Generated by Django 4.2.13 on 2024-05-22 11:39
+
+from django.db import migrations, models
+from olympia import amo
+from olympia.constants.categories import CATEGORIES
+
+
+def remove_accessibility_category(apps, schema_editor):
```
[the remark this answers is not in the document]
diox: As a developer, when you select the `Other` category you can't choose another category - if you choose that one it should be the only one for your add-on. If we did a replace here, potentially some add-ons would go against that and end up with `Other` and up to 2 other categories, and those add-ons would be a weird state (next time they try to edit metadata they would get an error and need to fix that)

Now, in some instances when doing category migrations we haven't been able to avoid that, but here we can so I figured I would.

src/olympia/addons/migrations/0051_remove_accessibility.py
diox: Category IDs are only defined in constants (and the constant is gone at this point) so we can't really query for that.

For constants of things we keep in db, like activity action ids, we keep a constant in the file and prefix it (and have a test making sure activity action ids are all unique), but here since I'm getting rid of all relevant `AddonCategory` rows entirely I figured I wouldn't need to worry about that. I'm open to alternatives though.

diox:
Verifying this locally:
- Submit a listed add-on with the `Accessibility` category set using `master` branch
- Switch to this branch, apply migrations: your add-on should now be in the `Other` category instead (can be seen on its detail page if it was public, or in devhub edit page if not). Other add-ons that may exist in your env should not be affected.
- Once on this branch, http://olympia.test/api/v5/addons/categories/ should not return `Accessibility` category anymore, and the `Accessibility` category should also be gone from the list of categories you can select from.

src/olympia/addons/migrations/0051_remove_accessibility.py
```diff
@@ -0,0 +1,44 @@
...
+    Addon = apps.get_model('addons', 'Addon')
+    AddonCategory = apps.get_model('addons', 'AddonCategory')
+    other_category_id = CATEGORIES[amo.ADDON_EXTENSION]['other'].id
+    old_accessibility_category_id = 74
+
+    # Remove accessibilty category from every add-on.
+    AddonCategory.objects.filter(category_id=old_accessibility_category_id).delete()
```
eviljeff, line 15: If we get a list of add-on IDs from this qs before deleting we can limit the create query to only those affected add-ons - so we aren't "fixing" listings that weren't complete because the developer didn't get that far in the submission flow on devhub.

eviljeff (approved):

src/olympia/addons/migrations/0051_remove_accessibility.py
diox: We're only fixing those with `status=amo.STATUS_APPROVED` so it should not impact those that didn't get that far in the submission flow

PR #22273, opened 2024-05-23
Fix Docker build and add concurrency limit

KevinMind:
fixes: mozilla/addons#14802

### Description

Fix the computed image tag in our docker compose file causing tags to be invalid and therrefore failing to push

### Context

We added (here) support for running our container via a tag or a digest, but the logic had a bug that meant if you defined DOCKER_VERSION in the shell environment, it would override the formattted value produced by calling make setup.

Essentially, the value in the .env is updated but not in the shel environment.

Fun facts: There is a different between an image Id, and image digest and an image configuration digest.

| Field                | Description                                                                                                                                                                                                                                                                                                                                                                               |
|----------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| **id**               | Image ID is a unique identifier for a Docker image that is different on every build, even if the contents are the same.                                                                                                                                                                                                                                                                   |
| **digest**           | A hash of the image layer content digests. A hash of hashes that is produced only when an image is pushed to a Docker registry. (To further confuse, it is also produced when building with buildkit, which is kind of like a pseudo-registry because it's running on its own Docker container that exports to the local daemon)                                                          |
| **configuration digest** | This is a hash of the image configuration object which is what you get when you inspect an image. It is not 1:1 with a digest because it doesn't (necessarily) take into account changes in the contents of the image, but in its configuration. For example, you could have a change in a RUN step that produces no change in the content of the layer, but it is still a change. Kind of confusing. |

### Testing

There are several flows to test.

First and formost. remove your .env file and create a new one!

`rm -f .env && make setup`

Not doing this likely won't cause any problems but better consistent than sorry.

#### Run with digest

1. Grab a digest from a built image in CI. This will NOT work with locally built images.
2. set the digest and up the environment `DOCKER_DIGEST=<value> make up`

This should run the container exactly from the build.

#### Run with tag

1. set the tag to latest `DOCKER_VERSION=latest make up`
2. This should run the latest tag from dockerhub.

#### Test concurrency.

If you have an open PR from this branch, with jobs in progress. Pushing a new commit should queue a new workflow run and cancel the previous one. There should only be one running at a time. This will save a huge amount of resources as the number of jobs per workflow increases.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,10 +1,6 @@
 name: Docs
 
 on:
-  # Build docs on pull request to verify changes
-  pull_request:
-    branches:
-      - master
   # Deploy to production on master
   push:
     branches:
@@ -12,41 +8,18 @@ on:
   # Deploy `master` on workflow dispatch.
   workflow_dispatch:
 
-jobs:
-  build_docs:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-      - id: build
-        uses: ./.github/actions/build-docker
-      - name: Setup Pages
-        uses: actions/configure-pages@v4
-      - name: Build Docs
-        uses: ./.github/actions/run-docker
-        with:
-          version: ${{ steps.build.outputs.digest }}
-          run: |
-            make docs
-      - name: Upload artifact
-        uses: actions/upload-pages-artifact@v3
-        with:
-          path: 'docs/_build/html'
-          name: "docs"
+# Only deploy the latest build artifact to GitHub Pages
+concurrency:
+  group: "pages"
+  cancel-in-progress: true
 
+jobs:
   deploy_docs:
-    # only deploy on push to the master branch
-    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
     # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
     permissions:
       contents: read
       pages: write
       id-token: write
-    # Only deploy the latest build artifact to GitHub Pages
-    concurrency:
-      group: "pages"
-      cancel-in-progress: true
-    needs: build_docs
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
@@ -56,4 +29,5 @@ jobs:
         id: deployment
         uses: actions/deploy-pages@v4
         with:
-          artifact_name: "docs"
+          # Keep this in sync with `build_docs` where we upload the artifact
+          artifact_name: ${{ secrets.DOCS_ARTIFACT_NAME}}
--- a/.github/workflows/verify-docker-image.yml
+++ b/.github/workflows/verify-docker-image.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - master
 
+concurrency:
+  group: verify-docker-image
+  cancel-in-progress: true
+
 jobs:
   docker_config_check:
     runs-on: ubuntu-latest
@@ -35,7 +39,7 @@ jobs:
         id: failure
         uses: ./.github/actions/run-docker
         with:
-          version: ${{ steps.build.outputs.digest }}
+          version: ${{ steps.build.outputs.version }}
           run: |
             exit 1
         continue-on-error: true
@@ -50,7 +54,7 @@ jobs:
       - name: Check (special characters in command)
         uses: ./.github/actions/run-docker
         with:
-          version: ${{ steps.build.outputs.digest }}
+          version: ${{ steps.build.outputs.version }}
           run: |
             echo 'this is a question?'
             echo 'a * is born'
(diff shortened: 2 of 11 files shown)


KevinMind:
Confirmed run of build-image with push that shows we are pushing the right tag. we don't have permission to push this tag it seems but we are pushing the "correct" one.

https://app.circleci.com/pipelines/github/mozilla/addons-server/31633/workflows/7c936b2e-547d-4bb3-b7a3-5380e599e584/jobs/215049

.github/workflows/docs.yml
```diff
@@ -1,52 +1,25 @@
 name: Docs
 
 on:
-  # Build docs on pull request to verify changes
```
KevinMind, line 4: I think it is simpler to move the "building" of the docs to the verify-docker-image job which should actually just be called the "ci" job or the "pull request" job. It is "everything" that runs on a pull request and uses a single build to run all tests and produce all artifacts.

KevinMind: Same rationale for extract locales. We "could" solve this by no specifying either version or digest and forcing the run action to build the image but this will build using a different mechanism than our "build-docker" job so not 100% guaranteed it will produce the same result. That's like a huge mega edge case though so happy to hear push back.

.github/workflows/verify-docker-image.yml
```diff
@@ -21,21 +25,34 @@ jobs:
           docker compose version
           npm exec jest -- ./tests/make --runInBand
 
+  build:
```
KevinMind, line 28: This is probably the most contraversial change. We are building the image and actually pushing it to dockerhub. This means the image is available for anyone to run locally (could be seen as a good thing) especially for testing you can just specify the tag and run the exact latest version of whatever is made in the pr locally. Kind of cool.

I have investigated whether it is possible to remove iamges from dockerhub programatically. It is using their api or a community action so we could utilize that if we don't want images to sit there forever.

Happy to hear feedback on this. It is totally possible to implement these changes without pushing the image. It just means we have to build locally (in the runner I mean) a lot more and we cannot be 100% certain that each job is running with the exact same build. This is the only way to guarantee that.

KevinMind:
Interesting result that counters what I've written above.

I reran the CI job which re "built" and pushed the image. I was expecting it to produce the same digest because literally nothing changed. I can verify that is true by container diffing the images.

```bash
kmeinhardt@kevins-mbp ~ % container-diff diff remote://mozilla/addons-server@sha256:1e0e0fd67bfec5e6ce4419961315fd13022c01337f5d0c10ebbe387dfd43c7e5 remote://mozilla/addons-server@sha256:01a25462865cfee0f91d2e542b83a3b631a1d52bd167ffe76911fd7d1b4b5a11 --type=history --type=file --type=pip --type=apt --type=node --json
ERRO[0001] error getting diff with HistoryAnalyzer: unsupported status code 404; body: 404 page not found
ERRO[0001] error getting diff with PipAnalyzer: unsupported status code 404; body: 404 page not found
[
  {
    "Image1": "mozilla/addons-server@sha256:1e0e0fd67bfec5e6ce4419961315fd13022c01337f5d0c10ebbe387dfd43c7e5",
    "Image2": "mozilla/addons-server@sha256:01a25462865cfee0f91d2e542b83a3b631a1d52bd167ffe76911fd7d1b4b5a11",
    "DiffType": "Apt",
    "Diff": {
      "Packages1": [],
      "Packages2": [],
      "InfoDiff": []
    }
  },
  {
    "Image1": "mozilla/addons-server@sha256:1e0e0fd67bfec5e6ce4419961315fd13022c01337f5d0c10ebbe387dfd43c7e5",
    "Image2": "mozilla/addons-server@sha256:01a25462865cfee0f91d2e542b83a3b631a1d52bd167ffe76911fd7d1b4b5a11",
    "DiffType": "File",
    "Diff": {
      "Adds": null,
      "Dels": null,
      "Mods": null
    }
  },
  {
    "Image1": "mozilla/addons-server@sha256:1e0e0fd67bfec5e6ce4419961315fd13022c01337f5d0c10ebbe387dfd43c7e5",
    "Image2": "mozilla/addons-server@sha256:01a25462865cfee0f91d2e542b83a3b631a1d52bd167ffe76911fd7d1b4b5a11",
    "DiffType": "Node",
    "Diff": {
      "Packages1": [],
      "Packages2": [],
      "InfoDiff": []
    }
  }
]%
```

No file/pip/apt/node changes. It is failing on the "history" so maybe there is a slight change in the configuration object but I was under the impression the configuration object wouldn't change if the exact same commit on the repo is re-ran twice.. how could it. maybe there is some time based data in the docker context that I don't know about... 

This isn't really a deal breaker, it's just weird.

.github/workflows/verify-docker-image.yml
diox: Can we put this in a separate PR to unblock builds now while we think about it ? I'm not sure we want to do that so I'd prefer if we focused on unblocking builds and take the time to think about the implications of pushing an image for every PR.

.github/workflows/docs.yml
diox: I agree - we should eventually rename that workflow, but for now this is fine.

.github/workflows/verify-docker-image.yml
KevinMind: Yep, can do. I'll rollback that part of the PR.

docker-compose.yml
```diff
@@ -27,7 +27,7 @@ x-env-mapping: &env
 services:
   worker: &worker
     <<: *env
-    image: mozilla/addons-server${DOCKER_VERSION:-}
+    image: ${DOCKER_TAG:-}
```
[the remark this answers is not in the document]
KevinMind: Yep. running make up will automatically fix and yes running raw docker commands will break.

.github/workflows/verify-docker-image.yml
KevinMind: Done. I used a force push to keep everything on a single commit.

KevinMind: I've kept the version reference as it is using a more "unique" version tag anyway that should drastically reduce likelihood of collisions unless you push an image with the exact PR number, ref and commit hash.

Makefile-os
```diff
@@ -7,7 +7,7 @@
 DOCKER_BUILDER ?= container
 DOCKER_PROGRESS ?= auto
 DOCKER_PUSH ?= false
-DOCKER_OUTPUT ?=
+BUILDX_BAKE_METADATA_FILE ?= buildx-bake-metadata.json
```
KevinMind, line 10: Giving this file a more explanatory name and a default value. It is ignored so we can create it by default. This means you don't have to expcitly opt in or know the location of the file. Useful for CI jobs that want to read this data.

diox:
> get the image ID from the build make docker_image_id

should be `make docker_image_digest` right ?

diox (approved):
Still unsure about whether or not the added complexity of supporting digests is worth it but we can try it out.

PR #22276, opened 2024-05-23
Explicitly raise an exception when trying to create an ActivityLog with no user

diox:
Let's see what this breaks...

Fixes mozilla/addons#14803


diox:
> Let's see what this breaks...

Pretty much everything then, since this breaks `addon_factory()` if used without setting a user in the thread...

PR #22277, opened 2024-05-23
Fix ActivityLog recording when flagging a version because of an abuse report

diox:
When we're manually creating the `ActivityLog`, we need to pass a user for the `ActivityLog` to be saved. This code could be called from somewhere where there is no user in the thread, so explicitly fall back on the task user when that's happening.

Fixes mozilla/addons#2010

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -8,13 +8,14 @@ import requests
 import waffle
 
 import olympia
-from olympia import activity, amo
+from olympia import activity, amo, core
 from olympia.amo.utils import (
     backup_storage_enabled,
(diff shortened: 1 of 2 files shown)


eviljeff (approved):
Can we have a line in one the existing tests that proves the activitylog is indeed created?

diox:
> Can we have a line in one the existing tests that proves the activitylog is indeed created?


`test_report()` in `TestCinderAddonHandledByReviewers` does it. It was passing because the test was doing a `set_user()` in `setUp()`, so I removed that, ensured test was failing without my fix, and also added another assertion verifying that `get_user()` returns `None` in that test now.

PR #22278, opened 2024-05-23
Implement resolving jobs as invalid in reviewer tools

eviljeff:
fixes mozilla/addons#14777 by:
- stopping using `clear_needs_human_review_multiple_versions` to resolve cinderjobs
- now only dropping needs human review for cinder jobs when making the decision (and preventing those NHR from being cleared in the reviewer tools otherwise)
- exposing CinderPolicy instances directly in the reviewer tools, and making them selectable instead of ReviewActionReasons (for now just for actions that are only on the CinderJob itself)
- Adding to the CinderPolicy model so Operations can make the policy available, and set a cinder_action for them.
- Linking the CinderPolicy-s selected to the ActivityLog, so the abuse models don't need to know/care about ReviewActionReasons.
- defining the cinder action in the activitylog details, in reviewers/utils.pu, because the logic to choose the action is more complicated, and it made it more understandable to set it there and just have CinderDecision consume it.  (Plus it enables future work to allow the action to be overridden by reviewers or selectable based on other criteria)

This patch doesn't include a migration to set up the CinderPolicies - it _could_ but it'd have to be pretty lose because the policy uuids are generated so are different on cinder stage and cinder prod.

image


diox:
Still testing this but I've encountered something weird/unexpected:

Context:
- I've reported an add-on multiple times for violating policies inside the add-on
- I see the add-on in the queue as expected
- I've synced policies in the admin, set a few of them as exposed in reviewer tools with a cinder action attached. 
- All dsa related waffle switches are on
- Cinder API credentials are set

When selecting "Comment/Resolve Jobs", I can see my reports and the policies that are meant to be exposed. I add a comment, select a ignore policy (action should be `AMO_IGNORE` and tick the box to resolve the jobs before clicking save...

... and although the review page says "Review successfully processed." at the top, the reports are still shown at the bottom when choosing "Comment/Resolve Jobs" action. The version is still flagged and my reviewer comment is nowhere to be seen.

eviljeff:
> Still testing this but I've encountered something weird/unexpected:

That is indeed a weird combination of outcomes.  The log entry should _always_ be created, even if nothing was being resolved; and that a `CinderDecision` exists implies that the task was correctly called and the log entry existed (!) with the correct `.details` json, or a `ImproperlyConfigured` exception would have been raised.

src/olympia/reviewers/forms.py
```diff
@@ -338,11 +344,35 @@ def is_valid(self):
...
+            actions = self.helper.handler.get_cinder_actions_from_policies(
+                self.cleaned_data.get('cinder_policies')
+            )
+            if len(actions) == 0:
+                raise ValidationError(
+                    'No policies selected with an associated cinder action.'
+                )
+            elif len(actions) > 1:
+                raise ValidationError(
+                    'Multiple policies selected with differnt cinder actions.'
```
diox, line 372:
```suggestion
                    'Multiple policies selected with different cinder actions.'
```

PR #22295, opened 2024-05-28
Use separate activity log action for automated rejections after delay expiry

wagnerand:
Fixes https://github.com/mozilla/addons/issues/14810

This PR introduces separate activity log actions for automated rejections after the delay expired, in order to distinguish them from instant rejections.

# How to test
1. Submit a new add-on.
2. Delay reject it in the reviewer tools (you need a `ReviewReason` and a `CinderPolicy` to exist for this).
   * Cinder policies can be synced from cinder stage (stick `CINDER_SERVER_URL='https://stage.cinder.nonprod.webservices.mozgcp.net/api/v1/'` and `CINDER_API_TOKEN='<whateverthatis>'` in your local settings, then sync them on `/admin/models/abuse/cinderpolicy/`)
   * If needed, create a new `ReviewReason` (`/admin/models/reviewers/reviewactionreason/`, reference an existing `CinderPolicy`)
3. Update the delay expiry to be in the past, so you don't have to wait: Run `make dbshell`, then something like the following. Note this will update all rows in the table, add a condition if you don't want that.
   ```
   UPDATE versions_versionreviewerflags set pending_rejection = "2024-05-25 00:00:00";
   ```
4. Close the review page and wait a few minutes for the reviewer lock to expire.
5. Run `make shell`, then `./manage.py auto_reject`
6. Reopen the review page, you should see something like this:
Screenshot 2024-05-28 at 14 13 20

### DIFF
--- a/src/olympia/constants/activity.py
+++ b/src/olympia/constants/activity.py
@@ -1015,6 +1015,31 @@ class NEEDS_HUMAN_REVIEW_CINDER(NEEDS_HUMAN_REVIEW_AUTOMATIC):
     review_queue = True
 
 
+class AUTO_REJECT_VERSION_AFTER_DELAY_EXPIRED(_LOG):
+    # takes add-on, version, reviewtype
+    id = 189
+    action_class = 'reject'
+    format = _('{addon} {version} rejected automatically after delay expired.')
+    short = _('Rejected automatically after delay expired')
+    keep = True
+    review_email_user = True
+    review_queue = True
+    reviewer_review_action = True
+    cinder_action = DECISION_ACTIONS.AMO_REJECT_VERSION_ADDON
+
+
+class AUTO_REJECT_CONTENT_AFTER_DELAY_EXPIRED(_LOG):
+    id = 190
+    action_class = 'reject'
+    format = _('{addon} {version} content rejected automatically after delay expired.')
+    short = _('Content rejected automatically after delay expired')
+    keep = True
+    review_email_user = True
+    review_queue = True
+    reviewer_review_action = True
+    cinder_action = DECISION_ACTIONS.AMO_REJECT_VERSION_ADDON
+
+
 LOGS = [x for x in vars().values() if isclass(x) and issubclass(x, _LOG) and x != _LOG]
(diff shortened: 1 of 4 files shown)


wagnerand:
Requesting review from @eviljeff specifically to consider any side effects for DSA and the cinder integration.

src/olympia/constants/activity.py
```diff
@@ -1015,6 +1015,31 @@ class NEEDS_HUMAN_REVIEW_CINDER(NEEDS_HUMAN_REVIEW_AUTOMATIC):
     review_queue = True
 
 
+class AUTO_REJECT_VERSION_AFTER_DELAY_EXPIRED(_LOG):
+    # takes add-on, version, reviewtype
+    id = 189
+    action_class = 'reject'
+    format = _('{addon} {version} rejected automatically after delay expired.')
+    short = _('Rejected automatically after delay expired')
```
eviljeff, line 1023: These `short` descriptions are a little non-short 😬 
I don't have a great alternative... is "Rejected automatically after expiration" too vague? (and only a little shorter anyway).
Or to reword completely to match how we label the delayed rejection event, "Scheduled automatic rejection"?

eviljeff: The other strings to be similarly adjusted if you agree.

eviljeff: Semi-nit comments I guess, but I can't remember everywhere we expose `short` as a string in the UI.

eviljeff (approved):

src/olympia/constants/activity.py
wagnerand: I like "Scheduled automatic rejection", but when I added it, I started to wonder if this can be read as "[someone] Scheduled [an] automatic rejection".  If we follow the other labels ("Rejection scheduled"), how about we use "Scheduled rejection due" or similar? What's a good way to express that?

wagnerand: After some discussion on slack, we will be moving back to the original copy.

wagnerand:
Reverted to initial commit, as discussed on slack (cc @diox as Andrew is out)

PR #22296, opened 2024-05-28
Hard-code super user name and email

wagnerand:
Fixes https://github.com/mozilla/addons/issues/14811 by hard-coding the superuser name, and more importantly, email, to always use a mozilla.com address.


scripts/setup.py
```diff
@@ -98,12 +91,8 @@ def get_docker_tag():
     {
         'DOCKER_TAG': docker_tag,
         'HOST_UID': get_value('HOST_UID', os.getuid()),
-        'SUPERUSER_EMAIL': get_value(
-            'SUPERUSER_EMAIL', git_config('user.email', 'admin@mozilla.com')
-        ),
-        'SUPERUSER_USERNAME': get_value(
-            'SUPERUSER_USERNAME', git_config('user.name', 'admin')
-        ),
+        'SUPERUSER_EMAIL': get_value('SUPERUSER_EMAIL', 'local_admin@mozilla.com'),
```
KevinMind, line 94: FWIW, if you set your email in the .env file, it will remain at the email you set no matter if you rerun `make setup` or `make up`.

The script prioritizes existing values in the .env, and values specified in the environment over the default.

So, technically, do we need to change this code at all? I don't mind changing it personally but just curious if you tried this and still had issues.

diox: The argument he made in the issue is that if you do the initial setup without knowing this, the default superuser that gets created is useless because it doesn't have a `@mozilla.com` email so it can never log in. Sure, you can fix it later but it's too late (and I don't think we want to override any existing user data if the env variable changes)

KevinMind: @diox is there a reason not to hard code the values then? Flipping it around, if we have specific requirements for the superuser email.. we could just hard code it in the make file and remove these two variables entirely..

@wagnerand do you have an opinion?

wagnerand: > Sure, you can fix it later but it's too late (and I don't think we want to override any existing user data if the env variable changes)

In addition to that, you first have to know what the issue is and how to fix it. We haven't documented either atm (afaik), and even if so, one would have to find that first. It would be better if it worked out of the box, I don't see a major downside to it.

I don't have a strong opinion where to hard-code the values.

diox: Hardcoding those to a `mozilla.com` email is fine with me. We likely don't need the variables, IMHO anyone that wants a custom admin user can just manually create/edit one in a shell or whatever.

wagnerand: I removed the variables. @diox r?

KevinMind (approved):
LGTM

PR #22299, opened 2024-05-30
Add configurable COMPOSE_FILE to setup flow.

KevinMind:
Relates to: [AMOENG-599](https://mozilla-hub.atlassian.net/browse/AMOENG-599)

### Description

This PR adds COMPOSE_FILE to the .env setup flow for running the project. This allows users to configure how the project runs in a predictable way.

### Context

You could already use COMPOSE_FILE (built in feature of docker compose) but our `make setup` flow doesn't consider this variable when managing the .env file and so it could be removed. adding it explicitly to the flow ensures a sensible default is maintained and that any user overrides are not erased unintentionally.

### Testing

This PR enables several new features locally.

1. Run web/worker containers without mounting local files. This is useful if you want to test a remote image and not mount your local file system.

```bash
DOCKER_DIGEST=<valid digest> COMPOSE_FILE=docker-compose.yml:docker-compose.ci.yml make up
```

This is the "default" behavior for CI preventing any host files in the runner from influencing the tests.

This will run the digest you pass and not mount local files. You can verify this by creatin a new file on the host, and checking it does NOT exist in the container.

2. Add additional override/private compose file.

Similarly to the above you can specify a COMPOSE_FILE environment variable either in the cli or in the .env file itself to specify additional compose files to add to your configuration. This is useful for those wanting to run `docker-compose.private.yml` @wagnerand CC please verify this works as expected.

### DIFF
--- a/.github/actions/run-docker/action.yml
+++ b/.github/actions/run-docker/action.yml
@@ -19,7 +19,7 @@ inputs:
   compose_file:
     description: 'The docker-compose file to use'
     required: false
-    default: 'docker-compose.yml'
+    default: 'docker-compose.yml:docker-compose.ci.yml'
 runs:
   using: 'composite'
   steps:
--- a/.github/workflows/extract-locales.yml
+++ b/.github/workflows/extract-locales.yml
@@ -19,6 +19,7 @@ jobs:
       - name: Extract Locales
         uses: ./.github/actions/run-docker
         with:
+          compose_file: docker-compose.yml
           run: |
             make extract_locales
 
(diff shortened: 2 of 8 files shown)


diox (approved):
> Fixes: [AMOENG-599](https://mozilla-hub.atlassian.net/browse/AMOENG-599)

Careful, this was split from the other PR and is not directly fixing this (it's fine since we don't have automation closing JIRA tickets from GitHub PRs, but still...)

KevinMind:
Updated.

PR #22305, opened 2024-05-31
Queue tasks before AutoReject'ing (like AutoApprove)

eviljeff:
fixes mozilla/addons#14818

I had to move `test_full_run` to a `TransactionTestCase` to make it work now, but I didn't add any specific new test - I couldn't see any similar test specifically checking for the order of tasks being executed for auto reject so I guess being in `TransactionTestCase` is enough?

Current status: tests not passing 😒

### DIFF
--- a/src/olympia/reviewers/management/commands/auto_reject.py
+++ b/src/olympia/reviewers/management/commands/auto_reject.py
@@ -4,6 +4,13 @@ from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.db import transaction
 
+from post_request_task.task import (
+    _discard_tasks,
+    _send_tasks_and_stop_queuing,
+    _start_queuing_tasks,
+    _stop_queuing_tasks,
+)
+
 import olympia.core.logger
 from olympia import amo
 from olympia.abuse.models import CinderJob
@@ -58,7 +65,6 @@ class Command(BaseCommand):
             .order_by('id')
         )
 
-    @transaction.atomic
     def reject_versions(self, *, addon, versions, latest_version):
         """Reject specific versions for an addon."""
         if self.dry_run:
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/management/commands/auto_approve.py
```diff
@@ -188,6 +188,7 @@ def process(self, version_id):
             # has been rolled back, or they may not have been processed by the
             # on commit handler yet).
             _stop_queuing_tasks()
+            _discard_tasks()
```
diox, line 191: The comment above is no longer accurate with that addition and needs to be updated.

src/olympia/reviewers/management/commands/auto_reject.py
```diff
@@ -97,46 +103,68 @@ def reject_versions(self, *, addon, versions, latest_version):
...
+                    addon=addon, versions=versions, latest_version=latest_version
+                )
         finally:
             # Always clear our lock no matter what happens.
             clear_reviewing_cache(addon.pk)
+            # Stop post request task queue before moving on (useful in tests to
+            # leave a fresh state for the next test. Note that we don't want to
+            # send or clear queued tasks (they may belong to a transaction that
+            # has been rolled back, or they may not have been processed by the
+            # on commit handler yet).
```
diox, line 165: As in `auto_approve.py` above that comment is no longer accurate now that we're also discarding any tasks left in the queue.

diox (approved):

src/olympia/reviewers/management/commands/auto_reject.py
eviljeff: https://github.com/mozilla/addons-server/pull/22305/commits/a2f16dbb07048d6fde39ce50b2547805cc9199fb

PR #22306, opened 2024-05-31
Refactor CI workflow and enable parallelization

KevinMind:
Fixes: https://github.com/mozilla/addons/issues/14821

### Description

This pull request refactors the CI workflow by renaming the `verify-docker-image` job to `CI` and enabling better parallelization of tests. 

### Context


src/olympia/devhub/tests/test_forms.py
```diff
@@ -1137,14 +1137,14 @@ def test_cannot_create_new_tags(self):
...
     def test_tags_limit(self):
-        extra = Tag.objects.count() - amo.MAX_TAGS
-        data = {**self.data, 'tags': [tag.tag_text for tag in Tag.objects.all()]}
+        data = {**self.data, 'tags': []}
         form = forms.AdditionalDetailsForm(
             data=data, request=self.request, instance=self.addon
         )
         assert not form.is_valid()
-        assert form.errors['tags'] == [f'You have {extra} too many tags.']
+        assert form.errors['tags'] == [f'You have {1} too many tags.']
```
diox, line 1147: This test has always been weird, not sure how it was passing on CircleCI, but I don't think we should patch `MAX_TAGS` to make it pass, and certainly not to a negative value.

How about just adding at the top:
```python
    for x in range(0, amo.MAX_TAGS + 2):
        Tag.objects.create(tag_text=f'tag{x}')
```

And not having any other modifications ?

KevinMind: It must have been running before other tests which depend on the number of tags that exist so that it could "determine" the correct amount. Not totally sure myself.

I don't have a strong opinion though I don't really see the issue with patching the setting which singularly controls if the test passes or not. I think it is logically the same as your suggestion, just doesn't require writing any data, which I think is better.

But since my opinion is weakly held, I'll happily go for any approach that solves the issue of flakiness.

diox: My issue with patching the constant to a negative value is that it's an unrealistic scenario. The constant is... a constant, and we should test behavior under normal circumstances. You're modifying the test in a way that no tags are submitted at all in the post data, which I would argue is not a very useful test.

KevinMind: Fixed it according to your original suggestion.

.circleci/config.yml
```diff
@@ -654,17 +100,8 @@ workflows:
...
-      - assets
-      - codestyle
-      - devhub
-      - main
       # Uncomment if you want to test the docker build
-      # - build-image
-      - reviewers-and-zadmin
-      - es-tests
-      - localization
+      - build-image
```
diox, line 104: Don't forget to re-comment this

diox (approved):

.circleci/config.yml
KevinMind: Cannot wait to remove that job so I don't have to keep enabling/disabling it.

PR #22314, opened 2024-06-04
More efficient utilization of github action cache

KevinMind:
Fixes: https://github.com/mozilla/addons/issues/14830

### Description

Enables more efficient usage of github action cache (which has a 10gb limit) by:
- reducing the number of builds and instead sharing images across CI jobs
- removing the need to cache ./deps by instead using a `DOCKER_TARGET` variable to include dev dependencies in a CI build

### Context

Both actions independently improve the cache usage, currently our biggest cache blobs are from larger dependency stages and from the cache-deps action. Since we are still including dependencies in the docker build, the dependency blobs will remained cached, however, by deduplicating the build step in our CI.yml workflow, we reduce the number of builds and therefore the maximum number of cached layers that can be pushed.

### Testing

You can expect much easier reproducability of CI workflows as a result of this PR. Try it out by running 1:1 ci builds locally.
- Click into one of the test jobs
- FInd the action that is using run-docker
- Copy the environment variables in the console to your local .env
- run `make up`
- Now you can either copy paste the heredoc command to execute the command in the container, or use `make shell` and copy without the heredoc. (note: if you use the heredoc, locally, you should add the `-T` argument.

```bash
  cat <<'EOF' | docker compose exec --user olympia -T web sh
    make extract_locales
  EOF
```

OR

```bash
make shell
```

and then 

```bash
make extract_locales
```


Dockerfile
```diff
@@ -181,11 +181,18 @@ COPY --from=assets --chown=olympia:olympia ${HOME}/site-static ${HOME}/site-stat
 # inside the docker image, thus it's copied there.
 COPY version.json /app/version.json
 
+# We have to reinstall olympia after copying source
+# to ensure the installation syncs files in the src/ directory
+RUN make -f Makefile-docker update_deps_olympia
+
+FROM sources as development
+
+# Copy dependencies from `pip_production`
```
diox, line 190:
```suggestion
# Copy dependencies from `pip_development`
```

diox (approved):

PR #22324, opened 2024-06-05
Assay Links Added for Yara and Customs, opens the flagged add-on version & specific file.  Links to code.AMO should direct to the corresponding file and line in VS Code

chrstinalin:
This PR is to create the links next to the existing code-manager ones for proof-of-concept purposes. These links should eventually fully replace code-manager. 

**Changes**
- Added Assay link to Yara results
- Modified existing Assay link to use helper.

image

### DIFF
--- a/src/olympia/reviewers/templatetags/assay.py
+++ b/src/olympia/reviewers/templatetags/assay.py
@@ -1,0 +1,10 @@
+from django import template
+
+register = template.Library()
+
+@register.simple_tag
+def assay_url(addon_guid, version, file=None):
+    url = f'vscode://mozilla.assay/review/{addon_guid}/{version}'
+    if file:
+        url = f'{url}?path={file}'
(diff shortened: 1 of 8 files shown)


src/olympia/reviewers/templatetags/assay.py
```diff
@@ -0,0 +1,12 @@
+from django import template
+
+
+register = template.Library()
+
+
+@register.simple_tag
+def assay_url(addon_guid, version, file=None):
```
eviljeff, line 8: I think these args can be a less ambiguous (is version a Version object?) if it's a version string, `version_string` is clearer. Similarly, `file` isn't a File object, its a path and filename, so `path` or `filepath`?

eviljeff (changes requested):

eviljeff (approved):
r+

PR #22327, opened 2024-06-06
Fix inaccuracies in email sent to developers when rejecting versions being blocked

diox:
Fixes mozilla/addons#14831

### DIFF
--- a/src/olympia/abuse/templates/abuse/emails/CinderActionRejectVersion.txt
+++ b/src/olympia/abuse/templates/abuse/emails/CinderActionRejectVersion.txt
@@ -6,7 +6,7 @@ Our review found that your content violates the following Mozilla policy or poli
 
 Affected versions: {{ version_list }}
 
-Based on that finding, those versions of your {{ type }} have been disabled on {{ target_url }} and are no longer available for download from Mozilla Add-ons, anywhere in the world. {% if not is_addon_being_blocked and not target.block %}Users who have previously installed those versions will be able to continue using them.{% endif %}
+Based on that finding, those versions of your {{ type }} have been disabled on {{ target_url }} and are no longer available for download from Mozilla Add-ons, anywhere in the world. {% if is_addon_being_blocked %}In addition, users who have previously installed those versions won’t be able to continue using them.{% else %}Users who have previously installed those versions will be able to continue using them.{% endif %}
 
-{% if not is_addon_being_disabled and not target.is_disabled %}You may upload a new version which addresses the policy violation(s).{% endif %}
+{% if not is_addon_disabled %}You may upload a new version which addresses the policy violation(s).{% endif %}
 {% endblock %}
--- a/src/olympia/abuse/templates/abuse/emails/CinderActionRejectVersion.txt
+++ b/src/olympia/abuse/templates/abuse/emails/CinderActionRejectVersion.txt
@@ -6,7 +6,7 @@ Our review found that your content violates the following Mozilla policy or poli
 
 Affected versions: {{ version_list }}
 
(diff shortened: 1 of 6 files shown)


src/olympia/abuse/templates/abuse/emails/CinderActionRejectVersion.txt
```diff
@@ -6,7 +6,7 @@ Our review found that your content violates the following Mozilla policy or poli
 
 Affected versions: {{ version_list }}
 
-Based on that finding, those versions of your {{ type }} have been disabled on {{ target_url }} and are no longer available for download from Mozilla Add-ons, anywhere in the world. Users who have previously installed those versions will be able to continue using them.
+Based on that finding, those versions of your {{ type }} have been disabled on {{ target_url }} and are no longer available for download from Mozilla Add-ons, anywhere in the world. {% if not is_addon_being_blocked and not target.block %}Users who have previously installed those versions will be able to continue using them.{% endif %}
```
diox, line 9: I've added a condition on `not target.block` to cover notifications being sent by a rejection if the add-on is _already_ blocked. Technically we don't check whether or not the block that exists covers the affected versions, but since we only use that condition to hide that extra sentence I think it's ok.

diox: Copy is being validated by Legal, I'm missing the `{% else %}Users who have previously installed those versions won’t be able to continue using them.{% endif %}`.

src/olympia/abuse/templates/abuse/emails/CinderActionRejectVersion.txt
```diff
@@ -6,7 +6,7 @@ Our review found that your content violates the following Mozilla policy or poli
...
 
-You may upload a new version which addresses the policy violation(s).
+{% if not is_addon_being_disabled and not target.is_disabled %}You may upload a new version which addresses the policy violation(s).{% endif %}
```
eviljeff, line 11: imo it'd be better if we weren't accessing `.block` and `.is_disabled` directly in the template.  
e.g. `'is_addon_blocked': log_entry.details.get('is_addon_being_blocked') or target.block`

eviljeff: But... we'd be always making that query, even though we only need it for the reject version template.

src/olympia/abuse/templates/abuse/emails/CinderActionRejectVersion.txt
eviljeff: That'd mean any block on any version of an add-on would lead to the email informing the developer that "Users who have previously installed those versions won’t be able to continue using them", which won't always be correct.

diox: True, this only works if we don't have the `else` clause in the mail (we didn't at the time I wrote the patch). I'm going to remove it.

src/olympia/abuse/templates/abuse/emails/CinderActionRejectVersion.txt
diox: I'll need to remove the `.block` condition because of that `else` clause anyway, so maybe I can do that just for the disabled part (and call the corresponding variable `is_addon_disabled`)

src/olympia/abuse/templates/abuse/emails/CinderActionRejectVersion.txt
diox: https://github.com/mozilla/addons-server/pull/22327/commits/2b7ff0ece9d758ae50e1d53f2ebef430cda81fac

eviljeff (approved):

PR #22328, opened 2024-06-06
Expose reviewers validation result API through JWT auth and document it

diox:
Fixes mozilla/addons#14833

## Context

That endpoint existed already but wasn't documented and was only available through the SessionID auth, which is the default, and not through JWT auth. That's because it was only needed for code-manager until now, which uses session auth. It's now needed for [Assay](https://github.com/mozilla/assay) which uses JWT so I've added the relevant authentication classes.

## Testing

See unit tests, we have used a similar method of duplicating the class and changing the `client_class` for other API tests, makes things easier even it's a little wasteful.

It's painful to test locally and I would just trust the unit tests, but for anyone adventurous, you'd need to:
- Upload an add-on, then grab the add-on and file id (from a shell is easier, i.e. `Addon.objects.latest('pk').pk` and `File.objects.latest('pk').pk`)
- Log in as a reviewer, generate API keys in devhub (need to look up fake emails to grab the confirmation link)
- Using those API keys to generate a JWT-authenticated request against `http://olympia.test/api/v5/reviewers/addon/{addon_id}/file/{file_id}/validation/` following our API docs

### DIFF
--- a/docs/topics/api/reviewers.rst
+++ b/docs/topics/api/reviewers.rst
@@ -9,9 +9,11 @@ Reviewers
     These APIs are not frozen and can change at any time without warning.
     See :ref:`the API versions available<api-versions-list>` for alternatives
     if you need stability.
-    The only authentication method available at
-    the moment is :ref:`the internal one<api-auth-internal>`, except for
-    :ref:`validation results<reviewers-validation>`.
+
+    The only authentication method available for these APIs is
+    :ref:`the internal one<api-auth-internal>`, except for the
+    :ref:`validation results<reviewers-validation>` endpoint, which allows both
+    internal and :ref:`external auth<api-auth>`.
 
 ---------
 Subscribe
--- a/docs/topics/api/reviewers.rst
+++ b/docs/topics/api/reviewers.rst
@@ -10,7 +10,8 @@ Reviewers
     See :ref:`the API versions available<api-versions-list>` for alternatives
     if you need stability.
     The only authentication method available at
-    the moment is :ref:`the internal one<api-auth-internal>`.
+    the moment is :ref:`the internal one<api-auth-internal>`, except for
+    :ref:`validation results<reviewers-validation>`.
 
 ---------
 Subscribe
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -21,6 +21,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Build Docker image
+        if: "!github.event.pull_request.head.repo.fork"
```
KevinMind, line 24: rebase.

docs/topics/api/reviewers.rst
```diff
@@ -10,7 +10,8 @@ Reviewers
     See :ref:`the API versions available<api-versions-list>` for alternatives
     if you need stability.
     The only authentication method available at
-    the moment is :ref:`the internal one<api-auth-internal>`.
+    the moment is :ref:`the internal one<api-auth-internal>`, except for
```
KevinMind, line 13: This is a confusing sentence. There are two auth methods, but the sentence starts with "the only one available" can you rephrase this?

src/olympia/reviewers/views.py
```diff
@@ -1197,6 +1201,10 @@ def clear_pending_rejections(self, request, **kwargs):
         detail=True,
         methods=['get'],
         permission_classes=[AllowAnyKindOfReviewer],
+        authentication_classes=[
```
KevinMind, line 1204: Have you explored what this would look like in the swagger docs? I know we kind of dropped the project of migrating api docs to swagger but could be an interesting exercise to add some anotations to the view class to see if you can explain what you do in the .rst file directly in the swagger.

KevinMind:
Curious @diox did you go through the steps you outlined in the test section? If so, I would be curious to try (another adventurous thing). What if you exported the database, which should include the addon, the files, the session etc and zipped it.

I could try to load that state onto my local and verify without having to repeat the steps. Could be fun. Don't have to, but would be curious what happens.

docs/topics/api/reviewers.rst
diox: Maybe 

```
The only authentication method available for these APIs is
:ref:`the internal one<api-auth-internal>`, except for the
:ref:`validation results<reviewers-validation>` endpoint, which allows both
internal and :ref:`external auth<api-auth>`.
```

? Ultimately internal auth remains the only auth mechanism available for the others APIs described in that doc, validation is the exception...

src/olympia/reviewers/views.py
diox: I haven't. It's a bad one for swagger from the start because it doesn't use a serializer and directly returns a `JsonResponse`, so it doesn't automatically document the response (which admittedly only contains a single field). Furthermore it looks like our swagger implementation is not smart enough to know which authentication classes are available for each endpoint, so it was already suggesting to use JWT for that endpoint...

I could add a docstring though.

diox:
> Curious @diox did you go through the steps you outlined in the test section?

I didn't, I was trusting the unit tests (and the fact that this is the standard way of adding JWT auth to our APIs). I shared the PR with @chrstinalin on slack so she could test locally with her build of assay that required this change.

KevinMind (approved):
Didn't test locally.

PR #22331, opened 2024-06-06
Add version suffix and enable push on master

KevinMind:
Follow up to https://github.com/mozilla/addons-server/pull/22330 fix build on master push

### DIFF
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -65,7 +65,7 @@ runs:
       id: meta
       uses: docker/metadata-action@v5
       with:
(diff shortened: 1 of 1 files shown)


.github/actions/build-docker/action.yml
```diff
@@ -75,7 +68,7 @@ runs:
         # Hard coding our dockerhub imnage name
```
diox, line 68: While you're at it...

```suggestion
        # Hard coding our dockerhub image name
```

diox (approved):
s/prefix/suffix on the commit message

PR #22332, opened 2024-06-07
Ditch the /deps volume

KevinMind:
Follow up of: https://github.com/mozilla/addons/issues/14821

### Description

Remove the ./deps:/deps volume as an unnecessary artifact of the previous build architecture. Currently dependencies are packaged with the docker image and mounting from the host is both unnecessary and complex to maintain.

Inspecting files using the local ./deps can have some limited benefits in some cases so a `docker-compose.deps.yml` file is retained to do so if a developer wants to, but it is no longer the default behavior to mount the ./deps directory.

### Testing

No changes to existing environments, once rebased any files in the ./deps directory will be ignored and should be cleaned up by developers on their machines.

#### Running with ./deps mount

To mount with deps run

```bash
make up COMPOSE_FILE=docker-compose.yml:docker-compose.deps.yml
```

Expect mounting from ./deps to occur.

### DIFF
--- a/docker-compose.deps.yml
+++ b/docker-compose.deps.yml
@@ -1,0 +1,9 @@
+services:
+  web: &web
+      volumes:
+      - ./deps:/deps
+      - ./package.json:/deps/package.json
+      - ./package-lock.json:/deps/package-lock.json
+
+  worker:
+    <<: *web
--- a/docker-compose.deps.yml
+++ b/docker-compose.deps.yml
@@ -1,3 +1,6 @@
+# Using this file enables a mount of ./deps to /deps in the web/worker containers.
+# This allows the /deps folder to be exposed on the host machine enabling editing of those files
+# from the host. It is disabled by default.
 services:
   web: &web
       volumes:
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -15,16 +15,19 @@ fi
 
 OLYMPIA_USER="olympia"
 
+function get_olympia_uid() { echo "$(id -u "$OLYMPIA_USER")"; }
+function get_olympia_gid() { echo "$(id -g "$OLYMPIA_USER")"; }
+
 if [[ -n "${HOST_UID:-}" ]]; then
   usermod -u ${HOST_UID} ${OLYMPIA_USER}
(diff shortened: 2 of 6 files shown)


docker/entrypoint.sh
```diff
@@ -15,16 +15,19 @@ fi
...
+
 if [[ -n "${HOST_UID:-}" ]]; then
   usermod -u ${HOST_UID} ${OLYMPIA_USER}
   echo "${OLYMPIA_USER} UID: ${OLYMPIA_UID} -> ${HOST_UID}"
-fi
 
-uid=$(id -u $OLYMPIA_USER)
-gid=$(id -g $OLYMPIA_USER)
+  # Ensure the olympia user has access to the /deps directory
+  time chown -R "$(get_olympia_uid):$(get_olympia_gid)" /deps
```
diox, line 26: As a follow-up to speed things up (*): If we stop mounting /deps maybe we could skip this and install deps as `root` when building the image - or any user really, including the uid=9500 - as long as the right `umask` is set at install to allow reading and executing by others.

The only benefits of having the `olympia` user actually owning those files are:
- Consistency with prod behavior (where deps are installed with `olympia` user with uid 9500 atm)
- Ability to edit the files in deps when entering the container as the `olympia` user

(*) I don't think we want to block the PR on this. Not sure what the implications would be for users of the `docker-compose.deps.yml` file. Plus, now that `/deps` is purely internal, this seems to be relatively fast.

docker-compose.deps.yml
```diff
@@ -0,0 +1,9 @@
+services:
```
diox, line 1: Needs a quick comment at the top to explain what this file is and the fact that it's not used by default.

diox (approved):

docker/entrypoint.sh
KevinMind: I've tried, haven't found a solution, but there could be one eventually.

Currently the chown takes around 30-45 seconds. Not catastrophically long, faster than the update_deps did. But yes, it would be nice to remove the need for this, we just have to figure out how to enable olympia user access to host mounted files and container created files that would have different owners. 

So far, when I tried to share ownership via group, pip and npm complained about permissions but happy to see a PR if you have a solution in mind.

PR #22344, opened 2024-06-11
Fix-pip-packages

KevinMind:
FIxes: https://github.com/mozilla/addons/issues/14840

### DIFF
--- a/scripts/check_pip_packages.sh
+++ b/scripts/check_pip_packages.sh
@@ -70,7 +70,7 @@ function check_package() {
     actual_name=$(get_package_name $actual_package)
 
     if [ "$expected_name" != "$actual_name" ]; then
-        fail "Package missing. Expected ${expected_name}. Recieved ${actual_name}"
+        fail "Package missing. Expected ${expected_name}. Received ${actual_name}"
     fi
 
     expected_version=$(get_package_version $expected_package)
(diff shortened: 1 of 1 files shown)


scripts/check_pip_packages.sh
```diff
@@ -70,14 +70,14 @@ function check_package() {
     actual_name=$(get_package_name $actual_package)
 
     if [ "$expected_name" != "$actual_name" ]; then
-        fail "Package $expected_package is missing."
+        fail "Package missing. Expected ${expected_name}. Recieved ${actual_name}"
```
diox, line 73:
```suggestion
        fail "Package missing. Expected ${expected_name}. Received ${actual_name}"
```

diox (commented):
Is this PR missing something ? It fixes the error message but not the underlying problem

KevinMind:
This PR fixes the error message, once this is merged, the error in the original PR will be clear that the root cause is a naming conflict. In our prod.txt we have 'prompt-tools' but pip is actually installing 'prompt_tools' we've seen that kind of issue before and merging this will make those issues more obvious.

diox:
Probably shouldn't mark the parent ticket as fixed if it's just to help us debug the problem further/make it more obvious though, as the issue remains.

diox (approved):

diox:
In https://github.com/mozilla/addons-server/pull/22341 ? The error message might be misleading but it's a legit failure, and you commented in there that it was a `/deps` permission problem ?

KevinMind:
> In #22341 ? The error message might be misleading but it's a legit failure, and you commented in there that it was a `/deps` permission problem ?

That was my theory at the time. Doesn't seem to be the issue though. The issue is definitely the package name conflict. Either way, this PR should land and once rebased the dependabot PR will give us an actionable error that can be fixed.

PR #22352, opened 2024-06-11
Merge docs/extract locales to CI workflow

KevinMind:
Split from: https://github.com/mozilla/addons-server/pull/22318

### Description

This PR adds a context job to specify the context in which CI.yml is running. The job can run either on a fork or the main repo, and can run on a push/pull_request or workflow_dispatch event. Each combination can result in different constraints in terms of how the job should run. This new job sets that context and displays the github context object in the logs for visibility

### Context

See what I did there?

### Testing

Forked PR: https://github.com/mozilla/addons-server/actions/runs/9471961075/job/26096429518?pr=22353
Forked push: https://github.com/KevinMind/addons-server/actions/runs/9471982768/job/26096487673

Dependabot PR: https://github.com/mozilla/addons-server/actions/runs/9483055672/job/26129267103?pr=22351 (I co-authored the PR so it returned false for is dependabot, but also that means it has access to secrets so this is expected behaviour. will have to test solo dependabots afteer merging..)

- rebase this commit on a fork both pushing to master and on a pull request
- run this commit on master (after merging)
- rebase this commit on a dependabot PR


.github/workflows/ci.yml
```diff
@@ -7,14 +7,69 @@ on:
...
+            # repository on a pull request refers to the base which is always mozilla/addons-server
+            is_fork=${{ github.event.pull_request.head.repo.fork }}
+          else
+            # Repository on a push refers to the head which would be the fork
+            is_fork="${{ github.event.repository.fork }}"
+          fi
+
+          echo "github.event.repository.full_name ${{ github.event.repository.full_name }}"
+
+          # Represetns a push to the master branch
```
diox, line 60:
```suggestion
          # Represents a push to the master branch
```

.github/workflows/ci.yml
```diff
@@ -7,14 +7,69 @@ on:
...
+        run: |
+          cat <<EOF
+          ${{ toJSON(github) }}
+          EOF
+      - name: Set context
+        id: context
+        shell: bash
+        run: |
+          is_push_master="${{ github.event_name == 'push' && github.ref_name == 'master' }}"
+          is_dependabot="${{ startsWith(github.head_ref, 'dependabot')}}"
```
diox, line 48: Should we also check that we're not on a fork here ? It doesn't matter now, but it would avoid us ever accidentally considering a branch on someone's fork as a `dependabot` PR, in case we add more logic around `is_dependabot` later.

(My own workaround that only checked the `actor` name suffered from this issue and I didn't like that)

diox (approved):
Haven't re-tested this but it makes sense

.github/workflows/ci.yml
diox: I mean specifically to figure out the `is_dependabot` variable value. It should never be true in forks IMHO, regardless of how they have named their branches.

KevinMind: I disagree, we have 2 variables 1 tells you if it is a fork, one if it is a dependabot PR. If you want to know if it is a fork/dependabot, just combine the conditions, just like we do in the build.

I think adding an implicit filter for forks on is_dependabot makes the logic less clear.

KevinMind: Rebased this PR on a dependabot PR, let's see if the user gives us an answer https://github.com/mozilla/addons-server/pull/22351

KevinMind: Pushed a commit using the `github.event.pull_request.user.login` to match the dependabot user.. this should work.

KevinMind: I actually think your original way is the best way, it is in the official documentation. Combining that with the new logic for forks, we should be good to go. https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions

Wdyt?

diox: But as it stands the condition is not specific enough, it does not guarantee that it's a dependabot PR: Any branch (from a fork or not) can start with `dependabot`, we could accidentally call a branch `dependabot-fix` or `dependabot-config` ourselves...

At the very least we should make that condition true if the branch name starts with `dependabot/` (with the slash) to avoid this IMHO.

diox: The problem with my original condition is that it prevents the `Build Docker image` job from running on commits from dependabot in `master`. Using the branch named worked around that.

I'm fine with going back to checking the `actor` but we need another check so that `Build Docker image` runs on `master` from the main repos no matter what

KevinMind: Yes, the check in the build step needs to be made more specific. I also found while testing this it is a bit confusing if the build doesn't run, that you don't know exactly why. So I've added a step that expalins the workflows reasoning as it goes. Wdyt?

KevinMind: Found a better way to check for default branch and split the branch we are on from the event type as they can be considered independently in many cases. I will retest everything as it is siginficantly changed.

.github/workflows/ci.yml
```diff
@@ -23,8 +86,33 @@ jobs:
...
+          is_fork="${{ needs.context.outputs.is_fork == 'true' }}"
+          is_default_branch="${{ needs.context.outputs.is_default_branch == 'true' }}"
+          is_dependabot="${{ needs.context.outputs.is_dependabot == 'true' }}"
+
+          # Default behaviour is to build images for any CI.yml run
+          should_build="true"
+
+          # Never run the build on a fork. Forks lack sufficient permissions
+          # to access secrets or push artifacts
+          if [ $is_fork == 'true' ]; then
```
diox, line 102:
```suggestion
          if [ "$is_fork" == 'true' ]; then
```

I think ? I can never remember the subtle differences between the various ways to do conditions in bash.

.github/workflows/ci.yml
```diff
@@ -23,8 +86,33 @@ jobs:
...
+            should_build="false"
+          fi
+
+          # Dependabot PRs are treated as if they are from forks (see above)
+          if [ $is_dependabot = 'true' && $is_default_branch = 'false']; then
```
diox, line 107: As above but also double equal sign for consistency 

```suggestion
          if [ "$is_dependabot" == 'true' && "$is_default_branch" == 'false']; then
```

?

PR #22360, opened 2024-06-12
Listen to all release events

KevinMind:
Following https://github.com/mozilla/addons-server/pull/22359

For some reason, release events are not triggering the workflow, so to simplify I am removing all filters and just triggering on any release event. We can figure out how to narrow down later.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,18 +2,6 @@ name: Release
 
 on:
   release:
-    types: [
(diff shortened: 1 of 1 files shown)


KevinMind:
@diox there goes the assertNumQueries again.. looks like it is flaky even without test duration splitting https://github.com/mozilla/addons-server/actions/runs/9486316072/job/26140465435?pr=22318

PR #22363, opened 2024-06-12
Fix flaky assertNumQueries test in TestAddonAdmin

diox:
Depending on test order content types were not preloaded yet when running the test_query_count().

Helps https://github.com/mozilla/addons/issues/14821

### DIFF
--- a/src/olympia/addons/tests/test_admin.py
+++ b/src/olympia/addons/tests/test_admin.py
@@ -2,6 +2,7 @@ from datetime import datetime, timedelta
 
 from django.conf import settings
 from django.contrib import admin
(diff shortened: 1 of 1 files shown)


KevinMind:
I guess this is hard to test because it is so seldom occurring. Did you say you can verify it locally?

KevinMind (approved):
I would like to test this locally, if you have a way to ensure the breakage "would" happen that would be great. Otherwise we will just have to wait and see if it comes back.

diox:
I verified locally yes. Running `TestAddonAdmin::test_query_count` alone (or first, but alone is easier) triggers the issue, I verified the patch fixes that.

PR #22364, opened 2024-06-13
Rename-release

KevinMind:
Relates to: https://github.com/mozilla/addons/issues/14823
Companion to: https://github.com/mozilla-sre-deploy/deploy-amo/pull/4

### Description

This PR introduces a `Release.yml` workflow that is triggered on release `published` events. This workflow currently builds our docker image using the tag that was linked to the release and will eventually be how we push our deployment image to a registry where it can then be shipped.

### Context

This job can only be triggered if it is included in the tag that creates the release (kind of weird gotcha in github releases)

This job doesn't currently push to a registry, though will likely eventually push directly to GAR.

This job doesn't currently have any logic to verify that the deployment is actually successful once the image is uploaded, but this could be done.

The workflow doesn't verify the image the way the ci.yml does. This should be considered in follow up work. Maybe integrate the two workflows, or find a way to triggere the ci.yml workflow as a part of the release. THere is too much to consider for here, how to make the release blocked by success, how to roll back if it fails etc. etc.

### Testing

You can test this by pushing a tag that has this commit and then publishing a release with that tag.


KevinMind:
[Release](github.com/mozilla/addons-server/releases/tag/rename-release-tag) caused workflow to [run](https://github.com/mozilla/addons-server/actions/runs/9499267953/job/26179806096)

.github/workflows/ci.yml
```diff
@@ -114,7 +117,7 @@ jobs:
           fi
 
           # Dependabot PRs are treated as if they are from forks (see above)
```
KevinMind, line 119: This is a more specific and accurate check.

.github/workflows/ci.yml
```diff
@@ -114,7 +117,7 @@ jobs:
...
-          if [[ "$is_dependabot" == 'true' && "$is_default_branch" == 'false' ]]; then
+          if [[ "$is_dependabot" == 'true' && "${{ github.event_name }}" == 'pull_request' ]]; then
```
KevinMind, line 120: This is a more specific check that we are running on a pull request, from the dependabot user

.github/workflows/ci.yml
```diff
@@ -63,24 +72,19 @@ jobs:
...
-          # Are we running on a push event
-          # Many jobs/steps only execute or change execution on push events
-          is_push="${{ github.event_name == 'push' }}"
 
           if [[ "${{ github.event_name }}" == 'pull_request' ]]; then
             # repository on a pull request refers to the base which is always mozilla/addons-server
             is_fork=${{ github.event.pull_request.head.repo.fork }}
           else
-            # Repository on a push refers to the head which would be the fork
+            # In most events, the epository refers to the head which would be the fork
```
KevinMind, line 80: the else statement applies to `workflow_dispatch`, `push` and `release` all of which share the same logic

KevinMind:
I'm going to merge this as it can only be tested on the default branch and does not significantly impact anything.

Issue #1741, opened 2016-02-23
CORS the API


andymckay:
We aren't sure where the admin API is going to be living. We do know that the disco pane is going to be on services.addons.mozilla.org, let's ensure the API is CORS'd so that's not going to be a problem.

In the marketplace we started off setting this explicitly on each API. I think CORS by default might be the way to go.

diox:
Is there any part of the *site* that we would not want to enable CORS for ? I'm thinking we could simply add it as a middleware, using https://github.com/ottoyiu/django-cors-headers/

andymckay:
On the marketplace I was always nervous about CORS'ing APIs that should only be used by the admin endpoints. But authentication should take care of that and I don't think the defence by depth is worth it.

diox:
We can add specific exceptions where needed. Do we have anything that lives on AMO, not just APIs, that we would not want to allow the outside world access to via CORS ?

andymckay:
I don't know if anyone can answer that question truthfully, there's a lot on AMO. Please limit it to the API if possible.

PR #22370, opened 2024-06-17
Handle abuse reports being associated with appeal jobs

eviljeff:
fixes mozilla/addons#1741 by handling abuse reports appearing at any level, rather than just assuming an appeal means reports only in the appealed job ... but in the meantime Cinder PM has queried if abuse reports being attached to appeal jobs is our preferred solution anyway.  It's unclear if a change in Cinder would land in time to make this patch redundant though.

Can be tested by first reproducing the issue in #1741 and then making a decision on the appeal job.

### DIFF
--- a/src/olympia/abuse/tests/test_models.py
+++ b/src/olympia/abuse/tests/test_models.py
@@ -972,6 +972,70 @@ class TestCinderJob(TestCase):
         ).exists()
         assert NeedsHumanReview.objects.filter(is_active=True).count() == 2
 
+    def test_resolve_job_appeal_with_new_report(self):
+        addon_developer = user_factory()
+        addon = addon_factory(users=[addon_developer])
+        appeal_job = CinderJob.objects.create(
+            job_id='999',
+        )
+        AbuseReport.objects.create(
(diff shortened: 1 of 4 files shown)


src/olympia/abuse/tests/test_models.py
```diff
@@ -1257,17 +1334,17 @@ def test_reporter_cant_appeal_appealed_decision(self):
             appellant_job=appeal_job,
         )
         self.decision.update(appeal_job=appeal_job)
-        # We should never end up in this situation where an AbuseReport is tied
-        # to a CinderJob from an appeal, but if that somehow happens we want to
-        # make sure it's impossible for a reporter to appeal an appeal.
+        # We can end up in this situation where an AbuseReport is tied
+        # to a CinderJob from an appeal, and if that somehow happens we want to
+        # make sure it's ossible for a reporter to appeal an appeal.
```
diox, line 1339:
```suggestion
        # make sure it's possible for a reporter to appeal an appeal.
```

diox (approved):

PR #22377, opened 2024-06-19
Replace post-request-task with transaction.on_commit on every delay/apply_async

diox:
## Description

According to [Django docs](https://docs.djangoproject.com/en/5.0/topics/db/transactions/#performing-actions-after-commit), `transaction.on_commit` should:
- only be called once the outer `transaction.atomic()` has been executed, guaranteeing that it works even if there are savepoints.
- be discarded on rollback

So it should be a drop-in replacement with the added benefit that it handles tasks and commands doing transactions outside of the request-response cycle.

This is the same approach celery has introduced in 5.4 with its [DjangoTask](https://docs.celeryq.dev/en/stable/reference/celery.contrib.django.task.html#celery.contrib.django.task.DjangoTask) except we keep our own class and apply that waiting for the transaction commit behavior to all tasks by default.

### Context

post-request-task is an app we wrote years ago to delay sending the tasks to celery broker until the current request sends the `request_finished` signal. The idea is that we're using django's `ATOMIC_REQUESTS` so that each request is wrapped in a transaction (unless it's specifically disabled for that view), so by waiting for that signal, we know the task we want to trigger is safe to execute, any database changes it's relying on will have been committed.

This works fine for requests, but doesn't do anything for management commands or tasks. Historically this hasn't been much of an issue, because we didn't trigger that many tasks from those, but lately it's become apparent we need to address that.

### Testing

Unit tests passing are not enough to prove this change works, because of 2 reasons:
- Our tests often expect tasks to be executed synchronously for convenience, so we use celery eager mode
- Django's `TestCase` wraps tests in its own transaction that is never committed and simply rolled back at the end of each test

Local testing should be done by looking at worker logs (`logs/docker-celery.log`) for the following scenarios to make sure the corresponding task(s) have been executed:
- Triggering `manage.py generate_error --celery --atomic`. You can repeat with or without `--atomic` and it shouldn't make a difference.
- Submitting a new add-on. This triggers a chord with our validation-related tasks, and should succeed and let you continue with your submission.
- Triggering `manage.py auto_approve` with some unreviewed versions. This should approve them, provided you don't have their review page opened in a tab and that the versions can be auto-approved (submitting a new unlisted one is the fastest way to do that)
- Replying as a reviewer in review page for an add-on. This should send a fake email. Note that you can also resolve cinder jobs if you have that set up.
- Making changes to an add-on name or description. That should trigger a reindex, which you can check by then searching for that change using the API or frontend.

Fixes mozilla/addons#14842

### DIFF
--- a/src/olympia/amo/celery.py
+++ b/src/olympia/amo/celery.py
@@ -61,8 +61,13 @@ class AMOTask(Task):
             args, kwargs = self._serialize_args_and_kwargs_for_eager_mode(
                 args=args, kwargs=kwargs, **options
             )
+            # In eager mode, immediately call original apply async as we are
+            # using eager mode for tests, where no transaction is ever actually
+            # committed so transaction.on_commit() is never called.
             return self.original_apply_async(args=args, kwargs=kwargs, **options)
         else:
+            # In normal mode, wait until the current transaction is committed
+            # to actually send the task.
             transaction.on_commit(
                 functools.partial(self.original_apply_async, *args, **kwargs)
             )
--- a/src/olympia/amo/celery.py
+++ b/src/olympia/amo/celery.py
@@ -8,14 +8,16 @@ is directly being run/imported by Celery)
 """
 
 import datetime
+import functools
 
 from django.core.cache import cache
+from django.db import transaction
 
 from celery import Celery, group
+from celery.app.task import Task
 from celery.signals import task_failure, task_postrun, task_prerun
 from django_statsd.clients import statsd
 from kombu import serialization
-from post_request_task.task import PostRequestTask
 
 import olympia.core.logger
 
@@ -23,10 +25,11 @@ import olympia.core.logger
 log = olympia.core.logger.getLogger('z.task')
 
 
-class AMOTask(PostRequestTask):
-    """A custom celery Task base class that inherits from `PostRequestTask`
-    to delay tasks and adds a special hack to still perform a serialization
-    roundtrip in eager mode, to mimic what happens in production in tests.
+class AMOTask(Task):
+    """A custom celery Task base class to always trigger tasks after the
+    current transaction has been committed, and also adds a special hack to
+    still perform a serialization roundtrip in eager mode, to mimic what
+    happens in production in tests.
 
     The serialization is applied both to apply_async() and apply() to work
     around the fact that celery groups have their own apply_async() method that
@@ -50,13 +53,21 @@ class AMOTask(PostRequestTask):
             args, kwargs = serialization.loads(data, content_type, content_encoding)
         return args, kwargs
 
+    def original_apply_async(self, args=None, kwargs=None, **extrakw):
+        return super().apply_async(args=args, kwargs=kwargs, **extrakw)
+
+    def delay(self, *args, **kwargs) -> None:
+        transaction.on_commit(functools.partial(super().delay, *args, **kwargs))
+
     def apply_async(self, args=None, kwargs=None, **options):
         if app.conf.task_always_eager:
             args, kwargs = self._serialize_args_and_kwargs_for_eager_mode(
                 args=args, kwargs=kwargs, **options
             )
 
-        return super().apply_async(args=args, kwargs=kwargs, **options)
+        transaction.on_commit(
+            functools.partial(self.original_apply_async, *args, **kwargs)
+        )
 
     def apply(self, args=None, kwargs=None, **options):
         if app.conf.task_always_eager:
--- a/src/olympia/amo/celery.py
+++ b/src/olympia/amo/celery.py
@@ -53,8 +53,8 @@ class AMOTask(Task):
             args, kwargs = serialization.loads(data, content_type, content_encoding)
         return args, kwargs
 
-    def original_apply_async(self, args=None, kwargs=None, **extrakw):
-        return super().apply_async(args=args, kwargs=kwargs, **extrakw)
+    def original_apply_async(self, *args, **kwargs):
+        return super().apply_async(*args, **kwargs)
 
     def apply_async(self, args=None, kwargs=None, **options):
         if app.conf.task_always_eager:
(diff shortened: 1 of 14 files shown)


src/olympia/amo/celery.py
```diff
@@ -50,13 +53,21 @@ def _serialize_args_and_kwargs_for_eager_mode(
             args, kwargs = serialization.loads(data, content_type, content_encoding)
         return args, kwargs
 
+    def original_apply_async(self, args=None, kwargs=None, **extrakw):
```
KevinMind: Can you add a comment why it would be necessary to define this method? What is the use case?

KevinMind: If the only use case is to call it within the apply_async method, can we not directly call super().apply_async() ? Is it possible to make this method private?

src/olympia/devhub/tasks.py
```diff
@@ -73,6 +73,7 @@ def validate(file_, *, final_task=None, theme_specific=False):
     if task_id:
         return AsyncResult(task_id)
     else:
+        breakpoint()
```
KevinMind: why?

diox: I since removed that breakpoint (was only there while I was debugging)

src/olympia/amo/celery.py
diox: This method used to be provided by django-post-request-task in case you wanted to bypass, well, django-post-request-task. It's used once elsewhere in our code to trigger a task immediately without waiting for the transaction on purpose. I'll document it.

src/olympia/amo/celery.py
```diff
@@ -50,13 +53,27 @@ def _serialize_args_and_kwargs_for_eager_mode(
...
         if app.conf.task_always_eager:
             args, kwargs = self._serialize_args_and_kwargs_for_eager_mode(
                 args=args, kwargs=kwargs, **options
             )
-
-        return super().apply_async(args=args, kwargs=kwargs, **options)
+            # In eager mode, immediately call original apply async as we are
+            # using eager mode for tests, where no transaction is ever actually
+            # committed so transaction.on_commit() is never called.
+            return self.original_apply_async(args=args, kwargs=kwargs, **options)
```
KevinMind, line 67: Not really sure what the impact of this is, but we sometimes return the value of the caller, and sometimes not? Is it not possible to use a lambda or some wrapper function to assign the value of self.original_apply_async  and return that? Or do we have to return in the other case? Why not have consistent return behavior?

KevinMind:
It's too bad the PR is from a fork. I was going to use the pushed image tag to test.. Schade. Need to solve that.

src/olympia/amo/celery.py
diox: I agree it's inconsistent, but we can't return anything in the transaction path: `transaction.on_commit()` registers a callback to be executed at a later time, and returns immediately.

The same issue was already present in django-post-request-task: https://github.com/mozilla/django-post-request-task/blob/4f1e33ec778bbba6210f00c7eb6d5820e0bc8d82/post_request_task/task.py#L98-L104 (`_append_task()` returns `None`)

diox:
Well the good news is that the PR easy to drop in your local env, don't need any migrations and you can even ignore the requirement change to test it locally since I'm just removing a package we no longer need.

KevinMind:
Testing:

> and it shouldn't make a difference.

I get additional logs when running with `--atomic` but otherwise I get 

```bash
olympia@265cef13c049:~$ python3 manage.py generate_error --celery --atomic
Inside an atomic block...
A RuntimeError exception was raised from a celery task. Check the logs!
```

I'm looking at `./logs/supervisord-celery.log` and don't see anything. What should I be looking for?

> Submitting a new add-on.

Can you share the resource for getting a dummy addon zip? We should have a way to automate that in development It's such a common use case for testing.

> Commenting in review page for an add-on.

Can you provide steps to do this?

diox:
> I'm looking at `./logs/supervisord-celery.log` and don't see anything. What should I be looking for?

Wrong log, it should be `docker-celery.log`, which contains the worker logs. The supervisor logs are irrelevant here.

You should see that we correctly triggered a celery task (that one happens to generate an error so it's highly visible in the logs)

>> Submitting a new add-on.
> Can you share the resource for getting a dummy addon zip? We should have a way to automate that in development It's such a common use case for testing.

https://staticfil.es/webext-generator/ or steal any add-on from production. I frequently use dummy ones from the generator but also larger, more complex ones from prod like ublock origin.

>> Commenting in review page for an add-on.
> Can you provide steps to do this?

(I edited my text to mention reply rather than comment, which is a silent action)

If you're loggued in as an admin with `ADDONS_EDIT` (should be the case by default in local envs if you're using the superadmin account created at initialization), the listed/unlisted review page should be linked from the left-side pane in the devhub page for your add-on for your convenience. From there you can select "Reviewer reply" action.

src/olympia/amo/celery.py
KevinMind: Yes but then if we have to return none, why not also return none in the always_eager mode.

```python
self.original_apply_async(args=args, kwargs=kwargs, **options)
return None
```

🤷 just seems like we are creating different behaviour in tests, which is already different. I'm not strongly advocating for this, just don't see why not if it's 1LOC

KevinMind (approved):
Verified all cases locally. Works! Left a few Nits but nothing stopping merging

src/olympia/amo/celery.py
diox: Ah yes, we could probably do that. It was a bit more controversial to do in django-post-request-task, which was somewhat generic (so it could have consumers of the library that wanted the return value in some cases) but we should be OK to do it now.

src/olympia/amo/celery.py
diox: 0d6dd7911864e6e2590592b5c24d770c0de195d0

src/olympia/amo/celery.py
diox: ddd44bade557c76c4cfc9e7887e67d7d58d1f859

PR #22381, opened 2024-06-19
Escape json blob in context job

KevinMind:
Fixes: mozilla/addons/issues/14857

### Description

cat <<EOF is to "" 

as 

cat <<'EOF' is  to ''

We want to prevent any interpolation of the values inside the github context blob as these can cause arbitrary execution errors in what should just be a plain log statement.


diox:
We also need that in `release.yml` right ?

diox (approved):

PR #22313, opened 2024-06-03
Make up the docs

KevinMind:
Fixes: https://github.com/mozilla/addons/issues/14789

### Description

This PR updates our docs site by replacing the already outdated before this PR installation docs with a new set of docs that aims to describe how to work with the repository in local and CI environments. This PR documents most of the work shipped in this epic https://mozilla-hub.atlassian.net/browse/AMOENG-561 

### Testing

To open the version of our docs on this PR you can download the docs artifact and open locally.

1. Click on the build_docs job in the status checks

image

2. Click into the upload artifact step

image

3. Find the artifact download url

image

Click on the URL. This will download a zipped folder of the build.

4. Unzip the directory locally and open the index.html file in your browser.

5. Enjoy the docs.

PR #22383, opened 2024-06-19
Remove deprecated documents, including references.

KevinMind:
Relates to: mozilla/addons/issues/14789

### Description

This PR removes deprecated and outdated documentation that will soon be replaced with more up to date docs.

### Context

This PR is meant to make this [PR](https://github.com/mozilla/addons-server/pull/22313) more manageable by removing documents before adding them. This way we can add docs in smaller chunks to make review more manageable.

One downside is we will temporarily be missing docs entirely but that is probably okay because everything being removed here is essentially invalid information anyway.

The process was:

- remove all deprecated docs
- run `make docs` follow each error for no longer valid references and 
a) remove the reference
b) update the reference


KevinMind:
Downloadable docs https://github.com/mozilla/addons-server/actions/runs/9584869650/artifacts/1617691342

diox (approved):

PR #22384, opened 2024-06-19
Remove egg-info without breaking locale extraction

KevinMind:
Relates to: mozilla/addons#14863

### Description

This PR removes the logic to install the `olympia` module. In order to resolve olympia as a python module we set the path to `$HOME` and `$HOME/src` in `PYTHONPATH` during the docker build.

In order to resolve the custom extractor used by babel, we have to explicitly map the `jinja2_custom`.

### Context

The way we were installing `olympia` was already "legacy" and since we do not distribute addons-server as a package there is no real reason to install.

Removing this logic doesn't degrade any functionality and significantly simplifies the architecture. Installing would have eventually needed to be upgraded and also results in files that need to be managed.

### Testing

Run this PR locally

```bash
make up DOCKER_TAG="mozilla/addons-server:pr-22384-ci"
```

Verify you can run `make update_assets` and `make extract_locales` successfully. Also verify you can run the app and everything functions normally.

You can verify that olympia is NOT installed anymore by running

```bash
pip show olympia
```


pyproject.toml
```diff
@@ -57,13 +85,9 @@ norecursedirs = [
   "__pycache__",
 ]
 DJANGO_SETTINGS_MODULE = "settings_test"
-# Ignoring csp deprecation warnings, we have control over the module and
-# currently it warns for child-src which is deprecated in CSPv3 but we're still
-# on CSP 2 while CSP 3 is still in working draft (sept 2018)
```
diox, line 62: This comment is still useful (but should be moved below)

diox:
FYI I still had an old `src/olympia.egg-info` laying around that I had to delete to make `pip show olympia` return nothing, but it didn't have any other impact.

diox (approved):

pyproject.toml
KevinMind: REverted

KevinMind:
> FYI I still had an old `src/olympia.egg-info` laying around that I had to delete to make `pip show olympia` return nothing, but it didn't have any other impact.

We still have logic in makefile-os to remove that directory. I think we should keep it around for a while to prevent that edge case.

PR #22401, opened 2024-06-21
Clear NeedsHumanReview selectively depending on job being resolved

diox:
Fixes: mozilla/addons#14872

### Description

If there are multiple jobs for a given add-on, when resolving a job we only want to clear the `NeedsHumanReview` flag if the job we're resolving is the last one of its kind: the add-on should stay in the queue until all jobs are resolved.

With that logic in place, because reports can be added to an ongoing appeal job, we need to also prevent NHR from being added following a report attached to an appeal job (otherwise we can't determine what NHR to clear). It's redundant anyway, there will already have a NHR for the appeal and that's enough.

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -517,12 +517,11 @@ class CinderAddonHandledByReviewers(CinderAddon):
             )
 
     def post_report(self, job):
-        if not (job.is_appeal and job.is_unresolved):
+        if not job.is_appeal:
             self.flag_for_human_review(appeal=False)
-        # If our report was added to an unresolved appeal job (i.e. an appeal
-        # was ongoing, and a report was made against the add-on), don't flag
-        # the add-on for human review again : we should already have one
-        # because of the appeal.
+        # If our report was added to an appeal job (i.e. an appeal was ongoing,
+        # and a report was made against the add-on), don't flag the add-on for
+        # human review again: we should already have one because of the appeal.
 
     def appeal(self, *args, **kwargs):
         self.flag_for_human_review(appeal=True)
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -517,11 +517,11 @@ class CinderAddonHandledByReviewers(CinderAddon):
             )
 
     def post_report(self, job):
-        if not job.is_appeal:
+        if not (job.is_appeal and job.is_unresolved):
             self.flag_for_human_review(appeal=False)
-        # If our report was added to an appeal job (i.e. an appeal was ongoing,
-        # not yet resolved, and a report was made against the add-on), don't
-        # flag the add-on for human review again : we should already have one
+        # If our report was added to an unresolved appeal job (i.e. an appeal
+        # was ongoing, and a report was made against the add-on), don't flag
+        # the add-on for human review again : we should already have one
         # because of the appeal.
 
     def appeal(self, *args, **kwargs):
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -40,9 +40,12 @@ class CinderEntity:
         return str(field_content or '').strip()
 
     def get_attributes(self):
+        """Return dict of attributes for this entity, to be sent to Cinder."""
         raise NotImplementedError
 
     def get_context_generator(self):
+        """Return a context generator containing dict of relationships for this
+        entity, to be sent to Cinder separately from the main attributes."""
         raise NotImplementedError
 
     def get_empty_context(self):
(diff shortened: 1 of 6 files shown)


diox:
The flaw with that approach is that it undoes https://github.com/mozilla/addons/issues/1741 : reports _can_ be attached to an appeal job, even if they are not part of the appeal at all. Maybe the fix is to prevent NHR from being recorded when there is an ongoing appeal already... Or link the NHR to the CinderJob. Both options require some refactoring...

src/olympia/abuse/cinder.py
```diff
@@ -496,9 +516,13 @@ def flag_for_human_review(self, appeal=False):
                 user=core.get_user() or get_task_user(),
             )
 
-    def report(self, *args, **kwargs):
-        self.flag_for_human_review(appeal=False)
-        return super().report(*args, **kwargs)
+    def post_report(self, job):
+        if not (job.is_appeal and job.is_unresolved):
```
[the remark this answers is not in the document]
eviljeff: yeah cinder only adds new reports to open jobs*

*(technically open and unassigned, but assignment isn't applicable here, because you can't have an open escalated job)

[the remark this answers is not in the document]
diox: https://github.com/mozilla/addons-server/pull/22401/commits/6bd6d3d3d8648a9327ab612cb94d5308c5c2e447 removes that code

src/olympia/abuse/models.py
```diff
@@ -291,22 +293,41 @@ def resolve_job(self, *, log_entry):
...
+                        appealed_decisions__isnull=False
+                    ).exists()
+                )
+                reason = NeedsHumanReview.REASONS.ADDON_REVIEW_APPEAL
+            else:
+                # If the job we're resolving was not an appeal or escalation
+                # then all abuse reports are considered dealt with.
+                has_unresolved_jobs_with_similar_reason = None
+                reason = NeedsHumanReview.REASONS.ABUSE_ADDON_VIOLATION
+            if not has_unresolved_jobs_with_similar_reason:
```
eviljeff, line 324: might be missing an edge case, but we could just not set `reason` to a non-`None` value unless the queryset exists, then we don't need this `has_unresolved_jobs_with_similar_reason` variable.

eviljeff (approved):
just a coding nit

src/olympia/abuse/models.py
diox: Like this?

```diff
diff --git a/src/olympia/abuse/models.py b/src/olympia/abuse/models.py
index 0118d97338..f4da6a7cf7 100644
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -302,26 +302,22 @@ class CinderJob(ModelBase):
                 .unresolved()
                 .resolvable_in_reviewer_tools()
             )
+            reason = None
             if was_escalated:
-                has_unresolved_jobs_with_similar_reason = (
-                    base_unresolved_jobs_qs.filter(
-                        decision__action=DECISION_ACTIONS.AMO_ESCALATE_ADDON
-                    ).exists()
-                )
-                reason = NeedsHumanReview.REASONS.CINDER_ESCALATION
+                if base_unresolved_jobs_qs.filter(
+                    decision__action=DECISION_ACTIONS.AMO_ESCALATE_ADDON
+                ).exists():
+                    reason = NeedsHumanReview.REASONS.CINDER_ESCALATION
             elif self.is_appeal:
-                has_unresolved_jobs_with_similar_reason = (
-                    base_unresolved_jobs_qs.filter(
-                        appealed_decisions__isnull=False
-                    ).exists()
-                )
-                reason = NeedsHumanReview.REASONS.ADDON_REVIEW_APPEAL
+                if base_unresolved_jobs_qs.filter(
+                    appealed_decisions__isnull=False
+                ).exists():
+                    reason = NeedsHumanReview.REASONS.ADDON_REVIEW_APPEAL
             else:
                 # If the job we're resolving was not an appeal or escalation
                 # then all abuse reports are considered dealt with.
-                has_unresolved_jobs_with_similar_reason = None
                 reason = NeedsHumanReview.REASONS.ABUSE_ADDON_VIOLATION
-            if not has_unresolved_jobs_with_similar_reason:
+            if reason:
                 NeedsHumanReview.objects.filter(
                     version__addon_id=cinder_decision.addon_id,
                     is_active=True,
```

I'm not sure this is better, it feels harder to read for me (though I agree the `has_unresolved_jobs_with_similar_reason = None` in my original version is not ideal)

eviljeff: yeah, like that.  But I do see it's not a whole lot less code with all the line wrapping and wordy queries and constants.  It'd probably read better if we had queryset methods for `.escalated` and `is_appeal` but that's even more extra code... so 🤷 your choice on whatever you think is best.

diox: Ok, going to keep my original, slightly more verbose version for now. I suspect this is not the last time we'll need to refactor this code anyway...

PR #22405, opened 2024-06-21
Login to GAR

KevinMind:
Relates to: <https://github.com/mozilla/addons/14823>

### Description

Add login and config for pushing production image from github actions

### Context

This PR proves all of the configurations for pushing the image correctly are working before actually attempting to push the image. We can add the logic for pushing to GAR bit by bit, verifying each step of the way in a series of PRs.

### Testing

CI is enough

Fork: https://github.com/mozilla/addons-server/actions/runs/9647365195/job/26605952760?pr=22411

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,10 @@ jobs:
       digest: ${{ steps.build.outputs.digest }}
       version: ${{ steps.build.outputs.version }}
 
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -133,6 +137,25 @@ jobs:
...
+        with:
+          token_format: access_token
+          service_account: ${{ secrets.GAR_PUSHER_SERVICE_ACCOUNT_EMAIL }}
+          workload_identity_provider: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}
+
+      - name: login to GAR
+        if: ${{ steps.should_build.outputs.result == 'true' }}
+        uses: docker/login-action@v3
+        with:
+          registry: us-docker.pkg.dev
```
diox, line 155: Where is that documented that we should be using that registry ? (I'm unfamiliar with GAR, is that the default or one we chose to use explicitly ?)

diox (approved):

.github/workflows/ci.yml
KevinMind: We chose explicitly. It's a question for @bqbn and I can add it in a follow up PR.

bqbn: `pkg.dev` to GCP is like `hub.docker.com` to Docker Inc., which means if you use GAR, you pretty much just use that name.

That said, there is some difference. GCP GAR has multiple domain names for multiple locations. The `us` in the name represents the US multi-region. You can read https://cloud.google.com/artifact-registry/docs/repositories/repo-locations for more details.

PR #22408, opened 2024-06-24
Bump pkcs7_digest algorithm to sha256

KevinMind:
Fixes: mozilla/addons/issues/14804

### Testing

1. Download a sample addon ([here](https://staticfil.es/webext-generator/))
2. Upload via devhub (unlisted is better due to no 24 block on approval)
3. Auto approve the addon via `.manage.py`
4. View the product listing page
5. Find the .zip file in storage
6. Upload the zip file to [xpidump](https://williamdurand.fr/xpidump/)

Expect:
- addon is signed with sha256

### Checklist

- [X] Add `#ISSUENUM` at the top of your PR to an existing open issue in the mozilla/addons repository.
- [x] Successfully verified the change locally.
- [x] The change is covered by automated tests, or otherwise indicated why doing so is unnecessary/impossible.
- [X] Add or update relevant [docs](https://github.com/mozilla/addons-server/blob/master/docs) reflecting the changes made.

PR #22409, opened 2024-06-24
Fix storage volume config for the nginx service

willdurand:
Fixes https://github.com/mozilla/addons/issues/14882

---

I tested locally and it worked for me but not sure if I didn't break anything else...

## Testing

This is _a_ scenario:

1. Upload an add-on locally
2. Get it signed (likely needs to be approved manually in reviewer tools because it will be a _new_ add-on + signed via `./manage.py auto_approve`)
3. Navigate to the detail page of this add-on (frontend)
4. Try to download the XPI

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -185,7 +185,6 @@ def test_pre_setup(request, tmpdir, settings):
     settings.SHARED_STORAGE = shared_storage = _path(storage_root, 'shared_storage')
 
     settings.ADDONS_PATH = _path(storage_root, 'files')
-    settings.GUARDED_ADDONS_PATH = _path(storage_root, 'guarded-addons')
     settings.GIT_FILE_STORAGE_PATH = _path(storage_root, 'git-storage')
     settings.MLBF_STORAGE_PATH = _path(storage_root, 'mlbf')
     settings.MEDIA_ROOT = _path(shared_storage, 'uploads')
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -74,10 +74,7 @@ services:
       - ./docker/nginx/addons.conf:/etc/nginx/conf.d/addons.conf
       - ./static:/srv/static
       - ./site-static:/srv/site-static
-      - storage:/shared_storage/uploads:/srv/user-media
-      - storage:/files:/srv/user-media/addons
-      - storage:/guarded-addons:/srv/user-media/guarded-addons
-      - storage:/sitemaps:/srv/user-media/sitemaps
+      - storage:/srv/user-media
     ports:
       - "80:80"
     networks:
--- a/docker/nginx/addons.conf
+++ b/docker/nginx/addons.conf
@@ -5,25 +5,9 @@ merge_slashes off;
 server {
     listen  80 default;
 
-    location /data/olympia/storage/files/ {
+    location /data/olympia/storage/ {
         internal;
-        # This matches where addons-server `docker-compose.yml` mounts
-        # `./storage/addons/` - as `/srv/user-media/addons/`
-        alias /srv/user-media/addons/;
-    }
-
-    location /data/olympia/storage/guarded-addons/ {
-        internal;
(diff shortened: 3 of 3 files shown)


conftest.py
```diff
@@ -185,7 +185,6 @@ def _path(*args):
     settings.SHARED_STORAGE = shared_storage = _path(storage_root, 'shared_storage')
 
     settings.ADDONS_PATH = _path(storage_root, 'files')
-    settings.GUARDED_ADDONS_PATH = _path(storage_root, 'guarded-addons')
```
willdurand, line 188: We don't have guarded add-ons anymore.

docker-compose.yml
```diff
@@ -74,10 +74,7 @@ services:
       - ./docker/nginx/addons.conf:/etc/nginx/conf.d/addons.conf
       - ./static:/srv/static
       - ./site-static:/srv/site-static
-      - storage:/shared_storage/uploads:/srv/user-media
-      - storage:/files:/srv/user-media/addons
-      - storage:/guarded-addons:/srv/user-media/guarded-addons
-      - storage:/sitemaps:/srv/user-media/sitemaps
+      - storage:/srv/user-media
```
KevinMind, line 77: I wonder if it is possible to test this. `./.github/actions/run-docker` has nginx running but not the `/etc/host` so would have to use localhost I would guess.

willdurand: Can we predict a file URL that goes through all of this?  In the meantime you'll need this to be able to complete #22408 so I suppose manual confirmation would be a good first step.

KevinMind: I think we could populate a file in an expected path in ./storage and then verify that a call to nginx returns the file. We don't have to add it here, but can you file a follow up ticket to do so. This was broken for a while and no one noticed.

KevinMind:
I'm getting when clicking add to firefox. Can you add more explicit test instructions. How exactly to verify this PR?

image

KevinMind:
> > I'm getting when clicking add to firefox. Can you add more explicit test instructions. How exactly to verify this PR?
> > image
> 
> I think you need to submit an actual add-on (and get it auto-approved) because the generated data probably lacks actual files.

I was following the steps outlined in the linked [issue](https://github.com/mozilla/addons/issues/14882). Can you please update those steps to be more specific of what exactly I need to do in order to verify this change, or add those steps to the `testing` section of the PR?

> Set up addons-server locally
> Go to the homepage: http://olympia.test/en-US/firefox/
> Try to install/download any add-on

Based on these steps I would not have thought I needed to upload an addon.

docker-compose.yml
willdurand: Yep, I filed https://github.com/mozilla/addons/issues/14886

KevinMind:
Thanks, I will try to verify this today.

KevinMind:
Tried verifying with an addon I uploaded, it looks like we no longer get a 404, but instead an empty document? Here is a video of what I managed to achieve.

https://github.com/mozilla/addons-server/assets/19595165/41d57b0d-e04d-44fe-a8c7-d44a632eaa83

Installing gives an error of a corrupted file.. could just be the file is crap I think it was one of the empty dummy addons but not sure if this concludes verification or not.

image

willdurand:
> Installing gives an error of a corrupted file.. could just be the file is crap I think it was one of the empty dummy addons but not sure if this concludes verification or not.

That looks good. Since we're fixing the server, getting the right file (or _a_ file) served is good enough. The corrupt file error is because the file is signed with a nonprod certificate, which isn't accepted by default (by Firefox). While we have prefs to allow add-ons from AMO -dev/stage to be installed, I am not sure we can install add-ons signed with the local autograph config.

KevinMind (approved):
Adding a follow up [ticket](https://github.com/mozilla/addons/issues/14895) for covering with tests but 👍 🚢

PR #22424, opened 2024-06-26
Push image via artifact upload/downlod

KevinMind:
Relates to: mozilla/addons#14823

### Description

- support forks pushing images via artifacts
- remove pushing from build action
- modify context to support more robust checks

### Context

TBD

### Testing

TBD

### Checklist


.github/workflows/release_tag.yml
```diff
@@ -0,0 +1,76 @@
...
+    # Add "id-token" with the intended permissions.
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - id: image
+        uses: ./.github/actions/download-docker
```
bqbn, line 51: I had a question in https://github.com/mozilla/addons-server/pull/22423, which got closed. But the question remains so I'll ask it here again. :)

> What I don't understand is why the image is pulled from Dockerhub and then pushed to GAR. Isn't this what the pull-verify-push script in the deploy-amo repo is already doing?

> I was expecting the workflow to be
> 1. build the image
> 2. push the image to both Dockerhub and GAR

Thanks.

KevinMind: In this case it is not being pulled from dockerhub, but from a github artifact. The reason it is being pulled from a github artifact is to solve a totally different problem, which is how to use pushed images in our CI with forked PRs, which have no access to secrets and so cannot authenticate.

KevinMind: To be clear, this is still WIP.

KevinMind:
Squashed: https://github.com/mozilla/addons-server/pull/22433

.github/actions/build-docker/action.yml
```diff
@@ -1,114 +1,183 @@
...
-      shell: bash
-      # The suffix is unique to a branch and commit
-      run: |
-        if [[ "${{ inputs.push }}" == "false" ]]; then
-          echo "suffix=${{ github.ref }}-${{ github.sha }}" >> $GITHUB_OUTPUT
+        if [[ "$push" == 'true' ]]; then
+          if [[ "$username" != '' && "$password" != '' ]]; then
+            should_push=true
+          else
+            echo "`push` set to true but missing username: $username and/or password: $password"
```
KevinMind, line 87: Cannot use backtick for strings, this is executed apparently

https://github.com/mozilla/addons-server/actions/runs/9758341983/job/26932695197?pr=22433

bqbn: Doesn't this have the potential to print out the actual password when the `username` is empty and the `password` is not? 

_maybe we don't want to echo `$password` at all._

PR #22434, opened 2024-07-02
Split test jobs to callable workflows

KevinMind:
Relates to: mozilla/addons#14823

### Description

Split some of our test jobs from ci.yml to callable workflows. Re-organize the _test.yml workflow to use a matrix strategy for simpler definition.

### Context

We have a lot of jobs in our CI. Previously each was defined independently in the ci.yml file. This PR splits the test jobs out into 2 separately callable workflows that are then called by the CI workflow.

Using `workflow_call` events allow workflows to be called from other workflows (see [docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows)). This gives 2 tangible benefits.

1. Callable workflows can keep CI jobs DRY by allowing you to define a set of jobs that should be executed by multiple workflows. This will be useful to us shortly when we introduce release workflows for master and tags that also want to run tests.

2. Jobs executed in a called workflow are collapsible in the jobs list view in the action run details page. This allows better organization in the view and cleaning up the UI by collapsing is just nice.

After / Before

image

Collapsed View:

image

### Testing

CI should pass and run essentially the same jobs as before.

### Checklist

### DIFF
--- a/.github/actions/context/action.yml
+++ b/.github/actions/context/action.yml
@@ -1,0 +1,114 @@
+name: 'Dump Context'
+description: 'Display context for action run'
+
+outputs:
+  # All github action outputs are strings, even if set to "true"
+  # so when using these values always assert against strings or convert from json
+  # \$\{{ needs.context.outputs.is_fork == 'true' }} // true
+  # \$\{{ fromJson(needs.context.outputs.is_fork) == false }} // true
+  # \$\{{ needs.context.outputs.is_fork == true }} // false
+  # \$\{{ needs.context.outputs.is_fork }} // false
+  is_fork:
+    description: ""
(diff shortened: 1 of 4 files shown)


KevinMind:
> not sure about the _naming - is that so they don't show in the list of workflows? - if they're effectively an action could they be in actions/ instead? r+wc

I would have liked to colocate the reusable workflows in a directory but unfortunately this is not supported, [docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow)

The _ is so we can distinguish between workflows that are called between ones that are triggered. do you have a better idea how to organize? I'm open to a follow up PR with better organization for sure.

PR #22438, opened 2024-07-02
Better handle unknown values for `addon_signature` in add-on abuse reports

willdurand:
Fixes https://github.com/mozilla/addons/issues/14896

---

## Testing

You can test this patch by (1) updating the `extensions.addonAbuseReport.url` pref to point to `http://olympia.test/api/v5/abuse/report/addon/` and (2) following QA's STRs in the bug.


src/olympia/abuse/models.py
```diff
@@ -370,6 +370,7 @@ class AbuseReport(ModelBase):
         ('SIGNED', 10, 'Signed'),
         ('SYSTEM', 11, 'System'),
         ('PRIVILEGED', 12, 'Privileged'),
+        ('NOT_REQUIRED', 13, 'Not required'),
```
willdurand, line 373: I am going to update some code in Firefox to report `NOT_REQUIRED` in the future so that's why this is added here.

src/olympia/abuse/serializers.py
```diff
@@ -257,6 +257,14 @@ def handle_unknown_install_method_or_source(self, data, field_name):
             value = 'other'
         return value
 
+    def handle_addon_signature(self, value):
+        # When the value is starting with `unknown: `, we force the value to
+        # UNKNOWN to account for any unknown value sent by Firefox, see:
+        # https://searchfox.org/mozilla-central/rev/b368ed8b48c0ea8ed2f1948e4776a6fbb5976dff/toolkit/mozapps/extensions/AbuseReporter.sys.mjs#218-219
+        if isinstance(value, str) and value.lower().startswith('unknown: '):
```
KevinMind, line 264: where is the literal string `unknown: ` coming from and how likely is it to change?

KevinMind: Why not save the value anyway? We can always add logic on our end to handle the class of strings tarting with unknown via a model method and this allows us to maintain the real truth of those values.

KevinMind: I see now in the code we have similar cases doing the same thing, `other` for example. 👍

KevinMind:
How can I get some dictionaries on my local environment?

image

src/olympia/abuse/serializers.py
```diff
@@ -257,6 +257,14 @@ def handle_unknown_install_method_or_source(self, data, field_name):
...
+            return 'unknown'
+        return value
+
     def to_internal_value(self, data):
         # We want to accept unknown incoming data for `addon_install_method`
```
KevinMind, line 269: Can you update the comment to include `addon_signature` just for consistency.

willdurand:
> How can I get some dictionaries on my local environment?

You'd need to submit new ones yourself, probably.

That being said, you don't need that for this bug. You can use AMO prod and install the add-on linked in the issue. The report will be sent via your local env when you set the pref mentioned in the PR description above.

KevinMind:
Verified submission successful on your PR. I followed these steps to verify.

1. set the pref `extensions.addonAbuseReport.url` to `http://olympia.test/api/v5/abuse/report/addon/`
2. Opened the addon report [page][report_url]
3. Submitted the following fields:

Screenshot 2024-07-03 at 09 01 45

4. Expect:

- on master: console error with error notification for unknown `addon_signature`

Screenshot 2024-07-03 at 09 13 02

- on branch: successfully submitted

Screenshot 2024-07-03 at 09 13 23

Notes:
- it is not clear to me if the addon needed to be installed for the flow to work as expected. I tried after installing but I did not verify with it uninstalled. That should be clarified in the ticket
- I did not see any network tab requests at any time related to submitting the report. I assume this is due to the request being processed via addon manager? TIL that this hides requests? That makes debugging a bit difficult. Could be something to reference for future in our `docs`.
- There is no criteria to verify anything on `addons-server` itself. I checked the only `AbuseReport` instance I have and it is missing the `addon_signature` field.

Screenshot 2024-07-03 at 09 26 16

[report_url]: https://addons.mozilla.org/en-US/firefox/feedback/addon/bambarabamanakan/

KevinMind (approved):
Approved with comments. Please update the `comment` and lmk about my follow ups but otherwise looks good.

For future, could you please use the PR template for the relevant bits of information? There are explainers for each relevant section and I think some of the information you included and some of the follow ups I've asked for could have gone in those sections.

It helps to have consistency on the PRs for reasons which should actually be documented somewhere (that's probably my job to do) but tldr;
- reduce cognitive load moving from PR to PR looking for something
- ensure critical information is captured
- encourage more documentation at the relevant place (in this case on the PR itself) for future reference.

willdurand:
> I did not see any network tab requests at any time related to submitting the report. I assume this is due to the request being processed via addon manager? TIL that this hides requests? That makes debugging a bit difficult. Could be something to reference for future in our `docs`.

Since Firefox makes the HTTP call, you need to look in the browser devtools (not the regular devtools).

willdurand:
> it is not clear to me if the addon needed to be installed for the flow to work as expected. I tried after installing but I did not verify with it uninstalled. That should be clarified in the ticket

The add-on needs to be installed yeah. When not installed, the code path in Firefox would be different.

willdurand:
> There is no criteria to verify anything on `addons-server` itself. I checked the only `AbuseReport` instance I have and it is missing the `addon_signature` field.

That is surprising. I have the "unknown" value when looking at the abuse report in the admin: http://olympia.test/en-US/admin/models/abuse/abusereport/

PR #22454, opened 2024-07-08
mark decision notes as safe so jinja doesn't escape them

eviljeff:
Fixes: mozilla/addons#14901

### Context

See issue; QA identified that `'` (and other characters) from the decision text are escaped with html entity codes in the emails.  It's rendered into a plain text email so it doesn't make sense to do that.  (And even if it wasn't, the content comes from ~inside the building~ our own reviewers and moderators, so should be regarded as safe anyway)

### Testing

Take any reviewer action in the reviewer tools and include a comment with it that has `'`.  (essentially every reviewer tools action with the comment box, apart from Comment, which doesn't send anything out).  Then check the email in fake mail in django admin.

### Checklist

### DIFF
--- a/src/olympia/abuse/tests/test_utils.py
+++ b/src/olympia/abuse/tests/test_utils.py
@@ -43,7 +43,7 @@ class BaseTestCinderAction:
         self.decision = CinderDecision.objects.create(
             cinder_id='ab89',
             action=DECISION_ACTIONS.AMO_APPROVE,
-            notes='extra notes',
+            notes="extra note's",
             addon=addon,
         )
         self.cinder_job = CinderJob.objects.create(
@@ -167,6 +167,7 @@ class BaseTestCinderAction:
         assert f'[ref:ab89/{self.abuse_report_auth.id}]' in mail.outbox[0].body
         assert '&quot;' not in mail.outbox[0].body
         assert '&lt;b&gt;' not in mail.outbox[0].body
+        assert '&#x27;' not in mail.outbox[0].body
         assert self.decision.notes in mail.outbox[0].body
 
     def _check_owner_email(self, mail_item, subject, snippet):
@@ -177,6 +178,7 @@ class BaseTestCinderAction:
         assert '[ref:ab89]' in mail_item.body
         assert '&quot;' not in mail_item.body
         assert '&lt;b&gt;' not in mail_item.body
+        assert '&#x27;' not in mail_item.body
         assert self.decision.notes in mail_item.body
 
     def _test_owner_takedown_email(self, subject, snippet):
@@ -198,6 +200,7 @@ class BaseTestCinderAction:
         )
         assert '&quot;' not in mail_item.body
         assert '&lt;b&gt;' not in mail_item.body
+        assert '&#x27;' not in mail_item.body
         assert self.decision.notes in mail_item.body
 
     def _test_owner_affirmation_email(self, subject):
@@ -206,6 +209,7 @@ class BaseTestCinderAction:
         assert 'right to appeal' not in mail_item.body
         notes = f'{self.decision.notes}. ' if self.decision.notes else ''
         assert f' was correct. {notes}Based on that determination' in (mail_item.body)
+        assert '&#x27;' not in mail_item.body
         if isinstance(self.decision.target, Addon):
             # Verify we used activity mail for Addon related target emails
             log_token = ActivityLogToken.objects.get()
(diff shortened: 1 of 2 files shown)


KevinMind:
> And even if it wasn't, the content comes from inside the building our own reviewers and moderators, so should be regarded as safe anyway

I'm not sure we can really say that. Even if the reviewers themselves are trusted, people copy paste all kinds of stuff into other stuff. I agree with the conclusion, if the text is rendered as a text email, then scripts are useless and would just be visible text, but trust alone probably isn't a good enough reason to remove sanitation.

src/olympia/abuse/tests/test_utils.py
```diff
@@ -43,7 +43,7 @@ def setUp(self):
         self.decision = CinderDecision.objects.create(
```
KevinMind, line 43: Since we are marking safe can you put an explicitly unsafe test case. notes=<script>jar=window.triggerExploit();"</script> e.g. and verify the script is rendered as plain text. Just kind of an exclamation point on the code that it is unsafely rendering.

src/olympia/abuse/utils.py
```diff
@@ -96,7 +96,9 @@ def notify_owners(self, *, log_entry_id=None, extra_context=None):
         reference_id = f'ref:{self.decision.get_reference_id()}'
         context_dict = {
             'is_third_party_initiated': self.decision.is_third_party_initiated,
-            'manual_reasoning_text': self.decision.notes or '',
+            # it's a plain-text email, and text from our own reveiewers, so we're safe
+            # - and we don't want ' or other charactors rendered with html escaping.
+            'manual_reasoning_text': mark_safe(self.decision.notes or ''),
```
KevinMind, line 101: Just curious, is this the right place to solve this problem? Why do we need to add mark_safe to email content if we are setting the type to plain text, could we not do this deeper in the stack where we are rendering the email?

```python
content = mark_safe(content) if email.type == 'plain' else content
```

Something like this?

KevinMind: In other words, could this bug exist elsewhere that we don't know about and how are we supposed to track this for the future? What stops it from just happening again?

src/olympia/abuse/tests/test_utils.py
KevinMind: It's not really about testing it per say, but drawing attention to it. Thanks.

src/olympia/abuse/utils.py
eviljeff: You can't `mark_safe` at the top level like that - at that point `content` has already been rendered down into a single string object.  

With django templates you've got these options:
A) mark each variable explicitly as safe (it becomes a string wrapped in a `SafeString` object) when the context is defined, like we're doing here with `mark_safe`.
B) in the template itself when it's used with `{{ manual_reasoning_text|safe }}` (though that doesn't work well inside a `{% blocktrans %}` tag, for some reason).
C) by using a `{% autoescape off %}` enclosing tag in the templates.  

**A** means a change in two places (notify_owners and notify_reporters). **B** means a change in 5 template files where `manual_reasoning` is referenced. **C** means a change in 5 files too, because some of the templates inherit from `base.html` or `base_reporter.html` and some don't.

> In other words, could this bug exist elsewhere that we don't know about 

It could.  Everything is supposed to be escaped by default normally, to avoid unescaped content ending up as html.  **C** would allow us to escape everything at a template level ... if we actually want that?  **A** means we can be deliberate about what we're marking safe or not.

> and how are we supposed to track this for the future? What stops it from just happening again?

Assuming you don't mean in general over our entire codebase (which would be a massive scope creep), then for these specific emails, the tests are checking that the `'` entity code doesn't exist in the email body, so that would apply to every variable inserted.  Beyond that, we're into proving a negative territory, imo.

src/olympia/abuse/tests/test_utils.py
eviljeff: https://github.com/mozilla/addons-server/pull/22454/commits/6811344e96b0fddb6bf2456411ccb93366f84ff1
 (and https://github.com/eviljeff/olympia/actions/runs/9854130982 to confirm the CI passes)

src/olympia/abuse/utils.py
KevinMind: Cool, thanks for the explanation!

KevinMind (approved):
🚢

PR #3460, opened 2016-09-09
Remove 'self' and https://addons.mozilla.org from script-src

muffinresearch:
Fixes #3459

PR #22478, opened 2024-07-16
Replace drf-yasg with drf_specatcular and make swagger enable-able in production via environment variable

KevinMind:
Relates to: mozilla/addons#15019

## Description

Replace `drf_yasg` with `drf_spectacular` as the more maintained and v3 supporting library. The new lib has better control over documenting serializers and supports the openapi spec v3 which natively has better type support.

https://drf-spectacular.readthedocs.io/en/latest/drf_yasg.html

### Dev

Run the project

```bash
```

Open the swagger urls:
- schema: <http://olympia.test/api/v5/schema?format=json>
- swagger <http://olympia.test/api/v5/swagger>
- redoc: <http://olympia.test/api/v5/redoc>

These URLs should work, load the appropriate content and there should be no failing asset requests.

Run the generate command

```bash
make shell
./manage.py generate_js_swagger_files
```

Expect files in `./static-build/js/swagger/{version}.js` to be created for each supported API version (v3,v4,v5)

### Prod

Run the project

```bash
```

Expect the same as dev but the files should already exist without running the generate command as it is run during the build.

> [!NOTE]
> Expect the files in both `static-build` and `site-static` directories in prod mode
> The files will be served from the nginx server and not the olympia server 

## Checklist


KevinMind:
Fails test https://github.com/mozilla/addons-server/pull/3460. We can load the package via CDN but this seems bad for local development as you'd need a persistent internet connection..

diox:
Should I wait until https://github.com/mozilla/addons-server/pull/22663 is merged before reviewing this ? Or could you remove that commit (https://github.com/mozilla/addons-server/pull/22478/commits/6b9dbdc577321129fd68610aaba5cb43131bee93) ? It makes the diff a lot bigger and harder to follow.

KevinMind:
@diox Reviving this since it has been sitting here and there is nothing stopping its merge. It is bottom of the list in priority but enables some of the work I'm doing on the data retention policy since I want to add better documentation of some of the models.

src/olympia/lib/settings_base.py
```diff
@@ -1108,6 +1110,7 @@ def get_db_config(environ_var, atomic_requests=True):
 CSP_OBJECT_SRC = ("'none'",)
 
 CSP_SCRIPT_SRC = (
+    "'self'",
```
diox, line 1113: Err, that's a deal breaker. For prod, we absolutely don't want to allow `self` as a `script-src` in our CSP, that's too dangerous.

diox (changes requested):

src/olympia/lib/settings_base.py
KevinMind: Another option, we can exclude `/api` or some subset of that from CSP... those URLs are not "really" relevant for CSP anyway right? Or are we worried about host spoofing code on specifically that subset.. can you explain in more detail the set of security vulnerabilities we are trying to defend against so it's easier to find a solution that meets the requirement?

diox: We're worried about any XSS on AMO domain, which would have typical wide consequences that XSS can have: stealing cookies, creating fake pages misleading users, etc.

This would be less likely on `/api/`... but still possible, in fact with swagger it's even more likely as it makes us host HTML content...

A strict CSP on all our responses is a non-negotiable requirement. If anything it should be stricter on `/api/` given that `/api/` shouldn't host anything that needs to load content (we gave up on it a while ago because it was low priority, but we did consider that at some point: https://github.com/mozilla/addons/issues/1855).

KevinMind: I've got another workaround which is handling this file the way we handle javascript catalog files. Currently the "dev" version generating on demand is working, but the management command to statically generate for production is failing.. in our production nginx, are requests to /static that are not resolved by our static assets bucket forwarded to django? I guess not right. That would really simplify this because then I could just handle it dynamically every time and call it a day.

diox: No fallback no. It's a bit messy, it goes through a CDN config and then nginx, ideally it would be CDN + Google Cloud Storage instead (as is already the case for addons-frontend static files), so I would rather not rely on adding a fallback mechanism.

KevinMind: I think I got it.. let's see. We have a per version url path based on /static/js/swagger/{version}.js. IN dev, this is rendered on the fly and during the build we pre generate each version in the static-build dir which should be collected and thus available in prod.

KevinMind: Works.

diox: You still have 'self' allowed though ?

KevinMind: Fixed. hopefully the verification still passes 🙃

diox:
Can you fix the conflicts too?

KevinMind:
@diox ready to go.

diox:
Multiple test failures https://github.com/mozilla/addons-server/actions/runs/13864297054/job/38799990651?pr=22478

src/olympia/amo/tests/test_csp_headers.py
```diff
@@ -105,7 +105,6 @@ def test_self_in_common_settings(self):
     def test_not_self_in_script_child_or_style_src(self):
         """script-src/style-src/child-src should not need 'self' or the entire
         a.m.o. domain"""
-        assert "'self'" not in base_settings.CSP_SCRIPT_SRC
```
diox, line 108: Keep this

src/olympia/api/urls.py
```diff
@@ -94,6 +102,13 @@ def get_versioned_api_routes(version, url_patterns):
     re_path(r'^shelves/', include('olympia.shelves.urls')),
 ]
 
+url_name = 'schema'
+
+try:
+    static_swagger_ui_js = staticfiles_storage.url('js/swagger/swagger_ui.js')
+except ValueError:
+    static_swagger_ui_js = None
+
```
diox, line 111: Should this be there? This is duplicated from what's in `get_versioned_api_routes()` except without the versioning

src/olympia/schema.py
```diff
@@ -0,0 +1,19 @@
...
+class MyAuthenticationScheme(OpenApiAuthenticationExtension):
+    target_class = 'olympia.api.authentication.SessionIDAuthentication'
+    name = 'SessionIDAuthentication'
+
+    def get_security_definition(self, auto_schema):
+        return {
+            'type': 'apiKey',
+            'in': 'header',
+            'name': 'Authorization',
+            'description': 'pase the sessionId cookie as "Session {token}"',
```
diox, line 18:
```suggestion
            'description': 'pass the sessionId cookie as "Session {token}"',
```

diox (approved):

PR #22479, opened 2024-07-17
Add pip-api to make hashin work out of the box

KevinMind:
Fixes: mozilla/addons#14914

### Checklist

### DIFF
--- a/docs/topics/development/dependencies.md
+++ b/docs/topics/development/dependencies.md
@@ -52,6 +52,18 @@ This prevents pip from installing transitive dependencies.
 
 We do this because it gives us control over the full dependency chain - we know exactly which version of what package is installed so we can fully reproduce & trust environments.
 
+In order to determine the dependencies a given package requires you can check
+
+```bash
+pip show <package-name>
+```
+
+To see the `requirements` field which lists the dependencies. Install missing dependencies manually.
+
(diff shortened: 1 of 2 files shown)


docs/topics/development/dependencies.md
```diff
@@ -52,6 +52,18 @@ This prevents pip from installing transitive dependencies.
...
+In order to determine the dependencies a given package requires you can check
+
+```bash
+pip show <package-name>
+```
+
+To see the `requirements` field which lists the dependencies. Install missing dependencies manually.
+
+```{admonition} Note
+Ensure to document in the requirements file which "main" dependency a "transitive" dependency is required by.
```
eviljeff, line 64:
```suggestion
Ensure to comment in the requirements file above transitive dependencies which direct dependency it is required by.
```
?

eviljeff: Although I'm not sure how reliable manually adding the information is.

eviljeff (approved):

docs/topics/development/dependencies.md
KevinMind: We have been doing it until now (AFAICS) and probably better to have it than not.. ideally we'd have a pip.lock.py type file to do it for us, but alas.

PR #22483, opened 2024-07-17
Add auto generated model documentation to AMO docs

KevinMind:
Fixes: <https://mozilla-hub.atlassian.net/browse/AMOENG-1285>

## Description 

1. Introduces django model level documentation pages auto generated from:
- model doc string
- model field class definition

2. Introduces application level dependency graphs (using graphviz)

## Context

By introducing the documentation to the docs website it makes our data models inspectable outside of our codebase. By automating the generation of the docs it ensures they are always correct and require little maintenance.

## Testing

1. Generate docs `make docs`
2. Expect embedded models in the pages.

### Testing ui behavior

navigate to an application page `topics/models/scanners/index.html` for example.

1. Initially you should simply see the embedded diagram.
2. if you click on the diagram or add `?zoom` to the url (plus refresh) then you will see the diagram in full screen and be able to pan/zoom with drag or wheel events.
3. You can click "go back" or remove `?zoom` from the url to reset to the initial state.

https://github.com/user-attachments/assets/b5f611e2-780b-45f4-a6dc-19abae447c0d

## Checklist


KevinMind:
> Where is `docs/_static/panzoom.min.js` coming from ? Should it be installed through `npm`'s `devDependencies` ?

I'm not totally sure how to do that with sphinx and it seemed easier to have an offline version rather than using a CDN.

diox:
Definitely don't want to depend on a CDN version, but we should make sure we can update it, rebuild it etc. Also obey licensing.

We're already installing npm dependencies in the image, so why can't we use that ? Even just copying the built file from `node_modules/panzoom/dist` when building docs would be enough.

diox (changes requested):
(re-request when done fixing)

KevinMind:
WDYT, I actually think the graph generated by GraphViz is better than the one by django-model-info

### GraphViz

Screenshot 2025-05-16 at 19 45 44

### Django Model Info

Screenshot 2025-05-16 at 19 38 15

KevinMind:
@diox done

diox:
> WDYT, I actually think the graph generated by GraphViz is better than the one by django-model-info

Agreed

diox (approved):

PR #22500, opened 2024-07-22
Avoid extra queries around banned user content in user admin

diox:
Before this change these queries would happen all the time: we display counts of disabled content on ban for each type, that's 4 queries, and for every user, even not banned, Django would execute them once to display the hidden instance used to create additional instances when the user adds one, and another time for the existing default instance displayed because of the `extra` bit.

Because those queries do JOINs _and_ there isn't even an actual banned_user_content row to find, there were extra slow on top of being useless.

Preventing adds (it's an internal model anyway that admins should look at, but not manually add) and removing the extra default instance fixes the problem.

Fixes: https://github.com/mozilla/addons/issues/2011

### Testing

This wasn't really noticeable locally (the queries would be fast enough on most environments that don't have a lot of data), but the unit test included demonstrate the extra queries.

### DIFF
--- a/src/olympia/users/admin.py
+++ b/src/olympia/users/admin.py
@@ -49,6 +49,7 @@ from .models import (
 class GroupUserInline(admin.TabularInline):
     model = GroupUser
     raw_id_fields = ('user',)
+    extra = 0
 
 
 class BannedUserContentInline(admin.TabularInline):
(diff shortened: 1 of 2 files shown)


src/olympia/users/admin.py
```diff
@@ -70,10 +71,21 @@ class BannedUserContentInline(admin.TabularInline):
...
+        # that would be caused by banned_content_link().
+        return False
 
     def banned_content_link(self, obj, related_class):
         return related_content_link(
-            obj, related_class, 'bannedusercontent', related_manager='unfiltered'
+            obj,
+            related_class,
+            'bannedusercontent',
+            related_manager='unfiltered',
```
eviljeff, line 88: nit, unnecessary change

src/olympia/users/tests/test_admin.py
```diff
@@ -477,6 +481,24 @@ def test_can_edit_with_users_edit_permission(self):
...
+        user = user_factory(email='someone@mozilla.com')
+        # We want to see absolutely everything, so make our user a superadmin.
+        self.grant_permission(user, '*:*')
+        self.client.force_login(user)
+        with self.assertNumQueries(21):
+            # - 4 savepoint/release
+            # - 2 current logged in user & groups
+            # - 2 target user & groups
+            # - 1 banned user content instance
+            # - 8 related content counts
```
eviljeff, line 494: What are these 8 queries - can you include which content we're currently expecting so if it changes it'll be clearer what's new?

eviljeff (approved):

src/olympia/users/tests/test_admin.py
diox: https://github.com/mozilla/addons-server/pull/22500/commits/a96944c88552cd49d7d375716b8a547bf18c2429

PR #22526, opened 2024-07-30
Bypass checks about remaining listed/owner authors in pending authors…

diox:
… API

This fixes an intermittent test error, because the methods we're overriding check AddonUser against the instance id, but the instance id in that case is not an AddonUser, it's an AddonUserPendingConfirmation.

Fixes: mozilla/addons#14933

Example of the test failing in CI without that change https://github.com/mozilla/addons-server/actions/runs/10162196826/job/28104173268?pr=22522

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -904,6 +904,18 @@ class AddonPendingAuthorSerializer(AddonAuthorSerializer):
         writeable_fields = ('addon', 'user_id', 'listed', 'role')
         read_only_fields = tuple(set(fields) - set(writeable_fields))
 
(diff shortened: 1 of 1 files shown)


eviljeff (approved):
Is there a way to reliably test this?  (given it's been passing for years and only now failing)

diox:
The failure I had is due to specific test ordering causing the ids for both tables to match - I can try to add a test triggering that, I'll look into it.

diox:
https://github.com/mozilla/addons-server/pull/22526/commits/1db2761cc95a6f9d7ffbc44ffe3c6c7b17f3b8a7 adds a regression test

PR #22530, opened 2024-07-31
Expose all the reasons why there is a due_date to the extensions queue (queryset)

eviljeff:
fixes: mozilla/addons#14946 to enable mozilla/addons#14921

### Description

Splits all the reasons that a version may get a due_date into `Q` objects, and adds a `VersionManager` method to return them.  The same Q objects are then used to calculate `should_have_due_date` and to annotate `Addon.get_queryset_for_pending_queues`, so they can be exposed as flags in the review queue.

Some of the (seemingly) unused reviewer tools icons were repurposed.

### Context

To show flags for why an add-on is in the review queue we need to show all the reasons, so to be consistent we should use the exact same queries. 

### Testing

- set up some add-ons with combinations of NeedsHumanReview reasons, and reasons they may need a pre-review.
- see all the relevant flags shown in the reviewer tools.
- bonus tests for weird combinations of nhrs on different versions that are inactive, or the versions disabled, etc.

### Checklist


src/olympia/versions/models.py
```diff
@@ -232,11 +232,80 @@ def should_have_due_date(self, negate=False):
...
+                    When(
+                        Exists(
+                            nhr_q.filter(
+                                needshumanreview__reason=NeedsHumanReview.REASONS.ABUSE_ADDON_VIOLATION
+                            )
+                        ),
+                        then=Version.DUE_DATE_REASON_NHR_ABUSE,
+                    ),
+                    default=0,
+                )
```
eviljeff, line 264: fwiw this can be rewritten as 
```suggestion
                + (
                    Exists(
                        nhr_q.filter(
                            needshumanreview__reason=NeedsHumanReview.REASONS.ABUSE_ADDON_VIOLATION
                        ),
                        output_field=models.IntegerField()
                    ) * Version.DUE_DATE_REASON_NHR_ABUSE
                )
```
but I'm not convinced it's more readable than "Case When Exists q then v default 0"

src/olympia/versions/models.py
```diff
@@ -1018,7 +1111,14 @@ def should_have_due_date(self):
...
         See VersionManager.should_have_due_date for logic."""
-        return Version.unfiltered.should_have_due_date().filter(id=self.id).exists()
+        obj = (
+            Version.unfiltered.should_have_due_date(annotate_reasons=True)
+            .filter(id=self.id)
+            .first()
+        )
+        if obj is not None:
+            self._due_date_reasons = obj._due_date_reasons
+        return obj is not None
```
eviljeff, line 1121: tbh I don't like this, because it's setting magical properties on the instance, in a way you may expect from a queryset with annotations, but would by accessing a `@property`.  But we _do_ need the value of `_due_date_reasons` if we're going to update `due_date` immediately after checking if there should be a due_date.  The alternative is we just delete this `@property` and directly call the manager method in the one place it's used (and update a bunch of tests 😒)

src/olympia/versions/migrations/0046_version_due_date_reasons.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('versions', '0045_versionprovenance'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='version',
+            name='due_date_reasons',
+            field=models.PositiveIntegerField(blank=True, null=True),
```
eviljeff, line 16: Could use PositiveSmallIntegerField actually, we're only using 6 bits and even a PositiveSmallIntegerField supports up to 15 [reliably](PositiveSmallIntegerField).

diox (approved):

PR #22531, opened 2024-08-01
Disable csrf checks on client_info testing view

diox:
Makes it easier to test POST requests against it to check difference in behavior with GET requests. That view only works on dev environment so it should be safe.

Helps mozilla/addons#14929


eviljeff (approved):

bqbn:
This is good. We should only allows this in the test environments!

PR #22533, opened 2024-08-02
Unify themes reviewers queue into a single queue

diox:
Fixes: https://github.com/mozilla/addons/issues/14935

### Testing

- Submit a new theme (using the theme creation wizard or not), find it in reviewer tools new "Themes" queue, approve it, submit a new version of the theme after that
- Submit another new theme
- You should see both themes in the reviewer tools themes queue

### DIFF
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -232,7 +232,7 @@ def dashboard(request):
     ):
         sections['Themes'] = [
             (
-                'Manual Review ({0})'.format(queue_counts['theme']),
+                'Awaiting Review ({0})'.format(queue_counts['theme']),
                 reverse('reviewers.queue_theme'),
             ),
             (
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -233,7 +233,7 @@ def dashboard(request):
         sections['Themes'] = [
             (
                 'Manual Review ({0})'.format(queue_counts['theme']),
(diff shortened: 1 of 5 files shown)


src/olympia/reviewers/views.py
```diff
@@ -233,12 +232,8 @@ def dashboard(request):
     ):
         sections['Themes'] = [
             (
-                'New ({0})'.format(queue_counts['theme_nominated']),
-                reverse('reviewers.queue_theme_nominated'),
-            ),
-            (
-                'Updates ({0})'.format(queue_counts['theme_pending']),
-                reverse('reviewers.queue_theme_pending'),
+                'Manual Review ({0})'.format(queue_counts['theme']),
```
eviljeff, line 235: Do we want to call it "Manual Review" when all themes are (right now) pre-review, so all manually reviewed? I don't have a great alternative title though... just "Review"?  "Awaiting Review"? 🤔

eviljeff (approved):
Wow.... I thought it would need some significant refactoring/reworking.  Yay past us for writing generic enough code I guess?

eviljeff:
I think we should confirm with Scott before merging this so he has a chance to notify the theme reviewers of the upcoming change.

src/olympia/reviewers/views.py
diox: `Awaiting Review` is probably better than `Manual Review` yeah.

diox:
We have decided to wait until after MozWeek.

PR #22535, opened 2024-08-05
Use nobody@mozilla.org as the `From` for all emails, even activity ones

diox:
Keep setting notifications@addons.mozilla.org as the `Reply-To` email (with logic to dispatch the replies to the correct activity according to prefix) but stop using that address as the `From`, as it will fail DMARC.

Fixes mozilla/addons#14944

### DIFF
--- a/src/olympia/abuse/utils.py
+++ b/src/olympia/abuse/utils.py
@@ -137,7 +137,12 @@ class CinderAction:
         if version := getattr(self, 'addon_version', None):
             unique_id = log_entry_id or random.randrange(100000)
             send_activity_mail(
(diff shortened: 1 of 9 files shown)


eviljeff (approved):
We could simplify the code by dropping `from_email` from `send_activity_mail`, if it's always the default.  

Or even go further down the chain, to `amo.utils.send_mail`?

diox:
I initially did that but it's a bit of a mess: `from_email` is RFC 2822, fully formatted address with name and email. This is consistent with django's `send_email`. While we probably always want to use the same email, sometimes we customize the name part, that's why I kept the parameter.

PR #22548, opened 2024-08-06
Show All Current Authors on the Review Page

chrstinalin:
Fixes: mozilla/addons#14923

### Description
List all current authors under an add-ons review page, marking listed ones in bold and non-listed ones in regular font weight.

**Before**
image

**After**
image

### Testing

1. From the landing page, navigate to 'Developer Hub' and submit a new add-on. Upload a [test extension](https://staticfil.es/webext-generator/).
2. On the addon 'edit' page,  navigate to `Listed Review Page`or `Unlisted Review Page` `(/reviewers/review/<id>`, `/reviewers/review-unlisted/<id>`)
3. You can directly add an author using the `AddonUser` in Django admin and uncheck the unlisted box to avoid the email confirmation.

### Checklist


KevinMind:
@chrstinalin could you add some steps to reproduce for verifying this change locally. It should include how to add listed/unlisted authors to an addon and then steps to open the review page for an addon to view the author list.

src/olympia/reviewers/templates/reviewers/review.html
```diff
@@ -24,9 +24,9 @@ <h2 class="addon"{{ addon.name|locale_html }} id="addon-name">
       <span class="version">{{ version.version }} ({{ version.get_channel_display() }})</span>
     {% endif %}
   </h2>
-  <h4 class="author">by
-    {% for author in addon.listed_authors %}
-    <a href="{{ url('reviewers.developer_profile', author.id) }}">{{ author.name }}</a>{% if not loop.last %},{% endif %}
+  <h4 class="author">by 
+    {% for author in addon.current_authors %}
+    <a href="{{ url('reviewers.developer_profile', author.id) }}" style="font-weight:{{ 'bold' if author in addon.listed_authors else 'normal' }}">{{ author.name }}</a>{% if not loop.last %},{% endif %}
```
diox, line 29: Could you avoid inline styles and create a class for this ? We currently allow inline styles in CSP because of existing old code but in general we try to avoid that.

chrstinalin:
@KevinMind I added some steps, but if it isn't clear enough let me know!

KevinMind:
IMO the distinction between listed and unlisted authors is pretty minimal and very easy to miss.

I think we should consider a different color or perhaps a lighter font weight for the unlisted.

maybe 300 for unlisted and bold for listed?

image

KevinMind (approved):
Code looks fine. I have a UX comment that we should look at with @wagnerand before merging.

Thanks for adding the test scenario. I found it helpful but I think you could add a bit more detail in the future to make sure it is clear exactly what steps should be followed to reproduce the issue. It's a delicate balance between too much info and not enough info but try to think from the perspective of someone who doesn't have any context on the issue or the codebase. how would you explain to them the steps? This not only ensures anyone can review and verify your PR but also that you yourself understand the exact flow and all of the potential edge cases there may or may not be.

Single example: I didn't know which "id" to use from the steps so I went to the review queue.. then realized the addon I uploaded isn't in a review queue. Spent some time thinking of how to add it to the queue and then realized there is a link to the review page from the addon edit page in devhub.

wagnerand:
> IMO the distinction between listed and unlisted authors is pretty minimal and very easy to miss.
> 
> I think we should consider a different color or perhaps a lighter font weight for the unlisted.
> 
> maybe 300 for unlisted and bold for listed?
> image

Thanks Kevin.

I agree the distinction is not super visible, I had assumed bold vs non-bold would be. It's hard to give a direct guidance without see what it looks like. We could try to make listed authors even more bold for instance. A different color has a drawback of not being clear what each color means. If we are further reducing the font-weight for non-listed authors, it might be too thin, but that too is hard to say without seeing it.

While looking at the image, a question came to my mind, acknowledging that I hadn't thought about that before: Are the listed authors always shown first? If not, could we add that? Also happy if we do it in a separate followup issue and patch.

chrstinalin:
@wagnerand It looks like the order is based on their Position attribute --

image
image
image

I could look into sorting it, but I'd be concerned about whether it conflicts with whatever the original intent of Position was.

chrstinalin:
@KevinMind @wagnerand And, just to toss ideas in for the UX, maybe a dashed underline and/or italics?

image
image

src/olympia/reviewers/templates/reviewers/review.html
```diff
@@ -24,9 +24,9 @@ <h2 class="addon"{{ addon.name|locale_html }} id="addon-name">
       <span class="version">{{ version.version }} ({{ version.get_channel_display() }})</span>
     {% endif %}
   </h2>
-  <h4 class="author">by
-    {% for author in addon.listed_authors %}
-    <a href="{{ url('reviewers.developer_profile', author.id) }}">{{ author.name }}</a>{% if not loop.last %},{% endif %}
+  <h4 class="author">by 
+    {% for author in addon.current_authors %}
+    <a href="{{ url('reviewers.developer_profile', author.id) }}"{% if author not in addon.listed_authors %}class="is_unlisted"{% endif %}>{{ author.name }}</a>{% if not loop.last %},{% endif %}
```
diox, line 29: The `Addon.attach_authors` adds a `role` and `listed` properties to the `UserProfile` instances it gathers, so you should be able to do this in a slightly more efficient way:

```suggestion
    <a href="{{ url('reviewers.developer_profile', author.id) }}"{% if not author.listed %}class="is_unlisted"{% endif %}>{{ author.name }}</a>{% if not loop.last %},{% endif %}
```

src/olympia/reviewers/tests/test_views.py
```diff
@@ -5637,12 +5638,33 @@ def test_links_to_developer_profile(self):
...
-        self.assertContains(response, profile_url)
+        author_profile_url = reverse('reviewers.developer_profile', args=(author.id,))
+        another_profile_url = reverse(
+            'reviewers.developer_profile', args=(another_author.id,)
+        )
+        unlisted_profile_url = reverse(
+            'reviewers.developer_profile', args=(unlisted_author.id,)
+        )
+
+        print(response.content)
```
diox, line 5654: Stray print btw

KevinMind:
@chrstinalin IMO bold + underline for listed authors and no bold no underline for unlisted is probably good enough. If it's not too much trouble could you post a screenshot with that variant as well as the same variant additionally reducing font weight for unlisted to 300. Curious to see if I agree with @wagnerand about the readability aspect.

Almost there!

chrstinalin:
@KevinMind The current version has no underline, I had forgotten to change the screenshot out (whoops) -- Here's what it is now:
image

And here's it with 300
image

KevinMind:
@wagnerand I vote option 2, wdyt? I think the lighter font makes it much clearer and is still readable IMO.

wagnerand:
Either works for me, thanks!

KevinMind (approved):
🚢

PR #22555, opened 2024-08-07
Add Submit Button to Review Due Date Changes

chrstinalin:
Fixes: mozilla/addons#14922

### Description

Changes the due date update functionality on the reviewer page to update `onblur` rather than on change.

**Before**
image

**After**
ScreenRecording2024-08-09at8 24 41AM-ezgif com-video-to-gif-converter
Flashes in the gif are tab refreshes. Changes don't persist across reload unless OK is pressed.

### Checklist


eviljeff:
> The button doesn't technically do anything (clicking it just unfocuses the date input and fires the `onblur` event). It's there as a visual aid to distinguish between 'saved' and 'unsaved' states.

I lead towards removing the button because it could be confusing that the date changed before the button was pressed, but it's @wagnerand 's call.  So the options are:
- trigger the XHR on the input blur, with a fake button (the PR now)
- trigger the XHR on the input blur (no button - the reviewer would have to click somewhere else to send the update)
- trigger the XHR on the button being pressed (no event on input blur)

src/olympia/reviewers/templates/reviewers/review.html
```diff
@@ -299,11 +299,14 @@ <h3 id="history">
     <ul class="admin_only">
       <li {% if not version.due_date %}class="hidden"{% endif %}>
         <label for="due_date_update">Change latest version review due date</label>
-        <input type="datetime-local" id="due_date_update"
+        <div style="display: flex; gap: 0.5em">
```
eviljeff, line 302: give a class to the div, and add this css to `reviewers.less`

wagnerand:
Let's do

> trigger the XHR on the button being pressed (no event on input blur)

eviljeff (approved):
r+

PR #22563, opened 2024-08-09
Prevent URLs from being added to add-on summaries 

diox:
Fixes https://github.com/mozilla/addons/issues/14953

Also simplify implementation of `NoURLsField` (already used for `Collection` descriptions and `UserProfile` biographies) to only look for URLs starting with http/https, to avoid being too aggressive while also preventing URLs with TLDs bleach's hardcoded regexp doesn't know about.


src/olympia/translations/models.py
```diff
@@ -251,11 +250,13 @@ def __str__(self):
...
-        # It's more effective/aggressive than something like r'http\S+', it can
-        # also detect things like foo.com.
-        self.localized_string_clean = URL_RE.sub('', self.localized_string).strip()
+        from olympia.amo.utils import URL_RE
+
+        self.localized_string_clean = (
+            URL_RE.sub('', self.localized_string).strip()
+            if self.localized_string
+            else self.localized_string
+        )
```
eviljeff, line 259:
```suggestion
        self.localized_string_clean = URL_RE.sub('', self.localized_string or '').strip()
```?

eviljeff (approved):

src/olympia/translations/models.py
diox: That would transform `None` into an empty string which I'm not sure we want to do at this point.

PR #22573, opened 2024-08-15
License related fixes

leoheitmannruiz:
https://github.com/mozilla/addons/issues/14976

### Checklist


diox:
1. I saw it too and I'm not sure either. They used to be before so I kept it as-is.
2. It's a base dataset for initializing a new copy of the project. We can update it to stay consistent but it doesn't have any impact in addons.mozilla.org production.

PR #22582, opened 2024-08-19
Re-implement cinder forwards (escalations) as a new job in the content infringement queue

eviljeff:
Fixes: mozilla/addons#14970

### Description

Reimplement forwards from cinder (nee escalations) without creating a CinderDecision, and creates a new CinderJob in the correct queue to track it. 

### Context

The renaming to forwarding to escalations was a low-priority Operations request that seemed easy to do at the same time.

More context in the issue and parent story ticket.  Essentially, forwarded jobs will be implemented by ~moving into~ _recreating the job in_ the special queue on cinder where AMO handled reviews are kept.

~The webhook implementation itself will be in a separate issue~ _There is no need for any new webhook implementation at this time.  We re-create the job in the content infringement queue ourselves.  There may be a future step where we re-implement this re-create step as a workflow in Cinder, and monitor a webhook instead._

~(The 3rd step will be to remove support for the AMO_ESCALATE_ADDON entirely, once T&S have updated whatever operational/training docs they have that references escalations)~

### Testing

The functionality should be the same as now, with any existing jobs that are escalated migrated.
Setup, before checking out the patch:
- Either escalate an existing job from cinder and apply the webhook JSON locally; or manually create the appropriate CinderDecision instance (`CinderDecision(action=AMO_ESCALATE_ADDON, notes='xxx', ...)`, update the job to be `handled_by_reviewers=True`) for a cinder handled job.

Then:
- ~run the migrations, see the CinderDecisions are deleted and the CinderJobs updated to `is_forwarded=True`, with the associated notes.~
  - ~bonus: reverse the migration, and re-apply it.~
- Escalate a different existing job from cinder and apply the webhook JSON locally
- See the job shows correctly as forwarded in the reviewer tools
- See the job has been recreated in cinder, in the content infringement queue

Footguns:
- make sure the waffle switch for escalations is enabled
- make sure the add-on you report has an is_signed=True current_version
### Checklist


diox:
Also a bit late but plz add story points to the jira ticket

src/olympia/abuse/cinder.py
```diff
@@ -470,63 +474,101 @@ class CinderAddonHandledByReviewers(CinderAddon):
...
+    def flag_for_human_review(
+        self, *, reported_versions, appeal=False, forwarded=False
+    ):
         from olympia.reviewers.models import NeedsHumanReview
 
         waffle_switch_name = (
-            'dsa-appeals-review' if appeal else 'dsa-abuse-reports-review'
+            'dsa-appeals-review'
+            if appeal
+            else 'dsa-cinder-escalations-review'
```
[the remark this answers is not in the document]
eviljeff: https://github.com/mozilla/addons-server/pull/22582/commits/d24115c015ac33d1e6cd1485a550df1ebd35302e

eviljeff:
Patch updated.  This is now the complete implementation of https://mozilla-hub.atlassian.net/browse/AMOENG-961 unless we decide to switch to use cinder's workflows in the future.

src/olympia/abuse/migrations/0038_auto_20240819_1343.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def migrate_escalate_action_to_is_forwarded(apps, schema_editor):
+    CinderDecision = apps.get_model('abuse', 'CinderDecision')
+    CinderJob = apps.get_model('abuse', 'CinderJob')
+
+    CinderJob.objects.filter(
+        decision__action=DECISION_ACTIONS.AMO_ESCALATE_ADDON
+    ).update(
+        is_forwarded=True,
```
diox, line 15: That migration should also set `resolvable_in_reviewer_tools=True` right ?

eviljeff: they should already have that - it would have been set when the `AMO_ESCALATE_ADDON` decision came in.

diox: Ah, right. I was testing with a manually created decision and I had set a bunch of fields but forgot to update the `CinderJob` I was using.

src/olympia/abuse/models.py
```diff
@@ -235,6 +227,21 @@ def notify_reporters(self, action_helper):
...
+            resolved_in_reviewer_tools=True,
+        )
+        self.notes = notes
+        job_id = entity_helper.workflow_recreate(job=self)
+        self.update(
+            job_id=job_id,
+            notes=notes,
+            resolvable_in_reviewer_tools=True,
+            is_forwarded=True,
+        )
```
diox, line 243: As discussed on slack, this feels wrong. We're potentially creating a new job in Cinder, but in AMO losing the reference to the old one. If there was anything linked to the old job AMO now see it as linked to the new one... 

In addition, this doesn't handle the fact that Cinder could be re-using an existing job, so we could end up failing some unique constraint. I think the proper way of doing this is to close the old job before opening the new one (maybe in `workflow_recreate` itself?).

Also... this should all be done in a celery task since it's going to call Cinder API.

diox (changes requested):

eviljeff:
All change - the escalate decision is retained along with the original job, and the forwarded job is created as a new instance.  (This means we don't need an explicit `is_forwarded` boolean property - there being an fk to the new job is the same thing; and we can leave the escalation notes on the previous decision)

src/olympia/abuse/models.py
```diff
@@ -235,6 +231,20 @@ def notify_reporters(self, action_helper):
...
+        new_job, _ = CinderJob.objects.update_or_create(
+            job_id=new_job_id,
+            defaults={
+                'resolvable_in_reviewer_tools': True,
+                'target_addon': self.target_addon,
+            },
+        )
+        # Update our fks to connected objects
+        AbuseReport.objects.filter(cinder_job=self).update(cinder_job=new_job)
+        AbuseReport.objects.filter(appellant_job=self).update(appellant_job=new_job)
```
eviljeff, line 244: I think this fk is unnecessary now we have the `CinderAppeal` model, but existing code relies on it so I'll remove in a follow-up.

src/olympia/abuse/models.py
```diff
@@ -235,6 +231,20 @@ def notify_reporters(self, action_helper):
...
+    def handle_job_recreated(self, new_job_id):
+        new_job, _ = CinderJob.objects.update_or_create(
+            job_id=new_job_id,
+            defaults={
+                'resolvable_in_reviewer_tools': True,
+                'target_addon': self.target_addon,
+            },
+        )
+        # Update our fks to connected objects
+        AbuseReport.objects.filter(cinder_job=self).update(cinder_job=new_job)
```
eviljeff, line 243: These fk updates feel a bit hacky... but the alternative is everywhere we check for reports attached to a job we have to also navigate down a possible `forwarded_to_job` fk - similarly for CinderDecision.appeal_job.  That seemed worse to me.

src/olympia/abuse/models.py
eviljeff: https://github.com/mozilla/addons/issues/14984

diox (approved):

PR #22583, opened 2024-08-19
Make reviewer theme queue orderable (w/ default order by due date)

diox:
Fixes: mozilla/addons#14960

### Context

Reviewer theme queue was not orderable at all (and therefore not ordered by default by due date like requested in the original ticket) because `ThemesQueueTable.Meta` was inheriting from `AddonQueueTable`, which defaults to `orderable = False`.

This change makes:
- `PendingManualApprovalQueueTable.Meta` inherit from `AddonQueueTable` as it should have been, instead of nothing, and force `orderable = True`. This is a drive-by change for consistency with 0 noticeable end-user impact.
- `ThemesQueueTable.Meta` inherit from `PendingManualApprovalQueueTable`, making it gain `orderable = True` in the process, fixing the issue.

In addition, the `verify` keyword argument of the `check_links()` function used to test queues ordering in reviewer tools have been removed - we were always passing `verify=False`...

### Testing

- Submit a few themes via devhub (you can use the `Submit a new theme` and `create a new theme version` options, which leads to the wizard, making this super easy - only need an image)
- Check that they show up in reviewer tools, ordered by `Target date` (we default to oldest first).
- Optionally, you can alter the due dates of these themes in the review page, admin tool or shell and check that the ordering still works. There is already an unit test included that does that though.

### DIFF
--- a/src/olympia/amo/tests/__init__.py
+++ b/src/olympia/amo/tests/__init__.py
@@ -147,7 +147,7 @@ def initial(form):
     return data
 
 
-def check_links(expected, elements, selected=None, verify=True):
+def check_links(expected, elements, selected=None):
     """Useful for comparing an `expected` list of links against PyQuery
     `elements`. Expected format of links is a list of tuples, like so:
 
@@ -159,9 +159,6 @@ def check_links(expected, elements, selected=None, verify=True):
 
     If you'd like to check if a particular item in the list is selected,
     pass as `selected` the title of the link.
-
-    Links are verified by default.
-
     """
     for idx, item in enumerate(expected):
         # List item could be `(text, link)`.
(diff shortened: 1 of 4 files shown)


src/olympia/amo/tests/__init__.py
```diff
@@ -171,19 +168,19 @@ def check_links(expected, elements, selected=None, verify=True):
         elif isinstance(item, str):
             text, link = None, item
 
-        e = elements.eq(idx)
+        elm = elements.eq(idx)
```
KevinMind, line 171: why? Just curious.

KevinMind (approved):
Works

https://github.com/user-attachments/assets/93aca257-65e7-4369-9ef1-9f0dc43a709d

src/olympia/amo/tests/__init__.py
diox: I hate single letter variables and tend to replace them whenever I see some (that were typically added 15 years ago). They often conflict with `ipdb` shortcuts and reduce readability for no benefit.

PR #22586, opened 2024-08-21
Refactor eula_policy action write path to be cleaner & safer

diox:
Fixes: mozilla/addons#14966

https://www.django-rest-framework.org/api-guide/viewsets/#routing-additional-http-methods-for-extra-actions

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -3015,7 +3015,7 @@ class TestVersionViewSetDetail(AddonAndVersionViewSetDetailMixin, TestCase):
         # not logged in
         assert 'source' not in self.client.get(self.url).data
 
-        user = UserProfile.objects.create(username='user')
+        user = user_factory()
         self.client.login_api(user)
 
         # logged in but not an author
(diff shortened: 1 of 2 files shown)


src/olympia/addons/tests/test_views.py
```diff
@@ -3015,7 +3015,7 @@ def test_developer_version_serializer_used_for_authors(self):
         # not logged in
         assert 'source' not in self.client.get(self.url).data
 
-        user = UserProfile.objects.create(username='user')
+        user = user_factory()
```
diox, line 3018: drive-by, I replaced all `UserProfile.objects.create(username='user')` in this test file with `user_factory()` for consistency, which amongst other things ensures the user always has an email (which in turn matters for some of the eula/policy tests because of `isSubmissionAllowedFor` permission check).

src/olympia/addons/views.py
```diff
@@ -346,20 +346,19 @@ def get_georestrictions(self):
...
-            kwargs['partial'] = True
-            method = self.update
-        return method(request, **kwargs)
+        return self.retrieve(request)
+
+    @eula_policy.mapping.patch
+    def update_eula_policy(self, request, pk=None):
+        kwargs = {'partial': True}
+        self.permission_classes = self.write_permission_classes
+        return self.update(request, **kwargs)
```
eviljeff, line 361: nit, the kwargs dict seems unnecessary?
```suggestion
        self.permission_classes = self.write_permission_classes
        return self.update(request, partial=True)
```

eviljeff (approved):

PR #22588, opened 2024-08-21
Remove chown because life is too short

KevinMind:
Fixes: mozilla/addons#14972

### Description

Trying it a different way. Disparate list of changes:
- remove "force-recreate" flag to save CPU and time restarting services that won't change
- remove supervisor PID before start to ensure we get the PID after restarting the service
- only build the worker container image, web reuses. This saves CPU and time
- remove 45s timeout on healthcheck we expect services to start quickly
- remove the chown command in entrypoint.sh as we don't need file ownership to make them executable.

### Context

We already support building either production or development versions of addons-server in the docker build. Previous PRs enabled running the web/worker services relying only on production dependencies (link) and this PR enables running without needing to install dependencies or modify file ownership.

> NOTE: by removing the ./deps:/deps volume and the chown command we are maintaining original ownership of the /deps directory. In development this means the olympia user will not "own" the files in that directory in the container. They will still be executable but running "make update_deps" will cause an error. Instead of running "update_deps" locally, instead run "make up" to rebuild the image with the newer dependencies. make up is so "thin" now, this is only marginally slower and more than makes up for it in the increased speed of make up.

Docs are updated explaining this as well as including a gotcha for the error that occurs if you do try to update deps on a running container.

### Testing

"make up" should work. First time expect 1-3 minutes. Second time should be less than 30 seconds. This will depend somewhat on your host architecture, memory CPU limitations etc. 

### Checklist


KevinMind:
https://github.com/mozilla/addons-server/blob/e8fee23c28f4d31cee65161e09b285340beaa467/src/olympia/urls.py#L110-L122

PR #22589, opened 2024-08-22
CI: Only run push locales on push events

KevinMind:
Follow up of failled deployment: https://github.com/mozilla/addons-server/actions/runs/10505053101

### Context

Currently our CI runs a locales job that extracts locales and runs the push locales command. On non master push it was running in dry mode and for any tag based from master this would be fine as it would include the automatically extracted locale strings.

But.... if you cherry pick a commit without cherry picking the extracted locales... then they are missing and locale job tries to push new strings.

This fails because we are on a detached head. There is no reason to run the push job at all on any event other than push so we can just avoid this class of bugs entirely.

We did NOT see this before because we didn't run any CI on tags before deploying them previously.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -193,7 +193,9 @@ jobs:
           run: make extract_locales
 
       - name: Push Locales
-        if:  github.event_name == 'push'
+        if:  |
+          github.event_name == 'push' ||
+          github.event_name == 'pull_request'
         shell: bash
         run: |
           is_fork="${{ needs.context.outputs.is_fork }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -206,7 +208,7 @@ jobs:
               Please submit a PR from the base repository if you are modifying l10n extraction scripts.
           EOF
           else
-            if [[ "$is_default_branch" == 'true' && "$is_push" == 'true' ]]; then
```
diox, line 209: Kinda weird edge case but if this is a PR from the default branch of the developer (e.g. PR from `master`) we still want the `dry-run`, so I think we should keep this ?

KevinMind: Fair.

KevinMind: When we originally had the condition to push only it made no sense to check true == true but now we have PR back it makes sense

diox (approved):

PR #22603, opened 2024-08-28
Set REMOTE_ADDR correctly when requests are going through a CDN Shield

diox:
Fixes: mozilla/addons#14977

### Context

We want to set `REMOTE_ADDR` from `HTTP_X_FORWARDED_FOR` depending on what the request goes through. 

Our CDN (CloudFront or Fastly) and our Load Balancer both append their own IP to `HTTP_X_FORWARDED_FOR` as the request goes through each layer. We can't just pick the leftmost value though, as the client can potentially provide that header. Thus, we have this logic that checks which layer we went through to grab the actual client IP by searching in the header from the right.

On Fastly, we have [Shielding](https://www.fastly.com/documentation/guides/concepts/shielding/) enabled. The shield is an additional PoP requests can go through, and when that happens, the IP of the shield is added to `X-Forwarded-For` so we need to shift where we're looking for the client IP one more step from the right. To figure out if we are in this situation, SRE has added a [special `X-AMO-Request-Shielded` header](https://github.com/mozilla-it/webservices-infra/pull/2261/commits/89f1f0d87f16beb6fdebd93146d9e8e9fa761a43) which is set to `true` when we are going through a shield.

### Testing

If you don't trust the unit tests, have a look at the docstring from the middleware to come up with scenarios to test locally. You can set `ENV` in your local settings to have access to `/services/client_info` and `/api/v5/services/client_info` which expose which `REMOTE_ADDR` is going to be used once we go through the middlewares and check that it's the value you're expecting to see.

### DIFF
--- a/src/olympia/amo/middleware.py
+++ b/src/olympia/amo/middleware.py
@@ -287,17 +287,24 @@ class SetRemoteAddrFromForwardedFor(MiddlewareMixin):
     Set REMOTE_ADDR from HTTP_X_FORWARDED_FOR if necessary.
 
     The request flow is either:
-    Client -> CDN -> Load balancer -> WSGI proxy -> addons-server.
+    Client -> CDN -> Load balancer -> WSGI proxy -> addons-server
     or
-    Client -> Load balancer -> WSGI proxy -> addons-server.
+    Client -> CDN -> CDN shield -> Load balancer -> WSGI proxy -> addons-server
+    or
+    Client -> Load balancer -> WSGI proxy -> addons-server
 
     Currently:
-    - CDN is CloudFront
+    - CDN is CloudFront or Fastly
+    - CDN shield is an additional PoP on the CDN that request can go through
+      (only enabled on Fastly)
     - Load Balancer is either a classic ELB (AWS) or GKE Ingress (GCP)
     - WSGI proxy is nginx + uwsgi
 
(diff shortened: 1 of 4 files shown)


src/olympia/amo/middleware.py
```diff
@@ -287,17 +287,24 @@ class SetRemoteAddrFromForwardedFor(MiddlewareMixin):
...
     or
-    Client -> Load balancer -> WSGI proxy -> addons-server.
+    Client -> CDN -> CDN shield -> Load balancer -> WSGI proxy -> addons-server
+    or
+    Client -> Load balancer -> WSGI proxy -> addons-server
 
     Currently:
-    - CDN is CloudFront
+    - CDN is CloudFront or Fastly
+    - CDN shield is an additional PoP on the CDN that request can go through
```
KevinMind, line 298: It's not clear exactly what "shielded" means in this context. Could you add a sentence explaining this?

KevinMind (approved):
Strong recommend to add a file "networking.md" in the docs and move the comments you've added there so we can surface this information more readily.

You can include a link to the middleware.py file from the doc file and at the top of the middleware file include a comment linking back to the doc.

Wdyt?

diox:
https://github.com/mozilla/addons-server/pull/22603/commits/2163e915f8a11834588e0ed92285f873d6624ae1 consolidates that documentation to a separate file in docs/

PR #22611, opened 2024-08-30
Use Bakefile to build image

KevinMind:

Fork PR: https://github.com/mozilla/addons-server/pull/22612

### Description

Using a bake file to define our build configuration accomplishes two things:

1. lifts the responsibility of if and when to builld the image during make up.
2. unifies the configuration between local and CI builds to a single file.

### Context

We have until now defined the build configuration as a part of docker compose. This led to building images locally and needing to control docker compose's behaviour via the DOCKER_PULL_POLICY. This worked, kind of.

It still leads to local images always being built, which sometimes isn't what we want.

It also means we have to define logic for the local build that doesn't always work for CI builds, spreading where the configurations need to be defined and managed.

Additionally, adding build can lead to mulitiple builds of images if multiple services inherit the build property. THis is largely solved by only applying build to the olympia service, but in a (soon) future PR I will remove olympia, reintroducing this issue. Removing build solves it for good.

Using a bake file means we can build our image with 1 exact mechanism via make up, manual make build or in CI. It also gives us more control over if and when an image build occurs.

### Testing

1. Run a local build

```bash
make up DOCKER_VERSION=local
```

Expect service is up and running with a local build.

2. After that run a "restart" skipping the build

```bash
make docker_compose_up
```

Should run really fast and just confirm the services are up. You can try manually stopping a service to verify it just reups that service. NOTE: if you stop web or worker, it will re-build.

### Checklist

### DIFF
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -66,12 +66,13 @@ runs:
 
     - name: Create .env and version.json files
       shell: bash
-      env:
-        DOCKER_TARGET: ${{ inputs.target }}
-        DOCKER_VERSION: ${{ steps.meta.outputs.version }}
-        DOCKER_COMMIT: ${{ steps.context.outputs.git_sha }}
-        DOCKER_BUILD: ${{ steps.context.outputs.git_build_url }}
-      run: make setup
+      run: |
+        echo "DOCKER_TARGET=${{ inputs.target }}" >> $GITHUB_ENV
+        echo "DOCKER_VERSION=${{ steps.meta.outputs.version }}" >> $GITHUB_ENV
+        echo "DOCKER_COMMIT=${{ steps.context.outputs.git_sha }}" >> $GITHUB_ENV
+        echo "DOCKER_BUILD=${{ steps.context.outputs.git_build_url }}" >> $GITHUB_ENV
+
+        make setup
 
     - name: Build Image
       id: build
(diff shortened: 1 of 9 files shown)


eviljeff:
> ```shell
> make up DOCKER_VERSION=local
> ```

is this change limited to when you override `DOCKER_VERSION`?  i.e. it doesn't affect (help) the normal case of running vanilla `make up`?

> After that run a "restart" skipping the build
> 
> ```shell
> make docker_compose_up
> ```

Similarly, does this only help the case where you're calling internal Makefile recipes, rather than repeating `make up ...`?

KevinMind:
> is this change limited to when you override DOCKER_VERSION? i.e. it doesn't affect (help) the normal case of running vanilla make up?

No, I only added that to ensure that your version would be "local" as that is required for a build to be triggered. Running make up, respects the previous value of the variables in .env so if you already had local as the verison, it wouldn't have been strictly necessary.

Same for the second snippet. Docker compose will infer the current values from .env when you execute the command. so running docker_compose_up will restart containers with the current configuration... what is different in this patch is it should not retrigger a build.

eviljeff (approved):
It works... and `make docker_compose_up` is fast, if you choose to run that exact command ... but `make up` (i.e what we advise as the default command) is considerably slower for me now - up from about 2.5 minutes to about 3.5 minutes. 😒

PR #22613, opened 2024-08-30
Enable Production Mode via DEBUG or docker-compose.ci.yml

KevinMind:

### Description

This PR achieves 2 goals.
1. Makes running addons-server in DEBUG=false i.e. "production" mode possible
2. Fixes existing issues with serving static files in development mode

The specific changes include:
- respecting host defined DEBUG variable to run the container in. when debug true enable serving static files from ./static directory mounted from host as well as from django static server to load dependency assets.
- modifying the way we mount static files to ensure that development mode does not require subsequent calls to update_assets in order to run addons-server and that production mode contains necessary and collected static files from site-static.
- using a package django-node-assets to load node_modules as static assets directly from our node_modules directory. This simplifies our setup code, and like the previous step ensures we do not need additional commands to run the containers in dev mode.
- move assets that we "build" ourselves to a separate directory, that is added to STATICFILES_DIRS. The allows us to freely mount ./static in development without overriding built assets.

### Context

**Why not just mount ./static and be done with it? A few problems**
- during our docker build we copy ./static but then we modify and add files. If you mount ./static again after building, the host files (which do not have the "build" assets) get removed.. so you end up back at the beginning.
- some files are available in ./static and some are in django dependencies or npm dependencies. We were using some complex scripting and mounting to make them available, but this can be simplified for development where we don't need collected static assets.

**Why not just use runserver or serve for all assest**

We use uwsgi to run django even in development because this more closely mirrors how our app runs in prod. We also cannot always rely on static_serve because in DEBUG=False (production mode) this will either not work, or even if we force it, it will be missing files.

The simpler solution is to 
1. load files from whatever is in nginx locally
2. fallback to a request to addons-server static serve.

In production mode, 1 will have all the files and so it's fine that 2 will always fail.
In development mode, 1 will have most of the files from ./static which is fast and any files that are missing will be served from serve_static.

### Testing

You should run several different iterations of addons-server and verify each is able to serve the expected static assets. This can be verified by running:
- login flow
- devhub (just load pages, it should all work)
- admin (/admin page should load)
- swagger (/api/v5/swagger. page should load)

Each scenario is toggling whether we target development or production dependencies, and whether to run in DEBUG mode or not. These are independent features and can be toggled accordingly.

Additionally you can set DEBUG directly in make commands by adding it as an argument, but the value is by default set to true and when `docker-compose.ci.yml` is included in the project the value is hard coded to "DEBUG=" regardless of the value set in the host environment or arguments.

> IMPORTANT: before running each test you should "make down" to clear any image/volume artifacts, make sure to use exact versions since you are changing a lot between each run. After running make up, you should run "make reindex" to repopulate elasticsearch indexes as they are cleared when you make down and currently not repopulated automatically

1. Development/DEBUG=True

```bash
make up DOCKER_VERSION=local DOCKER_TARGET=development COMPOSE_FILE=docker-compose.yml
```

2. Development/DEBUG=

```bash
make up DOCKER_VERSION=local DOCKER_TARGET=development COMPOSE_FILE=docker-compose.yml:docker-compose.ci.yml
```

3. Production/DEBUG=True

```bash
make up DOCKER_VERSION=local DOCKER_TARGET=production COMPOSE_FILE=docker-compose.yml
```

4. Production/DEBUG=

```bash
make up DOCKER_VERSION=local DOCKER_TARGET=production COMPOSE_FILE=docker-compose.yml:docker-compose.ci.yml
```

### Checklist


diox (changes requested):
As mentioned on slack, we need a way for assets to load using `DEBUG=False`. `SERVE_STATIC` variable is a good start but not enough: At the moment, with this patch, if I've got `DEBUG=False` set, then the URLs we are going to be referencing are going to be the ones with the hash that are in `site-static/`, but `serve()` only knows how to serve the source stuff from `static/`, not the built ones.

With `DEBUG=False` and `SERVE_STATIC=True` http://olympia.test/static/img/favicon.3cec07462a3d.ico fails for me for instance (and all js/css/images as well obviously).

We need some way of handling that scenario (where the developer wants a configuration closer to production - with "real" compiled assets, while still being able to develop locally, either to build some feature/fix some bug that requires real assets, or to reproduce an issue with production).

KevinMind:
@diox it is ready for review and QA. Note, you either need a stage account for fxa or you need to force use_fake_fxa to return true for the 2nd and 4th case where we use `docker-compose.ci.yml` as this sets DEBUG= and will trigger real auth.

src/olympia/lib/settings_base.py
```diff
@@ -1289,7 +1289,23 @@ def read_only_mode(env):
...
+    'django.contrib.staticfiles.finders.FileSystemFinder',
+    'django.contrib.staticfiles.finders.AppDirectoriesFinder',
+    'django_node_assets.finders.NodeModulesFinder',
+)
+
+NODE_MODULES_ROOT = os.path.join('/', 'deps', 'node_modules')
+NODE_PACKAGE_JSON = os.path.join('/', 'deps', 'package.json')
+NODE_PACKAGE_MANAGER_INSTALL_OPTIONS = ['--dry-run']
+
+STATIC_DIST_PATH = os.path.join('/', 'data', 'olympia', 'static-dist')
```
diox, line 1302: As discussed on slack, let's rename the constant and the directory to `static-build` (and `STATIC_BUILD_PATH`) and document what it is for.

src/olympia/urls.py
```diff
@@ -120,6 +123,15 @@
                 serve_static,
                 {'document_root': settings.MEDIA_ROOT},
             ),
+            re_path(r'^static/(?P<path>.*)$', serve_static_files),
```
diox, line 126: Document that it's only meant as a fallback

docker/nginx/addons.conf
```diff
@@ -11,29 +11,15 @@ server {
     }
 
     location /static/ {
-        alias /srv/static/;
-    }
+         alias /srv/site-static/;
 
-    location /site-static/ {
-        alias /srv/site-static/;
+         error_page 404 = @olympia;
```
diox, line 16: Add small comment explaining that this is the fallback for files not directly available

docker-compose.yml
```diff
@@ -87,8 +92,7 @@ services:
     image: nginx
     volumes:
       - ./docker/nginx/addons.conf:/etc/nginx/conf.d/addons.conf
-      - ./static:/srv/static
-      - ./site-static:/srv/site-static
+      - ./static:/srv/site-static
```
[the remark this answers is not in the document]
KevinMind: we mount either "./static" or "/data/olympia/site-static" depending on which set of compose files you use.

As far as nginx is concerned it serves from one location but the source of the files changes.

docs/topics/development/static-files.md
```diff
@@ -0,0 +1,76 @@
...
+2. Django's built-in development server
+
+In development, the nginx server will attempt to serve static files from the `./static` directory mounted into the nginx cointainer.
+If the file cannot be found there the request is forwarded to django.
+Nginx serves our own static files quickly and any vendor files can be fetched from django directly during development.
+
+In production mode, we mount a data volume both to `web` anb `nginx` containers.
+The `web` container exposes the `site-static` directory to nginx that includes the collected static files.
+
+> In actual production environments, we upload the static files to a cloud bucket and serve them directl from the static path.
```
diox, line 28:
```suggestion
> In actual production environments, we upload the static files to a cloud bucket and serve them directly from the static path.
```

docs/topics/development/static-files.md
```diff
@@ -0,0 +1,76 @@
...
+
+### npm Dependencies
+
+We have a (complex) set of npm static assets that are built by the `compress_assets` management command.
+During development, these assets are served directly from the node_modules directory using a custom static finder.
+
+## DEBUG Property and Static File Serving
+
+The behavior of static file serving can be controlled using the `DEBUG` environment variable.
+This is set in the environment, and in CI environments, it's controlled by the `docker-compose.ci.yml` file.
```
diox, line 66: Before your changes the way to modify `DEBUG` used to be through the `local_settings.py` file only. Which still works, but can be confusing because it will take precedence over the environment, so it will overwrite whatever is set in `docker-compose.ci.yml`... So I think it deserves a mention somewhere as a possible gotcha

diox (approved):

docs/topics/development/static-files.md
KevinMind: Done.

PR #22622, opened 2024-08-30
Create release with template and release notes

KevinMind:
Fixes: <https://mozilla-hub.atlassian.net/browse/AMOENG-616>

### Description

Adds a workflow to trigger the creation of a draft release. This allows us to have a standardized template for the information in our releases.

### Context

Previously we created release documentation manually and updated with commit diffs of tags and info related to a given release. Now that we can release via github releases we should automate creation of release docs in the same way.

### Testing

The criteria for the release is:
- the release hero is set to the current actor. It can be updated post release but typically the release hero will trigger this workflow 
- the tag is based on an input parameter which is validated:
- date YYYY.MM.DD with or without the -X suffix indicating a cherry-pick release.
- The tag is validated to not match the tag of an existing release. Cannot release the same tag twice (you can deploy the same tag twice, but that is different)
- Link to the latest published release is included.

> NOTE: Release notes with commits and diffs are only created when we click the "generate notes" button. we should only do that when we actually roll out to stage.

### Checklist

### DIFF
--- a/.github/release-template.md
+++ b/.github/release-template.md
@@ -1,0 +1,18 @@
+This week's push hero is @{{GITHUB_USER}}
+
+Latest Release: [{{LATEST_TAG}}]({{LAST_RELEASE_URL}})
+
+## Blockers:
+
+## Cherry-picks:
+
+<!-- Link to the actual commits, NOT merge commits. The commits need to appear
+in chronological order so that `git cherry-pick` will apply them correctly. -->
+
+## Before we push:
+
+## Before we start:
+
+## Before we promote:
+
+## After we're done:
--- a/.github/release-template.md
+++ b/.github/release-template.md
@@ -1,6 +1,6 @@
 This week's push hero is @{{GITHUB_USER}}
 
-Latest Release: [{{LATEST_TAG}}]({{LAST_RELEASE_URL}})
+Previous Release: [{{PREVIOUS_TAG}}]({{PREVIOUS_RELEASE_URL}})
 
 ## Blockers:
 
--- a/.github/release-template.md
+++ b/.github/release-template.md
@@ -16,3 +16,13 @@ in chronological order so that `git cherry-pick` will apply them correctly. -->
 ## Before we promote:
 
 ## After we're done:
+
+## Addons-Frontend Changelog:
+
+<!-- Link to the tag comparison of the target tag and the previous tag for addons-frontend
+https://github.com/mozilla/addons-frontend/compare/<PREVIOUS_TAG...<CURRENT_TAG>
+-->
+
+## Addons Server Changelog:
+
+<!-- This section will be automatically populated once the release notes are auto generated -->
--- a/.github/workflows/draft_release.yml
+++ b/.github/workflows/draft_release.yml
@@ -41,8 +41,8 @@ jobs:
             tag="${{ github.event.inputs.tag }}"
           fi
 
-          #Validate the tag is formatted correctly YYYY.MM.DD or YYYY.MM.DD-rcX
-          # where rcX is a whole number greater than zero
+          # Validate the tag is formatted correctly YYYY.MM.DD or YYYY.MM.DD-X
+          # where X is a whole number greater than zero
           if [[ ! $tag =~ ^[0-9]{4}\.[0-9]{2}\.[0-9]{2}(-[0-9]+)?$ ]]; then
             echo "Invalid tag format. Must be YYYY.MM.DD or YYYY.MM.DD-X"
             exit 1
--- a/.github/workflows/draft_release.yml
+++ b/.github/workflows/draft_release.yml
@@ -1,0 +1,72 @@
+name: Draft Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Release date YYYY.MM.DD Also used to generate the tag name.'
+        required: true
+
+jobs:
+  draft_release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Create Release Draft
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        shell: bash
+        run: |
+          # Format current date as YYYY.MM.DD
+
+          if [[ "${{ github.event_name}}" == 'workflow_dispatch' ]]; then
+            tag="${{ github.event.inputs.tag }}"
+          else
+            tag=$(date +'%Y.%m.%d')
+          fi
+
+          #Validate the tag is formatted correctly YYYY.MM.DD or YYYY.MM.DD-rcX
+          # where rcX is a whole number greater than zero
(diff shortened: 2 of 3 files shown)


KevinMind:
Won't recreate existing tag https://github.com/mozilla/addons-server/actions/runs/10635520149/job/29485287048

.github/workflows/draft_release.yml
```diff
@@ -0,0 +1,72 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        shell: bash
+        run: |
+          # Format current date as YYYY.MM.DD
+
+          if [[ "${{ github.event_name}}" == 'workflow_dispatch' ]]; then
+            tag="${{ github.event.inputs.tag }}"
+          else
+            tag=$(date +'%Y.%m.%d')
```
KevinMind, line 25: Comment this out and add an error. this should only be used for testing PRs.

eviljeff: I don't understand what you mean here - why would PRs be creating releases?

.github/workflows/draft_release.yml
```diff
@@ -0,0 +1,72 @@
...
+          fi
+
+          #Validate the tag is formatted correctly YYYY.MM.DD or YYYY.MM.DD-rcX
+          # where rcX is a whole number greater than zero
+          if [[ ! $tag =~ ^[0-9]{4}\.[0-9]{2}\.[0-9]{2}(-rc[0-9]+)?$ ]]; then
+            echo "Invalid tag format. Must be YYYY.MM.DD or YYYY.MM.DD-rcX"
+            exit 1
+          fi
```
eviljeff, line 33: What's the `YYYY.MM.DD-rcX` for?  It's new.

.github/release-template.md
```diff
@@ -0,0 +1,18 @@
+This week's push hero is @{{GITHUB_USER}}
+
+Latest Release: [{{LATEST_TAG}}]({{LAST_RELEASE_URL}})
```
eviljeff, line 3: Isn't this always going to be wrong, if we're creating the release notes before the tag? Or is the idea we'd update it manually afterwards?

.github/release-template.md
```diff
@@ -0,0 +1,18 @@
+This week's push hero is @{{GITHUB_USER}}
```
eviljeff, line 1: Shouldn't this be an input too?  The current push hero (who creates this doc) is very rarely also the person doing it next time.

eviljeff (commented):
Some things I'm not clear on:
- how we're going to handle addons-frontend releases - will we just manually add the diff link in here, or something else?
- How the generated full changelog can work when there is an existing release note, with potentially important information that can't be overwritten with an entirely freshly generated release notes.

.github/workflows/draft_release.yml
KevinMind: IF you need to work on this yml file, you might want to see what release would be produced. I think adding that code into master is a bad idea though as it lead to the incorrect date on the release doc you were asking about earlier.

.github/workflows/draft_release.yml
KevinMind: Need to support cherry pick releases. 2024.09.05-23 for example. Good catch though, I accidentally included rc in the regex.. will remove.

.github/release-template.md
KevinMind: That's a good point. Kind of unfortunate though because you could end up doing it wrong... I actually think you can use select inputs in workflow dispatch so maybe that would be better. Hardcode the github handles and then you just select the right one.

.github/release-template.md
KevinMind: Latest release points to the most recent release tagged "latest" which all of them are, so effectively it is the "last" release. Each time you create a new release, it links to the previous release. Should we call it "previous" instead of "latest"?

KevinMind:
> How the generated full changelog can work when there is an existing release note, with potentially important information that can't be overwritten with an entirely freshly generated release notes.

You generate the changelog when publishing the release. It appends the commits after any other content already on the release doc. It won't override anything.

Because releases are essentially immutable, you never change the tag on a release but create a new one, you never should need to update the changelog.

Does that clarify?

diox:
> You generate the changelog when publishing the release. It appends the commits after any other content already on the release doc. It won't override anything.
> Because releases are essentially immutable, you never change the tag on a release but create a new one, you never should need to update the changelog.

There are 2 kinds of cherry-picks though:
1. Before the push to prod happened. In this scenario we tagged and deployed release A to prod 2 weeks ago, then tagged release B, deployed it to stage, but then cherry-picked before deploying B to prod. Let's call that new tag B-1. When B-1 is created, we want the companion doc to contain the diff link between the tag from A to B-1, and all previous info that was in the original doc for B. Original B tag then no longer matters and will never be deployed to prod.
2. After the push to prod happened. In this scenario we already deployed B to prod, then cherry-picked, creating B-1. The release notes should be an entirely new doc with no previous text and diff link between B and B-1.

.github/release-template.md
eviljeff: Yeah.  A select on our 3 github handles would be ideal, but being able to manually edit to choose the correct one is fine.

.github/release-template.md
eviljeff: Previous is going to be correct every time we're creating a draft release, yeah.

eviljeff:
> It appends the commits after any other content already on the release doc. It won't override anything.

👍

KevinMind:
@diox both of your proposed diffs are achievable. I don't think it will be possible to get this granular with the automation, but that is probably okay. If you are making a cherry pick release I think it is up to you to verify your target and compare tags are correct, and you can update or remove the link as you wish.

I have included info like this in the PR updating our push duty docs. Do you think this is sufficient? Or would you suggest removing the link? Or trying to make the automation support more complex linking?

.github/release-template.md
KevinMind: Updated with the select.

.github/release-template.md
KevinMind: It's not clear if you are asking for a change or not? Is it fine to leave latest, or should I update to say "previous"?

eviljeff: I think it should be changed to previous release, at least for draft releases.

eviljeff (approved):
r+ if @diox concurs

.github/workflows/draft_release.yml
```diff
@@ -0,0 +1,88 @@
...
+
+          # Format current date as YYYY.MM.DD
+
+          if [[ "${{ github.event_name}}" == 'workflow_dispatch' ]]; then
+            tag="${{ github.event.inputs.tag }}"
+          fi
+
+          #Validate the tag is formatted correctly YYYY.MM.DD or YYYY.MM.DD-rcX
+          # where rcX is a whole number greater than zero
```
diox, line 45:
```suggestion
          # Validate the tag is formatted correctly YYYY.MM.DD or YYYY.MM.DD-X
          # where X is a whole number greater than zero
```

PR #22625, opened 2024-09-02
dont report abuse reports for unlisted versions to cinder

eviljeff:
Fixes: mozilla/addons#14985

### Description

The check for whether an abuse report should be reported to cinder has been extracted out to a function so we can test it more thoroughly.

### Context

There may also be future work to reduce the reports we are submitting, so having the logic for that in a contained place that can be tested is helpful for that also.  I additionally added a query on `guid`, because previously we would create a task for guids that are unknown to AMO, even though there wasn't anything we could possibly review from an unknown guid.

### Testing

- prerequisites
  -  `CINDER_API_TOKEN` set to a valid API token
  - waffle switch `dsa-job-technical-processing` enabled
- install an unlisted version of an add-on
  - either sign the unlisted version locally, and install that xpi (takes some setup in Firefox to be able to install them); or update a local Addon instance to have a matching guid, and a version with a matching version number (that is unlisted channel, and `is_signed=True`)
- report that add-on for abuse
- see the abuse report exists but no `CinderJob` instance (or errors raised in logs)
- repeat with a url pointing to guid that doesn't exist locally (same - AbuseReport instance, but no CinderJob)
- check the default case is still working - report for an abuse an add-on you _do_ have in your local database, and see there _is_ a CinderJob created, with a cinder_id set.

### Checklist

### DIFF
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -794,11 +794,14 @@ class AbuseReport(ModelBase):
                 not self.guid
                 or (
                     Addon.unfiltered.filter(guid=self.guid).exists()
-                    and not Version.unfiltered.filter(
-                        addon__guid=self.guid,
-                        version=self.addon_version,
-                        channel=amo.CHANNEL_UNLISTED,
-                    ).exists()
+                    and (
+                        not self.addon_version
+                        or Version.unfiltered.filter(
+                            addon__guid=self.guid,
+                            version=self.addon_version,
+                            channel=amo.CHANNEL_LISTED,
+                        ).exists()
+                    )
                 )
             )
         )
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -20,7 +20,7 @@ from olympia.constants.abuse import (
 )
 from olympia.ratings.models import Rating
 from olympia.users.models import UserProfile
-from olympia.versions.models import VersionReviewerFlags
+from olympia.versions.models import Version, VersionReviewerFlags
 
 from .cinder import (
     CinderAddon,
@@ -784,6 +784,24 @@ class AbuseReport(ModelBase):
             return self.collection
         else:
             return self.addon
+
+    @property
+    def is_reportable(self):
+        """Is this abuse report reportable under DSA, so should be sent to Cinder"""
+        return bool(
+            self.reason in AbuseReport.REASONS.REPORTABLE_REASONS
+            and (
+                not self.guid
+                or (
+                    Addon.unfiltered.filter(guid=self.guid).exists()
+                    and not Version.unfiltered.filter(
+                        addon__guid=self.guid,
+                        version=self.addon_version,
+                        channel=amo.CHANNEL_UNLISTED,
+                    ).exists()
+                )
+            )
+        )
 
     @property
     def is_handled_by_reviewers(self):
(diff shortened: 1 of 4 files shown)


src/olympia/abuse/models.py
```diff
@@ -785,6 +785,24 @@ def target(self):
         else:
             return self.addon
 
+    @property
+    def is_reportable(self):
```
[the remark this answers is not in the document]
diox: Alternatively... maybe `is_directly_actionable` or something along those lines to be more generic ?

src/olympia/abuse/models.py
```diff
@@ -785,6 +785,24 @@ def target(self):
...
+            self.reason in AbuseReport.REASONS.REPORTABLE_REASONS
+            and (
+                not self.guid
+                or (
+                    Addon.unfiltered.filter(guid=self.guid).exists()
+                    and not Version.unfiltered.filter(
+                        addon__guid=self.guid,
+                        version=self.addon_version,
+                        channel=amo.CHANNEL_UNLISTED,
+                    ).exists()
```
diox, line 801: This should be done the opposite way, looking for the version that exists in the listed channel. That way:
- If we ever add more channels, it will be an explicit decision to make them reportable (we are more likely to add channels that don't matter under the DSA than the opposite)
- Non-existing versions that are somehow reported by users won't be sent to Cinder (nothing can be done about them, since they don't exist, so they shouldn't be sent to Cinder)

(Also adjust tests accordingly)

src/olympia/abuse/models.py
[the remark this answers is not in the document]
eviljeff: It was supposed to be more "is reportable under DSA" but I agree that report is overused.  `is_individually_actionable` or ``is_individually_actioned`?

src/olympia/abuse/models.py
eviljeff: I wrote it deliberately that way as that's what @abyrne-moz specified in the issue.  I don't disagree with your logic though.

src/olympia/abuse/models.py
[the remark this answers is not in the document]
diox: `is_individually_actionable` sounds good to me

src/olympia/abuse/models.py
diox: I think inverting the logic like I'm suggesting also satisfies the criteria specified in the issue while having the advantages I've highlighted

eviljeff: It does, but it opens more cases.  Such as reporting an add-on without addon_version supplied (i.e. what happens if you don't have the add-on installed at the time you report it, or use another browser) - should those cases be treated as if you reported a listed version?  e.g. Should we add another query to verify there is at least one listed version?

src/olympia/abuse/models.py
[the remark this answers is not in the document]
eviljeff: https://github.com/mozilla/addons-server/pull/22625/commits/5917a2fc7163c7ee4c3666f5bbae8a1ab8a7e7aa

src/olympia/abuse/models.py
eviljeff: https://github.com/mozilla/addons-server/pull/22625/commits/88243020ed2247222c29ed4079a9500baa79fbd8

diox (approved):

PR #22627, opened 2024-09-03
Remove obsolete validator caching

diox:
This was only really relevant when validation could be triggered arbitrarily by reviewers for an existing File(Upload) that was laying around unvalidated.

That confusing and dangerous functionality no longer exists, FileUpload validation is always done once at upload and can't be triggered later.

When we still had post-request-task, that caching was broken for years and nobody noticed because the code path where we pulled the task id from the cache was never used.

Fixes mozilla/addons#14995

### DIFF
--- a/src/olympia/devhub/tasks.py
+++ b/src/olympia/devhub/tasks.py
@@ -9,7 +9,6 @@ from functools import wraps
 from zipfile import BadZipFile
 
 from django.conf import settings
-from django.core.cache import cache
 from django.core.files.storage import default_storage as storage
 from django.core.validators import ValidationError
 from django.db import transaction
@@ -19,7 +18,6 @@ from django.utils.encoding import force_str
 from django.utils.translation import gettext
 
 import waffle
-from celery.result import AsyncResult
 from django_statsd.clients import statsd
 
 import olympia.core.logger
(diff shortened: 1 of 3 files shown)


diox:
1. Validation is a complex process, it's actually several tasks grouped together that repack the zip, call external services etc, so it needs to happen asynchronously to avoid eating all CPU on webservers. We can't block the request for long enough (it would timeout/504 for large files, or accidentally be cancelled in case of network errors, etc) so we create the `FileUpload`, trigger the task, return the `FileUpload` uuid, and clients poll for results for results with that. The `Validator` class usefulness is debatable, but nowadays it's short enough that it doesn't bother me personally.
2. Running validation tasks modify the `FileUpload` instance, so it shouldn't be triggered twice. The `Validator` class actually has an assertion that prevents double-validation (`assert not file_.validation`). Furthermore when the `Version` is created we actually move the file away so it's not possible to re-use a `FileUpload` that has been used to create a `File`/`Version` already in any meaningful way.

All code to re-trigger validation has been removed in https://github.com/mozilla/addons/issues/7012 which I linked in the issue. I tweaked the wording of the milestone in the acceptance criteria, going through an upload and checking it's validated correctly (and allows you to create a new version, if you want to go deeper) should be enough to verify that everything works.

KevinMind (approved):

PR #22631, opened 2024-09-03
Use default buildx builder for faster build

KevinMind:

### Description

Switch to using the default builder for faster builds

### Context

We have been using a custom builder for building our docker image. THis dates back to how we built in circleCI, but the basic point is it is WAY faster to use the default builder on the host.

Essentially it means we don't have to create a new container, share data from the host to that container, and then also export an image back to the host. we can just run the build directly on the host.

Techincally, with the custom builder we have better control over caching and target architecture, but the bake file seems to handle specifying the target architecture and I have verified that caching is working as expected.

### Testing

> NOTE: this is a breaking change. to enable this build even after merged you will have to run

```bash
docker buildx use default
```

The make target removed in this PR `create_docker_builder` effectively told buildx to use the custom builder, even when not specified and so you will continue to use the builder until you manually switch back to the default.

I considered adding logic to automate that, but it's just more code and... well.. if you wantz blazing fast... you typez four words in terminal. 

### Checklist


.gitignore
```diff
@@ -50,6 +50,7 @@ storage/shared_storage/*
 storage/sitemaps/*
 tmp/*
 deps/*
+metadata.json
```
eviljeff, line 53: we're going to need some docs to explain what this metadata.json file is (and a less vague name) - I can see anywhere it's written, only read.

KevinMind (commented):
> we're going to need some docs to explain what this metadata.json file is (and a less vague name) - I can see anywhere it's written, only read. 

I've added some doc updates. LMKWYT.

.gitignore
KevinMind: I've added some doc updates. LMKWYT.

eviljeff: Can we have a more specific name?  metadata.json could be anything.

KevinMind: Done.

.github/actions/build-docker/action.yml
```diff
@@ -71,28 +61,25 @@ runs:
         echo "DOCKER_VERSION=${{ steps.meta.outputs.version }}" >> $GITHUB_ENV
         echo "DOCKER_COMMIT=${{ steps.context.outputs.git_sha }}" >> $GITHUB_ENV
         echo "DOCKER_BUILD=${{ steps.context.outputs.git_build_url }}" >> $GITHUB_ENV
+        echo "TAGS_FILE=${{ steps.meta.outputs.bake-file-tags }}" >> $GITHUB_ENV
+        echo "ANNOTATIONS_FILE=${{ steps.meta.outputs.bake-file-annotations }}" >> $GITHUB_ENV
+        echo "DOCKER_METADATA_FILE=buildx-bake-buildx-bake-buildx-bake-metadata.json" >> $GITHUB_ENV
```
diox, line 66: `buildx-bake-buildx-bake-buildx-bake-metadata.json` ? Why not just `buildx-bake-metadata.json` ?

.gitignore
```diff
@@ -61,7 +62,7 @@ private/
 !deps/.keep
 
 # Ignore buildx bake output files
-buildx-bake-metadata.json
+buildx-bake-buildx-bake-metadata.json
```
diox, line 65: What's this one ?

.github/actions/build-docker/action.yml
KevinMind: Good catch.

.gitignore
KevinMind: Looks like a broken find/replace

diox (approved):

PR #22640, opened 2024-09-04
Reviewer Tools Pastebin Re-implementation

chrstinalin:
Fixes: mozilla/addons#14997

### Description

Adds an `AttachmentLog` model and ability to either upload a file (currently limited to `.txt`) or directly paste a build log (stored as a `.txt` file) from any reviewer action. Attachments can then be downloaded in the add-on history.

image
image

### Checklist


src/olympia/reviewers/forms.py
```diff
@@ -299,7 +302,25 @@ def create_option(self, *args, **kwargs):
...
 
 
-class ReviewForm(forms.Form):
+class WithAttachmentMixin:
+    def get_invalid_attachment_file_type_message(self):
+        valid_extensions_string = '(%s)' % ', '.join(VALID_ATTACHMENT_EXTENSIONS)
+        return gettext(
+            'Unsupported file type, please upload an archive '
+            'file {extensions}.'.format(extensions=valid_extensions_string)
+        )
```
eviljeff, line 311: I think you can include this directly within `clean_attachment_file` - unless you're going to need to call it elsewhere and/or it's difficult to test otherwise

src/olympia/reviewers/forms.py
```diff
@@ -299,7 +302,25 @@ def create_option(self, *args, **kwargs):
         return option
 
 
-class ReviewForm(forms.Form):
+class WithAttachmentMixin:
```
eviljeff, line 305: Can you implement this as a [validator](https://docs.djangoproject.com/en/4.2/ref/validators/#writing-validators) instead?  (named `validate_review_attachment` or similar)

src/olympia/reviewers/tests/test_utils.py
```diff
@@ -1162,6 +1164,22 @@ def test_log_action_override_user(self):
         assert logs.count() == 1
         assert logs[0].user == task_user
 
+    def test_log_action_attachment_input(self):
+        assert AttachmentLog.objects.count() == 0
+        data = self.get_data()
+        data['attachment_input'] = 'This is input'
+        self.helper.set_data(data)
+        self.helper.handler.log_action(amo.LOG.REJECT_VERSION)
+        assert AttachmentLog.objects.count() == 1
```
eviljeff, line 1173: assert the content too

src/olympia/reviewers/tests/test_utils.py
```diff
@@ -1162,6 +1164,22 @@ def test_log_action_override_user(self):
...
+
+    def test_log_action_attachment_file(self):
+        assert AttachmentLog.objects.count() == 0
+        data = self.get_data()
+        data['attachment_file'] = ContentFile("I'm a text file", name='attachment.txt')
+        self.helper.set_data(data)
+        self.helper.handler.log_action(amo.LOG.REJECT_VERSION)
+        assert AttachmentLog.objects.count() == 1
```
eviljeff, line 1181: same, assert the content too

src/olympia/reviewers/tests/test_views.py
```diff
@@ -2611,6 +2611,52 @@ def test_reviewer_reply(self):
...
+        response = self.client.post(
+            self.url,
+            {
+                'action': 'reply',
+                'comments': 'hello sailor',
+                'attachment_file': attachment,
+            },
+        )
+        assert response.status_code != 302
+        assert AttachmentLog.objects.count() == 0
```
eviljeff, line 2658: assert the error message is in the response.

src/olympia/reviewers/tests/test_views.py
```diff
@@ -2611,6 +2611,52 @@ def test_reviewer_reply(self):
...
+        response = self.client.post(
+            self.url,
+            {
+                'action': 'reply',
+                'comments': 'hello sailor',
+                'attachment_input': 'babys first build log',
+            },
+        )
+        assert response.status_code == 302
+        assert AttachmentLog.objects.count() == 1
```
eviljeff, line 2630: assert the content is there

src/olympia/reviewers/tests/test_views.py
```diff
@@ -2611,6 +2611,52 @@ def test_reviewer_reply(self):
...
+        response = self.client.post(
+            self.url,
+            {
+                'action': 'reply',
+                'comments': 'hello sailor',
+                'attachment_file': attachment,
+            },
+        )
+        assert response.status_code == 302
+        assert AttachmentLog.objects.count() == 1
```
eviljeff, line 2644: assert the content is there

src/olympia/reviewers/utils.py
```diff
@@ -1017,6 +1018,17 @@ def log_action(
...
         self.log_entry = ActivityLog.objects.create(action, *args, **kwargs)
 
+        attachment = None
+        if self.data.get('attachment_file'):
+            attachment = self.data.get('attachment_file')
+        elif self.data.get('attachment_input'):
+            # The name will be overridden later by attachment_upload_path.
+            attachment = ContentFile(
+                self.data['attachment_input'], name='attachment.txt'
+            )
```
eviljeff, line 1028: So you can't provide both the text _and_ upload a file?

src/olympia/urls.py
```diff
@@ -43,6 +44,8 @@
     re_path(r'^uploads/', include(upload_patterns)),
     # Downloads.
     re_path(r'^downloads/', include(download_patterns)),
+    # Activity.
+    re_path(r'^activity/', include(attachment_patterns)),
```
eviljeff, line 48: I'd lean towards putting the url under `developers/` but don't know if @diox  has a counter opinion.

src/olympia/activity/models.py
```diff
@@ -39,6 +42,20 @@
 GENERIC_USER_NAME = gettext('Add-ons Review Team')
 
 
+def attachment_upload_path(instance, filename):
+    ext = os.path.splitext(filename)[1]
+    timestamp = datetime.now().replace(microsecond=0)
+    return os.path.join(
+        'activity_attachment',
+        id_to_path(instance.activity_log.pk, breadth=2),
```
eviljeff, line 50: the default `breath` of `1` is enough

src/olympia/activity/views.py
```diff
@@ -166,3 +171,26 @@ def check_content_length(request):
     spam_rating = request.data.get('SpamScore', 0.0)
     process_email.apply_async((message, spam_rating))
     return Response(data=validation_response, status=status.HTTP_201_CREATED)
+
+
+@non_atomic_requests
+def download_attachment(request, log_id):
```
eviljeff, line 177: this needs a test

eviljeff (changes requested):

src/olympia/reviewers/utils.py
chrstinalin: It was mentioned in Zoom before that only one file should be attached to each reply -- I don't see where you would need to do this?

I did consider placing a button beforehand to get the user to consciously choose between the two, but uploading a file has more conscious choices (upload -> select file -> submit) where if a file _did_ exist to take, is more likely to be the one the user wants to upload. But I could always just do the button.

eviljeff: @diox to clarify/confirm here.

If we _are_ limiting to either the textfield input or an attachment there should be some form validation added to error if both are provided - silently dropping something the reviewer has entered into the textfield isn't great.  (And there could be some better UX, that makes it clear when you provide one of them that you can't provide the other - e.g. disabling fields, etc - but that would be follow-up work)

src/olympia/urls.py
diox: `activity/` makes slightly more sense to me, it's an activity attachment after all, and it's available to reviewers, not just developers.

The cleaner way would then be to:
- Move the existing `activity/urls.py` to `activity/api_urls.py`
- Re-create the `activity/urls.py` file and add a fresh `urlpatterns` definition to it with that new URL
- Include it with `re_path(r'activity/', include('olympia.activity.urls'))` here.

Note that you would also need to add `activity` to `SUPPORTED_NONAPPS` and `SUPPORTED_NONLOCALES` settings (as well as the equivalent addons-frontend config variables)

src/olympia/reviewers/utils.py
diox: We should not allow both at the same time, and there should be some explicit error about it.

src/olympia/urls.py
eviljeff: (I was mainly thinking of a way of avoiding having to add another top level app/folder to configurations everywhere)

diox: It's a little bit annoying but I think that's worth doing to avoid for consistency (and the less reviewers functionality depend on developer URLs/views/etc the better IMHO)

chrstinalin: > Note that you would also need to add activity to SUPPORTED_NONAPPS and SUPPORTED_NONLOCALES settings (as well as the equivalent addons-frontend config variables)

I did notice that doing this caused the download link to no longer work, but I'm assuming its due to the changes not being synced?

src/olympia/reviewers/utils.py
```diff
@@ -1019,6 +1019,9 @@ def log_action(
         self.log_entry = ActivityLog.objects.create(action, *args, **kwargs)
 
         attachment = None
+        if self.data.get('attachment_file') and self.data.get('attachment_input'):
+            raise RuntimeError('Cannot upload both a file and input.')
+            ...
```
eviljeff, line 1024: `...`? 🤨

src/olympia/reviewers/tests/test_views.py
```diff
@@ -2656,6 +2665,7 @@ def test_attachment_invalid_upload(self):
         )
         assert response.status_code != 302
         assert AttachmentLog.objects.count() == 0
+        self.assertIn('Unsupported file type', response.content.decode('utf-8'))
```
eviljeff, line 2668: Can you assert the whole error?

I think you'll find it'd be something like `Unsupported file type, please upload an archive file (.txt)` - which is a little wrong as `.txt` files aren't archives.

src/olympia/reviewers/forms.py
```diff
@@ -302,25 +302,20 @@ def create_option(self, *args, **kwargs):
...
+                    'Unsupported file type, please upload an archive '
+                    'file {extensions}.'.format(extensions=valid_extensions_string)
                 )
-        return file
+            )
+    return value
 
 
-class ReviewForm(WithAttachmentMixin, forms.Form):
+class ReviewForm(forms.Form):
```
eviljeff, line 318: We still want validation that both attachment_file and attachment_input aren't provided simultaneously.  Add that check to `def clean`.

src/olympia/activity/urls.py
```diff
@@ -1,16 +1,9 @@
...
     re_path(r'^mail/', views.inbound_email, name='inbound-email-api'),
 ]
-
-attachment_patterns = [
-    re_path(
-        r'^attachment/(?P<log_id>\d+)',
-        views.download_attachment,
-        name='activity.attachment',
-    )
-]
```
eviljeff: This change isn't what Mat intended with his request.  

The end result should be 
```python
urlpatterns = [
    re_path(
        r'^attachment/(?P<log_id>\d+)',
        views.download_attachment,
        name='activity.attachment',
    )
]
```
in this file, and 
```python
urlpatterns = [
    re_path(r'^mail/', views.inbound_email, name='inbound-email-api'),
]
```
in api_urls.py.  

You'll then need to change all the `olympia.activity.urls` in https://github.com/mozilla/addons-server/blob/master/src/olympia/api/urls.py to be `olympia.activity.api_urls` instead

eviljeff (changes requested):

src/olympia/reviewers/forms.py
```diff
@@ -299,6 +302,19 @@ def create_option(self, *args, **kwargs):
...
 
 
+def validate_review_attachment(value):
+    if value:
+        if not value.name.endswith(VALID_ATTACHMENT_EXTENSIONS):
+            valid_extensions_string = '(%s)' % ', '.join(VALID_ATTACHMENT_EXTENSIONS)
+            raise forms.ValidationError(
+                gettext(
+                    'Unsupported file type, please upload an archive '
+                    'file {extensions}.'.format(extensions=valid_extensions_string)
```
eviljeff, line 312: This error string is incorrect - a `.txt` file is not an archive.

src/olympia/reviewers/utils.py
```diff
@@ -1017,6 +1018,19 @@ def log_action(
         kwargs = {'user': user or self.user, 'created': timestamp, 'details': details}
         self.log_entry = ActivityLog.objects.create(action, *args, **kwargs)
 
+        attachment = None
+        if self.data.get('attachment_file') and self.data.get('attachment_input'):
+            raise RuntimeError('Cannot upload both a file and input.')
```
eviljeff, line 1023: We don't need this extra validation here - the form is already validating the inputs. (If we did, we typically just use `assert xxx`)

eviljeff (approved):
r+wc - just two minor things to change.

PR #22641, opened 2024-09-05
stop ignoring ruff rule B904

eviljeff:
Fixes: mozilla/addons#1931

### Description

Chain all the exceptions raised within `except` blocks to the parent exception - everywhere we re-raise it seemed, to me, to be as a result of the parent exception.

### Context

I'd been meaning to finish up this last task from the ruff switchover, it was pretty much just a matter of pasting the same few characters in 200 places 🤷 

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,9 +35,7 @@ exclude = [
 line-length = 88
 
 [tool.ruff.lint]
-ignore = [
-  "B904",  # Within an `except` clause, raise exceptions with `raise ... from err|None`
-]
+ignore = []
 select = [
     "B",  # flake8-bugbear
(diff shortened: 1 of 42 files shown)


KevinMind (commented):
Will ruff auto fix these in the future?

KevinMind (approved):
Looks good to me. CI should cover testing for this as far as it's possible. I assume any unexpected behaviour would be covered in "stage" testing before the next release.

eviljeff:
> Will ruff auto fix these in the future?

No, it's not one of the fixable rules (I had to manually fix it).  I presume because there isn't a programmatic way to know if it should be `from None` or `from xxx`

PR #22645, opened 2024-09-05
Don't send auto-approval email for langpacks

eviljeff:
Fixes: mozilla/addons#1968

### Description

Don't trigger the task which would email the developers of langpacks for auto-approvals

### Context

All langpacks are from Mozilla releng these days and they don't need the emails.  The logic is in the reviewer tools because further along, for example in CinderDecision.notify_reviewer_decision, nothing else other than emailing the developers of the add-on happens anyway - for approvals without attached jobs: we don't create a CinderDecision instance; we don't notify Cinder; there are no cinder jobs that could be resolved; and there can't be any reporters to be emailed.

### Testing

- Submit a langpack (this will probably be easier if it's a new version of an existing langpack, so you don't hit any other restrictions/delays on new submissions)
- manually run `manage.py cron autoapprove`
- see no approval email in django admin
- (repeat with some other type of add-on that would be auto approved and see the email)

### Checklist

### DIFF
--- a/src/olympia/reviewers/tests/test_utils.py
+++ b/src/olympia/reviewers/tests/test_utils.py
@@ -1469,19 +1469,13 @@ class TestReviewHelper(TestReviewHelperBase):
         )
         assert activity.details['human_review'] is True
 
-    def test_nomination_to_public_not_human(self):
+    def _test_nomination_to_public_not_human(self):
         self.sign_file_mock.reset()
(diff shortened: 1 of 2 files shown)


diox:
> manually run manage.py cron autoapprove

`manage.py auto_approve`

diox (approved):

PR #22646, opened 2024-09-06
Update admin.py

KevinMind:
Fixes: mozilla/addons#15007

### Description

When updating the static paths for npm module static assets, I forgot to update this asset in admin, causing 500 for missing file.

### Context

This didn't trigger a build failure because we do not compress this asset, but just collect it. I'm also somewhat surprised collectstatic would allow a nonexistent file to quietly pass through.

### Testing

Make up and navigate to any admin page. should load

### Checklist

### DIFF
--- a/src/olympia/amo/admin.py
+++ b/src/olympia/amo/admin.py
@@ -183,7 +183,7 @@ class AMOModelAdmin(admin.ModelAdmin):
         js = (
             'js/admin/ip_address_search.js',
             'js/exports.js',
-            'js/node_lib/netmask.js',
+            'netmask/lib/netmask.js',
         )
         css = {'all': ('css/admin/amoadmin.css',)}
 


diox:
`collectstatic` doesn't look at python code referencing assets - it just goes through the staticfiles finders and collects everything it finds.

diox:
Which reminds me btw... now that we are using django-node-assets, our `site-static/` post-build is much bigger, because it will contain a copy of every node module js we have...

eviljeff:
I'm going to delegate this r? to @diox because he seems to have knowledge of the patch that caused the regression

diox (approved):
(Can be verified by running in `DEBUG=False` or with `findstatic` command)

KevinMind:
> Which reminds me btw... now that we are using django-node-assets, our site-static/ post-build is much bigger, because it will contain a copy of every node module js we have...

This is true. I don't think it's a fundamental problem though as there are multiple ways to help reduce that in the future.

diox:
Looks like `django-node-assets` does have a different finder class called `ManifestNodeModulesFinder` if we want to go that route: https://github.com/whitespy/django-node-assets/blob/master/django_node_assets/finders.py#L92

We would then explicitly add to `package.json` the npm packages we care about and it would only collect those.

PR #22661, opened 2024-09-11
Allow reviewers and developers to reply to older versions

diox:
Fixes: mozilla/addons#1981

### DIFF
--- a/src/olympia/activity/tests/test_views.py
+++ b/src/olympia/activity/tests/test_views.py
@@ -150,15 +150,22 @@ class ReviewNotesViewSetDetailMixin(LogMixin):
         self.version.delete()
 
         # There was a listed version, it has been deleted but still, it was
-        # there, so listed reviewers should still be able to access.
+        # there, so listed reviewers should still be able to access if they
+        # have Addons:ViewDeleted
         self._login_reviewer()
+        response = self.client.get(self.url)
+        assert response.status_code == 404
+
+        user = UserProfile.objects.get(username='reviewer')
+        self.grant_permission(user, 'Addons:ViewDeleted')
         response = self.client.get(self.url)
         assert response.status_code == 200
 
     def test_deleted_version_developer(self):
         self.version.delete()
         self._login_developer()
-        self._test_url()
+        response = self.client.get(self.url)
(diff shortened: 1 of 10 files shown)


src/olympia/activity/tests/test_views.py
```diff
@@ -143,15 +143,22 @@ def test_deleted_version_regular_reviewer(self):
...
         # There was a listed version, it has been deleted but still, it was
-        # there, so listed reviewers should still be able to access.
+        # there, so listed reviewers should still be able to access if they
+        # have Addons:ViewDeleted
         self._login_reviewer()
         response = self.client.get(self.url)
+        assert response.status_code == 404
+
+        user = UserProfile.objects.get(username='reviewer')
+        self.grant_permission(user, 'Addons:ViewDeleted')
```
diox, line 153: Removing the check on version being the latest version forced me to fix the `test_reply_to_deleted_version_is_400` test below, which was expecting a 400... but it really should have expected a 404 instead. Digging further, I found that developers and reviewers without `Addons:ViewDeleted` were able to view activities on deleted versions, which was not consistent with every other API/view. So I fixed that, which in turn forced me to adjust a few more tests.

src/olympia/devhub/templates/devhub/versions/list.html
```diff
@@ -89,9 +89,7 @@
          data-is-current="{{ (version == addon.current_version)|int }}">x</a>
     </td>
   </tr>
-  {% set can_request_review=addon.can_request_review() %}
-  {% set can_cancel=(not addon.is_disabled and addon.status==amo.STATUS_NOMINATED) %}
-  {% if full_info and check_addon_ownership(request.user, addon, allow_developer=True) and version.channel == amo.CHANNEL_LISTED and (can_request_review or can_cancel) %}
+  {% if full_info and version.channel == amo.CHANNEL_LISTED and (can_request_review or can_cancel) and check_addon_ownership(request.user, addon, allow_developer=True) %}
```
diox, line 92: We were redefining those 2 variables for each version, which is pointless and caused extra queries. I've moved their definition to the view.

eviljeff (approved):

PR #22662, opened 2024-09-12
Remove all reference to update_deps*

KevinMind:
Follow up: https://github.com/mozilla/addons-server/pull/22644

### Description

It is no longer possible to update dependencies in a running container. You must instead `make up` to get new dependencies.

This PR removes all the make references to update_deps and moves the logic to the docckerfile directly since it can only be triggered during a build.

Also cleaned up some lints for dockerfile

### Testing

If you run `make initialize_docker` it should work. 

> NOTE: this will wipe your current DB so make sure to run the `make data_export` first.

### Checklist


Makefile-docker
```diff
@@ -135,7 +97,7 @@ update_assets:
 	$(PYTHON_COMMAND) manage.py collectstatic --noinput
 
 .PHONY: update
-update: update_deps update_db update_assets ## update the dependencies, the database, and assets
+update: update_db update_assets ## update the dependencies, the database, and assets
```
eviljeff, line 100: Something I missed in the changes in https://github.com/mozilla/addons-server/commit/bc186195bc5da7fc0f7268a11f67a5470a64c454 was that `update_docker` (in Makefile-os) no longer calls this `update` recipe, so `update_db` doesn't get called anywhere automatically. Maybe `update_assets` should be executed with `up` and this recipe just removed, leaving `make update_db` to be stand-alone command.

I think calling `update_deps` too as part of `up` is risky though - if there are any db conflicts (which can happen often if you've been making changes in one of your patches) there wouldn't be an easy way to recover.

eviljeff (approved):

scripts/extract-l10n.sh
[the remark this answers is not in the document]
KevinMind: Yes. It also referenced update_deps so tangentially relevant.

Makefile-docker
KevinMind: It is impossible to call `update_deps` at all on a running container. The thing is.. `update_assets` and `update_deps` are both already called during `up` because, well you build an image and in it are the deps and the compiled/collected assets.

I have a separate PR (this one is split from) to run db updates as a part of up. It was getting to big and this part seemed relatively straight forward.

in the end I don't think there will need to be any commands starting with "update" as they all just get rolled into "up"

eviljeff: gah, I meant to write:

> I think calling **update_db** too as part of `up` is risky though

Migrations can easily fail if you've been making database changes yourself and `make up` can't fail or we can't get into the container to investigate/fix the problem

PR #22664, opened 2024-09-13
DevHub Pastebin Re-implementation

chrstinalin:
Fixes: mozilla/addons#14998

### Description

Surfaces attachments on the DevHub side, allowing developers to download reply attachments.
image

### Testing
1. Enable the `enable-activity-log-attachments` waffle switch.
2. Log in with local_admin@mozilla.com (for ease of navigation).
3. Submit a new add-on.
4. Make a reviewer reply (or any other action) from the review action section and include an attachment.
5. Navigate to the 'Manage Status & Versions'. Click 'Review History' under 'Status' of the replied-to version.

### Checklist


src/olympia/activity/tests/test_views.py
```diff
@@ -295,7 +295,7 @@ def test_queries(self):
             'fiiiine', amo.LOG.REVIEWER_REPLY_VERSION, self.days_ago(0)
         )
         self._login_developer()
-        with self.assertNumQueries(17):
+        with self.assertNumQueries(21):
```
diox, line 298: That's a problem we need to fix

static/css/zamboni/developers.css
```diff
@@ -702,6 +702,8 @@ div.history-container {
 div.review-entry {
     padding: 0.5em 1em;
     margin-bottom: 0.25em;
+    display: flex;
```
KevinMind, line 705: is default alignment working? Feel like it rarely does..

static/js/zamboni/devhub.js
```diff
@@ -602,6 +602,12 @@ function initVersions() {
       date.attr('datetime', note['date']);
       date.attr('title', note['date']);
       clone.find('pre:contains("$comments")')[0].textContent = note['comments'];
+      if (note['attachment_url']) {
```
KevinMind, line 605: Nit: why not insert the node only if the note has an attachment_url. You already do a bunch of dom modification why not just one insert only if needed.

Another alternative would be to se the href and text content in the template (to empty values if there is no attachment) and then let css decide if the element should be hidden based on the child content of the div?

I almost like that option better as it wouldn't require any javascript at all.

src/olympia/devhub/templates/devhub/versions/list.html
```diff
@@ -118,9 +118,15 @@
...
               <div class="review-entry-empty hidden">
-                  <p><strong class="action">$action_label</strong> {{ _('by') }}
-                  <em class="user_name">$user_name</em> <time class="timeago" datetime="$date">$date</time></p>
-                  <pre>$comments</pre>
+                  <div>
+                    <p><strong class="action">$action_label</strong> {{ _('by') }}
+                    <em class="user_name">$user_name</em> <time class="timeago" datetime="$date">$date</time></p>
+                    <pre>$comments</pre>
+                  </div>
+                  <div class="review-entry-attachment hidden">
```
KevinMind, line 126: Doubling back on my comment in the javascript file. You should be able to wrap the attachment div in a jinja condition to only render if there is an attachment. Then you can pass all the properties expecting them to exist, I think. Unless there is some reason why the attachment would  only be available client side (which I doubt)

static/css/zamboni/developers.css
chrstinalin: I mean, the css is working as expected for me?

With:
image

Without:
image

static/js/zamboni/devhub.js
chrstinalin: > Another alternative would be to se the href and text content in the template (to empty values if there is no attachment) and then let css decide if the element should be hidden based on the child content of the div?

I can see doing that, but I don't know if it would remove the need for the js (see the comment on .html), so I feel it would become more confusing if I were to split the logic between both a js and css file?

src/olympia/devhub/templates/devhub/versions/list.html
KevinMind: > Unless there is some reason why the attachment would only be available client side

This statement seems to be true. Yeah I get it now.

KevinMind:
Trying to verify I caught something.. how is the URL being defined? I see the get_attachment_url using revers() but also saw in the browser a relative URL.

@diox can you clarify I can't remember if we need special care for ensuring the site URL is included as the URL root.

image

Also clicking the URL in local dev resulted in infinite redirects and eventually  a failure.

Screenshot 2024-09-16 at 17 37 29

diox:
On local environments, where requests are proxied through addons-frontend, it should work fine if your `addons-frontend` container is up to date with the changes from https://github.com/mozilla/addons-frontend/commit/c09adf09fc38af8771714b8092f6614b92a9f5fa

The reason it works if you remove the constants is that then it makes a redirect to `/{locale}/activity/...`, but we want to avoid that.

On dev/stage/prod we'll need SRE to update nginx config in https://github.com/mozilla-it/webservices-infra/blob/main/amo/k8s/amo-proxy/configs/addons.conf.tpl to make `/activity` be served by addons-server.

KevinMind:
I just discovered we are currently not defaulting to pulling latest image of dependencies like `addons-frontend` that should probably be the case. I totally missed that my frontend was 7 days stale.

KevinMind:
@chrstinalin okay good news, new error. But still an error. now I'm getting `RelatedObjectDoesNotExist` looking for the attachment log.

I did the following:
- navigate to listed review page
- rejected the version with a "copy from pastebin" text I just entered random words
- nvagiate to status and versions page
- click view history
- I see the attachment link and click
- I see this error page.

Any idea what could be happening?

image

chrstinalin:
@KevinMind Whoops. Reverse is using the wrong ID

diox:
`/downloads` is not a bad idea, although currently, `/downloads` is defined from `olympia.versions.urls`, and this is not really tied to a version at all... and we already had `activity` views/URLs...

KevinMind (approved):
Verified working locally. LGTM

PR #22665, opened 2024-09-16
Use Creative Commons v4 licenses instead of v3 for new AMO themes

diox:
Fixes: mozilla/addons#1203

## Context

Built-in licenses exist in the database with a `builtin` attribute value matching the corresponding constant - then, the `on_form` attribute on the constant dictates whether or not to show them to the developer.

We're adding a bunch of constants + database entries through the migration, and making the old ones `on_form = False`, causing them to no longer be shown, unless the developer already had one of them selected.

## Testing

You can test that the new license choices appear at submission time for themes, or are shown when editing the license of the theme in devhub.


eviljeff (approved):

diox:
@leoheitmannruiz yes, they should. If you're up for it please send a PR set me as the reviewer.

PR #22670, opened 2024-09-16
Allow Zip Files to be Uploaded as Attachment

chrstinalin:
Fixes: mozilla/addons#14999

### Description

Builds on #22640 to allow .zip files to be uploaded as attachment.

image

**Note**: The local nginx configuration fails for large files. There's an issue filed for this [here](https://github.com/mozilla/addons/issues/15029). To test this feature locally, two lines in `addons.conf` need to be changed --
1. `try_files $uri @frontendamo;` -> `try_files $uri @olympia;`
2. `client_max_body_size 50m;`-> `client_max_body_size 500m;` (or anything else bigger than 200mb)

### Testing

1. Enable the `enable-activity-log-attachments` waffle switch.
3. Log in with local_admin@mozilla.com (for ease of navigation).
4. Submit a new add-on.
5. Make a reviewer reply (or any other action) from the review action section and include a zip attachment.

### Checklist

### DIFF
--- a/src/olympia/reviewers/forms.py
+++ b/src/olympia/reviewers/forms.py
@@ -316,9 +316,7 @@ def validate_review_attachment(value):
                 )
             )
         if value.size >= settings.MAX_UPLOAD_SIZE:
-            raise forms.ValidationError(
-                gettext('File too large.')
-            )
+            raise forms.ValidationError(gettext('File too large.'))
         try:
             if value.name.endswith('.zip'):
                 # See clean_source() in WithSourceMixin
@@ -395,8 +393,11 @@ class ReviewForm(forms.Form):
         widget=ReasonsChoiceWidget,
     )
     attachment_file = forms.FileField(
-        required=False, validators=[validate_review_attachment],
-        widget=forms.ClearableFileInput(attrs={'max_upload_size': settings.MAX_UPLOAD_SIZE})
+        required=False,
+        validators=[validate_review_attachment],
+        widget=forms.ClearableFileInput(
+            attrs={'max_upload_size': settings.MAX_UPLOAD_SIZE}
+        ),
     )
     attachment_input = forms.CharField(required=False, widget=forms.Textarea())
 
@@ -406,7 +407,6 @@ class ReviewForm(forms.Form):
         required=False,
         queryset=CinderJob.objects.none(),
         widget=CinderJobsWidget(attrs={'class': 'data-toggle-hide'}),
-        
     )
 
     cinder_policies = forms.ModelMultipleChoiceField(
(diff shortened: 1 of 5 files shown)


src/olympia/reviewers/forms.py
```diff
@@ -312,6 +315,18 @@ def validate_review_attachment(value):
                     'file {extensions}.'.format(extensions=valid_extensions_string)
                 )
             )
+        if value.size >= settings.MAX_ZIP_UNCOMPRESSED_SIZE:
```
diox, line 318: This setting has a specific meaning that is subtly different from maximum file size, as it's the maximum size of the zip files _contents_ once unzipped. We might change that value in the future, so let's use a new setting instead.

Call that new setting `MAX_UPLOAD_SIZE`, stick that in `lib/settings_base.py` with a comment explaining that this is our own setting and not Django's and it needs to be implemented by relevant forms. This would make https://github.com/mozilla/addons/issues/14835 a nice follow-up for you to work on after if you have the bandwidth.

Note that dev/stage/prod are currently configured with `client_max_body_size 200m` so the python code won't even see requests trying to upload files over 200 MB anyway, but we could bump it to 250 MB to give a little bit of leeway (we won't accept requests too big to avoid DOS attacks and similar with large requests).

diox:
Yes, that would be ideal. Doesn't have to be as fancy, but that would be a nice little quality of life feature.

static/js/zamboni/reviewers.js
```diff
@@ -242,6 +242,27 @@ function callReviewersAPI(apiUrl, method, data, successCallback) {
...
+  const file = this.files[0];
+  const max_upload_size = $(this).data('max-upload-size');
+  if (file) {
+    if (file.size > max_upload_size) {
+      error = $('<ul>')
+        .attr('class', 'errorlist')
+        .append(
+          $('<li>').append(
+            format(gettext('Your file exceeds the maximum size of {0}.'), [
+              '200MB',
```
diox, line 256: We don't have to hardcode this:
```
Intl.NumberFormat(document.documentElement.lang, {
  notation: 'compact',
  style: 'unit',
  unit: 'byte',
  unitDisplay: 'narrow',
}).format(max_upload_size);
```

(It gives 210MB because it's using SI units but that's fine)

diox (approved):

static/js/zamboni/reviewers.js
chrstinalin: Wouldn't it be confusing for a 208MB file to reject with this error?

diox: It already works actually - I think our python code is using SI units, so your patch already allows 210MB (200 MiB) files server-side.

chrstinalin:
> I did not try all the possible reviewer actions to check the emails, but that would be https://github.com/mozilla/addons/issues/15018. So far, I did notice that reviewer's reply email has this but I did not see a reference to it for approve or reject (will re-check anyway). (https://github.com/mozilla/addons/issues/14997#issuecomment-2374190777)

Template for approvals/rejections should also mention any attachments.

PR #22675, opened 2024-09-17
Fix CinderPolicy plural name

willdurand:
Fixes https://github.com/mozilla/addons/issues/15011


eviljeff (approved):
This is a bit of a weird bug - we _do_ define `verbose_name_plural` in `CinderPolicyAdmin` but it seems that the property isn't used for vanilla ModelAdmin - only for inlines.

It'd be nice to delete the attempted definition in `CinderPolicyAdmin` too, so it's not confusing.

PR #22678, opened 2024-09-18
Remove docker context switcher

KevinMind:
Fixes: mozilla/addons#15027

### Description

We do not need to explicitly set context. It shoud be set for most active developers and new devs will not need to switch as default should be... default.

### Context

We used to have a custom builder for docker images. We now use the default daemon owned context "default"

The switched caused this bug where docker sometimes did not like how we speecify the default context... we don't really need to do this anymore and can just remove the code and add a gotcha if anyone sees this again.

### Testing

```bash
make up DOCKER_VERSION=local
```

Should work.

### Checklist

### DIFF
--- a/Makefile-os
+++ b/Makefile-os
@@ -4,7 +4,6 @@
 # Variables referenced in docker-compose*.yml should be read from .env, exported and saved in .env
 ####################################################################################################
 
-DOCKER_BUILDER ?= default
 DOCKER_PROGRESS ?= auto
 DOCKER_METADATA_FILE ?= buildx-bake-metadata.json
 DOCKER_PUSH ?=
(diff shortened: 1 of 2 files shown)


diox:
This looks fine, but I don't understand, what was wrong with how we reset to the default context that caused Andreas's bug ?

KevinMind:
> This looks fine, but I don't understand, what was wrong with how we reset to the default context that caused Andreas's bug ?

we are calling `docker buildx use default` and somehow docker is complaining about this and suggesting to use `docker context use default`.

In essence they should be the same but since we are 1) using the default context already and 2) don't have any intentions of switching it.. I didn't see the point in spending the effort to figure out exactly why docker doesn't like that on some hosts.. 🤷 This is kind of a non issue at this point so I opted to remove the culprit entirely.

diox (approved):

PR #22679, opened 2024-09-18
Small fixes to docker setup (logs, static file routing, healthcheck, dockerignore)

KevinMind:
Fixes: mozilla/addons#15026
Fixes: mozilla/addons#15025

### Description

Semi related small commits fixing several known bugs in our local dev setup

### Context

I combined these for efficiency as it would be many tiny PRs that should be tested more or less in the same way.

### Testing

- [Only log docker compose if specified in CI run-docker step](https://github.com/mozilla/addons-server/pull/22679/commits/9744863fbbf9139ece3b50a12cdd07a9e209fb7b) check the CI for this PR in the failing steps there should be no long dumb of docker compose logs

- [Better and alphyabetized ignore paths for git and docker](https://github.com/mozilla/addons-server/pull/22679/commits/371837736a374c953860b909dbc69232429e461c) verify that files in the .dockerignore list are not included in the built image. Note: They will be included in the running container as we mount the entire root directory so test via running prod mode

```bash
make up COMPOSE_FILE=docker-compose.yml:docker-compose.ci.yml
```

- [Less aggressive post start health checks](https://github.com/mozilla/addons-server/pull/22679/commits/87239f046699da4243282a6ed471acdab91fafe4) Test this by running make up. The wait time for the containers to go from "starting" to "healthy" should still be fairly low. But after the container is "started" if you inspect the logs, there should only be a healthcheck request log every 90s.
- [Clean up redundant code](https://github.com/mozilla/addons-server/pull/22679/commits/3150e093bff0ad2d41414ebaf26dcf878f03eadd)

1. running `make up` should never build.. hard to test but it won't
2. anonymous volumes are recreated if running `make up` when containers are already up.
3. `make data_restore` works and doesn't fail with invalid index command
4. `make clean_docker` works and doesn't perform unnecessary logic

### Checklist


eviljeff:
Are these all the issues that are being fixed, and can you expand on the description?  There are 5 commits, 4 sets of the testing instructions, but only 2 linked issues. (None of which seem to be about cleaning logs - can you update the PR title too)

docker-compose.yml
```diff
@@ -66,15 +66,23 @@ services:
...
-      interval: 1s
-      retries: 100
+      # The interval is 90s after the start period of 60s
+      interval: 90s
+      # 3 failed attempts result in container failure
+      retries: 3
+      # While starting, ping faster to get the container healthy as soon as possible
+      start_interval: 1s
+      # The start period is 60s
+      start_period: 60s
```
eviljeff, line 76: Am I understanding this correctly that if start up takes longer than 60s the next healthcheck is at 2.5 minutes after start up?  What's the downside to having start_period be a longer time?

.dockerignore
```diff
@@ -1,15 +1,71 @@
...
+static/js/node_lib/*
+storage/files/*
+storage/git-storage/*
+storage/guarded-addons/*
+storage/mlbf/*
+storage/shared_storage/*
+storage/sitemaps/*
+tmp/*
+
+# Additionally ignore these files from the docker build that are not in .gitignore
```
eviljeff, line 59: Are these frequently updated?  (i.e. is there a real benefit to having these two lists of things to ignore, vs. just relying on .gitignore)

.github/workflows/ci.yml
```diff
@@ -145,6 +145,19 @@ jobs:
...
             fi
 
+      - name: Check ignored files
+        uses: ./.github/actions/run-docker
+        with:
+          digest: ${{ needs.build.outputs.digest }}
+          version: ${{ needs.build.outputs.version }}
+          run: |
+            # Verify that .dockerignore is working and the
+            # Makefile-os is not in the production container
```
eviljeff, line 155: what's the downside with `Makefile-os` being in the production container?  (or was that just a useful known file)?

docker/nginx/addons.conf
```diff
@@ -19,6 +19,23 @@ server {
...
+        alias /srv/user-media/shared_storage/uploads/;
+    }
+
+    location /user-media/addons/ {
+        alias /srv/user-media/files/;
+    }
+
+    location /user-media/guarded-addons/ {
+        alias /srv/user-media/guarded-addons/;
+    }
```
eviljeff, line 33: see https://github.com/mozilla/addons/issues/15025#issuecomment-2358690244

docker-compose.yml
KevinMind: The startup period defines how long to wait before "counting" towards the total retries. If the service starts before that time it will switch to normal healtcheck. 

So if startup takes longer than 60s the next healtcheck is indeed 2.5 after and will try 3 times. BUT. It is unfathomable that startup takes longer than 60s and you don't have a terminally ill container.

It will always be less than that which is why I set it there. Making it longer won't materially change anything.

Additionally you _want_ the normal interval to be long otherwise you get spammed with /__version__ requests in the container logs.

.dockerignore
KevinMind: When you say "these" you mean the files or the ignore file?

Dockerignore should include everything in gitignore because CI by definition only has what you committed.

Dockerignore should include other things not in gitignore because we don't necessarily want those files in the image. All of these fit into that category.

.github/workflows/ci.yml
KevinMind: It is useless in that container. Everything it does is designed to work on a host outside a container. Excluding it prevents you from accidentally running host commands inside a production container.

docker/nginx/addons.conf
KevinMind: I think it will be easier to remove this part and move that to a separate PR.

docker-compose.yml
eviljeff: We might have to revisit that "always less than 60s" after Andreas and I use it 😿

KevinMind: Would 120s be more realistic? give me a number and I'll bump it up.

.dockerignore
eviljeff: I mean maintaining these two files with different lists in them - and specifically the differences between them, like these additional files.  Is there any benefit beyond not necessarily wanting them in the image?

docker-compose.yml
eviljeff: I don't know yet, but we've consistently seen longer times.

.dockerignore
KevinMind: `.dockerignore` is a super set of `gitignore`. I thought about using a script to validate or auto generate the docker ignore but I think it's overkill.

The benefit is precisely excluding these files from the build. Excluding them from the build means fewer cache misses and faster builds and smaller images. I don't think we need an additional benefit as that is plenty.

KevinMind: I added a dockerignore section to the docs. lmk wyt if this addresses any doubts or confusion.

eviljeff (approved):
I didn't see any affect from changes to `start_period` - it took ~90 seconds to get to Healthy so I changed the start_period in docker-compose to 180 and it still took ~90 seconds 🤷

PR #22681, opened 2024-09-18
Show "pending" activity in devhub for all versions and not just the latest one

diox:
Fixes: mozilla/addons#15024

## Context

See the issue - without this change, for a given add-on in devhub, only the last version in each channel shows the "pending" activity count. This makes it shown on all versions, while not increasing the query count (plenty more to optimize there though, but out of scope)

## Testing

See the issue but also the added tests of the model method + test around the view that were added.

### DIFF
--- a/src/olympia/activity/models.py
+++ b/src/olympia/activity/models.py
@@ -325,21 +325,26 @@ class ActivityLogQuerySet(BaseQuerySet):
     def default_transformer(self, logs):
         ActivityLog.arguments_builder(logs)
 
-    def pending_for_developer(self):
+    def pending_for_developer(self, for_version=None):
+        if for_version is None:
+            for_version = models.OuterRef('versionlog__version_id')
         latest_reply_date = models.functions.Coalesce(
             models.Subquery(
                 self.filter(
                     action__in=NOT_PENDING_IDS,
-                    versionlog__version_id=models.OuterRef('versionlog__version_id'),
+                    versionlog__version=for_version,
                 )
                 .values('created')
                 .order_by('-created')[:1]
             ),
             date.min,
         )
-        return self.filter(
-            action__in=amo.LOG_REVIEW_QUEUE_DEVELOPER,
-            created__gt=latest_reply_date,
+        return (
+            # The subquery needs to appear after having already filtered by
+            # action.
+            self.filter(action__in=amo.LOG_REVIEW_QUEUE_DEVELOPER)
+            .annotate(latest_reply_date=latest_reply_date)
+            .filter(created__gt=models.F('latest_reply_date'))
         )
 
 
--- a/src/olympia/activity/models.py
+++ b/src/olympia/activity/models.py
@@ -3,7 +3,7 @@ import os
 import uuid
 from collections import defaultdict
 from copy import copy
-from datetime import datetime
+from datetime import date, datetime
 from inspect import isclass
 
 from django.apps import apps
@@ -40,6 +40,18 @@ log = olympia.core.logger.getLogger('z.amo.activity')
 MAX_TOKEN_USE_COUNT = 100
 
 GENERIC_USER_NAME = gettext('Add-ons Review Team')
+
+# Activity ids that are not considered as needing a reply from developers, so
+# they are never considered "pending".
+NOT_PENDING_IDS = (
+    amo.LOG.DEVELOPER_REPLY_VERSION.id,
+    amo.LOG.APPROVE_VERSION.id,
+    amo.LOG.REJECT_VERSION.id,
+    amo.LOG.PRELIMINARY_VERSION.id,
+    amo.LOG.PRELIMINARY_ADDON_MIGRATED.id,
+    amo.LOG.NOTES_FOR_REVIEWERS_CHANGED.id,
+    amo.LOG.SOURCE_CODE_UPLOADED.id,
+)
 
 
 def attachment_upload_path(instance, filename):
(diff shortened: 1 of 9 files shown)


src/olympia/activity/tests/test_models.py
```diff
@@ -99,6 +99,89 @@ def test_rejected_version_still_valid(self):
         assert self.token.is_valid()
 
 
+class TestActivityLogManager(TestCase):
+    def test_pending_for_developer(self):
```
diox, line 103: FIXME: need more tests, particularly when using the same activitylog for multiple replies across different versions, but also the special `for_version` arg

src/olympia/activity/models.py
```diff
@@ -313,6 +325,28 @@ class ActivityLogQuerySet(BaseQuerySet):
     def default_transformer(self, logs):
         ActivityLog.arguments_builder(logs)
 
+    def pending_for_developer(self, for_version=None):
```
diox, line 328: FIXME: needs a docstring

src/olympia/devhub/views.py
```diff
@@ -1267,7 +1267,19 @@ def check_validation_override(request, form, addon, version):
 
 @dev_required
 def version_list(request, addon_id, addon):
-    qs = addon.versions.order_by('-created')
+    unread_count = (
+        (
+            ActivityLog.objects.all()
+            .pending_for_developer(for_version=OuterRef(OuterRef('id')))
+            .filter(versionlog__version=OuterRef('id'))
```
diox, line 1274: Needs a comment explaining that this `filter()` cannot appear before `pending_for_developer()` as that method is doing a subquery, so `OuterRef` would be wrong.

src/olympia/activity/models.py
```diff
@@ -41,6 +41,18 @@
 
 GENERIC_USER_NAME = gettext('Add-ons Review Team')
 
+# Activity ids that are not considered as needing a reply from developers, so
+# they are never considered "pending".
+NOT_PENDING_IDS = (
```
KevinMind, line 46: I see it predates this PR but curious why a negative definition of NOT_PENDING instead of PENDING_IDS?

src/olympia/activity/tests/test_models.py
```diff
@@ -99,6 +99,129 @@ def test_rejected_version_still_valid(self):
...
+        to_create = (
```
KevinMind, line 104: curious why not using a parameterized test here? Again seems to predate but this would parallelize the tests and let each fail independently of the others.

src/olympia/devhub/tests/test_views_versions.py
```diff
@@ -658,26 +657,81 @@ def test_version_history_mixed_channels(self):
...
+        )
         ActivityLog.objects.create(
             amo.LOG.REVIEWER_REPLY_VERSION, v2.addon, v2, user=self.user
         )
         ActivityLog.objects.create(
             amo.LOG.REVIEWER_REPLY_VERSION, v2.addon, v2, user=self.user
         )
 
-        response = self.client.get(self.url)
+        with self.assertNumQueries(43):
```
KevinMind, line 675: This test is to prevent additional queries being added to the view? Instead of documenting here could you put this in a ticket and link the ticket in the code?

src/olympia/activity/models.py
diox: @eviljeff might remember but I don't know, I haven't touched that code before. I think I would have preferred a property on the activity constant like we do elsewhere but I didn't want to expand the scope too much.

src/olympia/activity/tests/test_models.py
diox: `@pytest.mark.parametrize` is not really usable with `TestCase` - it wants to feed all the arguments to the function, so that doesn't play well with methods that should receive `self` as the first argument.

src/olympia/devhub/templatetags/jinja_helpers.py
```diff
@@ -88,10 +87,12 @@ def summarize_validation(validation):
 
 @library.global_function
 def pending_activity_log_count_for_developer(version):
-    alog = ActivityLog.objects.for_versions(version).filter(
-        action__in=amo.LOG_REVIEW_QUEUE_DEVELOPER
-    )
-    return filter_queryset_to_pending_replies(alog).count()
+    # unread_count is an annotation set by version_list() view to do this once
```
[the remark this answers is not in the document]
diox: Sadly, at the moment the various versions this page cares about outside of the `versions` list are defined in the template, not the view, so this would have been a refactor too far IMHO. The template tag was already there so this was simpler to keep.

In some places we have that logic encapsulated in models, with a custom transformer that sets a "private" property for a whole queryset and then some method that either returns the hidden property or re-computes the value if the transformer was not used. That would be my preference for code that we definitely need to re-use, but this case felt fairly specific...

src/olympia/devhub/tests/test_views_versions.py
diox: This is a test to prove my own changes don't scale badly and make the page performance any worse (note that there aren't extra queries caused by the number of versions or number of activities on those versions). I'll file an issue about the various bits we could improve on that I discovered along the way (though it's very low priority/unlikely to be worked on any time soon, it has been that way forever)

src/olympia/activity/models.py
eviljeff: I don't remember either.  I think if I was writing this today I'd start with the activities that could need a reply instead (from properties in the constant).

src/olympia/devhub/tests/test_views_versions.py
KevinMind: Still better to track in an issue.

src/olympia/devhub/templatetags/jinja_helpers.py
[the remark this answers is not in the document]
KevinMind: Makes sense

src/olympia/activity/tests/test_models.py
KevinMind: Right. In those cases I've just used functional tests. Just wondering as I've used it before and wonder if we have a stigma against it or not.

src/olympia/devhub/tests/test_views_versions.py
diox: https://github.com/mozilla/addons/issues/15037

src/olympia/activity/tests/test_models.py
diox: I don't love `@parametrize` because of that limitation (and general awkwardness around how arguments are defined when using it) but nobody would stop you if you were to use it in new code. We do have several already.

KevinMind (approved):
Verified working. 🚢 

image

PR #22683, opened 2024-09-19
clear REASONS.ABUSE_ADDON_VIOLATION if a forwarded job is combined

eviljeff:
Fixes: mozilla/addons#15001

### Description

Clears NHR for an abuse report if a forwarded job is combined into it.

### Context

Cinder combines reports for the same queue together into single jobs.  So if a job existed in the content infringement queue for abuse report(s) then that would have a `ABUSE_ADDON_VIOLATION` NHR; if a job is forwarded and we recreate it in that infringement queue then it'll be combined into a single job (this is all fine).  But now we have a job that, effectively, has two NHR for it - one of the abuse report, one for the forward.  

I tried adjusting `CinderJob.clear_needs_human_review_flags` to clear both when the job was resolved but it turned out to be difficult, so this patch clears the unnecessary NHR just after the job has been recreated.

Ultimately this may all be code that is removed once AMO forwards are reimplemented using Cinder's new multi-queue escalations.

### Testing

Follow the STR in the issue - it's tricky to do though because it needs cinder api integration set up, and for the webhook responses to be applied locally.

### Checklist

### DIFF
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -235,13 +235,35 @@ class CinderJob(ModelBase):
         )
 
     def handle_job_recreated(self, new_job_id):
-        new_job, _ = CinderJob.objects.update_or_create(
+        from olympia.reviewers.models import NeedsHumanReview
+
+        new_job, created = CinderJob.objects.update_or_create(
             job_id=new_job_id,
             defaults={
                 'resolvable_in_reviewer_tools': True,
                 'target_addon': self.target_addon,
             },
         )
+        # If this forward has been combined with an existing non-forwarded job we need
+        # to clear the duplicate NHR
+        if not created and not new_job.forwarded_from_jobs.exists():
+            has_unresolved_abuse_report_jobs = (
+                self.__class__.objects.for_addon(new_job.target_addon)
+                .exclude(
+                    id=new_job.id,
+                    forwarded_from_jobs__isnull=True,
+                    appealed_decisions__isnull=True,
+                )
+                .unresolved()
+                .resolvable_in_reviewer_tools()
+            )
+            if not has_unresolved_abuse_report_jobs:
+                NeedsHumanReview.objects.filter(
+                    version__addon_id=new_job.target_addon.id,
+                    is_active=True,
+                    reason=NeedsHumanReview.REASONS.ABUSE_ADDON_VIOLATION,
+                ).update(is_active=False)
+                new_job.target_addon.update_all_due_dates()
         # Update our fks to connected objects
         AbuseReport.objects.filter(cinder_job=self).update(cinder_job=new_job)
         CinderDecision.objects.filter(appeal_job=self).update(appeal_job=new_job)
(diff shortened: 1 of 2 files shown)


src/olympia/abuse/models.py
```diff
@@ -235,13 +235,35 @@ def notify_reporters(self, action_helper):
...
+        if not created and not new_job.forwarded_from_jobs.exists():
+            has_unresolved_abuse_report_jobs = (
+                self.__class__.objects.for_addon(new_job.target_addon)
+                .exclude(
+                    id=new_job.id,
+                    forwarded_from_jobs__isnull=True,
+                    appealed_decisions__isnull=True,
+                )
+                .unresolved()
+                .resolvable_in_reviewer_tools()
```
diox, line 258: Add a `.exists()`

src/olympia/abuse/models.py
```diff
@@ -235,13 +235,35 @@ def notify_reporters(self, action_helper):
...
+
+        new_job, created = CinderJob.objects.update_or_create(
             job_id=new_job_id,
             defaults={
                 'resolvable_in_reviewer_tools': True,
                 'target_addon': self.target_addon,
             },
         )
+        # If this forward has been combined with an existing non-forwarded job we need
+        # to clear the duplicate NHR
```
diox, line 248: Could you clarify that comment for me and future readers of that code? I suppose you mean that the duplicate NHR has a different `reason` (`ABUSE_ADDON_VIOLATION` vs `CINDER_ESCALATION`), which is why we can safely clear the old one?

src/olympia/abuse/tests/test_models.py
```diff
@@ -1071,7 +1075,47 @@ def test_handle_job_recreated_existing_job(self):
...
+        assert old_job.forwarded_to_job == exisiting_report_job
+        assert list(exisiting_report_job.forwarded_from_jobs.all()) == [old_job]
+        assert list(exisiting_report_job.abusereport_set.all()) == [
+            existing_report,
+            report,
+        ]
+        assert report.reload().cinder_job == exisiting_report_job
+        assert not NeedsHumanReview.objects.filter(
+            is_active=True
+        ).exists()  # it's cleared
```
diox, line 1118: Related to the comment I made above: in real life, we'd have a `NeedsHumanReview` with `CINDER_ESCALATION` reason left, right ? We don't here because that's done by the webhook handler but I'd feel better if you added it manually before, and then adjusted that assertion to ensure it's still here and we only cleared the `ABUSE_ADDON_VIOLATION` one.

eviljeff: https://github.com/mozilla/addons-server/pull/22683/commits/6657358f084879812247fe4acb07b9ba6958b56a

src/olympia/abuse/models.py
eviljeff: https://github.com/mozilla/addons-server/pull/22683/commits/6657358f084879812247fe4acb07b9ba6958b56a

diox (approved):

PR #22691, opened 2024-09-20
Mention max allowable filesize on DevHub when uploading versions/sources

chrstinalin:
Fixes: mozilla/addons#14835

### Description

Mentions the max allowable filesize (200mb) on DevHub when uploading versions/sources.
image
image

### Checklist


src/olympia/devhub/templates/devhub/addons/submit/source.html
```diff
@@ -63,6 +63,11 @@ <h3>{{ _("You must submit your extension's source code.") }}</h3>
                 Source files (aside from open-source third party libraries) cannot be transpiled, concatenated, minified, or otherwise machine-generated.
             {% endtrans %}
         </p>
+        <p>
+            {% trans %}
+            The maximum file size accepted is 200MB.
```
KevinMind, line 68: Is there a setting or some constant determining what the actual max file size is? I would recommend a) linking to that in an html comment or b) (better but more work) using that value in the template so they are always synchronized.

src/olympia/devhub/templates/devhub/addons/submit/upload.html
```diff
@@ -31,6 +31,9 @@ <h3>{{ _('Upload Version') }}</h3>
     Use the fields below to upload your add-on package. After upload, a series
     of automated validation tests will be run on your file.
   {% endtrans %}
+  {% trans %}
+    The maximum file size accepted is 200 MB. If your add-on is larger than 200 MB, it will fail validation.
```
KevinMind, line 35: Why the extra explanation here and not in the other place?

chrstinalin: Just going off the linked issue 😅

src/olympia/devhub/templates/devhub/addons/submit/source.html
chrstinalin: The one attributed to the error shown to the user, as far as I could tell, was just a hard-coded variable in the JS (upload-addon.js#38 used in upload-base.js#83). I added a comment next to it

chrstinalin: ([This](https://github.com/mozilla/addons/issues/15039#issuecomment-2368479933) would change things, though)

src/olympia/devhub/templates/devhub/addons/submit/upload.html
```diff
@@ -31,6 +31,17 @@ <h3>{{ _('Upload Version') }}</h3>
     Use the fields below to upload your add-on package. After upload, a series
     of automated validation tests will be run on your file.
   {% endtrans %}
+  {% trans %}
```
KevinMind, line 34: @diox will know better than me how this will actually get extracted but is this the "right" way to do this? I know there is some way of injecting a "variable" in a string but I don't remember exactly how to do it.

I think this could result in several independent strings for translators that could be a bit confusing..

KevinMind: I think this might be an example. https://github.com/mozilla/addons-server/blob/8d00231f29890e4363f0b9e079823f4ff55a3429/src/olympia/amo/templates/amo/404.html#L7-L38

I guess you can pass arguments to the `{% trans foo="bar" %}` which allows you to encapsulate a single "string" with replaceable content. Then you can just reference the content.

Again @diox please correct me if I'm wrong.

diox: Correct, we don't want to break a translated string like this. It's not just about it not making sense for translators: some languages might need to change the order of the words, which can't happen if it's been broken down into separate strings.

src/olympia/devhub/templates/devhub/addons/submit/source.html
```diff
@@ -63,6 +63,12 @@ <h3>{{ _("You must submit your extension's source code.") }}</h3>
                 Source files (aside from open-source third party libraries) cannot be transpiled, concatenated, minified, or otherwise machine-generated.
             {% endtrans %}
         </p>
+        <p>
+            {% trans %}
+            The maximum file size accepted is 
+            {% endtrans %}
+            {{ max_upload_size|filesizeformat }}.
+        </p>
```
diox, line 71: See below, the variable needs to be passed to the `{% trans %}` block. You can do `{% trans max_size=max_upload_size|filesizeformat %}The maximum file size accepted is {{ max_size }}. {% endtrans %}`.

src/olympia/devhub/templates/devhub/addons/submit/upload.html
diox: As a side note that 404 page is a bit special because the entire page contents is in the `{% trans %}` block, which is fairly rare. That pretty much only happens for error pages. A simpler example:
https://github.com/mozilla/addons-server/blob/b8ffdae3c8a459ef61525aefa21c4debbb45896b/src/olympia/devhub/templates/devhub/suppressed_email.html#L4-L6

Another which is old and kinda bad, but shows that it works even with a `|filter` applied to the variable assignment:
https://github.com/mozilla/addons-server/blob/b8ffdae3c8a459ef61525aefa21c4debbb45896b/src/olympia/devhub/templates/devhub/addons/dashboard.html#L42-L44

KevinMind (approved):
🚀

PR #22692, opened 2024-09-20
Additional case handling for color parameter in search API

chrstinalin:
Fixes: mozilla/addons#15022

### Description

Handles cases as follows -- 
1. The empty color parameter is ignored. 
2. Otherwise, take the first 6 characters of the color parameter (after removing the #, if any). 3-character hex codes (ex. `#aaa`) still pass.
3. Throw ValueError if the value is still unparseable as a hex code.

### Testing

Directly use the search function for static themes, or to the API directly, ex.
`http://olympia.test/api/v5/addons/search/?type=statictheme&color=336699aa`

### Checklist

### DIFF
--- a/src/olympia/search/filters.py
+++ b/src/olympia/search/filters.py
@@ -442,12 +442,12 @@ class AddonColorQueryParam(AddonQueryParam):
         try:
             rgb = tuple(bytearray.fromhex(hexvalue))
         except ValueError:
-            rgb = (0, 0, 0)
+            raise ValueError(gettext('Expected a hex string color code.'))
         return colorgram.colorgram.hsl(*rgb)
 
     def get_value(self):
         color = self.query_data.get(self.query_param, '')
-        return self.convert_to_hsl(color.upper().lstrip('#'))
+        return self.convert_to_hsl(color.upper().lstrip('#')[:6]) if color else None
 
     def get_es_query(self):
         # Thresholds for saturation & luminosity that dictate which query to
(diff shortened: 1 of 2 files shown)


src/olympia/search/filters.py
```diff
@@ -441,13 +441,13 @@ def convert_to_hsl(self, hexvalue):
             hexvalue = ''.join(2 * c for c in hexvalue)
         try:
             rgb = tuple(bytearray.fromhex(hexvalue))
-        except ValueError:
-            rgb = (0, 0, 0)
+        except ValueError as err:
+            raise ValueError(gettext('Expected a hex string color code.')) from err
```
diox, line 445: The other filters in this file simply use `gettext('Invalid "%s" parameter.' % self.query_param)`, let's use that to be consistent.

(It doesn't specify the error but does specify the param name, which should be enough given that it's should be raised to users of the API, not end-users)

src/olympia/search/tests/test_filters.py
```diff
@@ -1059,6 +1071,11 @@ def test_search_by_color_grey(self):
             {'range': {'colors.ratio': {'gte': 0.25}}},
         ]
 
+    def test_search_by_color_invalid(self):
+        with self.assertRaises(serializers.ValidationError) as context:
+            self._filter(data={'color': '#gggggg'})
+        assert context.exception.detail == ['Expected a hex string color code.']
+
```
diox, line 1078: Also add a test for an empty param `'color': ''`

diox (approved):

PR #22693, opened 2024-09-23
Use db_backup + management commands to implement data dump/load/seed

KevinMind:
Relates to: mozilla/addons#15020
Unblocks: https://github.com/mozilla/addons-server/pull/22663

### Description

- migrate data backup management to use django manmagement commands and [db-backup](https://github.com/jazzband/django-dbbackup/tree/d8f09bda6a4d72eb94d3a5352c8d4d59f36f1f8a) under the hood
- add test coverage for the behaviour
- add synchronization of the storage volume with the database snapshot as those two data sources are tightly coupled.

### Context

In order to include data initialization into the make up flow, we need a better way to dump/load/seed data. This PR moves the mechanism for exporting/importing data to use the db-backup lib orchestrated by management commands, includes test coverage for these behaviours and removes the concept from our make files.

db-backup is a light wrapper around django database connectors to use mysql commands for dumping/loading. This side steps issues we have with using native django dump/load data commands. Before this patch, we had a db connector and could implement this logic ourselves.. but looking at this lib it is well maintained, has a lot of activity and is very transparent looking through the source code. I cannot imagine implementing something much different then what they already have.

This unblocks the initialization part because the command for initializing data in make up will use seeding (which uses dumping) and loading in order to make sure the database is in the "correct" state based on the the arguments passed by the developer.

This can be extracted as a separate PR to make review and testing easier as this part is relatively low impact.

From this PR we also get synchronization of the ./storage directory with the database backup which squashes a whole range of bugs related to mismatches from db file references and the underlying file system.

### Testing

1. run a data dump.

```sh
make data_dump
```

Expect a directory with the `timestamp` name is added to ./backups. It should include a file `db.sql` and a storage file `storage.tar` which is an exact replica of the ./storage directory.

Optionally pass an argument `ARGS=--name foo` to put the data in a specific named directory.

2. run data dump force

Assuming you dumped data to a directory with the name "foo" if you try to dump data there again, it will fail. You can make it pass by adding `ARGS="--name foo --force"` which will clean the foo directory before dumping.

3. Seed data

Now run the command

```sh
make initialize
```

This should reset your database to the initial data.

> Note. any data you previously had in your DB will now be gone. Expect no add-ons for the local_admin user.

5. Load data

You should be able to load data from a directory by passing the directory name. Use the name from step 1.

> Note, this should restore the data you had originally

```sh
make data_load ARGS=--name foo
```

If you don't pass a name or pass an invalid name the command will raise.

### Checklist


src/olympia/lib/settings_base.py
```diff
@@ -1325,6 +1328,9 @@ def read_only_mode(env):
 MEDIA_ROOT = os.path.join(SHARED_STORAGE, 'uploads')
 TMP_PATH = os.path.join(SHARED_STORAGE, 'tmp')
 
+DATA_BACKUP_DIR = path('backups')
+DATA_BACKUP_INIT = '_init'
```
eviljeff, line 1332: imo the setting should be named something more like `DATA_BACKUP_INITIAL_DIR_NAME`, but I'm not sure it even needs to be a setting - seems more like it should be a constant.

src/olympia/amo/management/commands/seed_data.py
```diff
@@ -0,0 +1,48 @@
...
+import os
+import shutil
+
+from django.conf import settings
+from django.core.management import call_command
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = 'Seed the _init data dir with fresh data from the database'
```
eviljeff, line 11: This help text really only makes sense if you know that `settings.DATA_BACKUP_INIT` defaults to `_init` and that's the name of the directory in the backups directory... at which point you don't need the help text 🙃

```suggestion
    help = 'Reset and seed the database with initial data, generated add-ons, and data from AMO prod'
```

(or something)

eviljeff (commented):
iirc @diox had concerns that django's `dumpdata` wasn't going to work for us, but I don't know the details.

docs/topics/development/data_management.md
```diff
@@ -35,30 +35,47 @@ The `make initialize` command, executed as part of `make initialize_docker`, per
...
-  make data_restore [RESTORE_DIR=<path>]
+  make data_load [ARGS="--name <name>"]
   ```
 
-  This command restores a MySQL database from a previously exported snapshot. The optional `RESTORE_DIR` argument allows you to specify the path of the import file.
-  This must be an absolute path. It defaults to the latest stored snapshot in the `backups` directory.
+  This command will load data from an existing backup directory. The name is required and must match a directory in the `DATA_BACKUP_DIR` directory.
 
-Refer to the Makefile for detailed instructions on these commands.
+  > NOTE: This command will NOT reindex elasticsearch. In most cases you should use the `make initialize_data` command instead.
```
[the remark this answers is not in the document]
KevinMind: This will be handled in a follow up PR where we introduce the `initialize` command. This will handle indexing. I wouldn't want to introduce that concept here as the load/dump commands are really just about managing the data and (for now at least) I do not know of a good way to dump/load the es index.

src/olympia/amo/management/commands/dump_data.py
```diff
@@ -0,0 +1,51 @@
...
+
+        logging.info(f'Dumping data to {dump_path}')
+
+        if os.path.exists(dump_path):
+            if force:
+                shutil.rmtree(dump_path)
+            else:
+                raise CommandError(
+                    f'Dump path {dump_path} already exists.'
+                    'Use --force to overwrite or --init to reseed the initial data.'
```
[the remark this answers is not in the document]
KevinMind: Fixed

src/olympia/lib/settings_base.py
KevinMind: Fixed

src/olympia/amo/management/commands/seed_data.py
KevinMind: Fixed

KevinMind:
> iirc @diox had concerns that django's `dumpdata` wasn't going to work for us, but I don't know the details.

I ran into these issues myself. I think it could be possible to mitigate but I found a library that does mysql level dumping/loading and also handles files.. so pretty much exactly what we are aiming for..

It's a dev dep and only installed on dev so I think it's relatively low risk to add. Also seems well maintained and I read through the code fairly easily.

docs/topics/development/data_management.md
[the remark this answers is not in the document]
eviljeff: This is about `data_load` though - that can be used at any time to load data from your backup folder -  not just during the first run `initialize`.  `make restore` used to call reindex so what's the argument against doing it for `make data_load`?

docs/topics/development/data_management.md
```diff
@@ -35,30 +35,47 @@ The `make initialize` command, executed as part of `make initialize_docker`, per
 
 ## Exporting and Loading Data Snapshots
 
-You can export and load data snapshots to manage data states across different environments or for backup purposes. The Makefile provides commands to facilitate this.
+You can export and load data snapshots to manage data states across different environments or for backup purposes. The Makefile provides commands to facilitate this. These commands rely internally on django `loaddata` and `dumpdata` commands
```
diox, line 38: The bit about using `loaddata` and `dumpdata` is no longer accurate

docs/topics/development/data_management.md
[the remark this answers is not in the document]
diox: I wouldn't worry about trying to dump/load ES and would straight up reindex (with `--wipe` if you want to ensure a clean slate).

I agree with @eviljeff that if you're loading data you should reindex to keep things consistent.

diox (commented):
In general

docs/topics/development/data_management.md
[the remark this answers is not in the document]
KevinMind: I don't want to make `data_load` the command you actually run.

If you want to run with a specific backup you will actually be able to run

```sh
make up INIT_LOAD=<backup_name>
```

Adding all of the functionality everywhere just adds complexity, more testing (can you opt out of indexing on data_load)... data_load is about "just" loading data. that's it.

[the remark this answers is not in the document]
KevinMind: You _could_ make the argument that reindex **only** get's called during `data_load` and then it would be totally internalized.

We would have to remove the ability to opt out of indexing though.. as it would be coupled to loading. IDK I can see both sides. @diox do you have an opinion?

docs/topics/development/data_management.md
KevinMind: Fixed

docs/topics/development/data_management.md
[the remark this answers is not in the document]
KevinMind: I'm coming around to your viewpoint actually. `load_data` totally resets the database and storage.. there really isn't a reason it should not also reindex. I will address the upstream issue in the next PR.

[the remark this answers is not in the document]
KevinMind: Fixed

src/olympia/amo/management/commands/data_seed.py
```diff
@@ -0,0 +1,42 @@
...
+            'createsuperuser',
+            '--no-input',
+            '--username',
+            settings.LOCAL_ADMIN_USERNAME,
+            '--email',
+            settings.LOCAL_ADMIN_EMAIL,
+        )
+        self.call_command('loaddata', 'zadmin/users')
+
+        self.logger.info('Generating add-ons...')
```
diox, line 35: You want the `reindex` with `--wipe` here as well.

Otherwise the add-ons being generated below are indexed using the automatic mapping, and without our aliases set up, which is wrong and would require a `--wipe` later to fix anyway.

KevinMind: Do we need it before and after? If we call `data_load` after running the seed commands that will end with a reindex. lmk if you think we still need to wipe before.

diox: As it stands, until the reindex that happens in `data_load`, the indexes will be incorrect. So technically we don't need both calls if we can guarantee that everything happens uninterrupted. But I'm not sure we have that guarantee...

I think it's safer to do it here as well. But then again I'm not sure I understand why you're calling dump and load in succession in the seed command...

KevinMind: Fixed. For now it's in both. I think this is generally fine and we can iterate more if we want to.

Makefile-docker
```diff
@@ -52,37 +49,26 @@ check_django: ## check if the django app is configured properly
...
-	# The superuser needs to have a mozilla.com address for admin tools access
-	$(PYTHON_COMMAND) manage.py createsuperuser \
-		--no-input \
-		--username "local_admin" \
-		--email "local_admin@mozilla.com"
-	$(PYTHON_COMMAND) manage.py loaddata zadmin/users
+	# Seed the database with initial data
+	./manage.py data_seed
+	# Load the data from the initial dump
+	$(MAKE) data_load ARGS="--name _init"
```
eviljeff, line 66: it's unfortunate we hard-code `_init` here _and_ in amo/management/base.py

src/olympia/amo/management/commands/data_seed.py
```diff
@@ -0,0 +1,46 @@
...
+        call_command('generate_addons', '--app', 'firefox', num_addons)
+        call_command('generate_addons', '--app', 'android', num_addons)
+        call_command('generate_themes', num_themes)
+
+        call_command('generate_default_addons_for_frontend')
+
+        call_command('data_dump', '--name', self.data_backup_init)
+        call_command('data_load', '--name', self.data_backup_init)
```
eviljeff, line 46: Is the `data_load` needed after the `data_dump`?  It'd be loading the same data (data_dump doesn't delete anything, does it?).

src/olympia/blocklist/tests/test_admin.py
```diff
@@ -1361,7 +1361,7 @@ def test_signoff_approve(self):
         log_entry = LogEntry.objects.last()
         assert log_entry.user == user
         assert log_entry.object_id == str(mbs.id)
-        other_obj = addon_factory(id=mbs.id)
+        other_obj = addon_factory()
```
eviljeff, line 1364: Seems like a random change in an unrelated part of the codebase? iirc it's using the same pk as mbs (the `BlockSubmission` instance - previously `MultiBlockSubmission`) to test the LogEntry shown on the page are being correctly filtered on content type, not just id.

src/olympia/blocklist/tests/test_admin.py
```diff
@@ -1437,7 +1437,7 @@ def test_signoff_reject(self):
...
```
eviljeff, line 1440: same question here

src/olympia/amo/management/base.py
```diff
@@ -0,0 +1,48 @@
+import logging
+import os
+import shutil
+
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+
+
+class BaseDataCommand(BaseCommand):
```
diox, line 9: There is already a base command class in `olympia/amo/management/__init__.py` so you don't need a new file

src/olympia/blocklist/tests/test_admin.py
```diff
@@ -1361,7 +1361,7 @@ def test_signoff_approve(self):
         log_entry = LogEntry.objects.last()
         assert log_entry.user == user
         assert log_entry.object_id == str(mbs.id)
-        other_obj = addon_factory(id=mbs.id)
```
diox, line 1364: That's an unrelated change that shouldn't be in this PR. Same below.

src/olympia/amo/management/commands/data_dump.py
```diff
@@ -0,0 +1,48 @@
...
+
+            call_command(
+                'mediabackup',
+                output_path=storage_path,
+                interactive=False,
+                compress=True,
+            )
+        except Exception as e:
+            self.clean_dir(dump_path)
+            raise e
```
eviljeff, line 48: nit, but
```suggestion
        finally:
            self.clean_dir(dump_path)
```

src/olympia/blocklist/tests/test_admin.py
eviljeff: hah, snap

src/olympia/amo/management/commands/data_seed.py
KevinMind: This guarantees that the data in the database is exactly the data that is dumped to `_init` it seems redundant but it is much more "predictable" this way. 

the `_init` seed can be created once and used infinitely and we want to make sure the database state after the first time matches the N time.

src/olympia/amo/management/commands/data_dump.py
KevinMind: We ONLY want to clean if there is an exception raised, this would remove the dump after we successfully created it.

Makefile-docker
KevinMind: I actually don't need that line because data_seed already loads the data. so that's totally redundant.

src/olympia/amo/management/base.py
KevinMind: TIL. I'll update it.

src/olympia/amo/management/commands/data_dump.py
eviljeff: ah, right.

eviljeff:
@KevinMind I'm attempting to follow the testing instructions but I don't think they're up to date?  `make data_seed` doesn't exist any longer, and there's no mention of the changes to `make initialize`

KevinMind:
> @KevinMind I'm attempting to follow the testing instructions but I don't think they're up to date? `make data_seed` doesn't exist any longer, and there's no mention of the changes to `make initialize`

@eviljeff updated. You should run `make initialize` this will in turn run the seed management command

src/olympia/amo/management/base.py
KevinMind: Fixed

Makefile-docker
KevinMind: Fixed

src/olympia/blocklist/tests/test_admin.py
KevinMind: Removed

src/olympia/blocklist/tests/test_admin.py
KevinMind: Removed

src/olympia/amo/management/commands/data_seed.py
eviljeff: I just can't think of a scenario where it would be different - if it is then it'd be a bug?

docs/topics/development/data_management.md
```diff
@@ -35,30 +35,46 @@ The `make initialize` command, executed as part of `make initialize_docker`, per
...
+and the actual data is stored in a persistent data volume. This enables data to persist across container restarts.
 
-By following these practices, developers can manage data effectively in the **addons-server** project. The use of persistent volumes, external mounts, data snapshots, and automated data population ensures a robust and flexible data management strategy. For more detailed instructions, refer to the project's Makefile and Docker Compose configuration in the repository.
+`addons-server` assumes that a database named `olympia` already exists and most data management commands will fail
+if it does not.
+
+If you need to hard reset the database (for example, to start with a fresh state), you can use the following command:
+
+```bash
+make down && docker_mysqld_volume_remove
```
eviljeff, line 73: This fails with "docker_mysql_volume_remove: command not found".  If you run the commands separately the second fails because `make down` stops the `web` container that `docker_mysqld_volume_remove` relies on.

src/olympia/amo/management/commands/data_seed.py
KevinMind: I can remove it.

docs/topics/development/data_management.md
eviljeff: Does that actually work for you? `make down` stops the `web` container that `make docker_mysqld_volume_remove` relies on

KevinMind: you do not need the container running to remove the volume, yes it works. To be clean I'm talking about the updated code

```sh
make down && make docker_mysqld_volume_remove
```

eviljeff: Okay, it's working now (not sure why it didn't earlier).  

It exposes another bug though - `make docker_mysqld_volume_remove` deletes the database, but we don't now call `create_db` anywhere so there isn't a database.

KevinMind: The DB is created by the mysql container on startup. You only need to create the volume which happens during make up automatically. Initialize then handles seeding and loading data based on the state of the DB and your arguments.

KevinMind: It failed the first time because of a missing `make` in the second command.

diox:
```
% make down && make docker_mysqld_volume_remove
docker compose down --rmi local --remove-orphans --volumes
[+] Running 17/17
 ✔ Container olympia-worker-1           Removed                                                                                                                                         10.4s 
 ✔ Container olympia-rabbitmq-1         Removed                                                                                                                                          6.9s 
 ✔ Container olympia-elasticsearch-1    Removed                                                                                                                                          0.7s 
 ✔ Container olympia-nginx-1            Removed                                                                                                                                          0.5s 
 ✔ Container olympia-memcached-1        Removed                                                                                                                                          0.5s 
 ✔ Container olympia-mysqld-1           Removed                                                                                                                                          4.0s 
 ✔ Container olympia-redis-1            Removed                                                                                                                                          0.5s 
 ✔ Container olympia-autograph-1        Removed                                                                                                                                          0.6s 
 ✔ Container olympia-web-1              Removed                                                                                                                                         10.4s 
 ✔ Container olympia-addons-frontend-1  Removed                                                                                                                                          0.4s 
 ✔ Volume olympia_storage               Removed                                                                                                                                          0.0s 
 ✔ Image olympia-customs:latest         Removed                                                                                                                                          0.0s 
 ✔ Volume olympia_data_rabbitmq         Removed                                                                                                                                          0.0s 
 ✔ Volume olympia_data_olympia          Removed                                                                                                                                          0.0s 
 ✔ Volume olympia_data_elasticsearch    Removed                                                                                                                                          0.0s 
 ✔ Volume olympia_data_redis            Removed                                                                                                                                          0.0s 
 ✔ Network olympia_default              Removed                                                                                                                                          0.2s 
docker image prune --filter "dangling=true" --force
Total reclaimed space: 0B
docker volume prune --force
Total reclaimed space: 0B
docker volume rm addons-server_data_mysqld
Error response from daemon: remove addons-server_data_mysqld: volume is in use - [ceb8902b2d39250b76abe6a411167e6c037acee37b83a25a8a05eb3c972a9411]
make: *** [Makefile-os:115: docker_mysqld_volume_remove] Error 1
```

Makefile-docker
```diff
@@ -52,37 +49,24 @@ check_django: ## check if the django app is configured properly
...
-	$(PYTHON_COMMAND) manage.py migrate --noinput
-	$(PYTHON_COMMAND) manage.py loaddata initial.json
-	$(PYTHON_COMMAND) manage.py import_prod_versions
-	# The superuser needs to have a mozilla.com address for admin tools access
-	$(PYTHON_COMMAND) manage.py createsuperuser \
-		--no-input \
-		--username "local_admin" \
-		--email "local_admin@mozilla.com"
-	$(PYTHON_COMMAND) manage.py loaddata zadmin/users
+	# Seed the database with initial data
```
eviljeff, line 63:
```suggestion
    ./manage.py create_db --force
	./manage.py migrate --noinput
	# Seed the database with initial data
```
From our call, this is what was missing

docs/topics/development/data_management.md
KevinMind: @diox had you already run make down? the volume is used by mysql so if the container is running docker will complain. so long as mysql is not running you should be able to remove the volume.

Makefile-docker
KevinMind: Fixed

docs/topics/development/data_management.md
KevinMind: Can you try running

```sh
docker ps -a --filter volume=addons-server_data_mysqld --format "{{.ID}}: {{.Names}}"
```

This should tell us which container is using it.

diox (approved):

PR #22699, opened 2024-09-24
Handle Non-Alphanumerical Edge Cases When Searching Themes By Colour

chrstinalin:
Fixes: mozilla/addons#15022

### Description

Moved all formatting to its own function (convert_to_hex) that either returns a valid 6-character hex string or raises a ValueError.

### Checklist

### DIFF
--- a/src/olympia/search/filters.py
+++ b/src/olympia/search/filters.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+import re
 
 from django.utils import translation
 from django.utils.encoding import force_str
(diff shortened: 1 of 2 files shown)


src/olympia/search/filters.py
```diff
@@ -449,7 +454,7 @@ def convert_to_hsl(self, hexvalue):
 
     def get_value(self):
         color = self.query_data.get(self.query_param, '')
-        return self.convert_to_hsl(color.upper().lstrip('#')[:6]) if color else None
+        return self.convert_to_hsl(self.convert_to_hex(color)) if color else None
```
diox, line 457: Need to call `convert_to_hex()` above and stick the result into `color` otherwise you'll end up potentially passing `None` to `convert_to_hsl()`, as the `if color` bit only applies to before the conversion to hex at the moment.

chrstinalin: When None is passed to the convert_to_hex(), it throws the ValueError. If its done beforehand and returns None here, `get_es_query` will return [] (which isn't necessarily what's expected, if it was an invalid input). I don't particularly like how I have it set up right now though. I'm modifying it to be more explicit.

diox (approved):

PR #22702, opened 2024-09-25
Add basic documentation about the various types of logs that AMO has

diox:
Fixes: mozilla/addons#15047

### DIFF
--- a/docs/topics/logs.md
+++ b/docs/topics/logs.md
@@ -1,0 +1,36 @@
+# Logs
+
+(logs)=
+
+We have various types of logs with different purposes in AMO. The following
+tables summarize their characteristics:
+
+
+| | `ActivityLog` / `LogEntry` |
+|-| -------------------------- |
+|Type| Database entry |
+|Purpose| Storing information about developers/reviewers/admin actions |
+|Stores IP| Depending on the action |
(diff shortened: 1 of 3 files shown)


docs/topics/logs.md
```diff
@@ -0,0 +1,36 @@
+# Logs
+
+(logs)=
+
+We have various types of logs with different purposes in AMO. The following
+tables summarize their characteristics:
+
+
+| | `ActivityLog` / `LogEntry` |
```
KevinMind, line 9: Did you consider using a markdown table or at least a multi column table? Seems the rows are repeated for each type and wondered if it's more readable as a single table? I downloaded the built docs site and it's unclear. 

Up to you.

KevinMind (approved):
LGTM. Left a comment but not a blocker.

image

docs/topics/logs.md
diox: Yes, I tried that, that's actually how I originally had it, but it overflowed (the theme we're using only allows a `max-width` of `800px` on the content) so I figured I would make separate tables for now.

KevinMind: Thought that might be the case 👍

PR #22705, opened 2024-09-26
reset File.original_status when rejected

eviljeff:
Fixes: mozilla/addons#15023

### Description

resets File.original_status to STATUS_NULL when rejecting versions

### Context

The issue scenario was only possible with unlisted versions (i.e. reject_multiple_versions) because the listed reject_version action works on current_version, and a disabled version (even if disabled by the developer) can't be current_version.

### Testing
(you can probably do 1-3 in django admin/shell, to skip the full STR)
1) have an unlisted version that's approved
2) delay reject that version in reviewer tools
3) developer disable that version in devhub
4) change the pending rejection date/time in django admin to a date/time in the past in `/admin/models/versions/version/<version-id>/change/`, "Version reviewer flags" section.
5) run `./manage.py cron auto_reject`
6) see in devhub you can't re-enable that version
7) (double check `original status` is 0 (STATUS_NULL) in django admin/shell)

### Checklist

### DIFF
--- a/src/olympia/reviewers/tests/test_utils.py
+++ b/src/olympia/reviewers/tests/test_utils.py
@@ -3672,6 +3672,33 @@ class TestReviewHelper(TestReviewHelperBase):
         }
         assert set(appeal_job2.reload().decision.policies.all()) == {policy_d}
 
+    def test_reject_multiple_versions_resets_original_status_too(self):
+        old_version = self.review_version
+        old_version.file.update(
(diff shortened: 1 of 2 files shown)


KevinMind (approved):
I think the test instructions are a bit off.

```sh
run ./manage.py cron auto_reject
```

should be 

```sh
run ./manage.py auto_reject
```
And it should specify we should see our version number in the logs of rejected versions.

Verified I cannot seem to undisable the version.

image

Verified the file original_status is 0

image

Top notch testing instructions mate 👌

PR #22706, opened 2024-09-26
Add BlockType enum to BlockVersion model

KevinMind:
Relates to: mozilla/addons#15014

### Description

TBD

### Checklist

### DIFF
--- a/src/olympia/blocklist/models.py
+++ b/src/olympia/blocklist/models.py
@@ -1,5 +1,6 @@
 from collections import defaultdict, namedtuple
 from datetime import datetime
+from enum import Enum
 
 from django.conf import settings
 from django.db import models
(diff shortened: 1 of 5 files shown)


src/olympia/blocklist/models.py
```diff
@@ -155,11 +155,34 @@ def get_blocks_from_guids(cls, guids):
...
+
+
+class BlockVersionManager(ManagerBase):
+    _queryset_class = BlockVersionQuerySet
+
+    def soft_blocked(self):
+        return self.get_queryset().soft_blocked()
+
+    def hard_blocked(self):
+        return self.get_queryset().hard_blocked()
```
diox, line 176: You might be able to replace that with just... nothing, and defining `objects` below with [`as_manager()`](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.as_manager), like this: `objects = BlockVersionQuerySet.as_manager()`.

For more complex cases where you do want the functionalty `ManagerBase` brings we can also define an empty manager inheriting from `ManagerBase` and use [`from_queryset()`]( (https://docs.djangoproject.com/en/4.2/topics/db/managers/#from-queryset)) like this: `BlockVersionManager.from_queryset(BlockVersionQuerySet)()`

If we have custom methods in the manager that somehow don't belong in the queryset but do call queryset stuff, then it's no longer appropriate, but most of the time that should be enough. We have a bunch of old code using the method you used, but for new cases without anything fancy we should try the modern Django way and avoid repetition.

KevinMind: This belongs in the docs.

PR #22708, opened 2024-09-27
Only include locale stage in production image

KevinMind:
fixes: https://github.com/mozilla/addons/issues/15055

### Description

Skip locale compilation outside of the production image build.
This is a difficult to cache operation that is not necessary for development

### Context

In order to speed up our local make up flow there are stages we can optimize and others we can totally skip. This is a low hanging fruit as we don't need compiled .mo files for dev at all.

### Testing

1. make up should work and skip the locale stage if `DOCKER_TARGET=development`
2. make up should work and include the locale stage if `DOCKER_TARGET=production`
3. in a production image you should see the .mo files in the container, in development not.
4. running `make test_needs_locale_compilation` will fail on development images if you don't first run the compilation step.

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -172,8 +172,6 @@ ENV DOCKER_VERSION=${DOCKER_VERSION}
 COPY docker/etc/mime.types /etc/mime.types
 # Copy the rest of the source files from the host
 COPY --chown=olympia:olympia . ${HOME}
-# Copy compiled locales from builder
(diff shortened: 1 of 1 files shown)


diox:
This and https://github.com/mozilla/addons-server/pull/22710 should have their own issues as they don't just potentially affect local dev and would benefit from a quick check from QA once they land.

diox (approved):
r+ but see comment about needing an issue

PR #22709, opened 2024-09-27
Fix sponsored/verified removal migration

diox:
Fixes: mozilla/addons#15052

The damage is done on dev, but we can still save stage & prod...


src/olympia/promoted/migrations/0021_auto_20240919_0952.py
```diff
@@ -1,13 +1,15 @@
...
+from olympia.constants.promoted import RECOMMENDED
 from django.db import migrations
 
 
 def delete_verified_sponsored_promoted_usage(apps, schema_editor):
     PromotedAddon = apps.get_model('promoted', 'PromotedAddon')
     PromotedApproval = apps.get_model('promoted', 'PromotedApproval')
-    PromotedAddon.objects.filter(group_id__in=(3, 4)).update(group_id=0)
-    PromotedApproval.objects.filter(group_id__in=(3, 4)).delete()
+    # 2 and 3 were Sponsored and Verified, respectively.
```
KevinMind, line 10: Can we avoid using the raw numbers and instead use something with a "name" like PROMOTED_GROUPS.VERIFIED or something?

KevinMind: We could maintain constants for them though.. maybe not needed 🤷

diox: `VERIFIED` no longer exists. I almost changed it to `.exclude(group_id__in=PROMOTED_GROUPS_BY_ID.keys())` but then it would have potentially introduced a bug down the line depending on when you run the migration, because the constant would not be the historical value from when the migration was written, so if its content changes again in the future and somehow that migration is re-ran it won't be accurate...

Looking back, I think we should have kept the original constant with a `_` prefix and a comment, but it's a bit late now... I could re-introduce it, but we're back to the original problem that having it around was confusing and added maintenance.

KevinMind: Can we do that. Like just a variable per id or a tuble of "DELETED_PROMOTION_GROUPS". It doesn't add maintenance because it should never change, unless we delete another one, in which case it's actually useful.

Could even verify there are no Addons with the group ID in those gouups for a next migration.

diox: https://github.com/mozilla/addons-server/pull/22709/commits/947aee76bce18a44ab62d8b2915e605961acb825 re-introduces dummy constants. Unfortunately we have to have a `name` and `api_name` set for `PromotedClass`.

KevinMind (approved):
🙈 LGTM

src/olympia/promoted/migrations/0021_auto_20240919_0952.py
eviljeff: The obsolete names are hard-coded in search/filters too, for the API_GATE - it would have been good to change that code to reference these reintroduced constants, somehow.

PR #22724, opened 2024-09-30
Stop using our remote settings "test server mode" for dev, it's broken

diox:
Fixes https://github.com/mozilla/addons/issues/1939

### Context

`REMOTE_SETTINGS_IS_TEST_SERVER` is a special mode designed to work around the fact that the Kinto server was being regularly wiped out, so we had this special user that created the actual user and collection dynamically, and then used its own special bucket. But nowadays we can just use the same user/bucket/collection everywhere, as the remote settings server has the same config everywhere.

### Testing

This is difficult to test without the credentials, but you can check that this is consistent with stage. I manually logged in to dev and used our `RemoteSettings` class, bypassing `REMOTE_SETTINGS_IS_TEST_SERVER` to ensure it works. This resulted in https://remote-settings-dev.allizom.org/v1/buckets/blocklists/collections/addons-bloomfilters/records (which was empty before, and has been for the past few years...)

### DIFF
--- a/src/olympia/conf/dev/settings.py
+++ b/src/olympia/conf/dev/settings.py
@@ -74,6 +74,6 @@ FXA_CONTENT_HOST = 'https://accounts.stage.mozaws.net'
 FXA_OAUTH_HOST = 'https://oauth.stage.mozaws.net/v1'
 FXA_PROFILE_HOST = 'https://profile.stage.mozaws.net/v1'
 
-REMOTE_SETTINGS_IS_TEST_SERVER = True
+REMOTE_SETTINGS_WRITER_BUCKET = 'staging'
 
 SITEMAP_DEBUG_AVAILABLE = True


eviljeff (commented):
Can we just delete support for test server mode entirely?

diox:
I wasn't confident enough to do that at first, but now that I have a better understanding of the state of things, yes, let's do that. https://github.com/mozilla/addons-server/pull/22724/commits/3ecfc82d810069f539e50189d94ef91bc041c7b8

src/olympia/conf/dev/settings.py
```diff
@@ -74,6 +74,6 @@
 FXA_OAUTH_HOST = 'https://oauth.stage.mozaws.net/v1'
 FXA_PROFILE_HOST = 'https://profile.stage.mozaws.net/v1'
 
-REMOTE_SETTINGS_IS_TEST_SERVER = True
+REMOTE_SETTINGS_WRITER_BUCKET = 'staging'
```
eviljeff, line 77: we can probably just change the default in settings_base to `staging` and drop the overrides in prod and stage too

src/olympia/lib/tests/test_remote_settings.py
```diff
@@ -16,114 +16,11 @@
...
-        )
-        responses.add(
-            responses.PUT,
-            settings.REMOTE_SETTINGS_WRITER_URL + 'buckets/foo/collections/baa',
-            content_type='application/json',
-            status=201,
-        )
-        server.setup_test_server_collection()
-
     @override_settings(REMOTE_SETTINGS_IS_TEST_SERVER=False)
```
eviljeff, line 19: as this patch drops all support for `REMOTE_SETTINGS_IS_TEST_SERVER` actually doing anything, we should delete the setting in settings_base (and this override) too.

eviljeff (approved):

src/olympia/lib/tests/test_remote_settings.py
diox: https://github.com/mozilla/addons-server/pull/22724/commits/addf8ff91cd776ed2463373c5240343cf483cdae

src/olympia/conf/dev/settings.py
diox: https://github.com/mozilla/addons-server/pull/22724/commits/addf8ff91cd776ed2463373c5240343cf483cdae

PR #22727, opened 2024-10-01
reset file.status_disabled_reason when rejecting too

eviljeff:
Fixes: mozilla/addons#15023 (again)

### Description
Resets `File.status_disabled_reason` to `File.STATUS_DISABLED_REASONS.NONE` in the reviewer tools when a file/version is rejected.

### Context

https://github.com/mozilla/addons-server/pull/22705 only reset `File.original_status` and I forgot that when `File.status_disabled_reason` was added (by me 😬) `Version.is_user_disabled` was changed to rely on it, so it also needs resetting to a null state.

### Testing

the testing steps from https://github.com/mozilla/addons-server/pull/22705 ... but I guess something was missing from it because the version should still have been possible to un-disable. in devhub - and then doing that would have set the File status to 0 (as visible in the reviewer tools).  So check that doesn't happen.

### Checklist

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -21,6 +21,7 @@ from olympia.activity.utils import notify_about_activity_log
 from olympia.addons.models import Addon, AddonApprovalsCounter, AddonReviewerFlags
 from olympia.constants.abuse import DECISION_ACTIONS
 from olympia.constants.promoted import RECOMMENDED
+from olympia.files.models import File
 from olympia.lib.crypto.signing import sign_file
 from olympia.reviewers.models import (
     AutoApprovalSummary,
@@ -867,6 +868,7 @@ class ReviewBase:
         file.status = status
         if status == amo.STATUS_DISABLED:
             file.original_status = amo.STATUS_NULL
+            file.status_disabled_reason = File.STATUS_DISABLED_REASONS.NONE
         file.save()
 
     def set_promoted(self, versions=None):
--- a/src/olympia/reviewers/tests/test_utils.py
+++ b/src/olympia/reviewers/tests/test_utils.py
@@ -3675,7 +3675,9 @@ class TestReviewHelper(TestReviewHelperBase):
     def test_reject_multiple_versions_resets_original_status_too(self):
         old_version = self.review_version
         old_version.file.update(
-            status=amo.STATUS_DISABLED, original_status=amo.STATUS_APPROVED
+            status=amo.STATUS_DISABLED,
+            original_status=amo.STATUS_APPROVED,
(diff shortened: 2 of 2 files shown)


src/olympia/reviewers/utils.py
```diff
@@ -867,6 +868,7 @@ def set_file(self, status, file):
         file.status = status
```
KevinMind, line 868: Now that I'm looking at it again, this should probably have been a model method. `file.reset_disabled()`

KevinMind:
@eviljeff I'm not sure I follow.

```sh
the testing steps from https://github.com/mozilla/addons-server/pull/22705 ... but I guess something was missing from it because the version should still have been possible to un-disable. in devhub
```

When I verified that PR, I was NOT able to undisable the version after running auto reject.. Am I now expecting to be able or not be able to undisable the version.

eviljeff:
> @eviljeff I'm not sure I follow.
> 
> ```shell
> the testing steps from https://github.com/mozilla/addons-server/pull/22705 ... but I guess something was missing from it because the version should still have been possible to un-disable. in devhub
> ```
> 
> When I verified that PR, I was NOT able to undisable the version after running auto reject.. Am I now expecting to be able or not be able to undisable the version?

QA found you should have been (erroneously) able to undisable the version after running auto reject after #22705 - that's part of the bug this patch is trying to solve. But I can't quite see what was wrong with my original testing instructions that led you to validate the patch as working when it wasn't.

You should not be able to undisable the version.

diox:
FWIW, on `master`, using the original testing instructions, I was able to undisable the version after the auto-rejection (it never got back to public, ended up with `status=0`, which is what we saw in prod IIRC)

diox (approved):

src/olympia/reviewers/utils.py
KevinMind: I'm talking about something different.

A new method on either the version or the file that essentially encapsulates all of the logic we have here in a utility (_recalling @diox  mentioning we should lean towards business logic in models_) into a method we can call for a particular file/version. 

(I guess the distinction is also somewhat opaque since previously it's been said there is little tangible difference between the two at this point)

No action needed unless @eviljeff agrees and is keen to implement but I thought it was worth mentioning there should be a way to add this on the model and that would probably be "better" in a latte art code kind of way.

KevinMind (approved):
If it works.. 🚢

src/olympia/reviewers/utils.py
eviljeff: There's an argument we should have methods to reject a version, to approve a version, etc, but that's not how the reviewer tools currently works.  Something to consider for future refactoring though.

PR #22728, opened 2024-10-01
Hide attachment upload on force disable/enable

chrstinalin:
Fixes: mozilla/addons#15063

### Description

Removes the option to upload attachments for force disable/enable add-ons.

image
image

### DIFF
--- a/static/js/zamboni/reviewers.js
+++ b/static/js/zamboni/reviewers.js
@@ -1,3 +1,5 @@
+NO_ATTACHMENT_ACTIONS = ['disable_addon', 'enable_addon'];
+
 $(document).ready(function () {
   if ($('.daily-message').length) {
     initDailyMessage();
@@ -110,6 +112,11 @@ function initReviewActions() {
     // the value we're interested in.
     $data_toggle_hide.show();
     $data_toggle_hide.filter('[data-value~="' + value + '"]').hide();
+
+    $('.review-actions-attachment-reply').toggleClass(
+      'hidden',
+      NO_ATTACHMENT_ACTIONS.includes(value),
+    );
   }
 
   $('#review-actions .action_nav #id_action > *:not(.disabled)').click(


static/js/zamboni/reviewers.js
```diff
@@ -1,3 +1,5 @@
+NO_ATTACHMENT_ACTIONS = ['disable_addon', 'enable_addon'];
```
KevinMind, line 1: Can you add a comment of where the source of these strings is, since they are literals.

static/js/zamboni/reviewers.js
```diff
@@ -110,6 +112,11 @@ function initReviewActions() {
     // the value we're interested in.
     $data_toggle_hide.show();
     $data_toggle_hide.filter('[data-value~="' + value + '"]').hide();
+
+    $('.review-actions-attachment-reply').toggleClass(
```
KevinMind, line 116: Just clarifying.. did you look for a way to add this to the template. Relying on JS to hide things can be tricky.

1. the element will be visible until js loads and this particular function is executed.
2. the element will still be in the dom, just not visible.

It would be much better to hide this element in the view (server side) OR to invert the logic and make it invisible by default and only make it "visible" on certain actions.

chrstinalin: All the toggling between action panels happens in this function above the line I added, including the other form inputs that need to be shown/hidden, so I don't think there's a way to avoid JS. I did realize I could just route the show/hide logic through that existing JS, though, so at least there isn't any _new_ JS.

KevinMind:
Code looks good. Could you add testing instructions in either the PR or the ticket. Are there any waffle flags or specific data requirements to verify this feature? Should it be for listed/unlisted/both etc etc.

This way I will be able to verify the PR and QA will also be able to verify when it is merged and shipped to Dev.

chrstinalin:
@KevinMind Just made it more explicit in the PR

KevinMind (approved):
Ship it 🚢

PR #22729, opened 2024-10-01
Add Waffle Flag for Submissions

chrstinalin:
Fixes: mozilla/addons#1830

### Description

Adds waffle flag `enable-submissions` to disable add-on related submissions (both new add-ons and versions) when toggled to `false` for a particular group on both DevHub and the API. Includes the endpoints:

1. POST `api/v5/addons/addon/`
2. PUT (CREATE) `api/v5/addons/addon/<guid>/`
3. POST `api/v5/addons/addon/<guid>/versions/`
4. POST `api/v5/addons/addon/<guid>/previews/`
5. PUT `api/v4/addons/`
6. POST `api/v4/addons/<guid>/versions/<version>`
7. PUT `api/v3/addons/`
8. POST `api/v3/addons/<guid>/versions/<version>`

The waffle flag's 'note' can be used to add additional information. The API response:
```
{
    "error": "Submissions are not currently available.",
    "reason": "This is the note reason."
}
```
And on DevHub:
image
image

### Testing
1. Deactivate the `enable-submissions` waffle flag (everyone should be = `True` by default).
2. On DevHub, all URLs pointing to a) uploading a new add-on, or b) uploading a new version are blocked.
3. Any API requests to endpoints that do the same return the error.

### Checklist


diox:
Can you refactor the conditions in the views into a decorator or two (one for APIs, one for "regular" views) to reduce duplicated code ?

src/olympia/addons/decorators.py
```diff
@@ -55,3 +58,15 @@ def addon_view_factory(qs):
...
+
+
+def require_submissions_enabled(f):
+    """Require the enable-submissions waffle flag to be enabled."""
+
+    @functools.wraps(f)
+    def wrapper(request, *args, **kw):
+        if waffle.flag_is_active(request, 'enable-submissions'):
+            return f(request, *args, **kw)
+        return submissions_disabled_response()
```
diox, line 70: Sorry, one more change with the logic: we could have a single decorator actually. Inside, you can return a `Response` (API) or a `TemplateResponse` (non-API) by doing a check on `request.is_api`. And I don't think we need to have the code in utils, it's short enough and never re-used anywhere else, we could have it directly in the decorator IMHO. That allows you to share the bit that retrieves the note between the 2 code paths as well.

src/olympia/amo/templates/amo/submissions_disabled.html
```diff
@@ -0,0 +1,18 @@
...
+
+{% block main_content %}
+<div role="main">
+  <section>
+    <h1>{{ _('Submissions are not currently available.') }}</h1>
+    <p>
+      {% if reason %}
+      {{ reason }}
+      {% else %}
+      Submissions have been temporarily disabled. Please check back later.
```
diox, line 13: Should be translated

diox: (I've asked for help from product/operations on the copy in the issue)

src/olympia/amo/templates/amo/submissions_disabled.html
```diff
@@ -0,0 +1,18 @@
...
+
+{% block title %}{{ _('Submissions Unavailable') }}{% endblock %}
+
+{% block main_content %}
+<div role="main">
+  <section>
+    <h1>{{ _('Submissions are not currently available.') }}</h1>
+    <p>
+      {% if reason %}
+      {{ reason }}
```
diox, line 11: indent

diox (changes requested):
I think we should be a little more subtle than that (sorry it was so poorly specified in the issue):
- We don't need to completely block all steps of the submission process. Let's only block the part where the add-on is being submitted, that's what matters - for the rest, it's too late anyway.
- On the page where the developer is selecting the file, let's allow access to the page if it's a `GET`, but disable the inputs/buttons and show the error message there. Or even replace the entire <form> with the message, whatever is easier. That way developers should never be surprised (and in theory shouldn't hit the 403 submissions_disabled.html page, unless they tried to be sneaky or submitted right when the flag was modified)

src/olympia/addons/decorators.py
```diff
@@ -55,3 +61,30 @@ def addon_view_factory(qs):
...
+            return f(request, *args, **kw)
+        flag = waffle.get_waffle_flag_model().get('enable-submissions')
+        reason = flag.note if hasattr(flag, 'note') else None
+        if getattr(request, 'is_api', True):
+            return Response(
+                {
+                    'error': gettext('Add-on uploads are temporarily unavailable.'),
+                    'reason': reason,
+                },
+                status=status.HTTP_403_FORBIDDEN,
```
diox, line 81: Here and below: I think we should use a 503, not a 403. It's a temporary error caused by us, not something the developer can do anything about.

src/olympia/devhub/views.py
```diff
@@ -1363,6 +1364,7 @@ def submit_version_agreement(request, addon_id, addon):
 
 
 @transaction.atomic
+@require_submissions_enabled
```
diox, line 1367: Not necessary to protect this page (see comment below about needing to protect `upload()` and how to integrate the error message better in the submit upload page, though)

diox (approved):

diox (changes requested):
Wrong button before, didn't mean to approve, definitely need changes.

diox (approved):
We can ship this, but create an issue to address these points in a follow-up:

- The Theme Submission page (`/en-US/developers/addon/submit/theme/upload-listed`) has the `Create a theme` button still enabled when the flag is off
- The Theme Wizard page (that the aforementioned button leads to - `/en-US/developers/addon/submit/wizard-listed`) also has its submit button still enabled
- Some warning should be displayed somewhere on the upload & theme wizard pages and not just on the `title` of the disabled buttons, to make it more obvious. With the full `note` for consistency.

PR #22732, opened 2024-10-02
Fix user-media loading in local environments

diox:
Fixes: mozilla/addons#15025

### Context

URLs with `/user-media/foo` should map to `<root_storage>/user-media/shared_storage/uploads/ `on the filesystem.

### Testing

You can try creating a theme through the theme wizard and then make sure its preview is visible in devhub. That shouldn't have affected downloads served through `X-Accel-Redirect` but trying to download an xpi is also a good regression test.


docker/nginx/addons.conf
```diff
@@ -20,7 +20,7 @@ server {
     }
 
     location /user-media/ {
-        alias /srv/user-media/;
+        alias /srv/user-media/shared_storage/uploads/;
```
KevinMind, line 23: Since this is such a tiny PR, could you add the basic test?

in _test.yml you can add another job that will just echo the file in the ./storage/shared_storage/uploads and then you should be able to curl it from nginx which will be running.

diox: First attempt in https://github.com/mozilla/addons-server/pull/22732/commits/ebf8616da061fb310628fce29ca1b97cd5b09274 works locally, but not in CI because of permissions: the `olympia` user is id 9500, but the host storage belongs to uid `1001`... I'm confused, this is using `run-docker`, shouldn't it be set correctly already ? How do unit test that write into storage even work ?

diox: We've got a uid=9500 because we're using `docker-compose.ci.yml`... but :bulb: unit tests are running with their own temporary media directory for the test so they don't care...

diox: Fixed in https://github.com/mozilla/addons-server/pull/22732/commits/92088e1cdd517a25d839398a65c8daefc2be2845

eviljeff (approved):
wfm :shipit:

PR #22734, opened 2024-10-07
Use Choices API for block_type choices + add by_block_type query set on BlockVersion

KevinMind:
Relates to: mozilla/addons#15014

### Description

add `block_type` property to get the display name of a block type. Use choices api to define the block type options. Add a block tyype queryset to easily grab different "kinds" of blocks.

### Testing

Look at the tests, you can recreate in djshell if you want.

### Checklist


diox:
Could you please make the title of that PR a bit more explicit ?

src/olympia/amo/utils.py
```diff
@@ -881,8 +881,13 @@ def _open(self, name, mode='rb'):
                     raise
         return super()._open(name, mode=mode)
 
-    def path(self, name):
-        return os.path.normpath(super().path(force_str(name)))
+    def path(self, *names):
```
diox, line 884: Do we need this ? I'm not sure we want to change the signature of a method that is shared by all Django storage classes (https://docs.djangoproject.com/en/5.1/ref/files/storage/#django.core.files.storage.Storage.path) Also this is not tested (we have `src/olympia/amo/tests/test_storage_utils.py`).

eviljeff: /agree - if you need this it should be a different(ly named) method

src/olympia/blocklist/models.py
```diff
@@ -155,14 +156,24 @@ def get_blocks_from_guids(cls, guids):
         return blocks
 
 
+class BlockType(Enum):
+    BLOCKED = 'blocked'
+    SOFT_BLOCKED = 'soft_blocked'
+
```
eviljeff, line 162: use `Choices` (see use elsewhere in the repo) to get some extra utility functions around this

eviljeff: may as well switch `soft` to be a small int too

src/olympia/blocklist/models.py
```diff
@@ -155,14 +156,24 @@ def get_blocks_from_guids(cls, guids):
...
-        blocktype = 'soft' if self.soft else 'hard'
-        return f'Block.id={self.block_id} ({blocktype}) -> Version.id={self.version_id}'
+        return (
+            f'Block.id={self.block_id} ({self.block_type}) '
+            f'-> Version.id={self.version_id}'
+        )
+
+    @property
+    def block_type(self):
+        return BlockType.SOFT_BLOCKED if self.soft else BlockType.BLOCKED
```
eviljeff, line 176: with a Choices object you can get the value from self.soft.display or similar, you don't need a `block_type`

eviljeff (changes requested):

src/olympia/amo/utils.py
KevinMind: Ah, looked in test_utils. It is a simple change and is fully backwards compatible. I'll add a test for multiple paths.

And to original question, yes we do.

src/olympia/blocklist/models.py
KevinMind: That was my original idea 😆 so we would rather have choices with int value? @diox if you agree I'll update in this PR.

src/olympia/amo/utils.py
KevinMind: @eviljeff can you explain why it would **need** to be a differently named method? Nothing about the current use of the method would change, and you would not need to consider should I use "path" or "path_join".

src/olympia/blocklist/models.py
eviljeff: any field is a choice field in django 😉- it's the use of Choices to declare the constants that's important.

src/olympia/amo/utils.py
eviljeff: it's a standard interface to the Storage class, so we wouldn't be able to swap it out as easily.

KevinMind: Fair enough. Removed. Added the test that was seemingly missing though for the original implementation.

src/olympia/blocklist/models.py
KevinMind: Kept Boolean field but added the BlockType as a choices.

eviljeff (changes requested):
wdyt?  (needs a `make format` too)

src/olympia/blocklist/models.py
```diff
@@ -154,15 +156,36 @@ def get_blocks_from_guids(cls, guids):
...
 
+class BlockVersionQuerySet(models.QuerySet):
+    def by_block_type(self, block_type):
+        # It's seemingly impossible to get a good type out of a Choices object
+        if block_type not in BlockVersion.BLOCK_TYPE_CHOICES:
+            raise ValueError(f'Invalid block type: {block_type}')
+
+        return self.filter(
+            soft=block_type, version__file__is_signed=True
+        )
```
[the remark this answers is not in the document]
KevinMind: Removed. Can add back if needed.

src/olympia/blocklist/migrations/0035_alter_blockversion_soft.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('blocklist', '0034_remove_blockversion_hard_blockversion_soft'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='blockversion',
+            name='soft',
+            field=models.BooleanField(choices=[(False, 'blocked'), (True, 'soft_blocked')], default=False),
```
diox, line 16: Regenerate the migration, it should be `field=models.BooleanField(choices=[(0, 'Blocked'), (1, 'Soft-Blocked')], default=False),`

diox (approved):

src/olympia/blocklist/migrations/0035_alter_blockversion_soft.py
KevinMind: Done.

PR #22738, opened 2024-10-08
Soft-block-tests

KevinMind:
Relates to: https://github.com/mozilla/addons/issues/15014

### Description

- Refactors the test_cron tests to match ([new code](https://github.com/mozilla/addons-server/pull/22718))) with minimal reverting to work with the existing code base (e.g existing MLBF/cron has no concept of a block_type)
- Refactors the existing code (minimally) to pass the new tests. The new tests were more aggressive and exposed a few edge cases that can be solved within the existing code base before adding the block_type functionality

### Context

This PR is an attempt to a) decouple https://github.com/mozilla/addons-server/pull/22718 by splitting the patch and b) ensure backwards compatibility with the existing codebase.

I would have ideally liked to change **literally** nothing ([first commit](https://github.com/mozilla/addons-server/pull/22738/commits/072bd011b1fe391b7a4b174405675a2a317f62b0)) but the tests, but as mentioned above, there are some bugs in the code that I've found through the new feature and it doesn't make sense to make the tests _not catch_ those bugs. Specifically:

- bloom filter fails if either `blocked` or `not_blocked` lists are empty. This is rare but found in testing and easy to cover by not adding error rates in those cases.
- the Store bloom filter does not check whether the stored files exist on initialization. This was techincally covered in the cron job, but should actually fail on initialization of the `MLBF` class as it is catastrophic
- the DataBase bloom filter does not store the json until you call either "write" stash or filter methods. This can lead to bugs where the data is queried by the class, but not stored, making it possible for subsequent Store filters to fail due to missing files. There is no reason not to store immediately. Additionally if you call both filter and stash methods, you would overwrite the files, which could lead to a filter with a given time stamp to have different data (especially since we can freely modify the underlying data store)
- remove the stash_json property and instead read the file where we need to (like we do with the filter) This guarantees the data comes from the stash file and not some arbitrary set of that property. Also fails if the file doesn't exist meaning we don't need to check in the cron job.

Small refactors that do not change underlying loogic:
- reverse the ordering of defining `base_filter` and `last_filter` (IMO this is just easier to reason about)
- encapsulate the timestamp getters to named methods (this is in fact mostly to make testing easier by mocking those methods)
- use existing `rel_location` and storage.path methods to derive paths instead of direct os.path methods
- rename self.id to self.created_at (that is what it is and it is important that we understand you should be able to do math with the value)

### Testing

Check the tests, this is the heart and soul of the PR. I've made the mocks for test_cron much more composable so it should be easier to understand what specific data is used in each test as well as what critical values cause the test to pass or fail.

You should be able to run the cron function or `./manage.py export_blocklist 1`

You should additionally be able to create and manipulate `MLBF` in the djshell.

### Checklist

### DIFF
--- a/src/olympia/blocklist/tests/test_cron.py
+++ b/src/olympia/blocklist/tests/test_cron.py
@@ -321,12 +321,12 @@ class TestTimeMethods(TestCase):
         assert isinstance(get_generation_time(), int)
 
     def test_get_last_generation_time(self):
-        assert get_last_generation_time() == None
+        assert get_last_generation_time() is None
         set_config(MLBF_TIME_CONFIG_KEY, 1)
         assert get_last_generation_time() == 1
 
     def test_get_base_generation_time(self):
-        assert get_base_generation_time() == None
+        assert get_base_generation_time() is None
         set_config(MLBF_BASE_ID_CONFIG_KEY, 1)
         assert get_base_generation_time() == 1
 
--- a/src/olympia/blocklist/tests/test_cron.py
+++ b/src/olympia/blocklist/tests/test_cron.py
@@ -24,7 +24,7 @@ from olympia.blocklist.cron import (
     process_blocklistsubmissions,
     upload_mlbf_to_remote_settings,
 )
-from olympia.blocklist.mlbf import MLBF, MLBFDataType
+from olympia.blocklist.mlbf import MLBF
 from olympia.blocklist.models import Block, BlocklistSubmission, BlockVersion
 from olympia.blocklist.utils import datetime_to_ts
 from olympia.constants.blocklist import MLBF_BASE_ID_CONFIG_KEY, MLBF_TIME_CONFIG_KEY
@@ -204,7 +204,6 @@ class TestUploadToRemoteSettings(TestCase):
             mock.call(
                 self.current_time,
                 is_base=force_base,
-                block_type_constant=MLBFDataType.BLOCKED.name,
             )
         ]
         mlbf = MLBF.load_from_storage(self.current_time)
@@ -223,7 +222,6 @@ class TestUploadToRemoteSettings(TestCase):
             mock.call(
                 self.current_time,
                 is_base=False,
-                block_type_constant=MLBFDataType.BLOCKED.name,
             )
         ]
         mlbf = MLBF.load_from_storage(self.current_time)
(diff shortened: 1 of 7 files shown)


src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -25,26 +27,43 @@
 
 
 class TestMLBF(TestCase):
+    def setUp(self):
+        self.user = user_factory()
+
+    def _blocked_addon(self, **kwargs):
```
KevinMind, line 33: This might seem odd, but in the next PR I will almost totally replace the existing tests and this is how I generate the data, so easier to just add it now as there are a few tests that were added in this PR. https://github.com/mozilla/addons-server/pull/22663

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -38,342 +38,298 @@
...
-            updated_by=user_factory(),
+        self.user = user_factory()
+        self.addon = addon_factory()
+        self.block = block_factory(guid=self.addon.guid, updated_by=self.user)
+
+        self.mocks: dict[str, mock.Mock] = {}
+        for mock_name in (
+            'olympia.blocklist.cron.statsd.incr',
+            'olympia.blocklist.cron.cleanup_old_files.delay',
+            'olympia.blocklist.cron.upload_filter.delay',
```
diox, line 49: I don't know about the rest but this specifically is a bad idea with the current setup, because `upload_filter` is not currently covered by dedicated tests - it was all covered by this class. So as it stands by mocking it, you are causing the entirety of `upload_filter` task to not be tested at all. Either that mock needs to go or you need to add additional tests for the task.

KevinMind: Done

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -38,342 +38,298 @@
...
-                mock.call(f'{STATSD_PREFIX}success'),
-            ]
+        assert (
+            mock.call(
+                'blocklist.cron.upload_mlbf_to_remote_settings.not_blocked_count', 1
+            )
+            in self.mocks['olympia.blocklist.cron.statsd.incr'].call_args_list
         )
 
-    @override_switch('blocklist_mlbf_submit', active=False)
```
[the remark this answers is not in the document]
KevinMind: Added.

diox (approved):

PR #22749, opened 2024-10-09
Allow URLs in developer replies to ratings

diox:
Fixes https://github.com/mozilla/addons/issues/15069

### DIFF
--- a/src/olympia/ratings/serializers.py
+++ b/src/olympia/ratings/serializers.py
@@ -184,7 +184,6 @@ class RatingSerializerReply(BaseRatingSerializer):
         allow_null=False,
         required=True,
         allow_blank=False,
-        validators=[NoURLsValidator()],
     )
 
     def to_representation(self, obj):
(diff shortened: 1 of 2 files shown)


KevinMind:
Can you add some testing steps?

KevinMind (approved):
Looks good. FWIW I could add `www.google.com` but not `https://www.google.com` Probably difficult to get lower level that protocol checking but wanted to shout.

KevinMind:
I was able to review with `www.google.com` but not `https://www.google.com`

Screenshot 2024-10-16 at 17 36 07

I could reply even with https...

Screenshot 2024-10-16 at 17 39 06

diox:
Yeah we're not trying to detect things that may look like an URL - we used to, when we were transforming them into links automatically, but we don't anymore so really the check is just to add an extra layer to keep our ratings as clean as possible.

when we were doing that detection it was very prone to errors. There are some many TLDs these days that it's too messy.

PR #22750, opened 2024-10-09
expose blocked version as blocked and soft_blocked in the API

eviljeff:
Fixes: mozilla/addons#15015

### Description

Renames `versions` to `blocked`, adds `soft_blocked`, and drops unused `max_version` and `min_version`.  Adds shims for v4 so functionality is as now.

### Context

Frontend consumes this API to display details of blocks, and in particular for this change, the versions that are blocked, to end-users.  Until deploy+1 after frontend is updated to consume `blocked` rather than `versions` we'll keep the shim `block-versions-list-shim` enabled for v5 too so it continues to work during the deploy, etc.  `min_version` and `max_version` have been unused for a long time on frontend - since we stopped implementing blocked versions as a range - but we overlooked removing and shim'ing them.

### Testing
- Add some blocked (signed) versions that are a mixture of hard and soft blocks, and see `/api/v5/blocked/<guid>/` lists them as `blocked` and `soft_blocked` respectively.
- See hard blocked are still there as `versions` as before
- check /v4/ and see `min_version` and `max_version` are still there, and are the max and min hard blocked versions.
- `is_all_versions` should be true if all signed versions are either hard or soft blocked AND the add-on is disabled to prevent further versions.

### Checklist

### DIFF
--- a/src/olympia/blocklist/serializers.py
+++ b/src/olympia/blocklist/serializers.py
@@ -3,6 +3,7 @@ from rest_framework import fields
 from olympia import amo
 from olympia.api.fields import OutgoingURLField, TranslationSerializerField
 from olympia.api.serializers import AMOModelSerializer
+from olympia.api.utils import is_gate_active
 from olympia.versions.models import Version
 
 from .models import Block
@@ -11,9 +12,8 @@ from .models import Block
 class BlockSerializer(AMOModelSerializer):
     addon_name = TranslationSerializerField(source='addon.name')
     url = OutgoingURLField()
-    min_version = fields.SerializerMethodField()
-    max_version = fields.SerializerMethodField()
-    versions = fields.SerializerMethodField()
+    blocked = fields.SerializerMethodField()
+    soft_blocked = fields.SerializerMethodField()
     is_all_versions = fields.SerializerMethodField()
 
     class Meta:
@@ -23,29 +23,32 @@ class BlockSerializer(AMOModelSerializer):
             'created',
             'modified',
             'addon_name',
+            'blocked',
+            'soft_blocked',
             'guid',
-            'min_version',
-            'max_version',
             'reason',
             'url',
-            'versions',
             'is_all_versions',
         )
 
-    def get_versions(self, obj):
+    def _get_blocked_for(self, obj, *, is_soft):
         if not hasattr(obj, '_blockversion_set_qs_values_list'):
             obj._blockversion_set_qs_values_list = sorted(
                 obj.blockversion_set.order_by('version__version').values_list(
-                    'version__version', flat=True
+                    'version__version', 'soft'
                 )
             )
-        return obj._blockversion_set_qs_values_list
+        return [
+            version
+            for version, soft in obj._blockversion_set_qs_values_list
+            if soft == is_soft
+        ]
 
-    def get_min_version(self, obj):
-        return versions[0] if (versions := self.get_versions(obj)) else ''
+    def get_blocked(self, obj):
+        return self._get_blocked_for(obj, is_soft=False)
 
-    def get_max_version(self, obj):
-        return versions[-1] if (versions := self.get_versions(obj)) else ''
+    def get_soft_blocked(self, obj):
+        return self._get_blocked_for(obj, is_soft=True)
 
     def get_is_all_versions(self, obj):
         cannot_upload_new_versions = not obj.addon or obj.addon.status in (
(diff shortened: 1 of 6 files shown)


KevinMind:
Testing instructions used the wrong URL. I actually found this very useful to test in the swagger UI as I can see the fields I expect and trigger that API to see the result.

The [URL](http://olympia.test/api/v5/swagger/#/blocklist/blocklist_block_read)

I have a setup with only one block, for an addon with a single signed version, here is the json

```json
{
  "id": 1,
  "created": "2024-10-16T17:17:03Z",
  "modified": "2024-10-16T17:17:03Z",
  "addon_name": {
    "en-US": "Addôn 8eaffd9aca3645399e98c316d3c78020"
  },
  "blocked": [
    "23609.517.8654.43211"
  ],
  "soft_blocked": [],
  "guid": "{0f1fbd83-d0fe-42ef-b6ce-365860d6ca10}",
  "reason": "",
  "url": null,
  "is_all_versions": false,
  "versions": [
    "23609.517.8654.43211"
  ]
}
```

I think I should be expecting 'is_all_versions': True on this one right?

Added another block version and still never getting is_all_versions to be true.

```json
{
  "id": 1,
  "created": "2024-10-16T17:17:03Z",
  "modified": "2024-10-16T17:17:03Z",
  "addon_name": {
    "en-US": "Addôn 8eaffd9aca3645399e98c316d3c78020"
  },
  "blocked": [
    "23609.517.8654.43211",
    "60110.12408.26118.60798"
  ],
  "soft_blocked": [],
  "guid": "{0f1fbd83-d0fe-42ef-b6ce-365860d6ca10}",
  "reason": "",
  "url": null,
  "is_all_versions": false,
  "versions": [
    "23609.517.8654.43211",
    "60110.12408.26118.60798"
  ]
}
```

What am I doing wrong?

I can see that updating the soft property appropriately moves the block to the right list.

KevinMind:
> AND the add-on is disabled to prevent further versions.

Missed that part.

KevinMind:
Got it after force disabling from reviewer tools. That should also be mentioned it must be force disabled not just a version disabled by the developer.

```json
{
  "id": 1,
  "created": "2024-10-16T17:17:03Z",
  "modified": "2024-10-16T17:17:03Z",
  "addon_name": {
    "en-US": "Addôn 8eaffd9aca3645399e98c316d3c78020"
  },
  "blocked": [
    "23609.517.8654.43211",
    "60110.12408.26118.60798"
  ],
  "soft_blocked": [],
  "guid": "{0f1fbd83-d0fe-42ef-b6ce-365860d6ca10}",
  "reason": "",
  "url": null,
  "is_all_versions": true,
  "versions": [
    "23609.517.8654.43211",
    "60110.12408.26118.60798"
  ]
}
```

src/olympia/blocklist/serializers.py
```diff
@@ -23,29 +23,32 @@ class Meta:
...
             'url',
-            'versions',
             'is_all_versions',
         )
 
-    def get_versions(self, obj):
+    def _get_blocked_for(self, obj, *, is_soft):
         if not hasattr(obj, '_blockversion_set_qs_values_list'):
             obj._blockversion_set_qs_values_list = sorted(
                 obj.blockversion_set.order_by('version__version').values_list(
```
KevinMind, line 37: Could you not add soft to the query here? I'm not totally sure if that's valid in django but 

```python
obj.blockversion_set.filter(soft=is_soft)...
```

doesn't work? That would be better than filtering after the fact from a query perf perspective and much cleaner to look at.

KevinMind (approved):
One nit. Looks good. 🚢

src/olympia/blocklist/tests/test_serializers.py
```diff
@@ -8,6 +12,7 @@
 
 class TestBlockSerializer(TestCase):
```
KevinMind, line 13: It's not clear to me from the diff but make sure there is a test covering the scenario where all are blocked and addon is NOT disabled.

eviljeff:
> Testing instructions used the wrong URL. I actually found this very useful to test in the swagger UI as I can see the fields I expect and trigger that API to see the result.

The [api docs](https://mozilla.github.io/addons-server/topics/api/blocklist.html) are correct I just wrote the testing instructions from memory so got the url slightly wrong

src/olympia/blocklist/tests/test_serializers.py
eviljeff: The test for that scenario was missing in the existing tests, looks like.  I'll add it for completeness.

PR #22755, opened 2024-10-11
Add recaptcha for new addon submission upload

KevinMind:
Fixes: mozilla/addons#15076

### Description

Adds recaptcha for new addon submissions, behind a waffle switch.

### Context

In order to combat spam uploads, we will enable the google recaptcha field on all new addon upload submissions.

This should NOT lead to a recaptcha being seen on each submission, but will include the field to decide if and when to display.

### Testing

You can enable seeing the recaptcha by flipping `developer-submit-addon-captcha` on

1. navigate to the new submission [page](http://olympia.test/en-US/developers/addon/submit/distribution)
2. choose "listed" or "unlisted" (You should test both)
3. click continue
4. Expect to see the captcha above the continue button.
5. upload a valid xpi
6. If you do not submit the recaptcha, expect the page to reload with an error that the field is required, otherwise you will continue.

Note: if the waffle flag is "disabled" expect 4 to not display the captcha and thus 6 will always let you continue (assuming other fields are valid)

### Checklist

### DIFF
--- a/src/olympia/devhub/forms.py
+++ b/src/olympia/devhub/forms.py
@@ -1056,11 +1056,17 @@ class NewUploadForm(CheckThrottlesFormMixin, forms.Form):
         'Please try again after some time.'
     )
     throttle_classes = addon_submission_throttles
+    recaptcha = ReCaptchaField(label='')
 
     def __init__(self, *args, **kw):
         self.request = kw.pop('request')
         self.addon = kw.pop('addon', None)
+        self.include_recaptcha = kw.pop('include_recaptcha', False)
         super().__init__(*args, **kw)
+
+        recaptcha_enabled = waffle.switch_is_active('developer-submit-addon-captcha')
+        if not recaptcha_enabled or not self.include_recaptcha:
+            del self.fields['recaptcha']
 
         # Preselect compatible apps based on the current version
         if self.addon and self.addon.current_version:
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -1486,7 +1486,13 @@ WIZARD_COLOR_FIELDS = [
 
 @transaction.atomic
 def _submit_upload(
-    request, addon, channel, next_view, wizard=False, theme_specific=False
+    request,
+    addon,
+    channel,
+    next_view,
+    wizard=False,
+    theme_specific=False,
+    include_recaptcha=False,
 ):
     """If this is a new addon upload `addon` will be None.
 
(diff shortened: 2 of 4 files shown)


src/olympia/devhub/views.py
```diff
@@ -1611,7 +1621,9 @@ def submit_addon_upload(request, channel):
     if not RestrictionChecker(request=request).is_submission_allowed():
         return redirect('devhub.submit.agreement')
     channel_id = amo.CHANNEL_CHOICES_LOOKUP[channel]
-    return _submit_upload(request, None, channel_id, 'devhub.submit.source')
+    return _submit_upload(
+        request, None, channel_id, 'devhub.submit.source', include_recaptcha=True
```
diox, line 1625: Note that themes have their own specific function (`submit_theme_upload()`, which passes `theme_specific=True` to bypass 2FA requirement) and you're not including the captcha there. Not sure if that was intentional (might need to check with Operations/Product what they want). In any case it would be good to cover it with tests.

diox (approved):
Missing a test around the captcha not being included (even if the waffle switch is active) for new versions submissions.

diox:
(Please update the PR title before merging too)

KevinMind:
@diox I'm not sure I understand this scenario?

> Missing a test around the captcha not being included (even if the waffle switch is active)

If the switch is active, we include the field... so the test is testing where the captcha is not included.. how would that happen?

diox:
"for new versions submissions." (not new add-ons). That's a different view, `submit_version_upload()`, where `include_recaptcha` is not passed.

src/olympia/devhub/views.py
KevinMind: Discussed and decided not to include themese initially

src/olympia/devhub/forms.py
```diff
@@ -1056,12 +1056,18 @@ class NewUploadForm(CheckThrottlesFormMixin, forms.Form):
...
+    recaptcha = ReCaptchaField(label='')
 
     def __init__(self, *args, **kw):
         self.request = kw.pop('request')
         self.addon = kw.pop('addon', None)
+        self.include_recaptcha = kw.pop('include_recaptcha', False)
         super().__init__(*args, **kw)
 
+        recaptcha_enabled = waffle.switch_is_active('developer-submit-addon-captcha')
+        if not recaptcha_enabled or not self.include_recaptcha:
```
diox, line 1068: For your consideration, I don't necessarily think this is better, but we could do the waffle check in the view, just passing `include_recaptcha=waffle.switch_is_active('developer-submit-addon-captcha')` ?

PR #22759, opened 2024-10-14
Implement APIKey as django form + recaptcha field

KevinMind:
Fixes: mozilla/addons#15083

### Description

This PR refactors the `APIKeyPage` view to utilize a django form, both to support using the recaptcha field as part of the flow, and to make the form logic more consolidated and easier to test/grok.

### Context

Previously, the logic for managing the API key page was kind of murky and spread/duplicated between the view and the template layer.

Now, there is a Django form `APIKeyForm` that is managing the state machine of which actions a developer has available, what is the current state of the relevant entities (confirmation + credentials) and which fields + validations should be in scope.

This is a dynamic form that can model the full state of the API key lifecycle from requesting a confirmation, confirming + generating, revoking, and regenerating credentials.

The underlying logic is largely the same as before, but to support the re-captcha a few key changes have been made.

1) the re-captcha field is now included in the "confirm" stage, when a user does not have a confirmation or credentials.

2) if a user has credentials but no confirmation, they are shown the credentials view with ONLY the revoke action. Previously we handled creating the confirmation to allow them to regenerate but now, we don't know if they have re-captcha complete and it is simpler to remove that action and force the developer to reenter the flow from the beginning.

3) We use GET parameters to prefill the confirmation code, but the API is exposed as a post endpoint so it is (unlikely but) possible to submit both a POST and GET token. this is explicitly handled to prefer the POST over the get and covered with a test.

### Testing

I'm going to document the entire flow in the issue since there are really a lot of edge cases with this form.

### Checklist

### DIFF
--- a/src/olympia/devhub/forms.py
+++ b/src/olympia/devhub/forms.py
@@ -1,6 +1,7 @@
 import os
 import tarfile
 import zipfile
+from functools import cached_property
 from urllib.parse import urlsplit
 
 from django import forms
@@ -17,6 +18,7 @@ from django.utils.translation import gettext, gettext_lazy as _, ngettext
 
 import waffle
 from django_statsd.clients import statsd
+from extended_choices import Choices
 
 from olympia import amo
 from olympia.access import acl
@@ -42,6 +44,7 @@ from olympia.amo.forms import AMOModelForm
 from olympia.amo.messages import DoubleSafe
 from olympia.amo.utils import slug_validator, verify_no_urls
 from olympia.amo.validators import OneOrMoreLetterOrNumberCharacterValidator
+from olympia.api.models import APIKey, APIKeyConfirmation
 from olympia.api.throttling import CheckThrottlesFormMixin, addon_submission_throttles
 from olympia.applications.models import AppVersion
 from olympia.constants.categories import CATEGORIES, CATEGORIES_BY_ID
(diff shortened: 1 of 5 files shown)


src/olympia/devhub/views.py
```diff
@@ -2016,66 +2015,89 @@ def api_key(request):
...
+            {
+                'label': _('Revoke'),
+                'action': form.ACTION_CHOICES.revoke,
+            },
+            {
+                'label': _('Revoke and regenerate credentials'),
+                'action': form.ACTION_CHOICES.regenerate,
+            },
+        ]
+    elif form.confirmation is not None:
```
KevinMind, line 2075: The initial value could be internalized to the form along with the help text.

src/olympia/devhub/views.py
```diff
@@ -2016,66 +2015,89 @@ def api_key(request):
...
+        if form.action == form.ACTION_CHOICES.confirm:
+            print('BAAAAAAm')
+            form.send_confirmation()
 
-        # In any case, redirect after POST.
         return redirect(reverse('devhub.api_key'))
 
+    submit_actions = []
+
+    if form.credentials is not None:
```
KevinMind, line 2058: This could still be in the template.. i have mixed feelings

src/olympia/devhub/views.py
```diff
@@ -2016,66 +2015,89 @@ def api_key(request):
...
+            )
+            send_key_revoked_email(request.user.email, form.credentials.key)
 
-    if request.method == 'POST':
-        has_confirmed_or_is_confirming = confirmation and (
-            confirmation.confirmed_once
-            or confirmation.is_token_valid(request.POST.get('confirmation_token'))
-        )
+            if form.action == form.ACTION_CHOICES.revoke:
+                form.confirmation.delete()
```
KevinMind, line 2034: this is the biggest departure from the previous flow.. but I think this is a good additional feature. If you explicitly revoke your own keys we should probably double check you're human and the account owner... otherwise you end up on a screen waiting for a confirmation token that might be really old and hard to find.

src/olympia/devhub/forms.py
```diff
@@ -1424,3 +1427,154 @@ def clean(self):
...
+            get_token_param = self.request.GET.get('token')
+
+            if (
+                self.confirmation.confirmed_once
+                or get_token_param is not None
+                or self.data.get('confirmation_token') is not None
+            ):
+                help_text = _(
+                    'Please click the confirm button below to generate '
+                    'API credentials for user <strong>{user}</strong>.'
```
diox, line 1515: Ideally keep the existing `name` parameter to avoid changes to the string so that translators don't have to re-translate it

src/olympia/devhub/forms.py
```diff
@@ -1424,3 +1427,154 @@ def clean(self):
...
+                initial=get_token_param,
+                required=False,
+                help_text=help_text,
+                validators=[self.validate_confirmation_token],
+            )
+
+        else:
+            if waffle.switch_is_active('developer-submit-addon-captcha'):
+                self.fields['recaptcha'] = ReCaptchaField(
+                    label='', help_text=_("You don't have any API credentials.")
```
diox, line 1537: By making it a `help_text` on the recaptcha field that text is no longer displayed if the captcha isn't enabled... I guess this is fine because it's not a very useful text anyway ?

diox (approved):

src/olympia/devhub/forms.py
KevinMind: Done.

PR #22765, opened 2024-10-14
Allow adding, viewing information about and deleting soft-blocks in the admin

diox:
Fixes https://github.com/mozilla/addons/issues/15013

### Description

Like other blocklist operations, everything goes through a `BlocklistSubmission`, which gains a new `block_type` field to determine what kind of block to apply for this particular submission. Users are not allowed to modify `Block`s directly.

Note: If a version is already soft-blocked or hard-blocked, that version won't be selectable when creating a new `BlocklistSubmission`, even of another type. However, the code underneath should handle overriding the block type for a given blocked version in a submission, and future actions will be added to soften or harden blocks.

### Testing

Play around with the admin and add/delete `BlocklistSubmission`, and then inspect the blocks until you're satisfied. Few things you can try to mix together:
- Make soft or hard blocks
- Make soft *and* hard blocks on add-ons with multiple versions (requires separate submissions)
- `BlocklistSubmission` for add-ons with `average_daily_users` above `settings.DUAL_SIGNOFF_AVERAGE_DAILY_USERS_THRESHOLD` are subject to approval as before. You can check that it still behaves as expected. You shouldn't be able to modify a block type for a submission you're approving (arbitrary decision on my part, we could change this later)
- Make soft and/or hard blocks for multiple guids at the same time, for add-ons containing multiple versions
- Etc


src/olympia/blocklist/admin.py
```diff
@@ -407,7 +409,6 @@ def add_view(self, request, **kwargs):
             'errors': admin.helpers.AdminErrorList(form, []),
             'preserved_filters': self.get_preserved_filters(request),
             # extra context we use in our custom template
-            'is_delete': is_delete,
```
diox, line 410: Was not actually used (`blocks.html` widget has its own separate context)

KevinMind:
> Note: If a version is already soft-blocked or hard-blocked, that version won't be selectable when creating a new BlocklistSubmission, even of another type. However, the code underneath should handle overriding the block type for a given blocked version in a submission, and future actions will be added to soften or harden blocks.

What is the rationale for this? If the logic to determine the underlying block type is internalized, why does blocksubmission need a block type at all?

src/olympia/blocklist/models.py
```diff
@@ -260,6 +259,10 @@ class BlocklistSubmission(ModelBase):
         help_text='The submission will not be published into blocks before this time.',
     )
     disable_addon = models.BooleanField(default=True)
+    block_type = models.IntegerField(
```
KevinMind, line 262: I'm not a huge fan of this. In BlockVersion the field is a boolean called soft, and in BlockSubmissionList it is an integerField. Also, since BLOCK_TYPE_CHOICES is shared between these models (and will be used in MLBF eventually) I think we should extract this to a higher level, maybe module level on the models.py or in utils.

Maybe we could create a field for this to use in both models?

Also, I'm still not totally sure why we need the block_type on the submission level if I can submit multiple versions in a submission and there is a real use case of submitting different block types for versions of the same addon.

Why isn't a BlockSubmissionList creating Blocks -> BlockVersions ? I guess because we sometimes delay or await signoff etc etc. Still feels like an overly complicated approach opposed to a boolean field on the block "active" or something.. thinking out loud here but this doesn't feel right.

diox: We need `BlocklistSubmission` to handle multi-guids submissions, which is a frequent use-case for operations as when they find some malware, they often find more than one add-on connected to the same entity, and want to block them all in one go (or unblock if the decision needs to be reverted).

Since we need the delay/approval/disable add-on/notes/etc bits for those multi-guids block operations and also single-guid ones, for consistency and to avoid having the logic in several places, the admin was set up in such a way to make everything go through `BlocklistSubmission`. I'm simply following the logic here.

diox: Regarding integer vs boolean: calling it `block_type` here allowed me to directly expose it to the admin as a field admins can tweak and easily understand but I agree it can be confusing to us looking at the code...

If I move the `BLOCK_TYPE_CHOICES` to `src/olympia/constants/blocklist.py` with a comment explaining how it's shared, is that enough for you?

src/olympia/blocklist/utils.py
```diff
@@ -194,19 +201,27 @@ def save_versions_to_blocks(guids, submission):
...
-                block_versions_to_create.append(block_version)
+            if version.id in submission.changed_version_ids:
+                if version.is_blocked:
+                    block_version = version.blockversion
+                    block_versions_to_update.append(block_version)
+                else:
+                    block_version = BlockVersion(block=block, version=version)
+                    block_versions_to_create.append(block_version)
+                block_version.soft = is_soft
                 version.blockversion = block_version
```
eviljeff, line 215: nit: you can move this into the `else` - in the `if ` we just set `block_version` to `version.blockversion`

src/olympia/blocklist/utils.py
```diff
@@ -36,18 +38,23 @@ def block_activity_log_save(
...
+    version_details = {}
     if submission_obj:
         details['signoff_state'] = submission_obj.SIGNOFF_STATES.get(
             submission_obj.signoff_state
         )
         if submission_obj.signoff_by:
             details['signoff_by'] = submission_obj.signoff_by.id
+        details['soft'] = version_details['soft'] = (
+            submission_obj.block_type == BlockVersion.BLOCK_TYPE_CHOICES.SOFT_BLOCKED
+        )
```
eviljeff, line 50: imo we need to change the content of `details` significantly - it's useful to know the `soft`/`hard` state of the submission, but `blocked_versions` and `added_versions` don't mean what they used to mean, in the world of soft blocks.

My suggestion would be more like
'hard_blocked_versions' : ...
'soft_blocked_versions': ...
'added_versions' ...
'changed_versions' ...

src/olympia/blocklist/models.py
eviljeff: yeah, pushing everything through `BlocklistSubmission` is necessary for a single submission that may be affecting multiple guid block(s), version(s), and other block details, but are gated by sign-off or delays.  It wouldn't _impossible_ to change everything around to do things differently, but it's a non-trivial amount of refactoring.

src/olympia/blocklist/templates/admin/blocklist/widgets/blocks.html
```diff
@@ -29,11 +29,10 @@ <h3>{{ blocks|length|intcomma }} Add-on GUIDs with {{ total_adu|intcomma }} user
...
               value="{{ version.id }}"
               {% if version.id in widget.value %}checked{% endif %}
-            > {% if is_delete %}Unblock{% else %}Block{% endif %} {{ version.version }}</label>
+            > {% if is_add_change %}Block{% else %}Unblock{% endif %} {{ version.version }} {% if version.blockversion %}({{ version.blockversion.get_soft_display }}){% endif %}</label>
           {% else %}
-            <span title="{% if version.is_blocked %}Blocked{% else %}Not blocked{% endif %}">
-              <!-- Red Hexagonal stop sign for Blocked; Green cirle for not blocked -->
-              {% if version.is_blocked %}&#x1F6D1;{% else %}&#x1F7E2;{% endif %}{{ version.version }}
+            <span>
+              {{ version.version }} ({% if version.is_blocked %}{{ version.blockversion.get_soft_display }}{% else %}&#x1F7E2; Not Blocked{% endif %})
```
eviljeff, line 35: Can you add a screenshot to the PR showing this with a list of versions?  imo showing it this way is going to make it harder to visually scan through a list of versions than previously.

I.e. it's currently (using `#` and `*` in place of green and red symbols):
```
# 12.5
# 13.2.1.567
* 14.3
# 15
* 16.5399834.0
```
and this change makes it:
```
12.5 (# Not Blocked)
13.2.1.567 (# Not Blocked)
14.3 (* Hard-Block)
15 (# Not Blocked)
16.5399834.0 (^ Soft-Block)
```

eviljeff (commented):
> If a version is already soft-blocked or hard-blocked, that version won't be selectable when creating a new BlocklistSubmission, even of another type. However, the code underneath should handle overriding the block type for a given blocked version in a submission, and future actions will be added to soften or harden blocks.

Do you know how this would work?  Isn't there a concern that to implement_that_ feature will require changes to what you're landing here?

src/olympia/blocklist/migrations/0036_blocklistsubmission_block_type_and_more.py
```diff
@@ -0,0 +1,23 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='blocklistsubmission',
+            name='block_type',
+            field=models.IntegerField(choices=[(0, 'Hard-Blocked'), (1, 'Soft-Blocked')], default=0),
+        ),
+        migrations.AlterField(
+            model_name='blockversion',
+            name='soft',
+            field=models.BooleanField(choices=[(0, 'Hard-Blocked'), (1, 'Soft-Blocked')], default=False),
```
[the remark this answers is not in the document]
diox: It is yeah, I'll fix that

src/olympia/blocklist/templates/admin/blocklist/widgets/blocks.html
diox: Here is a deletion of a complex case (I've manually increased the height of the list in devtools to display everything):
Screenshot 2024-10-24 at 18-07-47 Delete Blocks Django site admin

Here is an addition with the same add-on in the same state (again manually increased the height of the list to display everything):
Screenshot 2024-10-24 at 18-10-25 Block Add-ons Django site admin


IMHO it's necessary to do _something_ to properly visualize such cases.

diox:
> > If a version is already soft-blocked or hard-blocked, that version won't be selectable when creating a new BlocklistSubmission, even of another type. However, the code underneath should handle overriding the block type for a given blocked version in a submission, and future actions will be added to soften or harden blocks.
> 
> Do you know how this would work? Isn't there a concern that to implement _that_ feature will require changes to what you're landing here?

I'm not entirely sure how hardening/softening actions will work 100%, but the rough plan I've played with is separate explicit actions that allow you to select only relevant blocked versions, similar to how the add/delete actions work. Some tweaks will be required but allowing overrides will help (since ultimately soften/harden are overrides on top of an existing blockversion), even if it's not currently exposed in the admin in this PR.

eviljeff:
> > > If a version is already soft-blocked or hard-blocked, that version won't be selectable when creating a new BlocklistSubmission, even of another type. However, the code underneath should handle overriding the block type for a given blocked version in a submission, and future actions will be added to soften or harden blocks.
> > 
> > 
> > Do you know how this would work? Isn't there a concern that to implement _that_ feature will require changes to what you're landing here?
> 
> I'm not entirely sure how hardening/softening actions will work 100%, but the rough plan I've played with is separate explicit actions that allow you to select only relevant blocked versions, similar to how the add/delete actions work. Some tweaks will be required but allowing overrides will help (since ultimately soften/harden are overrides on top of an existing blockversion), even if it's not currently exposed in the admin in this PR.

The code appears to support Operations making changes to an existing version's hard/soft status, while blocking new versions (so a mix of add and change), but I can't see how the UI in this patch could allow that.  If you've got a plan to get there, then that's fine.

src/olympia/blocklist/templates/admin/blocklist/widgets/blocks.html
eviljeff: If @wagnerand is okay with that UX then fine

diox:
Yeah, no UI supporting softening/hardening is in this PR. It's only the small backend bit I've changed that supports the future softening/hardening actions.

src/olympia/blocklist/models.py
diox: Filed https://github.com/mozilla/addons/issues/15111

src/olympia/blocklist/utils.py
diox: I'm not sure how useful those keys are (I know `changed_versions` is used in the history we show) ... Do you think this is enough:

```diff
diff --git a/src/olympia/blocklist/utils.py b/src/olympia/blocklist/utils.py
index 896e5274f1..1502f58a2d 100644
--- a/src/olympia/blocklist/utils.py
+++ b/src/olympia/blocklist/utils.py
@@ -10,6 +10,32 @@ from olympia.versions.models import Version
 log = olympia.core.logger.getLogger('z.amo.blocklist')
 
 
+def blocked_versions_activity_details(obj, *, deleted=False):
+    hard_blocked_versions = (
+        []
+        if deleted
+        else sorted(
+            ver.version
+            for ver in obj.addon_versions
+            if ver.is_blocked and not ver.blockversion.soft
+        )
+    )
+    soft_blocked_versions = (
+        []
+        if deleted
+        else sorted(
+            ver.version
+            for ver in obj.addon_versions
+            if ver.is_blocked and ver.blockversion.soft
+        )
+    )
+
+    return {
+        'hard_blocked_versions': hard_blocked_versions,
+        'soft_blocked_versions': soft_blocked_versions,
+    }
+
+
 def block_activity_log_save(
     obj,
     change,
@@ -19,24 +45,25 @@ def block_activity_log_save(
 
     action = amo.LOG.BLOCKLIST_BLOCK_EDITED if change else amo.LOG.BLOCKLIST_BLOCK_ADDED
     addon_versions = {ver.id: ver.version for ver in obj.addon_versions}
-    blocked_versions = sorted(
-        ver.version for ver in obj.addon_versions if ver.is_blocked
-    )
+    blocked_versions_details = blocked_versions_activity_details(obj)
     changed_version_ids = (
         [v_id for v_id in submission_obj.changed_version_ids if v_id in addon_versions]
         if submission_obj
         else sorted(ver.id for ver in obj.addon_versions if ver.is_blocked)
     )
     changed_versions = sorted(addon_versions[ver_id] for ver_id in changed_version_ids)
+    total_versions_blocked = len(
+        blocked_versions_details['hard_blocked_versions']
+    ) + len(blocked_versions_details['soft_blocked_versions'])
 
     details = {
         'guid': obj.guid,
-        'blocked_versions': blocked_versions,
         'added_versions': changed_versions,
         'url': obj.url,
         'reason': obj.reason,
         'comments': f'{len(changed_versions)} versions added to block; '
-        f'{len(blocked_versions)} total versions now blocked.',
+        f'{total_versions_blocked} total versions now blocked.',
+        **blocked_versions_details,
     }
     version_details = {}
     if submission_obj:
@@ -72,11 +99,7 @@ def block_activity_log_save(
 def block_activity_log_delete(obj, deleted, *, submission_obj=None, delete_user=None):
     assert submission_obj or delete_user
     addon_versions = {ver.id: ver.version for ver in obj.addon_versions}
-    blocked_versions = (
-        sorted(ver.version for ver in obj.addon_versions if ver.is_blocked)
-        if not deleted
-        else []
-    )
+    blocked_versions_details = blocked_versions_activity_details(obj, deleted=deleted)
     changed_version_ids = (
         [v_id for v_id in submission_obj.changed_version_ids if v_id in addon_versions]
         if submission_obj
@@ -87,15 +110,18 @@ def block_activity_log_delete(obj, deleted, *, submission_obj=None, delete_user=
         for ver_id in changed_version_ids
         if ver_id in addon_versions
     )
+    total_versions_blocked = len(
+        blocked_versions_details['hard_blocked_versions']
+    ) + len(blocked_versions_details['soft_blocked_versions'])
 
     details = {
         'guid': obj.guid,
-        'blocked_versions': blocked_versions,
         'removed_versions': changed_versions,
         'url': obj.url,
         'reason': obj.reason,
         'comments': f'{len(changed_versions)} versions removed from block; '
-        f'{len(blocked_versions)} total versions now blocked.',
+        f'{total_versions_blocked} total versions now blocked.',
+        **blocked_versions_details,
     }
     action = (
         amo.LOG.BLOCKLIST_BLOCK_EDITED

```

eviljeff: The main use case is how the logs are used, yeah.... - in logs.html it looks like we just expose `added_versions` and `removed_versions` - so beyond that there probably isn't much use in adding anything, unless we want to change what's exposed in logs.html too...

src/olympia/blocklist/templates/admin/blocklist/widgets/blocks.html
wagnerand: This looks ok to me overall.

Minor suggestion: For the blocked versions, can we rename "Soft-Block" to "Soft-Blocked" (and same for hard), to clarify that this is a state, not an action.

Also, when adding a block, where would we indicate whether it is a soft-block or hard-block? (The screenshot only says "Block".) Is this on the page that isn't part of the screenshot?

wagnerand:
> I've manually increased the height of the list in devtools to display everything

@diox would you mind adding that to the PR. The box is too small for many cases.

diox:
> > I've manually increased the height of the list in devtools to display everything
> 
> @diox would you mind adding that to the PR? The box is too small for many cases.

Done in https://github.com/mozilla/addons-server/pull/22765/commits/8cc5c34ace1b289b2d9726c3d947322e04ed51c7

src/olympia/blocklist/templates/admin/blocklist/widgets/blocks.html
wagnerand: > This is also why btw I've renamed from "Soft-Blocked" to "Soft-Block", since it refers to a type of block, I figured this would work better in all situations. I'm happy to change it back to "Soft-Blocked" if you prefer though.

That makes sense. My comment was about modifying/deleting a block (not adding a new one, since there are no blocked versions). Here, we should distinguish between action (block) and state (blocked). Does that clarify things?

diox: Yes, but I was hoping to re-use the string as it was complicated enough as it is...

diox: Going back to the different actions (again those are the most complex cases, on purpose, showing all possible states), I think it should already be clear in all cases:

### Adding a block

Screenshot 2024-10-29 at 11-43-02 Block Add-ons Django site admin

You can see this add-on already has a soft and a hard blocked version, with the block type displayed in parenthesis. Since it's an addition, those don't get the checkbox for those. We only use the word "Block" next to the checkboxes, since the type of block is selected for the whole submission by the dropdown below.

(I've cheated and removed the block history in this screenshot as this block has been modified over and over and over and it was taking a bunch of space)

### Approving a pending block

Screenshot 2024-10-29 at 11-54-38 Pending Sign-off {105ca783-0522-4143-b8ed-062cba7028a9} Change blocklist submission Django site admin

Same add-on after I submitted one new hard blocked version that needs to be approved. The block type is no longer modifiable. Since this submission only contained a single version to be blocked, that's the only checkbox shown, and again only the word "Block" shows up for it, as the type of block would be determined by the block type for this submission.

### Deleting a block

Screenshot 2024-10-24 at 18-07-47 Delete Blocks Django site admin

This time, there is no global "Block type" for this submission since it's a deletion. We only show the checkboxes next to versions that are soft or hard blocked already. We show the word "Unblock" for those, and again the current block type for each version in parenthesis.

diox: I found a way to do what @wagnerand asked for in 4391116d2e3d6a73550798639c1429220c79cd3d.

So every time we display the state of a blocked version it will say "Hard/Soft-Blocked", but for the block type of a submission it will say "Hard/Soft Block".

src/olympia/blocklist/models.py
diox: 4391116d2e3d6a73550798639c1429220c79cd3d introduces a different Choices... that shares the same name and re-uses the constants/values from `BlockVersion`. The true fix will be in https://github.com/mozilla/addons/issues/15111 which will make `BlockVersion` use `block_type` instead of the boolean (but IMHO out of scope here)

src/olympia/blocklist/utils.py
diox: https://github.com/mozilla/addons-server/pull/22765/commits/ea79c09deb87c725761bdda3db0f7c82bf516e78

src/olympia/blocklist/migrations/0037_alter_blockversion_soft.py
eviljeff, line 1: Nit: you could squish this into one migration file.

eviljeff (approved):

src/olympia/blocklist/migrations/0037_alter_blockversion_soft.py
diox: https://github.com/mozilla/addons-server/pull/22765/commits/3123a66116e9ece02e1031a4e548b8fc40848f53

PR #22766, opened 2024-10-15
Hold actions on the back of decisions, rather than process them, if content is high profile

eviljeff:
Fixes: mozilla/addons#15074

### Description

This patch changes decision handling from Cinder, to check if the decision/content is high profile, and if so to not carry out the action - just log the action that should have happened.

### Context

This is the first part of larger epic - further work will expose the held actions and allow them to be carried out (or not).  No decisions are coming from Cinder at this time so we're pretty safe to land it half finished ... though it could be argued it should be behind a waffle switch :shrug: 

### Testing

Bit of a pain to test fully, because it relies on having your cinder auth enabled for the report, and manually applying webhook payloads. But roughly:
- find a promoted (recommended, line, spotlight, or notable) add-on locally
  - note: the add-on must be actively promoted (i.e. has at least one version approved for that Promoted class)
- report that add-on to cinder under a category that is handled in cinder (e.g. legal reason) 
- in cinder stage, make a decision to disable on that add-on and copy the payload json from the admin section
- apply the payload via curl, etc (easier if you comment out `@permission_classes((CinderInboundPermission,))`!)
- go to reviewer review page for that add-on and see add-on is not disabled, but that the disable was attempted _is_ displayed in the version history

### Checklist


KevinMind:
>  though it could be argued it should be behind a waffle switch 🤷

My very first thought. I think it is pretty low effort to add. Could make it default to true via migration.

src/olympia/abuse/migrations/0041_alter_decision_date.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 4.2.16 on 2024-10-14 13:48
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
KevinMind, line 6: Would be nice to have a reversible migration here. It's not clear what the field "used" to be and if you ever wanted to undo the migration this would be useful.. possible?

src/olympia/abuse/utils.py
```diff
@@ -239,12 +262,30 @@ class CinderActionBanUser(CinderAction):
     reporter_template_path = 'abuse/emails/reporter_takedown_user.txt'
     reporter_appeal_template_path = 'abuse/emails/reporter_appeal_takedown.txt'
 
+    def should_hold_action(self):
```
KevinMind, line 265: Is it not possibe to make this a method on the model of "target"? I'm probably way late to the party on this, but it feels like there's so much really complex logic in these utility classes.

src/olympia/abuse/utils.py
```diff
@@ -51,11 +51,34 @@ def __init__(self, decision):
...
+            *(self.decision.policies.all()),
+            *extra_args,
+            details={
+                'comments': self.decision.notes,
+                'cinder_action': self.decision.action,
+                **(extra_details or {}),
+            },
+        )
+
+    def should_hold_action(self):
```
KevinMind, line 67: the should_hold_action and hold_action are just empty methods that should be overridden by child classes right? Why not raise a not implemented error?

src/olympia/abuse/migrations/0041_alter_decision_date.py
eviljeff: built-in django migrations are reversible by default.

KevinMind: TIL.

src/olympia/abuse/utils.py
eviljeff: It is action dependant though - an approve decision on a UserProfile wouldn't be restricted, only a ban user.  The possible models of target are Addon, UserProfile, Rating, and Collection - I'm not really convinced it'd be an improvement to split up the logic about what is high profile among classes in 4 different files - especially in models like Addon that already have a lot of code in them.

I do agree they aren't really "utility" classes - I started a branch that renamed the file to actions, and dropped the "Cinder" from them, because they're intended to be the source of truth for review/moderation actions eventually - currently most of _that_ logic is spread among random functions in reviewers/utils.py instead.  (I never ended up merging the branch because it would always create too many conflicts with whatever patch I was working on at the time!)

KevinMind: I think at a certain point it starts to be a model on it's own. But not for this PR 🤷

eviljeff:
> > though it could be argued it should be behind a waffle switch 🤷
> 
> My very first thought. I think it is pretty low effort to add. Could make it default to true via migration.

Slack thread resolved to not do this because ERR_TOO_MANY_DSA_WAFFLE_SWITCHES

KevinMind (approved):
Code looks good (from what I can tell) 🚢  did not verify but agreed over slack it is easier to verify on dev

src/olympia/constants/activity.py
```diff
@@ -1059,6 +1059,39 @@ class DENY_APPEAL_JOB(_LOG):
...
+    id = 194
+    action_class = 'review'
+    format = _('Review {rating} for {addon} delete held for further review.')
+    reviewer_format = 'Held {user_responsible}s delete {rating} for {addon}'
+    admin_event = True
+
+
+class HELD_ACTION_COLLECTION_DELETED(_LOG):
+    id = 195
+    format = _('Collection {collection} deletion held for further review')
```
flodolo, line 1081: I'm guessing this should have ended with a period like the other messages?

eviljeff: We're not super consistent (`HELD_ACTION_FORCE_DISABLE` below, and the original `FORCE_DISABLE` are both missing `.` from the end of their format messages.).

Though as all of these are admin only events that shouldn't be exposed to end-users, they don't need to be localized anyway - they'll be a follow-up that drops them.

flodolo: > shouldn't be exposed to end-users, they don't need to be localized anyway - they'll be a follow-up that drops them.

These are being actively translated (see [latest commits](https://github.com/mozilla/addons-server/commits/master/)). If that's the intent, we should try not to expose them, vs dropping them later.

PR #22773, opened 2024-10-16
Tweak file upload limits: bump max uncompressed size, round down max upload size

diox:
Fixes: mozilla/addons#15093


KevinMind (approved):
I wish we could synchronize this with production somehow but alas.

PR #22779, opened 2024-10-17
Modify Button Behaviour When enable-submissions Flag is Disabled

chrstinalin:
Fixes: mozilla/addons#15090

### Description
when `enable-submissions` is disabled, now:
1. Disables the 'create a theme' button 
2. Addon submission buttons are disabled and a warning is displayed banner style with the flag note (if any).

image

image

### Testing
1. Deactivate the `enable-submissions` waffle flag (everyone should be = `True` by default).
2. User should be unable to make new submissions and warned via the disabled button.

### Checklist

### DIFF
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -1614,7 +1614,7 @@ def _submit_upload(
             'wizard_url': wizard_url,
             'max_upload_size': settings.MAX_UPLOAD_SIZE,
             'submissions_enabled': flag.is_active(request),
-            'reason': flag.note if hasattr(flag, 'note') else None
+            'reason': flag.note if hasattr(flag, 'note') else None,
         },
     )
 
--- a/src/olympia/devhub/views.py
+++ b/src/olympia/devhub/views.py
@@ -1574,7 +1574,19 @@ def _submit_upload(
         if existing_properties
         else []
     )
-    submit_notification_warning = get_config('submit_notification_warning')
+
+    flag = waffle.get_waffle_flag_model().get('enable-submissions')
+    warning = (
+        'Add-on uploads are temporarily unavailable' + ': ' + flag.note
+        if hasattr(flag, 'note')
+        else '.'
+    )
+
+    submit_notification_warning = (
+        warning
+        if not flag.is_active(request)
+        else get_config('submit_notification_warning')
+    )
     if not submit_notification_warning and addon:
         # If we're not showing the generic submit notification warning, show
         # one specific to pre review if the developer would be affected because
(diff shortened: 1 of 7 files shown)


KevinMind:
@chrstinalin could you reassign review. I'll be on PTO next week and don't want to block you.

eviljeff (commented):
I might be missing some context here, but why are we doing the disabling, and rendering the error, in javascript?  We know submissions are disabled when the template is rendered so add the logic to disable the buttons and display the message in the template?

We can test the error message is visible and the button disabled in tests too that way

chrstinalin:
I honestly can't recall if I had a reason for this, I think it may have been the natural progression of my original barebones fix... I do like that solution better, though

eviljeff (commented):
Some nits in the test code, and it works fine.  I want to get a second opinion on the UI though - exposing the reason in a tooltip you have to mouse-over to seee feels too hidden to me.  ~Can you include a screenshot of what it looks like for reference?~ (you've already included one)

diox:
I'm no UX/Product person but I think it should be a bit more obvious. My idea was to prevent submissions while at the same time still showing a page that looks somewhat normal, because the error page you otherwise get from the decorator is a bit too generic/empty (it's good as a fallback for anyone trying to bypass the more user-friendly error).

chrstinalin:
@diox My initial fear with directly putting the reason on the page is if the reason is longer than usual (which may not happen) and possibly look weird. 

Though IIRC there's somewhere in DevHub that use a banner-style information section, which might work better? I can't seem to find it, though.

diox:
Search for `site-notice` (it displays the content of the `zadmin.models.Config` named `site_notice` which can be set in the admin)

src/olympia/devhub/views.py
```diff
@@ -1574,7 +1574,17 @@ def _submit_upload(
         if existing_properties
         else []
     )
-    submit_notification_warning = get_config('submit_notification_warning')
+    flag = waffle.get_waffle_flag_model().get('enable-submissions')
+    warning = (
+        'Add-on uploads are temporarily unavailable' + ': ' + flag.note
+        if hasattr(flag, 'note')
+        else '.'
+    )
```
eviljeff, line 1582: Can you add a test to verify what's rendered when there is no note set, but the flag is enabled?  I missed that case before.

src/olympia/devhub/views.py
```diff
@@ -1574,7 +1574,17 @@ def _submit_upload(
         if existing_properties
         else []
     )
-    submit_notification_warning = get_config('submit_notification_warning')
+    flag = waffle.get_waffle_flag_model().get('enable-submissions')
+    warning = (
+        'Add-on uploads are temporarily unavailable' + ': ' + flag.note
```
eviljeff, line 1579: we need a `gettext` here too, or the warning won't be localised.

eviljeff (changes requested):

eviljeff (approved):

PR #22784, opened 2024-10-21
Make growth threshold before flagging relative instead of absolute

diox:
`growth_threshold_before_flagging` for a given `UsageTier` is now relative to the average growth of all non-disabled add-ons in that usage tier.

Fixes mozilla/addons#15049

### Testing

Play around with the unit test or locally. As you can see in the patch `UsageTier` has `lower_adu_threshold` and `upper_adu_threshold` fields governing which add-ons it targets, and then `growth_threshold_before_flagging` is now relative to the average `hotness` of all extensions in that tier. `growth_threshold_before_flagging` is still a percentage though, and needs to be converted back to a float when comparing.

So, a simplistic case is to have `Addon(average_daily_users=102, hotness=0.3)`, `Addon(average_daily_users=102, hotness=0.2)`, `UsageTier(lower_adu_threshold=100, upper_adu_threshold=1000, growth_threshold_before_flagging=1)`: Before the changes, this would have flagged both add-ons for having their `hotness` over `0.01` (`1/100`). Now, it should only flag the first one for being over `0.26` (`(0.3+0.2)/2 + 1/100`).

Flagging will result in an `active` `NeedsHumanReview` instance on the add-on latest versions in both channels with `REASONS.HOTNESS_THRESHOLD` reason.

### DIFF
--- a/src/olympia/addons/tasks.py
+++ b/src/olympia/addons/tasks.py
@@ -4,7 +4,7 @@ import os
 
 from django.core.files.storage import default_storage as storage
 from django.db import transaction
-from django.db.models import Q, Value
+from django.db.models import Avg, Q, Value
 from django.db.models.functions import Collate
 
 from elasticsearch_dsl import Search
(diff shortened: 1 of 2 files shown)


eviljeff (approved):
lgtm but can we have another test - essentially one that proves that an add-on (or add-ons) with the exact same growth are flagged under scenario  A (where the others in their group have one average hotness), and aren't flagged under scenario B (where the others in their group have a different average hotness)

PR #22789, opened 2024-10-24
Optimize number of queries made by devhub version list view & display block status

diox:
Fixes mozilla/addons#15037
Fixes mozilla/addons#15017

### Description

This replaces "Disabled by Mozilla" by "Blocked"/"Restricted" in devhub versions list, and optimizes queries made by that view as a drive-by (I was already in the neighborhood and was changing the test anyway...)

### Testing

Unit test, or play with blocks on versions that you have locally. Note that technically one can force a block without having the version disabled, but I figured this was an edge case not worth worrying about, that saves us a couple database queries that way because we are re-using the macro for various versions that are fetched without fetching the blockversion as well.

### DIFF
--- a/src/olympia/blocklist/models.py
+++ b/src/olympia/blocklist/models.py
@@ -15,6 +15,7 @@ from olympia.addons.models import Addon
 from olympia.amo.models import BaseQuerySet, ManagerBase, ModelBase
 from olympia.amo.templatetags.jinja_helpers import absolutify
 from olympia.amo.utils import chunked
+from olympia.constants.blocklist import BLOCK_TYPE_END_USERS_CHOICES
 from olympia.users.models import UserProfile
 from olympia.versions.models import Version
 
@@ -170,6 +171,10 @@ class BlockVersion(ModelBase):
             f'Block.id={self.block_id} ({self.get_soft_display()}) '
             f'-> Version.id={self.version_id}'
         )
+
+    def get_user_facing_block_type_display(self):
+        """Like get_soft_display(), but using strings meant for end-users."""
+        return BLOCK_TYPE_END_USERS_CHOICES.for_value(int(self.soft)).display
 
 
 class BlocklistSubmissionQuerySet(BaseQuerySet):
--- a/src/olympia/constants/blocklist.py
+++ b/src/olympia/constants/blocklist.py
@@ -1,3 +1,8 @@
+from django.utils.translation import gettext_lazy as _
+
+from extended_choices import Choices
+
+
 # How many guids should there be in the stashes before we make a new base.
 BASE_REPLACE_THRESHOLD = 5_000
 
@@ -6,3 +11,10 @@ MLBF_TIME_CONFIG_KEY = 'blocklist_mlbf_generation_time'
 MLBF_BASE_ID_CONFIG_KEY = 'blocklist_mlbf_base_id'
 
 REMOTE_SETTINGS_COLLECTION_MLBF = 'addons-bloomfilters'
+
+
+# As BlockVersion.BLOCK_TYPE_CHOICES except meant to be exposed to end-users.
+BLOCK_TYPE_END_USERS_CHOICES = Choices(
+    ('BLOCKED', 0, _('Blocked')),
+    ('SOFT_BLOCKED', 1, _('Restricted')),
+)
--- a/src/olympia/devhub/templates/devhub/versions/list.html
+++ b/src/olympia/devhub/templates/devhub/versions/list.html
@@ -46,7 +46,17 @@
     </td>
     <td class="file-status">
       <div>
-        {{ distro_tag(version.channel)}} {{ version.file.get_status_display() }}
+        {{ distro_tag(version.channel)}}
+        <span class="file-status-text">
+          {% if version.file.status == amo.STATUS_DISABLED and version.is_blocked %}
+            {# Slight optimization: don't try getting blocked status for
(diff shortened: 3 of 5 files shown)


src/olympia/devhub/templates/devhub/versions/list.html
```diff
@@ -87,7 +97,7 @@
          data-is-current="{{ (version == addon.current_version)|int }}">x</a>
     </td>
   </tr>
-  {% if full_info and version.channel == amo.CHANNEL_LISTED and (can_request_review or can_cancel) and check_addon_ownership(request.user, addon, allow_developer=True) %}
```
diox, line 90: The view is already decorated by `dev_required` which calls `check_addon_ownership(request.user, addon, allow_developer=True, allow_mozilla_disabled_addon)`. `can_cancel` and `can_request_review` will never be `True` if the add-on is disabled, so this extra check is not needed.

src/olympia/devhub/templates/devhub/versions/list.html
```diff
@@ -212,7 +221,7 @@ <h3>{{ _('All versions') }}</h3>
   <div class="item" id="version-list"
        data-stats="{{ url('devhub.versions.stats', addon.slug) }}">
     <div class="item_wrapper">
-      {% if check_addon_ownership(request.user, addon, allow_developer=True) and addon.status != amo.STATUS_DISABLED %}{# i.e. Admin disabled #}
+      {% if can_submit %}
```
diox, line 224: As above, since the view is already decorated with the appropriate check and `can_submit` already means the add-on is not disabled, we can simplify this.

src/olympia/devhub/templates/devhub/versions/list.html
```diff
@@ -46,7 +46,17 @@
...
-        {{ distro_tag(version.channel)}} {{ version.file.get_status_display() }}
+        {{ distro_tag(version.channel)}}
+        <span class="file-status-text">
+          {% if version.file.status == amo.STATUS_DISABLED and version.is_blocked %}
+            {# Slight optimization: don't try getting blocked status for
+               non-disabled versions - this avoids checking if for latest/current
+               versions, which are fetched without a select_related('blockversion') #}
+            {{ version.blockversion.get_user_facing_block_type_display() }}
+          {% else %}
+            {{ version.file.get_status_display() }}
```
eviljeff, line 57: I see we did this already, but I wonder why we weren't using `version.get_review_status_display` here?

eviljeff (approved):

src/olympia/blocklist/models.py
```diff
@@ -171,6 +172,10 @@ def __str__(self) -> str:
             f'-> Version.id={self.version_id}'
         )
 
+    def get_user_facing_block_type_display(self):
+        """Like get_soft_display(), but using strings meant for end-users."""
+        return BLOCK_TYPE_END_USERS_CHOICES.for_value(int(self.soft)).display
```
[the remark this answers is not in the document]
diox: It's tricky to override as it dynamically added by some metaclass stuff. You can't do this for instance:
```
    def get_soft_display(self, end_user_facing=False):
        if not end_user_facing:
            return super().get_soft_display()
        return BLOCK_TYPE_END_USERS_CHOICES.for_value(int(self.soft)).display
```

You'll get `AttributeError: 'super' object has no attribute 'get_soft_display'` when `end_user_facing` is `False`.

src/olympia/devhub/templates/devhub/versions/list.html
diox: Good question. I think `get_review_status_display` was meant for reviewers only. It can return `Disabled by Developer` which is technically more correct that what we do now (atm in devhub if the developer disables the version, we say "Disabled by Mozilla") but at the same time a bit weird to display to a developer maybe ?

src/olympia/constants/blocklist.py
```diff
@@ -6,3 +11,10 @@
 MLBF_BASE_ID_CONFIG_KEY = 'blocklist_mlbf_base_id'
 
 REMOTE_SETTINGS_COLLECTION_MLBF = 'addons-bloomfilters'
+
+
+# As BlockVersion.BLOCK_TYPE_CHOICES except meant to be exposed to end-users.
+BLOCK_TYPE_END_USERS_CHOICES = Choices(
+    ('BLOCKED', 0, _('Blocked')),
+    ('SOFT_BLOCKED', 1, _('Restricted')),
+)
```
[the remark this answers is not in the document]
diox: I agree. As much as I don't like switching the field to an integer, we could do it, that would solve the most confusing part. But I don't see a way to solve the strings, since UX want us to say "Restricted" to end-users, but internally "Restricted" is such a loaded term...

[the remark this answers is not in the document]
diox: https://github.com/mozilla/addons/issues/15111

src/olympia/devhub/templates/devhub/versions/list.html
diox: We have a thread with UX about the various statuses and what we show to developers, lets wait to see how that shakes out as it will probably require more changes/its own issue.

src/olympia/constants/blocklist.py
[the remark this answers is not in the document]
eviljeff: Re: the label we expose to internal vs end users, we _could_ use the constant name (i.e. SOFT_BLOCKED) but I'm not sure that's a great solution either 🤷

PR #22795, opened 2024-10-25
Respond to cinder moving a job into the amo handled queue

eviljeff:
Fixes: mozilla/addons#15035

### Description

Allowing monitoring of job.action webhook - if a job is moved into the AMO reviewer handled queue, treat it as a forward.

### Context

It's a small amount of extra code, but unlocks removing a lot of complexity in our Cinder related models (and somewhat in Cinder too). Currently, with AMO_ESCALATE, a decision is made to escalate, which closes the job.  We then have to a) create a new job in Cinder, and b) link up the new and old jobs in a our models via forwarded_job.  We can remove all that.

### Testing

As always we webhook integration, it's involved...
- Report an extension to be handled in Cinder (e.g. other reason).
- In Cinder, find the job in [AMO Dev Listings](https://stage.cinder.nonprod.webservices.mozgcp.net/queues/UXVldWVUeXBlOjM5/details/jobs?filters=%7B%7D) on cinder stage
- Escalate that job (under More Actions) to "AMO Dev Add-on Infringement"
- (The Escalate policies have been removed from that queue)
- see the job is now in the "[AMO Dev Add-on Infringement](https://stage.cinder.nonprod.webservices.mozgcp.net/queues/UXVldWVUeXBlOjI1/details/jobs?filters=%7B%7D)" queue in Cinder
- go into [webhooks in admin](https://stage.cinder.nonprod.webservices.mozgcp.net/settings/webhooks) and copy the job action webhook request; replay it via curl, etc, into your local AMO.  (as usual, probably easier if you comment out the auth)
- The job should now be escalated in AMO - i.e. it's got a NHR which causes it to show up in the manual review queue; and the job is shown as available to be resolved on the review page.

### Checklist

### DIFF
--- a/src/olympia/abuse/views.py
+++ b/src/olympia/abuse/views.py
@@ -261,9 +261,13 @@ def cinder_webhook(request):
     set_user(UserProfile.objects.get(pk=settings.TASK_USER_ID))
 
     try:
-        if not (payload := request.data.get('payload', {})):
-            log.info('No payload received: %s', str(request.data)[:255])
-            raise ValidationError('No payload')
+        if (
+            not (payload := request.data.get('payload', {}))
+            or not isinstance(payload, dict)
+            or not payload
+        ):
+            log.info('No payload dict received: %s', str(request.data)[:255])
+            raise ValidationError('No payload dict')
         match event := request.data.get('event'):
             case 'decision.created':
                 process_webhook_payload_decision(payload)
--- a/src/olympia/abuse/views.py
+++ b/src/olympia/abuse/views.py
@@ -32,6 +32,7 @@ from olympia.core import set_user
 from olympia.ratings.views import RatingViewSet
 from olympia.users.models import UserProfile
 
+from .cinder import CinderAddon
 from .forms import AbuseAppealEmailForm, AbuseAppealForm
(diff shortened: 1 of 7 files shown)


src/olympia/abuse/views.py
```diff
@@ -191,59 +261,17 @@ def cinder_webhook(request):
     set_user(UserProfile.objects.get(pk=settings.TASK_USER_ID))
 
     try:
-        if request.data.get('event') != 'decision.created':
-            log.info('Non-decision payload received: %s', str(request.data)[:255])
-            raise ValidationError('Not a decision')
-
         if not (payload := request.data.get('payload', {})):
```
diox, line 264: Could you add a check if payload is not a dict ? I got a 500 locally because I was calling the webhook with multipart data instead of json, and somehow that caused `payload` to be present, but as a string.... It wasn't happening before because the `request.data.get('event') != 'decision.created':` condition was causing us to skip the entire thing early.

diox (approved):

src/olympia/abuse/views.py
eviljeff: https://github.com/mozilla/addons-server/pull/22795/commits/61f4c6747a818a78e0affbc4b37f87a40e9b7f7f

PR #22806, opened 2024-10-30
Relax parsing of IP networks coming from the database

diox:
Even if technically the bits corresponding to the netmask should be set to 0, in practice it's fine to accept that they are not, it will result in the same network.

Fixes https://github.com/mozilla/addons/issues/15116

### DIFF
--- a/src/olympia/amo/fields.py
+++ b/src/olympia/amo/fields.py
@@ -120,7 +120,11 @@ class CIDRField(models.Field):
         value = value.strip()
 
         try:
-            return ipaddress.ip_network(value)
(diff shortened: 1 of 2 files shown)


eviljeff (approved):
Can we have a follow-up issue to fix this at the source - i.e. either not writing invalid ipv6 network addresses to the database, or parsing them as ipv6 network addresses?

r+ once CI passes

diox:
I'll file a follow-up yeah. Note that we already have validation preventing this (that I'm not changing with this patch), but automation was bypassing that.

diox:
Follow-up: https://github.com/mozilla/addons/issues/15123

PR #22811, opened 2024-10-31
Migrate from BlockVersion.soft to BlockVersion.block_type

bakulf:
Fixes: mozilla/addons#15111

### Description

This patch does the following:

- Introduce `PositiveTinyIntegerField` to support `tinyint unsigned` fields
- Rename `soft` to `block_type` via 2 separate migrations (add + remove)
- Introduce the `BlockType` enum and use it instead of the `soft` boolean
- Unify the string versions of enum values

### Context

The first `BlockVersion.soft` implementation used a boolean field type. For readability and future improvement, this PR replaces it with an enum and uses a `tinyint` data type, which is the same size as a boolean in MySQL. 

### Testing

No new tests are added, but all the existing tests are updated to use the new enum.

### Checklist


src/olympia/blocklist/migrations/0038_remove_blockversion_soft.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 4.2.16 on 2024-10-31 13:26
+
+from django.db import migrations
```
KevinMind, line 3: I wonder if this should be on a secondary PR as we probably want to land/run the first one and then separately the second one.

src/olympia/blocklist/models.py
```diff
@@ -157,25 +157,24 @@ def get_blocks_from_guids(cls, guids):
         return blocks
 
 
+class BlockType(models.IntegerChoices):
+    BLOCKED = 0, _('Blocked')
```
KevinMind, line 161: I know @diox had mentioned we might want to keep the emojified versions of the strings. Maybe @wagnerand also has an opinion. Is it possible we can use only one set of these, or it really must be two totally separate strings?

src/olympia/blocklist/migrations/0038_remove_blockversion_soft.py
KevinMind: Crazy idea, I've already rebased https://github.com/mozilla/addons-server/pull/22715 onto this patch.

I could cut out the part removing the soft field and we could land the first part in 2715. Then this patch could include the migration backfilling any rows from soft -> block_type and dropping the column.. just an idea.

src/olympia/blocklist/migrations/0036_blocklistsubmission_block_type_and_more.py
```diff
@@ -13,11 +13,11 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='blocklistsubmission',
             name='block_type',
-            field=models.IntegerField(choices=[(0, '🛑 Hard-Block'), (1, '⚠️ Soft-Block')], default=0),
+            field=models.IntegerField(choices=[(0, 'Blocked'), (1, 'Restricted')], default=0),
```
KevinMind, line 16: I don't think we should edit an existing migration.

src/olympia/blocklist/migrations/0038_remove_blockversion_soft.py
```diff
@@ -0,0 +1,27 @@
...
+import olympia.amo.fields
+
+
+def convert_block_type_to_soft(apps, schema_editor):
```
KevinMind, line 7: Wouldn't we want the opposite? to convert any remaining rows with soft populated and block_type NOT populated to use block_type, before dropping the column? What's the point of populating soft if we are just going to drop it?

bakulf: because if you see line 22, this code runs when rolling back this migration.

src/olympia/blocklist/migrations/0038_remove_blockversion_soft.py
bakulf: Yes, this is the idea behind having these 2 migrations.

* 0037 introduces the new `block_type` field. The code  still uses the existing `soft` field.
* 0038 drops the `soft` field because the new code uses the new `block_type`.

In theory we could have a race condition for `BlockVersion` records created between 0037 and 0038, but Mat said that we don't use this feature yet.

src/olympia/blocklist/migrations/0038_remove_blockversion_soft.py
KevinMind: Then we should also include another method to backfill soft to block_type no?

src/olympia/blocklist/migrations/0038_remove_blockversion_soft.py
KevinMind: The issue I see with this is that we have the migrations split, but the potential error condition is still there if the new code is deployed after 38 is executed.

We would want to switch from using the old code to the new code after 37 and before 38 and I'm not totally sure if or how we would do that in production environments. TBF I don't fully understand exactly how we run migrations anyhow so maybe that is trivial.

If we were to run both migrations before switching the code then it's still possible to have blocks created referencing the soft field after it's been dropped. That was the whole point of splitting it off.

**T0**: execute 37 adding the new field (still old code)
T1: new block created (uses soft field that is still there)
**T2**: deploy new code
T3: new block created (uses block type)
**T4**: execute 38 (would convert the block in T1 to block_type and drop the column)
T5: new block created (uses block_type)

If T2 goes before T0 you get errors at T1
If T2 goes after T5, then T4 will error because we are still expecting soft field to be in the db.

TLDR; I cannot understand how this currently solves the problem, unless we can finely control the execution along with the deployment (which maybe we can) does that make sense?

> This is also one of the reasons why I suggested we could split the PR to unblock the code change and 37 as that is relatively straight forward, then after we can figure out how to remove the soft field as a lower priority task.

If we are saying none of this matters, just make it one migration and let the errors happen if they do... but that could mean blocks in production not being created and idk what kind of retry/fallback mechanisms we have in that case.

src/olympia/blocklist/migrations/0038_remove_blockversion_soft.py
KevinMind: For context T4 in [this timeline](https://github.com/mozilla/addons-server/pull/22811#discussion_r1825753311) is why I think we need that back fill.

src/olympia/blocklist/migrations/0038_remove_blockversion_soft.py
bakulf: I hope "T2 before T0" cannot happen. If it does, there is a huge bug in the django deployment architecture. Same as "T2 goes after T5". Note that if this is the case, basically none of our migration is safe.

https://docs.djangoproject.com/en/5.1/topics/migrations/ 
"Once the migration is applied, commit the migration and the models change to your version control system as a single commit - that way, when other developers (or your production servers) check out the code, they’ll get both the changes to your models and the accompanying migration at the same time."

KevinMind: If there is one thing I've learned about addons-server is that things are way more subtle and definitely not always working how the latest django docs describe. addons-server is actually older than the concept of automatic migrations in django...

We'd have to ask @bqbn about more details of when and how we run migrations but I think it's entirely feasible that we would run the migrations before or after deploying the code but what is very suspicious would be running one migration before, and another after.. It's not obvious that the timeline described would just happen so in order to make it happen, we would have to have some level of control over the execution.

I do know we don't always run all migrations automatically (maybe none, I'm really not sure) but we should just be mindful of this. There are several ways this patch could break production if we are not careful and splitting 38 to a separate PR almost singularly removes that possibility.

diox: When running migrations, there is a (short) period of time where new migrations have all been ran but we're still on the old code from the previous tag, while the new pods are started. We don't have built-in capability to execute a specific migration before and another one after a deploy.

That means 2 things for migrations:
- A migration that adds a column needs to add it as `null=True`, or insertions will fail during the push window: old code doesn't know about that column, it doesn't have a database default (only `null=True` ones do in Django 4.x - in the future, Django 5.x, which we are not using yet, will allow us to have access to `db_default` to work around this). Sometimes this is acceptable, depending on how likely it is to have such insertions performed during the push.
- A migration that alters the type of a column requires making sure the old code won't try to insert incompatible data. Again, sometimes that risk is acceptable depending on how likely that is to happen.
- A migration that removes or renames a column requires making sure the existing code doesn't reference it at all: even if we're not actively using it, by default if a field is in a model, Django ORM will include it in every query for that model, even `SELECT`s. Here too this is sometimes acceptable for models not used in APIs or pages we expose to end-users, though it's a lot less likely.

- `Block` and `BlockVersion` are used in devhub, reviewers tools, admin tools & cron jobs, so we have to be backwards-compatible with old code at all times.
- `BlocklistSubmission` is used in reviewers tools, admin tools & cron jobs, so we could be less strict with it if needed

src/olympia/blocklist/models.py
diox: Yes, we explicitly didn't use "Restricted" in admin tools because we already use that term internally for other admin-related features (see all the `*Restriction` models in `users/models.py`). The emojis were also added to visually differentiate between soft and hard blocks and you'll need @wagnerand to weigh in if you want to change this (with screenshots of the admin tools for various scenarios after that change)

src/olympia/blocklist/migrations/0038_remove_blockversion_soft.py
KevinMind: So short answer, we need the first migration to set the new field nullable and the second migration should be in a separate PR so it can ship separately.

diox: Yes. The safest path is:

This PR should:
- Have the "new" field in the model (with `null=True`, but don't use `None` as a value anywhere)
- Have the migration to add the corresponding "new" column
- Remove the existing "old" field from the model
- Have a migration making the corresponding "old" column nullable (It should **not** be a migration removing the column)

A PR to be deployed in the following deploy, next sprint, should:
- Have a migration removing the "old" column
- Have the "new" field in the model no longer nullable
- Have a corresponding migration making the "new" column no longer nullable

KevinMind: Second PR should also backfill soft to block_type right?

src/olympia/blocklist/migrations/0037_alter_blockversion_block_type_and_more.py
```diff
@@ -0,0 +1,24 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='blockversion',
+            name='block_type',
+            field=olympia.amo.fields.PositiveTinyIntegerField(choices=[(0, 'Blocked'), (1, 'Restricted')], default=0, null=True),
+        ),
+        migrations.AlterField(
+            model_name='blocklistsubmission',
+            name='block_type',
+            field=olympia.amo.fields.PositiveTinyIntegerField(choices=[(0, 'Blocked'), (1, 'Restricted')], default=0, null=True),
```
KevinMind, line 22: Tiny deviation but I don't think we need to make the blocklistsubmission block_type field nullable. It wasn't already nullable and we are not adding it so it should probably remain required on DB level since we also provide a default value. or?

https://github.com/mozilla/addons-server/blob/master/src/olympia/blocklist/migrations/0036_blocklistsubmission_block_type_and_more.py#L13-L17

KevinMind (approved):
LGTM 🚀 (run make lint and fix the long line issue)

PR #22812, opened 2024-10-31
Filter reused guid:version strings from not blocked items in mlbf

KevinMind:
Fixes: mozilla/addons#15132

### Description

Ensure filtering of duplicate guid:version strings from not_blocked_items

### Context

After merging [commit](https://github.com/mozilla/addons-server/commit/97f3dbdbc230cd2ff346b521aa571852fd056458#diff-9d78bf2fa1a0d5d8446a4506e00278a05b8713e6bac8d430913150be293f43b5L240) it became clear in stage that we have duplicate hashed version strings in the mlbf data set.

### Testing

unit tested to cover the edge case

### Checklist

### DIFF
--- a/src/olympia/blocklist/tests/test_mlbf.py
+++ b/src/olympia/blocklist/tests/test_mlbf.py
@@ -1,6 +1,8 @@
 import json
 from functools import cached_property
 
+from olympia import amo
+from olympia.addons.models import GUID_REUSE_FORMAT
 from olympia.amo.tests import (
     TestCase,
     addon_factory,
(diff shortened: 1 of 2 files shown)


src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -351,3 +353,30 @@ def test_generate_filter_not_raises_if_all_versions_blocked(self):
...
+        reused_addon.update(guid=GUID_REUSE_FORMAT.format(addon.id))
+        reused_addon.addonguid.update(guid=addon.guid)
+
+        self._block_version(block, self._version(addon), soft=False)
+
+        mlbf = MLBF.generate_from_db('test')
+
+        (block_version,) = MLBF.hash_filter_inputs([(addon.guid, version)])
+        assert block_version in mlbf.data.blocked_items
+        assert block_version not in mlbf.data.not_blocked_items
```
eviljeff, line 382: being paranoid: can we assert that `(addon.guid, version)` is in `fetch_all_versions_from_db(MLBF._version_excludes)` ?  i.e. would have been included otherwise?

eviljeff (approved):
Something for the next time we touch this code (soon!) - we should have a check that there is no common items between the two sets and raise with an explicit error if it happens - we spent way too long wondering if there was some change in `filter_cascade` dependencies because the error we got from there was an obscure one about error levels, and not "Don't have duplicates!"

src/olympia/blocklist/tests/test_mlbf.py
KevinMind: I'd rather not test the underlying function but the final data set. Or am I missing something in your ask?

KevinMind:
> Something for the next time we touch this code (soon!) - we should have a check that there is no common items between the two sets and raise with an explicit error if it happens - we spent way too long wondering if there was some change in `filter_cascade` dependencies because the error we got from there was an obscure one about error levels, and not "Don't have duplicates!"

It's still not entirely clear if this bug is the cause of the error we saw.. I agree with you though we could have a more explicit check that no entry is in multiple arrays.. not totally clear when or where that check should be though.

src/olympia/blocklist/tests/test_mlbf.py
eviljeff: I was looking for a verification the test is set up correctly - that we have a case where there is a guid, version pair that is unblocked, and a pair that is blocked, but they don't both end up in final data set.  I suppose closer to a functional test than an integration test.

eviljeff:
> > Something for the next time we touch this code (soon!) - we should have a check that there is no common items between the two sets and raise with an explicit error if it happens - we spent way too long wondering if there was some change in `filter_cascade` dependencies because the error we got from there was an obscure one about error levels, and not "Don't have duplicates!"
> 
> It's still not entirely clear if this bug is the cause of the error we saw.. 

If I'm right we'll know soon 😄 

> I agree with you though we could have a more explicit check that no entry is in multiple arrays.. not totally clear when or where that check should be though.

Probably just before `cascade.initialize` errors in an obscure way, really.  If it's gotten that far we've got a bug elsewhere we can't (shouldn't try to) recover from, but at least the error would be comprehensible in sentry.

PR #22813, opened 2024-10-31
Show Pastebin Attachment if File/Input Already Uploaded

chrstinalin:
Fixes: mozilla/addons#15105

### Description

Makes it clearer that a file or text is still attached to an action and prevent defaulting to the initial menu.

### Testing

1. Enable waffle flag `enable-activity-log-attachments`
2. Upload a file or input text as attachment from reviewer actions
3. Refreshes that contain the cached input will continue to show that input. Hard refreshes will reset.

### Checklist

### DIFF
--- a/static/js/zamboni/reviewers.js
+++ b/static/js/zamboni/reviewers.js
@@ -308,18 +308,18 @@ function initExtraReviewActions() {
     $('#attachment-type-toggle').addClass('hidden');
     $('#attachment_input_wrapper').addClass('hidden');
     $('#attachment_file_wrapper').removeClass('hidden');
-  }
+  };
   const showInputWrapper = (e) => {
     e?.preventDefault();
     $('#attachment-type-toggle').addClass('hidden');
     $('#attachment_file_wrapper').addClass('hidden');
     $('#attachment_input_wrapper').removeClass('hidden');
-  }
+  };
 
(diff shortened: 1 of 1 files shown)


static/js/zamboni/reviewers.js
```diff
@@ -303,19 +303,22 @@ function initExtraReviewActions() {
...
-    $('#attachment_file_wrapper').addClass('hidden');
+  };
+  const showInputWrapper = (e) => {
+    e?.preventDefault();
+    $('#attachment-type-toggle, #attachment_file_wrapper').addClass('hidden');
     $('#attachment_input_wrapper').removeClass('hidden');
-  });
+  };
+
+  $('#id_attachment_file').prop('files').length && showFileWrapper();
```
KevinMind, line 317: My jquery days are long gone but my understanding is this code will execute one time on page load, or when that element is in the dom? It's not totally clear since this is no explicit event trigger. Is it possible to do that somehow or maybe add a comment indicating when this code is/should be executed?

chrstinalin: Once on page load, yeah -- `initExtraReviewActions()` is called on document ready. 

I could add a comment, but it feels like something that would really be relevant to the entire function rather than a comment singling out this particular line within it?

KevinMind (approved):
Verified working with a question 🚢

PR #22817, opened 2024-11-04
Record version rejections triggered by a block as a human review

diox:
The user that last updated the block should be the user set on the corresponding activity.

Fixes: mozilla/addons#15104

### Context

We want `human_review_date` to be set on versions we reject as part of a block, because that makes us consider the person that made the block as having performed a review of the version, which allows reviewers to unreject that version (on the listed channel, reviewers can only unreject versions that a human reviewer rejected in the first place). It's also more consistent in general - the person who made the block should have performed a review of the affected versions, so we should record that.

### Testing

Block a listed version from the admin: it should set a `human_review_date` on the `Version`, and the reject activity log should be coming from the user that recorded the block, not the task user.

This can then allow unrejecting the version in reviewer tools (assuming the add-on is re-enabled first if it was disabled).

### DIFF
--- a/src/olympia/blocklist/tests/test_utils.py
+++ b/src/olympia/blocklist/tests/test_utils.py
@@ -11,7 +11,7 @@ from olympia import amo
 from olympia.activity.models import ActivityLog
 from olympia.amo.tests import TestCase, addon_factory, block_factory, user_factory
 
-from ..models import Block, BlocklistSubmission, BlockType, BlockVersion
+from ..models import Block, BlocklistSubmission, BlockType
 from ..utils import datetime_to_ts, disable_versions_for_block, save_versions_to_blocks
 
 
@@ -226,7 +226,7 @@ class TestSaveVersionsToBlocks(TestCase):
             url=None,
             updated_by=user,
             disable_addon=True,
-            block_type=BlockVersion.BLOCK_TYPE_CHOICES.BLOCKED,
+            block_type=BlockType.BLOCKED,
             changed_version_ids=[addon.current_version.pk],
             signoff_state=BlocklistSubmission.SIGNOFF_PUBLISHED,
         )
@@ -252,7 +252,7 @@ class TestSaveVersionsToBlocks(TestCase):
             url=None,
             updated_by=self.task_user,
             disable_addon=True,
-            block_type=BlockVersion.BLOCK_TYPE_CHOICES.BLOCKED,
+            block_type=BlockType.BLOCKED,
             changed_version_ids=[addon.current_version.pk],
             signoff_state=BlocklistSubmission.SIGNOFF_PUBLISHED,
         )
@@ -282,7 +282,7 @@ class TestSaveVersionsToBlocks(TestCase):
             url=None,
             updated_by=self.task_user,
             disable_addon=True,
-            block_type=BlockVersion.BLOCK_TYPE_CHOICES.BLOCKED,
+            block_type=BlockType.BLOCKED,
             changed_version_ids=[addon.current_version.pk],
             signoff_state=BlocklistSubmission.SIGNOFF_PUBLISHED,
         )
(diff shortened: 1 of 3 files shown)


src/olympia/blocklist/utils.py
```diff
@@ -147,12 +147,15 @@ def disable_versions_for_block(block, submission):
     """Disable appropriate addon versions that are affected by the Block."""
     from olympia.reviewers.utils import ReviewBase
 
+    task_user = get_task_user()
```
KevinMind, line 150: Reading the PR and issue it's not super clear to me why we have the fallback to `updated_by`. I also don't really understand why the review is considered human **only** if the `task_user` is not the user that last updated the block... this seems kind of arbitrary and really easy to miss or get wrong when looking at this code in 2 weeks let alone 2 years.. thoughts?

src/olympia/blocklist/tests/test_utils.py
```diff
@@ -21,7 +21,7 @@ def test_datetime_to_ts():
 
 class TestSaveVersionsToBlocks(TestCase):
     def setUp(self):
```
KevinMind, line 23: Semi-serious nit. Looking only at the diff here it is, like, totally unclear how these changes to the test satisfactorily test the change in the underlying code. I've looked at the larger file and see that the tests being updated are large and test many things. It's not actually clear to me yet that you have covered all of the potential edge cases (you may have but I'll need to read a lot more code to validate that)

Can I ask for you to instead add a smaller test(s) that covers this specific case:
- ReviewBase.user is task user or fallback to updated_by from the block
- human review is true only if the task user is not the updated_by user.

These 2 test scenarios would clearly and concisely test this behavior, making it easy to have faith in the tests moving forward and east to identify what part of the code might be broken if something changes.

KevinMind:
Neither in the PR nor in the issue are there steps to `Block a listed version from the admin:` could you add those 🙏 so I can verify locally?

src/olympia/blocklist/utils.py
diox: `ActivityLog` needs a `user` no matter what, or it won't be saved (and that will be a silent error until https://github.com/mozilla/addons/issues/14803 is fixed).

If a block has no `updated_by` (which is technically possible, since the field is nullable/not required) we still want to record the activity. So it makes sense to fall back to the task user.

As a rule in addons-server, if the task user is responsible for something, then it was not a human action (see `auto_approve.py` or `auto_reject.py` for instance).

diox:
I'll add the unit test. Regarding `Block a listed version from the admin`, do you really need more instructions regarding how to do this ? In the admin, there is a "add block" button, click it and it should be fairly straightforward from there.

src/olympia/blocklist/tests/test_utils.py
diox: https://github.com/mozilla/addons-server/pull/22817/commits/d895354e38615b6aeff213b74966907efc9af8a3

KevinMind:
> I'll add the unit test. Regarding `Block a listed version from the admin`, do you really need more instructions regarding how to do this ? In the admin, there is a "add block" button, click it and it should be fairly straightforward from there.

I didn't even know where to start 🤷 I've never actually made a block from the admin. I'll give this a go today. This would be a great example of something Alan was asking about where having screeenshots can help. Not only does it help me know where to go for testing, but also validates what I expect to see. Just a thought. I'll give this a try without and see where I get to.

KevinMind:
@diox could you rebase this PR?

diox:
The jinja error you're seeing locally because `DEBUG` is `True` (https://docs.djangoproject.com/en/5.1/topics/templates/#django.template.backends.jinja2.Jinja2 see the `undefined` bit) - it's fine, just complaining about an undefined variable, it's been there forever.

The fact that the version no longer says "unreviewed" in reviewer tools should be proof that it's working - it won't have a due date because it has been reviewed and rejected. You should be able to test the last sentence of my instructions above as well: if the add-on is force-enabled, the unreject versions action should be available, and if that version is unrejected, it should become awaiting review again.

PR #22819, opened 2024-11-04
Skip reindex on normal make up where index alias exists

KevinMind:
Follow up: https://github.com/mozilla/addons-server/pull/22805

### Description

When running make up, if we are not explicitly cleaning and reseting the database and we have a non-empty database it is fair to assume that indexing may have already happened and we can safely skip reindex if the es index is already aliased

### Context

See: LINK TO COMMENT for explanation

### Testing

Run `make up` after running `make down` and you should expect ES indexing to happen. Then run `make up` again and expect it not to happen. Adding `INIT_CEAN=True` or wiping the database should also cause reindexing.

### Checklist

### DIFF
--- a/src/olympia/amo/management/commands/initialize.py
+++ b/src/olympia/amo/management/commands/initialize.py
@@ -62,4 +62,6 @@ class Command(BaseDataCommand):
         # We should reindex even if no data is loaded/modified
         # because we might have a fresh instance of elasticsearch
         else:
-            call_command('reindex', '--wipe', '--force', '--noinput')
+            call_command(
+                'reindex', '--wipe', '--force', '--noinput', '--skip-if-exists'
+            )
--- a/src/olympia/search/management/commands/reindex.py
+++ b/src/olympia/search/management/commands/reindex.py
@@ -117,6 +117,12 @@ class Command(BaseCommand):
             help=('Do not ask for confirmation before wiping. Default: False'),
             default=False,
         )
+        parser.add_argument(
+            '--skip-if-exists',
+            action='store_true',
+            help=('Skip the reindex if the index already exists.'),
+            default=False,
+        )
 
     def accepted_keys(self):
         return ', '.join(settings.ES_INDEXES.keys())
@@ -140,6 +146,9 @@ class Command(BaseCommand):
                 % (self.accepted_keys())
             )
         self.stdout.write('Starting the reindexation for %s.' % alias)
+        if kwargs['skip_if_exists'] and ES.indices.exists_alias(name=alias):
+            self.stdout.write('Index %s already exists. Skipping reindex.' % alias)
+            return
 
         if kwargs['wipe']:
             skip_confirmation = kwargs['noinput']
--- a/src/olympia/search/management/commands/reindex.py
+++ b/src/olympia/search/management/commands/reindex.py
@@ -120,7 +120,7 @@ class Command(BaseCommand):
         parser.add_argument(
             '--skip-if-exists',
             action='store_true',
-            help=('Skip the reindex if the index already exists.'),
+            help=('Skip the reindex if the alias already exists.'),
             default=False,
         )
 
--- a/src/olympia/search/management/commands/reindex.py
+++ b/src/olympia/search/management/commands/reindex.py
@@ -147,7 +147,7 @@ class Command(BaseCommand):
             )
         self.stdout.write('Starting the reindexation for %s.' % alias)
         if kwargs['skip_if_exists'] and ES.indices.exists_alias(name=alias):
-            self.stdout.write('Index %s already exists. Skipping reindex.' % alias)
+            self.stdout.write('Alias %s already exists. Skipping reindex.' % alias)
             return
 
         if kwargs['wipe']:
--- a/src/olympia/search/tests/test_commands.py
+++ b/src/olympia/search/tests/test_commands.py
@@ -113,7 +113,10 @@ class TestIndexCommand(ESTestCaseMixin, PatchMixin, TransactionTestCase):
                 # alias in setUpClass.
                 time.sleep(1)
                 management.call_command(
-                    'reindex', wipe=wipe, noinput=True, stdout=self.stdout
+                    'reindex',
+                    wipe=wipe,
+                    noinput=True,
+                    stdout=self.stdout,
                 )
 
         t = ReindexThread()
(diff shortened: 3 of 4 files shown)


diox (changes requested):
https://github.com/mozilla/addons-server/pull/22805#discussion_r1827860525 still applies:
- You should continue to pass `--wipe` (important for initial setup)
- You should only skip if the alias exists, not the index

KevinMind:
> [#22805 (comment)](https://github.com/mozilla/addons-server/pull/22805#discussion_r1827860525) still applies:
> 
>     * You should continue to pass `--wipe` (important for initial setup)
> 
>     * You should only skip if the alias exists, not the index

Is [this](https://github.com/mozilla/addons-server/pull/22819/files#diff-1b804f1cb77a57dec31314ab2ef3d26372f1e9b613a1411d6f7bf49b76151015R219) not addressing the second point?

For the first point. Looking back at the original point

> If add-ons are saved before the initial reindex is done, then ES will automatically create an index called "addons" that won't be an alias. It won't have the schema however, so some data will be incorrect, and subsequent calls to reindex without passing --wipe will fail.

So if I don't pass wipe, and addons are created before I call the "reindex" command there is a potential failure? It's not totally clear to me why that would be. If I take your words literally in that case ther would still not be an alias so we would recreate the index anyway.. so that is why we need both the skip if aliased and wipe so if we don't skip due to non aliased index we also remove the existing invalid index first? Do I get it? Not sure.. halllp 🤷

diox:
> Is [this](https://github.com/mozilla/addons-server/pull/22819/files#diff-1b804f1cb77a57dec31314ab2ef3d26372f1e9b613a1411d6f7bf49b76151015R219) not addressing the second point?

It's at the wrong place: it should be done for your new `skip_if_exists` argument a few lines above. Don't want to change the line that you changed, that has different consequences in terms of what index we touch...

> So if I don't pass wipe, and addons are created before I call the "reindex" command there is a potential failure? It's not totally clear to me why that would be. If I take your words literally in that case ther would still not be an alias so we would recreate the index anyway.. so that is why we need both the skip if aliased and wipe so if we don't skip due to non aliased index we also remove the existing invalid index first? Do I get it? Not sure.. halllp 🤷

If you don't pass `--wipe` and add-ons are created before we call the initial `reindex`, then `reindex` will never complete. It will create a timestamped `addons-XXXX` index, and after if has indexed all add-ons it will want to swap the `addons` alias to point to that index, and that will fail because it's not an alias. You'll get a traceback in the worker logs, and the `reindex` command will infinitely loop waiting for the task to successfully complete, which it never will.

BTW, the reason we do that alias switch is to allow a full reindex while still having a fully working search. The idea is:
- You execute the command to reindex, that creates a new index almost immediately, but `addons` alias is not updated to point to it just yet. A bunch of indexing tasks are triggered in the background.
- All reads still go to the old index, because `addons` still points to that old index.
- All writes go to both the old index and the new one (in addition to the indexing tasks that were just triggered).
- Eventually all indexing tasks finish, and the alias is switched to point to the new index, and we stop using the old index & delete it.

This is particularly important in production: we want both reads and writes to continue working transparently while we do a full reindex that might take a couple dozen minutes.

KevinMind:
@diox fixed

src/olympia/search/tests/test_commands.py
```diff
@@ -229,3 +237,12 @@ def test_create_workflow_addons(self):
         for addons.
         """
         self._test_workflow('default')
+
+    @mock.patch('olympia.search.management.commands.reindex.ES')
+    def test_reindex_skip_if_exists(self, mock_es):
+        """
+        Test reindex command with skip_if_exists option when index already exists.
+        """
+        mock_es.indices.exists.return_value = True
```
diox, line 246: Not the right mock - it should be `exists_alias` - why is this test passing ?

src/olympia/amo/management/commands/initialize.py
```diff
@@ -58,4 +58,4 @@ def handle(self, *args, **options):
         # We should reindex even if no data is loaded/modified
         # because we might have a fresh instance of elasticsearch
         else:
-            call_command('reindex', '--wipe', '--force', '--noinput')
```
diox, line 61: I wonder if we should keep the `--force` here. It prevents the error from being raise when there is already a reindexation in progress, which I think is what we want here, in case somehow it was triggered from elsewhere?

KevinMind: Done. Also moved the check, above the check for already indexing.

src/olympia/search/tests/test_commands.py
KevinMind: Looksl ike the condition was met regardless.. I simplified the test. Verifiable now that if you modify the mock for alias_exists, the test reliably fails.

KevinMind:
@diox I think this is ready to go now.

src/olympia/search/management/commands/reindex.py
```diff
@@ -117,6 +117,12 @@ def add_arguments(self, parser):
             help=('Do not ask for confirmation before wiping. Default: False'),
             default=False,
         )
+        parser.add_argument(
+            '--skip-if-exists',
+            action='store_true',
+            help=('Skip the reindex if the index already exists.'),
```
diox, line 123:
```suggestion
            help=('Skip the reindex if the alias already exists.'),
```

src/olympia/search/management/commands/reindex.py
```diff
@@ -140,6 +146,9 @@ def handle(self, *args, **kwargs):
                 % (self.accepted_keys())
             )
         self.stdout.write('Starting the reindexation for %s.' % alias)
+        if kwargs['skip_if_exists'] and ES.indices.exists_alias(name=alias):
+            self.stdout.write('Index %s already exists. Skipping reindex.' % alias)
```
diox, line 150:
```suggestion
            self.stdout.write('Alias %s already exists. Skipping reindex.' % alias)
```

src/olympia/search/tests/test_commands.py
```diff
@@ -229,3 +232,20 @@ def test_create_workflow_addons(self):
...
         self._test_workflow('default')
+
+    @mock.patch(
+        'olympia.search.management.commands.reindex.ES.indices.exists_alias',
+        return_value=True,
+    )
+    @mock.patch('olympia.search.management.commands.reindex.Command.create_workflow')
+    def test_reindex_skip_if_exists(self, mock_create_workflow, mock_exists_alias):
+        """
+        Test reindex command with skip_if_exists option when index already exists.
```
diox, line 243:
```suggestion
        Test reindex command with skip_if_exists option when alias already exists.
```

diox (approved):
w/ comments to make it clear we're checking the alias and not the index

PR #22820, opened 2024-11-04
Don't fail when displaying IPv6 and IPv4 together in the admin

diox:
Fixes: mozilla/addons#15134

### DIFF
--- a/src/olympia/amo/admin.py
+++ b/src/olympia/amo/admin.py
@@ -521,7 +521,7 @@ class AMOModelAdmin(admin.ModelAdmin):
             )
 
         contents = format_html_join(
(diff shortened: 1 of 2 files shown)


src/olympia/amo/admin.py
```diff
@@ -521,7 +521,7 @@ def known_ip_adresses(self, obj):
             )
 
         contents = format_html_join(
-            '', '<li>{}</li>', ((ip,) for ip in sorted(ip_addresses))
+            '', '<li>{}</li>', sorted((str(i),) for i in ip_addresses)
```
eviljeff, line 524: It'd be ideal if any ipv6 addresses came after all the ipv4 addresses really... but not failing is the priority so 🤷‍♂️

eviljeff (approved):

PR #22822, opened 2024-11-05
add basic page to proceed/escalate held actions

eviljeff:

### Checklist

### DIFF
--- a/src/olympia/reviewers/urls.py
+++ b/src/olympia/reviewers/urls.py
@@ -73,6 +73,11 @@ urlpatterns = (
         views.whiteboard,
         name='reviewers.whiteboard',
     ),
+    re_path(
+        r'^held-action-review/(?P<decision_id>[^/<>]+)$',
+        views.held_action_review,
+        name='reviewers.held_action_review',
+    ),
     re_path(r'^eula/%s$' % ADDON_ID, views.eula, name='reviewers.eula'),
     re_path(r'^privacy/%s$' % ADDON_ID, views.privacy, name='reviewers.privacy'),
     re_path(r'^motd$', views.motd, name='reviewers.motd'),
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -2,7 +2,7 @@ import functools
 import operator
 from collections import OrderedDict
 from datetime import date, datetime
-from urllib.parse import quote
+from urllib.parse import quote, urljoin
 
 from django import http
 from django.conf import settings
@@ -35,7 +35,8 @@ from rest_framework.viewsets import GenericViewSet
 
 import olympia.core.logger
 from olympia import amo
-from olympia.abuse.models import AbuseReport
+from olympia.abuse.models import AbuseReport, CinderJob, CinderPolicy, ContentDecision
+from olympia.abuse.tasks import handle_escalate_action
 from olympia.access import acl
 from olympia.activity.models import ActivityLog, CommentLog, DraftComment
 from olympia.addons.models import (
@@ -63,6 +64,7 @@ from olympia.api.permissions import (
     AnyOf,
     GroupPermission,
 )
+from olympia.constants.abuse import DECISION_ACTIONS
 from olympia.constants.reviewers import (
     REVIEWS_PER_PAGE,
     REVIEWS_PER_PAGE_MAX,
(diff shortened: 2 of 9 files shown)


src/olympia/reviewers/urls.py
```diff
@@ -73,6 +73,11 @@ def queue_urls():
         views.whiteboard,
         name='reviewers.whiteboard',
     ),
+    re_path(
```
KevinMind, line 76: Nit: maybe calling the url something like

`decision/<decision_id>`

src/olympia/reviewers/views.py
```diff
@@ -1604,3 +1607,44 @@ def queue_held_actions(request, tab):
...
+                    CinderPolicy.objects.filter(
+                        default_cinder_action=DECISION_ACTIONS.AMO_APPROVE
+                    )
+                )
+                decision.process_action(release_hold=True)
+            case 'forward':
+                decision.update(action_date=datetime.now())
+                for job in cinder_jobs_qs:
+                    handle_escalate_action.delay(job_pk=job.id)
+            case _:
```
KevinMind, line 1637: This should probably error no. Not sure how this state would even happen but if it does currently it will just re-render the page and be very unobvious to the user.

src/olympia/reviewers/urls.py
eviljeff: Like, instead of held_action_review? Or in addition to it?

KevinMind:
Setup test data:

### Connect to Cinder and download policies 

Make sure to setup your `CINDER_API_TOKEN` in `local_settings.py`. This is only needed if you don't have existing `CinderPolicy` instances locally.

Verify that cinder is connected via the <http://olympia.test/services/monitor.json> endpoint where `cinder` should have `state` set to `true`

Open the [cinder policy admin page](http://olympia.test/en-US/admin/models/abuse/cinderpolicy/) and click `Sync from cinder`

You should expect new policies to be added after refreshing the page. You can now select a cinder policy id to be used in the python script below. 

### Setup Test data

```python
from olympia.constants.abuse import DECISION_ACTIONS

addon_id=<addon>
cinder_policy_id=<cinder_policy>

decision, _ = ContentDecision.objects.get_or_create(action=DECISION_ACTIONS.AMO_DISABLE_ADDON, notes='some notes', addon_id=addon_id)
cinder_policy=CinderPolicy.objects.get(pk=cinder_policy_id)

decision.policies.add(cinder_policy)
```

### Verify decision without cinder job

> Make sure you do _not_ have a `CinderJob` associated with the `addon_id` you are using as this should only be true for the next scenario

```python
assert CinderJob.objects.filter(

Open <http://olympia.test/en-US/reviewers/held-action-review/1>

Expect to see the page including the relevant metadata and a form including only 2 options

<img width="1363" alt="image" src="https://github.com/user-attachments/assets/6fa8f501-7aa6-4d2b-a575-026463eb4c65">

### Verify decision with cinder job

First associate a `CinderJob` with your existing decision from before. In the same shell as before or using at least the same value for `addon_id`

```python
cinder_job, created = CinderJob.objects.get_or_create(target_addon_id=addon_id)
```

No open <http://olympia.test/en-US/reviewers/held-action-review/1>

Expect to see additional options to forward the decision

src/olympia/reviewers/views.py
KevinMind: Then let's remove it?

KevinMind:
@eviljeff is it possible to review the outcome of selecting all 3 possible options in the form using the same `CinderDecision` is it possible to "revert" the form on the shell level, or easier to just recreate again? Also, can you add:

- expectations for selecting each field and how to validate via the django admin/ reviewer tools
- additional details for how I would test this below.. It's not really clear what commands I should run in the shell to create the necessary state in the DB

> You can test Rating, Collection, and User similarly with a rating_id, and DECISION_ACTIONS.AMO_DELETE_RATING, etc (but no need to create a CinderJob, as there's no tool to forward those content types too anyway)

src/olympia/reviewers/urls.py
KevinMind: instead of.. this page is for showing a decision right? held, action, and review have nothing to do with cinder or decision so it feels like decision is the most descriptive word.. but just an singular opiniomation.

eviljeff: The thinking was it was a review for the action associated with the decision, which was held from being carried out pending that review.  I _do_ think we need "review" in the url, but "/decision_review"? 🤷

eviljeff:
> @eviljeff is it possible to review the outcome of selecting all 3 possible options in the form using the same `CinderDecision` is it possible to "revert" the form on the shell level, or easier to just recreate again? 

you can clear `action_date` on the `ContentDecision` instance to get it back to the same state - if the previous action was an add-on disable the easiest fix it to use the reviewer tools to force enable.  For other objects/actions you can use django shell to update them (it's a bit of a lucky dip how deletes and disables are handled on each type of content though!)

> Also, can you add:
> 
>     * expectations for selecting each field and how to validate via the django admin/ reviewer tools
> 
>     * additional details for how I would test this below.. It's not really clear what commands I should run in the shell to create the necessary state in the DB
> 
> > You can test Rating, Collection, and User similarly with a rating_id, and DECISION_ACTIONS.AMO_DELETE_RATING, etc (but no need to create a CinderJob, as there's no tool to forward those content types too anyway)

Done

KevinMind:
Is this the desired result if no decision exists? Would we not want a slightly prettier 404?

image

KevinMind:
Also when submitting the form should we not redirect to the listed/unlisted page for that addon? going to the review home page feels like the wrong place

eviljeff:
> Is this the desired result if no decision exists? Would we not want a slightly prettier 404? In production this would be the standard 404 that doesn't even tell you why you're getting that page.
> 
> image
> 

Yes, that's the standard result if your URL mangling references a nonexistent ID of some sort. We use `get_object_or_404` in quite a few places.

KevinMind:
### enable 2 waffle switches

See explanation [here](https://mozilla-hub.atlassian.net/browse/AMOENG-1199?focusedCommentId=960594)

- dsa-job-technical-processing (this gates the intake of new abuse reports into Cinder.)
- dsa-cinder-forwarded-review (enable forwarding a decision to another queue in reviewer tools)

### create an abuse report for the addon from the frontend
- navigate to <http://olympia.test/en-US/firefox/feedback/addon/1>
- enter any information for the form (it doesn't really matter)
- Note what text you enter for "Provide more details"

> If verifying this in a production-like environment, ensure that you create the report for a high_profile addon, so we know it will be a held decision

### Process in cinder

> In production you can create the decision for the abuse report in cinder. Cinder will send the decision back to amo. Make sure to choose policy `amo_disable_addon`

Follow the steps [above](https://github.com/mozilla/addons-server/pull/22822#issuecomment-2464240727) for creating a content decision

### Process in reviewer tools

1. http://olympia.test/en-US/reviewers/decision-review/1
2. Select "escalate" option and submit the form
3. Navigate to manual review [queue](http://olympia.test/en-US/reviewers/queue/extension)
4. Expect your addon to be in the manual review queue

image

5. Navigate to the review page <http://olympia.test/en-US/reviewers/review/1>
6. expect in the "resolved reports" tab to see the abuse report you filed for the addon

image

KevinMind (approved):
Verified for disable addon. works as expected with udpated testing instructions for the escalate flow.

PR #22826, opened 2024-11-05
Generate and write stash including soft blocked versions

KevinMind:
Fixes: mozilla/addons#15050

### Description

This PR updates the `generate_and _write_stash` method on `MLBF` class to consider the range of use cases covered in the linked jira issue document.

We have to add support for specifying a block_type when requesting a change count. This will be useful for creating type level filters as we need to know for each type if we have any changes and enough changes for a new filter.

We have to modify which guids are included in the now 3 different properties on the stash. This has to cover the set of use cases where blocks are added as hard or soft as well as moving from soft to hard and vice versa.

### Context

When updating our bloom filter we are now going to create a unified stash which will allow us to enable support for adding/removing items from the soon to be soft block filter, but, to do so in a way that is backwards compatible with legacy fx clients that have no concept of soft blocking. This raises some edge cases that increased the complexity of this patch. Read the doc for more info

### Testing

Look at the automated tests but essentially we should consider the following scenarios and you can independently verify them if you don't trust the tests.

Here is a [detailed break ](https://github.com/mozilla/addons-server/pull/22826/files#diff-9df60b77ee1f859628ea78fa4cbe1729ea4dc9d67d0a76c77da8003e52c1fba9R393-R448)down of the scenarios

### Checklist

### DIFF
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -6,6 +6,7 @@ from typing import Dict, List, Optional, Tuple
 
 from django.utils.functional import cached_property
 
+import waffle
 from filtercascade import FilterCascade
 from filtercascade.fileformats import HashAlgorithm
 
@@ -245,17 +246,52 @@ class MLBF:
         }
 
     def generate_and_write_stash(self, previous_mlbf: 'MLBF' = None):
-        # compare previous with current blocks
-        extras, deletes, _ = self.generate_diffs(previous_mlbf)[BlockType.BLOCKED]
+        """
+        Generate and write the stash file representing changes between the
+        previous and current bloom filters. See:
+        https://bugzilla.mozilla.org/show_bug.cgi?id=soft-blocking
+
+        In order to support Firefox clients that don't support soft blocking,
+        unblocked is a union of deletions from blocked and deletions from
+        soft_blocked, filtering out any versions that are in the newly blocked
+        list.
+
+        Versions that move from hard to soft blocked will be picked up by old
+        clients as no longer hard blocked by being in the unblocked list.
+
+        Clients supporting soft blocking will also see soft blocked versions as
+        unblocked, but they won't unblocked them because the list of
+        soft-blocked versions takes precedence over the list of unblocked
+        versions.
+
+        Versions that move from soft to hard blocked will be picked up by
+        all clients in the blocked list. Note, even though the version is removed
+        from the soft blocked list, it is important that we do not include it
+        in the "unblocked" stash (like for hard blocked items) as this would
+        result in the version being in both blocked and unblocked stashes.
+        """
+        diffs = self.generate_diffs(previous_mlbf)
+        blocked_added, blocked_removed, _ = diffs[BlockType.BLOCKED]
         stash_json = {
-            'blocked': extras,
-            'unblocked': deletes,
+            'blocked': blocked_added,
+            'unblocked': blocked_removed,
         }
+
+        if waffle.switch_is_active('mlbf-soft-blocks-enabled'):
+            soft_blocked_added, soft_blocked_removed, _ = diffs[BlockType.SOFT_BLOCKED]
+            stash_json['softblocked'] = soft_blocked_added
+            stash_json['unblocked'] = [
+                unblocked
+                for unblocked in (blocked_removed + soft_blocked_removed)
+                if unblocked not in blocked_added
+            ]
+
         # write stash
         stash_path = self.stash_path
         with self.storage.open(stash_path, 'w') as json_file:
             log.info(f'Writing to file {stash_path}')
             json.dump(stash_json, json_file)
+        return stash_json
 
     def blocks_changed_since_previous(
         self, block_type: BlockType = BlockType.BLOCKED, previous_mlbf: 'MLBF' = None
--- a/src/olympia/blocklist/tests/test_mlbf.py
+++ b/src/olympia/blocklist/tests/test_mlbf.py
@@ -1,6 +1,8 @@
 import json
 from functools import cached_property
 
+from waffle.testutils import override_switch
+
 from olympia import amo
 from olympia.addons.models import GUID_REUSE_FORMAT
 from olympia.amo.tests import (
@@ -494,6 +496,129 @@ class TestMLBF(_MLBFBase):
             ),
         }
 
+    def test_diff_all_possible_changes(self):
+        """
+        Simulates 6 guid:version combinations moving from one state to another
+        covering all possible movements and the final diff state.
+
+        1. Not blocked -> Soft blocked
+        2. Not blocked -> Blocked
+        3. Soft blocked -> Blocked
+        4. Soft blocked -> Not blocked
+        5. Blocked -> Soft blocked
+        6. Blocked -> Not blocked
+
+        """
+        # Create a version that isn't blocked yet.
+        one = addon_factory(guid='1', file_kw={'is_signed': True})
+        (one_hash,) = MLBF.hash_filter_inputs([(one.guid, one.current_version.version)])
+        # Create a second version not blocked yet.
+        two = addon_factory(guid='2', file_kw={'is_signed': True})
+        (two_hash,) = MLBF.hash_filter_inputs([(two.guid, two.current_version.version)])
+        # Create a soft blocked version.
+        three, three_block = self._blocked_addon(
+            guid='3', block_type=BlockType.SOFT_BLOCKED, file_kw={'is_signed': True}
+        )
+        (three_hash,) = MLBF.hash_filter_inputs(
+            [(three.guid, three.current_version.version)]
+        )
+        # Create another soft blocked version.
+        four, four_block = self._blocked_addon(
+            guid='4', block_type=BlockType.SOFT_BLOCKED, file_kw={'is_signed': True}
+        )
+        (four_hash,) = MLBF.hash_filter_inputs(
+            [(four.guid, four.current_version.version)]
+        )
+        # Create a hard blocked version.
+        five, five_block = self._blocked_addon(
+            guid='5', block_type=BlockType.BLOCKED, file_kw={'is_signed': True}
+        )
+        (five_hash,) = MLBF.hash_filter_inputs(
+            [(five.guid, five.current_version.version)]
+        )
+        # And finally, create another hard blocked version.
+        six, six_block = self._blocked_addon(
+            guid='6', block_type=BlockType.BLOCKED, file_kw={'is_signed': True}
+        )
+        (six_hash,) = MLBF.hash_filter_inputs([(six.guid, six.current_version.version)])
+
+        # At this point, we have 2 versions not blocked,
+        # and 4 versions blocked. We're going
+        # to generate a first MLBF from that set of versions.
+        first_mlbf = MLBF.generate_from_db('first')
+
+        # We expect the 4 blocked versions to be in the diff, sorted by block type.
+        assert first_mlbf.generate_diffs() == {
+            BlockType.BLOCKED: ([five_hash, six_hash], [], 2),
+            BlockType.SOFT_BLOCKED: ([three_hash, four_hash], [], 2),
+        }
+
+        # The first time we generate the stash, we expect 3 to 6 to be in the stash
+        # as they have some kind of block applied.
+        assert first_mlbf.generate_and_write_stash() == {
+            'blocked': [five_hash, six_hash],
+            'unblocked': [],
+        }
+
+        with override_switch('mlbf-soft-blocks-enabled', active=True):
+            assert first_mlbf.generate_and_write_stash() == {
+                'blocked': [five_hash, six_hash],
+                'softblocked': [three_hash, four_hash],
+                'unblocked': [],
+            }
+
+        # Update the existing blocks, and create new ones for
+        # the versions "one" and "two".
+
+        # The first version gets soft blocked now.
+        block_factory(
+            guid=one.guid, updated_by=self.user, block_type=BlockType.SOFT_BLOCKED
+        )
+        # The second version is hard blocked.
+        block_factory(guid=two.guid, updated_by=self.user, block_type=BlockType.BLOCKED)
+        # 3 was soft-blocked and is now hard blocked.
+        three_block.blockversion_set.first().update(block_type=BlockType.BLOCKED)
+        # 4 was soft blocked and is now unblocked.
+        four_block.delete()
+        # 5 was hard blocked and is now soft blocked.
+        five_block.blockversion_set.first().update(block_type=BlockType.SOFT_BLOCKED)
+        # 6 was hard blocked and is now unblocked.
+        six_block.delete()
+
+        # We regenerate another MLBF based on the updates we've just done
+        # to verify the final state of each version.
+        second_mlbf = MLBF.generate_from_db('second')
+
+        # The order is based on the ID (i.e. creation time) of the block,
+        # not the version so we expect two after three since two was
+        # blocked after three.
+        assert second_mlbf.generate_diffs(previous_mlbf=first_mlbf) == {
+            BlockType.BLOCKED: (
+                [three_hash, two_hash],
+                [five_hash, six_hash],
+                4,
+            ),
+            # Same as above, one had a block created after five so it comes second.
+            BlockType.SOFT_BLOCKED: (
+                [five_hash, one_hash],
+                [three_hash, four_hash],
+                4,
+            ),
+        }
+
+        assert second_mlbf.generate_and_write_stash(previous_mlbf=first_mlbf) == {
+            'blocked': [three_hash, two_hash],
+            # 4 is omitted because it's soft blocked and the waffle switch is off
+            'unblocked': [five_hash, six_hash],
+        }
+
+        with override_switch('mlbf-soft-blocks-enabled', active=True):
+            assert second_mlbf.generate_and_write_stash(previous_mlbf=first_mlbf) == {
+                'blocked': [three_hash, two_hash],
+                'softblocked': [five_hash, one_hash],
+                'unblocked': [five_hash, six_hash, four_hash],
+            }
+
     def test_generate_stash_returns_expected_stash(self):
         addon, block = self._blocked_addon()
         block_versions = [
@@ -513,6 +638,13 @@ class TestMLBF(_MLBFBase):
                 'unblocked': [],
             }
 
+        with override_switch('mlbf-soft-blocks-enabled', active=True):
+            assert mlbf.generate_and_write_stash() == {
+                'blocked': MLBF.hash_filter_inputs(expected_blocked),
+                'softblocked': [],
+                'unblocked': [],
+            }
+
         # Remove the last block version
         block_versions[-1].delete()
         expected_unblocked = expected_blocked[-1:]
@@ -523,6 +655,13 @@ class TestMLBF(_MLBFBase):
         with next_mlbf.storage.open(next_mlbf.stash_path, 'r') as f:
             assert json.load(f) == {
                 'blocked': [],
+                'unblocked': MLBF.hash_filter_inputs(expected_unblocked),
+            }
+
+        with override_switch('mlbf-soft-blocks-enabled', active=True):
+            assert next_mlbf.generate_and_write_stash(previous_mlbf=mlbf) == {
+                'blocked': [],
+                'softblocked': [],
                 'unblocked': MLBF.hash_filter_inputs(expected_unblocked),
             }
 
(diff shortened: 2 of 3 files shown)


KevinMind:
@willdurand and @diox 2 remaining questions from my side.

1. what logic should we use to determine the number of "changed" guids when including the differing handling of hard <-> soft blocks in legacy versus new clients. Some items are included in multiple stashes so should they equally double increase the counts? Seems intuitive enough but not sure if we care about that.
2. It is important that the time used for the stash is earlier than the time used for the filter in the event that we create both. Currently we use a single generation time for both. Do we want to always make the filter generation time +1 milisecond of the stash? or vice versa or how exactly should we handle this?

src/olympia/blocklist/mlbf.py
```diff
@@ -240,20 +240,37 @@ def generate_diffs(
         }
 
     def generate_and_write_stash(self, previous_mlbf: 'MLBF' = None):
-        # compare previous with current blocks
-        extras, deletes, _ = self.generate_diffs(previous_mlbf)[BlockType.BLOCKED]
+        """
+        Generate and write the stash file representing
+        changes between the previous and current MLBF Filters.
+        In order to support older FX clients that only understood blocked and unblocked
+        The unblocked list is a union of deletions from blocked and soft_blocked.
```
willdurand, line 247: Where is this coming from?

KevinMind: Unblocked includes versions which have gone from hard blocked or from soft blocked to unblocked. It also includes versions which have gone from hard blocked to soft blocked.

KevinMind: This is attempting to implement what is written in the google doc you wrote.

KevinMind: I updated the doc string to be a bit clearer. lmk

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -300,19 +311,217 @@ def test_diff_block_added_and_removed(self):
...
+    def test_diff_all_possible_changes(self):
+        """
+        Simulates 6 guid:version comibinations moving from one state to another
+        covering all possible movements and the final diff state.
+
+        1. Not blocked -> Soft blocked
+        2. Not blocked -> Blocked
+        3. Soft blocked -> Blocked
+        4. Soft blocked -> Not blocked
+        5. Blocked -> Soft blocked
```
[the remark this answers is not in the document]
KevinMind: Use of the word this should be forbidden in code reviews. Which this are you referring to here?

Each one of these edge cases is important to cover for either modern or legacy versions of FX and moreover we have to ensure the backward compatibility that we agreed on (that you proposed) does not break either legacy or modern Fx clients.

KevinMind:
@willdurand rebased

src/olympia/blocklist/tests/test_mlbf.py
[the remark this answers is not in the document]
willdurand: _This_ line 😬  Not sure if that makes the code more complex or anything, but there won't be hard to soft block transitions.

src/olympia/blocklist/mlbf.py
```diff
@@ -245,17 +245,42 @@ def generate_diffs(
         }
 
     def generate_and_write_stash(self, previous_mlbf: 'MLBF' = None):
-        # compare previous with current blocks
-        extras, deletes, _ = self.generate_diffs(previous_mlbf)[BlockType.BLOCKED]
+        """
+        Generate and write the stash file representing
+        changes between the previous and current MLBF Filters.
+
+        In order to support older FX clients that only understood blocked and unblocked
```
willdurand, line 252:
```suggestion
        In order to support older FX clients that only understood blocked and unblocked,
```

src/olympia/blocklist/mlbf.py
```diff
@@ -245,17 +245,42 @@ def generate_diffs(
...
+        unblocked is a union of deletions from blocked and additions to soft_blocked.
+
+        Versions that move from hard to soft blocked will be picked up by old clients
+        as no longer hard blocked by being in the unblocked list.
+
+        Newer clients will also see soft blocked items as unblocked
+        but then will check the soft blocked list and re-add them to soft blocked.
```
willdurand, line 259: I am reading the PR top to bottom and this is one of the first comments... Q: is this adding complexity in the patch? If so, we can probably do without it because Product + others said we didn't need to account for hard to soft transitions.

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +452,145 @@ def test_diff_block_soft_to_hard(self):
...
+        6. Blocked -> Not blocked
+
+        first stash
+        {
+          'blocked': [5,6],
+          'softblocked': [3,4],
+          # 1,2 are already unblocked so wouldn't be on the stash
```
willdurand, line 471: I am not clear on why that's `true`.

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +452,145 @@ def test_diff_block_soft_to_hard(self):
...
+        (five_hash,) = MLBF.hash_filter_inputs(
+            [(five.guid, five.current_version.version)]
+        )
+        # 6. Blocked -> Not blocked
+        six, six_block = self._blocked_addon(
+            guid='6', block_type=BlockType.BLOCKED, file_kw={'is_signed': True}
+        )
+        (six_hash,) = MLBF.hash_filter_inputs([(six.guid, six.current_version.version)])
+
+        first_mlbf = MLBF.generate_from_db('first')
```
willdurand, line 547: I am not clear on why the first mlbf would generate differences if we create all the blocks _before_ we generate the bloom filter from the DB.

willdurand (commented):
I have questions.

src/olympia/blocklist/tests/test_mlbf.py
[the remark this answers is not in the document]
KevinMind: There is a PR introducing that feature (hard->soft) now so AFAIK yes it will be possible to do. Even if we don't use hard->soft feature I still think it makes sense to account for the total possible transitions, MLBF shouldn't care about how we intend to use the feature it just works.. esp since the code covering this case is already here and tested.

hard to soft results in a version being in both unblocked and soft-blocked on the next stash

src/olympia/blocklist/tests/test_mlbf.py
KevinMind: What what is true? 1 and 2 were blocked before we re-generated the stash so we don't add them to "unblocked".

src/olympia/blocklist/tests/test_mlbf.py
KevinMind: If there is no previous filter to compare to, then the filter will consider everything in the DB a "change" because, well, it is. Nothing to something.

T0: we create all the blocks
T1: we create the first_mlbf (it includes all of the "changes" from T0. since there is no previous filter to compare to everything is a change)
T2: we create and update some records
T3: we create another filter and now track the changes from T1 to T3

KevinMind: FWIW that is existing behavior and is how regenerating the filter works.

src/olympia/blocklist/tests/test_mlbf.py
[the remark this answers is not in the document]
willdurand: Ok, that's fine.

src/olympia/blocklist/mlbf.py
willdurand: Well, it might not be the right question but that is still _my_ question, and a question that I want an answer for. That said, the code doesn't look too complicated so probably fine.

src/olympia/blocklist/tests/test_mlbf.py
willdurand: Comments in GitHub diffs are contextual. They apply to individual lines (or multiple lines, sometimes).

> 1 and 2 were blocked before we re-generated the stash so we don't add them to "unblocked".

That wasn't entirely clear to me. I read the comment then jumped to the generation of the MLBF and the assertion, and didn't quite get how we ended in this state. Maybe splitting the long python comment into inlined comments in the test would be easier to follow? I don't know.

src/olympia/blocklist/tests/test_mlbf.py
willdurand: OK so it looks like we don't create all the blocks for all the versions there, which is what I assumed. The first two versions are not blocked at this point. We block them after the initial MLBF is generated.

src/olympia/blocklist/mlbf.py
KevinMind: I was trying to answer your question.. hope that I did. To quote the python style guide "simple is better than complex" but "complex is better than complicated" I think this patch is somewhere between simple and complex, but at least not complicated. All the complexity is in one place and changes elsewhere will not impact or be impacted by this code. Closing this thread.

src/olympia/blocklist/tests/test_mlbf.py
KevinMind: Correct, following the spec in the test description. I think you should now understand [why 1,2 were already unblocked](https://github.com/mozilla/addons-server/pull/22826#discussion_r1837935103). Question is, how could I update the test/comments to make it more clear from the start?

src/olympia/blocklist/tests/test_mlbf.py
KevinMind: I know which line you referred to but it wasn't clear to me exactly which part of the statement being true is not clear. Would something like

"Versions that were already not hard or softblocked will not be included in any stash since their state has not changed"

Would that have helped do you think?

willdurand: Not really. I think it's more mapping parts of the large comment to what happens in the test that could be helpful.

KevinMind: Updated by removing the long comment and trying to make the inline comments more clear.

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +452,109 @@ def test_diff_block_soft_to_hard(self):
...
+        )
+        # 6. (Blocked) -> (Not blocked)
+        six, six_block = self._blocked_addon(
+            guid='6', block_type=BlockType.BLOCKED, file_kw={'is_signed': True}
+        )
+        (six_hash,) = MLBF.hash_filter_inputs([(six.guid, six.current_version.version)])
+
+        first_mlbf = MLBF.generate_from_db('first')
+
+        assert first_mlbf.generate_diffs() == {
```
willdurand, line 505:
```suggestion
        # We expect the 4 blocked versions to be in the diff, sorted by block type.
        assert first_mlbf.generate_diffs() == {
```

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +452,109 @@ def test_diff_block_soft_to_hard(self):
...
+            BlockType.BLOCKED: ([five_hash, six_hash], [], 2),
+            BlockType.SOFT_BLOCKED: ([three_hash, four_hash], [], 2),
+        }
+
+        # The first time we generate the stash, we expect 3-6 to be in the stash
```
willdurand, line 510:
```suggestion
        # The first time we generate the stash, we expect 3 to 6 to be in the stash
```

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +452,109 @@ def test_diff_block_soft_to_hard(self):
...
+        # as they have some kind of block applied
+        assert first_mlbf.generate_and_write_stash() == {
+            'blocked': [five_hash, six_hash],
+            'softblocked': [three_hash, four_hash],
+            'unblocked': [],
+        }
+
+        # Transform the addons to the next state
+
+        # 1. Not blocked -> (Soft blocked)
```
willdurand, line 526:
```suggestion
        # The first version gets soft blocked now.
```

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +452,109 @@ def test_diff_block_soft_to_hard(self):
...
+        block_factory(
+            guid=one.guid, updated_by=self.user, block_type=BlockType.SOFT_BLOCKED
+        )
+        # 2. Not blocked -> (Blocked)
```
willdurand, line 530:
```suggestion
        # The second version is hard blocked.
```

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +452,109 @@ def test_diff_block_soft_to_hard(self):
...
+        block_factory(guid=two.guid, updated_by=self.user, block_type=BlockType.BLOCKED)
+        # 3. Soft blocked -> (Blocked)
```
willdurand, line 532:
```suggestion
        # 3 was soft-blocked and is now hard blocked.
```

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +452,109 @@ def test_diff_block_soft_to_hard(self):
...
+        three_block.blockversion_set.first().update(block_type=BlockType.BLOCKED)
+        # 4. Soft blocked -> (Not blocked)
```
willdurand, line 534:
```suggestion
        # 4 was soft blocked and is now unblocked.
```

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +452,109 @@ def test_diff_block_soft_to_hard(self):
...
+        four_block.delete()
+        # 5. Blocked -> (Soft blocked)
```
willdurand, line 536:
```suggestion
        # 5 was hard blocked and is now soft blocked.
```

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +452,109 @@ def test_diff_block_soft_to_hard(self):
...
+        five_block.blockversion_set.first().update(block_type=BlockType.SOFT_BLOCKED)
+        # 6. Blocked -> (Not blocked)
```
willdurand, line 538:
```suggestion
        # 6 was hard blocked and is now unblocked.
```

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +452,109 @@ def test_diff_block_soft_to_hard(self):
...
+        six_block.delete()
+
+        second_mlbf = MLBF.generate_from_db('second')
```
willdurand, line 541:
```suggestion

        # We regenerate another MLBF based on the updates we've just done to verify the final
        # state of each version.
        second_mlbf = MLBF.generate_from_db('second')
```

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +452,109 @@ def test_diff_block_soft_to_hard(self):
...
+        # The order is based on the ID (i.e. creation time) of the block,
+        # not the version so we expect two after three
+        # since two was blocked after three
+        assert second_mlbf.generate_diffs(previous_mlbf=first_mlbf) == {
+            BlockType.BLOCKED: (
+                [three_hash, two_hash],
+                [five_hash, six_hash],
+                4,
+            ),
+            # Same as above, one had a block created after five so it comes second
```
willdurand, line 552: nit

```suggestion
            # Same as above, one had a block created after five so it comes second.
```

willdurand (changes requested):
This looks good overall but given the number of suggestions, I'm going to request changes to get a chance to re-read the entire diff once updated. Given that most of the tests are in the single `test_diff_all_possible_changes` test method, I made numerous suggestions to get it better documented for our future selves (not that it wasn't well documented initially).

Another thing that I am wondering is that we do a lot of comparisons between a MLBF (so a base bloom filter) and changes. Do we have test coverage to generate a stash after a previous stash was created?

src/olympia/blocklist/mlbf.py
```diff
@@ -245,17 +245,42 @@ def generate_diffs(
...
+        Generate and write the stash file representing
+        changes between the previous and current MLBF Filters.
+
+        In order to support older FX clients that only understood blocked and unblocked,
+        unblocked is a union of deletions from blocked and additions to soft_blocked.
+
+        Versions that move from hard to soft blocked will be picked up by old clients
+        as no longer hard blocked by being in the unblocked list.
+
+        Newer clients will also see soft blocked items as unblocked
```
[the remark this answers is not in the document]
KevinMind: Done

src/olympia/blocklist/mlbf.py
```diff
@@ -245,17 +245,42 @@ def generate_diffs(
...
+        but then will check the soft blocked list and re-add them to soft blocked.
+
+        Versions that move from soft to hard blocked will be picked up by old clients
+        and new clients in the blocked list. However, we prevent these versions from
+        also being in the unblocked list as this would produce a contradiction
+        since unblocked is shared by hard and soft blocked filters.
```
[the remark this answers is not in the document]
KevinMind: It's exactly that scenario that I was trying to describe. Updated the comment to be more explicit.

src/olympia/blocklist/mlbf.py
```diff
@@ -245,17 +245,42 @@ def generate_diffs(
...
+        blocked_added, blocked_removed, _ = diffs[BlockType.BLOCKED]
+        soft_blocked_added, soft_blocked_removed, _ = diffs[BlockType.SOFT_BLOCKED]
         stash_json = {
-            'blocked': extras,
-            'unblocked': deletes,
+            'blocked': blocked_added,
+            'softblocked': soft_blocked_added,
+            'unblocked': [
+                unblocked
+                for unblocked in (blocked_removed + soft_blocked_removed)
```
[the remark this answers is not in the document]
KevinMind: Fixed

KevinMind:
@willdurand I'm not sure I get it but

> Another thing that I am wondering is that we do a lot of comparisons between a MLBF (so a base bloom filter) and changes. Do we have test coverage to generate a stash after a previous stash was created?

Each time we run the cron job we check the current state of the database from the previous `cache.json` and if there are >0 < threshold number of changes we create a stash.

It's actually totally irrelevant if we created a stash the last time or not. It only matters what was the state of the DB during the last run. In the case where we don't have any filter to compare to, we cannot create a stash because we need to create a filter.

willdurand:
> @willdurand I'm not sure I get it but
> 
> > Another thing that I am wondering is that we do a lot of comparisons between a MLBF (so a base bloom filter) and changes. Do we have test coverage to generate a stash after a previous stash was created?
> 
> Each time we run the cron job we check the current state of the database from the previous `cache.json` and if there are >0 < threshold number of changes we create a stash.
> 
> It's actually totally irrelevant if we created a stash the last time or not. It only matters what was the state of the DB during the last run. In the case where we don't have any filter to compare to, we cannot create a stash because we need to create a filter.

You got it right, I think. Do we have test coverage that reads a previous `cache.json` file? Something a bit more end-to-end than calling individual methods.

src/olympia/blocklist/mlbf.py
```diff
@@ -245,17 +246,52 @@ def generate_diffs(
...
+        diffs = self.generate_diffs(previous_mlbf)
+        blocked_added, blocked_removed, _ = diffs[BlockType.BLOCKED]
         stash_json = {
-            'blocked': extras,
-            'unblocked': deletes,
+            'blocked': blocked_added,
+            'unblocked': blocked_removed,
         }
+
+        if waffle.switch_is_active('mlbf-soft-blocks-enabled'):
```
willdurand, line 277: Ha, I see. So there is the waffle switch. Given that the migration file is short, I would just add it in this patch, rather than in its own PR without anything else.

KevinMind: It's referenced in both PRs. Why not just land it?

willdurand (approved):
r+wc, thank you!

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +454,129 @@ def test_diff_block_soft_to_hard(self):
...
+            ),
+            # Same as above, one had a block created after five so it comes second.
+            BlockType.SOFT_BLOCKED: (
+                [five_hash, one_hash],
+                [three_hash, four_hash],
+                4,
+            ),
+        }
+
+        assert second_mlbf.generate_and_write_stash(previous_mlbf=first_mlbf) == {
```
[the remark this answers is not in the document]
KevinMind: I think we will remove the switch before flipping the default, and even if we did, I'd just update the tests then.

KevinMind:
@willdurand added the integration test for invalid cache, we already had for previously existing cache.

PR #22827, opened 2024-11-06
Add ability to harden / soften blocks in the admin

diox:
Fixes: mozilla/addons#15084

### Description

See title, this allows admins to harden or soften a block (either partially or completely).

### Context

Like with the rest of the block operations,  you're never directly modifying `Block` or `BlockVersion` for this in the admin, instead you go through a `BlocklistSubmission` with a specific `action`. This allows signoff and makes the UI consistent across the various blocklist-related features in the admin.

### Testing

There should be harden & soften buttons for a block. Once on the corresponding page, you should be able to select only the hard-blocked versions in that block for the soften action, and the opposite, soft-blocked versions for the harden action. Once you submit, the selected versions should have their block type changed according to the action performed. 

The signoff mechanism is reused, untouched. So, if the change requires signoff (average daily users for the add-on is above configured threshold of `settings.DUAL_SIGNOFF_AVERAGE_DAILY_USERS_THRESHOLD`), the changes are pending approval, which is performed as normal in the same admin (in production it would be a different user, if `DEBUG` is `True` we allow the same user)

### Checklist


src/olympia/blocklist/admin.py
```diff
@@ -658,6 +667,7 @@ def changeform_view(self, request, obj_id=None, form_url='', extra_context=None)
                 raise PermissionDenied
             ModelForm = self.get_form(request, obj, change=bool(obj_id))
             form = ModelForm(request.POST, request.FILES, instance=obj)
+            # FIXME: change action here depending on _soften/_harden ?
```
diox, line 670: Still need to decide the best way to do this.

diox:
### Screenshots
#### Slight copy changes on Delete Block page (call the action next to the blocked version "Unblock")
Screenshot 2024-11-08 at 17-45-27 Delete Block Django site admin

#### Hardening a complex block
Note that only the soft-blocked version is available to select
Screenshot 2024-11-08 at 17-45-10 Harden Block Django site admin

#### Softening a complex block
Opposite as before
Screenshot 2024-11-08 at 17-48-15 Soften Block Django site admin

#### Soften / Harden buttons shown on Block page
Both buttons available
Screenshot 2024-11-08 at 17-49-08 Block {105ca783-0522-4143-b8ed-062cba7028a9} View block Django site admin
Only Softening button available
Screenshot 2024-11-08 at 17-44-46 Block {ff560d64-a5f6-4027-92c7-2185a959bb61} View block Django site admin
Only Hardening button available
Screenshot 2024-11-08 at 17-44-36 Block {8598fe93-e71a-429a-9557-5bbb8c3e53dd} View block Django site admin

src/olympia/blocklist/models.py
```diff
@@ -485,7 +495,11 @@ def process_input_guids(
 
     def save_to_block_objects(self):
         assert self.is_submission_ready
-        assert self.action == self.ACTION_ADDCHANGE
+        assert self.action in (
+            self.ACTION_ADDCHANGE,
+            self.ACTION_HARDEN,
+            self.ACTION_SOFTEN,
+        )
```
eviljeff, line 502: nit: if this was all new code I'd suggest using django-extended-choices so we could have all the actions defined together, and a sub-group for actions that need saving.  (marginal benefit with the code already written 🤷 )

src/olympia/blocklist/forms.py
```diff
@@ -77,6 +77,19 @@ class BlocksWidget(forms.widgets.SelectMultiple):
...
+        """Return the verb to use when displaying a given version, depending
+        on the action.
+
+        Re-uses BlocklistSubmission action display strings, but in a shorter
+        form, adapted for displaying next to a single version."""
+        return {
+            BlocklistSubmission.ACTION_ADDCHANGE: 'Block',
+            BlocklistSubmission.ACTION_DELETE: 'Unblock',
+            BlocklistSubmission.ACTION_HARDEN: 'Harden',
+            BlocklistSubmission.ACTION_SOFTEN: 'Soften',
```
eviljeff, line 90: certainly not a hill I'd die on, but really these terms should all be defined in one place, imo

eviljeff (approved):
Just nits

diox:
@eviljeff went ahead and used extended choices in https://github.com/mozilla/addons-server/pull/22827/commits/c59d7425f09da5cf203153a012279e8b87e3b289 ; re-requesting approval because it's a large change (as you said, marginal benefits, but it does address the couple minor points you raised, so I figured it was worth a try and could be useful in the long run)

eviljeff (approved):
:shipit:

PR #22828, opened 2024-11-06
Soft-block-bloom-filter-filter

KevinMind:

### Checklist


src/olympia/blocklist/cron.py
```diff
@@ -68,72 +69,78 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
...
-            previous_filter is not None
-            and previous_filter.created_at < get_blocklist_last_modified_time()
-        )
-        or changes_count > 0
+        len(base_filters_to_update) > 0 or update_stash or previous_filter_is_stale
     )
     if not need_update:
         log.info('No new/modified/deleted Blocks in database; skipping MLBF generation')
-        return
+        return mlbf.delete()
```
willdurand, line 119: is this `delete()` call clearing the local (fs) cache?

src/olympia/blocklist/cron.py
```diff
@@ -68,72 +69,78 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
...
+        ):
+            update_stash = True
+
+    previous_filter_is_stale = (
+        previous_filter is not None
+        and previous_filter.created_at < get_blocklist_last_modified_time()
     )
+
     need_update = (
-        force_base
```
willdurand, line 99: So we cannot use `force_base` anymore?

src/olympia/blocklist/mlbf.py
```diff
@@ -207,32 +208,45 @@ def hash_filter_inputs(cls, input_list):
...
+        return self.storage.path(f'filter-{self.data_type_key(block_type)}')
 
     @property
     def stash_path(self):
         return self.storage.path('stash.json')
 
-    def generate_and_write_filter(self):
+    def delete(self):
+        if self.storage.exists(self.storage.base_location):
+            self.storage.rm_stored_dir(self.storage.base_location)
```
willdurand, line 220: Let's log a (debug or info) message here so that we know this deletion happened, unless the storage layer does it already.

src/olympia/blocklist/tasks.py
```diff
@@ -121,14 +111,64 @@ def upload_filter(generation_time, is_base=True):
...
             statsd.incr('blocklist.tasks.upload_filter.upload_stash')
 
+    oldest_base_filter_id = None
+
+    if len(filter_list) > 0:
+        for block_type in filter_list:
+            data = {
+                'key_format': MLBF.KEY_FORMAT,
+                'generation_time': generation_time,
+                'attachment_type': BLOCKLIST_RECORD_MLBF_BASE,
```
willdurand, line 121: For the soft block MLBF, this needs to be a different attachment type: `softblocks-bloomfilter-base` per https://searchfox.org/mozilla-central/rev/f732a1651018b7c32002981a9c8b8613975ffbf9/toolkit/mozapps/extensions/Blocklist.sys.mjs#908

src/olympia/blocklist/cron.py
KevinMind: Still [there](https://github.com/mozilla/addons-server/pull/22828/files#diff-2c1350128209aae0e622bd63c101e306f735752352fd0d5d8b8447785bd3a56fR93), but somewhere else in the code. There are tests verifying force_base recreates filters even if any filter should not have otherwise been recreated.

src/olympia/blocklist/cron.py
KevinMind: It is using the Storage class to delete the directory where all mlbf files for that id are stored. I'm not sure if that answers the question.. is there some cache in front of storage I need to know/worry about?

src/olympia/blocklist/tasks.py
KevinMind: What do we expect for the filename.. is it always `filter.bin` or should be based on the block_type? Currently it is different for each blocktype, but not sure the expectations.

src/olympia/blocklist/cron.py
willdurand: no, I am mostly interested in `storage/mlbf/*` stuff, which is another (probably pre-existing) bug I mentioned to you the other day.

src/olympia/blocklist/tasks.py
willdurand: The file name does not matter, the attachment type does, however.

src/olympia/blocklist/cron.py
KevinMind: self.storage in this case is pointing to `storage/mlbf/<current_id>` so this is deleting that, addressing the issue you raised.

KevinMind: Answer: yes

willdurand:
> Fixes: [mozilla/addons#15014](https://github.com/mozilla/addons/issues/15014)
> Fixes: [mozilla/addons#15166](https://github.com/mozilla/addons/issues/15166)
> Rleates to: [mozilla/addons#15155](https://github.com/mozilla/addons/issues/15155)

we should almost never fix two issues with a single PR, so please fix https://github.com/mozilla/addons/issues/15166 in a different PR.

KevinMind:
> > Fixes: [mozilla/addons#15014](https://github.com/mozilla/addons/issues/15014)
> > Fixes: [mozilla/addons#15166](https://github.com/mozilla/addons/issues/15166)
> > Rleates to: [mozilla/addons#15155](https://github.com/mozilla/addons/issues/15155)
> 
> we should almost never fix two issues with a single PR, so please fix [mozilla/addons#15166](https://github.com/mozilla/addons/issues/15166) in a different PR.

I've never heard of this rule and regularly do this. Why should that be a rule?

KevinMind:
In the interest of time, I moved to a separate PR https://github.com/mozilla/addons-server/pull/22870 @willdurand

KevinMind:
@willdurand dropped and closed as this PR introduces the reference to the new switch and so there is nothing to "update"

KevinMind:
Scenarios.

### Nothing to do

1. **Given**: No hard or soft blocks
2. **When**: [run the cron][link_cron_bash]
3. **Then**: 
- Expect successful run and no files created
- Expect [remote settings][link_remote_settings] to be empty

image

### Two hard blocked

1. **Given**: Two hard blocks since the last generation time
2. **When**: [run the cron](link_cron_bash)
3. **Then**: 
- Expect cache.json with two hard blocks
- Expect [remote settings][link_remote_settings] to contain a new attachment with a `blocked-bloomfilter-base` attatchment_type

image

### One hard blocked

1. **Given**:
- One hard block since the last generation
- There is already an uploaded bloom filter
2. **When**: [run the cron](link_cron_bash)
3. **Then**: 
- Expect cache.json with one hard block and stash.json with one hard block
- Expect [remote settings][link_remote_settings] to contain a new stash record with one `blocked` entry and an empty `unblocked` entry.

> if [softblocking][link_enable_softblocking] is enabled, then also expect an empty `softblocked` entry

image

### Two soft blocked

1. **Given**:
- Two softblocked entries since last generation time
- [softblockint][link_soft_blocking] is enabled
2. **When**: [run the cron](link_cron_bash)
3. **Then**: 
- Expect cache.json with 2 soft blocks and a `filter-softblocked` file
- Expect [remote settings[link_remote_settings] to contain a new attachment with `softblocked-bloomfilter-base` attatchment_type

### One soft blocked

1. **Given**:
- One soft blocked since last generation time
- [softblockint][link_soft_blocking] is enabled
3. **When**: [run the cron](link_cron_bash)
4. **Then**: 
- Expect cache.json and a `stash.json` file with one soft blocked entry
- Expect [remote settings[link_remote_settings] to contain a new record with one softblocked entry and empty unblocked/blocked entries

### One of each 

1. **Given**:
- One hard blocked since last generation time
- One soft blocked since last generation time
- [softblockint][link_soft_blocking] is enabled
3. **When**: [run the cron](link_cron_bash)
4. **Then**: 
- Expect cache.json and a stash.json file with the 2 blocked/softblocked entries respectively
- Expect [remote settings[link_remote_settings] to contain a new record with one softblocked entry, one blocked entry and empty unblocked entry

### Two of one, one of the other

1. **Given**:
- One hard blocked since last generation time
- Two soft blocked since last generation time
- [softblockint][link_soft_blocking] is enabled
3. **When**: [run the cron](link_cron_bash)
4. **Then**: 
- Expect cache.json and a stash.json file with the 1 hard blocked and a softblocked-filter file
- Expect [remote settings[link_remote_settings] to contain a new record with one blcoked entry,  and empty unblocked/blocked entry
Expect [remote settings[link_remote_settings] to contain a new attacthment with `softblocked-bloomfilter-base` attatchment_type

### Two of each

1. **Given**: 
- Two hard blocked since last generation time
- Two soft blocked since last generation time
- [softblockint][link_soft_blocking] is enabled
3. **When**: [run the cron](link_cron_bash)
4. **Then**: 
- Expect cache.json and, a softblocked-filter file and a blocked-filter file
Expect [remote settings[link_remote_settings] to contain a new attacthment with `softblocked-bloomfilter-base` attatchment_type
Expect [remote settings[link_remote_settings] to contain a new attacthment with `blocked-bloomfilter-base` attatchment_type

[link_remote_settings]: http://127.0.0.1:8888/v1/buckets/staging/collections/addons-bloomfilters/changeset?_expected=0
[link_cron_bash]: https://github.com/mozilla/addons-server/pull/22828#issuecomment-2488840809
[link_enable_softblocking]: https://github.com/mozilla/addons-server/pull/22828#issuecomment-2488840809

src/olympia/blocklist/mlbf.py
```diff
@@ -207,31 +208,44 @@ def hash_filter_inputs(cls, input_list):
...
+            log.info(f'Deleted {self.storage.base_location}')
+
+    def generate_and_write_filter(self, block_type: BlockType):
         stats = {}
 
+        blocked = self.data[block_type]
+
+        not_blocked = [
```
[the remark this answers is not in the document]
willdurand: Another thing I've noticed is that this generates a list of lists. I am not sure how that is encoded in the filter but I can see that the size of the filter is very small compared to what it used to be without this patch applied and, more serious, all add-ons seem to be blocked by Firefox with filters generated by this patch.

Let's flatten all that so that both `blocked` and `not_blocked` are plain lists of strings.

It isn't clear to me what this `verify()` method on the filter does, though, because we seem to have coverage for this `generate_and_write_filter` method.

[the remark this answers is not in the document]
willdurand: Try this maybe? That should address both comments above.

```diff
diff --git a/src/olympia/blocklist/mlbf.py b/src/olympia/blocklist/mlbf.py
index 26bd2264b6..14d51edf98 100644
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -224,12 +224,16 @@ class MLBF:
         stats = {}
 
         blocked = self.data[block_type]
-
-        not_blocked = [
-            self.data[_block_type]
-            for _block_type in BlockType
-            if _block_type != block_type
-        ]
+        not_blocked = (
+            self.data.not_blocked_items
+            + self.data[
+                (
+                    BlockType.BLOCKED
+                    if block_type == BlockType.SOFT_BLOCKED
+                    else BlockType.SOFT_BLOCKED
+                )
+            ]
+        )
 
         bloomfilter = generate_mlbf(
             stats=stats,
```

src/olympia/blocklist/tasks.py
```diff
@@ -88,28 +97,62 @@ def monitor_remote_settings():
...
+                    if 'generation_time' in record
+                    # If we don't have a generation time to use
+                    # then setting the base filter id will ensure
+                    # the record is not deleted.
+                    else oldest_base_filter_id
+                )
+                if record_time < oldest_base_filter_id:
+                    server.delete_record(record['id'])
+
+            cleanup_old_files.delay(base_filter_id=oldest_base_filter_id)
```
willdurand, line 151: Previous filters should also be deleted, and from what I can see locally, this isn't the case. We should only have a single record per filter.

willdurand (changes requested):
Thanks, that needs some changes but this looks promising.

src/olympia/blocklist/management/commands/export_blocklist.py
```diff
@@ -29,6 +30,11 @@ def add_arguments(self, parser):
             'the database',
             default=None,
         )
+        parser.add_argument(
+            '--block-type',
+            help='Block type to export',
+            default=None,
+        )
```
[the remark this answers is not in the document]
KevinMind: I didn't know we could do that. But sure.

src/olympia/blocklist/tasks.py
KevinMind: I'm not sure what you mean. This task is for cleaning up the local files and it only deletes them if they are both older than the given id and older than six months old. We can have multiple records per filter a `cache.json` a `stash.json` and or two filter files.

src/olympia/constants/blocklist.py
```diff
@@ -1,8 +1,15 @@
...
+
 BASE_REPLACE_THRESHOLD = 5_000
 
 # Config keys used to track recent mlbf ids
 MLBF_TIME_CONFIG_KEY = 'blocklist_mlbf_generation_time'
-MLBF_BASE_ID_CONFIG_KEY = 'blocklist_mlbf_base_id'
+
+
+def MLBF_BASE_ID_CONFIG_KEY(block_type: BlockType):
+    return f'blocklist_mlbf_base_id_{block_type.name.lower()}'
```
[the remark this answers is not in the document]
KevinMind: That's a good call out. We can manually set those values in our environments during the push and should just add that as a todo. Worst case scenario it would rcreate the base filters for both on the first run, then set the correct values.

src/olympia/blocklist/tasks.py
willdurand: ah sorry, I meant to comment where we do `server.delete_record()`.

src/olympia/blocklist/mlbf.py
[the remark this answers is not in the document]
KevinMind: I think that might not be quite enough. But I've expanded to a loop to be a bit more clear. Also there was a bug there as we need to include not just the BlockTypes but also the not blocked items in general.

willdurand (changes requested):
Some more feedback.

---

A note for myself too: I deleted enough hard blocks to force the generation of a new filter but it looks like only a stash was created. I am not entirely sure why.

src/olympia/blocklist/tasks.py
```diff
@@ -88,28 +97,62 @@ def monitor_remote_settings():
...
+                    if 'stash_time' in record
+                    else record['generation_time']
+                    if 'generation_time' in record
+                    # If we don't have a generation time to use
+                    # then setting the base filter id will ensure
+                    # the record is not deleted.
+                    else oldest_base_filter_id
+                )
+                if record_time < oldest_base_filter_id:
+                    server.delete_record(record['id'])
```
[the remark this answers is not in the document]
willdurand: Re-thinking about this - and I am not saying this isn't already what's happening - we should:

1. delete any filter that we are going to re-upload to keep always 1 filter (per block type)
2. delete everything when both filters are re-generated
3. delete stash records older than the oldest of the two filters

src/olympia/blocklist/mlbf.py
```diff
@@ -281,10 +313,27 @@ def generate_and_write_stash(self, previous_mlbf: 'MLBF' = None):
             stash_json['softblocked'] = soft_blocked_added
             stash_json['unblocked'] = [
                 unblocked
+                # unblocked should include any versions
+                # removed from blocked or soft blocked
```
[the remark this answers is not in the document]
willdurand: Re-thinking about this - and, again, not saying this isn't what's happening - we should create a stash record for blocks that are of a type whose filter isn't being regenerated. That sentence is horrible, ugh.

In practice, that would mean:

1. 1 hard block "change" and enough soft block changes to regenerate the soft block filter = we create a soft block filter and a stash that only contains this 1 hard block change (could we addition or deletion)
2. 1 soft block "change" and enough hard block changes to regenerate the hard block filter = we create a hard block filter and a stash that only contains this 1 soft block change (could we addition or deletion)
3. changes of soft and/or hard blocks that only lead to the creation of a stash = we only create a stash

src/olympia/blocklist/cron.py
```diff
@@ -64,76 +64,74 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
...
-    mlbf = MLBF.generate_from_db(generation_time)
-
-    base_filter = (
-        MLBF.load_from_storage(base_generation_time)
-        if base_generation_time is not None
-        else None
+    mlbf = MLBF.generate_from_db(get_generation_time())
+    previous_filter = MLBF.load_from_storage(
+        # This timestamp represents the last time the MLBF was generated and uploaded.
+        # It could have been a base filter or a stash.
```
willdurand, line 70: I suppose this answers one of my comments, though this isn't very accurate (a MLBF is a filter, not a stash). Not super important but it should probably say "last time we ran this task with block changes".

KevinMind: MLBF is a class which contains data about all blocks and not blocked versions, as well as methods to produce filters and or stashes... I think the name refers to the class and not the filter.

src/olympia/blocklist/tasks.py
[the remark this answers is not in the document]
KevinMind: - Added logic for 1. We check any pre-existing attachment that matches the `filter_list` and delete them.
- I did not include logic explicitly for 2 because it should happen anyway
- 3 was already implemented

src/olympia/blocklist/tasks.py
```diff
@@ -161,7 +161,16 @@ def upload_filter(generation_time, filter_list=None, create_stash=False):
...
+                # We subtract 1 from the stash time to ensure that FX
+                # considers all stashes to be older than base filters
+                # generated at the same time. The filter timestamp is
+                # tied to the MLBF creation time so we cannot change it.
+                # FX checks stashes before filters and only if a filter
+                # is more recent than the stash. So making the stash time
+                # 1 ms older than the filter ensures if an item is added
+                # via a new filter, and a stash was also created at the same
+                # time, the item will be found in the filter.
+                'stash_time': generation_time - 1,
```
willdurand, line 173: Let's reserve that kind of thing for https://github.com/mozilla/addons/issues/15200

src/olympia/blocklist/mlbf.py
```diff
@@ -287,23 +309,47 @@ def generate_and_write_stash(self, previous_mlbf: 'MLBF' = None):
...
+            stash_json[STASH_KEYS[block_type]] = added
+            # TODO: add a better test for this.
+            set_blocked_any_type.update(added)
+            # Items removed from that block type are included in the unblocked list
+            for item in removed:
+                # Items already on the hard block list should be excluded from the
+                # unblocked list to prevent double counting of a version that has moved
+                # from soft to hard blocked.
+                if item not in set_blocked_any_type:
+                    stash_json[UNBLOCKED_STASH_KEY].append(item)
```
willdurand, line 350: It's probably more of a note for myself than anything else: we should double check what stash we get when we have a block that goes from hard to soft AND enough other changes to make a new soft filter. In this PR, we should probably not account for "old" clients with the "super-mega-unified stash" approach documented in https://docs.google.com/document/d/1JmGUPD9fZ8ri_xdF1zBwOXsQHfBX-51fV2HYbBDsFQg/edit?usp=sharing

willdurand (changes requested):
I am done for today, but that makes it much easier to see what's going on, thanks.

To be continued...

src/olympia/blocklist/mlbf.py
```diff
@@ -298,6 +348,26 @@ def blocks_changed_since_previous(
...
+        )
+
+    def should_upload_stash(
+        self, block_type: BlockType = BlockType.BLOCKED, previous_mlbf: 'MLBF' = None
+    ):
+        return (
+            self.blocks_changed_since_previous(
+                block_type=block_type, previous_mlbf=previous_mlbf
+            )
+            > 0
```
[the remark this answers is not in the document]
KevinMind: We check them independently. I'd rather not inject the logic for updating filter to stash.

src/olympia/blocklist/mlbf.py
KevinMind: I'll add a test for this tomorrow.

src/olympia/blocklist/mlbf.py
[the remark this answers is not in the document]
willdurand: Fine. I had this comment last week in fact, but forgot to submit it.

src/olympia/blocklist/mlbf.py
KevinMind: I added the test and, perhaps interestingly, We might be thinking about this in a less than accurate way. When we run the cron, and a block has moved from hard to soft, and there are still (or now) enough soft blocks for a new filter, what happens is:

- if soft blocking is enabled, then we get a new soft blocked filter which will include the soft blocked items.
- else we don't get a soft block filter

What happens next depends on the number of hard blocks at the time of generation:

- if there are enough hard blocks for a filter, we get 2 filters
- else if there are enough hard blocks for a stash, we get a stash with those hard blocks and an empty soft block
- else we get no stash because there are no hard blocks and soft blocks are either ignored due to waffle switch or included in the filter instead due to the number of changes

Because there are (at the time of generation) enough soft blocks for a new filter we get the filter or not depending on the state of the waffle switch. The expected stash depends on the number of hard blocks and (partially) on the state of the waffle switch.

src/olympia/blocklist/cron.py
```diff
@@ -111,13 +116,21 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
...
+        len(mlbf.data.soft_blocked_items),
+    )
     statsd.incr(
         'blocklist.cron.upload_mlbf_to_remote_settings.not_blocked_count',
         len(mlbf.data.not_blocked_items),
     )
 
     if create_stash:
-        mlbf.generate_and_write_stash(previous_filter)
+        mlbf.generate_and_write_stash(
```
willdurand, line 129:
```suggestion
        # We generate unified stashes, which means they can contain data
        # for both soft and hard blocks. We need the base filters of each
        # block type to determine what goes in a stash.
        mlbf.generate_and_write_stash(
```

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -538,49 +525,215 @@ def test_upload_stash_unless_enough_changes(self):
...
+            enable_soft_blocking=True,
+            expected_stash={
+                'blocked': 1,
+                # Expect no soft blocks becuase there are enough for a filter
+                'softblocked': 0,
+                'unblocked': 0,
+            },
+            # Expect a soft blocked filter
+            expected_filters=[BlockType.SOFT_BLOCKED],
+        )
```
willdurand, line 628: Aren't these two tests the same? I don't think we need `test_upload_blocked_stash_and_softblock_filter_with_switch_enabled`, it is covered by the second call to `_test_upload_stash_and_filter` in `test_upload_blocked_stash_and_softblock_filter`, no?

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -538,49 +525,215 @@ def test_upload_stash_unless_enough_changes(self):
...
+                # and soft blocking is enabled
+                'softblocked': 1,
+                # There are no unblocked versions
+                'unblocked': 0,
+            },
+            # Expect a blocked filter
+            expected_filters=[BlockType.BLOCKED],
+        )
+
+    def test_upload_soft_blocked_stash_and_blocked_filter_2(self):
```
willdurand, line 663: Please find a better name :D

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -538,49 +525,215 @@ def test_upload_stash_unless_enough_changes(self):
...
+        self._block_version(is_signed=True, block_type=BlockType.SOFT_BLOCKED)
```
willdurand, line 664:
```suggestion
        # Enough soft blocked versions for a filter
        self._block_version(is_signed=True, block_type=BlockType.SOFT_BLOCKED)
```

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -538,49 +525,215 @@ def test_upload_stash_unless_enough_changes(self):
...
+        self._block_version(is_signed=True, block_type=BlockType.SOFT_BLOCKED)
+        self._block_version(is_signed=True, block_type=BlockType.BLOCKED)
```
willdurand, line 666:
```suggestion
        # Enough hard blocked versions for a filter
        self._block_version(is_signed=True, block_type=BlockType.BLOCKED)
```

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -538,49 +525,215 @@ def test_upload_stash_unless_enough_changes(self):
...
+
+        self._test_upload_stash_and_filter(
+            enable_soft_blocking=True,
+            expected_stash={
+                # Expect a blocked version now,
+                # because we converted a hard block to a soft block
```
willdurand, line 672:
```suggestion
                # Expect a blocked version
```

willdurand (changes requested):
Some more feedback. This is coming together.

src/olympia/blocklist/tests/test_cron.py
KevinMind: How about `test_upload_soft-blocked_stash_and_blocked_filter_deux`? 🙃

KevinMind: This test is a dupe of `test_upload_blocked_stash_and_softblock_filter` just removing it.

src/olympia/blocklist/mlbf.py
```diff
@@ -262,46 +285,59 @@ def generate_diffs(
...
+        excluded from the unblocked list to prevent double counting.
 
-        Versions that move from soft to hard blocked will be picked up by
-        all clients in the blocked list. Note, even though the version is removed
-        from the soft blocked list, it is important that we do not include it
-        in the "unblocked" stash (like for hard blocked items) as this would
-        result in the version being in both blocked and unblocked stashes.
+        If a block type needs a new filter, we do not include any items for that
+        block type in the stash to prevent double counting items.
         """
```
willdurand, line 308:
```suggestion

        We used to generate a list of `unblocked` versions as a union of deletions 
        from blocked and deletions from soft_blocked, filtering out any versions
        that are in the newly blocked list in order to support Firefox clients that
        don't support soft blocking. That, unfortunately, caused other issues so
        currently we are very conservative, and we do not fully support old clients.
        See: https://github.com/mozilla/addons/issues/15208
        """
```

willdurand (approved):
r+wc, thanks!

willdurand:
(note: it'd be good if you could clear the commit message before merging)

PR #22830, opened 2024-11-07
Split diff by block type

KevinMind:
Relates to: https://github.com/mozilla/addons-server/pull/22826
Relates to: https://github.com/mozilla/addons-server/pull/22828
Fixes: mozilla/addons#15147

### Description

Split `generate_diffs` to include diffs for both hard and soft blocks. This is required for both stashing and filter updates.

### Context

This patch is required by both of the linked PRs and to keep them independent we should land this (trivial) change first and rebase the other two.

### Testing

This is a backaward compatible patch so automated testing is fine.

### Checklist

### DIFF
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -135,6 +135,7 @@ class MLBFDataBaseLoader(BaseMLBFLoader):
     def _all_blocks(self):
         return (
             BlockVersion.objects.filter(version__file__is_signed=True)
+            .distinct()
             .order_by('id')
             .values_list(
                 'block__guid',
@@ -167,6 +168,7 @@ class MLBFDataBaseLoader(BaseMLBFLoader):
         all_blocks_ids = [version.version_id for version in self._all_blocks]
         not_blocked_items = MLBF.hash_filter_inputs(
             Version.unfiltered.exclude(id__in=all_blocks_ids or ())
+            .distinct()
             .order_by('id')
             .values_list('addon__addonguid__guid', 'version')
         )
@@ -258,7 +260,8 @@ class MLBF:
     def blocks_changed_since_previous(
         self, block_type: BlockType = BlockType.BLOCKED, previous_mlbf: 'MLBF' = None
     ):
-        return self.generate_diffs(previous_mlbf)[block_type][2]
+        _, _, changed_count = self.generate_diffs(previous_mlbf)[block_type]
+        return changed_count
 
     @classmethod
     def load_from_storage(
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -2,7 +2,7 @@ import json
 import os
 import secrets
 from enum import Enum
-from typing import Dict, List, Optional, Set, Tuple
+from typing import Dict, List, Optional, Tuple
 
 from django.utils.functional import cached_property
 
@@ -19,11 +19,12 @@ from olympia.versions.models import Version
 log = olympia.core.logger.getLogger('z.amo.blocklist')
 
 
-def diff_lists(
+def ordered_diff_lists(
     previous: List[str], current: List[str]
-) -> Tuple[Set[str], Set[str], int]:
-    extras = set(current) - set(previous)
-    deletes = set(previous) - set(current)
+) -> Tuple[List[str], List[str], int]:
+    # Use lists instead of sets to maintain order
+    extras = [x for x in current if x not in previous]
+    deletes = [x for x in previous if x not in current]
     changed_count = len(extras) + len(deletes)
     return extras, deletes, changed_count
 
@@ -172,9 +173,11 @@ class MLBFDataBaseLoader(BaseMLBFLoader):
         # even though we exclude all the version ids in the query there's an
         # edge case where the version string occurs twice for an addon so we
         # ensure not_blocked_items contain no blocked_items or soft_blocked_items.
-        return list(
-            set(not_blocked_items) - set(self.blocked_items + self.soft_blocked_items)
-        )
+        return [
+            item
+            for item in not_blocked_items
+            if item not in self.blocked_items + self.soft_blocked_items
+        ]
 
 
 class MLBF:
@@ -230,9 +233,9 @@ class MLBF:
 
     def generate_diffs(
         self, previous_mlbf: 'MLBF' = None
-    ) -> Dict[BlockType, Tuple[Set[str], Set[str], int]]:
+    ) -> Dict[BlockType, Tuple[List[str], List[str], int]]:
         return {
-            block_type: diff_lists(
+            block_type: ordered_diff_lists(
                 [] if previous_mlbf is None else previous_mlbf.data[block_type],
                 self.data[block_type],
             )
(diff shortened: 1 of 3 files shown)


src/olympia/blocklist/mlbf.py
```diff
@@ -19,6 +19,15 @@
 log = olympia.core.logger.getLogger('z.amo.blocklist')
 
 
+def diff_lists(
+    previous: List[str], current: List[str]
+) -> Tuple[Set[str], Set[str], int]:
+    extras = set(current) - set(previous)
+    deletes = set(previous) - set(current)
+    changed_count = len(extras) + len(deletes) if len(previous) > 0 else len(current)
```
eviljeff, line 27:
```suggestion
    changed_count = len(extras) + len(deletes)
```
the `if` is redundant?

KevinMind: Done.

src/olympia/blocklist/mlbf.py
```diff
@@ -148,16 +166,18 @@ def soft_blocked_items(self) -> List[str]:
...
         # even though we exclude all the version ids in the query there's an
         # edge case where the version string occurs twice for an addon so we
         # ensure not_blocked_items contain no blocked_items or soft_blocked_items.
-        return list(
-            set(not_blocked_items) - set(self.blocked_items + self.soft_blocked_items)
-        )
+        return [
+            item
+            for item in not_blocked_items
+            if item not in self.blocked_items + self.soft_blocked_items
```
willdurand, line 179: It looks like the second branch in this condition is not covered by tests.

With this diff applied, I would expect some tests to break but `test_mlbf.py` passes.
```diff
diff --git a/src/olympia/blocklist/mlbf.py b/src/olympia/blocklist/mlbf.py
index b5e7bbbfad..ed8d658452 100644
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -176,7 +176,7 @@ class MLBFDataBaseLoader(BaseMLBFLoader):
         return [
             item
             for item in not_blocked_items
-            if item not in self.blocked_items + self.soft_blocked_items
+            if item not in self.blocked_items # + self.soft_blocked_items
         ]
```

src/olympia/blocklist/mlbf.py
```diff
@@ -148,16 +166,18 @@ def soft_blocked_items(self) -> List[str]:
...
-            )
+            Version.unfiltered.exclude(id__in=all_blocks_ids or ())
+            .order_by('id')
+            .values_list('addon__addonguid__guid', 'version')
         )
         # even though we exclude all the version ids in the query there's an
         # edge case where the version string occurs twice for an addon so we
         # ensure not_blocked_items contain no blocked_items or soft_blocked_items.
-        return list(
-            set(not_blocked_items) - set(self.blocked_items + self.soft_blocked_items)
```
willdurand, line 159: I am a bit nervous about getting rid of the sets because of https://github.com/mozilla/addons/issues/15132 but OTOH `not_blocked_items` should only contain unique entries already.

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -141,13 +143,32 @@ def test_load_returns_expected_data(self):
...
+            [
+                (block_version.block.guid, block_version.version.version)
+                for block_version in blocked_versions
+            ]
+        )
+        assert mlbf_data[MLBFDataType.SOFT_BLOCKED] == MLBF.hash_filter_inputs(
+            [
+                (soft_blocked_version.block.guid, soft_blocked_version.version.version)
+                for soft_blocked_version in soft_blocked_versions
+            ]
```
willdurand, line 171: can we assert the `len()` as well? it should 10 for each type, right?

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -193,6 +214,47 @@ def test_get_data_from_db(self):
         mlbf_data = MLBFDataBaseLoader(mlbf.storage)
         assert mlbf.data._raw == mlbf_data._raw
 
+    def test_cache_json_is_sorted(self):
```
willdurand, line 217: I don't think this test is quite right because if I apply it to current `master`, it passes.

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -300,44 +365,105 @@ def test_diff_block_added_and_removed(self):
...
         with next_mlbf.storage.open(next_mlbf.stash_path, 'r') as f:
             assert json.load(f) == {
                 'blocked': [],
                 'unblocked': MLBF.hash_filter_inputs(
-                    [(block_version.block.guid, block_version.version.version)]
+                    [
+                        (block_version.block.guid, block_version.version.version)
+                        for i, block_version in enumerate(block_versions)
+                        if i % 2 == 0
+                    ]
```
willdurand, line 466: This is very hard to follow what's going on and what is being asserted. I'd suggest you delete one or two `block_version` before generating the next mlbf, and then you use that instead, but without a loop or modulo.

willdurand (changes requested):
This is looking promising. I have some comments.

src/olympia/blocklist/tests/test_mlbf.py
KevinMind: The len is asserted via the equality check but I cleaned up the code a bit to make it more obbvious by using a single 10 length loop

src/olympia/blocklist/mlbf.py
KevinMind: I tried reproducing this and both are covered by the duplicate test as we create a hard and soft block from the same reused addon

src/olympia/blocklist/tests/test_mlbf.py
KevinMind: If you change the order by in the queries the test fails.

src/olympia/blocklist/mlbf.py
KevinMind: I was looking at this and I'm not even sure how I would test it... I added `distinct()` to the queries but that's kind of moot as we have unique indexes for  all the relevant combinations.

src/olympia/blocklist/tests/test_mlbf.py
willdurand: Do the queries on `master` have ORDER BY clauses? The test doesn't capture the "broken" behavior that this PR is supposed to fix. I would expect a failing test on `master` and a passing test with this PR.

willdurand (approved):
[r+wc](https://forum.bors.tech/t/what-does-r-wc-mean-within-the-mozreview-system/101), thanks

src/olympia/blocklist/tests/test_mlbf.py
KevinMind: They do not.. and for me at least it fails on master. It could be that this test is actually flaky on master.. if we don't specify ordering it _could_ be returned in the right order and the fewer number of items in the list the higher the probability would be.

This is from latest master right now.

image

KevinMind: Currently running a 100 loop on it and it is indeed flaky. Sometimes passes.

PR #22832, opened 2024-11-07
Add command to delete erroneously added overgrowth NeedsHumanReview & reset due dates

diox:
Relates to mozilla/addons#15141

### Context

We introduced a new way to determine if an add-on should be flagged for abnormal growth but the new algorithm has a bug that caused a bunch of add-ons to be flagged erroneously. Operations want to delete the `NeedsHumanReview` flag from them all (can't just make it inactive, that would skew data in our stats about reviews) and reset their due dates to remove them from the queue.

Note that the algorithm hasn't been fixed yet, because we still need to agree on what the solution is - but we want to clean up the queue quickly before reviewers get to it.

### Testing

Unit test should be enough. We should only remove the NHR that were created for that specific reason, during the specific timeframe, that are still active.

### DIFF
--- a/src/olympia/addons/tasks.py
+++ b/src/olympia/addons/tasks.py
@@ -1,6 +1,7 @@
 import hashlib
 import mimetypes
 import os
+from datetime import date
 
 from django.core.files.storage import default_storage as storage
 from django.db import transaction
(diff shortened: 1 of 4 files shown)


src/olympia/addons/tasks.py
```diff
@@ -563,3 +564,26 @@ def flag_high_hotness_according_to_review_tier():
...
+    addons = Addon.unfiltered.filter(pk__in=addon_ids).no_transforms()
+    for addon in addons:
+        log.info(
+            'Deleting erroneously added NHR and updating due dates for %s', addon.pk
+        )
+        NeedsHumanReview.objects.filter(
+            version__addon=addon,
+            reason=NeedsHumanReview.REASONS.HOTNESS_THRESHOLD,
+            created__range=ERRONEOUSLY_ADDED_OVERGROWTH_DATE_RANGE,
+            is_active=True,
```
eviljeff, line 587: I'm not a fan of how we have to query on the same things in two places but I can't immediately think of a way around it 😒

eviljeff (approved):

src/olympia/addons/tasks.py
diox: Same. I could have generated some dictionaries of filters with an optional `versions__` prefix but I figured this wasn't worth doing for a one-off.

PR #22833, opened 2024-11-07
Add add-on created datetime as a column to the content review queue

bakulf:
Fixes: mozilla/addons#15077

### Description

This PR adds the creation date time to the content-review queue and make orderable this and the last-updated column.
It does not remove the 'flag' column because this seems a task for a separate bug.

### Testing

No extra tests added. I did not find tests for the columns in that queue. If there are, ping me and I'm happy to write a test.

### Checklist

Before:
image

After:
image

### DIFF
--- a/src/olympia/reviewers/utils.py
+++ b/src/olympia/reviewers/utils.py
@@ -206,16 +206,17 @@ class PendingRejectionTable(AddonQueueTable):
 
 class ContentReviewTable(AddonQueueTable):
     last_updated = tables.DateTimeColumn(verbose_name='Last Updated')
+    created = tables.DateTimeColumn(verbose_name='Created')
     title = 'Content Review'
     urlname = 'queue_content_review'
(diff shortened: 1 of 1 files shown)


diox (approved):
> Fixes: mozilla/addons/15077

Should be Fixes: mozilla/addons#15077 to link properly (`#` instead of `/`)

> No extra tests added. I did not find tests for the columns in that queue. If there are, ping me and I'm happy to write a test.

We don't have tests specifically for the columns in that queue, you're right (`TestContentReviewQueue` will test that specific add-ons are present using their names, so in a way the first column contents are tested though). This is probably fine for now, unless QA finds something weird.

Before merging, could you check with ops (@abhn or @wagnerand) how we should render the created date ?

diox:
(Also, remove "WIP" from the title)

bakulf:
image

PR #22834, opened 2024-11-07
Ensure enable-submissions Disabling Correctly Allows Bypassing

chrstinalin:
Fixes: mozilla/addons#15096

### Checklist

### DIFF
--- a/src/olympia/addons/views.py
+++ b/src/olympia/addons/views.py
@@ -399,7 +399,6 @@ class AddonViewSet(
             self.action = 'create'
             return self.create(request, *args, **kwargs)
 
-    @require_submissions_enabled
     @swagger_auto_schema(
         operation_description="""
             This endpoint allows a submission of an upload to create a new add-on
@@ -414,6 +413,9 @@ class AddonViewSet(
         """
     )
     def create(self, request, *args, **kwargs):
+        return require_submissions_enabled(self._create)(request, *args, **kwargs)
+
+    def _create(self, request, *args, **kwargs):
         response = super().create(request, *args, **kwargs)
         return response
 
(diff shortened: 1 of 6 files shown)


src/olympia/addons/views.py
```diff
@@ -414,6 +413,9 @@ def update(self, request, *args, **kwargs):
         """
     )
     def create(self, request, *args, **kwargs):
```
KevinMind, line 415: why can't this be a decorator? Seems odd..

chrstinalin: Authentication isn't being performed prior to when its used as a decorator, for some reason, so when its used as a decorator, `request.user` doesn't exist (reason the issue was filed in the first place) and so the API is disabled for everyone regardless of whether they have access via `users` or `superusers`.

KevinMind: I think we should try and figure out the root cause. If it turns out fixing the root cause is a huge headache, then we can ship this. But let's resolve the `for some reason` to `for reason X which is difficult to fix` first.

We can try to debug this together next week if you'd like.

diox: I suggested in the issue this could be a problem where the auth isn't performed at the right time, which can happen because DRF auth isn't performed through the regular Django middleware (*), but that's actually not the issue here.

The problem is simpler: you should be using `@method_decorator(require_submissions_enabled)`, otherwise the first parameter that the decorator sees is not a request, but the viewset instance instead.

(*) DRF instead does authentication following the `authentication_classes` of the view(set) instead, in a `perform_authentication()` method that is called fairly late. If you decorate `dispatch()` on a viewset, you'll run into this.

KevinMind: @chrstinalin will you update based on @diox suggestion? It's not totally clear to me the status/plan from here.

chrstinalin: @KevinMind I had made the commit at the same time you sent that 😆

PR #22846, opened 2024-11-13
Fix-missing-cache

KevinMind:
This reverts commit 2bd0783eb5218008b7420eb86c49b8a043de5c3e.

Fixes: mozilla/addons#15163

### Context

After https://github.com/mozilla/addons-server/pull/22715 we introduced `soft_blocked` key to the cache.json, however the first time we ran the filter it compared to a cache.json that did not have soft_blocked and this led to failures when generating diffs.

This should really never happen after the first iteration but it's good to have this fallback in place if for any reason we have a broken cache.json

### Testing

1. create a bloom filter
2. manually delete one of the keys from cache.json
3. create another bloom filter
4. generate a diff from the first filter and expect an empty array for the key that is missing

### Checklist


src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -133,6 +133,15 @@ def test_loads_data_from_file(self):
...
         assert loader._raw == self._data
 
+    def test_fallback_to_empty_list_for_missing_key(self):
+        for key in self._data.keys():
+            new_data = self._data.copy()
+            new_data.pop(key)
+            with self.storage.open('cache.json', 'w') as f:
+                json.dump(new_data, f)
+            loader = MLBFStorageLoader(self.storage)
+            assert loader._raw == {**new_data, key: []}
```
willdurand, line 143: I provided a test on slack that reproduces the bug found in Sentry, please add it. While this test verifies the underlying storage loader and some of its properties, that doesn't explain why we have a `NoneType` elsewhere.

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -133,6 +133,15 @@ def test_loads_data_from_file(self):
         loader = MLBFStorageLoader(self.storage)
         assert loader._raw == self._data
 
+    def test_fallback_to_empty_list_for_missing_key(self):
+        for key in self._data.keys():
+            new_data = self._data.copy()
+            new_data.pop(key)
```
willdurand, line 139:
```suggestion
            # Generate a `cache.json` file with a missing key
            # (we do this for each key).
            new_data = self._data.copy()
            new_data.pop(key)
```

willdurand (changes requested):
Add the regression test and we should be good to go.

src/olympia/blocklist/tests/test_mlbf.py
KevinMind: Added the test, even added an additional assumption we should have.. if there is a corrupted cache that _would_ have included a block, we should expect that block to be included in the diff because the previous data that would have filtered it out is missing.

willdurand: I would still add the test case that was derived from the actual bug.

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -452,6 +463,28 @@ def test_diff_block_soft_to_hard(self):
             ),
         }
 
+    def test_diff_invalid_cache(self):
```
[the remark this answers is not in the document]
KevinMind: It's not important which key is missing. You could reproduce the bug by removing "blocked" or "not_blocked", same error, regardless of which is missing.

I used an empty one to show the most extreme case where all keys are missing. Hard coding to test only if soft_blocked is missing is a lower coverage test.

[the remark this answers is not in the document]
KevinMind: I'll update it to drop the soft block, but actually create a soft block as well so at least we still validate the results of the diff.

PR #22853, opened 2024-11-14
Restrict the /64 for IPv6 when adding an automatic IP restriction from a scanner action

diox:
Fixes: mozilla/addons#15123

### Description

See issue: the `_delay_auto_approval_indefinitely_and_restrict` scanner action, for users on IPv6, was adding a `/32` restriction and doing it while keeping the hosts bits in the IP, both of which lead to an incorrect restriction.

### DIFF
--- a/src/olympia/scanners/tests/test_actions.py
+++ b/src/olympia/scanners/tests/test_actions.py
@@ -392,6 +392,45 @@ class TestActions(TestCase):
                 f'Addon {addon.pk} Version {addon.current_version.pk}.'
             )
 
+    def test_delay_auto_approval_indefinitely_and_restrict_with_ipv6(self):
+        user1 = user_factory(last_login_ip='2001:0db8:4815:1623:4200:1337:cafe:d00d')
+        user2 = user_factory(last_login_ip='')
+        user3 = user_factory()
+        addon = addon_factory(users=[user1, user2])
+        FileUpload.objects.create(
+            addon=addon,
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/tests/test_actions.py
```diff
@@ -392,6 +392,45 @@ def test_delay_auto_approval_indefinitely_and_restrict(self):
                 f'Addon {addon.pk} Version {addon.current_version.pk}.'
             )
 
+    def test_delay_auto_approval_indefinitely_and_restrict_with_ipv6(self):
```
eviljeff, line 395: nit, but this is a copy and paste of the previous test while not _just_ testing ipv6 - only one of the users has a ipv6 last_login_ip.  I don't have an easy suggestion of what to change but my "feeling" is the tests need restructuring/de-duplicating/refactoring.

eviljeff (approved):
r+ with a nit.

src/olympia/scanners/tests/test_actions.py
diox: Yeah, I also felt this wasn't ideal. I tried to remove extra cruft and focus only on the network restriction part while still keeping some non-ipv6 data to prove it didn't impact anything else. I could have re-used on the existing tests instead of creating a new one but I felt it was going to make it more confusing... Ultimately this does demonstrate the problem & fix

PR #22859, opened 2024-11-15
More efficient diff function removing quadratic iterable

KevinMind:
Fixes: mozilla/addons#15170

### Testing

Automated test should cover it.. the test without sets hangs forever

### Checklist

### DIFF
--- a/src/olympia/blocklist/tests/test_mlbf.py
+++ b/src/olympia/blocklist/tests/test_mlbf.py
@@ -48,7 +48,7 @@ class _MLBFBase(TestCase):
         )
 
 
-class TestOrderedDiffLists(_MLBFBase):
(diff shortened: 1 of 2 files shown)


src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -47,6 +48,27 @@ def _block_version(self, block, version, block_type=BlockType.BLOCKED):
         )
 
 
+class TestOrderedDiffLists(_MLBFBase):
```
diox, line 51: Don't need to depend on `_MLBFBase`, it could just be `TestCase` (or `SimpleTestCase` since you don't do database stuff, though we tend not to use that one)

diox (approved):

PR #22860, opened 2024-11-15
Fix data_seed migration data being skipped

KevinMind:
Fixes: mozilla/addons#15153

### Description

Makes sure if we are going to run `data_seed` in the initialize command that we do not pre migrate the database as this could lead to data being dropped post migration.

Additionally, use `reset_db` instead of `flush` as this will ensure the entire DB is dropped/created ensuring migrations actually run afterward.

Minor cleanup removing redundant uneeded code.

### Context

When running `data_seed` there are 2 issues. 1) migrations have already been run by `initialize` this can lead to migrations that should run not being run again after flushing the db 2) flushing the db does not reset migrations so if they already ran, will not run again.

This can lead to a scenario where the database migrates data and then removes it, and the second migration does not reintroduce the data.

### Testing

Run clean up

> **WARNING**: this will wipe your current database state.

```bash
make up INIT_CLEAN=True
```
Now in a shell run

```
./manage waffle_switch --list
```

Expect this output including waffle switches generated during migrations.

```bash
run-action-in-auto-approve: off
enable-mad: off
allow-deleted-guid-reuse: off
dsa-job-technical-processing: off
dsa-abuse-reports-review: off
dsa-appeals-review: on
dsa-cinder-forwarded-review: off
enable-soft-blocking: off
suppressed-email: off
enable-taar: on
enable-manifest-normalization: off
enable-mv3-submissions: off
enable-git-extraction-cron: off
enable-subscriptions-for-promoted-addons: off
record-install-origins: off
enable-cinder-reviewer-tools-integration: off
enable-activity-log-attachments: off
disable-linter-xpi-autoclose: off
return-to-amo-for-all-listed: off
disable-bigquery: off
```

You can verify other data created during migrations like `License.objects.count()` should return 26 licenses.

### Checklist


src/olympia/amo/management/commands/data_seed.py
```diff
@@ -13,56 +10,25 @@ class Command(BaseDataCommand):
...
         self.clean_storage()
-        # reindex --wipe will force the ES mapping to be re-installed.
-        call_command('reindex', '--wipe', '--force', '--noinput')
+        # Migrate the database
         call_command('migrate', '--noinput')
 
         self.logger.info('Loading initial data...')
         call_command('loaddata', 'initial.json')
         call_command('import_prod_versions')
-        call_command('import_licenses')
```
eviljeff, line 65: Can you link to the pr/issue where we made this command redundant?  (I thought Baku's work was still in progress)

src/olympia/core/apps.py
```diff
@@ -98,6 +99,26 @@ def static_check(app_configs, **kwargs):
     return errors
 
 
+@register(CustomTags.custom_setup)
+def db_charset_check(app_configs, **kwargs):
```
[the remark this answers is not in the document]
KevinMind: Because we are now using `reset_db` we rely on django-extensions to create the Database. This ensures that when that happens, it has the correct charset.

src/olympia/amo/management/commands/data_seed.py
KevinMind: <https://github.com/mozilla/addons-server/pull/22818>

eviljeff: @KevinMind this is the patch that added the line, 2 days ago.  How is it redundant after 2 days - it's not even be deployed to stage yet?

PR #22861, opened 2024-11-15
Remove all time inefficient loops from mlbf

KevinMind:
Fixes: mozilla/addons#15170

### Description

- extract the database querying logic from `MLBFDatabaseLoader` to module scope functions for easier mocking
- reduce number of loops required to build initial dataset in `MLBFDatabaseLoader` to 1
- filter `not_blocked_items` with an O(1) set instead of list
- refactor `generate_diffs` to accept a block_type and only diff one set of data at a time.
- add an lru_cache to `generate_diffs` to ensure computation is executed only one time per instance + block_type + previous_mlbf
- collect all mlbf tests under `test_mlbf` mark 

### Context

This patch includes logic that should reduce the number of long loops required to process our mlbf data sets and determine the cache and diff of blocks across different scenarios.

### Run the mlbf suite

```bash
make test_mlbf
```

This will run all of the mlbf related tests. Expect everything passes

### Test the timeout for the slow test

1. modify the `pytest.mark.timeout(120)` reducing the time to 3 (or some low number
2. run the test

```bash
pytest ./src/olympia/blocklist/tests/test_cron.py::TestUploadToRemoteSettings::test_large_data_set
```

3. expect the test to fail due to timeout

```bash
FAILED src/olympia/blocklist/tests/test_cron.py::TestUploadToRemoteSettings::test_large_data_set - Failed: Timeout >10.0s
```

### Test in local environment

IF you want to test with a real data set you will have to create a large amount of blocks and execute the cron

[Code snippet](https://github.com/mozilla/addons/issues/15171#issuecomment-2479388403)

Execute this code. You will need to execute many time or bump `1_000 up to between 1 and 2 million.. grab a coffee this will take a while.

After that is finished, execute `./manage.py cron upload_mlbf_to_remote_settings` and expect it to take up to 2 minutes to complete.

### Checklist

### DIFF
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -2,7 +2,8 @@ import json
 import os
 import secrets
 from enum import Enum
-from typing import Dict, List, Optional, Tuple
+from functools import lru_cache
+from typing import List, Optional, Tuple
 
 from django.utils.functional import cached_property
 
@@ -21,15 +22,47 @@ log = olympia.core.logger.getLogger('z.amo.blocklist')
 
 
 def ordered_diff_lists(
-    previous: List[str], current: List[str]
+    current: List[str],
+    previous: List[str],
 ) -> Tuple[List[str], List[str], int]:
-    current_set = set(current)
     previous_set = set(previous)
     # Use lists instead of sets to maintain order
-    extras = [x for x in current if x not in previous_set]
-    deletes = [x for x in previous if x not in current_set]
-    changed_count = len(extras) + len(deletes)
-    return extras, deletes, changed_count
+    return [x for x in current if x not in previous_set]
+
+
+def get_not_blocked_items(all_blocked_version_ids: List[int]):
+    """
+    Returns a list of tuples containing the guid, version of all not blocked
+    versions. We use distinct to avoid duplicates, order by ID to ensure
+    cache.json is always sorted consistently, and return the values as a tuple
+    to make it easier to mock in tests.
+    """
+    return (
+        Version.unfiltered.exclude(id__in=all_blocked_version_ids or ())
+        .distinct()
+        .order_by('id')
+        .values_list('addon__addonguid__guid', 'version')
+    )
+
+
+def get_all_blocked_items():
+    """
+    Returns a list of tuples containing the guid, version, version_id, and
+    block_type of all blocked items. We use distinct to avoid duplicates,
+    Order by ID to ensure cache.json is always sorted consistently,
+    and return the values as a tuple to make it easier to mock in tests.
+    """
+    return (
+        BlockVersion.objects.filter(version__file__is_signed=True)
+        .distinct()
+        .order_by('id')
+        .values_list(
+            'block__guid',
+            'version__version',
+            'version_id',
+            'block_type',
+        )
+    )
 
 
 def generate_mlbf(stats, blocked, not_blocked):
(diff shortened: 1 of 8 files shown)


willdurand:
Thanks for adding the new tests, that looks good at first glance. I'm going to focus on #22828 so I'll let someone else take care of this PR.

src/olympia/blocklist/mlbf.py
```diff
@@ -21,15 +22,47 @@
 
 
 def ordered_diff_lists(
```
eviljeff, line 24: Can we just use sets instead? They're much more efficient for comparison, and, to me, sticking to sets everywhere would be more straightforward.

We only need a specific order when we're writing to JSON, (because it doesn't support sets), so we could just sort the data just before writing.

src/olympia/blocklist/mlbf.py
```diff
@@ -21,15 +22,47 @@
...
     previous_set = set(previous)
     # Use lists instead of sets to maintain order
-    extras = [x for x in current if x not in previous_set]
-    deletes = [x for x in previous if x not in current_set]
-    changed_count = len(extras) + len(deletes)
-    return extras, deletes, changed_count
+    return [x for x in current if x not in previous_set]
+
+
+def get_not_blocked_items(all_blocks_ids: List[int]) -> List[Tuple[str, str]]:
```
eviljeff, line 33: strictly, this type definition isn't correct - the return is a type of queryset rather than `List`, unless the function is meant to cast the queryset into a `list(...)` before returning.

src/olympia/blocklist/mlbf.py
```diff
@@ -21,15 +22,47 @@
...
+    """
+    return (
+        Version.unfiltered.exclude(id__in=all_blocks_ids or ())
+        .distinct()
+        .order_by('id')
+        .values_list('addon__addonguid__guid', 'version')
+    )
+
+
+def get_all_blocked_items() -> List[Tuple[str, str]]:
```
eviljeff, line 48: same comment about type definition

src/olympia/blocklist/mlbf.py
```diff
@@ -136,50 +169,46 @@ def __init__(self, *args, **kwargs):
...
-                (version.block__guid, version.version__version)
-                for version in self._all_blocks
-                if version.block_type == block_type
-            ]
-        )
+        # We define get_all_blocked_items as a separate function to allow
+        # mocking the database query in tests to simulate large data sets.
+        for guid, version, id, block_type in get_all_blocked_items():
+            _blocked_version_ids.append(id)
+            _blocked[block_type].append((guid, version))
```
eviljeff, line 182:
```suggestion
        for guid, version_string, version_id, block_type in get_all_blocked_items():
            _blocked_version_ids.append(version_id)
            _blocked[block_type].append((guid, version_string))
```

src/olympia/blocklist/mlbf.py
```diff
@@ -136,50 +169,46 @@ def __init__(self, *args, **kwargs):
...
-        return self._format_blocks(BlockType.SOFT_BLOCKED)
+    def soft_blocked_items(self):
+        blocks_dict, _ = self._all_blocks
+        return MLBF.hash_filter_inputs(blocks_dict[BlockType.SOFT_BLOCKED])
 
     @cached_property
-    def not_blocked_items(self) -> List[str]:
-        all_blocks_ids = [version.version_id for version in self._all_blocks]
+    def not_blocked_items(self):
+        _, all_blocks_ids = self._all_blocks
```
eviljeff, line 198: Can we have a more accurate name for `all_blocks_ids` - they aren't Block ids, they're Version ids

src/olympia/blocklist/mlbf.py
eviljeff, line 33: re: `all_blocks_ids` - they aren't block ids, they're version ids.

src/olympia/blocklist/mlbf.py
eviljeff: Though I do wonder if it's more efficient to generate `_blocked_version_ids` and `_blocked` with list/set generator syntax, rather than loop through with a for loop and append one at a time.

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -379,6 +380,46 @@ def test_invalid_cache_results_in_diff(self):
...
+            for x in range(0, 2_000_000)
+        ]
+        one_million_not_blocked = [
+            (f'{x}@not_blocked', f'{x}@version') for x in range(2_000_000, 3_000_000)
+        ]
+
+        mock_get_all_blocked_items.return_value = two_million_blocked
+        mock_get_not_blocked_items.return_value = one_million_not_blocked
+
+        upload_mlbf_to_remote_settings()
```
eviljeff, line 413: is there a pytest-y way to measure the time spent on this call, so we can check it doesn't regress? (Obviously if it _really_ regressed we'd notice CI timing out, or something)

eviljeff (commented):
wdyt?

src/olympia/blocklist/tests/test_cron.py
KevinMind: There is a [pytest-timeout](https://pypi.org/project/pytest-timeout/) plugin we could add yes. Since this is a separate PR now, maybe it does make sense to add it.?

eviljeff: up to you really - it was just an idea

src/olympia/blocklist/mlbf.py
eviljeff: If we're just considering this function chatGPT may have a point, but I meant use sets throughout the code in this file; then we can use more efficient set arithmetic and drop this function.  (And do `sorted(...)` only when we're writing json)

src/olympia/blocklist/mlbf.py
eviljeff: I was thinking more than, in python anyway, it's generally more efficient to build a list with a generator than to build it by iterating and repeatedly appending items (because it needs resizing and reallocating more).  I don't have stats to back-up the difference between generating _two_ lists vs. iterating over the list once though.

src/olympia/blocklist/mlbf.py
```diff
@@ -233,16 +262,17 @@ def generate_and_write_filter(self):
 
         log.info(json.dumps(stats))
 
+    @lru_cache(maxsize=128)  # noqa: B019
```
eviljeff, line 265: nit: I don't think we're ever going to call the function 128 times 🙃 - it's typically like, 3 times, isn't it?

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -379,6 +380,48 @@ def test_invalid_cache_results_in_diff(self):
             in self.mocks['olympia.blocklist.cron.upload_filter.delay'].call_args_list
         )
 
+    # This test is very slow, so we increase the timeout to 180 seconds.
+    @pytest.mark.timeout(120)
```
eviljeff, line 384: 120 or 180?

eviljeff (approved):
... yeah?  I remain in the pro-set camp though. And I'd like someone who's been more involved with the intricacies of the current soft-blocking work (e..g @willdurand) to give this a once-over too.

(I understand this may not be merged for a while anyway, if we're soft-freezing blocklisting changes for QA)

src/olympia/blocklist/mlbf.py
KevinMind: FWIW I wasn't considering just this function and it doesn't make a difference in any case. Using set arithmetic is going to result in more space usage and the same amount of time usage. It simply does not benefit us in this use outside of perhaps looking nicer.

Additionally, one of the issues we had with the approach of sorting before writing is the built in sort method didn't seem to work predictably causing really flaky tests.. I think we'd have to figure that out if we wanted to switch to sets.

src/olympia/blocklist/mlbf.py
KevinMind: Yeah IDK about the perf difference there, but I imagine it would not be significant since we are comparing a slightly slower single iteration versus a slightly faster double iteration. We could give it a try and see if there is a big diff.

src/olympia/blocklist/mlbf.py
KevinMind: I think it can be more than three but yeah probably not 128. I think I copy pasted that from docs.. maybe there is a sensible default.

diox:
Just to make it clear: because QA is testing things right now and this is the last push of the year, we'd like to avoid big changes to blocklisting, so this should *NOT* land in this sprint, even if approved.

PR #22862, opened 2024-11-18
Remove time inefficient loop in not_blocked_items

KevinMind:
Relates to: mozilla/addons#15170

### Description

Removes time inefficient loop in get_not_blocked_items by filtering from a once created set instead of an inlined list

### Context

This patch adds a slow test 20 seconds that can be isolated from the rest of our tests in a follow up PR

### Testing

CI passes, this PR will be verified on staging as well.

### Checklist

### DIFF
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -175,14 +175,11 @@ class MLBFDataBaseLoader(BaseMLBFLoader):
             .order_by('id')
             .values_list('addon__addonguid__guid', 'version')
         )
+        blocked_items = set(self.blocked_items + self.soft_blocked_items)
         # even though we exclude all the version ids in the query there's an
         # edge case where the version string occurs twice for an addon so we
         # ensure not_blocked_items contain no blocked_items or soft_blocked_items.
-        return [
-            item
-            for item in not_blocked_items
-            if item not in self.blocked_items + self.soft_blocked_items
-        ]
+        return [item for item in not_blocked_items if item not in blocked_items]
 
 
 class MLBF:


src/olympia/blocklist/mlbf.py
```diff
@@ -136,25 +171,14 @@ def __init__(self, *args, **kwargs):
...
+        return get_all_blocked_items()
 
     def _format_blocks(self, block_type: BlockType) -> List[str]:
         return MLBF.hash_filter_inputs(
             [
-                (version.block__guid, version.version__version)
+                (version[0], version[1])
                 for version in self._all_blocks
-                if version.block_type == block_type
+                if version[3] == block_type
```
willdurand, line 181: could you switch back to _named_ values? or what's the problem otherwise?

src/olympia/blocklist/mlbf.py
```diff
@@ -168,21 +192,18 @@ def soft_blocked_items(self) -> List[str]:
...
+        blocked_items = set(self.blocked_items + self.soft_blocked_items)
         # even though we exclude all the version ids in the query there's an
         # edge case where the version string occurs twice for an addon so we
         # ensure not_blocked_items contain no blocked_items or soft_blocked_items.
-        return [
-            item
-            for item in not_blocked_items
-            if item not in self.blocked_items + self.soft_blocked_items
-        ]
+        newly_not_blocked_items, _, _ = ordered_diff_lists(
```
willdurand, line 203: I get why you're reusing this but is it necessary? vs. `return [item for item not in not_blocked_items if item not is blocked_items]`? isn't `ordered_diff_lists` computing two lists?

src/olympia/blocklist/mlbf.py
KevinMind: I can for this patch, but in order to mock the result using named values means I need to either create millions of BlockVersion instances (expensive) or create a custom class with these properties (expensive and ugly)

src/olympia/blocklist/mlbf.py
KevinMind: I don't see the issue reusing it. it's doing the exact same thing and better ensures that we use the most performant loop. Had we already been using that method, we wouldn't have needed this patch at all.

willdurand: isn't `ordered_diff_lists` computing two lists?

KevinMind: Yes, and that's what we want. we have `blocked_items` and `not_blocked_items` and we want to diff them. ordered_diff_lists returns 3 values `added`, `removed`, `count`. We want added, so we return that value.

willdurand: right, and what I am saying is that we could do without the computation of `removed`, no?

src/olympia/blocklist/mlbf.py
willdurand: I suppose a `@dataclass` might be a decent middle ground since I'd rather keep the production code easier to read than less obvious because of the tests. Let's continue the conversation in the follow-up patch.

src/olympia/blocklist/mlbf.py
willdurand: we can move this conversation elsewhere if the method call moves alongside. I don't disagree that it's a can vs. should situation. Since we're trying to fix prod, maybe we should change the minimum amount of code?

KevinMind: @willdurand this code passed the test, we are going to verify it on stage. It's arguably more code change to not reuse the existing tested, verified method. I don't want to change this any more than I already have.

willdurand: I am not clear on why you insist on doing something here that is more code changes while - at the same time - changing the code in https://github.com/mozilla/addons-server/pull/22861 to do exactly what I've been suggesting so far. It sounds to me that this is the simplest diff we can come up with:

```diff
diff --git a/src/olympia/blocklist/mlbf.py b/src/olympia/blocklist/mlbf.py
index d2cc578afb..fa3a742f7b 100644
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -175,14 +175,11 @@ class MLBFDataBaseLoader(BaseMLBFLoader):
             .order_by('id')
             .values_list('addon__addonguid__guid', 'version')
         )
+        blocked_items = set(self.blocked_items + self.soft_blocked_items)
         # even though we exclude all the version ids in the query there's an
         # edge case where the version string occurs twice for an addon so we
         # ensure not_blocked_items contain no blocked_items or soft_blocked_items.
-        return [
-            item
-            for item in not_blocked_items
-            if item not in self.blocked_items + self.soft_blocked_items
-        ]
+        return [item for item in not_blocked_items if item not in blocked_items]
```

I'll add that statements like "I don't want to change this any more than I already have" don't belong to code review discussions. This is a conversation where we both need to be convinced that we got to the best solution (for some definition of "best").

willdurand (changes requested):
Could you squash your commits to get a single commit in this PR? I know GitHub will do it eventually but it'd be easier to navigate the changes locally.

KevinMind:
Ok @willdurand let's do it your way. I have things to say about your feedback but PR comments is not the place, especially when the goal here is to "unbreak prod on the short term"

src/olympia/blocklist/mlbf.py
KevinMind: Done.

KevinMind: Fwiw I did not update the second PR to what you are suggesting but to a third option which addresses your concern and maintains my (legitimate) goal of re-using the single function for diffing.

willdurand (approved):
r+, thanks

PR #22868, opened 2024-11-19
Store NeedsHumanReview for promoted & auto-approval disabled when auto-approval is attempted

diox:
Fixes https://github.com/mozilla/addons/issues/15072

### Description

That removes the need for the due date code to check for promoted & auto-approval disabled flags, instead relying just on the NHR that is now added when we notice we can't auto-approve the version.

### Context

Previously, we duplicated the checks for whether an add-on had one of the auto-approval disabled flags set or was promoted in the due date handling code (called when a new version is submitted and a bunch of other places) and auto-approve code. 

With this change, that duplication is removed, and `auto_approve` becomes responsible for adding the needs human review flag, which in turn causes a due date to be added, which in turn causes the version to appear in the human review queue.

### Testing

This should be transparent to developers/reviewers, with 2 important differences if a version belongs to an add-on that belongs to a group for which versions are pre-reviewed by a human[^1] or has one of the auto-approval disabled flags [^2]:
- They won't get a due date immediately like they did before: instead if should happen once `auto_approve` runs on the version and finds the version can't be auto-approved
- They should get an active `NeedsHumanReview` flag once `auto_approve` has run on the version, with the corresponding `reason` (cumulative).

You can set those flags/states on an add-on with the admin and check what happens when a new version is submitted. This PR contains unit tests doing that too.

[^1]: listed recommended, listed by Firefox, listed spotlight, both channels for notable
[^2]: `auto_approval_disabled`, `auto_approval_disabled_until_next_approval` or `auto_approval_delayed_until` in `AddonReviewerFlag` (and their corresponding variants for unlisted channel, add `_unlisted` suffix).

### Checklist


src/olympia/addons/models.py
```diff
@@ -2013,13 +2013,6 @@ class AddonReviewerFlags(ModelBase):
     )
 
 
-@receiver(
-    dbsignals.post_save, sender=AddonReviewerFlags, dispatch_uid='addon_review_flags'
-)
-def update_due_date_for_auto_approval_changes(sender, instance=None, **kwargs):
-    instance.addon.update_all_due_dates()
-
-
```
diox, line 2022: No longer needed since the various `AddonReviewerFlags` no longer directly cause something to have a due date.

src/olympia/promoted/models.py
```diff
@@ -179,13 +180,6 @@ def update_es_for_promoted(sender, instance, **kw):
     update_search_index(sender=sender, instance=instance.addon, **kw)
 
 
-@receiver(
-    models.signals.post_save, sender=PromotedAddon, dispatch_uid='promoted_prereview'
-)
-def update_due_date_for_pre_review(sender, instance, **kw):
```
diox, line 185: No longer needed since belonging to a `PromotedAddon` no longer directly cause something to have a due date (it has to go through `auto_approve` first).

src/olympia/promoted/tests/test_models.py
```diff
@@ -287,25 +241,3 @@ def test_approve_for_addon(self):
         promo.addon.reload()
         assert promo.addon.promoted_group() == promoted.SPOTLIGHT
         assert promo.addon.current_version.version == '0.123a'
-
-    def test_signal(self):
```
diox, line 291: No longer needed since belonging to a `PromotedAddon` no longer directly cause something to have a due date (it has to go through `auto_approve` first).

src/olympia/reviewers/tests/test_views.py
```diff
@@ -6852,40 +6875,6 @@ def test_clear_pending_rejections(self):
             version__addon=self.addon, pending_content_rejection__isnull=False
         ).exists()
 
-    def test_clear_pending_rejections_triggers_reset_due_date(self):
```
diox, line 6855: Updating due dates directly when clearing pending rejection date no longer makes sense since it doesn't affect the individual NHR instances anyway.

src/olympia/files/tests/test_models.py
```diff
@@ -371,14 +371,16 @@ def test_get_review_status_display(self):
...
+        addon = addon_factory(file_kw={'status': amo.STATUS_DISABLED})
+        version = addon.versions.get()
+        version.needshumanreview_set.create()
         assert not version.due_date
 
         version.file.update(status=amo.STATUS_AWAITING_REVIEW)
         assert version.reload().due_date
 
-        version.file.update(status=amo.STATUS_APPROVED)
+        version.file.update(status=amo.STATUS_DISABLED)
```
eviljeff, line 383: Might be something obvious, but I'm missing why this test can to switch around the statuses leading to a due_date

eviljeff (approved):

src/olympia/files/tests/test_models.py
diox: All this test does is prove we trigger a due date reset on file status change. Previously, because the add-on had `auto_approval_disabled`, the file started approved, so it didn't have a due date because auto-approval being disabled only had an impact on versions awaiting review. It was then changed to awaiting review, got a due date, and changed back to approved, so it lost it.

I've changed it to prove the reset was still happening on status change, but because now the only thing that matters to get a due date is having a NHR and not being disabled (or being signed), the file starts disabled, then is changed to awaiting review, then back to disabled.

src/olympia/amo/tests/__init__.py
```diff
@@ -665,9 +665,17 @@ def _get_created(created):
...
 
 
-def addon_factory(status=amo.STATUS_APPROVED, version_kw=None, file_kw=None, **kw):
+def addon_factory(
+    status=amo.STATUS_APPROVED,
+    version_kw=None,
+    file_kw=None,
+    needshumanreview_kw=None,
+    **kw,
+):
```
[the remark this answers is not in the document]
diox: https://github.com/mozilla/addons-server/pull/22868/commits/b39fa494c4ee2c47720f0f63967ae0416c73abec

PR #13659, opened 2020-03-06
Submit bloom filter as kinto attachment with cron job

eviljeff:
fixes #13616 - this is using the (previously assumed to be enough) salt in the keys and the existing hash function the library provides.  I'll make the change to whatever the final encoding/hashing solution is in a follow-up.  Also, @Rob--W has requested a reliable date be included in the record metadata to compare signed dates too, which will also be added in a follow-up.

PR #22870, opened 2024-11-20
Use correct name for softblack waffle switch

KevinMind:
Fixes: mozilla/addons#15166

### Description

Use correct name for waffle switch

### Checklist


willdurand (changes requested):
It's my understanding that the `blocklist_mlbf_submit` waffle switch predates soft-blocking (I can see it in https://github.com/mozilla/addons-server/pull/13659) and its name makes me think that it's used to control the upload to Remote Settings.

Should we tie this to soft-blocking?

KevinMind:
@willdurand ha good catch. That's what happens when you open a PR while keeping a toddler from mayhem 😂 

There should be two separate switches, one for gating the entire cron and one for controlling the inclusion of soft blocking. I'll have to look again but it could be that the soft block flag is only introduced in the original PR and so it would make no sense to split the PR. Have to verify that.

PR #22874, opened 2024-11-21
Ensure TESTING_ENV is defined in all environments correctly

KevinMind:
Fixes: mozilla/addons#15185

### Description

Quick fix for the bug by defining the value to False in settings_base we ensure it can be referenced in any environment

### Context

We could consider wrapping this access in a util function that can be tested and guaranteed to never raise.. I think it might be overkill though.

### Testing

You have to run make up on a fresh environment:

```bash
make down
make docker_mysqld_volume_remove
make up
```

The migration should not raise.

### Checklist

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1606,3 +1606,7 @@ CINDER_QUEUE_PREFIX = 'amo-dev-'
 SOCKET_LABS_HOST = env('SOCKET_LABS_HOST', default='https://api.socketlabs.com/v2/')
 SOCKET_LABS_TOKEN = env('SOCKET_LABS_TOKEN', default=None)
 SOCKET_LABS_SERVER_ID = env('SOCKET_LABS_SERVER_ID', default=None)
+
+# Set to True in settings_test.py
+# This controls the behavior of migrations
+TESTING_ENV = False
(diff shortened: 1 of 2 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -1606,3 +1606,7 @@ def read_only_mode(env):
 SOCKET_LABS_HOST = env('SOCKET_LABS_HOST', default='https://api.socketlabs.com/v2/')
 SOCKET_LABS_TOKEN = env('SOCKET_LABS_TOKEN', default=None)
 SOCKET_LABS_SERVER_ID = env('SOCKET_LABS_SERVER_ID', default=None)
+
+# Set to True in settings_test.py
+# This controls the behavior of migrations
+TESTING_ENV = False
```
willdurand, line 1612: Do we still need https://github.com/mozilla/addons-server/blob/f6569245aa5d6e4b89ead29d01fc9c74bb1635a7/settings.py#L198-L199 if it's defined there?

willdurand:
> Relates to: mozilla/addons#22828

Fixes?

src/olympia/lib/settings_base.py
KevinMind: good catch.

KevinMind:
> > Relates to: mozilla/addons#22828
> 
> Fixes?

Don't wanna say fixes. See the ticket. I think there are better solutions but this just unbreaks dev.

willdurand (approved):
That looks reasonable, thanks. I haven't tried the patch locally but I trust you.

PR #22875, opened 2024-11-21
Change growth threshold algorithm do use percentage points increase

diox:
Also contains a refactor to make the code easier to test, as well as admin & reviewer tools tweaks to make the consequences of changing that threshold more obvious.

If the average hotness in a tier is 0.1, and the growth threshold is set to 50, we used to flag add-ons in that tier with hotness over 0.15 (percentage increase) now we'll flag over 0.51 (percentage points increase).

Fixes mozilla/addons#15178

### Screenshots

New admin features:
Screenshot 2024-11-21 at 14-45-47 low Change usage tier Django site admin

### Testing

You can add addons with various `hotness` and `average_daily_users` values to your local environment to test and trigger `flag_high_hotness_according_to_review_tier` task (if the add-on is flagged, it should get a `NeedsHumanReview` instance with `HOTNESS_THRESHOLD` reason attached to the latest signed version in each channel that hasn't already been explicitly reviewed by a human), or trust the unit tests that do that.

### DIFF
--- a/src/olympia/addons/tasks.py
+++ b/src/olympia/addons/tasks.py
@@ -5,7 +5,7 @@ from datetime import date
 
 from django.core.files.storage import default_storage as storage
 from django.db import transaction
-from django.db.models import Avg, Q, Value
+from django.db.models import Q, Value
 from django.db.models.functions import Collate
 
 from elasticsearch_dsl import Search
@@ -536,28 +536,13 @@ def flag_high_hotness_according_to_review_tier():
         # Need a hotness threshold to be set for the tier.
         growth_threshold_before_flagging__isnull=False,
     )
-    base_qs = Addon.unfiltered.exclude(status=amo.STATUS_DISABLED).filter(
-        type=amo.ADDON_EXTENSION
-    )
+    # Build a single queryset with all add-ons we want to flag for each tier.
+    base_qs = UsageTier.get_base_addons()
     hotness_per_tier_filters = Q()
     for usage_tier in usage_tiers:
-        tier_filters = {
-            'average_daily_users__gte': usage_tier.lower_adu_threshold,
-            'average_daily_users__lt': usage_tier.upper_adu_threshold,
-        }
-        mean_hotness_in_tier = (
-            base_qs.filter(**tier_filters)
-            .aggregate(Avg('hotness', default=0))
-            .get('hotness__avg')
-        )
-        if not mean_hotness_in_tier:
-            continue
-        hotness_ceiling_before_flagging = mean_hotness_in_tier * (
-            1 + usage_tier.growth_threshold_before_flagging / 100
-        )
-        hotness_per_tier_filters |= Q(
-            hotness__gt=hotness_ceiling_before_flagging, **tier_filters
-        )
+        # Note: get_growth_threshold_q_object can return an empty Q() object
+        # if the average hotness for that tier is exactly 0.
+        hotness_per_tier_filters |= usage_tier.get_growth_threshold_q_object()
     if not hotness_per_tier_filters:
         return
     qs = base_qs.filter(hotness_per_tier_filters)
--- a/src/olympia/addons/tests/test_tasks.py
+++ b/src/olympia/addons/tests/test_tasks.py
@@ -256,7 +256,7 @@ def test_flag_high_hotness_according_to_review_tier():
         name='D tier (below minimum usage for hotness)',
         lower_adu_threshold=0,
         upper_adu_threshold=100,
-        growth_threshold_before_flagging=0.1,
+        growth_threshold_before_flagging=1,
     )
     UsageTier.objects.create(
         name='C tier (no growth threshold)',
@@ -267,13 +267,13 @@ def test_flag_high_hotness_according_to_review_tier():
         name='B tier',
         lower_adu_threshold=200,
         upper_adu_threshold=250,
-        growth_threshold_before_flagging=35,
+        growth_threshold_before_flagging=11,
     )
     UsageTier.objects.create(
         name='A tier',
         lower_adu_threshold=250,
         upper_adu_threshold=1000,
-        growth_threshold_before_flagging=20,
+        growth_threshold_before_flagging=7,
     )
     UsageTier.objects.create(
         name='S tier (no upper threshold)',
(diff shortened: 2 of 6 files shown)


src/olympia/addons/tasks.py
```diff
@@ -536,28 +536,13 @@ def flag_high_hotness_according_to_review_tier():
...
-        mean_hotness_in_tier = (
-            base_qs.filter(**tier_filters)
-            .aggregate(Avg('hotness', default=0))
-            .get('hotness__avg')
-        )
-        if not mean_hotness_in_tier:
-            continue
-        hotness_ceiling_before_flagging = mean_hotness_in_tier * (
-            1 + usage_tier.growth_threshold_before_flagging / 100
-        )
```
diox, line 557: Old algorithm was defined here

KevinMind: I guess I'm wondering about the algorithm change? If the issue was about penalizing addons in tiers with negative average growth, could we invert the behaviour depending on if the average growth is above or below 0?


After talking it through with GPT something like this?

```python
def calculate_hotness(mean_hotness_in_tier, growth_threshold):
    if mean_hotness_in_tier > 0:
        return mean_hotness_in_tier * (1 + growth_threshold)
    elif mean_hotness_in_tier < 0:
        return mean_hotness_in_tier * (1 / (1 + growth_threshold))
    else:
        return 0
```

This would mean as the negative growth approaches zero we reduce the threshold to zero and as positive growth increases we increase it to infinity, I think.. 🙃 🤓

src/olympia/reviewers/models.py
```diff
@@ -798,6 +801,52 @@ class Meta:
...
+        """Return boundaries used to filter add-ons for that tier instance."""
+        return {
+            'average_daily_users__gte': self.lower_adu_threshold,
+            'average_daily_users__lt': self.upper_adu_threshold,
+        }
+
+    @cached_property
+    def average_growth(self):
+        return (
+            self.get_base_addons()
```
KevinMind, line 824: Does this result in the same number of queries in the end? It's a bit hard to tell given the refactor and changes to the logic.

src/olympia/addons/tasks.py
```diff
@@ -536,28 +536,13 @@ def flag_high_hotness_according_to_review_tier():
         # Need a hotness threshold to be set for the tier.
         growth_threshold_before_flagging__isnull=False,
     )
-    base_qs = Addon.unfiltered.exclude(status=amo.STATUS_DISABLED).filter(
-        type=amo.ADDON_EXTENSION
-    )
+    # Build a single queryset with all add-ons we want to flag for each tier.
```
KevinMind, line 539: It would be a lot easier to review this if we had 2 commits. 1) migrate the code to the new structure moving the logic to the model and then 2) modify the logic calculating the threshold filters.

Is that too big an ask? I feel I'm spending a lot of time trying to map against the two types of changes.

KevinMind (commented):
Looks good in principle.. I have a slight doubt/question about the algo change.

src/olympia/addons/tests/test_tasks.py
```diff
@@ -256,7 +256,7 @@ def test_flag_high_hotness_according_to_review_tier():
         name='D tier (below minimum usage for hotness)',
         lower_adu_threshold=0,
         upper_adu_threshold=100,
```
KevinMind, line 258: I would think we want to cover cases where growth is negative? just so it is included in the set of possible scenarios?

src/olympia/addons/tasks.py
diox: I originally had a variant of this using `abs()` but Operations / @abhn explicitly requested to use percentage points instead (https://mozilla-hub.atlassian.net/browse/ADDONSOPS-381).

src/olympia/reviewers/models.py
diox: Yes. That's why I added a `get_growth_threshold_q_object()` method and made `get_base_addons()` separate. We're looping over each `UsageTier`, computing the average for each (so one query per tier - we don't have a lot of them fortunately) and building the final queryset, then once we're done looping we execute the query to find all add-ons over threshold in one single query regardless of which tier they belong to.

src/olympia/addons/tasks.py
diox: I actually did that but rebased it away :sweat_smile: Ultimately the algorithm change is extremely straightforward, since it's just a single line in `get_growth_threshold()`.

src/olympia/addons/tests/test_tasks.py
diox: https://github.com/mozilla/addons-server/pull/22875/commits/b0a647158306cc65637d80ad5e73548bd91e428e

src/olympia/reviewers/models.py
diox: 3e0a058768e81ea1ad6f33b691ccbb6b0d8817e4 adds a test around the number of queries for the task itself

src/olympia/reviewers/models.py
```diff
@@ -798,6 +801,52 @@ class Meta:
...
+            .get('hotness__avg')
+        )
+
+    def get_growth_threshold(self):
+        """Return the growth threshold for that tier, or None.
+
+        The value is computed from the average growth (hotness) of add-ons in
+        that tier plus the growth_threshold_before_flagging converted to
+        decimal."""
+        return self.average_growth + self.growth_threshold_before_flagging / 100
```
diox, line 836: With the old algorithm this would have been:
```python
        return self.average_growth * (
            1 + self.growth_threshold_before_flagging / 100
        )

src/olympia/addons/tasks.py
diox: See my other comment: https://github.com/mozilla/addons-server/pull/22875/files#r1853839986

src/olympia/addons/tasks.py
KevinMind: Not trying to block, but does the function I shared above not handle zero correctly? Maybe you could describe what was incorrect about zero growth in a tier in the PR context?

diox: Your suggestion would have `0` as the threshold if the average hotness in the tier was `0`, which is undesirable. As I said, this was discussed with Operations and they decided, after having looked at the data, to go for a percentage point increase (an addition) instead of a percentage increase (a multiplication).

KevinMind:
@diox is it possble to get more specific criteria for a given addon, which values should be set to what in order to trigger or not trigger the flag?

PR #22880, opened 2024-11-21
Switch to Python 3.12

diox:
Fixes: mozilla/addons#15165

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # Read the docs/topics/development/docker.md file for more information about this Dockerfile.
 ####################################################################################################
 
-FROM python:3.11-slim-bookworm AS olympia
+FROM python:3.12-slim-bookworm AS olympia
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,7 +2,7 @@
 # Read the docs/topics/development/docker.md file for more information about this Dockerfile.
 ####################################################################################################
 
-FROM python:3.11-slim-bookworm AS olympia
+FROM python:3.12-slim-bookworm AS olympia
```
eviljeff, line 5: 😄 literally no changes?

eviljeff (approved):

Dockerfile
diox: Yeah. Looking at dependency updates the past couple years 99% of them added 3.12 to their compatibility matrix without any code changes.

Biggest breaking change is the removal of `distutils` by default but `setuptools` provides it for backwards-compatibility.

PR #22885, opened 2024-11-22
Add support for soft/hard block in upload_to_mllf_to_remote_settings cron

KevinMind:
Relates to: https://github.com/mozilla/addons/issues/15014

### Description

- adds support to process fort blocks in the cron
- does not actually support soft blocks, we skip them always
- adds test proving we always skip them

### Context

This commit makes actually adding them more clear and easier to grok

### Testing

Test exactly the same as <https://github.com/mozilla/addons-server/pull/22828>

Adding hard or soft blocks should produce only hard block filter/stash regardless of the soft block change count or the `enable-soft-blocking` waffle_switch.

### Checklist

### DIFF
--- a/src/olympia/blocklist/cron.py
+++ b/src/olympia/blocklist/cron.py
@@ -116,13 +116,6 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
         len(mlbf.data.not_blocked_items),
     )
 
-    # Until we are ready to enable soft blocking, it should not be possible
-    # to create a stash and a filter at the same iteration
-    # This should never happen, but we'll raise an exception just in case.
-    assert not (create_stash and len(base_filters_to_update) > 0), (
-        'Cannot upload stash and filter without implementing soft blocking'
-    )
-
     if create_stash:
         mlbf.generate_and_write_stash(previous_filter)
 
--- a/src/olympia/blocklist/cron.py
+++ b/src/olympia/blocklist/cron.py
@@ -1,11 +1,11 @@
 from datetime import datetime
+from typing import List
 
 import waffle
 from django_statsd.clients import statsd
 
 import olympia.core.logger
 from olympia.constants.blocklist import (
-    BASE_REPLACE_THRESHOLD,
     MLBF_BASE_ID_CONFIG_KEY,
     MLBF_TIME_CONFIG_KEY,
 )
@@ -13,7 +13,7 @@ from olympia.zadmin.models import get_config
 
 from .mlbf import MLBF
 from .models import Block, BlocklistSubmission, BlockType
-from .tasks import cleanup_old_files, process_blocklistsubmission, upload_filter
+from .tasks import process_blocklistsubmission, upload_filter
 from .utils import datetime_to_ts
 
 
@@ -28,9 +28,9 @@ def get_last_generation_time():
     return get_config(MLBF_TIME_CONFIG_KEY, None, json_value=True)
 
 
-def get_base_generation_time():
+def get_base_generation_time(block_type: BlockType):
     return get_config(
-        MLBF_BASE_ID_CONFIG_KEY(BlockType.BLOCKED, compat=True), None, json_value=True
+        MLBF_BASE_ID_CONFIG_KEY(block_type, compat=True), None, json_value=True
     )
 
 
@@ -66,48 +66,45 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
     # An add-on version/file from after this time can't be reliably asserted -
     # there may be false positives or false negatives.
     # https://github.com/mozilla/addons-server/issues/13695
-    generation_time = get_generation_time()
-    # This timestamp represents the last time the MLBF was generated and uploaded.
-    # It could have been a base filter or a stash.
-    last_generation_time = get_last_generation_time()
-    # This timestamp represents the point in time when
-    # the base filter was generated and uploaded.
-    base_generation_time = get_base_generation_time()
-
-    mlbf = MLBF.generate_from_db(generation_time)
-
-    base_filter = (
-        MLBF.load_from_storage(base_generation_time)
-        if base_generation_time is not None
-        else None
+    mlbf = MLBF.generate_from_db(get_generation_time())
+    previous_filter = MLBF.load_from_storage(
+        # This timestamp represents the last time the MLBF was generated and uploaded.
+        # It could have been a base filter or a stash.
+        get_last_generation_time()
     )
 
-    previous_filter = (
-        # Only load previoous filter if there is a timestamp to use
-        # and that timestamp is not the same as the base_filter
-        MLBF.load_from_storage(last_generation_time)
-        if last_generation_time is not None
-        and (base_filter is None or base_filter.created_at != last_generation_time)
-        else base_filter
-    )
+    base_filters_to_update: List[BlockType] = []
+    create_stash = False
 
-    changes_count = mlbf.blocks_changed_since_previous(
-        BlockType.BLOCKED, previous_filter
-    )
-    statsd.incr(
-        'blocklist.cron.upload_mlbf_to_remote_settings.blocked_changed', changes_count
-    )
-    need_update = (
-        force_base
-        or base_filter is None
-        or (
-            previous_filter is not None
-            and previous_filter.created_at < get_blocklist_last_modified_time()
-        )
-        or changes_count > 0
-    )
-    if not need_update:
+    # Determine which base filters need to be re uploaded
+    # and whether a new stash needs to be created
+    for block_type in BlockType:
+        # This prevents us from updating a stash or filter based on new soft blocks
+        if block_type == BlockType.SOFT_BLOCKED:
+            log.info(
+                'Skipping soft-blocks because enable-soft-blocking switch is inactive'
+            )
+            continue
+
+        base_filter = MLBF.load_from_storage(get_base_generation_time(block_type))
+
+        # add this block type to the list of filters to be re-uploaded
+        if (
+            force_base
+            or base_filter is None
+            or mlbf.should_upload_filter(block_type, base_filter)
+        ):
+            base_filters_to_update.append(block_type)
+        # only update the stash if we should AND if
+        # we aren't already reuploading the filter for this block type
+        elif mlbf.should_upload_stash(block_type, previous_filter or base_filter):
+            create_stash = True
+
+    skip_update = len(base_filters_to_update) == 0 and not create_stash
+    if skip_update:
         log.info('No new/modified/deleted Blocks in database; skipping MLBF generation')
+        # Delete the locally generated MLBF directory and files as they are not needed
+        mlbf.delete()
         return
 
     statsd.incr(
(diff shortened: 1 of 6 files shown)


src/olympia/blocklist/mlbf.py
```diff
@@ -207,13 +209,21 @@ def hash_filter_inputs(cls, input_list):
             for (guid, version) in input_list
         ]
 
-    def filter_path(self, _block_type: BlockType = BlockType.BLOCKED):
-        return self.storage.path('filter')
+    def filter_path(self, block_type: BlockType):
+        # TODO: explain / test
+        if block_type == BlockType.BLOCKED:
+            return self.storage.path('filter')
```
willdurand, line 215: Why do we need to special case this path? Backward compatibility?

KevinMind: Correct. Existing filters in the file system are going to have that name.

src/olympia/blocklist/tests/test_tasks.py
```diff
@@ -284,6 +284,7 @@ def test_skip_cleanup_when_no_filters_or_config_keys(self):
     def test_cleanup_records_with_matching_attachment(self):
         mlbf = MLBF.generate_from_db(self.generation_time)
         mlbf.generate_and_write_filter(BlockType.BLOCKED)
+        mlbf.generate_and_write_filter(BlockType.SOFT_BLOCKED)
```
willdurand, line 287: Why is this not changing anything?

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -398,9 +696,10 @@ def test_get_last_generation_time(self):
         assert get_last_generation_time() == 1
 
     def test_get_base_generation_time(self):
-        assert get_base_generation_time() is None
-        set_config(MLBF_BASE_ID_CONFIG_KEY(BlockType.BLOCKED, compat=True), 1)
-        assert get_base_generation_time() == 1
+        for block_type in BlockType:
+            assert get_base_generation_time(block_type) is None
+            set_config(MLBF_BASE_ID_CONFIG_KEY(block_type, compat=True), 1)
+            assert get_base_generation_time(block_type) == 1
```
willdurand, line 702: I forgot to comment on this before but I tend to avoid values such as `0` or `1` than can easily be transformed into something else (e.g. a boolean). Not a big deal.

```suggestion
            set_config(MLBF_BASE_ID_CONFIG_KEY(block_type, compat=True), 123)
            assert get_base_generation_time(block_type) == 123
```

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -327,36 +601,47 @@ def test_creates_base_filter_if_base_generation_time_invalid(self):
...
+        MLBF.generate_from_db(self.base_time)
+        # Soft block version is not accounted for in the base filter
+        # but accounted for in the last filter
+        self._block_version(block_type=BlockType.SOFT_BLOCKED)
+        MLBF.generate_from_db(self.last_time)
+
+        upload_mlbf_to_remote_settings(force_base=False)
+        assert not self.mocks['olympia.blocklist.cron.upload_filter.delay'].called
+
+        # delete the last filter, now the base filter will be used to compare
```
willdurand, line 620:
```suggestion
        # Delete the last filter, now the base filter will be used to compare
```

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -279,44 +526,71 @@ def test_upload_stash_unless_enough_changes(self):
...
+            self._block_version(is_signed=True, block_type=BlockType.BLOCKED)
+            self._block_version(is_signed=True, block_type=BlockType.BLOCKED)
+            upload_mlbf_to_remote_settings()
+            self.mocks['olympia.blocklist.cron.upload_filter.delay'].assert_called_with(
+                self.current_time,
+                filter_list=[BlockType.BLOCKED.name],
+                create_stash=False,
+            )
+            mlbf = MLBF.load_from_storage(self.current_time)
+            mlbf = MLBF.load_from_storage(self.current_time)
```
willdurand, line 583: Why do we generate twice?

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -84,49 +85,104 @@ def _block_version(
...
-    def test_skip_update_unless_no_base_mlbf(self):
+            # Check that both filters were created on the second attempt
+            mlbf = MLBF.load_from_storage(self.current_time)
+            self.assertTrue(
+                mlbf.storage.exists(mlbf.filter_path(BlockType.BLOCKED)),
+            )
+            self.assertEqual(
+                mlbf.storage.exists(mlbf.filter_path(BlockType.SOFT_BLOCKED)),
+                # Until we are ready to enable soft blocking
+                # there should never be a soft block filter.
```
willdurand, line 117: So even with the switch enabled, we don't produce any sfot-block filter, right?

src/olympia/blocklist/mlbf.py
```diff
@@ -207,13 +209,21 @@ def hash_filter_inputs(cls, input_list):
             for (guid, version) in input_list
         ]
 
-    def filter_path(self, _block_type: BlockType = BlockType.BLOCKED):
-        return self.storage.path('filter')
+    def filter_path(self, block_type: BlockType):
+        # TODO: explain / test
```
willdurand, line 213: 👻

src/olympia/blocklist/cron.py
```diff
@@ -119,28 +116,25 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
...
         mlbf.generate_and_write_stash(previous_filter)
 
+    for block_type in base_filters_to_update:
+        mlbf.generate_and_write_filter(block_type)
+
     upload_filter.delay(
-        generation_time,
-        filter_list=[BlockType.BLOCKED.name] if make_base_filter else [],
-        create_stash=not make_base_filter,
+        mlbf.created_at,
```
willdurand, line 133: I might have missed something. Why do we change the date here?

src/olympia/blocklist/cron.py
```diff
@@ -66,48 +66,45 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
...
-            and previous_filter.created_at < get_blocklist_last_modified_time()
-        )
-        or changes_count > 0
-    )
-    if not need_update:
+    base_filters_to_update: List[BlockType] = []
+    create_stash = False
+
+    # Determine which base filters need to be re uploaded
+    # and whether a new stash needs to be created
```
willdurand, line 80:
```suggestion
    # and whether a new stash needs to be created.
```

src/olympia/blocklist/cron.py
```diff
@@ -66,48 +66,45 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
...
+    for block_type in BlockType:
+        # This prevents us from updating a stash or filter based on new soft blocks
```
willdurand, line 82:
```suggestion
        # This prevents us from updating a stash or filter based on new soft blocks.
```

src/olympia/blocklist/cron.py
```diff
@@ -66,48 +66,45 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
...
+        if block_type == BlockType.SOFT_BLOCKED:
+            log.info(
+                'Skipping soft-blocks because enable-soft-blocking switch is inactive'
+            )
+            continue
+
+        base_filter = MLBF.load_from_storage(get_base_generation_time(block_type))
+
+        # add this block type to the list of filters to be re-uploaded
```
willdurand, line 91:
```suggestion
        # Add this block type to the list of filters to be re-uploaded.
```

src/olympia/blocklist/cron.py
```diff
@@ -66,48 +66,45 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
...
+        if (
+            force_base
+            or base_filter is None
+            or mlbf.should_upload_filter(block_type, base_filter)
+        ):
+            base_filters_to_update.append(block_type)
+        # only update the stash if we should AND if
+        # we aren't already reuploading the filter for this block type
```
willdurand, line 99:
```suggestion
        # Only update the stash if we should AND if we aren't already 
        # re-uploading the filter for this block type.
```

src/olympia/blocklist/cron.py
```diff
@@ -66,48 +66,45 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
...
+        elif mlbf.should_upload_stash(block_type, previous_filter or base_filter):
+            create_stash = True
+
+    skip_update = len(base_filters_to_update) == 0 and not create_stash
+    if skip_update:
         log.info('No new/modified/deleted Blocks in database; skipping MLBF generation')
+        # Delete the locally generated MLBF directory and files as they are not needed
```
willdurand, line 106:
```suggestion
        # Delete the locally generated MLBF directory and files as they are not needed.
```

src/olympia/blocklist/cron.py
```diff
@@ -119,28 +116,25 @@ def _upload_mlbf_to_remote_settings(*, force_base=False):
...
-        or previous_filter is None
-        or mlbf.blocks_changed_since_previous(BlockType.BLOCKED, base_filter)
-        > BASE_REPLACE_THRESHOLD
-    )
+    # Until we are ready to enable soft blocking, it should not be possible
+    # to create a stash and a filter at the same iteration
+    if create_stash and len(base_filters_to_update) > 0:
+        raise Exception(
+            'Cannot upload stash and filter without implementing soft blocking'
+        )
```
willdurand, line 124: Do we have test coverage for this error?

willdurand (changes requested):
Some preliminary (minor) feedback, this looks good.

src/olympia/blocklist/cron.py
KevinMind: They have the same value. In the original PR, I switched to using the creation date because this is the value that is actually assigned to the MLBF class. It's also the value we're passing to the initialize but it just makes sense to me to use the value from the class.

src/olympia/blocklist/tests/test_cron.py
KevinMind: Correct. The next PR, updates the tests to verify we **do** produce a soft block filter.

src/olympia/blocklist/cron.py
KevinMind: It's actually impossible to cover because we force skip soft block, and if hard block produces a filter, it cannot produce a stash.. so not really. I'd have to do some serious hacking. We never expect this error. I'll add a comment as such.

src/olympia/blocklist/tests/test_cron.py
KevinMind: We generate zero times (akkkshully) but yeah we don't need to load from storage twice.

src/olympia/blocklist/tests/test_tasks.py
KevinMind: On current master (since the first of this set of patches) `filter_path` always returns the path to the hard filter.. never the soft.

That's why the test wasn't failing before. Now in this patch, filter-path returns the correct path.

src/olympia/blocklist/cron.py
willdurand: This might be a case where we want to use `assert` since that sounds like an invariant?

src/olympia/blocklist/tests/test_cron.py
```diff
@@ -259,15 +487,34 @@ def test_upload_stash_unless_missing_base_filter(self):
...
+                    self.current_time,
+                    # Even if updating a soft block and the switch is active
+                    # don't expect a soft block filter update until we
+                    # implement support for that
+                    filter_list=[BlockType.BLOCKED.name],
+                    create_stash=False,
+                )
+            ) in self.mocks['olympia.blocklist.cron.upload_filter.delay'].call_args_list
+
+    # TODO: add test for soft blocks and ensure stash/filter compatibility
```
willdurand, line 506: Is this TODO addressed in the next PR?

willdurand (approved):
r+wc, thanks

This looks good, especially when we do _not_ enable the soft-blocking switch, haha

src/olympia/blocklist/cron.py
KevinMind: Got it.

src/olympia/blocklist/tests/test_cron.py
KevinMind: Covered by https://github.com/mozilla/addons/issues/15200

PR #22886, opened 2024-11-22
Add forward to legal option in reviewer tools

eviljeff:
Fixes: mozilla/addons#15078

### Description

Adds a Request legal action in the reviewer tools.  Under the hood the activity log is attached to `DECISION_ACTIONS.AMO_LEGAL_FORWARD`, which has an associated ContentAction class with a process action that resolves the job (if there was one); then creates a new job in the legal escalations queue.

image

### Context

There's some TODOs that will be resolved with future refactoring, as it's the first reviewer-tools-launched-action where the `process_action` in the ContentActionXXX needs to do something.  (currently all the actions from the reviewer tools do the necessary action in reviewers/utils.py first)

### Testing

- Make sure you have your CINDER_API defined
- In the reviewer tools choose an add-on (any state where you could make any changes should do) and choose Forward to Legal.  Fill in some notes and submit.
- In Cinder look in the Legal Escalations queue and see a report for that add-on, with your notes as the report reasoning
- (resolve the job for a clean slate)
- Report the add-on for abuse in the add-on, for a policy violation reason, to allow the job to be created in the reviewer tools
- In the reviewer tools go back to the review page for the add-on (it should be in the review queue if you have the relevant waffle switches enabled, but that's not important)
- Choose Forward to Legal, but this time select the open job too, then write some notes and submit.
- In Cinder look in the Legal Escalations queue and see a report for that add-on, with your notes as the report reasoning
- resolve the job in cinder.
- (if you resolved the job with a policy that lead to disable; and replayed that decision webhook payload from Cinder back in your local env, it should disable the add-on and send an email to the reporter of the abuse reports, as if the legal escalation hadn't gotten in the middle.  Testing that is possibly out of scope for this change though, because all of that is existing functionality)

### Checklist


src/olympia/abuse/tests/test_cinder.py
```diff
@@ -1510,19 +1512,19 @@ def test_workflow_recreate_no_versions_to_flag(self):
             json={'job_id': '2'},
             status=201,
         )
-        assert cinder_instance.workflow_recreate(job=cinder_job) == '2'
+        assert cinder_instance.workflow_recreate(notes='', job=cinder_job) == '2'
```
KevinMind, line 1515: In code land, it seems like the only real use case for worfklow_recreate is to pass the notes from the cinder_job as the notes argument. Is it possible that this property is None or otherwise false? What happens in this case? Might be worth a test scenario if you opt not to read that arg from the job argument (see comment above)

src/olympia/abuse/tests/test_cinder.py
```diff
@@ -1567,6 +1569,56 @@ def test_workflow_recreate_waffle_switch_off(self):
         assert ActivityLog.objects.count() == 0
 
 
+class TestCinderAddonHandledByLegal(TestCinderAddon):
+    CinderClass = CinderAddonHandledByLegal
```
KevinMind, line 1573: Question: why the indirection here? Shorter syntax?

src/olympia/abuse/tests/test_cinder.py
```diff
@@ -1567,6 +1569,56 @@ def test_workflow_recreate_waffle_switch_off(self):
...
+    # For rendering the payload to Cinder like CinderAddon:
+    # - 1 Fetch Addon authors
+    # - 2 Fetch Promoted Addon
+    expected_queries_for_report = 2
+    expected_queue_suffix = None
+
+    def test_queue(self):
+        extension = self._create_dummy_target()
+        assert self.CinderClass(extension).queue == 'legal-escalations'
```
KevinMind, line 1582: Nit: Is this string defined in a constant somewhere? Would make future updates a lot easier if we referenced the constant instead of the raw string.

src/olympia/abuse/tests/test_cinder.py
```diff
@@ -1567,6 +1569,56 @@ def test_workflow_recreate_waffle_switch_off(self):
...
+        # Contrary to reports handled by Cinder moderators, for reports handled
+        # by legal the queue should remain the same regardless of the
+        # addon-type.
+        target = self._create_dummy_target(type=amo.ADDON_STATICTHEME)
+        assert self.CinderClass(target).queue_appeal == 'legal-escalations'
+
+    def test_workflow_recreate(self):
+        addon = self._create_dummy_target()
+        listed_version = addon.current_version
+        listed_version.file.update(is_signed=True)
```
KevinMind, line 1598: Question: why is it important that the listed_version is signed? Is it not possible to create signed via _create_dummy_target? maybe add a comment for clarity?

src/olympia/abuse/tests/test_cinder.py
```diff
@@ -1567,6 +1569,56 @@ def test_workflow_recreate_waffle_switch_off(self):
...
+        unlisted_version = version_factory(
+            addon=addon, channel=amo.CHANNEL_UNLISTED, file_kw={'is_signed': True}
+        )
+        ActivityLog.objects.all().delete()
```
KevinMind, line 1602: Question: why do we have to delete all activity logs? That will never happen in real life so if it's just to make the test pass, then there could be an issue in the setup?

src/olympia/abuse/tests/test_cinder.py
```diff
@@ -1567,6 +1569,56 @@ def test_workflow_recreate_waffle_switch_off(self):
...
+        assert self.CinderClass(extension).queue_appeal == 'legal-escalations'
+
+    def test_queue_with_theme(self):
+        # Contrary to reports handled by Cinder moderators, for reports handled
+        # by legal the queue should remain the same regardless of the
+        # addon-type.
+        target = self._create_dummy_target(type=amo.ADDON_STATICTHEME)
+        assert self.CinderClass(target).queue_appeal == 'legal-escalations'
+
+    def test_workflow_recreate(self):
```
KevinMind, line 1595: This test need a doc string.. It's not clear why I'm expecting the results I am, and there are odd setup scenarios that should be explained for > 6 months from now humans.

src/olympia/abuse/tests/test_models.py
```diff
@@ -2356,7 +2337,7 @@ def test_get_action_helper(self):
                     'rating': None,
                     'collection': None,
                     'user': None,
-                    **targets[ActionClass],
+                    **targets.get(ActionClass, {'addon': addon}),
```
KevinMind, line 2340: Why? This is a big behavior change as the index accessor will raise if the key is not defined, now instead we default to addon? Could you add an explainer why we default to addon: addon?

src/olympia/abuse/tests/test_models.py
```diff
@@ -3136,6 +3118,27 @@ def test_notify_reviewer_decision_rejection_addon_already_disabled(self):
...
+        )
+        self._test_notify_reviewer_decision(
+            decision,
+            amo.LOG.REQUEST_LEGAL,
+            None,
+            expect_create_decision_call=False,
+            expect_create_job_decision_call=False,
+            expect_email=False,
+        )
+        assert CinderJob.objects.get().job_id == '123456'
```
KevinMind, line 3140: TIL you can just call get() is that different from first()? Also, probably better to asset that CinderJob with job_id = `<your id>` exists or is this django magic for the same thing?

src/olympia/abuse/tests/test_models.py
```diff
@@ -3136,6 +3118,27 @@ def test_notify_reviewer_decision_rejection_addon_already_disabled(self):
             not in mail.outbox[0].body
         )
 
+    def test_notify_reviewer_decision_legal_forward(self):
```
KevinMind, line 3121: This test probably needs a doc string as well.. why should the status be disabled? why do we expect no CinderJob at first? why do we expect those specific parameters on the sub test call?

src/olympia/reviewers/tests/test_utils.py
```diff
@@ -3708,6 +3767,44 @@ def test_reject_multiple_versions_resets_original_status_too(self):
             File.STATUS_DISABLED_REASONS.NONE
         )
 
+    def test_request_legal_review_no_job(self, *, data=None):
```
KevinMind, line 3770: how would data get passed here?

src/olympia/reviewers/tests/test_utils.py
```diff
@@ -3708,6 +3767,44 @@ def test_reject_multiple_versions_resets_original_status_too(self):
...
+            status=201,
+        )
+
+        self.helper.handler.request_legal_review()
+
+        assert len(mail.outbox) == 0
+        assert report_request.call_count == 1
+        assert CinderJob.objects.last().job_id == '5678'
+        request_body = json.loads(responses.calls[0].request.body)
+        assert request_body['reasoning'] == 'foo'
```
KevinMind, line 3791: where is foo coming from?

src/olympia/reviewers/tests/test_utils.py
```diff
@@ -3708,6 +3767,44 @@ def test_reject_multiple_versions_resets_original_status_too(self):
...
+
+    def test_request_legal_review_resolve_job(self):
```
KevinMind, line 3793: This test needs a doc string.. Because "resolvable_in_reviewer_tools is True, and because we call `test_request_legal_review_no_job (passing a job, odd) we then expect the job.forwarded_to_job to equal to the last created job, with no other criteria. It doesn't really add up to me. Like, I could say, sure it makes sense, but it's very fuzzy why this makes sense.

src/olympia/reviewers/tests/test_utils.py
```diff
@@ -3708,6 +3767,44 @@ def test_reject_multiple_versions_resets_original_status_too(self):
...
+        AbuseReport.objects.create(guid=self.addon.guid, cinder_job=job)
+        responses.add_callback(
+            responses.POST,
+            f'{settings.CINDER_SERVER_URL}jobs/1234/decision',
+            callback=lambda r: (201, {}, json.dumps({'uuid': uuid.uuid4().hex})),
+        )
+        self.test_request_legal_review_no_job(data={'cinder_jobs_to_resolve': [job]})
+
+        assert job.reload().decision.action == DECISION_ACTIONS.AMO_LEGAL_FORWARD
+        assert job.forwarded_to_job == CinderJob.objects.last()
```
KevinMind, line 3806: Is there a more explicit, narrow query here that could make use of existing known data?

src/olympia/reviewers/tests/test_views.py
```diff
@@ -2416,6 +2416,7 @@ def test_need_correct_reviewer_for_promoted_addon(self):
             'reject_multiple_versions',
```
KevinMind, line 2416: I was expecting an integration test here? You've tested all the units, but a reviewer clicks a button and a get/post? request comes to reviewers and what endpoint from reviewer tools is called with what arguments to forward to legal, creating or not creating a resolvable job?

src/olympia/reviewers/utils.py
```diff
@@ -800,7 +800,7 @@ def get_actions(self):
                 'add-on or versions.'
             ),
             'minimal': True,
-            'available': is_reviewer and has_unresolved_abuse_report_jobs,
+            'available': is_appropriate_reviewer and has_unresolved_abuse_report_jobs,
```
KevinMind, line 803: Is this covered, most of the test_utils are just adding the string to a dict, so it's not very clear and based on the diff we are replacing one existing variable with another.

KevinMind (commented):
Left some comments, mostly regarding the tests. It looks pretty good so far. I think some parts need more explanation to ensure this continues making sense in the future. I'll verify the behavior and approve.

src/olympia/abuse/cinder.py
```diff
@@ -210,12 +210,13 @@ def post_report(self, *, job):
...
         pass
 
-    def workflow_recreate(self, *, job):
+    def workflow_recreate(self, *, notes, job=None):
         """Recreate a job in a queue."""
         raise NotImplementedError
 
-    def workflow_move(self, *, job):
-        """Move job to a different queue."""
+    def post_queue_move(self, *, job):
```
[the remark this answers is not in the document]
eviljeff: It's because it doesn't actually move the job - that API functionality is still under development by Cinder - and the naming implies it did. The function is for what should happen after the move on our side.

src/olympia/abuse/models.py
```diff
@@ -1256,6 +1260,12 @@ def notify_reviewer_decision(
         action_helper = self.get_action_helper(
             overridden_action=overridden_action, appealed_action=appealed_action
         )
+
+        # TODO: generalize this hack -we shouldn't need to special case specific actions
+        # This will be rewritten for https://mozilla-hub.atlassian.net/browse/AMOENG-1125
```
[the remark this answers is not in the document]
eviljeff: It's the epic

KevinMind:
Here is my addon

image


1) ✅  I can see the request legal review

image

2) ✅ I've submitted the legal review

image

eviljeff:
image

src/olympia/abuse/cinder.py
```diff
@@ -583,18 +595,19 @@ def appeal(self, *args, **kwargs):
         self.flag_for_human_review(related_versions=related_versions, appeal=True)
         return super().appeal(*args, **kwargs)
 
-    def workflow_recreate(self, *, job):
```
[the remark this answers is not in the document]
eviljeff: it's moved up to `CinderAddon` (`CinderAddonHandledByReviewers`'s super class) so `CinderAddonHandledByLegal` can use it too.

src/olympia/abuse/tests/test_cinder.py
eviljeff: It's mainly because addon_factory and friends end up generating a lot of activity logs. We can't do `assert ActivityLog.objects.count() == 0` otherwise - we could filter by the specific activity log we expect not to find, but that leaves open the edge case of another activity log being generated of a different type.

`ActivityLog.objects.all().delete()` is a pretty common pattern though - maybe we should just do it in `addon_factory`? 🤷

src/olympia/abuse/tests/test_models.py
eviljeff: Add as code comments?  It was mainly as I kept having to add new `ContentAction...` classes to the dict but they all ended up using `addon` so it was redundant.  Made sense to me, but do you think it should remain an explicit list of all the possible targets?

src/olympia/abuse/tests/test_models.py
eviljeff: `.get()` only succeeds when there is a single match so it's giving us an extra validation there aren't 0, or more than 1 CinderJobs.

src/olympia/reviewers/tests/test_utils.py
eviljeff: see line 3803

src/olympia/reviewers/tests/test_utils.py
eviljeff: It's... me ~being lazy~ exercising DRY with tests - because the functionality for escalating _without_ a job is a subset of the expected functionality for escalating with a job.  I can refactor to make it clearer...

src/olympia/reviewers/utils.py
eviljeff: mmm, you're right, it was a drive-by "fix", because I suddenly thought it should have a more restrictive condition.  I'll revert it - if it needs restricting it should be done with a separate issue & PR

src/olympia/abuse/tests/test_actions.py
```diff
@@ -830,6 +834,51 @@ def test_hold_action(self):
...
+            json={'job_id': '1234-xyz'},
+            status=201,
+        )
+
+        action.process_action()
+
+        new_cinder_job = CinderJob.objects.last()
+        assert new_cinder_job.job_id == '1234-xyz'
+        request_body = json.loads(responses.calls[0].request.body)
+        assert request_body['reasoning'] == "extra note's"
```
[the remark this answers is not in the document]
eviljeff: it's defined in the setup for BaseTestContentAction

src/olympia/reviewers/tests/test_utils.py
eviljeff: it's from the default `get_data`

src/olympia/abuse/tests/test_actions.py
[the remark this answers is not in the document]
eviljeff: it's defined in the setup for BaseTestContentAction

src/olympia/reviewers/tests/test_views.py
eviljeff: I added an integration test in the style we have in the same test class - that tests the request goes through, and then mocks the task call.

src/olympia/abuse/models.py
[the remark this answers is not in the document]
KevinMind: Hmmm, yeah I guess we don't have a solution for that in github then. Not a huge fan of jira links in the code since that is not public info but I guess there's no alternative? Only think I could think of is creating a github issue that links to the epic, but that's the same problem with more steps 😆 🤷

src/olympia/abuse/cinder.py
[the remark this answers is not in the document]
KevinMind: Okay. 👍 might be worth a docstring to clarify this is a "reactive" action.

src/olympia/abuse/cinder.py
```diff
@@ -210,12 +210,13 @@ def post_report(self, *, job):
         a keyword argument."""
         pass
 
-    def workflow_recreate(self, *, job):
+    def workflow_recreate(self, *, notes, job=None):
```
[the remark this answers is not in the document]
KevinMind: Ah I missed that one, okay so we sometimes pass the old job notes and sometimes the current decision notes. got it.

src/olympia/reviewers/tests/test_utils.py
KevinMind: maybe reference that variable then. This is confusing and makes the tests harder to maintain

src/olympia/reviewers/tests/test_utils.py
KevinMind: I'm not asking to change the test implementation, just explain why the test, as written, should pass. I'm all in favor of dry tests otherwise we would have millions more lines of code.

src/olympia/abuse/tests/test_cinder.py
KevinMind: So then passing None should just pass right? Probably not a bad idea to add that then either here or in a separate test. That way the test is trying it's hardest to fail and doesn't because our code is so awesome.

src/olympia/abuse/tests/test_cinder.py
KevinMind: Achso. Okay.

src/olympia/abuse/tests/test_cinder.py
KevinMind: Probably. But that could be a separate issue.

src/olympia/abuse/tests/test_cinder.py
eviljeff: We can filter for the activity that's created for NeedsHumanReview., sure  Like I said though, it leaves open the edge case of other activity logs being generated of a different type, especially for future refactoring. Given we don't check anywhere for the presence of an ActivtyLog of a given type and do something different (outside of obvious places like the reviewer tools, where we may show "no activity", and gc/backfill/migrations), we aren't covering any edge cases by having extra log noise; but _are_ potentially opening an edge case by only checking for a particular type.

src/olympia/reviewers/tests/test_utils.py
eviljeff: I changed the test implementation because a clearer code is better than unclear code with an explanation, imo.

KevinMind: That was ~~ironic~~ pythonic.

src/olympia/abuse/cinder.py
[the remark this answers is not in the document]
KevinMind: Nope, didn't see it was already there since it is below the comment card.

src/olympia/abuse/tests/test_cinder.py
KevinMind: This is still going over my head. 

> the edge case of other activity logs being generated of a different type

Why are we expecting that there is "no activity" here. It makes sense that we don't expect activity logs relating to NHR, which would match the narrow query I suggested. But I'm hearing you say there is an edge case that we could be creating other activity logs and I don't understand why we would care about that?

If you could explain why it's bad to have activity logs other than the NHR log, then I could get on board with it, otherwise it doesn't seem like an edge case worth protecting and our test is better fitted with a narrow query for the stuff that is actually important.

src/olympia/abuse/tests/test_models.py
KevinMind: Nice!

src/olympia/abuse/tests/test_cinder.py
eviljeff: 🤷 not a bridge I'll die on, https://github.com/mozilla/addons-server/pull/22886/commits/a1d484f13855cac8d0fa13b4eced2052a10a876c

PR #22888, opened 2024-11-24
Allow Multiple Partner Addons

chrstinalin:
Fixes mozilla/addons#14986

### Description

Allows multiple `PromotedAddon`s to be created & associated to the same `Addon`.

image
image

**Note**: Frontend breaks without https://github.com/mozilla/addons-frontend/pull/13343

### Testing

1. Navigate to Django admin panel's `promoted addon` section.
2. Multiple `PromotedAddon`s can now be created & associated to the same `Addon`.
3. Any add-ons in the queue will have all the associated badges.
4. Review pages will list the promoted type under the flags, as well as under the banner.

### Checklist


KevinMind:
So to be clear:

> Note: Frontend breaks without

Means the frontend PR should land first. right?

diox:
> Means the frontend PR should land first. right?

Yes, and as discussed during our stand-up, we should consider pushing the frontend changes for the 2024.12.12 push but *not* the addons-server changes, which are a lot more risky and should wait until next year (can land after the tag).

src/olympia/addons/serializers.py
```diff
@@ -1562,20 +1569,25 @@ def fake_object(self, data):
             # set .approved_for_groups cached_property because it's used in
             # .approved_applications.
             approved_for_apps = promoted.get('approved_for_apps')
-            obj.promoted = PromotedAddon(
-                addon=obj,
-                approved_application_ids=approved_for_apps,
-                created=None,
-                group_id=promoted['group_id'],
-            )
+            obj.promoted = [
```
willdurand, line 1572: This requires an update to the docs: https://mozilla.github.io/addons-server/topics/api/overview.html#v5-api-changelog

src/olympia/addons/serializers.py
```diff
@@ -1562,20 +1569,25 @@ def fake_object(self, data):
...
             # .appproved_applications only cares about the current group
             obj._current_version.approved_for_groups = (
-                (obj.promoted.group, APP_IDS.get(app_id))
+                (PROMOTED_GROUPS_BY_ID.get(group_id), APP_IDS.get(app_id))
                 for app_id in approved_for_apps
+                for group_id in promoted['group_ids']
             )
         else:
-            obj.promoted = None
+            obj.promoted = []
```
willdurand, line 1590: ditto (IIUC this means the API cannot return `null` anymore)

KevinMind:
@chrstinalin is this PR still ready for review? I haven't seen or done anything since before the break so can you give a 👍 👎 if it is ready?

KevinMind:
@chrstinalin any update on this one?

src/olympia/addons/models.py
```diff
@@ -1555,37 +1559,114 @@ def _is_recommended_theme(self):
...
-        from olympia.promoted.models import PromotedAddon
 
         try:
-            promoted = self.promotedaddon
-        except PromotedAddon.DoesNotExist:
-            return NOT_PROMOTED
-        is_promoted = not currently_approved or promoted.approved_applications
-        return promoted.group if is_promoted else NOT_PROMOTED
+            promoted_addons = self.promoted_addons.all()
+        except ValueError:
```
KevinMind, line 1575: Why would we expect a value error here? Can you add a comment for why that would happen?

src/olympia/addons/models.py
```diff
@@ -1555,37 +1559,114 @@ def _is_recommended_theme(self):
...
+            return []
+
+        return [
+            promoted.group
+            for promoted in promoted_addons
+            if not currently_approved or promoted.approved_applications
+        ]
+
+    def group_name(self, *, currently_approved=True):
+        """Returns the string name of the currently groups, comma separated.
```
KevinMind, line 1585: typo "currently"

KevinMind (dismissed):
Looking good so far. Some comments.

src/olympia/addons/models.py
```diff
@@ -1555,37 +1559,114 @@ def _is_recommended_theme(self):
             ).exists()
         )
 
-    def promoted_group(self, *, currently_approved=True):
+    def promoted_groups(self, *, currently_approved=True):
```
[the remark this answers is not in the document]
chrstinalin: So the complication is with `PromotedAddon`. In `promoted_groups()`, we filter on `approved_applications`. Since `approved_applications` is a `@property` function, Django would be [unable to filter by this property](https://forum.djangoproject.com/t/django-filter-can-it-handle-property-attributes/13808). In which case a loop similar to how it's implemented in this PR is done.

However, this means that a Manager on PromotedAddon couldn't be accessed (as we're returning a list), and so I'm not confident a query manager would work out.

src/olympia/addons/models.py
```diff
@@ -1555,37 +1559,114 @@ def _is_recommended_theme(self):
...
-        is_promoted = not currently_approved or promoted.approved_applications
-        return promoted.group if is_promoted else NOT_PROMOTED
+            promoted_addons = self.promoted_addons.all()
+        except ValueError:
+            return []
+
+        return [
+            promoted.group
+            for promoted in promoted_addons
+            if not currently_approved or promoted.approved_applications
```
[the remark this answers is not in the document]
chrstinalin: This should actually the most elegant solution. Since `not currently_approved` is additive (if `true`, we include _all_ groups) we'd still need to loop through all the `promoted_addons` to get all the groups associated with the add-on.

i.e the alternative would be
``` 
return [promoted.group for promoted in promoted_addons] if not currently_approved else 
       [promoted.group for promoted in promoted_addons if promoted.approved_applications]
```

KevinMind (changes requested):
My current feeling on the PR is that something is off. This touches so much of the code base and includes a pretty large migration I don't feel confident shipping in the current state. Spid-ey senses are triggering with:
- the get method and its return signature feels like the wrong abstraction. The name is confusing and the fact it returns three totally different and incompatible types just feels like a recipe for headache. I'd like to look at how we are using this to see if we can simplify that abstraction.
- The promoted_group field is used in many places and since we have a big migration, we should try to find a way to break this down. Maybe we can create a new field with the new type, and add the behavior we want around it. Then after we are happy with the behavior we can migrate the data and remove the old field. This makes it a lot safer to push changes because you don't necessarily break the whole data model if you make a mistake.

What you could definitely do now is open a PR that includes a mgiration adding "promoted_groups" as a foreign key type and logic in the model to sync the fields. When we modify the promoted_group field, also add that group to the promoted_groups. Now we have the new field and we can freely play around with it and find the right solution to this problem.

I know you are probably pretty anxious to get this merged and it has been tossed around a lot even since last year. But I'd rather not merge something that 1) has such a big impact and that 2) is hard to revert and 3) adds significant complexity top an already extremely complex code base.

PR #22896, opened 2024-11-27
Make mlbf testable

KevinMind:
Fixes: mozilla/addons#15211

### Description

Adds buttons to recreate the bloom filter from the blocklist admin (with the right permissions)

Adds a configurable replace threshold to trigger running the cron job manually.

### Context

Adding these features drastically simplifies testing the bloom filter logic via the django admin, making it more accessible for non devs.

### Change the base threshold

1. open the zadmin page
2. set the value to a low number (2)
3. make 3 changes to blocks (add/delete/update) of a given type
4. run the cron job and expect a new filter instead of a stash.
5. you can try removing or changing the value. The default is 5k

### Manually trigger mlbf

If either of the following are true, expect instead to get a 503:
- `ENABLE_ADMIN_MLBF_UPLOAD` setting is `False`
- `blocklist_mlbf_submit` waffle switch disabled
- user does not have permissions to change blocks

image

Steps to reproduce:

1. open the zadmin page <http://olympia.test/en-US/admin/models/blocklist/block/>
2. make changes that would trigger a filter/stash update
3. click the "upload mlbf" button

image

4. expect the cron was run (if there changes to the filter you can see the files are changed

image

5. try again with the force base button and expect new filters even without changes

image

### Checklist


src/olympia/blocklist/tasks.py
```diff
@@ -48,6 +48,13 @@ def BLOCKLIST_RECORD_MLBF_BASE(block_type: BlockType):
             raise ValueError(f'Unknown block type: {block_type}')
 
 
+@task
+def upload_mlbf_to_remote_settings(force_base=False):
```
diox, line 52: So this is confusing because the cron has the same name, but doesn't use this task. I don't think we want to change the cron at this time, so maybe rename the task to `upload_mlbf_to_remote_settings_trigger` or `upload_mlbf_to_remote_settings_admin` or something, and add a short comment explaining why it's there. Then you don't need to do the `as` bit in the import below as it wouldn't conflict anymore.

src/olympia/lib/settings_base.py
```diff
@@ -893,6 +893,7 @@ def get_db_config(environ_var, atomic_requests=True):
     'olympia.amo.tasks.set_modified_on_object': {'queue': 'amo'},
     'olympia.bandwagon.tasks.collection_meta': {'queue': 'amo'},
     'olympia.blocklist.tasks.cleanup_old_files': {'queue': 'amo'},
+    'olympia.blocklist.tasks.upload_mlbf_to_remote_settings': {'queue': 'zadmin'},
```
diox, line 896: Move below with the other `zadmin` routes

src/olympia/blocklist/tasks.py
KevinMind: Done.

src/olympia/blocklist/tasks.py
```diff
@@ -48,6 +48,13 @@ def BLOCKLIST_RECORD_MLBF_BASE(block_type: BlockType):
...
+    from .cron import upload_mlbf_to_remote_settings as _upload_mlbf_to_remote_settings
```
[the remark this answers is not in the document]
KevinMind: Fixed

src/olympia/blocklist/templates/admin/blocklist/block_change_list.html
```diff
@@ -9,4 +9,13 @@
...
   </li>
+  <li>
+    {% url cl.opts|admin_urlname:'upload_mlbf' as upload_mlbf_url %}
+    <a href="{% add_preserved_filters upload_mlbf_url %}" class="addlink">
+      Upload MLBF
+    </a>
+    <a href="{% add_preserved_filters upload_mlbf_url %}?force_base=true" class="addlink">
+      Upload MLBF (Force Base)
+    </a>
+  </li>
```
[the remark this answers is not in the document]
KevinMind: Done.

diox (approved):

PR #22899, opened 2024-11-28
Update block email language to accomodate possible soft blocking

KevinMind:
Fixes: mozilla/addons#15180

### DIFF
--- a/src/olympia/abuse/templates/abuse/emails/ContentActionDisableAddon.txt
+++ b/src/olympia/abuse/templates/abuse/emails/ContentActionDisableAddon.txt
@@ -4,5 +4,5 @@ Your {{ type }} {{ name }} was manually reviewed by the Mozilla Add-ons team {% 
 Our review found that your content violates the following Mozilla policy or policies:
 {% include 'abuse/emails/includes/policies.txt' %}
 
-Based on that finding, your {{ type }} has been permanently disabled on {{ target_url }} and is no longer available for download from Mozilla Add-ons, anywhere in the world. {% if is_addon_being_blocked %}In addition, in some cases, users who have previously installed those versions won't be able to continue using them.{% else %}Users who have previously installed your add-on will be able to continue using it.{% endif %}
+Based on that finding, your {{ type }} has been permanently disabled on {{ target_url }} and is no longer available for download from Mozilla Add-ons, anywhere in the world. {% if is_addon_being_blocked %}In addition, in some cases, users who have previously installed the addon-on won't be able to continue using it.{% else %}Users who have previously installed your add-on will be able to continue using it.{% endif %}
 {% endblock %}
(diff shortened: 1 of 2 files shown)


src/olympia/abuse/templates/abuse/emails/ContentActionDisableAddon.txt
```diff
@@ -4,5 +4,5 @@ Your {{ type }} {{ name }} was manually reviewed by the Mozilla Add-ons team {%
 Our review found that your content violates the following Mozilla policy or policies:
 {% include 'abuse/emails/includes/policies.txt' %}
 
-Based on that finding, your {{ type }} has been permanently disabled on {{ target_url }} and is no longer available for download from Mozilla Add-ons, anywhere in the world. {% if is_addon_being_blocked %}In addition, users who have previously installed the add-on won't be able to continue using it.{% else %}Users who have previously installed your add-on will be able to continue using it.{% endif %}
+Based on that finding, your {{ type }} has been permanently disabled on {{ target_url }} and is no longer available for download from Mozilla Add-ons, anywhere in the world. {% if is_addon_being_blocked %}In addition, in some cases, users who have previously installed those versions won't be able to continue using them.{% else %}Users who have previously installed your add-on will be able to continue using it.{% endif %}
```
diox, line 7: So, I suspect we can't actually reach this:
- When an add-on gets force disabled, which would trigger this email, we don't set `is_addon_being_blocked`.
- When an add-on gets disabled as part of a block, we currently avoid calling `force_disable()`, so we don't send this email. Instead, we do call `reject_multiple_versions()` which would trigger the other email below.

That being said, if it was possible to reach it, in this case, we're disabling the entire add-on and not specifying affected versions like in the other case below. Therefore, unlike in the other email, the words `those versions` are not referencing anything, we should keep using `the add-on` and `it` (the entire add-on is being blocked).

src/olympia/abuse/templates/abuse/emails/ContentActionDisableAddon.txt
```diff
@@ -4,5 +4,5 @@ Your {{ type }} {{ name }} was manually reviewed by the Mozilla Add-ons team {%
 Our review found that your content violates the following Mozilla policy or policies:
 {% include 'abuse/emails/includes/policies.txt' %}
 
-Based on that finding, your {{ type }} has been permanently disabled on {{ target_url }} and is no longer available for download from Mozilla Add-ons, anywhere in the world. {% if is_addon_being_blocked %}In addition, users who have previously installed the add-on won't be able to continue using it.{% else %}Users who have previously installed your add-on will be able to continue using it.{% endif %}
-{% endblock %}
\ No newline at end of file
+Based on that finding, your {{ type }} has been permanently disabled on {{ target_url }} and is no longer available for download from Mozilla Add-ons, anywhere in the world. {% if is_addon_being_blocked %}In addition, in some cases, users who have previously installed the addon-on won't be able to continue using it.{% else %}Users who have previously installed your add-on will be able to continue using it.{% endif %}
```
diox, line 7:
```suggestion
Based on that finding, your {{ type }} has been permanently disabled on {{ target_url }} and is no longer available for download from Mozilla Add-ons, anywhere in the world. {% if is_addon_being_blocked %}In addition, in some cases, users who have previously installed the add-on won't be able to continue using it.{% else %}Users who have previously installed your add-on will be able to continue using it.{% endif %}
```

diox (approved):

PR #22900, opened 2024-11-28
Don't show duplicate flags for NeedsHumanReview reasons already flagged explicitly

diox:
Those new `NeedsHumanReview` reasons get a flag because they are in `VIEW_QUEUE_FLAGS`. That is useful for the form to filter the queue for those reasons, but they don't need to show the corresponding sprite icon, because we also already display one by checking `AddonReviewerFlags`/`PromotedAddon` state.

Fixes: mozilla/addons#15072 (see screenshots posted by QA)

### DIFF
--- a/static/css/zamboni/reviewers.less
+++ b/static/css/zamboni/reviewers.less
@@ -56,6 +56,11 @@
 .ed-sprite-needs-human-review-from-abuse { background-position: 0 -468px; }
 .ed-sprite-needs-human-review-from-cinder { background-position: 0 -484px; }
 .ed-sprite-needs-human-review-from-appeal { background-position: 0 -500px; }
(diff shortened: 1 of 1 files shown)


eviljeff (approved):
This works.... but really they shouldn't be in `VIEW_QUEUE_FLAGS` if there isn't a flag for them in the view queue.  i.e. we should structure the constants differently.

Given this is for a late cherry-pick I won't block r+ on it though.

diox:
Yeah I think the proper fix would be to make the form depend on something else than `VIEW_QUEUE_FLAGS` - perhaps it should call `get_due_date_reason_q_objects()` directly and extract all the keys ? But then we wouldn't have the descriptions... I wonder if we could somehow annotate what `get_due_date_reason_q_objects()` returns with descriptions...

I'll file a follow-up

PR #22906, opened 2024-12-02
Record counts for review queues, including detail per promoted class

diox:
Fixes https://github.com/mozilla/addons/issues/15173

### Testing

`manage.py cron record_reviewer_queues_counts` should populate the database with `QueueCount` instances for the current date. There should be one instance per day per queue, plus one per promoted class. Re-running the cron for a given day should have no effect.


src/olympia/reviewers/utils.py
```diff
@@ -166,6 +167,7 @@ class Meta(PendingManualApprovalQueueTable.Meta):
 
     @classmethod
     def get_queryset(cls, request, **kw):
+        # Note: this must match what record_metrics() is doing.
```
KevinMind, line 170: Add a reference to the file. `olympia.amo.cron.record_metrics` or something like that.

src/olympia/amo/cron.py
```diff
@@ -92,3 +94,31 @@ def write_sitemaps(section=None, app_name=None):
                 continue
             content = sitemap_object.render(app_name=_app_name, page=_page)
             sitemap_file.write(content)
+
+
+def record_metrics():
```
KevinMind, line 99: I think we can do better than the comment to tie these two interfaces to each other. This is just an idea but lmk what you think:

1. define an ENUM representing the "queue" names. `ReviewQueueEnum`
2. use that enum in the [table dictionary](https://github.com/mozilla/addons-server/blob/c192a390b92b6a412ac81eba37e8c2ed309a10b2/src/olympia/reviewers/views.py#L418-L426) to specify relevant queue tables

3. use the same enum here to define the counts.

Currently it's not clear to me which key here maps to which table. I would need to read each table to find that. But if you have an enum connecting them, then I can trace the enum to the 2 interfaces. In a world where we have a rest interface the same can apply there as well.

src/olympia/amo/models.py
```diff
@@ -659,3 +659,16 @@ class GroupConcat(models.Aggregate):
 class Inet6Ntoa(Func):
     function = 'INET6_NTOA'
     output_field = CharField()
+
+
+class Metric(models.Model):
```
KevinMind, line 664: I'm really not sure why we need a new model for this. It seems we are keeping track of counts so we can:
- use grafana counters
- logs
- google analytics (maybe not server side but you get the idea)

Models tend to expand and harden so if we can avoid adding one, I would.

KevinMind (commented):
Few comments.

src/olympia/amo/models.py
diox: If this was just for engineering grafana or logs would be fine, but it's for Operations/Product, and their tool of choice is Redash. Notably because it allows correlation with the rest of the data available there

src/olympia/amo/cron.py
```diff
@@ -92,3 +94,31 @@ def write_sitemaps(section=None, app_name=None):
...
+    for group in PROMOTED_GROUPS:
+        if group != NOT_PROMOTED:
+            querysets[f'review_queue_{group.api_name}'] = querysets[
+                'review_queue'
+            ].filter(promotedaddon__group_id=group.id)
+
+    # Execute a count for each queryset and record a Metric instance for it.
+    for key, qs in querysets.items():
+        Metric.objects.get_or_create(
+            name=key, date=today, defaults={'value': qs.count()}
```
diox, line 123: FIXME: need `construct_count_queryset_from_queryset` trick for perf

src/olympia/amo/cron.py
diox: https://github.com/mozilla/addons-server/pull/22906/commits/c9076a48aaf1b77a1cf7de476e296a37c7b09288 goes in that direction; not completely finished yet, need to clean up a couple things.

I've switched back to the tables being the source of truth, even though the manager/queryset method is the actual source of truth, the way I'm handling this in each table should be more DRY.

diox: So, now:
- `reviewer_tables_registry` is where we specify queues
- Each queue table has a `name` and a `get_queryset()` method, with an optional `request` parameter to filter the queue according to the request/user.
- The `count()` optimization has been moved to its own queryset method on our base queryset class, in a dedicated method, `optimized_count()`.
- `record_metrics` simply loops over the registry, and for each queue in there, gets the queryset and calls `optimized_count()` on it. It re-uses the queue `name` as the key under which to record the value.

src/olympia/lib/settings_base.py
```diff
@@ -1500,17 +1500,18 @@ def read_only_mode(env):
...
 CRON_JOBS = {
-    'update_addon_average_daily_users': 'olympia.addons.cron',
-    'update_addon_weekly_downloads': 'olympia.addons.cron',
     'addon_last_updated': 'olympia.addons.cron',
-    'update_addon_hotness': 'olympia.addons.cron',
     'gc': 'olympia.amo.cron',
-    'write_sitemaps': 'olympia.amo.cron',
     'process_blocklistsubmissions': 'olympia.blocklist.cron',
-    'upload_mlbf_to_remote_settings': 'olympia.blocklist.cron',
+    'record_metrics': 'olympia.amo.cron',
```
diox, line 1506: This is the only change, rest is re-ordering alphabetically

src/olympia/amo/models.py
KevinMind: Does redash not have access to other data sources? That's surprising to me but TIL. I want to open a slack thread about this but won't stop the PR for it. Just want to make sure we draw attention to this as it isn't great that this is the only way to get this data..

KevinMind: Nit if we do indeed go this route. Metric is a vague name, and this should not be a vague model. I don't think we should open flood gates for adding all kinds of metrics and such... feels like the creation of `ActivityLog` if I can imagine how that went. We should avoid like plague and make this a hyper-specific model that resists being used for other purposes `ReviewQueueOptimizedCount` feels specific enough. It should also be in the review app so as not to be considered a general purpose thing.

Should other purposes arise we should double down on finding a solution in the place where it belongs, which is in redash or big query or some tool made for collecting metrics on existing data not our production database.

src/olympia/reviewers/views.py
```diff
@@ -186,12 +186,16 @@ def dashboard(request):
     view_all = any(
         acl.action_allowed_for(request.user, perm) for perm in view_all_permissions
     )
-    queue_counts = fetch_queue_counts(request)
+    queue_counts = {
+        tab: queue.get_queryset(request).optimized_count()
```
KevinMind, line 190: here you are using the key from the `reviewer_tables_registry` and in the cron you are using the name property on the value.. why not use one or the other in both places? Do we actually need the name property if we already have a key from .items()

src/olympia/amo/models.py
diox: Redash does have access to other sources but they either expire (statsd pings) or are not more practical than our database (bigquery).

src/olympia/reviewers/utils.py
```diff
@@ -113,18 +114,24 @@ class Meta(AddonQueueTable.Meta):
...
 
     @classmethod
-    def get_queryset(self, request, *, upcoming_due_date_focus=False, **kw):
-        show_only_upcoming = upcoming_due_date_focus and not acl.action_allowed_for(
-            request.user, amo.permissions.ADDONS_ALL_DUE_DATES
-        )
-        qs = Addon.unfiltered.get_queryset_for_pending_queues(
-            admin_reviewer=is_admin_reviewer(request.user),
-            show_temporarily_delayed=acl.action_allowed_for(
+    def get_queryset(cls, request, *, upcoming_due_date_focus=False, **kwargs):
```
KevinMind, line 117: nit: is there a reason not to simply define the 3 relevant properties as args and let the view pass in the arguments from else and the cron pass in the hard coded values. IMO that makes it super clear at the call site what is going on, whereas now both are kind of obfuscated.. If there is a good reason to keep the request in context then fine, but otherwise that seems like the better way to go, don't you think?

src/olympia/amo/cron.py
KevinMind: This looks promising. I would still unify on the name property a bit more:
- move the formatted name passing the hard coded name string to the table class and just map the registry to produce the dict
- If we have the formatted name on the class, we could include this in the Metric instance and then dashboards can represent queue names exactly as they are in reviewer tools (ux benefit IMO)
- the cron should compile the querysets the same way the view does. Either use the tab, queue in registry.items() as in the view or ditch the tab entirely if we have the name. It's confusing to have both since they are not the same thing. e.g. is Metric.name referring to the tab key in the source of truth dict or the name of the table class?

src/olympia/reviewers/views.py
diox: Yeah I don't love this, but here is the reason to my madness:
- The registry being a dict matter because we need to easily be able to find a queue by name in `queue()` (it does `TableObj = reviewer_tables_registry[tab]`) and also to show the correct tab in `queue.html`
- We need a `name` on the table anyway because we need to name the URL in `reviewers/urls.py` - that is also referenced in `queue.html`:
```
<ul class="tabnav">
  {% for this, loc, text in queue_tabnav(registry) %}
  <li class="{% if tab==this %}selected{% endif %}"><a href="{{ url('reviewers.%s' % loc) }}">{{ text }}</a></li>
  {% endfor %}
</ul>
```

Previously, there was a bit of duplication between `urlname` being set and used for URLs, but the name of the tab defined elsewhere, in the registry, and sometimes the values were close, sometimes not, which was super confusing. So this is my attempt to reconcile the 2...

You're right that I made `dashboard` inconsistent with the cron. I could use `queue.name` instead of `tab` here for consistency, that seems better.

src/olympia/amo/models.py
KevinMind: >  not more practical than our database

What's the argument for that statement? From where I'm standing it sounds like this is literally the use case for big query.

src/olympia/amo/cron.py
diox: (I've made the querysets looping more consistent in https://github.com/mozilla/addons-server/pull/22906/commits/d7e318844145f8e8c7d7a9ea968ba6531ca1afce)

By formatted name, you mean the `title` ? We already have it on the classes, so that's easy to do. I see the benefit but I'm worried about it not being super stable, making data comparisons in the future harder if we decide to change the wording. The `name` is a completely internal value we are less likely to tweak.

src/olympia/amo/models.py
diox: AMO is not currently set up to write to BigQuery - at the moment it only uses it to read stats from, which are generated outside of AMO completely. And most of the data we'd want to correlate these metrics with exist on AMO side (activity log etc).

KevinMind: That's probably something we should look into.. I'm not sure it really matters where other data we want to correlate to is, that's the benefit of a "data lake" tool like redash. But fine for now then. Wdyt about the specificication of the model then?

diox: I had missed your comment about the model being too vague. That's fair, I was worried we'd immediately need more data, which is why I wanted to be quite generic from the beginning (though I did make `value` an integer in an attempt to make it too generic and end up like `ActivityLog`). I don't mind making the name more specific to force us to use it correctly, I'll think about it a bit more.

diox: 1ff5be491c57e4c5c09fc52b0707b18e289ae4bc renames and moves the model & the cron

src/olympia/reviewers/utils.py
KevinMind: I'm not sure why it needs to be 2 methods. it could just be one method with the defined arguments.

```python
get_queryset(
    admin_reviewer=False,
    show_temporarily_delayed = True,
    show_only_upcoming = False,
)
```

Then in the view we can pull the arguments from the request.

```python
admin_reviewer = is_admin_reviewer(request.user)
show_temporarily_delayed = acl.action_allowed_for(
    request.user, amo.permissions.ADDONS_TRIAGE_DELAYED
)
show_only_upcoming = upcoming_due_date_focus and not acl.action_allowed_for(
    request.user, amo.permissions.ADDONS_ALL_DUE_DATES
)

table = TheClass.get_queryset(
    admin_reviewer=admin_reviewer,
    show_temporarily_delayed=show_temporarily_delayed,
    show_only_upcoming=show_only_upcoming,
)

... do stuff with table
````

And in the cron we just directly set whatever arguments we want... would something like this not work and achieve my goal of keeping the control logic at the call site? You can say yes and also not implement it, but wondering if this would work in your view?

diox: Something like that would work, yes, but I was trying to avoid more duplication of code.  The code to check for `show_temporarily_delayed` / `admin_reviewer` is needed in the queue view... but also the dashboard, which fetches the counts (personalized for the current user according to what they'd be able to see). The single `get_queryset()` method avoids that duplication currently.

KevinMind (commented):
Okay. I won't die on this hill but I think that kind of code duplication is totally okay and arguably better than including additional (unnecessary) context into the "service" layer. Keeping that layer args only not only simplifies testing (just pass all combination of arguments and verify the results, no need to mock requests) it also exposes the actual expectations each caller has for the service, since the code is defined and readable at the point of dependency. This is totally worth sacrificing one unit of DRY imo.

I will verify and approve. Speaking of... how should I verify the counts themselves are correct. Are we expecting the length of the queue in reviewer tools should match that of the QueueCount instance? How would one verify this using the admin?

diox:
> I will verify and approve. Speaking of... how should I verify the counts themselves are correct. Are we expecting the length of the queue in reviewer tools should match that of the QueueCount instance? How would one verify this using the admin?

At the most basic level, you need a bunch of add-ons in the queues, which can be achieved by:
- Submitting themes / theme versions, for the theme queue
- Submitting extensions / extensions versions *and* forcing their versions not to be auto-approved and marked as needing human review, for the manual review queue (quickest way is probably to add `NeedsHumanReview` instance for each version - `reason` doesn't matter, as long as `is_active` stays `True` - and don't auto-approve them)

Then if you are a superuser (so you have all permissions) the count in the dashboard for each queue should match the `value` in the corresponding `QueueCount` for that day, as well as the number of visible items in the queue page itself (including pagination if needed)

(You can't verify this in the admin, have to open a django shell)

There are a lot more combinations since depending on permissions you should not see the same contents in some queues (but the `QueueCount` should stay the same) and more queues than that, but I think this would be a lot easier to fully verify on dev (which has all the relevant data) if the logic looks ok.

KevinMind (approved):
Agree, this requires a lot of preparation data already covered in the tests so let's verify on dev.

PR #22927, opened 2024-12-06
Record review queue history: when a version enters the queue, when they leave it

diox:
Fixes https://github.com/mozilla/addons/issues/15174

A future task could be to even replace `Version.due_date` by this new table entirely, but let's see if we can make that work first...

### Testing

https://github.com/mozilla/addons/issues/15174#issuecomment-2567795552


src/olympia/reviewers/migrations/0027_backfill_needshumanreview.py
```diff
@@ -5,13 +5,7 @@
...
 
 def migrate_needs_human_review_field(apps, schema_editor):
-    NeedsHumanReview = apps.get_model('reviewers', 'NeedsHumanReview')
-    Version = apps.get_model('versions', 'Version')
-
-    for version in Version.unfiltered.filter(needs_human_review=True).filter(
-        Q(needshumanreview__isnull=True) | Q(needshumanreview__is_active=False)
-    ):
-        NeedsHumanReview.objects.create(version=version)
+    pass
```
diox, line 8: I was getting migration conflicts because of this old migration, given it's a data migration and has already been executed in every env it shouldn't matter if it's blank.

eviljeff:
> A future task could be to even replace `Version.due_date` by this new table entirely

How would this work?  I understood `original_due_date` to be the original, unchanged, due_date, meaning `Version.due_date` could be updated?

diox:
>> A future task could be to even replace `Version.due_date` by this new table entirely
>
> How would this work? I understood `original_due_date` to be the original, unchanged, due_date, meaning `Version.due_date` could be updated?

I was thinking we'd rename `ReviewQueueHistory` to `ReviewQueue` and move `due_date` field from `Version` to that model (so we'd have both the original and the current due dates). But that's just a random idea, we don't have to do that at all, and certainly not now - it would be annoying to do for little to no benefit at the moment.

eviljeff (approved):

PR #22930, opened 2024-12-09
Embed build info into the docker image instead of relying on runtime variables.

KevinMind:
Fixes: mozilla/addons#15230

### Description

- Modified `Dockerfile` to expose build arguments and create a static build info file, ensuring that build variables are hard-coded and not overridden at runtime.
- Refactored `get_version_json` function in `utils.py` to read build information from the newly created static file, ensuring required keys are validated.
- Updated tests in `test_apps.py` and `test_utils.py` to check for required version keys and validate the build info retrieval process.

### Context

Previously relying on runtime environment variables led to less consistent checks because runtime variables can change, where the underlying image cannot. Since we are mostly checking the underlying image, we should not rely on the runtime variables but instead the image itself.

### Testing

1. verify the image contains the /build-info file with the correct information. Set values for all the variables to verify they can be configured appropriately
2. Modifying one of the variables after the build should not invalidate most caches.

### Checklist

### DIFF
--- a/Makefile-os
+++ b/Makefile-os
@@ -7,6 +7,9 @@
 DOCKER_PROGRESS ?= auto
 DOCKER_METADATA_FILE ?= buildx-bake-metadata.json
 DOCKER_PUSH ?=
+# Not in dot env saved,
+# Docker needs these values set,
+# Static, cache preserved.
 export DOCKER_COMMIT ?=
 export DOCKER_BUILD ?=
 export DOCKER_VERSION ?=
--- a/src/olympia/core/apps.py
+++ b/src/olympia/core/apps.py
@@ -12,7 +12,7 @@ from django.core.management.base import CommandError
 from django.db import connection
 from django.utils.translation import gettext_lazy as _
 
-from olympia.core.utils import get_version_json
+from olympia.core.utils import REQUIRED_VERSION_KEYS, get_version_json
 
 
 log = logging.getLogger('z.startup')
@@ -64,11 +64,9 @@ def host_check(app_configs, **kwargs):
 @register(CustomTags.custom_setup)
 def version_check(app_configs, **kwargs):
     """Check the (virtual) version.json file exists and has the correct keys."""
-    required_keys = ['version', 'build', 'commit', 'source']
-
     version = get_version_json()
 
-    missing_keys = [key for key in required_keys if key not in version]
+    missing_keys = [key for key in REQUIRED_VERSION_KEYS if key not in version]
 
     if missing_keys:
         return [
@@ -85,8 +83,10 @@ def version_check(app_configs, **kwargs):
 def static_check(app_configs, **kwargs):
     errors = []
     output = StringIO()
+    version = get_version_json()
 
-    if settings.DEV_MODE:
+    # We only run this check in production images.
+    if version.get('target') != 'production':
         return []
 
     try:
--- a/src/olympia/core/utils.py
+++ b/src/olympia/core/utils.py
@@ -5,15 +5,25 @@ import sys
 import django
 
 
-def get_version_json():
-    contents = {}
+# Keys required to be set in the version.json file.
+REQUIRED_VERSION_KEYS = ['target', 'version', 'source', 'commit', 'build']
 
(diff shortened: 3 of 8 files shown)


src/olympia/core/utils.py
```diff
@@ -5,15 +5,32 @@
...
+    # Read the build info from the docker image.
+    # This is static read only data that cannot
+    # be overridden at runtime.
+    if os.path.exists(build_info_path):
+        with open(build_info_path) as f:
+            for line in f:
+                key, value = line.strip().split('=', 1)
+                if value.startswith('"') and value.endswith('"'):
+                    value = value[1:-1]
+                contents[key] = value
```
eviljeff, line 33: 😬 I'm not a fan of this manually written parser... is there no scope for using a format that has a library to reliably parse it - e.g. yaml, json, etc?

src/olympia/core/utils.py
```diff
@@ -5,15 +5,32 @@
 import django
 
 
-def get_version_json():
-    contents = {}
+# Keys exempt from inspection in local images as they must be set.
+EXEMPT_REQUIRED_KEYS = ['target', 'version', 'source']
```
eviljeff, line 9: I don't see where this is used independently - what is it for?

KevinMind: Right you are.

src/olympia/core/apps.py
```diff
@@ -12,7 +12,10 @@
 from django.db import connection
 from django.utils.translation import gettext_lazy as _
 
-from olympia.core.utils import get_version_json
+from olympia.core.utils import (
+    REQUIRED_VERSION_KEYS,
+    get_version_json,
+)
```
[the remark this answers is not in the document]
KevinMind: Let's see

Makefile-os
```diff
@@ -7,6 +7,10 @@
 DOCKER_PROGRESS ?= auto
 DOCKER_METADATA_FILE ?= buildx-bake-metadata.json
 DOCKER_PUSH ?=
+# Values that are not saved to .env
+# but should be set in the docker image
+# default to static values to prevent
+# invalidating docker build cache
```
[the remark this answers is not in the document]
KevinMind: Not intended, but now it is an actual haiku 👍

src/olympia/core/utils.py
KevinMind: Replaced with json.

src/olympia/core/utils.py
```diff
@@ -5,15 +5,27 @@
...
+    contents = {key: '' for key in REQUIRED_VERSION_KEYS}
+
+    # Read the build info from the docker image.
+    # This is static read only data that cannot
+    # be overridden at runtime.
+    if os.path.exists(build_info_path):
+        with open(build_info_path) as f:
+            data = json.loads(f.read())
+            for key, value in data.items():
+                contents[key] = value
```
eviljeff, line 28: nit: you could probably replace this with something like:
```suggestion
            contents.update(json.loads(f.read()))
```
?

tests/make/make.spec.js
```diff
@@ -220,6 +220,36 @@ describe('docker-compose.yml', () => {
...
+
+    describe.each(['development', 'production'])(
+      'DOCKER_TARGET=%s',
+      (DOCKER_TARGET) => {
+        const FILTERED_KEYS = [
+          'DOCKER_COMMIT',
+          'DOCKER_VERSION',
+          'DOCKER_BUILD',
+        ];
+        it('.services.(web|worker).environment excludes build info variables', () => {
```
eviljeff, line 232: I forgot to add this last time: I'm not clear what this is testing, exactly.  What aren't we including in the web/worker envs, and why?

eviljeff (approved):

tests/make/make.spec.js
KevinMind: We exclude variables that could be interpreted as build info such as the DOCKER_COMMIT, DOCKER_VERSION and (eventually) DOCKER_TARGET. You don't want to live in a world where you build with one commit/version and then run the container with another value set (arbitrarily) and then your container thinks it is a different image.. This test confirms that the variables that are used in the build info file are NOT available on the environment of the container.

PR #22934, opened 2024-12-10
Split CI dependencies and add ./deps:/deps mount.

KevinMind:
Fixes: mozilla/addons#15234
Fixes: mozilla/addons#15034

### Description

- Introduced a new script `install_deps.py` to streamline the installation of pip and npm dependencies based on specified targets.
- Updated `docker-compose.yml` to include a new volume for dependencies and adjusted service configurations accordingly.
- Modified `Makefile-docker` to incorporate checks for CI dependencies and streamline the update process.

### Context

Adds a dependency mount to load dependencies from the host. Additionally optimizes the logic for installing dependencies in one batch.

Which dependencies to run is set during make up and not "mutable" because we pass it to the container via the .env file. to change it, re-run make up with a different value.

> [!NOTE]
> You can cheat and install development depenednecies in a production running container is you run the command inside the container shell, where make will accept your OLYMPIA_DEPS=development argument

### Testing

The value of `OLYMPIA_DEPS` defaults to the value of `DOCKER_TARGET` but can be overriden so there are 4 different scenarios.

### Standard development mode

```bash
```

Expect:
- volume called `data_deps` is created
- it is synced to your ./deps folder.
- containers start up and install prod/dev dependencies.
- Theoretically this should re-enable tracking in IDE
- running `make check` in the container should pass

### Dev mode with Prod dependencies

```bash
make up DOCKER_TARGET=development OLYMPIA_DEPS=production
```

Expect:
- everything from above, but missing development dependencies.
- pytest will not work

> [!IMPORTANT]
> You will see in the logs during make up that the ./deps directory is cleaned. That is because we are running a development image but specifying we want production dependencies. To ensure we don't end up with stale dependencies we have to clean up the directory and freshly install, otherwise some dev dependencies could be left behind. we DON'T do this on local images that are installing development dependencies as we should not need any removed

### Prod mode

This should happen if the `DOCKER_TARGET` is `production`

```bash
```

Expect:
- Everything from above AND
- ./deps directory is cleared and dependencies are freshly installed
- expect only prod dependencies to be installed
- verify this with `make check` and (for example) that debug toolbar is missing on the web
- no pytest or ruff
- verify with `make check`

### Prod mode with Dev dependencies

This does not currently work on a running container We set the OLYMPIA_DEPS variable during startup and make does not support passing arguments from the host makefile, into the container, and into the host again, at least not easily.

So to run prod mode with testing, run a production build with dev dependencies.

```bash
make up DOCKER_TARGET=production OLYMPIA_DEPS=development
```

This will build the production image and install the development dependencies at runtime.

Expect:
- Everything from above AND
- now you should be able to run pytest and ruff

### Checklist

### DIFF
--- a/Makefile-docker
+++ b/Makefile-docker
@@ -18,6 +18,13 @@ REQUIRED_FILES := \
 	/deps/package-lock.json \
 	/addons-server-docker-container \
 
+# Build list of dependencies to install
+DEPS = pip prod
+# If we're running a development image, then we should install the development dependencies
+ifeq ($(OLYMPIA_DEPS), development)
+DEPS += dev
+endif
+
 .PHONY: help_redirect
 help_redirect:
 	@$(MAKE) help --no-print-directory
@@ -28,12 +35,9 @@ check_debian_packages: ## check the existence of multiple debian packages
 
 .PHONY: check_pip_packages
 check_pip_packages: ## check the existence of multiple python packages
-	@ ./scripts/check_pip_packages.sh prod.txt
-# "production" corresponds to the "propduction" DOCKER_TARGET defined in the Dockerfile
-# When the target is "production" it means we cannot expect dev.txt dependencies to be installed.
-	@if [ "$(DOCKER_TARGET)" != "production" ]; then \
-		./scripts/check_pip_packages.sh dev.txt; \
-	fi
+	@for dep in $(DEPS); do \
+		./scripts/check_pip_packages.sh $$dep.txt; \
+	done
 
 .PHONY: check_files
 check_files: ## check the existence of multiple files
@@ -76,6 +80,10 @@ update_assets:
 	# Collect static files: This MUST be run last or files will be missing
 	$(PYTHON_COMMAND) manage.py collectstatic --noinput
 
+
+.PHONY: update_deps
+update_deps: ## Update the dependencies
+	$(HOME)/scripts/install_deps.py $(DEPS)
 
 # TOOD: remove this after we migrate addons-frontned to not depend on it.
 .PHONY: setup-ui-tests
--- a/.dockerignore
+++ b/.dockerignore
@@ -28,7 +28,7 @@
 backups
 build*.py
 buildx-bake-metadata.json
-deps
+deps/*
 docker*.yml
 docker/artifacts/*
 docs/_build
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -10,9 +10,6 @@ inputs:
   version:
     required: true
     description: The image version to tag with
-  target:
-    required: true
-    description: The stage to target in the build
   push:
     required: false
     description: Push the image?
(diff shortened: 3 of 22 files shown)


KevinMind:
@diox we can simplify this if we decide it's okay to add everything in ci.txt to prod.txt but that means those dependencies ship to production. I'm happy either way.

Makefile-docker
```diff
@@ -28,12 +28,12 @@ check_debian_packages: ## check the existence of multiple debian packages
...
-# When the target is "production" it means we cannot expect dev.txt dependencies to be installed.
-	@if [ "$(DOCKER_TARGET)" != "production" ]; then \
-		./scripts/check_pip_packages.sh dev.txt; \
-	fi
+	./scripts/check_pip_packages.sh pip.txt
+	./scripts/check_pip_packages.sh prod.txt
+# We expect dev dependencies only if DOCKER_TARGET is not production
+ifeq ($(DOCKER_TARGET), development)
+	./scripts/check_pip_packages.sh dev.txt
+endif
```
eviljeff, line 36: my Makefile foo is non-master level, but should these lines be indented like they were before?

Makefile-docker
```diff
@@ -77,6 +77,25 @@ update_assets:
...
+
+.PHONY: update_deps
+update_deps: ## Update the dependencies
+# If we're running a non local or production image, remove the existing deps and install clean
+ifneq ($(and $(findstring local, $(DOCKER_TAG)), $(findstring development, $(DOCKER_TARGET))),)
+	@echo "Updating existing deps"
+else
+	@echo "Removing existing deps"
+	rm -rf /deps/**
+endif
```
eviljeff, line 95: similarly these lines don't appear to be indented

Makefile-docker
```diff
@@ -77,6 +77,25 @@ update_assets:
 	$(PYTHON_COMMAND) manage.py collectstatic --noinput
 
 
+# Build list of dependencies to install
+DEPS = pip prod
+# If we're running a development image, then we should install the development dependencies
+ifeq ($(DOCKER_TARGET), development)
+DEPS += dev
+endif
```
eviljeff, line 85: Should these lines be at the top of the file with other file-level constants, or inside `update_deps`?  (they're only referenced in `update_deps`

eviljeff:
@KevinMind CI is failing with ```ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref d292b4cc-acf7-406d-b50f-82233a76aea4::5khdhdz6z7m0rspdr4p5sj9wn: "/requirements/ci.txt": not found```

Makefile-docker
KevinMind: The if statements don't have to be indendted and common make examples show these statements non indented. Make is quite strange in parsing but this works and I'd keep it this way.

Makefile-docker
KevinMind: Good point. I moved them up and also that means we can use them for the check command also.

Makefile-docker
KevinMind: conditionals inside of recipes should not be indented

KevinMind:
> @KevinMind CI is failing with `ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref d292b4cc-acf7-406d-b50f-82233a76aea4::5khdhdz6z7m0rspdr4p5sj9wn: "/requirements/ci.txt": not found`

I split off the second part of this patch as it was not strictly necessary and left that line. It's gone now.

eviljeff:
I just did `make up` and got

> Error response from daemon: failed to mount local volume: mount /run/desktop/mnt/host/wsl/docker-desktop-bind-mounts/Ubuntu/3827bdbcbfb76834bf7a50fd8746e67e97d66dbb404c4ef6a62eef6bf805c55f:/var/lib/docker/volumes/addons-server_data_deps/_data, flags: 0x1000: no such file or directory

(I didn't do anything from the testing section, to check the scenario for a typical use just after the patch has been merged onto master) (edited)

KevinMind:
@eviljeff can you try now? I added a .gitkeep in the ./deps directory. If that dir doesn't exist then docker will bail creating the volume.

eviljeff (approved):
works locally but:
- I expected `DOCKER_TARGET=production` to also remove the javascript formatting utilities (it didn't), not just `ruff`
- CI is failing

PR #22937, opened 2024-12-11
Mark orphaned Cinder Policies as such when syncing them, delete the ones we don't use

diox:
Fixes https://github.com/mozilla/addons/issues/15000


src/olympia/abuse/models.py
```diff
@@ -933,6 +933,7 @@ class CinderPolicy(ModelBase):
     default_cinder_action = models.PositiveSmallIntegerField(
         choices=DECISION_ACTIONS.choices, null=True, blank=True
     )
+    present_in_cinder = models.BooleanField(null=True)
```
eviljeff, line 936: mmm, what's the thinking behind the field being `present_in_cinder` (and `True` for policies in Cinder right now) vs. `orphaned` or similar (and `False` for policies in Cinder right now)?  

I don't think there's much/any different at a model/database level, but the operational logic and messaging seems to all be about orphaning but the field name is the opposite way around.

eviljeff (approved):

src/olympia/abuse/models.py
diox: I figured `orphaned` wasn't descriptive enough, and by default the admin shows booleans as green and red icons... greens for truthy, red for falsy, so I like to avoid using a negative boolean value to describe a "positive" or "normal" state.

PR #22946, opened 2024-12-13
Do not include current extension in 'Other popular extensions'

chrstinalin:
Fixes: mozilla/addons#1569

### Description

Filters out the current extension from 'other popular extensions' when using the fallback extensions.
image
image

### Testing

`TAAR_LITE_FALLBACKS` uses GUIDs from production, so I'm actually unsure what the proper way of testing this should be. You can mimick the behaviour locally by replacing the GUIDs in `TAAR_LITE_FALLBACKS` with local ones (similar to the screenshot)

### Checklist

### DIFF
--- a/src/olympia/addons/tests/test_utils_.py
+++ b/src/olympia/addons/tests/test_utils_.py
@@ -148,15 +148,19 @@ class TestGetAddonRecommendations(TestCase):
         assert reason == TAAR_LITE_FALLBACK_REASON_INVALID
         assert incr_mock.call_count == 0
         # Fallback filters out the current guid if it exists in TAAR_LITE_FALLBACKS
-        recommendations, _, _ = get_addon_recommendations_invalid(TAAR_LITE_FALLBACKS[0])
+        recommendations, _, _ = get_addon_recommendations_invalid(
+            TAAR_LITE_FALLBACKS[0]
+        )
         assert recommendations == TAAR_LITE_FALLBACKS[1:]
-    
+
     def test_get_filtered_fallbacks(self, _):
         # Fallback filters out the current guid if it exists in TAAR_LITE_FALLBACKS
         recommendations = get_filtered_fallbacks(TAAR_LITE_FALLBACKS[2])
         assert recommendations == TAAR_LITE_FALLBACKS[:2] + TAAR_LITE_FALLBACKS[3:]
         # Fallback returns the first four if it does not.
-        recommendations, outcome, reason = get_addon_recommendations_invalid('random-guid')
+        recommendations, outcome, reason = get_addon_recommendations_invalid(
+            'random-guid'
+        )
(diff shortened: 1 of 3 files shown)


src/olympia/addons/tests/test_utils_.py
```diff
@@ -103,7 +104,7 @@ def test_recommended(self, incr_mock):
     def test_recommended_no_results(self, incr_mock):
         self.recommendation_server_mock.return_value = []
         recommendations, outcome, reason = get_addon_recommendations('a@b', True)
-        assert recommendations == TAAR_LITE_FALLBACKS
+        assert recommendations == TAAR_LITE_FALLBACKS[:4]
```
KevinMind, line 107: This is a bit ambiguous.. I guess you are slicing the array because you expect the sliced item to be the "excluded" addon. Is there a more explicit way of doing that?

Maybe assert that the one you expect not to be there is not in there.

```python
assert "foo" not in recommendations
```

That is easier to understand and a more precise test.

KevinMind (commented):
A few comments

src/olympia/addons/tests/test_utils_.py
chrstinalin: I feel like the existing test works better in context, in that "if there's no results, it only takes the first four fallbacks".

I added a comment and I can still add that test, but arbitrarily setting the precedent to test for n-many fallbacks not being in it seems unneccessary?

src/olympia/addons/utils.py
```diff
@@ -90,22 +91,27 @@ def get_addon_recommendations(guid_param, taar_enable):
...
+def get_addon_recommendations_invalid(current_guid=None):
     return (
-        TAAR_LITE_FALLBACKS,
+        get_filtered_fallbacks(current_guid),
         TAAR_LITE_OUTCOME_REAL_FAIL,
         TAAR_LITE_FALLBACK_REASON_INVALID,
     )
 
 
+def get_filtered_fallbacks(current_guid=None):
```
KevinMind, line 110: It's unclear why we need to return only 4 items.  If this is a constraint on the FE we should handle it there. This just seems arbitrary to loop an entire list of static items and then slice the first 4 items.

src/olympia/addons/tests/test_utils_.py
diox: That was from the original product decision. We know the frontend should never display more than 4, and the fewer the API returns, the faster it's going to be (those responses can be quite verbose, so any kilobyte counts).

https://mozilla.github.io/addons-server/topics/api/addons.html#recommendations documents that it will return 4 max.

KevinMind: Ok. That's rather unclear then. Maybe a comment in the function linking to the docs would help.

KevinMind (approved):
Have not verified but can be done on dev/stage

PR #22951, opened 2024-12-16
Use monitors instead of custom healthcheck.py script to verify service dependencies' health

KevinMind:
Fixes: mozilla/addons#15249
Relates to: mozilla/addons#15248

### Description

the custom healthcheck.py script was returning false positives. Instead of fixing this scripts, it makes more sense to use the existing infrastructure we have from monitors to ensure services are running correctly.

### Context

We can use a management command running a set of monitors on a timeout to verify service dependencies are up. This is useful in the initializion command itself that depdends on the database being up, and after the initialization command where we want to ensure several critical dependencies are up before letting make up return successfully.

### Testing

With a running environment you can run

```bash
./manage.py monitors
```

1. stop the worker container.. the monitor should fail
2. add an exception to the __version__ view function.. the monitor should fail
3. stop the mysql container.. the monitor should fail

Any combination of these should result in a failure. You can adjust the attempts parameter or the services parameter to check different services.

### Checklist


docker-compose.yml
```diff
@@ -56,9 +56,10 @@ services:
       "DJANGO_SETTINGS_MODULE=settings",
       "watchmedo",
       "auto-restart",
-      "--directory=/data/olympia/src",
+      "--directory=/data/olympia",
```
diox, line 59: I know `src/` means we don't restart on `settings.py` (or `local_settings.py`) changes, but this was like that before because otherwise there are too many files to monitor as it includes `static/`, `node_modules/`, `storage/` ... (The limit of how many files you can monitor varies depending on the host but we need to keep it as low as possible)

src/olympia/amo/monitors.py
```diff
@@ -31,6 +31,28 @@ def execute_checks(checks: list[str]):
     return status_summary
 
 
+def web():
```
diox, line 34: As discussed on slack let's rename that to `localdev_web` to be less confusing

src/olympia/amo/monitors.py
```diff
@@ -31,6 +31,28 @@ def execute_checks(checks: list[str]):
...
+    return status, None
+
+
+def worker():
+    """
+    Check if the celery worker is healthy.
+    """
+    status = ''
+    result = celery.result.AsyncResult('healthcheck')
+    if result.failed():
```
diox, line 50: As discussed on slack: I'm Not sure this does anything useful, does it actually call the worker ?

diox (commented):
Still playing with it locally, some comments

src/olympia/amo/monitors.py
KevinMind: Done.

src/olympia/amo/monitors.py
KevinMind: Fixed. `celery.current_app.inspect().ping()` faithfully returns None if the worker is shut down.

docker-compose.yml
KevinMind: Reverted

diox: Looks like it's still here ?

src/olympia/core/apps.py
```diff
@@ -125,18 +125,26 @@ def static_check(app_configs, **kwargs):
...
-        cursor.execute("SHOW VARIABLES LIKE 'character_set_database';")
-        result = cursor.fetchone()
-        if result[1] != settings.DB_CHARSET:
-            errors.append(
-                Error(
-                    'Database charset invalid. '
-                    f'Expected {settings.DB_CHARSET}, '
-                    f'recieved {result[1]}',
-                    id='setup.E005',
+    try:
```
KevinMind, line 128: If you "stop" the mysql service and run

```bash
./manage.py monitors --service olympia_database
```

The command raises here as it is a check. we want to return the error object which means that check should not raise.

In general, this django check should also not raise but return the error object.

diox (approved):
r+ once CI pass

PR #22956, opened 2024-12-20
Support Markdown in Add-on Listing Fields

chrstinalin:
Fixes: mozilla/addons#15145

### Description

Replaces HTML syntax support with Markdown support in the `description`, `developer_comments`, `eula`, and `privacy_policy` fields of `Addon`. Existing HTML in those fields will remain functional until the field is edited. No new HTML will be functional.

Only the markdown-equivalent of [currently allowed HTML attributes](https://github.com/mozilla/addons-server/blob/ffd993045cef73fb7635e4a94ac95e1a3a9073ec/src/olympia/devhub/templates/devhub/includes/macros.html#L6) are allowed. `abbr` uses the [PHP Markdown Extra syntax](https://python-markdown.github.io/extensions/abbreviations/).

#### DevHub:

image
image
image

#### Frontend:

image
image
image

### Testing

**New Add-ons**
1. Create a listed add-on.
2. Under 'submit/details', description should have the tooltip and support markdown.

**Existing Add-ons**
1. Navigate to an add-on that already has HTML in its description. The HTML should still be functional. The HTML is still rendered on the frontend listing.
3. Under the 'edit' page, the description and developer comments should support markdown syntax.*
4. Under the 'Manage Authors & License' page, the EULA and Privacy Policy fields should support markdown syntax.*

\* Any user-inputted HTML saved at this point is rendered as plaintext. Any allowed markdown in the field is rendered as expected in the frontend. Disallowed markdown is stripped from the result and not visible at all (this is to avoid showing the user the markdown formatted into HTML).

### Checklist

### DIFF
--- a/src/olympia/devhub/templates/devhub/includes/macros.html
+++ b/src/olympia/devhub/templates/devhub/includes/macros.html
@@ -10,7 +10,7 @@
 </p>
 {% endmacro %}
 
-{% macro markdown_tip(title=None) %}
+{% macro supported_syntax_tip(title=None) %}
 <p class="syntax-support">
 {# L10n: %s is a list of markdown syntax. #}
 <span class="tooltip" title="{% if title %}{{ title }}{% else %}{{
--- a/src/olympia/devhub/templates/devhub/includes/macros.html
+++ b/src/olympia/devhub/templates/devhub/includes/macros.html
@@ -10,6 +10,17 @@
 </p>
 {% endmacro %}
 
+{% macro markdown_tip(title=None) %}
+<p class="html-support">
+{# L10n: %s is a list of HTML tags. #}
+<span class="tooltip" title="{% if title %}{{ title }}{% else %}{{
+      _('Allowed HTML: %(allowed_html)s', allowed_html='<a href title> <abbr title> <acronym title> <b>
+                             <blockquote> <code> <em> <i> <li> <ol> <strong> <ul>')|safe
+    }}{% endif %}">{{ _('Some Markdown supported.') }}</span>
+</p>
+{% endmacro %}
+
+
 {% macro empty_unless(truthy) %}
   {% if truthy %}
     {{ caller() }}
--- a/src/olympia/devhub/templates/devhub/includes/macros.html
+++ b/src/olympia/devhub/templates/devhub/includes/macros.html
@@ -1,7 +1,7 @@
 {% extends "includes/forms.html" %}
 
 {% macro some_html_tip(title=None) %}
-<p class="html-support">
+<p class="syntax-support">
 {# L10n: %s is a list of HTML tags. #}
 <span class="tooltip" title="{% if title %}{{ title }}{% else %}{{
       _('Allowed HTML: %(allowed_html)s', allowed_html='<a href title> <abbr title> <acronym title> <b>
@@ -11,11 +11,10 @@
 {% endmacro %}
 
 {% macro markdown_tip(title=None) %}
-<p class="html-support">
-{# L10n: %s is a list of HTML tags. #}
+<p class="syntax-support">
+{# L10n: %s is a list of markdown syntax. #}
 <span class="tooltip" title="{% if title %}{{ title }}{% else %}{{
-      _('Allowed HTML: %(allowed_html)s', allowed_html='<a href title> <abbr title> <acronym title> <b>
-                             <blockquote> <code> <em> <i> <li> <ol> <strong> <ul>')|safe
+      _('Allowed Markdown: %(allowed_markdown)s', allowed_markdown='links, abbr,  bold, italics, blockquote, code, ol, ul')
     }}{% endif %}">{{ _('Some Markdown supported.') }}</span>
 </p>
 {% endmacro %}
--- a/src/olympia/devhub/templates/devhub/includes/macros.html
+++ b/src/olympia/devhub/templates/devhub/includes/macros.html
@@ -1,15 +1,5 @@
 {% extends "includes/forms.html" %}
 
-{% macro some_html_tip(title=None) %}
-<p class="syntax-support">
-{# L10n: %s is a list of HTML tags. #}
-<span class="tooltip" title="{% if title %}{{ title }}{% else %}{{
-      _('Allowed HTML: %(allowed_html)s', allowed_html='<a href title> <abbr title> <acronym title> <b>
-                             <blockquote> <code> <em> <i> <li> <ol> <strong> <ul>')|safe
-    }}{% endif %}">{{ _('Some HTML supported.') }}</span>
-</p>
-{% endmacro %}
-
 {% macro supported_syntax_tip(allowed_markdown, title=None) %}
 <p class="syntax-support">
 {# L10n: %s is a list of markdown syntax. #}
--- a/src/olympia/devhub/templates/devhub/includes/macros.html
+++ b/src/olympia/devhub/templates/devhub/includes/macros.html
@@ -10,11 +10,11 @@
 </p>
 {% endmacro %}
 
-{% macro supported_syntax_tip(title=None) %}
+{% macro supported_syntax_tip(allowed_markdown, title=None) %}
 <p class="syntax-support">
 {# L10n: %s is a list of markdown syntax. #}
 <span class="tooltip" title="{% if title %}{{ title }}{% else %}{{
-      _('Allowed Markdown: %(allowed_markdown)s', allowed_markdown='links, abbr,  bold, italics, blockquote, code, ol, ul')
+      _('Allowed Markdown: %(allowed_markdown)s', allowed_markdown=allowed_markdown)
     }}{% endif %}">{{ _('Some Markdown supported.') }}</span>
 </p>
 {% endmacro %}
--- a/src/olympia/translations/models.py
+++ b/src/olympia/translations/models.py
@@ -232,9 +232,15 @@ class PurifiedMarkdownTranslation(PurifiedTranslation):
 
     def clean_localized_string(self):
         # bleach user-inputted html
-        cleaned = bleach.clean(self.localized_string, tags=[], attributes={})
+        cleaned = (
+            bleach.clean(self.localized_string, tags=[], attributes={})
+            if self.localized_string
+            else ''
+        )
         # hack; cleaning breaks blockquotes
-        markdown = md.markdown(cleaned.replace('&gt;', '>'), extensions=['abbr'])
+        markdown = md.markdown(
+            cleaned.replace('&gt;', '>'), extensions=['abbr', 'nl2br']
+        )
 
         # Keep only the allowed tags and attributes, strip the rest.
         cleaner = bleach.Cleaner(
--- a/src/olympia/translations/models.py
+++ b/src/olympia/translations/models.py
@@ -238,9 +238,8 @@ class PurifiedMarkdownTranslation(PurifiedTranslation):
             else ''
         )
         # hack; cleaning breaks blockquotes
-        markdown = md.markdown(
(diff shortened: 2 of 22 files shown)


chrstinalin:
There should probably be a more informative way to explain what syntax or Markdown is supported, possibly similar to how GitHub explains it via the 'Markdown is Supported' button by navigating to a [dedicated page](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax).

src/olympia/translations/models.py
```diff
@@ -209,21 +216,42 @@ def clean(self):
...
             tags=self.allowed_tags,
             attributes=self.allowed_attributes,
-            filters=[linkify_filter],
+            filters=[self.linkify_filter],
         )
 
         return cleaner.clean(str(self.localized_string))
 
 
+class PurifiedMarkdownTranslation(PurifiedTranslation):
```
KevinMind, line 229: Let's cover this with tests.

src/olympia/devhub/templates/devhub/includes/macros.html
```diff
@@ -10,6 +10,16 @@
 </p>
 {% endmacro %}
 
+{% macro markdown_tip(title=None) %}
+<p class="syntax-support">
+{# L10n: %s is a list of markdown syntax. #}
+<span class="tooltip" title="{% if title %}{{ title }}{% else %}{{
+      _('Allowed Markdown: %(allowed_markdown)s', allowed_markdown='links, abbr,  bold, italics, blockquote, code, ol, ul')
```
KevinMind, line 17: allowed_markdown should be in a constant somewhere. If these are in fact the only tags allowed then use that same constant in actual processing.

src/olympia/devhub/templates/devhub/includes/macros.html
```diff
@@ -10,6 +10,16 @@
 </p>
 {% endmacro %}
 
+{% macro markdown_tip(title=None) %}
```
KevinMind, line 13: I like what you have done elsewhere migrating from "html" to "syntax" let's do that here also "supported_syntax_tip" is a more generic name that won't again need to be migrated and encapsulates what you mean about markdown without coupling the name to the language.

src/olympia/translations/models.py
```diff
@@ -209,21 +216,42 @@ def clean(self):
...
+    def clean_localized_string(self):
+        # bleach user-inputted html
+        cleaned = (
+            bleach.clean(self.localized_string, tags=[], attributes={})
+            if self.localized_string
+            else ''
+        )
+        # hack; cleaning breaks blockquotes
+        text_with_brs = cleaned.replace('&gt;', '>')
+        markdown = md.markdown(text_with_brs, extensions=['abbr', 'fenced_code'])
```
KevinMind, line 242: It's not clear why we are passing an array with "abbr" and "fenced_code" here.. can you add a comment?

KevinMind (commented):
Looks promising. A few comments before testing.

KevinMind:
Please add testing steps and I can verify the PR.

src/olympia/devhub/templates/devhub/includes/macros.html
chrstinalin: Just since you pointed this out --

> There should probably be a more informative way to explain what syntax or Markdown is supported, possibly similar to how GitHub explains it via the 'Markdown is Supported' button by navigating to a [dedicated page](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax).

I mentioned this to Andrew offhandedly as, especially with the extensions, it may not be obvious what Markdown syntax is allowed in what way, and he mentioned maybe making it a page on the extension workshop over the tooltip. What do you think?

KevinMind: I'm definitely in favor of that yes. I don't think it is directly related to my ask though. I want to make sure that the tags specified in this tooltip are "bound" to the tags we actually filter for, so that if one were to change, the other is automatically synced.

It would be a bad UX if the tooltip allowed links, but actually we don't. Would it be possible to extract this list to a constant and the `", ".join(ALLOWED_MARKDOWN_TAGS)` in the template?

And then where we actually validate the input, we can use the same constant.

src/olympia/translations/tests/test_models.py
```diff
@@ -701,6 +702,57 @@ def test_external_text_link(self, get_outgoing_url_mock):
         assert doc('b')[0].text == 'markup'
 
 
+class PurifiedMarkdownTranslationTest(TestCase):
```
KevinMind, line 705: nit: Typically I think adding `Test` to the beginning is better than the end as this class now matches the class it is testing in a find/replace

image

KevinMind:
@chrstinalin what should I expect to happen if I enter invalid or disallowed markdown syntax? Can you add that to the test scenario. Additionally can you clarify what "functional" means? I think you mean the syntax should be "rendered" on the page right?

src/olympia/translations/tests/test_models.py
```diff
@@ -701,6 +702,57 @@ def test_external_text_link(self, get_outgoing_url_mock):
...
+    @patch('olympia.amo.urlresolvers.get_outgoing_url')
+    def test_external_text_link(self, get_outgoing_url_mock):
+        get_outgoing_url_mock.return_value = 'http://external.url'
+        s = '**markup** http://example.com'
+        x = PurifiedMarkdownTranslation(localized_string=s)
+        doc = pq(x.__html__())
+        links = doc('a[href="http://external.url"][rel="nofollow"]')
+        assert links[0].text == 'http://example.com'
+        assert doc('strong')[0].text == 'markup'
+
```
KevinMind, line 754: Can you add a test of mixed html and markdown, simulating someone editing an existing html string and adding markdown to it. We would expect the html to be stringified and the markdown to be formatted to html right?

src/olympia/devhub/templates/devhub/includes/macros.html
KevinMind: It looks like we have such a constant in the translations model file already "allowed_tags"

src/olympia/translations/tests/test_models.py
chrstinalin: `test_markdown` does have a case of mixed HTML and markdown, and yeah -- the HTML is stringified, the markdown is formatted.

src/olympia/devhub/templates/devhub/includes/macros.html
chrstinalin: I had meant it in a replace-the-tooltip-entirely way, but I figure there's no harm in having both.

src/olympia/translations/tests/test_models.py
KevinMind: ah I missed that in the name `test_mixed_markdown_html` maybe?

KevinMind:
After modifying existing html I can confirm:
- before modification the html is rendered
- after the modification it is stringified

image

Should we expect the "stringified" href attribute to still be rendered as a link though? it is.. I'm not sure if that is expected or not.

KevinMind:
I'm seeing a template tag instead of actually allowed markdown tags

image

KevinMind:
Do I still expect to see html supported for version notes?

image

KevinMind:
@chrstinalin I cannot find the developer comments section.. can you help navigating there?

KevinMind:
There is no way to preview content in the authors page. I can add markdown but after saving I still see the textarea so I have no idea what it would actually render as

image

KevinMind:
@chrstinalin something that would be very helpful when verifying this is if you could share a sample of both html and markdown that include a sample of all the supported tags.

I even wonder if having this available to developers on a page could be useful. There are many flavors of markdown and looking at the tag, it's hard for me (with context) to know what and how I should input the syntax to get the expected result.

Imagine how difficult it would be for a developer.

KevinMind (changes requested):
Still needs some work. A few bugs discovered and some edge cases needing addressing.

Also found a UX hangup that might be worth addressing in terms of usability of this feature.

chrstinalin:
> Should we expect the "stringified" href attribute to still be rendered as a link though? it is.. I'm not sure if that is expected or not.

Currently, if you were to put just a link in, it'd render similarly. If we're essentially treating user HTML as plaintext from now on, I can't see why this wouldn't be expected?

> Do I still expect to see html supported for version notes?

I had omitted it originally since it wasn't part of the original issue. Now that I think about it, though, there's probably no reason to do that. @abyrne-moz should add-on version notes also make the switch from HTML to Markdown?

> I cannot find the developer comments section.. can you help navigating there?

At the very bottom of 'Edit Product Page', under 'Technical Details'.

> There is no way to preview content in the authors page. I can add markdown but after saving I still see the textarea so I have no idea what it would actually render as

As it did with HTML before, too. It did bother me, though. It shouldn't hurt to just add the rendering.

> something that would be very helpful when verifying this is if you could share a sample of both html and markdown that include a sample of all the supported tags.

```
Some of these words *are emphasized*.
Some of these words _are emphasized also_.

Use two asterisks for **strong emphasis**.
Or, if you prefer, __use two underscores instead__.

Here's a link: [Google](https://www.google.com "Click me!").

**abbr, acronym**

The HTML specification is maintained by the W3C.
*[HTML]: Hyper Text Markup Language
*[W3C]:  World Wide Web Consortium

**codeblocks, blockquotes**

Here is code:

    tell application "Foo"
        beep
    end tell

> This is the first level of quoting.
> > This is nested blockquote.

**Lists**

*Ordered List:*

1. First item
2. Second item
3. Third item

*Unordered List:*

- Item 1
- Item 2
- Item 3

This is an [example link](http://example.com/)

<a href="www.google.com">This shouldnt work!</a>

<ol>
<li>Neither</li>
<li>Should</li>
<li>This</li>
</ol>
```

> I even wonder if having this available to developers on a page could be useful. There are many flavors of markdown and looking at the tag, it's hard for me (with context) to know what and how I should input the syntax to get the expected result.

Yeah, which is why I mentioned it [before](https://github.com/mozilla/addons-server/pull/22956#discussion_r1905792997). Though I imagined something like:
1) Merge this PR
2) Merge a PR to `extension-workshop` with syntax information
3) Create a PR to link to said information.
Being a smoother way about it?

KevinMind:
Thanks for the updates. Everything checks out except this part:

> Currently, if you were to put just a link in, it'd render similarly. If we're essentially treating user HTML as plaintext from now on, I can't see why this wouldn't be expected?

If I render this on a browser:

```
<html>
<body>
<a href="http://github.com/mozilla/addons-server">
Github repo
</a>
</body>
</html>
```

I do not expect a clickable link to appear on the screen. I expect (just like github is doing, to just render out all the markup as plain text.

I kind of have the same expectations for our "plain text" html where we render links. As you can see the html syntax is rendered as plain text, but somehow, the link is a clickable link.. that seems odd.

Is that actually expected behaviour? why plaint text the links then?

chrstinalin:
I do agree its a bit weird. But I'd still say its expected as a result of the [linkify](https://github.com/mozilla/addons-server/blob/52b004f660cf8bc8823839f33428d2aa87507822/src/olympia/translations/models.py#L213) bleach converting [any string that looks like a URL](https://bleach.readthedocs.io/en/latest/linkify.html#bleach.linkify).

KevinMind:
Do you mean the "whiteboard" if that also supports markdown I would expect the same tooltip we have in the other inputs.

image

KevinMind (approved):
Verified the specified input sources as correctly formatting markdown and plaint text ifying html

image

PR #22957, opened 2024-12-30
Introduce vite to transpile js/css assets for development and production + vitest for static file testing

KevinMind:
Relates to: mozilla/addons#2000

### Description

> [!NOTE]
> Don't be scared, most of the changes are package-lock.json 

- Introduce vite and vitest targeting only the `preload.js` and `analytics.js` scripts initially.
- add infrastructure to support building and serving vite files

### Context

This PR introduces the "plumbing" needed for compiling assets with vite. It does not migrate all assets to using vite. We should split this up after landing this and we can move assets over to esm with vite.

### Vite

Vite will now compile every file in `./static/js/*.js` and `./static/cs/*.less` into independent bundles. Those files can then be included in our django app via the helper method `vite_asset`.

For example:

```jinja
{{ vite_asset('js/preload.js') }}
{{ vite_asset('css/common.less') }}
```

This will render roughly to this:

```html
<script src="<vite_asset_url>" />
<link href="<vite_asset_url" />
```

In development mode, `vite_asset_url` will resolve to `/static/bundle/<path>` where path is the original path to the url.
Nginx routes `/static/bundle/` to the vite dev server which will then return the asset compiled in real time. LESS will return CSS and JS will return transpiled js.

In production mode, vite will compile the assets and save them in `./site-static/assets/` and a manifest will be used to resolve the url to the file path. Nginx will mount this path and serve files directly. In production environments, like before we will serve files from a CDN.

### Dev mode

Run the dev/prod version of the app and make sure that:
- assets are being served successfully
- no console errors or runtime errors in the modified javascript
- all functionality remains unregressed. 

```bash
```

### Prod mode

```bash
```

### Automated tests

Run the following tests which are now using vitest instead of jest

- `make test_setup`
- `make run_js_tests`

### Checklist


src/olympia/lib/settings_base.py
```diff
@@ -674,7 +680,6 @@ def get_db_config(environ_var, atomic_requests=True):
         'preload': (
```
diox, line 680: Could you rename this bundle to something like `jquery_base` or something and update references to it in `base.html` and `devhub/index.html` ? I find it confusing that it's still called `preload` but doesn't include the new `preload.js` file...

KevinMind: Yes. FWIW, the goal is to eventually include all of that into a single preload bundle but that doesn't work when we are not bundling everything with vite so would happen later.

src/olympia/lib/settings_base.py
```diff
@@ -1332,7 +1337,12 @@ def read_only_mode(env):
 NODE_PACKAGE_JSON = path('package.json')
 NODE_PACKAGE_MANAGER_INSTALL_OPTIONS = ['--dry-run']
 
+# The manifest file is created in static-build but copied into the static root
+# so we should expect to find it at /<static_root/<static_build>/manifest.json
 STATIC_BUILD_PATH = path('static-build')
+# This value should be kept in sync with vite.config.js
+# where the manifest will be written to
+STATIC_BUILD_MANIFEST_PATH = path(STATIC_BUILD_PATH, 'manifest.json')
```
diox, line 1345: A bit unfortunate that `vite` calls this `manifest.json`, same as webextensions manifest, but at least it doesn't conflict with django's `staticfiles.json` ...

diox (approved):

src/olympia/lib/settings_base.py
KevinMind: we can make the file whatever we want. I could make it "vite-manifest.json" or "banana.json" as long as the file names are aligned in both places it's up to us.

KevinMind: https://github.com/mozilla/addons/issues/15326

PR #22958, opened 2025-01-02
Fix python version used by dependabot

diox:
Fixes mozilla/addons#15265 (hopefully)


eviljeff:
Although we probably want to document this magic file/value somewhere we'd notice, for when python3.12 moves out of security support...

diox:
I've added a comment in the file, which [should be supported](https://github.com/dependabot/dependabot-core/pull/9519). I'm not too worried about finding that file in the future when upgrading Python, I would expect us to grep for `3.12` when upgrading, and that should happen way before 3.12 goes out of support.

PR #22967, opened 2025-01-03
Refactor ContentDecision to be used consistently

eviljeff:
Fixes: mozilla/addons#15220

### Description

Significant refactoring of how we create `ContentDecision`s, and how they're "processed", so we can a) use a single workflow for decisions that originate in Cinder and the reviewer tools.

With this patch we no longer call the task with an activity log and create the `ContentDecision` later - instead we create the `ContentDecision` in `record_decision` and pass that id.  

The main changes are 
- in abuse/models.py 
  - CinderJob.resolve_job is no more - all the code is in ContentDecision; 
  - the cinder functionality in notify_reviewer_decision is now in report_to_cinder; 
  - and execute_action_and_notify - nee process_action - can handle decisions from anywhere
- and then in reviewers/utils
  - record_decision creates the ContentDecision and then creates the activity log before triggering the task

### Context

One of the blockers to be able to hold and release decisions made via the reviewer tools.  Also contributes to us having less separate code-paths we have to maintain.

### Testing

Mainly these changes impact reviewer tools originated decisions, so to focus on that:
- reject a version in the reviewer tools
- see the developer receives the expected email notification, that includes the version numbers + review text (the policy text from the policies should not be repeated in the email text)
  - (more advanced case: delay reject and reject multiple versions instead)
- repeat while resolving a cinder job at the same time
  - to get a CinderJob, report an add-on for abuse inside the add-on, as a policy violation
  - the outstanding job should create a NeedsHumanReview that causes the add-on to be in the review queue
- the add-on should be cleared from the queue after the review
- abuse reporter should get an email (as before) too

To verify processing decisions originated in Cinder hasn't regressed:
- report an add-on for abuse on the AMO detail page or under another policy so it appears in the Cinder queue for extension abuse
- make a decision to disable the add-on in Cinder
- replay that decision locally
- check the add-on was disabled, and the emails were sent to the reporter and developer as now.

### Checklist


src/olympia/reviewers/utils.py
```diff
@@ -1441,10 +1463,24 @@ def reject_multiple_versions(self):
...
+                log.info('Sending email for %s' % (self.addon))
+                self.record_decision(
+                    action_id,
+                    log_action_kw={
+                        'versions': versions,
+                        'timestamp': now,
+                        'user': user,
+                        'extra_details': extra_details,
+                    },
+                )
```
eviljeff, line 1483: This is one part I'm not completely happy with: it works, but ends up creating separate decisions - and sending separate emails - for the different users and activity types.  I feel at least the different users can be combined.

diox: Ultimately, aren't those decisions separate anyway ? They were made at different times, we are just applying them together.

eviljeff: Originally, yes - and for each of the (delayed) rejection decisions they would have gotten an email at the time.  It's just whether we attempt to combine the eventual takedown into a single (or fewer) decisions.  But unless you feel very strongly that we need a single decision and email we can decide what else (if anything) to do in a follow-up.

diox: I feel (not very strongly) that multiple decisions is fine - the fact that it results in multiple emails is annoying but not the end of the world. IMHO this is an edge case anyway, that we don't really need to worry about too much.

src/olympia/abuse/models.py
```diff
@@ -1233,130 +1204,100 @@ def appeal(self, *, abuse_report, appeal_text, user, is_reporter):
...
-            action_helper.notify_owners(log_entry_id=getattr(log_entry, 'id', None))
-            self.save(update_fields=('action_date',))
-        else:
-            action_helper.hold_action()
+
+            if self.addon_id:
+                details = (log_entry and log_entry.details) or {}
+                # Using 'reviewtype' is a bit of a hack to allow us to differentiate
+                # decisions from the reviewer tools, because it's only set there.
+                from_reviewer_tools = 'reviewtype' in details
```
diox, line 1259: Don't love this... It's fairly fragile. Not that much better but could we use `human_review` instead ? `reviewtype` is kinda obsolete these days...

At the minimum add a comment in `ReviewBase.log_action()` to make sure we continue setting this in the future...

diox (approved):

src/olympia/abuse/models.py
eviljeff: mmm, forgot the comment.  I'll add it with https://github.com/mozilla/addons/issues/1982

PR #22972, opened 2025-01-06
Change Wording of Escalated Appeals in Version History

chrstinalin:
Fixes: mozilla/addons#14809

### Description

Adds CINDER_APPEAL_ESCALATION to distinguish appeal escalations.

### Checklist


src/olympia/abuse/cinder.py
```diff
@@ -515,10 +515,10 @@ def flag_for_human_review(self, *, related_versions, appeal=False, forwarded=Fal
         from olympia.reviewers.models import NeedsHumanReview
 
         waffle_switch_name = (
-            'dsa-appeals-review'
-            if appeal
-            else 'dsa-cinder-forwarded-review'
+            'dsa-cinder-forwarded-review'
             if forwarded
+            else 'dsa-appeals-review'
+            if appeal
```
eviljeff, line 521: mmm, was this change necessary?

eviljeff (changes requested):
We need to change VersionManager.get_due_date_reason_q_objects too (look for where CINDER_ESCALATION is referenced)

src/olympia/abuse/cinder.py
chrstinalin: If `CINDER_APPEAL_ESCALATION` is meant to behave virtually identical to `CINDER_ESCALATION` (beyond the history), it would make sense for this waffle switch name to be the same, no? If it's the other way around, `CINDER_APPEAL_ESCALATION` would have 'dsa-appeals-review' as the name instead.

eviljeff: I can see the logic (and don't disagree), but one of reasons we have these series of waffles switches it the rollout is being phased, and reviewers handling appeals is a later phase than handling forwarded reports.  So for now I we're going to have to stick with the inconsistency.

src/olympia/versions/models.py
```diff
@@ -223,7 +223,10 @@ def get_due_date_reason_q_objects(cls):
             # date even if the version has been disabled / not signed.
             'needs_human_review_from_cinder': Q(
                 needshumanreview__is_active=True,
-                needshumanreview__reason=NeedsHumanReview.REASONS.CINDER_ESCALATION,
+                needshumanreview__reason__in={
+                    NeedsHumanReview.REASONS.CINDER_ESCALATION,
+                    NeedsHumanReview.REASONS.CINDER_APPEAL_ESCALATION,
```
eviljeff, line 228: There could be a follow-up to break these out - so they can be filtered separately - but this is fine for now.

eviljeff (approved):
r+wc - just revert the waffle switch logic and we're gtg

src/olympia/users/utils.py
```diff
@@ -61,7 +61,10 @@ def get_task_user():
     """
     from olympia.users.models import UserProfile
 
-    return UserProfile.objects.get(pk=settings.TASK_USER_ID)
+    try:
+        return UserProfile.objects.get(pk=settings.TASK_USER_ID)
+    except UserProfile.DoesNotExist:
+        return None
```
eviljeff, line 67: We don't want this change - if there's a test failing because the task user doesn't exist, just create a user profile with `TASK_USER_ID` as the id in the test. (use `user_factory`)

eviljeff (approved):

PR #22975, opened 2025-01-07
flag correct versions for a developer appeal

eviljeff:
Fixes: mozilla/addons#15182 and fixes mozilla/addons#15247

### Description

Flag the versions that were affected by a version rejection if the developer appeals.

### Context

Until recently there was no easy way to reference the versions rejected by the ContentDecision, so we could only flag the current version if the developer appealed the rejection. 

### Testing

- enable all the DSA related waffle switches and have a Cinder API key in your settings
- Reject a version of an add-on in the reviewer tools - either an unlisted version or a non-latest version - i.e. a version that wouldn't be the addon's current_version (unlisted is better, because that's the use-case the issue was filed for)
- appeal that decision with the appeal link in the email
- go back to the reviewer tools and see that the appealed version is flagged for NHR

### Checklist

### DIFF
--- a/src/olympia/abuse/tests/test_cinder.py
+++ b/src/olympia/abuse/tests/test_cinder.py
@@ -1230,7 +1230,10 @@ class TestCinderAddonHandledByReviewers(TestCinderAddon):
         assert addon.current_version.reload().due_date
 
     def test_appeal_specific_version_from_report(self):
-        # version_string is set for reporter appeals on reports that specified a version
+        """For an appeal from a reporter, version_string is set on the CinderClass
+        instance, from AbuseReport.addon_version_string. If version_string is defined,
+        and the version exists, we should flag that version rather than current_version.
+        """
         addon = self._create_dummy_target()
         other_version = version_factory(
             addon=addon,
@@ -1252,7 +1255,9 @@ class TestCinderAddonHandledByReviewers(TestCinderAddon):
         assert other_version.reload().due_date
 
     def test_appeal_specific_version_from_action(self):
-        # for developer appeals we should use the versions that were affected
+        """For an appeal from a developer, version_string will be None on the
+        CinderClass instance. If version_string is falsely we collect and flag the addon
+        versions from the appealled decision rather the current_version."""
         addon = self._create_dummy_target()
         other_version = version_factory(
             addon=addon,
--- a/src/olympia/abuse/tests/test_cinder.py
+++ b/src/olympia/abuse/tests/test_cinder.py
@@ -136,9 +136,14 @@ class BaseTestCinderCase:
     def test_report(self):
         self._test_report(self._create_dummy_target())
 
-    def _test_appeal(self, appealer, cinder_instance=None):
-        fake_decision_id = 'decision-id-to-appeal-666'
-        cinder_instance = cinder_instance or self.CinderClass(
+    def _test_appeal(
+        self,
+        appealer,
+        *,
+        cinder_entity_instance=None,
+        appealed_decision_id='decision-id-to-appeal-666',
+    ):
+        cinder_entity_instance = cinder_entity_instance or self.CinderClass(
             self._create_dummy_target()
         )
 
@@ -149,8 +154,8 @@ class BaseTestCinderCase:
             status=201,
         )
         assert (
-            cinder_instance.appeal(
-                decision_cinder_id=fake_decision_id,
+            cinder_entity_instance.appeal(
+                decision_cinder_id=appealed_decision_id,
                 appeal_text='reason',
                 appealer=appealer,
             )
@@ -163,8 +168,8 @@ class BaseTestCinderCase:
             status=400,
         )
         with self.assertRaises(ConnectionError):
-            cinder_instance.appeal(
-                decision_cinder_id=fake_decision_id,
+            cinder_entity_instance.appeal(
+                decision_cinder_id=appealed_decision_id,
                 appeal_text='reason',
                 appealer=appealer,
             )
@@ -1204,7 +1209,8 @@ class TestCinderAddonHandledByReviewers(TestCinderAddon):
     def test_appeal_anonymous(self):
         addon = self._create_dummy_target()
         self._test_appeal(
-            CinderUnauthenticatedReporter('itsme', 'm@r.io'), self.CinderClass(addon)
+            CinderUnauthenticatedReporter('itsme', 'm@r.io'),
+            cinder_entity_instance=self.CinderClass(addon),
         )
         assert (
             addon.current_version.needshumanreview_set.get().reason
@@ -1214,14 +1220,17 @@ class TestCinderAddonHandledByReviewers(TestCinderAddon):
 
     def test_appeal_logged_in(self):
         addon = self._create_dummy_target()
-        self._test_appeal(CinderUser(user_factory()), self.CinderClass(addon))
+        self._test_appeal(
+            CinderUser(user_factory()), cinder_entity_instance=self.CinderClass(addon)
+        )
         assert (
             addon.current_version.needshumanreview_set.get().reason
             == NeedsHumanReview.REASONS.ADDON_REVIEW_APPEAL
         )
         assert addon.current_version.reload().due_date
 
-    def test_appeal_specific_version(self):
+    def test_appeal_specific_version_from_report(self):
+        # version_string is set for reporter appeals on reports that specified a version
         addon = self._create_dummy_target()
         other_version = version_factory(
             addon=addon,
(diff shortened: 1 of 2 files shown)


src/olympia/abuse/tests/test_cinder.py
```diff
@@ -1214,23 +1220,55 @@ def test_appeal_anonymous(self):
...
+            CinderUser(user_factory()), cinder_entity_instance=self.CinderClass(addon)
+        )
         assert (
             addon.current_version.needshumanreview_set.get().reason
             == NeedsHumanReview.REASONS.ADDON_REVIEW_APPEAL
         )
         assert addon.current_version.reload().due_date
 
-    def test_appeal_specific_version(self):
+    def test_appeal_specific_version_from_report(self):
```
KevinMind, line 1232: Can you add docstrings for these tests?

Regarding `test_appeal_specific_version_from_report` you mentioned a report.. I think that maps to the CinderClass instance you've defined that (specified a version) right? And therefore we expect the original version (current_version) not to be flagged, but only the one referenced in the CinderClass.

Regarding `test_appeal_specific_version_from_action` I'm a bit lost on what exactly this tests...

src/olympia/abuse/tests/test_cinder.py
```diff
@@ -149,8 +154,8 @@ def _test_appeal(self, appealer, cinder_instance=None):
             status=201,
         )
         assert (
-            cinder_instance.appeal(
-                decision_cinder_id=fake_decision_id,
+            cinder_entity_instance.appeal(
```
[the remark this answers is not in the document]
eviljeff: just that when I re-read the test I was initially confused that `cinder_instance` was a instance of the CinderEntity subclass, not a CinderJob.

src/olympia/abuse/tests/test_cinder.py
```diff
@@ -136,9 +136,14 @@ def test_build_report_payload(self):
...
 
-    def _test_appeal(self, appealer, cinder_instance=None):
-        fake_decision_id = 'decision-id-to-appeal-666'
-        cinder_instance = cinder_instance or self.CinderClass(
+    def _test_appeal(
+        self,
+        appealer,
+        *,
+        cinder_entity_instance=None,
+        appealed_decision_id='decision-id-to-appeal-666',
```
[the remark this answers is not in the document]
eviljeff: what method signature are you referring to?  `_test_appeal` is only defined once (here)

src/olympia/abuse/tests/test_cinder.py
eviljeff: I've attempted to do this in https://github.com/mozilla/addons-server/pull/22975/commits/8486ff498480e1df40103df854be42ef02a1eb6a - though the docstrings are pretty much what the comments at the top of the tests already said... let me know if this helps...

src/olympia/abuse/tests/test_cinder.py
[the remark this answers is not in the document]
KevinMind: The method for `.appeal` on the cinder entity.

src/olympia/abuse/tests/test_cinder.py
```diff
@@ -1214,23 +1220,60 @@ def test_appeal_anonymous(self):
...
+        assert not addon.current_version.needshumanreview_set.exists()
+        assert (
+            other_version.needshumanreview_set.get().reason
+            == NeedsHumanReview.REASONS.ADDON_REVIEW_APPEAL
+        )
+        assert not addon.current_version.reload().due_date
+        assert other_version.reload().due_date
+
+    def test_appeal_specific_version_from_action(self):
+        """For an appeal from a developer, version_string will be None on the
```
KevinMind, line 1258: You say "version_string will be None on...." and then you say "If version_string is falsely"... is this a different reference with the same name or otherwise if it's always none how is that a valid condition?

Also `false**l**y` I think drop the "l"

src/olympia/abuse/tests/test_cinder.py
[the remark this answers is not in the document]
eviljeff: Cinder (the webservice) would return an error if `decision_to_appeal_id` was not a valid uuid for a decision.  But that wouldn't be reflected in the tests here because we're mocking all the requests/responses to cinder - `None` or any other value is as valid or invalid as we decide (and we decide it's invalid and return a 400 error instead further down in this test).
`None` wouldn't be a typical value though, so using it in the test and having it "pass" seems wrong to me.

src/olympia/abuse/tests/test_cinder.py
eviljeff: The codepath will result in `version_string` being `None` for a developer appeal, because `abuse_report` will be `None`, so `getattr(abuse_report, 'version_string', None)` will be `None`.  But the test is `not version_string` to cover the case where there addon_version was specified as `''` in the report, but there _was_ a version associated with the decision (I don't know how likely a `''` value is to happen, but we account for it in other places).

src/olympia/abuse/tests/test_cinder.py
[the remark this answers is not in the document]
KevinMind: That is exactly my point though, the test should fail by default and so the default value for the _test_* methods should, if at all, use values which are likely to fail. The actual test calling this helper method can then pass the value that makes it pass.

I've seen over and over tests like this pass when they maybe shouldn't due to default arguments that are designed to pass. So in general, I have a reflex against that pattern. I would change it to None or provide no default at all (since the method we pass that to also doesn't define a default) and let the individual tests decide what the value should be.

src/olympia/abuse/tests/test_cinder.py
KevinMind: Okay, can you specify in the docstring _which_ version string is potentially "falsely" so it is more clear (since there are apparently two).

> If version_string is falsely we collect and flag the addon
        versions from the appealled decision rather the current_version.

Would a more descriptive version of this sentence (from your added context above) be

> If there is a version associated with the cinder decision, then we collect and flag that version for review even if the "addon_version" associated with the parent report is falsy (e.g. "")  

I think I conceptually understand what you mean for this test, but the test code itself seems to be hiding things that make it not clear how the test implements that scenario. I don't see a reference to a report and there is no place where you are setting an "addon_version" and then how is it set to an empty string "" 🤷 This is magic to me.

I know you understand how it works, but I think that is precisely why documenting the test is important. You or someone will look at this in even a few weeks and be confused.

src/olympia/abuse/tests/test_cinder.py
```diff
@@ -1214,23 +1220,60 @@ def test_appeal_anonymous(self):
...
         addon = self._create_dummy_target()
         other_version = version_factory(
             addon=addon,
             channel=amo.CHANNEL_UNLISTED,
             file_kw={'status': amo.STATUS_AWAITING_REVIEW},
         )
+        decision = ContentDecision.objects.create(
+            action=DECISION_ACTIONS.AMO_DISABLE_ADDON, cinder_id='some_id', addon=addon
+        )
+        ActivityLog.objects.create(
```
KevinMind, line 1270: Add a comment here (this is the secret sauce)

> # An activity log links the other_version to the decision, even though the version_string on the CinderClass below is set to None

src/olympia/abuse/tests/test_cinder.py
```diff
@@ -1214,23 +1220,60 @@ def test_appeal_anonymous(self):
...
+        )
+        assert not addon.current_version.reload().due_date
+        assert other_version.reload().due_date
+
+    def test_appeal_specific_version_from_action(self):
+        """For an appeal from a developer, version_string will be None on the
+        CinderClass instance. If version_string is falsely we collect and flag the addon
+        versions from the appealled decision rather the current_version."""
         addon = self._create_dummy_target()
         other_version = version_factory(
```
KevinMind, line 1262: nit: other_version -> flagged_version

KevinMind, line 1262: nit: other_version -> flagged_version

src/olympia/abuse/tests/test_cinder.py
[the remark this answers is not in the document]
eviljeff: That's _my_ point though 😬 - the test would _pass_ with `None` because there's nothing in `CinderEntity.appeal` that requires `decision_to_appeal_id` to be a non-`None` value - it just passes it along in the payload.  

And not explicitly in `CinderAddonHandledByReviewers.appeal` either - `None` would mean there wouldn't be any versions identified, but that's the exact same scenario as the default being `'decision-id-to-appeal-666'` and there not being a log entry associated with a decision with that id.

So all we'd have is a test that uses `None` as the uuid and passes, even though in the real world that wouldn't ever be a valid value, because the request to Cinder would fail.  Is that better?

src/olympia/abuse/tests/test_cinder.py
eviljeff: > Okay, can you specify in the docstring which version string is potentially "falsely" so it is more clear (since there are apparently two).

There is one.  `CinderAddonHandledByReviewers.version_string` is set from the `AbuseReport`.

> Would a more descriptive version of this sentence (from your added context above) be
>
>>    If there is a version associated with the cinder decision, then we collect and flag that version for review even if the "addon_version" associated with the parent report is falsy (e.g. "")

No, that's the opposite logic to the python code in `CinderAddonHandledByReviewers.appeal` - the if is checking `addon_version` first, _then_ falling back to versions associated with the decision.

> I think I conceptually understand what you mean for this test, but the test code itself seems to be hiding things that make it not clear how the test implements that scenario. I don't see a reference to a report 

The codepath starts with an abuse report in this scenario, but there is no report referenced in `appeal`, just the id for the decision being appealed, and the version string from the abuse report.

> and there is no place where you are setting an "addon_version" and then how is it set to an empty string "" 🤷 This is magic to me.

`self.addon_version` is set in the `__init__` for `CinderAddonHandledByReviewers` - you can trace back to the one place where `CinderAddonHandledByReviewers` is created (`CinderJob.get_entity_helper`); the `get_entity_helper` call in `ContentDecision.appeal` (the one place that calls `CinderEntity.appeal`) is made with the abuse report.  What's in the abuse report comes from api POST submissions.

> I know you understand how it works, but I think that is precisely why documenting the test is important. You or someone will look at this in even a few weeks and be confused.

I agree - I'm just not sure what context needs to be added, exactly.

KevinMind: I think this comment does a lot of heavy lifting https://github.com/mozilla/addons-server/pull/22975/files/8486ff498480e1df40103df854be42ef02a1eb6a#r1908428369 Could resolve this adding that.

KevinMind (approved):
The code and tests look good. I wasn't able to verify locally but that seems to be an unrelated issue. Happy to let further verification happen on dev.

PR #22977, opened 2025-01-09
Serve django locale bundles from django as a fallback in local dev

KevinMind:
Fixes: mozilla/addons#15272

### Context

We typically serve these files directly from nginx, but in dev mode you may or may not have them defined, so we should serve them from django as a fallback

### Testing

1. create an empty site static dir `mkdir -p ./site-static` (make sure it is empty)
2. run `make up DOCKER_TARGET=development`
3. verify there are no files in nginx `docker compose exec nginx ls /srv/site-static/js/i18n`
4. load dev hub `http://olympia.test/developers` (open the network tab)
5. Verify the `/static/js/i18n/en.js` file loads 200 with the javascript.

### Checklist

### DIFF
--- a/src/olympia/addons/tests/test_views.py
+++ b/src/olympia/addons/tests/test_views.py
@@ -7954,3 +7954,17 @@ class TestBrowserMapping(TestCase):
             'extension_id': extension_id_2,
             'addon_guid': addon_2.guid,
         }
+
+
+class TestJavascriptCatalog(TestCase):
+    @pytest.mark.needs_locales_compilation
+    def test_get_correct_catalog(self):
+        url = reverse('javascript-catalog', kwargs={'locale': 'fr'})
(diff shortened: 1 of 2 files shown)


diox (changes requested):
Can we have a test that has the `needs_locales_compilation` marker? Locally, after compiling locales with `make compile_locales`, even though I have the required `djangojs.mo` files, http://olympia.test/static/js/i18n/fr.js doesn't return the correct translations.

KevinMind:
> Can we have a test that has the `needs_locales_compilation` marker? Locally, after compiling locales with `make compile_locales`, even though I have the required `djangojs.mo` files, http://olympia.test/static/js/i18n/fr.js doesn't return the correct translations.

Done 👍

diox (approved):

PR #22982, opened 2025-01-10
Add new locales folders for the various es variants

diox:
See mozilla/addons#15232

Not using them yet, just adding the files so that they can be taken care of in pontoon.


src/olympia/addons/migrations/0054_update_default_locale_es_to_es-es.py
```diff
@@ -0,0 +1,23 @@
...
+from django.db import migrations
+
+def forwards_func(apps, schema_editor):
+    Addon = apps.get_model('addons', 'Addon')
+    Addon.unfiltered.filter(default_locale='es').update(default_locale='es-es')
+
+
+def backwards_func(apps, schema_editor):
+    Addon = apps.get_model('addons', 'Addon')
+    Addon.unfiltered.filter(default_locale='es-es').update(default_locale='es')
```
KevinMind, line 12: Is it possible currently for an `Addon` to have `es-es` as the default_locale? Could this run backwards blow that state away?

src/olympia/addons/tests/test_models.py
```diff
@@ -1067,7 +1067,7 @@ def test_newlines_inline(self):
         assert self.newlines_helper(before) == after
 
     def test_newlines_code_inline(self):
-        before = "Code tags aren't blocks.\n\n" '<code>alert(test);</code>\n\nSee?'
+        before = "Code tags aren't blocks.\n\n<code>alert(test);</code>\n\nSee?"
```
KevinMind, line 1070: Why does this test need changing?

src/olympia/amo/tests/test_amo_utils.py
```diff
@@ -178,7 +178,7 @@ def test_to_language(test_input, expected):
         ('en_US', 'en-US'),
         ('en', 'en-US'),
         ('FR', 'fr'),
-        ('es-ES', None),  # We don't go from specific to generic.
+        ('fr-FR', None),  # We don't go from specific to generic.
```
KevinMind, line 181: Such a large PR, is there a reason to go from spanish to french here?

src/olympia/amo/tests/test_utils.py
```diff
@@ -161,7 +161,7 @@ def test_multiple_objects_with_multiple_translations(self):
         assert set(addon2.translations[addon2.name_id]) == (
             {
                 ('en-us', 'English 2 Name'),
```
KevinMind, line 163: Oh god how I wish we had enums for locales...

src/olympia/api/fields.py
```diff
@@ -279,6 +279,7 @@ def fetch_single_translation(self, obj, field, requested_language):
         translations = self.fetch_all_translations(obj, field) or {}
         locale = None
         value = None
+        requested_language = to_language(requested_language)
```
KevinMind, line 282: It's not clear why this change is here.. could you add some context?

src/olympia/translations/management/commands/process_translations.py
```diff
@@ -1,28 +1,17 @@
...
-            'OLD_OUTGOING_URL', 'https://outgoing.prod.mozaws.net/v1/'
-        )
         return {
-            'update_outgoing_url': {
-                'task': update_outgoing_url,
-                'queryset_filters': [
-                    Q(localized_string_clean__icontains=old_outgoing_url),
-                ],
-                'kwargs': {'old_outgoing_url': old_outgoing_url},
+            'copy_spanish_translations': {
```
KevinMind, line 13: This change is also a bit opaque..

locale/es_AR/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,5440 @@
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: messages\n"
```
KevinMind, line 4: Is the rationale that it is easier for translators to modify the original spanish translation to variants than to start from scratch?

Also.. it's weird.. it's like did we support the different spanish variants already or is this introducing them. On the one hand, you're adding the locale files ... so they weren't there.. but onb the other, there was already reference to those locales ... can you explain that?

src/olympia/addons/tests/test_models.py
diox: Just my `ruff` formatting in my editor fixing this string to be a single string

src/olympia/amo/tests/test_amo_utils.py
diox: Yes, because this test checks that `xx-XX` when we only have a `xx` locale returns `xx`. We are introducing `es-ES` (and removing `es`) so I had to change it to something else.

src/olympia/translations/management/commands/process_translations.py
diox: Err I should remove this. I removed the old `update_outgoing_url` task because I had planned to add a management command to do the migration but decided against it eventually. I'll revert this bit, it doesn't matter either way.

locale/es_AR/LC_MESSAGES/django.po
diox: There were no references to the variants before, I'm adding them. Translator teams have expressed different preferences regarding what to do with the files, but the conclusion we reached when talking to @mathjazz (Pontoon lead) was to "seed" the new locales by copying the old file over and they'd sort them out in pontoon later.

src/olympia/api/fields.py
diox: This was a bug in the way we fetch translations coming from ES, it did not convert the requested language like we do when the translations are coming from the database. 

It was revealed by the tests when I changed them from `es` to `es-ES` - it would only occur when we're fetching a language with a "regional" variant (xx-XX) that is not the default locale for the add-on.

src/olympia/addons/migrations/0054_update_default_locale_es_to_es-es.py
diox: It's only possible if it was forced or is very, very old. As it turns out we have a single add-on in prod that does this, and it's... the `es-ES` language pack for Firefox.

In the unlikely event that we'd run the backwards migration for `addons` and not the one for `translations` in prod, that add-on would have no name/summary and would need to be manually fixed, which isn't the end of the world since it's really only for its `name` and `summary`. I could also have omitted the backwards migration or make it a no-op, I figured it was better to have a real one for local development.

src/olympia/translations/management/commands/process_translations.py
diox: 3724bb2ee40dcf49c4d5f55afd39c1f9d35fbf61

src/olympia/addons/migrations/0054_update_default_locale_es_to_es-es.py
KevinMind: I think the backwards compat is good for the migration, just wanted to understand the scope of the impact. I agree it's probably okay. Thanks for clarifying.

locale/es_AR/LC_MESSAGES/django.po
KevinMind: Got it. Probably worth dumping that in the "context" on the PR for posterity.

FYI there was a [reference](https://github.com/mozilla/addons-server/blob/3f154e06a92e9a2c0d85516eea3ef26f2c2e684e/src/olympia/core/languages.py#L149) to the language but I guess not sufficient enough to trigger creating a translation file for that locale.

KevinMind:
@diox how did you create 700k+ rows? Even better, you can zip your database and send it to me by running 

```bash
make data_dump ARGS="--name pr-22982"
```

Then zip the directory and send it to me. I can then load that directly on my instance and boom!

diox:
I'll generate the dump, but for the record, I essentially did:
```python
translations = []
for x in range(0, 700000):
    t = Translation.new('Lorem ipsum dolor sit amet, erat graece accusata eum te', 'es')
    translations.append(t)
Translation.objects.bulk_create(translations, batch_size=1000)
```

(`Translation.new()` is a custom hack in addons-server used behind the scenes when creating translations "the normal way", that is there to maintain the separate id sequence table. I've tried to reproduce that behavior here while still being efficient with a `bulk_create()` after, but I haven't checked if it completely works as expected in very aspect, I was only interested in creating garbage data)

src/olympia/api/fields.py
diox: I dug a bit more into what was happening here:

The broken test was `TestESTranslationSerializerField.test_attach_translations_target_name()`, which is meant to verify how we attach the translation dict of data coming from Elasticsearch to an object using `ESTranslationSerializerField.attach_translations()`. The test attaches the translations then verifies that the dummy Addon object
has both a correct `<field>_translations` and a `<field>` containing the
translation in the current locale.

The `<field>_translations` dict was correct, only that second part was broken. The reason we never noticed is that it's not actively used in our `ESAddonSerializer` right now: when we perform a search and call the serializer, we call attach_translations() but only really use the first part of what it does, setting `<field>_translations`. Even when a `lang` parameter is passed and we return a single translation, we still use only that `<field>_translations` dict.

Why does that second part exist then ? The way our `ESAddonSerializer` works, it creates fake `Addon` (and `Version`, `Preview`, `License` etc instances) but then shares most of its code with the `AddonSerializer` that deals with database instances. Setting the translated fields correctly ensures we don't accidentally trigger a database query or return bad data if some piece of code somewhere ends up directly using the field.

Ultimately the test was previously incorrect as it was setting up bad data - the translations dict keys were incorrect. I'll add some comments and make the test more explicit.

diox: 49ce11d105410e5feb74ba194482847134884ec6 adds even more tests/docs.

locale/es_AR/LC_MESSAGES/django.po
diox: Done

diox:
Updated, and on top of providing the snippet I used to generate extra data locally I've shared my database dump privately on slack.

src/olympia/api/fields.py
KevinMind: Very nice 👏 👏

KevinMind (approved):
Verified 🚢 

- migration took around 10 seconds with 700k translations.. creating the translations took a lot longer 😆 
- confirmed generic ES is not available on devhub selector

image

Issue #15261, opened 2020-08-11
Version with expired delay rejection is not automatically rejected if another version is awaiting review


AlexandraMoga:
### Describe the problem and steps to reproduce it:
1. Open the listed review page of an add-on with a reviewer account
2. Submit a delay rejection for the latest Approved add-on version - consider this as version `1.0`
3. Upload a new version for this add-on
4. Wait for the delay rejection deadline for the first version to expire and check the add-on review page

### What happened?
The first version has not been automatically rejected when the delay rejection period expired - its status is "Approved · Scheduled for rejection in 0 minutes". It looks like this happens when a new version is still Awaiting review.

### What did you expect to happen?
The version should have been rejected automatically when the grace period ended

### Anything else we should know?
Make sure you are not locking the add-on review by keeping the page open as this will prevent the automatic rejection cron to run on your version.
Add-on examples:
https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/delay-reje-approved-version
https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/cookie-autodelete23

wagnerand:
@AlexandraMoga  I thought the previous version was supposed be disabled immediately?

AlexandraMoga:
@wagnerand I believe this was also mentioned in the requirements doc, but it wasn't implemented (not yet at least).

wagnerand:
It seems we hold off on the delayed rejection until the next version has been reviewed.

Log output:
```
Skipping rejections for add-on 619699 until version 1691835 has been reviewed
Skipping rejections for add-on 620086 until version 1691814 has been reviewed
```

eviljeff:
> 
> 
> It seems we hold off on the delayed rejection until the next version has been reviewed.
> 
> Log output:
> 
> ```
> Skipping rejections for add-on 619699 until version 1691835 has been reviewed
> Skipping rejections for add-on 620086 until version 1691814 has been reviewed
> ```

so, it's by design?  (I guess so we don't reject - and take down the public version - until the next version is reviewed)

wagnerand:
@jvillalobos Can you clarify what the expected flow is for this scenario?

jvillalobos:
It sounds like it's working as expected. The old version can't be auto-disabled because the new version is still awaiting review. The reason it works this way is that the developer is presumably cooperating, so they shouldn't have their version/listing taken down until there's a resolution of the latest version.

[...]

PR #22983, opened 2025-01-13
Add MLBF validation logic and corresponding management command

KevinMind:
Fixes: mozilla/addons#15281

### Description

- Implemented a `validate` method in the `MLBF` class to check for duplicate items in the cache.json, ensuring each item occurs only once and in a single data type.
- Added a new management command `validate_mlbf` to facilitate validation of MLBF directories from the command line.
- Created unit tests to verify the validation logic, ensuring it raises errors for duplicate items within a single data type and across multiple data types.

### Context

We have discovered that MLBFs created since November 2024 are larger than expected. This validation is one step to ensure we do not see unchecked growth in the mlbf by verifying the underlying data set used to produce the filter.

### Testing

- Download the cache.json from the (privately) linked g-drive URL
- move the `mlbf` directory into your "storage" directory in the repository
- Run the below command to validate the data

```bash
./manage.py validate_mlbf.py <id>
```

Where `id` is the timestamp directory name `./mlbf/<id>/`.

### Checklist

### DIFF
--- a/src/olympia/blocklist/management/commands/validate_mlbf.py
+++ b/src/olympia/blocklist/management/commands/validate_mlbf.py
@@ -12,9 +12,19 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         """Handle command arguments."""
-        parser.add_argument('storage_id', help='The storage ID of the MLBF', metavar=('ID'))
+        parser.add_argument(
+            'storage_id', help='The storage ID of the MLBF', metavar=('ID')
+        )
+        parser.add_argument(
+            '--fail-fast',
+            action='store_true',
+            help='Fail fast if an error is found',
+            default=False,
+        )
 
     def handle(self, *args, **options):
         storage_id = options['storage_id']
-        mlbf = MLBF.load_from_storage(storage_id)
-        mlbf.validate()
+        fail_fast = options['fail_fast']
+        log.info(f'Validating MLBF {storage_id} with fail_fast={fail_fast}')
+        mlbf = MLBF.load_from_storage(storage_id, error_on_missing=True)
+        mlbf.validate(fail_fast=fail_fast)
--- a/src/olympia/blocklist/management/commands/validate_mlbf.py
+++ b/src/olympia/blocklist/management/commands/validate_mlbf.py
@@ -1,0 +1,28 @@
+import json
+import os
+
+from django.core.management.base import BaseCommand
+from django.conf import settings
+
+import olympia.core.logger
+from olympia.blocklist.mlbf import MLBF
+
+
+log = olympia.core.logger.getLogger('z.amo.blocklist')
+
+
+class Command(BaseCommand):
+    help = 'Validates a given MLBF directory'
+
+    def add_arguments(self, parser):
+        """Handle command arguments."""
+        parser.add_argument('storage_id', help='The storage ID of the MLBF', metavar=('ID'))
+
+    def load_json(self, json_path):
+        with open(json_path, 'r') as json_file:
+            return json.load(json_file)
+
+    def handle(self, *args, **options):
+        storage_id = options['storage_id']
+        mlbf = MLBF.load_from_storage(storage_id)
+        mlbf.validate()
--- a/src/olympia/blocklist/management/commands/validate_mlbf.py
+++ b/src/olympia/blocklist/management/commands/validate_mlbf.py
@@ -8,7 +8,7 @@ log = olympia.core.logger.getLogger('z.amo.blocklist')
 
 
 class Command(BaseCommand):
-    help = 'Validates a given MLBF directory'
+    help = 'Validates a given MLBF does not contain duplicate items'
 
     def add_arguments(self, parser):
         """Handle command arguments."""
--- a/src/olympia/blocklist/management/commands/validate_mlbf.py
+++ b/src/olympia/blocklist/management/commands/validate_mlbf.py
@@ -1,8 +1,4 @@
-import json
-import os
-
 from django.core.management.base import BaseCommand
-from django.conf import settings
 
 import olympia.core.logger
 from olympia.blocklist.mlbf import MLBF
@@ -17,10 +13,6 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         """Handle command arguments."""
         parser.add_argument('storage_id', help='The storage ID of the MLBF', metavar=('ID'))
-
-    def load_json(self, json_path):
-        with open(json_path, 'r') as json_file:
-            return json.load(json_file)
 
     def handle(self, *args, **options):
         storage_id = options['storage_id']
--- a/src/olympia/blocklist/management/commands/validate_mlbf.py
+++ b/src/olympia/blocklist/management/commands/validate_mlbf.py
@@ -15,16 +15,9 @@ class Command(BaseCommand):
         parser.add_argument(
             'storage_id', help='The storage ID of the MLBF', metavar=('ID')
         )
-        parser.add_argument(
-            '--fail-fast',
-            action='store_true',
-            help='Fail fast if an error is found',
-            default=False,
-        )
 
     def handle(self, *args, **options):
         storage_id = options['storage_id']
-        fail_fast = options['fail_fast']
-        log.info(f'Validating MLBF {storage_id} with fail_fast={fail_fast}')
+        log.info(f'Validating MLBF {storage_id}')
         mlbf = MLBF.load_from_storage(storage_id, error_on_missing=True)
-        mlbf.validate(fail_fast=fail_fast)
+        mlbf.validate()
--- a/src/olympia/blocklist/mlbf.py
+++ b/src/olympia/blocklist/mlbf.py
@@ -384,15 +384,8 @@ class MLBF:
             > 0
         )
 
-    def validate(self, fail_fast: bool = False):
-        errors = []
+    def validate(self):
         store: Dict[str, Dict[MLBFDataType, int]] = {}
-
-        def add_error(error: str):
-            if fail_fast:
-                raise ValueError(error)
-            log.error(error)
-            errors.append(error)
 
         # Create a map of each guid:version string in the cache.json
         # and the set of data types that contain it and the count of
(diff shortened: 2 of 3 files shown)


Rob--W:
> Now to find out how that happened in the first place and how to prevent.
> 
> FYI @Rob--W you're theory seems to be correct.

Note: although `cache.json` here does indeed have duplicates (which I had observed myself as well), it is not necessarily the root cause for the issues observed in #15261.

In my local testing that lead to https://github.com/mozilla/addons/issues/15261#issuecomment-2584947155, I generated a new MLBF solely based on the `cache.json` file, without de-duplication. This means that if there were any duplicates in `cache.json`, that it would directly contribute to the generated MLBF. The fact that my MLBF is still smaller than what was observed in production is a strong signal that there is additional duplicate data that is not stored in `cache.json`.

KevinMind:
>  I generated a new MLBF solely based on the cache.json file, without de-duplication. This means that if there were any duplicates in cache.json, that it would directly contribute to the generated MLBF. The fact that my MLBF is still smaller than what was observed in production is a strong signal that there is additional duplicate data that is not stored in cache.json.

I'm not sure I follow this logic.

> I generated a new MLBF solely based on the cache.json file, without de-duplication.

What do you mean by that? like what code did you actually run?

> The fact that my MLBF is still smaller than what was observed in production is a strong signal that there is additional duplicate data that is not stored in cache.json.

That makes sense assuming the way you generated the filter is the same as the way production generated it. My understanding of that is based on the answer to the first question.

src/olympia/blocklist/mlbf.py
```diff
@@ -157,28 +157,42 @@ def _all_blocks(self):
...
         )
 
-    def _format_blocks(self, block_type: BlockType) -> List[str]:
-        return MLBF.hash_filter_inputs(
+    def _format_blocks(self, versions: List[Tuple[str, str]]) -> List[str]:
+        unique_versions = set()
+        deduped_versions = []
+
+        for version in versions:
+            if version not in unique_versions:
```
Rob--W, line 165: Question: is there a benefit to looping over the input and checking it for each entry, over using `list(set(versions))` ?

src/olympia/blocklist/mlbf.py
```diff
@@ -77,7 +77,7 @@ def generate_mlbf(stats, include, exclude):
 # Extends the BlockType enum to include versions that have no block of any type
 MLBFDataType = Enum(
     'MLBFDataType',
-    [block_type.name for block_type in BlockType] + ['NOT_BLOCKED', 'not_blocked'],
```
Rob--W, line 80: This is the primary fix to the cause of https://github.com/mozilla/addons/issues/15261#issuecomment-2584947155

Rob--W (commented):
Thanks for putting this fix up so quickly!

The fix of the primary issue looks good to me.

I'll defer to AMO engineers for the approving sign-off, in case they have thoughts on the tests  or implementation details.

src/olympia/blocklist/mlbf.py
KevinMind: Converting to set and back to list breaks order in Python. And it is not faster to do so. So there is actually no real benefit other than code prettiness.

src/olympia/blocklist/tests/test_mlbf.py
```diff
@@ -971,6 +1031,39 @@ def test_generate_filter_returns_expected_data(self):
...
+            version_kw={'version': version},
+            file_kw={'is_signed': True},
+        )
+        addon = addon_factory(
+            version_kw={'version': version},
+            file_kw={'is_signed': True},
+        )
+
+        reused_addon.update(guid=GUID_REUSE_FORMAT.format(addon.id))
+        reused_addon.addonguid.update(guid=addon.guid)
```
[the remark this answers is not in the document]
KevinMind: I'm not sure how I would do that outside of what is already here. I'm explicitly setting the addonguid.guid of reused_addon to the guid of addon... neither are blocked so I cannot check for 2 non existent blocks.. so scratching my head on what exactly to verify?..

src/olympia/blocklist/mlbf.py
eviljeff: I've suggested this in other PRs too, but we could just keep everything as `set`s, and then `sorted(...)` before being written to the JSON file.  Python has efficient set handling (and the syntax is prettier too).

KevinMind: We have talked about this before and then same issue still is that sorting from sets seems to be unpredictable, as far as I've attempted it. I'm happy for someone to open an issue for that but I think changing that here would blow up the scope too much, especially since I'm not even confident it will work.

Also, still, there are zero performance benefits and actual performance costs associated with that approach so I'm not totally sure why we should prefer it in the first place.

src/olympia/blocklist/tests/test_mlbf.py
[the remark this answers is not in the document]
KevinMind: Done. we can verify the addon.addon_guid property are the same.

src/olympia/blocklist/mlbf.py
eviljeff: From previous issues I don't recall any concrete evidence of performance benefits or costs of using `set`s throughout these classes, just a language-agnostic ChatGPT analysis of a single function, that it wouldn't reduce the order of complexity of that function.
But without that evidence (e.g. timing) I agree there isn't justification to refactor in this PR.

willdurand: If that makes things easier, we could also drop the requirement to maintain ordered lists in the `cache.json` file.

Rob--W: > If that makes things easier, we could also drop the requirement to maintain ordered lists in the `cache.json` file.

For the MLBF to be reproducible, the cache.json file must either contain the entries in the order of generation (which is the case today), or the order should be deterministically derivable from the cache.json file.

src/olympia/blocklist/management/commands/validate_mlbf.py
```diff
@@ -0,0 +1,23 @@
...
+
+import olympia.core.logger
+from olympia.blocklist.mlbf import MLBF
+
+
+log = olympia.core.logger.getLogger('z.amo.blocklist')
+
+
+class Command(BaseCommand):
+    help = 'Validates a given MLBF directory'
```
eviljeff, line 11: nit: It'd be nice if the help text was a little bit descriptive

src/olympia/blocklist/mlbf.py
willdurand: That requirement had nothing to do with the reproducibility of MLBFs. It's nice and I didn't know about it, but if we can simplify code and avoid bugs like this one to happen again, then I'd lean towards simplicity.

src/olympia/blocklist/mlbf.py
```diff
@@ -384,6 +398,38 @@ def should_upload_stash(
             > 0
         )
 
+    def validate(self):
+        store: Dict[str, Dict[MLBFDataType, int]] = {}
```
eviljeff, line 402: nit: type annotation _within_ a function doesn't give us anything.

src/olympia/blocklist/mlbf.py
```diff
@@ -384,6 +398,38 @@ def should_upload_stash(
...
+        # how many times it occurs in each data type.
+        for key in MLBFDataType:
+            for item in self.data[key]:
+                if item in store:
+                    if key in store[item]:
+                        store[item][key] = store[item][key] + 1
+                    else:
+                        store[item][key] = 1
+                else:
+                    store[item] = {key: 1}
```
eviljeff, line 415: this reads like a use for `defaultdict`?
`store = defaultdict(lambda: defaultdict(int))` and then `store[item][key] = store[item][key] +1`

src/olympia/blocklist/mlbf.py
Rob--W: > That requirement had nothing to do with the reproducibility of MLBFs

It did. To identify the cause of the bug I had to reproduce the MLBF from its inputs, independently of AMO's code. An alternative way to achieve reproducibility in my verification tool is to "manually" sort the list after reading from cache.json.

As long as the steps are deterministic it does not really matter how cache.json looks like, and that is what I meant by "or the order should be deterministically derivable from the cache.json file.".

This thread here is rather long, and not critical in relation to the bug or operation. So it would be ideal if we don't block the PR on this discussion, and if there is any desire to make changes, to do that in a follow-up. The critical part of this patch is at line 80.

KevinMind: Regardless of any decision, I would not do that here or now. Way too much scope increase. I also don't really see the argument for removing a useful feature that is here and works... because we like sets?..

src/olympia/blocklist/mlbf.py
```diff
@@ -384,6 +398,38 @@ def should_upload_stash(
...
+
+        # Verify that each item occurs only one time and in only one data type
+        for item, data_types in store.items():
+            # We expect each item to only occur in one data type
+            if len(data_types) > 1:
+                formatted_data_types = ', '.join(key.name for key in data_types.keys())
+                raise ValueError(
+                    f'Item {item} found in multiple data types: '
+                    f'{formatted_data_types}'
```
eviljeff, line 424: nit
```suggestion
                    f'Item {item} found in multiple data types: {formatted_data_types}'
```

eviljeff (approved):
just some code nits.  I'll defer to Rob/Will's opinions for whether this is doing the "right" thing.

src/olympia/blocklist/mlbf.py
KevinMind: I think that fails linting for line to long but will try.

src/olympia/blocklist/mlbf.py
KevinMind: It helps me keep straight what the shape of the dictionary should be. I wish python would just let us use types properly but this is the closest I could get.

src/olympia/blocklist/management/commands/validate_mlbf.py
KevinMind: Better?

PR #22994, opened 2025-01-15
Bleach Add-on Summary of all HTML

chrstinalin:
Relates to mozilla/addons#15254 

### Description

Changes `NoURLsTranslation` to now escape all HTML, on top of stripping any links. Existing HTML is not escaped.

### Testing

Updating an add-on's summary with HTML will escape the HTML. 

### Checklist


src/olympia/bandwagon/models.py
```diff
@@ -61,8 +61,7 @@ class Collection(ModelBase):
     slug = models.CharField(max_length=30, blank=True, null=True)
 
     # description can (and sometimes does) contain html and other unsanitized
-    # content. It must be cleaned before display - NoURLsField just strips the
-    # URL without doing any escaping.
+    # content. It must be cleaned before display.
```
diox, line 64: We should reword that entire docstring now, as we are escaping, but at the same time it should be rendered as plain text by clients.

src/olympia/translations/models.py
```diff
@@ -267,29 +282,28 @@ class Meta:
...
     class Meta:
         proxy = True
 
     def __str__(self):
-        # Clean string if that hasn't been done already, like
-        # PurifiedTranslation does. Unlike PurifiedTranslation though, this
-        # class doesn't implement __html__(), because it's designed to contain
-        # only text. This means that it should be escaped by templates and API
-        # clients, as it can contain raw HTML.
+        # Clean string if that hasn't been done already.Unlike PurifiedTranslation,
```
diox, line 292:
```suggestion
        # Clean string if that hasn't been done already. Unlike PurifiedTranslation,
```

diox (approved):

PR #22998, opened 2025-01-18
Redirect incoming URLs with short languages to the long-language variant

diox:
Pre-requisite for https://github.com/mozilla/addons/issues/15232

### Context

That was previously only done for Accept-Language HTTP header in addons-server, not in URLs, despite being supported on addons-frontend.

### Testing

See unit test, you can try those URLs/redirects locally too.

### DIFF
--- a/src/olympia/amo/tests/test_url_prefix.py
+++ b/src/olympia/amo/tests/test_url_prefix.py
@@ -38,6 +38,11 @@ class MiddlewareTest(TestCase):
             '/es-PE/firefox/addon/1': '/es/firefox/addon/1',
             # /developers doesn't get an app.
             '/developers': '/en-US/developers',
+            # Shorter locales redirects to long form
+            '/pt/developers': '/pt-PT/developers',
+            '/pt/firefox': '/pt-PT/firefox/',
+            '/pt/firefox/addon/1': '/pt-PT/firefox/addon/1',
+            '/pt/addon/1': '/pt-PT/firefox/addon/1',
             # Check basic use-cases with a 'lang' GET parameter:
             '/?lang=fr': '/fr/firefox/',
             '/addon/1/?lang=fr': '/fr/firefox/addon/1/',
--- a/src/olympia/amo/urlresolvers.py
+++ b/src/olympia/amo/urlresolvers.py
@@ -36,8 +36,15 @@ class Prefixer:
         first_lower = first.lower()
         lang, dash, territory = first_lower.partition('-')
 
-        # Check language-territory first.
-        if first_lower in settings.LANGUAGE_URL_MAP:
+        # First test shorter languages shortcuts.
+        if not dash and first in settings.SHORTER_LANGUAGES:
+            first = settings.SHORTER_LANGUAGES[first]
+            if second in amo.APPS:
+                return first, second, rest
+            else:
+                return first, '', first_rest
+        # Then check language-territory.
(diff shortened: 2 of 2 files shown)


src/olympia/amo/urlresolvers.py
```diff
@@ -36,8 +36,15 @@ def split_path(path_):
...
-        if first_lower in settings.LANGUAGE_URL_MAP:
+        # First test shorter languages shortcuts.
+        if not dash and first in settings.SHORTER_LANGUAGES:
+            first = settings.SHORTER_LANGUAGES[first]
+            if second in amo.APPS:
+                return first, second, rest
+            else:
+                return first, '', first_rest
+        # Then check language-territory.
+        elif first_lower in settings.LANGUAGE_URL_MAP:
```
KevinMind, line 47: Nit: the elif could be returned to an if since the first if definitely returns right? I tend to prefer split if statements if they all return since elif makes me think we could continue executing after the final branch, but in this case there is a (not totally obvious) return

KevinMind: but I guess there are already other if/elif/elif branches so nvm 🤷

src/olympia/amo/urlresolvers.py
```diff
@@ -36,8 +36,15 @@ def split_path(path_):
         first_lower = first.lower()
         lang, dash, territory = first_lower.partition('-')
 
-        # Check language-territory first.
-        if first_lower in settings.LANGUAGE_URL_MAP:
+        # First test shorter languages shortcuts.
+        if not dash and first in settings.SHORTER_LANGUAGES:
+            first = settings.SHORTER_LANGUAGES[first]
```
KevinMind, line 41: Nit: the "first" renaming is blowing up my brain. I know what it's doing but it feels confusing. could you use a different name "first_short" for example?

src/olympia/amo/tests/test_url_prefix.py
```diff
@@ -38,6 +38,11 @@ def test_redirection(self):
             '/es-PE/firefox/addon/1': '/es/firefox/addon/1',
             # /developers doesn't get an app.
             '/developers': '/en-US/developers',
+            # Shorter locales redirects to long form
```
KevinMind, line 41: Curious how it would handle invalid urls like 

`/-pt/developers`

or 

`/pt--PT/developers`

or

`/-/developers`

Could you add some of those tests to verify sane handling of weird edge cases?

diox: addons-server handles all of these poorly (adding `/en-US/firefox/` prefix to all )... but adding a test would be confusing, because none of them match the regexp set in nginx to redirect traffic to addons-server, so they would all be handled by addons-frontend, which _does_ handle them somewhat correctly, redirecting to either `/en-US/developers/` or `/pt-PT/developers/`

Tbh we don't have to care at all - those are not supposed to work anyway.

src/olympia/amo/urlresolvers.py
diox: c25a40594bb1a88268eb63a7b7ba347c9cdd6e2b

src/olympia/amo/tests/test_url_prefix.py
KevinMind: Oh man so we have 3 different sources of redirection? Fun.

KevinMind (approved):
Looks good.

PR #23001, opened 2025-01-20
Make reviewer pending rejection input a datetime widget and allow changing it through an action

diox:
Fixes https://github.com/mozilla/addons/issues/14920

### Description

This allows reviewers to change the pending rejection date of versions pending rejection through a widget where they can select the date and time.

Because the actions menu is already pretty large, the clear pending rejection date action is reused, with an option to change or clear ; to tie everything together and make the UI more consistent, the rejection action use the same widgets, making the date choice absolute rather than relative.

#### New reject multiple versions UI
Screenshot 2025-01-03 at 17-17-44 Ui-Addon-Install – Add-ons for Firefox

#### Modified Clear pending rejection action (now Change pending rejection) UI
Screenshot 2025-01-03 at 17-18-37 Ui-Addon-Install – Add-ons for Firefox

## Testing

See tests. Non-exhaustive list of scenarios, after submitting an add-on and going to its review page:
- Use the reject multiple versions action with immediate effect
- Use the reject multiple versions action with a delay, and then...
  - Use the change pending rejection action to clear the delay
  - Use the change pending rejection action to change the delay to a date in the future

Changing the delay (not clearing it) should send an email to the developer following copy that can be found as a comment in the issue.

Changing a delay should require the new date to be at least one extra day in the future compared to now. Changing the delay of multiple versions at once is possible, but the existing pending rejection date should all be the same.

### DIFF
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -447,6 +447,14 @@ class ContentActionForwardToLegal(ContentAction):
         return self.log_action(amo.LOG.REQUEST_LEGAL)
 
 
+class ContentActionChangePendingRejectionDate(ContentAction):
+    description = 'Add-on pending rejection date has changed'
+    valid_targets = (Addon,)
+
+    def get_owners(self):
+        return self.target.authors.all()
+
+
 class ContentActionDeleteCollection(ContentAction):
     valid_targets = (Collection,)
     description = 'Collection has been deleted'
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -1,5 +1,5 @@
 import random
-from datetime import datetime, timedelta
+from datetime import datetime
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
@@ -383,10 +383,17 @@ class ContentActionRejectVersionDelayed(ContentActionRejectVersion):
 
     def __init__(self, decision):
         super().__init__(decision)
-        self.days = int(self.decision.metadata.get('delayed_rejection_days', 0))
+        self.delayed_rejection_date = datetime.fromisoformat(
+            self.decision.metadata.get('delayed_rejection_date')
+        )
 
     def log_action(self, activity_log_action, *extra_args, extra_details=None):
-        extra_details = {**(extra_details or {}), 'delayed_rejection_days': self.days}
+        extra_details = {
+            **(extra_details or {}),
+            'delayed_rejection_days': (
+                self.delayed_rejection_date - datetime.now()
+            ).days,
+        }
         return super().log_action(
             activity_log_action, *extra_args, extra_details=extra_details
         )
@@ -397,14 +404,13 @@ class ContentActionRejectVersionDelayed(ContentActionRejectVersion):
         if not self.decision.reviewer_user:
             # This action should only be used by reviewer tools, not cinder webhook
             raise NotImplementedError
-        pending_rejection_deadline = datetime.now() + timedelta(days=self.days)
 
         for version in self.decision.target_versions.all():
             # (Re)set pending_rejection.
             VersionReviewerFlags.objects.update_or_create(
                 version=version,
                 defaults={
-                    'pending_rejection': pending_rejection_deadline,
+                    'pending_rejection': self.delayed_rejection_date,
                     'pending_rejection_by': self.decision.reviewer_user,
                     'pending_content_rejection': self.content_review,
                 },
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -383,9 +383,15 @@ class ContentActionRejectVersionDelayed(ContentActionRejectVersion):
 
     def __init__(self, decision):
         super().__init__(decision)
-        self.delayed_rejection_date = datetime.fromisoformat(
-            self.decision.metadata.get('delayed_rejection_date')
-        )
+        if 'delayed_rejection_date' in self.decision.metadata:
+            self.delayed_rejection_date = datetime.fromisoformat(
+                self.decision.metadata.get('delayed_rejection_date')
+            )
+        else:
+            # Will fail later if we try to use it to log/process the action,
+            # but allows us to at least instantiate the class and use other
+            # methods.
+            self.delayed_rejection_date = None
 
     def log_action(self, activity_log_action, *extra_args, extra_details=None):
         extra_details = {
--- a/src/olympia/abuse/templates/abuse/emails/ContentActionChangePendingRejectionDate.txt
+++ b/src/olympia/abuse/templates/abuse/emails/ContentActionChangePendingRejectionDate.txt
@@ -1,0 +1,6 @@
+{% extends "abuse/emails/base.txt" %}{% block content %}
+
+As you are aware, your {{ type }} {{ name }} was manually reviewed by the Mozilla Add-ons team, at which point we found a violation of one or more Mozilla add-on policies.
+
+Our previous correspondence indicated that you would be required to correct the violation(s) by {{ details.old_deadline }}. However, after further assessing the circumstances - including the violation itself, the risks it presents, and the steps required to resolve it - we have determined that an alternative timeline is appropriate. Based on that determination, we have updated the deadline, and will now require you to correct your add-on violations no later than {{ details.new_deadline }}.
+{% endblock %}
--- a/src/olympia/abuse/tests/test_models.py
+++ b/src/olympia/abuse/tests/test_models.py
@@ -2835,7 +2835,11 @@ class TestContentDecision(TestCase):
             action=DECISION_ACTIONS.AMO_REJECT_VERSION_WARNING_ADDON,
             notes='some review text',
             reviewer_user=self.reviewer_user,
-            metadata={'delayed_rejection_days': 14},
+            metadata={
+                'delayed_rejection_date': (
+                    datetime.now() + timedelta(days=14, minutes=1)
+                ).isoformat()
+            },
         )
         decision.target_versions.set([addon.current_version])
         NeedsHumanReview.objects.create(
@@ -2870,7 +2874,11 @@ class TestContentDecision(TestCase):
             action=DECISION_ACTIONS.AMO_REJECT_VERSION_WARNING_ADDON,
             notes='some review text',
             reviewer_user=self.reviewer_user,
-            metadata={'delayed_rejection_days': 14},
+            metadata={
+                'delayed_rejection_date': (
+                    datetime.now() + timedelta(days=14, minutes=1)
+                ).isoformat()
+            },
         )
         decision.target_versions.set([version])
         assert decision.action_date is None
--- a/src/olympia/abuse/tests/test_models.py
+++ b/src/olympia/abuse/tests/test_models.py
@@ -1,6 +1,6 @@
 import json
 import uuid
-from datetime import datetime
+from datetime import datetime, timedelta
 from unittest import mock
 
 from django.conf import settings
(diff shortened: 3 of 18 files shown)


src/olympia/reviewers/views.py
```diff
@@ -1114,21 +1114,6 @@ def allow_resubmission(self, request, **kwargs):
             status_code = status.HTTP_409_CONFLICT
         return Response(status=status_code)
 
-    @drf_action(
-        detail=True,
-        methods=['post'],
-        permission_classes=[GroupPermission(amo.permissions.REVIEWS_ADMIN)],
-    )
-    def clear_pending_rejections(self, request, **kwargs):
```
diox, line 1122: Drive-by cleanup to reduce confusion - This API was already obsolete since 9f1e462ad03708cb7516b64aa08fcaf538028b0a

src/olympia/abuse/actions.py
```diff
@@ -383,10 +383,23 @@ class ContentActionRejectVersionDelayed(ContentActionRejectVersion):
 
     def __init__(self, decision):
         super().__init__(decision)
-        self.days = int(self.decision.metadata.get('delayed_rejection_days', 0))
+        if 'delayed_rejection_date' in self.decision.metadata:
```
KevinMind, line 386: Nit

```python
self.delayed_rejection_date = (
    self.decision.metadata.get('delayed_rejection_date')
    and datetime.fromisoformat(self.decision.metadata['delayed_rejection_date'])
)
```

src/olympia/abuse/actions.py
```diff
@@ -383,10 +383,23 @@ class ContentActionRejectVersionDelayed(ContentActionRejectVersion):
...
+            self.delayed_rejection_date = datetime.fromisoformat(
+                self.decision.metadata.get('delayed_rejection_date')
+            )
+        else:
```
KevinMind, line 390: Is the implicit assumption here that if the value actually is none, then later methods will raise? We should just raise explicitly here then no?

src/olympia/abuse/templates/abuse/emails/ContentActionChangePendingRejectionDate.txt
```diff
@@ -0,0 +1,6 @@
+{% extends "abuse/emails/base.txt" %}{% block content %}
+
+As you are aware, your {{ type }} {{ name }} was manually reviewed by the Mozilla Add-ons team, at which point we found a violation of one or more Mozilla add-on policies.
```
KevinMind, line 3: Just my read, but this sounds somewhat passive aggressive.. is the "As you are aware" necessary? I would think this type of email should be more matter of fact and not assume what people may or may not be aware of

src/olympia/abuse/tests/test_models.py
```diff
@@ -1170,11 +1170,10 @@ def test_process_decision(self):
         policy_a = CinderPolicy.objects.create(uuid='123-45', name='aaa', text='AAA')
         policy_b = CinderPolicy.objects.create(uuid='678-90', name='bbb', text='BBB')
 
-        with mock.patch.object(
-            ContentActionBanUser, 'process_action'
-        ) as action_mock, mock.patch.object(
-            ContentActionBanUser, 'notify_owners'
-        ) as notify_mock:
+        with (
```
KevinMind, line 1173: TIL 💪

src/olympia/reviewers/forms.py
```diff
@@ -497,8 +514,54 @@ def clean(self):
                 raise ValidationError(
                     'Multiple policies selected with different cinder actions.'
                 )
+
+        if self.helper.actions.get(selected_action, {}).get('delayable'):
+            if 'delayed_rejection' not in self.data:
```
KevinMind, line 519: It's not totally clear to me how this wouldn't be handled by setting the fields to required (which you do) or using required on the html?

src/olympia/reviewers/forms.py
```diff
@@ -497,8 +514,54 @@ def clean(self):
...
+                self.add_error(
+                    'delayed_rejection',
+                    forms.ValidationError('This field is required.'),
+                )
+            elif self.cleaned_data.get('delayed_rejection') and not self.data.get(
```
KevinMind, line 524: Nit: following your other refactorings you are querying the fields from cleaned_data repeatedly and could be extracted to vars..

src/olympia/reviewers/forms.py
```diff
@@ -497,8 +514,54 @@ def clean(self):
...
+                distinct_pending_rejection_dates = (
+                    self.cleaned_data['versions']
+                    .values_list('reviewerflags__pending_rejection')
+                    .distinct()
+                    .count()
+                )
+                if distinct_pending_rejection_dates > 1:
+                    self.add_error(
+                        'versions',
+                        forms.ValidationError(
```
KevinMind, line 548: Why is this a requirement? I could imagine having 3 vesrions with different dates and my goal is to set them all to the same single date??

src/olympia/reviewers/tests/test_forms.py
```diff
@@ -95,7 +95,7 @@ def test_actions_addon_status_null(self):
         actions = self.get_form().helper.get_actions()
         assert list(actions.keys()) == [
             'unreject_latest_version',
-            'clear_pending_rejection_multiple_versions',
+            'change_pending_rejection_multiple_versions',
```
KevinMind, line 98: nit: it is no longer clear in the name you can also "clear" via this action. Maybe "change_or_clear_pending_rejection_multiple_versions" is super long, but more clear? Or (for some reason) "set" "update" read a bit more broadly to me. Somehow from "change" i'm inferring that there is a value to change to.

KevinMind (commented):
Initial thoughts and feedback/ I'll try to setup and verify tomorrow.

src/olympia/abuse/actions.py
diox: My comment is probably poorly worded. There are times when we need to instantiate the class and call some methods that don't need the `delayed_rejection_date`, and we want those to continue to work.

An example is found in `TestCinderJob.test_report_with_outstanding_rejection` which sends an email to the reporter of an abuse report filed against an add-on with a delayed rejection already set (but that delayed rejection decision doesn't have that metadata - as might be the case for existing ones before this change)

src/olympia/abuse/templates/abuse/emails/ContentActionChangePendingRejectionDate.txt
diox: In my experience developers are very aware of rejections already :sweat_smile: - and I assume in most cases they are the ones asking for that date to be changed.

Regardless, this is the copy everyone approved (see [issue](https://github.com/mozilla/addons/issues/14920)) so I'm not in hurry to change it and go through validation with Operations, Product and Legal again.

src/olympia/reviewers/forms.py
diox: It's a requirement from the issue. I think the main problem is that we have to notify the developer of the old date in the email (see copy you commented on above), which makes it cumbersome if there are multiple different ones. Operations would know more, but @wagnerand noted the main use case would be when dates are all the same anyway so this limitation wouldn't be too bad.

src/olympia/reviewers/forms.py
diox: mmm looks like we already manipulate `required` in `is_valid()` above for various fields depending on the action so yeah, I'll try doing that there and drop the custom `ValidationError`

diox: Actually, can't do that because `NullBooleanField` accepts missing values as `None` without raising, even if the field is `required`. I added a comment and stopped hardcoding the error message in 407703a7f58465cf9f570e68eacfdc2534b664d6

KevinMind:
Some feedback after running through the steps

It looks like the change pending rejection is available as an action even if I don't have any pending rejections.. not sure if that is pre-existing but seems odd

Screenshot 2025-01-31 at 14 04 28
fc8-4387-91b3-5ea4e1f815b3" />

It is odd that I see an hr/mm at all, and it is actually html required. Can we remove that to only select a date and not time?

Screenshot 2025-01-31 at 13 27 15

Maybe this is to "react"-ive for what we can do with rev tools but I would expect if I modify the date field, I would exect on blur that we should change the radio to change date and not clear date. It would be super easy to think you are changing the date and actually you clear it. I actually did that before noticing the issue.

https://github.com/user-attachments/assets/e73b2bd3-b6fd-49c8-93a2-184ce850bdfe

diox:
> It looks like the change pending rejection is available as an action even if I don't have any pending rejections.. not sure if that is pre-existing but seems odd

That is pre-existing yeah, I didn't change that. We could, I'm not sure it's worth the extra database query (we're trying to minimize them on this page to keep it as fast as possible - every milliseconds counts given how many different things it's loading)

> When I schedule a rejection.. shouldn't I see when the pending rejection is for in the log?

You mean the date ? We don't show it in the log because it's not part of the format string, but we do show the current scheduled rejection date in the row of the version itself in the history if the pending rejection is still active.

> It is odd that I see an hr/mm at all, and it is actually html required. Can we remove that to only select a date and not time?

Maybe. I played with that a little bit, showing the time was a compromise: I was trying to avoid having all scheduled rejections end up at the same time, and not accidentally reduce the delay by a few hours for some developers depending on when the rejection was posted. My original idea was to only allow reviewers to select the date and automatically set the time to the hour/minute of the submission, and in fact that's shown in my screenshots in the description, but then I changed my mind for this implementation to keep things simpler. I don't feel strongly one way or the other though, but I figured, since we do have a full datetime internally, why not show it.

> Maybe this is to "react"-ive for what we can do with rev tools but I would expect if I modify the date field, I would exect on blur that we should change the radio to change date and not clear date. It would be super easy to think you are changing the date and actually you clear it. I actually did that before noticing the issue.

Also something I changed my mind on as I was working on this. My first implementation was more JavaScript-heavy and I went for something a bit simpler to maintain given the current codebase. But I see your point, maybe there is a middle-ground we can reach here. I'll see what I can do.

src/olympia/reviewers/forms.py
diox: a1128d64885be43823671b51a83f99ca5e212c7f

src/olympia/abuse/actions.py
KevinMind: Just curious if that could have been resolved with a migration? All good though no need for a change.

src/olympia/reviewers/forms.py
KevinMind: Ok

src/olympia/abuse/actions.py
diox: Maybe. We haven't tried to do that in the past, it's a bit of an edge case (we typically don't have to worry about executing past decisions...)

diox:
>> Maybe this is to "react"-ive for what we can do with rev tools but I would expect if I modify the date field, I would exect on blur that we should change the radio to change date and not clear date. It would be super easy to think you are changing the date and actually you clear it. I actually did that before noticing the issue.
>
> Also something I changed my mind on as I was working on this. My first implementation was more JavaScript-heavy and I went for something a bit simpler to maintain given the current codebase. But I see your point, maybe there is a middle-ground we can reach here. I'll see what I can do.

https://github.com/mozilla/addons-server/pull/23001/commits/f1f0ac5243df4fea90a6ad6a5449e683279dc5d1 does this

KevinMind (approved):
Verified 🚢

src/olympia/reviewers/tests/test_forms.py
diox: e1ffe0983d8264314fe1adbd8353815ef295e3ab

Kept the label the same to avoid wrapping, but updated (eh) the internal name, and the description to mention that changing the date will notify developers.

PR #23004, opened 2025-01-21
delegate review actions to cinder

eviljeff:
Fixes: mozilla/addons#1982

### Description

Refactors code in the ContentAction classes, and ReviewHelper to allow activity logging to be consistently generated;  migrates Disable Add-on action from the reviewer tools to ContentAction classes.

### Context

The primary motivation for this work (what this blocks) is 2nd level approvals for reviewer tools originated decisions - and only certain types of decision (add-on disables and addon version rejections) will be held - so not all actions need to be rewritten at this point.  This PR starts with disable_addon as it's a) already implemented for cinder originated decisions and b) relatively straightforward.

I had to tinker with some existing migrations to fix breakage caused by migrations executing in a different order now (though they were only a problem in CI or with a fresh database anyway)

### Testing

The main change is how Disable Addon is executed from the reviewer tools:
- For a "normal" add-on, select force disable action from reviewer tools, and choose a reason.
- verify that the addon was disabled; and the email we send out contains the reason text
- check the activity log created is linked to the addon, current_version of the addon, decision, policy, and reason.

Repeat with an add-on that would be considered high profile - e.g. a Recommended or Notable add-on.
- select force disable from the reviewer tools, etc.
- verify that the add-on was _not_ disabled, and the review history includes that the action was held.  (no email to developer either)
- Go to 2nd level approval queue and release the hold - proceed with the action.
- verify that the addon was disabled; and the email we send out contains the reason text
- check the activity log created is linked to the addon, current_version of the addon, decision, policy, and reason.

### Checklist


src/olympia/reviewers/utils.py
```diff
@@ -1009,15 +1012,25 @@ def create_decision():
...
+            log_entry = self.log_action(
+                activity_action,
+                decisions=decisions,
+                reasons=reasons,
+                policies=policies,
+                **(log_action_kw or {}),
+            )
+            self.update_queue_history(log_entry)
         for decision in decisions:
+            log_entry = decision.execute_action()
```
diox, line 1026: This is confusing me at first glance. Shouldn't it be inside the `if not action_completed` condition below ?

eviljeff: `execute_action` also clears or sets pending rejection fks with the cinder job, plus clears cinder related NHRs.  That logic _could_ be moved into this class instead (reversing the trend)

diox: Got it. I think it's fine to have that in `execute_action()`, it's just that while we're in this weird in-between state with stuff being performed on both sides depending on the action it's kinda confusing which part does what

diox (approved):

PR #23011, opened 2025-01-22
Migrate build scripts to python with single script to orchestrate during make up

KevinMind:
Fixes: mozilla/addons#15066

Also fixes the below issue regarding

Fixes: mozilla/addons#15235

### Description

- Replaced the locale compilation script from a shell script to a Python script (compile_locales.py) for better error handling and parallel processing.
- Updated the update_assets target in Makefile-docker to use the new update_assets.py script.
- Removed the obsolete compile-mo.sh script.
- Introduced sync_host_files.py to streamline dependency updates and asset synchronization.

### Context

This changes makes sure that the state of the host file system matches the state of the container before running the containers.

This enables simplifying the volumes we use to a single volume from the host to the /data/olympia directory.

On the flip side, the docker build uses the exact same scripts to execute build logic during a build. That is how we guarantee files match in and out of the container.

### Testing

Run the container in production mode.

```bash
```

I would have suggested a content hashing test to verify the site-static directory from the built image and the one from make up are identical, but django `staticfiles.json` is non-deterministically created.. changes the order of the json attributes. I think now it is enough that literally the same command is run.

### Checklist


Dockerfile
```diff
@@ -125,73 +125,15 @@ ENV DOCKER_TARGET=${DOCKER_TARGET}
...
-# This will shave nearly 1 minute off the best case build time
-RUN \
-    --mount=type=bind,src=src,target=${HOME}/src \
-    --mount=type=bind,src=Makefile-docker,target=${HOME}/Makefile-docker \
-    --mount=type=bind,src=manage.py,target=${HOME}/manage.py \
-<<EOF
-echo "from olympia.lib.settings_base import *" > settings_local.py
-DJANGO_SETTINGS_MODULE="settings_local" make -f Makefile-docker update_assets
-EOF
+# Copy build info from info
```
KevinMind, line 130: Because we now don't use the production image for development.. and due to the dependency graph here there is really no reason to maintain the split between locales/assets/production

They all just rely on one single command. That also bolsters the link between dev/prod because make up just re-runs that command to ensure host files match image files.

KevinMind: image

diox (approved):

PR #23018, opened 2025-01-23
Group celery and google packages together in dependabot config

diox:
Fixes: mozilla/addons#15304


diox:
We could also look into limiting the number of minor/patch level updates we get for some packages, but this is just a first step to make dependabot PRs more manageable.

eviljeff (approved):

PR #23023, opened 2025-01-24
delegate reject actions to ContentActions too

eviljeff:
Fixes: mozilla/addons#15306

~Depends on https://github.com/mozilla/addons-server/pull/23004 & needs rebasing after that is merged.~

### Description

Implements the last remaining actions that could need a hold for 2nd level approval - version rejections.

### Context

To ease implementation, I've focused on the simple case - the initial rejection of one or more versions - so ignored auto rejections and rejections that happen via blocks (and renamed the existing `reject_multiple_versions` function).  Auto rejections don't need another hold because it was already considered when the initial decision was made; blocks have their own hold criteria.  Follow-up work would be to continue migrating the remaining code that changes addon/version states in the reviewer tools to their equivalent ContentAction 

### Testing

Similarly to #15306, the change is how the action ends up being processed in the reviewer tools, but this time for all the reject actions:

- For a "normal" add-on, select reject action from reviewer tools, and choose a reason.
- verify that the version was disabled; and the email we send out contains the reason text
- check the activity log created is linked to the addon, the version rejected, decision, policy, and reason.

Repeat with an add-on that would be considered high impact - e.g. a Recommended or Notable add-on.

- select reject from the reviewer tools, etc.
- verify that the version was not disabled, and the review history includes that the action was held. (no email to developer either)
- Go to 2nd level approval queue and release the hold - proceed with the action.
- verify that the version was disabled; and the email we send out contains the reason text
- check the activity log created is linked to the addon, version rejected, decision, policy, and reason.

This can be further repeated with multiple version rejection; delayed rejections; and content rejections - they should all work as before when the add-on isn't high impact, but be held for review if the add-on is.

### Checklist


src/olympia/abuse/actions.py
```diff
@@ -324,25 +327,106 @@ def get_owners(self):
 class ContentActionRejectVersion(ContentActionDisableAddon):
     description = 'Add-on version(s) have been rejected'
 
-    def should_hold_action(self):
-        # This action should only be used by reviewer tools, not cinder webhook
-        # eventually, if add-on becomes non-public do as disable
-        raise NotImplementedError
+    def __init__(self, decision):
+        super().__init__(decision)
+        self.content_review = decision.metadata.get('content_review', False)
```
eviljeff, line 332: fwiw, I did consider using separate Action classes ("ContentActionRejectContent" and "ContentActionRejectContentDelayed"), but it's the same actual outcome, with the emails sent in either case so didn't seem worth the extra complexity - especially as content reviews are being changed soon(?).  

At that point we may want a different ContentAction class - and perhaps set AMO_DISABLE_ADDON instead as the cinder action as it's more what the outcome is (the listing being removed) than a version rejection(?)

src/olympia/constants/activity.py
```diff
@@ -1119,6 +1119,50 @@ class REQUEST_LEGAL(_LOG):
     cinder_action = DECISION_ACTIONS.AMO_LEGAL_FORWARD
 
 
+class HELD_ACTION_REJECT_VERSIONS(_LOG):
+    id = 199
+    action_class = 'reject'
+    format = _('{addon} {version} rejection held for further review.')
```
diox, line 1125: It won't be shown to developers (it's an admin event), so it doesn't need to be translated. Same for the others below (and also `REQUEST_LEGAL` above if you want to do a drive-by)

src/olympia/reviewers/utils.py
```diff
@@ -1383,7 +1382,66 @@ def confirm_auto_approved(self):
...
+        # same channel should be manually reviewed.
+        auto_approval_disabled_until_next_approval_flag = (
+            'auto_approval_disabled_until_next_approval'
+            if channel == amo.CHANNEL_LISTED
+            else 'auto_approval_disabled_until_next_approval_unlisted'
+        )
+        AddonReviewerFlags.objects.update_or_create(
+            addon=self.addon,
+            defaults={auto_approval_disabled_until_next_approval_flag: True},
+        )
```
diox, line 1430: Should this be part of the `ContentAction` ? It feels like a consequence of the action, something that we wouldn't do if the action is "cancelled" after being held up for approval.

diox (approved):

src/olympia/reviewers/utils.py
eviljeff: My thinking was that in either case we'd want to stop any further auto approvals. It also prevents any approvals while the rejection is held.

diox: mmm yeah preventing any approvals while the rejection is held certainly makes sense.

PR #23024, opened 2025-01-27
process jobs from legal escalations with no abuse reports

eviljeff:
Fixes: mozilla/addons#15278

### Description

Handles jobs from Cinder that have no abuse report attached.  Also correctly clears NHR from the add-on when the legal escalation is made (the other problem noted in the issue)

### Context

CinderJobs in general don't have any target information directly fk'd to them - the information is on the AbuseReport, or appealed decision, that led to the job being created.  A legal forward doesn't (necessarily) have any appeals or abuse reports associated with it.  But for add-on, we do set `target_addon`, so we can use that instead (which we did already as an optimization, but then failed because there was no abuse report or appeal to reference `.rating`, etc , which would always have been `None` for an addon abuse report/appeal anyway).

### Testing

- Forward an add-on with some NHRs (manually create them in django shell if it's easier)
- see them cleared with the "review" (i.e. the add-on is no longer in the manual review queue)
- see the job in Cinder legal escalations queue
- resolve the job in Cinder with an action that would disable the add-on (AUP, etc)
- playback the webhook response locally
- see the add-on is disabled

### Checklist

### DIFF
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -312,9 +312,9 @@ class CinderJob(ModelBase):
                 if self.target_addon_id
                 else abuse_report_or_decision.addon
             ),
-            rating=abuse_report_or_decision.rating,
-            collection=abuse_report_or_decision.collection,
-            user=abuse_report_or_decision.user,
+            rating=getattr(abuse_report_or_decision, 'rating', None),
+            collection=getattr(abuse_report_or_decision, 'collection', None),
+            user=getattr(abuse_report_or_decision, 'user', None),
             cinder_id=decision_cinder_id,
             action=decision_action,
             notes=decision_notes[: ContentDecision._meta.get_field('notes').max_length],
--- a/src/olympia/abuse/tests/test_models.py
+++ b/src/olympia/abuse/tests/test_models.py
@@ -1252,6 +1252,35 @@ class TestCinderJob(TestCase):
         assert new_job.target_addon == addon
         assert report.reload().cinder_job == new_job
 
+    def test_process_decision_for_legal_reviewed_job(self):
+        """An add-on forwarded to the legal queue for review will be a job that may not
+        contain any attached abuse reports (i.e. if an add-on was forwared without a
+        job)."""
+        target = addon_factory()
+        cinder_job = CinderJob.objects.create(job_id='1234', target_addon=target)
+        policy_a = CinderPolicy.objects.create(uuid='123-45', name='aaa', text='AAA')
+        policy_b = CinderPolicy.objects.create(uuid='678-90', name='bbb', text='BBB')
+
+        with mock.patch.object(
+            ContentActionDisableAddon, 'process_action'
+        ) as action_mock, mock.patch.object(
+            ContentActionDisableAddon, 'notify_owners'
+        ) as notify_mock:
+            action_mock.return_value = None
+            cinder_job.process_decision(
+                decision_cinder_id='12345',
+                decision_action=DECISION_ACTIONS.AMO_DISABLE_ADDON.value,
+                decision_notes='teh notes',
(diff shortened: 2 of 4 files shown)


src/olympia/reviewers/tests/test_utils.py
```diff
@@ -3738,7 +3738,10 @@ def _test_request_legal_review(self, *, data=None):
             json={'job_id': '5678'},
             status=201,
         )
-
+        NeedsHumanReview.objects.create(
```
KevinMind, line 3741: Might be overkill but my intuition is that this test is missing something on two ends.

1. You aren't checking the the specific instance you created above is deleted, and we rely on implicit default values that is_active is even true. That's a possible false positive
2. the function that you're testing has a more subtly complex query embedded. It won't clear certain types of reasons. That's a false negative.

I almost want to see you create 2 instances, the one we expect to be removed and the one we expect not to be removed and then verify those specific instances are deleted..

https://github.com/mozilla/addons-server/blob/97c56fdb0bb2372b9f6ad26e99d3cb6d4bd0c6ec/src/olympia/reviewers/utils.py#L1040

src/olympia/abuse/tests/test_models.py
```diff
@@ -1252,6 +1252,35 @@ def test_process_decision_escalate_addon_action(self):
         assert new_job.target_addon == addon
         assert report.reload().cinder_job == new_job
 
+    def test_process_decision_for_legal_reviewed_job(self):
+        """An add-on forwarded to the legal queue for review will be a job that may not
+        contain any attached abuse reports (i.e. if an add-on was forwared without a
+        job)."""
+        target = addon_factory()
+        cinder_job = CinderJob.objects.create(job_id='1234', target_addon=target)
```
KevinMind, line 1260: Is it possible to explicitly set "None" to the values that we are testing are in fact None.

That's probably the default but it helps clarify the intentions of the test and protects against any weird scenario where those values are defined internally.

src/olympia/abuse/models.py
```diff
@@ -312,9 +312,9 @@ def process_decision(
                 if self.target_addon_id
                 else abuse_report_or_decision.addon
             ),
-            rating=abuse_report_or_decision.rating,
-            collection=abuse_report_or_decision.collection,
-            user=abuse_report_or_decision.user,
+            rating=getattr(abuse_report_or_decision, 'rating', None),
```
KevinMind, line 315: Before this patch, would calling this method just raise? Or how would it have behaved before. That would be good to add to the context as well as testing to know how you'd expect it to behave outside of this patchl.

src/olympia/abuse/tests/test_models.py
eviljeff: the fields aren't on `CinderJob` - they're on `AbuseReport` and `ContentDecision` - it's that there aren't any `AbuseReport`s or `ContentDecision`s (both have a foreign key to the `CinderJob`)

src/olympia/reviewers/tests/test_utils.py
eviljeff: https://github.com/mozilla/addons-server/pull/23024/commits/95f23f427264acb6c910fda039dda052c5bae6f2

src/olympia/abuse/models.py
KevinMind: Ah I read it wrong then, I interpreted "failed" as the approach failed not the code.

src/olympia/abuse/tests/test_models.py
eviljeff: It's impossible for that assertion ever to fail, but as requested https://github.com/mozilla/addons-server/pull/23024/commits/c88399c50fb1b9dcdd9291a8b38d5f25c21f77f6

diox (approved):
:+1:

PR #23031, opened 2025-01-27
Don't flag add-ons for growth threshold if hotness is negative

diox:
Fixes: mozilla/addons#15262

### DIFF
--- a/src/olympia/addons/tests/test_tasks.py
+++ b/src/olympia/addons/tests/test_tasks.py
@@ -451,13 +451,14 @@ def test_flag_high_hotness_according_to_review_tier_threshold_check_negative():
         upper_adu_threshold=1000,
         growth_threshold_before_flagging=10,
     )
-    # Average hotness should be negative, -0.1.
+    # Average hotness should be negative, -0.2.
     # growth_threshold_before_flagging for that tier is 10 so we should flag
-    # add-ons with hotness above 0
+    # add-ons with hotness above -0.1, but we explicitly avoid flagging
+    # negative growth add-ons, so only the last one should be flagged.
 
-    addon_factory(average_daily_users=251, hotness=-0.1)
-    addon_factory(average_daily_users=251, hotness=-0.2)
-    addon = addon_factory(average_daily_users=251, hotness=0.0)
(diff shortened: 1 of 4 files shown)


src/olympia/addons/tests/test_tasks.py
```diff
@@ -451,13 +451,14 @@ def test_flag_high_hotness_according_to_review_tier_threshold_check_negative():
...
     # growth_threshold_before_flagging for that tier is 10 so we should flag
-    # add-ons with hotness above 0
+    # add-ons with hotness above -0.1, but we explicitly avoid flagging
+    # negative growth add-ons, so only the last one should be flagged.
 
-    addon_factory(average_daily_users=251, hotness=-0.1)
-    addon_factory(average_daily_users=251, hotness=-0.2)
-    addon = addon_factory(average_daily_users=251, hotness=0.0)
+    addon_factory(average_daily_users=251, hotness=-0.05)
+    addon_factory(average_daily_users=251, hotness=-0.65)
```
KevinMind, line 460: We should still include one with hotness 0.0 to make sure it is clear how that growth would be handled.

KevinMind:
@diox code looks good. Could you add a test section so I can verify locally. Feel free to include a db zip of the initial state if that is easier than setup instructions.

src/olympia/addons/tests/test_tasks.py
diox: Added more tests in 44ae308e7a8a485a26530014298bec90053a0e34

diox:
It's super painful to test locally because there is quite a bit of set up to do. Since the included unit tests replicate that setup I only relied on unit tests while writing this patch myself.

KevinMind:
> It's super painful to test locally because there is quite a bit of set up to do. Since the included unit tests replicate that setup I only relied on unit tests while writing this patch myself.

I'm not a fan of that response, but I understand if it's difficult. Question, assuming after merging this went straight to production, what would be the cost if there was a bug and it was totally wrong.

How often is the "flagging" that is happening, happening?

KevinMind (approved):
Didn't verify 🚢 🙈

PR #23035, opened 2025-01-28
Don't translate admin events / activities hidden to developers

diox:
Extra fix for https://github.com/mozilla/addons/issues/15306


KevinMind:
Can you add a context section? Is it typical not to translate admin text?

KevinMind (approved):
Looks good

diox:
> Can you add a context section? Is it typical not to translate admin text? Since the code changes are in a constants file, what would stop me from using it somewhere outside of that non-localized context?

Nothing would stop you, it's just a convention. We decided some time ago to not translate strings that are only shown to admins/reviewers because that was a lot of extra work for translators and reviewers/admins all work in English anyway.

PR #23036, opened 2025-01-29
Add GitHub Actions health check workflow

KevinMind:
Introduces a new GitHub Actions workflow to periodically check service health and version information for specified endpoints. The workflow includes:
- Scheduled runs every 5 minutes
- Checks version information
- Monitors service status
- Raises an error if any services are down

### Checklist


scripts/health_check.py
```diff
@@ -0,0 +1,87 @@
...
+        self.verbose = verbose
+
+    def _fetch(self, path: str) -> dict[str, str] | None:
+        url = f'{self.environment.value}/{path}'
+        if self.verbose:
+            print(f'Requesting {url} for {self.environment.name}')
+
+        try:
+            response = requests.get(url)
+            response.raise_for_status()
```
diox, line 34: We return a `500` error in the monitor view if one or more services is down alongside a valid json response, so if you raise for status, `data` is not set and that causes your script to raise `UnboundLocalError`. We want to do the `data = response.json()` bit first because it can work even if the response status is not 20x.

scripts/health_check.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+ENV_ENUM = Enum(
+    'ENV',
+    [
+        ('dev', 'https://addons-dev.allizom.org'),
+        ('stage', 'https://addons.allizom.org'),
+        ('prod', 'https://addons.mozilla.org'),
+        # TODO: maybe we could use the local environmnet here
+        ('test', ''),
```
diox, line 17: I would remove all the special `test` bits here and below and just use:
```suggestion
        # For local environments hit the nginx container as set in docker-compose.yml
        ('local', 'http://nginx'),
```

scripts/health_check.py
```diff
@@ -0,0 +1,87 @@
...
+            return {}
+        return self._fetch('__version__')
+
+    def monitors(self):
+        if self.environment.name == 'test':
+            return {
+                'up': {'state': True},
+                'down': {'state': False, 'status': 'something is wrong'},
+            }
+        return self._fetch('services/monitor.json')
```
diox, line 61: We likely want to check `__heartbeat__` and `services/__heartbeat__` (separately) instead

.github/workflows/health_check.yml
```diff
@@ -0,0 +1,67 @@
...
+            echo "Environment $environment is not reachable"
+            exit 1
+          fi
+
+          message=""
+
+          data=$(echo $monitors | jq -r 'to_entries[] | select(.value.state == false) | .key')
+          for monitor in $data; do
+            message="$message\n- $monitor: $(echo $monitors | jq -r ".[\"$monitor\"].status")"
+          done
```
diox, line 59: (goes with my suggestion below to not `raise_for_status()` too early)

I know this is the first step but we likely want to `exit 1` here too if one of the `state` values is `false` so that the health check action is recorded as failing

scripts/health_check.py
KevinMind: So  really a non 200 response is kind of not important for this we could just skip it entirely right? We will already catch the json error if we don't get a good response or how would you suggest I proceed?

diox: Haven't tried it but we should still account for the possibility of a 500 coming from an unhandled exception somewhere in the view, so just have to be careful about that - we wouldn't return a json or if somehow we are, we wouldn't have the `state` key that we check for, so maybe that's ok as long as it fails later in `main()`.

scripts/health_check.py
KevinMind: why?

scripts/health_check.py
KevinMind: So to clarify, should I check for status at all? I guess we should rely on the response being valid json and then inspect the values instead of relying on the http status code. wdyt?

scripts/health_check.py
diox: Because both matter (and `monitor.json` is just an alias for `services/__heartbeat__` - doesn't matter which URL you use, but if adding a check for `__heartbeat__` might as well be consistent and use `services/__heartbeat__` for the other one)

KevinMind: I see, they have different checks. That wasn't super clear until I looked closely at the views. At the risk of continuing that confusion. I've added a __healthcheck__ endpoint which checks all of the relevant monitors in a single go. I think it's worth the trade off since we don't want to disturb any existing callers to the original endpoints and this way our healtcheck does precisely what we want in a single request. wdyt?

diox: Yeah. The trick with these is that `__heartbeat__` is really mandatory for addons-server to function, where as `services/__heartbeat__` is just services, if they are down the site would still work.

scripts/health_check.py
diox: There is still a possibility to raise a 500 from an unhandled exception in the view while still returning json but that's something we currently only do for API exceptions, so I guess it's fine to ignore status for now, and only rely on the response being valid json (if it's not valid json, then you have an unhandled 500 on your hands)

scripts/health_check.py
diox: One issue with adding a new endpoint: dev/stage/prod nginx's config needs to be updated, (and frontend's list of URLs valid without an app/locale, though it's purely for cleanliness here, as we are not going to call that from the frontend).

If you kept just calling both separately you wouldn't have to deal with that.

KevinMind: Reverted to using the existing endpoints.

.github/workflows/health_check.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        environment: ${{fromJson(github.event_name == 'push' && '["local"]' || '["dev","stage","prod"]')}}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/actions/run-docker
```
diox, line 23: Needing the entire docker setup just to run a basic python script that doesn't have any dependencies other than `requests` feels like a waste of resources and is quite slow (even with docker build caching).

diox (changes requested):
While the [run I just did](https://github.com/mozilla/addons-server/actions/runs/13697781122/job/38304036129) is marked as successful if you dig through the logs it raised a 404 and logged an error:

```
Requesting https://addons-dev.allizom.org/services/monitors.json for dev
{'error': JSONDecodeError('Expecting value: line 1 column 1 (char 0)'), 'data': None, 'response': <Response [404]>}
```

scripts/health_check.py
```diff
@@ -0,0 +1,108 @@
...
+        return data
+
+    def version(self):
+        return self._fetch('__version__')
+
+    def heartbeat(self):
+        return self._fetch('__heartbeat__')
+
+    def monitors(self):
+        return self._fetch('services/monitors.json')
```
diox, line 62:
```suggestion
        return self._fetch('services/__heartbeat__')
```

(But first investigate why the 404+error raised in the run I did wasn't causing the health check to fail)

KevinMind:
> While the [run I just did](https://github.com/mozilla/addons-server/actions/runs/13697781122/job/38304036129) is marked as successful if you dig through the logs it raised a 404 and logged an error:
> 
> ```
> Requesting https://addons-dev.allizom.org/services/monitors.json for dev
> {'error': JSONDecodeError('Expecting value: line 1 column 1 (char 0)'), 'data': None, 'response': <Response [404]>}
> ```

typo in the URL its `services/monitor.json`

.github/workflows/health_check.yml
KevinMind: I hear you, but this way it ensures that we are running the same environment as when you run this locally. Otherwise we have to introduce a totally separate way of loading dependencies and I don't think that's worth the cost of maintaining it when we have a ready to go solution for "running stuff" wdyt?

KevinMind:
@diox the issue was in the retry mechanism in the healtcheck script. We were not raising on the final attempt. That is fixed now and I've verified it exits 1 with the incorrect URL. I've also fixed the URL and the related tests.

.github/workflows/health_check.yml
diox: Random observation: AFAICT we don't allow concurrent workflow runs so the slowness could have an impact, if one run takes more than 5 minutes, that might delay the others... And the recent ones all took over 4 minutes...

diox (approved):

.github/workflows/health_check.yml
KevinMind: This workflow doesn't have concurrency rules defined so it should be fine.

KevinMind: I think there are probably areas where we could make it more efficient. Could for example pull the image instead of building or instead of using a matrix job could iterate on the script level in a single job. I'd put those in a follow up after we get the slack notification working and let it run for a week or so to see what actually makes the most sense.

PR #23049, opened 2025-02-03
Remove DEV_MODE and USE_FAKE_FXA_AUTH, replace with PROD_MODE and FXA_CONFIG

KevinMind:
Fixes: mozilla/addons#1

### Description

This commit introduces several changes to simplify and improve the authentication and environment configuration:

- Replaced `DEV_MODE` with `PROD_MODE` derived from version JSON instead of environment variable
- Removed `USE_FAKE_FXA_AUTH` setting replaced with `FXA_CONFIG` to control fake authentication
- Simplified conditional logic around development and production environments to rely on stable build time variables

### Context

This patch simplifies the configuration of "prod" versus "dev" mode by relying on build time parameters that reliably inform what kind of server to run. Environment variables can be overwritten and have no direct connection to the build, but get_version_json can tell you exactly waht kind of image you are running.

Additionally, sometimes you want to run in prod mode, but don't want to setup a real authentication server. Now you can do that because we use a more granular check for fake FXA based on the presence of valid fxa credentials at runtime.

### Dev mode

Run in dev mode

```bash
```

App should run in dev mode with debug toolbar and non compiled assets

### Prod mode

Run in prod mode

```bash
```

Expect compiled assets, no debug toolbar (unless DEBUG=True is explicitly set)
and finally, fake fxa should still be used by default.

### Real auth

To run with real auth you must set the environment variables

```bash
TBD
```

and run in either dev or prod mode. Try both.

### Checklist


src/olympia/amo/utils.py
```diff
@@ -1161,10 +1161,10 @@ def extract_colors_from_image(path):
     return colors
 
 
-def use_fake_fxa():
+def use_fake_fxa(config):
     """Return whether or not to use a fake FxA server for authentication.
     Should always return False in production"""
-    return settings.DEV_MODE and settings.USE_FAKE_FXA_AUTH
+    return config.get('client_id') == '.'
```
diox, line 1167: The main benefit of the explicit settings checks was to avoid accidentally having fake FxA enabled on dev/stage/prod if an environnement variable wasn't set properly (something we can't easily see/test/control ourselves until after deployment).

We should at least keep something like `USE_FAKE_FXA_AUTH`, change it `ALLOW_FAKE_FXA_AUTH` and use it here on top of the check on `client_id` being `.` to keep the same guarantee.

KevinMind: I was thinking about a Django system check that raises if those values are not set in production. Wdyt?

KevinMind: I was actually thinking we could have a django check that ensures those values are not '.' in production. Would that also solve the issue? Production pods would just fail if they start without proper auth configured.

KevinMind: I found a potential compromise. we can re-use the existing `ENV` setting to check if we are running on a "local" environment or not and only allow fxa if both ENV == 'local' and FXA_CONFIG is as expected.

I've had bad experiences previously trying to add "custom" flags in settings. I think we should aim for a narrow waist and have a setting that tells us the information we actually care about "are you in prod or not" in a way that is clear and doesn't have any explicit meaning about how that information should be used.

Otherwise we end up having many of these bespoke flags that all should agree with each other but then become impossible to coordinate correctly.

src/olympia/urls.py
```diff
@@ -20,6 +21,22 @@
...
+def serve_static_files(request, path, **kwargs):
+    if settings.TARGET == 'production':
+        return serve_static(request, path, document_root=settings.STATIC_ROOT, **kwargs)
+    else:
+        return static_serve(request, path, insecure=True, show_indexes=True, **kwargs)
+
+
+def serve_javascript_catalog(request, locale, **kwargs):
+    with translation.override(locale):
+        return JavaScriptCatalog.as_view()(request, locale, **kwargs)
```
diox, line 34: I don't think we want these to be available on "real" environment - so why not use the same trick as for FxA and only make them available if `ENV == 'local'` ? Don't need to move the code around, can just replace the existing condition

src/olympia/lib/settings_base.py
```diff
@@ -1636,3 +1625,9 @@ def read_only_mode(env):
         'manifest_path': STATIC_BUILD_MANIFEST_PATH,
     }
 }
+
+# The environment in which the application is running.
+# This is set by the environment variables in production environments.
+# For local it is hard coded to "local" to guarantee a clear distinction
+# between local and non-local environments.
```
diox, line 1632:
```suggestion
# For local it is hard coded to "local" in `settings.py` to guarantee a clear
# distinction between local and non-local environments.
```

src/olympia/amo/templates/amo/fake_fxa_authorization.html
```diff
@@ -4,7 +4,12 @@
   <h2> {{ _('Register or Log in') }} </h2>
   <form id="fake_fxa_authorization" method="get" action="{{ url('auth:accounts.authenticate') }}">
     <h3> Don't panic! </h3>
-    <p> You're seeing this page instead of the regular authentication page because both <code>settings.DEBUG</code> and <code>settings.USE_FAKE_FXA_AUTH</code> are set to <code>True</code>.</p>
+    <p>
+      You're seeing this page instead of the regular authentication page because the
+      <code>FXA_CONFIG</code> being used has the client_id set to <code>.</code>.
```
diox, line 9:
```suggestion
      You're seeing this page instead of the regular authentication page because the
      <code>FXA_CONFIG</code> being used has the client_id set to <code>.</code> and
      `<code>ENV</code> is set to <code>local</code>.
```

src/olympia/urls.py
KevinMind: I tried that, but got failing tests. My thought was, it's not actually a problem to serve these paths as they effectively cannot be reached in production. and even if they were, that means CDN and NGINX could not serve the file... so wouldn't serving it here actually be an improvement?

If you feel strongly I'll add a gate but it would have to be that we register the URLS and then 404 if env !== local.

When testing this with the gate at the register level the "override_settings" is applied too late so it doesn't behave as expected unless you pass the environment variable when executing the test. That felt like overkill...

KevinMind:
Pausing this for now. Other higher priorities.

PR #23051, opened 2025-02-04
Refactor promoted groups to use APIChoicesWithNone

KevinMind:
fixes: mozilla/addons#15335

### Description

- Replacing hardcoded promoted group constants with an APIChoicesWithNone implementation
- Updating references across multiple files to use the new PROMOTED_GROUP_CHOICES
- Ensuring consistent group ID references and API-related attributes
- Simplifying imports and reducing redundant constant definitions

### Context

This is a step to simplifying the PromotedAddon data model by isolating the PromotedClass instances. If we don't import them all over the code base we can more easily make changes to the data model without breaking things.

### Testing

CI should pass, no changes to any logic, just string/number references.

### Checklist


src/olympia/versions/models.py
```diff
@@ -889,12 +888,13 @@ def transformer_promoted(cls, versions):
         for version in versions:
             v_id = version.id
             groups = [
-                (
-                    PROMOTED_GROUPS_BY_ID.get(approval.group_id),
-                    APP_IDS.get(approval.application_id),
+                (group, APP_IDS.get(approval.application_id))
```
KevinMind, line 891: extract approval_dict.get(v_id, []) to 889

src/olympia/constants/promoted.py
```diff
@@ -2,9 +2,22 @@
 
 from django.utils.translation import gettext_lazy as _
 
+from olympia.api.utils import APIChoicesWithNone
 from olympia.constants import applications
 
 
+PROMOTED_GROUP_CHOICES = APIChoicesWithNone(
+    ('NOT_PROMOTED', 0, 'Not Promoted'),
```
chrstinalin, line 10: I know `NOT_PROMOTED` is here so it's easier to refactor, but looking at this now -- what would distinguish 'None' from `NOT_PROMOTED`? What was the thought behind `APIChoicesWithNone` over `APIChoices` directly?

IIRC, `NOT_PROMOTED` exists to keep the PromotedAddon around even though the Addon is no longer promoted, but if `None` is an option, wouldn't that be functionally the same?

src/olympia/versions/migrations/0009_auto_20200717_1137.py
```diff
@@ -2,7 +2,6 @@
 
 from django.db import migrations, models
 
-from olympia.constants.promoted import RECOMMENDED
```
chrstinalin, line 5: Huh. Does the linter not check migrations?

src/olympia/constants/promoted.py
KevinMind: Eventually we will no longer need NOT_PROMOTED because the state is equivalent to the absence of any PromotedAddonPromotions.

That being said, even in that case, there will be no use case for a None value. I'll update it.

src/olympia/promoted/models.py
```diff
@@ -49,7 +47,9 @@ def __str__(self):
 
     @property
     def group(self):
-        return PROMOTED_GROUPS_BY_ID.get(self.group_id, NOT_PROMOTED)
+        return PROMOTED_GROUPS_BY_ID.get(
+            self.group_id, PROMOTED_GROUPS_BY_ID[PROMOTED_GROUP_CHOICES.NOT_PROMOTED]
+        )
```
chrstinalin, line 52: If I'm understanding this right, are we essentially 1) moving `GROUP_CHOICES` out of the model itself so we can access it independent of PromotedAddon, and 2) referencing `PROMOTED_GROUP_CHOICES`, which has an ID matching the PromotedClass, instead of directly using the PromotedClasses ID?

I'm struggling to see the benefit (specifically in terms of refactoring for multiple promoted groups) as it seems... functionally the same? And just adding a layer of syntactic abstraction (albeit easier to understand this way)?

And, thinking about the Excalidraw diagram some more, I'm still worried about the `promoted_group()` `approved_applications` reference that I mentioned in the existing multiple promoted addons PR, in that [value_list does not work with @property functions like `approved_applications`](https://stackoverflow.com/questions/2143438/is-it-possible-to-reference-a-property-using-djangos-queryset-values-list).

src/olympia/versions/migrations/0009_auto_20200717_1137.py
KevinMind: Correct.. it's terrible. Like probably one of the worst possible things.

src/olympia/promoted/models.py
KevinMind: In the new data model those will not be property functions but queries.

And yes, the point here is to allow the codebase to reference the same "thing" that represents a promoted group, without needing a direct reference to the class itself. If the goals is to remove those classes and replace them with django models, the first step is to isolate them. Right now they are imported everywhere and that makes the second and even more the third step difficult.

chrstinalin (approved):

PR #23073, opened 2025-02-13
Define the ENV setting for all environments (including test/build/local)

KevinMind:
Fixes: mozilla/addons#15358

### Description

Defines a gloabl ENV setting and documents its purpose and meaning.

### Testing

You can inspect for build,test,local.

### Build

Anywhere in the dockerfile add 

RUN echo "env: ${ENV}"

Run the build

```bash
make docker_build_web DOCKER_PROGRESS=plain
```

Expect to see "env: build"

### Test

In any test add

```python
print('env', os.environ.get('ENV'))
print('env_setting', settings.ENV)
```

Expect both to print "test"

### Local

run `make down && make up` (to ensure new compose settings are applied

Check in the terminal

`make shell`
`printenv`

Expect ENV=local in there

Check in settings

`make djshell`
`settings.ENV`

Expect "local" to be returned

### Checklist


eviljeff (approved):
I couldn't verify the `build` testing steps - I'd already done `make up` by the time I read the instructions so I guess everything was cached(?) because I didn't see any `print` output.

KevinMind:
> I couldn't verify the `build` testing steps - I'd already done `make up` by the time I read the instructions so I guess everything was cached(?) because I didn't see any `print` output.    

If you add a line in the docker file it will invalidate the cache. Should be pretty trivial to reproduce.

eviljeff:
> > I couldn't verify the `build` testing steps - I'd already done `make up` by the time I read the instructions so I guess everything was cached(?) because I didn't see any `print` output.
> 
> If you add a line in the docker file it will invalidate the cache. Should be pretty trivial to reproduce.

I added the `RUN echo "env: ${ENV}"` to the end and it didn't.

KevinMind:
> > > I couldn't verify the `build` testing steps - I'd already done `make up` by the time I read the instructions so I guess everything was cached(?) because I didn't see any `print` output.
> > 
> > 
> > If you add a line in the docker file it will invalidate the cache. Should be pretty trivial to reproduce.
> 
> I added the `RUN echo "env: ${ENV}"` to the end and it didn't.

Did you run with DOCKER_TARGET=production? If not, then it would've just skipped that stage. You could add it to the base stage and it will always run.

KevinMind:
Confirming it does work

image

eviljeff:
> > > > I couldn't verify the `build` testing steps - I'd already done `make up` by the time I read the instructions so I guess everything was cached(?) because I didn't see any `print` output.
> > > 
> > > 
> > > If you add a line in the docker file it will invalidate the cache. Should be pretty trivial to reproduce.
> > 
> > 
> > I added the `RUN echo "env: ${ENV}"` to the end and it didn't.
> 
> Did you run with DOCKER_TARGET=production? If not, then it would've just skipped that stage. You could add it to the base stage and it will always run.

I ran exactly what was specified in the testing steps above: `make docker_build_web DOCKER_PROGRESS=plain`

PR #23079, opened 2025-02-14
Add Forward to Legal to important changes history in review page

diox:
Also make that history look more version activity to reduce custom formatting logic.

Fixes mozilla/addons#15327

### Context

"Forward to legal" is one of those actions that is not tied to a version, so in order to show it to reviewers, we add it to the "Important add-on changes history".

#### With the new action
Screenshot 2025-02-14 at 13-17-01 New Tasty Salad 6 – Add-ons for Firefox
#### Some old actions in the new format
Screenshot 2025-02-14 at 13-16-51 New Exquisite Sandwich 90 – Add-ons for Firefox

### Testing

- Review an add-on as an admin, and select the action to forward it to legal
- "Important add-on changes history" should now be shown on review page for that add-on and should include the forward to legal action, with the comment that was made at the time

You can verify that other actions we're supposed to show there (force enable, force disable, author add, author remove, author role change, etc) still show up, but the unit test does that already.

### DIFF
--- a/src/olympia/constants/activity.py
+++ b/src/olympia/constants/activity.py
@@ -55,6 +55,7 @@ class ADD_USER_WITH_ROLE(_LOG):
     action_class = 'add'
     # L10n: {0} is the user role.
     format = _('{user} ({0}) added to {addon}.')
+    short = 'Author added'
     keep = True
     show_user_to_developer = True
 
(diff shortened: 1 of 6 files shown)


KevinMind:
LGTM 🚢 

image

KevinMind (approved):

PR #23081, opened 2025-02-14
Allow API authentication for source package downloads

chrstinalin:
Relates to https://github.com/mozilla/assay/issues/59

### Description

Enables Assay to access & install source code packages for review.

### Testing

- `GETs` to `/firefox/downloads/source/<version_id>` should return the source code if the file exists. 404 otherwise.
- Source code is only available to admin reviewers or developers of the add-on.
- A 404 is returned if there is no version source code, but the user had the right permissions.

### Checklist

### DIFF
--- a/src/olympia/versions/tests/test_views.py
+++ b/src/olympia/versions/tests/test_views.py
@@ -848,6 +848,13 @@ class TestDownloadSource(TestCase):
         xsendfile_header = decode_http_header_value(response[settings.XSENDFILE_HEADER])
         assert xsendfile_header == expected_path
 
+    def test_no_source_with_permission_should_go_in_404(self):
+        self.client.force_login(self.user)
+        self.version.source = None
+        self.version.save()
+        response = self.client.get(self.url)
+        assert response.status_code == 404
+
     def test_no_source_should_go_in_404(self):
         self.version.source = None
         self.version.save()
--- a/src/olympia/versions/tests/test_views.py
+++ b/src/olympia/versions/tests/test_views.py
@@ -810,8 +810,11 @@ class TestDownloadSource(TestCase):
         )
         self.url = reverse('downloads.source', args=(self.version.pk,))
 
+    def login(self, user):
+        return self.client.force_login(user)
+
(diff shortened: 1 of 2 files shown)


src/olympia/versions/views.py
```diff
@@ -183,6 +183,7 @@ def download_latest(request, addon, download_type=None, **kwargs):
 
 
 @non_atomic_requests
+@api_authentication
```
diox, line 186: Can you add tests for this ? Similar to how it was done when `api_authentication` was introduced in 4e800b83d58:
- Add a default `login()` method in some base class that calls  `self.client_force_login()`
- Make `TestDownloadSource` use that `login()` method instead of `self.client_force_login()` directly
- Add a new mixin that implements `login()` through API auth
- Add a new class inheriting from that mixin and `TestDownloadSource`

src/olympia/versions/tests/test_views.py
```diff
@@ -916,3 +926,11 @@ def test_download_for_admin(self):
         # Even deleted!
         self.addon.delete()
         assert self.client.get(self.url).status_code == 200
+
+
+class TestDownloadSourceSessionAPIAuth(TestDownloadSource, APILoginMixin):
```
diox, line 931: Subtle bug: `APILoginMixin` should be first, otherwise you're using the implementation for `login` provided by `TestDownloadSource`, which uses regular auth.

You can verify this by adding a `breakpoint()` inside `APILoginMixin.login()` and running tests for that class, it will never be triggered without changing that ordering.

src/olympia/versions/tests/test_views.py
```diff
@@ -916,3 +926,11 @@ def test_download_for_admin(self):
...
+    client_class = APITestClientSessionID
+
+
+class TestDownloadSourceJWTAPIAuth(TestDownloadSource, APILoginMixin):
```
diox, line 935: Same as above.

diox (changes requested):
Almost there :)

PR #23083, opened 2025-02-17
create override decision for denied 2nd level approvals

eviljeff:
Fixes: mozilla/addons#15094
### Description

Creates an override decision if a 2nd level approval action is to decline the negative action, and approve the content instead.

### Context

We've only recently been able to create override decisions via the API, so previously we just didn't notify Cinder if the 2nd level hold was denied instead of proceeded with.  The second commit is because we previously assumed that a decision could only be overridden once, which is now not a given (though is still the likely case).

Ideally, I'd like to have a decision return an iterator that would allow overridden decisions to be iterated through to the start (or end), rather than having to do a recursive look through `override_of` in 3 different places ( `originating_job`, and in `get_action_helper` and `report_to_cinder`; and `final` goes to the other end) but it'll take some playing around. It could be done as a follow-up.

### Testing

- Set-up a decision that is held for a 2nd level approval:
  -  i.e. reject or disable an add-on in a high impact group (notable, recommended, etc).  Either in reviewer tools or Cinder - though reviewer tools is easier because you don't have to replay the webhook payload
- deny the second level approval - i.e. don't reject/disable, mark the add-on as okay instead
- as before, the developer should not be emailed, because the rejection/disable didn't take place; if the rejection/disable was linked to a 3rd party job (i.e. from an abuse report), then the reporter should get an email saying we didn't take down the content, etc.
- in Cinder there should be a record of both the original decision to reject/disable _and_ the 2nd level decision to approve - with the latter decision as an override of the first.

### Checklist

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -173,8 +173,14 @@ class CinderEntity:
             **data,
             'reasoning': self.get_str(reasoning),
             'policy_uuids': policy_uuids,
-            'enforcement_actions_slugs': [action],
-            'enforcement_actions_update_strategy': 'set',
+            **(
+                {
+                    'enforcement_actions_slugs': [action],
+                    'enforcement_actions_update_strategy': 'set',
+                }
+                if action is not None
+                else {}
+            ),
         }
         response = requests.post(url, json=data, headers=self.get_cinder_http_headers())
         if response.status_code == 201:
@@ -196,6 +202,13 @@ class CinderEntity:
     def create_job_decision(self, *, action, reasoning, policy_uuids, job_id):
         url = f'{settings.CINDER_SERVER_URL}jobs/{job_id}/decision'
         return self._send_create_decision(url, {}, action, reasoning, policy_uuids)
+
+    def create_override_decision(self, *, action, reasoning, policy_uuids, decision_id):
+        url = f'{settings.CINDER_SERVER_URL}decisions/{decision_id}/override/'
+        # TODO: send action too once
+        # https://lindie.app/share/6a21d831b39351d7c6fe898f6d22619af62dde98/PLAT-1834
+        # implements the same parameters for overrides
+        return self._send_create_decision(url, {}, None, reasoning, policy_uuids)
 
     def close_job(self, *, job_id):
         url = f'{settings.CINDER_SERVER_URL}jobs/{job_id}/cancel'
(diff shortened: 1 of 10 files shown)


src/olympia/abuse/cinder.py
```diff
@@ -197,6 +203,13 @@ def create_job_decision(self, *, action, reasoning, policy_uuids, job_id):
         url = f'{settings.CINDER_SERVER_URL}jobs/{job_id}/decision'
         return self._send_create_decision(url, {}, action, reasoning, policy_uuids)
 
+    def create_override_decision(self, *, action, reasoning, policy_uuids, decision_id):
+        url = f'{settings.CINDER_SERVER_URL}decisions/{decision_id}/override/'
+        # TODO: send action too once
+        # https://lindie.app/share/6a21d831b39351d7c6fe898f6d22619af62dde98/PLAT-1834
+        # implements the same parameters for overrides
```
diox, line 210: Can we track this in a ticket on our side as well?

src/olympia/reviewers/utils.py
```diff
@@ -319,7 +319,9 @@ class HeldDecisionQueueTable:
 
     @classmethod
     def get_queryset(cls, request, **kwargs):
-        return ContentDecision.objects.filter(action_date=None).order_by('created')
+        return ContentDecision.objects.filter(
+            action_date=None, overridden_by=None
```
diox, line 323: So if the decision has been overridden in Cinder, we consider it doesn't need second level approval ?

eviljeff: Yes.  It still won't have an action date (so we know that the disable/reject wasn't carried out, and no emails were sent), but it doesn't need any further review.

src/olympia/abuse/cinder.py
eviljeff: https://github.com/mozilla/addons/issues/15371

src/olympia/reviewers/utils.py
diox: Could you refactor this in its own queryset/manager method on `ContentDecision`, like `held_for_second_approval()` or something, adding a short docstring with that info?

eviljeff: https://github.com/mozilla/addons-server/pull/23083/commits/42b4821208de59d36b61b50e567430bec927905a

diox (approved):

PR #23086, opened 2025-02-19
Update local dev docker images for autograph, elasticsearch, memcached

wagnerand:
Fixes: https://github.com/mozilla/addons/issues/15381
Fixes: https://github.com/mozilla/addons/issues/15379

### Description

This updates the local development docker images to match what we are currently using in production.

### Context

As an added benefit, some images are now multi-platform, which is mostly relevant for developers on Apple Silicon or other non-amd64 platforms.
Updating to a more recent autograph version also required some heartbeat parameters to be added to the config, I took those from https://github.com/mozilla-services/autograph/blob/master/autograph.yaml as per https://github.com/mozilla/addons-server/blob/1f3e27cc2bf6a64b64c4e75525df1d7bd05e183a/docker/autograph/autograph_localdev_config.yaml#L2

### Testing

Make sure the containers all start without an error, and perhaps you can reach the website. Not sure what else to test. Perhaps a quick look at the container log files might be useful to see if there are any errors.

### Checklist

### DIFF
--- a/docker/autograph/autograph_localdev_config.yaml
+++ b/docker/autograph/autograph_localdev_config.yaml
@@ -251,3 +251,7 @@ authorizations:
       key: 9vh6bhlc10y63ow2k4zke7k0c3l9hpr8mo96p92jmbfqngs9e7d
       signers:
           - webextensions-rsa-with-recommendation
+
+heartbeat:
+    hsmchecktimeout: 100ms
+    dbchecktimeout: 150ms
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -104,9 +104,7 @@ services:
       - web
 
   memcached:
-    image: memcached:1.4
-    # Remove this once we upgrade to a version that provides multi-platform images
-    platform: linux/amd64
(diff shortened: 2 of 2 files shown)


docker/autograph/autograph_localdev_config.yaml
```diff
@@ -251,3 +251,7 @@ authorizations:
       key: 9vh6bhlc10y63ow2k4zke7k0c3l9hpr8mo96p92jmbfqngs9e7d
       signers:
           - webextensions-rsa-with-recommendation
+
+heartbeat:
+    hsmchecktimeout: 100ms
```
diox, line 256: We're not using an HSM but I suppose it doesn't hurt to have it...

wagnerand: Yea I thought the same thing, I just copied both over from the example config. Fwiw, we are not using a database either, so the db timeout shouldn't apply as well, but I guess that could be an issue to be filed in autograph itself.

PR #23099, opened 2025-02-21
create approve 2nd level decision with non-addon entities too

eviljeff:
Fixes: mozilla/addons#15382

### Description

Correctly creates ContentDecision with all possible entities, not just `addon`.

### Testing

- set up a non-addon decision to require 2nd level approval - e.g. a delete collection owned by TaskUser, by reporting it; choosing to delete the collection in cinder; then replaying the webhook locally
  - or fake it by creating a `ContentDecision` without an `action_date` set
- navigate to the review page for the 2nd level approval via the reviewer tools
- choose to approve (i.e. deny the delete action)
- see: no 500 error; content remains non-deleted; record of Approve decision in Cinder

### Checklist

### DIFF
--- a/src/olympia/reviewers/views.py
+++ b/src/olympia/reviewers/views.py
@@ -1261,6 +1261,9 @@ def decision_review(request, decision_id):
             case 'no':
                 new_decision = ContentDecision.objects.create(
                     addon=decision.addon,
+                    rating=decision.rating,
+                    collection=decision.collection,
+                    user=decision.user,
                     action=DECISION_ACTIONS.AMO_APPROVE,
                     reviewer_user=request.user,
                     override_of=decision,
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/views.py
```diff
@@ -1261,6 +1261,9 @@ def decision_review(request, decision_id):
             case 'no':
                 new_decision = ContentDecision.objects.create(
```
diox, line 1262: I feel we should we have a `deny(action_to_use_instead)` method or something along those lines (with a better name) on `ContentDecision` to move all the code dealing with somebody "cancelling" a held decision.

Doesn't have to be in this PR though.

diox (approved):

src/olympia/reviewers/views.py
eviljeff: AMO_APPROVE is the exceptional case, where there's only one action and policy - to generalize it we'd need to build more of a review tool that would allow a policy selection and action.  (That would also be a replacement for the "forward" action, which is a bit of hack)

eviljeff: So, a `deny_held_action()` on `ContentDecision` is possible, which would be the same code in the view, but moved. A more general `override_action(action, policies)` is dependent on some more form/view work.  I'll log an issue.

eviljeff: https://github.com/mozilla/addons/issues/15396

PR #23107, opened 2025-02-24
Only hold REJECT_VERSION action if it affects signed versions

eviljeff:
Fixes: mozilla/addons#15346

### Description

Limits `should_hold_action` for ContentActionRejectVersion[Delayed] to decisions where `target_versions` contains one or more files that are signed.

### Context

diff is larger than it should be because I refactored some of the tests for ContentActionDisableAddon too, so we have more complete coverage of the functionality of the two classes (and the differences between them)

### Testing

- setup an add-on that may be held for review (e.g. make it Recommended)
- reject a signed version (you can use django admin/shell to tinker with the file is_signed flag if it's easier)
- see the action on the decision is held for 2nd level approval
- reject a version that isn't signed
- see the action on the decision _isn't_ held for 2nd level approval

### Checklist

### DIFF
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -332,8 +332,9 @@ class ContentActionRejectVersion(ContentActionDisableAddon):
         self.content_review = decision.metadata.get('content_review', False)
 
     def should_hold_action(self):
-        return super().should_hold_action() and bool(
-            self.decision.target_versions.filter(file__is_signed=True).exists()
+        return (
+            super().should_hold_action()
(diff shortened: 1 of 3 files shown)


src/olympia/abuse/actions.py
```diff
@@ -331,6 +331,11 @@ def __init__(self, decision):
         super().__init__(decision)
         self.content_review = decision.metadata.get('content_review', False)
 
+    def should_hold_action(self):
+        return super().should_hold_action() and bool(
```
diox, line 335: Don't need to `bool()`, `exists()` already returns a boolean

diox (approved):

PR #23108, opened 2025-02-24
add retry_unreported_abuse_reports command to retry AbuseReports

eviljeff:
Fixes: mozilla/addons#15376

### Description

Fires the `report_to_cinder` celery task for any AbuseReports that should have been reported to Cinder but don't have an associated `CinderJob` (e.g. the reporting to Cinder failed; or the waffle switch was disabled at the time)

### Context

Doesn't depend on https://github.com/mozilla/addons/issues/15359 but makes more sense after it's merged

### Testing

- toggle `dsa-job-technical-processing` waffle switch off
- submit an abuse report for an addon/collection, etc
- check AbuseRepot is created in database but it has no associated CinderJob
- execute `./manage.py retry_unreported_abuse_reports` 
- see AbuseReport now has an associated `.cinder_job` fk
- (re-enable `dsa-job-technical-processing` before you forget 😉)

### Checklist

### DIFF
--- a/src/olympia/abuse/management/commands/retry_unreported_abuse_reports.py
+++ b/src/olympia/abuse/management/commands/retry_unreported_abuse_reports.py
@@ -1,0 +1,20 @@
+from django.core.management.base import BaseCommand
+
+import olympia.core.logger
+from olympia.abuse.models import AbuseReport, AbuseReportManager
+from olympia.abuse.tasks import report_to_cinder
+
+
+class Command(BaseCommand):
+    log = olympia.core.logger.getLogger('z.abuse')
+
+    def handle(self, *args, **options):
+        qs = AbuseReport.objects.filter(
+            AbuseReportManager.is_individually_actionable_q(), cinder_job__isnull=True
+        )
+        self.stdout.write(f'{len(qs)} AbuseReports to report to Cinder')
+
+        for report in qs:
+            # call task to fire off cinder report
+            self.log.info('Created task for %s.', report)
+            report_to_cinder.delay(report.id)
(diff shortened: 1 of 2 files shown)


diox:
AFAICT from the logs, the problem with autoretry is that the exceptions before we give retrying are not logged as `error`s. They are just logged as `info`. (the `severity` - lowercase - is 200 on each attempt that is retried, 500 on the one where we give up)

That would render those exceptions completely invisible to us for 72 hours, they wouldn't be sent to Sentry.

src/olympia/abuse/management/commands/retry_unreported_abuse_reports.py
```diff
@@ -0,0 +1,20 @@
...
+from olympia.abuse.tasks import report_to_cinder
+
+
+class Command(BaseCommand):
+    log = olympia.core.logger.getLogger('z.abuse')
+
+    def handle(self, *args, **options):
+        qs = AbuseReport.objects.filter(
+            AbuseReportManager.is_individually_actionable_q(), cinder_job__isnull=True
+        )
```
diox, line 14: Add `.values_list('pk', flat=True)` IMHO. That makes the query slightly less nasty and since you're fetching it all in one go, consume less memory.

We don't care about the str representation of each report in the `log.info()` call below, logging pks will be enough. You'll need to adjust the `.delay()` call since the queryset would then only contain pks though.

diox (approved):

src/olympia/abuse/management/commands/retry_unreported_abuse_reports.py
eviljeff: https://github.com/mozilla/addons-server/pull/23108/commits/6544c582bf34b14838991da1fefd95f6318a3fd4

PR #23119, opened 2025-02-26
Add task to remove HTML from public add-on summaries

diox:
We switched to completely preventing HTML from being added, so any links that were previously there generated some HTML we need to remove now.

(Note that we're currently aggressively forbidding URLs too, but
 these existing ones are allowed to stay)

My initial approach was to use a migration to copy `localized_string` to `localized_string_clean`, but the command I eventually went for uses `bleach` instead to handle more cases.

Fixes: mozilla/addons#15369

Testing: see unit test. You can replicate locally by creating some translations in the same way that unit test does, and calling `manage.py process_translations --task=strip_html_from_summaries`


eviljeff (commented):
I think this needs a test (possibly with some refactoring if it's not easy to test in this state)

diox:
I tried to look into integrating https://pypi.org/project/django-test-migrations/ but it doesn't like our migrations, and causing weird failures, on top of being quite slow (since it needs to roll back the state of the entire db every time you want to test a single migration).

I'll move that to a command for now to make it testable.

src/olympia/translations/tasks.py
```diff
@@ -39,3 +41,25 @@ def update_outgoing_url(pks, *, old_outgoing_url, **kw):
...
+    # Note: can't just use PureTranslation.clean(), because we want to strip
+    # the HTML and not just escape it here - We used to create that HTML
+    # automatically, so it's out responsability to remove it, escaping it not
+    # enough.
+    cleaner = bleach.Cleaner(tags=[], attributes={}, strip=True)
+    translations = Translation.objects.filter(pk__in=pks)
+    for translation in translations:
+        translation.localized_string = cleaner.clean(translation.localized_string)
+        translation.localized_string_clean = translation.localized_string
+        translation.save()
```
eviljeff, line 65: Do we need to trigger ES reindexing, or is there already a signal connected to the Translation instances?

eviljeff (approved):

src/olympia/translations/tasks.py
eviljeff: Actually... either way, is there a way to verify that in tests?  ("no" is a valid answer 😬)

diox: That's a good point - since I'm updating the translations directly there wouldn't be an automatic reindex, we need to trigger that ourselves. Testing is easy, with a `ESTestCase` and performing a search after we've updated the translations. Implementing it is going to be annoying as we need to do the reverse query to find the add-on from the translation id, but that's doable, I'll look into it.

diox: https://github.com/mozilla/addons-server/pull/23119/commits/f240a5da3101fa926325ce8af29d7c2cadf970ee (ended up just mocking `index_addons`, no need for a real ES test case)

PR #23122, opened 2025-02-27
Split forwarded appeal into separate review queue filter

wagnerand:
Fixes: mozilla/addons#15414

### Description

This changes the filter in the reviewer tools queue so that appeals forwarded from cinder have their own filter item, rather than being merged into "Abuse report forwarded from Cinder".
Smaller changes include, in separate commits:
* Renaming some filter items for clarity
* Reordering the filter list

#### Before

Screenshot 2025-02-27 at 19 11 38

#### After 

Screenshot 2025-02-27 at 19 10 38

### Context

Given the anticipated amount of additional work and the assumed different priorities between cinder-fowarded abuse reports and appeals, reviewers should be able to filter them separately.

### Testing

I tested locally that the labels and order appear as expected. Testing whether the new filter item itself works would be good, but I didn't know how to fake an item in the queue appropriately. What you'd need to do is create an add-on that is in the queue because an a) abuse report, b) appeal has been forwarded from cinder. Then, using the checkboxes on the reviewer tools queue page, check if the respective add-on is shown/hidden.

### Checklist


eviljeff:
> I didn't know how to fake an item in the queue appropriately

The full procedure would be to submit a report; process in cinder; replay webhook; submit appeal; forward that appeal in cinder; then replay webhook, but for purposes of this pr just creating the NeedsHumanReview with that reason is enough.

eviljeff (approved):

src/olympia/versions/models.py
```diff
@@ -236,23 +245,17 @@ def get_due_date_reason_q_objects(cls):
...
             'is_from_theme_awaiting_review': is_from_theme_awaiting_review,
-            # Everything else shares the is_other_needs_human_review condition.
-            'needs_human_review_auto_approval_disabled': Q(
-                is_other_needs_human_review,
-                needshumanreview__reason=NeedsHumanReview.REASONS.AUTO_APPROVAL_DISABLED,
+            'needs_human_review_from_cinder_forwarded_abuse': Q(
+                needshumanreview__is_active=True,
+                needshumanreview__reason__in={
+                    NeedsHumanReview.REASONS.CINDER_ESCALATION,
+                },
```
eviljeff, line 254: this should have been changed to
```python
                needshumanreview__reason=NeedsHumanReview.REASONS.CINDER_ESCALATION,
```
I'll fix in the PR for https://github.com/mozilla/addons/issues/15413

wagnerand: Ah thanks. It's just a code-quality issue, right? The code should work either way? At least I did not see any errors in tests or during manual testing.

PR #23138, opened 2025-03-05
always detect if there is an attachment for a decision

eviljeff:
follow-up fix for: mozilla/addons#15412

### Description

changes how we detect if there is an attachment for a decision, to correctly include the template snippet about it.

### Context

It appears while fixing #15412 I fixed the attachment being available for download in developer hub, but inadvertently stopped the existence of the attachment being exposed in the email.
The more complete fix would be to refactor how `send_notifications` selects the appropriate activity log but to limit the scope of this pr, which is expected to be cherry-picked, I instead changed how `has_attachment` was determined.

### Testing

Follow the testing steps from https://github.com/mozilla/addons-server/pull/23121

### Checklist

### DIFF
--- a/src/olympia/abuse/tests/test_models.py
+++ b/src/olympia/abuse/tests/test_models.py
@@ -2886,23 +2886,6 @@ class TestContentDecision(TestCase):
             expect_create_override_call=True,
         )
 
-    def _check_notify_emails(self, decision, log_entry):
-        assert len(mail.outbox) == 1
-        assert mail.outbox[0].to == [decision.addon.authors.first().email]
-        assert str(log_entry.id) in mail.outbox[0].extra_headers['Message-ID']
-        assert 'days' not in mail.outbox[0].body
-        assert 'some review text' in mail.outbox[0].body
-        assert 'some policy text' not in mail.outbox[0].body
-        AttachmentLog.objects.create(
-            activity_log=log_entry,
-            file=ContentFile('Pseudo File', name='attachment.txt'),
-        )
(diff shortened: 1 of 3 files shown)


src/olympia/abuse/tests/test_models.py
```diff
@@ -2886,23 +2886,6 @@ def test_report_to_cinder_approve_override_previous_decision_has_cinder_id(self)
...
-        AttachmentLog.objects.create(
-            activity_log=log_entry,
-            file=ContentFile('Pseudo File', name='attachment.txt'),
-        )
-        decision.send_notifications()
-        assert 'An attachment was provided.' not in mail.outbox[0].body
-        assert 'To respond or view the file,' not in mail.outbox[0].body
-        assert 'An attachment was provided.' in mail.outbox[1].body
-        assert 'To respond or view the file,' in mail.outbox[1].body
-
```
eviljeff, line 2905: btw, this is just refactoring because I realised `_check_notify_emails` was only used in one test, so it was confusing - I initially thought I could change here to test multiple tests.  code has been cut&pasted into `test_execute_action_with_action_date_already` directly.

KevinMind (approved):
Still a bit confusing how this was passing tests before but verified.

eviljeff:
> Still a bit confusing how this was passing tests before but verified.

There wasn't coverage for it - in the previous PR I added tests for the changes in ContentActionDisableAddon, but this fail was coming from further up the call stack, in ContentDecision. Those tests didn't test attachments with held actions.

PR #23141, opened 2025-03-06
Migrate all admin js/css to vite compilation

KevinMind:
Fixes: mozilla/addons#15432

## Description

Migrate remaining admin pages to use vite for asset compilation

## Context

Removed redundant `AMOModelAdmin.Media.css['all']` and `AMOModelAdmin.Media.js` inclusions as django admin automatically merges `Media` class js/css properties.

## Testing

Run the following tests in prod/dev modes

```
make up DOCKER_TARGET=<production|development>
``` 

For each of these admin pages:
- verify there are no 404/500 type responses for all js/css assets
- verify there are no critical js console errors
- verify any additional requirements on the specific page

> [!NOTE]
> there are a few expected errors you can ignore CSP google analytics errors 

### [BlocklistSubmissionAdmin](http://olympia.test/en-US/admin/models/blocklist/blocklistsubmission/1/change/))

> [!NOTE]
> Assuming you have a blocklist submission

Expect the "add or change blocks" section has styled content, namely the border on the addon guid input box

image

### [DiscoveryItemAdmin](http://olympia.test/en-US/admin/models/discovery/discoveryitem/1/change/)

> [!NOTE]
> There should be at least one discovery item created, if not create one. 

Expect the preview text to be formatted correctly

> [!NOTE]
> I discovered an invalid css property "font-weight: medium" which has been replaced with the valid equivalent "500" to indicate a "light" boldness. 

image

### [PromotedAddon](http://olympia.test/en-US/admin/models/discovery/promotedaddon/)

> [!NOTE]
> Select a promoted addon, and on the hero section verify the below 

Expect the color gradient selection tool is styled correctly

image

> [!NOTE]
> expect some hero images to be uploaded [here](http://olympia.test/en-US/admin/models/discovery/primaryheroimageupload/)

Verify that the select image options are styled correctly

image

### [ScannerQueryResult](http://olympia.test/en-US/admin/models/scanners/scannerqueryresult/)

It's not totally clear to me how exactly to reproduce the conditions for the scanner results admin sections.

There is js and css that adds and styles custom classes depending on some logic... TBD

image

### [VersionAdmin](http://olympia.test/en-US/admin/models/versions/version/)

> [!NOTE]
> Make sure you have a version with a release note or other translated field

Verify that the translated fields (those that have .trans class) have the locale string prefix.

TBD add the screenshot

### [ActivityLog](http://olympia.test/en-US/admin/models/activity/activitylog/)

Verify that you can add/remove IP addresses via the search results page. Follow this video for details.

https://github.com/user-attachments/assets/3568a345-7120-4749-bf25-88e4f33548f0

## Checklist

### DIFF
--- a/static/css/admin-abuse-report.less
+++ b/static/css/admin-abuse-report.less
@@ -1,0 +1,1 @@
+@import url('./admin/abuse_reports.css');
--- a/static/js/admin-scanner-query-result.js
+++ b/static/js/admin-scanner-query-result.js
@@ -1,0 +1,1 @@
+import './admin/scannerqueryresult';
--- a/static/js/admin-versions.js
+++ b/static/js/admin-versions.js
@@ -1,0 +1,1 @@
+import './admin/l10n.js';
--- a/src/olympia/abuse/admin.py
+++ b/src/olympia/abuse/admin.py
@@ -16,6 +16,7 @@ from olympia import amo
 from olympia.access import acl
 from olympia.addons.models import Addon, AddonApprovalsCounter
 from olympia.amo.admin import AMOModelAdmin, DateRangeFilter, FakeChoicesMixin
+from olympia.amo.templatetags.jinja_helpers import vite_asset
 from olympia.ratings.models import Rating
 from olympia.translations.utils import truncate_text
 
@@ -91,13 +92,8 @@ class MinimumReportsCountFilter(FakeChoicesMixin, admin.SimpleListFilter):
 
 
 class AbuseReportAdmin(AMOModelAdmin):
-    class Media(AMOModelAdmin.Media):
-        css = {
-            'all': (
-                'css/admin/amoadmin.css',
-                'css/admin/abuse_reports.css',
-            )
-        }
+    class Media:
+        css = {'all': (vite_asset('css/admin-abuse-report.less'),)}
 
     date_hierarchy = 'modified'
     list_display = (
--- a/src/olympia/addons/admin.py
+++ b/src/olympia/addons/admin.py
@@ -176,12 +176,9 @@ class FileInline(admin.TabularInline):
 
 
 class AddonAdmin(AMOModelAdmin):
-    class Media(AMOModelAdmin.Media):
-        css = {
-            'all': AMOModelAdmin.Media.css['all']
-            + (vite_asset('css/admin-addon.less'),)
-        }
-        js = AMOModelAdmin.Media.js + (
+    class Media:
+        css = {'all': (vite_asset('css/admin-addon.less'),)}
+        js = (
             # TODO: This is probably a redundant dependency
             'admin/js/jquery.init.js',
             vite_asset('js/admin-addon.js'),
--- a/src/olympia/amo/admin.py
+++ b/src/olympia/amo/admin.py
@@ -4,7 +4,6 @@ import operator
 from collections import OrderedDict
 
 from django import forms
-from django.conf import settings
 from django.contrib import admin
 from django.contrib.admin.options import IncorrectLookupParameters
 from django.contrib.admin.views.main import (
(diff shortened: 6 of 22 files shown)


static/css/admin-abuse-report.less
```diff
@@ -0,0 +1 @@
+@import url('./admin/abuse_reports.css');
```
eviljeff, line 1: I don't understand why we have to do this - if there's only a single file why just just import it directly in the `vite_asset`?

static/js/admin-scanner-query-result.js
```diff
@@ -0,0 +1 @@
+import './admin/scannerqueryresult';
```
eviljeff, line 1:
```suggestion
import './admin/scannerqueryresult.js';
```
?

static/js/admin-versions.js
```diff
@@ -0,0 +1 @@
+import './admin/l10n.js';
```
eviljeff, line 1: similar question to the css files - why do we need to have this layer of misdirection for a single file?

static/css/admin-abuse-report.less
eviljeff: Or rename `abuse_reports.css` to `admin-abuse-report.less` if it's about consistent naming.

eviljeff: (and so on for the other css files - I won't repeat the comment)

eviljeff (commented):
This seems fine, but can you clarify the single files with single import files?

static/css/admin-abuse-report.less
KevinMind: The issue is that in django we can only specify files which are "definitely" included in the vite bundle. Those files are configured in `vite.config.js` to include any file in the root `static/js` or `static/css` folders.

So the files in those roots are included in vite as addressable bundles that django can include in the HTML.
Why do we have those files importing single files. The reason for that is to prevent changing the "existing" files during this migration process.

You'll notice that now that we have vite bundles which have direct css/js imports, we can actually use static analysis to figure out exactly what files import what other files, a dependency graph. Once we have all of the django bundles migrated to vite bundles, we can **then** clean up and rearrange things much more safe-er because vite will tell us if something is incorrect or moved out of place.

I could have configured vite to include ALL files in the static directory as bundles, but that would have created a huge number of unnecessary bundles bloating our static folder even more.

Once we have all of the files migrated, we will know exactly which files are not needed and can delete them, and we will know which files are "duplicates" and we can merge them.

static/js/admin-scanner-query-result.js
KevinMind: You don't need to specify the file extension. Vite will figure it out.

static/js/admin-versions.js
KevinMind: Addressed in the original comment.

static/js/admin-scanner-query-result.js
eviljeff: oh, okay. The file extension was specified in all the other files. I don't mind whatever codestyle we settle on, but it'd be good to start consistently importing.

static/css/admin-abuse-report.less
eviljeff: Okay.  From the description it seemed like _this_ was the final step, but if we have to finish all of it (i.e. the non-admin js/css too) then I guess it has to wait.

static/js/admin-scanner-query-result.js
KevinMind: Yes, I'll update the others. Vite will allow either. but right we should have consistency. This is also something we can enforce with eslint again once the migration is done.

eviljeff (approved):

PR #23153, opened 2025-03-10
Add command to check locales completion rates and send emails about it

diox:
Fixes: mozilla/addons#15175

### Context

We're going to use this new command to check how complete translations are in AMO (starting with a fairly low threshold of 40%) and warn us about anything weird, using pontoon API.

This will be added to AMO cronjobs (running from AMO infra gets us an email server and actual prod settings) by SRE, to be run monthly.

### Testing

Run `manage.py check_locales_completion_rate` to try it locally, checkout out the standard output as well as the fake email to see what it has generated - depends on pontoon API response.

Check included tests for more responses with forced fake content.


src/olympia/amo/management/commands/check_locales_completion_rate.py
```diff
@@ -0,0 +1,172 @@
...
+        'de',
+        'en-US',
+        'es-ES',
+        'fr',
+        'he',
+        'it',
+        'ja',
+        'pl',
+    )
+    EMAIL_RECIPIENTS = ('amo-notifications+locales@mozilla.com',)
```
diox, line 55: Need to create a new group and change that address

src/olympia/amo/templates/amo/emails/locales_completion.ltxt
```diff
@@ -0,0 +1,18 @@
...
+{% endif %}
+{% if locales_to_keep_despite_being_below_threshold %}
+The following locales are below threshold, but should never be removed from AMO:
+- {{ locales_to_keep_despite_being_below_threshold }}
+{% endif %}
+{% if locales_above_threshold %}
+The following locales are above threshold and not yet enabled:
+- {{ locales_above_threshold }}
+{% endif %}
+{% if not locales_below_threshold and not locales_above_threshold and not locales_to_keep_despite_being_below_threshold %}
```
KevinMind, line 14: We should just not send the email in this case or? What is the argument for regularly sending an email with no information in it?

src/olympia/amo/management/commands/check_locales_completion_rate.py
```diff
@@ -0,0 +1,175 @@
...
+                self.stdout.write(f'❌ {locale} is in both sets !')
+                locales_above_threshold.remove(locale)
+        return locales_below_threshold, locales_above_threshold
+
+    def fetch_data(self, project):
+        q = re.sub(r'\s', '', self.QUERY_TEMPLATE % {'project': project})
+        self.stdout.write(f'Calling pontoon with {q}')
+        response = requests.get(self.PONTOON_API, {'query': q})
+        response.raise_for_status()
+        data = response.json().get('data', {})
```
KevinMind, line 124: Graphql apis generally return errors in the response and usually response 200 even with errors.. we might need to check if pontoon might be returning an "error" object in the response.

src/olympia/amo/management/commands/check_locales_completion_rate.py
```diff
@@ -0,0 +1,175 @@
...
+        """Handle command arguments."""
+        parser.add_argument(
+            '--dry-run',
+            action='store_true',
+            dest='dry_run',
+            default=False,
+            help='Fetch data from Pontoon but do not send emails.',
+        )
+
+    def handle(self, *args, **options):
```
KevinMind, line 69: This is quite a large command and I wonder how it will actually run in the end if it is run via a cron job that executes the command.. do crons run on worker by default?

Wondering also because it might be nice to have some basic retry logic here in case of a temporary failure. using a task with basic retry could give us this out of the box.. just thoughts. I know this is kind of a "first step" but that decision is easier to make right the first time than to fix later.

src/olympia/amo/management/commands/check_locales_completion_rate.py
```diff
@@ -0,0 +1,175 @@
...
+                    [
+                        self.pretty_locale_name(locale)
+                        for locale in locales_to_keep_despite_being_below_threshold
+                    ]
+                )
+            ),
+        }
+        template = loader.get_template(self.EMAIL_TEMPLATE_PATH)
+        message = template.render(context)
+        send_mail(self.EMAIL_SUBJECT, message, recipient_list=self.EMAIL_RECIPIENTS)
```
KevinMind, line 175: AFAICT the `dry_run` is not used here.. it probably should be.

src/olympia/amo/management/commands/check_locales_completion_rate.py
```diff
@@ -0,0 +1,175 @@
...
+        # on our side.
+        for locale in settings.AMO_LANGUAGES:
+            if locale != settings.LANGUAGE_CODE and seen_locales[locale] != len(
+                self.PONTOON_PROJECTS
+            ):
+                self.stdout.write(
+                    f'❌ {locale} is only in {seen_locales[locale]} project(s) !'
+                )
+                locales_below_threshold.add(locale)
+        # Somewhat inefficient but good enough check to remove any locales that
```
KevinMind, line 110: Nit: You have sets so you could use set intersection to get a more efficient filter?

```python
duplicate_locales = locales_above_threshold & locales_below_threshold
locales_above_threshold = locales_above_threshold - duplicate_locales
```

src/olympia/amo/templates/amo/emails/locales_completion.ltxt
diox: It's only going to be sent monthly, so it's not going to be super spammy.

I wanted to still send it in this case to let people know everything is working properly. It's a cheap built-in monitoring system :)

src/olympia/amo/management/commands/check_locales_completion_rate.py
```diff
@@ -0,0 +1,175 @@
...
+            locales_above_threshold=locales_above_threshold,
+        )
+
+    def find_locales_below_and_above_threshold(self):
+        locales_below_threshold = set()
+        locales_above_threshold = set()
+        seen_locales = defaultdict(int)
+        for project in self.PONTOON_PROJECTS:
+            data = self.fetch_data(project)
+            for locale_data in data.get('project', {}).get('localizations', []):
```
KevinMind, line 84: nit: fetch_data could be named fetch_locales and just return the array since we are already doing dictionary digging. Could also format the locale/complettion in that method as well.

```python
for locale, completion in self.fetch_locales(project):
    # Do stuff

File: src/olympia/amo/management/commands/check_locales_completion_rate.py [hunk 8]
```diff
@@ -0,0 +1,175 @@
...
+    def fetch_data(self, project):
+        q = re.sub(r'\s', '', self.QUERY_TEMPLATE % {'project': project})
+        self.stdout.write(f'Calling pontoon with {q}')
+        response = requests.get(self.PONTOON_API, {'query': q})
+        response.raise_for_status()
+        data = response.json().get('data', {})
+        return data
+
+    def get_completion(self, locale_data):
+        return round(locale_data['approvedStrings'] / locale_data['totalStrings'] * 100)
```
KevinMind, line 128: if `totalStrings` is ever zero, this could raise. might want to have a quick check for it.

src/olympia/amo/management/commands/check_locales_completion_rate.py
```diff
@@ -0,0 +1,175 @@
...
+                    f'❌ {locale} is only in {seen_locales[locale]} project(s) !'
+                )
+                locales_below_threshold.add(locale)
+        # Somewhat inefficient but good enough check to remove any locales that
+        # both below and above threshold because of differences between the
+        # projects.
+        for locale in locales_above_threshold:
+            if locale in locales_below_threshold:
+                self.stdout.write(f'❌ {locale} is in both sets !')
+                locales_above_threshold.remove(locale)
```
KevinMind, line 116: Another issue here is modifying the set in place. AFAIK this _can_ be problematic. Another way could be 

```python
locales_above_threshold -= locales_above_threshold & locales_below_threshold
```

That can also be combined with the other case of using intersection by extracting the intersection to a variable.

src/olympia/amo/management/commands/check_locales_completion_rate.py
```diff
@@ -0,0 +1,175 @@
...
+        self.warn_about(
+            locales_below_threshold=locales_below_threshold,
+            locales_above_threshold=locales_above_threshold,
+        )
+
+    def find_locales_below_and_above_threshold(self):
+        locales_below_threshold = set()
+        locales_above_threshold = set()
+        seen_locales = defaultdict(int)
+        for project in self.PONTOON_PROJECTS:
```
KevinMind, line 82: GraphQL also typically supports multiple queries in a single request. There are only 2 projects but that could be less prone to random errors and _technically_ more efficient.

```python
QUERY = """
query {
  amo: project(slug: "amo") {
    ...projectFields
  }
  amoFrontend: project(slug: "amo-frontend") {
    ...projectFields
  }
}

fragment projectFields on Project {
  name
  localizations {
    locale {
      code
      name
    }
    totalStrings
    approvedStrings
  }
}
"""
```

That would probably do the trick.

src/olympia/amo/management/commands/check_locales_completion_rate.py
KevinMind: Also, could skip if both sets are empty to prevent needless email sending.

src/olympia/amo/management/commands/check_locales_completion_rate.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+
+    def handle(self, *args, **options):
+        locales_below_threshold, locales_above_threshold = (
+            self.find_locales_below_and_above_threshold()
+        )
+        self.warn_about(
+            locales_below_threshold=locales_below_threshold,
+            locales_above_threshold=locales_above_threshold,
+        )
```
KevinMind, line 76: Should we add a statsd counter here so we can get an alert if the job is for whatever reason not running at least one time in the last X days.

KevinMind (commented):
A few small-medium sized nits. I think I would strongly recommend the:
- task for core logic to benefit from standard retry (we should start making that the **default** way to implement async logic that has external dependencies.
- set intersection for duplicate/conflicting results
- single request for all the data and formatting in the fetch method (easier to read and test and fewer things can go wrong)

src/olympia/amo/management/commands/check_locales_completion_rate.py
diox: Ah yeah forgot the `dry_run`... I don't think it's needed in the end: we'll never do a dry run in prod, and locally we already have `FakeEmail`

src/olympia/amo/management/commands/check_locales_completion_rate.py
diox: We could yeah

src/olympia/amo/management/commands/check_locales_completion_rate.py
diox: It's running fairly quickly (seconds on my machine) and it's not a critical command... crons run on a special dedicated instance.

I don't think it's worth having this as a task, but I could add retries support through `requests` if we think that's worth it ? Ultimately it's not a critical command, and we'll see failures in Sentry so I wasn't particularly worried about it.

src/olympia/amo/management/commands/check_locales_completion_rate.py
diox: I had 400 for incorrect fields. For instance, this 400s:
```
curl -gi 'https://pontoon.mozilla.org/graphql?query={projecterrorpleaseignore(slug:"amo-frontend"){name,localizations{locale{code,name}totalStrings,approvedStrings}}}' 
```

src/olympia/amo/management/commands/check_locales_completion_rate.py
diox: And do what though ? If pontoon is reporting a `0` here I'm okay with raising: this should be reported in Sentry, it's unexpected behavior.

src/olympia/amo/management/commands/check_locales_completion_rate.py
diox: Added basic `requests` retrying (6 tries, with backoff)

src/olympia/amo/management/commands/check_locales_completion_rate.py
KevinMind: If it is 0 you could just return 0. That would indicate there are no translations for a locale right? Since this is a job that runs automatically and regularly might be good to avoid typical cases where it could raise.

src/olympia/amo/management/commands/check_locales_completion_rate.py
KevinMind: Since we recently talked about retries, I thought it a good opportunity to promote that as the _reflexive and default_ behavior for work that depends on external services. But ok, like you said it's not critical.

src/olympia/amo/management/commands/check_locales_completion_rate.py
KevinMind: Just wondering if it returns an error property in data. GraphQL apis are weird and can fail both with status or property. Depends on how pontoon implemented it but might be worth double checking.

diox: It does return `errors` but it would raise before we parse since it returns a proper error code.

src/olympia/amo/management/commands/check_locales_completion_rate.py
diox: I would prefer raising, because it's not expected behavior. If we return `0` then it will impact the output and if it's a global bug affecting multiple locales we'd never know which locales should legitimately be removed or which just had a bug

diox:
The standard output is really only useful for debugging or if we want to dig deeper in the results - it's more verbose, since it includes completion per project, something the people receiving the email won't care about.

PR #23154, opened 2025-03-11
Return minimal user profile for non-developer users.

eviljeff:
Fixes: mozilla/addons#15402

### Description

Changes accounts API endpoint to return a response for all users, but with less fields for a non-developer.  API shims are in-place so any consumer that expects the payload to either be a 404, or to contain all the data, don't fail - we're returning `null` values for the missing fields. 

### Context

Frontend is initially a consumer that will (likely?) fail if presented with a user profile response with most of the fields missing, so the API_GATE is enabled for v5 too.  Once frontend is patched we can drop the api_gate for v5.  Some fields aren't nullified because the api contract defines them as boolean, etc, with no valid null or empty values, so they're returned (though are going to empty/useless in all non-edge-cases anyway) - `created` is probably the only field that has a value we can't sanitise easily, so I just left it in.

~After writing this patch I'm pretty sure we no longer need to store the `has_full_profile` in the ES index - I could drop it here or in a follow-up?~ (dropped in this patch)  

And the need for a `has_full_profile` denormalized field on the model isn't particularly clear either these days - we access it in 2 places after this patch - to determine what serializer to use for the accounts response, and whether we want the url in the sitemap - both of those could be (likely) replaced with calculations based on preloaded values from the model instance, really.  That would certainly be a follow-up issue though.

### Testing

- set up a user who has a some combination of non-empty biography, occupation, location, or homepage (but isn't an author of any publicly listed add-ons).
- access that user's profile with `/api/v5/accounts/account/(int:user_id)/`
- see the user is returned, but biography (etc) returned as null values.
- also access that user's profile via frontend `/firefox/user/(int:user_id)/`
- repeat while authenticated and see the values returned on the profile page (because you're an admin)
  - you can go through the api authentication too, but it's a bit of a pain, and the local frontend should be sufficient.
- make a rating with some text on a random add-on, and see the url to their profile is linked on the ratings page for that add-on.
- to test the case without the API_GATE shim you have to change the `DRF_API_GATES` setting for `v5` to not have the gate enabled.
- then repeat above for a non-authenticated user and see the extra fields are absent entirely from the response - only `name`, `url`, `id`, `username` should be returned in the api response.
- make that user an author of a publicly listed add-on and the response should be as now - full fields with url, etc.

### Checklist

### DIFF
--- a/src/olympia/addons/indexers.py
+++ b/src/olympia/addons/indexers.py
@@ -443,7 +443,6 @@ class AddonIndexer:
                         'id': {'type': 'long'},
                         'name': {'type': 'text'},
                         'username': {'type': 'keyword'},
-                        'has_full_profile': {'type': 'boolean', 'index': False},
                     },
                 },
                 'modified': {'type': 'date', 'index': False},
@@ -650,12 +649,7 @@ class AddonIndexer:
         data['category'] = [cat.id for cat in obj.all_categories]
         data['current_version'] = cls.extract_version(obj, obj.current_version)
         data['listed_authors'] = [
-            {
-                'name': a.name,
-                'id': a.id,
-                'username': a.username,
-                'has_full_profile': a.has_full_profile,
-            }
+            {'name': a.name, 'id': a.id, 'username': a.username}
             for a in obj.listed_authors
         ]
 
--- a/src/olympia/addons/indexers.py
+++ b/src/olympia/addons/indexers.py
@@ -443,7 +443,7 @@ class AddonIndexer:
                         'id': {'type': 'long'},
                         'name': {'type': 'text'},
                         'username': {'type': 'keyword'},
-                        'is_public': {'type': 'boolean', 'index': False},
+                        'has_full_profile': {'type': 'boolean', 'index': False},
                     },
                 },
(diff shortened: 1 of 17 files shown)


src/olympia/addons/indexers.py
```diff
@@ -443,7 +443,7 @@ def get_mapping(cls):
                         'id': {'type': 'long'},
                         'name': {'type': 'text'},
                         'username': {'type': 'keyword'},
-                        'is_public': {'type': 'boolean', 'index': False},
+                        'has_full_profile': {'type': 'boolean', 'index': False},
```
diox, line 446: I would drop it entirely from the index and from `fake_object()` in `ESAddonSerializer`. We don't need that data to build the `authors` from ES data since that is using `BaseUserSerializer`, which no longer cares about whether we're dealing with a public profile or not (and I would argue since it's for listed authors it probably didn't make sense before)

eviljeff: https://github.com/mozilla/addons-server/pull/23154/commits/491bb38923e457b1ca65b2dcc681c1f75bab81f7

diox (approved):

PR #23164, opened 2025-03-12
Add py-call-uwsgi-fork-hooks to local development uwsgi config

diox:
This shouldn't have any impact, just making things more consistent as [we are going to want that enabled in dev/stage/prod](https://mozilla-hub.atlassian.net/browse/SVCSE-2554).

Helps mozilla/addons#1994

Recent versions of sentry sdk will warn if this is not set, it resolves some uwsgi workers dying with those versions of the sentry sdk.


KevinMind (approved):
🚢

PR #23166, opened 2025-03-12
Add Developer Experience Survey to DevHub

chrstinalin:
Fixes: mozilla/addons#15164

### Description

Adds a link to the developer experience survey via a banner on DevHub.

image

### Testing
The banner is only visible if the user has 1) uploaded an add-on in the past 30 days, and 2) has not interacted with the banner (i.e clicked the survey link or dismissed the banner) within 180 days.

1. Enable the waffle switch `enable_dev_experience_survey`.
2. Banner should not be visible when user is not logged in.
3. Log-in to DevHub.
4. Banner should not be visible if developer has not uploaded an add-on within the past 30 days.
5. Create or update an add-on.
6. Banner should be visible.
7. Dismiss (by closing the banner) or navigate to the survey link. The banner should be closed.
8. Subsequent refreshes should not show the banner.

### Checklist


diox (commented):
Could you rebase with the latest `master` to make sure CI is working properly on that PR ?

src/olympia/templates/base.html
```diff
@@ -56,6 +56,17 @@
         {% block bodyattrs %}{% endblock %}>
 
     <div id="main-wrapper">
+      {% if user.is_authenticated and user.is_survey_eligible(survey_id=amo.DEV_EXP_SURVEY_ALCHEMER_ID) and switch_is_active('dev_exp_survey') %}
```
diox, line 59: We don't want to add that extra database query to every reviewer page. Can you move that block to `devhub/base.html` ? You can re-use the `{% block site_notice %}` to insert the survey right where you had it, and add a `{{ block.super }}` to keep the site notice logic still.

diox:
> Enable the waffle switch dev-exp-survey.

This is actually `dev_exp_survey` :sweat_smile:

diox (changes requested):
Looking good, just a small change regarding which template to add the block in to avoid extra queries on non-devhub pages

src/olympia/devhub/migrations/0008_auto_20250313_0001.py
```diff
@@ -0,0 +1,15 @@
...
+from olympia.core.db.migrations import CreateWaffleSwitch
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('devhub', '0007_surveyresponse_surveyresponse_unique_survey_user'),
+    ]
+
+    operations = [
+        CreateWaffleSwitch('dev_exp_survey')
```
eviljeff, line 14: Could you change this to a more expressive name?

src/olympia/devhub/templates/devhub/base.html
```diff
@@ -35,3 +35,17 @@
...
+    <div id="dev-survey-banner" response-url="{{ url('devhub.survey_response', survey_id=amo.DEV_EXP_SURVEY_ALCHEMER_ID) }}">
+      <div>
+        Help us make things better!
+        <a target="_blank" href="{{amo.SURVEY_LINK[amo.DEV_EXP_SURVEY_ALCHEMER_ID]}}" class="survey-link">Take our quick survey</a> 🦊
+      </div>
+      <a href="#" class="survey-dismiss">
+        ✕
+      </a>
+    </div>
+  {% endif %}
```
diox, line 50: You need a `{{ block.super }}` else you won't see the site notice/read only message if it's active. In fact, I'm surprised this is not breaking any tests ? Maybe we should have one checking that this (site notice/ready only message) is displayed correctly in devhub when active!

chrstinalin:
I split it off into its own block instead. I'm not sure why the super wasn't working correctly... It's having a very odd blend of behaviours.

image

diox (commented):
Almost there :sweat_smile:

diox (approved):

PR #23168, opened 2025-03-13
Fix broken tests following locales tweaks

diox:
Follow-up for https://github.com/mozilla/addons/issues/15433

### DIFF
--- a/src/olympia/amo/tests/test_utils.py
+++ b/src/olympia/amo/tests/test_utils.py
@@ -241,7 +241,7 @@ def test_get_locale_from_lang(lang):
     locale = get_locale_from_lang(lang)
 
     ignored_languages = ('cak',)
-    long_languages = ('ast', 'dsb', 'hsb', 'kab')
+    long_languages = ('ast', 'dsb', 'fur', 'hsb', 'kab')
     expected_language = (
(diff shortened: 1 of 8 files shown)


eviljeff (changes requested):
can we have the linting fix in this PR too?  (run `ruff format`)

diox:
With the way the code was previously written it would warn for locales already removed, which is why the tests were broken. My changes in `src/olympia/amo/management/commands/check_locales_completion_rate.py` fix that, and I've adjusted the tests + contents of the fixture file to make it more apparent.

src/olympia/amo/tests/test_utils.py
```diff
@@ -258,12 +258,6 @@ def test_get_locale_from_lang(lang):
         assert locale.territory == territory
 
 
-@pytest.mark.parametrize('lang', settings.LANGUAGES_BIDI)
-def test_bidi_language_in_amo_languages(lang):
-    """Make sure all bidi marked locales are in AMO_LANGUAGES too."""
```
diox, line 263: I'm removing this test because that's no longer true, some bidi locales are no longer enabled on the site

eviljeff (approved):

PR #23171, opened 2025-03-14
Add some reserved guids

diox:
Fixes: mozilla/addons#15423
Fixes: mozilla/addons#15424

There was some pushback against having a wildcard in there during triage meeting so I didn't go that far.

### DIFF
--- a/src/olympia/access/tests.py
+++ b/src/olympia/access/tests.py
@@ -299,6 +299,8 @@ system_guids = pytest.mark.parametrize(
         'flop@search.mozilla.org',
         'user@mozillaonline.com',
         'tester@MoZiLlAoNlInE.CoM',
+        'foo@siteperms.mozilla.org',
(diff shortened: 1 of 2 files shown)


peidrao (approved):

KevinMind (approved):
I'm not totally sure what I'm looking at. Can you add some additional context?

diox:
`RESERVED_ADDON_GUIDS` are guids (technically guid suffixes) that regular users can't use for their add-ons on AMO. Uploading a new add-on (*) with such a guid results in an error, unless you belong to a group with the specific `SystemAddon:Submit` permission.

This is done for various reasons, but the one we care about here is to avoid any potential conflicts with internal/synthetic add-ons that are never going to be submitted on AMO.

(*) new versions of an existing add-on are always permitted, to account for the fact that some of these guids used to be accepted in the past.

PR #23080, opened 2025-02-14
Add ci_completed.yml workflow to notify slack when something goes wrong.

KevinMind:
Fixes: mozilla/addons#15065

### Checklist

- [ ] Add `#ISSUENUM` at the top of your PR to an existing open issue in the mozilla/addons repository.
- [ ] Successfully verified the change locally.
- [ ] The change is covered by automated tests, or otherwise indicated why doing so is unnecessary/impossible.
- [ ] Add before and after screenshots (Only for changes that impact the UI).
- [ ] Add or update relevant [docs](../docs/) reflecting the changes made.

PR #23172, opened 2025-03-14
Update ci_completed.yml

KevinMind:
Follow up: https://github.com/mozilla/addons-server/pull/23080

### Description

We don't want to cancel concurrent workflows for `ci_completed.yml` each and every commit should update the status even if the triggering job was cancelled.

### Context

This workflow was cancelled because another ci_completed workflow was triggered. This workflow should have waited and then ran, but instead it was cancelled so we are missing a notification.

https://github.com/mozilla/addons-server/actions/runs/13854810472

### Testing

Very difficult to test but we should be able to see in the ci_completed queue that jobs are waiting and running 1 at a time, but are not cancelling. Only visible after merging.

### Checklist

### DIFF
--- a/.github/workflows/ci_completed.yml
+++ b/.github/workflows/ci_completed.yml
@@ -7,7 +7,8 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event_name}}-${{ github.ref}}
-  cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


eviljeff (approved):
The "fixes" link is missing in the description - can we at associate this PR with some issue/ticket

PR #23191, opened 2025-03-19
Check If Promoted Is Array in AddonSerializers

chrstinalin:
Fixes: mozilla/addons#15478

### Description

Corrects ESAddonSerializer assumption of a list (and makes it compatible with multiple promotions in the meantime, instead of maintaining the assumption of a single promotion)

### Checklist

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -1144,9 +1144,7 @@ class AddonSerializer(AMOModelSerializer):
             data.pop('is_featured', None)
         if request and is_gate_active(request, 'promoted-groups-shim'):
             promoted = data.pop('promoted', None)
-            data['promoted'] = (
-                promoted[0] if promoted else None
-            )
+            data['promoted'] = promoted[0] if promoted else None
         return data
 
     def get_promoted(self, obj):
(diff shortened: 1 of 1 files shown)


src/olympia/addons/serializers.py
```diff
@@ -1144,7 +1144,9 @@ def to_representation(self, obj):
             data.pop('is_featured', None)
         if request and is_gate_active(request, 'promoted-groups-shim'):
             promoted = data.pop('promoted', None)
-            data['promoted'] = promoted[0] if promoted else None
+            data['promoted'] = (
+                promoted[0] if promoted and isinstance(promoted, list) else promoted
+            )
```
diox, line 1149: Is this necessary ? isn't what you're doing in `fake_object()` enough ? (just asking, don't change it if you're not sure, I might be wrong)

chrstinalin: This is for the v3/v4 shim to return only the object (`src/olympia/addons/tests/test_serializers.py::test_promoted_shim`)

diox: Right, but why the change? When is `promoted` not going to be a list ? Doesn't `fake_object()` take care of that already no matter what data was in ES ? At this point we're in `to_representation()` so `data` is the output of the serializer.

diox (approved):

PR #23192, opened 2025-03-19
Use MinimalUserProfileSerializer for AccountViewSet for non-developers

eviljeff:
Fixes: mozilla/addons#15479 - follow up for mozilla/addons#15402, fixing a bug in https://github.com/mozilla/addons-server/pull/23154

### Description

See issue.  We were using the wrong serializer that didn't have the API_GATE integration to shim the fields.

### Testing

access any `/api/v[3|4|5]/accounts/account/(int:user_id)/` url without auth; see all the fields present, but with biography, etc (see original issue and pr) shim'd as `null` rather than omitted.

### Checklist

### DIFF
--- a/src/olympia/accounts/tests/test_views.py
+++ b/src/olympia/accounts/tests/test_views.py
@@ -1268,7 +1268,28 @@ class TestAccountViewSet(TestCase):
         response = self.client.get(self.url)
         assert response.status_code == 404
 
-    def test_is_not_full_public_profile_because_not_developer(self):
+    def test_is_not_full_public_profile_because_not_developer_but_fields_present(self):
+        # TODO: when mimimal-profile-has-all-fields-shim is removed for v5, we should
+        # change self.url to use v4 or v3
+        assert not self.user.has_full_profile
+        response = self.client.get(self.url)  # No auth.
+        assert response.data['name'] == self.user.name
+        assert response.data['biography'] is None
+        assert 'email' not in response.data
+        assert response.data['url'] == absolutify(self.user.get_url_path())
+
+        # Login as a random user and check it's still not visible.
+        self.client.login_api(user_factory())
+        response = self.client.get(self.url)
+        assert response.data['name'] == self.user.name
+        assert response.data['biography'] is None
+        assert 'email' not in response.data
+        assert response.data['url'] == absolutify(self.user.get_url_path())
+
+    @override_settings(DRF_API_GATES={'v5': ()})
+    def test_is_not_full_public_profile_because_not_developer_no_fields(self):
+        # TODO: when mimimal-profile-has-all-fields-shim is removed for v5, we won't
+        # need the override_settings
         assert not self.user.has_full_profile
         response = self.client.get(self.url)  # No auth.
         assert response.data['name'] == self.user.name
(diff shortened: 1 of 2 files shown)


src/olympia/accounts/tests/test_views.py
```diff
@@ -1268,7 +1268,28 @@ def test_view_deleted(self):
         response = self.client.get(self.url)
         assert response.status_code == 404
 
-    def test_is_not_full_public_profile_because_not_developer(self):
+    def test_is_not_full_public_profile_because_not_developer_but_fields_present(self):
```
KevinMind, line 1271: Nit: I would rather this be two different tests, but since we are trying to cherry-pick no need.

eviljeff: It is two tests:
- `test_is_not_full_public_profile_because_not_developer_but_fields_present` tests the case with the shim
- `test_is_not_full_public_profile_because_not_developer_no_fields` tests without the shim is, but as it's identical to the original `test_is_not_full_public_profile_because_not_developer` it doesn't show any changed lines

KevinMind: I mean the 2 sets of assertions in the first test.

KevinMind (approved):
URL: http://olympia.test/api/v3/accounts/account/1/

Response:

```json
{
  "id": 1,
  "name": "Test",
  "url": "http://olympia.test/en-US/firefox/user/1/",
  "username": "local_admin",
  "average_addon_rating": null,
  "created": "2025-03-03T11:06:15Z",
  "biography": null,
  "has_anonymous_display_name": false,
  "has_anonymous_username": false,
  "homepage": null,
  "is_addon_developer": true,
  "is_artist": true,
  "location": null,
  "occupation": null,
  "num_addons_listed": 0,
  "picture_type": null,
  "picture_url": null
}
```

URL: http://olympia.test/api/v4/accounts/account/1/

```json
{
  "id": 1,
  "name": "Test",
  "url": "http://olympia.test/en-US/firefox/user/1/",
  "username": "local_admin",
  "average_addon_rating": null,
  "created": "2025-03-03T11:06:15Z",
  "biography": null,
  "has_anonymous_display_name": false,
  "has_anonymous_username": false,
  "homepage": null,
  "is_addon_developer": true,
  "is_artist": true,
  "location": null,
  "occupation": null,
  "num_addons_listed": 0,
  "picture_type": null,
  "picture_url": null
}
```

> [!NOTE]
> only on v5 I got an error when using an invalid sessionID cookie 

URL: http://olympia.test/api/v5/accounts/account/1/

```json
{
  "id": 1,
  "name": "Test",
  "url": "http://olympia.test/en-US/firefox/user/1/",
  "username": "local_admin",
  "average_addon_rating": null,
  "created": "2025-03-03T11:06:15Z",
  "biography": null,
  "has_anonymous_display_name": false,
  "has_anonymous_username": false,
  "homepage": null,
  "is_addon_developer": true,
  "is_artist": true,
  "location": null,
  "occupation": null,
  "num_addons_listed": 0,
  "picture_type": null,
  "picture_url": null
}
```

eviljeff:
> only on v5 I got an error when using an invalid sessionID cookie

@KevinMind sounds like something that needs further investigation (but not related to this change)

PR #23193, opened 2025-03-19
add linter for workflows

KevinMind:
Fixes: mozilla/addons#15481

### Description

Use [actionlint](https://github.com/rhysd/actionlint/) to lint our github workflow files.

### Testing

Before: **failing** <https://github.com/mozilla/addons-server/actions/runs/13952752375/job/39056293411?pr=23193>
After:  **passing** <https://github.com/mozilla/addons-server/actions/runs/13952805622/job/39056471261?pr=23193>

### Checklist

### DIFF
--- a/.github/workflows/_test_main.yml
+++ b/.github/workflows/_test_main.yml
@@ -61,15 +61,22 @@ jobs:
         id: result
         shell: bash
         run: |
-          splits="${{ inputs.splits }}"
-          echo "splits: $splits"
-          echo "splits=$splits" >> "$GITHUB_OUTPUT"
-
           # Construct the matrix input for test_main using the groups count
-          # the matrix.group should be an array of numbers from 1 to $splits
-          matrix=$(seq -s, 1 "$splits")
-          echo "matrix: $matrix"
+          # the matrix.group should be an array of numbers from 0 to ${{ inputs.splits }}
+          splits="${{ inputs.splits }}"
+          # the matrix group should be indexed from 1 to splits + 1
+          n=$((splits + 1))
+          matrix=$(jq -n -c --arg n "$n" '[range(1;($n|tonumber))]')
+
+          # validate that matrix is an array of numbers from 1 to splits + 1
+          if ! echo "$matrix" | jq empty; then
+            echo "matrix is not an array of numbers from 1 to splits + 1"
+            exit 1
+          fi
+
+          echo "splits=$splits" >> "$GITHUB_OUTPUT"
           echo "matrix=$matrix" >> "$GITHUB_OUTPUT"
+          cat "$GITHUB_OUTPUT"
 
   test_main:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -43,6 +42,19 @@ jobs:
         id: context
         uses: ./.github/actions/context
 
+  actionlint:
```
diox, line 45: Can we explicitly drop permissions for this ? It's probably a good idea to do that for every job, but especially here since you're using an external docker image...

.github/workflows/ci.yml
```diff
@@ -43,6 +42,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run actionlint
+        run: |
+          docker run \
+            --rm \
+            -v "$(pwd):/workdir" \
+            -w "/workdir" \
+            rhysd/actionlint:latest \
```
diox, line 55: We should use a specific version w/ hash (and have it updated through dependabot later)

.github/workflows/ci.yml
KevinMind: I would do that in a separate issue. Could you file it, not a bad idea to default no permissions.

diox: Added that to https://github.com/mozilla/addons/issues/15312 and raised its priority

diox (approved):

Makefile-os
```diff
@@ -81,6 +81,15 @@ help_submake:
 	@echo "\nAll other commands will be passed through to the docker 'web' container make:"
 	@make -f Makefile-docker help_submake
 
+.PHONY: actionlint
+actionlint: ## lint github action workflows
+	docker run \
+		--rm \
+		-v "$(PWD):/workdir" \
+		-w "/workdir" \
+		rhysd/actionlint@sha256:887a259a5a534f3c4f36cb02dca341673c6089431057242cdc931e9f133147e9 \
```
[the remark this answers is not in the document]
KevinMind: Added this comment to the issue https://github.com/mozilla/addons/issues/15307

.github/workflows/_test_main.yml
```diff
@@ -61,15 +61,22 @@ jobs:
...
-          matrix=[$(seq -s, 1 $splits)]
-          echo "matrix: $matrix"
-          echo "matrix=$matrix" >> $GITHUB_OUTPUT
+          # the matrix.group should be an array of numbers from 0 to ${{ inputs.splits }}
+          splits="${{ inputs.splits }}"
+          # the matrix group should be indexed from 1 to splits + 1
+          n=$((splits + 1))
+          matrix=$(jq -n -c --arg n "$n" '[range(1;($n|tonumber))]')
+
+          # validate that matrix is an array of numbers from 1 to splits + 1
```
KevinMind, line 71: Turns out you need to validate json data that is used to build a matrix job, otherwise the job just doesn't run and silently fails.. 🙃

PR #23197, opened 2025-03-20
Refactor UsageTier to avoid exceptions being raised in the admin

diox:
This avoids failing when:
- growth_threshold_before_flagging is not set
- lower or upper adu threshold is not set

In addition, this adds to the admin page the number of add-ons belonging to the tier that would be flagged for number of abuse reports

Fixes https://github.com/mozilla/addons/issues/15480


src/olympia/abuse/tasks.py
```diff
@@ -43,31 +40,13 @@ def flag_high_abuse_reports_addons_according_to_review_tier():
...
-        .annotate(guid_abuse_reports_count=Count('*'))
-        .values('guid_abuse_reports_count')
-        .order_by()
-    )
     qs = (
-        Addon.unfiltered.exclude(status=amo.STATUS_DISABLED)
-        .filter(type=amo.ADDON_EXTENSION)
-        .annotate(abuse_reports_count=Subquery(abuse_reports_count_qs))
+        UsageTier.get_base_addons()
+        .alias(abuse_reports_count=UsageTier.get_abuse_count_subquery())
```
eviljeff, line 49: how does `alias` differ in functionality to `annotate` here?

src/olympia/reviewers/models.py
```diff
@@ -832,10 +834,12 @@ def get_base_addons(cls):
...
         """Return boundaries used to filter add-ons for that tier instance."""
-        return {
-            'average_daily_users__gte': self.lower_adu_threshold,
-            'average_daily_users__lt': self.upper_adu_threshold,
-        }
+        # We always have a lower bound, but not necessarily a higher one, e.g.
+        # notable doesn't have a ceiling.
+        filters = {'average_daily_users__gte': self.lower_adu_threshold or 0}
+        if self.upper_adu_threshold is not None:
+            filters['average_daily_users__lt'] = self.upper_adu_threshold
```
eviljeff, line 841: idea: would it make it easier to to some MAX_INT type value so it always exists?

eviljeff (approved):

src/olympia/abuse/tasks.py
diox: Super minor but `alias()` doesn't end up in the fields returned by the query. So for annotations that are only useful as something to `filter()` or `exclude()` but that you don't need the value later, it's slightly more efficient.

PR #23208, opened 2025-03-21
Replace bleach

diox:
Fixes: https://github.com/mozilla/addons/issues/15291

### Description

This replaces `bleach` with `nh3`, keeping similar functionality (cleaning text but also allowing us to linkify, and ensure outgoing links go through our outgoing proxy).

### Context

`bleach` is deprecated and needs to go. `nh3` is not a direct 1:1 replacement though, so some details are changing, particularly around what we keep when we encounter "bad" text. Overall functionality should remain the same though, with some refactoring to keep things more DRY.

### Testing

Unit tests, plus entering text in various fields when editing add-ons as a developer, and seeing that the result looks correct.


src/olympia/amo/tests/test_amo_utils.py
```diff
@@ -287,17 +287,14 @@ def test_escape_all(test_input, expected):
...
 
     assert escape_all('http://firefox.com') == (
-        '<a href="https://outgoing.firefox.com">http://firefox.com</a>'
+        '<a href="https://outgoing.firefox.com" rel="nofollow">http://firefox.com</a>'
     )
 
-    assert escape_all('firefox.com') == (
-        '<a href="https://outgoing.firefox.com">firefox.com</a>'
-    )
+    assert escape_all('firefox.com') == 'firefox.com'
```
diox, line 297: Minor behavior change: we no longer try to detect `<something>.<tld>` as an URL, `<scheme>://` needs to be present (and only `http`/`https`, as we're using `lxml_html_clean` defaults).

src/olympia/amo/tests/test_helpers.py
```diff
@@ -283,49 +284,97 @@ def test_external_url():
...
+            (('https://out.going.com/http%3A//foo.com', 'http://foo.com'),),
+        ),
+        (
+            'some http://foo.com <b>bold</b>',
+            'some http://foo.com bold',
+            (('https://out.going.com/http%3A//foo.com', 'http://foo.com'),),
+        ),
+        (
+            'some http://foo.com <script>alert(42)</script> thing',
+            'some http://foo.com thing',
```
diox, line 321: Behavior change: if you tried to run that same test through the old implementation, it would escape `<script>alert(42)</script>`. `nh3` completely removes it.

src/olympia/translations/tests/test_commands.py
```diff
@@ -71,12 +71,12 @@ def test_basic(self, index_addons_mock):
...
         )
 
-        # That one should not have changed.
+        # We didn't double-escape that one.
         self.translation_with_already_escaped_html.reload()
         assert (
             self.translation_with_already_escaped_html.localized_string
             == self.translation_with_already_escaped_html.localized_string_clean
-            == 'foo &lt;a href=&quot;http://bar.com&quot;&gt;http://bar.com&lt;/a&gt;'
+            == 'foo &lt;a href="http://bar.com"&gt;http://bar.com&lt;/a&gt;'
```
diox, line 79: Minor behavior change: we no longer escape quotes that don't need to be escaped because they are not inside quotes themselves. This is something to consider in general, our escaping here is not context-aware, and shouldn't be relied upon to display HTML attributes...

diox: `nh3` has `clean_text()` for when we truly don't want HTML, but it would double-escape in this scenario. I'll check whether we have some places in addons-server that could benefit from that.

src/olympia/amo/tests/test_helpers.py
diox: This is because of https://github.com/rust-ammonia/ammonia/issues/145 and I'm not super happy about it, might need to find a workaround, at the very least for markdown content.

src/olympia/translations/models.py
```diff
@@ -249,34 +243,25 @@ class PurifiedMarkdownTranslation(PurifiedTranslation):
...
-            tags=self.allowed_tags,
-            attributes=self.allowed_attributes,
-            filters=[self.linkify_filter],
-            strip=True,
-        )
+        html = md.markdown(text_with_brs, extensions=['abbr', 'fenced_code'])
 
-        return cleaner.clean(markdown)
+        # Run through cleaning as normal
+        return super().clean_string(html)
```
diox, line 258: This is causing a subtle breakage: We used to get text with `<p>` elements inserted by `md.markdown()`, only to remove them with `bleach` (with `strip=True`) and have them replaced with newlines, which got the string back to the original number of newlines.

We are now running through regular cleaning with `nh3` which strips but doesn't "restore" the newlines... we can work around that, maybe converting `\n` to `br` early and allowing `<br>` after markdown, but need to be careful.

diox:
Closing for now - not worth the time investment at the moment.

PR #23210, opened 2025-03-24
Deactivate jitter for Cinder tasks we're retrying

diox:
Fixes https://github.com/mozilla/addons/issues/15447

### Context

As mentioned in the issue, with `retry_jitter` enabled the delay is too unpredictable, could potentially be `0` which is not ideal. Let's stick to regular exponential backoff delay.

### Testing

See https://github.com/mozilla/addons-server/pull/23098, except the delay between each retry should always increase due to exponential backoff.

### DIFF
--- a/src/olympia/abuse/tasks.py
+++ b/src/olympia/abuse/tasks.py
@@ -80,12 +80,11 @@ def retryable_task(f):
 
     @task(
         autoretry_for=retryable_exceptions,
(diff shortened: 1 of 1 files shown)


KevinMind (approved):
I think we already have pretty good coverage of this, essentially asserting that the delay is greater than 0. Since this is solving a flaky problem might not hurt to re-run the offending job a few times to ensure it is actually gone. Up to you.

PR #23215, opened 2025-03-24
Make "Cancel and Disable Version" behave like Disable Version ; remove "Cancel Review Request"

diox:
### Context

"Cancel Review Request" is a relic from before auto-approval. Developers can just disable the version instead.

"Cancel and Disable Version", in the submission flow, is potentially useful but should behave like Disable Version instead, which is reversible and logs the activity.

### Testing

- Submit an add-on, choosing "On this site" as the distribution method. After uploading, during the submission process, hit "Cancel and Disable Version". 

Expected results:
- This should take you to the add-on version list page
- There should have been an activity log recorded (visible from "Recent changes" in the sidebar) for the version being disabled.
- The add-on should be shown as "Incomplete", it should not be public (`status` should be `0`).
- The version should have been disabled.
- Re-enabling the version should work, and let you resume the submission process, filling out details for the listing.
- Once you do the version should be awaiting review, and one approved the add-on should be public.

- You can repeat for an unlisted submission or a new version, everything should work in the same way except of course you won't have to fill out missing details for the listing and the add-on itself won't become public.

Fixes: mozilla/addons#15395

### DIFF
--- a/src/olympia/devhub/templates/devhub/versions/list.html
+++ b/src/olympia/devhub/templates/devhub/versions/list.html
@@ -97,7 +97,7 @@
          data-is-current="{{ (version == addon.current_version)|int }}">x</a>
     </td>
   </tr>
-  {% if full_info and version.channel == amo.CHANNEL_LISTED and (can_request_review or can_cancel) %}
+  {% if full_info and version.channel == amo.CHANNEL_LISTED and can_request_review %}
     <tr>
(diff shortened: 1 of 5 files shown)


eviljeff (approved):
I've this feeling at the back of my mind there's a scenario that's missed here, but I couldn't find one... so either we're all good or QA will find it(!) :shipit:

PR #23224, opened 2025-03-25
Add command to give "By Firefox" badge to themes with "Firefox" as author

diox:
Fixes https://github.com/mozilla/addons/issues/14947

### Context

Prod has some themes that were added to the "Firefox" user (they were initially built for Firefox and later released on AMO), and for consistency we want all of them to have the "By Firefox" badge.

### Testing

See unit test. Can also do locally:

- Make your "Firefox" user (`UserProfile` with id `settings.TASK_USER_ID`) an author on a public theme
- Run `manage.py promote_by_firefox_themes` command
- Your theme should now have a "By Firefox" badge.

### DIFF
--- a/src/olympia/promoted/management/commands/promote_by_firefox_themes.py
+++ b/src/olympia/promoted/management/commands/promote_by_firefox_themes.py
@@ -3,7 +3,7 @@ from django.core.management.base import BaseCommand
 
 from olympia import amo
 from olympia.addons.models import Addon
-from olympia.constants.promoted import LINE
+from olympia.constants.promoted import PROMOTED_GROUP_CHOICES
 from olympia.promoted.models import (
     PromotedAddon,
 )
@@ -24,6 +24,9 @@ class Command(BaseCommand):
             self.stdout.write(f'Promoting {addon.slug}')
             promotion, created = PromotedAddon.objects.get_or_create(
                 addon=addon,
-                defaults={'application_id': amo.FIREFOX.id, 'group_id': LINE.id},
+                defaults={
+                    'application_id': amo.FIREFOX.id,
+                    'group_id': PROMOTED_GROUP_CHOICES.LINE,
+                },
             )
             promotion.approve_for_addon()
--- a/src/olympia/promoted/management/commands/promote_by_firefox_themes.py
+++ b/src/olympia/promoted/management/commands/promote_by_firefox_themes.py
@@ -1,0 +1,30 @@
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+from olympia import amo
+from olympia.addons.models import Addon
+from olympia.constants.promoted import LINE
+from olympia.promoted.models import (
+    PromotedAddon,
+    PromotedApproval,
+)
+from olympia.users.models import UserProfile
+
+
+class Command(BaseCommand):
+    help = 'Give themes with Firefox as author the "By Firefox" badge'
+
+    def handle(self, *args, **options):
+        firefox_user = UserProfile.objects.get(pk=settings.TASK_USER_ID)
+        addons = (
+            Addon.objects.public()
+            .filter(type=amo.ADDON_STATICTHEME, authors=firefox_user)
+            .no_transforms()
+        )
+        for addon in addons:
+            self.stdout.write(f'Promoting {addon.slug}')
+            promotion, created = PromotedAddon.objects.get_or_create(
+                addon=addon,
+                defaults={'application_id': amo.FIREFOX.id, 'group_id': LINE.id},
+            )
+            promotion.approve_for_addon()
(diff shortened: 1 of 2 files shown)


src/olympia/promoted/management/commands/promote_by_firefox_themes.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def handle(self, *args, **options):
+        firefox_user = UserProfile.objects.get(pk=settings.TASK_USER_ID)
+        addons = (
+            Addon.objects.public()
+            .filter(type=amo.ADDON_STATICTHEME, authors=firefox_user)
+            .no_transforms()
+        )
+        for addon in addons:
+            self.stdout.write(f'Promoting {addon.slug}')
```
KevinMind, line 24: Nit: could wrap the log in the `created` to only show the newly promoted addons in the log. otherwise it will be a huge amount of noise.

src/olympia/promoted/management/commands/promote_by_firefox_themes.py
```diff
@@ -0,0 +1,29 @@
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+from olympia import amo
+from olympia.addons.models import Addon
+from olympia.constants.promoted import LINE
```
KevinMind, line 6: We are planning to deprecate the classes after we finish multiple promoted groups. Please use the chocies object instead. `PROMOTED_GROUPS_BY_ID` keeps the access centralized so it will be easier to clean up.

src/olympia/promoted/tests/test_commands.py
```diff
@@ -412,3 +412,38 @@ def disable_post_save_promoted_addon(self):
             sender=self.promoted_addon_class,
             dispatch_uid='addons.sync_promoted.promoted_addon_proxy',
         )
+
+
+class TestPromoteByFirefoxThemesCommand(TestCase):
+    def test_basic(self):
+        user = user_factory(pk=settings.TASK_USER_ID)
+        non_affected = [
+            # Not By Firefox user.
```
KevinMind, line 421: Nit: more explicit would be instead of comment, set the user and the type like you are setting the status.

src/olympia/promoted/tests/test_commands.py
```diff
@@ -412,3 +412,38 @@ def disable_post_save_promoted_addon(self):
             sender=self.promoted_addon_class,
             dispatch_uid='addons.sync_promoted.promoted_addon_proxy',
         )
+
+
+class TestPromoteByFirefoxThemesCommand(TestCase):
+    def test_basic(self):
+        user = user_factory(pk=settings.TASK_USER_ID)
```
KevinMind, line 419: nit: `firefox_user` or `task_user` more explicit

src/olympia/promoted/tests/test_commands.py
```diff
@@ -412,3 +412,38 @@ def disable_post_save_promoted_addon(self):
...
+            addon_factory(type=amo.ADDON_STATICTHEME),
+            # Not a theme.
+            addon_factory(users=[user]),
+            # Not public.
+            addon_factory(
+                type=amo.ADDON_STATICTHEME, status=amo.STATUS_DISABLED, users=[user]
+            ),
+        ]
+        # Already promoted, should not cause an error.
+        addon_factory(type=amo.ADDON_STATICTHEME, users=[user], promoted_id=LINE.id)
```
KevinMind, line 431: Could assert this instance is not changed? idk.

src/olympia/promoted/tests/test_commands.py
```diff
@@ -412,3 +412,38 @@ def disable_post_save_promoted_addon(self):
...
+
+        expected_affected = addon_factory(type=amo.ADDON_STATICTHEME, users=[user])
+
+        call_command('promote_by_firefox_themes')
+
+        for addon in non_affected:
```
KevinMind, line 437: We have already migrating reads from `PromotedAddon` to `PromotedAddonPromotion`, so to be consistent, query against `PromotedAddonPromotion` and `PromotedApproval` instead.

src/olympia/promoted/tests/test_commands.py
```diff
@@ -1,12 +1,12 @@
...
+from django.conf import settings
 from django.core.management import call_command
 from django.db.models.signals import post_save
-from django.test import TestCase
 
 from olympia import amo
-from olympia.amo.tests import addon_factory
-from olympia.constants.promoted import PROMOTED_GROUP_CHOICES
+from olympia.amo.tests import TestCase, addon_factory, user_factory
+from olympia.constants.promoted import LINE, PROMOTED_GROUP_CHOICES
```
KevinMind, line 9: Same as above, use the choices or by id wrappers.

KevinMind (commented):
Looks good, a few nits and consistency requests.

src/olympia/promoted/management/commands/promote_by_firefox_themes.py
diox: We should only run the command once, and none of these are promoted already, so it shouldn't matter - but more importantly, I want the output to show which add-ons we considered - if we noticed any one of them is missing the badge, we'll be able to go back to the logs to check it was part of the ones the command looked at.

KevinMind (approved):
🚢

PR #23225, opened 2025-03-25
Add short format to enable/disable version activity logs

diox:
Follow-up for https://github.com/mozilla/addons/issues/15395

### DIFF
--- a/src/olympia/activity/tests/test_constants.py
+++ b/src/olympia/activity/tests/test_constants.py
@@ -1,0 +1,52 @@
+import pytest
+
+from django.utils.translation import gettext_lazy as _
+
+from olympia.constants.activity import (
+    LOG,
+    LOG_HIDE_DEVELOPER,
+    LOG_OBSOLETE,
+    LOG_REVIEW_QUEUE_DEVELOPER,
+)
+
+
+def is_translated(value):
+    """Return True if `value` is a translated string. It simply looks at the
+    type: since translated strings for constants should be lazy, their type
+    should be a lazy proxy, not `str`."""
+    return not isinstance(value, str)
+
+
+def test_is_translated():
+    assert is_translated(_('Add-on'))
+    assert not is_translated('Add-on')
+
+
+@pytest.mark.parametrize(
+    'action',
+    [
+        action
+        for action in LOG
+        if action.id not in LOG_HIDE_DEVELOPER and action.id not in LOG_OBSOLETE
+    ],
+)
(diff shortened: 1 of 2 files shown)


eviljeff (commented):
Should these strings should be localized?  (I can't remember now where we expose `short` vs. the full format, but the other activities seem to be localized if `format` is)

diox:
mmm it should only be meant for reviewers, but technically it's potentially exposed through the API, so might as well translate them...

eviljeff:
Random idea: add a test that checks all the activity log classes not in `LOG_HIDE_DEVELOPER` (or in `LOG_REVIEW_QUEUE_DEVELOPER`?  We have too many constants and ways of showing/hiding) have localised strings - and not exposed aren't localised.

diox:
Yes, that's a good idea, I'll look into that I think that in theory:
- Every log *not* in LOG_HIDE_DEVELOPER needs a translated `format` (because those are exposed in devhub)
- Every log in LOG_REVIEW_QUEUE_DEVELOPER needs a translated `format` and a translated `short` (because technically those are exposed through the activity API)

diox:
Added https://github.com/mozilla/addons-server/pull/23225/commits/1414e353e9f4135c6c3ba3db10531074a21564ba which revealed those two actions couldn't show up in the activity API (not in `LOG_REVIEW_QUEUE_DEVELOPER`) and therefore don't need their `short` value to be translated.

src/olympia/activity/tests/test_constants.py
```diff
@@ -0,0 +1,52 @@
...
+    assert is_translated(_('Add-on'))
+    assert not is_translated('Add-on')
+
+
+@pytest.mark.parametrize(
+    'action',
+    [
+        action
+        for action in LOG
+        if action.id not in LOG_HIDE_DEVELOPER and action.id not in LOG_OBSOLETE
```
eviljeff, line 30: Should we be excluding obsolete activities from this check?  Historic activities could still appear in existing views exposed to users.

eviljeff (approved):
nit on LOG_OBSOLETE.  

(Potentially we should be doing the opposite too - enforcing in CI that we _don't_ localise strings that aren’t exposed. But that's a larger scope)

src/olympia/activity/tests/test_constants.py
diox: I know, it's not exactly great, but better than nothing. There are a number of existing obsolete ones that don't have translations, but I'm pretty sure have never been used. I wanted to at least ensure the ones we are currently using are translated.

PR #23228, opened 2025-03-26
Log when an admin deletes/undeletes a collection through the admin

diox:
Also avoid showing `deleted` field if the collection isn't already deleted - that's pointless and confusing, admins should use the regular delete action if they want to delete something.

Fixes https://github.com/mozilla/addons/issues/15394

### Testing
- As an admin (needs `Collections:Edit` permission + access to the admin, or just `*:*`) look at a non-deleted collection detail page in the admin
- `Deleted` field should be absent
- Delete that collection through the button
- Collection should now be deleted - can't see it in frontend
- An `ActivityLog` with `COLLECTION_DELETED` action and the collection as argument should have been recorded in the database
- Go back to that collection admin detail page
- `Slug` field should not have changed (it would if the user would have deleted the collection themselves through the frontend)
- `Deleted` field should be present and ticked
- Untick `Deleted` field and hit save
- An `ActivityLog` with `COLLECTION_UNDELETED` action and the collection as argument should have been recorded in the database
- Collection should now be undeleted - can see it in frontend

### DIFF
--- a/src/olympia/bandwagon/admin.py
+++ b/src/olympia/bandwagon/admin.py
@@ -62,7 +62,7 @@ class CollectionAdmin(AMOModelAdmin):
         # something not yet deleted the admin should use the regular button
         # instead of the checkbox.
         if obj and obj.deleted:
-            return fields + ('deleted', )
+            return fields + ('deleted',)
         return fields
 
     # Permission checks:
@@ -95,7 +95,6 @@ class CollectionAdmin(AMOModelAdmin):
         return acl.action_allowed_for(
             request.user, amo.permissions.ADMIN_CURATION
         ) or super().has_add_permission(request)
-
 
     def delete_model(self, request, obj):
         ActivityLog.objects.create(amo.LOG.COLLECTION_DELETED, obj)
--- a/src/olympia/bandwagon/admin.py
+++ b/src/olympia/bandwagon/admin.py
@@ -99,7 +99,7 @@ class CollectionAdmin(AMOModelAdmin):
 
     def delete_model(self, request, obj):
         ActivityLog.objects.create(amo.LOG.COLLECTION_DELETED, obj)
-        obj.delete()
+        obj.delete(clear_slug=False)
 
     def save_model(self, request, obj, form, change):
         super().save_model(request, obj, form, change)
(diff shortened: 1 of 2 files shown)


src/olympia/bandwagon/admin.py
```diff
@@ -87,5 +96,14 @@ def has_add_permission(self, request):
             request.user, amo.permissions.ADMIN_CURATION
         ) or super().has_add_permission(request)
 
+    def delete_model(self, request, obj):
```
KevinMind, line 99: Is there a reason we don't want a deleted log in all cases? Why only via admin?

src/olympia/bandwagon/admin.py
```diff
@@ -87,5 +96,14 @@ def has_add_permission(self, request):
...
+        ActivityLog.objects.create(amo.LOG.COLLECTION_DELETED, obj)
+        obj.delete(clear_slug=False)
+
+    def save_model(self, request, obj, form, change):
+        super().save_model(request, obj, form, change)
+        if change and 'deleted' in form.changed_data and not obj.deleted:
+            ActivityLog.objects.create(amo.LOG.COLLECTION_UNDELETED, obj)
```
KevinMind, line 106: Same question.

KevinMind (commented):
Looks good. I'll verify after you resolve my comments.

src/olympia/bandwagon/admin.py
diox: I don't think there is a strong reason, it's just something that never came up (where as deletion by moderators is something we need to keep track of for the transparency report - we already had the `LogEntry` django generates automatically for actions happening in the admin, but that's not as centralized as our `ActivityLog`).

src/olympia/bandwagon/admin.py
diox: Undeletion is not a feature exposed outside of the admin

src/olympia/bandwagon/admin.py
diox: FWIW ratings behave in the same way, the only activity log we emit is for moderators/reviewers deletion

KevinMind (approved):

PR #23235, opened 2025-03-28
backfill ReviewActionReason and CinderPolicy for expired rejections

eviljeff:
Fixes: mozilla/addons#15434

### Checklist

### DIFF
--- a/src/olympia/reviewers/management/commands/backfill_reviewactionreasons_for_delayed_rejections.py
+++ b/src/olympia/reviewers/management/commands/backfill_reviewactionreasons_for_delayed_rejections.py
@@ -1,0 +1,91 @@
+from django.core.management.base import BaseCommand
+
+import olympia.core.logger
+from olympia import amo
+from olympia.activity.models import ActivityLog, CinderPolicyLog, ReviewActionReasonLog
+
+
+log = olympia.core.logger.getLogger(
+    'z.reviewers.backfill_reviewactionreasons_for_delayed_rejections'
+)
+
+
+class Command(BaseCommand):
+    help = 'Backfill ReviewActionReasons in eventual delayed Reject'
+    expired_from_delayed_action_ids = {
+        amo.LOG.AUTO_REJECT_CONTENT_AFTER_DELAY_EXPIRED.id: (
+            amo.LOG.REJECT_CONTENT_DELAYED.id
+        ),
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/management/commands/backfill_reviewactionreasons_for_delayed_rejections.py
```diff
@@ -0,0 +1,101 @@
...
+            amo.LOG.REJECT_VERSION_DELAYED.id
+        ),
+    }
+
+    def handle(self, *args, **options):
+        rejections = ActivityLog.objects.filter(
+            # one of the expired rejection reasons
+            action__in=self.expired_from_delayed_action_ids.keys(),
+            # with no review action reasons currently
+            reviewactionreasonlog__id=None,
```
diox, line 31: I think we also want to limit to rejection made before the [fix](https://github.com/mozilla/addons/issues/15352) went live, and possibly limit to 2025 as well as mentioned in the issue.

The risk with this backfill is to accidentally include reasons from cancelled delayed rejections - IMHO it's acceptable if we limit to the relatively narrow period, but slightly more problematic if we start doing it on a larger number of rejections.

eviljeff: But sure, I can add additional guardrails

diox: I saw that it would only take the most recent, but it wouldn't help all scenarios, so limiting to the specific (relatively short, first couple months of 2025 essentially) time period would be valuable IMHO

eviljeff: What scenario is missing?  I can add to it if it's worthwhile?

diox: The problem is that technically it's possible to issue multiple delayed rejections in a row for the same version, with or without cancelling earlier ones. With the way you've written the backfill command, it technically leads to missing the earlier rejections... which in turn prevents incorrect data being backfilled if the earlier ones were cancelled. But it looks like `auto_reject` doesn't care and does the opposite, happily recording them all... it's hard to choose which one is correct - I vote for keeping your approach, but with the time period filtering so that any potential damage is limited to a few instances.

PR #23236, opened 2025-03-28
Add validation for DOCKER_* variables:

KevinMind:
Fixes: mozilla/addons#15504

## Description

- add explicit make build command
- Refactored setup.py to improve metadata handling for Docker images, including validation of environment variables.

## Context

This PR solves the issue found in the bug in 2 separate and compatiible ways.

###  combined `build` command

Before we had to call `make setup` before calling `make docker_build_web` that is because the build command relied on values set in the .env file, but did not actually create the .env so it would either a) not exist or b) be whatever it happened to be before.

That could lead to very bizzare and unexpected results (like this bug) where we passed a value that is **relevant** for the build but not saved in the .env and therefore not available to the build command context.

A dedicated `make build` command that depends on setup will bass whatever environment variables and arguments you pass to both setup and docker build commands.

### validate setup .env results

We have always had implicit validation of the values created by setup.py but that is not really good enough as this bug was allowed to exist. In some cases we would (silently) override values passed to the command. That's not great because it might fix the issue but doesn't give the user feedback In other cases we just let invalid values go unnoticed.

Now we have explicit validation of every value set in the .env file ensuring it is impossible to run the project with an invalid setup config.

### 1) Run make setup

```bash
make setup
```

Here are some sample test scenarios, all covered by automated testing.

> [!NOTE]
> It is assumed that you `rm .env` in between each invocation. Some values are preserved and re-read from the .env file DOCKER_TARGET and DOCKER_TAG 

```
make setup DOCKER_TAG='' -> works because tag can be inferred from default version
make setup DOCKER_TAG='' DOCKER_VERSION='' -> still works because version defaults to local
make setup DOCKER_VERSION='' DOCKER_DIGEST='' -> surprisingly still works because version defaults to local
make setup DOCKER_VERSION='local' DOCKER_DIGEST='abc' -> also works.. expect digest to take precedence

make setup DOCKER_TARGET=development -> works because the default verssion is local
make setup DOCKER_TARGET=development DOCKER_VERSION=latest -> fails because remote images are always production
make setup DOCKER_TARGET=production DOCKER_VERSION=latest -> works because we use production for remote image
make setup DOCKER_TARGET=production DOCKER_DIGEST=abc -> works because we use production for remote image
make setup DOCKER_TARGET=production DOCKER_DIGEST=abc DOCKER_VERSION=local -> works because we use production for remote image and digest takes precedence

make setup DOCKER_TARGET=production DOCKER_COMMIT=abc -> works because we do not care about commit on local images and the default version is local
make setup DOCKER_TARGET=production DOCKER_BUILD=abc -> works because we do not care about build on local images and the default version is local

make setup DOCKER_TARGET=production DOCKER_COMMIT=abc DOCKER_VERSION=latest -> fails because we cannot override commit when running production images
make setup DOCKER_TARGET=production DOCKER_BUILD=abc DOCKER_VERSION=latest -> fails because we cannot override build when running production images

make setup DOCKER_TARGET=development DOCKER_COMMIT=abc -> works because we dont care about commit on local images
make setup DOCKER_TARGET=development DOCKER_BUILD=abc -> works because we dont care about build on local images

make setup SETUP_ARGS="--build" DOCKER_TARGET=production DOCKER_COMMIT="" -> works because we do not care about commit on local images
make setup SETUP_ARGS="--build" DOCKER_TARGET=production DOCKER_VERSION=latest DOCKER_COMMIT="abc" -> fails because we must set a build when building remote images
make setup SETUP_ARGS="--build" DOCKER_TARGET=production DOCKER_VERSION=latest DOCKER_COMMIT="abc" -> fails because we must set a build when building remote images
make setup SETUP_ARGS="--build" DOCKER_TARGET=production DOCKER_VERSION=latest DOCKER_COMMIT="abc" DOCKER_BUILD="build" -> works because we set a build and commit when building a remote image

make setup SETUP_ARGS="--build" DOCKER_TARGET=production DOCKER_VERSION=latest DOCKER_COMMIT="abc" DOCKER_BUILD="build"

make setup SETUP_ARGS="--build" DOCKER_TARGET=development -> fails because we cannot build a development image
```

### 2) Run make build

```bash
make build
```

Check the endpoint

```bash
curl http://olympia.test/__version__
```
Should contain content matching the `/build-info.json` file. Run make up with different values and it should reflect the updates

Will run `setup` first so same rules apply and then will run the build.

## Checklist

### DIFF
--- a/Makefile-os
+++ b/Makefile-os
@@ -97,7 +97,12 @@ test_setup:
 .PHONY: setup
 setup: ## create configuration files version.json and .env required to run this project
 	for path in $(CLEAN_PATHS); do rm -rf "$(PWD)/$$path" && echo "$$path removed"; done
-	./scripts/setup.py
+	./scripts/setup.py $(SETUP_ARGS)
+
+.PHONY: build
+build: ## build the docker image
+	$(MAKE) setup SETUP_ARGS="--build"
+	$(MAKE) docker_build_web
 
 .PHONY: push_locales
 push_locales: ## extracts and merges translation strings
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -56,16 +56,15 @@ runs:
       id: build
       shell: bash
       env:
+        # These values are irrelevant for local builds, so don't include them
+        # in the make command directly but set them on the environment to be picked up.
         DOCKER_TAGS_FILE: ${{ steps.meta.outputs.bake-file-tags }}
         DOCKER_ANNOTATIONS_FILE: ${{ steps.meta.outputs.bake-file-annotations }}
         DOCKER_METADATA_FILE: ${{ steps.context.outputs.metadata_file }}
         DOCKER_PUSH: ${{ inputs.push }}
       run: |
-        make setup \
-          DOCKER_TARGET="production" \
-          DOCKER_VERSION="${{ steps.meta.outputs.version }}"
-
-        make docker_build_web \
+        make build \
+          DOCKER_VERSION="${{ steps.meta.outputs.version }}" \
           DOCKER_COMMIT="${{ steps.context.outputs.git_sha }}" \
           DOCKER_BUILD="${{ steps.context.outputs.git_build_url }}"
 
--- a/.github/workflows/_test_check.yml
+++ b/.github/workflows/_test_check.yml
@@ -61,16 +61,9 @@ jobs:
           - production
     steps:
       - uses: actions/checkout@v4
-      - shell: bash
-        continue-on-error: true
-        run: |
-          cat <<EOF
-            Values passed to the action:
-            version: ${{ matrix.version }}
-            target: ${{ matrix.target }}
-            deps: ${{ matrix.deps }}
-          EOF
       - name: ${{ matrix.version == 'local' && 'Uncached Build' || 'Pull' }} Check
+        id: build
+        continue-on-error: true
         uses: ./.github/actions/run-docker
         # Set environment variables that are expected to be ignored
         env:
@@ -81,9 +74,22 @@ jobs:
           target: ${{ matrix.target }}
           deps: ${{ matrix.deps }}
           run: make check
+
+      - name: Verify build
+        id: verify
+        shell: bash
+        run: |
+          version="${{ matrix.version }}"
+          target="${{ matrix.target }}"
+          outcome="${{ steps.build.outcome }}"
+          if [[ "$version" != "local" && "$target" == "development" && "$outcome" == "success" ]]; then
+            echo "Non local images must run in production mode"
+            exit 1
+          fi
+
       - name: Cached Build Check
         uses: ./.github/actions/run-docker
-        if: ${{ matrix.version == 'local' }}
+        if: matrix.version == 'local' && steps.verify.outcome == 'success'
         with:
           version: ${{ matrix.version }}
           target: ${{ matrix.target }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,7 +96,6 @@ jobs:
           digest: ${{ needs.build.outputs.digest }}
           version: ${{ needs.build.outputs.version }}
           deps: development
-          target: development
           run: |
             make docs
 
--- a/.github/workflows/health_check.yml
+++ b/.github/workflows/health_check.yml
@@ -29,7 +29,6 @@ jobs:
         continue-on-error: true
         uses: ./.github/actions/run-docker
         with:
-          target: development
           version: local
           run: |
             # On local, we want to ensure there are failures to test
--- a/docs/topics/development/setup_and_configuration.md
+++ b/docs/topics/development/setup_and_configuration.md
@@ -118,6 +118,9 @@ make up DOCKER_VERSION=<version>
 
 Version is the published tag of addons-server and corresponds to `mozilla/addons-server:<version>` in [dockerhub][addons-server-tags].
 
+> **Important**: When using a remote image (via `DOCKER_VERSION` or `DOCKER_DIGEST`), the `DOCKER_TARGET` must be set to 'production'.
+> Running remote images in development mode is not supported and will fail validation during setup.
+
 Specifying a version will configure docker compose to set the [pull policy] to _always_ and specify the _image_ property
 in the docker compose config to pull the latest build of the specified `version`. Once you've specified a version,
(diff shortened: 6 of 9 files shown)


Makefile-os
```diff
@@ -39,7 +39,7 @@ DOCKER_BAKE_ARGS := \
 	--progress $(DOCKER_PROGRESS) \
 	--metadata-file $(DOCKER_METADATA_FILE) \
 
-ifeq ($(DOCKER_PUSH), true)
```
KevinMind, line 42: This is irrelevant to the main scope but other checks work with `ifneq` which will accept **any** value as true.

DOCKER_PUSH=1
DOCKER_PUSH=true
DOCKER_PUSH=True
DOCKER_PUSH=False

All mean the same thing (we want to push)

DOCKER_PUSH=

OR simply not adding the flag, means don't push

KevinMind:
@eviljeff assuming CI passes, this should be ready to review.

eviljeff:
Can you suggest some valid values for the testing too?  https://mozilla.github.io/addons-server/topics/development/building_and_running_services.html#build-process implies DOCKER_COMMIT and DOCKER_BUILD would both have automatic values if not present but now they have to be?

eviljeff:
I don't see any `build-info.json` file either, but I'm assuming that's only present for production builds, or something?

KevinMind:
> I don't see any `build-info.json` file either, but I'm assuming that's only present for production builds, or something?

`build-info.json` is on every image, produced at build time and is located in the root.  so it won't be visible from your host. that is by design as the values there should NOT be overridable when running an image, but only building.

After answering this question and your other for valid config inputs, it dawned on me that if DOCKER_VERSION!=local we should not allow you to pass any values for `DOCKER_TARGET`, `DOCKER_COMMIT` or `DOCKER_VERSION` as they should be defined already in the build.

I need to update the validation logic based on that. 

```bash
make up DOCKER_VERSION=latest DOCKER_COMMIT=foo
``

Is actually invalid.. you shouldn't be able to just **set** a commit. it is based on the commit that was used to build the version of `latest` you are pulling.

Back to draft.

**Comment by KevinMind** [comment]
And we're back. I'm not changing anything because CI is green and it does what I wanted it to do.

**Comment by eviljeff** [comment]
> Can you suggest some valid values for the testing too? https://mozilla.github.io/addons-server/topics/development/building_and_running_services.html#build-process implies DOCKER_COMMIT and DOCKER_BUILD would both have automatic values if not present but now they have to be?

I'm back at this point - the testing instructions seems to just have DOCKER_COMMIT as "abc" - is that valid?

> Should contain content matching the /build-info.json file.

How can I verify this if I can't see the /build-info.json file?

**Comment by KevinMind** [comment]
> > Can you suggest some valid values for the testing too? https://mozilla.github.io/addons-server/topics/development/building_and_running_services.html#build-process implies DOCKER_COMMIT and DOCKER_BUILD would both have automatic values if not present but now they have to be?
> 
> I'm back at this point - the testing instructions seems to just have DOCKER_COMMIT as "abc" - is that valid?
> 
> > Should contain content matching the /build-info.json file.
> 
> How can I verify this if I can't see the /build-info.json file?

Valid/Invalid examples are included in the test description there are over a dozen. The validation rule for DOCKER_COMMIT is that it must exist in some scenarios there is no deeper validation (overkill) You can verify the build-info.json by:
- `cat /build-info.json` in the container
- `/__version__` endpoint which reads from that file.

**Comment by eviljeff** [comment]
- `make setup SETUP_ARGS="--build" DOCKER_TARGET=production DOCKER_VERSION=latest DOCKER_COMMIT="abc" DOCKER_BUILD="build"`
- `make up`
- curl http://olympia.test/__version__
```{"target": "production", "version": "", "source": "https://github.com/mozilla/addons-server", "commit": "130f132d577baea4331ccaad5367f90a8f528c89", "build": "https://github.com/mozilla/addons-server/actions/runs/14216750902", "python": "3.12", "django": "4.2", "addons-linter": "7.9.0"}```

- cat /build-info.json 
```{  "commit": "130f132d577baea4331ccaad5367f90a8f528c89",  "version": "",   "build": "https://github.com/mozilla/addons-server/actions/runs/14216750902",   "target": "production",   "source": "https://github.com/mozilla/addons-server" } ```

**Review by eviljeff — APPROVED** [review]

**Comment by KevinMind** [comment]
> * `make setup SETUP_ARGS="--build" DOCKER_TARGET=production DOCKER_VERSION=latest DOCKER_COMMIT="abc" DOCKER_BUILD="build"`
> 
>     * `make up`
> 
>     * curl http://olympia.test/__version__
>       `{"target": "production", "version": "", "source": "https://github.com/mozilla/addons-server", "commit": "130f132d577baea4331ccaad5367f90a8f528c89", "build": "https://github.com/mozilla/addons-server/actions/runs/14216750902", "python": "3.12", "django": "4.2", "addons-linter": "7.9.0"}`
> 
>     * cat /build-info.json
>       `{  "commit": "130f132d577baea4331ccaad5367f90a8f528c89",  "version": "",   "build": "https://github.com/mozilla/addons-server/actions/runs/14216750902",   "target": "production",   "source": "https://github.com/mozilla/addons-server" } `

When you run the setup command the output is:

```
DOCKER_TAG=mozilla/addons-server:latest
DOCKER_VERSION=latest
DOCKER_TARGET=production
DOCKER_COMMIT=abc
DOCKER_BUILD=build
HOST_UID=501
DEBUG=False
OLYMPIA_DEPS=production
```

When you run `make up`, the environment is:

```
Environment:
DOCKER_TAG=mozilla/addons-server:latest
DOCKER_VERSION=latest
DOCKER_TARGET=production
HOST_UID=501
DEBUG=False
OLYMPIA_DEPS=production
```

> [!IMPORTANT]
> DOCKER_COMMIT is no longer there 

In this case, the first command would have set you up to build an image with tag `latest` and docker commit `abc`, but you didn't build, you up-ed and up is pulling the latest image instead which has the commit you then saw in the build-info.json

I know that's probably confusing but that's why I added the `build` command so you can pass the arguments directly when building and not get weird instances of `setup` (like this one) that then get overriden by another command that reruns setup under the hood.

eviljeff:
yes, it was very confusing - it's really not clear what `make build` does specifically when the `make setup` examples include `--build`, implying it also builds.

If `make up` was _not_ supposed to be run it would have been good to call that out, because it's the standard way of starting the project - I wouldn't have expected `olympia.test/__version__` to work at all if the `web` container wasn't running (and I'm still not sure how it was expected to work)

PR #23237, opened 2025-03-28
Use primary db in promote_by_firefox_themes and promoted models sync in general

diox:
Follow-up fix (hopefully) for https://github.com/mozilla/addons/issues/14947

(We ran the command on dev and it created incomplete models, likely because some reads where happening on the replica)


KevinMind (approved):
No real downside here right?

PR #23256, opened 2025-04-02
Make DB introspective regarding PII

KevinMind:
Fixes: mozilla/addons#15430

### Description

Adds a monkey patch for the django base filed class adding a "pii" argument and a "is_pii" property, allowing the entire database to be introspective with reference to pii information.

### Context

This enables much more fine grained control over data retention, data protection and all kinds of goodies.

We can parse the entire database schema and know which models contain pii and what PII that is.

This could allow for more predictable and declarative "cleaning" of PII during soft deletion and better protection of data in a number of different cases including but not limited to:
- logging
- error handling
- data retention

This is the "simplest" way to do this. We could instead subclass any fields requiring PII but this would be:
- a lot more work
- harder to gurantee correct usage as all fields would have to come from a custom location and not django itself

The risks associated with monkey patching here are quite low because we do it at the root app on the module scope. If anything were to go wrong, it would be immediate. Additionally we are only dealing with instance level logic, no changes to DB behaviour or featureset.

### Test PII

You can add the `pii=True` to any number of fileds.

Run the management command:

```bash
./manage.py pii
```

Expect to get a 2d matrix of models and the fields that have pii set. Expect all the fields you updated to be listed.

You should be able to do anything you normally would do with those models.

### Migrate

Run makemigrations for the app(s) you touched. Expect a migration created adding PII to those fields. This way we can track what is or isn't PII via migrations.

### Data index

Screenshot 2025-04-02 at 13 04 41

### PII

Screenshot 2025-04-02 at 13 04 51

## Checklist


diox:
We should avoid monkeypatching django fields IMHO. This is the sort of things that breaks easily and is hard to maintain. We can achieve a similar result by just making our own `SensitiveCharField` and `SensitiveTextField` classes that inherit from Django's.

KevinMind:
> We should avoid monkeypatching django fields IMHO. This is the sort of things that breaks easily and is hard to maintain. We can achieve a similar result by just making our own `SensitiveCharField` and `SensitiveTextField` classes that inherit from Django's.

@diox I generally agree but I think in this case subclassing fields would either:
- require us to subclass every field
- not be able to introspect consistently

Let's say there is a `CharField` that is **NOT** using the `SensitiveCharField` class, if you parse the schema, it would show up as not sensitive, but it is... even if you add the argument. You don't know if a false value is due to not setting the argument or not using the right class.

It's true that even if you provide the argument globally it's still up to individual field declarations to provide the argument, but at least this way you don't have to subclass every field type just to add a single static property.

I think the benefits here outweigh the risks. We are not overriding **ANY** existing behavior of the base field class.

One _other_ Idea I had was to use a name prefix for sensitive fields. if a field is sensitive it would get named `sensitive_<name` then we could introspect based on the name. This has some pitfalls of it's own. Would require large migrations, longer field names, what about typos, etc. It might not be a terrible alternative though because would avoid monkey patching and it really throws in your face that a piece of data is sensitive because it's literally in the name. wdyt?

diox:
Having to subclass field types where you want to have the argument is still better than monkeypatching base fields from Django IMHO.

Another approach would be to simply have a list of fields on the model (or the model's Meta class). We already have similar logic in place, for a different purpose, with `ANONYMIZED_FIELDS` on `UserProfile`.

diox:
> But how would we avoid false negative cases without making the argument available on all fields?

The only way to completely avoid it would be to make it required for all fields, which I think is way too invasive anyway. I don't think that should be part of our goals.

KevinMind:
@diox thoughts? This requires no monkey patching and makes the property abstracted from field definitions.. it has nothing to do with the particulars of the field anyhow.

There are a bunch of migrations here because I wanted to see what happens when you run make migrations before and after the change.

btw.. we have a lot of unmigrated changes to models.. we should probably start checking for unmigrated code chage in CI

diox:
> btw.. we have a lot of unmigrated changes to models.. we should probably start checking for unmigrated code chage in CI

We often do that on purpose to not ship migrations altering the db until the field is gone from the model, so we can't do a simple check in CI.

diox:
Using `Meta` sounds fine. Couple thoughts:
- There are probably models defining their own custom Meta that does not inherit from the `ModelBase` one.
- You don't need to stick that in `ModelBase`, it could be a separate mixin that `ModelBase` inherits from, so that you don't have to add `ModelBase` as the parent for a bunch of models that don't currently inherit from it.

PR #23261, opened 2025-04-03
Bump check locales completion rate to 70%

diox:
Fixes https://github.com/mozilla/addons/issues/15441

### Testing

We're only changing the threshold value, so not much to verify. You can run the command locally and check the logs and `FakeEmail` to see if we correctly considered locales below/above the threshold, but unit tests already do that.

### DIFF
--- a/src/olympia/amo/tests/test_commands.py
+++ b/src/olympia/amo/tests/test_commands.py
@@ -781,12 +781,19 @@ class TestCheckLocalesCompletionRate(TestCase):
 
     def _test_full_run_typical_response(self):
         expected_below = (
-            'The following locales are below threshold of 40% or completely '
+            'The following locales are below threshold of 70% or completely '
             'absent in one of our projects in Pontoon:\n- '
             + '\n- '.join(
                 (
+                    'Bulgarian [bg]',
+                    'Danish [da]',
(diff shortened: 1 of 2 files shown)


src/olympia/amo/tests/test_commands.py
```diff
@@ -781,12 +781,19 @@ def test_full_run_typical_response(self):
 
     def _test_full_run_typical_response(self):
         expected_below = (
-            'The following locales are below threshold of 40% or completely '
+            'The following locales are below threshold of 70% or completely '
             'absent in one of our projects in Pontoon:\n- '
             + '\n- '.join(
                 (
+                    'Bulgarian [bg]',
```
KevinMind, line 788: Is this testing the actual completion rate based on the current locales? Could that be mocked so we wouldn't need to change it when local ratios change?

KevinMind (approved):
One nit about the test but all good.

src/olympia/amo/tests/test_commands.py
diox: It's using mocked data instead of the actual source, which is pontoon API. Because that mocked data hasn't changed, the test had to be updated to match.

PR #23267, opened 2025-04-03
Limit height of notes for reviewers and replies in review page, add scroll

diox:
Fixes: mozilla/addons#15502

### Testing:

- As a developer, change "Notes for reviewers" for a version to something super long
- As a reviewer, open the review page for that add-on

The height of the text in the version history should have a limit, with a scrollbar available to see the rest. If there is some text to show after the scrollbar at the bottom (or at the top if you started scrolling) a subtle background gradient should show up to indicate that.

Repeat for developers and reviewer replies.

### Screenshot

Screenshot 2025-04-03 at 14-45-39 seafowala-chanter-girroach – Add-ons for Firefox

### DIFF
--- a/static/css/zamboni/reviewers.less
+++ b/static/css/zamboni/reviewers.less
@@ -1142,14 +1142,6 @@ pre.history-comment {
         linear-gradient(rgba(255,255,255,0), white 70%) 0 100%,
 
         /* Shadows */
-        radial-gradient(50% 0, farthest-side, rgba(0,0,0,.2), rgba(0,0,0,0)),
-        radial-gradient(50% 100%,farthest-side, rgba(0,0,0,.2), rgba(0,0,0,0)) 0 100%;
-    background:
-        /* Shadow covers */
-        linear-gradient(white 30%, rgba(255,255,255,0)),
-        linear-gradient(rgba(255,255,255,0), white 70%) 0 100%,
-
-        /* Shadows */
         radial-gradient(farthest-side at 50% 0, rgba(0,0,0,.2), rgba(0,0,0,0)),
         radial-gradient(farthest-side at 50% 100%, rgba(0,0,0,.2), rgba(0,0,0,0)) 0 100%;
     background-repeat: no-repeat;
--- a/static/css/zamboni/reviewers.less
+++ b/static/css/zamboni/reviewers.less
@@ -1129,9 +1129,33 @@ pre.history-comment {
     font-style: italic;
 }
 
-.history-notes {
-    max-width: 370px;
+.history-notes, .history-comment {
     overflow: auto;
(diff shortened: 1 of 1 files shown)


static/css/zamboni/reviewers.less
```diff
@@ -1129,9 +1129,33 @@ pre.history-comment {
     font-style: italic;
 }
 
-.history-notes {
-    max-width: 370px;
```
eviljeff, line 1133: I'm assuming this removal was deliberate, but noting because it's width and the PR is about height

static/css/zamboni/reviewers.less
```diff
@@ -1129,9 +1129,33 @@ pre.history-comment {
...
+        radial-gradient(50% 0, farthest-side, rgba(0,0,0,.2), rgba(0,0,0,0)),
+        radial-gradient(50% 100%,farthest-side, rgba(0,0,0,.2), rgba(0,0,0,0)) 0 100%;
+    background:
+        /* Shadow covers */
+        linear-gradient(white 30%, rgba(255,255,255,0)),
+        linear-gradient(rgba(255,255,255,0), white 70%) 0 100%,
+
+        /* Shadows */
+        radial-gradient(farthest-side at 50% 0, rgba(0,0,0,.2), rgba(0,0,0,0)),
+        radial-gradient(farthest-side at 50% 100%, rgba(0,0,0,.2), rgba(0,0,0,0)) 0 100%;
```
eviljeff, line 1154: duplicated rules?

eviljeff (approved):

static/css/zamboni/reviewers.less
diox: It's to be consistent between the way we display "approval notes" and comments

static/css/zamboni/reviewers.less
diox: It's not strictly duplicated, there are subtle differences. The first background is for compatibility with older syntax (it's ignored by browsers supporting the newer syntax, the second rule overwrites it). I don't think we care about backwards-compatibility here though, so I'll remove it.

eviljeff: ah, the first two lines (`linear-gradient`) are identical in each (right?), but I overlooked the difference in `radial-gradient'

diox: Yes, only the `radial-gradient` syntax differs (the second one using the more modern `at` version... well, more modern in 2012 when the article was written), but the entire `background` needs to be repeated for the compatibility trick to work.

I've removed the old version in https://github.com/mozilla/addons-server/pull/23267/commits/a63ea49d33ddccc250aadd249965ec0eb7e4f5f7

PR #23268, opened 2025-04-03
Allow Multiple Promoted Groups

chrstinalin:
Fixes: mozilla/addons#15511

### Description

Enables multiple promoted groups to be assigned to an add-on.

### Testing

1. Upload a new add-on.
2. Create or delete a `PromotedAddonPromotion` for multiple different promoted groups / applications via the admin panel for that add-on in its `DiscoveryAddon`.
3. a) Approve the groups for the add-on via the reviewer tools, under 'Approve', OR
b) Run the auto-approval script, and confirm auto-approval.
4. Banner should be visible on reviewer tools, listing every promoted type.
5. Only one badge should be visible at a time on `addons-frontend`, based on priority (See https://github.com/mozilla/addons-frontend/pull/13343)

image

### Checklist


src/olympia/addons/tests/test_models.py
```diff
@@ -1689,6 +1695,79 @@ def test_promoted_groups(self):
         assert not addon.promoted_groups()
         assert addon.promoted_groups(currently_approved=False)
 
+    def test_multiple_promoted_groups(self):
```
KevinMind, line 1698: This should be split into several tests. Long tests that test multiple behaviors like this have a few issues:
- they take a long time to read and understand, really easy to miss something
- they give poor signal when something fails. Assume the first assertion fails, you lose all info about the rest of the test. A lot less granular than multiple specific tests
- they are hard to maintain. Mostly because of the first point, but also because you have a lot of sequentially ordered logic and when the code changes it may or may not continue making sense in the order you have. Isolated tests are harder to break because they are sensitive to smaller parts of the code base.

src/olympia/amo/tests/__init__.py
```diff
@@ -589,19 +589,25 @@ def change_channel_for_addon(self, addon, listed):
             version.update(channel=channel)
 
     @classmethod
-    def make_addon_promoted(cls, addon, group_id, approve_version=False, apps=None):
+    def make_addon_promoted(
```
KevinMind, line 592: I think the main reason you need the `reset` argument is because of the large tests.. right? if they were smaller, you wouldn't need this. This abstraction is probably not worth it IMO.
- It doesn't save much as you can delete all promotions in a single line anyway and now you have to go back here to understand what it does.
- True by default means calls to `make_addon_promoted` implicitly delete all other promotions (very un-intuitive)

src/olympia/promoted/tests/test_models.py
```diff
@@ -615,16 +615,6 @@ def _test_unique_constraint(self, fields, should_raise=False):
 
         assert PromotedAddonPromotion.objects.create(**merged) is not None
 
-    def test_multiple_promoted_groups_per_addon_raises(self):
```
KevinMind, line 618: This should now test that `multiple_promoted_groups_per_addon_application_raises`

KevinMind (commented):
So far so good. Have a look at my comments.

src/olympia/amo/tests/__init__.py
chrstinalin: > I think the main reason you need the reset argument is because of the large tests.. right? 

No, its because of the existing tests that use `make_addon_promoted()` and are designed only to expect one promotion at a time. So if I were to do this:

> It doesn't save much as you can delete all promotions in a single line anyway and now you have to go back here to understand what it does.

...I'd have to touch a lot more test files.

src/olympia/promoted/tests/test_models.py
chrstinalin: Is that not tested by `self._test_unique_constraint({}, should_raise=True)`?

eviljeff: Personally, I'd rewrite these tests - `should_raise` is only True in one case now, so the content of _test_unique_constraint should be directly in that test.  And then the other two tests, where `should_raise=False`, would be 2 explicit lines of code, e.g.
```python
PromotedAddonPromotion.objects.create(**self.required_fields)
assert PromotedAddonPromotion.objects.create(**{**self.required_fields, 'application_id': applications.ANDROID.id}) is not None
```

src/olympia/amo/tests/__init__.py
eviljeff: I think I +1 to Kevin here - _most_ tests that use `make_addon_promoted` will only be setting a single promotion in that test, so we just need to update the handful that make multiple `make_addon_promoted` calls in a single test to delete in between.  Some of _those_ tests may actually benefit from being expanded to test where is now multiple groups anyway.

src/olympia/amo/tests/__init__.py
```diff
@@ -614,6 +620,21 @@ def make_addon_promoted(cls, addon, group_id, approve_version=False, apps=None):
...
+        }
+
+    def make_promoted(self, addon, group_ids=None, apps=None, approve_version=False):
+        group_ids = group_ids or []
+        for group_id in group_ids:
+            self.make_addon_promoted(
+                addon=addon, group_id=group_id, apps=apps, reset=False
+            )
+        if approve_version:
+            addon.approve_for_version()
```
eviljeff, line 636: I don't think you need these in `TestCase` - they're only used in one test class.

(I'm not sure I'd have added `make_promoted` either, when it's only saving you from having to do 
```python
self.make_addon_promoted(addon=addon, group_id=PROMOTED_GROUP_CHOICES.RECOMMENDED, reset=False)
self.make_addon_promoted(addon=addon, group_id=PROMOTED_GROUP_CHOICES.LINE, reset=False)
```
i.e. 1 line per extra group but 🤷‍♂️ )

chrstinalin:
> there isn't any reference to Partner in addons-frontend - I'm not sure there should be

That's expected, partner is hidden from the API (see https://github.com/mozilla/addons/issues/14986)

PR #23274, opened 2025-04-07
Fix broken statistics dashboard. (Enable no-undef and no-reassign-const eslint rules)

KevinMind:
Fixes: mozilla/addons#15515

### Description

Enable eslint rules preventing:
- re-defining constants
- using undefined values

### Context

The statistics dashbaord was broken due to 2 lines of code violating these rules. It led to blank pages without any error validation.

Should we consider adding additional validation when there is a critical error?

### Testing

1. setup dev credentials for bigquery.
2. set the guid for one of your addons to a dev enabled stats guid
3. open the statistics page for that addon

Expect:
- the default view to load.
- changing the date range should work and update the chart, including custom date range

### Checklist

### DIFF
--- a/static/js/stats/chart.js
+++ b/static/js/stats/chart.js
@@ -229,7 +229,7 @@ const dayMsecs = 24 * 3600 * 1000;
     }
 
     // Transform xAxis based on time grouping (day, week, month) and range.
-    let pointInterval = (dayMsecs = 1 * 24 * 3600 * 1000);
+    let pointInterval = dayMsecs;
     let dateRangeDays = (end - start) / dayMsecs;
     baseConfig.xAxis.min = start - dayMsecs; // Fix chart truncation.
     baseConfig.xAxis.max = end;
--- a/static/js/zamboni/reviewers.js
+++ b/static/js/zamboni/reviewers.js
@@ -86,14 +86,14 @@ $(document).ready(function () {
 
     $('#addon-queue-filter-form .select-all').click(function (e) {
       e.preventDefault();
-      jQuery('#addon-queue-filter-form input[type="checkbox"]').prop(
+      $('#addon-queue-filter-form input[type="checkbox"]').prop(
         'checked',
         true,
       );
     });
     $('#addon-queue-filter-form .select-none').click(function (e) {
       e.preventDefault();
-      jQuery('#addon-queue-filter-form input[type="checkbox"]').prop(
+      $('#addon-queue-filter-form input[type="checkbox"]').prop(
         'checked',
         false,
       );
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -85,11 +85,9 @@ const config = defineConfig([
       // Disable rules causing errors
       'simple-import-sort/imports': 'off',
       'no-unused-vars': 'off',
-      'no-undef': 'off',
       'no-var': 'off',
       'prettier/prettier': 'off',
       'no-prototype-builtins': 'off',
-      'no-const-assign': 'off',
       'no-dupe-keys': 'off',
     },
   },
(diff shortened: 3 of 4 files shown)


docs/topics/runbooks/dev/shell_into_dev_pod.md
```diff
@@ -0,0 +1,59 @@
+# Shell into a dev pod
```
eviljeff, line 1: This is duplicating SRE's instructions in https://mana.mozilla.org/wiki/display/SRE/AMO+Dev+Resources, no?  I don't think we want to have it in two places, and (hopefully) SRE would maintain theirs if they change something.

I'm sure they wouldn't mind if we want to improve their docs though.

eviljeff (commented):
Can we move all the runbooks stuff into a separate PR - it's bloating a bugfix with extra content that deserves careful attention.

docs/topics/runbooks/dev/shell_into_dev_pod.md
KevinMind: I'd like to have all of our runbooks in one place. We can coordinate with SRE but there's no reason this info couldn't live here instead of on confluence.

KevinMind:
> Can we move all the runbooks stuff into a separate PR - it's bloating a bugfix with extra content that deserves careful attention.

Can do. I wanted to make sure the information was available in this PR since it is highly relevant for testing it. But I'll split off and link back.

docs/topics/runbooks/dev/shell_into_dev_pod.md
KevinMind: We could also just link there from here.

static/js/stats/chart.js
```diff
@@ -229,7 +229,7 @@ const dayMsecs = 24 * 3600 * 1000;
     }
 
     // Transform xAxis based on time grouping (day, week, month) and range.
-    let pointInterval = (dayMsecs = 1 * 24 * 3600 * 1000);
+    let pointInterval = dayMsecs;
```
willdurand, line 232: Where is `1 * 24 * 3600 * 1000` now?

willdurand (commented):
> Enable no-undef and no-reassign-const eslint rules.

Can we have a more accurate commit title for this? It does look like just a code quality improvement when we're actually fixing a regression on prod.

static/js/zamboni/reviewers.js
```diff
@@ -86,14 +86,14 @@ $(document).ready(function () {
 
     $('#addon-queue-filter-form .select-all').click(function (e) {
       e.preventDefault();
-      jQuery('#addon-queue-filter-form input[type="checkbox"]').prop(
+      $('#addon-queue-filter-form input[type="checkbox"]').prop(
```
willdurand, line 89: It just occurred to me.. What does _that_ fix?

static/js/stats/chart.js
KevinMind: `dayMsecs` is set to 24 * 36 * 1000 at the top of the file, so redefining it to that * 1 is redundant. `pointInterval` will further multiple the **constant** number of milliseconds in a day by whatever values it wants.

static/js/zamboni/reviewers.js
KevinMind: It fixes compliance with the eslint rule that we should not reference undefined globals in our own code. We do define both `$` and `jQuery` because we have vendored code and node module code that relies on it. But we can hold ourselves to a stricter standard and `$` was already imported in the module scope.

willdurand:
I am good with the answers to the questions, thank you. I don't need to re-review.

PR #23290, opened 2025-04-09
Revert ./stats/js/stats files closer to original state before vite migration

KevinMind:
Fixes: mozilla/addons#15515

### Description

This does 4 things:
- revert the files in `./static/js/stats` to https://github.com/mozilla/addons-server/tree/bdbcbcdd635a106d25cb65c42897b3fbc814ce9d/static/js/stats
- reapply surgical updates to make module graph work with vite and maintain removal of "z." state etc.
- add logs to all `changeview` events to debug on dev. (will be reverted before pushing)
- replace `typeof module === 'undefined'` calling of top level functions to direct calls in our bundle entrypoint

### Context

The statistics dashboard is broken and this is the fastest way to get the code back to its original state

### Testing

Follow these steps

https://github.com/mozilla/addons-server/pull/23274

You can verify that using `typeof module === 'undefined'` checks are not valid in vite bundles.

```javascript
console.log('module', typeof module);
```

This will return 'object' in production and 'undefined' in development. That is because in development we are loading modules with native ESM and in production we are using rollup bundled es modules. Rollup defines module but vite does not.

### Checklist

### DIFF
--- a/static/js/stats.js
+++ b/static/js/stats.js
@@ -3,8 +3,13 @@ import './stats/helpers';
 import './stats/dateutils';
 import './stats/manager';
 import './stats/controls';
-import './stats/overview';
+import { stats_overview } from './stats/overview';
 import './stats/topchart';
 import './stats/chart';
 import './stats/table';
-import './stats/stats';
+import { stats_stats } from './stats/stats';
+import { capabilities } from './zamboni/capabilities';
+
+// Initialize the stats module.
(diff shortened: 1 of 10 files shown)


willdurand (approved):
I reviewed the changes against 6057c9b56a518e0216a92a40ea8030839f804677, the last commit where stats pages were working fine. This looks very similar to what we had, except for the IIFEs that are now gone.

That looks good enough for now.

PR #23294, opened 2025-04-09
load stats_overview before stats_stats

KevinMind:
Fixes: mozilla/addons#15515

### Description

Ensure we load stats overview before triggering the changeview event which stats over view dependson.

### Context

Various places rely on $().on('changeview') to update the chart UI client side (reactive programming eh)

The problem here is we rely on a single trigger of that event in the `stats_sats` function. This ensures all the event listeners are called at least once with the initial state to render the correct stuff. That trigger was happening before we setup the topChart listener so the top chart is not rendered initially but is rendered on subsequent events.

### Testing

TBD

### Checklist

### DIFF
--- a/static/js/stats.js
+++ b/static/js/stats.js
@@ -11,5 +11,5 @@ import { stats_stats } from './stats/stats';
 import { capabilities } from './zamboni/capabilities';
 
 // Initialize the stats module.
-stats_stats(window.sessionStorage, capabilities);
 stats_overview();
+stats_stats(window.sessionStorage, capabilities);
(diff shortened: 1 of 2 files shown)


willdurand (approved):
As mentioned on Slack, looking at 6057c9b56a518e0216a92a40ea8030839f804677 again, overview.js is loaded before stats.js (in [config](https://github.com/mozilla/addons-server/blob/6057c9b56a518e0216a92a40ea8030839f804677/src/olympia/lib/settings_base.py#L715-L727)), and we get this order of execution:

1. https://github.com/mozilla/addons-server/blob/6057c9b56a518e0216a92a40ea8030839f804677/static/js/stats/overview.js#L82
2. https://github.com/mozilla/addons-server/blob/6057c9b56a518e0216a92a40ea8030839f804677/static/js/stats/stats.js#L136

Looks sensible.

PR #23299, opened 2025-04-10
Update Elasticsearch local image to 8.x, enable compatibility mode

diox:
ELASTIC_CLIENT_APIVERSIONING=1 forces our client libraries to send headers requesting compatibility with 7.x requests/responses, so we don't have to upgrade them yet and can wait for dev/stage/prod to be switched over.

Fixes https://github.com/mozilla/addons/issues/15351

### Testing

Should be transparent, although the first `make up` might take longer (or even initially fail until re-triggered, because the cluster takes too long to come online initially) to update the cluster.

You can do a few searches, make some updates to add-ons and see the updates being reflected in search results (excluding API caching).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,7 @@ x-env-mapping: &env
     - CELERY_RESULT_BACKEND=redis://redis:6379/1
     - DATABASES_DEFAULT_URL=mysql://root:@mysqld/olympia
     - ELASTICSEARCH_LOCATION=elasticsearch:9200
+    - ELASTIC_CLIENT_APIVERSIONING=1
     - MEMCACHE_LOCATION=memcached:11211
     - MYSQL_DATABASE=olympia
     - MYSQL_ROOT_PASSWORD=docker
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -7,6 +7,7 @@ x-env-mapping: &env
     - CELERY_RESULT_BACKEND=redis://redis:6379/1
     - DATABASES_DEFAULT_URL=mysql://root:@mysqld/olympia
     - ELASTICSEARCH_LOCATION=elasticsearch:9200
+    - ELASTIC_CLIENT_APIVERSIONING=1
```
diox, line 10: We'll need SRE to add that to our addons-server dev/stage/prod environments as well before they can switch the cluster for those environments to 8.x.

KevinMind: Will you file a follow up ticket or what's the plan?

KevinMind (approved):
🚢

docker-compose.yml
diox: Thinking about this more... We should have that in the `Dockerfile` - and once all envs are on 8.x server, when we upgrade the client libs, we'll remove it.

diox: https://github.com/mozilla/addons-server/pull/23299/commits/e902122f8a156384f38db3633729d5a8d24ac55d

PR #23308, opened 2025-04-11
Record which versions were affected when force disabling/enabling add-ons

diox:
Also show versions attached to activity logs explicitly in reviewer tools when there are more than one, so that reviewers can see in both the important changes history and the version history what versions were affected by activities.

Fixes: mozilla/addons#15099

### Testing

- Force disable add-ons with some versions attached in reviewer tools: the recorded `ActivityLog` should have `versions` attached with all affected versions, and this should be shown in the review page, both under "Add-on Important Changes History" and in the "Add-on History" for all those versions
- Repeat with force-enable
- Repeat with an add-on that would be held for second approval, the versions should be attached and shown in the review page both before and after the action is approved.


src/olympia/reviewers/templates/reviewers/addon_details_box.html
```diff
@@ -53,7 +53,7 @@ <h2>{{ addon_name }}</h2>
           </tr>
           <tr class="addon-status">
             <th>Status</th>
-            <td>
+            <td class="status-{{ amo.STATUS_CHOICES_API.get(addon.status, 'unknown') }}">
```
diox, line 56: Sort of a drive-by, but one of the complaints was that it wasn't always clear what happened to the add-on, so hopefully having a dedicated class (that applies a `color: ref`) for the disabled status will help a bit.

src/olympia/abuse/admin.py
```diff
@@ -329,6 +329,7 @@ def addon_card(self, obj):
         context = {
             'addon': addon,
             'addon_name': addon.name,
+            'amo': amo,
```
diox, line 332: Needed because of the changes to `addon_details_box.html`

src/olympia/abuse/actions.py
```diff
@@ -610,12 +624,34 @@ class ContentActionTargetAppealApprove(
 ):
     description = 'Reported content is within policy, after appeal'
 
+    @property
+    def target_versions(self):
+        target = self.target
+        if isinstance(target, Addon) and target.status == amo.STATUS_DISABLED:
+            qs = target.versions(
+                manager='unfiltered_for_relations'
+            ).disabled_that_would_be_renabled_with_addon()
```
eviljeff, line 633: Is there an alternative that doesn't involve us adding another queryset method to VersionManager that is only used here, and is a duplicate of the same method on FileManager?
`File.objects.filter(version__addon=target).disabled_that_would_be_renabled_with_addon().values_list('version', flat=True)`?

eviljeff: Or if we're worried about the extra overhead of loading full-fat Version instances, return `version__version` and `version_id` and send `*target_versions` as `*((Version, version_id) for _, version_id in target_versions)` instead.  (Because the underlying ActivityLog.create can accept (Class, pk) tuples rather than instances)

eviljeff (commented):
wdyt?

src/olympia/abuse/actions.py
diox: I wanted to stay as consistent as possible (which meant returning `Version`s) with the rest of the code while at the same time using a query that performs well and loads as little as possible because there could be thousands of versions.

I'll play around with it and try to remove the extra method, I also don't like it

src/olympia/abuse/tests/test_actions.py
```diff
@@ -542,34 +543,34 @@ def setUp(self):
...
+        )
 
         # except if there is no current_version, where the latest version is used
-        first_version.file.update(status=amo.STATUS_DISABLED)
-        second_version.file.update(status=amo.STATUS_DISABLED)
+        File.objects.update(status=amo.STATUS_DISABLED)
         self.addon.update_version()
         assert not self.addon.current_version
-        assert self.ActionClass(self.decision).addon_version == second_version
+        assert self.ActionClass(self.decision).addon_version == self.another_version
```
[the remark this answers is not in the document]
diox: Need to revisit this. That test was broken because of the new versions I added to the `setUp` here and in `TestContentActionRejectVersion` - I wasn't so much trying to improve it than trying to fix it

src/olympia/abuse/actions.py
eviljeff: I figured, but you don't need to have a `target_versions` method on this class - it could be called something different, or the query inlined elsewhere

src/olympia/reviewers/tests/test_utils.py
```diff
@@ -3596,15 +3604,34 @@ def test_disable_addon(self):
         activity_log = ActivityLog.objects.latest('pk')
         assert activity_log.action == amo.LOG.FORCE_DISABLE.id
         assert activity_log.arguments[0] == self.addon
-
+        # FIXME: There is an inconsistency between ReviewHelper.log_action()
+        # and ContentAction.log_action() regarding where to put the
+        # ContentDecision. See test_enable_addon() for comparison.
```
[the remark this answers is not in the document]
diox: I discovered this a bit late, I wasn't sure which one is better. I'd rather address this as a follow-up, the ordering shouldn't matter in the end, it's mostly for us in tests.

src/olympia/reviewers/utils.py
```diff
@@ -998,7 +999,13 @@ def record_decision(
             )
         assert cinder_action
 
-        versions = versions or ([self.version] if self.version else [])
+        if version is not None:
```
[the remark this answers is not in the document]
diox: Only reason for `version` and `decision_versions` to exist is deal with actions that need to record the version as `version` (singular) in the activitylog `details` - which the confirm auto-approval action needs.

I really wanted to avoid passing `versions`/`version` all the time because `log_action()` has some logic that does things depending on `self.version`/`version`/`versions` already... But let me take another look to see if I can simplify some of that.

src/olympia/abuse/actions.py
diox: Can't inline completely, the main issue is that the query is used in 2 places:
- Here, in the `ContentActionTargetAppealApprove`
- In `reviewers/utils.py`'s `ReviewHelper.enable_addon()`

But I've pushed 2c952700a2d618c2a450a1c508576a134511d5f4 which should be good enough and removes the duplicate method

src/olympia/reviewers/utils.py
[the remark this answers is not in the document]
diox: Simplified in https://github.com/mozilla/addons-server/pull/23308/commits/1ae15038f335300cc14f296208cf261fc67b581d ; this was really only needed to fix `confirm_auto_approved()` now that we're creating `log_action_kw` automatically, so I've fixed it in that specific method.

static/css/zamboni/reviewers.less
```diff
@@ -1374,6 +1379,10 @@ p.is_promoted {
     padding: 0 1em;
 }
 
+.status-disabled {
+    color: red;
+}
```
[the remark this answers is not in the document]
diox: I'll file a follow-up for that cause I want to avoid bikeshedding about the color choices blocking this PR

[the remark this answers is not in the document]
diox: https://github.com/mozilla/addons/issues/15551

src/olympia/reviewers/tests/test_utils.py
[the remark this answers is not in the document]
diox: https://github.com/mozilla/addons/issues/15552

src/olympia/reviewers/utils.py
```diff
@@ -1671,10 +1673,28 @@ def change_or_clear_pending_rejection_multiple_versions(self):
...
+        versions = tuple(
+            self.addon.versions(manager='unfiltered_for_relations')
+            .filter(pk__in=files_qs.values_list('version'))
+            .no_transforms()
+            .only('id', 'version')
+            .order_by('-pk')
+        )
         log.info('Sending email for %s' % (self.addon))
-        self.record_decision(amo.LOG.FORCE_ENABLE)
+        # We can't use record_decision(..., action_completed=False) like we do
```
eviljeff, line 1691: Not for this patch, obvs, but would it be possible to reconcile this code with `ContentActionTargetAppealApprove.process_action` at some point?

eviljeff (approved):

src/olympia/reviewers/utils.py
diox: Ah yes, that's the 3rd issue I need to file

PR #23315, opened 2025-04-14
Prepare Drop of Old Promotion Models

chrstinalin:
Fixes: mozilla/addons#15386

### Description

- Prepares drop of the legacy `PromotedAddon` and `PromotedApproval` models.
- Renames `PromotedAddonPromotion` -> `PromotedAddon`, and `PromotedAddonVersion` -> `PromotedApproval`.

### Checklist


src/olympia/promoted/models.py
```diff
@@ -148,118 +145,8 @@ def badged_groups(self):
...
-        ):
-            self.approve_for_addon()
-        elif self.group.flag_for_human_review:
-            self.addon.set_needs_human_review_on_latest_versions(
-                due_date=due_date,
-                reason=NeedsHumanReview.REASONS.ADDED_TO_PROMOTED_GROUP,
-            )
-
-
 # TODO: Drop Promotion suffix after dropping PromotedAddon table
```
eviljeff, line 148: I think you did this TODO :)

src/olympia/promoted/migrations/0030_remove_promotedaddon_addon_and_more.py
```diff
@@ -0,0 +1,33 @@
...
+        ),
+        migrations.DeleteModel(
+            name='PromotedTheme',
+        ),
+        migrations.DeleteModel(
+            name='PromotedAddon',
+        ),
+        migrations.DeleteModel(
+            name='PromotedApproval',
+        ),
```
eviljeff, line 32: This is risky to do in the same deploy to prod as the code being removed - the underlying sql will drop the tables.  Remove it from the migration and add a follow-up to re-generate the migrations in a subsequent sprint.

(PromotedTheme is fine to delete, assuming the references to that model were dropped in a previous deployment)

src/olympia/promoted/migrations/0031_promotedaddon_and_more.py
```diff
@@ -0,0 +1,22 @@
...
+
+    operations = [
+        migrations.RenameModel(
+            old_name='PromotedAddonPromotion',
+            new_name='PromotedAddon',
+        ),
+        migrations.RenameModel(
+            old_name='PromotedAddonVersion',
+            new_name='PromotedApproval',
+        ),
```
eviljeff, line 21: This is also very risky to do - it will do an sql RENAME.  Instead we manually define the db_table name in the model Meta object, so while the model is renamed in django then underlying database table doesn't change it's name.

(In the rare cases that we care about the database table name there's a longer process of writing to a new table and copying over the old rows, then deleting the old table in a subsequent deploy)

eviljeff (changes requested):

eviljeff (approved):

PR #23254, opened 2025-04-01
Mark swagger as experimental

KevinMind:
Fixes: mozilla/addons#15510

image

PR #23332, opened 2025-04-15
Enable swagger on dev

KevinMind:
Follow up: https://github.com/mozilla/addons-server/pull/23254

### Description

Enable swagger UI on dev

### Checklist


diox:
We have a command to generate the js files now, but is that running as part of our assets generation ? And does that bypass that setting ? (I'm assuming it should, unless that setting itself is set when running the command in the image build step)

KevinMind:
The setting only controls if we should serve the swagger UI it does not impact the endpoint serving the schema itself and the swagger JS asset generation relies only on the schema endpoint. Effectively this change enables /api/<version/swagger or redoc endpoints which currently would return 404.

diox:
Yeah I was wondering how the command was run and whether it depends on the setting, but I see it now (`update_assets` is what triggers the command and the command itself doesn't care about the setting)

diox (approved):

PR #23333, opened 2025-04-16
migrate PrimaryHero model to use addon instead of promoted_addon in a migration safe way

KevinMind:
Fixes: mozilla/addons#15557

### Description

- Changed the `addon` field in the `PrimaryHero` model to be nullable to be able to continue querying before the migration is complete
- Introduced a new task `sync_primary_hero_addon` to synchronize the `addon` field with the `promoted_addon` where applicable.

### Context

We tried running the migraiton via a single migration script. This can fail if the RunPython script executes before the database commits the new field.

Additionally, if there are a lot of PrimaryHero records the script can take an arbitrarily long time to run. Safer to use a task.

That means we need to revert the null constraint on addon until after this commit lands and the task is run. That is fine for now since all records will still have the promoted_addon field.

### Testing

Look at the unit tests. There are 2 scenarios:

- hero with promoted_addon without addon (the standard before case)
- hero with promoted_addon and addon (the standard after case)
- hero with addon and no promoted addon (non-standard but okay since we only care about addon from now on)
- hero with neither (non standard edge case that should never happen

### Checklist

### DIFF
--- a/src/olympia/hero/tasks.py
+++ b/src/olympia/hero/tasks.py
@@ -1,0 +1,20 @@
+from olympia.hero.models import PrimaryHero
+from olympia.amo.celery import task
+from olympia.amo.decorators import use_primary_db
+
+@task()
+@use_primary_db
+def sync_primary_hero_addon():
+    invalid_heroes = []
+    for hero in PrimaryHero.objects.filter(addon__isnull=True):
+        promoted_addon = hero.promoted_addon
+
+        if promoted_addon is None:
+            invalid_heroes.append(hero)
+        else:
+            hero.addon = promoted_addon.addon
+            hero.save()
+
+    if len(invalid_heroes) > 0:
+        raise ValueError(f'{invalid_heroes} heroes have no addon or legacy promoted addon')
+
--- a/src/olympia/hero/tasks.py
+++ b/src/olympia/hero/tasks.py
@@ -1,8 +1,13 @@
-from olympia.hero.models import PrimaryHero
+import olympia.core.logger
 from olympia.amo.celery import task
 from olympia.amo.decorators import use_primary_db
+from olympia.hero.models import PrimaryHero
 
-@task()
+
+log = olympia.core.logger.getLogger('z.hero')
+
+
+@task
 @use_primary_db
 def sync_primary_hero_addon():
     invalid_heroes = []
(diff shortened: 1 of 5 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -779,6 +779,7 @@ def get_db_config(environ_var, atomic_requests=True):
     'olympia.users.tasks.resize_photo': {'queue': 'amo'},
     'olympia.users.tasks.update_user_ratings_task': {'queue': 'amo'},
     'olympia.versions.tasks.delete_preview_files': {'queue': 'amo'},
+    'olympia.hero.tasks.sync_primary_hero_addon': {'queue': 'amo'},
```
diox, line 782: You can use the `adhoc` queue for one-off tasks like these - it has dedicated workers. See a few lines above this block

src/olympia/hero/tasks.py
```diff
@@ -0,0 +1,20 @@
+from olympia.hero.models import PrimaryHero
+from olympia.amo.celery import task
+from olympia.amo.decorators import use_primary_db
+
+@task()
```
diox, line 5:
```suggestion
@task
```

(both should work but we use the plain syntax when we don't have arguments to pass)

src/olympia/hero/tasks.py
```diff
@@ -0,0 +1,20 @@
...
+        promoted_addon = hero.promoted_addon
+
+        if promoted_addon is None:
+            invalid_heroes.append(hero)
+        else:
+            hero.addon = promoted_addon.addon
+            hero.save()
+
+    if len(invalid_heroes) > 0:
+        raise ValueError(f'{invalid_heroes} heroes have no addon or legacy promoted addon')
```
diox, line 19: No need to change anything, but raising at this point is not super useful - we are not in a transaction, and I don't think we want to be (we want the changes we made earlier to apply). Might as well just `log.error()` instead

KevinMind:
@diox thinking about it, I could add a quick check in the save method for PrimaryHero to ensure that either addon or promoted_addon is set.. wdyt? Maybe overkill I'm not sure.

diox (approved):
We'll want a follow-up that removes the `null=True` from the model & the database, and deletes the task from our code

diox:
Fix formatting as well

KevinMind:
> We'll want a follow-up that removes the `null=True` from the model & the database, and deletes the task from our code

https://github.com/mozilla/addons/issues/15560

PR #23369, opened 2025-04-25
Allow selecting "AMO_CLOSED_NO_ACTION" policies when resolving jobs in reviewer tools

diox:
Also make error handling for the review form a little more explicit to make debugging such issues easier in the future.

Fixes https://github.com/mozilla/addons/issues/15574

### Context

We have a policy in prod that is configured to trigger `AMO_CLOSED_NO_ACTION` decision action, which should be used to close jobs that were created for content that has since already been moderated. It has been enabled for reviewers, but using it doesn't work because we have a specific allow-list on our side (that is a temporary safeguard that will eventually be removed, but it's still here at the moment).

On top of that, the error was hidden because we didn't show global form errors, making the issue very confusing for reviewers.

### Testing

- Enable all cinder-related waffle switches
- In the admin ensure you have a Cinder Policy with `"amo-closed-no-action"` as the only `enforcement_actions`, and make that policy exposed in Reviewer Tools by flipping the checkbox. (If you trigger a sync with dev right now, `Ignore, specifically Content already moderated` should be configured with that enforcement action already, you just need to enable it in reviewer tools)
- Report an add-on for policy violation inside the add-on
- Go to that add-on review page
- Select "Resolve reports", picking the job for that report you just made
- Select the Policy mentioned above
- Submit

Expected result:
- Job should be closed, action recorded in activity log

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/review.html
+++ b/src/olympia/reviewers/templates/reviewers/review.html
@@ -102,9 +102,6 @@
 <form method="POST" enctype="multipart/form-data" action="#review-actions" class="review-form">
   {% csrf_token %}
 
-  <input type="hidden" name="version_pk" value="{{ version.pk }}"/>
-  {{ form.version_pk.errors }}
-
   <div id="review-files-header">
     <h3 id="history">
       Add-on History
@@ -140,6 +137,12 @@
         You don't have permission to review it.
       {% endif %}
     </p>
+  {% endif %}
+
+  <input type="hidden" name="version_pk" value="{{ version.pk }}"/>
+  {{ form.version_pk.errors }}
+  {% if form.errors.__all__ %}
+    {{ form.errors.__all__ }}
   {% endif %}
 
   <div id="review-actions" class="review-actions">
--- a/src/olympia/reviewers/forms.py
+++ b/src/olympia/reviewers/forms.py
@@ -485,14 +485,14 @@ class ReviewForm(forms.Form):
 
     def clean(self):
         super().clean()
-        # If the user select a different type of job before changing actions there could
-        # be non-appeal jobs selected as cinder_jobs_to_resolve under resolve_appeal_job
-        # action, or appeal jobs under resolve_reports_job action. So filter them out.
         if self.cleaned_data.get('attachment_input') and self.cleaned_data.get(
             'attachment_file'
         ):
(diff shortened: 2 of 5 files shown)


src/olympia/reviewers/templates/reviewers/review.html
```diff
@@ -142,6 +139,12 @@ <h3 id="history">
     </p>
   {% endif %}
 
+  <input type="hidden" name="version_pk" value="{{ version.pk }}"/>
+  {{ form.version_pk.errors }}
+  {% if form.errors.__all__ %}
+    {{ form.errors.__all__ }}
+  {% endif %}
+
```
diox, line 147: Moved the error about the version here so that it's easier to spot, and added global form errors as well. That part can't be tested anymore now that I've also changed any of the existing global errors to be attached to the specific field they relate to, but it's useful for future-proofing.

KevinMind:
Created the abuse report, resolved in reviewer tools. Resolved with selected inputs below

Screenshot 2025-04-28 at 09 34 48

How can I verify from admin that the job is resolved?

diox:
If it's closed, you shouldn't see it anymore in the reviewer tools after the action has been performed (it's no longer outstanding)

KevinMind:
> If it's closed, you shouldn't see it anymore in the reviewer tools after the action has been performed (it's no longer outstanding)

It would just help as a reviewer to know the counterfactual. What would it look like if it were not closed for example. I guess in my case it is closed.

KevinMind (approved):
LGTM 🚢

diox:
In the current UI, reviewers only see open (outstanding) jobs - once a job is closed, until there is an appeal they won't see it anymore.

KevinMind:
FWIW I meant a code reviewer. Would there have been something in the /admin I could have seen? Or I guess I could have first verified the report is visible and then after resolving it should no longer be visible.

diox:
Yes, the `CinderJob` would have a `ContentDecision` attached with an `action` of `Content already removed (no action)` (`AMO_CLOSED_NO_ACTION`). You would also have had an `ActivityLog` with `action` set to `RESOLVE_CINDER_JOB_WITH_NO_ACTION`.

See unit test I've added:
https://github.com/mozilla/addons-server/pull/23369/files#diff-396c5c8e10f3f679d423384f7c16ff803786fc4488d91a029c142d70125b97adR1143-R1148

PR #23371, opened 2025-04-25
Fix broken settings import in review_reports command

diox:
Fixes https://github.com/mozilla/addons/issues/15543

### Testing

This is annoying to test locally or in CI: It fails in dev/stage/prod prod because `import settings` actually imports `./settings.py`, which in turn fails when we try to evaluate `SITE_URL` since it's set to `env('SITE_URL')` and there is no such environment variable on dev/stage/prod. Locally and in CI that variable _is_ set through `docker-compose.yml`.

I triggered the command on dev, it failed, then applied this patch, it worked.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ select = [
     "I",  # isort
     "Q",  # flake8-quotes
     "W",  # pycodestyle warnings
+    "TID251",  # banned imports
 ]
 
 [tool.ruff.lint.flake8-quotes]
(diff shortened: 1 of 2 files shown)


eviljeff (approved):
Fun that this bug has effectively existed since `review_reports` was committed, but importing `settings.py` didn't fail any longer.

... I feel like there is some underlying problem here - like we should never be importing `settings` directly (linting?), and/or `settings.py` shouldn't fail on dev/stage/page (or _should_ it?  Maybe it should _deliberately_ fail?)

diox:
Added a `lint` rule in https://github.com/mozilla/addons-server/pull/23371/commits/89961ddb83bd88c89a9d1e34684a35f0da80a4b4 ; I tested it successfully on `master` first.

PR #23372, opened 2025-04-25
Prevent non-public groups from appearing in DevHub status

chrstinalin:
Fixes: mozilla/addons#15548

### Description

Hides promoted groups with `is_public=False` from appearing in DevHub.

### Checklist


KevinMind (changes requested):
This should be a database field. Please don't add more logic based on the constants as we should be aiming to remove those from the codebase entirely relatively soon.

diox:
Is there any reason not to simply follow `is_public` here ? I don't think developers should see more information than what we're exposing in the API.

src/olympia/promoted/models.py
```diff
@@ -21,9 +21,9 @@ def __getattr__(self, attribute):
             return self.values_list(attribute, flat=True)
         raise AttributeError(f'PromotedGroup has no attribute: {attribute}')
 
-    @property
-    def name(self):
-        return ', '.join(self.__getattr__('name'))
+    def get_names(self, needs_badged=True):
```
KevinMind, line 24: NIT: I don't see the value in changing "badged" to "needs_badged" if they mean the same thing.

src/olympia/abuse/cinder.py
```diff
@@ -353,7 +353,7 @@ def get_attributes(self):
             'name': self.get_str(self.addon.name),
             'slug': self.addon.slug,
             'summary': self.get_str(self.addon.summary),
-            'promoted': self.get_str(promoted_group.name),
+            'promoted': self.get_str(promoted_groups.get_names(needs_badged=False)),
```
KevinMind, line 356: If `.get_names` is just filtering by badged or not badged and then returning the .name is it really beneficial to have this as a model method at all? How is it different/better than the shorter and built in?

```python
promoted_group.filter(badged=False).name
```

src/olympia/abuse/cinder.py
```diff
@@ -343,7 +343,7 @@ def get_attributes(self):
         # promoted in any way, but we don't care about the promotion being
         # approved for the current version, it would make more queries and it's
         # not useful for moderation purposes anyway.
-        promoted_group = self.addon.promoted_groups(currently_approved=False)
+        promoted_groups = self.addon.promoted_groups(currently_approved=False)
```
KevinMind, line 346: changing the variable name across all of the files from `promoted_group` to `promoted_groups` makes the diff harder to read. That would be a good candidate for a separate commit perhaps the first commit.

src/olympia/promoted/models.py
KevinMind: TBH adding the argument here feels like an unecessary abstraction. The queryset already has the ability to filter by properties. Just use the filter and return the name property where needed.

LIke I've said above, if we wanted a convenient way to query promoted groups that are badged we could add that on the `promoted_groups` method but even that is probably overkill.

KevinMind (commented):
Can you update the description as the core changes are pretty different than what was originally described. It doesn't mention anything about badged for example.

src/olympia/abuse/cinder.py
chrstinalin: I was concerned with the future use of `.name` and the possibility of something similar to what exists already occurring again (since its not immediately obvious that public names should be filtered out.) I do prefer this way and it's what I did initially, though. Given there's only two places where filtering is necessary, probably just better to go with this.

KevinMind (approved):
Verified with partner addon

image

image

PR #23373, opened 2025-04-25
Hide Strategic and Notable groups from the API

chrstinalin:
Fixes: mozilla/addons#14987

### Description

Prevents strategic and notable groups from being exposed through the API.

### Testing

1. `GETs` to `/api/v5/addons/addon/(int:id|string:slug|string:guid)/` should not expose the strategic or notable groups.
2. Other promoted groups (excluding partner) should still be exposed.

### Checklist


diox:
Spotlight should be public - we depend on that to [hide the installation warning in the frontend](https://github.com/mozilla/addons-frontend/blob/52f510ade61175f7538ebe20a3a6db4294af10c8/src/amo/constants.js#L340)

diox:
We're checking `publicly_promoted_groups` when indexing, so we are going to need a reindex of all add-ons in those promoted groups, right ? That could be triggered from the migration.

PR #23386, opened 2025-04-28
Ignore all decisions that originated from the api

eviljeff:
Fixes: mozilla/addons#15525

### Description

refactors `process_webhook_payload_decision` so we filter out all api originated decisions with a 200, and raise a different 400 error if we get a proactive decision _not_ from the api.

### Context

`process_webhook_payload_decision` was previously checking `job.resolvable_in_reviewer_tools` to filter out api_decisions, but only for jobs - not proactive decisions.  Instead we were treating all non-job decisions as overrides and then raising when the decision id didn't exist (which didn't exist for a new, non override, decision).  

There's a bonus performance win in that we won't try to query any CinderJobs or ContentDecisions for any api originated decision now - we previously queried all decisions to find a job, to then check if was `resolvable_in_reviewer_tools==True`.

### Testing

As always, annoying to test locally because you have to replay webhook payloads from Cinder.

To test proactive reviewer decisions are ignored:
1) Reject or disable any addon/version in the reviewer tools, with Cinder API integration set up
2) See the decision in `/dashboards/decisions-created` in Cinder
3) replay the webhook payload from `/settings/webhooks` with curl
4) get a 200 with "Decision already handled via reviewer tools" in the response

To test proactive cinder decisions are ignored:
1) identify any add-on in investigate (e.g. investigate/amo_addon/616992)
2) take a manual action (right hand 3 dot menu) with a supported enforcement action (e.g. AMO_DISABLE_ADDON)
3) replay the webhook payload from `/settings/webhooks` with curl
4) get a 400 with "Unsupported Manual decision" in the response

### Checklist

### DIFF
--- a/src/olympia/abuse/views.py
+++ b/src/olympia/abuse/views.py
@@ -208,25 +208,23 @@ def filter_enforcement_actions(enforcement_actions, cinder_job):
 def process_webhook_payload_decision(payload):
     source = payload.get('source', {})
     log.info('Valid Payload from AMO queue: %s', payload)
-    if source.get('decision').get('type') == 'api_decision':
+    if source.get('decision', {}).get('type') == 'api_decision':
         log.debug('Cinder webhook decision for api decision skipped.')
         raise CinderWebhookIgnoredError('Decision already handled via reviewer tools')
-    elif 'job' in source:
-        job_id = source.get('job', {}).get('id', '')
-
+    elif job_id := source.get('job', {}).get('id'):
         try:
             cinder_job = CinderJob.objects.get(job_id=job_id)
         except CinderJob.DoesNotExist as exc:
             log.debug('CinderJob instance not found for job id %s', job_id)
             raise CinderWebhookMissingIdError('No matching job id found') from exc
-    elif prev_decision_id := payload.get('previous_decision', {}).get('id', ''):
+    elif prev_decision_id := payload.get('previous_decision', {}).get('id'):
         try:
-            decision = ContentDecision.objects.get(cinder_id=prev_decision_id)
+            prev_decision = ContentDecision.objects.get(cinder_id=prev_decision_id)
         except ContentDecision.DoesNotExist as exc:
             log.debug('ContentDecision instance not found for id %s', prev_decision_id)
             raise CinderWebhookMissingIdError('No matching decision id found') from exc
 
-        cinder_job = decision.cinder_job
+        cinder_job = prev_decision.cinder_job
         if not cinder_job:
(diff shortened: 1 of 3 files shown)


src/olympia/abuse/views.py
```diff
@@ -208,31 +208,32 @@ def filter_enforcement_actions(enforcement_actions, cinder_job):
 def process_webhook_payload_decision(payload):
     source = payload.get('source', {})
     log.info('Valid Payload from AMO queue: %s', payload)
-    if 'job' in source:
+    if source.get('type') == 'api_decision':
```
diox, line 211: Looking at the payload I got from Cinder, `type` should be under `decision` in `source`, not `source` itself. Otherwise following your STR I'm getting a 400 Error with `Unsupported Manual decision`.

diox (changes requested):

src/olympia/abuse/views.py
eviljeff: 😬 corrected, and included a (sanitized) real payload to test against https://github.com/mozilla/addons-server/pull/23386/commits/8635168147d24be9decf8abfd777f164f0a930ae

src/olympia/abuse/views.py
```diff
@@ -208,31 +208,32 @@ def filter_enforcement_actions(enforcement_actions, cinder_job):
...
         except CinderJob.DoesNotExist as exc:
             log.debug('CinderJob instance not found for job id %s', job_id)
             raise CinderWebhookMissingIdError('No matching job id found') from exc
-    else:
-        decision_id = payload.get('previous_decision', {}).get('id', '')
-
+    elif prev_decision_id := payload.get('previous_decision', {}).get('id', ''):
         try:
-            decision = ContentDecision.objects.get(cinder_id=decision_id)
+            decision = ContentDecision.objects.get(cinder_id=prev_decision_id)
```
diox, line 224: For consistency now that you've renamed `decision_id` to `prev_decision_id`:
```suggestion
            prev_decision = ContentDecision.objects.get(cinder_id=prev_decision_id)
```

And then adjust below where it's used ?

src/olympia/abuse/views.py
```diff
@@ -208,31 +208,32 @@ def filter_enforcement_actions(enforcement_actions, cinder_job):
 def process_webhook_payload_decision(payload):
     source = payload.get('source', {})
     log.info('Valid Payload from AMO queue: %s', payload)
-    if 'job' in source:
+    if source.get('decision').get('type') == 'api_decision':
```
diox, line 211: Safer with:

```suggestion
    if source.get('decision', {}).get('type') == 'api_decision':
```

PR #23301, opened 2025-04-11
allow multiple ContentDecision to link to CinderJob; etc

eviljeff:
Fixes: mozilla/addons#15491

### Description

Allows any add-on decision held for 2nd level approval to be cancelled and re-flagged for NHR in the reviewer tools.

image

### Context

This PR contains some significant refactoring of how `ContentDecision`s are linked to `CinderJob`s - it reverses the foreign key, allowing multiple decisions to be associated with a job (this was possible in python previously by recursively looking up the `overridden_by`, but it's impractical to do in a query).  It also expands the concept of decisions that aren't really a decision (like how a decision with an action of AMO_ESCALATE wasn't a decision, but a way of moving queues) and allows jobs where the latest decision has an AMO_REQUEUE action to be seen as unresolved, so they can be re-resolved with a new decision (that would override the previous decision).

This does leave, effecitivey, 3 different ways a job can be forwarded - https://github.com/mozilla/addons/issues/15542 will remove one of them (the AMO_ESCALATE path), and potentially queue moves could also be represented in this way.

### Testing

First with a job
- make an extension promoted (e.g. recommended)
- either upload a version and approve it to sign it, or manually make the file is_signed=True with django admin|shell
- report the extension for abuse in the add-on, as a policy violation
- (find the job in Cinder, in the "AMO Reviewer escalations" queue)
- reject a version for that add-on via the reviewer tools, resolving the job
- in 2nd level approvals see that decision being held and choose "cancel and enqueue"
- see the add-on is re-flagged for NHR in the manual review queue
- reject it, resolving the job, again to see this can be done repeatedly
- check that the decision is reported correctly in Cinder - i.e. it has closed the job - only the final reject decision should be recorded

Then without a job:
- similarly, this should work when there _isn't_ a job to start with (i.e. a proactive decision)
- make an extension promoted (e.g. recommended)
- either upload a version and approve it to sign it, or manually make the file is_signed=True with django admin|shell
- reject a version for that add-on via the reviewer tools (without resolving any jobs, if there are any present)
- in 2nd level approvals see that decision being held and choose "cancel and enqueue"
- see the add-on is re-flagged for NHR in the manual review queue
- see there is now a forwarded job
- reject it, resolving the job this time, to see this can be done repeatedly
- check that the decision is reported correctly in Cinder - i.e. there is a stand-alone proactive decision - only the final reject decision should be recorded

### Checklist

[...]

PR #23387, opened 2025-04-28
allow null CinderJob.job_id

eviljeff:
Fixes: mozilla/addons#15571

### Description

Fixes #15571 in the most simple way - allows `null` values for `CinderJob.job_id`.  

### Context
#23301 creates CinderJob instances without `job_id`s, because they aren't reported to Cinder - and because not specifying a `job_id` in the `.create` worked without any complaints I missed that the field was getting a value of `''` - and that only works exactly one time, until the `unique` restriction prevents the second one.

### Testing
- Make decision that would be held for 2nd level approval (e.g. rejection on a promoted add-on), with no cinder job.
- then choose the action to requeue it in the reviewer tools (and see it requeued in the reviewer tools, with a cinder job for the requeue)
- Repeat with another add-on and get the same outcome.

### Checklist


src/olympia/abuse/models.py
```diff
@@ -1346,6 +1350,7 @@ def requeue_held_action(self, *, user, notes):
             job = CinderJob.objects.create(
                 target_addon=self.addon,
                 resolvable_in_reviewer_tools=True,
+                job_id=None,
```
KevinMind, line 1353: Is this for readability or do we have to specify none here?

KevinMind:
@eviljeff can you clarify for

> Make decision that would be held for 2nd level approval (e.g. rejection on a promoted add-on), with no cinder job.

What are the concrete steps for doing this? Should I manually create the records or is there a process to get there via admin/review-tools etc?

src/olympia/abuse/models.py
```diff
@@ -215,12 +215,16 @@ def report(cls, abuse_report):
             resolved_in_reviewer_tools=abuse_report.is_handled_by_reviewers,
         )
         job_id = entity_helper.report(report=report_entity, reporter=reporter_entity)
-        defaults = {
```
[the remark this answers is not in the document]
eviljeff: it's was just a refactoring: I initially couldn't find all the places where we were creating CinderJob instances (because it was using `cls.objects` instead).  Then after I started I went a little further because a `defaults` dict is redundant - it's only used in one place, and it's more explicit when the values are defined directly in the kwargs.  

`target if isinstance(target, Addon) else None` is the same as `abuse_report.addon` - you can trace what `abuse_report.addon` is, and how `abuse_report.target` is the same value when it's an Addon instance, and `None` otherwise.  More refactoring to the typical use elsewhere in the file.

None of it is vital, I just did it so the code - and how we set `job_id` looked visually similar to the other place we create a CinderJob.

src/olympia/abuse/models.py
eviljeff: it's necessary - the default is still `''` (though the logic for working out what a default value is when undefined is [complicated](https://github.com/django/django/blob/4.2.20/django/db/models/fields/__init__.py#L960))... though maybe I should set a default on the field and disallow any future empty string values...

eviljeff:
> @eviljeff can you clarify for
> 
> > Make decision that would be held for 2nd level approval (e.g. rejection on a promoted add-on), with no cinder job.
> 
> What are the concrete steps for doing this? Should I manually create the records or is there a process to get there via admin/review-tools etc?

reject a version of a promoted add-on in the reviewer tools.

KevinMind:
The first iteration worked, but for some odd reason, rejecting additional addons did not lead to them being included in the 2nd level approval queue, though they were rejected.

https://github.com/user-attachments/assets/d4833e41-7277-4a77-bdcd-362a751065e0

KevinMind (approved):
Tentatively approving, code looks good but I was not able to completely verify so please ensure it actually works before merging. Shared a video of the repro.

src/olympia/abuse/models.py
eviljeff: https://github.com/mozilla/addons/issues/15583

PR #23390, opened 2025-04-29
Enable Partner Group for High Profile Rating

chrstinalin:
Fixes: mozilla/addons#15544

### Description

Enables `high_profile_rating` for the partner promoted group. 

### Checklist


src/olympia/constants/promoted.py
```diff
@@ -159,6 +159,7 @@ def __bool__(self):
     name=_('Partner'),
     api_name=PROMOTED_GROUP_CHOICES.PARTNER.api_value,
     high_profile=True,
+    high_profile_rating=True,
```
eviljeff, line 162: Aren't these constants obsolete? Do we need to be keep them up to date any longer?  (shouldn't they be deleted all together? - not in this PR though)

src/olympia/promoted/migrations/0029_auto_20250429_1315.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 4.2.20 on 2025-04-29 13:15
+
+from django.db import migrations
+
+from olympia.constants.promoted import PROMOTED_GROUP_CHOICES
+
+def set_partner_group_high_profile_rating(apps, schema_editor):
+    PromotedGroup = apps.get_model('promoted', 'PromotedGroup')
+    group = PromotedGroup.objects.get(group_id=PROMOTED_GROUP_CHOICES.PARTNER)
```
eviljeff, line 9: is it possible for the situation to occur where this group does not exist? If so, you'll need to handle it by wrapping in a try (or otherwise handle it in a safe way).  Similarly with the reverse - database migrations should never fail.

eviljeff (approved):

src/olympia/constants/promoted.py
chrstinalin: Yeah, they will be -- this is just to prevent one of the sync tests from failing in the mean time.

src/olympia/promoted/migrations/0029_auto_20250429_1315.py
chrstinalin: Hm, it shouldn't with the dependencies, but doesn't hurt.

PR #23395, opened 2025-04-30
Remove locales below 80% threshold

diox:
Fixes: mozilla/addons#15558

### DIFF
--- a/src/olympia/devhub/tests/test_forms.py
+++ b/src/olympia/devhub/tests/test_forms.py
@@ -1108,7 +1108,7 @@ class TestAdditionalDetailsForm(TestCase):
 
     def test_locales(self):
         form = forms.AdditionalDetailsForm(request=self.request, instance=self.addon)
-        assert form.fields['default_locale'].choices[0][0] == 'ar'
+        assert form.fields['default_locale'].choices[0][0] == settings.LANGUAGES[0][0]
 
     def _get_tag_text(self):
         return [t.tag_text for t in self.addon.tags.all()]
--- a/src/olympia/files/tests/test_utils.py
+++ b/src/olympia/files/tests/test_utils.py
@@ -1205,7 +1205,6 @@ def test_extract_translations_simple():
         'de',
         'en-US',
         'ja',
-        'nb-NO',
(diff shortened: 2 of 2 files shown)


src/olympia/amo/management/commands/check_locales_completion_rate.py
```diff
@@ -47,12 +47,11 @@ class Command(BaseCommand):
     # that by querying amo and amoFrontend.
     PONTOON_PROJECTS = 2
```
KevinMind, line 48: Question. do we currently have a CI check that locales above the threshold are included? That would be the thing I'd expect to make a PR like this pass. Does that currently exist?

KevinMind (approved):
Approved with question/comment

src/olympia/addons/tests/test_views.py
```diff
@@ -2339,7 +2338,7 @@ def test_set_homepage_support_url_email(self):
         }
 
         data = {
-            'homepage': {'ro': f'{settings.EXTERNAL_SITE_URL}'},
```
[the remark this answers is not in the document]
diox: The constant with the locales that are never removed are in the command... but in this particular instance we did remove one such locale (`ar`) because it was really too low, so...

On the plus side, the locales that are left hardcoded in these tests, `fr`, `de`, `it`... are _really_ unlikely to ever be removed.

src/olympia/amo/management/commands/check_locales_completion_rate.py
diox: We don't have a check that calls pontoon in CI, so we don't check that all locales are below threshold in CI. This was explicitly not a goal of this project, multiple reasons for this:
- The command (and therefore the threshold) is only there to generate an email once a month. We then manually review the email to decide which locales to keep/add. We don't necessarily follow the threshold, it's more of a recommendation (though it will warn about the same locale the next month if things don't change)
- Things change quickly. New strings are added, translations are changed etc several times per week. We don't want CI to break just because we're suddenly at 79% for a given locale for a couple days, or even a couple weeks.

PR #23401, opened 2025-05-01
Expose the policy text for decisions in version|important-changes history

eviljeff:
Fixes: mozilla/addons#15503

### Description

Simple fix to expose `policy_texts` alongside review comments where they're shown.  If it's too much or too ugly can be iterated on.

Screenshot 2025-05-01 161253

Screenshot 2025-05-01 151937

### Context

`policy_text` is only be added to `details` for Cinder originated decisions while the reviewer tools is including all the policy text in the comments.  https://github.com/mozilla/addons/issues/15513 will address that.

The subtle different in formatting between the Hold and the actual Force Disable (I disabled a recommended add-on, so it was held for 2nd level approval) is because I resync'd the policies in the meantime and the policy text had been updated with some extra line breaks.  What's displayed is always the policy text at the time the activity log was created, so we have a immutable record of it.

### Testing

- report an add-on for abuse for something Cinder handled (e.g. legal)
- in Cinder resolve that job with some policies that disable the add-on.  Add some manual reasoning if you like.
- replay the webhook payload for that decision into your local web container with curl
- in reviewer tools navigate to the review page for that add-on and see the policy text shown for those decisions along side any manual reasoning as comments.

### Checklist

### DIFF
--- a/src/olympia/reviewers/templates/reviewers/includes/history.html
+++ b/src/olympia/reviewers/templates/reviewers/includes/history.html
@@ -22,6 +22,11 @@
               </details>
             {% endif %}
           {% endif %}
+          <ul>
+          {% for policy_text in record.details.policy_texts %}
+            <li class="light history-comment">{{ policy_text }}</li>
+          {% endfor %}
+          </ul>
           {# <pre> rather than <div> so that white space is preserved on copy #}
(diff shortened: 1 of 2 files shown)


KevinMind:
Verified with decision

image

KevinMind (approved):
LGTM

PR #23420, opened 2025-05-05
fix URL redirection from remote source on `decorators()`

odaysec:
Fixes: [mozilla/addons#poc-moz5-6-25.md](https://gist.github.com/odaysec/3a201ac91542b0804682792776526d28)

https://github.com/mozilla/addons-server/blob/355739b86579b8f68790ec1dbd23c87b9ee8ddd0/src/olympia/addons/decorators.py#L39-L39

fix the issue, we need to ensure that the constructed URL is safe for redirection. This can be achieved by validating the `addon.slug` and ensuring that the resulting URL is relative (i.e., does not include a host or scheme). We can use Django's `url_has_allowed_host_and_scheme` utility to validate the URL or ensure that the `netloc` and `scheme` attributes of the parsed URL are empty. Additionally, we should sanitize `request.path` to remove any backslashes (`\`) that could be misinterpreted by browsers.

The changes will be made in the `addon_view` function in `src/olympia/addons/decorators.py`. Directly incorporating user input into a URL redirect request without validating the input can facilitate phishing attacks. In these attacks, unsuspecting users can be redirected to a malicious site that looks very similar to the real site they intend to visit, but which is controlled by the attacker.

### POC
https://gist.github.com/odaysec/3a201ac91542b0804682792776526d28

### Checklist

### DIFF
--- a/src/olympia/addons/decorators.py
+++ b/src/olympia/addons/decorators.py
@@ -1,6 +1,7 @@
 import functools
 
 from django import http
+from urllib.parse import urlparse
 from django.shortcuts import get_object_or_404
 from django.template.response import TemplateResponse
 from django.utils.translation import gettext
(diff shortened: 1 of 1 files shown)


diox:
1. The POC in your gist doesn't work
2. Only very specific characters are allowed in `addons`.`slug`, so there is already validation preventing this class of attack in this decorator.
3. Any change would need to pass existing tests and be covered by new ones to ensure no regressions.

PR #23426, opened 2025-05-06
Re-order needs human review reasons, display them in the review queue for each add-on

diox:
Fixes https://github.com/mozilla/addons/issues/15536
Fixes https://github.com/mozilla/addons/issues/15527

### Screenshot

**Before**
Screenshot 2025-05-07 at 13-01-38 🛠️ Manual Review - Add-ons for Firefox

**After**

Screenshot 2025-05-07 at 11-27-55 🛠️ Manual Review - Add-ons for Firefox

### DIFF
--- a/src/olympia/reviewers/models.py
+++ b/src/olympia/reviewers/models.py
@@ -885,17 +885,32 @@ class NeedsHumanReview(ModelBase):
     review."""
 
     REASONS = ChoicesWithNHRAnnotationPrefix(
-        ('UNKNOWN', 0, 'Unknown'),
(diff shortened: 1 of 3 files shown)


diox:
I've added a before screenshot for comparison. The issues linked asked to re-order the reasons in the specific order given, and to make them visible in the queue for each row, as well as re-order the columns in the review queue, so you can check that it has indeed been changed and that it matches the ask from the ticket

KevinMind (approved):
FWIW what you described in the comment is a perfect and acceptable testing section. Absence indicates either you forgot or there are none, explicitly saying none (or in this case checking the list is correct via the screenshot) is more clear for reviewers and QA

PR #23478, opened 2025-05-20
Fix urlparams() when a query has multiple values for a given key

diox:
Fixes https://github.com/mozilla/addons/issues/15532

### Description

This fixes review queue filter breakage caused by pagination or sorting as those filters work by adding values to the same `due_date_reasons` key.

### Context

We use `urlparams()` in a bunch of places to add params to an URL to do pagination and things like that - it's very old code that was introduced 15 years ago and has been patched a couple times since. Because it was using a `dict` internally it didn't support keys with multiple values, which is a valid scenario that can happen when dealing with URLs created from forms with `<select multiple>` or `<input type=checkbox>`.

### Testing

I've added a couple unit tests demonstrating the following works:
```python
>>> urlparams('https://example.com/?foo=one&foo=two', bar='extra')
'https://example.com/?foo=one&foo=two&bar=extra'
```
To test that in the real world, you can:
- Make sure you have some add-ons in the review queue
- Visit the manual review queue
- Click link to uncheck all filters
- Check a few checkboxes (that match reasons shown for add-ons already in the queue)
- Submit
- If you have multiple results (lucky you, you have lots of add-ons waiting to be reviewed) and pagination is shown, click on page 2. Alternatively, sort by due date in reverse order.
- Make sure the same filters you selected before are still active on that new page

### DIFF
--- a/src/olympia/amo/tests/test_helpers.py
+++ b/src/olympia/amo/tests/test_helpers.py
@@ -202,6 +202,22 @@ def test_urlparams():
     assert s == url
 
 
+def test_urlparams_multiple_values():
+    url = 'https://example.com/?foo=one&foo=two'
+    assert (
+        utils.urlparams(url, bar='extra')
+        == 'https://example.com/?foo=one&foo=two&bar=extra'
+    )
+
+    # Replacing value for existing param with multiple values
+    url = 'https://example.com/?foo=one&foo=two'
+    assert utils.urlparams(url, foo='newvalue') == 'https://example.com/?foo=newvalue'
+
+    # Removing existing param with multiple values
+    url = 'https://example.com/?foo=one&foo=two'
+    assert utils.urlparams(url, foo=None) == 'https://example.com/'
+
+
 def test_urlparams_unicode():
     url = '/xx?evil=reco\ufffd\ufffd\ufffd\u02f5'
     utils.urlparams(url)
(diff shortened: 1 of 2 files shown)


src/olympia/amo/tests/test_helpers.py
```diff
@@ -226,8 +242,8 @@ def test_urlparams_returns_safe_string():
     s = render('{{ "https://foo.com/"|urlparams(param="a%20b") }}', {})
     assert s == 'https://foo.com/?param=a+b'
 
-    s = render('{{ "https://foo.com/"|urlparams(param="%AAA") }}', {})
```
diox, line 229: This wasn't valid, and now that I'm using a `QueryDict` it wouldn't work properly.

diox:
@KevinMind  That looks expected to me ? You have `foo=baz`, you want `bar=extra` and `foo=bar`. The original `foo=baz` gets overwritten by `foo=bar`, and we add `bar=extra`.

KevinMind:
And 

```python
urlparams('https://example.com?foo=baz', bar='extra', foo=['bar', 'baz'])
```

is expected to return an encoded array not multiple params?

diox:
That's undefined behavior - I'm not sure what would be expected, it's not a use case that has come up before. The bug was about preserving params with multiple values in the existing URL, not trying to pass new ones.

KevinMind (approved):
🚢

PR #23484, opened 2025-05-22
Handle version rejection appeals forwarded to legal

eviljeff:
Fixes: mozilla/addons#15296

### Checklist


src/olympia/reviewers/utils.py
```diff
@@ -1659,7 +1661,7 @@ def unreject_latest_version(self):
 
         self.set_file(amo.STATUS_AWAITING_REVIEW, self.version.file)
         self.log_action(amo.LOG.UNREJECT_VERSION)
-        self.addon.update_status(self.user)
+        self.addon.update_status()
```
eviljeff, line 1664: this (and below) were drive-by fixes when I was looking into update_status - the arg is supposed to be a `Version` instance to ignore, so passing `self.user` was always useless.

src/olympia/reviewers/utils.py
```diff
@@ -969,10 +969,12 @@ def set_file(self, status, file):
         file.datestatuschanged = datetime.now()
         if status == amo.STATUS_APPROVED:
             file.approval_date = datetime.now()
-        file.status = status
         if status == amo.STATUS_DISABLED:
+            file.original_status = file.status
+        else:
             file.original_status = amo.STATUS_NULL
-            file.status_disabled_reason = File.STATUS_DISABLED_REASONS.NONE
+        file.status_disabled_reason = File.STATUS_DISABLED_REASONS.NONE
```
eviljeff, line 976: I figured any time we're updating the status of a file we want to clear `status_disabled_reason` (?)

PR #23465, opened 2025-05-16
auto-resolve incoming reports where the addon version has been reviewed

eviljeff:
Fixes: mozilla/addons#15608
+resolves https://mozilla-hub.atlassian.net/browse/AMOENG-1678 about updating the email template (no github issue)

### Description

Adds automation to auto-resolve incoming reports (on criteria detailed in the issue/parent ticket/doc) as the content was already moderated by a human.

### Context

We already had some rudimentary automation, but only for removed content, so this expands on it.  Also those reports are now counted in aggregate abuse report thresholds.

### Testing

- manually approve a version of an add-on in the reviewer tools
- go to the listing page for that add-on and create an abuse report for it, for a policy violation reason in the add-on (i.e. so it would normally be resolvable by reviewers rather than Cinder).  
  - Either log in beforehand, or provide a reporter email address - we want to test the email
- Check the email in Fake mail
- Look in Cinder for the decision for that add-on (with the Ignore -> Content Already Moderated policy)
- (The add-on should not be flagged for NHR due to abuse)

Bonus testing - with a specific version
- install the version you approved in your Firefox
- upload a new version of that add-on and run auto-approve to make it the new current_version
- go to the listing page for that add-on and create an abuse report for it, for a policy violation reason in the add-on (i.e. so it would normally be resolvable by reviewers rather than Cinder).  
  - Either log in beforehand, or provide a reporter email address - we want to test the email
- Check the email in Fake mail
- Look in Cinder for the decision for that add-on (with the Ignore -> Content Already Moderated policy)
- (The add-on should not be flagged for NHR due to abuse)
- Now uninstall that add-on and go back to the listing page - you should now be reporting the current version, which wasn't human reviewed, so _shouldn't_ be auto resolved
- there should be no new email in Fake mail
- but the add-on _should_ now be flagged for NHR in the reviewer tools.

~Bonus-Bonus testing - thresholds~
- ~(I'm not super familiar with testing the threshold functionality, but it's something like:)~
- ~Create a UsageTier set to a (low) number of abuse reports~
- ~File more than &lt;the number in the UsageTier&gt; abuse reports, for policy violation in the add-on.~
- ~run `flag_high_abuse_reports_addons_according_to_review_tier` task; see the add-on isn't flagged with a NHR for high abuse reports~
- ~update the latest version of the add-on so it has a `human_review_date`~
[...]

PR #23485, opened 2025-05-22
Expose when jobs were created / forwarded / requeued in review page

diox:
Fixes https://github.com/mozilla/addons/issues/15575

### Testing

The unit test change should cover it, but a full test would look something like:

- Create a listed add-on, content-approve it, then run auto-approve to get it fully approved (*)
- Make a report against a signed version of that add-on for policy violations inside the add-on
- In the review page for that add-on, select an action that can resolve jobs (reject, resolve reports, disable add-on...)
- See the creation date for the cinder job being displayed
- Make another report for the same add-on for policy violations inside the add-on again
- In the review page, see the creation date hasn't changed since that new report it should be attached to the same job
- Make another report for the same add-on for hateful content in the add-on page itself
- In Cinder, find that job in the add-on listings queue, assign it to yourself and escalate that to reviewers
- (locally, grab the webhook request corresponding to that escalation and replay it locally)
- In the review page, notice you should have a second job displayed when selecting an action that resolves jobs, and both the creation and forward date are displayed
- Add the add-on to a high profile promoted group (don't manually approve the version)
- Reject the add-on while resolving the jobs
- See the add-on in second level approval queue
- In second level approval queue, requeue the add-on
- In the review page, see the requeue date in the jobs that can be resolved

(*) auto-approving vs manual approving doesn't change anything at the moment but it would matter [when auto-resolving reports against reviewed content](https://github.com/mozilla/addons-server/pull/23465) is merged.

### DIFF
--- a/src/olympia/reviewers/forms.py
+++ b/src/olympia/reviewers/forms.py
@@ -276,11 +276,7 @@ class CinderJobsWidget(forms.CheckboxSelectMultiple):
             *(decision.private_notes for decision in requeued_decisions),
         ]
         internal_notes = (
-            (
-                (
-                    f'Reasoning: {"; ".join(notes for notes in forwarded_or_requeued_notes)}',
-                ),
-            )
+            ((f'Reasoning: {"; ".join(forwarded_or_requeued_notes)}',),)
             if forwarded_or_requeued_notes
             else ()
         )
--- a/src/olympia/reviewers/forms.py
+++ b/src/olympia/reviewers/forms.py
@@ -21,6 +21,7 @@ from olympia import amo, ratings
 from olympia.abuse.models import CinderJob, CinderPolicy, ContentDecision
 from olympia.access import acl
 from olympia.amo.forms import AMOModelForm
+from olympia.amo.templatetags.jinja_helpers import format_datetime
 from olympia.constants.abuse import DECISION_ACTIONS
 from olympia.constants.reviewers import REVIEWER_DELAYED_REJECTION_PERIOD_DAYS_DEFAULT
 from olympia.files.utils import SafeZip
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/forms.py
```diff
@@ -268,30 +271,40 @@ def create_option(
...
                 for text, is_reporter in appeals
             ),
         ]
 
         label = format_html(
-            '{}{}{}<br/><span>{}</span>'
+            '(Created on {}) {}{}{}{}<br/><span>{}</span>'
             '<details><summary>Show detail on {} reports</summary><ul>{}</ul>'
             '</details>',
+            format_datetime(obj.created),
```
eviljeff, line 300: I would have stopped at just the date, but datetime is what Andreas specified 🤷‍♂️

src/olympia/reviewers/forms.py
```diff
@@ -250,13 +251,15 @@ def create_option(
...
-        ]
         is_appeal = obj.is_appeal
+        queue_moves = list(obj.queue_moves.order_by('-created'))
+        requeued_decisions = list(
+            obj.decisions.filter(action=DECISION_ACTIONS.AMO_REQUEUE).order_by(
+                '-created'
+            )
+        )
+        forwarded = queue_moves[0].created if queue_moves else None
+        requeued = requeued_decisions[0].created if requeued_decisions else None
```
eviljeff, line 262: Nit, but imo we don't need both forwarded and requeued datetimes - just whatever is the most recent.

eviljeff (approved):

src/olympia/reviewers/forms.py
diox: I wanted to split Forwarded from Requeued text to make things more explicit, potentially displaying both, so adding extra logic to hide the date on one of them in that case felt a bit too much. I'll keep it that way until someone complains...

PR #23515, opened 2025-05-28
Send `requires_payment` field to Cinder

diox:
Fixes https://github.com/mozilla/addons/issues/15596

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -404,6 +404,7 @@ class CinderAddon(CinderEntity):
                 self.addon.current_version.release_notes
             )
         data['privacy_policy'] = self.get_str(self.addon.privacy_policy)
+        data['requires_payment'] = self.addon.requires_payment
         # The URL/email fields can't be sent as empty strings as they would not
         # be considered valid by Cinder.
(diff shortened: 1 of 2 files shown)


KevinMind:
Can you clarify for posterity what settings specifically to set for

> Make sure your env is set to report add-ons to Cinder in your settings

You could also achieve this by adding a runbook for that and linking the PR here. That would even be preferable as future PRs needing the same setup could benefit from the info.

KevinMind:
I am also going to need help finding the job in cinder... How do I even get to a list view of the queue in cinder?

I attempted going [here](https://stage.cinder.nonprod.webservices.mozgcp.net/investigate) and typing the GUID.. got back a list of different guid addons.

I could not find a link to the queue other than a page that displays all of the queues..

KevinMind:
Verified

Screenshot 2025-06-02 at 20 30 10
Screenshot 2025-06-02 at 20 35 37

diox:
> You could also achieve this by adding a runbook for that and linking the PR here. That would even be preferable as future PRs needing the same setup could benefit from the info.

https://github.com/mozilla/addons/issues/15639

PR #23537, opened 2025-06-03
Push delayed rejection date forward after second level approval

diox:
Fixes https://github.com/mozilla/addons/issues/15584

### DIFF
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -95,7 +95,7 @@ class ContentAction:
         without further checks, and true if it should be held for further review."""
         return False
 
-    def process_action(self):
+    def process_action(self, release_hold=False):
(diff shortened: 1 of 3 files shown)


eviljeff (approved):
This is fine... but I do wonder if it wouldn't be a lot more straightforward to just update metadata[`delayed_rejection_date`] in views.decision_review - then you wouldn't need all these changes to pass `release_hold` everywhere.

diox:
> This is fine... but I do wonder if it wouldn't be a lot more straightforward to just update metadata[`delayed_rejection_date`] in views.decision_review - then you wouldn't need all these changes to pass `release_hold` everywhere.

I initially started doing that, but I figured the `ContentAction` classes are supposed to be responsible for what happens when the hold is released/the action is processed, so it felt less of a hack / less surprising that way.

PR #23538, opened 2025-06-04
Use specific, non-admin permission for 2nd level approval queue

diox:
Fixes https://github.com/mozilla/addons/issues/15601

### Testing

See issue: accessing or making decisions in 2nd level approval queue need a new permission, `Addons:HighImpactApprove`.

You can try this by doing the following:
- Pick a recommended add-on
- As a reviewer, force disable it in review page
- Create a Group with `Addons:HighImpactApprove,ReviewerTools:View` `rules` through the django admin or in a shell
- Add a user that isn't part of any other group to that group
- Log in with that user, and check that you can access the second level approval queue in reviewer tools. You should see that add-on you force-disabled.
- Proceed with the action, make sure the decision goes through and the add-on is force-disabled.

### DIFF
--- a/src/olympia/constants/permissions.py
+++ b/src/olympia/constants/permissions.py
@@ -49,6 +49,8 @@ ADDONS_RECOMMENDED_REVIEW = AclPermission('Addons', 'RecommendedReview')
 ADDONS_TRIAGE_DELAYED = AclPermission('Addons', 'TriageDelayed')
 # Can see add-ons with all due dates in the queue, rather than just upcoming ones
 ADDONS_ALL_DUE_DATES = AclPermission('Addons', 'AllDueDates')
+# Can view/make choices in 2nd level approval queue
+ADDONS_HIGH_IMPACT_APPROVE = AclPermission('Addons', 'HighImpactApprove')
 
 # Can edit all collections.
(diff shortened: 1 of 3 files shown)


KevinMind (approved):
Verified

image

KevinMind:
@diox it was a bit confusing when doing the verification because:
1) my goto user local_admin is a god user so I was able to see the review page before adding the other user to the group.
2) after adding the user and logging in I could still see the review page but like mentioned above, I already saw it so could either mean both users had the permissions or the permission is irrelevant.

A more clear STR would have me login as the not allowed user and verify I cannot see the page, then adde the group and re-check seeing that I can.

Edge case, what would we expect to happen if I had the permission to load the page, but before "proceed" button is clicked I remove the group.. maybe not an edge case we care about but probably worth verifying with a test that nothing irreversibly bad happens?

diox:
> Edge case, what would we expect to happen if I had the permission to load the page, but before "proceed" button is clicked I remove the group.. maybe not an edge case we care about but probably worth verifying with a test that nothing irreversibly bad happens?

The permission will be verified when we deal with the `POST` request - it goes to the same view. There is already a unit test proving that if you're missing the permission, it doesn't go through (`test_reviewer_viewer_cannot_submit()`)

PR #23545, opened 2025-06-05
Parse and store data collection permissions, behind a waffle switch

diox:
Fixes https://github.com/mozilla/addons/issues/15593
Fixes https://github.com/mozilla/addons/issues/15594

### Description

This adds initial support for data collection permissions in AMO at submission time, behind a `enable-data-collection-permissions` waffle switch that is disabled by default.

### Testing

- Grab the [xpi added in this PR](https://github.com/diox/olympia/blob/be7d00b582f8480e0a7fe103a96caf7ccac9089e/src/olympia/files/fixtures/files/webextension_data_collection.xpi) or create one with some `data_collection_permissions` in the manifest (under `browser_specific_settings` > `gecko`, it's an object containing 2 lists, `required` and `optional` - see https://docs.google.com/document/d/1UFVCzAr-MDrBzHNcxdv4vq8MoSIqeTKfrIXyjsLCXdI/edit?tab=t.0)
- Submit it ; if you haven't modified the waffle yet, it should be disabled, so this should fail validation
- Enable the waffle with `manage.py waffle_switch enable-data-collection-permissions on` or through the admin
- Submit the xpi again, it should pass validation
- After successful submission, look for the `File` in the admin to verify that we're storing the relevant `data_collection_permissions` and `optional_data_collection_permissions`. You can also check the `WebextPermission` manually in a shell.


willdurand:
> Grab the [xpi added in this PR](https://github.com/diox/olympia/blob/be7d00b582f8480e0a7fe103a96caf7ccac9089e/src/olympia/files/fixtures/files/webextension_data_collection.xpi) or create one with some `data_collection_permissions` in the manifest (under `browser_specific_settings` > `gecko`, it's an object containing 2 lists, `required` and `optional` - see [docs.google.com/document/d/1UFVCzAr-MDrBzHNcxdv4vq8MoSIqeTKfrIXyjsLCXdI/edit?tab=t.0](https://docs.google.com/document/d/1UFVCzAr-MDrBzHNcxdv4vq8MoSIqeTKfrIXyjsLCXdI/edit?tab=t.0))

Another example in case that's useful: https://github.com/willdurand/example-extensions/blob/main/data-collection-with-fallback/manifest.json

src/olympia/files/admin.py
```diff
@@ -20,7 +20,13 @@ def has_add_permission(self, request, obj=None):
 
 class WebextPermissionInline(admin.StackedInline):
     model = WebextPermission
-    fields = ('permissions', 'optional_permissions', 'host_permissions')
+    fields = (
+        'permissions',
+        'optional_permissions',
+        'host_permissions',
```
willdurand, line 26: Unrelated to this PR but we also have `optional_host_permissions`: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/optional_host_permissions

diox: Yeah, we don't currently support reading those in AMO - I guess that would happen in https://github.com/mozilla/addons/issues/15103 once it's triaged

src/olympia/files/migrations/0034_add_data_collection_fields.py
eviljeff, line 1: For the deployment window we need these fields to be nullable (right? And these defaults are Django defined defaults, that don't propagate to the schema sql)

diox: I wondered the same thing, but `JSONField` defaults _are_ database defaults as well. You can verify that by checking the `sqlmigrate` for that migration. This is why we didn't bother with nulls for the previous fields on that model.

```sql
--
-- Add field data_collection_permissions to webextpermission
--
ALTER TABLE `webext_permissions` ADD COLUMN `data_collection_permissions` json DEFAULT ('{}') NOT NULL;
--
-- Add field optional_data_collection_permissions to webextpermission
--
ALTER TABLE `webext_permissions` ADD COLUMN `optional_data_collection_permissions` json DEFAULT ('{}') NOT NULL;`
```

PR #23550, opened 2025-06-09
Put add-on hard-deletion for old deleted users behind a waffle switch

diox:
Fixes https://github.com/mozilla/addons/issues/15644

### Testing

Look at unit test I've added: when the switch is off (which is the default, the migration adds it as inactive) `clear_old_user_data` command will no longer hard-delete soft-deleted add-ons belonging to users deleted over 7 years ago. When it's on, it will.


src/olympia/users/management/commands/clear_old_user_data.py
```diff
@@ -57,5 +52,11 @@ def handle(self, *args, **options):
             ip_logs = IPLog.objects.filter(activity_log__user__in=users)
             ip_logs.delete()
 
-        if addon_ids:
-            delete_addons.delay(addon_ids, with_deleted=True)
+        if waffle.switch_is_active('enable-addons-hard-deletion'):
```
KevinMind, line 55: I'm curious if this is the right place to add the waffle check? This command, then runs a task, which then calls delete on the addons.

the delete method in addons already handles soft/hard deletion. so why not hook into the delete method?

Would soft deleting when the flag is off prevent future inclusion of the addons when we re-enable it?

diox: The `delete()` method of the `Addon` model doesn't really handle hard-deletion, it only does it in a very specific & narrow use-case. For this reason, the task actually bypasses that method, and calls `delete()` on the _queryset_, which by design ignores the model method, and cascades properly.

I could have added the waffle switch check in the task itself, but it's used in at least one other place where we do want to hard-delete add-ons, the `gc` cron, which hard-delete add-ons that don't have any versions in the database. This is something that can happen with the way the old (and long deprecated) signing API works, which creates an `Addon` before validation, so it could end up without a `Version` attached.

KevinMind: Could there be an argument that we **should** make the delete method handle the deletion of addons and any caller to that method should either let the delete method handle the logic for whether to hard/soft delete or should pass arguments specifying what kind of deletion is expected.

Your answer makes sense, kind of, but still (maybe even more) makes me want to see this logic centralized to the model itself?

There is a branch of the Addon.delete method that calls super.delete() (hard deletion, with cascade)

So why can't we call that delete method with the right context, or with appropriate arguments to result in a hard deletion?

Keeping the set of logic that is executed for deletion in one place makes it a lot easier to reason about how and when we delete... which is kind of the crux of the issue this PR relates to so why not solve this problem?

We have no rush to get this PR landed (at least we have many months of runway) so why rush it?

KevinMind: To put some context on this, the vision that I can see for how we handle soft/hard deletion generally is to implement ([this](https://github.com/mozilla/addons/issues/15430)) which means every model can tell you if an addon is safe to delete. We can take a similar approach to deletion where a set of arguments can control if a caller wants to hard delete a record, and this method can raise if the record is not safe to delete. A nice to have on top would be a standard mechanism for handling soft/hard deletion

Deletion can be explicit and centralized to each model. If you call Addon.delete(hard=True) we should either hard delete the record or raise if it is not safe to do so. Otherwise if you call Addon.delete() the assumption is it is a soft deletion.

I think there is a lot of room to debate about the shape of the API and exactly how we do this but generally I don't think we should move the logic for what kind of deletion is happening to various tasks, cron jobs, methods, functions, commands etc. It makes the system hard to understand and therefore hard to control. Every model should describe which records are safe to delete and should implement the logic necessary to soft or hard delete its own records. Ideally, this mechanism should be as standard as possible across models so that handling these situations follows a common set of expectations.

Wdyt?

diox: The ticket is a high priority (P1) and was evaluated as a 1 story point small super safe change we could easily get in the sprint - it was just to stop the bleeding while we re-assess policy. Doing the refactor you're suggesting changes this significantly, adding complexity/risk - I'm not opposed to doing that, it's probably a good idea, but IMHO it's out of scope for this.

KevinMind: Okay. Well this change doesn't significantly change the existing state so happy to approve so long as we have the context in mind. I don't want a small fix to make the problem seem small enough to ignore

diox: :+1: add your idea about the general mechanism (pass `hard=True` to individual object `delete()`, use it whenever we delete stuff, and add-ons in particular) in https://github.com/mozilla/addons/issues/15430 maybe ? Or should we file another issue and put that in the data retention epic ?

KevinMind (approved):

PR #23560, opened 2025-06-10
Add auto_resolve_reports command to run on reviewer tools backlog

eviljeff:
Fixes: mozilla/addons#15637

### Checklist

### DIFF
--- a/src/olympia/abuse/management/commands/auto_resolve_reports.py
+++ b/src/olympia/abuse/management/commands/auto_resolve_reports.py
@@ -1,0 +1,60 @@
+from collections import defaultdict
+
+from django.core.management.base import BaseCommand
+
+import olympia.core.logger
+from olympia.abuse.models import AbuseReport, CinderJob
+from olympia.abuse.tasks import handle_already_moderated
+from olympia.versions.models import Version
+
+
+class Command(BaseCommand):
+    log = olympia.core.logger.getLogger('z.abuse')
+
+    def handle(self, *args, **options):
+        jobs = (
+            CinderJob.objects.unresolved()
+            .resolvable_in_reviewer_tools()
+            .select_related('target_addon___current_version')
+            .prefetch_related('appealed_decisions')
+        )
+        abuse_report_dict = defaultdict(list)
+        abuse_reports = list(
+            AbuseReport.objects.filter(cinder_job__in=jobs).values_list(
+                'reason', 'addon_version', 'cinder_job_id', named=True
+            )
+        )
+        for report in abuse_reports:
+            abuse_report_dict[report.cinder_job_id].append(report)
+
+        addon_versions = list(
+            # Note this query can over-match Version objects if the version string
+            # exists in multiple add-ons (e.g. version "0.1" exists in addon-A and
+            # addon-B, but is not reported for addon-A, only for addon-B)
+            # - this is somewhat inefficient, but harmless - it just won't be looked
+            # up in CinderJob.should_auto_resolve.
+            Version.objects.filter(
+                addon__in=[job.target_addon for job in jobs],
+                version__in=[
+                    report.addon_version
+                    for report in abuse_reports
+                    if report.addon_version
+                ],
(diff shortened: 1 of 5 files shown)


src/olympia/abuse/management/commands/auto_resolve_reports.py
```diff
@@ -0,0 +1,60 @@
...
+        for report in abuse_reports:
+            abuse_report_dict[report.cinder_job_id].append(report)
+
+        addon_versions = list(
+            # Note this query can over-match Version objects if the version string
+            # exists in multiple add-ons (e.g. version "0.1" exists in addon-A and
+            # addon-B, but is not reported for addon-A, only for addon-B)
+            # - this is somewhat inefficient, but harmless - it just won't be looked
+            # up in CinderJob.should_auto_resolve.
+            Version.objects.filter(
```
diox: I assume we want deleted versions too, so that should be `unfiltered` instead of `objects`.

src/olympia/abuse/management/commands/auto_resolve_reports.py
```diff
@@ -0,0 +1,60 @@
...
+                addon__in=[job.target_addon for job in jobs],
+                version__in=[
+                    report.addon_version
+                    for report in abuse_reports
+                    if report.addon_version
+                ],
+            ).values_list('addon_id', 'version', 'human_review_date', named=True)
```
diox: probably also `.order_by()` to speed it up as you don't need ordering.

src/olympia/abuse/management/commands/auto_resolve_reports.py
```diff
@@ -0,0 +1,60 @@
...
+            # addon-B, but is not reported for addon-A, only for addon-B)
+            # - this is somewhat inefficient, but harmless - it just won't be looked
+            # up in CinderJob.should_auto_resolve.
+            Version.objects.filter(
+                addon__in=[job.target_addon for job in jobs],
+                version__in=[
+                    report.addon_version
+                    for report in abuse_reports
+                    if report.addon_version
+                ],
```
diox: At a minimum this should be a set, not a list of versions.

I tried running the `addon_versions` queryset in prod. The `version__in` part causes the query to take forever to execute, I had to abort it after 5 minutes. There are a lot of reports... I wasn't running 100% the exact same thing, I had to use subqueries for `addon__in` and `version__in`, but still, I'm concerned it would not execute properly.

It might be better to grab affected versions individually for each job in the loop later rather than trying to prefetch them all in one go...

src/olympia/abuse/management/commands/auto_resolve_reports.py
```diff
@@ -0,0 +1,60 @@
...
+            job._abusereports = abuse_report_dict.get(job.id, [])
+            job._addon_versions = addon_version_dict.get(job.target_addon_id, [])
+            if job.should_auto_resolve():
+                # if it should be auto resolved, fire a task to resolve it
+                self.log.info(
+                    'Found job#%s to auto resolve for addon#%s.',
+                    job.id,
+                    job.target_addon_id,
+                )
+                handle_already_moderated.delay(job_pk=job.id)
```
diox: Given the issues I had in prod executing the `addon_versions` queryset I'm not convinced the `_abusereports` / `_addon_versions` hacks are worth it.

Couldn't you extract this in its own task, fire it for each job, and let it execute the normal queries `should_auto_resolve()` does ? It would be slightly inefficient because we're be re-running the same queries multiple times across different tasks yes, but that would probably scale better than trying to run a huge query to grab all versions in one go. The whole patch would be a lot simpler, no need to fetch versions and reports in the command.

diox (changes requested):

src/olympia/abuse/management/commands/auto_resolve_reports.py
eviljeff: okay, all the optimisation dropped 😢 - it's now a much more simple task that executes as it would for a newly reported version or a forward from Cinder.

diox (approved):

PR #23562, opened 2025-06-11
Reset user sessions when users are banned

diox:
Fixes https://github.com/mozilla/addons/issues/15649

### Context

`auth_id` is used by `get_session_auth_hash()` to invalidate sessions. If set to any value different to the one the session had, the session won't be valid.

### Description

This patches ensures we clear `auth_id` on banning, but also, just in case, makes `get_session_auth_hash()` return `None` for any deleted user, no matter what `auth_id` value is. Since `auth_id` defaults to a non-null value, this effectively ensures past sessions are never valid for deleted users (which banned users are a subset of)

### Testing

See STR in the issue:
- Create a user
- In a [private window, other browser, other container] log in as an admin and ban that user
- Go back to the original user, refresh, you should have been logged out

### DIFF
--- a/src/olympia/users/models.py
+++ b/src/olympia/users/models.py
@@ -236,7 +236,9 @@ class UserQuerySet(BaseQuerySet):
             # To delete their photo, avoid delete_picture() that updates
             # picture_type immediately.
             delete_photo.delay(user.pk, banned=user.banned)
-        return self.bulk_update(users, fields=('auth_id', 'banned', 'deleted', 'modified'))
+        return self.bulk_update(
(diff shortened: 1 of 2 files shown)


KevinMind:
Would be good to have a follow up adding a runbook for banning/unbanning a user.

KevinMind:
I actually don't know how to ban a user.. could you add the runbook? Or describe it and I'll add it?

diox:
There is a ban (and unban) action available in the [user admin](http://olympia.test/en-US/admin/models/users/userprofile/) (can be done in bulk from the changelist or on a specific user profile)

PR #23565, opened 2025-06-11
define Config keys in constants, with safeguards

eviljeff:
Fixes: mozilla/addons#1932

### Description
Replace our config keys (and defaults, and types) with constants.  A call to `get_config` or `set_config` must use a valid config name from the constants; but because the constants are tuples it's still possible to override `get_config` to provide an arbitrary default (or type) - though we don't currently.

### Context

This... was supposed to be a quick and easy fix of an old code quality issue and ended up more complicated than it would have seemed - mainly because, to me, it didn't make sense to centralize the config _names_ while leaving defaults defined somewhere else, and the type of the config unenforced, and needing to specify it each time you get or set it. wdyt?  (directly requesting review as the code is all very inside baseball 😬)

### Testing

- test somewhere we set or get config, I guess? reviewer MOTD is an easy candidate; the blocklist code is a more complicated test.

### Checklist

### DIFF
--- a/src/olympia/blocklist/utils.py
+++ b/src/olympia/blocklist/utils.py
@@ -289,3 +289,16 @@ def delete_versions_from_blocks(guids, submission):
             block.delete()
 
     return blocks
+
+
+def get_mlbf_base_id_config_key(block_type, compat: bool = False):
+    """
+    We use compat to return the old singular config key for hard blocks.
+    This allows us to migrate writes to the new plural key right now without
+    losing access to the old existing key when deploying this patch.
+    """
+    from olympia.blocklist.models import BlockType
+
+    if compat and block_type == BlockType.BLOCKED:
+        return amo.config_keys.BLOCKLIST_MLBF_BASE_ID
+    return getattr(amo.config_keys, f'BLOCKLIST_MLBF_BASE_ID_{block_type.name.upper()}')
--- a/src/olympia/constants/config_keys.py
+++ b/src/olympia/constants/config_keys.py
@@ -1,0 +1,49 @@
+import json
+
+
+"""
+Keys are define either as a tuple of (str, value-type, default) or a str.
+`value-type` currently supported in get_config are int, json, and str (the default).
+The default `default` is None.
+"""
+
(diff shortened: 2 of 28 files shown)


src/olympia/constants/config_keys.py
```diff
@@ -0,0 +1,49 @@
+import json
```
eviljeff, line 1: is using the `json` package as a type, madness? 🤷

eviljeff: I guess I could have used  strings, i.e. `"json"`, but it _looks_ more natural ?

diox: I think it's too clever for its own good. Using `is` when comparing it is also extra-dangerous (it works, and in this specific super limited scope it's probably fine now, but there is no guarantee it won't misbehave if we start adding more weird types, or if python import module mechanism changes).

It's more verbose, but maybe you could have something like:
```python
class ConfigKey:
  type_ = str
  def load(self, value):
    return self.type_(value)
  def dump(self, value):
    return self.type_(value)

class IntConfigKey(ConfigKey):
  type_ = int

class JsonConfigKey(ConfigKey):
  type_ = None  # Don't use
  def load(self, value):
    return json.loads(value)
  def dump(self, value):
    return json.dumps(value)
```

Then you wouldn't need to check the type at all in the `get_config`/`set_config` code.

src/olympia/blocklist/utils.py
```diff
@@ -289,3 +289,16 @@ def delete_versions_from_blocks(guids, submission):
             block.delete()
 
     return blocks
+
+
+def get_mlbf_base_id_config_key(block_type, compat: bool = False):
```
KevinMind, line 294: Could we migrate to the new key since we no longer need the old one? That was supposed to happen after soft blocking launched. Could be a follow up but since you're down there?

eviljeff: Is there a ticket that explains how the migration is supposed to work?

src/olympia/constants/config_keys.py
eviljeff: https://github.com/mozilla/addons-server/pull/23565/commits/4686af7df4a8650ef852064b1a41076ea71ca40b

diox (approved):

PR #23566, opened 2025-06-12
Lock down github action permissions

KevinMind:
Fixes: mozilla/addons#15312

### Description

- [ ] Lock down top level workflow permissions to {} (empty)
- [ ] Add permissions back until all jobs pass
- [ ] Add linting for security holes in GHA workflows

### Context

By locking down default permissions and adding them back on a job or even step level gives us the most granular and limited by default security model for GHA.

Ensuring a step/job/workflow can only do what it intends and should be allowed to do.

### Testing

Tests Green

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event_name}}-${{ github.ref}}
   cancel-in-progress: true
 
+permissions: {}
+
 env:
   docs_artifact: docs
 
--- a/.github/workflows/_health_check.yml
+++ b/.github/workflows/_health_check.yml
@@ -4,6 +4,8 @@ run-name: |
   ref: ${{ github.ref_name }} |
   environment: ${{ inputs.environment }} |
 
+permissions: {}
+
 on:
   workflow_call:
     inputs:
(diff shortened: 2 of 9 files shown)


KevinMind:
Stable.

.github/workflows/ci.yml
```diff
@@ -240,21 +257,85 @@ jobs:
...
+      #     slack_channel: ${{ vars.SLACK_ADDONS_PRODUCTION_CHANNEL }}
+      #     emoji: ${{ steps.emoji.outputs.emoji }}
+      #     actor: ${{ env.actor }}
+      #     conclusion: ${{ matrix.result }}
+      #     workflow_id: ${{ env.workflow_id }}
+      #     workflow_url: ${{ env.workflow_url }}
+      #     event: ${{ env.event }}
+      #     env: ${{ env.env }}
+      #     ref: ${{ matrix.ref }}
+      #     ref_link: ${{ matrix.ref_link }}
```
diox: Why is this block commented?

KevinMind: I wanted to prevent actually sending any notifications. I will unblock now, though this won't run until after the PR lands.. technically possible to switch things up so that it would but it could trigger a dev deployment and probably not worth it.

KevinMind: Added dry_run so we can see what is passed to the action without actually sending a notification in slack. working on the latest push of this branch

diox (approved):

PR #23581, opened 2025-06-13
Expose data collection permissions in API

diox:
Fixes https://github.com/mozilla/addons/issues/15620

### Testing

- Follow instructions in https://github.com/mozilla/addons-server/pull/23545 to upload an add-on with data collection permissions, but make sure to submit it as listed
- Approve that add-on
- Look at its detail API (`http://olympia.test/api/v5/addons/addon/<pk_or_slug_or_guid>/`) and search API (`http://olympia.test/api/v5/addons/search/?guid=<guid>`) and check `data_collection_permissions` and `optional_data_collection_permissions` are there and correct

Bonus:
- Check search API still works for other add-ons you already had (the 2 data collection properties should be an empty array)

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -103,6 +103,10 @@ class FileSerializer(AMOModelSerializer):
     permissions = serializers.ListField(child=serializers.CharField())
     optional_permissions = serializers.ListField(child=serializers.CharField())
     host_permissions = serializers.ListField(child=serializers.CharField())
+    data_collection_permissions = serializers.ListField(child=serializers.CharField())
+    optional_data_collection_permissions = serializers.ListField(
+        child=serializers.CharField()
+    )
     is_restart_required = serializers.SerializerMethodField()
     is_webextension = serializers.SerializerMethodField()
 
@@ -122,6 +126,8 @@ class FileSerializer(AMOModelSerializer):
             'permissions',
             'optional_permissions',
             'host_permissions',
+            'data_collection_permissions',
+            'optional_data_collection_permissions',
         )
 
     def get_url(self, obj):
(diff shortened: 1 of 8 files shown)


src/olympia/addons/serializers.py
```diff
@@ -158,6 +164,8 @@ class LanguageToolFileSerializer(FileSerializer):
     permissions = serializers.SerializerMethodField()
     optional_permissions = serializers.SerializerMethodField()
     host_permissions = serializers.SerializerMethodField()
+    data_collection_permissions = serializers.SerializerMethodField()
```
eviljeff, line 167: Maybe we shouldn't collect them from langpacks in the first place, then we wouldn't need the special magic in the serializer (equally applies to the existing permission property too)

src/olympia/addons/serializers.py
```diff
@@ -103,6 +103,10 @@ class FileSerializer(AMOModelSerializer):
     permissions = serializers.ListField(child=serializers.CharField())
     optional_permissions = serializers.ListField(child=serializers.CharField())
     host_permissions = serializers.ListField(child=serializers.CharField())
+    data_collection_permissions = serializers.ListField(child=serializers.CharField())
+    optional_data_collection_permissions = serializers.ListField(
+        child=serializers.CharField()
+    )
```
eviljeff, line 109: I know it doesn't matter a great deal...  But these extra properties should be behind an API gate

src/olympia/addons/serializers.py
diox: The linter will prevent them from being in langpack manifests - this is just for us to avoid the queries when calling the API.

src/olympia/addons/serializers.py
diox: Should they ? It's just new properties so I don't really care. We didn't bother when we added `host_permissions`.

PR #23583, opened 2025-06-16
Don't add NHR for a version we couldn't auto-approve if it's waiting for 2nd level approval

diox:
Fixes mozilla/addons#15618

### Context

Before this patch, each time `auto_approve` is attempted for an add-on with auto-approval disabled or belonging to a promoted group that should prevent auto-approval, if the version was not pending rejection we'd add a NHR, unless there was already one. That in turn is what causes such add-ons to land in the manual review queue.

### Description

With that change, we also look if the version has a decision waiting for second level approval as well before attempting to add the NHR. If the version in the second level approval queue, that means someone already looked at it, we don't need to add it to the manual review queue again.

### Testing

- Submit a new version of a recommended or notable add-on with a listed version
- Run `auto_approve`, that version should get a new active NHR
- Go to the review page for that add-on and reject all versions, including that new one awaiting review. Note that the new version was marked as needing human review (orange background, has a due date)
- Since the add-on is part of a high-profile promoted group, that decision should be held for second level approval
- Leave the review page and run `auto_approve` again

Expected results:
- The add-on shouldn't get a new NHR

### DIFF
--- a/src/olympia/reviewers/management/commands/auto_approve.py
+++ b/src/olympia/reviewers/management/commands/auto_approve.py
@@ -218,11 +218,15 @@ class Command(BaseCommand):
         # could leave the review queue and re-enter it multiple times.
         for key in verdicts_to_reasons:
             reason = verdicts_to_reasons[key]
-            if not version.pending_rejection and (
-                getattr(version.autoapprovalsummary, key)
-                and not version.needshumanreview_set.filter(
-                    reason=reason, is_active=True
-                ).exists()
+            if (
+                not version.pending_rejection
+                and not version.contentdecision_set.awaiting_action().exists()
+                and (
+                    getattr(version.autoapprovalsummary, key)
+                    and not version.needshumanreview_set.filter(
+                        reason=reason, is_active=True
+                    ).exists()
+                )
             ):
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/management/commands/auto_approve.py
```diff
@@ -218,11 +218,15 @@ def disapprove(self, version):
         # could leave the review queue and re-enter it multiple times.
         for key in verdicts_to_reasons:
             reason = verdicts_to_reasons[key]
-            if not version.pending_rejection and (
-                getattr(version.autoapprovalsummary, key)
-                and not version.needshumanreview_set.filter(
-                    reason=reason, is_active=True
-                ).exists()
+            if (
```
KevinMind, line 230: Nit: This if statement is a bit too much to read, 3 negatives and a positive, not totally clear what everything means in the context without digging further. 1st one is clear enough, other 3 would be easier to follow as named variables.

Bonus points if the direction of all the binary checks could be one way but that might be harder.

diox: 85990f4a7cedc15108b672d21c611a8aedec5440

PR #23589, opened 2025-06-16
Only hold version rejections for 2nd level approval if they would lead to the listing being unavailable

diox:
Fixes: mozilla/addons#15568

### Context

Previously, we would hold for 2nd level approval any decision containing signed listed versions from a high-profile add-on. Now, we only hold the decision if it would lead to the add-on listing to become unavailable, essentially.

### Testing

- Using a recommended add-on with multiple public listed versions:
  - Try to reject all its already public listed versions: that decision should be held for 2nd level approval
  - Try to delay-reject all its already public listed versions except one: that decision should not be held of 2nd level approval, it should go through
    - From that state, try to delay-reject or plain reject the last public listed version you left out previously: that decision should be held for 2nd level approval, because all other versions, despite being still public for the moment, are pending rejection.

### DIFF
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -384,8 +384,25 @@ class ContentActionRejectVersion(ContentActionDisableAddon):
     def should_hold_action(self):
         return (
             super().should_hold_action()
-            and self.target_versions.filter(file__is_signed=True).exists()
+            # Only hold rejections for 2nd level approval if...
+            # At least one of the version was listed and signed already...
+            and self.target_versions.filter(
+                channel=amo.CHANNEL_LISTED, file__is_signed=True
+            ).exists()
+            # And no remaining public listed versions exists (pending rejection
+            # don't count) - i.e., going through with the rejection would make
+            # the listing go away.
+            and not self.remaining_public_listed_versions()
+            .exclude(reviewerflags__pending_rejection__isnull=False)
+            .exists()
         )
+
+    def remaining_public_listed_versions(self):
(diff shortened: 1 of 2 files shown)


src/olympia/abuse/actions.py
```diff
@@ -384,9 +384,26 @@ def __init__(self, decision):
...
+            # Only hold rejections for 2nd level approval if...
+            # At least one of the version was listed and signed already...
+            and self.target_versions.filter(
+                channel=amo.CHANNEL_LISTED, file__is_signed=True
+            ).exists()
+            # And no remaining public listed versions exists (pending rejection
+            # don't count) - i.e., going through with the rejection would make
+            # the listing go away.
+            and not self.remaining_public_listed_versions()
+            .exclude(reviewerflags__pending_rejection__isnull=False)
```
eviljeff, line 396: interesting - should we exclude pending rejection versions in general, I wonder?  (i.e. add the exclusion to `remaining_public_listed_versions` so it affects the stakeholder email too?)

eviljeff (approved):
Plus I found a bug https://github.com/mozilla/addons/issues/15668

src/olympia/abuse/actions.py
diox: It was an explicit requirement for this change, but I wondered the same thing about the stakeholder email and... I left it out of that because I thought it could be more confusing depending on what the delays are. On the other hand I see the argument that it would make things more consistent.

diox: Asked on slack and @wagnerand said this was fine, at least for now.

PR #23603, opened 2025-06-19
exclude 2nd level approval requeued jobs from being auto resolved

eviljeff:
Fixes: mozilla/addons#15671

### Description

Changes filtering in command to only process jobs that have no decisions (meaning we won't try to process any jobs that have had a decision that was subsequently held or overridden)

### Testing

A little tricky to set up, because we're testing for a situation that wouldn't occur (the relevant jobs would already have been closed).
- select an add-on that has a public listing - the current version should **not** have a human review date, but **should** be `is_signed=True`
  - use django admin to tinker with the version/file
- make the add-on promoted in django admin
- report an add-on for abuse for in the add-on, for a policy violation.
- check the job is open and the add-on is flagged for review in the review tools
- reject all the versions of the add-on while resolving the open job
  - it should be held for 2nd level approval as it's promoted; a signed version is being rejected; and the add-on will no longer be public.
- run `auto_resolve_reports` command - the job shouldn't be closed

### Checklist

### DIFF
--- a/src/olympia/abuse/management/commands/auto_resolve_reports.py
+++ b/src/olympia/abuse/management/commands/auto_resolve_reports.py
@@ -10,7 +10,9 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         job_ids = (
-            CinderJob.objects.unresolved().resolvable_in_reviewer_tools()
+            CinderJob.objects.filter(
+                decisions__isnull=True
+            ).resolvable_in_reviewer_tools()
         ).values_list('id', flat=True)
         self.stdout.write(f'{len(job_ids)} unresolved CinderJobs in reviewer tools')
 
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -216,6 +216,7 @@ class CinderJob(ModelBase):
             action_date=datetime.now(),
             reviewer_user_id=settings.TASK_USER_ID,
             cinder_job=self,
+            override_of=self.final_decision,
         )
         decision.policies.set(
             CinderPolicy.objects.filter(
--- a/src/olympia/abuse/tests/test_commands.py
+++ b/src/olympia/abuse/tests/test_commands.py
@@ -127,6 +127,7 @@ class TestAutoResolveReports(TestCase):
         )
         CinderQueueMove.objects.create(cinder_job=job_forwarded)
 
+        # the following jobs shouldn't be auto resolved
         addon2 = addon_factory(version_kw={'human_review_date': datetime.now()})
         job_legal_reason = CinderJob.objects.create(
             job_id='legal', resolvable_in_reviewer_tools=True, target_addon=addon2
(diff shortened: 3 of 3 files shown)


src/olympia/abuse/management/commands/auto_resolve_reports.py
```diff
@@ -10,7 +10,9 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         job_ids = (
-            CinderJob.objects.unresolved().resolvable_in_reviewer_tools()
+            CinderJob.objects.filter(
```
KevinMind, line 13: I don't know about this but it seems like the reason you switched from using unresolved() to plain objects is because unresolved currently filters for requeued jobs, which we explicitly want to exclude.

Did you consider adding a an argument to unresolved "include_requeued" and letting the queryset method still handle the logic?

I think there is precedent for this in the code and would keep the central condition localized to the model (good thing)

But maybe there is some other consideration that would make that a bad idea.. thoughts?

KevinMind:
```python
CinderJob.unresolved(
exclude_requeued=True
).resolvable_in_reviewer_tools()
```

To me at least reads much more clearly to me

KevinMind: Copilot agrees https://github.com/mozilla/addons-server/pull/23603#discussion_r2161126053

eviljeff: It's a one-time command to catch-up on jobs/reports, and it's the only place we need to do this.  If this was going to be used more than once, or we need to filter out held-decisions elsewhere I'd agree with you, but it's not worth the extra complexity in this case.

> Copilot agrees [#23603 (comment)](https://github.com/mozilla/addons-server/pull/23603#discussion_r2161126053)

That's not a plus point in my book 😛

KevinMind: Don't hate on the robots just because they are juniors (for now).

KevinMind: Doesn't the fact that your modifying the existing command preclude it from being a one time command? Or did this ship and then we found issues before actually running it? 

I'm also not convinced that we should have a lower bar for "temporary" things because once it merges it could be there for 15 years ** points to numerous random files in the repo that are just there ** I won't press the point so we can get this in the tag but generally I don't think that's a good rationale for this kind of change.

**Note to the void**: we should have some nomenclature for temporary commands or something like `Temp<MyCommand>` so it is clear.

If it is in fact temporary, could you link the issue that aims to remove this code so we already know we should clean it up after running it?

KevinMind:
Really? 😆 Now you're going to make the little guy feel bad. He's just trying to help.

image

src/olympia/abuse/management/commands/auto_resolve_reports.py
eviljeff: I ran it on dev and found a bug that needs to be fixed before it's ran on (stage and) prod - it's expected to be run once per env.  I'm pretty sure it's temporary because the other issues in the epic make it redundant - if was run again it wouldn't do anything because all the possible job candidates would have been closed when they occured (on report, or on forward from Cinder)

src/olympia/abuse/models.py
```diff
@@ -216,6 +216,7 @@ def handle_already_moderated(self, abuse_report, entity_helper):
             action_date=datetime.now(),
             reviewer_user_id=settings.TASK_USER_ID,
             cinder_job=self,
+            override_of=self.final_decision,
```
diox, line 219: What scenario is this covering ? When are we calling `handle_already_moderated()` with a previous decision already made ? Also this isn't covered by tests ?

diox (approved):

src/olympia/abuse/models.py
eviljeff: I can't think of a possible scenario where a job would already have a decision - all scenarios would close the job in Cinder first, so a new report would be a new job. 

If the PR for https://github.com/mozilla/addons/issues/15640 is finished and is merged (and it may not) then in that case there could be an existing decision.  If you like I can drop this line.

PR #23606, opened 2025-06-19
Refactor CI workflow: Remove unnecessary Git step and update job names and conditions for builds

KevinMind:
Fixes: mozilla/addons#15312

### Description

- Use the gar_push job to define the result of the build
- use event specific triggers for the notification so we don't get duplicate notifications on a push to gar
- remove git message entirely (slack just cannot deal with all of the myriad of characters that could be in a commit message

### Context

Before

image

After

image

### Testing

Take the dry run output in the notification job and paste it into slack UI https://app.slack.com/block-kit-builder/T027LFU12

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,25 +38,46 @@ jobs:
       is_release_tag: ${{ steps.context.outputs.is_release_tag }}
       docker_version: ${{ steps.context.outputs.docker_version }}
 
-      short_sha: ${{ steps.git.outputs.short_sha }}
-      short_message: ${{ steps.git.outputs.short_message }}
+      git_ref: ${{ steps.git.outputs.git_ref }}
+      git_ref_link: ${{ steps.git.outputs.git_ref_link }}
 
     steps:
       - name: Set context
         id: context
         uses: mozilla/addons/.github/actions/context@f1d4daa008d908d52815aa41257db39b8cdef958
 
-      - name: Git
+      - name: Git Reference
         id: git
         env:
           sha: ${{ github.sha }}
-          message: ${{ github.event.head_commit.message}}
+          event_name: ${{ github.event_name }}
+          pr_title: ${{ github.event.pull_request.title }}
+          pr_link: ${{ github.event.pull_request.html_url}}
+          push_title: ${{ github.event.head_commit.message }}
+          push_link: ${{ github.event.head_commit.url }}
+          release_title: ${{ github.event.release.name }}
+          release_link: ${{ github.event.release.html_url }}
+          git_url: ${{ github.server_url }}/${{ github.repository }}
         run: |
           short_sha=$(echo "${sha}" | cut -c1-7)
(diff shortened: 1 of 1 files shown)


diox:
> remove git message entirely (slack just cannot deal with all of the myriad of characters that could be in a commit message

Why ? This worked fine before, we had the first line of the commit message and the link to the commit, both are very useful.

KevinMind:
> > remove git message entirely (slack just cannot deal with all of the myriad of characters that could be in a commit message
> 
> Why ? This worked fine before, we had the first line of the commit message and the link to the commit, both are very useful.

We had it via a property `display_title` that was on the ci_complete workflow. workflows that run from other workflows are considered "fundamentally unsafe" and so that's why I removed that one in the scope of securing our workflows.

Now that we don't use that workflow trigger, we have to rely on the metadata available on the push event, which has different data like the full commit.

Sucks I know. There is probably a better solution but I didn't want to bloat the scope.

diox:
Instead of trying to truncate the message, taking the first line should be enough though, no ? I think trying to fit the entire message by deleting newlines is the problem here, instead we should just do something like `head -n 1` to only keep the first line.

KevinMind:
> > remove git message entirely (slack just cannot deal with all of the myriad of characters that could be in a commit message
> 
> Why ? This worked fine before, we had the first line of the commit message and the link to the commit, both are very useful.

We had it via a property `display_title` that was on the ci_complete workflow. workflows that run from other workflows are considered "fundamentally unsafe" and so that's why I removed that one in the scope of securing our workflows.

Now that we don't use that workflow trigger, we have to rely on the metadata available on the push event, which has different data like the full commit.

Sucks I know. There is probably a better solution but I didn't want to bloat the scope.

KevinMind:
> I think trying to fit the entire message by deleting newlines is the problem here, instead we should just do something like head -n 1 to only keep the first line.

I'll give this a shot.

KevinMind:
@diox It's tough to test that approach but can be done manually by feeding the commit message from a random push to master through the bash script and pasting the result into the slack ui playground.

Example:

```bash
kmeinhardt@mac addons-server % bash

The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
bash-3.2$ echo -e "${git_ref}" | head -n1 | tr -dc '[:print:]'
bash-3.2$ sha="7243c58e1b8a48cff2785de86745774ed38a82e3"
bash-3.2$ git_ref="Use commit sha as ref because commit message is too unpredictable (#23602)\n\n* Use commit sha as ref because commit message is too unpredictable\n\n* Fix permissions for locales job to push"
bash-3.2$ git_ref_link="https://github.com/mozilla/addons-server/commit/7243c58e1b8a48cff2785de86745774ed38a82e3"
bash-3.2$ git_ref=$(echo -e "${git_ref}" | head -n1 | tr -dc '[:print:]')
bash-3.2$ echo $git_ref
Use commit sha as ref because commit message is too unpredictable (#23602)
bash-3.2$ 
```

Result

image

PR #23608, opened 2025-06-20
send email to 2nd level approvers when something enters the queue

eviljeff:
Fixes: mozilla/addons#15602

### Checklist


KevinMind:
This is the 3rd PR that requires "2nd level approval" as a part of the testing process in my working memory (last 1-2 weeks)

Could you:
1) add a runbook to this PR with the rough steps for triggering a 2nd level approval flow
2) do 1 but in a follow up PR.

I think it should be in this PR since we don't need to track documentation for a feature in a separate bug (we even have a section in the PR checklist that docs are updated) but if you feel strongly to not do it in a follow up then would you open the ticket?

KevinMind:
> use django admin to tinker with the version/file

What specific changes should I make, or what does tinker mean in this context?

eviljeff:
> > use django admin to tinker with the version/file
> 
> What specific changes should I make, or what does tinker mean in this context?

As per the parent point
> the current version should should be **is_signed=True**

KevinMind:
So "use django admin to tinker with the version/file" is saying I should use django admin to ensure the version/file is_signed is equal to "true" if it is not already, or that I should just "tinker with the version/file" It's still not clear if it is a requirement that the version/file should "change" as a part of the scope or it the requirement is just that it has the specific state.

eviljeff:
> So "use django admin to tinker with the version/file" is saying I should use django admin to ensure the version/file is_signed is equal to "true" if it is not already, or that I should just "tinker with the version/file" It's still not clear if it is a requirement that the version/file should "change" as a part of the scope or it the requirement is just that it has the specific state.

the former - the version should be is_signed=True

eviljeff:
> This is the 3rd PR that requires "2nd level approval" as a part of the testing process in my working memory (last 1-2 weeks)
> 
> Could you:
> 
>     1. add a runbook to this PR with the rough steps for triggering a 2nd level approval flow

https://github.com/mozilla/addons-server/pull/23608/commits/b31df313b8b814c843aa67f9ab77bbb5abfb4cc0

KevinMind:
Got to the last step and hit a set back. I see the 2nd level approval in the queue, but no email.

Screenshot 2025-06-24 at 22 32 51
Screenshot 2025-06-24 at 22 31 34
Screenshot 2025-06-24 at 22 24 29
Screenshot 2025-06-24 at 22 24 14

Can confirm I'm in the right group. Does it matter if the user getting notified and the user taking the action are the same? Shouldn't right?

Screenshot 2025-06-24 at 22 35 38
Screenshot 2025-06-24 at 22 35 25

eviljeff:
> Can confirm I'm in the right group. 

Looks like it.

> Does it matter if the user getting notified and the user taking the action are the same? Shouldn't right?

It doesn't matter, and I've used the same account/user too.

The setup for the 2nd level approval must have worked because the version history in the screenshot shows that the decision was held, so I've just followed the instructions, starting with creating a group:
Screenshot 2025-06-25 102151

and got an email:
Screenshot 2025-06-25 102017

So I can't account for why you don't.  (I'm assuming you don't have something really obscure set, like `SEND_REAL_EMAIL=True` in your settings, or that email address added to a Group named `'Force Send Mail'`)

PR #23610, opened 2025-06-20
Update Makefile-docker

eviljeff:
I was aiming for the quick python only format command to be not super long to type 😬

### DIFF
--- a/Makefile-docker
+++ b/Makefile-docker
@@ -175,17 +175,20 @@ run_js_tests: ## Run the JavaScript test suite (requires compiled/compressed ass
 watch_js_tests: ## Run+watch the JavaScript test suite (requires compiled/compressed assets).
 	npm run test:watch
 
-.PHONY: formatjs
-formatjs: ## Autoformat our js and css files
+.PHONY: format_js
+format_js: ## Autoformat our js and css files
 	npm run fix
 
-.PHONY: formatpy
-formatpy: ## Autoformat our python files.
+.PHONY: format_py
+format_py: ## Autoformat our python files.
 	ruff check --fix-only .
 	ruff format .
 
+.PHONY: ruff
+ruff: format_py
+
 .PHONY: format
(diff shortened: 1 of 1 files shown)


Makefile-docker
```diff
@@ -175,12 +175,18 @@ run_js_tests: ## Run the JavaScript test suite (requires compiled/compressed ass
 watch_js_tests: ## Run+watch the JavaScript test suite (requires compiled/compressed assets).
 	npm run test:watch
 
-.PHONY: format
-format: ## Autoformat our codebase.
+.PHONY: formatjs
```
KevinMind: Nit: I would prefer format_js and format_py (makes it more clear they are subsets of format and helps distinguish between them.

Makefile-docker
```diff
@@ -175,12 +175,18 @@ run_js_tests: ## Run the JavaScript test suite (requires compiled/compressed ass
...
+formatjs: ## Autoformat our js and css files
 	npm run fix
+
+.PHONY: formatpy
+formatpy: ## Autoformat our python files.
 	ruff check --fix-only .
 	ruff format .
 
+.PHONY: format
+format: formatjs formatpy ## Autoformat our codebase.
```
KevinMind: you should reverse the order so formatpy runs first. It is faster and more relevant and if it fails will just bail the whole process.

KevinMind (approved):
Few nits, but mostly good. I would really like to see the `_` in the name and for speed sake better to reverse the order on format.

Makefile-docker
eviljeff: that was the order in `format`, but if there wasn't some logic behind it, then sure.

Makefile-docker
eviljeff: I was aiming for the minimum amount of extra keypresses 😉 - but I guess two more isn't too terrible...

diox (approved):

PR #23623, opened 2025-06-26
Fix overrides of negative reviewer/moderator actions for add-ons

diox:
Fixes https://github.com/mozilla/addons/issues/15653

This fixes overrides of negative actions but also remove the ability to override a decision about an add-on in second level approval (they remain available for other kind of content) - you have to use the review page (or Cinder) to make a new decision about add-ons.

### Testing

This is the full testing QA should do. I don't expect a reviewer to do all of that.

#### Overrides in general
- Report an [add-on, collection, user, rating] publicly available on AMO frontend for a reason like hateful content. Avoid content that would trigger 2nd level approval. For add-ons, specify that the location is on the page and not inside the add-on and don't pick a promoted one. For users/collections/ratings, pick content not tied to staff/developers of promoted add-ons.
- Find the report in Cinder, then disable the content
- Make sure the content has been disabled
- Using decisions tab or investigate feature in Cinder to find that decision, override it and approve it instead
- Make sure the content is back up

#### Second level approval of non add-ons
- Report a [collection, user, rating] publicly available on AMO frontend for a reason like hateful content. We want to trigger 2nd level approval, so when picking/creating the relevant content, the collection should be owned by the TASK_USER_ID user, the user should be a staff user or someone who is an author of add-ons belonging to a high-profile promoted group, the rating should be a developer reply on an add-ons belonging to a high-profile promoted group.
- Find the report in Cinder, then disable the content
- Make sure the decision is held for 2nd level approval
- In second level approval, find that decision and note that you can either approve the content instead of proceeding with the decision
- Test that both options work

#### Second level approval of add-ons
- Report an add-on publicly available on AMO for violating policies in the add-on. Pick an add-on that would trigger 2nd level approval, such as a recommended add-on.
- In reviewer tools, disable the add-on, or reject its versions
- Make sure that decision is held for 2nd level approval
- In second level approval, find that decision and note that you can no longer approve the content instead, but only proceed with the decision or enqueue the add-on back in reviewer tools.
- Test that both options work

### DIFF
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -761,7 +761,7 @@ class ContentActionTargetAppealApprove(
             elif self.previous_decisions.filter(
                 action=DECISION_ACTIONS.AMO_REJECT_VERSION_WARNING_ADDON
             ).exists():
-                for version in self.data['versions']:
+                for version in target_versions:
                     VersionReviewerFlags.objects.update_or_create(
                         version=version,
                         defaults={
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -714,25 +714,30 @@ class ContentActionTargetAppealApprove(
             qs = self.decision.target_versions.all()
         return qs
 
+    @property
+    def previous_decisions(self):
+        """Queryset with previous decisions made that this action would revert."""
+        return self.decision.cinder_job.appealed_decisions
+
     def process_action(self, release_hold=False):
         target = self.target
         log_entry = None
         if isinstance(target, Addon):
-            appealed = self.decision.cinder_job.appealed_decisions
             target_versions = (
                 self.target_versions.no_transforms()
                 .only('pk', 'version')
                 .order_by('-pk')
             )
-            if target.status == amo.STATUS_DISABLED and appealed.filter(
+            if self.previous_decisions.filter(
                 action=DECISION_ACTIONS.AMO_DISABLE_ADDON
-            ):
-                # it was a disable action that was appealed
+            ).exists():
                 target_versions = list(target_versions)
-                target.force_enable(skip_activity_log=True)
+                if target.status == amo.STATUS_DISABLED:
+                    target.force_enable(skip_activity_log=True)
                 activity_log_action = amo.LOG.FORCE_ENABLE
-            elif appealed.filter(action=DECISION_ACTIONS.AMO_REJECT_VERSION_ADDON):
-                # otherwise it was a reject action that was appealed
+            elif self.previous_decisions.filter(
+                action=DECISION_ACTIONS.AMO_REJECT_VERSION_ADDON
+            ).exists():
                 target_versions = list(
                     target_versions.filter(
                         # we only need to unreject disabled versions we rejected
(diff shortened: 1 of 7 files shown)


src/olympia/abuse/actions.py
```diff
@@ -753,6 +758,19 @@ def process_action(self, release_hold=False):
                     )
                 target.update_status()
                 activity_log_action = amo.LOG.UNREJECT_VERSION
+            elif self.previous_decisions.filter(
+                action=DECISION_ACTIONS.AMO_REJECT_VERSION_WARNING_ADDON
```
eviljeff, line 762: A developer can't appeal a delayed rejection - if it's not cancelled by a reviewer it's followed by a rejection, that they can appeal.  
(And an override can't come from Cinder of delayed rejection, because it's only emitted by reviewer tools)

... right?  Am I missing some case?

eviljeff: Or is it this to fix a 2nd level approval action?  Can you detail the scenario?

diox: You're right - I just wanted that to work in case 2nd level approval was ever allowed to "approve content instead" again for add-ons (when I started working on this, I was going to let 2nd level approvers continue to use that in some cases).

eviljeff: codepaths that can't be reached make me itch, but I understand the logic

src/olympia/abuse/actions.py
```diff
@@ -792,6 +810,18 @@ def process_action(self, release_hold=False):
...
+    def previous_decisions(self):
+        """Queryset with previous decisions made that this action would revert."""
+        # We want a queryset to be able to filter later, so create one from the
+        # overriden instance, or just return an empty one.
+        qs = self.decision.__class__.objects.all()
+        return (
+            qs.filter(pk=self.decision.override_of.pk)
+            if self.decision.override_of
+            else qs.none()
+        )
```
eviljeff: Does this work?
```suggestion
        return self.decision.__class__.objects.filter(pk=self.decision.override_of_id)    
```

wagnerand:
@diox asked me to take a look at the testing scenarios. They look good to me overall, I have a few questions/remarks:

> Second level approval of non add-ons
> ...
> In second level approval, find that decision and note that you can either approve the content instead of proceeding with the decision

Shouldn't there also be an option to "Cancel and enqueue in Reviewer Tools"?

There is a separate scenario, which is a combination of the above. It might be out of scope for this PR, but should be verified at some point. I am not sure how a job held in second-level approval is shown in cinder. Can an admin in cinder submit an override for a job while it's held on AMO? Would a cinder admin even see that a case is being held? What would happen if a decision is overridden on cinder while it is being held on AMO?

eviljeff:
> > Second level approval of non add-ons
> 
> Shouldn't there also be an option to "Cancel and enqueue in Reviewer Tools"?

There are no reviewer tools for non add-ons. 

> There is a separate scenario, which is a combination of the above. It might be out of scope for this PR, but should be verified at some point. I am not sure how a job held in second-level approval is shown in cinder. Can an admin in cinder submit an override for a job while it's held on AMO? 

Yes

> Would a cinder admin even see that a case is being held?

No

> What would happen if a decision is overridden on cinder while it is being held on AMO?

The overridden decision would take precedence. if that decision doesn't need to be held (e.g. Approve) then it would go ahead immediately; otherwise the new decision would be held.

wagnerand:
Thanks Andrew!

> The overridden decision would take precedence. if that decision doesn't need to be held (e.g. Approve) then it would go ahead immediately; otherwise the new decision would be held.

What happens to the original held decision? Is it being canceled/closed automatically on AMO? If not, I think it probably should be?

eviljeff:
> Thanks Andrew!
> 
> > The overridden decision would take precedence. if that decision doesn't need to be held (e.g. Approve) then it would go ahead immediately; otherwise the new decision would be held.
> 
> What happens to the original held decision? Is it being canceled/closed automatically on AMO? If not, I think it probably should be?

There's no need to do anything with it - once a decision is overridden it's resolved

src/olympia/abuse/actions.py
diox: It does - 922a8549881a155a95f0808d40a8ca54a8b93e01 (maybe slightly less efficient in the case there is no previous decision - the query will be sent to MySQL which will determine it's an impossible query instead of avoiding the call to MySQL altogether, but it's simpler)

PR #23625, opened 2025-06-27
Use updated github action for slack workflow notifications

KevinMind:
Relates to: mozilla/addons#15628

### Description

Use updated github action API for dynamic context/link sections in slack notifications.

### Context

The new github action allows us to send JSON objects for both `context` and `links` these will get mapped into the the separated single link message that is sent to slack. It means the action can be flexible in terms of what information we send to it and we can add/remove items without needing to make changes to the action itself.

### Testing

CI verified slack messages can be sent.

image

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -339,44 +339,26 @@ jobs:
             name: Release Build
             dry_run: ${{ needs.context.outputs.is_release_tag != 'true' }}
 
-    env:
-      event: ${{ github.event_name }}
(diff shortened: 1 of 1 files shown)


locale/nb_NO/LC_MESSAGES/django.po
eviljeff, line 1: Random extra file in your commit

KevinMind: Strange artifact from rebasing I guess.

PR #23626, opened 2025-06-27
remove NOT_PROMOTED group as a concept.

KevinMind:
Fixes: mozilla/addons#15616

### Description

Remove the `NOT_PROMOTED` group as we rely now on non-existence to represent the NOT_PROMOTED group.

### Context

The codebase was already prepped to handle this flow during the migration to multiple promoted groups, but we did not remove the NOT_PROMOTED group as this is a clean up task. Now that multi-promoted-groups is in prod and stable we can remove this.

### Testing

CI should be enough, but you could try to execute the code paths which are touched with:
- a promoted addon
- a not promoted addon

1. Specifically check that the discovery addon promoted addon inline table still works. Should be able to view and update promoted groups for addons.
2. Run `record_reviewer_queues_counts` should work no errors.

No errors and beavhior should be as expected.

### Checklist

### DIFF
--- a/src/olympia/constants/promoted.py
+++ b/src/olympia/constants/promoted.py
@@ -7,7 +7,6 @@ from olympia.constants import applications
 
 
 PROMOTED_GROUP_CHOICES = APIChoices(
-    ('NOT_PROMOTED', 0, 'Not Promoted'),
     ('RECOMMENDED', 1, 'Recommended'),
     ('LINE', 4, 'By Firefox'),
     ('SPOTLIGHT', 5, 'Spotlight'),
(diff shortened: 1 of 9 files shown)


src/olympia/constants/promoted.py
eviljeff, line 1: Is this entire file now irrelevant? (the source of truth is a db instance, right?)

src/olympia/promoted/models.py
```diff
@@ -117,13 +117,7 @@ class PromotedGroup(models.Model):
...
 
     def __bool__(self):
-        """
-        When we use a PromotedGroup in a boolean context, we should treat NOT_PROMOTED
-        as falsey. This is how the rest of the code base expects it. Eventually,
-        we should consider deprectaing the NOT_PROMOTED group which could yield
-        the same result via a database query simply not returning a promoted group.
-        """
-        return bool(self.group_id != PROMOTED_GROUP_CHOICES.NOT_PROMOTED)
+        return bool(self.group_id)
```
eviljeff: We should just delete this method - is there any use-case for a `PromotedGroup` instance not having a truthy value for `group_id`?

KevinMind: You're right, since it depends on internal group_id reference.. it must be truthy now.

I just confirmed we do not have any promottions where the group_id is 0 so we are safe to remove.

src/olympia/constants/promoted.py
KevinMind: Not quite yet. Currently we still use it for referencing promoted group_id values. As well as the choices API. Most of that could be moved to the model and then yes we could delete this file.

I'd put that outside of this scope though.

PR #23633, opened 2025-06-30
Handle decisions coming from webhook we already created

diox:
### Context

Cinder is sending all overrides without specifying that they are an API decision or not, so we can't filter out those we created from reviewer decisions, and so process_decision() is called from the webhook handler. Without this change it would fail with a duplicate entry error for such overrides: we have already created the decision when calling Cinder API.

To avoid this, when processing decisions, handle the case where a decision with the same cinder_id already exists.

This was just causing noise, no data/logic issues: we already had the same data and were raising an integrity error before causing duplicate notifications/actions.

Fixes https://github.com/mozilla/addons/issues/15657

### DIFF
--- a/src/olympia/abuse/models.py
+++ b/src/olympia/abuse/models.py
@@ -285,7 +285,8 @@ class CinderJob(ModelBase):
         decision_notes,
         policy_ids,
     ):
-        """This is called for cinder originated decisions only."""
+        """Process a decision as sent by the webhook. If a decision with that
+        `decision_cinder_id` already exists, do nothing."""
         # appeals on REJECT_VERSION_ADDON need target_versions redefining.
         if appealed_ids := self.appealed_decisions.filter(
             action=DECISION_ACTIONS.AMO_REJECT_VERSION_ADDON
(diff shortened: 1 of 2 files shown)


src/olympia/abuse/models.py
```diff
@@ -304,33 +305,43 @@ def process_decision(
         abuse_report_or_decision = (
             self.appealed_decisions.first() or self.abusereport_set.first()
         )
-        decision = ContentDecision.objects.create(
```
KevinMind, line 307: Curious, why not just add a return if the decision exists?

```python
if ContentDecision.objects.filter(cinder_id=decision_cinder_id).exists:
  return
```

Maybe there is a slight race condition there where you could still get the error, so you could just add a try catch for integritry error? 

```python
try:
    decision = ContentDecision.objects.create(
        ...
    )
except IntegrityError.Duplicate:
    return
```

I know its basically equivalent it just seems like using the get or create adds an uglier call.. I almost changed my mind while writing this but curious to hear your thoughts on it?

diox: I was trying to avoid the race condition, yeah. I find `get_or_create` a bit neater - we need to know if something was created or not anyway, so there isn't much to gain in doing the `try`/`except` ourselves.

KevinMind:
> In reviewer tools, force disable the add-on while resolving the report

I did not see any reports.. do I need an actual connection to cinder for this?

KevinMind:
Nit: Would be great if there was a runbook for replaying a cinder webhook 🙃

diox:
Sorry, on top of filing the report for violating policies inside the add-on, you'll need the usual:
- `dsa-job-technical-processing` waffle switch on
- `dsa-abuse-reports-review` waffle switch on
- A valid `CINDER_API_TOKEN` to talk to the default `CINDER_SERVER_URL` (`https://stage.cinder.nonprod.webservices.mozgcp.net/api/v1/`)

In the absence of a true runbook, replaying the cinder webhook is essentially:
- Go to Cinder settings from the menu by clicking on the bottom left corner (probably need to be an admin)
- Go to Webhooks
- Find the decision you want to grab the payload of and click "Show payload / response"
- Copy the payload
- Paste it in a file somewhere on your local machine
- Comment out the `@permission_classes((CinderInboundPermission,))` bit of `cinder_webhook` (easier than dealing with properly signing the request)
- Run something like (outside the container, or adjust if you want to run it inside) in a `python` shell:
```python
import json
import requests
data = json.load(open('./tmp/payload.json'))
response = requests.post('http://olympia.test/api/v5/abuse/response/', json=data)
print(response.status_code)
print(response.content)
```

KevinMind:
No bueno.

image

KevinMind:
Couldn't I short circuit this somehow by calling the process_webhook method and just pass in the data from an existing content decision?

KevinMind:
For posterity, we should have two management commands:

```python
./manage.py cinder_view_webhook --output payload.json {cinder_decision_id}
```

This should recreate the cinder webhook json we recieved including the "necessary parts" optionally saving to a file. Maybbe decision id is the wrong thing, maybe its job id.

```python
./manage.py cinder_post_webhook --input payload.json
```

This should replay a webhook from a payload input by calling the endpoint with the specified json.

diox:
I don't think I can alter your user to give you more rights - not sure who can, might need to ask SRE in charge of Cinder.

> Couldn't I short circuit this somehow by calling the process_webhook method and just pass in the data from an existing content decision?

You can - that's more or less what the unit tests I've added do (for different methods). Alternatively I can grab the payload for you, since I do have access - need to give me the `cinder_id` of the corresponding `ContentDecision`, or maybe the add-on `guid`, or something so I can find it.

PR #23638, opened 2025-06-30
Add duplicate_addon_version_for_rollback task

eviljeff:
Fixes: mozilla/addons#15658 and fixes mozilla/addons#15659 (more or less)

### Description

Adds a task that can be triggered to duplicate an existing version.  Note this is the first step so can only be triggered in a django console.

(We probably wouldn't allow a developer to "rollback" to the current version, but as a proof of concept, and showing what it would look like in reviewer tools)
Screenshot 2025-06-30 150718

### Context

I started writing this from scratch but realized the existing resigning task was almost what we need, so refactored that code, so we can reuse it (mostly). This first step differs a little than the spec in that: it talks about linking the review history in the UI (this links using `VersionLog`, so it shows up in both places); and doesn't trigger the validator and ignore any errors, it just copies the old results. If we want to follow the spec in these cases, we can address in follow-ups (I think these changes are preferable anyway)

### Testing

- Identify a version of an add-on to rollback.  Ideally (to test everything) it would:
    - have a source zip upload
    - be signed 
    - have been superseded by a newer approved version 
    - have been human reviewed (either manual approval, or confirmed after auto approval)
- in a django shell import `olympia.versions.tasks.duplicate_addon_version_for_rollback` and run `duplicate_addon_version_for_rollback.delay(version_pk=<the version to rollback>, new_version_number=<new version number>, user_pk=<an author*>)`
    -  i.e. `from olympia.versions.tasks import duplicate_addon_version_for_rollback`
    - ~note the~ new version number ~is expected to have been validated before the task is called (so the error can be fed back) so~ will accept anything.
    -  *an author of the add-on, or any user with sufficient perms to upload new version
- See the task has succeeded because the version you specified has been recreated with the new version you specified
- The reviewer tools page for the add-on will look similar to the screenshot - i.e. both versions will have the activity log, and the new version will have the history for old version.
  - if you specified a source file, that should be duplicated in the new version
  - if you chose a human reviewed version, the new version will be human reviewed
- in fake mail there should be an email to all authors of the add-on telling them of the success 

### Checklist

### DIFF
--- a/src/olympia/constants/activity.py
+++ b/src/olympia/constants/activity.py
@@ -1202,6 +1202,22 @@ class CHANGE_PENDING_REJECTION(_LOG):
     # Not hidden to developers.
 
 
+class VERSION_ROLLBACK(_LOG):
+    # takes add-on, version, VersionString
+    id = 204
+    format = _('{addon} {version} created, to rollback from {0}.')
+    short = _('Version rollback')
+    review_queue = True
+
+
+class VERSION_ROLLBACK_FAILED(_LOG):
+    # takes add-on, version
+    id = 205
+    format = _('{addon} rollback from {version} failed.')
+    short = _('Version rollback failure')
+    review_queue = True
+
+
 LOGS = [x for x in vars().values() if isclass(x) and issubclass(x, _LOG) and x != _LOG]
 # Make sure there's no duplicate IDs.
 assert len(LOGS) == len({log.id for log in LOGS})
--- a/src/olympia/lib/crypto/tasks.py
+++ b/src/olympia/lib/crypto/tasks.py
@@ -68,7 +68,7 @@ def update_version_in_json_manifest(content, new_version_number):
     return json.dumps(json_data, indent=2)
 
 
-def copy_bumping_version_number(src, dst, new_version_number):
+def copy_xpi_with_new_version_number(src, dst, new_version_number):
     """Copy a xpi while bumping its version number in the manifest."""
     # We can't modify the contents of a zipfile in-place, so instead of copying
     # the old zip file and modifying it, we open the old one, copy the contents
@@ -117,9 +117,7 @@ def bump_and_resign_addons(addon_ids):
             bump_addon_version(old_version)
 
 
-def bump_addon_version(old_version):
-    mail_subject, mail_message = MAIL_COSE_SUBJECT, MAIL_COSE_MESSAGE
-    task_user = get_task_user()
+def duplicate_addon_version(old_version, new_version_number, user):
     addon = old_version.addon
     old_file_obj = old_version.file
     carryover_groups = [
@@ -136,9 +134,6 @@ def bump_addon_version(old_version):
         )
         return
 
-    log.info(f'Bumping addon {old_version.addon}, version {old_version}')
-    bumped_version_number = get_new_version_number(old_version.version)
-
     if not old_file_obj.file or not os.path.isfile(old_file_obj.file.path):
         log.info(f'File {old_file_obj.pk} does not exist, skip')
         return
@@ -149,30 +144,24 @@ def bump_addon_version(old_version):
 
     try:
         # Copy the original file to a new FileUpload.
-        task_user = get_task_user()
-        # last login ip should already be set in the database even on the
-        # task user, but in case it's not, like in tests/local envs, set it
-        # on the instance, forcing it to be localhost, that should be
-        # enough for our needs.
-        task_user.last_login_ip = '127.0.0.1'
         original_author = addon.authors.first()
         upload = FileUpload.objects.create(
             addon=addon,
-            version=bumped_version_number,
-            user=task_user,
-            channel=amo.CHANNEL_LISTED,
+            version=new_version_number,
+            user=user,
+            channel=old_version.channel,
             source=amo.UPLOAD_SOURCE_GENERATED,
-            ip_address=task_user.last_login_ip,
+            ip_address=user.last_login_ip,
             validation=old_validation,
         )
-        upload.name = f'{upload.uuid.hex}_{bumped_version_number}.zip'
+        upload.name = f'{upload.uuid.hex}_{new_version_number}.zip'
         upload.path = upload.generate_path('.zip')
         upload.valid = True
         upload.save()
 
         # Create the xpi with the bumped version number.
-        copy_bumping_version_number(
-            old_file_obj.file.path, upload.file_path, bumped_version_number
+        copy_xpi_with_new_version_number(
+            old_file_obj.file.path, upload.file_path, new_version_number
         )
 
         # Parse the add-on. We use the original author of the add-on, not
@@ -189,7 +178,7 @@ def bump_addon_version(old_version):
                 upload,
                 old_version.addon,
                 compatibility=old_version.compatible_apps,
-                channel=amo.CHANNEL_LISTED,
+                channel=old_version.channel,
                 parsed_data=parsed_data,
             )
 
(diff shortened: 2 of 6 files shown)


KevinMind:
> (We probably wouldn't allow a developer to "rollback" to the current version, but as a proof of concept, and showing what it would look like in reviewer tools)

I totally think we should allow rollback to the current version. It is so profoundly useful for almost every type of test that involves a version. I don't have a super strong opinion for devhub exposure but if we lock it can we do it on the view/api layer and still allow via admin or console.

src/olympia/constants/activity.py
```diff
@@ -1202,6 +1202,22 @@ class CHANGE_PENDING_REJECTION(_LOG):
     # Not hidden to developers.
 
 
+class VERSION_ROLLBACK(_LOG):
```
KevinMind, line 1205: I hope one day we can stop adding more activity logs 🙈

KevinMind:
Just a random thought that occurred while reviewing this PR:

```markdown
Identify a version of an add-on to rollback. Ideally (to test everything) it would:

    - have a source zip upload
    - be signed
    - have been superseded by a newer approved version
    - have been human reviewed (either manual approval, or confirmed after auto approval)
```

Could you use the django admin to search for this? That would be really handy. I'm thinking generally having test criteria defined that can be executed purely via web interface is the ideal state, because it means the STR for the ticket and the PR can be effectively the same thing. If that filtering criteria existed you could literally paste a URL path that could be entered in any environment giving the matching versions.

src/olympia/lib/crypto/tasks.py
```diff
@@ -149,30 +144,24 @@ def bump_addon_version(old_version):
...
-        # enough for our needs.
-        task_user.last_login_ip = '127.0.0.1'
         original_author = addon.authors.first()
         upload = FileUpload.objects.create(
             addon=addon,
-            version=bumped_version_number,
-            user=task_user,
-            channel=amo.CHANNEL_LISTED,
+            version=new_version_number,
+            user=user,
```
KevinMind, line 151: Is it a requirement that the user is an actual addon author? Why not leave it as task_user?

 I could almost see some value in distinguishing between "machine" uploaded files and "user" uploaded files.. even if the user is "clicking the button".
 
 If we create the FileUpload entity with the actual user, is there any way (currently) to distinguish between files created via this duplication or normally uploaded files?
 
 I don't have a specific use case in mind for that information but it is at least worth considering? If something were to be wrong with these created files after creation, being able to query for them could be useful to fix DB state. Or if we want to clean up machine created files for some other reason?

eviljeff:
> Just a random thought that occurred while reviewing this PR:
> 
> ```
> Identify a version of an add-on to rollback. Ideally (to test everything) it would:
> 
>     - have a source zip upload
>     - be signed
>     - have been superseded by a newer approved version
>     - have been human reviewed (either manual approval, or confirmed after auto approval)
> ```
> 
> Could you use the django admin to search for this?

No, you can't search for that with django admin.  Similarly to your question on Slack yesterday (about an entirely different ticket), it's usually quicker to make the changes yourself than scour the existing data for the exact match.

src/olympia/lib/crypto/tasks.py
```diff
@@ -211,9 +200,27 @@ def bump_addon_version(old_version):
         log.exception(f'Failed re-signing file {old_file_obj.pk}', exc_info=True)
         # Next loop iteration will clear the task queue.
         return
+    return new_version
+
+
+def bump_addon_version(old_version):
```
KevinMind, line 206: I'm a bit confused about the restructing/reorganization here. It seems like you've split out the "core" logic of the creating the new version to `duplicate_addon_version`, and also created the new `duplicate_addon_version_for_rollback` which calls the core method using the addon author as a user and with extra activity logs etc. But also we are keeping the `bump_addon_version` that effectively is the same logic as it was.

1) Why even keep the bump_addon_version at all? That was really just a one time thing for the sha1 deprecation project? We don't actually need that? Can we just remove that?
2) generally speaking are`bump_addon_version` and `duplicate_addon_version` actually tasks? I don't see the @task decorator. Also, in the new setup, the `duplicate_addon_version` seems to be a) a utility function that is already executed by tasks so no need to be a task, and b) more a part of "version" than "crypto".

KevinMind (commented):
Code looks generally fine but I have some questions. I don't think we should keep the bump version task unless we really think we will need it.. I don't think we will.

src/olympia/constants/activity.py
eviljeff: there's always new activities though!

src/olympia/lib/crypto/tasks.py
eviljeff: > Is it a requirement that the user is an actual addon author?

No, but it _is_ one of the authors that executed the rollback, and I thought there was value in knowing which user that was.  

> If we create the FileUpload entity with the actual user, is there any way (currently) to distinguish between files created via this duplication or normally uploaded files?

yes, next-but-one line `source=amo.UPLOAD_SOURCE_GENERATED`

KevinMind:
The testing steps are definitely lacking.
1. It's not clear how to "validate" the version number, I used the `new_version_number` method.. I think that should work but would be nice for you to specify how to validate.
2. the import syntax you've added is not really valid
3. It's not clear what I'm expecting after running the task..
4. we should probably run the task with `.apply` instead of `.delay` so you see in the shell what is happening.

Can you update with expectations of the test?

Here is the code I executed to verify

```python
from olympia.versions.tasks import duplicate_addon_version_for_rollback
from olympia.lib.crypto.tasks import get_new_version_number

addon = Addon.objects.get(guid="<my_addon_guid>")
version = addon.current_version
user = addon.authors.first()
new_version_number = get_new_version_number(version)

duplicate_addon_version_for_rollback(
    version_pk=version.id,
    new_version_number=new_version_number,
    user_pk=user.id
)
```

KevinMind:
image

Nit: I think the addon and version strings should be quoted, it's hard to tell exactly where they start/stop.

src/olympia/lib/crypto/tasks.py
eviljeff: >  That was really just a one time thing for the sha1 deprecation project?

It's been used for resigning a number of different times, ([example ](https://github.com/mozilla/addons/issues/100)from 9 years ago), but the function is older still than that.  I'm not aware of any current project that requires it (though past us have thought that many times before!) but the refactoring was to be DRY as much as possible, while keeping the scope of this PR limited to the project at hand, and not divert into cleaning up (potentially) unused code.

> generally speaking are `bump_addon_version and duplicate_addon_version` actually tasks? 

They're utility functions that the task above uses, presumably to keep the amount of lines in a single function small-ish, and to enable the sub-functions to be tested independently.

eviljeff:
>     1. It's not clear how to "validate" the version number, I used the `new_version_number` method.. I think that should work but would be nice for you to specify how to validate.

The task will accept any version number.  There is no validation requirement.

>     2. the import syntax you've added is not really valid

it's not import syntax?  I'm saying that `olympia.versions.tasks.duplicate_addon_version_for_rollback` needs to be imported. 

>     3. It's not clear what I'm expecting after running the task..

Okay, I'll add that.

>     4. we should probably run the task with `.apply` instead of `.delay` so you see in the shell what is happening.

Does that lead to any different output?  It's not supposed to print anything because it's a celery task

eviljeff:
> image
> 
> Nit: I think the addon and version strings should be quoted, it's hard to tell exactly where they start/stop.

you're probably right - I didn't stray too far from what the resign email copy was. I'll add quotes but also I think we need someone to create, or at least sign-off, what we're sending (someone probably being @abyrne-moz - let us know if you want a dedicated issue if it's a more complex answer than 👍 on quotation marks and my attempt at copy)

abyrne-moz:
> Rolling back addon 85: grant-fronkey-dolve

1. Can we change it to `add-on` so it is consistent with the rest of the message?
2. What is 85? Is that part of the name or something else that will have meaning to the developer?
3. Why does it say "An add-on reviewer wrote"? Is that part of the template or can we easily remove that for this message? I don't think an add-on reviewer actually writes this.

KevinMind:
> Does that lead to any different output? It's not supposed to print anything because it's a celery task

```python
my_task.delay(...) # sends to celery for async execution
my_task() # executes as a function directly in terminal
my_task.apply() # executes as a task synchronously logging to the shell.
```

Essentially by not using .delay you will actually see the logs from the task right now.

eviljeff:
> > Rolling back addon 85: grant-fronkey-dolve
> 
>     1. Can we change it to `add-on` so it is consistent with the rest of the message?

Sure.

>     2. What is 85? Is that part of the name or something else that will have meaning to the developer?

It's the add-on's internal AMO numeric id - I think we might expose it some places to developers

>     3. Why does it say "An add-on reviewer wrote"? Is that part of the template or can we easily remove that for this message? I don't think an add-on reviewer actually writes this.

It's part of the template (we're basically triggering the code that would be used for reviewer replies).  It's not impossible to change (worst case we'd manually send the email instead), it was just a shortcut to rely on existing functionality.

For this PR I'll do 1) and Kevin's quotation mark suggestion; and create a dedicated issue to finalise email changes.

abyrne-moz:
I'm happy with 1) and Kevin's quotation. No need to raise a ticket for the rest.

eviljeff:
> I'm happy with 1) and Kevin's quotation. No need to raise a ticket for the rest.

I filed it straight away - https://github.com/mozilla/addons/issues/15690 - we can close it again if there really is nothing to do though

PR #23651, opened 2025-07-02
Make enable/disable auto-approval an action in review page

diox:
Fixes https://github.com/mozilla/addons/issues/15646

### Testing

In the (listed) review page for an add-on, users with `Reviews:Admin` should now see a new action to Enable or Disable auto-approval depending on the state of that reviewer flag. Hitting the action should flip that flag, and log an activity recording that action took place, shown in "Important Changes" section.

#### Example of a listed review page before the changes

The actions don't have enable/disable auto-approval and we have a button in the bottom right corner that does it.

Screenshot 2025-07-01 at 14-14-58 seafowala-chanter-girroach – Add-ons for Firefox

#### Example of a review page after the changes

(due date field is missing and remaining button is different color because I changed channel in the meantime)

##### When auto-approval is still enabled in channel
Screenshot 2025-07-04 at 12-04-15 seafowala-chanter-girroach – Add-ons for Firefox

##### When auto-approval is disabled in channel
Screenshot 2025-07-04 at 15-26-07 seafowala-chanter-girroach – Add-ons for Firefox

##### Add-on important changes history section showing the new activities
Screenshot 2025-07-04 at 12-04-24 seafowala-chanter-girroach – Add-ons for Firefox

### DIFF
--- a/src/olympia/constants/activity.py
+++ b/src/olympia/constants/activity.py
@@ -1218,6 +1218,26 @@ class VERSION_ROLLBACK_FAILED(_LOG):
     review_queue = True
 
 
+class ENABLE_AUTO_APPROVAL(_LOG):
+    id = 206
+    format = '{addon} auto-approval enabled.'
+    short = 'Auto-Approval enabled'
+    keep = True
+    reviewer_review_action = True
+    review_queue = True
+    hide_developer = True
+
+
+class DISABLE_AUTO_APPROVAL(_LOG):
+    id = 207
+    format = '{addon} auto-approval disabled.'
+    short = 'Auto-Approval disabled'
+    keep = True
+    reviewer_review_action = True
+    review_queue = True
+    hide_developer = True
+
+
 LOGS = [x for x in vars().values() if isclass(x) and issubclass(x, _LOG) and x != _LOG]
 # Make sure there's no duplicate IDs.
 assert len(LOGS) == len({log.id for log in LOGS})
--- a/src/olympia/reviewers/management/commands/auto_approve.py
+++ b/src/olympia/reviewers/management/commands/auto_approve.py
@@ -179,6 +179,7 @@ class Command(BaseCommand):
             addon=version.addon,
             version=version,
             human_review=False,
+            channel=version.channel,
         )
         approve_action = helper.actions.get('public')
         if not approve_action:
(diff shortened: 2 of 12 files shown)


KevinMind:
@diox iIt would help with verification and later QA if there were screenshots for:
- enabling a disabled auto approval
- disabling an enabled auto approval

And a "before" screenshot so we can compare the extra buttons. A few light questions. I suppose the reasoning for introducing the channel as a property on the review helper is because we need to know this context to determine if auto approval is enabled/disabled right? Worth adding in context as it's not clear until you read the whole PR.

Additionally you've removed a whole section of list item markup from some template could you include a before after screenshot of that or highlight in the test scenario.

Thanks 🙏

KevinMind:
> The new channel parameter in ReviewHelper.__init__ is mandatory without a default, which may break callers that don’t pass it; consider adding a default (e.g., channel=amo.CHANNEL_LISTED).

This seems potentially relevant though i have not yet read the reviews in detail. Maybe worth making sure there is a test for scenarios where it is not defined.. for other classes that don't override.

diox:
> I suppose the reasoning for introducing the channel as a property on the review helper is because we need to know this context to determine if auto approval is enabled/disabled right? Worth adding in context as it's not clear until you read the whole PR.

Yes. For actions that care about the channel but only do something on a specific version (that's most of them) we can determine the channel from that version. But here we can't, because of the (admittedly niche, but real) use-case where you want to disable auto-approval on a channel that doesn't have any versions yet.

I'll add a comment.

> The new channel parameter in ReviewHelper.init is mandatory without a default, which may break callers that don’t pass it; consider adding a default (e.g., channel=amo.CHANNEL_LISTED).

I preferred having callers pass it explicitly to avoid surprises. The only good default would be `None` really, which I'm passing in `auto_reject.py` as it's a special scenario where we're executing an action on several versions at once which could be in different channels (though it's quite difficult to reach the state where that would happen).

diox:
> Additionally you've removed a whole section of list item markup from some template could you include a before after screenshot of that or highlight in the test scenario.

This is just removing the old buttons that enable/disable auto-approval now that it's an action.

diox:
Added a couple extra screenshots.

PR #23689, opened 2025-07-14
Add Policies FAQ to links and restructure links to make them more obvious

KevinMind:
Fixes: mozilla/addons#15681

### Description

Add a link to the FAQ

### Checklist


wagnerand:
Since this is an important page for us, would you mind adding a screenshot of what it looks like with your patch applied?

wagnerand:
~Also, do we want to use this opportunity to change the links to go to EW directly, rather than taking a roundtrip through MDN?~

It looks like we did that, but only kind of. We updated the base domain, but the paths are also different. For instance, https://extensionworkshop.com/AMO/Policy/FAQ/ does not exist.

KevinMind:
@wagnerand added a video demonstrating the changes. LMK what you think. I know it is not precisely the ask from the ticket, but I believe it is better.. HTML check boxes are supposed to be clickable via their label.

KevinMind:
@eviljeff FYI just fixing some CI complaints. PR is generally ready to review/verify but after CI is 🟢 I'll flip it back to ready officially.

wagnerand:
@fjosephmoz can you take a look at the video above and let us know if you are ok with the design and copy?

src/olympia/devhub/views.py
```diff
@@ -1952,26 +1951,20 @@ def request_review(request, addon_id, addon):
...
+        ),
     }
 
-    if doc_name in mdn_docs:
-        return redirect(MDN_BASE + mdn_docs[doc_name], permanent=True)
+    if doc_name in developer_docs:
+        return redirect(
+            settings.EXTENSION_WORKSHOP_URL + developer_docs[doc_name],
+            permanent=True,
+        )
```
eviljeff: I'm not against dropping these ancient MDN redirects per-sea ... but this are hidden changes not even mentioned in the description, let alone discussed and agreed to in the issue.

My preference would be dropping this and dealing with it in a separate issue (i.e. keep the MDN redirects for everything other than the 3 new ones we're redirecting to EW), but at the very least it should be documented in this PR and have Andreas or Alan confirm they're okay with using dropping these old redirects.

src/olympia/devhub/templates/devhub/includes/agreement.html
```diff
@@ -1,11 +1,27 @@
 <form method="post">
   {% if not agreement_form.has_error('__all__') %}
   <p>
-    {{ agreement_message }}
+    {{ _('Before starting, please read and accept our <a href="{agreement_url}" target="_blank" rel="noopener noreferrer">Firefox Add-on Distribution Agreement</a> as well as our <a href="{policies_url}" target="_blank" rel="noopener noreferrer">Review Policies and Rules</a>, and our <a href="{faq_url}" target="_blank" rel="noopener noreferrer">FAQ</a>. The Firefox Add-on Distribution Agreement also links to our Privacy Notice which explains how we handle your information.')|format_html(
+      agreement_url=url('devhub.docs', 'policies/agreement'),
+      policies_url=url('devhub.docs', 'policies'),
+      faq_url=url('devhub.docs', 'policies/faq'),
+    ) }}
```
eviljeff: It's generally not a good idea to have the `<a ...` html markup in the middle of a translated string, especially one with multiple links and extra parameters.  

The best practise (imo) is to have `open_tag = '<a href="{url("devhub.docs", "policies/agreement")}" target="_blank" rel="noopener noreferrer">` (or whatever) and `close_tag = '</a>'` and just have those placeholder tokens in the string. 
Although in most places (e.g. `verify_email.html`) we do `blah <a {{ a_attrs }}>teh link</a>` instead, and that's also pretty clean.

And because it's a multiple sentence chunk of text I'd use `{% trans %}... ` rather than `{{ _('... ` - that way you can define the placeholders in the trans tag too, and don't have to use `|format_html` (see the 404.html template as a example)

eviljeff (commented):
code review r+wc 
- I'll check it out for a test once Frantz has okayed the copy, etc.

fjosephmoz:
Just looked at the video and I'd like to go back to the original layout.  
1. FAQ not used as an acronym. Using "Frequently Asked Questions" it is easier to localize and understand. 
2. Having the "frequently asked questions" link be next to the poilcy language checkbox. 

Just a note -  I want us to incentivize people to click on and read those links through design. this treatment seems like it would be less likely for a new developer to read the policies  or the distro agreement from the UX perspective. Since the link to the info and the action to click are separated and not gated. (i.e., I don't have to click the links in the paragraph to check the boxes).

src/olympia/devhub/views.py
KevinMind: I would lean towards dropping them and seeing if we get any 404s but I checked which of them are still working

```
requesting https://developer.mozilla.org/en-US/Add-ons
requesting https://developer.mozilla.org/en-US/Add-ons
requesting https://developer.mozilla.org/en-US/Add-ons
requesting https://developer.mozilla.org/en-US/Add-ons
requesting https://developer.mozilla.org/en-US/Add-ons
requesting https://developer.mozilla.org/en-US/Add-ons#Extensions
requesting https://developer.mozilla.org/en-US/Add-ons#Other_types_of_add-ons
requesting https://developer.mozilla.org/en-US/Add-ons#Application-specific
requesting https://developer.mozilla.org/en-US/Add-ons#Themes
requesting https://developer.mozilla.org/en-US/Add-ons/Themes/Background
/Themes/Background failed
requesting https://developer.mozilla.org/en-US/Add-ons/Themes/Background/FAQ
/Themes/Background/FAQ failed
requesting https://developer.mozilla.org/en-US/Add-ons/AMO/Policy
requesting https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Reviews
requesting https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Contact
/AMO/Policy/Contact failed
requesting https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Agreement
```

Results are:
- 3 are already 404ing on MDN
- 3 are redirecting to extensionworkshop
- 1 links to an actual subsection of the addons page on MDN
- all the rest (8) just show the same page on MDN.

I'll revert the change but we are keeping stuff around for no reason and we shouldn't do that.

src/olympia/devhub/templates/devhub/includes/agreement.html
KevinMind: FIxed, used the a_attr since we have to repeat several links with the same attributes.

KevinMind:
> Just looked at the video and I'd like to go back to the original layout.
> 
>     1. FAQ not used as an acronym. Using "Frequently Asked Questions" it is easier to localize and understand.
> 
>     2. Having the "frequently asked questions" link be next to the poilcy language checkbox.
> 
> 
> Just a note - I want us to incentivize people to click on and read those links through design. this treatment seems like it would be less likely for a new developer to read the policies or the distro agreement from the UX perspective. Since the link to the info and the action to click are separated and not gated. (i.e., I don't have to click the links in the paragraph to check the boxes).

@fjosephmoz That's a good point. making the link integrated to the checkbox effectively requires the developer to open the link.

One of my concerns with that approach was the fact we will (in the new version) now have 2 links on the same line. In the ticket we have the red text but in production we would have the same blue. There will be no way for a developer to understand that it is 2 links. It breaks the assumption that we "guarantee" they've opened the links and is also super confusing.

Could a hybrid approach be to hoist the FAQ link to the description paragraph (with full text like you suggested) and then we can keep the original links as they are one per line? Wdyt?

KevinMind:
Here are 3 options we could go with?

Edit: @fjosephmoz do you have any input? If you're dead set on the inline link I can make it happen, but I wanted to take the opportunity to suggest something that might be a bit more accessibility friendly than 2 side by side links.

### Option 1 (in the description)

image

### Option 2 (in the confirmation message before submit)

image

### Option 3 (in the more info link section)

image


Thoughts?

fjosephmoz:
Why would it be hard to understand that there are 2 links if we do <Policy link> <[FAQ link]> ? I think the addition of the brackets would denote that it is a different link. (btw the red text was just to make it stand out in the design)

@KevinMind I think the updated design with the FAQ link being the only link in the paragraph is fine. (go forward with the design you shared in the last post) I want to make sure it is easily visible and accessible.

fjosephmoz (approved):

KevinMind:
I think even adding the brackets is pretty subtle and I can imagine someone easily clicking on one or the other link and thinking that they have seen the single document that is relevant. 

If it is in fact, very important that the primary links get clicked then making sure that it's totally unambiguous how to access those links is probably equally important.

I think the easiest way to make it totally unambiguous is to ensure that there is one and only one link for each checkbox. 

Given we had multiple other places where we could easily put links, seemed like a natural fit. 

Thanks for the feedback. It's pretty easy to iterate on this, so if we get any signal that this link is not being clicked on or that it's confusing in some way, we can always add another patch.

fjosephmoz:
@KevinMind sounds good. Where can I see the click data for this screen? Are we just relying on referrer on the extensionworkshop analytics? (It is fine if you don't know off the top of your head I can go look it up).

KevinMind:
> @KevinMind sounds good. Where can I see the click data for this screen? Are we just relying on referrer on the extensionworkshop analytics? (It is fine if you don't know off the top of your head I can go look it up).

I will at a `utm_referrer` param to the links and we can as you said use analytics on extension workshop to track.

src/olympia/devhub/views.py
```diff
@@ -1952,26 +1951,38 @@ def request_review(request, addon_id, addon):
 
 
 def docs(request, doc_name=None):
+    def get_url(base, doc=None):
+        base_url = urljoin(base, doc)
```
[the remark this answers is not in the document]
KevinMind: Silly robot, tricks are for kids....

```python
In [22]: for url in [get_url(MDN_BASE, ''), get_url(MDN_BASE, False), get_url(MDN_BASE, None)]:
    ...:     print(url, 'is valid' if all([urlparse(url).scheme, urlparse(url).netloc]) else 'is malformed')
    ...: 
https://developer.mozilla.org/en-US/Add-ons?utm_referrer=amo is valid
https://developer.mozilla.org/en-US/Add-ons?utm_referrer=amo is valid
https://developer.mozilla.org/en-US/Add-ons?utm_referrer=amo is valid
```

eviljeff (approved):
I *think* this is working as expected, from reading the back and forth - the testing instructions and video relate to an earlier iteration.

KevinMind:
The video I updated last night and the solution we ended up going with is having the FAQ link in the form description and keeping all of the other links the same. 

I'll double check the test instructions to make sure they match.

PR #23697, opened 2025-07-15
Move email address for regional restriction notifications

wagnerand:
Fixes: mozilla/addons#15727

### Description

This changes the email address for regional restriction change notifications from amo-admins to amo-notifications.

### Context

Moving notifications to where they belong.

### Testing

1. Add, change or delete a regional restriction for an add-on.
2. Observe the recipient address of the outgoing email.

### Checklist

### DIFF
--- a/src/olympia/addons/admin.py
+++ b/src/olympia/addons/admin.py
@@ -616,7 +616,7 @@ class AddonRegionalRestrictionsAdmin(AMOModelAdmin):
         send_mail(
             f'Regional Restriction {action} for Add-on',
             message,
-            recipient_list=('amo-admins@mozilla.com',),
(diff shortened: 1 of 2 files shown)


src/olympia/addons/admin.py
```diff
@@ -616,7 +616,7 @@ def _send_mail(self, obj, action):
         send_mail(
             f'Regional Restriction {action} for Add-on',
             message,
-            recipient_list=('amo-admins@mozilla.com',),
+            recipient_list=('amo-notifications+regionrestrict@mozilla.com',),
```
eviljeff, line 619: we should probably define these email addresses in constants rather than inline, but it's not new so 🤷

wagnerand: If it's important, let's file a followup ticket, if you don't mind.

PR #23704, opened 2025-07-16
add fi and nb-NO to supported locales

eviljeff:
part of: mozilla/addons#15691


willdurand (approved):
r+ on green ci status

PR #23708, opened 2025-07-17
cinder returns 201 for override decision

eviljeff:
Fixes: mozilla/addons#15734

### Description

Relaxes the status code enforcement we have for a decision request to Cinder, to only raise for >=400 error codes.

### Context

See issue.  Cinder originally implemented the override decision endpoint to return a 200, rather than a 201 for the other decision endpoints.  Now it's returning a 201.

### Testing

It's such a trivial change it doesn't really need testing, but if you want to:
- set your setting CINDER_SERVER_URL to `https://mozilla-staging.cinderapp.com/api/v1/`
- in reviewer tools force disable a recommended add-on
- in 2nd level approval queue in reviewer tools, requeue the add-on in the reviewer tools
- back in reviewer tools, force disable it again, but this time resolve the (newly created, from requeue) job at the same time
- in 2nd level approval queue in reviewer tools, proceed with the disable
- check in fake mail for the email to the developer about the disable
- check in Cinder that both decisions were recorded - and the second is override of the first

### Checklist

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -169,9 +169,7 @@ class CinderEntity:
         else:
             raise HTTPError(response.content)
 
-    def _send_create_decision(
-        self, url, data, action, reasoning, policy_uuids, *, success_code=201
-    ):
+    def _send_create_decision(self, url, data, action, reasoning, policy_uuids):
         data = {
             **data,
             'reasoning': self.get_str(reasoning),
@@ -186,7 +184,7 @@ class CinderEntity:
             ),
         }
         response = requests.post(url, json=data, headers=self.get_cinder_http_headers())
-        if response.status_code == success_code:
+        if response.status_code == 201:
             return response.json().get('uuid')
         else:
             raise HTTPError(response.content)
@@ -212,7 +210,7 @@ class CinderEntity:
         # https://lindie.app/share/6a21d831b39351d7c6fe898f6d22619af62dde98/PLAT-1834
         # implements the same parameters for overrides
         return self._send_create_decision(
-            url, {}, None, reasoning, policy_uuids, success_code=200
+            url, {}, None, reasoning, policy_uuids
         )
 
     def close_job(self, *, job_id):
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -184,10 +184,8 @@ class CinderEntity:
             ),
         }
         response = requests.post(url, json=data, headers=self.get_cinder_http_headers())
-        if response.status_code == 201:
-            return response.json().get('uuid')
-        else:
-            raise HTTPError(response.content)
+        response.raise_for_status()
+        return response.json().get('uuid')
 
     def create_decision(self, *, action, reasoning, policy_uuids):
         if self.type is None:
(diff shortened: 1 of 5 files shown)


src/olympia/abuse/cinder.py
```diff
@@ -186,7 +184,7 @@ def _send_create_decision(
             ),
         }
         response = requests.post(url, json=data, headers=self.get_cinder_http_headers())
-        if response.status_code == success_code:
+        if response.status_code == 201:
```
KevinMind: Since we already had to change this value due to cinder changing their response code, we should definitely use `response.raise_for_status()` instead of checking for a specific exact code. We only care that it is a successful code and that we get the json. Who care if the number ends in 0 1 or 4.

eviljeff: note that it's a pretty broad set of statuses that are treated as a "success" with [`raise_for_status`](https://github.com/psf/requests/blob/v2.32.4/src/requests/models.py#L999) - effectively any response code from 0 to 399 - i.e. not just 200, 201, or 204.  But given we won't necessarily know in advance if a subtle change like this happens, I guess acceptable...

src/olympia/abuse/tests/test_models.py
```diff
@@ -2986,7 +2986,7 @@ def _test_report_to_cinder(
             responses.POST,
             f'{settings.CINDER_SERVER_URL}decisions/{overridden_id}/override/',
             json={'uuid': uuid.uuid4().hex},
-            status=200,
```
KevinMind: Can we have at least one test returning 200 so that we know our logic doesn't care about the specific code.?

KevinMind:
I believe this failure is due to the fact that this PR comes from a fork where the `${{ vars.SLACK_ADDONS_PRODUCTION_CHANNEL }}` variable is likely not defined on github.

I've already made a patch to fix that but could you define that variable to any non empty string to see if it fixes the issue (proving what I think is wrong is what is wrong)

```
gh variable set SLACK_ADDONS_PRODUCTION_CHANNEL --body "foo"
```

should do it. Then rerun the CI.

eviljeff:
> I believe this failure is due to the fact that this PR comes from a fork where the `${{ vars.SLACK_ADDONS_PRODUCTION_CHANNEL }}` variable is likely not defined on github.
> 
> I've already made a patch to fix that but could you define that variable to any non empty string to see if it fixes the issue (proving what I think is wrong is what is wrong)
> 
> ```
> gh variable set SLACK_ADDONS_PRODUCTION_CHANNEL --body "foo"
> ```
> 
> should do it. Then rerun the CI.

I ran `gh variable set SLACK_ADDONS_PRODUCTION_CHANNEL --body "foo"` locally and reran all the CI and it doesn't seem to have made a difference (the gh docs indicate it's set at a repo level anyway?)

KevinMind:
They are indeed set on the repo, but in this case there are 2 repos, yours and the origin. It looks like what is happening is that on the `pull_request` event, github forbids the usage of secrets (we already knew about) but also variables (didn't know).

My [patch](https://github.com/mozilla/addons-server/pull/23716) fixes the issue by making the workflow able to run if the variables are not defined, like in a fork PR.

KevinMind:
FYI I wouldn't let the bug or my patch delay this patch. As long as all of the other CI is passing, the commit will build just fine on the push since that runs on the origin repo.

eviljeff:
> Make sure to make at least one of the tests return 200 so we know the error handling is enforced to be lazy.

https://github.com/mozilla/addons-server/pull/23708/commits/399f32fa26d58e0f503d57edb68d52f4d0923c74 did that

PR #23709, opened 2025-07-18
update CINDER_SERVER_URL to saas cinder

eviljeff:
Fixes: mozilla/addons#15735

### Description

As issue, just updates the setting default

### Testing

Do something that communicates with Cinder, e.g. reject a version of an add-on.  You'll need an api key defined in settings

### Checklist

### DIFF
--- a/src/olympia/lib/settings_base.py
+++ b/src/olympia/lib/settings_base.py
@@ -1473,8 +1473,7 @@ DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
 SITEMAP_DEBUG_AVAILABLE = False
 
 CINDER_SERVER_URL = env(
-    'CINDER_SERVER_URL',
-    default='https://stage.cinder.nonprod.webservices.mozgcp.net/api/v1/',
+    'CINDER_SERVER_URL', default='https://mozilla-staging.cinderapp.com/api/v1/'
 )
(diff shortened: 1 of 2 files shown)


src/olympia/lib/settings_base.py
```diff
@@ -1473,8 +1473,7 @@ def read_only_mode(env):
 SITEMAP_DEBUG_AVAILABLE = False
 
 CINDER_SERVER_URL = env(
-    'CINDER_SERVER_URL',
-    default='https://stage.cinder.nonprod.webservices.mozgcp.net/api/v1/',
+    'CINDER_SERVER_URL', default='https://mozilla-staging.cinderapp.com/api/v1/'
```
KevinMind: It's not clear to me if this is the right approach, and if it is should it default to staging or production.

I would imagine we actually want `settings_base` to not have this value defined by default or default to None. and let the dev/stage/prod configs set it as well as the `settings.py` we use for local dev to set it to stage.

This could lead to traffic in production going to cinder staging if we somehow remove the override in the production config.. that could be very bad and difficult to detect immediately.

whereas if the value is simply undefined.. all kinds of stuff will just break immediately (in this case the preferable path)

wdyt?

eviljeff: This is just updating the url like-for-like, but sure 🤷

PR #23725, opened 2025-07-23
Log name|summary changes in more detail

eviljeff:
Fixes: mozilla/addons#15742

### Description

Name and summary addon metadata changes create `EDIT_ADDON_PROPERTY` activity logs, with the added and removed localized strings in a json blob in the arguments (not displayed).  

A simple name change from `en-US`: `good add-on` to `en-US`: `bad add-on` would result in the following activity log record:

action =  `amo.LOG.EDIT_ADDON_PROPERTY.id`
arguments = `[<Addon: id>, 'name',  '{"added": ["bad add-on"], "removed": ["good add-on"]}']`

or in the database:
action = `208`
arguments =  `'[{"addons.addon": 1}, {"str": "name"}, {"str": "{\\"added\\": [\\"bad add-on\\"], \\"removed\\": [\\"good add-on\\"]}"}]`

Added and removed are de-duplicated lists, and if a localized string hasn't changed it won't be included.

### Context

I originally intended to replace `EDIT_PROPERTIES` everywhere, and create seperate `EDIT_ADDON_PROPERTY` logs for each property change but ran into some difficulties with the devhub form (localized fields weren't being returned in changed_fields property) so, given the priority on the issue, limited the change to `name` and `summary` - and added the same limitation to `AddonSerializer` for consistency.

### Testing

- In devhub, edit the name (or summary) of an add-on with listed versions in en-US locale and save the form
- check:
  - the name (or summary) has changed
  - there is an activity log in the developer's activity log view for "{addon} name property edited."
  - in a django shell (or dbshell) check the `details` json blob contains the old and new names (or summaries) for the `EDIT_ADDON_PROPERTY` activity
- repeat via the API, by submitting a PATCH to [the addon endpoint](https://mozilla.github.io/addons-server/topics/api/addons.html#patch--api-v5-addons-addon-(int-id|string-slug|string-guid)-), changing the name (or summary)
- repeat the checks:
  - the name (or summary) has changed
  - there is an activity log in the developer's activity log view for "{addon} name property edited."
  - in a django shell (or dbshell) check the `details` json blob contains the old and new names (or summaries)

- Optional further testing:
  - add and/or delete name/summary in multiple, other locales
  - duplicate some strings (name is commonly duplicated because it's often a brand name)
  - change other properties at the same time and see the older `EDIT_PROPERTIES` is used for other properties, and name/summary in add-ons without listed versions.

### Checklist

### DIFF
--- a/src/olympia/addons/serializers.py
+++ b/src/olympia/addons/serializers.py
@@ -80,6 +80,7 @@ from .models import (
 from .tasks import resize_icon, resize_preview
 from .utils import (
     fetch_translations_from_addon,
+    get_translation_differences,
     validate_version_number_is_gt_latest_signed_listed_version,
 )
 from .validators import (
@@ -1298,7 +1299,7 @@ class AddonSerializer(AMOModelSerializer):
             remove_icons(self.instance)
             self.instance.update(icon_type='')
 
-    def log(self, instance, validated_data):
+    def log(self, instance, validated_data, changes):
         validated_data = {**validated_data}  # we want to modify it, so take a copy
         user = self.context['request'].user
 
@@ -1319,6 +1320,15 @@ class AddonSerializer(AMOModelSerializer):
             # version is always a new object, and not a property either
             validated_data.pop('version')
 
+        for field in changes:
+            ActivityLog.objects.create(
+                amo.LOG.EDIT_ADDON_PROPERTY,
+                instance,
+                field,
+                details={**changes.get(field, {})},
+                user=user,
+            )
+            validated_data.pop(field)
         if validated_data:
             ActivityLog.objects.create(
                 amo.LOG.EDIT_PROPERTIES,
@@ -1356,6 +1366,7 @@ class AddonSerializer(AMOModelSerializer):
             if instance.has_listed_versions()
             else None
         )
+        changes = {}
 
         old_slug = instance.slug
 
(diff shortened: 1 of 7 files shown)


KevinMind:
FYI you can link directly to the swagger and API calls can be made easily via the browser.

http://olympia.test/api/v5/swagger/#/addons/addons_addon_partial_update

KevinMind:
This also works in dev (soon prod)

KevinMind:
I was able to verify via devhub relatively straight forward. Via API I see the description was updated, and I see the activity log in devhub but when checking the shell for some reason I still only see one activity log record.

I'm doing something like this

```python
from olympia.constants.activity import LOG

addon = Addon.objects.get(slug="camellam-pheroach-ottle")
logs = ActivityLog.objects.for_addons(addon.id).filter(action=LOG.EDIT_ADDON_PROPERTY.id)
```

After updating via devhub, I got 1 record.. then when updating via API.. still one record.

Screenshot 2025-07-25 at 23 40 03
Screenshot 2025-07-25 at 23 32 03
Screenshot 2025-07-25 at 23 28 40
Screenshot 2025-07-25 at 23 23 05

eviljeff:
It's not easy to tell where one screenshot ends and another starts - some text labels would help break the images.

But I think from what I see it's because you're trying to [update](https://mozilla.github.io/addons-server/topics/api/addons.html#edit) a field that requires a object with a string.

KevinMind (approved):
I've verified the main flow, we can verify further on stage if we want to.

eviljeff:
> I've verified the main flow, we can verify further on stage if we want to.

What is the "main flow"?  Do you mean you've verified it works, with both the API and devhub?

KevinMind:
Responded offline. I was able to verify the patch according to the steps. I only ran into issues inspecting the activity logs via the djshell. I was able to see the first log for an addon, but after making further changes to name/description I still only saw the one log.

eviljeff:
> Responded offline. I was able to verify the patch according to the steps. I only ran into issues inspecting the activity logs via the djshell. I was able to see the first log for an addon, but after making further changes to name/description I still only saw the one log.

it's `name` and `summary` - `description` isn't one of the fields tracked for content review purposes so isn't included in this (though could easily be, if Product/Operations want to expand this to other metadata)

PR #23726, opened 2025-07-24
Add bulk add functionality for disposable email domains

KevinMind:
- Introduced a new CSV file containing disposable email domains.
- Implemented the `bulk_add_disposable_email_domains` task to handle the addition of these domains.
- Added a management command to facilitate bulk addition from a CSV file.
- Created unit tests to ensure the functionality works as expected, including handling of duplicate entries.

## Description

Add a task + command to execute the task, taking a csv file (via the command) path, extracting the (domain, provider) values and passing them to the task.

The task bulk adds the records in batches to handle gracefully large payloads. Currently, it logs errors that fail django model validation, removes the m from the list of domains to add and continues.

## Context

We want an easy way to process large numbers of domain block requests.

## Testing

TBD: see the automated tests, we should be covering the following edge cases:

### Task level:
- duplicate same domain and provider
- duplicate same domain different provider
- existing model validation for `DisposableEmailRestriction`

To run the task, from a django shell run something like

```python
from olympia.users.tasks import bulk_add_disposable_domains 

entries = [
('foo.com', 'fooprovider')
]

result = bulk_add_disposable_domains.apply(args=[entries])
```

You can inspect result and check the `result.status` to see if it was ran successfully, you can verify the records exist in the database.

### Command level:

- Invalid or missing file provided.

To execute the command run from a shell

```python
./manage.py bulk_add_disposable_domains disposable_domains_2025-07-22.csv
```

- invalid file (remove the file path before checking this) should be a required argument
- missing file (delete the file but pass the path) should raise command error that the file must exist.

### Checklist

### DIFF
--- a/src/olympia/users/management/commands/bulk_add_disposable_domains.py
+++ b/src/olympia/users/management/commands/bulk_add_disposable_domains.py
@@ -28,7 +28,7 @@ class Command(BaseCommand):
             for row in reader:
                 if len(row) >= 2:
                     domain, provider = row[0], row[1]
-                    records.append((domain, provider))
+                    records.append((domain.strip(), provider.strip()))
 
         result = bulk_add_disposable_email_domains.apply(args=[records])
         logger.info(result)
--- a/src/olympia/users/management/commands/bulk_add_disposable_domains.py
+++ b/src/olympia/users/management/commands/bulk_add_disposable_domains.py
@@ -11,7 +11,7 @@ logger = logger.getLogger('z.users')
 
 
 class Command(BaseCommand):
-    help = 'Bulk add disposable email domains (no-op)'
+    help = 'Bulk add disposable email domains'
 
     def add_arguments(self, parser):
         parser.add_argument('file', type=str)
--- a/src/olympia/users/management/commands/bulk_add_disposable_domains.py
+++ b/src/olympia/users/management/commands/bulk_add_disposable_domains.py
@@ -1,0 +1,34 @@
+import csv
+from pathlib import Path
+
+from django.core.management.base import BaseCommand, CommandError
+
+from olympia.core import logger
+from olympia.users.tasks import bulk_add_disposable_email_domains
+
+
+logger = logger.getLogger('z.users')
+
+
+class Command(BaseCommand):
+    help = 'Bulk add disposable email domains (no-op)'
+
+    def add_arguments(self, parser):
+        parser.add_argument('file', type=str)
+
+    def handle(self, *args, **options):
+        file_path = Path(options['file'])
+        if not file_path.exists():
+            raise CommandError(f'File {file_path} does not exist')
+
+        records = []
+        with file_path.open(newline='') as f:
+            reader = csv.reader(f)
+            next(reader, None)  # skip header row
+            for row in reader:
+                if len(row) >= 2:
+                    domain, provider = row[0], row[1]
+                    records.append((domain, provider))
+
+        result = bulk_add_disposable_email_domains.apply(args=[records])
+        logger.info(result)
--- a/src/olympia/users/tasks.py
+++ b/src/olympia/users/tasks.py
@@ -4,6 +4,8 @@ import tempfile
 import urllib.parse
 
 from django.conf import settings
+from django.core.exceptions import ValidationError
+from django.db.utils import InterfaceError, OperationalError
 from django.urls import reverse
 from django.utils.translation import gettext
 
@@ -24,6 +26,7 @@ from olympia.amo.utils import (
 
 from .models import (
     BannedUserContent,
+    DisposableEmailDomainRestriction,
     SuppressedEmail,
     SuppressedEmailVerification,
     UserProfile,
@@ -196,3 +199,53 @@ def send_suppressed_email_confirmation(suppressed_email_verification_id):
     )
 
     verification.save()
+
+
+@task(
+    autoretry_for=(OperationalError, InterfaceError), max_retries=5, retry_backoff=True
+)
+def bulk_add_disposable_email_domains(entries: list[tuple[str, str]], batch_size=1000):
+    task_log.info(f'Adding {len(entries)} disposable email domains')
+
+    records = []
+    errors = []
+
+    for entry in entries:
+        [domain, provider] = entry
+        record = DisposableEmailDomainRestriction(
+            domain=domain,
+            reason=f'Disposable email domain of {provider}',
+        )
+
+        try:
+            record.full_clean()
+            records.append(record)
+        except ValidationError as e:
+            errors.append(e)
+
+    if not records:
+        task_log.info('No valid entries provided')
+        return
+
+    processed_domains = []
+    records_iter = iter(records)
+
+    if not isinstance(batch_size, int) or batch_size <= 0:
+        raise ValueError('batch_size must be a positive integer')
+
+    while batch := list(itertools.islice(records_iter, batch_size)):
+        created_objects = DisposableEmailDomainRestriction.objects.bulk_create(
+            batch,
+            batch_size,
+            ignore_conflicts=True,
+        )
+        processed_domains.extend(created_objects)
+        task_log.info(
+            f'Successfully processed {len(created_objects)} '
+            f'of {len(batch)} domains in this batch'
+        )
+
+    task_log.info(
+        f'Processed {len(processed_domains)} domains: '
+        f'{[obj.domain for obj in processed_domains]}'
+    )
--- a/src/olympia/users/tasks.py
+++ b/src/olympia/users/tasks.py
@@ -205,6 +205,9 @@ def send_suppressed_email_confirmation(suppressed_email_verification_id):
     autoretry_for=(OperationalError, InterfaceError), max_retries=5, retry_backoff=True
 )
 def bulk_add_disposable_email_domains(entries: list[tuple[str, str]], batch_size=1000):
+    if not isinstance(batch_size, int) or batch_size <= 0:
+        raise ValueError('batch_size must be a positive integer')
+
     task_log.info(f'Adding {len(entries)} disposable email domains')
 
     records = []
@@ -228,9 +231,9 @@ def bulk_add_disposable_email_domains(entries: list[tuple[str, str]], batch_size
         return
 
     processed_domains = []
-    records_iter = iter(records)
 
-    while batch := list(itertools.islice(records_iter, batch_size)):
+    for i in range(0, len(records), batch_size):
+        batch = records[i : i + batch_size]
         created_objects = DisposableEmailDomainRestriction.objects.bulk_create(
             batch,
             batch_size,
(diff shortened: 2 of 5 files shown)


src/olympia/users/tests/test_tasks.py
```diff
@@ -414,3 +417,62 @@ def test_retry_existing_verification(
...
+            for j in range(100)
+        ]
+        # Ensure that the task runs with 2 batches by default
+        self.batch_size = len(self.entries) // 2
+
+    def _assert_count(self, count):
+        assert DisposableEmailDomainRestriction.objects.count() == count
+
+    def _assert_result(self, result, outcome):
+        assert result.status == outcome
```
eviljeff: I don't understand why we'd ever want this level of abstraction - outcome is always 'SUCCESS' and calling `self._assert_result(result, 'SUCCESS')` actually _more_ characters than `assert result.status == 'SUCCESS`.

src/olympia/users/tests/test_commands.py
```diff
@@ -473,3 +478,128 @@ def test_migrate_twice(self):
...
+        )
+
+        with tempfile.NamedTemporaryFile(mode='w+', delete=False, suffix='.csv') as tmp:
+            tmp.write(csv_content)
+            tmp.flush()
+            tmp_path = tmp.name
+
+        assert DisposableEmailDomainRestriction.objects.count() == 0
+
+        call_command('bulk_add_disposable_domains', tmp_path)
```
eviljeff: do this within the `with` so you don't need to manually `os.remove` it afterwards

src/olympia/users/tests/test_commands.py
```diff
@@ -473,3 +478,128 @@ def test_migrate_twice(self):
...
```
eviljeff: do all this within the `with` so you don't need to manually `os.remove` it afterwards

src/olympia/users/tests/test_commands.py
```diff
@@ -473,3 +478,128 @@ def test_migrate_twice(self):
...
+    def test_file_with_header_only(self):
+        """Test that a file with only a header row does not trigger any additions."""
+        csv_content = 'Domain,Provider\n'
+
+        with tempfile.NamedTemporaryFile(mode='w+', delete=False, suffix='.csv') as tmp:
+            tmp.write(csv_content)
+            tmp.flush()
+            tmp_path = tmp.name
+
+        call_command('bulk_add_disposable_domains', tmp_path)
```
eviljeff: do all this within the `with` so you don't need to manually `os.remove` it afterwards

src/olympia/users/management/commands/bulk_add_disposable_domains.py
```diff
@@ -0,0 +1,34 @@
...
+
+from olympia.core import logger
+from olympia.users.tasks import bulk_add_disposable_email_domains
+
+
+logger = logger.getLogger('z.users')
+
+
+class Command(BaseCommand):
+    help = 'Bulk add disposable email domains (no-op)'
```
eviljeff: `(no-op)` means nothing happens and no changes are made, but this command specifically _does_ add records to the database?

src/olympia/users/tasks.py
```diff
@@ -196,3 +199,53 @@ def send_suppressed_email_confirmation(suppressed_email_verification_id):
...
+
+    if not records:
+        task_log.info('No valid entries provided')
+        return
+
+    processed_domains = []
+    records_iter = iter(records)
+
+    if not isinstance(batch_size, int) or batch_size <= 0:
+        raise ValueError('batch_size must be a positive integer')
```
eviljeff: nit: if this is a value error we should be doing it at the top to avoid wasting the the time creating the instance from the records.  But it seems redundant anyway - this is only going to be called within our own codebase.

src/olympia/users/tasks.py
```diff
@@ -196,3 +199,53 @@ def send_suppressed_email_confirmation(suppressed_email_verification_id):
...
+
+    while batch := list(itertools.islice(records_iter, batch_size)):
+        created_objects = DisposableEmailDomainRestriction.objects.bulk_create(
+            batch,
+            batch_size,
+            ignore_conflicts=True,
+        )
+        processed_domains.extend(created_objects)
```
eviljeff: I've been scratching my head over this for a while - `records` is a `list`... that we've wrapped in an `iter`... that we're slicing with `islice`... that we're casting back to a `list`.  

Just slice `records` directly with something like
```
for i in range(0, len(records), batch_size):
    batch = records[i : i + batch_size]
    ...
```
?

src/olympia/users/management/commands/bulk_add_disposable_domains.py
```diff
@@ -0,0 +1,34 @@
...
+        records = []
+        with file_path.open(newline='') as f:
+            reader = csv.reader(f)
+            next(reader, None)  # skip header row
+            for row in reader:
+                if len(row) >= 2:
+                    domain, provider = row[0], row[1]
+                    records.append((domain, provider))
+
+        result = bulk_add_disposable_email_domains.apply(args=[records])
```
eviljeff, line 33: what is `result` expected to be?  `bulk_add_disposable_email_domains` doesn't return anything

src/olympia/users/tests/test_commands.py
```diff
@@ -473,3 +478,128 @@ def test_migrate_twice(self):
...
+        with tempfile.NamedTemporaryFile(mode='w+', delete=False, suffix='.csv') as tmp:
+            tmp.write(csv_content)
+            tmp.flush()
+            tmp_path = tmp.name
+
+        try:
+            call_command('bulk_add_disposable_domains', tmp_path)
+            mock_logger.info.assert_called_with(fake_result)
+        finally:
+            os.remove(tmp_path)
```
eviljeff: Tbh I'm not sure what this is actually testing with the mocks - seems like it would be more accurate to just mock the logger and call the command with correct data to check it returns that value.  Or add that logging mock to one of the other tests.

src/olympia/users/tests/test_tasks.py
eviljeff: (`_assert_count` seems also unneeded, but at least that's called a few times, and the line is quite long)

src/olympia/users/tasks.py
eviljeff: And, if we didn't need the separate `task_log.info` calls (do we _need_ them?) then we don't have to manually batch anyway, because setting `batch_size` does the batching inside `bulk_create` for us.

src/olympia/users/tests/test_tasks.py
KevinMind: I originally had logic for expecting failure.. but yeah I guess especialy with the arugment it's a bit redundant.

Personally I like the helper assert methods because they ensure I'm calling the assertion exactly the same way across all of the tests. Reduced error and makes it super clear which tests have similar assertions.

src/olympia/users/management/commands/bulk_add_disposable_domains.py
KevinMind: It does because it is a task execution so it returns a celery task result. In my experience with these kinds of commands it is best to log more than you ought to.

src/olympia/users/tasks.py
KevinMind: Referencing previous comment, I think with these commands/tasks it is better to over log at least in the beginning. Could you give an example of what you are suggesting in terms of changes?

src/olympia/users/tests/test_commands.py
KevinMind: > seems like it would be more accurate to just mock the logger and call the command with correct data to check it returns that value

AFAICT that is literally what the test does.. we mock the data and the logger and assert the log gets passed `fake_result`

eviljeff: No, it's mocking the task return too, so it's only testing that a mocked value is used on the following line.  You could set `fake_result` to be "BLAH BLAH"; or completely rewrite `bulk_add_disposable_email_domains` so something different is returned and it'd still pass.

src/olympia/users/management/commands/bulk_add_disposable_domains.py
eviljeff: what _is_ `result` expected to be then? (this ties into the `test_bulk_add_result_is_printed` test - you're not testing it actually returns a known value anywhere.)

src/olympia/users/tasks.py
eviljeff: If we didn't need all the `task_log.info` for each batch it'd just be
```python
processed_domains = DisposableEmailDomainRestriction.objects.bulk_create(
    records, batch_size, ignore_conflicts=True
)
```

src/olympia/users/tests/test_tasks.py
eviljeff: I disagree - it _adds_ complexity by redirecting the code path (and adds confusion because you have to look up what it's actually doing - and you then discover it's literally just `assert result.status == outcome` and you spend 10 minutes having to read all the code again to double check as it _can't_ **just** be doing that...)  

A wider point is this kind of enterprise style coding is completely at odds with the style in the rest of this codebase (and even the style within this patch - in test_commands.py, you also `assert DisposableEmailDomainRestriction.objects.count()` against a value 5 times but don't have an `_assert_result` in that class).

KevinMind: Do you want me to change it or can we merge? Just thinking practically I'm not sure this is the highest priority nit at the moment.

src/olympia/users/tasks.py
KevinMind: Cool. I've left the logs for reasons stated above. Resolving.

src/olympia/users/management/commands/bulk_add_disposable_domains.py
KevinMind: Result is the return of a celery task definition synchronous call. It is a dict with some attributes.

I'm logging it so that when we deploy and run this task we have more info rather than less..

I'm happy to remove the test if that's what's bothering you. These lines are not functional requirements of the command but just helper logs to show us what is happening while and after it runs.. not really sure it's worth the back and forth though tbh.

src/olympia/users/tests/test_tasks.py
eviljeff: This wasn't a nit (I'd have labelled it as a nit).  You can fix it at the same as changing `test_bulk_add_result_is_printed` to test a real scenario.

src/olympia/users/management/commands/bulk_add_disposable_domains.py
eviljeff: I think we _should_ test it - I'm picking at it because logging the result of an asynchronous call is unusual for this codebase - I couldn't find anywhere else we did it (with either `apply`, or the 99% usage of `delay) - so we should know what it does.

src/olympia/users/tests/test_tasks.py
KevinMind: Done.

src/olympia/users/management/commands/bulk_add_disposable_domains.py
KevinMind: Removed the mock, the test now returns a real value, which we expect to be an EagerResult.

PR #23747, opened 2025-07-29
refactor AMO_LANGUAGES to make it settable per env

eviljeff:
server part of mozilla/addons#15687

### Description

Refactors how we define `AMO_LANGUAGES`, the wider list of all languages, and the other language parts of our settings so we can define which languages are enabled per env, and so enable a different list on production than on dev.

### Context

This ended up being a lot more involved than i expected - e.g. we had hardcoded exceptions for a locale like `cak` that wasn't in babel, but not others in the long time that weren't in babel. Problems for follow issues include:
- Do we need to update what's in `ALL_LANGUAGES`? 
  - probably: there are locales in there that aren't in pontoon, and some teams in pontoon that aren't in the list
  - should we align what _possible_ languages, even on dev, we have with frontend (probably, also)
- `UNSUPPORTED_LANGUAGES` is a random list of locales we had in `ALL_LANGUAGES` but wouldn't pass tests because they're "weird" in some way. We can possibly patch our support for locales to support them?  Or delete some of them from ALL_LANGUAGES if they're too obscure? - afaik non have active teams in pontoon.

### Testing

- Not really sure - this touches a core setting.  
  - Make sure devhub, etc, works and you can select a different locale than `en-US` in the dropdown?
    - (note, locally, you have frontend in the way that will redirect any locales it doesn't know about, so only the prod locales will work, unless you edit the local frontend - or the nginx config to bypass it)
  - test redirects in the url work (/developers/ to /en-US/developers, etc)
- To test the prod override you could redefine `AMO_LANGUAGES` in your local_settings.py file (the settings inheritance is similar)

### Checklist


diox:
> To test the prod override you could redefine `AMO_LANGUAGES` in your `local_settings.py` file (the settings inheritance is similar)

Didn't work that easily for me. There seems to be a subtle difference in how `local_settings.py` works that seems to prevent this from working without also redefining `LANGUAGES`. I don't think that's a blocker though, we'll just need to be careful about testing stage.

src/olympia/core/languages.py
```diff
@@ -1,252 +1,245 @@
...
+    'csb',
+    'fj-FJ',
+    'son',
+    'tl',
+    'tsz',
+]
+
+# Languages we've historically supported for statistics, etc, but we can't use because
+# they have a format we don't support in urls
+UNSUPPORTED_LANGUAGES = [
```
diox, line 235: I think we should probably keep them around just to be aware of them, but prevent them from being ever accidentally enabled with a simple test proving they are not in `PROD_LANGUAGES`.

diox:
> should we align what possible languages, even on dev, we have with frontend (probably, also)

We should.

diox (approved):

eviljeff:
> > To test the prod override you could redefine `AMO_LANGUAGES` in your `local_settings.py` file (the settings inheritance is similar)
> 
> Didn't work that easily for me. There seems to be a subtle difference in how `local_settings.py` works that seems to prevent this from working without also redefining `LANGUAGES`. I don't think that's a blocker though, we'll just need to be careful about testing stage.

Ah, thanks for that - I’d introduced a bug in the lazy loading, fixed in https://github.com/mozilla/addons-server/pull/23747/commits/80d12df4c75deae15e1ea50d822053319e292ed1

src/olympia/core/languages.py
eviljeff: added in https://github.com/mozilla/addons-server/pull/23747/commits/80d12df4c75deae15e1ea50d822053319e292ed1

eviljeff:
> > should we align what possible languages, even on dev, we have with frontend (probably, also)
> 
> We should.

I've reached out to @peiying2 to see if there's a list we can standardize on.

PR #23768, opened 2025-08-06
soft-block versions on delete

eviljeff:
Fixes: mozilla/addons#15762

### Description

Adds a task that is triggered:
- on every version delete (with that version id)
- on every addon delete (with all the versions of the addon)
- via `./manage.py process_versions --task block_old_deleted_versions --with-deleted` (with all deleted versions, batched)

Adds messaging to:
- API docs for add-on and version deletes
- Devhub for add-on and version deletes

Screenshot 2025-08-07 162225

### Context

This is, as much as possible, reusing the existing utility functions for adding blocks.  `save_versions_to_blocks` was used by a bulk command that was deleted (which is why it has support for `blocksubmission` being an unsaved instance, used just to pass along metadata), and this is doing a similar thing.  (Creating a `BlocklistSubmission` instance for a single version delete would be ludicrous; and in the vasty majority of cases (where there are only 1 or a small number of versions) also redundant for add-ons.

Note: the task does not take into account signed status of the file as the issue does not specify that as a criteria. (From historical deleted versions, about 25% aren't signed before they are deleted)

### Testing

- Delete a version from an add-on in devhub
  - see a new `Block` instance for that guid has been added to `Block.objects`, and for a `BlockVersion` that version
  - no emails should be sent - the developer was warned about the blocking in devhub and they chose to do it.
- Block a version of an add-on in django admin (create a blocklistsubmisson, and then go ahead with it to create the Block/`BlockVersion`).  It should be a hard block.
  - Then delete that version (I'm pretty sure you can't do it via devhub - if not, use `version.delete()` in a django shell.
  - Check we've not replaced it with a soft block
- Select/set-up another add-on in devhub
  - ideally with a few versions
  - delete the add-on
  - see a new Block instance for that guid has been added to `Block.objects`, and for `BlockVersion`s for all the versions
  - no emails should be sent - the developer was warned about the blocking in devhub and they chose to do it.
- To test the command:
  - If you have an existing database filled with some existing deleted versions, great.
  - If not, the easiest method is probably to delete them via devhub/api and then manually delete the BlockVersion and Block instances via django shell.
  - run the command (without `--with-deleted` it obviously does nothing, because we're filtering on deleted)
  - check that all versions that are `deleted=True` have `BlockVersion` instances.

(Optional: trigger a new bloom filter - imo unnecessary as we're using exclusively existing code paths for adding the blocks)

### Checklist

### DIFF
--- a/src/olympia/blocklist/utils.py
+++ b/src/olympia/blocklist/utils.py
@@ -10,11 +10,7 @@ from olympia.versions.models import Version
 log = olympia.core.logger.getLogger('z.amo.blocklist')
 
 
-def block_activity_log_save(
-    obj,
-    change,
-    submission_obj=None,
-):
+def block_activity_log_save(obj, *, change, submission_obj):
     from .models import BlockType
 
     action = amo.LOG.BLOCKLIST_BLOCK_EDITED if change else amo.LOG.BLOCKLIST_BLOCK_ADDED
@@ -23,11 +19,9 @@ def block_activity_log_save(
     blocked_versions = sorted(
         ver.version for ver in obj.addon_versions if ver.is_blocked
     )
-    changed_version_ids = (
-        [v_id for v_id in submission_obj.changed_version_ids if v_id in addon_versions]
-        if submission_obj
-        else sorted(ver.id for ver in obj.addon_versions if ver.is_blocked)
-    )
+    changed_version_ids = [
+        v_id for v_id in submission_obj.changed_version_ids if v_id in addon_versions
+    ]
     changed_versions = sorted(addon_versions[ver_id] for ver_id in changed_version_ids)
 
     details = {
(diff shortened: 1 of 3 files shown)


src/olympia/devhub/templates/devhub/addons/listing/_delete_warning.html
```diff
@@ -3,6 +3,11 @@
   have submitted for this add-on, listed or not.
 {% endtrans %}
 
+{% trans %}
+  All versions will be Restricted, which will disable and prevent any further installation in Firefox.
```
diox, line 7: `Restricted` with a capital `R` ? Same below.

diox:
> Block a version of an add-on in django admin (create a blocklistsubmisson, and then go ahead with it to create the Block/BlockVersion). It should be a hard block.
> Then delete that version (I'm pretty sure you can't do it via devhub

Looks like you can, actually (if the add-on itself is not disabled). That doesn't matter though.

src/olympia/devhub/templates/devhub/addons/listing/_delete_warning.html
eviljeff: I capitalized it, as we use it as a named term in devhub and elsewheere.  Alan was [okay with the wording](https://mozilla.slack.com/archives/C06JB499BEC/p1754482617038099) (assuming he didn't just overlook the case of the text)

bpc-clone:
@eviljeff 

This also effects deleted older add-on versions (in AMO by dev) prior to this change and disables them and falsely flags them as 'restricted for violating Mozilla policies'.

Not actually a 'soft block' (you may call this a feature, but it's actually a bug) ...\
At least you could have first set an update warning, but too late now.

eviljeff:
> Not actually a 'soft block' 

soft and hard block is our internal nomenclature - "soft block" is labelled "restricted", because the block type restricts the version from being installed, and disables the version by default in Firefox (but the user can choose to re-enable)

> At least you could have first set an update warning, but too late now.

Yes, we overlooked that the block reason text (which was set as "version deleted") wasn't directly exposed in Firefox.

bpc-clone:
> Yes, we overlooked that the block reason text (which was set as "version deleted") wasn't directly exposed in Firefox.

Oh, if not directly when is the 'right' block reason exposed ...?

But you're not responding to the old versions being blocked which never got a block warning on deletion (by dev).\
Users better always use the latest version, but many stick to older versions because of new (host) permissions.

eviljeff:
> > Yes, we overlooked that the block reason text (which was set as "version deleted") wasn't directly exposed in Firefox.
> 
> Oh, if not directly when is the 'right' block reason exposed ...?

There's a "more info" or "details" button/link in Firefox, which opens a page on addons.mozilla.org that displays the block reason.

> But you're not responding to the old versions being blocked which never got a block warning on deletion (by dev). 

That was a decision to reduce the risk of older deleted versions to users.  

> Users better always use the latest version, but many stick to older versions because of new (host) permissions.

This wasn't about old versions - it's specifically about _deleted_ versions.

bpc-clone:
> That was a decision to reduce the risk of older deleted versions to users.

Sure 'security' is always the reason, but very user unfriendly (and apparently also bad testing) ...

PR #23775, opened 2025-08-08
Introduce new scanner: Name Add-on Regexp Checker, NARC

diox:
Fixes https://github.com/mozilla/addons/issues/15763
Prepares (does not fully fix, see below) https://github.com/mozilla/addons/issues/15765

### Description

This introduces a new scanner called NARC. Its rules are regular expressions that are applied to the add-on names from the XPI and database, in all locales, as well as the authors display names.

A difference with yara and customs is that we have to re-run the scanner whenever add-on name is changed, including re-running actions if applicable. This adds the groundwork to allow this (mostly to prove the design works), re-scanning the version when the add-on name is changed - adding to existing results, but without re-running the actions, that will be in https://github.com/mozilla/addons/issues/15765

No new actions are introduced at this point, all existing scanner actions are available.

### Testing

- Add a Scanner rule in the admin for the `narc` scanner with whatever regexp you want in the definition of the rule. Make the rule active and associate an action with it, for instance flag for human review.
- Enable `enable-narc` waffle switch
- Enable `run-action-in-auto-approve` waffle switch
- Submit a new add-on with a name in the manifest matching the regexp in your rule
- Once the add-on is successfully submitted, run `auto_approve`.
- The action from the rule you made earlier should be executed - notably, the add-on version should have been flagged for human review if you selected any action (see note below).
- You should be able to see information about the matched rule(s) in the review page, next to the version in the history

Note: scanners always execute the most severe rule if multiple matches are found. In addition, at the moment, all currently implemented scanner actions flag for human review in addition to whatever else the action does.

Other things that should be covered by unit tests but you or QA can play with:
- Have multiple rules matching, this should be reflected in the results visible in the admin
- Disabled rules shouldn't match
- Have the name come from translations in the XPI
- Submit a new version with no matching name from the manifest/XPI translations, but the name of the add-on in the database would match instead (in whatever locale)
- Submit a new add-on with no matching name, but the user display name of the uploader matches instead
- Submit a new version with no matching name, but the user display name of one of the authors (not necessarily the uploader) matches instead
- Any of the above through devhub, signing API, add-on API

### DIFF
--- a/src/olympia/amo/utils.py
+++ b/src/olympia/amo/utils.py
@@ -953,23 +953,20 @@ class SafeStorage(FileSystemStorage):
             self.delete(dn)
 
 
-def attach_trans_dict(model, objs, fields=None):
+def attach_trans_dict(model, objs, *, field_names=None):
     """Put all translations from either all or specified non-deferred
-    translated fields from objs into a translations dict on each instance.
-
-    fields defaults to None to fetch from all known translated fields on the
-    model, or a list of field objects."""
+    translated fields by names from objs into a translations dict on each
+    instance."""
     # Get the ids of all the translations we need to fetch.
     try:
         deferred_fields = objs[0].get_deferred_fields()
     except IndexError:
         return
-    if fields is None:
-        fields = [
-            field
-            for field in model._meta.translated_fields
-            if field.attname not in deferred_fields
-        ]
+    if field_names is not None:
+        fields = [model._meta.get_field(name) for name in field_names]
+    else:
+        fields = model._meta.translated_fields
+    fields = [field for field in fields if field.attname not in deferred_fields]
     ids = [
         getattr(obj, field.attname)
         for field in fields
--- a/src/olympia/scanners/tasks.py
+++ b/src/olympia/scanners/tasks.py
@@ -236,7 +236,7 @@ def _run_narc(*, upload, version):
     if addon := upload.addon if upload else version.addon:
         # If we have an add-on instance, find all translations for the name in
         # the database as well as the display names from its authors.
-        attach_trans_dict(Addon, [addon], fields=[addon._meta.get_field('name')])
+        attach_trans_dict(Addon, [addon], field_names=['name'])
         values_from_db = dict(addon.translations[addon.name_id])
         values_from_authors.update(
             addon.authors.all().values_list('display_name', flat=True)
--- a/src/olympia/amo/tests/test_utils.py
+++ b/src/olympia/amo/tests/test_utils.py
@@ -184,7 +184,7 @@ class TestAttachTransDict(TestCase):
         attach_trans_dict(
             Addon,
             [addon, addon2],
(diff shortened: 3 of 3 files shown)


src/olympia/scanners/models.py
```diff
@@ -72,7 +72,7 @@ def add_yara_result(self, rule, tags=None, meta=None):
     def extract_rule_names(self):
         """This method parses the raw results and returns the (matched) rule
         names. Not all scanners have rules that necessarily match."""
-        if self.scanner == YARA:
+        if self.scanner in (NARC, YARA):
```
eviljeff, line 75: nit - and certainly not in this PR -  but I feel the conditionals where we check the scanner throughout this class mean this would be better implemented as multiple classes inheriting from  base scanner clas - e.g. a narc or yara scanner class would know how to extract rule names, and customs would know how it extra _it's_ names, etc.  Similarly with the clean methods.

src/olympia/scanners/tasks.py
```diff
@@ -155,6 +160,129 @@ def run_customs(results, upload_pk):
...
+        )
+        # Not part of the submission process, so we can always raise.
+        raise exc
+    else:
+        statsd.incr('devhub.narc.success')
+    log.info('Ending scanner "narc" task for Version %s.', version_pk)
+
+
+def _run_narc(*, upload, version):
+    scanner_result = (
```
eviljeff, line 217: nit: `assert upload or version` at the top of the function?  (so we don't get weird failures if neither or non-None).

src/olympia/scanners/models.py
```diff
@@ -165,17 +165,27 @@ def __str__(self):
...
+    def clean_narc(self):
         if not self.definition:
+            raise ValidationError({'definition': 'Narc rules should have a definition'})
+        try:
+            re.compile(self.definition)
+        except Exception as exc:
             raise ValidationError(
-                {'definition': _('Yara rules should have a definition')}
-            )
+                {'definition': 'An error occurred when compiling the definition'}
```
eviljeff: Can we have a more specific "regex failed to compile" error message?

src/olympia/amo/utils.py
```diff
@@ -953,19 +953,23 @@ def rm_stored_dir(self, dir_path):
             self.delete(dn)
 
 
-def attach_trans_dict(model, objs):
-    """Put all translations from all non-deferred translated fields from objs
-    into a translations dict on each instance."""
+def attach_trans_dict(model, objs, fields=None):
```
eviljeff: nit: `field_names` is a more friendly interface - we could accept a list of names, so you don't have to do `addon._meta.get_field(name)` each time.

eviljeff (commented):
Recording some nits I found (none blocking approval) before checking out the patch

src/olympia/scanners/models.py
diox: Yeah, I agree it's a bit of a mess right now.

PR #23793, opened 2025-08-12
Re-run scanner actions if results changed when running narc scanner on version

diox:
Fixes https://github.com/mozilla/addons/issues/15765
(Also addresses follow-ups for https://github.com/mozilla/addons/issues/15763)

### Description

Initial implementation of https://github.com/mozilla/addons/issues/15763 already triggered a new scan when developer or add-on name changes outside of the XPI, but we didn't re-trigger the scanner actions. This does.

### Context

To handle the case where a scanner action is run multiple times (something that is specific to `NARC`, doesn't happen for other scanners), the actions now check for a `NeedsHumanReview` corresponding to a scanner action before doing anything: if there weren't any scanner action NHR, or if one or more active scanner action NHR exist, we proceed with the action, but otherwise we abort (reviewer has acknowledged the NHR and the scanner shouldn't override what the reviewer has done)

#### Initial setup
- Add a Scanner rule in the admin for the narc scanner with whatever regexp you want in the definition of the rule. Make the rule active and associate an action with it, for instance flag for human review.
- Enable enable-narc waffle switch
- Enable run-action-in-auto-approve waffle switch

#### API submission while providing a different name
- Craft an add-on XPI that would *not* match the rule created above.
- Submit that XPI as a new listed add-on with the add-on API while passing a `name` through the parameters that would match the rule above
- Your rule action should be triggered on that add-on

#### Changing the name afterwards
- Craft an add-on XPI that would *not* match the rule created above.
- Submit that XPI
- Change the add-on name with devhub or the API with a name that matches the rule above
- Your rule action should be triggered on that add-on

#### Changing the name afterwards but the add-on was reviewed
- Craft an add-on XPI that *would* match the rule created above.
- Submit that XPI
- Your rule action should be triggered on that add-on
- As a reviewer, clear the NHR in some way, like approving/confirming auto-approval.
- As the developer, change your display name to a name that matches the rule above (this doesn't re-trigger the rule yet, but will be taken into consideration when re-scanning in the next step)
- As the developer, change the add-on name with devhub or the API to a different name that matches the rule above
- Your rule action should *not* be triggered on that add-on (a reviewer approved the code already)

### DIFF
--- a/src/olympia/scanners/actions.py
+++ b/src/olympia/scanners/actions.py
@@ -15,12 +15,25 @@ def _no_action(*, version, rule):
 
 
 def _flag_for_human_review(*, version, rule):
-    """Flag the version for human review."""
+    """Flag the version for human review if it hasn't been flagged by a scanner
+    already."""
     from olympia.reviewers.models import NeedsHumanReview
 
-    NeedsHumanReview.objects.create(
-        version=version, reason=NeedsHumanReview.REASONS.SCANNER_ACTION
-    )
+    # Check if the version has already been flagged by a scanner action.
+    # If it has not, then we just need to create a flag.
+    # (Can't use get_or_create(), there is no uniqueness constraint)
+    if not version.needshumanreview_set.filter(
+        reason=NeedsHumanReview.REASONS.SCANNER_ACTION
+    ).exists():
+        version.needshumanreview_set.create(
+            reason=NeedsHumanReview.REASONS.SCANNER_ACTION
+        )
+        return True
+    # If it has been flagged already... then return True only if one of the
+    # flags is active still.
+    return version.needshumanreview_set.filter(
+        reason=NeedsHumanReview.REASONS.SCANNER_ACTION, is_active=True
+    ).exists()
 
 
 def _delay_auto_approval(*, version, rule):
@@ -28,8 +41,11 @@ def _delay_auto_approval(*, version, rule):
 
     If delay was already set for either channel, only override it if the new
     delay is further in the future."""
-    # Always flag for human review.
-    _flag_for_human_review(version=version, rule=rule)
+    # Always try to flag for human review. If that returns False it means we
+    # already flagged before so we don't want to repeat the rest.
+    if not _flag_for_human_review(version=version, rule=rule):
+        return False
+
     in_twenty_four_hours = datetime.now() + timedelta(hours=24)
     version.addon.set_auto_approval_delay_if_higher_than_existing(in_twenty_four_hours)
     # When introducing a short auto-approval delay, reset the due date to match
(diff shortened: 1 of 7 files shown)


eviljeff:
re:
> if there weren't any scanner action NHR, or if one or more active scanner action NHR exist, we proceed with the action, but otherwise we abort (reviewer has acknowledged the NHR and the scanner shouldn't override what the reviewer has done)

Do you mean "active"?  (an active NHR would mean the reviewer _hasn't_ looked at the addon/version; an inactive would mean they have)

diox:
> Do you mean "active"? (an active NHR would mean the reviewer hasn't looked at the addon/version; an inactive would mean they have)

My sentence was poorly worded, the code is less confusing :) I meant:
- If there are any NHR, apply the action as normal (including flagging)
- If there are any active NHR, apply the action without flagging again
- Otherwise (there are one or more NHR, but all inactive), skip the action entirely, as it means a reviewer already looked at it.

src/olympia/scanners/actions.py
```diff
@@ -15,21 +15,37 @@ def _no_action(*, version, rule):
...
+    ).exists():
+        version.needshumanreview_set.create(
+            reason=NeedsHumanReview.REASONS.SCANNER_ACTION
+        )
+        return True
+    # If it has been flagged already... then return True only if one of the
+    # flags is active still.
+    return version.needshumanreview_set.filter(
+        reason=NeedsHumanReview.REASONS.SCANNER_ACTION, is_active=True
+    ).exists()
```
eviljeff, line 36: Nit/optimization: This is logically correct, but we're running two database queries to check the NHRs exist - if we run a values or values_list of is_active (and group by is_active) we should be able to know in a single query if there any any active or inactive NHRs

src/olympia/scanners/tasks.py
```diff
@@ -281,12 +279,23 @@ def _run_narc(*, upload, version):
                     )
                 )
 
-    scanner_result.results = [json.loads(result) for result in sorted(results)]
+    run_action = False
+    new_results = [json.loads(result) for result in sorted(results)]
+    if version and new_results != scanner_result.results:
```
eviljeff, line 284: The results don't contain any datetimes or other unique values, right?

src/olympia/scanners/tasks.py
```diff
@@ -281,12 +279,23 @@ def _run_narc(*, upload, version):
...
+        # we're re running the scanner on a version and found more results than
+        # when it ran on the upload. We don't know whether the action has
+        # already been triggered but in case it has we need to do it again.
+        run_action = True
+    scanner_result.results = new_results
     scanner_result.save()
     if scanner_result.has_matches:
         statsd.incr('devhub.narc.has_matches')
     for scanner_rule in scanner_result.matched_rules.all():
         statsd.incr(f'devhub.narc.rule.{scanner_rule.id}.match')
```
eviljeff, line 295: Maybe we don't want to increase the statsd pings if the results haven't changed?  (it'd be inflating the match counts)

eviljeff (approved):

src/olympia/scanners/tasks.py
diox: They shouldn't, it should all be stable. At some point in the original NARC implementation I had the `trigger` in there (with value being `version` or `upload`) and removed it when I realized it would prevent this comparison from working.

src/olympia/scanners/actions.py
diox: You're right, I don't think it matters enough to refactor it at this point though so I'm going to leave as-is for now.

src/olympia/scanners/tasks.py
diox: I went back and forth on this and I'm still not sure so I kept it as-is. The problem is, if you want to compare the number of matches against the number of runs (`devhub.narc.success`) then you need to still increase the pings even if the results haven't changed.

I introduced `devhub.narc.results_differ` ping to at least monitor when results differ but maybe we'd want another to tell us when we ran the scanner on a version and results were the same...

diox: Filed https://github.com/mozilla/addons/issues/15776 as a follow-up.

PR #23801, opened 2025-08-14
Change NARC scanner to run on versions rather than uploads

diox:
Fixes https://github.com/mozilla/addons/issues/15780

### Description

This means triggering the scanner synchronously in auto_approve, but solves various edge cases as we're guaranteed to have an addon and translations attached at that point.

### Context

Scanners typically run on `FileUpload`s during validation but NARC needs more than just the XPI, it checks translations linked to the `Addon`, and unfortunately in the API `FileUploads` are not attached to an `Addon` during validation.

We already were running scanner _actions_ in `auto_approve`, this makes us run `narc` specifically with a similar logic (only running once, when an `AutoApprovalSummary` doesn't exist yet).

We still re-run the scan on name changes as before.

### DIFF
--- a/src/olympia/reviewers/management/commands/auto_approve.py
+++ b/src/olympia/reviewers/management/commands/auto_approve.py
@@ -103,11 +103,21 @@ class Command(BaseCommand):
                     str(version.version),
                 )
 
+                # We want to execute `run_action()`/`run_narc()` only once.
+                summary_exists = AutoApprovalSummary.objects.filter(
+                    version=version
+                ).exists()
+
+                if waffle.switch_is_active('enable-narc'):
+                    if not summary_exists:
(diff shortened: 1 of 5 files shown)


src/olympia/reviewers/management/commands/auto_approve.py
```diff
@@ -103,11 +104,22 @@ def process(self, version_id):
                     str(version.version),
                 )
 
+                # We want to execute `run_action()`/`run_narc()` only once.
+                summary_exists = AutoApprovalSummary.objects.filter(
+                    version=version
+                ).exists()
+
+                if waffle.switch_is_active('enable-narc'):
+                    if not summary_exists:
```
eviljeff, line 113: Nit: combine if statements? (And the next one)

PR #23806, opened 2025-08-19
Handle manifest parsing errors caused by narc in auto_approve

diox:
Follow-up for https://github.com/mozilla/addons/issues/15780

There are some weird add-ons in auto-approval queue sometime that have a broken XPI and should be ignored to avoid messing with the rest of the add-ons waiting for auto-approval.


eviljeff (commented):
r+

PR #23814, opened 2025-08-20
Add "Disable and block" as a scanner action

diox:
This is guarded by a new boolean on `UsageTier` - if the action is not meant to be available for the tier the add-on belongs to it disables auto-approval and flags the add-on instead.

Fixes: mozilla/addons#15769


src/olympia/addons/models.py
```diff
@@ -2043,6 +2043,19 @@ def rollbackable_versions_qs(self, channel):
...
 
+    def get_usage_tier(self):
+        """Return the UsageTier instance the add-on is a part of, or None.
+
+        Note that UsageTier has additional filtering on top of just usage, so
+        it's possible for some add-ons to not be part of any UsageTier even if
+        their average_daily_users value would match one."""
+        from olympia.reviewers.models import UsageTier
+
+        for tier in UsageTier.objects.all():
```
eviljeff, line 2054: It'd be great if there was a more efficient way of doing this than doing a query and then running a query with each result - it _seems_ like it should be something that could be done in a single query. 

No idea how practical that is though.

src/olympia/addons/models.py
```diff
@@ -2043,6 +2043,19 @@ def rollbackable_versions_qs(self, channel):
...
+            if tier.get_addons().filter(pk=self.pk).exists():
+                return tier
```
eviljeff, line 2056: What about if there are overlapping tiers? I.e. the add-on is in one tier that allows disabling, but it also in another one that does not.

(Maybe this isn't best practice, but iirc we don't _prevent_ it ...?)

eviljeff (approved):

src/olympia/addons/models.py
diox: It doesn't currently happen in production, but you're right, we don't prevent it at the moment, and we should. I'll file a follow-up for that - overlapping tiers are problematic in other ways, since all the automatic flagging/disabling tasks could also affect the same add-on multiple times if there is an overlap.

src/olympia/addons/models.py
diox: Yeah, I spent a couple minutes trying to find a way to do that and gave up. It feels doable, it probably is but I didn't have time.

diox:
~~Removed the fallback to flag & disable auto-approval if the tier doesn't allow disable & block per feedback from Operations: 8c0ec34f92b4dfcd1a346b04e32982e8f234550a~~ Reverted, it turns out we want the fallback actually.

src/olympia/addons/models.py
diox: https://github.com/mozilla/addons/issues/15788

src/olympia/addons/models.py
diox: https://github.com/mozilla/addons/issues/15788#issuecomment-3209881260

PR #23817, opened 2025-08-21
When grabbing a version to re-scan for name, pick the last by pk

diox:
Default order for `Version` is by `-created, -modified`, so `last()` is actually the oldest one, which is not what we want here.

Fixes https://github.com/mozilla/addons/issues/15768#issuecomment-3206048112


eviljeff (approved):
I guess you could have changed the `last()`s to `first()`s? Though this is a cleaner end solution I agree.

PR #23820, opened 2025-08-22
Call run_action() even on initial match when scanning for narc outside auto_approve

diox:
This ensures we trigger actions from NARC matches that happen outside auto-approve, even if they are the first time NARC is called for that version, for instance because it was already approved before we enabled NARC.

Follow-up for an edge case found when testing https://github.com/mozilla/addons/issues/15780


ioanarusiczki:
Wfm, I tried with 

[632486](https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/632486)
[635515](https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/633515)
[634056](https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/634056)

To add: 
author name edit -> [631975](https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/631975)
promoted addon is held in 2nd level approval for block-disable after name edit[ 631998](https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/631998)

PR #23823, opened 2025-08-25
Allow scanner rule definition to wrap if necessary in readonly view

diox:
Only matters when viewing the rule (users with `ScannersRulesView` but without `ScannersRulesEdit`)

### Screenshot

With the patch:

Screenshot 2025-08-27 at 13-50-11 longtext View scanner rule Django site admin

### DIFF
--- a/static/css/admin/scannerrule.css
+++ b/static/css/admin/scannerrule.css
@@ -8,3 +8,8 @@
     */
     visibility: hidden;
 }
+
+.field-formatted_definition .readonly pre {
+    overflow-wrap: anywhere;
+    white-space: pre-wrap;
+}


diox:
I wasn't convinced with text-wrapping but now I'm second-guessing myself. `text-wrap: wrap` on the `<pre>` might be better than scrolling...

diox:
I updated the patch to force word-wrapping anywhere and no longer have scrollbars (force-pushed overwriting the initial commit, and updated the title & screenshot accordingly).

I was initially concerned about wrapping the definition because it might be confusing to read but I think it's better than scrolling.

ioanarusiczki:
Checked on stage with Admin:ScannersRulesView permission - the  list of scanner rules is available (without scanner query rules)
in readonly

Screenshot 2025-09-04 at 16-17-26 Make a loooong rule View scanner rule Django site admin

PR #23832, opened 2025-08-26
Allow automated disable&block action to be appealed (and handle such appeals)

diox:
Fixes https://github.com/mozilla/addons/issues/15800

### Description

`ContentActionBlockAddon` action is used to automatically disable and block the add-on when over the ratings/abuse threshold for the tier, or in the scanner action "Force-disable and block". It should allow appeals, but didn't before this change.

### Testing

Like testing https://github.com/mozilla/addons/issues/15769 or https://github.com/mozilla/addons/issues/15764 but also check that the author can then appeal, and that the appeal does re-enable the add-on (without unblocking, that's still a manual process for now)

### DIFF
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -801,8 +801,13 @@ class ContentActionTargetAppealApprove(
                 .order_by('-pk')
             )
             if self.previous_decisions.filter(
-                action=DECISION_ACTIONS.AMO_DISABLE_ADDON
+                action__in=(
+                    DECISION_ACTIONS.AMO_DISABLE_ADDON,
+                    DECISION_ACTIONS.AMO_BLOCK_ADDON,
+                )
             ).exists():
+                # FIXME: we should also automatically revert the block if the
+                # previous decision was AMO_BLOCK_ADDON.
                 target_versions = list(target_versions)
                 if target.status == amo.STATUS_DISABLED:
                     target.force_enable(skip_activity_log=True)
--- a/src/olympia/abuse/tests/test_actions.py
+++ b/src/olympia/abuse/tests/test_actions.py
@@ -1783,11 +1783,15 @@ class TestContentActionRejectVersion(TestContentActionDisableAddon):
 
 class TestContentActionBlockAddon(TestContentActionDisableAddon):
     ActionClass = ContentActionBlockAddon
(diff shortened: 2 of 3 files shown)


src/olympia/abuse/tests/test_actions.py
```diff
@@ -1783,14 +1783,18 @@ def test_notify_stakeholders_with_policy_texts(self):
 
 class TestContentActionBlockAddon(TestContentActionDisableAddon):
     ActionClass = ContentActionBlockAddon
-    takedown_decision_action = DECISION_ACTIONS.AMO_DISABLE_ADDON
+    takedown_decision_action = DECISION_ACTIONS.AMO_BLOCK_ADDON
```
diox, line 1786: This makes `_test_owner_takedown_email`, which is used in several methods we inherit from in this test class, use the correct action corresponding to the class, and in turn makes a couple tests fail if `AMO_BLOCK_ADDON` wasn't added to the `APPEALABLE_BY_AUTHOR` list in `src/olympia/abuse/constants.py` below

src/olympia/abuse/actions.py
```diff
@@ -801,8 +801,13 @@ def process_action(self, release_hold=False):
                 .order_by('-pk')
             )
             if self.previous_decisions.filter(
-                action=DECISION_ACTIONS.AMO_DISABLE_ADDON
+                action__in=(
+                    DECISION_ACTIONS.AMO_DISABLE_ADDON,
+                    DECISION_ACTIONS.AMO_BLOCK_ADDON,
```
diox, line 806: Allows the appeal to actually re-enable the add-on

src/olympia/constants/abuse.py
```diff
@@ -37,6 +37,10 @@
         'AMO_DELETE_RATING',
         'AMO_DELETE_COLLECTION',
         'AMO_REJECT_VERSION_ADDON',
+        # Note: AMO_BLOCK_ADDON is appealable, but at the moment the blocking
+        # part can't be automatically reverted by a successful appeal and has
+        # to be done manually.
+        'AMO_BLOCK_ADDON',
```
diox, line 43: Allows the automated decision to be appealable

eviljeff: I thought I did that already, d'oh.

eviljeff (approved):

PR #23840, opened 2025-08-27
Display narc scanner results in review page

diox:
Fixes https://github.com/mozilla/addons/issues/15801

Example with 3 different scanner results:
Screenshot 2025-08-27 at 01-40-30 maguar-redge-alpan – Add-ons for Firefox

### DIFF
--- a/src/olympia/scanners/templates/admin/scanners/scannerresult/formatted_matched_rules_with_files.html
+++ b/src/olympia/scanners/templates/admin/scanners/scannerresult/formatted_matched_rules_with_files.html
@@ -24,6 +24,7 @@
       <a class="rule-link" href="{% url rule_change_urlname rule.pk %}" title="{{ rule.description }} ">{{ rule.name }}</a>
     </td>
     <td>
+      {% if display_data %}<details>{% endif %}
       <ul>
       {% for item in rule.files_and_data %}
         <li>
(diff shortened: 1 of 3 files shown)


eviljeff (approved):
I have UX concerns about the extra stuff we're adding the left hand column of the version history... in some (most?) cases it'll be expanding the size.  But 🤷 - if Operations are okay with it.

Idea: maybe we could collapse by default these results if the results aren't "bad"?

diox:
Only matching results will be displayed, so if the version doesn't match any rules it won't change anything. Still, we could wrap the entire `Files & data` portion in a `<details>` element if we prefer. @wagnerand thoughts ?

abhn:
Will the `string` and the `pattern` fields have ellipses when they're too wide to display? We might want to display them with a overflow. 

Collapsing the data in a details element sounds good to me (showing the name of the rule and scanner that flagged).

diox:
@abhn I'm now hiding `pattern`, but yes: all fields have ellipses, you can see that in my screenshot where I've purposefully made a bunch of the strings longer.

abhn:
> all fields have ellipses

Is it then displayed with title text? I just want the full string/regex available for reviewers to see, and not get cut off just because it is too long.

diox:
> Is it then displayed with title text? I just want the full string/regex available for reviewers to see, and not get cut off just because it is too long.

Not at the moment - the ellipsis is added automatically via CSS when it would be cut, there is no title text.

PR #23853, opened 2025-09-01
Allow product pages to be noindex'ed temporarily

willdurand:
Fixes: https://github.com/mozilla/addons/issues/15808

### Description

This PR does the following:

1. it allow product pages to be noindexed on content change
2. the API exposes a new boolean property (detail/search endpoints, but search will always return `is_noindexed=null`)
3. the sitemap excludes the noindexed product pages

### Checklist


src/olympia/shelves/tests/test_views.py
```diff
@@ -151,8 +151,8 @@ def test_only_enabled_shelves_in_view(self, tag_mock):
         # don't enable shelf_c
         self.shelf_d.update(enabled=True)
 
-        # would be 24 but we mocked Shelf.tag that does a query.
-        with self.assertNumQueries(26):
+        # would be 24 (??) but we mocked Shelf.tag that does a query.
+        with self.assertNumQueries(27):
```
willdurand, line 155: I tried to use select/prefetch related in `PrimaryHeroShelfViewSet` (`src/olympia/hero/views.py`) but that didn't change that particular number. Some other tests got less queries, though, so maybe there is some optimization to consider there. But not for that one...

src/olympia/addons/tests/test_views.py
```diff
@@ -2643,6 +2643,21 @@ def test_dont_trigger_narc_on_name_change_if_waffle_is_inactive(
         self._test_metadata_content_review()
         assert run_narc_on_version_mock.delay.call_count == 0
 
+    def test_noindex_on_content_change(self):
+        # Make sure the add-on is recent enough.
+        self.addon.update(created=self.days_ago(1))
+        assert not self.addon.is_listing_noindexed
+        self._test_metadata_content_review()
+        self.addon.refresh_from_db()
```
willdurand, line 2651: FWIW `self.addon.reload()` didn't work.

diox:
At a minimum we need something like:
```diff
diff --git a/src/olympia/addons/views.py b/src/olympia/addons/views.py
index f917c411af..9b23aa33e8 100644
--- a/src/olympia/addons/views.py
+++ b/src/olympia/addons/views.py
@@ -285,7 +285,7 @@ class AddonViewSet(
             # translations because at least the addon name is likely to be
             # needed in most cases.
             qs = qs.only_translations()
-        return qs
+        return qs.select_related('addonlistinginfo')
 
     def get_serializer_class(self):
         # Override serializer to use serializer_class_for_developers if the author or
diff --git a/src/olympia/bandwagon/views.py b/src/olympia/bandwagon/views.py
index 23e98c80f0..2a300a6715 100644
--- a/src/olympia/bandwagon/views.py
+++ b/src/olympia/bandwagon/views.py
@@ -202,6 +202,7 @@ class CollectionAddonViewSet(ModelViewSet):
             CollectionAddon.objects.filter(collection=self.get_collection())
             .prefetch_related(
                 'addon__promotedaddon',
+                'addon__addonlistinginfo',
                 'addon___current_version__file___webext_permissions',
             )
             .transform(self._transformer)
```


(Ideally the `CollectionAddonViewSet` would do a `select_related()` too but because it has to traverse down it seems to mess things up and trigger the default transformer... resulting in a lot more queries. I didn't want to spend time tracking why, so I've used `prefetch_related()` instead, but that means an extra query, so the tests need some tweaking)

src/olympia/shelves/tests/test_views.py
diox: That test is for secondary shelves, I don't think the primary hero is affected by your changes, since it's using serializers that only specific fields that don't include the one you've added.

The secondary shelves load via a collection (directly using the `CollectionAddonViewSet`!), so the `prefetch_related()` I've suggested in my other comment to make the collection view scale correctly would also help here.

That test doesn't have enough add-ons to prove it, but without my patch, this would scale badly: as it stands the test will do 27 queries but actually, without my patch that extra query would happen for each add-on, where as with my patch it only happens once thanks to the `prefetch_related()`.

willdurand: > That test is for secondary shelves

ha yeah... that explains a lot of things 🙄

docs/topics/api/overview.rst
```diff
@@ -477,6 +477,7 @@ These are `v5` specific changes - `v4` changes apply also.
 * 2024-08-22: restricted add-on eula_policy endpoint to non-themes only. https://github.com/mozilla/addons/issues/14937
 * 2024-10-17: replaced ``versions`` with ``blocked`` and ``soft_blocked`` in blocklist api; dropped unused ``min_version`` and ``max_version``. https://github.com/mozilla/addons/issues/15015
 * 2025-06-26: added ``data_collection_permissions`` and ``optional_data_collection_permissions`` to the version detail endpoint (and therefore addons/search as well). https://github.com/mozilla/addons/issues/15620
+* 2025-09-01: added ``is_noindexed`` to the add-on detail/search endpoint. Note that the value is unconditionally set to ``null`` in the search response. https://github.com/mozilla/addons/issues/15808
```
diox: That note about the search response should be elsewhere - probably in the main `addons.rst` doc

PR #23854, opened 2025-09-01
Implement searching by created/updated threshold values in the API

diox:
Fixes https://github.com/mozilla/addons/issues/15814

### Testing

See new API docs, this allows searching by add-on `created` or last `updated` date with all the threshold operators we already supported for `users` and `ratings`. The main addition is that for these new parameters, we support milliseconds timestamps, or `YYYY-MM-DD` date, or `YYYY-MM-DDTHH:MM`/`YYYY-MM-DDTHH:MM:SS` datetime strings.

So you can craft add-ons matching any arbitrary criteria you like to test this ; for instance, searching for all extensions created after 2025: `/api/v5/addons/search/?type=extension&created__gte=2025-01-01`


docs/topics/api/addons.rst
```diff
@@ -79,7 +81,7 @@ This endpoint allows you to search through public add-ons.
 
 .. _addon-threshold-param:
 
-    Threshold style parameters allow queries against numeric values using comparison.
+    Threshold style parameters allow queries against arbitrary values using comparison.
```
eviljeff: Not really _arbitrary_, but okay.

src/olympia/search/filters.py
```diff
@@ -141,12 +141,25 @@ class FooAddonThresholdQueryParam(AddonThresholdQueryParam):
         '__gte': 'gte',
         '': 'eq',
     }
+    valid_types = (int, float, str)
```
eviljeff: Do any of the threshold parameters support `str`?  It appears all the subclasses override this to only support int, float, date, or datetime

src/olympia/search/filters.py
```diff
@@ -141,12 +141,25 @@ class FooAddonThresholdQueryParam(AddonThresholdQueryParam):
...
+
+    def convert_value(self, value):
+        if value.isdecimal():
+            converter = int
+        elif value.replace('.', '', 1).isdecimal():
```
eviljeff: Technically this is a backwards incompatible change - previously any [valid float string](https://docs.python.org/3/library/functions.html#float) would be accepted - but 🤷.

eviljeff (approved):
Can you add a test to `TestAddonSearchView` to confirm the date, datetime, and timestamp filtering actually works in an ES query?  (we have `test_filter_by_ratings` which tests the existing functionality)

docs/topics/api/addons.rst
diox: Suggestions welcome, I just didn't want to keep "numeric" now that dates are allowed for some parameters

src/olympia/search/filters.py
diox: Not at moment no. I'll remove it.

src/olympia/search/filters.py
diox: Yeah originally I used to `str` fallback for dates so I had to add this condition to distinguish floats from strings. Eventually as you saw I didn't end up using `str` at all so this is no longer as useful, I'll see if it makes sense to keep it.

diox: 4488b424a40a7065530f3a44be8e730b586614f2

src/olympia/search/filters.py
diox: 4488b424a40a7065530f3a44be8e730b586614f2

docs/topics/api/addons.rst
eviljeff: Could just say "numeric or date-time"

PR #23865, opened 2025-09-02
Fix default_locale erroneously set to lowercase for spanish add-ons&collections

diox:
Fixes https://github.com/mozilla/addons/issues/15770


src/olympia/addons/migrations/0056_fix_default_locale_spanish.py
```diff
@@ -0,0 +1,28 @@
...
+# on some add-ons in previous migration instead of es-ES, this fixes that.
+
+def forwards_func(apps, schema_editor):
+    Addon = apps.get_model('addons', 'Addon')
+    Addon.unfiltered.filter(default_locale='es-es').update(default_locale='es-ES')
+
+
+def backwards_func(apps, schema_editor):
+    Addon = apps.get_model('addons', 'Addon')
+    Addon.unfiltered.filter(default_locale='es-es').update(default_locale='es-es')
```
eviljeff, line 17:
```suggestion
    Addon.unfiltered.filter(default_locale='es-ES').update(default_locale='es-es')
```
no?

src/olympia/bandwagon/migrations/0010_fix_default_locale_spanish.py
```diff
@@ -0,0 +1,27 @@
...
+# on some collection in previous migration instead of es-ES, this fixes that.
+
+def forwards_func(apps, schema_editor):
+    Collection = apps.get_model('bandwagon', 'Collection')
+    Collection.unfiltered.filter(default_locale='es-es').update(default_locale='es-ES')
+
+
+def backwards_func(apps, schema_editor):
+    Collection = apps.get_model('bandwagon', 'Collection')
+    Collection.unfiltered.filter(default_locale='es-es').update(default_locale='es-es')
```
eviljeff, line 16: also here
```suggestion
    Collection.unfiltered.filter(default_locale='es-ES').update(default_locale='es-es')
```

src/olympia/addons/migrations/0056_fix_default_locale_spanish.py
eviljeff: (Or just drop the backwards function)

diox: Doesn't matter, filtering is case-insensitive unless we specify a case-sensitive collation. The backwards function isn't really useful, it's just to be nice on local environments.

PR #23866, opened 2025-09-02
Expose add-on listing info to the django admin

willdurand:
Fixes: https://github.com/mozilla/addons/issues/15818


diox (approved):
Good enough for now

PR #23867, opened 2025-09-02
Don't let NARC fail on invalid xpi/manifest

diox:
Fixes: mozilla/addons#15812

### Context

This is a follow-up fixing issues caused by broken add-ons that we have on dev (either because we manipulated the data, or we bypassed validation for them)

In https://github.com/mozilla/addons-server/pull/23806 I worked around those add-ons at auto-approval time, but because we are now running NARC as a query rule as well, we need to move that handling deeper in NARC itself. So this change includes a revert of https://github.com/mozilla/addons-server/commit/6bb890217e9f54a311bc727d118c6ba1b6373748

### Testing

Easier to just trust the unit tests, but you can submit an add-on, then replace its file on the filesystem with an xpi that has no manifest or no name in the manifest, or a broken json manifest, and run a NARC query rule ; it should complete successfully.

### DIFF
--- a/src/olympia/reviewers/management/commands/auto_approve.py
+++ b/src/olympia/reviewers/management/commands/auto_approve.py
@@ -3,7 +3,6 @@ from collections import Counter
 from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.db import transaction
-from django.forms import ValidationError
 
 import waffle
 from django_statsd.clients import statsd
@@ -176,10 +175,6 @@ class Command(BaseCommand):
                 'Version %s was skipped because approval action was not available',
                 version,
             )
-            self.stats['error'] += 1
-        except ValidationError:
-            statsd.incr('reviewers.auto_approve.approve.failure')
-            log.info('Version %s was skipped because of a validation error', version)
             self.stats['error'] += 1
         except SigningError:
             statsd.incr('reviewers.auto_approve.approve.failure')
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/tasks.py
```diff
@@ -229,14 +230,20 @@ def _run_narc(*, scanner_result, version, rules=None):
...
+        # Something else should stop us if the manifest or the xpi itself is
+        # invalid, NARC shouldn't fail for that. This matters if validation was
+        # forced by an admin or something similar.
+        data = {}
     values_from_xpi = Addon.resolve_webext_translations(data, version.file.file).get(
         'name', {}
     )
-    if isinstance(values_from_xpi, str):
-        values_from_xpi = {None: values_from_xpi}
+    if values_from_xpi is None or isinstance(values_from_xpi, str):
```
eviljeff, line 245: Not sure how important this is, but I can't see how `values_from_xpi` would be `None` after the changes you've made - if there's a `ValidationError` data would be `{}` (not `None`) so `resolve_webext_translations` would return early with `data` (because there is no `default_locale` in the dict); and the `.get()` fallback would mean `values_from_xpi` would be `{}`

diox: The `ValidationError` would be raised for malformed JSON/XPI only - `test_run_invalid_manifest_somehow` covers this case.

We still need to handle the case where the manifest is valid json, but `name` and `default_locale` are absent or `null`. That doesn't raise (*), `data` ends up being a dict with `name` and `default_locale` in it set to `None`, we pass through `resolve_webext_translations` not changing the dict, and so `values_from_xpi` ends up being `None`. The `test_run_xpi_no_name()` covers that case.

(*) maybe it should but not willing to change that...

PR #23869, opened 2025-09-03
Fix rollback with only one channel available

eviljeff:
Fixes: mozilla/addons#15804

### Description

Fixes the scenario with an add-on with 2 channels, but only one channel available for rollback, where the channel selector was disabled.  Now the field on the form is disabled instead so the initial channel value is correctly used.

Also fixes a (separate) bug where the explanation label isn't available because there's no channel selector (and the javascript that sets it is on the channel selector)

### Context

I initially thought these two bugs were connected, but turns out they're for different scenarios (two channels, but only one channel available for rollback vs. only one channel)

### Testing

- Set-up an add-on with 2 or more approved versions in the unlisted channel (and nothing in the listed channel)
  - see the explanation text is present and changes with the version selection in the rollback form
  - rollback one of the versions
  - (refresh the versions page afterwards)
- upload and approve a listed version
  - now the channel selector should be present but disabled on the rollback form
  - rollback an unlisted version again

optional - repeat with listed
- upload and approve another listed version
- see the rollback form has a functional channel selector
- delete all but one unlisted version
- see the channel selector is disabled and defaulted to listed; the explanation text is there and correct
- rollback the latest listed version
- delete the remaining unlisted version
- see the channel selector is not present; the explanation text is there and correct

### Checklist

### DIFF
--- a/src/olympia/devhub/tests/test_forms.py
+++ b/src/olympia/devhub/tests/test_forms.py
@@ -1571,7 +1571,7 @@ class TestRollbackVersionForm(TestCase):
         )
         assert form.is_valid()
 
-    def test_clean_adds_version(self):
+    def test_clean_adds_version_for_listed(self):
         addon = addon_factory()
         lv1 = addon.current_version
         version_factory(addon=addon)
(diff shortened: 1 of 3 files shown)


src/olympia/devhub/tests/test_forms.py
```diff
@@ -1606,8 +1606,9 @@ def test_clean_adds_version(self):
             'version': uv1,
         }
 
-        # and when we select listed but there is no listed version availble: error
+        # and when we select listed but there is no listed version available: error
         data['channel'] = amo.CHANNEL_LISTED
+        del data['unlisted_version']
```
diox: Can we split this test into several separate ones instead ? There are 3 separate sets of assertions, we should probably just have 3 methods to reduce the chance of them affecting each other.

static/js/zamboni/devhub.js
```diff
@@ -652,6 +653,14 @@ function initVersions() {
       });
     })
     .trigger('change');
+
+    if ($hiddenChannelInput) {
+      if ($hiddenChannelInput.val() == '1') {
```
diox: We should avoid hardcoding `1` here

PR #23870, opened 2025-09-03
Optimize NARC scanner

diox:
Fixes https://github.com/mozilla/addons/issues/15819

This should be a transparent change, NARC scanner should produce same results as before, but much faster. To achieve that, we:
- De-duplicate strings to check - if an add-on contains the same string X times for different locales, check once (and record X results for each locale as before)
- Avoid parsing XPI, instead rely on `FileManifest` if possible (if `name` is translated in the manifest and a `default_locale` is declared, then we open the XPI)
- Cache character normalization table in memory as it's very expensive to compute

### DIFF
--- a/src/olympia/reviewers/tests/test_commands.py
+++ b/src/olympia/reviewers/tests/test_commands.py
@@ -724,9 +724,12 @@ class TestAutoApproveCommand(AutoApproveTestsMixin, TestCase):
         )
         # Fake file manifest data to avoid dealing with a real file. We want to
         # avoid resolve_webext_translations() which wants a real file...
-        FileManifest.objects.create(file=self.version.file, manifest_data={
-            'name': 'Foo',
-        })
+        FileManifest.objects.create(
+            file=self.version.file,
+            manifest_data={
+                'name': 'Foo',
+            },
+        )
 
         call_command('auto_approve')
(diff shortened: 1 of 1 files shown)


eviljeff:
Does this require manual testing?  If so, can haz testing instructions.

diox:
Up to you, there is no functionality change, so depends on how much you trust the code you see and the tests.

If you're feeling adventurous you can try:
- Enabling `enable-narc`, `run-action-in-auto-approve` waffle switches
- Creating a NARC rule that would match everything (definition set to `.*`) in the admin
- Uploading a new add-on in the unlisted channel (needs a true add-on)
- Running `auto_approve`
- From the admin page for the rule or the review page for the add-on, you should see it matched your rule
- Creating a NARC *query* rule in the admin that would match your add-on
- Running it after saving
- It should execute rather quickly and find your add-on (possibly more if you had more add-ons with `FileManifest` available that match whatever definition you set)

To see the speed difference in action, you can try running `TestRunNarc` tests, they should be noticeably quicker than on `master` (that particular suite is 5 times faster for me now). Or even just the full upload flow, because the normalization function is called at validation to check for Mozilla trademark violations (I'm seeing gains of ~1s per upload after the first one)

PR #23871, opened 2025-09-03
Add rollback release notes

eviljeff:
Fixes: mozilla/addons#15663

### Checklist


src/olympia/devhub/forms.py
```diff
@@ -1627,6 +1627,13 @@ class RollbackVersionForm(forms.Form):
         required=False,
     )
     new_version_string = forms.CharField(max_length=255)
+    release_notes = forms.CharField(
+        widget=forms.Textarea(attrs={'rows': 2}),
+        max_length=255,
+        required=False,
+        label=gettext('Release notes'),
+        initial=gettext('Automatic rollback based on version [m.m].'),
```
diox: Since it's going to be called when the file is parsed these should be using `_()` instead to keep the calls lazy.

PR #23874, opened 2025-09-04
Remove filemanifest backfill

diox:
Fixes https://github.com/mozilla/addons/issues/15821

### Context

We don't need that backfill anymore. Having `FileManifest` created in `file_factory()` when we pass a `filename` is going to be useful for future tests.


src/olympia/versions/tests/static_themes/theme_images.zip
diox, line 1: Had to be changed because it's used by `test_generate_static_theme_preview()` which calls the factory with that file, but it didn't contain a valid manifest. I added a dummy manifest in it , that was the easiest fix.

eviljeff (approved):
The change to file_factory is probably a more significant change than removing some old code but 🤷

diox:
I could have made a separate patch for it but it's really the driver behind this - I need it for the optimize NARC patch but the backfill was getting in the way.

PR #23907, opened 2025-09-11
Record correct user for version unblocking activity log

diox:
Fixes mozilla/addons#15777

### Testing
- In the admin (`/en-US/admin/models/blocklist/`), add a block for an entire add-on (all versions)
- Log out, log back in the admin with a different user (need to have/create a second user with a `mozilla.com` email that has relevant permissions or `*:*`)
- Delete the block
- In the reviewer page for the add-on (it's linked from the block admin page), check that the activity log for the unblocking created by deleting the block is associated with the second user and not the first.

### DIFF
--- a/src/olympia/blocklist/tests/test_utils.py
+++ b/src/olympia/blocklist/tests/test_utils.py
@@ -382,7 +382,9 @@ class TestDeleteVersionsFromBlocks(TestCase):
         version_factory(addon=addon)
         user = user_factory()
         block_factory(guid=addon.guid, updated_by=self.task_user)
-        changed_version_ids = list(addon.versions.values_list('pk', flat=True))
+        changed_version_ids = list(
+            addon.versions.values_list('pk', flat=True).order_by('pk')
+        )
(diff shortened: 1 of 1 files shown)


src/olympia/blocklist/utils.py
```diff
@@ -97,26 +96,25 @@ def block_activity_log_delete(obj, deleted, *, submission_obj=None, delete_user=
...
         details=details,
-        user=submission_obj.updated_by if submission_obj else delete_user,
+        user=submission_obj.updated_by,
     )
     log_create(
         amo.LOG.BLOCKLIST_VERSION_UNBLOCKED,
         *((Version, version_id) for version_id in changed_version_ids),
         obj,
-        user=obj.updated_by,
+        user=submission_obj.updated_by,
```
diox, line 114: This is the fix: `obj` here is the `Block` instance we're deleting - we want the `BlocklistSubmission` that holds the information about what/who/why we're deleting instead, and that's `submission_obj`. The rest of the code was already doing that correctly.

PR #23916, opened 2025-09-16
Restrict future approvals in "Disable and block" scanner action

diox:
Also to go with that:
- Link (exact) matching restrictions from the user admin to easily find automatically added restrictions when looking up a user in the admin, in order to make such restrictions easier to remove on successful appeals.
- Add activity log when automatically disabling auto-approval because of a restriction

Fixes https://github.com/mozilla/addons/issues/15833

### DIFF
--- a/src/olympia/users/models.py
+++ b/src/olympia/users/models.py
@@ -881,6 +881,21 @@ class IPNetworkUserRestriction(RestrictionAbstractBaseModel):
         return str(self.network)
 
     @classmethod
+    def network_from_ip(cls, ip):
+        """
+        Return the smallest meaningful network to restrict from an IP
+        """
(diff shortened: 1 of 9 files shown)


src/olympia/users/models.py
```diff
@@ -880,6 +880,21 @@ class Meta:
     def __str__(self):
         return str(self.network)
 
+    @classmethod
+    def network_from_ip(cls, ip):
```
eviljeff, line 884: now it's abstracted into a function, test please.

diox: https://github.com/mozilla/addons-server/pull/23916/commits/45320bafa8b47a216fff21502997bb8d64416d63

PR #23917, opened 2025-09-16
Add 7 days cooldown to dependabot

diox:
Add a 7 days cooldown delay to dependabot, making us wait until the dependencies are a little more stable before updating.

TODO:
- Add an exception for the linter
- Consider different delay for some packages (would a 7 day cooldown mean we'll never see a dependency PR for the google deps for instance?)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,21 +5,29 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 99
+  cooldown:
+    default-days: 7
 - package-ecosystem: "docker-compose"
   directory: "/"
   schedule:
     interval: weekly
   open-pull-requests-limit: 99
+  cooldown:
+    default-days: 7
 - package-ecosystem: "github-actions"
   directory: "/"
   schedule:
     interval: weekly
   open-pull-requests-limit: 99
+  cooldown:
+    default-days: 7
 - package-ecosystem: "npm"
   directory: "/"
   schedule:
     interval: daily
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,21 +5,31 @@ updates:
...
+    default-days: 7
 - package-ecosystem: "npm"
   directory: "/"
   schedule:
     interval: daily
   open-pull-requests-limit: 99
+  cooldown:
+    default-days: 7
+    exclude:
+      - addons-linter
```
eviljeff, line 36: if we're excluding packages we own, and addons-frontend?  (+ autograph?)

eviljeff (approved):
Needs an issue/ticket

.github/dependabot.yml
diox: Good idea. https://github.com/mozilla/addons-server/pull/23917/commits/1c1c1f5ee7db4ac84580b2f6ad5c0bd5034ae0c6

diox: `autograph` doesn't have a `mozilla/` parent but I think not immediately upgrading to the latest version of `autograph` locally isn't a bad thing anyway...

eviljeff: `mozilla-services/`, but 🤷 you could be right

diox: What matters is the package/image name. And actually I was wrong, `autograph` image on docker hub is under `mozilla/` like `addons-frontend` is, so it would get the exclusion as well.

PR #23926, opened 2025-09-17
Make various fields in django admin read-only

wagnerand-moz:
Fixes: mozilla/addons#15843

### Description

This make some fields in django admin read-only. For the specific list of fields, see the patch or the issue description.

### Context

We want to avoid making accidental or unintended (potentially breaking) changes using those fields.

### Testing

Open each django admin detail page for the models listed in the issue and make sure the fields mentioned are read-only.

### Checklist

### DIFF
--- a/src/olympia/addons/admin.py
+++ b/src/olympia/addons/admin.py
@@ -46,7 +46,7 @@ class AddonListingInfoInline(admin.TabularInline):
     view_on_site = False
 
 
-class AddonUserInline(admin.TabularInline):
+class AddonUserBase:
     model = AddonUser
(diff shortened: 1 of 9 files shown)


src/olympia/addons/admin.py
```diff
@@ -52,6 +52,14 @@ class AddonUserBase:
...
         'user',
     )
+    fields = (
+        'addon',
+        'user',
+        'role',
+        'listed',
+        'position',
+        'user_profile_link',
+    )
```
wagnerand-moz, line 62: This is to keep the order.

PR #23942, opened 2025-09-22
tidy up and add more django admin permissions

eviljeff:
Fixes: mozilla/addons#15837

### Description

Make the permission changes specified in the issue, along with adding a helper function for the common case of having a single permission for adding, changing, and deleting.

### Context

The tests for CinderPolicyAdmin were broken - the way we were creating a user, giving it permissions, and then logging in in the class setup, somehow broke something, meaning that any subsequent `force_login` - even with a user with `*:*` would fail as unauthenticated.

### Testing

Impractical to test every new permission, but some tests for the changes that caused some difficulties:
- create a group with `CinderPolicies:View` and assign your (non-admin) user to it, then try to list CinderPolicies; note you can't edit the policy to make it available in the reviewer tools; but can sync policies from Cinder
- ~create a group with Admin:Advanced, assign your user to it, and try to delete (not ban) a User that has Ratings attached.  Note you should not be a member of a group with `*:*` or `Ratings:Delete`~

### Checklist

### DIFF
--- a/src/olympia/constants/permissions.py
+++ b/src/olympia/constants/permissions.py
@@ -162,11 +162,6 @@ DJANGO_PERMISSIONS_MAPPING.update(
     {
         'abuse.change_abusereport': ABUSEREPORTS_EDIT,
         'abuse.view_cinderpolicy': CINDER_POLICIES_VIEW,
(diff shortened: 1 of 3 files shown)


eviljeff:
Added a further commit to prevent UserProfile deletes via django admin entirely, after issue discussion.

PR #23944, opened 2025-09-22
Run scanner query rules using replica, not primary db

diox:
Only use the primary to grab the rule to make sure it exists when we try to get it.

Fixes https://github.com/mozilla/addons/issues/15850

### Description

Now that we have more workers for the tasks routed to the `zadmin` queue like the scanners query rule tasks, we're causing a bunch of parallel connections to the primary db, which isn't ideal, we want the load to be distributed better.

This might not be enough to fix it, we might need to make getting the rule use a replica too, but this should be a safe&quick fix to try first.

### Testing

Can't be tested with a regular development environment. You can check that a simple scanner query rule still executes but unless you add a database replica you won't be able to test the change itself locally.

### DIFF
--- a/src/olympia/amo/tests/test_models.py
+++ b/src/olympia/amo/tests/test_models.py
@@ -449,8 +449,8 @@ class BaseQuerysetTestCase(TestCase):
             config = qs.get_with_primary_fallback(key='a')
         assert config
         assert config.value == 'Zero'
-        # Despite the query originally using 'replica', we should have used
-        # 'default' since the original get() call raised DoesNotExist.
+        # Despite the query originally using 'fake-replica', we should have
+        # used 'default' since the original get() call raised DoesNotExist.
         assert config._state.db == 'default'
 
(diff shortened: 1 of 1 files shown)


src/olympia/scanners/tasks.py
```diff
@@ -505,7 +504,7 @@ def run_scanner_query_rule_on_versions_chunk(version_pks, query_rule_pk):
         version_pks[0],
         version_pks[-1],
     )
-    rule = ScannerQueryRule.objects.get(pk=query_rule_pk)
+    rule = ScannerQueryRule.objects.using('default').get(pk=query_rule_pk)
```
willdurand: Nit: it might be good to add a comment like on `run_scanner_query_rule`.

diox:
@willdurand your comment gave me another idea to improve this, I've added https://github.com/mozilla/addons-server/pull/23944/commits/4cef52b780a9c6d54ef071c142f951d77695a807 which includes a test for the new method I'm adding for this. Can you re-review ?

PR #23947, opened 2025-09-23
Normalize emails before restricting them automatically in a scanner action

diox:
We use `get_or_create()` but that's not enough as emails are normalized by the `save()` method.

Running the following will always raise in `get_or_create()`:
```python
EmailUserRestriction.objects.create(email_pattern='foo+variant@example.com')
# At this point we recorded a restriction for foo@example.com
EmailUserRestriction.objects.get_or_create(email_pattern='foo+variant@example.com')
```

The fix is to do this instead:
```python
EmailUserRestriction.objects.create(email_pattern='foo+variant@example.com')
# At this point we recorded a restriction for foo@example.com
EmailUserRestriction.objects.get_or_create(
    email_pattern=EmailUserRestriction.normalize_email('foo+variant@example.com')
)
```

Fixes https://github.com/mozilla/addons/issues/15833 (follow-up found by QA)

### DIFF
--- a/src/olympia/scanners/actions.py
+++ b/src/olympia/scanners/actions.py
@@ -101,7 +101,7 @@ def _restrict_future_approvals(*, version, rule, restriction_type):
     }
     for user in users:
         EmailUserRestriction.objects.get_or_create(
-            email_pattern=EmailUserRestriction.normalize_email(user.email),
+            email_pattern=user.email,
             restriction_type=restriction_type,
             defaults=restriction_defaults,
         )
--- a/src/olympia/users/models.py
+++ b/src/olympia/users/models.py
@@ -994,6 +994,15 @@ class NormalizeEmailMixin:
         return normalized_email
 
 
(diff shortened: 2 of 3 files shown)


diox:
FYI the same fix was used before for creating restrictions automatically on bans: https://github.com/mozilla/addons-server/blob/d14c1c9aef8a772321d7a38dc309985f39456d8e/src/olympia/users/tasks.py#L55

eviljeff (approved):
This seems like a footgun - maybe we should address it in EmailUserRestrictionManager (i.e. define a Manager for the model)?  I guess override `get_or_create`

diox:
Did that in https://github.com/mozilla/addons-server/pull/23947/commits/4459f2d57b277e33b070b53ad8aa8b5906f9583a

ioanarusiczki:
I tried a scenario with str from @diox, plus testing https://github.com/mozilla/addons/issues/15856 ->  now rule is hit and the addon is also disabled.

PR #23948, opened 2025-09-24
Fix versions not being flagged for review after 2nd level approval requeue

diox:
Fixes mozilla/addons#15822

### Description

This refactors how versions are passed when flagging for review in `CinderAddonHandledByReviewers`, always passing down a `Version` queryset

### Context

We were expecting version strings in `flag_for_human_review()` (which is inefficient and confusing) but the requeue method was passing a Version queryset directly.

Later inside `flag_for_human_review()` there was a fallback that worked for the listed channel that caused us to flag the `current_version` (with further handling for the case where it's an appeal and there is no `current_version` anymore). But for unlisted, nothing, so we didn't flag anything.

### Testing

- Enable all DSA waffle switches
- Submit an add-on as unlisted (do not have a listed version!)
- Make that add-on promoted so that negative decisions are held for 2nd level approval
- Force-disable that add-on - the decision should be held for 2nd level approval
- In second level approval, requeue the decision

Expected result:
- You should see that add-on in the reviewer queue ; in the review page, the version should be flagged as needing human review, there should be an activity log explaining that it was requeued, and the version should have a due date

Actual result before this patch:
- None of that ^^

### DIFF
--- a/src/olympia/abuse/cinder.py
+++ b/src/olympia/abuse/cinder.py
@@ -16,7 +16,6 @@ from olympia.amo.utils import (
     create_signed_url_for_file_backup,
 )
 from olympia.users.utils import get_task_user
-from olympia.versions.models import Version
 
 
 log = olympia.core.logger.getLogger('z.abuse')
@@ -494,12 +493,22 @@ class CinderAddonHandledByReviewers(CinderAddon):
     # This queue is not monitored on cinder - reports are resolved via AMO instead
     queue_suffix = 'addon-infringement'
 
-    def __init__(self, addon, *, version_string=None):
+    def __init__(self, addon, *, versions_strings=None):
         super().__init__(addon)
-        self.version_string = version_string
+        self.versions_strings = versions_strings
+
+    def get_versions(self):
+        """Return Version queryset corresponding to the versions strings on
+        this instance."""
+        qs = self.addon.versions(manager='unfiltered_for_relations')
+        if self.versions_strings:
+            qs = qs.filter(version__in=self.versions_strings).no_transforms()
+        else:
+            qs = qs.none()
+        return qs
 
     def flag_for_human_review(
-        self, *, related_versions, appeal=False, forwarded=False, second_level=False
+        self, *, versions, appeal=False, forwarded=False, second_level=False
     ):
         """Flag an appropriate version for needs human review so it appears in reviewers
         manual revew queue.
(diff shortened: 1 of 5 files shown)


src/olympia/abuse/cinder.py
```diff
@@ -539,46 +548,33 @@ def flag_for_human_review(
...
-            )
-            if related_versions
-            else set()
-        )
-        # If we have more versions specified than versions we flagged, flag current
-        # to be safe. (Either because there was an unknown version, or a None)
-        if len(version_objs) != len(related_versions) or len(related_versions) == 0:
+        # If we don't have versions, flag current to be safe.
+        # (There were either no specific version reported, or an unknown one)
+        if not versions:
```
eviljeff: This is dropping one of the cases -`len(version_objs) != len(related_versions)`, when there was a version in the list that didn't exist, or was None - and only replaces the `len(related_versions) == 0` case.

There being a `None` in the list of versions is pretty common as it means no `addon_version` in the `AbuseReport`, i.e. the add-on wasn't installed/enabled in Firefox at the time the report was filed.

I don't think we care that much about version strings that don't exist, but it _would_ mean that a job with multiple reports where one report had a specified version (maybe an old version), and then other reports with no specified version, wouldn't lead to the current version being flagged.

eviljeff: It could be cleaner to re-add that logic in `get_versions` though (maybe just looking for a `None` in the list is sufficient and appending `current_version` 🤷).

diox: When would that scenario happen ? Is that only after a forwarding ? I've changed `post_queue_move()` to set `self.version_strings` before calling `flag_for_human_review()` and made it ignore reports with `addon_version` being `None`, I guess that's the actual issue here ? I should have kept the `None` and if there are any `None` version strings inside `get_versions()` add `current_version` ?

eviljeff: > When would that scenario happen ? Is that only after a forwarding ? 

Yeah, after a job has been forwarded from Cinder to AMO reviewers it could have multiple reports in it, and one of more of them could have no `addon_version` - so we would now flag the `current_version` in that scenario.

> and made it ignore reports with `addon_version` being `None`, I guess that's the actual issue here ? I should have kept the `None` and if there are any `None` version strings inside `get_versions()` add `current_version` ?

That'd work

diox: https://github.com/mozilla/addons-server/pull/23948/commits/ea34f2d68f60057bf5f24a8788b09d76399c962c (I didn't end up doing it inside `get_version()` - it didn't work well because we also needed the fallback in cases where we don't use `get_versions()` at all)

src/olympia/abuse/cinder.py
```diff
@@ -539,46 +548,36 @@ def flag_for_human_review(
...
-                    .filter(channel=amo.CHANNEL_LISTED)
-                    .no_transforms()
-                    .first()
+                and self.addon.find_latest_version(
+                    channel=amo.CHANNEL_LISTED, exclude=(), deleted=True
                 )
             )
             if latest_or_current:
-                version_objs.add(latest_or_current)
-        version_objs = sorted(version_objs, key=lambda v: v.id)
```
eviljeff, line 566: nit: This meant that the versions in `log.debug` and `activity.log_create` below were in a consistent order.  It'd be nice to retain that.

diox: https://github.com/mozilla/addons-server/pull/23948/commits/0beb3d413a22ef1de38391abdfaf23151ac28afb

PR #23956, opened 2025-09-26
Fix potential IntegrityError exception in EmailUserRestriction admin

diox:
### Description

Since we're normalizing patterns on save (and when checking at upload) if someone tried to record a non-normalized variant of an email that we already had in the database, it would raise an `IntegrityError` exception.

This fixes that problem by normalizing earlier, so that we raise a proper validation error to the user.

Slight drive-by: remove some unnecessary gettext calls around restrictions-related strings that only show up in the admin.

### Testing

- Through the admin, add an email user restriction for `foo+bar@example.com` ; that should record `foo@example.com` (unchanged)
- Add an email user restriction for `foo+else@example.com`. That should show a validation error to the user instead of raising an integrity error exception.

Fixes https://github.com/mozilla/addons/issues/15766

### DIFF
--- a/src/olympia/users/forms.py
+++ b/src/olympia/users/forms.py
@@ -55,7 +55,7 @@ class IPNetworkUserRestrictionForm(AMOModelForm):
 
         if ip_address and network:
             raise forms.ValidationError(
-                ('You can only enter one, either IP Address or Network.')
+                'You can only enter one, either IP Address or Network.'
             )
         elif ip_address is not None and not network:
             data['network'] = f'{ip_address}/32'


src/olympia/users/forms.py
```diff
@@ -34,7 +55,7 @@ def clean(self):
 
         if ip_address and network:
             raise forms.ValidationError(
-                _('You can only enter one, either IP Address or Network.')
+                ('You can only enter one, either IP Address or Network.')
```
eviljeff: nit: extra brackets

diox: https://github.com/mozilla/addons-server/pull/23956/commits/b6bf49cfc3d109d2b22d11ec9f641f441d525283

PR #23973, opened 2025-10-01
create missing migrations

eviljeff:
Fixes: mozilla/addons#15803

### Description

Adds all missing django migrations.

### Context

- 0062_rename_created_idx_abuse_reports_created_idx is a weird one - I'm assuming there was a bugfix in django at some point that lead to it not recognising the index renaming we did via our own migration class (that should have satisfied django with DropIndex and CreateIndex).  I did take a brief look to see if there was an obvious cause, but it was easier just to fake it in the end.
- I had to manually rewrite some of the promoted models changes - we should have created the migrations to delete the old models before renaming the new ones because django couldn't discover that (for example) `PromotedAddonPromotion` became `PromotedAddon` and thought it was a new model.

### Testing

- `./manage.py migrate`

### Checklist


src/olympia/addons/migrations/0057_alter_addon_description_and_more.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('translations', '0003_puretranslation_purifiedmarkdowntranslation'),
+        ('addons', '0056_fix_default_locale_spanish'),
+    ]
+
+    operations = [
```
diox, line 15: Looks like this is dropping and re-adding constraints because of the change to a `PurifiedMarkdownField`, but it's unnecessary. Can we fake that ?

src/olympia/promoted/migrations/0033_rename_promotedaddonversion_to_promotedapproval.py
```diff
@@ -0,0 +1,28 @@
...
+
+    operations = [
+        migrations.RenameModel(
+            old_name='PromotedAddonVersion',
+            new_name='PromotedApproval',
+        ),
+        migrations.AlterModelTable(
+            name='promotedapproval',
+            table='promoted_promotedaddonversion',
+        ),
```
diox: Can we fake that ? It's pointless and will break stuff between the 2 renames

src/olympia/promoted/migrations/0034_rename_promotedaddonpromotion_to_promotedaddon.py
```diff
@@ -0,0 +1,28 @@
...
+    dependencies = [
+        ('addons', '0057_alter_addon_description_and_more'),
+        ('promoted', '0033_rename_promotedaddonversion_to_promotedapproval'),
+    ]
+
+    operations = [
+        migrations.RenameModel(
+            old_name='PromotedAddonPromotion',
+            new_name='PromotedAddon',
+        ),
```
diox, line 22: As migration 33, this + the `AlterModelTable` below should be faked together

src/olympia/versions/migrations/0048_alter_version_release_notes.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('translations', '0003_puretranslation_purifiedmarkdowntranslation'),
+        ('versions', '0047_auto_20241031_1750'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
diox: As the add-on one, this is dropping and re-adding the same constraint, presumably because Django doesn't realize `PurifiedMarkDownField` is just a proxy model and there are no actual changes, can we fake that instead ?

src/olympia/promoted/migrations/0033_rename_promotedaddonversion_to_promotedapproval.py
eviljeff: The RenameModel should be effectively fake anyway... except for the table rename.  If we swap the order of the AlterModelTable (so the name is hard-coded before the model rename) it should be 2 no-ops at the db level.

eviljeff: https://github.com/mozilla/addons-server/pull/23973/commits/3b5860d952524f691cfc5474aa432120e375e5c4

src/olympia/addons/migrations/0057_alter_addon_description_and_more.py
eviljeff: https://github.com/mozilla/addons-server/pull/23973/commits/7494ad722acbc94543604adb483c51a717dab1d8

diox (approved):

PR #23984, opened 2025-10-03
Add additional filters to scanner query rules/results

diox:
Fixes https://github.com/mozilla/addons/issues/15831


src/olympia/amo/admin.py
```diff
@@ -642,6 +645,27 @@ def choices(self, changelist):
...
+    Custom rangefilter.filters.NumericRangeFilter class without the need for
+    inline CSS/JavaScript.
+
+    Needs FakeChoicesMixin for the fake choices the template will be using (the
+    upstream implementation depends on inline JavaScript for this, which we
+    want to avoid).
+    """
+
+    template = 'admin/amo/numeric_range_filter.html'
+
```
eviljeff, line 659: When testing the label tags were empty (the label tag existed, but there was no text value). Date range filter fields get "To" and "From" labels; numeric range filter fields did not - though the placeholder texts did say "To" and "From".

Do you need to override `_get_form_fields` too?

At that point, you probably want to inherit from DateRangeFilter, or a common ancestor - it's worth considering anyway as your `choices` seems to be a direct copy and paste.

src/olympia/files/models.py
```diff
@@ -112,7 +113,7 @@ class File(OnChangeMixin, ModelBase):
     status = models.PositiveSmallIntegerField(
         choices=STATUS_CHOICES.items(), default=amo.STATUS_AWAITING_REVIEW
     )
-    datestatuschanged = models.DateTimeField(null=True, auto_now_add=True)
+    datestatuschanged = models.DateTimeField(null=True, default=timezone.now)
```
eviljeff, line 116: Not saying this is a bad change to make ... but we could just special case `datestatuschanged` field in `file_factory` - pop the value and set it after the create, and so avoid any database/model change to prod.

eviljeff (approved):

PR #23991, opened 2025-10-06
rm MAD scanner

eviljeff:
Fixes: mozilla/addons#15849

### Description

Removes all MAD scanner integration in addons-server

### Context

There are three database columns (two already null-able; one with a migration to make it null-able) that will need removing in a subsequent push.

### Testing

Shouldn't be necessary, but you could:
- upload an xpi so scanners run (now without mad)
- access reviewer tools (mad review queue is gone)
- load a review page and perform an action that would have clear the mad flag (e.g. approve a version)

### Checklist


willdurand:
> There are three database columns (two already null-able; one with a migration to make it null-able) that will need removing in a subsequent push.

Do we already have a ticket for that?

eviljeff:
> > There are three database columns (two already null-able; one with a migration to make it null-able) that will need removing in a subsequent push.
> 
> Do we already have a ticket for that?

https://github.com/mozilla/addons/issues/15887

PR #24728, opened 2026-04-10
Allow multiple BlocklistSubmission in parallel for the same version

### LINKED ISSUE #16117: [Task]: Support having multiple block submissions in parallel
### Description

As part of implementing automated actions depending policy violations, we should make it possible to have 2 different delayed block submissions co-exist for the same guid + version ids, so that we can have in parallel:
- A BlocklistSubmission that soft-blocks delayed in X days
- A BlocklistSubmission that hard-blocks the same guid+versions, delayed in Y days

Currently, under the hood it mostly already works and https://github.com/mozilla/addons-server/pull/24570 makes use of this, but the admin tool doesn't allow you to create a submission for version ids that already have a pending one. In this ticket we should:
- Make sure we handle conflicting blocklistsubmissions in a predictable way. The main thing to worry about is ensuring we never downgrade a block through a `ADDCHANGE` action.
- Lift that restriction about multiple blocklist submissions for admins, possibly warning them about the potential conflict

### Acceptance Criteria

- [ ] item 1

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/AMOENG-2368)

diox:
Conflicts are resolved automatically when setting up the block depending on the BlocklistSubmission action: an ADDCHANGE action is never allowed to "downgrade" a block_type, only an expicit SOFTEN can do that. In addition, adding a delay on a SOFTEN action is disallowed.

Fixes https://github.com/mozilla/addons/issues/16117

### DIFF
--- a/src/olympia/blocklist/models.py
+++ b/src/olympia/blocklist/models.py
@@ -94,11 +94,8 @@ class Block(ModelBase):
             .annotate(**{GUID: models.F(GUID)})
             .select_related('blockversion')
         )
-        all_submission_versions = BlocklistSubmission.get_all_submission_versions()
-
         all_addon_versions = defaultdict(list)
         for version in qs:
-            version.blocklist_submission_id = all_submission_versions.get(version.id, 0)
             all_addon_versions[getattr(version, GUID)].append(version)
         for block in blocks:
             block.addon_versions = all_addon_versions[block.guid]
@@ -249,7 +246,6 @@ class BlocklistSubmission(ModelBase):
             'is_blocked',
             'is_hard_blocked',
             'is_soft_blocked',
-            'blocklist_submission_id',
         ),
     )
     FakeBlock = namedtuple(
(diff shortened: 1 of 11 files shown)


src/olympia/blocklist/models.py
```diff
@@ -368,12 +351,7 @@ def update_signoff_for_auto_approval(self, *, ignoring=None):
         is_pending = self.signoff_state == self.SIGNOFF_STATES.PENDING
         add_action = self.action == self.ACTIONS.ADDCHANGE
         if is_pending and (
-            (
-                self.all_adu_safe()
-                and not self.has_potential_conflicts(ignoring=ignoring)
-            )
-            or add_action
-            and not self.has_version_changes()
+            (self.all_adu_safe()) or add_action and not self.has_version_changes()
```
eviljeff: nit: now unneeded surrounding brackets.  Though, to me, the whole `if` could do with reformatting to make it clearer.

eviljeff (approved):
> In addition, adding a delay on a SOFTEN action is disallowed.

Do we need to enforce this elsewhere - e.g. in the models itself?  BlocklistSubmissions can be created automatically now/soon.

diox:
> Do we need to enforce this elsewhere - e.g. in the models itself? BlocklistSubmissions can be created automatically now/soon.

I don't think it's necessary (famous last words). We should never have automation adding a delay with a soften action in the first place, softening is really meant to fix a situation were we were too harsh, I don't see a situation where we would want to automatically do that with a delay.

PR #24731, opened 2026-04-13
Add option to scan add-on summaries with NARC

diox:
Fixes https://mozilla-hub.atlassian.net/browse/AMOENG-2359

### DIFF
--- a/src/olympia/addons/utils.py
+++ b/src/olympia/addons/utils.py
@@ -17,7 +17,7 @@ def generate_addon_guid():
     return '{%s}' % str(uuid.uuid4())
 
 
-def validate_addon_name(name, user, *, form=None):
+def validate_addon_name(name, *, user, form=None):
     """
     Validate that an add-on name is allowed.
 
@@ -51,7 +51,7 @@ def validate_addon_name(name, user, *, form=None):
                 )
                 raise forms.ValidationError(msg)
 
-    validate_name(name, check_function, gettext('This name cannot be used.'), form=form)
+    validate_name(name, check_function=check_function, form=form)
 
     return name
 
--- a/src/olympia/addons/utils.py
+++ b/src/olympia/addons/utils.py
@@ -17,6 +17,24 @@ def generate_addon_guid():
     return '{%s}' % str(uuid.uuid4())
 
 
+def validate_addon_summary(summary, user, form=None):
+    """
+    Validate that an add-on summary is allowed.
+    """
+    # For the moment we don't do anything other than checking the summary isn't
+    # too complex, so we apply this regardless of the user.
+    return validate_name(
+        summary,
+        check_function=None,
+        # Make the string generic so that it can be used in devhub but also
+        # when importing the summary from the 'description' field in the
+        # manifest.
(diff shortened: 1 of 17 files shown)


src/olympia/addons/utils.py
```diff
@@ -51,7 +69,7 @@ def check_function(normalized_name, variant):
                 )
                 raise forms.ValidationError(msg)
 
-    validate_name(name, check_function, gettext('This name cannot be used.'), form=form)
+    validate_name(name, check_function=check_function, form=form)
```
eviljeff, line 72: why drop the error_message arg/kwarg here and make it the default instead?

eviljeff: Seems like we're both making `validate_name` more generic (by making it useful for summary too), but at the same time making it more specialized (by adding defaults just for name in the function)?

src/olympia/files/utils.py
```diff
@@ -960,7 +960,8 @@ def check_xpi_info(
         # Make sure we pass in a copy of `xpi_info` since
         # `resolve_webext_translations` modifies data in-place
         translations = Addon.resolve_webext_translations(xpi_info.copy(), xpi_file)
-        validate_addon_name(translations['name'], user)
+        validate_addon_name(translations['name'], user=user)
+        validate_addon_summary(translations['summary'], user=user)
```
eviljeff, line 964: This would mean a problem in the summary would be hidden by a problem in the name - i.e. we'd fail for the name, and only after they fixed it and resubmitted would we fail for summary. Should we execute these calls in try... catch and combine the exceptions... somehow?

src/olympia/addons/utils.py
diox: My intent was to make it more generic. I noticed for collections we were re-using the same generic `This name cannot be used.` error message and that's why I added a default, but it does make it a bit more confusing. So maybe I can do back to repeating that error message and even force callers to pass an explicit `error_message`.

src/olympia/files/utils.py
diox: I agree, but this is already the case for a bunch of exceptions raised by this function. Do we really care? Only a handful of add-ons in prod raise an error for this, and they are all disabled add-ons or very old broken theme imports.

src/olympia/addons/utils.py
eviljeff: I guess whatever is the most DRY. Just seemed like a change that was out of place for the overall intent of the patch

src/olympia/files/utils.py
eviljeff: If it's not a problem in reality then 🤷

ioanarusiczki:
Mmm does this require qa ? let me know

diox:
@ioanarusiczki yes please. As the title suggests, this adds a configuration option to use NARC rules to scan summaries (and the `description` field in manifest) of add-ons. Because it's not enabled by default and we still have https://github.com/mozilla/addons/issues/16048 you would need to create a rule and then edit it to change the configuration and have the rule apply to listing summary and/or xpi `description`.

PR #24732, opened 2026-04-13
Added support for follow-up actions

### LINKED ISSUE #16118: [Task]: Implement ability to have follow-up blocks for negative enforcement actions
### Description

Implement new "follow-up" enforcement actions: “Delayed soft-block in X days”, “Delayed hard-block in Y days”, “Delayed hard-block in Z days”
Each correspond to a possible second and third action in [Add-on Policies violations enforcement actions](https://docs.google.com/document/d/1pRRDRyZ7fQi7a85Y6KIzOi0TtiJXSML13FTFqONNgoc/edit?tab=t.0#heading=h.rec62wnvc4ft) that can occur after a negative action taken for a policy violation.
  
X, Y and Z are hardcoded in constants for now according to what Operations need. The name of the action could be like `amo-delayed-short-soft-block` or `amo-delayed-mid-hard-block` and `amo-delayed-long-hard-block` for instance (or maybe `amo-hard-block-in-short-delay`, `amo-hard-block-in-medium-delay` etc) to allow us to tweak the constants independently.

Those blocks should use `BlocklistSubmission` with `disable_addon=False` as the add-on should already be disabled when they are executed. Similarly, whenever we execute such a `Blocklistsubmission`, relevant versions should have been rejected already, so possibly look into adding a safeguard to prevent us from calling `disable_versions_for_block()` when creating the block ?

Those actions should not happen on their own. Automatically unblocking of add-on/versions upon successful appeal will be handled in https://github.com/mozilla/addons/issues/16119 

### Acceptance Criteria

- [ ] item 1

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/AMOENG-2369)

eviljeff:
Fixes: mozilla/addons#16118

### Description

Adds support for specific follow-up actions as policy enforcements, and action classes for delayed block submissions.

### Context

As per the issue, this patch doesn't handle any override or appeals for these new actions.  We may need to address messaging (emails) in follow-ups too.  Right now the follow-up actions would be silent.

### Testing

- Add one of the new enforcement follow-up actions to a policy in Cinder.
- in Cinder, moderate an add-on with that policy
- replay the webhook
- see that the normal action is enforced immediately (assuming it doesn't need to be held for 2nd level)
- and that an BlocklistSubmission instance is created, delayed for whatever amount of time the action was set to (and hard/soft as per the action)

### Checklist


src/olympia/abuse/actions.py
```diff
@@ -695,6 +699,75 @@ def hold_action(self):
...
+        return None
+
+    def get_owners(self):
+        # TODO: Define our own email strategy and template copy
+        return ()
+
+
+class ContentActionDelayedShortSoftBlockAddon(_ContentActionDelayedBlockAddon):
+    block_type = BlockType.SOFT_BLOCKED
+    delay_days = 7
```
diox, line 743: We should use constants for the 3 delays to stay consistent IMHO

src/olympia/abuse/models.py
```diff
@@ -1503,6 +1515,26 @@ def get_policy_texts(self):
         )
 
 
+class ContentDecisionFollowupAction(ModelBase):
```
diox, line 1518: Since it will come up: we need a ticket to expose that table in Redash once it exists.

diox:
I noticed you added to Cinder staging:
```
AMO_FU_DELAY_LONG_HARD_BLOCK_ADDON	amo_fu_delay_long_hard_block_addon
AMO_FU_DELAY_LONG_SOFT_BLOCK_ADDON	amo_fu_delay_long_soft_block_addon
AMO_FU_DELAY_MID_HARD_BLOCK_ADDON	amo_fu_delay_mid_hard_block_addon
AMO_FU_DELAY_MID_SOFT_BLOCK_ADDON	amo_fu_delay_mid_soft_block_addon
AMO_FU_DELAY_SHORT_HARD_BLOCK_ADDON	amo_fu_delay_short_hard_block_addon
AMO_FU_DELAY_SHORT_SOFT_BLOCK_ADDON	amo_fu_delay_short_soft_block_addon
```

Should we use a more descriptive name in Cinder for the first column, and a slug with dashes instead of underscores for the second ?

eviljeff:
> I noticed you added to Cinder staging:
> 
> ```
> AMO_FU_DELAY_LONG_HARD_BLOCK_ADDON	amo_fu_delay_long_hard_block_addon
> AMO_FU_DELAY_LONG_SOFT_BLOCK_ADDON	amo_fu_delay_long_soft_block_addon
> AMO_FU_DELAY_MID_HARD_BLOCK_ADDON	amo_fu_delay_mid_hard_block_addon
> AMO_FU_DELAY_MID_SOFT_BLOCK_ADDON	amo_fu_delay_mid_soft_block_addon
> AMO_FU_DELAY_SHORT_HARD_BLOCK_ADDON	amo_fu_delay_short_hard_block_addon
> AMO_FU_DELAY_SHORT_SOFT_BLOCK_ADDON	amo_fu_delay_short_soft_block_addon
> ```
> 
> Should we use a more descriptive name in Cinder for the first column, and a slug with dashes instead of underscores for the second ?

mmm, yes I messed up by bulk-copying the enforcement actions over after adjusting the naming - the slugs need to have dashes.  

The naming is restricted though, because the slug is generated by Cinder from the name by lower-casing the letters, swapping out `-` for ` `, and stripping out most other punctuation. So for `AMO_FU_DELAY_LONG_HARD_BLOCK_ADDON`, we have to have a slug of `amo-fu-delay-long-hard-block-addon` for our integration to work, and pretty much the best we can name it, while keeping the slug, is something like `AMO [f/u] delay-long hard-block addon`

eviljeff:
We could adjust the constant/slug naming if you have a better scheme?

diox:
I'm happy with the names/slugs now that they are fixed

src/olympia/abuse/actions.py
eviljeff: I thought perhaps Operations would want different definitions of "short", "mid", "long" for soft and hard blocks.  E.g. the short delay for a soft block is 5 days, and the short delay for a hard bock is 10 days (or whatever).

src/olympia/abuse/models.py
eviljeff: https://mozilla-hub.atlassian.net/browse/AMOENG-2457

src/olympia/abuse/actions.py
diox: mmm fair. I hope that won't happen though, cause it would be quite confusing...

diox (approved):

PR #24785, opened 2026-04-20
Give ContentAction classes an action attribute

diox:
This will allow us to have some logic inside the content action classes that depend on the action being executed by that class, e.g. we could have `ContentActionDisableAddon` look at past `AMO_DISABLE_ADDON` actions when processing and deciding not to do it again.

Pre-requisite for https://github.com/mozilla/addons/issues/16123

### DIFF
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -344,14 +344,6 @@ class ContentActionDisableAddon(ContentAction):
     reporter_appeal_template_path = 'abuse/emails/reporter_appeal_takedown.txt'
     action = DECISION_ACTIONS.AMO_DISABLE_ADDON
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.target_versions = (
-            self.decision.target_versions.all()
-            if self.decision.id
-            else Version.objects.none()
-        )
-
     def should_hold_action(self):
         return bool(
             self.target.status != amo.STATUS_DISABLED
@@ -389,6 +381,10 @@ class ContentActionDisableAddon(ContentAction):
             attachment.update(activity_log=activity_log)
             activity_log.attachmentlog = attachment  # update fk
         return activity_log
+
+    @property
+    def target_versions(self):
+        return self.decision.target_versions.all()
 
     @property
     def versions_force_disable_will_affect(self):
@@ -790,9 +786,9 @@ class ContentActionRejectListingContent(ContentActionDisableAddon):
     description = 'Add-on listing content has been rejected'
     action = DECISION_ACTIONS.AMO_REJECT_LISTING_CONTENT
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.target_versions = self.decision.target_versions.none()
+    @property
+    def target_versions(self):
+        return self.decision.target_versions.none()
 
     def should_hold_action(self):
         return bool(
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -1,6 +1,7 @@
 import random
 from collections import defaultdict
 from datetime import datetime, timedelta
+from inspect import isclass
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
@@ -44,6 +45,7 @@ class ContentAction:
     second_level_notification_template_path = (
         'abuse/emails/second_level_notification.txt'
     )
+    action = None
 
     def __init__(self, decision):
         self.decision = decision
@@ -304,6 +306,7 @@ class ContentActionBanUser(ContentAction):
     valid_targets = (UserProfile,)
     reporter_template_path = 'abuse/emails/reporter_takedown_user.txt'
     reporter_appeal_template_path = 'abuse/emails/reporter_appeal_takedown.txt'
+    action = DECISION_ACTIONS.AMO_BAN_USER
 
     def should_hold_action(self):
         return bool(
@@ -339,6 +342,15 @@ class ContentActionDisableAddon(ContentAction):
     valid_targets = (Addon,)
     reporter_template_path = 'abuse/emails/reporter_takedown_addon.txt'
     reporter_appeal_template_path = 'abuse/emails/reporter_appeal_takedown.txt'
+    action = DECISION_ACTIONS.AMO_DISABLE_ADDON
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.target_versions = (
+            self.decision.target_versions.all()
+            if self.decision.id
+            else Version.objects.none()
+        )
 
     def should_hold_action(self):
         return bool(
@@ -379,10 +391,6 @@ class ContentActionDisableAddon(ContentAction):
         return activity_log
 
     @property
-    def target_versions(self):
-        return self.decision.target_versions.all()
-
-    @property
     def versions_force_disable_will_affect(self):
         return (
             Version.objects.all()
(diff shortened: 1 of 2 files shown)


src/olympia/abuse/tests/test_actions.py
```diff
@@ -300,8 +300,10 @@ def test_owner_content_approve_report_email(self):
         assert 'has been approved' in mail.outbox[-1].body
 
     def test_notify_reporters_reporters_provided(self):
-        action = self.ActionClass(self.decision)
-        action.notify_reporters(reporter_abuse_reports=[self.abuse_report_no_auth])
+        action_helper = self.ActionClass(self.decision)
```
diox, line 307: I tried to make the tests in this file more consistent by:
- Naming the content action class variable `ActionClass` everywhere
- Naming the variable with the instance of the action class `action_helper`
- Naming the variable with the actual action `action`

That's why I ended up with +261-252 in this test file.

src/olympia/abuse/actions.py
```diff
@@ -954,6 +971,7 @@ class ContentActionApproveListingContent(AnyTargetMixin, ContentAction):
     description = 'Reported content is within policy'
     reporter_template_path = 'abuse/emails/reporter_content_approve.txt'
     reporter_appeal_template_path = 'abuse/emails/reporter_appeal_approve.txt'
+    action = DECISION_ACTIONS.AMO_APPROVE
```
eviljeff, line 1049: This action is used by 2 classes, so the logic is ambiguos (this will overwrite `ContentActionTargetAppealApprove` if `vars.values` returns classes in the order declared in the file).  

I think we need an additional check (or test) somewhere to make sure we don't declare an action in more than one class.

src/olympia/abuse/actions.py
```diff
@@ -1032,23 +1054,8 @@ class ContentActionNotImplemented(AnyTargetMixin, NoActionMixin, ContentAction):
...
-        DECISION_ACTIONS.AMO_IGNORE: ContentActionIgnore,
-        DECISION_ACTIONS.AMO_CLOSED_NO_ACTION: ContentActionAlreadyModerated,
-        DECISION_ACTIONS.AMO_LEGAL_FORWARD: ContentActionForwardToLegal,
-        DECISION_ACTIONS.AMO_CHANGE_PENDING_REJECTION_DATE: (
-            ContentActionChangePendingRejectionDate
-        ),
-        DECISION_ACTIONS.AMO_REJECT_LISTING_CONTENT: ContentActionRejectListingContent,
+        local_.action: local_
+        for local_ in vars().values()
+        if isclass(local_) and issubclass(local_, ContentAction) and local_.action
```
eviljeff, line 1133: I almost added something like this in https://github.com/mozilla/addons-server/pull/24732 😃

src/olympia/abuse/actions.py
```diff
@@ -992,12 +1010,14 @@ class ContentActionApproveInitialDecision(
...
     reporter_template_path = 'abuse/emails/reporter_content_approve.txt'
     reporter_appeal_template_path = 'abuse/emails/reporter_appeal_approve.txt'
+    action = DECISION_ACTIONS.AMO_APPROVE_VERSION
 
 
 class ContentActionTargetAppealRemovalAffirmation(
     AnyTargetMixin, AnyOwnerEmailMixin, ContentAction
 ):
     description = 'Reported content is still offending, after appeal.'
+    action = None  # FIXME: or DECISION_ACTIONS.AMO_IGNORE? Does it matter?
```
eviljeff: `None` is correct - the action will depend on what the original action was; there isn't a specific appeal affirmation enforcement action.

src/olympia/abuse/actions.py
```diff
@@ -339,6 +342,15 @@ class ContentActionDisableAddon(ContentAction):
     valid_targets = (Addon,)
     reporter_template_path = 'abuse/emails/reporter_takedown_addon.txt'
     reporter_appeal_template_path = 'abuse/emails/reporter_appeal_takedown.txt'
+    action = DECISION_ACTIONS.AMO_DISABLE_ADDON
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.target_versions = (
```
eviljeff: I'm not sure I entirely follow the reasoning for why setting `self.target_versions` in the `__init__`, and allowing `target_versions` on the class instance to be set or get, gives a different outcome than deferring to the decision.target_versions, and allowing `.target_versions` to be calculated independently and/or get or set on decision instead.  We don't want `ContentDecision.process_action` to be called without a saved decision (or... do we?)

eviljeff: `ContentActionTargetAppealApprove` has a different implementation of `target_versions` which you aren't overriding here so how would that work?

eviljeff: also, `self.decision.target_versions` is set specifically to `self.versions_force_disable_will_affect` in `process_action`, (plus in `ContentActionBlockAddon`, to `self.versions_block_will_affect`) so after this patch `self.target_versions` and `self.decision.target_versions` will be different - which is the "correct" one?

eviljeff (commented):
The .action change is sensible; the .target_versions change seems to be introducing footguns with different values for target_versions on `self` and `self.decision` :/

src/olympia/abuse/actions.py
diox: I'll split this into a different PR. The main problem I had is that `target_versions` on the `ContentDecision` is a M2M, so it requires a `ContentDecision` that has been saved to the database, or throws an error.

I want to be able to instantiate `ContentAction` classes with a `ContentDecision` that hasn't been saved yet, and then call a new method that checks if their `target_versions` match a previous `ContentDecision`, to determine if we should proceed with that action from a scanner and possibly throw away that unsaved `ContentDecision` and the action if we are not.

src/olympia/abuse/actions.py
diox: mmm I guess appeals shouldn't set an `action` then ? or be excluded from the dict somehow ? They are a bit special...

eviljeff: Yeah, I think the appeal action class shouldn't define an action

src/olympia/abuse/actions.py
eviljeff: I don't think that's _impossible_ to achieve, but it'll need some refactoring so we aren't only setting the correct `target_versions` in `process_action` - it's too late at that point.

src/olympia/abuse/actions.py
diox: This conflicts with it btw, so I think I'll let you merge yours first...

src/olympia/abuse/actions.py
diox: Removed for now in https://github.com/mozilla/addons-server/pull/24785/commits/3a5e6fea07f6e33ab002f5fa67041ac686fcf5d3 ; I'll come back to that once this is merged.

src/olympia/abuse/tests/test_actions.py
```diff
@@ -2895,3 +2910,32 @@ def test_hold_action(self):
...
+            for v in vars(olympia.abuse.actions).values()
+            if isclass(v) and issubclass(v, ContentAction) and v.action
+        ]
+    )
+    assert len(actions_from_content_action_classes) == len(
+        set(actions_from_content_action_classes)
+    )
+
+    # There should be at least as many DECISION_ACTIONS as there are classes,
+    # except for AMO_ESCALATE_ADDON (which is obsolete) and AMO_REQUEUE (which
```
eviljeff, line 2931: just thinking out loud at this point, but we should probably create an empty class for AMO_REQUEUE (that possibly raises in it's process_action), and delete AMO_ESCALATE_ADDON.

eviljeff (approved):

src/olympia/abuse/tests/test_actions.py
diox: I almost did but decided it was not worth spending the time on that right now...

PR #24789, opened 2026-04-21
Add an explicit target_versions attribute on ContentAction classes

diox:
Having an actual target_versions attribute rather than a property depending on the decision will allow us to have logic that depends on the target versions inside the action class but still allow the target versions considered for that logic to be set by the caller before a decision has been made.

A `set_target_versions()` method is provided to keep the decision target versions in sync with the ContentAction.

Split from https://github.com/mozilla/addons-server/pull/24785
Helps mozilla/addons#16123

### DIFF
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -57,6 +57,11 @@ class ContentAction:
                 or self.target.current_version
                 or self.target.find_latest_version(channel=None, exclude=())
             )
+            self.target_versions = (
+                self.decision.target_versions.all()
+                if self.decision.id
+                else Version.objects.none()
+            )
 
         if not isinstance(self.target, self.valid_targets):
             raise ImproperlyConfigured(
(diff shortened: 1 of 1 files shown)


diox:
The end goal I'm after here is being able to add a `should_be_skipped_by_automation()` method on `ContentAction`. It would look like this in the base class (only caring about `Addon` entities for this demonstration):

```python
def should_be_skipped_by_automation(self):
    return ContentDecision.objects.filter(self.addon, action=self.action).exists()
```

But then `ContentActionRejectVersion` would have this:
```python
def should_be_skipped_by_automation(self):
    return ContentDecision.objects.filter(
        addon=self.target, versions__in=self.target_versions, action=self.action
    ).exists()
```

The problem is, I need to perform this check before the scanner decision has actually taken place. So I'm instantiating the relevant class without a saved `ContentDecision`, so I don't have access to `decision.target_versions`:

```python
ContentActionClass = CONTENT_ACTION_FROM_DECISION_ACTION[action]
action_helper = ContentActionClass(ContentDecision(addon=addon))
action_helper.target_versions = Version.unfiltered.filter(pk=version.pk)
if not action_helper.should_be_skipped_by_automation():
    # Proceed ...
```

This is done in a loop for one or more actions, the `action_helper` and its decision are thrown away after this - I just needed some place to stick the `should_be_skipped_by_automation()` logic, and since it is action-dependent, the `ContentAction` classes was the logical place for it, but since I don't have an actual decision at that point, I need the ability to override `target_versions` being considered.

Alternatively, I could give up on using `target_versions` for this, and just pass the `version` to my `should_be_skipped_by_automation()` method, and have it ignore `self.target_versions` entirely - automation is always processing versions one by one anyway. Is that better ? Let me know what you think.

diox:
> Alternatively, I could give up on using `target_versions` for this, and just pass the `version` to my `should_be_skipped_by_automation()` method, and have it ignore `self.target_versions` entirely - automation is always processing versions one by one anyway. Is that better ? Let me know what you think.

Picked that route instead after talking with @eviljeff on slack.

PR #24796, opened 2026-04-22
Implement policy-based scanner actions

### LINKED ISSUE #16123: [Task]: Add ability for `ScannerRule` to execute enforcement actions from a `CinderPolicy`
### Description

Like https://github.com/mozilla/addons/issues/16122 but for scanners: add ability for scanners rules to link to a `CinderPolicy` and have a special kind of "Automatic" action.

When we run scanner actions, execute the most impactful action as normal, but on top of this, for the "Automatic" action:
- Find all `CinderPolicy` all matched rules are attached to, and create and execute a decision with all the relevant enforcement actions (depends on https://github.com/mozilla/addons/issues/16120)

This would eventually replace the `Disable and block` scanner action, as this action would be made available to specific policies. It would also allow scanners to do a rejection for instance (and possible follow-up delayed blocks). 

Open question: how to re-implement the safeguards we have in that "Disable and block" scanner action? We want to continue avoiding automatic decisions if there are successful appeals or if the add-on belongs to a usage tier that should disallow that.

### Acceptance Criteria

- [ ] item 1

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/AMOENG-2375)

diox:
This allows scanners rules to have a policy attached. When a version is scanned and hits rules with policy attached, we gather the policies, find the most aggressive action attached to them, and execute it.

Fixes https://github.com/mozilla/addons/issues/16123


src/olympia/scanners/models.py
```diff
@@ -411,26 +427,117 @@ def save(self, *args, **kwargs):
...
     @classmethod
-    def run_action(cls, version):
-        """Try to find and execute an action for a given version, based on the
-        scanner results and associated rules.
+    def run_actions(cls, version):
+        cls.execute_workflow_action(version)
+        cls.execute_policy_enforcement_action(version)
+
+    @classmethod
+    def execute_policy_enforcement_action(cls, version):
```
diox, line 436: These (plus the addition of the `policy` FK and the admin changes that go with it) are the only real changes, but I had to rename `run_action` to `run_actions` since it does a bit more than before as well as write a bunch of tests, which inflates the diff size quite a bit.

src/olympia/abuse/actions.py
```diff
@@ -567,10 +568,12 @@ def __init__(self, decision):
                 self.delayed_rejection_date - self.decision.created
             ).days
         else:
-            # Will fail later if we try to use it to log/process the action,
-            # but allows us to at least instantiate the class and use other
-            # methods.
-            self.delayed_rejection_date = self.delayed_rejection_days = None
+            self.delayed_rejection_days = REVIEWER_DELAYED_REJECTION_PERIOD_DAYS_DEFAULT
```
diox, line 571: We need a default here since scanners won't know what delay to use otherwise.

src/olympia/scanners/models.py
```diff
@@ -411,18 +428,105 @@ def save(self, *args, **kwargs):
             self.matched_rules.add(scanner_rule)
 
     @classmethod
-    def run_action(cls, version):
-        """Try to find and execute an action for a given version, based on the
-        scanner results and associated rules.
+    def run_actions(cls, version):
+        cls.execute_workflow_action(version)
+        cls.execute_policy_enforcement_action(version)
```
eviljeff, line 433: Should we try to execute both a workflow action (old style rule) and policy enforcement (new style)?  It would be safer to skip one if the other succeeds in doing something (e.g. don't try execute_policy_enforcement_action if execute_workflow_action does anything other than NO_ACTION, or visa versa, depending on what we think should take precedence).

Alternatively, add a db constraint that means a rule can't have both a policy and action defined.

src/olympia/scanners/models.py
```diff
@@ -284,7 +289,19 @@ def save(self, *args, **kwargs):
...
+        help_text='Workflow action to execute',
+    )
+    policy = models.ForeignKey(
+        'abuse.CinderPolicy',
+        on_delete=models.SET_NULL,
+        null=True,
+        help_text=(
+            'Policy used to automatically derive an action from when the rule hit.'
+        ),
+        limit_choices_to={'expose_in_reviewer_tools': True},
```
eviljeff, line 304: Operations might want a policy that's specific to automation in some cases, that wouldn't be exposed in reviewer tools(?)

eviljeff (approved):

src/olympia/scanners/models.py
diox: I wanted to allow both at the same time as a feature, because we might want to apply restrictions or delay auto-approval at the same time we're executing a policy action.

src/olympia/scanners/models.py
diox: I would hope not, but we can cross that bridge when get get to it.

PR #24800, opened 2026-04-27
AMOENG-2475 - Use the correct FxA endpoints for the contact support form (stage/prod)

bakulf:
Fixes: AMOENG-2475

### Description

Dev/Stage/Prod need to use the correct FxA endpoints in order to create zendesk tickets via the contact support form.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -115,7 +115,7 @@ DATABASES = {
 FXA_CONTENT_HOST = 'https://accounts.stage.mozaws.net'
 FXA_OAUTH_HOST = 'https://oauth.stage.mozaws.net/v1'
 FXA_PROFILE_HOST = 'https://profile.stage.mozaws.net/v1'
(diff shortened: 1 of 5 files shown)


diox (changes requested):
`settings.py` (used for local dev) still references `https://accounts.stage.mozaws.net/v1` instead of `https://api-accounts.stage.mozaws.net/v1`

src/olympia/conf/stage/settings.py
```diff
@@ -68,6 +68,8 @@
     'REMOTE_SETTINGS_WRITER_URL', default='https://remote-settings.allizom.org/v1/'
 )
 
+FXA_SUPPORT_HOST = 'https://api.accounts.firefox.com/v1'
```
eviljeff: you don't need to redeclare this in stage settings if it's the same as settings_base - it inherits (imports) all the base settings.

diox (approved):

PR #24807, opened 2026-04-28
Add logic to find actions that can be automatically executed from a list of policies

diox:
Needed for https://github.com/mozilla/addons/issues/16123

This is split from the larger PR that implements the whole thing to make it easier to review. It contains all the logic, but isn't used yet.

Changes:
- Moves and splits `filter_enforcement_actions()` so that it focuses on filtering action enums, adding another util, `to_enforcement_actions()` to transform a list of api values into the enums. 
- Adds `should_be_skipped_by_automation()` to scanner actions to determine whether automated action can use this action for the given add-on and version
- Adds `find_automated_enforcement_actions_from_policies()` to find the appropriate automated action to take for a given add-on, version and list of policies.

That last function is what the scanners automated enforcement actions code will use later (`ScannerRule` will gain a FK to `CinderPolicy`).

### DIFF
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -822,13 +822,7 @@ class ContentActionRejectListingContent(ContentActionDisableAddon):
     @classmethod
     def should_be_skipped_by_automation(cls, *, addon, version, **kwargs):
         # Only consider this action in automation if the version is listed.
-        return (
-            super().should_be_skipped_by_automation(
-                addon=addon, version=version, **kwargs
-            )
-            if version.channel == amo.CHANNEL_LISTED
-            else True
-        )
+        return version.channel != amo.CHANNEL_LISTED
 
     def should_hold_action(self):
         return bool(
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -103,6 +103,11 @@ class ContentAction:
         without further checks, and true if it should be held for further review."""
         return False
 
+    @classmethod
+    def should_be_skipped_by_automation(cls, **kwargs):
+        """Return True if the action should be skipped by automation for any reason."""
+        return False
+
     def process_action(self, release_hold=False):
         """This method should return an activity log instance for the action,
         if available."""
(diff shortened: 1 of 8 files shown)


src/olympia/abuse/actions.py
```diff
@@ -644,6 +649,24 @@ def should_hold_action(self):
             and any(self.target.promoted_groups(currently_approved=False).high_profile)
         )
 
+    @classmethod
+    def should_be_skipped_by_automation(cls, *, addon, version, **kwargs):
```
diox, line 653: This replicates what the scanner action does in https://github.com/mozilla/addons-server/blob/ac6079943382e69e71ef85993903bdb0b7b75aa6/src/olympia/scanners/actions.py#L141-L172

src/olympia/abuse/actions.py
```diff
@@ -790,6 +819,17 @@ class ContentActionRejectListingContent(ContentActionDisableAddon):
...
+        # Only consider this action in automation if the version is listed.
+        return (
+            super().should_be_skipped_by_automation(
+                addon=addon, version=version, **kwargs
+            )
+            if version.channel == amo.CHANNEL_LISTED
+            else True
+        )
```
eviljeff: This currently resolves down to `return version.channel != amo.CHANNEL_LISTED`, right?

src/olympia/abuse/utils.py
```diff
@@ -67,3 +68,94 @@ def is_same_time(instance, created_string):
...
+            or ContentActionClass.should_be_skipped_by_automation(
+                addon=addon, version=version
+            )
+        ):
+            continue
+
+        if addon_negative_actions_priority_from_iterable(
+            (primary_enforcement_actions, followup_actions)
+        ) > addon_negative_actions_priority_from_iterable(most_aggressive_actions):
+            most_aggressive_actions = [primary_enforcement_actions, followup_actions]
```
eviljeff, line 160: I'm reading this as meaning that:
DISABLE_ADDON, FU_DELAY_SHORT_SOFT_BLOCK_ADDON would be seen as more aggressive than
DISABLE_ADDON, FU_DELAY_MID_HARD_BLOCK_ADDON.

Which... firstly is it? (it's more _immediate_, but is hard not ultimately more aggressive than soft?)  Secondly, should we be concatenating some followup actions instead - i.e. the end result would include both the soft and hard actions.

eviljeff (approved):
For the primary actions I agree it's appropriate (and necessary) to have a single action, and the ordering/logic seems sensible.  For the follow-up actions there's more nuance.

src/olympia/abuse/actions.py
diox: Yeah I guess I could simplify it down to that.

src/olympia/abuse/utils.py
diox: My thinking was that we shouldn't concatenate follow-up actions because that makes the decision consequences harder to follow for both developers and reviewers. 

Because it is more immediate, I considered the policy with the shorter delay as the "right" policy to use, but we could change that down the line if Operations disagree.

src/olympia/abuse/actions.py
diox: https://github.com/mozilla/addons-server/pull/24807/commits/407b3d29d264b8044232b0bfd5a773ca664e9625

PR #24808, opened 2026-04-28
Make configuration of scanner rules possible immediately at creation

### LINKED ISSUE #16048: [Task]: Make configuration of NARC scanner rules possible immediately at creation
### Description

https://github.com/mozilla/addons-server/pull/24388 added the ability to configure NARC rules, but there is a known limitation at the moment: the configuration is only available when addons-server "knows" what scanner the rule is for, so on the creation UI, the configuration bits don't show up, you need to submit the form first.

We should dynamically show the relevant configuration according to the scanner being selected, through some JavaScript + an endpoint that would return the default config schema and re-trigger the jsonform widget init

### Acceptance Criteria

- [ ] NARC rules configuration is available at creation

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/AMOENG-2282)

diox:
(Only NARC has configuration for the moment, but this should work for any scanner in theory)

Fixes: mozilla/addons#16048

### Description

This allows admins to see and change configuration of scanner rules and scanner query rules directly at creation, rather than have to submit the form first.

### Context

django-jsonform is only initialized once by default, but we have dynamic scanner selection, so dynamic initialization is needed.

### Testing

- In the admin, go to the add a scanner rule page
- Change the scanner to NARC

Expected results:
- Configuration should appear

Saving the rule should correctly save the configuration and display it again on edit.

Note: configuration is reset if you change scanner back and forth.

### DIFF
--- a/src/olympia/constants/scanners.py
+++ b/src/olympia/constants/scanners.py
@@ -140,7 +140,7 @@ NARC_RULE_CONFIGURATION_SCHEMA = {
             'type': 'boolean',
             'default': True,
             'helpText': (
-                'For each string being examined, also examine homoglyphs variants',
+                'For each string being examined, also examine homoglyphs variants'
             ),
         },
     },
--- a/src/olympia/scanners/templates/django_jsonform/editor.html
+++ b/src/olympia/scanners/templates/django_jsonform/editor.html
@@ -1,0 +1,5 @@
+{%  load scanners %}
+{% scanners_configuration_schemas %}
+<div data-django-jsonform-container="true" {% include "django_jsonform/attrs.html" %}></div>
+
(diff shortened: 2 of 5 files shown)


src/olympia/constants/scanners.py
```diff
@@ -140,7 +140,7 @@
             'type': 'boolean',
             'default': True,
             'helpText': (
-                'For each string being examined, also examine homoglyphs variants',
```
diox, line 143: Fun fact: this string was accidentally a tuple, yet somehow it still worked as a help text when displaying the configuration.

willdurand (approved):

PR #24812, opened 2026-04-28
Remove db_constraint on TranslatedField as translations are not unique

diox:
We have been using this quirky "feature" of MySQL that allows FKs to point to non-unique rows in the translations app for 15+ years, but MySQL 8.4 raises an error unless `--skip-restrict-fk-on-non-standard-key` is set. This parameter is not supported on GCP, so remove the constraints instead.

Unlocks https://github.com/mozilla/addons/issues/15918


docker-compose.yml
```diff
@@ -132,8 +132,6 @@ services:
       - --skip-name-resolve
       # Disable performance schema for faster startup
       - --performance-schema=OFF
-      # Allow nonstandard FKs (needed for translations)
-      - --skip-restrict-fk-on-non-standard-key
```
diox, line 136: https://github.com/mozilla/addons-server/pull/24894 proves just removing `--skip-restrict-fk-on-non-standard-key` in MySQL 8.4 without the changes from this PR loudly fails in CI.

src/olympia/translations/migrations/0004_remove_legacy_constraints.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Remove Foreign Key constraints to Translations.id - TranslatedField should
+    now be using db_constraint: False as there can be multiple translation rows
+    sharing the same id, one per locale per translated string.
+    """
+    Translation = apps.get_model('translations', 'Translation')
+    models_related_to_translations = {
+        f.related_model
+        for f in Translation._meta.get_fields(include_hidden=True)
+        if f.auto_created and not f.concrete and (f.one_to_one or f.one_to_many)
```
eviljeff: nit, it's only a migration, but: I don't think most of this is needed for Translation - `f.is_relation` should be sufficient

eviljeff (approved):

src/olympia/translations/migrations/0004_remove_legacy_constraints.py
diox: Ah I didn't know about that one...

diox: https://github.com/mozilla/addons-server/pull/24812/commits/891fe3308a5fd1ef1cf107c4c2f595a5b13c934b

PR #24820, opened 2026-04-30
Keep continue-on-error in _test_check.yml for now

diox:
Removing it in https://github.com/mozilla/addons-server/commit/3750d49fe4cabfc2881bc319a653a1cf053ed1ee broke the build on `master` because of the DOCKER_COMMIT and DOCKER_VERSION are set to 'not-expected' after... This is not happening in PRs...

I was hoping this issue would be a quick fix but we'll have to deal with it later.


eviljeff (approved):
:(

PR #24824, opened 2026-04-30
Always record Promoted Approval when approving add-on versions

### LINKED ISSUE #16165: [Bug]: Warning banner is shown for recommended add-ons if flipping `listed_pre_review` for the promoted group

### What happened?

STR:
- Change `listed_pre_review` on Recommended Promoted Group to `False`
- Submit a new version of a recommended add-on
- Let it be auto-approved (remove any other flag that would prevent it, don't manually approve)
- Visit the detail page: 
  - The warning banner about non monitored add-ons should not be shown
  - The recommended badge should be shown
- Change `listed_pre_review` on Recommended Promoted Group back to `True`
- Visit the detail page again

Actual result:
- The warning banner about non monitored add-ons is shown.
- The recommended badge is not shown.

Expected result:
- The warning banner about non monitored add-ons should not be shown.
- The recommended badge should be shown.

This is a consequence of the original promoted add-ons design: when an add-on version belonging to a promoted group gets approved, [if the promoted group required pre-review for the channel the version was submitted for, we record a promotion approval](https://github.com/mozilla/addons-server/blob/97b8e9149502f2d76f02c855cc85710593868705/src/olympia/reviewers/utils.py#L1065-L1071). This allows then to say that this version has been approved while the add-on was promoted, to distinguish it from versions submitted before that. If the promoted group doesn't require pre-review, we don't bother recording anything.

The API then exposes [what promotions the add-on current version was approved for](https://github.com/mozilla/addons-server/blob/97b8e9149502f2d76f02c855cc85710593868705/src/olympia/addons/models.py#L1695-L1710) to the frontend, which uses it to determine if we should show the warning banner telling users to be careful about add-ons that are not actively monitored. 

Because we only record that promoted approval information if the promoted group require pre-review for the channel, changing the fact that a promoted group requires pre-review results in the banner being shown.

There is a shortcut in the model code that considers the promoted approved for the version if both `listed_pre_review` and `unlisted_pre_review` are `False`, which is why in the STR above the banner it not shown until `listed_pre_review` is flipped back on.

Fixing this likely means:
- Always record promoted approval when an add-on gets approved
- Do that even if it's not an human review
- Do that for both channels (less impactful since there is no listing and therefore no banner, but for unlisted, `set_promoted()` isn't even called currently for non-human review)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/AMOENG-2429)

diox:
This allows the `listed_pre_review` and `unlisted_pre_review` flags to be flipped back and forth for promoted group without any negative consequences for the versions approved while they are off: they would still have their approval recorded, so they would still be considered approved for the promoted group if the flags are turned back on.

Fixes https://github.com/mozilla/addons/issues/16165

### DIFF
--- a/src/olympia/reviewers/tests/test_utils.py
+++ b/src/olympia/reviewers/tests/test_utils.py
@@ -46,6 +46,7 @@ from olympia.lib.crypto.tests.test_signing import (
 )
 from olympia.promoted.models import (
     PromotedApproval,
+    PromotedGroup,
 )
 from olympia.reviewers.models import (
     AutoApprovalSummary,
@@ -3886,27 +3887,39 @@ class TestReviewHelper(TestReviewHelperBase):
         assert PROMOTED_GROUP_CHOICES.LINE in self.addon.promoted_groups().group_id
 
     def test_autoapprove_promoted(self):
+        group = PromotedGroup.objects.get(group_id=PROMOTED_GROUP_CHOICES.RECOMMENDED)
+        group.listed_pre_review = False
+        group.save()
         self.make_addon_promoted(self.addon, PROMOTED_GROUP_CHOICES.RECOMMENDED)
-        assert not self.addon.promoted_groups()
         self.user = UserProfile.objects.get(id=settings.TASK_USER_ID)
 
         self.test_nomination_to_public()
         assert PromotedApproval.objects.filter(
             version=self.addon.current_version
         ).exists()
-        assert self.addon.promoted_groups()
+        assert group in self.addon.promoted_groups()
+
+        group.listed_pre_review = True
+        group.save()
+        assert group in self.addon.promoted_groups()
 
     def test_autoapprove_promoted_notable(self):
         # as above with notable
+        group = PromotedGroup.objects.get(group_id=PROMOTED_GROUP_CHOICES.NOTABLE)
+        group.listed_pre_review = False
+        group.save()
         self.make_addon_promoted(self.addon, PROMOTED_GROUP_CHOICES.NOTABLE)
-        assert not self.addon.promoted_groups()
         self.user = UserProfile.objects.get(id=settings.TASK_USER_ID)
 
         self.test_nomination_to_public()
         assert PromotedApproval.objects.filter(
             version=self.addon.current_version
         ).exists()
-        assert self.addon.promoted_groups()
+        assert group in self.addon.promoted_groups()
+
+        group.listed_pre_review = True
+        group.save()
+        assert group in self.addon.promoted_groups()
 
     def _test_block_multiple_unlisted_versions(self, redirect_url):
         old_version = self.review_version
--- a/src/olympia/reviewers/tests/test_utils.py
+++ b/src/olympia/reviewers/tests/test_utils.py
@@ -45,7 +45,6 @@ from olympia.lib.crypto.tests.test_signing import (
     _get_signature_details,
 )
 from olympia.promoted.models import (
-    PromotedAddon,
     PromotedApproval,
 )
 from olympia.reviewers.models import (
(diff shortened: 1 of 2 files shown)


src/olympia/reviewers/utils.py
```diff
@@ -1056,19 +1055,11 @@ def set_file(self, status, file):
...
         elif not versions:
             return
-        channel = versions[0].channel
-        if group and (
-            (channel == amo.CHANNEL_LISTED and any(group.listed_pre_review))
-            or (channel == amo.CHANNEL_UNLISTED and any(group.unlisted_pre_review))
-        ):
-            # These addons shouldn't be be attempted for auto approval anyway,
-            # but double check that the cron job isn't trying to approve it.
-            assert not self.user.id == settings.TASK_USER_ID
```
eviljeff, line 1071: We still _want_ this assertion, don't we? Perhaps not here, but somewhere. 

I.e. the issue/request is to record the promoted approval for groups that aren't pre_review=True too (so, all groups), not to actually auto approve pre_review=True promoted groups.

diox: Well that exists in `AutoApprovalSummary` already, `check_is_promoted_prereview()` will check and refuse to auto-approve already if necessary.

eviljeff: So we're assuming that AutoApprovalSummary would never end up triggering auto approval on an add-on in a promoted group with pre_review=True, so we don't need that additional assertion here.  (I'm fine with that, if that's what you want)

src/olympia/reviewers/tests/test_utils.py
```diff
@@ -3886,38 +3885,28 @@ def test_approved_update_other_promoted(self):
         ).exists()
         assert PROMOTED_GROUP_CHOICES.LINE in self.addon.promoted_groups().group_id
 
-    def test_autoapprove_fails_for_promoted(self):
+    def test_autoapprove_promoted(self):
```
eviljeff, line 3889: Just to point out, this test is testing a scenario which is currently not possible - Recommended is pre_review=True, so the auto approval command would not attempt to approve the nomination.

eviljeff (approved):

src/olympia/reviewers/utils.py
diox: My thinking was, auto-approval is not the responsibility of this class - it's what `AutoApprovalSummary` and `auto_approve` are for. But writing this comment, I see your initial point, I didn't have to remove the assertion... I'll look into adding it back.

src/olympia/reviewers/tests/test_utils.py
diox: 220e96b5df324a40abbb18a332e5cf293740a7a6 makes the test a little more realistic with the scenario we had.

src/olympia/reviewers/utils.py
diox: 220e96b5df324a40abbb18a332e5cf293740a7a6

src/olympia/reviewers/utils.py
```diff
@@ -1069,6 +1069,7 @@ def set_promoted(self, versions=None):
             # These addons shouldn't be be attempted for auto approval anyway,
             # but double check that the cron job isn't trying to approve it.
             assert not self.user.id == settings.TASK_USER_ID
+        if self.addon.promoted_groups(currently_approved=False):
```
eviljeff, line 1072: I can see you've already merged this, but

```suggestion
        if group:
```

diox: Doh missed this when reverting the previous part, I'll patch it.

PR #24830, opened 2026-05-05
Allow admins to include a soft-blocked version in their add/change blocklistsubmission

### LINKED ISSUE #16180: [Bug]: Delayed hard-block on top of a soft-block should hard-block the version when the delay expires

### What happened?

From https://github.com/mozilla/addons/issues/16117

Scenario is:
1. A version is soft-blocked without a delay
2. A hard-block with a delay is set on the same version

After the delay expires version should be hard-blocked

Examples:
on dev  https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/641597
on stage  https://reviewers.addons.allizom.org/en-US/reviewers/review/2253944

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/AMOENG-2451)

diox:
Hard-blocked versions remain read-only, they are only meant to be changed through a delete or soften action.

Fixes: mozilla/addons#16180

### DIFF
--- a/src/olympia/blocklist/forms.py
+++ b/src/olympia/blocklist/forms.py
@@ -198,7 +198,7 @@ class BlocklistSubmissionForm(AMOModelForm):
         """Return whether or not the given version should be available as a
         choice for the action we're currently doing."""
         conditions = {
-            BlocklistSubmission.ACTIONS.ADDCHANGE: not version.is_blocked,
+            BlocklistSubmission.ACTIONS.ADDCHANGE: not version.is_hard_blocked,
             BlocklistSubmission.ACTIONS.DELETE: version.is_blocked,
             BlocklistSubmission.ACTIONS.HARDEN: version.is_soft_blocked,
             BlocklistSubmission.ACTIONS.SOFTEN: version.is_hard_blocked,
--- a/src/olympia/blocklist/tests/test_admin.py
+++ b/src/olympia/blocklist/tests/test_admin.py
@@ -481,20 +481,16 @@ class TestBlocklistSubmissionAdmin(TestCase):
         doc = pq(response.content.decode('utf-8'))
         checkboxes = doc('input[name=changed_version_ids]')
 
-        assert len(checkboxes) == 4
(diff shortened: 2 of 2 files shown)


diox:
Originally missed this in https://github.com/mozilla/addons/issues/16117 but that was part of the intent stated in that issue: allow a `ADDCHANGE` with a hard-block block type to modify soft-blocks.

eviljeff:
> Originally missed this in [mozilla/addons#16117](https://github.com/mozilla/addons/issues/16117) but that was part of the intent stated in that issue: allow a `ADDCHANGE` with a hard-block block type to modify soft-blocks.

I thought this was what `ACTIONS.HARDEN` was for.  If ADDCHANGE can now harden too, do we need a more specific action that duplicates it?

diox:
We could do without it, but I think it helps when doing things in bulk:
- Version X and Y are soft-blocked already for whatever reason
- As an admin, you try to submit a new hard-block for version X, Y and Z
- You don't want to have to split your blocklist submission into 2 just because X and Y are already soft-blocked, you just want to submit your bulk-blocklistsubmission and have them all hard-blocked

eviljeff:
> We could do without it, but I think it helps when doing things in bulk:
> 
>     * Version X and Y are soft-blocked already for whatever reason
> 
>     * As an admin, you try to submit a new hard-block for version X, Y and Z
> 
>     * You don't want to have to split your blocklist submission into 2 just because X and Y are already soft-blocked, you just want to submit your bulk-blocklistsubmission and have them all hard-blocked

That reads like a justification for not having a duplicate HARDEN action.

diox:
Maybe... `HARDEN` has a couple advantages: it's more specific (it only "upgrades", doesn't allow new blocked versions), and it is straightforward to revert (basically just make a clone of the submission with the `SOFTEN` action), something we are taking advantage of in automated blocks on user ban.

I don't see the harm on having both available - it's already working under the hood, I'm just lifting an arbitrary restriction for humans in the admin.

eviljeff (approved):

PR #24838, opened 2026-05-07
Remove useless colon character in pull request template

diox:


diox:
Maybe I'm the only one who cares but I have always found that character annoying.

eviljeff (approved):
🤷

PR #24863, opened 2026-05-07
cache existing Blocks so reversing delayed block actions doesn't fail

### LINKED ISSUE #16204: [Bug]: reversing multiple follow-up block actions can fail
Follow-up from #16119 

### What happened?

https://mozilla.sentry.io/issues/7448848020/?query=is%3Aunresolved&referrer=issue-stream

```
(1452, 'Cannot add or update a child row: a foreign key constraint fails (`addons_allizom_org`.`log_activity_block`, CONSTRAINT `log_activity_block_block_id_6f5144af_fk_blocklist_block_id` FOREIGN KEY (`block_id`) REFERENCES `blocklist_block` (`id`) ON DELETE RESTRI)')
```

My theory is the `using` in `get_blocks_from_guids` means if we attempt to unblock twice in the same transaction, then we'll be trying to delete a block that was already deleted.  Which seems to be handed in the actual deletion code, but the logging function fails because the activity log is trying to fk to an already deleted block record.

We don't attempt to delete a block that is already deleted - or at least fail quietly.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/AMOENG-2481)

eviljeff:
Fixes mozilla/addons#16204

### Description

Works around the replica query for Blocks being out of date when we're in a transaction on the primary by running the query once and caching the results, so the same Block instances can be operated on/over.

### Context

I'm not a massive fan of this solution, but I think it works - the loop in `delete_versions_from_blocks` will skip over Block instances where the id is `None`, and that's one of things that happens when an instance is deleted.  I added a test but it's not really complete as it's still running against a single database in tests.

### Testing

- set up a policy in Cinder with multiple follow-up enforcement actions, both for the same block type and other types.
- report an add-on
- make a decision that a job in Cinder with that policy, and playback that decision payload locally
- edit the blocklistsubmission instances to remove their delays, so they expand into Block and BlockVersion instances immediately
- override the decision in Cinder, and playback the override decision
- all the blocked versions should be removed, along with the Block instance itself, with no failures.

### Checklist

### DIFF
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -19,7 +19,7 @@ from olympia.addons.models import Addon, AddonApprovalsCounter, AddonReviewerFla
 from olympia.amo.templatetags.jinja_helpers import absolutify
 from olympia.amo.utils import send_mail
 from olympia.bandwagon.models import Collection
-from olympia.blocklist.models import BlocklistSubmission, BlockType
+from olympia.blocklist.models import Block, BlocklistSubmission, BlockType
 from olympia.blocklist.utils import delete_versions_from_blocks, save_versions_to_blocks
 from olympia.constants.abuse import DECISION_ACTIONS
 from olympia.constants.permissions import ADDONS_HIGH_IMPACT_APPROVE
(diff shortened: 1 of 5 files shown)


diox (approved):
Seems like a reasonable approach actually, I like it.

PR #24871, opened 2026-05-11
Move auto_approval prevention in rejection code to the ContentAction

### LINKED ISSUE #16123: [Task]: Add ability for `ScannerRule` to execute enforcement actions from a `CinderPolicy`

diox:
Scanners performing a reject version action should prevent approval until the next manual approval just like if it was done in reviewer tools: This is not just consistent, it also prevents auto-approval of the very version we just rejected, since actions are ran in auto_approve.

Fixes https://github.com/mozilla/addons/issues/16123

### Context

With `run-action-in-auto-approve` set (`True` in dev/stage/prod) we run scanner workflow and enforcement actions in `auto_approve()`, after having fetched versions to auto-approve, but before determining if the version can actually be auto-approved.

This means that we should ensure actions that affect the version negatively also prevent its auto-approval, otherwise we can execute the action and then proceed with auto-approval normally, leading to inconsistent states.

Even without taking this scenario into consideration, rejecting a version should prevent auto-approval of the next submission regardless of the source of the rejection.

### Testing

See `test_run_actions_policy_force_disable()` and `test_run_actions_policy_reject()`:
- Make sure `run-action-in-auto-approve` waffle switch is enabled
- Set up a policy that either rejects or force-disable as the enforcement action, available in reviewer tools
- Set up a scanner rule that always matches (you can use yara with `a `condition: true`) with that policy 
- Submit a new version (Make sure `INITIAL_AUTO_APPROVAL_DELAY_FOR_LISTED` and `INITIAL_DELAY_FOR_UNLISTED` config keys are set to 0 depending on what channel you are using, or submit a new version of an existing approved add-on to not affect results - otherwise that delay will accidentally work around the issue by preventing auto-approval)
- Run `auto_approve`

Expected results:
- The add-on should be disabled if the enforcement action was force disable
- The version should be rejected and not be approved.

### DIFF
--- a/src/olympia/abuse/actions.py
+++ b/src/olympia/abuse/actions.py
@@ -397,7 +397,17 @@ class ContentActionDisableAddon(ContentAction):
             .order_by('-pk')
         )
 
+    def prevent_auto_approval(self):
+        AddonReviewerFlags.objects.update_or_create(
+            addon=self.target,
+            defaults={
+                'auto_approval_disabled': True,
+                'auto_approval_disabled_unlisted': True,
+            },
+        )
+
     def process_action(self, release_hold=False):
+        self.prevent_auto_approval()
         if self.target.status != amo.STATUS_DISABLED:
             # Set target_versions before executing the action, since the
             # queryset depends on the file statuses.
@@ -407,6 +417,7 @@ class ContentActionDisableAddon(ContentAction):
         return None
 
     def hold_action(self):
+        self.prevent_auto_approval()
         if self.target.status != amo.STATUS_DISABLED:
             self.decision.target_versions.set(self.versions_force_disable_will_affect)
             return self.log_action(amo.LOG.HELD_ACTION_FORCE_DISABLE)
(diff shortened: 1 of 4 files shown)


eviljeff (approved):
It's a little aggressive - we're pushing the next version for every add-on that is disabled, or has a version rejected, into manual review, even if the decision is overridden without it taking affect.  Plus an appeal wouldn't reverse it.

Maybe this is what we want... but it should probably be called out to Operations at least (I see the pr is addressing something QA commented on, so the justification for the change is buried in the issue comments)

diox:
I could work around the issue QA found in a different way - checking the status of the version and the add-on inside `AutoApprovalSummary` - but I figured this was consistent with current behavior for reviewer actions so it shouldn't be controversial. I'll talk with Operations folks about it.

PR #24885, opened 2026-05-12
Store automatic block reasons per version as constant

### LINKED ISSUE #16125: [Task]: Generate meaningful block reason when there are multiple submissions for a block
### Description

We have a single `Block` which has a single `reason` text field for a given `guid`. That reason is then exposed to end-users (and not translated, btw).

The block can be coming from multiple different blocklist submissions, with different versions being blocked at different times. On top of this, sometimes a block can be partially reverted, or soft-blocks can be hardened into hard-blocks, or the opposite.

Currently we overwrite the `reason` when saving a new submission, so the latest reason wins. With blocks being automated as part of bans and soon as part of follow-up actions after a reviewer selected a policy, we need a better way to generate the block reason.

### Acceptance Criteria

- [ ] item 1

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/AMOENG-2378)

eviljeff:
Fixes mozilla/addons#16125

### Description

Stores a reason constant (inelegantly named auto_block_reason because "reason" is an overused term) per BlockVersion, so the reason exposed to end users can be based on the current versions blocked, rather than a text string we have to keep up to date.

### Context

Initially I thought we could generate a fully descriptive reason, based on all the reasons, but the need to keep the frontend integration (that relies on deleted addons and versions having a specific reason string) means we can't do that - in cases where there are multiple reasons for versions being blocked, we need to return just one.  The numeric sorting won't scale if we add more reasons, but it works for now.

To allow admins still to be able to manually define a reason for a block, having one defined in Block.reason takes precedence. This would mean that only new blocks going forward would have their reasons dynamically updated - we could consider follow-up to attempt to migrate any existing blockreason to the new field, and clear the associated Block.reason field.

### Testing

- Delete a (signed) version of an add-on; see it blocked as usual
- check the admin and see the Block lists the reason constant next to the version numbers, but the Block.reason field is blank
- access the api endpoint, e.g. http://olympia.test/api/v5/blocklist/block/{guid} and see the reason string returned is "version deleted"
  - http://olympia.test/en-GB/firefox/blocked-addon/{guid} should display the more expansive explanation as now
- use django admin to update the auto_block_reason on the BlockVersion to `0`; see the api and frontend page display the text about the add-on being fraudulent, etc.
- set the Block.reason field to a string value; see that string value returned as the reason instead

### Checklist


diox (approved):
We'll likely need something more flexible but for now this works...

eviljeff:
> We'll likely need something more flexible but for now this works...

Can you file an issue with use-cases?

PR #24893, opened 2026-05-13
Prevent the creation of a new service account when a webhook scanner name is changed

### LINKED ISSUE #16215: [Bug]: Updating a webhook name shouldn't create a new service account

### What happened?

I edited the name of a webhook and a new service account was created for this webhook scanner.

Only the name changes but the service account doesn't.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/AMOENG-2503)

willdurand:
Fixes https://github.com/mozilla/addons/issues/16215

### DIFF
--- a/src/olympia/scanners/models.py
+++ b/src/olympia/scanners/models.py
@@ -332,19 +332,22 @@ class ScannerWebhook(ModelBase):
         db_table = 'scanners_webhooks'
 
     def save(self, *args, **kwargs):
-        service_account, created = UserProfile.objects.get_or_create_service_account(
-            name=f'webhook-{self.name}',
-            notes=(
-                'Service account automatically created for '
-                f'the "{self.name}" scanner webhook.'
-            ),
-        )
-        if created:
-            # Add service account to the scanner group.
-            group = Group.objects.get(name=SCANNER_SERVICE_ACCOUNTS_GROUP)
-            GroupUser.objects.get_or_create(group=group, user=service_account)
+        if not self.service_account:
+            service_account, created = (
+                UserProfile.objects.get_or_create_service_account(
+                    name=f'webhook-{self.name}',
+                    notes=(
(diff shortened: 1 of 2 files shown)


src/olympia/scanners/models.py
```diff
@@ -332,19 +332,22 @@ class Meta:
         db_table = 'scanners_webhooks'
 
     def save(self, *args, **kwargs):
```
eviljeff, line 334: These changes would work... but, imo, with the aim of leaving good code smell this needs a little more of a rewrite:

If there isn't an existing `service_account` then it's always a create, so we don't need an underlying get_or_create call, just a create, in `get_or_create_service_account` (and I'm not really sure there's a need for a specific queryset manager function either when it's only called in this one place, but meh)

src/olympia/scanners/tests/test_models.py
```diff
@@ -793,3 +793,18 @@ def test_save_creates_a_service_account(self):
...
+            api_key='secret',
+        )
+        webhook.save()
+        original_account = webhook.service_account
+        assert original_account is not None
+
+        webhook.name = 'some-new-name'
+        webhook.save()
+
+        assert webhook.service_account == original_account
```
eviljeff, line 812: strictly, if it's checking for the account not being recreated, it should check for object instances too - i.e. there are no new UserProfile instances created on the rename.

eviljeff (approved):

src/olympia/scanners/models.py
diox: You could have the FK unlinked but the user profile existing still for some reason if the scanner instance was manipulated through shell or whatever.

On that note, while we don't expose that to the admin etc, I've used `get_or_create_service_account()` recently to create the service account for the Zendesk integration - they are not specific to scanners.

PR #24914, opened 2026-05-19
Start services earlier in CI to improve speed

### LINKED ISSUE #16074: [Task]: Start services earlier in CI to be faster
### Description

https://github.com/mozilla/addons/issues/16061 introduced a new GitHub action to run tests in that doesn't spin up a full development environment.

Each job is waiting for MySQL to be healthy through `depends_on`, not as a separate step, so we don't see how long that takes. That may be an area to look into.

On top of that there is another 30 seconds to 1 minute to gain by being more clever with starting `autograph`: at the moment we are starting it for each job, separately, waiting for it to be healthy, which takes ~25 seconds. Given very few tests need autograph, we should look into marking them and running them in their own dedicated job, and stop starting autograph everywhere else.

### Acceptance Criteria

- [ ] item 1

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/AMOENG-2312)

diox:
Fixes mozilla/addons#16074

- Remove obsolete and unused internal routes allowed setting & test marker
- Remove obsolete `needs_locale_compilation`: we are building locales when building the `web` image anyway...
- Split out js tests that were hidden inside the `locales` job for some reason
- Start MySQL, memcached and redis in detached mode early, without waiting for them to be healthy.
  - Wait for them as you'd expect later through the compose file (so that check is done after building the image for the `web` container, by the time this happens the services are healthy)
- Rename `es_tests` marker to `requires_elasticsearch`, and introduce `requires_autograph` that does what it says on the tin. Split out tests requiring autograph to their own job, and stop spinning up an `autograph` container for all the other jobs.

For jobs inside `test_main` on a PR from an outside repos, this saves around 30 seconds on both average and median times, going from around 7 minutes to 6 and a half. Some jobs remain slow, especially outside `test_main`, and we could improve things further with a better split, but that should be good enough for now.

### DIFF
--- a/Makefile-docker
+++ b/Makefile-docker
@@ -95,7 +95,8 @@ initialize: ## ensure database exists
 PYTEST_SRC := src/olympia/
 
 .PHONY: test_js
-test_js: run_js_tests
+test_js:
+	npm run test tests/js
 
 .PHONY: test_main
 test_main:
@@ -144,10 +145,6 @@ test_failed: ## rerun the failed tests from the previous run
 	$(PYTEST_SRC) \
 	--lf \
 	$(ARGS)
-
-.PHONY: run_js_tests
-run_js_tests: ## Run the JavaScript test suite (requires compiled/compressed assets).
-	npm run test tests/js
 
 .PHONY: watch_js_tests
 watch_js_tests: ## Run+watch the JavaScript test suite (requires compiled/compressed assets).
--- a/Makefile-docker
+++ b/Makefile-docker
@@ -105,18 +105,7 @@ test_needs_locales_compilation:
 test_main:
 	pytest $(PYTEST_SRC) \
 		-n auto \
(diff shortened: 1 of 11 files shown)


Makefile-docker
```diff
@@ -94,36 +94,28 @@ initialize: ## ensure database exists
...
 PYTEST_SRC := src/olympia/
 
-.PHONY: test_needs_locales_compilation
-test_needs_locales_compilation:
-	pytest $(PYTEST_SRC) \
-		-m 'needs_locales_compilation' \
-		$(ARGS)
-	npm run test tests/js/
+.PHONY: test_js
+test_js: run_js_tests
```
eviljeff: do we need this? Seems redundant

eviljeff (approved):

Makefile-docker
diox: Agreed, if you don't use `run_js_tests` we can remove it

PR #24918, opened 2026-05-20
Enable policy->action selection in reviewer tools, for negative policies

### LINKED ISSUE #16122: [Task]: Have reviewers select a policy rather than an action
### Description

Depends on https://github.com/mozilla/addons/issues/16121

Change reviewer tools UI to have reviewer select a policy rather than an action, deriving the action(s) automatically from the policy depending on how the policy is configured, similar to how Cinder moderators select a policy rather than an action.

The relevant `enforcement_actions` from the policy should be executed with the decision.

This should be behind a dedicated waffle switch.

Could be split into a spike to investigate what the UI would look like first.

Challenges: 
- How to select whether the reviewer is acting on the add-on or the version(s) ?
- How to deal with add-on/version state checks ?
- How to deal with reverting an earlier action (with or without appeals) in the UI ?

### Acceptance Criteria

- [ ] item 1

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/AMOENG-2374)

eviljeff:
Fixes mozilla/addons#16122

### Description

Support policy selections directly in reviewer tools, which create decisions based on the enforcement actions of the policies selected.  In this (first) patch, only some reviewer tool actions are replaced (reject latest version, and disable addon)

Screenshot 2026-05-21 140053

### Context

Follow-up pieces for this:
- non-negative enforcement actions https://github.com/mozilla/addons/issues/16239
- version selection https://github.com/mozilla/addons/issues/16240
- appeal jobs https://github.com/mozilla/addons/issues/16241
- overriding decisions. https://github.com/mozilla/addons/issues/16242

### Testing

- enable waffle switch
- see force disable and reject version are unavailable, but new action replaces them
- sync cinder policies, and enable some in the reviewer tools
  - you may need to update some (in Cinder and sync, or with django shell) to get a good selection of enforcement actions.
- the "negative" policies should be available for selection
- selecting multiple policies should display on page what enforcement action(s) will take place
- the same enforcement should happen when you submit the form

### Checklist


src/olympia/reviewers/templatetags/jinja_helpers.py
```diff
@@ -83,3 +83,13 @@ def activity_changes(activity):
             'added': data.get('added', []),
         }
     )
+
+
+@library.global_function
+@jinja2.pass_context
+def toggle_when_action(context, definition_key, *, default=False):
```
eviljeff, line 90: I added this because I was bored of constructing extra `action_xxx` lists each time we needed to toggle on something in a single place.  It didn't scale well.

eviljeff:
Mmm, this needs more work - I overlooked that the reject_version action, that it replaces, still has some logic to clear needshumanreview we don't do in ContentActionRejectVersion.  We either need to add in some special casing in the `review_with_policy` method, or elsewhere in the ReviewHandler class, or migrate it over to ContentActionRejectVersion.

It's something that's going to trip up the expansion to multiple versions and non-negative enforcment actions too - that even after we implemented reviewer tools using ContentAction there's still some logic that is reviewer tools specific that stayed.

diox:
> Mmm, this needs more work - I overlooked that the reject_version action, that it replaces, still has some logic to clear needshumanreview we don't do in ContentActionRejectVersion. We either need to add in some special casing in the `review_with_policy` method, or elsewhere in the ReviewHandler class, or migrate it over to ContentActionRejectVersion.

I think it would make sense to migrate that logic over to `ContentActionRejectVersion` as well, if it's a human review... (and I guess there might be some special casing regarding when the NHR is cleared when something is held for 2nd level approval to pay attention to)

diox:
Note to myself: waffle switch for this PR is called `enable-policy-review-selection`.

eviljeff:
> Note to myself: waffle switch for this PR is called `enable-policy-review-selection`.

better naming welcome

diox:
The naming is fine, I just needed to write it down somewhere to remember later without having to search in the code

eviljeff:
> > Mmm, this needs more work - I overlooked that the reject_version action, that it replaces, still has some logic to clear needshumanreview we don't do in ContentActionRejectVersion. We either need to add in some special casing in the `review_with_policy` method, or elsewhere in the ReviewHandler class, or migrate it over to ContentActionRejectVersion.
> 
> I think it would make sense to migrate that logic over to `ContentActionRejectVersion` as well, if it's a human review... (and I guess there might be some special casing regarding when the NHR is cleared when something is held for 2nd level approval to pay attention to)

Moved it to ContentActionRejectVersion (and removed it from reject_latest_version and reject_multiple_versions, that use ContentActionRejectionVersion under the hood).  

+ added a migration to create the waffle switch

src/olympia/reviewers/forms.py
```diff
@@ -601,25 +610,49 @@ def clean(self):
                 for job in self.cleaned_data.get('cinder_jobs_to_resolve', ())
                 if not job.is_developer_appeal
             ]
-        if self.cleaned_data.get('cinder_jobs_to_resolve') and self.cleaned_data.get(
-            'cinder_policies'
-        ):
-            actions = CinderPolicy.get_decision_actions_from_policies(
-                self.cleaned_data.get('cinder_policies'),
-                for_entity=Addon,
```
diox, line 609: With the waffle switch off, previously I could reject a version with any combination of policies when not resolving a job at the same time. Now, with the waffle still off, I can't if their enforcement actions differ: I'm getting `Multiple policies selected with different cinder actions.` error.

diox (changes requested):

diox:
@wagnerand-moz can you let us know if the proposed UI in this PR suits you ? (see screenshot) ; it's going to be behind a waffle switch anyway.

wagnerand-moz:
> @wagnerand-moz can you let us know if the proposed UI in this PR suits you ? (see screenshot) ; it's going to be behind a waffle switch anyway.

This looks good overall, thanks.

One nit we could optimize (here or in a follow-up, leaving that up to you) is the "Enforcement actions" section. I am not sure the screenshot above has a made-up copy, but it looks like (from reading the labels), the that long-term action is mentioned before the mid-term. Can we order them accurately, and ideally show as a sequence? Perhaps a numbered list (+ time indication if we have it) could make sense, e.g.

Enforcement actions:
1. Now: Add-on disable
2. 2026-07-01: Soft-block
3. 2026-08-01: Hard block

Wdyt about that idea?

eviljeff:
> One nit we could optimize (here or in a follow-up, leaving that up to you) is the "Enforcement actions" section. I am not sure the screenshot above has a made-up copy, but it looks like (from reading the labels), the that long-term action is mentioned before the mid-term. Can we order them accurately, and ideally show as a sequence? Perhaps a numbered list (+ time indication if we have it) could make sense, e.g.

It's currently a list of actions, with the primary action (the Disable) first, and in bold, then the follow-up actions in no particular order (actually just the order of the list in the policy in Cinder)
 
> Enforcement actions:
> 
>     1. Now: Add-on disable
> 
>     2. 2026-07-01: Soft-block
> 
>     3. 2026-08-01: Hard block
> 
> 
> Wdyt about that idea?

Showing the number days delay rather than a date is more accurate because a hold for 2nd level approval would delay the enforcement.  But can do either.

wagnerand-moz:
> Showing the number days delay rather than a date is more accurate because a hold for 2nd level approval would delay the enforcement. But can do either.

I don't have a strong preference on that. In a perfect world we could show both, because depending on the reader and the specific case, you want to either see one or the other, but we can also start with folks looking at their calendar, and expand later if there is strong need.

src/olympia/reviewers/forms.py
eviljeff: Okay, I was trying to cover an edge case with the waffle switch off where if you select two policies with different enforcement actions the outcome is indeterminate, though I think it can only happen when resolving reports anyway (because most of the reviewer actions override the enforcement action)

eviljeff:
fwiw, I need to change how enforcement actions are rendered in https://github.com/mozilla/addons/issues/16240 so I'll do it that.

diox (approved):

eviljeff:
> fwiw, I need to change how enforcement actions are rendered in [mozilla/addons#16240](https://github.com/mozilla/addons/issues/16240) so I'll do it that.

Actually, this will have to be a separate issue - it's a significant departure from the current rendering, which is just the label attached to the constant.  i.e. there is no way to translate a constant `AMO_FU_DELAY_SHORT_SOFT_BLOCK_ADDON`, with a label `'[Follow-up] Short-delayed soft block'` to `"2026-07-01: Soft-block"`

PR #24931, opened 2026-05-20
Expose policies in devhub review history

### LINKED ISSUE #16232: [Bug]: Review history in devhub misses policy text

### What happened?

`cinder_policy_review_reasons_enabled` switched the policy text/reasons from being stored in `comments` to `policy_texts` in `ActivityLog.details`.

However, the review history in devhub is not yet aware of this change, looking at `comments` only.

STR:
1. Receive a rejection for your add-on.
2. On AMO, open the _Review History_ (under _Manage Status and Versions_) in the developer hub for the add-on. The policy items for which the submission was rejected are not shown. only _Comments_ are shown, which is the field that used to hold the policy text previously.
before

Both fields should be shown, in this order:
1) `policy_texts`
2) `comments`

Screenshot 2026-05-22 at 12 19 26

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/AMOENG-2523)

wagnerand-moz:
Fixes mozilla/addons#16232

### Description

Adds the policies (`policy_texts` in `ActivityLog.details`) to the review history.

Before:
before

After:
Screenshot 2026-05-22 at 12 19 26

### Context

When the waffle switch `cinder_policy_review_reasons_enabled` is enabled, the policies the reviewer selected are stored in `policy_texts`, rather than `comments`. That field needs to be exposed in the developer hub as well so developers can see the policies for which their submission was rejected.

### Testing

1. Submit an add-on.
2. Reject it, ideally with at least two policies and an additional note added to "Comments" in the reviewer tools.
3. Check the _Review History_ for that version in devhub. Make sure it contains both the selected policies and the note from "Comments".

### Checklist

#### Initial "After" screenshot

after

### DIFF
--- a/src/olympia/activity/serializers.py
+++ b/src/olympia/activity/serializers.py
@@ -40,11 +40,11 @@ class ActivityLogSerializer(AMOModelSerializer):
         self.to_highlight = kwargs.get('context', {}).get('to_highlight', [])
 
     def get_policies(self, obj):
-        # Some activity logs might not have `policy_texts`
-        policies = obj.details.get('policy_texts', []) if obj.details else []
         sanitize = getattr(obj.log(), 'sanitize', None)
         if sanitize is not None:
             return [sanitize]
+        # Some activity logs might not have `policy_texts`
+        policies = obj.details.get('policy_texts', []) if obj.details else []
         return policies
 
     def get_comments(self, obj):
--- a/src/olympia/activity/serializers.py
+++ b/src/olympia/activity/serializers.py
@@ -12,6 +12,7 @@ from olympia.api.utils import is_gate_active
 class ActivityLogSerializer(AMOModelSerializer):
     action = serializers.SerializerMethodField()
     action_label = serializers.SerializerMethodField()
+    policies = serializers.SerializerMethodField()
     comments = serializers.SerializerMethodField()
     date = serializers.DateTimeField(source='created')
     user = serializers.SerializerMethodField()
@@ -25,6 +26,7 @@ class ActivityLogSerializer(AMOModelSerializer):
             'id',
             'action',
             'action_label',
+            'policies',
             'comments',
             'user',
             'date',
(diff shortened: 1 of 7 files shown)


src/olympia/activity/serializers.py
```diff
@@ -37,6 +39,14 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.to_highlight = kwargs.get('context', {}).get('to_highlight', [])
 
+    def get_policies(self, obj):
+        # Some activity logs might not have `policy_texts`
+        policies = obj.details.get('policy_texts', []) if obj.details else []
+        sanitize = getattr(obj.log(), 'sanitize', None)
+        if sanitize is not None:
+            return [sanitize]
```
eviljeff: nit, the ordering of this means we still do the redundant getting of details, etc, even when the response is sanitized


```suggestion
        sanitize = getattr(obj.log(), 'sanitize', None)
        if sanitize is not None:
            return [sanitize]
        # Some activity logs might not have `policy_texts`
        policies = obj.details.get('policy_texts', []) if obj.details else []
```

src/olympia/devhub/templates/devhub/versions/list.html
```diff
@@ -124,6 +124,7 @@
                   <div>
                     <p><strong class="action">$action_label</strong> {{ _('by') }}
                     <em class="user_name">$user_name</em> <time class="timeago" datetime="$date">$date</time></p>
+                    <ul class="review-entry-policies"></ul>
```
eviljeff, line 127: I think this needs a little more to make it clear what the text is - for example in the email template we do:
```
Our review found that your content violates the following Mozilla policy or policies:
    - {{ policytext }}

{{ comments }}
```
i.e. an introduction line, each policy text with a bullet, and then an empty line before the comments.  That's probably too much for this view, but the result in the after screenshot is a little too simple, imo.

In the reviewer tools we use a different font color, and a line space, to distinguish the policy texts from comments too.

eviljeff (approved):
r+wc 
I think the UX needs some additional work, but I'll leave it to you to decide if that's done before merging this pr, or as a follow-up.

src/olympia/activity/serializers.py
wagnerand-moz: Thanks!

src/olympia/devhub/templates/devhub/versions/list.html
wagnerand-moz: Thanks, fixed in https://github.com/mozilla/addons-server/pull/24931/changes/c317e9ae05503e527fd9a2f79bac805a5f6673b0.

PR #24939, opened 2026-05-22
Add bulk banning functionality to the admin

### LINKED ISSUE #15580: [Bug]: 502 error on django admin for large requests
### What happened?

For large requests, e.g. banning a lot of users, a 502 error "Bad gateway" is thrown, but the ban is actually successful (as far as I can see).

This happens on any admin page where you submit a large request.

Request should go through without errors.

### LINKED ISSUE #16224: [Task]: Allow Operations to mass ban users

diox:
Also make banning through the changelist action use the same async task underneath to scale to a larger number of users.

Fixes https://github.com/mozilla/addons/issues/16224
Fixes https://github.com/mozilla/addons/issues/15580

### DIFF
--- a/src/olympia/users/forms.py
+++ b/src/olympia/users/forms.py
@@ -73,4 +73,4 @@ class BulkBanForm(forms.Form):
         user_ids = {id_ for id_ in data.splitlines() if id_.isdigit()}
         if not user_ids:
             raise forms.ValidationError('This field must contain a least one user id')
-        return user_ids
+        return sorted(user_ids)
--- a/src/olympia/users/forms.py
+++ b/src/olympia/users/forms.py
@@ -61,3 +61,16 @@ class IPNetworkUserRestrictionForm(AMOModelForm):
             data['network'] = f'{ip_address}/32'
 
         return data
+
+
+class BulkBanForm(forms.Form):
+    user_ids = forms.CharField(
+        label='', required=True, widget=forms.Textarea(attrs={'rows': 30, 'cols': 80})
+    )
+
+    def clean_user_ids(self):
+        data = self.cleaned_data.get('user_ids', '')
+        user_ids = {id_ for id_ in data.splitlines() if id_.isdigit()}
+        if not user_ids:
+            raise forms.ValidationError('This field must contain a least one user id')
+        return user_ids
--- a/src/olympia/users/forms.py
+++ b/src/olympia/users/forms.py
@@ -70,7 +70,7 @@ class BulkBanForm(forms.Form):
 
     def clean_user_ids(self):
         data = self.cleaned_data.get('user_ids', '')
-        user_ids = {id_ for id_ in data.splitlines() if id_.isdigit()}
+        user_ids = {id_ for id_ in data.strip().splitlines() if id_.isdigit()}
         if not user_ids:
             raise forms.ValidationError('This field must contain a least one user id')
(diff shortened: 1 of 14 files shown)


src/olympia/users/forms.py
```diff
@@ -61,3 +61,16 @@ def clean(self):
...
+
+
+class BulkBanForm(forms.Form):
+    user_ids = forms.CharField(
+        label='', required=True, widget=forms.Textarea(attrs={'rows': 30, 'cols': 80})
+    )
+
+    def clean_user_ids(self):
+        data = self.cleaned_data.get('user_ids', '')
+        user_ids = {id_ for id_ in data.splitlines() if id_.isdigit()}
```
eviljeff: We probably want to trim here too - a trailing space makes the number no longer `isdigit() is True`.

src/olympia/users/forms.py
```diff
@@ -61,3 +61,16 @@ def clean(self):
...
+        if not user_ids:
+            raise forms.ValidationError('This field must contain a least one user id')
```
eviljeff, line 77: I _do_ wonder if the form shouldn't raise on bad input rather than just silently ignore.  I don't see a use case of Operations pasting in content with a mix of valid user ids and other random text they want to be ignored - accidentaly adding an extra character on the same line as a user id seems a more common case, that they would want to know about and correct.

eviljeff (commented):
I expected there to be ADMIN_USER_BANNED entries in activitylog but there wasn't - potentially because it's being run in the task?

Also some comments about cleaning/validation.

diox:
> I expected there to be ADMIN_USER_BANNED entries in activitylog but there wasn't - potentially because it's being run in the task?

Oh, that's a good catch. I need to pass the user_id of the user who initiated the bulk ban to the task, and do `core.set_user()` so that the `ActivityLog` is recorded properly.

src/olympia/users/forms.py
diox: My reasoning was, it could be coming from a csv with headers and we should ignore those, but for the rest, the confirmation that shows how many users we found should be enough (Operations folks can ensure what they paste has unique ids to compare the number shown in the confirmation page with the number of lines from their input)

src/olympia/users/forms.py
diox: https://github.com/mozilla/addons-server/pull/24939/commits/c724247d6f7333025d7ec3efa7e2cb5053cedc44 and https://github.com/mozilla/addons-server/pull/24939/commits/aac0dccdd06aede7df3d6bd00142924fd812c9df

eviljeff (approved):

PR #24946, opened 2026-05-25
Add more packages to the google group in dependabot

diox:


eviljeff (approved):
I can't say I've ever noticed any cross-incompatibilities between these deps and google deps but 🤷

diox:
I'm mostly doing this to avoid merging those independently - they are often released together, or their updates only really make sense in context.

For example, see https://github.com/mozilla/addons-server/pull/24897 and https://github.com/mozilla/addons-server/pull/24899 : both released at the same time, cut from the same repos. Having separate PRs for these is a bit pointless.

